PR #538, opened 2022-02-25
`tcptracer`: Add guide and integration test

### LINKED ISSUE #463: [RFE] Create `tcptracer` guide and integration tests
## Current situation

Guide and integration tests are missing for the `tcptracer` gadget.

## Impact

Users do not have examples of usage for this gadget and it is not covered by the integration tests.

## Notes

Please remember to update the README with the link to the guide.

eiffel-fl:
Hi.

This PR adds a guide and an integration test for `tcptracer`.
It fixes #463.

Best regards.

### DIFF
--- a/docs/guides/tcptracer.md
+++ b/docs/guides/tcptracer.md
@@ -1,0 +1,116 @@
+---
+title: 'The "tcptracer" gadget'
+weight: 10
+---
+
+The `tcptracer` gadget is used to monitor tcp connections.
+
+## How to use it?
+
+First, we need to create one pod:
+
+```bash
+$ kubectl run busybox --image busybox:latest sleep inf
+pod/busybox created
+```
+
+You can now use the gadget, but output will be empty:
+
+```bash
+$ kubectl gadget tcptracer
+NODE             NAMESPACE        POD              CONTAINER        KPID   KCOMM            PAGES  TPID             TCOMM
+```
+
+Indeed, it is waiting for TCP connection to be established in the `default` namespace (you can use `-A` to monitor all namespaces and then be sure to not miss any event).
+So, in *an other terminal*, `exec` a container and run this `wget`:
+
+```bash
+$ kubectl exec -ti busybox -- wget https://www.kinvolk.io
+Connecting to www.kinvolk.io (188.114.97.3:443)
+wget: note: TLS certificate validation not implemented
+saving to 'index.html'
+index.html           100% |*************************************************************************************************| 42627  0:00:00 ETA
+'index.html' saved
+```
+
+Go back to *the first terminal* and see:
+
+```bash
+NODE             NAMESPACE        POD              CONTAINER        T PID    COMM             IP  SADDR            DADDR            SPORT   DPORT
+minikube         <>               <>               <>               C 16266  wget             4   172.17.0.3       188.114.97.3     34878   443
+```
+
+The printed lined correspond to TCP connection established with the socket.
+Here is the full legend of all the fields:
+
+* `T`: How the TCP connection was established, it can be one of the following values:
+	* `C`: The TCP connection was established after a `connect()` system call.
+	* `A`: The TCP connection was established after an `accept()` system call.
+	* `X`: The TCP connection was closed following the `close()` system call.
+	* `U`: The TCP connection was either established or closed following an unknown reason.
+* `PID`: The PID which established the TCP connection.
+* `COMM`: The command corresponding to the PID.
+* `IP`: The IP version (either 4 or 6).
+* `SADDR`: The source IP address.
+* `DADDR`: The destination IP address.
+* `SPORT`: The source port.
+* `DPORT`: The destination port.
+
+So, the above line should be read like this: "Command `wget`, which has PID 19981, established a TCP connection through IP version 4, using the `connect()` system call, from address 172.17.0.3 and port 16266 towards address 188.114.97.3 and port 433"
+
+Note that, IP 188.114.97.3 corresponds to `kinvolk.io` while port 443 is the port generally used for HTTPS.
+
+## Only print some information
+
+You can restrain the information printed using `-o custom-columns=column0,...,columnN`.
+This command will only show the PID and command:
+
+```bash
+$ kubectl gadget tcptracer -A -o custom-columns=pid,comm
+PID    COMM
+28489  wget
+```
+
+The following command is the same as default printing:
+
+```bash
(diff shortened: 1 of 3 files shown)


docs/guides/tcptracer.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+The printed lined correspond to TCP connection established with the socket.
+Here is the full legend of all the fields:
+
+* `T`: How the TCP connection was established, it can be one of the following values:
+	* `C`: The TCP connection was established after a `connect()` system call.
+	* `A`: The TCP connection was established after an `accept()` system call.
+	* `X`: The TCP connection was closed following the `close()` system call.
+	* `U`: The TCP connection was either established or closed following and unknown reasons.
```
mauriciovasquezbernal, line 49:
```suggestion
	* `U`: The TCP connection was either established or closed following an unknown reason.
```

docs/guides/tcptracer.md
```diff
@@ -0,0 +1,115 @@
...
+wget: note: TLS certificate validation not implemented
+wget: server returned error: HTTP/1.1 403 Forbidden
+command terminated with exit code 1
+```
+
+Go back to *the first terminal* and see:
+
+```bash
+NODE             NAMESPACE        POD              CONTAINER        T PID    COMM             IP  SADDR            DADDR            SPORT   DPORT
+minikube         <>               <>               <>               C 19981  wget             4   172.17.0.3       95.101.225.221   34434   443
```
mauriciovasquezbernal, line 39: What about also printing the line when the connection is closed?

docs/guides/tcptracer.md
```diff
@@ -0,0 +1,115 @@
...
+
+Indeed, it is waiting for TCP connection to be established in the `default` namespace (you can use `-A` to monitor all namespaces and then be sure to not miss any event).
+So, in *an other terminal*, `exec` a container and run this `wget`:
+
+```bash
+$ kubectl exec -ti busybox -- wget https://www.microsoft.com
+Connecting to www.microsoft.com (95.101.225.221:443)
+wget: note: TLS certificate validation not implemented
+wget: server returned error: HTTP/1.1 403 Forbidden
+command terminated with exit code 1
```
blanquicet, line 32: I would prefer to use a command that does not fail.

docs/guides/tcptracer.md
```diff
@@ -0,0 +1,115 @@
...
+
+The printed lined correspond to TCP connection established with the socket.
+Here is the full legend of all the fields:
+
+* `T`: How the TCP connection was established, it can be one of the following values:
+	* `C`: The TCP connection was established after a `connect()` system call.
+	* `A`: The TCP connection was established after an `accept()` system call.
+	* `X`: The TCP connection was closed following the `close()` system call.
+	* `U`: The TCP connection was either established or closed following and unknown reasons.
+* `PID`: The PID which calls `connect`.
```
blanquicet, line 50: This field is also valid for `accept()` and `close()`.

docs/guides/tcptracer.md
```diff
@@ -0,0 +1,115 @@
...
+* `COMM`: The command corresponding to the PID.
+* `IP`: The IP version (either 4 or 6).
+* `SADDR`: The sender IP address.
+* `DADDR`: The destination IP address.
+* `SPORT`: The sender port.
+* `DPORT`: The destination port.
+
+So, the above line should be read like this: "Command `wget`, which has PID 19981, established a TCP connection, using the `connect()` system call, from the IP version 4 IP 172.17.0.3 and port 34434 towards IP 95.101.255.221 and port 433"
```
blanquicet, line 58:
```suggestion
So, the above line should be read like this: "Command `wget`, which has PID 19981, established a TCP connection through IP version 4, using the `connect()` system call, from the IP 172.17.0.3 and port 34434 towards IP 95.101.255.221 and port 433".
```

docs/guides/tcptracer.md
```diff
@@ -0,0 +1,115 @@
+---
+title: 'The "tcptracer" gadget'
+weight: 10
+---
+
+The `tcptracer` gadget is used to monitor tcp connections.
+
+## How to use it?
+
+First, we need to create one pod with limited amount of memory:
```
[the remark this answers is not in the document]
eiffel-fl: Indeed, from `oomkill` guide :sweat_smile:

docs/guides/tcptracer.md
eiffel-fl: I do not understand your comment.
I am only getting one line when I run `wget`.

blanquicet: I think you get one line because the connection does not success. Now that you are using `kinvolk.io` you should get another entry for the `close()`.

eiffel-fl: I tested with `kinvolk.io` but I am still getting only one line.
Maybe it is because I use `podinformer`.

mauriciovasquezbernal: It's not related with podinformer at all. It's something with the `wget` implementation of busybox. I tried other images and two lines are printed.

eiffel-fl: Thank you for your test, let's stick with this output.

docs/guides/tcptracer.md
```diff
@@ -0,0 +1,116 @@
...
+* `IP`: The IP version (either 4 or 6).
+* `SADDR`: The source IP address.
+* `DADDR`: The destination IP address.
+* `SPORT`: The source port.
+* `DPORT`: The destination port.
+
+So, the above line should be read like this: "Command `wget`, which has PID 19981, established a TCP connection through IP version 4, using the `connect()` system call, from address 172.17.0.3 and port 34434 towards address 95.101.255.221 and port 433"
+
+Note that, IP 95.101.255.221 corresponds to `microsoft.com` while port 443 is the port generally used for HTTPS.
+
```
blanquicet, line 62: Could you please update it?

docs/guides/tcptracer.md
blanquicet: IMO, the container image should be changed to be able to show the `close()` entry in the output which will contain the Kubernetes metadata. I think it would improve the guide because with the current example you are hitting https://github.com/kinvolk/inspektor-gadget/issues/550 and no Kubernetes metadata is shown at all. WDYT?

eiffel-fl: I do not really see the point.
The important thing is that user gets the same output when they test than that of the guide.
So, as long as they test with `busybox` they will only get one line for `connect()`.

Regarding #550, we can either wait for it to be solved and I will update this guide or we can update the guide in #550.

blanquicet (approved):
LGTM!

eiffel-fl:
Thank you for your reviews!

Issue #541, opened 2022-03-02
grpc.WithDialer is deprecated, we should use WithContextDialer instead


blanquicet:
## Description

CI job `lint` is failing with the following error:

```
/inspektor-gadget/pkg/gadgettracermanager/containerutils/containerd/containerd.go:43:3 SA1019: grpc.WithDialer is deprecated: use WithContextDialer instead.  Will be supported throughout 1.x. (staticcheck)
/inspektor-gadget/pkg/gadgettracermanager/containerutils/crio/crio.go:44:3 SA1019: grpc.WithDialer is deprecated: use WithContextDialer instead.  Will be supported throughout 1.x. (staticcheck)
```

## Impact

CI is not passing and a specific exception was added to skip these errors, see `.golangci.yml` file.

## Environment and steps to reproduce

By removing the following exception from `.golangci.yml` file:

```yaml
    - text: "SA1019: grpc.WithDialer is deprecated: use WithContextDialer instead."
      linters:
        - staticcheck
```

The following errors will compare in the CI:
```
/inspektor-gadget/pkg/gadgettracermanager/containerutils/containerd/containerd.go:43:3 SA1019: grpc.WithDialer is deprecated: use WithContextDialer instead.  Will be supported throughout 1.x. (staticcheck)
/inspektor-gadget/pkg/gadgettracermanager/containerutils/crio/crio.go:44:3 SA1019: grpc.WithDialer is deprecated: use WithContextDialer instead.  Will be supported throughout 1.x. (staticcheck)
```

## Expected behavior

No errors are reported.

Issue #540, opened 2022-03-02
Invalid signal trap is reported by Go linter


blanquicet:
## Description

CI job `lint` is failing with the following error:

```
/inspektor-gadget/cmd/kubectl-gadget/utils/trace.go:588:117 SA1016: syscall.SIGKILL cannot be trapped (did you mean syscall.SIGTERM?) (staticcheck)
/inspektor-gadget/gadget-container/gadgets/networkpolicyadvisor/main.go:216:35 SA1016: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (staticcheck)
```

## Impact

CI is not passing and a specific exception was added to skip these two errors, see `.golangci.yml` file.

## Environment and steps to reproduce

By remove the following exception from `.golangci.yml` file:

    - text: "SA1016: (syscall.SIGKILL|os.Kill) cannot be trapped \\(did you mean syscall.SIGTERM\\?\\)"
      linters:
       - staticcheck

The following errors will compare in the CI:
```
/inspektor-gadget/cmd/kubectl-gadget/utils/trace.go:588:117 SA1016: syscall.SIGKILL cannot be trapped (did you mean syscall.SIGTERM?) (staticcheck)
/inspektor-gadget/gadget-container/gadgets/networkpolicyadvisor/main.go:216:35 SA1016: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (staticcheck)
```

## Expected behavior

No errors are reported.

PR #542, opened 2022-03-02
CI: Add go linters stylecheck, staticcheck and errorlint

blanquicet:
# Add go linters stylecheck, staticcheck and errorlint

The aim of this PR is to uniform the code style and speed up PR reviews. To do so, this PR adds three Go linters: stylecheck, staticcheck and errorlint to automate the work.

Regarding `errorlint`, check https://github.com/polyfloyd/go-errorlint for further details. There are three main checks:
- Use `%w` to format errors in `fmt.Errorf()`
- Comparison of errors is done using `errors.Is()`
- Type assertions of errors is done using `errors.As()`

Regarding `stylecheck` and `staticcheck`, check https://staticcheck.io/docs/checks/ for further details. Some examples of checks:
- ALL_CAPS should not be used in Go names; use CamelCase instead
- Underscores should not be used in Go names
- Error strings should not be capitalized
- Error strings should not end with punctuation or a newline
- Keywords like ID and JSON on variables and function names should be in uppercase, e.g. var `tracerId` should be `tracerID`.

## Warning

Only two errors were not fixed and were skipped by adding exceptions. They will be managed in separate PRs, see issues #540 and #541.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -75,9 +75,9 @@ jobs:
       run: |
         make -C gadget-container ebpf-objects
     - name: Lint
-      uses: golangci/golangci-lint-action@v2
+      uses: golangci/golangci-lint-action@v3.1.0
       with:
-        version: v1.43.0
+        version: v1.44.2
         working-directory: /home/runner/work/inspektor-gadget/inspektor-gadget
         # Workaround to display the output:
         # https://github.com/golangci/golangci-lint-action/issues/119#issuecomment-981090648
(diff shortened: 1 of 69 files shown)


mauriciovasquezbernal (approved):
I went through the changes quickly. I think it's fine to go as the changes are just style ones.

mauriciovasquezbernal:
https://github.com/kinvolk/inspektor-gadget/blob/bcf6b76afe652de9080b61817dc17312e2d392bd/cmd/kubectl-gadget/utils/errors.go#L69 has to be renamed, otherwise compilation of kubectl-gadget fails.

PR #548, opened 2022-03-04
Remove tools-mode flag

mauriciovasquezbernal:
The tools-mode flag was intended to let the user choose the kind of
BCC tools to use: python-based/standard vs libbpf-based/core.

We have ported the tools to have their control plane in Golang
and we implemented a logic to automatically fallback to standard mode
when it's not possible to use CO:RE. For this reason this flag is not
useful anymore. In the future we could think about a way to specify the
tool mode (mainly for tests).

Fixes https://github.com/kinvolk/inspektor-gadget/issues/397

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -12,11 +12,28 @@ node, as well as whether or not the kernel has
 [BTF](https://www.kernel.org/doc/html/latest/bpf/btf.html) enabled (via
 `CONFIG_DEBUG_INFO_BTF=y`).
 
-The required BTF information for some well known kernel versions is
-generated with [BTFGen](https://github.com/kinvolk/btfgen) and shipped
-within the gadget container image. If there are not shipped types for
-the running kernel then they are downloaded from
-[BTFhub](https://github.com/aquasecurity/btfhub/).
+Some of the gadgets have two different implementations. The first is
+based on the [traditional BCC
+tools](https://github.com/iovisor/bcc/tree/master/tools), in this case
+they are compiled at runtime on the target machine before loading. These
+tools usually work on older kernels (~4.15) and require to have the
+kernel headers available or the `IKHEADERS` module loaded. The second
+implementation uses the modern Compile Once - Run Everywhere (CO:RE)
+approach. These tools are not compiled at runtime and it's needed to
+have BTF information. This information is collected from three different
+sources, a fallback mechanism is implemented to try another source if the
+previous one was not available.
+1. The kernel already exposes it through `/sys/kernel/btf/vmlinux`: the
+   kernel was compiled with `CONFIG_DEBUG_INFO_BTF`).
+2. It's available in the gadget container image: we ship the BTF
+   information for some well known kernel versions using
+   [BTFGen](https://github.com/kinvolk/btfgen).
+3. It's downloaded from
+   [BTFHub](https://github.com/aquasecurity/btfhub/).
+
+Inspektor Gadget always tries to use the CO:RE implementation of the
+gadget (when available), if it fails then it falls back to the
+traditional BCC-based implementation.
 
 ## Required Kernel Versions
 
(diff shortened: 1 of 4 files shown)


blanquicet (commented):
I think we should describe this CORE ->Standard fallback mechanism in the [kernel-requirements](https://github.com/kinvolk/inspektor-gadget#kernel-requirements) section of the README or in the [requirements](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/requirements.md) doc itself. WDYT?

mauriciovasquezbernal:
Makes sense. I reworked that documentation to make it easier for the user to understand.

docs/requirements.md
```diff
@@ -12,11 +12,27 @@ node, as well as whether or not the kernel has
...
+approach. These tools are not compiled at runtime and it's needed to
+have the BTF information of the kernel where the tools run that can be
+gathered from three different sources:
+1. The kernel exposes it through `/sys/kernel/btf/vmlinux`: the kernel
+   was compiled with `CONFIG_DEBUG_INFO_BTF`).
+2. It's available in the gadget container image: we ship the BTF
+   information for some well known kernel versions using
+   [BTFGen](https://github.com/kinvolk/btfgen).
+3. It's downloaded from
+   [BTFhub](https://github.com/aquasecurity/btfhub/).
```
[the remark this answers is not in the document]
mauriciovasquezbernal: What do you mean by "ship it" exactly?

docs/requirements.md
```diff
@@ -26,28 +42,22 @@ and their shipped kernels, hence it's possible that some gadgets work in older
...
+| network policy advisor |                         |
+| oomkill                | 5.4                     |
+| opensnoop              | 4.15 (BCC), 5.4 (CO:RE) |
+| sigsnoop               | 5.4                     |
+| socket collector       | 5.10                    |
+| process collector      | 5.10                    |
+| tcpconnect             | 4.15 (BCC), 5.8 (CO:RE) |
+| tcptop                 | 4.15                    |
+| tcptracer              | 4.15                    |
+| traceloop              | 4.15                    |
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Right, it's more understandable than before.

docs/requirements.md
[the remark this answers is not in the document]
blanquicet: Oops, _ship_ is not the correct verb. What I meant is that we are downloading the vmlinux from BTFHub if `/sys/kernel/btf/vmlinux` is not available and the container does not contain the BTF file generated by BTFGen.

I am just suggesting to describe a little bit more this fallback mechanism 🙂.

docs/requirements.md
```diff
@@ -12,11 +12,27 @@ node, as well as whether or not the kernel has
...
+1. The kernel already exposes it through `/sys/kernel/btf/vmlinux`: the
+   kernel was compiled with `CONFIG_DEBUG_INFO_BTF`).
+2. It's available in the gadget container image: we ship the BTF
+   information for some well known kernel versions using
+   [BTFGen](https://github.com/kinvolk/btfgen).
+3. It's downloaded from
+   [BTFhub](https://github.com/aquasecurity/btfhub/).
+
+Inspektor Gadget always tries to use the CO:RE implementation of the
+gadget (when available), if it fails then fail then it fallbacks to the
```
alban, line 34:
```suggestion
gadget (when available), if it fails then it fallbacks to the
```

gadget-container/entrypoint.sh
```diff
@@ -215,29 +215,6 @@ else
...
-    echo "vmlinux not found. Using standard tools"
-    TOOLS_MODE="standard"
-  fi
-fi
-
-# Create symlinks for tools according to the value of TOOLS_MODE
-if [ "$TOOLS_MODE" = "core" ] ; then
-  ln -s /bin/libbpf-tools/ /bin/gadgets
-elif [ "$TOOLS_MODE" = "standard" ] ; then
-  ln -s /usr/share/bcc/tools/ /bin/gadgets
```
alban, line 238: Since `/bin/gadgets` is no longer created, the following comment should be updated:

https://github.com/kinvolk/inspektor-gadget/blob/40376f7173cede226eb132f6b9ffd745c74def93/cmd/kubectl-gadget/bcck8s.go#L35-L37

docs/requirements.md
mauriciovasquezbernal: :see_no_evil:

gadget-container/entrypoint.sh
mauriciovasquezbernal: I removed that command as all tools present in bcck8s.go use `/usr/share/bcc/tools`.

docs/requirements.md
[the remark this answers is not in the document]
mauriciovasquezbernal: I changed it a bit to make the fallback mechanism more clear.

docs/requirements.md
```diff
@@ -12,11 +12,28 @@ node, as well as whether or not the kernel has
...
+1. The kernel already exposes it through `/sys/kernel/btf/vmlinux`: the
+   kernel was compiled with `CONFIG_DEBUG_INFO_BTF`).
+2. It's available in the gadget container image: we ship the BTF
+   information for some well known kernel versions using
+   [BTFGen](https://github.com/kinvolk/btfgen).
+3. It's downloaded from
+   [BTFHub](https://github.com/aquasecurity/btfhub/).
+
+Inspektor Gadget always tries to use the CO:RE implementation of the
+gadget (when available), if it fails then then it falls back to the
```
blanquicet, line 35:
```suggestion
gadget (when available), if it fails then it falls back to the
```

blanquicet (approved):
LGTM

PR #500, opened 2022-02-09
Group gadgets by thematic subcommand.

eiffel-fl:
Hi.

In this PR, I grouped the gadgets under semantic commands, some of the gadgets were renamed:

```bash
$ ./kubectl-gadget advisor --help
Advise system configurations based on collected information

Usage:
  kubectl-gadget advisor [flags]
  kubectl-gadget advisor [command]

Available Commands:
  network-policy Generate network policies based on recorded network activity
  seccomp        Generate seccomp policies based on recorded syscalls activity

$ ./kubectl-gadget profiler --help
Profile different subsystems

Usage:
  kubectl-gadget profiler [flags]
  kubectl-gadget profiler [command]

Available Commands:
  biolatency  Generate a histogram with the distribution of block device I/O latency
  stacktrace  Profile CPU usage by sampling stack traces

$ ./kubectl-gadget snapshot --help
Take a snapshot of a subsystem and print it

Usage:
  kubectl-gadget snapshot [flags]
  kubectl-gadget snapshot [command]

Available Commands:
  process     Gather information about running processes
  socket      Gather information about network sockets

$ ./kubectl-gadget snoop --help
Trace and print system events

Usage:
  kubectl-gadget snoop [flags]
  kubectl-gadget snoop [command]

Available Commands:
  bind         Trace IPv4 and IPv6 bind() system calls
  capabilities Trace used capabilities
  dns          Trace DNS requests
  exec         Trace new processes
  mount        Trace mount and umount syscalls
  oomkill      Trace when OOM killer is triggered and kills a process
  open         Trace open() system calls
  tcpconnect   Trace TCP connect() system calls
  tcptracer    Trace tcp connect, accept and close

$ ./kubectl-gadget top --help
Gather, sort and print events according to a given criteria

Usage:
  kubectl-gadget top [flags]
  kubectl-gadget top [command]

Available Commands:
  bio         Trace block device I/O
  file        Trace reads and writes by file, with container details
  tcp         Trace TCP traffic in a specific pod
```

Best regards.

PR #551, opened 2022-03-07
Add golang control plane for biotop.

eiffel-fl:
Hi.

In this PR, I translated `biotop` to golang.
The CO-RE code is not merged at the moment in iovisor/bcc#3893.
This is now possible to get output from all the nodes:

```bash
$ ./kubectl-gadget biotop -A
NODE             NAMESPACE        POD              CONTAINER        PID     COMM             R/W MAJOR  MINOR  BYTES   TIME(µs) IOs
minikube         kube-system      etcd-minikube    etcd             1828    etcd             W   0      0      0       0       0

NODE             NAMESPACE        POD              CONTAINER        PID     COMM             R/W MAJOR  MINOR  BYTES   TIME(µs) IOs
minikube-m02     kube-system      kindnet-lrwql    kindnet-cni      1551    kindnetd         R   0      0      61440   148     2

NODE             NAMESPACE        POD              CONTAINER        PID     COMM             R/W MAJOR  MINOR  BYTES   TIME(µs) IOs
minikube         kube-system      coredns-78fcd69978-gvk8l coredns          3141    coredns          R   0      0      57344   1176    5
minikube-m04     kube-system      kindnet-5qjf8    kindnet-cni      1583    kindnetd         R   0      0      552960  372     7
minikube         kube-system      kube-apiserver-minikube kube-apiserver   3141    coredns          R   0      0      0       0       0
```

Best regards.


pkg/gadgets/biotop/gadget.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/biotop/types"
+
+	gadgetv1alpha1 "github.com/kinvolk/inspektor-gadget/pkg/apis/gadget/v1alpha1"
+)
+
+const (
+	maxRowsDefault  = 20
+	intervalDefault = 1
+	sortByDefault   = types.ALL
+)
```
blanquicet, line 37: I don't think they will change but what do you think about defining these default values on a "neutral" place like `pkg/gadgets/biotop/types/types.go` and then using those values from the CLI and the container side?

That was the strategy we used for [fsslower](https://github.com/kinvolk/inspektor-gadget/blob/ceea77517b9bfe7c3665967faec27046fe2208b7/pkg/gadgets/fsslower/types/types.go#L21-L23).

pkg/gadgets/biotop/gadget.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func (f *TraceFactory) Description() string {
+	t := `biotop shows command generating block I/O, with container details.
+
+The following parameters are supported:
+ - interval: Output interval, in seconds. (default %d)
+ - max_rows: Maximum rows to print. (default %d)
+ - sort: The field to sort the results by (%s). (default %s)`
```
blanquicet, line 62: Isn't it `sort_by`? I noticed the same problem on the filetop gadget, could you please fix both.

pkg/gadgets/biotop/gadget.go
```diff
@@ -0,0 +1,202 @@
...
+	sortBy := sortByDefault
+
+	if trace.Spec.Parameters != nil {
+		params := trace.Spec.Parameters
+		var err error
+
+		if val, ok := params["max_rows"]; ok {
+			maxRows, err = strconv.Atoi(val)
+			if err != nil {
+				trace.Status.OperationError = fmt.Sprintf("%q is not valid for ouput_rows", val)
```
blanquicet, line 122: Isn't it `max_rows`? I noticed the same problem on the filetop gadget, could you please fix both.

pkg/gadgets/biotop/gadget.go
```diff
@@ -0,0 +1,202 @@
...
+				return
+			}
+		}
+
+		if val, ok := params["interval"]; ok {
```
blanquicet, line 127: In the CLI it is named `printInterval`, isn't it?

pkg/gadgets/biotop/gadget.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+		}
+
+		if val, ok := params["sortby"]; ok {
+			sortBy, err = types.ParseSortBy(val)
+			if err != nil {
+				trace.Status.OperationError = fmt.Sprintf("%q is not valid for sortby", val)
+				return
+			}
+		}
```
blanquicet, line 141: Same here, in the CLI it is defined as `sort_by`.

I expect these two parameters to not work with the CLI or what did I miss?

cmd/kubectl-gadget/biotop.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+
+		return nil
+	},
+	Args: cobra.MaximumNArgs(1),
+}
+
+func init() {
+	biotopCmd.Flags().IntVarP(&maximumRows, "maximumRows", "r", 20, "Maximum rows to print")
+	biotopCmd.Flags().StringVarP(&sortTypeStr, "sort", "", "all", "Sort column")
```
blanquicet, line 100: What about showing the supported values with `strings.Join(types.SortBySlice, ",")`?

blanquicet (commented):
Thanks for working on this!

I did an initial review and I added some comments about the parameters but it already looks great. I will test it once these comments are managed.

cmd/kubectl-gadget/biotop.go
```diff
@@ -0,0 +1,274 @@
...
+		config := &utils.TraceConfig{
+			GadgetName:       "biotop",
+			Operation:        "start",
+			TraceOutputMode:  "Stream",
+			TraceOutputState: "Started",
+			CommonFlags:      &params,
+			Parameters: map[string]string{
+				"max_rows":      strconv.Itoa(maximumRows),
+				"printInterval": strconv.Itoa(printInterval),
+				"sort_by":       sortTypeStr,
```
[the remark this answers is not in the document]
eiffel-fl: Oops, too big `s///`!

pkg/gadgets/biotop/gadget.go
eiffel-fl: It is planned, I want to define them in `cmd/kubectl-gadget/top/top.go` of #500 once merged.

blanquicet: Mmmm but doing so you need to redefine then again in the "container" side and that's what I am suggesting to avoid, or what did I miss?

eiffel-fl: Oh OK! I did not understand your suggestion!

mauriciovasquezbernal (commented):
Overall looks very good! I'm facing some issues during the tests. There are cases when the gadget seems not to capture the event, for instance, if I run `dd if=/dev/zero of=/tmp/foo count=16384`, that event is not always printed by the gadget but it's printed by the BCC tool (either by the python and the libbpf based ones). 

---

After some investigation I found out the cause. I think the mount namespace ID can't be stored in blk_account_io_done`, most of the time is returns the mount namespace id of the host. (Probably because that operation is almost executed in the host context by the kernel?). If you gather the mountnsid in the same place (blk_account_io_start) where the pid is gathered it should work much better.

cmd/kubectl-gadget/bcck8s.go
```diff
@@ -178,16 +152,6 @@ func bccCmd(subCommand, bccScript string) func(*cobra.Command, []string) error {
...
-		case "capabilities":
-			if stackFlag {
-				gadgetParams += " -K"
-			}
-			if uniqueFlag {
-				gadgetParams += " --unique"
-			}
-			if params.Verbose {
-				gadgetParams += " -v"
-			}
```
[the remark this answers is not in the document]
eiffel-fl: Because it is dead code since `capabilities` is not using this path anymore.

pkg/gadgets/biotop/tracer/tracer.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return fmt.Errorf("error opening kprobe: %w", err)
+	}
+
+	t.startRequestLink, err = link.Kprobe("blk_mq_start_request", t.objs.BlkMqStartRequest)
+	if err != nil {
+		return fmt.Errorf("error opening kprobe: %w", err)
+	}
+
+	t.doneLink, err = link.Kprobe("blk_account_io_done", t.objs.BlkAccountIoDone)
```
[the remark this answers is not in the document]
eiffel-fl: I forgot to add this logic here.

pkg/gadgets/biotop/gadget.go
```diff
@@ -0,0 +1,196 @@
+// Copyright 2019-2021 The Inspektor Gadget authors
```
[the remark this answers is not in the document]
eiffel-fl: Updated but we should really think about `SPDX`.

blanquicet (commented):
> There are cases when the gadget seems not to capture the event
>
I tested it in an AKS cluster and I noticed this for the first events but after checking the gadget pod logs, as we discussed in https://github.com/kinvolk/inspektor-gadget/pull/558, sometimes it seems to require some seconds to really start the gadget:

```bash
time="2022-03-16T14:21:46Z" level=info msg="Reconcile trace gadget/biotop-9zrqm (gadget biotop, node aks-agentpool-31885719-vmss000000)"
time="2022-03-16T14:21:46Z" level=info msg="Gadget biotop operation \"start\" on gadget/biotop-9zrqm"
time="2022-03-16T14:21:51Z" level=info msg="Gadget completed operation. Trace status will be updated accordingly"
time="2022-03-16T14:21:51Z" level=info msg="Updating new status of trace \"gadget/biotop-9zrqm\": state=Started operationError=\"\" operationWarning=\"\" output=<0 characters>"
```

While in OpenShift (ARO) it does not seem to work at all 😢 .

cmd/kubectl-gadget/biotop.go
```diff
@@ -0,0 +1,274 @@
...
+
+			fmt.Printf("%-16s %-16s %-16s %-16s %-7d %-16s %-3c %-6d %-6d %-7d %-8d %d\n",
+				event.Node, event.Namespace, event.Pod, event.Container,
+				event.Pid, event.Comm, rw, event.Major, event.Minor, event.Bytes,
+				event.MicroSecs, event.Ios)
+		}
+	case utils.OutputModeJSON:
+		b, err := json.Marshal(stats)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error marshalling json: %v", err)
```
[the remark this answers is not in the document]
eiffel-fl: >I would be great to have a CI job that verifies this automatically.

I am thinking to it (as well as some scripts to check README.md output of `kubectl gadget --help` is in sync).
We are only using `utils.Error` on the CLI, right?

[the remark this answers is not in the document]
blanquicet: Yes, only from CLI.

cmd/kubectl-gadget/bcck8s.go
```diff
@@ -31,11 +31,9 @@ import (
...
 )
 
-var biotopCmd = &cobra.Command{
-	Use:   "biotop",
-	Short: "Trace block device I/O",
-	RunE:  bccCmd("biotop", "/usr/share/bcc/tools/biotop"),
-}
+// create the commands for the different gadgets. The gadgets that have CO-RE
+// support should use "/bin/gadgets/" as the path for the binary. Otherwise
+// "/usr/share/bcc/tools/" should be used.
```
mauriciovasquezbernal, line 36: Why to add this comment again? I think it's not needed anymore.

mauriciovasquezbernal (approved):
Some small comments. LGTM overall. Great contribution!

cmd/kubectl-gadget/biotop.go
```diff
@@ -0,0 +1,274 @@
...
+
+var (
+	// arguments
+	printInterval int = types.IntervalDefault
+
+	// flags
+	biotopMaximumRows int
+	biotopSortTypeStr string
+	biotopSortType    types.SortBy
+)
```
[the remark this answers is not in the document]
eiffel-fl: While it compiles like this, I will keep it.
I will work on this once #500 is merged.
For example, `mutex` and `finished` can be grouped for all `*top` gadgets.

eiffel-fl:
Thank you for the reviews!

PR #552, opened 2022-03-08
Update go fanotify

mauriciovasquezbernal:
Fixes https://github.com/kinvolk/inspektor-gadget/issues/294


mauriciovasquezbernal:
@alban would you mind reviewing this one please?

alban (approved):
LGTM

I tested this patch with local-gadget and starting/stopping containers:
* `/proc/$pid/fd` does not show any fd leaks
* `strace -p $pid -f -e close` does not show any double-close or invalid closes

PR #553, opened 2022-03-08
pkg/gadgets: Uniform perf buffer memory size for all gadgets

mauriciovasquezbernal:
Many gadgets were using different values for the memory size of the
perf ring buffer. This commit defines a const and uses the same value
for all of them.

### DIFF
--- a/pkg/gadgets/bindsnoop/tracer/core/tracer.go
+++ b/pkg/gadgets/bindsnoop/tracer/core/tracer.go
@@ -177,7 +177,7 @@ func (t *Tracer) start() error {
 		return fmt.Errorf("error opening ipv6 kprobe: %w", err)
 	}
 
-	t.reader, err = perf.NewReader(t.objs.bindsnoopMaps.Events, os.Getpagesize())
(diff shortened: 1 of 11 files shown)


pkg/gadgets/bindsnoop/tracer/core/tracer.go
```diff
@@ -177,7 +177,8 @@ func (t *Tracer) start() error {
 		return fmt.Errorf("error opening ipv6 kprobe: %w", err)
 	}
 
-	t.reader, err = perf.NewReader(t.objs.bindsnoopMaps.Events, os.Getpagesize())
+	t.reader, err = perf.NewReader(t.objs.bindsnoopMaps.Events,
+		gadgets.PerfBufferPages*os.Getpagesize())
```
blanquicet, line 181: Why did you break just this line 😝 ?

blanquicet (approved):
LGTM!

pkg/gadgets/bindsnoop/tracer/core/tracer.go
mauriciovasquezbernal: That's the first one I edited, then I chose not to break them.

PR #554, opened 2022-03-09
Avoid leaving processes running after integration tests

blanquicet:
# Avoid leaving processes running after integration tests

When running integration tests, the kubectl-gadget CLI process is not correctly terminated. In addition, if the option `-no-deploy` is used, the traces are not deleted.

This fix is particularly useful when running tests locally.

## How to use

It is enough to run the integration test locally and check the processes.

### Problem
When we run the `kubect-gadget` CLI to run a gadget during the integration tests, we do it by executing `/bin/sh -c kubectl-gadget <gadget-name> ...`, see [here](https://github.com/kinvolk/inspektor-gadget/blob/c8fe352aeb33b2366dec4df8a000ecd1d478e80a/integration/command.go#L160). It means that `/bin/sh` executes a child process for `kubectl-gadget`. The issue occurs because the current code is [killing the process of `/bin/sh`](https://github.com/kinvolk/inspektor-gadget/blob/c8fe352aeb33b2366dec4df8a000ecd1d478e80a/integration/command.go#L178) and not its child. As a result, the process of `kubect-gadget` says alive and if `-no-deploy` is used, the traces are not deleted.

### Solution
To be able to kill the process of `/bin/sh` and its child (the process of `kubectl-gadget`), we need to send the termination signal to their process group ID (PGID). However, child processes get the same PGID as their parents by default, so in order to avoid killing also the integration tests process, we set the fields `Setpgid=true` and `Pgid=0` of `syscall.SysProcAttr` before executing `/bin/sh`. Doing so, the PGID of `/bin/sh` (and its children) will be set to its process ID, see: https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/syscall/exec_linux.go;l=32-34.

Then, we just need to send the PID of `/bin/sh` (which is the same PGID) as a negative number to `syscall.Kill()`. As a result, the signal will be received by all the processes with such PGID, in our case, the process of `/bin/sh` and the one of `kubectl-gadget`.

I got the inspiration from https://medium.com/@felixge/killing-a-child-process-and-all-of-its-children-in-go-54079af94773.

### Run test

```bash
$ pwd
/path/to/source/of/inpektor-gadget
$ cd integration
$ export KUBECTL_GADGET=../kubectl-gadget
$ sudo -E go test -run TestBindsnoop -tags withebpf -integration -no-deploy -test.v ./...
```

### During execution

**Before this PR,** all the children share the same PGID:
```bash
$ ps -ejf
UID          PID    PPID    PGID     SID  C STIME TTY          TIME CMD
...
root       26420    9975   26420    9975  0 11:13 pts/3    00:00:00 sudo -E go test -run TestBindsnoop -tags withebpf -integration -no-deploy -test.v ./...
root       26421   26420   26420    9975  7 11:13 pts/3    00:00:00 go test -run TestBindsnoop -tags withebpf -integration -no-deploy -test.v ./...
root       26599   26421   26420    9975  0 11:13 pts/3    00:00:00 /tmp/go-build3062841243/b001/integration.test -test.paniconexit0 -test.timeout=10m0s -test.run=TestBindsnoop -integration -no-deploy -test.v=true ./...
root       26619   26599   26420    9975  0 11:13 pts/3    00:00:00 /bin/sh -c $KUBECTL_GADGET bindsnoop -n test-bindsnoop-1258208857824550840
root       26621   26619   26420    9975  1 11:13 pts/3    00:00:00 ../kubectl-gadget bindsnoop -n test-bindsnoop-1258208857824550840
...
```

**After this PR,** the process of `/bin/sh` and `kubectl-gadget` share a different PGID (the same PID of `/bin/sh`). Therefore, we can use that value to send a signal to both:

```bash
$ ps -ejf
UID          PID    PPID    PGID     SID  C STIME TTY          TIME CMD
...
root       12304    9975   12304    9975  0 11:06 pts/3    00:00:00 sudo -E go test -run TestBindsnoop -tags withebpf -integration -no-deploy -test.v ./...
root       12305   12304   12304    9975 14 11:06 pts/3    00:00:00 go test -run TestBindsnoop -tags withebpf -integration -no-deploy -test.v ./...
root       12494   12305   12304    9975  0 11:06 pts/3    00:00:00 /tmp/go-build1126395440/b001/integration.test -test.paniconexit0 -test.timeout=10m0s -test.run=TestBindsnoop -integration -no-deploy -test.v=true ./...
root       12516   12494   12516    9975  0 11:06 pts/3    00:00:00 /bin/sh -c $KUBECTL_GADGET bindsnoop -n test-bindsnoop-1332130641320470889
root       12518   12516   12516    9975  4 11:06 pts/3    00:00:00 ../kubectl-gadget bindsnoop -n test-bindsnoop-1332130641320470889
...
```

### After execution
**Before this PR,** the process of `kubectl-gadget` is still running thus trace was not deleted:
```bash
$ ps -efj | grep bindsnoop
root       26621    7034   26420    9975  0 11:13 pts/3    00:00:00 ../kubectl-gadget bindsnoop -n test-bindsnoop-1258208857824550840
jose       27021   11798   27020   11798  0 11:17 pts/2    00:00:00 grep --color=auto bindsnoop

$ kubectl get traces -n gadget
NAME              AGE
bindsnoop-xwmwb   83s
```
**After this PR,** all processes terminated and trace was deleted:
```bash
$ ps -ejf | grep -i bindsnoop
jose       14440   11798   14439   11798  0 11:08 pts/2    00:00:00 grep --color=auto -i bindsnoop

$ kubectl get traces -n gadget
No resources found in gadget namespace.

```

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"syscall"
 	"testing"
 
 	"github.com/kr/pretty"
@@ -158,6 +159,17 @@ func (c *command) runWithoutTest() error {
 func (c *command) start(t *testing.T) {
 	t.Logf("Start command: %s\n", c.cmd)
 	cmd := exec.Command("/bin/sh", "-c", c.cmd)
+
+	// To be able to kill the process of /bin/sh and its child (the process of
+	// c.cmd), we need to send the termination signal to their process group ID
+	// (PGID). However, child processes get the same PGID as their parents by
+	// default, so in order to avoid killing also the integration tests process,
+	// we set the fields Setpgid and Pgid of syscall.SysProcAttr before
+	// executing /bin/sh. Doing so, the PGID of /bin/sh (and its children)
+	// will be set to its process ID, see:
+	// https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/syscall/exec_linux.go;l=32-34.
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: 0}
+
 	cmd.Stdout = &c.stdout
 	cmd.Stderr = &c.stderr
 
@@ -175,7 +187,12 @@ func (c *command) start(t *testing.T) {
 // Cmd output is then checked with regard to expectedString and expectedRegexp
 func (c *command) stop(t *testing.T) {
 	t.Logf("Stop command: %s\n", c.cmd)
-	err := c.command.Process.Kill()
+
+	// Here, we just need to send the PID of /bin/sh (which is the same PGID) as
+	// a negative number to syscall.Kill(). As a result, the signal will be
+	// received by all the processes with such PGID, in our case, the process of
+	// /bin/sh and c.cmd.
+	err := syscall.Kill(-c.command.Process.Pid, syscall.SIGTERM)
 	if err != nil {
 		t.Fatal(err)
 	}


integration/command.go
```diff
@@ -175,23 +183,26 @@ func (c *command) start(t *testing.T) {
...
 	if c.expectedRegexp != "" {
 		r := regexp.MustCompile(c.expectedRegexp)
 		if !r.MatchString(stdout) {
-			fmt.Fprintf(os.Stderr, "%s\n", stderr)
 			t.Fatalf("regexp didn't match: %s\n%s\n", c.expectedRegexp, stdout)
 		}
-	}
-	if c.expectedString != "" && stdout != c.expectedString {
-		fmt.Fprintf(os.Stderr, "%s\n", stderr)
+	} else if c.expectedString != "" && stdout != c.expectedString {
```
mauriciovasquezbernal, line 205: I'm not sure if we really want to change this behavior. I think it's better to have two independent conditionals in the (extreme) case somebody sets both, `expectedRegexp` and `expectedString`. Otherwise one of them will be ignored silently.

integration/command.go
```diff
@@ -109,19 +109,19 @@ func (c *command) run(t *testing.T) {
 	t.Logf("command: %s\n", c.cmd)
 	cmd := exec.Command("/bin/sh", "-c", c.cmd)
 	output, err := cmd.CombinedOutput()
-	actual := string(output)
-	t.Logf("command returned:\n%s\n", actual)
```
mauriciovasquezbernal, line 113: I think it's good to print the command before making the test fail. Otherwise we won't know what's the output of the command if it fails (returns with exit code different than 0).

mauriciovasquezbernal (commented):
I agree with the solution to the problem this PR refers to. However, I have some concerns with other changes not related to that problem.

integration/command.go
blanquicet: Yeah, for me it is also "extreme", I do not expect them to be used together but it's ok. I will keep them independent.

integration/command.go
blanquicet: This log is not the command that is being executed but the output of the command which will be for sure empty if `cmd.CombinedOutput()` fails. 

Anyway, I will make a separate PR with a couple of changes like this because this logic of running the command and verifying the output is repeated in three different places.

mauriciovasquezbernal (approved):

PR #561, opened 2022-03-11
ci: Run integration tests on ARO

blanquicet:
# Run integration tests on ARO

This PR modifies the CI to also run the integration tests on a pre-created [Azure Red Hat OpenShift (ARO) cluster](https://docs.microsoft.com/en-us/azure/openshift/intro-openshift).

## Implementation details

- It is a pre-created ARO cluster because its creation takes about 35 minutes: [documentation](https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster#create-the-cluster).
- Integration tests on ARO was added to the `test-integration` job with the minikube (already existing) using a [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs). Notice it was configured with `fail-fast: false` to continue running on other Kubernetes distributions so that one can understand if it is a distro-specific issue or not.
- The [concurrency feature](https://docs.github.com/en/actions/using-jobs/using-concurrency) was used to ensure only one test-integration job runs at a time to avoid trying to deploy IG multiple times on the same Kubernetes cluster. It is specifically needed for ARO because it is a pre-created cluster, see _Improvements_ section.
- Authentication credentials are not associated with any user but the ones generated with [az](https://docs.microsoft.com/en-us/azure/openshift/tutorial-connect-cluster), see _Improvements_ section though. Notice that this information was stored in the repo secrets. See [REAMDE](https://github.com/kinvolk/inspektor-gadget/blob/jose/ci-aro/.github/workflows/README.md) file this PR adds.
- The kernel headers for the RedHat CoreOS are downloaded as one of the first steps when the gadget pod is started. It means that even if the pod is in `Ready` state, it is not yet able to run gadgets because it is downloading the kernel headers. As a workaround, I increased the initial delay after the pod is ready and before starting running the gadgets, see _Improvements_ section below.
- Not all the gadgets are working on ARO so they are being temporary skipped, see the _Improvements_ section though. So far, we are only sure that **SocketCollector** and **ProcessCollector** will never work on ARO because they require a kernel version newer than 5.10 and ARO uses 4.18.

## Improvements (In future PRs)

- [ ] Use a functional Service Account instead of admin credentials: https://github.com/kinvolk/inspektor-gadget/issues/574
- [x] By using the [concurrency feature](https://docs.github.com/en/actions/using-jobs/using-concurrency), when there are two workflows running in parallel, the entire `test-integration` job on one of them remains blocked waiting for the other one to finish. It means that the overall CI process is now taking longer while the test-integration job could run in parallel in different workflows when they are using the minikube-github: https://github.com/kinvolk/inspektor-gadget/issues/575  
- [ ] Improve the deploy command to return just when the pod is ready to run gadgets: https://github.com/kinvolk/inspektor-gadget/issues/123
- [ ] Make all gadgets work:
  - **Biolatency**: See issue https://github.com/kinvolk/inspektor-gadget/issues/572
  - **Dns** and **Snisnoop**: See issue https://github.com/kinvolk/inspektor-gadget/issues/569
  - **Capabilities**, **Fsslower**, **Tcptracer**: See issues https://github.com/kinvolk/inspektor-gadget/issues/570, https://github.com/kinvolk/inspektor-gadget/issues/571, https://github.com/kinvolk/inspektor-gadget/issues/568.
  - **Biotop**: See issue https://github.com/kinvolk/inspektor-gadget/issues/589.
  - **Audit-seccomp**: See issue https://github.com/kinvolk/inspektor-gadget/issues/631.

## TODO (This PR)

- [x] Create issues for improvements and not working gadgets. Then, specify then in the code, see [comments]( https://github.com/kinvolk/inspektor-gadget/pull/561#pullrequestreview-910500319).

## How to use it

This PR adds also the [docs/ci.md](https://github.com/kinvolk/inspektor-gadget/blob/jose/ci-aro/docs/ci.md) describing how to create an ARO cluster so that the `Inspektor Gadget CI` workflow will run the integration tests on it.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -321,67 +321,92 @@ jobs:
 
         make local-gadget-tests
 
+  check-secrets:
+    name: Check repo secrets
+    # level: 0
+    runs-on: ubuntu-latest
+    outputs:
+      aro: ${{ steps.set_output.outputs.aro }}
+    steps:
+      # Secrets cannot be used as if condition, use job output as workaround.
+      # https://github.com/actions/runner/issues/520
+      - id: set_output
+        run: |
+          if [[ "${{ secrets.OPENSHIFT_SERVER }}" != "" && \
+                "${{ secrets.OPENSHIFT_USER }}" != "" && \
+                "${{ secrets.OPENSHIFT_PASSWORD }}" != "" ]]; \
+          then
+            echo "Secrets to use an ARO cluster were configured in the repo"
+            echo "::set-output name=aro::true"
+          else
+            echo "Secrets to use an ARO cluster were not configured in the repo"
+            echo "::set-output name=aro::false"
+          fi
+
+  # Integration tests for ARO are separated from others distributions because it
+  # is a pre-created cluster. It implies that we need to use a concurrency group
+  # to ensure that only one test-integration-aro job runs at a time so that we
+  # never try to use IG on that unique ARO cluster from different workflow runs.
+  test-integration-aro:
+    name: Integration tests on ARO
+    # level: 1
+    needs: [check-secrets, test-unit, build-kubectl-gadget, build-local-gadget, build-docker-image]
+    # Run this job only if an ARO cluster is available on repo secrets. See
+    # docs/ci.md for further details.
+    if: needs.check-secrets.outputs.aro == 'true'
+    runs-on: ubuntu-latest
+    concurrency:
+      group: no-simultaneous-test-integration-aro
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v1
+    - name: Authenticate and set ARO cluster context
+      # NOTE: This action generates the Kubernetes config file in the current
+      # directory. Therefore, it must be run after checking out code otherwise
+      # the file will be cleaned up.
+      uses: redhat-actions/oc-login@v1
+      with:
+        # API Server URL
+        openshift_server_url: ${{ secrets.OPENSHIFT_SERVER }}
+        # Credentials (TODO: Use a functional Service Account, see issue #574)
+        openshift_username: ${{ secrets.OPENSHIFT_USER }}
+        openshift_password: ${{ secrets.OPENSHIFT_PASSWORD }}
+    - name: Run integration tests
+      uses: ./.github/actions/run-integration-tests
+      env:
+        KUBERNETES_DISTRIBUTION: "aro"
+      with:
+        registry: ${{ secrets.CONTAINER_REGISTRY }}
+        username: ${{ secrets.CONTAINER_REGISTRY_USERNAME }}
+        password: ${{ secrets.CONTAINER_REGISTRY_PASSWORD }}
+
   test-integration:
     name: Integration tests
     # level: 1
     needs: [test-unit, build-kubectl-gadget, build-local-gadget, build-docker-image]
     runs-on: ubuntu-latest
     steps:
+    - name: Check out code
+      uses: actions/checkout@v1
     - name: Setup Minikube
       uses: manusa/actions-setup-minikube@v2.4.3
       with:
         minikube version: 'v1.25.2'
         kubernetes version: 'v1.23.0'
         github token: ${{ secrets.GITHUB_TOKEN }}
-    - name: Set IMAGE_TAG
-      run: |
-        TMP1=${GITHUB_REF#*/}
-        TMP2=${TMP1#*/}
-        IMAGE_TAG=${TMP2//\//-}
-        if [ "$IMAGE_TAG" = "main" ]; then
-            IMAGE_TAG="latest"
-        fi
-        echo IMAGE_TAG=$IMAGE_TAG >> $GITHUB_ENV
-    - name: Check out code
-      uses: actions/checkout@v1
-    - name: Get kubectl-gadget-linux-amd64.tar.gz from artifact.
-      uses: actions/download-artifact@v2
-      with:
-        name: kubectl-gadget-linux-amd64-tar-gz
-        path: /home/runner/work/inspektor-gadget/
-    - name: Integration tests
-      run: |
-        echo "Using IMAGE_TAG=$IMAGE_TAG"
-
-        tar zxvf /home/runner/work/inspektor-gadget/kubectl-gadget-linux-amd64.tar.gz
-        mv kubectl-gadget kubectl-gadget-linux-amd64
-
-        TESTS_DOCKER_ARGS="-e KUBECONFIG=/root/.kube/config -v /home/runner/.kube:/root/.kube -v /home/runner/work/_temp/.minikube:/home/runner/work/_temp/.minikube" \
-            make -C integration build test
-
-        sed -i "s/latest/$IMAGE_TAG/g" integration/gadget-integration-tests-job.yaml
-    - name: Add integration asset as artifact.
-      if: startsWith(github.ref, 'refs/tags/v')
-      uses: actions/upload-artifact@master
-      with:
-        name: integration-asset
-        path: /home/runner/work/inspektor-gadget/inspektor-gadget/integration/gadget-integration-tests-job.yaml
-    - name: Login to Container Registry
-      if: startsWith(github.ref, 'refs/tags/v')
-      uses: docker/login-action@v1
+    - name: Run integration tests
+      uses: ./.github/actions/run-integration-tests
+      env:
+        KUBERNETES_DISTRIBUTION: "minikube-github"
       with:
         registry: ${{ secrets.CONTAINER_REGISTRY }}
         username: ${{ secrets.CONTAINER_REGISTRY_USERNAME }}
         password: ${{ secrets.CONTAINER_REGISTRY_PASSWORD }}
-    - name: Push Integration Test Image
-      if: startsWith(github.ref, 'refs/tags/v')
-      run: |
-        make -C integration push
 
   release:
     name: Release
     # level: 2
-    needs: [documentation-checks, lint, test-integration, test-local-gadget]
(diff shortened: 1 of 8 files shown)


blanquicet:
I think it is ready for a first review!

.github/workflows/inspektor-gadget.yml
```diff
@@ -364,13 +364,38 @@ jobs:
...
+    - if: matrix.k8s-distro == 'aro'
+      name: Authenticate and set ARO cluster context
+      uses: redhat-actions/oc-login@v1
+      with:
+        # API Server URL
+        openshift_server_url: ${{ secrets.OPENSHIFT_SERVER }}
+
+        # Credentials (TODO: Use a functional Service Account)
+        openshift_username: ${{ secrets.OPENSHIFT_USER }}
+        openshift_password: ${{ secrets.OPENSHIFT_PASSWORD }}
```
mauriciovasquezbernal, line 398: Just wondering, is it the piece that configures the kubeconfig so kubect-gadget commands are associated to that cluster?

blanquicet: Yeah, it is.

blanquicet:
> What about opening issues for the gadgets that are not working and adding a link to them in the printed message? (Basically replacing the "To be investigated part").
>
Absolutely! I still need to do it. I will do the same for [this TODO](https://github.com/kinvolk/inspektor-gadget/blob/bd671c5781eacdae08c41991b77935e2251a6da3/.github/workflows/inspektor-gadget.yml#L396).

> I think it's interesting to have a long running cluster where we install and remove IG multiple times, let's see if this keeps working for a long time.
>
Yeah, I expect some issues to come out after a while. Let's see 💪🏽!

.github/workflows/README.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+If we need to delete our cluster, it is enough to execute:
+```bash
+$ az aro delete --resource-group $RESOURCEGROUP --name $CLUSTER
+```
+
+Take into account that it will remove the entire resource group and all
+resources inside it.
+
```
blanquicet, line 61: Took this from [documentation](https://docs.microsoft.com/en-us/azure/openshift/tutorial-delete-cluster#delete-the-cluster) (and it is the same [upstream](https://github.com/MicrosoftDocs/azure-docs/blob/main/articles/openshift/tutorial-delete-cluster.md#delete-the-cluster)) but it seems strange to me that removing the `aro` resource removes the entire resource group 🤔. I will make a test tomorrow to confirm it.

mauriciovasquezbernal (approved):
This is a great improvement for integration testing!

.github/workflows/README.md
blanquicet: My tests confirmed the documentation is not accurate. I updated this PR to use the instructions I used in my tests and I created a PR to improve the documentation: https://github.com/MicrosoftDocs/azure-docs/pull/90207

blanquicet:
I rebased the PR after the integration of the audit-seccomp gadget.

integration/integration_test.go
```diff
@@ -457,6 +509,10 @@ func TestOpensnoop(t *testing.T) {
 }
 
 func TestProcessCollector(t *testing.T) {
+	if *k8sDistro == K8sDistroARO {
+		t.Skip("Skip running process-collector gadget on ARO: iterators are not supported on kernel 4.18.0-305.19.1.el8_4.x86_64")
```
alban, line 513: I wonder if this should be detected with the kernel version, so that the test would be automatically enabled when the ARO kernel supports it.

```
kubectl get node -o jsonpath='{.items[0].status.nodeInfo.kernelVersion}'
```

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
+# Inspektor Gadget CI Workflow
```
alban, line 1: Could this documentation be moved somewhere in `docs/`?

In this way, documentation websites (such as https://kinvolk.io/docs/inspektor-gadget/latest) could be automatically generated.

We could keep a small `.github/workflows/README.md` redirecting to the file in `docs/`.

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+#### From GitHub actions
+
+The `test-integration` job is already configured to authenticate and set the
+kubeconf context to the ARO cluster configured in the GitHub repository. So all
+we need to do is to add the following actions secrets:
+
+- `OPENSHIFT_SERVER`: The API server URL: `https://api.server.example.io:1234`.
```
alban, line 91: We have some documentation explaining the environment variables for the GitHub Action:

https://github.com/kinvolk/inspektor-gadget/blob/main/docs/CONTRIBUTING.md#github-actions

But those 2 docs don't refer to each other. Could it be unified somewhere?

.github/workflows/inspektor-gadget.yml
```diff
@@ -364,13 +364,39 @@ jobs:
     # level: 1
     needs: [test-unit, build-kubectl-gadget, build-local-gadget, build-docker-image]
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s-distro: [minikube-github, aro]
```
alban, line 370: Could `aro` be disabled if `${{ secrets.OPENSHIFT_SERVER }}` is not set? I tried it on a fork of IG, and I get the following:
image

Currently, the contributing guide does not list this environment variable in https://github.com/kinvolk/inspektor-gadget/blob/main/docs/CONTRIBUTING.md#github-actions

blanquicet: Absolutely! I will work on this.

.github/workflows/inspektor-gadget.yml
```diff
@@ -359,67 +359,92 @@ jobs:
...
-    - name: Push Integration Test Image
-      if: startsWith(github.ref, 'refs/tags/v')
-      run: |
-        make -C integration push
 
   release:
     name: Release
     # level: 2
-    needs: [documentation-checks, lint, test-integration, test-local-gadget]
+    needs: [documentation-checks, lint, test-integration, test-integration-aro, test-local-gadget]
```
blanquicet, line 447: Do you agree to make `test-integration-aro` a need of the `release` job? Take into account that `test-integration-aro` is now optional (based on secrets' existence) thus if we do not configure an ARO cluster on the repo secrets both `test-integration-aro` and `release` will be skipped.

.github/workflows/inspektor-gadget.yml
blanquicet: Done. Notice that secrets cannot be used as `if` condition so I used job output as a workaround. WDYT?

.github/workflows/inspektor-gadget.yml
```diff
@@ -359,67 +359,92 @@ jobs:
...
+            echo "Secrets to use an ARO cluster were not configured in the repo"
+            echo "::set-output name=aro::false"
+          fi
+
+  # Integration tests for ARO are separated from others distributions because it
+  # is a pre-created cluster. It implies that we need to use a concurrency group
+  # to ensure that only one test-integration-aro job runs at a time so that we
+  # never try to use IG on that unique ARO cluster from different workflow runs.
+  test-integration-aro:
+    name: Integration tests on ARO
```
blanquicet, line 389: This should fix https://github.com/kinvolk/inspektor-gadget/issues/575.

blanquicet:
PR rebased

blanquicet:
PR updated to skip integration test for biotop gadget because it is not woking anymore on ARO after https://github.com/kinvolk/inspektor-gadget/pull/551. I created a dedicated issue to investigate it: https://github.com/kinvolk/inspektor-gadget/issues/589.

integration/integration_test.go
```diff
@@ -48,7 +48,7 @@ func runCommands(cmds []*command, t *testing.T) {
 		for _, cmd := range cmds {
 			if cmd.startAndStop && cmd.started {
 				// Wait a bit before stopping the command.
-				time.Sleep(10 * time.Second)
+				time.Sleep(15 * time.Second)
```
eiffel-fl, line 61: Please indicate in your commit message why do you need this.

integration/integration_test.go
```diff
@@ -733,7 +734,7 @@ func TestTcptop(t *testing.T) {
 
 	tcptopCmd := &command{
 		name:           "Start tcptop gadget",
-		cmd:            fmt.Sprintf("$KUBECTL_GADGET tcptop --node $(kubectl get node --no-headers | cut -d' ' -f1) -n %s -p test-pod", ns),
```
eiffel-fl, line 736: Good catch but I think I am also dealing with it in #500.

.github/workflows/inspektor-gadget.yml
eiffel-fl: I am not sure to understand, I would imagine if a job is skipped then it fufills needs condition.

eiffel-fl (commented):
It is cool!
I have some minor comments and I mainly reviewed this one to have better understanding of #566.

integration/integration_test.go
eiffel-fl: Indeed, I should have lost it during the rebase process or maybe I only thought to deal with it.
So, thank you for dealing with it here.

.github/workflows/inspektor-gadget.yml
blanquicet: No, if the `test-integration-aro` job is skipped, all the jobs that have it as a `needs` condition will be skipped too. However, in this repo, we will always have the ARO secrets configured so it should never be skipped.

blanquicet:
Rebased and updated to skip audit-seccomp test given that it is always failing. I created a dedicated issue to investigate why it is failing, see https://github.com/kinvolk/inspektor-gadget/issues/631.

PR #491, opened 2022-02-04
Increase traceloop opened file limit.

eiffel-fl:
Hi.

In this PR, I increased the gadget container opened file limit from 1024 to 2048, using `ulimit -n`, right before running `traceloop`.
This is needed in CBL Mariner to avoid this type of error:

```
I0204 11:03:10.222056  197541 podinformer.go:282] Starting Pod controller
error while loading "tracepoint/raw_syscalls/sys_enter" (too many open files):
processed 2529 insns (limit 1000000) max_states_per_insn 4 total_states 209 peak_states 209 mark_read 7
```

Since now all gadgets work on CBL Mariner, merging this PR will fix #401. 

Best regards.

PR #565, opened 2022-03-14
Correct traceloop opened file limit increase.

eiffel-fl:
Hi.

This PR fixes #491, because we cannot use `unlimited`.
So, we change shell to `bash` to be able to use `hard`.

Best regards.

### DIFF
--- a/pkg/gadgets/traceloop/gadget.go
+++ b/pkg/gadgets/traceloop/gadget.go
@@ -110,7 +110,7 @@ func (f *TraceFactory) Operations() map[string]gadgets.TraceOperation {
 }
 
 func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
-	t.cmd = exec.Command("/bin/sh", "-c", `
+	t.cmd = exec.Command("/bin/bash", "-c", `
 		# gobpf currently uses global kprobes via debugfs/tracefs and not the Perf
 		# Event file descriptor based kprobe (Linux >=4.17). So unfortunately, kprobes
 		# can remain from previous executions. Ideally, gobpf should implement Perf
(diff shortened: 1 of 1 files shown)


pkg/gadgets/traceloop/gadget.go
```diff
@@ -126,7 +126,7 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
 		# Indeed, traceloop creates a lof of tracers which each has maps and events:
 		# https://github.com/kinvolk/traceloop/blob/6f4efc6fca46d92c75f4ec4e6c6e1d829bdeaddf/bpf/straceback-guess-bpf.h#L27-L28
 		# So, this can generate a lof ot opened files.
-		ulimit -n unlimited
+		ulimit -n hard
```
alban, line 129: What is the current hard limit on the system where you have a problem?

I wonder if both the hard and soft limit should be changed instead:

```suggestion
		ulimit -Hn unlimited
		ulimit -Sn unlimited
```

eiffel-fl: The current hard limit is 524288.
I cannot use unlimited:

```bash
root [ /home/renaud/inspektor-gadget/integration ]# kubectl exec -ti -n gadget gadget-5mk6h -- bash
root@cbl-mariner:/# ulimit -Sn
1024
root@cbl-mariner:/# ulimit -Hn
524288
root@cbl-mariner:/# ulimit -n
1024
root@cbl-mariner:/# ulimit -Sn unlimited
bash: ulimit: open files: cannot modify limit: Invalid argument
root@cbl-mariner:/# ulimit -Hn unlimited
bash: ulimit: open files: cannot modify limit: Operation not permitted
```

Hence, the use of `hard` (which is big enough to satisfy our needs).

alban: The default value of the hard limit can be configured differently on different systems & different container runtimes. Here they mention a default hard limit of 2560:
https://github.com/containers/common/blob/8eeaf9f5c09a212d14891cb8a0f2a3889d05ae4d/pkg/config/containers.conf#L79

In Podman 1.9, the hard limit used to be 1024 before being changed to 524288: https://github.com/containers/podman/issues/6093#issuecomment-624254340

So I am worried it might not be enough. But I don't have better suggestions for now. I suggest  to merge this patch and revisit if it causes problems.

eiffel-fl: Clearly, it does not seem to have a silver bullet.
If we want to increase this limit, we can run the container as `privileged`, but we worked on this to restrain its privileges as much as possible so I do not want to choose this solution.

Anyway, I am not really worried about it: one day we will rework `traceloop` so this solution is only temporary.

alban (approved):

PR #566, opened 2022-03-15
ci: Push to an Azure Container Registry

blanquicet:
# Push to an Azure Container Registry

This PR modifies the Inspektor Gadget CI workflow to push generated images to the Azure Container Registry. 

Notice that this initial use of the Azure Container Registry does not replace the uses of the current registry but starts by pushing the images to both registries.

## How to use

This PR adds a README file that describes how the Azure Container Registry was created and configured. And, what actions secrets need to be added to make this feature work.


.github/workflows/inspektor-gadget.yml
```diff
@@ -216,7 +216,7 @@ jobs:
         path: /home/runner/work/inspektor-gadget/inspektor-gadget/${{ matrix.local-gadget-target }}.tar.gz
 
   build-docker-image:
```
mauriciovasquezbernal, line 218: nit: even if not related to this commit. `build-docker-image` is not precise as it's a container image and not a docker image.

.github/workflows/inspektor-gadget.yml
```diff
@@ -415,6 +423,19 @@ jobs:
...
         make -C integration push
+    - name: Login to Azure Container Registry
+      if: startsWith(github.ref, 'refs/tags/v')
+      uses: docker/login-action@v1
+      with:
+        registry: ${{ secrets.AZURE_REGISTRY_CONTAINER_REPO }}
+        username: ${{ secrets.AZURE_REGISTRY_USERNAME }}
+        password: ${{ secrets.AZURE_REGISTRY_PASSWORD }}
+    - name: Push Integration Test Image to Azure Container Registry
+      if: startsWith(github.ref, 'refs/tags/v')
```
mauriciovasquezbernal, line 434: I think it'll fail because there is not a `$REGISTRYtest.azurecr.io/gadget` created -> https://github.com/kinvolk/inspektor-gadget/blob/main/integration/Makefile#L15. 

I'm not sure how we should handle, different options I see are:

- create a second `...test` registry 
- don't push this image ti ACR 
- push on the same registry but appending a suffix (like `-test`)

@alban do you have any opinion?

mauriciovasquezbernal (commented):
I'm wondering if we really need to push images to both repositories, it seems to me that switching repository doesn't present a big risk and we can just change without duplicating the logic. However I think we should delay that change until next release is done, so we have the whole next release development period to test if this registry is working fine. 

ping @marga-kinvolk and @alban as you could have more comments here.

.github/workflows/inspektor-gadget.yml
blanquicet: I agreed. I will take advantage of this PR to change it.

.github/workflows/inspektor-gadget.yml
blanquicet: > I think it'll fail because there is not a `$REGISTRYtest.azurecr.io/gadget` created
>
I had tested it and it worked because it is not `$REGISTRYtest.azurecr.io/gadget` but `$REGISTRY.azurecr.io/gadgettest`. It means that `gadgettest` is a different repository (which does not need to already exist) within the same registry `$REGISTRY.azurecr.io`, where we are pushing the gadget container image (which uses the repository `gadget`). It is the same for the Docker registry.

blanquicet: Maybe what is creating confusion is that I didn't use a specific secret to store the Azure registry. I mean, currently, we use the following secrets:

- `CONTAINER_REGISTRY`: The URL of the registry which we use to login.
- `CONTAINER_REPO`: The URL of the repository which we use to push the gadget container images and the integration test image after appending `test`.

Instead, for the Azure registry, I used one single secret named `AZURE_REGISTRY_CONTAINER_REPO` to store the repository which I used for both, login and push.

alban:
> I'm wondering if we really need to push images to both repositories, it seems to me that switching repository doesn't present a big risk and we can just change without duplicating the logic.

I agree.

Also, I would like the push mechanism to be usable by forks and it should be optional. Does the code before this PR work on forks when the GitHub secrets are selecting an Azure Container Registry?

blanquicet:
> Does the code before this PR work on forks when the GitHub secrets are selecting an Azure Container Registry?
>
Yes, I tested it and it worked. To test it, I set `CONTAINER_REGISTRY`, `CONTAINER_REGISTRY_USERNAME`, `CONTAINER_REGISTRY_PASSWORD` as I described in the [README](https://github.com/kinvolk/inspektor-gadget/blob/jose/azure-cr-push/.github/workflows/README.md) file of this PR. And, set `CONTAINER_REPO` equal to `CONTAINER_REGISTRY`.

> I would like the push mechanism to be usable by forks and it should be optional. 
>
I agree it would be useful for people testing IG on forks.

.github/workflows/README.md
```diff
@@ -0,0 +1,53 @@
...
+Now, we need to add the following actions secrets to the GitHub repository:
+
+- `AZURE_REGISTRY_CONTAINER_REPO`: The `gadget` repository server name of our
+  registry (all lowercase): `$REGISTRY.azurecr.io/gadget`.
+- `AZURE_REGISTRY_USERNAME`: The `clientId` from the JSON output from the
+  service principal creation.
+- `AZURE_REGISTRY_PASSWORD`: The `clientSecret` from the JSON output from the
+  service principal creation.
+
+With the secrets on place, the `build-docker-image` and `test-integration` jobs
```
mauriciovasquezbernal, line 39:
```suggestion
With the secrets on place, the `build-container-image` and `test-integration` jobs
```

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: I have no idea why I was confused. Probably I was thinking that the commands in the readme were to create a container **repository** instead of a **registry**, so what both repositories have to be created by hand, but it turns out that the registry is created manually and repositories are created automatically when pushing.

.github/workflows/README.md
```diff
@@ -0,0 +1,53 @@
+# Inspektor Gadget CI Workflow
```
mauriciovasquezbernal, line 1: I suppose this file should be moved to docs/ as done in https://github.com/kinvolk/inspektor-gadget/pull/561.

blanquicet: Done

.github/workflows/README.md
blanquicet: Done

blanquicet:
I rebased the PR on top of https://github.com/kinvolk/inspektor-gadget/pull/561 which already modifies the CI. 

As agreed, I also updated the PR to not push to both registries but only ACR once the repo secrets will be updated. I updated the [docs/ci.md](https://github.com/kinvolk/inspektor-gadget/blob/jose/azure-cr-push/docs/ci.md) describing how to configure the repo secrets to push to ACR.

I also took the opportunity to make all the `CONTAINER_*` secrets optional so that CI workflow does not fail on forks and its output is cleaner. However, if the gadget container image is not pushed to a registry, the integration tests will not be able to run, forcing the contributors to either use the official IG repo or configure the secrets on their forks. We have been discussing about several improvements to this process in the `#inspektor-gadget` Kubernetes Slack channel but none of them have been implemented in this PR. They will be managed in future PRs.

.github/actions/run-integration-tests/action.yml
```diff
@@ -2,14 +2,18 @@ name: "Run integration tests"
 description: "Run Inspektor Gadget integration tests"
 
 inputs:
+  push:
+    description: 'Defines whether the generated integration tests image has to be push to the container registry or not'
+    required: true
+    default: 'false'
```
eiffel-fl, line 8: What about using boolean [`false`](https://docs.github.com/en/actions/learn-github-actions/expressions#literals) instead?
Then, you do not need later to compare string.

.github/workflows/inspektor-gadget.yml
```diff
@@ -414,15 +427,19 @@ jobs:
...
   test-integration:
     name: Integration tests
-    # level: 1
-    needs: [test-unit, build-kubectl-gadget, build-local-gadget, build-docker-image]
+    # level: 2
+    needs: [check-secrets, test-unit, build-kubectl-gadget, build-local-gadget, build-container-image]
     runs-on: ubuntu-latest
+    # It is useless to run the integration tests if the container image was not
+    # pushed to a container registry. They will indeed fail.
+    if: needs.check-secrets.outputs.container_registry == 'true'
```
eiffel-fl, line 442: I checked the CI graph because I was afraid that making image building a level 1 job would increase the whole pipeline time and I saw the integration tests were skipped:
https://github.com/kinvolk/inspektor-gadget/actions/runs/2064286450
Was this on purpose to gain development time?

eiffel-fl (commented):
It seems cool even though I do not know what to do regarding it.
I think it is OK to get it because we will use `${{ secrets.* }}` while `ghcr.io` will use `$GITHUB_TOKEN` and I think we decided to use both.

Please see my comment regarding skipped integration tests:

.github/workflows/inspektor-gadget.yml
blanquicet: It was skipped because we are currently pushing to DockerHub and it does not need to have the `CONTAINER_REGISTRY` secret set. Therefore, I will remove it from [this check](https://github.com/kinvolk/inspektor-gadget/blob/213d485b60d8c4add0a6c63766c907c80c0004aa/.github/workflows/inspektor-gadget.yml#L333) because if the user decides to use DockerHub it will not work.

On the other hand, the whole pipeline time was increased by the time required for its need to be completed, in this case, it is the `check-secrets` job, which means 2 seconds.

.github/actions/run-integration-tests/action.yml
blanquicet: It doesn't work actually, see https://github.com/actions/runner/issues/1483. There are some workarounds but I found it easier to just check against `'true'`.

blanquicet:
I added `state/hold` flag because we decided to not use Azure Container Registry but GitHub Container Registry given that it is a neutral location for an OSS project.

However, the documentation about how to push to an ACR repo could be useful so we need to decide whether to keep it or not. Maybe a Wiki could be a better place.

marga-kinvolk:
We don't have a wiki, I think the "next best" is to just add it as an md file in our docs directory.

marga-kinvolk:
BTW, the move to GHCR is in #597

mauriciovasquezbernal:
Closing this one as we're moving to GHCR instead -> https://github.com/kinvolk/inspektor-gadget/pull/597.

PR #582, opened 2022-03-21
Fix tests for network-policy-advisor

alban:
It was not testing anything because the source files were in a different
directory than expected. Moreover, it was dependent on the current
working directory, so the following was running tests differently:

* `go test ./pkg/gadgets/networkpolicy/...`
* `cd pkg/gadgets/networkpolicy ; go test ./...`

Fixes 7b68ec33b594934fed7c4d8ad1db5e2e8e351d00


alban:
The CI failure seems unrelated to my patch.

blanquicet:
> The CI failure seems unrelated to my patch.
>
Yeah, it happens sometimes and it is quite frustrating to need to re-run the CI without knowing what happened. 

I have been thinking about it and I think it would be useful to have the logs of the gadget pods to be sure there wasn't an error that wasn't sent to the CLI (I added it in PR https://github.com/kinvolk/inspektor-gadget/pull/556). And, the logs or output of the `test-pod` to be sure that the command that has to create the events for triggering the gadget actually worked (see issue https://github.com/kinvolk/inspektor-gadget/issues/567). 

What do you think? What else do you think we can do to have enough information to debug this situation?

alban:
> What do you think? What else do you think we can do to have enough information to debug this situation?

I think your suggestions are good for this failure.

For other tests, it might make sense to print the list of pods (`kubect get pods -A`) and the last events (`kubectl get events -A`) like in [command.go#L110](https://github.com/kinvolk/inspektor-gadget/blob/1b990430457ff17665fdf44ffc705104e0ebfc83/integration/command.go#L110). That could include explanations why a pod didn't start.

alban:
The network-policy-advisor tests are being rewritten and fixed in https://github.com/kinvolk/inspektor-gadget/pull/577. If PR 577 is merged, we don't need this PR anymore.

PR #584, opened 2022-03-25
Remove libbpf-tools from container image

mauriciovasquezbernal:
Those are not needed anymore as we implemented such tools directly
in Inspektor Gadget. This changes reduces the container image size by about 60MB.


eiffel-fl (approved):
A negative patch which reduces size of our container image?! Let's merge it!
Thank you for it!

I was just a bit concerned by the profile gadget which is still using `bcck8s.go` file but I tested it and it still works!

alban (approved):

PR #585, opened 2022-03-25
cmd/kubectl-gadget: Wait for gadget namespace to be removed

mauriciovasquezbernal:
This commit adds logic to wait until the gadget namespace is removed
before returning. This is useful to guarantee that Inspektor Gadget
is completely remove when the command returns. This behaviour is
enabled by default and can be turn off by using '--wait=false'

Fixes https://github.com/kinvolk/inspektor-gadget/issues/573

### DIFF
--- a/cmd/kubectl-gadget/undeploy.go
+++ b/cmd/kubectl-gadget/undeploy.go
@@ -24,8 +24,14 @@ import (
 	"github.com/kinvolk/inspektor-gadget/pkg/k8sutil"
 	"github.com/spf13/cobra"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
 
 	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 )
@@ -37,8 +43,21 @@ var undeployCmd = &cobra.Command{
 	SilenceUsage: true,
 }
 
+var undeployWait bool
+
+const (
+	gadgetNamespace string = "gadget"
+	timeout         int    = 30
+)
+
 func init() {
 	rootCmd.AddCommand(undeployCmd)
+	undeployCmd.PersistentFlags().BoolVarP(
+		&undeployWait,
+		"wait", "",
+		true,
+		"wait for all resources to be deleted before returning",
+	)
 }
 
 func runUndeploy(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


cmd/kubectl-gadget/undeploy.go
```diff
@@ -150,9 +177,36 @@ again:
...
 	}
 
+	if undeployWait && watcher != nil {
+		ch := watcher.ResultChan()
+		defer watcher.Stop()
+
+		timer := time.NewTimer(time.Duration(timeout) * time.Second)
+
+	loop:
+		for {
```
eiffel-fl, line 187: Isn't there an easy way to avoid busy waiting?
Something like what is done in `waitForCondition()` but for NameSpace (hoping the API will be easier to handle)?

mauriciovasquezbernal: It's not busy waiting, notice that there is a channel read on L189. I used a loop in case we receive multiple events for the `gadget` namespace before the delete one. Actually this code is adapted from [untilWithouRetry](https://github.com/kinvolk/inspektor-gadget/blob/8c5811050cb8cf9dcd588b520d5f36e0a44da055/cmd/kubectl-gadget/utils/trace.go#L414), let me see if we can reuse that function there.

mauriciovasquezbernal: I reworked the PR to use `watchtools.Until()`, so I don't have to implement this logic by hand.

eiffel-fl (approved):
I already gave you my approval so this enhanced version is even more approved!
I may consider using `watchtools.Until()` instead of `watchtools.UntilWithoutRetry()` because we would not have to keep a copy of this function in our code.

Just see my comment regarding `watch.Error` (I never know if we should deal with them particularly or treat them like other events).

mauriciovasquezbernal:
> I may consider using `watchtools.Until()` instead of `watchtools.UntilWithoutRetry()` because we would not have to keep a copy of this function in our code.

Right. Also  `watchtools.Until()` seems to be more reliable than `watchtools.UntilWithoutRetry()` .

cmd/kubectl-gadget/undeploy.go
```diff
@@ -135,24 +151,63 @@ again:
...
+
+	if len(list.Items) == 0 {
+		goto out
+	}
+
 	err = k8sClient.CoreV1().Namespaces().Delete(
 		context.TODO(), "gadget", metav1.DeleteOptions{},
 	)
 	if err != nil {
 		errs = append(errs, fmt.Sprintf("failed to remove \"gadget\" namespace: %s", err))
```
blanquicet, line 180: Why not?

```suggestion
		errs = append(errs, fmt.Sprintf("failed to remove \"gadget\" namespace: %s", err))
		goto out
```

blanquicet (approved):
I tested it and it is working as expected. Some minor comments but LGTM!

I was wondering if we should print the steps we are executing (maybe in another PR):
```bash
$ kubectl gadget undeploy
Removing traces ...
Removing CRD ...
Removing cluster role binding ...
Removing cluster role ...
Removing namespace ...
Inspektor Gadget successfully removed
```

mauriciovasquezbernal:
> I was wondering if we should print the steps we are executing (maybe in another PR):

I put them there so the user has more information about what's going on.

cmd/kubectl-gadget/undeploy.go
```diff
@@ -112,46 +132,99 @@ again:
...
 
 	// 3. gadget cluster role binding
+	fmt.Println("Removing cluster role binding...")
 	err = k8sClient.RbacV1().ClusterRoleBindings().Delete(
 		context.TODO(), "gadget-cluster-role-binding", metav1.DeleteOptions{},
 	)
-	if err != nil {
+	if err != nil && !errors.IsNotFound(err) {
 		errs = append(
 			errs, fmt.Sprintf("failed to remove \"gadget\" cluster role bindings: %s", err),
```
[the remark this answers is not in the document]
mauriciovasquezbernal: no actually. It's the name oft eh cluster role and not the namespace. It's a coincidence that they both are named the same.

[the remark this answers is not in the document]
blanquicet: Oops, you are right!

mauriciovasquezbernal:
I think it's a good improvement to have in the next release. I'm merging for v0.5.0.

PR #593, opened 2022-03-31
kubectl-gadget: Add missing capabilities needed in ARO

mauriciovasquezbernal:
- BCC python-based tools were failing to start because the container was missing the sys_mod capability. 
It seems it was only happening in ARO because that system don't have the kernel headers available.
- NET_RAW: Needed by gadgets that open a raw sock like dns and snisnoop

Fixes https://github.com/kinvolk/inspektor-gadget/issues/570
Fixes https://github.com/kinvolk/inspektor-gadget/issues/568
Fixes https://github.com/kinvolk/inspektor-gadget/issues/572
Fixes https://github.com/kinvolk/inspektor-gadget/issues/569


alban (approved):
LGTM!

Then some `t.Skip` will need to be removed from #561.

mauriciovasquezbernal:
I'm merging before the next release as it solves and bug and should not create any regressions.

PR #594, opened 2022-03-31
kubectl-gadget: Fix typo in fsslower command

mauriciovasquezbernal:
The parameter is called "type" and not filesystem.

### DIFF
--- a/cmd/kubectl-gadget/trace/fsslower.go
+++ b/cmd/kubectl-gadget/trace/fsslower.go
@@ -42,7 +42,7 @@ var fsslowerCmd = &cobra.Command{
 	Short: "Trace open, read, write and fsync operations slower than a threshold",
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		if fsslowerFilesystem == "" {
-			return utils.WrapInErrMissingArgs("--filesystem")
+			return utils.WrapInErrMissingArgs("--filesystem / -f")
 		}
 
 		found := false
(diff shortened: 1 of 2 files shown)


cmd/kubectl-gadget/fsslower.go
```diff
@@ -41,7 +41,7 @@ var fsslowerCmd = &cobra.Command{
 	Short: "Trace open, read, write and fsync operations slower than a threshold",
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		if fsslowerFilesystem == "" {
-			return utils.WrapInErrMissingArgs("--filesystem")
+			return utils.WrapInErrMissingArgs("--type")
```
alban, line 44:
```suggestion
			return utils.WrapInErrMissingArgs("--type / -t")
```

For consistency with `utils.WrapInErrInvalidArg` below, line 58? Or use `"--type"` everywhere.

blanquicet, line 44: IMO, we should also change the Trace CR parameter. Otherwise, users using the kubectl-gadget CLI and trace will have different name parameters. WDYT?

mauriciovasquezbernal: I changed it but I'm not sure `--type` is the best name here, however I want to keep it exactly the same as BCC fsslower https://github.com/iovisor/bcc/blob/5f2a64ec101b2916558bb934c436d4a6d4ed157c/libbpf-tools/fsslower.c#L107.

mauriciovasquezbernal: Makes sense, thanks!

blanquicet: Yeah, `type` seems too generic 🙁.

blanquicet: Proposal: `fs-type`

blanquicet: @mauriciovasquezbernal what is your plan for this PR?

blanquicet (approved):
LGTM

PR #596, opened 2022-04-01
gadgettracermanager: Fix possible deadlock

mauriciovasquezbernal:


eiffel-fl (approved):
One line and it removes a potential deadlock: perfect!
I do not know when you plan to merge it, but adding this fix to the release seems a good idea :D.

mauriciovasquezbernal:
Merging as this fixes a potential problem.

PR #602, opened 2022-04-05
Provide dockerfile for only CO-RE gadgets

mauriciovasquezbernal:
Inspektor Gadget has two different kind of gadgets:
1: standard gadgets. These tools come from BCC, use Python and are
   compiled in the target machine before loading. They require to ship a
   whole compiling chain (clang) and libbcc. These tools have the
   advantage that they only require the kernel headers to be installed
   and work on systems without BTF support.
2: CO-RE gadgets: These tools use Compile Once - Run Everywhere. They
   are more lighter but require BTF info to be present.

A lot of distributions are enabling BTF and tools like BTFHub and BTFGen
make it possible to run those gadgets in systems without BTF. This
commit introduces a new dockerfile that provides support only for CO-RE
gadgets. The advantages of this approach are:
- The size of the image is reduced from ~800MB to ~200MB (when using
  alpine).
- A lot of things is removed from the container reducing the attack
  surface.
- The compilation is faster.
- Other images like fedora, debian, alpine can be used as base.

--- 

NOTE: This PR only provides the docker file, it's not used by the CI nor when deploying IG. We can consider those steps later on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,12 @@ gadget-container:
 	docker build -t $(CONTAINER_REPO):$(IMAGE_TAG) -f gadget.Dockerfile \
 		--build-arg ENABLE_BTFGEN=$(ENABLE_BTFGEN) .
 
+# gadget container core
+.PHONY: gadget-container-core
+gadget-container-core:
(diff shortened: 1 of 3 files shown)


mauriciovasquezbernal:
Replaced by https://github.com/kinvolk/inspektor-gadget/pull/608.

PR #603, opened 2022-04-07
entrypoint.sh: Set current directory to a valid path

mauriciovasquezbernal:
The previous logic was changing the directory to a path and then
removing it, which was causing issues when calling "getcwd" in BCC
python scripts.

This commit fixes the logic by using pushd/popd instead of plain cd, it
also changes the current directoy to / before executing
gadgettracermanager to be sure this problem doesn't happen again.

This solves some issues when running standard tools on ARO clusters. 

cc  @blanquicet

### DIFF
--- a/gadget-container/entrypoint.sh
+++ b/gadget-container/entrypoint.sh
@@ -104,7 +104,9 @@ if [ "$ID" = "rhcos" ] && [ ! -d "/usr/src/kernels/$KERNEL" ]; then
     curl --max-time 40 -fsSLo $RPMDIR/$RPM $REPO/$RPM || rm -f $RPMDIR/$RPM
 
     if test -f $RPMDIR/$RPM; then
-      cd $RPMDIR && rpm2cpio $RPM | cpio --quiet -i
+      pushd $RPMDIR
+      rpm2cpio $RPM | cpio --quiet -i
+      popd
 
       # In the gadget image, /usr/src is a symlink to /host/usr/src (see gadget.Dockerfile).
       # But in the case of RHCOS, remove the symlink and install files in the container.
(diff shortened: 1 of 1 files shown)


gadget-container/entrypoint.sh
```diff
@@ -104,7 +104,9 @@ if [ "$ID" = "rhcos" ] && [ ! -d "/usr/src/kernels/$KERNEL" ]; then
     curl --max-time 40 -fsSLo $RPMDIR/$RPM $REPO/$RPM || rm -f $RPMDIR/$RPM
 
     if test -f $RPMDIR/$RPM; then
-      cd $RPMDIR && rpm2cpio $RPM | cpio --quiet -i
+      pushd $RPMDIR
+      rpm2cpio $RPM | cpio --quiet -i
+      popd
```
blanquicet, line 109: > The previous logic was changing the directory to a path and then removing it, which was causing issues when calling "getcwd" in BCC python scripts.
>
Oops, I didn't know it could be a problem. Good to know for next time. Thanks!

blanquicet (approved):
I run Biolantecy, Capabilities and Tcptracer integration tests locally and they are now passing. I also rebased https://github.com/kinvolk/inspektor-gadget/pull/561 on top of this PR and the CI is now passing! 

After this PR, the integration tests on ARO are only skipping Biotop (https://github.com/kinvolk/inspektor-gadget/issues/589) and Fsslower (https://github.com/kinvolk/inspektor-gadget/issues/571) tests. And of course socket and process collector but it is because of that kernel does not support iterators.

I have just one concern, otherwise it LGTM!

gadget-container/entrypoint.sh
```diff
@@ -219,6 +221,8 @@ else
 fi
 
 echo "Starting the Gadget Tracer Manager..."
+# change directory before running gadgettracermanager
+cd /
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's fine to change it before so we're sure we don't have the same issue again.

gadget-container/entrypoint.sh
mauriciovasquezbernal: I was a bit surprising for me too and took some time to realize what was going on.

mauriciovasquezbernal:
Merging as this fixes a bug.

PR #604, opened 2022-04-08
containerutils: Create CRIClient that implements the common CRI methods

blanquicet:
# Create CRIClient that implements the container runtime methods

This PR does not change the current behaviour. It just tries to reorganize the current code by using a `CRIClient` that will allow us to avoid maintaining the same code separately for cri-o and contained. 

For what concerns Docker, it implements its own methods because the docker CRI (dockershim), which would have used the `CRIClient`'s methods, does not provide a way to retrieve the container PID. Therefore, we had to use directly the docker API.


pkg/gadgettracermanager/containerutils/containerutils.go
```diff
@@ -72,9 +76,33 @@ uint64_t get_cgroupid(char *path) {
...
+		runtimeClient = crio.NewCrioClient(crio.DefaultRuntimeEndpoint)
+	default:
+		return nil, fmt.Errorf("unknown container runtime: %s (available %s)",
+			runtime, strings.Join(ContainerRuntimes, ", "))
+	}
+
+	err := runtimeClient.Initialize()
+	if err != nil {
+		return nil, err
+	}
```
mauriciovasquezbernal, line 104: I'm not a big fan of this `Initialize` approach. What about moving this logic to the `NewXXX()` functions? Something like -> https://github.com/kinvolk/inspektor-gadget/commit/54eeee93f553cd3889b9a9476012053af93bbfc2?

blanquicet: Yeah, I like your proposal. Applied!

pkg/gadgettracermanager/containerutils/containerd/containerd.go
```diff
@@ -15,80 +15,28 @@
...
 
-func NewContainerdClient(path string) (*ContainerdClient, error) {
-	conn, err := grpc.Dial(
-		path,
-		grpc.WithInsecure(),
-		grpc.WithDialer(func(addr string, timeout time.Duration) (net.Conn, error) {
-			return net.DialTimeout("unix", path, DefaultTimeout)
-		}),
-	)
+func NewContainerdClient(endPoint string) (runtimeclient.ContainerRuntimeClient, error) {
```
mauriciovasquezbernal: I know I suggested it but `endpoint` is better than `endPoint`.

pkg/gadgettracermanager/containerutils/runtime-client/interface.go
```diff
@@ -0,0 +1,142 @@
...
+// ContainerRuntimeClient defines the interface to communicate with the
+// different container runtimes.
+type ContainerRuntimeClient interface {
+	// PidFromContainerID returns the pid1 of the container identified by the
+	// specified ID. In case of errors, it can return -1 if there is any problem
+	// retrieving the container information or parsing the response. Or, 0 if
+	// the pid is not present within the retrieved information.
+	PidFromContainerID(containerID string) (int, error)
+
+	// Close tears down the connection with the client.
```
mauriciovasquezbernal: `with the client` -> `with the container runtime`?

mauriciovasquezbernal (approved):
Besides the naming of a variable LGTM.

blanquicet:
Rebased

pkg/gadgettracermanager/containerutils/runtime-client/interface.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return -1, fmt.Errorf("failed extracting pid from container status reply: %w", err)
+	}
+
+	return infoContent.Pid, nil
+}
+
+func (c *CRIClient) PidFromContainerID(containerID string) (int, error) {
+	// If ID contains a prefix, it must match the runtime name: "<name>://<ID>"
+	split := regexp.MustCompile(`://`).Split(containerID, -1)
```
alban: You could just use strings.SplitN:
https://go.dev/play/p/JVvPYNR3C9Q

```
	id := "docker://abcde"
	fmt.Printf("%+v\n", strings.SplitN(id, "://", 2))
```

Ditto in the other occurrences of regexp.MustCompile(`://`)

pkg/gadgettracermanager/k8s/k8s.go
```diff
@@ -59,9 +57,17 @@ func NewK8sClient(nodeName string) (*K8sClient, error) {
 		return nil, fmt.Errorf("failed to get node %w", err)
 	}
 
-	// get a CRI client to talk to the CRI handling pods in this node
-	// TODO: when to close it?
-	criClient, err := newCRIClient(node)
+	runtimeVersion := node.Status.NodeInfo.ContainerRuntimeVersion
+	list := strings.Split(runtimeVersion, "://")
```
alban, line 61:
```suggestion
	list := strings.SplitN(runtimeVersion, "://", 2)
```

pkg/gadgettracermanager/containerutils/docker/docker.go
```diff
@@ -19,52 +19,59 @@ import (
...
-	}
-
-	return nil
-}
-
 func (c *DockerClient) PidFromContainerID(containerID string) (int, error) {
-	if !strings.HasPrefix(containerID, "docker://") {
-		return -1, fmt.Errorf("invalid CRI %s, it should be docker", containerID)
+	// If ID contains a prefix, it must match the runtime name: "<name>://<ID>"
+	split := regexp.MustCompile(`://`).Split(containerID, -1)
```
alban, line 64: Could use strings.SplitN

alban (approved):
Small suggestions. LGTM after that

pkg/gadgettracermanager/k8s/k8s.go
```diff
@@ -59,9 +57,10 @@ func NewK8sClient(nodeName string) (*K8sClient, error) {
 		return nil, fmt.Errorf("failed to get node %w", err)
 	}
 
-	// get a CRI client to talk to the CRI handling pods in this node
-	// TODO: when to close it?
-	criClient, err := newCRIClient(node)
+	// Get a runtime client to talk to the container runtime handling pods in
+	// this node. TODO: when to close it?
```
mauriciovasquezbernal, line 61: No related to this PR: This TODO hasn't been solved, I think were leaking the connection to the client in https://github.com/kinvolk/inspektor-gadget/blob/a2a95ec9729b0203cb30fb60ff4073aa64b88df1/pkg/container-collection/options.go#L219, there should be a call to `k8sClient.CloseRuntimeClient()` there before returning. The ContainerCollectionsOptions are also leaking as they never call any Close() method, I opened https://github.com/kinvolk/inspektor-gadget/issues/624 to track this down.

mauriciovasquezbernal (commented):
Few more comments, but I think it's almost ready to be merged.

pkg/gadgettracermanager/containerutils/runtime-client/interface.go
```diff
@@ -0,0 +1,149 @@
...
+		Pid int `json:"pid"`
+	}
+
+	var infoContent InfoContent
+	err := json.Unmarshal([]byte(info), &infoContent)
+	if err != nil {
+		return -1, fmt.Errorf("failed extracting pid from container status reply: %w", err)
+	}
+
+	return infoContent.Pid, nil
```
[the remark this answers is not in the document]
blanquicet: Yeah, it is because the [caller](https://github.com/kinvolk/inspektor-gadget/blob/4cfdb72d4edd149a8314311088ffab846b5b7b2d/pkg/gadgettracermanager/k8s/k8s.go#L119-L122) is discarding it. I will move this check inside `PidFromContainerID()` so that caller just needs to check for no error.

pkg/gadgettracermanager/containerutils/runtime-client/interface.go
```diff
@@ -0,0 +1,149 @@
...
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+)
+
+// ContainerRuntimeClient defines the interface to communicate with the
+// different container runtimes.
+type ContainerRuntimeClient interface {
+	// PidFromContainerID returns the pid1 of the container identified by the
+	// specified ID. In case of errors, it can return -1 if there is any problem
+	// retrieving the container information or parsing the response. Or, 0 if
+	// the pid is not present within the retrieved information.
```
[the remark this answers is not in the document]
blanquicet: Yeah. I will use only -1 in all the cases. See also this [comment](https://github.com/kinvolk/inspektor-gadget/pull/604#discussion_r858725694).

mauriciovasquezbernal (approved):
I think it's good to go!

PR #605, opened 2022-04-08
integration-tests: Print the output before making the test fail

blanquicet:
# Print the output before making the test fail

Printing the output no matter whether the test failed or not allows us to understand why it failed if it did.

### Before this PR
```bash
$ sudo -E go test -tags withebpf -run TestBiolatency -integration -no-deploy ./...
using random seed: 1649428227236645175
--- FAIL: TestBiolatency (15.37s)
    command.go:204: Run command: id=$($KUBECTL_GADGET biolatency start --node $(kubectl get node --no-headers | cut -d' ' -f1)); sleep 15; $KUBECTL_GADGET biolatency stop $id
    command.go:207: exit status 1
FAIL
exit status 1
FAIL    github.com/kinvolk/inspektor-gadget/integration 15.369s
```

### After this PR
```bash
$ sudo -E go test -tags withebpf -run TestBiolatency -integration -no-deploy ./...
using random seed: 1649428324109800923
--- FAIL: TestBiolatency (15.44s)
    command.go:204: Run command: id=$($KUBECTL_GADGET biolatency start --node $(kubectl get node --no-headers | cut -d' ' -f1)); sleep 15; $KUBECTL_GADGET biolatency stop $id
    command.go:207: Command returned:
        Error: unknown command "aks-agentpool-25084143-vmss000001" for "kubectl-gadget biolatency start"
        Error: missing required arguments: <trace-id>


    command.go:210: exit status 1
FAIL
exit status 1
FAIL    github.com/kinvolk/inspektor-gadget/integration 15.438s
```

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -203,11 +203,12 @@ func (c *command) run(t *testing.T) {
 
 	t.Logf("Run command: %s\n", c.cmd)
 	err := c.command.Run()
+
+	t.Logf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
+
 	if err != nil {
 		t.Fatal(err)
 	}
-
-	t.Logf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
 
 	err = c.verifyOutput()
 	if err != nil {
@@ -221,11 +222,12 @@ func (c *command) runWithoutTest() error {
 
 	c.createExecCmd()
 	err := c.command.Run()
+
+	fmt.Printf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
+
 	if err != nil {
 		return err
 	}
-
-	fmt.Printf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
 
 	return c.verifyOutput()
 }
@@ -275,11 +277,12 @@ func (c *command) stop(t *testing.T) {
 	// received by all the processes with such PGID, in our case, the process of
 	// /bin/sh and c.cmd.
 	err := syscall.Kill(-c.command.Process.Pid, syscall.SIGTERM)
+
+	t.Logf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
+
 	if err != nil {
 		t.Fatal(err)
 	}
-
-	t.Logf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
 
 	err = c.verifyOutput()
 	if err != nil {


eiffel-fl (commented):
Your goal is to print more output when tests are runned with `-test.v`, right?
If so, maybe you should move this one too:
https://github.com/kinvolk/inspektor-gadget/pull/605/commits/4d73bfc25d6537634f7a02ddd63af377d8f15709#diff-aaade3782dd35eb7898f3403d9c0ae1345621a6442ddff5145b0fdae304ed6afR285

blanquicet:
> Your goal is to print more output when tests are runned with `-test.v`, right?
>
No, it also works without using `-test.v`. My goal is to better understand what was the error when the `command.Run()` or `command.Start()` failed. I added an example on the PR description so that it is clearer 🙂.

> If so, maybe you should move this one too: [4d73bfc#diff-aaade3782dd35eb7898f3403d9c0ae1345621a6442ddff5145b0fdae304ed6afR285](https://github.com/kinvolk/inspektor-gadget/commit/4d73bfc25d6537634f7a02ddd63af377d8f15709#diff-aaade3782dd35eb7898f3403d9c0ae1345621a6442ddff5145b0fdae304ed6afR285)
> 
Mmmm the link brings me the commit of this PR, it doesn't show any additional diff 🙁 .

integration/command.go
```diff
@@ -252,6 +254,7 @@ func (c *command) start(t *testing.T) {
 
 	err := c.command.Start()
 	if err != nil {
+		fmt.Printf("Command returned:\n%s\n%s\n", c.stderr.String(), c.stdout.String())
```
mauriciovasquezbernal, line 257: Do `c.stderr` and `c.stdout` contain data at this point?

mauriciovasquezbernal (commented):
I think the line on `start()` is useless as at that point there is not any data on those buffers. `cmd.Start()` returns without waiting for the process. I think the correct point to add that information is in [stop()](https://github.com/kinvolk/inspektor-gadget/blob/4d73bfc25d6537634f7a02ddd63af377d8f15709/integration/command.go#L285). Btw,  I think we're missing a [Wait()](https://pkg.go.dev/os/exec#Cmd.Wait) call after `syscall.Kill()`. :thinking:

integration/command.go
blanquicet: Yeah, you are right. This change is useless.

eiffel-fl (approved):
Thank you for your explaination, it is clearer now.
As long as printing is only done when things go wrong, I am fine with it.

You can merge it now, it is small and I think you need it, so better to have it upstream sooner.

PR #607, opened 2022-04-08
integration: add some delay to commands execution

mauriciovasquezbernal:
Many of the integration tests run commands in a loop to generate events.
This commit add some delay between consecutive executions of the
commands to avoid creating so many events making it difficutl to handle
the test log.


eiffel-fl (approved):
This is a good idea!
Otherwise what we can do is to use something like `for unused in {1..X}; do ...; done` instead of `while true;` but all the question is: which value should we give to X.

PR #608, opened 2022-04-08
Create dockerfiles for kubectl-gadget and CO-RE tools

mauriciovasquezbernal:
Please see individual commits for more information.

--- 

It's still not clear if we want to push these images to a container registry, so this PR doesn't handle it.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -193,8 +193,8 @@ jobs:
         name: ${{ matrix.local-gadget-target }}-tar-gz
         path: /home/runner/work/inspektor-gadget/inspektor-gadget/${{ matrix.local-gadget-target }}.tar.gz
 
-  build-docker-image:
-    name: Build gadget container image
+  build-gadget-default-container-image:
+    name: Build gadget default container image
     # level: 0
     runs-on: ubuntu-latest
     steps:
@@ -205,9 +205,9 @@ jobs:
       uses: actions/cache@v2
       with:
         path: /tmp/.buildx-cache
-        key: ${{ runner.os }}-single-buildx-${{ hashFiles('gadget.Dockerfile') }}
+        key: ${{ runner.os }}-docker-default-${{ hashFiles('gadget-default.Dockerfile') }}
         restore-keys: |
-          ${{ runner.os }}-single-buildx
+          ${{ runner.os }}-docker-default-
     - name: Login to Container Registry
       uses: docker/login-action@v1
       with:
@@ -225,17 +225,64 @@ jobs:
             IMAGE_TAG="latest"
         fi
         echo IMAGE_TAG=$IMAGE_TAG >> $GITHUB_ENV
-    - name: Build gadget container
+    - name: Build gadget default container
       uses: docker/build-push-action@v2
       with:
         context: /home/runner/work/inspektor-gadget/inspektor-gadget/
-        file: /home/runner/work/inspektor-gadget/inspektor-gadget/gadget.Dockerfile
+        file: /home/runner/work/inspektor-gadget/inspektor-gadget/gadget-default.Dockerfile
         build-args: |
           ENABLE_BTFGEN=true
         # TODO: how to avoid pushing a container before running integration tests
         # Answer: push to runner registry first, if integration tests are OK
         # push to final registry
         push: true
+        tags: ${{ secrets.CONTAINER_REPO }}:${{ env.IMAGE_TAG }}
+        cache-from: type=local,src=/tmp/.buildx-cache
+        cache-to: type=local,dest=/tmp/.buildx-cache-new
+
+  build-gadget-core-container-image:
+    name: Build gadget CO-RE container image
+    # level: 0
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Cache Docker layers
+      uses: actions/cache@v2
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-docker-core-${{ hashFiles('gadget-core.Dockerfile') }}
+        restore-keys: |
+          ${{ runner.os }}-docker-core-
+    - name: Login to Container Registry
+      uses: docker/login-action@v1
+      with:
+        registry: ${{ secrets.CONTAINER_REGISTRY }}
+        username: ${{ secrets.CONTAINER_REGISTRY_USERNAME }}
+        password: ${{ secrets.CONTAINER_REGISTRY_PASSWORD }}
+    - name: Check out code
+      uses: actions/checkout@v1
+    - name: Set IMAGE_TAG
+      run: |
+        TMP1=${GITHUB_REF#*/}
+        TMP2=${TMP1#*/}
+        IMAGE_TAG=${TMP2//\//-}
+        if [ "$IMAGE_TAG" = "main" ]; then
+            IMAGE_TAG="latest"
+        fi
+        echo IMAGE_TAG=$IMAGE_TAG >> $GITHUB_ENV
+    - name: Build gadget CO-RE container
+      uses: docker/build-push-action@v2
+      with:
+        context: /home/runner/work/inspektor-gadget/inspektor-gadget/
+        file: /home/runner/work/inspektor-gadget/inspektor-gadget/gadget-core.Dockerfile
+        build-args: |
+          ENABLE_BTFGEN=true
+        # TODO: how to avoid pushing a container before running integration tests
+        # Answer: push to runner registry first, if integration tests are OK
+        # push to final registry
+        push: false # TODO: just build the image for the time being
         tags: ${{ secrets.CONTAINER_REPO }}:${{ env.IMAGE_TAG }}
         cache-from: type=local,src=/tmp/.buildx-cache
         cache-to: type=local,dest=/tmp/.buildx-cache-new
@@ -350,7 +397,7 @@ jobs:
   test-integration-aro:
     name: Integration tests on ARO
     # level: 1
-    needs: [check-secrets, test-unit, build-kubectl-gadget, build-local-gadget, build-docker-image]
+    needs: [check-secrets, test-unit, build-kubectl-gadget, build-local-gadget, build-gadget-default-container-image]
     # Run this job only if an ARO cluster is available on repo secrets. See
     # docs/ci.md for further details.
     if: needs.check-secrets.outputs.aro == 'true'
@@ -383,7 +430,7 @@ jobs:
   test-integration:
     name: Integration tests
     # level: 1
-    needs: [test-unit, build-kubectl-gadget, build-local-gadget, build-docker-image]
+    needs: [test-unit, build-kubectl-gadget, build-local-gadget, build-gadget-default-container-image]
     runs-on: ubuntu-latest
     steps:
     - name: Check out code
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -35,10 +35,19 @@ kubectl-gadget-linux-amd64` or `make kubectl-gadget-darwin-amd64`.
 
 ### Building the gadget container image
 
+Inspektor Gadget provides two different container images:
+- gadget-default: Contains CO-RE and BCC gadgets
+- gadget-core: Containes only CO-RE gadgets that are integrated with the gadget tracer manager
+  * traceloop and the network policy advisor gadgets aren't included,
+    they will be included once
+    [#371](https://github.com/kinvolk/inspektor-gadget/issues/371) and
+    [#172](https://github.com/kinvolk/inspektor-gadget/issues/172) are
+    fixed.
+
 You can build and push the container gadget image by running the following commands:
 
 ```bash
-$ make gadget-container
+$ make gadget-default-container # or make gadget-core-container
 $ make push-gadget-container
 ```
 
@@ -46,13 +55,14 @@ The BPF code is built using a Docker container, so you don't have to worry
 installing the compilers to build it.
 
 ### Building notes
+
 - The compilation uses `tools/image-tag` to choose the tag of the container
 image to use according to the branch that you are compiling.
 - The container repository is set with the `CONTAINER_REPO` env variable.
 - You can push the container images to another registry and use the `--image`
 argument when deploying to the Kuberentes cluster.
 - If you wish to make changes to traceloop program, update
-`gadget.Dockerfile` to pick your own image of traceloop.
+`gadget-default.Dockerfile` to pick your own image of traceloop.
 - As for traceloop, it is also possible to change the BCC to be used as
 described in [BCC](#Updating-BCC-from-upstream) section.
 - See the [minikube](#Development-environment-on-minikube)
@@ -121,6 +131,7 @@ it's highly recommended to open an issue first to discuss it with the team.
 ## BCC
 
 ### Porting BCC gadgets
+
 This project uses some gadgets from [BCC](https://github.com/iovisor/bcc/).
 Instead of keeping our patched versions, we prefer to make those gadgets
 suitable to be used with Inspektor Gadget by contributing to the upstream project.
@@ -143,7 +154,8 @@ The [adding new BCC-based gadgets in Inspektor Gadget](https://kinvolk.io/blog/2
 blogpost presents some more details about this process.
 
 ### Updating BCC from upstream
-As you can see in `gadget.Dockerfile`, the gadget container image
+
+As you can see in `gadget-default.Dockerfile`, the gadget container image
 uses the BCC container image as its parent image.
 Given that there is not an official container repository to get that BCC image,
 we keep a synchronised [Kinvolk BCC fork](https://github.com/kinvolk/bcc)
@@ -158,7 +170,7 @@ it is necessary to first update the
 [Kinvolk BCC fork](https://github.com/kinvolk/bcc)
 so that the Github actions are triggered, and a new image is published.
 Once the image is available in registries, you have to update
-`gadget.Dockerfile` so that it uses the just created image, same goes for local
+`gadget-default.Dockerfile` so that it uses the just created image, same goes for local
 compilation with `gadget-local.Dockerfile`. The
 [Update BCC container image](https://github.com/kinvolk/inspektor-gadget/pull/190)
 PR is an example of it.
--- a/gadget-container/entrypoint.sh
+++ b/gadget-container/entrypoint.sh
@@ -47,7 +47,7 @@ echo $INSPEKTOR_GADGET_VERSION
 # mounted, passing /sys/fs/bpf from the pseudo-host does not work.
 # See also:
 # https://github.com/kubernetes/minikube/blob/99a0c91459f17ad8c83c80fc37a9ded41e34370c/deploy/kicbase/entrypoint#L76-L81
-BPF_MOUNTPOINT_TYPE="`stat -f --format=%T /sys/fs/bpf`"
+BPF_MOUNTPOINT_TYPE="`stat -f -c %T /sys/fs/bpf`"
 if [ "$BPF_MOUNTPOINT_TYPE" != "bpf_fs" ] ; then
   echo "/sys/fs/bpf is of type $BPF_MOUNTPOINT_TYPE. Remounting."
   mount -t bpf bpf /sys/fs/bpf/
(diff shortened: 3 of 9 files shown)


.dockerignore
```diff
@@ -5,3 +5,6 @@
 !gadget-container
 !pkg
 !tools
+!Makefile
+!*.mk
+!.git
```
alban, line 10: Could we use a different dockerignore file for the kubectl-gadget image? Otherwise, the addition of `.git` would invalidate the cache too often for the main image.

How to use a different dockerignore: https://github.com/moby/moby/issues/12886#issuecomment-480575928

gadget.core.Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+COPY --from=builder /gadget/gadget-container/bin/nrigadget /opt/hooks/nri/
+COPY gadget-container/hooks/nri/conf.json /opt/hooks/nri/
+
+## Hooks Ends
+
+# BTF files
+COPY --from=builder /tmp/btfs /btfs/
+
+# Mitigate https://github.com/kubernetes/kubernetes/issues/106962.
+RUN rm -f /var/run
```
alban, line 100: `gadget.Dockerfile` does not have `-f`. I think the 2 Dockerfiles should be consistent.

alban (commented):
> It's still not clear if we want to push these images to a container registry, so this PR doesn't handle it.

What about pushing the images to the same container registry but with the tag prefix 'core-'?

.dockerignore
mauriciovasquezbernal: Updated. I left the rules for `Makefile` and `*.mk` in case BuildKit is not available.

gadget.core.Dockerfile
mauriciovasquezbernal: Updated both of them to have `-f`.

mauriciovasquezbernal:
I'll wait for https://github.com/kinvolk/inspektor-gadget/pull/597 to go in first before updating this PR.

mauriciovasquezbernal:
> I'll wait for #597 to go in first before updating this PR.

Actually I changed my mind. I wanted to tackle https://github.com/kinvolk/inspektor-gadget/issues/625 in this PR too but it seems it's too much, so I propose to merge it as it's.

gadget.core.Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+RUN set -ex; \
+	if command -v tdnf; then \
+		tdnf install -y libseccomp wget curl; \
+	elif command -v yum; then \
+		yum install -y libseccomp wget curl; \
+	elif command -v apt-get; then \
+		apt-get update && \
+		apt-get install -y seccompwget curl ; \
+	elif command -v apk; then \
+		apk add gcompat libseccomp bash wget curl ; \
```
eiffel-fl, line 67: Do we forcefully need `bash` there?
I think adding bash to an alpine image increase the image size and so we loose the benefit of using alpine.

kubectl-gadget.Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG BUILDER_IMAGE=golang:1.18-bullseye
```
eiffel-fl, line 1: What is the use case of this Dockerfile?
To enable having a portable `kubectl-gadget` where user only needs to give the `.kubeconfig` to interact with an existing cluster?

eiffel-fl (commented):
Hi!

Thank you for it, it seems really cool, particularly the image for CO-RE.
I nonetheless have two questions:

Best regards.

gadget.core.Dockerfile
mauriciovasquezbernal: Yes, we need bash there because our entrypoint is a bash script. Anyway, I don't think adding bash will affect the image size that much.

kubectl-gadget.Dockerfile
mauriciovasquezbernal: Yes, it's another way to ship kubectl-gadget, if you already have docker or a similar runtime you can just pull the container image and run kubectl-gadget from there without having to download the binary by hand.

gadget.core.Dockerfile
eiffel-fl: > affect the image size that much.

What is our image size with alpine?
Because `bash` seems to add around 4.5 MB:

```
alpine-bash                             latest                                                   dea7fd1962e5   4 minutes ago   10MB
alpine                                  latest                                                   0ac33e5f5afa   2 weeks ago     5.57MB
```

kubectl-gadget.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Cache go modules so they won't be downloaded at each build
+COPY go.mod go.sum /gadget/
+RUN cd /gadget && go mod download
+
+# This COPY is limited by .dockerignore
+COPY ./ /gadget
+RUN cd /gadget && make kubectl-gadget
+
+FROM ${BASE_IMAGE}
+COPY --from=builder /gadget/kubectl-gadget /kubectl-gadget
```
alban, line 22: Why not `/bin/kubectl-gadget`?

Is there any documentation how to use that image, or how to extract kubectl-gadget from that image?

alban (changes requested):
Some notes:
* Add comments near the new Makefile target and at the top of the 3 Dockerfiles to explain the difference between them (info you've written in the commit messages)
* The image names will need to be renamed after #597 gets merged (if it gets merged first)

kubectl-gadget.Dockerfile
mauriciovasquezbernal: Updated it to be /bin/kubectl-gadget.

mauriciovasquezbernal:
I updated the PR to build the core container image. However it's not pushed nor tested. I think it could be handled in https://github.com/kinvolk/inspektor-gadget/issues/625. Also to avoid more conflicts with https://github.com/kinvolk/inspektor-gadget/pull/597.

docs/CONTRIBUTING.md
```diff
@@ -143,7 +149,8 @@ The [adding new BCC-based gadgets in Inspektor Gadget](https://kinvolk.io/blog/2
 blogpost presents some more details about this process.
 
 ### Updating BCC from upstream
-As you can see in `gadget.Dockerfile`, the gadget container image
+
+As you can see in `gadget.bccDockerfile`, the gadget container image
```
marga-kinvolk, line 153:
```suggestion
As you can see in `gadget.bcc.Dockerfile`, the gadget container image
```

docs/CONTRIBUTING.md
```diff
@@ -35,24 +35,29 @@ kubectl-gadget-linux-amd64` or `make kubectl-gadget-darwin-amd64`.
 
 ### Building the gadget container image
 
+Inspektor Gadget provides two different container images:
+- gadget-bcc: Contains CO-RE and BCC gadgets
+- gadget-core: Containes only CO-RE gadgets
```
marga-kinvolk, line 40: What about non CO-RE gadgets that are not BCC (traceloop, network-policy advisor)?

gadget-container/entrypoint.sh
```diff
@@ -108,7 +108,7 @@ if [ "$ID" = "rhcos" ] && [ ! -d "/usr/src/kernels/$KERNEL" ]; then
       rpm2cpio $RPM | cpio --quiet -i
       popd
 
-      # In the gadget image, /usr/src is a symlink to /host/usr/src (see gadget.Dockerfile).
+      # In the gadget image, /usr/src is a symlink to /host/usr/src (see sDockerfile).
```
marga-kinvolk, line 111:
```suggestion
      # In the gadget image, /usr/src is a symlink to /host/usr/src (see Dockerfile).
```

kubectl-gadget.Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Dockerfile for the kubectl-gadget plugin.
```
marga-kinvolk, line 1: Maybe add one more sentence regarding the purpose of this container? Given that the plugin is a statically compiled binary, it's not immediately obvious why someone might want to have a container with it.

gadget-container/entrypoint.sh
```diff
@@ -59,7 +59,7 @@ if grep -q '^1:name=systemd:.*/crio-[0-9a-f]*\.scope$' /proc/self/cgroup > /dev/
     CRIO=1
 fi
 
-# In the gadget image, /usr/src is a symlink to /host/usr/src (see gadget.Dockerfile).
+# In the gadget image, /usr/src is a symlink to /host/usr/src (see Dockerfile).
```
marga-kinvolk, line 62: Why did you remove the prefix here? There isn't a "Dockerfile" file. Do you mean to point to any of the Dockerfiles? I think maybe gadget.*.Dockerfile would be clearer in that case? Otherwise, if I'm reading this and then find that there is no Dockerfile, I'd assume something got broken in between.

gadget-container/entrypoint.sh
marga-kinvolk: Or maybe gadget.*.Dockerfile, depending on the reply to my comment above.

marga-kinvolk (commented):
Some comments about documentation

kubectl-gadget.Dockerfile
blanquicet, line 1: Why you didn't create a make target to build this container?

kubectl-gadget.Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Since the plugin is a statically compiled golang application, any base
+# image is valid, even scratch. Alpine is used as a tradeoff between
+# size and tools available in the image.
```
blanquicet, line 4: I was expecting it to be used like this:

```bash
$ docker run --rm -it kubectl-gadget trace exec
NODE             NAMESPACE        POD              CONTAINER        PCOMM            PID    PPID   RET ARGS
...
```

In that case we could use scratch as base image.

gadget.core.Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+RUN set -ex; \
+	if [ "$ENABLE_BTFGEN" = true ]; then \
+		cd /btf-tools && \
+		BTFHUB=/tmp/btfhub INSPEKTOR_GADGET=/gadget ./btfgen.sh; \
+	fi
+
+# Main gadget image
+FROM ${BASE_IMAGE}
+
+# install libseccomp according to the package manager available on the base image
```
blanquicet, line 53: It is missing `rpm2cpio` and `cpio` for OpenShift.

blanquicet (commented):
I think it looks like a good starting point. There are some things that need to be improved/missing like the documentation of the kubectl-gadget image but I think we can manage it in future PRs.

docs/CONTRIBUTING.md
mauriciovasquezbernal: Added a comment clarifying what happens with them.

gadget-container/entrypoint.sh
mauriciovasquezbernal: I wanted to make it generic and valid for both Dockerfiles, changed to `gadget.*.Dockerfile`.

gadget.core.Dockerfile
```diff
@@ -0,0 +1,94 @@
+# Dockerfile for Inspektor Gadget CO-RE container image gadgets.
+# This is a reduced gadget container image that supports CO-RE gadgets
+# only, i.e, it doesn't depend on BCC. This image is much smaller than
+# the traditional image and is designed to be used on systems that
+# support BTF (CONFIG_DEBUG_INFO_BTF).
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Makes sense. Reworded it in a similar way to what you suggested.

kubectl-gadget.Dockerfile.dockerignore
```diff
@@ -0,0 +1,10 @@
+*
+!go.mod
+!go.sum
+!cmd
+!gadget-container
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's not needed. I'll remove it.

kubectl-gadget.Dockerfile
mauriciovasquezbernal: That's one option. Another one is to run `docker run --rm -it kubectl-gadget /bin/sh` and then execute kubectl-gadget from there, it provides lower latency as it doesn't need to create a new container instance for each kubectl-gadget invocation. Alpine is around 6MB, so I'd say it's better to provide an image with some more tools than just providing the kubectl-gadget binary.

gadget.core.Dockerfile
mauriciovasquezbernal: These are needed for downloading the kernel headers but this image is about CO-Re tools, hence those should not be needed. What happens if they are not installed?, is there a warning or the whole gadget container creation fails?

kubectl-gadget.Dockerfile
mauriciovasquezbernal: Added.

gadget.core.Dockerfile
blanquicet: You are right. Currently, it prints an error but it goes ahead with the container initialization. Once we have CORE implementation for all the gadgets, we can avoid downloading the kernel headers if `CONFIG_DEBUG_INFO_BTF` is set.

mauriciovasquezbernal:
For the time being the Github Actions are only compiling the image but it's not pushed anywhere. We need to understand how to publish those images, I have thought about two different possibilities:

1. Use the same registry but different tags
    - ghcr.io/kinvolk/inspektor-gadget:v0.5.0 -> points to default image 
    - ghcr.io/kinvolk/inspektor-gadget:v0.5.0-core
    - ghcr.io/kinvolk/inspektor-gadget:v0.5.0-bcc
    - ghcr.io/kinvolk/inspektor-gadget:v0.5.0-kubectl-gadget
2. Have different registries for each image (TODO: we can use / or - as name separator depending on the container registry)
   - ghcr.io/kinvolk/inspektor-gadget/kubectl-gadget:v0.5.0
   - ghcr.io/kinvolk/inspektor-gadget-core:v0.5.0
   - ghcr.io/kinvolk/inspektor-gadget-bcc:v0.5.0

Anyway, I think it's something we can tackle in https://github.com/kinvolk/inspektor-gadget/issues/625 and we can move ahead with this PR. Any comments?

.github/workflows/inspektor-gadget.yml
```diff
@@ -240,6 +240,53 @@ jobs:
...
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Cache Docker layers
+      uses: actions/cache@v2
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-single-buildx-${{ hashFiles('gadget.core.Dockerfile') }}
+        restore-keys: |
+          ${{ runner.os }}-single-buildx
```
alban, line 257:
```suggestion
      uses: actions/cache@v2
      with:
        path: /tmp/.buildx-cache
        key: ${{ runner.os }}-buildx-gadget-${{ hashFiles('gadget.core.Dockerfile') }}
        restore-keys: |
          ${{ runner.os }}-buildx-gadget-
          ${{ runner.os }}-buildx-
```

We could use different cache prefix for the core image build and the non-core image build, since they are not based on the same image. Otherwise, in case of non-exact match on the key, we risk picking up the wrong one, depending on what is faster to build.

Similarly, the non-core image build could use `${{ runner.os }}-buildx-noncore-${{ hashFiles('gadget.Dockerfile') }}`

See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

gadget.bcc.Dockerfile
```diff
@@ -1,12 +1,15 @@
-# Prepare and build gadget artifacts in a container
+# Dockerfile for Inspektor Gadget container image.
+# This image contains CO-RE and BCC-based gadgets. It uses as base the
+# BCC image. It's the default image that is deployed in Inspektor Gadget.
```
alban, line 3:
```suggestion
# Dockerfile for Inspektor Gadget container image.
# This image contains CO-RE and BCC-based gadgets. Its base image is the
# BCC image. It's the default image that is deployed in Inspektor Gadget.
```

docs/CONTRIBUTING.md
```diff
@@ -35,24 +35,34 @@ kubectl-gadget-linux-amd64` or `make kubectl-gadget-darwin-amd64`.
 
 ### Building the gadget container image
 
+Inspektor Gadget provides two different container images:
+- gadget-bcc: Contains CO-RE and BCC gadgets
+- gadget-core: Containes only CO-RE gadgets that are integrated with the gadget tracer manager
```
alban, line 40: Given that `gadget-bcc` contains both bcc-based gadgets and core-based gadgets, it is a bit confused to call it bcc only. What about calling the two files:
1. `gadget-default`
2. `gadget-coreonly`

(and renaming the Dockerfiles accordingly)

alban (commented):
Thanks!

eiffel-fl:
Sorry I did not review it, but from my understanding it looks quite cool.

> We need to understand how to publish those images, I have thought about two different possibilities:

Regarding this, I would prefer the second except I think GHCR does not handle nested registries (`inspektor-gadget/kubectl-gadget`).
Anyway, this PR is about "**[c]reat**[ing] dockerfiles for kubectl-gadget and CO-RE tools", so I think it can be merged and we will address GHCR troubles in future(s?) PR.

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: I'm no t sure I follow. What you're saying is that `restore-keys` should be different for both images to avoid picking up a wrong cache when there is not a match on `key`? 

Will using 

```        
restore-keys: |
          ${{ runner.os }}-core-single-buildx
```

and 

```        
restore-keys: |
          ${{ runner.os }}-bcc-single-buildx
```

fix the issue?

Also, do we really need these `restore-keys`?

kubectl-gadget.Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Dockerfile for the kubectl-gadget plugin.
+# This image is useful for cases where downloading and running a binary
+# is not possible or is complicated. In this cases, this container image
+# provides an alternative to download and run kubectl-gadget.
+# Since the plugin is a statically compiled golang application, any base
+# image is valid, even scratch. Alpine is used by default as a tradeoff
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I just tested and it worked fine. It seems the lack of `etc/nsswitch.conf` doesn't affect it. 

```
sudo docker run --rm -it -v /home/vagrant/.kube/config:/.kube/config kubectl-gadget:latest kubectl-gadget trace exec
```

docs/CONTRIBUTING.md
mauriciovasquezbernal: Renamed it to be `gadget-default` and `gadget-core`.

kubectl-gadget.Dockerfile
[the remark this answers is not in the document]
alban: Thanks for testing!

.github/workflows/inspektor-gadget.yml
alban: > What you're saying is that restore-keys should be different for both images to avoid picking up a wrong cache when there is not a match on key?

Yes.

If a PR modifies the Dockerfile in some layer, the `${{ hashFiles(...) }}` will be different, but it is still worth falling back to an older cache from some older PR, but it has to be the same image, otherwise there will not be any Docker layer in common.

I think your suggested `restore-keys` should work.

> Also, do we really need these restore-keys?

As far as I understand, the `restore-keys` is just a fallback if the preferred `key` does not match. Having them will improve the perf if there is no exact match.

mauriciovasquezbernal: I checked the examples at https://github.com/actions/cache/blob/main/examples.md and it seems the restore keys are always used like the main key without `${{ hashFiles('...') }}`. I updated this to follow the same approach.

mauriciovasquezbernal:
Is there anything else I should make to get this in? @marga-kinvolk  @alban @blanquicet @eiffel-fl

eiffel-fl (approved):
I did not test them but I think it has enough reviews and I trust you so let's merge it.
In case of trouble, we can still address them in future PR.

Just be sure you addressed the requested changes before hitting the merge button.

docs/CONTRIBUTING.md
```diff
@@ -35,24 +35,34 @@ kubectl-gadget-linux-amd64` or `make kubectl-gadget-darwin-amd64`.
...
 ### Building the gadget container image
 
+Inspektor Gadget provides two different container images:
+- gadget-default: Contains CO-RE and BCC gadgets
+- gadget-core: Containes only CO-RE gadgets that are integrated with the gadget tracer manager
+  * traceloop and the network policy advisor gadgets aren't included,
+    they will be included
+    [#371](https://github.com/kinvolk/inspektor-gadget/issues/371) and
+    [#172](https://github.com/kinvolk/inspektor-gadget/issues/172) are
+    fixed.
```
alban, line 45:
```suggestion
  * traceloop and the network policy advisor gadgets aren't included,
    they will be included once
    [#371](https://github.com/kinvolk/inspektor-gadget/issues/371) and
    [#172](https://github.com/kinvolk/inspektor-gadget/issues/172) are
    fixed.
```

alban (approved):
1 suggestion & LGTM. Thanks!

PR #613, opened 2022-04-14
Remove local-gadget from the default build target

marga-kinvolk:
The local-gadget target in our Makefile requires having CLang and libbpf-dev installed locally, which are not considered required dependencies for building Inspektor Gadget. The default build target should only include targets that can be built with the listed tools.

This change removes local-gadget from "build", and creates an "all" target that includes local-gadget

## How to use / Testing done

Run `make` and see how it doesn't call CLang outside of docker.
Run `make all` and see how it does (or fails to, if you machine doesn't have CLang and/or libbpf-dev).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,10 @@ LDFLAGS := "-X main.version=$(VERSION) \
 
 .DEFAULT_GOAL := build
 .PHONY: build
-build: manifests generate local-gadget kubectl-gadget gadget-container
+build: manifests generate kubectl-gadget gadget-container
+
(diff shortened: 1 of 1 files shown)


blanquicet:
CI is failing because of the integration tests but the provided output is not enough to determine what was the cause. I had already filed an issue to improve the output (https://github.com/kinvolk/inspektor-gadget/issues/567). For now, please re-run the failed job.

blanquicet (approved):
LGTM!

PR #617, opened 2022-04-20
cmd/undeploy: Remove old Inspektor Gadgets components too

mauriciovasquezbernal:
It could happen that a user deploys Inspektor Gadget v0.5.0 without
first removing a previous version. This condition creates a lot of
problems because there will be two gadget daemons running (kube-system
and gadget namespaces), hence two traces controllers that will race
each other creating failures some times.

If the user has a kubectl-gadget binary >= v0.5.0 they won't be able
to remove the old Inspektor Gadget from the cluster. This commit extends
the current undeploy command to remove also components from old
versions.

It's a real problem and already hit one of our users -> https://kubernetes.slack.com/archives/CSYL75LF6/p1647025053942309?thread_ts=1644399681.035839&cid=CSYL75LF6.

### DIFF
--- a/cmd/kubectl-gadget/undeploy.go
+++ b/cmd/kubectl-gadget/undeploy.go
@@ -164,6 +164,26 @@ again:
 		)
 	}
 
+	// Let's try to remove components of IG versions before v0.5.0,
+	// just in case somebody has a newer CLI but is trying to remove
+	// an old version of Inspektor Gadget from the cluster. Given
+	// that this is a best effort work, we don't track any error.
+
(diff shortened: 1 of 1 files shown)


eiffel-fl (approved):
Tested and validated!
Thank you for it!

eiffel-fl:
Since it is best effort, isn't it possible to do it using `defer`?

mauriciovasquezbernal:
That function does not return on other errors (they're only collected in the errs slice), so the I don't see how `defer` could have an impact on the behavior.

PR #618, opened 2022-04-20
pkg/gadget/-*collector: Mark trace as completed when no container matches.

eiffel-fl:
There was a problem in the CLI where the collect operation would timeout.
Particularly, if there was, e.g., 3 nodes but target namespace only has one pod
on a given node.
Then the traces created for the nodes without trace would make the CLI code
timeout because they would never reach Completed state.

### DIFF
--- a/pkg/gadgets/process-collector/gadget.go
+++ b/pkg/gadgets/process-collector/gadget.go
@@ -73,6 +73,7 @@ func (t *Trace) Collect(trace *gadgetv1alpha1.Trace) {
 
 	if len(events) == 0 {
 		trace.Status.OperationWarning = "No container matches the requested filter"
+		trace.Status.State = "Completed"
 		return
 	}
 
--- a/pkg/gadgets/socket-collector/gadget.go
+++ b/pkg/gadgets/socket-collector/gadget.go
@@ -77,6 +77,7 @@ func (t *Trace) Collect(trace *gadgetv1alpha1.Trace) {
 	filteredContainers := t.resolver.GetContainersBySelector(selector)
 	if len(filteredContainers) == 0 {
 		trace.Status.OperationWarning = "No container matches the requested filter"
+		trace.Status.State = "Completed"
 		return
 	}
 


eiffel-fl:
Error messages will indeed be printed but the trace output as well:

```
$ kubectl get node  
NAME           STATUS   ROLES                  AGE   VERSION
minikube       Ready    control-plane,master   71m   v1.22.2
minikube-m02   Ready    <none>                 71m   v1.22.2
minikube-m03   Ready    <none>                 70m   v1.22.2
$ kubectl get pod -o=wide
NAME        READY   STATUS    RESTARTS   AGE   IP           NODE           NOMINATED NODE   READINESS GATES
nginx-app   1/1     Running   0          67m   10.244.1.3   minikube-m02   <none>           <none>
$ ./kubectl-gadget snapshot socket
Error: failed to run gadget on node "minikube": No container matches the requested filter
Error: failed to run gadget on node "minikube-m03": No container matches the requested filter
NODE            NAMESPACE    POD          PROTOCOL    LOCAL         REMOTE       STATUS
minikube-m02    default      nginx-app    TCP         0.0.0.0:80    0.0.0.0:0    LISTEN
```

mauriciovasquezbernal:
> Notice that if you have a cluster with 10 nodes, you will have 9 lines of error

Yes, it messes the output too much when there are not real errors.

mauriciovasquezbernal:
After some considerations I think we should only make sure the gadget changes its status to `Completed` even when the filter doesn't match:

```diff
diff --git pkg/gadgets/process-collector/gadget.go pkg/gadgets/process-collector/gadget.go
index 78c5f18d..2803be95 100644
--- pkg/gadgets/process-collector/gadget.go
+++ pkg/gadgets/process-collector/gadget.go
@@ -73,6 +73,7 @@ func (t *Trace) Collect(trace *gadgetv1alpha1.Trace) {
 
        if len(events) == 0 {
                trace.Status.OperationWarning = "No container matches the requested filter"
+               trace.Status.State = "Completed"
                return
        }
 
diff --git pkg/gadgets/socket-collector/gadget.go pkg/gadgets/socket-collector/gadget.go
index d919bc17..67cf6c11 100644
--- pkg/gadgets/socket-collector/gadget.go
+++ pkg/gadgets/socket-collector/gadget.go
@@ -77,6 +77,7 @@ func (t *Trace) Collect(trace *gadgetv1alpha1.Trace) {
        filteredContainers := t.resolver.GetContainersBySelector(selector)
        if len(filteredContainers) == 0 {
                trace.Status.OperationWarning = "No container matches the requested filter"
+               trace.Status.State = "Completed"
                return
        }
```

With that change the trace will move to the desired status, and not timeout will happen. A drawback of this solution is that no warning will be printed when the filter doesn't match any pod, an empty list will be printed instead:

```
$ ./kubectl-gadget snapshot socket -p nonexistingpod
NODE    NAMESPACE    POD    PROTOCOL    LOCAL    REMOTE    STATUS
```

It's not a perfect solution but I think it's the way to go, by having this change we're sure we don't affect other gadgets, and removing that warning is not a big deal IMO.  What do you think @eiffel-fl , @blanquicet ?

blanquicet:
I agree. This is the simplest/quickest solution to unblock the release. We can think of a proper solution directly when working on the GlobalTrace task.

eiffel-fl:
I am not really happy with either solution.
Considering "Completed" a trace which is not is not a perfect solution.
On the other side, printing error messages is not a good thing and dealing particularly with this error to know if there are effectively an error is just no way (_e.g._ checking if a pod existed on this node but is now dead and so this is a "true" error while if there was never a pod on this node this is a "false" one).

eiffel-fl:
I change the code to use Maurio suggestion.
I do not think this is this gadget responsibility to report crashed pods, so better to not print error messages there.

pkg/gadgets/process-collector/gadget.go
```diff
@@ -73,6 +73,10 @@ func (t *Trace) Collect(trace *gadgetv1alpha1.Trace) {
 
 	if len(events) == 0 {
 		trace.Status.OperationWarning = "No container matches the requested filter"
+		// Mark the trace as complete there to avoid a timeout in CLI.
+		// Indeed, since no container matches the filter, the trace would never hit
+		// Completed state set below.
```
blanquicet, line 78: IMO, it is not a workaround to avoid timeout in CLI but a bug fixing. The trace indeed completed its job, it was requested to snapshot the current sockets using a given filter but given that there is no container that matched the requested filter, it produced an empty output thus its job is done. The warning message is just additional information the trace is giving to the user to explain why the output is empty.

mauriciovasquezbernal: I think that comment is misleading. The trace is in fact completed, it completed the tasks without producing any output because there was not any matching pod.

mauriciovasquezbernal:
The PR title and description should be updated to reflect the actual changes we're doing.

mauriciovasquezbernal (approved):

mauriciovasquezbernal:
> I change the code to use Maurio suggestion. 

That's a new one to add to the list: [Mauricio, Mauro, Mauri, **Maurio**] :D

mauriciovasquezbernal:
Merging to unblock v0.5.0 release.

PR #621, opened 2022-04-20
cmd: Align top commands

mauriciovasquezbernal:
These is most a renaming commit to align the different top commands:
- move common variables to top.go file
- add prefix to specific gadget variables

There is still a lot of room for improvement. We could create a generic
top command to reduce code duplication but it's TBD in a different
commit.


cmd/kubectl-gadget/top/top.go
```diff
@@ -15,16 +15,35 @@
...
+	"fmt"
+	"strings"
+	"sync"
+
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
 
 	"github.com/spf13/cobra"
 )
 
 // All the gadgets within this package use this global variable, so let's
```
eiffel-fl, line 27: `global variableS`.
(Or you can even remove the comment.)

eiffel-fl (approved):
Thank you, I like the "globalization" of the resources this PR added.
I tested it and the 3 gadgets are OK.
There is just one small nit, but you can even merge it like this.

> We could create a generic
top command to reduce code duplication but it's TBD in a different
commit.

I agree on this point, I need to think more about the API we can provide and if it is possible to give one for, at least, each category.

cmd/kubectl-gadget/top/top.go
mauriciovasquezbernal: I removed the comment.

Issue #620, opened 2022-04-20
[RFE] Do not create the `containers` eBPF map for local-gadget


blanquicet:
## Current situation

The `containers` eBPF map is created at `/sys/fs/bpf/gadget/containers` by the package `containersmap` to expose the container details for each mount namespace.

This map is used by the Python-based gadgets to enrich the output with the Kubernetes metadata of the containers.

## Impact

As described in https://github.com/kinvolk/inspektor-gadget/issues/619, it is a problem for systems where `/sys/fs/bpf` is not `bpf`.

## Ideal future situation

Given that the `local-gadget` does not use the Python-based gadgets thus it does not use the `containers` map, we should not create it for `local-gadget`. It will remove also the need of having a directory of type `bpf` and all the issues it involves.

alban:
Note that the audit-seccomp gadget uses the containers map too.

mauriciovasquezbernal:
> Note that the audit-seccomp gadget uses the containers map too.

Right. I think the point here is not to pin the map, but it can still be created and passed as an *ebpf.Map to the gadgets that need it.

alban:
Fixed by https://github.com/kinvolk/inspektor-gadget/pull/654

Issue #619, opened 2022-04-20
[RFE] Do not pin the filtering eBPF map


blanquicet:
## Current situation

Most gadgets can generate events for a specific container or for all the containers within a given pod, namespace or labels. To do so, the gadgets use an eBPF map with the mount namespace of the containers they are interested in. In particular, this map is created by the userspace code with the list of mount namespace (i.e. containers) requested by the user and it is used by the kernel space code to propagate only the events from the requested containers.
 
Currently, the userspace code has to pin this eBPF map and pass its path as a string to the kernel space code instead of passing the map itself, as it should be. Example:

https://github.com/kinvolk/inspektor-gadget/blob/e2a88779e8ce5212f6c2230a8e3cfbcd2f6b5ea1/pkg/gadgets/bindsnoop/tracer/interface.go#L22-L25

As mentioned in the comment, it had to be done in this way because of two issues found in the cilium/ebpf library:

- [ ] https://github.com/cilium/ebpf/issues/515
- [ ] https://github.com/cilium/ebpf/issues/517

## Impact

To be able to pin eBPF maps, it is necessary to do it in a directory of type `bpf`. Inspektor Gadget uses the directory `/sys/fs/bpf` because, [from v238](https://github.com/systemd/systemd/commit/43b7f24b5e0dd048452112bfb344739764c58694), systemd automatically mounts it as `bpf`. However, we need to apply a workaround for systems where such directory is not `bpf` due to they use an older version of systemd (for instance, AKS currently uses v237) or special systems like minikube (see [here](https://github.com/kinvolk/inspektor-gadget/blob/b50d93f8812a0ce049c1289b50aa060f228efd7b/gadget-container/entrypoint.sh#L45-L47)).

## Ideal future situation

eBPF maps are passed as maps and they are not pinned.

## Implementation options

Notice that issue 515 was already solved upstream and maybe by following the suggestion maintainers did on it we could avoid issue 517. 

## Important information

Solving this issue is very important for `local-gadget` because it runs in the node itself and the workaround we use for Inspektor Gadget (i.e. within the `gadget` pod, see [here](https://github.com/kinvolk/inspektor-gadget/blob/b50d93f8812a0ce049c1289b50aa060f228efd7b/gadget-container/entrypoint.sh#L45-L54)), are too invasive and should be avoided directly in the node.

mauriciovasquezbernal:
[...]

PR #622, opened 2022-04-21
local-gadget: Ensure /sys/fs/bpf is of type bpf

blanquicet:
# local-gadget: Ensure /sys/fs/bpf is of type bpf

Currently, the `local-gadget` creates pinning eBPF maps for filtering events and to expose the container details for each mount namespace. However, this is something that is not actually necessary for the `local-gadget` as it does not support the use of python-based gadgets (so the `containers` map is not needed, see issue #620) and it could pass directly the filtering map to the eBPF code instead of passing the path of the pinned map (once cilium/ebpf issues are fixed, see issue #619).

This PR adds a temporary workaround to be able to run `local-gadget` on systems where `/sys/fs/bpf` is not of type `bpf`. For instance, on AKS nodes. This workaround can be fixed once #619 and #620 are done. This is the same workaround we apply when running Inspektor Gadget but inside the `gadget` pod, see [here](https://github.com/kinvolk/inspektor-gadget/blob/ec20f3c5772901f319ff484a31a88d589c2f62a6/gadget-container/entrypoint.sh#L45-L54).

I took advantage of this PR to also add the `--verbose / -v` flag to the `local-gadget`.

## How to use

Try to run `local-gadget` on a system where `/sys/fs/bpf` is not of type `bpf`.

## Testing done

Check the type of `/sys/fs/bpf`, in this case it is not `bpf`:
```bash
$ stat -f --format=%T /sys/fs/bpf
sysfs
```

Run `local-gadget`:
```bash
$ local-gadget
INFO[0000] /sys/fs/bpf is of type sysfs. Remounting with type bpf
» list-containers
k8s_shell01_net_default_25be0e27-bc3d-4c38-b0a7-518b751ba898_9
...
```

Check what is the type of `/sys/fs/bpf` now:
```bash
$ stat -f --format=%T /sys/fs/bpf
bpf_fs
```

If we run it again in the same system or any other system where  `/sys/fs/bpf` is type `bpf`, it does nothing, not even print a message:
```bash
$ local-gadget
» list-containers
k8s_shell01_net_default_25be0e27-bc3d-4c38-b0a7-518b751ba898_9
...
```

Check the `--verbose / -v` flag:

```bash
$ local-gadget -v
DEBU[0000] Checking "/usr/bin/runc": done
DEBU[0000] Checking "/usr/sbin/runc": fanotify: mark error, no such file or directory
DEBU[0000] Checking "/usr/local/sbin/runc": fanotify: mark error, no such file or directory
DEBU[0000] Checking "/run/torcx/unpack/docker/bin/runc": fanotify: mark error, no such file or directory
DEBU[0000] Checking "/host/usr/bin/runc": fanotify: mark error, no such file or directory
DEBU[0000] Checking "/host/usr/sbin/runc": fanotify: mark error, no such file or directory
DEBU[0000] Checking "/host/usr/local/sbin/runc": fanotify: mark error, no such file or directory
DEBU[0000] Checking "/host/run/torcx/unpack/docker/bin/runc": fanotify: mark error, no such file or directory
» list-gadgets
audit-seccomp
dns
seccomp
snisnoop
socket-collector
```


pkg/local-gadget-manager/local-gadget-manager.go
```diff
@@ -302,6 +304,25 @@ func NewManager() (*LocalGadgetManager, error) {
 		return nil, err
 	}
 
+	// Ensure /sys/fs/bpf is of type bpf. It is necessary to be able to pin eBPF
+	// maps. TODO: Remove the need of using pinning, see issues #619 and #620.
+	cmd := exec.Command("/bin/sh", "-c", "stat -f --format=%T /sys/fs/bpf")
```
alban, line 309: Could you use https://pkg.go.dev/golang.org/x/sys/unix#Statfs?

BPF_FS_MAGIC = 0xcafe4a11 according to manual page statfs.2.

pkg/local-gadget-manager/local-gadget-manager.go
```diff
@@ -302,6 +304,25 @@ func NewManager() (*LocalGadgetManager, error) {
...
+	output, err := cmd.Output()
+	if err != nil {
+		return nil, fmt.Errorf("error checking type of /sys/fs/bpf (%s): %w",
+			cmd.String(), err)
+	}
+	fsType := strings.TrimSpace(string(output))
+	if fsType != "bpf_fs" {
+		log.Infof("/sys/fs/bpf is of type %s. Remounting with type bpf", fsType)
+
+		cmd := exec.Command("/bin/sh", "-c", "mount -t bpf bpf /sys/fs/bpf/")
```
alban, line 319: Could use https://pkg.go.dev/golang.org/x/sys/unix#Mount

[Example of usage](https://github.com/opencontainers/runc/blob/238717ddd31cff54b36a6ae937c4016d501caaa8/libcontainer/mount_linux.go#L57)

pkg/local-gadget-manager/local-gadget-manager.go
blanquicet: Yeah, I should have expected that they exist, sorry.

I update the PR to use `unix.Mount` and `unix.Statfs`.

pkg/local-gadget-manager/local-gadget-manager.go
```diff
@@ -302,6 +304,24 @@ func NewManager() (*LocalGadgetManager, error) {
...
+		log.Debugf("%s is of type %#x but we were expecting %#x (BPF_FS_MAGIC)",
+			bpfPinPath, buf.Type, unix.BPF_FS_MAGIC)
+
+		log.Infof("Remounting %q as bpf", bpfPinPath)
+
+		if err := unix.Mount("bpf", bpfPinPath, "bpf", 0, ""); err != nil {
+			return nil, fmt.Errorf("error remounting %s as bpf: %w", bpfPinPath, err)
+		}
+	}
+
```
alban, line 324: Could this be moved in a separate function, even if it is in the same file for now?
```
func ensureBPFMount() error
```

cmd/local-gadget/main.go
```diff
@@ -273,124 +274,151 @@ func newRootCmd() *cobra.Command {
...
+			}
 
-	homedir, err := os.UserHomeDir()
-	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
-	}
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
```
alban, line 290: Since this function is quite long, could it be moved outside of main() and be referenced here?

alban (commented):
Thanks! I am adding some suggestions but this should be good after that.

cmd/local-gadget/main.go
blanquicet: You are rigth. The new version looks much better!

pkg/local-gadget-manager/local-gadget-manager.go
```diff
@@ -286,6 +288,31 @@ func (l *LocalGadgetManager) Dump() string {
...
+	if buf.Type != unix.BPF_FS_MAGIC {
+		log.Debugf("%s is of type %#x but we were expecting %#x (BPF_FS_MAGIC)",
+			bpfPinPath, buf.Type, unix.BPF_FS_MAGIC)
+
+		log.Infof("Remounting %q as bpf", bpfPinPath)
+
+		if err := unix.Mount("bpf", bpfPinPath, "bpf", 0, ""); err != nil {
+			return fmt.Errorf("error remounting %s as bpf: %w", bpfPinPath, err)
+		}
+	}
```
mauriciovasquezbernal, line 311: nit: usually doing an early return simplifies this kind of code:

```

if buf.Type == unix.BPF_FS_MAGIC {
  // All good, nothing to do! 
  return nil
}

// logic to remount it. 
```

pkg/gadgets/tcptop/tracer/tcptop_bpfel.go
```diff
@@ -62,7 +62,6 @@ type tcptopProgramSpecs struct {
 //
 // It can be passed ebpf.CollectionSpec.Assign.
 type tcptopMapSpecs struct {
-	CgroupMap  *ebpf.MapSpec `ebpf:"cgroup_map"`
```
mauriciovasquezbernal, line 65: Just out of curiosity: What happened here? An old version of this file was committed? :thinking:

mauriciovasquezbernal (approved):
I haven't tested it myself but LGTM from a code review.

Just two comments that could be ignored.

pkg/local-gadget-manager/local-gadget-manager.go
blanquicet: I wasn't using this format because that code was inside `NewManager()` but now that we moved it out as an independent function, I agree it should follow the early return approach. Thanks!

pkg/gadgets/tcptop/tracer/tcptop_bpfel.go
blanquicet: I think it is because it wasn't updated after cgroup map was removed.

mauriciovasquezbernal (approved):

PR #623, opened 2022-04-21
Retitle usage guides in the docs (plus some minor updates)

marga-kinvolk:
# Retitle usage guides

Some of our usage guides were still using the old CLI names, and even those that weren't were using a weird title format (The "audit seccomp" gadget).  This change updates all titles to the new names, and changes the title template, e.g. Using `audit seccomp`

Also, it moves the "Gadget specs" title to "CR Usage Reference", as those specs are not about how to use the gadget, but rather what the custom resources take as parameters and so on.

There are also some other minor fixes that I ran across when reviewing the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Inspektor Gadget is a collection of tools (or gadgets) to debug and inspect Kube
 
 Inspektor Gadget tools are known as gadgets. You can deploy one, two or many gadgets.
 
-Exploring the following documentation will best help you learn which tools can help you in your investigations.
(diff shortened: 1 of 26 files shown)


eiffel-fl (approved):
I like the "Using `<category> <name>`" naming.
I did not rename them because I did not find a good way and was convinced by the "The `<category> <name>` gadget" as a title.

mauriciovasquezbernal:
I fixed the typo and updated the branch for you @marga-kinvolk to speed up a bit.

PR #626, opened 2022-04-22
integration: Make SPO deployment optional

mauriciovasquezbernal:
The SPO is required for TestAuditSeccomp(), make it optional in case
somebody wants to run tests that don't require this component.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -29,7 +29,8 @@ var (
 	// image such as docker.io/kinvolk/gadget:latest
 	image = flag.String("image", "", "gadget container image")
 
-	doNotDeploy = flag.Bool("no-deploy", false, "don't deploy Inspektor Gadget")
+	doNotDeployIG  = flag.Bool("no-deploy-ig", false, "don't deploy Inspektor Gadget")
+	doNotDeploySPO = flag.Bool("no-deploy-spo", false, "don't deploy the Security Profiles Operator (SPO)")
 )
 
 func runCommands(cmds []*command, t *testing.T) {
@@ -85,30 +86,35 @@ func testMain(m *testing.M) int {
 	rand.Seed(seed)
 	fmt.Printf("using random seed: %d\n", seed)
 
-	initCommands := []*command{
-		deployInspektorGadget,
-		deploySPO,
-		waitUntilInspektorGadgetPodsDeployed,
-		waitUntilInspektorGadgetPodsInitialized,
+	initCommands := []*command{}
+
+	if !*doNotDeploySPO {
+		initCommands = append(initCommands, deploySPO)
+		defer func() {
+			fmt.Printf("Clean SPO:\n")
+			cleanupSPO.runWithoutTest()
+		}()
 	}
 
-	if !*doNotDeploy {
+	if !*doNotDeployIG {
+		initCommands = append(initCommands, deployInspektorGadget)
+		initCommands = append(initCommands, waitUntilInspektorGadgetPodsDeployed)
+		initCommands = append(initCommands, waitUntilInspektorGadgetPodsInitialized)
+
 		// defer the cleanup to be sure it's called if the test
 		// fails (hence calling runtime.Goexit())
 		defer func() {
 			fmt.Printf("Clean inspektor-gadget:\n")
 			cleanupInspektorGadget.runWithoutTest()
-			fmt.Printf("Clean SPO:\n")
-			cleanupSPO.runWithoutTest()
 		}()
+	}
 
-		fmt.Printf("Setup inspektor-gadget:\n")
-		for _, cmd := range initCommands {
-			err := cmd.runWithoutTest()
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "%v\n", err)
-				return -1
-			}
+	fmt.Printf("Running init commands:\n")
+	for _, cmd := range initCommands {
+		err := cmd.runWithoutTest()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			return -1
 		}
 	}
 
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -64,19 +64,17 @@ func runCommands(cmds []*command, t *testing.T) {
 	}
 }
 
-func TestMain(m *testing.M) {
+func testMain(m *testing.M) int {
 	flag.Parse()
 
 	if !*integration {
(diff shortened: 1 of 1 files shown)


eiffel-fl (commented):
It seems cool but I have one concern:

integration/integration_test.go
```diff
@@ -128,6 +135,10 @@ func TestMain(m *testing.M) {
 }
 
 func TestAuditSeccomp(t *testing.T) {
+	if *deploySPOFlag == false {
+		t.Fatal("The Security Profiles Operator was not deployed. Please check that the deploy-spo flag is true")
+	}
+
```
blanquicet, line 141: I would prefer to let it fail if the SPO is not already installed in the system, don't you think?

integration/integration_test.go
```diff
@@ -30,6 +30,8 @@ var (
 	image = flag.String("image", "", "gadget container image")
 
 	doNotDeploy = flag.Bool("no-deploy", false, "don't deploy Inspektor Gadget")
+
+	deploySPOFlag = flag.Bool("deploy-spo", true, "deploy the Security Profiles Operator (SPO)")
```
blanquicet, line 34: What about following the same approach for both and making it `no-deploy-spo`?

integration/integration_test.go
```diff
@@ -89,16 +91,21 @@ func TestMain(m *testing.M) {
...
 		deployInspektorGadget,
-		deploySPO,
 		waitUntilInspektorGadgetPodsDeployed,
 		waitUntilInspektorGadgetPodsInitialized,
 	}
 
+	if *deploySPOFlag {
+		initCommands = append(initCommands, deploySPO)
+	}
+
```
blanquicet, line 101: What about decoupling `deploySPOFlag` from `doNotDeploy`? With this PR, the user cannot only deploy the SPO without deploying IG. I would suggest making them independent.

integration/integration_test.go
```diff
@@ -30,6 +30,8 @@ var (
 	image = flag.String("image", "", "gadget container image")
 
 	doNotDeploy = flag.Bool("no-deploy", false, "don't deploy Inspektor Gadget")
```
blanquicet, line 32: nit: Maybe changing `doNotDeploy` to `doNotDeployIG` would make the code clearer now that we also have `deploySPOFlag`?

blanquicet (commented):
I think this PR will make our lives easier! 🙂 ... Some comments but looks good overall.

integration/integration_test.go
mauriciovasquezbernal: Yeah, I think it makes sense to let it fail.

integration/integration_test.go
mauriciovasquezbernal: Right, updated to follow this approach.

integration/integration_test.go
mauriciovasquezbernal: Makes sense, it was a bit complicated but I think it's fine now.

mauriciovasquezbernal:
I updated the code following the suggestions. Please review with care, specially the "integration: Simplify logic around deferred calls" commit as it touches parts that have causes problems to us in the past.

integration/integration_test.go
blanquicet: Yeah, it is clearer now. Thanks for managing this!

integration/integration_test.go
```diff
@@ -64,19 +64,17 @@ func runCommands(cmds []*command, t *testing.T) {
 	}
```
eiffel-fl, line 65: Commit message:
`s/(guarante)/$1e/`
`s/(functions )all( called)/$1are$2/`

integration/integration_test.go
```diff
@@ -29,7 +29,8 @@ var (
 	// image such as docker.io/kinvolk/gadget:latest
 	image = flag.String("image", "", "gadget container image")
 
-	doNotDeploy = flag.Bool("no-deploy", false, "don't deploy Inspektor Gadget")
+	doNotDeployIG  = flag.Bool("no-deploy-ig", false, "don't deploy Inspektor Gadget")
+	doNotDeploySPO = flag.Bool("no-deploy-spo", false, "don't deploy the Security Profiles Operator (SPO)")
```
eiffel-fl, line 33: Instead of `*DeploySPO`, can't we call this `*InstallSPO`?
So, we are not bothered regarding IG.

eiffel-fl (commented):
I do not understand why we do not make `TestAuditSeccomp` conditionnal regarding the value of `doNotDeploySPO`.
It is a bit strange to know the test will fail but still run it if its requirements are not met...

blanquicet:
> I do not understand why we do not make `TestAuditSeccomp` conditional regarding the value of `doNotDeploySPO`.
> It is a bit strange to know the test will fail but still run it if its requirements are not met...

I suggested this change because the fact that `-no-deploy-spo` is set does not mean the SPO is not installed in the cluster. Indeed, that's how contributors can make some quick local tests during the development. They install the SPO manually and then just execute `sudo -E go test -run TestAuditSeccomp -tags withebpf -test.v -integration -no-deploy-spo ./...` after each change to the `audit-seccomp` gadget. It will be indeed quicker given that installing/unsintalling the SPO takes a while.

eiffel-fl:
> > I do not understand why we do not make `TestAuditSeccomp` conditional regarding the value of `doNotDeploySPO`.
> > It is a bit strange to know the test will fail but still run it if its requirements are not met...
> 
> I suggested this change because the fact that `-no-deploy-spo` is set does not mean the SPO is not installed in the cluster. Indeed, that's how contributors can make some quick local tests during the development. They install the SPO manually and then just execute `sudo -E go test -run TestAuditSeccomp -tags withebpf -test.v -integration -no-deploy-spo ./...` after each change to the `audit-seccomp` gadget. It will be indeed quicker given that installing/unsintalling the SPO takes a while.

I understand it is possible to install the SPO by yourself.
But I think we should add a check or at least an informative error message like "perhaps this is failing because you did not install the SPO?".

integration/integration_test.go
mauriciovasquezbernal: Then why not "InstallIG"? I think both of them should have the prefix.

mauriciovasquezbernal:
> But I think we should add a check or at least an informative error message like "perhaps this is failing because you did not install the SPO?".

It's deployed by default so I don't think it's a big problem. As Jose said I expected that somebody playing with the `no-deploy-spo` know what they are doing.

PR #627, opened 2022-04-25
Remove formatting from titles

marga-kinvolk:
The documentation rendered that we're using doesn't support formatting in the titles of our docs. :(


mauriciovasquezbernal (approved):
LGTM. 

For the record, it was how it was shown in our docs site:
Selection_089

PR #633, opened 2022-04-27
integration: Make tests faster and improve code 

mauriciovasquezbernal:
- use grace period to speed up namespace deletion (brings ~5m of integration tests improvement) 
- reorganize busyboxcommand function to remove duplicated code


integration/command.go
```diff
@@ -312,6 +318,11 @@ spec:
     - %s
 EOF
 `, namespace, cmd)
+
+	return &command{
+		name: "Run test-pod",
```
eiffel-fl, line 323: What about making this a parameter and so keeping the old names as argument?

integration/command.go
```diff
@@ -312,6 +318,11 @@ spec:
...
+		cmd:  cmdStr,
+	}
```
eiffel-fl, line 325: Almost all the tests checked for `expectedString: "pod/test-pod created\n"`, is it possible to get it back?

eiffel-fl (commented):
Thank you for it, this is really cool.
I am just a bit disappointed to know than 66% of percent of each test was used to wait for the pod to terminate...

I have some remarks on the second commits.
Nonetheless, the idea is good, as it was strange than this function does not return a command.

integration/command.go
mauriciovasquezbernal: I checked the old names and most of them are generic descriptions that IMO don't add any useful information. By the way, we're not using command.name anywhere...

integration/command.go
mauriciovasquezbernal: I added it back. However I'm not 100% sure it's needed. Having a return value == 0 from the invocation of kubectl should be enough, shouldn't it?

integration/command.go
eiffel-fl: Isn't it printed when running with `-test.v`?
If not, then your modification is fine.

integration/command.go
eiffel-fl: Indeed:
https://github.com/kinvolk/inspektor-gadget/blob/67fce08e2f7f7fceb6aebdb0499176e071aee8af/integration/command.go#L209

integration/command.go
mauriciovasquezbernal: It's not printed anywhere. We're just assigning it but not printing.

eiffel-fl: OK, so better to remove it then.

eiffel-fl (approved):
LGTM!

PR #634, opened 2022-04-27
Fix bad json event and provide more information when unmarshal fails

mauriciovasquezbernal:
- Fix bad json sent to client in gadget tracer manager
- Add extra information to WrapInErrUnmarshalOutput() on the client


blanquicet (approved):
LGTM from a code inspection but I didn't test it. Did you find a way to reproduce the unmarshall issue?

mauriciovasquezbernal:
I don't have a stable reproducer for it. I saw it few times while running integration tests locally.

PR #635, opened 2022-04-27
docs: Fix installation instructions

mauriciovasquezbernal:
File is now called kubectl-gadget instead of inspektor-gadget


marga-kinvolk (approved):
LGTM

PR #636, opened 2022-04-28
Use hook-mode=auto when deploying to minikube

marga-kinvolk:
fanotify is already used if available, there's no reason why this should
be forced. Putting it in auto allows this rule to be used on
minikube installations that don't support fanotify (like the one used in
the cloud native bpf workshop)


blanquicet (approved):
LGTM!

PR #637, opened 2022-04-28
Review some commands short descriptions

marga-kinvolk:
Commands from a given category should have similar verbiage, matching
what they are doing and not what other categories do. i.e. the commands
in `top` should say that they periodically print their data, and `audit`
should say that it audits the given subsystem.


alban (approved):
This looks good to me. Let's see if others agree :)

blanquicet (approved):
LGTM!

PR #441, opened 2021-12-15
Build inspektor-gadget docker image for arm in our CI

eiffel-fl:
Hi.

This PR cross-builds `inspektor-gadget` docker image for arm64 architecture.
I needed to add a step inside our Dockerfile to build `libbpf-dev` from source because it does not exist a package version of this library for arm64.

The build will only be triggered when pushing to main branch.
The build was already tested and arm64 image were successfully generated:
https://hub.docker.com/layers/kinvolk/gadget/francis-arm/images/sha256-ce3704558e2f42f19b21356a8d0d1e53f2386d9f68656808c5ecda3887d9189c?context=explore

I am currently testing it and it seems a majority of our gadget work:

```
```

TODO:
[ ] Wait for traceloop/66 to be merged and adapt Dockerfile.
[ ] Wait for bcc/16 to be merged and adapt Dockerfile.
[ ] Change `ref/heads` to main.

Best regards.

PR #640, opened 2022-04-29
Build CO-RE gadgets depending on the underlying container architecture

eiffel-fl:
Hi.

In this PR, I added `BPF_ARCH` variable to all eBPF related makefiles.
This variable is set depending on the container architecture and will be used to both create correct `__TARGET_ARCH_*` used by `bpf2go` and to include the `vmlinux.h` for the corresponding architecture.

This PR, applied on top of #441 and with the following patch applied:

```patch
diff --git a/gadget.Dockerfile b/gadget.Dockerfile
index b7af526..bee2453 100644
--- a/gadget.Dockerfile
+++ b/gadget.Dockerfile
@@ -103,3 +103,4 @@ COPY --from=builder /tmp/btfs /btfs/
 
 # Mitigate https://github.com/kubernetes/kubernetes/issues/106962.
 RUN rm /var/run
+RUN rm -r /usr/share/bcc/tools
```

leads to the following gadgets to fail on arm64 hardware:

```
--- FAIL: TestExecsnoop (17.71s): silent fail
--- FAIL: TestProfile (25.44s) : no CORE
--- FAIL: TestTraceloop (42.83s): syscall numbers hardcoded
--- FAIL: TestTcptracer (17.89s): no CORE
--- FAIL: TestBiolatency (20.21s): no CORE
--- FAIL: TestCapabilities (18.17s): no CORE
```

That is to say, all CO-RE gadgets are working except `execsnoop`.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -200,7 +200,7 @@ jobs:
     steps:
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
-      if: github.ref == 'refs/heads/francis/arm'
+      if: github.ref == 'refs/heads/francis/arm-co-re'
     - name: Set up Docker Buildx
       id: buildx
       uses: docker/setup-buildx-action@v1
@@ -231,7 +231,7 @@ jobs:
     - name: Cross build gadget container for amd64 and arm64
       uses: docker/build-push-action@v2
       # We cross build only on main because it takes a lot of time.
-      if: github.ref == 'refs/heads/francis/arm'
+      if: github.ref == 'refs/heads/francis/arm-co-re'
       with:
         context: /home/runner/work/inspektor-gadget/inspektor-gadget/
         file: /home/runner/work/inspektor-gadget/inspektor-gadget/gadget-default.Dockerfile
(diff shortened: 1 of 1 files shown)


eiffel-fl:
> How are you testing it?

Indeed, the image on docker hub is the one corresponding to #441.
I just clone IG, jump to this branch and run `make gadget-container` on an arm64 hardware.

eiffel-fl:
> I tested some gadgets manually and them work fine. It's a great improvement IMO so I'm approving it!
> 
> Thanks for this work, it's amazing.

Thank you for the review.
Nonetheless, I will not merge for the moment because I would like to merge #441 before (I built this one on top of #441).

> NOTE: We still need to figure out btfgen (generated files are only for x86), but I think we can make this on another release. Perhaps we can just disable it on arm64 on the time being to avoid creating weird errors?

I will check how to deactivate it for arm64 and maybe add a commit there.
I am totally fine with the `arm64` support to require `CONFIG_DEBUG_INFO_BTF`.

eiffel-fl:
I rebased on top of #608 and when running `./integration.test -skip-no-co-re -integration -test.timeout 20m &> core.log`, I have the following:

```
--- FAIL: TestExecsnoop (22.88s)
--- FAIL: TestNetworkpolicy (23.75s)
--- FAIL: TestTraceloop (42.67s)
FAIL
```

Failures for `execsnoop` and `traceloop` are normal as there is a [bug in the kernel]() for the first one and the second hardcodes syscall number.
But `networkpolicy` fails with the following:

```
failed to init perf map for IPv4 events: error guessing offsets: invalid guessing state while guessing source address, got checking expected checked
```

I will investigate and try to address this failure in a future PR as `networkpolicy` is mainly a dependency.

PR #646, opened 2022-05-04
Fix breaking typo

marga-kinvolk:
My previous change somehow had removed one hyphen from the file format,
causing the doc renderer to choke.


eiffel-fl (approved):
LGTM!

PR #652, opened 2022-05-06
Fix GetContainers with multiple Docker containers

alban:
local-gadget was reporting wrong names when there are several Docker
containers:
```
» list-containers
zealous_robinson
zealous_robinson
```


blanquicet (approved):
Oops. Sorry!

LGTM. Thanks for managing it!

PR #653, opened 2022-05-06
ci/integration-tests: Clean up resources when the workflow run is cancelled

blanquicet:
# Clean up resources when the workflow run is cancelled

Fixes: https://github.com/kinvolk/inspektor-gadget/issues/641 

Based on the [GitHub documentation](https://docs.github.com/en/actions/managing-workflow-runs/canceling-a-workflow#steps-github-takes-to-cancel-a-workflow-run), this is the proposed solution:

1. Capture the `SIGINT` sent from bash on runner: [here](https://github.com/kinvolk/inspektor-gadget/blob/b47a55db25c1f976e9ac73200e34034512737b58/.github/actions/run-integration-tests/action.yml#L58-L63).
2. Notify the container that the workflow run is being cancelled, [here](https://github.com/kinvolk/inspektor-gadget/blob/b47a55db25c1f976e9ac73200e34034512737b58/.github/actions/run-integration-tests/action.yml#L41-L56). Notice that it is done only when the `SIGINT` is received. For `SIGTERM`, we just capture the signal so that we have that additional 2.5s to finish the cleanup.
3. The integration-tests binary captures the signal [here](https://github.com/kinvolk/inspektor-gadget/blob/b47a55db25c1f976e9ac73200e34034512737b58/integration/integration_test.go#L183-L189) and starts the cleanup procedure (see [here](https://github.com/kinvolk/inspektor-gadget/blob/b47a55db25c1f976e9ac73200e34034512737b58/integration/integration_test.go#L199-L215)):
    a. First, it kills the init commands (in case they are still running).
    b. Wait until the init commands to finish (in case they are still running) so that we do not start undeploying things that are being deployed.
    c. Run the cleanup commands: Remove the possible remaining namespaces, undeploy IG and SPO.

Given that we are not stopping the ongoing tests, there is still the possibility that the tests continue running and it creates the test namespace after we try to delete all the remaining namespaces. This will be managed in another PR.
 

## How to use

There are two options:

- As described in the issue, force-pushing a commit while it is running the integration tests or manually cancelling it on the Actions section of the repo. 
- The other option is to execute the integration-tests locally and hit Ctrl+C.

## Testing done

Check resources are cleaned up when:
- Integration tests are executed successfully
- One of the init commands fails 
- Cancel request (Ctrl+C) arrives when the init commands are being executed
- Cancel request (Ctrl+C) arrives when running the integration tests themselves

I also tested the case when two cancel requests (Ctrl+C) are sent. It is useful when running the integration-tests locally. Of course, in this case, it is not guarantee that the resources will be correctly cleaned up.

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"math/rand"
 	"os/exec"
@@ -25,6 +26,11 @@ import (
 	"testing"
 
 	"github.com/kr/pretty"
+)
+
+const (
+	namespaceLabelKey   string = "scope"
+	namespaceLabelValue string = "ig-integration-tests"
 )
 
 type command struct {
@@ -149,6 +155,16 @@ func (c *command) createExecCmd() {
 	cmd.Stdout = &c.stdout
 	cmd.Stderr = &c.stderr
 
+	// To be able to kill the process of /bin/sh and its child (the process of
+	// c.cmd), we need to send the termination signal to their process group ID
+	// (PGID). However, child processes get the same PGID as their parents by
+	// default, so in order to avoid killing also the integration tests process,
+	// we set the fields Setpgid and Pgid of syscall.SysProcAttr before
+	// executing /bin/sh. Doing so, the PGID of /bin/sh (and its children)
+	// will be set to its process ID, see:
+	// https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/syscall/exec_linux.go;l=32-34.
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: 0}
+
 	c.command = cmd
 }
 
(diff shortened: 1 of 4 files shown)


integration/integration_test.go
```diff
@@ -75,9 +80,41 @@ func runCommands(cmds []*command, t *testing.T) {
...
 }
 
+func cleanupFunc(cleanupDone chan bool, cleanupCommands []*command) {
+	if cancelling {
+		// Wait until the other call of cleanupFunc() is done.
+		<-cleanupDone
+		return
+	}
+
+	cancelling = true
```
mauriciovasquezbernal, line 90: I don't get why you need a channel here. If I understand correctly the behavior should be to avoid running cleanupFunc multiple times. Isn't enough to use the cancelling variable? (the atomic version of it)?

mauriciovasquezbernal (commented):
I tried it and in all the cases the cleanup was done successfully. I think there are still some corner cases were things could fail but IMO it's a very good improvement.

integration/command.go
```diff
@@ -273,13 +375,7 @@ func (c *command) stop(t *testing.T) {
...
 
 	t.Logf("Stop command: %s\n", c.cmd)
-
-	// Here, we just need to send the PID of /bin/sh (which is the same PGID) as
-	// a negative number to syscall.Kill(). As a result, the signal will be
-	// received by all the processes with such PGID, in our case, the process of
-	// /bin/sh and c.cmd.
-	err := syscall.Kill(-c.command.Process.Pid, syscall.SIGTERM)
-
+	err := kill(c.command, c.started)
```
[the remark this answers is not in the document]
blanquicet: I modified the function to use `func (c *command) kill()` and I used internally `c.started`.

integration/command.go
```diff
@@ -194,44 +210,141 @@ func (c *command) verifyOutput() error {
 	return nil
 }
 
-// run runs the command on the given as parameter test.
-func (c *command) run(t *testing.T) {
-	c.createExecCmd()
+// kill kills a command by sending SIGKILL because we want to stop the process
+// immediatly and avoid that the signal is trapped.
+func kill(cmd *exec.Cmd, wait bool) error {
```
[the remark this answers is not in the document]
blanquicet: No, I modified it to follow your suggestion

integration/integration_test.go
blanquicet: For instance, if the workflow run is cancelled during the `initCommands`, the `cleanupFunc()` will be called twice: Once from the [defer](https://github.com/kinvolk/inspektor-gadget/blob/f847eb2daa1773392ff7897ce8f550dae9ee1239/integration/integration_test.go#L237) and the other from [the signal handler](https://github.com/kinvolk/inspektor-gadget/blob/f847eb2daa1773392ff7897ce8f550dae9ee1239/integration/integration_test.go#L214). At this point, `cancelling` will avoid running `cleanupFunc()` multiple times but if we let one of them finish, it will make the main goroutine to finish without waiting for the `cleanupFunc()`  to finish. And that's where the `cleanupDone` channel helps. Does it make sense for you?

blanquicet: As agreed, I kept that channel but moved it outside the `cleanupFunc()` as it can be managed outside and makes the code clearer.

integration/command.go
```diff
@@ -198,44 +214,146 @@ func (c *command) verifyOutput() error {
...
+// to be used in TestMain().
+func (c *command) killWithoutTest() error {
+	fmt.Printf("Kill command(%s)\n", c.name)
+
+	if err := c.kill(); err != nil {
+		return fmt.Errorf("failed to kill command(%s): %w", c.name, err)
+	}
+
+	return nil
+}
```
mauriciovasquezbernal: How is it different than just calling `c.kill()`, is it all about the extra logging information that is printed or are there more differences?

mauriciovasquezbernal (approved):
LGTM once conflicts are solved.

integration/command.go
blanquicet: The extra logging and for keeping all the commands align with the naming `{start|wait|run|kill}WithoutTest()`

PR #655, opened 2022-05-09
integration: Add a flag to skip tests which do not have CO-RE version.

eiffel-fl:
Hi.

In this PR, I added a new flag to skip integration tests for gadgets which do not have a CO-RE version.
This can be useful for #640 and #608.

Best regards.


eiffel-fl:
Thank you for the review!
I just hope we will revert it soon!

PR #656, opened 2022-05-09
containercollection: Implement ContainerCollectionClose()

mauriciovasquezbernal:
Implement method to release resources when they aren't needed anymore.

Fixes https://github.com/kinvolk/inspektor-gadget/issues/624

### DIFF
--- a/pkg/container-collection/container-collection.go
+++ b/pkg/container-collection/container-collection.go
@@ -51,6 +51,12 @@ type ContainerCollection struct {
 
 	// initialized tells if ContainerCollectionInitialize has been called.
 	initialized bool
+
+	// closed tells if ContainerCollectionClose has been called.
+	closed bool
+
+	// functions to be called on ContainerCollectionClose()
+	cleanUpFuncs []func()
 }
 
 // ContainerCollectionOption are options to pass to
@@ -310,4 +316,18 @@ func (cc *ContainerCollection) Unsubscribe(key interface{}) {
 		panic("ContainerCollection's pubsub uninitialized")
 	}
 	cc.pubsub.Unsubscribe(key)
+}
+
+func (cc *ContainerCollection) ContainerCollectionClose() {
+	if !cc.initialized || cc.closed {
+		panic("ContainerCollection is not initialized or has been closed")
+	}
+
+	for _, f := range cc.cleanUpFuncs {
+		f()
+	}
+
+	// TODO: it's not clear if we want/can allow to re-initialize
+	// this instance yet, so we don't set cc.initialized = false.
+	cc.closed = true
 }
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -114,6 +114,12 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
 			return containerRuntimeEnricher(runtime.Name, runtimeClient, container)
 		})
 
+		cc.cleanUpFuncs = append(cc.cleanUpFuncs, func() {
+			if err := runtimeClient.Close(); err != nil {
+				log.Warnf("failed to close container runtime %s: %s", runtime.Name, err)
+			}
+		})
+
 		// Enrich already running containers
 		containers, err := runtimeClient.GetContainers()
 		if err != nil {
@@ -177,13 +183,15 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 			return fmt.Errorf("failed to create Kubernetes client: %w", err)
 		}
 
-		createdChan := make(chan *v1.Pod)
-		deletedChan := make(chan string)
-
-		_, err = k8s.NewPodInformer(nodeName, createdChan, deletedChan)
+		podInformer, err := k8s.NewPodInformer(nodeName)
 		if err != nil {
 			return fmt.Errorf("failed to create pod informer: %w", err)
 		}
+
+		cc.cleanUpFuncs = append(cc.cleanUpFuncs, func() {
+			k8sClient.Close()
+			podInformer.Stop()
+		})
 
 		go func() {
 			// containerIDsByKey keeps track of container ids for each key. This is
@@ -196,13 +204,19 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 
 			for {
 				select {
-				case d := <-deletedChan:
+				case d, ok := <-podInformer.DeletedChan():
+					if !ok {
+						return
+					}
 					if containerIDs, ok := containerIDsByKey[d]; ok {
 						for containerID := range containerIDs {
 							cc.RemoveContainer(containerID)
 						}
 					}
-				case c := <-createdChan:
+				case c, ok := <-podInformer.CreatedChan():
+					if !ok {
+						return
+					}
 					key, _ := cache.MetaNamespaceKeyFunc(c)
 					containerIDs, ok := containerIDsByKey[key]
 					if !ok {
@@ -247,8 +261,6 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 					}
 				}
 			}
-			// TODO: ContainerCollection does not have a Stop() method
-			// podInformer.Stop()
 		}()
 
 		return nil
@@ -266,6 +278,7 @@ func WithInitialKubernetesContainers(nodeName string) ContainerCollectionOption 
 		if err != nil {
 			return fmt.Errorf("failed to create Kubernetes client: %w", err)
 		}
+		defer k8sClient.Close()
 
 		containers, err := k8sClient.ListContainers()
 		if err != nil {
@@ -572,6 +585,10 @@ func WithRuncFanotify() ContainerCollectionOption {
 		if err != nil {
 			return fmt.Errorf("cannot start runc fanotify: %w", err)
 		}
+
+		cc.cleanUpFuncs = append(cc.cleanUpFuncs, func() {
+			runcNotifier.Close()
+		})
 
 		// Future containers
 		cc.containerEnrichers = append(cc.containerEnrichers, func(container *pb.ContainerDefinition) bool {
--- a/pkg/gadgettracermanager/containers-map/tracer.go
+++ b/pkg/gadgettracermanager/containers-map/tracer.go
@@ -62,6 +62,8 @@ type ContainersMap struct {
 	// exposing container details for each mount namespace.
 	containersMap *ebpf.Map
 
+	coll *ebpf.Collection
+
 	pinPath string
 }
 
@@ -96,7 +98,8 @@ func NewContainersMap(pinPath string) (*ContainersMap, error) {
 	}
 	return &ContainersMap{
 		containersMap: m,
-		pinPath:       filepath.Join(pinPath, BPFMapName),
+		pinPath:       pinPath,
+		coll:          coll,
 	}, nil
 }
 
@@ -149,4 +152,5 @@ func (cm *ContainersMap) Close() {
 		return
 	}
 	os.Remove(filepath.Join(cm.pinPath, BPFMapName))
+	cm.coll.Close()
 }
--- a/pkg/gadgettracermanager/k8s/podinformer.go
+++ b/pkg/gadgettracermanager/k8s/podinformer.go
@@ -21,6 +21,7 @@ package k8s
 
 import (
 	"errors"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -45,9 +46,10 @@ type PodInformer struct {
 	stop           chan struct{}
 	createdPodChan chan *v1.Pod
 	deletedPodChan chan string
+	wg             sync.WaitGroup
 }
 
-func NewPodInformer(node string, createdPodChan chan *v1.Pod, deletedPodChan chan string) (*PodInformer, error) {
+func NewPodInformer(node string) (*PodInformer, error) {
 	config, err := rest.InClusterConfig()
 	if err != nil {
 		return nil, err
@@ -90,8 +92,8 @@ func NewPodInformer(node string, createdPodChan chan *v1.Pod, deletedPodChan cha
 		queue:          queue,
 		informer:       informer,
 		stop:           make(chan struct{}),
-		createdPodChan: createdPodChan,
-		deletedPodChan: deletedPodChan,
+		createdPodChan: make(chan *v1.Pod),
+		deletedPodChan: make(chan string),
 	}
 
 	// Now let's start the controller
@@ -101,7 +103,23 @@ func NewPodInformer(node string, createdPodChan chan *v1.Pod, deletedPodChan cha
 }
 
 func (p *PodInformer) Stop() {
+	// tell all workers to end
 	close(p.stop)
+
+	// wait for workers to end before closing channels to avoid
+	// writing to closed channels
+	p.wg.Wait()
+
+	close(p.createdPodChan)
+	close(p.deletedPodChan)
+}
+
+func (p *PodInformer) CreatedChan() <-chan *v1.Pod {
+	return p.createdPodChan
+}
+
+func (p *PodInformer) DeletedChan() <-chan string {
+	return p.deletedPodChan
 }
 
 func (p *PodInformer) processNextItem() bool {
@@ -151,6 +169,7 @@ func (p *PodInformer) Run(threadiness int, stopCh chan struct{}) {
 	}
 
 	for i := 0; i < threadiness; i++ {
+		p.wg.Add(1)
 		go wait.Until(p.runWorker, time.Second, stopCh)
 	}
 
@@ -159,6 +178,8 @@ func (p *PodInformer) Run(threadiness int, stopCh chan struct{}) {
 }
 
 func (p *PodInformer) runWorker() {
+	defer p.wg.Done()
+
 	for p.processNextItem() {
 	}
 }
--- a/pkg/local-gadget-manager/local-gadget-manager_test.go
+++ b/pkg/local-gadget-manager/local-gadget-manager_test.go
@@ -48,6 +48,8 @@ func TestBasic(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to start local gadget manager: %s", err)
 	}
+	defer localGadgetManager.Close()
+
 	gadgets := localGadgetManager.ListGadgets()
 	if len(gadgets) == 0 {
 		t.Fatalf("Failed to get any gadgets")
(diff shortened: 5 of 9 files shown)


alban:
Do you think it is worth to add a test to check for file descriptor & stack leaks?

pkg/local-gadget-manager/local-gadget-manager_test.go already has tests for that. LocalGadgetManager could expose a Close() function that would call ContainerCollectionClose(). Then existing tests could be adapted to test this.

alban (commented):
Apart from the question about testing and the check about closing ContainerCollectionClose twice, the code LGTM.

pkg/container-collection/container-collection.go
```diff
@@ -51,6 +51,9 @@ type ContainerCollection struct {
 
 	// initialized tells if ContainerCollectionInitialize has been called.
 	initialized bool
+
+	// functions to be called on ContainerCollectionClose()
+	cleanUpFuncs []func()
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Added it.

mauriciovasquezbernal:
> Do you think it is worth to add a test to check for file descriptor & stack leaks?
> 
> pkg/local-gadget-manager/local-gadget-manager_test.go already has tests for that. LocalGadgetManager could expose a Close() function that would call ContainerCollectionClose(). Then existing tests could be adapted to test this.

What a great suggestion! I introduced a new test case for this and while doing so I found other cases where we're leaking resources (specially the runcfanotify package). Another review is needed for it then. Thanks!

mauriciovasquezbernal:
I think I'm hitting one of those weird os.File issues here. 

I have the following code compiled using the version of this branch:

```go
package main

import (
	"fmt"

	"github.com/kinvolk/inspektor-gadget/pkg/runcfanotify"
)

func main() {
	runcNotifier, err := runcfanotify.NewRuncNotifier(func(notif runcfanotify.ContainerEvent) {
		switch notif.Type {
		case runcfanotify.EventTypeAddContainer:
			fmt.Printf("add container: %s\n", notif.ContainerID)
		case runcfanotify.EventTypeRemoveContainer:
			fmt.Printf("removecontainer: %s\n", notif.ContainerID)
		}
	})
	if err != nil {
		fmt.Printf("cannot start runc fanotify: %s", err)
		return
	}

	// Uncommenting these two lines causes close() not to be called
	fmt.Print("print enter to finish\n")
	fmt.Scanln()

	runcNotifier.Close()
}
```

If I execute commenting out the two mentioned lines I get this:

```bash
$ go build . && sudo strace -f -e trace='open,openat,fanotify_init,close' -e signal='!SIGURG' ./foo
openat(AT_FDCWD, "/sys/kernel/mm/transparent_hugepage/hpage_pmd_size", O_RDONLY) = 3
close(3)                                = 0
strace: Process 498404 attached
strace: Process 498405 attached
strace: Process 498406 attached
strace: Process 498407 attached
[pid 498403] fanotify_init(FAN_CLASS_CONTENT|FAN_CLOEXEC|FAN_UNLIMITED_QUEUE|FAN_UNLIMITED_MARKS, O_RDONLY|O_CLOEXEC) = 3
[pid 498403] close(3)                   = 0
[pid 498407] +++ exited with 0 +++
[pid 498406] +++ exited with 0 +++
[pid 498405] +++ exited with 0 +++
[pid 498404] +++ exited with 0 +++
+++ exited with 0 +++
```

fanotify_init returns 3 and then it's closed with close(3), all good so far. If I enable those two lines, then:

```
$ go build . && sudo strace -f -e trace='open,openat,fanotify_init,close' -e signal='!SIGURG' ./foo
openat(AT_FDCWD, "/sys/kernel/mm/transparent_hugepage/hpage_pmd_size", O_RDONLY) = 3
close(3)                                = 0
strace: Process 498022 attached
strace: Process 498023 attached
strace: Process 498024 attached
strace: Process 498025 attached
[pid 498021] fanotify_init(FAN_CLASS_CONTENT|FAN_CLOEXEC|FAN_UNLIMITED_QUEUE|FAN_UNLIMITED_MARKS, O_RDONLY|O_CLOEXEC) = 3
print enter to finish

[pid 498025] +++ exited with 0 +++
[pid 498024] +++ exited with 0 +++
[pid 498023] +++ exited with 0 +++
[pid 498022] +++ exited with 0 +++
+++ exited with 0 +++
```

**There is not call to close(3)**. (This missing call to close() is causing the check on the open file descriptors to fail in the local gadget unit tests). 

I have no idea what is going on here. @alban  do you have any idea what could be wrong?

alban:
> @alban do you have any idea what could be wrong?

It is because another thread is doing a read:
```
[pid 1605065] read(3,  <unfinished ...>
```

And Golang will not close it if it still has such a reference ([fd_unix.go;l=105-107](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/internal/poll/fd_unix.go;l=105-107;drc=17dc7b487f2406261d638350e4652fc5df224cc2)):

```
	// The call to decref will call destroy if there are no other
	// references.
	err := fd.decref()
```

I guess the read comes from [fanotify.go](https://github.com/s3rj1k/go-fanotify/blob/9c00a300bb7a91475ab99519109e7aa4b2ef7e6c/fanotify/fanotify.go#L194):
```
	if err := binary.Read(handle.Rd, binary.LittleEndian, event); err != nil {
		return nil, fmt.Errorf("fanotify: event error, %w", err)
	}
```

alban:
It seems to close the file with the following:
```diff
diff --git a/pkg/runcfanotify/runcfanotify.go b/pkg/runcfanotify/runcfanotify.go
index 29271ac7..28d57436 100644
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -92,7 +92,7 @@ var runcPaths = []string{
 
 // initFanotify initializes the fanotify API with the flags we need
 func initFanotify() (*fanotify.NotifyFD, error) {
-       fanotifyFlags := uint(unix.FAN_CLOEXEC | unix.FAN_CLASS_CONTENT | unix.FAN_UNLIMITED_QUEUE | unix.FAN_UNLIMITED_MARKS)
+       fanotifyFlags := uint(unix.FAN_CLOEXEC | unix.FAN_CLASS_CONTENT | unix.FAN_UNLIMITED_QUEUE | unix.FAN_UNLIMITED_MARKS | unix.FAN_NONBLOCK)
        openFlags := os.O_RDONLY | unix.O_LARGEFILE | unix.O_CLOEXEC
        return fanotify.Initialize(fanotifyFlags, openFlags)
 }
```

pkg/container-collection/container-collection.go
```diff
@@ -51,6 +51,12 @@ type ContainerCollection struct {
 
 	// initialized tells if ContainerCollectionInitialize has been called.
 	initialized bool
+
+	// initialized tells if ContainerCollectionInitialize has been closed.
+	closed bool
```
alban, line 56:
```suggestion
	// closed tells if ContainerCollectionInitialize has been closed.
	closed bool
```

pkg/container-collection/container-collection.go
```diff
@@ -311,3 +317,17 @@ func (cc *ContainerCollection) Unsubscribe(key interface{}) {
...
+	if !cc.initialized || cc.closed {
+		panic("ContainerCollection is not initialized or has been closed")
+	}
+
+	for _, f := range cc.cleanUpFuncs {
+		f()
+	}
+
+	// TODO: it's not clear if we want/can allow to re-initialize
+	// this instance yet, so we don't set cc.initialized = false.
```
alban, line 331: I think it is fine. This can be added later if the need comes.

pkg/gadgettracermanager/containers-map/tracer.go
```diff
@@ -75,6 +77,7 @@ func NewContainersMap(pinPath string) (*ContainersMap, error) {
 		return nil, fmt.Errorf("failed to load asset: %w", err)
 	}
 
+	// TODO: can we just defer coll.Close()?
```
alban, line 80: We can't: it would close all the programs and maps.

pkg/local-gadget-manager/local-gadget-manager_test.go
```diff
@@ -159,6 +161,26 @@ func checkFdList(t *testing.T, initialFdList string, attempts int, sleep time.Du
...
+
+	localGadgetManager, err := NewManager([]*containerutils.RuntimeConfig{{Name: "docker"}})
+	if err != nil {
+		t.Fatalf("Failed to start local gadget manager: %s", err)
+	}
+
+	localGadgetManager.Close()
+
+	// We need a quite long try period here because the runc
+	// notifier takes a bit long to release all resources
```
alban, line 180: How long does it take? It is because a file descriptor is closed when the GC releases a os.File? If so, that would be a bug.

pkg/local-gadget-manager/local-gadget-manager_test.go
```diff
@@ -213,6 +236,7 @@ func TestAuditSeccomp(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to start local gadget manager: %s", err)
 	}
+	defer localGadgetManager.Close()
```
alban, line 239: How does that work? With a defer, it is executed after the call to `checkFdList`.

pkg/runcfanotify/runcfanotify.go
```diff
@@ -199,14 +202,17 @@ func (n *RuncNotifier) watchContainerTermination(containerID string, containerPI
...
+		}
 		fds := []unix.PollFd{
 			{
 				Fd:      int32(pidfd),
 				Events:  unix.POLLIN,
 				Revents: 0,
 			},
 		}
-		count, err := unix.Poll(fds, -1)
+		count, err := unix.Poll(fds, 1000)
```
alban, line 215: Is there a way to poll on the fd and on a channel to avoid the timeout?

pkg/container-collection/container-collection.go
mauriciovasquezbernal: ops. Thanks for catching this up.

pkg/local-gadget-manager/local-gadget-manager_test.go
mauriciovasquezbernal: That comment is actually outdated as this long period is not needed anymore. It was fixed by your suggestion to the `fanotifyFlags` and by some closing logic added to `func (n *RuncNotifier) Close()`.

pkg/local-gadget-manager/local-gadget-manager_test.go
mauriciovasquezbernal: `initialFdList` is gathered *after* `NewManager`, hence this test does not take into consideration fds allocated by the local gadget manager. I make this way because I think this test is about the gadget not leaking fds. `TestClose()` tests the local gadget manager.

pkg/gadgettracermanager/containers-map/tracer.go
alban: I think the map would still exist on the bpffs but the fd would be closed in the process. And with #654 we will need to keep it open to pass it around to other gadgets.

pkg/runcfanotify/runcfanotify.go
alban: If a node has 100 containers, that's 100 Go-routines that restart a Poll every second. Probably ok performance wise, but somewhat ugly.

mauriciovasquezbernal: Ok, let's hold this PR for a bit then. I have some ideas to explore:
- Use a single goroutine to poll multiple FDs.
- Save the pidfds on a map and close them on `func (n *RuncNotifier) Close()` making Unix.Poll() return.

pkg/gadgettracermanager/containers-map/tracer.go
mauriciovasquezbernal: You're right. I'll remove this comment then.

pkg/runcfanotify/runcfanotify.go
alban: If you poll multiple FDs, you will need to store the list in `struct RuncNotifier`. Then, this goes a bit in the direction of https://github.com/kinvolk/inspektor-gadget/pull/487#discussion_r871622918: you would need to keep track of containers. So that would simplify the commit 6c17fbfd7f33 about bundleDir.

mauriciovasquezbernal:
@alban I updated the PR to have a single goroutine that watches for all containers terminations. Now the CI is passing (ARO failures are unrelated to this PR), hence another review is needed :D

pkg/container-collection/container-collection.go
eiffel-fl: I still see:

```golang
	// initialized tells if ContainerCollectionInitialize has been closed.
	closed bool
```

pkg/container-collection/options.go
```diff
@@ -266,6 +283,7 @@ func WithInitialKubernetesContainers(nodeName string) ContainerCollectionOption
 		if err != nil {
 			return fmt.Errorf("failed to create Kubernetes client: %w", err)
 		}
+		defer k8sClient.Close()
```
eiffel-fl, line 286: I am not sure to understand.
This functions is already in `cc.cleanUpFuncs`, so it will be called twice?
Should it be removed from the array?

pkg/runcfanotify/runcfanotify.go
```diff
@@ -92,9 +104,33 @@ var runcPaths = []string{
...
+	pid := uintptr(os.Getpid())
+	pidfd, _, errno := unix.Syscall(unix.SYS_PIDFD_OPEN, pid, 0, 0)
+	if errno == unix.ENOSYS {
+		log.Debug("SYS_PIDFD_OPEN is not available")
+		pidfdOpenAvailable = false
+		return
+	}
+
+	if errno != 0 {
+		pidfdOpenAvailable = false
```
eiffel-fl, line 120: Hum...
I am not convinced by the name of this variable, shouldn't it be `pidfdOpenWorking` or something to reflect this idea?
Because in this case, it is available but not working.

eiffel-fl (commented):
Hi.


I did not test it for the moment but there are some points I do not understand.


Best regards.

pkg/container-collection/container-collection.go
mauriciovasquezbernal: You're right. I messed up with a force push and pushed the wrong version. It should be fixed now.

pkg/container-collection/options.go
mauriciovasquezbernal: Those invocations are for different objects. `k8sClient.Close()` on L195 closes the object created on L181. This invocation on L286 closes the object created on L282.

pkg/runcfanotify/runcfanotify.go
mauriciovasquezbernal: I think for practical effects on this code having `SYS_PIDFD_OPEN` available but not working is the same and the `pidfdOpenAvailable` name works fine.

pkg/local-gadget-manager/local-gadget-manager_test.go
mauriciovasquezbernal: Comment removed and timeout updated to be the same as in other tests.

pkg/container-collection/container-collection.go
```diff
@@ -51,6 +51,12 @@ type ContainerCollection struct {
 
 	// initialized tells if ContainerCollectionInitialize has been called.
 	initialized bool
+
+	// closed tells if ContainerCollectionInitialize has been closed.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I went for the first one. thanks!

pkg/runcfanotify/runcfanotify.go
```diff
@@ -72,8 +79,13 @@ type RuncNotifier struct {
 	//
 	// Keys: Container ID
 	// Value: dummy struct
-	containers map[string]struct{}
+	containers map[string]*runcContainer
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Removed that line as I think it's clear from the code what the value is.

pkg/runcfanotify/runcfanotify.go
```diff
@@ -174,80 +216,116 @@ func (n *RuncNotifier) AddWatchContainerTermination(containerID string, containe
...
+		if n.closed {
+			// close all pidfds
+			for _, c := range n.containers {
+				unix.Close(c.pidfd)
+			}
+			return
+		}
+
+		fds := make([]unix.PollFd, len(n.containers))
+		ids := make([]string, len(n.containers))
```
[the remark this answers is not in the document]
mauriciovasquezbernal: You're right. I moved  `n.mu.Lock()` to be before these lines.

pkg/container-collection/options.go
```diff
@@ -180,11 +186,18 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
...
+		podInformer, err := k8s.NewPodInformer(nodeName, createdChan, deletedChan)
 		if err != nil {
 			return fmt.Errorf("failed to create pod informer: %w", err)
 		}
 
+		cc.cleanUpFuncs = append(cc.cleanUpFuncs, func() {
+			k8sClient.Close()
+			podInformer.Stop()
+			close(createdChan)
+			close(deletedChan)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Great catch. I changed the logic a bit then, now created and deleted channels are allocated in `NewPodInformer()` and closed on `podinformer.Stop()`. I also used a waiting group to be sure they are only closed when no goroutine can write to them.

pkg/runcfanotify/runcfanotify.go
```diff
@@ -174,80 +216,116 @@ func (n *RuncNotifier) AddWatchContainerTermination(containerID string, containe
...
 		}
-		count, err := unix.Poll(fds, -1)
-		if err == nil && count == 1 {
+
+		for i, fd := range fds {
+			if fd.Revents == 0 {
+				continue
+			}
+
+			c := n.containers[ids[i]]
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I changed the approach a bit to taking an additional lock here. I used an array to save pointers to containers directly avoiding to access the map here.

pkg/gadgettracermanager/k8s/podinformer.go
```diff
@@ -101,7 +103,23 @@ func NewPodInformer(node string, createdPodChan chan *v1.Pod, deletedPodChan cha
...
+
+	// wait for workers to end before closing channels to avoid
+	// writing to closed channels
+	p.wg.Wait()
+
+	close(p.createdPodChan)
+	close(p.deletedPodChan)
+}
+
+func (p *PodInformer) CreatedChan() chan *v1.Pod {
```
alban, line 117: We could specify a channel directions to increase type safety:

```suggestion
func (p *PodInformer) CreatedChan() <-chan *v1.Pod {
```

pkg/gadgettracermanager/k8s/podinformer.go
```diff
@@ -101,7 +103,23 @@ func NewPodInformer(node string, createdPodChan chan *v1.Pod, deletedPodChan cha
...
+	return p.createdPodChan
+}
+
+func (p *PodInformer) DeletedChan() chan string {
```
alban, line 121:
```suggestion
func (p *PodInformer) DeletedChan() <-chan string {
```

alban (approved):
1 suggestion about type safety, LGTM otherwise. Thanks

pkg/gadgettracermanager/k8s/podinformer.go
mauriciovasquezbernal: TIL! Thanks!

PR #657, opened 2022-05-09
entrypoint: Do not load headers when kheaders mod is available

mauriciovasquezbernal:
Fixes https://github.com/kinvolk/inspektor-gadget/issues/611

### DIFF
--- a/gadget-container/entrypoint.sh
+++ b/gadget-container/entrypoint.sh
@@ -61,63 +61,67 @@ fi
 
 # In the gadget image, /usr/src is a symlink to /host/usr/src (see gadget-*.Dockerfile).
 # If the kernel headers are already available on the gadget pod via the symlink,
-# no need to download them.
+# or the kheaders module is available no need to download them.
 if [ "$ID" = "rhcos" ] && [ ! -d "/usr/src/kernels/$KERNEL" ]; then
-  # Using Centos RPMs because RHCOS ones are not publicly available (they are the same).
-  echo "Fetching kernel-devel from CentOS Vault Mirror..."
+  if grep -q '^kheaders ' /proc/modules || (modprobe kheaders && rmmod kheaders); then
+    echo "kheaders module available, kernel headers won't be downloaded"
+  else
+    # Using Centos RPMs because RHCOS ones are not publicly available (they are the same).
+    echo "Fetching kernel-devel from CentOS Vault Mirror..."
 
-  # CentOS version is composed by <RHEL_VERSION>.<RELEASE_YEAR><RELEASE_MONTH>
-  # https://access.redhat.com/articles/3078#RHEL8
-  case $RHEL_VERSION in
-    "8")
-      CENTOS_VERSION="8.0.1905"
-      ;;
-    "8.1")
-      CENTOS_VERSION="8.1.1911"
-      ;;
-    "8.2")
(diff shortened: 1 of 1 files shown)


gadget-container/entrypoint.sh
```diff
@@ -60,9 +60,11 @@ if grep -q '^1:name=systemd:.*/crio-[0-9a-f]*\.scope$' /proc/self/cgroup > /dev/
 fi
 
 # In the gadget image, /usr/src is a symlink to /host/usr/src (see gadget.Dockerfile).
-# If the kernel headers are already available on the gadget pod via the symlink,
-# no need to download them.
-if [ "$ID" = "rhcos" ] && [ ! -d "/usr/src/kernels/$KERNEL" ]; then
+# If the kernel headers are already available on the gadget pod via the
+# symlink, or the kheaders module is available no need to download them.
+if [ "$ID" = "rhcos" ] && [ ! -d "/usr/src/kernels/$KERNEL" ] && modprobe kheaders > /dev/null 2>&1; then
```
alban, line 65: modprobe has `--quiet`

gadget-container/entrypoint.sh
```diff
@@ -60,9 +60,11 @@ if grep -q '^1:name=systemd:.*/crio-[0-9a-f]*\.scope$' /proc/self/cgroup > /dev/
...
+  rrmod kheaders || true
```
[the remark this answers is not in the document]
alban: If the module was already loaded before the `modprobe` call, this would unload it...
Something like this keeps kheaders loaded/unloaded:
```
if grep -q kheaders /proc/modules || (sudo modprobe kheaders && sudo rmmod kheaders); then
  echo kheaders available
fi
```

gadget-container/entrypoint.sh
mauriciovasquezbernal: Nice!

gadget-container/entrypoint.sh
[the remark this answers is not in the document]
mauriciovasquezbernal: Updated to follow this logic, thanks for the suggestion!

gadget-container/entrypoint.sh
```diff
@@ -61,63 +61,67 @@ fi
...
-      popd
-
-      # In the gadget image, /usr/src is a symlink to /host/usr/src (see gadget-*.Dockerfile).
-      # But in the case of RHCOS, remove the symlink and install files in the container.
-      test ! -L /usr/src || rm -f /usr/src
-      mkdir -p /usr/src/kernels/
-      mv $RPMDIR/usr/src/kernels/$KERNEL /usr/src/kernels/$KERNEL
-    else
-      echo "Failed to fetch kernel-devel package $REPO/$RPM" >&2
+  if grep -q kheaders /proc/modules || (modprobe kheaders && rmmod kheaders); then
```
alban, line 66:
```suggestion
  if grep -q '^kheaders ' /proc/modules || (modprobe kheaders && rmmod kheaders); then
```

alban (approved):
LGTM!

gadget-container/entrypoint.sh
mauriciovasquezbernal: Manually applied. Thanks for the suggestion!

PR #663, opened 2022-05-13
Use k8s golang API to create everything expect trace.

eiffel-fl:
Hi.

In this PR, I removed the YAML template inside `deploy.go` to use k8s golang API to create the objects we need.
Sadly, we still to use `kubectl gadget deploy | kubectl apply -f -` because I did not deal with trace.

Best regards.


cmd/kubectl-gadget/deploy.go
```diff
@@ -346,6 +96,586 @@ type parameters struct {
...
+			},
+			Scope:    apiv1.NamespaceScoped,
+			Versions: []apiv1.CustomResourceDefinitionVersion{
+				apiv1.CustomResourceDefinitionVersion{
+					Name:  "v1alpha1",
+					Schema: &apiv1.CustomResourceValidation{
+						OpenAPIV3Schema: &apiv1.JSONSchemaProps{
+							Description: "Trace is the Schema for the traces API",
+							Properties: map[string]apiv1.JSONSchemaProps{
+								"apivVersion": apiv1.JSONSchemaProps{
```
mauriciovasquezbernal, line 265:
```suggestion
								"apiVersion": apiv1.JSONSchemaProps{
```

cmd/kubectl-gadget/deploy.go
```diff
@@ -346,6 +96,586 @@ type parameters struct {
...
+										"gadget": apiv1.JSONSchemaProps{
+											Description: "Gadget is the name of the gadget such as \"seccomp\"",
+											Type: "string",
+										},
+										"node": apiv1.JSONSchemaProps{
+											Description: "Node is the name of the node on which this trace should run",
+											Type: "string",
+										},
+										"output": apiv1.JSONSchemaProps{
+											Description: "Output allows a gadget to output the results in the specified location. * With OutputMode=Status|Stream, Output is unused * With OutputMode=File, Output specifies the file path * With OutputMode=ExternalResource, Output specifies the external   resource (such as   seccompprofiles.security-profiles-operator.x-k8s.io for the seccomp gadget)",
```
mauriciovasquezbernal, line 314: Break this long line?

mauriciovasquezbernal (commented):
Looks very good as a first version, some comments:
- I think it'll be nice to add a logic to wait until gadget pods are ready, it'd guarantee the user can start using it just after deploying
- I think in this case it's easier if we put everything in a single commit 


I tried it and it didn't work for me, do you know what's wrong with the traces?

alban (changes requested):
What is the rationale of switching to Go code?

Is it possible for the tool to still output the full yaml (including the comments)? That could be useful for integrating Inspektor Gadget with other systems when we don't want to deploy IG with the kubectl-gadget command.

With this PR, the trace CRD spec is now hardcoded in Go instead of being automatically generated. Previously, the workflow was:
- Edit Go structs like TraceSpec directly in pkg/apis/gadget/v1alpha1/types.go
- Run 'make manifests generate' to generate pkg/resources/crd/bases/gadget.kinvolk.io_traces.yaml
- It is included in pkg/resources/embed.go
- cmd/kubectl-gadget/deploy.go automatically inserts resources.TracesCustomResource

But with this PR, it seems the synchronisation between TraceSpec and createTraceCustomResource() is manual, so it is easy to make mistakes.

eiffel-fl:
> * I think it'll be nice to add a logic to wait until gadget pods are ready, it'd guarantee the user can start using it just after deploying

It seems a good idea, I will try to add this to this PR.

> * I think in this case it's easier if we put everything in a single commit

I prefer to have several commits, it is easier to review and easier to blame.

> I tried it and it didn't work for me, do you know what's wrong with the traces?

It is on purpose, for this PR, the trace will still be created in YAML.
I will address trace in #666.

eiffel-fl:
> What is the rationale of switching to Go code?
> 
> Is it possible for the tool to still output the full yaml (including the comments)? That could be useful for integrating Inspektor Gadget with other systems when we don't want to deploy IG with the kubectl-gadget command.
> 
> With this PR, the trace CRD spec is now hardcoded in Go instead of being automatically generated. Previously, the workflow was:
> 
> * Edit Go structs like TraceSpec directly in pkg/apis/gadget/v1alpha1/types.go
> * Run 'make manifests generate' to generate pkg/resources/crd/bases/gadget.kinvolk.io_traces.yaml
> * It is included in pkg/resources/embed.go
> * cmd/kubectl-gadget/deploy.go automatically inserts resources.TracesCustomResource
> 
> But with this PR, it seems the synchronisation between TraceSpec and createTraceCustomResource() is manual, so it is easy to make mistakes.

I understand your point of view, but as Mauricio stated the goal is to avoid having to pipe `kubectl gadget deploy` to actually deploy it.
A simple solution could be to keep the YAML and do something like this in the code:

```golang
exec.command(fmt.Sprintf("echo %s | kubect apply", yaml))
```

But I do not really like having call to `exec()` in code.

cmd/kubectl-gadget/deploy.go
eiffel-fl: Good catch, thank you!

cmd/kubectl-gadget/deploy.go
eiffel-fl: I prefer to leave it like this as it is easier for someone to grep it.

mauriciovasquezbernal:
What's the advantage of splitting this in two PRs? IMO this change should go altogether in a single PR to avoid an intermediate step where some things are created using the API and others through `kubectl apply`

eiffel-fl:
> What's the advantage of splitting this in two PRs?

For now, one part is working and the other not and they adress different stuff (one for CRD, the other for k8s related components) hence the split in two PR.

eiffel-fl:
I was able to debug the small problem for trace CRD and added its commit on top of this PR.

mauriciovasquezbernal:
(Second time writing this comment as I closed my tab by mistake :angry: )

> But with this PR, it seems the synchronisation between TraceSpec and createTraceCustomResource() is manual, so it is easy to make mistakes.

I think this PR is still missing this point. I was thinking that we can use a different strategy:
1. Have yaml manifests to deploy IG on our repo. These are very similar to what we have on the template in the deploy.go file, but filling the template parameters with some default values. Having a yaml file will allow users to deploy IG without using kubectl-gadget.
2. Embed that file file in go and apply it programmatically to the server. (I was looking and it seems possible from a technical point of view https://stackoverflow.com/questions/36307950/kubernetes-api-call-equivalent-to-kubectl-apply, https://stackoverflow.com/questions/60510569/the-equivalent-way-of-doing-kubectl-apply-in-golang, https://github.com/kubernetes/client-go/issues/193#issuecomment-363318588). 

What do you all think?

PR #664, opened 2022-05-14
generate demo videos (asciinema gifs)

alban:
# generate demo videos (asciinema gifs) from Github Action

This reuses https://github.com/thockin/micro-demos to automatically generate demo videos from scripts.

demo

## How to use

Two choices:
* Run `make -C demos`
* Download the demos artifact from the GitHub action

## Testing done

## TODO

- [ ] Add demo scripts for other gadgets
- [ ]


marga-kinvolk:
The video you pasted here shows nothing in the `get pods` output, any idea why? I would have expected to at least see the pod in ContainerCreating state. Also, it goes away rather fast, can we make it slow down?

alban:
I don't know why the pod is missing. I would have expected ContainerCreating as well.

mauriciovasquezbernal (commented):
I think it's a good approach, i.e. having a source code of the video to make it easier to update them. I don't fully understand the goal of the github actions here. I think the work flow is:
- developer updates the "source video script"
- they run "make videos" (or similar) to update / generate the video
- both, the source and the generated video are committed to the repo

Am I missing something on the use case?

It's true that we could have some testing to be sure the video is updated, but I'm not sure it's that important.

blanquicet:
> developer updates the "source video script"
>
@mauriciovasquezbernal It would be great if we can automate this. Developers can easily forget about it, we have seen it 😄 . IIUC, Alban is already generating the demo GIF in the CI, we "just" need to find a way to verify if it is updated (maybe a regexp of the `demo.cast`?). And then, create a PR (it already exists, e.g. https://github.com/marketplace/actions/create-pull-request).

> It's true that we could have some testing to be sure the video is updated, but I'm not sure it's that important.
>
IMO, it is quite important to try to find a way to keep the documentation aligned with the developments.


> Also, it goes away rather fast, can we make it slow down?
>
@marga-kinvolk, yeah!

demo

demos/gen.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -ex
+
+command -v pv
+
+kubectl gadget deploy | kubectl delete -f - || true
```
blanquicet, line 7: Please use the `undeploy` command.

mauriciovasquezbernal:
> @mauriciovasquezbernal It would be great if we can automate this. Developers can easily forget about it, we have seen it smile . IIUC, Alban is already generating the demo GIF in the CI, we "just" need to find a way to verify if it is updated (maybe a regexp of the `demo.cast`?). And then, create a PR (it already exists, e.g. https://github.com/marketplace/actions/create-pull-request).

The idea is then to generate the GIF on the CI and automatically open a PR for them? IMHO it's too complex, I'd say that only checking if the GIFs are updated is enough (btw, we need to do something similar for the compiled BPF objects https://github.com/kinvolk/inspektor-gadget/pull/691). But if you think this approach makes sense and not against it, we can try it out and see how it works.

blanquicet:
BTW, the video about the installation should also show how to install `kubectl-gadget`.

alban:
Some progress with videos for:
* install
* trace-bind
* trace-exec

Check the "demos" artifact: https://github.com/alban/inspektor-gadget/actions/runs/2671865732#artifacts

demos/install/init.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+kubectl gadget undeploy || true
+kubectl delete ns demo || true
```
mauriciovasquezbernal, line 4: What is this file supposed to do? Execute this commands **before** running the demo?

mauriciovasquezbernal (commented):
It's looking very nice. I made some small comments.

Do you think it'll be possible to upload the source .cast files to asciinema and link them from the corresponding gif? (Just in case somebody wants to copy a command instead of having to type it all).

demos/install/init.sh
alban: Yes

alban:
I updated the PR with a few more demos:

https://github.com/kinvolk/inspektor-gadget/blob/alban_demos/docs/guides/trace/bind.gif
https://github.com/kinvolk/inspektor-gadget/blob/alban_demos/docs/guides/trace/capabilities.gif
https://github.com/kinvolk/inspektor-gadget/blob/alban_demos/docs/guides/trace/dns.gif
https://github.com/kinvolk/inspektor-gadget/blob/alban_demos/docs/guides/trace/exec.gif
https://github.com/kinvolk/inspektor-gadget/blob/alban_demos/docs/guides/trace/fsslower.gif

tools/demos/install/demo.sh
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. $(dirname ${BASH_SOURCE})/../util.sh
+
+desc "Let's install Inspektor Gadget"
+run "kubectl gadget deploy | kubectl apply -f -"
+
+desc "Now, wait for the pods to be ready"
+run "kubectl wait --for=condition=ready -n gadget pod -l k8s-app=gadget"
```
mauriciovasquezbernal, line 22: It can now be replaced with `kubectl gadget deploy`

mauriciovasquezbernal (commented):
I think the generated gifs are great. Do you think we could put those at the beginning of each guide?

alban:
I'll switch from [asciicast2gif](https://github.com/asciinema/asciicast2gif) to [agg](https://github.com/asciinema/agg) because asciicast2gif is very slow and can from my experience needs more than 10GB of memory in some videos. See https://github.com/asciinema/asciicast2gif/issues/51#issuecomment-1214385482

alban:
> I think the generated gifs are great. Do you think we could put those at the beginning of each guide?

Done

tools/demos/trace-bind/demo.sh
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+. $(dirname ${BASH_SOURCE})/../util.sh
+
+desc "Let's start a pod that's listening on TCP port 4242"
+run "kubectl run -n demo test-pod --image busybox -- sh -c 'while true ; do nc -l -p 4242 -w 1 ; done'"
+run "kubectl wait -n demo --for=condition=ready pod/test-pod"
+
+desc "Let's trace"
+run "kubectl gadget trace bind -n demo --timeout 3"
```
mauriciovasquezbernal, line 23: Is it possible to add a sleep/delay at the end. I find that it terminates very quickly and one it not able to ready the result of the trace. Also having a comment indicating where the demo ends would be useful since it's played in a loop. (I suppose it applies for all gadgets too).

tools/demos/install/demo.sh
```diff
@@ -0,0 +1,26 @@
...
+
+. $(dirname ${BASH_SOURCE})/../util.sh
+
+desc "Let's install Inspektor Gadget"
+run "kubectl gadget deploy"
+
+run "kubectl get pod -n gadget"
+
+desc "And check the installed version"
+run "kubectl logs -n gadget -l k8s-app=gadget --tail=-1 | grep 'Inspektor Gadget version'"
```
mauriciovasquezbernal, line 24: nit (for another PR): It should be handled by "kubectl gadget version" as kubectl does -> https://github.com/kinvolk/inspektor-gadget/issues/902.

tools/demos/trace-bind/demo.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Copyright 2016-2022 The Kubernetes Authors.
```
mauriciovasquezbernal, line 2: Is this line correct?

mauriciovasquezbernal (approved):
Besides a minor comment to add some more delay at the end of the demo it LGTM!

alban:
The CI fails even after restarting it 2 times. But it does not seem related to this PR.

blanquicet (approved):
I think it is awesome @alban! 

I just don't enjoy seeing some misaligned columns and some entries larger than the terminal so that it goes to the next line 🙁. However, we can go ahead with this PR and create a couple of issues for some improvements and generate the gif for the missing gadgets. In addition, we could use the same approach for demoing `local-gadget` 🎉!

Some improvements we can consider:

- I would suggest moving all gifs within a specific folder. I would look more organised. 
- We could unify all the `fini.sh` and `init.sh` scripts as they seems to be quite the same.

demos/trace-exec/demo.sh
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. $(dirname ${BASH_SOURCE})/../util.sh
+
+desc "Let's trace new processes"
+run "kubectl get pod -n demo"
+run "kubectl gadget trace exec -n demo --timeout 4"
```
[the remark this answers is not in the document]
blanquicet: Maybe using a cluster with shorter node names? 🤣

PR #667, opened 2022-05-23
integration: Make sure default service account is created

mauriciovasquezbernal:
Sometimes we're hitting the following issue:
pods "test-pod" is forbidden: error looking up service account xxx: serviceaccount "default" not found

Let's add a bit of extra logic after the namespace creation to be sure
the default service account for the namespace is created before going
on.


blanquicet (approved):
LGTM!

PR #670, opened 2022-05-27
ci/integration-tests: Install a fixed version of SPO and CertManager

blanquicet:
# Install a fixed version of SPO and CertManager

This PR modifies the integration tests to not install the SPO from the `main` branch but the latest released version to ease the maintenance of the CI. In this way, we just need to verify that the integration tests are working correctly after each new release instead of dealing with all the changes introduced to the `main` branch.

This PR also avoids limiting the resources when the integration tests are running on ARO where we have 6 nodes and we don't have resource limitations like in the case of Minikube running on a GH runner.

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -92,9 +92,9 @@ var deploySPO *command = &command{
 	// security-profiles-operator-webhook to be started, hence the long
 	// timeout
 	cmd: `
-	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.yaml
+	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.2/cert-manager.yaml
 	kubectl --namespace cert-manager wait --for condition=ready pod -l app.kubernetes.io/instance=cert-manager
-	curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/main/deploy/operator.yaml | \
+	curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.2/deploy/operator.yaml | \
 		sed 's/replicas: 3/replicas: 1/'|grep -v cpu: | \
 		kubectl apply -f -
 	for i in $(seq 1 120); do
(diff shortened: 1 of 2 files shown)


integration/command.go
```diff
@@ -82,32 +82,39 @@ var waitUntilInspektorGadgetPodsDeployed *command = &command{
...
-	kubectl --namespace security-profiles-operator wait --for condition=ready pod -l app=security-profiles-operator || (kubectl get pod -n security-profiles-operator ; kubectl get events -n security-profiles-operator ; false)
-	`,
-	expectedRegexp: "pod/security-profiles-operator-.*-.* condition met",
+	// security-profiles-operator-webhook to be started, hence the long timeout
+	cmdStr := fmt.Sprintf(`
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.2/cert-manager.yaml
+kubectl --namespace cert-manager wait --for condition=ready pod -l app.kubernetes.io/instance=cert-manager
+curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.2/deploy/operator.yaml | \
+  %s \
+  kubectl apply -f -
```
alban, line 101: I wonder if it would be more readable if the Go code and bash code were not interleaved like this. Something like:
```
curl https://... | \
if [ %v = true ] ; then 
  sed 's/replicas: 3/replicas: 1/'|grep -v cpu:
fi | \
kubectl apply -f -
```
When fmt.Sprintf() is given a boolean like `limit` with %v, it renders the string "true" or "false".

What do you think?

integration/integration_test.go
```diff
@@ -117,7 +117,12 @@ func testMain(m *testing.M) int {
 	initCommands := []*command{}
 
 	if !*doNotDeploySPO {
-		initCommands = append(initCommands, deploySPO)
+		limit := false
+		if *k8sDistro == K8sDistroMinikubeGH {
+			limit = true
```
alban, line 122: Is it tested on Minikube outside of GitHub Actions? I wonder if Minikube on developers' computers might need the same limitation.

alban (commented):
Some questions first, but other than that it looks good to me.

integration/command.go
blanquicet: Yeah, I think it makes the code more readable.

integration/integration_test.go
blanquicet: However, I tried with Docker driver because it is the one used by the CI.

blanquicet: Never mind. I had a problem during the tests. As we were expecting, the limitation is due to the limitation of the GH runners where Minikube is running and not on Minikube itself. Therefore, I will not make any changes to this part of the code.

mauriciovasquezbernal (commented):
Just a nit. I haven't tested it myself, but I see the integration tests on GHA are passing.

integration/command.go
```diff
@@ -82,34 +82,38 @@ var waitUntilInspektorGadgetPodsDeployed *command = &command{
 	done`,
 }
 
-var deploySPO *command = &command{
-	name: "Deploy Security Profiles Operator (SPO)",
-	// The security-profiles-operator-webhook deployment is not part of the
-	// yaml but created by SPO. We cannot use kubectl-wait before it is
-	// created, see also:
-	// https://github.com/kubernetes/kubernetes/issues/83242
+func deploySPO(limit bool) *command {
```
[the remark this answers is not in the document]
blanquicet: Make sense.

mauriciovasquezbernal (approved):

PR #671, opened 2022-05-31
Use gadget-default-container as minikube-install dependency.

eiffel-fl:
Hi.

In this PR, I `s/gadget-container/gadget-default-container/` for `minikube-install` Makefile target.
For now, I prefer to use default instead of CO-RE.

Best regards.


blanquicet (approved):
I didn't test it but LGTM from code review.

eiffel-fl:
Thank you for the review!

PR #672, opened 2022-06-01
ci/integration-tests: Improve SPO installation and enable AuditSeccomp test on ARO

blanquicet:
# Improve SPO installation and enable AuditSeccomp test on ARO

This PR fixes two issues:

1. The SPO installation was sometimes failing with timeout (also in Minikube) because we were not waiting correctly for the SPO daemon to be ready. 
2. The AuditSeccomp test was failing on ARO because the seccomp profile took around 2 minutes to be installed into the nodes. After investigating this issue, I found that those 2 minutes are not needed to install the seccomp profile but to wait until the SPO daemon is ready. That's why these two issues are correlated. Therefore, I decided to create one single PR to solve them.

Proposed solutions:

1. Manually wait for the SPO daemon to be created (It is not installed as part of the YAML but by the SPO so we cannot use kubectl-wait). Then, if needed, remove the resource management from the SPO webhook and daemon and let Kubernetes use the BestEffor QoS approach. And finally, wait for the daemon set to be available.
2. Given that we are now waiting for the SPO daemon to be ready. The seccomp profile is installed in a couple of seconds as expected. I took advantage of this PR to modify the test to use `--timeout`.

## How to use

Just run the integration tests on different k8s distro. I tested ARO, AKS and Minikube on GH.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -169,11 +169,13 @@ func testMain(m *testing.M) int {
 
 	if !*doNotDeploySPO {
 		limitReplicas := false
+		bestEffortResourceMgmt := false
 		if *k8sDistro == K8sDistroMinikubeGH {
 			limitReplicas = true
+			bestEffortResourceMgmt = true
 		}
 
-		initCommands = append(initCommands, deploySPO(limitReplicas))
+		initCommands = append(initCommands, deploySPO(limitReplicas, bestEffortResourceMgmt))
 		cleanupCommands = append(cleanupCommands, cleanupSPO)
 	}
 
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -269,11 +269,8 @@ func TestMain(m *testing.M) {
 }
 
 func TestAuditSeccomp(t *testing.T) {
-	if *k8sDistro == K8sDistroARO {
-		t.Skip("Skip running audit-seccomp gadget on ARO: see issue #631")
-	}
-
 	ns := generateTestNamespaceName("test-audit-seccomp")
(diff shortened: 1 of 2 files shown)


integration/integration_test.go
```diff
@@ -300,9 +299,13 @@ spec:
     names:
     - mkdir
 EOF
-			`, ns),
+			`, spName, ns),
 			expectedRegexp: "seccompprofile.security-profiles-operator.x-k8s.io/log created",
```
mauriciovasquezbernal, line 303: Nit: this line should use `spName` instead of hardcoding "log".

integration/integration_test.go
```diff
@@ -330,7 +333,7 @@ EOF
 		waitUntilTestPodReadyCommand(ns),
 		{
 			name:           "RunAuditSeccompGadget",
-			cmd:            fmt.Sprintf("$KUBECTL_GADGET audit seccomp -n %s & sleep 5; kill $!", ns),
+			cmd:            fmt.Sprintf("$KUBECTL_GADGET audit seccomp -n %s --timeout 15", ns),
```
mauriciovasquezbernal, line 336: Is the change from 5 to 15 intended?

mauriciovasquezbernal (approved):
I'm not an expert on this area so I can't provide so much feedback. The changes LGTM and the tests are passing, so I'm approving, but I think it needs a review from @alban who originally introduced this. 

btw, it fixes https://github.com/kinvolk/inspektor-gadget/issues/631, right?

alban (approved):
LGTM, thanks!

PR #680, opened 2022-06-07
Add CO-RE for biolatency

eiffel-fl:
Hi.

In this PR, I added CO-RE code for biolatency.

Best regards.


pkg/gadgets/biolatency/tracer/core/tracer.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	return sb.String()
+}
+
+// slotsToString convert array of values to an histogram.
+// It is a golang translation of iovisor/bcc function print_log2_hist as time of
+// commit:
+// 3787a0d23f75 ("libbpf-tools: improve check for fentry program support")
+func slotsToString(slots []uint32) string {
```
mauriciovasquezbernal, line 119: It seems to me this is following the "standard" implementation but I think we should make it differently here. I think we should define a type structure for this gadget (as we do for other gadgets), then we should marshal to json (as done in other gadgets as well) and the client should contain this logic to "pretty print" the event. 

The standard tool already supports a -j output, I think we can take some inspiration from there to define the new type struct for this gadget. 

In summary, the changes I'm proposing are:
1. Define a data for the events generated by this gadget 
2. Store the output in json format in the trace (use -j for the standard tool and adapt the core one)
3. Move the "slotsToString" logic to the client

mauriciovasquezbernal (changes requested):
I think we should move more logic to the client side, to make this tool more usable by external components like mentioned here https://github.com/kinvolk/inspektor-gadget/issues/285.

pkg/gadgets/biolatency/tracer/core/tracer.go
eiffel-fl: Done!

pkg/gadgets/biolatency/tracer/core/bpf/biolatency.bpf.c
```diff
@@ -78,44 +68,19 @@ int trace_rq_start(struct request *rq, int issue)
...
-	/**
-	 * commit a54895fa (v5.11-rc1) changed tracepoint argument list
-	 * from TP_PROTO(struct request_queue *q, struct request *rq)
-	 * to TP_PROTO(struct request *rq)
-	 */
-	if (LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 11, 0))
-		return trace_rq_start((void *)ctx[1], false);
-	else
-		return trace_rq_start((void *)ctx[0], false);
+	return trace_rq_start((void *)ctx[1], false);
```
mauriciovasquezbernal, line 71: ditto.

pkg/gadgets/biolatency/tracer/core/tracer.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return report, nil
+}
+
+func (t *Tracer) Stop() (string, error) {
+	t.blockRqCompleteLink = gadgets.CloseLink(t.blockRqCompleteLink)
+	t.blockRqInsertLink = gadgets.CloseLink(t.blockRqInsertLink)
+	t.blockRqIssueLink = gadgets.CloseLink(t.blockRqIssueLink)
+
+	report, err := getReport(t.objs.biolatencyMaps.Hists)
```
mauriciovasquezbernal, line 100: This can cause crashes if `Stop()` is called before this map is created. (It could happen if the program fails to load on L112 below). An easy solution is to add a conditional checking if `t.objs.Hists != nil`, it's not cleanest way but I think it's fine.

cmd/kubectl-gadget/profile/block-io.go
```diff
@@ -94,11 +157,13 @@ func runBiolatency(cmd *cobra.Command, args []string) error {
...
 
-		// remove message printed by BCC tracer to avoid printing it twice
-		ret := strings.ReplaceAll(results[0].Status.Output,
-			"Tracing block device I/O... Hit Ctrl-C to end.\n", "")
+		var report types.Report
+		if err := json.Unmarshal([]byte(results[0].Status.Output), &report); err != nil {
+			return utils.WrapInErrUnmarshalOutput(err, results[0].Status.Output)
+		}
+
+		fmt.Printf("%s", reportToString(report))
```
mauriciovasquezbernal, line 165: Are you missing the logic to print the results in json format?

```
./kubectl-gadget profile block-io --node ubuntu-hirsute -o json
Tracing block device I/O... Hit Ctrl-C to end.^C
     usecs               : count    distribution
         1 -> 1          : 0        |                                        |
         2 -> 3          : 0        |                                        |
         4 -> 7          : 0        |                                        |
         8 -> 15         : 0        |                                        |
        16 -> 31         : 0        |                                        |
        32 -> 63         : 7        |****************                        |
        64 -> 127        : 17       |****************************************|
       128 -> 255        : 6        |**************                          |
```

mauriciovasquezbernal (changes requested):
After applying the following changes it worked fine on my machine (kernel 5.11.0-49).

```diff
index 2eeffde0..b389e1f8 100644
diff --git pkg/gadgets/biolatency/tracer/core/bpf/biolatency.bpf.c pkg/gadgets/biolatency/tracer/core/bpf/biolatency.bpf.c
index 9dd02f9b..f10e79f3 100644
--- pkg/gadgets/biolatency/tracer/core/bpf/biolatency.bpf.c
+++ pkg/gadgets/biolatency/tracer/core/bpf/biolatency.bpf.c
@@ -68,13 +68,13 @@ int trace_rq_start(struct request *rq, int issue)
 SEC("tp_btf/block_rq_insert")
 int block_rq_insert(u64 *ctx)
 {
-       return trace_rq_start((void *)ctx[1], false);
+       return trace_rq_start((void *)ctx[0], false);
 }
 
 SEC("tp_btf/block_rq_issue")
 int block_rq_issue(u64 *ctx)
 {
-       return trace_rq_start((void *)ctx[1], true);
+       return trace_rq_start((void *)ctx[0], true);
 }
 
 SEC("tp_btf/block_rq_complete")
diff --git pkg/gadgets/biolatency/tracer/core/tracer.go pkg/gadgets/biolatency/tracer/core/tracer.go
index e4e17e53..4cb9f599 100644
--- pkg/gadgets/biolatency/tracer/core/tracer.go
+++ pkg/gadgets/biolatency/tracer/core/tracer.go
@@ -97,8 +97,12 @@ func (t *Tracer) Stop() (string, error) {
        t.blockRqInsertLink = gadgets.CloseLink(t.blockRqInsertLink)
        t.blockRqIssueLink = gadgets.CloseLink(t.blockRqIssueLink)
 
-       report, err := getReport(t.objs.biolatencyMaps.Hists)
-       t.objs.Close()
+       defer t.objs.Close()
+
+       if t.objs.Hists == nil {
+               return "", nil
+       }
+       report, err := getReport(t.objs.Hists)
        if err != nil {
                return "", err
        }
@@ -106,6 +110,7 @@ func (t *Tracer) Stop() (string, error) {
        output, err := json.Marshal(report)
 
        return string(output), err
+
 }
 
 func (t *Tracer) start() error {
```

Once that problem is fixed it can be merged. It's great to see how the --json support was implemented.

mauriciovasquezbernal (commented):
I like the idea of compiling for different kernel version (I'm confident we'll be able to remove this workaround soon once the right supports lands on cilium/ebpf). 

What about using https://pkg.go.dev/github.com/moby/moby@v20.10.17+incompatible
/pkg/parsers/kernel to check the kernel version?

mauriciovasquezbernal (approved):
Just one additional comment regarding the moby library, however I'll approve supposing it'll be fixed.

cmd/kubectl-gadget/profile/block-io.go
```diff
@@ -76,14 +137,20 @@ func runBiolatency(cmd *cobra.Command, args []string) error {
 			time.Sleep(time.Duration(params.Timeout) * time.Second)
 			c <- os.Interrupt
 		}()
-		fmt.Printf("Tracing block device I/O...")
+		if params.OutputMode != utils.OutputModeJSON {
+			fmt.Printf("Tracing block device I/O...\n")
+		}
 	} else {
 		fmt.Printf("Tracing block device I/O... Hit Ctrl-C to end.")
 	}
```
mauriciovasquezbernal, line 145: I think this should be printed when output mode is not json, right? Otherwise a tool parsing this could fail.

mauriciovasquezbernal: I think everybody knows that the standard way to stop a running command is Ctrl + C, so I think it's not a problem.

PR #691, opened 2022-06-10
gadgets: Remove bpf makefiles

mauriciovasquezbernal:
The "ebpf-objects" target on gadget-container/Makefile forces to
recompile the ebpf objects, i.e, we're re-compiling all the eBPF objects
again when building the gadget container image, making it inefficient
and, in some cases, causing weird problems becuase the clang version
used inside the container is different than the one used to compiled the
shipped object.

We're now using go generate to invoke bpf2go in all the gadgets, hence
the makefiles in the bpf folders of all gadgets and the target mentioned
above are not needed anymore. Running "go generate ./..." on the root
folder will update all genereted files (invoking clang to compile the
ebpf programs.

Btw, the first commit of this PR shows this problem, there are some
updated BPF objects on the repo but we weren't aware because they were
compiled always when building the container image and we explicitly when
building the dependencies for local-gadget.


eiffel-fl (commented):
Hi.

I am not 100% convinced by this contribution.
Normally, I welcome all contributions which goal is to reduce the number of files and/or lines.
But this one will make CO-RE arm64 cross-compilation really tricky as  it relies on Makefile to do so (d0fb162d3478345d0cc186d7d22cf0f4b70bdc06).

Also, it means we will not update the `.o` when shipping a new container image, so we will need to be sure that these files are up to date.

Best regards.

Makefile
```diff
@@ -68,7 +68,7 @@ local-gadget-%: phony_explicit
 	if [ "$(shell go env GOOS)" = "$${GOOS}" -a "$(shell go env GOARCH)" = "$${GOARCH}" ] ; then \
 		export CGO_ENABLED=1 && \
 		export GO111MODULE=on && \
-		make -C gadget-container ebpf-objects && \
+		make -C gadget-container && \
```
alban, line 71: We should just delete this line (and possibly replace it with `go generate ./...`?). Since the default target of `gadget-container/Makefile` is `gadget-container-deps`, this change will now build `gadgettracermanager` and other binaries, which does not make sense for local-gadget.

Ditto below line 152 in the `local-gadget-tests` target.

mauriciovasquezbernal:
> But this one will make CO-RE arm64 cross-compilation really tricky as it relies on Makefile to do so ([d0fb162](https://github.com/kinvolk/inspektor-gadget/commit/d0fb162d3478345d0cc186d7d22cf0f4b70bdc06)).

I don't think the individual makefiles on each gadget are needed by that PR. One could do `BPF_ARCH=$(BPF_ARCH) go generate ./...` in https://github.com/kinvolk/inspektor-gadget/pull/640/files#diff-5404b12decf531eab5ad4fe9888d2fd13003355ad2c839e881bf3cd40d34d07aR10 and it should work fine. 

> Also, it means we will not update the `.o` when shipping a new container image, so we will need to be sure that these files are up to date.

I think in anycase we must be sure object files are updated, so it's not a concern to me.

Makefile
mauriciovasquezbernal: Good catch. For the time being I think the right solution is to remove those files, as replacing with `go generate ./...` will cause the ebpf objects to be recompiled again, what I'm not sure is a good idea.

mauriciovasquezbernal:
> What is the difference between `-target` and `-D__TARGET`, does the first supersedes the second?

Those are different. `-target` is the clang target and `-D__TARGET` is used in https://github.com/libbpf/libbpf/blob/master/src/bpf_tracing.h to define the correct `__PT_*` macros. 

bpf2go automatically defines  `-D__TARGET` (https://github.com/cilium/ebpf/blob/d17ebbefb05d7ebc1f1bac3b1c445369ff859f09/cmd/bpf2go/main.go#L295) according to the target, so we don't need to define it ourselves.

mauriciovasquezbernal:
I pushed a commit removing -D__TARGET, if this works fine it'll simplify things a lot, we could easily compile objects for arm64 only by adding that target.

pkg/gadgets/process-collector/tracer/bpf/process-collector.c
```diff
@@ -12,11 +12,6 @@
 #include <vmlinux/vmlinux.h>
 
 #include <bpf/bpf_helpers.h>
-/* libbpf v0.4.0 introduced BPF_SEQ_PRINTF in bpf_tracing.h.
- * In future versions, it will be in bpf_helpers.h.
- */
-#include <bpf/bpf_tracing.h>
```
alban, line 18: I tried `go generate ./...` on this branch, and I get:
```
/home/alban/go/src/github.com/kinvolk/inspektor-gadget/pkg/gadgets/process-collector/tracer/bpf/process-collector.c:37:2: error: implicit declaration of function 'BPF_SEQ_PRINTF' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        BPF_SEQ_PRINTF(seq, "%d %d %s %llu\n", task->tgid, task->pid, task->comm, mntns_id);
        ^
1 error generated.
Error: clang: exit status 1
```

It works when I reintroduce `bpf/bpf_tracing.h`.

```
$ rpm -qa|grep  libbpf
libbpf-0.4.0-1.fc34.x86_64
libbpf-devel-0.4.0-1.fc34.x86_64
```

Do we want to keep supporting libbpf 0.4.0?

Ditto in pkg/gadgets/socket-collector/tracer/bpf/tcp4-collector.c

pkg/gadgets/seccomp/tracer/tracer.go
```diff
@@ -23,7 +23,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-//go:generate sh -c "echo $CLANG_OS_FLAGS; GOOS=$(go env GOHOSTOS) GOARCH=$(go env GOHOSTARCH) go run github.com/cilium/ebpf/cmd/bpf2go -target bpfel -cc clang seccomp ./bpf/seccomp.c -- $CLANG_OS_FLAGS -I./bpf/ -I../../.. -target bpf -D__TARGET_ARCH_x86"
+//go:generate sh -c "echo $CLANG_OS_FLAGS; GOOS=$(go env GOHOSTOS) GOARCH=$(go env GOHOSTARCH) go run github.com/cilium/ebpf/cmd/bpf2go -target amd64 -cc clang seccomp ./bpf/seccomp.c -- $CLANG_OS_FLAGS -I./bpf/ -I../../.."
```
alban, line 26: Here we don't define `$CLANG_OS_FLAGS`, so that could be removed, right?

Did you forget to remove `$GOOS` and `$GOARCH` like in other gadgets?

Ditto in some other gadgets.

tools/btfgen.sh
```diff
@@ -21,24 +21,26 @@ fi
 # We do not need to add network eBPF programs like dns or snisnoop here because
 # they do not rely on BTF (i.e., they do not include vmlinux.h).
```
alban, line 22: snisnoop was somehow added back despite the comment.

To detect socket programs that don't use BTF:
```
for f in $(git ls-files |grep 'tracer/.*\.c$') ; do grep -q 'SEC("socket' $f && echo ${f} || true ; done
```

pkg/gadgets/process-collector/tracer/bpf/process-collector.c
mauriciovasquezbernal: > Do we want to keep supporting libbpf 0.4.0?

libbpf v0.4.0 was released more than a year ago (https://github.com/libbpf/libbpf/releases/tag/v0.4.0) and the latest release now is v0.8.0 (https://github.com/libbpf/libbpf/releases/tag/v0.8.0). IMO we could require a newer version. 

Also, libbpf is only needed for the bpf headers, there is https://github.com/cilium/ebpf/issues/449 that could even remove this dependency for us.

pkg/gadgets/seccomp/tracer/tracer.go
mauriciovasquezbernal: You're right, I was missing this one. Thanks!

pkg/gadgets/process-collector/tracer/bpf/process-collector.c
alban: I use Fedora 34 and it only has libbpf 0.4.0. Since this PR removes the ebpf build in a container, it means I have to update libbpf on my laptop. I guess it is time to update it :)

pkg/gadgets/seccomp/tracer/tracer.go
mauriciovasquezbernal: > Ditto in some other gadgets.

Do you have a specific one? I can only see that `$CLANG_OS_FLAGS` is used in dns and sni where it's actually needed.

tools/btfgen.sh
```diff
@@ -21,24 +21,26 @@ fi
...
-    -o ${INSPEKTOR_GADGET}/pkg/gadgets/capabilities/tracer/core/capabilities_bpfel.o                \
+    -o ${INSPEKTOR_GADGET}/pkg/gadgets/audit-seccomp/tracer/auditseccomp_bpfel_x86.o                \
+    -o ${INSPEKTOR_GADGET}/pkg/gadgets/audit-seccomp/tracer/auditseccompwithfilters_bpfel_x86.o     \
+    -o ${INSPEKTOR_GADGET}/pkg/gadgets/bindsnoop/tracer/core/bindsnoop_bpfel_x86.o                  \
+    -o ${INSPEKTOR_GADGET}/pkg/gadgets/biotop/tracer/biotop_bpfel_x86.o                             \
+    -o ${INSPEKTOR_GADGET}/pkg/gadgets/capabilities/tracer/core/capabilities_bpfel_x86.o            \
     -o ${INSPEKTOR_GADGET}/pkg/gadgets/execsnoop/tracer/core/execsnoop_bpfel.o                      \
-    -o ${INSPEKTOR_GADGET}/pkg/gadgets/filetop/tracer/filetop_bpfel.o                               \
-    -o ${INSPEKTOR_GADGET}/pkg/gadgets/fsslower/tracer/core/fsslower_bpfel.o                        \
+    -o ${INSPEKTOR_GADGET}/pkg/gadgets/filetop/tracer/filetop_bpfel_x86.o                           \
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > Could we include bpf_tracing.h in all bpf sources so that it is consistent?

Gadgets using [bpf-tracing.h](`https://github.com/libbpf/libbpf/blob/29869d6ef0893cc1c907705f919fdfd6feb7cc13/src/bpf_tracing.h`) need to be compiled for each different architecture. This PR proposes a solution where gadgets that don't use this file are compiled only two times: big and little endian machines. Finally, btfgen.sh is removed in https://github.com/kinvolk/inspektor-gadget/pull/710 to avoid having to add each single gadget there.

mauriciovasquezbernal:
I'll merge as integration tests are passing. Tests on ARO are failing but it seems unrelated to this PR as those are failing on most PRs now.

PR #695, opened 2022-06-15
cmd/kubectl-gadget: Avoid repeated code on snapshot gadgets

blanquicet:
# Avoid repeated code on snapshot gadgets

This PR factorizes the code of the snapshot gadgets to avoid repeating patterns on each of them. The common code was moved to a snapshot source code, and each gadget kept only its specific code.

## Implementation details

I used generics to simplify the code. It implies updating the required Go to v1.18.

## Notes for reviewers

- I tried to make progressive changes at each commit so that it is easier to do a review per commit.
- Please consider that the final goal of this series of PRs is to move most of the code to other packages and then use it from local-gadget.

### DIFF
--- a/cmd/kubectl-gadget/snapshot/process.go
+++ b/cmd/kubectl-gadget/snapshot/process.go
@@ -15,31 +15,28 @@
 package snapshot
 
 import (
-	"encoding/json"
 	"fmt"
-	"os"
 	"sort"
 	"strings"
-	"text/tabwriter"
 
 	"github.com/spf13/cobra"
 
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
-	gadgetv1alpha1 "github.com/kinvolk/inspektor-gadget/pkg/apis/gadget/v1alpha1"
 	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/process-collector/types"
-	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 )
 
 type ProcessFlags struct {
 	showThreads bool
 }
 
-func init() {
-	processCmd := initProcessCmd()
-	SnapshotCmd.AddCommand(processCmd)
+type ProcessParser struct {
+	BaseSnapshotParser
+
+	outputConfig *utils.OutputConfig
+	processFlags *ProcessFlags
 }
 
-func initProcessCmd() *cobra.Command {
+func newProcessCmd() *cobra.Command {
 	var commonFlags utils.CommonFlags
 	var processFlags ProcessFlags
 
@@ -47,69 +44,14 @@ func initProcessCmd() *cobra.Command {
 		Use:   "process",
 		Short: "Gather information about running processes",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			config := &utils.TraceConfig{
-				GadgetName:       "process-collector",
-				Operation:        "collect",
-				TraceOutputMode:  "Status",
-				TraceOutputState: "Completed",
-				CommonFlags:      &commonFlags,
+			processGadget := &SnapshotGadget[types.Event]{
+				name:        "process-collector",
+				commonFlags: &commonFlags,
+				parser:      NewProcessParser(&commonFlags.OutputConfig, &processFlags),
+				params:      nil,
 			}
 
-			callback := func(results []gadgetv1alpha1.Trace) error {
-				allEvents := []types.Event{}
-
-				for _, i := range results {
-					if len(i.Status.Output) == 0 {
-						continue
-					}
-
-					var events []types.Event
-					if err := json.Unmarshal([]byte(i.Status.Output), &events); err != nil {
-						return utils.WrapInErrUnmarshalOutput(err, i.Status.Output)
-					}
-					allEvents = append(allEvents, events...)
-				}
-
-				allEvents = sortProcessEvents(allEvents, &processFlags)
-
-				switch commonFlags.OutputMode {
-				case utils.OutputModeJSON:
-					b, err := json.MarshalIndent(allEvents, "", "  ")
-					if err != nil {
-						return utils.WrapInErrMarshalOutput(err)
-					}
-
-					fmt.Printf("%s\n", b)
-					return nil
-				case utils.OutputModeColumns:
-					fallthrough
-				case utils.OutputModeCustomColumns:
-					// In the snapshot gadgets it's possible to use a tabwriter because
-					// we have the full list of events to print available, hence the
-					// tablewriter is able to determine the columns width. In other
-					// gadgets we don't know the size of all columns "a priori", hence
-					// we have to do a best effort printing fixed-width columns.
-					w := tabwriter.NewWriter(os.Stdout, 0, 0, 4, ' ', 0)
-
-					fmt.Fprintln(w, getProcessColsHeader(&processFlags, commonFlags.CustomColumns))
-					for _, e := range allEvents {
-						if e.Type != eventtypes.NORMAL {
-							utils.ManageSpecialEvent(e.Event, commonFlags.Verbose)
-							continue
-						}
-
-						fmt.Fprintln(w, transformProcessEvent(&e, &processFlags, &commonFlags.OutputConfig))
-					}
-
-					w.Flush()
-				default:
-					return utils.WrapInErrOutputModeNotSupported(commonFlags.OutputMode)
-				}
-
-				return nil
-			}
-
-			return utils.RunTraceAndPrintStatusOutput(config, callback)
+			return processGadget.Run()
 		},
 	}
 
@@ -126,9 +68,7 @@ func initProcessCmd() *cobra.Command {
 	return cmd
 }
 
-// getProcessColsHeader returns a header with the default list of columns
-// when it is not requested to use a subset of custom columns.
-func getProcessColsHeader(processFlags *ProcessFlags, requestedCols []string) string {
+func NewProcessParser(outputConfig *utils.OutputConfig, processFlags *ProcessFlags) SnapshotParser[types.Event] {
 	availableCols := map[string]struct{}{
 		"node":      {},
 		"namespace": {},
@@ -139,24 +79,33 @@ func getProcessColsHeader(processFlags *ProcessFlags, requestedCols []string) st
 		"pid":       {},
 	}
 
+	return &ProcessParser{
+		BaseSnapshotParser: BaseSnapshotParser{
+			availableCols: availableCols,
+		},
+		outputConfig: outputConfig,
+		processFlags: processFlags,
+	}
+}
+
+func (p *ProcessParser) GetColumnsHeader() string {
+	requestedCols := p.outputConfig.CustomColumns
 	if len(requestedCols) == 0 {
 		requestedCols = []string{"node", "namespace", "pod", "container", "comm", "pid"}
-		if processFlags.showThreads {
+		if p.processFlags.showThreads {
 			requestedCols = []string{"node", "namespace", "pod", "container", "comm", "tgid", "pid"}
 		}
 	}
 
-	return buildSnapshotColsHeader(availableCols, requestedCols)
+	return p.BuildSnapshotColsHeader(requestedCols)
 }
 
-// transformProcessEvent is called to transform an event to columns
-// format according to the parameters.
-func transformProcessEvent(e *types.Event, processFlags *ProcessFlags, outputConf *utils.OutputConfig) string {
+func (p *ProcessParser) TransformEvent(e *types.Event) string {
 	var sb strings.Builder
 
-	switch outputConf.OutputMode {
+	switch p.outputConfig.OutputMode {
 	case utils.OutputModeColumns:
-		if processFlags.showThreads {
+		if p.processFlags.showThreads {
 			sb.WriteString(fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t%d\t%d",
 				e.Node, e.Namespace, e.Pod, e.Container,
 				e.Command, e.Tgid, e.Pid))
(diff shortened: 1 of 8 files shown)


blanquicet:
Related to https://github.com/kinvolk/inspektor-gadget/issues/545. It will be completely solved once we implement this refactoring for all the categories.

cmd/kubectl-gadget/snapshot/snapshot.go
```diff
@@ -15,16 +15,118 @@
...
+	// The Go compiler does not support accessing a struct field x.f where x is
+	// of type parameter type even if all types in the type parameter's type set
+	// have a field f. We may remove this restriction in Go 1.19. See
+	// https://tip.golang.org/doc/go1.18#generics.
+	GetBaseEvent() eventtypes.Event
+}
+
+type GadgetFlags interface {
+	SocketFlags | ProcessFlags
+}
```
blanquicet, line 50: Defining the [`processFlags`](https://github.com/kinvolk/inspektor-gadget/blob/e9cc720de5e12bf926bebc42a69417afe4959d31/cmd/kubectl-gadget/snapshot/process.go#L45) and [`socketFlags`](https://github.com/kinvolk/inspektor-gadget/blob/e9cc720de5e12bf926bebc42a69417afe4959d31/cmd/kubectl-gadget/snapshot/socket.go#L46) as global variables would remove the need of having this interface. On the one hand, I would prefer not to use global variables unless it is strictly needed. However, I am aware that this interface makes the code more complex. What do you think?

cmd/kubectl-gadget/snapshot/snapshot.go
```diff
@@ -15,16 +15,118 @@
...
+
+// getSnapshotCallback returns the callback that will be called when a snapshot
+// gadget finishes without errors and generates a list of results per node. This
+// function merges, sorts and print all of them in the requested mode.
+func getSnapshotCallback[Event GadgetEvent, Flags GadgetFlags](
+	outputConf *utils.OutputConfig,
+	gadgetFlags *Flags,
+	sortEvents func([]Event, *Flags),
+	getColsHeader func(*Flags, []string) string,
+	transformEvent func(*Event, *Flags, *utils.OutputConfig) string,
```
blanquicet, line 60: I am investigating how to create an `interface` with these methods to avoid passing all these arguments but I am not sure it is possible using generics.

eiffel-fl (commented):
Hi.

I like the idea, code is nicer now so thank you for it!
I nonetheless have some comments (particularly regarding `-t`):

Best regards.

.github/workflows/inspektor-gadget.yml
```diff
@@ -17,12 +17,12 @@ jobs:
     # level: 0
     runs-on: ubuntu-latest
     steps:
-    - name: Setup go 1.17
+    - name: Setup Go
       uses: actions/setup-go@v1
       with:
-        go-version: 1.17
+        go-version: 1.18
```
alban, line 23: This breaks "go get" in crd.mk because of the change of semantics in Go 1.18:
https://go.dev/doc/go-get-install-deprecation

It seems to work on the upstream repository but it does not work on a fork:
https://github.com/alban/inspektor-gadget/runs/6900841081?check_suite_focus=true

```
/bin/sh: 1: /home/runner/work/inspektor-gadget/inspektor-gadget/bin/controller-gen: not found
```

I am not sure if it is because something is cached...

I am trying to fix this with something like [this patch](https://github.com/alban/inspektor-gadget/commit/4ea50d19c936466f98e7295e902c4e3ecd1213e9) but there are other errors behind. I am investigating it...

alban: The CI tests are fixed with the following commits:
https://github.com/alban/inspektor-gadget/commit/45f31810d4dfcdb120bcc95a741c2b529f247135
https://github.com/alban/inspektor-gadget/commit/b74cdff3eb683499ef79b67c64f865b0f1dc5392

.github/workflows/inspektor-gadget.yml
```diff
@@ -17,12 +17,12 @@ jobs:
...
       id: go
-    - name: Cache go 1.17
+    - name: Cache Go
```
alban, line 25: The new version of `actions/setup-go` can manage the Go cache automatically, so we don't need this section anymore. 
This requires to move the `checkout` above the `setup-go`:
https://github.com/alban/inspektor-gadget/commit/2126f17d80f77eeb1628e844eee592fa3b932191

cmd/kubectl-gadget/snapshot/process.go
```diff
@@ -128,32 +166,32 @@ func processTransformEvent(e types.Event) string {
...
+			sb.WriteRune('\t')
 		}
 	}
 
 	return sb.String()
 }
 
-func printProcesses(allProcesses []types.Event) error {
-	if !processCollectorParamThreads {
+func sortProcessEvents(allProcesses []types.Event, processFlags *ProcessFlags) {
```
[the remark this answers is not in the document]
blanquicet: Oops. You are right! I didn't realise deep enough that functionality. I followed your suggestion about returning the new slice, but I was wondering if it would be better to use pointers 🤔

cmd/kubectl-gadget/snapshot/snapshot.go
blanquicet: I found a way to use interfaces, and it improved the code!

cmd/kubectl-gadget/utils/flags.go
```diff
@@ -78,14 +90,8 @@ type CommonFlags struct {
...
-	// OutputMode specifies the way output should be printed
-	OutputMode string
-
-	// Verbose prints additional information
-	Verbose bool
-
-	// List of columns to print (only meaningful when OutputMode is "columns=...")
-	CustomColumns []string
+	// OutputConf specifies the way output should be printed
+	OutputConf OutputConfig
```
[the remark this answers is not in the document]
blanquicet: I agree it doesn't look that good, but I am not sure if embedding it is the best solution. AFAIU, embedding should be used when the inner struct is like a "base version" for the outer (e.g. [here](https://github.com/kinvolk/inspektor-gadget/blob/a7237570b55aa7d20e067c41a992615436370f33/pkg/gadgets/execsnoop/types/types.go#L22)). And it does not seem to be the case, don't you think?

Anyway, I modified the code to use an interface, see https://github.com/kinvolk/inspektor-gadget/pull/695#discussion_r898087131. Let me know if it looks better now.

.github/workflows/inspektor-gadget.yml
blanquicet: It is being managed in https://github.com/kinvolk/inspektor-gadget/pull/696.

cmd/kubectl-gadget/snapshot/snapshot.go
blanquicet: Using an interface resolved this situation.

cmd/kubectl-gadget/snapshot/process.go
```diff
@@ -15,110 +15,97 @@
...
 
-	processCollectorCmd.PersistentFlags().BoolVarP(
-		&processCollectorParamThreads,
+	cmd.PersistentFlags().BoolVarP(
+		&processFlags.threadGroupID,
 		"threads",
 		"t",
 		false,
-		"Show all threads",
+		"Show the thread group ID",
```
alban, line 73: I think this doc change is confusing and the previous text ("Show all threads") was better.

When this flag is set, it will show an entry for each thread. Without this flag, only one entry per process. In multi-threaded processes, there is typically one main thread that has pid == tid.

I guess you renamed the flag to `threadGroupID` it seems to change whether the tgid is printed. But its main change is to filter out the threads (see code near `allProcessesTrimmed`).

Also, "thread group ID" is the term given in kernel code (or BPF code since it runs in the kernel). But users would normally call it "process ID" or pid:

|  Concept       | In kernel code       | In userspace code |
|---------|----------------------|-------------------|
| Process | tgid (task group id) | pid (from `getpid()`)               |
| Thread  | pid                  | tid (from `gettid()`)               |

alban:
> Please consider that the final goal of this series of PRs is to move most of the code to other packages and then use it from local-gadget

Could you give examples of generic functions that would be added in a shared package between kubectl-gadget and local-gadgets? Would getSnapshotCallback() be one of them?

cmd/kubectl-gadget/snapshot/process.go
blanquicet: You are right. I got confused (again 😢 ). Thanks for the explanation!

I think my confusion comes from the name of the columns because they describe the data in "kernel terms" and not "userspace terms". Let me describe it with an example. 

When `-t` is not specified, we have the following output:

```bash
$ ./kubectl-gadget snapshot process -n gadget -p gadget-fgkll
NODE           NAMESPACE    POD             CONTAINER    COMM               PID
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559
```

It shows one process with ID `10559` (PID in terms of kernel and userspace). So far, so good. 

Now, if I want to know if that process has more than one thread, I use the `-t` option:

```bash
$ ./kubectl-gadget snapshot process -n gadget -p gadget-fgkll -t
NODE           NAMESPACE    POD             CONTAINER    COMM               TGID     PID
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10559
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10586
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10587
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10588
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10589
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10590
vm-minikube    gadget       gadget-fgkll    gadget       gadgettracerman    10559    10591
...
```

Here is where I don't think the name of the columns describes very well the situation in terms of userspace. I mean, I read it as the column PID now represents the thread ID and the TGID, which is kernel terminology, represents the PID. Don't you think?

IMO, after thinking about it better, the description should remain "Show all threads" as you said. Then, if our idea is to show the information in terms of userspace, I think we should change the name of the last two columns to be PID and TID (thread ID):

```bash
$ ./kubectl-gadget snapshot process -n gadget -p gadget-fgkll -t
NODE           NAMESPACE    POD             CONTAINER    COMM               PID     TID
```

What do you think?

BTW, I found [this](https://stackoverflow.com/a/9306150) comment very useful.

blanquicet:
> Could you give examples of generic functions that would be added in a shared package between kubectl-gadget and local-gadgets? Would getSnapshotCallback() be one of them?

My first idea is to move all the functions within `socket.go` to a shared package and have this code in the CLIs:

kubectl-gadget:
```go
func init() {
	socketCmd := initSocketCmd(utils.RunTraceAndPrintStatusOutput)
	SnapshotCmd.AddCommand(socketCmd)
}
```

local-gadget:
```go
func init() {
	socketCmd := initSocketCmd(<function that runs the gadget and returns the results (still to be developed)>)
	SnapshotCmd.AddCommand(socketCmd)
}
```

The same applies to `process.go`. While for `snapshot.go`, all the code within can also be moved to a shared package except for the definition of `SnapshotCmd`.

What do you think? I think we can still improve it, but we will see once we start developing the local-gadget side. I decided not to do everything in one PR because it would have been very difficult to review.

cmd/kubectl-gadget/snapshot/process.go
alban: I agree, the columns should use the userspace terminology

eiffel-fl: I agree too, but this can be dealt in another PR.

blanquicet: I drop that change from this PR, and I will change the columns header in another PR.

cmd/kubectl-gadget/snapshot/process.go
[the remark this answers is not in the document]
eiffel-fl: > I followed your suggestion about returning the new slice, but I was wondering if it would be better to use pointers thinking

Yes sorry, I suggested the first thing which came to my mind to solve it but indeed using pointers would be better as it would avoid allocating a new slice.

cmd/kubectl-gadget/utils/flags.go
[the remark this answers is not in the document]
blanquicet: TBH, I don't like `params.OutputConfig.Verbose`, so I would be more for using embedding. Let's first see what other folks think about it.

cc @alban @mauriciovasquezbernal

cmd/kubectl-gadget/snapshot/process.go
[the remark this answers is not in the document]
blanquicet: > it would avoid allocating a new slice
>
We still need to do it because it is easier to create a new slice where we only add the main thread per process than to remove the secondary threads from the original slice.

My only concern is that a pointer to a slice (`*[]types.Event`) doesn't look very go-style 😅 .

cmd/kubectl-gadget/utils/flags.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think embedding is fine.

blanquicet:
Rebased

blanquicet:
Related https://github.com/kinvolk/inspektor-gadget/issues/642

blanquicet:
I created the [`SnapshotGadget` struct](https://github.com/kinvolk/inspektor-gadget/blob/d665c09aa53a505eeaf75e5561867ed57572a17e/cmd/kubectl-gadget/snapshot/snapshot.go#L77-L83) and used composition for the [parsers](https://github.com/kinvolk/inspektor-gadget/blob/d665c09aa53a505eeaf75e5561867ed57572a17e/cmd/kubectl-gadget/snapshot/process.go#L33) to make the code more readable as suggested by @mauriciovasquezbernal. Therefore, PTAL.

cmd/kubectl-gadget/snapshot/snapshot.go
```diff
@@ -15,16 +15,158 @@
...
+		}
+
+		return nil
+	}
+
+	return utils.RunTraceAndPrintStatusOutput(config, callback)
+}
+
+func NewSnapshotCmd() *cobra.Command {
+	SnapshotCmd := &cobra.Command{
```
mauriciovasquezbernal, line 163: I think local variables as usually lowercase, right? (I got confused for a second thinking it was a global variable).

mauriciovasquezbernal (approved):
Just a very minor nit. Looks great, thanks for handling this!

PR #696, opened 2022-06-15
go: Update to version 1.18

alban:
# go: Update to version 1.18

Go 1.18 is useful because:
- it introduces generics (see #695)
- we depends on cilium/ebpf which requires Go 1.18

This PR updates to Go 1.18, update the GitHub Action with the cache feature from `setup-go` and add fixes required by Go 1.18.

## How to use

No changes.

## Testing done

None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,23 +1,21 @@
 module github.com/kinvolk/inspektor-gadget
 
+go 1.18
+
 require (
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/cilium/ebpf v0.8.2-0.20220502122259-96aa1a7a929f
-	github.com/containerd/containerd v1.5.11 // indirect
 	github.com/containerd/nri v0.1.1-0.20210619071632-28f76457b672
 	github.com/containers/common v0.46.0
-	github.com/docker/distribution v2.8.0+incompatible // indirect
 	github.com/docker/docker v20.10.8+incompatible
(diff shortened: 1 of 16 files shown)


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/docker/docker v20.10.8+incompatible
```
blanquicet, line 10: go.mod still needs to be updated to require 1.18. 

BTW, in the go.mod we cannot specify 1.18.3 as in the CI, so I was wondering if we should change the CI also to use 1.18 instead of 1.18.3. WDYT?

alban: > go.mod still needs to be updated to require 1.18.

Oh, I've broken my rebase. Let me fix this.

> BTW, in the go.mod we cannot specify 1.18.3 as in the CI, so I was wondering if we should change the CI also to use 1.18 instead of 1.18.3. WDYT?

I think we should not: the go.mod version is for backward compatibility (they don't change the semantics between minor versions) but we can still take the last Go version.

blanquicet (approved):
LGTM, just please update the installation requirements [here](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/install.md#compile-from-source).

alban:
> LGTM, just please update the installation requirements [here](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/install.md#compile-from-source).

Thanks. Updated. I will merge when the CI is green.

PR #698, opened 2022-06-16
go: Update to version 1.18 in Dockerfiles

alban:
The Dockerfiles were forgotten in the previous update to Go 1.18

Follow-up to https://github.com/kinvolk/inspektor-gadget/pull/696

### DIFF
--- a/gadget-core.Dockerfile
+++ b/gadget-core.Dockerfile
@@ -22,8 +22,8 @@ RUN set -ex; \
 	add-apt-repository -y ppa:tuxinvader/kernel-build-tools && \
 	apt-add-repository -y ppa:longsleep/golang-backports && \
 	apt-get update && \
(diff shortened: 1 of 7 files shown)


blanquicet (approved):
LGTM. Should we update the integration-tests Docker file?

alban:
integration-tests Dockerfile updated.

@blanquicet Could you please check the errors on ARO? I try to re-run the job but it fails with the same error:

> Warning: Detected changes to resource security-profiles-operator which is currently being deleted.
> Error from server (Forbidden): error when creating "STDIN": serviceaccounts "security-profiles-operator" is forbidden: unable to create new content in namespace security-profiles-operator because it is being terminated

blanquicet (approved):
New changes LGTM!

PR #703, opened 2022-06-21
Adds support for multi containers pod for seccomp-profile.

### LINKED ISSUE #215: [RFE] Merge seccomp profiles for pods with several containers
## Current situation

With the seccomp gadget (#208), users can specify a container and generate a seccomp policy based on the syscalls used by the container. But only one container can be selected at a time.

## Impact

Users cannot easily generate a seccomp policy for pods containing several containers: they would have to generate a seccomp policy for each container individually, and merge the seccomp policies manually.

## Ideal future situation

Ideally, users can select a pod and the gadget would automatically merge the seccomp policies of each individual container into one seccomp policy for the pod.

eiffel-fl:
Hi.

With this PR, I enabled `seccomp-profile` to work on pods with several containers inside.
The idea is to generate a seccomp profile which contains the union of all syscalls used by containers within this pod.
The union works like this:
Container X: [syscallA] = 0, [syscallB] = 1, [syscallC] = 0, [syscallD] = 0
Container Y: [syscallA] = 1, [syscallB] = 0, [syscallC] = 1, [syscallD] = 0
Union:          [syscallA] = 1, [syscallB] = 1, [syscallC] = 1, [syscallD] = 0

I check the above behavior running a pod with two containers (nginx and mariadb) and running them separately in dedicated pods:

```bash
# test contains two containers
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget advise seccomp-profile start -p test        
YgbVs4UQgmpUk0ar
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget advise seccomp-profile stop YgbVs4UQgmpUk0ar
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    {
      "names": [
        "futex"
      ],
      "action": "SCMP_ACT_ALLOW"
    }
  ]
}
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget advise seccomp-profile start -p nginx       
aSoqfaTyjoCF3XX1
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget advise seccomp-profile stop aSoqfaTyjoCF3XX1
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    {
      "names": [],
      "action": "SCMP_ACT_ALLOW"
    }
  ]
}
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget advise seccomp-profile start -p mariadb     
4IYtOXK53ROoZCrv
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget advise seccomp-profile stop 4IYtOXK53ROoZCrv
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    {
      "names": [
        "futex"
      ],
      "action": "SCMP_ACT_ALLOW"
    }
  ]
}
```

It fixes #215.

Best regards.

### DIFF
--- a/pkg/gadgets/seccomp/gadget.go
+++ b/pkg/gadgets/seccomp/gadget.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -60,6 +59,9 @@ type TraceSingleton struct {
 }
 
 var traceSingleton TraceSingleton
+
+// Must be equal to C.SYSCALLS_COUNT
+const syscallsCount = 500
 
 func NewFactory() gadgets.TraceFactory {
 	return &TraceFactory{
@@ -420,6 +422,31 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
 	trace.Status.State = "Started"
 }
 
+// mergeSyscallArrays merge two syscalls array into first argument.
+// After this function, the first argument will be the union of present
+// syscalls of each arrays.
+// For example, the merge result of:
+// syscalls: [syscallA] = 0, [syscallB] = 1, [syscallC] = 0, [syscallD] = 0
+// b:        [syscallA] = 1, [syscallB] = 0, [syscallC] = 1, [syscallD] = 0
+// result:   [syscallA] = 1, [syscallB] = 1, [syscallC] = 1, [syscallD] = 0
+func mergeSyscallArrays(syscalls []byte, b []byte) error {
+	if len(b) != len(syscalls) {
+		return fmt.Errorf("syscalls array lenghts mistmatch (%d != %d), it should be %d",
+			len(b), len(syscalls), syscallsCount,
+		)
+	}
+
+	for syscallNumber, isPresent := range b {
+		if isPresent != 1 {
+			continue
+		}
+
+		syscalls[syscallNumber] = isPresent
+	}
+
+	return nil
+}
+
 func (t *Trace) Generate(trace *gadgetv1alpha1.Trace) {
 	if traceSingleton.tracer == nil {
 		log.Errorf("Seccomp tracer is nil")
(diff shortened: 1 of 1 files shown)


alban:
With pods containing several containers, it's possible to add a different seccomp profile for each container of the pod. Shouldn't we recommend that instead?

X-Ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#seccompprofile-v1-core
There is a `SeccompProfile` field both in:
* PodSecurityContext (seccomp profile at the pod level)
* SecurityContext (seccomp profile at the container level)

blanquicet:
> > With pods containing several containers, it's possible to add a different seccomp profile for each container of the pod. Shouldn't we recommend that instead?
> 
> This seems to be a good idea! So, instead, trace output would contain as many JSON as containers, right?

Yeah, I think it is a good idea. The problem is how to distinguish which container the seccomp profile belongs to. Maybe we could retry a map where the key is the name of the container and the value is the seccomp profile? I don't see any other field in [the struct](https://pkg.go.dev/github.com/opencontainers/runtime-spec/specs-go#LinuxSeccomp) where we can specify the container name.

mauriciovasquezbernal:
What's the status of this PR? Is there anything we should do to move this forward or should we close it?

mauriciovasquezbernal:
Right, I also agree we can close this. Let's ask @alban since he has more context on this.

mauriciovasquezbernal:
Closing this as we don't want this behavior as stated in https://github.com/inspektor-gadget/inspektor-gadget/issues/215.

PR #713, opened 2022-06-27
cmd/kubectl-gadget: Avoid repeated code on trace gadgets

blanquicet:
# Avoid repeated code on trace gadgets

This PR continues the series of PRs to factorize the code of the different categories so that we avoid repeating patterns on all the gadgets within the same category. Please take into account that the final goal of this series of PRs is to move most of the code to specific packages and then use it from kubectl-gadget and local-gadget.

Related https://github.com/kinvolk/inspektor-gadget/pull/695 and https://github.com/kinvolk/inspektor-gadget/issues/642


cmd/kubectl-gadget/trace/trace.go
```diff
@@ -15,16 +15,148 @@
...
+
+	if err := utils.RunTraceAndPrintStream(config, transformEvent); err != nil {
+		return utils.WrapInErrRunGadget(err)
+	}
+
+	return nil
+}
+
+func NewTraceCmd() *cobra.Command {
+	TraceCmd := &cobra.Command{
```
mauriciovasquezbernal, line 143: same nit about case as in the other PR.

integration/integration_test.go
```diff
@@ -468,7 +468,7 @@ func TestExecsnoop(t *testing.T) {
 	execsnoopCmd := &command{
 		name:           "StartExecsnoopGadget",
 		cmd:            fmt.Sprintf("$KUBECTL_GADGET trace exec -n %s", ns),
-		expectedRegexp: fmt.Sprintf(`%s\s+test-pod\s+test-pod\s+date`, ns),
+		expectedRegexp: fmt.Sprintf(`%s\s+test-pod\s+test-pod\s+\d+\s+\d+\s+date`, ns),
```
mauriciovasquezbernal, line 471: Could you please describe why these changes are needed? Do we need to update some documents as well?

cmd/kubectl-gadget/trace/mount.go
```diff
@@ -15,52 +15,92 @@
...
+				"op",
+				"ret",
+				"lat",
+				"fs",
+				"src",
+				"target",
+				"data",
+				"call",
+				"flags",
+			},
```
mauriciovasquezbernal, line 54: I think it's too much for the default arguments. Any reason no to keep using the same as before? https://github.com/kinvolk/inspektor-gadget/blob/08d49ba693a7c734993afe32f0ee7af797a83682/cmd/kubectl-gadget/trace/mount.go#L39-L41

mauriciovasquezbernal (commented):
Looks very good overall. I'm wondering why the default output of some gadgets was changed.

cmd/kubectl-gadget/trace/mount.go
blanquicet: Oops, my fault. I added the full list of fields and I didn't take into account that not all of them were printed. I will update the PR to continue printing the same list of columns. 

The other changes in the integration-tests was because I changed the order of the columns in some gadgets because I was trying to keep the same order in all of them: node, namespace, pod, container, pid, comm (if available) and the rest.

cmd/kubectl-gadget/trace/trace.go
```diff
@@ -15,16 +15,148 @@
...
+	}
+
+	fmt.Println(sb.String())
+}
+
+type TraceEvent interface {
+	bindtypes.Event | capabilitiestypes.Event | dnstypes.Event |
+		exectypes.Event | fsslowertypes.Event | mounttypes.Event |
+		oomkilltypes.Event | opentypes.Event | signaltypes.Event |
+		snitypes.Event | tcptypes.Event | tcpconnecttypes.Event
```
mauriciovasquezbernal, line 68: I was thinking that we could use `any` instead of explicitly listing all types here, however I'm not totally sure about the implications. Do you have any opinion on this?  /cc @alban

mauriciovasquezbernal (approved):
It looks great. I like that now the gadgets are more uniform (common columns are printed in the same order using the same width). I only have a comment, but it's not so important hence LGTM. 

Thinking about the future, it seems to me we can still move some more code to common places for common columns. We can create an issue for that once it's merged.

cmd/kubectl-gadget/trace/trace.go
blanquicet: I know, it doesn't look great. 

I would say that the only advantage of listing them is that we are type-safe. However, I don't expect people to use `TraceEvent` directly but it is something we will only use internally so `any` would be ok as well. I think it is more about how it looks 😄 .

blanquicet: Updated with `any`.

mauriciovasquezbernal:
Still LGTM. But I didn't try it out.

PR #728, opened 2022-07-01
Use current user for linting container

eiffel-fl:
Hi.

This PR changes the container user from root to current user.
It avoids problem where modified files belong to root after linting.

Best regards.

### DIFF
--- a/tools/check-readme.pl
+++ b/tools/check-readme.pl
@@ -26,7 +26,7 @@ $kubectl_gadget_path = shift or die "Usage: $0 relative_path_to_kubectl_gadget p
 $readme_path = shift or die "Usage: $0 path_to_kubectl_gadget path_to_README";
 
 if (substr($kubectl_gadget_path, 0, 1) ne '.') {
-	die "kubectl-gadget must be absolute path: ${kubectl_gadget_path}";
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -168,7 +168,7 @@ generate-documentation:
 
 lint:
 	# This version number must be kept in sync with CI workflow lint one.
-	docker run --rm -v $(shell pwd):/app -w /app golangci/golangci-lint:v1.46.2 golangci-lint run --fix
+	docker run --rm --user $(shell id -u):$(shell id -g) -v $(shell pwd):/app -w /app golangci/golangci-lint:v1.46.2 golangci-lint run --fix
```
blanquicet, line 171: This change is needed indeed because the fixed files also change their owner to `root`. However, in my case, after this change it is not working:

```bash
$ make lint
# This version number must be kept in sync with CI workflow lint one.
docker run --rm --user 1000:1000 -v /home/jose/Documents/kinvolk/src/inspektor-gadget:/app -w /app golangci/golangci-lint:v1.46.2 golangci-lint run --fix
2022/07/05 15:05:26 failed to initialize build cache at /.cache/golangci-lint: mkdir /.cache: permission denied
make: *** [Makefile:171: lint] Error 1
```

Is there something wrong in my env?

PR #729, opened 2022-07-04
Remove signal KILL when used in signal handler.

eiffel-fl:
Hi.

This PR removes SIGKILL from handler as it cannot be trapped (this was pointed by static checker).

Best regards.


blanquicet (approved):
From code inspection and given that CI is passing, LGTM.

PR #730, opened 2022-07-05
local-gadget: Support running snapshot-process in a single call

blanquicet:
# Support running snapshot-process in a single call

This is a PoC where the goal is to discuss how to implement the support of running gadgets with local-gadget in a single call. Like this:

```bash
$ sudo local-gadget snapshot process
```

I started implementing the snapshot-process gadget to discuss the general approach. Then, once we agree and https://github.com/kinvolk/inspektor-gadget/pull/713 is merged, we can work on the trace category.

## Implementation details

- The idea was to reuse all the code already implemented for `kubectl-gadget`. Including specific gadget flags.
- I created the `cmd/common` folder where I am moving all the code that is shared between `local-gadget` and `kubectl-gadget`.
- The only differences between the implementation of gadgets on `local-gadget` and `kubectl-gadget` are:
  - The columns to be printed (`availableColumns` and the default value for `CustomColumns`). For now, for all the gadgets that we will support on `local-gadget`, I would suggest to removed the node, namespace and pod columns. Please notice they will be added in the future when we will enrich the containers from the runtimes also with this information, see more details at https://github.com/kinvolk/inspektor-gadget/issues/644. 
  - The `CommonFlags`, `local-gadget` and `kubectl-gadget` has flags that do not make sense for the other, e.g. `--runtimes` in `local-gadget`.
  - The function that actually runs the gadgets: `utils.RunTraceAndPrintStatusOutput`.
- I kept the interactive mode using the command `interactive`. However, I don't know how much sense it has to keep this mode once we support other gadgets in this single-call mode.
- I didn't enable the socket gadget because the output without the node, namespace and pod doesn't have sense. It is visiting the same network namespace multiple times. We have to either include container information (if possible) or ensure we visit the same network namespace only once. The discussion was already initiated here: https://github.com/kinvolk/inspektor-gadget/pull/245#discussion_r713242794. See #744.

I tried to make progressive changes so that it can be reviewed by commit.

This PR is part of the effort for EPIC https://github.com/kinvolk/inspektor-gadget/issues/643. In particular, task https://github.com/kinvolk/inspektor-gadget/issues/638.

## How to use

```bash
$ sudo local-gadget list-containers --runtimes containerd
calico-kube-controllers
calico-node
coredns
coredns
etcd
gadget
kube-apiserver
kube-controller-manager
kube-proxy
kube-scheduler

$ sudo local-gadget snapshot process --runtimes containerd
CONTAINER                  COMM               PID
calico-kube-controllers    kube-controller    4942
calico-node                bird               4062
calico-node                bird6              4063
calico-node                calico-node        3801
calico-node                calico-node        3803
calico-node                calico-node        3804
calico-node                calico-node        3807
calico-node                calico-node        3808
calico-node                calico-node        3809
calico-node                runsv              3793
calico-node                runsv              3794
calico-node                runsv              3795
calico-node                runsv              3796
calico-node                runsv              3797
calico-node                runsv              3798
calico-node                runsv              3799
calico-node                runsv              3800
calico-node                runsvdir           3697
coredns                    coredns            4938
coredns                    coredns            5024
etcd                       etcd               1608
gadget                     gadgettracerman    3320
kube-apiserver             kube-apiserver     1605
kube-controller-manager    kube-controller    1624
kube-proxy                 kube-proxy         2972
kube-scheduler             kube-scheduler     1641

# All the flags already implemented in kubectl-gadget are also supported in local-gadget
$ sudo local-gadget snapshot process -c gadget --runtimes containerd -t
CONTAINER    COMM               TGID    PID
gadget       gadgettracerman    3320    3320
gadget       gadgettracerman    3320    3342
gadget       gadgettracerman    3320    3343
gadget       gadgettracerman    3320    3344
gadget       gadgettracerman    3320    3345
gadget       gadgettracerman    3320    3346
gadget       gadgettracerman    3320    3347
gadget       gadgettracerman    3320    3348
gadget       gadgettracerman    3320    3349
gadget       gadgettracerman    3320    3350
gadget       gadgettracerman    3320    3622
```


cmd/common/utils/trace.go
```diff
@@ -0,0 +1,52 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+// BaseTraceConfig is used to contain information used to manage a trace.
+type BaseTraceConfig struct {
```
mauriciovasquezbernal, line 18: I think this is too related to the CRD. I know local-gadget is reuses many of the logic around the trace, but I think this is something that we'll reduce in the future and this PR is a good opportunity not to include more logic related to it here.

My specific point is: I think this `BaseTraceConfig` shouldn't be the base for the trace config used in kubectl-gadget. It's possible that they share some info, but I'd like to have them as separated as possible.

mauriciovasquezbernal (commented):
I have some worries regarding the `gadgetv1alpha1.Trace` type being so heavily used in local-gadget. I know we can't remove it completely now, but can we avoid this as much as possible:

- Can we avoid changes in [cmd/common: Create common BaseTraceConfig struct](https://github.com/kinvolk/inspektor-gadget/pull/730/commits/bbf59e9f006574e45df5687ce8f8c859578ea72e).
- Code in cmd/common should not depend on this type (if possible).

cmd/kubectl-gadget/snapshot/process.go
```diff
@@ -79,96 +52,19 @@ func NewProcessParser(outputConfig *utils.OutputConfig, processFlags *ProcessFla
...
-			}
-			sb.WriteRune('\t')
-		}
+		return utils.RunTraceAndPrintStatusOutput(config, callback)
 	}
 
-	return sb.String()
-}
+	// It is not currently useful to pass processFlags here but it keeps
+	// uniformity with other gadgets.
```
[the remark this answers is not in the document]
blanquicet: I meant that `processFlags` could be directly defined inside `NewCommonProcessCmd` because it is not used in  [`newProcessCmd`](https://github.com/kinvolk/inspektor-gadget/blob/25fbe811b2f4f5811f10c2bb4094067f0fa5eb7e/cmd/kubectl-gadget/snapshot/process.go#L57) as socket-collector [does](https://github.com/kinvolk/inspektor-gadget/blob/25fbe811b2f4f5811f10c2bb4094067f0fa5eb7e/cmd/kubectl-gadget/snapshot/socket.go#L59).

cmd/common/utils/trace.go
blanquicet: You are right. Given that we are planning to make `local-gadget` independent from the CRD, it is better to keep that code separated so that when we introduce such a change, there will be no need to modify `kubectl-gadget` or `common` packages but only `local-gadget` itself.

blanquicet:
I addressed all the comments and rebased the PR on top of https://github.com/kinvolk/inspektor-gadget/pull/742.

I introduced one change with respect to the initial PR: I added the command `interactive` to run the `local-gadget` with the already existing interactive mode because it was quite strange to automatically enter the interactive mode when the user does not specify any command.

cc @alban

cmd/local-gadget/containers/containers.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+
+	"github.com/kinvolk/inspektor-gadget/cmd/local-gadget/utils"
+	localgadgetmanager "github.com/kinvolk/inspektor-gadget/pkg/local-gadget-manager"
+	"github.com/spf13/cobra"
+)
+
+func NewListContainers() *cobra.Command {
```
mauriciovasquezbernal, line 25: nit: `NewListContainersCmd`?

mauriciovasquezbernal (approved):
I tested it and LGTM. However I'd like to get @alban opinion on this one too.

blanquicet:
Rebased to solve conflicts created after merging https://github.com/kinvolk/inspektor-gadget/pull/746.

alban:
I think it goes in the right direction. Thanks!

I am a bit worried we need to add a new file for each gadget to support in local-gadget instead of reusing the common code. I checked the difference between the two files:

```
$ diff -u  cmd/kubectl-gadget/snapshot/process.go cmd/local-gadget/snapshot/process.go
...
-		"node":      {},
-		"namespace": {},
-		"pod":       {},
```

Could we have something that automatically removes node, namespace and pod for local-gadget? In this way, we would not need to duplicate this file for each gadget.

> I added the command interactive to run the local-gadget with the already existing interactive mode

docs/local-gadget.md needs to be updated to use this "interactive" command.

Btw, the command "create biolatency trace1" from the documentation (docs/local-gadget.md) does not work anymore. I am not sure if it biolatency was removed from pkg/gadget-collection/gadget-collection.go or if it was never added.

blanquicet:
> Could we have something that automatically removes node, namespace and pod for local-gadget? In this way, we would not need to duplicate this file for each gadget.
>

I think you missed another difference:
```bash
-       "github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
+       "github.com/kinvolk/inspektor-gadget/cmd/local-gadget/utils"
```

It makes that most of the used functions and structures come from two different packages even though they use the same name: `utils.CommonFlags`, `utils.RunTraceAndPrintStatusOutput` and `utils.AddCommonFlags`. Therefore, I didn't find a way to move the remaining part to the common code without making it unreadable. Any suggestion?

> docs/local-gadget.md needs to be updated to use this "interactive" command.
>
Oops, you are right. I missed that.

> Btw, the command "create biolatency trace1" from the documentation (docs/local-gadget.md) does not work anymore. I am not sure if it biolatency was removed from pkg/gadget-collection/gadget-collection.go or if it was never added.
>
It was never added. I will update the documentation in another PR to not iterate on top of this one that was already reviewed and avoid creating conflicts with other PRs.

blanquicet:
> ```
> $ diff -u  cmd/kubectl-gadget/snapshot/process.go cmd/local-gadget/snapshot/process.go
> ...
> -		"node":      {},
> -		"namespace": {},
> -		"pod":       {},
> ```
> 
> Could we have something that automatically removes node, namespace and pod for local-gadget? In this way, we would not need to duplicate this file for each gadget.
>
Regarding this, also Mauricio commented that we could have some basic columns, mainly the k8s metadata, and then each `kubectl-gadget` gadget extends it with its own list of columns. On the other hand, the `local-gadget` gadgets just need to use the gadget list without the basics ones.

That sounds good, but I'm not sure if a gadget will always use the same columns in `kubectl-gadget` and `local-gadget`. I don't know, I would like to import more gadgets to `local-gadget`, add the `local-gadget` specific fields and then decide how to manage this repeated code. I want to avoid implementing something that we will change/remove in a near future 🙂 .

PR #736, opened 2022-07-06
local-gadget: static build in container

alban:
Build local-gadget statically so it can be copied to Minikube.

I renamed the Dockerfile because I reuse the same for different
architectures.

Tested with:
```
make local-gadget
minikube cp ./local-gadget /bin/local-gadget
minikube ssh
sudo chmod +x /bin/local-gadget
sudo local-gadget
```

Depends on: https://github.com/kinvolk/inspektor-gadget/pull/441


alban:
I rebased. The CI failure does not seem related to my patch.

PR #739, opened 2022-07-07
ci: Use a matrix to build container image.

eiffel-fl:
Hi.

This PR replaces the two jobs we have to create docker images by a single one with a matrix.

Best regards.


alban (commented):
One question, LGTM after that

alban (approved):
LGTM

PR #740, opened 2022-07-08
ci: Speed up arm64 build

eiffel-fl:
Hi.

In this PR, I used cross compilation to reduce the time taken to cross-build the arm64 images.
Basically, all the steps in the builder stage will occur on the build platform (amd64 in our case) while the other stage will be run for the target platform (so emulated in our case).
It permitted to decrease the build time from around 35 minutes to approximately 15 minutes which I think is acceptable so I enable the arm64 cross building for all branches (instead of main and tag previously).

I sadly cannot use a matrix for the platform as it will lead to only one image created on our registry.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -166,7 +166,6 @@ jobs:
     - uses: actions/checkout@v3
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
-      if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
     - name: Set up Docker Buildx
       id: buildx
       uses: docker/setup-buildx-action@v1
@@ -208,8 +207,7 @@ jobs:
         cache-from: type=local,src=/tmp/.buildx-cache
         cache-to: type=local,dest=/tmp/.buildx-cache-new
         # For the moment, we only support these two platforms.
-        # Cross building for arm64 only happens on main branch.
-        platforms: ${{ fromJSON('["linux/amd64", "linux/amd64,linux/arm64"]')[github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')] }}
+        platforms: linux/amd64,linux/arm64
 
   build-examples:
     name: Build examples


mauriciovasquezbernal:
Looks very nice. Could you describe the problems with Ubuntu? I'd like to avoid switching building image if possible.

gadget-container/Makefile
```diff
@@ -2,12 +2,14 @@
 gadget-container-deps: ocihookgadget gadgettracermanager nrigadget
 
 
+TARGET ?= $(shell go env GOHOSTARCH=)
```
mauriciovasquezbernal, line 5: TARGET_ARCH?

mauriciovasquezbernal (approved):
LGTM.  But I'd like to have another review on this one, specially because we're introducing a new build dependency: buildx. @alban what do you think about it? 

I'm fine with adding buildx, it's not installed all the times but it's also not so difficult to install. What about adding a note about it on https://github.com/kinvolk/inspektor-gadget/blob/main/docs/CONTRIBUTING.md#setup-developer-environment?

alban:
> > specially because we're introducing a new build dependency: buildx
> 
> I also would have liked to avoid using `buildx` and use conventionnal `build`, but the `BUILTPLATFORM` and `TARGETARCH` variables are only available with `buildx`.
> So, it would have mean having two versions of the Dockerfiles: 
> 
> * one for native build
> * another for cross-building 
> 
> which I would have found heavy for no really added value.

You can use `docker build --build-arg` to pass the platform and the arch, like it was done for native + cross-building local-gadget in a single Dockerfile (see [Makefile#L77-L78](https://github.com/kinvolk/inspektor-gadget/blob/48415e3c0211c966bbd63f12749261cea1f61579/Makefile#L77-L78) and [local-gadget.Dockerfile#L27](https://github.com/kinvolk/inspektor-gadget/blob/48415e3c0211c966bbd63f12749261cea1f61579/local-gadget.Dockerfile#L27))

docs/CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ Inspektor Gadget.
 
 - [Fork](https://github.com/kinvolk/inspektor-gadget/fork) and clone this repo:
     - `git clone git@github.com:your_account/inspektor-gadget.git`.
-- Install [Docker](https://docs.docker.com/get-docker/) and [Golang](https://golang.org/doc/install).
+- Install [Docker](https://docs.docker.com/get-docker/), [Docker Buildx](https://docs.docker.com/buildx/working-with-buildx/#static-binaries) and [Golang](https://golang.org/doc/install).
```
alban, line 20:
```suggestion
- Install [Docker](https://docs.docker.com/get-docker/), [Docker Buildx](https://docs.docker.com/buildx/working-with-buildx/) and [Golang](https://golang.org/doc/install).
```

gadget-core.Dockerfile
```diff
@@ -5,28 +5,30 @@
...
 
 # Prepare and build gadget artifacts in a container
-FROM ${BUILDER_IMAGE} as builder
+FROM --platform=${BUILDPLATFORM} ${BUILDER_IMAGE} as builder
 
 ARG ENABLE_BTFGEN=false
 ENV ENABLE_BTFGEN=${ENABLE_BTFGEN}
 
+ARG TARGETARCH
+# We need a cross compiler and libraries for TARGETARCH due to CGO.
```
alban, line 18: Should we have a default value?

```suggestion
ARG TARGETARCH=amd64
# We need a cross compiler and libraries for TARGETARCH due to CGO.
```

gadget-container/Makefile
```diff
@@ -2,12 +2,14 @@
 gadget-container-deps: ocihookgadget gadgettracermanager nrigadget
 
 
+TARGET_ARCH ?= $(shell go env GOHOSTARCH)
```
alban, line 5: Shouldn't this be 
```suggestion
TARGET_ARCH ?= $(shell go env GOARCH)
```

(not relevant when called from the Dockerfiles, or on usual setups where GOHOSTARCH=GOARCH)

alban (approved):
LGTM with some questions

PR #741, opened 2022-07-08
ci: Add back local-gadget-linux-arm64.

### LINKED ISSUE #471: Fail to cross-compile local-gadget for arm64
Hi.

This bug was first found in the [CI](https://github.com/kinvolk/inspektor-gadget/runs/4844370754?check_suite_focus=true#step:7:802) but I was able to reproduce it locally:

```bash
make local-gadget-linux-arm64
...
Step 7/7 : RUN  CGO_ENABLED=1   CC=aarch64-linux-gnu-gcc        GOOS=linux GOARCH=arm64         PKG_CONFIG_PATH=/usr/lib/aarch64-linux-gnu/pkgconfig     go build        -tags withebpf  -o local-gadget-linux-arm64     github.com/kinvolk/inspektor-gadget/cmd/local-gadget
 ---> Running in ac4d1286438b
pkg/gadgets/dns/tracer/embed-ebpf.go:24:12: pattern bpf/dns.o: no matching files found
pkg/gadgets/process-collector/tracer/embed-ebpf.go:27:12: pattern bpf/process-collector-with-filter.o: no matching files found
pkg/gadgets/seccomp/tracer/embed-ebpf.go:24:12: pattern bpf/seccomp.o: no matching files found
```

Indeed, these files do not exist inside the cross-container.
The fix would need two things:
1. Use `make local-gadget-linux-arm64` instead of `go build`.
2. Cross compile eBPF file (_i.e._ to modify pkg/*/bpf/Makefile to add a BPF_CC value we can override).

Best regards.

eiffel-fl:
Hi.

In this PR, I added back compilation local-gadget on linux/arm64.
It fixes #471.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -122,8 +122,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # TODO add local-gadget-linux-arm64
-        local-gadget-target: [local-gadget-linux-amd64]
+        local-gadget-target: [local-gadget-linux-amd64, local-gadget-linux-arm64]
     steps:
     - uses: actions/checkout@v3
     - name: Setup go


mauriciovasquezbernal (approved):
LGTM! Thanks for working on this.

PR #742, opened 2022-07-11
kubectl-gadget: Remove trace CR dependency from CLI code

blanquicet:
# Remove trace CR dependency from CLI code

This PR removes the dependency of the Trace CR from the CLI-specific code. Now, the only code that imports `gadgetv1alpha1` is [utils/trace](https://github.com/kinvolk/inspektor-gadget/blob/d31164c5726530fd81eb50bbb4156b92ab474fad/cmd/kubectl-gadget/utils/trace.go#L40) which is indeed the module in charge of transforming the user request (through the CLI) to a Trace CR to communicate with the gadget pod.

## Implementation details

The advise-seccomp gadget could be still improved to use `PrintTraceOutputFromStatus` only when it is using the `Status` output mode. It would also allow to remove the `traceOutputMode` from the callback, see [here](https://github.com/kinvolk/inspektor-gadget/blob/da286f7ed1ed3769840d402cada5f4172bb9cd03/cmd/kubectl-gadget/utils/trace.go#L716-L719).

## How to use

It does not have any impact on the client-side.

### DIFF
--- a/cmd/kubectl-gadget/advise/seccomp-profile.go
+++ b/cmd/kubectl-gadget/advise/seccomp-profile.go
@@ -23,7 +23,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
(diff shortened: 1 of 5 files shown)


mauriciovasquezbernal (approved):
Good cleanup!

PR #743, opened 2022-07-12
CI: Use two different container image repositories

eiffel-fl:
Hi.

This PR introduces two different container image repositories:

* one (ghcr.io/kinvolk/inspektor-gadget) used for push on main and push of tags
* the other (ghcr.io/kinvolk/inspektor-gadget-dev) for any other push.

Best regards.

### DIFF
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -23,8 +23,15 @@ To run the integration tests, it is necessary to have the gadget container image
 available on a container repository so that it can be installed in the
 Kubernetes cluster where the tests will run.
 
-As a default, `ghcr.io/${{ github.repository }}` is used to store images created
-in the CI pipeline.
+As a default, `ghcr.io/${{ github.repository }}-dev` is used to store images
+created in the CI pipeline for all branches except main and tags.
+When the target branch correspond to the main or the push refers to a tag, the
+default repository is `ghcr.io/${{ github.repository }}`.
+This permits a clear separation between "in development" images and production
+ones.
+During a release, integration test container image will be pushed to
+`ghcr.io/${{ github.repository }}-test`.
+
 Note that, you need to [set repository packages as public](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility#configuring-visibility-of-container-images-for-your-personal-account) to allow anonymous pull.
 
 ## Run integration tests on an ARO cluster


.github/workflows/inspektor-gadget.yml
```diff
@@ -1,6 +1,8 @@
 name: Inspektor Gadget CI
 env:
-  CONTAINER_REPO: ghcr.io/${{ github.repository }}
+  REGISTRY: ghcr.io
+  RELEASE_CONTAINER_REPO: ${{ github.repository }}
```
blanquicet, line 4: Did you verify pushing to main or a tag in a private repo uses the correct repo?

.github/workflows/inspektor-gadget.yml
```diff
@@ -1,6 +1,8 @@
...
+  DEVELOPMENT_CONTAINER_REPO: ${{ github.repository }}-dev
```
blanquicet, line 5: Could you please update the documentation with this new approach?

blanquicet, line 5: Wasn't it `-ci`? 

BTW, could you please also add a `-` in the `$(CONTAINER_REPO)test` so that we have:
- `ghcr.io/kinvolk/inspektor-gadget`
- `ghcr.io/kinvolk/inspektor-gadget-ci`
- `ghcr.io/kinvolk/inspektor-gadget-test`

If you prefer, you can do it in another PR.

blanquicet (changes requested):
I haven't gone through all the details yet and didn't test it, but it looks great already. 

I noticed you simplified the PR by not touching the computation of the `IMAGE_TAG`. In addition, you found that we don't need to modify the `CONTAINER_REPO` for the integration tests because it will only be pushed for tags (releases). It simplified the PR. Good catch!

.github/workflows/inspektor-gadget.yml
blanquicet: Could you please rebase this PR after 740 was merged to verify this problem is gone?

mauriciovasquezbernal (commented):
Sorry for the late comment here, I've forgotten to click "Submit review"

.github/workflows/inspektor-gadget.yml
```diff
@@ -85,6 +87,13 @@ jobs:
             IMAGE_TAG="latest"
         fi
         echo IMAGE_TAG=$IMAGE_TAG >> $GITHUB_ENV
+    - name: Choose container repository
+      id: choose-repo
+      uses: ./.github/actions/choose-container-repo
+      with:
+        registry: ${{ env.REGISTRY }}
+        container-image-release: ${{ env.RELEASE_CONTAINER_REPO }}
+        container-image-dev: ${{ env.DEVELOPMENT_CONTAINER_REPO }}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > I really want to keep this as simple as possible and I am not convinced than hiding things in env make it easier to understand.

I don't understand what is the concern here. We can have a step like similar to `Set IMAGE_TAG` but for `CONTAINER_REPO:. 

```
        if [[ ${{ github.ref }} == 'refs/heads/main' || ${{ github.ref }} == 'refs/tags/v'* ]]; then
          CONTAINER_REPO=$env.RELEASE_CONTAINER_REPO
        else
          CONTAINER_REPO=$env.DEVELOPMENT_CONTAINER_REPO
        fi
        echo CONTAINER_REPO=$CONTAINER_REPO >> $GITHUB_ENV
```

It's true that we have to repeat this many times, but we're doing the same for `Choose container repository` too, hence not difference. Does it make sense to you?

docs/ci.md
```diff
@@ -23,8 +23,13 @@ To run the integration tests, it is necessary to have the gadget container image
...
 Kubernetes cluster where the tests will run.
 
-As a default, `ghcr.io/${{ github.repository }}` is used to store images created
-in the CI pipeline.
+As a default, `ghcr.io/${{ github.repository }}-dev` is used to store images
+created in the CI pipeline for all branches except main and tags.
+When the target branch correspond to the main or the push refers to a tag, the
+default repository is `ghcr.io/${{ github.repository }}`.
+This permits a clear separation between "in development" images and production
+ones.
```
blanquicet, line 31: The description of the `ghcr.io/${{ github.repository }}-test` is missing here.

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
blanquicet: My concern in the previous PR was that it was setting the environment variables inside a custom action. Instead, using the action's output makes things more clear. I think it would make things even better to move the image-tag computation inside the `choose-container-repo` and have:

```yaml
CONTAINER_REPO=${{ steps.new-fancy-name.outputs.container-repo }} IMAGE_TAGE=${{ steps.new-fancy-name.outputs.image-tag }} make $kubectl_gadget
```

integration/Makefile
```diff
@@ -1,4 +1,5 @@
 CONTAINER_REPO ?= ghcr.io/kinvolk/inspektor-gadget
+TEST_CONTAINER_REPO ?= $(CONTAINER_REPO)-test
```
blanquicet, line 2: It is missing to update https://github.com/kinvolk/inspektor-gadget/blob/3da7b2cfc76a75b28095c246c5593314dacc4815/integration/gadget-integration-tests-job.yaml#L38

blanquicet (approved):
IMO, after resolving this comment, this PR is ready to go. We can improve the `IMAGE_TAG` management in another PR. @mauriciovasquezbernal WDYT?

PR #745, opened 2022-07-12
pkg/gadgets: Move code related to Kubernetes operator to different package

mauriciovasquezbernal:
Code in pkg/gadgets is composed by:
- tracer/core: gadget tracer implementation using CO-RE
- tracer/standard: gadget tracer implementation on top of BCC
- gadget.go: glue code that implements the TraceFactory interface

The code in gadget.go is not related to the tracers itself but it's
used to "glue" the trace controller and the trace implementation. In
order to improve the code organization (by providing a better separation
of concerns) this commit moves all the gadget.go files to
pkg/gadget-collection.

There is not any change in functinality introduced by this commit, all
changes are files movements and renaming of import paths.

This PR is part of https://github.com/kinvolk/inspektor-gadget/issues/723. There are another changes I'd like to propose but I realized I was changing too many things on a single PR, so I'll be opening rather small ones.


blanquicet (approved):
I think it is a good move because it clarifies the code's structure for those just importing a gadget tracer (e.g. `local-gadget` could be a use-case for this). In fact, I wrote an example, and the imported packages are all related to the tracer gadget are packages strictly related to the tracer itself and do not include any CRD logic (We will be able to remove the gRPC API dependency with https://github.com/kinvolk/inspektor-gadget/pull/748 :partying_face:):

https://github.com/kinvolk/inspektor-gadget/blob/37d6020482b58036865c1d0300b4da0a9ea64562/examples/gadget-standalone/main.go#L21-L23

Look at the the output of the example. Just great! :100:
```bash
$ sudo ./gadget-standalone
{"type":"normal","node":"foo","namespace":"default","pod":"calico-kube-controllers","container":"calico-kube-controllers","pid":165034,"ppid":165022,"uid":999,"mountnsid":4026532441,"pcomm":"check-status","args":["/usr/bin/check-status","-l"]}
{"type":"normal","node":"foo","namespace":"default","pod":"calico-node","container":"calico-node","pid":165050,"ppid":3501,"mountnsid":4026532295,"pcomm":"iptables-legacy","args":["/usr/sbin/iptables-legacy-save","-t","nat"]}
{"type":"normal","node":"foo","namespace":"default","pod":"gadget","container":"gadget","pid":165061,"ppid":165051,"mountnsid":4026532517,"pcomm":"gadgettracerman","args":["/bin/gadgettracermanager","-liveness"]}

$ sudo ./gadget-standalone  --containername gadget
{"type":"normal","node":"foo","namespace":"default","pod":"gadget","container":"gadget","pid":166904,"ppid":166892,"mountnsid":4026532517,"pcomm":"gadgettracerman","args":["/bin/gadgettracermanager","-liveness"]}
{"type":"normal","node":"foo","namespace":"default","pod":"gadget","container":"gadget","pid":166989,"ppid":166978,"mountnsid":4026532517,"pcomm":"gadgettracerman","args":["/bin/gadgettracermanager","-liveness"]}
```


I am just wondering if we should reorder the `pkg/gadgets` folder by categories as we did for `cmd/kubectl-gadget`, `cmd/local-gadget` and `pkg/gadgets-collection`. I think it could make navigation through the code more consistent, even though I know that users will only import one gadget at a time so they could not care about categories in that case. 

Otherwise, LGTM!

mauriciovasquezbernal:
> I am just wondering if we should reorder the `pkg/gadgets` folder by categories as we did for `cmd/kubectl-gadget`, `cmd/local-gadget` and `pkg/gadgets-collection`.

This is something I'll do but in another PR. I tried to put all these changes on a single one and it was too much.

mauriciovasquezbernal:
> This is something I'll do but in another PR. I tried to put all these changes on a single one and it was too much.

Done in https://github.com/kinvolk/inspektor-gadget/pull/749.

alban (approved):
LGTM.

This will require @flyth's https://github.com/kinvolk/inspektor-gadget/pull/754 to be rebased.

PR #754, opened 2022-07-18
New gadget: ebpftop

### LINKED ISSUE #704: [RFE] Measure CPU overhead of eBPF programs
## Current situation

Inspektor Gadget installs eBPF programs that are going to use CPU but there are no easy way to measure the CPU usage of those programs.

Kubernetes has tools to measure CPU usage of pods (kubectl top, metrics-server, cAdvisor, see [documentation](https://kubernetes.io/docs/tasks/debug/debug-cluster/resource-usage-monitoring/)) but those tools don't measure eBPF programs.

## Impact

Users don't know if Inspektor Gadget would impact the performance of their Kubernetes clusters.

## Ideal future situation

There should be a simple way to measure CPU usage of Inspektor Gadget's eBPF programs.

Additionally, it would be good if Inspektor Gadget could measure CPU usage of eBPF programs installed by other projects (Falco, Cilium, etc).

### Methods to avoid:

- :x: BPF helper function bpf_ktime_get_ns() BPF helper function to measure the time at the beginning and end of BPF programs. This is not reliable and it would require to modify all eBPF programs to measure.
- :x: Linux' BPF_PROG_TEST_RUN facility. This does not run realistic scenario.

### Methods to use:

https://qmonnet.github.io/whirl-offload/2021/09/23/bpftool-features-thread/#programs-statistics

> Linux 5.1 introduced statistics for attached eBPF programs: it can collect the total run time and the run count for each program. When available, this information is displayed by bpftool when listing the programs:
> `# bpftool prog show`
> But gathering statistics slightly impacts performance of the program execution (~10 to 30 nanoseconds per run), so it is disabled by default. Activate it with:
> `# sysctl -w kernel.bpf_stats_enabled=1`

cilium/ebpf has support for this: https://github.com/cilium/ebpf/blob/7e5a2461e928e1a41cfa7cd95af904f56b10fbdb/info.go#L121

Implementations:
1. `kubectl gadget top ebpf -o json`
2. Enable the sysctl on worker nodes (`sysctl -w kernel.bpf_stats_enabled=1` but in Golang)
3. List all eBPF programs and which processes are using them (`sudo bpftool prog` but in Golang)
4. Aggregate stats by pid (e.g. `gadgettracerman`, `systemd`...)
5. Print stats every second

## Additional information

Interesting articles to read:
- https://qmonnet.github.io/whirl-offload/2021/09/23/bpftool-features-thread/#programs-statistics
- https://developers.redhat.com/articles/2022/06/22/measuring-bpf-performance-tips-tricks-and-best-practices
- https://ebpf.io/summit-2020-slides/eBPF_Summit_2020-Lightning-Bryce_Kahle-How_and_When_You_Should_Measure_CPU_Overhead_of_eBPF_Programs.pdf
- https://docs.kernel.org/admin-guide/sysctl/kernel.html?highlight=bpf_stats_enabled#bpf-stats-enabled
- https://github.com/torvalds/linux/commit/492ecee892c2a4ba6a14903d5d586ff750b7e805
- https://github.com/DataDog/ebpfbench/blob/main/stats.go#L48

## Alternative: BPF_ENABLE_STATS

BPF_ENABLE_STATS is an alternative to `sysctl -w kernel.bpf_stats_enabled=1`, but requires Linux >= 5.8: https://github.com/torvalds/linux/commit/d46edd671a147032e22cfeb271a5734703093649

[Example of code in Goland using BPF_ENABLE_STATS](https://github.com/DataDog/ebpfbench/blob/de7131d7ee1d19d79668fb1ad9f6067afe7c3a59/syscall.go#L92)

flyth:
# New gadget: ebpftop

This new top gadget will list runtime and runcount for each installed bpf program. It does so by enabling stats collection using either BPF(BPF_ENABLE_STATS) or via sysctl (as a fallback) and iterating over the programs using the cilium/ebpf library. Afterwards it collects information on the processes with open fds to the programs by either running an ebpf iterator (Linux >= 5.8) or going through the /proc filesystem.

## How to use

The gadget can be run using `kubectl gadget top ebpf`.

## Testing done

- [x] Test on amd64
- [x] Test on arm64
- [x] Manual tests of the fallback situations

## Todo
- [x] Complete documentation & guide
- [x] Compile eBPF program on amd64

## References
Fixes #704


pkg/gadgets/ebpftop/gadget.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+func (f *TraceFactory) OutputModesSupported() map[string]struct{} {
+	return map[string]struct{}{
+		"Stream": {},
+	}
+}
+
+func deleteTrace(name string, t interface{}) {
+	// TODO: need to stop something in here?
```
alban, line 72: The trace could be deleted without `Stop()` being called before, so if the trace was started, you need to stop the Go routine (`t.donechan <- true`) and stop the stats (`bpfstats.DisableBPFStats`) like you wrote in `Stop()`.

pkg/gadgets/network-graph/gadget.go gives an example of code where some code in `Stop()` has been factorized in `stop()` to avoid code duplication in `deleteTrace()`. Some other gadgets duplicated the code in `Stop()` and `deleteTrace()`.

alban (commented):
Thanks! Some initial comments.

pkg/gadgets/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,175 @@
...
+
+func getSpec() (*ebpf.CollectionSpec, error) {
+	cspecLock.Lock()
+	defer cspecLock.Unlock()
+
+	if triedGetSpec {
+		return cspec, cspecErr
+	}
+
+	triedGetSpec = true
```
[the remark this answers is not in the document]
flyth: You're right, it's pretty fast (0.4ms on my test machine) and might add unnecessary complexity. In contrast to the other gadgets, this ebpf program gets called/loaded every second, that's why I optimized a bit further (saving the elf parsing/relocation etc.).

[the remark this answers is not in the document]
alban: Oh, I didn't realise it is loaded every second. Up to you then.

pkg/gadgets/ebpftop/gadget.go
flyth: Thanks! This should be handled correctly, now.

alban:
I see it works with local-gadget and I can see the CPU time used by the dns gadget:
```
$ sudo ./local-gadget
» create ebpftop trace1 
State: Started
» create dns trace2
State: Started
» stream trace1 --follow 
...
      {
         "name" : "bpf_prog1",
         "node" : "local",
         "pids" : [
            {
               "comm" : "local-gadget",
               "pid" : 1249402
            }
         ],
         "progid" : 415,
         "runcount" : 1240,
         "runtime" : 472682,
         "type" : "SocketFilter"
      },
```

pkg/gadgets/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+
+	return res, nil
+}
+
+// IteratePids returns an array of PidIterEntry containing information
+// on which pid (and comm) has an open fd to which eBPF Program ID.
+func IteratePids() ([]*PidIterEntry, error) {
+	return (&pidIter{}).iteratePids()
```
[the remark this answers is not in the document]
alban: With this, the optimisation with `triedGetAddr` and `triedGetSpec` become unnecessary because this code would not be executed every second but only when the trace is created.

pkg/gadgets/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,190 @@
...
+	spec, err := getSpec()
+	if err != nil {
+		return nil, err
+	}
+
+	opts := ebpf.CollectionOptions{}
+
+	if err := spec.LoadAndAssign(&p.objs, &opts); err != nil {
+		return nil, fmt.Errorf("failed to load ebpf program: %w", err)
+	}
```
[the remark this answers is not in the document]
flyth: Changed, thanks!

pkg/gadgets/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,183 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package piditer
+
+// #include <linux/types.h>
+// #include "./bpf/pid_iter.h"
+import "C"
```
alban, line 19: You can ask bpf2go to generate a Go struct equivalent to `C.struct_pid_iter_entry` so you don't need to have `import "C"` here:

```diff
--- a/pkg/gadgets/ebpftop/piditer/bpf/pid_iter.bpf.c
+++ b/pkg/gadgets/ebpftop/piditer/bpf/pid_iter.bpf.c
@@ -8,8 +8,14 @@
 
 const volatile __u64 bpf_prog_fops_addr = 0;
 
+/**
+ * Emit struct type info into the ELF's BTF so bpf2go 
+ * can generate a Go type from it. 
+ */ 
+struct pid_iter_entry *unused __attribute__((unused));
+
 SEC("iter/task_file")
--- a/pkg/gadgets/ebpftop/piditer/iter.go
+++ b/pkg/gadgets/ebpftop/piditer/iter.go
@@ -34,7 +34,7 @@ import (
        "golang.org/x/sys/unix"
 )
 
-//go:generate go run github.com/cilium/ebpf/cmd/bpf2go -target $TARGET -cc clang piditer ./bpf/pid_iter.bpf.c -- -I./bpf/ -I../../../${TARGET}
+//go:generate go run -mod vendor github.com/cilium/ebpf/cmd/bpf2go -target $TARGET -cc clang -type pid_iter_entry piditer ./bpf/pid_iter.bpf.c -- -I./bpf/ -I../../../${TARGET}
 
 type pidIter struct {
        objs piditerObjects
```

Then piditer_bpfel_x86.go will contain the following:
```
type piditerPidIterEntry struct {
       Id   uint32
       Pid  uint32
       Comm [16]int8
}
```

pkg/gadgets/ebpftop/piditer/bpf/pid_iter.bpf.c
```diff
@@ -0,0 +1,36 @@
...
+    // TODO: Currently cilium/ebpf doesn't support .ksyms, this is why we get the info from userspace right now
+    if (bpf_prog_fops_addr != 0 && (__u64) file->f_op != bpf_prog_fops_addr)
+        return 0;
+
+	__builtin_memset(&e, 0, sizeof(e));
+	e.pid = task->tgid;
+	e.id = BPF_CORE_READ((struct bpf_prog *)(file->private_data), aux, id);
+
+	bpf_probe_read_kernel_str(&e.comm, sizeof(e.comm), task->group_leader->comm);
+	bpf_seq_write(ctx->meta->seq, &e, sizeof(e));
```
alban, line 31: I didn't know about `bpf_seq_write()`. Other gadgets use `bpf_seq_printf()`. Either is fine for me, but at some point we should probably be consistent. Using `bpf_seq_printf()` like other gadgets would have the advantage to not hardcode the size 24, or to convert the C string to a Go string.

pkg/gadgets/ebpftop/piditer/bpf/pid_iter.bpf.c
```diff
@@ -0,0 +1,36 @@
...
+	struct task_struct *task = ctx->task;
+	struct pid_iter_entry e;
+
+	if (!file || !task)
+		return 0;
+
+    // If we do have an address of bpf_prog_fops, use it; otherwise... well...
+    // TODO: Currently cilium/ebpf doesn't support .ksyms, this is why we get the info from userspace right now
+    if (bpf_prog_fops_addr != 0 && (__u64) file->f_op != bpf_prog_fops_addr)
+        return 0;
```
alban, line 24: inconsistent alignment (tab vs spaces)

pkg/gadgets/ebpftop/piditer/bpf/pid_iter.bpf.c
flyth: I did some performance tests and it seems that the `bpf_seq_printf()` variant is much slower. I don't think it matters too much (it's still pretty fast) in this case, but there might be other gadgets (in the future) that would profit from the `bpf_seq_write()` version. I replaced the hardcoded values and removed the "C" import, though.

pkg/gadgets/ebpftop/piditer/iter.go
flyth: Changed, thanks!

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,273 @@
...
+func (f *TraceFactory) Description() string {
+	t := `ebpftop shows cpu time used by ebpf programs.
+
+The following parameters are supported:
+ - %s: Output interval, in seconds. (default %d)
+ - %s: Maximum rows to print. (default %d)
+ - %s: The field to sort the results by (%s). (default %s)`
+	return fmt.Sprintf(t, types.IntervalParam, types.IntervalDefault,
+		types.MaxRowsParam, types.MaxRowsDefault,
+		types.SortByParam, strings.Join(types.SortBySlice, ","), types.SortByDefault)
```
alban, line 63: We have not detailed parameters like this in other gadgets. This requires `types.MaxRowsDefault` to be defined in  `pkg/gadgets/ebpftop/types` even though it's specific to the kubectl-gadget CLI, so it makes me uneasy...

Also this Description() function generates the documentation for interacting with the Trace CR directly, it's not about the documentation for how to use the `kubectl-gadget` CLI (which would be in `docs/guides/top/` instead).

Arguably, separating the documentation for the different Trace CR and for the different `kubectl-gadget` commands is not ideal, but I think it should be addressed for all gadgets in a separate PR.

cc @blanquicet

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,273 @@
...
+	// program. If that doesn't work, we will fall back to scanning
+	// all used fds in all processes /proc/$pid/fdinfo/$fd.
+	iter, err := piditer.NewPidIter()
+	if err != nil {
+		useFallbackIterator = true
+	}
+
+	t.started = true
+
+	go func() {
```
alban, line 130: Since this go function is 120 lines long, it could be moved in a separate function.
```
go t.run()
```

This will require moving `useFallbackIterator` and `iter` as fields of the `Trace struct`.

pkg/gadgets/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,191 @@
...
+	if err := spec.LoadAndAssign(&p.objs, &opts); err != nil {
+		return nil, fmt.Errorf("failed to load ebpf program: %w", err)
+	}
+
+	iter, err := link.AttachIter(link.IterOptions{
+		Program: p.objs.GadgetEbpftop,
+		Map:     nil,
+	})
+	if err != nil {
+		p.objs.Close()
```
alban, line 126: This `p.objs.Close()` would be easy to forget if we add more `if err != nil` sections below. We cannot just use `defer p.objs.Close()` because we don't want to do it in the success path.

I've started to use the following code pattern to solve this error management issue in various gadgets. In this way, the `defer p.objs.Close()` would not be forgotten if we add more sections below.

* https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/pkg/gadgets/dns/tracer/tracer.go#L93-L109
* https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/pkg/gadgets/network-graph/tracer/tracer.go#L109-L117
* https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/pkg/gadgets/snisnoop/tracer/tracer.go#L97-L109

pkg/gadgets/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		if n%iterEntrySize != 0 {
+			return nil, fmt.Errorf("invalid format: %d", n)
+		}
+		for i := 0; i < n/iterEntrySize; i++ {
+			entry := (*piditerPidIterEntry)(unsafe.Pointer(&buf[i*iterEntrySize]))
+			res = append(res, &PidIterEntry{
+				ProgID: entry.Id,
+				Pid:    entry.Pid,
+				Comm:   string(entry.Comm[:cstrlen(entry.Comm[:])]),
```
alban, line 173: In various gadgets, we didn't take special care for checking the NULL-terminated string in the `comm[16]` field:
```
$ git grep -n 'GoString.*comm'
pkg/gadgets/audit-seccomp/tracer/tracer.go:165:                 Comm:      C.GoString(&eventC.comm[0]),
pkg/gadgets/execsnoop/tracer/core/tracer.go:162:                        Comm:      C.GoString(&eventC.comm[0]),
pkg/gadgets/filetop/tracer/tracer.go:187:                       Comm:       C.GoString(&fileStat.comm[0]),
pkg/gadgets/mountsnoop/tracer/core/tracer.go:178:                       Comm:      C.GoString(&eventC.comm[0]),
pkg/gadgets/oomkill/tracer/tracer.go:154:                       TriggeredComm: C.GoString(&eventC.fcomm[0]),
pkg/gadgets/oomkill/tracer/tracer.go:156:                       KilledComm:    C.GoString(&eventC.tcomm[0]),
pkg/gadgets/opensnoop/tracer/core/tracer.go:198:                        Comm:      C.GoString(&eventC.comm[0]),
pkg/gadgets/sigsnoop/tracer/core/tracer.go:234:                 Comm:      C.GoString(&eventC.comm[0]),
```

https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/pkg/gadgets/oomkill/tracer/tracer.go#L154-L156

Last time I checked, the kernel was not allowed the `comm[16]` field to be non NULL-terminated. But I agree this is fragile, I don't know if future kernel could change that.

In any case, I think it should be consistent among various gadgets. So either:
* use `C.GoString(&entry.comm[0])` and remove `cstrlen`, or
* write a helper function to convert the `comm[16]` to a Go string in some shared package and use it in all gadgets

cc @mauriciovasquezbernal Any opinion?

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,273 @@
...
+						RunCount:  rc,
+					})
+
+					prog.Close()
+				}
+
+				pidmap := make(map[uint32][]*types.PidInfo)
+
+				if !useFallbackIterator {
+					pids, err = iter.DumpPids()
```
alban, line 183: It seems `var pids []*piditer.PidIterEntry` could be declared just before here since it's not reused at the next tick.

pkg/bpfstats/bpfstats.go
```diff
@@ -0,0 +1,123 @@
...
+	defer mutex.Unlock()
+
+	refCnt++
+
+	if refCnt != 1 {
+		return nil
+	}
+
+	// Actually enable
+	s, err := ebpf.EnableStats(unix.BPF_STATS_RUN_TIME)
```
alban, line 65: If you want to add unit tests in pkg/bpfstats/bpfstats_test.go... (not sure if it's worth it)

It could test both methods. The unit test could overwrite the global variable `method` to choose what to test.

Since the test needs to be run as root, it needs to be guarded. Example of unit tests guarded by a flag:
https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/pkg/local-gadget-manager/local-gadget-manager_test.go#L43

And then add a target in the Makefile to use `-root-test`:
https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/Makefile#L148-L152

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
blanquicet: > We have not detailed parameters like this in other gadgets. 
>
All the gadget in the top category has the same detailed parameters, e.g., in in [file gadget](https://github.com/kinvolk/inspektor-gadget/blob/04c2a3664d830cfc39c294f9a37b1b73159de008/pkg/gadget-collection/gadgets/top/file/gadget.go#L53-L57). I think it is useful for third-party applications that will explore the documentation on [docs/gadgets](https://github.com/kinvolk/inspektor-gadget/tree/main/docs/gadgets), don't you think?

However, I agree that we it is not idea to have it in two places. Maybe we could move it to a common place where it keeps sync between kubectl-gadget and the CRD description? Of course, in another PR.

> This requires types.MaxRowsDefault to be defined in pkg/gadgets/ebpftop/types even though it's specific to the kubectl-gadget CLI
>
It shouldn't be specific for kubectl-gadget because `types.IntervalParam`, `types.MaxRowsParam` and  `types.SortByParam` should be used to extract the params, e.g. in [file gadget](https://github.com/kinvolk/inspektor-gadget/blob/e8250746a361a67bf90cdb2dee48cfe730693bba/pkg/gadget-collection/gadgets/top/file/gadget.go#L113-L148).

alban: Oh, I didn't realise those parameters were in the Trace custom resource. Ok then, you can ignore my comment.

docs/gadgets/ebpftop.md
```diff
@@ -0,0 +1,50 @@
...
+# Code generated by 'make generate-documentation'. DO NOT EDIT.
+title: Gadget ebpftop
+---
+
+ebpftop shows cpu time used by ebpf programs.
+
+The following parameters are supported:
+ - interval: Output interval, in seconds. (default 1)
+ - max_rows: Maximum rows to print. (default 20)
+ - sort_by: The field to sort the results by (all,runtime,runcount,programid). (default all)
```
alban, line 11: Could you use those parameters in `pkg/resources/samples/trace-ebpftop.yaml` to show how they can be used?

There is an example in another gadget: pkg/resources/samples/trace-socket-collector.yaml

I think numbers in the yaml have to be in quotes since `Parameters` is a `map[string]string`.

flyth: Yeah, I will add that. Just saw that I was missing a block of code that is using these values in the gadget itself (and not only in the kubectl-gadget). Will add that as well :)

pkg/gadgets/ebpftop/piditer/iter.go
mauriciovasquezbernal: Good question. I think it's a potential bug and we should handle the case when comm is not null terminated. I'd suggest to use `C.GoString(&entry.comm[0])` on this PR and open an issue to track and fix it in all gadgets in a single pass. 

> write a helper function to convert the comm[16] to a Go string in some shared package and use it in all gadgets

It seems like the easiest / safest way to go.

pkg/gadgets/top/ebpftop/piditer/iter.go
```diff
@@ -0,0 +1,191 @@
...
+				Comm:   C.GoString((*C.char)(unsafe.Pointer(&entry.Comm[0]))),
+			})
+		}
+	}
+
+	return res, nil
+}
+
+func (p *PidIter) Close() error {
+	err := p.iter.Close()
```
blanquicet, line 182: I haven't analysed the code yet, but I just ran it to see it working, and it is systematically crashing only on one node when deleting the trace. I am testing it on a multi-node cluster:

```
time="2022-07-27T10:58:32Z" level=info msg="Reconcile trace gadget/ebpftop-lxk2w (gadget ebpftop, node worker)"
time="2022-07-27T10:58:32Z" level=info msg="Gadget ebpftop operation \"start\" on gadget/ebpftop-lxk2w"
time="2022-07-27T10:58:32Z" level=info msg="Gadget completed operation. Trace status will be updated accordingly"
time="2022-07-27T10:58:32Z" level=info msg="Updating new status of trace \"gadget/ebpftop-lxk2w\": state=Started operationError=\"\" operationWarning=\"\" output=<0 characters>"
time="2022-07-27T10:58:32Z" level=info msg="Reconcile trace gadget/ebpftop-lxk2w (gadget ebpftop, node worker)"
time="2022-07-27T10:58:32Z" level=info msg="No annotations. Nothing to do."
time="2022-07-27T11:00:38Z" level=info msg="Reconcile trace gadget/ebpftop-lxk2w (gadget ebpftop, node worker)"
time="2022-07-27T11:00:38Z" level=info msg="Deleting gadget/ebpftop-lxk2w"
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x20b0f79]

goroutine 176 [running]:
github.com/kinvolk/inspektor-gadget/pkg/gadgets/top/ebpftop/piditer.(*PidIter).Close(0x0)
  /gadget/pkg/gadgets/top/ebpftop/piditer/iter.go:182 +0x19
github.com/kinvolk/inspektor-gadget/pkg/gadget-collection/gadgets/top/ebpf.(*Trace).run(0xc0002a6de0)
  /gadget/pkg/gadget-collection/gadgets/top/ebpf/gadget.go:210 +0x36b
created by github.com/kinvolk/inspektor-gadget/pkg/gadget-collection/gadgets/top/ebpf.(*Trace).Start
  /gadget/pkg/gadget-collection/gadgets/top/ebpf/gadget.go:178 +0x56c

```

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -0,0 +1,296 @@
...
+				comm = event.Pids[0].Comm
+			}
+			fmt.Printf("%-16s %-8d %-16s %-16s %-7s %-20s %12v %6d\n",
+				event.Node,
+				event.ProgramID,
+				event.Type,
+				event.Name,
+				pid, comm,
+				time.Duration(event.Runtime),
+				event.RunCount,
```
blanquicet, line 215: Is there any way to align the time to the left? It looks wear:

```bash
NODE             PROG     TYPE             NAME             PID     COMM                 TIME         COUNT
master           32       Tracing          gadget_ebpftop   129449  gadgettracerman           37.18µs   1447
master           11       CGroupSKB                         1       systemd                        0s      0
master           15       CGroupSKB                         1       systemd                        0s      0
master           16       CGroupSKB                         1       systemd                        0s      0

```

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -0,0 +1,296 @@
...
+
+func ebpfPrintHeader() {
+	switch params.OutputMode {
+	case commonutils.OutputModeColumns:
+		if term.IsTerminal(int(os.Stdout.Fd())) {
+			utils.ClearScreen()
+		} else {
+			fmt.Println("")
+		}
+		fmt.Printf("%-16s %-8s %-16s %-16s %-7s %-20s %-12s %-6s\n",
```
blanquicet, line 163: BTW, you are not using the same format here for the time column: %s vs %v (when using custom-columns)

blanquicet (commented):
There are just a few initial comments after a first run, but it looks great! I will take a look at the code later today.

cmd/kubectl-gadget/top/ebpf.go
flyth: Hmm, I think aligning the time to the right makes it much more readable/comparable (especially when you sort by time), same goes for the count.

Specifically regarding the time: go will use whatever time unit suits best with %v, so sometimes the suffix will be "s", sometimes it's "µs" (like in your example) and so on. That makes it even more tricky to comparable it to other rows. However, not using the unit automation (always using "ms" for example) could lead to massive or always rounded to zero numbers.

Unsorted it would look something like that (left vs right aligned):
```
TIME         TIME
1.13ms     1.13ms
1.05s       1.05s
10ms         10ms
8s             8s
```

Sorted:
```
TIME         TIME
8s             8s
1.05s       1.05s
10ms         10ms
1.13ms     1.13ms
```

I would always prefer the right aligned values, IMHO.

Maybe the header should be aligned to the right as well? What do you think?

blanquicet: You're right! I think aligning the columns in the same way would improve the view.

pkg/gadgets/top/ebpftop/piditer/iter.go
flyth: Looks like it used the fallback iterator and I missed checking for that. Thanks!

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,336 @@
+// Copyright 2019-2022 The Inspektor Gadget authors
```
mauriciovasquezbernal, line 1: We are trying to make the tracers usable by 3rd party applications: https://github.com/kinvolk/inspektor-gadget/issues/723. I think most of the logic on this file should be moved to `pkg/gadgets/top/ebpf/tracer.go` and this file should only contain logic related to the trace resource. (I think you can take inspiration from top file, that uses a callback to pass events from tracer.go to gadget.go).

mauriciovasquezbernal (changes requested):
Some blocking comments regarding the tracer.go / gadget.go separation.

pkg/gadgets/top/ebpf/piditer/iter.go
```diff
@@ -0,0 +1,191 @@
...
+	err := p.iter.Close()
+	if err != nil {
+		return err
+	}
+	err = p.objs.Close()
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
flyth: You're right - better clean up in any case.

pkg/gadgets/top/ebpf/piditer/iter.go
```diff
@@ -0,0 +1,191 @@
...
+
+func NewTracer() (iter *PidIter, err error) {
+	p := &PidIter{}
+	defer func() {
+		if err != nil {
+			if p.iter != nil {
+				p.iter.Close()
+			}
+			p.objs.Close()
```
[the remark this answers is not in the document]
flyth: p.objs is not a pointer but an embedded struct.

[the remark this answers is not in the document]
alban: The embedded struct contains a pointer:
```
GadgetEbpftop *ebpf.Program `ebpf:"gadget_ebpftop"`
```

But the method [ebpf.Program.Close()](https://github.com/cilium/ebpf/blob/a293887a364da3506568f9bc3c627854f5af9a2c/prog.go#L468-L470) already checks if the pointer is nil, so this is safe to call `p.objs.Close()`.

pkg/gadgets/top/ebpf/piditer/bpf/pid_iter.bpf.c
```diff
@@ -0,0 +1,38 @@
...
+int gadget_ebpftop(struct bpf_iter__task_file *ctx)
+{
+	struct file *file = ctx->file;
+	struct task_struct *task = ctx->task;
+	struct pid_iter_entry e;
+
+	if (!file || !task)
+		return 0;
+
+	// If we do have an address of bpf_prog_fops, use it; otherwise... well...
```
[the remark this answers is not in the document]
flyth: They say "there be dragons".

Seriously though:
If `bpf_prog_fops_addr` is 0, the program will run on more than just programs. In tests it worked and didn't have any weird outcome (`BPF_CORE_READ` is the only action specific to bpf programs and it will return 0 if it accesses invalid memory. But I can imagine that it also potentially could return garbage if the address of `file->private_data` is valid and contains enough data - that would lead to a wrong program id and possibly result in ebpf programs showing wrong pid/comm information).

This was implemented in the first version where we had no fallback for the iterator. Now that we have one, it might be better to entirely skip the ebpf iterator and fallback to the /proc/./fdinfo scanning when the address of `bpf_prog_fops_addr` can't be found.

pkg/gadgets/top/ebpf/piditer/iter.go
```diff
@@ -0,0 +1,185 @@
+// Copyright 2019-2022 The Inspektor Gadget authors
```
blanquicet, line 1: This code could be useful for other gadgets so maybe we can reorganize this in another way and move it out from this folder. Anyway, it will be done once at least another gadget starts using it.

blanquicet (commented):
It looks really good overall. Thanks! I just added a couple of comments.

pkg/gadgets/top/ebpftop/piditer/iter.go
blanquicet: Resolved. Thanks!

pkg/gadgets/top/ebpf/piditer/iter.go
flyth: Yeah, I was thinking kind of the same, however I didn't really know where to put it. For all other gadgets, the ebpf part holds the main functionality. In ebpftop it's only a helper that is used to get the processes that are using the ebpf programs. Then, however, we would have ebpf code in other packages than pkg/gadgets (and generate in there), so I'm not sure what to do. But maybe it's something we can discuss in another PR?

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,339 @@
...
+func (t *Trace) stop() error {
+	t.donechan <- true
+	t.started = false
+	return bpfstats.DisableBPFStats()
+}
+
+func (t *Trace) run() {
+	if !t.useFallbackIterator {
+		defer t.iter.Close()
+	}
```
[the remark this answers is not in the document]
flyth: I can't think of a reason why not. :)

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,339 @@
...
+				if err != nil {
+					prog.Close()
+					continue
+				}
+				rt, _ := pi.Runtime()
+				rc, _ := pi.RunCount()
+
+				stats = append(stats, types.Stats{
+					ProgramID: uint32(curID),
+					Pids:      nil,
```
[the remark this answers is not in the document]
flyth: I'll try to remember to remove the nils - there was already another comment on such a case.
Let's just blame my IDE for now, as it automatically adds the full struct with default values for me...

pkg/bpfstats/bpfstats.go
```diff
@@ -0,0 +1,123 @@
...
+	case MethodBPFFunc:
+		err := statsSock.Close()
+		statsSock = nil
+		if err != nil {
+			return fmt.Errorf("could not disable stat collection using BPF(): %w", err)
+		}
+	case MethodSysctl:
+		err := ioutil.WriteFile("/proc/sys/kernel/bpf_stats_enabled", []byte("0"), 0o644)
+		if err != nil {
+			refCnt = 0
```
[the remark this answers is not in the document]
flyth: Hmm, could you clarify, please?
If you call `EnableBPFStats()`, it will set it to "1", `DisableBPFStats()` just sets it back to "0".

pkg/gadgets/top/ebpf/piditer/bpf/pid_iter.bpf.c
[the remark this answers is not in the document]
mauriciovasquezbernal: > Now that we have one, it might be better to entirely skip the ebpf iterator and fallback to the /proc/./fdinfo scanning when the address of bpf_prog_fops_addr can't be found.

 I agree. `getBpfProgFopsAddr()` can return an error, then it can be checked on `NewTracer()`.

pkg/bpfstats/bpfstats.go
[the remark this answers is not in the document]
blanquicet: Yes, but `refCnt` was already decreased [here](https://github.com/kinvolk/inspektor-gadget/blob/e1f99b807b8bf9e9010a48e3068357a1c0756786/pkg/bpfstats/bpfstats.go#L87). So, my question is, why to explicitly set `refCnt` to `0` only for `MethodSysctl`?

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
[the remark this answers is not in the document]
blanquicet: nit: IMO, it would be better to have all the clean-up stuff in the `stop()` 🙂

pkg/gadgets/top/ebpf/piditer/iter.go
blanquicet: Absolutely, it will be manage in another PR. I just wanted to mention it.

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
flyth: I moved the logic and implemented similarly to top file like you suggested.

pkg/bpfstats/bpfstats.go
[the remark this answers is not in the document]
flyth: Ah, sorry, I thought you were talking about the contents of /proc/.../bpf_stats_enabled. :)

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,339 @@
...
+					Name:      pi.Name,
+					Type:      pi.Type.String(),
+					Runtime:   int64(rt),
+					RunCount:  rc,
+				})
+
+				prog.Close()
+			}
```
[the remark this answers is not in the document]
flyth: Yeah, totally agree. Especially since introducing the fallback, this has grown. I have extracted the pid aggregation into separate functions - should be more readable now.

pkg/bpfstats/bpfstats.go
```diff
@@ -0,0 +1,122 @@
...
+			return fmt.Errorf("could not enable stat collection: %w", err)
+		}
+		method = MethodSysctl
+	} else {
+		statsSock = s
+		method = MethodBPFFunc
+	}
+
+	return nil
+}
```
mauriciovasquezbernal, line 80: What about only increasing `refCnt` after we're done enabling it? (to avoid having to reset it back to 0 wen there is an error)

```suggestion
func EnableBPFStats() error {
	mutex.Lock()
	defer mutex.Unlock()

	if refCnt != 0 {
		return nil
	}

	// Actually enable
	s, err := ebpf.EnableStats(unix.BPF_STATS_RUN_TIME)
	if err != nil {
		// Use fallback method
		err = ioutil.WriteFile("/proc/sys/kernel/bpf_stats_enabled", []byte("1"), 0o644)
		if err != nil {
			return fmt.Errorf("could not enable stat collection: %w", err)
		}
		method = MethodSysctl
	} else {
		statsSock = s
		method = MethodBPFFunc
	}

	refCnt++

	return nil
}
```

pkg/bpfstats/bpfstats.go
```diff
@@ -0,0 +1,122 @@
...
+
+// GetMethod returns the currently used method to enable stats collection. If
+// EnableBPFStats() has not yet been called, it will return MethodNone.
+func GetMethod() BPFStatsMethod {
+	mutex.Lock()
+	defer mutex.Unlock()
+	return method
+}
```
mauriciovasquezbernal, line 122: Is this function useful?

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,209 @@
...
+
+	statsCallback := func(stats []types.Stats) {
+		ev := types.Event{
+			Stats: stats,
+		}
+		ev.Node = trace.Spec.Node
+
+		r, err := json.Marshal(ev)
+		if err != nil {
+			log.Warnf("Gadget %s: Failed to marshall event: %s", trace.Spec.Gadget, err)
```
mauriciovasquezbernal, line 157: I know it's wrong on all other gadgets, but it's "marshal" (with a single `l`).

pkg/gadgets/top/ebpf/piditer/bpf/pid_iter.bpf.c
```diff
@@ -0,0 +1,38 @@
...
+
+	if (!file || !task)
+		return 0;
+
+	// We need to have an address of bpf_prog_fops to run
+	// TODO: Currently cilium/ebpf doesn't support .ksyms, this is why we get the info from userspace right now
+	if (bpf_prog_fops_addr == 0 || (__u64)(file->f_op) != bpf_prog_fops_addr)
+		return 0;
+
+	__builtin_memset(&e, 0, sizeof(e));
```
mauriciovasquezbernal, line 28: What's the difference compare to have `struct pid_iter_entry e = {};` on L18?

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -0,0 +1,299 @@
...
+
+var ebpfCmd = &cobra.Command{
+	Use:   fmt.Sprintf("ebpf [interval=%d]", types.IntervalDefault),
+	Short: "Periodically report ebpf runtime stats",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var err error
+
+		if params.NamespaceOverridden {
+			return commonutils.WrapInErrInvalidArg("--namespace / -n",
+				fmt.Errorf("this gadget can not filter by namespace"))
```
mauriciovasquezbernal, line 45: We're using `cannot` in almost all the places, so better to stick with it. :D

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+		if params.Podname != "" {
+			return commonutils.WrapInErrInvalidArg("--podname / -p",
+				fmt.Errorf("this gadget can not filter by pod name"))
+		}
+		if params.Containername != "" {
+			return commonutils.WrapInErrInvalidArg("--containername / -c",
+				fmt.Errorf("this gadget can not filter by container name"))
+		}
+		if params.LabelsRaw != "" || len(params.Labels) > 0 {
```
mauriciovasquezbernal, line 55: I think only checking `len(params.Labels)` should be enough.

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -0,0 +1,249 @@
...
+		if err != nil {
+			continue
+		}
+		pi, err := prog.Info()
+		if err != nil {
+			prog.Close()
+			continue
+		}
+		rt, _ := pi.Runtime()
+		rc, _ := pi.RunCount()
```
mauriciovasquezbernal, line 190: It seems this logic is wrong here as it'll get the numbers from the time `EnableBPFStats()` instead of the last interval. I think it's needed to keep a map with the id of the program and the values when the gadget started to be able to calculate the difference.

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -0,0 +1,299 @@
...
+	switch params.OutputMode {
+	case commonutils.OutputModeColumns:
+		if term.IsTerminal(int(os.Stdout.Fd())) {
+			utils.ClearScreen()
+		} else {
+			fmt.Println("")
+		}
+		fmt.Printf("%-16s %-8s %-16s %-16s %-7s %-20s %12s %8s\n",
+			"NODE",
+			"PROG",
```
mauriciovasquezbernal, line 165: `PROGID` or `ID`? (not sure, `PROG` seems not that bad too).

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -0,0 +1,299 @@
...
+		for idx, event := range stats {
+			if idx == maxRows {
+				break
+			}
+			pid := ""
+			comm := ""
+			if len(event.Pids) > 0 {
+				pid = strconv.Itoa(int(event.Pids[0].Pid))
+				comm = event.Pids[0].Comm
+			}
```
mauriciovasquezbernal, line 210: I don't know if this has been discussed before. It seems it's too misleading to only print a pid. Would it be possible to concatenate and print them? (or at least, to have something that makes it clear to the user that not all pids were printed).

mauriciovasquezbernal (commented):
It looking great. Thanks a lot for the code reorganization, looks much better!

Most of my comments are nits, the only important is about bad statistics being reported.

cmd/kubectl-gadget/top/ebpf.go
flyth: Hmm, true. I think we could just use a new line with only the pid/comm columns filled out so that you can directly see that it refers to the line above. It'll probably not happen often anyways that a program is referenced more than once and that way we can save space (width).

pkg/gadgets/top/ebpf/tracer/tracer.go
flyth: Hmm. Right now the runtime is used the same way that the system command `top` uses it - it shows the total cpu time - (but yes, in our case only since collecting the stats were enabled/when they were enabled).

But yeah, then, there's a difference: by default, the system command `top` sorts by current cpu usage and not the total cpu time. Current usage is a value that we currently don't have - and for that, as you said, we would need to store the old value in a map and compare. We could then show the time used in the current interval or - most likely taking into account the number of CPU cores - calculate the cpu percentage as well.

I'll look into it, if we want to have both values - but I wouldn't scratch the total runtime from the stats, as I think that it's more helpful to have the absolute numbers when you look at longer timespans.

pkg/bpfstats/bpfstats.go
flyth: Since it's a package that could be used from third parties as well, I thought it could be helpful to know which method is being used. I can remove it, if you don't think it is.

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -0,0 +1,249 @@
...
+			Comm: e.Comm,
+		})
+	}
+	return pidmap
+}
+
+func getPidMapFromProcFs() map[uint32][]*types.PidInfo {
+	pidmap := make(map[uint32][]*types.PidInfo)
+	processes, err := os.ReadDir("/proc/")
+	if err == nil {
```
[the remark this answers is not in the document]
flyth: Agree with the early return :)

About the error: as we're "just" enriching data here, I'm not too sure about what we should do when there's an error. If we're still getting all the other info, it might be worth to show it without the pid info. Maybe we should do the same with the `t.iter.DumpPids()` and ignore that error as well and try the fallback solution in case of an error.

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -0,0 +1,249 @@
...
+		pidmap = getPidMapFromPids(pids)
+	} else {
+		// Fallback...
+		pidmap = getPidMapFromProcFs()
+	}
+
+	for i := range stats {
+		if a, ok := pidmap[stats[i].ProgramID]; ok {
+			stats[i].Pids = a
+		}
```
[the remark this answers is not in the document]
flyth: `pids` would shadow the var from L164, that's probably why I left it anonymous. Will call it `tmpPids`, ok?

cmd/kubectl-gadget/top/ebpf.go
flyth: I just talked to @blanquicet about this. It would probably make sense to postpone adding the other pids until #782 is done - I'll open an issue for that, if that's okay.

cmd/kubectl-gadget/top/ebpf.go
flyth: Changed it to `PROGID` - and also the custom column name.

pkg/gadgets/top/ebpf/tracer/tracer.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Yes, `tmpPids` is better than `a`.

pkg/bpfstats/bpfstats.go
mauriciovasquezbernal: We can leave it there, I just wanted to be sure that it wasn't a leftover.

pkg/gadgets/top/ebpf/tracer/tracer.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I agree that the information produced in this step is "extra" information and not fundamental. My concern is how do we indicate that we failed to enrich the output, i.e., how can a user distinguish the case when no pids are using a program vs the case when we failed to enrich the data?  (I don't want to produce any misleading information). Maybe it's easier to fail in this case to avoid ambiguity?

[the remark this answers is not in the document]
mauriciovasquezbernal: I see. I'd propose to fail for the time being, then we can implement the mechanism you described. (I'd like to get this PR in, I think it's almost complete and we can make other iterations later on to improve it, otherwise it becomes too complicated to review it).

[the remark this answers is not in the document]
blanquicet: One option could be to make `getPidMapFromProcFs` a `Tracer`'s method so it can call `errorCallback`. Of course, this proposal can be explored in another PR.

pkg/gadgets/top/ebpf/tracer/tracer.go
flyth: I left the values for the total numbers in the event/stats, but removed them from the default output. Instead it now uses the "last interval" values for everything by default.

cmd/kubectl-gadget/top/ebpf.go
flyth: Will be handled in #793

pkg/gadgets/top/ebpf/piditer/bpf/pid_iter.bpf.c
flyth: I guess not zeroing when returning early (whenever it is not an ebpf program), which happens a lot more often than not.

pkg/gadgets/top/ebpf/tracer/tracer.go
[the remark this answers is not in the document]
flyth: We can discuss further in #794.

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -0,0 +1,288 @@
...
+
+		// calculate delta, if possible
+		if oldrt, ok := t.prevRuntimes[uint32(curID)]; ok {
+			curRuntime = int64(totalRuntime) - oldrt
+		}
+		if oldctr, ok := t.prevRunCounts[uint32(curID)]; ok {
+			curRunCount = totalRunCount - oldctr
+		}
+
+		curRuntimes[uint32(curID)] = int64(totalRuntime)
```
mauriciovasquezbernal, line 220: I'm a bit concerned that the ID of a program can be reused and it leads to a wrong result. What about using the tag to be sure the statistics are from the same program? (It's already a string, so concatenating it somehow with the programID should be straightforward.

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
```diff
@@ -0,0 +1,209 @@
...
+		if err != nil {
+			log.Warnf("Gadget %s: Failed to marshal event: %s", trace.Spec.Gadget, err)
+			return
+		}
+		t.helpers.PublishEvent(t.traceName, string(r))
+	}
+
+	errorCallback := func(err error) {
+		ev := types.Event{
+			Error: fmt.Sprintf("Gadget failed with: %v", err),
```
mauriciovasquezbernal, line 165: You can directly use `eventtypes.Err(fmt.Sprintf("Gadget failed with: %v", err))` . -> https://github.com/kinvolk/inspektor-gadget/blob/v0.7.0/pkg/types/types.go#L77

docs/guides/top/ebpf.md
```diff
@@ -0,0 +1,74 @@
...
+...
+```
+
+The output should already contain the `top ebpf` gadget itself.
+
+Now, in a second terminal, start `top file` on the _gadget_ namespace to get some output.
+
+```bash
+$ kubectl gadget top file -n gadget
+NODE             PROGID   TYPE             NAME             PID     COMM                      RUNTIME   RUNCOUNT
```
mauriciovasquezbernal, line 30: Remove this line?

README.md
```diff
@@ -138,6 +138,7 @@ Usage:
 
 Available Commands:
   block-io    Periodically report block device I/O activity
+  ebpf        Periodically report ebpf runtime stats
```
mauriciovasquezbernal, line 141: Please also add an entry pointing to the guide around L24 of this file.

mauriciovasquezbernal (commented):
I think it's almost ready to go. Just two small nits on the code.

Could you also please squash all the commits?

Thanks!

pkg/gadgets/top/ebpf/tracer/tracer.go
flyth: The previous runtimes/run counts are only valid for the next call of nextStats(), because I had the same concern. But I'll add the tag as well.

pkg/gadget-collection/gadgets/top/ebpf/gadget.go
flyth: This is the same way it's used in the other gadgets (copied it from there -> https://github.com/kinvolk/inspektor-gadget/blob/v0.7.0/pkg/gadget-collection/gadgets/top/block-io/gadget.go#L167). Maybe it makes more sense to change it later everywhere in one go.

pkg/gadgets/top/ebpf/types/types.go
```diff
@@ -0,0 +1,126 @@
...
+			return a.ProgramID > b.ProgramID
+		default:
+			return a.CurrentRuntime > b.CurrentRuntime && a.CurrentRunCount > b.CurrentRunCount
+		}
+	})
+}
+
+type Event struct {
+	eventtypes.Event
+	Error string  `json:"error,omitempty"`
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think the mistake here is that this structure should not extend `eventtypes.Event`, this is what we're doing in other gadgets:

https://github.com/kinvolk/inspektor-gadget/blob/v0.7.0/pkg/gadgets/top/file/types/types.go#L73-L82
https://github.com/kinvolk/inspektor-gadget/blob/v0.7.0/pkg/gadgets/top/tcp/types/types.go#L83-L92

mauriciovasquezbernal (approved):
LGTM. 

Thanks a lot @flyth for handling all the comments, impressive work!

flyth:
Thanks a lot for all the reviews and support!

PR #755, opened 2022-07-18
Use k8s golang API to deploy Inspektor Gadget

eiffel-fl:
Hi.

This PR uses k8s golang API to deploy Inspektor Gadget instead of piping to `kubectl apply -f -`.

Best regards.

### DIFF
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -65,6 +65,7 @@ var (
 	deployTimeout             time.Duration
 	fallbackPodInformer       bool
 	printOnly                 bool
+	quiet                     bool
 	wait                      bool
 )
 
@@ -114,7 +115,19 @@ func init() {
 		"timeout", "",
 		120*time.Second,
 		"timeout for deployment")
+	deployCmd.PersistentFlags().BoolVarP(
+		&quiet,
+		"quiet", "q",
+		false,
+		"supress information messages")
 	rootCmd.AddCommand(deployCmd)
+}
+
+func info(format string, args ...any) {
+	if quiet {
+		return
+	}
+	fmt.Printf(format, args...)
 }
 
 // parseK8sYaml parses a k8s YAML deployment file content and returns the
@@ -192,6 +205,7 @@ func createResource(client dynamic.Interface, mapper meta.RESTMapper, object run
 		dynamicInterface = client.Resource(mapping.Resource)
 	}
 
+	info("Creating %s/%s...\n", unstruct.GetKind(), unstruct.GetName())
 	_, err = dynamicInterface.Create(context.TODO(), unstruct, metav1.CreateOptions{})
 	if err != nil {
 		return fmt.Errorf("failed to create %q: %w", groupVersionKind.Kind, err)
@@ -289,7 +303,7 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	fmt.Println("Waiting for gadget pod(s) to be ready...")
+	info("Waiting for gadget pod(s) to be ready...\n")
 	k8sClient, err := k8sutil.NewClientsetFromConfigFlags(utils.KubernetesConfigFlags)
 	if err != nil {
 		return utils.WrapInErrSetupK8sClient(err)
@@ -323,7 +337,7 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 			daemonSet, _ := event.Object.(*appsv1.DaemonSet)
 			status := daemonSet.Status
 
-			fmt.Printf("%d/%d gadget pod(s) ready\n", status.NumberReady, status.DesiredNumberScheduled)
+			info("%d/%d gadget pod(s) ready\n", status.NumberReady, status.DesiredNumberScheduled)
 
 			return status.DesiredNumberScheduled == status.NumberReady, nil
 		case watch.Error:
--- a/integration/command.go
+++ b/integration/command.go
@@ -72,6 +72,7 @@ type command struct {
 var deployInspektorGadget *command = &command{
 	name:           "DeployInspektorGadget",
 	cmd:            "$KUBECTL_GADGET deploy $GADGET_IMAGE_FLAG",
+	expectedRegexp: `\d\/\d gadget pod\(s\) ready`,
 }
 
 var waitUntilInspektorGadgetPodsDeployed *command = &command{


cmd/kubectl-gadget/deploy.go
```diff
@@ -362,35 +192,105 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+				case "INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER":
+					gadgetContainer.Env[i].Value = strconv.FormatBool(fallbackPodInformer)
+				}
+			}
+
+			err = createResource(dynamicClient, mapper, object)
+			if err != nil {
+				return err
+			}
+		}
```
mauriciovasquezbernal, line 293: What about something like: 

```suggestion
	for _, object := range objects {
		if daemonSet, ok := object.(*appsv1.DaemonSet); ok {
			gadgetContainer := &daemonSet.Spec.Template.Spec.Containers[0]

			gadgetContainer.Image = image

			policy, err := stringToPullPolicy(imagePullPolicy)
			if err != nil {
				return err
			}
			gadgetContainer.ImagePullPolicy = policy

			if !livenessProbe {
				gadgetContainer.LivenessProbe = nil
			}

			for i, _ := range gadgetContainer.Env {
				switch gadgetContainer.Env[i].Name {
				case "GADGET_IMAGE":
					gadgetContainer.Env[i].Value = image
				case "INSPEKTOR_GADGET_VERSION":
					gadgetContainer.Env[i].Value = version
				case "INSPEKTOR_GADGET_OPTION_HOOK_MODE":
					gadgetContainer.Env[i].Value = hookMode
				case "INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER":
					gadgetContainer.Env[i].Value = strconv.FormatBool(fallbackPodInformer)
				}
			}
		}

		err = createResource(dynamicClient, mapper, object)
		if err != nil {
			return err
		}
	}
```

pkg/resources/gadget/gadget.yaml
```diff
@@ -0,0 +1,262 @@
+apiVersion: v1
```
mauriciovasquezbernal, line 1: I'm not sure it's the best place for this file (same for already existing files). I think we should move the whole `pkg/resources` to a `manifests` folder or similar. (I think we don't expected external users to import this golang package). 

@alban what do you think?

mauriciovasquezbernal (commented):
It's looking great. Should it also include logic to wait until the gadget pods are running?, otherwise the command will return immediately but it won't be possible to use kubectl-gadget right away as the pod could still be being created.

pkg/resources/gadget/gadget.yaml
alban: I think it is fine to keep it in `pkg/resources`, to keep it simple with `pkg/resources/embed.go`.

It could be called `pkg/resources/manifests/gadget.yaml`

Makefile
```diff
@@ -195,9 +195,7 @@ minikube-install: gadget-default-container kubectl-gadget
 	./kubectl-gadget undeploy || true
 	./kubectl-gadget deploy --hook-mode=auto \
 		--liveness-probe=$(LIVENESS_PROBE) \
-		--image-pull-policy=Never | \
-		sed 's/initialDelaySeconds: [0-9]*/initialDelaySeconds: '$(LIVENESS_PROBE_INITIAL_DELAY_SECONDS)'/g' | \
-		kubectl apply -f -
+		--image-pull-policy=Never
```
alban, line 198: Since you removed `LIVENESS_PROBE_INITIAL_DELAY_SECONDS`, you can remove it line 175 too.

cmd/kubectl-gadget/deploy.go
```diff
@@ -362,35 +196,70 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("invalid argument %q for --hook-mode=[auto,crio,podinformer,nri,fanotify]", hookMode)
 	}
 
-	t, err := template.New("deploy.yaml").Parse(deployYamlTmpl)
+	objects, err := parseK8sYaml(resources.GadgetDeployment)
 	if err != nil {
-		return fmt.Errorf("failed to parse template: %w", err)
+		return err
```
alban, line 201: I am not sure about setting the parameters on the objects instead of the yaml text before it is parsed. It will make it more difficult when we want to use `kubectl gadget deploy --dry-run` to output the yaml with the substituted `image:` and other parameters, so that users can see what they are applying before doing so.

I would like `.github/workflows/inspektor-gadget.yml` to add a release artifact so users or third party software can install it via:
```
$ kubectl apply -f https://github.com/kinvolk/inspektor-gadget/releases/download/v0.7.0/gadget-deployment.yaml
```
So that it would be possible to write a Headlamp plugin to install Inspektor Gadget without requiring the dependency on the kubectl-gadget binary.

And to make it possible to write a Gardener extension (see [example of Helm chart for a Gardener extension](https://github.com/gardener/gardener-extension-networking-calico/blob/master/charts/gardener-extension-networking-calico/templates/deployment.yaml)). We don't have to prepare Helm charts in this PR of course, but I'm wondering if this PR is making it more difficult.

What do you think?

alban: > > I am not sure about setting the parameters on the objects instead of the yaml text before it is parsed.
> 
> I was pragmatic and it is easier to to set the parameters on the objects instead of dealing with strings.
> I can indeed convert the `string` to sort of YAML struct then set the parameters and convert it again to string but I find this operation heavy.

The double conversion would lose the comments too.

I was suggesting to apply the parameters on the Go template like it was done before, and then convert the yaml text with the parameters to objects. In that way, we can have the yaml text with the comments and with the parameters.

> > It will make it more difficult when we want to use kubectl gadget deploy --dry-run to output the yaml with the substituted image: and other parameters, so that users can see what they are applying before doing so.
> 
> I do not understand this point.
> There is no `--dry-run` option and the `kubectl gadget deploy` does not print anything with this modification.
> Can you please clarify it?

I was suggesting that this PR should implement this `--dry-run` option. Today, users can look at the yaml to see what they are about to deploy. With this PR, users can no longer do that. So adding `--dry-run` option would resolve this.

> > I would like .github/workflows/inspektor-gadget.yml to add a release artifact so users or third party software can install it via:
> 
> I think this the goal of `pkg/resources/manifests/deploy.yaml`, people can just apply it and there a ready to play with Inspektor Gadget.

But this file has:
```
        image: "ghcr.io/kinvolk/gadget"
```
without the version attached to it. So users will have to edit the file to write the version they want to use, and this is not documented. If the yaml was attached as a release artifact, they can just use it with `kubectl apply -f https://...` without downloading it and fixing it first.

> Note that this file only deals with k8s resources related to Inspektor Gadget and does not deal with trace CRD.

Understood.

alban: > I am really not sure to understand.
> Do you want to keep the `| kubectl apply -f -` to deploy Inspektor Gadget?
> I really want to get rid of it and I think getting rid of it would mean to remove the Go template.

I don't want to keep the `| kubectl apply -f -` but to have an option for users to double check what is going to be deployed, if they want to. I imagine that on production systems, one would want to inspect the yaml. I would like to be able to:
1. `kubectl gadget deploy --print-only` (or `--dry-run`, naming to be decided) to print the yaml on stdout so I can check if it looks ok.
2. `kubectl gadget deploy` to install Inspektor Gadget on Kubernetes.

Before this PR, it was possible. I just don't want to lose this feature.

> As a consequence, people who want to install Inspektor Gadget with only YAML files will need to do the following:
> 
> ```bash
> # Deploy everything related to k8s
> $ kubectl apply -f .../deploy.yaml
> # Deploy everything related to trace CRD
> $ kubectl apply -f .../gadget.kinvolk.io_traces.yaml
> ```
> 
> And `gadget.kinvolk.io_traces.yaml` is not published as release artifact somewhere.

Oh, I didn't understand actually. I guess gadget.kinvolk.io_traces.yaml could be added in the release artifact? But it does not have parameters, so not sure if it's necessary.

mauriciovasquezbernal: > without the version attached to it. So users will have to edit the file to write the version they want to use, and this is not documented. If the yaml was attached as a release artifact, they can just use it with kubectl apply -f https://... without downloading it and fixing it first.

What about updating those files before tagging a release to make sure they point to the right image?, then those files could be used directly to install the right version:

```
$ kubectl apply -f https://raw.githubusercontent.com/kinvolk/inspektor-gadget/main/manifests/gadget.yaml
--> installs latest version

$ kubectl apply -f https://raw.githubusercontent.com/kinvolk/inspektor-gadget/v0.7.0/manifests/gadget.yaml
--> installs v0.7.0

$ cat https://raw.githubusercontent.com/kinvolk/inspektor-gadget/v0.7.0/manifests/gadget.yaml | grep image 
        image: "ghcr.io/kinvolk/inspektor-gadget:v0.7.0"
```

mauriciovasquezbernal:
> I can add an option to wait until the gadget pod is ready but I prefer to not make it the default (for the moment).

`kubectl gadget undeploy` is already waiting for the resources to be removed. IMO having `kubectl gadget deploy` waiting by default is better, in the end it's the most common case, you deploy it and then you want to use it afertwards.

cmd/kubectl-gadget/deploy.go
alban: > What about updating those files before tagging a release to make sure they point to the right image?, then those files could be used directly to install the right version:

It's a bit more complicated to make a release because we can't just git-tag the last commit on `main` but we need to create a commit with the version and then revert it... Do you suggest to create those commits manually, with a script or with a GitHub Action somehow?

pkg/resources/gadget/gadget.yaml
alban: Either way works for me. Whatever is the easiest to implement :)

cmd/kubectl-gadget/deploy.go
```diff
@@ -64,293 +89,115 @@ func init() {
...
-`
-
-type parameters struct {
-	Image               string
-	ImagePullPolicy     string
-	Version             string
-	HookMode            string
-	LivenessProbe       bool
-	FallbackPodInformer bool
+// parseK8sYaml parses a k8s YAML deployement file content and returns the
```
blanquicet, line 115:
```suggestion
// parseK8sYaml parses a k8s YAML deployment file content and returns the
```

cmd/kubectl-gadget/deploy.go
```diff
@@ -64,293 +89,115 @@ func init() {
...
+// corresponding objects.
+// It was adapted from:
+// https://github.com/kubernetes/client-go/issues/193#issuecomment-363318588
+func parseK8sYaml(content string) ([]runtime.Object, error) {
+	acceptedK8sTypes := regexp.MustCompile(`(Namespace|ServiceAccount|Role|RoleBinding|ClusterRole|ClusterRoleBinding|DaemonSet|CustomResourceDefinition)`)
```
blanquicet, line 120: nit: Do we really need to check this? In the end, we are parsing our own YAML. Additionally, every time we add a new resource, we will need to update this line.

cmd/kubectl-gadget/deploy.go
```diff
@@ -64,293 +89,115 @@ func init() {
...
+	case "IfNotPresent":
+		return v1.PullIfNotPresent, nil
+	default:
+		return "", fmt.Errorf("invalid pull policy: %s. Possible values are [Always, Never, IfNotPresent]",
+			imagePullPolicy)
+	}
+}
+
+// createResource creates the resource corresponding to the object given as
+// parameter usinga dynamic client an RESTMapper to get the corresponding
```
blanquicet, line 170:
```suggestion
// parameter using a dynamic client and a RESTMapper to get the corresponding
```

cmd/kubectl-gadget/deploy.go
```diff
@@ -362,35 +209,140 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+
+				if v1pod.IsPodReady(pod) {
+					readyPods++
+				}
+
+				// We cannot list the pods after we asked for their creation because the
+				// list could be empty.
+				// But, we deploy one pod per node, so we return succesfully when the
+				// number of ready pods equals the number of nodes.
+				return len(nodes.Items) == readyPods, nil
```
blanquicet, line 333: I think the advantage of watching each pod is to have access to the progress. Why we are not showing it as we do for the `undeploy` command? I expect the output to be similar to the events:

```bash
$ k alpha events -n gadget pod gadget
LAST SEEN   TYPE     REASON             OBJECT             MESSAGE
5s          Normal   SuccessfulCreate   DaemonSet/gadget   Created pod: gadget-kfxwk
5s          Normal   SuccessfulCreate   DaemonSet/gadget   Created pod: gadget-pncg6
4s          Normal   Scheduled          Pod/gadget-kfxwk   Successfully assigned gadget/gadget-kfxwk to master
4s          Normal   Scheduled          Pod/gadget-pncg6   Successfully assigned gadget/gadget-pncg6 to worker
4s          Normal   Pulling            Pod/gadget-kfxwk   Pulling image "192.168.1.105:5000/gadget:francis-new-deploy"
4s          Normal   Pulled             Pod/gadget-kfxwk   Successfully pulled image "192.168.1.105:5000/gadget:francis-new-deploy" in 12.511295ms
4s          Normal   Created            Pod/gadget-kfxwk   Created container gadget
4s          Normal   Pulling            Pod/gadget-pncg6   Pulling image "192.168.1.105:5000/gadget:francis-new-deploy"
4s          Normal   Pulled             Pod/gadget-pncg6   Successfully pulled image "192.168.1.105:5000/gadget:francis-new-deploy" in 15.011706ms
4s          Normal   Created            Pod/gadget-pncg6   Created container gadget
4s          Normal   Started            Pod/gadget-pncg6   Started container gadget
4s          Normal   Started            Pod/gadget-kfxwk   Started container gadget
```

If we decide to not do that, I think it would be easier to verify the status of the daemonset instead of every single pod? I expect it to be simpler, and we could avoid relying on the number of nodes.

It would be the same as the `rollout status` command:

```bash
$ ./kubectl-gadget deploy --deploy-wait=false

$ kubectl rollout status -n gadget ds gadget --timeout=30s
Waiting for daemon set "gadget" rollout to finish: 0 of 2 updated pods are available...
Waiting for daemon set "gadget" rollout to finish: 1 of 2 updated pods are available...
daemon set "gadget" successfully rolled out
```

However, I am not sure if it verifies if the pods are ready or only if they were scheduled. 

cc @alban @mauriciovasquezbernal

blanquicet (commented):
I think it looks great. The code is very readable and easy to follow! 

I just would like to add some "in-progress" messages to let the user knows what is happening. WDYT?

cmd/kubectl-gadget/deploy.go
```diff
@@ -64,293 +89,115 @@ func init() {
...
-		&file,
-		"file", "",
-		"",
-		"file to save the generated yaml template")
+		"use pod informer as a fallback for the main hook")
+	deployCmd.PersistentFlags().BoolVarP(
+		&printOnly,
+		"print-only", "",
+		false,
+		"only print YAML of resources")
```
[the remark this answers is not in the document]
blanquicet: You are missing the `-o yaml` flag:

```bash
$ kubectl create --dry-run=client namespace foo -o yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: null
  name: foo
spec: {}
status: {}
```

cmd/kubectl-gadget/deploy.go
blanquicet: > I prefer to keep the command silent and only prints if something went wrong.
>
Actually, we agreed the opposite in our team meeting: https://github.com/kinvolk/inspektor-gadget/issues/123#issuecomment-1068950640. In addition, this is what `undeploy` is doing thus for consistency, it should do the same.

cmd/kubectl-gadget/deploy.go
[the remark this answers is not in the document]
blanquicet: Yeah, I agree. I let you decide 🙂

cmd/kubectl-gadget/deploy.go
```diff
@@ -64,293 +89,115 @@ func init() {
...
+	deployCmd.PersistentFlags().BoolVarP(
+		&deployWait,
+		"deploy-wait", "",
+		true,
+		"wait for gadget pod to be ready")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: `kubectl` also uses `--wait`, so I'd stick to it.

cmd/kubectl-gadget/deploy.go
```diff
@@ -362,35 +209,140 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+	objects = append(objects, traceObjects...)
 
-	if file != "" {
-		f, err := os.OpenFile(file, os.O_CREATE|os.O_WRONLY, 0o644)
-		if err != nil {
-			return fmt.Errorf("failed to create file: %w", err)
-		}
-		defer f.Close()
-		w = f
+	config, err := utils.KubernetesConfigFlags.ToRESTConfig()
```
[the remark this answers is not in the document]
mauriciovasquezbernal: `rest.InClusterConfig()` (as stated in the name) is intended to be used inside the cluster, i.e. in a pod, in that case it uses the secrets mount on the pod and some env variables to authenticate against the api server.

cmd/kubectl-gadget/deploy.go
```diff
@@ -362,36 +203,129 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			options.LabelSelector = "k8s-app=gadget"
+
+			return daemonSetInterface.Watch(context.TODO(), options)
+		},
+	}
+
+	ctx, cancel := watchtools.ContextWithOptionalTimeout(context.TODO(), time.Duration(timeout)*time.Second)
+	defer cancel()
```
mauriciovasquezbernal, line 310: It's using the timeout from the undeploy command that is 30 seconds, I think it's too short for creation. Maybe 1 or 2 minutes work better?

cmd/kubectl-gadget/deploy.go
blanquicet: Ok. It means we need another issue to track what this PR is not managing from what was discussed in https://github.com/kinvolk/inspektor-gadget/issues/123.

pkg/resources/manifests/deploy.yaml
```diff
@@ -0,0 +1,262 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
mauriciovasquezbernal: What about the users that want to deploy Inspektor Gadget directly with kubectl? We're forcing them to deploy two different yamls.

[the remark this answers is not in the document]
mauriciovasquezbernal: > A middle ground could be to release one big file, which is basically the concatenation of the two files with some modifications (_e.g._ hardcoding the version of Inspektor Gadget docker image to deploy). It permits users to have only file to apply and for us we still have the two files which simplify future modifications. What do you think?

What do you mean to release? If you mean to include as a release artifact I think it's too complicated. The resulting file or merging those two is about 300 lines long, it seems to me it's not complicated to handle.

cmd/kubectl-gadget/deploy.go
```diff
@@ -362,36 +209,131 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
+			return daemonSetInterface.List(context.TODO(), options)
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			options.LabelSelector = "k8s-app=gadget"
+
+			return daemonSetInterface.Watch(context.TODO(), options)
+		},
+	}
+
+	ctx, cancel := watchtools.ContextWithOptionalTimeout(context.TODO(), time.Duration(deployTimeout)*time.Second)
```
blanquicet, line 315: I am wondering if we should undo (call `undeploy`) to clean up things when the timeout is reached. If we decide to do so, we should manage it in another PR. This one is ready.

blanquicet (commented):
Just two concerns, but I think it is ready to be merged! 🎊

blanquicet (approved):
Oops, my previous review was intended to be an "Approval" 😅

cmd/kubectl-gadget/deploy.go
```diff
@@ -64,293 +89,115 @@ func init() {
...
+
+	unstruct := &unstructured.Unstructured{Object: unstructuredObj}
+
+	var dynamicInterface dynamic.ResourceInterface
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		dynamicInterface = client.Resource(mapping.Resource).Namespace(unstruct.GetNamespace())
+	} else {
+		dynamicInterface = client.Resource(mapping.Resource)
+	}
+
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Sounds like a good idea to me.

cmd/kubectl-gadget/deploy.go
mauriciovasquezbernal: It's a good question. The same applies when there is an error deploying, should we try to revert the changes? I think just letting the user know there was an error is good enough for the time being.

PR #757, opened 2022-07-19
cmd/kubectl-gadget: Remove requestedColumns parameter from TransformEvent()

mauriciovasquezbernal:
The implementations of TraceParser already have this information when
they are instantiated. (BaseParser on this case). Hence, remove that
parameter to make this interface simpler.


blanquicet (approved):
LGTM!

I added this optimization to snapshot gadgets but I forgot to do it also for trace gadgets. Thanks!

PR #758, opened 2022-07-20
Use UntilWithSync() instead of untilWithoutRetry() for traces watching

eiffel-fl:
Hi.

In this PR, I replaced `untilWithoutRetry()` by `UntilWithSync()`.
So, I was able to delete the local copy of `untilWithoutRetry()` code and we do not need to worry about it being [made private](https://github.com/kubernetes/client-go/blob/c6bd30b9ec5f668df191bc268c6f550c37726edb/tools/watch/until.go#L55) one day.

Best regards.

### DIFF
--- a/cmd/kubectl-gadget/utils/trace.go
+++ b/cmd/kubectl-gadget/utils/trace.go
@@ -34,8 +34,9 @@ import (
 	types "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 
-	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/tools/cache"
 	watchtools "k8s.io/client-go/tools/watch"
 
 	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/utils/trace.go
```diff
@@ -507,23 +469,23 @@ func waitForCondition(traceID string, conditionFunction func(*gadgetv1alpha1.Tra
...
 		if err != nil {
 			return nil, err
 		}
 
 		ctx, cancel := watchtools.ContextWithOptionalTimeout(context.Background(), TraceTimeout)
-		_, err = untilWithoutRetry(ctx, watcher, func(event watch.Event) (bool, error) {
+		defer cancel()
+
+		createdTraces := 0
+		_, err = watchtools.UntilWithSync(ctx, traceListerWatcher, &gadgetv1alpha1.Trace{}, nil, func(event watch.Event) (bool, error) {
```
blanquicet, line 488: Why we were not using `UntilWithSync` from the very beginning?

cmd/kubectl-gadget/utils/trace.go
```diff
@@ -507,23 +469,23 @@ func waitForCondition(traceID string, conditionFunction func(*gadgetv1alpha1.Tra
...
+		// We will need to list and watch events on them.
+		// For this, we will get a ListerWatcher on all the traces which share the
+		// same ID.
+		// NOTE all the traces on different nodes but linked to one gadget share
 		// the same ID.
 		// We will also begin to monitor events since the above GET of the traces
-		// list.
-		watcher, err = getTraceWatcher(traceID, traceList.ListMeta.ResourceVersion)
+		// list thanks to resource version.
+		traceListerWatcher, err := getTraceListerWatcher(traceID, traceList.Items[tracesNumber-1].ResourceVersion)
```
blanquicet, line 479: What is the difference between `traceList.ListMeta.ResourceVersion` and `traceList.Items[tracesNumber-1].ResourceVersion`?

blanquicet (commented):
I'm still not sure about this PR. Could you please describe why this change in needed in the PR's description? I think I am missing some context 🙂 .

cmd/kubectl-gadget/utils/trace.go
blanquicet: It is clear now.

blanquicet (approved):
I tested it, and it is working correctly. This PR only modifies the logic for the `Added` case, which, in my opinion, is not necessary to manage.

Let's see if someone else has an opinion about it, so please don't merge it yet.

mauriciovasquezbernal (approved):
LGTM. I'm also not sure if handling the added event is needed, however it doesn't hurt to have this code there.

PR #759, opened 2022-07-21
go.mod: update cilium/ebpf and remove GC workaround

alban:
cilium/ebpf v0.9.1 has the following fix:
https://github.com/cilium/ebpf/commit/54f34552d7e3ddc690722fd1f1dcc7685962c216

So the ebpf file descriptors are correctly closed and it is no longer
necessary to call runtime.GC() explicitly.


mauriciovasquezbernal (approved):
LGTM!

flyth (approved):
Yup, LGTM!

PR #761, opened 2022-07-22
dockerignore: Add one dockerignore per Dockerfile.

eiffel-fl:
Hi.

The idea behind this PR is to not store the `cmd` directory into the gadget images because they do not need this directory for their build process.
As a consequence, it is possible to modify CLI code without having to build again the image.

Sadly, modifying `.dockerignore` will impact all the Dockerfiles and `local-gadget.Dockerfile` needs this directory.
The only solution is to add one `*.Dockerfile.dockerignore` for each `*.Dockerfile` which suits the corresponding `Dockerfile`.

Best regards.


mauriciovasquezbernal (approved):
Approving given that it's not a big change and it should not require a lot of maintenance.

PR #762, opened 2022-07-25
cmd/common: Merge `BaseSnapshot` and `commonutils.BaseParser`

blanquicet:
# Merge `BaseSnapshot` and `commonutils.BaseParser`

When working on https://github.com/kinvolk/inspektor-gadget/pull/763, I wanted to reuse the logic already implemented for snapshot gadgets to print columns separated by taps. Therefore, I thought to move the `BaseSnapshot` implementation to the common utils library. However, once there, I realized that it was pretty similar to the already existing `BaseParser` used by other gadgets (trace category and profile/cpu gadget). The only difference is that one uses tabs and the other uses a predefined width. Therefore, I decided to [merge them](https://github.com/kinvolk/inspektor-gadget/blob/661aab572fa2554d3b158d9bac6561cba41a095b/cmd/common/utils/parser.go#L22-L36) and add a `SeparateWithTabs` which allows specifying whether the parser has to print using tabs or the predefined width.

## How to use

Tests gadgets using `BaseParser` (profile/cpu and all the traces) and the gadgets that were previously using the `BaseSnapshot` (snapshot category).

### DIFF
--- a/cmd/kubectl-gadget/snapshot/process.go
+++ b/cmd/kubectl-gadget/snapshot/process.go
@@ -41,14 +41,14 @@ func newProcessCmd() *cobra.Command {
 		},
 	}
 
-	availableColumns := map[string]struct{}{
-		"node":      {},
-		"namespace": {},
-		"pod":       {},
-		"container": {},
-		"comm":      {},
-		"tgid":      {},
-		"pid":       {},
+	availableColumns := []string{
+		"node",
+		"namespace",
+		"pod",
+		"container",
+		"comm",
+		"tgid",
+		"pid",
 	}
 
 	customRun := func(callback func(traceOutputMode string, results []string) error) error {
--- a/cmd/common/snapshot/process.go
+++ b/cmd/common/snapshot/process.go
@@ -33,23 +33,20 @@ type ProcessFlags struct {
 }
 
 type ProcessParser struct {
-	BaseSnapshotParser
+	commonutils.BaseParser
 
 	processFlags *ProcessFlags
 }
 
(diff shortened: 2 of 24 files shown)


cmd/kubectl-gadget/snapshot/process.go
```diff
@@ -41,14 +41,14 @@ func newProcessCmd() *cobra.Command {
...
 
-	availableColumns := map[string]struct{}{
-		"node":      {},
-		"namespace": {},
-		"pod":       {},
-		"container": {},
-		"comm":      {},
-		"tgid":      {},
-		"pid":       {},
+	availableColumns := map[string]int{
```
mauriciovasquezbernal, line 44: Maybe add a comment indicating why to use 0 as width?

mauriciovasquezbernal (approved):
LGTM. I think commits should be squashed.

cmd/kubectl-gadget/snapshot/process.go
blanquicet: I created two wrapper functions to simplify the parser API and keep this logic of width/taps inside the parser, and it is transparent for the gadgets:

https://github.com/kinvolk/inspektor-gadget/blob/f1870b7f3c3a6f61cba3d48a00286632d16910d9/cmd/common/utils/parser.go#L46-L65

WDYT?

mauriciovasquezbernal: I like this idea of having to different functions, hence we don't need to pass "true/false" from all the callers.

PR #763, opened 2022-07-25
cmd/local-gadget: Enrich list-containers with ID and Runtime information

blanquicet:
# Enrich `list-containers` with ID and Runtime information

This PR enriches the `list-containers` command to show the ID and the name of the runtime managing the container:

```bash
$ sudo local-gadget list-containers
RUNTIME       ID               NAME
containerd    7766d32caded4    calico-kube-controllers
containerd    2e3e4968b456f    calico-node
containerd    d3be7741b94ff    coredns
containerd    e7be3e4dc1bb4    coredns
containerd    fb4fe41921f30    etcd
containerd    136e7944d2077    kube-apiserver
containerd    ad8709a2c2ded    kube-controller-manager
containerd    66cf05654a47f    kube-proxy
containerd    a68bed42aa6b2    kube-scheduler
docker        95b814bb82b9e    myContainer
```

The flag `--no-trunc` was also implemented:

```bash
$ sudo local-gadget list-containers --no-trunc
RUNTIME       ID                                                                  NAME
containerd    34e68b2dd2f9f1ac516ec695de79f2a6489213c54131e3de20128bd94251e5f7    calico-kube-controllers
containerd    b66b5c2a756fc48a3056b8754704121bd4f1f71ee85d609b920c13a7995796e9    calico-node
...
```

Notice that now that the `list-containers` uses the `BaseParser`, it also supports JSON and custom-columns output mode. In addition, it supports the `--containername` flag.

Before reviewing this PR, please review https://github.com/kinvolk/inspektor-gadget/pull/762 as it introduces changes used by this PR.

### DIFF
--- a/cmd/local-gadget/containers/containers.go
+++ b/cmd/local-gadget/containers/containers.go
@@ -15,34 +15,177 @@
 package containers
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
 
-	"github.com/kinvolk/inspektor-gadget/cmd/local-gadget/utils"
-	localgadgetmanager "github.com/kinvolk/inspektor-gadget/pkg/local-gadget-manager"
 	"github.com/spf13/cobra"
+
+	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
+	"github.com/kinvolk/inspektor-gadget/cmd/local-gadget/utils"
+	containercollection "github.com/kinvolk/inspektor-gadget/pkg/container-collection"
+	localgadgetmanager "github.com/kinvolk/inspektor-gadget/pkg/local-gadget-manager"
 )
 
+type ContainerFlags struct {
+	noTrunc bool
+}
+
+type ContainerParser struct {
+	commonutils.BaseParser
+
+	containerFlags *ContainerFlags
+}
+
 func NewListContainersCmd() *cobra.Command {
-	var commonFlags utils.CommonFlags
+	var containerFlags ContainerFlags
+
+	availableColumns := []string{
+		"runtime",
+		"id",
+		"name",
+		"pid",
+		"mntns",
+		"netns",
+		"namespace",
+		"podname",
+		"poduid",
+		"cgrouppath",
+		"cgroupid",
+		"cgroupv1",
+		"cgroupv2",
+	}
+
+	commonFlags := &utils.CommonFlags{
+		OutputConfig: commonutils.OutputConfig{
+			// The columns that will be used in case the user does not specify
+			// which specific columns they want to print.
+			CustomColumns: []string{
+				"runtime",
+				"id",
+				"name",
+			},
+		},
+	}
 
 	cmd := &cobra.Command{
 		Use:   "list-containers",
 		Short: "List all containers",
-		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


cmd/local-gadget/containers/containers.go
```diff
@@ -15,34 +15,145 @@
...
+
 func NewListContainersCmd() *cobra.Command {
-	var commonFlags utils.CommonFlags
+	var containerFlags ContainerFlags
+
+	availableColumns := map[string]int{
+		"runtime": 0,
+		"id":      0,
+		"name":    0,
+	}
```
mauriciovasquezbernal, line 50: Shouldn't it include all columns from the container struct? 

Something like should work:

```
$ sudo ./local-gadget list-containers -o json
[
  {
    "ID": "11d5de97dbf1fabbcd93807fc172d8be422ab40bfc531fc6f106058d111eb527",
    "Pid": 5934,
    "OciConfig": null,
    "Bundle": "",
    "Mntns": 4026533360,
    "Netns": 4026533365,
    "CgroupPath": "/sys/fs/cgroup/unified/system.slice/docker-11d5de97dbf1fabbcd93807fc172d8be422ab40bfc531fc6f106058d111eb527.scope",
    "CgroupID": 2225,
    "CgroupV1": "/system.slice/docker-11d5de97dbf1fabbcd93807fc172d8be422ab40bfc531fc6f106058d111eb527.scope",
    "CgroupV2": "/system.slice/docker-11d5de97dbf1fabbcd93807fc172d8be422ab40bfc531fc6f106058d111eb527.scope",
    "Namespace": "default",
    "Podname": "local-registry",
    "Name": "local-registry",
    "Labels": null,
    "OwnerReference": null,
    "PodUID": "",
    "Runtime": "docker"
  }
]
$ sudo ./local-gadget list-containers -o custom-columns=id,netns
ID               <--- netns column not printed but it's available on the json.
11d5de97dbf1f    
```

pkg/container-utils/runtime-client/interface.go
```diff
@@ -39,6 +39,11 @@ type ContainerData struct {
 
 	// Running defines whether or not the container is in the running state
 	Running bool
+
+	// Runtime is the name of the runtime (e.g. docker, cri-o, containerd). It
+	// is useful to distinguish who is the "owner" of each container in a list
+	// of container collected from multiples runtimes.
```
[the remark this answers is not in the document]
blanquicet: I don't think so. The "collected from multiples runtimes" goes all together. IMHO, your suggestion would be true if it were just "in a list of collected containers". Don't you think?

cmd/local-gadget/containers/containers.go
blanquicet: You are right. I think I would still omit the ones that we already know will always be nil. Do you agree? The documentation already says it will be always nil.

pkg/container-utils/runtime-client/interface.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I'm not sure either. Anyway, I think "container" must be plural on your original sentence.

cmd/local-gadget/containers/containers.go
mauriciovasquezbernal: Yes, I agree. I think we're missing the `omitempty` tag on the container structure declaration to avoid that behavior, but it's not something we need to make on this PR.

PR #764, opened 2022-07-25
cmd: Manage JSON output mode outside of utils/trace

blanquicet:
# Manage JSON output mode outside of utils/trace

Before this PR, all the gadgets using `RunTraceAndPrintStream` don't have the full control over the JSON output mode because this function was internally managing that case [here](https://github.com/kinvolk/inspektor-gadget/blob/0f5fd1688e67494beb52dd995d2dbb69f2e4ad53/cmd/kubectl-gadget/utils/trace.go#L928-L930). However, with the introduction of the gadget parsers, keeping this logic on the utils/trace doesn't make sense anymore. The full management of the output format must be controlled by the parsers.

In this PR, the logic to print the received lines was moved from [cmd/kubectl-gadget/utils/trace.go](https://github.com/kinvolk/inspektor-gadget/blob/0f5fd1688e67494beb52dd995d2dbb69f2e4ad53/cmd/kubectl-gadget/utils/trace.go#L928-L930) to [cmd/common/utils/parser.go](https://github.com/kinvolk/inspektor-gadget/blob/bcf4267b548c712193a42800dd9a6e53f4fdbdbd/cmd/common/utils/parser.go#L83-L90). However, it was done in two steps:

1. Move the logic from the utils/trace to the gadgets
2. Move the logic from the gadgets to parsers. In particular, to the `BaseParser` to avoid duplicated code on all the gadgets.

This PR also introduce some tests for the `BaseParser`. However, additional tests need to be added for every gadget parser.

## How to use

This PR does not introduce any change for the users.

## Testing done

Integration tests for all the gadgets using the `BaseParser` are passing.

## TODO in future PRs

Modify the Audit and Top categories to take advantage of the `BaseParser`. Once it is done, we will be easier to implement common features like: combine the `OutputModeColumns` and `OutputModeCustomColumns`, print the available columns that can be used with `-o custom-columns` (see https://github.com/kinvolk/inspektor-gadget/pull/664#discussion_r923422310).

### DIFF
--- a/cmd/kubectl-gadget/audit/seccomp.go
+++ b/cmd/kubectl-gadget/audit/seccomp.go
@@ -81,11 +81,17 @@ func auditSeccompTransformLine(line string) string {
 	}
 
 	switch params.OutputMode {
+	case commonutils.OutputModeJSON:
+		b, err := json.Marshal(e)
+		if err != nil {
+			fmt.Fprint(os.Stderr, fmt.Sprint(commonutils.WrapInErrMarshalOutput(err)))
+			return ""
+		}
+		sb.WriteString(string(b))
 	case commonutils.OutputModeColumns:
 		sb.WriteString(fmt.Sprintf("%-16s %-16s %-16s %-16s %-16s %-6d %-16s %-16s",
 			e.Node, e.Namespace, e.Pod, e.Container,
 			e.Comm, e.Pid, e.Syscall, e.Code))
-
 	case commonutils.OutputModeCustomColumns:
 		for _, col := range params.CustomColumns {
 			switch col {
--- a/cmd/local-gadget/containers/containers.go
+++ b/cmd/local-gadget/containers/containers.go
@@ -108,7 +108,7 @@ func NewListContainersCmd() *cobra.Command {
 				fmt.Fprintln(w, parser.BuildColumnsHeader())
 
 				for _, c := range containers {
-					fmt.Fprintln(w, parser.TransformContainerToColumns(c))
+					fmt.Fprintln(w, parser.TransformToColumns(c))
 				}
 
 				w.Flush()
@@ -146,7 +146,7 @@ func (p *ContainerParser) SortContainers(containers []*containercollection.Conta
 	})
 }
 
-func (p *ContainerParser) TransformContainerToColumns(c *containercollection.Container) string {
+func (p *ContainerParser) TransformToColumns(c *containercollection.Container) string {
 	var sb strings.Builder
 
 	for _, col := range p.OutputConfig.CustomColumns {
--- a/cmd/local-gadget/containers/containers.go
+++ b/cmd/local-gadget/containers/containers.go
@@ -35,7 +35,7 @@ type ContainerFlags struct {
 }
 
 type ContainerParser struct {
-	commonutils.BaseParser
+	commonutils.BaseParser[containercollection.Container]
 
 	containerFlags *ContainerFlags
 }
@@ -82,7 +82,7 @@ func NewListContainersCmd() *cobra.Command {
 			defer localGadgetManager.Close()
 
 			parser := &ContainerParser{
-				BaseParser:     commonutils.NewBaseTabParser(availableColumns, &commonFlags.OutputConfig),
+				BaseParser:     commonutils.NewBaseTabParser[containercollection.Container](availableColumns, &commonFlags.OutputConfig),
 				containerFlags: &containerFlags,
 			}
 
--- a/cmd/common/snapshot/process.go
+++ b/cmd/common/snapshot/process.go
@@ -87,7 +87,7 @@ func NewCommonProcessCmd(
 	return cmd
 }
 
-func (p *ProcessParser) TransformEvent(e *types.Event) string {
+func (p *ProcessParser) TransformToColumns(e *types.Event) string {
 	var sb strings.Builder
 
 	for _, col := range p.OutputConfig.CustomColumns {
--- a/cmd/common/snapshot/process.go
+++ b/cmd/common/snapshot/process.go
@@ -33,7 +33,7 @@ type ProcessFlags struct {
 }
 
 type ProcessParser struct {
-	commonutils.BaseParser
+	commonutils.BaseParser[types.Event]
 
 	processFlags *ProcessFlags
 }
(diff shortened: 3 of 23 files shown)


docs/guides/snapshot/process.md
```diff
@@ -74,32 +74,30 @@ minikube    demo         mypod    mypod        sleep    584294
...
+  "type": "normal",
+  "node": "minikube",
+  "namespace": "demo",
+  "pod": "mypod",
+  "container": "mypod",
+  "tgid": 582333,
+  "pid": 582333,
+  "comm": "nginx",
+  "mntns": 4026532507
+}
```
mauriciovasquezbernal, line 99: TL;DR: I think this result is wrong and we should keep the previous one.

We need to understand what it means to have json output on the different gadget categories. For gadgets that produce events, like trace, top, audit, the full output of the tool is not a json array of events, instead each single line is a json itself that represents the event. On the other hand, gadgets in the snapshot category produce a full json array as result. 

I think it's useful to have snapshotters to produce a valid json output, I'm not sure what to do for tracers and similar, as they don't have all the output available at one hence it's not possible to print the event in a single call. 
Maybe we can print a `[` at the beginning, separate lines by `,`, and print a `]` before exiting? (Or maybe it's too complicated and we shouldn't worry about it?).  

@alban do you have any opinions on this?

mauriciovasquezbernal (commented):
I have some doubts about the way we're printing json, so I stopped there until we clarify it.

docs/guides/snapshot/process.md
mauriciovasquezbernal: Even if I agree that having the same behavior would be nice, I also think that different categories can have different behaviors:

> $ kubectl gadget trace tcp -o json | myapp

In this case myapp should be prepared to process one event at the time (currently each new line represents a new event). 

> $ kubectl gadget snapshot socket -o json | myapp

In this case it's expected to received all the information on a single step, so probably myapp would read all and parse as a single json structure. 

I agree adding `[` to the tracers is too much, but I think snapshotters should keep the current behavior.

alban: > TL;DR: I think this result is wrong and we should keep the previous one.

I agree with that.
* Snapshotters: with should be able to use `| jq`, so we need valid json.
* Streams cannot be parsed as valid json and adding `[` would not help because a partial json document would still not be parsable. I think the separation by line was good.

blanquicet: Make sense. I love uniformity 😝, but it is also true that one of the reasons for dividing the gadgets into categories is that in some cases, like this one, we want to have different formats between categories while keeping uniformity within each category. I will revert commits for the snapshot category.

blanquicet: BTW, `jq` can still be used even if each line is a JSON by itself (like in the case of streams):

```bash
$ printf '{"a": 1}\n{"a": 2}\n'|jq .a
1
2
```

cmd/local-gadget/containers/containers.go
```diff
@@ -92,30 +91,18 @@ func NewListContainersCmd() *cobra.Command {
 
 			parser.SortContainers(containers)
 
-			switch commonFlags.OutputMode {
-			case commonutils.OutputModeJSON:
-				b, err := json.MarshalIndent(containers, "", "  ")
-				if err != nil {
-					return commonutils.WrapInErrMarshalOutput(err)
-				}
```
mauriciovasquezbernal, line 100: I have the same concern as before. Why not to keep this behavior and print a json array with all the containers??

blanquicet: 🤦🏽 ... Yep, I had to revert these changes as well. I am sorry!

cmd/kubectl-gadget/audit/seccomp.go
```diff
@@ -81,11 +81,17 @@ func auditSeccompTransformLine(line string) string {
 	}
 
 	switch params.OutputMode {
+	case commonutils.OutputModeJSON:
+		b, err := json.Marshal(e)
+		if err != nil {
+			fmt.Fprint(os.Stderr, fmt.Sprint(commonutils.WrapInErrMarshalOutput(err)))
+			return ""
+		}
+		sb.WriteString(string(b))
```
mauriciovasquezbernal, line 90: I suppose this change is needed because we still don't have a parser implemented for this gadget, right?

mauriciovasquezbernal (approved):
I think the changes are well organized and the result code is clear enough. I implemented this example (after rebasing this branch to main) and worked fine:

```golang
package main

import (
	"fmt"
	"os"
	"os/signal"

	"github.com/cilium/ebpf/rlimit"
	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/exec/tracer"
	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/exec/types"

	"github.com/kinvolk/inspektor-gadget/cmd/common/utils"
	cmdtrace "github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/trace"
)

func main() {
	if err := rlimit.RemoveMemlock(); err != nil {
		return
	}

	config := &utils.OutputConfig{
		// Configure the output mode to use.
		OutputMode: utils.OutputModeJSON,
	}

	// Create a parser. It's the component that converts events to
	// strings according to the configuration above.
	execparser := cmdtrace.NewExecParser(config)

	// Define a callback to be called each time there is an event.
	eventCallback := func(event types.Event) {
		// Convert the event to string and print to the terminal.
		fmt.Println(execparser.TransformEvent(&event))
	}

	// Create the tracer. An empty configuration is passed as we are
	// not interesting on filtering by any container. For the same
	// reason, no enricher is passed.
	tracer, err := tracer.NewTracer(&tracer.Config{}, nil, eventCallback)
	if err != nil {
		fmt.Printf("error creating tracer: %s\n", err)
		return
	}

	// Wait for SIGINT
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)

	<-c

	// Clean up everything before exiting.
	tracer.Stop()
}
```


```bash
$ sudo ./trace 
{"type":"normal","pid":30277,"ppid":1083,"mountnsid":4026531840,"pcomm":"runc","args":["/usr/sbin/runc","--version"]}
{"type":"normal","pid":30283,"ppid":1083,"mountnsid":4026531840,"pcomm":"docker-init","args":["/usr/bin/docker-init","--version"]}
```

cmd/kubectl-gadget/audit/seccomp.go
blanquicet: Exactly. I implemented the parser for that gadget in https://github.com/kinvolk/inspektor-gadget/pull/770.

PR #767, opened 2022-07-25
Cleanup resolver interface

mauriciovasquezbernal:
This PR contains two commits to cleanup the resolver interface. The whole goal of this change is to make tracers usable by 3rd party apps (https://github.com/kinvolk/inspektor-gadget/issues/723). 

```
    pkg/gadget-collection: Rename Resolver interface to GadgetHelpers
    
    The old called "Resolver" interface is not a resolver anymore. It does
    a bunch of different things. This commit renames it to something more
    generic.
```

```
    pkg/gadgets: Create EventEnricher interface
    
    Create an EventEnricher interface that is used by the tracers to enrich
    events with container information. This commit also removes references
    to the resolver in standard tracers given that it's not used at all.

```

Fixes https://github.com/kinvolk/inspektor-gadget/issues/685

### DIFF
--- a/pkg/gadgets/audit/seccomp/tracer/tracer.go
+++ b/pkg/gadgets/audit/seccomp/tracer/tracer.go
@@ -143,20 +143,21 @@ func (t *Tracer) run() {
 		event := types.Event{
 			Event: eventtypes.Event{
 				Type: eventtypes.NORMAL,
-				Node: t.node,
-
-				// Get 'Namespace', 'Pod' and 'Container' from
-				// BPF and not from the resolver because the
-				// container might be terminated immediately
-				// after the BPF kprobe on audit_seccomp() is
-				// executed (e.g. with SCMP_ACT_KILL), so by
-				// the time the event is read from the perf
-				// ring buffer, we might not be able to get the
-				// Kubernetes metadata from the mount namespace
-				// id.
-				Namespace: C.GoString(&eventC.container.namespace[0]),
-				Pod:       C.GoString(&eventC.container.pod[0]),
-				Container: C.GoString(&eventC.container.container[0]),
+				CommonData: eventtypes.CommonData{
+					Node: t.node,
+					// Get 'Namespace', 'Pod' and 'Container' from
+					// BPF and not from the gadget helpers  because the
+					// container might be terminated immediately
+					// after the BPF kprobe on audit_seccomp() is
+					// executed (e.g. with SCMP_ACT_KILL), so by
+					// the time the event is read from the perf
+					// ring buffer, we might not be able to get the
+					// Kubernetes metadata from the mount namespace
+					// id.
+					Namespace: C.GoString(&eventC.container.namespace[0]),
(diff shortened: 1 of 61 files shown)


pkg/gadgets/profile/cpu/tracer/tracer.go
```diff
@@ -254,19 +254,15 @@ func getReport(t *Tracer, kAllSyms []kernelSymbol, stack *ebpf.Map, keyCount key
 	}
 
 	report := types.Report{
-		Node:        t.node,
```
blanquicet, line 257: From a quick analysis, I saw that removing this assignment makes that the tracer doesn't need to receive the node anymore. I notice it is the same for several gadgets, the ones in the profile and top categories (snapshot is a special case). Did I miss something? Maybe you were planning to do it in another PR?

alban:
I like this change. I agree with @blanquicet. I think the `node` parameters can be removed in this PR, for all gadgets that don't need it (all of them?).

mauriciovasquezbernal:
@alban @blanquicet I'm planning to remove the node parameter in a different PR. (I tried to make all these changes in a single PR but I realized they were too many, so I'm going step by step).

pkg/gadgets/audit/seccomp/tracer/tracer.go
```diff
@@ -143,20 +143,21 @@ func (t *Tracer) run() {
...
-				// ring buffer, we might not be able to get the
-				// Kubernetes metadata from the mount namespace
-				// id.
-				Namespace: C.GoString(&eventC.container.namespace[0]),
-				Pod:       C.GoString(&eventC.container.pod[0]),
-				Container: C.GoString(&eventC.container.container[0]),
+				CommonData: eventtypes.CommonData{
+					Node: t.node,
+					// Get 'Namespace', 'Pod' and 'Container' from
+					// BPF and not from the resolver because the
```
alban, line 149:
```suggestion
					// BPF and not from the gadget helpers because the
```

blanquicet (approved):
LGTM!

pkg/types/types.go
```diff
@@ -17,10 +17,33 @@ package types
...
 	"fmt"
+	"os"
 )
 
 type EventType string
 
+var node string
+
+func init() {
+	node = os.Getenv("NODE_NAME")
```
alban, line 28: `$NODE_NAME` is defined in the pod spec, but this package `pkg/types` is used outside of the gadget pod (e.g. in the `kubectl-gadget` CLI) so I feel uneasy about using environment variables there.

What about
```
func Init(n string) {
    node = n
}
```
And getting the gadget-tracer-manager to call `Init()`?

alban (approved):
A couple of suggestions but otherwise LGTM!

pkg/types/types.go
mauriciovasquezbernal: Thanks for catching this up. I'll handle this comment on https://github.com/kinvolk/inspektor-gadget/pull/774 as this change actually belongs to that PR and not here.

PR #770, opened 2022-07-26
cmd/kubectl-gadget/audit: Use BaseParser on seccomp gadget

blanquicet:
# Use BaseParser to avoid repeated code

With this PR, the audit-seccomp gadget starts using the `BaseParser` to reuse the logic that prints the header columns and transforms the events into columns.

In addition, using `BaseParser` for all the gadgets that support JSON and columns output modes will make easier the implementation of some common features like: combine the `OutputModeColumns` and `OutputModeCustomColumns` and printing the columns that can be used with `custom-columns`.

## How to use

The only impact of these changes is that the columns now follow the order we defined for other gadgets: node, podname, container, pid, comm and then the rest.

## Testing done

Integration tests were updated with the changes in the order of the columns, and it is passing.

### DIFF
--- a/cmd/kubectl-gadget/audit/audit.go
+++ b/cmd/kubectl-gadget/audit/audit.go
@@ -15,16 +15,16 @@
 package audit
 
 import (
-	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
-
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal (approved):
I haven't tested it but it seems it's following the same approach as in other gadgets, then I'll approve.

PR #774, opened 2022-07-26
gadgets: Remove node parameter when creating tracer

mauriciovasquezbernal:
The node parameter is not needed anymore. This changes makes it easier
to import and use the tracers directly as there is one less parameter
to be passed.

### DIFF
--- a/pkg/gadgets/trace/bind/tracer/tracer.go
+++ b/pkg/gadgets/trace/bind/tracer/tracer.go
@@ -72,7 +72,6 @@ type Tracer struct {
 	config        *Config
 	enricher      gadgets.DataEnricher
 	eventCallback func(types.Event)
-	node          string
 
 	objs      bindsnoopObjects
 	ipv4Entry link.Link
@@ -83,13 +82,12 @@ type Tracer struct {
 }
 
 func NewTracer(config *Config, enricher gadgets.DataEnricher,
-	eventCallback func(types.Event), node string,
+	eventCallback func(types.Event),
 ) (*Tracer, error) {
 	t := &Tracer{
 		config:        config,
 		enricher:      enricher,
 		eventCallback: eventCallback,
-		node:          node,
 	}
 
 	if err := t.start(); err != nil {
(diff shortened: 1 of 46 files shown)


alban (approved):
LGTM

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -260,13 +258,13 @@ func (t *Tracer) run() {
 			}
 
 			msg := fmt.Sprintf("Error reading perf ring buffer: %s", err)
-			t.eventCallback(types.Base(eventtypes.Err(msg, t.node)))
+			t.eventCallback(types.Base(eventtypes.Err(msg)))
```
blanquicet, line 261: For `local-gadget` or third-party apps that use directly the tracer the node information is not useful. However, it is when using the gadget from Kubernetes. Therefore, what if we add it to the CRD code?

I know it is not very clean/pretty but we could do something like this [here](https://github.com/kinvolk/inspektor-gadget/blob/61a37c65c326517bc29b63fc8065a85c9bdb55db/pkg/gadget-collection/gadgets/trace/bind/gadget.go#L100-L107):

```golang
        eventCallback := func(event types.Event) {
                // Add the node information in case it is missing
		if event.Node == "" {
			event.Node = trace.Spec.Node
		}

		r, err := json.Marshal(event)
		if err != nil {
			log.Warnf("Gadget %s: error marshalling event: %s", trace.Spec.Gadget, err)
			return
		}
		t.helpers.PublishEvent(traceName, string(r))
	}
```

Does it look that bad? 😅

blanquicet (approved):
Just one comment but LGTM!

pkg/gadgets/trace/bind/tracer/tracer.go
mauriciovasquezbernal: I created some confusion here because I was made a wrong change in the other PR (https://github.com/kinvolk/inspektor-gadget/pull/767#discussion_r930175408). I'm not removing the information about the node, but given it's always the same, I defined a variable in the types package that automatically sets this to the event created when calling the `Err(), Warn()` and other functions.

PR #775, opened 2022-07-26
examples: Add gadgets examples

mauriciovasquezbernal:
Add some examples showing how to use the Golang packages of our gadgets.

Part of https://github.com/kinvolk/inspektor-gadget/issues/723.


.github/workflows/inspektor-gadget.yml
```diff
@@ -243,6 +243,18 @@ jobs:
...
+    name: Build Gadgets examples
+    # level: 0
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build example binaries
+      run: |
+        go build ./examples/gadgets/trace/exec
+        go build ./examples/gadgets/top/file
+        go build ./examples/gadgets/snapshot/process
```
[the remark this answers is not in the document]
alban: We already have the job `build-examples` (line 211) that builds `examples/kube-container-collection` and `examples/runc-hook` in Docker.

If we use the `find` command here, then they would be built twice.

examples/README.md
```diff
@@ -0,0 +1,24 @@
+# examples
+
+This folder contains some examples showing the usage of the Golang
+packages provided by Inspektor Gadget.
```
alban, line 4: You could add a paragraph explaining that the examples are for developers who want to reuse code from Inspektor Gadget. End-users don't need to use those: they can use local-gadget or kubectl-gadget instead.

examples/gadgets/trace/exec/exec.go
```diff
@@ -0,0 +1,59 @@
...
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+
+	"github.com/cilium/ebpf/rlimit"
+
+	// Import tracer packages
```
alban, line 24:
```suggestion
```

For consistency: the file.go example didn't have this comment

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
alban: Fine for me but Mauricio would need to write a Dockerfile for the new examples.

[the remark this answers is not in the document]
mauriciovasquezbernal: > ```shell
> for dir in $(dirname $(find examples/gadgets -name README.md)); do
>   go build $dir
> done
> ```

I used a slightly modified version of this.

mauriciovasquezbernal:
I think I handled all the comments. Please review again :smile:

examples/README.md
```diff
@@ -0,0 +1,27 @@
+# examples
+
+This folder contains some examples showing the usage of the Golang
+packages provided by Inspektor Gadget. These examples are designed for
+developers that want to use the Golang exposed by Inspektor Gadget
```
blanquicet, line 5:
```suggestion
developers that want to use the Golang packages exposed by Inspektor Gadget
```
?

examples/README.md
```diff
@@ -0,0 +1,27 @@
...
+- gadgets
+  - [trace/exec](gadgets/trace/exec/): basic example showing how to use
+    [trace/exec](https://github.com/kinvolk/inspektor-gadget/tree/main/pkg/gadgets/trace/exec)
+    package to trace the creation of new processes in the host.
+  - [top/file](gadgets/top/file/): shows how to use
+    [top/file](https://github.com/kinvolk/inspektor-gadget/tree/main/pkg/gadgets/top/file)
+    package to get a list of the files with a higher number of write operations
+    on the host.
+  - [snapshot/process](gadgets/snapshot/process/): shows how to use
+    [snapshot/process](https://github.com/kinvolk/inspektor-gadget/tree/main/pkg/gadgets/snapshot/    process)
```
blanquicet, line 26:
```suggestion
    [snapshot/process](https://github.com/kinvolk/inspektor-gadget/tree/main/pkg/gadgets/snapshot/process)
```
Otherwise the link is broken.

examples/README.md
```diff
@@ -0,0 +1,27 @@
...
+- [runc-hook](runc-hook/): Use of the runcfanotify package to
+  - receive notifications when a container starts or terminates
+  - execute PreStart and PostStop hooks
+- [kube-container-collection](kube-container-collection/): Use the
+  container-collection package
+  ("github.com/kinvolk/inspektor-gadget/pkg/container-collection") in
+  order to be notified when a new container is started and to attach the
+  OCI config.json as a Kubernetes event.
+- gadgets
+  - [trace/exec](gadgets/trace/exec/): basic example showing how to use
```
blanquicet, line 18: As for other descriptions?

```suggestion
  - [trace/exec](gadgets/trace/exec/): shows how to use
```

examples/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+- [runc-hook](runc-hook/): Use of the runcfanotify package to
+  - receive notifications when a container starts or terminates
+  - execute PreStart and PostStop hooks
+- [kube-container-collection](kube-container-collection/): Use the
+  container-collection package
+  ("github.com/kinvolk/inspektor-gadget/pkg/container-collection") in
+  order to be notified when a new container is started and to attach the
+  OCI config.json as a Kubernetes event.
+- gadgets
```
blanquicet, line 17:
```suggestion
- gadgets:
```

examples/gadgets/snapshot/process/process.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/cilium/ebpf/rlimit"
+
+	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/snapshot/process/tracer"
+)
+
+func main() {
+	if err := rlimit.RemoveMemlock(); err != nil {
```
blanquicet, line 26: As in other examples:

```suggestion
        // In some kernel versions it's needed to bump the rlimits to
	// use run BPF programs.
	if err := rlimit.RemoveMemlock(); err != nil {
```

examples/gadgets/trace/exec/exec.go
```diff
@@ -0,0 +1,58 @@
...
+	// Define a callback to be called each time there is an event.
+	eventCallback := func(event types.Event) {
+		fmt.Printf("A new %q process with pid %d was executed\n",
+			event.Comm, event.Pid)
+	}
+
+	// Create the tracer. An empty configuration is passed as we are
+	// not interesting on filtering by any container. For the same
+	// reason, no enricher is passed.
+	tracer, err := tracer.NewTracer(&tracer.Config{}, nil, eventCallback)
```
blanquicet, line 44: BTW, it would be nice if we could pass `nil` instead of `&tracer.Config{}`.

examples/gadgets/trace/exec/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## How to run
+
+```bash
+$ sudo ./exec
+A new "calico" process with pid 118594 was executed
+A new "portmap" process with pid 118606 was executed
+A new "bandwidth" process with pid 118611 was executed
+A new "runc" process with pid 118616 was executed
+A new "docker-init" process with pid 118623 was execute
```
alban, line 24: execute -> executed ?

alban (approved):

examples/gadgets/snapshot/process/process.go
```diff
@@ -0,0 +1,40 @@
...
+		return
+	}
+
+	processes, err := tracer.RunCollector(nil, nil)
```
blanquicet, line 30: When passing all these `nil`, I cannot avoid thinking about how we can avoid them 😅. In this case, I immediately tought "How would it looks like using the _option pattern_?"

Definition:
```golang
func RunCollector(options ...Options) ([]processcollectortypes.Event, error) {
    ...
}
```

Usages:
```golang
processes, err := tracer.RunCollector(WithEnricher(t.helpers))
```
```golang
processes, err := tracer.RunCollector(WithEnricher(t.helpers), WithFiltering(mountNsMap))
```
```golang
processes, err := tracer.RunCollector()
```

It looks better, but I am not sure if it is worth using it in this case.

examples/gadgets/trace/exec/README.md
blanquicet: OMG, how did you notice that!? hahahaha

mauriciovasquezbernal: ops, missed the `d` while Copying and pasting.

examples/README.md
mauriciovasquezbernal: I'm wondering how it happened :thinking:

examples/gadgets/snapshot/process/process.go
mauriciovasquezbernal: I agree that it looks better, but it seems too much work now. How will it work for other categories? 

```
tracer, err := tracer.NewTracer(tracer.WithConfig(c), tracer.WithEnricher(foo))
```

?

examples/gadgets/trace/exec/exec.go
mauriciovasquezbernal: Good point. I was thinking that we could make the `enricher` and `callback` part of the configuration, then it will only receive a single parameter making it easier to extend.

examples/gadgets/snapshot/process/process.go
blanquicet: For other categories, the only parameter that is optional is the enricher so that it would be:

For `top` category:

```golang
tracer, err := tracer.NewTracer(config, statsCallback, errorCallback)
```
```golang
tracer, err := tracer.NewTracer(config, statsCallback, errorCallback, WithEnricher(t.helpers))
```

For `trace` category:
```golang
t.tracer, err = tracer.NewTracer(config, eventCallback)
```
```golang
t.tracer, err = tracer.NewTracer(config, eventCallback, WithEnricher(t.helpers))
```

Same for the `profile/cpu` gadget.

The implementation would be something like this:

```golang
type Option func(*Tracer)

func WithEnricher(enricher gadgets.DataEnricher) Option {
	return func(t *Tracer) {
		t.enricher = enricher
	}
}

func NewTracer(
	config *Config,
	eventCallback func(types.Event),
	options ...Option,
) (*Tracer, error) {
	t := &Tracer{
		config:        config,
		eventCallback: eventCallback,
	}

	for _, opts := range options {
		opts(t)
        }
	
        ...
}
```

However, given that the `Tracer` is different for each gadget, we would have to redefine `WithEnricher` within each gadget. Generics could solve this problem, but maybe it starts becoming too much just to avoid passing a `nil` 😅. It is more evident with the `RunCollector` where we pass three `nil`, but I don't have a better proposal for it now. Anyway, let's keep this in mind if we need to add more optional parameters to the tracers.

examples/gadgets/trace/exec/exec.go
blanquicet: Yeah, that could also resolve [this discussion](https://github.com/kinvolk/inspektor-gadget/pull/775/files#r931213502).

blanquicet (approved):

PR #778, opened 2022-07-27
pkg/resources: Fix image tag in manifest

mauriciovasquezbernal:
`kubectl apply -f pkg/resources/manifests/deploy.yaml` wasn't working because the image was wrong.


blanquicet (approved):
LGTM

PR #780, opened 2022-07-28
cmd/kubectl-gadget: Do not fail if tracer controller is dead on node

mauriciovasquezbernal:
Add logic to handle the case when some the trace controller is dead
in some nodes, i.e. some traces don't reach the expected state.

This is important because it should be possible to get some results
from the cluster even if the gadget pod is not running fine in all of
the nodes.

## How to test 

It's not so easy to kill the gadgettracermanager on a node because kubernetes restarts it, for this reason I have to use this hack:

1. Modify the gadget tracer manager not to start on a specific node:
```diff
$ git diff
diff --git gadget-container/gadgettracermanager/main.go gadget-container/gadgettracermanager/main.go
index 7325c875..777d4ee9 100644
--- gadget-container/gadgettracermanager/main.go
+++ gadget-container/gadgettracermanager/main.go
@@ -214,6 +214,10 @@ func main() {
                        log.Fatalf("Environment variable NODE_NAME not set")
                }
 
+               if node == "worker" {
+                       time.Sleep(24 * 60 * 60 * time.Second)
+               }
+
                lis, err := net.Listen("unix", socketfile)
                if err != nil {
                        log.Fatalf("failed to listen: %v", err)
```

2. Deploy without liveness probes to avoid kubernetes restarting it

```
$ ./kubectl-gadget deploy --liveness-probe=false
```

3. Run a gadget
```bash
./kubectl-gadget trace exec -A
NODE             NAMESPACE        POD                            CONTAINER        PID     PPID    PCOMM            RET  ARGS                     
Timeout waiting for condition on node "worker"
ubuntu-hirsute   kube-system      calico-node-4mx6g              calico-node      906422  5290    ipset            0    /usr/sbin/ipset list  
ubuntu-hirsute   gadget           gadget-kq7rc                   gadget           906456  906446  sh               0    /bin/sh -c exec gadgettracermanager -call receive-stream -tracerid trace_gadget_execsnoop-lr69l  
ubuntu-hirsute   gadget           gadget-kq7rc                   gadget           906456  906446  gadgettracerman  0    /usr/bin/gadgettracermanager -call receive-stream -tracerid trace_gadget_execsnoop-lr69l 
```

Fix https://github.com/kinvolk/inspektor-gadget/issues/632.

### DIFF
--- a/cmd/kubectl-gadget/utils/trace.go
+++ b/cmd/kubectl-gadget/utils/trace.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	types "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -152,7 +153,7 @@ func deleteTraces(traceClient *clientset.Clientset, traceID string) {
 		context.TODO(), metav1.DeleteOptions{}, listTracesOptions,
 	)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error deleting traces: %q", err)
+		fmt.Fprintf(os.Stderr, "Error: deleting traces: %q", err)
 	}
 }
 
@@ -349,7 +350,7 @@ func CreateTrace(config *TraceConfig) (string, error) {
 			deleteError := DeleteTrace(traceID)
 
 			if deleteError != nil {
-				fmt.Fprintf(os.Stderr, "%v\n", err)
+				fmt.Fprintf(os.Stderr, "Error: deleting trace: %s\n", err)
 			}
 
 			return "", err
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/utils/trace.go
```diff
@@ -597,7 +598,18 @@ func waitForCondition(traceID string, conditionFunction func(*gadgetv1alpha1.Tra
...
+		if !errors.Is(err, wait.ErrWaitTimeout) {
+			return nil, err
+		}
+
+		// Print a message for traces that timed out
+		for _, trace := range traceList.Items {
+			_, satisfied := satisfiedTraces[trace.ObjectMeta.Name]
+			_, errored := erroredTraces[trace.ObjectMeta.Name]
+			if !satisfied && !errored {
+				fmt.Fprintf(os.Stderr, "Timeout waiting for condition on node %q\n", trace.Spec.Node)
```
blanquicet, line 610: Beyond this change, I was thinking that `Timeout waiting for condition` (which is identical to the current error message), maybe is not that user-friendly. I think users will ask themselves, "_What condition?_". Maybe something like `Timeout trying to run gadget on node`? WDYT?

cmd/kubectl-gadget/utils/trace.go
```diff
@@ -597,7 +598,18 @@ func waitForCondition(traceID string, conditionFunction func(*gadgetv1alpha1.Tra
...
+			}
+		}
```
blanquicet, line 612: You should manage the case when all the traces timed out and return an error. Otherwise, the CLI will remain blocked, waiting to print events that will never arrive. I don't expect the CLI to be able to recover the communication with the node and resume the work when the node is up again. Or what did I miss?

blanquicet (commented):
I was able to reproduce the issue and it is now working as expected. Thanks!

Just a couple of comments about the error message and I think that maybe there is a use case that is not covered.

cmd/kubectl-gadget/utils/trace.go
mauriciovasquezbernal: The problem is that `waitForCondition` is used in `waitForTraceState` and `waitForNoOperation`, so `Timeout trying to run gadget on node` is misleading in the second case. I have no more ideas to fix it without a big change.

cmd/kubectl-gadget/utils/trace.go
mauriciovasquezbernal: ops, you're right. I wanted to do it but got distracted and pushed the commit without this.

cmd/kubectl-gadget/utils/trace.go
blanquicet: Let's keep this for now.

blanquicet (approved):

mauriciovasquezbernal:
Do you have more than one node on your cluster? I just tested on a minikube cluster with two nodes and worked fine.

PR #786, opened 2022-07-28
kubectl-gadget: Print error when undeploy command timeouts

mauriciovasquezbernal:
The undeploy command wasn't printing any error when there was a timeout
trying to remove the gadget namespace. It was misleading as Inspektor
Gadget was not actually removed and the command exited with 0.

### DIFF
--- a/cmd/kubectl-gadget/undeploy.go
+++ b/cmd/kubectl-gadget/undeploy.go
@@ -234,12 +234,15 @@ again:
 
 		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
 		defer cancel()
-		watchtools.Until(ctx, list.ResourceVersion, watcher, conditionFunc)
+		_, err := watchtools.Until(ctx, list.ResourceVersion, watcher, conditionFunc)
+		if err != nil {
+			errs = append(errs, fmt.Sprintf("failed waiting for %q namespace to be removed: %s", gadgetNamespace, err))
+		}
 	}
 
 out:
 	if len(errs) > 0 {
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/undeploy.go
```diff
@@ -234,7 +234,10 @@ again:
 
 		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
 		defer cancel()
-		watchtools.Until(ctx, list.ResourceVersion, watcher, conditionFunc)
+		_, err := watchtools.Until(ctx, list.ResourceVersion, watcher, conditionFunc)
+		if err != nil {
+			errs = append(errs, fmt.Sprintf("Failed waiting for IG to be removed: %s", err))
```
blanquicet, line 239: I would instead use `namespace` or `Inspektor Gadget`:

```suggestion
			errs = append(errs, fmt.Sprintf("Failed waiting for namespace to be removed: %s", err))
```

And then error should be printed:

```go
return fmt.Errorf("error removing Inspektor Gadget:\n%s", strings.Join(errs, "\n"))
```

blanquicet (approved):
One comment about the error message but LGTM

cmd/kubectl-gadget/undeploy.go
mauriciovasquezbernal: Fixed. Thanks for the suggestions.

mauriciovasquezbernal: Fixed. Thanks for the suggestions.

PR #788, opened 2022-07-29
audit/seccomp: Remove node parameter from tracer

blanquicet:
# Remove node parameter from audit/seccomp tracer

Enriching the event with node information on the gadget-collection side allows using the tracer from outside a Kubernetes context. For instance, from local-gadget or 3rd party apps using IG packages. 

## How to use

No changes for users.


mauriciovasquezbernal (approved):
LGTM!

PR #796, opened 2022-08-02
cmd/common: Handle `ManageSpecialEvent` inside `BaseParser.Transform`

blanquicet:
# Handle `ManageSpecialEvent` inside `BaseParser.Transform`

This PR aims to handle `ManageSpecialEvent` inside `BaseParser.Transform`. The main issue with doing so is that it doesn't need to be done in all cases. Therefore, we need to identify whether `ManageSpecialEvent` needs to be called or not. My proposal in this PR is to base the decision on the value returned by `GetBaseEvent`.

## Implementation details

The gadgets using `BaseParser.Transform` are the ones supporting columns output format:

- The `audit/seccomp` gadget
- The `profile/cpu` gadget
- The `trace` category
- The `top` category is in the process: https://github.com/kinvolk/inspektor-gadget/pull/782. BTW, we need to synchronize these two PRs and rebase once one of them is merged.
- **NOTE**: The `snapshot` category and the `list-container` command of `local-gadget` don't use `BaseParser.Transform` as they print the JSON output format differently (as discussed [here](https://github.com/kinvolk/inspektor-gadget/pull/764#discussion_r930044431)). It means they will need to manage special events (if needed) by themselves. 

However, the `profile/cpu` gadget and the `top` category don't need to handle any "special events" as they are not event-based, but reports, and they manage the errors differently:
- The `profile/cpu` gadget reports errors when starting or stopping the gadget.
- The `top` category reports errors per report (e.g., [here](https://github.com/kinvolk/inspektor-gadget/blob/355f0295cef04a40cdd7df50e6ffaf9036ab8d7f/cmd/kubectl-gadget/top/tcp.go#L198-L201)) and not per stats (which is what we transform in columns using `BaseParser.Transform`).

In addition, moving `ManageSpecialEvent` into `BaseParser.Transform`, implies that whoever wants to use the `BaseParser`, needs to implement `GetBaseEvent`, which is strange but I think is acceptable. 

## How to use

It should now be easier for `BaseParser` users to use it because they don't need to manage these special cases alone. See https://github.com/kinvolk/inspektor-gadget/pull/789#discussion_r934738002.

## Testing done

I enhenced the unit-tests to cover this.

### DIFF
--- a/cmd/common/utils/events.go
+++ b/cmd/common/utils/events.go
@@ -21,10 +21,10 @@ import (
 	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 )
 
-// ManageSpecialEvent prints the special events. Notice it does not receive the
+// HandleSpecialEvent prints the special events. Notice it does not receive the
 // OutputMode so it will always print the special events messages without any
 // specific format to standard error.
-func ManageSpecialEvent(e *eventtypes.Event, verbose bool) {
+func HandleSpecialEvent(e *eventtypes.Event, verbose bool) {
 	switch e.Type {
 	case eventtypes.READY:
 		// TODO
--- a/cmd/common/utils/events.go
+++ b/cmd/common/utils/events.go
@@ -21,6 +21,9 @@ import (
 	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 )
 
+// ManageSpecialEvent prints the special events. Notice it does not receive the
+// OutputMode so it will always print the special events messages without any
+// specific format to standard error.
 func ManageSpecialEvent(e *eventtypes.Event, verbose bool) {
 	switch e.Type {
 	case eventtypes.READY:
--- a/cmd/common/utils/events.go
+++ b/cmd/common/utils/events.go
@@ -21,7 +21,7 @@ import (
 	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 )
 
-func ManageSpecialEvent(e eventtypes.Event, verbose bool) {
+func ManageSpecialEvent(e *eventtypes.Event, verbose bool) {
 	switch e.Type {
 	case eventtypes.READY:
 		// TODO
--- a/cmd/kubectl-gadget/trace/trace.go
+++ b/cmd/kubectl-gadget/trace/trace.go
@@ -89,7 +89,7 @@ func (g *TraceGadget[Event]) Run() error {
 
 		baseEvent := e.GetBaseEvent()
 		if baseEvent.Type != eventtypes.NORMAL {
-			commonutils.ManageSpecialEvent(baseEvent, g.commonFlags.Verbose)
+			commonutils.ManageSpecialEvent(&baseEvent, g.commonFlags.Verbose)
 			return ""
 		}
 
--- a/cmd/kubectl-gadget/trace/trace.go
+++ b/cmd/kubectl-gadget/trace/trace.go
@@ -33,7 +33,7 @@ type TraceEvent interface {
 	// of type parameter type even if all types in the type parameter's type set
 	// have a field f. We may remove this restriction in Go 1.19. See
 	// https://tip.golang.org/doc/go1.18#generics.
-	GetBaseEvent() eventtypes.Event
+	GetBaseEvent() *eventtypes.Event
 }
 
 // TraceParser defines the interface that every trace-gadget parser has to
@@ -89,7 +89,7 @@ func (g *TraceGadget[Event]) Run() error {
 
 		baseEvent := e.GetBaseEvent()
 		if baseEvent.Type != eventtypes.NORMAL {
-			commonutils.ManageSpecialEvent(&baseEvent, g.commonFlags.Verbose)
+			commonutils.ManageSpecialEvent(baseEvent, g.commonFlags.Verbose)
 			return ""
 		}
 
--- a/cmd/kubectl-gadget/trace/trace.go
+++ b/cmd/kubectl-gadget/trace/trace.go
@@ -21,19 +21,12 @@ import (
 
 	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
-	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 
 	"github.com/spf13/cobra"
 )
 
 type TraceEvent interface {
 	any
-
-	// The Go compiler does not support accessing a struct field x.f where x is
-	// of type parameter type even if all types in the type parameter's type set
-	// have a field f. We may remove this restriction in Go 1.19. See
-	// https://tip.golang.org/doc/go1.18#generics.
-	GetBaseEvent() *eventtypes.Event
 }
 
 // TraceParser defines the interface that every trace-gadget parser has to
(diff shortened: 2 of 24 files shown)


pkg/gadgets/profile/cpu/types/types.go
```diff
@@ -32,3 +32,9 @@ type Report struct {
 	KernelStack []string `json:"kernel_stack,omitempty"`
 	Count       uint64   `json:"count,omitempty"`
 }
+
+// GetBaseEvent is defined to implement commonutils.BaseElement interface
+// required to use commonutils.BaseParser.
+func (e Report) GetBaseEvent() *eventtypes.Event {
+	return nil
```
mauriciovasquezbernal, line 39: hmm, this is weird. Probably report should contain `eventtypes.Event`.

cmd/kubectl-gadget/trace/trace.go
```diff
@@ -21,19 +21,12 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
 type TraceEvent interface {
 	any
-
-	// The Go compiler does not support accessing a struct field x.f where x is
-	// of type parameter type even if all types in the type parameter's type set
-	// have a field f. We may remove this restriction in Go 1.19. See
-	// https://tip.golang.org/doc/go1.18#generics.
```
mauriciovasquezbernal, line 35: What about adding this comment on `BaseElement`?

mauriciovasquezbernal (commented):
I'm not fully convinced about having all types to implement this `GetBaseEvent()` method, however, if we move the implementation of it to `eventtypes.Types` it'll avoid to repeat this definition again and again.

cmd/common/utils/events.go
```diff
@@ -21,6 +21,9 @@ import (
 	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 )
 
+// ManageSpecialEvent prints the special events. Notice it does not receive the
+// OutputMode so it will always print the special events messages without any
+// specific format to standard output.
 func ManageSpecialEvent(e *eventtypes.Event, verbose bool) {
```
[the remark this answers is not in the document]
blanquicet: Agree

cmd/kubectl-gadget/trace/trace.go
blanquicet: Agree

cmd/common/utils/events.go
```diff
@@ -44,6 +46,8 @@ func ManageSpecialEvent(e *eventtypes.Event, verbose bool) {
 			podMsgSuffix = ", pod " + e.Namespace + "/" + e.Pod
 		}
 
-		fmt.Fprintf(os.Stderr, "%s: node %s%s: %s\n", e.Type, e.Node, podMsgSuffix, e.Message)
+		fmt.Fprintf(os.Stderr, "%s: node %s%s: %s\n",
+			cases.Title(language.English, cases.Compact).String(string(e.Type)),
```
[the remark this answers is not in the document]
blanquicet: [`strings.Title`](https://pkg.go.dev/strings#Title) is simpler but it was deprecated. I didn't find any simpler way. Do you any simpler suggestions?

cc @alban

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -37,6 +37,6 @@ func Base(ev eventtypes.Event) Event {
 	}
 }
 
-func (e Event) GetBaseEvent() eventtypes.Event {
-	return e.Event
+func (e Event) GetBaseEvent() *eventtypes.Event {
+	return &e.Event
 }
```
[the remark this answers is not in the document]
blanquicet: I didn't know it can be done. Applied. Thanks for the suggestion!

pkg/gadgets/profile/cpu/types/types.go
blanquicet: Mmmm I don't think `Report` should contain `eventtypes.Event` because:

1. The profile cpu gadget is not an event-based gadget so it would be misleading. 
2. We should always initialize the `event.Type` as `NORMAL` so that it passes [this check](https://github.com/kinvolk/inspektor-gadget/blob/a6db7da168dfa7c4c47b9adac9f8459d731ecf01/cmd/common/utils/parser.go#L109). And, in my opinion, it could create confusion because, again, it is not an event-based gadget. 
3. If we make this change for `Report`, should we also add `eventtypes.Event` to [Container](https://github.com/kinvolk/inspektor-gadget/blob/a6db7da168dfa7c4c47b9adac9f8459d731ecf01/pkg/container-collection/containers.go#L77-L81)? And to the `Stats` of all the top gadgets ([#782](https://github.com/kinvolk/inspektor-gadget/pull/782) where we start using `BaseParser`). They all use `BaseParser` but don't support special events.

I know it is not the cleanest solution, but I think it could be ok with better comments. Check this:

```golang
// GetBaseEvent is defined to implement the commonutils.BaseElement interface so
// that we can use the commonutils.BaseParser methods to parse Report. In
// commonutils.BaseParser.Transform(), we call GetBaseEvent() to check whether
// the element being parsed is a special event or not. Given that the profile
// cpu gadget is not event-based, this method simply returns nil.
func (e Report) GetBaseEvent() *eventtypes.Event {
	return nil
}
```

```golang
// GetBaseEvent is defined to implement the commonutils.BaseElement interface so
// that we can use the commonutils.BaseParser methods to parse Container. In
// commonutils.BaseParser.Transform(), we call GetBaseEvent() to check whether
// the element being parsed is a special event or not. Given that Container is
// not an event, this method simply returns nil.
func (e Container) GetBaseEvent() *eventtypes.Event {
	return nil
}
```

WDYT?

mauriciovasquezbernal: TL;DR: I think we're messing up with `eventtypes.Event` and we need to fix it.

`eventtypes.Event` is used to carry **messages** from the tracer to the client. Each time a gadget produces output it's marshaled into a structure that contains this type, then it's unmarshaled on the client and according to the kind of gadget it could be printed to the screen (trace, snapshot), or saved for future processing (top). I think the `Event` name is misleading, as this is actually used for gadgets that are not event based, like snapshot. 

I see that `eventtypes.Event` is like an "message header" that indicates the kind of event and some debugging information. I'm thinking to rework this in a different way:

1. Rename Event to `DataHeader` and remove `CommonData` from it, leaving only the type and message.
2. Modify types of different gadgets to embed this structure within their types. 
   - Tracers on their Event types (as currently done, they'll have to add CommonData since it's not on eventtypes.Event anymore)
   - Toppers on their "[Event](https://github.com/kinvolk/inspektor-gadget/blob/120ffd1b1ff12012ab7ca1ee6f93b2fea18843de/pkg/gadgets/top/file/types/types.go#L75)" (to be renamed?) type too
   - Profilers on their [Report](https://github.com/kinvolk/inspektor-gadget/blob/120ffd1b1ff12012ab7ca1ee6f93b2fea18843de/pkg/gadgets/profile/cpu/types/types.go#L26) type
   - I think snapshotters should define a new type that wraps the current event:
```golang
type Event struct {
  eventtypes.DataHeader

  Processes/Sockets []Process/Socket
}

type Process/Socket struct {̣ <-- current Event type
```

3. We can know handle the special event in the same way in all gadgets. (Their "Event" have the same header).

blanquicet: I like your proposal! It would make things clearer. I will pause this PR until then.

blanquicet:
Moving this PR to draft after [this](https://github.com/kinvolk/inspektor-gadget/pull/796#discussion_r949431381) discussion.

pkg/gadgets/profile/cpu/types/types.go
blanquicet: @mauriciovasquezbernal there already exists an issue for this? If it doesn't, I will create one, move your proposal there and close this PR.

mauriciovasquezbernal: I don't think we have an issue for it.

blanquicet:
I opened https://github.com/inspektor-gadget/inspektor-gadget/issues/1130 to think about a better solution for this problem, as this PR is too tricky and not clear enough as a solution.

PR #803, opened 2022-08-03
Test Inspektor Gadget scaling

eiffel-fl:
Hi.

In this PR, I tested Inspektor Gadget scaling regarding number of nodes and so number of generated events.
All the details are in [experiment.pdf](https://github.com/kinvolk/inspektor-gadget/blob/fe88772d97fd3d72a50c5b5ef03ee4d2fa16e9d0/test-scaling/experiment.pdf).

Best regards.


mauriciovasquezbernal:
First of all, thanks a lot for these tests, I think they provide valuable information.

stress-ng produces a lot of events on a short period of time (burst). For each call to exec() we need to (1) run the eBPF program (2) read the event from the gadget tracer manager (3) pass it to the kubectl-gadget client (4) parse and print it. I think if an application is trying to produce as many as possible events we'll for sure miss some of them, even when tracing a single node.

Maybe we should consider testing a scenario where all nodes produce a lower number of events (100 per second?) and then check how many nodes we're able to support with that?

> Inspektor Gadget reports less events than actual number of generated exec(), this can be caused by difference of counting between stress-ng and Inspektor Gadget

I think it's fundamental to understand if the difference is because they count the events on a different way or because Inspektor Gadget is losing events. Did you see any warning message indicating events were lost? I think it's something we should check, losing events should be ok but we must inform the user about it. 

I think a future work for this is to make a more granular testing to understand where the bottleneck is. However, I don't know if this is something we want to do now or better to wait on the outcome of https://github.com/kinvolk/inspektor-gadget/issues/673. 
- single node scalability: check how many events per second are we able to catch on a single node by only using the tracer. (maybe reusing this example https://github.com/kinvolk/inspektor-gadget/tree/main/examples/gadgets/trace/exec). 
- [gadget tracer manager stream](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/pkg/gadgettracermanager/stream/stream.go): How many events per second it's able to handle with a single subscriber? What about more subscribers? 
- gadget tracer manager (server) <-> gadget tracer manager (client): To stream events, we execute the gadget tracer manager as [client](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/gadget-container/gadgettracermanager/main.go#L129) that connects to the [server](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/pkg/gadgettracermanager/gadgettracermanager.go#L77). It'd be nice to check the performance of this interface based on grpc over a unix socket. 
- kubectl-gadget <-> gadget tracer manager: We use the [kube exec](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/cmd/kubectl-gadget/utils/trace.go#L970) API to stream those events from the node to the client. 
- kubectl-gadget parsing performance: How much does it cost to parse and print the event?

test-scaling/script.sh
```diff
@@ -0,0 +1,121 @@
...
+function post_run {
+	local namespace=$1
+
+	kubectl delete ns $namespace
+	../kubectl-gadget undeploy
+}
+
+prepare
+
+for nodes_nr in 2 5 11 17 23 31 39; do
```
mauriciovasquezbernal, line 105: Why these numbers? I usually see powers-of-two are used.

test-scaling/script.sh
```diff
@@ -0,0 +1,121 @@
...
+    matchLabels:
+      k8s-app: stress-exec
+  template:
+    metadata:
+      labels:
+        k8s-app: stress-exec
+    spec:
+      initContainers:
+      - name: stress-exec
+        image: $full_image_name
```
mauriciovasquezbernal, line 64: Any reason not  to use `ghcr.io/colinianking/stress-ng:master` directly here and pass the command?

flyth:
A couple of thoughts about your stress tests - and I might be wrong on some of them... :)
Also, I started writing this parallel to @mauriciovasquezbernal's feedback, so a couple of things will overlap with his feedback.

From what I've seen, I think there's only a couple of bottlenecks that have directly to do with scaling (as right now communication is done through the kubernetes api server via exec):
* network bandwidth between the client node and the kubernetes api server
* network bandwidth between the kubernetes api server and the nodes
* if network bandwidth is not an issue, CPU usage might become one - first on your client node (as it has to do the heavy json decoding + outputting/saving to disk), and if that's not the bottleneck, the network stack will probably be next

I think all other bottlenecks won't apply more to a scaled installation than they would on a single node. I totally agree with @mauriciovasquezbernal that we should use less events to test scalability (basically making sure we never hit 100% in total, so the stresstester + gadget handling).

### Constraints
You're using execsnoop for your tests, so the following should be some constraints for that specific gadget:

* the BPF_MAP_TYPE_PERF_EVENT_ARRAY `events` gets a buffer assigned for each CPU core that is 262.144 bytes long (assuming the pagesize is 4096)
* an `event` takes 7726 bytes, for simplicity let's assume it's 8192 bytes - that means each event buffer can only store 32 entries after which it has to be read by the gadget or gets thrown away - the user won't be notified in this case.
* the `execs` map, which "caches" information on the event can hold 10240 entries; so there can be 10240 exec calls simultaneously, which should be more than fine (I think compared to the event buffer, this is probably too much, we should maybe shift more from here (10240 * 8192 = 84MB) to the event array buffer)
* whenever these event get read from IG, they will be serialized and pushed throgh the pub/sub system, which currently holds a maximum of `SubChannelSize` (currently 250) entries, after which events will be dropped for the subsciber - this will however lead to a notification for the user.
* a serialized event takes approx. 176 bytes, so 10k events would result in ~1.7MB + network overhead

### Bottlenecks
I think the biggest bottleneck is the buffer length of the `events` BPF_MAP_TYPE_PERF_EVENT_ARRAY, especially in a situation where you stress the CPU to 100%, which probably leads to most of the unreported events as the gadget has just not enough time to run and do its stuff (json encoding / sending etc.). If you would increase the buffer size to hold 15k events (from your burst execing), I'm sure it will look a lot better (as said above, I think those 10240 entries of the `execs` map would be better spent here).

After that it's probably network speed again that would be the limiting factor (not fast enough sending would lead into dropping events in the pub/sub manager). More likely in case of scaling on the api server than on the individual nodes, though.

And only after that CPU would be a bottleneck again - probably the JSON encoding which could be sped up a bit by using protocol buffers / gRPC.

flyth:
> The user is normally notified about this:

Ah, I missed that one, sorry :) Good to know, thanks!

mauriciovasquezbernal:
> an event takes 7726 bytes, for simplicity let's assume it's 8192 bytes - that means each event buffer can only store 32 entries after which it has to be read by the gadget or gets thrown away - the user won't be notified in this case.

hmm, exec events are huge and 32 is indeed a very small number. We should increase the size of that buffer. 

> So, the problem showed by this experiment is surely a problem of network/communication because the gadgettracermanagers, which live on each nodes, are able to collect the events for their respective node

I remember that we internally discussed you were performing some experiments by using a local application that only prints to the terminal (to avoid the network bottleneck). What are the results of those experiments? Do the number reported by stress-ng and the "local application" match? 

I did a similar test using trace open. It's able to receive all the events without loses when there is a sleep of some nanoseconds, otherwise it loses events. (code available in https://github.com/kinvolk/inspektor-gadget/tree/mauricio/trace-open-performance-experiment).

mauriciovasquezbernal:
> So, the problem showed by this experiment is surely a problem of network/communication because the gadgettracermanagers, which live on each nodes, are able to collect the events for their respective node.

Where did you execute kubectl-gadget? On the cluster or in your local machine?

mauriciovasquezbernal:
I'm wondering what to do with this PR. I think it adds value but I'm not to convinced of merging +17k LOC to test the scalability. Maybe we can generate the report in markdown format and include it under the documentation?

mauriciovasquezbernal:
But should we close it? It seems a bit weird to let a PR opened there to share some findings, maybe an issue / doc / entry on the wiki is better? (Anyway, it's not very relevant, we can decide later on what to do).

mauriciovasquezbernal:
> For the moment, I think we can let it as it is. Particularly because it would be good to have the numbers for #1096.

Yes, I think we need to gather numbers again once https://github.com/inspektor-gadget/inspektor-gadget/pull/1396 is merged.

mauriciovasquezbernal:
I defined a "testing plan" on https://github.com/inspektor-gadget/inspektor-gadget/issues/1429. What do you think about reusing the work done in this PR to implement a testing framework, then we can start running the actual tests?

mauriciovasquezbernal:
> Moreover, this PR will celebrate its birthday this week, so I guess this is time to review it!

TBH I don't have plans to review it. IMO it's useless as we want to have a more comprehensive testing as stated in https://github.com/inspektor-gadget/inspektor-gadget/issues/1429.

test-scaling/script.sh
```diff
@@ -0,0 +1,123 @@
+#! /usr/bin/env bash
+
+registry='francisregistry.azurecr.io'
+image_tag='francis-test-scaling'
```
vfalico, line 4: If it's in the tree - the user must be able to run it.

Also, having "test-scaling" in the root is not ideal, as we'll have different types of tests.
tests/scaling-nodes?

vfalico (changes requested):
I see more harm than good in adding this PR in its current state.

Harm - it measures the kernel, not Inspektor Gadget, scaling. And it incorporates testing results into the tree - which need to be maintained and clutter the tree.
Good - it's a very good base for eventual future test cases - where stress-ng can be changed to do IO stress etc.

So I'd first address these issues before committing it.

PR #810, opened 2022-08-10
Align JSON naming for Types

eiffel-fl:
Hi.

This PR aligns JSON naming for Types by using `lowerCamelCase`.

Best regards.

### DIFF
--- a/docs/guides/profile/block-io.md
+++ b/docs/guides/profile/block-io.md
@@ -127,18 +127,18 @@ You can also use JSON as output for this gadget:
 ```bash
 $ kubectl gadget profile block-io --node worker-node -o json > /tmp/output.json; jq < /tmp/output.json
 {
-  "val_type": "usecs",
+  "valType": "usecs",
   "data": [
     {
       "count": 0,
-      "interval-start": 1,
-      "interval-end": 1
+      "intervalStart": 1,
+      "intervalEnd": 1
     },
     ...
     {
       "count": 0,
-      "interval-start": 4096,
-      "interval-end": 8191
+      "intervalStart": 4096,
+      "intervalEnd": 8191
     }
   ]
 }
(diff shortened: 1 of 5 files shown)


pkg/gadgets/trace/network/types/types.go
```diff
@@ -24,31 +24,31 @@ import (
...
 	/* pod, svc or other */
-	RemoteKind string `json:"remote_kind,omitempty"`
+	RemoteKind string `json:"remoteKind,omitempty"`
 
-	PodHostIP string            `json:"pod_host_ip,omitempty"`
-	PodIP     string            `json:"pod_ip,omitempty"`
-	PodOwner  string            `json:"pod_owner,omitempty"`
-	PodLabels map[string]string `json:"pod_labels,omitempty"`
+	PodHostIP string            `json:"podHostIp,omitempty"`
+	PodIP     string            `json:"podIp,omitempty"`
```
mauriciovasquezbernal, line 36: `Ip` -> `IP`. (as done in Kubernetes https://github.com/kubernetes/api/blob/v0.24.3/core/v1/types.go#L3839).

mauriciovasquezbernal (commented):
Besides that comment LGTM. (Willa approve later on)

pkg/gadgets/trace/network/types/types.go
mauriciovasquezbernal: I see that in Kubernetes it's called with the full name always: `Service`, maybe we should switch to that.

mauriciovasquezbernal: I was taking about it. Kubernetes always uses `Service` instead of the abbreviation, so if we want to be fully aligned we should name those like `remoteServiceNamespace` and so on.

mauriciovasquezbernal (approved):
LGTM! Thanks!

PR #820, opened 2022-08-16
Add `Dockerfile` to build eBPF object files

eiffel-fl:
Hi.

In this PR, I added a new `Dockerfile` which provides everything needed to build eBPF object files.

Best regards.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -60,6 +60,19 @@ installing the compilers to build it.
 If you push the container images to another registry, you can use the `--image`
 argument when deploying to the Kubernetes cluster.
 
+### Building the eBPF object files
+
+If you need to compile the eBPF code of the gadgets, the ebpf-objects target
+will help you in this task:
+
+```bash
+$ make ebpf-objects
+...
+go: downloading github.com/giantswarm/crd-docs-generator v0.7.1
+...
+Wrote /work/pkg/gadgettracermanager/containers-map/containersmap_bpfel.go
+```
+
 ### Building notes
 
 - The compilation uses `tools/image-tag` to choose the tag of the container


Dockerfiles/build-ebpf-objects.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.18
+# clang and llvm are needed by bpf2go.
+# gcc-multilib is needed for <asm/types.h>.
+# libelf-dev is needed to compile libbpf.
+RUN apt-get update && apt-get install -y clang llvm gcc-multilib libelf-dev
```
mauriciovasquezbernal, line 5: Can we stick to specific versions of those packages? (at least, clang-11, llvm-11)

Dockerfiles/build-ebpf-objects.Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+# Let's install cutting edge libbpf.
+RUN git clone https://github.com/libbpf/libbpf.git && cd libbpf/src && mkdir build && OBJDIR=build make install -j
```
mauriciovasquezbernal, line 7: Same here, please stick to a fixed version of libbpf.

mauriciovasquezbernal (commented):
What about publishing this image and creating a rule on the makefile with the run command, so developers don't need to take care of building nor running the container:

```
ebpf-objects:
	docker run --rm -t --name ebpf-object-builder -v $(pwd):/work ghcr.io/kinvolk/inspektor-gadget-ebpf-builder
```

?

Dockerfiles/build-ebpf-objects.Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /work
+# The resulting docker image should be run with:
+# docker run --rm -t --name ebpf-object-builder -v/path/to/inspektor-gadget:/work name-of-this-image
+ENTRYPOINT make ebpf-objects
```
mqasimsarfraz, line 11: Wouldn't this only generate bpf objects for a single target like amd64? or the arm64 support is handled differently?

Dockerfiles/build-ebpf-objects.Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.18
+# clang and llvm are needed by bpf2go.
+# gcc-multilib is needed for <asm/types.h>.
+# libelf-dev is needed to compile libbpf.
+RUN apt-get update \
+	&& apt-get install -y clang-11 llvm-11 gcc-multilib libelf-dev \
+	&& update-alternatives --install /usr/local/bin/llvm-strip llvm-strip $(which llvm-strip-11) 100 \
+	&& update-alternatives --install /usr/local/bin/clang clang $(which clang-11) 100
+# Let's install cutting edge libbpf.
```
mqasimsarfraz, line 9: Minor nit. Comment needs an update.

Dockerfiles/build-ebpf-objects.Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.18
```
mauriciovasquezbernal, line 1: Call this file `ebpf-builder.Dockerfile`?

mauriciovasquezbernal:
Besides some nits the PR looks very good. I think there are further improvements that we can do later on: (I can create issues for it after it's merged)
- Automatically build and push the builder image using Github Actions
- Pre-download go packages in the image to avoid having to download them each time it's run. (RUN --mount could help)
- Provide documentation to indicate how to compile a single gadget (compiling all gadgets takes too long and could be annoying when trying modifications on a single gadget)
- Compile all the objects with this image and push them to the repo
- Check that compiled objects are updated on the CI

mauriciovasquezbernal (commented):
Right, I again wrote the comment in the wrong box...

mauriciovasquezbernal:
I also requested a review from @alban as he has experience building things inside containers :D.

Makefile
```diff
@@ -53,7 +55,11 @@ all: build local-gadget
 phony_explicit:
 
 ebpf-objects:
-	TARGET=$(GOHOSTARCH) go generate ./...
+	docker run --rm --name ebpf-object-builder -v $(shell pwd):/work ghcr.io/kinvolk/inspektor-gadget-ebpf-builder
```
alban, line 58: You could add:
```
--user $(shell id -u):$(shell id -g)
```
so that files don't belong to root.

(as it is done in the `lint` Makefile target below)

Dockerfiles/ebpf-builder.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# To avoid hitting this:
+# failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
+# when run as non root.
+ENV GOCACHE=/tmp/
+# Create a directory which can be read, written and executed by everyone, this
+# avoid trouble when running as non root.
+RUN mkdir -m 777 /work
+WORKDIR /work
+# The resulting docker image should be run with:
+# docker run --rm --name ebpf-object-builder --user $(shell id -u):$(shell id -g) -v $(shell pwd):/work name-of-this-image
```
mauriciovasquezbernal, line 23: `shell` is only needed when running from the Makefile.

mauriciovasquezbernal:
> What do you mean by this?

We should use this image to compile and update the objects checked on the repo. Running `make ebpf-objects` makes a lot of them as modified and it shouldn't happen if I didn't change the source code.

Makefile
```diff
@@ -53,7 +53,11 @@ all: build local-gadget
 phony_explicit:
 
 ebpf-objects:
-	TARGET=$(GOHOSTARCH) go generate ./...
+	docker run --rm --name ebpf-object-builder --user $(shell id -u):$(shell id -g) -v $(shell pwd):/work ghcr.io/kinvolk/inspektor-gadget-ebpf-builder
```
mauriciovasquezbernal, line 56: break this line?

mauriciovasquezbernal: No a big deal, I was just trying to make it more readable since IMO a 140 long line is difficult to handle.

mauriciovasquezbernal (approved):
> But it explodes because it cannot overwrite some files which were created by root. I will fix this with some chown

I'm hitting the same problem. I fixed it with:

```
$ sudo chown 1000:1000 pkg/gadgets/*/*/tracer/*.o
```

Maybe we should put a note on the documentation about it, otherwise I think many folks will hit this same problem and it's not immediate to realize what's going on.

LGTM otherwise.

PR #822, opened 2022-08-18
pkg/gadgets: Uniform naming for mount namespace filtering map

### LINKED ISSUE #662: gadgets: Uniform filtering map name
All gadgets have an eBPF map used to filter the events. This map is sometimes called `mount_ns_set` and others `filter`. We should use the same name everywhere. 

Ref https://github.com/kinvolk/inspektor-gadget/pull/654#discussion_r871551754

mqasimsarfraz:
This PR makes the mount namespace filtering map name consistent across all the gadgets.

## Fixes
#662 

## Testing done
Successful run for integration tests and manually verifying some gadgets e.g `profile/cpu`, `top/block-io` etc


mauriciovasquezbernal:
Looks very good overall. Two minor nits:
- Please squash both commits into a single one. Otherwise the first commit would be broken making it difficult to use `git bisect` in the future. 
- Add a prefix to the commit indicating the area where the change happens, like `pkg/gadgets` in this case.

mqasimsarfraz:
@mauriciovasquezbernal thanks for the review. I updated the PR!

mauriciovasquezbernal (approved):
Actions on https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/2883315717 are passing, so I'll approve.

PR #823, opened 2022-08-18
ci: Push core images to our container repository.

### LINKED ISSUE #807: CI: Push CO-RE container image to container registry
https://github.com/kinvolk/inspektor-gadget/pull/608 introduced a different container image with only CO-RE compatible gadgets. We aren't pushing that image anywhere, but I think it's a good idea to start pushing it to the container registry. 

To avoid a name collision, we can add a `core` suffix to the name:

```
docker pull ghcr.io/kinvolk/inspektor-gadget:v0.7.1

docker pull ghcr.io/kinvolk/inspektor-gadget:v0.7.1-core
```

eiffel-fl:
Hi.

In this PR, I enabled pushing for CO-RE only container image.
CO-RE only container images are pushed with ["-core" concatenated to their image tag](https://github.com/orgs/kinvolk/packages/container/inspektor-gadget-dev/36530150?tag=francis-push-core-image-core), this avoid creating a new repository for this image type.
It fixes #807.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -197,9 +197,8 @@ jobs:
         file: /home/runner/work/inspektor-gadget/inspektor-gadget/gadget-${{ matrix.type }}.Dockerfile
         build-args: |
           ENABLE_BTFGEN=true
-        # At the moment, we only build core image and do not publish them.
-        push: ${{ matrix.type != 'core' }}
-        tags: ${{ steps.choose-repo-determine-image-tag.outputs.container-repo }}:${{ steps.choose-repo-determine-image-tag.outputs.image-tag }}
+        push: true
+        tags: ${{ steps.choose-repo-determine-image-tag.outputs.container-repo }}:${{ steps.choose-repo-determine-image-tag.outputs.image-tag }}${{ matrix.type == 'core' && '-core' || ''}}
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
LGTM but I'd like to have a second opinion on this. Particularly because that `-core` suffix could create confusion.

alban (commented):
I think it is fine to have the `-core` suffix in the tag. Other container images (https://hub.docker.com/_/golang/) have similar suffix for different kind of builds.

But we have to be careful not to name our branches with a `-core` suffix to avoid possible overlap. Alternatively, it would be safer to always have a suffix, e.g. `foo-default` and `foo-core` instead of `foo` and `foo-core`.

blanquicet:
> Particularly because that -core suffix could create confusion.
>
Yeah, it could, but I don't see better options. To mitigate that, we should describe this image in the [documentation](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/ci.md#container-repository).

alban:
> In this case, I would expect the image to be foo-core-core, so this should not be a problem.

My concern is if we have two branches named as follow:
1. Branch `foo`
 a. Default image: `foo`
 b. Core image: `foo-core`
2. Branch `foo-core`
 a. Default image: `foo-core`
 b. Core image: `foo-core-core`

The images 1.b. and 2.a. collide.

PR #825, opened 2022-08-18
cmd/kubectl-gadget: Rename maxRows to max-rows

mauriciovasquezbernal:
We're using snake-case for CLI parameters.


flyth (approved):
Looks good.

PR #854, opened 2022-08-23
pkg/bpfstats: introduce and use HOST_ROOT for bpf_stats_enabled fallback

### LINKED ISSUE #850: top ebpf fails on AKS because of read-only /proc
## Description

`kubectl gadget top ebpf` fails on AKS (Linux 5.4.0-1089-azure):
```
failed to create tracer: could not enable stat collection: open /proc/sys/kernel/bpf_stats_enabled: read-only file system
```

* With the command `kubectl gadget top ebpf`, the error message is difficult to read because the screen gets refreshed every second.
* With `-o json`, the error message is easier to read

## Impact

Cannot use `top ebpf` on AKS

1. **Set-up**: AKS (Linux Linux 5.4.0-1089-azure)
2. **Task**:
3. **Action(s)**: [ sequence of actions that triggered the bug, see example below ]
  a. `kubectl gadget top ebpf`
  b. `kubectl gadget top ebpf -o json`
4. **Error**:
```
Error: failed to run gadget on node "aks-agentpool-37135575-vmss000000": failed to create tracer: could not enable stat collection: open /proc/sys/kernel/bpf_stats_enabled: read-only file system
```

## Expected behavior

* Errors are displayed correctly without being erased even without using `-o json`
* Fall back to `bpf_stats_enabled` works fine.

## Additional information

`BPF_ENABLE_STATS` requires Linux 5.8 so it falls back on `/proc/sys/kernel/bpf_stats_enabled`. However, the gadget pod has `/proc` mounted read-only, so this fails.

Possible solution:
1. Mount /proc in read-write mode
2. Use `/host/proc/sys/kernel/bpf_stats_enabled`, which is mounted in read-write mode.

flyth:
# Introduce and use HOST_ROOT for bpf_stats_enabled fallback

Instead of using `/proc/sys/kernel/bpf_stats_enabled` directly, this PR prefixes it with the environment variable `HOST_ROOT`, which defaults to `/host`.

## How to use

This will be enabled by default and used whenever the fallback for enabling ebpf stat collection is used.

Fixes #850

### DIFF
--- a/pkg/bpfstats/bpfstats.go
+++ b/pkg/bpfstats/bpfstats.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/cilium/ebpf"
@@ -63,7 +65,7 @@ func EnableBPFStats() error {
 	s, err := ebpf.EnableStats(unix.BPF_STATS_RUN_TIME)
 	if err != nil {
 		// Use fallback method
-		err = ioutil.WriteFile("/proc/sys/kernel/bpf_stats_enabled", []byte("1"), 0o644)
+		err = ioutil.WriteFile(filepath.Join(os.Getenv("HOST_ROOT"), "/proc/sys/kernel/bpf_stats_enabled"), []byte("1"), 0o644)
 		if err != nil {
 			return fmt.Errorf("could not enable stat collection: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


flyth:
@alban: Do we need to address setting the environment variable to something else (https://github.com/kinvolk/inspektor-gadget/blob/main/cmd/kubectl-gadget/deploy.go#L307)?

alban (approved):
LGTM. Thanks!

PR #859, opened 2022-08-24
Add GetContainerDetails

idangutman-ms:
# Add `GetContainerDetails` function

This function provides container details, such as PID, mounts and CgroupsPath.

This PR continues previous PR #819


mauriciovasquezbernal (approved):
IMO It's good to go. I'd like to get another review from @alban  or @blanquicet  since they're mot familiar with this code.

alban (approved):
Thanks!

PR #860, opened 2022-08-24
[RFC] Generate documentation for json types returned by gadgets

alban:
# Generate documentation for json returned by gadgets

Each gadget can return a stream of json events of different types and it is not consistently documented.

This PR is an attempt to automatically generate documentation for it.

Related: https://github.com/kinvolk/inspektor-gadget/issues/856

### DIFF
--- a/cmd/gen-doc/gen-doc.go
+++ b/cmd/gen-doc/gen-doc.go
@@ -20,8 +20,11 @@ import (
 	"html/template"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/giantswarm/crd-docs-generator/pkg/crd"
 	"github.com/giantswarm/crd-docs-generator/pkg/metadata"
@@ -40,6 +43,7 @@ func init() {
 
 type GadgetData struct {
 	Name        string
+	Pkg         string
 	Description string
 	OutputModes []string
 	Operations  []GadgetOperation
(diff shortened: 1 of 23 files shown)


cmd/gen-doc/gen-doc.go
```diff
@@ -57,8 +61,11 @@ var gadgetTemplate string
 
 func getTraceFactories() (ret []GadgetData) {
 	for name, factory := range gadgetcollection.TraceFactories() {
+		t := reflect.TypeOf(factory).Elem()
+		pkgPath := strings.TrimPrefix(t.PkgPath(), "github.com/kinvolk/inspektor-gadget/pkg/gadget-collection/gadgets/")
 		ret = append(ret, GadgetData{
 			Name:        name,
+			Pkg:         pkgPath,
```
alban, line 68: I think a better implementation would be to add a method `GetEventPrototype()` to the Factory and then use `reflect` to find the package name. That would avoid to hard code the string replacement of `pkg/gadget-collection/gadgets/` by `pkg/gadgets/.../types`.

mauriciovasquezbernal:
What's the advantage on using this over the documentation generated by godoc?, for instance https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.10.0/pkg/gadgets/snapshot/process/types#Event?

mauriciovasquezbernal:
> What's the advantage on using this over the documentation generated by godoc?, for instance https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.10.0/pkg/gadgets/snapshot/process/types#Event?

@alban do you think this PR is still relevant?

PR #874, opened 2022-08-25
pkg/container-utils/docker: Fix bad state

mauriciovasquezbernal:
The correct variable to use is State instead of Status. It was causing
to always have the State as "unknown", causing
"./local-gadget list-containers" not to show any docker containers.

Fixes: 6444282dc5d1 ("Replace `PidFromContainerID` for `GetContainerDetails`, providing more details")


mqasimsarfraz (approved):
LGTM. I tried it locally and I could see the running docker runtime containers after this fix.

blanquicet:
cc @idangutman-ms

PR #876, opened 2022-08-26
pkg/gadgets/top/ebpf: add information about referenced maps and their size

### LINKED ISSUE #875: [RFE] Add memory usage of referenced maps to top/ebpf
## Current situation

Right now, top/ebpf tracks runtime and run count of ebpf programs. It could be helpful to see the size of maps associated with the programs as well.

## Ideal future situation

The user is able to also see the size of all maps referenced by a program.

## Implementation options

cilium/ebpf lets one already get all associated maps ids - we just need to get the list, resolve the ids and add up the information on the maps.

## Additional information

Maps might be referenced by more than one program, so the sum of all maps might be larger than the used space. Also, not all memory has to be reserved right away.

flyth:
# Add information about number and size of referenced maps

This PR will add the number and the maximum size of the referenced maps to the programs. It does so by getting the referenced map ids using the cilium/ebpf library and then getting information about the defined sizes for that map (then it's basically (keysize+valuesize)*maxentries).

## How to use

```
$ kubectl gadget top ebpf --sort mapmemory
NODE             PROGID   TYPE             NAME             PID     COMM                      RUNTIME   RUNCOUNT      MAPMEMORY MAPCOUNT
minikube         3460     Kprobe           vfs_write_entry  4169649 gadgettracerman         467.908µs        257     40.953 MiB        4
minikube         3459     Kprobe           vfs_read_entry   4169649 gadgettracerman         380.821µs        414     40.953 MiB        4
minikube         3458     Tracing          gadget_ebpftop   4169649 gadgettracerman          90.351µs       1088         12   B        1
minikube         60       CGroupDevice                      1765    systemd                        0s          0          0   B        0
...
```

Fixes #875

### DIFF
--- a/cmd/kubectl-gadget/top/ebpf.go
+++ b/cmd/kubectl-gadget/top/ebpf.go
@@ -28,6 +28,7 @@ import (
 	gadgetv1alpha1 "github.com/kinvolk/inspektor-gadget/pkg/apis/gadget/v1alpha1"
 	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/top/ebpf/types"
 
+	"github.com/docker/go-units"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
 )
@@ -62,6 +63,8 @@ func newEbpfCmd() *cobra.Command {
 				"comm",
 				"runtime",
 				"runcount",
+				"mapmemory",
+				"mapcount",
 			},
 		},
 	}
@@ -79,6 +82,8 @@ func newEbpfCmd() *cobra.Command {
 		"totalruncount": 13,
 		"cumulruntime":  12,
 		"cumulruncount": 13,
+		"mapmemory":     14,
+		"mapcount":      8,
 	}
 
 	cmd := &cobra.Command{
@@ -281,6 +286,10 @@ func (p *EbpfParser) TransformStats(stats *types.Stats) string {
 				sb.WriteString(fmt.Sprintf("%*v", p.ColumnsWidth[col], time.Duration(stats.CumulativeRuntime)))
 			case "cumulruncount":
 				sb.WriteString(fmt.Sprintf("%*d", p.ColumnsWidth[col], stats.CumulativeRunCount))
+			case "mapmemory":
+				sb.WriteString(fmt.Sprintf("%*s", p.ColumnsWidth[col], units.BytesSize(float64(stats.MapMemory))))
+			case "mapcount":
+				sb.WriteString(fmt.Sprintf("%*d", p.ColumnsWidth[col], stats.MapCount))
 			}
 			sb.WriteRune(' ')
 		}
--- a/docs/guides/top/ebpf.md
+++ b/docs/guides/top/ebpf.md
@@ -14,10 +14,10 @@ So first, start `top ebpf` in a terminal. You should see something like:
 
 ```bash
 $ kubectl gadget top ebpf
-NODE             PROGID   TYPE             NAME             PID     COMM                      RUNTIME   RUNCOUNT
-minikube         6344     Tracing          ig_top_ebpf_it   540621  gadgettracerman          86.911µs       1071
-minikube         62       CGroupDevice                      2378    systemd                        0s          0
-minikube         48       CGroupSKB                         2378    systemd                        0s          0
+NODE             PROGID   TYPE             NAME             PID     COMM                      RUNTIME   RUNCOUNT      MAPMEMORY MAPCOUNT
+minikube         503      Tracing          ig_top_ebpf_it   573222  gadgettracerman           54.09µs       1069            12B        1
+minikube         187      CGroupDevice                                                        2.292µs          1             0B        0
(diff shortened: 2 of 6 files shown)


flyth:
I wasn't too sure about the column name `MapMemory` - we could also use MapSizes (but I would probably expect a list of sizes, then) or MapsSize (looks strange). Also, it doesn't necessarily reflect the real size, but the maximum size that all referenced maps could take in total...

alban (commented):
I was initially thinking that it should be a different view for ebpf maps (possibly in a different gadget?) because:
* Maps might not be attached to a ebpf program
* Maps could be attached to several ebpf programs (e.g. in the network-graph gadget or in traceloop)

I hadn't thought about adding the map data to the program view. But I think it's useful too. Maybe both?

Is it possible to see if the flag `BPF_F_NO_PREALLOC` is set on maps?

flyth:
> I hadn't thought about adding the map data to the program view. But I think it's useful too. Maybe both?

Yeah, I agree. A dedicated view for maps (in a separate gadget/mode) would make sense, but I too thought this is already helpful as it gives an instant view on what a program has access to (it shows the ~85MB for the exec tracer).

It seems possible to check for `BPF_F_NO_PREALLOC` since we can get the flags, but for that we would have to add yet another value to difference between the map types. But then it might be also interesting to see how much memory is actually allocated by iterating over the map. Maybe that might be too much to check in this case where it only is additional side info.

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -242,6 +246,19 @@ func (p *EbpfParser) PrintStats() {
 	}
 }
 
+func formatBytes(bytes uint64) string {
```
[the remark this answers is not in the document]
flyth: It does! I changed it. Did a quick search before, found nothing and didn't want to introduce a new dependency just for that. :)

docs/guides/top/ebpf.md
```diff
@@ -40,17 +40,18 @@ Some eBPF programs of type `Kprobe` should pop up, including their runtime and r
...
+minikube         3460     Kprobe           vfs_write_entry  4169649 gadgettracerman         614.455µs        647     40.953 MiB        4
+minikube         24       CGroupDevice                                                        2.458µs          2          0   B        0
+minikube         60       CGroupDevice                      1765    systemd                        0s          0          0   B        0
+minikube         51       CGroupDevice                                                             0s          0          0   B        0
 ...
 ```
 
-So in this case for example, in the past second `vfs_write_entry` has been called 378 times, which took 3.948619ms.
+So in this case for example, in the past second `vfs_write_entry` has been called 647 times, which took 614.455µs.
+The program references 4 maps that have a total maximum size of 40.953 MB.
```
alban, line 54: I think the documentation should make it clearer that the size is not the current memory allocation but the maximum, calculated with `max_entries * (key_size + value_size)`, with those 3 parameters as visible from userspace. Some of the consequences:
* BPF_MAP_TYPE_PERF_EVENT_ARRAY: value_size is not counting the ring buffers, but only their file descriptors (i.e. sizeof(int) = 4 bytes)
* BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS: value_size is not counting the inner maps, but only their file descriptors (i.e. sizeof(int) = 4 bytes)
* PERCPU maps: the reported size is for one cpu, the multiplication by the number of cpus is left as an exercise to the reader

alban (approved):
LGTM

docs/guides/top/ebpf.md
flyth: Added 👍

docs/guides/top/ebpf.md
```diff
@@ -67,3 +68,14 @@ minikube         60       CGroupDevice                      1765    systemd
 minikube         48       CGroupDevice                      1765    systemd                        0s             0
 ...
 ```
+
+### A note about memory usage of maps
+
+The shown value for MapMemory is calculated by using the parameters of the map: `max_entries * (key_size + value_size)`.
+This is the maximum size the map can have, but it doesn't necessarily reflect its actual size. Additionally, maps can
```
alban, line 75:
```suggestion
This is the maximum size the map can have, but it doesn't necessarily reflect its current memory allocation. Additionally, maps can
```

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -176,12 +176,40 @@ func (t *Tracer) nextStats() ([]types.Stats, error) {
...
+	for {
+		nextMapID, err = ebpf.MapGetNextID(curMapID)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				break
+			}
+			return nil, fmt.Errorf("could not get next map ID: %w", err)
+		}
+		if nextMapID <= curMapID {
+			break
```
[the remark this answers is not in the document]
flyth: I have seen this on another iterator (over the programs), so I added that check as well to prevent infinite loops. I'll read up on this later on.

PR #879, opened 2022-08-29
examples: Add missing defer call

mauriciovasquezbernal:


blanquicet:
What do you think about making the [TraceCollection.Close()](https://github.com/kinvolk/inspektor-gadget/blob/3df3dfd34a3e50afcc7dfe9bd0a3fcfe581fe349/pkg/tracer-collection/tracer-collection.go#L192-L193) close all the tracers? If you agree, I will create a simple issue for it.

PR #897, opened 2022-08-30
Run integration tests on AKS with default OS SKU for `amd64` and `arm64`

### LINKED ISSUE #785: Run integration tests on ARM64
https://github.com/kinvolk/inspektor-gadget/issues/359 was about manually checking if the gadgets were working on such platform, but we're not running any integration tests yet. 

The goal of this issue is to also run the integration tests on the ARM64 platform. It's not clear yet if we should run it on all the platforms (minikube, AKS, EKS, GKE).

eiffel-fl:
Hi.

In this PR, I added tests on AKS with default OS SKU for `amd64` and `arm64`.
The tests should run at 9:30 UTC.
The pipeline was tested there:
https://github.com/kinvolk/inspektor-gadget/actions/runs/3021525878

It also fixes #785.

Best regards.

### DIFF
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -81,7 +81,7 @@ $ az aro create --resource-group $RESOURCEGROUP --name $CLUSTER --vnet $VNET --m
 Considerations:
 - After executing the `az aro create` command, it normally takes about 35
   minutes to create a cluster.
-- For the sack of simplicity, we are not providing a [Red Hat pull
+- For the sake of simplicity, we are not providing a [Red Hat pull
   secret](https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster#get-a-red-hat-pull-secret-optional)
   during the cluster creation, so our cluster will not include samples or
   operators from Red Hat or certified partners. However, it is not a requirement
@@ -153,3 +153,49 @@ Notice that it configures the kubectl configuration with a new context.
 
 Please take into account that any change done on this cluster could cause issues
 with the integration tests running on GitHub actions at that moment.
+
+
+## Run integration tests on an AKS cluster
+
+When secrets described in the below sections are set, the integration tests
+will also run on
+[AKS clusters](https://docs.microsoft.com/en-us/azure/aks/intro-kubernetes).
+
+### Create the clusters
+
+As Inspektor Gadget support both `amd64` and `arm64`, two clusters will be
+created:
+
+```bash
+$ subscription=<mySubscription>
+$ subscription_id=<mySubscriptionID>
+$ resourcegroup=<myResourceName>
+$ location=<myLocation>
+
+# Set subscription so that we don't need to specify it at every command
+$ az account set --subscription $subscription
+
+# Create resource group
+$ az group create --name $resourcegroup --location $location
+
+# Create the needed secrets to be able to "az login" from the CI.
+$ az ad sp create-for-rbac --name inspektor-gadget-ci --role contributor \
+	--scopes /subscriptions/${subscription_id}/resourceGroups/${resourcegroup} \
+	--sdk-auth
+{
+# The JSON object containing a secret should be stored as CI secret.
+}
+```
+
+After doing this, you will need to create several secrets:
+
+1. `AZURE_AKS_CREDS`: It stores the JSON object outputted by ``.
+1. `AZURE_AKS_RESOURCE_GROUP`: It stores the name of the resource group where
+the clusters will be created.
+
+The workflow will create the `amd64` and `arm64` clusters for you by using the
+above information.
+By default, each of this cluster features
+[3 nodes](https://learn.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-create).
+Once created, the integration tests will be run on these clusters.
+Finally, the clusters are deleted, whatsoever is the result of the tests.


.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+    - name: Run integration tests
+      uses: ./.github/actions/run-integration-tests
+      env:
+        KUBERNETES_DISTRIBUTION: "aks-${{ matrix.os-sku }}"
+      with:
+        registry: ${{ env.REGISTRY }}
+        container_repo: ${{ env.CONTAINER_REPO }}
+        image_tag: 'latest'
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
```
blanquicet, line 141: As it already succeeded, I suppose we don't need to set any permission for the package for this job.

cc @mqasimsarfraz

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+        creds: ${{ secrets.AZURE_AKS_CREDS }}
+    - name: Create AKS cluster
+      shell: bash
+      run: |
+        cluster_name=${{ secrets.AZURE_AKS_CLUSTER_PREFIX }}${{ matrix.arch }}-${{ matrix.os-sku }}
+
+        node_size='Standard_D2s_v5'
+        if [ ${{ matrix.arch }} = 'arm64' ]; then
+          node_size='Standard_D2ps_v5'
+        fi
```
blanquicet, line 119: Could you please roughly describe in the documentation why we need two diff node sizes?

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18.3
+          cache: true
+        id: go
+      - name: Install debian packages
+        uses: ./.github/actions/install-debian-packages
+      - name: Build kubectl-gadget-${{ matrix.os }}-${{ matrix.arch }}
+        run: |
+          git checkout
```
blanquicet, line 65: I know it is already present in the current implementation but I was wondering why we need to execute this.

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+                "${{ secrets.AZURE_AKS_CLUSTER_PREFIX }}" != "" ]]; \
+          then
+            echo "Secrets to use an ARO cluster were configured in the repo"
+            echo "::set-output name=aks::true"
+          else
+            echo "Secrets to use an ARO cluster were not configured in the repo"
+            echo "::set-output name=aks::false"
+          fi
+
+  build-kubectl-gadget:
```
blanquicet, line 40: Is it possible to create a specific action for this that receives the container image and the tag to be used? Is it possible to upload the binary to the artifact from there?

For this workflow, it will pass `CONTAINER_REPO=${{ env.REGISTRY }}/${{ env.CONTAINER_REPO }}` and `IMAGE_TAG=latest`. While for the _Inspektor Gadget CI_ workflow, they would be the output of the `set-repo-determine-image-tag` step.

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+
+          # Prepare assets for release and actions artifacts
+          platform=$(echo ${kubectl_gadget} | cut -d- -f3-4)
+          mkdir $platform
+          cp $kubectl_gadget $platform/kubectl-gadget
+          cp LICENSE $platform/
+          tar --sort=name --owner=root:0 --group=root:0 \
+            -czf ${kubectl_gadget}.tar.gz -C $platform \
+            kubectl-gadget LICENSE
+          rm -rf $platform
```
blanquicet, line 89: Is this part necessary in this case? I think we could simply upload the kubectl-gadget binary, isn't it?

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
+name: Test cloud platforms
```
blanquicet, line 1:
```suggestion
name: Test Inspektor Gadget on different Kubernetes distributions
```
?

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+env:
+  REGISTRY: ghcr.io
+  CONTAINER_REPO: ${{ github.repository }}
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+on:
+  schedule:
+    # Each week day at 9:30 UTC (so 3:30 MDT and 2:30 MST).
+    - cron: "30 9 * * MON-FRI"
```
blanquicet, line 11: Why should we schedule this while ARO is running when merging on `main` or making a release? The main problem I see with this scheduling is that it will not run on the tagged commit before making a release.

.github/workflows/test-cloud-platforms.yml
```diff
@@ -0,0 +1,149 @@
...
+      shell: bash
+      run: |
+        cluster_name=${{ secrets.AZURE_AKS_CLUSTER_PREFIX }}${{ matrix.arch }}-${{ matrix.os-sku }}
+
+        node_size='Standard_D2s_v5'
+        if [ ${{ matrix.arch }} = 'arm64' ]; then
+          node_size='Standard_D2ps_v5'
+        fi
+
+        # Let's keep thing in the US to avoid data crossing the Atlantic.
```
blanquicet, line 121: Just out of curiosity, is it because you found that everything else (the GH registry and runner) are also running in the US?

blanquicet (commented):
Awesome!! It is great we are adding AKS to our tested distro!!! 🎉 🎉 

I think the workflow is clear and easy and clear, so easy to review! Well done! 

My main concern is when it is triggered. Otherwise, looks great! 💯

.github/workflows/test-cloud-platforms.yml
mqasimsarfraz: That is correct. We only needed `package: write` for releases and since we don't trigger this for any tags we should we good.

.github/workflows/test-cloud-platforms.yml
blanquicet: > I am concerned about the time it takes to setup the cluster and then running the tests.
> But running them only on main should be fine.
>
Running the integration tests on ARO takes around 10 minutes, and according to [this](https://github.com/kinvolk/inspektor-gadget/actions/runs/3021525878), setting up the AKS cluster + running the integration tests took around 12 minutes. I think two additional minutes on the CI, only when merging or releasing, is ok.

> In this case, I think we do not need to have a dedicated file and can add the job to the workflow file we already have.
>
Yeah, in that case, we can keep everything on one single workflow, and we don't duplicate the compilation of `kubectl-gadget`. Thus, this PR should become even simpler.

.github/workflows/test-cloud-platforms.yml
blanquicet: You are right. Maybe it is better to have a comment here.

docs/ci.md
```diff
@@ -153,3 +153,52 @@ Notice that it configures the kubectl configuration with a new context.
...
+# The JSON object containing a secret should be stored as CI secret.
+}
+```
+
+After doing this, you will need to create 3 secrets:
+
+1. `AZURE_AKS_CREDS`: It stores the JSON object outputted by ``.
+1. `AZURE_AKS_RESOURCE_GROUP`: It stores the name of the resource group where
+the clusters will be created.
+1. `AZURE_AKS_CLUSTER_PREFIX`: It stores the prefix which will be used as
```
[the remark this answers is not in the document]
blanquicet: As it is not anymore a secret, the documentation needs to be updated.

.github/workflows/inspektor-gadget.yml
```diff
@@ -451,6 +464,67 @@ jobs:
...
+      shell: bash
+      run: |
+        cluster_name=${{ env.AZURE_AKS_CLUSTER_PREFIX }}${{ matrix.arch }}-${{ matrix.os-sku }}
+
+        echo "CLUSTER_NAME=${$cluster_name}" >> GITHUB_ENV
+    - name: Create AKS cluster ${{ env.CLUSTER_NAME }}
+      shell: bash
+      run: |
+        cluster_name=${{ env.AZURE_AKS_CLUSTER_PREFIX }}${{ matrix.arch }}-${{ matrix.os-sku }}
+
```
blanquicet, line 492:
```suggestion
```
Not needed anymore.

.github/workflows/inspektor-gadget.yml
```diff
@@ -451,6 +464,67 @@ jobs:
...
+    - uses: azure/login@v1
+      name: Login to Azure
+      with:
+        creds: ${{ secrets.AZURE_AKS_CREDS }}
+    - name: Craft cluster name
+      shell: bash
+      run: |
+        cluster_name=${{ env.AZURE_AKS_CLUSTER_PREFIX }}${{ matrix.arch }}-${{ matrix.os-sku }}
+
+        echo "CLUSTER_NAME=${$cluster_name}" >> GITHUB_ENV
```
blanquicet, line 487: nit: Can't this be merged in one single line?

.github/workflows/inspektor-gadget.yml
```diff
@@ -451,6 +464,67 @@ jobs:
 
         perl tools/check-readme.pl ./kubectl-gadget README.md
 
+  test-integration-aks:
+    name: Integration tests on AKS
+    # level: 1
```
blanquicet, line 469:
```suggestion
    # level: 2
```

.github/workflows/inspektor-gadget.yml
```diff
@@ -451,6 +464,67 @@ jobs:
...
+      with:
+        ref: 'main'
+    - name: Run integration tests
+      uses: ./.github/actions/run-integration-tests
+      env:
+        KUBERNETES_DISTRIBUTION: "aks-${{ matrix.os-sku }}"
+      with:
+        registry: ${{ env.REGISTRY }}
+        container_repo: ${{ env.CONTAINER_REPO }}
+        image_tag: 'latest'
```
blanquicet, line 519: Shouldn't you use `./.github/actions/set-container-repo-and-determine-image-tag` to get this info?

blanquicet (commented):
It looks great and, most important, simple! Loving it!

.github/workflows/inspektor-gadget.yml
```diff
@@ -454,6 +466,68 @@ jobs:
 
         perl tools/check-readme.pl ./kubectl-gadget README.md
 
+  test-integration-aks:
+    name: Integration tests on AKS
+    # level: 2
+    needs: [check-secrets, test-unit, build-kubectl-gadget, build-local-gadget, build-gadget-container-images]
+    #needs: [check-secrets, test-unit, build-kubectl-gadget, build-local-gadget, build-gadget-container-images, publish-gadget-images-manifest]
+    if: needs.check-secrets.outputs.aks == 'true'
+    runs-on: ubuntu-latest
```
blanquicet, line 475:
```suggestion
    runs-on: ubuntu-latest
    permissions:
      contents: read
      # Required to be able to push the integration test image
      packages: write
```

Now that it will also run when making a release, I think it is needed. cc @mqasimsarfraz @mauriciovasquezbernal

blanquicet (approved):
After adding the permissions, I think it is ready to go. However, I would like to have another review as it is an important change.

.github/workflows/inspektor-gadget.yml
blanquicet: Could you please rebase this PR now that https://github.com/kinvolk/inspektor-gadget/pull/955 has been integrated?

.github/workflows/inspektor-gadget.yml
```diff
@@ -456,6 +468,70 @@ jobs:
...
+        node_size='Standard_D2s_v5'
+        if [ ${{ matrix.arch }} = 'arm64' ]; then
+          # 'p' means the node size corresponds to arm64 hardware.
+          node_size='Standard_D2ps_v5'
+        fi
+
+        # Let's keep thing in the US to avoid data crossing the Atlantic as
+        # GitHub data centers are in the US:
+        # https://github.blog/2017-10-12-evolution-of-our-data-centers/
+        az aks create -l westcentralus -g ${{ secrets.AZURE_AKS_RESOURCE_GROUP }} -n ${{ env.CLUSTER_NAME }} -s $node_size --os-sku ${{ matrix.os-sku }} --no-ssh-key
```
blanquicet, line 517: It creates a 3 node cluster by default, right?

blanquicet: I think we should add this information also in the documentation.

integration/integration_test.go
```diff
@@ -159,9 +159,9 @@ func testMain(m *testing.M) int {
 		if *k8sDistro == K8sDistroARO {
 			livenessProbe = false
 		}
-		if *k8sDistro == K8sDistroMinikubeGH {
-			imagePullPolicy = "Never"
-		}
+// 		if *k8sDistro == K8sDistroMinikubeGH {
+// 			imagePullPolicy = "Never"
+// 		}
```
blanquicet, line 164: Out of curiosity, why do you need to do this to test AKS?

mqasimsarfraz: I wonder if that is related to the tag fix we did [here](https://github.com/kinvolk/inspektor-gadget/pull/955/files#diff-bda633e49a0d242b6842493c3896d4f766f361166fa4c08e1c2e01d827544cadR211). If the run failed on loading the image, reverting [image_tag](https://github.com/kinvolk/inspektor-gadget/pull/897/files#diff-bda633e49a0d242b6842493c3896d4f766f361166fa4c08e1c2e01d827544cadL546) and the changes here will fix it.

mqasimsarfraz: Makes sense RTFM for me :laughing: Wanted to make sure I didn't break anything :laughing:

PR #901, opened 2022-08-31
pkg/gadgets: Fix Kprobe name for execsnoop on arm64.

eiffel-fl:
Hi.

This PR fixes `K.*probe` names for `execsnoop` on arm64.
With modifications added in 4130c004fead, the gadget would not print anything on arm64.

Best regards.


mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #911, opened 2022-09-05
docs: videos: use minikube

alban:
The 'minikube' node name is smaller, so it looks nicer in the gif videos

I used the following commands:
```
minikube start --driver=kvm2 --container-runtime=containerd --cni=cilium
make build install/kubectl-gadget push-gadget-container
kubectl-gadget deploy
make -C tools/demos/ all-gadgets
make -C tools/demos/ install/demo.gif
make -C tools/demos/ install
```

Follow up from https://github.com/kinvolk/inspektor-gadget/pull/664


blanquicet (approved):
Thanks, they look nicer! 🙂 

For the fsslower example, each entry is still using two lines. What if we use JSON format in that one?

alban:
> For the fsslower example, each entry is still using two lines. What if we use JSON format in that one?

What about implementing `-o wide` to show all the fields, but the default `-o columns` would only select a small selection fields so that all gadgets fit on a small terminal? I think `kubectl` does that already.

blanquicet:
> > For the fsslower example, each entry is still using two lines. What if we use JSON format in that one?
> 
> What about implementing `-o wide` to show all the fields, but the default `-o columns` would only select a small selection fields so that all gadgets fit on a small terminal? I think `kubectl` does that already.

Sounds good! We can start thinking about how to implement it by taking advantage of the formatter that is being introduced in https://github.com/kinvolk/inspektor-gadget/pull/877. BTW, could you please take another look at that PR? It was marked as ready.

PR #913, opened 2022-09-05
Remove "io/ioutil"

eiffel-fl:
Hi.

In this PR, I removed `"io/ioutil"` and use `os` and `io` which provides alternatives since 1.16.

Best regards.


mauriciovasquezbernal (approved):
LGTM, thanks for handling it!

Issue #744, opened 2022-07-12
[RFE] Enhance snapshot socket gadget with PID and container info


blanquicet:
## Current situation

The snapshot socket gadget does not include information about the PIDs/Containers using the socket. 

## Impact

It is particularly useful for `local-gadget` when the POD information cannot be available because the container might not be created via Kubernetes, making the output of `local-gadget` not very useful.

## Ideal future situation

As discussed with @alban:

> In usual use cases, the fd would belong to only one process and therefore one container. But it is possible to have the fd in several pods, for example if the different pods have a host volume to access the same unix socket file and pass fds around. That would be unusual though. In other words, the netns of a socket could be different than the netns of the processes that hold the socket fd.
>

However, we decided to implement only the simplest case for now and create an issue to track that limitation. Therefore, we will start by iteratating over `/proc` until we get the first PID using the inode within the same network namespace where we found the inode.

For IG, it also ensures that the container found will always be coherent with the pod so we can have a clean UI (with Kubernetes metadata for `kubectl-gadget`):

```
$ sudo local-gadget snapshot socket --proto tcp -e
CONTAINER    PROTOCOL    LOCAL                 REMOTE                STATUS         INODE    PID
myCont1      TCP         192.168.0.41:58202    10.96.0.1:443         ESTABLISHED    39380    1234
myCont2      TCP         10.0.2.15:22          10.0.2.2:45256        ESTABLISHED    32683    1455
myCont3      TCP         10.0.2.15:22          10.0.2.2:45264        ESTABLISHED    40037    8743
```

Of course, in the documentation of the gadget, we should specify that the container and PID information will be only displayed if the socket is being used by a process within the same network namespace of the pod. And in the case of multiple processes within the same network namespace using the same fd, only one of them will be displayed.

## Implementation options

Study how we can iterate over `/proc` by using eBPF iterators. @flyth

mauriciovasquezbernal:
Perhaps we can use https://github.com/inspektor-gadget/inspektor-gadget/pull/1201 for this one?

mauriciovasquezbernal:
I think it can be done with the socket enricher (pkg/gadgets/internal/socketenricher). The pid-related information will be from the process creating the socket, but I think that's good enough. @alban any special considerations on this one?

PR #919, opened 2022-09-06
local-gadget: Run snapshot gadgets using tracers instead of the Tracer CR

blanquicet:
# Run snapshot gadgets using tracers instead of the Tracer CR

This PR fixes https://github.com/kinvolk/inspektor-gadget/issues/862.

## Implementation details
- I reverted all the modifications introduced in [#730](https://github.com/kinvolk/inspektor-gadget/pull/730) to run the snapshot gadgets using the Tracer CR.
- I moved all the TraceCR-related code back from the `common` folder to the `kubectl-gadget` folder as it is not going to be used anymore by `local-gadget`.
- I implemented the logic for iterating over network namespaces in the socket gadget, but I keep it disabled until [#744](https://github.com/kinvolk/inspektor-gadget/issues/744) is implemented. Otherwise, we will show a list of sockets without any information about the container where they are running.

## How to use

See new [process](https://github.com/kinvolk/inspektor-gadget/blob/jose/use-tracers-for-snapshot/docs/local-gadget.md#snapshotprocess) examples.


cmd/local-gadget/snapshot/socket.go
```diff
@@ -15,54 +15,79 @@
...
-		"pod",
-		"protocol",
-		"local",
-		"remote",
-		"status",
-		"inode",
-	}
+				// Given that the tracer works per network namespace, we only
+				// need to run it once per namespace.
+				visitedNetNs := make(map[string]struct{})
```
mauriciovasquezbernal, line 45: Can you use uint14 as the map key?

mauriciovasquezbernal (approved):
LGTM. Just a minor comment but otherwise it can go in IMO.

cmd/local-gadget/snapshot/socket.go
blanquicet: You are right. There is no reason to transform it into a string.

Beyond this, I think it is a pity to have this gadget disabled, everything is in place to add `local-gadget snapshot socket`, but we need to solve https://github.com/kinvolk/inspektor-gadget/issues/744. Maybe @alban is interested in working on it?

blanquicet:
Rebased to solve conflicts.

PR #920, opened 2022-09-07
Test and push CO-RE based images

eiffel-fl:
Hi.

This PR enables testing and pushing of CO-RE based images in our CI.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -504,11 +504,14 @@ jobs:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
 
-  test-integration:
+  test-integration-minikube:
     name: Integration tests
     # level: 1
     needs: [test-unit, build-kubectl-gadget, build-local-gadget, build-gadget-container-images]
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        type: [default, core]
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
I think this is right but IMO it's still missing a logic to force standard tracers to be used in order to close https://github.com/kinvolk/inspektor-gadget/issues/808. (In the current state both tests will use the CO-RE version because the kernel is new enough). I agree that it can be done in another PR but then https://github.com/kinvolk/inspektor-gadget/issues/808 should not be closed by this.

mauriciovasquezbernal (approved):
I'm approving but please take a lot at my previous comment.

mauriciovasquezbernal:
> now we are saying we should make it back.

I'm not saying that we should expose this logic to the user, but we should have a way to allow to control it from the tests. Otherwise we don't have a way to test the standard tracers because the fallback mechanism will always use the CO-RE ones in the minikube versions we're running.

mauriciovasquezbernal:
Put some logic in all gadgets to check for an env variable around this? https://github.com/kinvolk/inspektor-gadget/blob/ced763e2987f10bd4abd54126bf9dc5d094f2140/pkg/gadget-collection/gadgets/trace/exec/gadget.go#L117-L130 I think we can easily define an env variable when deploying Inspektor Gadget for the integration tests.

PR #928, opened 2022-09-08
btfgen: Check BTF file exists before calling btfgen.

eiffel-fl:
Hi.

This PR adds a check to verify BTF directory is not empty before calling btfgen.
It is related to aquasecurity/btfhub-archive#6.

Best regards.


mauriciovasquezbernal (approved):
LGTM. Thanks for the fix!

mauriciovasquezbernal:
Perhaps change the title to "btfgen: Check if BTF file exists before calling btfgen"?

mauriciovasquezbernal:
I don't have strong opinions. Sometimes I think failing is good because we realize there's something bad upstream, another times I think this check is good because we avoid our build to fail. I think we can merge to be sure it won't happen again.

Issue #719, opened 2022-06-29
[RFE] ci/integration-tests: Use JSON output format for integration tests


blanquicet:
## Current situation

We execute `kubectl-gadget` using the columns format for integration tests and then we verify the output using regular expressions. The problem is that we are not currently verifying the type of all the fields because the current regular expressions do not match all the columns. 

## Impact

We could miss regressions on those fields that we are not verifying.

## Proposal

Given that the output generated by the `gadget` pod is in JSON format by default, the proposal is that the integration tests run `kubectl-gadget` with JSON output format and then unmarshall the output using the specific types. For instance, for execsnoop, using structs of type [Event](https://github.com/kinvolk/inspektor-gadget/blob/d4000354caac5fb58df556942b3539dbadf41013/pkg/gadgets/execsnoop/types/types.go#L21-L31). Doing so, we are verifying the type of all the fields. In addition, we must verify as many fields as possible. In particular, node, namespace, podname and container name (if available) must always be verified.

To test the transformation from JSON to columns done by the CLI, I propose to test it with unit-tests. After the refactoring of CLI (e.g. PR https://github.com/kinvolk/inspektor-gadget/pull/695), it means to unit-test the implementation of the gadget parsers, e.g. [SocketParser](https://github.com/kinvolk/inspektor-gadget/blob/4786d91d27d229f88850a44ed0d1ebf565a98d17/cmd/kubectl-gadget/snapshot/socket.go#L33) for the snapshot-socket gadget.

## PR tracker 
- Advise: It doesn't apply.
- Audit:
  - [x] Seccomp https://github.com/inspektor-gadget/inspektor-gadget/pull/1104
- Profile:
  - [x] Block-IO https://github.com/inspektor-gadget/inspektor-gadget/pull/1089
  - [x] CPU https://github.com/inspektor-gadget/inspektor-gadget/pull/1089
- Snapshot:
  - [x] Process https://github.com/inspektor-gadget/inspektor-gadget/pull/1087
  - [x] Socket https://github.com/inspektor-gadget/inspektor-gadget/pull/1083
- Top:
  - [x] Block-IO https://github.com/inspektor-gadget/inspektor-gadget/pull/1080
  - [x] ebpf https://github.com/inspektor-gadget/inspektor-gadget/pull/1079
  - [x] file https://github.com/inspektor-gadget/inspektor-gadget/pull/1077
[...]

PR #934, opened 2022-09-09
cli: Integrate columns package on signal and exec trace gadgets

blanquicet:
# Integrate columns package on signal and exec trace gadgets

This PR aims to start the discussion about integrating the `tableformatter` (https://github.com/kinvolk/inspektor-gadget/pull/877) in the CLI. I transformed two gadget to start testing the integration. 

## Implementation details

Given that the `tableformatter` only manages the case where the gadget output needs to be printed in columns, the final implementation cannot use it directly as an implementation of [`TraceParser`](https://github.com/kinvolk/inspektor-gadget/blob/ced763e2987f10bd4abd54126bf9dc5d094f2140/cmd/common/trace/trace.go#L33-L44) (as we did during experiments: [1](https://github.com/kinvolk/inspektor-gadget/commit/51df2bbce6649cfc61a39ea92088d9860e90df5f) and [2](https://github.com/kinvolk/inspektor-gadget/blob/e93399cc5f3cb01c5390969ccf214e005aceed04/cmd/kubectl-gadget/trace/signal.go)) because `TraceParser.TransformEvent` is expected to manage both cases JSON and columns.

Therefore, in the first commit, I started by modifying the `TraceParser.TransformEvent` to `TraceParser.TransformToColumns` and moved the management of JSON to its caller (`TraceGadget`). As described in the commit message, it is an intermediate step needed to introduce the usage of the columns library progressively. Otherwise, we would need to update all
the trace gadgets at once, and it's what we are trying to avoid. Once all the trace gadgets use the columns library, we will move the logic to print JSON or columns to a common place (parser-tableformatter.go that will become parser.go).

After this change, I modified the `signal` trace gadget to use the columns library, which was pretty simple. But then, I did the same for the `exec` trace gadget that is used by [`kubectl-gadget`](https://github.com/kinvolk/inspektor-gadget/blob/cf5d3587f710aadc5e017b1886ebdcfcf4601c2c/cmd/kubectl-gadget/trace/exec.go#L30), [`local-gadget`](https://github.com/kinvolk/inspektor-gadget/blob/cf5d3587f710aadc5e017b1886ebdcfcf4601c2c/cmd/local-gadget/trace/exec.go#L34) and in [a standalone example](https://github.com/kinvolk/inspektor-gadget/blob/cf5d3587f710aadc5e017b1886ebdcfcf4601c2c/examples/gadgets/parser/trace/exec/exec.go#L102). Therefore, we have all use case to test [the new parser](https://github.com/kinvolk/inspektor-gadget/blob/cf5d3587f710aadc5e017b1886ebdcfcf4601c2c/cmd/common/trace/exec.go#L26-L65).

## How to use

Test:

- `bind`, `exec`, `signal` and `tcp` trace gadgets with `kubectl-gadget`.
- `bind`, `exec` and `tcp` trace gadgets with `kubectl-gadget` (`signal` is not yet supported).
- Run examples/gadgets/parser/trace/exec/exec.go

## To be defined

- [x] **Q**: How should we retrieve the gadget-specific columns? I am proposing [`WithNoTags`](https://github.com/kinvolk/inspektor-gadget/blob/cf5d3587f710aadc5e017b1886ebdcfcf4601c2c/cmd/common/utils/parser-tableformatter.go#L65-L68).
  - **A**: Yes, going for `WithNoTags`. It was already included in [#877](https://github.com/kinvolk/inspektor-gadget/pull/877).
- [x] **Q**: How to ensure that the Kubernetes and ContainerRuntime columns are always shown before the gadget-specifics? We are now relying on the fact that the Kubernetes and ContainerRuntime columns will always be defined first in the event struct.
  - **A**: For now, we will rely on the order they are defined.
- [x] **Q**: What to do when custom-columns include invalid columns name? I propose to print a warning message and ignore them.
  - **A**: We agreed to return an error.
- [x] Define width and ellipsis for Kubernetes metadata.
  - Width: We agreed to use the same for all of them: 30. And enable autoscale so that it also works well on small screens.
  - Ellipsis:
    - Node and Pod: `middle`.
    - Namespace and Container: `end` (Default).
- [x] We need to start using JSON format on integration-tests to avoid failing because some columns use the ellipsis ([#719](https://github.com/kinvolk/inspektor-gadget/issues/719)).
  - Resolved by https://github.com/kinvolk/inspektor-gadget/pull/945.

```bash
NODE             NAMESPACE        POD                            CONTAINER                 PID     TPID    SIGNAL           RET    COMM            
fv-az28-19       test-sigsnoop-9… test-pod                       test-pod                  15269   15269   SIGURG           0      runc:[2:INIT]   
fv-az28-19       test-sigsnoop-9… test-pod                       test-pod                  15269   1       SIGURG           0      runc:[2:INIT]   
fv-az28-19       test-sigsnoop-9… test-pod                       test-pod                  15269   15283   SIGTERM          0      sh              
```

Let's discuss this in the comments.


cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, fmt.Errorf("creating columns: %w", err)
+	}
+
+	parser := &GadgetParser[T]{
+		columns: cols,
+
+		// By default, use only the columns with no specific tag. In other
+		// words, the gadget-specific columns. Notice the filter could be
+		// modified using WithCustomConfig.
+		columnsFilter: columns.WithNoTags(),
```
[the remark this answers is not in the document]
flyth: I think it makes sense - if we ever run into a situation where this isn't enough, we could add a "gadget" tag.
For the interface, I think it is more common to call the function HasTags() (instead of HasNoTags()) and then negate that in WithNoTags().

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,98 @@
...
+
+func WithMetadata[T any](tag string) func(*GadgetParser[T]) error {
+	return func(parser *GadgetParser[T]) error {
+		if _, ok := SupportedTags[tag]; !ok {
+			return fmt.Errorf("unknown value for metadata: %q", tag)
+		}
+
+		// The resulting set of columns will be the gadget-specific columns plus
+		// the ones with the requested tag.
+		parser.columnsFilter = columns.Or(columns.WithNoTags(), columns.WithTag(tag))
```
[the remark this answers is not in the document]
flyth: I think that's the best way without having to know any field specifics.

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
blanquicet: > For the interface, I think it is more common to call the function HasTags() (instead of HasNoTags()) and then negate that in WithNoTags().

Mmmm, but in WithNoTags(), we don't have the information about the "available" tags. So, which tags should I use to call HasTag()? Maybe I am missing something.

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
blanquicet: Applied!

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,98 @@
...
+		err = o(parser)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	parser.formatter = textcolumns.NewFormatter(
+		parser.columns.GetColumnMap(parser.columnsFilter),
+		textcolumns.WithDefaultColumns(outputConfig.CustomColumns),
+	)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Using json in the integration tests is like priority number 1 now. I'm wondering if we really need to look for a workaround or maybe we can start working on https://github.com/kinvolk/inspektor-gadget/issues/719 first.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,110 @@
...
+		opts.metadataTag = metadataTag
+		return nil
+	}
+}
+
+type GadgetOptions struct {
+	metadataTag string
+}
+
+// GadgetParser is a parser helps printing the gadget output in columns using
```
mauriciovasquezbernal, line 50:
```suggestion
// GadgetParser is a parser that helps printing the gadget output in columns using
```

?

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,110 @@
...
+	ContainerRuntimeTag: {},
+}
+
+type Option func(*GadgetOptions) error
+
+func WithMetadata(metadataTag string) Option {
+	return func(opts *GadgetOptions) error {
+		if _, ok := SupportedTags[metadataTag]; !ok {
+			return fmt.Errorf("invalid metadata tag: %q", metadataTag)
+		}
```
mauriciovasquezbernal, line 40: Do we really need to check this?

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var defaultColumns []string
+	if len(outputConfig.CustomColumns) != 0 {
+		defaultColumns = outputConfig.CustomColumns
+	} else {
+		// Here the order matters, thus it is important to append the
+		// kubernetes/runtime (tagged) columns before the gadget-specific ones.
+		defaultColumns = append(
+			cols.GetColumnNames(columns.WithTag(opts.metadataTag)),
+			cols.GetColumnNames(columns.WithNoTags())...)
```
mauriciovasquezbernal, line 88:
```suggestion
			cols.GetColumnNames(columns.WithNoTags())...,
		)
```

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Isn't it that bad to rely the way things are defined in the event struct? IMO it's just fine to suppose runtime/kubernetes tags are defined first.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, fmt.Errorf("creating columns: %w", err)
+	}
+
+	// If no tag is provided, we use only the columns with no specific tag. In
+	// other words, the gadget-specific columns. Otherwise, we also include the
+	// columns with the requested tag.
+	filteredColumns := cols.GetColumnMap(columns.WithNoTags())
+	if opts.metadataTag != "" {
+		filteredColumns = cols.GetColumnMap(columns.Or(columns.WithTag(opts.metadataTag), columns.WithNoTags()))
+	}
```
mauriciovasquezbernal, line 78: Perhaps this is a bit more efficient and easier to understand:
```suggestion
	var filter columns.ColumnFilter
	if opts.metadataTag == "" {
		filter = columns.WithNoTags()
	} else {
		filter = columns.Or(columns.WithTag(opts.metadataTag), columns.WithNoTags())
	}

	filteredColumns := cols.GetColumnMap(filter)
```

Also `filteredColumns` could be removed and could be called directly when creating the parser?

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+var SupportedTags = map[string]struct{}{
+	KubernetesTag:       {},
+	ContainerRuntimeTag: {},
+}
+
+type Option func(*GadgetOptions) error
+
+func WithMetadata(metadataTag string) Option {
```
mauriciovasquezbernal, line 36: It seems to me that "Metadata" is too generic here. What about? 

```suggestion
func WithTag(metadataTag string) Option {
```


```suggestion
func WithMetadataTag(metadataTag string) Option {
```

?

examples/gadgets/parser/trace/exec/exec.go
```diff
@@ -107,7 +108,22 @@ func main() {
...
+			}
+
+			fmt.Println(string(b))
+		case utils.OutputModeColumns:
+			fallthrough
+		case utils.OutputModeCustomColumns:
+			fmt.Println(execparser.TransformToColumns(&event))
+		default:
+			fmt.Fprint(os.Stderr, utils.WrapInErrOutputModeNotSupported(config.OutputMode))
+		}
```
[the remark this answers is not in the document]
blanquicet: Sure!

cmd/common/utils/parser-tableformatter.go
blanquicet: When I did it, I was expecting that we would be able to verify the tags used in the event structs, but I just realized it is impossible. I removed it.

cmd/common/trace/exec.go
```diff
@@ -16,104 +16,52 @@ package trace
...
-		"node":      -16,
-		"namespace": -16,
-		"pod":       -30,
-		"container": -16,
-		"pid":       -7,
-		"ppid":      -7,
-		"pcomm":     -16,
-		"ret":       -4,
-		"args":      -24,
+func newExecParser(outputConfig *commonutils.OutputConfig, metadataTag string) (TraceParser[execTypes.Event], error) {
```
mauriciovasquezbernal, line 26: Maybe you can received `opts...` here and use `return newExecParser(outputConfig, WithMetadata(commonutils.KubernetesTag))` in the caller?

mauriciovasquezbernal (commented):
Some comments. I like where it's going, even if this is an intermediate step it's already looking very good!

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
blanquicet: Mmmm ensuring the order was achieved by adding these 3 lines:

https://github.com/kinvolk/inspektor-gadget/blob/841b0d8f511bc879385e1ae13102f409ca2d3fd8/cmd/common/utils/parser-tableformatter.go#L84-L88

However, I just realized that if we configure a specific set of columns through `textcolumns.WithDefaultColumns()`, it will also include the columns with `hide` (e.g. the mntns columns in exec and signal gadgets). Therefore, I will remove these lines for now. If we find a problem with the order, we know how what we need to do.

cmd/common/utils/parser-tableformatter.go
blanquicet: Let's go for `WithMetadataTag`.

cmd/common/trace/exec.go
blanquicet: I considered it, but I thought that it would have been strange to use options from another library here. However, let's say that [`Option`](https://github.com/kinvolk/inspektor-gadget/blob/841b0d8f511bc879385e1ae13102f409ca2d3fd8/cmd/common/utils/parser-tableformatter.go#L34) is the generic option for all the parsers, and it will indeed make the code clear. In addition, I really don't like passing `""` as an argument, so I will do it! 😝

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
flyth: I meant this interface here: https://github.com/kinvolk/inspektor-gadget/pull/934/files#diff-fe87426b97ca95399ca7e32757043489f69a9c9d90011c501a90230b2aad90b5R43 - I would just reverse the functionality here https://github.com/kinvolk/inspektor-gadget/pull/934/files#diff-398759e651e4fd0d1715f49a7ed93ce34347c1a7086cef5f78761035de23bd42R216 and call it `HasTags()` (and use `!HasTags()` in `WithNoTags()`.) But that's really just my opinion, as I would rather expect `Has` than `HasNo` in the API.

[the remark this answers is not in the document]
blanquicet: `WithNoTags` was introduced in https://github.com/kinvolk/inspektor-gadget/pull/877.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,92 @@
...
+	} else {
+		filter = columns.Or(columns.WithTag(opts.metadataTag), columns.WithNoTags())
+	}
+
+	var formatter *textcolumns.TextColumnsFormatter[T]
+	if len(outputConfig.CustomColumns) != 0 {
+		validCols, invalidCols := cols.VerifyColumnNames(outputConfig.CustomColumns)
+		if len(invalidCols) != 0 {
+			fmt.Fprintf(os.Stderr, "Warn: Ignoring invalid columns: %s\n", strings.Join(invalidCols, ", "))
+		}
```
blanquicet, line 71: Do you agree to print a warning message in this case?

mauriciovasquezbernal (commented):
I have some concerns with the `MustGetColumns()` methods defined in the types package. IMO it should be moved somewhere else.

pkg/types/types.go
```diff
@@ -29,18 +29,18 @@ func Init(nodeName string) {
...
+	Namespace string `json:"namespace,omitempty" column:"namespace,width:16" columnTags:"kubernetes"`
 
 	// Pod where the event comes from, or empty for host-level event
-	Pod string `json:"pod,omitempty"`
+	Pod string `json:"pod,omitempty" column:"pod,width:30" columnTags:"kubernetes"`
 
 	// Container where the event comes from, or empty for host-level or
 	// pod-level event
-	Container string `json:"container,omitempty"`
+	Container string `json:"container,omitempty" column:"container,width:25" columnTags:"kubernetes,runtime"`
```
mauriciovasquezbernal, line 43: Do (K8s) containers usually have such a long name?

pkg/types/types.go
```diff
@@ -29,18 +29,18 @@ func Init(nodeName string) {
...
+	Node string `json:"node,omitempty" column:"node,width:16" columnTags:"kubernetes"`
 
 	// Pod namespace where the event comes from, or empty for host-level
 	// event
-	Namespace string `json:"namespace,omitempty"`
+	Namespace string `json:"namespace,omitempty" column:"namespace,width:16" columnTags:"kubernetes"`
 
 	// Pod where the event comes from, or empty for host-level event
-	Pod string `json:"pod,omitempty"`
+	Pod string `json:"pod,omitempty" column:"pod,width:30" columnTags:"kubernetes"`
```
mauriciovasquezbernal, line 39: Is this fine to have a width for the Pod almost twice than we have for namespace and node?

cmd/common/utils/parser-tableformatter.go
```diff
@@ -0,0 +1,92 @@
...
+	ContainerRuntimeTag string = "runtime"
+)
+
+type Option func(*GadgetParserOptions)
+
+func WithMetadataTag(metadataTag string) Option {
+	return func(opts *GadgetParserOptions) {
+		opts.metadataTag = metadataTag
+	}
+}
```
flyth, line 37: What about if we need more than one tag? Maybe change to []string?

flyth (commented):
Just a first pass, I really like where this is going!

pkg/gadgets/trace/signal/types/types.go
```diff
@@ -15,18 +15,25 @@
 package types
 
 import (
+	"github.com/kinvolk/inspektor-gadget/pkg/columns"
```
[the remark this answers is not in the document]
flyth: I talked to @blanquicet yesterday especially about this and I would like to have the columns definition inside the types.go of the corresponding package: Whenever we need special handling of columns (like custom extractors or virtual columns), this way we can also have that definition in the same file. So you can be sure to have the Columns on every gadget and it will give you information about the metadata available. Also, the library will probably be used by not only the frontend, but also by the gadgets themselves (e.g. sorting, filtering, grouping on the server side before cutting the result set to a limited amount like in the top gadgets).

cmd/common/trace/signal.go
```diff
@@ -0,0 +1,75 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+import (
+	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
+	signalTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/signal/types"
+	"github.com/spf13/cobra"
```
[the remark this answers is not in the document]
blanquicet: Sure! I'm also trying to do that on the files I am modifying/creating.

pkg/gadgets/trace/exec/types/types.go
```diff
@@ -15,19 +15,40 @@
...
+	MountNsID uint64   `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+	Retval    int      `json:"ret,omitempty" column:"ret,width:4"`
+	Comm      string   `json:"pcomm,omitempty" column:"comm,width:16"`
+	Args      []string `json:"args,omitempty" column:"args,width:24"`
+}
+
+// execColumns is defined as global variable so that the callers panic if they
+// are importing this package with invalid column tags. However, force callers
+// to use MustGetColumns to include additional configuration to the columns.
+var execColumns = columns.MustCreateColumns[Event]()
```
mauriciovasquezbernal, line 39: Is it a really good idea to put this on an external variable? What about users that don't need it at all, it seems like a waste of memory to me.

pkg/gadgets/trace/signal/types/types.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Ok, let's keep going this way. Later on we can split out the types if need to do so.

cmd/common/trace/signal.go
```diff
@@ -0,0 +1,75 @@
...
+	return newSignalParser(outputConfig, commonutils.WithMetadataTag(commonutils.KubernetesTag))
+}
+
+func NewSignalParserWithRuntimeInfo(outputConfig *commonutils.OutputConfig) TraceParser[signalTypes.Event] {
+	return newSignalParser(outputConfig, commonutils.WithMetadataTag(commonutils.ContainerRuntimeTag))
+}
+
+func NewSignalParser(outputConfig *commonutils.OutputConfig) TraceParser[signalTypes.Event] {
+	return newSignalParser(outputConfig)
+}
```
flyth, line 43: How about using generic functions instead? Avoids having dedicated functions for each gadget by just passing one more argument instead.

```golang
func NewParser[T TraceEvent](outputConfig *commonutils.OutputConfig, columns *columns.Columns[T]) TraceParser[T] {
	return commonutils.NewGadgetParser[T](outputConfig, columns)
}

func NewParserWithK8sInfo[T TraceEvent](outputConfig *commonutils.OutputConfig, columns *columns.Columns[T]) TraceParser[T] {
	return commonutils.NewGadgetParser[T](outputConfig, columns, commonutils.WithMetadataTag(commonutils.KubernetesTag))
}

func NewParserWithRuntimeInfo[T TraceEvent](outputConfig *commonutils.OutputConfig, columns *columns.Columns[T]) TraceParser[T] {
	return commonutils.NewGadgetParser[T](outputConfig, columns, commonutils.WithMetadataTag(commonutils.ContainerRuntimeTag))
}
```

cmd/common/utils/parser-tableformatter.go
blanquicet: Applied.

cmd/common/trace/exec.go
```diff
@@ -16,104 +16,42 @@ package trace
...
 	}
 
-	if len(outputConfig.CustomColumns) == 0 {
-		outputConfig.CustomColumns = GetExecDefaultColumns()
-		if len(prependColumns) != 0 {
-			outputConfig.CustomColumns = append(prependColumns, outputConfig.CustomColumns...)
+	err = execParser.SetColumnExtractor("args", func(event *execTypes.Event) (ret string) {
+		for _, arg := range event.Args {
+			ret += arg + " "
 		}
```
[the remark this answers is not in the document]
blanquicet: I modified it to use `strings.Join` 🙂

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -15,20 +15,27 @@
...
+	Addr      string `json:"addr,omitempty" column:"addr,width:16"`
+	Port      uint16 `json:"port,omitempty" column:"port,width:7"`
+	Options   string `json:"opts,omitempty" column:"opts,width:7"`
+	Interface string `json:"if,omitempty" column:"if,width:7"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+}
+
+var bindColumns = columns.MustCreateColumns[Event]()
+
+func MustGetColumns() *columns.Columns[Event] {
```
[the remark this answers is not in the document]
blanquicet: Yep, let's go for `GetColumns`.

cmd/common/trace/signal.go
blanquicet: Even though I was trying to hide the call to `types.GetColumns`, having these four functions for each gadget creates a lot of duplicated code and it is not worth it.

pkg/gadgets/trace/exec/types/types.go
blanquicet: Yeah, it is not worth it for those who are not interested in using the column format.

cmd/common/utils/parser-tableformatter.go
blanquicet: I would like to keep it as simple as possible knowing that it can be easily extended and I don't expect any use-case where it could be needed.

pkg/types/types.go
blanquicet: I have two examples: `calico-kube-controllers` and `kube-controller-manager`, which seem to me quite common. WDYT?

pkg/types/types.go
blanquicet: As per the discussion about container width, if we use 25 for containers, if they belong to a pod that is part of a daemon set, then they will be for sure longer than 25. Do you think it happens very few times?

blanquicet: BTW, I added the `ellipsis:left`.

flyth: I think pod is a good candidate for `ellipsis:middle` - the start is usually the informative part (what is it) and the end is has got the unique identifier.

Regarding 25/30 characters, what we should keep in mind: when autoscaling is enabled, the columns will be extended by their ratios (if a column takes 1/5 of the size, it will do so after scaling as well).

mauriciovasquezbernal: I think nodes and namespaces can be very long as well. What about using the same width for node, namespace, pod and container?

blanquicet: That's true. AKS nodes: `aks-agentpool-31885719-vmss00000X`. 

Yeah, we could have 30 for all of them, but enable autoscaling by default so that it also works well on small screens. Then, for the ellipsis, we could start using `ellipsis:middle` for the pod and namespace, and `ellipsis:end` for the node and container name. WDYT?

flyth: Hmm, I think `ellipsis:middle` would also make sense for the node (because those often start or end with the unique part). I don't know if it's common for namespaces to have some kind of unique identifier as well - if not, I would suggest using `ellipsis:end`.

blanquicet: Maybe @alban can make some suggestions about this topic

blanquicet:
Rebased after merging changes on trace/tcp gadget on https://github.com/kinvolk/inspektor-gadget/pull/937 that created conflicts.

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
blanquicet: I started working on this https://github.com/kinvolk/inspektor-gadget/pull/945.

pkg/gadgets/trace/signal/types/types.go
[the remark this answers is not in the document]
mauriciovasquezbernal: > Also, the library will probably be used by not only the frontend, but also by the gadgets themselves (e.g. sorting, filtering, grouping on the server side before cutting the result set to a limited amount like in the top gadgets).

I was thinking about this better. I'm not fully convinced to use a library based on reflection in the backend to perform such a tasks. I agree that the library is very flexible to use in the frontend, but I'd like to avoid to use in the backend as much as possible because of performance reasons.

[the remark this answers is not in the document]
flyth: I understand the concern of handling things via reflection and I see it more as a last resort for when the user wants to do specific things we didn't optimize for. That way the user can still do all the fancy things, even if it comes at some cost  - and for those additional features (that only cost when used), we don't have to write additional (struct specific) code at all.

Let's start with some benchmarks to see that reflection isn't all that bad (at least for our use cases):
```
BenchmarkFilterNumberGte-10        	49086614	        23.80 ns/op	       8 B/op	       1 allocs/op
BenchmarkFilterNumberEq-10         	50282662	        23.28 ns/op	       8 B/op	       1 allocs/op
BenchmarkFilterNumberNeq-10        	51283602	        23.31 ns/op	       8 B/op	       1 allocs/op
BenchmarkFilterStringEq-10         	34612514	        33.03 ns/op	      16 B/op	       1 allocs/op
BenchmarkFilterStringNeq-10        	35924289	        33.17 ns/op	      16 B/op	       1 allocs/op
BenchmarkFilterStringRegex-10      	31654264	        37.61 ns/op	       0 B/op	       0 allocs/op
BenchmarkFilterStringRegexCI-10    	17073394	        69.74 ns/op	       0 B/op	       0 allocs/op
BenchmarkGroup1000to100-10         	    6188	    189329 ns/op	  102683 B/op	     712 allocs/op
BenchmarkGroup1000to10-10          	    8006	    148145 ns/op	   63567 B/op	     105 allocs/op
BenchmarkSortReflection1000-10     	    3033	    393848 ns/op	     168 B/op	       4 allocs/op
BenchmarkSortNative1000-10         	    8763	    133296 ns/op	      56 B/op	       2 allocs/op
BenchmarkJSON-10                   	 3154509	       381.1 ns/op	     208 B/op	       1 allocs/op
```

These benchmarks are written for this struct:
```golang
type TestStruct struct {
	Int      int    `column:"int" json:"int"`
	Int2     int    `column:"int2" json:"int2"`
	Int3     int    `column:"int3" json:"int3"`
	Int4     int    `column:"int4" json:"int4"`
	String   string `column:"string" json:"string"`
	String2  string `column:"string2" json:"string2"`
	String3  string `column:"string3" json:"string3"`
	String4  string `column:"string4" json:"string4"`
	GroupID1 string `column:"groupID1" json:"groupID1"`
	GroupID2 string `column:"groupID2" json:"groupID2"`
}
```

(Source: https://github.com/flyth/inspektor-gadget/blob/michael/integrate-benchmark-tableformatter/pkg/columns/examples/benchmark/benchmark_test.go)

Now applying that to our use cases:

1. Heavy throughput gadgets sending events (e.g. trace gadgets)

The only real "feature" that is applicable here, is filtering. We already do filtering on pods, namespaces etc in a heavily optimized way for most gadgets - we should of course not change that. However, for all fields that are considered neither in BPF nor in native code, we could apply filtering through the columns lib (or even for fields like `pod` where we the optimized version support only matching for a specific name but not excluding a name / doing regex etc.)

The cost for a simple filter like `field >= int` in my tests was at about 23ns. If such a filter filters out an entry, we save about 390ns (for the rather simple struct I've tested with) for the JSON encoding and additional time and traffic for decoding / transferring. Given that the user chooses the filter wisely, I think this can help a lot actually decreasing the load on the nodes.

Regular expressions are of course slower, but that's where the overhead of using reflection becomes even smaller.

2. Snapshots (e.g. top gadgets)

For these, we need to do the same operations on the backend that we do on the frontend, as we only transfer a limited number of rows back to the client. So this is where we would need to handle grouping, sorting etc. This would happen every second (or less frequent) for a variable number of entries. The group/sort benchmark above handled 1000 pseudorandomly sorted entries. Those took ~0.4ms for sorting using reflection vs. ~0.13ms when doing it natively. Grouping 1000 entries down to 100 or 10 entries takes 0.19ms and 0.15ms - I have not written a native replacement for that.

Maybe we would need to do more testing with actual data, but I don't think using reflection looks that bad.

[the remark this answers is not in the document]
mauriciovasquezbernal: I got your point. Yes, I agree that if the performance impact is not that big we can use this library on the backend to avoid having to write a lot of code for implementing some features. If we find that the reflection approach is too heavy we can always switch to a "native" implementation.

pkg/types/types.go
blanquicet: To conclude:

- Width: We agreed to use the same for all of them: `30`. And enable autoscale so that it also works well on small screens.
- Ellipsis: 
  - Node and Pod: `middle`.
  - Namespace and Container: `end` (Default).

Do we all agree?

cmd/common/utils/parser-tableformatter.go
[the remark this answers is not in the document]
blanquicet: Rebased now that #945 has been merged.

blanquicet:
I rebased it to include #945 which solves the problems in the integration tests. This PR is now ready for the final review.

cc @mauriciovasquezbernal @flyth

flyth (approved):
Looks great, thanks! LGTM!

One thing, maybe: in the meantime (until all gadgets are using the columns lib), maybe it's better not to use autoscale, since it would be inconsistent between the gadgets.

blanquicet:
> One thing, maybe: in the meantime (until all gadgets are using the columns lib), maybe it's better not to use autoscale, since it would be inconsistent between the gadgets.
>
Yeah, it will add (another) inconsistency, but the problem with not enabling autoscale is that we are setting the k8s columns to `30`, so those columns will end up taking all the space on small screens. In addition, this PR is already introducing inconsistencies because the columns of the gadgets using the library will be truncated, and will use the ellipsis, while the current implementation doesn't. Thus, at this point, what is the difference between having 2 or 3 inconsistencies? I prefer to take advantage of all the potential of the library and speed up the transformation of the rest of the gadgets. WDYT?

flyth:
> > One thing, maybe: in the meantime (until all gadgets are using the columns lib), maybe it's better not to use autoscale, since it would be inconsistent between the gadgets.
> 
> Yeah, it will add (another) inconsistency, but the problem with not enabling autoscale is that we are setting the k8s columns to `30`, so those columns will end up taking all the space on small screens. In addition, this PR is already introducing inconsistencies because the columns of the gadgets using the library will be truncated, and will use the ellipsis, while the current implementation doesn't. Thus, at this point, what is the difference between having 2 or 3 inconsistencies? I prefer to take advantage of all the potential of the library and speed up the transformation of the rest of the gadgets. WDYT?

Yeah, you're right :)

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Options   string `json:"opts,omitempty"`
-	Interface string `json:"if,omitempty"`
-	MountNsID uint64 `json:"mountnsid,omitempty"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16"`
+	Protocol  string `json:"proto,omitempty" column:"proto,width:6"`
+	Addr      string `json:"addr,omitempty" column:"addr,width:16"`
+	Port      uint16 `json:"port,omitempty" column:"port,width:7"`
+	Options   string `json:"opts,omitempty" column:"opts,width:7"`
+	Interface string `json:"if,omitempty" column:"if,width:7"`
```
mauriciovasquezbernal, line 31: This is the name of a network interface. I think 7 is a bit short. What about 12?

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Port      uint16 `json:"port,omitempty"`
-	Options   string `json:"opts,omitempty"`
-	Interface string `json:"if,omitempty"`
-	MountNsID uint64 `json:"mountnsid,omitempty"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16"`
+	Protocol  string `json:"proto,omitempty" column:"proto,width:6"`
+	Addr      string `json:"addr,omitempty" column:"addr,width:16"`
+	Port      uint16 `json:"port,omitempty" column:"port,width:7"`
+	Options   string `json:"opts,omitempty" column:"opts,width:7"`
```
mauriciovasquezbernal, line 30:
```suggestion
	Options   string `json:"opts,omitempty" column:"opts,width:5,fixed"`
```

?

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Comm      string `json:"comm,omitempty"`
-	Protocol  string `json:"proto,omitempty"`
-	Addr      string `json:"addr,omitempty"`
-	Port      uint16 `json:"port,omitempty"`
-	Options   string `json:"opts,omitempty"`
-	Interface string `json:"if,omitempty"`
-	MountNsID uint64 `json:"mountnsid,omitempty"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16"`
+	Protocol  string `json:"proto,omitempty" column:"proto,width:6"`
```
mauriciovasquezbernal, line 27: Possible values to print are `UDP` or `TCP`.  Hence I think we can use fixed there. 

```suggestion
	Protocol  string `json:"proto,omitempty" column:"proto,width:5,fixed"`
```

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Pid       uint32 `json:"pid,omitempty"`
-	Comm      string `json:"comm,omitempty"`
-	Protocol  string `json:"proto,omitempty"`
-	Addr      string `json:"addr,omitempty"`
-	Port      uint16 `json:"port,omitempty"`
-	Options   string `json:"opts,omitempty"`
-	Interface string `json:"if,omitempty"`
-	MountNsID uint64 `json:"mountnsid,omitempty"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16"`
```
mauriciovasquezbernal, line 26: Use also fixed here? (The maximum length in the kernel is 16)

mauriciovasquezbernal (commented):
I have some minor comments about the parameters of the types. Please consider the following cases: 

```
$ tput cols
158
$ ./kubectl-gadget trace exec -A
NODE                       NAMESPACE                  POD                        CONTAINER                  PID   PPID  COMM          RET ARGS                
ubuntu-hirsute             gadget                     gadget-h9l6n               gadget                     9171… 9171… sh            0   /bin/sh -c exec gad…
ubuntu-hirsute             gadget                     gadget-h9l6n               gadget                     9171… 8663… gadgettracer… 0   /usr/bin/gadgettrac…
ubuntu-hirsute             default                    mypod-wbwst                mypod                      9172… 11670 ping          0   /bin/ping -c 1 8.8.…
ubuntu-hirsute             default                    mypod-wbwst                mypod                      9172… 11670 sleep         0   /bin/sleep 1        
^C
Terminating...
```

- PID, PPID and COMM are trimmed. Should we used `fixed` for them? 
- It seems to me that the ARGS column is too narrow. It usually will contain a lot of info, can we give it a 30 (or more) on its `width`?

```$ ./kubectl-gadget trace exec -A -o custom-columns=pod,args
POD                                                                                      ARGS                                                                 
gadget-h9l6n                                                                             /bin/sh -c exec gadgettracermanager -call receive-stream -tracerid t…
gadget-h9l6n                                                                             /usr/bin/gadgettracermanager -call receive-stream -tracerid trace_ga…
mypod-wbwst                                                                              /bin/ping -c 1 8.8.8.8                                               
mypod-wbwst                                                                              /bin/sleep 1                                                         
mypod-wbwst                                                                              /bin/ping -c 1 8.8.8.8                                               
mypod-wbwst                                                                              /bin/sleep 1                                                         
^C
Terminating...
```

I know the autoscale gives the weight to each column based on its width. But in this case it doesn't seem right, the pod is getting a lot of space (89 characters) and probably there is not a pod with such long name, instead args is getting 70 characters and is being trimmed. @flyth do you think it's possible to implement a logic like: if the terminal is very wide give the maximum to some columns and use all the remaining for a specific one. (In this case, give 30 to pod and the remaining space to args)?. 

```
$ tput cols
95
$ ./kubectl-gadget trace bind -A
NODE           NAMESPACE      POD            CONTAINER      PID  COMM    P… ADDR    PO… OP… IF 
ubuntu-hirsute default        mypod          mypod          954… nc      T… 0.0.0.0 42… .R… 0  
^C
Terminating...
```

I suppose we have to use fixed for those short columns like pid, comm, proto, port, opts? Otherwise in small terminals they will be trimmed.

flyth:
>  @flyth do you think it's possible to implement a logic like: if the terminal is very wide give the maximum to some columns and use all the remaining for a specific one. (In this case, give 30 to pod and the remaining space to args)?.

Yeah, good idea to have a maximum and probably a minimum as well that are considered when auto-scaling. Also, when it's a int|uint(8|16|32|64), we can set a fixed size/maximum automatically, maybe with something like `width:auto` or `width:type`. I'll open another issue/PR for that later/tomorrow :)

pkg/gadgets/trace/bind/types/types.go
blanquicet: Aahh, I thought it was the interface index because, [by default, it is `0`](https://github.com/kinvolk/inspektor-gadget/blob/0d7c7f0d0403cee0b2394f671940f606878fc687/pkg/gadgets/trace/bind/tracer/tracer.go#L275). I think we should change that default, it should be an empty string, don't you think?

Beyond that, you are right. I will change it to 12.

blanquicet:
@mauriciovasquezbernal, I applied your suggestions and set almost all the int|uint(8|16|32) to `fixed`. I added some examples of output in the comments. 

BTW, I had to rebase as https://github.com/kinvolk/inspektor-gadget/pull/961 created some conflicts.

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Comm      string `json:"comm,omitempty"`
-	Protocol  string `json:"proto,omitempty"`
-	Addr      string `json:"addr,omitempty"`
-	Port      uint16 `json:"port,omitempty"`
-	Options   string `json:"opts,omitempty"`
-	Interface string `json:"if,omitempty"`
-	MountNsID uint64 `json:"mountnsid,omitempty"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7,fixed"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16,fixed"`
+	Protocol  string `json:"proto,omitempty" column:"proto,width:5,fixed"`
```
blanquicet, line 27: This is how it looks now. The first event was generated with a pod and container named `test-pod`. While for the second, I changed it to `test-pod-long-name-to-test`.

```bash
$ tput cols
95
$ ./kubectl-gadget trace bind -A
NODE       NAMESPACE  POD        CONTAINER PID     COMM             PROTO ADDR  PORT  OPTS  IF
worker     test-ns    test-…test test-pod… 263307  nc               TCP   ::    4242  .R... 0
worker     test-ns    test-pod   test-pod  263789  nc               TCP   ::    4242  .R... 0
```

pkg/gadgets/trace/exec/types/types.go
```diff
@@ -15,19 +15,32 @@
...
-	MountNsID uint64   `json:"mountnsid,omitempty"`
-	Retval    int      `json:"ret,omitempty"`
-	Comm      string   `json:"pcomm,omitempty"`
-	Args      []string `json:"args,omitempty"`
+	Pid       uint32   `json:"pid,omitempty" column:"pid,width:7,fixed"`
+	Ppid      uint32   `json:"ppid,omitempty" column:"ppid,width:7,fixed"`
+	Comm      string   `json:"pcomm,omitempty" column:"comm,width:16,fixed"`
+	Retval    int      `json:"ret,omitempty" column:"ret,width:3,fixed"`
+	Args      []string `json:"args,omitempty" column:"args,width:40"`
+	UID       uint32   `json:"uid,omitempty" column:"uid,width:10,fixed,hide"`
```
blanquicet, line 32:
```bash
$ tput cols
95
$ ./kubectl-gadget trace exec -n test-ns
NODE        NAMESPACE  POD        CONTAINER  PID     PPID    COMM             RET ARGS
worker      test-ns    test-…test test-pod-… 195427  155649  sleep            0   /bin/sleep 0…
worker      test-ns    test-pod   test-pod   195443  100031  sleep            0   /bin/sleep 0…
```

I think COMM is taking too much space after making it fixed. Maybe we could do something like mix=10(?) and max=16?

While for ARGS, we definitely need a minimum.

pkg/gadgets/trace/signal/types/types.go
```diff
@@ -15,18 +15,23 @@
...
-	TargetPid uint32 `json:"tpid,omitempty"`
-	Signal    string `json:"signal,omitempty"`
-	Retval    int    `json:"ret,omitempty"`
-	Comm      string `json:"comm,omitempty"`
-	MountNsID uint64 `json:"mountnsid,omitempty"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7,fixed"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16,fixed"`
+	Signal    string `json:"signal,omitempty" column:"signal,width:6,fixed,ellipsis:start"`
+	TargetPid uint32 `json:"tpid,omitempty" column:"tpid,width:7,fixed"`
+	Retval    int    `json:"ret,omitempty" column:"ret,width:3,fixed"`
```
blanquicet, line 29: For SIGNAL, the most common signals SIGKILL, SIGTERM, SIGINT are 6, so I used that value fixed with `ellipsis:start` as all of them start with `SIG`. However, I think it would be better to have mix=6 and max=11 (SIGRTMIN+13).

```bash
$ tput cols
95
$ ./kubectl-gadget trace signal -A
NODE        NAMESPACE   POD         CONTAINER  PID     COMM             SIGNAL      TPID    RET
master      gadget      gadge…9zn9s gadget     93865   gadgettracerman  SIGURG      106393  0
master      gadget      gadge…9zn9s gadget     93865   gadgettracerman  SIGURG      106393  0
```

pkg/gadgets/trace/tcp/types/types.go
```diff
@@ -15,21 +15,43 @@
...
-	Daddr     string `json:"daddr,omitempty"`
-	Sport     uint16 `json:"sport,omitempty"`
-	Dport     uint16 `json:"dport,omitempty"`
-	Operation string `json:"operation,omitempty"`
+	Operation string `json:"operation,omitempty" column:"t,width:1,fixed"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7,fixed"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:16,fixed"`
+	IPVersion int    `json:"ipversion,omitempty" column:"ip,width:2,fixed"`
+	Saddr     string `json:"saddr,omitempty" column:"saddr,width:22"`
+	Daddr     string `json:"daddr,omitempty" column:"daddr,width:22"`
```
blanquicet, line 30: What do we want to do with the IP addresses? For IPv4 it should have a max=15 to cover this case: `192.168.112.123`. But, what about the min and IPv6?

```bash
$ ./kubectl-gadget trace tcp -n test
-ns
NODE      NAMESPACE POD       CONTAINER T PID     COMM             IP SADDR  DADDR  SPORT DPORT
worker    test-ns   test-pod  test-pod  X 200004  wget             4  192.1… 1.1.1… 51072 80
worker    test-ns   test…test test-pod… C 200008  wget             4  192.1… 1.1.1… 38380 80
```

pkg/gadgets/trace/bind/types/types.go
mauriciovasquezbernal: Yes, this was on my TODO list -> https://github.com/kinvolk/inspektor-gadget/pull/978.

pkg/gadgets/trace/exec/types/types.go
mauriciovasquezbernal: Yes, that can be a good idea. Anyway, I think we can handle those tune ups later on when we convert all gadgets to use this library.

mauriciovasquezbernal (approved):
I think it's ready to go. We can improve the output once the min/max feature in the columns library is implemented.

Issue #944, opened 2022-09-15
[RFE] Decide what to do with the integration test image


eiffel-fl:
## Current situation

To run integration tests, we first build a docker image which contains a compiled binary containing our integration tests.
On release, this image is then pushed as [`inspektor-gadget-test`](https://github.com/kinvolk/inspektor-gadget/pkgs/container/inspektor-gadget-test).

## Impact

To keep our sanity, the integration tests jobs are factorized running in the [`run-integration-tests action`](https://github.com/kinvolk/inspektor-gadget/blob/3af35cb183c576b7a62f3887157e6b7701b5725a/.github/actions/run-integration-tests/action.yml).
As a consequence, this will mean the above image will built for each call to this action but also pushed on release.

## Ideal future situation

The goal here is to discuss what should we do about this particular image.
Should we stop publishing it somewhere? Should we only publish it once when all the integration tests suceeded?

mauriciovasquezbernal:
I'd like to discuss better the goal of publishing that image. IIUC the initial idea was to provide an easy way for Kubernetes administrators to test if Inspektor Gadget works fine on a given cluster by using the `gadget-integration-test-job.yaml` release manifest. I'm not totally convinced that's useful, I think if somebody needs to test Inspektor Gadget they could run the integration tests manually by cloning the repo and running `make -C integration test` (or `make integration-test`).

I checked the [inspektor-gadget-test](https://github.com/kinvolk/inspektor-gadget/pkgs/container/inspektor-gadget-test) package and it has zero pulls for all images. 
![Selection_142](https://user-images.githubusercontent.com/13031921/190509572-0496673a-d5fb-46b4-8195-1df21fdf4f7f.png)

@alban I think you originally introduced this, do you have any opinion?

blanquicet:
> > I checked the inspektor-gadget-test package and it has zero pulls for all images.
> 
> Based on this information, I would tend toward not pushing this image (and delete the corresponding repository) unless there is a really good use case for this image.

Agree

mqasimsarfraz:
> > I checked the inspektor-gadget-test package and it has zero pulls for all images.
> 
> Based on this information, I would tend toward not pushing this image (and delete the corresponding repository) unless there is a really good use case for this image.

+1

alban:
> OK, can you please explain what was the idea you had in mind with this image?

It was to run the IG tests on a specific Kubernetes environment without having to set up a CI. But it ended up not being used.

[...]

PR #945, opened 2022-09-16
integration-tests: Use the JSON mode to verify the gadgets' output

blanquicet:
# Use the JSON output mode in integration-tests

This PR starts using JSON instead of columns output mode in the integration-tests. It starts by implementing just some gadgets:
- Trace/Bind
- Trace/Exec
- Trace/Signal
- Trace/Tcp

Notice I had to upgrade the Go version for the integration-tests jobs because the helper functions created to verify the JSON output use generics, which was introduced in v1.18.

Fixes https://github.com/kinvolk/inspektor-gadget/issues/719.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -2,6 +2,7 @@ name: Inspektor Gadget CI
 env:
   REGISTRY: ghcr.io
   CONTAINER_REPO: ${{ github.repository }}
+  GO_VERSION: 1.18.3
 concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
@@ -29,7 +30,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
       id: go
     - name: Check if generated files are updated
@@ -46,7 +47,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
       id: go
     - name: Install debian packages
@@ -79,7 +80,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
       id: go
     - name: Install debian packages
@@ -136,7 +137,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
       id: go
     - name: Install debian packages
@@ -340,7 +341,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
     - name: Build example binaries
       run: |
@@ -357,7 +358,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
       id: go
     - name: Install debian packages
@@ -381,7 +382,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
       id: go
     - name: Install debian packages
@@ -474,7 +475,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
     - name: Authenticate and set ARO cluster context
       # NOTE: This action generates the Kubernetes config file in the current
@@ -514,7 +515,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.18.3
+        go-version: ${{ env.GO_VERSION }}
         cache: true
     - name: Setup Minikube
       uses: manusa/actions-setup-minikube@v2.4.3
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -471,6 +471,11 @@ jobs:
       group: no-simultaneous-test-integration-aro
     steps:
     - uses: actions/checkout@v3
+    - name: Setup go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.3
+        cache: true
     - name: Authenticate and set ARO cluster context
       # NOTE: This action generates the Kubernetes config file in the current
       # directory. Therefore, it must be run after checking out code otherwise
@@ -506,6 +511,11 @@ jobs:
         type: [default, core]
     steps:
     - uses: actions/checkout@v3
+    - name: Setup go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.3
+        cache: true
     - name: Setup Minikube
       uses: manusa/actions-setup-minikube@v2.4.3
       with:
--- a/integration/command.go
+++ b/integration/command.go
@@ -56,6 +56,9 @@ type command struct {
 	// expectedRegexp contains a regex used to match against the command output.
 	expectedRegexp string
 
+	// expectedOutputFn is a function used to verify the output.
+	expectedOutputFn func(output string) error
+
 	// cleanup indicates this command is used to clean resource and should not be
 	// skipped even if previous commands failed.
 	cleanup bool
@@ -225,6 +228,13 @@ func (c *command) verifyOutput() error {
 	if c.expectedString != "" && output != c.expectedString {
 		return fmt.Errorf("output didn't match the expected string: %s\n%v\n%s",
 			c.expectedString, pretty.Diff(c.expectedString, output), getInspektorGadgetLogs())
+	}
+
+	if c.expectedOutputFn != nil {
+		if err := c.expectedOutputFn(output); err != nil {
+			return fmt.Errorf("verifying output with custom function: %w\n%s",
+				err, getInspektorGadgetLogs())
+		}
 	}
 
 	return nil
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -1,0 +1,107 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/google/go-cmp/cmp"
+
+	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
+)
+
+func parseOutput[T any](output string, normalize func(*T)) ([]*T, error) {
+	ret := []*T{}
+
+	scanner := bufio.NewScanner(strings.NewReader(output))
+	for scanner.Scan() {
+		var entry T
+
+		line := scanner.Text()
+		if err := json.Unmarshal([]byte(line), &entry); err != nil {
+			return nil, fmt.Errorf("unmarshaling line %q: %w", line, err)
+		}
+
+		// To be able to use reflect.DeepEqual and cmp.Diff, we need to
+		// "normalize" the output so that it only includes non-default values
+		// for the fields we are able to verify.
+		if normalize != nil {
+			normalize(&entry)
+		}
+
+		ret = append(ret, &entry)
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("parsing output: %w", err)
+	}
+
+	return ret, nil
+}
+
+// expectAllToMatch verifies that the expectedEntry is matched by all the
+// entries in the output.
+func expectAllToMatch[T any](output string, normalize func(*T), expectedEntry *T) error {
+	entries, err := parseOutput(output, normalize)
+	if err != nil {
+		return err
+	}
+
+	for _, entry := range entries {
+		if !reflect.DeepEqual(expectedEntry, entry) {
+			return fmt.Errorf("unexpected output entry:\n%s",
+				cmp.Diff(expectedEntry, entry))
+		}
+	}
+
+	return nil
+}
+
+// expectEntriesToMatch verifies that all the entries in expectedEntries are
+// matched by at least one entry in the output.
+func expectEntriesToMatch[T any](output string, normalize func(*T), expectedEntries ...*T) error {
+	entries, err := parseOutput(output, normalize)
+	if err != nil {
+		return err
+	}
+
+out:
+	for _, expectedEntry := range expectedEntries {
+		for _, entry := range entries {
+			if reflect.DeepEqual(expectedEntry, entry) {
+				continue out
+			}
+		}
+		return fmt.Errorf("output doesn't contain the expected entry: %+v", expectedEntry)
+	}
+
+	return nil
+}
+
+func buildBaseEvent(namespace string) eventtypes.Event {
+	return eventtypes.Event{
+		Type: eventtypes.NORMAL,
+		CommonData: eventtypes.CommonData{
+			Namespace: namespace,
+			// Pod and Container name are defined by busyboxPodCommand.
+			Pod:       "test-pod",
+			Container: "test-pod",
+			// TODO: Include the Node
+		},
+	}
+}
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -25,6 +25,11 @@ import (
 	"syscall"
 	"testing"
 	"time"
+
+	bindTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/bind/types"
+	execTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/exec/types"
+	signalTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/signal/types"
+	tcpTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/tcp/types"
 )
 
 const (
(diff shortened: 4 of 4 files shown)


integration/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.18
 require github.com/kr/pretty v0.3.0
 
 require (
+	github.com/kinvolk/inspektor-gadget v0.8.0 // indirect
```
blanquicet, line 8: TODO: How to use the events of the commit we are testing?

integration/integration_test.go
```diff
@@ -109,6 +115,45 @@ func cleanupFunc(cleanupCommands []*command) {
...
+	for scanner.Scan() {
+		var obj T
+
+		line := scanner.Text()
+		if err := json.Unmarshal([]byte(line), &obj); err != nil {
+			return fmt.Errorf("unmarshaling line %q: %w", line, err)
+		}
+
+		// Can't use reflect.DeepEqual as we don't have all the expected fields.
+		// For now, each test verifies manually their subset of expected fields.
```
blanquicet, line 130: This is the main problem with this approach. My first idea is to use this custom match where each test verifies its known fields. WDYT?

integration/integration_test.go
```diff
@@ -109,6 +115,45 @@ func cleanupFunc(cleanupCommands []*command) {
 	}
 }
 
+func expectAtLeastOneMatch[T any](output string, expected T, match func(T) bool) error {
```
blanquicet, line 118: I want to avoid making the `command` struct to use generics as it doesn't make sense for all commands. Therefore, I tried to hide the generics by calling `customOutputVerifier` (No generics) -> `expectAtLeastOneMatch[T any]`. WDYT? I am still thinking in another way because it seems quite complex.

integration/integration_test.go
```diff
@@ -469,10 +544,32 @@ func TestExecsnoop(t *testing.T) {
...
-		startAndStop:   true,
+		name:         "StartExecsnoopGadget",
+		cmd:          fmt.Sprintf("$KUBECTL_GADGET trace exec -n %s -o json", ns),
+		startAndStop: true,
+		customOutputVerifier: func(output string) error {
+			expected := execTypes.Event{
+				Event: buildBaseEvent(ns),
+				Comm:  "date",
+				Args:  []string{"/bin/date"},
+			}
```
blanquicet, line 555: Another cool thing about this change + this approach is that we could verify multiple entries. For instance, for the trace `exec` gadget:
 
1.` ..."pid":23861,"ppid":23841,"mountnsid":4026532375,"pcomm":"sh","args":["/bin/sh","-c","while true; do date \u0026\u0026 sleep 0.1; done"]}`
2. `... "pid":23877,"ppid":23861,"mountnsid":4026532375,"pcomm":"date","args":["/bin/date"]}`
3. `... "pid":23878,"ppid":23861,"mountnsid":4026532375,"pcomm":"sleep","args":["/bin/sleep","0.1"]}`

Currently, I just added the one for `date`, but I will add the other two.

integration/integration_test.go
```diff
@@ -109,6 +115,45 @@ func cleanupFunc(cleanupCommands []*command) {
...
+	}
+
+	return fmt.Errorf("output doesn't contain the expected object: %+v", expected)
+}
+
+func buildBaseEvent(namespace string) eventtypes.Event {
+	return eventtypes.Event{
+		Type: eventtypes.NORMAL,
+		CommonData: eventtypes.CommonData{
+			// TODO: Include node
```
blanquicet, line 149: It would be great to be able to verify this info. But maybe on a future PR.

integration/go.mod
mqasimsarfraz: I know we can now use [go workspaces](https://go.dev/doc/tutorial/workspaces) for situations like these but if we decide to get rid of integration test image via #944 we don't really need this module. Right?

blanquicet: Yep. that's also my understanding. If we are not missing something, that's another good reason to get rid of that image.

integration/integration_test.go
mauriciovasquezbernal: What about removing the fields we can't compare from the event and using reflect.DeepEqual -> https://github.com/kinvolk/inspektor-gadget/commit/3dbf544a05f05e972f5a284d5c1a1fcb976decc9. The resulting code is shorter and we' could provide detailed information of which is the field not matching by using `cmp.Diff()`.

integration/integration_test.go
```diff
@@ -342,10 +387,40 @@ func TestBindsnoop(t *testing.T) {
...
+				Event:     buildBaseEvent(ns),
+				Comm:      "nc",
+				Protocol:  "TCP",
+				Addr:      "::",
+				Port:      9090,
+				Options:   ".R...",
+				Interface: "0",
+			}
+
+			return expectAtLeastOneMatch(output, expected, func(e bindTypes.Event) bool {
```
mauriciovasquezbernal, line 404: Is `expectAtLeastOneMatch` really what we want for this test? It seems to me that all generated events should match. (We're running the command in a pod and filtering by namespace where there is a single pod, hence I wouldn't expect to capture any event that we didn't generate).

integration/integration_test.go
mauriciovasquezbernal: IMO it's a good solution,  `customOutputVerifier` is generic enough because it receives the output of the gadget, then each test can use helpers like `expectAtLeastOneMatch` to perform the validation.

integration/command.go
```diff
@@ -56,6 +56,9 @@ type command struct {
 	// expectedRegexp contains a regex used to match against the command output.
 	expectedRegexp string
 
+	// customOutputVerifier is a function that defines a custom verifier.
+	customOutputVerifier func(output string) error
```
mauriciovasquezbernal, line 60: I'm not really good a naming things, but what about the following options?

`expectedOutput`
`expectedOutputFn`
`expectedOutputVerifier`
`expectedOutputChecker`

?

integration/integration_test.go
blanquicet: Great suggestion, looks better!

integration/integration_test.go
blanquicet: Yeah, it is not but it was just the quickest way to implement a solution and start this discussion. The final solution would be to have multiple flavours. For instance:

- For execsnoop, we are triggering three different events and we would like to verify that every generated event is one of the expected ones. So, we would need something like `expectOneFrom`, which would receive a slice of expected events and every event must match one of them:

```bash
{"node":"worker","namespace":"test-execsnoop-2760712051242727130","pod":"test-pod","container":"test-pod","type":"normal","pid":245543,"ppid":245522,"mountnsid":4026532375,"pcomm":"sh","args":["/bin/sh","-c","while true; do date \u0026\u0026 sleep 0.1; done"]}
{"node":"worker","namespace":"test-execsnoop-2760712051242727130","pod":"test-pod","container":"test-pod","type":"normal","pid":245559,"ppid":245543,"mountnsid":4026532375,"pcomm":"date","args":["/bin/date"]}
{"node":"worker","namespace":"test-execsnoop-2760712051242727130","pod":"test-pod","container":"test-pod","type":"normal","pid":245560,"ppid":245543,"mountnsid":4026532375,"pcomm":"sleep","args":["/bin/sleep","0.1"]}
[... data and sleep event are repeated ...]
```
- For bindsnoop, they are all the same. So there we will need `expectOnly`, which would receive one expected event and every generated must match it:
```bash
{"node":"worker","namespace":"test-bindsnoop-3509468263802846636","pod":"test-pod","container":"test-pod","type":"normal","pid":239853,"comm":"nc","proto":"TCP","addr":"::","port":9090,"opts":".R...","if":"0","mountnsid":4026532375}
{"node":"worker","namespace":"test-bindsnoop-3509468263802846636","pod":"test-pod","container":"test-pod","type":"normal","pid":239881,"comm":"nc","proto":"TCP","addr":"::","port":9090,"opts":".R...","if":"0","mountnsid":4026532375}
[... All the following are equal ...]
```

WDYT?

integration/command.go
blanquicet: Yeah, they follow the slipstream from `expectedRegexp` and `expectedString`. I will go with `expectedOutputFn`.

integration/integration_test.go
mauriciovasquezbernal: Makes sense.

integration/Makefile
```diff
@@ -4,15 +4,9 @@ IMAGE_TAG ?= $(shell ../tools/image-tag branch)
...
-KUBERNETES_DISTRIBUTION ?= ""
-
 .PHONY: build
 build:
-	mkdir -p bin
-	cp ../kubectl-gadget-linux-amd64 bin/kubectl-gadget
-	docker build -t $(TEST_CONTAINER_REPO):$(IMAGE_TAG) -f Dockerfile \
-		--build-arg KUBERNETES_DISTRIBUTION=$(KUBERNETES_DISTRIBUTION) .
-	rm -f bin/kubectl-gadget
+	make -C ../ build-integration-tests
```
blanquicet, line 9: After setting up a [go workspace](https://go.dev/doc/tutorial/workspaces), including both modules IG and integration-tests, it is now possible to use the local code of IG on the integration-tests module. However, given that we are building the integration-tests in a container, we need to include the IG code into that building container (we are currently copying only the code of the integration-tests). Notice it is not possible to modify the `integration/Dockerfile` to do something like `COPY ../ .` because `../` (IG code) is outside the build context. To solve this, I am now building the integration-tests from the IG folder so that the IG code is part of the build context.

I don't like having the building on the main Makefile and the rest here. Maybe I should move all the tasks to the main one. WDYT?
 
Please consider that all this will be simplified/removed once https://github.com/kinvolk/inspektor-gadget/issues/944 is solved. I am just trying to unblock this PR because it is blocking #934, and solving 944 could take some time.

cc @mqasimsarfraz

integration/integration_test.go
```diff
@@ -810,10 +858,24 @@ func TestSigsnoop(t *testing.T) {
...
+		cmd:          fmt.Sprintf("$KUBECTL_GADGET trace signal -n %s -o json", ns),
+		startAndStop: true,
+		expectedOutputFn: func(output string) error {
+			expectedEntry := &signalTypes.Event{
+				Event:  buildBaseEvent(ns),
+				Comm:   "sh",
+				Signal: "SIGTERM",
+			}
+
+			return expectOnly(output, expectedEntry, func(e *signalTypes.Event) {
```
blanquicet, line 871: The CI is sometimes failing because, in some cases, the gadget captures the event:

```json
{"node":"fv-az39-947","namespace":"test-sigsnoop-2744583553906244966","pod":"test-pod","container":"test-pod","type":"normal","pid":16075,"tpid":16075,"signal":"SIGURG","comm":"runc:[2:INIT]","mountnsid":4026532724}
```

From my understanding, this signal happens at the container creation, and sometimes the gadget takes more time to start capturing events than others, so it sometimes loses this event. I added a delay of 5 seconds after launching the gadget, and it still failed 7/30 tests I ran locally.

I am wondering what is the best approach here. Should I implement a helper `expectAtLeastOne` for this test? WDYT?

integration/go.mod
blanquicet: I implemented a solution using go workspaces, but it could still be improved once #944 is solved. See https://github.com/kinvolk/inspektor-gadget/pull/945#discussion_r974174068.

integration/Makefile
mauriciovasquezbernal: I think #944 will be solved shortly. So we can avoid all these things.

integration/helpers.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("parsing output: %w", err)
+	}
+
+	return nil
+}
+
+func expectOnly[T any](output string, expected *T, normalize func(*T)) error {
```
mauriciovasquezbernal, line 62: Does something more specific like `expectOneEvent` work better?

integration/helpers.go
```diff
@@ -0,0 +1,112 @@
...
+	return iterOutput(output, normalize, func(entry *T) error {
+		if !reflect.DeepEqual(expected, entry) {
+			return fmt.Errorf("unexpected output entry:\n%s\n", cmp.Diff(expected, entry))
+		}
+		return nil
+	})
+}
+
+func expectAllFrom[T any](output string, expectedEntries []T, normalize func(*T)) error {
```
mauriciovasquezbernal, line 71: This name is not clear to me. Does it mean all expectedEntries should be present on the output?

integration/integration_test.go
mauriciovasquezbernal: I think having a `expectAtLeastOne` helps in the case events are generated outside our control.

integration/helpers.go
```diff
@@ -0,0 +1,112 @@
...
+		if err := compare(&entry); err != nil {
+			return err
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("parsing output: %w", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Something like https://github.com/kinvolk/inspektor-gadget/commit/d9b44b0985a8d76f9db5f31cb33d71881ab88e7b?

mauriciovasquezbernal (commented):
Looks very good overall. Some comments about the iterOutput function.

integration/helpers.go
mauriciovasquezbernal: OK, now I realized it means all the events should match. Maybe `expectAllToMatch` is more clear?

integration/Makefile
blanquicet: Yeah, I expect to remove that commit.

integration/helpers.go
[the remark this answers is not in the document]
blanquicet: I didn't used this approach because I wanted to avoid iterating twice over the output. Notice also that both `expectOnly` and `expectAllFrom` will always iterate over all the events. Only `expectAtLeastOne` will return earlier. However, we are not generating to many events, so maybe your suggestion is more suitable in this case. I will change it.

integration/helpers.go
blanquicet: Yes, I changed it to:

```golang
// expectListIsMatched verifies that all the entries in expectedEntries is
// matched by at least one entry in the output.
func expectListIsMatched[T any](output string, expectedEntries []*T, normalize func(*T)) error
```

WDYT?

integration/helpers.go
blanquicet: Applied:

```golang
// expectAllToMatch verifies that all the entries in the output matches the
// expected entry.
func expectAllToMatch[T any](output string, expectedEntry *T, normalize func(*T)) error
```

integration/helpers.go
mauriciovasquezbernal: What about `expectListMatch`?

integration/helpers.go
```diff
@@ -0,0 +1,124 @@
...
+			return nil
+		}
+	}
+
+	return fmt.Errorf("output doesn't contain the expected entry: %+v", expectedEntry)
+}
+
+// expectListIsMatched verifies that all the entries in expectedEntries is
+// matched by at least one entry in the output.
+func expectListIsMatched[T any](output string, expectedEntries []*T, normalize func(*T)) error {
```
mauriciovasquezbernal, line 94: Can we combine `expectAtLeastOneToMatch` and `expectListIsMatched` into a single function: `expectEventsToMatch(output string, normalize func(*T), expectedEntries ...*T)`?

mauriciovasquezbernal (commented):
One final comments about merging two functions in a single one.

integration/helpers.go
mauriciovasquezbernal: Forget that, I made another suggestion in my last review.

integration/helpers.go
blanquicet: Agree! I think we are converging on a solution:

```golang
// expectAllToMatch verifies that the expectedEntry is matched by all the
// entries in the output.
func expectAllToMatch[T any](output string, normalize func(*T), expectedEntry *T) error
```

```golang
// expectAtLeastOneToMatch verifies that all the elements in expectedEntries are
// matched by at least one entry in the output.
func expectAtLeastOneToMatch[T any](output string, normalize func(*T), expectedEntries ...*T)
```

Regarding `expectAllToMatch` (or `expectAllEntriesToMatch`), if we expect all the entries to match, it is obvious that it receives only one `expectedEntry`, so it is fine. Instead, in the case of `expectAtLeastOneToMatch` (or `expectAtLeastOneEntryToMatch`), we are more flexible, and it could receives multiple elements that must be matched by at least one entry in the output. WDYT?

mauriciovasquezbernal: I think `One` in `expectAtLeastOneEntryToMatch` is a bit confusing when multiple entries are passed. I'd go for a more generic name like `expectEntriesToMatch`. (Also, if we really want we can provide a `expectAtLeastOneToMatch` wrapper that only accepts a single entry).

blanquicet: > I think `One` in `expectAtLeastOneEntryToMatch` is a bit confusing when multiple entries are passed. 
Good point.

blanquicet:
I would say that this PR is ready, it is just waiting for https://github.com/kinvolk/inspektor-gadget/pull/955 to be merged so that we can remove commit https://github.com/kinvolk/inspektor-gadget/pull/945/commits/071c6bafc2a352c1ebeabf80c09326ce5fd5f478.

integration/Makefile
blanquicet: I removed it thanks to https://github.com/kinvolk/inspektor-gadget/pull/955.

blanquicet:
Rebased on top of https://github.com/kinvolk/inspektor-gadget/pull/955. I removed all the go workspaces stuff as it is not needed anymore.

blanquicet:
I had to upgrade the Go version for the integration-tests jobs because the helper functions created to verify the JSON output use generics, which was introduced in v1.18.

.github/workflows/inspektor-gadget.yml
```diff
@@ -471,6 +471,11 @@ jobs:
       group: no-simultaneous-test-integration-aro
     steps:
     - uses: actions/checkout@v3
+    - name: Setup go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.3
```
mqasimsarfraz, line 477: Thanks for this. I should have done this in my change. At some point we should `env` for this so we are sure we use same version everywhere.

integration/helpers.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2022 The Inspektor Gadget authors
```
mqasimsarfraz, line 1: I wonder if naming the files with `_test.go` prefix is a better idea like `command_test.go/helpers_test.go`. WDYT?

mqasimsarfraz (approved):
LGTM, this so useful for the work I wanted to do for #946

.github/workflows/inspektor-gadget.yml
blanquicet: Great suggestion. Don!

integration/helpers.go
blanquicet: Mmmm, I noticed you did that on https://github.com/kinvolk/inspektor-gadget/pull/946. I am not sure, using `_test.go` makes sense because they should be only considered by `go test`. However, they don't include tests at all, so it doesn't look natural to use the `_test.go` convention.

mqasimsarfraz: Make sense. I will follow the same convention in #946 then!

integration/integration_test.go
blanquicet: https://github.com/kinvolk/inspektor-gadget/issues/967

PR #946, opened 2022-09-16
integration-tests: add integration test for local-gadget

### LINKED ISSUE #738: [RFE] local-gadget: Create integration tests
## Current situation

There are no integration-tests for the `local-gadget`. Once we agree on how to implement the single call run https://github.com/kinvolk/inspektor-gadget/pull/730, we should start implementing the tests for them.  

As said in https://github.com/kinvolk/inspektor-gadget/issues/719, these tests should test the JSON format and the transformation from JSON to columns will be done by unit-testing the GadgetParsers' method, e.g. for [the socket gadget](https://github.com/kinvolk/inspektor-gadget/blob/65a398e8ef72887c9b36550de64e7b00cfa8451f/cmd/kubectl-gadget/snapshot/socket.go#L33-L38).

mqasimsarfraz:
Add support for running integration test for local-gadget against multiple container runtimes.

Fixes #738 

### Idea
The idea is based on following changes:
- Make `command.go` API public to share between different test environments.
- Use minikube to spin-up different container runtime environments.
- Hook the test in CI to run on each commit.
- Add couple of test to ensure the setup works properly. 

### Testing Done
Did a manual run of integration test against an AKS cluster.


integration/local-gadget/helpers_test.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, fmt.Errorf("running command=%v: %w", commands, err)
+	}
+	if stderr.Len() > 0 {
+		return nil, fmt.Errorf("error running command: %v", commands)
+	}
+
+	return stdout.Bytes(), nil
+}
+
+func runCRIContainer(ctx context.Context, runtime, prefix, image string, command []string) error {
```
mqasimsarfraz, line 36: I haven't used it anywhere yet but since docker also supports CRI so having this generic to way to create/remove test container should work fine for us.

integration/local-gadget/helpers_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"os/exec"
+	"time"
+
+	"google.golang.org/grpc"
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+)
+
+var ErrNoCommandProvided = errors.New("no command provided")
+
+func run(commands ...string) ([]byte, error) {
```
mqasimsarfraz, line 18: It would be nice to use [command](https://github.com/kinvolk/inspektor-gadget/blob/main/integration/command.go#L36) here. But I wanted to avoid adding multi module dependency like [this](https://github.com/kinvolk/inspektor-gadget/pull/945#discussion_r972908078). Also #945 will be super helpful for JSON validation.

mqasimsarfraz (commented):
@blanquicet I looked into the initial idea regrading running test against multiple runtimes. Would be great to get feedback if you think the change is heading in right direction?

blanquicet (commented):
It is just great @mqasimsarfraz! I really like how it is already looking 💯 

The setup is very complete, clear and easy to understand. The part I like more is that it creates a solid base for the full feature. I think it is great you already thought about creating a "framework" generic enough that can support all the runtimes. Awesome! 🎉 

I think we definitely need to tackle https://github.com/kinvolk/inspektor-gadget/issues/944 first, and then we can create a set of helper functions to verify the JSON output, see my comment below.

integration/local-gadget/Makefile
```diff
@@ -0,0 +1,70 @@
...
+	export MINIKUBE_PROFILE=minikube-$* && \
+	echo "Checking minikube with profile $${MINIKUBE_PROFILE} is running ..." && \
+	$(CONTAINER_CLI) inspect $${MINIKUBE_PROFILE} >/dev/null 2>&1 || (echo "Error: $${MINIKUBE_PROFILE} not running, exiting ..." && exit 1) && \
+	echo "Preparing minikube with profile $${MINIKUBE_PROFILE} for testing ..." && \
+	$(CONTAINER_CLI) cp ../../local-gadget-linux-amd64 $${MINIKUBE_PROFILE}:/bin/local-gadget >/dev/null 2>&1 && \
+	go test -c -o local-gadget-integration.test ./... && \
+	$(CONTAINER_CLI) cp local-gadget-integration.test $${MINIKUBE_PROFILE}:/bin/ >/dev/null 2>&1 && \
+	rm local-gadget-integration.test && \
+	echo "Running test in minikube with profile $${MINIKUBE_PROFILE} ..." && \
+	$(CONTAINER_CLI) exec $${MINIKUBE_PROFILE} local-gadget-integration.test -integration -container-runtime $*
```
blanquicet, line 70: I propose adding `-test.v` in this initial phase, we can remove it once the tests are more stable:

```suggestion
	$(CONTAINER_CLI) exec $${MINIKUBE_PROFILE} local-gadget-integration.test -integration -test.v -container-runtime $*
```

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestListContainers(t *testing.T) {
+	ctx := context.Background()
+	prefix := "list-containers"
+	podID, err := runTestContainer(ctx, *containerRuntime, prefix, "docker.io/library/alpine", "sleep inf")
+	if err != nil {
+		t.Fatalf("failed to run container: %v", err)
+	}
+	defer removeTestContainer(ctx, *containerRuntime, podID)
+
+	out, err := run("local-gadget", "list-containers", "-o", "json", "--runtimes", *containerRuntime)
```
blanquicet, line 74: I propose adding another test here using the `--containername` flag and then using a helper function that verifies there is only one result, and it matches the expected one. While the existing test (without the filtering) expects multiple results but only one must match the expected one.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		t.Fatalf("failed to run commamd: %v", err)
+	}
+
+	var containers []containercollection.Container
+	if err = json.Unmarshal(out, &containers); err != nil {
+		t.Fatalf("failed to unmarshal: %v", err)
```
blanquicet, line 81: Could you please start using the new notation we want to adopt: https://github.com/kinvolk/inspektor-gadget/issues/881#issuecomment-1231424009?

integration/local-gadget/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+.PHONY: phony_explicit
+test-%:
+	export MINIKUBE_PROFILE=minikube-$* && \
+	echo "Checking minikube with profile $${MINIKUBE_PROFILE} is running ..." && \
+	$(CONTAINER_CLI) inspect $${MINIKUBE_PROFILE} >/dev/null 2>&1 || (echo "Error: $${MINIKUBE_PROFILE} not running, exiting ..." && exit 1) && \
+	echo "Preparing minikube with profile $${MINIKUBE_PROFILE} for testing ..." && \
+	$(CONTAINER_CLI) cp ../../local-gadget-linux-amd64 $${MINIKUBE_PROFILE}:/bin/local-gadget >/dev/null 2>&1 && \
+	go test -c -o local-gadget-integration.test ./... && \
+	$(CONTAINER_CLI) cp local-gadget-integration.test $${MINIKUBE_PROFILE}:/bin/ >/dev/null 2>&1 && \
```
blanquicet, line 67: I think stderr shouldn't be sent to redirected, otherwise, it will just fail, and we will not know why.

integration/local-gadget/helpers_test.go
```diff
@@ -0,0 +1,144 @@
...
+		socketPath(runtime),
+		grpc.WithInsecure(),
+		grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+			d := net.Dialer{Timeout: 2 * time.Second}
+			return d.DialContext(ctx, "unix", socketPath(runtime))
+		}),
+	)
+}
+
+func runTestContainer(ctx context.Context, runtime, prefix, image string, command string) (string, error) {
```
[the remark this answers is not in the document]
mqasimsarfraz: Yeah, makes much more sense and also the workflow for both the integration tests will be similar. I will switch to using `busyboxPodCommand` after #944 is merged, will make everything simpler.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,93 @@
...
+	ContainerRuntimeDocker     = "docker"
+	ContainerRuntimeContainerd = "containerd"
+	ContainerRuntimeCRIO       = "cri-o"
+)
+
+var supportedContainerRuntimes = []string{ContainerRuntimeDocker, ContainerRuntimeContainerd, ContainerRuntimeCRIO}
+
+var (
+	integration      = flag.Bool("integration", false, "run integration tests")
+	containerRuntime = flag.String("container-runtime", "", "allows to skip tests that are not supported on a given container-runtime")
```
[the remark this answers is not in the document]
mqasimsarfraz: hehe. Side effect of ctrl-c + ctrl-v :smile: But my initial idea was to do some validation like the docker version since we need specific one for cri-o

mqasimsarfraz (commented):
@blanquicet thanks for an awesome review. I will rework the PR to address the suggestions!

.github/workflows/inspektor-gadget.yml
```diff
@@ -454,6 +454,42 @@ jobs:
...
+          tar zxvf /home/runner/work/inspektor-gadget/local-gadget-linux-amd64.tar.gz
+          mv local-gadget local-gadget-linux-amd64
+      # https://github.com/kubernetes/minikube/issues/12928#issuecomment-1245458044
+      - name: Upgrade docker for ${{ matrix.runtime }}
+        if: matrix.runtime == 'cri-o'
+        run: |
+          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+          sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu  $(lsb_release -cs)  test"
+          sudo apt-get update
+          sudo apt-get install docker-ce
```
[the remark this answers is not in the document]
blanquicet: Yeah, I like your solution. It is pretty simple and doesn't have any external depency. I was just wondering how were they managing this issue.

integration/local-gadget/Makefile
```diff
@@ -0,0 +1,70 @@
+CONTAINER_CLI ?= docker
```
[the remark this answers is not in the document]
blanquicet: Yeah, probably removed.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,93 @@
...
+	if err = json.Unmarshal(out, &containers); err != nil {
+		t.Fatalf("failed to unmarshal: %v", err)
+	}
+
+	var found bool
+	for _, container := range containers {
+		if strings.Contains(container.Name, prefix) {
+			found = true
+		}
+	}
```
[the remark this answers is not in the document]
blanquicet: For instance, once https://github.com/kinvolk/inspektor-gadget/pull/973 is merged, we can verify here that containers created with `kubectl-gadget` were enriched with the correct values: namespace, pod and poduid. Such verification should be omitted if containers were created via `docker`, `ctr` or ... I don't know how to create standalone CRI-O containers 😅 .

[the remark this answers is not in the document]
mqasimsarfraz: Nice idea. As of now I don't think it would be possible to create non CRI complaint container uing CRI-O. I used runtime service or `crictl` to create such containers and that obviously have the concept of pod/sandbox. It would be so weird if it does support something like that! :smile:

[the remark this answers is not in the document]
blanquicet: Yeah, I also think it is not possible. Maybe it is not even possible by definition 😅.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+			var containers []*containercollection.Container
+			if err := json.Unmarshal([]byte(output), &containers); err != nil {
+				return err
+			}
+
+			for _, gotContainer := range containers {
+				normalize(gotContainer)
+				if !reflect.DeepEqual(gotContainer, expectedContainer) {
+					return fmt.Errorf("expect all containers to match, expected=%+v, got=%+v", gotContainer, expectedContainer)
```
blanquicet, line 203:
```suggestion
					return fmt.Errorf("expect all containers to match, expected=%+v, got=%+v", expectedContainer, gotContainer)
```

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,262 @@
...
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+
+	. "github.com/kinvolk/inspektor-gadget/integration"
```
blanquicet, line 14: TIL: https://go.dev/ref/spec#Import_declarations

blanquicet (commented):
I like the result. It is simple and reuses most of the code already created for the inspektor-gadget tests. As I already said, I think it defines a solid base to start adding more complex tests.

Thanks for also updating the documentation. It is clear and really helps.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,263 @@
...
+				Options:  ".R...",
+			}
+
+			normalize := func(e *bindTypes.Event) {
+				// TODO: Handle it once we support getting K8s container name for docker
+				e.Container = "test-pod"
+				e.Node = ""
+				e.Pid = 0
+				e.MountNsID = 0
+				e.Interface = ""
```
blanquicet, line 247: Why can `Interface` not be verified?

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,263 @@
...
+				Event:    BuildBaseEvent(ns),
+				Comm:     "nc",
+				Protocol: "TCP",
+				Addr:     "::",
+				Port:     9090,
+				Options:  ".R...",
+			}
+
+			normalize := func(e *bindTypes.Event) {
+				// TODO: Handle it once we support getting K8s container name for docker
```
blanquicet, line 242: Once we have the Kubernetes metadata available, I think we should start considering moving the shared code between inspektor-gadget and local-gadget to a common place. For instance, this test is identical except for the `Cmd` that calls `kubectl-gadget` or `local-gadget`.

Also because I expect to add more tests like this for each gadget that will be identical again. For example, to verify that the flag `--pid`, `--ports` and `--ignore-errors` are working properly.

cc @mauriciovasquezbernal

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,263 @@
...
+				e.Container = "test-pod"
+				e.Node = ""
+				e.Pid = 0
+				e.MountNsID = 0
+				e.Interface = ""
+			}
+
+			return ExpectAllToMatch(output, normalize, expectedEntry)
```
blanquicet, line 250: The only filtering available is `--containername`, and given that it cannot be used for docker yet, we are capturing the events of all the containers. Given that, I suspect `ExpectAllToMatch` is working in this case because the bind events are events that are not triggered very frequently. To be sure the tests will succeed, I think we should use `ExpectEntriesToMatch` until we are able to filter the output. WDYT?

integration/local-gadget/integration_test.go
blanquicet: Yeah, there was a fix to change the default from "0" to "". Probably that's the issue you were hitting.

integration/local-gadget/integration_test.go
blanquicet: Mmmm you can mention https://github.com/kinvolk/inspektor-gadget/issues/644.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,262 @@
...
+			if err := json.Unmarshal([]byte(output), &containers); err != nil {
+				return err
+			}
+
+			for _, gotContainer := range containers {
+				normalize(gotContainer)
+				if reflect.DeepEqual(gotContainer, expectedContainer) {
+					return nil
+				}
+			}
```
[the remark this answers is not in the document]
mqasimsarfraz: Sounds good. I want to get out of this rebase loop I am in right now :smile:

integration/local-gadget/Makefile
```diff
@@ -0,0 +1,79 @@
...
+
+# minikube
+
+MINIKUBE = $(shell pwd)/bin/minikube-$(MINIKUBE_VERSION)
+.PHONY: minikube
+minikube:
+	mkdir -p bin
+	test -e bin/minikube-$(MINIKUBE_VERSION) || \
+	(cd bin && curl -Lo ./minikube-$(MINIKUBE_VERSION) https://github.com/kubernetes/minikube/releases/download/$(MINIKUBE_VERSION)/minikube-linux-amd64)
+	chmod +x bin/minikube-$(MINIKUBE_VERSION)
```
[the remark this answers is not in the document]
mqasimsarfraz: Definitely, makes sense to do that. Sounds good, I can handle this a follow up PR.

mqasimsarfraz:
@blanquicet thanks for all the reviews. I have updated the PR to handle all the feedback!

integration/local-gadget/Makefile
```diff
@@ -0,0 +1,81 @@
...
+	test-containerd \
+	test-cri-o
+
+.PHONY: test-all
+test-all: $(TEST_TARGETS) test
+
+test: test-$(CONTAINER_RUNTIME)
+
+.PHONY: phony_explicit
+test-%: build
```
blanquicet, line 67: Could you please add here a comment with an example of how `INTEGRATION_TESTS_PARAMS` can be used like you did for the inpektor-gadget tests?

```bash
# INTEGRATION_TESTS_PARAMS="-test.run TestListContainers" CONTAINER_RUNTIME=containerd make -C integration/local-gadget test
```

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+			for _, gotContainer := range containers {
+				normalize(gotContainer)
+				if !reflect.DeepEqual(gotContainer, expectedContainer) {
+					return fmt.Errorf("expect at leat one containers to match, expected=%+v, got=%+v", expectedContainer, gotContainer)
+				}
+			}
+
+			return nil
```
blanquicet, line 184: This logic will make pass the test if the list of the returned containers is empty.

blanquicet (commented):
IMO, after fixing the logic in `TestListSingleContainer`, it is ready to be merged.

What do other folks think about it?

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,255 @@
...
+				c.PodUID = ""
+			}
+
+			var containers []*containercollection.Container
+			if err := json.Unmarshal([]byte(output), &containers); err != nil {
+				return err
+			}
+
+			if len(containers) == 0 {
+				return fmt.Errorf("expect at least one container in output=%s", output)
```
blanquicet, line 179: nit: `output` is already being printed by `Command.Run`.

blanquicet (approved):
LGTM! I really like the work you did here @mqasimsarfraz 💯 !!

Let's give some time before merging it to see if other folks have comments.

integration/local-gadget/Makefile
```diff
@@ -0,0 +1,83 @@
...
+	export MINIKUBE_PROFILE=minikube-$* && \
+	echo "Checking minikube with profile $${MINIKUBE_PROFILE} is running ..." && \
+	docker inspect $${MINIKUBE_PROFILE} >/dev/null || (echo "Error: $${MINIKUBE_PROFILE} not running, exiting ..." && exit 1) && \
+	echo "Preparing minikube with profile $${MINIKUBE_PROFILE} for testing ..." && \
+	docker cp ../../local-gadget-linux-amd64 $${MINIKUBE_PROFILE}:/bin/local-gadget >/dev/null && \
+	go test -c -o local-gadget-integration.test ./... && \
+	docker cp local-gadget-integration.test $${MINIKUBE_PROFILE}:/bin/ >/dev/null 2>&1 && \
+	rm local-gadget-integration.test && \
+	echo "Running test in minikube with profile $${MINIKUBE_PROFILE} ..." && \
+	docker exec \
```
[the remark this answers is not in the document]
mqasimsarfraz: I looked up `minikube exec` and stopped there :smile:  I switched it to `minikube ssh` now, I like it that we are doing everything with minikube now. Thanks for the pointer.

mqasimsarfraz:
@eiffel-fl Thanks for the review and the feedback. I will keep a note of this for future PRs and will split into multiple commits. Sorry for the pain :laughing: I have updated the PR to address some of the comments you had!

integration/local-gadget/Makefile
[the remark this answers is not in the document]
blanquicet: Agree! Great suggestion Francis!

mqasimsarfraz:
Sounds good. I will go ahead and merge this by EOD if no one has objections. I plan to create a follow up PR for suggested improvements!

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+			if *containerRuntime == val {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			fmt.Fprintf(os.Stderr, "Error: invalid argument '-container-runtime': %q. Valid values: %s\n",
+				*containerRuntime, strings.Join(supportedContainerRuntimes, ", "))
+			os.Exit(-1)
```
mauriciovasquezbernal, line 49: Any reason to use -1? We're using 1 in all places.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+	} else {
+		fmt.Println("Error: '-container-runtime' must be specified")
+		os.Exit(1)
+	}
```
mauriciovasquezbernal, line 54: nit: You can move this to L36 and avoid the `*containerRuntime != ""` check. Something like:

```go 
	if *containerRuntime == "" {
		fmt.Println("Error: '-container-runtime' must be specified")
		os.Exit(1)
	}

	found := false
	for _, val := range supportedContainerRuntimes {
		if *containerRuntime == val {
			found = true
			break
		}
	}

	if !found {
		fmt.Fprintf(os.Stderr, "Error: invalid argument '-container-runtime': %q. Valid values: %s\n",
			*containerRuntime, strings.Join(supportedContainerRuntimes, ", "))
		os.Exit(-1)
	}
```

integration/local-gadget/integration_test.go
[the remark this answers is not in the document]
mauriciovasquezbernal: +1 on implementing a helper. We'll get crazy with the naming of the two different cases but let's see.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+				}
+			}
+
+			return fmt.Errorf("output doesn't contain the expected container: %+v", expectedContainer)
+		},
+	}
+
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		BusyboxPodRepeatCommand(ns, "sleep inf"),
```
mauriciovasquezbernal, line 108:
```suggestion
		BusyboxPodCommand(ns, "sleep inf"),
```

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+		DeleteTestNamespaceCommand(ns),
+	}
+
+	RunCommands(commands, t)
+}
+
+func TestListSingleContainer(t *testing.T) {
+	t.Parallel()
+	prefix := "test-list-single"
+	po := fmt.Sprintf("%s-pod", prefix)
```
mauriciovasquezbernal, line 120: nit: is `po` and easy to understand variable name?

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+		Cmd:  fmt.Sprintf("local-gadget list-containers -o json --runtimes=%s", *containerRuntime),
+		ExpectedOutputFn: func(output string) error {
+			expectedContainer := &containercollection.Container{
+				Podname:   "test-pod",
+				Runtime:   *containerRuntime,
+				Namespace: ns,
+			}
+
+			normalize := func(c *containercollection.Container) {
+				c.Name = ""
```
mauriciovasquezbernal, line 75: I suppose this will be also handled when https://github.com/kinvolk/inspektor-gadget/issues/737 is fixed.

integration/local-gadget/integration_test.go
mauriciovasquezbernal: Yes, I agree. In this case we're testing Kubernetes containers I expected both tests to be almost identical.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+			ExpectedString: fmt.Sprintf("pod/%s condition met\n", po),
+		},
+		listContainersCmd,
+		DeleteTestNamespaceCommand(ns),
+	}
+
+	RunCommands(commands, t)
+}
+
+func TestTraceTCPBind(t *testing.T) {
```
mauriciovasquezbernal, line 212: Why `TraceTCPBind` instead of only `TraceBind`?

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
+package main
```
mauriciovasquezbernal, line 1: nit: license header missing.

integration/local-gadget/integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+
```
mauriciovasquezbernal, line 2: I think splitting this file into a per gadget file will help us (https://github.com/kinvolk/inspektor-gadget/issues/974). What about splitting the file now before merging this PR?

integration/local-gadget/integration_test.go
mqasimsarfraz: I saw that we are using `ns` so I thought using short form from `kubectl` for pods i.e `po` would be okay. What do you say?

mauriciovasquezbernal: I didn't know `po` was common. Let's keep po, that should be fine.

mqasimsarfraz:
@mauriciovasquezbernal thanks for the review. I have adjusted the nits. Good point regrading introducing `k8s` directory, Currently  I see we can have `docker` or `k8s` based test but wanted to have deeper look into it. Would it be okay if handle this in follow up PR ?

mauriciovasquezbernal (approved):

mauriciovasquezbernal:
@mqasimsarfraz , yes, that's fine to handle in a different PR.

PR #956, opened 2022-09-19
advise/seccomp: Fix logic around container termination notification

mauriciovasquezbernal:
It's possible to create multiple instances of the seccomp advisor gadget, however it internally uses a single eBPF program that captures all the syscalls on the host.

The current logic is wrong as it only subscribes to the container termination events taking into consideration the filter spec of the first seccomp advisor instance. Hence, the following case doesn't work:

```
$ kubectl run mypod1 --restart=Never --image=praqma/network-multitool \
	-- bash -c "while true; do ping -c 1 8.8.8.8 && sleep 1; done"
$ kubectl run mypod2 --restart=Never --image=praqma/network-multitool \
	-- bash -c "while true; do ping -c 1 8.8.8.8 && sleep 1; done"

$ kubectl -n gadget get seccompprofile
No resources found in gadget namespace.

$ ./kubectl-gadget advise seccomp-profile start -m seccomp-profile -p mypod1

$ ./kubectl-gadget advise seccomp-profile start -m seccomp-profile -p mypod2

$ kubectl delete pod mypod2

$ kubectl -n gadget get seccompprofile
No resources found in gadget namespace
```

After this commit, a profile is created for mypod2 by following the same steps.

### DIFF
--- a/pkg/gadget-collection/gadgets/advise/seccomp/gadget.go
+++ b/pkg/gadget-collection/gadgets/advise/seccomp/gadget.go
@@ -411,34 +411,35 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
 			trace.Status.OperationError = fmt.Sprintf("Failed to start seccomp tracer: %s", err)
 			return
 		}
-
-		// 'trace' is owned by the controller and could be modified
-		// outside of the gadget control. Make a copy for the callback.
-		traceCopy := trace.DeepCopy()
-
-		// Subscribe to container creation and termination
-		// events. Termination is used to generate a
-		// SeccompProfile when a container terminates. Creation
-		// is used to fetch the owner reference of the
-		// containers to be sure this field is set when the
-		// container terminates.
-		containers := t.helpers.Subscribe(
-			genPubSubKey(trace.ObjectMeta.Namespace+"/"+trace.ObjectMeta.Name),
-			*gadgets.ContainerSelectorFromContainerFilter(trace.Spec.Filter),
-			func(event containercollection.PubSubEvent) {
-				switch event.Type {
-				case containercollection.EventTypeAddContainer:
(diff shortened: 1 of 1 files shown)


pkg/gadget-collection/gadgets/advise/seccomp/gadget.go
```diff
@@ -392,27 +392,28 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
...
+	// outside of the gadget control. Make a copy for the callback.
+	traceCopy := trace.DeepCopy()
+
+	// Subscribes to container termination events in order to
+	// generate a SeccompProfile when a container terminates. We
+	// don't need the list of existing containers, so the return
+	// value is ignored.
+	_ = t.helpers.Subscribe(
+		genPubSubKey(trace.ObjectMeta.Namespace+"/"+trace.ObjectMeta.Name),
+		*gadgets.ContainerSelectorFromContainerFilter(trace.Spec.Filter),
```
blanquicet, line 407: With this change, the second instance of the seccomp gadget will completely overwrite the filter, which means that we will not generate the profile for `mypod1`. Is it what we want?

mauriciovasquezbernal: Each trace (`gadgetv1alpha1.Trace`) object has a different filter that is used in `Subscribe()`. What this PR is making is that filters from another seccomp advisor instances are also taken into consideration, so when those pods terminate the profile is created.

blanquicet: Aahhh ok, it is because each trace subscribes with a diff key, so there will be no overwritting at all. Got it!

blanquicet (approved):
Tested and working as expected. Thanks!

PR #961, opened 2022-09-20
Implement local-gadget version of tcpconnect

wedaly:
# Implement local-gadget version of tcpconnect

Add tcpconnect trace to local-gadget.
Resolves: https://github.com/kinvolk/inspektor-gadget/issues/895

## How to use

1. `sudo local-gadget trace tcpconnect`
2. `docker run -it --rm --name test-container busybox /bin/sh -c "wget http://www.example.com"`

## Testing done

Tested locally:
image

### DIFF
--- a/docs/local-gadget.md
+++ b/docs/local-gadget.md
@@ -246,6 +246,24 @@ CONTAINER        T  PID     COMM             IP  SADDR                  DADDR   
 test-container   C  11039   wget             4   172.17.0.2             188.114.96.7           57560   443
 ```
 
+### Trace/TcpConnect
+
+The tcpconnect trace gadget traces IPv4 and IPv6 TCP connections.
+
+```bash
+$ docker run -it --rm --name test-container busybox /bin/sh -c "wget http://www.example.com"
+Connecting to www.example.com (93.184.216.34:80)
+saving to 'index.html'
+index.html           100% |************************************************************************************************|  1256  0:00:00 ETA
+'index.html' saved
+```
+
+```bash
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal (approved):
I tested and it worked as expected. LGTM!

docs/local-gadget.md
```diff
@@ -246,6 +246,24 @@ CONTAINER        T  PID     COMM             IP  SADDR                  DADDR
...
+```bash
+$ docker run -it --rm --name test-container busybox /bin/sh -c "wget http://www.example.com"
+Connecting to www.example.com (93.184.216.34:80)
+saving to 'index.html'
+index.html           100% |************************************************************************************************|  1256  0:00:00 ETA
+'index.html' saved
+```
+
+```bash
+$ sudo local-gadget trace tcpconnect
```
blanquicet, line 262: What do you think about using the `--containername` to ensure we just capture the connection we generated? Otherwise, the output could contain many other entries 🙂 . I would suggest adding it for the trace/tcp gadget as well.

```suggestion
$ sudo local-gadget trace tcpconnect --containername test-container
```

blanquicet (approved):
Just one comment on the documentation, but I tested and LGTM! Thanks!

docs/local-gadget.md
blanquicet: Could you please squash last two commits?

wedaly: sure thing! just updated the branch. thanks!

PR #963, opened 2022-09-20
Fix runcfanotify dropping events for docker-ce on WSL2 / Ubuntu 22.04

wedaly:
# Fix runcfanotify missing events on WSL2 Ubuntu 22.04

On Ubuntu 22.04 WSL2, /var/run is symlinked to /run.
Docker-CE generates a containerd configuration that sets the state path
to /var/run/docker/containerd/daemon, which containerd uses to construct
runc's --pid-file path. However, the fanotify event path uses /run,
and since the paths don't match, runcfanotify returns false.

This had the effect of "dropping" the container create event,
so that (1) the container did not show up in trace results, and
(2) on SIGINT, the runcfanotify Close() method would hang waiting on
the read sycall in `pidFileDirNotify.GetEvent()`.

Fix it by using `os.Samefile` to compare the device/inode of the
two files instead of comparing paths.

Resolves: https://github.com/kinvolk/inspektor-gadget/issues/958

## How to use

Follow the steps to reproduce from https://github.com/kinvolk/inspektor-gadget/issues/958
After this change, TCP trace events appear for containers started after `local-gadget`, and CTRL-C exits `local-gadget` without hanging.

## Testing done

Manually tested following the steps to reproduce in https://github.com/kinvolk/inspektor-gadget/issues/958

### DIFF
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -371,7 +371,15 @@ func (n *RuncNotifier) watchPidFileIterate(pidFileDirNotify *fanotify.NotifyFD, 
 	if err != nil {
 		return false, err
 	}
-	if path != pidFile {
+
+	// Consider files identical if they have the same device/inode,
+	// even if the paths differ due to symlinks (for example,
+	// the event's path is /run/... but the runc --pid-file argument
+	// uses /var/run/..., where /var/run is a symlink to /run).
+	filesAreIdentical, err := checkFilesAreIdentical(path, pidFile)
+	if err != nil {
+		return false, err
+	} else if !filesAreIdentical {
 		return false, nil
 	}
 
@@ -422,6 +430,20 @@ func (n *RuncNotifier) watchPidFileIterate(pidFileDirNotify *fanotify.NotifyFD, 
 		Bundle:          bundleDir,
 	})
 	return true, nil
+}
+
+func checkFilesAreIdentical(path1, path2 string) (bool, error) {
+	f1, err := os.Stat(path1)
+	if err != nil {
+		return false, err
+	}
+
+	f2, err := os.Stat(path2)
+	if err != nil {
+		return false, err
+	}
+
+	return os.SameFile(f1, f2), nil
 }
 
 func (n *RuncNotifier) monitorRuncInstance(bundleDir string, pidFile string) error {


pkg/runcfanotify/runcfanotify.go
```diff
@@ -371,7 +371,15 @@ func (n *RuncNotifier) watchPidFileIterate(pidFileDirNotify *fanotify.NotifyFD,
 	if err != nil {
 		return false, err
 	}
-	if path != pidFile {
+
+	// Some platforms (e.g. Ubuntu 22.04 on WSL2) symlink /var/run to /run.
+	// We need to evaluate the symlink before comparing with the event path below.
+	pidFileEvalSymlinks, err := filepath.EvalSymlinks(pidFile)
```
wedaly, line 377: not sure if it makes more sense to put this here or higher up when `pidFile` is first constructed. Thoughts?

wedaly: also not sure if it makes sense to also eval symlinks for `path` or not. In my testing, that seemed to use `/run` (no symlink), but maybe safer to normalize both strings being compared?

alban:
On Fedora and other Linux distros, /var/run is also a symlink to /run:
```
$ readlink /var/run
../run
```

I think the difference is that on my system, runc is started with `--pid-file /run/containerd/` and on yours it might be started with `--pid-file /var/run/containerd/`. In `/etc/containerd/config.toml` I have `state = "/run/containerd"`. Is yours containing `/var/run/...`?

```
$ sudo /usr/share/bcc/tools/execsnoop
PCOMM            PID    PPID   RET ARGS
runc             149588 149566   0 /usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/70080e7c9f8a94053c26f548334bd00fc4468982495fabd0ba7d743561ad47f7/log.json --log-format json create --bundle /run/containerd/io.containerd.runtime.v2.task/moby/70080e7c9f8a94053c26f548334bd00fc4468982495fabd0ba7d743561ad47f7 --pid-file /run/containerd/io.containerd.runtime.v2.task/moby/70080e7c9f8a94053c26f548334bd00fc4468982495fabd0ba7d743561ad47f7/init.pid --console-socket /tmp/pty244409015/pty.sock 70080e7c9f8a94053c26f548334bd00fc4468982495fabd0ba7d743561ad47f7
```

It might be safer to compare the files by checking the device & inode with [os.SameFile()](https://pkg.go.dev/os#SameFile), so that it would work even if local-gadget and runc resolve symlinks in a different ways (e.g. if they run in different mount namespaces with different mounts).

```
fileinfo1, err := os.Stat(path)
fileinfo2, err := os.Stat(filepath.Join("/proc/self/fd", fmt.Sprint(dataFile.Fd())))

if !os.SameFile(fileinfo1, fileinfo2) {
        return false, nil
}
```
(untested code)

wedaly:
> I think the difference is that on my system, runc is started with --pid-file /run/containerd/ and on yours it might be started with --pid-file /var/run/containerd/. In /etc/containerd/config.toml I have state = "/run/containerd". Is yours containing /var/run/...?

Ah, yes, that makes sense. I just checked on my system, and I'm seeing --pid-file set to /var/run
image

I checked /etc/containerd/config.toml, and I'm seeing the "state" key commented out:
image
Since the --pid-file path starts with /var/run/docker/containerd I'm guessing it's getting set by docker somewhere, but I'm not sure where.

It would be interesting to see if this reproduces on Ubuntu 22.04 and docker-ce without WSL2. Might spin up a VM later to test it.

wedaly:
> It might be safer to compare the files by checking the device & inode with [os.SameFile()](https://pkg.go.dev/os#SameFile), so that it would work even if local-gadget and runc resolve symlinks in a different ways (e.g. if they run in different mount namespaces with different mounts).

Nice, I'll try that out!

wedaly:
> Since the --pid-file path starts with /var/run/docker/containerd I'm guessing it's getting set by docker somewhere, but I'm not sure where.

Found it!
image


/var/run/docker/containerd/containerd.toml
```
version = 0
root = "/var/lib/docker/containerd/daemon"
state = "/var/run/docker/containerd/daemon"
```

wedaly:
Updated the PR and commit description to clarify that the /var/run is being set by docker.
Updated the code to use `os.SameFile` for the file comparison and verified that this also resolves the issue on Ubuntu 22.04 and WSL2.

I ended up using `os.Stat` on both paths without constructing a path to `/proc/self/fd`. Is there a reason to prefer using `/proc/self/fd`?

alban (approved):
LGTM if the CI is green.

I wonder if the runcfanotify Close() method should also call `pidFileDirNotify.File.Close()` for each pending runc call... but that's a different issue.

PR #965, opened 2022-09-21
trace/capabilities: Use *bool for InsetID

mauriciovasquezbernal:
InsetID represents the INSETID used in cap_capable() for kernel >= 5.1. It's possible to represent it by using *bool instead of string. nil indicates that the running kernel is < 5.1.

### DIFF
--- a/pkg/gadgets/trace/capabilities/tracer/tracer.go
+++ b/pkg/gadgets/trace/capabilities/tracer/tracer.go
@@ -25,7 +25,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strconv"
 	"unsafe"
 
 	"github.com/cilium/ebpf"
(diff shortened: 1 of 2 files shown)


alban:
`InsetID` does not seem to be displayed by the CLI:
```
$ kubectl gadget trace capabilities --selector name=set-priority
NODE             NAMESPACE        POD                            CONTAINER        PID     COMM             UID     CAP  NAME             AUDIT  VERDICT
minikube         default          set-priority-5646554d9d-pk4gg  set-priority     110385  nice             0       23   CAP_SYS_NICE     1      Deny
```

Do we need to keep this field at all?

mauriciovasquezbernal:
I think we can keep it. Even if this is not shown it can be useful for consumers parsing the json output. Also, I think in the future we could add a `--verbose` option for gadgets to print more information in the CLI.

mauriciovasquezbernal:
> In this case, how `nil` will be rendered once printed?

We could show it as `N/A` as done in bcc.

mauriciovasquezbernal:
That's right because we're using `omitempty`.

PR #966, opened 2022-09-21
Add Tilt support

### LINKED ISSUE #952: Add Tilt support for rapid development
Looking at https://github.com/kinvolk/inspektor-gadget/blob/main/docs/CONTRIBUTING.md, it seems we are currently relying on pushing images to a registry during development. This can greatly slow down development.

[Tilt](https://tilt.dev/) allows rapid development by automatically updating local code changes on a remote k8s cluster, thus saving the need to rebuild a container image, push it and re-deploy it to the cluster. It is free and open source.

P. S. Not sure if "local gadget" addresses this concern. Still, even if it does, there are probably some cases where there is no substitute for running against a real k8s cluster, in which case Tilt support can be useful.

johananl:
# Add Tilt support

This adds initial support for rapid iterative development on Gadget Tracer Manager using [Tilt](https://tilt.dev/). We may want to extend Tilt support to other relevant components (if there are any) later on.

>NOTE: Tilt relies heavily on the ability to rebuild container images efficiently. Right now rebuilding the image is rather slow (more than 1 minute), so we probably want to optimize the Dockerfile to better utilize caching etc. Ideally, Tilt should be able to complete the build->push->deploy cycle in a few seconds for the best dev experience.

Fixes #952.

## How to use

Follow the instructions in `docs/tilt.md`.

## Testing done

I verified I can automatically deploy changes to Gadget Tracer Manager to my dev cluster by following the instructions included in this PR.


docs/tilt.md
```diff
@@ -0,0 +1,30 @@
...
+Set your environment:
+
+```shell
+# Name of your cluster's k8s context (check kubectl config get-contexts).
+export IG_K8S_CLUSTER=my-cluster
+# A Docker registry to use for development. Tilt will push temporary images there.
+export IG_REGISTRY=quay.io
+# A Docker repository to use for development. Tilt will push temporary images there.
+export IG_REPOSITORY=me/inspektor-gadget
+```
```
alban, line 19: In the Makefile and in the CI, we use this variable:
```
CONTAINER_REPO ?= ghcr.io/kinvolk/inspektor-gadget
```

Could we reuse the same convention?

docs/tilt.md
```diff
@@ -0,0 +1,30 @@
...
+Ensure you are logged into your Docker registry (run `docker login` if necessary).
+
+Run Tilt:
+
+```shell
+tilt up
+```
+
+Hit Space and watch the progress in your browser. If you leave Tilt running in the background, any
+Tilt would rebuild, push and redeploy any code changes automatically when you modify source files.
```
alban, line 30: > when you modify source files

Which source files does it consider?
*.go and *.c?
Does it consider the .dockerignore files? (any changes in `cmd/*` is about the CLI so no need to rebuild the gadget container)

johananl: I think it considers all files by default. I'm not sure how smart it is, i.e. maybe it triggers for doc changes, too.
Tilt does support ignoring files. Let me look into that and ensure we're filtering correctly.

johananl: Also, I suspect we might need to trigger an eBPF rebuild in Tilt to ensure we bake up-to-date object files in the dev image.

docs/tilt.md
johananl: Done.

alban:
Out of curiosity, which Kubernetes distribution and which container registry did you use for testing this? AKS and quay.io?

If it is Minikube, we have `make minikube-install` so we don't need an external container registry.

johananl:
> Out of curiosity, which Kubernetes distribution and which container registry did you use for testing this? AKS and quay.io?
> 
> If it is Minikube, we have `make minikube-install` so we don't need an external container registry.

Yes, AKS and quay.io.

Tilt actually supports Minikube and leverages its local registry features for fast pushes. Is Minikube a complete replacement of a "real" k8s cluster? If so, we can use Tilt with Minikube. IIUC Tilt is even smart enough to figure out the type of k8s cluster so maybe we can use Minikube like a generic cluster without special config.

alban:
> Is Minikube a complete replacement of a "real" k8s cluster in the context of IG?

Almost, AFAIR Minikube works with all gadgets (both with the docker driver or the VM driver) except for traceloop where it currently requires the VM driver, see [install.md#minikube](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/install.md#minikube). But @eiffel-fl is working on a new implementation of traceloop which would not have this limitation.

johananl:
Cool, then I'd say let's focus on Minikube for rapid iteration and adjust Tilt accordingly. For example, we can get rid of the need for a hosted registry and Docker authentication 👏

alban:
> Cool, then I'd say let's focus on Minikube for rapid iteration and adjust Tilt accordingly. For example, we can get rid of the need for a hosted registry and Docker authentication clap

Note that with `make minikube-install`, we don't need hosted registry or Docker authentication.

But I think Tilt support can still be useful in general (there can be more than one good developer workflow).

docs/tilt.md
mauriciovasquezbernal: I personally have the (bad?) habit of hitting `Ctrl+S` many times to be sure I don't lose any change. Will it trigger the whole redeploy process each time I save a file?

PR #969, opened 2022-09-21
Rerun make ebpf-objects

johananl:
# Rerun make ebpf-objects

In f3aa2fbfa8acca7e039808aecf43eb09f07e050d we forgot to run make ebpf-objects, thus leading to a diff on main whenever the command is run. This fixes the problem.

## How to use

Run `make ebpf-objects` and ensure there is no diff.

## Testing done

I did the above.


alban (approved):
LGTM.
I tested it as well, and the diff is gone.

PR #971, opened 2022-09-21
integration: Switch trace open tests to use json

mauriciovasquezbernal:
While reviewing https://github.com/kinvolk/inspektor-gadget/pull/945 I tried to implement some tests by myself to check how easy it was. Opening a PR to not lose that work.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -28,6 +28,7 @@ import (
 
 	bindTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/bind/types"
 	execTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/exec/types"
+	openTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/open/types"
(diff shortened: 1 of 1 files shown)


integration/integration_test.go
```diff
@@ -775,16 +776,34 @@ func TestOpensnoop(t *testing.T) {
...
+			}
+
+			normalize := func(e *openTypes.Event) {
+				e.Node = ""
+				e.MountNsID = 0
+				e.Pid = 0
+				e.UID = 0
+			}
+
+			return expectEntriesToMatch(output, normalize, expectedEntry)
```
blanquicet, line 799: Why not `expectAllToMatch`?

mauriciovasquezbernal: I wasn't sure all events would match (there could be other open() calls in the container). I tried and it seems to work fine, so let's use that instead.

blanquicet: I tested it as well and I didn't see any other entry.

blanquicet (approved):
LGTM!

integration/integration_test.go
mauriciovasquezbernal: My initial concern was right after all -> https://github.com/kinvolk/inspektor-gadget/pull/984. :D

PR #973, opened 2022-09-21
container-collection: Enrich container with k8s info from container runtime

blanquicet:
# Enrich container with k8s info from container runtime

https://github.com/kinvolk/inspektor-gadget/pull/824 retrieves the Kubernetes information from the container runtime, but we were not enriching the containers with such information when using the `WithContainerRuntimeEnrichment` option.

## How to use

Check the namespace, pod and pod-uid fields when using `WithContainerRuntimeEnrichment` option. For instance, `local-gadget`  uses this option.

### Before this PR
Both k8s and no-k8s containers were enriched with the default info: namespace="default" and podname=container-name.

```bash
$ sudo /inspektor-gadget/local-gadget list-containers -o json --containername etcd
[
  {
    "id": "3ea21dfe1b9191c5dca7c3e1b4b102b7e6edd0f456d5651b9c1899e18eccccdf",
    [...]
    "namespace": "default",
    "podname": "etcd",
    "name": "etcd",
    "runtime": "containerd"
  }
]
```
```bash
$ sudo /inspektor-gadget/local-gadget list-containers -o json --containername test-container
[
  {
    "id": "749c588b1aef5997ed5b8ae7d5eaade350f6f2723c097e2faf618af04cbb5e20",
    [...]
    "namespace": "default",
    "podname": "test-container",
    "name": "test-container",
    "runtime": "docker"
  }
]
```

### After this PR
Now, the k8s containers are now enriched with the correct information. Notice those default values are needed because some gadgets use the Trace CRD approach (`local-gadget` interactive mode).

```bash
$ sudo /inspektor-gadget/local-gadget list-containers -o json --containername etcd
[
  {
    "id": "3ea21dfe1b9191c5dca7c3e1b4b102b7e6edd0f456d5651b9c1899e18eccccdf",
    [...]
    "namespace": "kube-system",
    "podname": "etcd-master",
    "name": "etcd",
    "podUID": "87a960e902bbb19289771a77e4b07353",
    "runtime": "containerd"
  }
]
```
```bash
$ sudo /inspektor-gadget/local-gadget list-containers -o json --containername test-container
[
  {
    "id": "749c588b1aef5997ed5b8ae7d5eaade350f6f2723c097e2faf618af04cbb5e20",
    [...]
    "namespace": "default",
    "podname": "test-container",
    "name": "test-container",
    "runtime": "docker"
  }
]
```

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -36,6 +36,31 @@ import (
 	"github.com/kinvolk/inspektor-gadget/pkg/runcfanotify"
 )
 
+func enrichContainerWithContainerData(containerData *runtimeclient.ContainerData, container *Container) {
+	// Runtime
+	container.ID = containerData.ID
+	container.Runtime = containerData.Runtime
+
+	// Kubernetes
+	container.Namespace = containerData.PodNamespace
+	container.Podname = containerData.PodName
+	container.PodUID = containerData.PodUID
+
+	// Notice we are temporarily using the runtime container name as the
+	// Kubernetes container name because the Container struct doesn't have that
+	// field, and we don't support filtering by runtime container name yet.
+	container.Name = containerData.Name
+
+	// Some gadgets using the Trace CRD approach in local-gadget require the
+	// namespace and pod name to be set.
+	if container.Namespace == "" {
+		container.Namespace = "default"
+	}
+	if container.Podname == "" {
+		container.Podname = containerData.Name
+	}
+}
+
 func containerRuntimeEnricher(
 	runtimeName string,
 	runtimeClient runtimeclient.ContainerRuntimeClient,
@@ -47,19 +72,14 @@ func containerRuntimeEnricher(
 		return true
 	}
 
-	c, err := runtimeClient.GetContainer(container.ID)
+	containerData, err := runtimeClient.GetContainer(container.ID)
 	if err != nil {
 		log.Debugf("Runtime enricher (%s): failed to get container: %s",
 			runtimeName, err)
 		return true
 	}
 
-	container.Name = c.Name
-	container.Runtime = c.Runtime
-	container.ID = c.ID
-	// Some gadgets require the namespace and pod name to be set
-	container.Namespace = "default"
-	container.Podname = container.Name
+	enrichContainerWithContainerData(containerData, container)
 
 	return true
 }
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (commented):
I tested and worked fine. I'm wondering if this is an intermediate step until we add the information about the runtime container name & namespace to the container structure? Also, why do we need this PR before implementing it?

pkg/container-collection/options.go
```diff
@@ -36,6 +36,31 @@ import (
...
+
+	// Kubernetes
+	container.Namespace = containerData.PodNamespace
+	container.Podname = containerData.PodName
+	container.PodUID = containerData.PodUID
+
+	// Notice we are temporarily using the runtime container name as the
+	// Kubernetes container name because the Container struct doesn't have that
+	// field, and we don't support filtering by runtime container name yet.
+	container.Name = containerData.Name
```
blanquicet, line 52: Take into account that this PR is not introducing this change, this is already the current implementation.

blanquicet:
> I'm wondering if this is an intermediate step until we add the information about the runtime container name & namespace to the container structure? Also, why do we need this PR before implementing it?
>
Mmmmm, it is an intermediate step, but it is not temporary because it won't get completely lost once we add the runtime container name & namespace. In particular, my next PR will:

- Add the `Kubernetes` or `Runtime` prefix to the fields in the `Container` and `ContainerData` structs.
- Add the `RuntimeContainerName` to `Container` and the `KubernetesContainerName` to `ContainerData`.
- And finally, add the support to the container-collections library for filtering by `RuntimeContainerName`. At this point, we will be able to remove [this](https://github.com/kinvolk/inspektor-gadget/blob/4ac2aea2e47d42d0ec7562abe7d354407c30705d/pkg/container-collection/options.go#L49-L52) workaround. But [the actual change](https://github.com/kinvolk/inspektor-gadget/blob/4ac2aea2e47d42d0ec7562abe7d354407c30705d/pkg/container-collection/options.go#L44-L47) that is introducing this PR will remain there.

Therefore, the next PR will be pretty big (even though mostly because of the renaming), while the changes introduced by this PR were very few, well delimited and easy to review/test. That's why I decided to create it. In addition, this PR can already be used by the integration tests, see https://github.com/kinvolk/inspektor-gadget/pull/946#discussion_r976845823.

Note: I will manage the runtime container namespace in a third PR.

Does it make sense to you? Sorry for not explaining the plan from the very beginning.

PR #977, opened 2022-09-22
cmd/version: Additionally show server version

burak-ok:
# `version` command outputs server version next to the client version

The `version` command now additionally outputs the deployed Inspektor Gadget version.
The information is from the deployed image-tag of the pods.
Solves https://github.com/kinvolk/inspektor-gadget/issues/902

## Testing done

_Versions are commit hashes, since these commits are not tagged._

```
➭ ./kubectl-gadget version
Client version: af1512e
Server version: v0.8.0
```

## Further possibilities
1. Additional message prompting the user to take action. For Example: `Please update Inspektor Gadget on the pods using kubectl gadget deploy`, ....

### DIFF
--- a/cmd/kubectl-gadget/version.go
+++ b/cmd/kubectl-gadget/version.go
@@ -15,12 +15,17 @@
 package main
 
 import (
+	"context"
 	"fmt"
+	"strings"
 
 	"github.com/blang/semver"
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
+	"github.com/kinvolk/inspektor-gadget/pkg/k8sutil"
 )
 
 // This variable is used by the "version" command and is set during build.
@@ -35,7 +40,47 @@ func init() {
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Show version",
-	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println(version)
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Println("Client version:", version)
+
+		client, err := k8sutil.NewClientsetFromConfigFlags(utils.KubernetesConfigFlags)
+		if err != nil {
+			return commonutils.WrapInErrSetupK8sClient(err)
+		}
+
+		opts := metav1.ListOptions{LabelSelector: "k8s-app=gadget"}
+		pods, err := client.CoreV1().Pods("gadget").List(context.TODO(), opts)
+		if err != nil {
+			return commonutils.WrapInErrListNodes(err)
+		}
+
+		serverVersions := make(map[string]struct{})
+		for _, pod := range pods.Items {
+			image := pod.Spec.Containers[0].Image
+
+			// Get the image tag
+			parts := strings.Split(image, ":")
+			if len(parts) < 2 {
+				continue
+			}
+
+			versionStr := parts[len(parts)-1]
+			if _, ok := serverVersions[versionStr]; !ok {
+				serverVersions[versionStr] = struct{}{}
+			}
+		}
+
+		if len(serverVersions) == 0 {
+			fmt.Println("Server version:", "not installed")
+		} else {
+			if len(serverVersions) > 1 {
+				fmt.Println("Warning: Multiple deployed versions detected")
+			}
+			for version := range serverVersions {
+				fmt.Println("Server version:", version)
+			}
+		}
+
+		return nil
 	},
 }


cmd/kubectl-gadget/deploy.go
```diff
@@ -296,6 +296,8 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 				delete(daemonSet.Spec.Template.Spec.NodeSelector, "kubernetes.io/os")
 			}
 
+			daemonSet.Annotations[utils.GadgetVersionAnnotation] = version
```
mauriciovasquezbernal, line 299: `version` refers to the kubectl-gadget version in this case. If you deploy something like `kubectl gadget --image=fooo:v0.8.0`, version isn't v0.8.0 but the kubectl-gadget version used to deploy it. Infact, the `INSPEKTOR_GADGET_VERSION` is wrong/misleading. 

https://github.com/kinvolk/inspektor-gadget/blob/585dda7d7924fa2f6830943130cd88044620ea5c/cmd/kubectl-gadget/deploy.go#L317-L318

What about basing the version check on the image tag as I did in https://github.com/kinvolk/inspektor-gadget/pull/936? I know `latest` can be a bit difficult to handle, in that case we can only print this without performing any version comparison. 

Also, it'd be nice to check the versions of all pods and print a different message if there are pods running different versions. (in case the installation is really messed up).

alban: > Also, it'd be nice to check the versions of all pods and print a different message if there are pods running different versions. (in case the installation is really messed up).

This should not happen with a single DaemonSet. I think it is simpler to check the DS, also because with the node selector, the gadget pods might only exist on a subset of nodes (e.g. with a mix of Linux & Windows nodes).

burak-ok: As far as i understood, when checking the DaemonSet there is still the problem with the version mismatch if someone runs `kubectl gadget --image=fooo:v0.8.0`.
The DaemonSet is going to be tagged with the `clientVersion` even though the image/server version can be a different one.

mauriciovasquezbernal: > This should not happen with a single DaemonSet

By "in case the installation is really messed up" I meant cases where the user could be running multiple daemonsets of Inspektor Gadget. I know it's not very probably, but if we're able to print a warning message in that case it would be great. 

> the gadget pods might only exist on a subset of nodes (e.g. with a mix of Linux & Windows nodes).

Is this a problem? I was thinking to get pods by labels (`k8s-app=gadget`) and verify those only.

mauriciovasquezbernal: > As far as i understood, when checking the DaemonSet there is still the problem with the version mismatch if someone runs kubectl gadget --image=fooo:v0.8.0.

I think by verifying the daemon set Alban intended to check the image on the daemonset spec:

```
$ kubectl get -n gadget daemonset gadget -o=jsonpath='{.spec.template.spec.containers[0].image}'
```

burak-ok:
Updated the PR

Changes:
1. The version is being read from the image tag
2. Querying all gadget pods

cmd/kubectl-gadget/version.go
```diff
@@ -36,6 +40,45 @@ var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Show version",
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println(version)
+
```
mauriciovasquezbernal, line 43: nit: extra empty line.

mauriciovasquezbernal (commented):
Some minor comments but it's looking very good!

cmd/kubectl-gadget/version.go
```diff
@@ -36,6 +40,45 @@ var versionCmd = &cobra.Command{
...
+		fmt.Println("Client version:", version)
+
+		client, err := k8sutil.NewClientsetFromConfigFlags(utils.KubernetesConfigFlags)
+		if err != nil {
+			return
```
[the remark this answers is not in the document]
burak-ok: Is there a standard way to log errors or warnings in IG?
Just println like i did in my latest version of this commit?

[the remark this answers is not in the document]
mauriciovasquezbernal: We use `fmt.Fprintf(os.Stderr` in the CLI to print errors. On the server part we use `"github.com/sirupsen/logrus"`.

cmd/kubectl-gadget/version.go
```diff
@@ -36,6 +40,46 @@ var versionCmd = &cobra.Command{
...
+
+		serverVersions := make(map[string]struct{})
+		for _, pod := range pods.Items {
+			image := pod.Spec.Containers[0].Image
+
+			// Get the image tag
+			parts := strings.Split(image, ":")
+			if len(parts) != 2 {
+				continue
+			}
```
blanquicet, line 66: This approach fails to extract the image tag in my development environment:

```bash
$ ./kubectl-gadget version
Client version: v0.8.0-55-g4289e7627b24-dirty
Server version: not installed
```

It is because I am using a local registry at `192.168.1.105:5000`, so there are two `:` 😅 :

```bash
$ k get pods -l k8s-app=gadget -n gadget -o jsonpath={.items[0].spec.containers[0].image}
192.168.1.105:5000/gadget:version
```

I think you can verify that `parts > 0`, and simply consider that `parts[len(parts)-1]` is the image tag. What do you think?

cmd/kubectl-gadget/version.go
```diff
@@ -36,6 +40,46 @@ var versionCmd = &cobra.Command{
...
+			if _, ok := serverVersions[versionStr]; !ok {
+				serverVersions[versionStr] = struct{}{}
+			}
+		}
+
+		if len(serverVersions) == 0 {
+			fmt.Println("Server version:", "not installed")
+		} else {
+			if len(serverVersions) > 1 {
+				fmt.Println("Warning: Multiple deployed versions detected")
```
blanquicet, line 78: What about printing the node, pod and version?

blanquicet (commented):
Thanks @burak-ok! I have some comments, but it looks good already 🙂.

cmd/kubectl-gadget/deploy.go
blanquicet: I am wondering if there is a way to make it more useful during the development phase where the image-tag is not the version but the name of the branch:

```bash
$ ./kubectl-gadget version
Client version: v0.8.0-55-g4289e7627b24-dirty
Server version: version
```

We can think/work on it in another PR.

cmd/kubectl-gadget/version.go
burak-ok: I think that parts[len(parts)-1] would be always the tag. (I am no Docker expert :D )

So yes, i would agree

cmd/kubectl-gadget/version.go
burak-ok: When do we print the node, pod and version?
Only if they are different?

What about if 50 pods are version X and 50 pods are version Y?
Do we print all of them?

These are the kind of questions that i would have

blanquicet: Good point. Forget about my proposal 🤣.

cmd/kubectl-gadget/version.go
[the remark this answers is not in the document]
burak-ok: For both failure scenarios i found a function in common errors, so i used them.
Personally i would prefer to have the failure message change in a separate commit and PR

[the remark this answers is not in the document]
blanquicet: Just one thing. If you use `RunE` instead of [`Run`](https://github.com/burak-ok/inspektor-gadget/blob/b4a2af37861e8556b9434651523806acf7061a64/cmd/kubectl-gadget/version.go#L44), the `cobra` library will take care of sending the error message to the `stderr`, and you just need to return what `WrapInErrSetupK8sClient` returns (what a pun 🤣).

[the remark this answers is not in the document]
burak-ok: Done. Also seems way better :D

cmd/kubectl-gadget/version.go
```diff
@@ -35,7 +40,47 @@ func init() {
...
+
+		client, err := k8sutil.NewClientsetFromConfigFlags(utils.KubernetesConfigFlags)
+		if err != nil {
+			return commonutils.WrapInErrSetupK8sClient(err)
+		}
+
+		opts := metav1.ListOptions{LabelSelector: "k8s-app=gadget"}
+		pods, err := client.CoreV1().Pods("gadget").List(context.TODO(), opts)
+		if err != nil {
+			return commonutils.WrapInErrListNodes(err)
```
blanquicet, line 54: I just realized we were incorrectly using this error when there is an error listing the **pods** but it actually talks about **nodes** (`failed to list nodes`) 🤦🏽. I think we can just make it more generic, something like `failed to list Kubernetes resources`. WDYT?

cmd/kubectl-gadget/version.go
burak-ok: Yes, that is intended to enumerate all detected versions

cmd/kubectl-gadget/version.go
burak-ok: Yep, that is a ListNodes error message, not a ListPods.
About generalizing the errors: Depends on how we would like to proceed with these kind of things. More generic or having many specific messages

mauriciovasquezbernal (approved):
I think it's ready to go. @blanquicet  do you want to review again?

blanquicet (approved):
LGTM!

cmd/kubectl-gadget/version.go
blanquicet: We can manage this in another PR 🙂

cmd/kubectl-gadget/deploy.go
burak-ok: It already does that (for me)
```
➭ ./kubectl-gadget version
Client version: 29968d8-dirty
Server version: burak-column-open
```

Or did you mean the branchname should also be in the Client version?

alban: kubectl gives more information than just one string:

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.2", GitCommit:"9d142434e3af351a628bffee3939e64c681afa4d", GitTreeState:"clean", BuildDate:"2022-01-19T17:35:46Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:23:26Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
```

Maybe we could reuse a similar structure with the GitCommit and GitTreeState fields?

blanquicet: I created https://github.com/inspektor-gadget/inspektor-gadget/issues/1060 to discuss this.

PR #978, opened 2022-09-22
trace/bind: Use empty string for iface as default instead of "0"

mauriciovasquezbernal:
The Interface field is the name of the interface that the socket is bound to. If the socket is not bound to any interface we can have an empty string as default.


blanquicet (approved):
LGTM

PR #980, opened 2022-09-22
local-gadget: Use GadgetParser to print list-containers columns

blanquicet:
# Use GadgetParser to print `local-gadge list-containers` output

This PR modifies the `list-containers` command to use the GadgetParser introduced in [in the CLI](https://github.com/kinvolk/inspektor-gadget/pull/934).

## Testing done

```bash
$ sudo /inspektor-gadget/local-gadget list-containers
RUNTIME    ID                         NAME
containerd 4a673ed444eda3b2d2e3dcde9… calico-kube-controllers
containerd c4bccfa8d9cf3475eb1b5fb8f… calico-node
containerd 2a764736c33d81431bc5868ce… coredns
containerd 6809883de8dd4295c516036f0… coredns
containerd b409beb4993e90cfc3cd68ab1… etcd
containerd 939915fdae6ecdbd4ccb7bcaf… gadget
containerd f6f90a48036477a86c194f62a… kube-apiserver
containerd 3642ab7e5d4d6b01d58badbc1… kube-controller-manager
containerd ff066db87808e91b7bf53f720… kube-proxy
containerd 9429cafe24382fa458cecc998… kube-scheduler
```

```bash
$ sudo /inspektor-gadget/local-gadget list-containers --no-trunc
RUNTIME    ID                                                               NAME
containerd 4a673ed444eda3b2d2e3dcde971bc98f1e2589559a6fd97205ffc651f28bc34d calico-kube-contro…
containerd c4bccfa8d9cf3475eb1b5fb8f35b97009df28212f3476e413a5a00967fe8052d calico-node
containerd 2a764736c33d81431bc5868ce24ce7748705031c393b045f34b9106166de97d6 coredns
containerd 6809883de8dd4295c516036f0d0d0e21271aedf8299f3a2b2d6613872f023419 coredns
containerd b409beb4993e90cfc3cd68ab1e60bc03a8deb7603e769e0b12c9056091b25d0a etcd
containerd 939915fdae6ecdbd4ccb7bcaf0cbb304d9f3719f5c487c395b1e5494ce801374 gadget
containerd f6f90a48036477a86c194f62a7583cef26b78c0241ba461e633865254332e4f7 kube-apiserver
containerd 3642ab7e5d4d6b01d58badbc1f103015c314ffbd5bcb1e63075d29ca8985a6e0 kube-controller-ma…
containerd ff066db87808e91b7bf53f720d14184daeae24ae47eba9b4d23da3736958c0c0 kube-proxy
containerd 9429cafe24382fa458cecc998b456f3bd8008019e39b5faf47d000edcc60691a kube-scheduler
```

```bash
$ sudo /inspektor-gadget/local-gadget list-containers -o custom-columns=runtime,id,name,pid,mntns,netns
RUNTIME    ID              NAME                             PID     MNTNS         NETNS
containerd 4a673ed444eda3… calico-kube-controllers          12182   4026532516    4026532280
containerd c4bccfa8d9cf34… calico-node                      4084    4026532445    4026531992
containerd 6809883de8dd42… coredns                          3852    4026532443    4026532366
containerd 2a764736c33d81… coredns                          3827    4026532441    4026532299
containerd b409beb4993e90… etcd                             2063    4026532277    4026531992
containerd 939915fdae6ecd… gadget                           93865   4026532294    4026531992
containerd f6f90a48036477… kube-apiserver                   2051    4026532275    4026531992
containerd 3642ab7e5d4d6b… kube-controller-manager          2031    4026532273    4026531992
containerd ff066db87808e9… kube-proxy                       3063    4026532284    4026531992
containerd 9429cafe24382f… kube-scheduler                   1985    4026532271    4026531992
```

Wit longer container names:

```bash
$ sudo /inspektor-gadget/local-gadget list-containers
RUNTIME    ID                         NAME
docker     8b85b8351041adede50a3a912… k8s_POD_calico-node-kdjhq_kube-system_f7caa3c1-abdb-4346…
docker     ab4588e32e8cca2f09e553683… k8s_POD_gadget-zqdhq_gadget_40bce549-d313-481d-a35c-f853…
docker     05f548a2f35cd43d71edcd167… k8s_POD_kube-proxy-f8brs_kube-system_8b626311-a595-4b1e-…
docker     a319ce7589eed5d77633e6152… k8s_POD_test-pod-long-name-to-test_test-ns_f3fbd03d-db5b…
docker     8aa7813e8930eebd2a95cc932… k8s_POD_test-pod_test-ns_e4efcff3-5c39-4851-a054-658714f…
docker     a3afac4df1fdbe2e6de9892cc… k8s_calico-node_calico-node-kdjhq_kube-system_f7caa3c1-a…
docker     341444c3bfd1f647c1148bda0… k8s_gadget_gadget-zqdhq_gadget_40bce549-d313-481d-a35c-f…
docker     375da4232a34d75b9ea623a4c… k8s_kube-proxy_kube-proxy-f8brs_kube-system_8b626311-a59…
docker     e2de121169b1f7fe77f4ddad7… k8s_test-pod-long-name-to-test_test-pod-long-name-to-tes…
docker     7d392725f646ff82c9d4e24b5… k8s_test-pod_test-pod_test-ns_e4efcff3-5c39-4851-a054-65…
```

```bash
$ sudo /inspektor-gadget/local-gadget list-containers --no-trunc
RUNTIME    ID                                                               NAME
docker     8b85b8351041adede50a3a9128fe7f10f335ec1f99e51dc805463a9be3c13d62 k8s_POD_calico-nod…
docker     ab4588e32e8cca2f09e553683dce7fcfff2000375e4f70f9c32884273e761914 k8s_POD_gadget-zqd…
docker     05f548a2f35cd43d71edcd1679ae56d592b9c6ca3923bbe3addb70aa1654edbe k8s_POD_kube-proxy…
docker     a319ce7589eed5d77633e6152d2b1366e330360e6e578a3ad1841221399f877d k8s_POD_test-pod-l…
docker     8aa7813e8930eebd2a95cc932c08b0f992620006aae5f3285d681999d1fa72f4 k8s_POD_test-pod_t…
docker     a3afac4df1fdbe2e6de9892cc1b45db6d3c619cc90841dc4618ffac65eb80d91 k8s_calico-node_ca…
docker     341444c3bfd1f647c1148bda0b1db053e58961003c9518a9d1fa9c5022fe7803 k8s_gadget_gadget-…
docker     375da4232a34d75b9ea623a4cdedc4fe6877efc09b4f504ab8ee993e91e22dc7 k8s_kube-proxy_kub…
docker     e2de121169b1f7fe77f4ddad79e41abacc33c23bc0be60ebe6111ed8448ce592 k8s_test-pod-long-…
docker     7d392725f646ff82c9d4e24b546afc2c508c5902785c02e035b0589db7fa00af k8s_test-pod_test-…
```

```bash
$ sudo /inspektor-gadget/local-gadget list-containers -o custom-columns=runtime,id,name,pid,mntns,netns
RUNTIME    ID              NAME                             PID     MNTNS         NETNS
docker     8b85b8351041ad… k8s_POD_calico-node-kdjhq_kube-… 1882    4026532277    4026531992
docker     ab4588e32e8cca… k8s_POD_gadget-zqdhq_gadget_40b… 88993   4026532259    4026531992
docker     05f548a2f35cd4… k8s_POD_kube-proxy-f8brs_kube-s… 1630    4026532261    4026531992
docker     a319ce7589eed5… k8s_POD_test-pod-long-name-to-t… 155476  4026532367    4026532372
docker     8aa7813e8930ee… k8s_POD_test-pod_test-ns_e4efcf… 99907   4026532276    4026532296
docker     a3afac4df1fdbe… k8s_calico-node_calico-node-kdj… 2650    4026532281    4026531992
docker     341444c3bfd1f6… k8s_gadget_gadget-zqdhq_gadget_… 89044   4026532269    4026531992
docker     375da4232a34d7… k8s_kube-proxy_kube-proxy-f8brs… 1702    4026532265    4026531992
docker     e2de121169b1f7… k8s_test-pod-long-name-to-test_… 155649  4026532437    4026532372
docker     7d392725f646ff… k8s_test-pod_test-pod_test-ns_e… 100031  4026532364    4026532296
```


cmd/local-gadget/containers/containers.go
```diff
@@ -127,65 +92,30 @@ func NewListContainersCmd() *cobra.Command {
...
+		return nil, fmt.Errorf(`"id" column doesn't exist`)
+	}
+
+	// By default, the "id" column is configured with width=13 and no fixed. In
+	// this way, the auto-scale will do its magic when --no-trunc is not set.
+	// Instead, when it is set, we must ensure auto-scale doesn't truncate it.
+	if flags.noTrunc {
+		idCol.Width = 64
+		idCol.FixedWidth = true
 	}
```
blanquicet, line 118: @flyth What do you think about this approach? If you agree with this approach, maybe having an API `containerColumns.SetWidth` and `containerColumns.SetFixed` would be useful. Don't you think?

pkg/container-collection/containers.go
```diff
@@ -22,15 +22,20 @@ import (
...
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
+
+	"github.com/kinvolk/inspektor-gadget/pkg/columns"
 )
 
 // Container represents a container with its metadata.
 type Container struct {
+	// Container Runtime
+	Runtime string `json:"runtime,omitempty" column:"runtime,width:10,fixed" columnTags:"runtime"`
```
blanquicet, line 32: Once min/max feature is ready, it will become min=5 (cri-o) and max=10 (containerd).

cmd/local-gadget/containers/containers.go
flyth: So the result would be a still consistent table (headers and rows look uniformly) with the id field set to 64? But what about other fields that we don't know the maximum width of? They'd still get truncated, right?
Regarding the API: do you mean it should look like `containerColumns.SetWidth(columnName, width)` and `containerColumns.SetFixed(columnName, fixedBool)` and then mark the auto-scaler as dirty to recalculate?

Maybe instead of setting different widths for columns in case of `noTrunc`, we can use an option for the formatter telling it to use the maxWidths (fixed) instead of the normal widths. This doesn't solve the issue with the fields we don't know the limit of, but it avoids extra code.

flyth: Btw. - as you're printing a list there and all rows are known beforehand: you can call `formatter.AdjustWidthsToContent(entries, false, 0, false)` and it should adjust the widths automatically.

blanquicet: > So the result would be a still consistent table (headers and rows look uniformly) with the id field set to 64? 
>
Yes, it would be a consistent table, as shown in the PR description. 

> But what about other fields that we don't know the maximum width of? They'd still get truncated, right? 
>
Yes, they will still be truncated. The `--no-trunc` flag [applies only to the ID field](https://github.com/kinvolk/inspektor-gadget/blob/81a9ce5a2584e7bd17c8c8b1b7166f9cf084a1a6/cmd/local-gadget/containers/containers.go#L123-L128). It is the same as `docker ps --no-trunc`.

> Regarding the API: do you mean it should look like containerColumns.SetWidth(columnName, width) and containerColumns.SetFixed(columnName, fixedBool) and then mark the auto-scaler as dirty to recalculate?
> 
Is it mandatory to mark auto-scale as dirty? It is my point. If I modify those values manually and everything works fine (as it seems), then it is ok for me to export them and allow users to modify them. However, if it is needed to do any additional step, like mark it as dirty, to make the auto-scale feature work correctly, then I think we should have a helper.

> Maybe instead of setting different widths for columns in case of noTrunc, we can use an option for the formatter telling it to use the maxWidths (fixed) instead of the normal widths. This doesn't solve the issue with the fields we don't know the limit of, but it avoids extra code.
> 
I am unsure how it applies in this case. Maybe you thought about it when you didn't know we wanted to avoid truncating only one column.

blanquicet: > Btw. - as you're printing a list there and all rows are known beforehand: you can call `formatter.AdjustWidthsToContent(entries, false, 0, false)` and it should adjust the widths automatically.
>
I completely forgot about this feature. I will apply it.

flyth: > It is the same as docker ps --no-trunc.
Ah, I missed that part. You're right then a global "use maxWidth" won't apply.

> Is it mandatory to mark auto-scale as dirty?
Yes. Unless you change things before the autoscaler get's called for the first time (which happens when you output a header or call it manually). So either a SetDirty() or a couple of Setters for the formatter - I'm fine with both.

blanquicet: > > Is it mandatory to mark auto-scale as dirty?
> 
> Yes. Unless you change things before the autoscaler get's called for the first time (which happens when you output a header or call it manually). So either a SetDirty() or a couple of Setters for the formatter - I'm fine with both.
>
OK, the current code is organized perform these operations before calling `NewGadgetParser` (who calls `NewFormatter`). That's the reason it is working and should keep working. We can keep it like this for now.

On the other hand, I tried `AdjustWidthsToContent`, but let's resolve this discussion and start another.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -97,3 +103,27 @@ func (p *GadgetParser[T]) BuildColumnsHeader() string {
...
 }
+
+func (p *GadgetParser[T]) TransformToTable(entries []*T) string {
+	maxWidth := 0
+	if term.IsTerminal(int(os.Stdout.Fd())) {
+		terminalWidth, _, err := term.GetSize(0)
+		if err == nil {
+			maxWidth = terminalWidth
+		}
+	}
```
blanquicet, line 114: As we discussed, it would be great if we find a way to integrate this into the library. As you suggest, a helper like `GetTerminalWidth` could be enough for now.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -97,3 +103,27 @@ func (p *GadgetParser[T]) BuildColumnsHeader() string {
...
+
+	// Disable auto-scaling as AdjustWidthsToContent will already manage the
+	// screen size with the maxWidth we are passing.
+	// p.formatter.SetAutoScaling(false)
```
blanquicet, line 118: Given the current structure of the code, I think it is necessary to have this helper. Otherwise, each gadget that is creating the parser needs to specify when calling `NewGadgetParser` whether it will print the full table or entry-by-entry so that we initialize the formatter with or without autoscaling. I would like to avoid that extra code, keep hiding such distinction from the gadget and manage all this in the parser.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -97,3 +103,27 @@ func (p *GadgetParser[T]) BuildColumnsHeader() string {
...
+
+	p.formatter.AdjustWidthsToContent(entries, true, maxWidth, true)
+
+	var sb strings.Builder
+	p.formatter.WriteTable(&sb, entries)
+	return sb.String()
```
blanquicet, line 124: I think `WriteTable` should follow the same logic as `FormatEntry` and `FormatHeader`. I meant, it should be `FormatTable` and should not print the table.

Notice also that `FormatTable` should not add a new line at the end of the table, like `FormatEntry` and `FormatHeader`.

pkg/container-collection/containers.go
blanquicet: Done.

cmd/common/utils/parser-tableformatter.go
```diff
@@ -43,6 +47,7 @@ type GadgetParserOptions struct {
 // using the columns and formatter/textcolumns packages.
 type GadgetParser[T any] struct {
```
mauriciovasquezbernal, line 48: Do you think `GadgetParser` is something we'll continue using in the future? I thought it was something temporal until the migration to use the columns library was completed on all gadgets, but now I see you're introducing new features to it. 

Is it something we really need? IIUC it's a wrapper/adapter around the columns library. I'm asking because I think many of the logic on this could go directly to the columns library, for instance, the `TransformToTable` function you implemented here.

pkg/container-collection/containers.go
```diff
@@ -22,15 +22,20 @@ import (
...
+
 	// ID is the container id, typically a 64 hexadecimal string
-	ID string `json:"id,omitempty"`
+	ID string `json:"id,omitempty" column:"id,width:13" columnTags:"runtime"`
 
 	// Pid is the process id of the container
-	Pid uint32 `json:"pid,omitempty"`
+	Pid uint32 `json:"pid,omitempty" column:"pid,width:7,fixed,hide"`
```
mauriciovasquezbernal, line 38:
```suggestion
	Pid uint32 `json:"pid,omitempty" column:"pid,minWidth:7,hide"`
```

pkg/container-collection/containers.go
```diff
@@ -22,15 +22,20 @@ import (
...
 )
 
 // Container represents a container with its metadata.
 type Container struct {
+	// Container Runtime
+	Runtime string `json:"runtime,omitempty" column:"runtime,width:10,fixed" columnTags:"runtime"`
+
 	// ID is the container id, typically a 64 hexadecimal string
-	ID string `json:"id,omitempty"`
+	ID string `json:"id,omitempty" column:"id,width:13" columnTags:"runtime"`
```
mauriciovasquezbernal, line 35:
```suggestion
	ID string `json:"id,omitempty" column:"id,width:13,maxWidth:64" columnTags:"runtime"`
```

mauriciovasquezbernal (commented):
Generally speaking LGTM. I'm wondering about the future of the "GadgetParser".

cmd/common/utils/parser-tableformatter.go
blanquicet: Solve using https://github.com/kinvolk/inspektor-gadget/pull/993.

cmd/common/utils/parser-tableformatter.go
blanquicet: Solve using https://github.com/kinvolk/inspektor-gadget/pull/993.

cmd/common/utils/parser-tableformatter.go
blanquicet: Solve using https://github.com/kinvolk/inspektor-gadget/pull/993.

blanquicet:
Rebased on top of https://github.com/kinvolk/inspektor-gadget/pull/993.

cmd/local-gadget/containers/containers.go
```diff
@@ -127,65 +82,30 @@ func NewListContainersCmd() *cobra.Command {
...
+	if !ok {
+		return nil, fmt.Errorf(`"id" column doesn't exist`)
+	}
+
+	// By default, the "id" column is configured with width=13 and no fixed. In
+	// this way, the auto-scale will do its magic when --no-trunc is not set.
+	// Instead, when it is set, we must ensure auto-scale doesn't truncate it.
+	if flags.noTrunc {
+		idCol.Width = 64
+		idCol.FixedWidth = true
```
flyth, line 107: This isn't needed anymore now that the table is adjusted by content, right? But maybe use the noTrunc as parameter to the AdjustByContent() to negate the `force` parameter, so that when the terminal is too small, it would output everything nevertheless.

pkg/container-collection/containers.go
```diff
@@ -22,15 +22,20 @@ import (
...
+
+	"github.com/kinvolk/inspektor-gadget/pkg/columns"
 )
 
 // Container represents a container with its metadata.
 type Container struct {
+	// Container Runtime
+	Runtime string `json:"runtime,omitempty" column:"runtime,minWidth:5,maxWidth:10" columnTags:"runtime"`
+
 	// ID is the container id, typically a 64 hexadecimal string
```
flyth, line 34: These could also be added as `columnDesc` later on.

flyth (approved):
Looks good to me as well! Thanks!

pkg/container-collection/containers.go
blanquicet: Agree. I think that's one of the next features we need to work on: Show the available (and description) columns in `--help`.

mauriciovasquezbernal: Issue for ^ https://github.com/kinvolk/inspektor-gadget/issues/998.

cmd/common/utils/parser-tableformatter.go
blanquicet: > Do you think `GadgetParser` is something we'll continue using in the future? 
>
I think we will. In particular, I like the idea of having this bridge that, on the one hand, avoids gadgets dealing with different packages: columns, textcolumns, sort, group, etc. And, on the other hand, it keeps those packages generic. Please notice that at some point, I expect to reach this:

```golang
func (p *GadgetParser[T]) Transform(entries ...*T) string {
	numEntries := len(entries)

	switch p.outputConfig.OutputMode {
	case OutputModeJSON:
		var b []byte
		var err error
		if numEntries == 1 {
			b, err = json.Marshal(entries[0])
		} else {
			b, err = json.MarshalIndent(entries, "", "  ")
		}

		if err != nil {
			fmt.Fprint(os.Stderr, fmt.Sprint(WrapInErrMarshalOutput(err)))
			return ""
		}

		return string(b)
	case OutputModeColumns:
		fallthrough
	case OutputModeCustomColumns:
		if numEntries == 1 {
			return p.formatter.FormatEntry(entries[0])
		} else {
			// Disable auto-scaling as AdjustWidthsToContent will already manage the
			// screen size.
			p.formatter.SetAutoScale(false)
			p.formatter.AdjustWidthsToContent(entries, true, textcolumns.GetTerminalWidth(), true)
			return p.formatter.FormatTable(entries)
		}
	default:
		fmt.Fprint(os.Stderr, fmt.Sprint(WrapInErrOutputModeNotSupported(p.outputConfig.OutputMode)))
		return ""
	}
}
```

> Is it something we really need? IIUC it's a wrapper/adapter around the columns library. I'm asking because I think many of the logic on this could go directly to the columns library, for instance, the TransformToTable function you implemented here.
>
As the previous example of a generic `Transform()`, I think the logic that involves the CLI flags can't be moved into the library. For instance, check my implementation proposal for adding `--k8s-metadata` flag. WDYT?

mauriciovasquezbernal: Right. I think it's fine to have this small "glue" code there.

mauriciovasquezbernal (approved):
LGTM!

PR #983, opened 2022-09-22
pkg/columns: add support for minWidth/maxWidth to help with auto-scaling

flyth:
This PR introduces two new parameters to columns, `minWidth` and `maxWidth`. Whenever the auto-scaler is in use, it will prevent columns from shrinking to less than minWidth or expanding to more than maxWidth. This is helpful when you know beforehand that a field will not cross those boundaries and instead give the remaining space to other columns.

This PR also introduces a new special value for `width`, `minWidth` and `maxWidth` called `type`. This can be used on columns that have an underlying integer value and will automatically use the maximum space that can be occupied for the specific type - e.g. a uint8 will result in a width of 3 (as 255 is the maximum value, holding a maximum of three digits).

I also added some leftover cleanups (mostly regarding the tests) from the initial PR - if it's too much, let me know and I'll move it to a separate PR.

### DIFF
--- a/pkg/columns/columninfo.go
+++ b/pkg/columns/columninfo.go
@@ -23,9 +23,23 @@ import (
 	"github.com/kinvolk/inspektor-gadget/pkg/columns/ellipsis"
 )
 
+const (
+	MaxCharsUint8  = 3  // 255
+	MaxCharsInt8   = 4  // -128
+	MaxCharsUint16 = 5  // 65535
+	MaxCharsInt16  = 6  // -32768
+	MaxCharsUint32 = 10 // 4294967295
+	MaxCharsInt32  = 11 // -2147483648
+	MaxCharsUint64 = 20 // 18446744073709551615
+	MaxCharsInt64  = 20 // −9223372036854775808
+	MaxCharsBool   = 5  // false
+)
+
 type Column[T any] struct {
 	Name         string                // Name of the column; case-insensitive for most use cases
 	Width        int                   // Width to reserve for this column
+	MinWidth     int                   // MinWidth will be the minimum width this column will be scaled to when using auto-scaling
+	MaxWidth     int                   // MaxWidth will be the maximum width this column will be scaled to when using auto-scaling
 	Alignment    Alignment             // Alignment of this column (left or right)
 	Extractor    func(*T) string       // Extractor to be used; this can be defined to transform the output before retrieving the actual value
 	Visible      bool                  // Visible defines whether a column is to be shown by default
@@ -43,11 +57,57 @@ type Column[T any] struct {
 	columnType    reflect.Type // cached type info from reflection
 }
 
+func (ci *Column[T]) getWidthFromType() int {
+	switch ci.kind {
+	case reflect.Uint8:
+		return MaxCharsUint8
+	case reflect.Int8:
+		return MaxCharsInt8
+	case reflect.Uint16:
+		return MaxCharsUint16
+	case reflect.Int16:
+		return MaxCharsInt16
+	case reflect.Uint32:
+		return MaxCharsUint32
+	case reflect.Int32:
+		return MaxCharsInt32
+	case reflect.Uint64, reflect.Uint:
+		return MaxCharsUint64
+	case reflect.Int64, reflect.Int:
+		return MaxCharsInt64
+	case reflect.Bool:
+		return MaxCharsBool
+	}
+	return 0
+}
+
+func (ci *Column[T]) getWidth(params []string) (int, error) {
+	if len(params) == 1 {
+		return 0, fmt.Errorf("missing %q value for field %q", params[0], ci.Name)
+	}
+	if params[1] == "type" {
+		// Special case, we get the maximum length this field can have by its type
+		w := ci.getWidthFromType()
+		if w > 0 {
+			return w, nil
+		}
+		return 0, fmt.Errorf("special value %q used for field %q is only available for integer and bool types", params[1], ci.Name)
+	}
+
+	res, err := strconv.Atoi(params[1])
+	if err != nil {
+		return 0, fmt.Errorf("invalid width %q for field %q: %w", params[1], ci.Name, err)
+	}
+
+	return res, nil
+}
+
 func (ci *Column[T]) fromTag(tag string) error {
 	tagInfo := strings.Split(tag, ",")
 	ci.Name = tagInfo[0]
 
 	tagInfo = tagInfo[1:]
+	var err error
 	for _, subTag := range tagInfo {
 		params := strings.SplitN(subTag, ":", 2)
 		paramsLen := len(params)
@@ -129,14 +189,20 @@ func (ci *Column[T]) fromTag(tag string) error {
 			}
 			ci.Precision = w
 		case "width":
-			if paramsLen == 1 {
-				return fmt.Errorf("missing width value for field %q", ci.Name)
-			}
-			w, err := strconv.Atoi(params[1])
+			ci.Width, err = ci.getWidth(params)
 			if err != nil {
-				return fmt.Errorf("invalid width %q for field %q: %w", params[1], ci.Name, err)
+				return err
 			}
-			ci.Width = w
+		case "maxWidth":
+			ci.MaxWidth, err = ci.getWidth(params)
+			if err != nil {
+				return err
+			}
+		case "minWidth":
+			ci.MinWidth, err = ci.getWidth(params)
+			if err != nil {
+				return err
+			}
 		default:
 			return fmt.Errorf("invalid column parameter %q for field %q", params[0], ci.Name)
 		}
--- a/pkg/columns/columninfo_test.go
+++ b/pkg/columns/columninfo_test.go
@@ -15,34 +15,63 @@
 package columns
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/kinvolk/inspektor-gadget/pkg/columns/ellipsis"
 )
 
+func expectColumnsSuccess[T any](t *testing.T, options ...Option) *Columns[T] {
+	cols, err := NewColumns[T](options...)
+	if err != nil {
+		t.Fatalf("Failed to initialize: %v", err)
+	}
+	return cols
+}
+
+func expectColumnsFail[T any](t *testing.T, name string, options ...Option) {
+	t.Run(name, func(t *testing.T) {
+		_, err := NewColumns[T](options...)
+		if err == nil {
+			t.Errorf("Succeeded to initialize but expected error")
+		}
+	})
+}
+
+func expectColumn[T any](t *testing.T, cols *Columns[T], columnName string) *Column[T] {
+	col, ok := cols.GetColumn(columnName)
+	if !ok {
+		t.Fatalf("Expected column with name %q", columnName)
+	}
+	return col
+}
+
+func expectColumnValue[T any](t *testing.T, col *Column[T], fieldName string, expectedValue interface{}) {
+	columnValue := reflect.ValueOf(col).Elem()
+	fieldValue := columnValue.FieldByName(fieldName)
+	if !fieldValue.IsValid() {
+		t.Errorf("Expected field %q", fieldName)
+		return
+	}
+	if fieldValue.Interface() != expectedValue {
+		t.Errorf("Expected field %q to equal %+v, got %+v", fieldName, expectedValue, fieldValue.Interface())
+	}
+}
+
 func TestColumnsInvalid(t *testing.T) {
 	type testFail1 struct {
-		Unknown string `column:"left,unknown"` // unknown parameter
+		Unknown string `column:"left,unknown"`
 	}
 	type testFail2 struct {
 		Unknown1 string `column:"unknown"`
-		Unknown2 string `column:"unknown"` // double name
+		Unknown2 string `column:"unknown"`
 	}
 	type testFail3 struct {
 		testFail2
 	}
-	_, err := NewColumns[testFail1]()
-	if err == nil {
-		t.Errorf("succeeded to initialize but expected error")
-	}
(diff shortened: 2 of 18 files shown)


pkg/columns/columns.go
```diff
@@ -198,6 +198,15 @@ func (c *Columns[T]) iterateFields(t reflect.Type, sub []int) error {
...
 		}
 
+		// fall back to struct field name if column name is empty
+		if column.Name == "" {
+			column.Name = f.Name
+		}
+
+		if column.MinWidth > column.Width {
+			column.Width = column.MinWidth
+		}
```
blanquicet, line 208: I would prefer this to return an error.

pkg/columns/columns.go
```diff
@@ -198,6 +198,15 @@ func (c *Columns[T]) iterateFields(t reflect.Type, sub []int) error {
 			return fmt.Errorf("error parsing tag for %q on field %q: %w", t.Name(), f.Name, err)
 		}
 
+		// fall back to struct field name if column name is empty
+		if column.Name == "" {
+			column.Name = f.Name
+		}
```
blanquicet, line 204: When could this happen?

pkg/columns/columninfo.go
```diff
@@ -43,11 +56,47 @@ type Column[T any] struct {
 	columnType    reflect.Type // cached type info from reflection
 }
 
+func (ci *Column[T]) getWidth(params []string) (int, error) {
+	if len(params) == 1 {
+		return 0, fmt.Errorf("missing %q value for field %q", params[0], ci.Name)
+	}
+	if params[1] == "type" {
```
blanquicet, line 63: What about setting `MaxWidth` by default to "type" if it is an integer?

blanquicet (commented):
A couple of comments after a quick check

pkg/columns/columns.go
flyth: This also applies when no width was set (default being used), but I can add an error specifically for the case that both minWidth and width have been set and then minWidth is > width.

pkg/columns/columns.go
flyth: If you define `Field string column:",width:20"` for example.  In this case the name then would become `Field`.

blanquicet: Mmmm not sure, I would prefer to return an error also here.

pkg/columns/columninfo.go
flyth: You mean when defining like so: `column:"intField,width,maxWidth,minWidth"`?

pkg/columns/columns.go
flyth: It's the same way that `encoding/json` handles that case - you can specify `json:",omitempty"` and it'll do the same.

blanquicet: And what happens if more than one field doesn't specify the name?

flyth: They all get their respective field names - if you used that field name before for another field, it'll return the "duplicate name" error.

pkg/columns/columninfo.go
blanquicet: No, I meant that if the user doesn't explicitly specify the `maxWtidth` in the `column` tag of an `int` field, then we automatically set `maxWtidth` to the max for that type of int.

flyth: Hmm, I think it's a good to set it as default, but how would you go back to scaling "naturally" for those fields, then - in case you don't want to use the `type` defaults?

blanquicet: Is there any case where someone won't want `types` defaults? I mean, if the user is using auto-scaling, they want to take advantage of all the screen to get those long fields displayed almost completely, I don't think they want to "waste" the space for fields that we are sure will not be longer than the max we are setting by default. Don't you think? The only case I can see that they won't want this default is, maaaybe for "styling" reasons 😅.

flyth: Hehe, _I_ wouldn't to that as well, but yeah, "styling" reasons. Maybe we should just go with it and re-add it somehow if it's really needed.

blanquicet: Anyway, I was thinking that for the IG use-case, we will set the `minWidth=types` for all the `int` fields, but there could be other use cases where it won't be like that. So, I am fine wither way 🙂

pkg/columns/columninfo.go
```diff
@@ -23,9 +23,22 @@ import (
...
 
+const (
+	MaxCharsUint8  = 3  // 255
+	MaxCharsInt8   = 4  // -128
+	MaxCharsUint16 = 5  // 65535
+	MaxCharsInt16  = 6  // -32768
+	MaxCharsUint32 = 10 // 4294967295
+	MaxCharsInt32  = 11 // -2147483648
+	MaxCharsUint64 = 20 // 18446744073709551615
+	MaxCharsInt64  = 20 // −9223372036854775808
```
mauriciovasquezbernal, line 34: Add booleans as well?

mauriciovasquezbernal:
Hey @flyth, would you mind squashing the commits where you handle the feedback? Otherwise this PR is very difficult to review as it includes changes not related.

flyth:
> Hey @flyth, would you mind squashing the commits where you handle the feedback? Otherwise this PR is very difficult to review as it includes changes not related.

I've rearranged the commits - I hope it's easier to read now.

pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -58,46 +62,88 @@ func (tf *TextColumnsFormatter[T]) RecalculateWidths(maxWidth int, force bool) {
...
 
 	// Get total width of all printed columns
-	totalWidth := 0
-	spaces := 0
+	totalWidthNotFixed := 0
+	totalWidthFixed := 0
 	for i, column := range tf.showColumns {
 		if i > 0 {
-			spaces += len([]rune(tf.options.ColumnDivider))
+			totalWidthFixed += len([]rune(tf.options.ColumnDivider))
```
mauriciovasquezbernal, line 69: Why do you have to use `len([]rune(tf.options.ColumnDivider))` instead of `len(tf.options.ColumnDivider)`?

pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -58,46 +62,88 @@ func (tf *TextColumnsFormatter[T]) RecalculateWidths(maxWidth int, force bool) {
...
 		}
```
mauriciovasquezbernal, line 70: Maybe you can initialize `totalWidthFixed := (len(tf.showColumns) - 1) * len([]rune(tf.options.ColumnDivider))` to avoid having this conditional on the loop?

pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -58,46 +62,88 @@ func (tf *TextColumnsFormatter[T]) RecalculateWidths(maxWidth int, force bool) {
...
+			if !force {
+				if column.col.MaxWidth > 0 && column.calculatedWidth > column.col.MaxWidth {
+					column.calculatedWidth = column.col.MaxWidth
+					column.treatAsFixed = true
+					satisfied = false
+
+					addToFixed += column.calculatedWidth
+					removeFromNotFixed += column.col.Width
+					continue
+				} else if column.col.MinWidth > 0 && column.calculatedWidth < column.col.MinWidth {
```
mauriciovasquezbernal, line 115: nit: else not needed because of "continue" above.

mauriciovasquezbernal (approved):
Some minor comments but generally LGTM. I didn't review the whole algorithm in `RecalculateWidths` as I find it difficult to follow. I see it has good tests and by trying the whole thing with kubectl gadget I see the result is the expected.

pkg/columns/formatter/textcolumns/scaler.go
flyth: We have to use []rune, because the string ColumnDivider could contain UTF encoded characters that would count as more characters than actually printed.

pkg/columns/columninfo_test.go
```diff
@@ -15,93 +15,91 @@
...
 	"testing"
 
 	"github.com/kinvolk/inspektor-gadget/pkg/columns/ellipsis"
 )
 
+func initializeColumns[T any](t *testing.T, expectSuccess bool, name string, options ...Option) *Columns[T] {
+	cols, err := NewColumns[T](options...)
+	t.Run(name, func(t *testing.T) {
+		if err != nil && expectSuccess {
+			t.Fatalf("failed to initialize: %v", err)
```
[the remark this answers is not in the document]
flyth: Ah, I missed that discussion. Changed it!

pkg/columns/columninfo.go
```diff
@@ -43,11 +57,57 @@ type Column[T any] struct {
...
+		return MaxCharsUint32
+	case reflect.Int32:
+		return MaxCharsInt32
+	case reflect.Uint64, reflect.Uint:
+		return MaxCharsUint64
+	case reflect.Int64, reflect.Int:
+		return MaxCharsInt64
+	case reflect.Bool:
+		return MaxCharsBool
+	}
```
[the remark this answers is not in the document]
flyth: You mean change it for readability? Since this is a non-hot code path, I'm happy to do it if it helps - generally (but especially for hotter code) I would prefer the switch as it is much faster.

pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -58,46 +62,88 @@ func (tf *TextColumnsFormatter[T]) RecalculateWidths(maxWidth int, force bool) {
...
+
+	for {
+		satisfied := true
+		addToFixed := 0
+		removeFromNotFixed := 0
+
+		occurrences = make(map[string]int)
+
+		totalAdjustedWidthNotFixed = 0
+		for _, column := range tf.showColumns {
```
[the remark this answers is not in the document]
flyth: I added a bunch of comments to clarify things.

pkg/columns/formatter/textcolumns/scaler.go
mauriciovasquezbernal: TIL. Thanks!

pkg/gadgets/trace/exec/types/types.go
```diff
@@ -24,12 +24,12 @@ import (
 type Event struct {
 	eventtypes.Event
 
-	Pid       uint32   `json:"pid,omitempty" column:"pid,width:7,fixed"`
-	Ppid      uint32   `json:"ppid,omitempty" column:"ppid,width:7,fixed"`
+	Pid       uint32   `json:"pid,omitempty" column:"pid,minWidth:7"`
+	Ppid      uint32   `json:"ppid,omitempty" column:"ppid,minWidth:7"`
 	Comm      string   `json:"pcomm,omitempty" column:"comm,width:16,fixed"`
```
blanquicet, line 29: As discussed [here](https://github.com/kinvolk/inspektor-gadget/pull/934#discussion_r977458270), I think COMM should be: `column:"comm,maxWidth:16"`. We can decide later whether we should add a `minWidth` or not.

pkg/gadgets/trace/bind/types/types.go
```diff
@@ -22,11 +22,11 @@ import (
 type Event struct {
 	eventtypes.Event
 
-	Pid       uint32 `json:"pid,omitempty" column:"pid,width:7,fixed"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,minWidth:7"`
 	Comm      string `json:"comm,omitempty" column:"comm,width:16,fixed"`
```
blanquicet, line 26: dito

pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -32,81 +32,161 @@ func (tf *TextColumnsFormatter[T]) RecalculateWidths(maxWidth int, force bool) {
...
 		for _, column := range tf.showColumns {
-			if column.col.FixedWidth && !force {
+			if (column.col.FixedWidth || column.treatAsFixed) && !force {
 				continue
 			}
-			if occurrences[column.col.Name] > 1 {
+
+			// we can only distribute to columns that are used more than once if we have leftover space for all
+			// occurrences
+			if occ := occurrences[column.col.Name]; occ > 1 {
```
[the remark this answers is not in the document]
flyth: Yes, it's possible - we give defaults (with no duplicate columns of course), but in the end the user can decide what to print. That's why I think we should be generic enough to handle it.

pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -32,81 +32,161 @@ func (tf *TextColumnsFormatter[T]) RecalculateWidths(maxWidth int, force bool) {
...
+
+					if leftover == 0 {
+						break
+					}
+
+					alreadySpent[column.col.Name] = struct{}{}
+					continue
+				}
 				// cannot redistribute here, since it would be used more than just once
 				continue
```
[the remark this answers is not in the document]
flyth: Not exactly: the columns and their calculated widths are only referenced using pointers - so if a column gets used twice, we need to make sure we have two leftover characters, since increasing the width by 1 for that column will also affect the output twice. And to make sure that in a pass a column doesn't get double that amount (but is distributed evenly), we mark that column.

flyth:
> Is my understanding correct?

That sounds about right for 1. and 2. Regarding columns being used more than once, please see the comment.

pkg/columns/columninfo_test.go
```diff
@@ -15,93 +15,91 @@
 package columns
```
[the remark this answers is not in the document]
flyth: As discussed, I will take this into account for the next PR.

flyth:
Thanks for the reviews everyone!

PR #986, opened 2022-09-22
integration: Add workaround for trace capabilities test on ARO

mauriciovasquezbernal:
Use --audit-only=false to avoid issue on ARO.

Ref https://github.com/kinvolk/inspektor-gadget/issues/985

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -432,8 +432,10 @@ func TestCapabilities(t *testing.T) {
 	t.Parallel()
 
 	capabilitiesCmd := &command{
-		name:           "StartCapabilitiesGadget",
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
You're right, that problem is not fixed by this. I tried to run this 10 times on an ARO cluster and worked fine. What do you think about merging this and checking if we face further problems later on?

blanquicet (approved):
LGTM

blanquicet:
Merging this PR as initial mitigation for https://github.com/kinvolk/inspektor-gadget/issues/985. Also because we need to verify if ARO is having problems deleting the SPO (AKS seems to have), after merging https://github.com/kinvolk/inspektor-gadget/pull/970.

PR #987, opened 2022-09-23
Remove seccomp architecture filter

eiffel-fl:
Hi.

Following this [discussion](https://kubernetes.slack.com/archives/CSYL75LF6/p1663926363524629?thread_ts=1663923900.580219&cid=CSYL75LF6), I wrote some patches to set the seccomp arch according to the underlying architecture.

Best regards.

### DIFF
--- a/docs/guides/audit/seccomp.md
+++ b/docs/guides/audit/seccomp.md
@@ -30,8 +30,6 @@ metadata:
     description: "Log some syscalls"
 spec:
   defaultAction: SCMP_ACT_ALLOW
-  architectures:
-  - SCMP_ARCH_X86_64
   syscalls:
   - action: SCMP_ACT_KILL
     names:
@@ -91,9 +89,6 @@ default          mypod            unshare          kill_thread
 ```json
 {
   "defaultAction": "SCMP_ACT_ALLOW",
-  "architectures": [
-    "SCMP_ARCH_X86_64"
-  ],
   "syscalls": [
     {
       "action": "SCMP_ACT_KILL",


docs/guides/audit/seccomp.md
```diff
@@ -41,6 +41,9 @@ spec:
     - mkdir
 ```
 
+Note that, on `arm64` based computer, you will need to replace
+`SCMP_ARCH_X86_64` by `SCMP_ARCH_AARCH64`.
```
alban, line 45: The `architectures` list is optional according to runtime-spec:
https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#seccomp

Man seccomp_arch_add.3:
> When a seccomp filter is initialized with the call to seccomp_init(3) the native architecture is automatically added to the filter.

And runc only calls AddArch but never RemoveArch:
https://github.com/opencontainers/runc/blob/4ff59f81fded69a6d5f9e85d7eda5b23587db985/libcontainer/seccomp/seccomp_linux.go#L79-L88

So we could remove the `architectures` section from the yaml for simplicity.

Some of the examples from SPO don't have `architectures`:
https://github.com/kubernetes-sigs/security-profiles-operator/blob/de6b5a05bafbf86c65b6ef5b772d77ba4d6a72d0/examples/seccompprofile-allowed-syscalls-validation.yaml#L42-L49

integration/integration_test.go
```diff
@@ -276,6 +276,11 @@ func TestAuditSeccomp(t *testing.T) {
 
 	t.Parallel()
 
+	seccompArch, err := k8sArchToSeccompArch(*k8sArch)
```
alban, line 279: There are some arch conversions in [libseccomp.GetArchFromString()](https://github.com/seccomp/libseccomp-golang/blob/main/seccomp.go#L264-L266).

But it does not seem to be the ones you want. I wonder if the missing conversion function could be added in libseccomp-golang.

Or more easily, maybe just remove the `architectures` list completely, since it is optional?

PR #988, opened 2022-09-23
Allow terminating ebpf-objects Make target

johananl:
# Allow terminating ebpf-objects Make target

Hitting Ctrl+C doesn't work for `make ebpf-objects`. This fixes the problem by making the Docker command interactive.

## How to use

Run `make ebpf-objects`, hit Ctrl+C and ensure the command returns.

## Testing done

The above.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ all: build local-gadget
 phony_explicit:
 
 ebpf-objects:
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
LGTM. Thanks for the PR!

PR #990, opened 2022-09-23
Use columns library for trace dns gadget

mauriciovasquezbernal:


cmd/kubectl-gadget/trace/dns.go
```diff
@@ -15,95 +15,40 @@
...
-	"strings"
+	"github.com/spf13/cobra"
 
 	commontrace "github.com/kinvolk/inspektor-gadget/cmd/common/trace"
 	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
-	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/dns/types"
-
-	"github.com/spf13/cobra"
+	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/capabilities/types"
```
blanquicet, line 23: 😅

cmd/kubectl-gadget/trace/dns.go
```diff
@@ -15,95 +15,40 @@
...
+	dnsTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/dns/types"
 )
 
 type DNSParser struct {
 	commonutils.BaseParser[types.Event]
 }
```
blanquicet, line 29: Not needed anymore.

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -15,15 +15,25 @@
...
 
 type Event struct {
 	eventtypes.Event
 
-	DNSName string `json:"name,omitempty"`
-	PktType string `json:"pktType,omitempty"`
-	QType   string `json:"qtype,omitempty"`
+	PktType string `json:"pktType,omitempty" column:"type,width:9,fixed"`
+	QType   string `json:"qtype,omitempty" column:"qtype,width:10,fixed"`
+	DNSName string `json:"name,omitempty" column:"name,width:24"`
```
blanquicet, line 27: What do you think about using 30 (Same as k8s columns)? In this way, they will be autoscaled identically. Current situation:

```bash
$ ./kubectl-gadget trace dns -A
NODE                      NAMESPACE                 POD                       TYPE      QTYPE      NAME
worker                                                                        OUTGOING  A          registry-1.docker.i…
worker                                                                        HOST      A          registry-1.docker.i…
master                    kube-system               coredns-78fcd69978-rj79d  HOST      AAAA       www.inspektor-gadge…
```

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -15,15 +15,25 @@
...
+}
+
+func GetColumns() *columns.Columns[Event] {
+	cols := columns.MustCreateColumns[Event]()
+
+	col, _ := cols.GetColumn("container")
+	col.Visible = false
```
blanquicet, line 34: As I suggested [here](https://github.com/kinvolk/inspektor-gadget/pull/980#discussion_r977633141). I think adding the API to perform these operations would be useful: `ColumnMap.SetVisible()`.

integration/integration_test.go
```diff
@@ -457,16 +458,32 @@ func TestDns(t *testing.T) {
...
+
+			return expectAllToMatch(output, normalize, expectedEntry)
+		},
 	}
 
 	commands := []*command{
 		createTestNamespaceCommand(ns),
 		dnsCmd,
-		busyboxPodRepeatCommand(ns, "nslookup microsoft.com"),
+		busyboxPodRepeatCommand(ns, "nslookup -query=A inspektor-gadget.io"),
```
blanquicet, line 486: Why do you prefer generating only IPv4 DNS requests? We could use expectEntriesToMatch to verify `A` and `AAAA` requests are generated.

cmd/kubectl-gadget/trace/dns.go
mauriciovasquezbernal: I'd blame visual studio.

pkg/gadgets/trace/dns/types/dns.go
mauriciovasquezbernal: I agree. Let's give it more "weight".

pkg/gadgets/trace/dns/types/dns.go
mauriciovasquezbernal: I don't have a strong opinion. It seems too verbose to create all those setters, we can leave the member public.

integration/integration_test.go
mauriciovasquezbernal: Right. I didn't think about `expectEntriesToMatch`.

pkg/gadgets/trace/dns/types/dns.go
blanquicet: Yeah, let's keep the current implementation as also discussed [here](https://github.com/kinvolk/inspektor-gadget/pull/980#discussion_r981096998).

integration/integration_test.go
```diff
@@ -457,16 +458,42 @@ func TestDns(t *testing.T) {
...
+			// DNS gadget doesn't provide container data. Remove it.
+			for _, entry := range expectedEntries {
+				entry.Container = ""
+			}
+
+			normalize := func(e *dnsTypes.Event) {
+				e.Node = ""
+			}
+
+			return expectEntriesToMatch(output, normalize, expectedEntries...)
```
blanquicet, line 489: Out of scope for this PR: I don't remember if we ready discussed this, but do you think it would make sense to create another helper function where **all** the generated events must match one of the entries in `expectedEntries`? Right now, we stop at the first match.

blanquicet (approved):
I didn't test it again but from a code inspection LGTM!

integration/integration_test.go
mauriciovasquezbernal: I'm not sure I follow here. Do you mean that for each generated event there should be at least one match on `expectedEntries`?

mauriciovasquezbernal: Maybe it's useful, I'm not sure. I think we'll understand it when implementing more tests.

PR #993, opened 2022-09-27
pkg/columns/formatter/textcolumns: Add a couple of helper functions

flyth:
This PR adds a couple of helper functions discussed [in this PR](https://github.com/kinvolk/inspektor-gadget/pull/980).

* `FormatTable()` to return a full table as string
* `SetAutoScale()` to enable/disable auto-scaling on demand
* `GetTerminalWidth()` to easily get the current width of the terminal

This PR also prevents a potential crash when using AdjustWidthsToContent().

### DIFF
--- a/pkg/columns/formatter/textcolumns/scaler.go
+++ b/pkg/columns/formatter/textcolumns/scaler.go
@@ -307,6 +307,9 @@ func (tf *TextColumnsFormatter[T]) AdjustWidthsToContent(entries []*T, considerH
 		return
 	}
 
+	// Force RecalculateWidths() to run
+	tf.currentMaxWidth = -1
+
 	// We did our best, but let's resize to fit to maxWidth
 	tf.RecalculateWidths(maxWidth, force)
 }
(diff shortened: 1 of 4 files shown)


pkg/columns/formatter/textcolumns/scaler.go
```diff
@@ -297,6 +297,8 @@ func (tf *TextColumnsFormatter[T]) AdjustWidthsToContent(entries []*T, considerH
 		totalWidth += column.calculatedWidth
 	}
 
+	tf.buildFillString()
+
```
blanquicet, line 301: Could you please describe why it is needed?

flyth: Sure - it takes the longest possible string and builds a string with whitespace to copy from. This is used to quickly fill strings to a specific length without having to do a for loop every time. But whenever the calculatedWidths are adjusted, it needs to be rebuilt - and that was missing here.

blanquicet (approved):
Tested in https://github.com/kinvolk/inspektor-gadget/pull/980 and working as expected.

PR #994, opened 2022-09-27
examples: Use syscall.SIGINT instead of os.Interrupt

mauriciovasquezbernal:
Cosmetic change to only use syscall package for signals in the examples.


blanquicet (approved):
LGTM

PR #995, opened 2022-09-27
examples fix parser example

mauriciovasquezbernal:
- Use formatter directly 
- Rename folder from parser to formatter


examples/gadgets/parser/trace/exec/exec.go
```diff
@@ -90,20 +89,20 @@ func main() {
...
 	defer containerCollection.Close()
 
-	// Create a parser. It's the component that converts events to columns.
-	execParser, err := utils.NewGadgetParser(&utils.OutputConfig{}, types.GetColumns())
-	if err != nil {
-		fmt.Printf("failed to create parser: %s\n", err)
-		return
-	}
+	// Create a formatter. It's the component that converts events to columns.
+	colNames := []string{"container", "pid", "ppid", "pcomm", "ret", "args"}
```
blanquicet, line 93: `pcomm` -> `comm`. In fact, it is missing in the output.

mauriciovasquezbernal: We need to fix it in other places as well. The member name is comm but it's pcomm in json:

https://github.com/kinvolk/inspektor-gadget/blob/2bd017d5915820529ca34480fbdba5fd44b372d4/pkg/gadgets/trace/exec/types/types.go#L29

And we also have PCOMM on our guides -> https://github.com/kinvolk/inspektor-gadget/blob/2bd017d5915820529ca34480fbdba5fd44b372d4/docs/guides/trace/exec.md :(

examples/README.md
```diff
@@ -39,7 +39,7 @@ or `local-gadget` directly.
     packages to filter events by container.
     - [trace/exec](gadgets/withfilter/trace/exec/): traces creation of
       new processes inside a particular container.
-  - [parser](gadgets/parser/): Examples showing how to use a parser to
-    format the events.
-    - [trace/exec](gadgets/parser/trace/exec/): traces creation of
+  - [formatter](gadgets/formatter/): Examples showing how to use a formatter to
+    print events in a column format.
+    - [trace/exec](gadgets/formatter/trace/exec/): traces creation of
```
blanquicet, line 44: > Parser is a concept we won't be using anymore.
>
I am not completely sure, see https://github.com/kinvolk/inspektor-gadget/pull/980#discussion_r982391991. In any case, we think the best way to go is to use directly the formatter for the examples. WDYT?

mauriciovasquezbernal: I changed the commit message. I agree it's better to use the formatter on the examples.

PR #997, opened 2022-09-28
integration: Test minikube with docker driver on 2 nodes.

eiffel-fl:
Hi.

In this PR, I enabled integration tests on several nodes using minikube with `docker` driver.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -587,6 +587,7 @@ jobs:
     strategy:
       matrix:
         type: [default, core]
+        driver: [none, docker]
     steps:
     - uses: actions/checkout@v3
     - name: Setup go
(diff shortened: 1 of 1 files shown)


.github/workflows/inspektor-gadget.yml
```diff
@@ -600,6 +601,11 @@ jobs:
         minikube version: 'v1.25.2'
         kubernetes version: 'v1.23.0'
         github token: ${{ secrets.GITHUB_TOKEN }}
+        driver: ${{ matrix.driver }}
+        # When using docker driver, run the integration tests on two nodes.
+        # As GitHub Runner has only 2 CPU and 7 GB of DRAM, we cannot use a big
+        # number here.
+        start args: ${{ matrix.driver == 'docker' && '-n 2' || '' }}
```
mqasimsarfraz, line 608: I hope this isn't evaluating to `true` rather then '-n 2'. Did we try it already?

mqasimsarfraz (approved):
The changes are looking great! Concise and to the point!

.github/workflows/inspektor-gadget.yml
```diff
@@ -587,6 +587,7 @@ jobs:
     strategy:
       matrix:
         type: [default, core]
+        driver: [none, docker]
```
blanquicet, line 590: Does it mean that with this PR, we are now running the integration test in a multi-node minikube cluster (using docker driver) and also in a single-node cluster (using none driver)?

.github/workflows/inspektor-gadget.yml
```diff
@@ -587,6 +587,7 @@ jobs:
     strategy:
```
blanquicet, line 587: Great, it seems like the SPO works correctly in a 2 nodes cluster using just [one replica](https://github.com/kinvolk/inspektor-gadget/blob/7fe0cefd66c74511727b0565d1223b8f3e579032/integration/command.go#L96) ... Let's hope it keeps working like this 🤞🏽

PR #1005, opened 2022-10-03
integration: Re-enable liveness probe on ARO

mauriciovasquezbernal:
Revert b9db9360d8b8 ("integration: Disable liveness probes when running on ARO") given that https://github.com/kinvolk/inspektor-gadget/issues/799 is now fixed.


blanquicet (approved):
LGTM!

Issue #1003, opened 2022-10-03
[EPIC] Use columns library in gadgets


mauriciovasquezbernal:
A new library to format the output data in an easier way was implemented in https://github.com/kinvolk/inspektor-gadget/pull/877. However, it's not being used in all gadgets yet. The purpose of this issue is to port all gadgets to use this library. https://github.com/kinvolk/inspektor-gadget/pull/989 implemented this for the "trace capabilities" gadget. 
 
## Heads up 

A gadget needs to use JSON for the integration tests before porting to use the columns library -> https://github.com/kinvolk/inspektor-gadget/issues/719. 

## PR tracker:
- advise: Doesn't apply
- audit:
	- [x] seccomp: https://github.com/inspektor-gadget/inspektor-gadget/pull/1133
- profile:
	- [x] block-io Outputs a graph. Doesn't apply
	- [x] cpu https://github.com/inspektor-gadget/inspektor-gadget/pull/1128
- snapshot:
	- [x] process https://github.com/inspektor-gadget/inspektor-gadget/pull/1123
	- [x] socket https://github.com/inspektor-gadget/inspektor-gadget/pull/1123
- top:
	- [ ] block-io https://github.com/inspektor-gadget/inspektor-gadget/pull/1138
	- [ ] ebpf https://github.com/inspektor-gadget/inspektor-gadget/pull/1138
	- [ ] file https://github.com/inspektor-gadget/inspektor-gadget/pull/1138
	- [ ] tcp https://github.com/inspektor-gadget/inspektor-gadget/pull/1138
- trace:
	- [x] bind (https://github.com/kinvolk/inspektor-gadget/pull/934)
	- [x] capabilities (https://github.com/kinvolk/inspektor-gadget/pull/989)
	- [x] dns (https://github.com/kinvolk/inspektor-gadget/pull/990)
	- [x] exec (https://github.com/kinvolk/inspektor-gadget/pull/934)
	- [x] Fsslower https://github.com/kinvolk/inspektor-gadget/pull/1017
	- [x] Mount https://github.com/kinvolk/inspektor-gadget/pull/1009
	- [x] oomkill (https://github.com/kinvolk/inspektor-gadget/pull/1027)
	- [x] open (https://github.com/kinvolk/inspektor-gadget/pull/1030)
	- [x] network (https://github.com/inspektor-gadget/inspektor-gadget/pull/1042)
[...]

PR #1009, opened 2022-10-05
trace/mount: Use columns library

burak-ok:
# trace/mount: Use columns library

The trace mount gadget now uses the columns library

## Testing done

Columnwidths are not the same as the old one. Currently it looks like this:
image

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -31,6 +31,7 @@ import (
 	capabilitiesTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/capabilities/types"
 	dnsTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/dns/types"
 	execTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/exec/types"
+	mountTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/mount/types"
 	openTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/open/types"
 	signalTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/signal/types"
 	tcpTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/tcp/types"
@@ -629,10 +630,33 @@ func TestMountsnoop(t *testing.T) {
 	t.Parallel()
 
 	mountsnoopCmd := &Command{
-		Name:           "StartMountsnoopGadget",
-		Cmd:            fmt.Sprintf("$KUBECTL_GADGET trace mount -n %s", ns),
-		ExpectedRegexp: fmt.Sprintf(`%s\s+test-pod\s+test-pod+\s+mount\s+\d+\s+\d+\s+\d+\s+mount\("/mnt", "/mnt", .*\) = -2`, ns),
-		StartAndStop:   true,
+		Name:         "StartMountsnoopGadget",
+		Cmd:          fmt.Sprintf("$KUBECTL_GADGET trace mount -n %s -o json", ns),
+		StartAndStop: true,
+		ExpectedOutputFn: func(output string) error {
+			expectedEntry := &mountTypes.Event{
+				Event:     BuildBaseEvent(ns),
+				Comm:      "mount",
+				Operation: "mount",
+				Retval:    -2,
+				Source:    "/mnt",
+				Target:    "/mnt",
+			}
+
+			normalize := func(e *mountTypes.Event) {
+				e.Node = ""
+				e.Pid = 0
+				e.Tid = 0
+				e.MountNsID = 0
+				e.Latency = 0
+				e.Fs = ""
+				e.Data = ""
+				e.Flags = nil
+				e.FlagsRaw = 0
+			}
+
+			return ExpectEntriesToMatch(output, normalize, expectedEntry)
+		},
 	}
 
 	commands := []*Command{
--- a/pkg/gadgets/trace/mount/types/types.go
+++ b/pkg/gadgets/trace/mount/types/types.go
@@ -15,25 +15,59 @@
 package types
 
 import (
+	"fmt"
+	"strings"
+
+	"github.com/kinvolk/inspektor-gadget/pkg/columns"
 	eventtypes "github.com/kinvolk/inspektor-gadget/pkg/types"
 )
 
 type Event struct {
 	eventtypes.Event
 
-	MountNsID uint64   `json:"mntnsid,omitempty"`
-	Pid       uint32   `json:"pid,omitempty"`
-	Tid       uint32   `json:"tid,omitempty"`
-	Comm      string   `json:"comm,omitempty"`
-	Operation string   `json:"operation,omitempty"`
-	Retval    int      `json:"ret,omitempty"`
-	Latency   uint64   `json:"latency,omitempty"`
-	Fs        string   `json:"fs,omitempty"`
-	Source    string   `json:"source,omitempty"`
-	Target    string   `json:"target,omitempty"`
-	Data      string   `json:"data,omitempty"`
-	Flags     []string `json:"flags,omitempty"`
+	Comm      string   `json:"comm,omitempty" column:"comm,maxWidth:16"`
+	Pid       uint32   `json:"pid,omitempty" column:"pid,minWidth:7"`
+	Tid       uint32   `json:"tid,omitempty" column:"tid,minWidth:7"`
+	MountNsID uint64   `json:"mntnsid,omitempty" column:"mntns,width:12"`
+	Operation string   `json:"operation,omitempty" column:"op,minWidth:5,maxWidth:7,hide"`
+	Retval    int      `json:"ret,omitempty" column:"ret,width:3,fixed,hide"`
+	Latency   uint64   `json:"latency,omitempty" column:"latency,minWidth:3,hide"`
+	Fs        string   `json:"fs,omitempty" column:"fs,minWidth:3,maxWidth:8,hide"`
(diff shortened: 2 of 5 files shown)


burak-ok:
The integration test fails because the regex can not find the testname `test-mountsnoop-6657685279158859080`, since it gets shortened to `test-mountsnoop-6657685279158…`

Is the testname vital to this integration test?
Many of these column library transformation PRs will have this kind of issue if the columns are too large or the window is too small

blanquicet:
Hi @burak-ok, thanks for working on this!

> The integration test fails because the regex can not find the testname `test-mountsnoop-6657685279158859080`, since it gets shortened to `test-mountsnoop-6657685279158…`
> 
> Is the testname vital to this integration test? Many of these column library transformation PRs will have this kind of issue if the columns are too large or the window is too small

It is because you first need to modify the integration tests to use JSON output mode. This PR is an example of updating the tests and starting using the columns library: https://github.com/kinvolk/inspektor-gadget/pull/990. I will update the description of https://github.com/kinvolk/inspektor-gadget/issues/1003 to mention this step.

blanquicet:
> I will update the description of #1003 to mention this step.
>
Actually, I just realized it is already mentioned in the _Heads up_ section 🙂.

burak-ok:
Additionally changed the test to use the JSON output.

Should this be a separate PR?
Or is it okay when i order the commit in front of my changes like i did here

pkg/gadgets/trace/mount/types/types.go
```diff
@@ -15,25 +15,63 @@
...
+	MountNsID uint64   `json:"mntnsid,omitempty" column:"mnt_ns,maxWidth:12"`
+	Operation string   `json:"operation,omitempty" column:"op,maxWidth:10,hide"`
+	Retval    int      `json:"ret,omitempty" column:"ret,width:3,fixed,hide"`
+	Latency   uint64   `json:"latency,omitempty" column:"latency,minWidth:3,hide"`
+	Fs        string   `json:"fs,omitempty" column:"fs,minWidth:3,hide"`
+	Source    string   `json:"source,omitempty" column:"src,minWidth:3,hide"`
+	Target    string   `json:"target,omitempty" column:"dst,minWidth:3,hide"`
+	Data      string   `json:"data,omitempty" column:"data,width:40,hide"`
+	Flags     []string `json:"flags,omitempty" column:"flagsArr,width:40,hide"`
+	FlagsRaw  uint64   `json:"flagsRaw,omitempty" column:"flagsRaw,width:40,hide"`
```
blanquicet, line 40: We want to have only one column named `FLAG`:

```suggestion
	Flags     []string `json:"flags,omitempty" column:"flags,width:24,hide"`
	FlagsRaw  uint64   `json:"flagsRaw,omitempty"`
```

pkg/gadgets/trace/mount/types/types.go
```diff
@@ -15,25 +15,63 @@
...
+func GetColumns() *columns.Columns[Event] {
+	cols := columns.MustCreateColumns[Event]()
+
+	cols.MustAddColumn(columns.Column[Event]{
+		Name:    "call",
+		Width:   100,
+		Visible: true,
+		Order:   1000,
+		Extractor: func(e *Event) string {
+			return getCall(e)
```
blanquicet, line 52: nit: I think there is no need to define `getCall`, but we can put the code here directly.

integration/integration_test.go
```diff
@@ -664,10 +665,35 @@ func TestMountsnoop(t *testing.T) {
...
 	mountsnoopCmd := &command{
-		name:           "StartMountsnoopGadget",
-		cmd:            fmt.Sprintf("$KUBECTL_GADGET trace mount -n %s", ns),
-		expectedRegexp: fmt.Sprintf(`%s\s+test-pod\s+test-pod+\s+mount\s+\d+\s+\d+\s+\d+\s+mount\("/mnt", "/mnt", .*\) = -2`, ns),
-		startAndStop:   true,
+		name:         "StartMountsnoopGadget",
+		cmd:          fmt.Sprintf("$KUBECTL_GADGET trace mount -n %s -o json", ns),
+		startAndStop: true,
+		expectedOutputFn: func(output string) error {
+			expectedEntries := []*mountTypes.Event{
```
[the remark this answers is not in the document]
burak-ok: Did not know that 👍

pkg/gadgets/trace/mount/types/types.go
```diff
@@ -15,25 +15,63 @@
...
+	Tid       uint32   `json:"tid,omitempty" column:"tid,maxWidth:7"`
+	MountNsID uint64   `json:"mntnsid,omitempty" column:"mnt_ns,maxWidth:12"`
+	Operation string   `json:"operation,omitempty" column:"op,maxWidth:10,hide"`
+	Retval    int      `json:"ret,omitempty" column:"ret,width:3,fixed,hide"`
+	Latency   uint64   `json:"latency,omitempty" column:"latency,minWidth:3,hide"`
+	Fs        string   `json:"fs,omitempty" column:"fs,minWidth:3,hide"`
+	Source    string   `json:"source,omitempty" column:"src,minWidth:3,hide"`
+	Target    string   `json:"target,omitempty" column:"dst,minWidth:3,hide"`
+	Data      string   `json:"data,omitempty" column:"data,width:40,hide"`
+	Flags     []string `json:"flags,omitempty" column:"flagsArr,width:40,hide"`
```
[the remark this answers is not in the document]
burak-ok: For call i think that a bigger value seems to be better, since it is probably the main source of information that people would want to see

pkg/gadgets/trace/mount/types/types.go
```diff
@@ -15,27 +15,57 @@
...
-	Latency   uint64   `json:"latency,omitempty"`
-	Fs        string   `json:"fs,omitempty"`
-	Source    string   `json:"source,omitempty"`
-	Target    string   `json:"target,omitempty"`
-	Data      string   `json:"data,omitempty"`
-	Flags     []string `json:"flags,omitempty"`
+	Comm      string   `json:"comm,omitempty" column:"comm,maxWidth:16"`
+	Pid       uint32   `json:"pid,omitempty" column:"pid,minWidth:7"`
+	Tid       uint32   `json:"tid,omitempty" column:"tid,minWidth:7"`
+	MountNsID uint64   `json:"mntnsid,omitempty" column:"mnt_ns,width:12"`
```
blanquicet, line 31: We are using `mntns` for other gadgets:

```suggestion
	MountNsID uint64   `json:"mntnsid,omitempty" column:"mntns,width:12"`
```

I think this changes implies you will have to update the [documentation](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/trace/mount.md).

pkg/gadgets/trace/mount/types/types.go
```diff
@@ -15,27 +15,57 @@
...
+	Fs        string   `json:"fs,omitempty" column:"fs,minWidth:3,maxWidth:8,hide"`
+	Source    string   `json:"source,omitempty" column:"src,width:16,hide"`
+	Target    string   `json:"target,omitempty" column:"dst,width:16,hide"`
+	Data      string   `json:"data,omitempty" column:"data,width:16,hide"`
+	Flags     []string `json:"flags,omitempty" column:"flags,width:24,hide"`
 	FlagsRaw  uint64   `json:"flagsRaw,omitempty"`
 }
 
+func GetColumns() *columns.Columns[Event] {
+	cols := columns.MustCreateColumns[Event]()
```
blanquicet, line 44: I think it is missing configuring the extractor for flags:

https://github.com/kinvolk/inspektor-gadget/blob/5b93ab18b3b22741b26ae326d259156928d0cd54/cmd/kubectl-gadget/trace/mount.go#L149

integration/integration_test.go
```diff
@@ -664,10 +665,33 @@ func TestMountsnoop(t *testing.T) {
...
+
+			normalize := func(e *mountTypes.Event) {
+				e.Node = ""
+				e.Pid = 0
+				e.Tid = 0
+				e.MountNsID = 0
+				e.Latency = 0
+				e.Fs = ""
+				e.Data = ""
+				e.Flags = nil
```
blanquicet, line 689: I noticed in the output that we `Flags` is always `["MS_SILENT"]` (`mount` command seems to use it by default). I wonder if we should also verify it, but maybe it is better to specify the option `-o silent` to be sure it will always be present.

```bash
busyboxPodRepeatCommand(ns, "mount /mnt /mnt -o silent"),
```

Anyway, we can manage it in another PR.

blanquicet (commented):
I think it is ready to go after these comments are managed  🚀

integration/inspektor-gadget/integration_test.go
```diff
@@ -629,10 +630,35 @@ func TestMountsnoop(t *testing.T) {
...
 	mountsnoopCmd := &Command{
-		Name:           "StartMountsnoopGadget",
-		Cmd:            fmt.Sprintf("$KUBECTL_GADGET trace mount -n %s", ns),
-		ExpectedRegexp: fmt.Sprintf(`%s\s+test-pod\s+test-pod+\s+mount\s+\d+\s+\d+\s+\d+\s+mount\("/mnt", "/mnt", .*\) = -2`, ns),
-		StartAndStop:   true,
+		Name:         "StartMountsnoopGadget",
+		Cmd:          fmt.Sprintf("$KUBECTL_GADGET trace mount -n %s -o json", ns),
+		StartAndStop: true,
+		ExpectedOutputFn: func(output string) error {
+			expectedEntries := []*mountTypes.Event{
```
blanquicet, line 637: It doesn't need to be an array.

blanquicet (approved):
LGTM! Thanks for managing all the comments 🙂

mauriciovasquezbernal (approved):
LGTM. Works as expected !

PR #1012, opened 2022-10-05
manifest: remove startup probe

mauriciovasquezbernal:
Inspektor Gadget now takes less time to start up. Less remove that probe that shouldn't be needed anymore.

Ref https://github.com/kinvolk/inspektor-gadget/pull/942
Ref https://github.com/kinvolk/inspektor-gadget/issues/799

---

@blanquicet any comments on this one?


blanquicet:
I am only afraid of this case:

https://github.com/kinvolk/inspektor-gadget/blob/b22f18e61ecca795e06f483c372609e1551d69bf/gadget-container/entrypoint.sh#L104-L107

Even if we have seen that RH is backporting the BTF feature and enabling `kheaders` in most of the kernels of OpenShift (tested in ARO). BTW, [this](https://github.com/kinvolk/inspektor-gadget/blob/b22f18e61ecca795e06f483c372609e1551d69bf/gadget-container/entrypoint.sh#L65-L68) logic needs to be updated to consider BTF/CORE, or maybe we should delete it (ARO doesn't need it anymore)? If we decide **not** to delete it, we should keep the `initialDelaySeconds` for the liveness probe.

mauriciovasquezbernal:
Good point. I think we can remove that logic, IMHO it adds few value and complicates us. Let me open an issue to discus it.

mauriciovasquezbernal:
That logic was removed in https://github.com/kinvolk/inspektor-gadget/pull/1020, so I think we can move ahead with this.

blanquicet (approved):
LGTM!

PR #1013, opened 2022-10-05
Use sigHandler() as unique signal handler

eiffel-fl:
Hi.

In this PR, I removed the `sigs` channel in `genericStreams()`, thus all the signals are now handled in `sigHandler`.
This prevents having race condition regarding which channel will receive the signal first.

Best regards.

### DIFF
--- a/cmd/kubectl-gadget/utils/trace.go
+++ b/cmd/kubectl-gadget/utils/trace.go
@@ -701,7 +701,7 @@ var sigIntReceivedNumber = 0
 // On reception of this signal, the given trace will be deleted.
 // This function fixes trace not being deleted when calling:
 // kubectl gadget process-collector -A | head -n0
-func sigHandler(traceID *string) {
+func sigHandler(traceID *string, printTerminationMessage bool) {
 	c := make(chan os.Signal)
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/utils/trace.go
```diff
@@ -1037,7 +1037,12 @@ func genericStreams(
 	for {
 		select {
 		case msg := <-completion:
-			fmt.Fprintln(os.Stderr, msg)
+			output := os.Stdout
+			if strings.HasPrefix(msg, "Error") {
+				output = os.Stderr
+			}
+			fmt.Fprintln(output, msg)
+
```
mauriciovasquezbernal, line 1045: It seems to hacky to me. I'd rather remove the "completed" message than having this logic here.

cmd/kubectl-gadget/utils/trace.go
```diff
@@ -701,7 +701,7 @@ var sigIntReceivedNumber = 0
 // On reception of this signal, the given trace will be deleted.
 // This function fixes trace not being deleted when calling:
 // kubectl gadget process-collector -A | head -n0
-func sigHandler(traceID *string) {
+func sigHandler(traceID *string, willPrintTerminatingMessage bool) {
```
mauriciovasquezbernal, line 704:
```suggestion
func sigHandler(traceID *string, printTerminatingMessage bool) {
```

?

mauriciovasquezbernal (approved):
This change LGTM, it fixes the problem that we have because we're adding multiple handlers for SIGINT.

cmd/kubectl-gadget/utils/trace.go
blanquicet: `printTerminationMessage`? 😆

blanquicet (approved):
LGTM!

PR #1016, opened 2022-10-06
makefile: Some changes on `minikube-install` target

blanquicet:
# makefile: Some changes on `minikube-install` target

This PR adds the following changes:

1. Remove the hardcoded minikube profile name so that users can run:
```bash
MINIKUBE="minikube -p minikube-containerd" make minikube-install
```

2. If the image is not correctly loaded into Minikube, then we shouldn't go ahead and try to redeploy IG.


burak-ok (approved):
LGTM
And i learned a new bash trick with the `||` to run only when the previous command failed. I knew that the operator existed but did not think to use it in that way :D

PR #1017, opened 2022-10-06
trace/fsslower: Use columns library

burak-ok:
# trace/fsslower: Use columns library

The trace fsslower gadget now uses the columns library

## Testing done

Columnwidths are not the same as the old one. Currently it looks like this:
image


pkg/gadgets/trace/fsslower/types/types.go
```diff
@@ -25,14 +26,18 @@ const (
...
-	Latency   uint64 `json:"latency,omitempty"`
-	File      string `json:"file,omitempty"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mnt_ns_id,hide"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,width:5,maxWidth:7"`
+	Comm      string `json:"comm,omitempty" column:"comm,width:10"`
+	Op        string `json:"op,omitempty" column:"T,maxWidth:3"`
+	Bytes     uint64 `json:"bytes,omitempty" column:"bytes,width:10"`
+	Offset    int64  `json:"offset,omitempty" column:"offset,width:10"`
+	Latency   uint64 `json:"latency,omitempty" column:"lat,width:10"`
+	File      string `json:"file,omitempty" column:"file,width:30"`
```
flyth, line 36: Just a couple of remarks on the column settings:

[Here's](https://github.com/kinvolk/inspektor-gadget/blob/main/pkg/gadgets/trace/exec/types/types.go#L27) an example of names/widths/settings we're already using for some of the fields that are also present here, like `mountns` / `pid` / `comm` - I think we should reuse those here as well.

I'd also right align `bytes`, `offset` and `latency` using `align:right` for better readability. `T`/`Op` seems as if it can use a fixed width of 1 (`width:1,fixed`).

MaxLength of `file` can be set to 32 (see [here](https://github.com/kinvolk/inspektor-gadget/blob/784fa014c715fd4f1f338f4a3122f6b066b0edee/pkg/gadgets/trace/fsslower/tracer/bpf/fsslower.h#L5)).

burak-ok: Updated with your suggestions.
Would it be possible to defines these values somewhere once and use it in every setting?

flyth: Yeah, thought about that too while writing the comment. We could either embed structs or somehow give the opportunity to register defaults/templates based on the column names with the lib itself. But that sounds pretty hacky.

blanquicet: Yeah, it would be too hacky with the lib itself, but I agree we should have it in a common place to avoid redefining it in every gadget.

mauriciovasquezbernal (commented):
Two minor comments. Looks very good!

cmd/common/trace/fsslower.go
```diff
@@ -0,0 +1,70 @@
...
+
+	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
+	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/fsslower/types"
+	"github.com/spf13/cobra"
+)
+
+type FsSlowerFlags struct {
+	FsSlowerMinLatency uint
+	FsSlowerFilesystem string
+}
```
blanquicet, line 29: I know it was already like that, but could you please remove the `FsSlower` prefix, it is redudant:

```suggestion
type FsSlowerFlags struct {
	MinLatency uint
	Filesystem string
}
```

blanquicet (commented):
Could you please update the [gif](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/trace/fsslower.gif)?

integration/inspektor-gadget/integration_test.go
```diff
@@ -611,25 +611,23 @@ func TestFsslower(t *testing.T) {
...
 				e.MountNsID = 0
 				e.Pid = 0
-				e.Op = ""
 				e.Bytes = 0
 				e.Offset = 0
 				e.Latency = 0
 			}
 
-			return ExpectEntriesToMatch(output, normalize, expectedEntries...)
+			return ExpectEntriesToMatch(output, normalize, expectedEntry)
```
mauriciovasquezbernal, line 630: nit: I think you amended this change to the wrong commit. All these changes should be on the first commit of this PR.

burak-ok:
The newly generated `.gif` has a different size.
Is that okay? How can i configure it exactly pixel perfect?

Another remark: There should be a discussion about the default column widths after everything is using the library.
Does not look that good currently

mauriciovasquezbernal:
> Is that okay? How can i configure it exactly pixel perfect?

I hit the same problem some days ago. I think we need to update the script to use a fixed terminal size, otherwise we'll end with a lot of gifs with different size. Would you mind opening an issue for it?

> Another remark: There should be a discussion about the default column widths after everything is using the library.
Does not look that good currently

Do you have any specific feedback now? What's is not looking good?

mauriciovasquezbernal:
> IMO the first 4 columns of almost every gadget (Name for Node, Namespace, Pod, Container) are too wide in relation to the following columns.
Maybe they should be have a lower weight(width) or we increase the weight of the gadget specific columns.

I agree that we can finish porting the gadgets to that approach and then we can see if there is any other tuning we can make.

pkg/gadgets/trace/fsslower/types/types.go
```diff
@@ -25,14 +26,18 @@ const (
...
-	Latency   uint64 `json:"latency,omitempty"`
-	File      string `json:"file,omitempty"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,minWidth:7"`
+	Comm      string `json:"comm,omitempty" column:"comm,maxWidth:16"`
+	Op        string `json:"op,omitempty" column:"T,width:1,fixed"`
+	Bytes     uint64 `json:"bytes,omitempty" column:"bytes,width:10,align:right"`
+	Offset    int64  `json:"offset,omitempty" column:"offset,width:10,align:right"`
+	Latency   uint64 `json:"latency,omitempty" column:"lat,width:10,align:right"`
+	File      string `json:"file,omitempty" column:"file,width:7,maxWidth:32"`
```
mauriciovasquezbernal, line 36: I think 7 is too short for file. Maybe something like 24 would work better?

cmd/common/trace/fsslower.go
```diff
@@ -0,0 +1,70 @@
...
+
+package trace
+
+import (
+	"fmt"
+	"strings"
+
+	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
+	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/fsslower/types"
+	"github.com/spf13/cobra"
```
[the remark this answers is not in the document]
blanquicet: Still not resolved

blanquicet (approved):
Just one comment but LGTM!

mauriciovasquezbernal:
https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/trace/fsslower.md does need to be updated as well?

mauriciovasquezbernal (changes requested):
Please update imports names before merging. (https://github.com/inspektor-gadget/inspektor-gadget/pull/1040).

PR #1020, opened 2022-10-07
gadget-container: Remove logic to download kernel headers for CentOS

mauriciovasquezbernal:
In 30f49a6 we added logic to download and install the kernel headers for CentOS when they were not available on the host. Inspektor Gadget and the ecosystem has changed a lot since that, all of our gadgets use BTF and it's available on a lot of distros now, even in CentOS.

It's time to remove this logic and concentrate the efforts on supporting BTF enabled kernels.

Fixes https://github.com/kinvolk/inspektor-gadget/issues/1019


blanquicet (approved):
LGTM!

Issue #451, opened 2022-01-03
[RFE] Make container runtime socket path configurable


invidian:
## Current situation

As a follow up to https://github.com/kinvolk/inspektor-gadget/pull/434#issuecomment-998840919, as a prerequisite to #450 and to partially address #184, it would be nice if container runtime socket paths could be configurable.

## Impact

Simpler configuration in custom scenarios without increased complexity in default installations.

## Ideal future situation

One can use custom container runtime socket paths when using IG locally or in the pod.

alban:
[@flyth](https://github.com/flyth) has some branch for this. If anyone wants to work on this, please contact him to avoid double work.

Issue #644, opened 2022-05-02
[EPIC] Improve local-gadget filtering 


blanquicet:
## Current situation

Currently, it is only possible to filter gadgets by container name. So, we would like to provide the possibility to filter by all these parameters: 
- Container Name: 
  - Already supported. 
- Container ID:  
  - There could be multiple containers with the same name. 
  - Some containers do not have a name (e.g., pause containers on Containerd) so the container ID is also needed. 
- Container Runtime (Currently: Docker, Containerd or CRI-O) 
- Kubernetes metadata: 
  - Users that are debugging containers created through Kubernetes, could need to trace all the containers belonging to the same Kubernetes pod or namespace. This information is provided by the container runtime in the labels, so we don’t need the Kubernetes API Server to be up. For instance: 
    - `"io.kubernetes.pod.name": "coredns-78fcd69978-zp5gx"`
    - `"io.kubernetes.pod.namespace": "kube-system"`
- Container Runtime namespace: 
  - Some container runtimes, like Containerd, support the concept of "namespace". Users that have containers created by Kubernetes and others created directly with ctr (containerd CLI) could need to trace only the containers created by Kubernetes, this could be achieved by filtering the `k8s` namespace of Containerd. 

This epic is composed of the tasks to support this kind of filtering: 

1. Currently, we already support this for Docker, so we need to do the same for Containerd and CRIO. 
    - [ ] TODO: Create an issue to add support to non-Kubernetes containers with Containerd and CRI-O. 

2. Improve enrichment of containers done by WithMultipleContainerRuntimesEnrichment:
    - [ ] TODO: Create an issue for enrichment. This enricher needs to be improved to add the following information: Container Runtime Name, Kubernetes Metadata: Pod and namespace, Container runtime Namespace. 
    - [ ] TODO: Create an issue to create an internal data structure to store the Container definition and do not use the one from gRPC. Wait after @alban finishes the PR where he is adding more fields to it. 
  
3. Integration tests:
    - [ ] TODO: Create an issue to add integration test for local-gadget

mqasimsarfraz:
Closing, since all the sub-issues have been merged!

PR #1021, opened 2022-10-07
cmd/deploy: Add container socketpath configuration

burak-ok:
# cmd/deploy: Add container socketpath configuration

When the socketpath of the underlying container runtime is not the default it can be configured through these newly added cmd options of the `deploy` cmd
See #451. This change bases on https://github.com/flyth/inspektor-gadget/tree/michael/socketpaths

The newly added 3 subcommands to the `deploy` command are the following:
`--socketpath-docker`, `socketpath-containerd` and `--socketpath-containerd`

# Usage
`kubectl gadget deploy --socketpath-docker /burak.docker.sock` for docker.
Same for `crio` and `containerd`

# Testing
## When configuring another (but wrong) socket path
```
time="2022-10-07T14:59:46Z" level=warning msg="Skip pod default/kubernetes-bootcamp-75c5d958ff-f2s5l: cannot find container (ID: docker://1fb5a0580223858c8ee7d105be6b74e01472630abaf28c760a29cfef02e340e6): Cannot connect to the Docker daemon at unix:///burak.docker.sock. Is the docker daemon running?"
time="2022-10-07T14:59:46Z" level=warning msg="Skip pod gadget/gadget-tkww7: cannot find container (ID: docker://c316045e652f6059f9015718cad81efc33c3738e055bb86391943e026075802c): Cannot connect to the Docker daemon at unix:///burak.docker.sock. Is the docker daemon running?"
time="2022-10-07T14:59:46Z" level=warning msg="Skip pod kube-system/coredns-565d847f94-vxw6m: cannot find container (ID: docker://368d30eee7c4c7152e12263ff0b071d18e898e1e38c3798402c81cf7827b7912): Cannot connect to the Docker daemon at unix:///burak.docker.sock. Is the docker daemon running?"
time="2022-10-07T14:59:46Z" level=warning msg="Skip pod kube-system/etcd-minikube: cannot find container (ID: docker://617eaa50f09506d55ed29003c274d2e16019182c7f2c6ccc5f0343584515ad90): Cannot connect to the Docker daemon at unix:///burak.docker.sock. Is the docker daemon running?"
```
## Same deployment when the socket `/burak.docker.sock` is there:
```
time="2022-10-07T15:00:29Z" level=info msg="GadgetTracerManager: enabling fallback podinformer"
time="2022-10-07T15:00:59Z" level=info msg="Starting Pod controller"
time="2022-10-07T15:00:59Z" level=info msg="Serving on gRPC socket /run/gadgettracermanager.socket"
time="2022-10-07T15:00:59Z" level=info msg="Starting trace controller manager"
```
-> IG works

### DIFF
--- a/cmd/common/utils/flags.go
+++ b/cmd/common/utils/flags.go
@@ -20,6 +20,9 @@ import (
 	"strings"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 )
 
 const (
@@ -40,6 +43,23 @@ type OutputConfig struct {
 
 	// Verbose prints additional information
 	Verbose bool
+}
+
+func AddOutputFlags(command *cobra.Command, outputConfig *OutputConfig) {
+	command.PersistentFlags().StringVarP(
+		&outputConfig.OutputMode,
+		"output",
+		"o",
+		OutputModeColumns,
+		fmt.Sprintf("Output format (%s).", strings.Join(SupportedOutputModes, ", ")),
+	)
+
+	command.PersistentFlags().BoolVarP(
+		&outputConfig.Verbose,
+		"verbose", "v",
+		false,
+		"Print debug information",
+	)
 }
 
 func (config *OutputConfig) ParseOutputConfig() error {
@@ -74,4 +94,33 @@ func (config *OutputConfig) ParseOutputConfig() error {
 		return WrapInErrInvalidArg("--output / -o",
 			fmt.Errorf("%q is not a valid output format", config.OutputMode))
 	}
+}
+
+type RuntimesSocketPathConfig struct {
+	Docker     string
+	Containerd string
+	Crio       string
+}
+
+func AddRuntimesSocketPathFlags(command *cobra.Command, config *RuntimesSocketPathConfig) {
+	command.PersistentFlags().StringVarP(
+		&config.Docker,
+		"docker-socketpath", "",
+		runtimeclient.DockerDefaultSocketPath,
+		"Docker Engine API Unix socket path",
+	)
+
+	command.PersistentFlags().StringVarP(
+		&config.Containerd,
+		"containerd-socketpath", "",
+		runtimeclient.ContainerdDefaultSocketPath,
+		"containerd CRI Unix socket path",
+	)
+
+	command.PersistentFlags().StringVarP(
+		&config.Crio,
+		"crio-socketpath", "",
+		runtimeclient.CrioDefaultSocketPath,
+		"CRI-O CRI Unix socket path",
+	)
 }
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -74,9 +74,12 @@ var (
 	quiet               bool
 	debug               bool
 	wait                bool
+	runtimesConfig      commonutils.RuntimesSocketPathConfig
 )
 
 func init() {
+	commonutils.AddRuntimesSocketPathFlags(deployCmd, &runtimesConfig)
+
 	deployCmd.PersistentFlags().StringVarP(
 		&image,
 		"image", "",
@@ -320,6 +323,12 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 					gadgetContainer.Env[i].Value = hookMode
 				case "INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER":
 					gadgetContainer.Env[i].Value = strconv.FormatBool(fallbackPodInformer)
+				case utils.GadgetEnvironmentContainerdSocketpath:
+					gadgetContainer.Env[i].Value = runtimesConfig.Containerd
+				case utils.GadgetEnvironmentCRIOSocketpath:
+					gadgetContainer.Env[i].Value = runtimesConfig.Crio
+				case utils.GadgetEnvironmentDockerSocketpath:
+					gadgetContainer.Env[i].Value = runtimesConfig.Docker
 				}
 			}
 
--- a/cmd/kubectl-gadget/utils/flags.go
+++ b/cmd/kubectl-gadget/utils/flags.go
@@ -144,6 +144,8 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
 	// No 'Namespace' flag because it's added automatically by
 	// KubernetesConfigFlags.AddFlags(rootCmd.PersistentFlags())
 
+	commonutils.AddOutputFlags(command, &params.OutputConfig)
+
 	command.PersistentFlags().StringVarP(
 		&params.LabelsRaw,
 		"selector",
@@ -183,26 +185,9 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
 		"Show data from pods in all namespaces",
 	)
 
-	command.PersistentFlags().StringVarP(
-		&params.OutputMode,
-		"output",
-		"o",
-		commonutils.OutputModeColumns,
-		fmt.Sprintf("Output format (%s).", strings.Join(commonutils.SupportedOutputModes, ", ")),
-	)
-
-	command.PersistentFlags().BoolVarP(
-		&params.Verbose,
-		"verbose",
-		"",
-		false,
-		"Print additional information",
-	)
-
-	command.PersistentFlags().IntVarP(
+	command.PersistentFlags().IntVar(
 		&params.Timeout,
 		"timeout",
-		"",
 		0,
 		"Number of seconds that the gadget will run for",
 	)
--- a/pkg/container-utils/docker/docker.go
+++ b/pkg/container-utils/docker/docker.go
@@ -30,9 +30,7 @@ import (
 )
 
 const (
-	Name              = "docker"
-	DefaultSocketPath = "/run/docker.sock"
-	DefaultTimeout    = 2 * time.Second
+	DefaultTimeout = 2 * time.Second
 )
 
 // DockerClient implements the ContainerRuntimeClient interface but using the
@@ -47,7 +45,7 @@ type DockerClient struct {
 
 func NewDockerClient(socketPath string) (runtimeclient.ContainerRuntimeClient, error) {
 	if socketPath == "" {
-		socketPath = DefaultSocketPath
+		socketPath = runtimeclient.DockerDefaultSocketPath
 	}
 
 	cli, err := client.NewClientWithOpts(
@@ -121,7 +119,7 @@ func (c *DockerClient) GetContainer(containerID string) (*runtimeclient.Containe
 }
 
 func (c *DockerClient) GetContainerDetails(containerID string) (*runtimeclient.ContainerDetailsData, error) {
-	containerID, err := runtimeclient.ParseContainerID(Name, containerID)
+	containerID, err := runtimeclient.ParseContainerID(runtimeclient.DockerName, containerID)
 	if err != nil {
 		return nil, err
 	}
@@ -149,7 +147,7 @@ func (c *DockerClient) GetContainerDetails(containerID string) (*runtimeclient.C
 			ID:      containerJSON.ID,
 			Name:    strings.TrimPrefix(containerJSON.Name, "/"),
 			State:   containerStatusStateToRuntimeClientState(containerJSON.State.Status),
-			Runtime: Name,
+			Runtime: runtimeclient.DockerName,
 		},
 		Pid:         containerJSON.State.Pid,
 		CgroupsPath: string(containerJSON.HostConfig.Cgroup),
@@ -221,7 +219,7 @@ func DockerContainerToContainerData(container *dockertypes.Container) *runtimecl
 		ID:      container.ID,
 		Name:    strings.TrimPrefix(container.Names[0], "/"),
 		State:   containerStatusStateToRuntimeClientState(container.State),
-		Runtime: Name,
+		Runtime: runtimeclient.DockerName,
 	}
 
 	// Fill K8S information.
--- a/pkg/container-utils/docker/docker.go
+++ b/pkg/container-utils/docker/docker.go
@@ -24,9 +24,10 @@ import (
 	dockertypes "github.com/docker/docker/api/types"
 	dockerfilters "github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cgroups"
 	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
-	log "github.com/sirupsen/logrus"
 )
 
 const (
--- a/pkg/resources/manifests/deploy.yaml
+++ b/pkg/resources/manifests/deploy.yaml
@@ -162,6 +162,13 @@ spec:
             value: "auto"
           - name: INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER
             value: "true"
+          # Make sure to keep these settings in sync with pkg/container-utils/runtime-client/interface.go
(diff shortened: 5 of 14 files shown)


cmd/kubectl-gadget/deploy.go
```diff
@@ -64,16 +64,19 @@ var deployCmd = &cobra.Command{
...
+	livenessProbe        bool
+	deployTimeout        time.Duration
+	fallbackPodInformer  bool
+	printOnly            bool
+	quiet                bool
+	debug                bool
+	wait                 bool
+	socketPathContainerd string
+	socketPathCRIO       string
+	socketPathDocker     string
```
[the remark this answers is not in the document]
blanquicet: > Since, we can only select one container runtime
>
Actually, in a multi-node cluster, we could have different runtimes on each node. So, if we want to manage such a case, we need all three flags, even if it makes the CLI more complex 🙁

cmd/kubectl-gadget/deploy.go
```diff
@@ -128,6 +131,21 @@ func init() {
 		"debug", "d",
 		false,
 		"show extra debug information")
+	deployCmd.PersistentFlags().StringVar(
+		&socketPathContainerd,
+		"socketpath-containerd",
+		"",
```
blanquicet, line 137: I think we should put here the default value we are using to let users know what path we are currently using. For instance, for Docker ["/run/docker.sock"](https://github.com/kinvolk/inspektor-gadget/blob/5b93ab18b3b22741b26ae326d259156928d0cd54/pkg/container-utils/docker/docker.go#L34).

cmd/kubectl-gadget/utils/flags.go
```diff
@@ -191,19 +200,38 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
...
 		"timeout",
-		"",
 		0,
 		"Number of seconds that the gadget will run for",
 	)
+
+	command.PersistentFlags().StringVar(
+		&params.DockerSocketPath,
+		"docker-socketpath",
+		"",
```
blanquicet, line 220: dito

pkg/resources/manifests/deploy.yaml
```diff
@@ -172,6 +172,12 @@ spec:
             value: "auto"
           - name: INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER
             value: "true"
+          - name: INSPEKTOR_GADGET_SOCKETPATH_CONTAINERD
+            value: ""
```
blanquicet, line 176: dito

cmd/kubectl-gadget/deploy.go
```diff
@@ -128,6 +131,21 @@ func init() {
...
+	deployCmd.PersistentFlags().StringVar(
+		&socketPathCRIO,
+		"socketpath-crio",
+		"",
+		"socket path to use for CRI-O runtime")
+	deployCmd.PersistentFlags().StringVar(
+		&socketPathDocker,
+		"socketpath-docker",
+		"",
+		"socket path to use for docker runtime")
```
blanquicet, line 148: IMO, we should align the naming notation with what local-gadget is doing. Maybe `socket-<runtime>` is a better naming:

https://github.com/kinvolk/inspektor-gadget/blob/5b93ab18b3b22741b26ae326d259156928d0cd54/cmd/local-gadget/utils/flags.go#L134-L153

cmd/kubectl-gadget/utils/flags.go
```diff
@@ -191,19 +200,38 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
...
+		"containerd CRI Unix socket path",
+	)
+
+	command.PersistentFlags().StringVar(
+		&params.CrioSocketPath,
+		"crio-socketpath",
+		"",
+		"CRI-O CRI Unix socket path",
+	)
```
blanquicet, line 236: I think you should move this to a dedicated function in a common place. Then, we can use it here, in the `deploy` command, and in `local-gadget`. Something similar to what we did for [the output flags](https://github.com/kinvolk/inspektor-gadget/blob/main/cmd/common/utils/flags.go). In this way, we can keep them aligned. WDYT?

cmd/kubectl-gadget/utils/flags.go
```diff
@@ -191,19 +200,38 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
...
-	command.PersistentFlags().IntVarP(
+	command.PersistentFlags().IntVar(
 		&params.Timeout,
 		"timeout",
-		"",
 		0,
 		"Number of seconds that the gadget will run for",
 	)
+
+	command.PersistentFlags().StringVar(
```
blanquicet, line 217: I wonder if it is also necessary to define these flags for each gadget, or if it is ok just for the `deploy` command.

cmd/kubectl-gadget/utils/flags.go
burak-ok: Seems good to me to move it to that place

cmd/kubectl-gadget/deploy.go
burak-ok: Yes, definitely they should be the same.
But like you asked: Which one should we pick :D `<runtime>-socket` or `socket-<runtime>`

cmd/kubectl-gadget/deploy.go
burak-ok: Had thought about that as well. What i wanted was to reference the `DefaultSocketPath` from `container-utils/docker/docker.go` where the default socket path was already defined, since i didn't want to have 2 lines with default socket paths.
Problem was that it introduced a circular import and so it did not compile :)

So we would also need to move that default socketpath definition from `docker.go` (same for crio and containerd) to another independent file, right?

cmd/kubectl-gadget/deploy.go
blanquicet: Mmmm I am ok either way. Pick the one you prefer! 🙂

cmd/kubectl-gadget/deploy.go
blanquicet: Mmmmm I don't see what is the circular import 🤔

cmd/kubectl-gadget/deploy.go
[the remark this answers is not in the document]
blanquicet: They will increase if we add support for other runtimes, so I think that's a price to pay for making IG project compatible with more systems out there 🙂

cmd/kubectl-gadget/utils/flags.go
burak-ok: Yep. It makes only sense for the `deploy` command

burak-ok:
New addtional changes for unification:
1. new struct `CommonFlags` in `cmd/common/utils/flags.go` (Rename struct to `BaseCommonFlags?)
    - Every variable that `local-gadget` and `kubectl-gadget` needs in their `CommonFlags`
1. new function `AddCommonFlags` in the same file as above to add the commands relating to the above varaibles to cobra

Question:
Since @blanquicet rightfully mentioned that we need the socketpath options only for the `deploy` command of `kubectl-gadget`: Should the socketpath variables be in this central `CommonFlags`? Or would it be better to have them in one time in `local-gadget` und a second time in the `deploy` flags of `kubectl-gadget`

cmd/common/utils/flags.go
```diff
@@ -30,6 +31,69 @@ const (
 
 var SupportedOutputModes = []string{OutputModeColumns, OutputModeJSON, OutputModeCustomColumns}
 
+type CommonFlags struct {
+	// Containername allows to filter containers by name.
+	Containername string
```
blanquicet, line 36: I suggest not adding this here. It will diverge between `kubectl-gadget` and `local-gadget` after https://github.com/inspektor-gadget/inspektor-gadget/issues/644 is implemented.

cmd/common/utils/flags.go
```diff
@@ -30,6 +31,69 @@ const (
 
 var SupportedOutputModes = []string{OutputModeColumns, OutputModeJSON, OutputModeCustomColumns}
 
+type CommonFlags struct {
```
blanquicet, line 34: Maybe `BaseCommonFlags`? I don't know ... It seems strange to me to have a `CommonFlags` that has an embedded member also called `CommonFlags`, don't you think?

pkg/container-utils/docker/docker.go
```diff
@@ -24,15 +24,15 @@ import (
...
 	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 	log "github.com/sirupsen/logrus"
 )
 
 const (
-	Name              = "docker"
-	DefaultSocketPath = "/run/docker.sock"
-	DefaultTimeout    = 2 * time.Second
+	Name           = "docker"
+	DefaultTimeout = 2 * time.Second
```
blanquicet, line 35: I think it is useful for people using only the `container-utils` package to have a default value here.

pkg/container-utils/docker/docker.go
```diff
@@ -24,15 +24,15 @@ import (
 	dockertypes "github.com/docker/docker/api/types"
 	dockerfilters "github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
+
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cgroups"
 	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 	log "github.com/sirupsen/logrus"
```
blanquicet, line 30: Could you please take advantage of the opportunity to move `log` to the correct place? :)

cmd/common/utils/consts.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+const (
+	DockerDefaultSocketPath     = "/run/docker.sock"
+	CrioDefaultSocketPath       = "/run/crio/crio.sock"
+	ContainerdDefaultSocketPath = "/run/containerd/containerd.sock"
+)
```
blanquicet, line 21: Still not possible to get these values from the docker/containerd/crio packages?

blanquicet (commented):
I like the idea of creating a `BaseCommonFlags` for flags shared between `kubectl-gadget` and `local-gadget`!

Consider that making changes progressively is a good practice (It also makes the review easier). For instance, one commit to crate the `BaseCommonFlags` in common/utils, and another commit to making kubectl use that new struct 🙂

blanquicet:
> we need the socketpath options only for the `deploy` command of `kubectl-gadget`
>
That's the main question, do we want to provide this option for the gadgets or only for the deploy command?

@mauriciovasquezbernal

cmd/kubectl-gadget/utils/flags.go
mauriciovasquezbernal: What do you mean for each gadget? 

I think those are needed for each gadget in local-gadget, but not for Inspektor-Gadget in those case the path should already be configured when deploying it. 

Also, are command line parameters the best UI here? It seems cumbersome to have to specify them for each invocation of local-gadget:

```
$ local-gadget snapshot process --containerd-socketpath=/bar/foo
$ local-gadget trace exec --containerd-socketpath=/bar/foo
```

Perhaps those should be defined as env variables too? (similar to how kubectl handles kubeconfig) (In another PR for sure). 

```
export CONTAINERD_SOCKETPATH=/bar/foo
$ local-gadget snapshot process 
$ local-gadget trace exec 
```

It seems it's possible by using viper + cobra https://github.com/kubereboot/kured/pull/464/files.

blanquicet: > What do you mean for each gadget?
>
If, as for local-gadget, we want to provide the possibility of using a different path each time you call a gadget:

```bash
$ kubectl gadget trace exec --containerd-socketpath=/bar/foo
```

I proposed to add them only for the `kubectl gadget deploy` command, and I see you agree.

On the other hand, I created https://github.com/inspektor-gadget/inspektor-gadget/issues/1044 to track this improvement:

> It seems it's possible by using viper + cobra https://github.com/kubereboot/kured/pull/464/files.

pkg/resources/manifests/deploy.yaml
```diff
@@ -162,6 +162,12 @@ spec:
             value: "auto"
           - name: INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER
             value: "true"
+          - name: INSPEKTOR_GADGET_CONTAINERD_SOCKETPATH
+            value: "/run/containerd/containerd.sock"
+          - name: INSPEKTOR_GADGET_CRIO_SOCKETPATH
+            value: "/run/crio/crio.sock"
+          - name: INSPEKTOR_GADGET_DOCKER_SOCKETPATH
+            value: "/run/docker.sock"
```
blanquicet, line 170: Please add a comment here to please keep these values aligned with the ones in the runtime-client package (where the default values are actually defined)

cmd/common/utils/flags.go
```diff
@@ -75,3 +95,32 @@ func (config *OutputConfig) ParseOutputConfig() error {
...
+func AddRuntimesSocketPathFlags(command *cobra.Command, config *RuntimesSocketPathConfig) {
+	command.PersistentFlags().StringVarP(
+		&config.Docker,
+		"docker-socketpath", "",
+		runtimeclient.DockerDefaultSocketPath,
+		"Docker Engine API Unix socket path",
+	)
+
+	command.PersistentFlags().StringVarP(
+		&config.Crio,
```
blanquicet, line 114:
```suggestion
		&config.Containerd,
```

It causes the following issue. Notice I don't have CRI-O on the system I am testing 🙂:

```bash
$ sudo local-gadget list-containers -r cri-o
RUNTIME ID NAME

$ sudo local-gadget list-containers -r cri-o --containerd-socketpath /run/containerd/containerd.sock
RUNTIME ID                                                               NAME
cri-o   a02b7270da188366a3f8e45c81e4bd4f7a406cc7a9e653543c87dfb2a3d96be3 calico-kube-controllers
cri-o   c939449f0be102ef541fdc3bd53701f90d5b5752ab94a5b221e9e870b0114e10 calico-node
cri-o   4bfc88dcbb0c76529929c1c0eee71016070ee298c40a7677d9fdf0ea003be74c coredns
cri-o   aaa3d1df1d3c17e544d58235e68c9fb98da03d4a8670e6274530c241190010f0 coredns
cri-o   4de1d33e372ec3a31f94327aaa99a576bc609bd286a91470b3ff19bea1073092 etcd
cri-o   83fd1238f1facb7c2fcd2e003da28eb80dcb93784668524dfd4371a9d35bc228 gadget
cri-o   37dc8f960b631c7f239b2e0c2cf0ffc7499462b71d5a941b3b40168ace4f3fab kube-apiserver
cri-o   70415e780e48356f41cdcf264e5492bd65fad73e829e8e2c34056de2b7c4e083 kube-controller-manager
cri-o   be344c75a9fe460951ee0088781785db009b37209b8481d6627b2864a966c2c6 kube-proxy
cri-o   4e1878a766921c25eff18ffb801063d78caa7e281dc81db70419cbe6fdb59531 kube-scheduler
```

blanquicet (changes requested):
After fixing the typo, I think it is ready to go 🚀

cmd/kubectl-gadget/utils/flags.go
```diff
@@ -183,26 +185,9 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
...
-		fmt.Sprintf("Output format (%s).", strings.Join(commonutils.SupportedOutputModes, ", ")),
-	)
-
-	command.PersistentFlags().BoolVarP(
-		&params.Verbose,
-		"verbose",
-		"",
-		false,
-		"Print additional information",
-	)
```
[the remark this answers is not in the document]
burak-ok: Make sense, I will try to split up my future commits even more

cmd/common/utils/flags.go
burak-ok: More diverse test configurations would be better :D

blanquicet (approved):
LGTM! Thanks for managing all the comments. 

I don't know if @mauriciovasquezbernal wants to have another look.

mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #1024, opened 2022-10-10
columns: Update `signal` and `tcp` to use (min|max)Width feature

blanquicet:
# Update `signal` and `tcp` columns to use (min|max)Width feature

As discussed in https://github.com/kinvolk/inspektor-gadget/pull/934#discussion_r977580600 and https://github.com/kinvolk/inspektor-gadget/pull/934#discussion_r977459998, update `signal` and `tcp` to use `(min|max)Width` feature.


mauriciovasquezbernal (approved):
LGTM!

PR #1027, opened 2022-10-10
trace/oomkill: Use columns library

burak-ok:
# trace/mount: Use columns library

The trace mount gadget now uses the columns library.

For the Test: The triggered command is most likely `tail` itself but technically it is not guaranteed. Should i modify the `exptectedEntry`?

## Testing done

image


cmd/kubectl-gadget/trace/oomkill.go
```diff
@@ -15,107 +15,39 @@
...
-	"fmt"
-	"strings"
-
 	"github.com/spf13/cobra"
 
 	commontrace "github.com/kinvolk/inspektor-gadget/cmd/common/trace"
 	commonutils "github.com/kinvolk/inspektor-gadget/cmd/common/utils"
 	"github.com/kinvolk/inspektor-gadget/cmd/kubectl-gadget/utils"
-	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/oomkill/types"
+	"github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/fsslower/types"
```
blanquicet, line 23: fsslower? 🙂

cmd/kubectl-gadget/trace/oomkill.go
```diff
@@ -15,107 +15,39 @@
...
+	oomkillTypes "github.com/kinvolk/inspektor-gadget/pkg/gadgets/trace/oomkill/types"
 )
 
 type OOMKillParser struct {
 	commonutils.BaseParser[types.Event]
 }
```
blanquicet, line 30: Not needed anymore.

integration/inspektor-gadget/integration_test.go
```diff
@@ -764,10 +765,27 @@ func TestOomkill(t *testing.T) {
...
+
+			normalize := func(e *oomkillTypes.Event) {
+				e.Node = ""
+				e.KilledPid = 0
+				e.Pages = 0
+				e.TriggeredPid = 0
+				e.MountNsID = 0
+			}
+
+			return ExpectEntriesToMatch(output, normalize, expectedEntry)
```
blanquicet, line 787: I don't expect other OOM to happen, so what about using `ExpectAllToMatch`?

integration/inspektor-gadget/integration_test.go
```diff
@@ -764,10 +765,27 @@ func TestOomkill(t *testing.T) {
...
-		ExpectedRegexp: `\d+\s+tail`,
-		StartAndStop:   true,
+		Name:         "StartOomkilGadget",
+		Cmd:          fmt.Sprintf("$KUBECTL_GADGET trace oomkill -n %s -o json", ns),
+		StartAndStop: true,
+		ExpectedOutputFn: func(output string) error {
+			expectedEntry := &oomkillTypes.Event{
+				Event:         BuildBaseEvent(ns),
+				KilledComm:    "tail",
+				TriggeredComm: "tail",
```
blanquicet, line 775: > The triggered command is most likely tail itself but technically it is not guaranteed. Should i modify the exptectedEntry?
>
@burak-ok, I think @eiffel-fl could help you with this question.

cmd/kubectl-gadget/trace/oomkill.go
burak-ok: That is correct. And VSCode imported `fsslower/types` automagically because of `types.Event`

integration/inspektor-gadget/integration_test.go
burak-ok: Would be possible if `TriggeredComm` is always `tail` (The question)

integration/inspektor-gadget/integration_test.go
blanquicet: > I think we should try to stick with the regex when converting tests to JSON.
>
I don't follow you. In which way?

blanquicet: I think the new implemenation Burak just pushed is the correct way to go. What are you proposing?

blanquicet (approved):
LGTM!

integration/inspektor-gadget/integration_test.go
burak-ok: The previous test with `regex` tested on only one occurence of `tail`

pkg/gadgets/trace/oomkill/types/types.go
```diff
@@ -15,18 +15,25 @@
...
+	Pages         uint64 `json:"pages,omitempty" column:"pages,width:6"`
+	TriggeredPid  uint32 `json:"tpid,omitempty" column:"tpid,minWidth:7"`
+	TriggeredComm string `json:"tcomm,omitempty" column:"tcomm,maxWidth:16"`
+	MountNsID     uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+}
+
+func GetColumns() *columns.Columns[Event] {
+	cols := columns.MustCreateColumns[Event]()
+
+	return cols
```
mauriciovasquezbernal, line 36: nit:

```suggestion
	return columns.MustCreateColumns[Event]()
```

mauriciovasquezbernal (commented):
A minor comment. Looks good otherwise.

PR #1030, opened 2022-10-11
trace/open: Use columns library

burak-ok:
# trace/open: Use columns library

The trace open gadget now uses the columns library

## Testing done
image

### DIFF
--- a/pkg/gadgets/trace/open/types/types.go
+++ b/pkg/gadgets/trace/open/types/types.go
@@ -15,20 +15,25 @@
 package types
 
 import (
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
 type Event struct {
 	eventtypes.Event
 
-	MountNsID uint64 `json:"mountnsid,omitempty"`
-	Pid       uint32 `json:"pid,omitempty"`
-	UID       uint32 `json:"uid,omitempty"`
-	Comm      string `json:"pcomm,omitempty"`
-	Fd        int    `json:"fd,omitempty"`
-	Ret       int    `json:"ret,omitempty"`
-	Err       int    `json:"err,omitempty"`
-	Path      string `json:"path,omitempty"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,minWidth:7"`
+	UID       uint32 `json:"uid,omitempty" column:"uid,minWidth:10,hide"`
+	Comm      string `json:"pcomm,omitempty" column:"comm,maxWidth:16"`
+	Fd        int    `json:"fd,omitempty" column:"fd,minWidth:2,width:3"`
+	Ret       int    `json:"ret,omitempty" column:"ret,width:3,fixed,hide"`
+	Err       int    `json:"err,omitempty" column:"err,width:3,fixed"`
+	Path      string `json:"path,omitempty" column:"path,minWidth:24,width:32"`
+}
+
+func GetColumns() *columns.Columns[Event] {
+	return columns.MustCreateColumns[Event]()
 }
 
 func Base(ev eventtypes.Event) Event {
--- a/cmd/common/trace/open.go
+++ b/cmd/common/trace/open.go
@@ -1,0 +1,27 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewOpenCmd(runCmd func(*cobra.Command, []string) error) *cobra.Command {
+	return &cobra.Command{
+		Use:   "open",
+		Short: "Trace open system calls",
+		RunE:  runCmd,
+	}
+}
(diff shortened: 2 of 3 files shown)


pkg/gadgets/trace/open/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Comm      string `json:"pcomm,omitempty"`
-	Fd        int    `json:"fd,omitempty"`
-	Ret       int    `json:"ret,omitempty"`
-	Err       int    `json:"err,omitempty"`
-	Path      string `json:"path,omitempty"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,minWidth:7,align:right"`
+	UID       uint32 `json:"uid,omitempty" column:"uid,minWidth:10,align:right,hide"`
+	Comm      string `json:"pcomm,omitempty" column:"comm,maxWidth:16"`
+	Fd        int    `json:"fd,omitempty" column:"fd,width:3,align:right,fixed"`
```
mauriciovasquezbernal, line 29: I think it should not use fixed. It's not impossible for a process to have more than 999 open files.

pkg/gadgets/trace/open/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	Pid       uint32 `json:"pid,omitempty"`
-	UID       uint32 `json:"uid,omitempty"`
-	Comm      string `json:"pcomm,omitempty"`
-	Fd        int    `json:"fd,omitempty"`
-	Ret       int    `json:"ret,omitempty"`
-	Err       int    `json:"err,omitempty"`
-	Path      string `json:"path,omitempty"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,minWidth:7,align:right"`
+	UID       uint32 `json:"uid,omitempty" column:"uid,minWidth:10,align:right,hide"`
```
mauriciovasquezbernal, line 27: ditto about `align:right`.

pkg/gadgets/trace/open/types/types.go
```diff
@@ -15,20 +15,25 @@
...
+	Comm      string `json:"pcomm,omitempty" column:"comm,maxWidth:16"`
+	Fd        int    `json:"fd,omitempty" column:"fd,width:3,align:right,fixed"`
+	Ret       int    `json:"ret,omitempty" column:"ret,width:3,align:right,fixed,hide"`
+	Err       int    `json:"err,omitempty" column:"err,width:3,align:right,fixed"`
+	Path      string `json:"path,omitempty" column:"path,minWidth:50"`
```
mauriciovasquezbernal, line 32: This one is difficult to choose. However I think 50 as minimum is too much. Maybe 32? Let's see others opinion on this.

mauriciovasquezbernal (commented):
Some comments about the parameters.

pkg/gadgets/trace/open/types/types.go
blanquicet: Yeah, maybe 24 or, at most, 32. We should define a value for the paths, what other gadgets have this column? Mount and fsslower?

pkg/gadgets/trace/open/types/types.go
```diff
@@ -15,20 +15,25 @@
...
-	MountNsID uint64 `json:"mountnsid,omitempty"`
-	Pid       uint32 `json:"pid,omitempty"`
-	UID       uint32 `json:"uid,omitempty"`
-	Comm      string `json:"pcomm,omitempty"`
-	Fd        int    `json:"fd,omitempty"`
-	Ret       int    `json:"ret,omitempty"`
-	Err       int    `json:"err,omitempty"`
-	Path      string `json:"path,omitempty"`
+	MountNsID uint64 `json:"mountnsid,omitempty" column:"mntns,width:12,hide"`
+	Pid       uint32 `json:"pid,omitempty" column:"pid,minWidth:7,align:right"`
```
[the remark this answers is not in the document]
blanquicet: IMO, `align:right` looks good for those columns representing a measurement (latency, offset, etc..), but not for the IDs (pid, uid, mountns, etc..). What do you think?

@flyth

pkg/gadgets/trace/open/types/types.go
blanquicet: Mmmm the `minWidth` would be 2 (because of the column name "fd"), and the `maxWidth`, as you said, is 3. Maybe at this point is better to have it fixed to 3?

pkg/gadgets/trace/open/types/types.go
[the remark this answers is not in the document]
burak-ok: Then again what do we do about the return value or error code. Right align or not.

Until now we did not align anything to the right except measurements (afaik)

pkg/gadgets/trace/open/types/types.go
burak-ok: Mauricio meant that `Fd` could be bigger than 999 and therefore the maximum is not 3. Theoretically it could be as big as the `int` that holds it

pkg/gadgets/trace/open/types/types.go
burak-ok: It differs.
For example fsslower has a `maxWidth` of `32`. That makes sense for that specific gadget since the eBPF program only captures `32` characters.
With this gadget the filepath can be up to `255` characters.

It is truly difficult to choose every columnwidth perfect for every gadget. Maybe it can't be unified as we had thought it could be.

Reason why i had chosen 50 here was just by trying different widths and running the gadget. When i felt that the gadget displayed enough of the filepath i stopped increasing the value

pkg/gadgets/trace/open/types/types.go
blanquicet: Oh, he said, "It is not impossible" ... @mauriciovasquezbernal, you confused me with this double negation hahahaha

pkg/gadgets/trace/open/types/types.go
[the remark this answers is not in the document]
blanquicet: I suggest keeping them like they are right now (default align: left). Then, using the templates [feature](https://github.com/inspektor-gadget/inspektor-gadget/pull/1037), we could define the default for all of them.

blanquicet (approved):
LGTM!

PR #1033, opened 2022-10-11
pkg/runcfanotify: Fix bug and improve detection logic 

mauriciovasquezbernal:
This PR brings two improvements for the runcfanotify package:
- Make Supported() logic more robust
- Use HOST_ROOT env variable

## What does change with this PR? 

- The logic doesn't try to use fanotify on systems where it can't work, for instance, systems without CONFIG_FANOTIFY_ACCESS_PERMISSIONS. (like minikube using the kvm2 driver for instance)
- This fixes a bug when the bundle path of a container is in a path that can't be accessed by the gadget pod. 

## How to test?

- Deploy this version in a system without CONFIG_FANOTIFY_ACCESS_PERMISSIONS and check how it doesn't use fanotify by default when hook-mode = auto. 
- Use this version in a system that creates the pods in a path that is not directly accessible from the gadget pod, like `/var/run`.

### DIFF
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -138,24 +138,14 @@ func initFanotify() (*fanotify.NotifyFD, error) {
 
 // Supported detects if RuncNotifier is supported in the current environment
 func Supported() bool {
-	// Test that runc is available
-	runcFound := false
-	for _, path := range runcPaths {
-		if _, err := os.Stat(filepath.Join(hostRoot, path)); err == nil {
-			runcFound = true
-			break
-		}
+	notifier, err := NewRuncNotifier(func(notif ContainerEvent) {})
+	if notifier != nil {
+		notifier.Close()
 	}
-	if !runcFound {
-		return false
-	}
-	// Test that it's possible to run fanotify
-	notifyFD, err := initFanotify()
 	if err != nil {
-		return false
+		log.Debugf("Runcfanotify: not supported: %s", err)
 	}
-	notifyFD.File.Close()
-	return true
+	return err == nil
 }
 
 // NewRuncNotifier uses fanotify to detect when runc containers are created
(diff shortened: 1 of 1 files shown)


pkg/runcfanotify/runcfanotify.go
```diff
@@ -175,14 +162,26 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
 	}
 	n.runcBinaryNotify = runcBinaryNotify
 
+	runcFound := false
+
 	for _, runcPath := range runcPaths {
 		file := filepath.Join(hostRoot, runcPath)
-		err = runcBinaryNotify.Mark(unix.FAN_MARK_ADD, unix.FAN_OPEN_EXEC_PERM, unix.AT_FDCWD, file)
-		if err == nil {
-			log.Debugf("Checking %q: done", file)
```
blanquicet, line 179: I think it is useful to print which path it is trying

pkg/runcfanotify/runcfanotify.go
```diff
@@ -175,14 +162,26 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
...
+			return nil, fmt.Errorf("fanotify mark: %w", err)
 		}
+
+		log.Debugf("Runcfanotify: Monitoring runc at %s", file)
+	}
+
+	if !runcFound {
+		runcBinaryNotify.File.Close()
+		return nil, fmt.Errorf("runc not found")
 	}
```
blanquicet, line 185: Or print the list of paths that were considered here.

alban (approved):

pkg/runcfanotify/runcfanotify.go
```diff
@@ -175,14 +162,26 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
...
+			runcBinaryNotify.File.Close()
+			return nil, fmt.Errorf("fanotify mark: %w", err)
 		}
+
+		log.Debugf("Runcfanotify: Monitoring runc at %s", file)
+	}
+
+	if !runcFound {
+		runcBinaryNotify.File.Close()
+		return nil, fmt.Errorf("runc not found")
```
blanquicet, line 184: nit: 

```suggestion
		return nil, errors.New("runc not found")
```

pkg/runcfanotify/runcfanotify.go
```diff
@@ -138,24 +138,11 @@ func initFanotify() (*fanotify.NotifyFD, error) {
...
-	runcFound := false
-	for _, path := range runcPaths {
-		if _, err := os.Stat(filepath.Join(hostRoot, path)); err == nil {
-			runcFound = true
-			break
-		}
-	}
-	if !runcFound {
-		return false
+	notifier, err := NewRuncNotifier(func(notif ContainerEvent) {})
```
blanquicet, line 141: What about adding a log here to print the error?

pkg/runcfanotify/runcfanotify.go
```diff
@@ -175,14 +162,26 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
...
-			log.Debugf("Checking %q: done", file)
-		} else {
-			log.Debugf("Checking %q: %s", file, err)
+		if _, err := os.Stat(file); errors.Is(err, os.ErrNotExist) {
+			continue
+		}
+		runcFound = true
+
+		if err := runcBinaryNotify.Mark(unix.FAN_MARK_ADD, unix.FAN_OPEN_EXEC_PERM, unix.AT_FDCWD, file); err != nil {
+			runcBinaryNotify.File.Close()
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > Maybe defer this?

In order to use defer I'll have to change the return values to be named and check if err != nil. IMO it's not worth it in this case. 

>  do you want to try all the paths or you want to stop at the first found? In this case, you can break after the below log.Debugf.

We should check all of them. (It could happen that there are two different runc instances installed on the system and we don't want to miss the containers from one of them).

pkg/runcfanotify/runcfanotify.go
mauriciovasquezbernal: I changed the logic a bit to always print the paths.

pkg/runcfanotify/runcfanotify.go
```diff
@@ -173,13 +165,28 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
...
+		if _, err := os.Stat(runcPath); errors.Is(err, os.ErrNotExist) {
+			log.Debugf("Runcfanotify: runc at %s not found", runcPath)
+			continue
 		}
+		runcFound = true
+
+		if err := runcBinaryNotify.Mark(unix.FAN_MARK_ADD, unix.FAN_OPEN_EXEC_PERM, unix.AT_FDCWD, runcPath); err != nil {
+			runcBinaryNotify.File.Close()
+			return nil, fmt.Errorf("fanotify mark: %w", err)
+		}
```
blanquicet, line 184: I suppose you are returning here because there is no way this operation fails on one path but succeeds on another, right?

blanquicet (approved):
LGTM

PR #1035, opened 2022-10-12
capabilities: remove CAP_ prefix

alban:
The capabilities trace gadget gave caps like CAP_SYS_ADMIN but security contexts in Kubernetes requires the cap name without the CAP_  prefix:
```
 capabilities:
   add: ["SETPCAP", "SYS_ADMIN", "SYS_RESOURCE", "SYSLOG", "IPC_LOCK", "DAC_OVERRIDE"]
```
Remove the CAP_ prefix so that it's easier to copy/paste.

The new width of the field is 18 because the larger capability is: CHECKPOINT_RESTORE


pkg/gadgets/trace/capabilities/types/types.go
```diff
@@ -38,7 +38,7 @@ type Event struct {
 	Comm      string `json:"comm,omitempty" column:"comm,maxWidth:16"`
 	UID       uint32 `json:"uid,omitempty" column:"uid,minWidth:6"`
 	Cap       int    `json:"cap,omitempty" column:"cap,width:3,fixed"`
-	CapName   string `json:"capName,omitempty" column:"capName,width:20,ellipsis:start"`
+	CapName   string `json:"capName,omitempty" column:"capName,minWidth:18,maxWidth:18,ellipsis:start"`
```
blanquicet, line 41:
```suggestion
	CapName   string `json:"capName,omitempty" column:"capName,width:18,fixed"`
```

`minWidth:18,maxWidth:18` means to have a fixed width. With this change, `ellipsis` is not needed anymore as it will never be truncated. @flyth

flyth: It could still be truncated if there's no more room anywhere else - however, removing the ellipsis parameter will let it fall back to the default (which should be `end`), which makes more sense now that the common prefix of `CAP_` has been removed.

alban: Ok, I marked it as fixed, and I removed the ellipsis to use the default (`end`).

blanquicet (approved):
I didn't test it, but we have multiple tests for this gadget in the CI. So, LGTM from code inspection!

mauriciovasquezbernal:
@alban would you mind updating https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/trace/capabilities.gif? :D

PR #1036, opened 2022-10-12
trace/capabilities: Implement units tests for unique flag

mauriciovasquezbernal:
Implement unit tests for `--unique` flag introduced in https://github.com/kinvolk/inspektor-gadget/pull/1032.

### DIFF
--- a/pkg/gadgets/trace/capabilities/tracer/tracer_test.go
+++ b/pkg/gadgets/trace/capabilities/tracer/tracer_test.go
@@ -218,6 +218,52 @@ func TestCapabilitiesTracer(t *testing.T) {
 				}
 			},
 		},
+		"unique_false": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *tracer.Config {
+				return &tracer.Config{
+					MountnsMap: utilstest.CreateMntNsFilterMap(t, info.MountNsID),
+					Unique:     false,
+				}
+			},
+			generateEvent: repeatChown,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, _ interface{}, events []types.Event) {
+				nfound := 0
+				for _, event := range events {
+					if event.CapName == "CHOWN" {
+						nfound++
+					}
+				}
+
+				if nfound <= 1 {
+					t.Fatalf("Capability not found")
+				}
+			},
+		},
+		"unique_true": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *tracer.Config {
+				return &tracer.Config{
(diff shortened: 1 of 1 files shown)


pkg/gadgets/trace/capabilities/tracer/tracer_test.go
```diff
@@ -218,6 +218,48 @@ func TestCapabilitiesTracer(t *testing.T) {
...
+				return &tracer.Config{
+					MountnsMap: utilstest.CreateMntNsFilterMap(t, info.MountNsID),
+					Unique:     true,
+				}
+			},
+			generateEvent: repeatChown,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, _ interface{}, events []types.Event) {
+				nfound := 0
+				for _, event := range events {
+					if event.CapName == "CAP_CHOWN" {
```
alban, line 253: If we merge https://github.com/kinvolk/inspektor-gadget/pull/1035, it should be changed to "CHOWN".

pkg/gadgets/trace/capabilities/tracer/tracer_test.go
```diff
@@ -218,6 +218,48 @@ func TestCapabilitiesTracer(t *testing.T) {
...
+						nfound++
+					}
+				}
+
+				if nfound > 1 {
+					t.Fatalf("Same capability found multiple times")
```
alban, line 259: Also return an error for the case nfound == 0:

```suggestion
				if nfound != 1 {
					t.Fatalf("Capability not unique: found %d times", nfound)
```

alban (commented):
Thanks! LGTM, pending on a few comments

alban (approved):

pkg/gadgets/trace/capabilities/tracer/tracer_test.go
```diff
@@ -218,6 +218,52 @@ func TestCapabilitiesTracer(t *testing.T) {
...
+
+				if nfound <= 1 {
+					t.Fatalf("Capability not found")
+				}
+			},
+		},
+		"unique_true": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *tracer.Config {
+				return &tracer.Config{
+					MountnsMap: utilstest.CreateMntNsFilterMap(t, info.MountNsID),
```
blanquicet, line 245: Maybe too complicated, but it would be useful also to implement a test where we generate the same capability event from two differente mount namespaces (both included in the `MountnsMap`) and verify that one event is generated per mout namespace.

Issue #1038, opened 2022-10-13
gadgets/types: inconsistent field comm vs pcomm


alban:
## Description

Since [#934](https://github.com/kinvolk/inspektor-gadget/pull/934/files#diff-d076c830bee7d56b6efa5d94936180122d6a914f7e481e18709f4c7425964053R29), the json field in pkg/gadgets/trace/exec/types/types.go is named `pcomm` but the column is named `comm`:
```
Comm      string   `json:"pcomm,omitempty" column:"comm,maxWidth:16"`
```

## Impact 1

Users use `kubectl gadget trace exec -o json` to learn about the columns and find the `pcomm` field:
```
{"node":"minikube","namespace":"demo","pod":"mypod","container":"mypod","type":"normal","pid":643158,"ppid":641892,"pcomm":"ls","args":["/bin/ls"],"mountnsid":4026533258}
```
Then, trying `-o custom-columns=pcomm` fails:
```
Error: creating parser: invalid columns: pcomm
```

## Impact 2

The script [demo.sh](https://github.com/kinvolk/inspektor-gadget/blob/af9630eb772979da5e6a15a3012823534188d2ed/tools/demos/trace-exec/demo.sh#L21) uses `-o custom-columns=pcomm` and it returns `Error: creating parser: invalid columns: pcomm`

## Environment and steps to reproduce

.

## Expected behavior

Easy to deduce column names from json fields.

## Additional information

Introduced by #934.

I suggest to use `comm`. There will be other changes to do:

```
$ git grep pcomm
```

PR #1039, opened 2022-10-13
capabilities: update documentation

alban:
Follow-up to https://github.com/kinvolk/inspektor-gadget/pull/1035

While rebuilding the gif videos, I encountered some problems, see individual commits.

See also #1038.


tools/demos/Makefile
```diff
@@ -23,6 +23,11 @@ all-gadgets: $(addsuffix /demo.gif,$(DEMO_LIST_EXCEPT_INSTALL))
 	agg $*/demo.cast $*/demo.gif
 	ls -lh $*/demo.gif
 
+.PHONY: check-terminal
+check-terminal:
+	test `tput lines` -eq 24
```
mauriciovasquezbernal, line 28: What about using https://stackoverflow.com/a/8308063 to try to set the size before testing it?

mauriciovasquezbernal:
Given that you're updating all the gifs. What do you think about using the new "kubectl gadget version" command (https://github.com/kinvolk/inspektor-gadget/pull/977) instead of https://github.com/kinvolk/inspektor-gadget/blob/main/tools/demos/install/demo.sh#L25?

alban:
Updated

blanquicet:
> It seems to show the versions of the client and the server in a different way:
> 
> ```
> $ kubectl gadget version
> Client version: v0.9.1-47-g269ceca6-dirty
> Server version: alban_caps_prefix2
> ```
> 
> Should I use it anyway?

It will have to be update it to use the next released CLI and image. In that way, both will be tagged and the output would be:

$ kubectl gadget version
Client version: v0.10.0
Server version: v0.10.0

blanquicet (approved):
LGTM! But I think having the capabilities column with a fixed size is not the best solution. It is taking up too much space without the need.

alban:
> LGTM! But I think having the capabilities column with a fixed size is not the best solution. It is taking up too much space without the need.

Right, in most cases it takes too much space because the capability with the long name is not the one frequently used. But it is the most important column though... Any suggestions?

PR #1042, opened 2022-10-14
trace/network: Use columns library

burak-ok:
# trace/network: Use columns library

The trace network gadget now uses the columns library

### Naming of gadgets
The names of the gadget is `network` but the gadgetname of the deployed pod is `network-graph` (See modified `network.md`). This is also true for other gadgets like `mount` <-> `mountsnoop`. 

Is this by design or should i open a separate issue for this to discuss it?

### Annotating every variable in the `Event` struct with `column:`
This is the first PR where i did not anootate every variable with a `column:...`
Do we want to do that or are we leaving these hidden fields without an annotation?

## Testing done
image

### DIFF
--- a/pkg/gadgets/trace/network/types/types.go
+++ b/pkg/gadgets/trace/network/types/types.go
@@ -18,39 +18,69 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns/ellipsis"
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
 type Event struct {
 	eventtypes.Event
 
-	PktType string `json:"pktType,omitempty"`
-	Proto   string `json:"proto,omitempty"`
-	Addr    string `json:"addr,omitempty"`
-	Port    uint16 `json:"port,omitempty"`
+	PktType string `json:"pktType,omitempty" column:"type,maxWidth:9"`
+	Proto   string `json:"proto,omitempty" column:"proto,maxWidth:5"`
+	Addr    string `json:"addr,omitempty" column:"addr,template:ipaddr,hide"`
+	Port    uint16 `json:"port,omitempty" column:"ipport,template:ipport"`
 
 	/* pod, svc or other */
-	RemoteKind string `json:"remoteKind,omitempty"`
+	RemoteKind string `json:"remoteKind,omitempty" column:"kind,maxWidth:5"`
 
-	PodHostIP string            `json:"podHostIP,omitempty"`
-	PodIP     string            `json:"podIP,omitempty"`
-	PodOwner  string            `json:"podOwner,omitempty"`
-	PodLabels map[string]string `json:"podLabels,omitempty"`
+	PodHostIP string            `json:"podHostIP,omitempty" column:"podhostip,template:ipaddr,hide"`
+	PodIP     string            `json:"podIP,omitempty" column:"podip,template:ipaddr,hide"`
+	PodOwner  string            `json:"podOwner,omitempty" column:"podowner,hide"`
+	PodLabels map[string]string `json:"podLabels,omitempty" column:"padlabels,hide"`
 
 	/* if RemoteKind = svc */
-	RemoteSvcNamespace     string            `json:"remoteServiceNamespace,omitempty"`
-	RemoteSvcName          string            `json:"remoteServiceName,omitempty"`
-	RemoteSvcLabelSelector map[string]string `json:"remoteServiceLabelSelector,omitempty"`
+	RemoteSvcNamespace     string            `json:"remoteServiceNamespace,omitempty" column:"remotesvcns,hide"`
+	RemoteSvcName          string            `json:"remoteServiceName,omitempty" column:"remotesvcname,hide"`
+	RemoteSvcLabelSelector map[string]string `json:"remoteServiceLabelSelector,omitempty" column:"remotesvclabel,hide"`
 
 	/* if RemoteKind = pod */
-	RemotePodNamespace string            `json:"remotePodNamespace,omitempty"`
-	RemotePodName      string            `json:"remotePodName,omitempty"`
-	RemotePodLabels    map[string]string `json:"remotePodLabels,omitempty"`
+	RemotePodNamespace string            `json:"remotePodNamespace,omitempty" column:"remotepodns,hide"`
+	RemotePodName      string            `json:"remotePodName,omitempty" column:"remotepodname,hide"`
+	RemotePodLabels    map[string]string `json:"remotePodLabels,omitempty" column:"remotepodlabel,hide"`
 
 	/* if RemoteKind = other */
-	RemoteOther string `json:"remoteOther,omitempty"`
+	RemoteOther string `json:"remoteOther,omitempty" column:"remoteother,template:ipaddr,hide"`
+}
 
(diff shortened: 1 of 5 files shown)


mauriciovasquezbernal:
> Is this by design or should i open a separate issue for this to discuss it?

These are leftover of our all naming schema. I think we should update all of them to the new naming. However, for network-graph I think we intentionally named it in that way). 

> Do we want to do that or are we leaving these hidden fields without an annotation?

I think we should annotate most of them. For instance, it's possible that somebody wants to get the raw IP address instead of pod related info: `/kubectl-gadget trace network -o custom-columns=node,namespace,pod,ip`

flyth (commented):
Besides the discussion about the naming it looks good to me.

pkg/gadgets/trace/network/types/types.go
```diff
@@ -18,16 +18,17 @@ import (
...
 type Event struct {
 	eventtypes.Event
 
-	PktType string `json:"pktType,omitempty"`
-	Proto   string `json:"proto,omitempty"`
+	PktType string `json:"pktType,omitempty" column:"type,maxWidth:9"`
+	Proto   string `json:"proto,omitempty" column:"proto,maxWidth:6"`
 	IP      string `json:"ip,omitempty"`
-	Port    int    `json:"port,omitempty"`
+	Port    int    `json:"port,omitempty" column:"port,maxWidth:6"`
```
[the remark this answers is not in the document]
blanquicet: Actually, I think we should make Port `uint16`. IMO, there is no reason to be `int`: 

https://github.com/inspektor-gadget/inspektor-gadget/blob/42d19b503bc3debde1a601f38dcb04d962898223/pkg/gadgets/trace/network/tracer/tracer.go#L217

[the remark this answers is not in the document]
flyth: Totally agree, seemed out-of-scope for this PR, though.

pkg/gadgets/trace/network/types/types.go
```diff
@@ -18,16 +18,17 @@ import (
...
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
 type Event struct {
 	eventtypes.Event
 
-	PktType string `json:"pktType,omitempty"`
-	Proto   string `json:"proto,omitempty"`
+	PktType string `json:"pktType,omitempty" column:"type,maxWidth:9"`
+	Proto   string `json:"proto,omitempty" column:"proto,maxWidth:6"`
```
blanquicet, line 29: We are using `maxWidth: 5` for the `Proto` in trace/bind as it is usually only TCP/UDP.

pkg/gadgets/trace/network/types/types.go
```diff
@@ -18,16 +18,17 @@ import (
...
 	IP      string `json:"ip,omitempty"`
```
blanquicet, line 30: Please also define the column for the IP address but with `hide`. BTW, We are using IP to describe if it is IPv4 or IPv6 in tcp/tcpconnect gadget. Instead, we use saddr or daddr or addr for the IP address (e.g. in bind, tcp, tcpconnect). I think this gadget should align with them: Use `Addr` instead of `IP` for the IP address.

BTW, we already define a sort of template for IP addresses:

```golang
// Min: XXX.XXX.XXX.XXX (IPv4) = 15
// Max: 0000:0000:0000:0000:0000:ffff:XXX.XXX.XXX.XXX (IPv4-mapped IPv6 address) = 45
Addr     string `json:"addr,omitempty" column:"addr,minWidth:15,maxWidth:45"`
```

pkg/gadgets/trace/network/types/types.go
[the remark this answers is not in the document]
blanquicet: I did it in https://github.com/inspektor-gadget/inspektor-gadget/pull/1050.

pkg/gadgets/trace/network/types/types.go
blanquicet: I aligned naming in I did it in https://github.com/inspektor-gadget/inspektor-gadget/pull/1050.

Please use `template:ipaddr` now that it has been [merged](https://github.com/inspektor-gadget/inspektor-gadget/pull/1048).

pkg/gadgets/trace/network/types/types.go
blanquicet: Maybe it is worth adding a template for it as well?

```golang
columns.MustRegisterTemplate("proto", "maxWidth:5")
```

cc @flyth

burak-ok:
`Event.Debug` doesn't seem to be written anywhere. Should i remove it?

pkg/gadgets/trace/network/types/types.go
```diff
@@ -18,39 +18,68 @@ import (
...
+	PodHostIP string            `json:"podHostIP,omitempty" column:"podhostip,template:ipaddr,hide"`
+	PodIP     string            `json:"podIP,omitempty" column:"podip,template:ipaddr,hide"`
+	PodOwner  string            `json:"podOwner,omitempty" column:"podowner,hide"`
+	PodLabels map[string]string `json:"podLabels,omitempty" column:"padlabels,hide"`
 
 	/* if RemoteKind = svc */
-	RemoteSvcNamespace     string            `json:"remoteServiceNamespace,omitempty"`
-	RemoteSvcName          string            `json:"remoteServiceName,omitempty"`
-	RemoteSvcLabelSelector map[string]string `json:"remoteServiceLabelSelector,omitempty"`
+	RemoteSvcNamespace     string            `json:"remoteServiceNamespace,omitempty" column:"remotesvcns,template:ns,hide"`
```
blanquicet, line 43: `template:ns` is used for Linux namespaces ID, and not for Kubernetes namespaces name.

cc @flyth maybe the template should be called `linuxNs` so that it is clearer?

pkg/gadgets/trace/network/types/types.go
```diff
@@ -18,39 +18,68 @@ import (
...
+	RemoteSvcName          string            `json:"remoteServiceName,omitempty" column:"remotesvcname,hide"`
+	RemoteSvcLabelSelector map[string]string `json:"remoteServiceLabelSelector,omitempty" column:"remotesvclabel,hide"`
 
 	/* if RemoteKind = pod */
-	RemotePodNamespace string            `json:"remotePodNamespace,omitempty"`
-	RemotePodName      string            `json:"remotePodName,omitempty"`
-	RemotePodLabels    map[string]string `json:"remotePodLabels,omitempty"`
+	RemotePodNamespace string            `json:"remotePodNamespace,omitempty" column:"remotepodns,template:ns,hide"`
```
blanquicet, line 48: dito

pkg/gadgets/trace/network/types/types.go
```diff
@@ -53,6 +54,34 @@ type Event struct {
 	Debug string `json:"debug,omitempty"`
 }
 
+func GetColumns() *columns.Columns[Event] {
+	cols := columns.MustCreateColumns[Event]()
+
+	cols.MustAddColumn(columns.Column[Event]{
+		Name:    "remote",
+		Width:   32,
```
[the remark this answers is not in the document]
blanquicet: I saw you changed the `Width`, but I was just proposing to add a `minWidth` given that it is the most important field of the gadget.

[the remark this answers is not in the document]
burak-ok: Well that makes even more sense :D

blanquicet (approved):
LGTM!

mauriciovasquezbernal (approved):
LGTM too!

PR #1045, opened 2022-10-17
ci: Use new way to set output values.

eiffel-fl:
Hi.

In this PR, I replaced the previous `::set-output` by the new way because the first one will be deprecated.

Best regards


blanquicet (approved):
LGTM!

PR #1046, opened 2022-10-17
integration-tests/networkgraph: Add network gadget test

burak-ok:
Added a test for the `network` gadget, since there was none yet.

Could not use `BuildBaseEvent(ns)` for the `Event`, since the `network` gadget doesn't capture the `CommonData.Container` string. (Is this intentional or did i do the trace wrong?)

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -33,6 +33,7 @@ import (
 	execTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/exec/types"
 	fsslowerType "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/fsslower/types"
 	mountTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/mount/types"
+	networkTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/network/types"
 	oomkillTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/oomkill/types"
 	openTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/open/types"
 	signalTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/signal/types"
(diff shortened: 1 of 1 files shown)


integration/inspektor-gadget/integration_test.go
```diff
@@ -909,6 +911,57 @@ func TestOpensnoop(t *testing.T) {
...
+					CommonData: eventtypes.CommonData{
+						Namespace: ns,
+						Pod:       "test-pod",
+					},
+				},
+				PktType:    "OUTGOING",
+				Proto:      "tcp",
+				Port:       443,
+				RemoteKind: "other",
+			}
```
blanquicet, line 936: It should be able to also capture the DNS request to the `svc kube-system/kube-dns`.

integration/inspektor-gadget/integration_test.go
```diff
@@ -909,6 +911,57 @@ func TestOpensnoop(t *testing.T) {
...
+
+			return ExpectEntriesToMatch(output, normalize, expectedEntry)
+		},
+	}
+
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		networkGraphCmd,
+		BusyboxPodRepeatCommand(ns, "wget -q -O /dev/null https://kinvolk.io"),
```
blanquicet, line 957: You could use the same test-pod used in [the guide](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/guides/trace/network-graph.md#on-kubernetes) so that you can verify also the IP "1.1.1.1"

integration/inspektor-gadget/integration_test.go
```diff
@@ -909,6 +911,57 @@ func TestOpensnoop(t *testing.T) {
...
+			}
+
+			normalize := func(e *networkTypes.Event) {
+				e.Node = ""
+				e.Container = ""
+				e.IP = ""
+				e.PodHostIP = ""
+				e.PodIP = ""
+				e.PodOwner = ""
+				e.PodLabels = nil
```
blanquicet, line 945: You can also verify the `PodLabels` as the pod we create using `BusyboxPodRepeatCommand` to generate the events has the label `run=test-pod`:

https://github.com/inspektor-gadget/inspektor-gadget/blob/b586002292b6ac79295d9657496d30ffb9c635a4/integration/command.go#L499-L500

blanquicet (commented):
> Could not use `BuildBaseEvent(ns)` for the `Event`, since the `network` gadget doesn't capture the `CommonData.Container` string. (Is this intentional or did i do the trace wrong?)
>
You can use it, and then remove the container info as the DNS gadget does:

https://github.com/inspektor-gadget/inspektor-gadget/blob/b586002292b6ac79295d9657496d30ffb9c635a4/integration/inspektor-gadget/integration_test.go#L472-L475

blanquicet (approved):
LGTM if CI is green. I expect it to work also on ARO and AKS. Let's see.

mauriciovasquezbernal (approved):
Nice. LGTM!

PR #1048, opened 2022-10-18
pkg/types: register column templates

flyth:
This PR adds the first batch of column templates discussed in #1037. I'll create another PR where I migrated column definitions to templates for already to columns converted gadgets.


blanquicet (approved):
LGTM!

PR #1050, opened 2022-10-18
gadgets/network: Align name and type notation with the rest of gadgets

blanquicet:
This PR introduces two changes to the network trace gadget:

1. Use `unint16` for the port field.
2. Currently, we are using `ip` to describe if it is IPv4 or IPv6 in tcp and tcpconnect gadget. Instead, we use `saddr`, `daddr` or `addr` for the IP addresses, e.g. in bind, tcp and tcpconnect. This commit aligns the network gadget to bind, tcp and tcpconnect gadgets.


flyth (approved):
LGTM! Thanks!

burak-ok (approved):
LGTM

PR #1051, opened 2022-10-18
pkg/types: move column template registration into init()

flyth:
This PR fixes the need to manually call Init(nodeName) for column template registration by putting it into an automatically called init() function.


blanquicet (approved):
LGTM!

PR #1055, opened 2022-10-19
integration-tests/networkgraph: Don't rely on svc names

burak-ok:
A previous PR https://github.com/inspektor-gadget/inspektor-gadget/pull/1046 breaks the ARO CI Test.

This PR loosens the requirements for the test to be successfull so it does not check the following fields anymore:
1. RemoteSvcNamespace
2. RemoteSvcName
3. RemoteSvcLabelSelector

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -932,15 +932,12 @@ func TestNetworkGraph(t *testing.T) {
 					RemoteOther: "1.1.1.1",
 				},
 				{
-					Event:                  BuildBaseEvent(ns),
-					PktType:                "OUTGOING",
-					Proto:                  "udp",
-					Port:                   53,
-					PodLabels:              map[string]string{"run": "test-pod"},
-					RemoteKind:             "svc",
-					RemoteSvcNamespace:     "kube-system",
-					RemoteSvcName:          "kube-dns",
-					RemoteSvcLabelSelector: map[string]string{"k8s-app": "kube-dns"},
+					Event:      BuildBaseEvent(ns),
+					PktType:    "OUTGOING",
+					Proto:      "udp",
+					Port:       53,
+					PodLabels:  map[string]string{"run": "test-pod"},
+					RemoteKind: "svc",
 				},
 			}
 			// Network gadget doesn't provide container data. Remove it.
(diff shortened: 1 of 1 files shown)


integration/inspektor-gadget/integration_test.go
```diff
@@ -932,15 +932,12 @@ func TestNetworkGraph(t *testing.T) {
...
-					RemoteKind:             "svc",
-					RemoteSvcNamespace:     "kube-system",
-					RemoteSvcName:          "kube-dns",
-					RemoteSvcLabelSelector: map[string]string{"k8s-app": "kube-dns"},
+					Event:      BuildBaseEvent(ns),
+					PktType:    "OUTGOING",
+					Proto:      "udp",
+					Port:       53,
+					PodLabels:  map[string]string{"run": "test-pod"},
+					RemoteKind: "svc",
```
blanquicet, line 940: I wonder if we could still verify these fields, but conditioned for ARO. Something like this:

```golang
			dnsEvent := networkTypes.Event{
				Event:                  BuildBaseEvent(ns),
				PktType:                "OUTGOING",
				Proto:                  "udp",
				Port:                   53,
				PodLabels:              map[string]string{"run": "test-pod"},
				RemoteKind:             "svc",
				RemoteSvcNamespace:     "kube-system",
				RemoteSvcName:          "kube-dns",
				RemoteSvcLabelSelector: map[string]string{"k8s-app": "kube-dns"},
			}

			// ARO cluster uses a custom DNS service
			if *k8sDistro == K8sDistroARO {
				dnsEvent.RemoteSvcNamespace = "openshift-dns"
				dnsEvent.RemoteSvcName = "dns-default"
				dnsEvent.RemoteSvcLabelSelector = map[string]string{"dns.operator.openshift.io/daemonset-dns": "default"}
			}
```

I would like to still verify them because the computation of those fields is done outside the tracer, so they cannot be tested on the unit-tests. The integration tests look like the correct place where we can verify they are computed correctly. WDYT?

burak-ok: If these are stable names and selectors, yes, why not.

mauriciovasquezbernal: I'm not so sure about relying on the name of the DNS service, it's K8s distribution dependent and could make it difficult to run integration tests in different distributions. (Do not think only about the ones that we run the CI in, but any distribution that a user can run the tests on). 

If we really want to test those fields we can create a different test scenario. For instance, to deploy a service and send requests to it.

blanquicet: > If we really want to test those fields we can create a different test scenario. For instance, to deploy a service and send requests to it.
>
Makes more sense.

blanquicet (approved):
LGTM!

Issue #849, opened 2022-08-22
[RFE] Avoid repeated code on gadgets within the top category


blanquicet:
## Current situation

We have the same code repeated on each gadget within the top category. Follow the same approach of https://github.com/kinvolk/inspektor-gadget/pull/713 to refactor the code and avoid such a situation. PR https://github.com/kinvolk/inspektor-gadget/pull/782 already prepared things for this.

PR #1063, opened 2022-10-20
Generalize parts of kubectl-gadget top gadgets

burak-ok:
# Generalize parts of kubectl-gadget top gadgets

All of the `top` gadgets of `kubectl-gadget` have similar variables, methods and code. This is an attempt to extract the similarities and put them into a central location `top.go`

1. All `top` gadgets use the sorting mechanism from the `colums` library
2. `TopParser` and `TopGadget` were added

Idea about this came when i started converting the gadgets to use the `columns` library. (Therefor the branch name 😅)

Found out later: https://github.com/inspektor-gadget/inspektor-gadget/issues/849


burak-ok:
Hi @blanquicet and @flyth ,

This whole PR is still a draft but i wanted to get some comments on the first commit `top/<gadget>/types.go: Generalized SortBy`.
Main changes are in `pkg/gadgets/top/top.go` and `pkg/gadgets/top/<gadget>/types/types.go`

### What i tried to solve
`SortedBy` looks like a type/struct which could be shared between all of the `top` gadgets.
Also there were 2 methods which were the same.

### What i did
Instead of having an own Datatype for the enumerations, they are all ints. Bad for type checking. Is there another possiblity to have a common type shared or an easy way with generics?

Additionally i introduced a `SortBy` interface which i maybe have to use in a future commit when i generalize the `top` gadgets even more (Should have added it with a later commit then probably)

`basicSortBy` is now a struct and implements all needed functions. The only thing for the gadgets to do in `type.go` is to pass a map into the function

### Questions from me

What possibilites does `Go` give me as a programmer to make this `SortBy` type/struct/whatever to be more generic?
What can be implemented more elegent and efficient?

flyth:
Thanks for creating more generic code!

Regarding a generic solution for the sorting that happens on the client as well as on the server side:
The columns library already has a subpackage `sort` to deal with that - so once we migrate the top gadgets to use columns, we can just use that in a pretty generic way.

In the past, we have discussed performance implications of using reflection - sorting was the part that was most affected, as values have to be read and compared quite often. I had a couple of ideas in mind that I now have implemented in #1081, which speeds up sorting (and filtering) by a lot - it's getting much closer to native speed. As far as I see it, we can use this code now for both sides (inside the gadget and inside kubectl gadget) and further reduce the code needed. If you need any help with that, please let me know.

burak-ok:
Will adapt to use the columns sort as soon as possible.

Another question: Is the sorting in `tracer.go` of the deployed gadgets really necessary? Currently the statistics are sorted once on the gadget and another time on `kubectl-gadget`.
Is the output of these `top` gadgets used anywhere else where the consumer does not sort it themself?

flyth:
> Will adapt to use the columns sort as soon as possible.
> 
> 
> 
> Another question: Is the sorting in `tracer.go` of the deployed gadgets really necessary? Currently the statistics are sorted once on the gadget and another time on `kubectl-gadget`.
> 
> Is the output of these `top` gadgets used anywhere else where the consumer does not sort it themself?

The output is limited to a number of entries per node (and later on in total) - that's why we need sorting in both places.

burak-ok:
Prepended commits which uses the `sort` method of the `colums` library. This way it was easier to generalize.

### Notes/Questions
1. I can't seem to totally generalize the `Callback` method since i have no type information about the `Event`. When i try add that via generics i get errors when unmarshalling the JSON
2. I can't seem to totally generalize the `PrintStats` method since the `top ebpf` gadget adds node information to the stats and the others don't

flyth (commented):
A couple more comments from a second pass. It looks really good already! I'll look into the JSON unmarshalling.

pkg/gadget-collection/gadgets/top/block-io/gadget.go
```diff
@@ -103,35 +107,46 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
...
+				colToTest := col
+				if len(col) > 0 && col[0] == '-' {
+					colToTest = colToTest[1:]
+				}
+				_, ok := cols.GetColumn(colToTest)
+				if !ok {
+					trace.Status.OperationError = fmt.Sprintf("%q is not valid for %q", val, top.SortByParam)
+					return
+				}
+				sortBy[i] = col
```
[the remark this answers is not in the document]
burak-ok: Well it looks like that because it is the same :D
Was the first version of this check before i found `VerifyColumnNames`. Forgot to update it in the `pkg/gadget-collection` 

Will change

pkg/gadget-collection/gadgets/top/block-io/gadget.go
```diff
@@ -48,15 +50,17 @@ func NewFactory() gadgets.TraceFactory {
...
 The following parameters are supported:
  - %s: Output interval, in seconds. (default %d)
  - %s: Maximum rows to print. (default %d)
  - %s: The field to sort the results by (%s). (default %s)`
-	return fmt.Sprintf(t, types.IntervalParam, types.IntervalDefault,
-		types.MaxRowsParam, types.MaxRowsDefault,
-		types.SortByParam, strings.Join(types.SortBySlice, ","), types.SortByDefault)
+	return fmt.Sprintf(t, top.IntervalParam, top.IntervalDefault,
+		top.MaxRowsParam, top.MaxRowsDefault,
+		top.SortByParam, strings.Join(cols.GetColumnNames(), ","), strings.Join(types.SortByDefault, ","))
```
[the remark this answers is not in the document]
burak-ok: I also think that would be better. Should be the done in kind of the same way as in the `kubectl-gadget` `top` gadgets in this PR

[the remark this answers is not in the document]
blanquicet: I consider it is useful to have this info here because it is used to generate [this doc](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadgets/filetop.md) which wouuld be read by people using directly the CRD and not passing through the CLI. Even though we were thinking to unify it with [the guides](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/docs/guides), until then, I would suggest to keep it updated (BTW, it looks like some gadgets are missing 😅).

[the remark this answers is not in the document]
burak-ok: Yes, they are missing :D
The CI doesnt complain when new files are generated by `make manifests generate generate-documentation`.
Intentional or not?

cmd/kubectl-gadget/top/top.go
```diff
@@ -47,11 +59,153 @@ func addCommonTopFlags(
...
+	defer g.Unlock()
+
+	stats, err := g.parser.UnmarshalStats(line, node)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s", err.Error())
+		return
+	}
+
+	g.nodeStats[node] = stats
+}
```
flyth, line 192: This could be replaced by a generic Events type like so:
```suggestion
type Event[T any] struct {
	Error string `json:"error,omitempty"`
	Stats []*T   `json:"stats,omitempty"`
}

func (g *TopGadget[Stats]) Callback(line string, node string) {
	var ev Event[Stats]

	err := json.Unmarshal([]byte(line), &ev)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Error: %s", commonutils.WrapInErrUnmarshalOutput(err, line))
		return
	}
	if ev.Error != "" {
		fmt.Fprintf(os.Stderr, "Error: failed on node %q: %s", node, ev.Error)
	}

	g.Lock()
	defer g.Unlock()

	g.nodeStats[node] = ev.Stats
}
```

burak-ok: Did not thought about making the `Event` generic.

Would it be even better then if i generalize all `Event` structs of the `top` gadgets and put one definition of the Event struct in `pkg/gadgets/top/top.go`? Maybe even calling it `TopEvent`?

Then it would be only defined once and we could use it in this file as well

cmd/kubectl-gadget/top/top.go
```diff
@@ -47,11 +59,153 @@ func addCommonTopFlags(
...
+		if err != nil {
+			return commonutils.WrapInErrInvalidArg("<interval>",
+				fmt.Errorf("%q is not a valid value", args[0]))
+		}
+	} else {
+		g.commonTopFlags.OutputInterval = top.IntervalDefault
+	}
+
+	sortByColumns := strings.Split(g.commonTopFlags.SortBy, ",")
+	cols := columns.MustCreateColumns[Stats]()
```
burak-ok, line 108: @flyth do you have any idea if i could use the `GetColumns` of that specific `types.go` here as well? Or how it could be solved when we have custom columns as you mentioned

cmd/kubectl-gadget/top/top.go
flyth: Yeah, I would remove the specific ones and replace them with this generic one, too.

cmd/kubectl-gadget/top/top.go
flyth: You already have a field `colMap         columns.ColumnMap[Stats]` in your `TopGadget` struct - that could be used here as well, right? As long as the initialization happens in the gadget specific code (where you can call a types.GetColumns()), it should work.

But to answer your question: I don't think it's possible to call that function generically without an interface. I would love to see an interface for all gadgets serving all kinds of metadata (like name, description, possible parameters, columns instance etc.) Hopefully we can add that later as it would help a lot with other/third party interfaces.

burak-ok: Yes, i do have that in my `TopGadget`. Actually did not notice it until you mentioned it.

cmd/kubectl-gadget/top/ebpf.go
```diff
@@ -86,168 +73,36 @@ func newEbpfCmd() *cobra.Command {
...
-					fmt.Errorf("this gadget cannot filter by pod name"))
-			}
-			if commonFlags.Containername != "" {
-				return commonutils.WrapInErrInvalidArg("--containername / -c",
-					fmt.Errorf("this gadget cannot filter by container name"))
-			}
-			if len(commonFlags.Labels) > 0 {
-				return commonutils.WrapInErrInvalidArg("--selector / -l",
-					fmt.Errorf("this gadget cannot filter by selector"))
-			}
```
flyth, line 169: These checks should probably not be removed.

cmd/kubectl-gadget/top/top.go
```diff
@@ -47,11 +60,157 @@ func addCommonTopFlags(
...
+		}
+	} else {
+		g.commonTopFlags.OutputInterval = top.IntervalDefault
+	}
+
+	sortByColumns := strings.Split(g.commonTopFlags.SortBy, ",")
+	_, invalidCols := g.colMap.VerifyColumnNames(sortByColumns)
+
+	if len(invalidCols) > 0 {
+		return commonutils.WrapInErrInvalidArg("--sort", fmt.Errorf("\"%v\" is/are not a recognized column(s) to sort by", strings.Join(invalidCols, ",")))
```
flyth, line 109: Nit: Maybe use "invalid columns to sort by: %q".

flyth (commented):
This removes so much duplicated code, I really like it! I'm excited to see how much more will be saved when the printing also moves to the columns lib in a future PR.

One thing since you're touching many files in gadgets/top/*/tracer already: the imports are mostly in the wrong order (local before external), so it'd be great if you could reorder those.

Also thanks for streamlining the `top/ebpf` gadget regarding the node info!

pkg/gadgets/top/ebpf/types/types.go
```diff
@@ -39,94 +37,28 @@ const (
 const (
 	MaxRowsDefault  = 20
 	IntervalDefault = 1
```
blanquicet, line 39: They can be removed.

cmd/kubectl-gadget/top/top.go
```diff
@@ -15,18 +15,31 @@
...
+
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/utils"
+	gadgetv1alpha1 "github.com/inspektor-gadget/inspektor-gadget/pkg/apis/gadget/v1alpha1"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top"
 )
 
 type CommonTopFlags struct {
+	commonFlags utils.CommonFlags
```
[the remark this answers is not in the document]
burak-ok: The `TraceConfig` wants a pointer to a  `utils.CommonFlags` in line `125` of file `top/top.go`. Therefore i have to declare it as a member, right?
Or is there any other way?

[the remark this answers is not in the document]
flyth: You should still be able to get a pointer using @commonTopFlags.CommonFlags.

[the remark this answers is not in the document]
blanquicet: I meant that you could [embed it](https://go.dev/doc/effective_go#embedding):

```golang
type CommonTopFlags struct {
	*utils.CommonFlags
	...
}
```

In this way, instead of `g.commonTopFlags.commonFlags.Timeout`, you could use `g.commonTopFlags.Timeout`, which looks better, IMHO 🙂.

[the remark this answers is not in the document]
blanquicet: And yes, as Mick said, you can still get it.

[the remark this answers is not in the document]
burak-ok: Seems i tried with a lowercase `commonFlags` instead of a uppercase `CommonFlags`. Indeed you can still get it 😅. 
I also like the embedded access way more 👍

cmd/kubectl-gadget/top/top.go
```diff
@@ -47,11 +60,157 @@ func addCommonTopFlags(
...
+	_, invalidCols := g.colMap.VerifyColumnNames(sortByColumns)
+
+	if len(invalidCols) > 0 {
+		return commonutils.WrapInErrInvalidArg("--sort", fmt.Errorf("\"%v\" is/are not a recognized column(s) to sort by", strings.Join(invalidCols, ",")))
+	}
+	g.commonTopFlags.ParsedSortBy = sortByColumns
+
+	if g.params == nil {
+		g.params = make(map[string]string)
+	}
```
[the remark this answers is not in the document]
burak-ok: When a gadget does not prepopulate the `params` it is `nil` and `go` panics on the next line.
`panic: assignment to entry in nil map`

blanquicet (approved):
LGTM! Great job!

cmd/kubectl-gadget/top/tcp.go
```diff
@@ -15,63 +15,53 @@
...
+					"sent",
+					"received",
+				},
 			},
 		},
 	}
 
+	var (
+		FilteredPid uint
+		Family      uint
```
mauriciovasquezbernal, line 62: nit: Usually those variable start with lower case.

```suggestion
		filteredPid uint
		family      uint
```

cmd/kubectl-gadget/top/file.go
```diff
@@ -15,62 +15,50 @@
...
+					"r_kb",
+					"w_kb",
+					"t",
+					"file",
+				},
 			},
 		},
 	}
 
+	var AllFiles bool
```
mauriciovasquezbernal, line 60: I think it's good to have a `FileFlags` struct as it was before. It would help to generalize it once we want to support it on local-gadget. (I'm find with removing the struct now and adding it back when needed)

mauriciovasquezbernal (approved):
Great work. I found some minor issues but IMO we can merge and fix later on if those delay this to be merged. I'm looking forward to see all of these gadgets using the columns library.

mauriciovasquezbernal:
Nit not related to coding: Present tense is preferred for the commit titles -> https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181.

PR #1064, opened 2022-10-20
integrationtest/TestTcptop: Validate json output

burak-ok:
### integrationtest/TestTcptop: Validate json output

Uses the json output in order to validate as much information as possible from the `top tcp` gadget

### DIFF
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -93,15 +93,19 @@ out:
 	return nil
 }
 
+func BuildCommonData(namespace string) eventtypes.CommonData {
+	return eventtypes.CommonData{
+		Namespace: namespace,
+		// Pod and Container name are defined by BusyboxPodCommand.
+		Pod:       "test-pod",
+		Container: "test-pod",
+		// TODO: Include the Node
+	}
+}
+
 func BuildBaseEvent(namespace string) eventtypes.Event {
 	return eventtypes.Event{
-		Type: eventtypes.NORMAL,
-		CommonData: eventtypes.CommonData{
-			Namespace: namespace,
-			// Pod and Container name are defined by BusyboxPodCommand.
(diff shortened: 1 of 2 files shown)


integration/helpers.go
```diff
@@ -105,3 +105,13 @@ func BuildBaseEvent(namespace string) eventtypes.Event {
 		},
 	}
 }
+
+func BuildCommonData(namespace string) eventtypes.CommonData {
```
blanquicet, line 109: Please update `BuildBaseEvent` to use `BuildCommonData`.

integration/helpers.go
```diff
@@ -93,15 +93,19 @@ out:
...
-			Namespace: namespace,
-			// Pod and Container name are defined by BusyboxPodCommand.
-			Pod:       "test-pod",
-			Container: "test-pod",
-			// TODO: Include the Node
-		},
+		Type:       eventtypes.NORMAL,
+		CommonData: BuildCommonData(namespace),
 	}
 }
```
blanquicet, line 111: Hey @mauriciovasquezbernal, I will create an issue to organize these structures so that we avoid having the top gadgets defining their own [Error](https://github.com/inspektor-gadget/inspektor-gadget/blob/0ca2f8466f8f27d5fd627fb20c6c91cfc78ad931/pkg/gadgets/top/tcp/types/types.go#L86) field and error mechanism, instead of using [these](https://github.com/inspektor-gadget/inspektor-gadget/blob/0ca2f8466f8f27d5fd627fb20c6c91cfc78ad931/pkg/types/types.go#L85-L89).

blanquicet (approved):
LGTM!

integration/helpers.go
mauriciovasquezbernal: Sure, makes sense.

Issue #1000, opened 2022-09-29
gadgets/dns: Enhance gadget to provide more information


mauriciovasquezbernal:
The DNS trace gadget provides information about DNS requests. It provides the following:

- node, namespace, pod: pod where the request is coming from  / to 
- pktType: direction of the request (ONGOING / INCOMING) 
- qtype: type of request 
- name: host of the request

There is more information we could add to this:
- nameserver: DNS server where the request is going to 
- check also DNS replies to understand if a request worked fine

PR #1071, opened 2022-10-24
dns: show nameserver in DNS queries and responses

alban:
# dns: show nameserver in DNS queries and responses

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1000

## How to use in Kubernetes

```
$ kubectl gadget trace dns
NODE          NAMESPACE     POD       QR NAMESERVER      TYPE      QTYPE      NAME                        
minikube      default       test      Q  8.8.4.4         OUTGOING  A          magic-1-2-3-4.nip.io.       
minikube      default       test      R  8.8.4.4         HOST      A          magic-1-2-3-4.nip.io.       
```

```
$ kubectl run -ti --restart=Never --rm \
        --image docker.io/tutum/dnsutils test \
        -- dig @8.8.4.4 magic-1-2-3-4.nip.io
```

## How to use with local-gadget

```
$ sudo ./local-gadget interactive -r docker
» create dns trace1
State: Started
» stream trace1 --follow
{"node":"local","namespace":"default","pod":"determined_allen","type":"debug","message":"tracer attached"}
{"node":"local","namespace":"default","pod":"determined_allen","type":"normal","qr":"Q","nameserver":"8.8.4.4","pktType":"OUTGOING","qtype":"A","name":"magic-1-2-3-4.nip.io."}
{"node":"local","namespace":"default","pod":"determined_allen","type":"normal","qr":"R","nameserver":"8.8.4.4","pktType":"HOST","qtype":"A","name":"magic-1-2-3-4.nip.io."}
{"node":"local","namespace":"default","pod":"determined_allen","type":"debug","message":"tracer detached"}
```

```
$ docker run -ti --rm docker.io/tutum/dnsutils dig @8.8.4.4 magic-1-2-3-4.nip.io
```

## Testing done

```
$ go test ./pkg/gadgets/trace/dns/...
ok  	github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/dns/tracer	0.009s
?   	github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/dns/types	[no test files]
```

```
$ make local-gadget-tests LOCAL_GADGET_TESTS_PARAMS='-test.run DNS'
...
--- PASS: TestDNS (6.01s)
...
PASS
```

```
$ INTEGRATION_TESTS_PARAMS="-run TestDns -v -no-deploy-ig -no-deploy-spo" \
    make integration-tests
...
--- PASS: TestDns (19.45s)
```


pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -104,19 +105,26 @@ int ig_trace_dns(struct __sk_buff *skb)
...
+		return 0;
 
 	struct event_t event = {0,};
-	if (len > 0)
-		bpf_skb_load_bytes(skb, DNS_OFF + sizeof(struct dnshdr), event.name, len);
+	event.af = AF_INET;
+	event.daddr_v4 = load_word(skb, ETH_HLEN + offsetof(struct iphdr, daddr));
+	event.saddr_v4 = load_word(skb, ETH_HLEN + offsetof(struct iphdr, saddr));
+	// load_word converts from network to host endianness. Convert back to
+	// network endianness.
```
blanquicet, line 110: We need to have it in network byte order because we are using `inet_ntop` to transform it into text form. right?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -33,7 +33,41 @@ import (
...
+//}
+//
+//static char *get_src_addr(const struct event_t *ev) {
+//	if (ev->af == AF_INET)
+//		return addr_str(&ev->saddr_v4, ev->af);
+//	else if (ev->af == AF_INET6)
+//		return addr_str(&ev->saddr_v6, ev->af);
+//	else
+//		return NULL;
+//}
```
blanquicet, line 61: nit: It is not used in this commit.

integration/inspektor-gadget/integration_test.go
```diff
@@ -488,7 +506,9 @@ func TestDns(t *testing.T) {
 	commands := []*Command{
 		CreateTestNamespaceCommand(ns),
 		dnsCmd,
-		BusyboxPodRepeatCommand(ns, "nslookup inspektor-gadget.io 8.8.4.4"),
+		BusyboxPodRepeatCommand(ns,
+			"nslookup -type=a inspektor-gadget.io. 8.8.4.4 ;"+
+				"nslookup -type=aaaa inspektor-gadget.io. 8.8.4.4"),
```
blanquicet, line 511: Why did you need to split the request, doesn't `nslookup` perform both requests by default?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -295,16 +295,24 @@ func parseLabelSequence(sample []byte) (ret string) {
 	return ret
 }
 
-func parseDNSEvent(rawSample []byte) (nameserver string, ret string, pktType string, qType string) {
+func parseDNSEvent(rawSample []byte) (qr string, nameserver string, ret string, pktType string, qType string) {
```
blanquicet, line 298: What about making `parseDNSEvent` return directly `types.Event`? Or use another struct for this?

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -71,20 +71,14 @@ int ig_trace_dns(struct __sk_buff *skb)
 	union dnsflags flags;
 	flags.flags = load_half(skb, DNS_OFF + offsetof(struct dnshdr, flags));
```
blanquicet, line 72: `Skip non DNS Query packets` should be removed.

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -22,6 +22,7 @@ import (
 type Event struct {
 	eventtypes.Event
 
+	Qr         string `json:"qr,omitempty" column:"qr,width:2,fixed"`
```
blanquicet, line 25: We should use an enum for this so that users that inspect this file, easily understand that `Qr` can be either `R` or `Q`.

pkg/local-gadget-manager/local-gadget-manager_test.go
```diff
@@ -461,6 +462,20 @@ func TestDNS(t *testing.T) {
...
 
+	// check dns response is traced
+	result = <-ch
+	event = dnstypes.Event{}
+	if err := json.Unmarshal([]byte(result), &event); err != nil {
+		t.Fatalf("failed to unmarshal json: %s", err)
+	}
+
+	expectedEvent.PktType = "HOST"
+	expectedEvent.Qr = "R"
```
blanquicet, line 473: I think this should at least verify the nameserver as, unlike queries, it is read from a different field for reponses.

integration/inspektor-gadget/integration_test.go
```diff
@@ -458,16 +458,34 @@ func TestDns(t *testing.T) {
...
 					DNSName:    "inspektor-gadget.io.",
 					QType:      "A",
 				},
 				{
 					Event:      BuildBaseEvent(ns),
+					Qr:         "R",
+					Nameserver: "8.8.4.4",
+					PktType:    "HOST",
+					DNSName:    "inspektor-gadget.io.",
+					QType:      "A",
```
blanquicet, line 473: I don't know how useful it is to have a response entry without the actual list of responses. However, I suppose it is an intermediate step, right?

Maybe this link could be useful for parsing the response: https://git.kernel.org/pub/scm/network/connman/connman.git/tree/src/dnsproxy.c#n972

docs/guides/trace/dns.md
```diff
@@ -21,37 +21,29 @@ Start the dns gadget:
...
-minikube         demo             mypod            OUTGOING  AAAA       e15316.a.akamaiedge.net.
+NODE                          NAMESPACE                     POD                           QR NAMESERVER      TYPE      QTYPE      NAME
+minikube                      demo                          mypod                         Q  8.8.4.4         OUTGOING  A          inspektor-gadget.io.
+minikube                      demo                          mypod                         R  8.8.4.4         HOST      A          inspektor-gadget.io.
+minikube                      demo                          mypod                         Q  8.8.4.4         OUTGOING  AAAA       inspektor-gadget.io.
+minikube                      demo                          mypod                         R  8.8.4.4         HOST      AAAA       inspektor-gadget.io.
+minikube                      demo                          mypod                         Q  8.8.4.4         OUTGOING  MX         inspektor-gadget.io.
+minikube                      demo                          mypod                         R  8.8.4.4         HOST      MX         inspektor-gadget.io.
+
```
blanquicet, line 46:
```suggestion
```

docs/guides/trace/dns.md
```diff
@@ -2,13 +2,13 @@
 title: 'Using trace dns'
 weight: 20
 description: >
-  Trace DNS requests.
+  Trace DNS queries and responses.
 ---
 
 ![Screencast of the trace dns gadget](dns.gif)
```
blanquicet, line 8: I don't know. I think using 80 as the width is too small.

integration/inspektor-gadget/integration_test.go
blanquicet: I wonder if the DNS ID could be useful for users to associate a DNS response to a query. Maybe we could add it but hidden by default?

blanquicet (commented):
Awesome @alban! I have some comments, but it already looks great 🤩 !!

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -104,19 +105,26 @@ int ig_trace_dns(struct __sk_buff *skb)
...
+	bpf_skb_load_bytes(skb, DNS_OFF + sizeof(struct dnshdr), event.name, len);
 
 	event.pkt_type = skb->pkt_type;
 
 	// Read QTYPE right after the QNAME
 	// https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.2
 	event.qtype = load_half(skb, DNS_OFF + sizeof(struct dnshdr) + len + 1);
 
-	// TODO: we should not send the event when len == 0. But the verifier
-	// won't let us.
```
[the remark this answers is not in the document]
alban: I don't know. I wonder if it is some older version of clag/llvm... But it seems to work fine now.

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -33,7 +33,41 @@ import (
...
+//	size_t size = af == AF_INET ? INET_ADDRSTRLEN : INET6_ADDRSTRLEN;
+//	char *str;
+//
+//	str = malloc(size);
+//	if (!str)
+//		return NULL;
+//
+//	inet_ntop(af, addr, str, size);
+//
+//	return str;
```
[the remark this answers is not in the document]
alban: It is the same code style in other gadgets:
```
pkg/gadgets/top/tcp/tracer/tracer.go:// inet_ntop(ip_type, addr, ip, size);
pkg/gadgets/trace/bind/tracer/tracer.go://      inet_ntop(ip_type, &event->addr, ip, size);
pkg/gadgets/trace/tcp/tracer/tracer.go://       inet_ntop(af, addr, str, size);
pkg/gadgets/trace/tcpconnect/tracer/tracer.go://        inet_ntop(af, addr, str, size);
```

I think @flyth was trying to remove cgo but meanwhile, I think it is better to keep the code consistent.

integration/inspektor-gadget/integration_test.go
alban: To guarantee the ordering of the 4 UDP packets: the busybox implementation of nslookup sends both requests without waiting for the responses. So depending how fast is the network, the first response could arrive between the two requests.

The nslookup implementation from bind-utils does not do that.

pkg/gadgets/trace/dns/tracer/tracer.go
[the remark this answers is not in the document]
blanquicet: Yeah. In fact, I was thinking that we can do that in Go, like in [the socket-collector gadget](https://github.com/inspektor-gadget/inspektor-gadget/blob/89d68a4a6da56fefaa2637015b84b0b09958f73b/pkg/gadgets/snapshot/socket/tracer/tracer.go#L33-L42).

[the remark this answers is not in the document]
flyth: Yeah, I've already converted everything to Go for the existing modules, I need to fix a couple of commits still (the repo renaming mixed up my branches/commits quite a bit) before I push it. For me it's okay to have it in cgo here and I'll port the parts again, then.

blanquicet:
BTW, could you please update [the example](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/examples/gadgets/basic/trace/dns)? At least to print "query" or "response" based on `Qr`.

pkg/local-gadget-manager/local-gadget-manager_test.go
alban: It is verifying that: I reuse the `expectedEvent` variable which is initialized that way:
```
	// select a nameserver that is not symmetrical with endianness
	nameserver := "8.8.4.4"
...
	expectedEvent = dnstypes.Event{
...
		Nameserver: nameserver,

```

integration/inspektor-gadget/integration_test.go
alban: Yes, showing the responses would be good. But I was shocked to see how the responses are encoded with offsets relative to the beginning of the header. Because of this, it cannot easily be done from userspace without access to the full DNS packet. And parsing this in eBPF seems potentially difficult with the verifier, so we can keep it for another PR.

Good idea about the DNS ID, I will add it.

pkg/local-gadget-manager/local-gadget-manager_test.go
blanquicet: Ops, I missed that.

docs/guides/trace/dns.md
alban: I changed it to 80 for consistency in https://github.com/inspektor-gadget/inspektor-gadget/blob/89d68a4a6da56fefaa2637015b84b0b09958f73b/tools/demos/Makefile#L26-L30

I would like to have an option `kubectl gadget -o short` which looks nice on 80 columns by removing/reducing some fields...

pkg/gadgets/trace/dns/types/dns.go
alban: Done!

blanquicet (commented):
I tested it, but something is not working with the columns format. The following command doesn't produce any output:

```bash
./kubectl-gadget trace dns -A
```

But using `-o json` works properly.

We must add unit-tests for all gadgets to verify that transformation from JSON to columns in the CLI is working properly.

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -19,12 +19,22 @@ import (
...
+)
+
 type Event struct {
 	eventtypes.Event
 
-	PktType string `json:"pktType,omitempty" column:"type,minWidth:7,maxWidth:9"`
-	QType   string `json:"qtype,omitempty" column:"qtype,minWidth:5,maxWidth:10"`
-	DNSName string `json:"name,omitempty" column:"name,width:30"`
+	ID         string     `json:"id,omitempty" column:"id,width:2,fixed,hide"`
+	Qr         DNSPktType `json:"qr,omitempty" column:"qr,width:2,fixed"`
```
blanquicet, line 33: > I tested it, but something is not working with the columns format. The following command doesn't produce any output:
> 
> ```bash
> ./kubectl-gadget trace dns -A
> ```
> 
> But using `-o json` works properly.
> 

The issue seems related to the `DNSPktType` type in the columns library. Changing this field to use the `string` type makes things work again. cc @flyth

examples/gadgets/basic/trace/dns/README.md
```diff
@@ -23,28 +23,19 @@ Then, perform some DNS requests:
 
 
 ```bash
-$ nslookup microsoft.com
-$ nslookup google.com
+$ nslookup -querytype=a microsoft.com.
+$ nslookup -querytype=a google.com.
```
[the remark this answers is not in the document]
alban: Without the final dot, it will try all search options:

```
$ grep search /etc/resolv.conf 
search default.svc.cluster.local svc.cluster.local cluster.local
```
So it would search
* microsoft.com.default.svc.cluster.local.
* microsoft.com.svc.cluster.local.
* microsoft.com.cluster.local.

I think it clutters the documentation.

docs/guides/trace/dns.md
flyth: Sounds like responsive design on a CLI :) We could add something like `hide:sm` to hide columns by default when the screen width is smaller than a threshold value like, say, 80 - and that can be manually triggered by `-short` and `-wide`. I'm not too sure how much "magic" is welcome on a CLI, though...

examples/gadgets/basic/trace/dns/README.md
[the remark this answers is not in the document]
blanquicet: Agree

blanquicet (approved):
LGTM! Awesome @alban!!

PR #1077, opened 2022-10-25
integrationtest/TestFiletop: Validate json output

burak-ok:
### integrationtest/TestFiletop: Validate json output

Uses the json output in order to validate as much information as possible from the `top file` gadget

### Question/Comment
General comment/question about the `FileType` member variable of the `Stats struct` in `top/types.go`.
Should the type of `FileType` be a char/enum, since in the eBPF program only `R`,  `S` or `O` will get assigned.
And where can we put this documentation. Is it right when it is in the same line for such a small comment?

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -28,6 +28,7 @@ import (
 
 	. "github.com/inspektor-gadget/inspektor-gadget/integration"
 	ebpftopTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/ebpf/types"
+	filetopTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/file/types"
 	tcptopTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/tcp/types"
 	bindTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/types"
 	capabilitiesTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/capabilities/types"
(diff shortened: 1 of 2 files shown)


pkg/gadgets/top/file/types/types.go
```diff
@@ -90,7 +90,7 @@ type Stats struct {
 	MountNsID  uint64 `json:"mountnsid,omitempty"`
 	Filename   string `json:"filename,omitempty"`
 	Comm       string `json:"comm,omitempty"`
-	FileType   byte   `json:"fileType,omitempty"`
+	FileType   byte   `json:"fileType,omitempty"` // R = Regular File, S = Socket, O = Other
```
blanquicet, line 93: >Is it right when it is in the same line for such a small comment?
>
IMO, it is ok for this PR. However, this struct should be self-descriptive, and we shouldn't need this comment. To solve that, I agree we need to use an `enum` or something similar to restrict the values `FileType` can take, and it should be synch with the BPF program. Of course, you can manage it in another PR 🙂 .

integration/inspektor-gadget/integration_test.go
```diff
@@ -585,10 +586,30 @@ func TestFiletop(t *testing.T) {
...
+			}
+
+			normalize := func(e *filetopTypes.Stats) {
+				e.Node = ""
+				e.Writes = 0
+				e.WriteBytes = 0
+				e.Pid = 0
+				e.Tid = 0
+				e.MountNsID = 0
+				e.Comm = ""
```
blanquicet, line 608: IMO, we could verify that the command is `sh` as far as we continue using [`BusyboxPodRepeatCommand`](https://github.com/inspektor-gadget/inspektor-gadget/blob/7167da66277a05ba3a617180d17e9cdb54407a64/integration/command.go#L507-L509) to start the pod and [`echo`](https://github.com/inspektor-gadget/inspektor-gadget/blob/7167da66277a05ba3a617180d17e9cdb54407a64/integration/inspektor-gadget/integration_test.go#L618) to write into the file.

blanquicet (approved):
LGTM after CI is green!

pkg/gadgets/top/file/types/types.go
burak-ok: Is there a way to do that?
Like defined an enum in a C header and "import" these values?

PR #1080, opened 2022-10-25
integrationtest/TestBiotop: Validate json output

burak-ok:
### integrationtest/TestBiotop: Validate json output

Uses the json output in order to validate as much information as possible from the `top block-io` gadget

### Question

Is the `Operations` always guaranteed to be 2?

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	. "github.com/inspektor-gadget/inspektor-gadget/integration"
+	biotopTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/block-io/types"
 	ebpftopTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/ebpf/types"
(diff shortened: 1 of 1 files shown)


integration/inspektor-gadget/integration_test.go
```diff
@@ -382,10 +383,29 @@ func TestBiotop(t *testing.T) {
...
+		Name:         "StartBiotopGadget",
+		Cmd:          fmt.Sprintf("$KUBECTL_GADGET top block-io -n %s -o json", ns),
+		StartAndStop: true,
+		ExpectedOutputFn: func(output string) error {
+			expectedEntry := &biotopTypes.Stats{
+				CommonData: BuildCommonData(ns),
+				Write:      true,
+				Comm:       "dd",
+				Bytes:      2097152,
+				Operations: 2,
```
blanquicet, line 395: > Is the Operations always guaranteed to be 2?
> 
I don't think so. @eiffel-fl any comment?

burak-ok: Thanks for all the information :)
I'll keep the deviations from the regex to a minimum in future PRs

blanquicet (approved):
LGTM!

PR #1083, opened 2022-10-26
integrationtest/TestSocketCollector: Validate json output

burak-ok:
### integrationtest/TestSocketCollector: Validate json output

Uses the json output in order to validate as much information as possible from the `snapshot socket` gadget

### Question

The returned json of snapshot is in the form of 
```
[
{
"node": "...",
....
},
...
]
```

`ExpectEntriesToMatch` expects it one JSON object per line
```
{"node":"...",...}
{...}
```

The solution in this commit is a kind of hackish way. It converts the output to the "object per line" format.
Optimally there should be 2 different functions for it. But then `ExpectEntriesToMatch` would need to be doubled, right?

`ExpectEntriesToMatch` on its own can't know what kind of json it receives, except it tries to do both. Would this be a good way?


blanquicet:
> Question
>
> The returned json of snapshot is in the form of 
> ...

We already started this discussion [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/946#discussion_r986610126). Please, synch with @mqasimsarfraz about it. You can manage it here in a specific commit, or in a separate PR that would need to be merged before this one, I am ok either ways 🙂.

mqasimsarfraz:
> We already started this discussion [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/946#discussion_r986610126). Please, synch with @mqasimsarfraz about it. You can manage it here in a specific commit, or in a separate PR that would need to be merged before this one. I am ok either way slightly_smiling_face.

I haven't yet looked into adding the helper yet so sounds good to take care of it in this PR. We should combine this with changing [list_containers_test](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/local-gadget/list_containers_test.go#L63) so all the usages are updated.

integration/helpers.go
```diff
@@ -54,6 +54,23 @@ func parseOutput[T any](output string, normalize func(*T)) ([]*T, error) {
 	return ret, nil
 }
 
+func UnpackJsonArray(output string) (string, error) {
+	var objects []map[string]any
```
mqasimsarfraz, line 58: Just for understanding why do we need a map here? I see we are ignoring the keys when merging the objects?

burak-ok: This is done in order to handle any type of input in the form of `[{...},{...}]`. Just an abstract way of telling that we have an array of JSON objects and we done care what they are.
We are going to extract these objects and write them in their own line

The validation will happen of the type will happen later in `parseOutput`.

This is inefficient since we are unmarshalling the JSON twice (and marshalling it an additional time)
That is why I asked if it would be better when we have two functions `ExpectEntriesToMatch` which handle the JSON output differently or how else we could generalize it. Like trying both methods

blanquicet: Yeah, I propose to have two functions. 

The existing one, for which we should be more specific in the description. Maybe something like this:

```golang
// ExpectEntriesToMatch verifies that all the entries in expectedEntries are
// matched by at least one entry in the output (Independent JSON objects).
func ExpectEntriesToMatch[T any](output string, normalize func(*T), expectedEntries ...*T) error
```

And a new one:
```golang
// ExpectEntriesInArrayToMatch verifies that all the entries in expectedEntries
// are matched by at least one entry in the output (JSON array).
func ExpectEntriesInArrayToMatch[T any](output string, normalize func(*T), expectedEntries ...*T) error
```

WDYT?

integration/inspektor-gadget/integration_test.go
```diff
@@ -1151,9 +1152,34 @@ func TestSocketCollector(t *testing.T) {
...
+					Event:         BuildBaseEvent(ns),
+					Protocol:      "TCP",
+					LocalAddress:  "0.0.0.0",
+					LocalPort:     9090,
+					RemoteAddress: "0.0.0.0",
+					RemotePort:    0,
+					Status:        "LISTEN",
+				}
+
+				expectedEntry.Container = ""
```
blanquicet, line 1168:
```suggestion
				// Socket gadget doesn't provide container data yet. See issue #744.
				expectedEntry.Container = ""
```

integration/helpers.go
mqasimsarfraz: I was thinking if we can use something like [this](https://stackoverflow.com/a/55014220/7274380) but that might be bit magic? :smile:  Also, sounds good having two functions, makes it more clear!

burak-ok: That was also my thought :D Magic or not.
Some people like it - others not. But i also think its more beneficial to have it clearer

blanquicet: Mmmm, IMO, it's non that "magic" 😝. However, I consider that having two helpers would make more evident (during the development and reveiwing) a change to the output format of any of the gadgets (array to indentent objects, or viceversa). In genereal, I think it is a good practice to not have generic things on the testing infrastructure, so that one easily realized about (desired or indesired) changes introduced with a PR. I don't know, I have always have this idea, but I maybe it is just me and my "control mania" 😆

burak-ok:
> Looking good now. Either we should adopt the [list_containers_test](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/local-gadget/list_containers_test.go) in the same PR or create an issue for it so we don't lose track of it. What do you say?

Personally i would have made a second PR and waited for this PR to go through first

mqasimsarfraz (approved):
Let's keep it open for a bit to get more reviews otherwise looking good!

integration/helpers.go
```diff
@@ -54,32 +54,56 @@ func parseOutput[T any](output string, normalize func(*T)) ([]*T, error) {
 	return ret, nil
 }
 
-// ExpectAllToMatch verifies that the expectedEntry is matched by all the
-// entries in the output.
-func ExpectAllToMatch[T any](output string, normalize func(*T), expectedEntry *T) error {
-	entries, err := parseOutput(output, normalize)
-	if err != nil {
-		return err
+func parseArrayOutput[T any](output string, normalize func(*T)) ([]*T, error) {
```
blanquicet, line 57: I suggest making the names more descriptive. Something like this, maybe with one line of description:

```golang
func parseMultiJSONOutput[T any](output string, normalize func(*T)) ([]*T, error)
```

```golang
func parseJSONArrayOutput[T any](output string, normalize func(*T)) ([]*T, error) {
```

blanquicet (approved):
LGTM, just one comment.

mauriciovasquezbernal (approved):
LGTM!

Issue #1028, opened 2022-10-10
[RFE] local-gadget: Add `--watch` option to the `list-containers` command


blanquicet:
## Current situation

Currently, the `list-containers` command shows the containers currently running in the system. However, there is no option to know when a given container is created/deleted.

## Impact

It is not possible to debug the creation/deletion of containers using `local-gadget`.

## Ideal future situation

`local-gadget` should provide an option like `--watch` to trace the containers events:

```bash
$ sudo local-gadget list-containers --watch
TIME                 EVENT   RUNTIME    ID                                                               NAME
2022-10-10T18:33:04Z CREATED containerd 942ed1e839818822794247a5b3f55356c33cb72c6332fa66c008e4a99b3aec4e coredns
2022-10-10T18:33:04Z CREATED containerd c8a65cf1c8650cf25427774494d45f00f81b243e3842142d41d4417c7892fc0d etcd
2022-10-10T18:33:04Z CREATED containerd ac36877bb8acf26ebd5186d747b83d2c12304a5cbc16a73e09e6beebbc424960 gadget
2022-10-10T18:33:04Z CREATED containerd e03544daba437aecbe23efc4df791d9ee1c2a07a5a121f6d6686f33b70a56454 kindnet-cni
2022-10-10T18:33:04Z CREATED containerd eaeef87de99228e97f87f06bcec023d54242363cf11a22bb76a78acc193ddc78 kube-apiserver
2022-10-10T18:33:04Z CREATED containerd 598e2c3d1a58524c08a7c5e45c1b38424b7c2d03789528c85b313788e9904197 kube-controller-manager
2022-10-10T18:33:04Z CREATED containerd c18d844a8c0b57024205687595dc921ba99e9b2e0880a3f9dac6ee8afadf2949 kube-proxy
2022-10-10T18:33:04Z CREATED containerd 53d749a30929eb76b4a53857001431b5579d916101b507b05098ce6c79beee7e kube-scheduler
2022-10-10T18:33:04Z CREATED containerd f032e55c72bc0a88dd72fa6d17c0f5ead70037dd97b4462a0b85416e36b3ac16 storage-provisioner
# Create a container in another terminal
2022-10-10T18:34:00Z CREATED containerd b8e94b0e5297cb6d3f8a86b01c4fd8221628cadd9156036ca81e72accd95d01d myContainer
# Delete a container in another terminal
[...]

PR #1084, opened 2022-10-26
local-gadget: fix tracing new crio containers

### LINKED ISSUE #1018: local-gadget: Trace events not shown for new cri-o containers started after trace command
Currently, if you start a trace on a node e.g `local-gadget trace bind`, all the cri-o containers started after the command aren't traced and no trace events are printed. Terminating and starting the `local-gadget` again fixes the issue. This happens only on `CRI-O` based environment.

## Impact

We will miss events from the containers started after the trace. It only impacts CRI-O based environments and local-gadget.

1. Setup a minikube with cri-o using: `make -C integration/local-gadget setup-cri-o`
2. Ensure you don't hit [CAP apply](https://github.com/kubernetes/minikube/issues/12928#issuecomment-1245458044) issue. 
3. Copy the local-gadget to the container using: `docker cp local-gadget minikube-cri-o:/bin/local-gadget`
4. Start the the tracing using: `docker exec -it minikube-cri-o local-gadget trace bind`
5. In other terminal start the test container and try some socket binding
```
kubectl run test-pod --image busybox:latest sleep inf
kubectl exec -ti test-pod -- nc -l -p 4242 -w 1
```
6.  No events are shown by trace started by 4. 

## Expected behavior

Trace events should be printed.

## Additional information

Initial finding suggest adding runc path for cri-o to runcPaths might work but it needs to be investigated.

mqasimsarfraz:
`runcfanotify` doesn't work for `cri-o` containers. The idea here is to adopt it to support tracing `cri-o` containers. 

Fixes #1018
Related #1028

### Idea
During `runc` notification container creation is blocked on `pidfile` so enriching the container using the container runtime won't work because we hit `Runtime enricher (cri-o): failed to get container: container "2899fb3ce898a2ed1d4426b0cb6df81ee72cd630ce0fb9e10eb53f04d5eb4e9b" not found`. In order to avoid this we are using annotations fetched via `OCIConfig` to enrich containers. 

### Usage
```
$ docker exec -it minikube-cri-o bash
$ local-gadget list-containers --watch 
# Create a container in another terminal e.g kubectl run test-pod --image busybox:latest sleep inf
...
cri-o      a0ddd82147f5536d38c449f39edff4151ae0aa39b9653045449302915c85… test-pod
...
```
JSON mode should also work here. Before this PR this will only works for `docker/containerd` but now it should work for `cri-o` as well.

### Future work
Consider using annotations for other runtimes or other information like `cgroups` etc. Having said that, `containerd` has some annotations but `docker` has none.

### DIFF
--- a/cmd/local-gadget/containers/containers.go
+++ b/cmd/local-gadget/containers/containers.go
@@ -17,6 +17,9 @@ package containers
 import (
 	"encoding/json"
 	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/spf13/cobra"
 
@@ -26,8 +29,11 @@ import (
 	localgadgetmanager "github.com/inspektor-gadget/inspektor-gadget/pkg/local-gadget-manager"
 )
 
+const localGadgetSubKey = "local-gadget-key"
+
 func NewListContainersCmd() *cobra.Command {
 	var commonFlags utils.CommonFlags
+	var optionWatch bool
 
 	cmd := &cobra.Command{
 		Use:   "list-containers",
@@ -44,31 +50,87 @@ func NewListContainersCmd() *cobra.Command {
 				return commonutils.WrapInErrParserCreate(err)
 			}
 
-			containers := localGadgetManager.GetContainersBySelector(&containercollection.ContainerSelector{
-				Name: commonFlags.Containername,
-			})
+			containers := localGadgetManager.ContainerCollection.Subscribe(
+				localGadgetSubKey,
+				containercollection.ContainerSelector{
+					Name: commonFlags.Containername,
+				},
+				func(event containercollection.PubSubEvent) {
+					switch event.Type {
+					case containercollection.EventTypeAddContainer:
+						if err = printContainer(parser, commonFlags, event.Container); err != nil {
+							fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
+						}
+					}
+				})
+			defer localGadgetManager.ContainerCollection.Unsubscribe(localGadgetSubKey)
 
 			parser.Sort(containers, []string{"runtime", "name"})
-
-			switch commonFlags.OutputMode {
-			case commonutils.OutputModeJSON:
-				b, err := json.MarshalIndent(containers, "", "  ")
-				if err != nil {
-					return commonutils.WrapInErrMarshalOutput(err)
+			if optionWatch {
+				if commonFlags.OutputMode != commonutils.OutputModeJSON {
+					fmt.Println(parser.BuildColumnsHeader())
+				}
+				for _, container := range containers {
+					if err = printContainer(parser, commonFlags, container); err != nil {
+						return err
+					}
 				}
 
-				fmt.Printf("%s\n", b)
-			case commonutils.OutputModeColumns:
-				fallthrough
-			case commonutils.OutputModeCustomColumns:
-				fmt.Println(parser.TransformIntoTable(containers))
+				stop := make(chan os.Signal, 1)
+				signal.Notify(stop, syscall.SIGINT)
+				<-stop
+			} else {
+				if err = printContainers(parser, commonFlags, containers); err != nil {
+					return err
+				}
 			}
 
 			return nil
 		},
 	}
 
+	cmd.Flags().BoolVarP(
+		&optionWatch,
+		"watch", "w",
+		false,
+		"After listing the containers, watch for new containers")
+
 	utils.AddCommonFlags(cmd, &commonFlags)
 
 	return cmd
+}
+
+func printContainers(parser *commonutils.GadgetParser[containercollection.Container], commonFlags utils.CommonFlags, containers []*containercollection.Container) error {
+	switch commonFlags.OutputMode {
+	case commonutils.OutputModeJSON:
+		b, err := json.MarshalIndent(containers, "", "  ")
+		if err != nil {
+			return commonutils.WrapInErrMarshalOutput(err)
+		}
+
+		fmt.Printf("%s\n", b)
+	case commonutils.OutputModeColumns:
+		fallthrough
+	case commonutils.OutputModeCustomColumns:
+		fmt.Println(parser.TransformIntoTable(containers))
+	}
+
+	return nil
+}
+
+func printContainer(parser *commonutils.GadgetParser[containercollection.Container], commonFlags utils.CommonFlags, container *containercollection.Container) error {
+	switch commonFlags.OutputMode {
+	case commonutils.OutputModeJSON:
+		b, err := json.MarshalIndent(container, "", "  ")
+		if err != nil {
+			return commonutils.WrapInErrMarshalOutput(err)
+		}
+		fmt.Printf("%s\n", b)
+	case commonutils.OutputModeColumns:
+		fallthrough
+	case commonutils.OutputModeCustomColumns:
+		fmt.Println(parser.TransformIntoColumns(container))
+	}
+
+	return nil
 }
--- a/integration/command.go
+++ b/integration/command.go
@@ -624,3 +624,11 @@ func WaitUntilPodReadyCommand(namespace string, podname string) *Command {
 func WaitUntilTestPodReadyCommand(namespace string) *Command {
 	return WaitUntilPodReadyCommand(namespace, "test-pod")
 }
+
+// SleepForSecondsCommand returns a Command which sleeps for given seconds
+func SleepForSecondsCommand(seconds int) *Command {
+	return &Command{
+		Name: "SleepForSeconds",
+		Cmd:  fmt.Sprintf("sleep %d", seconds),
+	}
+}
--- a/integration/local-gadget/list_containers_test.go
+++ b/integration/local-gadget/list_containers_test.go
@@ -15,9 +15,7 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
-	"reflect"
 	"testing"
 
 	. "github.com/inspektor-gadget/inspektor-gadget/integration"
@@ -60,19 +58,7 @@ func TestListContainers(t *testing.T) {
 				c.PodUID = ""
 			}
 
-			var containers []*containercollection.Container
-			if err := json.Unmarshal([]byte(output), &containers); err != nil {
-				return err
-			}
-
-			for _, gotContainer := range containers {
-				normalize(gotContainer)
-				if reflect.DeepEqual(gotContainer, expectedContainer) {
-					return nil
-				}
-			}
-
-			return fmt.Errorf("output doesn't contain the expected container: %+v", expectedContainer)
+			return ExpectEntriesInArrayToMatch(output, normalize, expectedContainer)
 		},
 	}
 
@@ -143,23 +129,7 @@ spec:
 				c.PodUID = ""
 			}
 
-			var containers []*containercollection.Container
-			if err := json.Unmarshal([]byte(output), &containers); err != nil {
-				return err
-			}
-
-			if len(containers) == 0 {
-				return fmt.Errorf("expect at least one container in output")
-			}
-
-			for _, gotContainer := range containers {
-				normalize(gotContainer)
-				if !reflect.DeepEqual(gotContainer, expectedContainer) {
-					return fmt.Errorf("expect at least one container to match, expected=%+v, got=%+v", expectedContainer, gotContainer)
-				}
-			}
-
-			return nil
+			return ExpectAllToMatch(output, normalize, expectedContainer)
 		},
 	}
 
--- a/integration/local-gadget/list_containers_test.go
+++ b/integration/local-gadget/list_containers_test.go
@@ -73,43 +73,24 @@ func TestListContainers(t *testing.T) {
 	RunCommands(commands, t)
 }
 
-func TestListSingleContainer(t *testing.T) {
+func TestFilterByContainerName(t *testing.T) {
 	t.Parallel()
-	prefix := "test-list-single"
-	po := fmt.Sprintf("%s-pod", prefix)
-	cn := fmt.Sprintf("%s-container", prefix)
-	ns := GenerateTestNamespaceName(fmt.Sprintf("%s-namespace", prefix))
+	cn := "test-filtered-container"
+	ns := GenerateTestNamespaceName(cn)
 
 	// TODO: Handle it once we support getting K8s container name for docker
 	// Issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/737
 	if *containerRuntime == ContainerRuntimeDocker {
-		t.Skip("Skip TestListSingleContainer on docker since we don't propagate the Kubernetes pod container name")
+		t.Skip("Skip TestFilterByContainerName on docker since we don't propagate the Kubernetes pod container name")
 	}
 
-	TestPodYaml := fmt.Sprintf(`
-apiVersion: v1
-kind: Pod
-metadata:
-  namespace: %s
-  name: %s
-spec:
-  restartPolicy: Never
-  terminationGracePeriodSeconds: 0
-  containers:
-  - name: %s
-    image: busybox
-    command: ["/bin/sh", "-c"]
-    args:
-    - "sleep inf"
-`, ns, po, cn)
-
 	listContainersCmd := &Command{
-		Name: "RunListSingleContainer",
+		Name: "RunFilterByContainerName",
 		Cmd:  fmt.Sprintf("local-gadget list-containers -o json --runtimes=%s --containername=%s", *containerRuntime, cn),
 		ExpectedOutputFn: func(output string) error {
 			expectedContainer := &containercollection.Container{
 				Name:      cn,
-				Podname:   po,
+				Podname:   cn,
 				Runtime:   *containerRuntime,
 				Namespace: ns,
 			}
@@ -129,23 +110,69 @@ spec:
 				c.PodUID = ""
 			}
 
-			return ExpectAllToMatch(output, normalize, expectedContainer)
+			return ExpectAllInArrayToMatch(output, normalize, expectedContainer)
 		},
 	}
 
 	commands := []*Command{
 		CreateTestNamespaceCommand(ns),
-		{
-			Name:           "RunTestListSinglePod",
-			Cmd:            fmt.Sprintf("echo '%s' | kubectl apply -f -", TestPodYaml),
-			ExpectedRegexp: fmt.Sprintf("pod/%s created", po),
-		},
-		{
-			Name:           "WaitForTestListSinglePod",
-			Cmd:            fmt.Sprintf("kubectl wait pod --for condition=ready -n %s %s", ns, po),
-			ExpectedString: fmt.Sprintf("pod/%s condition met\n", po),
-		},
+		PodCommand(cn, "busybox", ns, `["sleep", "inf"]`, ""),
+		WaitUntilPodReadyCommand(ns, cn),
 		listContainersCmd,
+		DeleteTestNamespaceCommand(ns),
+	}
+
+	RunCommands(commands, t)
+}
+
+func TestWatchContainers(t *testing.T) {
+	t.Parallel()
+	cn := "test-watched-container"
+	ns := GenerateTestNamespaceName(cn)
+
+	// TODO: Handle it once we support getting K8s container name for docker
+	// Issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/737
+	if *containerRuntime == ContainerRuntimeDocker {
+		t.Skip("Skip TestWatchContainers on docker since we don't propagate the Kubernetes pod container name")
+	}
+
+	watchContainersCmd := &Command{
+		Name:         "RunWatchContainers",
+		Cmd:          fmt.Sprintf("local-gadget list-containers -o json --runtimes=%s --containername=%s --watch", *containerRuntime, cn),
+		StartAndStop: true,
+		ExpectedOutputFn: func(output string) error {
+			expectedContainer := &containercollection.Container{
+				Name:      cn,
+				Podname:   cn,
+				Runtime:   *containerRuntime,
+				Namespace: ns,
+			}
+
+			normalize := func(c *containercollection.Container) {
+				c.ID = ""
+				c.Pid = 0
+				c.OciConfig = nil
+				c.Bundle = ""
+				c.Mntns = 0
+				c.Netns = 0
+				c.CgroupPath = ""
+				c.CgroupID = 0
+				c.CgroupV1 = ""
+				c.CgroupV2 = ""
+				c.Labels = nil
+				c.PodUID = ""
+			}
+
+			return ExpectEntriesToMatch(output, normalize, expectedContainer)
+		},
+	}
+
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		watchContainersCmd,
+		SleepForSecondsCommand(2), // wait to ensure local-gadget has started
+		PodCommand(cn, "busybox", ns, `["sleep", "inf"]`, ""),
+		WaitUntilPodReadyCommand(ns, cn),
 		DeleteTestNamespaceCommand(ns),
 	}
 
--- a/pkg/container-collection/annotations.go
+++ b/pkg/container-collection/annotations.go
@@ -1,0 +1,39 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package containercollection
+
+const (
+	// container types used by the runtimes
+	containerTypeSandbox   = "sandbox"
+	containerTypeContainer = "container"
+
+	// cri-o container annotations to get container information
+	// https://github.com/containers/podman/blob/main/pkg/annotations/annotations.go
+	// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/types/labels.go
+	crioContainerManagerAnnotation = "io.container.manager"
+	crioPodNameAnnotation          = "io.kubernetes.pod.name"
+	crioPodNamespaceAnnotation     = "io.kubernetes.pod.namespace"
+	crioPodUIDAnnotation           = "io.kubernetes.pod.uid"
+	crioContainerNameAnnotation    = "io.kubernetes.container.name"
+	crioContainerTypeAnnotation    = "io.kubernetes.cri-o.ContainerType"
+
+	// containerd container annotations to get container information
+	// https://github.com/containerd/containerd/blob/main/pkg/cri/annotations/annotations.go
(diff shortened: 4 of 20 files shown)


pkg/container-collection/options.go
```diff
@@ -585,6 +581,30 @@ func WithLinuxNamespaceEnrichment() ContainerCollectionOption {
 	}
 }
 
+// WithOCIConfigEnrichment enriches container using provided OCI config
+func WithOCIConfigEnrichment() ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
+			if container.OciConfig != nil {
```
blanquicet, line 588: What about using early return?

```golang
	if container.OciConfig == nil {
		return true
	}
```

pkg/container-collection/options.go
```diff
@@ -585,6 +581,30 @@ func WithLinuxNamespaceEnrichment() ContainerCollectionOption {
...
+				if cn, ok := container.OciConfig.Annotations[crioContainerNameAnnotation]; ok {
+					container.Name = cn
+				}
+				if pn, ok := container.OciConfig.Annotations[crioPodNameAnnotation]; ok {
+					container.Podname = pn
+				}
+				if ns, ok := container.OciConfig.Annotations[crioPodNamespaceAnnotation]; ok {
+					container.Namespace = ns
+				}
+			}
```
blanquicet, line 601: It should also enrich with the Pod UID.

pkg/container-collection/options.go
blanquicet, line 588: Shouldn't we verify that container is not already enriched as we are doing [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/f6b8c02bd5d1515e022fbf8cec220e7fad09d57f/pkg/container-collection/options.go#L67-L71)? BTW, we should also verify if the Pod UID and the Runtime.

cmd/local-gadget/containers/containers.go
```diff
@@ -44,33 +48,69 @@ func NewListContainersCmd() *cobra.Command {
...
-
 			return nil
 		},
 	}
 
+	cmd.Flags().BoolVarP(
+		&optionWatch,
+		"watch", "w",
+		false,
+		"After listing the containers, watch for new containers")
```
blanquicet, line 86: Please add a new test in `integration/local-gadget/list_containers_test.go` to verify this feature.

blanquicet (commented):
Thanks for working on this! These are just some comments, but it indeed fixed the issue.

pkg/container-collection/annotations.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package containercollection
+
+const (
+	// cri-o container annotations to get container information
+	crioContainerManagerAnnotation = "io.container.manager"
+	crioPodNameAnnotation          = "io.kubernetes.pod.name"
+	crioPodNamespaceAnnotation     = "io.kubernetes.pod.namespace"
+	crioContainerNameAnnotation    = "io.kubernetes.container.name"
```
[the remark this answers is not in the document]
mqasimsarfraz: I checked what containerd was doing and found out they have annotations named "io.kubernetes.cri.container-name", "io.kubernetes.cri.sandbox-namespace" so decided to put the crio prefix for our clarity.

pkg/container-collection/options.go
```diff
@@ -585,6 +581,34 @@ func WithLinuxNamespaceEnrichment() ContainerCollectionOption {
...
 
+// WithOCIConfigEnrichment enriches container using provided OCI config
+func WithOCIConfigEnrichment() ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
+			if container.OciConfig != nil {
+				// skip for pause container
+				if ct, ok := container.OciConfig.Annotations[crioContainerType]; ok && ct == "sandbox" {
+					return false
+				}
```
blanquicet, line 592: I don't think we should skip pause containers, as local-gadget users may also want to debug those. In fact, the ContainerRuntime enricher doesn't drop them.

pkg/container-collection/annotations.go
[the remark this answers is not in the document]
blanquicet: Oh, then the name is ok. Thanks for the clarification.

pkg/container-collection/options.go
mqasimsarfraz: Got it. I realized that pause containers were being printed with `list-containers --watch` I feel like since we aren't listing them with non-watched listing we should skip them during watch as well. What do you think?

blanquicet: Mmmm I remember to have seen the pause containers only for Docker on both cases, watched and non-watched. What runtime were you using when you saw this behaviour?

mauriciovasquezbernal:
> During runc notification container creation is blocked on pidfile so enriching the container using the container runtime won't work because we hit Runtime enricher (cri-o): failed to get container: container "2899fb3ce898a2ed1d4426b0cb6df81ee72cd630ce0fb9e10eb53f04d5eb4e9b" not found. In order to avoid this we are using annotations fetched via OCIConfig to enrich containers.

I don't understand what is the problem here. Why does it work with other runtimes? 

Btw, this commit also fixes the issue with inspektor gadget but only when `--hook-mode=fanotify` is used. I'll investigate why CRI-O hooks aren't working fine.

blanquicet:
> > During runc notification container creation is blocked on pidfile so enriching the container using the container runtime won't work because we hit Runtime enricher (cri-o): failed to get container: container "2899fb3ce898a2ed1d4426b0cb6df81ee72cd630ce0fb9e10eb53f04d5eb4e9b" not found. In order to avoid this we are using annotations fetched via OCIConfig to enrich containers.
> 
> I don't understand what is the problem here. Why does it work with other runtimes?
> 
From what I understood, unlike docker and containerd, when runc-fanotify notifies that a container is being created, CRI-O has not exposed the container through the CRI yet. Therefore, the ContainerRuntime enricher fails to gather the information needed to enrich the internal container structure.

blanquicet:
> > > During runc notification container creation is blocked on pidfile so enriching the container using the container runtime won't work because we hit Runtime enricher (cri-o): failed to get container: container "2899fb3ce898a2ed1d4426b0cb6df81ee72cd630ce0fb9e10eb53f04d5eb4e9b" not found. In order to avoid this we are using annotations fetched via OCIConfig to enrich containers.
> > 
> > 
> > I don't understand what is the problem here. Why does it work with other runtimes?
> 
> From what I understood, unlike docker and containerd, when runc-fanotify notifies that a container is being created, CRI-O has not exposed the container through the CRI yet. Therefore, the ContainerRuntime enricher fails to gather the information needed to enrich the internal container structure.
> 
> Before introducing this patch, I think we should try to understand why it behaves differently from other runtimes.

BTW, I didn't see the notification from runc-fanotify about the pause container when using CRI-O. Is there something wrong with my setup?

mqasimsarfraz:
> BTW, I didn't see the notification from runc-fanotify about the pause container when using CRI-O. Is there something wrong with my setup?

I only see them for `kvm` driver so need to investigate that is the difference between the 2 setups. I am setting it up using: `minikube start -p minikube-kvm --driver=kvm --iso-url=https://storage.googleapis.com/minikube-builds/iso/15232/minikube-v1.27.0-1666976405-15232-amd64.iso --container-runtime=cri-o`.

pkg/container-collection/options.go
mqasimsarfraz: CRI-O on kvm driver. I am bit confused regrading pause container and pod sandbox because the notifications I was seeing are related to pod sandbox. We don't need to show them to users. Right?

mqasimsarfraz:
> > I don't understand what is the problem here. Why does it work with other runtimes?
> 
> From what I understood, unlike docker and containerd, when runc-fanotify notifies that a container is being created, CRI-O has not exposed the container through the CRI yet. Therefore, the ContainerRuntime enricher fails to gather the information needed to enrich the internal container structure.
> 
+1 and just adding to that you can verify this behavior on your setup by starting a verbose trace from `main` e.g `local-gadget trace bind -v -r cri-o`
 and try starting a pod in another terminal. You would see the "not found" error thrown by CRI client (even if you trim the `userdata` in path). Same workflow works for `containerd`. 
> Before introducing this patch, I think we should try to understand why it behaves differently from other runtimes.

From what I understood, since we are blocking the `pidfile` to capture the first event. `conmon` gets blocked till we have enriched and then only the container is available via CRI. But definitely makes to confirm this.

pkg/container-collection/options.go
mqasimsarfraz: I think I have an idea what confused me. We aren't listing the pod sandboxes in the [CRI client](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/cri/cri.go#L80) so they are dropped during enrichment even for `contrainerd`. If we decided to handle them here we should  adopt the cri client to handle them as well.

blanquicet:
> > Before introducing this patch, I think we should try to understand why it behaves differently from other runtimes.
> 
> From what I understood, since we are blocking the `pidfile` to capture the first event. `conmon` gets blocked till we have enriched and then only the container is available via CRI. But definitely makes to confirm this.
>
Yep, but why it doesn't happen with other runtimes?

pkg/container-collection/options.go
```diff
@@ -585,6 +581,34 @@ func WithLinuxNamespaceEnrichment() ContainerCollectionOption {
 	}
 }
 
+// WithOCIConfigEnrichment enriches container using provided OCI config
+func WithOCIConfigEnrichment() ContainerCollectionOption {
```
alban, line 585: Since the name of this function is not specific to cri-o, should you also enrich from containerd annotations?

pkg/container-collection/options.go
blanquicet: Ahh, you are right. We should use [`ListPodSandbox`](https://pkg.go.dev/k8s.io/cri-api@v0.20.6/pkg/apis/runtime/v1alpha2#RuntimeServiceClient.ListPodSandbox) to get the pause containers. While using the Docker API Engine, we get all of them with [the same API](https://pkg.go.dev/github.com/docker/docker@v20.10.8+incompatible/client#Client.ContainerList).

I propose to continue ignoring the pause containers for CRI-O and containerd in this PR, and manage that [issue](https://github.com/inspektor-gadget/inspektor-gadget/issues/1095) in another one.

pkg/container-collection/options.go
mqasimsarfraz: Yes it is available but just for crio via annotation: `io.kubernetes.pod.uid`. I will go ahead and add it.

mqasimsarfraz:
> Yep but, why doesn't happen with other runtimes?

I added some delay to `GetContainer` to compare the state between `containerd` and `cri-o` while blocked and found not that in case of `containerd` container is available via CRI right after `runc create` but in case of `cri-o` it is only available after the `conmon` communicates the container_pid to the `cri-o` server. This seem to be the limitation of the design since `containerd` takes care of creating container by itself, it has more awareness of container states while `cri-o` only knows about container state once it [hears back](https://github.com/containers/conmon/blob/main/src/conmon.c#L396) from `conmon`, side effect of delegating the creation process.

```
	container_pid = atoi(contents);
	ndebugf("container PID: %d", container_pid);

	g_hash_table_insert(pid_to_handler, (pid_t *)&container_pid, container_exit_cb);

	/* Send the container pid back to parent
	 * Only send this pid back if we are using the current exec API. Old consumers expect
	 * conmon to only send one value down this pipe, which will later be the exit code
	 * Thus, if we are legacy and we are exec, skip this write.
	 */
	if ((opt_api_version >= 1 || !opt_exec) && sync_pipe_fd >= 0)
		write_sync_fd(sync_pipe_fd, container_pid, NULL);
```
or with the logs:

```
Nov 07 17:04:01 minikube-cri-o crio[423]: time="2022-11-07 17:04:01.980982415Z" level=info msg="Creating container: default/test-pod-25833/test-pod-25833" id=b90317f1-aeb6-46c0-84a8-3bec0a399770 name=/runtime.v1.RuntimeService/CreateContainer
Nov 07 17:04:01 minikube-cri-o crio[423]: time="2022-11-07 17:04:01.981058169Z" level=warning msg="Allowed annotations are specified for workload []"
Nov 07 17:04:02 minikube-cri-o crio[423]: time="2022-11-07 17:04:02.124023823Z" level=warning msg="Unable to find container ID ef99145814d2ae9e3e0ede9fd58cf9eb7f06985756e0322070c81da24f2c611e" id=a163e4e8-66a8-4b30-8581-b45dbd57f1c6 name=/runtime.v1alpha2.RuntimeService/ListContainers
Nov 07 17:04:02 minikube-cri-o crio[423]: time="2022-11-07 17:04:02.161124862Z" level=info msg="Created container ef99145814d2ae9e3e0ede9fd58cf9eb7f06985756e0322070c81da24f2c611e: default/test-pod-25833/test-pod-25833" id=b90317f1-aeb6-46c0-84a8-3bec0a399770 name=/runtime.v1.RuntimeService/CreateContainer
Nov 07 17:04:02 minikube-cri-o crio[423]: time="2022-11-07 17:04:02.161552331Z" level=info msg="Starting container: ef99145814d2ae9e3e0ede9fd58cf9eb7f06985756e0322070c81da24f2c611e" id=333c6b4f-ae07-4ff4-b5a1-188a5ac85ed7 name=/runtime.v1.RuntimeService/StartContainer
Nov 07 17:04:02 minikube-cri-o crio[423]: time="2022-11-07 17:04:02.197041419Z" level=info msg="Started container" PID=2765 containerID=ef99145814d2ae9e3e0ede9fd58cf9eb7f06985756e0322070c81da24f2c611e description=default/test-pod-25833/test-pod-25833 id=333c6b4f-ae07-4ff4-b5a1-188a5ac85ed7 name=/runtime.v1.RuntimeService/StartContainer sandboxID=d86c2b34c8ca9c838b9505b41c69c18dac51ffa12d4924aa109e2f3643cc6265
```
[Creation](https://github.com/cri-o/cri-o/blob/main/server/container_create.go#L443) only resumes after `enrichers` are done and pidfile is free. With this we know why we need this change for `cri-o` and not for `containerd`.

mqasimsarfraz:
I have updated the PR to handle the comments. Would be great to get another review!

integration/local-gadget/list_containers_test.go
```diff
@@ -89,41 +75,22 @@ func TestListContainers(t *testing.T) {
 
 func TestListSingleContainer(t *testing.T) {
 	t.Parallel()
-	prefix := "test-list-single"
-	po := fmt.Sprintf("%s-pod", prefix)
-	cn := fmt.Sprintf("%s-container", prefix)
-	ns := GenerateTestNamespaceName(fmt.Sprintf("%s-namespace", prefix))
+	cn := "test-list-single"
```
blanquicet, line 78: nit: I would use a diff name for the pod.

integration/local-gadget/list_containers_test.go
```diff
@@ -143,39 +110,68 @@ spec:
...
-			ExpectedRegexp: fmt.Sprintf("pod/%s created", po),
-		},
-		{
-			Name:           "WaitForTestListSinglePod",
-			Cmd:            fmt.Sprintf("kubectl wait pod --for condition=ready -n %s %s", ns, po),
-			ExpectedString: fmt.Sprintf("pod/%s condition met\n", po),
-		},
-		listContainersCmd,
+		watchContainersCmd,
+		PodCommand(cn, "busybox", ns, `["sleep", "inf"]`, ""),
```
blanquicet, line 173: What can we do to ensure `local-gadget` is already running when we start the pod? Maybe a simple `sleep` 😅? I mean, from the output, it is difficult to understand whether `local-gadget` captured it from [the already running containers](https://github.com/inspektor-gadget/inspektor-gadget/blob/5a9c080fc0ac14c63694d91e825b5d6566661a3d/cmd/local-gadget/containers/containers.go#L77-L80), or [the `EventTypeAddContainer` event](https://github.com/inspektor-gadget/inspektor-gadget/blob/5a9c080fc0ac14c63694d91e825b5d6566661a3d/cmd/local-gadget/containers/containers.go#L65-L68). So, we need another way to ensure that, don't you think?

pkg/container-collection/options.go
```diff
@@ -585,6 +582,65 @@ func WithLinuxNamespaceEnrichment() ContainerCollectionOption {
...
+					container.Podname = pn
+				}
+				if ns, ok := container.OciConfig.Annotations[containerdPodNamespaceAnnotation]; ok {
+					container.Namespace = ns
+				}
+				container.Runtime = runtimeclient.ContainerdName
+				return true
+			}
+
+			return true
```
blanquicet, line 638: I thought something like this could look better, but I am not that sure anymore:

```suggestion
			var containerTypeAnnotation, containerNameAnnotation, podNameAnnotation, podNamespaceAnnotation string

			if cm := container.OciConfig.Annotations[crioContainerManagerAnnotation]; cm == runtimeclient.CrioName {
				container.Runtime = runtimeclient.CrioName

				containerTypeAnnotation = crioContainerTypeAnnotation
				containerNameAnnotation = crioContainerNameAnnotation
				podNameAnnotation = crioPodNameAnnotation
				podNamespaceAnnotation = crioPodNamespaceAnnotation

				if pUID, ok := container.OciConfig.Annotations[crioPodUIDAnnotation]; ok {
					container.PodUID = pUID
				}
			} else if _, isContainerd := container.OciConfig.Annotations[containerdContainerTypeAnnotation]; isContainerd {
				container.Runtime = runtimeclient.ContainerdName

				containerTypeAnnotation = containerdContainerTypeAnnotation
				containerNameAnnotation = containerdContainerNameAnnotation
				podNameAnnotation = containerdPodNameAnnotation
				podNamespaceAnnotation = containerdPodNamespaceAnnotation
			} else {
				return true
			}

			// TODO: handle this once we support pod sandboxes via WithContainerRuntimeEnrichment
			// Issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/1095
			if ct := container.OciConfig.Annotations[containerTypeAnnotation]; ct == containerTypeSandbox {
				return false
			}

			if cn, ok := container.OciConfig.Annotations[containerNameAnnotation]; ok {
				container.Name = cn
			}
			if pn, ok := container.OciConfig.Annotations[podNameAnnotation]; ok {
				container.Podname = pn
			}
			if ns, ok := container.OciConfig.Annotations[podNamespaceAnnotation]; ok {
				container.Namespace = ns
			}

			return true
```

blanquicet (commented):
I will test it tomorrow, but I think it is ready to go. Thanks for managing all the comments and investigating the root cause. Awesome work!

integration/local-gadget/list_containers_test.go
mqasimsarfraz: We definitely need to ensure this. I hit it couple of times while running tests. For now we do a good old `sleep` and look into other ways latter (if there are any) ?

cmd/local-gadget/containers/containers.go
```diff
@@ -44,33 +50,73 @@ func NewListContainersCmd() *cobra.Command {
 				return commonutils.WrapInErrParserCreate(err)
 			}
 
-			containers := localGadgetManager.GetContainersBySelector(&containercollection.ContainerSelector{
-				Name: commonFlags.Containername,
-			})
+			w := bufio.NewWriter(os.Stdout)
```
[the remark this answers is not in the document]
blanquicet: OK, but please update the [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/a6cd04153237b493e7b2c79627eef33d38d66420/docs/local-gadget.md?plain=1#L119-L137).

blanquicet (approved):
LGTM after updating the documentation. Thanks!

mqasimsarfraz:
> LGTM after updating the documentation. Thanks!

Thanks a lot with all the reviews. Super helpful. Once `main` is fixed, let's us get a green run and merge this?

blanquicet:
> > LGTM after updating the documentation. Thanks!
> 
> Thanks a lot with all the reviews. Super helpful. Once `main` is fixed, let's us get a green run and merge this?

Yep, let's fix first the CI, so that we can verify this PR is green before merging.

cmd/local-gadget/containers/containers.go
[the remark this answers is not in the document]
mauriciovasquezbernal: > We changed from writing a list of containers to printing one container at a time

Why to change that? As a user I'd expect it to be a list (when not using the -w option). 

Please notice that `kubectl get pods -o json -w` prints a json object that has a list inside, so it's not exactly the same case.

pkg/local-gadget-manager/local-gadget-manager.go
```diff
@@ -348,6 +348,7 @@ func NewManager(runtimes []*containerutils.RuntimeConfig) (*LocalGadgetManager,
 
 	err = l.ContainerCollection.Initialize(
 		containercollection.WithPubSub(containerEventFuncs...),
+		containercollection.WithOCIConfigEnrichment(),
```
mauriciovasquezbernal, line 351: I think it should also be added to `pkg/gadgettracermanager/gadgettracermanager.go`, otherwise this fix won't be available to Inspektor Gadget in Kubernetes.

integration/command.go
```diff
@@ -598,3 +598,11 @@ func WaitUntilPodReadyCommand(namespace string, podname string) *Command {
 func WaitUntilTestPodReadyCommand(namespace string) *Command {
 	return WaitUntilPodReadyCommand(namespace, "test-pod")
 }
+
+// WaitForSecondsCommand returns a Command which waits for given seconds
+func WaitForSecondsCommand(seconds int) *Command {
```
mauriciovasquezbernal, line 603: Use `SleepForSecondsCommand` to be more explicit?

mauriciovasquezbernal (commented):
Some comments about the json output format.

cmd/local-gadget/containers/containers.go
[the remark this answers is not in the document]
blanquicet: > As a user I'd expect it to be a list (when not using the -w option).
>
I'm not sure. Considering how other gadgets works, I wouldn't. In addition, I will definitely not expect the output format changes when using this kind of option (`--watch`). IMO, it's annoying for people to parse the output, don't you think?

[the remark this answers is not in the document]
mauriciovasquezbernal: > it's annoying for people parsing the output to have to manage different formats of the same command, don't you think?

I agree, but I also think it's annoying that the output of a `list-containers` command is not something I can parse as a json object (with this PR):

```bash
$ sudo ./local-gadget list-containers -o json | jsonlint-php 
Parse error on line 14:
... "local-registry"}{  "runtime": "dock
--------------------^
Expected one of: 'EOF', '}', ',', ']' (stdin)
```

For gadgets that print multiple times to the terminal (tracers, top, audit, ...) what we do is that **each** event is printed as a json object to the terminal in a new line, then the client can consume each line as a valid json. 

To summarize:
1. Gadgets that print multiple objects a once and then return (snapshot) print a valid json array.
2. Gadgets that print multiple events print an event per line using json. 

The point is that `list-containers` falls on the first category but `list-containers -w` falls on the second one. 

Are my concerns clear?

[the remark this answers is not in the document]
mqasimsarfraz: Okay so I see the point here. Just to summarize my understanding: for most of the gadgets we are **waiting** for the json events so it is okay to stream a single json object at a time but for "print and exit" commands we should output a valid json?  Having said that, it then makes sense to revert to printing json list for `list-containers` and printing stream of json objects for `list-containers -w` ?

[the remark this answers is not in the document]
blanquicet: > Are my concerns clear?
>
Yep, they are.

> The point is that `list-containers` falls on the first category but `list-containers -w` falls on the second one. 
>
However, I still find this pretty strange. I can't recall any other CLI doing something similar. 

Anyway, let's move this PR on. I am OK printing a JSON array for `list-containers` and a stream of JSON objects for `list-containers -w`.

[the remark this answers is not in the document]
mqasimsarfraz: > > The point is that `list-containers` falls on the first category but `list-containers -w` falls on the second one.
> 
> However, I still find this pretty strange. I can't recall any other CLI doing something similar.

I also thought the same but it seem `kubectl` falls in this category with different outputs.
```
→ kubectl get pods -o json 
{
    "apiVersion": "v1",
    "items": [
       ...
    ],
    "kind": "List",
    "metadata": {
        "resourceVersion": ""
    }
}

```
and 

```
→ kubectl get pods -o json -w
{
    "apiVersion": "v1",
    "kind": "Pod",
    ...
}
```
Anyhow, let me update the PR.

pkg/local-gadget-manager/local-gadget-manager.go
mqasimsarfraz: True. But I prefer not to hanldle it in the same PR. I wanted to have look into `crio` hook mode and do some testing for IG with this.  I opened the [issue](https://github.com/inspektor-gadget/inspektor-gadget/issues/1142) to track this. What do you think?

pkg/container-collection/annotations.go
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+package containercollection
+
+const (
+	// container types used by the runtimes
+	containerTypeSandbox   = "sandbox"
+	containerTypeContainer = "container"
+
+	// cri-o container annotations to get container information
```
alban, line 22: Could you add a comment saying where it comes from? I can see some of them at:

https://github.com/containers/podman/blob/main/pkg/annotations/annotations.go
https://github.com/containerd/containerd/blob/master/pkg/cri/annotations/annotations.go

But not all of them.

pkg/container-collection/options.go
```diff
@@ -585,6 +582,54 @@ func WithLinuxNamespaceEnrichment() ContainerCollectionOption {
...
+// WithOCIConfigEnrichment enriches container using provided OCI config
+func WithOCIConfigEnrichment() ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
+			if container.OciConfig == nil || container.IsEnriched() {
+				return true
+			}
+
+			var containerTypeAnnotation, containerNameAnnotation, podNameAnnotation, podNamespaceAnnotation string
+			if cm := container.OciConfig.Annotations[crioContainerManagerAnnotation]; cm == runtimeclient.CrioName {
```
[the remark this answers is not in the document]
mqasimsarfraz: > I wonder if it could even be a separate package, so it could be reused by [seccompagent's kuberesolver.go#L30-L41](https://github.com/kinvolk/seccompagent/blob/main/pkg/kuberesolver/kuberesolver.go#L30-L41) without importing all container-collection but that would complicate `enrichFromAnnotations()`.

True, that would be nice. How about we return a custom `struct` with all the currently supported values in a standalone package? Later on it can be mapped to `containercollection.Container` in `enrichFromAnnotations` ?

mauriciovasquezbernal (approved):
LGTM.

pkg/container-collection/options.go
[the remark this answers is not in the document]
alban: That sounds good. Could this be done in a different PR? So that discussions on the details don't postpone this improvement on crio support.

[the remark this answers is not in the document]
mqasimsarfraz: I opened the [PR](https://github.com/inspektor-gadget/inspektor-gadget/pull/1147) with the changes. Should we go ahead and merge this one?

mqasimsarfraz:
Thanks for all the reviews! I will go ahead and merge this!

PR #1085, opened 2022-10-26
pkg/columns: fix handling of fields with alternative type definitions

flyth:
This PR fixes an issue with alternative string types in `pkg/columns/formatter/textcolumns` (seen in #1071)  and adds tests for it. The issue could only happen for strings as they have been cast directly to string instead of using the `.String()` function of the reflect package.

The issue also exists in `pkg/columns/filter` but should be fixed once #1081 is merged.


alban (approved):
Tested in my branch.
LGTM, thanks!

PR #1086, opened 2022-10-27
ci: Fix cleanup process for integration-tests

mqasimsarfraz:
We moved the inspektor-gadget integration tests to a new directory resulting in the test process named differently i.e inspektor-gadget.test rather than integration.test. This PR uses the correct process name in the cleanup. 

### Testing Done
A [workflow run](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3338706275/jobs/5527308288) with cancellation.

### DIFF
--- a/.github/actions/run-integration-tests/action.yml
+++ b/.github/actions/run-integration-tests/action.yml
@@ -40,7 +40,7 @@ runs:
             # Forward the SIGINT directly to test process but wait for current
             # active jobs since we can only wait for current shell child process.
             echo "IntegrationTestsJob: Notifying the integration tests process about the cancellation"
-            kill -2 $(pidof integration.test) > /dev/null
+            kill -2 $(pidof inspektor-gadget.test) > /dev/null
             echo "IntegrationTestsJob: Waiting for the integration tests process to finish"
             wait $(jobs -p)
             echo "IntegrationTestsJob: We are done with the clean-up. Let the job exit"


.github/actions/run-integration-tests/action.yml
```diff
@@ -40,7 +40,7 @@ runs:
             # Forward the SIGINT directly to test process but wait for current
             # active jobs since we can only wait for current shell child process.
             echo "IntegrationTestsJob: Notifying the integration tests process about the cancellation"
-            kill -2 $(pidof integration.test) > /dev/null
+            kill -2 $(pidof inspektor-gadget.test) > /dev/null
```
blanquicet, line 43: Is it possible to set the name? If so, I would suggest to do that to avoid it changing every time we change the folder name.

mqasimsarfraz: I didn't find a way to directly do it with `go test` without splitting it into compile and run phase. I have updated the PR let me know if it is looking good now.

blanquicet: Oh, I thought there was any way to do that more quickly. Mmmm, I'm not sure about this two-step approach. But if we decide to go with this approach, we should remove the binary after running it.

mqasimsarfraz: Yes, definitely we should remove the binary but I agree it looks bit too much. Other thought I had was to signal using `pgrep -P $(pidof go)` which removes the issue as well. What do you say?

mauriciovasquezbernal: I think just renaming the process should be enough. I don't expect that folder to be renamed too often.

mqasimsarfraz: Cool. I will revert the changes.

blanquicet (approved):
LGTM!

mauriciovasquezbernal (approved):

mqasimsarfraz:
Thanks for the reviews!

PR #1091, opened 2022-11-02
tests/tcp: Only test against a cluser local IP

burak-ok:
Many TCP Tests sometimes fail.
These flaky tests are not productive for the team, since we should rely on their results.

Outside services are not always up or can change their behavior. Furthermore we should not check for both ports `80` and `443` when we only specificed one.

This PR creates a new `nginx` pod and tries to connect to its port `80`

See:
https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3378520742/jobs/5610001881#step:7:499
https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3378520742/jobs/5610002721#step:8:603

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -571,12 +571,18 @@ func DeleteRemainingNamespacesCommand() *Command {
 	}
 }
 
+// WaitUntilPodReadyCommand returns a Command which waits until pod with the specified name in
+// the given as parameter namespace is ready.
+func WaitUntilPodReadyCommand(namespace string, podname string) *Command {
+	return &Command{
+		Name:           "WaitForTestPod",
+		Cmd:            fmt.Sprintf("kubectl wait pod --for condition=ready -n %s %s", namespace, podname),
+		ExpectedString: fmt.Sprintf("pod/%s condition met\n", podname),
+	}
+}
+
 // WaitUntilTestPodReadyCommand returns a Command which waits until test-pod in
 // the given as parameter namespace is ready.
 func WaitUntilTestPodReadyCommand(namespace string) *Command {
-	return &Command{
-		Name:           "WaitForTestPod",
-		Cmd:            fmt.Sprintf("kubectl wait pod --for condition=ready -n %s test-pod", namespace),
-		ExpectedString: "pod/test-pod condition met\n",
-	}
+	return WaitUntilPodReadyCommand(namespace, "test-pod")
 }
--- a/integration/command.go
+++ b/integration/command.go
@@ -481,6 +481,42 @@ func (c *Command) Stop(t *testing.T) {
 	c.Started = false
 }
 
+// PodCommand returns a Command that starts a pid with a specified image, command and args
+func PodCommand(podname, image, namespace, command, commandArgs string) *Command {
+	cmdLine := ""
+	if command != "" {
+		cmdLine = fmt.Sprintf("\n    command: %s", command)
+	}
+
+	commandArgsLine := ""
+	if commandArgs != "" {
+		commandArgsLine = fmt.Sprintf("\n    args:\n    - %s", commandArgs)
+	}
+
+	cmdStr := fmt.Sprintf(`kubectl apply -f - <<"EOF"
+apiVersion: v1
+kind: Pod
+metadata:
+  name: %s
+  namespace: %s
+  labels:
+    run: %s
+spec:
+  restartPolicy: Never
+  terminationGracePeriodSeconds: 0
+  containers:
+  - name: %s
+    image: %s%s%s
+EOF
+`, podname, namespace, podname, podname, image, cmdLine, commandArgsLine)
+
+	return &Command{
+		Name:           fmt.Sprintf("Run%s", podname),
+		Cmd:            cmdStr,
+		ExpectedString: fmt.Sprintf("pod/%s created\n", podname),
+	}
+}
+
 // BusyboxPodRepeatCommand returns a Command that creates a pod and runs
 // "cmd" each 0.1 seconds inside the pod.
 func BusyboxPodRepeatCommand(namespace, cmd string) *Command {
@@ -490,31 +526,7 @@ func BusyboxPodRepeatCommand(namespace, cmd string) *Command {
 
 // BusyboxPodCommand returns a Command that creates a pod and runs "cmd" in it.
 func BusyboxPodCommand(namespace, cmd string) *Command {
-	cmdStr := fmt.Sprintf(`kubectl apply -f - <<"EOF"
-apiVersion: v1
-kind: Pod
-metadata:
-  name: test-pod
-  namespace: %s
-  labels:
-    run: test-pod
-spec:
-  restartPolicy: Never
-  terminationGracePeriodSeconds: 0
-  containers:
-  - name: test-pod
-    image: busybox
-    command: ["/bin/sh", "-c"]
-    args:
-    - %s
-EOF
-`, namespace, cmd)
-
-	return &Command{
-		Name:           "RunTestPod",
-		Cmd:            cmdStr,
-		ExpectedString: "pod/test-pod created\n",
-	}
+	return PodCommand("test-pod", "busybox", namespace, `["/bin/sh", "-c"]`, cmd)
 }
 
 // GenerateTestNamespaceName returns a string which can be used as unique
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"os/exec"
 	"reflect"
 	"strings"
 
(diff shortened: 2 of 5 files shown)


integration/inspektor-gadget/integration_test.go
```diff
@@ -1300,14 +1291,14 @@ func TestTcpconnect(t *testing.T) {
...
-			return ExpectEntriesToMatch(output, normalize, expectedEntries...)
+			return ExpectEntriesToMatch(output, normalize, expectedEntry)
 		},
 	}
 
 	commands := []*Command{
 		CreateTestNamespaceCommand(ns),
 		tcpconnectCmd,
-		BusyboxPodRepeatCommand(ns, "wget -q -O /dev/null -T 3 http://1.1.1.1"),
+		BusyboxPodRepeatCommand(ns, "wget -q -O /dev/null -T 3 http://1.1.1.1:80"),
```
mqasimsarfraz, line 1301: We don't need to specify port here or you are trying to make it explicit?

mqasimsarfraz:
I wonder if cloudflare changed something and suddenly we are seeing so many errors. :worried:

integration/inspektor-gadget/integration_test.go
burak-ok: Just making things even more explicit. `http://` should be enough for port 80

burak-ok:
I actually don't know what happened.
But in the CI run of this PR the following error happend: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3379201568/jobs/5610524602#step:7:570

The test for `TestTraceTCP` can't find the destination addr of `1.1.1.1` in the traces. Indeed it is not in there.
What is going on?

mqasimsarfraz:
> > The test for TestTraceTCP can't find the destination addr of 1.1.1.1 in the traces. Indeed it is not in there.
> > What is going on?
> 
> What about deploying a `nginx` pod, monitoring it with `trace tcp` and doing a `wget` (from the pod) toward the `nginx` IP? This should generate TCP traffic and we would not depend on external resources, so I would say this test should be more reliable.

+1 or how about doing these tests against api server i.e `https://kubernetes.default.svc.cluster.local` (`KUBERNETES_SERVICE_HOST/KUBERNETES_SERVICE_PORT`) or even `kube-dns`.

integration/inspektor-gadget/integration_test.go
```diff
@@ -1325,49 +1314,31 @@ func TestTcptracer(t *testing.T) {
...
-					Operation: "connect",
-				},
-				{
-					Event:     BuildBaseEvent(ns),
-					Comm:      "wget",
-					IPVersion: 4,
-					Daddr:     "1.1.1.1",
-					Dport:     80,
-					Operation: "close",
-				},
```
mauriciovasquezbernal, line 1344: Any idea why do we have to remove this?

integration/inspektor-gadget/integration_test.go
```diff
@@ -1325,49 +1314,31 @@ func TestTcptracer(t *testing.T) {
...
+				IPVersion: 4,
+				Dport:     443,
+				Operation: "connect",
 			}
 
 			normalize := func(e *tcpTypes.Event) {
 				e.Node = ""
 				e.Pid = 0
 				e.Saddr = ""
+				e.Daddr = ""
```
mauriciovasquezbernal, line 1329: I suppose we have to make this because we're accessing the API server by using a URL. I feel the idea of testing with a server (nginx) or similar will allow us to perform a deeper test without relying on DNS.

mqasimsarfraz: True. Now that I am thinking how to propagate this address/port, deploying own pod sounds simpler.

mauriciovasquezbernal: I implemented a function to get the IP of a pod in https://github.com/inspektor-gadget/inspektor-gadget/commit/d012b5cfabb99370a0ff0b737c9361815dea643d, this could help in this case too. (probably has to be changed a bit to allow different pod names but that's the idea).

burak-ok:
New version bases on @mauriciovasquezbernal branch with his Helperfunction to get a Pods IP.

We can now probably vaildate even more in these tests since we capture the nginx pod too.
I would say that we should do that in another PR

integration/helpers.go
```diff
@@ -152,3 +153,14 @@ func BuildBaseEvent(namespace string) eventtypes.Event {
...
+func GetTestPodIP(ns string, podname string) string {
+	cmd := exec.Command("kubectl", "-n", ns, "get", "pod", podname, "-o", "jsonpath='{.status.podIP}'")
+	r, err := cmd.Output()
+	if err != nil {
+		return ""
+	}
+
+	ip := string(r)
+	return ip[1 : len(ip)-1]
+}
```
mauriciovasquezbernal, line 166: Feel free to squash this with my previous commit. I only wanted to show that as an example, so there is not need for my commit to be there.

integration/helpers.go
```diff
@@ -152,3 +153,14 @@ func BuildBaseEvent(namespace string) eventtypes.Event {
 		CommonData: BuildCommonData(namespace),
 	}
 }
+
+func GetTestPodIP(ns string, podname string) string {
+	cmd := exec.Command("kubectl", "-n", ns, "get", "pod", podname, "-o", "jsonpath='{.status.podIP}'")
+	r, err := cmd.Output()
+	if err != nil {
+		return ""
```
mauriciovasquezbernal, line 161: nit: `return fmt.Sprintf("failed to get Pod IP: %s", err)`. 

Ideally it should return an error but it becomes to annoying to handle it over all the places, so having that will allow us to understand when it fails.

integration/local-gadget/trace_tcpconnect_test.go
```diff
@@ -31,21 +31,16 @@ func TestTraceTcpconnect(t *testing.T) {
...
-				},
-				{
-					Event:     BuildBaseEvent(ns),
-					Comm:      "wget",
-					IPVersion: 4,
-					Daddr:     "1.1.1.1",
-					Dport:     443,
-				},
+			saddr := GetTestPodIP(ns, "test-pod")
+			NginxIP := GetTestPodIP(ns, "nginx-pod")
```
mauriciovasquezbernal, line 35: nit: if you move the `commandsPreTest` before this, then you can reuse the NginxIP from L68 without having to call `GetTestPodIP(ns, "nginx-pod")` again.

integration/command.go
```diff
@@ -481,42 +481,54 @@ func (c *Command) Stop(t *testing.T) {
...
+func PodCommand(podname, image, namespace, command, commandArgs string) *Command {
+	cmdLine := ""
+	if command != "" {
+		cmdLine = fmt.Sprintf("\n    command: %s", command)
+	}
+
+	commandArgsLine := ""
+	if commandArgs != "" {
+		commandArgsLine = fmt.Sprintf("\n    args:\n    - %s", commandArgs)
+	}
```
mauriciovasquezbernal, line 494: Not related to this PR: I wonder if we should consider using the k8s-apiserver directly instead of invoking kubectl. It seems the things we have are getting complicated and using the K8s API could simplify that.

mauriciovasquezbernal (commented):
Looks very good. Some minor comments around.

mauriciovasquezbernal (approved):
LGTM. Thanks for fixing this!

gadget-container/cleanup.sh
```diff
@@ -22,8 +22,12 @@ for i in ocihookgadget prestart.sh poststop.sh ; do
...
 
 # CRIO hooks
-rm -f /host/etc/containers/oci/hooks.d/gadget-prestart.json
-rm -f /host/etc/containers/oci/hooks.d/gadget-poststop.json
+for HOOK_PATH in "/host/etc/containers/oci/hooks.d" \
+                 "/host/usr/share/containers/oci/hooks.d/"
+do
+  rm -f $HOOK_PATH/gadget-prestart.json
+  rm -f $HOOK_PATH/gadget-poststop.json
+done
```
mauriciovasquezbernal, line 30: Why is this showing these changes here? Probably you have to rebase again?

burak-ok: Yep, another rebase fixed that

burak-ok:
The following test still sometimes fails: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3395755607/jobs/5646037941

That test was not part of this PR and the behavior seems to be a little bit different. On my TODO list

blanquicet:
> The following test still sometimes fails: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3395755607/jobs/5646037941
> 
Hey @alban, I just wanted to let you know about this failure on the test 🙂.

alban:
It seems the event about 1.1.1.1:443 is received twice. Any idea about the cause? Do you think it is because the first connection failed and `wget` tried again?

burak-ok:
It is probablty twice. Once on Port `80` and another one on Port `443`.
Why this is I currently do not know

alban:
It seems to be 443 twice because 
* it gets the first 443 as expected at pkg/local-gadget-manager/local-gadget-manager_test.go line 639
* it receives the second 443 and fails at local-gadget-manager_test.go:678

mauriciovasquezbernal:
I'm hitting this as well in https://github.com/inspektor-gadget/inspektor-gadget/pull/1099

integration/command.go
blanquicet: Tracking this in https://github.com/inspektor-gadget/inspektor-gadget/issues/1112

PR #1092, opened 2022-11-02
Remove cgo where possible from gadgets

### LINKED ISSUE #872: pkg/gadgets: Use Go types generated by bpf2go instead of C structs
bpf2go generates Go types for map keys and values -> https://github.com/cilium/ebpf/tree/master/cmd/bpf2go#generated-types. (example [here](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/pkg/gadgets/trace/exec/tracer/execsnoop_bpfel_x86.go#L16-L28) & [here](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/pkg/gadgets/trace/capabilities/tracer/capabilities_bpfel_x86.go#L16-L25))

In the gadgets we're [using](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/pkg/gadgets/trace/exec/tracer/tracer.go#L21) the C representation of that struct, which forces us to use cgo https://github.com/kinvolk/inspektor-gadget/blob/v0.7.1/pkg/gadgets/trace/exec/tracer/tracer.go#L161. 

The goal of this issue is to use the Go type generated instead of the C struct to avoid using cgo there. 

Things to take into consideration:
- Check why we're using `-no-global-types` in some tracers
- Look for a way to convert `[N]int8` to `string` without using cgo

flyth:
As we want to enable third parties to use our gadgets directly in their own environments / tools, we should make it as easy as possible. Sometimes, a significant hurdle in building tools written in Go is the requirement of CGO - hence we should try to remove as many dependencies on that as possible. Also, switching from Go to C code at runtime always comes with increased costs in terms of CPU usage, so that is another reason to avoid mixed code.

This is the first PR laying the groundwork for all our gadgets. The updated gadgets themselves will follow in separate PRs by gadget category.

* #1093

Fixes #872

### DIFF
--- a/pkg/container-utils/cgroups/cgroups.go
+++ b/pkg/container-utils/cgroups/cgroups.go
@@ -21,56 +21,9 @@ import (
 	"path/filepath"
 	"strings"
 	"unsafe"
+
+	"golang.org/x/sys/unix"
 )
-
-/*
-#define _GNU_SOURCE
-#include <stdlib.h>
-#include <stdio.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
-#include <stdint.h>
-
-struct cgid_file_handle
-{
-  //struct file_handle handle;
-  unsigned int handle_bytes;
-  int handle_type;
-  uint64_t cgid;
-};
-
-uint64_t get_cgroupid(char *path) {
-  struct cgid_file_handle *h;
-  int mount_id;
-  int err;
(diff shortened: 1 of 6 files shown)


mauriciovasquezbernal (commented):
Any reason for it to be draft? IMO it's ready to go!

mauriciovasquezbernal (approved):

PR #483, opened 2022-01-28
entrypoint: Install OCI hooks also in /etc/containers/oci/hooks.d

blanquicet:
# Install OCI hooks also in /etc/containers/oci/hooks.d

Testing IG on [OpenShift 4.9](https://developers.redhat.com/courses/explore-openshift/openshift-49-playground) and [Azure Red Hat Openshift](https://docs.microsoft.com/en-us/azure/openshift/) (OpenShift 4.8), the gadget pod is not able to install the hooks configurations because the folder `/host/usr/share/containers/oci/hooks.d/` is read-only:

```bash
$ kubectl logs -n gadget <gadget-name>
OS detected: "Red Hat Enterprise Linux CoreOS 48.84.202110270303-0 (Ootpa)"
Kernel detected: 4.18.0-305.19.1.el8_4.x86_64
bcc detected: 0.24.0-1
Gadget image: joseregistry.azurecr.io/gadget:latest
Deployment options:
INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER=true
INSPEKTOR_GADGET_OPTION_TOOLS_MODE=core
INSPEKTOR_GADGET_OPTION_HOOK_MODE=auto
Inspektor Gadget version: v0.4.2-115-g40b27d56ec9d-dirty
CRI-O detected.
hook mode cri-o detected.
Installing hooks scripts on host...
Installing ocihookgadget...
Installing prestart.sh...
Installing poststop.sh...
Installing OCI hooks configuration in /usr/share/containers/oci/hooks.d
cp: cannot create regular file '/host/usr/share/containers/oci/hooks.d/gadget-prestart.json': Read-only file system
```

## How to use

Try IG on  OpenShift 4.9 or ARO.

## Testing done

After applying this PR, the hooks are installed correctly:

```bash
$ kubectl logs -n gadget <gadget-name>
OS detected: "Red Hat Enterprise Linux CoreOS 48.84.202110270303-0 (Ootpa)"
Kernel detected: 4.18.0-305.19.1.el8_4.x86_64
bcc detected: 0.24.0-1
Gadget image: joseregistry.azurecr.io/gadget:jose-oci-hooks
Deployment options:
INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER=true
INSPEKTOR_GADGET_OPTION_TOOLS_MODE=core
INSPEKTOR_GADGET_OPTION_HOOK_MODE=auto
[...]

PR #1094, opened 2022-11-02
hooks: Install cri-o hooks in multiple folders

mauriciovasquezbernal:
We don't have a deterministic way to understand where these hooks should be installed, then try to install in the well-known paths.

Installing multiple times the same hook is not an issue as they are only executed once https://github.com/containers/podman/blob/v3.0.1/pkg/hooks/docs/oci-hooks.5.md#directories

## Ref 

It seems we got this wrong on https://github.com/inspektor-gadget/inspektor-gadget/pull/483. 

### Before this change

```
# Create a cluster with minikube
$ minikube start --container-runtime=cri-o --driver=kvm2

# start a tracer
$ kubectl gadget trace exec 

# run a pod (in another terminal)
$ kubectl run mypod$RANDOM --image=praqma/network-multitool -- sleep 9999

# the tracer terminal won't show any event 
```

### After this change 

```
# Create a cluster with minikube
$ minikube start --container-runtime=cri-o --driver=kvm2

# start a tracer
$ kubectl gadget trace exec 

# run a pod (in another terminal)
$ kubectl run mypod$RANDOM --image=praqma/network-multitool -- sleep 9999

# the tracer terminal will print the first events of the container
```

### DIFF
--- a/gadget-container/entrypoint.sh
+++ b/gadget-container/entrypoint.sh
@@ -80,16 +80,20 @@ if [ "$HOOK_MODE" = "crio" ] ; then
     cp /opt/hooks/oci/$i /host/opt/hooks/oci/
   done
 
-  echo "Installing OCI hooks configuration in /etc/containers/oci/hooks.d"
-  mkdir -p /host/etc/containers/oci/hooks.d
-  cp /opt/hooks/crio/gadget-prestart.json /host/etc/containers/oci/hooks.d/ 2>/dev/null || true
-  cp /opt/hooks/crio/gadget-poststop.json /host/etc/containers/oci/hooks.d/ 2>/dev/null || true
+  for HOOK_PATH in "/host/etc/containers/oci/hooks.d" \
+                   "/host/usr/share/containers/oci/hooks.d/"
+  do
+    echo "Installing OCI hooks configuration in /etc/containers/oci/hooks.d"
+    mkdir -p $HOOK_PATH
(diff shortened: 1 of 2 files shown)


gadget-container/entrypoint.sh
```diff
@@ -80,16 +80,20 @@ if [ "$HOOK_MODE" = "crio" ] ; then
...
-  cp /opt/hooks/crio/gadget-poststop.json /host/etc/containers/oci/hooks.d/ 2>/dev/null || true
-
-  if ! ls /host/etc/containers/oci/hooks.d/gadget-{prestart,poststop}.json > /dev/null 2>&1; then
-    echo "Couldn't install OCI hooks configuration" >&2
-  else
-    echo "Hooks installation done"
-  fi
+  for HOOK_PATH in "/host/etc/containers/oci/hooks.d" \
+                   "/host/usr/share/containers/oci/hooks.d/"
+  do
```
blanquicet, line 85: So, it means that minikube with CRI-O uses a different path than OpenShift (because it is working). Do you think adding a test in the integration test for this configuration is worth adding?

blanquicet (approved):
LGTM from code inspection.

gadget-container/entrypoint.sh
mauriciovasquezbernal: Yes, those use different folders and that's why it wasn't working on minikube. I think adding an integration test for it would be great, but I don't know how difficult it's. Created https://github.com/inspektor-gadget/inspektor-gadget/issues/1098 to track this down.

PR #1096, opened 2022-11-03
use gRPC via port-forwarding instead of exec to retrieve events from traces

flyth:
This PR changes how event streams are retrieved from the gadgettracermanager.

Until now, whenever a stream is requested (e.g. trace/top gadgets), kubectl gadget would connect to the API server and exec a new process (another instance of gadgettracermanager that is only used as a client) on each node that connects to a unix socket where main gadgettracermanager is listening on using gRPC. It would then print out line by line to stdout, which then is received by the kubectl gadget.

```mermaid
 flowchart TD
	 subgraph Old
		 direction TB
	     user[User]
	     kubectl[kubectl gadget]
	     apisrv{k8s API Server}
	     user <-- kubectl gadget trace tcp --> kubectl
	     kubectl <-.-> apisrv
	     apisrv -- "exec gadgettracermgr(traceID=123)" --> gadgettracermgr-client
	     gadgettracermgr-client -- "JSON encoded lines via stdout" --> apisrv
		 subgraph Node
		     direction TB
		     gadgettracermgr-client <-- gRPC --> gadgettracermgr-server
		 end
	end
 ```

After this PR, the API server will still be used, but instead of execing, a port forwarding is established and used for a gRPC connection directly to the gadgettracermanager.

```mermaid
 flowchart TD
	subgraph New
		direction TB
	    user[User]
	    kubectl[kubectl gadget]
	    apisrv{k8s API Server}
	    user <-- kubectl gadget trace tcp --> kubectl
	    kubectl <-. "gRPC getTrace(traceID=123) in Tunnel" .-> apisrv
		apisrv <-. JSON encoded lines via gRPC .-> gadgettracermgr-server
	end
```

Please note that right now we're still using JSON encoding inside the gRPC messages - so gRPC essentially just replaces the transport.

#### Binary compatible interface

While we're not using binary encoded messages right now, we should consider moving at least gadgets with higher throughput to a more lightweight encodings like protocol buffers. Also, if we want to transport whole packets (or parts) to the kubectl gadget side, JSON becomes really inefficient as it uses base64 encoding in that case.

#### Cleaner interfaces

Using gRPC, we don't have to decode messages ourselves like we do right now when splitting the stdin by lines. Defining alternative messages (proper error handling on the transport layer) becomes much easier using dedicated messages.

Also, this is the first step towards a fanning-out mechanism, where we can use the nodes itself to forward trace requests t other nodes (more on that in future PRs).

### Authentication

On deployment, a CA certificate will be generated and saved into a secret. This secret will be read by the gadgettracermanager on start and a new node (server) certificate will be generated. When using kubectl gadget, the client does the same for a client certificate to establish a secure gRPC connection with the server. Fancier mechanisms using cert-manager etc. might be added later on.

### DIFF
--- a/cmd/kubectl-gadget/utils/exec.go
+++ b/cmd/kubectl-gadget/utils/exec.go
@@ -17,17 +17,27 @@ package utils
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
 
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/tools/remotecommand"
+	"k8s.io/client-go/transport/spdy"
 
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	pb "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgettracermanager/api"
 )
 
 func ExecPodSimple(client *kubernetes.Clientset, node string, podCmd string) string {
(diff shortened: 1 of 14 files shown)


cmd/kubectl-gadget/utils/exec.go
```diff
@@ -100,3 +109,108 @@ func ExecPod(client *kubernetes.Clientset, node string, podCmd string, cmdStdout
...
+
+	fw, err := portforward.New(dialer, []string{fmt.Sprintf("%d:%d", 0, 7080)}, stopChan, readyChan, nil, os.Stderr)
+	if err != nil {
+		return fmt.Errorf("creating port forwarder: %w", err)
+	}
+
+	go fw.ForwardPorts()
+	defer func() {
+		stopChan <- struct{}{}
+	}()
```
mauriciovasquezbernal, line 172: `defer close(stopCh)`?

cmd/kubectl-gadget/utils/exec.go
```diff
@@ -100,3 +109,108 @@ func ExecPod(client *kubernetes.Clientset, node string, podCmd string, cmdStdout
...
+
+	if len(ports) != 1 {
+		return errors.New("unexpected result from GetPorts()")
+	}
+
+	conn, err := grpc.Dial(fmt.Sprintf("127.0.0.1:%d", ports[0].Local), grpc.WithInsecure())
+	if err != nil {
+		return fmt.Errorf("connecting to grpc: %w", err)
+	}
+
```
mauriciovasquezbernal, line 190: `defer conn.Close()` ?

mauriciovasquezbernal (commented):
Some minor comments. IMHO it's going to the right direction. btw, I had implemented a PoC of this before in https://github.com/inspektor-gadget/inspektor-gadget/commit/0d5c2db3999dc151aba6efc1cc43ddbc8650101b#diff-4842476e654447ce562259a42a83a1de242283a763ad097e32168b199c7e309e, nice to see that your implementation and mine are very similar :D.

flyth:
Yeah, looks similar :) I'm not quite happy about the fact that we open a socket for each forward on the local machine, though. However, it should in theory be fairly easy to get the forward stream directly and use it for the grpc calls instead, I'll look more into it.

alban:
Given that the gadget pods use `hostNetwork: true`, this will listen to port 7080 on the host (if no other apps use the same port) and we cannot use Kubernetes Network Policies to restrict it. Even if is restricted to the loopback interface, this seems too much to me. I think we cannot afford to use `grpc.WithInsecure()` anymore. Would it be possible to use some credentials linked to the gadget service account?

I think the grpc server needs to continue to listen on the unix socket for compatibility with Headlamp plugins (unless you update the Headlamp plugin at the same time as this PR?).

flyth:
> Given that the gadget pods use `hostNetwork: true`, this will listen to port 7080 on the host (if no other apps use the same port) and we cannot use Kubernetes Network Policies to restrict it. Even if is restricted to the loopback interface, this seems too much to me. I think we cannot afford to use `grpc.WithInsecure()` anymore. Would it be possible to use some credentials linked to the gadget service account?

For the fan-out situation (forwarding of traces) we also need the security aspect, so let's add it directly in here. As a first solution for this, I  implemented a CA that's being added as a secret when deploying and each node will generate its own certificate using that CA. From the kubectl side, whenever connecting for the first time (or after the certificate has expired), the client will do the same automatically.

> I think the grpc server needs to continue to listen on the unix socket for compatibility with Headlamp plugins (unless you update the Headlamp plugin at the same time as this PR?).

Yeah, let's just add it separately, then. There are more RPCs in future PRs that will then only be added to the (authenticated) tcp gRPC server.

gadget-container/gadgettracermanager/auth.go
```diff
@@ -0,0 +1,86 @@
...
+			return status.Errorf(codes.Unauthenticated, "invalid certificate")
+		}
+		tlsInfo := p.AuthInfo.(credentials.TLSInfo)
+		subject := tlsInfo.State.VerifiedChains[0][0].Subject
+		log.Infof("stream request from %s", subject)
+		return handler(srv, serverStream)
+	}
+}
+
+func unaryInterceptor() func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
```
mauriciovasquezbernal, line 49: I'm wondering if this would make sense to declare function directly and avoid returning a lambda? 

```
func unaryInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
	p, ok := peer.FromContext(ctx)
	if !ok {
		return nil, status.Errorf(codes.Unauthenticated, "invalid certificate")
	}
	tlsInfo := p.AuthInfo.(credentials.TLSInfo)
	subject := tlsInfo.State.VerifiedChains[0][0].Subject
	log.Infof("request from %s", subject)
	return handler(ctx, req)
}
```

And use `unaryInterceptor` instead of `unaryInterceptor()` in main.go.

internal/cert-helpers/cert-helpers.go
```diff
@@ -0,0 +1,123 @@
...
+	privateKeyDER, err := x509.MarshalPKCS8PrivateKey(privateKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("self-signing cert: %w", err)
+	}
+
+	return certDER, privateKeyDER, nil
+}
+
+// GenerateCertificate generates a new private key and a certificate signed by the given CA
+func GenerateCertificate(name string, usage x509.ExtKeyUsage, validFor time.Duration, caCertDER, caPrivateKeyDER []byte) ([]byte, []byte, error) {
```
mauriciovasquezbernal, line 82: nit: breaks those too long lines? (We don't have a specific number but lines > 120 chars seems difficult to handle for me)

internal/cert-helpers/cert-helpers.go
```diff
@@ -0,0 +1,123 @@
+// Copyright 2022 The Inspektor Gadget authors
```
mauriciovasquezbernal, line 1: Any reason to have this as a different package. Maybe we can fusion `internal/certs` and `internal/cert-helpers` packages?

internal/certs/certs.go
```diff
@@ -0,0 +1,69 @@
...
+
+	cert, privateKey, err := certhelpers.GenerateCertificate(node, keyUsage, duration, caCert, caPrivateKey)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("generate server key: %w", err)
+	}
+
+	return cert, privateKey, ca, nil
+}
+
+func GenerateClientCertificate(ctx context.Context, node string, clientset *kubernetes.Clientset) ([]byte, []byte, *x509.Certificate, error) {
```
mauriciovasquezbernal, line 63: What's the performance overhead of generating a different certificate each time? I was wondering if we could generate this certificate when deploy is run and save in the kubeconfig file or a similar one? (but maybe it gets complicated for the user to manage two different authentication files, kubeconfig for kubectl and another one for kubectl-gadget).

mauriciovasquezbernal (commented):
Looks very good. Some comments on the new authentication logic.

gadget-container/gadgettracermanager/auth.go
flyth: Yeah, I had those for token authentication earlier where I would use the outer function to have that as its parameter - doesn't make sense this way for these functions, but they're only here for debugging right now (but we could use it to prepend the logs with the CN of the incoming certificate for example).

internal/certs/certs.go
flyth: Really neglectable - it's using elliptic curve keys, it's pretty fast. If you test the branch your client generates them with each trace call and you probably won't notice that it's doing it. Nevertheless, I'd like to add some kind of caching later on for the user side at least (in ~/.ig or something like that).

mauriciovasquezbernal:
@flyth can we close this one now that we have https://github.com/inspektor-gadget/inspektor-gadget/pull/1396?

Issue #881, opened 2022-08-29
Do not use "fail to ...", "error ...(ing)" in error messages


mauriciovasquezbernal:
Having the "fail to", "error ..." prefixes causes error messages to be weird as they end up being something like: "fail to foo: error while X: failed to ...", we can remove those prefixes and only indicate the reason. 

Ref https://github.com/kinvolk/inspektor-gadget/pull/821/files#r957269992.

invidian:
Here is some simple [semgrep](https://semgrep.dev/) rules which I always use to catch things like that:
```yaml
rules:
  - id: bad-error-wrap-wording
    patterns:
      - pattern-regex: fmt\.Errorf\(".?([Ff]ailed|[]Uu]nable|[Cc]ould|([Ee]rror)|([Cc]an[ ]?not))
    message: Phrases "fail", "failed", "unable", "could not", "cannot", "can not" should not be used when wrapping errors. Use verb in continuous form instead!
    languages: [go]
    severity: ERROR
  - id: printed-messages-must-be-captialized
    patterns:
      - pattern-regex: (fmt.Print(ln|f)|\bt\.(Error|Fatal|Log)(|f))\("[a-z]
    message: Printed messages must be capitalized.
    languages: [go]
    severity: ERROR
  - id: strings-must-be-quoted-using-q
    patterns:
      - pattern-regex: "'%s'"
    message: When formatting, strings must be quoted using %q instead of e.g. '%s'.
    languages: [go]
    severity: ERROR
  - id: error-wrapping-must-be-preceded-by-space
    patterns:
      - pattern-regex: '[^ ]%w'
    message: When wrapping errors, %w must be preceded by space.
    languages: [go]
    severity: ERROR
  - id: files-must-not-have-trailing-whitespace
    patterns:
      - pattern-regex: '[[:blank:]]$'
    message: Files must not have any trailing whitespace.
    languages: [generic]
    severity: ERROR
  - id: files-must-not-have-trailing-newlines
    patterns:
      - pattern-regex: '\n\n\Z'
    message: Files must not have any trailing newlines.
    languages: [generic]
    severity: ERROR
  - id: all-lines-must-end-with-newline
    patterns:
      - pattern-regex: '\S\z'
    message: All lines must end with newline.
    languages: [generic]
    severity: ERROR
```

blanquicet:
What do you suggest using? For instance here:

```golang
	t, err := tracer.NewTracer(config, enricher, eventCallback)
[...]

PR #1097, opened 2022-11-03
introduce new gadget: tcpdump

flyth:
This PR introduces a new gadget that allows distributed packet capturing and inspection. It uses libpcap to compile rules like the ones used in tcpdump (hence the name) into cBPF and filters them using a raw socket with the compiled filter attached.

The received stream will by default be analyzed using gopacket but can optionally be forwarded to a local tcpdump or wireshark instance (it uses the pcap format).

Right now packets are encoded using JSON (like all the other gadgets), but I hope that we get better support for binary content using #1096.

### Additional Features

The gadget uses the pcapng file format and uses two features not yet commonly used:

#### Interface Names

Each container will have its own interface registered in the pcapng file - that makes it easy (especially in Wireshark) to differentiate the traffic. Additionally, as metadata, we add podname and namespace.

#### Name Resolution Blocks

Before starting to stream traffic into the pcapng file (or to tcpdump/WireShark), we generate DNS records for all pods and services and populate the file with them. That way you can easier see what is communicating with what inside the cluster.
Afterwards it will add updates as they come in (new pods, new services etc.)

### Sample output

Using gopacket (default)

Using gopacket

Using tcpdump

Using tcpdump

Using tcpdump with additional parameters

Using tcpdump with parameters

Using WireShark with interface names and name resolution blocks inserted (need to enable name resolution in WireShark for that)

wireshark

### Todo

* [x] add snaplen parameter
* [ ] use binary encoding
* [x] add option to save to pcap file
* [ ] add tests
* [ ] decide whether to have a new category "tools" or just place it into "trace"
* [ ] add more information about dropped packets
* [x] use pcapng as file format and enrich it with interface names and name resolution blocks
* [ ] make it work with local-gadget
* [x] optimize for speed as this is probably a high throughput gadget

### DIFF
--- a/cmd/kubectl-gadget/trace/tcpdump.go
+++ b/cmd/kubectl-gadget/trace/tcpdump.go
@@ -1,0 +1,399 @@
+// Copyright 2019-2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+import (
+	"bytes"
+	"context"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"net"
+	"os"
+	"os/exec"
+	"runtime"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcapgo"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	commontrace "github.com/inspektor-gadget/inspektor-gadget/cmd/common/trace"
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/tcpdump/types"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
+	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+)
+
+type Decoder string
+
+type nameEvent struct {
+	IP   net.IP
+	Name string
+}
+
+type TCPDumpParser struct {
+	commonutils.BaseParser[types.Event]
+	pcapngWriter   *pcapgo.NgWriter
+	writer         io.Writer
+	decoder        Decoder
+	snapLen        int
+	filter         string
+	interfaces     map[string]int
+	interfacesLock sync.RWMutex
+	events         chan *types.Event // channel for incoming packets
+	nameEvents     chan *nameEvent   // channel for dns resolution information
+}
+
+const (
+	DecoderWireshark = "wireshark"
+	DecoderTCPDump   = "tcpdump"
+	DecoderExternal  = "external"
+	DecoderInternal  = "internal"
+	DecoderFile      = "file"
+)
+
+var decoderCmd *exec.Cmd
+
+// populate DNS entries
+func (p *TCPDumpParser) populateDNS(ctx context.Context) ([]byte, error) {
+	client, err := k8sutil.NewClientsetFromConfigFlags(utils.KubernetesConfigFlags)
+	if err != nil {
+		return nil, err
+	}
+
+	go func() {
+		list, _ := client.CoreV1().Pods("").Watch(ctx, metav1.ListOptions{})
+		for info := range list.ResultChan() {
+			pod := info.Object.(*corev1.Pod)
+			for _, ip := range pod.Status.PodIPs {
+				if parsedIP := net.ParseIP(ip.IP); parsedIP != nil {
+					p.nameEvents <- &nameEvent{
+						IP:   parsedIP,
+						Name: fmt.Sprintf("%s.%s.pod", strings.Replace(ip.IP, ".", "-", -1), pod.Namespace),
+					}
+				}
+			}
+		}
+	}()
+	go func() {
+		list, _ := client.CoreV1().Services("").Watch(ctx, metav1.ListOptions{})
+		for info := range list.ResultChan() {
+			svc := info.Object.(*corev1.Service)
+			for _, ip := range svc.Spec.ClusterIPs {
+				if parsedIP := net.ParseIP(ip); parsedIP != nil {
+					p.nameEvents <- &nameEvent{
+						IP:   parsedIP,
+						Name: fmt.Sprintf("%s.%s.svc", svc.Name, svc.Namespace),
+					}
+				}
+			}
+			for _, ip := range svc.Spec.ExternalIPs {
+				if parsedIP := net.ParseIP(ip); parsedIP != nil {
+					p.nameEvents <- &nameEvent{
+						IP:   parsedIP,
+						Name: fmt.Sprintf("%s.%s.svc", svc.Name, svc.Namespace),
+					}
+				}
+			}
+			if svc.Spec.LoadBalancerIP != "" {
+				if parsedIP := net.ParseIP(svc.Spec.LoadBalancerIP); parsedIP != nil {
+					p.nameEvents <- &nameEvent{
+						IP:   parsedIP,
+						Name: fmt.Sprintf("%s.%s.svc", svc.Name, svc.Namespace),
+					}
+				}
+			}
+		}
+	}()
+	return nil, nil
+}
+
+func getDNSBlock(ip net.IP, name string) []byte {
+	// TODO: add support for IPv6
+	ipv4 := ip.To4()
+	if ipv4 == nil || len(ipv4) != net.IPv4len {
+		return nil
+	}
+
+	nsblock := bytes.NewBuffer(nil)
+
+	header := make([]byte, 8)
+	binary.LittleEndian.PutUint32(header[0:], 0x00000004) // Block type
+	nsblock.Write(header)
+
+	plen := 4 + 4 + len([]byte(name)) + 1 // header + ip + name + \0 terminator
+	if plen%4 != 0 {
+		// pad to 32bit
+		plen += 4 - plen%4
+	}
+	record := make([]byte, plen)
+
+	binary.LittleEndian.PutUint16(record[0:2], 0x0001)                        // ipv4
+	binary.LittleEndian.PutUint16(record[2:4], uint16(4+len([]byte(name))+1)) // length of name + \0 terminator
+
+	copy(record[4:8], ipv4[0:4])
+	copy(record[8:], []byte(name))
+	nsblock.Write(record)
+
+	footer := make([]byte, 4)
+	nsblock.Write(footer) // nrb_record_end + length
+
+	nsblock.Write(footer) // Block size
+
+	out := nsblock.Bytes()
+	binary.LittleEndian.PutUint32(out[4:], uint32(len(out)))
+	binary.LittleEndian.PutUint32(out[len(out)-4:], uint32(len(out)))
+	return out
+}
+
+func newTCPDumpCmd() *cobra.Command {
+	commonFlags := &utils.CommonFlags{
+		OutputConfig: commonutils.OutputConfig{
+			OutputMode:    commonutils.OutputModeCustom,
+			CustomColumns: []string{},
+		},
+	}
+
+	var decoderParam string
+	var decoderArgsParam string
+	var decoderBinaryParam string
+	var snapLen int
+	var filenameParam string
+	var disableDNSPopulation bool
+
+	cmd := &cobra.Command{
+		Use:   "tcpdump",
+		Short: "Trace packets",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			decoderArgs := []string{}
+			var decoder Decoder
+			var decoderBinary string
+			var ngw *pcapgo.NgWriter
+
+			// Writer, used for external output (pcapng)
+			var out io.Writer
+
+			switch Decoder(decoderParam) {
+			case DecoderWireshark:
+				decoder = DecoderExternal
+				decoderArgs = []string{"-k", "-i", "-"}
+				decoderBinary = "wireshark"
+			case DecoderTCPDump:
+				decoder = DecoderExternal
+				decoderArgs = []string{"-r", "-"}
+				decoderBinary = "tcpdump"
+			case DecoderInternal:
+				decoder = DecoderInternal
+			case DecoderFile:
+				decoder = DecoderFile
+			default:
+				return errors.New("unknown decoder")
+			}
+
+			if decoder == DecoderFile {
+				if filenameParam == "" {
+					return fmt.Errorf("no filename specified")
+				}
+				f, err := os.Create(filenameParam)
+				if err != nil {
+					return fmt.Errorf("creating file: %w", err)
+				}
+				out = f
+			}
+
+			if decoder == DecoderExternal {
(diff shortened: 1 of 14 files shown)


pkg/gadgets/trace/tcpdump/tracer/tracer.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func NewTracer(config *Config) (*Tracer, error) {
+	t := &Tracer{
+		attachments: map[string]*link{},
+	}
+
+	ins, err := TcpdumpExprToBPF(config.FilterString, layers.LinkTypeEthernet, config.SnapLen)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compile tcpdump expression to bpf")
```
mauriciovasquezbernal, line 59: nit: Avoid `failed to ` in error messages -> https://github.com/inspektor-gadget/inspektor-gadget/issues/881

pkg/gadgets/trace/tcpdump/tracer/tracer.go
```diff
@@ -0,0 +1,158 @@
...
+	// value: Tracelet
+	attachments map[string]*link
+}
+
+func NewTracer(config *Config) (*Tracer, error) {
+	t := &Tracer{
+		attachments: map[string]*link{},
+	}
+
+	ins, err := TcpdumpExprToBPF(config.FilterString, layers.LinkTypeEthernet, config.SnapLen)
```
mauriciovasquezbernal, line 57: Perhaps `TcpdumpExprToBPF()` could return `[]bpf.RawInstruction` directly to avoid the `Disassemble()`/`Assemble()` calls?

cmd/kubectl-gadget/trace/tcpdump.go
```diff
@@ -0,0 +1,244 @@
...
+			return tcpdumpGadget.Run()
+		},
+	}
+
+	utils.AddCommonFlags(cmd, commonFlags)
+	cmd.Flags().StringVar(&decoderParam, "decoder", "internal", "name of the decoder to use (either tcpdump, wireshark, internal or file)")
+	cmd.Flags().StringVar(&decoderArgsParam, "decoder-args", "", "arguments to forward to decoder")
+	cmd.Flags().StringVar(&decoderBinaryParam, "decoder-binary", "", "path to decoder binary (defaults to 'wireshark' or 'tcpdump' depending on decoder)")
+	cmd.Flags().StringVar(&filenameParam, "out-file", "", "output file name")
+	cmd.Flags().IntVar(&snapLen, "snaplen", 68, "number of bytes to capture")
```
mauriciovasquezbernal, line 179: bytes per packet, right?

cmd/kubectl-gadget/trace/tcpdump.go
```diff
@@ -0,0 +1,244 @@
...
+				name:        "tcpdump",
+				commonFlags: commonFlags,
+				parser:      NewTCPDump(&commonFlags.OutputConfig, filter, snapLen, decoder, ngw),
+				params: map[string]string{
+					types.FilterStringParam: filter,
+					types.SnapLenParam:      strconv.Itoa(snapLen),
+				},
+			}
+
+			return tcpdumpGadget.Run()
```
mauriciovasquezbernal, line 170: How about using `utils.RunTraceAndPrintStream` here? I think we're not interested in providing columns nor json support, so I think it's fine not to reuse the common logic for all gadgets.

cmd/kubectl-gadget/trace/tcpdump.go
```diff
@@ -0,0 +1,244 @@
...
+		},
+	}
+
+	utils.AddCommonFlags(cmd, commonFlags)
+	cmd.Flags().StringVar(&decoderParam, "decoder", "internal", "name of the decoder to use (either tcpdump, wireshark, internal or file)")
```
mauriciovasquezbernal, line 175: Would use `-o tcpdump`, `-o wireshark`, `-o file=/tmp/foo.pcap`? (It probably requires some extra work but maybe it's cleaner?)

cmd/kubectl-gadget/trace/tcpdump.go
```diff
@@ -0,0 +1,244 @@
...
+	p.interfacesLock.RLock()
+	if id, ok := p.interfaces[event.Container]; ok {
+		p.interfacesLock.RUnlock()
+		return id
+	}
+	p.interfacesLock.RUnlock()
+	// Define new interface
+	p.interfacesLock.Lock()
+	id, err := p.pcapngWriter.AddInterface(pcapgo.NgInterface{
+		Name:        event.Container,
```
mauriciovasquezbernal, line 206: offtopic: Is there anyway to print this interface name in tcpdump? I found https://serverfault.com/questions/224698/how-to-display-interface-in-tcpdump-output-flow but it seems rather complex.

mauriciovasquezbernal (commented):
Looks very good overall, nice work. I made some comments about different parts, I'll make another round later on.

cmd/kubectl-gadget/trace/tcpdump.go
flyth: Yeah, makes sense to leave out some of the unnecessary logic there. But maybe we should wait for #1096 (and a binary addition) to optimize flow of the data.

cmd/kubectl-gadget/trace/tcpdump.go
flyth: tcpdump seems to not handle the additional information (well), I haven't found a clean solution either. WireShark does however fully support it.

mauriciovasquezbernal (commented):
I'm trying it but I get the following error:

```
./kubectl-gadget trace tcpdump --decoder file --out-file /ebpf/capture.pcap
panic: runtime error: slice bounds out of range [:4] with capacity 0

goroutine 82 [running]:
github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/trace.getDNSBlock({0x0, 0x0, 0x0}, {0xc000a825d0, 0x15})
	/home/mvb/kinvolk/ebpf/inspektor-gadget/cmd/kubectl-gadget/trace/tcpdump.go:144 +0x317
github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/trace.(*TCPDumpParser).run(0xc00019d830)
	/home/mvb/kinvolk/ebpf/inspektor-gadget/cmd/kubectl-gadget/trace/tcpdump.go:410 +0x410
created by github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/trace.NewTCPDump
	/home/mvb/kinvolk/ebpf/inspektor-gadget/cmd/kubectl-gadget/trace/tcpdump.go:339 +0x2ea
```

Any idea of what can be wrong?

pkg/gadgets/trace/tcpdump/tracer/compiler/compiler_tcpdump.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/google/gopacket/layers"
+	"golang.org/x/net/bpf"
+)
+
+func TcpdumpExprToBPF(filterExpr string, linkType layers.LinkType, snapLen int) ([]bpf.RawInstruction, error) {
+	if linkType != layers.LinkTypeEthernet {
+		return nil, errors.New("unsupported linkType")
+	}
+
+	// We'll use an installed tcpdump to compile the filter
```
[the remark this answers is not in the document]
flyth: In general, I also think a dependency on a lib would be better - but this file is only used for local-gadget that we compile statically. If depending on the lib, this would mean local-gadget wouldn't start without a (compatible) libpcap version installed - compiling libpcap statically into local-gadget seems non-trivial and could also be problematic regarding licenses (I didn't check). That's why I think this is the better solution here. But yes, that means that even if you would only want to capture to a file that you'd need tcpdump installed. Maybe we can provide an option to read the compiled filter from a file so that you could compile it on another computer and just copy/paste the instructions to the target machine. We could also provide a couple of precompiled generic rules where we just replace some variables (like ports, ips), but I'm not sure how far we should go that route...

flyth:
> Any idea of what can be wrong?

getDNSBlock({0x0, 0x0, 0x0}, ...) - That is the IP (3 zero bytes) the func got - that looks odd. Probably need to add a check in populateDNS() to validate the IPs to make sure this doesn't happen.

pkg/gadget-collection/gadgets/trace/tcpdump/gadget.go
```diff
@@ -0,0 +1,281 @@
...
+		genPubSubKey(trace.ObjectMeta.Namespace+"/"+trace.ObjectMeta.Name),
+		*gadgets.ContainerSelectorFromContainerFilter(trace.Spec.Filter),
+		containerEventCallback,
+	)
+
+	for _, c := range existingContainers {
+		err := attachContainerFunc(c)
+		if err != nil {
+			log.Warnf("Warning: couldn't attach BPF program: %s", err)
+			break
```
mqasimsarfraz, line 261: Do we need to call `detachContainerFunc` here as well?

cmd/kubectl-gadget/trace/tcpdump.go
```diff
@@ -0,0 +1,399 @@
...
+		},
+	}
+
+	utils.AddCommonFlags(cmd, commonFlags)
+	cmd.Flags().StringVar(&decoderParam, "decoder", "internal", "name of the decoder to use (either tcpdump, wireshark, internal or file)")
+	cmd.Flags().StringVar(&decoderArgsParam, "decoder-args", "", "arguments to forward to decoder")
+	cmd.Flags().StringVar(&decoderBinaryParam, "decoder-binary", "", "path to decoder binary (defaults to 'wireshark' or 'tcpdump' depending on decoder)")
+	cmd.Flags().StringVar(&filenameParam, "out-file", "", "output file name")
+	cmd.Flags().BoolVar(&disableDNSPopulation, "disable-dns", false, "disable DNS population from kubernetes")
+	cmd.Flags().IntVar(&snapLen, "snaplen", 68, "number of bytes to capture per packet")
```
mqasimsarfraz, line 292: While tracing DNS requests I am seeing error because of packet truncation:
```
PACKET: 68 bytes, truncated
- Layer 1 (14 bytes) = Ethernet	{Contents=[..14..] Payload=[..54..] SrcMAC=02:42:ac:11:00:03 DstMAC=02:42:ac:11:00:02 EthernetType=IPv4 Length=0}
- Layer 2 (20 bytes) = IPv4	{Contents=[..20..] Payload=[..34..] Version=4 IHL=5 TOS=0 Length=175 Id=34611 Flags=DF FragOffset=0 TTL=64 Protocol=UDP Checksum=64653 SrcIP=10.96.0.10 DstIP=172.17.0.2 Options=[] Padding=[]}
- Layer 3 (08 bytes) = UDP	{Contents=[..8..] Payload=[..26..] SrcPort=53(domain) DstPort=54502 Length=155 Checksum=46889}
- Layer 4 (26 bytes) = DecodeFailure	Packet decoding error: dns name uncomputable: invalid index
```
Setting snaplen to higher values gets rid of errors.

cmd/kubectl-gadget/trace/tcpdump.go
```diff
@@ -0,0 +1,399 @@
...
+	filter         string
+	interfaces     map[string]int
+	interfacesLock sync.RWMutex
+	events         chan *types.Event // channel for incoming packets
+	nameEvents     chan *nameEvent   // channel for dns resolution information
+}
+
+const (
+	DecoderWireshark = "wireshark"
+	DecoderTCPDump   = "tcpdump"
```
mqasimsarfraz, line 71: While tracing with `./kubectl-gadget trace tcpdump --decoder localhost` I am seeing `localhost` everywhere:
```
→ ./kubectl-gadget trace tcpdump --decoder tcpdump
reading from file -, link-type EN10MB (Ethernet)
17:29:20.556811 IP localhost.60785 > localhost.domain: 24363+[|domain]
17:29:20.556913 IP localhost.60785 > localhost.domain: 25724+[|domain]
17:29:20.557411 IP localhost.domain > localhost.60785: 24363*-[|domain]
17:29:20.557566 IP localhost.domain > localhost.60785: 25724*-[|domain]

```

cmd/kubectl-gadget/trace/tcpdump.go
flyth: I first thought these would be panics - but those errors are just coming from truncated packets, so I think this is fine.

cmd/kubectl-gadget/trace/tcpdump.go
flyth: I also got a lot of localhost packets - but those seem valid (from 127.0.0.1 to 127.0.0.1). Could you please recheck to see if you also see inbound/outbound packets when doing a curl or such?

The interface name would greatly help to show from which container those local packets are coming, but that's only visible in WireShark as it seems.

mqasimsarfraz: I am running ` nslookup kubernetes.default.svc.cluster.local` in a test pod with `alpine` image. If try to see the IP addresses I am seeing following:
```
→ ./kubectl-gadget trace tcpdump --decoder tcpdump --decoder-args -n 
reading from file -, link-type EN10MB (Ethernet)
16:49:59.081782 IP 172.17.0.2.33583 > 10.96.0.10.53: 60708+[|domain]
16:49:59.081894 IP 172.17.0.2.33583 > 10.96.0.10.53: 61948+[|domain]
16:49:59.100071 IP 10.96.0.10.53 > 172.17.0.2.33583: 60708 1/0/0 (54)
16:49:59.100731 IP 10.96.0.10.53 > 172.17.0.2.33583: 61948 1/0/0 (66)
```
So I do see a DNS requests going towards `kube-dns` service but not sure why they are named localhost. I will check again if the setup has some weird config. It is minikube node.

pkg/gadgets/trace/tcpdump/tracer/compiler/compiler_tcpdump.go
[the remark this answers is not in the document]
mauriciovasquezbernal: > If depending on the lib, this would mean local-gadget wouldn't start without a (compatible) libpcap version installed

Do we have any special requirements on the libpcap version? I'd think that libpcap offers a quite stable API and that in most of the cases it's not a problem for us. 

> compiling libpcap statically into local-gadget seems non-trivial

I agree. Perhaps we can have a mix in local-gadget, most of the functionality is statically compiled in it, but some extra functionality could require to have installed some dependencies. I think we could use dlopen() to check at runtime if a given library is available. 

Anyway, this seems too much for this PR, I'm totally fine with the dependency on the tcpdump binary for now.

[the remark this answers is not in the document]
flyth: > I think we could use dlopen() to check at runtime if a given library is available.

Yeah, true. That would mean we have to rewrite a couple of things from the go pcap library, though. I'll look into it later on, then.

alban:
What do you think about translating the port name as defined by users in the Kubernetes service instead of using /etc/services? Note that the mapping between ports and names will be different for each service/endpoint.

```
$ kubectl get svc -n kube-system   kube-dns -o yaml
...
  ports:
  - name: dns
    port: 53
    protocol: UDP
    targetPort: 53
  - name: dns-tcp
    port: 53
    protocol: TCP
    targetPort: 53
  - name: metrics
    port: 9153
    protocol: TCP
    targetPort: 9153
```

flyth:
> What do you think about translating the port name as defined by users in the Kubernetes service instead of using /etc/services? Note that the mapping between ports and names will be different for each service/endpoint.
> 
> ```
> $ kubectl get svc -n kube-system   kube-dns -o yaml
> ...
>   ports:
>   - name: dns
>     port: 53
>     protocol: UDP
>     targetPort: 53
>   - name: dns-tcp
>     port: 53
>     protocol: TCP
>     targetPort: 53
>   - name: metrics
>     port: 9153
>     protocol: TCP
>     targetPort: 9153
> ```

I like the idea - I'll look at the pcapng spec to see if there's actually a proper way to add that info (without using an optional description field) - usually those portnames are resolved by tcpdump or wireshark and not included in the capture file.

alban:
The following code is translating the cBPF code generated by libpcap into a eBPF function that is replaced with freplace at runtime:

https://github.com/retis-org/retis/blob/53e42a5a802b08c51ac23d3a5a0e32228ef96589/src/core/filters/packets/ebpf.rs#L315

flyth:
> The following code is translating the cBPF code generated by libpcap into a eBPF function that is replaced with freplace at runtime...

Nice! I've looked at [this](https://github.com/cloudflare/cbpfc) previously, but didn't manage to get it to work instantly, so I went with cBPF for the first iteration. But I'd love to get this updated and extended some day :)

Issue #887, opened 2022-08-30
local-gadget: Run trace/fsslower gadgets using tracers instead of the Tracer CR


blanquicet:
## Current situation

`local-gadget` does not support `fsslower` trace gadget.

## Ideal future situation

Add `local-gadget trace fsslower` by:

- Resuing the command definition (description and flags) already created for `kubectl-gadget`.
- Verifying no regression tests are introduced to `kubectl-gadget`
- Updating the documentation.

The PRs https://github.com/kinvolk/inspektor-gadget/pull/896 and https://github.com/kinvolk/inspektor-gadget/pull/937 added support for `bind`, `exec` and `tcp` trace gadgets, so they can be used as a reference to solve this issue.

See https://github.com/kinvolk/inspektor-gadget/issues/864 for further details.

PR #1102, opened 2022-11-04
localgadget: Support trace fsslower gadget

mauriciovasquezbernal:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/887

### DIFF
--- a/integration/local-gadget/trace_fsslower_test.go
+++ b/integration/local-gadget/trace_fsslower_test.go
@@ -1,0 +1,73 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


integration/local-gadget/trace_fsslower_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+			return ExpectEntriesToMatch(output, normalize, expectedEntry)
+		},
+	}
+
+	// TODO: traceFsslowerCmd should moved up the list once we can trace new cri-o containers.
+	// Issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/1018
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		BusyboxPodCommand(ns, cmd),
```
blanquicet, line 69: nit:

```suggestion
		BusyboxPodCommand(ns, "echo 'this is foo' > foo && while true; do cat foo && sleep 0.1; done"),
```

For `exec`, we defined the cmd on top because it was used by `ExpectedOutputFn`. In this case, I don't see the advantage of defining it there.

blanquicet (approved):
LGTM

PR #1104, opened 2022-11-07
integrationtest/TestAuditSeccomp: Validate json output

burak-ok:
# integrationtest/TestAuditSeccomp: Validate json output

Last test to be ported over to json. See https://github.com/inspektor-gadget/inspektor-gadget/issues/719

Note: The container name was different in this test. Was that on purpose?


blanquicet (approved):
LGTM!

mauriciovasquezbernal:
> Note: The container name was different in this test. Was that on purpose?

I don't think so. Probably as this is created using a yaml we ended with two different names.

PR #1105, opened 2022-11-07
Use `raw_tp` for `profile block-io`

eiffel-fl:
Hi.

Since the change added in #1089, the CI does not pass for [AKS](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3388129577/jobs/5638848469#step:9:296).
Indeed, AKS relies on the standard tracer, which prints some empty line when used with `--json`.
As a consequence, the integration test cannot translate an empty line to an event and so it fails.

This PR modifies `biolatency` eBPF code to use `raw_tp` rather than `tp_btf`, so this tool can work on AKS (and other older kernels too).

Best regards.


alban (approved):
This patch LGTM, so our CO-RE gadget supports more kernels.

But shouldn't we also consider one of the following changes?
- Don't let the bcc gadget print empty lines
- Fix integration tests so they don't fail on empty lines from the bcc gadget

PR #1106, opened 2022-11-08
pkg/local-gadget-manager: Use nginx container to test network-graph.

eiffel-fl:
Hi.

In this PR, rather than querying `1.1.1.1` which leaded to buggy behavior recently, I run an `nginx` container which will be queried.

Best regards.

### DIFF
--- a/pkg/local-gadget-manager/local-gadget-manager_test.go
+++ b/pkg/local-gadget-manager/local-gadget-manager_test.go
@@ -593,92 +593,90 @@ func TestNetworkGraph(t *testing.T) {
 		t.Fatalf("Failed to start the tracer: %s", err)
 	}
 
-	runTestContainer(t, containerName, "docker.io/library/alpine", "wget https://1.1.1.1", "")
+	runTestContainer(t, containerName, "docker.io/library/nginx", "nginx && curl 127.0.0.1", "")
 
 	stop := make(chan struct{})
+	defer close(stop)
 
 	ch, err := localGadgetManager.StreamTraceResourceOutput("my-tracer", stop)
 	if err != nil {
 		t.Fatalf("Failed to get stream: %s", err)
 	}
 
-	var event networktypes.Event
-	var expectedEvent networktypes.Event
-	var result string
-
-	// check that attached message is sent
-	result = <-ch
-	event = networktypes.Event{}
-	if err := json.Unmarshal([]byte(result), &event); err != nil {
-		t.Fatalf("failed to unmarshal json: %s", err)
-	}
-
-	expectedEvent = networktypes.Event{
-		Event: eventtypes.Event{
-			Type: eventtypes.DEBUG,
-			CommonData: eventtypes.CommonData{
-				Node:      "local",
-				Namespace: "default",
-				Pod:       "test-local-gadget-network-graph001",
+	expectedEvents := []networktypes.Event{
+		{
+			Event: eventtypes.Event{
+				Type: eventtypes.DEBUG,
+				CommonData: eventtypes.CommonData{
+					Node:      "local",
+					Namespace: "default",
+					Pod:       "test-local-gadget-network-graph001",
+				},
+				Message: "tracer attached",
 			},
-			Message: "tracer attached",
+		},
+		{
+			Event: eventtypes.Event{
+				Type: eventtypes.NORMAL,
+				CommonData: eventtypes.CommonData{
+					Node:      "local",
+					Namespace: "default",
+					Pod:       "test-local-gadget-network-graph001",
+				},
+			},
+			PktType: "OUTGOING",
+			Proto:   "tcp",
+			// There's no place like
+			Addr: "127.0.0.1",
+			Port: 80,
+		},
+		{
+			Event: eventtypes.Event{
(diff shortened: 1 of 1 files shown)


pkg/local-gadget-manager/local-gadget-manager_test.go
```diff
@@ -647,8 +647,35 @@ func TestNetworkGraph(t *testing.T) {
 		},
 		PktType: "OUTGOING",
 		Proto:   "tcp",
-		Addr:    "1.1.1.1",
-		Port:    443,
+		// There's no place like
```
blanquicet, line 650: I don't get this comment.

pkg/local-gadget-manager/local-gadget-manager_test.go
```diff
@@ -593,7 +593,7 @@ func TestNetworkGraph(t *testing.T) {
 		t.Fatalf("Failed to start the tracer: %s", err)
 	}
 
-	runTestContainer(t, containerName, "docker.io/library/alpine", "wget https://1.1.1.1", "")
+	runTestContainer(t, containerName, "docker.io/library/nginx", "nginx & sleep 1; curl 127.0.0.1", "")
```
alban, line 596: It seems you can get rid of the sleep 1 and just run:
```
nginx ; curl 127.0.0.1
```
because the nginx process will daemonize and go in the background when it's ready.

Tested with:
```
$ docker run -ti --rm nginx sh -c 'nginx ; curl 127.0.0.1'
2022/11/08 15:46:46 [notice] 7#7: using the "epoll" event method
2022/11/08 15:46:46 [notice] 7#7: nginx/1.23.2
2022/11/08 15:46:46 [notice] 7#7: built by gcc 10.2.1 20210110 (Debian 10.2.1-6) 
2022/11/08 15:46:46 [notice] 7#7: OS: Linux 5.18.9-200.fc36.x86_64
2022/11/08 15:46:46 [notice] 7#7: getrlimit(RLIMIT_NOFILE): 1073741816:1073741816
2022/11/08 15:46:46 [notice] 8#8: start worker processes
2022/11/08 15:46:46 [notice] 8#8: start worker process 9
2022/11/08 15:46:46 [notice] 8#8: start worker process 10
2022/11/08 15:46:46 [notice] 8#8: start worker process 11
2022/11/08 15:46:46 [notice] 8#8: start worker process 13
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a href="http://nginx.com/">nginx.com</a>.</p>

<p><em>Thank you for using nginx.</em></p>
</body>
</html>
127.0.0.1 - - [08/Nov/2022:15:46:46 +0000] "GET / HTTP/1.1" 200 615 "-" "curl/7.74.0" "-"
```

blanquicet, line 596: nit: AFAIK, `nginx` command returns when it finishes to start up the server, so it could simply be `nginx; curl 127.0.0.1`.

alban:
> What are the guarantees we have regarding the order of events we receive?
> If we have none, we should not test the order but rather the presence.

Correct, the network-graph does not keep the order. So the test must be modified to accept any order.

pkg/local-gadget-manager/local-gadget-manager_test.go
alban: The line below is 127.0.0.1, aka the 'home' IP. The comment says 'There's no place like home'.

blanquicet: I haven't ever heard about it 😅 . I have always called it localhost.

pkg/local-gadget-manager/local-gadget-manager_test.go
```diff
@@ -647,8 +647,35 @@ func TestNetworkGraph(t *testing.T) {
...
+	expectedEvent = networktypes.Event{
+		Event: eventtypes.Event{
+			Type: eventtypes.NORMAL,
+			CommonData: eventtypes.CommonData{
+				Node:      "local",
+				Namespace: "default",
+				Pod:       "test-local-gadget-network-graph001",
+			},
+		},
+		PktType: "HOST",
```
blanquicet, line 675: > Correct, the network-graph does not keep the order. So the test must be modified to accept any order.
>
I also reproduced it. Maybe we could capture all the `eventtypes.NORMAL` events and reuse this to verify the expected entries:

https://github.com/inspektor-gadget/inspektor-gadget/blob/14da590a4d5edf7d825f36cbb20fc1ef0612d184/integration/helpers.go#L107-L118

blanquicet (changes requested):

blanquicet (approved):
LGTM.

PR #1110, opened 2022-11-09
integration-tests: Improve deployment workflow for IG/SPO

### LINKED ISSUE #976: integration: Improve deployment workflow for SPO
Currently we have scenarios in which if a user runs integration test against a cluster with already deployed SPO we can end up with cleaning everything even if we didn't deploy it. Possible solution:

- Check for already deployed SPO and skip cleaning in such scenarios.
- Change default make target to not deploy such components.

Ref [#970 Comment ](https://github.com/kinvolk/inspektor-gadget/pull/970#discussion_r976599658)

mqasimsarfraz:
If we have a cluster with pre-installed IG/SPO we end up removing them if '-no-deploy-ig/-no-deploy-spo' isn't specified. Here we add some logic to detect if these components are already installed and skip deployment/cleanup. This can also be helpful to skip SPO installation against ARO cluster without extra config. 

Fixes #976

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -67,9 +67,6 @@ var (
 	// image such as ghcr.io/inspektor-gadget/inspektor-gadget:latest
 	image = flag.String("image", "", "gadget container image")
 
-	doNotDeployIG  = flag.Bool("no-deploy-ig", false, "don't deploy Inspektor Gadget")
-	doNotDeploySPO = flag.Bool("no-deploy-spo", false, "don't deploy the Security Profiles Operator (SPO)")
-
 	k8sDistro = flag.String("k8s-distro", "", "allows to skip tests that are not supported on a given Kubernetes distribution")
 	k8sArch   = flag.String("k8s-arch", "amd64", "allows to skip tests that are not supported on a given CPU architecture")
 )
@@ -137,7 +134,8 @@ func testMain(m *testing.M) int {
 	initCommands := []*Command{}
 	cleanupCommands := []*Command{DeleteRemainingNamespacesCommand()}
 
-	if !*doNotDeployIG {
+	deployIG := !CheckNamespace("gadget")
+	if deployIG {
 		imagePullPolicy := "Always"
 		if *k8sDistro == K8sDistroMinikubeGH {
 			imagePullPolicy = "Never"
@@ -148,7 +146,8 @@ func testMain(m *testing.M) int {
 		cleanupCommands = append(cleanupCommands, CleanupInspektorGadget)
 	}
 
-	if !*doNotDeploySPO {
+	deploySPO := !CheckNamespace("security-profiles-operator")
+	if deploySPO {
 		limitReplicas := false
 		patchWebhookConfig := false
 		bestEffortResourceMgmt := false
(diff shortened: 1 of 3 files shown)


mqasimsarfraz:
> If my understanding is correct, you check for specific namespace to detect if we need to deploy Inspektor Gadget and the SPO. Then the idea is good and the logic too, I would like to test it before giving my approval though.
> 
Yes, that is correct.

> Nonetheless, this only tackles the deployment but what about the removal?

If we don't deploy, we don't remove, as cleanups are only [performed]((https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/inspektor-gadget/integration_test.go#L149)) in case we deploy the components during given test run. So, it should handle removal as well.

integration/inspektor-gadget/integration_test.go
```diff
@@ -137,7 +134,8 @@ func testMain(m *testing.M) int {
 	initCommands := []*Command{}
 	cleanupCommands := []*Command{DeleteRemainingNamespacesCommand()}
 
-	if !*doNotDeployIG {
+	deployIG := !CheckNamespace("gadget")
+	if deployIG {
```
blanquicet, line 138: I see a problem with this approach. If developers forget to remove IG before running the test by mistake, they can't easily realize that tests run with an "old" (and probably wrong) IG version. That's why I prefer the approach of failing the tests if IG is already installed unless developers manually specify `-no-deploy-ig`, which means that they are aware that they are running the tests with an "old" (and this time, probably correct) IG version.

integration/inspektor-gadget/integration_test.go
```diff
@@ -148,7 +146,8 @@ func testMain(m *testing.M) int {
 		cleanupCommands = append(cleanupCommands, CleanupInspektorGadget)
 	}
 
-	if !*doNotDeploySPO {
+	deploySPO := !CheckNamespace("security-profiles-operator")
+	if deploySPO {
```
blanquicet, line 150: I frequently run tests that don't need the SPO. So, I run:

```bash
$ INTEGRATION_TESTS_PARAMS="-run TestExecsnoop -v -no-deploy-spo" make integration-tests
```

How can I do this now?

blanquicet (commented):
I'm sorry I couldn't review it before.

integration/inspektor-gadget/integration_test.go
mqasimsarfraz: That is a fair point. How about printing a warning in case we use pre-installed components? If that isn't enough we can revert the `-no-deploy-ig/-no-deploy-spo` flags and only do this detection for `spo` since we don't really care of spo version? Should take care of this [issue](https://github.com/inspektor-gadget/inspektor-gadget/pull/1110#discussion_r1023830835) as well?

mqasimsarfraz:
> I'm sorry I couldn't review it before.

No worries at all. I should have waited a bit more :laughing:

integration/inspektor-gadget/integration_test.go
mauriciovasquezbernal: oh, I use those flags a lot in my system to run integration tests locally. Can we add them back?

mqasimsarfraz: Sorry about it. I am just testing my changes will open the PR soon.

PR #1113, opened 2022-11-09
integration-tests: Patch seccomp profiles to force SPO cleanup

### LINKED ISSUE #1111: Implement forced cleanup for SPO
We have been hitting issue regarding dangling finalizers on SPO resources in CI. We should take a look into implementing forced cleanup for SPO.

### Implementation details
Go through the resources stuck on `Terminating` state and patch the blocking finalizers to allow clean up.

mqasimsarfraz:
We have seen multiple times that SPO is stuck on 'Terminating' state because of dangling finalizers this commit fixes the issue by forcefully removing finalizers allowing SPO cleanup, since we don't really want to test if the cleanup is working fine.

Fixes #1111 

### Testing
- Start an integration test run: 
```
$ INTEGRATION_TESTS_PARAMS="-run TestExecsnoop -v" make IMAGE_TAG=latest  integration-tests
```
- During test run delete all SPO pods to reproduce finalizers issue with:

```
$ kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io spo-mutating-webhook-configuration ; kubectl delete deployments.apps -n  security-profiles-operator  security-profiles-operator  security-profiles-operator-webhook; kubectl delete pods -l name=spod -n ssecurity-profiles-operator; kubectl delete ds -n security-profiles-operator spod
```
### Before the fix
The test run should timeout because of dangling finalizers

### After the fix
The test should complete successful with output as:
```
seccompprofile.security-profiles-operator.x-k8s.io/log-enricher-trace patched
seccompprofile.security-profiles-operator.x-k8s.io/nginx-1.19.1 patched
```

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -162,21 +162,47 @@ kubectl rollout status -n security-profiles-operator ds spod --timeout=180s || \
 }
 
 // CleanupInspektorGadget cleans up inspector gadget in Kubernetes
-var CleanupInspektorGadget *Command = &Command{
+var CleanupInspektorGadget = &Command{
 	Name:    "CleanupInspektorGadget",
 	Cmd:     "$KUBECTL_GADGET undeploy",
 	Cleanup: true,
 }
 
 // CleanupSPO cleans up security profile operator in Kubernetes
-var CleanupSPO *Command = &Command{
-	Name: "RemoveSecurityProfilesOperator",
-	Cmd: `
-	kubectl delete seccompprofile --all --all-namespaces
-	kubectl delete -f https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.3/deploy/operator.yaml --ignore-not-found
-	kubectl delete -f https://github.com/jetstack/cert-manager/releases/download/v1.8.0/cert-manager.yaml --ignore-not-found
-	`,
-	Cleanup: true,
+var CleanupSPO = []*Command{
+	{
+		Name: "RemoveSecurityProfilesOperator",
+		Cmd: `
+		kubectl delete seccompprofile --all --all-namespaces
+		kubectl delete -f https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.3/deploy/operator.yaml --ignore-not-found
+		kubectl delete -f https://github.com/jetstack/cert-manager/releases/download/v1.8.0/cert-manager.yaml --ignore-not-found
+		`,
+		Cleanup: true,
+	},
+	{
+		Name: "PatchSecurityProfilesOperatorProfiles",
+		Cmd: `
+		while true; do
+		  # Ensure we have profiles to clean, otherwise just exit.
+		  NAMESPACES=$(kubectl get seccompprofile --all-namespaces --no-headers --ignore-not-found -o custom-columns=":metadata.namespace" | uniq)
+		  if [ -z $NAMESPACES ]; then
+		    break
+		  fi
+ 
+		  # Patch profiles in each namespace, ignore any errors since it can already be deleted.
+		  for NAMESPACE in $NAMESPACES; do
+		    PROFILES=$(kubectl get seccompprofile --namespace $NAMESPACE -o name)
+		    for PROFILE in $PROFILES; do
+		      kubectl patch $PROFILE -n $NAMESPACE -p '{"metadata":{"finalizers":null}}' --type=merge || true
+		    done
+		  done
+
+		  # Give some time before starting next cycle.
+		  sleep 1
+		done
+		`,
+		Cleanup: true,
+	},
 }
 
 // RunCommands is used to run a list of commands with stopping/clean up logic.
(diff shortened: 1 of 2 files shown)


integration/inspektor-gadget/integration_test.go
```diff
@@ -160,7 +160,7 @@ func testMain(m *testing.M) int {
 			patchWebhookConfig = true
 		}
 		initCommands = append(initCommands, DeploySPO(limitReplicas, patchWebhookConfig, bestEffortResourceMgmt))
-		cleanupCommands = append(cleanupCommands, CleanupSPO)
+		cleanupCommands = append(cleanupCommands, CleanupSPO, PurgeSPO)
```
blanquicet, line 163: Mmmm, what about adding this forcing before trying to delete the SP in the `CleanupSPO` operation? Right before this:

https://github.com/inspektor-gadget/inspektor-gadget/blob/7ba006c31b5cf51ec0ded76570ab17bcaf143813/integration/command.go#L175 

Having this step within the `CleanupSPO` operation would also simplify also the effort to avoid deleting the SPO if we didn't deploy it (https://github.com/inspektor-gadget/inspektor-gadget/pull/1110).

BTW, I think you can patch all the SP (regardless of the namespace) as the `CleanupSPO` operation will try to remove all of them.

What do you think?

mqasimsarfraz: > Mmmm, what about adding this forcing before trying to delete the SP in the `CleanupSPO` operation? Right before this:
> 
> https://github.com/inspektor-gadget/inspektor-gadget/blob/7ba006c31b5cf51ec0ded76570ab17bcaf143813/integration/command.go#L175
> 
> Having this step within the `CleanupSPO` operation would also simplify also the effort to avoid deleting the SPO if we didn't deploy it (#1110).
> 
Agreed. I was also thinking the same, The main reason we can't run this before `kubectl delete seccompprofile --all --all-namespaces` is spo operator will add the finalizers back so we really need to run both the things in parallel. But I have an idea to combine it, let me update the PR and we see how it is looking?

> BTW, I think you can patch all the SP (regardless of the namespace) as the `CleanupSPO` operation will try to remove all of them.
> 
> What do you think?

The `kubectl patch` doesn't support `--all-namespaces` so I have go over all the namespaces one by one or am I missing something?

blanquicet: > The kubectl patch doesn't support --all-namespaces so I have go over all the namespaces one by one or am I missing something?
>
Oops, you are right. Maybe we could get all the ns where the sp are installed using `kubectl get seccompprofile -A`. IIRC, it also shows the ns.

mqasimsarfraz: Thanks for pointing out. Yes, makes sense to get the namespaces via `kubectl get seccompprofile -A`, directly iterating with `sh` is tricky so let me do it other way around.

integration/command.go
```diff
@@ -162,21 +162,36 @@ kubectl rollout status -n security-profiles-operator ds spod --timeout=180s || \
...
+		Name: "RemoveSecurityProfilesOperator",
+		Cmd: `
+		kubectl delete seccompprofile --all --all-namespaces
+		kubectl delete -f https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.3/deploy/operator.yaml --ignore-not-found
+		kubectl delete -f https://github.com/jetstack/cert-manager/releases/download/v1.8.0/cert-manager.yaml --ignore-not-found
+		`,
+		Cleanup: true,
+	},
+	{
+		Name: "PatchSecurityProfilesOperatorProfiles",
```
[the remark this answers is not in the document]
mqasimsarfraz: > I have only one concern, is it guaranteed that `PatchSecurityProfilesOperatorProfiles` will always be executed after `RemoveSecurityProfilesOperator`?

By after you mean the case in which we patch the profiles even before termination has started? How about we only do this patching once all the `SPOD` pods are gone so we are sure we really removing the dangling finalizers?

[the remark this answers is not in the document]
blanquicet: > How about we only do this patching once all the SPOD pods are gone so we are sure we really removing the dangling finalizers?
>
That's is what I mean 🙂. I would like to be sure that we patch the SPs after we run `RemoveSecurityProfilesOperator` (`SPOD` pods are gone). Otherwise, I remember you told me that the `SPOD` will add again the finalizers. I am concern about this because the slice doesn't guarantee the order.

However, maybe your second idea is more robust ... What if it runs in the background and keep patching the SPs until all the SPs are gone? In this way, it doesn't matter whether it is executed before or after the `SPOD` is gone.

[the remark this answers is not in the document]
mqasimsarfraz: > However, maybe your second idea is more robust ... What if it runs in the background and keep patching the SPs until all the SPs are gone? In this way, it doesn't matter whether it is executed before or after the `SPOD` is gone.

Sounds good. I agree that is our best option. I will go ahead and update the PR.

blanquicet (approved):
I didn't test it, but LGTM from code inspection. IMO, we can go ahead and try it in the CI.

Issue #1112, opened 2022-11-09
[RFE] Use K8s API directly instead of exec kubectl ...


burak-ok:
## Current situation

At least in `integration/command.go` we are calling `kubectl` for new pods, waiting on something, gathering information, etc...
We should use the k8s API directly

See https://github.com/inspektor-gadget/inspektor-gadget/pull/1091#discussion_r1014264671

mauriciovasquezbernal:
We realized it doesn't make sense to implement it. It's too much work for little benefit. Closing.

PR #1116, opened 2022-11-11
Use the K8s API directly intead of exec kubectl

burak-ok:
# Use the K8s API directly intead of exec kubectl

Resolves issue https://github.com/inspektor-gadget/inspektor-gadget/issues/1112

This PR introduces a new interface `Command`. The old `Command` struct was renamed to `CmdCommand` to reflect its new usecase better.
A new struct was added too: `K8sCommand`, which represents commands that use the K8s API of go directly.

TODO:
- [ ] Check which Interface methods are really needed
- [ ] Move these structs + methods into their own file
- [ ] Implement more Commands as `K8sCommand`
- [ ] Think about their lifecycle. `CmdCommand` also has `Kill()`, etc... Needed for `K8sCommand` or not?
- [ ] General cleanup of my development testing


burak-ok:
Hi @mauriciovasquezbernal 

These are some preliminary to usethe K8s golang library instead of `exec kubectl ...`. See https://github.com/inspektor-gadget/inspektor-gadget/pull/1116/commits/5d54b6ba43fe78820b009a64b3278827bcddd1fd for a small example

What do you think about it? Seems to be much more code. Should i instead expand and work on http://inspektor-gadget/pkg/k8sutil/ having more utils and features?

integration/command.go
```diff
@@ -517,6 +564,56 @@ EOF
...
+// PodCommandThroughAPI returns a Command that starts a pid with a specified image, command and args
+func PodCommandThroughAPI(podname, image, namespace string, command, commandArgs []string) *K8sCommand {
+	return &K8sCommand{
+		name: fmt.Sprintf("Creating pod %s", podname),
+		runFunc: func(t *testing.T) {
+			k8sClient, err := k8sutil.NewClientsetFromConfigFlags(utils.KubernetesConfigFlags)
+			if err != nil {
+				t.Fatalf("Creating pod (%s): %s", podname, commonutils.WrapInErrSetupK8sClient(err).Error())
+				return
+			}
```
mauriciovasquezbernal, line 577: It would be nice to create this once and share in all invocations to these functions.

mauriciovasquezbernal (commented):
I think it goes in the right direction. It seems to be a lot of upfront work but IMO in the long term that will pay off as the code using the API directly will be easier to maintain / extend than the logic around kubectl. 

Some thoughts:
- Do we need WaitUntilTestPodReadyCommand()? Perhaps it can be part of `PodCommand()`. 
- We'd need to think better about the lifecycle as you mentioned. It seems to me much of the logic we built is for terminal commands (cmdcommand), but it's not required for k8s commands.

burak-ok:
I am still unsure about the implicit waiting of a command.
I searched for all `PodCommand` and we always do a `Wait()` on it right after creating it. So it makes sense to have that option.
While i would implement that should i remove the `Wait()` commands outright. At least we do not need them for the tests.

Regarding passing the `k8sClient` instead of creating it new in every command: Should we pass it or create a `K8sTestClient struct` which has the `func (c *K8sTestClient) PodCommand...` attached to it? WDYT?

Back to waiting implicitly: We need a set of command which does not wait for the commands completion. For example the `DeleteRemainingNamespacesCommand` should not wait, since we try to clean up as much resources as possible.

mauriciovasquezbernal:
> I searched for all PodCommand and we always do a Wait() on it right after creating it. So it makes sense to have that option.
While i would implement that should i remove the Wait() commands outright. At least we do not need them for the tests.

I agree to have one version that implicitly waits for now. Then we could have an "async" version if needed. 

> Regarding passing the k8sClient instead of creating it new in every command: Should we pass it or create a K8sTestClient struct which has the func (c *K8sTestClient) PodCommand... attached to it? WDYT?

I'm not sure, I think passing it will suffice. 

> But doing this right now in this PR is too much?

Yes, that seems too much. I think this PR is complicated enough already.

burak-ok:
Other projects use `kubectl` instead of the API:
- [`kubernetes/test-infra`](https://github.com/kubernetes/test-infra/blob/ee1e7c845bb25177944bb5de37545f007f529059/kubetest/kubernetes.go#L33)
- [`kubernetes-sigs/security-profiles-operator` 1](https://github.com/kubernetes-sigs/security-profiles-operator/blob/d416113646d55fce7286122d037da96dcbdc535d/test/suite_test.go#L583)
- [`kubernetes-sigs/security-profiles-operator` 2](https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/test/e2e_test.go#L211)
- [`cilium/cilium`](https://github.com/cilium/cilium/blob/fcf2b7ad1c47cc7309e752eb087ecfbda7ff8a6e/test/helpers/kubectl.go#L341)

mauriciovasquezbernal:
> Other projects use kubectl instead of the API:

Do you suggest to keep using kubectl? 

It's fine for me. I thought using the API will bring some advantages, but it seems too much work :(

Issue #924, opened 2022-09-07
top/ebpf: Map size calculation is not accurate 


mauriciovasquezbernal:
Current implementation computes the size of the map manually by using the key size, value size and number of entries. 

https://github.com/kinvolk/inspektor-gadget/blob/ad87c3f5407ae5bd30d9fc486e79df4662ec19ae/pkg/gadgets/top/ebpf/tracer/tracer.go#L208

This calculation is not totally correct as the allocated map memory on the kernel is different. For instance, key sizes are rounded up to 8 bytes, so this calculation gives a misleading result when the key size is not a multiple of 8.

bpftool gets this information using procfs -> https://github.com/libbpf/bpftool/blob/3e8e34bcc34287dc2eba6abec8313b8cc980dfa6/src/map.c#L466, think we can follow the same approach here.

/cc @alban @flyth

alban:
That's great, I didn't know that fdinfo was giving this. It was introduced in Linux 4.10 via https://github.com/torvalds/linux/commit/21116b7068b9b66ac16b2fe3675469f459968c3f. The top-ebpf gadget requires Linux 5.1 so that's fine.

The documentation will need updating:
https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/top/ebpf.md#a-note-about-memory-usage-of-maps

Using the values from fdinfo takes into account percpu maps, so that's an improvement but BPF_MAP_TYPE_PERF_EVENT_ARRAY maps will unfortunately still not account for the buffers in the map.

```
127: perf_event_array  name events  flags 0x0
	key 4B  value 4B  max_entries 4  memlock 4096B
	pids execsnoop(501155)
```
```
$ sudo cat /proc/501155/fdinfo/4
...
memlock:	4096
map_id:	127
```

PR #1120, opened 2022-11-15
gadgets/top/ebpf: Read fdinfo of map to get its size

burak-ok:
# gadgets/top/ebpf: Read fdinfo of map to get its size

Instead of calculating the size ourself we read it from `fdinfo`. See https://github.com/inspektor-gadget/inspektor-gadget/issues/924

This does not solve the _issue_ regarding to perf maps or maps of maps. Their value is still a descriptor and therefore 4 bytes.
Is there any possiblity to query the perf buffer sizes too? Do we want to include them in the size too?

### DIFF
--- a/pkg/gadgets/top/ebpf/tracer/tracer.go
+++ b/pkg/gadgets/top/ebpf/tracer/tracer.go
@@ -185,6 +185,29 @@ func getPidMapFromProcFs() (map[uint32][]*types.PidInfo, error) {
 	return pidmap, nil
 }
 
+func getMemoryUsage(m *ebpf.Map) (uint64, error) {
+	f, err := os.Open(fmt.Sprintf("/proc/self/fdinfo/%d", m.FD()))
+	if err != nil {
+		return 0, fmt.Errorf("could not read fdinfo: %w", err)
+	}
+	defer f.Close()
+
+	sc := bufio.NewScanner(f)
+	for sc.Scan() {
+		if strings.HasPrefix(sc.Text(), "memlock:\t") {
+			lineSplit := strings.Split(sc.Text(), "\t")
+			if len(lineSplit) == 2 {
+				size, err := strconv.ParseUint(lineSplit[1], 10, 64)
+				if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -205,9 +205,27 @@ func (t *Tracer) nextStats() ([]types.Stats, error) {
...
+				if len(lineSplit) == 2 {
+					mapSizes[curMapID], err = strconv.ParseUint(lineSplit[1], 10, 64)
+					if err != nil {
+						return nil, fmt.Errorf("could not read memlock of map ID (%d): %w", curMapID, err)
+					}
+					break
+				}
+			}
+		}
 	}
```
mauriciovasquezbernal, line 229: How about putting all of this in a helper function? `func getMemoryUsage(m *ebpf.Map)`?

mauriciovasquezbernal (commented):
Looks very good, thanks for working on this. Just a comment to better organize the code.

pkg/gadgets/top/ebpf/tracer/tracer.go
burak-ok: Done.
I should do that way more often

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -205,9 +205,27 @@ func (t *Tracer) nextStats() ([]types.Stats, error) {
 		if err != nil {
 			continue
 		}
-		mapSizes[curMapID] = (uint64(mapData.KeySize()) + uint64(mapData.ValueSize())) * uint64(mapData.MaxEntries())
+		defer mapData.Close()
```
burak-ok, line 208: Review of myself:
`defer mapData.Close()` doesn't do what I think here. It is not bound to block scope, but to function scope.
Not what i want here. Changing...

mauriciovasquezbernal (approved):

PR #1123, opened 2022-11-16
gadgets/snapshot: Use columns library

burak-ok:
# gadgets/snapshot: Use columns library

With this PR both `snapshot` gadgets are using the `columns` library.
Specifically it is using the `TransformIntoTable` function which finds the perfect width, since we already have all events. (Like not using extra width on a wide column which currently only has very small entries)
Example:
```
NODE     NAMESPACE POD                                  CONTAINER           COMM PID
minikube default   kubernetes-bootcamp-75c5d958ff-rntpk kubernetes-bootcamp sh   648715
minikube default   kubernetes-bootcamp-75c5d958ff-rntpk kubernetes-bootcamp node 648781
minikube default   kubernetes-bootcamp-75c5d958ff-rntpk kubernetes-bootcamp bash 650668
```

Additionally the sorting was also modified to use the `columssort` instead of doing it ourselves in the different gadgets

### DIFF
--- a/cmd/common/snapshot/process.go
+++ b/cmd/common/snapshot/process.go
@@ -15,13 +15,12 @@
 package snapshot
 
 import (
-	"fmt"
 	"sort"
-	"strings"
 
 	"github.com/spf13/cobra"
 
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/types"
 )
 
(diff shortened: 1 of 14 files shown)


cmd/common/snapshot/process.go
```diff
@@ -64,36 +63,19 @@ func (p *ProcessParser) GetOutputConfig() *commonutils.OutputConfig {
 	}
 }
 
-func (p *ProcessParser) SortEvents(allProcesses *[]types.Event) {
+func (p *ProcessParser) SortEvents(allProcesses []*types.Event) {
```
mauriciovasquezbernal: This change is reverted in the next commit. Can you make the right change in this commit to avoid changing it back in the next one?

cmd/common/snapshot/snapshot.go
```diff
@@ -42,15 +40,10 @@ type SnapshotEvent interface {
 // implement.
 type SnapshotParser[Event any] interface {
 	// SortEvents sorts a slice of events based on a predefined prioritization.
-	SortEvents(*[]Event)
+	SortEvents(*[]*Event)
```
mauriciovasquezbernal, line 43: I think using `SortEvents([]*Event)` will make our lives easier.

mauriciovasquezbernal (commented):
Looks great. Please look at my comment about the signature of SortEvents, I think that could help with the clarify of the code around.

cmd/common/snapshot/snapshot.go
mauriciovasquezbernal: I see. I'm fine for me to fix that in a following PR.

cmd/common/snapshot/snapshot.go
```diff
@@ -78,26 +71,19 @@ func (g *SnapshotGadgetPrinter[Event]) PrintEvents(allEvents []Event) error {
...
+		allEventsTrimmed := []*Event{}
 		for _, e := range allEvents {
-			baseEvent := e.GetBaseEvent()
+			baseEvent := (*e).GetBaseEvent()
 			if baseEvent.Type != eventtypes.NORMAL {
 				commonutils.ManageSpecialEvent(baseEvent, outputConfig.Verbose)
 				continue
+			} else {
+				allEventsTrimmed = append(allEventsTrimmed, e)
 			}
```
mauriciovasquezbernal, line 82: nit: not else needed since there is a continue in previous conditional.

mauriciovasquezbernal (approved):

PR #1127, opened 2022-11-18
pkg/gadget/trace/capabilities: Sync default audit-only value with CLI

blanquicet:
# cabapilities: Sync default `audit-only` value with CLI

For people using the CLI, the default is true. Let's set the same default for people using the trace CR directly.


mauriciovasquezbernal (approved):
LGTM!

PR #1128, opened 2022-11-18
profile/cpu: Use columns library

burak-ok:
# profile/cpu: Use columns library

Used the `columns` library wherever possible. Can't generalize for the `profile` gadgets, since `profile block-io` is outputting a graph. Also `profile cpu` also has its special case where the stack is printed on its separate line


mauriciovasquezbernal (approved):
Even if there is a small hack to support the stack traces, I think this PR brings good improvements to the code. LGTM!

blanquicet (approved):
LGTM!

PR #1129, opened 2022-11-18
trace/capabilities: Fix unit tests

mauriciovasquezbernal:
The TestCapabilitiesTracer/audit_only_false test was failing in some kernel versions. I found out that "/bin/cat /dev/null" doesn't always generate a non audit capability check. For this reason, I switched to "/bin/cat /proc/kallsyms" that seems to generate that in more cases.


blanquicet (approved):
LGTM

PR #1131, opened 2022-11-18
cmd/common: Clean up special event management

blanquicet:
# cmd/common: Clean up special event management

- There is no need to make a copy of the Event to print special events, instead, use pointers.
- Move `GetBaseEvent` into the embedded struct so that it doesn't need to be implemented by every gadget.
- Rename ManageSpecialEvent to HandleSpecialEvent as in this case, "Handle" is more accurate than "Manage".

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -99,6 +99,12 @@ type Event struct {
 	Message string `json:"message,omitempty"`
 }
 
+// GetBaseEvent is needed to implement commonutils.BaseElement and
+// snapshot.SnapshotEvent interfaces.
+func (e Event) GetBaseEvent() *Event {
+	return &e
+}
+
 func Err(msg string) Event {
 	return Event{
 		CommonData: CommonData{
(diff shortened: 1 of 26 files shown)


pkg/types/types.go
```diff
@@ -89,6 +89,12 @@ type Event struct {
 	Message string `json:"message,omitempty"`
 }
 
+// GetBaseEvent is needed to implement commonutils.BaseElement and
+// snapshot.SnapshotEvent interfaces.
+func (e Event) GetBaseEvent() *Event {
+	return &e
+}
+
```
flyth, line 97:
```suggestion
// GetBaseEvent is needed to implement commonutils.BaseElement and
// snapshot.SnapshotEvent interfaces.
func (*e Event) GetBaseEvent() *Event {
	return e
}
```

I think we should use a pointer receiver here as well, otherwise the struct is still being copied.

flyth (commented):
Much cleaner this way, thanks!

pkg/types/types.go
blanquicet: Yeah, that would also help. However, doing so with generics became quite complex. I will let it for another PR.

blanquicet:
Rebased to ensure the CI is all green after merging https://github.com/inspektor-gadget/inspektor-gadget/pull/1141.

mauriciovasquezbernal (approved):
Nice cleanup. Regarding pointers, I think we should check if this is possible to pass pointers in more places, it seems to me we're making a lot of unneeded copies, but of course this is something to explore on a different PR. 

(I'm approving now but please check that there are some instances of ManageSpecialEvent that need to be updated in cmd/kubectl-gadget/traceloop.go.

PR #1135, opened 2022-11-21
Remove extra output mode validation

mauriciovasquezbernal:
The output mode is validated in ParseOutputConfig() when building the common flags. Let's remove some extra validations around as they're not needed.

### DIFF
--- a/cmd/kubectl-gadget/trace/trace.go
+++ b/cmd/kubectl-gadget/trace/trace.go
@@ -48,13 +48,7 @@ func (g *TraceGadget[Event]) Run() error {
 		Parameters:       g.params,
 	}
 
-	// Print header
-	switch g.commonFlags.OutputMode {
-	case commonutils.OutputModeJSON:
-		// Nothing to print
-	case commonutils.OutputModeColumns:
-		fallthrough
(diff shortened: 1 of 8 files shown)


cmd/kubectl-gadget/trace/trace.go
```diff
@@ -85,10 +79,9 @@ func (g *TraceGadget[Event]) Run() error {
 			fallthrough
 		case commonutils.OutputModeCustomColumns:
 			return g.parser.TransformIntoColumns(&e)
-		default:
-			fmt.Fprint(os.Stderr, commonutils.WrapInErrOutputModeNotSupported(g.commonFlags.OutputMode))
-			return ""
 		}
```
blanquicet, line 82: I think we can change this `switch` for an `if` statement:

```golang
	if commonFlags.OutputMode == commonutils.OutputModeJSON {
		b, err := json.Marshal(event)
		if err != nil {
			fmt.Fprintf(os.Stderr, "Error: %s", fmt.Sprint(commonutils.WrapInErrMarshalOutput(err)))
			return
		}

		fmt.Println(string(b))
	} else {
		fmt.Println(parser.TransformIntoColumns(&event))
	}
```

WDYT?

PR #1137, opened 2022-11-22
[RFC] New example: shellloop

alban:
# New example: shellloop

With shellloop, you can monitor a shell with traceloop without using any container. shellloop starts the shell with traceloop enabled.

## How to use

```
$ cd examples/shellloop/
$ go build shellloop.go
$ sudo ./shellloop 
sh-5.1# echo '3*4*7'|bc
84
sh-5.1# 
exit
CPU PID        COMM     SYSCALL              PARAMS                                                RET
...
2   1333655    bc       read                 fd=0, buf="3*4*7\n", count=8192                       6  
2   1333655    bc       write                fd=1, buf="84\n", count=3                             3  
...
$
```

## Testing done

None

### DIFF
--- a/examples/shellloop/shellloop.go
+++ b/examples/shellloop/shellloop.go
@@ -1,0 +1,107 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


alban:
@eiffel-fl What do you think if we make this a feature of local-gadget? It could be called with:
```
$ sudo local-gadget traceloop --shell
```

alban:
Closing because I have not worked on this for a while.
If someone wants to complete this, I think this could be implemented outside of Inspektor Gadget by reusing IG's Golang packages.

PR #1140, opened 2022-11-23
Remove `Cgo` for `traceloop`

eiffel-fl:
Hi.

In this PR, I removed the need of `Cgo` for `traceloop`.

Best regards.


pkg/gadgets/traceloop/tracer/bpf/traceloop.bpf.c
```diff
@@ -48,6 +48,10 @@
 #define bpf_error_printk(fmt, ...)
 #endif /* !defined(SHOW_ERROR) */
 
+/* These are needed to not make use of cgo. */
```
mauriciovasquezbernal, line 51: I think this comment is misleading. Even if this is true the final purpose of them is not to use cgo, the immediate reason whey they are here is to avoid the compiler optimizing out  those definitions from the elf file.

mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #1141, opened 2022-11-23
integration/TestEbpftop: Check for any captured event

burak-ok:
# integration/TestEbpftop: Check for any captured event

Currently the `TestEbpftop` integration test is flaky.
It checked for a eBPF program name of `ig_top_ebpf_it`. But it could happen that it would not be in the `top ebpf` statistics, since other tests are running in parallel and in the `top` gadgets there is a maximum number of rows (default `20`).

Until we have a way to filter `top ebpf` by progam name or other things, this quickfix should suffice.

### Tested
The test will fail if `top ebpf` does not produce any output (empty)


integration/inspektor-gadget/integration_test.go
```diff
@@ -595,10 +595,7 @@ func TestEbpftop(t *testing.T) {
 		Cmd:          "$KUBECTL_GADGET top ebpf -o json",
 		StartAndStop: true,
 		ExpectedOutputFn: func(output string) error {
-			expectedEntry := &ebpftopTypes.Stats{
-				Name: "ig_top_ebpf_it",
-				Type: "Tracing",
-			}
+			expectedEntry := &ebpftopTypes.Stats{}
```
alban, line 598: You could add a comment:
```
// the gadget truncates the output to 20 entries,
// so we might not find ig_top_ebpf_it if it was truncated away
```

alban (approved):
LGTM with one suggestion

PR #1143, opened 2022-11-23
ci: Do not cancel in progress jobs on main.

eiffel-fl:
Hi.

In this PR, I made `cancel-in-progress` to be set only when branch is different than main.
Indeed, we want to catch regression as soon as they are introduced, so we do not want to cancel in progress CI jobs on main branch.

Best regards.


burak-ok (approved):
LGTM, hopefully we will spend less time on bughunting this way

blanquicet (approved):
LGTM!

mqasimsarfraz (approved):
Tested on my fork, LGTM. Seem like the PR needs to be rebased!

PR #1148, opened 2022-11-24
go.mod: update Kubernetes packages

alban:
# go.mod: update Kubernetes packages

In preparation for a future gadget, I need up-to-date Kubernetes packages.

New versions of Kubernetes packages marshall network policies in a different way. Update the unit tests accordingly.

## How to use

No changes.

## Testing done

`make test`

PR #1150, opened 2022-11-24
lint: update github action to fix deprecation warning

alban:
Patch initially proposed in https://github.com/inspektor-gadget/inspektor-gadget/pull/1148


alban:
This PR is similar to what dependabot did in commit b088df47cdf07e46f6fde4902b410c9c5f25ff67.

burak-ok (approved):
LGTM.

I played a little bit around and we can remove the `install-debian-packages` step from `Lint` since we do not need it there.
I'm doing that in another PR

PR #1152, opened 2022-11-25
integraton-tests: Fail if no entries provided to expectAllToMatch

mqasimsarfraz:
It seems like if no entries are passed to `expectAllToMatch` the test still goes and passes. We should explicitly fail if no entries are given.


mqasimsarfraz:
It seems `oomkiller` isn't working as expected for `containerd`. I see process getting killed for `docker`/`cri-o` but for containerd it is never killed. Need to be investigated.

burak-ok (approved):
LGTM
Very important fix

mauriciovasquezbernal (approved):
nice!

PR #1153, opened 2022-11-25
Remove Lint errors

burak-ok:
# Remove Lint errors

This PR removes the Lint Errors that we see on the summary page of Workflows and the Lint page itself.
Example see https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3540176371
Even though everything was ok, we see errors.

With this change we do not use the cache of the `setup go` step on the Lint step anymore

Related https://github.com/golangci/golangci-lint-action/issues/23


burak-ok:
> I am not sure to understand why it works. Is `actions/checkout` running `go get`?

I changed the order back. The only problem was the caching since it also tries to cache our code

PR #1156, opened 2022-11-28
local-gadget: Add network-graph trace gadget

blanquicet:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/889


pkg/container-collection/container-collection.go
```diff
@@ -311,6 +325,31 @@ func (cc *ContainerCollection) Enrich(event *eventtypes.CommonData, mountnsid ui
...
+	if containers[0].Podname != "" && containers[0].Namespace != "" {
+		// Kubernetes containers within the same pod.
+		// TBD: Should we verify that all containers belong to the same pod?
+		event.Pod = containers[0].Podname
+		event.Namespace = containers[0].Namespace
+	}
+	// else {
+	// 	TODO: Non-Kubernetes containers sharing the same network namespace.
+	// 	What should we do here?
+	// }
```
blanquicet, line 350: Suggestion?

pkg/container-collection/container-collection.go
```diff
@@ -311,6 +325,31 @@ func (cc *ContainerCollection) Enrich(event *eventtypes.CommonData, mountnsid ui
...
+	}
+	if len(containers) == 1 {
+		event.Container = containers[0].Name
+		event.Pod = containers[0].Podname
+		event.Namespace = containers[0].Namespace
+		return
+	}
+	if containers[0].Podname != "" && containers[0].Namespace != "" {
+		// Kubernetes containers within the same pod.
+		// TBD: Should we verify that all containers belong to the same pod?
```
blanquicet, line 343: IMO, it is not necessary. WDYT?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -373,22 +374,26 @@ func (t *Tracer) listen(
...
+	delete(t.attachments, netns)
 }
 
-func (t *Tracer) Detach(key string) error {
-	if l, ok := t.attachments[key]; ok {
+func (t *Tracer) Detach(pid uint32) error {
+	netns, err := containerutils.GetNetNs(int(pid))
+	if err != nil {
+		return fmt.Errorf("getting network namespace of pid %d: %w", pid, err)
+	}
```
blanquicet, line 388: Wasn't `containercollection.EventTypeRemoveContainer` occurring before the container was gone? The tests are failing because of:

```bash
    local-gadget-manager_test.go:502: Received: {{{local default test-local-gadget-dns001 } err stop tracing container "test-local-gadget-dns001": getting network namespace of pid 9356: stat /proc/9356/ns/net: no such file or directory}      }, Expected: {{{local default test-local-gadget-dns001 } debug tracer detached}      }
```

So, it seems like the container was gone when `Detach` was called. If it is confirmed, I was considering returning an `attachKey` to be used for detaching:

```golang
	attachKey, err := tracer.Attach(pid, eventCallback); 
	if err != nil {
		fmt.Printf("error attaching tracer: %s\n", err)
		return
	}
	defer tracer.Detach(attachKey)
```

WDYT?

pkg/container-collection/networktracer/connect.go
```diff
@@ -16,16 +16,28 @@ package networktracer
...
+type CallbackApproach[Event any] interface {
+	Attach(pid uint32, eventCallback func(Event)) error
+}
+
+// PopApproach represents network tracers that don't return the events through a
+// callback but instead they provide a method to retrieve them.
+type PopApproach[Event any] interface {
+	Attach(pid uint32) error
+	Pop() ([]*Event, error)
 }
```
blanquicet, line 41: With this implementation, the compiler doesn't fail if the tracer doesn't implement any of the `Attach` approaches, but it will fail at runtime. WDYT?

mauriciovasquezbernal:
Looks great. We need to handle the container termination issue (avoid using it to get the namespace), and it should be all fine.

pkg/container-collection/container-collection.go
burak-ok: As we want local-gadget to be used for all containers outside of k8s too, i think we should not verify as you suggested

pkg/container-collection/container-collection.go
burak-ok: Maybe leave TODO until we support all containers?

pkg/container-collection/networktracer/connect.go
burak-ok: Even though i dislike the idea of ditching compile time safety, i currently also don't know another way to support the 2 differenct `Tracer`s through one `interface` easily.

mauriciovasquezbernal (commented):
Another instance of me putting the comment on the wrong box and forgetting to summit the proper review.

pkg/container-collection/container-collection.go
blanquicet: We already support all containers (Kubernetes and non-Kubernetes). The use case here is to have more than one non-Kubernetes container sharing the same network namespace. In that case, we should not enrich the event with the container name of none because we are not sure which of them actually generated the event. And being non-Kubernetes containers, there will not be Kubernetes metadata (namespace / pod). Therefore, the event will have no reference at all.

I don't know what we can do here, we could show all the possible containers where the event could have been generated? But how to present it? Using the group feature of the columns library? ... I think we can manage this in another PR letter as it is a very specific use case.

pkg/gadgets/trace/dns/tracer/tracer.go
blanquicet: I tried it, but the result didn't convince me. It looks strange from the caller's point of view because the "attachment" object returned by the `tracer.Attach()` method seems to be something that should be managed by the tracer code internally. The caller should only call tracer methods: `tracer.Attach` and `tracer.Detach`. Instead of calling `tracer.Attach` and `attachment.Close()`.

 I pushed a proposal to do so. Let me know what you think about it.

pkg/gadgets/helpers.go
```diff
@@ -16,10 +16,20 @@ package gadgets
```
mauriciovasquezbernal, line 16: IMO all additions to this file should be moved to a new `networkgadgets.go` file as they don't have a direct relation with the helpers implemented here.

pkg/gadgets/helpers.go
```diff
@@ -32,6 +42,8 @@ const (
 	PerfBufferPages = 64
 )
 
+var ErrEmptyEvent = errors.New("empty event")
```
mauriciovasquezbernal, line 45: I'm wondering if we can replace it by returning a `nil` event?

pkg/gadgets/helpers.go
```diff
@@ -91,3 +103,190 @@ func IPStringFromBytes(ipBytes [16]byte, ipType int) string {
...
+	// 2. pods with networkHost=true
+	// In both cases, we want to attach the BPF program only once.
+	users map[uint32]struct{}
+}
+
+func NewNetworkTracerAttachment(
+	pid uint32,
+	spec *ebpf.CollectionSpec,
+	bpfProgName string,
+	bpfMapName string,
```
mauriciovasquezbernal, line 133: bpfMapName -> bpfPerfMapName? (to be a bit more explicit?)

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: I think something like:

```golang 
  attachment := tracer.Attach(...)

...
  attachment.Close()
```

is a totally valid approach. 

It's very similar to the one used by the link package of cilium/ebpf https://pkg.go.dev/github.com/cilium/ebpf/link#Tracepoint

Anyway, I think the  approach that you proposed is fine too, so I'm fine with either solution.

mauriciovasquezbernal (commented):
I like where this is going, however I'm having some issues reviewing this big PR, specially because it touches many different files. What do you think about splitting it?

pkg/gadgets/trace/dns/tracer/tracer.go
blanquicet: Yep, the usage also seems ok for me. But when I implemented it, I found it strange to have a map of attachments per netns within the `Tracer`, and another map in the `ConnectionToContainerCollection` with the same attachments but per pid. Don't you think? I don't know, I will have a look at the example you pointed out just for curiosity, but I will go ahead with the current approach 🙂.

pkg/gadgets/helpers.go
```diff
@@ -91,3 +103,190 @@ func IPStringFromBytes(ipBytes [16]byte, ipType int) string {
 		return ""
 	}
 }
+
+type NetworkTracerAttachment struct {
```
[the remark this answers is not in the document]
blanquicet: Yep. Also the `newNetworkTracerAttachment`.

pkg/gadgets/helpers.go
blanquicet: Yep, I think we should move to a separate file. What about creating a package in pkg/gadgets/internal/networktracer?

pkg/gadgets/helpers.go
blanquicet: I also considered that, but I thought that one normally expects that if `err` is not `nil`, the event is valid, and a `nil` check is not needed:

```golang
	event, err := parseEvent(record.RawSample)
	if err != nil {
		eventCallback(baseEvent(types.Err(err.Error())))
		continue
	}
	if event == nil {
		continue
	}
	eventCallback(*event)
```

WDYT?

pkg/gadgets/helpers.go
blanquicet: Doing so, we can do the following within the DNS and SNI tracers:

```golang
type Tracer struct {
	*networktracer.Tracer[types.Event]
}
```

blanquicet:
> I like where this is going, however I'm having some issues reviewing this big PR, specially because it touches many different files. What do you think about splitting it?

Agree. I divided it into three PRs:

- https://github.com/inspektor-gadget/inspektor-gadget/pull/1180
- https://github.com/inspektor-gadget/inspektor-gadget/pull/1182
- This one.

pkg/gadgets/helpers.go
mauriciovasquezbernal: I think returning nil is better. btw, in this case that check would be removed when we switch to use pointers everywhere for passing events.

blanquicet: Applied in https://github.com/inspektor-gadget/inspektor-gadget/pull/1180

blanquicet:
> > I like where this is going, however I'm having some issues reviewing this big PR, specially because it touches many different files. What do you think about splitting it?
> 
> Agree. I divided it into three PRs:
> 
>     * [x]  [dns/sni: Reduce duplicated code in network tracers #1180](https://github.com/inspektor-gadget/inspektor-gadget/pull/1180)
> 
>     * [x]  [pkg/gadget/tracer/network: Enrich events using `EnrichByNetns` #1182](https://github.com/inspektor-gadget/inspektor-gadget/pull/1182)
> 
>     * This one.

This PR is now ready to be reviewed.

cmd/local-gadget/trace/network.go
```diff
@@ -0,0 +1,166 @@
...
+		parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, cols)
+		if err != nil {
+			return commonutils.WrapInErrParserCreate(err)
+		}
+
+		// This callback is used by the ConnectionToContainerCollection to
+		// notify when containers are attached and detached, or any error during
+		// that operations. The actual events generated by the tracer will be
+		// retrieved using the Pop() method.
+		eventCallback := func(container *containercollection.Container, event networkTypes.Event) {
```
mauriciovasquezbernal, line 66: Please update now that https://github.com/inspektor-gadget/inspektor-gadget/pull/1248 is merged.

integration/inspektor-gadget/integration_test.go
```diff
@@ -1068,41 +1069,63 @@ func TestNetworkGraph(t *testing.T) {
 	}
```
mauriciovasquezbernal, line 1069: I'd suggest to rebase on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/1251 given that it's going to be merged. (Sorry for the conflicts, but it happens when we're working on many things in parallel).

integration/inspektor-gadget/integration_test.go
```diff
@@ -1068,41 +1069,63 @@ func TestNetworkGraph(t *testing.T) {
...
+					Proto:           "tcp",
+					PodIP:           nginxIP,
+					PodLabels:       map[string]string{"run": "nginx-pod"},
+					Port:            80,
+					RemoteKind:      networkTypes.RemoteKindPod,
+					RemoteAddr:      testPodIP,
+					RemoteNamespace: ns,
+					RemoteName:      "test-pod",
+					RemoteLabels:    map[string]string{"run": "test-pod"},
+				},
```
mauriciovasquezbernal, line 1114: Nice that it checks both sides now!

mauriciovasquezbernal (approved):
LGTM. I tested and works as expected. 

Please handle the two comments before merging.

cmd/local-gadget/trace/network.go
blanquicet: Done.

integration/inspektor-gadget/integration_test.go
blanquicet: Done.

PR #1158, opened 2022-11-29
profile/cpu: Add suffix "stack" to kernel/user parameter name

burak-ok:
# profile/cpu: Add suffix "stack" to kernel/user parameter name

A kubernetes parameter already uses the `--user` parameter.
The following screenshot (yellow highlight by me) shows that `-U`/`--user` is missing for `profile cpu` which would only show the userspace stack.
image

# Changed behavior on latests version of `cobra`
While getting a minimal testcode i discovered that the behavior changed. On our version the global flag (the kubernetes one) has precedence.
With the newest version `1.6.1` the local parameter (the userspace stack one) will be shown and shadows the global flag

Should i open an Issue/PR with a potential fix on the cobra github?


mauriciovasquezbernal (approved):
Good catch!

Issue #1001, opened 2022-09-30
Some integration tests report events for other namespace


eiffel-fl:
## Description

Some integration tests report events for other namespace than the one dedicated to this test.

## Impact

As a consequence, the test fails because it except only one namespace to be printed.

## Environment and steps to reproduce

In ARO, the integration tests for `capabilities` and `biotop` fail because some events are printed for `test-tcptop-*` namespace:

* [`capabilities`](https://github.com/kinvolk/inspektor-gadget/actions/runs/3157606929/jobs/5140732930#step:6:22645):

```
2022-09-30T12:03:54.2901764Z     command.go:429: Stop command(StartCapabilitiesGadget)
2022-09-30T12:03:54.2902359Z     command.go:431: Command returned(StartCapabilitiesGadget):
2022-09-30T12:03:54.2902744Z         
2022-09-30T12:03:54.2904562Z         {"node":"arotestingcluster-kmcjg-worker-eastus2-nnb52","namespace":"test-capabilities-2800655059608903620","pod":"test-pod","container":"test-pod","type":"normal","pid":2444819,"comm":"runc:[2:INIT]","cap":21,"capName":"CAP_SYS_ADMIN","verdict":"Allow","mountnsid":4026533142}
2022-09-30T12:03:54.2906846Z         {"node":"arotestingcluster-kmcjg-worker-eastus2-nnb52","namespace":"test-capabilities-2800655059608903620","pod":"test-pod","container":"test-pod","type":"normal","pid":2444819,"comm":"runc:[2:INIT]","cap":21,"capName":"CAP_SYS_ADMIN","verdict":"Allow","mountnsid":4026533142}
2022-09-30T12:03:54.2909061Z         {"node":"arotestingcluster-kmcjg-worker-eastus2-nnb52","namespace":"test-capabilities-2800655059608903620","pod":"test-pod","container":"test-pod","type":"normal","pid":2444819,"comm":"runc:[2:INIT]","cap":21,"capName":"CAP_SYS_ADMIN","verdict":"Allow","mountnsid":4026533142}
[...]

PR #1160, opened 2022-11-30
pkg/runcnotify: Use self-pipe trick to wake up poll()

mauriciovasquezbernal:
The runcnotify package uses poll() to monitor the file descriptors (pidfd) of containers to detect when they are removed. A 1000ms timeout is used to make poll() return and update the list of file descriptors to monitor. This solution is not perfect since after a creation of a container, this mechanism could take up to 1 second to start monitoring for its deletion, also, from a performance point of view to unblock each second to rebuild the file descriptors list is not optimal.

This commit uses the self-pipe trick[0]: an end of the pipe is monitored by poll() and the other end is written by another goroutine to unblock it. This approach updates the list of watched fds faster as we don't have to wait for the timeout, also this timeout is completed removed and poll() only returns when there is an event on some file descriptors.

[0]: https://man7.org/linux/man-pages/man2/pselect.2.html

This PR fixes **some** of the issues described in https://github.com/inspektor-gadget/inspektor-gadget/issues/1001#issuecomment-1329868007. There are still some possible race conditions but I'll investigate them separately.

### DIFF
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -91,6 +91,8 @@ type RuncNotifier struct {
 	closed bool
 
 	wg sync.WaitGroup
+
+	pipeFds []int
 }
 
 // runcPaths is the list of paths where runc could be installed. Depending on
@@ -158,6 +160,7 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
 	n := &RuncNotifier{
 		callback:   callback,
 		containers: make(map[string]*runcContainer),
+		pipeFds:    []int{-1, -1},
 	}
 
 	runcBinaryNotify, err := initFanotify()
@@ -191,12 +194,17 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
 	}
 
 	n.wg.Add(2)
-	go n.watchRunc()
 	if pidfdOpenAvailable {
+		if err := unix.Pipe2(n.pipeFds, unix.O_NONBLOCK); err != nil {
+			runcBinaryNotify.File.Close()
+			return nil, fmt.Errorf("creating pipe: %w", err)
+		}
+
 		go n.watchContainersTermination()
 	} else {
 		go n.watchContainersTerminationFallback()
 	}
+	go n.watchRunc()
 
 	return n, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/runcfanotify/runcfanotify.go
```diff
@@ -240,14 +243,41 @@ func (n *RuncNotifier) AddWatchContainerTermination(containerID string, containe
...
+
 // watchContainerTermination waits until the container terminates using
 // pidfd_open (Linux >= 5.3), then sends a notification.
 func (n *RuncNotifier) watchContainersTermination() {
 	defer n.wg.Done()
 
+	// create a pipe to use the self-pipe trick to "unblock" poll
+	pipeFds := make([]int, 2)
+
+	if err := unix.Pipe2(pipeFds, unix.O_NONBLOCK); err != nil {
```
alban, line 270: It bothers me a bit that the pipe is created in the go routine because awakeWatchContainersTermination() might be called before the pipe is created. I know that you test `if n.pipeFdW != -1` but `pipeFdW` is not protected by a mutex so I am not sure if the compiler would retain the instruction ordering in this function.

What about creating the pipe in the caller NewRuncNotifier() before creating the go routine:

```
	n := &RuncNotifier{
		callback:   callback,
		containers: make(map[string]*runcContainer),
                pipeFds: []int{-1, -1}
...
	if pidfdOpenAvailable {
		// XXX: call Pipe2() here
		go n.watchContainersTermination()
```

and closing the `pipeFds` in RuncNotifier.Close() (after wg.Wait) instead of using a defer in the go routine.

alban: I think that would also fix the unit test `TestDNS` because at the time we call `initialFdList := currentFdList`, the pipes would have been created for sure. With the current code, the pipes might or might not be created, depending on how fast the go routine starts.

pkg/runcfanotify/runcfanotify.go
```diff
@@ -240,14 +243,41 @@ func (n *RuncNotifier) AddWatchContainerTermination(containerID string, containe
...
+		return
+	}
+
+	pipeFdR := pipeFds[0]
+	n.pipeFdW = pipeFds[1]
+	defer func() {
+		unix.Close(pipeFds[0])
+		unix.Close(pipeFds[1])
```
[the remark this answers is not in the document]
burak-ok: What if we created the pipe in `NewRuncNotifier` and closed it only in `Close()`. Downside is an "unused" pipe while watchContainersTermination is not running.

pkg/runcfanotify/runcfanotify.go
mauriciovasquezbernal: Yes, that's much better. It also helps to handle the return error from Pipe2(). Updated the PR.

pkg/runcfanotify/runcfanotify.go
```diff
@@ -606,6 +639,12 @@ func (n *RuncNotifier) watchRuncIterate() (bool, error) {
 
 func (n *RuncNotifier) Close() {
 	n.closed = true
+	n.awakeWatchContainersTermination()
+	for _, fd := range n.pipeFds {
+		if fd != -1 {
+			unix.Close(fd)
```
alban, line 645: Should we make `RuncNotifier.Close()` safe if it is called several times?

alban (approved):
LGTM!

I think that `n.containers` needs to be protected by a mutex but that's not related to this PR...

pkg/runcfanotify/runcfanotify.go
mauriciovasquezbernal: I don't think we should care about that case in all places. I'd leave this up to the caller.

mauriciovasquezbernal:
> I think that n.containers needs to be protected by a mutex but that's not related to this PR... 

Could you place clarify in which case a mutex is needed? (there is a mutex to protect that and IIUC it's safe now).

pkg/runcfanotify/runcfanotify.go
```diff
@@ -289,6 +315,13 @@ func (n *RuncNotifier) watchContainersTermination() {
 				continue
 			}
 
+			// if this is the pipe, read and continue processing
+			if i == 0 {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Done!

Issue #866, opened 2022-08-24
local-gadget: Run advise/seccomp gadget using tracers instead of the Tracer CR


blanquicet:
## Current situation

See https://github.com/kinvolk/inspektor-gadget/issues/861 for further details.

## Ideal future situation

We can run `local-gadget advise seccomp ...` and it uses [this package](https://github.com/kinvolk/inspektor-gadget/tree/main/pkg/gadgets/advise/seccomp/tracer) instead of using the Trace CR. 

For sure, network-policy doesn't make sense for `local-gadget`, so it doesn't have to be implemented. Instead, for `advise/seccomp`, it could make sense. To be investigated.

PR #1161, opened 2022-11-30
local-gadget: Add seccomp-profile advise gadget

blanquicet:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/866

### DIFF
--- a/cmd/local-gadget/advise/seccomp-profile.go
+++ b/cmd/local-gadget/advise/seccomp-profile.go
@@ -1,0 +1,140 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package advise
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/spf13/cobra"
+
+	commonadvise "github.com/inspektor-gadget/inspektor-gadget/cmd/common/advise"
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/local-gadget/utils"
+	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	seccompAdviseTracer "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/advise/seccomp/tracer"
+	localgadgetmanager "github.com/inspektor-gadget/inspektor-gadget/pkg/local-gadget-manager"
+)
+
+const localGadgetSubKey = "local-gadget-key"
+
+func newSeccompProfileCmd() *cobra.Command {
+	var commonFlags utils.CommonFlags
+
+	runCmd := func(cmd *cobra.Command, args []string) error {
+		if commonFlags.Containername == "" {
+			return commonutils.WrapInErrMissingArgs("--containername / -c")
+		}
+
+		localGadgetManager, err := localgadgetmanager.NewManager(commonFlags.RuntimeConfigs)
+		if err != nil {
+			return commonutils.WrapInErrManagerInit(err)
+		}
+		defer localGadgetManager.Close()
+
+		tracer, err := seccompAdviseTracer.NewTracer()
+		if err != nil {
+			return fmt.Errorf("creating tracer: %w", err)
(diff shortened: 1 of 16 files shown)


pkg/gadget-collection/gadgets/advise/seccomp/gadget.go
```diff
@@ -532,7 +532,7 @@ func (t *Trace) Generate(trace *gadgetv1alpha1.Trace) {
 
 	switch trace.Spec.OutputMode {
 	case gadgetv1alpha1.TraceOutputModeStatus:
-		policy := syscallArrToLinuxSeccomp(b)
+		policy := seccomptracer.SyscallArrToLinuxSeccomp(b)
```
mauriciovasquezbernal, line 535: I was thinking that the tracer should return the syscalls names directly. What do you think?

cmd/local-gadget/advise/seccomp-profile.go
```diff
@@ -0,0 +1,141 @@
...
+		// EventTypeAddContainer event.
+		if n == 1 {
+			mntns = containers[0].Mntns
+		}
+
+		stop := make(chan os.Signal, 1)
+		signal.Notify(stop, syscall.SIGINT, syscall.SIGTERM)
+
+		// The profile can be generated upon container termination or when user
+		// requests it through a signal.
```
mauriciovasquezbernal, line 102: I have some doubts about the UX. What's the advantage of generating the policy when the container terminates compared to only generating it when the users press "Ctrl + C"?

mauriciovasquezbernal (commented):
One comment in the UX before reviewing more.

cmd/local-gadget/advise/seccomp-profile.go
blanquicet: IMHO, when the pod terminates, and the gadget generates the policy, I believe that "Ctrl + C" is unnecessary. In addition, it doesn't make sense to me that it continues running after having already generated the policy (i.e. doing nothing). Don't you think?

pkg/gadget-collection/gadgets/advise/seccomp/gadget.go
blanquicet: Yep, that makes sense. Let me try.

pkg/gadgets/advise/seccomp/tracer/tracer.go
```diff
@@ -78,24 +80,41 @@ func NewTracer() (*Tracer, error) {
...
+	sort.Strings(names)
+	return names
+}
+
+func (t *Tracer) Peek(mntns uint64) []string {
 	b, err := t.seccompMap.LookupBytes(mntns)
 	if err != nil {
 		log.Errorf("Error while looking up the seccomp map: %s", err)
-		return make([]byte, C.SYSCALLS_COUNT)
+		return make([]string, 0)
```
mauriciovasquezbernal, line 104: I think this should return the error here so the caller can handle it. (I know it's not related to this PR but I think we can take the opportunity to fix it). Same in other places.

cmd/local-gadget/advise/seccomp-profile.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	cmd := commonadvise.NewSeccompProfileCmd(runCmd)
+
+	utils.AddCommonFlags(cmd, &commonFlags)
+
+	return cmd
+}
+
+func generateProfile(tracer *seccompAdviseTracer.Tracer, mntns uint64) error {
```
mauriciovasquezbernal, line 126: nit: `printProfile`?

cmd/local-gadget/advise/seccomp-profile.go
mauriciovasquezbernal: Right, makes sense to me then. I was thinking about cases where one wants to run multiple times a container with the same name and create a profile that includes syscalls from all runs but seems weird and too specific.

cmd/local-gadget/advise/seccomp-profile.go
```diff
@@ -0,0 +1,141 @@
...
+			select {
+			case <-stop:
+				break WaitForEvent
+			case err := <-containerNotifier:
+				if err != nil {
+					return err
+				}
+				break WaitForEvent
+			}
+		}
```
mauriciovasquezbernal, line 114: I suppose something like works the same?

```suggestion
		select {
		case <-stop:
		case err := <-containerNotifier:
			if err != nil {
				return err
			}
		}
```

mauriciovasquezbernal (commented):
Some small comments around. 

I see you removed the interactive mode and the tests related, do you plan to add tests in this PR?

cmd/local-gadget/advise/seccomp-profile.go
blanquicet: Mmmmm yep, I think that's a weird use case. I would expect users will trigger all the syscalls in a single run.

mauriciovasquezbernal (approved):
LGTM. Only missing comment is about tests. I'm ok with adding them in a follow up PR.

blanquicet:
> Only missing comment is about tests. I'm ok with adding them in a follow up PR.

Oops. Sorry, I missed it. Yeah, I plan to add that test back once https://github.com/inspektor-gadget/inspektor-gadget/pull/1162 is merged.

PR #1162, opened 2022-11-30
integration: Add docker containers integration-tests

mqasimsarfraz:
Add support for running integration-tests for non-Kubernetes containers. The implementation is divided into following steps:
- Introduce a package [`container-utils/testutils`](https://github.com/inspektor-gadget/inspektor-gadget/pull/1162/commits/794b10e222c73b5f3a7ea86e5c8d368b1553ae94) to facilitate management of non-Kubernetes containers. Currently, only `docker` is supported. Use functional options to make creation API cleaner. 
- Introduce an TestStep [interface](https://github.com/inspektor-gadget/inspektor-gadget/pull/1162/commits/7889b2adbb2b78261d69e30bad5d72dbaaa7da81) allowing a  consistent test workflow (Courtesy @blanquicet )
-  Adds tests for docker containers and hook them in the CI.


integration/local-gadget/docker/list_containers_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// start a long-running command i.e with StartAndStop
+	watchContainersCmd.Run(t)
+	defer watchContainersCmd.Stop(t)
+	time.Sleep(2 * time.Second)
+
+	testutils.RunDockerContainer(context.Background(), t,
+		testutils.WithName(cn),
+	)
```
blanquicet, line 119: I was wondering if we could reuse the logic we already have to run commands to run these together. What about using an interface like [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/c6d69f16061f2fe0c12a3152b019f5f8e6e68bfe)? WDYT?

blanquicet (commented):
Great to see this initial support! Thanks! I added some comments, but it looks great.

integration/local-gadget/docker/list_containers_test.go
mqasimsarfraz: Great suggestion. If we could utilize the `RunCommands` it would keep writing tests simple. I see we are trying to introduce an interface in #1116 as well. We can coordinate both the changes and simply this. 

cc: @burak-ok

blanquicet: I think #1116 will not be used because we didn't see a real advantage. So, you can go ahead here.

blanquicet: In any case, in case we need to do something similar, the interface I'm proposing is generic enough to add any additional operation in the future.

mqasimsarfraz: > In any case, in case we need to do something similar, the interface I'm proposing is generic enough to add any additional operation in the future.

+1, I can start with something minimal and can always be extended later on.

pkg/container-utils/testutils/options.go
```diff
@@ -0,0 +1,89 @@
...
+	return func(opts *containerOptions) {
+		opts.removal = false
+	}
+}
+
+func WithRemoval() Option {
+	return func(opts *containerOptions) {
+		opts.removal = true
+	}
+}
```
blanquicet, line 83: Do we really need these two functions? Maybe mix them into one or just define the one that sets the opposite to the default value?

integration/command.go
```diff
@@ -285,6 +269,14 @@ func getInspektorGadgetLogs() string {
 	return sb.String()
 }
 
+// detectKubernetes checks if we are running in a Kubernetes environment
+func detectKubernetes() bool {
+	if _, err := os.Stat(filepath.Join(homedir.HomeDir(), "kube", "config")); err != nil {
+		return false
+	}
+	return true
+}
```
blanquicet, line 278: Mmmm cc @alban?

blanquicet (approved):
I like it. Some comments, but it looks good to me.

integration/opeartions.go
```diff
@@ -0,0 +1,81 @@
...
+// limitations under the License.
+
+package integration
+
+import (
+	"testing"
+	"time"
+)
+
+type Operation interface {
```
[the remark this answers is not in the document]
mqasimsarfraz: True, naming is so hard :D I actually liked `Operation` but adding `Test` prefix (`TestOperation`) might be bit more clear.  Also fine with `TestStep`. What do you say?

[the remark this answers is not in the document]
blanquicet: If we add the `Test` prefix, I will go for `TestStep`.

integration/command.go
blanquicet: But if I am running local-gadget to debug a Kubernetes cluster, the test will try to print the IG logs, isn't it? If so, I think you can think about another solution for this in another PR to avoid blocking this one.

blanquicet (approved):
IMO, it is ready to go. I let you the decision of the naming :)

integration/docker.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// DockerContainer implements Operation for docker containers
+type DockerContainer struct {
+	Name    string
+	Cmd     string
+	Options []dockerOption
+
+	Cleanup      bool
+	Started      bool
```
blanquicet, line 31: It doesn't need to be exposed, isn't it?

```suggestion
	started      bool
```

integration/command.go
blanquicet: I didn't test it, but do you think something like https://github.com/inspektor-gadget/inspektor-gadget/commit/52e7e35ee696289827637ef05076711b40b37d7e may work?

mqasimsarfraz: This is looking like something we can work with. I will handle it like that!

mqasimsarfraz:
> IMO, it is ready to go. I let you the decision of the naming :)

After bit more thinking I felt that `TestStep` seems better so I updated the PR. :) Can you please have a look again before I merge this?

integration/docker.go
```diff
@@ -0,0 +1,90 @@
...
+package integration
+
+import (
+	"context"
+	"testing"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/testutils"
+)
+
+// DockerContainer implements Operation for docker containers
```
blanquicet, line 24:
```suggestion
// DockerContainer implements TestStep for docker containers
```

blanquicet (approved):
LGTM! Thanks for managing all comments!

PR #1163, opened 2022-11-30
local-gadget: Add profile gadgets

burak-ok:
# local-gadget: Add profile gadgets

This is my first time working with `local-gadget`. Any kind of feedback is welcome :)

A new kind of change is introduced in this PR: I imported `github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/types` in `cmd/common/profile/block-io.go`.
Is this okay? Both local and kubectl -gadget use it the same way and i wanted to generalize as much as possible.


cmd/common/profile/cpu.go
```diff
@@ -15,14 +15,28 @@
...
 import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
 	"github.com/spf13/cobra"
+
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/cpu/types"
```
mauriciovasquezbernal, line 26: nit: we're trying to name then `<gadget>Types`, then it should be `cpuTypes`.

cmd/local-gadget/profile/profile.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/local-gadget/utils"
+	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-collection/gadgets/profile"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	localgadgetmanager "github.com/inspektor-gadget/inspektor-gadget/pkg/local-gadget-manager"
+)
+
+// ProfileGadget represents a gadget belonging to the profile category.
+type ProfileGadget[Report any] struct {
+	gadgetName    string
```
mauriciovasquezbernal, line 38: This is also not needed.

cmd/local-gadget/profile/block-io.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/tracer"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/types"
+)
+
+func newBlockIOCmd() *cobra.Command {
+	var commonFlags utils.CommonFlags
+
+	runCmd := func(*cobra.Command, []string) error {
+		blockIOGadget := &ProfileGadget[types.Report]{
```
mauriciovasquezbernal, line 33: Since this gadget doesn't support filtering, shouldn't we check if any filter is set? (this is done on the server side for the kubectl-gadget case https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadget-collection/gadgets/profile/block-io/gadget.go#L89)

cmd/common/profile/cpu.go
burak-ok: Ok. I was under the impression that we should only name then when package name clashes. Changed

cmd/local-gadget/profile/block-io.go
burak-ok: Like `--containername` and `--runtimes` ?

mauriciovasquezbernal: Right, and then `--namespace`, `--podname` in the future.

mauriciovasquezbernal (commented):
LGTM once the extra checks on the filtering for the block-io gadget is done.

mauriciovasquezbernal (approved):
LGTM, thanks!

cmd/local-gadget/profile/cpu.go
```diff
@@ -0,0 +1,75 @@
...
+		if cpuFlags.ProfileUserOnly && cpuFlags.ProfileKernelOnly {
+			return commonutils.WrapInErrArgsNotSupported("-U and -K can't be used at the same time")
+		}
+
+		parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, types.GetColumns())
+		if err != nil {
+			return commonutils.WrapInErrParserCreate(err)
+		}
+
+		profileGadget := &ProfileGadget[types.Report]{
```
blanquicet, line 50: nit:
```suggestion
		cpuGadget := &ProfileGadget[types.Report]{
```

cmd/local-gadget/profile/cpu.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/cpu/tracer"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/cpu/types"
+)
+
+type CPUParser struct {
+	commonutils.GadgetParser[types.Report]
+	commonutils.OutputConfig
+	cpuFlags *commonprofile.CPUFlags
+}
```
blanquicet, line 34: It can be removed, it is not used.

cmd/kubectl-gadget/profile/profile.go
```diff
@@ -20,25 +20,20 @@ import (
 	"os/signal"
 	"time"
 
+	commonprofile "github.com/inspektor-gadget/inspektor-gadget/cmd/common/profile"
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/utils"
 	gadgetv1alpha1 "github.com/inspektor-gadget/inspektor-gadget/pkg/apis/gadget/v1alpha1"
 	"github.com/spf13/cobra"
```
blanquicet, line 27: Could you please take the opportunity to fix the importing order?

cmd/local-gadget/profile/block-io.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/spf13/cobra"
+
+	commonprofile "github.com/inspektor-gadget/inspektor-gadget/cmd/common/profile"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/local-gadget/utils"
+	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-collection/gadgets/profile"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/tracer"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/types"
```
blanquicet, line 30: nit: we are using notation <gadget>Types and <gadget>Tracer:

```suggestion
	bioTracer "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/tracer"
	bioTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/types"
```

cmd/local-gadget/profile/cpu.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/spf13/cobra"
+
+	commonprofile "github.com/inspektor-gadget/inspektor-gadget/cmd/common/profile"
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/local-gadget/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-collection/gadgets/profile"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/cpu/tracer"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/cpu/types"
```
blanquicet, line 27: dito

cmd/local-gadget/profile/profile.go
```diff
@@ -0,0 +1,118 @@
...
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/local-gadget/utils"
+	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-collection/gadgets/profile"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	localgadgetmanager "github.com/inspektor-gadget/inspektor-gadget/pkg/local-gadget-manager"
+)
+
+// ProfileGadget represents a gadget belonging to the profile category.
+type ProfileGadget[Report any] struct {
```
blanquicet, line 37: Why does it need to be a generic type? I don't think it is needed as tracers generate the output in string format, and `DisplayResultsCallback` is managing types internally for CPU gadget.

cmd/local-gadget/profile/profile.go
```diff
@@ -0,0 +1,118 @@
...
+	containerSelector := containercollection.ContainerSelector{
+		Name: g.commonFlags.Containername,
+	}
+
+	// Create mount namespace map to filter by containers
+	mountnsmap, err := localGadgetManager.CreateMountNsMap(containerSelector)
+	if err != nil {
+		return commonutils.WrapInErrManagerCreateMountNsMap(err)
+	}
+	defer localGadgetManager.RemoveMountNsMap()
```
blanquicet, line 66: All this is not needed for block-io gadget. What about moving it into the gadget itself? Something like [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/a7e3fbc0f94fe616f6f5fd6cc929be33303a92dd)?

blanquicet (commented):
I added some comments but, in general, it looks good. Notice that integration tests are missing, but you can add them in another PR if you want.

cmd/local-gadget/profile/profile.go
burak-ok: Removing it. I think it is a remnant of a previous Idea

cmd/local-gadget/profile/cpu.go
```diff
@@ -0,0 +1,88 @@
...
+				containerSelector := containercollection.ContainerSelector{
+					Name: commonFlags.Containername,
+				}
+
+				// Create mount namespace map to filter by containers
+				mountnsmap, err := localGadgetManager.CreateMountNsMap(containerSelector)
+				if err != nil {
+					return nil, commonutils.WrapInErrManagerCreateMountNsMap(err)
+				}
+				defer localGadgetManager.RemoveMountNsMap()
```
blanquicet, line 70: Mmmmm I think this should go within the `runCmd` context to avoid `defers` are called after `NewTracer` returns.

burak-ok: Now that you have mentioned it: Doesn't it have to stay inside of the `Run()` method of `cmd/local-gadget/profile/profile.go`?

The `mountnsmap` and `localGadgetMangaer` needs to exist while the tracer is running, or not?

blanquicet: Yep, that's what I am saying that we need to move it one level up, in the `runCmd` instead of `createAndRunTracer` context. In this way, they will exist while the gadget is running.

burak-ok: Correct, didn't see that the first time. `cpuGadget.Run()` calls the `Run()` of the `ProfileGadget` so everything is good

burak-ok:
I added a `--timeout` parameter for the `profile` gadgets in `local-gadget`. Should we add it for other gadgets in `local-gadget` in another PR too?

blanquicet:
> Should we add it for other gadgets in `local-gadget` in another PR too?

Yep, but it can be done in another PR. I noticed we are missing it when reviewing the PR that adds the top category.

burak-ok:
Issue created: https://github.com/inspektor-gadget/inspektor-gadget/issues/1219

PR #1167, opened 2022-12-01
local-gadget: Add support for printing container deletion events

### LINKED ISSUE #1028: [RFE] local-gadget: Add `--watch` option to the `list-containers` command

mqasimsarfraz:
Add support for watching for container deletion events

Fixes #1028 

### Testing done
```
$ sudo ./local-gadget list-containers --watch
# start a container in other terminal using docker run --rm alpine sleep 5
TIMESTAMP                      EVENT      RUNTIME    ID                                                     NAME                                                                                                                        
2022-12-01T19:13:10+01:00      CREATED    docker     c6b7cb5f7ab2c4f7db33c22c77d219010c7e0a2512fd3f3c65b31… busy_knuth                                                                                                                  
2022-12-01T19:13:16+01:00      DELETED    docker     c6b7cb5f7ab2c4f7db33c22c77d219010c7e0a2512fd3f3c65b31… busy_knuth   
```

### DIFF
--- a/cmd/local-gadget/containers/containers.go
+++ b/cmd/local-gadget/containers/containers.go
@@ -20,11 +20,13 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/local-gadget/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
 	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
 	localgadgetmanager "github.com/inspektor-gadget/inspektor-gadget/pkg/local-gadget-manager"
 )
@@ -45,45 +47,62 @@ func NewListContainersCmd() *cobra.Command {
 			}
 			defer localGadgetManager.Close()
 
-			parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, containercollection.GetColumns())
-			if err != nil {
-				return commonutils.WrapInErrParserCreate(err)
+			selector := containercollection.ContainerSelector{
+				Name: commonFlags.Containername,
 			}
 
-			containers := localGadgetManager.ContainerCollection.Subscribe(
-				localGadgetSubKey,
-				containercollection.ContainerSelector{
-					Name: commonFlags.Containername,
-				},
-				func(event containercollection.PubSubEvent) {
-					switch event.Type {
-					case containercollection.EventTypeAddContainer:
-						if err = printContainer(parser, commonFlags, event.Container); err != nil {
-							fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
-						}
-					}
-				})
-			defer localGadgetManager.ContainerCollection.Unsubscribe(localGadgetSubKey)
-
-			parser.Sort(containers, []string{"runtime", "name"})
-			if optionWatch {
-				if commonFlags.OutputMode != commonutils.OutputModeJSON {
-					fmt.Println(parser.BuildColumnsHeader())
+			if !optionWatch {
+				parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, containercollection.GetColumns())
+				if err != nil {
+					return commonutils.WrapInErrParserCreate(err)
 				}
-				for _, container := range containers {
-					if err = printContainer(parser, commonFlags, container); err != nil {
-						return err
-					}
-				}
+				containers := localGadgetManager.GetContainersBySelector(&selector)
 
-				stop := make(chan os.Signal, 1)
-				signal.Notify(stop, syscall.SIGINT)
-				<-stop
-			} else {
+				parser.Sort(containers, []string{"runtime", "name"})
 				if err = printContainers(parser, commonFlags, containers); err != nil {
 					return err
 				}
+				return nil
 			}
+
+			cols := columns.MustCreateColumns[containercollection.PubSubEvent]()
+			cols.SetExtractor("event", func(event *containercollection.PubSubEvent) string {
+				return event.Type.String()
+			})
+
+			parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, cols)
+			if err != nil {
+				return commonutils.WrapInErrParserCreate(err)
+			}
+			containers := localGadgetManager.ContainerCollection.Subscribe(
+				localGadgetSubKey,
+				selector,
+				func(event containercollection.PubSubEvent) {
+					if err = printPubSubEvent(parser, commonFlags, &event); err != nil {
+						fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
+					}
+				},
+			)
+			defer localGadgetManager.ContainerCollection.Unsubscribe(localGadgetSubKey)
+
+			if commonFlags.OutputMode != commonutils.OutputModeJSON {
+				fmt.Println(parser.BuildColumnsHeader())
+			}
+			timestamp := time.Now().Format(time.RFC3339)
+			for _, container := range containers {
+				e := containercollection.PubSubEvent{
+					Timestamp: timestamp,
+					Type:      containercollection.EventTypeAddContainer,
+					Container: container,
+				}
+				if err = printPubSubEvent(parser, commonFlags, &e); err != nil {
+					return err
+				}
+			}
+
+			stop := make(chan os.Signal, 1)
+			signal.Notify(stop, syscall.SIGINT)
+			<-stop
 
 			return nil
 		},
@@ -118,10 +137,10 @@ func printContainers(parser *commonutils.GadgetParser[containercollection.Contai
 	return nil
 }
 
-func printContainer(parser *commonutils.GadgetParser[containercollection.Container], commonFlags utils.CommonFlags, container *containercollection.Container) error {
+func printPubSubEvent(parser *commonutils.GadgetParser[containercollection.PubSubEvent], commonFlags utils.CommonFlags, event *containercollection.PubSubEvent) error {
 	switch commonFlags.OutputMode {
 	case commonutils.OutputModeJSON:
-		b, err := json.MarshalIndent(container, "", "  ")
+		b, err := json.MarshalIndent(event, "", "  ")
 		if err != nil {
 			return commonutils.WrapInErrMarshalOutput(err)
 		}
@@ -129,7 +148,7 @@ func printContainer(parser *commonutils.GadgetParser[containercollection.Contain
 	case commonutils.OutputModeColumns:
 		fallthrough
 	case commonutils.OutputModeCustomColumns:
-		fmt.Println(parser.TransformIntoColumns(container))
+		fmt.Println(parser.TransformIntoColumns(event))
 	}
 
 	return nil
--- a/integration/local-gadget/list_containers_test.go
+++ b/integration/local-gadget/list_containers_test.go
@@ -125,9 +125,9 @@ func TestFilterByContainerName(t *testing.T) {
 	RunCommands(commands, t)
 }
 
-func TestWatchContainers(t *testing.T) {
+func TestWatchCreatedContainers(t *testing.T) {
 	t.Parallel()
-	cn := "test-watched-container"
+	cn := "test-created-container"
 	ns := GenerateTestNamespaceName(cn)
 
 	// TODO: Handle it once we support getting K8s container name for docker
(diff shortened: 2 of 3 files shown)


pkg/container-collection/containers.go
```diff
@@ -37,6 +37,9 @@ type Container struct {
 	// Pid is the process id of the container
 	Pid uint32 `json:"pid,omitempty" column:"pid,template:pid,hide"`
 
+	// CreatedAt represent the creation time of the container in time.RFC3339
+	CreatedAt string `json:"createdAt,omitempty"`
```
mauriciovasquezbernal, line 41: Do we need this timestamp here? Or is it something we can calculate in `cmd/local-gadget/containers.go` when we receive the notification? 

IMO we should try to keep our representation of containers as small as possible and only add new members when totally required. 

@alban any opinions on this?

pkg/container-collection/containers.go
```diff
@@ -67,6 +70,13 @@ type Container struct {
 	ownerReference *metav1.OwnerReference
 }
 
+type ContainerEvent struct {
+	Timestamp string `json:"timestamp,omitempty" column:"timestamp,maxWidth:30" columnTags:"runtime"`
+	Event     string `json:"event,omitempty" column:"event,maxWidth:10" columnTags:"runtime"`
+
+	Container
+}
```
mauriciovasquezbernal, line 78: How about using https://github.com/inspektor-gadget/inspektor-gadget/blob/3429862cec69552f890c975c2d3817a13b737ab0/pkg/container-collection/pubsub.go#L30-L33

cmd/local-gadget/containers/containers.go
```diff
@@ -45,44 +46,57 @@ func NewListContainersCmd() *cobra.Command {
...
+			} else {
+				parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, containercollection.GetContainerEventColumns())
+				if err != nil {
+					return commonutils.WrapInErrParserCreate(err)
+				}
+				containers := localGadgetManager.ContainerCollection.Subscribe(
+					localGadgetSubKey,
+					containercollection.ContainerSelector{
+						Name: commonFlags.Containername,
+					},
```
mauriciovasquezbernal, line 70: nit: You can extract this part as it's common in both cases.

pkg/container-collection/containers.go
```diff
@@ -167,3 +177,15 @@ func GetColumns() *columns.Columns[Container] {
...
+	return columns.MustCreateColumns[ContainerEvent]()
+}
+
+func ContainerEventFromContainer(container Container, event, timestamp string) *ContainerEvent {
+	return &ContainerEvent{
+		Container: container,
+		Timestamp: timestamp,
+		Event:     event,
+	}
+}
```
mauriciovasquezbernal, line 191: I think this function is too specific to being added to the package. The callers could implement it.

mauriciovasquezbernal (commented):
The functionality looks good but I think we should not put that much code in the containers collection package to implement it. I think the use case is very specific and almost all modifications could be implemented directly in `cmd/local-gadget/containers.go`.

pkg/container-collection/containers.go
mqasimsarfraz: I wasn't sure about this as well so I agree it should be on the client side. I can move this to `cmd/local-gadget/containers`

pkg/container-collection/containers.go
mqasimsarfraz: > Do we need this timestamp here? Or is it something we can calculate in `cmd/local-gadget/containers.go` when we receive the notification?

We mainly need it for initial containers like to show users when the runtime created the container (it can be a week ago time) so tricky to calculate it but for new containers we can just show the current time. 

Having said above, I was wondering if we can skip showing timestamp or do you think it is nice to have it in `ContainerEvents`?

mqasimsarfraz: We can print the current time in [timestamp](https://github.com/inspektor-gadget/inspektor-gadget/pull/1167/files#diff-4649a03815777bd0b520ade836043a3bb4baad34f2ec2e22f8d2b8277110082cR40). Should simplify all the changes to `containers-utils`.

pkg/container-collection/containers.go
mqasimsarfraz: I had a look into this and we need add support for handling pointers to struct in `columns` library. Is this a use case we want to support? 

cc: @flyth

pkg/container-collection/containers.go
mauriciovasquezbernal: I agree with printing the current timestamp.

mauriciovasquezbernal (approved):
I think it's good to go. @blanquicet do you want to review this given that you opened the issue?

cmd/local-gadget/containers/containers.go
```diff
@@ -118,19 +129,27 @@ func printContainers(parser *commonutils.GadgetParser[containercollection.Contai
...
 	return nil
 }
+
+func eventFromPubSubEvent(e containercollection.PubSubEvent) event {
+	return event{
+		Timestamp: time.Now().Format(time.RFC3339),
+		Event:     e.Type.String(),
+		Container: *e.Container,
+	}
+}
```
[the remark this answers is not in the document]
mqasimsarfraz: Yep, it should be like that. #1179 is looking ready so let me base my PR on that. Let's handle this in the same PR since I also wanted to use `containercollection.PubSubEvent` directly. Also, I am afraid it is easy to lose a change in which we just have to play with '&'/'*' :laughing:

integration/local-gadget/list_containers_test.go
```diff
@@ -178,3 +185,63 @@ func TestWatchContainers(t *testing.T) {
...
+				e.Netns = 0
+				e.CgroupPath = ""
+				e.CgroupID = 0
+				e.CgroupV1 = ""
+				e.CgroupV2 = ""
+				e.Labels = nil
+				e.PodUID = ""
+			}
+
+			return ExpectEntriesToMatch(output, normalize, expectedEvent)
```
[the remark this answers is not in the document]
mqasimsarfraz: Since the container is running we do get the running container in the list so need to do `ExpectEntriesToMatch`

cmd/local-gadget/containers/containers.go
```diff
@@ -45,7 +53,7 @@ func NewListContainersCmd() *cobra.Command {
 			}
 			defer localGadgetManager.Close()
 
-			parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, containercollection.GetColumns())
+			parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, columns.MustCreateColumns[event]())
 			if err != nil {
 				return commonutils.WrapInErrParserCreate(err)
 			}
```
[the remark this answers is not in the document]
mqasimsarfraz: Done

integration/local-gadget/list_containers_test.go
```diff
@@ -178,3 +182,60 @@ func TestWatchContainers(t *testing.T) {
...
+		},
+	}
+
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		PodCommand(cn, "busybox", ns, `["sleep", "5"]`, ""),
+		WaitUntilPodReadyCommand(ns, cn),
+		watchContainersCmd,
+		DeleteTestNamespaceCommand(ns),
+	}
```
blanquicet, line 238: Isn't it a little bit tricky? I mean, it relies on the fact that the `watchContainersCmd` will last 10 seconds to capture the deletion. What about:

```suggestion
	commands := []*Command{
		CreateTestNamespaceCommand(ns),
		PodCommand(cn, "busybox", ns, `["sleep", "inf"]`, ""),
		WaitUntilPodReadyCommand(ns, cn),
		watchContainersCmd,
		SleepForSecondsCommand(2), // wait to ensure local-gadget has started
		{
			Name:         "DeletePod",
			Cmd:          fmt.Sprintf("kubectl delete pod %s -n %s", cn, ns),
		},
		DeleteTestNamespaceCommand(ns),
	}
```

cmd/local-gadget/containers/containers.go
```diff
@@ -45,33 +47,42 @@ func NewListContainersCmd() *cobra.Command {
...
-				},
-				func(event containercollection.PubSubEvent) {
-					switch event.Type {
-					case containercollection.EventTypeAddContainer:
-						if err = printContainer(parser, commonFlags, event.Container); err != nil {
+			if optionWatch {
+				cols := columns.MustCreateColumns[containercollection.PubSubEvent]()
+				cols.SetExtractor("event", func(event *containercollection.PubSubEvent) string {
+					return event.Type.String()
+				})
```
blanquicet, line 58: I like how the columns library makes this work that easy. @flyth

cmd/local-gadget/containers/containers.go
```diff
@@ -80,6 +91,13 @@ func NewListContainersCmd() *cobra.Command {
 				signal.Notify(stop, syscall.SIGINT)
 				<-stop
 			} else {
+				parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, containercollection.GetColumns())
```
blanquicet, line 94: nit: I prefer the early return approach.

```golang
	if !optionWatch {
		parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, containercollection.GetColumns())
		[...]
		return nil
	}

	cols := columns.MustCreateColumns[containercollection.PubSubEvent]()
	cols.SetExtractor("event", func(event *containercollection.PubSubEvent) string {
		return event.Type.String()
	})
	[...]
```

pkg/container-collection/pubsub.go
```diff
@@ -27,9 +28,41 @@ const (
...
+}
+
+func (e *EventType) MarshalText() (text []byte, err error) {
+	return []byte(e.String()), nil
+}
+
+func (e *EventType) UnmarshalText(bytes []byte) error {
+	*e = EventTypeFromString(string(bytes))
+	return nil
+}
```
blanquicet, line 60: Mmmm, where is this used?

blanquicet (commented):
Awesome! I really like the result!

blanquicet (approved):

integration/local-gadget/list_containers_test.go
mqasimsarfraz: Agree, this is much more clear. Thanks for the suggestion.

PR #1168, opened 2022-12-02
ci: Invalidate install-debian-packages cache when files are changed.

eiffel-fl:
Before this commit, each time you changed the installed packages, you needed to manually update the cache key.
With this commit, the cache key is now crafted as the hash of the install-debian-packages action file and the main workflow file. So, any modifications to these two files will invalidate the cache


mqasimsarfraz (approved):
LGTM! Nice, I confirmed the cache was added [here](https://github.com/inspektor-gadget/inspektor-gadget/actions/caches) with hashed key!

PR #1171, opened 2022-12-05
pkg/trace-collection: Avoid capturing events from wrong container

mauriciovasquezbernal:
Inspektor Gadget uses the mnt namespace inode id of the containers to implement the logic that determines if a given event produced in the kernel should be traced or not.  There is a corner case when this logic traces the events from the wrong container:

1. A tracer is filtering events from the "foo" container
2. The foo container is removed
3. A bar container is created
4. The tracer displays events from the "bar" container

This can happen because:
- The bar container can reuse the mnt ns inode id from the foo container
- The deletion of mount namespace is asynchronous, we can't block it
- Hence there is a window when the mnt ns inode id from the bar container is in the filtering map and hence events coming from it are wrongly captured.

This commit fixes it by keeping a file descriptor of the mnt namespace of the container, guaranteeing that its inode id is not reused until the tracer has updated all eBPF maps.

### How to reproduce

The race condition described here is theoretical, it seems it has happened in https://github.com/inspektor-gadget/inspektor-gadget/issues/1001 but I haven't been able to reproduce it. However, we can use a small trick to alter Inspektor Gadget behavior and reproduce it:

1. Introduce a delay before updating the tracer ebpf maps
```diff 
diff --git pkg/tracer-collection/tracer-collection.go pkg/tracer-collection/tracer-collection.go
index ffe612cf..a4498648 100644
--- pkg/tracer-collection/tracer-collection.go
+++ pkg/tracer-collection/tracer-collection.go
@@ -17,6 +17,7 @@ package tracercollection
 import (
        "fmt"
        "os"
+       "time"
 
        "github.com/cilium/ebpf"
        log "github.com/sirupsen/logrus"
@@ -88,6 +89,8 @@ func (tc *TracerCollection) TracerMapsUpdater() containercollection.FuncNotify {
                        }
 
                case containercollection.EventTypeRemoveContainer:
+                       time.Sleep(2 * time.Second) // simulate the scheduler takes a bit long to run us here
+
                        for _, t := range tc.tracers {
                                if containercollection.ContainerSelectorMatches(&t.containerSelector, event.Container) {
                                        mntnsC := uint64(event.Container.Mntns)
```

2. Run this script in a terminal

```bash 
#! /bin/bash

while true
do
        docker run --rm --name we-broke-it ubuntu readlink /proc/self/ns/mnt &
        docker run --rm --name no-yet ubuntu cat /dev/null &

        for job in `jobs -p`
        do
        wait $job
        done
done
``` 

3. Start trace exec using local-gadget

```bash
sudo ./local-gadget trace exec -c no-yet
[sudo] password for mvb:           
CONTAINER                                         PID        PPID       COMM             RET ARGS                                                             
no-yet                                            334284     334238     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            334546     334518     cat              0   /usr/bin/cat /dev/null                                           
we-broke-it01                                     334811     334788     cat              0   /usr/bin/cat /dev/null                                           
we-broke-it01                                     335085     335047     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            335376     335328     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            335642     335613     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            335900     335872     readlink         0   /usr/bin/readlink /proc/self/ns/mnt                              
no-yet                                            335907     335852     cat              0   /usr/bin/cat /dev/null  
```

It can see that the tracer starts to report events from the wrong container. 

Thanks to @alban for the idea to fix it! 

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1001

### DIFF
--- a/pkg/gadgettracermanager/gadgettracermanager.go
+++ b/pkg/gadgettracermanager/gadgettracermanager.go
@@ -318,5 +318,8 @@ func (g *GadgetTracerManager) Close() {
 	if g.containersMap != nil {
 		g.containersMap.Close()
 	}
+	if g.tracerCollection != nil {
+		g.tracerCollection.Close()
+	}
 	g.ContainerCollection.Close()
 }
--- a/pkg/local-gadget-manager/local-gadget-manager.go
+++ b/pkg/local-gadget-manager/local-gadget-manager.go
@@ -367,5 +367,10 @@ func NewManager(runtimes []*containerutils.RuntimeConfig) (*LocalGadgetManager, 
 
 func (l *LocalGadgetManager) Close() {
 	l.ContainerCollection.Close()
-	l.containersMap.Close()
+	if l.tracerCollection != nil {
+		l.tracerCollection.Close()
+	}
+	if l.containersMap != nil {
+		l.containersMap.Close()
+	}
 }


mauriciovasquezbernal:
I updated the PR to include a mutex given that `TracerMapsUpdater` can be called in parallel.

pkg/tracer-collection/tracer-collection.go
```diff
@@ -190,6 +219,9 @@ func (tc *TracerCollection) TracerExists(id string) bool {
 }
 
 func (tc *TracerCollection) Close() {
+	for _, fd := range tc.containerMntNsFds {
+		unix.Close(fd)
+	}
```
alban, line 224:
```suggestion
	tc.mu.Lock()
	defer tc.mu.Unlock()
	for _, fd := range tc.containerMntNsFds {
		unix.Close(fd)
	}
	tc.containerMntNsFds = nil
```
To support multiple calls to Close().
And also take the mutex in case there is a container event during the Close() call.
I think it is cleaner.

mauriciovasquezbernal: I'm wondering about `tc.containerMntNsFds = nil`, this will cause a crash on `TracerMapsUpdater` if there is a container event after `Close()`. Probably the right solution is to include a `closed` boolean, but it seems to much for me. WDYT?

mauriciovasquezbernal:
Finally I was able to implement a test that simulates the issue this PR is fixing. Feel free to revert the `pkg/trace-collection: Avoid capturing events from wrong container` and see how the test fails without it.

pkg/tracer-collection/tracer-collection.go
alban: Fair enough.

Now that this function is non-empty, we should it in:
* pkg/local-gadget-manager/local-gadget-manager.go LocalGadgetManager.Close().
* pkg/gadgettracermanager/gadgettracermanager.go GadgetTracerManager.Close().

mauriciovasquezbernal: Done!

PR #1180, opened 2022-12-06
dns/sni: Reduce duplicated code in network tracers

blanquicet:
# Reduce duplicated code in network tracers

This PR also moves the attachment key's management internally into the tracer itself to avoid callers generating a key that was always the network namespace (it can be computed internally from the PID).

### DIFF
--- a/pkg/gadgets/internal/networktracer/tracer.go
+++ b/pkg/gadgets/internal/networktracer/tracer.go
@@ -1,0 +1,215 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package networktracer
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"syscall"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/perf"
+	"golang.org/x/sys/unix"
+
+	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/rawsock"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+)
+
+type attachment struct {
+	collection *ebpf.Collection
+	perfRd     *perf.Reader
+
+	sockFd int
+
+	// users keeps track of the users' pid that have called Attach(). This can
+	// happen for two reasons:
+	// 1. several containers in a pod (sharing the netns)
+	// 2. pods with networkHost=true
+	// In both cases, we want to attach the BPF program only once.
+	users map[uint32]struct{}
+}
+
+func newAttachment(
+	pid uint32,
+	spec *ebpf.CollectionSpec,
+	bpfProgName string,
+	bpfPerfMapName string,
+	bpfSocketAttach int,
+) (_ *attachment, err error) {
+	a := &attachment{
+		sockFd: -1,
+		users:  map[uint32]struct{}{pid: {}},
+	}
+	defer func() {
+		if err != nil {
+			if a.perfRd != nil {
+				a.perfRd.Close()
+			}
+			if a.sockFd != -1 {
+				unix.Close(a.sockFd)
+			}
+			if a.collection != nil {
+				a.collection.Close()
+			}
+		}
+	}()
+
+	a.collection, err = ebpf.NewCollection(spec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create BPF collection: %w", err)
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/cilium/ebpf/perf"
+	"golang.org/x/sys/unix"
+
+	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/rawsock"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+)
+
+type Attachment struct {
```
alban, line 33: Does it need to be exported?
```suggestion
type attachment struct {
```

mauriciovasquezbernal (approved):
I was a bit confused because there is a map inside a structure that is saved in another map. Perhaps we can move that inside map to the outer structure and make the code a bit simpler?

I tried it and it's the result:
```diff
$ git diff
diff --git pkg/gadgets/internal/networktracer/tracer.go pkg/gadgets/internal/networktracer/tracer.go
index 1fcc8423..0e38d6a6 100644
--- pkg/gadgets/internal/networktracer/tracer.go
+++ pkg/gadgets/internal/networktracer/tracer.go
@@ -36,12 +36,7 @@ type attachment struct {
 
        sockFd int
 
-       // users keeps track of the users' pid that have called Attach(). This can
-       // happen for two reasons:
-       // 1. several containers in a pod (sharing the netns)
-       // 2. pods with networkHost=true
-       // In both cases, we want to attach the BPF program only once.
-       users map[uint32]struct{}
+       users int
 }
 
 func newAttachment(
@@ -53,7 +48,7 @@ func newAttachment(
 ) (_ *attachment, err error) {
        a := &attachment{
                sockFd: -1,
-               users:  map[uint32]struct{}{pid: {}},
+               users:  1,
        }
        defer func() {
                if err != nil {
@@ -102,6 +97,7 @@ type Tracer[Event any] struct {
        // key: network namespace inode number
        // value: Tracelet
        attachments map[uint64]*attachment
+       pidToNetNs  map[uint32]uint64
 
        bpfProgName     string
        bpfPerfMapName  string
@@ -122,6 +118,7 @@ func NewTracer[Event any](
        return &Tracer[Event]{
                spec:            spec,
                attachments:     make(map[uint64]*attachment),
+               pidToNetNs:      make(map[uint32]uint64),
                bpfProgName:     bpfProgName,
                bpfPerfMapName:  bpfPerfMapName,
                bpfSocketAttach: bpfSocketAttach,
@@ -136,7 +133,8 @@ func (t *Tracer[Event]) Attach(pid uint32, eventCallback func(Event)) error {
                return fmt.Errorf("getting network namespace of pid %d: %w", pid, err)
        }
        if a, ok := t.attachments[netns]; ok {
-               a.users[pid] = struct{}{}
+               t.pidToNetNs[pid] = netns
+               a.users++
                return nil
        }
 
@@ -145,6 +143,7 @@ func (t *Tracer[Event]) Attach(pid uint32, eventCallback func(Event)) error {
                return fmt.Errorf("creating network tracer attachment for pid %d: %w", pid, err)
        }
        t.attachments[netns] = a
+       t.pidToNetNs[pid] = netns
 
        go t.listen(netns, a.perfRd, t.baseEvent, t.parseEvent, eventCallback)
 
@@ -196,16 +195,21 @@ func (t *Tracer[Event]) releaseAttachment(netns uint64, a *attachment) {
 }
 
 func (t *Tracer[Event]) Detach(pid uint32) error {
-       for netns, a := range t.attachments {
-               if _, ok := a.users[pid]; ok {
-                       delete(a.users, pid)
-                       if len(a.users) == 0 {
-                               t.releaseAttachment(netns, a)
-                       }
-                       return nil
-               }
+       netns, ok := t.pidToNetNs[pid]
+       if !ok {
+               return fmt.Errorf("pid %d is not attached", pid)
        }
-       return fmt.Errorf("pid %d is not attached", pid)
+
+       delete(t.pidToNetNs, pid)
+
+       a := t.attachments[netns]
+       a.users--
+
+       if a.users == 0 {
+               t.releaseAttachment(netns, a)
+       }
+
+       return nil
 }
 
 func (t *Tracer[Event]) Close() {
```

It's up to you. I'll approve because the PR brings a lot of improvements!

blanquicet:
> I was a bit confused because there is a map inside a structure that is saved in another map. Perhaps we can move that inside map to the outer structure and make the code a bit simpler?
>
Mmmm yeah, we could, but the problem with that approach is that the variables `Tracer.pidToNetNs` and `attachment.users` keep track of basically the same information, so it is a little bit redundant. My idea was to have this:

- Tracer keeps track of the `attachments`, one for each network namespace.
- Each `attachment` keeps track of its users, one for each PID.

That's why we have a map with key netns and value an `attachment`. Then, each `attachment`, among other fields, has a map with key PID and the value `struct{}` (it is just because the lookup is quicker than a slice).

What do you think?

mauriciovasquezbernal:
I think the only difference is in `Detach()`, your proposal has to perform multiple lookups in the different `users` map of each attachment, while the change I suggested only needs two lookups, one in `pidToNetNs` and another one in `attachments`. 

I'm fine with both approaches, if you feel more comfortable with the current one go and merge it!

blanquicet:
> I think the only difference is in `Detach()`, your proposal has to perform multiple lookups in the different `users` map of each attachment, while the change I suggested only needs two lookups, one in `pidToNetNs` and another one in `attachments`.
>
Agree. The lookup is faster, but we also need to update both `users` and `pidToNetNs` at every attach/detach. That's what I find redundant. I think both have pros and cons. I will go ahead with the current approach, but let's monitor it.

Issue #900, opened 2022-08-31
[RFE] Organise `local-gadget` documentation


blanquicet:
## Current situation

As discussed in https://github.com/kinvolk/inspektor-gadget/issues/717#issuecomment-1169809064, the `local-gadget` documentation is not well organized. For instance, the example for the `audit-seccomp` gadget is in [the main guides](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/audit/seccomp.md#with-local-gadget), but it is not present in [the local-gadget doc](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/local-gadget.md).

## Impact

It is unclear to users where to look for the `local-gadget` examples.

## Ideal future situation

We need to define where to put such examples. Together with the Inspektor-Gadget (`kubectl-gadget`) [examples](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides)? Or in [local-gadget.md](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/local-gadget.md)? Or in a dedicated folder for `local-gadget`? Consider that we are working to support all the gadgets (https://github.com/kinvolk/inspektor-gadget/issues/861), so we need to add more and more examples.

mauriciovasquezbernal:
> Together with the Inspektor-Gadget (kubectl-gadget) [examples](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides)?

I think it's better to put examples on the gadget guide, otherwise there will be a lot of repetition. Generally speaking, I think the user experience for each gadget should be very similar on kubectl-gadget and local-gadget, so a single guide per gadget should be enough.

blanquicet:
Notice we are also missing a section on the README for the local-gadget.

blanquicet:
Some other suggestions here: https://github.com/kinvolk/inspektor-gadget/pull/1014#discussion_r988718265.

> Notice we are also missing a section on the README for the local-gadget.
>
Done in https://github.com/kinvolk/inspektor-gadget/pull/1006

PR #1181, opened 2022-12-06
docs: Reorganize docs structure

mauriciovasquezbernal:
This is a PR that updates a bit the code structure. The idea is to make the documentation for gadgets more visible. This is a first step before https://github.com/inspektor-gadget/inspektor-gadget/issues/900. 

- Move gadgets specific crd docs to crds folder
- Rename "guides" to "gadgets"
- Remove repetitive information from gadgets docs


blanquicet (approved):
I like the proposal! Thanks for working on this.

blixtra:
Thanks, @mauriciovasquezbernal! This is a clear improvement. So LGTM.

In the future, we should also add more information about each category which describes common usage, output formats, flags, etc.

Issue #1178, opened 2022-12-06
Event enrichment fails in some situations


mauriciovasquezbernal:
This issue is based on https://github.com/inspektor-gadget/inspektor-gadget/issues/1001#issuecomment-1335943053. 

Different gadgets use the [DataEnricher](https://github.com/inspektor-gadget/inspektor-gadget/blob/1c2f1abff6f6f4d3b13addf6fb3f6e16bc2ae164/pkg/gadgets/helpers.go#L45-L47) to add container information to the events received from the eBPF programs. Example [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/1c2f1abff6f6f4d3b13addf6fb3f6e16bc2ae164/pkg/gadgets/trace/exec/tracer/tracer.go#L174). This is possible that by the time the tracer calls `.Enrich()` the container is gone and then the container information from the event is missing.

### Reproducer

1. Execute the following script that creates and removes containers:
```bash
#! /bin/bash

while true
do
        docker run --rm --name we-broke-it ubuntu readlink /proc/self/ns/mnt &
        docker run --rm --name no-yet ubuntu cat /dev/null &

        for job in `jobs -p`
        do
        wait $job
        done
done
```

2. Stress one core on the CPU
```
$ taskset 0x4 stress-ng -c 1
```
3. Execute local-gadget bound to a single thread with the lowest priority 
```
$  sudo taskset 0x4 nice -20 ./local-gadget trace exec -c no-yet

...
no-yet                                            191525     191477     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            191785     191751     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            192043     191998     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            192300     192238     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            192557     192512     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            192829     192791     cat              0   /usr/bin/cat /dev/null                                           
                                                  193109     193074     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            193366     193304     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            193622     193570     cat              0   /usr/bin/cat /dev/null                                           
no-yet                                            193878     193848     cat              0   /usr/bin/cat /dev/null                                     
``` 

Some of the events miss the container data. 

### Cause

[...]

Issue #1095, opened 2022-11-03
[RFE] container-utils: Include pause containers when using CRI-O and containerd 


blanquicet:
## Current situation

Unlike Docker, when using CRI-O and containerd, neither `GetContainer` nor `GetContainers` report the pause containers. 

**NOTE**: Verify if `GetContainerDetails` behaves similarly. 

## Impact

The container runtime enricher that uses the `runtimeclient.ContainerRuntimeClient` discards the pause containers when using CRI-O or containerd. It means that the container-collection doesn't report the creation/deletion of those containers.

## Ideal future situation

All the `ContainerRuntimeClient` methods should include the pause containers also for CRI-O and containerd. 

## Implementation options

Check  [`ListPodSandbox`](https://pkg.go.dev/k8s.io/cri-api@v0.20.6/pkg/apis/runtime/v1alpha2#RuntimeServiceClient.ListPodSandbox) API.

PR #1182, opened 2022-12-06
pkg/gadget/tracer/network: Enrich events using `EnrichByNetns`

blanquicet:
# Enrich events using `EnrichByNetns`

With this PR, the network-graph gadget starts using `container-collection/networktracer` and changes the strategy to enrich the events, which now uses a new `EnrichByNetns`.


pkg/gadgets/trace/network/tracer/tracer.go
```diff
@@ -113,7 +119,7 @@ func (t *Tracer) Attach(key string, pid uint32) (err error) {
 	}
 
 	consts := map[string]interface{}{
-		"container_quark": t.nextContainerQuark,
+		"container_quark": netns,
```
alban, line 122:
```suggestion
		"current_netns": netns,
```
Would that be a better name for the variable in the ebpf program?

pkg/gadgets/trace/network/tracer/tracer.go
```diff
@@ -43,6 +44,7 @@ type link struct {
 
 	sockFd int
 
+	// Identifies the belonging network namespace of the event
 	containerQuark uint64
```
alban, line 48:
```suggestion
	// Identifies the belonging network namespace of the event
	netnsid uint64
```
would that be a better variable name with this PR?

pkg/gadgets/trace/network/tracer/tracer.go
```diff
@@ -256,21 +249,25 @@ func (t *Tracer) Pop() ([]*types.Event, error) {
...
 	unix.Close(l.sockFd)
 	l.networkGraphObjects.Close()
-	delete(t.attachments, key)
+	delete(t.attachments, netns)
 }
 
-func (t *Tracer) Detach(key string) error {
-	if l, ok := t.attachments[key]; ok {
+func (t *Tracer) Detach(pid uint32) error {
+	netns, err := containerutils.GetNetNs(int(pid))
```
alban, line 259: How does that work? When Detach() is called, the container has probably terminated, so the pid does not exist and `/proc/$pid/ns/net` does not exist.

blanquicet: Oops. Sorry, I forgot to update this. I will embed the [Tracer](https://github.com/inspektor-gadget/inspektor-gadget/blob/2727379283b37d566b3932bb9f39804329f7640b/pkg/gadgets/internal/networktracer/tracer.go#L99) to reuse the `Detach` and `Close` but reimplementing the `Attach` method. WDYT?

blanquicet: Actually, I will not use the `networktracer.Tracer` in this PR as it needs a rework to work also for the network-graph tracer. For this PR, I will just apply the same logic inside the network-graph tracer, but for sure, we could make it work also with the `networktracer.Tracer` in a future PR. WDYT?

pkg/gadgets/trace/network/tracer/tracer.go
```diff
@@ -256,27 +247,27 @@ func (t *Tracer) Pop() ([]*types.Event, error) {
...
 	}
+	return fmt.Errorf("pid %d is not attached", pid)
 }
 
 func (t *Tracer) Close() {
 	for key, l := range t.attachments {
-		t.releaseLink(key, l)
+		t.releaseAttachment(key, l)
 	}
-	t.networkGraphMapObjects.Close()
```
mauriciovasquezbernal, line 281: Why to remove this line? This is created in `NewTracer` and won't be released by `releaseAttachment`.

mauriciovasquezbernal (commented):
Only one blocking comment on `t.networkGraphMapObjects.Close()`. Otherwise LGTM!

pkg/gadgets/trace/network/tracer/tracer.go
mauriciovasquezbernal: I agree, let's move forward with this one and then we can generalize `networktracer.Tracer` to make it work on this case too.

pkg/gadgets/trace/network/tracer/tracer.go
blanquicet: Oops, it wasn't on purpose. Re-added.

pkg/gadget-collection/gadgets/trace/network/gadget.go
```diff
@@ -245,16 +173,6 @@ func (t *Trace) run(trace *gadgetv1alpha1.Trace) {
...
-				return
-			}
-			err := t.tracer.Detach(key)
-			if err != nil {
-				t.publishMessage(trace, eventtypes.ERR, key, fmt.Sprintf("failed to detach tracer: %s", err))
-				return
-			}
-			t.publishMessage(trace, eventtypes.DEBUG, key, "tracer detached")
 		case <-ticker.C:
 			if !t.update(trace, traceBeforePatch) {
```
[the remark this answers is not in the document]
blanquicet: I'm not sure if it is a leftover from the initial proposal, which also put the output in the `trace.Status.Output`.

mauriciovasquezbernal (approved):
LGTM, nice work!

blanquicet:
There is a problem with the network-graph gadget, and it seems some events are missing. I'm investigating it. Check local-gadget unit tests output.

blanquicet:
> There is a problem with the network-graph gadget, and it seems some events are missing. I'm investigating it. Check local-gadget unit tests output.

After a long investigation, I found the issue. The problem was that the gadget was not able to enrich the events generated less than 1 second ([our predefined timeslot](https://github.com/inspektor-gadget/inspektor-gadget/blob/5a3852a171e8c7aedc6bab468bbc05f014a73eb3/pkg/gadget-collection/gadgets/trace/network/gadget.go#L241)) before the termination of the container that triggered those events. It was because we were calling the `EnrichByNetNs` on the detach, and at this time, the container was already removed from the container-collection. Therefore, I had to modify the timing for the notification (see [this commit](https://github.com/inspektor-gadget/inspektor-gadget/pull/1182/commits/b32559899046a5fc88f8e162b53a8d6ee8f89799)) and add a cache (as suggested by @alban) to store the already enriched events from the terminated containers, until the next time `Pop` is called. 

The issue was reproducible by running the gadget test of the `local-gadget-manager`:
```bash
$ go test -c ./pkg/local-gadget-manager -tags withebpf
$ sudo ./local-gadget-manager.test -test.v -root-test -test.run TestNetworkGraph
```
Or by running the network-graph gadget with the local-gadget and using this test container:

```bash
$ docker run -it --rm --name test-container nginx /bin/sh -c "nginx && curl 127.0.0.1"
```

Now, something else is failing in the local-gadget integration tests only for docker. I will investigate it tomorrow, but please take a look to my proposal for the previously described issue.

pkg/container-collection/container-collection.go
```diff
@@ -119,16 +119,23 @@ func (cc *ContainerCollection) GetContainer(id string) *Container {
...
 	if cc.pubsub != nil {
 		cc.pubsub.Publish(EventTypeRemoveContainer, v.(*Container))
 	}
+
+	// Remove the container from the collection after publishing the event as
+	// subscribers might need to use the different collection's lookups during
+	// the notification handler, and they expect the container to still be
+	// present.
+	cc.containers.Delete(id)
 }
```
mauriciovasquezbernal, line 139: I think it's fine to assume that `RemoveContainer()` and `AddContainer()` won't be called with the same container ID concurrently, otherwise a race can happen:

```
Goroutine A                    | Goroutine B 
                               |
RemoveContainer("foo")         | AddContainer("foo")
                               | ...
                               |   cc.containers.LoadOrStore("foo", container)
                               |   ^^^ -> loads existing container 
...                            |
  cc.containers.Delete("foo")  | 
  ^^^ -> removes container     |
  
  
...
GetContainer("foo") -> fails as container is not present
```

mauriciovasquezbernal (commented):
I think changing the order of operations in `RemoveContainer()` is fine, there is a low risk of a race condition there but as mentioned in the other comment I won't worry about it. 

Regarding the cache, there is a [similar issue](https://github.com/inspektor-gadget/inspektor-gadget/issues/1178) for other gadgets. I was considering keeping the container available for enrichment a bit after it's removed, so gadgets are able to enrich events even after the container is removed. However, we have to pay attention not to reintroduce https://github.com/inspektor-gadget/inspektor-gadget/issues/1001. Anyway, that looks like work for another PR in the future and it's not directly related to this PR.

pkg/container-utils/docker/docker.go
```diff
@@ -81,7 +81,18 @@ func listContainers(c *DockerClient, filter *dockerfilters.Args) ([]dockertypes.
...
+		if c.Labels["io.kubernetes.docker.type"] == "podsandbox" {
+			// Temporarily ignoring pod sandbox containers. Otherwise, they will
+			// be considered as containers and EnrichByNetNs will incorrectly
+			// think that they are using a given network namespace. See issue
+			// https://github.com/inspektor-gadget/inspektor-gadget/issues/1095.
+			continue
+		}
+		noPauseContainers = append(noPauseContainers, c)
+	}
+	return noPauseContainers, nil
```
blanquicet, line 95: The proper solution for this should be:

- Always include pause containers: Do not ignore them here in Docker, and include them with Containard and CRI-O as described in https://github.com/inspektor-gadget/inspektor-gadget/issues/1095. 
- Add a boolean in the `Container` struct that says whether the container is a sandbox container.
- `EnrichByNetNs` should consider whether the container is a sandbox or not to decide whether to include the container information.

pkg/gadget-collection/gadgets/trace/network/gadget.go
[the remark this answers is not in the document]
blanquicet: I removed it.

alban (approved):
One typo in the commit message:
> with pop apprach

And it needs a rebase.

LGTM after that!

PR #1183, opened 2022-12-07
ci: Do not push RC to krew.

eiffel-fl:
The goal of RC is to test the pipeline before pushing the effective release. As pushing to krew is at the end of the pipeline, the RC would already have tested everything.
This is a good idea to not push RC to krew to avoid noise for krew maintainers as PR containing RC will not be merged automatically.


mauriciovasquezbernal (approved):
LGTM!

Issue #1164, opened 2022-12-01
[RFE] Tracing network connections and aggregated statistics


alban:
## Current situation

We have the following gadgets that can list network connections:
* [tcpdump](https://github.com/inspektor-gadget/inspektor-gadget/pull/1097) (not merged yet). This can trace all packets but too intensive to trace everything all the time.
* [tcp trace gadget](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/guides/trace/tcp.md): only tcp, and does not aggregate events.
* [network gadget](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/guides/trace/network.md). This aggregates connections but does not gives statistics and some data is missing (source IP, process names).

## Impact

There are no ideal gadgets to analyse all network connections (udp + tcp) with enough metadata (process names, packet counts, byte counts, source IP) and with low performance impact to enable all the time.

## Ideal future situation

It would be good to have something intermediate. It would need to have the following metadata:
- process: uid, pid, ppid, comm, exe, cwd
- network protocol
- local ip
- remote ip
- remote port
- remote pod / service
- container metadata (container id, pod name & namespace)
- container image (see also #1310)
- packet count, byte count

## Implementation options

Network packet capture would not give the process metadata directly.

mauriciovasquezbernal:
We need to scope this better to understand the purpose of this issue.

mauriciovasquezbernal:
@alban I' wondering what format should this gadget present the information? I see that the `top tcp` is very similar to what you described here:

```bash
CONTAINER                               PID         COMM             IP LOCAL                 REMOTE                SENT                  RECV                
mycontainer                             90482       wget             4  172.17.0.3:47282      159.89.192.182:80     0B                    27.65MiB 
```

PR #1193, opened 2022-12-12
[RFC] sni: use socketenricher

alban:
# [RFC] sni: use socketenricher

In #1151, I implemented SocketEnricher so that custom networking bpf programs written by users for the bring-your-own-bpf gadget could print process metadata like `pid` and `comm`.

This PR is showing how it can be useful for existing networking gadgets like sni.

The SocketEnricher could be a step for #1164.

This uses bpf extensions (bpf functions replaced dynamically) to implement helper functions such as:
```
__u64 mount_ns_id = gadget_skb_get_mntns(skb);
__u64 pid_tgid = gadget_skb_get_pid_tgid(skb);
```

## How to use

```
$ go run -exec sudo ./cmd/local-gadget/... trace sni -r docker
CONTAINER   PID        TID        COMM     NAME
test        70022      70022      curl     wikipedia.org
```
```
$ sudo ./local-gadget trace sni -r docker
CONTAINER   PID        TID        COMM     NAME                                        
test        93945      93945      curl     wikipedia.org
```

The SocketEnricher currently has the following limitations:
- does not work yet with UDP, so no demo for the dns gadget yet
- sockets created before the tracer are not visible

## Discussion

Do you think this is a good approach? Should I prepare a PR for this separately from the bring-your-own-bpf gadget?

cc @flyth


alban:
I'll file a new PR for this shortly without pulling the bring-your-own-bpf gadget.

PR #1194, opened 2022-12-12
Add a linter docker image

eiffel-fl:
Hi.

This PR adds a linter docker image to avoid the linter complaining because we lack `libseccomp-dev`.

Best regards.


burak-ok (approved):
LGTM. Liking how every `make` command gets more stable :)

PR #1195, opened 2022-12-13
integration-tests: Disable verbose mode for local-gadget

mqasimsarfraz:
Since the test-infra is working fine and we don't have any false positive/negative we want to go ahead and disable verbose mode for tests. Also, add tests to release job 'needs'.


blanquicet (approved):
LGTM!

Issue #1192, opened 2022-12-09
Regex in tools/image-tag triggers a warning in grep 3.8+


invidian:
This change in grep: https://git.savannah.gnu.org/cgit/grep.git/commit/?h=v3.8&id=15466174352557a5f9993aba8afb0cf9397b5b2c started producing a warning for this regex `'^(?:[^?][^ ]|[^ ][^?])\s'` in `tools/image-tag` script, which uses `-qE` grep arguments.

```
$ make local-gadget-linux-amd64
grep: warning: ? at start of expression
```

Either regex should be changed or arguments should be changed to `-qP`, which suppresses it, I think because it changes the regex engine used.

See https://kubernetes.slack.com/archives/CSYL75LF6/p1670442533217329 for more context.

CC @eiffel-fl

PR #1197, opened 2022-12-13
tools: Fix git status regular expression.

eiffel-fl:
In tools/image-tag, the output of git status --porcelain is analyzed to determine if we should append "-WIP" to the image tag. Sadly, grep 3.8 prints a warning message when running this regular expression [1, 2].

This commit simplifies and fixes the existing regex. Indeed, the previous regex was wrong as it tries to use non capturing group ('?:' operator), which are not available in grep with the extended regex engine. As a consequence, the colon was treated as a character but this character is never printed by git status --porcelain.
Then, the left part of the or operator (i.e. ?:[^?][^ ]) was always wrong and only the right one could be right ([^ ][^?]).
So, this commit simplifies the regex by only applying the previous right part on git status --porcelain output.

[1] https://git.savannah.gnu.org/cgit/grep.git/commit/?h=v3.8&id=154661743525 [2] https://github.com/inspektor-gadget/inspektor-gadget/issues/1192#issue-1487457701

### DIFF
--- a/tools/image-tag
+++ b/tools/image-tag
@@ -12,13 +12,12 @@ if [ -n "$IMAGE_TAG" ]; then
   exit 0
 fi
 
-WORKING_SUFFIX=$(if git status --porcelain | grep -qE '^(?:[^?][^ ]|[^ ][^?])\s'; then echo "-WIP"; else echo ""; fi)
 BRANCH_PREFIX=$(git rev-parse --abbrev-ref HEAD)
 if [ "$1" = "branch" ] ; then
   if [ "$BRANCH_PREFIX" = "main" ] ; then
     BRANCH_PREFIX="latest"
   fi
-  echo "${BRANCH_PREFIX//\//-}$WORKING_SUFFIX"
+  echo "${BRANCH_PREFIX//\//-}"
 else
-  echo "${BRANCH_PREFIX//\//-}-$(git rev-parse --short HEAD)$WORKING_SUFFIX"
+  echo "${BRANCH_PREFIX//\//-}-$(git rev-parse --short HEAD)"
 fi


tools/image-tag
```diff
@@ -12,7 +12,7 @@ if [ -n "$IMAGE_TAG" ]; then
   exit 0
 fi
 
-WORKING_SUFFIX=$(if git status --porcelain | grep -qE '^(?:[^?][^ ]|[^ ][^?])\s'; then echo "-WIP"; else echo ""; fi)
+WORKING_SUFFIX=$(if git status --porcelain | grep -qE '^[^ ][^?]\s'; then echo "-WIP"; else echo ""; fi)
```
burak-ok, line 15: The current regex won't match on my `git status --procelain` output.
```bash
$ git status --porcelain
 M crd.mk
 D tests.mk
?? new_file
```
So we still need some kind of grouping with or. The old regex didn't match on ?? too. What was the original intention of this regex?

burak-ok (approved):

tools/image-tag
alban: I think the goal was to add "-WIP" either when
1. there are local changes (`git diff` shows something), or
2. if there is changes in the index (`git diff --cached` shows something).

Your patch only captures the 2nd case. What about the following instead?

```
if git status --porcelain | grep -qE '^([^?][^ ]|[^ ][^?])\s'; then echo "-WIP"; else echo ""; fi
```

alban: In that case, what about removing `$WORKING_SUFFIX` completely? It is weird to me to add -WIP only for changes in the index.

alban (approved):

PR #1202, opened 2022-12-14
cmd: Group supported hooks into an array.

eiffel-fl:
Hi.

In this PR, I grouped the supported hooks into an array.
The goal is to avoid repetition and so ease maintaining.

Best regards.


mauriciovasquezbernal (approved):
LGTM, thanks!

burak-ok (approved):
Nice :)

PR #1205, opened 2022-12-15
Add timestamps with nano-precision to events from gadgets

alban:
# Add timestamps with nano-precision to events from gadgets

I get the timestamp from the bpf code to get the date as precise as possible.

## How to use

The timestamp column is hidden by default. Use one of those options to see it:

```
$ sudo ./local-gadget trace exec -r docker -o json
{"timestamp":"2022-12-15T16:51:41.586709715+01:00","namespace":"default","pod":"test","container":"test","type":"normal","pid":1871800,"ppid":1866717,"comm":"cat","args":["/usr/bin/cat","/dev/null"],"mountnsid":4026532888}
```

```
$ sudo ./local-gadget trace exec -r docker -o custom-columns=timestamp,comm
TIMESTAMP                              COMM
2022-12-15T16:55:32.997502865+01:00    cat
```

## Testing done

The timestamp seems to give the correct result in my test on the trace exec gadget.

## TODO

- [x] Add timestamps to all gadgets
- [x] Sort the timestamp column: it might not work because [the doc says](https://pkg.go.dev/time) "The RFC3339Nano format removes trailing zeros from the seconds field and thus may not sort correctly once formatted."


pkg/types/types.go
```diff
@@ -53,6 +54,9 @@ func Init(nodeName string) {
 }
 
 type CommonData struct {
+	// Timestamp
+	Timestamp string `json:"timestamp,omitempty" column:"timestamp,template:timestamp"`
```
mauriciovasquezbernal, line 58: I'm wondering about the format to use here. Isn't it better to use an integer (uint64) and let the consumers do the formatting?

alban: I pushed a second patch using an integer.

@flyth But I am not sure how to use `MustSetExtractor` in a generic way to display the date without duplicating the code for every gadget. Is there a way to define an extractor in the template?

flyth: A pretty generic solution could be to make the columns lib aware of the `fmt.Stringer` interface:

```golang
type Stringer interface {
	String() string
}
```

And then let the Timestamp field have a type Int64 that implements that interface:
```golang
func (v Int64) String() string {
	return time.Format(...)
}
```

[Here's](https://github.com/inspektor-gadget/inspektor-gadget/commit/c9c3c705befbdc62d8e2c6c946eb78e4244cf84b) an implementation of that for the lib - if you think this is a good way (although not 100% optimized, yet), let me know and I'll create a PR.

In that case you would just have to add `,stringer` to the column definition.

alban: @flyth That sounds good. But I didn't understand in which package would the function `func (v Int64) String() string` be implemented.

flyth: I'd suggest in [pkg/types/types.go](https://github.com/inspektor-gadget/inspektor-gadget/pull/1205/files#diff-8e49337a1417a39a62ebf427e84fb2faa566c96e37ae627ac484c05dbe65c483), something like this:

```golang
type Time int64

func (t Time) String() string {
	return time.Format(...)
}

type CommonData struct {
 	// Timestamp in nanoseconds since January 1, 1970 UTC. A int64 is big
 	// enough to represent time between the year 1678 and 2262.
 	Timestamp Time `json:"timestamp,omitempty" column:"timestamp,template:timestamp,stringer"`
}
```

mauriciovasquezbernal (commented):
It's looking good. Some comments about how to reduce code duplication in the future but those are not relevant for this PR. 

I'll provide a full review once it's marked as ready for review.

pkg/types/types.go
alban: I updated this PR with that approach. I picked up c9c3c705befb and added another patch to fix a panic in the column package.

pkg/gadgets/traceloop/tracer/tracer.go
```diff
@@ -432,8 +434,8 @@ func (t *Tracer) Read(containerID string) ([]*types.Event, error) {
 	// but they will be incomplete.
 	// One possible reason would be that the buffer is full and so it only remains
 	// some exit events and not the corresponding enter/
-	for _, enterTimestampEvents := range syscallEnterEventsMap {
-		for enterTimestamp, enterEvent := range enterTimestampEvents {
+	for enterTimestamp, enterTimestampEvents := range syscallEnterEventsMap {
+		for _, enterEvent := range enterTimestampEvents {
```
alban, line 438: FYI @eiffel-fl, I think the `enterTimestamp` variable is declared in the wrong `for` loop.

alban: Right, I am just focusing on the timestamps in this PR.

pkg/gadgets/helpers.go
```diff
@@ -92,3 +94,33 @@ func IPStringFromBytes(ipBytes [16]byte, ipType int) string {
...
+
+var timeDiff time.Duration
+
+func init() {
+	var t unix.Timespec
+	err := unix.ClockGettime(unix.CLOCK_BOOTTIME, &t)
+	if err != nil {
+		panic(err)
+	}
+	timeDiff = time.Duration(time.Now().UnixNano() - t.Sec*1000*1000*1000 - t.Nsec)
```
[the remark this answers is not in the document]
alban: The variable `t` is defined a few lines above, it is not one second and one nanosecond.

I'm calculating the difference between two clocks (see [man clock_gettime](https://www.man7.org/linux/man-pages/man3/clock_gettime.3.html)):
- epoch_time (wall clock provided by time.Now())
- boot_time (provided by CLOCK_BOOTTIME)

So I know how to offset the time I get from BPF.

See https://github.com/xdp-project/xdp-tutorial/issues/204#issuecomment-819419800

alban:
@eiffel-fl Thanks! I updated the PR.

integration/inspektor-gadget/integration_test.go
```diff
@@ -855,7 +865,7 @@ func TestNetworkpolicy(t *testing.T) {
 					sleep 10
 					kill $!
 					head networktrace-client.log | sort | uniq`, nsClient),
-			ExpectedRegexp: fmt.Sprintf(`{"node":".*","namespace":"%s","pod":"test-pod","type":"normal","pktType":"OUTGOING","proto":"tcp","port":9090,"podHostIP":".*","podIP":".*","podLabels":{"run":"test-pod"},"remoteKind":"svc","remoteAddr":".*","remoteName":"test-pod","remoteNamespace":"%s","remoteLabels":{"run":"test-pod"}}`, nsClient, nsServer),
+			ExpectedRegexp: fmt.Sprintf(`{.*"node":".*","namespace":"%s","pod":"test-pod","type":"normal","pktType":"OUTGOING","proto":"tcp","port":9090,"podHostIP":".*","podIP":".*","podLabels":{"run":"test-pod"},"remoteKind":"svc","remoteAddr":".*","remoteName":"test-pod","remoteNamespace":"%s","remoteLabels":{"run":"test-pod"}}`, nsClient, nsServer),
```
burak-ok, line 868: Not because of your changes:
Why do we still have a Regexp for checking the output of a gadget?

burak-ok (approved):
Looks good to me.
Didn't find anything besides knowledge to learn :)

pkg/columns/columninfo.go
```diff
@@ -223,6 +223,20 @@ func (ci *Column[T]) parseTagInfo(tagInfo []string) error {
 				return fmt.Errorf("no template specified for field %q", ci.Name)
 			}
 			ci.template = params[1]
+		case "stringer":
+			if ci.Extractor != nil {
+				break
```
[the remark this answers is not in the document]
alban: The `break` applies to the inner `switch` and not to the outer `for` loop:
https://go.dev/play/p/Leh178DLT0C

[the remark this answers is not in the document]
flyth: @eiffel-fl You're right, sorry, mixed up the loops ;)

integration/inspektor-gadget/integration_test.go
alban: I think this test was written before we started switching our tests to check structs instead of json strings.

alban:
Rebased to fix conflict with main

mauriciovasquezbernal (commented):
`ktime_get_boot_ns` was added in 5.7 (https://github.com/torvalds/linux/commit/71d19214776e61b33da48f7c1b46e522c7f78221). It seems that it's a too new requirement to have for all gadgets :(.  Can't we use `ktime_get_ns()`?

alban:
ktime_get_ns() is more difficult to use because it uses the monotonic clock and that doesn't account for the time when the computer is suspended. This means we can't cache the difference between the wall clock and the monotonic clock in a init() function.

We could refresh the cache regularly, but the function to calculate the difference is not instantaneous so we take the risk of getting events with timestamps showing causal paradoxes.

Here is an attempt to make the calculation fast with runtime.LockOSThread():
https://github.com/keyval-dev/opentelemetry-go-instrumentation/blob/c4799f7d5fbedcd7533daf0e4103449c6ed156e5/pkg/opentelemetry/controller.go#L144

There was an attempt to unify the monotonic and boot clocks but it was reverted:
https://www.spinics.net/lists/linux-tip-commits/msg43709.html

mauriciovasquezbernal:
There is another possibility and it's something I always wanted to do but never had a real case for it: modify the ebpf program instructions before loading into the kernel.

The idea is to check if the kernel supports the `ktime_get_boot_ns()` helper by looking at the BTF info, then, if it's not supported to remove the helper call from the eBPF program before loading it. We can replace the result of the call by a constant like 0, then the userspace implementation can act and implement a fallback mechanism.  I implemented a PoC of this in https://github.com/inspektor-gadget/inspektor-gadget/commit/e6df59cdab5bdf4323c91e4a64964eb337e4c4fc, I tested and it seems to work fine. The fallback mechanism in userspace to fill the timestamp with something else is missing.

pkg/types/types.go
```diff
@@ -52,7 +54,17 @@ func Init(nodeName string) {
...
+type Time int64
+
+func (t Time) String() string {
+	return time.Unix(0, int64(t)).Format(time.RFC3339Nano)
+}
+
 type CommonData struct {
+	// Timestamp in nanoseconds since January 1, 1970 UTC. A int64 is big
+	// enough to represent time between the year 1678 and 2262.
+	Timestamp Time `json:"timestamp,omitempty" column:"timestamp,template:timestamp,stringer"`
```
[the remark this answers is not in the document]
alban: Fixed

alban:
@mauriciovasquezbernal I used your patch to support kernels without ktime_get_boot_ns.

pkg/gadgets/helpers.go
```diff
@@ -100,3 +109,88 @@ func IPStringFromBytes(ipBytes [16]byte, ipType int) string {
...
+	btfSpec, err := btf.LoadKernelSpec()
+	if err != nil {
+		return false
+	}
+
+	enum := btf.Enum{}
+
+	err = btfSpec.TypeByName("bpf_func_id", &enum)
+	if err != nil {
+		return false
```
alban, line 161: Here I get the error:
```
*btf.Enum does not satisfy Type interface
```

It works when `TypeByName()` has a double pointer (`**btf.Enum`). Fixing...

alban:
> Is there any limitation which impedes you from using bpf_ktime_get_ns() inside eBPF code and then
> WallTimeFromBootTime() in userspace to add the difference between "time" and "boot time"?

I mentioned the issue here: https://github.com/inspektor-gadget/inspektor-gadget/pull/1205#issuecomment-1360136575. Basically the difference between the 2 times can change over time (e.g. with computer suspend)

pkg/types/types.go
```diff
@@ -92,6 +100,10 @@ const (
 type Event struct {
 	CommonData
 
+	// Timestamp in nanoseconds since January 1, 1970 UTC. A int64 is big
```
mauriciovasquezbernal, line 103: nit -> "A int64" -> "An int64"

mauriciovasquezbernal (approved):
I reviewed everything but traceloop as I'm not totally familiar with this. What about moving the traceloop changes to a different PR as they don't see to fully match what is changed here? 

Is the audit seccomp gadget missing this change?

mauriciovasquezbernal:
Also, feel free to squash the commit I suggested to avoid having a commit that is changed in a following one in the same PR.

alban:
Thanks for the review! I am removing the changes in traceloop and I will merge this PR when the CI is green without the traceloop changes.

Remaining TODOs for future PRs:
- [ ] the traceloop timestamp and fixes
- [ ] the audit seccomp gadget
- [ ] the TODO for the sni gadget

alban:
Actually, it's more difficult than I thought to remove the traceloop changes from this PR because traceloop events already have a `Timestamp` field with a different semantic, so I have to rename that field anyway.

@eiffel-fl Could you review the traceloop changes (the last 2 commits)? Thank you :)

pkg/gadgets/traceloop/tracer/bpf/traceloop.bpf.c
```diff
@@ -152,7 +152,7 @@ int ig_traceloop_e(struct bpf_raw_tracepoint_args *ctx)
 	 * https://github.com/iovisor/bcc/issues/2623#issuecomment-560214481
 	 */
 	struct syscall_event_t sc = {};
-	u64 ts = bpf_ktime_get_ns();
+	u64 ts = bpf_ktime_get_boot_ns();
```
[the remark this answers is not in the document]
alban: Makes sense. I moved it.

pkg/gadgets/traceloop/tracer/tracer.go
```diff
@@ -478,7 +492,7 @@ func (t *Tracer) Read(containerID string) ([]*types.Event, error) {
 			incompleteExitEvent := &types.Event{
 				Event: eventtypes.Event{
 					Type:      eventtypes.NORMAL,
-					Timestamp: gadgets.WallTimeFromBootTime(exitTimestamp),
+					Timestamp: wallTimeFromBootTime(exitEvent.bootTimestamp),
```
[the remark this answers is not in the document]
alban: > Let me summarize this commit so I am sure to understand it:
> 1. You are using the monotonic clock to link events between themselves (mainly an exit or continued event to an enter one).
> 2. The boot clock is used for the `BaseEvent` timestamp, if this code is run on kernel older than `5.7`, there will be no `BaseEvent` timestamp.

Correct.

> Nonetheless, if run on kernel older than 5.7, the events will not be sorted on line 516.
> So, should this sorting be done on the monotonic clock?

Well spotted, I missed it. I pushed a new version to fix this.

Issue #1196, opened 2022-12-13
[RFE] gadgets/dns: Enhance gadget to provide more information about failed requests and latency


alban:
This is a follow-up to #1000.

## Current situation

The DNS trace gadget provides information about DNS requests. It provides the following:

- node, namespace, pod: pod where the request is coming from / to
- pktType: direction of the request (ONGOING / INCOMING)
- qtype: type of request
- name: host of the request
- nameserver: DNS server where the request is going to (added in #1071)
- id: to match queries & responses  (added in #1071)

There is more information we could add to this:
- [x] timestamp: when the packet was observed
  - #1206
  - #1205
- [x] latency: time difference between a request and a response by matching the id
- [x] failed requests: we can see responses but we don't know if there is an error, nxdomain or timeout
  - #1217 
- [x] #1415 

## Impact

This makes the dns gadget less useful to debug.

## Ideal future situation

The dns gadget displays the additional information listed above.

alban:
>  latency: time difference between a request and a response by matching the id

I am not sure if matching the dns id field between the dns request and response should be done in ebpf or in the pkg/gadgets/trace/dns/tracer Golang package but I think we should avoid doing it in the CLI to ensure it has a consistent result regardless of the client (kubectl-gadget, local-gadget, Headlamp plugin...).

The dns ids should be forgotten after a timeout or a maximum to ensure that the memory used to track it cannot grow without limits.
* If the timeout is implemented in ebpf, we could use the [bpf timers](https://github.com/torvalds/linux/commit/b00628b1c7d595ae5b544e059c27b1f5828314b4) but this requires Linux 5.15. 
* Alternatively, we could have a ticker in Golang to remove old entries. This could be done in one syscall with a bpf iterator (Linux 5.8).
* This could also be done with a LRU hash (Linux 4.10). Maybe the easiest and most compatible with older Linux versions?

alban:
@wedaly would like to work on this (see https://github.com/inspektor-gadget/inspektor-gadget/pull/1206#issuecomment-1357781484). It seems I can't update the "assignees" field accordingly, so I'm posting a comment here instead.

wedaly:
While working on "failed requests", I noticed that the response code flag wasn't being set correctly. I opened a PR that I think should address that issue: https://github.com/inspektor-gadget/inspektor-gadget/pull/1212

mauriciovasquezbernal:
[...]

PR #1206, opened 2022-12-16
trace/dns: add timestamp field

wedaly:
# trace/dns: add timestamp field

Add a timestamp field to DNS trace events so we can see when the DNS packet was observed. The timestamp comes from bpf_ktime_get_ns(), which reports the time in nanoseconds since system boot, excluding suspend time (CLOCK_MONOTONIC).

Related issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/1196

## How to use

Follow the steps in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/guides/trace/dns.md

## Testing done
Tested in minikube by running `./kubectl-gadget trace dns -n demo -o custom-columns=timestamp,qtype,type,name`

image


pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -35,6 +35,7 @@ type Event struct {
 	PktType    string     `json:"pktType,omitempty" column:"type,minWidth:7,maxWidth:9"`
 	QType      string     `json:"qtype,omitempty" column:"qtype,minWidth:5,maxWidth:10"`
 	DNSName    string     `json:"name,omitempty" column:"name,width:30"`
+	Timestamp  uint64     `json:"timestamp,omitempty" column:"timestamp"`
```
wedaly, line 38: I wasn't sure what other properties to set for `column` here. I see that [traceloop sets `width:7,fixed,hide`](https://github.com/inspektor-gadget/inspektor-gadget/blob/e3546b642cefcbb5df31b11e3592219c0118facd/pkg/gadgets/traceloop/types/types.go#L34), but that truncated the timestamps. Thoughts?

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -105,6 +105,7 @@ int ig_trace_dns(struct __sk_buff *skb)
 		return 0;
 
 	struct event_t event = {0,};
+	event.timestamp = bpf_ktime_get_ns();
```
wedaly, line 108: Is this the right helper to use? Wasn't sure if `bpf_ktime_get_coarse_ns` or `bpf_ktime_get_boot_ns` might be more appropriate.

wedaly:
looks like this overlaps with https://github.com/inspektor-gadget/inspektor-gadget/pull/1205

alban:
Hi @wedaly, sorry for the duplicate work.
#1205 does not implement everything, for example it does not show the duration between a dns request and the dns response. Would you like to work on that once #1205 is merged? I am going to post some thoughts on #1196.

wedaly:
> Hi @wedaly, sorry for the duplicate work. #1205 does not implement everything, for example it does not show the duration between a dns request and the dns response. Would you like to work on that once #1205 is merged? I am going to post some thoughts on #1196.

No problem! Yes, I'd like to work on #1196. Thanks!

PR #1207, opened 2022-12-16
gadgets/snapshot/process: Rename columns to match user space

mauriciovasquezbernal:
The naming around process / thread / task in kernel and user space
    is different and it's very confusing. In the kernel, each thread has an
    ID, called PID, and also has a TGID (Task Group ID) that is the PID of
    the first thread in the group. From user space we have PID
    (Process Identifier), that matches TGID on the kernel, and TID
    (Thread ID), that matches PID on the kernel.
    
    In summary[0]:
    
    User  | Kernel
    ------+-------
    PID   | TGID
    TID   | PID
    
    Then, TGID is a kernel concept and we shouldn't show this to the user,
    hence this commit uses PID and TID on the snapshot process gadget.

### DIFF
--- a/cmd/common/snapshot/process.go
+++ b/cmd/common/snapshot/process.go
@@ -63,7 +63,7 @@ func (p *ProcessParser) SortEvents(allProcesses *[]*types.Event) {
 	if !p.flags.showThreads {
 		allProcessesTrimmed := []*types.Event{}
 		for _, i := range *allProcesses {
-			if i.Tgid == i.Pid {
+			if i.Tid == i.Pid {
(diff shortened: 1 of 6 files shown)


blanquicet (approved):
LGTM!

alban (approved):
LGTM, thanks!

@eiffel-fl where is the file `integration/local-gadget/snapshot_process_test.go` coming from?

burak-ok (approved):
LGTM.

@alban : @eiffel-fl merged his PR 5 hours ago and it contains the new file which he referred to

PR #1208, opened 2022-12-17
gadgets/snapshot/process: Add support for older kernels

mauriciovasquezbernal:
Add a fallback mechanism that works on older kernels too. This reads
    the information about the processes from /proc/, even if this is slower
    than the eBPF-based implementation, it can be useful in cases that
    eBPF-iterators are not available.
    
    The following benchmark shows that the procfs-based implementation is
    ~30% slower than the ebpf-based one:
    
    ```
    $ go test -exec sudo -bench=. -benchtime=100x . -v
    [sudo] password for mvb:
    goos: linux
    goarch: amd64
    pkg: github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/tracer
    cpu: AMD Ryzen 7 3700X 8-Core Processor
    BenchmarkEBPFTracer
    BenchmarkEBPFTracer-16               100          21743119 ns/op
    BenchmarkProcfsTracer
    BenchmarkProcfsTracer-16             100          31169492 ns/op
    PASS
    ok      github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/tracer        8.340s
    ```

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/843

### DIFF
--- a/pkg/gadgets/snapshot/process/tracer/tracer.go
+++ b/pkg/gadgets/snapshot/process/tracer/tracer.go
@@ -16,12 +16,18 @@ package tracer
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
 
+	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	processcollectortypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/types"
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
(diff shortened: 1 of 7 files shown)


pkg/gadgets/snapshot/process/tracer/tracer.go
```diff
@@ -120,3 +143,88 @@ func RunCollector(config *Config, enricher gadgets.DataEnricher) ([]*processcoll
...
+		}
+
+		events = append(events, &event)
+	}
+
+	return events, nil
+}
+
+func runProcfsCollector(config *Config, enricher gadgets.DataEnricher) ([]*processcollectortypes.Event, error) {
+	items, err := os.ReadDir("/proc/")
```
alban, line 203: I wonder if we should use this:
```
filepath.Join(os.Getenv("HOST_ROOT"), "/proc/")
```

The packages `runcfanotify` and `bpfstats` started to use `$HOST_ROOT`.

pkg/gadgets/snapshot/process/tracer/tracer.go
```diff
@@ -120,3 +143,88 @@ func RunCollector(config *Config, enricher gadgets.DataEnricher) ([]*processcoll
...
+
+		pid64, err := strconv.ParseUint(item.Name(), 10, 32)
+		if err != nil {
+			continue
+		}
+		pid := int(pid64)
+
+		pidEvents, err := getPidEvents(config, enricher, pid)
+		if err != nil {
+			return nil, err
```
alban, line 223: It should `continue` and silently ignore the error if it cannot read a file "/proc/%d/*": the process might have terminated between the `os.ReadDir("/proc/")` and the subsequent reads.

alban (commented):
It looks generally good. Some comments...

pkg/gadgets/snapshot/process/tracer/tracer.go
mauriciovasquezbernal: Oh right, forgot about it. Done!

pkg/gadgets/snapshot/process/tracer/tracer_test.go
```diff
@@ -0,0 +1,178 @@
...
+					MountNsID: info.MountNsID,
+				}
+			}),
+		},
+		"captures_no_events_with_no_matching_filter": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *Config {
+				// We can't use 0 as the mntnsid becuase the tracer will collect
+				// some process that have this as mntnsid. Then, we set the network
+				// namespace inode id to be sure there is not any mount namespace
+				// using that same inode.
```
alban, line 98: What about using the mntns of `os.Gettid()` (the current go routine is not the thread of the runner)?

alban (approved):
LGTM after fixing some typos

pkg/gadgets/snapshot/process/tracer/tracer_test.go
mauriciovasquezbernal: I don't think it'll work because the goal is not to capture processes at all. If we use the inode of a valid namespace will capture the processes running there.

PR #1212, opened 2022-12-19
Fix byte order in DNS flags

wedaly:
# Fix byte order in DNS flags

The two bytes in the DNS flags struct were loaded in reverse order, which caused the "RA" flag to be interpreted as the "QR" flag.

This appeared to work correctly for recursive nameservers, since the queries always set RA=0 and the nameserver responses always set RA=1.

For non-recursive nameservers, however, inspektor-gadget would interpret the RA=0 response as a query rather than a response. If the response had at least one answer, then the DNS trace BPF program would exit early without emitting an event.

~~Proposed fix in this commit swaps the two bytes in the DNS flags struct back to big endian to match the struct definition.~~ (updated based on https://github.com/inspektor-gadget/inspektor-gadget/pull/1212#issuecomment-1359189741)  

Proposed fix in this commit reorders the bitfields in the dnsflags struct to match rfc1035 in both big and little endian architectures.

## Testing done

To reproduce the issue:

1. Create a minikube cluster (I used the Docker backend running on WSL2).
2. `make install-minikube` to install IG.
3. `./kubectl-gadget trace dns`
3. `k run testpod -it --image=docker.io/library/alpine -- /bin/sh`
4. In the testpod shell, run `nslookup google.com` a few times.

Expected behavior:
```
QR NAMESERVER                   TYPE      QTYPE      NAME
Q  10.96.0.10                   OUTGOING  A          google.com.
Q  10.96.0.10                   OUTGOING  AAAA       google.com.
R  10.96.0.10                   HOST      A          google.com.
R  10.96.0.10                   HOST      AAAA       google.com.
```

Actual behavior:
```
QR NAMESERVER                   TYPE      QTYPE      NAME
Q  10.96.0.10                   OUTGOING  A          google.com.
Q  10.96.0.10                   OUTGOING  AAAA       google.com.
```

Note that running `nslookup google.com 1.1.1.1` works as expected, since Cloudflare is a recursive server. Without specifying `1.1.1.1` explicitly, though, nslookup uses coredns (svc IP 10.96.0.10), and since coredns isn't a recursive DNS server, inspektor-gadget misinterprets the responses as queries.

I confirmed that with this commit, inspektor-gadget shows the responses for google.com.


alban:
Thanks for the fix!

Your patch reorders the 2 bytes for endianness correctly, but does not reorder individual bits in the bitfield definitions. Since you are working at it, let's try to fix it completely.

I think the only way is to use #ifdefs when defining the bitfield. There is an example of doing that in [include/uapi/linux/ip.h](https://github.com/torvalds/linux/blob/6feb57c2fd7c787aecf2846a535248899e7b70fa/include/uapi/linux/ip.h#L86-L95).

In the context of eBPF, I don't know if we have access to the `__LITTLE_ENDIAN_BITFIELD` macros from the Linux example but we could use (see [bpf_endian.h](https://github.com/cilium/ebpf/blob/1e326d1d24eb2608e3e37e5e994760ad81f17f38/examples/headers/bpf_endian.h#L50)):

```
#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
...
#elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
...
#else
# error "Fix your compiler's __BYTE_ORDER__?!"
#endif
```

When I defined `union dnsflags`, I tried to reorder the bits but I got it wrong as you found out, and it does not work for both little endian and big endian:
```
	struct {
		__u8 rd :1;	// recursion desired
		__u8 tc :1;	// truncation
		__u8 aa :1;	// authoritive answer
		__u8 opcode :4;	// kind of query
		__u8 qr :1;	// 0=query; 1=response
		__u8 rcode :4;	// response code
		__u8 z :3;	// reserved
		__u8 ra :1;	// recursion available
	};
```
For reference, the correct sequence is defined in [rfc1035](https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1):
```
                                    1  1  1  1  1  1
      0  1  2  3  4  5  6  7  8  9  0  1  2  3  4  5
    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
    |QR|   Opcode  |AA|TC|RD|RA|   Z    |   RCODE   |
    +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
```

There are different ways to load bytes from a packet:
- load_half() and load_word() automatically swap bytes from network order to host order.
- bpf_skb_load_bytes() does a memcpy without swapping bytes.

Since there are byte swaps at different points (load_half, bpf_htons, bitfield definition), I got lost trying to think about it.

wedaly:
thanks @alban for the detailed explanation! I've updated this PR to use __BYTE_ORDER__  #ifdef's instead of `bpf_hotns`. Tested it again to verify that this fixes the issue (DNS responses for google.com appear in the trace events).

alban (approved):
LGTM!

I'll wait until the CI completes its job.

alban:
For a future PR:

It would be nice to test the eBPF code path in pkg/gadgets/trace/dns/tracer/tracer_test.go in the same way other gadgets do, see examples at:
```
$ ls -1 pkg/gadgets/trace/*/tracer/tracer_test.go
pkg/gadgets/trace/bind/tracer/tracer_test.go
pkg/gadgets/trace/capabilities/tracer/tracer_test.go
pkg/gadgets/trace/dns/tracer/tracer_test.go
pkg/gadgets/trace/exec/tracer/tracer_test.go
pkg/gadgets/trace/open/tracer/tracer_test.go
```
But I suspect it's more difficult to write for networking gadgets.

It might be more easy to add this test case in `integration/local-gadget/trace_dns_test.go`.

PR #1214, opened 2022-12-20
[RFC/WIP] EndpointCollection to enrich IPs with metadata in eBPF

alban:
# New: EndpointCollection

EndpointCollection defines a eBPF LPM map that gadgets can use to get an enpoint id from an IP.

The goal is to then implement endpoint enrichers, e.g. to tell that IP 192.168.0.1 is the service 'coredns'.

For the current networking gadgets (dns, sni), this could have been done in userspace. However, future networking gadgets would need to be able to aggregate metrics in eBPF instead of sending an event in the ring buffer for each network packet of interest.

## How to use (WIP)

```
$ docker run --name test --rm -ti busybox
/ # cat /etc/resolv.conf
nameserver 192.168.0.1
```

* Convert the DNS server IP to hex: c0 a8 00 01
* IPv4 has 32 bits (0x20 bits, or 20 00 00 00 in uint32 hexa little endian)
* Set the endpoint id to 7 (in uint64 hex little endian: 07 00 00 00 00 00 00 00)

```
$ sudo bpftool map update name endpoints key hex 20 00 00 00 c0 a8 00 01 value hex 07 00 00 00 00 00 00 00
```

In the container:
```
/ # wget inspektor-gadget.io
```

On the host:
```
$ go run -exec 'sudo' ./cmd/local-gadget/... trace dns -c test -r docker
CONTAINER                                  QR NAMESERVER              TYPE      QTYPE      NAME                                      ENDPOINT1    ENDPOINT2
test                                       Q  192.168.0.1             OUTGOING  A          inspektor-gadget.io.                      0            7
test                                       Q  192.168.0.1             OUTGOING  AAAA       inspektor-gadget.io.                      0            7
test                                       R  192.168.0.1             HOST      A          inspektor-gadget.io.                      7            0
test                                       R  192.168.0.1             HOST      AAAA       inspektor-gadget.io.                      7            0
```

local-gadget can identify the dns server as endpoint id 7.

## Testing done

None


alban:
We now partially have this feature with the following operators:
- kubeipresolver
- kubenameresolver

The aforementioned operators don't aggregate metrics in eBPF but this will be addressed by #1598 and a different implementation might be chosen.

PR #1223, opened 2022-12-21
gadgets/snapshot/process: Handle showthreads in server side

mauriciovasquezbernal:
snapshot process provides a ShowThreads flag. This commit moves the filtering logic to the tracer to avoid sending a bunch of data to the client that will be discarded later on.


mauriciovasquezbernal:
I moved it to the eBPF code because it was easier than I initially thought.

pkg/gadgets/snapshot/process/tracer/bpf/process-collector.bpf.c
```diff
@@ -32,6 +33,9 @@ int ig_snap_proc(struct bpf_iter__task *ctx)
 	if (task == NULL)
 		return 0;
 
+	if (!show_threads && task->tgid != task->pid)
+		return 0;
```
burak-ok, line 37: Thinking out loud / Question:

**Before** moving this to `eBPF` the following would be true, right? :
If a thread (which is not the main thread -> `tid != pid`) gets detected here it was put in the `eBPF` map. In `tracer.go` it would still generate an event for the main thread `tid == pid`.

**Now** with the `eBPF` filtering this is not the case anymore.

Am i thinking right?

mauriciovasquezbernal: Both cases will produce the same result. When `config.ShowThreads` is `false`, the following happens:

**Before**: The eBPF program generated "an entry" for each thread (task in kernel terminology), then user space filtered out the threads where `tid != pid`.

**Now**: The eBPF program only generates entries for threads who its `tid == pid` and the user space part doesn't do any filtering. 

Is that clear?

burak-ok: Haha, i should have looked where this `eBPF` program is attached to. Somehow i thought it was a kind of `kprobe` or so (Where we could maybe have not captured all relevant events or so).

Yeah, when we are iterating over every task this works as expected

Issue #424, opened 2021-12-03
CI: Avoid running BTFGen when not needed


mauriciovasquezbernal:
The current building system always runs BTFGen when creating the gadget container. 

https://github.com/kinvolk/inspektor-gadget/blob/1d48c8b25eaf653dead1ff3abd558b4dfca85e31/gadget.Dockerfile#L42-L46

We could improve it by caching the generated BTF files and only generating them again when there is a change on any bpf object or in the BCC base container image we use. 

ref https://github.com/kinvolk/inspektor-gadget/pull/387

mauriciovasquezbernal:
> BCC container image

BCC container image is not relevant anymore:

https://github.com/kinvolk/inspektor-gadget/blob/79af7fc247ec15df63c1dec3d06036fbdf11522b/tools/btfgen.sh#L21-L34

> How would you detect that there is a change in the bpf object files 

We can check if a given PR is modifying those files by looking at the git diff. However it doesn't seem too easy to implement, so perhaps we can look at another solution.

mauriciovasquezbernal:
I think it's more difficult to handle since we'll have to update that image each time an ebpf program is updated. Also, given https://github.com/kinvolk/inspektor-gadget/issues/752 we should probably include the BTFs files directly into the binary, so having an additional container image won't work on that case neither.

PR #1228, opened 2022-12-23
CI: Cache BTFGen

mauriciovasquezbernal:
BTFGen takes a lot of time to execute because it has to:
1. clone the btfhub-archive repo (some GBs)
2. uncompress all files from btfhub-archive
3. run btfgen for all btf files uncompressed above

It's only required to run this command when the bpf objects are changed or new BTF files are pushed to btfhub-archive.

This commit modifies the building system to use the Github actions to cache the generated BTF files. The changes done here are:
1. Execute btfgen outside the building container: It's hard to extract the BTF files from the container to save them in the cache.
2. Define a cache for the btf files: It uses as key the architecture, the latest commit of btfhub-archive and the hash of all bpf objects.

### Testing done

First run to create the cache -> https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/3762003013/jobs/6394277168: Takes 18m to build the container image

Another round that used the cache created above -> https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/3762175305/jobs/6394629214: ~4 minutes

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/424

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -227,6 +227,27 @@ jobs:
         registry: ${{ env.REGISTRY }}
         container-image: ${{ env.CONTAINER_REPO }}
         co-re: ${{ matrix.type == 'core' }}
+    - name: Get btfhub-archive last commmit
+      id: get-btfhub-head
+      run: |
+        echo "head=$(git ls-remote https://github.com/aquasecurity/btfhub-archive/ HEAD | cut -f1)" >> $GITHUB_OUTPUT
+      shell: bash
+    - name: Cache BTFGen
+      id: cache-btfgen
+      uses: actions/cache@v3.0.11
+      with:
+        path: hack/btfs
+        # key is composed by
+        # - arch: amd64 or arm64
+        # - lastest commit of btfhub
+        # - hash of all bpf objects
+        key: btfgen-${{ matrix.platform }}-${{ steps.get-btfhub-head.outputs.head }}-${{ hashFiles('pkg/gadgets/**/*.o') }}
+    - name: BTFGen
+      if: ${{ steps.cache-btfgen.outputs.cache-hit != 'true' }}
+      run: |
+          ./tools/getbtfhub.sh
+          make btfgen BPFTOOL=$HOME/btfhub/tools/bin/bpftool.x86_64 \
+              BTFHUB_ARCHIVE=$HOME/btfhub-archive/ OUTPUT=$GITHUB_WORKSPACE/hack/btfs -j$(nproc)
     # we are using cache-to mode=min (default) implying that only final image layers are cached, using cache
     # mode=max results in builder image layer of ~7GB because of btfhub files in a layer, which is too
     # large (gloabal limit 10GB) to work with GH caches. (TODO: if we can work with mode=max in future?)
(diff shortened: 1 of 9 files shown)


.github/workflows/inspektor-gadget.yml
```diff
@@ -227,6 +227,27 @@ jobs:
...
+    - name: Cache BTFGen
+      id: cache-btfgen
+      uses: actions/cache@v3.0.11
+      with:
+        path: hack/btfs
+        # key is composed by
+        # - arch: amd64 or arm64
+        # - lastest commit of btfhub
+        # - hash of all bpf objects
+        key: ${{ matrix.platform }}-${{ steps.get-btfhub-head.outputs.head }}-${{ hashFiles('pkg/gadgets/*/*/tracer/*.o') }}
```
mqasimsarfraz, line 244: I found couple of locations that might need to handled as well [`piditer`](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/top/ebpf/piditer) and [`container-map`](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgettracermanager/containers-map). (I wonder if we have can a check to ensure above pattern does catches all the object files. Probably can be handled in separate PR)

mqasimsarfraz: nit: Prefix the key with `bpfgen-` it is easier to validate it in [caches](https://github.com/inspektor-gadget/inspektor-gadget/actions/caches).

mqasimsarfraz (commented):
This is already looking great. I thought we could handle cloning/generation in single image layer but it just gets more complicated.

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: I changed the pattern to match all object files within pkg/gadgets. The container-map doesn't need this information, so it's not included. I also added the btfgen prefix to the cache key.

mauriciovasquezbernal:
I wanted to highly the quantity of time this PR saves. I tested it on my personal fork with the following results and the time required to build the container image is:

- **5min** for a PR that doesn't change the eBPF code: https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/3886415935/jobs/6631500238
- **18min** for a PR that changes the eBPF code: https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/3886385255/jobs/6631425075

It means that for most of the PRs will be saving ~13 min per run.

mauriciovasquezbernal:
> I thought we could handle cloning/generation in single image layer but it just gets more complicated.

TBH I don't wan to change it anymore. It seems like a "simple" PR but I expend almost a full day trying different approaches until I found something that works fine.

mqasimsarfraz:
> > I thought we could handle cloning/generation in single image layer but it just gets more complicated.
> 
> TBH I don't wan to change it anymore. It seems like a "simple" PR but I expend almost a full day trying different approaches until I found something that works fine.

Totally agree. I tried couple of approaches (when I was playing with image caching) and they all had bad edge cases but this looks simple and better. Great work!

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: It seems `piditer` files are still missing:

```
→ la -la pkg/gadgets/*/*/tracer/*.o  | wc -l
40
→ la -la pkg/gadgets/*/*/*/*.o  | wc -l
42

```

Should we change the pattern to ` pkg/gadgets/*/*/*/*.o ` ?

mqasimsarfraz (approved):
LGTM! I tested it on my fork and it seems to be working as expected!

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: It turns out I pushed the wrong branch. I tested with the `mauricio/backups/cachebtfgen` branch on https://github.com/mauriciovasquezbernal/inspektor-gadget/pull/4/files but this PR uses the `mauricio/cache-btfgen` branch. I pushed the right code for this branch.

mqasimsarfraz:
> When a cache is created by a workflow run triggered on a pull request, the cache is created for the merge ref (refs/pull/.../merge). Because of this, the cache will have a limited scope and can only be restored by re-runs of the pull request. It cannot be restored by the base branch or other pull requests targeting that base branch.

TIL: The cache will be restored from [`default` to `feature`](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) branch but not from [pull request](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) to `default` branch. Putting it out here to share it!

PR #1229, opened 2022-12-23
Fix `trace signal` printing events twice

eiffel-fl:
Hi.

In this PR, I fixed `trace signal` bug where events were printed twice.

Best regards.

### DIFF
--- a/docs/gadgets/trace/signal.md
+++ b/docs/gadgets/trace/signal.md
@@ -65,11 +65,12 @@ With the following option, you can restrict the output:
 * `--pid` only prints events where a signal is sent by the given PID.
 * `--signal` only prints events where the given signal is sent.
 * `-f/--failed-only` only prints events where signal failed to be delivered.
+* `-k/--kill-only` only prints events where signal was sent by using kill syscall.
 
-For example, this command will only print failed attempts to send `SIGKILL` by PID `42`:
+For example, this command will only print failed attempts to send `SIGKILL` by PID `42` which were initiated by calling kill syscall:
 
 ```bash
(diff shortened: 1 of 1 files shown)


pkg/gadget-collection/gadgets/trace/signal/gadget.go
```diff
@@ -139,6 +140,17 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
 		failedOnly = failedParsed
 	}
 
+	killOnly := false
+	if kill, ok := params["kill-only"]; ok {
+		killParsed, err := strconv.ParseBool(kill)
+		if err != nil {
+			trace.Status.OperationError = fmt.Sprintf("%q is not valid for failed", kill)
```
burak-ok, line 147: Should be `... not valid for kill-only`

burak-ok (commented):
LGTM. Two little comments:
1. I think squashing the two commits would be better. Updating the Markdown belongs to the first commit imo
2. Is there a reason why you trace the `sys_enter_kill` and `sys_exit_kill` in the case when `KillOnly == true`? The link you provided to `bpf-tools` use the `signal_generate` in that case.
This is more like a exploratory question for me to understand the difference between the two different tracing approaches. From your commit message it seems like that it does not make a difference

burak-ok:
> Good point, I just updated the code as I was mistaken by the fact the `bcc/libbpf-tools` actually deactivates the `signal_generate` probe when given `--kill-only`

Yeah, the `bpf_program__set_autoload(..., false)` fooled me too :D

burak-ok (approved):

PR #1230, opened 2022-12-23
kubectl-gadget: Ensure context works everywhere

### LINKED ISSUE #132: Ensure --context works across the board
While testing kubectl gadget on a machine configured to have several kubernetes contexts, I discovered that the `--context` flag is not always taken into account.

For example, `kubectl gadget --context my-context traceloop list` correctly lists the traceloops gathered in `my-context`. However, `kubectl gadget --context my-context traceloop show <traceid>` tries to connect to the kube API of the current context (and not `my-context`).

When the contexts don't match, the call responds with an error like `pods "gadget-kg4qz" not found`.  For this to work, one needs to change the setting of the current context in the config file, which is not ideal for operators working with several contexts.

mqasimsarfraz:
It seems that we live without creating a our own `kubeRestConfig` and use the global one to allow sharing kubernetes config. 

Fixes #132

### Test done

```
# create couple of clusters
minikube -p foo start
minikube -p bar start

# install gadget on 'foo' cluster
kubectl config use-context foo
make MINIKUBE="minikube -p foo" minikube-install

# ensure you are using 'bar' context and try tracing
kubectl config use-context bar
./kubectl-gadget trace open -A # should fail
./kubectl-gadget trace open -A --context foo #  should work (should return an error on 'main')
```


mauriciovasquezbernal (approved):

cmd/kubectl-gadget/utils/exec.go
```diff
@@ -62,7 +62,7 @@ func ExecPod(client *kubernetes.Clientset, node string, podCmd string, cmdStdout
 	}
 	podName := pods.Items[0].Name
 
-	restConfig, err := kubeRestConfig()
+	restConfig, err := kubeRestConfig(KubernetesConfigFlags)
```
mauriciovasquezbernal, line 65: Does it make sense to pass that variable, perhaps kubeRestConfig() could take it directly? 

Also, I see it's the only place where `kubeRestConfig()` is used, other places use `utils.KubernetesConfigFlags.ToRESTConfig()` directly. Perhaps we can remove this function and use that directly as in other places?

mqasimsarfraz: Thanks for the suggestion. Yeah, it makes sense to remove this function since it looks more clear e.g why we were doing `factory.SetKubernetesDefaults(restConfig)` etc. I updated the PR!

mauriciovasquezbernal (approved):
LGTM!

PR #1231, opened 2022-12-24
Add latency field to DNS trace gadget

wedaly:
# Add latency field to DNS trace gadget

Derive DNS latency from the request and response timestamps. To avoid unbounded memory usage, set a limit on the number of outstanding requests, dropping old requests to make space if necessary.

DNS requests are tracked in userspace. I tried using BPF_MAP_TYPE_LRU_HASH in the eBPF program, but discovered that the LRU cache would evict entries when the program was executed on different CPUs, even when there were only a few entries in the map. For example, if nslookup queried for A and AAAA records, the second request executing on a different CPU could cause the first request to be evicted from the map. The userspace implementation avoids this issue.

Part of issue https://github.com/inspektor-gadget/inspektor-gadget/issues/1196

Depends on changes in https://github.com/inspektor-gadget/inspektor-gadget/pull/1205 to set the timestamp on the DNS event.

## How to use

https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadgets/trace/dns.md

## Testing done

image


pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -226,7 +243,7 @@ func parseDNSEvent(rawSample []byte) (*types.Event, error) {
 	}
 
 	// Convert name into a string with dots
-	event.DNSName = parseLabelSequence(rawSample[unsafe.Offsetof(bpfEvent.Name):])
+	event.DNSName = parseLabelSequence(bpfEvent.Name[:])
```
wedaly, line 246: I think this is equivalent -- is there a reason to prefer reading the data from `rawSample`?

wedaly:
> I am nonetheless not so sure about using golang `map`. Can you please give me more details regarding `BPF_MAP_TYPE_LRU_HASH`? I am not sure to understand clearly this part:
> 
> > but discovered that the LRU cache would evict  when the program was executed on different CPUs, even when were only a few entries in the map
> 
> Particularly, when would the eviction occurs? Does the whole map is cleared if the eBPF program is scheduled on a different CPU than before?

The behavior I observed was that some entries would get evicted, but not the whole map. There are more details in these Slack discussions:

* [#inspector-gadget](https://kubernetes.slack.com/archives/CSYL75LF6/p1671729501829919)
* [cilium #ebpf](https://cilium.slack.com/archives/C4XCTGYEM/p1671743048479469)

@mauriciovasquezbernal and I were both surprised by this behavior.

> 
> If so, what about using a simple eBPF hash map (_i.e._ not the LRU one)?

I think that's worth exploring! @alban [suggested maybe using a BPF timer or golang ticker to expire old entries](https://github.com/inspektor-gadget/inspektor-gadget/issues/1196#issuecomment-1357418679), but seems like that would require kernel versions > 4.10. Thoughts?

wedaly:
updated to:
* fix linter violations (var/func naming)
* use embedded `sync.Mutex`
* fix TestDNS local-gadget unit test

one of the integration tests is failing, looking into it

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -44,6 +47,15 @@ func GetColumns() *columns.Columns[Event] {
 	col, _ := cols.GetColumn("container")
 	col.Visible = false
 
+	cols.MustSetExtractor("latency", func(event *Event) string {
+		if event.Latency > 0 {
+			return event.Latency.String()
+		} else {
+			// latency <= 0 means none reported (e.g. for requests or responses whose request was evicted)
+			return ""
```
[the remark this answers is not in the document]
wedaly: updated to return "UNKNOWN" for responses where the lookup fails (either due to eviction or the DNS packet had an invalid ID)

blanquicet (commented):
Happy to see you again with this fantastic contribution! 🚀 ... I added some comments.

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: I suppose it's a leftover from previous iterations. It should be fine to change it.

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -44,6 +47,21 @@ func GetColumns() *columns.Columns[Event] {
 	col, _ := cols.GetColumn("container")
 	col.Visible = false
 
+	cols.MustSetExtractor("latency", func(event *Event) string {
```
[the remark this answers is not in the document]
wedaly: Sounds good to me. Maybe we could mark latency as "hidden" and let the user opt-in using `-o custom-columns`?

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func newDNSLatencyCalculator() *dnsLatencyCalculator {
+	return &dnsLatencyCalculator{
+		currentReqTSMap: make(map[dnsReqKey]uint64),
+		prevReqTSMap:    make(map[dnsReqKey]uint64),
+	}
+}
+
+func (c *dnsLatencyCalculator) storeDNSRequestTimestamp(saddr [16]uint8, id uint16, timestamp uint64) {
```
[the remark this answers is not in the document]
wedaly: ack, updated

integration/inspektor-gadget/integration_test.go
```diff
@@ -569,6 +571,9 @@ func TestDns(t *testing.T) {
 			normalize := func(e *dnsTypes.Event) {
 				e.Node = ""
 				e.ID = "0000"
+				if e.Latency > 0 {
+					e.Latency = 1
+				}
```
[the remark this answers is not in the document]
wedaly: I think if we always set it to zero, then the test would pass even if the tracer failed to set the `Latency` field (leaving the zero value). This also helps assert that `Latency > 0` only for the DNS responses, not the requests.

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -44,14 +44,36 @@ func NewTracer() (*Tracer, error) {
...
+			return nil, errors.New("invalid sample size")
+		}
+
+		event, err := bpfEventToDNSEvent(bpfEvent)
+		if err != nil {
+			return nil, err
+		}
+
+		// Derive latency from the request/response timestamps.
+		if bpfEvent.Qr == 0 {
```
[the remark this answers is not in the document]
wedaly: thanks, updated!

pkg/gadgets/trace/dns/types/dns.go
```diff
@@ -44,6 +47,21 @@ func GetColumns() *columns.Columns[Event] {
...
+			// Latency reported only for responses, not queries.
+			return ""
+		}
+
+		if event.Latency > 0 {
+			return event.Latency.String()
+		} else {
+			// This means lookup for the request's timestamp failed, possibly because the request was evicted
+			// or the DNS packet contained an invalid ID.
+			return "UNKNOWN"
```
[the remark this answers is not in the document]
wedaly: yeah, definitely open to suggestions here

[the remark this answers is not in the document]
wedaly: hmm, that's a good point. I'd imagine we'd want to continue using a numeric type in the JSON so clients don't need to parse the value from a string. We could use a magic value like `-1` to mean "unknown", but that also seems confusing. Maybe `null` is clearer?

wedaly:
> @wedaly what's the memory usage limit you want for this gadget? Perhaps we can use a bigger sized on the LRU ebpf map and it'll avoid evicting the entries?

I think the exact limit doesn't matter so much, as long as it's bounded. I tried setting `max_entries` in the LRU map to 4096, which seemed to help, but I'd be nervous about relying on that without understanding the mechanism that's evicting the keys.

> Otherwise, what about using an LRU golang map on user space instead of implementing our own logic? (something like https://github.com/hashicorp/golang-lru could work).

Yep, definitely open to that if you're comfortable taking the dependency.

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,100 @@
...
+	c.Lock()
+	defer c.Unlock()
+
+	// If the current map is full, drop the previous map and allocate a new one to make space.
+	if len(c.currentReqTSMap) == dnsLatencyMaxMapSize {
+		c.prevReqTSMap = c.currentReqTSMap
+		c.currentReqTSMap = make(map[dnsReqKey]uint64)
```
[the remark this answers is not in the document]
wedaly: thanks for putting that together, it's helpful to compare different approaches! I guess if we keep increasing N we'll eventually end up building an LRU cache :) 

Per @mauriciovasquezbernal 's comment below, I'm kind of leaning towards using an off-the-shelf LRU cache rather than writing something custom here.

wedaly:
last commit reimplements `dnsLatencyCalculator` using https://github.com/hashicorp/golang-lru to reduce the amount of custom code

integration/inspektor-gadget/integration_test.go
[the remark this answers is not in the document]
blanquicet: Ahh ok, I got you. The idea is to ensure the Latency was set for DNS responses. Then, the code is OK. Thanks for the clarification. Could you please add a comment like "Ensuring latency was set for DNS responses"?

pkg/gadgets/trace/dns/tracer/latency.go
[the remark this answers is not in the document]
blanquicet: Make sense. It was fun to make those experiments anyway :).

pkg/gadgets/trace/dns/types/dns.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I agree to hide it by default in the meanwhile. We should not add a lot of columns by default, otherwise it'll need a very large terminal to see it correctly.

pkg/gadgets/trace/dns/types/dns.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think using a pointer is a good idea. If the pointer is nil then it means the latency is not available.

mauriciovasquezbernal (commented):
I like it. Few small comments to handle, it should ready after that. 

Please squash the commits to get ready to be merged. 

@alban since you have worked in this gadget a lot, would you like to review this PR?

pkg/gadgets/trace/dns/types/dns.go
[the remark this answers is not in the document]
blanquicet: Mmmmm not sure, maybe `N/A`? What do other people think?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -44,14 +44,40 @@ func NewTracer() (*Tracer, error) {
...
+		}
+
+		event, err := bpfEventToDNSEvent(bpfEvent)
+		if err != nil {
+			return nil, err
+		}
+
+		// Derive latency from the request/response timestamps.
+		if event.Qr == types.DNSPktTypeQuery {
+			latencyCalc.storeDNSQueryTimestamp(bpfEvent.SaddrV6, bpfEvent.Id, bpfEvent.Timestamp)
```
blanquicet, line 65: BTW, not related to this PR but `bpfEvent.SaddrV6` and `bpfEvent.DaddrV6`, should be just `bpfEvent.Saddr` and `bpfEvent.Daddr`. We can do it in another PR.

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,74 @@
...
+	// Store the timestamp of the query so we can calculate the latency once the response arrives.
+	c.queryCache.Add(dnsQueryKey{saddr, id}, timestamp)
+}
+
+// If there is no corresponding DNS query (either never received or evicted to make space), then this returns zero.
+func (c *dnsLatencyCalculator) calculateDNSResponseLatency(daddr [16]uint8, id uint16, timestamp uint64) time.Duration {
+	key := dnsQueryKey{daddr, id}
+	reqTS, ok := c.queryCache.Get(key)
+	if !ok {
+		// Either an invalid ID or we evicted the query from the map to free space.
```
blanquicet, line 58: Should we log this at the beginning to give us an idea of how many times this happens?

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,74 @@
...
+		return 0
+	}
+
+	c.queryCache.Remove(key)
+
+	if reqTS > timestamp {
+		// Should never happen assuming timestamps are monotonic, but handle it just in case.
```
blanquicet, line 65: dito

blanquicet (commented):
I think it is ready after a rebase and squashing the fixup commits.

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,74 @@
...
+
+// dnsLatencyCalculator calculates the latency between a query and its response.
+// It uses an LRU cache to bound memory usage.
+// All operations are thread-safe.
+type dnsLatencyCalculator struct {
+	queryCache *lru.Cache[dnsQueryKey, uint64] // This is thread-safe.
+}
+
+func newDNSLatencyCalculator() (*dnsLatencyCalculator, error) {
+	queryCache, err := lru.New[dnsQueryKey, uint64](dnsLatencyCacheSize)
```
blanquicet, line 41: Just one curiosity, I check the documentation, and it seems there is no need to clean/remove the lru once we are done. It is all left to the GC?

pkg/gadgets/trace/dns/types/dns.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Just `-`? Or perhaps leaving it empty?

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: It seems we have a mixed approach:

https://github.com/inspektor-gadget/inspektor-gadget/blob/91c1786200489a56acc530c202aad82acf5dd60a/pkg/gadgets/trace/dns/tracer/bpf/dns-common.h#L10-L17

https://github.com/inspektor-gadget/inspektor-gadget/blob/91c1786200489a56acc530c202aad82acf5dd60a/pkg/gadgets/trace/tcp/tracer/bpf/tcptracer.h#L22-L31

https://github.com/inspektor-gadget/inspektor-gadget/blob/91c1786200489a56acc530c202aad82acf5dd60a/pkg/gadgets/trace/tcpconnect/tracer/bpf/tcpconnect.h#L27-L34

pkg/gadgets/trace/dns/tracer/latency.go
wedaly: bit concerned about perf implications of adding a log call here, since this code runs on every DNS event.

A user could infer that this is happening from the event stream directly (count number of DNS responses without a latency field).

pkg/gadgets/trace/dns/tracer/latency.go
wedaly: I think the event stream is sufficient to detect that this is happening, since each event includes the timestamp.

pkg/gadgets/trace/dns/tracer/latency.go
wedaly: yep, that's my understanding too -- should get cleaned up by the garbage collector

pkg/gadgets/trace/dns/types/dns.go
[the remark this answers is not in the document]
wedaly: updated to leave it empty. That's consistent with the JSON output for zero value (since we're using `omitempty` to omit the key if it's zero)

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -44,14 +44,40 @@ func NewTracer() (*Tracer, error) {
 		return nil, fmt.Errorf("failed to load asset: %w", err)
 	}
 
+	latencyCalc, err := newDNSLatencyCalculator()
```
alban, line 47: The same LRU cache is used for matching packets from different container / different network namespaces. I am concerned of the interaction it could have. I will run some tests.

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,74 @@
...
+// If there is no corresponding DNS query (either never received or evicted to make space), then this returns zero.
+func (c *dnsLatencyCalculator) calculateDNSResponseLatency(daddr [16]uint8, id uint16, timestamp uint64) time.Duration {
+	key := dnsQueryKey{daddr, id}
+	reqTS, ok := c.queryCache.Get(key)
+	if !ok {
+		// Either an invalid ID or we evicted the query from the map to free space.
+		return 0
+	}
+
+	c.queryCache.Remove(key)
```
alban, line 62: Here the cache is removed. Since the cache is global and not per-netns, we get weird results:
```
$ ./kubectl-gadget trace dns -A -o custom-columns=timestamp,pod,id,qr,type,qtype,latency,name
TIMESTAMP                           POD                                  ID   QR TYPE      QTYPE      LATENCY              NAME                               
2023-01-16T12:16:30.882868135+01:00 normal-pod-cnl4l                     7da6 Q  OUTGOING  A                               default.svc.cluster.local.         
2023-01-16T12:16:30.882932625+01:00 coredns-565d847f94-bkdpt             7da6 Q  HOST      A                               default.svc.cluster.local.         
2023-01-16T12:16:30.882887281+01:00                                      7da6 Q  OTHERHOST A                               default.svc.cluster.local.         
2023-01-16T12:16:30.882928220+01:00                                      7da6 Q  OUTGOING  A                               default.svc.cluster.local.         
2023-01-16T12:16:30.883202854+01:00 coredns-565d847f94-bkdpt             7da6 R  OUTGOING  A          274.634µs            default.svc.cluster.local.         
2023-01-16T12:16:30.883216137+01:00                                      7da6 R  OTHERHOST A                               default.svc.cluster.local.         
2023-01-16T12:16:30.883235411+01:00                                      7da6 R  OUTGOING  A          367.276µs            default.svc.cluster.local.         
2023-01-16T12:16:30.883244999+01:00                                      7da6 R  OUTGOING  A                               default.svc.cluster.local.         
2023-01-16T12:16:30.883249613+01:00 normal-pod-cnl4l                     7da6 R  HOST      A                               default.svc.cluster.local.         
^C
Terminating...
```

When my pod `normal-pod-cnl4l` receives the dns response, the latency is not displayed because the cache has already been removed in a previous entry when the same packet has been detected in another network namespace.

Tested with:

```
$ kubectl exec -ti -n default normal-pod-cnl4l -- nslookup -type=a default.svc.cluster.local
Server:		10.96.0.10
Address:	10.96.0.10:53
```

alban (commented):
Some comments about the cache being global vs per-netns

wedaly:
@alban oof, thanks for catching that. I've attempted to address the issue of sharing the LRU cache across netns by including the netns in the LRU cache key. I've also removed the IP address from the cache key, since I think netns + DNS ID should be sufficient to correlate queries/responses.

While testing with `-A` to include all namespaces, I noticed another problem. For DNS queries from the host netns, we see four events, all in the same host netns with the same DNS ID:

1. Query with pkt type OUTGOING
2. Query with pkt type HOST
3. Reply with pkt type OUTGOING
4. Reply with pkt type HOST

The second query event resets the timestamp, and the first reply event deletes the cache entry.

Maybe there's a better way to handle this, but I'm thinking maybe we store the timestamp only for queries with pkt type OUTGOING, and calculate latency only for replies with pkt type HOST. I'd imagine that's what users of this tracer would care about: the latency between when a container sent the query and got a reply.

With these changes, the output running a trace with  `-A` looks like this: 
image

I made those changes in the last two commits so they're easier to review. Thoughts on this approach?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -95,6 +97,11 @@ var pktTypeNames = []string{
 	"KERNEL",
 }
 
+const (
+	pktTypeHost     uint8 = 0
+	pktTypeOutgoing uint8 = 4
```
alban, line 102:
```
unix.PACKET_HOST
unix.PACKET_OUTGOING
```
From https://pkg.go.dev/golang.org/x/sys/unix#pkg-constants

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -60,10 +60,12 @@ func NewTracer() (*Tracer, error) {
...
 
-		// Derive latency from the request/response timestamps.
-		if event.Qr == types.DNSPktTypeQuery {
+		// Derive latency from the query/response timestamps.
+		// Filter by packet type (OUTGOING for queries and HOST for responses) to exclude cases where
+		// the packet is forwarded between containers in the host netns.
+		if bpfEvent.Qr == 0 && bpfEvent.PktType == pktTypeOutgoing {
 			latencyCalc.storeDNSQueryTimestamp(netns, bpfEvent.Id, bpfEvent.Timestamp)
-		} else {
+		} else if bpfEvent.Qr == 1 && bpfEvent.PktType == pktTypeHost {
```
alban, line 68: I wonder if we can get the two following latencies by just filtering for OUTGOING/HOST (excluding ) regardless of the `Qr` field:
1. in the target pod, time between the request is sent and the reply is received
2. in the coredns pod, time between the request is received and the reply is sent (that latency should be smaller than the previous one)

I have not tested if the output would make sense in that case.

I think we might need the interface (`eth0`, `docker0`) to distinguish duplicated entries on the host network namespace. It should be available from `(struct __sk_buff *skb)->ifindex`. Then we have [net.InterfaceByIndex(ifindex)](https://pkg.go.dev/net#InterfaceByIndex)->Name but it should be called from the right netns (with `NetnsEnter()`) and it should be cached, so that should be for another PR.

alban (changes requested):
I think it's good.

I added some comments but they don't necessarily need to be addressed in this PR.

> With these changes, the output running a trace with -A looks like this:

I noticed that events 3 and 4 are not in the right order, so it makes it more difficult to interpret (it's unrelated to this PR though).

pkg/gadgets/trace/dns/tracer/tracer.go
wedaly: thanks, updated!

pkg/gadgets/trace/dns/tracer/tracer.go
wedaly: > I wonder if we can get the two following latencies by just filtering for OUTGOING/HOST (excluding ) regardless of the `Qr` field:
> 
>     1. in the target pod, time between the request is sent and the reply is received
> 
>     2. in the coredns pod, time between the request is received and the reply is sent (that latency should be smaller than the previous one)
> 
> I have not tested if the output would make sense in that case.

The behavior with the `Qr` check is:

1. if coredns queries an upstream DNS server, latency for the query from coredns->upstream is reported.
2. if coredns serves the query directly, then latency is not reported for the coredns pod (since it doesn't send any query).

I think if we removed the `Qr` check, then the behavior is the same for case (1), but for case (2) we'd instead report the latency for coredns to serve the request.

My feeling is that it's clearer if latency always means the time between a query/reply, and I'd imagine that's what someone investigating a DNS issue would care most about.
 
> I think we might need the interface (`eth0`, `docker0`) to distinguish duplicated entries on the host network namespace. It should be available from `(struct __sk_buff *skb)->ifindex`. Then we have [net.InterfaceByIndex(ifindex)](https://pkg.go.dev/net#InterfaceByIndex)->Name but it should be called from the right netns (with `NetnsEnter()`) and it should be cached, so that should be for another PR.

That makes sense. I wonder if we could skip `InterfaceByIndex` and just make the DNS cache key (netns, ifindex, dnsID).

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2022 The Inspektor Gadget authors
```
alban, line 1:
```suggestion
// Copyright 2023 The Inspektor Gadget authors
```

pkg/gadgets/trace/dns/tracer/latency_test.go
```diff
@@ -0,0 +1,129 @@
+// Copyright 2022 The Inspektor Gadget authors
```
alban, line 1:
```suggestion
// Copyright 2023 The Inspektor Gadget authors
```

pkg/gadgets/trace/dns/tracer/latency.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"time"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+const dnsLatencyCacheSize int = 1024
+
+// dnsQueryKey is a unique identifier for a DNS query.
+// netns is the network namespace where the packet was received.
```
alban, line 26:
```suggestion
// netns is the network namespace where the packet was observed.
```

alban (approved):
LGTM after a couple of small changes. Thanks for implementing this.

wedaly:
seeing some test failures, but I think they're unrelated:
* Integration tests (core, none) TestTraceTcpconnect failing with `Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "binding.spo.io": failed to call webhook: Post "[https://webhook-service.security-profiles-operator.svc:443/mutate-v1-pod-binding?timeout=10s](https://webhook-service.security-profiles-operator.svc/mutate-v1-pod-binding?timeout=10s)": dial tcp 10.102.30.75:443: i/o timeout`
* Integration tests for local gadget (non-K8s, docker) TestTraceDns is failing with a missing event, but it passes locally for me.

Going to rebase onto the main branch to see if that helps.

mqasimsarfraz:
> * Integration tests for local gadget (non-K8s, docker) TestTraceDns is failing with a missing event, but it passes locally for me.

This should be unrelated as well since from the [logs](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3949717373/attempts/1#summary-10727271605) here I believe we might need to wait for tracer to be ready like [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/local-gadget/k8s/trace_dns_test.go#L112) before making the requests!

cheneytianx:
> I tried using BPF_MAP_TYPE_LRU_HASH in the eBPF program, but discovered that the LRU cache would evict entries when the program was executed on different CPUs, even when there were only a few entries in the map.

Any other references about the LRU map? I can not see the content in the slack. 

Thanks.

[#inspector-gadget](https://kubernetes.slack.com/archives/CSYL75LF6/p1671729501829919)
[cilium #ebpf](https://cilium.slack.com/archives/C4XCTGYEM/p1671743048479469)

mauriciovasquezbernal:
> I can not see the content in the slack.

Oh yes, this is the thing I don't like about discussing on Slack, there is a barrier to access that information and old messages are removed in some cases. 

Any way, this is the full discussion in [cilium #ebpf](https://cilium.slack.com/archives/C4XCTGYEM/p1671743048479469):

Will Daly: 
 
> Hi, I've observed BPF_MAP_TYPE_LRU_HASH with max_entries set to 256 evicting an entry when there are only a few other entries in the map. The [docs](https://www.kernel.org/doc/html/latest/bpf/map_hash.html) say "An LRU hash will automatically evict the least recently used entries when the hash table reaches capacity" so I expected no entries would be evicted until the number of entries approached max_entries.
I originally reproduced this working on inspektor-gadget ([discussion here](https://kubernetes.slack.com/archives/CSYL75LF6/p1671729501829919)) but I'm also able to reproduce it using the "xdp" example in cilium/ebpf which uses an [LRU hash with IP addresses as keys](https://github.com/cilium/ebpf/blob/1e326d1d24eb2608e3e37e5e994760ad81f17f38/examples/xdp/xdp.c#L12):
> - In the cilium/ebpf "examples" directory: go run -exec sudo ./xdp lo to start tracing packets on the loopback interface.
> - for i in {0..15}; do ping 127.0.0.${i} -c 1; sleep 1; done
> I'd expect to see all 16 IP addresses in the LRU map, but the actual output shows keys missing (can see the same using bpftool map dump):
        127.0.0.1 => 17
        127.0.0.15 => 1
        127.0.0.0 => 1
        127.0.0.14 => 1
        127.0.0.9 => 1
        127.0.0.8 => 1

> one of my colleagues was also able to reproduce this behavior with the test program here: https://github.com/mauriciovasquezbernal/lru_hash_experiment

> could someone help us understand why BPF_MAP_TYPE_LRU_HASH might evict entries when the number of entries is much smaller than max_entries?

> testing a bit more, it seems that when using this command, all the IPs show up in the map:
`taskset -c 1 ping 127.0.0.${i} -c 1`
so maybe this behavior has something to do with map updates from different CPUs?


Joe Stringer:

> Yeah it's likely a multi-CPU thing. If you wanted perfect LRU, that could technically be done but it would be horribly inefficient when running the algorithm across various CPUs. So, the Linux implementation makes some tradeoffs around how strictly to enforce the LRU behaviour (ie not strictly) in order to improve performance.

> I did a bit of a breakdown of some of the ways this works at Linux Plumbers Conference this year, that could be helpful for a bit of insight around the details on that platform.

> (slides/video should be available, I don't have the links on hand but you might find them either with a slack search for Linux Plumbers / LPC, or through an internet search)

Will Daly:

> thanks 
[@Joe Stringer](https://cilium.slack.com/team/U6B7RKMMX)
, I'll take a look!


I hope it helps.

cheneytianx:
@mauriciovasquezbernal Thanks, this helps a lot.

Issue #734, opened 2022-07-06
[RFE] Add support for non-Kubernetes containers with Containerd


blanquicet:
## Current situation

Currently, In [the runtime-client](https://github.com/kinvolk/inspektor-gadget/blob/main/pkg/container-utils/runtime-client/interface.go) package, the Docker implementation does not use the Docker CRI but directly the Docker Engine API so it is able to retrieve both kinds of containers: Kubernetes and non-Kubernetes. On the contrary, with Contained, we are using the CRI so we don't have visibility of the containers that weren't created via Kubernetes.

## Impact

With Contained, we don't have visibility of the containers that weren't created via Kubernetes as we have with Docker. For the use-case of `local-gadget`, it is a big disadvantage because we are missing part or even all the containers running in the system.

## Ideal future situation

We should use with the dedicated API for Containerd so that we have visibility of all the containers running in the system.

burak-ok:
For `CRI-O` i searched and found this https://pkg.go.dev/github.com/cri-o/cri-o@v1.26.0/internal/client
But this is in a directory called "internal" which means that it could be quite unstable? Should we keep using the `CRI` Client for `CRI-O`?

blanquicet:
Yep, according to CRI-O documentation: [What is the scope of this project?](https://github.com/cri-o/cri-o#what-is-the-scope-of-this-project) and [What is not in the scope of this project?](https://github.com/cri-o/cri-o#what-is-not-in-the-scope-of-this-project), the scope of CRI-O is tied to the scope of the CRI. Therefore, I would say that it doesn't make sense to support non-Kubernetes containers for CRI-O.

PR #1233, opened 2022-12-27
container-utils: Use the containerd API

burak-ok:
# container-utils: Use the containerd API

See https://github.com/inspektor-gadget/inspektor-gadget/issues/734

Usage of the containerd API modifies some behavior of the
ContainerdClient. Instead of failing on the first API Call it will fail
when the ContainerdClient is created.
The default namespace for the ContainerdClient will remain "k8s.io" for
now. This assures backwards compatiblity.
Additionally all "sandbox" containers are filtered out.

I tried to keep the behavior mostly the same.
Differences:
- On client creation via `ContainerdClient` the method will fail if there are any problems. Via `CRI` or `docker` nothing fails on the creation until the first request is actually made.

Notes:
- Default namespace is `k8s.io`, which should be configurable later
- Containers with the label `io.cri-containerd.kind=sandbox` are getting filtered out. These are the sleep containers etc
- We should check on all `ContainerRuntimeClient` what happens when the creation fails, etc... Then align the behavior of all clients with our expectations

----- OLD: -----

# Comments / Problems / ...
- I feel the API is somehow clunky to use. There are some structs with methods which i can use for e.g. `Task` and then i need to talk directly to the GRPC Wrapper for other `Task` related stuff
- We need to have a new parameter which specifies a `containerd` namespace. Last time i used the `k8s.io` namespace, but others work too.
- Getting a Containername is somehow hard/ambiguous. As far as i could see `containerd` has no concept like Names. I searched in `nerdctl` and it stores the name in a label. This is what i try to search for. Same for the k8s name label


pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,185 @@
...
 	if socketPath == "" {
 		socketPath = runtimeclient.ContainerdDefaultSocketPath
 	}
 
-	criClient, err := criclient.NewCRIClient(runtimeclient.ContainerdName, socketPath, DefaultTimeout)
+	client, err := containerd.New(socketPath,
+		containerd.WithTimeout(DefaultTimeout),
+		containerd.WithDefaultNamespace("k8s.io"),
+		containerd.WithDialOpts([]grpc.DialOption{
+			grpc.WithBlock(),
```
blanquicet, line 52: I had to remove this to make it works.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,185 @@
...
+
+	labels, err := containers[0].Labels(context.TODO())
+	if err != nil {
+		return nil, err
+	}
+
+	containerData := &runtimeclient.ContainerData{
+		ID:      proc.ID,
+		Name:    guessContainerName(containers[0]),
+		State:   processStatusStateToRuntimeClientState(proc.Status),
```
blanquicet, line 194: Getting the task and then getting the container means we need two, instead of one request, to enrich each container. It seems to be needed to get the `State`, but I am not sure it is worth it. `nerdctl` is directly using `client.Containers()` for the [ps command](https://github.com/containerd/nerdctl/blob/667307635d06f38616699a0d52f96f75389fcefa/cmd/nerdctl/ps.go#L91) and it somehow infers the status. Did you check it?

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,185 @@
...
+}
+
+func guessContainerName(container containerd.Container) string {
+	labels, err := container.Labels(context.TODO())
+	if err != nil {
+		return container.ID()
+	}
+
+	if nerdctlName, ok := labels[ContainerNameNerdctl]; ok {
+		return nerdctlName
```
blanquicet, line 169: What is the nerdctlName? When is it set?

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,185 @@
...
+	} else if k8sName, ok := labels[ContainerNameK8s]; ok {
+		return k8sName
+	}
+
+	return container.ID()
```
blanquicet, line 174: I think here you are matching the case of pause containers. I would suggest dropping them for now and we can manage it later, see https://github.com/inspektor-gadget/inspektor-gadget/issues/1095.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,185 @@
...
+		return nil, err
+	}
+
+	containers, err := c.client.Containers(context.TODO(), fmt.Sprintf("id==%s", containerID))
+	if err != nil {
+		return nil, err
+	}
+	if len(containers) != 1 {
+		return nil, fmt.Errorf("expected 1 container with id %q, got %d", containerID, len(containers))
+	}
```
blanquicet, line 123: It is redundant. I see three calls here to enrich every container. It is too much from my point of view, we need to make it simpler and faster. We don't want to delay the startup during container enrichment.

blanquicet (commented):
> I feel the API is somehow clunky to use. There are some structs with methods which i can use for e.g. `Task` and then i need to talk directly to the GRPC Wrapper for other `Task` related stuff
> 
Mmmm what do you mean? I actually had to remove the gRPC configuration you had to create for the client to make the local-gadget work.

> We need to have a new parameter which specifies a `containerd` namespace.
> 
Yep, this is something we could add in the future. For now, we can focus only on Kubernetes containers and they will all be created by default tin the `k8s.io` namespace.

> Last time i used the `k8s.io` namespace, but others work too.
>
If you don't set this namespace, you shouldn't be able to see the Kubernetes containers.

> Getting a Containername is somehow hard/ambiguous. As far as i could see `containerd` has no concept like Names. I searched in `nerdctl` and it stores the name in a label. This is what i try to search for. Same for the k8s name label
>
Yep, it looks strange. Regarding Kubernetes containers, containerd doesn't seem to give a name like Docker. It is true that `nerdctl` shows something like `k8s://<namespace>/<podname>/<containername>`. However, unlike Docker name (k8s_...), it cannot be used for filtering to do something like `nerdctl inspect k8s://...`. I think it is just the way containerd found to show the k8s metadata for each container. And, given that we have our own, I suggest to don't using this `k8s://...` naming. Instead, I propose to use the Kubernetes container name as the container name.

pkg/container-utils/containerd/containerd.go
burak-ok: Me too, forgot to push the changes though

pkg/container-utils/containerd/containerd.go
burak-ok: Directly in the `if` statement above.

pkg/container-utils/containerd/containerd.go
burak-ok: This could also happen with manually created `containerd` containers which do not have the labels. For paused containers i probably have to check if a `containerd container` has a running `task` associated with it. Looking into it

pkg/container-utils/containerd/containerd.go
burak-ok: Did not check it before.
As far as i can see from looking at the code they also read the `Task` in [their ContainerStatus formatter function](https://github.com/containerd/nerdctl/blob/667307635d06f38616699a0d52f96f75389fcefa/pkg/formatter/formatter.go#L46)

blanquicet: I see. However, for our use case, I would still try to avoid that additional call and have to deal with `Task`. 

I checked what we do with the `State`, and we only use it for discarding non-running containers after calling `GetContainers`. Instead, for `GetContainer` and `GetContainerDetails`, we don't use it at all. Maybe from `GetContainers`, we could filter the `client.Containers` call to return only the running containers? Is it possible?

pkg/container-utils/containerd/containerd.go
blanquicet: I mean, what does it represent? A different name from the Kubernetes container name?

pkg/container-utils/containerd/containerd.go
blanquicet: We are filtering by `k8s.io` namespace so we won't get non-Kubernetes containers. We can think about it later. 

Regarding the pause containers, I suggest discarding them for now to keep this PR simple, and we can manage them for all the runtimes later in a proper way in https://github.com/inspektor-gadget/inspektor-gadget/issues/1095.

pkg/container-utils/containerd/containerd.go
burak-ok: `nerdctl` saves the containername in that label when it creates a `containerd` container.
But if we currently only support K8s namespace it does not make that much sense.

pkg/container-utils/containerd/containerd.go
burak-ok: The filter string bases on the protobuf fields. They do not appear to give us any informations regarding the state of the container or its task :(
https://github.com/containerd/containerd/blob/main/api/services/containers/v1/containers.pb.go

pkg/container-utils/containerd/containerd.go
burak-ok: I reduced it to 2 calls.
`.TaskService()` to get the running tasks and then `.Containers(id)` to get the rest of the information.

burak-ok:
Updated the commit.
See first Post for a new informations

pkg/container-utils/containerd/containerd.go
blanquicet: Did you check if `client.Containers` maybe already returns only the running containers?

burak-ok: Unfortunately, it returns all containers

blanquicet (commented):
I am concerned mainly by the impact this change can have on the performance, but I wonder if it is really the case. What do other folks think?

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,213 @@
...
+func (c *ContainerdClient) GetContainerDetails(containerID string) (*runtimeclient.ContainerDetailsData, error) {
+	response, err := c.client.TaskService().Get(context.TODO(), &tasks.GetRequest{
+		ContainerID: containerID,
+	})
+	if err != nil {
+		return nil, err
+	}
+	proc := response.Process
+
+	container, err := c.getContainerdContainer(proc.ID)
```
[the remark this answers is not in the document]
burak-ok: I also expected it to be `ContainerID` but that field is empty for containers i created via `ctr` or `kubernetes`.

I can't recall what that field was used for. But it would be useful to know what it represents or had represented in the past?

```bash
$ sudo ctr c create docker.io/library/redis:alpine redis_container_id
$ sudo ctr c ls
CONTAINER             IMAGE                              RUNTIME                  
redis_container_id    docker.io/library/redis:alpine     io.containerd.runc.v2
$ sudo ctr t start -d redis_container_id
$ sudo ctr t ls
TASK                  PID       STATUS    
redis_container_id    620763    RUNNING
```

Source for `ctr t ls` is https://github.com/containerd/containerd/blob/main/cmd/ctr/commands/tasks/list.go#L62 and that is reads the field `ID`

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,213 @@
...
+
+func (c *ContainerdClient) GetContainer(containerID string) (*runtimeclient.ContainerData, error) {
+	response, err := c.client.TaskService().Get(context.TODO(), &tasks.GetRequest{
+		ContainerID: containerID,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	containerData, err := c.ContainerdTaskToContainerData(response.Process)
```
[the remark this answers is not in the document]
burak-ok: Hmmm, do we want verification here? What if its a sandbox container? Should we `return nil, nil` ?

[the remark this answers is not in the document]
blanquicet: Yep, we need to do that verification here. Otherwise, the containers created at runtime will be added without enrichment unless another enricher enriches it or drops it, which is precisely what is happening here. 

After adding a few logs https://github.com/inspektor-gadget/inspektor-gadget/commit/fc5418fe6e2c31f8bfc73a649ddddc981373aa7c, I realized that it was the `OCIConfigEnricher` who was dropping the sandbox containers and enriching the normal containers (because the containerd enricher was failing). My test:

```bash
$ sudo ./local-gadget list-containers -w -v -r containerd
[...]
# Already existing containers
[...]
# Create a pod in another terminal, e.g. kubectl run --rm --image=busybox mypod -it --restart='Never' -- sh
DEBU[0003] runc fanotify: add container "27a08e8e49d1f222b6aa27fa2f47f8418439a2da91cb7bf860f859ee7c18fb5f"
DEBU[0003] OCIConfig enricher: temporary dropping sandbox container "27a08e8e49d1f222b6aa27fa2f47f8418439a2da91cb7bf860f859ee7c18fb5f"
DEBU[0004] runc fanotify: add container "9dd55e85a2596eec5f852cdb616c2b48618a67997d469ac6c6f71ef36b4bf4e9"
DEBU[0004] OCIConfig enricher: container "9dd55e85a2596eec5f852cdb616c2b48618a67997d469ac6c6f71ef36b4bf4e9" was enriched
DEBU[0004] Runtime enricher (containerd): failed to get container: listing task for container "9dd55e85a2596eec5f852cdb616c2b48618a67997d469ac6c6f71ef36b4bf4e9": rpc error: code = NotFound desc = task 9dd55e85a2596eec5f852cdb616c2b48618a67997d469ac6c6f71ef36b4bf4e9 not found
2023-01-10T09:31:49Z           CREATED    containerd 9dd55e85a2596eec5f852cdb616c2b48618a67997d469ac6c6f71ef… mypod
```

Notice that if you remove the [OCIConfigEnricher](https://github.com/inspektor-gadget/inspektor-gadget/blob/98649ab9ea69ff9b7915347882dcd88385fa267d/pkg/local-gadget-manager/local-gadget-manager.go#L351), the containerd enricher will add the sandbox and the normal container without enrichments:

```bash
$ sudo ./local-gadget list-containers -w -v -r containerd -o json | jq .container.podname
[...]
# Already existing containers
"kube-controller-manager"
"coredns"
"kube-proxy"
[...]
# Create a pod in another terminal, e.g. kubectl run --rm --image=busybox mypod -it --restart='Never' -- sh
DEBU[0007] runc fanotify: add container "7761902bd3ff3e333fa3a2c0f443bd5e75debfa9fdd1d3f8c8c5cad42a868849"
DEBU[0007] Runtime enricher (containerd): failed to get container: listing task for container "7761902bd3ff3e333fa3a2c0f443bd5e75debfa9fdd1d3f8c8c5cad42a868849": rpc error: code = NotFound desc = task 7761902bd3ff3e333fa3a2c0f443bd5e75debfa9fdd1d3f8c8c5cad42a868849 not found
null
DEBU[0008] runc fanotify: add container "14e6f198d600361dbd9ffb9f451a65ff9be15e1d41670dea34971eb3fdb0590f"
DEBU[0008] Runtime enricher (containerd): failed to get container: listing task for container "14e6f198d600361dbd9ffb9f451a65ff9be15e1d41670dea34971eb3fdb0590f": rpc error: code = NotFound desc = task 14e6f198d600361dbd9ffb9f451a65ff9be15e1d41670dea34971eb3fdb0590f not found
null
```

This is why I said we should add some tests to verify that each enricher is working correctly. It is because in the previous test, the `OCIConfigEnricher` was hiding the bug in the container enricher.

To summarize, the previous tests show two issues:

1. Containerd enricher should drop sandbox containers also in `GetContainer`. You could use the same strategy used in [Docker](https://github.com/inspektor-gadget/inspektor-gadget/blob/98649ab9ea69ff9b7915347882dcd88385fa267d/pkg/container-utils/docker/docker.go#L95-L97), or directly solve the [main issue](https://github.com/inspektor-gadget/inspektor-gadget/issues/1095) to avoid adding another workaround.
2. When `GetContainer` is called, the container task hasn't been created yet, so the containerd enricher fails to enrich the containers created at runtime. The task is only needed to get the container status, and for `GetContainer`, we don't need the status as we already know that the container is being created. Therefore, in `GetContainer` we can avoid getting the task and we can get the rest of the information with `client.Containers(context.TODO(), fmt.Sprintf("id==%s", id))`. WDYT?

pkg/container-utils/containerd/containerd.go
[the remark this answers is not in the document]
blanquicet: Mmmm, I think it is a problem of how `ctr` show things because I can see the ID behaves as we expect for Kubernetes containers using `nerdctl`:

```bash
$ sudo ./local-gadget list-containers  -v -w -r containerd -c mypod
# Create a pod in another terminal, e.g. kubectl run --rm --image=busybox mypod -it --restart='Never' -- sh
DEBU[0041] runc fanotify: add container "22f48fc5e4d40051b2b3cb11bd39ce109591e6c18e235a70daf7a54eac1ff32c"
DEBU[0041] Runtime enricher (containerd): failed to get container: listing task for container "22f48fc5e4d40051b2b3cb11bd39ce109591e6c18e235a70daf7a54eac1ff32c": rpc error: code = NotFound desc = task 22f48fc5e4d40051b2b3cb11bd39ce109591e6c18e235a70daf7a54eac1ff32c not found
DEBU[0042] runc fanotify: add container "4d8e614ec3ce2a8a571b94ffd1e13697a7d7c52133bcb634e7f3a8c271b7eb89"
DEBU[0042] Runtime enricher (containerd): failed to get container: listing task for container "4d8e614ec3ce2a8a571b94ffd1e13697a7d7c52133bcb634e7f3a8c271b7eb89": rpc error: code = NotFound desc = task 4d8e614ec3ce2a8a571b94ffd1e13697a7d7c52133bcb634e7f3a8c271b7eb89 not found
```

The ID reported by runc-fanotify, which is the one that will be used to call `GetContainer(s)`, is the same that `nerdctl` uses to identify the containers:

```bash
$ sudo nerdctl -n k8s.io ps --filter id=4d8e614ec3ce2a8a571b94ffd1e13697a7d7c52133bcb634e7f3a8c271b7eb89
CONTAINER ID    IMAGE                               COMMAND    CREATED          STATUS    PORTS    NAMES
4d8e614ec3ce    docker.io/library/busybox:latest    "sh"       2 minutes ago    Up                 k8s://default/mypod/mypod
```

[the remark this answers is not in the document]
burak-ok: I think the `ContainerId` field of [Task.Process](https://pkg.go.dev/github.com/containerd/containerd/api/types/task#Process) just could be an old field that is now unused?

When creating a container through `nerctl` i get the following values for `Task.Process.ID` and `Task.Process.ContaienerID`:
```bash
$ sudo nerdctl run -d --name nginxtest -p 8080:80 library/nginx:1.22.1-alpine 
docker.io/library/nginx:1.22.1-alpine:                                            resolved       |++++++++++++++++++++++++++++++++++++++| 
index-sha256:2366ede62d2e26a20f7ce7d0294694fe52b166107fd346894e4658dfb5273f9c:    done
....
layer-sha256:71a3f979d23b00e5731837eef4583847c2d56e35fded467d0a79d5853373b533:    done           |++++++++++++++++++++++++++++++++++++++| 
elapsed: 3.9 s                                                                    total:  9.8 Mi (2.5 MiB/s)                                       
9a9531aefbf7e18b873db8d852f43e0d83bbed9b38e658c7de63289a61d421bb

$ sudo nerdctl ps                                                            
CONTAINER ID    IMAGE                                    COMMAND                   CREATED           STATUS    PORTS                   NAMES
9a9531aefbf7    docker.io/library/nginx:1.22.1-alpine    "/docker-entrypoint.…"    25 seconds ago    Up        0.0.0.0:8080->80/tcp    nginxtest                  

$ sudo ./local-gadget list-containers -v --runtimes=containerd   (LOCAL DEBUG OUTPUT)
....
Container proc id: 9a9531aefbf7e18b873db8d852f43e0d83bbed9b38e658c7de63289a61d421bb
Container proc container id: 
....
``` 

Containers that kubernetes created:
```bash
$ sudo ./local-gadget list-containers -v --runtimes=containerd   (LOCAL DEBUG OUTPUT in Minikube)
....
Container proc id: f806e6391201048ff67266f94e608eaabee9d0ec34b4c60c7373a640e2b9b0c2
Container proc container id: 
....
containerd f806e6391201048ff67266f94e608eaabee9d0ec34b4c60c7373a640e2b9b0c2 kindnet-cni
```

[Task.Process.ContainerID](https://pkg.go.dev/github.com/containerd/containerd/api/types/task#Process) is somehow always empty. And [Task.Process.ID](https://pkg.go.dev/github.com/containerd/containerd/api/types/task#Process) seems to be same as [Container.ID()](https://pkg.go.dev/github.com/containerd/containerd#Container)

pkg/container-utils/containerd/containerd.go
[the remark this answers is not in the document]
burak-ok: Thanks for the 

1. I think that it should be done in another PR. Will do that later, will add this temp workaround.
2. I remember that `nerdctl` did something. The [comment in this line](https://github.com/containerd/nerdctl/blob/b0a75d880ef9e6d1f1a8752804c9087ee2d02f73/pkg/formatter/formatter.go#L48) rightfully mentions that the status should be at least "created" instead of failing to enrich. I'll change the code to handle that gracefully

pkg/container-utils/containerd/containerd.go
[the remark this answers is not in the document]
burak-ok: Another possibility could be the following:
Instead of asking the API to get all current running Tasks we could list all containers.
Then we could query the container if a Task is attached to the container.

This way we don't need to handle instances of `Task.Process`. But there is the possibility that we are calling more API calls than in this current implementation.

pkg/container-utils/containerd/containerd.go
mauriciovasquezbernal: > To avoid such redundancy and improve the performance, we could move the logic of steps 1 and 2 inside the runtime implementations.

Yes, I think we need that. It seems to me the needed for Container and ContainerDetails comes from docker and CRI API. 

> What do you think?

I think it's fine to have 4 calls for the time being, but we should keep an eye to improve it later on.

pkg/container-utils/containerd/containerd.go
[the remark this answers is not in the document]
mauriciovasquezbernal: > Instead of asking the API to get all current running Tasks we could list all containers.
Then we could query the container if a Task is attached to the container.

I think this is a clearer approach.

mauriciovasquezbernal (commented):
I'm a bit late to this PR so I don't know if I got all the details. My main comment is to first get the containers and from them get the tasks. I agree with the performance concerns but IMO we can handle those later on.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,248 @@
...
+		Pid:           int(proc.Pid),
+		CgroupsPath:   spec.Linux.CgroupsPath,
+		Mounts:        mountData,
 	}, nil
 }
+
+// getContainerdContainer returns the corresponding container.Container instance to
+// the given id
+func (c *ContainerdClient) getContainerdContainer(id string) (containerd.Container, error) {
+	containers, err := c.client.Containers(context.TODO(), fmt.Sprintf("id==%s", id))
```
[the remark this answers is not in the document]
burak-ok: Unfortunatly `ContainerService().Get(..)` returns an instance of `containers.Container` but that has no additional methods.
With `Containers(...)` we get instances of `containerd.Container` which has all methods we need

pkg/container-utils/containerd/containerd.go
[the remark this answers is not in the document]
burak-ok: Implemented that approach

burak-ok:
Current implementation:
1. We now collect the running containers by listing all containers and then querying their tasks and status
2. Containers without tasks are considered as "Created"
3. Sandbox containers are filtered out temporarily

This is just a quick push to show the direction. Will try to generalize some parts of the methods later today

pkg/container-utils/containerd/containerd.go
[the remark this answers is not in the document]
mauriciovasquezbernal: What about [`LoadContainer()`](https://pkg.go.dev/github.com/containerd/containerd#Client.LoadContainer) then?

mqasimsarfraz (commented):
Nice that we will be able to support `containerd` containers. I found couple of things:
- When I tried to deploy IG on `minikube-containerd` I am seeing an error:

```
time="2023-01-13T14:57:34Z" level=warning msg="Skip pod gadget/gadget-vmzpt: cannot find container (ID: containerd://2b8fee8012fe23146457fbc92b6a86c7872128a20415e938af13260b9d8b15ff): listing container with id \"containerd://2b8fee8012fe23146457fbc92b6a86c7872128a20415e938af13260b9d8b15ff\": container \"containerd://2b8fee8012fe23146457fbc92b6a86c7872128a20415e938af13260b9d8b15ff\" in namespace \"k8s.io\": not found"
```

maybe we need trim `containerd://` prefix somewhere? 

- CLI feels little bit slow on large setup i.e 100+ containers (probably can be handled in future PR)

```
$ time local-gadget list-containers 
# PR
real    0m0.419s
user    0m0.308s
sys     0m0.121s
# main
real    0m0.163s
user    0m0.107s
sys     0m0.033s

```

go.mod
```diff
@@ -40,6 +40,7 @@ require (
 
 require (
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/containerd/containerd v1.5.1
```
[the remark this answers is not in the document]
burak-ok: No reason. That was just what i automatically got probably. Updated to `v1.6.15`

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,237 @@
...
+	case containerd.Created:
+		runtimeClientState = runtimeclient.StateCreated
+	case containerd.Running:
+		runtimeClientState = runtimeclient.StateRunning
+	case containerd.Stopped:
+		runtimeClientState = runtimeclient.StateExited
+	default:
+		runtimeClientState = runtimeclient.StateUnknown
+	}
+	return
```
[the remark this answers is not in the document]
burak-ok: Didn't like that either. But when removing that return the static analyzer/compiler complains about a missing return.

Restructured that section to an unnamed return value and returning directly in the case statements (I like this now better, but is that also more go-like?)

mqasimsarfraz:
> `runtimeclient.ParseContainerID` was additionally added to remove the prefix if it is present and check for `containerd://` in the `containerd` case

Thanks for handling it quickly. It works fine for me now!

blanquicet:
> When I tried to deploy IG on minikube-containerd I am seeing an error
>
@burak-ok maybe before merging, you could run the inspektor-gadget integration tests in an AKS cluster (which uses containerd) to be sure this PR is not introducing regressions.

Maybe we could consider to add another matrix in the `test-integration-minikube` job, to run it against the different runtimes: docker, containerd and CRI-O? It will be all done in parallel so shouldn't impact the time CI takes to complete.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+
+func (c *ContainerdClient) Close() error {
+	if c.client != nil {
+		return c.client.Close()
+	}
+	return nil
+}
+
+func (c *ContainerdClient) GetContainers() ([]*runtimeclient.ContainerData, error) {
+	containers, err := c.client.Containers(context.TODO(), "")
```
blanquicet, line 67:
```suggestion
	containers, err := c.client.Containers(context.TODO())
```

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+// getContainerdTask returns the corresponding container.Task instance to
+// the given containerd.Container. If there is no task, nil is returned without an error
+func getContainerdTask(container containerd.Container) (containerd.Task, error) {
+	task, err := container.Task(context.TODO(), nil)
+	if err != nil {
+		if !errdefs.IsNotFound(err) {
+			return nil, fmt.Errorf("getting task for container %q: %w", container.ID(), err)
+		}
+		// else it is created, but not running
+		log.Debugf("No task for %q", container.ID())
```
blanquicet, line 185: Could you please add the prefix `ContainerdClient` to all logs?

```suggestion
		log.Debugf("ContainerdClient: No task for %q", container.ID())
```

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+		log.Debugf("ContainerdClient: container %q is a sandbox container. Temporary skipping it", container.ID())
+		return nil, runtimeclient.ErrPauseContainer
+	}
+
+	return container, nil
+}
+
+// getContainerdTask returns the corresponding container.Task instance to
+// the given containerd.Container. If there is no task, nil is returned without an error
+func getContainerdTask(container containerd.Container) (containerd.Task, error) {
```
blanquicet, line 178: We are in the containerd package, so having it in the function names seems unnecessary, maybe `getContainerdTask` -> `getContainerTask`.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	return containerData, container, task, nil
+}
+
+// getContainerdContainer returns the corresponding container.Container instance to
+// the given id
+func (c *ContainerdClient) getContainerdContainer(id string) (containerd.Container, error) {
```
blanquicet, line 162: dito

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+
+// getContainerdTask returns the corresponding container.Task instance to
+// the given containerd.Container. If there is no task, nil is returned without an error
+func getContainerdTask(container containerd.Container) (containerd.Task, error) {
+	task, err := container.Task(context.TODO(), nil)
+	if err != nil {
+		if !errdefs.IsNotFound(err) {
+			return nil, fmt.Errorf("getting task for container %q: %w", container.ID(), err)
+		}
+		// else it is created, but not running
```
blanquicet, line 184:
```suggestion
		// else it is created, but not running according to nerdctl:
		// https://github.com/containerd/nerdctl/blob/b0a75d880ef9e6d1f1a8752804c9087ee2d02f73/pkg/formatter/formatter.go#L48-L55
```

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+		mount := spec.Mounts[i]
+		mountData[i] = runtimeclient.ContainerMountData{
+			Source:      mount.Source,
+			Destination: mount.Destination,
+		}
+	}
+
+	return &runtimeclient.ContainerDetailsData{
+		ContainerData: *containerData,
+		Pid:           int(task.Pid()),
```
blanquicet, line 135: `task` might be `nil`, and in case it is, `GetContainerDetails` should return an error `got zero pid` as others `ContainerRuntimeClient` do.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+// The extra containerd.Container parameter saves an additional call to the API
+func taskAndContainerToContainerData(task containerd.Task, container containerd.Container) (*runtimeclient.ContainerData, error) {
+	labels, err := container.Labels(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("listing labels of container %q: %w", container.ID(), err)
+	}
+
+	// If the container exists, but there is no running task/proc, the state is created
+	status := runtimeclient.StateCreated
+	if task != nil {
```
blanquicet, line 200: I think this logic should be managed inside `getContainerdTask`. What about something like this?

```go
// containerTask represents the task information for a given container.
type containerTask struct {
	status string
	pid    uint32
}

// getContainerTask returns the containerTask information for a given container.
// If the container is not running yet, it returns a containerTask with status
// StateCreated and pid 0.
func getContainerTask(container containerd.Container) (*containerTask, error) {
	task, err := container.Task(context.TODO(), nil)
	if err != nil {
		// According to nerdctl, if there is no task, we can assume the
		// container was just created but it is not running yet:
		// https://github.com/containerd/nerdctl/blob/b0a75d880ef9e6d1f1a8752804c9087ee2d02f73/pkg/formatter/formatter.go#L48-L55
		if !errdefs.IsNotFound(err) {
			return nil, fmt.Errorf("getting task for container %q: %w", container.ID(), err)
		}

		t := &containerTask{
			status: runtimeclient.StateCreated,
			pid:    0,
		}
		log.Debugf("ContainerdClient: No task for %q. Assuming it is in %q status",
			container.ID(), t.status)

		return t, nil
	}

	containerdStatus, err := task.Status(context.TODO())
	if err != nil {
		return nil, fmt.Errorf("getting status of task for container %q: %w", container.ID(), err)
	}

	return &containerTask{
		status: processStatusStateToRuntimeClientState(containerdStatus.Status),
		pid:    task.Pid(),
	}, nil
}
```

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,246 @@
...
+
+	return &runtimeclient.ContainerDetailsData{
+		ContainerData: *containerData,
+		Pid:           int(task.Pid()),
+		CgroupsPath:   spec.Linux.CgroupsPath,
+		Mounts:        mountData,
 	}, nil
 }
+
+func (c *ContainerdClient) getContainerDataAndContainerAndTask(containerID string) (*runtimeclient.ContainerData, containerd.Container, containerd.Task, error) {
```
blanquicet, line 141: nit: I'm not a big fan of this function, so I would remove it as it is called twice and one of them is discarding most of the returned values. Anyway, it is more code styling, so it is up to you.

blanquicet (commented):
I think we are almost ready, just a comment for the task but I think the code and the logic now is clearer. Thanks for managing all the comments.

pkg/container-utils/containerd/containerd.go
burak-ok: What i wanted to avoid to these 3 calls to have in both locations. Doesn't matter for me too (Since it is only used twice)

I also don't like that one of the two usages discard everything except the `containerData` :sweat_smile:

burak-ok:
Solved a race condition where a container that we got was removed in the meantime while we queried its task/labels.

There is still a bug which i am investigating. `make -C integration/local-gadget/k8s test-containerd` fails, but running single test don't

burak-ok:
So the problems drills down to the following:
1. When creating a container a mutex is locked in https://github.com/containerd/containerd/blob/main/runtime/v2/runc/task/service.go#L119
2. `RuncFanotify` interrupts the creating process and gives the ContainerEnrichers every information it currently has
3. The `containerD` enricher asks the `containerD` API for more information about that specific container
4. When the API server tries to serve our request it tries to lock the same mutex as in 1. https://github.com/containerd/containerd/blob/main/runtime/v2/runc/task/service.go#L597
5. A timeout of 2 seconds is hit and an unknown error is sent back

burak-ok:
In order to not get into the deadlock situation the following approach was chosen:
In `GetContainer` the state of the container is not actually queried from the `containerd` API, since this method is only getting called when a new container is detected -> State is always Created anyway.

Should we also modify the methodnames in `ContainerRuntimeClient interface` to reflect these informations and requirements? For example renaming the `GetContainer` to `GetContainerDataOnNewContainer` or something else (and smaller :D)

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,281 @@
...
+func getContainerName(container containerd.Container) string {
+	labels, err := container.Labels(context.TODO())
+	if err != nil {
+		return container.ID()
+	}
+
+	if k8sName, ok := labels[LabelK8sContainerName]; ok {
+		return k8sName
+	}
+
```
mauriciovasquezbernal, line 293: nit (for another PR): We could also look for `nerdctl/name` to support containers created with nerdctl.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/inspektor-gadget/inspektor-gadget
 go 1.19
 
 require (
-	github.com/cilium/ebpf v0.0.0-20230406090554-73acad5e3906
+	github.com/cilium/ebpf v0.9.1
```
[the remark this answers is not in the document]
burak-ok: `go mod tidy` changes that line

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,281 @@
...
+	ret := make([]*runtimeclient.ContainerData, 0, len(containers))
+	for _, container := range containers {
+		if isSandboxContainer(container) {
+			log.Debugf("%scontainer %q is a sandbox container. Temporary skipping it", logPrefix, container.ID())
+			continue
+		}
+
+		task, err := getContainerTask(container)
+		if err != nil {
+			log.Debugf("%sgetting containerTask for container %q: %s", logPrefix, container.ID(), err)
```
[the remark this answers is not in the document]
burak-ok: The `logPrefix` already has the colon and space. Is this way prefered?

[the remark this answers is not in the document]
mauriciovasquezbernal: Ah, that's fine then.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,281 @@
...
+		return runtimeclient.StateExited
+	default:
+		return runtimeclient.StateUnknown
+	}
+}
+
+// getContainerName returns the name of the container. If the container is
+// managed by Kubernetes, it returns the name of the container as defined in
+// Kubernetes. Otherwise, it returns the container ID.
+func getContainerName(container containerd.Container) string {
```
mqasimsarfraz, line 284: nit: I see we already have labels available whenever we call `getContainerName`. Why not reuse that?

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,281 @@
...
+
+	return &runtimeclient.ContainerDetailsData{
+		ContainerData: *containerData,
+		Pid:           int(task.pid),
+		CgroupsPath:   spec.Linux.CgroupsPath,
+		Mounts:        mountData,
+	}, nil
+}
+
+func (c *ContainerdClient) getContainerDataAndContainerAndTask(containerID string) (*runtimeclient.ContainerData, containerd.Container, *containerTask, error) {
```
mqasimsarfraz, line 160: nit: If we aren't planning to reuse it I feel this is too specific and we should move all of this in `GetContainerDetails` directly. For me that would be more readable!

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,281 @@
...
 
 func NewContainerdClient(socketPath string) (runtimeclient.ContainerRuntimeClient, error) {
 	if socketPath == "" {
 		socketPath = runtimeclient.ContainerdDefaultSocketPath
 	}
 
-	criClient, err := criclient.NewCRIClient(runtimeclient.ContainerdName, socketPath, DefaultTimeout)
+	client, err := containerd.New(socketPath,
+		containerd.WithTimeout(DefaultTimeout),
+		containerd.WithDefaultNamespace("k8s.io"),
```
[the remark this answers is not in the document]
alban: I don't know if the containerd client API allows it. If not, we will have to list namespaces and iterate (and cleanup the client object if we can't reuse it)...

pkg/container-utils/containerd/containerd.go
burak-ok: I created the method not only for reusability (which we don't reuse anymore :D).
The reason was also to separate the code block from the function that calls it. For readability and keeping the function body small

pkg/container-utils/containerd/containerd.go
burak-ok: Nice catch

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,281 @@
...
 )
 
 const (
 	DefaultTimeout = 2 * time.Second
+
+	LabelK8sContainerName         = "io.kubernetes.container.name"
+	LabelK8sContainerdKind        = "io.cri-containerd.kind"
+	LabelK8sContainerdKindSandbox = "sandbox"
+
+	logPrefix = "ContainerdClient: "
```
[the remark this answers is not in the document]
burak-ok: Cool idea

mqasimsarfraz (approved):
Thanks for it, it till help greatly in adding support for non-Kuberentes containers. Also, awesome that integration test can now run for different runtimes. :partying_face: 

Since we now check for each task while listing containers it can slow down the startup time of `ig` on a node with a lot of containers. We should keep an eye if that is problematic.

pkg/container-utils/containerd/containerd.go
```diff
@@ -15,31 +15,275 @@
...
+	client *containerd.Client
 }
 
 func NewContainerdClient(socketPath string) (runtimeclient.ContainerRuntimeClient, error) {
 	if socketPath == "" {
 		socketPath = runtimeclient.ContainerdDefaultSocketPath
 	}
 
-	criClient, err := criclient.NewCRIClient(runtimeclient.ContainerdName, socketPath, DefaultTimeout)
+	client, err := containerd.New(socketPath,
```
mqasimsarfraz, line 48: I am getting this error on a node w/o containerd:

```
Error: initializing manager: failed to dial "/run/containerd/containerd.sock": context deadline exceeded: connection error: desc = "transport: error while dialing: dial unix:///run/containerd/containerd.sock: timeout"
```
This means I won't be able to use `ig` on a node w/o containerd without specifying `runtimes` flag.

mqasimsarfraz: @blanquicet do you have any thoughts on this? I think `ig` should keep working even if one of the runtime isn't available. What do you say?

burak-ok: Oh no. It totally should work without `containerd`. That is not the intended behavior. Thanks for catching this

burak-ok: Now I remember what happens:
All other clients have a lazy initialization. When creating the `docker` or `crio` client nothing happens even if there is no socket. Only on the first call they try to connect to the socket and then an error happens. See the following output:
```
WARN[0000] Runtime enricher (docker): couldn't get current containers: listing containers with options {Size:false All:true Latest:false Since: Before: Limit:0 Filters:{fields:map[]}}: Cannot connect to the Docker daemon at unix:///run/docker.sock. Is the docker daemon running? 
WARN[0002] Runtime enricher (containerd): failed to initialize container runtime: failed to dial "/run/containerd/containerd.sock": context deadline exceeded: connection error: desc = "transport: error while dialing: dial unix:///run/containerd/containerd.sock: timeout" 
```
With `docker` the error appears when getting the current containers. Here we only print the error. With the `containerd` client the error instantly happens when initializing the client. Here we return the error.

So should i change that to only printing a warning?

mqasimsarfraz: > So should i change that to only printing a warning?

I believe so like `podman` also has lazy client but I am trying to understand if that is how runtime clients are supposed to be used.

blanquicet: Yep, it is. From the beginning of `ig`, we decided that even if the client fails to initialize, `ig` shouldn't fail and it should continue trying to use that client to enrich future containers because maybe the runtime was not temporarily working when we launched `ig`.

burak-ok: > It should continue trying to use that client to enrich future containers because maybe the runtime was not temporarily working when we launched `ig`.

Currently it is now working this way thought for any container enricher. I tested it with a docker container and renamed the `docker.sock` to something else. Started `ig` and then renamed `docker.sock` back to its original name. No events are showing up

mqasimsarfraz: You mean it working not working for future containers ? For me it works fine for docker:

```
$ sudo systemctl stop docker && sudo systemctl stop docker.socket
$ sudo ig version
v0.17.0
$ sudo ig list-containers -r docker -w
WARN[0000] Runtime enricher (docker): couldn't get current containers: listing containers with options {Size:false All:true Latest:false Since: Before: Limit:0 Filters:{fields:map[]}}: Cannot connect to the Docker daemon at unix:///run/docker.sock. Is the docker daemon running? 
TIMESTAMP                      EVENT      RUNTIME    ID
# events from future
2023-06-23T12:05:56+02:00      CREATED    docker     59818fa298d961fb47c21bdf27202c50813dd6f46a8679e… angry_kirch8                                                                                                 
2023-06-23T12:05:58+02:00      DELETED    docker     59818fa298d961fb47c21bdf27202c50813dd6f46a8679e… angry_kirch8  

# goto another terminal
$ sudo systemctl start docker && sudo systemctl start docker.socket
$ docker run -it busybox

Comment (burak-ok) [reply pkg/container-utils/containerd/containerd.go:48]: > You mean it working not working for future containers ? For me it works fine for docker:

Yes i meant not working for future containers. But apparently it works for some cases (like you tested it with a stopped docker service.
Whereas I tested it with renaming `/run/docker.sock` to `/run/docker.sock.foo`. Starting `ig` and while it runs renaming `/run/docker.sock.foo` back to /run/docker.sock`

Comment (burak-ok) [reply pkg/container-utils/containerd/containerd.go]: Thanks to @mqasimsarfraz the behavior is now the same as before and all other clients.
When creating our own `grpc` connection and giving it to the `containerd` client the connection seems to be lazily initialized.

File: pkg/container-utils/containerd/containerd.go [hunk 27]
```diff
@@ -15,31 +15,293 @@
...
 	if socketPath == "" {
 		socketPath = runtimeclient.ContainerdDefaultSocketPath
 	}
 
-	criClient, err := criclient.NewCRIClient(runtimeclient.ContainerdName, socketPath, DefaultTimeout)
+	dialCtx, cancelFunc := context.WithTimeout(context.TODO(), DefaultTimeout)
+	defer cancelFunc()
+	grpcConn, err := grpc.DialContext(
+		dialCtx,
+		"unix://"+socketPath,
```
mqasimsarfraz, line 56:
```suggestion
		"unix:"+socketPath,
```
How about using ["unix:"](https://github.com/grpc/grpc/blob/master/doc/naming.md?plain=1#L37) so it works for both absolute and relative paths like before?

mqasimsarfraz (approved):
LGTM! Just one one minor comment, tested that it works fine on non containerd hosts.

PR #1235, opened 2022-12-28
Add `pstree` like output to `snapshot process`

eiffel-fl:
Hi.

In this PR, I added `pstree` like output to `snapshot process`:

```bash
$ sudo ./local-gadget snapshot process
builder (default)
|-bash(155746)-+
               |-sleep(169590)
               |-bash(169595)-+
                              |-sleep(169596)
                              |-sleep(169598)
                              |-sleep(169597)
                              |-bash(169599)-+
                                             |-sleep(169602)
                                             |-sleep(169601)
                                             |-sleep(169600)
               |-sleep(169591)
               |-sleep(169592)
$ ./kubectl-gadget snapshot process -A -o tree
storage-provisioner (kube-system)
|-storage-provisi(2901)
gadget (gadget)
|-gadgettracerman(5315)
kube-controller-manager (kube-system)
|-kube-controller(1751)
kube-apiserver (kube-system)
|-kube-apiserver(1771)
kube-scheduler (kube-system)
|-kube-scheduler(1807)
etcd (kube-system)
|-etcd(1840)
kube-proxy (kube-system)
|-kube-proxy(2388)
coredns (kube-system)
|-coredns(2613)
```

The output is not exactly like `pstree` because I suppose in `pstree` they are doing some backtracking to continue one branch while printing children.
Also, I would like to get your opinion because this change breaks a bit the genericity of the `snapshot` command code.

Best regards.

### DIFF
--- a/pkg/gadgets/snapshot/process/tracer/gadget.go
+++ b/pkg/gadgets/snapshot/process/tracer/gadget.go
@@ -15,6 +15,9 @@
 package tracer
 
 import (
+	"fmt"
+	"strings"
+
 	gadgetregistry "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-registry"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/types"
@@ -62,6 +65,29 @@ func (g *GadgetDesc) Parser() parser.Parser {
 
 func (g *GadgetDesc) EventPrototype() any {
 	return &types.Event{}
+}
+
+func (g *GadgetDesc) OutputFormats() (gadgets.OutputFormats, string) {
+	return gadgets.OutputFormats{
+		"tree": gadgets.OutputFormat{
+			Name:        "Tree",
+			Description: "A pstree like output",
+			Transform: func(data any) ([]byte, error) {
+				processes, ok := data.([]*types.Event)
+				if !ok {
+					return nil, fmt.Errorf("type must be []*types.Event and is: %T", data)
+				}
+
+				var builder strings.Builder
+				err := types.WriteTree(&builder, processes)
+				if err != nil {
+					return nil, err
+				}
+
+				return []byte(builder.String()), nil
+			},
+		},
+	}, "columns"
 }
 
 func (g *GadgetDesc) SortByDefault() []string {


cmd/kubectl-gadget/snapshot/process.go
```diff
@@ -41,6 +41,10 @@ func newProcessCmd() *cobra.Command {
 			},
 		}
 
+		if processGadget.Parser.GetOutputConfig().OutputMode == "tree" {
+			processGadget.printer = commonsnapshot.PrintTree
+		}
```
blanquicet, line 46: Implementation is repeated in kubectl-gadget and local-gadget. I tried this https://github.com/inspektor-gadget/inspektor-gadget/commit/ec447a3957a758a9e64f111c9d3aa2bc839fcb23 to make it more generic. WDYT?

cmd/kubectl-gadget/utils/flags.go
```diff
@@ -131,7 +132,8 @@ func AddCommonFlags(command *cobra.Command, params *CommonFlags) {
 		}
 
 		// Output Mode
-		if err := params.ParseOutputConfig(); err != nil {
+		err := params.ParseOutputConfig()
+		if err != nil && command.Name() != "process" && !errors.Is(err, errors.New("\"tree\" output mode is not supported")) {
```
blanquicet, line 136: IMO, it is too specific for one gadget to have it here, and it is repeated for kubectl-gadget and local-gadget. I tried this https://github.com/inspektor-gadget/inspektor-gadget/commit/a69bc1fe60fb7c571283ff64ff3d3afcc407b16a to make it more generic. WDYT?

blanquicet (commented):
I made some experiments to avoid the duplicated code between kubectl-gadget and local-gadget, and avoid having gadget-specific code around. Check my comments and let me know what you think.

I still have to check the actual logic to print the tree 😅 .

cmd/common/snapshot/process.go
```diff
@@ -35,6 +38,85 @@ type ProcessParser struct {
...
+	front += strings.Repeat(" ", len(line)+1)
+	fmt.Println(printedLine)
+
+	for _, child := range node.children {
+		printNode(child, front)
+	}
+}
+
+func printTree(tree *nodeEvent) {
+	fmt.Printf("%s (%s)\n", tree.event.Container, tree.event.Namespace)
```
blanquicet, line 98: IMO, we should use a different format for the container information from the PID information. I suggest changing `container (namespace)` to `container/pod/namespace` and only printing the `/pod` and `/namespace` if available. In this way, it will be compatible with non-Kubernetes containers in local-gadget.

cmd/common/snapshot/process.go
```diff
@@ -35,6 +38,85 @@ type ProcessParser struct {
...
 }
 
+type nodeEvent struct {
+	event    *types.Event
+	children []*nodeEvent
+	parent   *nodeEvent
+}
+
+func createTree(processes []*types.Event) *nodeEvent {
+	nodes := make(map[int]*nodeEvent, 0)
```
blanquicet, line 48: We can preallocate this as we already know how many nodes we will create.

cmd/common/snapshot/process.go
```diff
@@ -35,6 +38,85 @@ type ProcessParser struct {
 	outputConfig *commonutils.OutputConfig
 }
 
+type nodeEvent struct {
```
blanquicet, line 41:
```suggestion
type nodeProcess struct {
```

cmd/common/snapshot/process.go
```diff
@@ -35,6 +38,85 @@ type ProcessParser struct {
...
+
+	for _, child := range node.children {
+		printNode(child, front)
+	}
+}
+
+func printTree(tree *nodeEvent) {
+	fmt.Printf("%s (%s)\n", tree.event.Container, tree.event.Namespace)
+	printNode(tree, "")
+}
```
blanquicet, line 100: I don't see the need of having `printTree` and `printNode`. They can be one single function by moving the container to `PrintTree`.

Also because `printNode` is not accurate, it is actually printing the full tree after all the recursion and not a node.

cmd/common/snapshot/process.go
```diff
@@ -35,6 +38,85 @@ type ProcessParser struct {
...
+	}
+
+	// Find the root, i.e. find the node without parent.
+	for _, node := range nodes {
+		if node.parent == nil {
+			return node
+		}
+	}
+
+	return nil
```
blanquicet, line 75: Shouldn't this be an error? How couldn't a root exist?

blanquicet, line 75: I think all this can be simplified:

```suggestion
	// Link all nodes together.
	var root *nodeEvent
	for _, node := range nodes {
		ppid := node.event.ParentPid
		if _, ok := nodes[ppid]; !ok {
			root = node
			continue
		}
		nodes[ppid].children = append(nodes[ppid].children, node)
	}

	if root == nil {
		// TODO: return an error instead of panicking.
		panic("container without root process")
	}

	return root
```

BTW, we wouldn't need the `parent` field in `nodeEvent`.

Did I miss something?

blanquicet (commented):
Looks really nice already! I think it is important to add some unit tests for the functions that convert the process into a tree.

cmd/common/snapshot/process.go
```diff
@@ -35,6 +40,91 @@ type ProcessParser struct {
...
 
+type nodeProcess struct {
+	event    *types.Event
+	children []*nodeProcess
+}
+
+func createTree(processes []*types.Event) (*nodeProcess, error) {
+	var root *nodeProcess
+
+	nodes := make(map[int]*nodeProcess, 0)
```
blanquicet, line 51:
```suggestion
	nodes := make(map[int]*nodeProcess, len(processes))
```

cmd/common/snapshot/process.go
```diff
@@ -35,6 +40,91 @@ type ProcessParser struct {
...
+	if root == nil {
+		return nil, fmt.Errorf("container has no root process")
+	}
+
+	return root, nil
+}
+
+func treeToString(node *nodeProcess, front string) string {
+	line := fmt.Sprintf("|-%s(%d)", node.event.Command, node.event.Pid)
+	completeLine := front + line
```
blanquicet, line 81:
```suggestion
	completeTree := front + line
```

cmd/common/snapshot/process.go
```diff
@@ -35,6 +40,91 @@ type ProcessParser struct {
...
+	return completeLine
+}
+
+func PrintTree(processes []*types.Event) error {
+	containers := make(map[string][]*types.Event, len(processes))
+	for _, process := range processes {
+		_, ok := containers[process.Container]
+		if !ok {
+			containers[process.Container] = make([]*types.Event, 0)
+		}
```
blanquicet, line 103:
```suggestion
		if _, ok := containers[process.Container]; !ok {
			containers[process.Container] = make([]*types.Event, 0)
		}
```

cmd/common/snapshot/process.go
```diff
@@ -35,6 +40,91 @@ type ProcessParser struct {
...
+		containers[process.Container] = append(containers[process.Container], process)
+	}
+
+	for _, container := range containers {
+		tree, err := createTree(container)
+
```
blanquicet, line 109:
```suggestion
```

cmd/common/snapshot/process.go
```diff
@@ -23,6 +26,8 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/types"
 )
 
+const OutputModeTree = "tree"
```
blanquicet, line 29: nit: Maybe it is more accurate

```suggestion
const OutputModeProcessTree = "tree"
```

blanquicet (commented):
It is getting ready 🚀 ... Some quick comments, I will take another look later when the tests will be ready

cmd/common/snapshot/process.go
blanquicet: Why they are different semantically if the second line is checking the error that was just generated in the first one?

cmd/common/snapshot/process_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+type stsTestCase struct {
+	val      uint64
+	valMax   uint64
+	width    uint64
+	expected string
+}
+
+func TestCreateTree(t *testing.T) {
```
blanquicet, line 33: Maybe add another test where `createTree` cannot find the root event, something where all processes are children of processes in the list.

cmd/common/snapshot/process_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	_, err = createTree(nil)
+	if err == nil {
+		t.Fatalf("error is nil while it was expected when creating empty tree")
+	}
+}
+
+func TestTreeToString(t *testing.T) {
+	nodes := make([]*nodeProcess, 5)
```
blanquicet, line 117: `t.Parallel()`?

cmd/common/snapshot/process_test.go
```diff
@@ -0,0 +1,152 @@
...
+	nodes[2].children = append(nodes[2].children, nodes[3])
+	nodes[2].children = append(nodes[2].children, nodes[4])
+
+	i = 0
+	treeString := treeToString(nodes[0], "")
+	scanner := bufio.NewScanner(strings.NewReader(treeString))
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		expected := fmt.Sprintf("%s(%d)", nodes[i].event.Command, nodes[i].event.Pid)
```
blanquicet, line 145: This test relies on the fact that the above appends will add the elements in order. And, as far as I know, it is not guaranteed, is it?

blanquicet (commented):
LGTM, just one concern about the tests.

cmd/common/snapshot/process_test.go
blanquicet: Aaahh you are right, it is an array. Order is not guaranteed when we add elements to a map. Sorry, never mind.

cmd/common/snapshot/process.go
```diff
@@ -35,6 +40,90 @@ type ProcessParser struct {
...
+		nodes[process.Pid] = &nodeProcess{
+			event:    process,
+			children: make([]*nodeProcess, 0),
+		}
+	}
+
+	// Link all nodes together.
+	for _, node := range nodes {
+		ppid := node.event.ParentPid
+		if _, ok := nodes[ppid]; !ok {
```
flyth, line 63: Is it safe to assume that there will be exactly one process that doesn't have an entry for its parent - if the iterator also works by iterating over a process tree, everything should be fine, but I don't know if that's the case.

flyth (commented):
Really cool addition, improves the usability of the gadget a lot!

cmd/common/snapshot/process.go
```diff
@@ -35,6 +40,90 @@ type ProcessParser struct {
 	outputConfig *commonutils.OutputConfig
 }
 
+type nodeProcess struct {
+	event    *types.Event
+	children []*nodeProcess
+}
+
+func createTree(processes []*types.Event) (*nodeProcess, error) {
```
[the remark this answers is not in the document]
flyth: Yeah, either tracer or types of the gadget - I think I'd rather go with types, as the tracer has more dependencies that might not be needed (e.g. for kubectl gadget) and the types are already included anyways.

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -32,3 +35,91 @@ type Event struct {
...
+	if len(node.children) == 0 {
+		return
+	}
+
+	for _, child := range node.children {
+		treeToStringBuilder(child, builder, depth+1)
+	}
+}
+
+func PrintTree(processes []*Event) error {
```
flyth, line 97: How about having an `io.Writer` parameter here and `fmt.Fprintf` to it instead of writing to stdout directly. Instead, `os.Stdout` can be given as argument in the `cmd/` packages calling PrintTree. That way third parties (and ourselves) can output the tree by other means.

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -32,3 +35,91 @@ type Event struct {
...
+			children: make([]*nodeProcess, 0),
+		}
+	}
+
+	// Link all nodes together.
+	for _, node := range nodes {
+		ppid := node.event.ParentPid
+		if _, ok := nodes[ppid]; !ok {
+			if root != nil {
+				return nil, fmt.Errorf("container has two root process: %v and %v", root, node)
```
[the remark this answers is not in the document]
flyth: Your proposal would still just show an error which might confuse the user. If (and I'm still not sure whether that can happen at all) there were orphaned entries, I would just have them appear next to the "real" root. That way we wouldn't drop processes from the output and also not just error out.

Something like this:
```
Container 1
	Real Root Process without Parent
		- Process #1
		- Process #2
			- Subprocess #2.1
	Orphan#1 (another process without parent)
		- Subprocess Orphan#1
	...
```

The goal is just to not confuse the user while also giving the information we have in the best possible way. "there are several roots" is pretty vague and users could misinterpret it as if it was their fault (not giving correct settings or whatever), when in fact it is only a small (and probably temporary) inconsistency in the data, that we could handle like I described above.

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -32,3 +37,100 @@ type Event struct {
...
+	treeToStringBuilder(node, &builder, 0)
+
+	return builder.String()
+}
+
+func treeToStringBuilder(node *nodeProcess, builder *strings.Builder, depth int) {
+	fmt.Fprintf(builder, "%s|-%s(%d)\n", strings.Repeat("\t", depth), node.event.Command, node.event.Pid)
+	if len(node.children) == 0 {
+		return
+	}
```
mauriciovasquezbernal, line 97: This is not needed. The for loop on next line will not iterate if the slice is empty.

pkg/gadgets/snapshot/process/types/process-collector_test.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	_, err := createTree(events)
+	if err == nil {
+		t.Fatalf("error is nil while it was expected when creating tree with several roots")
+	}
+}
+
+func TestTreeToString(t *testing.T) {
+	t.Parallel()
```
mauriciovasquezbernal, line 153: Any reason why it's not used in other tests?

mauriciovasquezbernal (commented):
Some minor comments but looks very good!

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -36,3 +41,96 @@ type Event struct {
...
+
+func createTree(processes []*Event) (*processTree, error) {
+	var root *processTree
+
+	nodes := make(map[int]*processTree, len(processes))
+	// Create a node for each process.
+	for _, process := range processes {
+		nodes[process.Pid] = &processTree{
+			process:  process,
+			children: make([]*processTree, 0),
```
mauriciovasquezbernal, line 58: nit: I think this is not needed as `append(nil, foo)` works fine in golang.

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -36,3 +41,96 @@ type Event struct {
...
+		// https://elixir.bootlin.com/linux/v6.1.3/source/kernel/exit.c#L653
+		// Note that above code is both called when calling exit() system call
+		// and when receiving a fatal signal.
+		return nil, fmt.Errorf("container has no root process")
+	}
+
+	return root, nil
+}
+
+func treeToString(node *processTree) string {
```
mauriciovasquezbernal, line 90: Define this as `func (t *processTree) String()` to implement the [Stringer](https://go.dev/tour/methods/17) interface?

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -36,3 +41,96 @@ type Event struct {
...
+	var builder strings.Builder
+	treeToStringBuilder(node, &builder, 0)
+
+	return builder.String()
+}
+
+func treeToStringBuilder(node *processTree, builder *strings.Builder, depth int) {
+	fmt.Fprintf(builder, "%s|-%s(%d)\n", strings.Repeat("\t", depth), node.process.Command, node.process.Pid)
```
mauriciovasquezbernal, line 98: nit: In my system pstree uses 8 spaces instead of a tab. Should we replicate the same?

pkg/gadgets/snapshot/process/types/process-collector.go
```diff
@@ -36,3 +41,96 @@ type Event struct {
...
+	}
+}
+
+func WriteTree(output io.Writer, processes []*Event) error {
+	containers := make(map[string][]*Event, len(processes))
+	for _, process := range processes {
+		_, ok := containers[process.Container]
+		if !ok {
+			containers[process.Container] = make([]*Event, 0)
+		}
```
mauriciovasquezbernal, line 110: I think these lines are not needed. The line below handles well the case when `containers[process.Container]` is nil.

pkg/gadgets/snapshot/process/types/process-collector_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+	tree, err := createTree(events)
+	if err != nil {
+		t.Fatalf("fail to create tree: %v", err)
+	}
+
+	expectedPid := 42
+	if tree.process.Pid != expectedPid {
+		t.Fatalf("root PID is wrong, expected %d got %d", expectedPid, tree.process.Pid)
+	}
```
mauriciovasquezbernal, line 64: I know you created this PR before we started using `github.com/stretchr/testify`, however it could improve these tests a lot. 

`require.Equal(t, 42, tree.process.Pid, "root PID is wrong")`

pkg/gadgets/snapshot/process/types/process-collector_test.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+type stsTestCase struct {
+	val      uint64
+	valMax   uint64
+	width    uint64
+	expected string
+}
+
+func TestCreateTree(t *testing.T) {
```
mauriciovasquezbernal, line 31: nit: An alternative could be to test the whole created tree. 

```golang
func TestCreateTree(t *testing.T) {
	t.Parallel()

	pid42 := Event{
		Pid: 42,
	}

	pid43 := Event{
		Pid:       43,
		ParentPid: 42,
	}

	pid44 := Event{
		Pid:       44,
		ParentPid: 42,
	}

	pid45 := Event{
		Pid:       45,
		ParentPid: 42,
	}

	pid46 := Event{
		Pid:       46,
		ParentPid: 43,
	}

	events := []*Event{
		&pid42,
		&pid43,
		&pid44,
		&pid45,
		&pid46,
	}

	tree, err := createTree(events)
	if err != nil {
		t.Fatalf("fail to create tree: %v", err)
	}

	expectedTree := &processTree{
		process: &pid42,
		children: []*processTree{
			{
				process: &pid43,
				children: []*processTree{
					{
						process:  &pid46,
						children: []*processTree{},
					},
				},
			},
			{
				process:  &pid44,
				children: []*processTree{},
			},
			{
				process:  &pid45,
				children: []*processTree{},
			},
		},
	}

	require.Equal(t, expectedTree, tree, "tree is wrong")
}
```

The disadvantage is that the provided information is less verbose, but IMO the test is much more readable.

mauriciovasquezbernal (commented):
Nice to see this rebased!

I think commits `pkg/gadgets: Add functions to print processes like pstree.` and `pkg/gadgets: Add tests for functions generating pstree like output.` can be squashed. 

Some minor comments around, but IMO it's almost ready to go.

pkg/gadgets/interface.go
```diff
@@ -78,13 +78,24 @@ type GadgetDescSkipParams interface {
...
 type OutputFormat struct {
-	Name        string `json:"name"`
-	Description string `json:"description"`
-	Transform   func([]byte) ([]byte, error)
+	Name                   string                       `json:"name"`
+	Description            string                       `json:"description"`
+	TransformType          TransformType                `json:"transformType"`
+	RequiresCombinedResult bool                         `json:"requiresCombinedResult"`
+	Transform              func([]byte) ([]byte, error) `json:"-"`
+	TransformArray         func(any) ([]byte, error)    `json:"-"`
```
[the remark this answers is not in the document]
flyth: You are right - we could reuse the `func(any) ([]byte, error)` for all cases and just cast to whatever type is appropriate (the TransformType should make sure that it's used where it's supposed to).

cmd/common/registry.go
```diff
@@ -318,7 +318,30 @@ func buildCommandFromGadget(
...
+					return fmt.Errorf("invalid output mode %q", outputModeName)
+				}
+
+				format := formats[outputModeName]
+
+				if format.RequiresCombinedResult {
+					parser.EnableCombiner()
+				}
+
+				transformResult := format.TransformArray
```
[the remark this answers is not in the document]
flyth: Yeah, there should be checks in `registry.go` to check which transform could be used at certain places (result, single events or an array of events) - right now we only use result + array of events, though.

pkg/gadgets/snapshot/process/tracer/gadget.go
```diff
@@ -64,6 +66,26 @@ func (g *GadgetDesc) EventPrototype() any {
 	return &types.Event{}
 }
 
+func (g *GadgetDesc) OutputFormats() (gadgets.OutputFormats, string) {
+	return gadgets.OutputFormats{
+		"tree": gadgets.OutputFormat{
+			Name:        "Tree",
+			Description: "A pstree like output",
+			Transform: func(data any) ([]byte, error) {
+				processes := data.([]*types.Event)
```
flyth, line 75: Could you please add a type check here as well and return an error, if it's not `[]*types.Event`?

flyth (approved):
Looks great, thanks for unifying the transformers as well!

I just found one small thing, would be great if you could address that as well before merging.

PR #1236, opened 2022-12-30
kubectl-gadget: Check kubernetes.io/os label only if --print-only was not given

eiffel-fl:
```
In commit b8fc17b561d2 ("kubectl-gadget: Only support Linux nodes explicitly"), we added an os label to ensure Inspektor Gadget is only deployed on Linux node. Sadly, this label is not available to all k8s version, so for older version we remove this label.
But, to check the kubernetes version we need to communicate with remove cluster, this leaded to error when running kubectl gadget deploy --print-only was not all the YAML would be printed.

This commits removes the k8s version check is --print-only was given.
```

### DIFF
--- a/cmd/kubectl-gadget/deploy_test.go
+++ b/cmd/kubectl-gadget/deploy_test.go
@@ -1,0 +1,32 @@
+// Copyright 2019-2021 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"testing"
+)
+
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/deploy.go
```diff
@@ -376,17 +376,19 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
 			daemonSet.Spec.Template.Annotations["inspektor-gadget.kinvolk.io/option-hook-mode"] = hookMode
 
-			// The "kubernetes.io/os" node label was introduced in v1.14.0
-			// (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md.)
-			// Remove this if the cluster is older than that to allow Inspektor Gadget to work there.
-			serverInfo, err := discoveryClient.ServerVersion()
-			if err != nil {
-				return fmt.Errorf("getting server version: %w", err)
-			}
+			if !printOnly {
```
mqasimsarfraz, line 379: nit: just wondering if adding a unit test as following makes sense here?

```
func TestPrintOnly(t *testing.T) {
	cmd := rootCmd
	cmd.SetArgs([]string{"deploy", "--print-only"})

	var stdErr bytes.Buffer
	cmd.SetErr(&stdErr)
	cmd.Execute()
	if stdErr.Len() != 0 {
		t.Fatalf("Error while running command: %s", stdErr.String())
	}
}
```

mqasimsarfraz (approved):
LGTM!

mqasimsarfraz (approved):

PR #1242, opened 2023-01-04
cmd/snapshot: Remove unnecessary pointer to slice

burak-ok:
# cmd/snapshot: Remove unnecessary pointer to slice

This was needed since the filtering logic for the thread ids in `snapshot process` was on the client side and the slice had to get modified.

With https://github.com/inspektor-gadget/inspektor-gadget/pull/1223 this changes and we can remove the pointer to the slice making the code more readable and maintainable

### DIFF
--- a/cmd/common/snapshot/process.go
+++ b/cmd/common/snapshot/process.go
@@ -70,8 +70,8 @@ func (p *ProcessParser) GetOutputConfig() *commonutils.OutputConfig {
 	return p.outputConfig
 }
 
-func (p *ProcessParser) SortEvents(allProcesses *[]*types.Event) {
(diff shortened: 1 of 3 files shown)


mauriciovasquezbernal:
> I am not sure it is unnecessary. Indeed, without pointer, a copy of the slice header will be done and using the pointer avoids this. On another hand, we can indeed avoid using pointer as we do not append data to slices in these cases.

Sort functions in the standard library (like [sort.Strings](https://pkg.go.dev/sort#Strings)) don't use a pointer, hence I think we should follow the same.

mauriciovasquezbernal (approved):
LGTM, thanks!

Issue #1011, opened 2022-10-05
docs: Update architecture document


mauriciovasquezbernal:
Our [architecture](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/architecture.md) document is outdated. It includes references to things like bcc-wrapper.sh, OCI hooks and other that are not relevant anymore.

PR #1245, opened 2023-01-04
Modernize architecture figures

eiffel-fl:
Hi.

I needed to use this figures and I modernize them so they are more accurate and represent better the current architecture.

Best regards.


mauriciovasquezbernal:
> I can expand this one to fix https://github.com/inspektor-gadget/inspektor-gadget/issues/1011.

Great!

flyth:
I think those figures are a bit confusing as they mix up the deployment of IG and running of gadgets a bit. I understand it a bit like that whenever a gadget should run, it's being deployed on the server as daemonset and then directly run (and then undeployed) - but that might be just me.

Regarding the layout/styling: the fonts used in the original are [Overpass](https://www.fontsquirrel.com/fonts/overpass) and [Overpass Mono](https://www.fontsquirrel.com/fonts/overpass-mono). I'd also add some padding to the red "gadget pod" box, lighten up the box around kernel or change the text to white to get more contrast and also darken the cloud a bit (maybe more blue-ish), as it's hardly visible on a white background right now.

docs/images/architecture/k8s_deploy.svg
```diff
@@ -0,0 +1,369 @@
...
+    <text
+       xml:space="preserve"
+       style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:18.66666603px;line-height:1.25;font-family:'Overpass Mono';-inkscape-font-specification:'Overpass Mono';letter-spacing:0px;word-spacing:0px;fill:#000000;fill-opacity:1;stroke:none"
+       x="139.17116"
+       y="353.49573"
+       id="text879"><tspan
+         sodipodi:role="line"
+         id="tspan877"
+         x="139.17116"
+         y="353.49573">$ kubectl gadget deploy</tspan></text>
```
mauriciovasquezbernal, line 367: It seems to be this font is different than the one in "kubectl-gadget". Also, why is `kubectl-gadget` an image and not a text element?

docs/architecture.md
```diff
@@ -10,19 +10,22 @@ events related to specific kernel subsystems.
...
 
-On Kubernetes, Inspektor Gadget is deployed to each node as a privileged
-DaemonSet.
-It uses in-kernel eBPF helper programs to monitor events from userspace programs
-in a pod. The eBPF programs are run by the kernel and gather the log data.
-Inspektor Gadget's userspace utilities fetch the log data from ring buffers
-and display it. What eBPF programs are and how Inspektor Gadget uses them is
-briefly explained here.
+Before using Inspektor Gadget, it first need to be deployed to each node as a
+privileged `DaemonSet`.
```
mauriciovasquezbernal, line 14: I think we should not tell this as a "story" given that this document is not a guide. 

```suggestion
On Kubernetes, Inspektor Gadget is deployed to each node as a DaemonSet.
```

It can be extended with a bit more of information of how it's deployed:

"It's usually deployed by the kubectl gadget plugin. It interacts with the kube-api server to create all the resources needed for Inspektor Gadget to work. Another way of deploying Inspektor Gadget is to directly apply the [manifests](pkg/resources/manifests/deploy.yaml) using kubectl".

mauriciovasquezbernal (commented):
I made some comments about this PR. I think it's an improvement of the current situation, however I think there is still a lot of work to be done in this document. I think the document should describe the following:

- General architecture
  - Different components involved 
- Filtering mechanism 
- Trace CR
  - Describe how information is sent back to user
- Tracer architecture
  - perf ring buffer, filtering logic, etc.  
- Container hooks
- ...


However, I'd wait for @flyth work before updating it more. Regarding this PR, I think we can merge it once you handle some of the comments I made. Otherwise, if you think it doesn't make sense to keep updating it, it's also fine for me, we can do all the work later on.

docs/images/architecture/k8s_deploy.svg
flyth: It's actually important to either embed the font or have the text elements converted to paths, because otherwise users that don't have the font installed will see the text rendered by some standard font that looks totally different.

For me right now it looks like this:
Rendered SVG

mauriciovasquezbernal (approved):
LGTM. Thanks for working on this.

PR #1246, opened 2023-01-05
cmd/local-gadget: Add global flag `--timeout`

burak-ok:
# cmd/local-gadget: Add global flag `--timeout`

This PR adds a new flag called `--timeout` to `local-gadget`.

A helper function was introduced which waits for `SIGTERM` or `SIGINT` or an elapsed timeout in order to reduce code duplication. Of course there are special cases like for `advise seccomp-profile`

### Questions/Comments/...

1. Naming things is still not easy for me :smile: Suggestions welcome, be it for the function name or filename
1. Since now every gadget in `local-gadget` has the `--timeout` flag available, it has needed to "hide" it in some gadgets like `snapshot`. Is this a good way to solve that?
1. Should i create another following PR which unifies this `--timeout` flag, since now `kubectl-gadget` and `local-gadget` has it? There is currently a Flagset which is shared called `OutputConfig`. `--timeout` would not fit in there -> Rename the Flagset? Create a new one? WDYT?
1. Rewrite some of the `local-gadget` tests to use `--timeout` instead of setting `startAndStop` to true? Not all since we want to test both paths

### DIFF
--- a/cmd/local-gadget/advise/seccomp-profile.go
+++ b/cmd/local-gadget/advise/seccomp-profile.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -98,9 +99,15 @@ func newSeccompProfileCmd() *cobra.Command {
 		stop := make(chan os.Signal, 1)
 		signal.Notify(stop, syscall.SIGINT, syscall.SIGTERM)
 
+		timeoutChannel := make(<-chan time.Time)
+		if commonFlags.Timeout != 0 {
+			timeoutChannel = time.After(time.Duration(commonFlags.Timeout) * time.Second)
+		}
+
 		// The profile can be generated upon container termination or when user
-		// requests it through a signal.
+		// requests it through a signal or on a timeout
 		select {
+		case <-timeoutChannel:
 		case <-stop:
 		case err := <-containerNotifier:
 			if err != nil {
--- a/cmd/local-gadget/snapshot/socket.go
+++ b/cmd/local-gadget/snapshot/socket.go
@@ -94,6 +94,7 @@ func newSocketCmd() *cobra.Command {
 	cmd := commonsnapshot.NewSocketCmd(runCmd, &flags)
 
 	utils.AddCommonFlags(cmd, &commonFlags)
+	utils.HideFlagTimeout(cmd)
 
 	return cmd
 }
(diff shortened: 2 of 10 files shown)


cmd/local-gadget/snapshot/socket.go
```diff
@@ -94,6 +94,7 @@ func newSocketCmd() *cobra.Command {
 	cmd := commonsnapshot.NewSocketCmd(runCmd, &flags)
 
 	utils.AddCommonFlags(cmd, &commonFlags)
+	utils.HideFlagTimeout(cmd)
```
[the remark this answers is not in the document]
burak-ok: Unfortunately this is not possible :(. I need to have the `cmd` instance of `snapshot process`

blanquicet (approved):
LGTM!

cmd/local-gadget/utils/utils.go
```diff
@@ -0,0 +1,43 @@
...
+// is sent to us
+func WaitForEnd(f *CommonFlags) {
+	timeoutChannel := make(chan os.Signal, 1)
+	signal.Notify(timeoutChannel, os.Interrupt)
+	if f.Timeout != 0 {
+		go func() {
+			time.Sleep(time.Duration(f.Timeout) * time.Second)
+			timeoutChannel <- os.Interrupt
+		}()
+	}
```
mauriciovasquezbernal, line 34: Why do you also need to attach the signal to this channel?  

What about something like the following?

```suggestion
	timeoutChannel := make(chan struct{})
	if f.Timeout != 0 {
		go func() {
			time.Sleep(time.Duration(f.Timeout) * time.Second)
			close(timeoutChannel)
		}()
	}
```

mauriciovasquezbernal: Thinking better, we can use [time.After](https://pkg.go.dev/time#After) to avoid creating the goroutine and handling it outselves. What do you think?

```suggestion
	timeoutChannel := make(<-chan time.Time)
	if f.Timeout != 0 {
		timeoutChannel = time.After(time.Duration(f.Timeout) * time.Second)
	}
```

cmd/local-gadget/advise/seccomp-profile.go
```diff
@@ -98,9 +99,19 @@ func newSeccompProfileCmd() *cobra.Command {
...
 
+		timeoutChannel := make(chan os.Signal, 1)
+		signal.Notify(timeoutChannel, os.Interrupt)
+		if commonFlags.Timeout != 0 {
+			go func() {
+				time.Sleep(time.Duration(commonFlags.Timeout) * time.Second)
+				timeoutChannel <- os.Interrupt
+			}()
+		}
+
```
mauriciovasquezbernal, line 110: Please update if you accept my comment in the other part of the code.

mauriciovasquezbernal (approved):
LGTM. thanks!

PR #1249, opened 2023-01-06
integration: Use /bin/sleep rather than sleep.

eiffel-fl:
```
Busybox commit 58598eb70935 ("ash: optional sleep builtin") compiles by default sleep as a builtin [1].
This commit is part of release 1.36.0 which was released 3 days ago [2].

As a consequence, TestExecsnoop which awaits an events with command sleep was failing.
This commit forces using /bin/sleep rather than the builtin so our integration test passes.

[1] https://git.busybox.net/busybox/commit/?id=58598eb7093561d914a6254697e137b815f1fdfc [2] https://git.busybox.net/busybox/commit/?id=70f77e4617e06077231b8b63c3fb3406d7f8865d
```

### DIFF
--- a/integration/inspektor-gadget/integration_test.go
+++ b/integration/inspektor-gadget/integration_test.go
@@ -640,7 +640,7 @@ func TestExecsnoop(t *testing.T) {
 
 	t.Parallel()
 
-	cmd := "while true; do date ; sleep 0.1; done"
+	cmd := "while true; do date ; /bin/sleep 0.1; done"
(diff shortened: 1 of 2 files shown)


mqasimsarfraz (commented):
Thanks for a quick fix. We need to do this for [local-gadget](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/local-gadget/k8s/trace_exec_test.go#L29) as well.

mqasimsarfraz (approved):
LGTM!

PR #1255, opened 2023-01-10
cmd: Remove OutputModeColumns

### LINKED ISSUE #908: CLI: Combine `columns` and `custom-columns` output mode
## Current situation

After implementing the `parser` approach on all the gadgets (where it makes sense) and defining a default list of columns to be printed when `-o custom-columns=a,b,c` is not provided, it came out that [`OutputModeColumns`](https://github.com/kinvolk/inspektor-gadget/blob/3a31bca668f37a7c0cf04ee83a697f91b81446a1/cmd/common/utils/flags.go#L26) is not needed anymore.  

In fact, in all the places where `OutputModeColumns` is used, it is applied the same strategy as `OutputModeCustomColumns`. Examples:

- https://github.com/kinvolk/inspektor-gadget/blob/3a31bca668f37a7c0cf04ee83a697f91b81446a1/cmd/kubectl-gadget/trace/trace.go#L77-L80
- https://github.com/kinvolk/inspektor-gadget/blob/3a31bca668f37a7c0cf04ee83a697f91b81446a1/cmd/common/utils/parser.go#L111-L114
- https://github.com/kinvolk/inspektor-gadget/blob/3a31bca668f37a7c0cf04ee83a697f91b81446a1/cmd/common/snapshot/snapshot.go#L99-L101

## Impact

Still keeping `OutputModeColumns` creates confusion in the code as it is being used identically to `OutputModeCustomColumns`. 

## Ideal future situation

Remove `OutputModeColumns` option. Notice it has to be done in the common package so that the result is applied to `kubectl-gadget` and `local-gadget`.

burak-ok:
# Remove OutputModeColumns

This PR removes `OutputModeColumns` from the `SupportedOutputModes`.

As the default value of `--output` there is an empty string now. Internally it will get parsed as `OutputModeCustomColumns` with no `CustomColums` selected.
Achieving this combination is manually not possible, which means that `--output OutputModeCustomColums=` is not possible. This behavior is the same before this PR

Is this the way we should implement this or do we want a now option like `OutputModeDefault` (Which internally maps to `OutputModeCustomColumns` with no `CustomColums` again, etc...) or else?

resolves #908

### DIFF
--- a/cmd/common/profile/cpu.go
+++ b/cmd/common/profile/cpu.go
@@ -108,8 +108,6 @@ func (p *CPUParser) TransformReport(report *cpuTypes.Report) string {
 		}
 
 		return string(b)
-	case utils.OutputModeColumns:
-		fallthrough
 	case utils.OutputModeCustomColumns:
 		otherCols := p.TransformIntoColumns(report)
 		if p.CPUFlags.ProfileUserOnly {
--- a/cmd/common/snapshot/snapshot.go
+++ b/cmd/common/snapshot/snapshot.go
@@ -68,8 +68,6 @@ func (g *SnapshotGadgetPrinter[Event]) PrintEvents(allEvents []*Event) error {
 
 		fmt.Printf("%s\n", b)
 		return nil
-	case commonutils.OutputModeColumns:
-		fallthrough
 	case commonutils.OutputModeCustomColumns:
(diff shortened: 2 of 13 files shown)


mauriciovasquezbernal (approved):
I think the proposed solution is great. Thanks!

PR #1256, opened 2023-01-11
Add debug container

eiffel-fl:
Hi.

In this PR, I added a new Dockerfile which executes the `gadgettracermanager` with `delve`.
So, we can attach to the remote `gadgettracermanager` and debug it with `dlv`.

Best regards.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -124,6 +124,53 @@ $ export KUBECONFIG=... # not needed if valid config in $HOME/.kube/config
 $ make integration-tests
 ```
 
+### Debugging `gadgettracermanager`
+
+You can debug the `gadgettracermanager` running inside the container with the
+following commands:
+
+```bash
+# You will need to deploy the debug container inside the minikube cluster
+$ make minikube-deploy-debug
+...
+# Get the gadget pod name you want to attach
+$ kubectl get pod -n gadget
+NAME           READY   STATUS    RESTARTS   AGE
+gadget-p5bc7   1/1     Running   0          18s
+# You need to port forward the gadget pod port 4040 to any of your localhost
+# available port.
+$ kubectl port-forward -n gadget pod/gadget-p5bc7 8081:4040 &
+# You can now connect to remove delve
+$ dlv connect :8081
+Handling connection for 8081
+Type 'help' for list of commands.
+# You will need to configure the substitute path to be able to browse code.
+(dlv) config substitute-path /gadget /absolute/path/to/inspektor-gadget
+# Let's break when the trace signal gadget is run.
+(dlv) b github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/signal/tracer.run
+(dlv) c
+# In another terminal run ./kubectl-gadget trace signal -A.
+# Breakpoint is hit, you can now begin your debugging session.
+> github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/signal/tracer.(*Tracer).run() ./pkg/gadgets/trace/signal/tracer/tracer.go:203 (hits goroutine(412):1 total:1) (PC: 0x312f9f2)
+   198:         go t.run()
+   199:
+   200:         return nil
+   201: }
+   202:
+=> 203: func (t *Tracer) run() {
+   204:         for {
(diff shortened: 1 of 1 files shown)


blanquicet:
Maybe @johananl is interested in this 🙂

johananl:
Thanks, I'm aware of this @blanquicet.

Dockerfiles/gadget-default-debug.Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+		make btfgen BPFTOOL=/tmp/btfhub/tools/bin/bpftool.x86_64 \
+			BTFHUB_ARCHIVE=/tmp/btfhub-archive/ OUTPUT=/tmp/btfs/ -j$(nproc); \
+	fi
+
+# Main gadget image
+
+FROM bcc
+
+# This will be used by dlv server.
+ENV DEBUG_PORT=4040
```
alban, line 66: Instead of creating a new Dockerfile for the container with the debug feature enabled, could you reuse the main Dockerfile but add a `ARG DEBUG=false` and have an option in the Makefile to turn DEBUG=true?

Dockerfiles/gadget-default.Dockerfile
```diff
@@ -39,17 +42,36 @@ RUN cd /gadget/gadget-container && \
...
 	make -j$(nproc) TARGET_ARCH=${TARGETARCH} gadget-container-deps
 
+# Execute BTFGen
+RUN set -ex; \
+	if [ "$ENABLE_BTFGEN" = true ]; then \
+		cd /gadget && \
+		make btfgen BPFTOOL=/tmp/btfhub/tools/bin/bpftool.x86_64 \
+			BTFHUB_ARCHIVE=/tmp/btfhub-archive/ OUTPUT=/tmp/btfs/ -j$(nproc); \
+	fi
+
```
mauriciovasquezbernal, line 52: Why is this back? (It was moved to the host in https://github.com/inspektor-gadget/inspektor-gadget/pull/1228 to speed it up).

mauriciovasquezbernal:
Do we need to replicate all changes in `Dockerfiles/gadget-core.Dockerfile` too? (I know it's a PITA to have multiple Dockerfiles but I didn't find a better solution).

Dockerfiles/gadget-default.Dockerfile
```diff
@@ -43,13 +48,24 @@ RUN cd /gadget/gadget-container && \
...
 RUN set -ex; \
 	export DEBIAN_FRONTEND=noninteractive; \
 	apt-get update && \
 	apt-get install -y --no-install-recommends \
 		ca-certificates jq wget xz-utils binutils && \
 		rmdir /usr/src && ln -sf /host/usr/src /usr/src && \
-		rm -f /etc/localtime && ln -sf /host/etc/localtime /etc/localtime
+		rm -f /etc/localtime && ln -sf /host/etc/localtime /etc/localtime; \
+	if [ ${DEBUG_PORT} ]; then \
```
mqasimsarfraz, line 62: We can also have a look at the idea of  [`target`](https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage). Moving these in the end as `dev` stage and if target isn't configured we build a debug image and when we specify a target like `prod` we stop without doing any debug stuff e.g I got this idea from [here](https://github.com/cilium/cilium/blob/master/images/cilium/Dockerfile#L75). Just an idea, what do you think about it?

Makefile
```diff
@@ -137,6 +140,10 @@ gadget-%-container:
 push-gadget-%-container:
 	docker push $(CONTAINER_REPO):$(IMAGE_TAG)$(if $(findstring core,$*),-core,)
 
+gadget-%-container-debug:
+	docker buildx build -t $(CONTAINER_REPO):$(IMAGE_TAG)$(if $(findstring core,$*),-core,) --build-arg DEBUG=1 --build-arg DEBUG_PORT=$(DEBUG_PORT) -f Dockerfiles/gadget-$*.Dockerfile \
```
alban, line 144: In my opinion, the choice of `$DEBUG_PORT` should not be at build-time but at run-time. What about removing `--build-arg DEBUG_PORT` here and let the user add it in the yaml if they want to change the default from 4040 to something else?

PR #1225, opened 2022-12-22
Remove `io/ioutil` from `snapshot process` tracer

eiffel-fl:
Hi.

This PR removes `io/ioutil` as this package is deprecated.

Best regards.

PR #1258, opened 2023-01-12
Remove deprecated ioutil.ReadFile

alban:
Symptoms: 'make lint' gives the following:

> pkg/gadgets/snapshot/process/tracer/tracer.go:21:2: SA1019: "io/ioutil"
> has been deprecated since Go 1.16: As of Go 1.16, the same functionality
> is now provided by package io or package os, and those implementations
> should be preferred in new code. See the specific function documentation
> for details. (staticcheck)
> 	"io/ioutil"
> 	^

It seems to have been fixed by #1225 but it came back.


mauriciovasquezbernal:
I really don't understand the linter, why it's not complaining on PRs?

PR #1262, opened 2023-01-12
timestamp column: set maximum width

alban:
This is not using the 'fixed' keyword so that the width can be reduced if necessary.

Example of output:
```
$ ./kubectl-gadget trace open -A -o custom-columns=timestamp,comm,path
TIMESTAMP                           COMM             PATH
2023-01-12T16:31:47.311311188+01:00 kube-apiserver   /var/lib/minikube/certs/apiserver-kubelet-client.crt
2023-01-12T16:31:47.311342955+01:00 kube-apiserver   /var/lib/minikube/certs/apiserver-kubelet-client.key
```

With this patch, there is more space for the path.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -28,7 +28,7 @@ var node string
 
 func init() {
 	// Register column templates
-	columns.MustRegisterTemplate("timestamp", "width:35,hide")
+	columns.MustRegisterTemplate("timestamp", "width:35,maxWidth:35,hide")
 	columns.MustRegisterTemplate("node", "width:30,ellipsis:middle")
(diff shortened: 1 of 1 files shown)


flyth (approved):
LGTM!

However: do we really need this kind of precision for the timestamp? I guess in most cases HH:MM:SS.sss would be more than enough and would save quite some space. If we don't want to change that globally, maybe we should add an option for it (need to think about how and where to place that, though).

alban:
That could be a virtual column called "time" that extracts its data from the same field.

flyth:
> That could be a virtual column called "time" that extracts its data from the same field.

Yeah, good idea, that's much easier!

PR #1264, opened 2023-01-13
gadgets: Fix networking gadgets on old kernels

mauriciovasquezbernal:
There is a missing call to FixBpfKtimeGetBootNs() in networking gadgets to enable them to work in old kernels.

Fixes: ffb7d9910f2d ("gadgets: get timestamp from userspace on Linux <5.7")
Fixes: 7982f833d3ba ("trace sni gadget: add timestamps")

---
[Integration tests on AKS (Ubuntu, amd64)](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3906312046/jobs/6674433125#logs) were failing because that uses an old kernel.

### DIFF
--- a/pkg/gadgets/internal/networktracer/tracer.go
+++ b/pkg/gadgets/internal/networktracer/tracer.go
@@ -119,6 +119,8 @@ func NewTracer[Event any](
 	baseEvent func(ev types.Event) *Event,
 	parseEvent func([]byte) (*Event, error),
 ) *Tracer[Event] {
+	gadgets.FixBpfKtimeGetBootNs(spec.Programs)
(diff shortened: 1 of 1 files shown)


alban:
The fix can be done in NewTracer so it's not executed for each pod but only once.

mauriciovasquezbernal:
> The fix can be done in NewTracer so it's not executed for each pod but only once.

Great, updated!

mauriciovasquezbernal:
I missed trace network on this PR. Opened a new one -> https://github.com/inspektor-gadget/inspektor-gadget/pull/1266.

PR #1270, opened 2023-01-16
CI: Do not truncate failure logs

mqasimsarfraz:
It is better to skip some logs then to truncate all of them. In this PR we skip the logs if job summary can't handle but by default we will retain all the logs. 

- [sample report](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/3929358173/attempts/1#summary-10671123474). (skipped test `TestTraceFsslower`)

Should help with the [cases](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3912844172/attempts/1#summary-10634393081) like here.

### DIFF
--- a/tools/testjson2md/testjson2md.go
+++ b/tools/testjson2md/testjson2md.go
@@ -28,12 +28,17 @@ import (
 	"github.com/medyagh/gopogh/pkg/report"
 )
 
+const (
+	// summaryLimitInBytes is the maximum size of the summary in bytes.
+	// https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#step-isolation-and-limits
+	summaryLimitInBytes = 1000000
+)
+
 var ErrInvalidContent = fmt.Errorf("invalid content")
 
 var (
-	inPath        = flag.String("in", "", "path to JSON file produced by go tool test2json")
-	outPath       = flag.String("out", "", "path to output file")
-	truncateLines = flag.Int("truncate-lines", 50, "number of failed test log lines to truncate to")
+	inPath  = flag.String("in", "", "path to JSON file produced by go tool test2json")
+	outPath = flag.String("out", "", "path to output file")
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


tools/testjson2md/testjson2md.go
```diff
@@ -28,12 +28,19 @@ import (
 	"github.com/medyagh/gopogh/pkg/report"
```
blanquicet, line 28: Not related to this PR, but please run "go mod tidy"

tools/testjson2md/testjson2md.go
```diff
@@ -95,14 +102,21 @@ func markdownForContent(content report.DisplayContent) ([]byte, error) {
...
 		fmt.Fprintf(&buf, "#### Failed Tests\n")
 		for _, test := range content.Results["fail"] {
-			fmt.Fprintf(&buf, "<details><summary>%s</summary>\n\n", test.TestName)
-			fmt.Fprintf(&buf, "```code\n")
-			fmt.Fprintf(&buf, "%s", testEventToString(lastN(test.Events, *truncateLines)))
-			fmt.Fprintf(&buf, "```\n")
-			fmt.Fprintf(&buf, "</details>\n")
+			s, d := testEventToDetailsBlock(test.TestName, test.Events)
+			// check if we are over the limit
+			if totalSize+s > (summaryLimitInBytes - summaryReservedInBytes) {
```
blanquicet, line 111: What if you simply leave the Failed Tests by last (moving the `Test Durations` section to the beginning maybe)?

blanquicet (approved):
One comment but LGTM!

tools/testjson2md/testjson2md.go
mqasimsarfraz: It drops the entire summary with error annotation: `$GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of 1024k, got 1265k.`.  We can definitely truncate returning slice but that might result in broken markdown. Since it one off case I am fine with that. Let me know if you think that will be better?

PR #1274, opened 2023-01-17
go.mod: Update fork of cilium/ebpf

alban:
This updates cilium/ebpf to take in the following changes:
- [v0.10.0](https://github.com/cilium/ebpf/releases/tag/v0.10.0) (released 2023-01-16)
- Definition of MissingConstantsError (merged upstream)
- Patches to support perf over writable & backward bufferd (not merged upstream)


alban:
The following failed:
```
FAIL: TestTraceExec

        {"node":"fv-az193-372","namespace":"test-trace-exec-2856498984653892754","pod":"test-pod","container":"test-pod","timestamp":1673970546962800382,"type":"normal","pid":19634,"ppid":19169,"comm":"sleep","args":["/bin/sleep","0.1"],"mountnsid":4026532819}
        
    command.go:493: invalid command output(StartTraceExecGadget): verifying output with custom function: output doesn't contain the expected entry: &{Event:{CommonData:{Node: Namespace:test-trace-exec-2856498984653892754 Pod:test-pod Container:test-pod} Timestamp:1970-01-01T00:00:00.000000000Z Type:normal Message:} Pid:0 Ppid:0 Comm:sh Retval:0 Args:[/bin/sh -c while true; do date ; /bin/sleep 0.1; done] UID:0 MountNsID:0}
```

I am restarting the CI to see if it was a single failure or not.

alban:
It works with this additional patch... But I think it is a race condition which is not related to the cilium/ebpf update.

PR #1275, opened 2023-01-18
ci: Compute checksums and publish file containing them as release asset.

eiffel-fl:
Hi.

While playing with `cosign` to test how we can use this tools to sign our release assets, I wrote this commit to compute checksums for our release assets.
The result can be found [here](https://github.com/eiffel-fl/inspektor-gadget/releases/tag/vtest3) and the corresponding pipeline is this [one](https://github.com/eiffel-fl/inspektor-gadget/actions/runs/3948084444).
Note that, with this commit, only the `_checksums.txt` file will be published as asset and no signature or certificate files.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -816,6 +816,13 @@ jobs:
         for i in *-gadget-*-*-tar-gz/*-gadget-*-*.tar.gz; do
           mv $i $(dirname $i)/$(basename $i .tar.gz)-${{ github.ref_name }}.tar.gz
         done
+    - name: Compute checksums for all artifacts
+      shell: bash
+      run: |
+        for i in *-gadget-*-*-tar-gz/*-gadget-*-*.tar.gz inspektor-gadget-${{ github.ref_name }}.yaml; do
+          hash=$(sha256sum $i | cut -d' ' -f1)
+          echo "${hash}  $(basename $i)" >> inspektor-gadget-${{ github.ref_name }}_checksums.txt
+        done
     - name: Upload Gadget Release *.tar.gz
       uses: csexton/release-asset-action@v2
       with:
@@ -826,6 +833,12 @@ jobs:
       uses: csexton/release-asset-action@v2
       with:
         file: inspektor-gadget-${{ github.ref_name }}.yaml
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        release-url: ${{ steps.create_release.outputs.upload_url }}
+    - name: Upload checksums file
+      uses: csexton/release-asset-action@v2
+      with:
+        file: inspektor-gadget-${{ github.ref_name }}_checksums.txt
         github-token: ${{ secrets.GITHUB_TOKEN }}
         release-url: ${{ steps.create_release.outputs.upload_url }}
     - name: Update new version in krew-index


.github/workflows/inspektor-gadget.yml
```diff
@@ -828,6 +834,12 @@ jobs:
         file: inspektor-gadget-${{ github.ref_name }}.yaml
         github-token: ${{ secrets.GITHUB_TOKEN }}
         release-url: ${{ steps.create_release.outputs.upload_url }}
+    - name: Upload checksums file
+      uses: csexton/release-asset-action@v2
+      with:
+        file: inspektor-gadget-${{ github.ref_name }}_checksums.txt
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        release-url: ${{ steps.create_release.outputs.upload_url }}
```
blanquicet, line 842: What about printing the checksum of the checksum file so that we add it to the release notes? For instance, [nerdctl](https://github.com/containerd/nerdctl/blob/241ee42035ace85f955a128c8450de2a59e1456c/.github/workflows/release.yml#L38) does it.

.github/workflows/inspektor-gadget.yml
```diff
@@ -816,6 +816,12 @@ jobs:
         for i in *-gadget-*-*-tar-gz/*-gadget-*-*.tar.gz; do
           mv $i $(dirname $i)/$(basename $i .tar.gz)-${{ github.ref_name }}.tar.gz
         done
+    - name: Compute checksums for all artifacts
+      shell: bash
+      run: |
+        for i in *-gadget-*-*-tar-gz/*-gadget-*-*.tar.gz inspektor-gadget-${{ github.ref_name }}.yaml; do
+          sha256sum $i >> inspektor-gadget-${{ github.ref_name }}_checksums.txt
```
blanquicet, line 823: The file looks like this:

```
91110ae0ece46da55c15176f587f249ee67f8629150c970e0453d1d5ed76ccf0  kubectl-gadget-darwin-amd64-tar-gz/kubectl-gadget-darwin-amd64-vtest3.tar.gz
be92463b78835b37ba073641fa99a43e660b5b002aedc7894a99a2e68c2a2ca9  kubectl-gadget-darwin-arm64-tar-gz/kubectl-gadget-darwin-arm64-vtest3.tar.gz
[...]
```

But, IMO, it should look like this:
```
91110ae0ece46da55c15176f587f249ee67f8629150c970e0453d1d5ed76ccf0  kubectl-gadget-darwin-amd64-vtest3.tar.gz
be92463b78835b37ba073641fa99a43e660b5b002aedc7894a99a2e68c2a2ca9  kubectl-gadget-darwin-arm64-vtest3.tar.gz
[...]
```

blanquicet (commented):
Thanks for working on this!

.github/workflows/inspektor-gadget.yml
blanquicet: I thought people could have a way to verify the integrity of the checksum file they get, but yeah, they should do it manually, which is not that great advantage. Let's go ahead as it is. Thanks!

blanquicet (approved):
LGTM! 

It would be great if you can run it again on your fork to verify that it is working properly after the change. We don't want surprises for the next release 🙂.

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal, line 842: Any specific reason to have a single file with all cheksums instead of a file per checksum? 

I don't have any preference, I see both approaches are used in different repos, but I wanted to confirm if there is any specific reason to use one or another way. 

https://github.com/containerd/containerd/releases/tag/v1.6.15 vs https://github.com/containers/podman/releases/tag/v4.3.1

PR #1278, opened 2023-01-18
integration: Ensure local-gadget is ready before running test

mqasimsarfraz:
We have seen [scenario](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3949717373/attempts/1#summary-10727271605) in CI here `local-gadget` isn't ready when test events are generated resulting in missing initial events. Add a sleep to ensure local-gadget is ready before we generate test events.

### DIFF
--- a/integration/local-gadget/non-k8s/trace_dns_test.go
+++ b/integration/local-gadget/non-k8s/trace_dns_test.go
@@ -108,6 +108,7 @@ func TestTraceDns(t *testing.T) {
 
 	testSteps := []TestStep{
 		traceDNSCmd,
+		SleepForSecondsCommand(2), // wait to ensure local-gadget has started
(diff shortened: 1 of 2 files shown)


mqasimsarfraz:
> But I think kubectl-gadget and local-gadget should provide a way to tell when it has initialized the tracers. Without `-o json`, it is visible when the column titles are displayed. But with `-o json`, maybe we should add an event of type [READY](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/types/types.go#L99) for all gadgets.

Thanks for the approval. Definitely, makes sense to handle it other way around. I [opened](https://github.com/inspektor-gadget/inspektor-gadget/issues/1279) the issue to track it.

PR #1282, opened 2023-01-20
Give back demos from past presentation

eiffel-fl:
Hi.

In this PR, I added two demos GIF I recorded for a past presentation.

Best regards.


docs/gadgets/snapshot/socket.md
```diff
@@ -5,6 +5,8 @@ description: >
   Gather information about TCP and UDP sockets.
 ---
 
+![Screencast of snapshot process compared to calling kubectl exec ps](capabilities.gif)
+
```
mauriciovasquezbernal, line 9: What happened here? :laughing: Shouldn't it to include the `process.gif` in `process.md`?

docs/gadgets/trace/capabilities.md
```diff
@@ -177,6 +177,11 @@ You can now delete the pod you created:
 $ kubectl delete -f docs/examples/app-set-priority-locked-down.yaml
 ```
 
+The following animation shows a real world example of debugging an `nginx` deployment:
+
+![Screencast of using trace capabilities to debug an nginx deployment](nginx_capabilities.gif)
+
+
```
mauriciovasquezbernal, line 184: I'm wondering if we should add this demo to this guide. It already has a gif on the top and some instructions to run it. I think the new gif is nice, but:
- There is overlap with the existing one: even if this is more complex, the base use of the gadget is already shown before 
- Do we want to handle the maintenance of this new gif? (If something changes we'll have to update multiple guides per each gadget)

mauriciovasquezbernal (approved):
LGTM. Thanks!

PR #1284, opened 2023-01-20
Fix race condition on enrichers

mauriciovasquezbernal:
Different gadgets use the different enricher interfaces to add
container information to the events received from the eBPF programs.
It could happen that by the time the tracer calls the enricher, the
container is gone and then the container information from the event
is missing.

This case is specially relevant when the container is killed after
executing the event, like using a capability not allowed or called
a syscall blocked by seccomp.

This commit fixes this by creating a container cache system: containers
are retained and made available to enrichers for a while after those
are removed from the container collection package, this allows the
enrichers to provide the container data even if the container is gone.

The container collection package is generic enough, hence this behavior
is only enabled when "WithTracerCollection" is used.

### How to test

Check https://github.com/inspektor-gadget/inspektor-gadget/issues/1178#issuecomment-1396238600. (Check that you're able to reproduce without this PR first, because it's possible that the issue doesn't reproduce in all environments)

### TODO
- [x] Implement test for WithTracerCollection
- [x] Update [audit seccomp](https://github.com/inspektor-gadget/inspektor-gadget/blob/00b892c97a95e0215cde92ec1e8c11d9349c0ee6/pkg/gadgets/audit/seccomp/tracer/tracer.go#L143-L151) to use this mechanism instead of saving container names in ebpf code? (cc @alban)

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1178


pkg/container-collection/container-collection.go
```diff
@@ -127,8 +134,20 @@ func (cc *ContainerCollection) RemoveContainer(id string) {
...
+		cc.pubsub.Publish(EventTypeRemoveContainer, container)
+	}
+
+	// Save the container in the cache as enrichers might need the container some time after it
+	// has been removed.
+	if cc.cachedContainers != nil {
+		cc.cachedContainers.Store(id, v)
+		time.AfterFunc(cc.cacheDelay, func() {
+			cc.cachedContainers.Delete(id)
+			unix.Close(container.mntNsFd)
```
alban, line 149: This is fine for gadgettracermanager and local-gadget because they use `WithTracerCollection` but in general mntNsFd is not initialized if `WithTracerCollection` is not used.

pkg/container-collection/containers.go
```diff
@@ -65,6 +65,12 @@ type Container struct {
 	PodUID    string            `json:"podUID,omitempty"`
 
 	ownerReference *metav1.OwnerReference
+
+	// We keep an open file descriptor of the containers mount namespace to be sure the kernel
+	// doesn't reuse the inode id before we get rid of this container. This logic avoids a race
+	// condition when the mnt ns inode id is reused by a new container and we erroneously pick
+	// events from it.
+	mntNsFd int
```
alban, line 73: It is not obvious that this field is only used with `WithTracerCollection` when `cc.cachedContainers != nil`. Could you add it in the comment or alternatively initialize the field to `-1`?

pkg/container-collection/container-collection.go
mauriciovasquezbernal: It's only executed if `cc.cachedContainers != nil` and there is a condition on L145 checking it.

pkg/container-collection/containers.go
mauriciovasquezbernal: Sure. Done!

alban:
> Update [audit seccomp](https://github.com/inspektor-gadget/inspektor-gadget/blob/00b892c97a95e0215cde92ec1e8c11d9349c0ee6/pkg/gadgets/audit/seccomp/tracer/tracer.go#L143-L151) to use this mechanism instead of saving container names in ebpf code? (cc @alban)

Ideally yes, we should avoid using different mechanisms in different gadgets for the same thing when we can avoid it.

pkg/container-collection/options.go
```diff
@@ -642,3 +645,33 @@ func WithNodeName(nodeName string) ContainerCollectionOption {
...
 	}
 }
+
+type TracerCollection interface {
+	TracerMapsUpdater() FuncNotify
+}
+
+func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.cacheDelay = 2 * time.Second
```
alban, line 655: I wonder if there is a way to make it not time-based, so that it would work even if the `local-gadget` is suspended (ctrl-Z) for a while... Some ideas:
* Increase the cacheDelay to a minute
* Use a LRU cache with a fixed size
* Have a synchronization mechanism with ring buffer readers so we can remove the container from the cache when all pending events have been enriched.

I think your `2 * time.Second` is better than the ideas above but it still make me uneasy.

pkg/container-collection/container-collection.go
```diff
@@ -339,7 +370,12 @@ func (cc *ContainerCollection) EnrichByMntNs(event *eventtypes.CommonData, mount
 func (cc *ContainerCollection) EnrichByNetNs(event *eventtypes.CommonData, netnsid uint64) {
 	event.Node = cc.nodeName
 
-	containers := cc.LookupContainersByNetns(netnsid)
+	var containers []*Container
+
+	containers = append(containers, lookupContainersByNetns(&cc.containers, netnsid)...)
+	if cc.cachedContainers != nil {
```
alban, line 376: For `EnrichByMntNs`, you also test `container == nil` but not here. It seems reasonable because there could be several containers in the same netns, so we should look into `cc.cachedContainers` even if we find a container in `cc.containers` already. But that makes the semantic subtle.

`EnrichByNetNs` is only used by the network tracer (pkg/gadgets/trace/network/tracer/tracer.go) and it has its own cache for enriching events of terminated containers. That cache works differently than this one. Do you intend to keep both?

pkg/container-collection/options.go
mauriciovasquezbernal: > I wonder if there is a way to make it not time-based, so that it would work even if the `local-gadget` is suspended (ctrl-Z) for a while... 

Do we really want to support that case? What other things happen when the process is suspended? I imagine perf buffer will be full and events will be discarded. (I'm not saying we shouldn't support the case, however I think it needs more investigation as they are other things that could break too)

Some ideas:
> 
>     * Increase the cacheDelay to a minute
> 

It'll improve the situation, by why one minute, why not 2, 5, 10? I'm also a bit worried about the extra memory consumption it implies to have containers one more minute around.

>     * Use a LRU cache with a fixed size
> 

I'm also worried about memory consumption there. Perhaps using an LRU with expiration (https://github.com/hashicorp/golang-lru/pull/116) can be a trade off, however it's not yet supported by the library we use. I also think creating a reduced container structure to save in the cache could help as we don't need all the fields to enrich events. 

>     * Have a synchronization mechanism with ring buffer readers so we can remove the container from the cache when all pending events have been enriched.
> 

mmm, how will it work? How can the eBPF program mark that it was the last event sent on a perf ring buffer? Also, what about the case when we don't have tracer running?
> 

> I think your `2 * time.Second` is better than the ideas above but it still make me uneasy.

I think the solution proposed in this PR fixes the issue in many cases, I'm open to improvements but I'd delay those to a future PR in order to get this soon, that do you think?

pkg/container-collection/container-collection.go
```diff
@@ -339,7 +370,12 @@ func (cc *ContainerCollection) EnrichByMntNs(event *eventtypes.CommonData, mount
 func (cc *ContainerCollection) EnrichByNetNs(event *eventtypes.CommonData, netnsid uint64) {
 	event.Node = cc.nodeName
```
flyth, line 371: I'm a bit worried about performance here, since the code is pretty hot: adding a second map to range over for every incoming event (unlike in EnrichByMntNs, where it's probably used rarely when the container hasn't been found) including handling locks etc. seems pretty expensive.

What about adding a `deleted` flag to the container struct and filter using that instead of using a different map?

However, in the long term, we should probably add two new maps for directly looking up by MntNs (e.g. `map[uint64]*Container`) & NetNs (something like `map[uint64]struct{Container *Container, useCount int}`, where useCount is used to do skip the ContainerName when enriching) which are updated when containers are added/deleted, as ranging over potentially a couple of hundred containers in a production environment could really hit performance.

pkg/container-collection/container-collection.go
mauriciovasquezbernal: hum, it's a good question. I didn't take into account the cache mechanism implemented in trace network. I don't think we can combine two approaches easily as the one used in this PR is to hold the container in memory long enough to allow the tracer to read and enrich an event from the ring buffer. On the other hand, in the trace network gadget that's different, we need to have a cache until the user calls `Pop()`, and it could be a much longer time span.

mauriciovasquezbernal: Also, the current design I proposed here is wrong, it could happen that a pod is removed and then a new one is created reusing the same network namespace inode id, hence there is a timestamp where both pods will be in memory and the enrichment will give wrong results. A possibility is to also keep a reference of the net ns id to avoid it being reused, but I'm getting worried about the number of opened file descriptors we keep. 

I'd propose to remove the changes introduced to `EnrichByNetNs` in this PR and discuss that later on, specially because I'm wondering how https://github.com/inspektor-gadget/inspektor-gadget/pull/1201 will impact all the enrich logic in networking gadgets.  WDYT?

pkg/container-collection/options.go
mauriciovasquezbernal: I'm not sure a mutex will help a lot here. As you mentioned, the major issue is removing the container when the enrichment process hasn't started yet. Using a mutex to protect when the enrichment has started will cover only few cases.

pkg/local-gadget-manager/local-gadget-manager.go
```diff
@@ -91,17 +91,15 @@ func NewManager(runtimes []*containerutils.RuntimeConfig) (*LocalGadgetManager,
 	if err != nil {
 		return nil, fmt.Errorf("error creating containers map: %w", err)
 	}
-	containerEventFuncs := []containercollection.FuncNotify{}
-	containerEventFuncs = append(containerEventFuncs, l.containersMap.ContainersMapUpdater())
-	containerEventFuncs = append(containerEventFuncs, l.tracerCollection.TracerMapsUpdater())
 
 	err = l.ContainerCollection.Initialize(
-		containercollection.WithPubSub(containerEventFuncs...),
+		containercollection.WithPubSub(l.containersMap.ContainersMapUpdater()),
```
blanquicet, line 96: I don't think the localgadgetmanager needs the ContainerMap.

blanquicet (commented):
There is still space for improvements, like avoiding iterating over the sync-maps using dedicated maps and avoiding depending on time. However, I agree we can do it step-by-step, and this PR is indeed fixing the issue, so I propose going ahead with this.

pkg/container-collection/container-collection_test.go
```diff
@@ -0,0 +1,123 @@
...
+			Name:      fmt.Sprintf("name%d", i),
+			Namespace: fmt.Sprintf("namespace%d", i),
+			Podname:   fmt.Sprintf("pod%d", i),
+			Mntns:     runner.Info.MountNsID,
+			Pid:       uint32(runner.Info.Pid),
+		}
+		cc.AddContainer(containers[i])
+	}
+
+	require.Equal(t, nContainers, len(f.containers), "number of containers should be equal")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: https://github.com/stretchr/testify seems a lot more popular than https://github.com/gotestyourself/gotest.tools.

pkg/container-collection/container-collection.go
```diff
@@ -39,6 +41,11 @@ type ContainerCollection struct {
 	// Values: container   Container
 	containers sync.Map
 
+	// Saves containers for "cacheDelay" to be able to enrich events after the container is
+	// removed. This is enabled by using WithTracerCollection().
+	cachedContainers *sync.Map
```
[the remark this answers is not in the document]
mauriciovasquezbernal: This option is not enabled all the times, then I used a pointer to avoid needing a boolean to check whether this is enabled or not.

pkg/container-collection/container-collection_test.go
```diff
@@ -0,0 +1,123 @@
...
+	utilstest.RequireRoot(t)
+
+	cc := ContainerCollection{}
+	f := &fakeTracerMapsUpdater{containers: make(map[string]*Container)}
+
+	opts := []ContainerCollectionOption{
+		WithTracerCollection(f),
+	}
+
+	if err := cc.Initialize(opts...); err != nil {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Right, tests were not working at all and this change was one of my desperate attempts to fix it.

pkg/local-gadget-manager/local-gadget-manager.go
mauriciovasquezbernal: It's not needed even in kubectl-gadget because the audit seccomp gadget doesn't use this anymore after this PR. However I plan to remove that in a following PR to make reviewing this easier.

pkg/container-collection/container-collection_test.go
```diff
@@ -0,0 +1,123 @@
...
+	containers map[string]*Container
+}
+
+func (f *fakeTracerMapsUpdater) TracerMapsUpdater() FuncNotify {
+	return func(event PubSubEvent) {
+		switch event.Type {
+		case EventTypeAddContainer:
+			f.containers[event.Container.ID] = event.Container
+		case EventTypeRemoveContainer:
+			delete(f.containers, event.Container.ID)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'll avoid them as IMO they don't add too much value to this test.

pkg/container-collection/container-collection_test.go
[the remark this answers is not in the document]
blanquicet: So, I suggest to move to it but not keep both.

pkg/local-gadget-manager/local-gadget-manager.go
blanquicet: But for the gadgettracermanager it is used by standard tools.

mauriciovasquezbernal: Ah, I almost forgot we still have those.

blanquicet (approved):
LGTM! I reproduced it with a1e2b601c006f240126e7f1578619e94276df9c1 and [these steps](https://github.com/inspektor-gadget/inspektor-gadget/issues/1178#issuecomment-1396238600) on an AKS cluster after a few seconds. Then, I ran the same steps with this PR for 20 minutes and didn't reproduce it.

I know this issue was complicated to reproduce and isolate from other race conditions. So, fantastic work!

alban (approved):
I didn't review all the tests but the fix looks good to me.

PR #1285, opened 2023-01-23
integration: Upgrade SPO to v0.6.0

mqasimsarfraz:
We have seen errors related to SPO webhook not available in CI, resulting in failing tests. It seems they have made SPO webhook opt-in [upstream](https://github.com/kubernetes-sigs/security-profiles-operator/pull/1207) to avoid such situation. Upgrading so we don't hit such issues in the CI.

```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "recording.spo.io": failed to call webhook: Post "[https://webhook-service.security-profiles-operator.svc:443/mutate-v1-pod-recording?timeout=10s](https://webhook-service.security-profiles-operator.svc/mutate-v1-pod-recording?timeout=10s)": context deadline exceeded
```
Also, I tried to check why `webhook` wasn't available and it seems for some reason `webhook` was being restarted maybe because of low resources in runner?

Related upstream change:
- https://github.com/kubernetes-sigs/security-profiles-operator/pull/1207
> The SPO webhooks have been cluster-wide by default. This is great for
functionality as both recording and binding just works, but might be
dangerous - if there is a bug in the webhook or the webhook is not
replying, then all namespaces might be affected.

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -113,9 +113,9 @@ func DeployInspektorGadget(image, imagePullPolicy string) *Command {
 
 func DeploySPO(limitReplicas, patchWebhookConfig, bestEffortResourceMgmt bool) *Command {
 	cmdStr := fmt.Sprintf(`
-kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.8.0/cert-manager.yaml
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.10.0/cert-manager.yaml
 kubectl --namespace cert-manager wait --for condition=ready pod -l app.kubernetes.io/instance=cert-manager
-curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.3/deploy/operator.yaml | \
+curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.6.0/deploy/operator.yaml | \
   if [ %v = true ] ; then
     sed 's/replicas: 3/replicas: 1/' | grep -v cpu:
   else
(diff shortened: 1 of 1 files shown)


integration/command.go
```diff
@@ -113,9 +113,9 @@ func DeployInspektorGadget(image, imagePullPolicy string) *Command {
 
 func DeploySPO(limitReplicas, patchWebhookConfig, bestEffortResourceMgmt bool) *Command {
 	cmdStr := fmt.Sprintf(`
-kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.8.0/cert-manager.yaml
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.10.1/cert-manager.yaml
```
blanquicet, line 116: Why v1.10.1? The [documentation](https://github.com/kubernetes-sigs/security-profiles-operator/blob/v0.6.0/installation-usage.md#install-operator) suggests 1.10.0.

integration/command.go
```diff
@@ -113,9 +113,9 @@ func DeployInspektorGadget(image, imagePullPolicy string) *Command {
...
 kubectl --namespace cert-manager wait --for condition=ready pod -l app.kubernetes.io/instance=cert-manager
-curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.4.3/deploy/operator.yaml | \
+curl https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-operator/v0.6.0/deploy/operator.yaml | \
```
alban, line 118: The change `v0.4.3` -> `v0.6.0` must be done line 199 too.

Ditto for `v1.8.0` -> `v1.10.1` line 200.

integration/command.go
```diff
@@ -113,9 +113,9 @@ func DeployInspektorGadget(image, imagePullPolicy string) *Command {
```
blanquicet, line 113: IIUC [the documentation](https://github.com/kubernetes-sigs/security-profiles-operator/blob/v0.6.0/installation-usage.md#label-namespaces-for-binding-and-recording), the webhook is only necessary for `ProfileBindings` and `ProfileRecordings` operations. However, given that we are using none of those operations, we don't need to apply any special label to our namespaces. In this way, the webhook will not infer with any operation in our namespaces (and sometimes makes the pod creation fail). Am I right?

blanquicet (commented):
Great! I hope this solves the issues we have having with the webhook.

integration/command.go
mqasimsarfraz: Ah I was checking `main` where they already changed to `v1.10.1`.

integration/command.go
mqasimsarfraz: Or maybe we leave everything as it is since we are doing [`bestEffortResourceMgmt`](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/command.go#L160) already.

blanquicet: I suggest making one change at a time. Let's keep it running with this new release and see how it works.

PR #1287, opened 2023-01-23
tests: Make GetTestPodIP() return an error too

mauriciovasquezbernal:
It's possible that GetTestPodIP() fails as in https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/3986343020/jobs/6834925048. This PR introduces a better way to handle errors in this function, hence avoiding weird test failures messages like:

```
invalid command output(StartTraceTcpconnectGadget): verifying output with custom function: output doesn't contain the expected entry: &{Event:{CommonData:{Node: Namespace:test-tcpconnect-2634895312520334714 Pod:test-pod Container:test-pod} Timestamp:1970-01-01T00:00:00.000000000Z Type:normal Message:} Pid:0 UID:0 Comm:wget IPVersion:4 Saddr:failed to get Pod IP: exit status 1 Daddr:172.17.0.11 Dport:80 MountNsID:0}
```

### DIFF
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -16,6 +16,7 @@ package integration
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os/exec"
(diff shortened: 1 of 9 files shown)


mqasimsarfraz:
This is looking good. Just wondering if we can use something like ["nginx && curl 127.0.0.1"](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/local-gadget/non-k8s/trace_network_test.go#L75) here as well ? I somehow find it simple.

mqasimsarfraz (approved):
LGTM!

PR #1290, opened 2023-01-24
trace/bind: add unit test for UDP on IPv6

alban:
The previous tests were only testing UDP4, TCP4 and TCP6.

Thanks @mauriciovasquezbernal  for the code suggestion.


mauriciovasquezbernal (approved):
My idea there was to keep the tests focused, that's why I implemented ipv4, ipv6, tcp and udp tests. However I think testing all combinations is good as well. Thanks!

PR #1292, opened 2023-01-25
gadgets: call btf.LoadKernelSpec once

alban:
When loading a ebpf module, btf.LoadKernelSpec() is called for each ebpf program. Even if cilium/ebpf has a cache mechanism, btf.(*Spec).Copy() is still called and it is taking a significant time (e.g. 237ms).

Cache mechanism:
https://github.com/cilium/ebpf/blob/dc0c82204f210e73e853bd0ad96060bb33ec3901/btf/btf.go#L290

Functions spec.LoadAndAssign(), ebpf.NewCollectionWithOptions() and loadSeccompObjects() take a ebpf.CollectionOptions parameter where it is possible to pass our own kernel spec. When doing so, btf.(*Spec).Copy() does not need to be called and we improve performance significantly.

This patch introduces a ebpfoptions package to provide a reasonable ebpf.CollectionOptions for all gadgets with a kernel spec created only one time without copying it for each program.

------

See https://github.com/inspektor-gadget/inspektor-gadget/pull/1201#issuecomment-1403598719

TODO:
- [ ] Tests

### DIFF
--- a/pkg/gadgets/advise/seccomp/tracer/tracer.go
+++ b/pkg/gadgets/advise/seccomp/tracer/tracer.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/internal/ebpfoptions"
 	libseccomp "github.com/seccomp/libseccomp-golang"
 )
 
@@ -60,7 +61,7 @@ func (t *Tracer) start() error {
 		return fmt.Errorf("failed to load asset: %w", err)
 	}
 
-	if err := spec.LoadAndAssign(&t.objs, nil); err != nil {
+	if err := spec.LoadAndAssign(&t.objs, ebpfoptions.CollectionOptions()); err != nil {
 		return fmt.Errorf("failed to load ebpf program: %w", err)
 	}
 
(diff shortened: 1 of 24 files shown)


alban:
This fails with:
```
=== CONT  TestEventEnrichmentRaceCondition
    gadgets_test.go:247: failed generating events: failed to run command: running command: fork/exec /bin/cat: too many open files
--- FAIL: TestEventEnrichmentRaceCondition (1.68s)
=== CONT  TestContainerRemovalRaceCondition
    gadgets_test.go:176: failed generating events: failed to run command: running command: open /dev/null: too many open files
--- FAIL: TestContainerRemovalRaceCondition (1.80s)
```

Is it caused by https://github.com/inspektor-gadget/inspektor-gadget/pull/1284? We keep containers in the container-collection with fds open for longer now.

alban:
Convert to draft because I didn't measure how much the performances change with the current version of this patch.
Previous measurements: https://github.com/inspektor-gadget/inspektor-gadget/pull/1201#issuecomment-1403598719

mauriciovasquezbernal:
I think this issue should be fixed directly in the library. I understand that a copy of the BTF spec has to be made in some cases to avoid the user modifying it, but it's not needed for internal use IMO.

mauriciovasquezbernal:
Opened https://github.com/cilium/ebpf/issues/917 to handle it upstream.

mauriciovasquezbernal:
@alban can we close this now that https://github.com/cilium/ebpf/issues/917 is fixed?

PR #1293, opened 2023-01-25
Add figure depicting all gadgets to README

eiffel-fl:
Hi.

This PR moves the existing figure from `architecture.md` to `README.md`.

Best regards.

### DIFF
--- a/docs/images/architecture/inspektor_gadget_tools.svg
+++ b/docs/images/architecture/inspektor_gadget_tools.svg
@@ -2,22 +2,36 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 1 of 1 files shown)


flyth:
This is probably a darkmode-only thing, but for me, the outter text of the svg is barely readable, because it has a transparent/no background:

transparent background in svg

burak-ok:
I think this transparency topic is always there :D
For me it looks like Michaels screenshot. But then when i want to use the figure in a non white background presentation i dislike that it has a default background.

But with `svg` one can easily remove the background when needed for other stuff. And many ppl use dark themes or plugins

burak-ok:
> > For me it looks like Michaels screenshot. But then when i want to use the figure in a non white background presentation i dislike that it has a default background.
> 
> `Ctrl+F5` on the README page when on my branch should do the trick!

Yep. refreshed. Looks "good" now. (Now there is a white rectangle in my all black webpage. Whatever good means :P)
I think i had the page open when you updated your branch :sweat_smile:

blixtra:
LGTM. But as a warning I'm working on some new messaging based on the new functionality that has been added that will likely have this moving again fairly soon.

mauriciovasquezbernal (approved):
LGTM. Let's merge it even if this could be changed soon.

PR #1294, opened 2023-01-26
Add a seccomp profile for Inspektor Gadget pods

eiffel-fl:
Hi.

In this PR, I added a seccomp profile for Inspektor Gadget pods.
I added an option to `kubectl gadget deploy --use-seccomp-profile` to deploy the pods with the associated seccomp profile.

Best regards.


pkg/resources/seccomp-profile/profile.yaml
```diff
@@ -0,0 +1,114 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gadget-seccomp-profile
```
alban, line 4: Why creating a new namespace? We could reuse the `gadget` namespace.

pkg/container-collection/match.go
```diff
@@ -29,6 +29,9 @@ func ContainerSelectorMatches(s *ContainerSelector, c *Container) bool {
 	if s.Podname != "" && s.Podname != c.Podname {
 		return false
 	}
+	if strings.HasSuffix(s.Name, "*") && strings.HasPrefix(c.Name, s.Name[:len(s.Name) - 1]) {
+		return true
```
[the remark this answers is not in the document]
alban: > I do not really want this patch to be included.
> This was more about what I needed to do to generate the profile rather than a polished patch to include.

Did you use local-gadget to inspect Inspektor Gadget?

In gadgettracermanager, `c.Name` does not contain the docker name but the container name as seen in the pod yaml:
```
# docker ps
5f1e29001f84   5185b96f0bec                  "/coredns -conf /etc…"   19 hours ago    Up 19 hours              k8s_coredns_coredns-565d847f94-vjl8g_kube-system_ec3920c7-2cac-4183-be79-124ea7488408_0
```
=> We see the docker name `k8s_coredns_coredns-565d847f94...`
```
# gadgettracermanager -dump
... Namespace:kube-system Podname:coredns-565d847f94-vjl8g Name:coredns
```
=> We see that `c.Name` = `coredns`

pkg/resources/seccomp-profile/profile.yaml
alban: I think it is preferable to use the same `gadget` namespace.

If it makes it easier, instead of a single file `pkg/resources/manifests/deploy.yaml`, we could have:
- `pkg/resources/manifests/namespace.yaml`
- `pkg/resources/manifests/seccomp.yaml`
- `pkg/resources/manifests/rbac.yaml` (sa, role, rolebinding, etc.)
- `pkg/resources/manifests/ds.yaml` (not sure why it was called deploy; there are no deployment, only a daemon set)

cmd/kubectl-gadget/deploy.go
```diff
@@ -336,17 +350,23 @@ func runDeploy(cmd *cobra.Command, args []string) error {
...
-	}
+	if useSeccompProfile {
+		resourcesYAML = append(resourcesYAML[:1], resourcesYAML...)
 
-	traceObjects, err := parseK8sYaml(resources.TracesCustomResource)
-	if err != nil {
-		return err
+		// We need to add the secomp profile creation after the namespace because it
+		// lives in the gadget namespace.
+		resourcesYAML[1] = resources.GadgetSeccompProfile
```
alban, line 358: The append() and replacement is confusing.

If `resourcesYAML` was a `[]*string`, you could just set the item to nil:
```
resourcesYAML[resourcesSeccompIndex] = nil
```

And add a `continue` in the loop below if `resource==nil`.

cmd/kubectl-gadget/deploy.go
```diff
@@ -336,17 +354,25 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("it's not possible to use --quiet and --debug together")
 	}
 
-	objects, err := parseK8sYaml(resources.GadgetDeployment)
-	if err != nil {
-		return err
+	if !useSeccompProfile {
+		resourcesYAML[resourcesSeccompIndex] = ""
```
burak-ok, line 358: nit: to avoid the `if` in the loop
```suggestion
		resourcesYAML = append(resourcesYAML[:resourcesSeccompIndex], resourcesYAML[resourcesSeccompIndex+1:]...)
```

burak-ok: _I didn't read the resolved conversations. Maybe i should do that._

Performance wise it shouldn't matter as you stated. And this is not a hot path anyway.
My nit was more of a discussion question just to see what is the most go-idiomatic way to do this.

Would be fine either way

mauriciovasquezbernal:
How do we plan to keep the this profile up to date?

mauriciovasquezbernal:
I think it's a nice to have feature, however we should have some CI in place to guarantee that the list is updated. 

> But maybe we can add a CI job to do it even though it will be not be simplest to write as it needs local-gadget to be deployed on a worker node and the integration tests to be run while local-gadget runs.

We can do it with a single node k8s cluster too:

```bash
./kubectl-gadget deploy

$ PODNAME=$(kubectl -n gadget get pod -o jsonpath='{.items[*].metadata.name}' | awk '{print $1;}')
$ ./kubectl-gadget advise seccomp-profile -n gadget start --podname $PODNAME
rYE4lqux95nx5sIc

INTEGRATION_TESTS_PARAMS="-no-deploy-ig" make integration-tests

$ ./kubectl-gadget advise seccomp-profile stop rYE4lqux95nx5sIc
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    {
      "names": [
        "accept4",
        "access",
        "arch_prctl",
        "bind",
        "bpf",
        "brk",
        "capget",
        "clock_gettime",
        "clone",
        "close",
        "connect",
        "dup",
        "dup2",
        "dup3",
        "epoll_create1",
        "epoll_ctl",
        "epoll_pwait",
        "epoll_wait",
        "eventfd2",
        "execve",
        "exit_group",
        "faccessat",
        "fanotify_init",
        "fanotify_mark",
        "fcntl",
        "fstat",
        "futex",
        "getdents64",
        "geteuid",
        "getpeername",
        "getpid",
        "getrandom",
        "getrlimit",
        "getsockname",
        "gettid",
        "ioctl",
        "kill",
        "lseek",
        "madvise",
        "mmap",
        "mprotect",
        "munmap",
        "name_to_handle_at",
        "nanosleep",
        "newfstatat",
        "openat",
        "perf_event_open",
        "pidfd_open",
        "pipe2",
        "ppoll",
        "pread64",
        "prlimit64",
        "read",
        "readlinkat",
        "recvfrom",
        "rt_sigaction",
        "rt_sigprocmask",
        "rt_sigreturn",
        "sched_getaffinity",
        "sched_yield",
        "seccomp",
        "select",
        "set_robust_list",
        "set_tid_address",
        "setns",
        "setsockopt",
        "shutdown",
        "sigaltstack",
        "socket",
        "socketpair",
        "stat",
        "statfs",
        "tgkill",
        "uname",
        "wait4",
        "waitid",
        "write"
      ],
      "action": "SCMP_ACT_ALLOW"
    }
  ]
}
```

Another idea is to create an integration tests that deploys IG with the ` --use-seccomp-profile` flag, if this fails when we can manually update the list.

mauriciovasquezbernal:
Oh right, it can't trace itself on this case.

PR #1295, opened 2023-01-26
[internal]: Suggestions to https://github.com/inspektor-gadget/inspektor-gadget/pull/1281

mauriciovasquezbernal:
@flyth what do you think about these changes?


flyth:
Looks good to me!

Perhaps an additional (u)int64Slice would be useful - if one needs another int type, then that would be big enough to hold all possible values for manual casting.

mauriciovasquezbernal:
> Perhaps an additional (u)int64Slice would be useful - if one needs another int type, then that would be big enough to hold all possible values for manual casting.

Done!

PR #1298, opened 2023-01-27
[internal]: pkg/params: add more explicit validation checks and unit tests

mauriciovasquezbernal:


pkg/params/params.go
```diff
@@ -91,7 +90,10 @@ func (p *ParamDesc) GetTitle() string {
 }
 
 func (p *ParamDesc) ToParam() *Param {
-	return &Param{ParamDesc: p}
+	return &Param{
+		ParamDesc: p,
+		value:     p.DefaultValue,
```
mqasimsarfraz, line 95: Nice. This fixes the issue I was hitting regarding default value not being set [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/michael/gadget-registry-local/pkg/gadgets/top/ebpf/tracer/tracer.go#L442) causing panics!

mauriciovasquezbernal: Oh yes: I think it's the same as https://github.com/inspektor-gadget/inspektor-gadget/pull/1281/files#r1089072222

mqasimsarfraz: Just tested it, `runtimes` flags are working fine as well!

pkg/params/params_test.go
```diff
@@ -0,0 +1,363 @@
...
+		},
+		{
+			name:     "Uint64()",
+			value:    "2222222222",
+			expected: uint64(2222222222),
+			getter:   func(p *Param) any { return p.AsUint64() },
+		},
+		{
+			name:     "String()",
+			value:    "eW91J3JlIGN1cmlvdXM=",
```
mqasimsarfraz, line 107: :laughing:

pkg/params/validator_test.go
```diff
@@ -0,0 +1,378 @@
...
+			bitSize:       bitSize,
+			value:         strconv.FormatInt(-(1<<(bitSize-1))-1, 10),
+			expectedError: true,
+			validateFn:    ValidateInt,
+		})
+	}
+
+	// ValidateUint
+	for _, bitSize := range bitSizes {
+		// test max and min numbers allowed
```
mqasimsarfraz, line 97:
```suggestion
		// test max number allowed
```

mqasimsarfraz (approved):

pkg/params/validators.go
```diff
@@ -23,20 +23,56 @@ import (
...
+	TypeUint32 TypeHint = "uint32"
+	TypeUint64 TypeHint = "uint64"
 )
 
 var typeHintValidators = map[TypeHint]ParamValidator{
-	TypeInt:  ValidateNumber, // TODO: more specific
-	TypeUint: ValidateNumber, // TODO: more specific
-	TypeBool: ValidateBool,
+	TypeBool:   ValidateBool,
+	TypeInt:    ValidateInt(strconv.IntSize),
```
flyth, line 42: I'm not totally sure whether we should use `IntSize` here - maybe it's better to stick with a hardcoded 64 in that case. Reason being: if used in a 32 bit kubectl-gadget binary, this would be different to what runs on the server and be too strict. It's a bit edge casey, I admit ;)

flyth (approved):
Awesome, thanks! Those tests are really comprehensive!

LGTM

pkg/params/validators.go
mauriciovasquezbernal: I had a offline talk with @flyth, the plan is to remove the TypeInt / TypeUint as those types with a specific size are difficult to handle in distributed systems as client and server could be compiled on different architectures. I'll merge this PR and those will be removed in https://github.com/inspektor-gadget/inspektor-gadget/pull/1281.

PR #1301, opened 2023-01-30
A different way to manage parameters instances

blanquicet:
See the whole discussion [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/1281#discussion_r1090595880).


blanquicet:
Agreed to go in another direction.

PR #1305, opened 2023-02-01
Making gadget-runner a gadget-context

blanquicet:
In this PR, I started experimenting with removing `{Get|Set}Result` from gadget-runner and making the registry call directly the runtime for running the gadget. This will allow us to simplify the gadget-runner and make it just hold the information of the running gadget (becoming a gadget-context).


cmd/common/registry.go
```diff
@@ -172,90 +177,96 @@ func buildCommandFromGadget(gadget gadgets.Gadget,
...
-				if len(filters) > 0 {
-					err = parser.SetFilters(filters)
-					if err != nil {
-						return err // TODO: Wrap
-					}
-					gadgetParams.AddKeyValuePair("columns_filters", strings.Join(filters, ",")) // TODO: maybe encode?! difficult for CRs though
+			if parser == nil {
+				// This kind of gadgets return directly the result instead of
+				// using the parser
+				result, err := runtime.RunGadget(gadgetContext, operatorsPerGadgetParamCollection)
```
mauriciovasquezbernal, line 193: Any reason not to include `operatorsPerGadgetParamCollection` in the context too?

internal/gadget-context/gadget-context.go
```diff
@@ -45,80 +44,55 @@ type GadgetRunner struct {
...
+) *GadgetContext {
+	return &GadgetContext{
+		ctx:          ctx,
+		id:           id,
+		runtime:      runtime,
+		gadget:       gadget,
+		gadgetParams: gadgetParams,
+		parser:       parser,
+		logger:       logger,
+		operators:    operators.GetOperatorsForGadget(gadget),
```
mauriciovasquezbernal, line 64: What to you think about passing them in the constructor? (In that way `GadgetContext` will become just a struct that holds values).

mauriciovasquezbernal (approved):
LGTM.
Two minor comments.

internal/gadget-context/gadget-context.go
flyth:
```golang
func (r *GadgetContext) OperatorsPerGadgetParamCollection() params.Collection {
	return r.operatorsPerGadgetParamCollection
}
```

Instead of this then, we could actually name it OperatorParamCollection(), I think. And maybe instead of `PerGadgetParams()` in operators we go to `ParamDescs()` - we also have the `GlobalParamDescs()` so it should be clear that these are not global (and that can be reinforced by proper documentation). Wdyt?

flyth: This was meant to be an answer to the comment above, sorry ;)

mauriciovasquezbernal: I think having `ParamDescs()` and `GlobalParamDescs()` is better than having `PerGadget`.

mauriciovasquezbernal:
I think these changes are already included in https://github.com/inspektor-gadget/inspektor-gadget/pull/1281. Can we close this one?  @flyth @blanquicet

blanquicet:
Changes were included in #1281

PR #1277, opened 2023-01-18
sort events by timestamps

alban:
Events can be unsorted for different reasons:
- perf ring buffers use a different buffer for each cpu
- networking gadgets use a different perf ring buffer for each network
  namespace

How to reproduce the bug easily:

* In the first terminal:
```
  $ go run -exec sudo ./cmd/local-gadget/... trace dns -r docker -o custom-columns=timestamp,qr,name
  TIMESTAMP                           QR NAME
  ^Z
```
* In the second terminal:
```
  $ docker run -ti --rm --name test1 busybox
  / # for i in `seq 10` ; do taskset -c $(($i % 4)) nslookup -type=a svc$i.10.0.0.1.nip.io ; done
```

* Back to the first terminal:
```
  $ fg %1
  2023-01-18T15:08:08.677820284+01:00 Q  svc3.10.0.0.1.nip.io.
  2023-01-18T15:08:08.790717728+01:00 Q  svc7.10.0.0.1.nip.io.
  2023-01-18T15:08:08.613460083+01:00 Q  svc1.10.0.0.1.nip.io.
  2023-01-18T15:08:08.709986563+01:00 Q  svc5.10.0.0.1.nip.io.
  2023-01-18T15:08:08.831027415+01:00 Q  svc9.10.0.0.1.nip.io.
  2023-01-18T15:08:08.624624594+01:00 R  svc1.10.0.0.1.nip.io.
  2023-01-18T15:08:08.674216648+01:00 R  svc2.10.0.0.1.nip.io.
  2023-01-18T15:08:08.692204281+01:00 R  svc3.10.0.0.1.nip.io.
  2023-01-18T15:08:08.695384170+01:00 Q  svc4.10.0.0.1.nip.io.
  2023-01-18T15:08:08.707398320+01:00 R  svc4.10.0.0.1.nip.io.
  2023-01-18T15:08:08.747448019+01:00 R  svc5.10.0.0.1.nip.io.
  2023-01-18T15:08:08.788357359+01:00 R  svc6.10.0.0.1.nip.io.
  2023-01-18T15:08:08.806841824+01:00 R  svc7.10.0.0.1.nip.io.
  2023-01-18T15:08:08.809255967+01:00 Q  svc8.10.0.0.1.nip.io.
  2023-01-18T15:08:08.828446626+01:00 R  svc8.10.0.0.1.nip.io.
  2023-01-18T15:08:08.846491396+01:00 R  svc9.10.0.0.1.nip.io.
[...]

PR #1306, opened 2023-02-02
New pkg to sort events by timestamps

alban:
This is the same code as the first patch of #1277 but without touching the gadgets. I am filing a separate PR in order to keep the other patches from #1277 easily accessible. This PR should be mergeable before #1281 without causing conflicts. The other changes from #1277 can be re-implemented later using an operator as defined in #1281.

-----

Events can be unsorted for different reasons:
- perf ring buffers use a different buffer for each cpu
- networking gadgets use a different perf ring buffer for each network namespace

How to reproduce the bug easily:

* In the first terminal:
```
$ go run -exec sudo ./cmd/local-gadget/... trace dns -r docker -o custom-columns=timestamp,qr,name
TIMESTAMP                           QR NAME
^Z
```
* In the second terminal:
```
$ docker run -ti --rm --name test1 busybox
/ # for i in `seq 10` ; do taskset -c $(($i % 4)) nslookup -type=a svc$i.10.0.0.1.nip.io ; done
```

* Back to the first terminal:
```
$ fg %1
2023-01-18T15:08:08.677820284+01:00 Q  svc3.10.0.0.1.nip.io.
2023-01-18T15:08:08.790717728+01:00 Q  svc7.10.0.0.1.nip.io.
2023-01-18T15:08:08.613460083+01:00 Q  svc1.10.0.0.1.nip.io.
2023-01-18T15:08:08.709986563+01:00 Q  svc5.10.0.0.1.nip.io.
2023-01-18T15:08:08.831027415+01:00 Q  svc9.10.0.0.1.nip.io.
2023-01-18T15:08:08.624624594+01:00 R  svc1.10.0.0.1.nip.io.
2023-01-18T15:08:08.674216648+01:00 R  svc2.10.0.0.1.nip.io.
2023-01-18T15:08:08.692204281+01:00 R  svc3.10.0.0.1.nip.io.
2023-01-18T15:08:08.695384170+01:00 Q  svc4.10.0.0.1.nip.io.
2023-01-18T15:08:08.707398320+01:00 R  svc4.10.0.0.1.nip.io.
2023-01-18T15:08:08.747448019+01:00 R  svc5.10.0.0.1.nip.io.
2023-01-18T15:08:08.788357359+01:00 R  svc6.10.0.0.1.nip.io.
2023-01-18T15:08:08.806841824+01:00 R  svc7.10.0.0.1.nip.io.
2023-01-18T15:08:08.809255967+01:00 Q  svc8.10.0.0.1.nip.io.
2023-01-18T15:08:08.828446626+01:00 R  svc8.10.0.0.1.nip.io.
2023-01-18T15:08:08.846491396+01:00 R  svc9.10.0.0.1.nip.io.
2023-01-18T15:08:08.899635475+01:00 R  svc10.10.0.0.1.nip.io.
2023-01-18T15:08:08.633105917+01:00 Q  svc2.10.0.0.1.nip.io.
2023-01-18T15:08:08.750021549+01:00 Q  svc6.10.0.0.1.nip.io.
2023-01-18T15:08:08.851341853+01:00 Q  svc10.10.0.0.1.nip.io.
```
This patch will allow to sort events correctly.

## Testing done

None, but it was tested in #1277.

### DIFF
--- a/pkg/event-sorter/doc.go
+++ b/pkg/event-sorter/doc.go
@@ -1,0 +1,55 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+Package eventsorter is a library that helps to sort streams of events
+chronologically according to their timestamps.
+
+This is motivated by events coming from a eBPF program via perf ring buffers.
+Since perf ring buffer maps have one buffer per cpu, it might be that the
+userland reader reads events out-of-order if they come from different per-cpu
+buffers.
+
+This package is generic and does not need to know the type of your events.
+Instead of using references to an event, it just calls a callback. The caller
+can use closures to handle the events.
+
+This package re-order events based on a uint64 timestamp. It does not need to
+know the clock used for the timestamp. It only compares timestamps between
+themselves and not with other clocks such as time.Now(). In this way, users of
+this library are free to use bpf_ktime_get_ns(), bpf_ktime_get_boot_ns() or
+event a simple counter.
+
+# How does it work?
+
+	// Create the event sorter
+	es = eventsorter.NewEventSorter()
+
+	// Add events. The callbacks will be called asynchronously in the
+	// right order after a short delay
+	event := types.Event{
+		Event: eventtypes.Event{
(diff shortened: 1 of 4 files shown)


alban:
If gadgets migrate from BPF_MAP_TYPE_PERF_EVENT_ARRAY to BPF_MAP_TYPE_RINGBUF (Linux 5.8), this issue will be automatically fixed.

include/gadget/buffer.h has macro `GADGET_TRACER_MAP` to automatically choose the best ring buffer implementation at run-time.

Another source of wrong ordering was using multiple ring buffers for network programs installed in each network namespace. But since this PR was filed, we use a single network program called in a tail-call, so this is not a problem anymore:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/network-gadget-dispatcher.png

So, if we need to support Linux < 5.8, we need to fix this issue. Otherwise, we could just wait until nobody uses Linux < 5.8. Wdyt?

alban:
In the builtin dns gadget, we use a perf ring buffer to pass the packet to userspace. This can't be done with the ebpf ring buffer. For this kind of use cases, we still need the perf ring buffer and therefore we will still need to find a way to sort events by timestamps.

Issue #968, opened 2022-09-21
[RFE] Run integration test on a multi-node cluster before merging a PR


blanquicet:
## Current situation

Now that we are running the integration tests only on Minikube before merging a PR, we are not verifying that the changes introduced by such PR are not introducing regression on a multi-node cluster environment. Therefore, I propose to modify the configuration of minikube in the integration tests to create a multi-node cluster.

blanquicet:
Thanks @eiffel-fl for sharing this. Yeah, 2 would be OK. Thanks!

PR #1308, opened 2023-02-03
integration: Run IG tests on containerd and crio

### LINKED ISSUE #1283: [RFE] Run IG Minikube integration tests with containerd and CRI-O
As discussed [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/1233#issuecomment-1382077327), we should add another matrix in the `test-integration-minikube` job, to run it against the different runtimes: docker, containerd and CRI-O.

mqasimsarfraz:
This PR allows running inspektor-gadget integration tests on `containerd` and `cri-o` on PR runs.

Fixes #1283

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -541,18 +541,14 @@ jobs:
         run: |
           tar zxvf /home/runner/work/inspektor-gadget/local-gadget-linux-amd64.tar.gz
           mv local-gadget local-gadget-linux-amd64
-      # https://github.com/kubernetes/minikube/issues/12928#issuecomment-1245458044
-      - name: Upgrade docker for ${{ matrix.runtime }}
-        if: matrix.runtime == 'cri-o'
-        run: |
-          curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-          sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu  $(lsb_release -cs)  test"
-          sudo apt-get update
-          sudo apt-get install docker-ce
+      - name: Setup minikube
+        uses: ./.github/actions/setup-minikube
+        with:
+          runtime: ${{ matrix.runtime }}
       - name: Run integration for container runtime ${{ matrix.runtime }}
         run: |
           set -o pipefail
-          make -C integration/local-gadget/k8s CONTAINER_RUNTIME=${{ matrix.runtime }} -o build minikube-start test |& tee integration.log
+          make -C integration/local-gadget/k8s CONTAINER_RUNTIME=${{ matrix.runtime }} -o build test |& tee integration.log
       - name: Prepare and publish test report for container runtime ${{ matrix.runtime }}
         if: always()
         continue-on-error: true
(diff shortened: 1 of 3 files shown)


burak-ok (approved):
LGTM. Will hopefully catch bugs earlier with all these different runtimes being tested  

Also thanks for unifying the creation of the minikube env

.github/workflows/inspektor-gadget.yml
```diff
@@ -716,40 +712,35 @@ jobs:
     strategy:
       matrix:
         type: [default, core]
-        driver: [none, docker]
+        profile: [docker, containerd, cri-o, multi-node]
```
mauriciovasquezbernal, line 715: What's special about `multi-node` to be a different profile? Is it possible to always test "multi-node"?

mqasimsarfraz: Good question. I think it helps to identify if something is specifically broken for multi node installation. Can't think of exact use-case so sounds fine to me if we choose to run all profiles with multi-node. (need to be careful with the resources on GH runner)

cc @blanquicet since we requested this as part of #968

blanquicet: Absolutely, all the runtime must run on a multi-node cluster.

> need to be careful with the resources on GH runner
>
We have been already running multi-node for docker runtime for a while, I don't expect others runtime to consume a lot more.

mqasimsarfraz: Cool. I will update the PR to run all runtime with multi-node setup. This also covers single node scenario so we should be fine there as well!

mauriciovasquezbernal (approved):
LGTM!

minikube.mk
```diff
@@ -47,6 +47,6 @@ minikube-start: minikube-start-$(CONTAINER_RUNTIME)
 .PHONY: phony_explicit
 minikube-start-%: minikube-install
 	$(MINIKUBE) status -p minikube-$* -f {{.APIServer}} >/dev/null || \
-	$(MINIKUBE) start -p minikube-$* --driver=$(MINIKUBE_DRIVER) --kubernetes-version=$(KUBERNETES_VERSION) --container-runtime=$* --wait=all
+	$(MINIKUBE) start -p minikube-$* --driver=$(MINIKUBE_DRIVER) --kubernetes-version=$(KUBERNETES_VERSION) --container-runtime=$* --wait=all $${MINIKUBE_PARAMS}
```
blanquicet, line 50: IIUC, we are not using `minikube-start` anymore on the CI, so why is this needed?

mqasimsarfraz: We are using it in the [action](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/actions/setup-minikube/action.yml#L28) now and [`MINIKUBE_PARAMS`](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/actions/setup-minikube/action.yml#L33) are used to pass custom config to minikube. Does that make sense?

blanquicet: Aahh we are using `minikube-start` in the action. I thought we were using `manusa/actions-setup-minikube@v2.7.2`. Go it now. Thanks!

PR #1311, opened 2023-02-07
socketenricher: Fix cleaning up the sockets map

mauriciovasquezbernal:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/pull/1201

PR #1312, opened 2023-02-07
socketenricher: Fix cleaning up the sockets map & fix tests

alban:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/pull/1201
Supersedes https://github.com/inspektor-gadget/inspektor-gadget/pull/1311

The family was missing, hence the cleanup logic wasn't working right.


pkg/gadgets/internal/socketenricher/tracer_test.go
```diff
@@ -203,32 +204,35 @@ func TestSocketEnricherBind(t *testing.T) {
 			if test.validateEvent != nil {
 				test.validateEvent(t, runner.Info, port, entries)
 			}
+
+			if fd != -1 {
+				unix.Close(fd)
+			}
```
mauriciovasquezbernal, line 210: What about using `t.Cleanup()` to run it on L189. Otherwise it won't run if the test fails leaking the fd.

alban: Updated.

pkg/gadgets/internal/socketenricher/tracer_test.go
```diff
@@ -179,56 +180,84 @@ func TestSocketEnricherBind(t *testing.T) {
...
+			continue
+		}
+		entries = append(entries, entry)
+	}
+	if err := iter.Err(); err != nil {
+		t.Fatal("Cannot iterate over socket enricher map:", err)
+	}
+	return entries
+}
+
```
mauriciovasquezbernal, line 239: What about implementing a `ExpectNoEventToMatch()` in helpers.go? I think that function can be useful for other testing as well.

mauriciovasquezbernal (approved):
One comment about creating a helper function, but LGTM already. I tested by reverting the commit that fixes the issues and the tests indeed fail!

pkg/gadgets/internal/socketenricher/tracer_test.go
alban: I'll try to implement that in a separate PR.

PR #1313, opened 2023-02-07
githubactions: Disable fail-fast

mauriciovasquezbernal:
From documentation [0]:

"jobs.<job_id>.strategy.fail-fast applies to the entire matrix. If jobs.<job_id>.strategy.fail-fast is set to true, GitHub will cancel all in-progress and queued jobs in the matrix if any job in the matrix fails. This property defaults to true."

"fast-failing" hides the result of other jobs on the same matrix, hence we don't know if the issue is specific to that job or something more generic. For instance, in [1] we not that the "Integration tests on AKS (Ubuntu, amd64)" test failed, however we don't know the result of other OS / arch combinations, and that information could help to have a better understanding of what's going on.

[0]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast
[1]: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4115766320/jobs/7105471579

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -86,6 +86,7 @@ jobs:
     # level: 0
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         os: [linux, darwin, windows]
         arch: [amd64, arm64]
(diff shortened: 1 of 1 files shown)


blanquicet (approved):
LGTM!

PR #1314, opened 2023-02-07
socketenricher: Fix reading sk_protocol field

mauriciovasquezbernal:
sk_protocol used to be a bitfield until[0], hence using BPF_CORE_READ_INTO() won't work on older kernels as it'd read a wrong value. From the BPF CO-RE reference guide[1] we have that:

"""
It's worth noting another important property of BPF_CORE_READ_BITFIELD() and BPF_CORE_READ_BITFIELD_PROBED(). They can read not just bitfields, but any integer field as well. Whatever the actual nature of a field (bitfield or integer of any size up to 8 bytes), macros return properly signed-extended 8-byte integers back. It keeps working even if a field changes from integer to bitfield and vice versa. It keeps working if the field changes from int to u8. As such, BPF_CORE_READ_BITFIELD() macros is a universal way to read any integer field regardless of its nature or size.
"""

Hence, this commit switches to use BPF_CORE_READ_BITFIELD_PROBED() to be able to read that field on older kernels.

[0]: https://github.com/torvalds/linux/commit/bf9765145b856fa2e238a5b8a54453795ba30ad6
[1]: https://nakryiko.com/posts/bpf-core-reference-guide/#reading-bitfields-and-integers-of-varying-sizes


alban (approved):
Nice finding!
LGTM

PR #1330, opened 2023-02-16
Update to golang 1.19

mauriciovasquezbernal:
The only specific reason to update is to have [0] in that automatically increases the number of open file descriptors we can have. I didn't updated to 1.20 because the base docker images we use don't have this version available yet.

### DIFF
--- a/pkg/client/clientset/versioned/fake/register.go
+++ b/pkg/client/clientset/versioned/fake/register.go
@@ -35,14 +35,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
-//     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//     aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//   )
+//	import (
+//	  "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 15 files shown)


pkg/client/clientset/versioned/fake/register.go
```diff
@@ -35,14 +35,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
...
+//	import (
+//	  "k8s.io/client-go/kubernetes"
+//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
+//	)
 //
-//   kclientset, _ := kubernetes.NewForConfig(c)
-//   _ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
+//	kclientset, _ := kubernetes.NewForConfig(c)
+//	_ = aggregatorclientsetscheme.AddToScheme(clientsetscheme.Scheme)
```
alban, line 45: Should that be in a separate PR? Ditto for register.go

alban (approved):
LGTM, modulo the change to generated files.

Thanks for fixing the issue

pkg/client/clientset/versioned/fake/register.go
mauriciovasquezbernal: I can move that to a different commit, however this should be on the same PR otherwise the tests won't pass.

alban: Oh ok, I didn't realise.

PR #1332, opened 2023-02-17
CI: Fix docker upgrade for cri-o

mqasimsarfraz:
Currently we need `docker-ce` >= 23.0.0 for minikube to work for `cri-o` runtime. Something changed in upstream docker packages and now we need to specify `containerd.io` in packages to avoid conflicts. Should have done in the first place as mentioned on their [guide](https://docs.docker.com/engine/install/ubuntu/). :see_no_evil:  

Hopefully, GH action will upgrade to `docker-ce` 23.0.0 and we can get rid of this workaround.

## Failed logs:

```
The following packages have unmet dependencies:
 containerd.io : Conflicts: containerd
                 Conflicts: runc
 moby-containerd : Conflicts: containerd
                   Conflicts: containerd.io
 moby-runc : Conflicts: runc
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
Error: Process completed with exit code 100.
```


mqasimsarfraz:
Thanks for the quick review!

PR #1331, opened 2023-02-17
ci: Add dependabot CVE checks

eiffel-fl:
Hi.

This PR does 3 things:

1. It enables `dependabot` to check our `golang` dependencies do not contain CVE daily as suggested by Burak.
1. It enables `dependabot` to check our `docker` dependencies do not contain CVE daily as suggested by Burak.
1. When a PR is opened, it adds a check to ensure the dependencies added by this PR do not container CVE.

Best regards.

PR #1335, opened 2023-02-17
CI: Use trivy for docker image vulnerability scanning

mqasimsarfraz:
Use [trivy](https://github.com/aquasecurity/trivy) to scan container images, job will fail in case we find any `CRITICAL/HIGH` vulnerabilities. We can also ignore the known vulnerabilities using following in case we want the CI to be green:

```
# ignore-unfixed: true
# ignore-policy: .github/trivy/ignore-policy.yaml
```

## Samples:
- [Alpine](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/4204302104/jobs/7294829278#step:6:33)
- [Ubuntu](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/4204302104/jobs/7294829178#step:6:34)


mqasimsarfraz:
> Can you please `cherry-pick` this one: [642cb3e](https://github.com/inspektor-gadget/inspektor-gadget/commit/642cb3e22a35c71544610b02c99970c07b218889) It should address the CVEs in our binaries.

Great. The CVES seem to be fixed now. Alpine is all green now! :green_circle:

mauriciovasquezbernal:
How is this related to https://github.com/inspektor-gadget/inspektor-gadget/pull/1331? Is there any overlap in the CVE scan for the container images? Do we need / want to keep both of them?

mqasimsarfraz:
> How is this related to #1331? Is there any overlap in the CVE scan for the container images? Do we need / want to keep both of them?

IIUC #1331 takes care of dependencies via dependabot and code scanning via codeql. But we still need to address the OS in docker images. For simple images like `alpine` I feel like `dependabot`  might be good enough but for images like `bcc` this is bit out of scope for `dependabot`. 

Having said that, I am not sure if we should run this on each PR so other options would be:
- Running this weekly in a separate workflow.
- Move this to https://github.com/kinvolk/bcc, scanning the base image directly.

What do you think?

mauriciovasquezbernal:
> Having said that, I am not sure if we should run this on each PR so other options would be:

I think it's better to have it on this repo.

mauriciovasquezbernal:
> which is not the case with kinvolk/bcc as we hardcode the image hash.

I suppose we can change it to push to `latest` and always fetch it?

mauriciovasquezbernal:
I don't get it, if we build and push the BCC image daily, and use latest in the Inspektor Gadget dockerfile, won't it work?

mqasimsarfraz:
> I think it is better to have both merged and we can revert one later if there is any overlap.

Agreed, better to start with both. I [cherry-picked](https://github.com/inspektor-gadget/inspektor-gadget/commit/2dcca772d2374015e569440970a02e58408bee84) to [verify](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4231145988/jobs/7349420111#step:6:34) the vulnerability results.

Do you want to merge the cherry-picks in the same PR? I think once we agree on bcc image tag this PR is ready to go in?

.github/workflows/inspektor-gadget.yml
```diff
@@ -296,6 +296,42 @@ jobs:
...
+    name: Scan gadget container images
+    # level: 1
+    needs: build-gadget-container-images
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        type: [ default, core ]
+        os: [ linux ]
+        platform: [ amd64 ]
```
mauriciovasquezbernal, line 309: Out of curiosity: why not arm64 too?

mqasimsarfraz: Nothing particular. I assumed doing it just for `amd64` would be enough. What do you think?

mauriciovasquezbernal: Yes, I think only checking a specific architecture should be good enough. AFAIU it'd very very weird to have a CVE only for one architecture.

Issue #1334, opened 2023-02-17
Network tracing gadget doesn't capture traffic from host network pods


johananl:
The `trace network` gadget currently doesn't capture traffic from pods with `hostNetwork` set to `true`. This is intentional:

https://github.com/inspektor-gadget/inspektor-gadget/blob/459878ae5d829370722752439697ed3d7c2f66ee/pkg/gadget-collection/gadgets/trace/network/gadget.go#L198-L201

Other similar tracing gadgets don't filter out host network traffic. For consistency, this gadget shouldn't either (according to @alban).

PR #1336, opened 2023-02-20
trace network: don't hide host netns traffic

alban:
The trace network gadget currently doesn't capture traffic from pods with hostNetwork set to true. This was intentional because this gadget was initially designed for the needs of the network policy advisor. Kubernetes network policies don't apply on the host network namespace, so capturing the host network traffic was not necessary. However, the trace network gadget has use cases outside of the network policy advisor.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/1334

TODO:
- [x] Check if it works both with single-container pods and multiple-container pods

cc @johananl

### DIFF
--- a/pkg/gadgets/trace/network/types/types.go
+++ b/pkg/gadgets/trace/network/types/types.go
@@ -41,7 +41,7 @@ type Event struct {
 	PodHostIP string            `json:"podHostIP,omitempty" column:"podhostip,template:ipaddr,hide"`
 	PodIP     string            `json:"podIP,omitempty" column:"podip,template:ipaddr,hide"`
 	PodOwner  string            `json:"podOwner,omitempty" column:"podowner,hide"`
-	PodLabels map[string]string `json:"podLabels,omitempty" column:"padlabels,hide"`
+	PodLabels map[string]string `json:"podLabels,omitempty" column:"podlabels,hide"`
 
 	/* Remote */
 	RemoteKind RemoteKind `json:"remoteKind,omitempty" column:"remoteKind,maxWidth:5,hide"`
--- a/cmd/local-gadget/trace/network.go
+++ b/cmd/local-gadget/trace/network.go
@@ -142,11 +142,6 @@ func printEvents(
 	}
 
 	for _, event := range newEvents {
-		// for now, ignore events on the host netns
-		if event.Container == "" {
-			continue
-		}
-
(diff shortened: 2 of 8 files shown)


johananl:
LGTM. I've tested the following:

- Capture traffic on a host network pod with a single container.
- Capture traffic on a host network pod with two containers.

Note that when specifying a pod using `./kubectl-gadget trace network -p` and that pod is a host network pod, **all** host network traffic from the relevant node is captured.

alban:
Thanks for testing! I'll wait for another review on the code. Some points of concern:
- The HostNetwork is added in CommonData so it's available in all gadgets. But it is initialized only in networking gadgets via EnrichByNetNs(). It impacts the documentation generation. I didn't see how to do it differently because EnrichByNetNs() does not know about the specific event types. Any ideas how to improve it? Should I just add one line in EnrichByMntNs() and assume that all gadgets call either EnrichByMntNs() or EnrichByNetNs()?

blanquicet (commented):
I'm unsure how urgent we want to merge this as it could be done more cleanly after merging https://github.com/inspektor-gadget/inspektor-gadget/pull/1281. See my comments.

pkg/gadgets/trace/network/types/types.go
```diff
@@ -41,7 +41,7 @@ type Event struct {
 	PodHostIP string            `json:"podHostIP,omitempty" column:"podhostip,template:ipaddr,hide"`
 	PodIP     string            `json:"podIP,omitempty" column:"podip,template:ipaddr,hide"`
 	PodOwner  string            `json:"podOwner,omitempty" column:"podowner,hide"`
-	PodLabels map[string]string `json:"podLabels,omitempty" column:"padlabels,hide"`
+	PodLabels map[string]string `json:"podLabels,omitempty" column:"podlabels,hide"`
```
alban, line 44: It seems this PR will take a bit more time. Meanwhile, I am fixing this typo in a separate PR:
https://github.com/inspektor-gadget/inspektor-gadget/pull/1339

blanquicet:
We can restart working on this PR

blanquicet:
This is my proposal: https://github.com/inspektor-gadget/inspektor-gadget/commit/c3b87340df454dbcb9fd1db6e2fb79665caa624c

Issue #1319, opened 2023-02-13
[RFE] trace capabilities gadget: add context such as stacks or syscall


alban:
## Current situation

`kubectl gadget trace capabilities` tells which capability is exercised by processes & containers but it does not currently say which syscall triggered it or what is the kernel or user stack during the check.

## Impact

Users can see `SYS_ADMIN` but they don't know if it was during a `mount` call, or any other reason. It is difficult to understand why the capability was exercised.

## Ideal future situation

From the easiest to the more difficult:

1. An option `--print-stack` would give the kernel stack.
2. Optional because more difficult: users would get the user stack too.
3. There is a "SYSCALL" field telling which syscall trigger the capability check, possibly with the arguments, as follows:

```
$ kubectl gadget trace capabilities --selector name=set-priority
... CONTAINER        PID     COMM   UID  CAP  CAPNAME   AUDIT  VERDICT SYSCALL
... set-priority     110385  nice   0    23   SYS_NICE  1      Deny    setpriority(PRIO_PROCESS, 0, -20)
```

### Stack field

We used to have that option, see:
```
$ git grep print-stack
```
But it was removed in https://github.com/inspektor-gadget/inspektor-gadget/pull/524

bcc has this features in both version of the `capable` tool:
* [libbpf-tools/capable.bpf.c](https://github.com/iovisor/bcc/blob/47502c2350b5de01dbcd35170e5542f2bc91118f/libbpf-tools/capable.c#L112-L113)
* [tools/capable.py](https://github.com/iovisor/bcc/blob/master/tools/capable.py#L43-L46)

### Syscall field

For the "SYSCALL" field, we could use a separate tracepoint on sys_enter, record the syscall id & parameters in a hash map indexed by the pid_tgid and use it in the cap_capable kprobe if necessary.

## Additional information

See also related discussion #173

cc @kranurag7

PR #1337, opened 2023-02-20
trace capabilities: add new fields

alban:
# trace capabilities: add new fields

This patch adds the following fields:
- Cap Effective: the effective capability bitfield from the objective context of the process
- Current User Ns: the user namespace inode of the process
- Target User Ns: the user namespace that the kernel used to test the capability

The new fields are not printed by default (column's attribute 'hide').

## How to use

Example of output:
```
$ go run -exec sudo -tags withebpf ./cmd/local-gadget/... trace capabilities -r docker -c test -o json
{"container":"test","timestamp":1676828381315749882,"type":"normal","pid":3997838,"comm":"chroot","syscall":"chroot","cap":18,"capName":"SYS_CHROOT","audit":1,"verdict":"Allow","insetid":false,"mountnsid":4026532914,"targetuserns":4026531837,"currentuserns":4026531837,"capeffective":"3fffffffff"}
```
In this example, targetuserns and currentuserns are the same. This is necessarily the case for chroot because the kernel tests the capability in this way:
```
    if (!ns_capable(current_user_ns(), CAP_SYS_CHROOT))
```
Other example:
```
$ go run -exec sudo -tags withebpf ./cmd/local-gadget/... trace capabilities -r docker -c test -o json
{"container":"test","timestamp":1676828926093698387,"type":"normal","pid":4016919,"comm":"unshare","syscall":"unshare","cap":21,"capName":"SYS_ADMIN","audit":1,"verdict":"Allow","insetid":false,"mountnsid":4026532894,"targetuserns":4026533311,"currentuserns":4026531837,"capeffective":"3fffffffff"}
{"container":"test","timestamp":1676828926095002774,"type":"normal","pid":4016949,"comm":"mount","syscall":"mount","cap":21,"capName":"SYS_ADMIN","audit":1,"verdict":"Deny","insetid":false,"mountnsid":4026532894,"targetuserns":4026531837,"currentuserns":4026533311,"capeffective":"0"}
```
Generated with the following command:
```
$ docker run --rm -ti --privileged --name=test busybox unshare -Upf mount -t tmpfs tmpfs /tmp
```
In this last line, targetuserns and currentuserns are different. This is because the kernel tests the capability with regards to other user namespaces:
```
    if (!ns_capable(mnt_ns->user_ns, CAP_SYS_ADMIN) || ...
```
The Cap Effective field can be decoded with the following command:
```
$ capsh --decode=3fffffffff
0x0000003fffffffff=cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,cap_wake_alarm,cap_block_suspend,cap_audit_read
```

## Testing done

TBD

## TODO

- [x] Update guide

------

This PR addresses a part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1319

cc @kranurag7


alban:
The error in the CI seems unrelated to this PR:
```
+ kubectl logs -n gadget pod/gadget-jlgvt
/entrypoint.sh: line 34: dpkg-query: not found
```
Is it a recent change?

pkg/gadgets/trace/capabilities/tracer/tracer_test.go
```diff
@@ -142,18 +150,22 @@ func TestCapabilitiesTracer(t *testing.T) {
...
+					Pid:           uint32(info.Pid),
+					UID:           uint32(info.UID),
+					Comm:          info.Comm,
+					Syscall:       "bind",
+					CapName:       "NET_BIND_SERVICE",
+					Cap:           10,
+					Audit:         1,
+					InsetID:       &false_,
+					Verdict:       "Allow",
+					CurrentUserNs: 1,
```
mauriciovasquezbernal, line 166: We can also test this by using something like -> https://github.com/inspektor-gadget/inspektor-gadget/commit/6b490ea269c704240a3fd1162fe475403afc1d64. Our runner isn't unsharing the user ns, hence we can assume `TargetUserNs` is always equal to `CurrentUserNs` in these tests? 

I think we can do the same for CapEffective, I was checking https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap#Set but didn't find a way to get the bitmask of the capabilities.

pkg/gadgets/trace/capabilities/tracer/bpf/capable.h
```diff
@@ -9,6 +9,9 @@
 
 struct cap_event {
 	__u64	mntnsid;
+	__u64	current_userns;
+	__u64	target_userns;
+	__u64   cap_effective;
```
mauriciovasquezbernal, line 14: It's defined as u32 in the kernel, any reason to use a different width?

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
```diff
@@ -129,6 +132,9 @@ int BPF_KPROBE(ig_trace_cap_e, const struct cred *cred, struct user_namespace *t
 
 
 	struct args_t args = {};
+	args.current_userns = (u64) BPF_CORE_READ(task, real_cred, user_ns, ns.inum);
+	args.target_userns = (u64) BPF_CORE_READ(targ_ns, ns.inum);
+	BPF_CORE_READ_INTO(&args.cap_effective, task, real_cred, cap_effective.cap[0]);
```
mauriciovasquezbernal, line 137: Just wondering why `BPF_CORE_READ` is not used in all cases? is there any special on the last one?

mauriciovasquezbernal (commented):
I like it. It's a nice improvement for users!

I left some comments, will make another pass later on.

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
```diff
@@ -176,6 +184,8 @@ int BPF_KRETPROBE(ig_trace_cap_x)
 
 	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, &event, sizeof(event));
 
+	bpf_map_delete_elem(&start, &pid_tgid);
```
alban, line 187: If a ebpf program reads a hash map value via the pointer returned by `bpf_map_lookup_elem` after the entry has been deleted via bpf_map_delete_elem, then it might read garbage, unless we use BPF_F_NO_PREALLOC.

Source:
https://lore.kernel.org/lkml/CAG48ez1-WZH55+Wa2vgwZY_hpZJfnDxMzxGLtuN1hG1z6hKf5Q@mail.gmail.com/T/

In this patch, I do both:
- move bpf_map_delete_elem at the end of the function
- use BPF_F_NO_PREALLOC: on my computer there was about 800 threads sleeping in futex/poll syscalls, so the previous limit of 1024 can easily be reached. I increased it to 1024*1024. This map is only used in process context were it is allowed to allocate, so it is safe to use BPF_F_NO_PREALLOC.

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
alban: The amount of bytes copied is different between BPF_CORE_READ and BPF_CORE_READ_INTO:
* BPF_CORE_READ looks at the type of the source variable
* BPF_CORE_READ_INTO looks at the type of the destination variable

For ns.inum, we use BPF_CORE_READ to read "unsigned int inum"  (size: 4 bytes), cast to u64 (size: 8 bytes). The cast just set the 4 most significant bytes to zero. If we were using BPF_CORE_READ_INTO, we would read 4 bytes of the next field in the struct ns_common (`count`):
```
struct ns_common {
        atomic_long_t stashed;
        const struct proc_ns_operations *ops;
        unsigned int inum;
        refcount_t count;
};
```

For cap_effective, we use BPF_CORE_READ_INTO to read 8 bytes (size of the destination variable "args.cap_effective"). We cannot use BPF_CORE_READ to read `cap_effective` in one instruction because of the way it is defined: there are no 8 bytes field but an array. I couldn't get it to work with BPF_CORE_READ.

```
struct cred {
    ...
    kernel_cap_t    cap_effective;
}
#define _LINUX_CAPABILITY_U32S_3     2
#define _KERNEL_CAPABILITY_U32S    _LINUX_CAPABILITY_U32S_3
typedef struct kernel_cap_struct {
        __u32 cap[_KERNEL_CAPABILITY_U32S];
} kernel_cap_t;
```

pkg/gadgets/trace/capabilities/tracer/bpf/capable.h
alban: I can change the *_userns fields to u32. But the cap_effective needs to stay 8 bytes.

alban:
Thanks for the review! I will address the comments tomorrow.

Meanwhile I updated the guide in the documentation.

pkg/gadgets/trace/capabilities/types/types.go
```diff
@@ -34,16 +34,19 @@ const (
...
+	UID           uint32 `json:"uid,omitempty" column:"uid,minWidth:6"`
+	Cap           int    `json:"cap,omitempty" column:"cap,width:3,fixed"`
+	CapName       string `json:"capName,omitempty" column:"capName,width:18,fixed"`
+	Audit         int    `json:"audit,omitempty" column:"audit,minWidth:5"`
+	Verdict       string `json:"verdict,omitempty" column:"verdict,width:7,fixed"`
+	InsetID       *bool  `json:"insetid,omitempty" column:"insetid,width:7,fixed,hide"`
+	MountNsID     uint64 `json:"mountnsid,omitempty" column:"mntns,template:ns"`
+	TargetUserNs  uint64 `json:"targetuserns,omitempty" column:"targetuserns,template:ns"`
+	CurrentUserNs uint64 `json:"currentuserns,omitempty" column:"currentuserns,template:ns"`
+	CapEffective  string `json:"capeffective,omitempty" column:"capeffective,hide"`
```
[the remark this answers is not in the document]
alban: Using a hexadecimal string is nice because it can be copy/pasted into `capsh --decode=`  or even read manually by seeing bit 0 to bit 40 set (41 bits total) set:

```
$ cat /proc/sys/kernel/cap_last_cap
40
```

And it would be difficult for a custom extractor to print the very long string like the following:
> cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,cap_wake_alarm,cap_block_suspend,cap_audit_read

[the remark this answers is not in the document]
mauriciovasquezbernal: What I meant is to transport the data using an unsigned integer and then print to the user as hexadecimal. (It's actually not that important, but I feel bad when we use strings to transport something that can be represented with an integer).

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
mauriciovasquezbernal: Ah, I was confused. I thought that we're reason the first element of `cap` in `kernel_cap_struct`, but for some reason the caps are represented as an array of two u32s in the kernel. This also answers my question about the data type above. Thanks!

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
mauriciovasquezbernal: > This map is only used in process context were it is allowed to allocate, so it is safe to use BPF_F_NO_PREALLOC.

I'm not sure I follow it. AFAIU it was not safe to use BPF_F_NO_PREALLOC (https://github.com/torvalds/linux/commit/94dacdbd5d2d), is there a condition where it's totally safe to use them?

alban: Thanks for the link, I didn't know.

What implementation should we do? Some options:
1. Don't use BPF_F_NO_PREALLOC and increases max_entries to a higher value (1024*1024?) and accept the big memory usage
2. Use BPF_MAP_TYPE_TASK_STORAGE but requires Linux 5.11
3. Avoid the sys_enter tracepoint entirely and get the syscall nr via other means (see [example](https://github.com/torvalds/linux/blob/89f5349e0673322857bd432fa23113af56673739/lib/syscall.c#L20-L29))

I would like to try option 3 at some point but for this PR, I will try option 1 to keep the changes small.

pkg/gadgets/trace/capabilities/types/types.go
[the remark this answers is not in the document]
alban: I made it in two columns:
* caps: as uint64, with a column extractor to display the hexadecimal
* capsnames, with a comma-separated list, following the same model as the existing columns cap and capname, so other GUIs don't have to decode it. It's hidden by default, so the long list is not bothering.

pkg/gadgets/trace/capabilities/tracer/tracer_test.go
alban: > We can also test this by using something like -> https://github.com/inspektor-gadget/inspektor-gadget/commit/6b490ea269c704240a3fd1162fe475403afc1d64.

Thanks! I applied your patch.

> Our runner isn't unsharing the user ns, hence we can assume TargetUserNs is always equal to CurrentUserNs in these tests?

Probably: neither the runner nor the workload (generateEvent) create user namespaces. But I didn't change this. Do you want me to add a test for this?

> I think we can do the same for CapEffective, I was checking https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap#Set but didn't find a way to get the bitmask of the capabilities.

We can probably do that with the gocapability package. I now use this in the capsNames() function. Could this be done in a follow-up PR?

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
mauriciovasquezbernal: I think it's fine to go with 1 for the time being.

pkg/gadgets/trace/capabilities/tracer/tracer_test.go
mauriciovasquezbernal: > But I didn't change this. Do you want me to add a test for this?

Yes, I think it's better to check if those are equal instead of relying that they are not empty. (If for any reason we mess up the tracer to read garbage the tests will pass anyway). 

> We can probably do that with the gocapability package. I now use this in the capsNames() function. Could this be done in a follow-up PR?

For sure, would you mind opening an issue to avoid forgetting about this?

go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/moby/moby v20.10.20+incompatible
 	github.com/stretchr/testify v1.8.0
+	github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
```
mauriciovasquezbernal, line 47: https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap seems to be the official package from the kernel. Is it possible to use it?

pkg/gadgets/trace/capabilities/tracer/tracer.go
```diff
@@ -282,16 +302,20 @@ func (t *Tracer) run() {
...
+			Cap:           int(bpfEvent.Cap),
+			UID:           bpfEvent.Uid,
+			Audit:         audit,
+			InsetID:       insetID,
+			Comm:          gadgets.FromCString(bpfEvent.Task[:]),
+			Syscall:       syscall,
+			CapName:       capabilityName,
+			Verdict:       verdict,
+			Caps:          bpfEvent.CapEffective,
+			CapsNames:     capsNames(bpfEvent.CapEffective),
```
mauriciovasquezbernal, line 318: I think it's interesting to send the full list of capabilities so the caller doesn't have to parse them, but, at the same time I'm a bit confused about performance, it seems we're sending too much data for each event and some clients could parse it to reduce the network usage. 

Perhaps this could  be configurable, on by default, and kubectl-gadget could disable it. @flyth do you have any opinions on this kind of situation?

btw, I'm completely fine if we handle this later on.

mauriciovasquezbernal (commented):
Only one blocking comment on the libraries for caps.

go.mod
mauriciovasquezbernal: This is also true that https://github.com/syndtr/gocapability is used by a lot of projects, so up to you which one to use.

pkg/gadgets/trace/capabilities/tracer/tracer.go
flyth: I also worry a bit about these extremely long strings. Making it configurable sounds good - we can easily do that with the upcoming kubectl upgrade PR and have it convert to strings on the client by default. So I'd propose to leave it as it is for now and optimize once the other PR is merged.

alban: Ok, I will keep it as is.

I am not sure it is worth optimising for this specific gadget: the stream of capability events are not high throughput like it would be for a tcpdump gadget.

go.mod
alban: syndtr/gocapability  is also used in runc.

The other lib has a String() method but it seems only have to describe all the full capability sets (effective, inheritable, etc.). That could be useful to add in a future version of the gadget but for now I only give the effective set since it is the set used by the kernel to check the capability.

So I will keep syndtr/gocapability for now.

pkg/gadgets/trace/capabilities/types/types.go
```diff
@@ -34,16 +34,20 @@ const (
...
+	Cap           int    `json:"cap,omitempty" column:"cap,width:3,fixed"`
+	CapName       string `json:"capName,omitempty" column:"capName,width:18,fixed"`
+	Audit         int    `json:"audit,omitempty" column:"audit,minWidth:5"`
+	Verdict       string `json:"verdict,omitempty" column:"verdict,width:7,fixed"`
+	InsetID       *bool  `json:"insetid,omitempty" column:"insetid,width:7,fixed,hide"`
+	MountNsID     uint64 `json:"mountnsid,omitempty" column:"mntns,template:ns"`
+	TargetUserNs  uint64 `json:"targetuserns,omitempty" column:"targetuserns,template:ns"`
+	CurrentUserNs uint64 `json:"currentuserns,omitempty" column:"currentuserns,template:ns"`
+	Caps          uint64 `json:"caps,omitempty" column:"caps,hide"`
+	CapsNames     string `json:"capsNames,omitempty" column:"capsnames,hide"`
```
mauriciovasquezbernal, line 50: Could it make sense to use `[]string` and let each "frontend" to format as they want?

mauriciovasquezbernal:
I'll review again once the conflicts are fixed. Thanks!

docs/gadgets/trace/capabilities.md
```diff
@@ -142,6 +142,117 @@ You can now delete the pod you created:
...
+  "syscall": "chroot",
+  "cap": 18,
+  "capName": "SYS_CHROOT",
+  "audit": 1,
+  "verdict": "Allow",
+  "insetid": false,
+  "mountnsid": 4026533229,
+  "targetuserns": 4026531837,
+  "currentuserns": 4026531837,
+  "caps": "3fffffffff",
```
mauriciovasquezbernal, line 183: I think it needs to be updated after last change, now it's printed as an integer. Perhaps it's better to use `-o columns` to avoid user dealing with json.

mauriciovasquezbernal (approved):
One minor comment on the documentation and also wondering about https://github.com/inspektor-gadget/inspektor-gadget/pull/1337/files#r1114234263 

I'll approve since it already works fine.

docs/gadgets/trace/capabilities.md
alban: With -o custom-columns I cannot get the full field because it gets ellipsed. I don't have a monitor big enough to enlarge the terminal. And `export COLUMNS=400` does not seem to work.
cc @flyth 

I will update the output with json.

alban:
Updates:
- rebase & fix conflicts
- use `[]string` for `CapsNames`
- update documentation

pkg/gadgets/trace/capabilities/types/types.go
```diff
@@ -35,15 +36,19 @@ type Event struct {
...
+	UID           uint32   `json:"uid,omitempty" column:"uid,minWidth:6"`
+	Cap           int      `json:"cap,omitempty" column:"cap,width:3,fixed"`
+	CapName       string   `json:"capName,omitempty" column:"capName,width:18,fixed"`
+	Audit         int      `json:"audit,omitempty" column:"audit,minWidth:5"`
+	Verdict       string   `json:"verdict,omitempty" column:"verdict,width:7,fixed"`
+	InsetID       *bool    `json:"insetid,omitempty" column:"insetid,width:7,fixed,hide"`
+	TargetUserNs  uint64   `json:"targetuserns,omitempty" column:"targetuserns,template:ns"`
+	CurrentUserNs uint64   `json:"currentuserns,omitempty" column:"currentuserns,template:ns"`
+	Caps          uint64   `json:"caps,omitempty" column:"caps,hide"`
+	CapsNames     []string `json:"capsNames,omitempty" column:"capsnames,hide"`
```
alban, line 51: @flyth FYI, I named the column `targetuserns`, `currentuserns`, `capsnames` in lower case because it was done like this for some columns like `cumulruncount`. But there are other columns with different casing. Do you have a recommended casing?

docs/gadgets/trace/capabilities.md
```diff
@@ -142,6 +142,134 @@ You can now delete the pod you created:
...
+$ kubectl run -ti --rm --restart=Never \
+    --image busybox --privileged testcaps -- \
+    chroot /
+```
+
+```
+$ kubectl gadget trace capabilities \
+    -o custom-columns=comm,syscall,capName,verdict,targetuserns,currentuserns,caps,capsnames
+COMM             SYSCALL                      CAPNAME            VERDICT TARGETUSERNS        CURRENTUSERNS       CAPS                 CAPSNAMES
+chroot           chroot                       SYS_CHROOT         Allow   4026531837          4026531837          3fffffffff           chown,dac_override,dac_…
```
alban, line 167: In a future PR, I´d like to switch cap names like `SYS_CHROOT` to lower case to be consistent with other columns.

pkg/gadgets/trace/capabilities/types/types.go
flyth: The columns library itself is case insensitive and the headers are always uppercase when displayed right now - so the exact name can only be seen in the defaults / available columns output (or: it doesn't really matter right now). But we should definitely agree on a standard format and keep it consistent across all gadgets. I personally like camel case more than having all lower caps.

alban:
A weird issue, but seems unrelated to this PR:

```
$ kubectl run -ti --rm --restart=Never --image busybox testcaps -- /bin/sh
If you don't see a command prompt, try pressing enter.
/ # /bin/mount -t tmpfs tmpfs /tmp
mount: permission denied (are you root?)
```

```
$ ./kubectl-gadget trace capabilities     -o custom-columns=comm,syscall,capName,audit,verdict,targetuserns,currentuserns,caps,capsnames
PID        COMM             SYSCALL              CAPNAME            AUDIT              VERDICT TARGETUSERNS  CURRENTUSERNS CAPS              CAPSNAMES        
3388586    mount            mount                SYS_ADMIN          1                  Allow   4026531837    4026531837    a80425fb          chown,dac_overri…
3388586    mount            mount                SYS_ADMIN          1                  Deny    4026531837    4026531837    a80425fb          chown,dac_overri…
```

Notice that we (erroneously?) have two events for the mount() syscall: one time supposedly allowed and one time denied. This behaviour is only spotted when `kubectl-gadget` is started before the pod is started. If I restart the `kubectl-gadget` command before typing `/bin/unshare` in the other terminal, only one event is generated.

alban:
With bcc's capable, I could get the stacks of the two mount events:

```
21:58:48  0      3700134 mount            21   CAP_SYS_ADMIN        1     
        b'cap_capable+0x1 [kernel]'
        b'security_capable+0x40 [kernel]'
        b'capable+0x2f [kernel]'
        b'xattr_permission+0x192 [kernel]'
        b'vfs_getxattr+0x33 [kernel]'
        b'ovl_check_origin_xattr+0x2c [overlay]'
        b'ovl_get_inode+0x4d7 [overlay]'
        b'ovl_lookup+0x731 [overlay]'
        b'__lookup_slow+0x75 [kernel]'
        b'walk_component+0x103 [kernel]'
        b'path_lookupat+0x67 [kernel]'
        b'filename_lookup+0xbd [kernel]'
        b'user_path_at_empty+0x36 [kernel]'
        b'__x64_sys_mount+0x9e [kernel]'
        b'do_syscall_64+0x5b [kernel]'
        b'entry_SYSCALL_64_after_hwframe+0x44 [kernel]'
```
This first event happens when it looks up the target directory for the first time in the container overlayfs to cache the path.
The current process temporarily gets the credentials of the overlayfs superblock with override_creds/revert_creds. So the current process real credentials are irrelevant here. This event is not useful for the trace capabilities gadget because this is not influenced by the process real capabilities.

```
21:58:48  0      3700134 mount            21   CAP_SYS_ADMIN        1     
        b'cap_capable+0x1 [kernel]'
        b'security_capable+0x40 [kernel]'
        b'ns_capable+0x2b [kernel]'
        b'path_mount+0xa5 [kernel]'
        b'__x64_sys_mount+0xe2 [kernel]'
        b'do_syscall_64+0x5b [kernel]'
        b'entry_SYSCALL_64_after_hwframe+0x44 [kernel]'
```

This event is the real one.

See also:
* objective context and subjective context in [Credentials in Linux](https://kernel.org/doc/html/v6.2-rc8/security/credentials.html#accessing-task-credentials)
* struct task_struct's cred and real_cred fields:
  * task->cred: subjective creds (subjective=subject as in the subject who acts), temporarily replaced by other creds in overlayfs with override_creds/revert_creds.
  * task->real_cred: objective creds (objective=object as in the object that is being acted upon) / what's visible in procfs

In this current PR, I was using task->real_cred, so the results were confusing because those creds were not the ones used in the capability check. I will rework that.

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
```diff
@@ -95,6 +98,14 @@ int BPF_KPROBE(ig_trace_cap_e, const struct cred *cred, struct user_namespace *t
 	if (filter_by_mnt_ns && !bpf_map_lookup_elem(&mount_ns_filter, &mntns_id))
 		return 0;
 
+	const struct cred *real_cred = BPF_CORE_READ(task, real_cred);
+	if (cred != real_cred) {
+	    // the subjective credentials are in an overridden state with
+	    // override_creds/revert_creds (e.g. during overlayfs cache or copyup)
+	    // https://kernel.org/doc/html/v6.2-rc8/security/credentials.html#overriding-the-vfs-s-use-of-credentials
+	    return 0;
```
mauriciovasquezbernal, line 106: nit: it's using spaces instead of tabs.

mauriciovasquezbernal (approved):
LGTM with a small nit. 

Great catch regarding objective and subjective credentials.

PR #1338, opened 2023-02-20
entrypoint.sh: fix dpkg error on alpine

alban:
Symptoms:
```
+ kubectl logs -n gadget pod/gadget-jlgvt
/entrypoint.sh: line 34: dpkg-query: not found
```

See https://github.com/inspektor-gadget/inspektor-gadget/pull/1337#issuecomment-1436849662

### DIFF
--- a/Dockerfiles/gadget-core.Dockerfile
+++ b/Dockerfiles/gadget-core.Dockerfile
@@ -82,3 +82,5 @@ COPY hack/btfs /btfs/
 
 # Mitigate https://github.com/kubernetes/kubernetes/issues/106962.
 RUN rm -f /var/run
+
+ENV GADGET_IMAGE_FLAVOUR=core
--- a/gadget-container/entrypoint.sh
+++ b/gadget-container/entrypoint.sh
@@ -30,11 +30,16 @@ KERNEL=$(uname -r)
 echo -n "Kernel detected: "
 echo $KERNEL
 
-echo -n "bcc detected: "
-dpkg-query --show libbcc | awk '{print $2}' || true
+# The gadget-core image does not provide bcc.
+if [ "$GADGET_IMAGE_FLAVOUR" = "default" ] ; then
+	echo -n "bcc detected: "
+	dpkg-query --show libbcc | awk '{print $2}' || true
+fi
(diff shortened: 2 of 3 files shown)


gadget-container/entrypoint.sh
```diff
@@ -31,7 +31,13 @@ echo -n "Kernel detected: "
...
 
 echo -n "bcc detected: "
-dpkg-query --show libbcc | awk '{print $2}' || true
+if command -v dpkg-query &> /dev/null ; then
+  dpkg-query --show libbcc | awk '{print $2}' || true
+else
+  # The gadget-core image is based on alpine (without dpkg-query) and does
+  # not provide bcc
+  echo "not available"
+fi
```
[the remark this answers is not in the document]
alban: I would like to have something in the logs saying if it is using the default or core image. Maybe I could add that log line, and then use your suggestion to only print the bcc version on the bcc-supporting image?

Dockerfiles/gadget-core.Dockerfile
```diff
@@ -82,3 +82,5 @@ COPY hack/btfs /btfs/
 
 # Mitigate https://github.com/kubernetes/kubernetes/issues/106962.
 RUN rm -f /var/run
+
+ENV GADGET_IMAGE_FLAVOUR=core
```
mauriciovasquezbernal, line 86: nit: please add a line break at the end. (Same for the other dockerfile)

gadget-container/entrypoint.sh
```diff
@@ -30,12 +30,19 @@ KERNEL=$(uname -r)
 echo -n "Kernel detected: "
 echo $KERNEL
 
-echo -n "bcc detected: "
-dpkg-query --show libbcc | awk '{print $2}' || true
+# The gadget-core image is based on alpine (without dpkg-query) and does
+# not provide bcc.
```
mauriciovasquezbernal, line 34: I think we can just mention that the gadget-core image doesn't provide BCC.

mauriciovasquezbernal (approved):
Seems like good debug info. Two minor nits but LGTM.

PR #1339, opened 2023-02-21
trace network gadget: fix typo

alban:
Currently the column is called padlabels:
```
$ kubectl gadget trace network -o custom-columns=node,padlabels
NODE  PADLABELS
```

This patch renames it to podlabels.


blanquicet (approved):
LGTM!

Issue #311, opened 2021-09-23
Improve `--dump` command 


mauriciovasquezbernal:
The `--dump` command of the gadget tracer manager is intended to provide useful debugging information. It currently prints information about the containers, tracers and go routine stacks (https://github.com/kinvolk/inspektor-gadget/pull/307). The current format we use is difficult to understand, hence we should something more user friendly like yaml. Additionally it'd be nice to be able to specify the items to print, for instance by having a `--containers`, `--traces`, `--stacks` parameters.

hvm0901:
Can I work on this, @mauriciovasquezbernal? However, I am new here, so could you please direct me to the files/source code that need modifications? I am currently understanding the codebase and architecture of Inspektor-Gadget. Thank you :)

HarshvMahawar:
Can I work on this, @mauriciovasquezbernal? However, I am new here, so could you please direct me to the files/source code that need modifications? I am currently understanding the codebase and architecture of Inspektor-Gadget and believe working on this issue can help me greatly. Thank you :)

mauriciovasquezbernal:
Sure, I'll assign it to you.

The implementation of this command is on https://github.com/inspektor-gadget/inspektor-gadget/blob/bf0836272a615a373f81934796dc24207ff92419/gadget-container/gadgettracermanager/main.go#L207-L236, and the server part is on https://github.com/inspektor-gadget/inspektor-gadget/blob/bf0836272a615a373f81934796dc24207ff92419/pkg/gadgettracermanager/gadgettracermanager.go#L217-L235

Currently this command is only available on the gadgettracermanager, so it has to be executed like:

```bash 
$ kubectl -n gadget exec -it $PODNAME -- /bin/gadgettracermanager -dump all
....
```

hvm0901:
In what formats should the information about the containers, tracers, and stacks be printed (with and without using different flags for different items)? Can you suggest some?

HarshvMahawar:
Implementing a yaml-like output format is a great idea! We can do it like these:
1. Saving the output in a dump file
```bash
kubectl -n gadget exec -it $PODNAME -- /bin/gadgettracermanager -dump all > dump.yaml
```
2. Adding an extra flag `--output` to directly write the output to the specified file within the command.

What do you suggest?

mauriciovasquezbernal:
I think it can always go to standard output. The user can redirect it to a file if they want. (So I'll go for option 1 you mentioned)

[...]

PR #1341, opened 2023-02-21
Improve gadgettracermanager -dump

claudiamarcubina:
# Adding `containers, traces, stacks, all` options to `gadgettracermanager -dump=`

This PR addresses a part  of [#311](https://github.com/inspektor-gadget/inspektor-gadget/issues/311). The `gadgettracermanager -dump=` command is enhanced with the following three options for specifying what item to print:
* `gadgettracermanager -dump=containers`
* `gadgettracermanager -dump=traces`
* `gadgettracermanager -dump=stacks`
* `gadgettracermanager -dump=all`

## Testing done

Changes have been tested locally with minikube using: `make minikube-deploy`. Example of output:

image


alban (approved):
LGTM, thanks!

@flyth You had other changes planned in grpc. Does this PR fit with your plans?

flyth:
> @flyth You had other changes planned in grpc. Does this PR fit with your plans?

The other PR doesn't touch the existing interfaces and also not the dump functionality, so it should be fine. 👍

mauriciovasquezbernal (approved):
Seems good. Thanks!

gadget-container/gadgettracermanager/main.go
```diff
@@ -184,13 +185,37 @@ func main() {
...
+
+		case "all":
+			fmt.Println(out.Containers, out.Traces, out.Stacks)
+			os.Exit(0)
+
+		default:
+			fmt.Printf("invalid method %q\n", method)
+			flag.PrintDefaults()
+			os.Exit(1)
+		}
```
blanquicet, line 217: I'm okay with either solution, but just for curiosity, why not pass what we want (containers, traces, stacks or all) in the `DumpStateRequest` and make the `DumpState` return a status (succeed or error) and a general string with the actual output?

claudiamarcubina: Thanks for the suggestion, Jose. It's a good idea and I'll keep that in mind for further approaches. Honestly, the reason why I didn't do that is because I didn't think of it. I was just happy to understand something for the first contribution 😄

blanquicet: Absolutely! I think your solution works perfectly fine and I'm also happy for your first contribution 🚀 ... It is just open question because I don't know what would be the best approach and would like to have the opinion of other folks for future contributions. I don't have much experience with protobuf + gRPC and this kind of discussion is one of my main source of knowledge 🙂

mauriciovasquezbernal: It'll be more efficient to tell the server which information the client is interested in and only return it. However, this API is for debugging purposes (not part of any "hot-path"), and additionally it's used locally through a unix socket, hence performance is not a big deal on this one.

Issue #880, opened 2022-08-29
Handle when KUBECONFIG contains several paths


eiffel-fl:
## Description

An error is thrown when `KUBECONFIG` contains several paths.

## Impact

Running Inspektor Gadget with `KUBECONFIG` containing several paths leads to an error.

## Environment and steps to reproduce

```bash
$ export KUBECONFIG="foo:bar:~/.kube/config"
$ ./kubectl-gadet trace exec -A
Error: failed to receive stream on node "node": stat /tmp/kubeconfig-1eb78c20f155a35a973c9452f25dc546:foo:bar:/home/francis/.kube/config: no such file or directory
```

## Expected behavior

Inspktor Gadget should run even if `KUBECONFIG` contains several paths.

## Additional information

We need to tweak the following code:
https://github.com/kinvolk/inspektor-gadget/blob/7c33be3122bdef9ac0fcc96d980b0ef62d3eb94c/pkg/k8sutil/client.go#L29

PR #1351, opened 2023-02-22
kubectl-gadget: Remove viper

mqasimsarfraz:
As part of fixing #132 #880 we stopped using `viper` but I somehow missed removing it as a dependency. Also, I don't see any occurrence of `viper.GetString()` in the code base so I assume it is safe to remove this? 

# Testing done

```bash
$ mv ~/.kube/config ~/kube/config.test
$ ./kubectl-gadget trace open # fails since no default config found
NODE                                  NAMESPACE                             POD                                   CONTAINER                             PID        COMM             FD   ERR PATH                                   
Error: failed to run gadget: error creating trace: listing pods: Get "http://localhost:8080/api/v1/namespaces/gadget/pods?labelSelector=k8s-app%3Dgadget": dial tcp 127.0.0.1:8080: connect: connection refused
$ KUBECONFIG=~/.kube/config.test ./kubectl-gadget trace open # works fine
$ KUBECONFIG=~/.kube/config.test:~/.kube/config.test ./kubectl-gadget trace open # works fine
$  ./kubectl-gadget trace open --kubeconfig ~/.kube/config.test # works fine

```

### DIFF
--- a/cmd/kubectl-gadget/utils/flags.go
+++ b/cmd/kubectl-gadget/utils/flags.go
@@ -19,24 +19,18 @@ import (
 	"fmt"
 	"strings"
 
-	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
 )
 
 var KubernetesConfigFlags = genericclioptions.NewConfigFlags(false)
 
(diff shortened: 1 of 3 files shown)


mauriciovasquezbernal (approved):
I see that `KUBECONFIG=/foo ./kubectl-gadget version` works fine too. So LGTM. 

However, do you know why it work if `AutomaticEnv` is not used?

mqasimsarfraz:
> However, do you know why it work if `AutomaticEnv` is not used?

IIUC [client-go](https://github.com/kubernetes/client-go/blob/v0.25.5/tools/clientcmd/loader.go#L143) is already taking care of  using [`KUBECONFIG`](https://github.com/kubernetes/client-go/blob/v0.25.5/tools/clientcmd/loader.go#L42) env internally so we don't need to bind any env to `kubeconfig` flag. In past we were using `viper` to [get the binded value from env](https://github.com/inspektor-gadget/inspektor-gadget/pull/926/files#diff-af6a6df6cbac4e1255d903ab2c3ad2c942d5575489d76a05ff98cc100fc6a6b8L29) but that was not necessary and removed.

mauriciovasquezbernal:
Great, thanks for the clarification. https://github.com/kubernetes/client-go/blob/v0.25.5/tools/clientcmd/loader.go#L143 is what I was looking for.

PR #1352, opened 2023-02-22
ci: Only run dependencies checks for PR on PR.

eiffel-fl:
Hi.

This PR fixes a commit introduced in a previous PR to only run dependencies checks for PR on... PR :sweat_smile:.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -84,6 +84,8 @@ jobs:
   dependencies-checks:
     name: Dependencies checks
     # level: 0
+    # Only run this on PR.
+    if: ${{ github.event.pull_request }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


mauriciovasquezbernal (approved):
LGTM!

.github/workflows/inspektor-gadget.yml
```diff
@@ -84,6 +84,8 @@ jobs:
   dependencies-checks:
     name: Dependencies checks
     # level: 0
+    # Only run this on PR.
+    if: ${{ github.event.pull_request }}
```
blanquicet, line 88: Just to be sure I got the goal of this PR. The whole CI workflow only runs on (1) PRs, (2) releases (tagged `v*`) and (3) pushes to the `main` branch. It means that the aim of this new conditional is to speed up the runs for (2) and (3) as the check was already done on PRs before merging/releasing?

blanquicet: It would be great if the comment "Only run this on PR" is extended to describe the reason you just said.

PR #1354, opened 2023-02-22
Better validation for -o columns

mauriciovasquezbernal:
Print a warning in case the user passes columns that are not available:

Before:
```
$ sudo ./local-gadget  trace exec -o columns=container,sdasadafd
CONTAINER
...
```

Now:
```
$ sudo ./local-gadget  trace exec -o columns=container,sdasadafd
WARN[0000] column "sdasadafd" is invalid
CONTAINER
```

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -238,7 +238,15 @@ func buildCommandFromGadget(
 
 			formatter := parser.GetTextColumnsFormatter()
 			if outputModeParams != "" {
-				formatter.SetShowColumns(strings.Split(outputModeParams, ","))
+				valid, invalid := parser.VerifyColumnNames(strings.Split(outputModeParams, ","))
+
+				for _, c := range invalid {
+					log.Warnf("column %q not found", c)
+				}
+
+				if err := formatter.SetShowColumns(valid); err != nil {
+					return err
+				}
 			}
 			parser.SetLogCallback(fe.Logf)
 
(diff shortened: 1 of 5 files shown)


cmd/common/registry.go
```diff
@@ -238,7 +238,9 @@ func buildCommandFromGadget(
 
 			formatter := parser.GetTextColumnsFormatter()
 			if outputModeParams != "" {
-				formatter.SetShowColumns(strings.Split(outputModeParams, ","))
+				if err := formatter.SetShowColumns(strings.Split(outputModeParams, ",")); err != nil {
```
flyth, line 241: Should we really error out in this case, though? Maybe a warning would suffice? If for example I'm switching between gadgets but defined a list of all columns I'm interested in, this would punish and force me to "clean up" that list. We have a VerifyColumns() that could be used to soft-fail (only forwarding the valid columns and reporting the invalid ones). This was btw also the intention behind the soft-fail inside SetShowColumns() in the first place.

However, I like your version (returning an error) more because it's cleaner - and we don't lose the soft-fail as said when combining with the verifier function.

mauriciovasquezbernal: Makes sense to me. I updated the PR.

flyth (approved):
Looks good to me, nice QoL improvement!

IMHO, I liked the `column %q not found` you had in an earlier commit more than the `column %q is invalid`.

PR #1355, opened 2023-02-22
RFC: pkg/columns: Use testify

mauriciovasquezbernal:
This library makes the tests more compact and avoids a lot of conditionals.

---

What do you all folks think? IMO we could start using this library in the tests as I think the result tests are more readable.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -192,6 +192,10 @@ if you need further help.
 If you want to propose a new feature or do a big change in the architecture
 it's highly recommended to open an issue first to discuss it with the team.
 
+### Writing tests
+
+We use `github.com/stretchr/testify` to make tests less verbose.
+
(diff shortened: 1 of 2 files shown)


flyth:
I agree, it looks much cleaner. I don't think it's worth rewriting all the tests, but if we have to update tests anyways or write new ones, I'll gladly write less but more readable code :)

blanquicet:
Yep, I like it. I checked the repo and I saw a lot of people is using it and even if there hasn't been new releases since Jan 2021, the repo seems to be active enough in case we need something. So, yeah, let's start using it.

mauriciovasquezbernal:
Great, I agree we should not rewrite all tests now, but when updating implementing new ones we should use this library. I added a sentence in the contributing guide to refer to it.

blanquicet (approved):
LGTM!

PR #1315, opened 2023-02-08
docs: Add releasing document

mauriciovasquezbernal:

PR #1356, opened 2023-02-23
docs: Add guide for 'How to write a Gadget'

flyth:
This PR adds a small and (hopefully) simple guide on how to write a new gadget using the updated architecture. Operators are not used for now, we can add that later in a more in-depth guide.

### DIFF
--- a/docs/devel/how-to-write-a-gadget.md
+++ b/docs/devel/how-to-write-a-gadget.md
@@ -1,0 +1,156 @@
+---
+title: How to write a Gadget?
+weight: 100
+description: >
+How to write a Gadget?
+---
+
+So you want to write a Gadget for Inspektor Gadget? Great!
+
+A gadget consists of a __Gadget Descriptor__ that provides metadata about your gadget and a __Gadget__ implementation
+(sometimes called tracer in our codebase).
+
+## Gadget Descriptor
+
+The first step to writing a new gadget is providing a _Gadget Descriptor_. Simply create a new struct that implements
+the [`gadgets.GadgetDesc`]( https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@main/pkg/gadgets#GadgetDesc)
+interface, providing all required information, including a name, description, category and so on.
+
+sIn our example, we call this struct like the interface, `GadgetDesc`. If you want your gadget to return records of
+a specific known type, for example system events as they come in, you can define a type for it (we call it
+EventDataType in our example) and return a pointer to an empty version of it in the `EventPrototype()` methods and
+initialize a Parser inside the `Parser()` function. This typically looks like this:
+
+```go
+func (g *GadgetDesc) EventPrototype() {
+	return &EventDataType{}
+}
+
+func (g *GadgetDesc) Parser() {
+	return parser.NewParser[EventDataType](nil)
+}
+```
+
+If you want support for a human-readable output in the form of text columns later on,
+you should annotate the fields of your data struct like explained in our
+[Columns library documentation](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget/pkg/columns).
+The annotation works similar to the JSON tags for struct members that you probably already know.
+In that case, you should instantiate a columns helper in the `Parser()` function like so:
+
+```go
+func (g *GadgetDesc) Parser() {
+	return parser.NewParser[EventDataType](columns.MustCreateColumns[EventDataType]())
+}
+```
+
+In addition to the implementation of the `gadgets.GadgetDesc` interface, you must also provide a
+`NewInstance() (gadgets.Gadget, error)` method on the same type, to satisfy the
+[`gadgets.GadgetInstantiate`](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@main/pkg/gadgets#GadgetInstantiate)
+interface. This method should return an instance of your gadget.
+
+> In order to avoid compilation errors on systems not supporting ebpf and kernel functions, you can move the
+> `NewInstance()` method and your actual gadget code to a separate file with the special build tag `!withoutebpf`
+> (the negation is important).
+
+To be able to use the gadget afterwards, we need to register the newly created GadgetDesc with the gadget-registry.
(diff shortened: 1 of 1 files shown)


docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (g *GadgetDesc) Parser() {
+	return parser.NewParser[EventDataType](nil)
+}
+```
+
+If you want support for a human-readable output in the form of text columns later on,
+you should annotate the fields of your data struct like explained in our
+[Columns library documentation](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.13.0/pkg/columns).
```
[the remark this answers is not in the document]
flyth: Not really, I will change it to latest :)

docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+In this case, we store the casted eventCallback in our Gadget struct, so we can use it later to send some data.
+
+## Gadget Lifecycle Overview
+
+This is a list of a default lifecycle of a gadget with all interfaces implemented. It also contains handling operators
+for the sake of completeness, but that was not subject of this article.
+
+```
```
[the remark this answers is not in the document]
flyth: It's not really go but rather pseudo-code. I'll try, but I guess this will make it look weird.

flyth:
> I think this guide is good, but I would like to try using it to convert an existing gadget to the new code you added in your refactoring.
> Do you have any suggestion there?

Thanks! All gadgets have already been converted and use what's described in this guide, so this is really rather for new gadgets and to understand the lifecycle of the gadgets and operators.

docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
...
+
+operators.PostGadgetRun()
+
+gadget.Close() (gadgets.CloseGadget interface)
+out, err := gadget.Result() (gadgets.GadgetResult interface)
+```
+
+## Conclusion
+
+This article showed you how you can implement a simple gadget fairly easy. We will go into more advanced topics
```
mauriciovasquezbernal, line 138:
```suggestion
This article showed you how you can implement a simple gadget. We will go into more advanced topics
```

docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
...
+
+So you want to write a Gadget for Inspektor Gadget? Great!
+
+A gadget consists of a __Gadget Descriptor__ that provides metadata about your gadget and a __Gadget__ implementation
+(sometimes called tracer in our codebase).
+
+## Gadget Descriptor
+
+The first step to writing a new gadget is providing a _Gadget Descriptor_. Simply create a new struct that implements
+the `gadgets.GadgetDesc` interface, providing all required information, including a name, description, category and so
```
mauriciovasquezbernal, line 11: Add some links to those interfaces -> https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@main/pkg/gadgets#GadgetDesc?

docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+The last thing to do, is to make sure that the file containing this registration is imported somehow. You can do this
+for example by adding it to `pkg/all-gadgets/gadgets.go`. (TBD)
+
+## Gadget Implementation
+
+Your actual code, that does the work, should live on the type you returned in the `NewInstance()` method of your
+`GadgetDesc` implementation. The only mandatory method is `Init(gadgetCtx gadgets.GadgetContext) error` to satisfy the
+`gadgets.Gadget` interface. This method will be called as soon as your gadget is executed and handed over a
```
mauriciovasquezbernal, line 63: https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@main/pkg/gadgets#Gadget

docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
+# How to write a Gadget?
```
mauriciovasquezbernal, line 1: Please remove the title and add a header like this generate this on our website. 

```
---
title: How to write a Gadget?
weight: 100
description: >
  How to write a Gadget?
---
```

mauriciovasquezbernal (commented):
In https://github.com/inspektor-gadget/inspektor-gadget/pull/1315 I created a devel folder, I think that's the best place for this file.

Besides that LGTM.

docs/guides/how-to-write-a-gadget.md
```diff
@@ -0,0 +1,141 @@
...
+you should annotate the fields of your data struct like explained in our
+[Columns library documentation](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.13.0/pkg/columns).
+The annotation works similar to the JSON tags for struct members that you probably already know.
+In that case, you should instantiate a columns helper in the `Parser()` function like so:
+
+```go
+func (g *GadgetDesc) Parser() {
+	return parser.NewParser[EventDataType](columns.MustCreateColumns[EventDataType]())
+}
+```
```
blanquicet, line 37: I was wondering if the gadget should have a method `GetColumns()` instead of `Parser()` and the caller (`buildCommandFromGadget` in our case) is in charge of creating the parser as it already knows the type from `EventPrototype()`. In this way, the gadget developer doesn't have to know about the `Parser` but just to return what `columns.MustCreateColumns[EventDataType]()` (`types.GetColumns()` in our case) returns if they want to support "human-readable" output or return `nil` otherwise. WDYT?

flyth: I'd really like that, but afaiu it's not as simple as that, because types that use generics have to be known at compile time and this would make it more runtime-ish. But maybe I fail to see how it can be done right now - will definitely look more into it.

mauriciovasquezbernal (approved):
LGTM!

PR #1359, opened 2023-02-23
pkg/operators: Add tests for SortOperators

burak-ok:
Tests the sorting of Operators

### DIFF
--- a/pkg/operators/operators_test.go
+++ b/pkg/operators/operators_test.go
@@ -1,0 +1,195 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


pkg/operators/operators_test.go
```diff
@@ -0,0 +1,195 @@
...
+		createOp("a", []string{}),
+	}
+
+	sortedOps, err := SortOperators(ops)
+	if assert.NoError(t, err) {
+		checkDependencies(t, ops, sortedOps)
+	}
+}
+
+func Test_SortOperatorsLargaGraph(t *testing.T) {
```
flyth, line 144: nit:
```suggestion
func Test_SortOperatorsLargeGraph(t *testing.T) {
```

pkg/operators/operators_test.go
```diff
@@ -0,0 +1,195 @@
+// Copyright 2022 The Inspektor Gadget authors
```
flyth, line 1: nit:
```suggestion
// Copyright 2022-2023 The Inspektor Gadget authors
```

Sorry that it took so long :)

flyth (approved):
Thanks! This is also a good base for other future tests concerning operators.

I like that you're also using the testify lib already that we just discussed in #1355.

pkg/operators/operators_test.go
burak-ok: Haha, typical copy paste error :D

PR #1361, opened 2023-02-23
top/ebpf: Print all information about processes

mauriciovasquezbernal:
Instead of only printing the information for the first process, let's print all the PIDs and comms that are using a given ebpf program. The columns library will take care of "cutting" the output if it's too long:

```
$ sudo ./local-gadget top ebpf
PROGID      TYPE        NAME        PID         COMM           RUNTIME RUNCO…  MAPMEMORY MAPCOUNT
75          Tracing     ig_top_ebp… 159953      local-gadge  784.377µs 8684         4KiB 1
3           CGroupSKB               1           systemd             0s 0              0B 0
68          Kprobe      kprobe_exe… 156948,159… kprobe,foo          0s 0            4KiB 1
```

### DIFF
--- a/pkg/gadgets/top/ebpf/types/types.go
+++ b/pkg/gadgets/top/ebpf/types/types.go
@@ -16,6 +16,7 @@ package types
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/docker/go-units"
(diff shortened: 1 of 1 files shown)


flyth (approved):
LGTM

PR #1362, opened 2023-02-23
top/ebpf: Rename Pids to Processes in Stats struct

mauriciovasquezbernal:
PidInfo was a struct with a Pid and a Comm. It seems a bit misleading. This commit renames PidInfo to Process to make it clear that this includes the PID and the command.

--- 

This is somehow related to https://github.com/inspektor-gadget/inspektor-gadget/pull/1361, but I prefer to have separated PRs to keep the discussion independent.

### DIFF
--- a/docs/crds/gadgets/ebpftop.md
+++ b/docs/crds/gadgets/ebpftop.md
@@ -8,7 +8,7 @@ ebpftop shows cpu time used by ebpf programs.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (node,namespace,pod,container,progid,type,name,pid,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount). (default -runtime,-runcount)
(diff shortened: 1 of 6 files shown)


docs/crds/gadgets/ebpftop.md
```diff
@@ -8,7 +8,7 @@ ebpftop shows cpu time used by ebpf programs.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (node,namespace,pod,container,progid,type,name,pid,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount). (default -runtime,-runcount)
+ - sort_by: The field to sort the results by (node,namespace,pod,container,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount). (default -runtime,-runcount)
```
mauriciovasquezbernal, line 11: It's not possible anymore to sort by pid, however I don't think it's a big deal.

flyth: We could add sorting for virtual columns, if we really need it, later on - but that won't help for the pid field in this case anyways.

flyth (approved):
LGTM, thanks for cleaning up!

PR #1363, opened 2023-02-24
ci: Explain why dependencies checks for PR should only be run on PR.

eiffel-fl:


.github/workflows/inspektor-gadget.yml
```diff
@@ -88,7 +88,9 @@ jobs:
   dependencies-checks:
     name: Dependencies checks
```
burak-ok, line 89: The comments make sense, thanks.
Maybe we could additionally rename the `name` and even the target to `pr-dependencies-check`
```suggestion
    name: PR dependencies checks
```

burak-ok (approved):
LGTM, thanks :)

Issue #1342, opened 2023-02-21
[EPIC] Add new gadget for `bpftrace`


mauriciovasquezbernal:
[bpftrace](https://github.com/iovisor/bpftrace) is an open-source project that implements a "High-level tracing language for Linux eBPF". Its main feature is to implement an LLVM backend that compiles scripts to eBPF bytecode, this allows users to write [one-liner](https://github.com/iovisor/bpftrace#one-liners) eBPF programs to extract information from a Linux host. 

bpftrace is designed to work on a single host and doesn't provide any integration with higher level concepts like containers or pods. Integrating bpftrace in Inspektor Gadget could enable new use cases, like aggregating/filtering data by container/pod/namespace and enriching the captured events with such information. [kubectl-trace](https://github.com/iovisor/kubectl-trace) allows users to "schedule bpftrace programs on kubernetes cluster using the kubectl", however it doesn't provide new filtering / enriching capabilities to it. 

### Implementation Phases

This seems to be a huge work, then we should split in different phases:

1. Implement a basic support that allows to run `bpftrace` scripts on a cluster. (very similar to current kubectl-gadget support)
```
# run on a single node 
$ kubectl gadget bpftrace --node foo -e `tracepoint...`

# run on multiple nodes 
# TODO: think about how to show results. 
$ kubectl gadget bpftrace -e `tracepoint...`
```

2. Integrate bpftrace and Inspektor Gadget to enable it to be used in Kubernetes. The following are some ideas of the support we have for now, it's still needed to study them more in detail.

```bash
# Filter events by container name 
$ kubectl gadget bpftrace –containername foo -e `tracepoint...` 
# or 
$ kubectl gadget bpftrace -e `tracepoint .. /containername=="foo"/ ...` 

# Print k8s related info  
$ kubectl gadget bpftrace -e 'tracepoint:syscalls:sys_enter_read { printf(“%s\n”, containername)}'
```

alban:
I started an experimental branch adding a new builtin function. I'd like to make it call a bpf extension (not done yet).
https://github.com/alban/bpftrace/commits/alban/external

alban:
I got a PoC working with the following branches:
- https://github.com/alban/freplace/commits/79b6b9f0508ec942a4eb3f5395cf16279d0b782a
- https://github.com/alban/bpftrace/commits/e323867778d3fcb7124aa919e065afb63056109e

How to use:

1. create bpf extension program
```
sudo ./freplace
[...]

Issue #250, opened 2021-09-03
[RFE] gadgets: Separate Event data structure by related fields 


blanquicet:
## Current situation

As discussed in  https://github.com/kinvolk/inspektor-gadget/pull/238#discussion_r700284527, the socket-collector gadget uses the Event data structure which is not really suitable because socket-collector is not an event-based gadget.

## Impact

It is not really an issue but could cause confusion.

## Ideal future situation

An initial proposal is:

Split `eventtypes.Event` into one named `Kubernetes` with `Node`, `Namespace`, `Pod` and `Container`, `Host` fields. Then, I would create another one named `Event` which includes `Kubernetes` and adds `Err`, `Ready` and `Notice` fields. This new data structure would replace the old `eventtypes.Event`.
 
Starting from that, the DNS gadget can create its own (Maybe `DnsEvent`) based on `Event` as it was before, while socket-collector would create one based on `Kubernetes` and adds its own fields, resulting in something that would finally deserve to be called `Socket`.

mauriciovasquezbernal:
https://github.com/kinvolk/inspektor-gadget/commit/bbb2524ce621ed6dbb7aaa3fe044f88e75f3bab6 introduced a `CommonData` structure as you described it here. It's still missing to update the socket collector to use it.

mauriciovasquezbernal:
I think this is not needed anymore as we're focusing on image-based gadgets now (#1929). @blanquicet WDYT?

blanquicet:
Agree. Closing based on previous comment.

PR #1370, opened 2023-02-24
gadgets: Initial bpftrace support

mauriciovasquezbernal:
This PR implements a basic support for bpftrace. 

```bash
$ kubectl gadget bpftrace -e 'tracepoint:syscalls:sys_enter_openat { printf("%s %s\n", comm, str(args->filename)); }'
NODE                             OUTPUT
minikube-m02                     Attaching 1 probe...
minikube-m02                     bpftrace /sys/devices/system/cpu/online
minikube-m02                     bpftrace /sys/devices/system/cpu/online
minikube-m03                     Attaching 1 probe...
minikube-m02                     bpftrace /dev/null
minikube-m02                     bpftrace /sys/kernel/debug/tracing/events/syscalls/sys_enter_openat/id
minikube-m02                     bpftrace /sys/devices/system/cpu/online
minikube-m03                     bpftrace /sys/devices/system/cpu/online
minikube-m03                     bpftrace /sys/devices/system/cpu/online
minikube-m02                     bpftrace /sys/devices/system/cpu/online
minikube                         Attaching 1 probe...
minikube-m02                     bpftrace /dev/null
minikube-m02                     bpftrace /sys/kernel/debug/tracing/events/syscalls/sys_enter_openat/id
minikube-m02                     runc /usr/bin/runc
minikube-m02                     runc /proc/sys/kernel/cap_last_cap
minikube                         runc /proc/sys/kernel/cap_last_cap
minikube-m03                     runc /proc/sys/kernel/cap_last_cap
minikube-m02                     runc
...
```

TODO: (for future PRs?)
- [ ] Support arbitrary parameters, i.e, make  kubectl-gadget transparent to bpftrace, just pass parameters around. 
- [ ] Support `-o json`
- [X] Documentation
- [X] Tests

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1432

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -841,6 +841,7 @@ jobs:
         kubernetes_architecture: "${{ matrix.arch }}"
         container_repo: ${{ steps.set-repo-determine-image-tag.outputs.container-repo }}
         image_tag: ${{ steps.set-repo-determine-image-tag.outputs.image-tag }}
+        image_flavour: "default"
     - name: Delete AKS cluster ${{ env.CLUSTER_NAME }}
       if: always()
       shell: bash
@@ -940,6 +941,7 @@ jobs:
         kubernetes_architecture: "amd64"
         container_repo: ${{ steps.set-repo-determine-image-tag.outputs.container-repo }}
         image_tag: ${{ steps.set-repo-determine-image-tag.outputs.image-tag }}
+        image_flavour: ${{ matrix.type }}
         dnstester_image: ${{ needs.build-helper-images.outputs.dnstester_image }}
     - if: ${{ matrix.type == 'default' && matrix.runtime == 'docker' }}
       name: Check that README is up-to-date
--- a/gadget-container/gadgettracermanager/main.go
+++ b/gadget-container/gadgettracermanager/main.go
@@ -38,6 +38,8 @@ import (
 
 	// This is a blank include that actually imports all gadgets
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/all-gadgets"
+	// The script gadget is designed only to work in k8s, hence it's not part of all-gadgets
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/script"
 
 	gadgetservice "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgettracermanager"
--- a/.github/actions/run-integration-tests/action.yml
+++ b/.github/actions/run-integration-tests/action.yml
@@ -8,6 +8,10 @@ inputs:
   image_tag:
     description: 'The image tag used as inspektor gadget deployment image tag.'
     required: true
+  image_flavour:
+    description: 'The image flavour used.'
+    required: true
+    default: "default"
   kubernetes_distribution:
     description: 'The kubernetes distribution used to select distro specific config in tests.'
     required: true
@@ -67,6 +71,7 @@ runs:
           CONTAINER_REPO=${{ inputs.container_repo }} \
           IMAGE_TAG=${{ inputs.image_tag }} \
           DNSTESTER_IMAGE=${{ inputs.dnstester_image }} \
+          IMAGE_FLAVOUR=${{ inputs.image_flavour }} \
           -o kubectl-gadget integration-tests |& tee integration.log & wait $!
         echo "IntegrationTestsJob: Done"
     - name: Prepare and publish test reports
--- a/Dockerfiles/gadget-default.Dockerfile
+++ b/Dockerfiles/gadget-default.Dockerfile
@@ -8,7 +8,11 @@ ARG BUILDER_IMAGE=golang:1.19
 # See BCC section in docs/devel/CONTRIBUTING.md for further details.
 ARG BCC="quay.io/kinvolk/bcc:gadget"
 
+# bpftrace upstream image
+ARG BPFTRACE="quay.io/iovisor/bpftrace:v0.17.0"
+
 FROM ${BCC} as bcc
+FROM ${BPFTRACE} as bpftrace
 FROM --platform=${BUILDPLATFORM} ${BUILDER_IMAGE} as builder
 
 ARG TARGETARCH
@@ -68,6 +72,13 @@ COPY gadget-container/hooks/nri/conf.json /opt/hooks/nri/
 
 # BTF files
 COPY hack/btfs /btfs/
+
+# bpftrace binary
+# TODO: this work because both bcc and bpftrace are based on the same ubuntu version:
+# https://github.com/kinvolk/bcc/blob/72b4247d7333df499a727987fde4e7903fc344d0/.github/workflows/publish.yml#L28
+# https://github.com/iovisor/bpftrace/blob/cd8c1b188df149277c09ebfb208de623a1aeaebb/docker/Dockerfile.focal#L1
+# We should consider building bpftrace ourselves to avoid this issue.
+COPY --from=bpftrace /usr/bin/bpftrace /usr/bin/bpftrace
 
 # Mitigate https://github.com/kubernetes/kubernetes/issues/106962.
 RUN rm -f /var/run
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ EBPF_BUILDER ?= ghcr.io/inspektor-gadget/inspektor-gadget-ebpf-builder
 
 DNSTESTER_IMAGE ?= "ghcr.io/inspektor-gadget/dnstester:latest"
 
+IMAGE_FLAVOUR ?= "default"
+
 # Adds a '-dirty' suffix to version string if there are uncommitted changes
 changes := $(shell git status --porcelain)
 ifeq ($(changes),)
@@ -193,6 +195,7 @@ integration-tests: kubectl-gadget
 			-k8s-arch $(KUBERNETES_ARCHITECTURE) \
 			-image $(CONTAINER_REPO):$(IMAGE_TAG) \
 			-dnstester-image $(DNSTESTER_IMAGE) \
+			-image-flavour $(IMAGE_FLAVOUR) \
 			$$INTEGRATION_TESTS_PARAMS
 
 .PHONY: generate-documentation
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Explore the following documentation to find out which tools can help you in your
 	- [`tcp`](docs/gadgets/trace/tcp.md)
 	- [`tcpconnect`](docs/gadgets/trace/tcpconnect.md)
 	- [`tcpdrop`](docs/gadgets/trace/tcpdrop.md)
+- [`script`](docs/gadgets/script.md)
 - [`traceloop`](docs/gadgets/traceloop.md)
 
 ## Installation
@@ -95,6 +96,7 @@ Available Commands:
   deploy         Deploy Inspektor Gadget on the cluster
   help           Help about any command
   profile        Profile different subsystems
+  script         Run a bpftrace-compatible scripts
   snapshot       Take a snapshot of a subsystem and print it
   top            Gather, sort and periodically report events according to a given criteria
   trace          Trace and print system events
--- a/cmd/kubectl-gadget/main.go
+++ b/cmd/kubectl-gadget/main.go
@@ -29,6 +29,8 @@ import (
 	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
 
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/all-gadgets"
+	// The script is not included in the all gadgets package.
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/script"
 )
 
 // common params for all gadgets
--- a/docs/gadgets/script.md
+++ b/docs/gadgets/script.md
@@ -1,0 +1,108 @@
+---
+title: 'Using script gadget'
+weight: 30
+description: >
+  Run bpftrace-compatible scripts using Inspektor Gadget.
+---
+
+### On kubernetes
+
+The script gadget allows to run [bpftrace](https://github.com/iovisor/bpftrace)-compatible scripts
+on a Kubernetes cluster. To get more information about bpftrace and its features please check the
+bpftrace documentation.
+
+Currently, only running one-liners is supported:
+
+```bash
+# Files opened by process
+$ kubectl gadget script -e 'tracepoint:syscalls:sys_enter_openat { printf("%s %s\n", comm, str(args->filename)); }'
+NODE                             OUTPUT
+minikube-m02                     Attaching 1 probe...
+minikube-m02                     bpftrace /sys/devices/system/cpu/online
+minikube-m02                     bpftrace /sys/devices/system/cpu/online
+minikube-m03                     Attaching 1 probe...
+minikube-m02                     bpftrace /dev/null
+minikube-m02                     bpftrace /sys/kernel/debug/tracing/events/syscalls/sys_enter_openat/id
+minikube-m02                     bpftrace /sys/devices/system/cpu/online
+minikube-m03                     bpftrace /sys/devices/system/cpu/online
+minikube-m03                     bpftrace /sys/devices/system/cpu/online
+minikube-m02                     bpftrace /sys/devices/system/cpu/online
+minikube                         Attaching 1 probe...
+minikube-m02                     bpftrace /dev/null
+minikube-m02                     bpftrace /sys/kernel/debug/tracing/events/syscalls/sys_enter_openat/id
+minikube-m02                     runc /usr/bin/runc
+minikube-m02                     runc /proc/sys/kernel/cap_last_cap
+minikube                         runc /proc/sys/kernel/cap_last_cap
+minikube-m03                     runc /proc/sys/kernel/cap_last_cap
+minikube-m02                     runc
+...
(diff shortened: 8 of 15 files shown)


alban:
Can't the gadget return two columns:
- node
- payload: generic string directly from bpftrace's output?

mauriciovasquezbernal:
> Can't the gadget return two columns:
> 
>     * node
> 
>     * payload: generic string directly from bpftrace's output?

I think it can work when bpftrace produces single line events, but what about cases like `-e 'tracepoint:syscalls:sys_exit_read /pid!=1/ { @[comm] = sum(args->ret); }'` that produce multiple lines? Perhaps we can use `ExtraLines()` for that? I'll think more about it.

flyth:
I really like your proposal, as it also solves the issue when we'd like to have different serialization (e.g. tcpdump binary stream doesn't need JSON). I had planned to do that within the parser, but this is much simpler (without the parser) and might in some cases suffice. I'd like to still think about the wiring, though - right now this feels a bit weird regarding the gadgetCtx.

For the specific use case, maybe we could start with this easy solution and improve upon later on once we extract more data that can be put into columns as well.

pkg/gadgets/bpftrace/tracer.go
```diff
@@ -0,0 +1,102 @@
...
+
+type Tracer struct {
+	eventCallback func([]byte)
+	cmd           *exec.Cmd
+	config        Config
+	l             logger.Logger
+}
+
+func (t *Tracer) Start() error {
+	t.cmd = exec.Command("bpftrace", "-e", t.config.Program)
```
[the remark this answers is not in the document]
alban: bpftrace is written in C++ and it uses libllvm, libbpf, libbcc. But it does not offer a library itself.

pkg/gadgets/bpftrace/gadget.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (g *GadgetDesc) EventPrototype() any {
+	return &event{}
+}
+
+func (g *GadgetDesc) OutputFormats() (gadgets.OutputFormats, string) {
+	return gadgets.OutputFormats{
+		"text": gadgets.OutputFormat{
+			Name:        "Text",
+			Description: "The output of the gadget is returned as raw text",
```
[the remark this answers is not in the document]
alban: Correct. For k8s metadata, we would need changes in bpftrace.

pkg/gadgets/bpftrace/tracer.go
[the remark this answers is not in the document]
mauriciovasquezbernal: This is something we discussed but seems too complex as bpftrace is not designed to have an API. Alban is exploring an implementation by using bpf extensions -> https://github.com/inspektor-gadget/inspektor-gadget/issues/1342#issuecomment-1455989538.

mauriciovasquezbernal:
I was able to implement something that works with kubetcl-gadget on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/1396, however I'll stop here and continue once that PR is merged.

mauriciovasquezbernal:
After talking privately to @alban, I decided that it was better to use the columns library from the beginning and allow running the script on multiple nodes. I updated the PR to reflect that. 

This is built on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/1396, even if I could build this independently, part of my work will be throw away as that PR reduces the quantity of code needed to implement a gadget on the client side.

mauriciovasquezbernal:
I'm marking this a ready to review. It's still based on https://github.com/inspektor-gadget/inspektor-gadget/pull/1396 but I think we can start reviewing it as we don't expect major changes in https://github.com/inspektor-gadget/inspektor-gadget/pull/1396.

pkg/gadgets/bpftrace/tracer.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return fmt.Errorf("getting stdout pipe: %w", err)
+	}
+
+	if err := cmd.Start(); err != nil {
+		return fmt.Errorf("running bpftrace: %w", err)
+	}
+
+	// TODO: how to enrich with the operator!
+	node := os.Getenv("NODE_NAME")
```
mauriciovasquezbernal, line 61: @flyth while implementing this gadget I found this situation, how to enrich only the node? (this gadget doesn't support filtering by k8s related things, yet)

I tried to implement this in `KubeManager` but I feel it doesn't fit very well: kubemanager provides parameters for filtering, logic for attaching containers, etc. using it to only enrich the node name seems too much. Another idea was to implement an operator that only enriches the node, but I'm not sure either, seems like it'll add overhead as it'll called for all events that are also handled by `KubeManager`. Do you have any opinion on this?

examples/kube-container-collection/main.go
```diff
@@ -105,7 +105,7 @@ func main() {
 		*kubeconfig = os.Getenv("KUBECONFIG")
 	}
 
-	if *node == "" && os.Getenv("NODE_NAME") != "" {
+	if *node == "" && os.("NODE_NAME") != "" {
```
alban, line 108: I don't understand the syntax `os.("NODE_NAME")`

pkg/gadgets/bpftrace/types/types.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (ev *Event) SetNode(node string) {
+	ev.Node = node
+}
+
+//func Base(ev eventtypes.Event) *Event {
+//	return &Event{
+//		Event: ev,
+//	}
+//}
```
alban, line 37: Can be removed

pkg/gadgets/bpftrace/tracer.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type Tracer struct {
+	eventCallback func(ev *types.Event)
+	config        Config
+}
+
+func (t *Tracer) Run(gadgetCtx gadgets.GadgetContext) error {
+	params := gadgetCtx.GadgetParams()
+	t.config.Program = params.Get(ParamProgram).AsString()
```
alban, line 41: Shouldn't it check that the string is not empty?

docs/gadgets/bpftrace.md
```diff
@@ -0,0 +1,99 @@
...
+minikube-m03                     @[EMT-7]: 15583
+minikube-m03                     @[dockerd]: 22937
+minikube-m03                     @[kubelet]: 27455
+...
+```
+
+### Future improvements
+
+We plan to make the bpftrace and Inspektor Gadget more powerful in the future. Please check TODO
+issue that contains more details about the planned features.
```
alban, line 99: I assume uprobes are not working. This could be added in the list of limitations.

alban (commented):
The documentation commit should also list the gadget in:
- README.md
- requirements.md

Thanks!

examples/kube-container-collection/main.go
mauriciovasquezbernal: Neither do I :D. Fixed!

gadget-container/gadgettracermanager/main.go
```diff
@@ -39,6 +39,8 @@ import (
 	gadgetservice "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgettracermanager"
 	pb "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgettracermanager/api"
+
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/bpftrace"
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Right now other gadgets are working because they are included in the gadget collection package. For this one we need a explicit include to make it work. I added a comment.

docs/gadgets/bpftrace.md
mauriciovasquezbernal: Actually they work -> `./kubectl-gadget bpftrace -e 'uretprobe:/host/bin/bash:readline { printf("read a line\n"); }'`  (note the /host prefix for the bin path). I added a comment on the documentation.

pkg/gadgets/bpftrace/tracer.go
mauriciovasquezbernal: I handled it using a validator in gadget.go. We're still discussing how to perform server-side validation of parameters.

pkg/gadgets/bpftrace/tracer.go
flyth: It can be done in the `grpc-runtime` directly, as an additional enricher. `runGadget` has the `node` inside GadgetPod, so you could check for the `SetNode()` interface on the prototype and hand over the enricher.

Something like that (maybe a new interface for only the NodeSetter might be needed, then):

```golang
	if parser != nil {
		var enrichers []func(any) error
		ev := gadgetCtx.GadgetDesc().EventPrototype()
		if _, ok := ev.(operators.ContainerInfoSetters); ok {
			enrichers = append(enrichers, func(ev any) error {
				ev.(operators.ContainerInfoSetters).SetNode(pod.node)
				return nil
			})
		}
		
		jsonHandler = parser.JSONHandlerFunc(enrichers...)
		jsonArrayHandler = parser.JSONHandlerFuncArray(pod.node, enrichers...)
	}
```

pkg/params/validators.go
```diff
@@ -123,3 +123,10 @@ func ValidateSlice(validator ParamValidator) func(value string) error {
 		return nil
 	}
 }
+
+func ValidateNonEmptyString(value string) error {
```
flyth, line 127: I think/hope the `IsMandatory` field of ParamDesc does the same (making sure there's a value).

pkg/gadgets/bpftrace/tracer.go
flyth: Server side validation is handled in `service.go` for the grpc way (`.CopyFromMap()`). So if not called manually (or through the current CR way), it's should have already been taken care of.

pkg/params/validators.go
mauriciovasquezbernal: :sweat_smile: you're right, that works!

pkg/gadgets/bpftrace/tracer.go
mauriciovasquezbernal: right, I opened https://github.com/inspektor-gadget/inspektor-gadget/pull/1486 to keep gadgets aligned.

pkg/gadgets/bpftrace/tracer.go
flyth: IMHO it should only be overwritten if not set already - if we implement fanning out requests, the info might be already there from the routing node . I'd prefer having it always set by the grpc-runtime, as it can potentially save _a bit_ of bandwidth and CPU time.


```mermaid
flowchart TD
  subgraph Flow
    direction TB
    client[Client]
    node1[Node 1]
    node2[Node 2]
    client <-- "via grpc-runtime\nsets to node 'node1' if not set" --- node1
    node1 <-- "via grpc-runtime\nsets to node 'node2' if not set" --- node2
  end
```

mauriciovasquezbernal: I see. Given that this discussion is not too related to this PR, do you agree if we leave it as it's and then we handle in a different one?

flyth: Absolutely!

pkg/gadgets/bpftrace/tracer.go
```diff
@@ -0,0 +1,111 @@
...
+	if flavour != "default" {
+		return fmt.Errorf("bpftrace is not supported on the %q flavour of the container image. Only default is supported for now",
+			flavour)
+	}
+
+	params := gadgetCtx.GadgetParams()
+	t.config.Program = params.Get(ParamProgram).AsString()
+	log := gadgetCtx.Logger()
+
+	cmd := exec.Command("bpftrace", "-e", t.config.Program)
```
blanquicet, line 50: What about using `CommandContext`, starting the go routines for the stdout and stderr, and then just calling `cmd.Run()`. Having used `CommandContext`, the `Run()` will automatically start the command, wait for its completion or kill the process when the context is done. WDYT?

blanquicet (commented):
I didn't test it but go through the code. It looks great. I'm happy to see how easy it is to add a new gadget now.

.github/workflows/inspektor-gadget.yml
```diff
@@ -841,6 +841,7 @@ jobs:
         kubernetes_architecture: "${{ matrix.arch }}"
         container_repo: ${{ steps.set-repo-determine-image-tag.outputs.container-repo }}
         image_tag: ${{ steps.set-repo-determine-image-tag.outputs.image-tag }}
+        image_flavour: "default"
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I prefer to be explicit and not rely on a the default value as we really want this to use the `default` image.

pkg/gadgets/bpftrace/gadget.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/bpftrace/types"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/parser"
+)
+
+const (
+	ParamProgram = "program"
+)
+
+type GadgetDesc struct{}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: The event for this gadget is defined as 

```
type Event struct {
	// Node where the event comes from
	Node   string `json:"node,omitempty" column:"node,template:node" columnTags:"kubernetes"`
	Output string `json:"output" column:"output,width:120"`
}
```

hence it doesn't support the KubeManager operator and not k8s related flags are added. 

```
$ ./kubectl-gadget bpftrace -h
Run a bpftrace program

Usage:
  kubectl-gadget bpftrace [flags]

Flags:
  -F, --filter strings   Filter rules
                           A filter can match any column using the following syntax
                             columnName:value       - matches, if the content of columnName equals exactly value
                             columnName:!value      - matches, if the content of columnName does not equal exactly value
                             columnName:>=value     - matches, if the content of columnName is greater or equal to the value
                             columnName:~value      - matches, if the content of columnName matches the regular expression 'value'
                         
  -h, --help             help for bpftrace
      --node string      Comma-separated list of nodes to run the gadget on (default "")
  -o, --output string    Output format (json, columns).
                         
                         Columns (columns)
                             Available columns:
                               node
                               output
                         
                         JSON (json)
                           The output of the gadget is returned as raw JSON
                         
                          (default "columns=node,output")
  -e, --program string   program to run (default "")
  -t, --timeout int      Number of seconds that the gadget will run for, 0 to disable
  -v, --verbose          Print debug information

```

pkg/gadgets/bpftrace/tracer.go
mauriciovasquezbernal: Before golang 1.20, the process is only killed when then context is done and it hasn't terminated, in this case we don't want to kill the process but interrupt it (otherwise it won't produce the output). In golang 1.20, it's possible to have a custom cancel function to interrupt the process. I added a note there to study that possibility once we supoprt go 1.20.

blanquicet (commented):
LGMT from code inspection. I'll test it later today or tomorrow.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -286,8 +287,17 @@ func (r *Runtime) runGadget(gadgetCtx runtime.GadgetContext, pod gadgetPod, allP
...
-		jsonHandler = parser.JSONHandlerFunc()
-		jsonArrayHandler = parser.JSONHandlerFuncArray(pod.node)
+		var enrichers []func(any) error
+		ev := gadgetCtx.GadgetDesc().EventPrototype()
+		if _, ok := ev.(operators.NodeSetter); ok {
+			enrichers = append(enrichers, func(ev any) error {
+				ev.(operators.NodeSetter).SetNode(pod.node)
+				return nil
+			})
+		}
```
blanquicet, line 297: This enrichment will only be needed by gadgets that doesn't use a parser?

pkg/gadgets/trace/tcpdrop/types/types.go
```diff
@@ -20,7 +20,20 @@ import (
...
+
+	// Timestamp in nanoseconds since January 1, 1970 UTC. An int64 is big
+	// enough to represent time between the year 1678 and 2262.
+	Timestamp eventtypes.Time `json:"timestamp,omitempty" column:"timestamp,template:timestamp,stringer"`
+
+	// Type indicates the kind of this event
+	Type eventtypes.EventType `json:"type"`
+
+	// Message when Type is ERR, WARN, DEBUG or INFO
+	Message string `json:"message,omitempty"`
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I tried different approaches to clean it up and it's a big mess with a lot of changes, so I decided to move forward with a very simple solution and handle this later on.  We have to think better these types, it's a problem we have from the very beginning -> https://github.com/inspektor-gadget/inspektor-gadget/issues/250

pkg/runtime/grpc/grpc-runtime.go
mauriciovasquezbernal: I think so, gadgets that don't use a parser only return `[]bytes` and we don't have any way to enrich it.

mauriciovasquezbernal:
> I tested it and it does the job, let's merge it, so we can polish it later.
> 
> Since we are into polishing error message a lot, the following could be modified (in a later PR):
> 
> ```shell
> $ ./kubectl-gadget script
> NODE                               OUTPUT                                                                                                                                  
> Error: running gadget: rpc error: code = Unknown desc = setting gadget parameters: expected value for "program"
> ```

Let me handle that in another PR.

PR #1371, opened 2023-02-24
cmd: Check if output mode is valid before running the gadget

mauriciovasquezbernal:
Print the error message as early as possible in case the output mode passed by the user is not valid.

Before:
```
$ sudo ./local-gadget profile block-io -o foobar
-> waiting some seconds before hitting ctrl + c
^CError: invalid output mode "foobar"
```

Now:
```
$ sudo ./local-gadget profile block-io -o foobar
Error: invalid output mode "foobar"
```

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -191,12 +191,7 @@ func buildCommandFromGadget(
 			}
 
 			if parser == nil {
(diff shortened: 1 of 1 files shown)


blanquicet (approved):
LGTM from code inspection

PR #1372, opened 2023-02-27
Dockerfiles: Use latest kinvolk/bcc as base image for gadget default one

eiffel-fl:
```
We recently updated kinvolk/bcc image to be built each day [1]. Hence, the latest tag will be up to date, so we can use it without troubles.

[1] https://github.com/iovisor/bcc/commit/a863186796aa439c1b8b3d1386477b1761619797
```


mauriciovasquezbernal (approved):
LGTM!

PR #1385, opened 2023-03-01
testutils: Print stderr logs for docker containers

mqasimsarfraz:
It seems we were only printing `stdout` in case of docker tests. This PR will allow printing `stderr` so we can get complete test logs.


mauriciovasquezbernal (approved):
LGTM. Thanks!

Issue #905, opened 2022-09-02
Consider renaming `NewTracer` function on gadget tracers


blanquicet:
## Current situation

Currently, [the packages](https://github.com/kinvolk/inspektor-gadget/tree/main/pkg/gadgets) we are exposing to be able to run the gadget tracers directly provides the following API:

- `func NewTracer(*Config, gadgets.DataEnricher, func(types.Event)) (*Tracer, error)`: Creates and run the tracer. Example in [trace/bind](https://github.com/kinvolk/inspektor-gadget/blob/256eb8fd9c3142151d84b020b98d5f43648dd6e1/pkg/gadgets/trace/bind/tracer/tracer.go#L84), [top/tcp](https://github.com/kinvolk/inspektor-gadget/blob/256eb8fd9c3142151d84b020b98d5f43648dd6e1/pkg/gadgets/top/tcp/tracer/tracer.go#L86).
- `func (t *Tracer) Stop()`: Stops the tracer.

## Impact

It creates some confusion as `NewTracer` is actually creating and running the trace, while its name only describes the creation. It creates some confusion for the caller. See [this discussion](https://github.com/kinvolk/inspektor-gadget/pull/896#discussion_r961562416).

## Ideal future situation

Naming should be more accurate with what the function is actually doing. 

## Additional information

Take into account that changing this naming will also imply changing the gadget examples, thus updating the blog-post that describes them.

mauriciovasquezbernal:
After https://github.com/inspektor-gadget/inspektor-gadget/pull/1281 these functions aren't used anymore, we're keeping them in order to avoid breaking users that use our gadgets directly from their golang code. Also, we're leaning towards supporting [containerized gadgets](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md) as the default way to run gadgets, so this code will be moved to legacy later on. 

I don't think it makes sense to expend time renaming those functions if they are going to be deprecated.

PR #1408, opened 2023-03-08
runtime/local: Move context and timeout management from runtime to tracers

blanquicet:
# runtime/local: Move context and timeout management from runtime to tracers

This PR tries to uniform and make more evident the lifecycle of gadgets by defining the following four gadget types:

1. Gadgets that collect insights from the system and, as soon as a new event is captured, they emit it through a callback. An example of this kind of gadget is the audit/seccomp gadget and the tracer category gadgets (except the dns, network and sni):

	<details>
	<summary><b>Code skeleton for Run() gadget with timeout</b></summary>

	```go
	func (t *Tracer) run(ctx context.Context) {
		for {
			record, err := t.reader.Read()
			if err != nil {
				if errors.Is(err, perf.ErrClosed) {
					// nothing to do, we're done
					return
				}

				// Report error
				return
			}
			
			[...]
			// Parse and emit events through the callback
			[...]
		}
	}

	func (t *Tracer) Run(gadgetCtx gadgets.GadgetContext) error {
		// Parse parameters
		[...]

		defer t.cleanup()
		if err := t.install(); err != nil {
			return nil, fmt.Errorf("installing tracer: %w", err)
		}

		go t.run()
		gadgetCtx.WaitForTimeoutOrDone()

		return nil
	}
	```

	</details>

	Notice that the gadget lifecycle is clearly defined in `Run()` :

	1. It parses the parameters if any.
	2. Install the eBPF program in the system.
	3. Runs a new go routine to communicate with the eBPF program and emit the events through the callback.
	4. Block the main go routine until the context is done/cancelled. For `ig`, users hit Ctrl+C (managed in front-end), or the timeout is reached.

2. A second type of gadget is the one the top category gadgets follow, which doesn't emit events as soon as they occur but emit them every X time (interval). So, they also use the `Run()` interface, which follows mostly the same flow as (1) but doesn't start the timer for the timeout to avoid dealing with two timers. Instead, it uses an internal counter and passes a `cancel` function to the `t.run()` so that it can notify when the counter (timeout) is reached:

	<details>
	<summary><b>Code skeleton for Run() gadget with ticker and counter</b></summary>

	```go
	func (t *Tracer) run(ctx context.Context) error {
		// Do not use a context with a timeout but use a counter
		count := t.config.Count
		ticker := time.NewTicker(t.config.Interval)
		defer ticker.Stop()

		for {
			select {
			case <-ctx.Done():
				return nil
			case <-ticker.C:
				[...]
				# Capture and emit events through the callback
				[...]

				if count > 0 {
					count--
					if count == 0 {
						return nil
					}
				}
			}
		}
	}

	func (t *Tracer) Run(gadgetCtx gadgets.GadgetContext) error {
		if err := t.init(gadgetCtx); err != nil {
			return nil, fmt.Errorf("initializing tracer: %w", err)
		}
		
		defer t.cleanup()
		if err := t.install(); err != nil {
			return nil, fmt.Errorf("installing tracer: %w", err)
		}

		return t.run(gadgetCtx.Context())
	}

	func (t *Tracer) init(gadgetCtx gadgets.GadgetContext) error {
		// Parse parameters
		[...]

		var err error
		if t.config.Count, err = top.ComputeCount(t.config.Interval, gadgetCtx.Timeout()); err != nil {
			return err
		}

		[...]
	}

	func ComputeCount(interval, timeout time.Duration) (int, error) {
		if timeout <= 0 {
			return 0, nil
		}
		if timeout < interval {
			return 0, fmt.Errorf("timeout must be greater than interval")
		}
		if timeout%interval != 0 {
			return 0, fmt.Errorf("timeout must be a multiple of interval")
		}
		return int(timeout / interval), nil
	}
	```

	</details>

3. The third set of gadgets doesn't emit events as soon as they occur but not through a callback. Instead, they return the result only when the gadget is stopped. An example of this kind of gadget is the advise/seccomp and profile/block-io gadgets:

	<details>
	<summary><b>Code skeleton for RunWithResult</b></summary>

	```go
	func (t *Tracer) collectResult() ([]byte, error) {
		[...]
		# Collect data from eBPF program
		[...]

		return json.Marshal(result)
	}

	func (t *Tracer) RunWithResult(gadgetCtx gadgets.GadgetContext) ([]byte, error) {
		// Parse parameters
		[...]
		
		defer t.cleanup()
		if err := t.install(); err != nil {
			return nil, fmt.Errorf("installing tracer: %w", err)
		}

		gadgetCtx.WaitForTimeoutOrDone()
		
		return t.collectResult()
	}
	```

	</details>

4. Additionally, The `RunGadget` and `RunWithResultGadget` gadgets can also implement the `InitCloseGadget`. It is useful for gadgets that uses the `AttachContainer`/`DetachContainer` mechanism to trace containers. They are different because they must be initialized and configured before the operators are installed. It is because the operators will start calling `AttachContainer`/`DetachContainer` during their installation.

	<details>
	<summary><b>Code skeleton for InitClose</b></summary>

	```go
	func (t *Tracer) Init(gadgetCtx gadgets.GadgetContext) error {
		// Parse parameters
		[...]

		if err := t.install(); err != nil {
			t.cleanup()
			return fmt.Errorf("installing tracer: %w", err)
		}
		
		// Context must be created before the first call to AttachContainer
		t.ctx, t.cancel = gadgetcontext.WithTimeout(gadgetCtx.Context(), gadgetCtx.Timeout())
		return nil
	}

	func (t *Tracer) Close() error {
		t.cancel()
		t.cleanup()
		return nil
	}

	func (t *Tracer) Run() error { // It can also be RunWithResult
		<-t.ctx.Done()
		return nil
	}

	func (t *Tracer) AttachContainer(container *containercollection.Container) error {
		[...]
		# Setup things for this new container that needs to be traced
		[...]
		
		go func() {
			for {
				if t.ctx.Err() != nil {
					return
				}
				
				[...]
				# Capture and emit events through the callback (or store it to send it later when the context is done in RunWithResult)
				[...]
		}()
	}

	func (t *Tracer) DetachContainer(container *containercollection.Container) error {
		[...]
		# Cleanup for this specific container
		[...]
	}
	```

	</details>

And finally, this is how the local runtime looks with these new interfaces:

<details>
<summary><b>Code skeleton for local.RunGadget</b></summary>

```go
func (r *Runtime) RunGadget(gadgetCtx runtime.GadgetContext) (out []byte, err error) {
	[...]

	// Create gadget instance
	gadgetInstance, err := gadget.NewInstance()
	if err != nil {
		return out, fmt.Errorf("instantiating gadget: %w", err)
	}

	// Initialize gadgets, if needed
	if initClose, ok := gadgetInstance.(gadgets.InitCloseGadget); ok {
		log.Debugf("calling gadget.Init()")
		err = initClose.Init(gadgetCtx)
		if err != nil {
			return out, fmt.Errorf("initializing gadget: %w", err)
		}
		defer func() {
			log.Debugf("calling gadget.Close()")
			initClose.Close()
		}()
	} else {
		log.Debugf("gadget does not implement gadget.Init() and gadget.Close()")
	}

	[...]
	# All the things we need to do with the operators
	[...]

	if run, ok := gadgetInstance.(gadgets.RunGadget); ok {
		log.Debugf("calling gadget.Run()")

		err := run.Run(gadgetCtx)
		if err != nil {
			return out, fmt.Errorf("running gadget: %w", err)
		}
	} else if runWithResult, ok := gadgetInstance.(gadgets.RunWithResultGadget); ok {
		log.Debugf("calling gadget.RunWithResult()")

		out, err = runWithResult.RunWithResult(gadgetCtx)
		if err != nil {
			return out, fmt.Errorf("running (with result) gadget: %w", err)
		}
	} else {
		return nil, errors.New("gadget not runnable")
	}

	return
}
```

</details>

TODO (It will be managed in another PR):

- [ ] Add this description to the developer [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/jose/ig-fix-intervarl-timeout/docs/devel/how-to-write-a-gadget.md).

### DIFF
--- a/integration/ig/k8s/top_block_io_test.go
+++ b/integration/ig/k8s/top_block_io_test.go
@@ -23,47 +23,82 @@ import (
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
+func newTopBlockIOCmd(ns string, cmd string, startAndStop bool) *Command {
+	expectedOutputFn := func(output string) error {
+		expectedEntry := &types.Stats{
+			CommonData: eventtypes.CommonData{
+				Namespace: ns,
+				Pod:       "test-pod",
+			},
+			Comm:  "dd",
+			Write: true,
+		}
+
+		normalize := func(e *types.Stats) {
+			e.Container = ""
+			e.Pid = 0
+			e.MountNsID = 0
+			e.Major = 0
+			e.Minor = 0
+			e.Bytes = 0
+			e.MicroSecs = 0
+			e.Operations = 0
+		}
+
+		return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
+	}
+
+	return &Command{
+		Name:             "TopBlockIO",
+		ExpectedOutputFn: expectedOutputFn,
+		Cmd:              cmd,
+		StartAndStop:     startAndStop,
+	}
+}
+
 func TestTopBlockIO(t *testing.T) {
 	t.Parallel()
 	ns := GenerateTestNamespaceName("test-top-block-io")
 
-	topBlockIOCmd := &Command{
-		Name:         "TopBlockIO",
-		Cmd:          fmt.Sprintf("ig top block-io -o json -m 999 --runtimes=%s", *containerRuntime),
-		StartAndStop: true,
-		ExpectedOutputFn: func(output string) error {
-			expectedEntry := &types.Stats{
-				CommonData: eventtypes.CommonData{
-					Namespace: ns,
-					Pod:       "test-pod",
-				},
-				Comm:  "dd",
-				Write: true,
-			}
-
-			normalize := func(e *types.Stats) {
-				e.Container = ""
-				e.Pid = 0
-				e.MountNsID = 0
-				e.Major = 0
-				e.Minor = 0
-				e.Bytes = 0
-				e.MicroSecs = 0
-				e.Operations = 0
-			}
-
-			return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
-		},
-	}
-
-	commands := []*Command{
+	commandsPreTest := []*Command{
 		CreateTestNamespaceCommand(ns),
-		topBlockIOCmd,
-		SleepForSecondsCommand(2), // wait to ensure ig has started
-		BusyboxPodRepeatCommand(ns, "dd if=/dev/zero of=/tmp/test count=4096"),
+		// Adding an additional sleep time to generate less events and avoid
+		// interference with other tests. See TestTopFile for more details.
+		BusyboxPodRepeatCommand(ns, "dd if=/dev/zero of=/tmp/test count=4096; sleep 0.2"),
 		WaitUntilTestPodReadyCommand(ns),
-		DeleteTestNamespaceCommand(ns),
 	}
+	RunTestSteps(commandsPreTest, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
 
-	RunTestSteps(commands, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	t.Cleanup(func() {
+		commandsPostTest := []*Command{
+			DeleteTestNamespaceCommand(ns),
+		}
+		RunTestSteps(commandsPostTest, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	t.Run("StartAndStop", func(t *testing.T) {
+		t.Parallel()
+
+		cmd := fmt.Sprintf("ig top block-io -o json -m 999 --runtimes=%s", *containerRuntime)
+		topBlockIOCmd := newTopBlockIOCmd(ns, cmd, true)
+		RunTestSteps([]*Command{topBlockIOCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	t.Run("Timeout", func(t *testing.T) {
+		t.Parallel()
+
+		cmd := fmt.Sprintf("ig top block-io -o json -m 999 --runtimes=%s --timeout %d",
+			*containerRuntime, timeout)
+		topBlockIOCmd := newTopBlockIOCmd(ns, cmd, false)
+		RunTestSteps([]*Command{topBlockIOCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	t.Run("Interval=Timeout", func(t *testing.T) {
+		t.Parallel()
+
+		cmd := fmt.Sprintf("ig top block-io -o json -m 999 --runtimes=%s --timeout %d --interval %d",
+			*containerRuntime, timeout, timeout)
+		topBlockIOCmd := newTopBlockIOCmd(ns, cmd, false)
+		RunTestSteps([]*Command{topBlockIOCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
 }
--- a/integration/ig/k8s/top_file_test.go
+++ b/integration/ig/k8s/top_file_test.go
@@ -1,4 +1,4 @@
-// Copyright 2022 The Inspektor Gadget authors
+// Copyright 2022-2023 The Inspektor Gadget authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,49 +23,88 @@ import (
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
+func newTopFileCmd(ns string, cmd string, startAndStop bool) *Command {
+	expectedOutputFn := func(output string) error {
+		expectedEntry := &types.Stats{
+			CommonData: eventtypes.CommonData{
+				Namespace: ns,
+				Pod:       "test-pod",
+			},
+			// echo is built-in
+			Comm:     "sh",
+			Filename: "bar",
+			FileType: 'R',
+		}
+
+		normalize := func(e *types.Stats) {
+			e.Container = ""
+			e.Pid = 0
+			e.Tid = 0
+			e.MountNsID = 0
+			e.Reads = 0
+			e.ReadBytes = 0
+			e.Writes = 0
+			e.WriteBytes = 0
+		}
+
+		return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
+	}
+
+	return &Command{
+		Name:             "TopFile",
+		ExpectedOutputFn: expectedOutputFn,
+		Cmd:              cmd,
+		StartAndStop:     startAndStop,
+	}
+}
+
 func TestTopFile(t *testing.T) {
 	t.Parallel()
 	ns := GenerateTestNamespaceName("test-top-file")
 
-	topFileCmd := &Command{
-		Name:         "TopFile",
-		Cmd:          fmt.Sprintf("ig top file -o json -m 999 --runtimes=%s", *containerRuntime),
-		StartAndStop: true,
-		ExpectedOutputFn: func(output string) error {
-			expectedEntry := &types.Stats{
-				CommonData: eventtypes.CommonData{
-					Namespace: ns,
-					Pod:       "test-pod",
-				},
-				// echo is built-in
-				Comm:     "sh",
-				Filename: "bar",
-				FileType: 'R',
-			}
-
-			normalize := func(e *types.Stats) {
-				e.Container = ""
-				e.Pid = 0
-				e.Tid = 0
-				e.MountNsID = 0
-				e.Reads = 0
-				e.ReadBytes = 0
-				e.Writes = 0
-				e.WriteBytes = 0
-			}
-
-			return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
-		},
-	}
-
-	commands := []*Command{
+	commandsPreTest := []*Command{
 		CreateTestNamespaceCommand(ns),
-		topFileCmd,
-		SleepForSecondsCommand(2), // wait to ensure ig has started
 		BusyboxPodRepeatCommand(ns, "echo foo > bar"),
 		WaitUntilTestPodReadyCommand(ns),
-		DeleteTestNamespaceCommand(ns),
 	}
+	RunTestSteps(commandsPreTest, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
 
-	RunTestSteps(commands, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	t.Cleanup(func() {
+		commandsPostTest := []*Command{
+			DeleteTestNamespaceCommand(ns),
+		}
+		RunTestSteps(commandsPostTest, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	// TODO: Filter by namespace to avoid interferences with events from other
+	// tests. In the meanwhile, given that we are generating events by writing
+	// into a file:
+	// 	(1) Increase max-rows to 999 (default is 20)
+	// 	(2) Sort by "-writes,-wbytes" (default is "-reads,-writes,-rbytes,-wbytes")
+
+	t.Run("StartAndStop", func(t *testing.T) {
+		t.Parallel()
+
+		cmd := fmt.Sprintf("ig top file -o json --sort-by -writes,-wbytes -m 999 --runtimes=%s", *containerRuntime)
+		topFileCmd := newTopFileCmd(ns, cmd, true)
+		RunTestSteps([]*Command{topFileCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	t.Run("Timeout", func(t *testing.T) {
+		t.Parallel()
+
+		cmd := fmt.Sprintf("ig top file -o json --sort-by -writes,-wbytes -m 999 --runtimes=%s --timeout %d",
+			*containerRuntime, timeout)
+		topFileCmd := newTopFileCmd(ns, cmd, false)
+		RunTestSteps([]*Command{topFileCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	t.Run("Interval=Timeout", func(t *testing.T) {
+		t.Parallel()
+
+		cmd := fmt.Sprintf("ig top file -o json --sort-by -writes,-wbytes -m 999 --runtimes=%s --timeout %d --interval %d",
+			*containerRuntime, timeout, timeout)
+		topFileCmd := newTopFileCmd(ns, cmd, false)
+		RunTestSteps([]*Command{topFileCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
 }
--- a/internal/benchmarks/benchmarks_test.go
+++ b/internal/benchmarks/benchmarks_test.go
@@ -271,6 +271,7 @@ func BenchmarkAllGadgetsWithContainers(b *testing.B) {
 							b.Fatalf("running gadget: %s", err)
 						}
 
+						gadgetCtx.Cancel()
 						cancel()
 					}
 				})
--- a/internal/benchmarks/benchmarks_test.go
+++ b/internal/benchmarks/benchmarks_test.go
@@ -263,6 +263,7 @@ func BenchmarkAllGadgetsWithContainers(b *testing.B) {
 							operatorsParamCollection,
 							parser,
 							logger.DefaultLogger(),
+							0,
 						)
 
 						_, err := runtime.RunGadget(gadgetCtx)
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -20,6 +20,7 @@ package gadgetcontext
 
 import (
 	"context"
+	"time"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
@@ -43,6 +44,7 @@ type GadgetContext struct {
 	logger                   logger.Logger
 	result                   []byte
 	resultError              error
+	timeout                  time.Duration
 }
 
 func New(
@@ -54,6 +56,7 @@ func New(
 	operatorsParamCollection params.Collection,
 	parser parser.Parser,
 	logger logger.Logger,
+	timeout time.Duration,
 ) *GadgetContext {
 	return &GadgetContext{
 		ctx:                      ctx,
@@ -65,6 +68,7 @@ func New(
 		logger:                   logger,
 		operators:                operators.GetOperatorsForGadget(gadget),
 		operatorsParamCollection: operatorsParamCollection,
+		timeout:                  timeout,
 	}
 }
 
@@ -102,4 +106,15 @@ func (r *GadgetContext) GadgetParams() *params.Params {
 
 func (r *GadgetContext) OperatorsParamCollection() params.Collection {
 	return r.operatorsParamCollection
+}
+
+func (r *GadgetContext) Timeout() time.Duration {
+	return r.timeout
+}
+
+func WithTimeoutOrCancel(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
+	if timeout == 0 {
+		return context.WithCancel(ctx)
+	}
+	return context.WithTimeout(ctx, timeout)
 }
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -72,44 +72,44 @@ func New(
 	}
 }
 
-func (r *GadgetContext) ID() string {
-	return r.id
+func (c *GadgetContext) ID() string {
+	return c.id
 }
 
-func (r *GadgetContext) Context() context.Context {
-	return r.ctx
+func (c *GadgetContext) Context() context.Context {
+	return c.ctx
 }
 
-func (r *GadgetContext) Parser() parser.Parser {
-	return r.parser
+func (c *GadgetContext) Parser() parser.Parser {
+	return c.parser
 }
 
-func (r *GadgetContext) Runtime() runtime.Runtime {
-	return r.runtime
+func (c *GadgetContext) Runtime() runtime.Runtime {
+	return c.runtime
 }
 
-func (r *GadgetContext) GadgetDesc() gadgets.GadgetDesc {
-	return r.gadget
+func (c *GadgetContext) GadgetDesc() gadgets.GadgetDesc {
+	return c.gadget
 }
 
-func (r *GadgetContext) Operators() operators.Operators {
-	return r.operators
+func (c *GadgetContext) Operators() operators.Operators {
+	return c.operators
 }
 
-func (r *GadgetContext) Logger() logger.Logger {
-	return r.logger
+func (c *GadgetContext) Logger() logger.Logger {
+	return c.logger
 }
 
-func (r *GadgetContext) GadgetParams() *params.Params {
-	return r.gadgetParams
+func (c *GadgetContext) GadgetParams() *params.Params {
+	return c.gadgetParams
 }
 
-func (r *GadgetContext) OperatorsParamCollection() params.Collection {
-	return r.operatorsParamCollection
+func (c *GadgetContext) OperatorsParamCollection() params.Collection {
+	return c.operatorsParamCollection
 }
 
-func (r *GadgetContext) Timeout() time.Duration {
-	return r.timeout
+func (c *GadgetContext) Timeout() time.Duration {
+	return c.timeout
 }
 
 func WithTimeoutOrCancel(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
@@ -117,4 +117,10 @@ func WithTimeoutOrCancel(ctx context.Context, timeout time.Duration) (context.Co
 		return context.WithCancel(ctx)
 	}
 	return context.WithTimeout(ctx, timeout)
+}
+
+func WaitForTimeoutOrDone(c gadgets.GadgetContext) {
+	ctx, cancel := WithTimeoutOrCancel(c.Context(), c.Timeout())
+	defer cancel()
+	<-ctx.Done()
 }
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -34,6 +34,7 @@ import (
 // instance and communicates with gadget and runtime.
 type GadgetContext struct {
 	ctx                      context.Context
+	cancel                   context.CancelFunc
 	id                       string
 	gadget                   gadgets.GadgetDesc
 	gadgetParams             *params.Params
@@ -58,8 +59,11 @@ func New(
 	logger logger.Logger,
 	timeout time.Duration,
 ) *GadgetContext {
+	gCtx, cancel := context.WithCancel(ctx)
+
 	return &GadgetContext{
-		ctx:                      ctx,
+		ctx:                      gCtx,
+		cancel:                   cancel,
 		id:                       id,
 		runtime:                  runtime,
 		gadget:                   gadget,
@@ -78,6 +82,10 @@ func (c *GadgetContext) ID() string {
 
 func (c *GadgetContext) Context() context.Context {
 	return c.ctx
+}
+
+func (c *GadgetContext) Cancel() {
+	c.cancel()
 }
 
 func (c *GadgetContext) Parser() parser.Parser {
--- a/pkg/gadgets/advise/seccomp/tracer/tracer.go
+++ b/pkg/gadgets/advise/seccomp/tracer/tracer.go
@@ -147,13 +147,10 @@ func (t *Tracer) RunWithResult(gadgetCtx gadgets.GadgetContext) ([]byte, error) 
 		return nil, fmt.Errorf("installing tracer: %w", err)
 	}
 
-	ctx, cancel := gadgetcontext.WithTimeoutOrCancel(gadgetCtx.Context(), gadgetCtx.Timeout())
-	defer cancel()
-
 	// Notice this Tracer starts collecting data for all containers as soon as
 	// it is installed, and uses the attach/detach mechanism to know which
 	// containers to report data from.
-	<-ctx.Done()
+	gadgetcontext.WaitForTimeoutOrDone(gadgetCtx)
 
 	return t.collectResult()
 }
--- a/pkg/gadgets/advise/seccomp/tracer/tracer.go
+++ b/pkg/gadgets/advise/seccomp/tracer/tracer.go
@@ -26,6 +26,7 @@ import (
 	libseccomp "github.com/seccomp/libseccomp-golang"
 
 	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 )
 
@@ -55,7 +56,7 @@ type Tracer struct {
 func NewTracer() (*Tracer, error) {
 	t := &Tracer{}
 
-	if err := t.start(); err != nil {
+	if err := t.install(); err != nil {
 		t.Close()
 		return nil, err
 	}
@@ -63,7 +64,7 @@ func NewTracer() (*Tracer, error) {
 	return t, nil
 }
 
-func (t *Tracer) start() error {
+func (t *Tracer) install() error {
 	spec, err := loadSeccomp()
 	if err != nil {
 		return fmt.Errorf("failed to load asset: %w", err)
@@ -124,6 +125,8 @@ func (t *Tracer) Delete(mntns uint64) {
 	t.objs.SyscallsPerMntns.Delete(mntns)
 }
 
+// Close closes the tracer
+// TODO: Unexport this function when the refactoring is done
 func (t *Tracer) Close() {
 	t.progLink = gadgets.CloseLink(t.progLink)
 	t.objs.Close()
@@ -138,19 +141,21 @@ func (g *GadgetDesc) NewInstance() (gadgets.Gadget, error) {
 	return t, nil
 }
 
-func (t *Tracer) Init(gadgetCtx gadgets.GadgetContext) error {
-	return nil
-}
-
-func (t *Tracer) Stop() {
-}
-
-func (t *Tracer) Start() error {
-	if err := t.start(); err != nil {
-		t.Stop()
-		return err
+func (t *Tracer) RunWithResult(gadgetCtx gadgets.GadgetContext) ([]byte, error) {
+	defer t.Close()
+	if err := t.install(); err != nil {
+		return nil, fmt.Errorf("installing tracer: %w", err)
 	}
-	return nil
+
+	ctx, cancel := gadgetcontext.WithTimeoutOrCancel(gadgetCtx.Context(), gadgetCtx.Timeout())
+	defer cancel()
+
+	// Notice this Tracer starts collecting data for all containers as soon as
+	// it is installed, and uses the attach/detach mechanism to know which
+	// containers to report data from.
+	<-ctx.Done()
+
+	return t.collectResult()
 }
 
 func (t *Tracer) AttachContainer(container *containercollection.Container) error {
@@ -168,7 +173,7 @@ func (t *Tracer) DetachContainer(container *containercollection.Container) error
 	return nil
 }
 
-func (t *Tracer) Result() ([]byte, error) {
+func (t *Tracer) collectResult() ([]byte, error) {
 	out := make(map[string][]string)
 	for container, result := range t.containers {
 		out[container.Name] = result
--- a/pkg/gadgets/interface.go
+++ b/pkg/gadgets/interface.go
@@ -66,10 +66,6 @@ type GadgetDesc interface {
 	EventPrototype() any
 }
 
-type GadgetResult interface {
-	Result() ([]byte, error)
-}
-
 type OutputFormats map[string]OutputFormat
 
 // OutputFormat can hold alternative output formats for a gadget. Whenever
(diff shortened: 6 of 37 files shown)


pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -54,6 +54,7 @@ type Tracer struct {
 	eventCallback    func(*top.Event[types.Stats])
```
blanquicet, line 54: Mmmm I was wondering what we should do if the interval is greater than the timeout. Should it return an error? With this PR, it will print what it has collected when the timeout is reached. WDYT?

integration/ig/k8s/top_block_io_test.go
```diff
@@ -23,47 +23,80 @@ import (
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
+func newTopBlockIOCmd(ns string, cmd string, startAndStop bool) *Command {
```
blanquicet, line 26: I think this is the function that we could share between `integration/ig` and `integration/inspektor-gadget`, maybe moving it to `integration/common` or so. I think it just needs to receive the `normalize` function because the rest is identical.

Notice I added two new tests to verify that (1) timeout and (2) the case when `timeout == interval` are working correctly. They should not impact the time CI is taking as both run in parallel with the one running previously.

mauriciovasquezbernal:
> If users use an interval that is not multiple of the timeout, we will print a last stats update. IMO, it is not necessarily a disadvantage:

This could be misleading as all printed stats don't share the same interval. All stats printed should share the same interval so the user can compare them. 

Regarding this solution, it seems racy as nothing guarantees that the call to `pushStats()` will end before the timeout timer ends, it can work in most of the cases but we still have that race there. 

I'm not sure how to fix it. I'm wondering if changing the update approach from push (the gadget pushes the updates) to pull (the user pulls the updates) will fix this, as the user (ig) will handle and could synchronize both timers.

pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -334,23 +349,9 @@ func (g *GadgetDesc) NewInstance() (gadgets.Gadget, error) {
...
-	}
-	t.colMap = statCols.GetColumnMap()
-
-	if err := t.start(); err != nil {
-		t.Stop()
-		return err
-	}
-
-	return nil
-}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Not removed, just moved above.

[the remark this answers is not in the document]
blanquicet: I was just trying to align all the tracers to make them behave similarly. So, having the functions organized in the same order just helped me to compare the files and easily spot the differences.

pkg/gadgets/top/block-io/tracer/tracer.go
blanquicet: Agree. Applied!

blanquicet:
>> If users use an interval that is not multiple of the timeout, we will print a last stats update. IMO, it is not necessarily a disadvantage:
>>
> This could be misleading as all printed stats don't share the same interval. All stats printed should share the same interval so the user can compare them.
>
This point can be solved by forcing users to enter an interval multiple of the timeout. WDYT?

> Regarding this solution, it seems racy as nothing guarantees that the call to pushStats() will end before the timeout timer ends, it can work in most of the cases but we still have that race there.
>
Yeah, that's actually the only real disadvantage of this approach. I just thought it works in more cases than the current implementation. I'm not saying it is a final solution 🙂.

> I'm not sure how to fix it. I'm wondering if changing the update approach from push (the gadget pushes the updates) to pull (the user pulls the updates) will fix this, as the user (ig) will handle and could synchronize both timers.
>
Let me give it a try. I think I had an idea.

blanquicet:
> I am not highly sure about but if this urgent I could give an approval.
> 
> I would also like to keep everything synchronized. Would it be possible to add a new trace operation to ask for the stats? So, periodically, the CLI asks the `gadgettracermanager` to send it the stats and when data are sent, the buffer is emptied?
>
Yeah, that's similar to the approach also Mauro is proposing: passing from push to pull approach. Let me try 🙂 .

flyth:
> > I am not highly sure about but if this urgent I could give an approval.
> > I would also like to keep everything synchronized. Would it be possible to add a new trace operation to ask for the stats? So, periodically, the CLI asks the `gadgettracermanager` to send it the stats and when data are sent, the buffer is emptied?
> 
> Yeah, that's similar to the approach also Mauro is proposing: passing from push to pull approach. Let me try 🙂 .

IMHO, we should not pull from gadgets, as that can skew the data (thinking about network hiccups for example). I think the timing should be handled by what is most precise, and that would be the gadget (as it knows about its initialization and so on). I have similar worries about this "almost 6 seconds".

I tried out another approach, by moving the timeout into the gadget context [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/7c6307a6885c6c4009b6bdcfa94af73e9d8558c5). This had the downside, that the `context.Context` inside the `GadgetContext` wasn't aware of the timeout anymore - but that could be resolved by changing the method `Context() context.Context` of the `GadgetContext` interface to `Context() (context.Context, context.CancelFunc)` and returning a new context based on the original one via `WithTimeout()`, if there's a timeout set in the `GadgetContext`. If we would then change the top gadgets from using `Start()` and `Stop()` to `Run()`, which then checks for `ctx.Done()` in addition to the interval timer in a loop, we could be fine (the context wouldn't be cancelled from the runtime anymore, as Run() is blocking). That basically would hand the timing completely over to the gadget (and also make it aware of the timeout, so you could still to extra steps if `timeout == interval`). WDYT?

mauriciovasquezbernal:
> IMHO, we should not pull from gadgets, as that can skew the data (thinking about network hiccups for example). I think the timing should be handled by what is most precise, and that would be the gadget (as it knows about its initialization and so on). I have similar worries about this "almost 6 seconds".

I think with both approaches we'll skew the data in case of a "network hiccup", for instance, it could happen that the update from a single node is delayed when the gadget pushes the data and the client will combine data from different nodes generated at different times. I don't see how a push approach is better on this matter than a pull approach. 

I suggested that the pull approach could help to fix it, but it we find a different alternative that would be great, changing for push to pull seems a long discussion and perhaps we can avoid it.

> I tried out another approach, by moving the timeout into the gadget context [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/7c6307a6885c6c4009b6bdcfa94af73e9d8558c5). This had the downside, that the `context.Context` inside the `GadgetContext` wasn't aware of the timeout anymore - but that could be resolved by changing the method `Context() context.Context` of the `GadgetContext` interface to `Context() (context.Context, context.CancelFunc)` and returning a new context based on the original one via `WithTimeout()`, if there's a timeout set in the `GadgetContext`. 

> change the top gadgets from using `Start()` and `Stop()` to `Run()`, which then checks for `ctx.Done()` 

This is interesting. Perhaps we can do it for **all** gadgets and get rid of `Start()` and `Stop()`. I mean, all gadgets will implement `Run()` that blocks and we'll use the context to indicate the gadget when to stop, because of a timeout or because the user hits Ctrl + C.

 WDYT?

flyth:
> > IMHO, we should not pull from gadgets, as that can skew the data (thinking about network hiccups for example). I think the timing should be handled by what is most precise, and that would be the gadget (as it knows about its initialization and so on). I have similar worries about this "almost 6 seconds".
> 
> I think with both approaches we'll skew the data in case of a "network hiccup", for instance, it could happen that the update from a single node is delayed when the gadget pushes the data and the client will combine data from different nodes generated at different times. I don't see how a push approach is better on this matter than a pull approach.

Agree, it's still not perfect. But the difference is that you don't skew the data directly but rather the presentation of it in some cases (especially if you're looking at the columns). However, if you export to JSON/CSV, the data would still be correct, whereas with the polling that would not necessarily be the case.

> 
> I suggested that the pull approach could help to fix it, but it we find a different alternative that would be great, changing for push to pull seems a long discussion and perhaps we can avoid it.
> 
> > I tried out another approach, by moving the timeout into the gadget context [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/7c6307a6885c6c4009b6bdcfa94af73e9d8558c5). This had the downside, that the `context.Context` inside the `GadgetContext` wasn't aware of the timeout anymore - but that could be resolved by changing the method `Context() context.Context` of the `GadgetContext` interface to `Context() (context.Context, context.CancelFunc)` and returning a new context based on the original one via `WithTimeout()`, if there's a timeout set in the `GadgetContext`.
> 
> > change the top gadgets from using `Start()` and `Stop()` to `Run()`, which then checks for `ctx.Done()`
> 
> This is interesting. Perhaps we can do it for **all** gadgets and get rid of `Start()` and `Stop()`. I mean, all gadgets will implement `Run()` that blocks and we'll use the context to indicate the gadget when to stop, because of a timeout or because the user hits Ctrl + C.
> 
> WDYT?

Yeah, totally agree, it would make things much cleaner.

blanquicet:
> > This is interesting. Perhaps we can do it for **all** gadgets and get rid of `Start()` and `Stop()`. I mean, all gadgets will implement `Run()` that blocks and we'll use the context to indicate the gadget when to stop, because of a timeout or because the user hits Ctrl + C.
> > WDYT?
> 
> Yeah, totally agree, it would make things much cleaner.

I updated the PR with it. PTAL.

cc @mauriciovasquezbernal @flyth

pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -273,32 +281,65 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
...
+		return fmt.Errorf("installing tracer: %w", err)
+	}
+
+	// Don't use a context with a timeout but a counter to avoid having to deal
+	// with two timers: one for the timeout and another for the ticker.
+	ctx, cancel := context.WithCancel(gadgetCtx.Context())
+	defer cancel()
+
+	go t.run(ctx, cancel)
+	<-ctx.Done()
```
mauriciovasquezbernal, line 340: What about something like https://github.com/inspektor-gadget/inspektor-gadget/commit/fd33b0a9b955efbb933595b1491eb4d4a4b7bdcf to avoid creatin gan extra goroutine here that is not needed?

pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -273,32 +281,65 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
...
+			if count > 0 {
+				count--
+				if count == 0 {
+					if cancel != nil {
+						cancel()
+					}
+					return
 				}
-				t.eventCallback(&top.Event[types.Stats]{Stats: stats[:n]})
 			}
```
mauriciovasquezbernal, line 319: I personally find this code a bit difficult to follow (too many conditionals involved). What about using the for loop structure to control it? 

option 1:
```
for i := 0; i < count || count == 0; i++ {
```

option 2: 
```
	if count == 0 {
		count = math.MaxInt
	}

	for i := 0; i < count; i++ {
```

mauriciovasquezbernal (commented):
I like it. IMO it goes in the right direction. Some comments around.

pkg/gadget-context/gadget-context.go
```diff
@@ -103,3 +107,14 @@ func (r *GadgetContext) GadgetParams() *params.Params {
 func (r *GadgetContext) OperatorsParamCollection() params.Collection {
 	return r.operatorsParamCollection
 }
+
+func (r *GadgetContext) Timeout() time.Duration {
+	return r.timeout
+}
+
+func WithTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Btw, I don't think we need this: passing 0 to `context.WithTimeout` should work the same as the golang handles it for us:

```golang
func WithTimeout(parent Context, timeout time.Duration) (Context, CancelFunc) {
	return WithDeadline(parent, time.Now().Add(timeout))
}
....
func WithDeadline(parent Context, d time.Time) (Context, CancelFunc) {
	return WithDeadlineCause(parent, d, nil)
}

// WithDeadlineCause behaves like WithDeadline but also sets the cause of the
// returned Context when the deadline is exceeded. The returned CancelFunc does
// not set the cause.
func WithDeadlineCause(parent Context, d time.Time, cause error) (Context, CancelFunc) {
	if parent == nil {
		panic("cannot create context from nil parent")
	}
	if cur, ok := parent.Deadline(); ok && cur.Before(d) {
		// The current deadline is already sooner than the new one.
		return WithCancel(parent)
	}
```

From https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/context/context.go;l=499;drc=82bf12902f2abecdbe2cb5fa38299ac70cf9c67e

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -274,7 +274,17 @@ func (t *Tracer[Event]) Detach(pid uint32) error {
...
+// Close closes the tracer
+// TODO: Remove after refactoring
 func (t *Tracer[Event]) Close() {
+	t.Cleanup()
+}
+
+// ---
+
+// Cleanup cleans up the tracer
+// TODO: It doesn't need to be exported after refactoring
```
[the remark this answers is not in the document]
blanquicet: I'm sorry. The comment was actually out of date.

pkg/gadgets/trace/exec/tracer/tracer.go
```diff
@@ -183,11 +190,22 @@ func (t *Tracer) run() {
...
+
+	defer t.cleanup()
+	if err := t.install(); err != nil {
+		return fmt.Errorf("installing tracer: %w", err)
+	}
+
+	ctx, cancel := gadgetcontext.WithTimeout(gadgetCtx.Context(), gadgetCtx.Timeout())
+	defer cancel()
+
+	go t.run()
```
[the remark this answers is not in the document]
blanquicet: Discussion in https://github.com/inspektor-gadget/inspektor-gadget/pull/1408#discussion_r1136158925.

pkg/gadgets/advise/seccomp/tracer/tracer.go
```diff
@@ -168,10 +179,14 @@ func (t *Tracer) DetachContainer(container *containercollection.Container) error
...
 	out := make(map[string][]string)
 	for container, result := range t.containers {
 		out[container.Name] = result
 	}
 	return json.MarshalIndent(out, "", "  ")
 }
+
+func (t *Tracer) init(gadgetCtx gadgets.GadgetContext) error {
+	return nil
+}
```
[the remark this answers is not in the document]
blanquicet: Actually, I forgot to remove all the `init`s that are empty (like this one) 😅 . I will do it.

I introduced it to make the parameters parsing and keep the `Run*` clean. In particular, because the one for the top gadget is quite long:

https://github.com/inspektor-gadget/inspektor-gadget/blob/0817ae862c7126af1428dc027d1dc1da0675b67a/pkg/gadgets/top/block-io/tracer/tracer.go#L376-L394

Maybe we could call it `parseParams()` and implement it only when the gadget has parameters. WDYT?

pkg/gadgets/top/block-io/tracer/tracer.go
blanquicet: Yep, I like it!

cc @flyth

flyth:
Here's the summary of an idea/suggestion that came up when brainstorming yesterday:

### Objective

Be able to

* have the timeout handled as part of Run() - that basically leaves it up to the gadget to decide when initialization is done and the timer should start
* (still) be able to cancel all contexts from inside the gadget, so that all created go-routines are stopped as well (even when done so before creating the `context.WithTimeout()` inside `Run()`) - for example inside operators like `AttachContainer()`/`DetachContainer()` that can only access the "main" ctx that was there _before_ `Run()` is called.

### Implementation

* inside `gadgetcontext.New()`, take the given `ctx` and create a new `context.WithCancel()` from it. Store the cancel func and the new ctx inside the `GadgetContext`.
* `Context()` returns this new context
* add new method `Cancel()` that calls the `cancel()` func created above
* inside `runtime.RunGadget()`, call `gadgetCtx.Cancel()` directly after `gadget.Run()` to clean up the context
* optionally: call cancel() inside the finalizer of GadgetContext to have third-parties not necessarily worry about calling `Cancel()` manually

Then, inside `Run()` of the gadget, you can simply create a new context with the timeout and wait for the timeout or cancellation to happen. After returning, the runtime will make sure that all remaining contexts are cancelled, but you could optionally do it inside the gadget as well, if more control over the lifecycle is necessary.

### Why is the extra call to `Cancel()` needed?

It will make sure that all descendant contexts of the main gadget context will be cancelled before de-initialization of  operators or `Close()` is called on the gadget. If you need to be 100% sure that everything is also *done* before closing operators (`PostGadgetRun()`) and the gadget using `Close()`, you could use a `sync.WaitGroup` and register/unregister go routines with it. Same applies to operators that could synchronize using a WaitGroup by waiting for it in `PostGadgetRun()`.

### Bonus

We could create a helper method `gadgetCtx.WaitForTimeoutOrDone()`, that will handle creating a new context with timeout, waiting for it to be done and clean it up afterwards. This could also alternatively be called automatically by the runtime, if the gadget has no `Run()` method.

### Advantages

* Timing is more precise by letting the gadget decide when it's done with initialization and the timer should start
* Gagdets can cancel themselves

### Disadvantages

* Gadgets *MUST* take care of the timer in `Run()` (gadget developers have to be made aware of this)
* Additional `Cancel()` call

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -288,11 +295,24 @@ func (t *Tracer) run() {
...
+	defer t.cleanup()
+	if err := t.install(); err != nil {
+		return fmt.Errorf("installing tracer: %w", err)
+	}
+
+	ctx, cancel := gadgetcontext.WithTimeout(gadgetCtx.Context(), gadgetCtx.Timeout())
+	defer cancel()
+
+	// TODO: Should we pass the cancel function to the run() function so it can
+	//       cancel the context when there is an error?
```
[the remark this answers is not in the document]
blanquicet: Yeah, we can introduce this improvement later on.

pkg/gadget-context/gadget-context.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Actually I was **totally** wrong. The code I linked there is related to the parent and doesn't implement the logic I wanted. Passing 0 as timeout to `WithTimeout` will cause the context to fire immediately.

pkg/gadgets/advise/seccomp/tracer/tracer.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Perhaps we could only have this for top gadgets. IMO it doesn't make sense in other gadgets as it's always few lines of code, so I feel it's cleaner to just put them directly on `Run*`. However this is just a personal preference, if you think it's better to have an independent function I'm fine with it.

pkg/gadgets/snapshot/process/tracer/tracer.go
```diff
@@ -289,10 +289,15 @@ func (t *Tracer) SetMountNsMap(mntnsMap *ebpf.Map) {
 	t.config.MountnsMap = mntnsMap
 }
 
-func (t *Tracer) Run() error {
+func (t *Tracer) Run(gadgetCtx gadgets.GadgetContext) error {
+	// TODO: Use context to cancel the collector
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'd say it depends on each specific gadget implementation. If there is a gadget that could potentially become stuck, the developer should guarantee it terminates somehow (when the external context is done or perhaps using a differnt internal timer). 

However it seems a very corner case to me and IIUC it doesn't happen with the current implementations.

pkg/gadget-context/gadget-context.go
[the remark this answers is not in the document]
blanquicet: I'll just rename it `WithTimeoutOrCancel` then.

pkg/gadgets/top/block-io/tracer/tracer.go
blanquicet: I don't know. They both seem also hacky. 

> I personally find this code a bit difficult to follow (too many conditionals involved). 
>
Even now that it is just?

```go
			if count > 0 {
				count--
				if count == 0 {
					return nil
				}
			}
```

mauriciovasquezbernal: It's a bit better. Anyway, it's also a personal preference, if you think using those conditionals is better I'm totally fine with this.

pkg/gadgets/top/block-io/tracer/tracer.go
blanquicet: Actually, we can avoid creating a new context with the cancel:

```go
	func (t *Tracer) run(ctx context.Context) error {
		// Do not use a context with a timeout but use a counter
		count := t.config.Count
		ticker := time.NewTicker(t.config.Interval)
		defer ticker.Stop()

		for {
			select {
			case <-ctx.Done():
				return nil
			case <-ticker.C:
				[...]
				# Capture and emit events through the callback
				[...]

				if count > 0 {
					count--
					if count == 0 {
						return nil
					}
				}
			}
		}
	}

	func (t *Tracer) Run(gadgetCtx gadgets.GadgetContext) error {
		if err := t.init(gadgetCtx); err != nil {
			return nil, fmt.Errorf("initializing tracer: %w", err)
		}
		
		defer t.cleanup()
		if err := t.install(); err != nil {
			return nil, fmt.Errorf("installing tracer: %w", err)
		}

		return t.run(gadgetCtx.Context())
	}
```

I updated the PR with this.

blanquicet:
Thanks for sharing this detailed summary @flyth!

I like the idea of having a mechanism to cancel the main gadget-context and ensure all descendant contexts will be cancelled. However, I think it should be only a "protection mechanism" if the developer forgets to cancel a descendant context (which should be fixed asap once detected). Still, IMO, it should not be used as a shortcut to "just cancel everything" or to allow gadgets to cancel themselves (which, IMO, would be a like bypass the whole architecture we have created). I consider that whoever creates a new descendant is responsible for cancelling it and shouldn't use this main gadget-context to do it. WDYT?

### Bonus
> We could create a helper method gadgetCtx.WaitForTimeoutOrDone(), that will handle creating a new context with timeout, waiting for it to be done and clean it up afterwards. 
>
Great idea. I implemented this!

> This could also alternatively be called automatically by the runtime, if the gadget has no Run() method.
>
I prefer to avoid that. IMO, we should force gadgets to implement one of our approaches: `Run`, `RunWithResult` or `InitRunClose`.

blanquicet:
TODO before merging:

- [ ] Update [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/jose/ig-fix-intervarl-timeout/docs/devel/how-to-write-a-gadget.md).

pkg/gadget-context/gadget-context.go
```diff
@@ -65,41 +72,63 @@ func New(
...
-	return r.operatorsParamCollection
+func (c *GadgetContext) Timeout() time.Duration {
+	return c.timeout
+}
+
+func (c *GadgetContext) WaitForTimeoutOrDone() {
+	ctx, cancel := WithTimeoutOrCancel(c.ctx, c.timeout)
+	defer cancel()
+	<-ctx.Done()
+}
```
mauriciovasquezbernal, line 127: I have mixed feelings about this function. I know it's useful in many cases, however it seems that it's too specific to be in   `GadgetContext`. I was wondering if this could be a helper function like, so we don't have to add this method to interfaces around. 

```golang
func WaitForTimeoutOrDone(c gadgets.GadgetContext) {
	ctx, cancel := WithTimeoutOrCancel(c.Context(), c.Timeout())
	defer cancel()
	<-ctx.Done()
}
```

and gadgets will use it like:

```golang 
	gadgetcontext.WaitForTimeoutOrDone(gadgetCtx)
```

but I don't know if that's a good idea. 

Also, this function won't work in cases like https://github.com/inspektor-gadget/inspektor-gadget/pull/1408#discussion_r1136158925

internal/benchmarks/benchmarks_test.go
```diff
@@ -263,13 +263,15 @@ func BenchmarkAllGadgetsWithContainers(b *testing.B) {
...
 							logger.DefaultLogger(),
+							0,
 						)
 
 						_, err := runtime.RunGadget(gadgetCtx)
 						if err != nil {
 							b.Fatalf("running gadget: %s", err)
 						}
 
+						gadgetCtx.Cancel()
```
mauriciovasquezbernal, line 274: This seems weird to cancel two contexts there. Can't we rely on the parent being cancelled in the line below?

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -274,7 +274,15 @@ func (t *Tracer[Event]) Detach(pid uint32) error {
...
 
+// Close closes the tracer
+// TODO: Remove after refactoring
 func (t *Tracer[Event]) Close() {
+	t.Cleanup()
+}
+
+// ---
+
+func (t *Tracer[Event]) Cleanup() {
```
mauriciovasquezbernal, line 285: Why this change? What's wrong with having a `Close()` method? I think Close() is very well-known in go as the method that performs the cleanup for a resource, so probably we want to keep using that name.

pkg/gadgets/advise/seccomp/tracer/tracer.go
```diff
@@ -125,6 +125,10 @@ func (t *Tracer) Delete(mntns uint64) {
 }
 
 func (t *Tracer) Close() {
+	t.cleanup()
+}
+
+func (t *Tracer) cleanup() {
```
mauriciovasquezbernal, line 131: I'm not totally convinced about calling it `cleanup()`, if you look at the implementations around, all of them call `*Close()` methods, so why not to call it `close()`? I know in some cases there will be `Close()` and `close()`, but the first one is temporal and will be removed later on. WDYT?

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -39,6 +40,7 @@ import (
 type Config struct {
 	MaxRows  int
 	Interval time.Duration
+	Count    int
```
mauriciovasquezbernal, line 43: Count -> Iterations?

integration/ig/k8s/top_file_test.go
```diff
@@ -23,49 +23,88 @@ import (
...
-
-			return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
-		},
-	}
-
-	commands := []*Command{
+	commandsPreTest := []*Command{
 		CreateTestNamespaceCommand(ns),
-		topFileCmd,
-		SleepForSecondsCommand(2), // wait to ensure ig has started
```
mauriciovasquezbernal, line 64: Don't we need this anymore?

mauriciovasquezbernal (commented):
Some more comments. Some concerns about `WaitForTimeoutOrDone`. Besides that I think it's good to go.

pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -273,32 +281,54 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
...
+			}
+			t.eventCallback(&top.Event[types.Stats]{Stats: stats[:n]})
 
-				n := len(stats)
-				if n > t.config.MaxRows {
-					n = t.config.MaxRows
+			if count > 0 {
+				count--
+				if count == 0 {
+					return nil
```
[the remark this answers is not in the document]
blanquicet: Yep, it is zero when the timeout is not set, so we don't need to keep track of the iterations. I added a comment.

pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -273,32 +281,54 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
...
-						Error: err.Error(),
-					})
-					return
-				}
+	for {
+		select {
+		case <-t.done:
+			// TODO: Once we completely move to use Run instead of NewTracer,
+			// we can remove this as nobody will directly call Stop (cleanup).
+			return nil
```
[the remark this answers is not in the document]
blanquicet: It is not possible to use `fallthrough` in this case.

pkg/gadgets/top/block-io/tracer/tracer.go
```diff
@@ -66,22 +68,30 @@ func NewTracer(config *Config, enricher gadgets.DataEnricherByMntNs,
...
 
 	statCols, err := columns.NewColumns[types.Stats]()
 	if err != nil {
-		t.Stop()
+		t.cleanup()
 		return nil, err
 	}
 	t.colMap = statCols.GetColumnMap()
 
+	go t.run(context.TODO())
```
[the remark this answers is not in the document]
blanquicet: No, this PR doesn't change the current behaviour. Let me try to explain the two scenarios quickly:

### Using `kubectl-gadget`
#### Currently
It uses `NewTracer` to create **and start** the tracer, as `start()` calls [`run()`](https://github.com/inspektor-gadget/inspektor-gadget/blob/50835b9f6f0561e70f7226eef1ce47a676fa4327/pkg/gadgets/top/block-io/tracer/tracer.go#L198). Then, everything was stopped and cleaned up by calling `Stop()`. The `Start()` was not involved in this case. 

_NOTE: Yeah, I agree the name NewTracer wasn't accurate. I had proposed renaming it https://github.com/inspektor-gadget/inspektor-gadget/issues/905)_

#### With this PR
It will still use `NewTracer` to create and start the tracer and stop for cleanup. Take into account this path will not be used anymore after https://github.com/inspektor-gadget/inspektor-gadget/pull/1396, but it will use what I described here just for `ig`.

### Using `ig`
#### Currently
It uses `NewInstance()` to **only** create the tracer and then `Start()` and `Stop()`.

#### With this PR
It will still use `NewInstance()` to create the tracer, but `Run` will fully replace `Start()` and `Stop()`.

I hope I explained myself well 🙂. Please let me know if that's not the case.

pkg/gadget-context/gadget-context.go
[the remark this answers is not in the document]
blanquicet: Yeah, I'll try, but I didn't want to waste time solving a lot of conflicts 🙈 .

pkg/gadgets/interface.go
```diff
@@ -108,33 +104,31 @@ type EventEnricherSetter interface {
...
+type RunWithResultGadget interface {
+	RunWithResult(GadgetContext) ([]byte, error)
 }
 
-type CloseGadget interface {
+// InitRunClose is used when gadgets need to be initialized before the operators
+// are installed. This is useful when operators need to interact with the
+// gadget during PreGadgetRun(). Notice Init() is not blocking but it returns
+// immediately after initializing the gadget. Run() must be called to wait for
+// the gadget to run, and Close() to clean up things.
```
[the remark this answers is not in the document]
blanquicet: Done!

pkg/gadget-context/gadget-context.go
blanquicet: TBH, I also had the same mixed feelings. And not that you mentioned it, I will make it a helper function.

integration/ig/k8s/top_file_test.go
blanquicet: We don't because we are now running `ig` after creating the pod.

internal/benchmarks/benchmarks_test.go
blanquicet: Mmmm, yes, we can, but is it a good practice to always call the `cancel()`?

pkg/gadgets/advise/seccomp/tracer/tracer.go
blanquicet: I tried to change it from `Close` to `Cleanup` because I want to avoid overlapping between gadgets implementing `RunGadget`/`RunWithResultGadget` with gadgets implementing `InitRunClose` (which requires `Close()` to be implemented).

I don't want to go to a `RunGadget` and see the `Run` and `Close` functions and think _"Where is the `Init`, is it missing?"_. WDYT?

internal/benchmarks/benchmarks_test.go
blanquicet: I wonder if we should support this zero timeout in the gadget-context. We could pass `timeout=-1` by default when it needs to wait forever, and any other value (including `0`) will be respected.

pkg/gadgets/interface.go
```diff
@@ -108,33 +104,47 @@ type EventEnricherSetter interface {
...
+	// and emit the events using the EventHandler. And, also also in this case,
+	// it is also expected to be blocking and return only when the context is
+	// done.
+	Run(GadgetContext) error
 
-type CloseGadget interface {
+	// Close is expected to clean up all the resources allocated by the
+	// gadget. It will be called after Run() returns and after the operators
+	// have been uninstalled.
 	Close()
```
flyth, line 144: Maybe add `GadgetContext` as parameter here as well? It's nothing expensive or confusing, but it might help not storing things, if we can get them inside `Close()` as well.

pkg/gadgets/interface.go
```diff
@@ -108,33 +104,47 @@ type EventEnricherSetter interface {
...
+	// EventHandler to emit the events, it returns the result of the gadget as a
+	// byte array after the context is done.
+	RunWithResult(GadgetContext) ([]byte, error)
+}
+
+// InitRunClose is a gadget interface that must be initialized before the
+// operators are installed. This is useful when the gadget needs to be kept
+// up-to-date with the containers that need to be traced, as the operators will
+// start sending notifications of this as soon as it is installed.
+type InitRunClose interface {
```
flyth, line 130: I think we should support this in a version with `RunWithResult` as well to keep it consistent. WDYT?

Having said that, I'd prefer to have an (optional) `InitClose` interface and then a combined mandatory `Run() ([]byte, error)` for both cases (with and without a result). I know not every gadget returns something here, but in that case you could just pass `nil` (and it's the same signature we use for `RunGadget()`). The advantage would be to have at least one interface that is bound to the `Gadget` interface (which otherwise goes back to being `any`).

pkg/runtime/local/local.go
```diff
@@ -55,31 +55,18 @@ func (r *Runtime) RunGadget(gadgetCtx runtime.GadgetContext) (out []byte, err er
...
-	defer func() {
-		if closer, ok := gadgetInstance.(gadgets.CloseGadget); ok {
-			log.Debugf("calling gadget.Close()")
-			closer.Close()
-		}
-
-		// No need to get results if gadget failed
+	// Initialize gadgets, if needed
+	if initRunClose, ok := gadgetInstance.(gadgets.InitRunClose); ok {
+		log.Debugf("calling gadget.Init()")
```
flyth, line 60: We could add more debug messages in the runtime to clarify things that don't happen, like "gadget.Init() not present, skipping" (in addition to the messages we already print), to better visualize the lifecycle of the gadget.

pkg/gadgets/interface.go
flyth: I don't have a good example right now (maybe params only needed in `Close()` or checking the embedded context) and in general, I agree that the interfaces should be as clean and simple as possible. However, we're defining an API here that we hopefully don't have to change often - having the GadgetContext also for `Close()` doesn't hurt IMHO as it doesn't introduce anything new and is merely a pointer - but it may come in handy someday (maybe even after we extend the context functionality itself).

blanquicet: Make sense. Done.

pkg/gadgets/interface.go
blanquicet: I try to see how it would look like in https://github.com/inspektor-gadget/inspektor-gadget/pull/1408/commits/e5a586a604794bd5dfe3d74cf67a45194b524df5.

pkg/gadgets/advise/seccomp/tracer/tracer.go
mauriciovasquezbernal: > I don't want to go to a RunGadget and see the Run and Close functions and think "Where is the Init, is it missing?". WDYT?

IMO it's not a big issue. I think it's better to call all functions that have a similar intention with the same name, otherwise we'll end with `cleanup()`,`close()`, `uninstall()` when all of them serve to the same purpose. 

However I don't have a strong opinion, if you think another naming makes more sense that's fine to me.

pkg/gadgets/interface.go
mauriciovasquezbernal: I was actually thinking the other way around: to remove `RunWithResult()` and only keep `Run()`, the result can be sent through the event handler. Could it work?

I feel like returning `[]byte` is very specific, so making it mandatory for all gadgets to return something seems weird, specially because they can't return any type but only `[]byte`.

flyth: Hmm. The good thing about the `[]byte` is that there's a guarantee that there can only be one per gadget run. And it can be transformed to whatever output type is supported/requested afterwards (the transformers accept `[]byte` as input, which also makes it easy to transport using gRPC). Using the `EventHandler` for that feels weird, as there's no such guarantee.
The cleanest solution would probably be to have something similar to the `EventHandler`, but called `SetResult` with a `ResultSetter` interface. But I guess that would feel clunky, so this is IMHO the next best solution.

blanquicet: I think we all have the same goal: Unify `Run` and `RunWithResult`. However, there are several options. So I suggest to unblock this PR and go ahead with both of them and unify them on a consecutive PR. WDYT?

pkg/gadgets/interface.go
blanquicet: I couldn't add it because there are some gadgets that needs to implement `Close(gadgetContext)` but they already implement `Close()` for the Tracer CR. I added a TODO on the interface and we can manage it on #1396.

pkg/gadgets/advise/seccomp/tracer/tracer.go
blanquicet: I moved back to `close()` because it will be unexported and it will not receive the gadget-context as one in the `InitClose` interface will do (see https://github.com/inspektor-gadget/inspektor-gadget/pull/1408#discussion_r1144074019). Therefore, it should be easy for developer to distinguish them.

pkg/gadgets/interface.go
mauriciovasquezbernal: > which also makes it easy to transport using gRPC

If the returned types can be serialized to json / binary we're all good, aren't we? 

Some questions and worries I have about returning `[]byte`:

- How does the enrichment work in that case? 
- Can we use it for all snapshotters and profilers?
- What about users that use the tracers directly? Do they need to parse the returned `[]byte` to have a golang type to work on?

mauriciovasquezbernal: > I suggest to unblock this PR and go ahead with both of them and unify them on a consecutive PR. WDYT?

I agree, let's unblock this one as soon as possible.

pkg/gadgets/trace/sni/tracer/tracer.go
```diff
@@ -92,17 +97,41 @@ func parseSNIEvent(sample []byte, netns uint64) (*types.Event, error) {
...
+		t.Close()
+		return fmt.Errorf("installing tracer: %w", err)
+	}
+
+	t.ctx, t.cancel = gadgetcontext.WithTimeoutOrCancel(gadgetCtx.Context(), gadgetCtx.Timeout())
+	return nil
+}
+
+func (t *Tracer) install() error {
+	// TODO: It is ugly. Clean it up.
```
mauriciovasquezbernal, line 117: I tried to handle this on https://github.com/inspektor-gadget/inspektor-gadget/commit/c4e1a45a1fae98d5e5e124046de2c28cdcacc235. Feel free to take and squash it if you agree.

pkg/runtime/local/local.go
```diff
@@ -106,47 +95,31 @@ func (r *Runtime) RunGadget(gadgetCtx runtime.GadgetContext) (out []byte, err er
...
 			return out, fmt.Errorf("running gadget: %w", err)
 		}
+	} else if runWithResult, ok := gadgetInstance.(gadgets.RunWithResultGadget); ok {
+		log.Debugf("calling gadget.RunWithResult()")
+		out, err = runWithResult.RunWithResult(gadgetCtx)
+		if err != nil {
+			return out, fmt.Errorf("running (with result) gadget: %w", err)
+		}
+	} else {
+		return nil, errors.New("gadget not runnable")
```
mauriciovasquezbernal, line 121: nit: There is a mix of  `return out,...`  and `return nil, ...`. Do we really need to make the return values "named" now?

mauriciovasquezbernal (approved):
LGTM, great job!

Some minor comments and a suggestion. Feel free to accept / reject them.

pkg/runtime/local/local.go
blanquicet: Yep, "named" return is not needed anymore. I removed it.

PR #1410, opened 2023-03-08
cmd: Check ig is run as root.

eiffel-fl:
```
Otherwise an error message is printed explaining why the tool should be run as root.
```

### DIFF
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
One issue with this approach is that we can't even print the help without running as root, that seems a bit weird.

```bash
./ig -h
./ig must be run as root to be able to run eBPF programs
```

Perhaps something like the following works better:

```go
	rootCmd := &cobra.Command{
		Use:   "ig",
		Short: "Collection of gadgets for containers",
		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
			if os.Geteuid() != 0 {
				return fmt.Errorf("%s must be run as root to be able to run eBPF programs", os.Args[0])
			}

			return nil
		},
	}
```

Issue #1378, opened 2023-02-27
`TraceDNS` test is failing many times


mauriciovasquezbernal:
I'm not sure if there is a problem with the gadget or the test itself is flaky, however this test is failing too often in different tests:
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4282772548/jobs/7457556219
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4282783360/jobs/7457751374
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4285322001/jobs/7463524979

We should investigate what's going and fix it.

mqasimsarfraz:
I created a [sandbox environment](https://github.com/mqasimsarfraz/actions-dns-testing/blob/main/.github/workflows/ci.yml#L11) to see if we can find any further details. I wanted to rule out following options:
- Docker Image (`busybox` vs `ubuntu`)
- DNS Server (`1.1.1.1` vs `8.8.4.4`)
- GitHub Actions

All the [jobs passed](https://github.com/mqasimsarfraz/actions-dns-testing/actions/runs/4313197392) multiple times so the DNS itself seems to be working fine in GH actions.

mauriciovasquezbernal:
In runs like https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4313357882/jobs/7525161469 I see it fails to contact the DNS server:

```
2023-03-02T11:20:52.6413834Z     command.go:266: Logs in namespace test-trace-dns-8868091403515467483:
2023-03-02T11:20:52.8584023Z     command.go:267: + kubectl get pods -n test-trace-dns-8868091403515467483 -o wide
2023-03-02T11:20:52.8584927Z         NAME       READY   STATUS    RESTARTS   AGE   IP            NODE                  NOMINATED NODE   READINESS GATES
2023-03-02T11:20:52.8585773Z         test-pod   1/1     Running   0          13s   10.244.1.12   minikube-docker-m02   <none>           <none>
2023-03-02T11:20:52.8586496Z         + kubectl get pods -n test-trace-dns-8868091403515467483 -o name
2023-03-02T11:20:52.8587184Z         + kubectl logs -n test-trace-dns-8868091403515467483 pod/test-pod
2023-03-02T11:20:52.8587747Z         ;; connection timed out; no servers could be reached
2023-03-02T11:20:52.8588123Z         
[...]

PR #1411, opened 2023-03-09
integration: Use 'dnstester' for k8s tests

mqasimsarfraz:
As mentioned [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/1407#issuecomment-1461623922) we had issues getting `kube-dns` service on ARO so I switched to using local dns in all the tests.

Related #1378

### DIFF
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -205,19 +205,6 @@ func GetTestPodIP(ns string, podname string) (string, error) {
 	return ip[1 : len(ip)-1], nil
 }
 
-func GetServiceIP(ns string, svcname string) (string, error) {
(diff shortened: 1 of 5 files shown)


mauriciovasquezbernal (approved):
Tests are passing hence LGTM!

(I also think it's more reliable to use our own binary than relying on any DNS server)

mauriciovasquezbernal:
Ah, I think we need an arm64 image of dnsenter too.

mqasimsarfraz:
> Ah, I think we need an arm64 image of dnsenter too.

Oh! Yeah, let me have a look!

PR #1414, opened 2023-03-09
tools: dnstester for arm64

mqasimsarfraz:
As mentioned [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/1411#issuecomment-1462359438) we need to create arm64 images for dnstester. This PR reworks the image to handle multi platform support. The corresponding arm64 [jobs](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4376025214/jobs/7661323843) seems to be passing fine now.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -416,6 +416,58 @@ jobs:
               ${{ steps.set-repo-determine-image-tag.outputs.container-repo }}@${{ needs.build-gadget-container-images.outputs.digest-default-arm64 }}
           fi
 
+  build-helper-images:
+    # level: 0
+    name: Build helper images
+    if: github.repository == 'inspektor-gadget/inspektor-gadget'
+    runs-on: ubuntu-latest
+    permissions:
+      # allow publishing container image
+      # in case of public fork repo/packages permissions will always be read
+      contents: read
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
+          - name: "dnstester"
+            dockerfile-dir: "tools/dnstester"
+            platform: "linux/amd64,linux/arm64"
+            filter-pattern: "tools/dnstester/*"
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          pattern: ${{ matrix.image.filter-pattern }}
+    - name: Set up Docker Buildx
+      if: steps.filter.outputs.pattern == 'true'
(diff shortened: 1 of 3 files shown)


mauriciovasquezbernal:
What do you think about adding a job in github actions to automatically update this image if changed? We can use https://github.com/dorny/paths-filter to only trigger this when the relevant files (tools/dnstester/*) are changed. 

It can be seem as some work upfront but in the long term I think that'd save us some headaches.

.github/workflows/inspektor-gadget.yml
```diff
@@ -416,6 +416,39 @@ jobs:
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build ${{ matrix.image.name }} image
+      uses: docker/build-push-action@v4
+      with:
+        context: ${{ matrix.image.dockerfile-dir }}
+        file: ${{ matrix.image.dockerfile-dir }}/Dockerfile
+        push: true
+        tags: ghcr.io/inspektor-gadget/${{ matrix.image.name }}:${{ matrix.image.tag }}
+        platforms: linux/amd64,linux/arm64
```
mauriciovasquezbernal, line 450: nit: probably this should be also in the matrix as I suspect we're going to push other images that don't need to be multiplatform using this action in the future, like ebpf-builder.

mqasimsarfraz:
> What do you think about adding a job in github actions to automatically update this image if changed? We can use https://github.com/dorny/paths-filter to only trigger this when the relevant files (tools/dnstester/*) are changed.
> 
> It can be seem as some work upfront but in the long term I think that'd save us some headaches.

Thanks for https://github.com/dorny/paths-filter, I included the filter. 

Also, my idea is to still push images for PRs with different tags to allow testing if needed and only do a push to `latest` when changes are merged. Does that make sense?

mauriciovasquezbernal:
> Also, my idea is to still push images for PRs with different tags to allow testing if needed and only do a push to latest when changes are merged. Does that make sense?

I don't have a strong opinion on this one. I see you already implemented this so I think we can keep this behavior. From time to time we could perform some manual cleanup of those tags but it's not so important as those images are intended for internal usage.

mauriciovasquezbernal (approved):
I see the run succeeded in https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/4384930120/jobs/7677008837 and the image was pushed https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/dnstester/76425713?tag=1414-merge. Approving it.

Thanks a lot for handling all the comments.

PR #1417, opened 2023-03-10
cmd: Replace -U and -K by --stack/-S for profile cpu.

eiffel-fl:
Hi.

In this PR, I replaced the two boolean flags `-U` and `-K` in favor of a string flag which takes the stack we want to print as argument.

Best regards.


alban:
Hi @eiffel-fl ,
Do you intend to fix the user stack as well? At the moment, it only prints 'unknown'. It seems our gadget only tries to resolve the kernel symbols with `readKernelSymbols()` regardless if it is a user or kernel stack.

Issue #1394, opened 2023-03-02
[RFE] Automatic tests with lots of pods on a node (scaling)


alban:
## Current situation

Kubernetes nodes typically support up to 110 pods per node but we don't have automatic tests checking whether Inspektor Gadget works fine with that amount of pods on a node.

## Impact

We won't notice if IG does not work with that amount of pods on a node. Regression bugs could go unnoticed

## Ideal future situation

We have integration tests that deploy 100 to 110 pods on a node and check the following:
- [ ] IG's memory usage stay reasonable
- [ ] IG's file descriptors don't reach a limit where it stops detecting new containers
- [ ] Filtering events (by pod name, labels...) continue to work
- [ ] Event enrichments continue to work
- [ ] IG does not slow down the creation of new containers
- [ ] Network gadgets such as dns (loading a new bpf module for each netns) continue to work
  - [ ] with reasonable performance (we get a new event fast without too much cpu consumption)
  - [ ] with reasonable memory consumption (e.g. ebpf maps)

## Additional information

This issue is about scaling the number of pods but only on one node. There is another discussion about scaling with many nodes in https://github.com/inspektor-gadget/inspektor-gadget/issues/519.

mqasimsarfraz:
It seem to be overlapping with #950, should we close #950 since this one has greater scope?

alban:
Hi @mqasimsarfraz, sorry I didn't see #950. Fine for me to close #950.

alban:
I tested the following manually with 100 pods on Minikube:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: testdeploy
  labels:
    app: testdeploy
spec:
  replicas: 100
  selector:
    matchLabels:
      app: testdeploy
  template:
    metadata:
      labels:
        app: testdeploy
    spec:
      containers:
      - name: testdeploy
        image: busybox
        command:
          - sleep
          - "infinity"
```

I chose the dns gadget because it loads a ebpf program for each pod:
```
$ time ./kubectl-gadget trace dns -v
```
Initially, it was able to start correctly in 2.5s. And I can get good output when testing:
```
$ kubectl exec -ti testdeploy-776799bb58-vcjwn -- wget wikipedia.org
```

But sometimes, it fails:
```
$ time ./kubectl-gadget trace dns -v
NODE               NAMESPACE          POD                PID        TID        COMM       QR NAMESERVER      TYPE       QTYPE     NAME              RCODE    
[...]

PR #1419, opened 2023-03-10
CI: benchmarks all gadgets and publish graphs

alban:
# CI: benchmarks all gadgets and publish graphs

This PR runs benchmarks for (almost) all gadgets with 100 fake containers. It uses the GitHub Action https://github.com/benchmark-action/github-action-benchmark to publish the graphs on a website automatically.

### Locally

Run benchmarks locally:
```
$ make gadgets-benchmarks
```

### GitHub Action

Use the CI, and check the results:

image
[screenshot source](https://github.com/alban/inspektor-gadget/actions/runs/4387055722/jobs/7681888513)

### Graphs

Check the graphs automatically updated for each commit pushed on the repository:
https://alban.github.io/ig-benchmarks/dev/bench/

## Testing done

```
BenchmarkAllGadgetsManyContainers/-traceloop-4         	       1	2023389297 ns/op
BenchmarkAllGadgetsManyContainers/advise-seccomp-4     	      26	  68125372 ns/op
BenchmarkAllGadgetsManyContainers/audit-seccomp-4      	      15	  67188176 ns/op
BenchmarkAllGadgetsManyContainers/profile-cpu-4        	       3	 522453773 ns/op
BenchmarkAllGadgetsManyContainers/snapshot-process-4   	       4	 496787210 ns/op
BenchmarkAllGadgetsManyContainers/snapshot-socket-4    	       1	82992027975 ns/op
BenchmarkAllGadgetsManyContainers/trace-bind-4         	       4	 285769542 ns/op
BenchmarkAllGadgetsManyContainers/trace-capabilities-4 	       4	 598973148 ns/op
BenchmarkAllGadgetsManyContainers/trace-dns-4          	       1	5628080722 ns/op
BenchmarkAllGadgetsManyContainers/trace-exec-4         	      12	  90112159 ns/op
BenchmarkAllGadgetsManyContainers/trace-fsslower-4     	       2	 624439805 ns/op
BenchmarkAllGadgetsManyContainers/trace-mount-4        	       7	 187561950 ns/op
BenchmarkAllGadgetsManyContainers/trace-network-4      	       2	 871403534 ns/op
BenchmarkAllGadgetsManyContainers/trace-oomkill-4      	      18	  63202527 ns/op
BenchmarkAllGadgetsManyContainers/trace-open-4         	       7	 259903084 ns/op
BenchmarkAllGadgetsManyContainers/trace-signal-4       	      10	 176863255 ns/op
BenchmarkAllGadgetsManyContainers/trace-sni-4          	       1	10987273427 ns/op
BenchmarkAllGadgetsManyContainers/trace-tcp-4          	       2	 736935149 ns/op
BenchmarkAllGadgetsManyContainers/trace-tcpconnect-4   	       5	 376750061 ns/op
```

We can see that the slowest gadgets are:
- traceloop
- snapshot socket
- trace dns
- trace sni

This is because they currently load one ebpf program per container.
See explanation in https://github.com/inspektor-gadget/inspektor-gadget/issues/1394#issuecomment-1456421534

## TODO

- [ ] Enable benchmarks for missing gadgets (see TODO in the code)
- [ ] Publish the results in a different repository than https://alban.github.io/ig-benchmarks/dev/bench/
- [ ] Update [index.html](https://github.com/alban/ig-benchmarks/blob/gh-pages/dev/bench/index.html) to add a graph to compare gadgets with each other

cc @flyth

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -548,6 +548,65 @@ jobs:
       run: |
         make controller-tests
 
+  benchmarks:
+    name: Benchmarks
+    # level: 0
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup go
+      uses: actions/setup-go@v3
+      with:
+        go-version: ${{ env.GO_VERSION }}
+        cache: true
+      id: go
+    - name: Install debian packages
+      uses: ./.github/actions/install-debian-packages
+    - name: Show repository setup
+      run: |
+        echo "github.event_name: ${{ github.event_name }}"
+        echo "github.event.pull_request.head.repo.full_name: ${{ github.event.pull_request.head.repo.full_name }}"
+        echo "github.repository: ${{ github.repository }}"
+    - name: Run benchmarks
+      run: go test -exec sudo -bench=. -run=Benchmark ./pkg/gadgets/... ./internal/benchmarks/... | tee output.txt
+    #- name: Download previous benchmark data
+    #  uses: actions/cache@v1
+    #  with:
+    #    path: ./cache
+    #    key: ${{ runner.os }}-benchmark
+    - name: Store benchmark result
+      uses: benchmark-action/github-action-benchmark@v1
+      # Disable push from forks or PR from forks.
+      # $BENCHMARKS_TOKEN will not be available in those cases.
+      if: |
+        (github.event_name == 'push' &&
+          github.repository == 'inspektor-gadget/inspektor-gadget') ||
+        (github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.full_name == 'inspektor-gadget/inspektor-gadget')
+      with:
+        name: Gadget benchmarks
+        # What benchmark tool the output.txt came from
+        tool: 'go'
+        # Where the output from the benchmark tool is stored
+        output-file-path: output.txt
+        # Where the previous data file is stored
+        # external-data-json-path: ./cache/benchmark-data.json
+        # Workflow will fail when an alert happens
+        fail-on-alert: false
+        # GitHub API token to make a commit comment
+        github-token: ${{ secrets.BENCHMARKS_TOKEN }}
+        # Enable alert commit comment
+        comment-on-alert: true
+        # Enable Job Summary for PRs
+        # summary-always: true
+        # Mention people in the commit comment
+        alert-comment-cc-users: '@alban'
+        # Push and deploy GitHub pages branch automatically
+        auto-push: ${{ github.repository == 'inspektor-gadget/inspektor-gadget' }}
+        gh-pages-branch: gh-pages
+        gh-repository: github.com/inspektor-gadget/ig-benchmarks
+        benchmark-data-dir-path: dev/bench
+
   test-ig:
     name: Unit tests for ig
     # level: 0
--- a/internal/benchmarks/benchmarks_test.go
+++ b/internal/benchmarks/benchmarks_test.go
@@ -1,0 +1,279 @@
+// Copyright 2019-2021 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package gadgets provides benchmarks for all gadgets. They can be executed
+// with the command:
+//
+//	$ make gadgets-benchmarks
+//
+// Results are published automatically by the GitHub Action:
+// https://inspektor-gadget.github.io/ig-benchmarks/dev/bench/index.html
+package gadgets
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/cilium/ebpf"
+
+	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
+	gadgetregistry "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-registry"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
+	tracercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/tracer-collection"
+
+	// TODO: find out why those gadgets don't work in the tests
+	//_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/block-io/tracer"
+	//_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/block-io/tracer"
+
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/advise/seccomp/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/audit/seccomp/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/profile/cpu/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/process/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/snapshot/socket/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/ebpf/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/file/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/tcp/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/tracer"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/capabilities/tracer"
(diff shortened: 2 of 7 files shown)


internal/test/runner.go
```diff
@@ -25,6 +25,8 @@ import (
 	"runtime"
 	"syscall"
 
+	"github.com/vishvananda/netlink"
+
```
mauriciovasquezbernal, line 29: nit: remove this empty line.

pkg/params/params.go
```diff
@@ -301,6 +301,9 @@ func (p *Param) AsFloat64() float64 {
 }
 
 func (p *Param) AsInt() int {
+	if p == nil {
+		return 0
+	}
```
mauriciovasquezbernal, line 306: I suppose we need to do it on all `As*` method. I'm wondering why somebody will call these on nil.

internal/benchmarks/benchmarks_test.go
```diff
@@ -0,0 +1,230 @@
...
+			runnerConfig := &utilstest.RunnerConfig{}
+			var containers []*containercollection.Container
+			for i := 0; i < containerCount; i++ {
+				runner := utilstest.NewRunnerWithTest(b, runnerConfig)
+				container := &containercollection.Container{
+					Netns: runner.Info.NetworkNsID,
+					Pid:   uint32(runner.Info.Tid),
+				}
+				containers = append(containers, container)
+			}
```
mauriciovasquezbernal, line 179: I'm wondering if this can be moved above to reuse the same fake containers for all gadgets.

internal/benchmarks/benchmarks_test.go
```diff
@@ -0,0 +1,230 @@
...
+		containers:   gadgetContext.Context().Value(TestOperatorContext("containers")).([]*containercollection.Container),
+		testingTB:    gadgetContext.Context().Value(TestOperatorContext("testing.TB")).(testing.TB),
+	}, nil
+}
+
+type TestOperatorInstance struct {
+	*TestOperator
+	attacher   Attacher
+	containers []*containercollection.Container
+	testingTB  testing.TB
```
mauriciovasquezbernal, line 122: Not used?

internal/benchmarks/benchmarks_test.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (t *TestOperatorInstance) PreGadgetRun() error {
+	for i := range t.containers {
+		t.attacher.AttachContainer(t.containers[i])
+	}
+	return nil
```
mauriciovasquezbernal, line 129: I'm wondering if this should also create a fake mntns map and call `SetMountNsMap` on the gadgets that support it. I know the cost of updating an eBPF map is much lower than the cost of attaching a new program, however to make the comparison fair I think we should also make it.

mauriciovasquezbernal (commented):
It looks very nice. I did a very quick pass. Will check later on with more detail.

pkg/params/params.go
alban: It is the `top block-io` gadget:
```
	t.config.MaxRows = params.Get(gadgets.ParamMaxRows).AsInt()
	t.config.SortBy = params.Get(gadgets.ParamSortBy).AsStringSlice()
```

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x6a118f]

goroutine 1379 [running]:
github.com/inspektor-gadget/inspektor-gadget/pkg/params.(*Param).AsInt(...)
	/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/params/params.go:307
github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/block-io/tracer.(*Tracer).Init(0xc0025ef400, {0x7f817ae384c0?, 0xc004656000?})
	/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/block-io/tracer/tracer.go:337 +0xaf
github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local.(*Runtime).RunGadget(0x1e28610?, {0x1e2d4d0, 0xc004656000})
	/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local/local.go:58 +0x20e
github.com/inspektor-gadget/inspektor-gadget/internal/benchmarks.BenchmarkAllGadgetsManyContainers.func2(0xc003d67680)
	/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/internal/benchmarks/benchmarks_test.go:232 +0x677
testing.(*B).runN(0xc003d67680, 0x1)
	/home/alban/programs/golang/go/src/testing/benchmark.go:193 +0x102
testing.(*B).run1.func1()
	/home/alban/programs/golang/go/src/testing/benchmark.go:233 +0x59
created by testing.(*B).run1
	/home/alban/programs/golang/go/src/testing/benchmark.go:226 +0x9c
exit status 2
```

alban: I am fixing this in benchmarks_test.go by fetching the missing params from the gadget type (`top` in this case).

Thanks @flyth for the explanation

internal/benchmarks/benchmarks_test.go
alban: It is used now :)

internal/benchmarks/benchmarks_test.go
alban: Done

alban:
> Is it possible to get stats on the code? Or only the mean?

@eiffel-fl What do you mean with stats on the code?

Here is an example of comment it posted on my fork:
https://github.com/alban/inspektor-gadget/commit/47a7ccb4280203501f1644c1dc4e5e5c4b7291cd#commitcomment-104269045

-----

For now the GitHub Action fails on this PR because I didn't configure the GitHub token in the repository.
- Is it ok if I configure a GitHub token for the benchmarks?
- Where should we publish the results? I suggest https://inspektor-gadget.github.io/ig-benchmarks/dev/bench/

alban:
> 99th percentile, mean, standard deviation and all this sort of stuff. From my understanding, `go test` only gives you an average given in `ns/op`. Average is cool, average plus standard deviation is cooler and 99th percentile is the best.

Ah, I see. Yes, the Go benchmarks has limitations... I've just read [this article](https://stonecode.ca/benchmarking-with-go/) which explains the limitations.

I used `benchstat` (see https://github.com/inspektor-gadget/inspektor-gadget/pull/1430) which gives additional stats like some confidence interval but unfortunately not all the stats you ask for.

I suggest to keep this PR as is. This can be improved later if needed.

flyth:
I really like having these kinds of stats over time 👍 

As discussed, [here's](https://gist.github.com/flyth/9eaeb144be34f9b4a90d7645639be47c) a patched version of the index.html handling the aggregation of stats. It's a bit hacky for now and can still be improved a lot, but I think it already helps to give a general overview.

Bildschirm­foto 2023-03-14 um 12 14 35

.github/workflows/inspektor-gadget.yml
```diff
@@ -548,6 +548,54 @@ jobs:
...
+        # What benchmark tool the output.txt came from
+        tool: 'go'
+        # Where the output from the benchmark tool is stored
+        output-file-path: output.txt
+        # Where the previous data file is stored
+        # external-data-json-path: ./cache/benchmark-data.json
+        # Workflow will fail when an alert happens
+        fail-on-alert: false
+        # GitHub API token to make a commit comment
+        github-token: ${{ secrets.BENCHMARKS_TOKEN }}
```
mqasimsarfraz, line 585: Why can't we use `GITHUB_TOKEN` here directly? Or `BENCHMARKS_TOKEN` offers something else?

mqasimsarfraz: Ah, makes sense. Thanks for clarifying. The reason I asked was, we could widen the [permission at job level](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/workflows/inspektor-gadget.yml#L227) for inspektor-gadget but it doesn't work for another repository.

internal/benchmarks/benchmarks_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+		err = validOperators.Init(nil)
+		if err != nil {
+			b.Fatalf("initializing operators: %s", err)
+		}
+		defer validOperators.Close()
+
+		b.Run(categoryAndName, func(b *testing.B) {
+			for n := 0; n < b.N; n++ {
+				ctx, cancel := context.WithTimeout(context.TODO(), 0)
```
mauriciovasquezbernal: Do you plan to pass a timeout there?

mauriciovasquezbernal (approved):
Two minor comments. LGTM. 

I didn't review anything related to html as it's not part of this PR. 

Would you mind adding some info about these tests in contributing as well linking to https://alban.github.io/ig-benchmarks/dev/bench/?

internal/benchmarks/benchmarks_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+				// Skip unwanted gadgets
+				categoryAndName := fmt.Sprintf("%s-%s", gadgetDesc.Category(), gadgetDesc.Name())
+				skipRegex := os.Getenv("IG_BENCHMARKS_GADGET_REGEX")
+				if skipRegex != "" {
+					matched, _ := regexp.Match(skipRegex, []byte(categoryAndName))
+					if !matched {
+						continue
+					}
+				}
```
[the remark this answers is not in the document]
alban: Good point. I didn't know how to make the regex work. It seems it is a list of `/`-separated regex: I have to use `^` and `$` on each section of the path to really select what I want, otherwise `container10` will also capture `container100`:
```
$ go test -exec sudo -bench='^BenchmarkAllGadgetsWithContainers$/^container10$/trace-tcpconnect' -run=Benchmark ./internal/benchmarks/...
```
I will add an example in the Makefile and remove the custom Golang regex code.

internal/benchmarks/benchmarks_test.go
alban: Timeout of 0s is necessary so it immediately stops the gadget once it has been started. I'll add a comment to explain.

Issue #801, opened 2022-08-03
Get rid / consider alternatives to entrypoint.sh


mauriciovasquezbernal:
The entrypoint of our container image is a bash script -> https://github.com/kinvolk/inspektor-gadget/blob/v0.7.0/gadget-container/entrypoint.sh. This script does the following:
- Print debug information like OS/kernel version, bcc version, IG version.
- Remount /sys/fs/bpf if it's not bpffs
- ~~Download and install kernel headers for rhcos when needed~~
- Install container hooks (OCI/CRI-O/NRI)
- ~~Install BTF information from BTFHub or from shipped BTF files when BTF is not available on the system~~
- Remove previous unix socket
- Start the gadget tracer manager

In order to perform those actions, the following binaries are needed in the container: (yes, I'm also surprised by the long list)
- ~~bash~~
- dpkg-query 
- awk
- grep 
- stat 
- mount 
- ~~modprobe~~
- ~~rmmod~~ 
- mkdir 
- ~~curl~~ 
- rm 
- ~~rpm2cpio~~
- ~~cpio~~
- mv 
- cp 
- ls 
- jq 
- tar
- ~~wget~~
- ~~objcopy~~

Having those in the ["default" image](https://github.com/kinvolk/inspektor-gadget/blob/v0.7.0/gadget-default.Dockerfile) is not a big deal, most of them are in the BCC base image we use, and we install the missing ones when building the container image. On the other hand, it's difficult to install all those dependencies on the ["core" image](https://github.com/kinvolk/inspektor-gadget/pull/608), specially when using a distroless base image. 

I think most of the logic of this file can be implemented directly on the gadget tracer manager, making it directly the entrypoint of the 
container. Specifically, I'd propose to implement the following directly on the gadget tracer manager:

- Print debug information like OS/kernel version, bcc version, IG version
- Remount /sys/fs/bpf if it's not bpffs
- Install container hooks (OCI/CRI-O/NRI) <-- not 100% sure 
- ~~Install BTF information from BTFHub or from shipped BTF files when BTF is not available on the system~~ (no needed after https://github.com/inspektor-gadget/inspektor-gadget/pull/2044)
- Remove previous unix socket <-- not sure about this one

ref https://github.com/kinvolk/inspektor-gadget/issues/625.

PR #1421, opened 2023-03-13
[RFC] Dockerfiles: Use busybox:glibc as base image for default image.

eiffel-fl:


Dockerfiles/gadget-core.Dockerfile
```diff
@@ -42,26 +42,18 @@ RUN cd /gadget/gadget-container && \
...
 	rm -f /etc/localtime && ln -sf /host/etc/localtime /etc/localtime
 
 COPY gadget-container/entrypoint.sh gadget-container/cleanup.sh /
 
 COPY --from=builder /gadget/gadget-container/bin/gadgettracermanager /bin/
 
+# We need to copy libseccomp.so and libdl.so from builder image as
+# gadgettracermanager relies on them.
+COPY --from=builder /usr/lib/x86_64-linux-gnu/libseccomp.so.2 /lib/x86_64-linux-gnu/libdl.so.2 /lib
+
```
mauriciovasquezbernal, line 56: Is this safe? Can we always use this regardless of the base image we use to build our container?

alban: I think it is not safe in general: `*.so` files can be compiled with different options across-distros so they might not have the same dependencies.

@eiffel-fl Did you have a specific issue with DNS in the gadget-core image? With wget in entrypoint.sh downloading from btfhub?

Dockerfiles/gadget-core.Dockerfile
```diff
@@ -42,26 +42,18 @@ RUN cd /gadget/gadget-container && \
...
-	if command -v tdnf; then \
-		tdnf install -y libseccomp wget util-linux; \
-	elif command -v yum; then \
-		yum install -y libseccomp wget util-linux; \
-	elif command -v apt-get; then \
-		apt-get update && \
-		apt-get install -y seccomp wget util-linux; \
-	elif command -v apk; then \
-		apk add gcompat libseccomp wget util-linux; \
-	fi && \
```
mauriciovasquezbernal, line 57: I think it'll break if a base image that doesn't provide the utilities needed in entrypoint.sh is used (#801).

mauriciovasquezbernal (commented):
I'm not sure if we should change this right now. Inspektor Gadget doesn't make a heavy use of DNS: (1) to download some files at startup on entrypoint.sh and (2) to contact the kubeapi-server, hence the lack of DNS over TCP support doesn't seem too relevant for us. Also, you mentioned this is being implemented...

I'd rather take the opportunity to discuss about distroless, once https://github.com/inspektor-gadget/inspektor-gadget/issues/801 is done, can we go distroless? Do we want to do it or is it better to have some utilities (like a package manager) available on our image?

alban:
In general, I am fine with switching to busybox if it works.

It removes socat from the image (`apt-get install -y util-linux socat` line 49), so this PR needs to be on hold until we have an alternative solution to socat.

Current error:

> time="2023-09-25T09:21:11Z" level=warning msg="connecting to gadget service on node \"minikube-containerd-m02\": Internal error occurred: error executing command in container: failed to exec in container: failed to start exec \"1258a9ac460cc719019e70954fcfd4968008c378b325706f9154594cd659dd3e\": OCI runtime exec failed: exec failed: unable to start container process: exec: \"/usr/bin/socat\": stat /usr/bin/socat: no such file or directory: unknown"

PR #1430, opened 2023-03-13
networking gadgets: instantiate ebpf prog once per netns

alban:
# networking gadgets: instantiate ebpf prog once per netns

In the tests with 100 containers on a node, the networking gadgets were
sometimes failing by timeout because they were too slow. Profiling
showed that most of the time is spent in the ebpf verifier. That is
because networking gadgets load the ebpf program once per network
namespace.

This patch refactors networking gadgets to use only one instance of ebpf
gadget. It is used for all network namespace. Unfortunately, we cannot
attach the ebpf program directly on the raw socket for each network
namespace because we need to know the netns in the ebpf program to get
the enrichment working properly.

So we have a new, very small ebpf program named 'dispatcher',
instantiated for each netns. Each instance is attached to a raw socket
on that netns. Each instance knows the current_netns then doing a tail
call on the singleton ebpf program. Since we cannot pass parameters in
tail calls, we use the skb->cb[] buffer to pass the context.

Since the dispatcher is very small, the ebpf verifier is still fast
enough, even with 100 instances.

Thanks @mauriciovasquezbernal for the idea to use skb->cb[] and tail calls.

The [documentation for skb->cb[]](http://vger.kernel.org/~davem/skb.html) says:
> This is the SKB control block. It is an opaque storage area usable by protocols, and even some drivers, to store private per-packet information. TCP uses this, for example, to store sequence numbers and retransmission state for the frame.

In the context of ebpf programs running on socket filters, the skb given to the ebpf program is a [skb clone](http://vger.kernel.org/~davem/skb.html):

> In order to make quick references to SKB data, Linux has the concept of SKB clones. When a clone of an SKB is made, all of the 'struct sk_buff' structure members of the clone are private to the clone. The data, however, is shared between the primary SKB and it's clone. When an SKB is cloned, the 'cloned' field will be set in both the primary and clone SKB. Otherwise is will be zero.

So reading/writing on skb->cb[] from the socket filter will not interfere with the networking stack.

## How to use

Run the benchmarks:
```
IG_BENCHMARKS_GADGET_REGEX='trace-[ds]n[si]' go test -exec 'sudo -E' -bench=. -run=Benchmark ./internal/benchmarks/... -count 8 |tee main.bench
IG_BENCHMARKS_GADGET_REGEX='trace-[ds]n[si]' go test -exec 'sudo -E' -bench=. -run=Benchmark ./internal/benchmarks/... -count 8 |tee patched.bench
```

Check the results:
```
$ benchstat main.bench patched.bench
goos: linux
goarch: amd64
                                     │ main.bench  │           patched.bench           │
                                     │   sec/op    │   sec/op    vs base               │
AllGadgetsManyContainers/trace-dns-4   2.945 ± 11%   1.584 ± 8%  -46.21% (p=0.000 n=8)
AllGadgetsManyContainers/trace-sni-4   4.735 ± 11%   1.523 ± 9%  -67.82% (p=0.000 n=8)
geomean                                3.734         1.553       -58.40%
```

Benchmarks showed a 58% performance improvement when starting a
networking gadget with 100 containers.

## TODO

- [x] Get https://github.com/inspektor-gadget/inspektor-gadget/pull/1419 merged first
- [x] Test the 3 networking gadgets
  - [x] dns
  - [x] sni
  - [x] network
- [x] Test with different frontends
  - [x] kubectl-gadget
  - [x] ig
- [x] Test the enrichment (node, pod, namespace...)
  - [x] in the main events
  - [x] in the debug events ("tracer attached" messages)

-----

This PR is part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1394

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1612


alban:
I ran more benchmarks with 0, 1, 10 and 100 containers:

```
IG_BENCHMARKS_GADGET_REGEX='[sd]n[is]' go test -exec 'sudo -E' -bench=. -run=Benchmark ./internal/benchmarks/... -count 8 | tee main.bench 
IG_BENCHMARKS_GADGET_REGEX='[sd]n[is]' go test -exec 'sudo -E' -bench=. -run=Benchmark ./internal/benchmarks/... -count 8 | tee patched.bench 

$ benchstat main.bench patched.bench
goos: linux
goarch: amd64
                                                  │  main.bench   │            patched.bench            │
                                                  │    sec/op     │    sec/op     vs base               │
AllGadgetsWithContainers/container0/trace-dns-4      854.2m ± 76%   797.1m ± 16%        ~ (p=0.721 n=8)
AllGadgetsWithContainers/container0/trace-sni-4      923.2m ± 22%   927.8m ± 16%        ~ (p=0.721 n=8)
AllGadgetsWithContainers/container1/trace-dns-4      878.3m ± 11%   806.2m ±  7%        ~ (p=0.279 n=8)
AllGadgetsWithContainers/container1/trace-sni-4      839.9m ± 27%   834.6m ± 21%        ~ (p=0.878 n=8)
AllGadgetsWithContainers/container10/trace-dns-4    1072.8m ± 16%   881.4m ± 22%  -17.84% (p=0.028 n=8)
AllGadgetsWithContainers/container10/trace-sni-4    1388.1m ± 52%   932.0m ± 11%  -32.86% (p=0.000 n=8)
AllGadgetsWithContainers/container100/trace-dns-4     3.345 ± 16%    1.511 ± 10%  -54.83% (p=0.000 n=8)
AllGadgetsWithContainers/container100/trace-sni-4     4.693 ± 14%    1.516 ± 12%  -67.70% (p=0.000 n=8)
geomean                                               1.386         991.8m        -28.43%
```

alban:
With the current state of this PR, the container enrichment does not work everywhere:
- **ig**
  - ✔️ **dns**
  - ✔️ **sni**
  - ❌ **network**: see explanation below
- ❌ **kubectl-gadget**: it does not use the new operators. I didn't investigate more; I am wondering whether to wait for #1396 to fix it.

Currently, `ig trace network -o json` shows events with netnsid=0. This is because the trace-network gadget does not use the `internal/networktracer` package like dns and sni, so it does not use the new dispatcher, so `skb->cb[0]` is not set. I see two choices to fix it:
1. Update the trace-network gadget to use `internal/networktracer`, so it uses the dispatcher, benefit from the perf improvements.
2. Continue to use `const volatile u64 container_netns` in `pkg/gadgets/trace/network/tracer/bpf/graph.c`.

I am thinking choice 1 is better, because we will need it eventually to have good performances in the network-policy advisor. Maybe this could be done before this PR? Then this PR will be a bit smaller to review.

mauriciovasquezbernal:
> I am thinking choice 1 is better

I agree. That'd be great if we're able to uniform those networking gadgets even more. It worries me a little bit that we're creating some dependencies between PRs but I think it's the only way to keep this organized and avoid a lot of conflicts.

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -189,36 +205,48 @@ func (t *Tracer[Event]) Attach(pid uint32, eventCallback func(*Event)) error {
...
+func (t *Tracer[Event]) EventCallback(event any) {
+	e, ok := event.(*Event)
+	if !ok {
+		panic("event handler argument invalid")
+	}
+	if t.eventHandler == nil {
+		return
+	}
+	t.eventHandler(e)
 }
```
mauriciovasquezbernal, line 239: I think it's kind of legacy to support `pkg/gadget-collection`, if that's right, can we put a comment stating that?

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -33,13 +34,16 @@ import (
...
+//go:generate bash -c "source ./clangosflags.sh; go run github.com/cilium/ebpf/cmd/bpf2go -target bpfel -cc clang dispatcher ./bpf/dispatcher.bpf.c -- $CLANG_OS_FLAGS -I./bpf/ -I../socketenricher/bpf"
+
 const (
 	SocketsMapName = "sockets"
 )
 
 type attachment struct {
-	collection *ebpf.Collection
-	perfRd     *perf.Reader
+	dispatcherSpec       *ebpf.CollectionSpec
```
mauriciovasquezbernal, line 44: Do we really need to keep the spec on the attachment struct? It seems it's only used in `newAttachment`, hence a local variable can work better.

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -33,13 +34,16 @@ import (
...
+	dispatcherCollection *ebpf.Collection
```
mauriciovasquezbernal, line 45: What about using `dispatcherObjects` to avoid having to manually access the maps and programs within the collection?

mauriciovasquezbernal (approved):
Some minor nits but looks great. Only blocking comment is missing call to `socketEnricher.Close()` when cleanup failed call to `NewTracer`. Approving since I think you can handle that without further discussion.

alban:
Thanks for the review, I updated the branch and I will merge once the CI is green.

PR #1438, opened 2023-03-14
CI: benchmarks all gadgets and publish graphs

alban:
Same as https://github.com/inspektor-gadget/inspektor-gadget/pull/1419

But the base of the PR is the upstream repository. This way I can check the GitHub action in both cases.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -548,6 +548,65 @@ jobs:
       run: |
         make controller-tests
 
+  benchmarks:
+    name: Benchmarks
+    # level: 0
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


alban:
It works:
https://inspektor-gadget.github.io/ig-benchmarks/dev/bench/index.html

mauriciovasquezbernal:
I have no idea how it works but both PRs (this and https://github.com/inspektor-gadget/inspektor-gadget/pull/1419) were merged to main but only a merge commit was created https://github.com/inspektor-gadget/inspektor-gadget/commit/901c9d58d23ebaaaf28801c2b626732d1b4f5933.

Commenting just in case somebody gets confused as me looking at this.

alban:
I only clicked on the merge button from the other PR. But since both PRs had exactly the same commits, I guess github noticed and marked this PR as "merged".

PR #1439, opened 2023-03-14
trace/network: Use ticker to query results

mauriciovasquezbernal:
The trace network gadget is designed to give a summary of the network connections performed by a pod / container. The eBPF program updates a map that is then read by the user space part. This operation happens each second, this is a tradeoff between reporting new connections fast enough and not using too many CPU to poll the maps.

This mechanism was wrongly removed in
9a0941298c63 ("update gadgets to use new interfaces"), this commit adds it again.

Fixes: 9a0941298c63 ("update gadgets to use new interfaces")


alban:
Well spotted.

But shouldn't the method Tracer.Start() start a go routine with the Ticker? The way it is done now, it seems that Tracer.Start() will not return until the gadget's context is cancelled. Other gadgets' Start() methods don't work like that: they return as soon as the gadget has been started successfully.

mauriciovasquezbernal:
You're right Alban, actually I took a shortcut as this is being changed in https://github.com/inspektor-gadget/inspektor-gadget/pull/1408, now `Run()` will be blocking (we'll get rid of start / stop).

alban:
Does it mean that #1408 should be merged first?

pkg/gadgets/trace/network/tracer/tracer.go
```diff
@@ -368,15 +369,20 @@ func (t *Tracer) Init(gadgetCtx gadgets.GadgetContext) error {
...
-		}
-		// Pop, but we don't need the results as we're handing the events over to the callback in cbMap
-		_, err := t.Pop()
-		if err != nil {
-			t.gadgetCtx.Logger().Debugf("pop() returned with: %v", err)
+		select {
+		case <-ctx.Done():
 			return nil
+		case <-ticker.C:
+			// Pop, but we don't need the results as we're handing the events over to the callback in cbMap
```
alban, line 380: What is `cbMap`? That comment was there before your patch but I don't understand what it refers to.

blanquicet: It refers to a callback map Mike was using during the refactoring but that we removed after a cleanup. So, the comment can be removed as it's completely out-of-date.

mauriciovasquezbernal:
> Does it mean that #1408 should be merged first?

Yes, I rebased on top of that and changed the target branch to that PR.

mauriciovasquezbernal:
This is now ready to review!

blanquicet (approved):
LGTM from code inspection

PR #1449, opened 2023-03-17
Use clang-15 to compile eBPF files

mauriciovasquezbernal:
While working on https://github.com/inspektor-gadget/inspektor-gadget/pull/1448 @wedaly found that there was a verifier error with older clang versions. So, we decided to update and use clang-15.


mauriciovasquezbernal:
I'll manually push the ebpf-builder image once we merge this PR.

PR #1450, opened 2023-03-17
pkg: Output DNS query and response at the same time

eiffel-fl:
There are two potential root cause:
1. Either this code is never called:

But I am highly doubtful because the condition should be true as we have `R(HOST)` in the output log
2. Or, there is something not good in this function:
https://github.com/inspektor-gadget/inspektor-gadget/blob/50835b9f6f0561e70f7226eef1ce47a676fa4327/pkg/gadgets/trace/dns/tracer/latency.go#L54
Indeed, if `Latency` is 0, it will never be printed:
https://go.dev/play/p/a6zIE6lWZql

Now, I am not sure if the problem comes from the above function to return 0 or for the computation to effectively be 0...


pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -143,7 +154,33 @@ int ig_trace_dns(struct __sk_buff *skb)
...
+
+		// Store the query in queries map.
+		err = bpf_map_update_elem(&queries_map, &key, &event, BPF_NOEXIST);
+		if (err != 0)
+			return 1;
+	} else {
+		struct event_t *query;
+
+		query = bpf_map_lookup_elem(&queries_map, &key);
+		if (query != NULL) {
```
mauriciovasquezbernal, line 173: What would happen if there is never a reply?

mauriciovasquezbernal (commented):
This reordering seems to be a well-known issue with perf ring buffers, @alban even wrote a blog post about it some years ago -> https://kinvolk.io/blog/2018/02/timing-issues-when-using-bpf-with-virtual-cpus/

I don't think the right solution for it is on eBPF, I think something like https://github.com/inspektor-gadget/inspektor-gadget/pull/1306 will fix this problem too. 

btw, I don't think your proposed solution is right because:
- It introduces a delay for all queries as they're only reported when there is a reply
- What about when there is not any reply?

PR #1461, opened 2023-03-21
Use cilium/ebpf API to read overwritable perf buffers

eiffel-fl:
Hi.

This PR fixed `traceloop` `AttachContainer()` by bringing back the previous "flight-recorder" behavior. 
It also makes use of upstream `cilium/ebpf` API to read overwritable perf buffers.

Best regards.


go.mod
```diff
@@ -149,4 +149,4 @@ replace (
 	github.com/gogo/protobuf v1.3.1 => github.com/gogo/protobuf v1.3.2
 )
 
-replace github.com/cilium/ebpf => github.com/inspektor-gadget/ebpf v0.0.0-20230306203928-c17e33ba8bac
+replace github.com/cilium/ebpf => github.com/cilium/ebpf v0.0.0-20230320181651-5ea5368f9823
```
alban, line 152: We can completely remove the `replace` line and specify the commit in the `require` declaration above.

pkg/gadgets/traceloop/tracer/bpf/traceloop.bpf.c
```diff
@@ -230,8 +230,11 @@ int ig_traceloop_e(struct bpf_raw_tracepoint_args *ctx)
 	}
 
 	bpf_debug_printk("Perf event output: sc.id: %d; sc.comm: %s; sizeof(sc): %d\n", sc.id, sc.comm, sizeof(sc));
-	bpf_perf_event_output(ctx, perf_buffer, BPF_F_CURRENT_CPU, &sc,
+	ret = bpf_perf_event_output(ctx, perf_buffer, BPF_F_CURRENT_CPU, &sc,
 			      sizeof(sc));
+	if (ret != 0) {
+		bpf_error_printk("Problem outputting perf event: %d", ret);
+	}
```
alban, line 237: Only half of the bpf_perf_event_output in this file are checked in this way. Is it intended?
Other gadgets don't check the result of bpf_perf_event_output. Did you have a problem with it?

pkg/gadgets/traceloop/tracer/tracer.go
```diff
@@ -607,7 +668,7 @@ func (t *Tracer) AttachContainer(container *containercollection.Container) error
 			}
 			evs, err := t.Read(container.ID)
 			if err != nil {
-				t.gadgetCtx.Logger().Debugf("error reading from container %s: %v", container.ID, err)
+				t.gadgetCtx.Logger().Errorf("error reading from container %s: %v", container.ID, err)
```
alban, line 671: @flyth I see that commit 9a0941298c639e4a52d1d98b9c69b1cb603f113a changed AttachContainer to immediately start a go routine to busy loop reading the buffer, making the traceloop gadget giving a stream of event instead of the flight recorder behaviour.

When I try this command:
```
$ sudo ./ig traceloop -c netem -r docker
```
It takes all the cpu (check with `top`) even without events.

Events are displayed instead of staying in the perf ring buffer until the user requests to get them by pressing ctrl-c as explained in [the traceloop documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadgets/traceloop.md#with-ig).


I think it should be fixed first before additional changes.

alban: It seems to work for me :)

PR #1480, opened 2023-03-29
Fix error handling

mauriciovasquezbernal:
Handle error events
    
    Add logic to handle the events that are use to provide debug information.
    Before this commit those events were erroneously handled as normal events
    printing wrong data to the terminal. For instance, for the DNS gadget:
    
    ```
    $ go build . && sudo ./ig trace dns
    CONTAINER                       PID        TID        COMM             QR NAMESERVER      TYPE      QTYPE      NAME                           RCODE
                                    0          0
                                    0          0
                                    0          0
    ```

### Testing

"broke" the dns gadget by applying the following diff

```diff
$ git diff
diff --git pkg/gadgets/trace/dns/tracer/tracer.go pkg/gadgets/trace/dns/tracer/tracer.go
index be5a5af5..9842ae02 100644
--- pkg/gadgets/trace/dns/tracer/tracer.go
+++ pkg/gadgets/trace/dns/tracer/tracer.go
@@ -286,7 +286,7 @@ func (t *Tracer) install() error {
 
        parseAndEnrichDNSEvent := func(rawSample []byte, netns uint64) (*types.Event, error) {
                bpfEvent := (*dnsEventT)(unsafe.Pointer(&rawSample[0]))
-               if len(rawSample) < int(unsafe.Sizeof(*bpfEvent)) {
+               if len(rawSample) != int(unsafe.Sizeof(*bpfEvent)) {
                        return nil, errors.New("invalid sample size")
                }
```

With this patch messages are now printed to the terminal:

```bash
$ go build . && sudo ./ig trace dns 
CONTAINER                       PID        TID        COMM             QR NAMESERVER      TYPE      QTYPE      NAME                           RCODE           
ERRO[0003] invalid sample size                          
ERRO[0003] invalid sample size                          
ERRO[0003] invalid sample size                          
```

```bash 
$ go build . && sudo ./ig trace dns -o json
{"type":"err","message":"invalid sample size"}
{"type":"err","message":"invalid sample size"}
{"type":"err","message":"invalid sample size"}
```

### DIFF
--- a/pkg/parser/outputhelpers.go
+++ b/pkg/parser/outputhelpers.go
@@ -18,6 +18,8 @@ import (
 	"encoding/json"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns/formatter/textcolumns"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
 // TextColumnsFormatter is the interface used for outputHelper
@@ -33,6 +35,11 @@ type TextColumnsFormatter interface {
 
 type ExtraLines interface {
 	ExtraLines() []string
+}
+
+type ErrorGetter interface {
+	GetType() types.EventType
+	GetMessage() string
 }
 
 // outputHelpers hides all information about underlying types from the application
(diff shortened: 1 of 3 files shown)


mauriciovasquezbernal:
> I'd really prefer to move towards using the logger in the `gadgetCtx` for most of these messages, once all refactoring is done and remove the event types like this for most of the gadgets.

I totally agree. One question I have is for the json output. Currently those events are printed using json, so a tool ingesting the data could handle them, if we use the logger, can we support this too or that will be lost?

flyth:
> > I'd really prefer to move towards using the logger in the `gadgetCtx` for most of these messages, once all refactoring is done and remove the event types like this for most of the gadgets.
> 
> I totally agree. One question I have is for the json output. Currently those events are printed using json, so a tool ingesting the data could handle them, if we use the logger, can we support this too or that will be lost?

It should be easy to rewire that in the registry.go when output is set to JSON, but I haven't tried it. We could also think about still outputting those to stderr, then, so that external tools could digest log messages (in a unified way) more easily.

PR #1483, opened 2023-03-30
ci: switch CI region

mauriciovasquezbernal:


mqasimsarfraz (approved):
LGTM! Worth a shot to run CI in different region! (AKS arm64 worked fine for me in `eastus`)

Issue #1484, opened 2023-03-31
ig not work on containerd CRI


vsxen:
## Environment and steps to reproduce

```
ig list-containers --containerd-socketpath  /run/containerd/containerd.sock -r containerd -v
WARN[0000] Runtime enricher (containerd): couldn't get current containers
DEBU[0000] Runcfanotify: trying runc at /usr/bin/runc
DEBU[0000] Runcfanotify: runc at /usr/bin/runc not found
DEBU[0000] Runcfanotify: trying runc at /usr/sbin/runc
DEBU[0000] Runcfanotify: trying runc at /usr/local/sbin/runc
DEBU[0000] Runcfanotify: runc at /usr/local/sbin/runc not found
DEBU[0000] Runcfanotify: trying runc at /usr/lib/cri-o-runc/sbin/runc
DEBU[0000] Runcfanotify: runc at /usr/lib/cri-o-runc/sbin/runc not found
DEBU[0000] Runcfanotify: trying runc at /run/torcx/unpack/docker/bin/runc
DEBU[0000] Runcfanotify: runc at /run/torcx/unpack/docker/bin/runc not found
RUNTIME ID NAME
root@ubuntu-focal:/home/vagrant# crictl -r /run/containerd/containerd.sock ps
I0331 18:58:59.936226    2811 util_unix.go:103] "Using this endpoint is deprecated, please consider using full URL format" endpoint="/run/containerd/containerd.sock" URL="unix:///run/containerd/containerd.sock"
I0331 18:58:59.941446    2811 util_unix.go:103] "Using this endpoint is deprecated, please consider using full URL format" endpoint="/run/containerd/containerd.sock" URL="unix:///run/containerd/containerd.sock"
CONTAINER           IMAGE               CREATED             STATE               NAME                      ATTEMPT             POD ID
4b5e4ea4e88ac       c1cfbd59f7747       5 minutes ago       Running             kube-proxy                0                   b4a4ac6fb0a1e
1330171e83a71       398b2c18375df       6 minutes ago       Running             kube-scheduler            42                  6453081689e84
673818c4a67b1       0048118155842       6 minutes ago       Running             etcd                      35                  9ccc8f21dac22
6d6d38304157d       c0d565df2c900       6 minutes ago       Running             kube-apiserver            16                  68f95ca3a65c0
4d70c7657166b       41ff053508988       6 minutes ago       Running             kube-controller-manager   44                  10e4e1ebb2afb
[...]

PR #1485, opened 2023-03-31
use k8s.io/cri-api v0.25.x

mauriciovasquezbernal:
v0.26.x breaks compatibility with containerd v1.5. Revert https://github.com/inspektor-gadget/inspektor-gadget/pull/1392 and use the latest v0.25.x release of k8s.io/cri-api. 

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1484


mqasimsarfraz:
The change is looking good but we should plan when do we want to switch to cri api v1 ? I see following state for different projects: 

- Kubernetes - [Removed support for v1alpha2 from `v1.26.0`](https://github.com/kubernetes/kubernetes/pull/110618)
- CRI-O - [Removed support for v1alpha2 from `v1.26.0`](https://github.com/cri-o/cri-o/pull/6347/commits/ef3bed00b9dbf62e80a18a36afb512f2a284a786)
- Containerd - [Plans to remove support of v1alpha2 from v2.0](https://github.com/containerd/containerd/blob/main/RELEASES.md#kubernetes-support)

Since we talking to cri-o/containerd using CRI makes sense to define a timeline to support `containerd v2.0` and `cri-o v1.26.0+`.    How about creating a support matrix for supported CRI version for `ig` and drop `v1alpha2` support after couple of releases with a deprecation notice?

blanquicet (approved):
I just realized I was suffering from this issue in one of my dev environments 😅 . This PR indeed solves the issue.

mauriciovasquezbernal:
Great analysis @mqasimsarfraz. This is a tricky situation for Inspektor Gadget, we want to support as many as possible scenarios. I'm wondering if there is a way to keep supporting both (v1 and v1alpha2) at the same time. I'd be great to keep supporting users that can't update to v1 too. I'll merge this and open an issue to study this.

PR #1488, opened 2023-03-31
gadgets: add helper function for mapReplacement

claudiamarcubina:

## How to use

## Testing done


pkg/gadgets/helpers.go
```diff
@@ -213,3 +214,31 @@ func FixBpfKtimeGetBootNs(programSpecs map[string]*ebpf.ProgramSpec) {
 		removeBpfKtimeGetBootNs(s)
 	}
 }
+
+func FilterMapReplacement(MountnsMap *ebpf.Map, spec *ebpf.CollectionSpec, objs interface{}) error {
```
mauriciovasquezbernal, line 218: Some thoughts about this:
- What about making this function more generic, like `LoadeBPFSpec` since it's not only about replacing the filtering map?
- I think it can have a `consts map[string]interface{}` parameter so it can be used in other gadgets too. 
- It could also call `FixBpfKtimeGetBootNs` to avoid each gadget doing the same 
- https://github.com/inspektor-gadget/inspektor-gadget/pull/1504 could be implemented here to avoid code duplication

claudiamarcubina:
``` 
> What about making this function more generic, like LoadeBPFSpec since it's not only about replacing the filtering map?
> I think it can have a consts map[string]interface{} parameter so it can be used in other gadgets too.
```
Added, Mauricio, could you please check?

```
> It could also call FixBpfKtimeGetBootNs to avoid each gadget doing the same
```
Good hint. I included `FixBpfKtimeGetBootNs` in the helper as well. 
I just have one mention, top/block-io gadget wasn't using `FixBpfKtimeGetBootNs`, but I think it makes no big difference that it's called now through `LoadeBPFSpec `.

CC: @mauriciovasquezbernal

claudiamarcubina:
Solving the merge conflicts.

pkg/gadgets/helpers.go
```diff
@@ -224,3 +225,57 @@ func FixBpfKtimeGetBootNs(programSpecs map[string]*ebpf.ProgramSpec) {
...
+	}
+
+	if consts == nil {
+		consts = map[string]interface{}{}
+	}
+
+	consts["filter_by_mnt_ns"] = filterByMntNs
+
+	if err := spec.RewriteConstants(consts); err != nil {
+		return fmt.Errorf("error RewriteConstants: %w", err)
```
mauriciovasquezbernal, line 271: nit: about adding `error`, `fail`, etc in these messages -> https://github.com/inspektor-gadget/inspektor-gadget/issues/881

(I know we have a bunch of code using that, but we're trying not to introduce new code with that issue)

mauriciovasquezbernal (commented):
I like it. Do you plan to use it in all / most of the gadgets?

pkg/gadgets/helpers.go
mauriciovasquezbernal: > I just have one mention, top/block-io gadget wasn't using FixBpfKtimeGetBootNs, but I think it makes no big difference that it's called now through LoadeBPFSpec .

Yes, I don't think it adds a huge overhead, so it's fine to call it for all gadgets.

pkg/gadgets/helpers.go
claudiamarcubina: Good to know about this effort. Re-worded that ✅

claudiamarcubina:
> I like it. Do you plan to use it in all / most of the gadgets?

Hm, used it in the gadgets that seemed most obvious to me so far. Checking where else it could be used,

claudiamarcubina:
> > I like it. Do you plan to use it in all / most of the gadgets?
> 
> Hm, used it in the gadgets that seemed most obvious to me so far. Checking where else it could be used,

I think I covered all gadgets that could use the new helper now

mauriciovasquezbernal:
Hi @claudiamarcubina, unfortunately I have to merge https://github.com/inspektor-gadget/inspektor-gadget/pull/1580 before this PR and it caused some conflicts. I hope they're not superdifficult to fix. I'll check this again once those are fixed.

pkg/gadgets/helpers.go
```diff
@@ -232,3 +233,57 @@ func FixBpfKtimeGetBootNs(programSpecs map[string]*ebpf.ProgramSpec) {
...
+	}
+	if opts.MapReplacements == nil {
+		opts.MapReplacements = map[string]*ebpf.Map{}
+	}
+	if mountnsMap != nil {
+		opts.MapReplacements[MntNsFilterMapName] = mountnsMap
+	}
+
+	return opts
+}
```
mauriciovasquezbernal, line 251: I'm not sure this function is needed. I think we can put this code directly into `LoadeBPFSpec`

pkg/gadgets/helpers.go
```diff
@@ -232,3 +233,57 @@ func FixBpfKtimeGetBootNs(programSpecs map[string]*ebpf.ProgramSpec) {
...
+
+// LoadeBPFSpec collects operations that are duplicated across gadgets.
+// It replaces filter map and calls the necessary function to load
+// Maps and Programs into the kernel
```
mauriciovasquezbernal, line 255: What about?

```suggestion
// LoadeBPFSpec is a helper to load an eBPF spec from gadgets.
// It replaces filter map and calls the necessary functions to load
// Maps and Programs into the kernel
```

pkg/gadgets/profile/cpu/tracer/tracer.go
```diff
@@ -327,30 +327,13 @@ func (t *Tracer) install() error {
...
-	opts := ebpf.CollectionOptions{
-		MapReplacements: mapReplacements,
+		"kernel_stacks_only": t.config.KernelStackOnly,
+		"user_stacks_only":   t.config.UserStackOnly,
 	}
 
-	if err := spec.LoadAndAssign(&t.objs, &opts); err != nil {
-		return fmt.Errorf("failed to load ebpf program: %w", err)
+	if err := gadgets.LoadeBPFSpec(t.config.MountnsMap, spec, consts, &t.objs); err != nil {
+		return fmt.Errorf("loading ebpf program: %w", err)
```
mauriciovasquezbernal, line 336: nit: program -> spec. (It could happen that the error is related to a map, so program will be misleading in that case)

pkg/gadgets/helpers.go
```diff
@@ -232,3 +233,57 @@ func FixBpfKtimeGetBootNs(programSpecs map[string]*ebpf.ProgramSpec) {
...
+	objs interface{},
+) error {
+	FixBpfKtimeGetBootNs(spec.Programs)
+
+	mapReplacements := map[string]*ebpf.Map{}
+	filterByMntNs := false
+
+	if mountnsMap != nil {
+		filterByMntNs = true
+		mapReplacements["mount_ns_filter"] = mountnsMap
```
mauriciovasquezbernal: nit: The `"mount_ns_filter` -> `MntNsFilterMapName` should be done in this commit. (Not in the second one). However, you can squash both commits if you want to avoid dealing with this issue.

mauriciovasquezbernal (commented):
It's looking great and almost ready to go. To very small comments, will approve once they're addressed.

pkg/gadgets/helpers.go
claudiamarcubina: ✅

pkg/gadgets/profile/cpu/tracer/tracer.go
claudiamarcubina: ✅  Perfect. I wasn't 100% convinced that the error message is the right one.

pkg/gadgets/helpers.go
claudiamarcubina: Right, thanks. I moved it to the first one 😄

mauriciovasquezbernal (approved):
LGTM, thanks!

PR #1489, opened 2023-04-03
cmd/kubectl-gadget/advise: add locks to writes for policy output

### LINKED ISSUE #1487: Corrupted network policy monitor output
## Description

When running the network policy monitor the json output seems to be getting corrupted.  As an example `Error: cannot parse line 577: invalid character 's' looking for beginning of value
`
## Impact

Unable to run network policy report without removing the corrupted lines which may be losing data for the corrupted line.

1. **Set-up**: EKS with gadget v0.14.0
2. **Task**: Running a network policy monitor and report
3. **Action(s)**:
  a. `k gadget advise network-policy monitor  --output np.json`
  b. `kubectl gadget advise network-policy report --input np.json`
4. **Error**: Error: cannot parse line 577: invalid character 's' looking for beginning of value

## Expected behavior

The json output for the monitor should not be corrupted and the report should run fine.

## Additional information

Please add any information here that does not fit the above format.

flyth:
Concurrent writes (from different nodes) to the underlying file/bufio.Buffer could lead to scrambled output. Adding a mutex makes sure that a line gets written as whole, before the next line is written.

We should check this behavior also for the new architecture, but in the meantime, this should fix #1487.

### DIFF
--- a/cmd/kubectl-gadget/advise/network-policy.go
+++ b/cmd/kubectl-gadget/advise/network-policy.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/advise/network-policy.go
```diff
@@ -94,11 +95,16 @@ func runNetworkPolicyMonitor(cmd *cobra.Command, args []string) error {
...
 	}
+
+	var mu sync.Mutex
+
 	count := 0
 	transform := func(line string) string {
 		line = strings.Replace(line, "\r", "\n", -1)
+		mu.Lock()
 		w.Write([]byte(line))
 		w.Flush()
```
burak-ok, line 106: nit: Can the `.Flush()` be moved after the `.Unlock()`?

burak-ok (approved):
LGTM. Comment is a nit anyway since we will have a new architecture

cmd/kubectl-gadget/advise/network-policy.go
flyth: I think the operation of writing and flushing should be "atomic" - it seems intended that a flush should happen after each line (and it's also important because the underlying buffer is limited) - which would not be guaranteed when calling it outside the lock.

PR #1492, opened 2023-04-03
pkg/ig-manager: Check ig manager is run as root.

eiffel-fl:
```
ig version does not run any eBPF program so it does not need to be run as root.

Fixes: 13cf80d62765 ("cmd: Check ig is run as root.")
```

### DIFF
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
@@ -37,13 +36,6 @@ func main() {
 	rootCmd := &cobra.Command{
 		Use:   "ig",
 		Short: "Collection of gadgets for containers",
-		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			if os.Geteuid() != 0 {
-				return fmt.Errorf("%s must be run as root to be able to run eBPF programs", os.Args[0])
-			}
-
-			return nil
-		},
 	}
 
 	rootCmd.AddCommand(
--- a/pkg/runtime/local/local.go
+++ b/pkg/runtime/local/local.go
@@ -17,6 +17,8 @@ package local
 import (
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	gadgetregistry "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-registry"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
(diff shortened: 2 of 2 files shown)


cmd/ig/main.go
```diff
@@ -38,7 +38,8 @@ func main() {
 		Use:   "ig",
 		Short: "Collection of gadgets for containers",
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			if os.Geteuid() != 0 {
+			// We allow ig version to not be run as root.
+			if os.Geteuid() != 0 && os.Args[1] != "version" {
```
flyth, line 42: Should check for `len(os.Args) > 1` before accessing `os.Args[1]` here. Maybe you can use `args[0]` instead (also with a proper length check).

alban:
Some other commands should be accepted without root:
```
$ ig completion bash # and fish, powershell, zsh
Error: ./ig must be run as root to be able to run eBPF programs
$ ig help
Error: ./ig must be run as root to be able to run eBPF programs
```

pkg/ig-manager/ig-manager.go
```diff
@@ -83,6 +85,11 @@ func NewManager(runtimes []*containerutils.RuntimeConfig) (*IGManager, error) {
 		return nil, err
 	}
 
+	// Check this now otherwise below RemoveMemlock will fail.
+	if os.Geteuid() != 0 {
+		return nil, errors.New("ig must be run as root to be able to run eBPF programs")
```
alban, line 90: I tested it and it looks like this:
```bash
$ ./ig trace exec
Error: initializing operators: initializing operator "LocalManager": initializing manager: ig must be run as root to be able to run eBPF programs
```

Before this patch, it looked like this:
```bash
$ ./ig trace exec
Error: ./ig must be run as root to be able to run eBPF programs
```

I think the more concise output is better.

alban: Indeed, nothing is perfect. I think your suggestion 3 is the best.

```go
import  "golang.org/x/exp/slices"

verbsAllowedByNonRoot := []string{"help", "completion", "version"}
if os.Geteuid() != 0 &&  len(os.args) > 1 && !slices.Contains(verbsAllowedByNonRoot, os.args[1]) {
}
```

flyth: What about adding the check in [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/runtime/local/local.go#L53) - that is always called before running an actual gadget and in the error case it should return `Error: initializing runtime: ig must be run as root to be able to run eBPF programs`. I think that could be a good compromise.

flyth: > I just remember about the existing of os.Args[0] for this case :p!

You might want to wrap that in a `filepath.Base(os.Args[0])`, then - otherwise, if called with a path, that would show up as well.

mauriciovasquezbernal (approved):
Nice! Thanks all for the discussion on this one!

PR #1502, opened 2023-04-07
Dockerfile: Use golang:1.19 as builder image.

eiffel-fl:
Before this commit, we used debian:bullseye as base image. This image was used because it permits easily to cross-build by installing package for foreign architecture as done in commit 8b261be730ab ("Dockerfile: Use debian as builder instead of ubuntu.") Nonetheless, we would need to download golang from a PPA which does not deliver an up to date golang version as pointed by docker scout [1].

So, in this commit, we use golang:1.19 as base image. This offers us the best of both world as we would have an up to date golang and it is also easy to cross build as golang docker image are based on debian.

[1]: https://github.com/kubernetes/minikube/pull/15869#issuecomment-1499649317

### DIFF
--- a/Dockerfiles/gadget-core.Dockerfile
+++ b/Dockerfiles/gadget-core.Dockerfile
@@ -5,7 +5,7 @@
 # and is designed to be used on systems that support BTF
 # (CONFIG_DEBUG_INFO_BTF).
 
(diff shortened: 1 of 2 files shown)


mauriciovasquezbernal (approved):
Tests are passing fine and I don't see any reason not to merge this. LGTM!

PR #1518, opened 2023-04-12
kubeipresolver: refactor in preparation for use in tcpdrop

alban:
Currently, kubeipresolver is only used by the 'trace network' gadget and it can resolve 1 local IP and 1 remote IP. In the 'trace tcpdrop' gadget, we will need to resolve 2 IPs (source IP and destination IP) and they might not be local IPs (i.e. not tied to a pod).

This refactoring does not change anything for the 'trace network' gadget but will allow kubeipresolver to be used by other gadgets.

## How to use

No changes.

## Testing done

Only the CI

### DIFF
--- a/pkg/operators/kubeipresolver/kubeipresolver.go
+++ b/pkg/operators/kubeipresolver/kubeipresolver.go
@@ -43,20 +43,11 @@ const (
 )
 
 // KubeNetworkInformation is for now a specific interface for `trace network` gadget
-// TODO: More granular and "shareable" interfaces
 type KubeNetworkInformation interface {
-	// Local/Host
-	SetPodOwner(string)
-	SetPodHostIP(string)
-	SetPodIP(string)
-	SetPodLabels(map[string]string)
+	SetLocalPodDetails(owner, hostIP, podIP string, labels map[string]string)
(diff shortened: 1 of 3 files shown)


pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -235,62 +226,71 @@ func (m *KubeIPResolverInstance) enrich(ev any) {
...
+			if svc.Spec.ClusterIP == remoteIP {
+				endpoints[j].Kind = types.RemoteKindService
+				endpoints[j].Name = svc.Name
+				endpoints[j].Namespace = svc.Namespace
+				endpoints[j].PodLabels = svc.Labels
+			}
 		}
 	}
+
+	if foundRemote > 0 {
```
burak-ok, line 290: If we can't resolve any IPs to K8s resources, shouldn't `.SetEndpointDetails(endpoints)` still be called?
So the tracer gets the info that all IPs are of type `types.RemoteKindOther`

burak-ok (approved):
LGTM and I like the `EndpointDetails` struct that summarizes everything :)

PR #1528, opened 2023-04-14
registry: Make mandatory parameters required cmd flags

mauriciovasquezbernal:
Before:
    
    ```
    $ ./kubectl-gadget script
    NODE                               OUTPUT
    Error: running gadget: rpc error: code = Unknown desc = setting gadget parameters: expected value for "program"
    ```
    
    ^^^ validation was performed in the server side
    
    Now:
    
    ```
    $ ./kubectl-gadget script
    Error: required flag(s) "program" not set
    ```
    
    Validation is done by cobra on the client side.

Ref https://github.com/inspektor-gadget/inspektor-gadget/pull/1370#pullrequestreview-1383795267

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -550,6 +550,9 @@ func addFlags(cmd *cobra.Command, params *params.Params, skipParams []params.Val
 		}
 
 		flag := cmd.PersistentFlags().VarPF(p, p.Key, p.Alias, desc)
+		if p.IsMandatory {
+			cmd.MarkPersistentFlagRequired(p.Key)
+		}
 
 		// Allow passing a boolean flag as --foo instead of having to use --foo=true
 		if p.IsBoolFlag() {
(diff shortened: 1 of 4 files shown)


cmd/common/registry.go
```diff
@@ -550,6 +550,9 @@ func addFlags(cmd *cobra.Command, params *params.Params, skipParams []params.Val
 		}
 
 		flag := cmd.PersistentFlags().VarPF(p, p.Key, p.Alias, desc)
+		if p.IsMandatory {
```
flyth, line 553: I saw that you changed a couple of mandatory flags, because they had default values - still, I think we're also using `IsMandatory` to check for non-empty values. Maybe this is cleaner (can have a default value but still can't be set to empty value):

```suggestion
		if p.IsMandatory && p.String() != "" {
```

But I'm fine either way.

flyth (approved):
Thanks!

cmd/common/registry.go
mauriciovasquezbernal: The purpose of this check in L553 and `MarkPersistentFlagRequired()` is to make the CLI flag required if the parameter is mandatory. 

```bash 
$ ./kubectl-gadget script
Error: required flag(s) "program" not set
```

The check to be sure a mandatory parameter is not empty is done in https://github.com/inspektor-gadget/inspektor-gadget/blob/9dd4b9248ab171266879c030335cab4d7e27d895/pkg/params/params.go#L112-L114

```bash
$ ./kubectl-gadget script --program ""
Error: invalid argument "" for "-e, --program" flag: expected value for "program"
```

So I think we're all good.

Issue #1530, opened 2023-04-14
[RFE] Find a way to provide `bpftrace` for `arm64`


eiffel-fl:
## Current situation

[ Please describe the current situation you would like to have improved ]

## Impact

[ Please describe the impact the lack of the feature requested is creating ]

## Ideal future situation

[ Please describe the future situation after the improvement was implemented ]

## Implementation options

[ Optional: please provide one or more options for implementing the feature requested ]

## Additional information

[ Please Add any information that does not fit into any of the above sections here ]

mauriciovasquezbernal:
I made some progress in https://github.com/inspektor-gadget/bpftrace/commit/655f19550458c1f16e1acfd8ba2b891df6794779, feel free to take it if you think it's useful.

mauriciovasquezbernal:
After 4 hours of cross compilation on my machine, I was able to push the image to https://hub.docker.com/repository/docker/mauriciovasquezbernal/bpftrace/general. I haven't try it out with Inspektor Gadget yet.

mauriciovasquezbernal:
I tested it and works fine on arm64, the image I used is in https://hub.docker.com/repository/docker/mauriciovasquezbernal/gadget/. 

We need to think about the compilation time, specially for the llvm image since it takes  time is 4hours in my machine. I don't think it's suitable for github actions. 
-

mauriciovasquezbernal:
I need to be more clear about my terminology. I used `docker buildx` with qemu to **emulate** the whole compilation on arm64.

PR #1531, opened 2023-04-14
Show error for bpftrace when not running on amd64

mauriciovasquezbernal:
So far only amd64 is supported, skip the tests and show a clear error in that platform.


mauriciovasquezbernal:
ref https://github.com/inspektor-gadget/inspektor-gadget/issues/1530

Issue #1515, opened 2023-04-11
ig does not work with podman due to lack of crun support


blixtra:
## Description

I was trying to run `ig list-containers` on a Ubuntu 22.04 with podman and podman-docker installed. But this failed because ig only works with runc, not crun which is the OCI Runtime that podman uses, at least in Ubuntu 22.04. 

When trying I get the following output.

```$ sudo ig list-containers
WARN[0000] Runtime enricher (containerd): couldn't get current containers
WARN[0000] Runtime enricher (cri-o): couldn't get current containers
Error: initializing manager: cannot start runc fanotify: no runc instance can be monitored with fanotify
```

@rata pointed out that we only support runc here: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/runcfanotify/runcfanotify.go#L104-L110

## Impact

ig does not recognize containers when running on Ubuntu 22.04 using podman, or likely any container engine that uses crun.

blixtra:
For those listening, one can get around this by using the runtime flag of podman before the command to force podman to use runc. For example, `podman --runtime /usr/sbin/runc <COMMAND>` on Ubuntu.

Issue #1507, opened 2023-04-10
Support CRI `v1` and `v1alpha` at the same time


mauriciovasquezbernal:
#1392 switched the CRI api from `v1alpha2` to `v1`, however it created an issue (#1484) as containerd v1.5 doesn't support `v1`, #1485 reverted it back to `v1alpha2` to support containerd v1.5. 

Different runtimes removed (or are removing) support for v1alpha (https://github.com/inspektor-gadget/inspektor-gadget/pull/1485#issuecomment-1494116963), hence we should make Inspektor Gadget compatible with both versions. containerd and cri-o support(ed) both versions at the same time https://github.com/cri-o/cri-o/pull/4408 & https://github.com/containerd/containerd/pull/5619, hence I think this is doable from a technical point of view.

mqasimsarfraz:
In case we want to bump the k8s.io/cri-api to >= v0.26.0 we need to copy cri-api v1alpha2 locally as done [here](https://github.com/containerd/containerd/pull/7656). In case we want to support v1alpha2 for longer period ?

mauriciovasquezbernal:
IMO we need to continue supporting v1alpha2 for a long period. If the solution you proposed above is not too difficult I'd suggest to implement this now to be able to update the k8s.io/cri-api version.

mqasimsarfraz:
> IMO we need to continue supporting v1alpha2 for a long period. If the solution you proposed above is not too difficult I'd suggest to implement this now to be able to update the k8s.io/cri-api version.

The proposed solution shouldn't be hard to implement. Makes sense to be able to update k8s.io/cri-api version. I reopened the issue to track it.

PR #1532, opened 2023-04-17
Add support for crun and podman

alban:
# Add support for crun and podman

* runcfanotify adds support for crun.
* runtime-client adds support for podman.

## Testing done

```
$ make ig && sudo ./ig list-containers -w
...
...
TIMESTAMP                      EVENT      RUNTIME    ID                                               NAME                                                                                                         
2023-05-02T17:47:41+02:00      CREATED    docker     ba4b260ec89864a1cd2de86dc0b918c04c9da7bc1d8139a… boring_ride                                                                                                  
2023-05-02T17:47:41+02:00      DELETED    docker     ba4b260ec89864a1cd2de86dc0b918c04c9da7bc1d8139a… boring_ride                                                                                                  
2023-05-02T17:47:54+02:00      CREATED    podman     dd781ba32f86a4240a0cc1197e8fb6dfb1557fc131c9393… sad_goldwasser                                                                                               
2023-05-02T17:47:54+02:00      DELETED    podman     dd781ba32f86a4240a0cc1197e8fb6dfb1557fc131c9393… sad_goldwasser                                                                                               
2023-05-02T17:48:04+02:00      CREATED    podman     941d3176bb6e9584a2b700a3a669f83de69dd69c950b453… wizardly_curie                                                                                               
2023-05-02T17:48:04+02:00      DELETED    podman     941d3176bb6e9584a2b700a3a669f83de69dd69c950b453… wizardly_curie                                                                                                                                
```

```
$ docker run -ti --rm busybox echo
$ podman --runtime=crun run -ti busybox echo
$ podman --runtime=runc run -ti busybox echo
```

## Notes

* It is not possible to call "podman inspect" when fanotify is blocking the runc/crun execution: this leads to a deadlock. To solve this, the runcfanotify package returns the additional field "ContainerName" found in the conmon command line so that runtime-client does not need to find it. runtime-client would only call "podman inspect" for initial containers, not for containers found via runcfanotify.

* The `ig list-containers` command has a field "RUNTIME". We should distinguish OCI Runtime (crun, runc) and high level runtime (docker, containerd, cri-o).

* I wanted to use the podman Go package but I didn't get it to work, so for this proof-of-concept, I use a workaround by executing an external command (e.g. `podman inspect ...`).

* I use `sudo -u alban` because by default podman uses rootless containers and `ig` runs as root, so they don't find the same containers. I am not sure what's the right way to do that. Maybe it will be solved when using the podman Go library?

## Updates:

> The ig list-containers command has a field "RUNTIME". We should distinguish OCI Runtime (crun, runc) and high level runtime (docker, containerd, cri-o). 

For now I used `io.container.manager` annotation to detect if the container is coming from `podman` and we can open a separate issue to track support for OCI Runtime in container events to make things simpler in this PR ? 

>> I wanted to use the podman Go package but I didn't get it to work, so for this proof-of-concept, I use a workaround by executing an external command (e.g. podman inspect ...).

> Switching to [podman bindings](https://github.com/containers/podman/tree/main/pkg/bindings) seems to be working fine but had to introduce [build tags](https://github.com/inspektor-gadget/inspektor-gadget/pull/1532/files#diff-c23efbf33144b7235d4b1139e17791167fb0de9ddadd72eab4a8cbfe54ee111eR98) to keep the build time dependencies simple. 

I used [podman restapi ](https://docs.podman.io/en/latest/_static/api.html) directly to keep things simple. It allows to use the same API as podman bindings but we can avoid introducing [build-tags](https://github.com/inspektor-gadget/inspektor-gadget/pull/1532/commits/781f4ee2531527f538d94b8c9ba0520e2068b99c) everywhere. Also, we can set timeouts for the `http.Client` to avoid running into deadlocks. 

> I use sudo -u alban because by default podman uses rootless containers and ig runs as root, so they don't find the same containers. I am not sure what's the right way to do that. Maybe it will be solved when using the podman Go library?

Introducing `podman-socketpath` helps supporting rootless containers by specifying socketPath `/run/user/$UID/podman/podman.sock` as:

```bash
# ensure user socket is present
$ systemctl start --user podman.socket
$ sudo ./ig list-containers --podman-socketpath /run/user/$UID/podman/podman.sock -r podman
```

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1515

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -138,13 +138,21 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
 			return err
 		}
 
-		// Add the enricher for future containers even if enriching the current
-		// containers fails. We do it because the runtime could be temporarily
-		// unavailable and once it is up, we will start receiving the
-		// notifications for its containers thus we will be able to enrich them.
-		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
-			return containerRuntimeEnricher(runtime.Name, runtimeClient, container)
-		})
+		switch runtime.Name {
+		case runtimeclient.PodmanName:
+			// Podman only supports runtime enrichment for initial containers otherwise it will deadlock.
+			// As a consequence, we need to ensure that new podman containers will be enriched with all
+			// the information via other enrichers e.g. see RuncNotifier.futureContainers implementation
+			// to see how container name is enriched.
+		default:
+			// Add the enricher for future containers even if enriching the current
+			// containers fails. We do it because the runtime could be temporarily
+			// unavailable and once it is up, we will start receiving the
+			// notifications for its containers thus we will be able to enrich them.
+			cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
+				return containerRuntimeEnricher(runtime.Name, runtimeClient, container)
+			})
+		}
 
 		cc.cleanUpFuncs = append(cc.cleanUpFuncs, func() {
 			if err := runtimeClient.Close(); err != nil {
@@ -539,6 +547,7 @@ func WithRuncFanotify() ContainerCollectionOption {
 					ID:        notif.ContainerID,
 					Pid:       notif.ContainerPID,
 					OciConfig: notif.ContainerConfig,
+					Name:      notif.ContainerName,
 				}
 				cc.AddContainer(container)
 			case runcfanotify.EventTypeRemoveContainer:
@@ -630,6 +639,10 @@ func WithOCIConfigEnrichment() ContainerCollectionOption {
 		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
 			if container.OciConfig == nil || container.IsEnriched() {
 				return true
+			}
+
+			if cm, ok := container.OciConfig.Annotations["io.container.manager"]; ok && cm == "libpod" {
+				container.Runtime = runtimeclient.PodmanName
 			}
 
 			resolver, err := ociannotations.NewResolverFromAnnotations(container.OciConfig.Annotations)
--- a/pkg/container-utils/podman/podman.go
+++ b/pkg/container-utils/podman/podman.go
@@ -1,0 +1,172 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package podman
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
+)
+
+const (
+	defaultConnectionTimeout = 2 * time.Second
+	containerListAllURL      = "http://d/v4.0.0/libpod/containers/json?all=true"
+	containerInspectURL      = "http://d/v4.0.0/libpod/containers/%s/json"
+)
+
+type PodmanClient struct {
+	client http.Client
+}
+
+func NewPodmanClient(socketPath string) runtimeclient.ContainerRuntimeClient {
+	if socketPath == "" {
+		socketPath = runtimeclient.PodmanDefaultSocketPath
+	}
+
+	return &PodmanClient{
+		client: http.Client{
+			Transport: &http.Transport{
+				DialContext: func(ctx context.Context, _, _ string) (conn net.Conn, err error) {
+					return net.Dial("unix", socketPath)
+				},
+			},
+			Timeout: defaultConnectionTimeout,
+		},
+	}
+}
+
+func (p *PodmanClient) listContainers(containerID string) ([]*runtimeclient.ContainerData, error) {
+	var filters string
+	if containerID != "" {
+		f, err := json.Marshal(map[string][]string{"id": {containerID}})
+		if err != nil {
+			return nil, fmt.Errorf("setting up filters: %w", err)
+		}
+		filters = "&filters=" + url.QueryEscape(string(f))
+	}
+
+	resp, err := p.client.Get(containerListAllURL + filters)
+	if err != nil {
+		return nil, fmt.Errorf("listing containers: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("listing containers via rest api: %s", resp.Status)
+	}
+
+	var containers []struct {
+		ID    string   `json:"Id"`
+		Names []string `json:"Names"`
+		State string   `json:"State"`
+	}
+	if err = json.NewDecoder(resp.Body).Decode(&containers); err != nil {
+		return nil, fmt.Errorf("decoding containers: %w", err)
+	}
+
+	ret := make([]*runtimeclient.ContainerData, len(containers))
+	for i, c := range containers {
+		ret[i] = &runtimeclient.ContainerData{
+			ID:      c.ID,
+			Name:    c.Names[0],
+			State:   containerStatusStateToRuntimeClientState(c.State),
+			Runtime: runtimeclient.PodmanName,
+		}
+	}
+	return ret, nil
+}
+
+func (p *PodmanClient) GetContainers() ([]*runtimeclient.ContainerData, error) {
+	return p.listContainers("")
+}
+
+func (p *PodmanClient) GetContainer(containerID string) (*runtimeclient.ContainerData, error) {
+	containers, err := p.listContainers(containerID)
+	if err != nil {
+		return nil, err
+	}
+	if len(containers) == 0 {
+		return nil, fmt.Errorf("container %q not found", containerID)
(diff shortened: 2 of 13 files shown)


pkg/container-utils/cri/cri.go
```diff
@@ -23,11 +23,12 @@ import (
 	"strings"
 	"time"
 
-	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
-	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	pb "k8s.io/cri-api/pkg/apis/runtime/v1"
```
alban, line 29: We will need to support both, see:
https://github.com/inspektor-gadget/inspektor-gadget/issues/1507

pkg/container-collection/options.go
```diff
@@ -146,6 +146,11 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
 			return containerRuntimeEnricher(runtime.Name, runtimeClient, container)
 		})
 
+		// skip enriching future containers for podman to avoid deadlocks
+		if runtime.Name == runtimeclient.PodmanName {
+			cc.containerEnrichers = nil
```
alban, line 151: I think you are removing all enrichers, including those unrelated to podman, such as WithTracerCollection(), WithLinuxNamespaceEnrichment().

mqasimsarfraz:
I added specifying custom socket-path for `podman` as well. Was able to see containers rootless containers as well:

```bash
# ensure user socket is present
$ systemctl start --user podman.socket
# root
→ sudo ./ig snapshot process -r podman
CONTAINER                                                                                                                                                                COMM             PID                  UID        GID       
goofy_aryabhata                                                                                                                                                          sleep            175092               0          0   
# rootless
→ sudo ./ig snapshot process -r podman --podman-socketpath /run/user/$UID/podman/podman.sock
CONTAINER                                                                                                                                                                COMM             PID                  UID        GID       
intelligent_goldwasser                                                                                                                                                   sleep            175203               1000       1000  
```

go.mod
```diff
@@ -150,3 +224,7 @@ replace (
 )
 
 replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.13.1-0.20230315234915-a26de2d610c3
+
+// Required by traceloop
+// https://github.com/cilium/ebpf/commit/5ea5368f9823c13e71146a6c2cca84ad60a5e26a
```
alban, line 229: Could add:
```
// Will be in future release cilium/ebpf v0.10.1
```

pkg/container-utils/podman/podman.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/containers/podman/v4/pkg/bindings"
+	"github.com/containers/podman/v4/pkg/bindings/containers"
+	log "github.com/sirupsen/logrus"
+
+	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
+)
+
+type PodmanClient struct {
+	conn       context.Context
+	socketPath string
```
alban, line 31: The field `socketPath` does not seem to be used. It could be removed.

alban (commented):
Some comments. I will review the rest after.

But it would be good if someone else review it too since I started the PR.

.github/workflows/inspektor-gadget.yml
```diff
@@ -530,7 +531,7 @@ jobs:
     - name: Build example binaries
       run: |
         for dir in $(dirname $(find examples/gadgets -name README.md)); do
-          go build ./$dir;
+          go build -tags "${{ env.BUILDTAGS }}" ./$dir;
```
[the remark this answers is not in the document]
mqasimsarfraz: Ah thanks for catching it. I will update the PR. I was thinking if there is a way to avoid introducing all these build tags and it seems [podman api](https://docs.podman.io/en/latest/_static/api.html) is simple:

```
$ curl --unix-socket /run/podman/podman.sock http://d/v4.0.0/libpod/containers/json
[{"AutoRemove":false,"Command":null,"Created":"2023-05-08T12:18:06.652224626+02:00","CreatedAt":"","Exited":false,"ExitedAt":-62135596800,"ExitCode":0,"Id":"79fc5967c6585b409443dc3a3bb82fcf6b25ef1ac33513eaaeae080fbbed6b14","Image":"localhost/podman-pause:4.5.0-0","ImageID":"24ba8382d076c5a49417dddef16eee88ac38dd5b9480f80a1e8fe6a1120c62d2","IsInfra":true,"Labels":{"io.buildah.version":"1.30.0"},"Mounts":[],"Names":["f77561f23a35-infra"],"Namespaces":{},"Networks":["podman"],"Pid":108601,"Pod":"f77561f23a354b93a2f43ee63f11b5f2d903274297b919da09edbf47af660632","PodName":"test","Ports":null,"Size":null,"StartedAt":1683541144,"State":"running","Status":""}]
...
```
Would need our own http client for list/inspect containers. What do you think about it?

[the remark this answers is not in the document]
alban: > Would need our own http client for list/inspect containers. What do you think about it?

I don't know what's best between the 2 approaches. I can accept either.

If we implement our own http client for this, we could implement a timeout in order to avoid accidental deadlock.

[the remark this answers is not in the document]
mqasimsarfraz: Makes sense. Using the API directly simplifies the changes and it isn't a lot of work. Also, I was able to test the timeout to avoid accidental deadlock.

pkg/container-collection/options.go
```diff
@@ -135,16 +135,26 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
...
-		// Add the enricher for future containers even if enriching the current
-		// containers fails. We do it because the runtime could be temporarily
-		// unavailable and once it is up, we will start receiving the
-		// notifications for its containers thus we will be able to enrich them.
-		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
-			return containerRuntimeEnricher(runtime.Name, runtimeClient, container)
-		})
+		switch runtime.Name {
+		case runtimeclient.PodmanName:
+			// Podman only supports runtime enrichment for initial containers otherwise it will deadlock
```
burak-ok, line 143: Is that the same issue that I had in #1233 ?

pkg/container-utils/podman/podman.go
```diff
@@ -16,100 +16,162 @@ package podman
...
 	if err != nil {
-		return nil, fmt.Errorf("error listing containers via podman: %w", err)
+		return nil, fmt.Errorf("listing containers: %w", err)
 	}
+	defer resp.Body.Close()
 
-	ret := make([]*runtimeclient.ContainerData, len(ctrs))
-	for i, c := range ctrs {
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("listing containers via api: %s", resp.Status)
```
burak-ok, line 75: nit: `via REST api:`

burak-ok (commented):
Didn't try it out, but it looks good to me

pkg/runcfanotify/runcfanotify.go
```diff
@@ -611,37 +766,21 @@ func (n *RuncNotifier) watchRuncIterate() (bool, error) {
...
 
-	if createFound && bundleDir != "" && pidFile != "" {
-		err := n.monitorRuncInstance(bundleDir, pidFile)
-		if err != nil {
-			log.Errorf("error monitoring runc instance: %v\n", err)
-		}
+	switch comm {
+	case "conmon":
+		// conmon is a special case because it is not a child of the container
+		n.parseConmonCmdline(cmdlineArr)
```
mauriciovasquezbernal, line 779: Perhaps I'm asking too much, but it'll be nice to have a short comment indicating how components call each other. Is it something like podman -> conmon -> crun?

mauriciovasquezbernal (commented):
Some nits and questions. From code inspection looks very good. I'll test it later on today.

mauriciovasquezbernal:
I was able to make it work after following the commands in https://github.com/inspektor-gadget/inspektor-gadget/pull/1532#issuecomment-1533362638 to enable the sock for user. 

I'm wondering what's the path forward to support rootless containers?

pkg/runcfanotify/runcfanotify.go
```diff
@@ -76,6 +80,13 @@ type runcContainer struct {
...
 }
 
+type futureContainer struct {
+	id        string
+	name      string
+	bundleDir string
+	pidFile   string
+}
+
 type RuncNotifier struct {
```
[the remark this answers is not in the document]
mqasimsarfraz: That is good question. Also this effects the package name `runcfanotify` as well. I know renaming might be nicer but might be breaking change for [examples](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/examples/container-collection/container-collection.go#L50) ?

cc @alban

mqasimsarfraz:
> I'm wondering what's the path forward to support rootless containers?

Currently the focus was to get the podman root support in and then look into the best way to support rootless containers in follow up PR. We need to knows `UID` for all the users running rootless containers so we can probably have a flag `podman-rootless-uids` and merge all the containers.

mauriciovasquezbernal (commented):
LGTM mostly. I think commits should be better organized, `podman: Use podman rest api directly ` reverts `Add support for crun and podman`, should those be squashed? 

I'll give a final review once those are organized.

pkg/runcfanotify/runcfanotify.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I talked internally with Alban and we agreed to move this PR forward. Later on we can think about renaming this component or even splitting it.

mauriciovasquezbernal (approved):
LGTM! I didn't test it again but I don't see major changes in the code. Great job!

blanquicet:
We are missing ig integration tests for podman as we have for docker, containerd and cri-o.

cc @mqasimsarfraz @mauriciovasquezbernal

PR #1538, opened 2023-04-18
cri: Support CRI API v1/v1alpha2 simultaneously

### LINKED ISSUE #1507: Support CRI `v1` and `v1alpha` at the same time

mqasimsarfraz:
containerd/crio will only support CRI API v1 in future releases so it makes sense to support v1/v1alpha2 simultaneously. This change will make v1 as the default version to talk to CRI and fallback to v1alpha2 in case v1 is unimplemented. It is mainly based on the implementation in [kubelet](https://github.com/kubernetes/kubernetes/commit/de37b9d293613aac194cf522561d19ee1829e87b).

Fixes #1507 

### older containerd version (v1alpha2 only)

```bash
$ make MINIKUBE_VERSION=v1.25.0 KUBERNETES_VERSION=v1.23.0 CONTAINER_RUNTIME=containerd minikube-start
# build/copy/verify ig
$ make ig && docker cp ig minikube-containerd:/bin/ig
$ docker exec -it minikube-containerd ig list-containers -r containerd -v
...
DEBU[0000] Using CRI v1alpha2 for containerd CRI client 
...
$ make MINIKUBE_VERSION=v1.25.0  minikube-clean
```

### newer containerd version (v1alpha2/v1)

```bash
$ make CONTAINER_RUNTIME=containerd minikube-start
# build/copy/verify ig
$ make ig && docker cp ig minikube-containerd:/bin/ig
$ docker exec -it minikube-containerd ig list-containers -r containerd -v
...
DEBU[0000] Using CRI v1 for containerd CRI client  
...
$ make minikube-clean
```

### DIFF
--- a/pkg/container-utils/cri/cri.go
+++ b/pkg/container-utils/cri/cri.go
@@ -23,11 +23,15 @@ import (
 	"strings"
 	"time"
 
-	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
-	pb "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	"google.golang.org/grpc/status"
+	runtime "k8s.io/cri-api/pkg/apis/runtime/v1"
+	runtimeV1alpha2 "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+
+	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 )
 
 // CRIClient implements the ContainerRuntimeClient interface using the CRI
@@ -37,8 +41,9 @@ type CRIClient struct {
 	SocketPath  string
 	ConnTimeout time.Duration
 
-	conn   *grpc.ClientConn
-	client pb.RuntimeServiceClient
+	conn           *grpc.ClientConn
+	client         runtime.RuntimeServiceClient
+	clientV1alpha2 runtimeV1alpha2.RuntimeServiceClient
 }
 
 func NewCRIClient(name, socketPath string, timeout time.Duration) (CRIClient, error) {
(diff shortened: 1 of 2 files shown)


pkg/container-utils/cri/cri.go
```diff
@@ -54,17 +59,46 @@ func NewCRIClient(name, socketPath string, timeout time.Duration) (CRIClient, er
...
+	if c.useV1alpha2() {
+		request := &runtimeV1alpha2.ListContainersRequest{}
+		if filter != nil {
+			request.Filter = &runtimeV1alpha2.ContainerFilter{
+				Id: filter.Id,
+			}
+		}
+		res, err := c.clientV1alpha2.ListContainers(context.Background(), request)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list containers with request %+v: %w",
```
mauriciovasquezbernal, line 94:
```suggestion
			return nil, fmt.Errorf("listing containers with request %+v: %w",
```

pkg/container-utils/cri/cri.go
```diff
@@ -54,17 +59,46 @@ func NewCRIClient(name, socketPath string, timeout time.Duration) (CRIClient, er
...
+
+	// determine CRI API version to use
+	if _, err = cc.client.Version(context.Background(), &runtime.VersionRequest{}); err == nil {
+		log.Debugf("Using CRI v1 for %s CRI client", name)
+	} else if status.Code(err) == codes.Unimplemented {
+		log.Debugf("Using CRI v1alpha2 for %s CRI client", name)
+		cc.client = nil
+		cc.clientV1alpha2 = runtimeV1alpha2.NewRuntimeServiceClient(conn)
+	} else {
+		log.Warnf("Failed to determine CRI API version: %v", err)
```
mauriciovasquezbernal, line 78:
```suggestion
		log.Warnf("Failed to determine CRI API version: %v. Using v1", err)
```

alban (approved):

pkg/container-utils/cri/cri.go
```diff
@@ -54,17 +59,46 @@ func NewCRIClient(name, socketPath string, timeout time.Duration) (CRIClient, er
...
 
-func listContainers(c *CRIClient, filter *pb.ContainerFilter) ([]*pb.Container, error) {
-	request := &pb.ListContainersRequest{}
+func listContainers(c *CRIClient, filter *runtime.ContainerFilter) ([]*runtime.Container, error) {
+	if c.useV1alpha2() {
+		request := &runtimeV1alpha2.ListContainersRequest{}
+		if filter != nil {
+			request.Filter = &runtimeV1alpha2.ContainerFilter{
+				Id: filter.Id,
+			}
```
mauriciovasquezbernal, line 90: I see that both versions define this structure in the same way 
https://github.com/kubernetes/cri-api/blob/3cbaf0d76da3429637d3eeb6de47045f50ebdc8b/pkg/apis/runtime/v1alpha2/api.pb.go#L4913 vs https://github.com/kubernetes/cri-api/blob/3cbaf0d76da3429637d3eeb6de47045f50ebdc8b/pkg/apis/runtime/v1/api.pb.go#L4948

Is it possible to define a function like `fromV1alpha2ListContainerFilter` to perform the conversion keep all the fields? (I think it can use the same exact approach as in `fromV1alpha2ListContainersResponse`)

mauriciovasquezbernal (commented):
Looks great! Some comments.

pkg/container-utils/cri/cri.go
mqasimsarfraz: Ah! Right. Thanks for spotting it. I see it also in [kubelet/convertion.go](https://github.com/kubernetes/kubernetes/blob/v1.25.9/pkg/kubelet/cri/remote/conversion.go#L108) as well, somehow I missed it. I updated the PR now

mauriciovasquezbernal (approved):
LGTM, thanks!

PR #1549, opened 2023-04-19
pkg/histogram: Create new package for all gadgets using histograms

blanquicet:
This PR moves the functions that allow generating and printing a histogram from the profile block-io gadget to an internal package. It also adds unit-tests for such package.

This new package will be used also by the upcoming gadget: [profile tcprtt](https://github.com/inspektor-gadget/inspektor-gadget/pull/1517).

### DIFF
--- a/pkg/histogram/histogram.go
+++ b/pkg/histogram/histogram.go
@@ -1,0 +1,129 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package histogram provides a Histogram struct that represents a histogram of
+// the number of events that occurred in each interval. It also provides a way
+// to transform a Histogram struct into a graphical representation. In addition,
+// it allows to create a Histogram struct from an exp-2 histogram.
+package histogram
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Unit string
+
+const (
+	UnitMilliseconds Unit = "ms"
+	UnitMicroseconds Unit = "µs"
+)
+
+type Interval struct {
+	Count uint64 `json:"count"`
+	Start uint64 `json:"start"`
+	End   uint64 `json:"end"`
+}
+
+// Histogram represents a histogram of the number of events that occurred in
+// each interval.
+type Histogram struct {
+	Unit      Unit       `json:"unit,omitempty"`
+	Intervals []Interval `json:"intervals,omitempty"`
+}
+
+// NewIntervalsFromExp2Slots creates a new Interval array from an exp-2
+// histogram represented in slots.
+func NewIntervalsFromExp2Slots(slots []uint32) []Interval {
+	if len(slots) == 0 {
+		return nil
+	}
+
+	intervals := make([]Interval, 0, len(slots))
+	indexMax := 0
+	for i, val := range slots {
+		if val > 0 {
+			indexMax = i
+		}
+
+		start := uint64(1) << i
+		end := 2*start - 1
+		if start == 1 {
+			start = 0
+		}
+
+		intervals = append(intervals, Interval{
+			Count: uint64(val),
+			Start: start,
+			End:   end,
+		})
+	}
+
+	// The element parsedIntervals[indexMax] is the last element with a non-zero
+	// value. So, we need to use parsedIntervals[:indexMax+1] to include it in
+	// the returned array.
+	return intervals[:indexMax+1]
+}
+
+// String returns a string representation of the histogram. It is a golang
+// adaption of iovisor/bcc print_log2_hist():
+// https://github.com/iovisor/bcc/blob/13b5563c11f7722a61a17c6ca0a1a387d2fa7788/libbpf-tools/trace_helpers.c#L895-L932
+func (h *Histogram) String() string {
+	if len(h.Intervals) == 0 {
+		return ""
+	}
+
+	valMax := uint64(0)
+	for _, b := range h.Intervals {
+		if b.Count > valMax {
+			valMax = b.Count
+		}
+	}
+
+	spaceBefore := 8
+	spaceAfter := 16
+	width := 10
+	stars := 40
+
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("%*s%-*s : count    distribution\n", spaceBefore,
+		"", spaceAfter, h.Unit))
+
+	for _, b := range h.Intervals {
+		sb.WriteString(fmt.Sprintf("%*d -> %-*d : %-8d |%s|\n", width,
+			b.Start, width, b.End, b.Count,
+			starsToString(b.Count, valMax, uint64(stars))))
+	}
+
+	return sb.String()
+}
+
+// starsToString returns a string with the number of stars and spaces needed to
+// represent the value in the histogram. It is a golang adaption of iovisor/bcc
+// print_stars():
+// https://github.com/iovisor/bcc/blob/13b5563c11f7722a61a17c6ca0a1a387d2fa7788/libbpf-tools/trace_helpers.c#L878-L893
+func starsToString(val, valMax, width uint64) string {
+	if valMax == 0 {
+		return strings.Repeat(" ", int(width))
+	}
+
+	stars := val * width / valMax
+	spaces := width - stars
+
+	var sb strings.Builder
+	sb.WriteString(strings.Repeat("*", int(stars)))
+	sb.WriteString(strings.Repeat(" ", int(spaces)))
+
+	return sb.String()
+}
--- a/pkg/histogram/histogram_test.go
+++ b/pkg/histogram/histogram_test.go
@@ -1,0 +1,232 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package histogram
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHistogram_NewIntervalsFromExp2Slots(t *testing.T) {
+	t.Parallel()
+
+	const unit = UnitMicroseconds
+
+	testTable := []struct {
+		description string
+		slots       []uint32
+		expected    *Histogram
+	}{
+		{
+			description: "Nil slots",
+			slots:       nil,
+			expected: &Histogram{
+				Unit:      unit,
+				Intervals: nil,
+			},
+		},
+		{
+			description: "Empty slots",
+			slots:       []uint32{},
+			expected: &Histogram{
+				Unit:      unit,
+				Intervals: nil,
+			},
+		},
+		{
+			description: "With 1 slot",
+			slots:       []uint32{333},
+			expected: &Histogram{
+				Unit: unit,
+				Intervals: []Interval{
+					{Count: 333, Start: 0, End: 1},
+				},
+			},
+		},
+		{
+			description: "With 2 slots",
+			slots:       []uint32{777, 555},
+			expected: &Histogram{
+				Unit: unit,
+				Intervals: []Interval{
+					{Count: 777, Start: 0, End: 1},
+					{Count: 555, Start: 2, End: 3},
+				},
+			},
+		},
+		{
+			description: "With zero slots",
+			slots:       []uint32{222, 0, 111},
+			expected: &Histogram{
+				Unit: unit,
+				Intervals: []Interval{
+					{Count: 222, Start: 0, End: 1},
+					{Count: 0, Start: 2, End: 3},
+					{Count: 111, Start: 4, End: 7},
+				},
+			},
+		},
+		{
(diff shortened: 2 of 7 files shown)


pkg/gadgets/internal/histogram/histogram.go
```diff
@@ -0,0 +1,146 @@
...
+	Slots []Slot `json:"slots,omitempty"`
+}
+
+// New creates a new Histogram.
+func New(unit string, slots []Slot) *Histogram {
+	return &Histogram{
+		Unit:  unit,
+		Slots: slots,
+	}
+}
```
mauriciovasquezbernal, line 47: Do we really need this function if Histogram is an exported type?

pkg/gadgets/internal/histogram/histogram_test.go
```diff
@@ -0,0 +1,179 @@
...
+			description: "Empty histogram",
+			histogram:   New("usecs", []Slot{}),
+			expected:    "",
+		},
+		{
+			description: "With 1 slot value 1",
+			histogram:   NewFromCSlots("usecs", []uint32{1}),
+			expected: "" +
+				"     usecs               : count    distribution\n" +
+				"         1 -> 1          : 1        |****************************************|\n",
```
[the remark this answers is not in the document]
blanquicet: I don't know. I know the concatenation of string doesn't look that great, but IMO, it makes things more organized and easy to create. It took me a while to match the spaces at the beginning of the line when using the backtick. I prefer keeping the current implementation 🙈.

pkg/gadgets/internal/histogram/histogram.go
blanquicet: Not really, but it looked weird to me to have `NewFromCSlots` and not `New` 😅. I'm ok removing it.

pkg/gadgets/internal/histogram/histogram.go
```diff
@@ -0,0 +1,146 @@
...
+	IntervalStart uint64 `json:"intervalStart"`
+	IntervalEnd   uint64 `json:"intervalEnd"`
+}
+
+// Histogram represents a histogram of the number of events that occurred in
+// each interval.
+type Histogram struct {
+	Unit  string `json:"unit,omitempty"`
+	Slots []Slot `json:"slots,omitempty"`
+}
```
blanquicet, line 39: I considered defining a `Unit` type with its possible values like this:

```suggestion
type Unit string

const (
	UnitMilliseconds Unit = "msecs"
	UnitMicroseconds Unit = "usecs"
)

// Histogram represents a histogram of the number of events that occurred in
// each interval.
type Histogram struct {
	Unit  Unit   `json:"unit,omitempty"`
	Slots []Slot `json:"slots,omitempty"`
}
```

It will simplify the caller's life by allowing this:

```golang
h := histogram.NewFromCSlots(histogram.UnitMicroseconds, []uint32{1}),
```

Instead of:

```golang
h := histogram.NewFromCSlots("usecs", []uint32{1}),
```

However, it'll require exporting this package in `pkg` to make it usable also by integration tests. WDYT?

cc @eiffel-fl @mauriciovasquezbernal

blanquicet: > I was wandering about the unit but so far it seems to only be printed. I mean, if we change the unit it does not have any impact on what is drawn (i.e. the underlying computation will not be done in another unit than µs), right?
>
The profile block-io gadget only supports microseconds, so this change is just a matter of using a variable instead of a literal string. Please notice that the bpf code is able to provide info in micro and milliseconds, but we are not using such a flag for this gadget.

> It could nonetheless be a good idea to introduce this new type as it would ease adding a PR to make the whole configurable if we want to use other time.
>
Exactly. I added support for micro and milliseconds in the profile tcprtt gadget that I'm working so this new type would help (also when we will add the same support to profile block-io).

> Nonetheless, rather than declaring our own type, would it be possible to use something from time golang package?
> 
I didn't find anything. The time package uses another notation: `µs`, and don't use constants, check [this](https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/time/time.go;l=644). Do you know any package that provides it?

pkg/gadgets/internal/histogram/histogram.go
mauriciovasquezbernal: A possibility is to make `parseCSlots` -> `NewSlots()` and remove `New` and `NewFromCSlots`. Then the user could create a histogram as:

```
myHist := &Histogram{
		Unit:  unit,
		Slots: NewSlots(cSlots),
	}
```

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+type Unit string
+
+const (
+	UnitMilliseconds Unit = "msecs"
+	UnitMicroseconds Unit = "usecs"
```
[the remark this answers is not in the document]
blanquicet: I'm not sure, "msecs" and "usecs" comes from BCC. But it is also true that we created a generic histogram representation at this point. I'm ok either way.

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+
+		parsedSlots = append(parsedSlots, Slot{
+			Count:         uint64(val),
+			IntervalStart: start,
+			IntervalEnd:   end,
+		})
+	}
+
+	// Slots are 0-indexed, so we need to increment indexMax to get the number
+	// of slots.
```
mauriciovasquezbernal, line 76: I'm not sure about this comment. I think this is because the expression `s[a:b]` will take the elements `s[a]`, ... , `s[b-1]`, so if we want to take up to the `s[x]` element we have to use `s[:x+1]`.  (Somehow this is related to the fact that arrays are 0-indexed but the comment is not straightforward for me)

pkg/histogram/histogram.go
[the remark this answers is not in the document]
mauriciovasquezbernal: IMO it's better to deviate from BCC and keep it aligned with other time representations that we have and use `time.Duration`

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+
+	parsedSlots := make([]Slot, 0, len(cSlots))
+	indexMax := 0
+	for i, val := range cSlots {
+		if val > 0 {
+			indexMax = i
+		}
+
+		start := uint64(1) << i
+		end := (uint64(1) << (i + 1)) - 1
```
mauriciovasquezbernal, line 63:
```suggestion
		start := uint64(1) << i
		end := 2*start - 1
```

Perhaps that helps a bit with code readability?

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+type Unit string
+
+const (
+	UnitMilliseconds Unit = "msecs"
+	UnitMicroseconds Unit = "usecs"
+)
+
+type Slot struct {
```
mauriciovasquezbernal, line 35: nit: I don't know if `Slot` is the better terminology here. It seems that `Bucket` is the standard. I also think `Interval` can work.

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package histogram provides a Histogram struct that represents a histogram of
+// the number of events that occurred in each interval. The intervals are powers
+// of two, so the slots array is indexed by the power of two minus one. For
+// example, the first slot is the number of events that occurred in the interval
+// [1, 1], the second slot is [2, 3], the third slot is [4, 7], and so on.
```
mauriciovasquezbernal, line 19: `[1, 1]` -> `[0, 1]` ? :exploding_head:

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+	if valMax == 0 {
+		return strings.Repeat(" ", int(width))
+	}
+
+	minVal := uint64(0)
+	if val < valMax {
+		minVal = val
+	} else {
+		minVal = valMax
+	}
```
mauriciovasquezbernal, line 129: Is this logic needed given that there is some logic to determine valMax in L88?

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+
+	stars := minVal * width / valMax
+	spaces := width - stars
+
+	var sb strings.Builder
+	sb.WriteString(strings.Repeat("*", int(stars)))
+	sb.WriteString(strings.Repeat(" ", int(spaces)))
+	if val > valMax {
+		sb.WriteByte('+')
+	}
```
mauriciovasquezbernal, line 139: ditto.

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+
+	valMax := uint64(0)
+	for _, slot := range h.Slots {
+		if slot.Count > valMax {
+			valMax = slot.Count
+		}
+	}
+
+	// reportEntries maximum value is C.MAX_SLOTS which is 27, so we take the
+	// value when idx_max <= 32.
```
mauriciovasquezbernal, line 96: Is this a comment from the original code that doesn't make sense now?

pkg/histogram/histogram.go
mauriciovasquezbernal, line 19: I think this histogram struct can be generic (linear, exp2, etc). This comment is more related to the `NewSlots` function.

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+// Histogram represents a histogram of the number of events that occurred in
+// each interval.
+type Histogram struct {
+	Unit  Unit   `json:"unit,omitempty"`
+	Slots []Slot `json:"slots,omitempty"`
+}
+
+// NewSlots creates a new Slot array from a C-Slots array returned by the BPF
+// program.
+func NewSlots(cSlots []uint32) []Slot {
```
mauriciovasquezbernal, line 50: As mentioned above, I think we need to make it clear that the `cSlots` correspond to a exp2 histogram. (We calculate start and end based on that assumption). Perhaps this can be renamed to `NewExp2Slots` ?

mauriciovasquezbernal (commented):
I didn't intended to review this into too much detail, but once I started I couldn't stop. Sorry if I added some many detail comments, I know your intention was only to move some code, but we can take the opportunity to improve it.

pkg/histogram/histogram.go
blanquicet: Yeah, the comment is not clear enough:

```suggestion
	// The element parsedSlots[indexMax] is the last element with a non-zero
	// value. So, we need to use parsedSlots[:indexMax+1] to include it in the
	// returned array.
```

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+	for i, val := range cSlots {
+		if val > 0 {
+			indexMax = i
+		}
+
+		start := uint64(1) << i
+		end := (uint64(1) << (i + 1)) - 1
+		if start == end {
+			start -= 1
+		}
```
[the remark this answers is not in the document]
blanquicet: Yep, it is only valid in that case.

pkg/histogram/histogram.go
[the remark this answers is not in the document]
blanquicet: It is fine for me. Just one thing. Should we use "us" or "µs"? The time package uses the Greek letter.

pkg/histogram/histogram.go
blanquicet: Yeah, [bucket](https://en.wikipedia.org/wiki/Data_binning) seems to be the standard.

pkg/histogram/histogram.go
mauriciovasquezbernal: Yes, it is.

pkg/histogram/histogram.go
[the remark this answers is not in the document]
mauriciovasquezbernal: `µs` should be fine.

pkg/histogram/histogram.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Should it be `start = 0` on your code?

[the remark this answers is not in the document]
blanquicet: Yep, fixed.

pkg/histogram/histogram.go
```diff
@@ -0,0 +1,142 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package histogram provides a Histogram struct that represents a histogram of
+// the number of events that occurred in each interval. The intervals are powers
+// of two, so the slots array is indexed by the power of two minus one. For
+// example, the first slot is the number of events that occurred in the interval
+// [1, 1], the second slot is [2, 3], the third slot is [4, 7], and so on.
+
```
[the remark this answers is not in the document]
blanquicet: It is not required so I removed it to avoid problems.

pkg/histogram/histogram.go
blanquicet: At least for me, it doesn't make sense now.

pkg/histogram/histogram.go
blanquicet: Yep, IMO, the computation of `stars` can be done directly with `val` as `val` will never be greater than `valMax`. It means `stars := val * width / valMax`.

pkg/histogram/histogram.go
blanquicet: Same, `val` will never be greater than `valMax`. Removed.

pkg/histogram/histogram_test.go
```diff
@@ -0,0 +1,224 @@
...
+		expected    *Histogram
+	}{
+		{
+			description: "Empty slots",
+			slots:       []uint32{},
+			expected: &Histogram{
+				Unit:      unit,
+				Intervals: nil,
+			},
+		},
```
mauriciovasquezbernal, line 40: nit: add a test with `slots: nil` to be sure it doesn't crash.

pkg/histogram/histogram_test.go
```diff
@@ -0,0 +1,224 @@
...
+		{
+			description: "With 1 slot",
+			slots:       []uint32{1},
```
mauriciovasquezbernal, line 43: nit: I'd use higher numbers than 1 in this case, to avoid those being messed up with indexes. (8888, 9999, ...)

mauriciovasquezbernal (approved):
LGTM!

PR #1554, opened 2023-04-20
Simplify `ig` build process

eiffel-fl:
Hi.

This PR simplifies the whole process to build `ig`.
Before, we would need to pull the `ig-builder` as we would rely on this image in another `Dockerfile` to actually build the image.
Nonetheless, this image was not available for `arm64` and even if we would have published a version for this architecture it would have needed some modifications as `aarch64-linux-gnu-gcc` is not available on `arm64`.

So, this commit modifies the `ig` image to make it `buildx` aware.
Note that, this image can be extended to `ln -s` the `ig` binary, so in a future where it would have sense people would be able to `docker run --rm -ti --privileged ig trace exec`.

We can go even further if the whole simplification of this to just have:

```Makefile
ig-$*:
        GOOS=$(parse $*) GOARCH=$(parse $*) go build \
		-ldflags "-X main.version=${VERSION} -extldflags '-static'" \
		-o ig-${GOOS}-${GOARCH} \
		github.com/inspektor-gadget/inspektor-gadget/cmd/ig
```

And it would just been called with:

```bash
# Build for host OS and architecture:
$ make ig
# Cross building:
$ CC=cross-compiler make ig-crossos-crossarch
```

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -204,6 +204,11 @@ jobs:
       uses: ./.github/actions/install-debian-packages
     - name: Build ${{ matrix.ig-target }}
       run: |
+        if [ "${{ matrix.ig-target}}" = 'ig-linux-arm64' ]; then
+          sudo apt-get update
+          sudo apt-get install qemu-user-static
+        fi
+
         make ${{ matrix.ig-target }}
 
         # Prepare assets for release and actions artifacts
--- a/Dockerfiles/ig.Dockerfile
+++ b/Dockerfiles/ig.Dockerfile
@@ -1,23 +1,42 @@
-FROM ghcr.io/inspektor-gadget/ig-builder:latest
+ARG BUILDER_IMAGE=golang:1.19
+ARG BASE_IMAGE=gcr.io/distroless/static-debian11
 
-WORKDIR /go/src/github.com/inspektor-gadget/inspektor-gadget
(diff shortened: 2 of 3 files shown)


mqasimsarfraz:
I believe we need to handle the case for [ig-tests.Dockerfile](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/Dockerfiles/ig-tests.Dockerfile#L1). The [reason](https://github.com/inspektor-gadget/inspektor-gadget/pull/1211) was to use same way to build ig and ig-tests. How do you think we should handle it? 

Having said that I agree we need to simplify the build process because I hit some issues when adding some dependencies in #1532

mqasimsarfraz:
> What type of issues? I would suggest to abstract them as much as possible, for example to not really worry regarding the tests in a first step.

We can ignore that now. I wanted to install `libgpgme-dev` for both `arm64/amd64` in `ig-builder` but was running into conflicts. But we don't need to do that anymore.

> I forgot this on purpose, basically because I wanted to get opinion first on how to simplify the whole building.

Makes sense. Let me play/review the changes. Currently we don't even push the arm image for ig-builder so would be nice to simplify/fix it!

burak-ok:
As @mqasimsarfraz mentioned we did that on purpose on another PR. But tbh I don't prefer any way over the other.

But if we want to accept this PR we should do the same cleanup for the `ig-test` Dockerfile

PR #1555, opened 2023-04-20
tree-wide: Rename UID fields to Uid

mauriciovasquezbernal:
The official go packages call it Uid rather than UID:
- https://github.com/golang/go/blob/bdccb85f509d24789694df198fe7bde948aa7955/src/syscall/ztypes_linux_amd64.go#L106
- https://github.com/golang/go/blob/bdccb85f509d24789694df198fe7bde948aa7955/src/os/user/user.go#L39

I also consider Uid is way better than UID, so I prefer to make this change now before it's impossible to make in the feature.

ref https://github.com/golang/lint/issues/89

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,10 @@ issues:
     - text: "ST1000: at least one file in a package should have a package comment"
       linters:
         - stylecheck
+    # Ignore rule about ID vs Id: https://github.com/golang/lint/issues/89
+    - text: "ST1003:"
+      linters:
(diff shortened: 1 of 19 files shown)


mauriciovasquezbernal:
`GID` and `UID` are added to a explicit list of initialisms in https://staticcheck.io/docs/configuration/options/#initialisms.

alban (approved):
I think it's good to observe the same practice as Goland.
LGTM

PR #1557, opened 2023-04-21
Add makefile target to check build requirements

eiffel-fl:
Hi.

This PR adds a `Makefile` target to check the host machine has everything needed to build `kubectl-gadget`, `ig` and the container images.

Best regards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 TAG := `git describe --tags --always`
 VERSION :=
 
+GOLANG_VERSION=go1.19
+
 CONTAINER_REPO ?= ghcr.io/inspektor-gadget/inspektor-gadget
 IMAGE_TAG ?= $(shell ./tools/image-tag branch)
 
@@ -41,6 +43,14 @@ ifeq ($(pvpath),)
 	PV :=
 else
 	PV := | $(pvpath)
+endif
+
+ifneq ($(shell go help > /dev/null; echo $$?), 0)
+$(warning "You need to have golang version $(GOLANG_VERSION) installed to build kubectl_gadget")
+endif
+
+ifneq ($(shell docker buildx --help > /dev/null; echo $$?), 0)
+$(warning 'You need to have docker and docker buildx installed to build ig and container images')
 endif
 
 # export variables that are used in Makefile.btfgen as well.
--- a/docs/devel/CONTRIBUTING.md
+++ b/docs/devel/CONTRIBUTING.md
@@ -38,7 +38,7 @@ kubectl-gadget-linux-amd64` or `make kubectl-gadget-darwin-amd64`.
 
 Inspektor Gadget provides two different container images:
 - gadget-default: Contains CO-RE and BCC gadgets
-- gadget-core: Containes only CO-RE gadgets that are integrated with the gadget tracer manager
+- gadget-core: Contains only CO-RE gadgets that are integrated with the gadget tracer manager
   * traceloop gadget is not included, it will be included once
     [#371](https://github.com/inspektor-gadget/inspektor-gadget/issues/371) is
     fixed.


Makefile
```diff
@@ -266,6 +268,11 @@ minikube-deploy: minikube-start gadget-default-container kubectl-gadget
 btfgen:
 	+make -f Makefile.btfgen
 
+.PHONY: check_build_requirements
+check_build_requirements:
+	@which go > /dev/null && go version | grep $(GOLANG_VERSION) || echo 'You need to have golang version $(GOLANG_VERSION) to build kubectl_gadget' >&2
+	@which docker > /dev/null && docker buildx --help > /dev/null || echo 'You need to have docker and docker buildx installed to build ig and container images' >&2
+
```
mauriciovasquezbernal, line 275: I'm not sure if this is worth it to check those things automatically. Perhaps we can only document them and that's it? (It could be more complicated to things versions of installed libraries and so on in case they're needed).

Makefile
```diff
@@ -266,6 +268,11 @@ minikube-deploy: minikube-start gadget-default-container kubectl-gadget
 btfgen:
 	+make -f Makefile.btfgen
 
+.PHONY: check_build_requirements
+check_build_requirements:
+	@which go > /dev/null && go version | grep $(GOLANG_VERSION) || echo 'You need to have golang version $(GOLANG_VERSION) to build kubectl_gadget' >&2
```
mauriciovasquezbernal, line 273: AFAIU that's the minimal golang version. For instance, the current code should compile fine with golang 1.20.

Makefile
alban: The check could be done outside of a rule so that it is always executed. This is already done for the `pv` command. We could do something like:

```
GOLANG_VERSION=1.19

ifneq (0, $(shell go version | grep -q $(GOLANG_VERSION) &>/dev/null; echo $$?))
$(error "go $(GOLANG_VERSION) not available")
endif

ifneq (0, $(shell docker buildx --help &>/dev/null; echo $$?))
$(error "docker buildx not available")
endif
```

claudiamarcubina:
> > In addition to the gcc and docker-buildx check, #1571 suggests to check for gcc and pkg-config. There is libseccomp-dev too, but a check is less trivial so we could leave it out.
> 
> Hum... I am not so sure about it, from my understanding #1571 seems to be about compiling directly using `go build`, while here the goal is to ensure people have everything needed to build using a `Makefile` target.

The intend for #1571 is also related to compiling by using the makefile targets.

mauriciovasquezbernal:
> build using a Makefile target

AFAIU this PR will only cover cases that are compiled within a container, but I see that there are a lot of targets in the Makefile that have other requirements. `generate`, `btfgen`, ..., what's the plan with them?

Makefile
mauriciovasquezbernal: > This is already done for the pv

IIUC the logic we have is to use the pv command if this is available. It's not a strict requirement.

I'm still not convinced about this kind of checks. Is there really an advantage of printing such warnings over letting the command fail:

```
$ make gadget-default-container
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
if false == "true" ; then \
	./tools/getbtfhub.sh && \
	make -f Makefile.btfgen BPFTOOL=/root/btfhub/tools/bin/bpftool. \
		BTFHUB_ARCHIVE=/root/btfhub-archive/ OUTPUT=hack/btfs/ -j; \
fi
docker buildx build --load -t ghcr.io/inspektor-gadget/inspektor-gadget:latest \
	-f Dockerfiles/gadget-default.Dockerfile .
make: docker: No such file or directory
make: *** [Makefile:142: gadget-default-container] Error 127
```

Also, I think those warnings are misleading as they'll be printed even if the target doesn't require them:

```
$ make btfgen
/bin/sh: 1: go: not found
Makefile:49: "You need to have golang version go1.19 installed to build kubectl_gadget"
/bin/sh: 1: docker: not found
Makefile:53: 'You need to have docker and docker buildx installed to build ig and container images'
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make: go: No such file or directory
make -f Makefile.btfgen
...
```

Issue #1425, opened 2023-03-13
Implement gadget to inspect tcp connect latency at the syscall level


mauriciovasquezbernal:
From #1199 

BCC has the tcpconnlat tool https://github.com/iovisor/bcc/blob/master/tools/tcpconnlat_example.txt, it's also available as libbpf-tool https://github.com/iovisor/bcc/blob/master/libbpf-tools/tcpconnlat.c.

PR #1559, opened 2023-04-21
gadgets: Implement trace tcpconnlat  gadget

mauriciovasquezbernal:
The trace tcpconnlat gadget tool traces the kernel function performing active TCP connections (eg,
via a connect() syscall), and shows the latency (time) for the connection as measured locally: the
time from SYN sent to the response packet.

TCP connection latency is a useful performance measure showing the time taken to establish a
connection. This typically involves kernel TCP/IP processing and the network round trip time, and
not application runtime.

trace tcpconnlat measures the time from any connection to the response packet, even if the response
is a RST (port closed).

Reuses the eBPF code from the BCC tcpconnlat tool:
- https://github.com/iovisor/bcc/blob/master/libbpf-tools/tcpconnlat.c
- https://github.com/iovisor/bcc/blob/master/tools/tcpconnlat_example.txt

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1425


pkg/gadgets/trace/tcpconnlat/tracer/tracer.go
```diff
@@ -0,0 +1,204 @@
...
+			Event: eventtypes.Event{
+				Type:      eventtypes.NORMAL,
+				Timestamp: gadgets.WallTimeFromBootTime(bpfEvent.Timestamp),
+			},
+			WithMountNsID: eventtypes.WithMountNsID{MountNsID: bpfEvent.MntnsId},
+			Pid:           bpfEvent.Tgid,
+			Comm:          gadgets.FromCString(bpfEvent.Comm[:]),
+			Dport:         gadgets.Htons(bpfEvent.Dport),
+			Sport:         bpfEvent.Lport,
+			Latency:       time.Duration(int64(bpfEvent.DeltaUs * 1000)),
```
alban, line 156: The latency is initially calculated in nanoseconds, then converted to microseconds, then back to nanoseconds here. Could we keep nanoseconds everywhere to avoid both conversions?

pkg/gadgets/trace/tcpconnlat/tracer/gadget.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (g *GadgetDesc) ParamDescs() params.ParamDescs {
+	return params.ParamDescs{
+		{
+			Key:          ParamMin,
+			Title:        "min",
+			DefaultValue: "0",
+			Description:  "Show only connect events with higher latency than min µs",
+			TypeHint:     params.TypeUint64,
```
alban, line 54: I am wondering about using `params.TypeString` and accepting a duration with the syntax described in:
https://pkg.go.dev/time#ParseDuration

That would be more friendly.

pkg/gadgets/trace/tcpconnlat/tracer/tracer.go
```diff
@@ -0,0 +1,204 @@
...
+
+		bpfEvent := (*tcpconnlatEvent)(unsafe.Pointer(&record.RawSample[0]))
+
+		event := types.Event{
+			Event: eventtypes.Event{
+				Type:      eventtypes.NORMAL,
+				Timestamp: gadgets.WallTimeFromBootTime(bpfEvent.Timestamp),
+			},
+			WithMountNsID: eventtypes.WithMountNsID{MountNsID: bpfEvent.MntnsId},
+			Pid:           bpfEvent.Tgid,
```
alban, line 152: Some gadgets have both Pid and Tid. Should we start having both for all gadgets?

pkg/gadgets/trace/tcpconnlat/tracer/bpf/tcpconnlat.bpf.c
```diff
@@ -0,0 +1,154 @@
...
+//int BPF_PROG(fentry_tcp_v6_connect, struct sock *sk)
+//{
+//	return trace_connect(sk);
+//}
+//
+//SEC("fentry/tcp_rcv_state_process")
+//int BPF_PROG(fentry_tcp_rcv_state_process, struct sock *sk)
+//{
+//	return handle_tcp_rcv_state_process(ctx, sk);
+//}
```
alban, line 152: To implement or to remove :)

alban (commented):
Thanks. This looks almost ready.

alban:
Could you add an entry for tcpconnlat in the following files?
- docs/requirements.md
- internal/benchmarks/benchmarks_test.go

docs/gadgets/trace/tcpconnlat.md
```diff
@@ -0,0 +1,116 @@
...
+...
+```
+
+The first terminal show all those connections and their latency. In my case both pods are running on
+the same node, so it's very low:
+
+```bash
+NODE             NAMESPACE        POD              CONTAINER        PID       COMM     IP SADDR           DADDR           SPORT DPORT LATENCY
+ubuntu-hirsute   default          myclientpod      myclientpod      363550    curl     4  172.16.118.147  10.109.132.206  47078 80    121µs
+ubuntu-hirsute   default          myclientpod      myclientpod      363654    curl     4  172.16.118.147  10.109.132.206  60028 80    116µs
```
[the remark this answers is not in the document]
mauriciovasquezbernal: What do you think about handling this together for all networking gadgets like tcptrace, tcpconnect and so on?

pkg/gadgets/trace/tcpconnlat/tracer/gadget.go
mauriciovasquezbernal: Added this support to the params library.

pkg/gadgets/trace/tcpconnlat/tracer/bpf/tcpconnlat.bpf.c
mauriciovasquezbernal: Added reference to https://github.com/inspektor-gadget/inspektor-gadget/issues/1566

pkg/gadgets/trace/tcpconnlat/tracer/bpf/tcpconnlat.bpf.c
```diff
@@ -0,0 +1,157 @@
...
+	__uint(max_entries, 1024);
+	__uint(key_size, sizeof(u64));
+	__uint(value_size, sizeof(u32));
+} mount_ns_filter SEC(".maps");
+
+static int trace_connect(struct sock *sk)
+{
+	u64 pid_tgid = bpf_get_current_pid_tgid();
+	u32 tgid = pid_tgid >> 32;
+	u32 pid = (pid_t)pid_tgid;
```
alban, line 51:
```suggestion
	u32 pid = (u32)pid_tgid;
```
(other gadgets do a simple cast)

docs/gadgets/trace/tcpconnlat.md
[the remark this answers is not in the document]
mauriciovasquezbernal: https://github.com/inspektor-gadget/inspektor-gadget/issues/1568 :)

pkg/gadgets/trace/tcpconnlat/tracer/bpf/tcpconnlat.bpf.c
```diff
@@ -0,0 +1,157 @@
...
+				__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
+		BPF_CORE_READ_INTO(&event.daddr_v6, sk,
+				__sk_common.skc_v6_daddr.in6_u.u6_addr32);
+	}
+	event.timestamp = bpf_ktime_get_boot_ns();
+	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
+			&event, sizeof(event));
+
+cleanup:
+	bpf_map_delete_elem(&start, &sk);
```
alban, line 116: It seems `tcp_rcv_synsent_state_process()` is only called when receiving a reply to the first SYN packet. What happens if the client never receives a reply and the connection is closed by timeout? Will the entries in the `start` eBPF map never be removed?

mauriciovasquezbernal: Good catch!!! I added an additional kprobe in `tcp_v{4,6}_destroy_sock` to fix it.

mauriciovasquezbernal: BCC upstream PR -> https://github.com/iovisor/bcc/pull/4588

pkg/gadgets/trace/tcpconnlat/tracer/bpf/tcpconnlat.bpf.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+SEC("kprobe/tcp_rcv_state_process")
+int BPF_KPROBE(ig_tcp_rsp, struct sock *sk)
+{
+	return handle_tcp_rcv_state_process(ctx, sk);
+}
+
+SEC("kprobe/tcp_v4_destroy_sock")
+int BPF_KPROBE(ig_tcp4_destroy, struct sock *sk)
```
alban, line 144: Instead of `tcp_v4_destroy_sock`, is it possible to use a kprobe on tcp_set_state() and check for `state == TCP_CLOSE`?
That would reduce the number of kprobes.

Another reason is to reduce the possibility of a malicious (or buggy) container from connecting many times to an unreachable endpoint but keeping the file descriptors open. If we only clean up on `tcp_v4_destroy_sock`, the eBPF map will be full and the tcpconnlat will stop working even for independent containers. If the cleanup is done after the connection timeout, that possibility is somewhat reduced.

mauriciovasquezbernal:
After talking with @alban, we agreed that it's better to integrate this functionality into the tcpconnect gadget as they are very similar. I opened a new PR for that https://github.com/inspektor-gadget/inspektor-gadget/pull/1574

PR #1562, opened 2023-04-24
tcpdrop: get netns from struct sock

alban:
The tracepoint `skb/kfree_skb` gives a `struct sk_buff *skb`.

Before this patch, we get the network namespace via:
`skb->dev->nd_net.net->ns.inum`

This is unreliable because 'dev' is part of an union:
```
struct sk_buff {
        union {
                struct {
                        struct sk_buff          *next;
                        struct sk_buff          *prev;
                        union {
                                struct net_device       *dev;
                                ...
```
Depending on where the `sk_buff` is dropped, the 'dev' pointer might or might not be set correctly. In some code paths, it could be NULL or other fields of the unions might be in use instead.

This patch fetches the network namespace with the following instead:
`skb->sk->__sk_common.skc_net.net->ns.inum`


mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #1569, opened 2023-04-24
Test non-k8s containerd API

burak-ok:
# Test non-k8s containerd API

This will be merged into a branch that is an open PR #1233 

The `non-k8s` intergration tests for `ig` were restructured to generalize around the runtime.

### DIFF
--- a/integration/container_interface.go
+++ b/integration/container_interface.go
@@ -1,0 +1,88 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"fmt"
+	"testing"
+
+	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/testutils"
+)
+
+type ContainerSpec struct {
+	Name         string
+	Cmd          string
+	Options      []containerOption
+	Cleanup      bool
+	StartAndStop bool
+}
+
+type ContainerFactory interface {
+	NewContainer(ContainerSpec) ContainerInterface
+}
+
+type ContainerInterface interface {
+	Run(t *testing.T)
+	Start(t *testing.T)
+	Stop(t *testing.T)
+	IsCleanup() bool
+	IsStartAndStop() bool
+	Running() bool
+}
+
+func NewContainerFactory(containerRuntime string) (ContainerFactory, error) {
+	switch containerRuntime {
+	case runtimeclient.DockerName:
+		return &DockerManager{}, nil
+	case runtimeclient.ContainerdName:
+		return &ContainerdManager{}, nil
+	default:
+		return nil, fmt.Errorf("unknown container runtime %q", containerRuntime)
+	}
+}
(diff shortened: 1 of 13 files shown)


pkg/container-utils/testutils/options.go
```diff
@@ -52,6 +55,12 @@ func WithImage(image string) Option {
 	}
 }
 
+func WithImageTag(tag string) Option {
```
mauriciovasquezbernal, line 58: What is this useful for? Can't we pass the tag within the image name instead (`WithImage("foo:latest")`)?

integration/container_interface.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2022 The Inspektor Gadget authors
```
mauriciovasquezbernal, line 1: nit: 2022 -> 2023 (in other files too)

mauriciovasquezbernal (approved):
I did a very quick review. Everything seems to be working fine. If nobody else wants to take a deeper look, I'll propose to merge as it is.

go.mod
```diff
@@ -147,3 +169,8 @@ replace (
 	github.com/gogo/protobuf v1.3.0 => github.com/gogo/protobuf v1.3.2
 	github.com/gogo/protobuf v1.3.1 => github.com/gogo/protobuf v1.3.2
 )
+
+// Required by ??? TODO
+replace k8s.io/cri-api => k8s.io/cri-api v0.25.8
```
mqasimsarfraz, line 174: This shouldn't be needed anymore after we update to [latest version](https://github.com/inspektor-gadget/inspektor-gadget/pull/1538). Does `containerd` complains about it?

go.mod
```diff
@@ -147,3 +169,8 @@ replace (
...
+
+replace github.com/cilium/ebpf => github.com/cilium/ebpf v0.0.0-20230320181651-5ea5368f9823
```
mqasimsarfraz, line 176: Please add a comment like:

```
// Required by traceloop, Will be in future release cilium/ebpf v0.10.1
// https://github.com/cilium/ebpf/commit/5ea5368f9823c13e71146a6c2cca84ad60a5e26a
```

cc @alban

integration/container_interface.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func NewContainerFactory(runtime string) ContainerFactory {
+	switch runtime {
+	case "docker":
+		return &DockerManager{}
+	case "containerd":
+		return &ContainerdManager{}
+	default:
+		panic("Unknown runtime " + runtime)
```
mqasimsarfraz, line 51: Since this is a test util do you think it makes sense to use [runtime.Goexit](https://pkg.go.dev/runtime#Goexit) here ?

integration/ig/non-k8s/integration_test.go
```diff
@@ -39,6 +42,8 @@ func TestMain(m *testing.M) {
 		os.Exit(0)
 	}
 
+	containerFactory = NewContainerFactory(*runtime)
```
mqasimsarfraz, line 45: Should we explicity add check for [supported runtime](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/ig/k8s/integration_test.go#L59) or you want it to panic ?

mqasimsarfraz (approved):
LGTM! Some minor comments but I am fine if we address them in the PR going to `main`

Also, I tested the changes as well!

integration/container_interface.go
burak-ok: I would like to have a message somewhere that the runtime given is unknown. `Goexit` doesn't take a string. I also looked into having a `testing.T.Errorf` but the time where we call `NewContainerFactory` doesn't allow that? We only have a `testing.M` instance there.
Maybe just `fmt.Println`?

pkg/container-utils/testutils/options.go
burak-ok: Intent was an easy way to switch tags without having to deconstruct the image string

go.mod
burak-ok: Not needed anymore after a rebase

go.mod
burak-ok: Added it in #1233

PR #1571, opened 2023-04-25
docs: add unit tests requirements

claudiamarcubina:
# Add minimal build requirements for compiling from source

I suggest adding these minimal build requirements for compiling from source for avoiding spending too much time debugging 
 in case one of the requirements is missing. It was the case for me when I did an upgrade to golang 1.20.3 and tried to run the tests. The error message wasn't clearly stating that gcc was missing from my system. 

**LE:** Realized that the these requirements were actually needed for the unit tests and moved them to contributing.

### DIFF
--- a/docs/devel/CONTRIBUTING.md
+++ b/docs/devel/CONTRIBUTING.md
@@ -109,6 +109,12 @@ registry.
 
 ### Unit tests
 
+#### Requirements
+
+For running unit tests, the following additional requirements need to be installed and configured on your system: 
+- gcc compiler
+- `pkg-config` and `libseccomp-dev` libraries
+
 You can run the different unit tests with:
 
 ```bash


docs/install.md
```diff
@@ -55,9 +56,16 @@ $ kubectl gadget version
 ```
 
 ### Compile from source
+#### Build requirements
 
-To build Inspektor Gadget from source, you'll need to have a Golang version
-1.19 or higher installed.
+To build Inspektor Gadget from source, the following requirements need to be installed and configured on your system:
+- Golang version 1.19 or higher
+- Docker Engine
```
alban, line 63:
```suggestion
- Docker or compatible
```

docs/install.md
```diff
@@ -55,9 +56,16 @@ $ kubectl gadget version
...
+- gcc compiler
+- make
```
alban, line 65: Did you also need the following?

- pkg-config
- libseccomp-dev

docs/install.md
```diff
@@ -55,9 +56,16 @@ $ kubectl gadget version
...
+
+#### Compile
+For compiling, clone this repository within the following directory structure: `go/src/github.com/inspektor-gadget/`:
```
alban, line 68: I suggest adding the explanation in commands:
```
mkdir -p ~/go/src/github.com/inspektor-gadget/
cd ~/go/src/github.com/inspektor-gadget/
git clone ...
```

docs/install.md
claudiamarcubina: Yees, missed these ones. Added now. Thanks for checking.

docs/install.md
claudiamarcubina: I also like this idea better. Added ✅

mauriciovasquezbernal:
@claudiamarcubina could you please clarify what and how were you trying to compile? I agree that most of the dependencies are not needed if directly using "make" because the whole compilation process happens inside Docker.

claudiamarcubina:
> The intention is good but I am not sure to understand the final goal.
> Does this cover when we want to build without using any provided `Dockerfile`?



> @claudiamarcubina could you please clarify what and how were you trying to compile? I agree that most of the dependencies are not needed if directly using "make" because the whole compilation process happens inside Docker.

On it already. Right now I am trying to reproduce locally the same workflow where I needed these dependencies. I'll get back soon.

claudiamarcubina:
> > > @claudiamarcubina could you please clarify what and how were you trying to compile? I agree that most of the dependencies are not needed if directly using "make" because the whole compilation process happens inside Docker.
> > 
> > 
> > On it already. Right now I am trying to reproduce locally the same workflow where I needed these dependencies. I'll get back soon.
> 
> Aren't those needed when you want to run `unit-tests`? Then, I would rather add them to `CONTRIBUTING.md` instead of `README.md`.

You're right, I came across the issues related to `gcc`, `libseccomp` and `pkg-config` when running unit tests.

I was aware this may not be the right place to mention these dependencies, but wanted to start the discussion. I think it would be helpful to have them mentioned somewhere. I agree that `CONTRIBUTING.md` may be more appropriate.

docs/install.md
```diff
@@ -66,9 +66,11 @@ To build Inspektor Gadget from source, the following requirements need to be ins
 - `pkg-config` and `libseccomp-dev` libraries
 
 #### Compile
-For compiling, clone this repository within the following directory structure: `go/src/github.com/inspektor-gadget/`:
+For compiling, follow the next steps:
 
 ```bash
+$ mkdir -p ~/go/src/github.com/inspektor-gadget/
+$ cd ~/go/src/github.com/inspektor-gadget/
```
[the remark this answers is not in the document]
claudiamarcubina: Hm, indeed. Not needed for compiling or testing.

docs/install.md
```diff
@@ -63,6 +63,7 @@ To build Inspektor Gadget from source, the following requirements need to be ins
 - Docker or compatible
 - gcc compiler
 - make
+- `pkg-config` and `libseccomp-dev` libraries
```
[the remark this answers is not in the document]
claudiamarcubina: Yes, `gcc`, `pkg-config` and `libseccomp-dev` are needed when testing. Will move them to contributing.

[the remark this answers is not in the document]
claudiamarcubina: Moved. Please check :)

claudiamarcubina:
> > > > @claudiamarcubina could you please clarify what and how were you trying to compile? I agree that most of the dependencies are not needed if directly using "make" because the whole compilation process happens inside Docker.
> > > 
> > > 
> > > On it already. Right now I am trying to reproduce locally the same workflow where I needed these dependencies. I'll get back soon.
> > 
> > 
> > Aren't those needed when you want to run `unit-tests`? Then, I would rather add them to `CONTRIBUTING.md` instead of `README.md`.
> 
> You're right, I came across the issues related to `gcc`, `libseccomp` and `pkg-config` when running unit tests.
> 
> I was aware this may not be the right place to mention these dependencies, but wanted to start the discussion. I think it would be helpful to have them mentioned somewhere. I agree that `CONTRIBUTING.md` may be more appropriate.

Moved them to `CONTRIBUTING.md`. Please have a look :)

Issue #1573, opened 2023-04-25
socket enricher in tcpdrop stops working server side after 1 connection


alban:
## Description

The socket enricher stops working on TCP, server-side after one connection is closed.

## Impact

The socket enricher appears unreliable.

## Environment and steps to reproduce

Use Minikube with the Docker driver. This will make it easier to run the bpftool and tc commands on the Kubernetes worker node because they can just be run from the host.

This is inspired from the guide to use tcpdrop on Kubernetes. Follow the guide to install the nginx service.
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadgets/trace/tcpdrop.md#on-kubernetes

Make sure the nginx pod is started after the tcpdrop gadget. If necessary, restarts the nginx pod with:
```
$ kubectl delete pod -l app=nginx --force
```

Just after the nginx pod has been started, notice that it listens to 2 sockets (port 80 on IPv4 and port 80 on IPv6):
```
$ sudo bpftool map dump name sockets|grep nginx
            "task": "nginx"
            "task": "nginx"
```

Install the network emulator in the nginx pod:
```
sudo nsenter -n -t $PID_NGINX tc qdisc add dev eth0 root netem drop 25%
```

Run curl from the `shell` pod:
```
root@shell:/# curl nginx
```
Notice that the sockets ebpf map lost one of its entry. This is the bug:
```
$ sudo bpftool map dump name sockets|grep nginx
            "task": "nginx"
```

Try the curl command again. This time, the socket enricher will not work properly for the nginx process:
```
minikube-docker             default                     nginx-8f458dc5b-l8ckq       nginx                       0                          4  p/default/nginx-8f458dc5b- p/default/busyshell:46308  ESTABLISHED  PSH|ACK        QDISC_DROP    
minikube-docker             default                     nginx-8f458dc5b-l8ckq       nginx                       0                          4  p/default/nginx-8f458dc5b- p/default/busyshell:46314  ESTABLISHED  PSH|ACK        QDISC_DROP    
```

## Expected behavior

* The sockets eBPF map does not lose one entry after a first connection
* The socket enricher continues to enrich nginx (TCP server side) correctly after one connection

## Additional information

nginx creates 2 passive sockets to listen on port 80 for IPv4 and IPv6. But after one connection, the accept() syscall returns an active socket with the same source port. When that active socket is released, the entry is removed from the sockets ebpf map.

[...]

PR #1576, opened 2023-04-26
socket enricher: distinguish active and passive sockets

alban:
The accept() syscall on a passive socket returns an active socket with the same source port. When that active socket is released, the entry must not be removed from the sockets ebpf map.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1573


burak-ok (approved):
I guess that this now fixes `dup/dup2` duplicated sockets too. Nice catch. LGTM

Future question: What happens in the following scenario:
1. accepting socket listens on socket `X`
2. client connects and accepting socket returns new socket `Y`
3. listening socket `X` now gets closed
4. socket `Y` is still in usage. socketenricher now doesn;t know anything about the socket

Can this situation occur?

alban:
@burak-ok I didn't think about the scenario where the listening socket is closed early. You're right, the socket enricher would not work server-side in that scenario. Usually it will not happen like this (e.g. nginx and other servers would not close the listening socket) but it can happens with `nc`, depending on the options. I suggest we don't have to fix this now.

Did you have a specific issue with dup/dup2? The socket enricher is working with the underlying sockets objects, so it should not care if a file descriptor is dup()ed because the underlying socket stays the same.

burak-ok:
> Usually it will not happen like this (e.g. nginx and other servers would not close the listening socket) but it can happens with `nc`, depending on the options. I suggest we don't have to fix this now.

Yep, that is not the normal situation with listening sockets so i also see this as a non/small issue

> Did you have a specific issue with dup/dup2? 

No, I didn't test anything. Was just thinking about other situations where a socket could refer to the same underlying `socket`. But the condition you added resolves this hypothetical issue

Issue #1572, opened 2023-04-25
trace tcpconnect stops working due to ebpf map leak


alban:
## Description

The `sockets` eBPF map is not clean up in exit_tcp_connect() when the filter by mount namespace skips the connection.

## Impact

The tcpconnect gadget stops working after MAX_ENTRIES=8192 such connections

## Environment and steps to reproduce

Start the tcpconnect gadget in terminal 1:
```
$ sudo ./ig trace tcpconnect -c mycontainer -r docker
```

Exhaust the eBPF map in terminal 2: (note that the container name is different than the one given to the tcpconnect gadget, so this is not generating any event)
```
$ docker run -d --name=nginx nginx
$ docker exec -ti nginx bash -c 'for i in `seq 8200` ; do echo -n " $i " ; curl localhost &>/dev/null ; done'
```

Try to generate events from terminal 3:
```
$ docker run --rm -ti --name=mycontainer busybox
/ # nc -l -p 8080 &
/ # echo OK|nc 127.0.0.1 8080
OK
```

## Expected behavior

The event is detected correctly, such as:

```
$ sudo ./ig trace tcpconnect -c mycontainer -r docker
CONTAINER    PID      COMM  IP SADDR      DADDR      DPORT
mycontainer  2541616  nc    4  127.0.0.1  127.0.0.1  8080
```

## Additional information

tcpconnect.bpf.c: exit_tcp_connect():
```
	if (filter_by_mnt_ns && !bpf_map_lookup_elem(&mount_ns_filter, &mntns_id))
		return 0;
```

PR #1579, opened 2023-04-26
trace/tcpconnect: Fix leakage of entries in eBPF map

mauriciovasquezbernal:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1572


mauriciovasquezbernal:
Holding as I'm proposing a more complete solution in https://github.com/inspektor-gadget/inspektor-gadget/pull/1580. If the discussion on that PR takes long then we can move ahead with this one.

alban:
I agree, the solution in #1580 is more elegant.

Issue #1578, opened 2023-04-26
gadgets: Enable/disable columns according to flags 


mauriciovasquezbernal:
There are some cases in which the columns we want to print are dependent on the flags passed to the gadget:
- snapshot process --threads  -> enable the TID column (https://github.com/inspektor-gadget/inspektor-gadget/pull/1396#issuecomment-1489250861)
- trace tcpconnect --latency -> enable the latency column (https://github.com/inspektor-gadget/inspektor-gadget/pull/1574#discussion_r1177659199)

The purpose of this issue is to implement this support to bring a better user experience. 

### Implementation options

I think this is a gadget specific logic, so each gadget should decide which columns to enable / disable based on this configuration. One option is to pass the gadget parameters to `Parser()` to let each gadget hide / show columns. 

### Questions

Should it be allowed to change the gadget config based on the set columns. For instance, `snapshot process -o columns=container,pid,tid` should implicitly enable the `--threads` flag?

/cc @flyth

flyth:
Regarding the implementation for the first part: as long as we're only enabling features (and therefore columns by bool params), we could simplify and generalize it by adding for example a column tag:

```golang
type Event {
...
	Latency   time.Duration `json:"latency,omitempty" column:"latency,minWidth:8" columnTags:"param:latency"`
...
}
```

Then we could reduce the default columns automatically when a given param is not true. Something like [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/23e458bf859412b06d9252a5799e65eb80f409ab) could do it without having to add code to gadgets. However, this only works for bool params (we could add a negation easily, like `param:!latency` and maybe even check for certain values like `param:mode=abc`, but if we would need more complicated rules, then it would probably be better having dedicated functions inside the gadgets).

Regarding the second part: I'm not really sure whether specifying columns should control feature flags. I'll think more about it.

mauriciovasquezbernal:
I think we can implement a very simple logic by using the `param` tag you suggested. It later on we find out that more complex cases are needed we can rethink it.

flyth:
Solved in #1587

PR #1587, opened 2023-04-27
Allow hiding columns based on selected features / parameters

flyth:
Right now, the `snapshot/process` gadget has the parameter `--threads`, which, if enabled, would allow displaying a new column showing `TID`. This column however doesn't get shown automatically right now, if said parameter is `true`. Instead, the user would have to specify the column explicitly by using `-o columns=...,tid`.

This PR allows adding `columnTags` in the form of `param:threads` to columns. These tags then will be compared to boolean parameters available for a gadget (like said `--threads`) - and if those parameters are falsy, columns with corresponding tags will be hidden.

To improve UX, required parameters will now also be shown in the `--help` for the columns:

```
  -o, --output string          Output format (jsonpretty, yaml, tree, json, columns).

                               Columns (columns)
                                   Available columns:
                                     ...
                                     ppid
                                     tid (enabled when using --threads)
                                     ...
```

To be able to access the tag information in a generic way (NOT using generics), I've moved out all column attributes to a separate non-generic struct that can easily be passed around.

Solves #1578

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -310,17 +310,36 @@ func buildCommandFromGadget(
 			}
 
 			formatter := parser.GetTextColumnsFormatter()
-			if outputModeParams != "" {
-				valid, invalid := parser.VerifyColumnNames(strings.Split(outputModeParams, ","))
 
-				for _, c := range invalid {
-					log.Warnf("column %q not found", c)
-				}
+			requestedStandardColumns := outputModeParams == ""
+			requestedColumns := strings.Split(outputModeParams, ",")
 
-				if err := formatter.SetShowColumns(valid); err != nil {
-					return err
+			// If the standard columns are requested, hide columns that would be empty without specific features
+			// (bool params) enabled
+			if requestedStandardColumns {
+				var hiddenTags []string
+				if gadgetParams != nil {
+					for _, param := range *gadgetParams {
+						if param.TypeHint == params.TypeBool {
+							if !param.AsBool() {
+								hiddenTags = append(hiddenTags, "param:"+strings.ToLower(param.Key))
+							}
+						}
+					}
 				}
+				requestedColumns = parser.GetDefaultColumns(hiddenTags...)
 			}
+
+			valid, invalid := parser.VerifyColumnNames(requestedColumns)
+
+			for _, c := range invalid {
+				log.Warnf("column %q not found", c)
+			}
+
+			if err := formatter.SetShowColumns(valid); err != nil {
+				return err
+			}
+
 			parser.SetLogCallback(fe.Logf)
 
 			// Wire up callbacks before handing over to runtime depending on the output mode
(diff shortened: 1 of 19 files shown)


cmd/common/registry.go
```diff
@@ -310,17 +310,37 @@ func buildCommandFromGadget(
 			}
 
 			formatter := parser.GetTextColumnsFormatter()
-			if outputModeParams != "" {
-				valid, invalid := parser.VerifyColumnNames(strings.Split(outputModeParams, ","))
 
-				for _, c := range invalid {
-					log.Warnf("column %q not found", c)
+			requestedStandardColumns := outputModeParams == ""
```
mauriciovasquezbernal, line 314: I tried to make it work for the tcpconnect gadget and I wasn't able. It turns out that the following logic is wrong:

https://github.com/inspektor-gadget/inspektor-gadget/blob/e2c674b962e71b08bb264a8df0eda389dbbd2306/cmd/common/registry.go#L486-L490

It'll change the value of `defaultOutputFormat` from  `columns=foo,bar,zas` to `columns` if the gadget declares that as default format. In other words, this `outputModeParams == ""` check is only valid for the `snapshot process` as it implements `OutputFormats` and returns `columns` ->  https://github.com/inspektor-gadget/inspektor-gadget/blob/b9eea85c9d1ed8ee8f7231c7a84a6950e00846e0/pkg/gadgets/snapshot/process/tracer/gadget.go#L90

I suppose this should be something like `outputModeParams == strings.Join(parser.GetDefaultColumns(), ",")`, but we need to be sure parser is not nil, etc...

mauriciovasquezbernal (commented):
Looks very good, some comments around.

What about squashing first and second commit to be sure it compiles fine?

pkg/parser/parser.go
```diff
@@ -42,8 +42,8 @@ type Parser interface {
 	// GetTextColumnsFormatter returns the default formatter for this columns instance
 	GetTextColumnsFormatter(options ...textcolumns.Option) TextColumnsFormatter
 
-	// GetColumnNamesAndDescription returns a map of column names to their respective descriptions
-	GetColumnNamesAndDescription() map[string]string
+	// GetColumnNamesAndAttributes returns a map of column names to their respective attributes
+	GetColumnNamesAndAttributes() map[string]columns.Attributes
```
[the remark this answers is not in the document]
flyth: Agree. And it might also be better to just return an array, so it'll also keep the order (the name can be accessed from the attributes).

cmd/common/registry.go
flyth: Ahh, I had it that way first, but when I tested it, the outputModeParams were empty, so I changed it 😆. Will fix it and do some more tests. Thanks!

flyth: I've changed the default to be empty now - that way we can differentiate between the value being set explicitly by the user or not. Otherwise we would have also adjusted, if the user specified exactly the default columns.

mauriciovasquezbernal (approved):
LGTM. Please rebase and fix also the tcpconnect gadget:


```diff
diff --git pkg/gadgets/trace/tcpconnect/types/types.go pkg/gadgets/trace/tcpconnect/types/types.go
index 0f07bf42..b7f69ae2 100644
--- pkg/gadgets/trace/tcpconnect/types/types.go
+++ pkg/gadgets/trace/tcpconnect/types/types.go
@@ -33,7 +33,7 @@ type Event struct {
        Daddr     string        `json:"daddr,omitempty" column:"daddr,template:ipaddr"`
        Sport     uint16        `json:"sport,omitempty" column:"sport,template:ipport"`
        Dport     uint16        `json:"dport,omitempty" column:"dport,template:ipport"`
-       Latency   time.Duration `json:"latency,omitempty" column:"latency,minWidth:8"`
+       Latency   time.Duration `json:"latency,omitempty" column:"latency,minWidth:8" columnTags:"param:latency"`
 }
 
 func GetColumns() *columns.Columns[Event] {
```

Thanks!

flyth:
Thanks for the updated review!

PR #1588, opened 2023-04-27
Check version at runtime for `profile block-io`

eiffel-fl:
Hi.

In this PR, I use `LINUX_KERNEL_VERSION` to avoid having two eBPF object files.

Best regards.


pkg/gadgets/profile/block-io/tracer/tracer.go
```diff
@@ -119,19 +105,19 @@ func (t *Tracer) install() error {
 
 	blockRqCompleteLink, err := link.AttachRawTracepoint(link.RawTracepointOptions{Name: "block_rq_complete", Program: t.objs.IgProfioDone})
 	if err != nil {
-		return fmt.Errorf("error attaching tracing: %w", err)
+		return fmt.Errorf("attaching tracing block_rq_complete: %w", err)
```
mauriciovasquezbernal, line 108: attaching tracing block_rq_complete -> attaching tracepoint block_rq_complete

?

Same in other locations.

mauriciovasquezbernal (approved):
LGTM. 

One minor nit: I think last two commits should be squashed: the one that modifies the eBPF code and the one that modifies the golang code. Those are IMO part of the same logical change and hence should be part of the same commit.

PR #1591, opened 2023-04-28
Mark release as draft and add dedicated release pipeline

eiffel-fl:
Hi.

In this PR, I marked all releases as draft and added a dedicated release pipeline to push to `krew` once the draft release was published.

Best regards.

### DIFF
--- a/docs/devel/release-process.md
+++ b/docs/devel/release-process.md
@@ -25,34 +25,22 @@ $ git pull origin
 
 Double check that the latest commit is the right one to make the release.
 
-2. Tag a release canditate version
+2. Tag the new version.
 
 ```bash
-$ git tag v0.x.0-rc
+$ git tag v0.x.0
 ```
 
 3. Push tag to remote
 
 ```bash
-$ git push origin v0.x.0-rc
+$ git push origin v0.x.0
 ```
 
 4. Verify that the CI for the tag passed
 
-5. Tag the real release
-
-```bash
-$ git push origin v0.x.0
-```
-
-6. Push tag for real release
-
-```bash
-$ git push origin v0.x.0
-```
-
-7. Once the CI is done, go to https://github.com/inspektor-gadget/inspektor-gadget/releases and edit
-   the created release. Then click `Generate release notes` and update them to match the format used
+5. Once the CI is done, go to https://github.com/inspektor-gadget/inspektor-gadget/releases and edit
+   the created draft release. Then click `Generate release notes` and update them to match the format used
    for other releases:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/devel/release-process.md
```diff
@@ -25,19 +25,19 @@ $ git pull origin
...
 
 ```bash
-$ git push origin v0.x.0-rc
+$ git push origin v0.x.0-draft
 ```
 
-4. Verify that the CI for the tag passed
+4. Verify that the CI for the draft tag passed
 
 5. Tag the real release
```
blanquicet, line 42: Not part of this PR but this step is wrong, it should be `git tag v0.x.0` instead of `git push origin v0.x.0` 🙂

mauriciovasquezbernal (approved):
LGTM. If you're able to fix Jose's comment that will be great as well.

I'll throw an idea here that we could implement later on, but please notice I don't want to block this PR.

We edit the release notes manually after the release is created, hence the notification email that is sent doesn't contain them:

Selection_237

If we mark the real release as draft, we'll be able to edit the release notes, and the notification email will include them once we publish the release. One issue with this approach is that the krew-index CI needs to access the release artifacts to be able to check the manifests. One possible solution is to move this out of the release pipeline and trigger is only when the release is published. (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release).

In summary, my proposal is:
- Don't do a draft release 
  - Save some time to the one that makes the release as we don't need to wait for this CI run to finish
- Mark the release as draft
  - Avoid sending and release notification email without the notes
- Move the `rajatjindal/krew-release-bot` action to be triggered when a release is published
  - We only update krew index once we have the release totally ready

If something goes wrong with a release, we can just delete it (it's a draft anyway), fix and try again. We can decide if we want to bump the minor version or force push the tag, but IMO this should be very rare so it really doesn't matter. WDYT?

docs/devel/release-process.md
```diff
@@ -25,34 +25,22 @@ $ git pull origin
...
 ```bash
-$ git tag v0.x.0-rc
+$ git tag v0.x.0
 ```
 
 3. Push tag to remote
 
 ```bash
-$ git push origin v0.x.0-rc
+$ git push origin v0.x.0-draft
```
mauriciovasquezbernal, line 37:
```suggestion
$ git push origin v0.x.0
```

?

docs/devel/release-process.md
```diff
@@ -64,11 +52,13 @@ $ git push origin v0.x.0
 ### Testing and Continue Integration
 ```
 
-8. Verify that the CI created a pull request in
+6. Once satisfied with the release notes, you can publish the draft release as public release.
```
mauriciovasquezbernal, line 55:
```suggestion
6. Once satisfied with the release notes, publish the draft release as public release.
```

mauriciovasquezbernal (approved):
LGTM (again).

Did you try this on a fork to be sure it works as expected? I know we can't try the krew-index part on a fork, but a least to be sure it's being triggered when the release is published.

Issue #1513, opened 2023-04-11
Design support for Prometheus metrics


mauriciovasquezbernal:
Part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1405

The scope of this issue is discuss and design the solution for supporting Prometheus metrics in Inspektor Gadget. The result should be a design document that covers the following aspects:
- Where to expose the metrics (ig, kubectl-gadget, standalone binary)
- Which metrics to expose? 
  - Make them configurable?
- Labels granularity 

Note: There is a PoC available in https://github.com/inspektor-gadget/inspektor-gadget/pull/1384. The design document could base on that approach or propose something different too.

flyth:
I'm currently extending the first PoC to also add support for counters that are (almost) entirely handled in eBPF and thus providing higher throughput. Basically, it's working also through the operator that hands over an interface to the gadget, which the gadget can then use to register certain metrics and get a per-cpu eBPF array/map in return. The operator then polls the map(s) every interval and exposes the values through the prometheus exporter.

The interface basically looks like:
```
RegisterMetrics([]*Metric)
RegisterMetricsPerNS(nsType, []*Metric)
```

The latter method `RegisterMetricsPerNS()` can use the LocalOperator/KubeOperator to get retrieve container, pod and namespace information and add them as labels.

Open question:

There should probably be a metrics-only mode for gadgets, so that they don't emit events but just provide updates to the counters. How can we add this as generic as possible?

> Part of #1405
> 
> The scope of this issue is discuss and design the solution for supporting Prometheus metrics in Inspektor Gadget. The result should be a design document that covers the following aspects:
> 
> * Where to expose the metrics (ig, kubectl-gadget, standalone binary)

In this case, both ig and gadgettracermanager could both support exporting metrics (not kubectl-gadget, though - should it?).

> * Which metrics to expose?

We should look at different implementations of services (like for example DNS servers) and look at what metrics they collect as a starting point.

>   
>   * Make them configurable?

With the current approach, they are plain gadgets, so configurable just like normal (including filtering etc.).

> * Labels granularity

The `RegisterMetricsPerNS()` adds the usual labels, but if we want to go father than this, we'd need to come up with even more flexible stuff (not handled by my update for now).

> 
> Note: There is a PoC available in #1384. The design document could base on that approach or propose something different too.

PR #1592, opened 2023-04-28
design: Add prometheus design doc

mauriciovasquezbernal:
Add document with the design of the Prometheus support in IG

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1513

### DIFF
--- a/docs/design/001-prometheus.md
+++ b/docs/design/001-prometheus.md
@@ -1,0 +1,447 @@
+# Prometheus support in Inspektor Gadget
+
+Inspektor Gadget has a lot of tools that hook into the kernel to capture different events like file
+opened, process created, DNS requests, etc. Currently it's mostly designed as a troubleshooting
+tool: it prints those events as they happen to the terminal. However, it's an easy win to provide
+metrics through Prometheus. The whole logic to capture the data is already in place, we only need to
+aggregate and expose this in a Prometheus format.
+
+This document contains a design proposal for supporting Prometheus metrics in Inspektor Gadget.
+Upstream issue:
+[https://github.com/inspektor-gadget/inspektor-gadget/issues/1513](https://github.com/inspektor-gadget/inspektor-gadget/issues/1513)
+
+# Goals
+
+This document is written with the following goals in mind in descending order of priority.
+
+- Bring this support soon to market
+- Metrics to expose should be configurable
+- The solution should be performant
+
+# Design Decisions
+
+## Metrics to expose
+
+In order to be as flexible as possible, the user should be able to configure the metrics they want to expose
+for each gadget. Most gadgets emit events from eBPF including several fields of data and send them to the
+userspace part of IG for processing. In a generic solution, most of these fields should be selectable for metric collection, aggregation and filtering. However, due to handling all events in userspace, this could negatively
+impact performance. In order to improve that, we also propose a way to handle collection of the most commonly
+used metrics directly in eBPF.
+
+## Labels Granularity
+
+High cardinality (a lot of distinct label combinations) can be problematic as it increases the memory
+usage of both the collector (IG) and the consumer (Prometheus). As stated above, users should still be able
+to configure the granularity they want to have and so should consider the cardinality themselves.
+
+## Filtering
+
+Inspektor Gadget already provides a mechanism to filter out events we're not interested in. This
+mechanism should be reused by the Prometheus integration to avoid handling metrics for objects the
+user is not interested in.
+
+# User Experience
+
+The metric collection and export to Prometheus should be supported in both cases, a) when running in Kubernetes (ig-k8s), and b) when running on Linux hosts (ig).
+This is possible by implementing this using a new Prometheus gadget/operator as it makes the code automatically shareable between ig, ig-k8s and external applications. This gadget/operator provides
+start / stop operations to enable / disable collection of metrics.
+
+```bash
+$ kubectl gadget prometheus start --config <path>
+$ kubectl gadget prometheus stop
+```
+
+TODO: need to think about not having a start operation on this one
+
+```bash
+$ ig prometheus --config <path>
+```
+
+It should also be possible to configure the metrics using a CR - supporting a [static
+configuration](https://github.com/inspektor-gadget/inspektor-gadget/issues/1401) could be
+implemented in the future as well.
+
+## Configuration File
+
+Given that we want this to be flexible, allowing the user to control which metrics to capture
+and how to aggregate them, we will use configuration files to define those aspects. This takes inspiration from
+[https://github.com/cloudflare/ebpf_exporter](https://github.com/cloudflare/ebpf_exporter).
+
+### Filtering (aka Selectors)
+
+The user should be able to provide a set of filters indicating the events that should be taken into
+consideration when collecting the metrics. The mechanism should provide the following features:
+
+- Equal operator
+  - "columnName:value"
+- Different operator (!)
+  - "columnName:!value"
+- Greater than, less than operators (<, >, <=, >=)
+  - "columnName:>value"
+- Match regex (~)
+  - "columnName:~regex"
+
+In the future, we could consider introducing more advanced operators like:
+
+- Set based
+  - In
+  - NotIn
+
+It's similar to the existing [Labels and
+Selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) mechanism, but
+we still need to understand if we can reuse that or if we need a completely new implementation.
+
+Some examples of possible filters are:
+
+# Only metrics for default namespace
+
+```yaml
+# Only metrics for default namespace
+selector:
+  - k8s.namespace: default
+
+# Count only events with retval != 0
+selector:
+  - "retval:!0"
+```
+
+The configuration file defines the different metrics to collect.
+
+### Counters
+
+This is probably the most intuitive metric: "A _counter_ is a cumulative metric that represents a
+single [monotonically increasing counter](https://en.wikipedia.org/wiki/Monotonic_function) whose
+value can only increase or be reset to zero on restart. For example, you can use a counter to
+represent the number of requests served, tasks completed, or errors." from
+[https://prometheus.io/docs/concepts/metric_types/#counter](https://prometheus.io/docs/concepts/metric_types/#counter).
+
+The following are examples of counters we can support with the existing gadgets. The first one
+counts the number of executed processes.
+
+```yaml
+metrics:
+  # executed processes by namespace, pod and container.
+  - name: executed_processes
+    type: counter
+    category: trace
+    gadget: exec
+    labels:
+      - k8s.namespace
+      - k8s.pod
+      - k8s.container
+```
+
+The category and gadget fields define which gadget to use. The labels indicate how metrics are
+aggregated, i.e., the cardinality of the exposed metric. In this case, we'll have a counter for each
+namespace, pod and container combination.
+
+Another example that will report the number of executed processes, aggregated by comm and namespace:
+
+# executed processes by comm and namespace
+
+```yaml
+- name: executed_processes_by_comm
+  type: counter
+  category: trace
+  gadget: exec
+  labels:
+    - k8s.namespace
+    - comm
+```
+
+It is possible to count events based on matching criteria. For instance, the following counter
+will only consider events in the default namespace.
+
+```yaml
+# executed processes by pod and container in the default namespace
+- name: executed_processes
+  type: counter
+  category: trace
+  gadget: exec
+  labels:
+    - k8s.pod
+    - k8s.container
+  selector:
+    - "k8s.namespace:default"
+```
+
+Or only count events for a given command:
+
+```yaml
+# cat executions by namespace, pod and container
+- name: executed_cats # ohno!
+  type: counter
+  category: trace
+  gadget: exec
+  labels:
+    - k8s.namespace
+    - k8s.pod
+    - k8s.container
+  selector:
+    - "comm:cat"
+```
+
+And finally, we can provide counters for failed operations:
+
+```yaml
+# failed execs by namespace, pod and container
+- name: failed_execs
+  type: counter
+  category: trace
+  gadget: exec
+  labels:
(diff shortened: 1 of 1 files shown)


docs/design/001-prometheus.md
```diff
@@ -0,0 +1,445 @@
...
+value can only increase or be reset to zero on restart. For example, you can use a counter to
+represent the number of requests served, tasks completed, or errors." from
+[https://prometheus.io/docs/concepts/metric_types/#counter](https://prometheus.io/docs/concepts/metric_types/#counter).
+
+The following are examples of counters we can support with the existing gadgets. The first one
+counts the number of executed processes.
+
+```yaml
+metrics:
+  counters:
```
flyth, line 127: IMHO, it would be better to have these keys (counters, gauges, histograms) removed and instead use a `type` field inside the actual metric configuration with a value of counter, gauge and histogram respectively. I know that the proposed change can be a bit trickier to parse, but from a user perspective it seems better and cleaner to be able to group metrics that belong together, regardless of their type.

mauriciovasquezbernal: Makes sense. I updated the document to follow this approach.

mauriciovasquezbernal:
I updated the proposal to clarify the filtering supported. I think we should focus on simple filtering for now (==, !=, >,<...), then we can think about a more advanced mechanism if needed. 

@flyth can we merge this PR?

@alban do you want to review it?

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+# Design Decisions
+
+The following are key aspects for this support.
+
+## Metrics to Expose
+
+The user should be able to configure the metrics they want to expose for each gadget. This makes
+this support much more flexible and lets the responsibility of which metrics to expose to the user.
+One drawback of this is the performance, however in the implementation section we propose a way to
+increase performance for the most common metrics.
```
flyth, line 30:
```suggestion
In order to be as flexible as possible, the user should be able to configure the metrics they want to expose
for each gadget. Most gadgets emit events from eBPF including several fields of data and send them to the
userspace part of IG for processing. In a generic solution, most of these fields should be selectable for metric collection, aggregation and filtering. However, due to handling all events in userspace, this could negatively
impact performance. In order to improve that, we also propose a way to handle collection of the most commonly
used metrics directly in eBPF.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+## Labels Granularity
+
+High cardinality (a lot of active label combinations) can be problematic as it increases the memory
+usage of Prometheus. As above, users should be able to configure the granularity they want to have;
+this will make this more flexible and will avoid us taking all considerations while defining the
+granularity of the metrics.
```
flyth, line 37:
```suggestion
High cardinality (a lot of distinct label combinations) can be problematic as it increases the memory
usage of both the collector (IG) and the consumer (Prometheus). As stated above, users should still be able
to configure the granularity they want to have and so should consider the cardinality themselves.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+TODO: need to think about not having a start operation on this one
+
+```bash
+$ ig prometheus --config <path>
+```
+
+This is also possible to configure the metrics using a CR and supporting a [static
+configuration](https://github.com/inspektor-gadget/inspektor-gadget/issues/1401) could be
+implemented in the future as well.
```
flyth, line 66:
```suggestion
It should also be possible to configure the metrics using a CR - supporting a [static
configuration](https://github.com/inspektor-gadget/inspektor-gadget/issues/1401) could be
implemented in the future as well.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+## Configuration File
+
+Given that we want this support to be flexible allowing the user to control which metrics to capture
+and how to aggregate them. The configuration file defines those aspects. It takes inspiration from
+[https://github.com/cloudflare/ebpf_exporter](https://github.com/cloudflare/ebpf_exporter).
```
flyth, line 72:
```suggestion
Given that we want this to be flexible, allowing the user to control which metrics to capture
and how to aggregate them, we will use configuration files to define those aspects. This takes inspiration from
[https://github.com/cloudflare/ebpf_exporter](https://github.com/cloudflare/ebpf_exporter).
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+    gadget: exec
+    labels:
+      - k8s.namespace
+      - k8s.pod
+      - k8s.container
+```
+
+The category and trace define which is the gadget to use. The labels session indicates how they are
+aggregated, i.e., the cardinality of the exposed metric. In this case, we'll have a counter for each
+namespace, pod and container combination.
```
flyth, line 140:
```suggestion
The category and gadget fields define which gadget to use. The labels indicate how metrics are
aggregated, i.e., the cardinality of the exposed metric. In this case, we'll have a counter for each
namespace, pod and container combination.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+Another example that will report number of executed processes by comm and namespace:
```
flyth, line 142:
```suggestion
Another example that will report the number of executed processes, aggregated by comm and namespace:
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+  type: counter
+  category: trace
+  gadget: exec
+  labels:
+    - k8s.namespace
+    - comm
+```
+
+This is possible to count events based on a matching criteria. For instance, the following counter
+will only take into consideration events in the default namespace
```
flyth, line 157:
```suggestion
It is possible to count events based on matching criteria. For instance, the following counter
will only consider events in the default namespace.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+    - k8s.container
+  selector:
+    - "k8s.namespace:default"
+```
+
+Or only count events for a given command:
+
+```yaml
+# cat executions by namespace, pod and container
+- name: executed_cats # ohno!
```
flyth, line 176: :/

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+  aggregate:
+    - k8s.namespace
+    - k8s.pod
+    - k8s.container
+  selector:
+    - "status:CLOSE_WAIT"
+```
+
+TODO: This is not totally clear how this should work since there gadget doesn't provide a stream of
+events. In this case we should execute the gadget each time prometheus scrapes the endpoint.
```
flyth, line 292: I guess they could be run like the top category in that case - using an interval. I wouldn't do things "on scrape", because of the delay this could cause (and thus skew the other metrics).

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Histograms
+
+The histogram definition is a more complex than the previous one, hence please check the Prometheus
```
flyth, line 296:
```suggestion
The histogram definition is a bit more complex than the previous ones, hence please check the Prometheus
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+## Gadgets supported
+
+We want to make Prometheus supported by as many gadgets as possible, however it's currently not
+possible to support all of them. The initial implementation covers these gadgets:
+
+- Tracers: counters and histograms
+- Snapshot: gauges
+
+There are some categories that we don't know if can be supported altogether, so probably we should
+also define a per gadget support:
```
flyth, line 332:
```suggestion
There are some categories that we don't know if they can be supported altogether, so we probably should
also define per gadget support:
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Collection in user space
+
+In this case the counting / aggregation happens on user space. This is the simplest way to collect
+metrics, but also the most expensive one. It leverages the whole functionality of our gadgets as it
+is right now. Events are still collected in eBPF and sent to user-space as they occur, where they
+are evaluated, i.e., aggregated and filtered according to the user's configuration.
+
+This option is defined to be flexible rather than performant. For critical metrics users should use
+the metrics backed by eBPF.
```
flyth, line 353:
```suggestion
This option is defined to be flexible rather than performant. For metrics with high throughput, users should use
the metrics collection backed by eBPF (see below).
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+The implementation is based on the existing parser that uses reflection underneath, a PoC is
+implemented in
+[https://github.com/inspektor-gadget/inspektor-gadget/tree/mauricio/experiments/prometheus](https://github.com/inspektor-gadget/inspektor-gadget/tree/mauricio/experiments/prometheus.)
+
+### Collection in eBPF
+
+We should extend the gadgets to collect some common metrics in eBPF to make this solution more
+performant. This should be the preferred way of collecting metrics, even if it isn't as flexible as
+the implementation using the events and harder to implement. In this case, we'll have to define a
```
flyth, line 363:
```suggestion
the implementation using the events (and it's harder to implement). In this case, we'll have to define a
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+Gadgets' eBPF code would require two new constants:
+
+```c
+bool enable_events
+bool enable_metrics
+```
+
+When metric collection is enabled, it expects maps that it can fill. The operator will provide such
+maps with a layout depending on the users' configuration.
```
flyth, line 374:
```suggestion
When metric collection is enabled, it expects maps that it can fill. The gadget/operator will provide such
maps with a layout depending on the users' configuration.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+could then create required functions (like offset helpers) and volatile consts that will be set upon
+starting the gadget.
+
+# Compatibility with script and BYOB (bring your own bpf) gadgets
+
+These two gadgets allow the user to inject custom eBPF programs. This should be possible to
+integrate them with the Prometheus integration. The structure of the eBPF maps that collect the
+metrics should be well defined to create a contract between the eBPF programs and the
+Operator/Gadget in user space.
```
flyth, line 429:
```suggestion
These two gadgets allow the user to inject custom eBPF programs. It should be possible to
also support metric collection using this solution. The structure of the eBPF maps that collect the
metrics should be well defined to create a contract between the eBPF programs and the
Operator/Gadget in user space.
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+
+## BYOB
+
+We have to document what is the structure of the maps and the use will be responsible for creating
+and filling them in their eBPF programs
```
flyth, line 434:
```suggestion
We have to document what is the structure of the maps and the user will be responsible for creating
and filling them in their eBPF programs
```

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,451 @@
...
+[https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#8-hist-log2-histogram](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#8-hist-log2-histogram).
+
+# Out of Scope
+
+- Providing a Golang package for 3rd party applications: The support will be implemented as a
+  Gadget/Operator, hence it'll be available also for 3rd party applications. It's not in our plans
+  to provide a golang package with this support, however this should be easy to refactor the code
+  and create such a package if needed in the future. At that point we'll need to determine the
+  format used by this package to expose the data. One possibility is to expose Otel metrics and the
+  user then will convert them to the needed format, Prometheus for instance.
```
flyth, line 451:
```suggestion
- Providing a Golang package for 3rd party applications: This solution will be implemented as a
  Gadget/Operator, hence it'll also be available for 3rd party applications. It's not on our roadmap
  to provide a golang package supporting this, however it should be easy to refactor the code
  and create such a package if needed in the future. At that point we'll need to determine the
  format used by this package to expose the data. One possibility is to expose Otel metrics and the
  user then will convert them to the needed format, Prometheus for instance.
```

flyth (approved):
Mostly nits, take whatever you think is appropriate :)

Thanks!

docs/design/001-prometheus.md
mauriciovasquezbernal: This is something we should think more about, specially  regarding the "cache" to avoid long times on scrapping. I remember I read somewhere clients shouldn't implement that kind of logic, but I can't find it again.

mauriciovasquezbernal:
Accepted your suggestions, thanks!

docs/design/001-prometheus.md
```diff
@@ -0,0 +1,447 @@
...
+
+These two gadgets allow the user to inject custom eBPF programs. It should be possible to
+also support metric collection using this solution. The structure of the eBPF maps that collect the
+metrics should be well defined to create a contract between the eBPF programs and the
+Operator/Gadget in user space.
+
+## BYOB
+
+We have to document what is the structure of the maps and the user will be responsible for creating
+and filling them in their eBPF programs
```
alban, line 430: Another possibility is using a bpf extension defined by Inspektor Gadget. Then the byob code would call the bpf extension:
```
struct event event = {};
event.field1 = ...;
ig_count_event(ctx, event);
```

The bpf extension would be built dynamically after inspecting the btf type of `struct event`.

That would reduce the surface of the contract between the eBPF programs and the Operator/Gadget in user space.

But we can explore that in a future PR.

alban (approved):
LGTM

I am unconvinced about the use of macros to define the keys of interest, but I think it is ok at this stage for this proposal. We can revisit if a better idea come up.

mauriciovasquezbernal:
I'll merge and we can have a further look at the eBPF part when implementing it.

PR #1593, opened 2023-05-02
socket enricher: initial sockets and delayed socket deletion

alban:
# socket enricher: initial sockets and delayed socket deletion

This uses a bpf iterator to capture initial sockets (sockets created before the socket enricher is started). Note that it's not possible to get the thread id that created the socket because all threads of the same process share the same file descriptor table.

This refactors the kallsyms code in a single package to avoid duplication.

Upon socket termination, the socket is only removed after a few seconds. This allow the socket enricher to work on recently terminated connections, for example with TCP FIN packets in the tcpdrop gadget.

## How to use

-

## Testing done

- Start a nginx container before starting Inspektor Gadget. Use netem to drop some packets and check that the socket enricher works with the tcpdrop gadget.
- Test the tcpdrop until it catches some TCP FIN packets and check that the socket enricher works in that case.


alban:
Following the [benchmark documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/CONTRIBUTING.md#benchmarks) on my laptop, I got the following:
```
goos: linux
goarch: amd64
                                                    │ main.bench  │            patched.bench            │
                                                    │   sec/op    │   sec/op     vs base                │
AllGadgetsWithContainers/container1/trace-tcpdrop-4   1.092 ± 37%   1.513 ± 21%  +38.52% (p=0.005 n=10)
```

The changes on the socket enricher decreases the performance of installing the tracer by 38%. But I think it is worth having the socket enricher working with initial sockets.

pkg/gadgets/internal/socketenricher/tracer.go
```diff
@@ -51,11 +53,14 @@ func NewSocketEnricher() (*SocketEnricher, error) {
 }
 
 func (se *SocketEnricher) start() error {
+
```
mauriciovasquezbernal: Remove this extra empty line?

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
```diff
@@ -130,6 +130,34 @@ int ig_sockets_it(struct bpf_iter__task_file *ctx)
...
+	__u64 now = bpf_ktime_get_boot_ns();
+	__u64 deletion_timestamp = socket_value->deletion_timestamp;
+	__u64 socket_expiration_ns = 1000ULL*1000ULL*1000ULL*5ULL; // 5 seconds
+
+	if (deletion_timestamp != 0 && deletion_timestamp + socket_expiration_ns < now) {
+        // The socket is expired, remove it from the map.
+        __builtin_memcpy(&tmp_key, socket_key, sizeof(struct sockets_key));
+        bpf_map_delete_elem(&sockets, &tmp_key);
+        return 0;
+    }
```
mauriciovasquezbernal, line 156: nit: spaces vs tabs.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
```diff
@@ -130,6 +130,34 @@ int ig_sockets_it(struct bpf_iter__task_file *ctx)
...
+		return 0;
+
+	__u64 now = bpf_ktime_get_boot_ns();
+	__u64 deletion_timestamp = socket_value->deletion_timestamp;
+	__u64 socket_expiration_ns = 1000ULL*1000ULL*1000ULL*5ULL; // 5 seconds
+
+	if (deletion_timestamp != 0 && deletion_timestamp + socket_expiration_ns < now) {
+        // The socket is expired, remove it from the map.
+        __builtin_memcpy(&tmp_key, socket_key, sizeof(struct sockets_key));
+        bpf_map_delete_elem(&sockets, &tmp_key);
```
mauriciovasquezbernal, line 154: Does the verifier complains if you pass `socket_key` directly to `bpf_map_delete_elem`?

pkg/gadgets/internal/socketenricher/tracer.go
```diff
@@ -162,10 +166,55 @@ func (se *SocketEnricher) start() error {
...
+	file, err := cleanupIter.Open()
+	if err != nil {
+		return fmt.Errorf("open BPF iterator: %w", err)
+	}
+	defer file.Close()
+	_, err = io.ReadAll(file)
+	if err != nil {
+		return fmt.Errorf("read BPF iterator: %w", err)
+	}
 	return nil
```
mauriciovasquezbernal, line 210: offtopic: I'm wondering if cilium/ebpf should provide an easier way to run iterators. Something like `cleanupIter.Run()` would be nice for cases the iter doesn't generate any output.

mauriciovasquezbernal (commented):
Some initial comments. I'll do another pass tomorrow.

pkg/kallsyms/kallsyms.go
```diff
@@ -0,0 +1,209 @@
...
+func NewKAllSymsFromReader(reader io.Reader) (*KAllSyms, error) {
+	symbols := []kernelSymbol{}
+	symbolsMap := map[string]uint64{}
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		fields := strings.Fields(line)
+		if len(fields) < 3 {
```
mauriciovasquezbernal, line 65: Should this return an error instead?

pkg/kallsyms/kallsyms.go
```diff
@@ -0,0 +1,209 @@
...
+		addr = k.symbols[mid].addr
+
+		if addr <= ip {
+			start = mid
+		} else {
+			end = mid - 1
+		}
+	}
+
+	if start == end && k.symbols[start].addr <= addr {
```
mauriciovasquezbernal, line 119: I know you didn't implement this, but I should check `k.symbols[start].addr <= ip` instead as done in BCC. ip here is called addr and I think it caused confusion when doing the porting. 

With the current check, a ksysall like 
```
		"ffffffff00000000 A fixed_percpu_data",
		"ffffffffb4231f40 D bpf_prog_fops",
		"ffffffffb43723e0 d socket_file_ops",
```

would return `fixed_percpu_data` for `LookupByInstructionPointer(0)` instead of `unknown`.

pkg/kallsyms/kallsyms_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/btf"
+
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+)
+
+func TestCustomKAllSyms(t *testing.T) {
+	kAllSymsStr := strings.Join([]string{
+		"0000000000000000 A fixed_percpu_data",
```
mauriciovasquezbernal, line 32: nit: test when `unknown` is return. (see my other comment.)

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
```diff
@@ -25,25 +25,111 @@ struct {
...
+
+	// At this point, the endpoint was already present, we need to determine
+	// the best entry between the existing one and the new one.
+
+	// When iterating on initial sockets, we get both passive and active
+	// sockets (server side). We want the passive socket because we don't
+	// want the endpoint to be removed from the map when just one
+	// connection is terminated. We cannot determine if an active socket
+	// is server side or client side, so we add active socket anyway on the
+	// chance that it is client side. It will be fine server side too,
```
mauriciovasquezbernal, line 91: It will be fine server side -> It will be fine for server side ?

pkg/gadgets/internal/socketenricher/tracer_test.go
```diff
@@ -246,7 +296,27 @@ func bindSocket(ipStr string, domain, typ int, port int) (uint16, int, error) {
...
+	//
+	// Example how to test this:
+	//
+	//	$ go test -c ./pkg/gadgets/internal/socketenricher/...
+	//	$ sudo ./socketenricher.test
+	//	PASS
+	//	$ mv socketenricher.test se.test
+	//	$ sudo ./se.test
+	//	FAIL
+	setProcessName("socketenricher.")
```
mauriciovasquezbernal, line 318: Check error just in case?

pkg/gadgets/internal/socketenricher/tracer_test.go
```diff
@@ -189,28 +193,75 @@ func TestSocketEnricherBind(t *testing.T) {
...
 
-			entries := socketsMapEntries(t, tracer, nil)
+			// Start the late tracer after the event has been generated
+			lateTracer, err := NewSocketEnricher()
+			if err != nil {
+				t.Fatal(err)
+			}
+			t.Cleanup(lateTracer.Close)
+
+			normalize := func(entry *socketEnricherMapEntry) {
```
mauriciovasquezbernal, line 203: normalize -> earlyNormalize?

mauriciovasquezbernal (approved):
Great job. I tried and it works perfectly. Most of my comments are nits, so I'll approve it now.

pkg/kallsyms/kallsyms.go
```diff
@@ -0,0 +1,209 @@
...
+
+// NewKAllSyms reads /proc/kallsyms and returns a KAllSyms.
+func NewKAllSyms() (*KAllSyms, error) {
+	file, err := os.Open("/proc/kallsyms")
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	return NewKAllSymsFromReader(file)
```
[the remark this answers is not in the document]
alban: I didn't cache the content of `/proc/kallsyms` permanently because it is big (10MB) and not all gadgets need all the entries in cache.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
alban: Yes, the verifier failed.
The same technique is used in this sample code:
https://github.com/torvalds/linux/blob/v6.3/tools/testing/selftests/bpf/progs/bpf_iter_bpf_hash_map.c#L70

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
```diff
@@ -25,25 +25,111 @@ struct {
...
 
+static __always_inline void
+insert_socket_from_iter(struct sock *sock, struct task_struct *task)
+{
+	struct sockets_key socket_key = {0,};
+	prepare_socket_key(&socket_key, sock);
+
+	struct sockets_value socket_value = {0,};
+	// use given task
+	socket_value.pid_tgid = ((u64)task->tgid) << 32 | task->pid;
```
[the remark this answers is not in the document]
alban: I tried to remove BPF_CORE_READ everywhere for iterator, but the following ones failed:
- `struct sock *sock = BPF_CORE_READ(socket, sk);`
- `__u16 family = BPF_CORE_READ(sock, __sk_common.skc_family);`
- `if (BPF_CORE_READ(sock, __sk_common.skc_state) == TCP_LISTEN)`

Example:

```
// file->private_data is a struct socket because we checked f_op.
struct socket *socket = (struct socket *) file->private_data;
struct sock *sock = socket->sk;
```

```
; struct socket *socket = (struct socket *) file->private_data;
11: (79) r1 = *(u64 *)(r1 +200)       ; R1_w=scalar()
; struct sock *sock = socket->sk;
12: (79) r7 = *(u64 *)(r1 +24)
R1 invalid mem access 'scalar'

processed 12 insns (limit 1000000) max_states_per_insn 0 total_states 1 peak_states 1 mark_read 1
Error: running gadget: running gadget: installing tracer: failed to load ebpf program: field
IgSocketsIt: program ig_sockets_it: load program: permission denied: 12: (79) r7 = *(u64 *)(r1 +24):
R1 invalid mem access 'scalar' (24 line(s) omitted)
```

Printing errors with:
```
	if err := spec.LoadAndAssign(&se.objs, nil); err != nil {
		var errVerifier *ebpf.VerifierError
		if errors.As(err, &errVerifier) {
			fmt.Printf("loading ebpf program:\n%+v\n", errVerifier)
		}
		return fmt.Errorf("failed to load ebpf program: %w", err)
	}
```

It seems to work for the structs that were passed as arguments to the iterator (`struct bpf_iter__task_file`) such as `struct task_struct`, `struct file`... but it does not work after the cast via `(struct socket *) file->private_data`.

pkg/kallsyms/kallsyms_test.go
alban: But since there is an entry with address 0, it will never return unknown in practice.

PR #1595, opened 2023-05-03
localmanager: Reintroduce custom socket paths for container runtimes

mqasimsarfraz:
This PR reintroduces the flags related to custom socket paths for container runtimes:

## Before the PR

```
→ ig -h
...
...
Flags:
  -h, --help              help for ig
  -r, --runtimes string   Container runtimes to be used separated by comma. Supported values are: docker, containerd, cri-o (default "docker,containerd,cri-o")
...
```

## After the PR
```
→ ./ig -h
...
...
Flags:
      --containerd-socketpath string   Containerd CRI Unix socket path (default "/run/containerd/containerd.sock")
      --crio-socketpath string         CRI-O CRI Unix socket path (default "/var/run/crio/crio.sock")
      --docker-socketpath string       Docker Engine API Unix socket path (default "/var/run/docker.sock")
  -h, --help                           help for ig
  -r, --runtimes string                Container runtimes to be used separated by comma. Supported values are: docker, containerd, cri-o (default "docker,containerd,cri-o")
...
```

## Testing done

```
→ sudo ./ig snapshot process -r docker
CONTAINER                                                                                                                                                                COMM             PID                  UID        GID       
loving_chatelet                                                                                                                                                          sleep            125182               0          0         
quizzical_galois 

→ sudo ./ig snapshot process -r docker --docker-socketpath /run/docker.sock
CONTAINER                                                                                                                                                                COMM             PID                  UID        GID       
loving_chatelet                                                                                                                                                          sleep            125182               0          0         
quizzical_galois 
```

### DIFF
--- a/pkg/operators/localmanager/localmanager.go
+++ b/pkg/operators/localmanager/localmanager.go
@@ -37,6 +37,9 @@ const (
 	OperatorInstanceName = "LocalManagerTrace"
 	Runtimes             = "runtimes"
 	ContainerName        = "containername"
+	DockerSocketPath     = "docker-socketpath"
+	ContainerdSocketPath = "containerd-socketpath"
+	CrioSocketPath       = "crio-socketpath"
 )
 
 type MountNsMapSetter interface {
@@ -74,6 +77,21 @@ func (l *LocalManager) GlobalParamDescs() params.ParamDescs {
 			Description: fmt.Sprintf("Container runtimes to be used separated by comma. Supported values are: %s",
 				strings.Join(containerutils.AvailableRuntimes, ", ")),
 			// PossibleValues: containerutils.AvailableRuntimes, // TODO
+		},
+		{
+			Key:          DockerSocketPath,
+			DefaultValue: runtimeclient.DockerDefaultSocketPath,
+			Description:  "Docker Engine API Unix socket path",
+		},
+		{
+			Key:          ContainerdSocketPath,
+			DefaultValue: runtimeclient.ContainerdDefaultSocketPath,
+			Description:  "Containerd CRI Unix socket path",
+		},
+		{
+			Key:          CrioSocketPath,
+			DefaultValue: runtimeclient.CrioDefaultSocketPath,
+			Description:  "CRI-O CRI Unix socket path",
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


alban (approved):
This looks good.

Do you plan to add the podman socket in the other PR?

mqasimsarfraz:
> Do you plan to add the podman socket in the other PR?

Exactly, I will rebase and add the support in #1532 after merge.

burak-ok (commented):
So the previous implementation was not completely removed:
There is this common function that was used for both commands `ig` and `kubectl-gadget`: [flags.go](https://github.com/inspektor-gadget/inspektor-gadget/blob/1fdd282f050c21117e81851f42dbe7dddea1b2cf/cmd/common/utils/flags.go#L104)

1. `kubectl-gadget`:
    This is only done for the `deploy` [subcommand](https://github.com/inspektor-gadget/inspektor-gadget/blob/9573897e5169cc7c133951c2885af725e4187ff0/cmd/kubectl-gadget/deploy.go#L87) (rightfully)
3. `ig`:
    This seems to be only called for the `list-containers` [subcommand](https://github.com/inspektor-gadget/inspektor-gadget/blob/dc3c4f23644cebe7d4afd3dac1c8ecd03e1b6358/cmd/ig/containers/containers.go#L112). That doesn't make any sense so your approach of putting in the `localmanager` is better. But you would should also remove the above usage in `list-containers`.

If it is possible a common file with the default values would be nice to have. So default values for `ig` and `kubectl-gadget` is the same without modifying 2 places (Don't know if it is easily possible)

mqasimsarfraz:
> 2. `ig`:
>    This seems to be only called for the `list-containers` [subcommand](https://github.com/inspektor-gadget/inspektor-gadget/blob/dc3c4f23644cebe7d4afd3dac1c8ecd03e1b6358/cmd/ig/containers/containers.go#L112). That doesn't make any sense so your approach of putting in the `localmanager` is better. But you would should also remove the above usage in `list-containers`.

I believe since `list-containers` isn't really a gadget but a special case so it is handled differently and hence requires its own way of doing. Maybe we can use `operators.GlobalParamsCollection()` to get the params but we will to register the flags afterwards as well. Not sure if it was our intention to refactor it. @flyth any thoughts on this?

burak-ok:
> I believe since `list-containers` isn't really a gadget but a special case so it is handled differently and hence requires its own way of doing.

That is right. I forgot about that

flyth:
I'd just prefer to have the actual default values in one place (`runtimeclient.DockerDefaultSocketPath`, etc.) and re-use them - I would not refactor `list-containers`, yet (actually, there were plans to move that and the debug-commands to kind of "internally used" gadgets later on (not labelled as gadgets) - just to use the same infrastructure and reduce code.)

mqasimsarfraz:
> I'd just prefer to have the actual default values in one place (`runtimeclient.DockerDefaultSocketPath`, etc.) and re-use them

Ah! I missed them. Perfect, we can use the default paths from `runtimeclient`. I will update the PR, thanks for the hint!

mqasimsarfraz:
Thanks for the reviews!

Issue #1316, opened 2023-02-09
[RFE] Add Inspektor Gadget as Minikube Addon


alban:
## Current situation

Users who want to try Inspektor Gadget on Minikube needs to install it manually.

Minikube has [addons](https://minikube.sigs.k8s.io/docs/commands/addons/) but Inspektor Gadget is not among them:

<details>

```
$ minikube addons list
|-----------------------------|----------|--------------|--------------------------------|
|         ADDON NAME          | PROFILE  |    STATUS    |           MAINTAINER           |
|-----------------------------|----------|--------------|--------------------------------|
| ambassador                  | minikube | disabled     | 3rd party (Ambassador)         |
| auto-pause                  | minikube | disabled     | Google                         |
| cloud-spanner               | minikube | disabled     | Google                         |
| csi-hostpath-driver         | minikube | disabled     | Kubernetes                     |
| dashboard                   | minikube | disabled     | Kubernetes                     |
| default-storageclass        | minikube | enabled ✅   | Kubernetes                     |
| efk                         | minikube | disabled     | 3rd party (Elastic)            |
| freshpod                    | minikube | disabled     | Google                         |
| gcp-auth                    | minikube | disabled     | Google                         |
| gvisor                      | minikube | disabled     | Google                         |
| headlamp                    | minikube | disabled     | 3rd party (kinvolk.io)         |
| helm-tiller                 | minikube | disabled     | 3rd party (Helm)               |
| inaccel                     | minikube | disabled     | 3rd party (InAccel             |
|                             |          |              | [info@inaccel.com])            |
| ingress                     | minikube | disabled     | Kubernetes                     |
| ingress-dns                 | minikube | disabled     | Google                         |
| istio                       | minikube | disabled     | 3rd party (Istio)              |
| istio-provisioner           | minikube | disabled     | 3rd party (Istio)              |
| kong                        | minikube | disabled     | 3rd party (Kong HQ)            |
| kubevirt                    | minikube | disabled     | 3rd party (KubeVirt)           |
| logviewer                   | minikube | disabled     | 3rd party (unknown)            |
| metallb                     | minikube | disabled     | 3rd party (MetalLB)            |
| metrics-server              | minikube | disabled     | Kubernetes                     |
| nvidia-driver-installer     | minikube | disabled     | Google                         |
| nvidia-gpu-device-plugin    | minikube | disabled     | 3rd party (Nvidia)             |
[...]

PR #1605, opened 2023-05-08
runcfanotify: add missing /host for pid file

alban:
When ig or gadgettracermanager run inside a container, runcfanotify monitors execution of runc on the host (e.g. /host/usr/bin/runc) and then monitors the pid file such as:
```
/run/containerd/io.containerd.runtime.v2.task/moby/d29.../init.pid
```

The Go fanotify package provides GetPath() to gives the path of a file descriptor of the pid file. But it provides the path from the host point of view, without the /host part.

This is usually fine anyway because gadgettracermanager runs in a container with a volume /run:/run, so it can find the pid file correctly.

But in some conditions, it does not work: the Docker Desktop OS:
- uses a path like /var/run/desktop-containerd/daemon/io.containerd.runtime.v2.task/moby/1c5.../init.pid
- does not have the symlink /var/run->/run but the gadget container has it.

So when Inspektor Gadget tries to read /var/run/desk... in its container, it follows its symlink /var/run->/run, then /run is the volume. But host /run/desk... directory does not exist.

Symptoms:
```
WARN[0010] error watching pid: stat /var/run/desktop-containerd/daemon/io.containerd.runtime.v2.task/moby/1c5.../init.pid: no such file or directory
```

This patch access the pid file via the /host volume when configured, so it does not rely on the /run:/run volume or the /var/run->/run symlink on the host OS.

This patch will be useful for #1316.


pkg/runcfanotify/runcfanotify.go
```diff
@@ -416,6 +416,7 @@ func (n *RuncNotifier) watchPidFileIterate(pidFileDirNotify *fanotify.NotifyFD,
 	if err != nil {
 		return false, err
 	}
+	path = filepath.Join("/", os.Getenv("HOST_ROOT"), path)
```
mqasimsarfraz, line 419: nit: We can use [`hostRoot`](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/snapshot/process/tracer/tracer.go#L48) here as well.

mqasimsarfraz (approved):
LGTM! IIUC the docker extension container (if it runs in one) will be required to set `HOST_ROOT` for this to work correctly. Right?

pkg/runcfanotify/runcfanotify.go
alban: The hostRoot is in the snapshot process package whereas the runcfanotify package should be usable outside of IG. But I will reuse the same code pattern.

PR #1608, opened 2023-05-08
docs/devel/release-process.md: add post-release instructions

alban:
We should not keep the milestone for the release open. We should update the Inspektor Gadget Addon in Minikube.


mauriciovasquezbernal (approved):
LGTM, thanks!

alban:
Should I add a post-release task to update the vscode-aks-tools extension?
https://github.com/Azure/vscode-aks-tools/pull/191

mauriciovasquezbernal:
I think so. It's good to keep those versions updated, even if this has to be done manually for now.

PR #1619, opened 2023-05-10
pkg/types: Create ip version column template

blanquicet:
This PR creates a template for the ip version column and applies it to all gadgets with such a column. In particular, notice that the top tcp gadget was using `maxWidth` instead of `width:2,fixed`, thus this PR uniforms it with other gadgets.

### DIFF
--- a/pkg/gadgets/top/tcp/types/types.go
+++ b/pkg/gadgets/top/tcp/types/types.go
@@ -49,7 +49,7 @@ type Stats struct {
 
 	Pid       int32  `json:"pid,omitempty" column:"pid,template:pid"`
 	Comm      string `json:"comm,omitempty" column:"comm,template:comm"`
-	IPVersion uint16 `json:"ipversion,omitempty" column:"ip,maxWidth:2"`
+	IPVersion uint16 `json:"ipversion,omitempty" column:"ip,template:ipversion"`
 	Saddr     string `json:"saddr,omitempty" column:"saddr,template:ipaddr,hide"`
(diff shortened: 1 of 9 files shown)


pkg/gadgets/top/tcp/types/types.go
```diff
@@ -49,7 +49,7 @@ type Stats struct {
 
 	Pid      int32  `json:"pid,omitempty" column:"pid,template:pid"`
 	Comm     string `json:"comm,omitempty" column:"comm,template:comm"`
-	Family   uint16 `json:"family,omitempty" column:"ip,maxWidth:2"`
+	Family   uint16 `json:"family,omitempty" column:"ip,template:ip"`
```
alban, line 52:
```suggestion
	IPVersion   uint16 `json:"ipversion,omitempty" column:"ip,template:ip"`
```

pkg/types/types.go
```diff
@@ -44,6 +44,7 @@ func init() {
 	// Max: 0000:0000:0000:0000:0000:ffff:XXX.XXX.XXX.XXX (IPv4-mapped IPv6 address) = 45
 	columns.MustRegisterTemplate("ipaddr", "minWidth:15,maxWidth:45")
 	columns.MustRegisterTemplate("ipport", "minWidth:type")
+	columns.MustRegisterTemplate("ip", "width:2,fixed")
```
mauriciovasquezbernal, line 47: IMO `ip` is confusing. It should be `ipversion`.

mauriciovasquezbernal (approved):
LGTM after handling two minor nits.

Issue #1514, opened 2023-04-11
Implement support for Prometheus metrics


mauriciovasquezbernal:
Part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1405

Implement the design done in #1513. 

TODO: Add further information once we have a design in place.

PR #1527, opened 2023-04-12
New gadget: byob

burak-ok:
Based on https://github.com/inspektor-gadget/inspektor-gadget/pull/1151:

# New gadget: byob

Based on https://github.com/solo-io/bumblebee

### ig
```
$ go run -exec sudo ./cmd/ig/... trace byob --oci-image="ghcr.io/burak-ok/tcpconnect:dev" -r docker
CONTAINER    PAYLOAD
             {"daddr":"192.168.49.2","saddr":"192.168.49.2"}
             {"daddr":"127.0.0.1","saddr":"127.0.0.1"}
             {"daddr":"192.168.49.2","saddr":"192.168.49.2"}
```

## How to use

* Clone https://github.com/burak-ok/bumblebee/tree/burak/update_cilium
* Push your ebpf program in a OCI registry
```
$ bee build \
    examples/tcpconnect/tcpconnect.c \
    ghcr.io/burak-ok/tcpconnect:dev \
    -i ghcr.io/solo-io/bumblebee/builder:0.0.9
$ bee push ghcr.io/burak-ok/tcpconnect:dev
```

* Use `ig trace byob ----oci-image ...`
* When using the executable produced by `make ig`: Use `GODEBUG=netdns=go ig trace byob ----oci-image ...`

## Testing done

Tested with the [tcpconnect](https://github.com/burak-ok/bumblebee/blob/burak/update_cilium/examples/tcpconnect/tcpconnect.c) ebpf program above.

## TODO

- [ ] container enrichment in userspace (pass mntnsid from ebpf to userspace)
- [ ] container enrichment in kernel (provide a "containers" ebpf map to the ebpf program updated with container-collection)
- [ ] container filtering
- [ ] generate columns dynamically
- [x] support in kubectl-gadget (in addition to local-gadget)
- [ ] support ebpf programs other than kprobes/kretprobes
- [ ] support ebpf maps other than ring buffer
- [ ] upstream cilium/ebpf patches
- [ ] upstream bumblebee patches

PR #1620, opened 2023-05-10
Prometheus: Implement metrics collection in user space

mauriciovasquezbernal:
This commit implements a new Prometheus gadget that collects metrics
and exposes them to Prometheus.

The Prometheus gadget reuses the existing gadgets to collect the metrics.
It creates instances of them and registers a callback to get the events
they generate, then based on that it updates the metrics according to
the user configuration.

Check docs/design/001-prometheus.md to get more information about the
design of this feature

### Implementation

One key of this implementation is to be gadget neutral, it means there
is not any gadget-specific logic in this. It's generic and is able to
collect metrics from all gadgets. For this, we have to rely on the parser
(and hence columns) packages that abstract the types of the events that
each gadget provides.

#### Counters

The prometheus gadget registers a callback that is called each time the
gadget generates and event. For each event we need to:

1. Filter out the event: This is done by two different components,
well-known fields like namespace, pod and container are filtered by
the operators in eBPF. More specific fields are filtered by the already
existing filtering implementation in the parser.

2. Calculate attributes: Based on the labels configured by the user, we
need to get the attributes for the current event. This is implemented
by having a `AttrsGetter` method that returns a callback that generates
the attributes for a given event.

3. Increase the counter: Once we have the attributes, we can increase the
counter by using the otel meter.

#### Gauges:

The prometheus gadget uses the snapshot gadgets to get a list of a given
resource. The filtering is configured as the counters, but in this case
the prometheus gadget itself aggregates entries as configured by the user.

### How to use

Please check the documentation (docs/gadgets/prometheus.md) that contains some examples. 

### TODO: (from easy to more difficult)
- [X] other filtering operators (it was already implemented -> https://github.com/inspektor-gadget/inspektor-gadget/pull/1622)
- [x] float gauges
- [X] unit tests
- [x] integration tests? 
- [x] documentation for ig case

### TODO in future PRs
- [ ] histograms 

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1514


mqasimsarfraz (commented):
I just did some basic testing and was able to see metrics in my prometheus installation. 

I will continue to review other parts now!

docs/gadgets/prometheus.md
```diff
@@ -0,0 +1,316 @@
...
+
+```bash
+$ kubectl gadget prometheus --config /tmp/myconfig.yaml --listen-address 127.0.0.1:2223 --metrics-path /metrics
+INFO[0000] ubuntu-hirsute       | metrics server listening on 127.0.0.1:2223
+```
+
+We can see that none pod is executing cat.
+
+```bash
+$ curl http://127.0.0.1:2223/metrics
```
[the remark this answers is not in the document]
flyth: I think we should move the prometheus exporter (listener) to an operator. Operators have their global part that works as a singleton with the lifespan of the process (rather than the lifespan of the gadget run). The gadget should IMHO then register the metrics with the operator and expose them through it. That way we can also use annotations like you mentioned for autodiscovery. (Basically the exporting part from [my earlier PoC](https://github.com/inspektor-gadget/inspektor-gadget/pull/1384)).

pkg/gadgets/prometheus/tracer/tracer.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (t *Tracer) Run(gadgetCtx gadgets.GadgetContext) error {
+	ctx := gadgetCtx.Context()
+
+	params := gadgetCtx.GadgetParams()
+	metricsConfig := params.Get(ParamConfig).AsString()
+	metricsPath := params.Get(ParamMetricsPath).AsString()
+	listenAddress := params.Get(ParamListenAddress).AsString()
+
+	config, err := igprometheus.ParseConfig(filepath.Join(hostRoot, metricsConfig))
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Good point. Yes, I think the file should be on the client machine and somehow we should transfer its contents to the remote machine. 

I implemented a solution for this in https://github.com/inspektor-gadget/inspektor-gadget/pull/1620/commits/61df4a663378c8dda2cb44718380f9ff4fd28147, what do you think about this approach?

@flyth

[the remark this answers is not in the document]
flyth: We had a similar discussion in #1527 - I don't think the Params library should read files itself, as that might become a security issue pretty fast. Instead, what we discussed in the other PR was to use a prefix '@' to point at a local file and then transfer it transparently as string (compressed in that case) - the file is locally being read by the cobra integration (registry.go). IMHO we should use that behavior globally.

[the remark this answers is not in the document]
flyth: Ah, yeah - it would probably make sense to wrap the params before calling 
```
flag := cmd.PersistentFlags().VarPF(p, p.Key, p.Alias, desc)
```

Something like [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/778ec5e1e5144cd26d968559788dbe7956334c11) could hook into that, leaving the verification intact.

/cc @burak-ok

docs/gadgets/prometheus.md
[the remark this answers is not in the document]
mauriciovasquezbernal: I updated to use a prometheus operator as suggested. The discoverability part is still to be done.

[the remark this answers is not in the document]
mauriciovasquezbernal: @flyth an issue I noticed after this change is that there is not any way to set global params for operators when using kubectl-gadget. I think those should be (somehow) part of the deploy command. Do you have any thought on how to make it work?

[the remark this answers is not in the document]
flyth: > an issue I noticed after this change is that there is not any way to set global params for operators when using kubectl-gadget. I think those should be (somehow) part of the deploy command. Do you have any thought on how to make it work?

This applies to operator & runtime global params - both should be bound to the deployment IMHO. I think [viper](https://github.com/spf13/viper#working-with-flags) is commonly used in combination with cobra to parse environment variables and/or config files, so we could use that as well. We should be able to push all our params/flags through it. The only thing that's again a bit weird (again), is having those flags also added to the deploy command, as the operators aren't compiled in. It's similar to the issue we have when there's no catalog. Maybe it makes sense to have a build step that extracts all those information from the gadget image and then include that into the kubctl-gadget binary. Maybe @alban has some ideas as well.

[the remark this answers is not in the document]
mauriciovasquezbernal: After an internal discussion we agreed not to tackle this issue now given that https://github.com/inspektor-gadget/inspektor-gadget/pull/1648 is merged and hence not being able to configure the port is not a big issue.

pkg/resources/manifests/deploy.yaml
```diff
@@ -102,6 +102,10 @@ spec:
         # (For reference, see: https://github.com/inspektor-gadget/inspektor-gadget/runs/3966318270?check_suite_focus=true#step:20:221)
         container.apparmor.security.beta.kubernetes.io/gadget: "unconfined"
         inspektor-gadget.kinvolk.io/option-hook-mode: "auto"
+        # keep aligned with values in pkg/operators/prometheus/prometheus.go
```
mqasimsarfraz, line 105: Nice. Probably we should also document how to disable scrapping with something:

```bash
$ kubectl annotate daemonsets -l k8s-app=gadget -n gadget prometheus.io/scrape=false --overwrite
$ kubectl annotate pod -l k8s-app=gadget -n gadget prometheus.io/scrape=false
```

pkg/resources/manifests/deploy.yaml
```diff
@@ -102,6 +102,10 @@ spec:
...
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "2223"
```
mqasimsarfraz, line 107: Does this work without listing the [port](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports) in our gadget pod ?

mqasimsarfraz (approved):
This is looking great. I feel once we update the documentation this should be ready to go in!

pkg/gadgets/prometheus/tracer/tracer.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/1656. Validation is working fine too.

docs/gadgets/prometheus.md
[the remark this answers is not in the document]
mauriciovasquezbernal: I also added the pod annotations.

pkg/resources/manifests/deploy.yaml
mauriciovasquezbernal: Or make it a deploy flag? `.. deploy --disable-prometheus-autodiscovery`?

pkg/resources/manifests/deploy.yaml
mauriciovasquezbernal: Yes, it does! I think listing the ports is only needed when endpoints or service is configured.

pkg/prometheus/config.go
```diff
@@ -0,0 +1,42 @@
...
+	Field    string   `yaml:"field,omitempty"`
+	Labels   []string `yaml:"labels,omitempty"`
+	Selector []string `yaml:"selector,omitempty"`
+}
+
+type Config struct {
+	Metrics []Metric `yaml:"metrics,omitempty"`
+}
+
+func ParseConfig(configBytes []byte) (*Config, error) {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Added more strict checks for mandatory fields.

docs/gadgets/prometheus.md
```diff
@@ -0,0 +1,318 @@
...
+      - namespace
+      - pod
+      - container
+```
+
+Start the gadget
+
+```bash
+$ kubectl gadget prometheus --config @myconfig.yaml
+INFO[0000] ubuntu-hirsute       | metrics server listening on 127.0.0.1:2223
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I added another info message, however I'm not totally sure about it.

pkg/resources/manifests/deploy.yaml
mauriciovasquezbernal: I added some new documentation that uses this approach. Could you check this out?

README.md
```diff
@@ -61,6 +61,7 @@ Explore the following documentation to find out which tools can help you in your
 	- [`tcpconnect`](docs/gadgets/trace/tcpconnect.md)
 	- [`tcpdrop`](docs/gadgets/trace/tcpdrop.md)
 	- [`tcpretrans`](docs/gadgets/trace/tcpretrans.md)
+- [`prometheus`](docs/gadgets/prometheus.md)
```
mqasimsarfraz, line 64: Need to update ["How to use"](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/README.md#how-to-use) section as well!

mauriciovasquezbernal: oh thanks. I was assuming it was a flake test given that it's run in the integration tests.

docs/gadgets/prometheus.md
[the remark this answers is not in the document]
mqasimsarfraz: > time="2023-05-26T10:17:20Z" level=info msg="minikube-containerd  | Metrics will be updated while this gadget is running."
> time="2023-05-26T10:17:20Z" level=info msg="minikube-containerd-m02 | Metrics will be updated while this gadget is running."

What do you think about a message like:

```bash
time="2023-05-26T10:17:20Z" level=info msg="minikube-containerd  | Publishing metrics..."
time="2023-05-26T10:17:20Z" level=info msg="minikube-containerd-m02  | Publishing metrics..."
Hit Ctrl-C to end
```

docs/gadgets/prometheus.md
```diff
@@ -0,0 +1,364 @@
...
+# prometheus.yaml
+scrape_configs:
+- job_name: ig
+  scrape_interval: 1s
+  static_configs:
+  - targets:
+    - localhost:2223
+```
+
+Start prometheus with such configuration
```
mqasimsarfraz, line 333:
```suggestion
Start prometheus with above configuration (please refer to [docker installation](https://prometheus.io/docs/prometheus/latest/installation/#using-docker) in case you want to run prometheus in a container)
```

docs/gadgets/prometheus.md
```diff
@@ -0,0 +1,364 @@
...
+
+```bash
+$ prometheus --config.file prometheus.yaml
+```
+
+Then, start the prometheus gadget with the same configuration as above:
```
mqasimsarfraz, line 339:
```suggestion
Then, start the prometheus gadget with the same configuration as above Kubernetes section:
```

docs/gadgets/prometheus.yaml
mqasimsarfraz, line 1: We should probably move this file to https://github.com/inspektor-gadget/inspektor-gadget/tree/main/docs/examples in prometheus. Just to be clear this is just an example?

docs/gadgets/prometheus.md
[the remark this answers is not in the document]
mauriciovasquezbernal: I did something like:

```
INFO[0000] minikube             | Running. Publishing metrics...
INFO[0000] minikube-m02         | Running. Publishing metrics...
```

It seems printing the ctrl + c message after the gadget's one is difficult. @flyth what do you think about the change on my  the last commit of this PR?

pkg/gadgets/prometheus/example.yaml
```diff
@@ -0,0 +1,54 @@
...
+    labels:
+      - namespace
+      - pod
+      - container
+    #selector:
+    # - "comm:cat"
+
+
+#  counters:
+#    - name: "mymetric_root_executions"
```
mqasimsarfraz, line 15: nit: These examples needs to be updated.

mauriciovasquezbernal: ah right. That file is not intended to be committed. I'll remove it.

docs/gadgets/prometheus.md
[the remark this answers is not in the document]
flyth: Yeah, it makes sense. Maybe also add a check to not display it if a timeout is set.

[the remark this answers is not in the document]
mqasimsarfraz: It is looking better now, the updated message needs to be synced with docs as well!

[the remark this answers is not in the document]
mauriciovasquezbernal: Keeping all those messages in sync is a challenge. Thanks for the reminder.

mqasimsarfraz (approved):
LGTM! Thanks a lot for handling all the comments! 

It is so great to have Prometheus support!

docs/gadgets/prometheus.md
```diff
@@ -0,0 +1,369 @@
...
+docker run --rm -ti --name=mycontainer busybox sh -c 'for i in $(seq 0 1 1000); do cat /dev/null ; ping -c 1 localhost > /dev/null; done'
+```
+
+We can see how the counter for `mycontainer` is increased.
+![ig counter](../images/prometheus_ig_counter1.png)
+
+### Limitations
+
+- The `kubectl gadget` instance has to keep running in order to update the metrics.
+- Histograms aren't supported
```
flyth, line 368:
```suggestion
```

See above. IMHO, let's not mention stuff we don't support in the first place.

pkg/parser/parser.go
```diff
@@ -358,3 +387,278 @@ func (p *parser[T]) SetFilters(filters []string) error {
...
+		if !ok {
+			return nil, fmt.Errorf("unknown column: %s", colName)
+		}
+
+		var getter func(*T) attribute.Value
+
+		switch col.Kind() {
+		case reflect.Int:
+			getter = func(a *T) attribute.Value {
+				ff := columns.GetFieldFunc[int, T](col)
```
flyth, line 410: `ff` can be cached by moving it outside of the getter. See also below.

pkg/parser/parser.go
```diff
@@ -358,3 +387,278 @@ func (p *parser[T]) SetFilters(filters []string) error {
...
+	if !ok {
+		return nil, fmt.Errorf("colunm %s not found", colName)
+	}
+
+	var getter func(any) int64
+
+	switch col.Kind() {
+	case reflect.Int:
+		getter = func(a any) int64 {
+			ff := columns.GetFieldFunc[int, T](col)
```
flyth, line 583: Same as above, can be moved outside the getter for caching.

flyth (approved):
Awesome work, great to see this introduced in IG, thanks!

I've got just a couple of comments, mostly regarding the docs, nothing blocking - we can also adjust everything later on as well.

docs/gadgets/prometheus.md
```diff
@@ -0,0 +1,369 @@
...
+    category: trace
+    gadget: exec
+    labels:
+      - namespace
+      - pod
+      - container
+```
+
+By default, a counter is increased by one each time there is an event, however it's possible to
+increase a counter using a field on the event too.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I had a TODO there to create such example but I didn't find it neither. It's useful for increasing network bytes for instance, but we don't have any gadget that generates an event for each network packet.

docs/gadgets/prometheus.md
mauriciovasquezbernal: I removed the mention to histograms above, but do you think we should remove this line too?

pkg/parser/parser.go
mauriciovasquezbernal: (Not relevant anymore as used `GetFieldAsNumberFunc` as suggested by you)

docs/gadgets/prometheus.md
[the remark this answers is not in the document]
flyth: Yeah, let's just leave it as it is for now, then.

docs/gadgets/prometheus.md
flyth: I don't really mind it being there, it just forces us to remember removing it once we add support for histograms.

integration/inspektor-gadget/testdata/prometheus/counter.yaml
```diff
@@ -0,0 +1,11 @@
+metrics:
```
mqasimsarfraz, line 1: This is missing `metrics_name`. Also, we need to update the documentation to include `metrics_name` everywhere?

mauriciovasquezbernal: oh right, I only updated the part of the guide. Thanks!

pkg/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,633 @@
...
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctx, cancel := context.WithCancel(context.Background())
+			t.Cleanup(cancel)
+
+			wg := &sync.WaitGroup{}
+			wg.Add(len(test.config.Metrics))
```
alban, line 566: The test fails after the 10-minute timeout for me.

@mauriciovasquezbernal @flyth Who is calling `wg.Done()` for metrics of type `nonvalidtype` (used in TestMetrics)?

mauriciovasquezbernal: For those cases the code doesn't arrive to this check but returns early in 
https://github.com/inspektor-gadget/inspektor-gadget/blob/787ecf3e17a9404a0cfa52003abc6bc9faa134d8/pkg/prometheus/prometheus_test.go#L574-L577

mauriciovasquezbernal: Do you have more details about your issues / environment? I checked and they work fine in my machine:

```bash
$ cd pkg/prometheus
$ go test  ./... -count 1
ok  	github.com/inspektor-gadget/inspektor-gadget/pkg/prometheus	0.004s
```

alban: It was during my tests in #1809.

When pkg/runtime/local/local.go RunGadget() returns an error (because of my local changes), then it times out.

mauriciovasquezbernal: Unfortunately it's not easy to handle as RunGadget() is called in a different goroutine https://github.com/inspektor-gadget/inspektor-gadget/blob/787ecf3e17a9404a0cfa52003abc6bc9faa134d8/pkg/prometheus/prometheus.go#L272 and its return error is ignored. 

Perhaps a solution is to implementing something like https://stackoverflow.com/questions/32840687/timeout-for-waitgroup-wait to avoid waiting for the whole test suite to timeout. WDYT?

alban: Yes, the timeout solution in the test sounds fine.

mauriciovasquezbernal: Opened https://github.com/inspektor-gadget/inspektor-gadget/pull/1833.

PR #1621, opened 2023-05-11
pkg: Unwrap verifier error.

eiffel-fl:
When gadgets failed to run and returns a verifier error, we unwrap it so get the whole verifier stack.

Suggested-by: Michael Friese <mfriese@microsoft.com>

### DIFF
--- a/pkg/operators/kubemanager/kubemanager.go
+++ b/pkg/operators/kubemanager/kubemanager.go
@@ -15,6 +15,7 @@
 package kubemanager
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


flyth:
Thanks! You could use the `gadgetCtx.Logger().Debugf()` directly to output there instead of wrapping and returning the error again. Also, the same error handling would be needed for other entry points like the operators ([LocalManager](https://github.com/inspektor-gadget/inspektor-gadget/blob/d2aae0501b9d10d65afcb4caf2eb6623219d7973/pkg/operators/localmanager/localmanager.go#L252), [KubeManager](https://github.com/inspektor-gadget/inspektor-gadget/blob/d2aae0501b9d10d65afcb4caf2eb6623219d7973/pkg/operators/kubemanager/kubemanager.go#L245)), where you can use `m.gadgetCtx.Logger().Debugf()`.

mauriciovasquezbernal:
@eiffel-fl can we merge it?

PR #1622, opened 2023-05-11
Fully document filtering operators

mauriciovasquezbernal:
Add documentation for >=, <, <= operators too.

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -460,7 +460,10 @@ func buildCommandFromGadget(
   A filter can match any column using the following syntax
     columnName:value       - matches, if the content of columnName equals exactly value
     columnName:!value      - matches, if the content of columnName does not equal exactly value
-    columnName:>=value     - matches, if the content of columnName is greater or equal to the value
+    columnName:>=value     - matches, if the content of columnName is greater than or equal to the value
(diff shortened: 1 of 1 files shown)


cmd/common/registry.go
```diff
@@ -461,6 +461,9 @@ func buildCommandFromGadget(
     columnName:value       - matches, if the content of columnName equals exactly value
     columnName:!value      - matches, if the content of columnName does not equal exactly value
     columnName:>=value     - matches, if the content of columnName is greater or equal to the value
+    columnName:>value      - matches, if the content of columnName is greater than the value
+    columnName:<=value     - matches, if the content of columnName is lower or equal to the value
+    columnName:<value      - matches, if the content of columnName is lower than the value
```
flyth, line 466:
```suggestion
    columnName:>=value     - matches, if the content of columnName is greater than or equal to the value
    columnName:>value      - matches, if the content of columnName is greater than the value
    columnName:<=value     - matches, if the content of columnName is less than or equal to the value
    columnName:<value      - matches, if the content of columnName is less than the value
```

nit: "less" seems more commonly used for numbers.

flyth (approved):
Thanks!

Issue #1568, opened 2023-04-24
gadgets: Use KubeIPResolver in networking gadgets


mauriciovasquezbernal:
The [KubeIPResolver](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/kubeipresolver/kubeipresolver.go) operator enriches events by providing context to the IP addresses (like service / pod name). Currently this is only used in a couple of gadgets (trace network, trace tcpdrop), however we should extend and use it in more gadgets. 

It's not clear yet if we should enable this by default or not given the overhead it has. Perhaps we should disable it and only enable once we optimized this resolver? 

https://github.com/inspektor-gadget/inspektor-gadget/commit/229c772c0e6d49c543c623af11d2a53786b09d49 gives an example of how it's used by a gadget.

PR #1626, opened 2023-05-12
trace/tcp: Use KubeIPResolver

burak-ok:
# trace/tcp: Use KubeIPResolver

See #1568 

# Discussion
To not end in a copy-paste party: How should this be generalized the best way?
I think creating a struct for all things resolved through the `KubeIPResolver` might be the best idea.
And then embed it into the `Event struct` of that gadget.
Furthermore provide a function which creates the virtual columns.

But doing this generalization removes the flexibility.
Maybe some gadget only need part of the `KubeIPResolver`.
Or different ordering of the columns


mauriciovasquezbernal:
> I think creating a struct for all things resolved through the KubeIPResolver might be the best idea.
And then embed it into the Event struct of that gadget.

I agree. I was thinking that something like could work good:
- Extend `types.EndpointDetails` to also contain the IP and the port. Perhaps it can only be called `Endpoint`.
- Embed that into the types structure of different gadgets.
- Remove `GetRemoteIPs` and use instead `GetEndpoints()`. The gadget has to fill the ip address and port and the enricher will fill out the other fields.

> Furthermore provide a function which creates the virtual columns.

Yeah, I think it's good. Perhaps it'll need to define common extractor that takes `types.EndpointDetails`, but I haven't checked the implementation details.

> But doing this generalization removes the flexibility.
Maybe some gadget only need part of the KubeIPResolver.

I think in all the cases converting from an IP address to svc/pod name is good, so I think it's useful for most of the gadgets.

> Or different ordering of the columns

In that case they could define their own columns and not use the generic function to create them.

pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -49,8 +49,8 @@ const (
 type KubeNetworkInformation interface {
 	SetLocalPodDetails(owner, hostIP, podIP string, labels map[string]string)
```
mauriciovasquezbernal, line 50: I'm wondering about splitting this interface. It seems this is only useful for the trace network (for network policy) generation gadget. I see that it's not used by others and it's even missing in tcpretrnas. 

Also, it seems to me `ContainerInfoGetters` is only required for this, so maybe we can change the logic a bit to not require all gadgets to implement `ContainerInfoGetters`, only the ones that have `SetLocalPodDetails`, by doing so we'll be able to enrich the endpoint information from the IP address regardless `ContainerInfoGetters`.

pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -49,8 +49,8 @@ const (
...
 
-	GetRemoteIPs() []string
-	SetEndpointsDetails(endpoints []types.EndpointDetails)
+	GetEndpoints() []types.Endpoint
```
mauriciovasquezbernal, line 52: We can have a single `GetEndpoints() []*types.Endpoint` method, by returning the pointer we allow the enricher to set its fields without having to use `SetEndpoints()` later on.

pkg/types/types.go
```diff
@@ -120,11 +120,35 @@ func (c *CommonData) GetContainer() string {
...
+	PodLabels map[string]string `column:"podLabels,hide"`
+}
+
+func AddVirtualColumn[Event any](cols *columns.Columns[Event], attr columns.Attributes, getEndpoint func(*Event) Endpoint) {
+	err := cols.AddColumn(
+		attr,
+		func(e *Event) string {
+			endpoint := getEndpoint(e)
+			switch endpoint.Kind {
+			case RemoteKindPod:
```
mauriciovasquezbernal, line 139: nit: RemoteKindPod (and others) should be renamed to EndpointKindPod, etc...

pkg/gadgets/trace/network/types/types.go
```diff
@@ -54,23 +47,21 @@ func (e *Event) SetLocalPodDetails(owner, hostIP, podIP string, labels map[strin
...
-	e.RemoteKind = endpoints[0].Kind
+	e.DstEndpoint = endpoints[0]
 }
 
 func GetColumns() *columns.Columns[Event] {
-	cols := columns.MustCreateColumns[Event]()
+	cols := columns.MustCreateColumns[Event](columns.WithPrependEmbeddedNames(true))
 
+	// Can't use eventtypes.AddVirtualColumn here because we don't capture port information and dont' want to display it
 	cols.MustAddColumn(columns.Attributes{
```
mauriciovasquezbernal, line 65: Shouldn't it use `eventtypes.AddVirtualColumn` too?

mauriciovasquezbernal (commented):
I like where it's going. I added some comments but will review with more care tomorrow.

mauriciovasquezbernal (commented):
I checked it again but I don't have other comments besides the ones I did yesterday. Please note that unit tests need to be updates as well.

pkg/columns/columns_test.go
```diff
@@ -442,3 +442,44 @@ func TestVerifyColumnNames(t *testing.T) {
...
+		embeddedStructUnnamed
+		embeddedStructNamed `column:"named"`
+	}
+
+	cols := MustCreateColumns[testStruct](WithPrependEmbeddedNames(true))
+
+	_, found := cols.GetColumn("embeddedStructUnnamed.foo")
+	if found {
+		t.Errorf("Expected column embeddedStructUnnamed.foo to be not found")
+	}
```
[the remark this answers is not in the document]
flyth: This has already been taken care of for this particular file (and also this test) in #1664 this morning :)

pkg/gadgets/trace/network/types/types.go
burak-ok: The printing is here done differently. Here the port is not displayed and not captured at all. The output would always show a port `1.1.1.1:0` or `127.0.0.1:0`

pkg/operators/kubeipresolver/kubeipresolver.go
burak-ok: Indeed `trace network` is the only gadget that needs that interface currently. I thought that we could create another operator just for gadgets that want to get network enrichment from pod namespace/name.

The cached container and service list which `KubeIPResolver` has can also be shared between the two operators. Or maybe the `KubeIPResolver` can check on two interfaces and selectively use them.
1. pod namespace/name -> network information
2. ip -> network information

burak-ok:
I found another issue:
For `ig` the whole mechanism doesn't work and didn't work in the past. Do we want it to work?

Until now the `KubeIPResolver` wasn't even included for `ig`.
Additionally there is a problem when we are creating a k8s client. When calling `ig` with `sudo` the default place to search is at `/root/.kube/config: no such file or directory` and the initialization errors out. That is the desired behavior for `kubectl-gadget`, but for `ig` it doesn't make sense maybe? How could we differentiate in an elegant way

alban:
I think that for `ig`, connecting to the Kubernetes API has to be an opt-in. `--enable-kubernetes` maybe? Then users can specify `--kubeconfig`, `$KUBECONFIG` or use the downward api. How should `ig` know the name of the node where it is running? With gadgettracermanager, it knows via `os.Getenv("NODE_NAME")`. Should it be an explicit parameter with `ig`?

pkg/operators/kubeipresolver/kubeipresolver.go
blanquicet: I agree that `SetLocalPodDetails` should be moved to a new interface that embeds `ContainerInfoGetters`. I would keep `ContainerInfoGetters` separated as it could be helpful for other operators in the future but for the kubeipresolver, ensure it is just required if the gadget also implements `SetLocalPodDetails`. Notice there is no verification right now. That's why tcpretrans works even without implementing it. 

I suggest something like this:

```go
// KubeNetworkInformation is for now a specific interface for `trace network` gadget
type KubeNetworkInformation interface {
	GetEndpoints() []*types.Endpoint
}

type AnotherFancyName interface {
	operators.ContainerInfoGetters
	SetLocalPodDetails(owner, hostIP, podIP string, labels map[string]string)
}
```

```go
func (k *KubeIPResolver) CanOperateOn(gadget gadgets.GadgetDesc) bool {
	km := kubemanager.KubeManager{}
	if !km.CanOperateOn(gadget) {
		return false
	}

	_, hasNetworkInf := gadget.EventPrototype().(KubeNetworkInformation)
	_, hasAnotherFancyInterface := gadget.EventPrototype().(AnotherFancyName)
	return hasNetworkInf || hasAnotherFancyInterface
}
```

```
func (m *KubeIPResolverInstance) enrich(ev any) {
	additionalInfo, enrichKubeNetInfo := ev.(KubeNetworkInformation)
	anotherFancyInfo, enrichAnotherFancyInfo := ev.(AnotherFancyName)

	[...]
	// Modify this function to do things based on enrichKubeNetInfo and
	// enrichAnotherFancyInfo
	[...]
```

If `enrich` becomes too complicated or unreadable, I would suggest dividing it into two functions and calling them from `EnrichEvent` based on `enrichKubeNetInfo` and `enrichAnotherFancyInfo`. Or we can consider having two operators as you suggested.

docs/gadgets/trace/tcp.md
```diff
@@ -13,34 +13,35 @@ connect, accept and close events related to TCP connections.
 First, we need to create one pod:
 
 ```bash
-$ kubectl run busybox --image busybox:latest sleep inf
```
blanquicet, line 16: Please update guides of other gadgets too.

blanquicet (commented):
It looks great!

> I think that for `ig`, connecting to the Kubernetes API has to be an opt-in. `--enable-kubernetes` maybe? Then users can specify `--kubeconfig`, `$KUBECONFIG` or use the downward api.
>
@alban I agree with you. Everything related to Kubernetes must be optional in `ig`. Burak, please consider that those flags should be available only for `ig` and only if the kubeipresolver operator was imported in `ig`. I'm not sure how to do that, maybe @flyth can suggest something.

> How should `ig` know the name of the node where it is running? With gadgettracermanager, it knows via `os.Getenv("NODE_NAME")`. Should it be an explicit parameter with `ig`?
> 
Why `ig` needs to know the node name? In one case, the kubeipresolver operator uses namespace/pod as a source which is unique in the cluster. And in the other, it uses a pod/svc IP, which is also unique. Am I missing something? Maybe you wanted to limit it to resolving only local (within the same node) things?

flyth:
> Burak, please consider that those flags should be available only for ig and only if the kubeipresolver operator was imported in ig. I'm not sure how to do that, maybe @flyth can suggest something.

Hmm, we could use `pkg/environment` for that and add the param dynamically using it.

alban:
> Why ig needs to know the node name? In one case, the kubeipresolver operator uses namespace/pod as a source which is unique in the cluster. And in the other, it uses a pod/svc IP, which is also unique. Am I missing something? Maybe you wanted to limit it to resolving only local (within the same node) things?

You're right. I thought it needed the node name to check the local pods to match the source IP address. But it's not needed.

integration/ig/non-k8s/trace_network_test.go
```diff
@@ -52,10 +55,13 @@ func TestTraceNetwork(t *testing.T) {
...
-					PktType:    "HOST",
-					Proto:      "tcp",
-					Port:       80,
-					RemoteAddr: "127.0.0.1",
+					PktType: "HOST",
+					Proto:   "tcp",
+					Port:    80,
+					DstEndpoint: eventtypes.Endpoint{
+						Addr: "127.0.0.1",
+						Kind: eventtypes.EndpointKindOther,
```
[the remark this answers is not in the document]
burak-ok: Indeed this is weird. What about renaming the `EndpointKindOther` alltogether to something it actually represents: A simple ip(+port) string.
Maybe something like `EndpointKindRaw` or `EndpointKindIP`/`EndpointKindIPPort`

pkg/gadgets/trace/network/types/types.go
burak-ok: Theoretically port 0 is allowed and could be used by someone (Even if it is not seen good). So i guess a pointer would be needed here?

Regarding the difference of `MustAddVirtualEndpointColumn` and the formatting in `trace network`: I left the network formatting untouched so the output wouldn't change at all. But I don't have anything against aligning it

pkg/gadgets/trace/network/types/types.go
```diff
@@ -79,15 +63,15 @@ func GetColumns() *columns.Columns[Event] {
...
+		switch e.DstEndpoint.Kind {
+		case eventtypes.EndpointKindPod:
+			return fmt.Sprintf("pod %s/%s", e.DstEndpoint.Namespace, e.DstEndpoint.Name)
+		case eventtypes.EndpointKindService:
+			return fmt.Sprintf("svc %s/%s", e.DstEndpoint.Namespace, e.DstEndpoint.Name)
+		case eventtypes.EndpointKindOther:
+			return fmt.Sprintf("endpoint %s", e.DstEndpoint.Addr)
 		default:
-			return e.RemoteAddr
+			return e.DstEndpoint.Addr
```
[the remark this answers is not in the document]
burak-ok: Correct. The `default` part here will be used for `ig` without any k8s info currently

pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -226,72 +104,51 @@ func (m *KubeIPResolverInstance) PostGadgetRun() error {
 }
 
 func (m *KubeIPResolverInstance) enrich(ev any) {
-	additionalInfo, _ := ev.(KubeNetworkInformation)
```
blanquicet, line 229: BTW, it's unrelated to this PR, but could you please update `KubeIPResolverInstance.Name()`?

pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -226,72 +104,51 @@ func (m *KubeIPResolverInstance) PostGadgetRun() error {
...
 	}
-	if foundRemote == len(remoteIPs) {
-		additionalInfo.SetEndpointsDetails(endpoints)
+	if foundRemote == len(endpoints) {
 		return
 	}
 
 	svcs := m.manager.k8sInventory.GetSvcs()
 
 	for _, svc := range svcs.Items {
```
blanquicet, line 142: I think also here we should add a check to avoid iterating over all services:

```go
func (m *KubeIPResolverInstance) enrich(ev any) {
	additionalInfo, _ := ev.(KubeIPResolverInterface)
	endpoints := additionalInfo.GetEndpoints()
	for j := range endpoints {
		// initialize to this default value if we don't find a match
		endpoints[j].Kind = types.EndpointKindRaw
	}

	found := 0
	pods := m.manager.k8sInventory.GetPods()
	for _, pod := range pods.Items {
		if pod.Spec.HostNetwork {
			continue
		}

		for j, endpoint := range endpoints {
			if pod.Status.PodIP == endpoint.Addr {
				endpoints[j].Kind = types.EndpointKindPod
				endpoints[j].Name = pod.Name
				endpoints[j].Namespace = pod.Namespace
				endpoints[j].PodLabels = pod.Labels

				found++
				if found == len(endpoints) {
					break
				}
			}
		}
	}

	svcs := m.manager.k8sInventory.GetSvcs()
	for _, svc := range svcs.Items {
		for j, endpoint := range endpoints {
			if svc.Spec.ClusterIP == endpoint.Addr {
				endpoints[j].Kind = types.EndpointKindService
				endpoints[j].Name = svc.Name
				endpoints[j].Namespace = svc.Namespace
				endpoints[j].PodLabels = svc.Labels

				found++
				if found == len(endpoints) {
					break
				}
			}
		}
	}
}
```

pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -21,22 +21,9 @@
 package kubeipresolver
```
blanquicet, line 21: Please update the list of gadgets using it. Or make it more generic not to need to update it every time a new gadget start using it.

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -0,0 +1,145 @@
...
+	return []string{kubemanager.OperatorName}
+}
+
+func (k *KubeNameResolver) CanOperateOn(gadget gadgets.GadgetDesc) bool {
+	km := kubemanager.KubeManager{}
+	if !km.CanOperateOn(gadget) {
+		return false
+	}
+	_, hasNameResolverInterface := gadget.EventPrototype().(KubeNameResolverInterface)
+	return hasNameResolverInterface
```
blanquicet, line 71: I know it was already like that in the KubeIPResolver, but I can't remember why we need to ensure that the `KubeManager` can operate on the gadget instead of simply doing:

```suggestion
	_, hasContainerInfoGetters := gadget.EventPrototype().(operators.ContainerInfoGetters)
	_, hasNameResolverInterface := gadget.EventPrototype().(KubeNameResolverInterface)
	return hasNameResolverInterface && hasContainerInfoGetters
```

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (k *KubeNameResolver) Init(params *params.Params) error {
+	log.Warnf("Init %s", OperatorName)
```
blanquicet, line 75: Debug?

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -0,0 +1,145 @@
...
+	return nil
+}
+
+func (k *KubeNameResolver) Close() error {
+	k.k8sInventory.Close()
+	return nil
+}
+
+func (k *KubeNameResolver) Instantiate(gadgetCtx operators.GadgetContext, gadgetInstance any, params *params.Params) (operators.OperatorInstance, error) {
+	log.Warnf("Instantiate %s", OperatorName)
```
blanquicet, line 87: dito

In particular, for the instance you have the gadget context, you could do the following:

```suggestion
	gadgetCtx.Logger().Debugf("Instantiating %s", OperatorName)
```

BTW, We need a way to pass the logger to the operator as well. cc @flyth

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -0,0 +1,145 @@
...
+			// shorter name without the random suffix. That will be used to
+			// generate the network policy name.
+			if pods.Items[i].OwnerReferences != nil {
+				nameItems := strings.Split(pods.Items[i].Name, "-")
+				if len(nameItems) > 2 {
+					owner = strings.Join(nameItems[:len(nameItems)-2], "-")
+				}
+			}
+			additionalInfo.SetLocalPodDetails(owner, pod.Status.HostIP, pod.Status.PodIP, pod.Labels)
+			break
```
blanquicet, line 133: `return`?

pkg/gadget-collection/gadgets/trace/network/gadget.go
```diff
@@ -115,14 +116,20 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
...
 	// - delete this file pkg/gadget-collection/gadgets/trace/network/gadget.go
-	op := operators.GetRaw(kubeipresolver.OperatorName).(*kubeipresolver.KubeIPResolver)
-	op.Init(nil)
-	inst, err := op.Instantiate(nil, nil, nil)
+	kubeIPOp := operators.GetRaw(kubeipresolver.OperatorName).(*kubeipresolver.KubeIPResolver)
+	kubeIPOp.Init(nil)
+	kubeIPInst, err := kubeIPOp.Instantiate(nil, nil, nil)
 	if err == nil {
-		inst.PreGadgetRun()
+		kubeIPInst.PreGadgetRun()
```
blanquicet, line 126: Missing call to `PostGadgetRun()` when stopping the tracer.

pkg/gadgets/trace/tcp/types/types.go
```diff
@@ -29,10 +29,17 @@ type Event struct {
...
-	Saddr     string `json:"saddr,omitempty" column:"saddr,template:ipaddr"`
-	Daddr     string `json:"daddr,omitempty" column:"daddr,template:ipaddr"`
-	Sport     uint16 `json:"sport,omitempty" column:"sport,template:ipport"`
-	Dport     uint16 `json:"dport,omitempty" column:"dport,template:ipport"`
+
+	SrcEndpoint eventtypes.Endpoint `json:"src,omitempty" column:"src"`
+	DstEndpoint eventtypes.Endpoint `json:"dst,omitempty" column:"dst"`
+}
+
+func (e *Event) SetLocalPodDetails(owner, hostIP, podIP string, labels map[string]string) {
```
blanquicet, line 37: No longer needed.

pkg/gadgets/trace/tcpdrop/types/types.go
```diff
@@ -34,90 +32,43 @@ type Event struct {
...
-	/* Destination IP resolved by kubeipresolver  */
-	DstKind      eventtypes.RemoteKind `json:"dstKind,omitempty" column:"dstKind,maxWidth:5,hide,order:3100"`
-	DstNamespace string                `json:"dstNamespace,omitempty" column:"dstns,hide,order:3101"`
-	DstName      string                `json:"dstName,omitempty" column:"dstname,hide,order:3102"`
+	SrcEndpoint eventtypes.Endpoint `json:"src,omitempty" column:"src"`
+	DstEndpoint eventtypes.Endpoint `json:"dst,omitempty" column:"dst"`
 }
 
 func (e *Event) SetLocalPodDetails(owner, hostIP, podIP string, labels map[string]string) {
 	// Unused
```
blanquicet, line 44: No longer needed.

docs/gadgets/trace/tcp.md
blanquicet: But the columns' names are different now: `DST.X` and `SRC.Y`.

blanquicet (commented):
Just a few comments, but it's looking good. After managing these comments, I think we can merge it.

integration/inspektor-gadget/trace_network_test.go
```diff
@@ -53,20 +53,22 @@ func TestTraceNetwork(t *testing.T) {
```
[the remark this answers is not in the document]
burak-ok: Just did that. And some other modifications for that gadget was needed aswell

pkg/operators/kubenameresolver/kubenameresolver.go
burak-ok: Reason is that we need `KubeManager` to work at all. That is the operator which sets the information which can be retrieved by `ContainerInfoGetters`. This way we always stay aligned with `KubeManager`

pkg/operators/kubenameresolver/kubenameresolver.go
burak-ok: Yes, debug. Warn was the debug for me during development :D

pkg/operators/kubenameresolver/kubenameresolver.go
flyth: Everything not running in the instance should use a server side logger - I think for now we could use `logger.DefaultLogger()`, but it would be great if we could somehow pass around an already initialized logger, maybe even with prefixes / additional fields (for the operator using it in this case).

/cc @mqasimsarfraz as he wanted to improve the logger as well :)

pkg/operators/kubenameresolver/kubenameresolver.go
blanquicet: > Afaik, LocalManager will soon also be able to enrich Kubernetes metadata as well, right?
>
Actually, the LocalManager is already enriching with the k8s metadata. The problem is that the container name is not the one from k8s but from the container runtime. https://github.com/inspektor-gadget/inspektor-gadget/pull/1702 uses the correct container name (btw, reviews are welcome there). 

> Importing `KubeManager` here directly installs it as operator
>
It's already done in the gadget-service, so it shouldn't be necessary here. Or did I miss/misunderstand something?

> That is the operator which sets the information which can be retrieved by `ContainerInfoGetters`. This way we always stay aligned with `KubeManager`
>
IMO, this is something the `KubeNameResolverInstance` shouldn't care about. I mean, in practice, it just requires an event from which it can extract the namespace and the pod. It doesn't care how such information was added to the event. I think we `CanOperateOn` should be:

```go
func (k *KubeNameResolver) CanOperateOn(gadget gadgets.GadgetDesc) bool {
	_, hasContainerInfoGetters := gadget.EventPrototype().(operators.ContainerInfoGetters)
	_, hasNameResolverInterface := gadget.EventPrototype().(KubeNameResolverInterface)
	return hasNameResolverInterface && hasContainerInfoGetters
}
```

At most, we can do this:

```go
func (m *KubeNameResolver) enrich(ev any) {
	additionalInfo, _ := ev.(KubeNameResolverInterface)
	containerInfo, _ := ev.(operators.ContainerInfoGetters)

	if containerInfo.GetNamespace() == "" || containerInfo.GetPod() == "" {
		log.Debugf("Impossible to enrich event with empty namespace/pod info")
		return
	}

	[...]
	}
```

And the `Dependencies` shouldn't be a specific operator but something like "I need an operator that enriches the event with this data" ... No idea how to do it, though 🤣.

Anyway, we can manage this in another PR.

burak-ok: > > That is the operator which sets the information which can be retrieved by `ContainerInfoGetters`. This way we always stay aligned with `KubeManager`
> 
> IMO, this is something the `KubeNameResolverInstance` shouldn't care about. I mean, in practice, it just requires an event from which it can extract the namespace and the pod. It doesn't care how such information was added to the event. I think we `CanOperateOn` should be:
> 

I think you are right here. See below

> And the `Dependencies` shouldn't be a specific operator but something like "I need an operator that enriches the event with this data" ... No idea how to do it, though rofl.

Guess this is the hard part :D. Probably can't be done with the current mechanisms

docs/gadgets/trace/tcp.md
burak-ok: Ahhh yes. For `trace tcpdrop` and `trace tcp` I had to update them

pkg/operators/kubeipresolver/kubeipresolver.go
mauriciovasquezbernal: IMO ^^^ that's too much overhead. I'd return the pointer and document what the operator fills.

pkg/gadgets/trace/tcpretrans/types/types.go
```diff
@@ -34,89 +32,38 @@ type Event struct {
...
+		func(e *Event) eventtypes.Endpoint { return e.SrcEndpoint })
+	eventtypes.MustAddVirtualEndpointColumn(
+		cols,
+		columns.Attributes{
+			Name:    "dst",
+			Visible: true,
+			Width:   30,
+			Order:   3000,
+		},
+		func(e *Event) eventtypes.Endpoint { return e.DstEndpoint })
```
mauriciovasquezbernal, line 66: I think Jose made a similar comment before, not sure. I strongly think the last `)` should go in a new line to make this more readable.

```go
	eventtypes.MustAddVirtualEndpointColumn(
		tcpColumns,
		columns.Attributes{
			Name:    "src",
			Visible: true,
			Width:   30,
			Order:   2000,
		},
		func(e *Event) eventtypes.Endpoint { return e.SrcEndpoint },
	)
```

pkg/gadgets/trace/network/types/types.go
mauriciovasquezbernal: This one is difficult to handle. IMO we should consider the case of port 0 being valid. I also think adding a pointer creates some complications for the user of those types, as they need to check if this is nil before reading and when assigning it has to be saved in a variable in order to be able to get its address.

I think a solution is to split Endpoint into two types: L3 (IP) and L4 (ports). I did exploration in https://github.com/inspektor-gadget/inspektor-gadget/commit/12e216a59509f754c3c7de93ed6d624a180fdde4 and I think the result is good. What do you think?

mauriciovasquezbernal (commented):
It's looking really really good. Only blocking comment is about splitting the Endpoint type. After that I think it's ready to go.

pkg/gadgets/trace/network/types/types.go
burak-ok: Looks nice @mauriciovasquezbernal , thanks. I cherry-picked it

mauriciovasquezbernal (approved):
LGTM. My only comment is about organization of commits.

I'd squash `Generalize network Endpoints` and `types/Endpoint: Split into l3 and l4 types` to avoid introducing changes on the former that are then modified by the later. 

Nice job!

PR #1628, opened 2023-05-15
docs: CONTRIBUTING.md: profiling benchmarks

alban:
The documentation for contributors already explained how to run benchmarks but not how to profile them in order to understand where the time is spent and how the memory is consumed.

This patch adds explanations how to do that with both the bcc profile tool and the Go pprof tools.


mauriciovasquezbernal (approved):
LGTM from a quick inspection. I didn't try it out myself.

PR #1629, opened 2023-05-15
docs: installation on minikube

alban:
Inspektor Gadget is now available as an Addon in minikube: https://minikube.sigs.k8s.io/docs/handbook/addons/inspektor-gadget/

This patch adds this in the documentation as an option in the possible ways to install Inspektor Gadget on minikube.

cc @yolossn


mauriciovasquezbernal (approved):
LGTM!

Issue #1612, opened 2023-05-08
[RFE] Performances: networking gadgets (dns, sni...)


alban:
## Current situation

In the tests with 100 containers on a node, the networking gadgets were
sometimes failing by timeout because they were too slow. Profiling
showed that most of the time is spent in the ebpf verifier. That is
because networking gadgets load the ebpf program once per network
namespace.

## Impact

Networking gadgets could fail on nodes with a lot of containers due to timeout.

## Ideal future situation

Networking gadgets are fast like other gadgets.

## Implementation options

instantiate ebpf prog once for all netns:
https://github.com/inspektor-gadget/inspektor-gadget/pull/1430

PR #1630, opened 2023-05-16
trace network: use same code pattern as other networking gadgets

alban:
# trace network: use same code pattern as other networking gadgets

- Use pkg/gadgets/internal/networktracer package instead of 
  - Use a perf ring buffer like other tracing gadgets instead of a custom hash map polled every second. This can be slower with a lot of new connections but I am hoping we can get the performances back with the Prometheus work.
- Use kubeipresolver instead of its own enricher
- Use socket enricher

First part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1612
Refactoring suggested in https://github.com/inspektor-gadget/inspektor-gadget/pull/1430#issuecomment-1470794513
Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1334
Supersedes https://github.com/inspektor-gadget/inspektor-gadget/pull/1336

Depends on:
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/1636
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/1680

I will rebase this PR once the dependencies are merged.

## How to use

## Testing done


alban:
I am testing this in two ways:

1. With trace network:
```
$ ./kubectl-gadget trace network -o json|grep tcp
{"node":"minikube-containerd","namespace":"default","pod":"shell02","container":"shell02","timestamp":1685466705588526838,"type":"normal","mountnsid":4026533430,"netnsid":4026533256,"pid":255628,"tid":255628,"comm":"nc","pktType":"OUTGOING","proto":"tcp","port":9091,"podHostIP":"192.168.49.2","podIP":"10.244.0.7","podLabels":{"run":"shell02"},"remoteKind":"svc","remoteAddr":"10.99.179.186","remoteName":"shell","remoteNamespace":"default","remoteLabels":{"run":"shell"}}
```

2. With advise network-policy
```
$ ./kubectl-gadget advise network-policy monitor|grep tcp
{"node":"minikube-containerd","namespace":"default","pod":"shell02","timestamp":1685466740985560531,"type":"normal","mountnsid":4026533430,"netnsid":4026533256,"pid":256061,"tid":256061,"comm":"nc","pktType":"OUTGOING","proto":"tcp","port":9091,"remoteAddr":"10.99.179.186"}
```

As you can see, KubeIPResolver enriches the event correctly with 'trace network' but not with 'advise network-policy'. This is because:
1. 'trace network' uses the operators to enrich the events but 'advise network-policy' still uses pkg/gadget-collection/gadgets/trace/network/gadget.go.
2. In this PR, I no longer pass an enricher interface to netTracer.NewTracer(). And I removed the file pkg/gadget-collection/gadgets/trace/network/enricher.go since it is a duplicate of pkg/operators/kubeipresolver/kubeipresolver.go.

Is there a simple way I could get a reference to the KubeIPResolver from pkg/gadget-collection/gadgets/trace/network/gadget.go in order to enrich the events from there?

Alternatively, cmd/kubectl-gadget/advise/network-policy.go could be rewritten to use the gRPC interface instead of the CR. But I would like to avoid refactoring too much in a single PR.

cc @flyth

alban:
> Is there a simple way I could get a reference to the KubeIPResolver from pkg/gadget-collection/gadgets/trace/network/gadget.go in order to enrich the events from there?
> 
> Alternatively, cmd/kubectl-gadget/advise/network-policy.go could be rewritten to use the gRPC interface instead of the CR. But I would like to avoid refactoring too much in a single PR.

Updated by getting the KubeIPResolver operator with GetRaw(), based on an idea from @flyth.

pkg/gadget-collection/gadgets/trace/network/gadget.go
```diff
@@ -113,6 +115,19 @@ func (t *Trace) Start(trace *gadgetv1alpha1.Trace) {
 		return
 	}
 
+	// TODO: Don't access the operator directly
+	// - update cmd/kubectl-gadget/advise/network-policy.go to use the gRPC interface instead of the CR one
+	// - delete this file pkg/gadget-collection/gadgets/trace/network/gadget.go
+	op := operators.GetRaw(kubeipresolver.OperatorName).(*kubeipresolver.KubeIPResolver)
+	op.Init(nil)
+	var inst operators.OperatorInstance
+	if op != nil {
```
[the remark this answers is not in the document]
alban: Good point. I removed it.

alban:
Rebased and added the uid/gid feature like it was done for other gadgets.

alban:
> Note that this PR worsens the performances. It is partially expected because it sends all events to userspace instead of aggregating results in a hash map. I was planning to reuse the work for Prometheus in ebpf (https://github.com/inspektor-gadget/inspektor-gadget/issues/1598) to make it fast again but reusing the general code pattern instead of having something specific for trace-network.
> 
> However, I am not sure why the benchmarks are so bad because so far the benchmarks don't generate any events but just load+unload the ebpf program

```
$ go test -exec sudo -ldflags="-s=false" -bench='^BenchmarkAllGadgetsWithContainers$/^container1$/trace-network' -run=Benchmark ./internal/benchmarks/... -count 10 | tee patched.bench
$ git checkout main
$ go mod vendor
$ go test -exec sudo -ldflags="-s=false" -bench='^BenchmarkAllGadgetsWithContainers$/^container1$/trace-network' -run=Benchmark ./internal/benchmarks/... -count 10 | tee main.bench
$ benchstat main.bench patched.bench|tee benchstat.log
goos: linux
goarch: amd64
                                                    │  main.bench  │              patched.bench               │
                                                    │    sec/op    │     sec/op      vs base                  │
AllGadgetsWithContainers/container1/trace-network-4   12.43m ± 11%   1220.34m ± 47%  +9715.57% (p=0.000 n=10)
```

This PR decreases the performances of trace network by +9715.57% and bring it to the same "bad" level as trace dns.

This is due to the socket enricher, which was not used in the trace network gadget before the PR:

```
$ sudo /usr/share/bcc/tools/profile -p $(pidof benchmarks.test)
...
    github.com/cilium/ebpf/btf.LoadKernelSpec
    github.com/cilium/ebpf.findTargetInKernel
    github.com/cilium/ebpf.newProgramWithOptions
    github.com/cilium/ebpf.(*collectionLoader).loadProgram
    github.com/cilium/ebpf.(*CollectionSpec).LoadAndAssign.func1
    github.com/cilium/ebpf.assignValues
    github.com/cilium/ebpf.(*CollectionSpec).LoadAndAssign
    github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/internal/socketenricher.(*SocketEnricher).start
    github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/internal/socketenricher.NewSocketEnricher
    github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/internal/networktracer.NewTracer[...]
    github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/network/tracer.(*Tracer).install
    github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/network/tracer.(*Tracer).Init
    github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local.(*Runtime).RunGadget
    github.com/inspektor-gadget/inspektor-gadget/internal/benchmarks.BenchmarkAllGadgetsWithContainers.func1.1
    testing.(*B).runN
    testing.(*B).run1.func1
    runtime.goexit.abi0
...
```

Since the time is mostly spent in the socket enricher's LoadAndAssign and LoadKernelSpec, we can do the following optimization for gadgettracermanager:
* Load the socket enricher when the process starts, but without attaching the kprobes
* When the socket enricher is used by a gadget, initialize it (~RewriteConstants) and attach the kprobes
* When the socket enricher is not used anymore, detach the kprobes but don't release the eBPF programs so it can be reused for the next gadget.

This would reduce the time to load a gadget using the socket enricher by about 1.2 seconds, so I think it is worth it for a better user experience. This is unrelated to this PR though, since it is the same problem in trace dns, sni, tcpdrop, etc.

integration/inspektor-gadget/advise_networkpolicy_test.go
```diff
@@ -44,8 +44,9 @@ func TestAdviseNetworkpolicy(t *testing.T) {
 			Cmd: fmt.Sprintf(`$KUBECTL_GADGET advise network-policy monitor -n %s --output ./networktrace-client.log &
 					sleep 10
 					kill $!
-					head networktrace-client.log | sort | uniq`, nsClient),
-			ExpectedRegexp: fmt.Sprintf(`{"node":".*","namespace":"%s","pod":"test-pod","container":"test-pod","timestamp":.*,"type":"normal","netnsid":.*,"pktType":"OUTGOING","proto":"tcp","port":9090,"podHostIP":".*","podIP":".*","podLabels":{"run":"test-pod"},"remoteKind":"svc","remoteAddr":".*","remoteName":"test-pod","remoteNamespace":"%s","remoteLabels":{"run":"test-pod"}}`, nsClient, nsServer),
```
mauriciovasquezbernal, line 48: I always wanted to update these tests not to use regexp. I did that in https://github.com/inspektor-gadget/inspektor-gadget/commit/d3956b7c8e8339866ff215bea7f94eb93f177e0b. Feel free to pick up that, otherwise I can open a PR with that one too.

docs/crds/gadgets/filetop.md
```diff
@@ -8,7 +8,7 @@ filetop shows reads and writes by file, with container details.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (node,namespace,pod,container,mntns,pid,tid,comm,reads,writes,rbytes,wbytes,T,file). (default -reads,-writes,-rbytes,-wbytes)
+ - sort_by: The field to sort the results by (node,namespace,pod,container,hostnetwork,mntns,pid,tid,comm,reads,writes,rbytes,wbytes,T,file). (default -reads,-writes,-rbytes,-wbytes)
```
mauriciovasquezbernal, line 11: It's now called `sort` and the help is something like: `--sort string            Sort by columns. Join multiple columns with ','. Prefix a column with '-' to sort in descending order. (default "-reads,-writes,-rbytes,-wbytes")`

IMO we can remove them from the documentation. It's too difficult to keep that updated.

mauriciovasquezbernal (approved):
Most of my comments are nits. Great work, LGTM!

docs/crds/gadgets/filetop.md
alban: This is the CRD documentation. The text is automatically generated with `make generate-documentation`.

integration/inspektor-gadget/advise_networkpolicy_test.go
alban: Thanks! Please open a separate PR.

PR #1633, opened 2023-05-16
GitHub Action: use shorter names

alban:
Long names are not visible in the web UI, making it impossible to distinguish the different actions.

Example of too long names:

<details>

image

</details>

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -121,7 +121,7 @@ jobs:
         uses: actions/dependency-review-action@v3
 
   build-kubectl-gadget:
-    name: Build kubectl-gadget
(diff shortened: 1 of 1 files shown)


mqasimsarfraz (approved):
LGTM!

.github/workflows/inspektor-gadget.yml
```diff
@@ -222,7 +222,7 @@ jobs:
         path: /home/runner/work/inspektor-gadget/inspektor-gadget/${{ matrix.ig-target }}.tar.gz
 
   build-gadget-container-images:
-    name: Build gadget container images
+    name: gadget img
```
blanquicet, line 225: As agreed for the new naming convention, it should be `ig-k8s img`. Same in scan and publish jobs.

PR #1636, opened 2023-05-17
socket enricher: add socket earlier at tcp_connect

alban:
# socket enricher: add socket earlier at tcp_connect

Before this patch, the first TCP-SYN packet could not be enriched because the socket is added in the sockets map only when tcp_v4_connect() and tcp_v6_connect() returns.

This patch changes this to update the sockets map before executing tcp_v4_connect() and tcp_v6_connect(). If the connection fails, the kretprobe detects it and marks the socket for removal after the timeout.

## How to use

No changes

## Testing done

None yet. I'll test it in https://github.com/inspektor-gadget/inspektor-gadget/pull/1630


mauriciovasquezbernal (approved):
LGTM. 

I suppose creating an unit test for this it too difficult.

alban:
Yes, the tests will have to wait for #1630 because the current networking gadgets (dns or sni) are not sending events for the TCP SYN packet.

PR #1648, opened 2023-05-18
Don't use hostNetwork=true

alban:
hostNetwork=true was initially necessary for the traceloop gadget. But since it was refactored, it is not necessary anymore.

This patch changes the gadget DaemonSet to use hostNetwork=false.

Since the gadget pod is no longer running in the host network namespace, changes the code to get the inode of the host network namespace with `$HOST_ROOT/proc/1/ns/net`.

We cannot use `$HOST_ROOT/proc/$(os.Getpid)/ns/net` anymore because:
- gadgettracermanager might not be in the host network namespace
- os.Getpid() will not return a valid pid in the host pid namespace.

Reading `$HOST_ROOT/proc/1/ns/net` requires `CAP_SYS_PTRACE` if pid 1 runs as a different uid than Inspektor Gadget. We should avoid that from an init() function because it makes 'make generate-documentation' fail.

Part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1658

### DIFF
--- a/examples/kube-container-collection/deploy.yaml
+++ b/examples/kube-container-collection/deploy.yaml
@@ -60,7 +60,7 @@ spec:
     spec:
       serviceAccount: gadget-container-collection
       hostPID: true
-      hostNetwork: true
+      hostNetwork: false
       containers:
       - name: gadget
         terminationMessagePolicy: FallbackToLogsOnError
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal (approved):
This is great. Not having to run Inspektor Gadget on the host network open a lot of new possibilities, like exposing its API on TCP. (It was not possible before because we didn't have a way to protect this API, now we can achieve it by using network policies). 

Please handle the comments by @eiffel-fl before merging.

PR #1656, opened 2023-05-19
Registry/Columns: Read from file + new `bytes` type

burak-ok:
Sometimes a parameter can become long or one wants to set the whole content of a file as the parameter. For this use case one can now specify the filepath if a `@` is used as a prefix 

Example: `echo docker > test.txt && sudo ./ig -r @./test.txt snapshot process`

Furthermore the `bytes` type was introduced. For best transportation it is compressed and then b64 encoded (because of CR)

### DIFF
--- a/cmd/common/params.go
+++ b/cmd/common/params.go
@@ -1,0 +1,42 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+const FilePrefix = "@"
+
+// Param is a wrapper around params.Param. It's used to implement the logic that reads parameters from files.
+type Param struct {
+	*params.Param
+}
+
+func (p *Param) Set(val string) error {
+	if strings.HasPrefix(val, FilePrefix) {
+		filepath := strings.TrimPrefix(val, FilePrefix)
+		data, err := os.ReadFile(filepath)
+		if err != nil {
+			return fmt.Errorf("reading file %q for parameter %q: %w", filepath, p.Key, err)
+		}
+		return p.Param.Set(string(data))
+	}
+	return p.Param.Set(val)
+}
--- a/cmd/common/params_test.go
+++ b/cmd/common/params_test.go
@@ -1,0 +1,45 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+func TestHandleFileArguments(t *testing.T) {
+	t.Parallel()
+
+	pd := &params.ParamDesc{
+		Key:      "test",
+		TypeHint: "string",
+	}
+	p := Param{
+		Param: pd.ToParam(),
+	}
+
+	const fileContent = "This is a magic String. 123$$##@!"
+	const filename = "./file_param_testfile"
+	require.Nil(t, os.WriteFile(filename, []byte(fileContent), 0o600))
+	defer os.Remove(filename)
+
+	err := p.Set(FilePrefix + filename)
+	require.Nil(t, err, "Error while handling file argument")
+	require.Equal(t, p.AsString(), fileContent)
+}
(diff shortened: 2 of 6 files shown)


pkg/params/params.go
```diff
@@ -238,16 +242,60 @@ func (p *Params) ValidateStringMap(cfg map[string]string) error {
...
+		return "" // fmt.Errorf("failed to copy file %s to zlib.Writer: %w", path, err)
+	}
+	// Close the zlib.Writer to ensure that all data has been written
+	if err := w.Close(); err != nil {
+		return "" // fmt.Errorf("failed to close zlib.Writer: %w", err)
+	}
+	return base64.StdEncoding.EncodeToString(b.Bytes())
+}
+
+func b64DecodeAndDecompress(s string) []byte {
```
flyth, line 261: IMHO, this should return an error, if decompression fails (and so should `CopyFromMap` below.

cmd/common/params.go
```diff
@@ -0,0 +1,41 @@
...
+type Param struct {
+	*params.Param
+}
+
+func (p *Param) Set(val string) error {
+	if strings.HasPrefix(val, FilePrefix) {
+		filepath := strings.TrimPrefix(val, FilePrefix)
+		data, err := os.ReadFile(filepath)
+		if err != nil {
+			return fmt.Errorf("failed to open file %q for parameter %q: %w", filepath, p.Key, err)
```
flyth, line 36:
```suggestion
			return fmt.Errorf("reading file %q for parameter %q: %w", filepath, p.Key, err)
```

pkg/params/params.go
```diff
@@ -238,16 +242,60 @@ func (p *Params) ValidateStringMap(cfg map[string]string) error {
...
 }
 
+func compressAndB64Encode(s string) string {
+	// Create a new zlib.Writer, which will write to a bytes.Buffer
+	var b bytes.Buffer
+	w := zlib.NewWriter(&b)
+
+	// Write the contents of the file to the zlib.Writer
+	if _, err := io.Copy(w, strings.NewReader(s)); err != nil {
+		return "" // fmt.Errorf("failed to copy file %s to zlib.Writer: %w", path, err)
```
flyth, line 252: nit: remove these comments

I don't think these errors are likely to occur (other than on hardware failures or oom situations), but in that case I'd prefer a panic to at least tell you that there's something very bad happening. But that might just be my preference.

cmd/common/params_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+func TestHandleFileArguments(t *testing.T) {
+	t.Parallel()
+
+	pd := &params.ParamDesc{
+		Key:      "test",
+		TypeHint: "string",
```
flyth, line 30: Shouldn't this be `TypeBytes`?

flyth (commented):
Looks good! Maybe we should switch to `[]byte` as the underlying type for the values, but that can be done later, if necessary. Thanks!

cmd/common/params_test.go
flyth: Ah, this isn't also checking the base64 compression/decompression, sorry. Maybe there should be some test cases for that as well, though.

pkg/params/params.go
burak-ok: Agree, in case of hardware failure or oom there are other huge problems/indicators

cmd/common/params_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	p := Param{
+		Param: pd.ToParam(),
+	}
+
+	expected := "This is a magic String. 123$$##@!"
+	err := p.Set(FilePrefix + "./params_test.go")
+	assert.Nil(t, err, "Error while handling file argument")
```
mauriciovasquezbernal, line 39: Use `require` instead of `assert`? (if err is not nil it doesn't make sense to continue with the test)

cmd/common/params_test.go
```diff
@@ -0,0 +1,41 @@
...
+		Key:      "test",
+		TypeHint: "string",
+	}
+
+	p := Param{
+		Param: pd.ToParam(),
+	}
+
+	expected := "This is a magic String. 123$$##@!"
+	err := p.Set(FilePrefix + "./params_test.go")
```
mauriciovasquezbernal, line 38: I'd rather create a temp file with expected and check that it fully matches.

cmd/common/params.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+const FilePrefix = "@"
+
```
mauriciovasquezbernal, line 26: What about a short comment indicating why this wrapper is used?

```
// Param is a wrapper aroun params.Param. It's used to implement the logic that reads parameters from files. 
```

pkg/params/params.go
```diff
@@ -363,6 +363,10 @@ func (p *Param) AsString() string {
 	return p.value
 }
 
+func (p *Param) AsBytes() []byte {
```
mauriciovasquezbernal, line 414: Please add a test for it in https://github.com/inspektor-gadget/inspektor-gadget/blob/cd4bc3599a8d389f73cea4498d97ec822beab897/pkg/params/params_test.go#L25.

pkg/params/params.go
mauriciovasquezbernal: I agree to add a panic. It's better not to hide this error.

mauriciovasquezbernal (commented):
Looks very good. Some nits around.

cmd/common/params_test.go
burak-ok: I thought that `assert` would also do that. Thanks for the info

pkg/params/params.go
```diff
@@ -238,16 +242,64 @@ func (p *Params) ValidateStringMap(cfg map[string]string) error {
...
+		return nil, fmt.Errorf("reading from zlib reader:: %w", err)
+	}
+	return bytes, nil
+}
+
 func (p *Params) CopyToMap(target map[string]string, prefix string) {
 	for _, param := range *p {
-		target[prefix+param.Key] = param.String()
+		if param.TypeHint == TypeBytes {
+			target[prefix+param.Key] = compressAndB64Encode(param.String())
```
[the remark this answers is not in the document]
flyth: I think having `TypeBytes` is enough - this signals that the value contains binary data that has to be handled as such. So whatever is handling/transporting the data should take care of making it safe to use. The `CopyFromMap`/`CopyToMap` is only used when transferring the parameter values remotely, so I think this is fine for now. Maybe later on we could optimize it in a way that compression+encoding is only used for CRs, while compression only could still be used for gRPC when the gRPC connection itself doesn't have compression enabled (which is something we could think about as well).

Issue #1658, opened 2023-05-22
[RFE] Avoid using host namespaces (hostNetwork=true and hostPID=true)


alban:
## Current situation

Inspektor Gadget is deployed with:
```
hostNetwork: true
hostPID: true
```

## Impact

* Inspektor Gadget uses more privileges than necessary.
* Users hesitate to deploy Inspektor Gadget due to those privileges
* Developers hesitate before using IG's Go packages because of those privileges

## Ideal future situation

1. Inspektor Gadget is deployed with:
```
hostNetwork: false
hostPID: false
```

2. IG's Go packages can be used by third-party projects without those privileges

## Implementation

Work in progress implementation: https://github.com/inspektor-gadget/inspektor-gadget/pull/1665

1. hostNetwork=false: https://github.com/inspektor-gadget/inspektor-gadget/pull/1648
2. hostPID=false
  a. Various packages read /proc without using /host/proc: https://github.com/inspektor-gadget/inspektor-gadget/pull/1659
  b. runcfanotify uses pidfd_open which requires host pidns: https://github.com/inspektor-gadget/inspektor-gadget/pull/1662
  c. https://github.com/inspektor-gadget/inspektor-gadget/pull/1710
  d. https://github.com/inspektor-gadget/inspektor-gadget/pull/1708
  e. https://github.com/inspektor-gadget/inspektor-gadget/pull/1709

## Additional information

I explored the [Kernel Connector API](https://docs.kernel.org/driver-api/connector.html) which can send process exit events through a netlink socket. Unfortunately, it requires the [host user namespaces, the host pid namespace and CAP_NET_ADMIN](https://github.com/torvalds/linux/blob/v6.2/drivers/connector/cn_proc.c#L360-L368), so it does not help this issue.

alban:
## Problem with ebpf iterators and `hostPID=false`

Using `hostPID=false` will break ebpf iterators of type iter/task_file and
iter/task because they work within the current pidns. This would break:

- the 'snapshot process' gadget
- the future 'snapshot files' gadget discussed in https://github.com/inspektor-gadget/inspektor-gadget/pull/1597#issuecomment-1543729214
- the socket enricher
- the 'top ebpf' gadget

This is difficult to workaround because a process cannot reenter a parent
pidns.

The problem between those ebpf iterators and pidns is not limited to
deployments to Kubernetes using `hostPID=false` but also with minikube setups
where the minikube node is in a container (e.g. the docker driver).

[...]

PR #1659, opened 2023-05-22
Access /proc via $HOST_ROOT

alban:
In preparation for https://github.com/inspektor-gadget/inspektor-gadget/issues/1658

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -38,6 +39,7 @@ import (
 	ociannotations "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/oci-annotations"
 	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runcfanotify"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
 )
 
 func enrichContainerWithContainerData(containerData *runtimeclient.ContainerData, container *Container) {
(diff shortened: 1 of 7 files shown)


gadget-container/hooks/oci/main.go
```diff
@@ -48,6 +48,12 @@ func init() {
 	flag.StringVar(&hook, "hook", "", "OCI hook: prestart or poststop")
 }
 
+var hostRoot string
+
+func init() {
+	hostRoot = os.Getenv("HOST_ROOT")
+}
+
```
flyth, line 56: Maybe put this into a (new) common library to avoid code repetition?

pkg/host/host-root.go
mauriciovasquezbernal, line 1: What about putting this under pkg/utils/host. Or perhaps under internal? (I think this is not useful for external users and pkg/host is too generic).

pkg/container-collection/options.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/host"
+
```
mauriciovasquezbernal, line 28: Move this import below with other inspektor-gadget packages.

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -27,6 +27,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/host"
+
```
mauriciovasquezbernal, line 31: ditto.

pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -196,7 +198,7 @@ func getPidMapFromProcFs() (map[uint32][]*types.Process, error) {
 }
 
 func getMemoryUsage(m *ebpf.Map) (uint64, error) {
-	f, err := os.Open(fmt.Sprintf("/proc/self/fdinfo/%d", m.FD()))
+	f, err := os.Open(fmt.Sprintf("%s/proc/self/fdinfo/%d", host.HostRoot, m.FD()))
```
mauriciovasquezbernal, line 201: Some places use `filepath.Join` and others `fmt.Sprintf`. Can we uniform it? (Given that this code is linux specific, can we only use fmt.Sprintf?)

pkg/host/host-root.go
alban: Moved to pkg/utils/host.

pkg/gadgets/top/ebpf/tracer/tracer.go
alban: I switched to filepath.Join, I find it more clear for semantics.

mauriciovasquezbernal (approved):
LGTM!

nit: I'd be nice to have a small unit test for pkg/utils/host/host.go, but it's up to you.

alban:
Thanks! I will merge as soon as I understand the issue behind https://github.com/inspektor-gadget/inspektor-gadget/pull/1665#issuecomment-1561374258.

PR #1664, opened 2023-05-23
pkg/columns: add support for embedded struct handling, optimize raw access

flyth:
This change enhances support for embedded structs (pointer and non-pointer types). Previously, only non-pointer types were supported.

Additionally, embedded struct members can now be prefixed automatically with the name of the parent field to avoid duplicate field names.

```go
type Demo struct {
	Parent *struct {
		Child int `column:"int"`
	} `column:"parent" columnTags:"abc"`
}
```

So in this case, the inner field will be accessible using `parent.int`. Additionally, `columnTags` will now be inherited by members of the sub-structs. In the case above, `parent.int` will also have the columnTag `abc`.

This is in preparation of gadgets that want to embed certain struct types in their event types more than once (like src/destination address information).

I also updated some test files to use the testify package.

/cc @burak-ok

### DIFF
--- a/pkg/columns/columninfo.go
+++ b/pkg/columns/columninfo.go
@@ -1,4 +1,4 @@
-// Copyright 2022 The Inspektor Gadget authors
+// Copyright 2022-2023 The Inspektor Gadget authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -36,12 +36,15 @@ const (
 )
 
 type subField struct {
-	index int
-	isPtr bool
+	index       int     // number of the referenced field inside the struct
+	offset      uintptr // offset of the referenced field inside the struct
+	parentIsPtr bool    // true, if the referenced field is a member of a pointer type
+	isPtr       bool    // true, if the referenced field is a pointer type
 }
 
 type Attributes struct {
-	Name         string                // Name of the column; case-insensitive for most use cases
+	Name         string                // Name of the column; case-insensitive for most use cases; includes inherited prefixes
+	RawName      string                // Name of the columns without inherited prefixes
 	Width        int                   // Width to reserve for this column
 	MinWidth     int                   // MinWidth will be the minimum width this column will be scaled to when using auto-scaling
 	MaxWidth     int                   // MaxWidth will be the maximum width this column will be scaled to when using auto-scaling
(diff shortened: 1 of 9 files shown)


burak-ok:
Regarding the embedded struct handling I think something along like https://github.com/inspektor-gadget/inspektor-gadget/pull/1626/commits/43a4b3e1010261217c9596336de28430d9b7dead 

Changes:
1. The naming of the struct can be changed via the `columns:ColumnNameHere`. So it will print as `ColumnNameHere.foo` instead of `StructNameHere.foo`
2. For embedded struct which do not have a `columns:ColumnNameHere` the members will be embedded directly. So instead of `StructNameHere.foo` it will just be `foo`

What do you think about that? (I think the test from the commit is better than reading the above list :D)

flyth:
Agree with everything you both wrote. I thought overriding the name of the parent would already work, I'll double check on that :)

Having tag inheritance really makes sense for many cases - maybe we'll have to add functionality later on that allows preventing that (would probably be something like "noinherit" in the columns tag).

Regarding making it the default behavior: from a library perspective that feels a bit weird. If you don't specify a column name for a field, the field name will be used as a fallback. But in this case not specifying a column name on a parent would result in not falling back to its name but instead embed without a prefix (so completely different behavior/functionality). But for the project it makes total sense and also feels more intuitive, so I'll change it.

blanquicet (commented):
I tested it with [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/c30854dffa65aa081816cd1db626f7bdcf1484ea) example (with some changes), and fields within the embedded struct did not inherit the tag:

```bash
K8S.NODE                     K8S.PODNAME                 K8S.CONTAINERNAME           RUNTIME.RUNTIMENAME         RUNTIME.CONTAINERID   RUNTIME.CONTAINERNAME       GADGETDATA
Node 1                       Pod 1                       Container 1                 Runtime 1                   1                     Container 1                 Data 1
Node 2                       Pod 2                       Container 2                 Runtime 2                   2                     Container 2                 Data 2
Node 3                       Pod 3                       Container 3                 Runtime 3                   3                     Container 3                 Data 3

GADGETDATA
Data 1
Data 2
Data 3
```

flyth:
> I tested it with [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/c30854dffa65aa081816cd1db626f7bdcf1484ea) example (with some changes), and fields within the embedded struct did not inherit the tag

Thanks! I was just adding a test for it :) Should be working now.

blanquicet:
> > I tested it with [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/c30854dffa65aa081816cd1db626f7bdcf1484ea) example (with some changes), and fields within the embedded struct did not inherit the tag
> 
> Thanks! I was just adding a test for it :) Should be working now.

Working now!

blanquicet (approved):
I didn't go into detail through the code, but I tested all the cases that came to my mind and the unit tests are quite exhaustive, so LGTM!

PR #1665, opened 2023-05-23
Don't use hostNetwork=true or hostPID=true

alban:
# Don't use hostNetwork=true or hostPID=true

Deploy Inspektor Gadget as regular containers without using the host network/pid namespaces.

This PRs includes different PRs to make it work. They should be merged first.
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/1648
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/1659
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/1662

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1658

This uses https://github.com/vishvananda/netns/pull/76

## How to use

No changes.

## Testing done

In progress.

### DIFF
--- a/examples/kube-container-collection/deploy.yaml
+++ b/examples/kube-container-collection/deploy.yaml
@@ -59,7 +59,7 @@ spec:
         k8s-app: gadget-container-collection
     spec:
       serviceAccount: gadget-container-collection
-      hostPID: true
+      hostPID: false
       hostNetwork: false
       containers:
       - name: gadget
--- a/examples/runc-hook/deploy.yaml
+++ b/examples/runc-hook/deploy.yaml
@@ -56,7 +56,7 @@ spec:
         k8s-app: runc-hook
     spec:
       serviceAccount: runc-hook
-      hostPID: true
+      hostPID: false
       hostNetwork: false
       containers:
       - name: gadget
--- a/gadget-container/entrypoint.sh
+++ b/gadget-container/entrypoint.sh
@@ -30,6 +30,21 @@ KERNEL=$(uname -r)
 echo -n "Kernel detected: "
 echo $KERNEL
 
+echo "/proc/self/cgroup"
+cat /proc/self/cgroup
+
+echo "/host/proc/self/cgroup"
+cat /host/proc/self/cgroup
+
+echo "/host/proc/1/cgroup"
+cat /host/proc/1/cgroup
+
+echo /host/proc/self/ns
+ls -l /host/proc/self/ns
+
+echo /proc/self/ns
+ls -l /proc/self/ns
+
 # The gadget-core image does not provide bcc.
 if [ "$GADGET_IMAGE_FLAVOUR" = "default" ] ; then
 	echo -n "bcc detected: "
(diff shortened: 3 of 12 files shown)


alban:
The tests pass both on #1659 and on #1662 individually but they fail once both PR are merged together:

```
$ ./kubectl-gadget snapshot socket -A
NODE                            NAMESPACE                      POD                            PROTOCOL LOCAL                 REMOTE                STATUS      
Error: running gadget: rpc error: code = Unknown desc = running gadget: running gadget: snapshotting sockets in netns 4026535046: no such file or directory
```

I will investigate this.

alban:
The issue is vendor/github.com/vishvananda/netns/netns_linux.go:
```
	return GetFromPath(fmt.Sprintf("/proc/%d/ns/net", pid))
```
So it is not caused by #1659 or #1662, but by this PR when I set `hostPID: false`.

alban:
Fixed by using https://github.com/vishvananda/netns/pull/76

The CI still fail in the traceloop, trace-exec and snapshot-process tests. I notice the following messages in the CI. I don't know if it is the cause.
```
level=error msg="cgroup enricher: failed to get cgroup paths on container f07942096096508bdb828c1ce81f99d782a44059d350cb1833b199582ba8ec3f: cannot find cgroup path in /proc/PID/cgroup"
level=warning msg="failed to get cgroups info of container f07942096096508bdb828c1ce81f99d782a44059d350cb1833b199582ba8ec3f from /proc/3044/cgroup: cannot find cgroup path in /proc/PID/cgroup"
```

This is because the `/host/proc/$pid/cgroup` files look like this on the CI node:
```
$ cat /proc/self/cgroup
0::/
$ cat /host/proc/self/cgroup
0::/
```
This is because the pod is running in a different cgroup namespace (/proc/$pid/ns/cgroup).

Kubernetes pods run in a cgroupns (cgroup namespace), unless it has privileged: true. Source: [cgroup v2 KEP](https://github.com/kubernetes/enhancements/pull/1370) (found via [podman#5341](https://github.com/containers/podman/issues/5341)):

> Privileged pods will still use the host cgroup namespace

See also:
* https://github.com/cri-o/cri-o/blob/main/server/container_create_linux.go#L552-L553
* https://github.com/containerd/containerd/blob/5997a6589686479e0bf7074218bd7bddf09d2d8f/pkg/cri/server/container_create_linux.go#L338

The gadget pod is no longer running with privileged: true  (with https://github.com/kinvolk/inspektor-gadget/pull/348). But it was already the case before this PR, so I don't know if it is related.

alban:
I don't have this cgroupns issue on my laptop (my minikube setup does not use a separate cgroupns) but the integration test for `TestTraceExec` still fail, so the `cgroupns` issue is unrelated.
```
$ INTEGRATION_TESTS_PARAMS="-run TestTraceExec -no-deploy-ig -no-deploy-spo" make integration-tests
--- FAIL: TestTraceExec (22.25s)
```

It seems runcfanotify does not detect containers anymore; it only kind of worked thanks to `--fallback-podinformer`.

alban:
Superseded by https://github.com/inspektor-gadget/inspektor-gadget/pull/1709

PR #1666, opened 2023-05-23
gadgets: Add gid in generated events

mauriciovasquezbernal:
Many gadgets have the uid available but not the gid. Add also this.

gadgets:
 - [X] bind
 - [X] capabilities
 - [ ] dns
 - [X] exec
 - [ ] fsslower
 - [ ] mount
 - [ ] oomkill
 - [X] open
 - [X] signal
 - [ ] sni
 - [ ] tcp
 - [X] tcpconnect
 - [ ] tcpdrop
 - [ ] tcpretrans


mauriciovasquezbernal:
This is ready for review. There are some gadgets that don't implement this support but in those cases it doesn't make sense or that information is not available, so I'd propose to review for the current gadgets and tackle the other ones in the future.

mauriciovasquezbernal:
I can't understand why the unit tests are failing. If I execute the new test I introduced alone it works fine:

```bash
cd /pkg/gadgets/internal/socketenricher
$ go test -exec sudo -run 'TestSocketEnricherBind/tcp_uid_gid' .
ok  	github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/internal/socketenricher	3.379s
```

If I execute it with any other test it fails:

```bash
$ go test -exec sudo -run 'TestSocketEnricherBind/tcp_uid_gid|TestSocketEnricherBind/tcp6' .
--- FAIL: TestSocketEnricherBind (0.00s)
    --- FAIL: TestSocketEnricherBind/tcp_uid_gid (3.63s)
        tracer_test.go:240: Testing if early tracer noticed the event
        tracer_test.go:244: Testing if late tracer noticed the event
        helpers.go:126: Event wasn't captured 
FAIL
FAIL	github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/internal/socketenricher	3.650s
FAIL
```

The same happens with `--parallel=1`. 

I tried for a bit to understand if there any obvious race but it doesn't seem. @alban do you have any idea of what can be going wrong on this case?

alban:
It's an interaction between the way fake containers in the test framework and the initialisation of the socket enricher with the bpf iterator.

Fake containers are just one thread in the process: only one thread had uid/gid 1000, the other threads are still root.

Threads of the same process share the same file descriptor table. So when the bpf iterator of the socket enricher iterates over file descriptors, it only iterates over the group leaders (main threads of the processes). But that might not be the correct thread of the fake container.

The early socket enricher should work but the late socket enricher might enrich the sockets with the wrong thread in the process.

mauriciovasquezbernal:
I was going to comment that it has to be related to the limitation that the late enricher can't distinguish threads but you already pointed that out. I'll think better about what are the implications of this limitation tomorrow.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
```diff
@@ -63,6 +64,7 @@ insert_socket_from_iter(struct sock *sock, struct task_struct *task)
 	struct sockets_value socket_value = {0,};
 	// use given task
 	socket_value.pid_tgid = ((u64)task->tgid) << 32 | task->pid;
+	socket_value.uid_gid = ((u64)task->cred->gid.val) << 32 | task->cred->uid.val;
```
alban, line 67: I think it should be `s/cred/real_cred`:

Here task is not the current task but another task that might be running VFS calls. The VFS code might temporary substitute `task->cred` by other creds during overlayfs copyup. In this case, we want the real creds of the process, not the creds temporarily substituted by VFS overlayfs copyup.

https://github.com/inspektor-gadget/inspektor-gadget/commit/690f74bd11a91fbfda6ba8f8cdbe22f8daee8176#diff-6baf489ee945d6f76e3e02d7755a2a890b548321c57f414fc3af36b47a0ab33aR101-R106

alban (commented):
This looks good. A few comments.

For the socket enricher issues, I think we should just make the tests pass by cleaning up uid/gid in lateNormalize(). We can't test whether a late socket enricher pass the right uid/gid with the fake containers because the information is lost. I think it's ok because we already test this in integration tests.

> mount (mount is for only root)

A regular user could be given CAP_SYS_ADMIN too, even if unusual.

> network (it doesn't event expose pid information)

Right. After refactorisation, it will be possible. This is fine to skip this gadget in this PR.

> oomkill (there wasn't a way to find uid and gid)

I think you could add uid/gid for the [TriggeredPID](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadgets/trace/oomkill.md), even if we can't add it for the KilledPID

pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c
mauriciovasquezbernal: makes sense. Updated!

mauriciovasquezbernal:
Test were failing because of https://github.com/inspektor-gadget/inspektor-gadget/issues/1692.

alban (approved):
LGTM! Just a small comment about the documentation.

About the mount syscall: I couldn't make it work without root. I think it is not important for now.

PR #1668, opened 2023-05-24
Add support for flags and mode for trace open

eiffel-fl:
Hi.

This PR adds support for flags and mode for `trace open`:

```
COMM             PATH                                                                        FLAGS                                                    MODE                 
bash             bar                                                                         [O_WRONLY O_CREAT O_TRUNC]                               rw-rw-rw-            
touch            /etc/ld.so.cache                                                            [O_RDONLY O_CLOEXEC]                                     ---------            
touch            /lib/x86_64-linux-gnu/libc.so.6                                             [O_RDONLY O_CLOEXEC]                                     ---------            
touch            quux                                                                        [O_WRONLY O_CREAT O_NOCTTY O_NONBLOCK]                   rw-rw-rw-            
cat              /etc/ld.so.cache                                                            [O_RDONLY O_CLOEXEC]                                     ---------            
cat              /lib/x86_64-linux-gnu/libc.so.6                                             [O_RDONLY O_CLOEXEC]                                     ---------            
cat              bar                                                                         [O_RDONLY]                                               ---------            
```

Best regards.

### DIFF
--- a/integration/ig/k8s/trace_open_test.go
+++ b/integration/ig/k8s/trace_open_test.go
@@ -40,6 +40,8 @@ func TestTraceOpen(t *testing.T) {
 				Path:  "/dev/null",
 				Uid:   1000,
 				Gid:   1111,
+				Flags: []string{"O_RDONLY"},
+				Mode:  "----------",
 			}
 
 			normalize := func(e *openTypes.Event) {
--- a/integration/inspektor-gadget/trace_open_test.go
+++ b/integration/inspektor-gadget/trace_open_test.go
@@ -42,6 +42,8 @@ func TestTraceOpen(t *testing.T) {
 				Path:  "/dev/null",
 				Uid:   1000,
 				Gid:   1111,
+				Flags: []string{"O_RDONLY"},
+				Mode:  "----------",
 			}
 
 			normalize := func(e *traceopenTypes.Event) {


pkg/gadgets/trace/open/tracer/utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// We first need to deal with the first 3 bits which indicates access mode.
+	switch{
+	case flags&1 == 0:
+		flagsStr = append(flagsStr, "O_RDONLY")
+	case flags&1 == 1:
+		flagsStr = append(flagsStr, "O_WRONLY")
+	case flags&2 == 2:
+		flagsStr = append(flagsStr, "O_RDWR")
+	}
```
flyth, line 45: `case flags&2 == 2:` will never happen, because one of the two cases above always matches, right?

```suggestion
	switch flags&0b11{
	case 0:
		flagsStr = append(flagsStr, "O_RDONLY")
	case 1:
		flagsStr = append(flagsStr, "O_WRONLY")
	case 2:
		flagsStr = append(flagsStr, "O_RDWR")
	}
```

flyth (commented):
Great, that looks really useful!

pkg/gadgets/trace/open/tracer/tracer_test.go
```diff
@@ -134,6 +136,8 @@ func TestOpenTracer(t *testing.T) {
 					Ret:           fd,
 					Err:           0,
 					Path:          "/dev/null",
+					Flags:         []string{"O_RDONLY"},
+					Mode:          "----------",
```
blanquicet, line 140: What do you think about adding tests to test these two fields?

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -74,6 +74,7 @@ int ig_open_e(struct trace_event_raw_sys_enter* ctx)
 		struct args_t args = {};
 		args.fname = (const char *)ctx->args[0];
 		args.flags = (int)ctx->args[1];
+		args.mode = (__u16)ctx->args[2];
```
blanquicet, line 77: What about something like this to avoid having to modify things in two places:

```c
static int handle_enter(const char *fname, int flags, __u16 mode)
{
	struct args_t args = {};
	u64 id = bpf_get_current_pid_tgid();
	/* use kernel terminology here for tgid/pid: */
	u32 tgid = id >> 32;
	u32 pid = id;

	/* store arg info for later lookup */
	if (trace_allowed(tgid, pid)) {
		args.fname = fname;
		args.flags = flags;
		args.mode = mode;
		bpf_map_update_elem(&start, &pid, &args, 0);
	}
	return 0;
}

SEC("tracepoint/syscalls/sys_enter_open")
int ig_open_e(struct trace_event_raw_sys_enter* ctx)
{
	handle_enter((const char *)ctx->args[0], (int)ctx->args[1],
		     (__u16)ctx->args[2]);
}

int ig_openat_e(struct trace_event_raw_sys_enter* ctx)
{
	handle_enter((const char *)ctx->args[1], (int)ctx->args[2],
		     (__u16)ctx->args[3]);
}
```

NOTE: I didn't test it.

blanquicet (approved):
LGTM!

PR #1670, opened 2023-05-25
logs: Print ig warnings only with custom config

### LINKED ISSUE #1537: [RFE] ig: Manage warning messages better for errors with runtimes
## Current situation

Currently, the `--runtime` parameter is optional. If you don't specify the runtime, it will try to check all supported runtimes. In addition, if any of the runtimes get an error retrieving the containers, it shows a warning. 

It means that in a use case where CRI-O is not available and users don't specify the `--runtime` parameter, users will get the following error at every `ig` command they will execute:

```bash
$ sudo ig list-containers
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: failed to list containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory"
RUNTIME    ID                                                               NAME
containerd 4e17427345c9e512ba3f0deb415ba66630ad429fbca842658ce57018b71b2332 calico-kube-controllers
docker     36b6eee2b92c67c6c56e9c7822dfc5bc5d85afd0b06f58fbb8b7e4e46345c8db myc
```

```bash
$ sudo ig trace exec
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: failed to list containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory"
CONTAINER                                                       PID        PPID       COMM             RET ARGS
gadget                                                          134743     134732     gadgettracerman  0   /bin/gadgettracermanager -liveness
gadget                                                          134766     134749     gadgettracerman  0   /bin/gadgettracermanager -liveness
```

```bash
$ sudo ./ig snapshot process
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: failed to list containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory"
CONTAINER                                                                                                                                                COMM             PID
myc                                                                                                                                                      bash             129380
gadget                                                                                                                                                   gadgettracerman  3080
calico-kube-controllers                                                                                                                                  kube-controller  5208
calico-node                                                                                                                                              bird             4006
calico-node                                                                                                                                              bird6            4005
[...]
```

## Impact

It is annoying for users to get that warning message at every execution.

## Ideal future situation

One possible solution is to hide those warning messages by default unless the user wants --verbose, or nothing is found, and we want to hint that users might want to specify the runtime path.

NOTE: No matter the chosen solution, it must provide a clear message when there is an error getting the containers of the runtime users requested.

```bash
$ ig list-containers
RUNTIME    ID                                                               NAME
containerd 4e17427345c9e512ba3f0deb415ba66630ad429fbca842658ce57018b71b2332 calico-kube-controllers
docker     36b6eee2b92c67c6c56e9c7822dfc5bc5d85afd0b06f58fbb8b7e4e46345c8db myc
```

```bash
$ ig list-containers -r containerd
RUNTIME    ID                                                               NAME
containerd 4e17427345c9e512ba3f0deb415ba66630ad429fbca842658ce57018b71b2332 calico-kube-controllers
```

```bash
$ ig list-containers -r docker
RUNTIME    ID                                                               NAME
docker     36b6eee2b92c67c6c56e9c7822dfc5bc5d85afd0b06f58fbb8b7e4e46345c8db myc
```

```bash
$ ig list-containers -r cri-o
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: failed to list containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory"
```

mqasimsarfraz:
We want to disable container runtime warnings in case user is running with default container runtime configuration to improve the user experience. Also, I don't want to disable warnings by default and allow it to be configured via an option e.g we should keep these warnings for `gadgettracermanager` 

Fixes #1537 

### before the PR
```bash
→ sudo ./ig trace open
WARN[0000] Failed to determine CRI API version: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory" using v1 
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: listing containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory" 
```
### after the PR

```
# default config
→ sudo ./ig trace open 
CONTAINER                                                                                  PID        COMM             FD        ERR PATH                                                                                          

# custom runtime name
→ sudo ./ig trace open -r cri-o
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: listing containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/crio/crio.sock: connect: no such file or directory" 
CONTAINER                                                                                  PID        COMM             FD        ERR PATH   

# custom socket path
→ sudo ./ig trace open  --crio-socketpath foo.sock
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: listing containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix foo.sock: connect: no such file or directory" 
CONTAINER                                                                                  PID        COMM             FD        ERR PATH  
```

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -87,6 +87,14 @@ func containerRuntimeEnricher(
 	return true
 }
 
+// WithDisableContainerRuntimeWarnings disables the warnings about container runtime.
+func WithDisableContainerRuntimeWarnings() ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.disableContainerRuntimeWarnings = true
+		return nil
+	}
+}
+
 // WithMultipleContainerRuntimesEnrichment is a wrapper for
 // WithContainerRuntimeEnrichment() to allow caller to add multiple runtimes in
 // one single call.
@@ -124,8 +132,10 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
 	return func(cc *ContainerCollection) error {
 		runtimeClient, err := containerutils.NewContainerRuntimeClient(runtime)
 		if err != nil {
-			log.Warnf("Runtime enricher (%s): failed to initialize container runtime: %s",
-				runtime.Name, err)
+			if !cc.disableContainerRuntimeWarnings {
+				log.Warnf("Runtime enricher (%s): failed to initialize container runtime: %s",
+					runtime.Name, err)
+			}
 			return err
 		}
 
@@ -154,9 +164,10 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
 		// Enrich already running containers
 		containers, err := runtimeClient.GetContainers()
 		if err != nil {
-			log.Warnf("Runtime enricher (%s): couldn't get current containers: %s",
-				runtime.Name, err)
-
+			if !cc.disableContainerRuntimeWarnings {
+				log.Warnf("Runtime enricher (%s): couldn't get current containers: %s",
+					runtime.Name, err)
+			}
 			return nil
 		}
 		for _, container := range containers {
--- a/pkg/container-collection/container-collection.go
+++ b/pkg/container-collection/container-collection.go
@@ -70,6 +70,9 @@ type ContainerCollection struct {
 
 	// functions to be called on Close()
 	cleanUpFuncs []func()
+
+	// disableContainerRuntimeWarnings is used to disable warnings about container runtimes.
+	disableContainerRuntimeWarnings bool
 }
 
 // ContainerCollectionOption are options to pass to
--- a/pkg/container-utils/cri/cri.go
+++ b/pkg/container-utils/cri/cri.go
@@ -75,7 +75,7 @@ func NewCRIClient(name, socketPath string, timeout time.Duration) (CRIClient, er
 		cc.client = nil
 		cc.clientV1alpha2 = runtimeV1alpha2.NewRuntimeServiceClient(conn)
 	} else {
-		log.Warnf("Failed to determine CRI API version: %v using v1", err)
+		log.Debugf("Failed to determine CRI API version: %v using v1", err)
 	}
 
 	return cc, nil
--- a/pkg/ig-manager/ig-manager.go
+++ b/pkg/ig-manager/ig-manager.go
@@ -19,9 +19,11 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/rlimit"
+	log "github.com/sirupsen/logrus"
 
 	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
 	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
+	runtimeclient "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/runtime-client"
 	containersmap "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgettracermanager/containers-map"
 	tracercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/tracer-collection"
 )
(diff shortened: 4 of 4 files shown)


mauriciovasquezbernal:
I was thinking something different for this one. What about only printing those warnings when the `--runtimes` flag is specified? In that way, we don't print any warning when the user just runs `ig foo bar`, but we print those when the user specifies any custom configuration. I did some experimentation in https://github.com/inspektor-gadget/inspektor-gadget/commit/11c53976dd322f54b8037770581a1e9041f95af9. 

My point is, those messages are annoying, I don't think any user will have all the runtimes available, so there will always we a warning, but at the same time, we shouldn't disable them completely to avoid hiding the error.

mqasimsarfraz:
> I was thinking something different for this one. What about only printing those warnings when the `--runtimes` flag is specified? In that way, we don't print any warning when the user just runs `ig foo bar`, but we print those when the user specifies any custom configuration. I did some experimentation in [11c5397](https://github.com/inspektor-gadget/inspektor-gadget/commit/11c53976dd322f54b8037770581a1e9041f95af9).
> 

The only reason I wasn't sure about hiding warnings was since by default we pass in all the `--runtimes` so as a user I would expect warnings in case a certain runtime has issues. But then again we don't expect users to have all the runtimes to so sounds good if we only enable warnings with `--verbose` or custom `--runtimes`. I will update the PR accordingly and thanks the [experiment](https://github.com/inspektor-gadget/inspektor-gadget/commit/11c53976dd322f54b8037770581a1e9041f95af9).

> My point is, those messages are annoying, I don't think any user will have all the runtimes available, so there will always we a warning, but at the same time, we shouldn't disable them completely to avoid hiding the error.

+1

mauriciovasquezbernal:
Please take into consideration that specifying any runtime endpoint different than the default one should enable the warnings as well.

pkg/container-collection/options.go
```diff
@@ -124,8 +132,10 @@ func WithContainerRuntimeEnrichment(runtime *containerutils.RuntimeConfig) Conta
 	return func(cc *ContainerCollection) error {
 		runtimeClient, err := containerutils.NewContainerRuntimeClient(runtime)
 		if err != nil {
-			log.Warnf("Runtime enricher (%s): failed to initialize container runtime: %s",
-				runtime.Name, err)
+			if !cc.disableContainerRuntimeWarnings || log.IsLevelEnabled(log.DebugLevel) {
```
mauriciovasquezbernal, line 135: What about moving the `IsLevelEnabled` check to ig-manager.go
```
	if !log.IsLevelEnabled(log.DebugLevel) && isDefaultContainerRuntimeConfig(runtimes) {
		warnings := []containercollection.ContainerCollectionOption{containercollection.WithDisableContainerRuntimeWarnings()}
		opts = append(warnings, opts...)
	}
```

mauriciovasquezbernal (approved):
Just one minor comment about log level check logic. LGTM otherwise.

pkg/container-collection/options.go
mqasimsarfraz: Done. Looks better like that!

mqasimsarfraz:
Thanks for all the feedback!

Issue #1640, opened 2023-05-17
`ig`: Provide container image


mauriciovasquezbernal:
We provide the `ig` binary and different packages to install it as release artifacts. However, we don't have any container image with this available. It can be useful for some folks that don't want to download a binary but just execute `docker run ...` to get it on their systems. 

The scope of this issue is: 
- [x] Build and publish `ig` as a container image. (If we can make it a distroless one, that'd be great)
- [x] Document how to run this image
  - [x] By using docker (--privileged or caps, required mounts, etc)

mauriciovasquezbernal:
We are missing pushing this to the container repo and documenting how to use it.

alban:
#1787 addresses part 1 of this issue.

mauriciovasquezbernal:
Closing as https://github.com/inspektor-gadget/inspektor-gadget/pull/1809 is merged.

PR #1674, opened 2023-05-26
Remove `ig-builder.Dockerfile`

eiffel-fl:
Hi.

Following up #1554, this PR modifies the Dockerfile used to build `ig` integration test to make it cross build aware.
As a consequence, `ig-builder.Dockerfile` is no more needed, hence removed.

Best regards.


alban:
Will you be working on https://github.com/inspektor-gadget/inspektor-gadget/issues/1640 after that?

I assumed that Dockerfiles/ig-builder.Dockerfile would serve as a base to build the ig container image, but since it is removed here, will you create a new ig.Dockerfile?

Currently, `Dockerfiles/ig.Dockerfile` starts with `FROM ghcr.io/inspektor-gadget/ig-builder:latest`, but if `Dockerfiles/ig-builder.Dockerfile` is removed, something needs to be updated.

Issue #1340, opened 2023-02-21
snapshot/socket: Support IPv6


mauriciovasquezbernal:
Support for IPv6 was never implemented in this gadget: https://github.com/inspektor-gadget/inspektor-gadget/pull/227.

PR #1680, opened 2023-05-30
socket enricher: add support for IPv6

alban:
IPv6 support is useful even if users care only about IPv4 because a process can listen on a socket using both IPv4 and IPv6, and internally it will be IPv6 packets even if the client is connecting using IPv4.

The problem was found while working on https://github.com/inspektor-gadget/inspektor-gadget/pull/1630


mauriciovasquezbernal:
Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1340

alban:
> Fixes #1340

It does not fix the snapshot socket gadget. The snapshot socket gadget has this kind of filter that I didn't remove:

```
        /* Filter out IPv6 for now */
        if (sk_common->skc_family != AF_INET)
                return 0;
```

mauriciovasquezbernal:
Nevermind, I read the title too fast.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
```diff
@@ -81,26 +81,38 @@ struct {
...
-	if (load_half(skb, offsetof(struct ethhdr, h_proto)) != ETH_P_IP)
-		return 0;
-
 	struct sockets_key key = {0,};
+	int l4_off;
+
 	key.netns = current_netns;
-	key.family = AF_INET;
-	key.proto = load_byte(skb, ETH_HLEN + offsetof(struct iphdr, protocol));
+	key.family = load_half(skb, offsetof(struct ethhdr, h_proto));
```
mauriciovasquezbernal, line 88: nit: use another variable? It's a bit confusing to have the same variable to store the ether proto and the family.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
```diff
@@ -81,26 +81,38 @@ struct {
...
+	switch (key.family) {
+	case ETH_P_IP:
+		key.family = AF_INET;
+		key.proto = load_byte(skb, ETH_HLEN + offsetof(struct iphdr, protocol));
+		l4_off = ETH_HLEN + sizeof(struct iphdr);
```
mauriciovasquezbernal, line 93: It's missing the options field. I think we can use the [IHL (Internet Header Length)](https://en.wikipedia.org/wiki/Internet_Protocol_version_4#IHL), however we need to check that the verifier is happy.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
```diff
@@ -81,26 +81,38 @@ struct {
...
+		break;
+	case ETH_P_IPV6:
+		key.family = AF_INET6;
+		key.proto = load_byte(skb, ETH_HLEN + offsetof(struct ipv6hdr, nexthdr));
+		l4_off = ETH_HLEN + sizeof(struct ipv6hdr);
```
mauriciovasquezbernal, line 98: Same here. However in this case it's not too bad because the current implementation just ignores those packets as nexthdr is not `IPPROTO_TCP` nor `IPPROTO_UDP`.

mauriciovasquezbernal (approved):
LGTM. Some comments about variable IP headers length. Maybe you can add a comment in the commit message indicating ipv6 sockets were already collected by ebpf probes but this change is only about supporting in on `gadget_socket_lookup`?

pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
alban: It is still 2 variables for 2 different things:
1. key.family: layer 3 protocol (aka family): IPv4 or IPv6
2. key.proto: layer 4 protocol: tcp or udp

My PR didn't change that. The patch might be confusing because I moved the assignment of `key.proto` below.

I could rename the fields to `l3_proto` and `l4_proto` to avoid confusion? (in another PR)

pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
alban: We have the sni gadget that parses the IHL:
https://github.com/inspektor-gadget/inspektor-gadget/blob/427603a57dff2d265f10d9a864fc093427dd12de/pkg/gadgets/trace/sni/tracer/bpf/snisnoop.c#L142-L146

I can try to do something similar.

pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
mauriciovasquezbernal: My point is, in L88 it's reading **`h_proto`** (i.e. [EtherType](https://en.wikipedia.org/wiki/EtherType)), but then it's being converted to family, AF_INET in L91 and AF_INET6 in L96. I find it a bit confusing that h_proto is saved in a variable named "family", when those are two different things. 

That change I propose is:

```diff
$ git diff
diff --git pkg/gadgets/internal/socketenricher/bpf/sockets-map.h pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
index 935d3502..f68ef38a 100644
--- pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
+++ pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
@@ -83,10 +83,11 @@ gadget_socket_lookup(const struct __sk_buff *skb)
 {
        struct sockets_key key = {0,};
        int l4_off;
+       __u16 h_proto
 
        key.netns = current_netns;
-       key.family = load_half(skb, offsetof(struct ethhdr, h_proto));
-       switch (key.family) {
+       h_proto = load_half(skb, offsetof(struct ethhdr, h_proto));
+       switch (h_proto) {
        case ETH_P_IP:
                key.family = AF_INET;
                key.proto = load_byte(skb, ETH_HLEN + offsetof(struct iphdr, protocol));
```

alban:
Updated

Issue #1310, opened 2023-02-06
[RFE] Filter containers by container image + Enrich events by container image


alban:
## Current situation

Users can filter containers with the following selectors:
- `--node`
- `--namespace`
- `--podname`
- `--containername`

But they cannot filter by container image.

Additionally, events are enriched with node, namespace, pod, container but not with the container image.

## Impact

Sometimes a micro service application uses several Kubernetes deployments with the same image and it is not possible to get events from all of them.

Example: https://github.com/microservices-demo/microservices-demo/tree/master/deploy/kubernetes/manifests
This has several deployments with the `mongo` image

If we notice that `mongo` has an issue but we don't know yet which  namespace, pod or deployment it is, it would be useful to have a `--image` filter.

Additionally, reporting container images in events has been requested by users in #1164.

```
$ kubectl gadget trace exec -A -o custom-columns=node,namespace,pod,container,image,comm,pid
Error: creating parser: invalid columns: image
```

## Ideal future situation

Implement `--image=` to most gadgets where it makes sense.

There should be a way to match both `busybox`, `library/busybox`, `busybox:1.35`. It could be done with Golang's [regexp](https://pkg.go.dev/regexp) or [glob](https://pkg.go.dev/github.com/gobwas/glob).

Additionally, we can add the image in events:
```
$ ./kubectl-gadget trace exec -A -o custom-columns=container,image
CONTAINER  IMAGE
busybox    busybox:latest
```

## Implementation options

The different backends for the container-collection packages need to feed the container image:

* WithContainerRuntimeEnrichment
  * CRI-O starts OCI containers with the following annotation:
    * "io.kubernetes.cri-o.ImageName": "k8s.gcr.io/coredns/coredns:v1.8.6"
    * See pkg/container-utils/oci-annotations/resolver_crio.go
  * containerd starts OCI containers with the following annotation:
    * "io.kubernetes.cri.image-name": "k8s.gcr.io/coredns/coredns:v1.8.6"
    * See pkg/container-utils/oci-annotations/resolver_containerd.go
* WithPodInformer
  * The pod spec contains a `image` field.

`ContainerSelectorMatches()` needs to be updated to support images.

## Additional information

Sysdig can filter events based on the container image with the `%container.image` filter.

alban:
Additional requirements for this feature request:
1. There should be fields for both the image tag and the image digest
2. We should be able to filter by prefix, for example filter all container image with the prefix "localhost:5000/".

[...]

Issue #737, opened 2022-07-06
[RFE] Enrich containers with k8s metadata the runtime information


blanquicet:
### Current situation

The [containerRuntimeEnricher](https://github.com/kinvolk/inspektor-gadget/blob/65a398e8ef72887c9b36550de64e7b00cfa8451f/pkg/container-collection/options.go#L43) is enriching the containers with the container name and the PID (only at start-up when retrieving the already running containers). However, there is a lot of information that we can gather from the runtimes to improve the information display per container and the filtering.

### Ideal future situation

We should start by adding the following fields:

- **Container runtime name**: Given that we are collecting the container from different runtimes, we should show the owner of each container. 
- **Container ID**: Two containers can have the same name so the only way to distinguish them is to use the ID.
- **Kubernetes metadata**: For the use-case of `ig` (formerly known as `local-gadget`), it would be very useful to also get the Kubernetes information: Namespace and Pod.

#### Current view
```bash
$ local-gadget list-containers
myContainer1 
shell01     
shell01     
```

#### Future view
```bash
$ local-gadget list-containers --k8s-metadata
CONTAINER RUNTIME    ID               NAME            K8S-POD    K8S-NAMESPACE
cri-o                6af6dd37ab878    myContainer1    myPod      myNamepsace      # Container created with k8s  
containerd           4c45eb2fb0211    shell01                                     # Rouge container
docker               0dbab37ba6429    shell01                                     # Rouge container
```

### Additional improvements

Once https://github.com/kinvolk/inspektor-gadget/issues/734 is merged, we could also add the container runtime namespace. It is supported by Containerd and a user may want to see only the containers created via Kubernetes and not the other ones. To do that it could use the runtime namespace `k8s.io` that is used by Containerd.

blanquicet:
https://github.com/inspektor-gadget/inspektor-gadget/pull/1702 added the k8s information to the events, but it can be displayed only in json/yaml format, not in columns yet. So this issue can't be closed yet.

mqasimsarfraz:
[...]

PR #1682, opened 2023-05-30
Enrich new containers with ContainerImageName

claudiamarcubina:
# Enrich events with container image (ImageName)
- Implements #1310 

This PR partially implements #1310. The current iteration addresses only **new containers enrichment** for **containerd and cri-o** and adds `ContainerImageName` attribute to events as per the following table:

|                               | Initial containers            | New containers              |
|-------------------------------|-------------------------------|-----------------------------|
| With k8s: containerd or cri-o | TODO  | OCI annotations             |
| With k8s: docker              | TODO   | *Blocker: No OCI annotations* |
| W/o k8s: containerd           | *Blocker: no k8s, but may be solved by  #737*   | OCI annotations             |
| W/o k8s: podman               | *Blocker: no k8s, but may be solved by  #737*                             | OCI annotations             |
| W/o k8s: docker                 | *Blocker: no k8s, but may be solved by  #737* | *Blocker: No OCI annotations* |

## How to use

## Testing done

`imageName` field in gadgets output is populated for new containers :
`$ kubect-gadget trace exec -A -o json`

image

`IMAGENAME` column in `ig list-containers` output is populated for new containers:

image

### DIFF
--- a/docs/crds/gadgets/ebpftop.md
+++ b/docs/crds/gadgets/ebpftop.md
@@ -8,7 +8,7 @@ ebpftop shows cpu time used by ebpf programs.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,node,namespace,pod,container,hostnetwork,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount,totalcpu,percpu). (default -runtime,-runcount)
+ - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,runtime.containerImageName,node,namespace,pod,container,hostnetwork,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount,totalcpu,percpu). (default -runtime,-runcount)
 
 ### Example CR
 
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -193,6 +193,15 @@ func WithRuntimeMetadata(runtime string) CommonDataOption {
 	}
 }
 
+// WithContainerImageName sets the ContainerImageName to facilitate the tests
+func WithContainerImageName(imageName string, isDockerRuntime bool) CommonDataOption {
+	return func(commonData *eventtypes.CommonData) {
+		if !isDockerRuntime {
+			commonData.Runtime.ContainerImageName = imageName
+		}
+	}
+}
+
 func BuildCommonData(namespace string, options ...CommonDataOption) eventtypes.CommonData {
 	e := eventtypes.CommonData{
 		K8s: eventtypes.K8sMetadata{
@@ -268,4 +277,21 @@ func GetPodUID(ns, podname string) (string, error) {
 func CheckNamespace(ns string) bool {
 	cmd := exec.Command("kubectl", "get", "ns", ns)
 	return cmd.Run() == nil
+}
+
+// IsDockerRuntime checks whether the container runtime of the first node in the Kubernetes cluster is Docker or not.
+func IsDockerRuntime() (error, bool) {
+	cmd := exec.Command("kubectl", "get", "node", "-o", "jsonpath={.items[0].status.nodeInfo.containerRuntimeVersion}")
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	r, err := cmd.Output()
+	if err != nil {
+		return fmt.Errorf("%w: %s", err, stderr.String()), false
+	}
+	ret := string(r)
+
+	if strings.Contains(ret, "docker") {
+		return nil, true
+	}
+	return nil, false
 }
--- a/integration/ig/k8s/trace_bind_test.go
+++ b/integration/ig/k8s/trace_bind_test.go
@@ -32,8 +32,12 @@ func TestTraceBind(t *testing.T) {
 		Cmd:          fmt.Sprintf("ig trace bind -o json --runtimes=%s", *containerRuntime),
 		StartAndStop: true,
 		ExpectedOutputFn: func(output string) error {
+			isDockerRuntime := *containerRuntime == ContainerRuntimeDocker
 			expectedEntry := &bindTypes.Event{
-				Event:    BuildBaseEvent(ns, WithRuntimeMetadata(*containerRuntime)),
+				Event: BuildBaseEvent(ns,
+					WithRuntimeMetadata(*containerRuntime),
+					WithContainerImageName("docker.io/library/busybox:latest", isDockerRuntime),
+				),
 				Comm:     "nc",
 				Protocol: "TCP",
 				Addr:     "::",
--- a/integration/inspektor-gadget/advise_networkpolicy_test.go
+++ b/integration/inspektor-gadget/advise_networkpolicy_test.go
@@ -15,10 +15,7 @@
 package main
 
 import (
-	"bytes"
 	"fmt"
-	"os/exec"
-	"strings"
 	"testing"
 
 	networkTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/network/types"
@@ -32,6 +29,12 @@ func TestAdviseNetworkpolicy(t *testing.T) {
 	nsClient := GenerateTestNamespaceName("test-advise-networkpolicy-client")
 
 	t.Parallel()
+
+	// TODO: Handle it once we support getting container image name from docker
+	errIsDocker, isDockerRuntime := IsDockerRuntime()
+	if errIsDocker != nil {
+		t.Fatalf("checking if docker is current runtime: %v", errIsDocker)
+	}
 
 	commands := []*Command{
 		CreateTestNamespaceCommand(nsServer),
@@ -58,7 +61,7 @@ func TestAdviseNetworkpolicy(t *testing.T) {
 			Cmd:  fmt.Sprintf(`$KUBECTL_GADGET advise network-policy monitor -n %s --timeout 5 --output - | tee ./networktrace-client.log`, nsClient),
 			ExpectedOutputFn: func(output string) error {
 				expectedEntry := &networkTypes.Event{
-					Event:     BuildBaseEvent(nsClient),
+					Event:     BuildBaseEvent(nsClient, WithContainerImageName("docker.io/library/busybox:latest", isDockerRuntime)),
 					Comm:      "nc",
 					Uid:       0,
 					Gid:       0,
@@ -89,7 +92,9 @@ func TestAdviseNetworkpolicy(t *testing.T) {
 					e.K8s.Node = ""
 					e.K8s.ContainerName = ""
 					// TODO: Verify container runtime and container name
-					e.Runtime = eventtypes.BasicRuntimeMetadata{}
+					e.Runtime.RuntimeName = ""
+					e.Runtime.ContainerName = ""
+					e.Runtime.ContainerID = ""
 				}
 
 				return ExpectEntriesToMatch(output, normalize, expectedEntry)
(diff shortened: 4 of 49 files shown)


claudiamarcubina:
> The following patch should help you for the CI but I suspect you will need to tackle some corner cases where we do not use `busybox` as the image:
> 
> ```diff
> diff --git a/integration/helpers.go b/integration/helpers.go
> index 7fab694c..76165a67 100644
> --- a/integration/helpers.go
> +++ b/integration/helpers.go
> @@ -185,6 +185,7 @@ func BuildCommonData(namespace string) eventtypes.CommonData {
>                 // Pod and Container name are defined by BusyboxPodCommand.
>                 Pod:       "test-pod",
>                 Container: "test-pod",
> +               ImageName: "docker.io/library/busybox"
>                 // TODO: Include the Node
>         }
>  }
> ```

I also thought so, but testing it the second time now and it make even more tests fail.

blanquicet:
Hi @claudiamarcubina, thanks for working on this.

I was looking at the ways you found to get the Image info and I have a proposal for the k8s cases. On top of #1702, if you get the image name from the container runtimes, it would be available automatically enriched to the containers with the following: [withPodInformer](https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/options.go#L222C6-L222C21) -> [k8sClient.GetRunningContainers](https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/options.go#L282C20-L282C55) -> [RuntimeMetadata](https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/k8s.go#L148). 

I think it would be valid for new and initial containers. WDYT?

cc @alban

pkg/types/types.go
```diff
@@ -33,6 +33,7 @@ func init() {
 	columns.MustRegisterTemplate("namespace", "width:30")
 	columns.MustRegisterTemplate("pod", "width:30,ellipsis:middle")
 	columns.MustRegisterTemplate("container", "width:30")
+	columns.MustRegisterTemplate("imageName", "width:30")
```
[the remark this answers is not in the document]
claudiamarcubina: Do you refer just to the name of the column or also the member of the structs i.e `CommonData struct`?

I thought about `imageName` as it follows to add imageDigest too.

integration/inspektor-gadget/advise_networkpolicy_test.go
```diff
@@ -33,6 +30,9 @@ func TestAdviseNetworkpolicy(t *testing.T) {
 
 	t.Parallel()
 
+	// TODO: Handle it once we support getting container image name from docker
+	errIsDocker, isDocker := IsDockerRuntime()
```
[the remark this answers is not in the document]
claudiamarcubina: That's a good idea! However, in order to return the error here I think I would need to change the function arguments from   `TestAdviseNetworkpolicy(t *testing.T) { }`  to `TestAdviseNetworkpolicy(t *testing.T) error { }`. Am I missing somehting?

pkg/container-utils/oci-annotations/types.go
```diff
@@ -26,6 +26,8 @@ type Resolver interface {
 	ContainerName(annotations map[string]string) string
 	// ContainerType returns the type of the container i.e "container" or "sandbox"
 	ContainerType(annotations map[string]string) string
+	// ContainerImage returns the image name of the container
```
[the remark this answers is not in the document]
claudiamarcubina: Good hint. Added, thanks :)

pkg/container-collection/containers.go
```diff
@@ -65,7 +65,7 @@ type Container struct {
 	Name      string            `json:"name,omitempty" column:"name,width:30" columnTags:"runtime"`
```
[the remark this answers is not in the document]
claudiamarcubina: Right. Updated it.

claudiamarcubina:
> Hi @claudiamarcubina, thanks for working on this.
> 
> I was looking at the ways you found to get the Image info and I have a proposal for the k8s cases. On top of #1702, if you get the image name from the container runtimes, it would be available automatically enriched to the containers with the following: [withPodInformer](https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/options.go#L222C6-L222C21) -> [k8sClient.GetRunningContainers](https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/options.go#L282C20-L282C55) -> [RuntimeMetadata](https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/k8s.go#L148).
> 
> I think it would be valid for new and initial containers. WDYT?
> 
> cc @alban

HI, Jose :)

Thanks again for the idea. As talking to you and Alban, podInformer is asynchronous and provides info about new containers with some delay. So it can happen that the first events generated by a container have no container info associated. 

When a new container is detected by container-hook, it is added synchronously. Then, when podinformer notices it exists after a while, it would go to this code path: https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/options.go#L297 and do nothing because the container was already added.

blanquicet:
> When a new container is detected by container-hook, it is added synchronously. Then, when podinformer notices it exists after a while, it would go to this code path:
> 
> https://github.com/inspektor-gadget/inspektor-gadget/blob/a39e7b00cc2d114259f3d9121aa4232ac7f6e407/pkg/container-collection/options.go#L297
> and do nothing because the container was already added.
>
Oh, I had completely missed that drop. Thanks for the detailed answer! 

At this point, I'm wondering why not use `WithContainerRuntimeEnrichment` in ig-k8s to enrich the new containers?

claudiamarcubina:
> WithContainerRuntimeEnrichment

Hm, I am planning to test using `WithContainerRuntimeEnrichment` for adding the image digest. Let's see how that goes in a next PR.

pkg/container-collection/container-collection.go
```diff
@@ -442,6 +442,7 @@ func (cc *ContainerCollection) EnrichByMntNs(event *eventtypes.CommonData, mount
 		event.Runtime.RuntimeName = container.Runtime.RuntimeName
 		event.Runtime.ContainerName = container.Runtime.ContainerName
 		event.Runtime.ContainerID = container.Runtime.ContainerID
+		event.Runtime.ImageName = container.Runtime.ImageName
```
blanquicet, line 445: You're missing to do the same in `EnrichEventByMntNs` and `EnrichEventByNetNs`.

pkg/types/types.go
```diff
@@ -116,6 +117,10 @@ type BasicRuntimeMetadata struct {
 	// ContainerName is the container name. In the case the container runtime
 	// response with multiple containers, ContainerName contains only the first element.
 	ContainerName string `json:"containerName,omitempty" column:"containerName,template:container"`
+
+	// ImageName is the name of the container image where the event comes from
+	// i.e. docker.io/library/busybox:latest
+	ImageName string `json:"imageName,omitempty" column:"imageName" columnTags:"runtime"`
```
blanquicet, line 123: We can `ContainerName` and `ContainerID`, so I think this should be `ContainerImage`.

blanquicet, line 123: `runtime` container-tag is not necessary anymore as it inherits it from the definition of `BasicRuntimeMetadata` in `CommonData`.

pkg/container-collection/container-collection.go
claudiamarcubina: I thought this is automatically done on [this line](https://github.com/claudiamarcubina/inspektor-gadget/blob/14233a54cf36cf253a4b2f5d6e9641297544d898/pkg/container-collection/operator.go#L30C3-L30C3) and on [this one](https://github.com/claudiamarcubina/inspektor-gadget/blob/14233a54cf36cf253a4b2f5d6e9641297544d898/pkg/container-collection/operator.go#L46), isn't it?

blanquicet: You're right.

integration/ig/k8s/trace_bind_test.go
```diff
@@ -40,6 +40,11 @@ func TestTraceBind(t *testing.T) {
 				Options:  ".R...",
 			}
 
+			//TODO: Handle it once we support getting container image name for docker
+			if *containerRuntime == ContainerRuntimeDocker {
+				expectedEntry.CommonData.ImageName = ""
+			}
+
```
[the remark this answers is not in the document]
claudiamarcubina: Done. Thanks :)

claudiamarcubina:
After rebasing, initial containers enrichment was broken. For a better understanding I need to run more tests for that, so decided to merge this first part. Initial containers enrichment will come in a future PR.

integration/helpers.go
```diff
@@ -193,6 +193,19 @@ func WithRuntimeMetadata(runtime string) CommonDataOption {
...
+// WithCommonDataContainerImage sets the ContainerImageName to facilitate the tests
+func WithCommonDataContainerImage(containerImage string, isDockerRuntime bool) CommonDataOption {
+	return func(commonData *eventtypes.CommonData) {
+		commonData.Runtime.ContainerImageName = containerImage
+		// TODO: Handle it once we support getting container image name for docker.
+		// As docker do not provide OCI annotations, we need to find another way
+		// to retrieve image name for new containers.
+		if isDockerRuntime {
+			commonData.Runtime.ContainerImageName = ""
+		}
```
blanquicet, line 205:
```suggestion
		// TODO: Handle it once we support getting container image name for docker.
		// As docker do not provide OCI annotations, we need to find another way
		// to retrieve image name for new containers.
		if !isDockerRuntime {
			commonData.Runtime.ContainerImageName = containerImage
		}
```
?

integration/ig/k8s/trace_bind_test.go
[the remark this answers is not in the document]
blanquicet: > In this way, once we support getting the image name for docker, we just have the above branch and not one in each test file.
>
Mmmm @eiffel-fl, isn't it true only if the computation of `isDockerRuntime` was done inside `WithCommonDataContainerImage` (which seems possible)? With your suggestion, once we start supporting Docker, we will need to clean up this function and the computation and passage of `isDockerRuntime` from each test to this function, don't you think?

@claudiamarcubina In general, it is just a matter of making it simple to be removed in the near future. However, you already did it like this and the tests are passing, I don't think it is worth changing it.

integration/helpers.go
```diff
@@ -193,6 +193,19 @@ func WithRuntimeMetadata(runtime string) CommonDataOption {
 	}
 }
 
+// WithCommonDataContainerImage sets the ContainerImageName to facilitate the tests
+func WithCommonDataContainerImage(containerImage string, isDockerRuntime bool) CommonDataOption {
```
blanquicet, line 197: nit: 
```suggestion
func WithContainerImageName(imageName string, isDockerRuntime bool) CommonDataOption {
```

blanquicet (commented):
Awesome work @claudiamarcubina! It looks good from a code inspection. I will run a couple of tests on Monday and give my approval.

integration/helpers.go
claudiamarcubina: Good idea! Thank you. Added in a separate commit.

integration/helpers.go
claudiamarcubina: Added in the same commit as previous.

docs/crds/gadgets/ebpftop.md
```diff
@@ -8,7 +8,7 @@ ebpftop shows cpu time used by ebpf programs.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,node,namespace,pod,container,hostnetwork,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount,totalcpu,percpu). (default -runtime,-runcount)
+ - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,runtime.containerImageName,node,namespace,pod,container,hostnetwork,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount,totalcpu,percpu). (default -runtime,-runcount)
```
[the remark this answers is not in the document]
claudiamarcubina: Do you mean to have tests update and field rename in the same commit? I thought it would be a bit more clear to have it separated. WDYT?

blanquicet (approved):
LGTM! Awesome!

PR #1698, opened 2023-06-01
columns: Support prefix when using templates

blanquicet:
# columns: Support prefix when using templates

Currently, in the following case, the prefix `runtime` is not applied to the `Container` field because it uses a template. While the prefix is correctly applied to the `ID` field:

```golang
type RuntimeMetadata struct {
	ID        string `column:"id"`
	Container string `column:"container,template:container"`
}

type GadgetData struct {
	RuntimeMetadata RuntimeMetadata `column:"runtime" columnTags:"runtime"`
	GadgetData      string          `column:"gadgetData"`
}
```

This PR makes the columns library to support applying prefixes also when templates are used.


flyth (approved):
Looks good! Thanks for also adding another test!

Issue #1490, opened 2023-04-03
[RFE] Trace exec: additional fields: cwd, exepath, tty, sid, gid, loginuid


alban:
## Current situation

The trace exec gadget already has the following fields:
```
$ ig trace exec -h
...
Available columns:
  args
  comm
  container
  mntns
  pid
  ppid
  ret
  timestamp
  uid
```

But it does not have the following columns:
- container (separate issue for this: [#1310](https://github.com/inspektor-gadget/inspektor-gadget/issues/1310))
  - container id (IG has this field internally but it is not exposed to the user yet)
  - image tag
  - image digest
- process info
  - [x] cwd (#1700)
  - [x] exepath
  - [x] tty (not needed)
  - [x] sid (#1705)
- user info
  - [x] gid (#1666)
  - [x] loginuid (#1704)

sysdig provides those fields but not Inspektor Gadget.

## Impact

The context around the exec events are not detailed enough for some of our users.

## Ideal future situation

- All of the additional fields are added.
- Some of the fields can be generic and should be available in all trace gadgets.

## Implementation options

Some of the fields can be straightforward to get in eBPF.

sysdig gets the `cwd` by keeping track of the directory for all processes in a eBPF map and using a kprobe on chdir to update that map.

## Additional information

[ Please Add any information that does not fit into any of the above sections here ]

mauriciovasquezbernal:
I tried to check how easy it's to get the different fields:

- [sid](https://github.com/inspektor-gadget/inspektor-gadget/blob/d192da3b4ce9de65584e82726d9299406e790828/pkg/arm64/vmlinux/vmlinux.h#L1035-L1036)
- gid through the `bpf_get_current_uid_gid` heper
- [loginuid](https://github.com/inspektor-gadget/inspektor-gadget/blob/d192da3b4ce9de65584e82726d9299406e790828/pkg/arm64/vmlinux/vmlinux.h#L1034-L1035) 
- `exepath` is `pathname` in execve, it's currently read in args[0]. 
[...]

PR #1700, opened 2023-06-01
trace/exec: add cwd field

mauriciovasquezbernal:
See https://github.com/inspektor-gadget/inspektor-gadget/issues/1490

## TBD
- [ ] Investigate how much does it increase the complexity
The feature is now disabled by default. Complexity shouldn't be a problem as the kernel removes the dead code. 
- [ ] How to avoid creating multiple instances of the percpu buffer?
TBD later as it's not that simple.

## Questions
- Should this be enabled by default?
R: Not, it's only enabled when the `--cwd` flag is passed.
- Should we implement this in other gadgets?
R: Not for now.


pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
```diff
@@ -84,6 +89,12 @@ int ig_execve_e(struct trace_event_raw_sys_enter* ctx)
 	event->args_size = 0;
 	event->mntns_id = mntns_id;
 
+	if (get_cwd) {
+		fs = BPF_CORE_READ(task, fs);
+		cwd = get_path_str(__builtin_preserve_access_index(&fs->pwd));
```
alban, line 94: Why do we need __builtin_preserve_access_index?

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
```diff
@@ -84,6 +89,12 @@ int ig_execve_e(struct trace_event_raw_sys_enter* ctx)
...
+		bpf_probe_read(event->cwd, MAX_STRING_SIZE, cwd);
```
alban, line 95: bpf_probe_read_kernel_str()
- Use the _kernel suffix since it is kernel memory
- Use _str for strings to avoid copying too much

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.h
```diff
@@ -24,6 +24,7 @@ struct event {
 	int retval;
 	int args_count;
 	unsigned int args_size;
+	__u8 cwd[MAX_STRING_SIZE];
 	__u8 comm[TASK_COMM_LEN];
 	__u8 args[FULL_MAX_ARGS_ARR];
```
alban, line 29: Before the patch, the event was 7752 bytes. This adds 4kB to the event. And events sent to a perf ring buffer are rounded up to the next power of 2 size, so that probably makes it 16kB.

```
PerfBufferPages = 64
gadgets.PerfBufferPages*os.Getpagesize() = 256 kB
```
This means the perf per-cpu ring buffer can hold about 16 events before being full.

Should we send a smaller event if `get_cwd` is `false`? Maybe it is ok for this PR, but something to think about if we add more strings.

alban (commented):
Does it work on older kernels?

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
mauriciovasquezbernal: I just wanted to make it explicit a relocation happens there, but it's not needed since we're using vmlinux.h. I removed it.

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.h
mauriciovasquezbernal: > This adds 4kB to the event.

you're right :( 

> Should we send a smaller event if get_cwd is false? 

I think we should optimize it. I tried to follow the same idea as with args: to use a memory zone to send null terminated strings as str0\0str1\0str2, etc. however I wasn't able to make it work. 

I also tried to send a smaller event when get_cwd is false, but I didn't find an easy way without having to change a lot of the code. I'll try again tomorrow.

mauriciovasquezbernal: I handled it by using conditional compilation. I tried to use a constant but it's difficult because the last field of the events structure (args) has already variable length, so modifying the code not to send the cwd is not easy. I think we can use this for now in this way.

mauriciovasquezbernal:
> Does it work on older kernels?

I tried in 5.4 and it worked fine. I tried in 4.15 but nothing works there, no even without this PR.

alban (approved):
LGTM after the issue about the `args` field is handled.

Note: The documentation in docs/gadgets/trace/exec.md should be updated, but it can be done in a following PR.

I tested the PR with the following commands:

```
$ docker run --privileged -v /:/host -ti --rm --name shell01 busybox
/ # cd /host/proc/1/cwd/var/run && /bin/ls
```

```
$ sudo ./ig trace exec --cwd -c shell01
CONTAINER    PID        PPID       COMM     RET ARGS     CWD
shell01      3506503    3505764    ls       0   /bin/ls  /run
```

* It follows symlinks
* The paths for foreign mounts are generated compared to the root in the foreign mount namespace

pkg/gadgets/trace/exec/tracer/tracer.go
```diff
@@ -134,6 +143,7 @@ func (t *Tracer) run() {
 			continue
 		}
 
+		// this works regardless the kind of event because cwd is defined at the end of the structure
 		bpfEvent := (*execsnoopEvent)(unsafe.Pointer(&record.RawSample[0]))
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Args has to be the last field because the ebpf program has some optimization not to send the full event but only the bytes that are actually needed. I added a comment clarifying that cwd is at the end but before args.

alban (approved):

PR #1704, opened 2023-06-01
trace/exec: Add loginuid field

mauriciovasquezbernal:
Part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1490. 

Should we add this on all gadgets too? 
/cc @alban


pkg/gadgets/trace/exec/types/types.go
```diff
@@ -25,13 +25,14 @@ type Event struct {
...
-	Uid    uint32   `json:"uid" column:"uid,template:uid,hide"`
-	Gid    uint32   `json:"gid" column:"gid,template:gid,hide"`
+	Pid      uint32   `json:"pid,omitempty" column:"pid,template:pid"`
+	Ppid     uint32   `json:"ppid,omitempty" column:"ppid,template:pid"`
+	Comm     string   `json:"comm,omitempty" column:"comm,template:comm"`
+	Retval   int      `json:"ret,omitempty" column:"ret,width:3,fixed"`
+	Args     []string `json:"args,omitempty" column:"args,width:40"`
+	Uid      uint32   `json:"uid" column:"uid,template:uid,hide"`
+	Gid      uint32   `json:"gid" column:"gid,template:gid,hide"`
+	LoginUid uint32   `json:"loginuid,omitempty" column:"loginuid,template:uid,hide"`
```
alban, line 35:
```suggestion
	LoginUid uint32   `json:"loginuid" column:"loginuid,template:uid,hide"`
```

alban (approved):
LGTM after the comment about omitempty is addressed.

> Should we add this on all gadgets too?

Ideally yes, but this is low priority, so you can also wait to see if users start requesting it.

PR #1705, opened 2023-06-02
trace/exec: Add sessionId

mauriciovasquezbernal:
See https://github.com/inspektor-gadget/inspektor-gadget/issues/1490


pkg/gadgets/trace/exec/types/types.go
```diff
@@ -25,14 +25,15 @@ type Event struct {
...
-	LoginUid uint32   `json:"loginuid,omitempty" column:"loginuid,template:uid,hide"`
+	Pid       uint32   `json:"pid,omitempty" column:"pid,template:pid"`
+	Ppid      uint32   `json:"ppid,omitempty" column:"ppid,template:pid"`
+	Comm      string   `json:"comm,omitempty" column:"comm,template:comm"`
+	Retval    int      `json:"ret,omitempty" column:"ret,width:3,fixed"`
+	Args      []string `json:"args,omitempty" column:"args,width:40"`
+	Uid       uint32   `json:"uid" column:"uid,template:uid,hide"`
+	Gid       uint32   `json:"gid" column:"gid,template:gid,hide"`
+	LoginUid  uint32   `json:"loginuid,omitempty" column:"loginuid,template:uid,hide"`
+	SessionId uint32   `json:"sessionid,omitempty" column:"sessionid,minWidth:10,hide"`
```
alban, line 36: Why omitempty? I think it is similar to uid/gid where we should not use omitempty.

alban (approved):
1 comment about omitempty. LGTM otherwise.

pkg/gadgets/trace/exec/types/types.go
mauriciovasquezbernal: Right, I just forgot about it. Thanks for catching it up!

Issue #414, opened 2021-11-24
[RFE] Provide a HELM chart


margamanterola:
## Current situation

Deploying Inspektor Gadget on the cluster requires running the deploy subcommand and applying the output of that command. There's no HELM chart

## Impact

Operators that are used to deploying with HELM need to work outside of their workflow both to deploy and remove the deployed pods.

## Ideal future situation

We should provide a HELM chart and make sure it's published in the right places so that people that use HELM charts can find it.

## Implementation options

To avoid duplication of work, it would make sense if both the HELM chart and the output of the deploy subcommand have the same source of truth.

mqasimsarfraz:
Helm Charts for Inspektor Gadget have been published [here](https://inspektor-gadget.github.io/charts), closing the issue!

PR #1707, opened 2023-06-02
charts: Add helm charts for Inspektor Gadget

mqasimsarfraz:
This PR introduces helm charts for Inspektor Gadget. The idea is to have a single source of truth for `deploy` command as helm charts to avoid running into issues in future. The current approach will reuse the CRDs in both installation workflows. In case of manifest, helm will be source of truth and helm generated manifest is used as base (`deploy.yaml`) for `deploy` command. In terms of versioning, I want to start with 1:1 versioning between chart version and app version since it is simple and [recommend](https://codefresh.io/docs/docs/ci-cd-guides/helm-best-practices/) in the start and app version will used for container image tag unless specified.

For publishing we will setup a index like [here](https://github.com/mqasimsarfraz/charts/blob/main/index.yaml) using `helm repo index ...` but pointing to charts in our release assets. I want this change to go in and then automate that step. 

Related: #414 

## Testing done

```bash
$ make minikube-start
$ make -C charts APP_VERSION=latest install
NAME: gadget
LAST DEPLOYED: Fri Jun  2 17:31:35 2023
NAMESPACE: gadget
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Inspektor Gadget latest installed successfully!

For usage details, visit https://www.inspektor-gadget.io
$ kubectl get pods -n gadget
NAME           READY   STATUS    RESTARTS   AGE
gadget-nsrx2   1/1     Running   0          2m30s
```
You can verify the charts using:

```bash
$ make -C charts [build/package]
$ ls -la charts/bin/gadget
...
-rw-rw-r-- 1 qasim qasim  529 Jun  2 17:04 Chart.yaml
drwxrwxr-x 2 qasim qasim 4096 Jun  2 17:04 crds
-rw-rw-r-- 1 qasim qasim  358 Jun  2 17:04 .helmignore
-rw-r--r-- 1 root  root  2887 Jun  2 17:04 README.md
-rw-rw-r-- 1 qasim qasim 1549 Jun  2 17:04 README.md.gotmpl
drwxrwxr-x 2 qasim qasim 4096 Jun  2 17:04 templates
-rw-rw-r-- 1 qasim qasim 1173 Jun  2 17:04 values.yaml
```

### Ensuring no regression in deploy.yaml

```bash
# deploy IG on minikube via 'main'
$ git checkout main
$ make kubectl-gadget
$ ./kubectl-gadget deploy
$ git checkout qasim/helm-charts
$ cat pkg/resources/manifests/*.yaml | kubectl diff -f -
...
```

## Todo

- [x] Add a test in CI that generated manifest has no changes. 
- [x] Add a step in release job to publish helm package.
- [ ] Add an action to publish charts index to charts repository (possibly inspektor-gadget/charts). [Post Merge]
- [ ] Verify helm charts are available on https://artifacthub.io with generated documentation [Post Merge]


charts/Makefile
```diff
@@ -0,0 +1,37 @@
...
+build: clean
+	@echo "Building chart:"
+	mkdir -p $(BUILD_DIR)
+	cp -r gadget $(BUILD_DIR)
+	cp -r ../pkg/resources/crd/bases $(CHART_DIR)/crds
+	mv $(CHART_DIR)/Chart.yaml.tmpl $(CHART_DIR)/Chart.yaml
+	sed -i "s/%VERSION%/$(CHART_VERSION)/g" $(CHART_DIR)/Chart.yaml
+	sed -i "s/%APP_VERSION%/$(IMAGE_TAG)/g" $(CHART_DIR)/Chart.yaml
+	@echo "Preparing docs"
+	docker run --user $(shell id -u):$(shell id -g) -v $(shell pwd)/$(CHART_DIR):/helm-docs jnorwood/helm-docs:v1.11.0 -s file
```
invidian, line 20: `jnorwood/helm-docs:v1.11.0` could also be in a variable in case an override is needed.

charts/Makefile
```diff
@@ -0,0 +1,37 @@
...
+	@echo "Charts available at: $(CHART_DIR)"
+
+install: build
+	$(HELM) install gadget $(CHART_DIR) --namespace gadget --create-namespace
```
invidian, line 24: Maybe namespace could be a variable as well? Is it supported to deploy IG in other namespaces?

charts/gadget/templates/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+{{ if .Values.createNamespace }}
```
invidian, line 1: I think this manifest is not necessary. IIRC Helm will complain if target namespace does not exist and `--create-namespace` flag is not specified.

charts/Makefile
invidian, line 24: Maybe it would make sense to use `helm upgrade --install` + `kubectl apply` for CRDs (as `helm upgrade` does not update CRDs) to be able to iterate on chart when developing it? `install` is not idempotent.

invidian (commented):
I'm not very familiar with charts publishing process, so I did not review this part of the PR, but manifests etc looks good. Left some questions/suggestions.

mqasimsarfraz (commented):
@invidian Thank you so much for the review!

charts/Makefile
invidian: If that's the case, perhaps all resources should have this namespace hardcoded then? As installing in other namespace will result in broken scenario.

charts/gadget/templates/_helpers.tpl
```diff
@@ -0,0 +1,49 @@
...
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "gadget.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Namespace used by all resources.
+*/}}
+{{- define "gadget.namespace" -}}
```
invidian, line 36: Great idea for still using a template for the namespace in case this changes in the future!

Makefile
```diff
@@ -267,6 +267,14 @@ minikube-deploy: minikube-start gadget-default-container kubectl-gadget
 btfgen:
 	+make -f Makefile.btfgen
 
+.PHONY: generate-manifests
+generate-manifests:
+	echo "---" > pkg/resources/manifests/deploy.yaml
+	echo "# This file is generated by 'make generate-manifests'; DO NOT EDIT." >> pkg/resources/manifests/deploy.yaml
+	cat pkg/resources/manifests/namespace.yaml >> pkg/resources/manifests/deploy.yaml
+	make -C charts APP_VERSION=latest template
```
mauriciovasquezbernal, line 275: nit: should it use `IMAGE_TAG` instead of latest in case you're trying it on a different branch than main?

However it'll require to update the logic that checks if this is updated. (I did a quick try and break other things, so if this is difficult we can leave as it's)

charts/Makefile
```diff
@@ -0,0 +1,44 @@
...
+	cp -r ../pkg/resources/crd/bases $(CHART_DIR)/crds
+	mv $(CHART_DIR)/Chart.yaml.tmpl $(CHART_DIR)/Chart.yaml
+	sed -i "s/%VERSION%/$(CHART_VERSION)/g" $(CHART_DIR)/Chart.yaml
+	sed -i "s/%APP_VERSION%/$(APP_VERSION)/g" $(CHART_DIR)/Chart.yaml
+	@echo "Preparing docs"
+	docker run --user $(shell id -u):$(shell id -g) -v $(shell pwd)/$(CHART_DIR):/helm-docs $(HELM_DOCS_IMAGE) -s file
+	@echo "Charts available at: $(CHART_DIR)"
+
+# install uses 'helm upgrade --install' to make chart installation idempotent,
+# also helm upgrade does not update CRDs so we need to apply them separately
```
mauriciovasquezbernal, line 28: :( this is too bad as users will have to handle additional commands manually. 

According to https://github.com/helm/helm/issues/8668#issuecomment-683925646:

> If you're chart doesn't install custom resources just the CRDs... you can put the CRDs right into the templates directory.

That seems to be our case. Perhaps we can investigate more to see if we can improve this. Anyway, I'll agree this an improvement and not a blocker for this PR.

mauriciovasquezbernal (commented):
Some comments. I'm a bit worried about the handling of the `latest` version. 

I was on this branch, I did:

```bash
$ make -C charts/ install 
...
Inspektor Gadget 0.17.0 installed successfully 
```

That was surprising as I was expecting the current version to be installed. Is that the expected version to install the last release?

pkg/resources/manifests/deploy.yaml
```diff
@@ -1,91 +1,99 @@
...
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: gadget
   namespace: gadget
 ---
+# Source: gadget/templates/clusterrole.yaml
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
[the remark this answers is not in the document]
mqasimsarfraz: In general, It might be problem like if `namespace` is moved in the end we will run into errors. But here we only swap `ClusterRole` and `Role` which aren't related so should be fine. Helm decide to order them like that as mentioned in this [section](https://helm.sh/docs/intro/using_helm/#helm-install-installing-a-package).

charts/Makefile
```diff
@@ -0,0 +1,44 @@
+HELM ?= helm
+HELM_DOCS_IMAGE ?= jnorwood/helm-docs:v1.11.0
+BUILD_DIR ?= bin
+OUTPUT_DIR ?= $(BUILD_DIR)
+CHART_DIR := $(BUILD_DIR)/gadget
+
+IMAGE_TAG := $(shell git describe --tags --abbrev=0 --match "v*")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: How do other folks handle the chart version of the "latest" app? Or they aren't charts available for "latest" apps?

mauriciovasquezbernal (approved):
I think it's ready to go! Thanks for handling this!

charts/Makefile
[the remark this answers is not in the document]
mqasimsarfraz: Most of the charts are [living outside](https://github.com/bitnami/charts/tree/main/bitnami) so they perhaps don't care about `latest`. Having said that, one can have dev versions (`1.0.0-dev`) for both chart/app and use `--set image.tag=$IMAGE_TAG` (e.g [cilium](https://github.com/cilium/cilium/blob/0447095e6e6bb877a06efd1d8802f6330c2bf786/contrib/scripts/kind-shell-helpers.sh#L68)) installation and worry about versioning only for releases. What we did now should work fine but in case it gets confusing we can rework it in future?

charts/Makefile
```diff
@@ -0,0 +1,43 @@
+HELM ?= helm
+HELM_DOCS_IMAGE ?= jnorwood/helm-docs:v1.11.0
+BUILD_DIR ?= bin
+OUTPUT_DIR ?= $(BUILD_DIR)
+CHART_DIR := $(BUILD_DIR)/gadget
+
+# CHART_VERSION must be a valid SemVer string so strip the leading 'v'
+CHART_VERSION ?= $(shell git describe --tags --always | sed 's/^v//')
```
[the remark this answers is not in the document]
mqasimsarfraz: Done.

mqasimsarfraz:
Thanks for all the reviews! :)

PR #1708, opened 2023-06-05
Read bpf iterators from host pid ns

alban:
# Read bpf iterators from host pid ns

In #1658, we are trying to remove `hostPID=true` . One issue with switching to `hostPID=false` is that bpf iterators of type `iter/task` or `iter/task_file` are working in the context of the current pidns (see https://github.com/inspektor-gadget/inspektor-gadget/issues/1658#issuecomment-1567114802).

This PR changes that by opening the bpf iterators from the host pidns by creating a systemd transient unit on the host by talking to systemd's unix socket (`/run/systemd/private`). Since `/run` is a host volume, Inspektor Gadget has access to it.

See systemd documentation:
* https://pkg.go.dev/github.com/coreos/go-systemd/v22/dbus#Conn.StartTransientUnitContext
* https://www.freedesktop.org/wiki/Software/systemd/dbus/

The only programs executed in the systemd unit are `sh` and `cat`, so this PR does not require any new dependencies to be installed on the host.

## How to use

Commands to test:

```
$ make ig
$ docker run --privileged -ti --rm \
        -v /:/host \
        -v /sys/kernel/debug:/sys/kernel/debug \
        -v /run:/run \
        -v /sys/fs/bpf:/sys/fs/bpf \
        -v $PWD:/src \
        -e HOST_ROOT=/host \
        ubuntu /src/ig snapshot process -r docker
```
```
go run -exec 'sudo unshare -p -f' ./cmd/ig/... snapshot process -r docker
```

## Testing done

Gadgets using bpf iterators of type `iter/task` or `iter/task_file`:
- [x] process snapshot gadget
- [ ] top ebpf gadget
- [ ] socket enricher

Depends on:
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/1662


pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -0,0 +1,197 @@
...
+		return false, errors.New("not a syscall.Stat_t")
+	}
+
+	return selfStat.Ino == systemdStat.Ino, nil
+}
+
+// Read reads the iterator in the host pid namespace.
+// It will test if the current pid namespace is the host pid namespace.
+func Read(iter *link.Iter) ([]byte, error) {
+	isHostPidNs, err := IsHostPidNs()
```
mauriciovasquezbernal, line 67: Can it be calculated in `init()` to avoid checking it each time?

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -0,0 +1,197 @@
...
+func ReadOnHostPidNs(iter *link.Iter) ([]byte, error) {
+	selfPidOnHost, err := os.Readlink(host.HostProcFs + "/self")
+	if err != nil {
+		return nil, fmt.Errorf("readlink /proc/self: %w", err)
+	}
+	if selfPidOnHost == "" {
+		return nil, fmt.Errorf("empty /proc/self symlink")
+	}
+
+	tmpPinDir, err := os.MkdirTemp("/sys/fs/bpf", "ig-iter-")
```
mauriciovasquezbernal, line 106: Shouldn't it prepend `host.HostProcFs` un case /sys/fs/bpf is not mounted?

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -0,0 +1,197 @@
...
+		systemdDbus.PropDescription("Inspektor Gadget job on host pidns"),
+		// Type=oneshot ensures that StartTransientUnitContext will only return "done" when the job is done
+		systemdDbus.PropType("oneshot"),
+		systemdDbus.PropExecStart([]string{
+			"/bin/sh",
+			"-c",
+			fmt.Sprintf("cat %s > %s", pinPath, writerPath),
+		}, true),
+	}
+	properties = append(properties, systemdDbus.PropDescription("Inspektor Gadget job on host pidns"))
```
mauriciovasquezbernal, line 159: What's the difference with L150?

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -0,0 +1,197 @@
...
+	select {
+	case s := <-statusChan:
+		close(statusChan)
+
+		// Close writer first: this will unblock the go routine reading from the pipe
+		w.Close()
+		wg.Wait()
+
+		if errReader != nil {
+			return nil, fmt.Errorf("reading from pipe: %w", err)
```
mauriciovasquezbernal, line 178: err -> errReader ?

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -0,0 +1,197 @@
...
+	conn, err := systemdDbus.NewSystemdConnectionContext(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("connecting to systemd: %w", err)
+	}
+	defer conn.Close()
+
+	randomID := make([]byte, 8)
+	for i := 0; i < 8; i++ {
+		randomID[i] = byte(65 + rand.Intn(25))
+	}
```
mauriciovasquezbernal, line 145: Use `github.com/google/uuid` to avoid handling this "by hand"?

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -0,0 +1,197 @@
...
+		}, true),
+	}
+	properties = append(properties, systemdDbus.PropDescription("Inspektor Gadget job on host pidns"))
+
+	_, err = conn.StartTransientUnitContext(context.TODO(),
+		unitName, "fail", properties, statusChan)
+	if err != nil {
+		return nil, fmt.Errorf("starting transient unit: %w", err)
+	}
+	timeout := time.NewTimer(5 * time.Second)
```
mauriciovasquezbernal, line 166: Is this big enough to cover reading a lot of entries in a busy system?

pkg/gadgets/top/ebpf/piditer/iter.go
```diff
@@ -15,17 +15,15 @@
...
-	"io"
 	"unsafe"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
-	"golang.org/x/sys/unix"
-
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/kallsyms"
+
```
mauriciovasquezbernal, line 25: Remove this empty line?

mauriciovasquezbernal (commented):
From a code point of view it looks good mostly. Some question I have:
- Is  not running in the host pid namespace safer even if we're able to contact systemd to perform things like this? 
- I suppose this mode is a "compatibility" mode and in most of the cases systemd won't be used, right?

alban:
> Is not running in the host pid namespace safer even if we're able to contact systemd to perform things like this?

I think `hostPID=false` does not improve security if a malicious user already has full control of the gadget container. But with this PR, we will be able to reduce the amount of code that is running in the host pidns. And often, security attacks only give partial control of the container, so it is still valuable to reduce the attack surface.

> I suppose this mode is a "compatibility" mode and in most of the cases systemd won't be used, right?

systemd will not be used in this PR, but later when we actually switch to `hostPID=false`, then systemd will be used for gadgets that require it (socket enricher initialization, top ebpf, snapshot process). But it will not be used for most gadgets.

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
alban: Yes (but host.HostRoot, not host.HostProcFs)

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
alban: None, I forgot to remove it.

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
alban: I think it should be fast, but I can change it to 10s.

alban (commented):
Branch updated.

pkg/gadgets/snapshot/process/tracer/tracer.go
```diff
@@ -86,15 +87,15 @@ func runeBPFCollector(config *Config, enricher gadgets.DataEnricherByMntNs) ([]*
 	}
 	defer dumpTaskIter.Close()
 
-	file, err := dumpTaskIter.Open()
+	buf, err := bpfiterns.Read(dumpTaskIter)
 	if err != nil {
-		return nil, fmt.Errorf("opening BPF iterator: %w", err)
+		fmt.Printf("error reading iterator: %v\n", err)
```
mauriciovasquezbernal, line 92: letfover? Also, missing error wrapping in next line?

mauriciovasquezbernal (approved):
Some nits around. LGTM from code inspection.

PR #1710, opened 2023-06-05
Open network namespaces from /host/proc (preparation for hostPID=false)

alban:
# Open network namespaces from /host/proc (preparation for hostPID=false)

As part of https://github.com/inspektor-gadget/inspektor-gadget/issues/1658, we want Inspektor Gadget to support deployments with hostPID=false.

Since Inspektor Gadget can run in a container with a new pid namespace, pid numbered in the host pid namespace have to be looked up in /host/proc instead of /proc.

There are a couple of places where /proc was used instead of /host/proc:
- OpenRawSock() for socket filters in networking gadgets (dns, sni)
- NetnsEnter() for the snapshot socket gadgets: tcp/udp iterators have to be opened from the correct netns

Both were using the vishvananda/netns library. We need a patch in that library:
https://github.com/vishvananda/netns/pull/76

## How to use

No changes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -149,3 +149,5 @@ require (
 )
 
 replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.13.1-0.20230315234915-a26de2d610c3
+
+replace github.com/vishvananda/netns => github.com/inspektor-gadget/netns v0.0.5-0.20230524185006-155d84c555d6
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -149,3 +149,5 @@ require (
 )
 
 replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.13.1-0.20230315234915-a26de2d610c3
+
+replace github.com/vishvananda/netns => github.com/alban/netns v0.0.5-0.20230524185006-155d84c555d6
```
mauriciovasquezbernal, line 153: What about moving that fork into this organization?

mauriciovasquezbernal (approved):
LGTM. Just one comment about the location of the fork.

go.mod
alban: Ok!

PR #1715, opened 2023-06-07
integration: Don't use regexp on advise network tests

mauriciovasquezbernal:


integration/inspektor-gadget/advise_networkpolicy_test.go
```diff
@@ -41,25 +47,92 @@ func TestAdviseNetworkpolicy(t *testing.T) {
...
+					Proto:           "tcp",
+					PodLabels:       map[string]string{"run": "test-pod"},
+					Port:            9090,
+					RemoteKind:      eventtypes.RemoteKindService,
+					RemoteNamespace: nsServer,
+					RemoteName:      "test-pod",
+					RemoteLabels:    map[string]string{"run": "test-pod"},
+				}
+
+				expectedEntry.Container = ""
```
alban, line 67: Note: the "trace network" gadget sets the field `Container` via enrichers, but "advise network-policy" does not because it does not use the operators yet (it still uses the CR path). We can remove this line from the test once the gadget gets converted.

alban (approved):
LGTM if the tests pass!

PR #1716, opened 2023-06-07
[WIP] design: Add containerized gadgets (aka byob) design doc

alban:
Add document with the design of the containerized gadget support in IG.

## How to use

## Testing done

### DIFF
--- a/docs/design/002-containerized-gadgets.md
+++ b/docs/design/002-containerized-gadgets.md
@@ -1,0 +1,226 @@
+# Support for containerized gadgets in Inspektor Gadget
+
+TODO: intro
+
+# Goals
+
+TODO
+
+# Categories of eBPF programs
+
+Supported categories of eBPF programs:
+
+| ebpf program type     | Support | Difficulty |
+|-----------------------|---------|------------|
+| socket                | 👷      |            |
+| sk_reuseport/migrate  |         |            |
+| sk_reuseport          |         |            |
+| kprobe/               | ✅      |            |
+| uprobe/               |         |            |
+| kretprobe/            | ✅      |            |
+| uretprobe/            |         |            |
+| tc                    |         |            |
+| classifier            |         |            |
+| action                |         |            |
+| tracepoint/           | 👷      |            |
+| tp/                   |         |            |
+| raw_tracepoint/       |         |            |
+| raw_tp/               |         |            |
+| raw_tracepoint.w/     |         |            |
+| raw_tp.w/             |         |            |
+| tp_btf/               |         |            |
+| fentry/               |         |            |
+| fmod_ret/             |         |            |
+| fexit/                |         |            |
+| fentry.s/             |         |            |
+| fmod_ret.s/           |         |            |
+| fexit.s/              |         |            |
+| freplace/             |         |            |
+| lsm/                  |         |            |
+| lsm.s/                |         |            |
+| iter/                 | 📅      |            |
+| iter.s/               | 📅      |            |
+| syscall               |         |            |
+| xdp_devmap/           |         |            |
+| xdp_cpumap/           |         |            |
+| xdp                   |         |            |
+| perf_event            |         |            |
+| lwt_in                |         |            |
+| lwt_out               |         |            |
+| lwt_xmit              |         |            |
+| lwt_seg6local         |         |            |
+| cgroup_skb/ingress    |         |            |
+| cgroup_skb/egress     |         |            |
+| cgroup/skb            |         |            |
+| cgroup/sock_create    |         |            |
+| cgroup/sock_release   |         |            |
+| cgroup/sock           |         |            |
+| cgroup/post_bind4     |         |            |
+| cgroup/post_bind6     |         |            |
+| cgroup/dev            |         |            |
+| sockops               |         |            |
+| sk_skb/stream_parser  |         |            |
+| sk_skb/stream_verdict |         |            |
+| sk_skb                |         |            |
+| sk_msg                |         |            |
+| lirc_mode2            |         |            |
+| flow_dissector        |         |            |
+| cgroup/bind4          |         |            |
+| cgroup/bind6          |         |            |
+| cgroup/connect4       |         |            |
+| cgroup/connect6       |         |            |
+| cgroup/sendmsg4       |         |            |
+| cgroup/sendmsg6       |         |            |
+| cgroup/recvmsg4       |         |            |
+| cgroup/recvmsg6       |         |            |
+| cgroup/getpeername4   |         |            |
+| cgroup/getpeername6   |         |            |
+| cgroup/getsockname4   |         |            |
+| cgroup/getsockname6   |         |            |
+| cgroup/sysctl         |         |            |
+| cgroup/getsockopt     |         |            |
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (commented):
Some comments. Will continue reviewing it.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| profile tcprtt         |                                                                                 |            |
+| snapshot process       |                                                                                 |            |
+| snapshot socket        |                                                                                 |            |
+| top block-io           |                                                                                 |            |
+| top ebpf               |                                                                                 |            |
+| top file               |                                                                                 |            |
+| top tcp                |                                                                                 |            |
+| trace bind             | field_filter, bitfield_column, enum_convert, ifindex                            |            |
+| trace capabilities     |                                                                                 |            |
```
mauriciovasquezbernal, line 159: - `enum_convert`
- int -> syscall name: Can we implement this as enum_convert given that syscalls are different per architecture?

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+
+| Gadget                 | Feature needed                                                                  | Difficulty |
+|------------------------|---------------------------------------------------------------------------------|------------|
+| advise network-policy  |                                                                                 |            |
+| advise seccomp-profile |                                                                                 |            |
+| audit seccomp          |                                                                                 |            |
+| profile block-io       |                                                                                 |            |
+| profile cpu            |                                                                                 |            |
+| profile tcprtt         |                                                                                 |            |
+| snapshot process       |                                                                                 |            |
```
mauriciovasquezbernal, line 152: It supports a compatibility mode to get processes when iterators are not available -> https://github.com/inspektor-gadget/inspektor-gadget/blob/e91a93bb2c0a7d05ed6ac2689b89ec7465923dcf/pkg/gadgets/snapshot/process/tracer/tracer.go#L216. I support we'll lose that in the containerized version.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| snapshot socket        |                                                                                 |            |
```
mauriciovasquezbernal, line 153: I think it's the most challenging one since its mechanism to run the iterator for each network namespace.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| top block-io           |                                                                                 |            |
```
mauriciovasquezbernal, line 154: - It has some logic to change the kprobe name based on the kernel support https://github.com/inspektor-gadget/inspektor-gadget/blob/9d8f024001c38d5677308b8d67289c5bcaf90d43/pkg/gadgets/top/block-io/tracer/tracer.go#L117-L130

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| top ebpf               |                                                                                 |            |
+| top file               |                                                                                 |            |
```
mauriciovasquezbernal, line 156: Nothing to worry about.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| top tcp                |                                                                                 |            |
```
mauriciovasquezbernal, line 157: - `ip_ver`

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| advise seccomp-profile |                                                                                 |            |
+| audit seccomp          |                                                                                 |            |
+| profile block-io       |                                                                                 |            |
+| profile cpu            |                                                                                 |            |
+| profile tcprtt         |                                                                                 |            |
+| snapshot process       |                                                                                 |            |
+| snapshot socket        |                                                                                 |            |
+| top block-io           |                                                                                 |            |
+| top ebpf               |                                                                                 |            |
```
mauriciovasquezbernal, line 155: This gadget doesn't use eBPF, so I think it can't be ported to the containerized approach.

mauriciovasquezbernal (commented):
I checked some more gadgets to see what features they need.

docs/design/002-containerized-gadgets.md
alban: It uses pkg/gadgets/top/ebpf/piditer/bpf/pid_iter.bpf.c to link the ebpf programs with processes. But I also don't see how the gadget could be done with the containerized approach.

By the way, it has:
```
const volatile __u64 bpf_prog_fops_addr = 0;
```
with initialization from userspace. But no real gadgets use that.

docs/design/002-containerized-gadgets.md
alban: I think it is simple to do:
* iter/bpf_map_elem: run in current namespace
* iter/task_file and iter/task: run in host pidns (since pidns are hierarchical)
* iter/tcp and iter/udp: run in all netns (since netns are independent)

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+| top tcp                |                                                                                 |            |
+| trace bind             | field_filter, bitfield_column, enum_convert, ifindex                            |            |
+| trace capabilities     |                                                                                 |            |
+| trace dns              | socket_filter, socket_enricher, enum_convert, string_edit, ip_ver, latency_calc |            |
+| trace exec             |                                                                                 |            |
+| trace fsslower         |                                                                                 |            |
+| trace mount            |                                                                                 |            |
+| trace oomkill          |                                                                                 |            |
+| trace open             |                                                                                 |            |
```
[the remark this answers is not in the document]
alban: > Attach probe based on architecture

The bpf source code would just need to add the correct #ifdef to compile out the program that's not used.

OCI images can have different layers depending on the architecture, so IG would download either the arm64 or the x86 OCI image, and that would contain a different ELF file.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,226 @@
...
+Alternatively, the bpf module can contain a global variable with the same name as the iterator plus a suffix "_output".
+So the BTF information can be attached to that variable.
+```
+const volatile event_t myiterator_output = {};
+
+SEC("iter/task_file")
+int myiterator(struct bpf_iter__task_file *ctx)
+```
+
+Problem: if that variable is not used by any program, the cilium/ebpf loader will complain.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: ^^^ didn't work. Creating a constant seems to work: `const volatile struct myevent foo = {};`.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: mmm, good point. We can define that approach and implement that logic inside Inspektor Gadget.

mauriciovasquezbernal:
Closing in favor of https://github.com/inspektor-gadget/inspektor-gadget/pull/1856.

PR #1718, opened 2023-06-07
pkg: Log verifier error at debug level for gadget with result

eiffel-fl:
Fixes: 829717b6a68d ("pkg: Log verifier error at debug level.")


mauriciovasquezbernal (approved):
Thanks for the fix!

PR #1719, opened 2023-06-07
bpf-iter-ns: fix bpf pin path

alban:
I didn't set the correct path when addressing this comment:
https://github.com/inspektor-gadget/inspektor-gadget/pull/1708#discussion_r1220175307

Fixes: e91a93bb2c0a ("Read bpf iterators from host pid ns")

### DIFF
--- a/pkg/utils/bpf-iter-ns/bpf-iter-ns.go
+++ b/pkg/utils/bpf-iter-ns/bpf-iter-ns.go
@@ -96,16 +96,22 @@ func ReadOnHostPidNs(iter *link.Iter) ([]byte, error) {
 		return nil, fmt.Errorf("empty /proc/self symlink")
 	}
 
-	tmpPinDir, err := os.MkdirTemp(host.HostRoot+"/sys/fs/bpf", "ig-iter-")
+	// Create a temporary directory in the host bpffs
+	bpfFS := "/sys/fs/bpf"
+	tmpPinDir, err := os.MkdirTemp(host.HostRoot+bpfFS, "ig-iter-")
 	if err != nil {
 		return nil, fmt.Errorf("creating temporary directory in bpffs: %w", err)
 	}
+	defer os.RemoveAll(tmpPinDir)
+
+	// Prepare the pin path from the container and host point of view
 	pinPath := filepath.Join(tmpPinDir, "iter")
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
Does it fail because `iter.Pin()` must be under `"/sys/fs/bpf"`?

Changes LGTM!

alban:
> Does it fail because iter.Pin() must be under "/sys/fs/bpf"?

No, `iter.Pin()` didn't fail with the "/host/sys/fs/bpf..." path. But then the systemd unit was trying to access it with the same path, which is wrong because /host does not exist from the systemd unit's point of view.

pkg/utils/bpf-iter-ns/bpf-iter-ns.go
```diff
@@ -96,16 +96,22 @@ func ReadOnHostPidNs(iter *link.Iter) ([]byte, error) {
...
+	bpfFS := "/sys/fs/bpf"
+	tmpPinDir, err := os.MkdirTemp(host.HostRoot+bpfFS, "ig-iter-")
 	if err != nil {
 		return nil, fmt.Errorf("creating temporary directory in bpffs: %w", err)
 	}
+	defer os.RemoveAll(tmpPinDir)
+
+	// Prepare the pin path from the container and host point of view
 	pinPath := filepath.Join(tmpPinDir, "iter")
+	pinPathHost := filepath.Join(bpfFS, filepath.Base(tmpPinDir), "iter")
```
mauriciovasquezbernal, line 109: I got a bit confused, I thought `Host` meant it had the `/host` prefix, but it means it's from the host point of view. it doesn't seem a big deal to me.

alban:
Oops, I merged this too early because that does not work in some setups: the host might not have bpffs mounted but the gadget container always has bpffs mounted, see:
https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.17.0/gadget-container/entrypoint.sh#L50-L59

So I'll need to do it the other way around:
- in the container, pin on /sys/fs/bpf directly
- on the host, read the iterator via /proc/$GADGET_PID/root/sys/fs/bpf

Thankfully, this PR does not break main because hostPID is not yet disabled.

PR #1723, opened 2023-06-08
gha: Run AKS and ARO integration tests for `citest/` branches too.

mauriciovasquezbernal:
It's useful for debugging purposes to avoid having to setup all of this
    in a personal fork or pushing to main.
    
    Signed-off-by: Mauricio Vásquez <mauriciov@microsoft.com>

---

I tried to implement a solution to automatically remove the image in the container repository but I'm too afraid that I can get this wrong and remove important images, hence I prefer to leave this as a manual step. 

---

Tested in https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/5212558595, I don't have AKS secrets configured but the image was pushed to https://github.com/mauriciovasquezbernal/inspektor-gadget/pkgs/container/inspektor-gadget, so it should work all fine :tm: .

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -793,7 +793,13 @@ jobs:
   test-integration-aks:
     name: Integration tests on AKS
     # level: 2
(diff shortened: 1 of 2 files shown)


alban (approved):
LGTM

PR #1726, opened 2023-06-09
kubectl-gadget/top: Ensure timeout=interval prints event before exiting

### LINKED ISSUE #716: top gadgets don't print information when using the same interval and timeout
## Summary 

It's possible to run the top gadgets in a "single shot" fashion, i.e, when the interval and timeout values are the same (for instance `./kubectl-gadget top file 2 --timeout 2`). In this configuration the gadget should print a single summary of the events and exit, however it could happen that in some cases nothing is printed. This was implemented in https://github.com/kinvolk/inspektor-gadget/pull/583. 

## The issue
In the top gadgets each node pushes a summary of the events each interval https://github.com/kinvolk/inspektor-gadget/blob/008667ba8192f53c8af25d52e2324e2d06708020/pkg/gadgets/filetop/tracer/tracer.go#L216-L237 and https://github.com/kinvolk/inspektor-gadget/blob/008667ba8192f53c8af25d52e2324e2d06708020/pkg/gadgets/filetop/gadget.go#L176, then those are collected 
https://github.com/kinvolk/inspektor-gadget/blob/008667ba8192f53c8af25d52e2324e2d06708020/cmd/kubectl-gadget/top/file.go#L114 and printed by the kubectl-gadget client  https://github.com/kinvolk/inspektor-gadget/blob/008667ba8192f53c8af25d52e2324e2d06708020/cmd/kubectl-gadget/top/file.go#L140. 

There is not any synchronization used among kubectl-gadget and the nodes, it can happen that some updates are missed but it is not a big problem, those will be considered in the next iteration, however this creates some problems when `--timeout` is used as it could happen that some events are lost, printing nothing to the terminal. 

I've been able to reproduce it when loading the CPU of the node. Please notice that this is a one node setup and kubectl-gadget is run on the same node too. The only difference between the following two cases is that one sleeps on each iteration. 

## Case that works fine

```bash
$ kubectl run ubuntu --image ubuntu:latest -- /bin/bash -c 'while true; do date >> foo.txt && sleep 0.1; done'
pod/ubuntu created
$ kubectl wait --for=condition=Ready pod/ubuntu
pod/ubuntu condition met
./kubectl-gadget top file 2 --timeout 2

NODE             NAMESPACE        POD              CONTAINER        PID     COMM             READS  WRITES R_Kb    W_Kb    T FILE
ubuntu-hirsute   default          ubuntu           ubuntu           3831900 sleep            5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831916 sleep            5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831931 date             5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831889 date             5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831938 sleep            5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831926 date             0      1      0       0       R foo.txt
ubuntu-hirsute   default          ubuntu           ubuntu           3831924 date             0      1      0       0       R foo.txt
ubuntu-hirsute   default          ubuntu           ubuntu           3831925 sleep            5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831939 date             5      0      2       0       R libc.so.6
ubuntu-hirsute   default          ubuntu           ubuntu           3831899 date             5      0      2       0       R libc.so.6
[...]
$ kubectl delete pod ubuntu
```

## Reproducer
```bash
$ kubectl run ubuntu --image ubuntu:latest -- /bin/bash -c 'while true; do date >> foo.txt; done'
pod/ubuntu created
$ kubectl wait --for=condition=Ready pod/ubuntu
pod/ubuntu condition met
$ ./kubectl-gadget top file 2 --timeout 2

NODE             NAMESPACE        POD              CONTAINER        PID     COMM             READS  WRITES R_Kb    W_Kb    T FILE
[ nothing is printed]

$ kubectl delete pod ubuntu
```

## Possible solution

One possibility would be to add some synchronization between kubectl-gadget and the nodes, but it seems rather difficult to implement. A complicated algorithm waiting for results of all nodes should be implemented. Another possibility can be to switch from a "push" to a "pull" approach, i.e, instead of waiting for the nodes to "push" the results, the client could "pull" the results from all nodes each interval, this solution seems easier to implement in the long term, however it requires big changes to Inspektor Gadget as we don't have a way to implement it now. 

Some discussion on slack regarding the top gadgets implementation -> https://kubernetes.slack.com/archives/CSYL75LF6/p1655846371253189?thread_ts=1655287566.910119&cid=CSYL75LF6.

mqasimsarfraz:
As mentioned in the comment [here](https://github.com/inspektor-gadget/inspektor-gadget/issues/716#issuecomment-1581315275) the [snapshot ticker](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/parser/parser.go#L159) can go off bit early compared to when we [receive](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/runtime/grpc/grpc-runtime.go#L310) events. Which should be fine in case user is running gadget for multiple intervals since the missed events will be printed in following interval. But this is problematic when gadget is only ran for single interval (`timeout==interval`). 

The idea with this change it to introduce a flush mechanism to ensure all the outstanding events are sent downstream for handling before exiting. Timeout timer is running on the server side, ensuring we receive all the events from server before calling `Flush()`. Following diagram (courtesy: @flyth) shows the flow with the fix.

```mermaid
sequenceDiagram
    participant Local as Local (kubectl gadget)
    participant Remote as Remote (gadget pods)
    Local->>Local: Start Snapshot Timer
    Local->>Remote: Start Gadget
    Remote->>Remote: Start Gadget Timer
    loop every interval
        Local->>Local: SnapshotCombiner Timer fires
        Local->>Local: Get Events from SnapshotCombiner
        Remote->>Local: New Event
        Local->>Local: Add Events to SnapshotCombiner
    end
    alt end
        Remote->>Remote: Timer done
        Remote->>Local: Send last results
    end
    Local->>Local: Wait for all connections to be done
    Local->>Local: Flush "last results"
```

Fixes #716 

## Testing done

Integration test adapted to test `timeout==interval`, Also failure results before the fix:

- [Failing CI w/o fix](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/5220885409) - [Summary](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/5220885409/attempts/1#summary-14133592143)

### DIFF
--- a/integration/inspektor-gadget/top_blockio_test.go
+++ b/integration/inspektor-gadget/top_blockio_test.go
@@ -18,53 +18,82 @@ import (
 	"fmt"
 	"testing"
 
-	topblockioTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/block-io/types"
-
 	. "github.com/inspektor-gadget/inspektor-gadget/integration"
+	topblockioTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/top/block-io/types"
 )
+
+func newTopBlockIOCmd(ns string, cmd string, startAndStop bool) *Command {
+	expectedOutputFn := func(output string) error {
+		expectedEntry := &topblockioTypes.Stats{
+			CommonData: BuildCommonData(ns),
+			Write:      true,
+			Comm:       "dd",
+		}
+
+		normalize := func(e *topblockioTypes.Stats) {
+			e.Node = ""
+			e.Major = 0
+			e.Minor = 0
+			e.MicroSecs = 0
+			e.MountNsID = 0
+			e.Pid = 0
+			e.Operations = 0
+			e.Bytes = 0
+		}
+
+		return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
+	}
+	return &Command{
+		Name:             "TopBlockIO",
+		ExpectedOutputFn: expectedOutputFn,
+		Cmd:              cmd,
+		StartAndStop:     startAndStop,
+	}
+}
 
 func TestTopBlockIO(t *testing.T) {
 	if *k8sDistro == K8sDistroARO {
 		t.Skip("Skip running top block-io gadget on ARO: see issue #589")
 	}
 
+	t.Parallel()
 	ns := GenerateTestNamespaceName("test-top-block-io")
 
-	t.Parallel()
-
-	topBlockIOCmd := &Command{
-		Name:         "StartTopBlockIOGadget",
-		Cmd:          fmt.Sprintf("$KUBECTL_GADGET top block-io -n %s -o json", ns),
-		StartAndStop: true,
-		ExpectedOutputFn: func(output string) error {
-			expectedEntry := &topblockioTypes.Stats{
-				CommonData: BuildCommonData(ns),
-				Write:      true,
-				Comm:       "dd",
-			}
-
-			normalize := func(e *topblockioTypes.Stats) {
-				e.Node = ""
-				e.Major = 0
-				e.Minor = 0
-				e.MicroSecs = 0
-				e.MountNsID = 0
-				e.Pid = 0
-				e.Operations = 0
-				e.Bytes = 0
-			}
-
-			return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
-		},
-	}
-
-	commands := []*Command{
+	commandsPreTest := []*Command{
 		CreateTestNamespaceCommand(ns),
-		topBlockIOCmd,
 		BusyboxPodRepeatCommand(ns, "dd if=/dev/zero of=/tmp/test count=4096"),
 		WaitUntilTestPodReadyCommand(ns),
-		DeleteTestNamespaceCommand(ns),
 	}
+	RunTestSteps(commandsPreTest, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
 
-	RunTestSteps(commands, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	t.Cleanup(func() {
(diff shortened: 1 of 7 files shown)


integration/inspektor-gadget/top_blockio_test.go
```diff
@@ -23,48 +23,78 @@ import (
...
-			return ExpectEntriesInMultipleArrayToMatch(output, normalize, expectedEntry)
-		},
-	}
-
-	commands := []*Command{
+	commandsPreTest := []*Command{
 		CreateTestNamespaceCommand(ns),
-		topBlockIOCmd,
-		BusyboxPodRepeatCommand(ns, "dd if=/dev/zero of=/tmp/test count=4096"),
+		BusyboxPodRepeatCommand(ns, "dd if=/dev/zero of=/tmp/test count=4096; sleep 0.2"),
```
flyth, line 65: Can you please add a comment regarding the `sleep 0.2`?

integration/inspektor-gadget/top_ebpf_test.go
```diff
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
```
flyth, line 18: nit: since you're editing imports here, maybe reorder the imports below as well

flyth (approved):
Looks great! Thanks for adding tests specifically for this issue. Hopefully we won't run into it again :)

PR #1738, opened 2023-06-13
cmd: move verbose flag to top level

flyth:
This change adds a global verbose flag that affects all logging operations using logrus instead of implementing it per command.


mauriciovasquezbernal (approved):
LGTM!

PR #1739, opened 2023-06-13
cmd/kubectl-gadget: add support to dynamically deploy multiple embedded CRD

flyth:
This change will import multiple CRD yaml definitions stored in `resources/crd/bases` (which are generated using `crd.mk`/kubebuilder). Additionally, rules will be added to our ClusterRole to allow permissions for all verbs on our custom resources.
On undeploy, it will retrieve a list the installed CRD and remove those belonging to Inspektor Gadget.

## Why?

For now, this is a prerequisite to add back the persistent gadget behavior using custom resources (a new CRD is being used for that). As we might in the future implement more custom resources for "external gadgets" as well, this hopefully helps adding them more easily to our deploy/undeploy routines.

## Other notes

Additionally to our current API group `gadget.kinvolk.io`, this PR adds support for the API group `inspektor-gadget.io`, which IMHO we should use in the future.

### DIFF
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -19,6 +19,7 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"reflect"
 	"strconv"
 	"strings"
@@ -27,6 +28,7 @@ import (
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


mqasimsarfraz:
I like the idea but I think we should use `kubebuilder` generation since most of the controller follow that approach. I did an experiment with generating our cluster role [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/96f7a064901781a478b84e94ae8fe74b60ce5f92). Also, It is already embedded to [RbacRole](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/resources/embed.go#L25) but not used yet. Maybe we can also use that approach for CRDs [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/apis/gadget/v1alpha1/groupversion_info.go#L17). What do you say?

flyth:
> I like the idea but I think we should use kubebuilder generation since most of the controller follow that approach. I did an experiment with generating our cluster role [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/96f7a064901781a478b84e94ae8fe74b60ce5f92). Also, It is already embedded to [RbacRole](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/resources/embed.go#L25) but not used yet. Maybe we can also use that approach for CRDs [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/apis/gadget/v1alpha1/groupversion_info.go#L17). What do you say?

If that can replace everything we do in `pkg/resources/manifests/deploy.yaml` regarding the role, then sure! Sounds like a much cleaner/better solution.

flyth:
Closing in favor of @mqasimsarfraz's suggestion.

PR #1741, opened 2023-06-14
pkg: Enable to capture data for only host processes.

eiffel-fl:
Hi.

In this PR, I added support to capture data for host processes only:

```bash
$ sudo ig exec trace --host
CONTAINER                                               PID        PPID       COMM             RET ARGS                                                                    
                                                        79765      2167       runc             0   /usr/bin/runc --version                                                 
                                                        79771      2167       docker-init      0   /usr/libexec/docker/docker-init --version                               
                                                        79777      4304       ls               0   /usr/bin/ls --color=auto                                                
```

Best regards.

### DIFF
--- a/docs/ig.md
+++ b/docs/ig.md
@@ -114,6 +114,8 @@ simplicity, they are not demonstrated in each command guide:
   `--output` flag.
 - It is possible to filter events by container name using the `--containername`
   flag.
+- It is possible to trace events from all the running processes, even though
+  they were not generated from containers, using the `--host` flag.
 
 For instance, for the `list-containers` command:
 
@@ -137,3 +139,21 @@ $ sudo ig list-containers -o json --containername etcd
   }
 ]
 ```
+
+For example, with `--host`, you can get the following output:
+
+```bash
+$ sudo ig trace exec --host
+CONTAINER                                               PID        PPID       COMM             RET ARGS
+
+# Open another terminal.
+$ cat /dev/null
+$ docker run --name test-host --rm -t debian sh -c 'ls > /dev/null'
+# Go back to first terminal.
+CONTAINER                                               PID        PPID       COMM             RET ARGS
+                               24640            4537             cat              0   /usr/bin/cat /dev/null
+test-host                      24577            24553            sh               0   /bin/sh -c cat /dev/null
+test-host                      24598            24577            cat              0   /bin/cat /dev/null
+```
+
+Events generated from containers have their container field set, while events which are generated from the host do not.


burak-ok:
I like the idea, thanks.
Regarding the now empty column `Container`. Maybe we could expand on https://github.com/inspektor-gadget/inspektor-gadget/pull/1587 and use it here too.

For me in an Ubuntu docker container when running sleep I still get the following output. Am I doing something wrong:
```
{12:04}~/codes/inspektor-gadget-2:francis/host-events ✓ ➭ go run -exec sudo ./cmd/ig/... trace exec --host
CONTAINER                                             PID        PPID       COMM             RET ARGS                                                                 
nostalgic_driscoll                                    359339     310815     sleep            0   /usr/bin/sleep 1                                                     
....                                               
nostalgic_driscoll                                    356891     310815     sleep            0   /usr/bin/sleep 1              
```

alban:
We have different categories of mount namespaces:
1. The so-called "host" mount namespace, from pid 1.
2. Mount namespaces of containers supported by ig (such as docker, runc, podman, crun)
3. Mount namespaces of containers unsupported by ig (such as bubblewrap, systemd-nspawn)
4. Other mount namespaces on the host, such as the ones created by systemd for some systemd unit with mount namespace settings such as `PrivateTmp`.
5. The "host of the host", e.g. when running minikube with the docker driver

Category 3, 4 and 5 cannot be distinguished by ig. But ig should be able to display them.

alban:
This can't be done with only the `mountnsmap` ebpf map since it is an "opt-in" for mount namespaces and this map can't express "all namespaces, including the ones we don't know exist".

We need to use the boolean `gadget_filter_by_mntns` (already defined in `pkg/gadgets/common/mntns_filter.h` so we don't need to patch the ebpf code).

alban:
> * `--mntnsid`: it takes a mount namespace ID to filter by, it catches the above case pointed by Alban.
> * `--host`: a boolean, it filter out all the events which do not occur in the root mount namespace.

I don't think `--mntnsid` should be visible to users. We should have the same flags in kubectl-gadget and `mntnsid` is not something stable across nodes.

And I think `--host` should include all events from the host including both in the systemd mount namespace and the mount namespaces of all systemd units.  What matters to users is whether it is in a container or not. Using the mount namespace id is an implementation detail of Inspektor Gadget.

@mauriciovasquezbernal What do you think?

mauriciovasquezbernal:
> I don't think --mntnsid should be visible to users. We should have the same flags in kubectl-gadget and mntnsid is not something stable across nodes.

I agree. That should be internal to IG. 

> And I think --host should include all events from the host including both in the systemd mount namespace and the mount namespaces of all systemd units. What matters to users is whether it is in a container or not. Using the mount namespace id is an implementation detail of Inspektor Gadget.

I think the main confusion here is the meaning of `host`. It can be either (related to 5 cases described by Alban above)
1. All events happening in the host (including containers, systemd units, etc)
2. Events happening in the mount ns of pid 1
3. Events happening outside containers we support

I think in the long term we should support the following possibilities:
1. Gathering all events that happen in the host (first point above). I slightly leaned for using `--host` as it's clear what the host is and also because `--all` collides with `-A` from Kubernetes.
2. Events outside containers. I think it's useful to understand the activity done in the host but that is not part of a container. For this case I think we should start building on the `-F/--filter` support, IMO we can't keep adding flags for each use case we find. For instance, this case could be something like `-F runtime.ContainerID:!""` (exact syntax to be defined) -> show all events that don't have the containerID set. This is currently implemented in userspace but we can optimize it to support some filtering directly into eBPF. 
3. Filtering by systemd units. (TBD because I have no idea what it could look like)

For this PR implementing point 1 is enough. We can handle the other case later on. What do you all think?

/cc @flyth as you're interested in the filtering part.

cmd/ig/utils/flags.go
```diff
@@ -107,6 +111,13 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		"",
 		"Show only data from containers with that name",
 	)
+	command.PersistentFlags().BoolVarP(
+		&commonFlags.Host,
+		"host",
+		"",
+		false,
+		"Show data from all processes",
```
alban, line 119:
```suggestion
		"Show data from both the host and containers",
```
Or:
```suggestion
		"Include events from the host",
```

docs/ig.md
```diff
@@ -137,3 +139,21 @@ $ sudo ig list-containers -o json --containername etcd
...
 ```
+
+For example, with `--host`, you can get the following output:
+
+```bash
+$ sudo ig trace exec --host
+CONTAINER                                               PID        PPID       COMM             RET ARGS
+
+# Open another terminal.
+$ ls > /dev/null
```
alban, line 150: I suggest to use a different command in the container and outside to make it easier to understand the `ig` output.

```suggestion
$ cat /dev/null
```

docs/ig.md
```diff
@@ -137,3 +139,21 @@ $ sudo ig list-containers -o json --containername etcd
...
+$ docker run --name test-host --rm -t debian sh -c 'ls > /dev/null'
+# Go back to first terminal.
+CONTAINER                                               PID        PPID       COMM             RET ARGS
+                               5671             4040             ls               0   /usr/bin/ls --color=auto
+test-host                      5757             5735             sh               0   /bin/sh -c ls > /dev/null
+test-host                      5778             5757             ls               0   /bin/ls
+```
+
+As you can see, events generated from containers have their container field set, while events which are generated from root mount namespace do not.
```
alban, line 159:
```suggestion
As you can see, events generated from containers have their container field set, while events which are generated from the host do not.
```

I would avoid saying "root mount namespace" because that's also true for other mount namespaces as long as they are not from containers.

integration/ig/k8s/trace_exec_test.go
```diff
@@ -89,3 +90,64 @@ func TestTraceExec(t *testing.T) {
...
+					Comm:  "sleep",
+					Args:  sleepArgs,
+				},
+			}
+
+			normalize := func(e *execTypes.Event) {
+				// Event is running on the host, so no containers related data.
+				e.Namespace = ""
+				e.Pod = ""
+				e.Container = ""
```
alban, line 126: Since you expect those fields to be empty, it should not be reset to "" in the normalize function, otherwise the test would not fail if they are not empty.

pkg/operators/localmanager/localmanager.go
```diff
@@ -111,6 +112,12 @@ func (l *LocalManager) ParamDescs() params.ParamDescs {
 			Description: "Show only data from containers with that name",
 			ValueHint:   gadgets.LocalContainer,
 		},
+		{
+			Key:          Host,
+			Description:  "Show only data from all processes",
```
alban, line 117:
```suggestion
			Description:  "Include events from the host",
```

alban (commented):
Thanks, that looks good so far. I've made some phrasing suggestions. I will do another review when the CI is green.

docs/ig.md
```diff
@@ -137,3 +139,21 @@ $ sudo ig list-containers -o json --containername etcd
...
+CONTAINER                                               PID        PPID       COMM             RET ARGS
+
+# Open another terminal.
+$ cat /dev/null
+$ docker run --name test-host --rm -t debian sh -c 'ls > /dev/null'
+# Go back to first terminal.
+CONTAINER                                               PID        PPID       COMM             RET ARGS
+                               5671             4040             ls               0   /usr/bin/cat /dev/null
+test-host                      5757             5735             sh               0   /bin/sh -c ls > /dev/null
+test-host                      5778             5757             ls               0   /bin/ls
```
mauriciovasquezbernal, line 156: Please fix the aligment of the columns header and the content

pkg/operators/localmanager/localmanager.go
```diff
@@ -111,6 +112,12 @@ func (l *LocalManager) ParamDescs() params.ParamDescs {
 			Description: "Show only data from containers with that name",
 			ValueHint:   gadgets.LocalContainer,
 		},
+		{
+			Key:          Host,
+			Description:  "Include events from the host",
```
mauriciovasquezbernal, line 117: Please keep is aligned with the description in cmd/ig/utils/flags.go.

docs/ig.md
```diff
@@ -137,3 +139,21 @@ $ sudo ig list-containers -o json --containername etcd
...
+```
+
+As you can see, events generated from containers have their container field set, while events which are generated from the host do not.
```
mauriciovasquezbernal, line 159:
```suggestion
Events generated from containers have their container field set, while events which are generated from the host do not.
```

mauriciovasquezbernal (approved):
LGTM. Great!

mauriciovasquezbernal:
@eiffel-fl please hold this a bit. I'm checking a possible issue. I'll be back soon.

mauriciovasquezbernal:
This is not working for gadgets that rely on the `Attacher` interface, like trace dns, sni, etc. because they use the Subscribe mechanism from the container collection that takes a container selector. I hacked it to make it work for some cases with the following diff:
```diff
$ git diff
diff --git pkg/operators/localmanager/localmanager.go pkg/operators/localmanager/localmanager.go
index 32a25c18..5897e164 100644
--- pkg/operators/localmanager/localmanager.go
+++ pkg/operators/localmanager/localmanager.go
@@ -317,12 +317,21 @@ func (l *localManagerTrace) PreGadgetRun() error {
                for _, container := range containers {
                        attachContainerFunc(container)
                }
+
+               // attach fake container with host pid
+               if host {
+                       attachContainerFunc(&containercollection.Container{
+                               Pid: 1,
+                       })
+               }
        }
 
        return nil
 }
 
 func (l *localManagerTrace) PostGadgetRun() error {
+       host := l.params.Get(Host).AsBool()
+
        if l.mountnsmap != nil {
                log.Debugf("calling RemoveMountNsMap()")
                l.manager.igManager.RemoveMountNsMap(l.subscriptionKey)
@@ -335,6 +344,12 @@ func (l *localManagerTrace) PostGadgetRun() error {
                for container := range l.attachedContainers {
                        l.attacher.DetachContainer(container)
                }
+
+               if host {
+                       l.attacher.DetachContainer(&containercollection.Container{
+                               Pid: 1,
+                       })
+               }
        }
        return nil
 }
```

However, we need to think more about it. Perhaps we should extend the container selector to container this `host` field. 

Also, this is only available for `ig`, IMO this can also be useful for `kubectl-gadget`, we can argue that kubectl-gadget is only for Kubernetes, but if the tool is already deployed in the cluster and it has the capabilities to trace host event, why not to expose this to the user too? (this can be done in a folow up PR)

Issue #1692, opened 2023-05-31
ig: Doesn't support non-root k8s containers with cri-o 


mauriciovasquezbernal:
`ig` is not able to capture events from non-root k8s containers when using cri-o

```bash 
# build ig (can also use make ig)
$ cd cmd/ig
$ go build . 

# create a minikube cluster with cri-o
$ minikube start --container-runtime cri-o

$ minikube cp ig /ig
$ minikube ssh "sudo chmod +x /ig && sudo /ig trace exec -r cri-o"
CONTAINER                                         PID        PPID       COMM             RET ARGS
```

## working case

Deploy the following pod:

```yaml
# foo.yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
spec:
  securityContext:
  restartPolicy: Never
  terminationGracePeriodSeconds: 0
  containers:
  - name: test-pod-container
    image: busybox
    command: ["/bin/sh", "-c"]
    args:
    - while true; do cat /dev/null ; sleep 0.1; done
```

```bash
$ kubectl apply -f foo.yaml 
pod/test-pod created
```

Events are captured by ig:

```bash
.... 
test-pod-container                                82476      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82477      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82478      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82479      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82480      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82499      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82506      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82507      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82508      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82509      82037      cat              0   /bin/cat /dev/null                                               
test-pod-container                                82510      82037      cat              0   /bin/cat /dev/nul
...
```

```bash
$ kubectl delete -f foo.yaml
```

### not working case

Deploy this foo: (notice the `runAsUser: 1000`)

```yaml
# foo.yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
spec:
  securityContext:
    runAsUser: 1000
  restartPolicy: Never
[...]

PR #1742, opened 2023-06-14
runcfanotify/container-hook: Ignore events from 'passwd' in pidfile directory

mqasimsarfraz:
As mentioned in the comment [here](https://github.com/inspektor-gadget/inspektor-gadget/issues/1692#issuecomment-1591486516) we need to handle `passwd` file in case of non root crio containers. The idea is to ignore events from `passwd` similar to `config.json`.

Fixes: #1692 

## Testing done
Added an integration test: [TestPodWithSecurityContext](https://github.com/inspektor-gadget/inspektor-gadget/blob/69f61a3df1bbecbcad775a708c66ea74b6f4634a/integration/ig/k8s/list_containers_test.go#L247). Also, can be tested manually as:

```bash
$ make CONTAINER_RUNTIME=cri-o minikube-start
$ make ig && docker cp ig minikube-cri-o:/bin/ig
$ docker exec -it minikube-cri-o ig list-containers -w
# start a container with security context in other terminal
$ kubectl run test-pod --image=busybox --overrides='{ "apiVersion": "v1", "spec": { "securityContext": { "runAsUser": 1000 } } }' sleep inf
...
2023-06-16T10:57:59Z           CREATED    cri-o      ff389d1241831c24c25ea845307d039a3aaa218cec98b0d… test-pod
...

### DIFF
--- a/pkg/container-hook/tracer.go
+++ b/pkg/container-hook/tracer.go
@@ -560,6 +560,16 @@ func (n *ContainerNotifier) monitorRuntimeInstance(bundleDir string, pidFile str
 		return fmt.Errorf("marking %s: %w", configJSONPath, err)
 	}
 
+	// similar to config.json, we ignore passwd file if it exists
+	passwdPath := filepath.Join(bundleDir, "passwd")
+	if _, err := os.Stat(passwdPath); !errors.Is(err, os.ErrNotExist) {
+		err = pidFileDirNotify.Mark(unix.FAN_MARK_ADD|unix.FAN_MARK_IGNORED_MASK, unix.FAN_ACCESS_PERM, unix.AT_FDCWD, passwdPath)
+		if err != nil {
+			pidFileDirNotify.File.Close()
+			return fmt.Errorf("marking %s: %w", passwdPath, err)
+		}
+	}
+
 	n.wg.Add(1)
 	go func() {
 		defer n.wg.Done()
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -429,6 +429,16 @@ func (n *RuncNotifier) monitorRuncInstance(bundleDir string, pidFile string) err
 		return fmt.Errorf("ignoring %s: %w", configJSONPath, err)
 	}
 
+	// similar to config.json, we ignore passwd file if it exists
+	passwdPath := filepath.Join(bundleDir, "passwd")
+	if _, err := os.Stat(passwdPath); !errors.Is(err, os.ErrNotExist) {
+		err = pidFileDirNotify.Mark(unix.FAN_MARK_ADD|unix.FAN_MARK_IGNORED_MASK, unix.FAN_ACCESS_PERM, unix.AT_FDCWD, passwdPath)
+		if err != nil {
+			pidFileDirNotify.File.Close()
+			return fmt.Errorf("marking %s: %w", passwdPath, err)
+		}
+	}
+
 	n.wg.Add(1)
 	go func() {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/runcfanotify/runcfanotify.go [hunk 1]
```diff
@@ -413,16 +413,19 @@ func (n *RuncNotifier) monitorRuncInstance(bundleDir string, pidFile string) err
...
-	err = pidFileDirNotify.Mark(unix.FAN_MARK_ADD|unix.FAN_MARK_IGNORED_MASK, unix.FAN_ACCESS_PERM, unix.AT_FDCWD, configJSONPath)
-	if err != nil {
-		pidFileDirNotify.File.Close()
-		return fmt.Errorf("ignoring %s: %w", configJSONPath, err)
+	ignoreFiles := []string{"config.json", "passwd"}
+	for _, ignoreFile := range ignoreFiles {
+		err = pidFileDirNotify.Mark(unix.FAN_MARK_ADD|unix.FAN_MARK_IGNORED_MASK, unix.FAN_ACCESS_PERM, unix.AT_FDCWD, filepath.Join(pidFileDir, ignoreFile))
+		if err != nil {
+			pidFileDirNotify.File.Close()
+			return fmt.Errorf("ignoring %s: %w", ignoreFiles, err)
```
alban, line 427: What if the file `passwd` does not exist? Should we ignore the ENOENT error in that case?

mqasimsarfraz: Good point. Yes, we should. I see the CI failing because of this as well. I will update it.

mqasimsarfraz: Done. I was wondering if we should have a fallback check in `watchPidFileIterate` to handle this e.g `filepath.Base(path) == filepath.Base(pidFileDir)` not sure how robust that is!

pkg/runcfanotify/runcfanotify.go
```diff
@@ -413,16 +413,22 @@ func (n *RuncNotifier) monitorRuncInstance(bundleDir string, pidFile string) err
 
 	// watchPidFileIterate() will read config.json and it might be in the
 	// same directory as the pid file. To avoid getting events unrelated to
-	// the pidfile, add an ignore mask.
+	// the pidfile, add an ignore mask. Same for passwd in case non-root
+	// crio containers
```
burak-ok, line 417: This comment implies that `passwd` is only ignored for non-root crio containers. If I understand the code correct it ignores it for every configuration. So maybe change the wording.

burak-ok (commented):
Looks good. Just a small comment

pkg/container-hook/tracer.go
```diff
@@ -560,6 +560,16 @@ func (n *ContainerNotifier) monitorRuntimeInstance(bundleDir string, pidFile str
 		return fmt.Errorf("marking %s: %w", configJSONPath, err)
 	}
 
+	// similar to config.json, we ignore passwd file if it exists
+	passwdPath := filepath.Join(bundleDir, "passwd")
+	if _, err := os.Stat(passwdPath); err == nil {
```
alban, line 565: With `config.json`, we test the error with `errors.Is(err, os.ErrNotExist)` but with `passwd`, we test the error with `err == nil`. Is there a reason for testing differently? Ditto in runcfanotify.

mqasimsarfraz: Done. `passwd` file is optional compared to `config.json` but makes sense to use `!errors.Is(err, os.ErrNotExist)` for clarity and let it return an `error` in case it happens in following calls.

PR #1749, opened 2023-06-15
Dockerfiles: Set PKG_CONFIG_PATH for arm64.

eiffel-fl:
```
On newer version of golang:1.19 which are based on debian bookworm, the build will fail without this variable as pkg-config will try to find libseccomp for amd64 using the default personality.

By setting this variable, it will search it for the corresponding architecture.
```


mauriciovasquezbernal (approved):
This fixed the compilation issue on my machine.

Issue #1733, opened 2023-06-12
kubectl-gadget/top: Output isn't sorted correctly for multi-node cluster


mqasimsarfraz:
When running top gadget it seems the generated events aren't sorted. For example in following example `-sent` column isn't in expected order for some intervals:

![top](https://github.com/inspektor-gadget/inspektor-gadget/assets/10230890/81634af5-368b-4720-90e3-fe84003e0be9)

```bash
NODE                        NAMESPACE                   POD                         CONTAINER                  PID         COMM          IP REMOTE                LOCAL                 SENT          RECV         
aks-agentpool…87-vmss000002 kube-system                 metrics-serve…49bc669-nz6qc metrics-server             6219        metrics-serv… 6  ::ffff:10.244.0.4:372 ::ffff:10.244.2.4:444 1.157KiB      222B         
aks-agentpool…87-vmss000002 kube-system                 metrics-serve…49bc669-nz6qc metrics-server             6219        metrics-serv… 6  ::ffff:10.244.0.3:412 ::ffff:10.244.2.4:444 1.156KiB      222B         
aks-agentpool…87-vmss000002 kube-system                 metrics-serve…49bc669-nz6qc metrics-server             6219        metrics-serv… 6  ::ffff:10.244.0.4:373 ::ffff:10.244.2.4:444 798B          148B         
aks-agentpool…87-vmss000002 kube-system                 metrics-serve…49bc669-nz6qc metrics-server             6219        metrics-serv… 6  ::ffff:10.244.0.4:373 ::ffff:10.244.2.4:444 448B          113B         
aks-agentpool…87-vmss000005 kube-system                 metrics-serve…49bc669-rvkd6 metrics-server             5915        metrics-serv… 6  ::ffff:10.244.0.4:553 ::ffff:10.244.3.4:444 1.514KiB      296B         
aks-agentpool…87-vmss000005 kube-system                 metrics-serve…49bc669-rvkd6 metrics-server             5915        metrics-serv… 6  ::ffff:10.244.0.4:553 ::ffff:10.244.3.4:444 1.158KiB      292B         
aks-agentpool…87-vmss000005 kube-system                 metrics-serve…49bc669-rvkd6 metrics-server             5915        metrics-serv… 6  ::ffff:10.244.0.3:535 ::ffff:10.244.3.4:444 815B          187B         
[...]

PR #1753, opened 2023-06-16
parser: Sort events after combining for top/snapshot gadgets

mqasimsarfraz:
Currently we don't do any sorting after combining events related to top gadgets from different node. I believe this isn't an expected behavior so I sort events after combining from all the nodes. 

Fixes: #1733 

## Testing done

```bash
# create multi node cluster with ig-k8s
$ MINIKUBE_PARAMS="-n 3" make minikube-deploy
# run the gadget to ensure events are sorted correctly
# if same command is ran w/o the fix you should see output only sorted per node.
$ make kubectl-gadget 
$ ./kubectl-gadget top tcp --all-namespaces --interval 5 --max-rows 20 --sort -sent
```

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -161,6 +161,9 @@ func (p *parser[T]) EnableSnapshots(ctx context.Context, interval time.Duration,
 			select {
 			case <-ticker.C:
 				out, _ := p.snapshotCombiner.GetSnapshots()
+				if p.sortSpec != nil {
+					p.sortSpec.Sort(out)
+				}
 				p.eventCallbackArray(out)
 			case <-ctx.Done():
 				return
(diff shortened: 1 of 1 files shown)


flyth (approved):
Great catch, thanks!

Do we need the same for the `Flush()` function as well?

mqasimsarfraz:
> Do we need the same for the `Flush()` function as well?

You mean the `Flush()` logic we wanted to implement in other PR? If yes then indeed. I was hoping to get this in and then update the other PR.

flyth:
> > Do we need the same for the `Flush()` function as well?
> 
> You mean the `Flush()` logic we wanted to implement in other PR? If yes then indeed. I was hoping to get this in and then update the other PR.

The current version already has a `Flush()` that reads from the `combinedResult` - that should probably also be sorted if a sort spec is present (gadgets of TypeOneShot).

mqasimsarfraz:
> The current version already has a `Flush()` that reads from the `combinedResult` - that should probably also be sorted if a sort spec is present (gadgets of TypeOneShot).

Ah! thanks for catching it. I updated the PR and now snapshot gadgets works fine for multi nodes!

Issue #1757, opened 2023-06-17
[RFE] Add cpuusage in top ebpf gadget


middaywords:
## Current situation

Currently, the `ig top ebpf` has shown cpu overheads like runtime. 

```
$ sudo ./ig top ebpf
NODE             PROGID   TYPE             NAME             PID     COMM                      RUNTIME   RUNCOUNT      MAPMEMORY MAPCOUNT
minikube         506      Kprobe           ig_topfile_rd_e  573222  gadgettracerman         824.589µs       2076       40.95MiB        4
minikube         505      Tracing          ig_top_ebpf_it   573222  gadgettracerman          47.171µs       1103            12B        1
minikube         507      Kprobe           ig_topfile_wr_e  573222  gadgettracerman         609.645µs        836       40.95MiB        4
```

However, for those who want to compare ebpf prog overheads with whole other processes, runtime is not intuitive.
Thus, It would be better if we have a column `CPU` to show sth like what we see in top.

## Ideal future situation

We want to add a column to show the cpu usage. example output like:
```
$ sudo ./ig top ebpf -o columns=progid,type,name,pid,comm,runtime,cpu
PROGID     TYPE                 NAME                 PID                  COMM                              RUNTIME CPU
4035       Tracing              ig_top_ebpf_it       242591               ig                               269.82µs 0.0270%
4026       SchedCLS             cil_from_contai                                                            229.87µs 0.0230%
3988       SchedCLS             cil_from_contai                                                           139.297µs 0.0139%
3941       SchedCLS             cil_to_containe                                                            120.62µs 0.0121%
4020       SchedCLS             cil_to_containe                                                            92.557µs 0.0093%
```

middaywords:
close the issue since related PR has been merged

Issue #1526, opened 2023-04-12
Cleanup custom-columns CLI flag


alban:
## Current situation

We used to have a CLI flag `custom-columns` but it does not exist anymore: it was renamed to `columns`. However, `git grep custom-columns` shows left overs in comments, documentation, tools and code.

`custom-columns` was renamed to `columns` by #1281.

## Impact

Suggested commands from the code base actually don't work anymore:
```
$ ./kubectl-gadget trace dns -n demo --timeout 3 -o custom-columns=pod,comm,qr,qtype,name
Error: invalid output mode "custom-columns"
```

## Ideal future situation

`git grep custom-columns` does not give any results.

## Implementation options

Remove `custom-columns` or rename as appropriate.

## Additional information

```
$ git grep custom-columns
cmd/common/utils/flags.go:30:   OutputModeCustomColumns = "custom-columns"
docs/gadgets/audit/seccomp.md:65:$ kubectl gadget audit seccomp -o custom-columns=namespace,pod,syscall,code
docs/gadgets/common-features.md:50:This can be overridden with either `json` or `custom-columns`.
docs/gadgets/common-features.md:78:Using `-o custom-columns=column1,column2` we can choose which columns to
docs/gadgets/common-features.md:87:$ kubectl gadget trace oomkill -A -o custom-columns=kpid,kcomm
docs/gadgets/profile/cpu.md:89:$ kubectl gadget profile cpu --timeout 1 --podname random -o custom-columns=node,pod
docs/gadgets/profile/cpu.md:100:$ kubectl gadget profile cpu --timeout 1 --podname random -o custom-columns=node,namespace,pod,container,pid,comm,count,stack
docs/gadgets/top/ebpf.md:59:you can call the gadget with the custom-columns option and specify the cumulruntime and cumulruncount columns.
docs/gadgets/top/ebpf.md:64:$ kubectl-gadget top ebpf -o custom-columns=node,progid,type,name,pid,comm,cumulruntime,cumulruncount --sort cumulruntime --timeout 60
docs/gadgets/trace/capabilities.md:165:    -o custom-columns=comm,syscall,capName,verdict,targetuserns,currentuserns,caps,capsnames
docs/ig.md:113:- JSON format and `custom-columns` output mode are supported through the
integration/command.go:209:               NAMESPACES=$(kubectl get seccompprofile --all-namespaces --no-headers --ignore-not-found -o custom-columns=":metadata.namespace" | uniq)
[...]

PR #1758, opened 2023-06-17
pkg/gadgets/top/ebpf: introduce cpu usage

middaywords:
# pkg/gadgets/top/ebpf: introduce cpu usage

This PR introduces cpu usage column in top ebpf gadget. Related issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/1757

For those who want to compare ebpf prog overheads with other processes, runtime is not intuitive. Thus, It would be better if we have a column `CPU` to show sth like what we see in top.

The definition for cpuusage here is similar to that in `top`. It is defined as: The ebpf prog's share of the elapsed CPU time since the last update, expressed as a percentage of total CPU time. Because the ebpf prog's runtime is adding up runtime on each core in kernel, it's possible that the cpuusage here is over 100%.

## How to use

1. build ig: `make ig`
2. run ig ebpf top: `sudo ./ig top ebpf -o columns=progid,type,name,pid,comm,runtime,cpu`
3. check test case: `make test`

## Testing done

1. make test
4. run ig to check the output. (The column is disabled by default.)
```
$ sudo ./ig top ebpf -o columns=progid,type,name,pid,comm,runtime,cpu
PROGID     TYPE                 NAME                 PID                  COMM                              RUNTIME CPU
4035       Tracing              ig_top_ebpf_it       242591               ig                               269.82µs 0.0270%
4026       SchedCLS             cil_from_contai                                                            229.87µs 0.0230%
3988       SchedCLS             cil_from_contai                                                           139.297µs 0.0139%
3941       SchedCLS             cil_to_containe                                                            120.62µs 0.0121%
```

PTAL, feel free to let me know if you have any comments or concerns on this PR.

### DIFF
--- a/docs/gadgets/top/ebpf.md
+++ b/docs/gadgets/top/ebpf.md
@@ -104,3 +104,11 @@ be used by more than one program and would account towards the MapMemory of all 
 Also note:
 * BPF_MAP_TYPE_PERF_EVENT_ARRAY: value_size is not counting the ring buffers, but only their file descriptors (i.e. sizeof(int) = 4 bytes)
 * BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS: value_size is not counting the inner maps, but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+
+### A note about CPU usage
+
+There are two types of cpu usage metrics available in top ebpf gadget:
+* TotalCPUUsage: It means all ebpf progs' share of the elapsed CPU time since the last update, expressed as a percentage of total CPU time. This value can be >100% because it summed up cpu time in all cpu cores, similar to linux `top`.
+* PerCPUUsage: This value is `TotalCPUUsage / CPUCores` and `CPUCores` is the number of cpu cores in a node. PerCPUUsage will be always be <=100%. When using ebpf top gadget to check cpu usage for many nodes, it's suggested to use PerCPUUsage.
+
+To show them, you can run the gadget with `percpu` and `totalcpu` specified in `-o columns` option.
--- a/pkg/gadgets/top/ebpf/tracer/tracer.go
+++ b/pkg/gadgets/top/ebpf/tracer/tracer.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/cilium/ebpf"
+	"github.com/tklauser/numcpus"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/bpfstats"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
@@ -237,6 +238,11 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
 
 	mapSizes := make(map[ebpf.MapID]uint64)
 
+	numOnlineCPUs, err := numcpus.GetOnline()
+	if err != nil {
+		return nil, fmt.Errorf("getting number of online cpu: %w", err)
+	}
+
 	// Get memory usage by maps
 	for {
 		nextMapID, err = ebpf.MapGetNextID(curMapID)
@@ -322,6 +328,8 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
 			runCount: totalRunCount,
 		}
 
+		totalCpuUsage := 100 * float64(curRuntime) / float64(t.config.Interval.Nanoseconds())
+
 		stat := &types.Stats{
 			ProgramID:          uint32(curID),
 			Name:               pi.Name,
@@ -334,6 +342,8 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
 			CumulativeRunCount: cumulativeRunCount,
 			MapMemory:          totalMapMemory,
 			MapCount:           uint32(len(mapIDs)),
+			TotalCpuUsage:      totalCpuUsage,
+			PerCpuUsage:        totalCpuUsage / float64(numOnlineCPUs),
 		}
 
 		if t.enricher != nil {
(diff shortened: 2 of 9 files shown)


pkg/gadgets/top/ebpf/tracer/tracer.go
```diff
@@ -334,6 +334,7 @@ func (t *Tracer) nextStats() ([]*types.Stats, error) {
 			CumulativeRunCount: cumulativeRunCount,
 			MapMemory:          totalMapMemory,
 			MapCount:           uint32(len(mapIDs)),
+			CpuUsage:           100 * float64(curRuntime) / float64(t.config.Interval.Nanoseconds()),
```
mauriciovasquezbernal, line 337: should this be divided by the number of online CPUs to avoid it being > 100% as you mentioned in the commit message?

middaywords: The motivation is to compare the cpu usage of ebpf progs to the cpu usage shown in linux `top` command. So we adopt the same way as what they did in linux `top` command. linux `top` also summed up cpu overheads in all cpu cores, it can be >100% as well.

mauriciovasquezbernal: Make sense.

mauriciovasquezbernal (approved):
LGTM. Thanks a lot for the contribution.

@alban @flyth do you want to take another look since you're more familiar with this gadget?

alban (approved):
Thanks!

pkg/gadgets/top/ebpf/tracer/tracer.go
flyth: Hmm, I'm not too sure about this. I get that this is what top also does, but top only ever works on one node. If you have a cluster with different hardware for the nodes, this would make it even trickier to compare and find bottlenecks, right? Maybe we should at least provide another value where the time is divided by the number of cores?

mauriciovasquezbernal: What could be the name of that second value?

middaywords: > Hmm, I'm not too sure about this. I get that this is what top also does, but top only ever works on one node. If you have a cluster with different hardware for the nodes, this would make it even trickier to compare and find bottlenecks, right? Maybe we should at least provide another value where the time is divided by the number of cores?

Yeah, your concerns make sense. I agree maybe we can provide both two metrics.

> What could be the name of that second value?

maybe `avg_cpu_utilization` or `avg_cpu_util`?

mauriciovasquezbernal: What about having `TotalCpuUsage` and `PerCpuUsage`? (and adding the correct documentation to explain what they are about)

docs/gadgets/top/ebpf.md
```diff
@@ -104,3 +104,11 @@ be used by more than one program and would account towards the MapMemory of all
...
 * BPF_MAP_TYPE_PERF_EVENT_ARRAY: value_size is not counting the ring buffers, but only their file descriptors (i.e. sizeof(int) = 4 bytes)
 * BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS: value_size is not counting the inner maps, but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+
+### A note about CPU usage
+
+There are two types of cpu usage metrics available in top ebpf gadget:
+* TotalCPUUsage: It means all ebpf progs' share of the elapsed CPU time since the last update, expressed as a percentage of total CPU time. This value can be >100% because it summed up cpu time in all cpu cores, similar to linux `top`.
+* PerCPUUsage: This value is `TotalCPUUsage / CPUCores` and `CPUCores` is the number of cpu cores in a node. PerCPUUsage will be always be <=100%. When using ebpf top gadget to check cpu usage for many nodes, it's suggested to use PerCPUUsage.
+
+To show them, you can run the gadget with `per_cpu` and `total_cpu` specified in custom-columns option.
```
mauriciovasquezbernal, line 114:
```suggestion
To show them, you can run the gadget with `per_cpu` and `total_cpu` specified in  `-o columns` option. 
```

(we're getting rid of custom-columns https://github.com/inspektor-gadget/inspektor-gadget/issues/1526)

pkg/gadgets/top/ebpf/types/types.go
```diff
@@ -102,6 +104,12 @@ func GetColumns() *columns.Columns[Stats] {
 	cols.MustSetExtractor("mapmemory", func(stats *Stats) (ret string) {
 		return fmt.Sprint(units.BytesSize(float64(stats.MapMemory)))
 	})
+	cols.MustSetExtractor("total_cpu", func(stats *Stats) (ret string) {
+		return fmt.Sprintf("%.4f%%", stats.TotalCpuUsage)
+	})
+	cols.MustSetExtractor("per_cpu", func(stats *Stats) (ret string) {
+		return fmt.Sprintf("%.4f%%", stats.PerCpuUsage)
+	})
```
mauriciovasquezbernal, line 112: I think we can do the same by using `precision:4` in the column tag.

pkg/utils/host/host.go
```diff
@@ -62,6 +65,27 @@ func init() {
...
+
+		if key == "CPU(s)" {
+			HostNumCPU, err = strconv.Atoi(value)
+			if err != nil {
+				HostNumCPU = 0
+			}
+			break
+		}
+	}
 }
```
mauriciovasquezbernal, line 89: Some comments:
- I don't think parsing the output of `lscpu` is the right approach. That command couldn't be available in some environments, so it's better to get this information directly from golang.
- We need to take the "online cpus" number. This can change at runtime, so we'll need to calculate it each time instead of caching it. 

We can use https://pkg.go.dev/github.com/tklauser/numcpus#GetOnline instead of having to parse any file manually.

mauriciovasquezbernal (changes requested):
Thanks a lot for the update. Few comments regarding the logic around getting CPUs.

pkg/gadgets/top/ebpf/types/types.go
mauriciovasquezbernal: ops, as noted by @flyth, the extractor is actually needed in this case to print the `%` sign. Sorry for the noise.

mauriciovasquezbernal (approved):
I pushed some small changes directly to your fork.

This LGTM, thanks a lot for your contribution. I'll merge once tests are green.

flyth (approved):
Great work, thanks a lot for the contribution!

mauriciovasquezbernal:
CI failure is not related to this PR and we're aware of them. Merging.

PR #1759, opened 2023-06-18
socketenricher: add support for older kernels

alban:
# socketenricher: add support for older kernels

The socket enricher currently fails on Linux 5.4.0:

> creating socket enricher: loading ebpf program: field IgFreeIpv4E:
> program ig_free_ipv4_e: load program: invalid argument: invalid func
> unknown#125 (133 line(s) omitted)

func 125 is bpf_ktime_get_boot_ns:
```
$ sudo bpftool btf dump id 1|grep BPF_FUNC_.*val=125
	'BPF_FUNC_ktime_get_boot_ns' val=125
```

IG has a workaround for bpf_ktime_get_boot_ns. Let's use it in the socket enricher.

## How to use

```
$ kubectl gadget trace dns
^C
$ kubectl logs -n gadget -l k8s-app=gadget
...
level=error msg="creating socket enricher: loading ebpf program: field IgFreeIpv4E: program ig_free_ipv4_e: load program: invalid argument: invalid func unknown#125 (133 line(s) omitted)"
```

## Testing done

It works on Linux 5.4.0 but in degraded mode (without initial sockets, and without TCP-FIN packets):

```
$ kubectl get node -o wide
NAME           STATUS   ROLES           AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION      CONTAINER-RUNTIME
controlplane   Ready    control-plane   38d   v1.27.1   172.30.1.2    <none>        Ubuntu 20.04.5 LTS   5.4.0-131-generic   containerd://1.6.12

$ ./kubectl-gadget trace network
NODE                   NAMESPACE              POD                   PID        TID        COMM        TYPE      PROTO PORT  REMOTE               
controlplane           default                sh01                  77350      77350      wget        OUTGOING  tcp   8086  r/127.0.0.1          
```


alban:
> It looks good, but what are the consequences of having time set to 0?

You're right, this PR is not correct: if the `deletion_timestamp` is set to zero, the sockets will never be cleaned up from the map. I need to use bpf_ktime_get_ns() or another method.

alban:
I updated the patch to use bpf_ktime_get_ns (Linux 4.1) instead of bpf_ktime_get_boot_ns (Linux 5.8). However, this is still using a bpf iterator (Linux 5.8) to initialize the sockets map and clean up the deleted sockets from the map. So it does not help.

We could implement a slower initialization and cleanup by iterating from userspace, but I don't know if it is worth it.

alban:
For the record, with this patch, I get the following error on Ubuntu 20.04.5 LTS Linux 5.4.0 (as expected):

> creating socket enricher: loading ebpf program: field IgSkCleanup: program ig_sk_cleanup: attach Tracing/TraceIter: bpf_map_elem iterator not supported

alban:
> > For the record, with this patch, I get the following error on Ubuntu 20.04.5 LTS Linux 5.4.0 (as expected):
> > > creating socket enricher: loading ebpf program: field IgSkCleanup: program ig_sk_cleanup: attach Tracing/TraceIter: bpf_map_elem iterator not supported
> 
> Then, should we simply try to load the program and if this particular error is fired, inform the user that this gadget relies on this?

Like a warning message visible with `kubectl-gadget -v`?

"Warning: this gadgets uses the socket enricher, only available from Linux 5.8. Some events might have missing details."

That sounds good to me.

Another thing: at the moment, the socket enricher is either enabled fully or completely disabled.  I think we should also enable some features of the socket enricher on older kernels: only bpf iterators require Linux 5.8 and they are only used for initializing existing sockets and postponing deletion of sockets after a timeout. When bpf iterators cannot be used, we should still enable the basic features, even if some events cannot be enriched.

alban:
> I am nonetheless a bit worried regarding the memory footprint of not deleting socket periodically. Can you please provide context regarding this?

If iterators are not supported, the socket will be deleted immediately instead of waiting for some seconds. So it will actually use less memory.

See [ig_sk_cleanup() L73](https://github.com/inspektor-gadget/inspektor-gadget/pull/1759/files#diff-5638853504b028e47a75a911cb465f120f5d7271ddbe37ea1ef607bdd264af06R73):

```
if (disable_bpf_iterators) {
    bpf_map_delete_elem(&sockets, &socket_key);
```

pkg/gadgets/internal/socketenricher/bpf/sockets-iter.bpf.c
```diff
@@ -0,0 +1,126 @@
...
+	// connection is terminated. We cannot determine if an active socket
+	// is server side or client side, so we add active socket anyway on the
+	// chance that it is client side. It will be fine for server side too,
+	// because the passive socket will be added later, overwriting the
+	// active socket.
+	u64 flags = BPF_NOEXIST;
+	if (BPF_CORE_READ(sock, __sk_common.skc_state) == TCP_LISTEN)
+		flags = BPF_ANY;
+
+	bpf_map_update_elem(&sockets, &socket_key, &socket_value, flags);
```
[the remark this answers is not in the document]
alban: I think your code is easier. There might have been a difference that your code does not handle the race condition if `probe_release_entry` is called in parallel for removing another connection to the same listening socket.

But now that we delete the sockets after a timeout from tracer.go in the go routine `go se.cleanupDeletedSockets(cleanupIter)` and that go routine is only started after the initialization of initial sockets, the race condition can't happen anymore. So I will try your version.

pkg/gadgets/internal/socketenricher/tracer.go
```diff
@@ -58,14 +62,36 @@ func NewSocketEnricher() (*SocketEnricher, error) {
...
 
 func (se *SocketEnricher) start() error {
+	specIter, err := loadSocketsiter()
+	if err != nil {
+		return fmt.Errorf("loading socketsiter asset: %w", err)
+	}
+
+	kallsyms.SpecUpdateAddresses(specIter, []string{"socket_file_ops"})
+
+	disable_bpf_iterators := false
```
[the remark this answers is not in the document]
alban: Yes. I'll update.

pkg/gadgets/internal/socketenricher/tracer.go
```diff
@@ -58,14 +62,36 @@ func NewSocketEnricher() (*SocketEnricher, error) {
...
+	if err != nil {
+		return fmt.Errorf("loading socketsiter asset: %w", err)
+	}
+
+	kallsyms.SpecUpdateAddresses(specIter, []string{"socket_file_ops"})
+
+	disableBPFIterators := false
+	if err := specIter.LoadAndAssign(&se.objsIter, nil); err != nil {
+		disableBPFIterators = true
+		log.Debugf("Socket enricher: skip loading iterators: %v", err)
```
mauriciovasquezbernal, line 75: According to [this](https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels) SO answer:

"Warn - Anything that can potentially cause application oddities, but for which I am automatically recovering. (Such as switching from a primary to backup server, retrying an operation, missing secondary data, etc.)"

I think it perfectly matches this case, so this (and other similar ones in this commit) should be switched to warn.

pkg/gadgets/internal/socketenricher/bpf/socket-enricher.bpf.c
```diff
@@ -25,17 +26,7 @@ struct {
...
-static __always_inline void
-prepare_socket_key(struct sockets_key *socket_key, struct sock *sock)
-{
-	struct inet_sock *inet_sock = (struct inet_sock *)sock;
-	BPF_CORE_READ_INTO(&socket_key->netns, sock, __sk_common.skc_net.net, ns.inum);
-	BPF_CORE_READ_INTO(&socket_key->family, sock, __sk_common.skc_family);
-	socket_key->proto = BPF_CORE_READ_BITFIELD_PROBED(sock, sk_protocol);
-	socket_key->port = bpf_ntohs(BPF_CORE_READ(inet_sock, inet_sport));
-}
+const volatile __u64 disable_bpf_iterators = 0;
```
mauriciovasquezbernal, line 29: Can it be a boolean?

mauriciovasquezbernal (approved):
Some minor comments but LGTM from code inspection.

pkg/gadgets/internal/socketenricher/bpf/socket-enricher.bpf.c
alban: I am not sure: `bool` is not a built-in type, and userspace programs are supposed to `#include <stdbool.h>` to make use of it. In eBPF, we have vmlinux.h defining `bool`:
```
typedef _Bool bool;
```
`_Bool` is defined in C99.

I have not tested if cilium/ebpf's `RewriteConstants()` works with this type.
I see cilium/ebpf's `marshalBytes()` uses `binary.Write()`, which according to its documentation:
> Boolean values encode as one byte: 1 for true, and 0 for false.

Should I try it?

mauriciovasquezbernal: It should work fine. Other gadgets are using bool too:

```bash
$ git grep "const volatile bool"
pkg/gadgets/common/mntns_filter.h:const volatile bool gadget_filter_by_mntns = false;
pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c:const volatile bool filter_cg = false;
pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c:const volatile bool targ_per_disk = false;
pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c:const volatile bool targ_per_flag = false;
pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c:const volatile bool targ_queued = false;
pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c:const volatile bool targ_ms = false;
pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c:const volatile bool filter_dev = false;
pkg/gadgets/profile/cpu/tracer/bpf/profile.bpf.c:const volatile bool kernel_stacks_only = false;
pkg/gadgets/profile/cpu/tracer/bpf/profile.bpf.c:const volatile bool user_stacks_only = false;
pkg/gadgets/profile/cpu/tracer/bpf/profile.bpf.c:const volatile bool include_idle = false;
pkg/gadgets/profile/tcprtt/tracer/bpf/tcprtt.bpf.c:const volatile bool targ_laddr_hist = false;
pkg/gadgets/profile/tcprtt/tracer/bpf/tcprtt.bpf.c:const volatile bool targ_raddr_hist = false;
pkg/gadgets/profile/tcprtt/tracer/bpf/tcprtt.bpf.c:const volatile bool targ_ms = false;
pkg/gadgets/snapshot/process/tracer/bpf/process-collector.bpf.c:const volatile bool show_threads = false;
pkg/gadgets/top/file/tracer/bpf/filetop.bpf.c:const volatile bool regular_file_only = true;
pkg/gadgets/trace/bind/tracer/bpf/bindsnoop.bpf.c:const volatile bool ignore_errors = true;
pkg/gadgets/trace/bind/tracer/bpf/bindsnoop.bpf.c:const volatile bool filter_by_port = false;
pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c:const volatile bool audit_only = false;
pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c:const volatile bool unique = false;
pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c:const volatile bool ignore_failed = true;
pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c:const volatile bool targ_failed = false;
pkg/gadgets/trace/signal/tracer/bpf/sigsnoop.bpf.c:const volatile bool failed_only = false;
pkg/gadgets/trace/tcpconnect/tracer/bpf/tcpconnect.bpf.c:const volatile bool do_count = 0;
pkg/gadgets/trace/tcpconnect/tracer/bpf/tcpconnect.bpf.c:const volatile bool calculate_latency = false;
```

alban:
Thanks! I will merge when the tests are completed.

PR #1762, opened 2023-06-19
docs: tcpdrop: listing similar tools

alban:
Listing tools similar to the tcpdrop gadget.

### DIFF
--- a/docs/gadgets/trace/tcpdrop.md
+++ b/docs/gadgets/trace/tcpdrop.md
@@ -203,3 +203,11 @@ This table can be generated with:
 ```bash
 $ go run ./pkg/gadgets/trace/tcpdrop/tracer/dropreasongen/...
 ```
+
+### Other tools showing dropped packets
+
+The following tools can be used to show dropped packets but they are not focused on containers or Kubernetes:
+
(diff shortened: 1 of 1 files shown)


mqasimsarfraz (approved):
LGTM!

docs/gadgets/trace/tcpdrop.md
```diff
@@ -202,4 +202,12 @@ This table can be generated with:
...
\ No newline at end of file
+```
+
+### Other tools showing dropped packets
+
+The following tools can be used to show dropped packets but they are not focused on containers or Kubernetes:
+
+* [dropwatch](https://github.com/nhorman/dropwatch): interactive tool using [Netlink Devlink Trap](https://www.kernel.org/doc/html/latest/networking/devlink/devlink-trap.html) to see drops packets by the NIC.
+* [BCC's tcpdrop](https://github.com/iovisor/bcc/blob/master/tools/tcpdrop_example.txt): tool using eBPF and kprobes/tracepoints to show when a socket buffer is released by the kernel.
+* [Retis' skb-drop collector](https://github.com/retis-org/retis): tool using various collectors (eBPF, skb-drop, nftables, ovs) to show the flow of packets in the kernel.
```
alban, line 213: Hi @amorenoz, is this an accurate one line summary for Retis in comparison with others?

amorenoz: Hi @alban , thanks for including retis!
Just a small nit: eBPF is not a collector, it's the core technology we use for all collectors :-D. Otherwise lgtm.

PR #1764, opened 2023-06-19
Optimize `LookupContainerByMntNs()` using a map

eiffel-fl:
Hi.

This PR modifies how `LookupContainerByMntNs()` works by using a map rather than ranging through all containers.
It increases a bit the memory footprint by time is drastically reduced.

Best regards.


flyth:
This seems similar to #1616 - I wanted to add the benchmarks @mauriciovasquezbernal requested prior to merging, but forgot somehow to do it :/

PR #1765, opened 2023-06-19
container-hook: don't search fanotify_fops in kallsyms

alban:
`fanotify_fops` is a static constant. It is not exposed in `/proc/kallsyms` on Linux `5.10.102.1-microsoft-standard-WSL2`.

Instead, use `fanotify_show_fdinfo`, which is not static.

Follow-up to https://github.com/inspektor-gadget/inspektor-gadget/pull/1709

---

On WSL2:
```
/ # uname -r
5.10.102.1-microsoft-standard-WSL2
/ # grep -wE 'fanotify_fops|fanotify_show_fdinfo' /proc/kallsyms
ffffffff8131dd10 T fanotify_show_fdinfo
/ #
```

On Fedora:
```
$ uname -r
6.2.10-100.fc36.x86_64
$ sudo grep -wE 'fanotify_fops|fanotify_show_fdinfo' /proc/kallsyms 
ffffffffac49d7c0 T fanotify_show_fdinfo
ffffffffad284c00 d fanotify_fops
$
```


mauriciovasquezbernal (approved):
LGTM from code inspection. Do you know why it's exposed on some versions only? Does it only depends on the kernel version?

alban:
I tried to find out but I didn't see any difference in the code between the two versions.

PR #1766, opened 2023-06-20
Add IP as param type

eiffel-fl:
Hi.

This PR enable usng `net.IP` as `params`.

Best regards.


mauriciovasquezbernal (commented):
Code LGTM. But what about using this on some gadgtets, like profile tcprtt?

mauriciovasquezbernal (approved):
LGTM!

PR #1773, opened 2023-06-22
deploy: Add 'seLinuxOptions' for gadget container

mqasimsarfraz:
It seems the pods with hostNetwork/hostPid [1] aren't confined by SELinux. But recently we disabled both hostNetwork/hostPid for gadget pod hence we need to set appropiate SELinux context allowing us to mount host paths with correct persmission. This change also introduces a flag 'skip-selinux-opts' to allow user to skip this option if needed. Please see the [comments](https://github.com/inspektor-gadget/inspektor-gadget/pull/1709#issuecomment-1599080208) for details.

1 - https://github.com/cri-o/cri-o/blob/v1.27.0/server/sandbox_run_linux.go#L537

### DIFF
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -82,6 +82,7 @@ var (
 	runtimesConfig      commonutils.RuntimesSocketPathConfig
 	nodeSelector        string
 	experimentalVar     bool
+	skipSELinuxOpts     bool
 )
 
 var supportedHooks = []string{"auto", "crio", "podinformer", "nri", "fanotify", "fanotify+ebpf"}
@@ -155,6 +156,11 @@ func init() {
 		"experimental",
 		false,
 		"enable experimental features")
+	deployCmd.PersistentFlags().BoolVarP(
+		&skipSELinuxOpts,
+		"skip-selinux-opts", "",
+		false,
+		"skip setting SELinux options on the gadget pod")
 	rootCmd.AddCommand(deployCmd)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/resources/manifests/deploy.yaml
```diff
@@ -176,6 +176,10 @@ spec:
           - name: HOST_ROOT
             value: "/host"
         securityContext:
+          # We need this to be able mount host directories with correct permissions on OpenShift
+          # The option is ignored for non SELinux systems.
```
alban, line 180: Maybe you can add the cri-o link in a comment somewhere:
https://github.com/cri-o/cri-o/blob/v1.27.0/server/sandbox_run_linux.go#L537

mqasimsarfraz: Done. I updated the comment to include more context!

cmd/kubectl-gadget/deploy.go
```diff
@@ -445,6 +451,11 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 				daemonSet.Spec.Template.Spec.Affinity = affinity
 			}
 
+			// skip SELinux options if the user explicitly requests it for
```
alban, line 454:
```suggestion
			// skip SELinux options if the user explicitly requests it
```

pkg/resources/manifests/deploy.yaml
```diff
@@ -176,6 +176,14 @@ spec:
           - name: HOST_ROOT
             value: "/host"
         securityContext:
+          # With hostPID/hostNetwork/privliged [1] set to false, we need to set appropriate
```
alban, line 179:
```suggestion
          # With hostPID/hostNetwork/privileged [1] set to false, we need to set appropriate
```

alban (approved):
LGTM if the CI is green, after fixing the typos.
Thanks!

PR #1778, opened 2023-06-23
container-utils: Add TestNewContainerRuntimeClient

mqasimsarfraz:
We should be able to initialze runtimeclient.ContainerRuntimeClient even if the underlying runtime isn't available e.g socketPath not found etc. I added a test to ensure we can track this intention.

### DIFF
--- a/pkg/container-utils/containerutils_test.go
+++ b/pkg/container-utils/containerutils_test.go
@@ -18,7 +18,22 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
+
+func TestNewContainerRuntimeClient(t *testing.T) {
+	nonExistingSocketPath := filepath.Join(t.TempDir(), "non-existing-socket")
+	for _, runtime := range AvailableRuntimes {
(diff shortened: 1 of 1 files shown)


burak-ok (approved):
I think this is all that we need currently. Thanks

pkg/container-utils/containerutils_test.go
```diff
@@ -20,6 +20,23 @@ import (
...
+			Name:       runtime,
+			SocketPath: nonExistingSocketPath,
+		}
+		c, err := NewContainerRuntimeClient(&rc)
+		if err != nil {
+			t.Errorf("Cannot create client for %s: %s", runtime, err)
+		}
+		if c == nil {
+			t.Errorf("Client for %s is nil", runtime)
+		}
```
blanquicet, line 36: Please is "github.com/stretchr/testify" library:

```suggestion
		require.Nil(t, err)
		require.NotNil(t, c)		
```

blanquicet (approved):
LGTM!

mqasimsarfraz:
Thanks for the reviews!

PR #1779, opened 2023-06-23
container-collection: Keep net ns inode id reference

mauriciovasquezbernal:
We're hitting, again, the issue where the events are enriched with the wrong container data, this time in networking gadgets. This is happenning because in some situations the container collection can keep the reference to a container that is not present anymore in the kernel (yes, I know containers aren't a kernel concept but you get the point). In this case, the kernel can reuse the net ns inode in a new container, making the enrichment process to use the wrong container.

The solution proposed here is the same as 4dedafc8565d ("Fix race condition on enrichers") but this time for the net ns.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1001

TODO
- [ ] Implement a test like `TestEventEnrichmentRaceCondition` for this case.

### DIFF
--- a/pkg/container-collection/container-collection.go
+++ b/pkg/container-collection/container-collection.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"golang.org/x/sys/unix"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
@@ -108,6 +107,7 @@ initialContainersLoop:
 			ok := enricher(container)
 			if !ok {
 				// Enrichers can decide to drop a container
+				container.close()
 				continue initialContainersLoop
 			}
 		}
@@ -168,6 +168,7 @@ func (cc *ContainerCollection) AddContainer(container *Container) {
 		ok := enricher(container)
 		// Enrichers can decide to drop a container
 		if !ok {
+			container.close()
 			return
 		}
 	}
@@ -217,11 +218,8 @@ func (cc *ContainerCollection) LookupContainerByMntns(mntnsid uint64) *Container
 	return lookupContainerByMntns(&cc.containers, mntnsid)
 }
 
-// LookupContainersByNetns returns a slice of containers that run in a given
-// network namespace. Or an empty slice if there are no containers running in
-// that network namespace.
-func (cc *ContainerCollection) LookupContainersByNetns(netnsid uint64) (containers []*Container) {
-	cc.containers.Range(func(key, value interface{}) bool {
+func lookupContainersByNetns(m *sync.Map, netnsid uint64) (containers []*Container) {
+	m.Range(func(key, value interface{}) bool {
 		c := value.(*Container)
 		if c.Netns == netnsid {
 			containers = append(containers, c)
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal:
Regarding the TODO. This test is really difficult to implement in this case , `TestEventEnrichmentRaceCondition` [uses trace-exec,](https://github.com/inspektor-gadget/inspektor-gadget/blob/1c51eb93cab637ddf50755f7bec79e629f93d091/pkg/gadgets/test/gadgets_test.go#L88) it [doesn't depend on an operator to enrich the data](https://github.com/inspektor-gadget/inspektor-gadget/blob/1c51eb93cab637ddf50755f7bec79e629f93d091/pkg/gadgets/trace/exec/tracer/tracer.go#L169). On the other hand, trace sni fully relies on an operator to enrich the data, that makes this test more difficult to implement. Since this PR is fixing an issue, I'll propose to move ahead and implement that test later on.

pkg/container-collection/containers.go
```diff
@@ -73,6 +73,7 @@ type Container struct {
 	// events from it.
 	// This is only used when cachedContainers are enabled through WithTracerCollection().
 	mntNsFd int
+	netNsFd int
```
alban, line 76: The comment needs to be updated

pkg/container-collection/options.go
```diff
@@ -796,15 +796,23 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
...
 			}
+			container.mntNsFd = mntNsFd
 
-			container.mntNsFd = fd
+			netNsPath := filepath.Join(host.HostProcFs, fmt.Sprint(container.Pid), "ns", "net")
+			netNsFd, err := unix.Open(netNsPath, unix.O_RDONLY|unix.O_CLOEXEC, 0)
+			if err != nil {
+				log.Warnf("WithTracerCollection: failed to open netns reference for container %s: %s",
+					container.ID, err)
+				return false
```
alban, line 813: Shouldn't we close `mntNsFd` here?

alban (commented):
Thanks!

pkg/container-collection/container-collection.go
```diff
@@ -452,6 +459,10 @@ func (cc *ContainerCollection) Close() {
 			unix.Close(c.mntNsFd)
 			c.mntNsFd = 0
 		}
+		if c.netNsFd != 0 {
+			unix.Close(c.netNsFd)
+			c.netNsFd = 0
```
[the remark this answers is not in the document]
mauriciovasquezbernal: And around line 822 too. It doesn't scale well. I introduced a `close()` method on container to factorize this in a single place.

pkg/container-collection/options.go
```diff
@@ -796,26 +793,32 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
...
+			container.mntNsFd = mntNsFd
 
-			container.mntNsFd = fd
+			netNsPath := filepath.Join(host.HostProcFs, fmt.Sprint(container.Pid), "ns", "net")
+			netNsFd, err := unix.Open(netNsPath, unix.O_RDONLY|unix.O_CLOEXEC, 0)
+			if err != nil {
+				log.Warnf("WithTracerCollection: failed to open netns reference for container %s: %s",
+					container.ID, err)
+				unix.Close(mntNsFd)
+				return false
```
alban, line 811: What about removing unix.Close() from here and add the new container.close() function in two places:
* https://github.com/inspektor-gadget/inspektor-gadget/blob/2a2630c73bac7c7cd51727785dc61167d9bb4406/pkg/container-collection/container-collection.go#L108-L110
* https://github.com/inspektor-gadget/inspektor-gadget/blob/2a2630c73bac7c7cd51727785dc61167d9bb4406/pkg/container-collection/container-collection.go#L168-L170

```
// Enrichers can decide to drop a container
```

alban (approved):
LGTM after the previous comment is addressed (and if tests pass). Thanks!

PR #1780, opened 2023-06-23
podinformer: Fix removing non-running containers

mauriciovasquezbernal:
This commit fixes few issues with the logic to remove non-runinng
containers:
1. Remove container that weren't added by the pod informer: it can happen
   that a container is added by a different hook but such hooks fails to
   remove it. In this case the podinformer can act as a backup mechanism
   to remove such container.
   This cause the enrichment to fail in some cases as the cotainer
   collection keep a container with a mount or net ns reference that
   doesn't exist anymore, then it can be reused by a new container making
   the enrichment to enrich with the wrong data.
2. Remove runtime prefix from id: The container's ID in the container
   collection is stored without a container runtime prefix. Remove this
   before calling cc.RemoveContainer().
3. Remove elements from containerIDsByKey map when pod is removed.

---

This PR is a mitigation for https://github.com/inspektor-gadget/inspektor-gadget/issues/1001 (I wasn't able to reproduce the issue anymore with this), but the real solution is in https://github.com/inspektor-gadget/inspektor-gadget/pull/1779.

---

## Testing

0. Deploy Inspektor Gadget in a cluster running cri-o:

```bash
$ minikube start --container-runtime=cri-o -n 1
$ ./kubectl-gadget deploy
```

The following commands assume the cluster only has a single node, update
them in case it isn't true.

### Before this commit

1. Create pod that finishes and it's restarted

```bash
$ kubectl run --restart=Never --image busybox mypod -- sh -c "sleep 5 && echo foo"
```

2. Container is present for a while

```bash
$ PODNAME=$(kubectl -n gadget get pod -o jsonpath='{.items[*].metadata.name}' | awk '{print $1;}')
$ kubectl -n gadget exec -it $PODNAME -- bash -c "gadgettracermanager -dump containers | grep mypod | wc -l"
1
```

3. After the pod is done, the container is still present:
```bash
$ kubectl -n gadget exec -it $PODNAME -- bash -c "gadgettracermanager -dump containers | grep mypod | wc -l"
1
```

The container is marked as `Exited` by cri-o:

```bash
$ minikube ssh "sudo crictl ps -a | grep mypod"
de22f2a3d8414       docker.io/library/busybox@sha256:1b0a26bd07a3d17473d8d8468bea84015e27f87124b283b91d781bce13f61370                45 seconds ago      Exited              mypod                     0                   0492288eb0fcf
```

After some minutes the container is removed from cri-o and then
from Inspektor Gadget too (by using the post-stop hook):

```bash
$ minikube ssh "sudo crictl ps -a | grep mypod"
ssh: Process exited with status 1
$ kubectl -n gadget exec -it $PODNAME -- bash -c "gadgettracermanager -dump containers | grep mypod | wc -l"
0
```

### After this commit

1. Create pod that finishes and it's restarted

```bash
$ kubectl run --restart=Never --image busybox mypod -- sh -c "sleep 5 && echo foo"
```

2. Container is present for a while

```bash
$ PODNAME=$(kubectl -n gadget get pod -o jsonpath='{.items[*].metadata.name}' | awk '{print $1;}')
$ kubectl -n gadget exec -it $PODNAME -- bash -c "gadgettracermanager -dump containers | grep mypod | wc -l"
1
```

3. After the pod is done, the container is removed:
```bash
$ kubectl -n gadget exec -it $PODNAME -- bash -c "gadgettracermanager -dump containers | grep mypod | wc -l"
0
```

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -247,15 +247,16 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 
 			for {
 				select {
-				case d, ok := <-podInformer.DeletedChan():
+				case key, ok := <-podInformer.DeletedChan():
 					if !ok {
 						return
 					}
-					if containerIDs, ok := containerIDsByKey[d]; ok {
+					if containerIDs, ok := containerIDsByKey[key]; ok {
 						for containerID := range containerIDs {
 							cc.RemoveContainer(containerID)
 						}
 					}
+					delete(containerIDsByKey, key)
 				case c, ok := <-podInformer.CreatedChan():
 					if !ok {
 						return
@@ -274,7 +275,6 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 						if _, ok := containerIDs[id]; !ok {
 							continue
 						}
-
 						cc.RemoveContainer(id)
 					}
 
@@ -286,6 +286,7 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 						if _, ok := containerIDs[container.ID]; ok {
 							continue
 						}
+						containerIDs[container.ID] = struct{}{}
 
 						// Make a copy instead of passing the same pointer at
 						// each iteration of the loop
@@ -299,8 +300,6 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 								container.Namespace, container.Podname, container.Name)
 						}
 						cc.AddContainer(&newContainer)
-
-						containerIDs[container.ID] = struct{}{}
 					}
 				}
 			}
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -257,11 +257,11 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
 						}
 					}
 					delete(containerIDsByKey, key)
-				case c, ok := <-podInformer.CreatedChan():
+				case pod, ok := <-podInformer.UpdatedChan():
 					if !ok {
 						return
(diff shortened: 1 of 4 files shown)


pkg/container-collection/options.go
```diff
@@ -270,11 +271,10 @@ func withPodInformer(nodeName string, fallbackMode bool) ContainerCollectionOpti
...
-						// container had not been added, no need to remove it
-						if _, ok := containerIDs[id]; !ok {
-							continue
-						}
-
+						// Try to remove the container even if it wasn't
+						// captured by the pod informer. It could happen
+						// that it's captured by a start-container hook but
+						// the stop-container one is having issues.
 						cc.RemoveContainer(id)
```
alban, line 278: Could it happen that `k8sClient.GetNonRunningContainers` returns a container id that has not been started yet? Maybe the container has been started but the API Server or the cache does not know that it is already started, so we would wrongly remove the container immediately after its startup if the API server is a bit slow to update the `Running->StartedAt` field?

mauriciovasquezbernal: I think that could happen. I changed the logic a bit, now the container is only removed if this was previously detected by the pod informer, in this way I think we avoid the situation you described.

mauriciovasquezbernal:
Updated the PR. I also added a commit renaming some variables to make this code more clear.

alban (approved):
The second commitmsg needs to be updated: the change in point 1 was removed.

Otherwise, LGTM. Thanks!

mauriciovasquezbernal:
> The second commitmsg needs to be updated: the change in point 1 was removed.

Actually that's still true. I moved the `containerIDs[container.ID] = struct{}{}` line up, so it's executed even if the code continues around L297 when in fallback mode.

PR #1786, opened 2023-06-26
pkg/operators: Enable tracing hosts events for attacher gadgets.

eiffel-fl:
Hi.

This PR follow ups #1741.
Can you please provide more feedback because I am not able to make it work with:

```bash
$ sudo ./ig trace dns --host                                                             francis/host-events-attacher *%
CONTAINER                               PID        TID        COMM             QR TYPE      QTYPE      NAME                                   RCODE                NUMANSW…
# Other terminal
$ $ dig inspektor-gadget.io                                                                francis/host-events-attacher *%

; <<>> DiG 9.18.12-0ubuntu0.22.04.2-Ubuntu <<>> inspektor-gadget.io
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 53613
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;inspektor-gadget.io.           IN      A

;; ANSWER SECTION:
inspektor-gadget.io.    168     IN      A       188.114.96.2
inspektor-gadget.io.    168     IN      A       188.114.97.2

;; Query time: 0 msec
;; SERVER: 127.0.0.53#53(127.0.0.53) (UDP)
;; WHEN: Mon Jun 26 17:23:18 CEST 2023
;; MSG SIZE  rcvd: 80
# Go back to first terminal but nothing was printed.
```

Best regards.

### DIFF
--- a/integration/ig/k8s/trace_dns_test.go
+++ b/integration/ig/k8s/trace_dns_test.go
@@ -21,6 +21,7 @@ import (
 
 	. "github.com/inspektor-gadget/inspektor-gadget/integration"
 	dnsTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/dns/types"
+	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
 func TestTraceDns(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/operators/localmanager/localmanager.go
```diff
@@ -314,6 +314,12 @@ func (l *localManagerTrace) PreGadgetRun() error {
 			},
 		)
 
+		if host {
+			// We need to attach this fake container for gadget which rely only on the
+			// Attacher concept.
+			containers := append(containers, &containercollection.Container{Pid: 1,})
```
alban, line 320:
```suggestion
			containers = append(containers, &containercollection.Container{Pid: 1,})
```

mauriciovasquezbernal: After this change it works on my system. 


```bash 
$ go run -exec sudo ./cmd/ig/. trace dns --host 
CONTAINER                         PID        TID        COMM             QR TYPE      QTYPE      NAME                              RCODE              NUMANSW…
                                  309974     309975     isc-worker0000   Q  OUTGOING  A          inspektor-gadget.io.                                 0       
                                  1126       1126       systemd-resolve  Q  HOST      A          inspektor-gadget.io.                                 0       
                                  1126       1126       systemd-resolve  Q  OUTGOING  A          inspektor-gadget.io.                                 0       
                                  1126       1126       systemd-resolve  R  HOST      A          inspektor-gadget.io.              NoError            2       
                                  1126       1126       systemd-resolve  R  OUTGOING  A          inspektor-gadget.io.              NoError            2       
                                  309974     309975     isc-worker0000   R  HOST      A          inspektor-gadget.io.              NoError            2   
```

integration/ig/k8s/trace_dns_test.go
```diff
@@ -164,3 +165,95 @@ func TestTraceDns(t *testing.T) {
...
+		Name:         "TraceDnsHost",
+		Cmd:          "ig trace dns -o json --host",
+		StartAndStop: true,
+		ExpectedOutputFn: func(output string) error {
+			expectedEntries := []*dnsTypes.Event{
+				{
+					Event: eventtypes.Event{
+						Type: eventtypes.NORMAL,
+					},
+					Comm:       "isc-worker0000",
```
mauriciovasquezbernal, line 195: Why do we need to use `isc-worker0000` and why isn't nslookup reported there? Is this a bug on the enrichment?

alban: nslookup has several threads and one of them is called `isc-worker0000`:
```
$ cat /proc/$(pidof nslookup)/task/*/comm
nslookup
isc-net-0000
isc-timer
isc-socket-0
```

But it does dependent on the nslookup binary. If you use the busybox nslookup, it is not multithreaded. If you use the host nslookup, depending on the Linux distribution, it might be.

The socket enricher currently looks at the `comm` of the thread and not the process because it uses `bpf_get_current_comm`. Except for initial sockets, where we cannot distinguish the threads that created the socket: in that case we take the `comm` from the main thread.

mauriciovasquezbernal: Thanks a lot for the explanation. It'd be useful for include this on a comment on the test.

mauriciovasquezbernal (approved):
LGTM after handling the comment about isc-worker...

integration/ig/k8s/trace_dns_test.go
```diff
@@ -225,6 +222,10 @@ func TestTraceDnsHost(t *testing.T) {
 
 			normalize := func(e *dnsTypes.Event) {
 				e.Timestamp = 0
+				// nslookup has several threads, so we prefer to normalize the command
+				// as it can be hard to guess which thread (nslookup, isc-net-0000,
+				// isc-timer or isc-socket-0) will do the DNS request.
+				e.Comm = ""
```
mauriciovasquezbernal, line 228: I suppose that by design it's always isc-worker0000 doing the request? I have tried multiple times and I always get the same thread.

integration/ig/k8s/trace_dns_test.go
mauriciovasquezbernal: I'd prefer not to normalize (remove) fields from the testing. Each time we "normalize" a field we reduce our test coverage. Let's keep `isc-worker0000`, if this starts causing issues we can easily normalize it.

PR #1787, opened 2023-06-26
ci: Publish ig container images.

eiffel-fl:
Hi.

This PR publishes ig container images when merged on main.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -349,11 +349,96 @@ jobs:
       if: github.event_name != 'pull_request'
       run: |
           echo "${{ matrix.type }}-${{ matrix.platform }}=${{ steps.publish-gadget-container-images.outputs.digest }}" >> $GITHUB_OUTPUT
-
     # old cache entries aren’t deleted, so the cache size keeps growing
     # remove old cache and move new cache to cache path to workaround the issue
     # https://github.com/docker/build-push-action/issues/252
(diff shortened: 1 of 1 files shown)


burak-ok:
> This one already uses `distroless` as it is quite easy in this case since this is just a golang binary:

I meant the other way around. Isn't it better to (additionally) provide an image which includes a shell, etc..?

alban:
> I meant the other way around. Isn't it better to (additionally) provide an image which includes a shell, etc..?

I would like to have an image with a shell.

When the image is published, I plan to test it on WSL2 and Docker Desktop for Windows. During my tests, I found out that the environment is quite different there. For example, I need to remount bpffs, debugfs and tracefs. And the pidns used in /host/proc is not a parent pidns of the container. So I added a [few workarounds in entrypoint.sh](https://github.com/inspektor-gadget/inspektor-gadget/blob/c9bdb667efc9443105f6420ce18b4fc63db65fe7/desktop-extension/entrypoint.sh) to make it work in this environment. For this, I needed busybox with stat, mount, pidof, etc. I think it is easier to implement those workarounds in a shell script rather than add them in the `ig` binary.

blanquicet:
Hi @alban, IIUC, all the use cases you mentioned are focused on development, and it will be helpful to us as maintainers/developers rather than for the final user. If that is the case, what do you think about keeping the image as small/simple as possible and, in case of need, from a developer pov, we can quickly rebuild the image by modifying the `BASE_IMAGE` and adding all the tools we need? We could even pass a variable to the make to build a debug image with shell and any other tool we might need for debugging.

PR #1796, opened 2023-06-28
Fix make kubectl-gadget-container

alban:
When the --experimental flag was added, it added a internal/deployinfo package that is needed to build kubectl-gadget in a container.

Symptoms:
```
cmd/kubectl-gadget/undeploy.go:25:2: no required module provides package github.com/inspektor-gadget/inspektor-gadget/internal/deployinfo; to add it:
  go get github.com/inspektor-gadget/inspektor-gadget/internal/deployinfo
make: *** [Makefile:121: kubectl-gadget-linux-amd64] Error 1
```

xref https://github.com/inspektor-gadget/inspektor-gadget/pull/1728

### DIFF
--- a/Dockerfiles/kubectl-gadget.Dockerfile.dockerignore
+++ b/Dockerfiles/kubectl-gadget.Dockerfile.dockerignore
@@ -4,6 +4,7 @@
 !cmd
 !pkg
 !internal/thirdparty
+!internal/deployinfo
 !tools
 !Makefile
 !*.mk


Dockerfiles/kubectl-gadget.Dockerfile.dockerignore
```diff
@@ -4,6 +4,7 @@
 !cmd
 !pkg
 !internal/thirdparty
+!internal/deployinfo
```
mauriciovasquezbernal, line 7: Why not everything in `internal` right away to avoid this breaking again in the future once we add another package?

alban: No good reason, I should have done that indeed.

PR #1803, opened 2023-06-30
integration: Skip RunTraceOpen on arm64.

eiffel-fl:
BYOB does not filter out unexisting trace points.
    Sadly, arm64 does not define open() syscall, only openat().
    Then, when we try to load the corresponding eBPF object file, it will fail due
    to this.
    
 Indeed, at the moment, we do not filter out unexisting tracepoints in BYOB like
    we do in trace open [1].
    
 Fixes: 34e6e36d3e0b ("integration: Add basic test for "run trace open"")


burak-ok (approved):
Thanks for fixing the test. I totally forgot about `arm64` 🙈

PR #1804, opened 2023-06-30
ci: test run trace open for arm64

eiffel-fl:


burak-ok (approved):
Again thanks for the quick troubleshoot and fix. LGTM

PR #1806, opened 2023-06-30
Use KubeIPResolver in remaining networking gadgets

### LINKED ISSUE #1568: gadgets: Use KubeIPResolver in networking gadgets

burak-ok:
# Use KubeIPResolver in remaining networking gadgets

With #1626 it got easier to add the `KubeIPResolver` to gadgets. This PR adds it to the remaining three gadgets:
1. trace/tcpconnect
2. top/tcp
3. snapshot/socket

Closes #1568


docs/gadgets/top/tcp.md
```diff
@@ -32,9 +32,9 @@ $ kubectl exec -ti test-pod -- wget kinvolk.io
 On *the first terminal*, you should see:
 
 ```
-NODE            NAMESPACE       POD             CONTAINER       PID     COMM    IP REMOTE                LOCAL                 SENT    RECV
-minikube        default         test-pod        test-pod        134110  wget    4  188.114.96.3:443      172.17.0.2:38190      0       2
-minikube        default         test-pod        test-pod        134110  wget    4  188.114.96.3:80       172.17.0.2:33286      0       1
+NODE             NAMESPACE        POD              CONTAINER        PID       COMM      IP SRC                   DST                   SENT     RECV
+minikube-docker  default          test-pod         test-pod         2176620   wget      4  p/default/test-pod:41 r/1.1.1.1:443         296B     58.12KiB
+minikube-docker  default          test-pod         test-pod         2176618   wget      4  p/default/test-pod:47 r/1.1.1.1:80          70B      381B
```
mauriciovasquezbernal, line 37: Is this being trimmed without appending the ellipsis? Or did it used 41 or 47 as source ports? (I don't believe so)

pkg/gadgets/snapshot/socket/types/socket-collector.go
```diff
@@ -60,25 +62,26 @@ func GetColumns() *columns.Columns[Event] {
...
-	}, func(e *Event) string {
-		return fmt.Sprintf("%s:%d", e.RemoteAddress, e.RemotePort)
-	})
+	eventtypes.MustAddVirtualL4EndpointColumn(
+		cols,
+		columns.Attributes{
+			Name:    "src",
+			Visible: true,
+			Width:   40,
+			Order:   1000,
```
mauriciovasquezbernal, line 71: I'm wondering if we should have a `src` and `dst` templates for this.

mauriciovasquezbernal (approved):
LGTM. I really like this new approach to handling endpoints.

docs/gadgets/top/tcp.md
burak-ok: I think I made a mistake during copy/paste. Fixed it

pkg/gadgets/snapshot/socket/types/socket-collector.go
burak-ok: We should definitely do that +1

pkg/columns/columninfo.go
```diff
@@ -156,6 +156,23 @@ func (ci *Column[T]) fromTag(tag string) error {
 	return ci.parseTagInfo(tagInfo[1:])
 }
 
+func (ci *Column[T]) applyTemplate() error {
+	// Apply template, if present
+	if ci.Template != "" {
```
flyth, line 161: nit: reverse this (return nil early if empty)

pkg/columns/columninfo.go
```diff
@@ -156,6 +156,23 @@ func (ci *Column[T]) fromTag(tag string) error {
 	return ci.parseTagInfo(tagInfo[1:])
 }
 
+func (ci *Column[T]) applyTemplate() error {
```
flyth, line 159: Maybe also use this [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/fedee8c233071d29a29992b445eccb560e1b617a/pkg/columns/columns.go#L331)?

flyth (approved):
Thanks for the updates, LGTM!

PR #1810, opened 2023-07-03
pkg/gadgets: Use paramIP for profile tcprtt.

eiffel-fl:


burak-ok (approved):
LGTM thanks for finding this

EDIT: Seems that I commented without looking at the tests too. Seems like we somehow need to allow empty strings of course. Do we generally always want to allow empty strings as IP addresses?

pkg/params/validators.go
```diff
@@ -137,7 +137,7 @@ func ValidateDuration(value string) error {
 }
 
 func ValidateIP(value string) error {
-	if net.ParseIP(value) == nil {
+	if value != "" && net.ParseIP(value) == nil {
```
burak-ok, line 140: I would additionally add a testcase for that so the wanted behavior is "documented"

burak-ok: Even better :D

Issue #1776, opened 2023-06-22
`TestTraceNetwork` failing sometimes in CI


mauriciovasquezbernal:
The test failed a couple of times in https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/5323662284/attempts/2#summary-14407553312. 

It seems that some entries are missing the container metadata:

```
{"timestamp":1687275447766044955,"type":"normal","netnsid":4026532224,"uid":0,"gid":0,"pktType":"OUTGOING","proto":"tcp","port":80,"remoteAddr":"10.244.0.12"}
```

It could be related to https://github.com/inspektor-gadget/inspektor-gadget/pull/1630, @alban suggested that "It seems the socket enricher fails sometimes but not always, and only client-side... maybe the sockets map is full? That would explain why it works for nginx because its sockets are created before wget's sockets"

Reported in https://kubernetes.slack.com/archives/CSYL75LF6/p1687274504983709

alban:
In the logs linked above, the `trace network` gadget shows output from both various Kubernetes namespaces:
- test-profile-tcprtt-*
- test-top-tcp-*

This is because the test `TestTraceNetwork` runs in parallel with other tests. This is the `ig` test (and not the `kubectl-gadget` test), so it shows events from all Kubernetes namespaces because it does not have a `-n` filter.

The socket enricher cleans up sockets from the map [5 seconds](https://github.com/inspektor-gadget/inspektor-gadget/blob/53631a7ed2ae0d102bb25b4c839b82701067270a/pkg/gadgets/internal/socketenricher/bpf/sockets-map.bpf.c#L182) after they are deleted. This is checked every [2 seconds](https://github.com/inspektor-gadget/inspektor-gadget/blob/53631a7ed2ae0d102bb25b4c839b82701067270a/pkg/gadgets/internal/socketenricher/tracer.go#L168), so at worst the socket is deleted after `~7` seconds. The sockets map can have [16384 entries](https://github.com/inspektor-gadget/inspektor-gadget/blob/53631a7ed2ae0d102bb25b4c839b82701067270a/pkg/gadgets/internal/socketenricher/bpf/sockets-map.h#L83).

[...]

PR #1812, opened 2023-07-03
Add IPv6 support to `snapshot socket`

eiffel-fl:
Hi.

This PR adds IPv6 support to `snapshot socket`:

```bash
$ python
>>> import socket
>>> s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
>>> s.bind(('::1', 8082, 0, 0))
>>> s.listen(5)
zsh: suspended  python3
$ sudo ig snapshot socket --host
CONTAINER                                                                PROTOCOL LOCAL                                  REMOTE                                STATUS
...
                                                                         TCP      ::1:8082                               :::0                                  LISTEN 
...
```

Best regards.

### DIFF
--- a/pkg/gadgets/snapshot/socket/tracer/tracer_test.go
+++ b/pkg/gadgets/snapshot/socket/tracer/tracer_test.go
@@ -90,6 +90,41 @@ func TestSnapshotSocket(t *testing.T) {
 			},
 		},
 		{
+			name:  "listen_tcp_v6",
+			proto: types.TCP,
+			addr:  "::1",
+			port:  8082,
+			expectedEvent: func(info *utilstest.RunnerInfo, _ any) *types.Event {
+				return &types.Event{
+					Event:       eventtypes.Event{Type: eventtypes.NORMAL},
+					WithNetNsID: eventtypes.WithNetNsID{NetNsID: info.NetworkNsID},
+					Protocol:    "TCP",
+					Status:      "LISTEN",
+					SrcEndpoint: eventtypes.L4Endpoint{
+						L3Endpoint: eventtypes.L3Endpoint{
+							Addr: "::1",
+						},
+						Port: 8082,
+					},
+					DstEndpoint: eventtypes.L4Endpoint{
+						L3Endpoint: eventtypes.L3Endpoint{
+							// There is no connection in this test, so the remote address is null.
+							Addr: "::",
+						},
+					},
+				}
+			},
+			socketCreator: func(addr string, port int) error {
+				conn, err := net.Listen("tcp", fmt.Sprintf("[%s]:%d", addr, port))
+				if err != nil {
+					return fmt.Errorf("listening to %s: %w", addr, err)
+				}
+				t.Cleanup(func() { conn.Close() })
+
+				return nil
+			},
+		},
+		{
 			name:  "listen_udp_v4",
 			proto: types.UDP,
 			addr:  "127.0.0.1",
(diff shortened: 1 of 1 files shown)


pkg/gadgets/snapshot/socket/tracer/bpf/socket-common.h
```diff
@@ -66,13 +73,36 @@ static __always_inline void socket_bpf_seq_print(struct seq_file *seq,
...
+     * in this format:
+     *
+     * protocol: "TCP" or "UDP"
+     * family: AF_INET
+     * IP addresses: Printed as IPv6 notation (https://www.kernel.org/doc/html/latest/core-api/printk-formats.html#ipv6-addresses).
+     * ports: Hexadecimal in host-byte order.
+     * state: Hexadecimal of https://github.com/torvalds/linux/blob/v5.13/include/net/tcp_states.h#L12-L24
+     * ino: unsigned long.
+     */
+    BPF_SEQ_PRINTF(seq, "%s %04X %pI6c %04X %pI6 %04X %02X %lu\n",
```
alban, line 103: Why different formats for the source and destination IPs?
`%pI6c` vs  `%pI6`

alban:
While you're working on this, it would be nice to print the `v6only` bool from IPv6 sockets.
```
sk->__sk_common.skc_ipv6only
```
See https://github.com/inspektor-gadget/inspektor-gadget/issues/1776#issuecomment-1603030119

(could be done in a different PR)

pkg/gadgets/snapshot/socket/tracer/bpf/socket-common.h
```diff
@@ -66,13 +73,36 @@ static __always_inline void socket_bpf_seq_print(struct seq_file *seq,
...
+     * protocol: "TCP" or "UDP"
+     * family: AF_INET
+     * IP addresses: Printed as IPv6 notation (https://www.kernel.org/doc/html/latest/core-api/printk-formats.html#ipv6-addresses).
+     * ports: Hexadecimal in host-byte order.
+     * state: Hexadecimal of https://github.com/torvalds/linux/blob/v5.13/include/net/tcp_states.h#L12-L24
+     * ino: unsigned long.
+     */
+    BPF_SEQ_PRINTF(seq, "%s %04X %pI6 %04X %pI6 %04X %02X %lu\n",
+        protocol, AF_INET6, src, bpf_ntohs(srcp),
+        dest, bpf_ntohs(destp), state, ino);
```
mauriciovasquezbernal, line 105: This is not fully related to your PR but at this point I'm wondering if we should switch to a binary representation (use bpf_seq_write as in pkg/gadgets/top/ebpf/piditer/bpf/pid_iter.bpf.c) instead of parsing a string. IMO that's more efficient and easier to handle. 

IIUC the advantage of using `BPF_SEQ_PRINTF` is that one can pin the iterator and then any application can ready it getting a text representation, but for us this is not a valuable use case as it's the gadget itself reading and parsing it.

alban: In general I would agree when the struct has a fixed size, which is the case in all our current snapshot gadgets, but not in the new "snapshot file" gadget using `bpf_seq_printf_btf()` to output lengthy text.

Other considerations:
- Ideally, if we change the parsing, it should be something practical for containerized gadgets (using either BTF or some generic text parsing method)
- We still have to fix the issue with the `comm` field contains a `\n`.

mauriciovasquezbernal:
I'll suggest to rebase this on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/1895, the result code will be much simpler IMO.

mauriciovasquezbernal:
I don't have any issue. Let's get this in and then I'll rebase mine on top.

mauriciovasquezbernal:
I also can implement ipv6 support on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/1895, so we don't waste time reviewing this.

blanquicet:
From my point of view, it makes much more sense to rebased this on top of the refactoring introduced in https://github.com/inspektor-gadget/inspektor-gadget/pull/1895, which IMO, is ready to go.

mauriciovasquezbernal (approved):
LGTM. Thanks for handling all the comments.

PR #1815, opened 2023-07-04
pkg/gadgets: Add unit test for snapshot socket.

eiffel-fl:


pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+func TestSocketTracerCreate(t *testing.T) {
+	t.Parallel()
+
+	utilstest.RequireRoot(t)
+
+	tracer, err := NewTracer(types.ALL)
+	if err != nil {
+		t.Fatalf("creating tracer: %v", err)
+	}
```
mauriciovasquezbernal, line 39: nit: Please use `github.com/stretchr/testify` as suggested in our contribution guidelines.

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+	}
+
+	for _, event := range events {
+		// Normalize few fields before comparing:
+		// 1. There is no connection in this test, so there is no remote address.
+		// 2. This is hard to guess the inode number, let's normalize it for the
+		// moment.
+		// 3. We do not want to get the net namespace ID associated to PID 1, so
+		// let's normalize it too.
```
mauriciovasquezbernal, line 86: What about using https://github.com/inspektor-gadget/inspektor-gadget/blob/2da10c9b67234afc067016374d41f6ecb9806caf/internal/test/runner.go#L71 to run this on a different namespace and check it's able to gather the right information there?

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,156 @@
...
+		event.DstEndpoint.Addr = ""
+		event.InodeNumber = 0
+		event.NetNsID = 0
+
+		if reflect.DeepEqual(*event, expectedEvent) {
+			return
+		}
+	}
```
mauriciovasquezbernal, line 94: IMO this test can be refactored in the following way:
1. Use a table driven approach: It'll help us to easily introduce other tests (like connected sockets) in the future
2. Reuse existing helpers like [ExpectAtLeastOneEvent](https://github.com/inspektor-gadget/inspektor-gadget/blob/2da10c9b67234afc067016374d41f6ecb9806caf/internal/test/helpers.go#L114)

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
mauriciovasquezbernal: Right, unfortunately it's not so straightforward, in this case `runner.Info.Tid` has to be used because the runner testing infrastructure only switches network namespace for a single thread and not the entire process. I was able to make it work on https://github.com/inspektor-gadget/inspektor-gadget/commit/b6a89c23d0ce2068a7432a235b087cdb0fc0d8dc.

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func TestSnapshotSocket(t *testing.T) {
+	t.Parallel()
+
+	utilstest.RequireRoot(t)
+
+	cases := []testCase{
+		{
+			name:  "tcp_v4",
```
mauriciovasquezbernal, line 58: Perhaps be a more specific in the name. `tcp_v4_listening`?

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,151 @@
...
+				c.expectedEvent.DstEndpoint = eventtypes.L4Endpoint{
+					L3Endpoint: eventtypes.L3Endpoint{
+						// There is no connection in this test, so there remote address is null.
+						Addr: "0.0.0.0",
+					},
+				}
+				c.expectedEvent.NetNsID = runner.Info.NetworkNsID
+
+				return &c.expectedEvent
+			})(t, nil, extra{c.addr, uint16(c.port)}, events)
```
mauriciovasquezbernal, line 148: I think this function should be in the test definition (like [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/70010a391b9d5b9f908bc75d5da08b8c03ef7ceb/pkg/gadgets/internal/socketenricher/tracer_test.go#L77)). I agree this will cause some more code duplication but I think the extra flexibility is worth it.

mauriciovasquezbernal (commented):
It looks much better. Thanks for handling the comments. Only one blocking comment now, besides that LGTM!

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,176 @@
...
+			require.ErrorIsf(t, nil, err, "creating tracer: %v", err)
+			defer tracer.CloseIters()
+
+			evs, err := tracer.RunCollector(uint32(runner.Info.Tid), "", "", "")
+			require.ErrorIsf(t, nil, err, "running collector: %v", err)
+
+			type extra struct {
+				addr string
+				port uint16
+			}
```
mauriciovasquezbernal, line 157: not used.

pkg/gadgets/snapshot/socket/tracer/tracer_test.go
```diff
@@ -0,0 +1,176 @@
...
+	require.Nil(t, err, "creating tracer: %v", err)
+
+	tracer.CloseIters()
+}
+
+type extraInformation struct {
+	addr    string
+	port    uint16
+	netNsID uint64
+}
```
mauriciovasquezbernal, line 46: The original idea of having this "extra" structure was to be able to pass information that is dynamic, for instance, the port a socket uses when it's bound to `0`. In this case, `addr` and `port` are known at test creation, and `netNsID` is available on `runner.Info`. I used the following to make it work without this structure:

```diff
diff --git pkg/gadgets/snapshot/socket/tracer/tracer_test.go pkg/gadgets/snapshot/socket/tracer/tracer_test.go
index 41b067d4..6c713395 100644
--- pkg/gadgets/snapshot/socket/tracer/tracer_test.go
+++ pkg/gadgets/snapshot/socket/tracer/tracer_test.go
@@ -39,18 +39,12 @@ func TestSocketTracerCreate(t *testing.T) {
        tracer.CloseIters()
 }
 
-type extraInformation struct {
-       addr    string
-       port    uint16
-       netNsID uint64
-}
-
 type testCase struct {
        name          string
        proto         types.Proto
        addr          string
        port          int
-       expectedEvent func(info *utilstest.RunnerInfo, e extraInformation) *types.Event
+       expectedEvent func(info *utilstest.RunnerInfo, _ any) *types.Event
        socketCreator func(addr string, port int) error
 }
 
@@ -65,17 +59,17 @@ func TestSnapshotSocket(t *testing.T) {
                        proto: types.TCP,
                        addr:  "127.0.0.1",
                        port:  8082,
-                       expectedEvent: func(info *utilstest.RunnerInfo, e extraInformation) *types.Event {
+                       expectedEvent: func(info *utilstest.RunnerInfo, _ any) *types.Event {
                                return &types.Event{
                                        Event:       eventtypes.Event{Type: eventtypes.NORMAL},
-                                       WithNetNsID: eventtypes.WithNetNsID{NetNsID: e.netNsID},
+                                       WithNetNsID: eventtypes.WithNetNsID{NetNsID: info.NetworkNsID},
                                        Protocol:    "TCP",
                                        Status:      "LISTEN",
                                        SrcEndpoint: eventtypes.L4Endpoint{
                                                L3Endpoint: eventtypes.L3Endpoint{
-                                                       Addr: e.addr,
+                                                       Addr: "127.0.0.1",
                                                },
-                                               Port: e.port,
+                                               Port: 8082,
                                        },
                                        DstEndpoint: eventtypes.L4Endpoint{
                                                L3Endpoint: eventtypes.L3Endpoint{
@@ -100,17 +94,17 @@ func TestSnapshotSocket(t *testing.T) {
                        proto: types.UDP,
                        addr:  "127.0.0.1",
                        port:  8082,
-                       expectedEvent: func(info *utilstest.RunnerInfo, e extraInformation) *types.Event {
+                       expectedEvent: func(info *utilstest.RunnerInfo, _ any) *types.Event {
                                return &types.Event{
                                        Event:       eventtypes.Event{Type: eventtypes.NORMAL},
-                                       WithNetNsID: eventtypes.WithNetNsID{NetNsID: e.netNsID},
+                                       WithNetNsID: eventtypes.WithNetNsID{NetNsID: info.NetworkNsID},
                                        Protocol:    "UDP",
                                        Status:      "INACTIVE",
                                        SrcEndpoint: eventtypes.L4Endpoint{
                                                L3Endpoint: eventtypes.L3Endpoint{
-                                                       Addr: e.addr,
+                                                       Addr: "127.0.0.1",
                                                },
-                                               Port: e.port,
+                                               Port: 8082,
                                        },
                                        DstEndpoint: eventtypes.L4Endpoint{
                                                L3Endpoint: eventtypes.L3Endpoint{
@@ -151,11 +145,6 @@ func TestSnapshotSocket(t *testing.T) {
                        evs, err := tracer.RunCollector(uint32(runner.Info.Tid), "", "", "")
                        require.ErrorIsf(t, nil, err, "running collector: %v", err)
 
-                       type extra struct {
-                               addr string
-                               port uint16
-                       }
-
                        events := make([]types.Event, len(evs))
                        for i, ev := range evs {
                                events[i] = *ev
@@ -165,12 +154,7 @@ func TestSnapshotSocket(t *testing.T) {
                                events[i].InodeNumber = 0
                        }
 
-                       utilstest.ExpectAtLeastOneEvent(c.expectedEvent)(t, nil, extraInformation{
-                               addr:    c.addr,
-                               port:    uint16(c.port),
-                               netNsID: runner.Info.NetworkNsID,
-                       },
-                               events)
+                       utilstest.ExpectAtLeastOneEvent(c.expectedEvent)(t, runner.Info, nil, events)
                })
        }
 }

```

mauriciovasquezbernal (approved):
One last comment to avoid using `extraInformation`. LGTM! 

Thanks for implementing tests!

PR #1816, opened 2023-07-04
socket enricher: support dual stack sockets

alban:
When gadget_socket_lookup() is trying to do a lookup for an IPv4 socket, it also needs to do a second lookup for IPv6 and checks for skc_ipv6only in case there is a dual stack socket.

## How to use

After this patch, the socket enricher is able to handle the kube-apiserver sockets:

```
$ go run -exec 'sudo' ./cmd/ig/... -r docker trace network -c minikube-docker
CONTAINER                    PID        TID        COMM            TYPE      PROTO PORT  REMOTE
minikube-docker              692650     692704     kubelet         OUTGOING  tcp   8443  192.168.58.2
minikube-docker              693180     693180     kube-apiserver  HOST      tcp   8443  192.168.58.2
```

See https://github.com/inspektor-gadget/inspektor-gadget/issues/1776#issuecomment-1603030119


pkg/gadgets/internal/socketenricher/bpf/sockets-map.h
```diff
@@ -77,6 +77,7 @@ struct sockets_value {
 	__u64 uid_gid;
 	char task[TASK_COMM_LEN];
 	__u64 sock;
+	__u64 ipv6only;
```
mauriciovasquezbernal, line 80: Do we really need 64bits for this? Can't we move this as the last element and use a bitfield in case we want to other small fields in the future?

pkg/gadgets/internal/socketenricher/bpf/sockets-iter.bpf.c
```diff
@@ -31,6 +31,8 @@ insert_socket_from_iter(struct sock *sock, struct task_struct *task)
 	__builtin_memcpy(&socket_value.task, task->comm, sizeof(socket_value.task));
 	socket_value.mntns = (u64) task->nsproxy->mnt_ns->ns.inum;
 	socket_value.sock = (__u64) sock;
+	if (socket_key.family == AF_INET6)
```
mauriciovasquezbernal, line 34: What do you think about removing this conditional as `skc_ipv6only` should be false when `socket_key.family != AF_INET6`? Do you prefer to keep this just in case?

mauriciovasquezbernal (approved):
LGTM!

I also added some unit tests for checking if this is working properly, feel free to cherry-pick / squash them if you want -> https://github.com/inspektor-gadget/inspektor-gadget/commit/6aa9767a260fab6b3f5eef88d0847b086e876307

integration/inspektor-gadget/advise_networkpolicy_test.go
```diff
@@ -35,15 +35,23 @@ func TestAdviseNetworkpolicy(t *testing.T) {
 
 	commands := []*Command{
 		CreateTestNamespaceCommand(nsServer),
-		BusyboxPodRepeatCommand(nsServer, "nc -lk -p 9090 -e /bin/cat"),
+		BusyboxPodRepeatCommand(
+			nsServer,
+			// Dual stack socket
+			"nc -lk -p 9090 -e /bin/cat",
+		),
```
mauriciovasquezbernal, line 42: I suppose this will also work on ipv4 only systems as the traffic we're generating is not only ipv6, right?

pkg/gadgets/internal/socketenricher/bpf/sockets-iter.bpf.c
alban: Yes, it should work. Let me try.

integration/inspektor-gadget/advise_networkpolicy_test.go
alban: Yes; in this test, the traffic is ipv4.

alban:
Thanks. I've picked up your unit test and I made the other changes.

PR #1817, opened 2023-07-04
gadgets/run: Add JSON output

burak-ok:
# gadgets/run: Add JSON output

This PR adds a new formatter which outputs json "through" the columns library.

This has the advantage that the columns and json output will always be aligned and the same. Additionally we won't need duplicate logic for the different output modes in the `run` gadget

## Last commit about testing 
**WIP and proposal. Ideas really welcome :)**

I just added another commit which adds some integration test helpers and modifies the test for `TestRunTraceOpen`.

We can't use the traditional functions for the containerized gadgets.
They were directly unmarshaling into the Event object which we don't have with containerized gadgets.

Most of the columns are dynamically created and therefore the output json is also mostly dynamic and depends solely on the BTF and config from the input.

So with this commit the expected entry is now a `map[string]interface{}` which gets constructed through `RunEventToJsonObj(...)` and `MergeJsonObjs(...)` where the actual payload of the eBPF program is merged into the json object returned by `RunEventToJsonObj(...)` 

The rational behind creating helper functions such as `SetEventTimestamp` is that we can control the json path in a single location. So when the field changes in the future we don't have to change every test, only that function.

### DIFF
--- a/integration/run_helpers.go
+++ b/integration/run_helpers.go
@@ -1,0 +1,176 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
+	columns_json "github.com/inspektor-gadget/inspektor-gadget/pkg/columns/formatter/json"
+	runtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/run/types"
+	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+)
+
+func SetEventTimestamp(jsonObj map[string]interface{}, timestamp eventtypes.Time) {
+	jsonObj["timestamp"] = timestamp.String()
+}
+
+func SetEventMountNsID(jsonObj map[string]interface{}, mountNsID uint64) {
+	jsonObj["mntns"] = mountNsID
+}
+
+func SetEventRuntimeName(jsonObj map[string]interface{}, runtimeName eventtypes.RuntimeName) {
+	runtimeMetadata := jsonObj["runtime"].(map[string]interface{})
+	if runtimeMetadata != nil {
+		runtimeMetadata["runtimeName"] = runtimeName
+	}
+}
+
+func SetEventRuntimeContainerID(jsonObj map[string]interface{}, s string) {
+	runtimeMetadata := jsonObj["runtime"].(map[string]interface{})
+	if runtimeMetadata != nil {
+		runtimeMetadata["containerId"] = s
+	}
+}
+
+func SetEventRuntimeContainerName(jsonObj map[string]interface{}, s string) {
+	runtimeMetadata := jsonObj["runtime"].(map[string]interface{})
+	if runtimeMetadata != nil {
+		runtimeMetadata["containerName"] = s
+	}
+}
+
+func SetEventK8sNode(jsonObj map[string]interface{}, s string) {
+	jsonObj["node"] = s
+}
+
+func SetEventK8sNamespace(jsonObj map[string]interface{}, s string) {
+	jsonObj["namespace"] = s
+}
+
+func SetEventK8sPod(jsonObj map[string]interface{}, s string) {
+	jsonObj["pod"] = s
+}
+
+func SetEventK8sContainer(jsonObj map[string]interface{}, s string) {
+	jsonObj["container"] = s
+}
+
+func SetEventK8sHostNetwork(jsonObj map[string]interface{}, b bool) {
+	jsonObj["hostnetwork"] = b
+}
+
+func RunEventToObj(t *testing.T, ev *runtypes.Event) map[string]interface{} {
+	cols := columns.MustCreateColumns[runtypes.Event]()
+	jsonFormatter := columns_json.NewFormatter[runtypes.Event](cols.GetColumnMap())
+
+	jsonStr := jsonFormatter.FormatEntry(ev)
+
+	var jsonObj map[string]interface{}
+	err := json.Unmarshal([]byte(jsonStr), &jsonObj)
+	require.NoError(t, err, "unmarshalling event json")
+
+	delete(jsonObj, "raw_data")
+	delete(jsonObj, "data")
+
+	return jsonObj
+}
+
+// MergeJsonObjs merges two JSON objects (map[string]interface{}) and returns a copy.
+// The original objects are not modified in order to use the base for multiple merges.
+// If a key exists in both objects, the function fails.
+func MergeJsonObjs(t *testing.T, base map[string]interface{}, additionalFields map[string]interface{}) map[string]interface{} {
+	// First copy the base map
+	target := map[string]interface{}{}
+	for k, v := range base {
+		target[k] = v
+	}
+	// Then merge the additional fields and fail if a field already exists
+	for k, v := range additionalFields {
+		_, ok := target[k]
+		require.False(t, ok, fmt.Sprintf("key %s already exists in target map", k))
+		target[k] = v
+	}
+	return target
+}
+
+// parseMultiJSONOutputToObj parses a string containing multiple JSON objects
+// (one per line) into a slice of maps.
+func parseMultiJSONOutputToObj(output string, normalize func(map[string]interface{})) ([]map[string]interface{}, error) {
+	ret := []map[string]interface{}{}
+
+	decoder := json.NewDecoder(strings.NewReader(output))
+	for decoder.More() {
+		var entry map[string]interface{}
+		if err := decoder.Decode(&entry); err != nil {
+			return nil, fmt.Errorf("decoding json: %w", err)
+		}
+		if normalize != nil {
+			normalize(entry)
+		}
+
+		// Marshal & Unmarshal to have the right types in the map
+		bytes, err := json.Marshal(entry)
+		if err != nil {
+			return nil, fmt.Errorf("marshalling json: %w", err)
+		}
+		err = json.Unmarshal(bytes, &entry)
+		if err != nil {
+			return nil, fmt.Errorf("unmarshalling json: %w", err)
+		}
+		ret = append(ret, entry)
+	}
+	return ret, nil
+}
+
+// expectEntriesToMatchObj verifies that all the entries in expectedEntries are
+// matched by at least one entry in the output
+func expectEntriesToMatchObj(t *testing.T, entries []map[string]interface{}, expectedEntries ...map[string]interface{}) error {
+out:
+	for _, expectedEntry := range expectedEntries { // Marshal & Unmarshal to have the right types in the map
+		bytes, err := json.Marshal(expectedEntry)
+		require.NoError(t, err, "marshalling expectedEntry")
+		err = json.Unmarshal(bytes, &expectedEntry)
+		require.NoError(t, err, "unmarshalling expectedEntry")
+
+		for _, entry := range entries {
+			if entry["fname"] == "/dev/null" {
+				fmt.Println("actual entry:", entry)
+			}
+			if reflect.DeepEqual(expectedEntry, entry) {
+				continue out
+			}
+		}
+		expectedJsonEntry, _ := json.Marshal(expectedEntry)
+		return fmt.Errorf("output doesn't contain the expected entry: %+v", string(expectedJsonEntry))
+	}
+	return nil
+}
+
+// ExpectEntriesToMatchObj verifies that all the entries in expectedEntries are
+// matched by at least one entry in the output (Lines of independent JSON objects).
+func ExpectEntriesToMatchObj(t *testing.T, output string, normalize func(map[string]interface{}), expectedEntries ...map[string]interface{}) error {
+	entries, err := parseMultiJSONOutputToObj(output, normalize)
+	if err != nil {
+		return err
+	}
+	return expectEntriesToMatchObj(t, entries, expectedEntries...)
+}
--- a/pkg/columns/columns.go
+++ b/pkg/columns/columns.go
@@ -15,6 +15,7 @@
 package columns
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"reflect"
@@ -502,6 +503,26 @@ func GetFieldFuncExt[OT any, T any](column ColumnInternals, raw bool) func(entry
 	}
 }
 
+// GetFieldAsArrayFunc returns a helper function to access an array of type OT of a struct T
+// without using reflection. It does not differentiate between direct members of the struct and
+// members of embedded structs.
+func GetFieldAsArrayFunc[OT any, T any](column ColumnInternals) func(entry *T) []OT {
+	l := column.(*Column[T]).RawType().Len()
+
+	return func(entry *T) []OT {
+		entryStart := unsafe.Pointer(entry)
+		if column.(*Column[T]).getStart != nil {
+			entryStart = column.(*Column[T]).getStart(entry)
+		}
+
+		fieldStart := unsafe.Add(entryStart, column.getOffset())
+		srcSlice := unsafe.Slice((*OT)(fieldStart), l)
+		r := make([]OT, l)
+		copy(r, srcSlice)
+		return r
+	}
+}
+
 // SetFieldFunc returns a helper function to set the value of type OT to a member of struct T
 // without using reflection. It differentiates between direct members of the struct and
 // members of embedded structs. If any of the embedded structs being accessed is a nil-pointer,
@@ -580,28 +601,16 @@ func GetFieldAsStringExt[T any](column ColumnInternals, floatFormat byte, floatP
 			return "false"
 		}
 	case reflect.Array:
-		l := column.(*Column[T]).Type().Len()
 		s := column.(*Column[T]).Type().Elem().Size()
 		// c strings: []char null terminated
 		if s == 1 {
 			return func(entry *T) string {
-				r := []byte{}
-				entryStart := unsafe.Pointer(entry)
-				if column.(*Column[T]).getStart != nil {
-					entryStart = column.(*Column[T]).getStart(entry)
+				arr := GetFieldAsArrayFunc[byte, T](column)(entry)
+				i := bytes.IndexByte(arr, 0)
+				if i != -1 {
+					arr = arr[:i]
 				}
-
-				fieldStart := unsafe.Add(entryStart, column.getOffset())
-
-				for i := 0; i < int(l); i++ {
-					b := *(*byte)(unsafe.Add(fieldStart, i))
-					if b == 0 {
-						break
-					}
-					r = append(r, b)
-				}
-
-				return string(r)
+				return string(arr)
 			}
 		}
 
--- a/pkg/columns/formatter/json/json.go
+++ b/pkg/columns/formatter/json/json.go
@@ -35,6 +35,7 @@ type column[T any] struct {
 type Formatter[T any] struct {
 	options *Options
 	columns []*column[T]
+	printer func(entry *T, buf *encodeState)
 }
 
 // NewFormatter returns a Formatter that will turn entries of type T into JSON representation
@@ -126,13 +127,16 @@ func NewFormatter[T any](cols columns.ColumnMap[T], options ...Option) *Formatte
 		options: opts,
 		columns: ncols,
 	}
+	tf.printer = tf.getPrinter(0, ncols)
 	return tf
 }
 
-func (f *Formatter[T]) printCols(entry *T, buf *encodeState, cols []*column[T], level int) {
+func (f *Formatter[T]) getPrinter(level int, cols []*column[T]) func(entry *T, buf *encodeState) {
 	indent := strings.Repeat("  ", level+1)
 	colsWithParent := map[string][]*column[T]{}
 	colsWithoutParent := []*column[T]{}
+
+	funcs := make([]func(entry *T, buf *encodeState), 0)
 
 	for _, col := range cols {
 		// Save columns which have a parent in its name
@@ -153,15 +157,19 @@ func (f *Formatter[T]) printCols(entry *T, buf *encodeState, cols []*column[T], 
 	// Better name?
 	handlePrefix := func() {
 		if !first {
-			buf.WriteByte(',')
-			if f.options.prettyPrint {
-				buf.WriteString("\n" + indent)
-			} else {
-				buf.WriteByte(' ')
-			}
+			funcs = append(funcs, func(entry *T, buf *encodeState) {
+				buf.WriteByte(',')
+				if f.options.prettyPrint {
+					buf.WriteString("\n" + indent)
+				} else {
+					buf.WriteByte(' ')
+				}
+			})
 		} else {
 			if f.options.prettyPrint {
-				buf.WriteString("\n" + indent)
+				funcs = append(funcs, func(entry *T, buf *encodeState) {
+					buf.WriteString("\n" + indent)
+				})
 			}
 			first = false
 		}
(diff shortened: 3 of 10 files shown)


mauriciovasquezbernal:
I see this flattens the results! Great!

blanquicet:
I rebased it to test it with https://github.com/inspektor-gadget/inspektor-gadget/pull/1702 and this is the output:

```bash
$ sudo IG_EXPERIMENTAL=true ./ig run --prog @./gadgets/trace_tcpconnect_x86.bpf.o --definition @./gadgets/trace_tcpconnect.yaml -o json
{
  "runtime": {
    "runtimeName": "docker", "containerId": "c00c1df110fa46c87deb793ae9de5321c0a2b0f6931dea6088829f5ac96fed3a", "containerName": "test-container"}, "node": "", "namespace": "", "pod": "", "container": "", "hostnetwork": false, "timestamp": "1970-01-01T00:00:00.000000000Z", "mntns": 4026532265, "pid": 87921, "task": "wget", "uid": 0, "gid": 0, "version": 4, "src": "172.17.0.2", "dst": "1.1.1.1", "sport": 60550}
{
  "runtime": {
    "runtimeName": "docker", "containerId": "c00c1df110fa46c87deb793ae9de5321c0a2b0f6931dea6088829f5ac96fed3a", "containerName": "test-container"}, "node": "", "namespace": "", "pod": "", "container": "", "hostnetwork": false, "timestamp": "1970-01-01T00:00:00.000000000Z", "mntns": 4026532265, "pid": 87921, "task": "wget", "uid": 0, "gid": 0, "version": 4, "src": "172.17.0.2", "dst": "1.1.1.1", "sport": 56216}
```

The format is strange and it doesn't have the k8s sub-struct 🤔

blanquicet:
> > and it doesn't have the k8s sub-struct thinking
> 
> Since the `K8sMetadata` field in `CommonData` doesn't have a column name it will not show it as a substruct. This PR "retrieves" the JSON fields from the columns library. So the column name is important for this.

The problem with adding the column name to the `K8sMetadata` field in `CommonData` is that it implies adding `K8S.` prefix to all its columns: `K8S.NODE`, `K8S.PODNAME`, etc. It's okay for `ig` but not sure if it is what we want for `kubeclt-gadget`. I'm not sure how we want to manage this.

cc @mauriciovasquezbernal

burak-ok:
**WIP and proposal. Ideas really welcome :)**


I just added another commit which adds some integration test helpers and modifies the test for `TestRunTraceOpen`.

We can't use the traditional functions for the containerized gadgets.
They were directly unmarshaling into the Event object which we don't have with containerized gadgets.

Most of the columns are dynamically created and therefore the output json is also mostly dynamic and depends solely on the BTF and config from the input.

So with this commit the expected entry is now a `map[string]interface{}` which gets constructed through `RunEventToJsonObj(...)` and `MergeJsonObjs(...)` where the actual payload of the eBPF program is merged into the json object returned by `RunEventToJsonObj(...)` 

The rational behind creating helper functions such as `SetEventTimestamp` is that we can control the json path in a single location. So when the field changes in the future we don't have to change every test, only that function.

flyth:
> > > and it doesn't have the k8s sub-struct thinking
> > 
> > 
> > Since the `K8sMetadata` field in `CommonData` doesn't have a column name it will not show it as a substruct. This PR "retrieves" the JSON fields from the columns library. So the column name is important for this.
> 
> The problem with adding the column name to the `K8sMetadata` field in `CommonData` is that it implies adding `K8S.` prefix to all its columns: `K8S.NODE`, `K8S.PODNAME`, etc. It's okay for `ig` but not sure if it is what we want for `kubeclt-gadget`. I'm not sure how we want to manage this.
> 
> cc @mauriciovasquezbernal

IMHO it's okay to have the `k8s` substruct in JSON also for kubectl-gadget - in that environment it's maybe a bit too specific, but at least the 3rd party tools digesting that JSON can rely on the same data structure for both environments.

blanquicet:
> > > > and it doesn't have the k8s sub-struct thinking
> > > 
> > > 
> > > Since the `K8sMetadata` field in `CommonData` doesn't have a column name it will not show it as a substruct. This PR "retrieves" the JSON fields from the columns library. So the column name is important for this.
> > 
> > 
> > The problem with adding the column name to the `K8sMetadata` field in `CommonData` is that it implies adding `K8S.` prefix to all its columns: `K8S.NODE`, `K8S.PODNAME`, etc. It's okay for `ig` but not sure if it is what we want for `kubeclt-gadget`. I'm not sure how we want to manage this.
> > cc @mauriciovasquezbernal
> 
> IMHO it's okay to have the `k8s` substruct in JSON also for kubectl-gadget - in that environment it's maybe a bit too specific, but at least the 3rd party tools digesting that JSON can rely on the same data structure for both environments.

Yeah, it is already like that. The `runtime` and `k8s` substructs in JSON are already present for `kubectl-gadget`. The question is to add the `K8S.` prefix to the columns name in `kubeclt-gadget`.

flyth:
> > > > > and it doesn't have the k8s sub-struct thinking
> > > > 
> > > > 
> > > > Since the `K8sMetadata` field in `CommonData` doesn't have a column name it will not show it as a substruct. This PR "retrieves" the JSON fields from the columns library. So the column name is important for this.
> > > 
> > > 
> > > The problem with adding the column name to the `K8sMetadata` field in `CommonData` is that it implies adding `K8S.` prefix to all its columns: `K8S.NODE`, `K8S.PODNAME`, etc. It's okay for `ig` but not sure if it is what we want for `kubeclt-gadget`. I'm not sure how we want to manage this.
> > > cc @mauriciovasquezbernal
> > 
> > 
> > IMHO it's okay to have the `k8s` substruct in JSON also for kubectl-gadget - in that environment it's maybe a bit too specific, but at least the 3rd party tools digesting that JSON can rely on the same data structure for both environments.
> 
> Yeah, it is already like that. The `runtime` and `k8s` substructs in JSON are already present for `kubectl-gadget`. The question is to add the `K8S.` prefix to the columns name in `kubeclt-gadget`.

Ah, I see. We could add something like `WithStripPrefixes("k8s")` for the columns output (columnsformatter), if we want to get rid of it.

mauriciovasquezbernal:
> It's okay for ig but not sure if it is what we want for kubeclt-gadget. I'm not sure how we want to manage this.

It'd be better not to have it IMO. But if that's needed for this PR to work and removing it is difficult I'll just add.

pkg/columns/formatter/json/json.go
```diff
@@ -46,7 +47,12 @@ func NewFormatter[T any](cols columns.ColumnMap[T], options ...Option) *Formatte
 
 	ncols := make([]*column[T], 0)
 	for _, col := range cols.GetOrderedColumns() {
-		name, _ := json.Marshal(col.Name)
+		colName := col.Name
+		if strings.Contains(colName, ".") {
+			hierarchy := strings.Split(colName, ".")
+			colName = hierarchy[len(hierarchy)-1]
+		}
+		name, _ := json.Marshal(colName)
```
mauriciovasquezbernal, line 55: What about columns having a `a.b.c` name, will those be supported?

pkg/columns/formatter/json/json.go
```diff
@@ -124,39 +130,92 @@ func NewFormatter[T any](cols columns.ColumnMap[T], options ...Option) *Formatte
...
 
-	buf := bufpool.Get().(*encodeState)
-	buf.Reset()
-	defer bufpool.Put(buf)
-
-	buf.WriteByte('{')
-	if f.options.prettPrint {
-		buf.WriteByte('\n')
+	for _, col := range cols {
+		// Save columns which has a parent in its name
```
mauriciovasquezbernal, line 139: has -> have

pkg/columns/formatter/json/json.go
```diff
@@ -124,39 +130,92 @@ func NewFormatter[T any](cols columns.ColumnMap[T], options ...Option) *Formatte
...
+		if strings.Count(col.column.Name, ".") > level {
+			hierarchy := strings.Split(col.column.Name, ".")
+			parentName := hierarchy[level]
+			colsWithParent[parentName] = append(colsWithParent[parentName], col)
+			continue
+		}
+		// Don't print columns without a parent directly
+		// It is possible that their name is the same as a name of an object in the same level
+		// For example: {"a": {"bbb": 1}, "a": 2}
```
mauriciovasquezbernal, line 148: I don't get this point. Wouldn't it be an error to have two things under the same key? (Even if one of them is an object and the other is a column).

pkg/columns/formatter/json/json_test.go
```diff
@@ -141,3 +142,47 @@ func TestDynamicFields(t *testing.T) {
...
+	assert.NoError(t, cols.AddColumn(columns.Attributes{
+		Name: "parent.child2.grandchild1",
+	}, func(t *testStruct) string { return "parent.child2.grandchild1_text" }))
+	assert.NoError(t, cols.AddColumn(columns.Attributes{
+		Name: "parent.child3",
+	}, func(t *testStruct) string { return "parent.child3_text" }))
+
+	assert.NoError(t, cols.AddColumn(columns.Attributes{
+		Name: "parent.child1",
+	}, func(t *testStruct) string { return "This should be skipped/filtered" }))
```
mauriciovasquezbernal, line 166: Use `require` instead of `assert`: if there is an error here we want the test to fail immediately.

pkg/columns/formatter/json/json_test.go
```diff
@@ -141,3 +142,47 @@ func TestDynamicFields(t *testing.T) {
...
+	}, func(t *testStruct) string { return "parent.child1.grandchild2_text" }))
+	assert.NoError(t, cols.AddColumn(columns.Attributes{
+		Name: "parent.child2.grandchild1",
+	}, func(t *testStruct) string { return "parent.child2.grandchild1_text" }))
+	assert.NoError(t, cols.AddColumn(columns.Attributes{
+		Name: "parent.child3",
+	}, func(t *testStruct) string { return "parent.child3_text" }))
+
+	assert.NoError(t, cols.AddColumn(columns.Attributes{
+		Name: "parent.child1",
```
mauriciovasquezbernal, line 165: IMO this is a programmer issue. Perhaps this should return an error here instead of handling this when serializing the json.

pkg/columns/formatter/json/json_test.go
```diff
@@ -141,3 +142,47 @@ func TestDynamicFields(t *testing.T) {
...
+		}
+	}`
+
+	formatter := NewFormatter[testStruct](cols.ColumnMap)
+	actual := formatter.FormatEntry(&testStruct{})
+	assert.JSONEq(t, expected, actual)
+
+	prettyFormatter := NewFormatter[testStruct](cols.ColumnMap, WithPrettyPrint())
+	actual = prettyFormatter.FormatEntry(&testStruct{})
+	assert.JSONEq(t, expected, actual)
```
mauriciovasquezbernal, line 187: Shouldn't this check that the printed json has the same format? I mean, this is for testing pretty printed json, so I think it should compare the string itself and not only the json, otherwise the test will pass even if it's not pretty printed.

pkg/gadgets/run/tracer/run.go
```diff
@@ -272,10 +272,49 @@ func (g *GadgetDesc) CustomParser(params *params.Params, args []string) (parser.
...
 
+func (g *GadgetDesc) CustomParser(params *params.Params, args []string) (parser.Parser, error) {
+	cols, err := g.getColumns(params, args)
+	if err != nil {
+		return nil, err
+	}
 	return parser.NewParser[types.Event](cols), nil
 }
 
+func (g *GadgetDesc) CustomJsonParser(params *params.Params, args []string, options ...columns_json.Option) (*columns_json.Formatter[types.Event], error) {
```
mauriciovasquezbernal, line 286: Does it need to be exported?

pkg/gadgets/run/tracer/run.go
```diff
@@ -272,10 +272,49 @@ func (g *GadgetDesc) CustomParser(params *params.Params, args []string) (parser.
...
+	if err != nil {
+		return nil, err
+	}
+	return columns_json.NewFormatter(cols.ColumnMap, options...), nil
+}
+
+func (g *GadgetDesc) JSONConverter(params *params.Params, printer gadgets.Printer) func(ev any) {
+	formatter, err := g.CustomJsonParser(params, []string{})
+	if err != nil {
+		printer.Logf(logger.WarnLevel, "creating json formatter: %s", err)
```
mauriciovasquezbernal, line 297: mmm, I think we should find a way to propagate these errors, but we can handle it in a following PR.

mauriciovasquezbernal (commented):
Some comments. I need to make another pass and test it on different situations, specially with https://github.com/inspektor-gadget/inspektor-gadget/pull/1825. Btw, I'd propose to first merge this PR, so I'll rebase mine on top of this.

pkg/columns/formatter/json/json_test.go
```diff
@@ -59,6 +59,21 @@ func TestJSONFormatter_FormatEntry(t *testing.T) {
...
+		"{\n  \"name\": \"Eve\",\n  \"age\": 99,\n  \"size\": 5.12,\n  \"balance\": 1000000,\n  \"canDance\": false\n}",
+		"",
+	}
+	formatter := NewFormatter(testColumns, WithPrettyPrint())
+	for i, entry := range testEntries {
+		if res := formatter.FormatEntry(entry); res != expected[i] {
+			t.Errorf("got %s, expected %s", res, expected[i])
+		}
+	}
+}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I just realized it's about pretty print, I personally prefer to use ``` ` ``` and define those as multiline strings but it's up to you.

mauriciovasquezbernal (commented):
I rebased https://github.com/inspektor-gadget/inspektor-gadget/pull/1825 on top of this. 

Almost everything worked fine, but the port as it's printed as a string:

```json
{
  "dst": {
    "v": "4",
    "namespace": "",
    "kind": "",
    "addr": "1.1.1.1",
    "name": "",
    "port": "80"
  },
  "src": {
    "v": "4",
    "namespace": "",
    "addr": "172.17.0.3",
    "port": "49770",
    "kind": "",
    "name": ""
  },
  "runtime": {
    "runtimeName": "docker",
    "containerId": "2798f8dce13885fa2a19527659256b0e2c29f5751a37ce811df2d6c3033d57a8",
    "containerName": "mycontainer"
  },
  "node": "",
  "namespace": "",
  "pod": "",
  "container": "",
  "hostnetwork": false,
  "timestamp": "1969-12-31T19:00:00.000000000-05:00",
  "mntns": 4026536699,
  "pid": 69750,
  "task": "wget",
  "uid": 0,
  "gid": 0
}
```

The port there is taken from the enrichment endpoint (and not from the C struct), hence the extractor only returns a string. Given that the column is not only about "columns" anymore, I think it's time to make the extractor more generic and return `any`, so different types are correctly encoded to json. I'd propose to make it right away on this PR. @flyth do you think it's feasible?

pkg/columns/columns.go
```diff
@@ -502,6 +502,31 @@ func GetFieldFuncExt[OT any, T any](column ColumnInternals, raw bool) func(entry
 	}
 }
 
+// GetFieldAsArrayFunc returns a helper function to access an array of type OT of a struct T
+// without using reflection. It does not differentiates between direct members of the struct and
+// members of embedded structs.
+func GetFieldAsArrayFunc[OT any, T any](column ColumnInternals) func(entry *T) []OT {
+	l := column.(*Column[T]).Type().Len()
+	s := column.(*Column[T]).Type().Elem().Size()
```
flyth, line 510: `RawType` should be better suited here, as `Type` could be `string` if an extractor has been set.

pkg/columns/columns.go
```diff
@@ -502,6 +502,31 @@ func GetFieldFuncExt[OT any, T any](column ColumnInternals, raw bool) func(entry
...
+	return func(entry *T) []OT {
+		r := []OT{}
+		entryStart := unsafe.Pointer(entry)
+		if column.(*Column[T]).getStart != nil {
+			entryStart = column.(*Column[T]).getStart(entry)
+		}
+
+		fieldStart := unsafe.Add(entryStart, column.getOffset())
+
+		for i := 0; i < int(l); i += int(s) {
```
flyth, line 521: I guess since we're doing heavy unsafe magic here already, we could also just copy() the entire thing in one go to the preallocated `r`.

pkg/columns/columns.go
```diff
@@ -502,6 +502,31 @@ func GetFieldFuncExt[OT any, T any](column ColumnInternals, raw bool) func(entry
...
+			b := *(*OT)(unsafe.Add(fieldStart, i))
+			r = append(r, b)
+		}
```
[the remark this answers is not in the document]
burak-ok: > How does it work when `s != 1`? IIUC `l` is the number of elements in the array and `s` is the size (in bytes) of each element, so the loop should be something like:
> Am I missing something?

No, I just missed something and mixed up `l` and `s` :sweat_smile: . Thanks for the finding

pkg/columns/formatter/json/json.go
burak-ok: Yes, this is done recursively. For support see the `TestJSONFormatter` test which has a `parent.child1.grandchild1` test case and more

pkg/columns/formatter/json/json.go
burak-ok: For JSON it is an error. For the columns library it is not an error.

See the KubeIPResolver PR: It has a virtual column called `src` which displays something like `p/test-pod:8080`. But the Endpoint struct is also called `src` and its therefore its children are called `src.port: 8080`, etc...

pkg/columns/formatter/json/json_test.go
burak-ok: In theory yes - it should have the same structure.
The problem is with the order of the children. This is (currently) not ordered lexicographical. Therefore I don't know which children comes first and which comes last. And if it is different between runs or host systems

pkg/columns/formatter/json/json.go
```diff
@@ -0,0 +1,335 @@
...
+
+type column[T any] struct {
+	column    *columns.Column[T]
+	formatter func(*encodeState, *T)
+}
+
+type Formatter[T any] struct {
+	options *Options
+	columns []*column[T]
+	scratch [64]byte
```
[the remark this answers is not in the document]
burak-ok: There are usages for `scratch`. See same file the stringification for integers and float. Did you mean another field?

pkg/columns/formatter/json/json.go
```diff
@@ -0,0 +1,335 @@
...
+			}
+			first = false
+		}
+	}
+
+	// Now print columns with their parents
+	for key, val := range colsWithParent {
+		handlePrefix()
+
+		parentName, _ := json.Marshal(strings.Split(key, ".")[0])
```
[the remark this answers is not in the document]
burak-ok: Correct, not needed anymore. Thanks

pkg/columns/columns.go
burak-ok: Thanks, I also added a test for this where an extractor gets set

pkg/columns/columns.go
burak-ok: Do we actually need to copy it?
If not then something like 
```golang
fieldStart := unsafe.Add(entryStart, column.getOffset())
arrPtr := (*[1 << 30]OT)(fieldStart)
return (*arrPtr)[:l:l]
```
should be better since we are not copying anything

pkg/columns/formatter/json/json.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think the one used is https://github.com/inspektor-gadget/inspektor-gadget/blob/9b84dfe03a0029f95121ff976379af3ae0e2dd37/pkg/columns/formatter/json/buffer.go#L30. 

The code compiled fine after removing this.

pkg/columns/formatter/json/json.go
mauriciovasquezbernal: You're right. Then the current logic makes sense.

pkg/columns/columns.go
mauriciovasquezbernal: It should be fine to not copy it, but is it worth it? Perhaps the copy is not too expensive and we void future headaches?

flyth: I'm not sure about this, tbh. Afaiu it, generically it should be okay to do something like this (although `unsafe.Slice` would be preferred), however: we don't really know what the source of the memory is and whether it's safe after calling this function. For example, we could be accessing a source that does it's own garbage collection or frees the accessed memory block after some time by other means.

mauriciovasquezbernal (commented):
First 6 commits LGTM.

I'm still wondering about the last one, I made comments but I need to make another pass to check it out.

integration/helpers.go
```diff
@@ -182,6 +188,159 @@ func ExpectEntriesInMultipleArrayToMatch[T any](output string, normalize func(*T
...
+		}
+		ret = append(ret, entry)
+	}
+	return ret, nil
+}
+
+// TODO: Find a name for this. This is only a dev name
+func expectEntriesToMatchFoobar(t *testing.T, entries []map[string]interface{}, expectedEntries ...map[string]interface{}) error {
+out:
+	for _, expectedEntry := range expectedEntries { // Marshal & Unmarshal to have the right types in the map
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I tried this:

```golang
func expectEntriesToMatchFoobar(t *testing.T, entries []map[string]interface{}, expectedEntries ...map[string]interface{}) error {
	fixIntegers := func(m map[string]interface{}) {
		for k, v := range m {
			switch typed := v.(type) {
			case int, uint, int8, uint8, int16, uint16, int32, uint32, int64, uint64:
				m[k] = cast.ToFloat64(typed)
			}
		}
	}

out:
	for _, expectedEntry := range expectedEntries { // Marshal & Unmarshal to have the right types in the map
		fmt.Println("Searching for entry:", expectedEntry)
		for _, entry := range entries {

			fixIntegers(expectedEntry)
			fixIntegers(entry)

			if entry["fname"] == "/dev/null" {
				fmt.Println("actual entry:", entry)
			}
			if reflect.DeepEqual(expectedEntry, entry) {
				continue out
			}

			fmt.Printf("diff: %s\n", cmp.Diff(expectedEntry, entry))
		}
		expectedJsonEntry, _ := json.Marshal(expectedEntry)
		return fmt.Errorf("output doesn't contain the expected entry: %+v", string(expectedJsonEntry))
	}
	return nil
}
```

(cast is "github.com/spf13/cast")

Perhaps that helps to clean the code a bit?

integration/helpers.go
```diff
@@ -182,6 +188,159 @@ func ExpectEntriesInMultipleArrayToMatch[T any](output string, normalize func(*T
...
+
+func SetEventRuntimeContainerName(jsonObj map[string]interface{}, s string) {
+	runtimeMetadata := jsonObj["runtime"].(map[string]interface{})
+	if runtimeMetadata != nil {
+		runtimeMetadata["containerName"] = s
+	}
+}
+
+func SetEventK8sNode(jsonObj map[string]interface{}, s string) {
+	jsonObj["node"] = s
```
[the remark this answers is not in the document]
burak-ok: Correct. That was the rationale behind having this as an helper method. So we don't need to update every test, only these helper functions

integration/helpers.go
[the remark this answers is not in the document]
burak-ok: I also thought about casting everything to `float64` but then there are digit precision issues. `float64` roughly has 16 digit precision and especially for more digits there are issues with that.

So the problem would occur with large numbers in `int64`(max 19 digits) and `uint64` (max 20 digits)

integration/run_helpers.go
```diff
@@ -0,0 +1,182 @@
...
+
+func SetEventK8sContainer(jsonObj map[string]interface{}, s string) {
+	jsonObj["container"] = s
+}
+
+func SetEventK8sHostNetwork(jsonObj map[string]interface{}, b bool) {
+	jsonObj["hostnetwork"] = b
+}
+
+func RunEventToJsonObj(t *testing.T, ev *runtypes.Event) map[string]interface{} {
```
mauriciovasquezbernal, line 87: I have some doubts about this function's name because it only takes into consideration "hardcoded" columns and not the ones each gadget uses. In fact it just ignores (and removes) the `raw_data` and `data` fields. 
Can't we use the standard json marshaller for that?

mauriciovasquezbernal (commented):
Some more comments about the testing functions. In my opinion it's just a matter of cleaning them up a bit and then it's good to go.

integration/run_helpers.go
burak-ok: Unfortunately we can't use the standard json marshaller since we need to use the column names instead of the tagged json names

flyth:
> What's the plan for the YAML support? If possible I'd like to remove the `genericConverter` (as it uses a different layout for the data, takes a bumblebee dependency, etc.). I think we can translate JSON to YAML. I see there are different options to do that ->https://pkg.go.dev/github.com/ghodss/yaml#JSONToYAML, https://github.com/itchyny/json2yaml. WDYT?
> 
> (For sure native YAML support would be better, but IMO that's too much work at this point)
> 
> WDYT @flyth ?

I also think we should opt for a generic converter for now - IMHO speed is much more important for JSON than for yaml.

https://pkg.go.dev/github.com/ghodss/yaml#JSONToYAML == https://github.com/kubernetes-sigs/yaml which we already import (it says "kubernetes-sigs/yaml is a permanent fork of [ghodss/yaml](https://github.com/ghodss/yaml)."), so we could use just that.

mauriciovasquezbernal:
> What's the plan for the YAML support?

It turns out that using JSONToYAML is easier than I anticipated. Feel free to cherry-pick https://github.com/inspektor-gadget/inspektor-gadget/commit/a3c384f6652237721360f46035ababb6d87ebf71 so we implement that right away on this PR.

pkg/columns/formatter/json/json.go
```diff
@@ -0,0 +1,334 @@
...
+			// We already have an object with this key, skip this column
+			continue
+		}
+		handlePrefix()
+		col.formatter(buf, entry)
+	}
+}
+
+// FormatEntry returns an entry as a formatted string, respecting the given formatting settings
+func (f *Formatter[T]) FormatEntry(entry *T) string {
```
mauriciovasquezbernal, line 202: While experimenting with snapshotter support in containerized gadgets, I found out that a `FormatEntries` function will be useful. Let's keep going with this PR as it's but keep in mind that'd be needed very soon.

mauriciovasquezbernal (approved):
LGTM. Thanks for handling all the comments!

mauriciovasquezbernal:
There is only one question left, (for later on)

> Also, for the future, do we want to support omitempty?

pkg/columns/formatter/json/json.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	tf := &Formatter[T]{
+		options: opts,
+		columns: ncols,
+	}
+	return tf
+}
+
+func (f *Formatter[T]) printCols(entry *T, buf *encodeState, cols []*column[T], level int) {
```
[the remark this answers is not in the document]
burak-ok: > WDYT?

What I think? It is amazing :rocket:
Nice idea to offload the heavy parent "tracing" logic to the start of the program

Issue #1261, opened 2023-01-12
tests/snapshot/socket: Enable test once AKS supports Kubernetes 1.26


mauriciovasquezbernal:
We don't tun the on AKS with Ubuntu and amd64 because the kernel is too old to support tcp iterators:

https://github.com/inspektor-gadget/inspektor-gadget/blob/e68acfed636eee83904749c15d591d888bd31a81/integration/inspektor-gadget/snapshot_socket_test.go#L31-L33

AKS now uses Ubuntu 22.04 from Kubernetes 1.25 -> https://azure.microsoft.com/fr-fr/updates/generally-available-aks-support-for-ubuntu-2204/, then we could remove that conditional, however, by default, `az aks create` uses a N-1 minor version when creating the cluster as stated in the [documentation](https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#azure-portal-and-cli-versions).

To avoid hardcoding the Kubernetes version we use, we should remove this conditional once AKS supports Kubernetes 1.26 (and uses 1.25 by default). 

ref https://github.com/inspektor-gadget/inspektor-gadget/pull/1257

PR #1820, opened 2023-07-05
integration: Run snapshot socket test on AKS.

eiffel-fl:
AKS now runs on Kubernetes 1.25 which comes with kernel 5.15.0-1040-azure: $ kubectl describe node
Name:               aks-nodepool1-38109605-vmss000000
...
System Info:
...
  Kernel Version:             5.15.0-1040-azure
  OS Image:                   Ubuntu 22.04.2 LTS
  Operating System:           linux
  Architecture:               amd64
  Container Runtime Version:  containerd://1.7.1+azure-1
  Kubelet Version:            v1.25.6
  Kube-Proxy Version:         v1.25.6

### DIFF
--- a/integration/inspektor-gadget/snapshot_socket_test.go
+++ b/integration/inspektor-gadget/snapshot_socket_test.go
@@ -28,10 +28,6 @@ func TestSnapshotSocket(t *testing.T) {
 		t.Skip("Skip running snapshot socket gadget on ARO: iterators are not supported on kernel 4.18.0-305.19.1.el8_4.x86_64")
 	}
 
-	if *k8sDistro == K8sDistroAKSUbuntu && *k8sArch == "amd64" {
-		t.Skip("Skip running snapshot socket gadget on AKS Ubuntu amd64: iterators are not supported on kernel 5.4.0-1089-azure")
-	}
-
 	ns := GenerateTestNamespaceName("test-socket-collector")
 
 	t.Parallel()


mauriciovasquezbernal (approved):
LGTM from code inspection. Thanks!

mauriciovasquezbernal:
Fixes  https://github.com/inspektor-gadget/inspektor-gadget/issues/1261

mauriciovasquezbernal:
ARO has its own concurrent group https://github.com/inspektor-gadget/inspektor-gadget/blob/2da10c9b67234afc067016374d41f6ecb9806caf/.github/workflows/inspektor-gadget.yml#L1055 

Regarding AKS, I think the cluster name can collide https://github.com/inspektor-gadget/inspektor-gadget/blob/2da10c9b67234afc067016374d41f6ecb9806caf/.github/workflows/inspektor-gadget.yml#L996 
But, I think we shouldn't be running these tests twice, so that shouldn't be an issue. 

> This is not about concurrency-group but rather to not do tests on PR for citest* branches.

I think that's one way to fix that. When I enabled the `citest` branch I was thinking to only test a branch without opening a PR for it, but it totally makes sense to open a PR for the same branch. 

Another idea I have is to use a label on the PR (like `ci/run-all-tests`) to indicate to run all the tests, specially because we could want to run the tests on ARO / AKS after the PR has been opened, with the current approach it'll require to recreate the PR with a new base branch. But I never investigated this.

PR #1822, opened 2023-07-05
cmd: Check --help was given in RunE.

eiffel-fl:
Setting DisableFlagParsing caused issues where --help would not be taken into
    account:
    $ sudo ./ig profile block-io --help
    
    To fix this, we need to check that --help was given in RunE and call the
    corresponding Help().
    
    Fixes: 8511d3b25dd7 ("gadgets: Add custom formatting options")

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -216,6 +216,10 @@ func buildCommandFromGadget(
 			return cmd.ParseFlags(args)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if showHelp, _ := cmd.Flags().GetBool("help"); showHelp {
+				return cmd.Help()
+			}
+
 			if c, ok := gadgetDesc.(gadgets.GadgetDescCustomParser); ok {
 				var err error
 				parser, err = c.CustomParser(gadgetParams, cmd.Flags().Args())


flyth:
The intention behind setting DisableFlagParsing was to be able to add/modify flags at a later stage (iirc in this case it was to be able to update the available columns). For `--help` to still work, it has to be manually handled, though - I know this was implemented in an earlier PoC, but it must somehow have been removed/forgotten.

Re-adding this code inside `RunE` after all changes/additions have been made to the flags, should also update the help output:

```go
			if showHelp, _ := cmd.Flags().GetBool("help"); showHelp {
				return cmd.Help()
			}
```

/cc @mauriciovasquezbernal

flyth:
> I find the whole stuff overcomplicated but your solution does the trick, thank you.

Yeah, I agree - we sadly occasionally have a chicken+egg problem with the cobra library. For containerized gadgets we can only know all details at runtime (after initializing and parsing the containerized gadget's metadata) - and this hack was the only solution I found to make it work more dynamically.

cmd/common/registry.go
```diff
@@ -216,6 +216,10 @@ func buildCommandFromGadget(
 			return cmd.ParseFlags(args)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if showHelp, _ := cmd.Flags().GetBool("help"); showHelp {
```
flyth, line 219: Ah, I just re-checked: these lines actually already exist [below](https://github.com/inspektor-gadget/inspektor-gadget/pull/1822/files#diff-9ae11145e0ddd34f2787a98a7b403de21c7315c218ef54b66440bc8d9cbd2818R240) - they just have to be moved [after (but outside) the `if parser != nil {` section](https://github.com/inspektor-gadget/inspektor-gadget/pull/1822/files#diff-9ae11145e0ddd34f2787a98a7b403de21c7315c218ef54b66440bc8d9cbd2818R244).

Sorry, can't do full suggestion in here with the change.

flyth: [This](https://github.com/inspektor-gadget/inspektor-gadget/commit/3be19ad95552ae22805d6c7a8978663c51046b62) fixes it.

mauriciovasquezbernal: ops right, it must be outside the conditional. Thanks!

mauriciovasquezbernal:
I just realized `DisableFlagParsing` also breaks `--verbose` :(

flyth:
> > I just realized `DisableFlagParsing` also breaks `--verbose` :(
> 
> Yes, setting this flag actually broke few things and I would like to avoid us having to implement the same logic the library handles for us in `RunE()`.

Is there anything else you know of that broke? While I'd also prefer to not re-implement logic, we probably have to. But at least the registry is handling ~95% of all IG commands at one place... :)

cmd/common/registry.go
flyth: Sure, can do. Will fix the `--verbose` as well, then. I didn't want to hijack this, though.

Issue #1722, opened 2023-06-08
Design containerized gadgets support


mauriciovasquezbernal:
Create a design document for the containerized gadgets proposal in https://github.com/inspektor-gadget/inspektor-gadget/issues/1669.

Issue #718, opened 2022-06-29
bpf: There is not a defined style for BPF code 


mauriciovasquezbernal:
It's not clear which code style we show follow for BPF programs. Many of our tools are based on BCC gadgets (https://github.com/iovisor/bcc/tree/master/libbpf-tools) but it seems they also don't have defined a style. This leads to discussions like https://github.com/kinvolk/inspektor-gadget/pull/692#discussion_r895509531. 

The scope of this issue is to define which style to use (maybe the kernel one?), propose this to BCC (and hopefully update tools there) and implement a CI to check it here.

PR #1837, opened 2023-07-09
Changing trace/open gadget to send absolute file path with symlink resolution in the events

slashben:
# Trace/open gadget to resolve full file paths and symlinks

Open trace gadget has been using the file name from the user-space side in the system call. 

The open file name argument can be:
1. Relative path: in this case, the observability tool must be able to point to the full path that was open
2. Symlink: the observability tool doesn't need to resolve symlinks (since it can be done outside of its scope). However, this creates a burden on the event receiver side. Therefore it is better systemwide to implement it on the event source side. 

In this PR we are implementing a real file path resolution within the trace/open gadget which uses the file description and kernel side file structure to resolve the full file path including symlinks and relative paths.

## How to use

This implementation simply adds the resolution to the existing gadget, therefore everyone who uses this gadget will automatically receive this improvement.

## Testing done

Added unit tests to cover both cases.


pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -121,6 +123,12 @@ int trace_exit(struct trace_event_raw_sys_exit* ctx)
 	event.mntns_id = gadget_get_mntns_id();
 	event.timestamp = bpf_ktime_get_boot_ns();
 
+	// Attempting to extract the full file path with symlink resolution
+	if (0 <= ret)
+	{
+		read_full_path_of_open_file_fd(ret, (char*)event.fname, sizeof(event.fname));
```
alban, line 129: I agree with @eiffel-fl, we should use a different field for the given path and the resolved path.

pkg/gadgets/trace/open/types/types.go could have both `Path` and `FullPath`. One of them could have the column attribute "hide" if it is too long to print on one line.

slashben:
> I am nonetheless wondering if we should not keep both information, _i.e._ the path which was opened, _e.g._ a relative one, and the underlying corresponding path, _e.g._ the absolute path corresponding.

@eiffel-fl great comment and I have to tell you that I have been thinking about it. These were the options I saw:
1. The gadget adds `cwd` to the event. This enables the event user to calculate the absolute path using the `filename` and `cwd`.
2. Resolve the full path (as in the current version) on the Kernel side and use it to the event as `filename`
3. Send both the `filename` from the system call and the full resolved path (as in option 2)

These were the factors for my decision:
* Sending two path-type strings in a single event inflates the event record. In general, paths are limited by PATH_MAX (4k) and in our case 256 bytes. AFAIK we have a hard time creating events larger than 512, no? This means that options 1 and 3 are creating a load on the ring buffer and the receiver side.
* `cwd` is not a very useful part of the event record, the only reason to have it is to reconstruct the absolute path. If I want to track the `cwd` I will follow the corresponding system call since the applications are not changing it so many times. Sending it in every event is suboptimal from a memory footprint perspective.
* Option number 1 does not resolve symbolic links. It is a big minus.
* Sending the original system call `filename` value helps to give a full picture of application behavior, and option 2 does not cover it.

Maybe it is due to my specific application needs, but I think that keeping the event pipeline memory friendly and the need for symbolic link resolution is more important than the other factors.

WDYT?

pkg/gadgets/common/filesystem.h
```diff
@@ -162,4 +162,44 @@ static __always_inline void *get_path_str(struct path *path)
...
+    if (task == NULL) {
+        return NULL;
+    }
+
+	// extact the file vector from the task_struct
+	struct file **fd = BPF_CORE_READ(task,files,fdt,fd);
+	
+	// extract the file pointer from the file vector
+	struct file *f;
+	bpf_core_read((void *) &f, sizeof(f), &fd[fd_num]);
```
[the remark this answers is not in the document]
slashben: Agreed, changing it.

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
slashben: @alban , see my comments below. Don't you think it will create pressure on the ring buffer and event structures? This would effectively double the size of an event record. Besides I am not sure about how useful the `Path` would be in this case.

alban:
@slashben Thanks for explaining your considerations. Some additional points:

- I think it is better to resolve the paths in eBPF rather than userspace because:
  - By the time the event reaches userspace, the filesystem might have changed and the path resolution might be different. Even though it is not race-free in eBPF, it is still more likely to be correct.
  - We are trying to add support for containerized gadgets (#1722). If we reimplement trace-open as a containerized gadget, it's easier if most of the features are implemented in eBPF
- The limit to 512 bytes can be solved by allocating the event struct in a hash map indexed by the pid instead of the stack. There is an example of code pattern in [trace_exec.bpf.c#L90-L95](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_exec.bpf.c#L90-L95).
- If size of the event struct is a concern, we could have one field with a size limit such as 256 or more and add the two strings one after another. There is an example of this with args of execve in [execsnoop.h#L28](https://github.com/inspektor-gadget/inspektor-gadget/blob/3992c6dc120aaa0969c7f4e8bb35e1c095bee9f2/pkg/gadgets/trace/exec/tracer/bpf/execsnoop.h#L28) and [execsnoop.bpf.c#L110-L111](https://github.com/inspektor-gadget/inspektor-gadget/blob/3992c6dc120aaa0969c7f4e8bb35e1c095bee9f2/pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c#L110-L111)
- If memory or cpu performance becomes a problem, it is possible to enable/disable features in eBPF with parameters and `spec.RewriteConstants`. Example in [tracer.go#L210-L222](https://github.com/inspektor-gadget/inspektor-gadget/blob/3992c6dc120aaa0969c7f4e8bb35e1c095bee9f2/pkg/gadgets/profile/tcprtt/tracer/tracer.go#L210-L222) and [tcprtt.bpf.c#L18-L26](https://github.com/inspektor-gadget/inspektor-gadget/blob/3992c6dc120aaa0969c7f4e8bb35e1c095bee9f2/pkg/gadgets/profile/tcprtt/tracer/bpf/tcprtt.bpf.c#L18-L26)
  - We've got a user request for cwd for trace-exec (#1700) but not for trace-open yet. I wonder if we could get some numbers showing how much the performance is impacted or not by those additional fields.

pkg/gadgets/common/filesystem.h
```diff
@@ -0,0 +1,165 @@
...
+    unsigned int off;
+    int sz;
+
+    // Get per-cpu string buffer
+    buf_t *string_p = get_buf(STRING_BUF_IDX);
+    if (string_p == NULL)
+        return NULL;
+
+#pragma unroll
+    for (int i = 0; i < MAX_PATH_COMPONENTS; i++) {
```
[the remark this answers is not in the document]
slashben: Same here, I don't claim that I understand the code completely, but the structure of the `for` loop is looking for path components, not character level parsing.

slashben:
@alban @eiffel-fl , I have changed the gadget to send the original path in the `Path` field (as before) and use `FullPath` for the resolved path. 
This is what you had in mind?

pkg/gadgets/trace/open/tracer/tracer_test.go
```diff
@@ -69,6 +70,26 @@ func createTracer(
...
+	if !filepath.IsAbs(fileName) {
+		return fileName
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return "" // This should never happen
+	}
+
+	// Count the number of components in the path
```
alban, line 84: Comment to be updated

pkg/gadgets/trace/open/tracer/tracer_test.go
```diff
@@ -174,13 +197,97 @@ func TestOpenTracer(t *testing.T) {
...
+		},
+		"test_symbolic_links": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *tracer.Config {
+				return &tracer.Config{
+					MountnsMap: utilstest.CreateMntNsFilterMap(t, info.MountNsID),
+				}
+			},
+			generateEvent: func() (int, error) {
+				// Create a symbolic link to /dev/null
+				err := os.Symlink("/dev/null", "/tmp/test")
```
alban, line 257: Tests are executed in parallel (`t.Parallel()` line 364) and the path `/tmp/test` is reused by several tests to create a file, create a symlink, or delete it. We need to use different paths in different tests to avoid race conditions.

This problem was not introduced in this PR, but I just noticed it.

alban (commented):
This looks good! Just a couple of comments in the tests.

I didn't review pkg/gadgets/common/filesystem.h. Waiting from @mauriciovasquezbernal about it.

pkg/gadgets/trace/open/tracer/tracer_test.go
slashben: done

pkg/gadgets/trace/open/tracer/tracer_test.go
```diff
@@ -173,14 +195,98 @@ func TestOpenTracer(t *testing.T) {
...
+					},
+					WithMountNsID: eventtypes.WithMountNsID{MountNsID: info.MountNsID},
+					Pid:           uint32(info.Pid),
+					Uid:           uint32(info.Uid),
+					Comm:          info.Comm,
+					Fd:            fd,
+					Ret:           fd,
+					Err:           0,
+					Path:          "/tmp/test_symbolic_links",
+					FullPath:      "/dev/null",
```
mauriciovasquezbernal, line 284: Nice!

mauriciovasquezbernal (commented):
Some initial comments. Will do a deeper pass soon.

pkg/gadgets/common/filesystem.h
```diff
@@ -0,0 +1,208 @@
...
+    if (task == NULL) {
+        return NULL;
+    }
+
+	// extact the file vector from the task_struct
+	struct file **fd = BPF_CORE_READ(task,files,fdt,fd);
+	
+	// extract the file pointer from the file vector
+	struct file *f = NULL;
+    uint max_fds = BPF_CORE_READ(task,files,fdt,max_fds);
```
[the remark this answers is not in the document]
mauriciovasquezbernal: This applies to few other places as well.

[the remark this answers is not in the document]
slashben: @mauriciovasquezbernal do you have a [built-in Clang formatting style](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#basedonstyle) you are using?

pkg/gadgets/trace/open/tracer/bpf/opensnoop.h
```diff
@@ -24,6 +24,7 @@ struct event {
 	__u16 mode;
 	__u8 comm[TASK_COMM_LEN];
 	__u8 fname[NAME_MAX];
+	__u8 full_fname[NAME_MAX];
```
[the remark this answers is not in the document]
slashben: I think I understand what you mean in general, but I am not sure about the practice. Do you mean that if the last member of the C struct is the `full_name`, since it is a null-terminated string it can be truncated in the ring-buffer? Is this working with giving in the last argument `bpf_perf_event_output` less than `sizeof(struct event)`?

pkg/gadgets/common/filesystem.h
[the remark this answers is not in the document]
mauriciovasquezbernal: Unfortunately we haven't defined it yet -> https://github.com/inspektor-gadget/inspektor-gadget/issues/718. Now that I think better, you can just ignore this comment given that we don't have a style defined and requiring it is a non-sense.

pkg/gadgets/trace/open/tracer/bpf/opensnoop.h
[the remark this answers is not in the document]
mauriciovasquezbernal: The last argument to `bpf_perf_event_output` should be a variable that takes the return value of `bpf_probe_read_user_str` into consideration. The [trace exec](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c) gadget has a more complex example of this as in that case multiple strings are saved at the end.

pkg/gadgets/common/filesystem.h
```diff
@@ -0,0 +1,209 @@
...
+    }
+    
+    struct task_struct *task = (struct task_struct *) bpf_get_current_task();
+    if (task == NULL) {
+        return NULL;
+    }
+
+	// extract the file vector from the task_struct
+	struct file **fd = BPF_CORE_READ(task, files, fdt, fd);
+
```
mauriciovasquezbernal, line 180: nit: please remove trailing white spaces.

pkg/gadgets/common/filesystem.h
```diff
@@ -0,0 +1,209 @@
...
+{
+	struct file *file = get_struct_file_for_fd(fd_num);
+	if (file == NULL) {
+		return 1;
+	}
+
+	struct path f_path = BPF_CORE_READ(file, f_path);
+
+	// Extract the full path string
+	char* c_path = get_path_str(__builtin_preserve_access_index(&f_path));
```
mauriciovasquezbernal, line 201: We use vmlinux.h that marks all structs with [preserve_access_index](https://github.com/alban/inspektor-gadget/blob/d135fa553e5a6957b9aee2274762db55b36db517/pkg/amd64/vmlinux/vmlinux.h#L5), so we don't need to make this manually. We can also use `get_path_str(&file->f_path)` directly to avoid creating the temp variable.

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -22,7 +23,7 @@ struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 10240);
 	__type(key, u32);
-	__type(value, struct args_t);
+	__type(value, struct event);
```
mauriciovasquezbernal, line 26: Remove `struct args_t` from opensnoop.h since it's not needed anymore after this change.

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -95,35 +105,38 @@ int ig_openat_e(struct trace_event_raw_sys_enter* ctx)
...
+	event->pid = bpf_get_current_pid_tgid() >> 32;
+	event->uid = (u32) uid_gid;
+	event->gid = (u32) (uid_gid >> 32);
+	bpf_get_current_comm(&event->comm, sizeof(event->comm));
+	event->ret = ret;
+	event->mntns_id = gadget_get_mntns_id();
+	event->timestamp = bpf_ktime_get_boot_ns();
+
+	// Attempting to extract the full file path with symlink resolution
+	if (0 <= ret)
```
mauriciovasquezbernal, line 132: I know some people prefer to have the constant first in a comparison, but all our code uses the other way around. Please swap them.

mauriciovasquezbernal (commented):
It's looking really good. Some minor comments around.

I'm a bit concerned about the performance impact of this, I think we should disable it by default. (@alban do you have comments on this?) You can get some inspiration about how to make it from https://github.com/inspektor-gadget/inspektor-gadget/pull/1700. But basically you have to:
- Introduce a new parameter (gadget.go). I'd suggest `full-path`
- Introduce a new global constant on the eBPF program and only get that field when it's true.
- Set the constant to true when the above parameter is true
- Enable the feature in integration tests.

Please also put a sentence in docs/gadgets/trace/open.md describing this new field (and how to enable it with the parameter introduced above).

Once that's complete, I'll approve. Thanks!

pkg/gadgets/trace/open/tracer/bpf/opensnoop.h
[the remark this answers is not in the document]
slashben: @mauriciovasquezbernal check out the code I have pushed, I have changed `full_fname` to be at PATH_MAX size and also putting variable size (depending on the string size) on the ring buffer.

pkg/gadgets/common/filesystem.h
slashben: done

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
slashben: OK

pkg/gadgets/common/filesystem.h
slashben: gotcha

mauriciovasquezbernal:
@slashben it looks good! Please organize the commits then I'll make a final pass.

slashben:
@mauriciovasquezbernal I have reorganized the commits

slashben:
```       cols := types.GetColumns()
          col, ok := cols.GetColumn("fullPath")
          if ok {
               col.GetAttributes().Visible = true
          }
```
@eiffel-fl  
Shouldn't this be conditional based on the `t.config.FullPath` flag?

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -95,35 +106,50 @@ int ig_openat_e(struct trace_event_raw_sys_enter* ctx)
...
+	} else {
+		// If the open failed, we can't get the full path
+		event->full_fname[0] = '\0';
+		full_fname_len = 1;
+	}
 
 	/* emit event */
 	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
-			      &event, sizeof(event));
+			      event, sizeof(struct event) - (PATH_MAX - full_fname_len));
```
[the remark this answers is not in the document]
slashben: Credits to @mauriciovasquezbernal

pkg/gadgets/common/filesystem.h
```diff
@@ -0,0 +1,208 @@
...
+
+
+#define BPF_PERCPU_ARRAY(_name, _value_type, _max_entries)                                         \
+	BPF_MAP(_name, BPF_MAP_TYPE_PERCPU_ARRAY, u32, _value_type, _max_entries)
+
+BPF_PERCPU_ARRAY(bufs, buf_t, MAX_BUFFERS);                        // percpu global buffer variables
+
+// undef as we don't want to use this in our gadgets. yet?
+#undef BPF_MAP
+#undef BPF_PERCPU_ARRAY
```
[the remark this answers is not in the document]
slashben: See the global variable declaration in line 49, it is in use. Since this is not my code, I'd prefer not to touch it :(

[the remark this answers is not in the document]
mauriciovasquezbernal: Yes, it can't be removed.

mauriciovasquezbernal:
> I just would like to have the `fullPath` colum being printed when using the corresponding flag:

This can be done by adding `columnTags:"param:full-path` in pkg/gadgets/trace/open/types/types.go. 

```
FullPath string      `json:"fullPath,omitempty" column:"fullPath,minWidth:24,width:32" columnTags:"param:full-path"`
```

```bash
$ go run -exec "sudo -E" ./cmd/ig/. trace open --full-path
CONTAINER                               PID        COMM             FD   ERR PATH                                     FULLPATH                                
minikube-containerd                     12139      kubelet          25   0   /proc/290884/net/dev                     /proc/290884/net/dev                    
minikube-containerd                     12139      kubelet          25   0   /sys/fs/cgroup/cpu,cpuacct/system.slice… /sys/fs/cgroup/cpu,cpuacct/system.slice…
minikube-containerd                     12139      kubelet          25   0   /proc/290884/fd                          /proc/290884/fd                         
minikube-containerd                     12139      kubelet          25   0   /proc/290884/limits                      /proc/290884/limits                     
minikube-containerd                     12139      kubelet          25   0   /sys/fs/cgroup/cpu,cpuacct/kubepods/bes… /sys/fs/cgroup/cpu,cpuacct/kubepods/bes…
minikube-containerd                     12139      kubelet          32   0   /sys/fs/cgroup/cpu,cpuacct/kubepods/bes… /sys/fs/cgroup/cpu,cpuacct/kubepods/bes…
minikube-containerd                     12139      kubelet          25   0   /sys/fs/cgroup/cpu,cpuacct/kubepods/bes… /sys/fs/cgroup/cpu,cpuacct/kubepods/bes…
minikube-containerd                     12139      kubelet          25   0   /proc/1310/net/dev                       /proc/1310/net/dev                      
minikube-containerd                     12139      kubelet          25   0   /sys/fs/cgroup/cpu,cpuacct/kubepods/bes… /sys/fs/cgroup/cpu,cpuacct/kubepods/bes…
minikube-containerd                     12139      kubelet          25   0   /proc/1310/fd                            /proc/1310/fd                           
minikube-containerd                     12139      kubelet          25   0   /proc/1310/limits                        /proc/1310/limits                       
minikube-containerd                     12139      kubelet          25   0   /sys/fs/cgroup/cpu,cpuacct/kubepods/bes… /sys/fs/cgroup/cpu,cpuacct/kubepods/bes…
^Cmvb@mvb-desktop:~/kinvolk/ebpf/inspektor-gadget$ 
```

mauriciovasquezbernal (approved):
LGTM. Thanks @slashben for your contribution. 

I added `columnTags:"param:full-path"` to types.go, fixed a commit message and pushed to your branch. I'll merge once CI is green.

alban:
FYI, this PR made the `trace open` gadget 5x slower to start, even when `--full-path` is not given: about 1 second instead of 176ms. See comment generated by the bot at the bottom of the commit:
https://github.com/inspektor-gadget/inspektor-gadget/commit/204c97f7a72661c5fa08f07a4e4730244b6615d8#commitcomment-121440900

image

[source of the graph](https://inspektor-gadget.github.io/ig-benchmarks/dev/bench/index.html)

If someone is interested in profiling it, there is documentation how to do that step by step:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/CONTRIBUTING.md#benchmarks

My guess is it's the ebpf verifier (in kernel) and the cilium/ebpf BTF relocations (in userspace) and there is not much we can do about it.

PR #1839, opened 2023-07-10
cmd/common/registry: fix `--help` for gadgets without parser, fix `--verbose`

flyth:
There was a bug, that prevented gadgets that don't use a parser, from outputting help when requested. This fixes that. This also fixes `--verbose` not working properly on gadgets anymore by manually checking the flag at a later stage.

Reason for the changed behavior is that we delayed flag parsing, so that we can handle them more dynamically and still have proper `--help` output.


mauriciovasquezbernal (approved):
Works for me too. Thanks for cleaning my mess.

PR #1843, opened 2023-07-11
Fix Inspektor Gadget environment variables for container runtime

eiffel-fl:
Hi.

This PR fixes the `INSPEKTOR_GADGET_*SOCKETPATH`.

Best regards.

### DIFF
--- a/Dockerfiles/ig.Dockerfile
+++ b/Dockerfiles/ig.Dockerfile
@@ -39,4 +39,8 @@ LABEL org.opencontainers.image.licenses=Apache-2.0
 
 COPY --from=builder /go/src/github.com/inspektor-gadget/inspektor-gadget/ig-${TARGETOS}-${TARGETARCH} /usr/bin/ig
 ENV HOST_ROOT=/host
+ENV INSPEKTOR_GADGET_DOCKER_SOCKETPATH=/host/run/docker.sock
+ENV INSPEKTOR_GADGET_CONTAINERD_SOCKETPATH=/host/run/containerd/containerd.sock
+ENV INSPEKTOR_GADGET_CRIO_SOCKETPATH=/host/run/crio/crio.sock
+ENV INSPEKTOR_GADGET_PODMAN_SOCKETPATH=/host/run/podman/podman.sock
 ENTRYPOINT ["/usr/bin/ig"]
--- a/cmd/common/utils/flags.go
+++ b/cmd/common/utils/flags.go
@@ -31,6 +31,11 @@ import (
 const (
 	OutputModeJSON          = "json"
 	OutputModeCustomColumns = "custom-columns"
+
+	DockerSocketPathFlag     = "docker-socketpath"
+	ContainerdSocketPathFlag = "containerd-socketpath"
+	CrioSocketPathFlag       = "crio-socketpath"
+	PodmanSocketPathFlag     = "podman-socketpath"
 )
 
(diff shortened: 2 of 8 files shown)


alban (commented):
Thanks for working on that.

Usually, CLI tools have this precedence for configuration:
1. CLI flags: highest priority
2. environment variable: next priority
3. configuration file: low priority

So I would like `--docker-socketpath` to take precedence over `INSPEKTOR_GADGET_DOCKER_SOCKETPATH`. But if the user does not specify `--docker-socketpath` on the command line, it should take the environment variable, even if the `--docker-socketpath` flag has a default value.

The issue here is that the environment variables were never taken into account because the variables for the CLI flags were never empty due to them having a default value.

I am not sure how to organise the code for this. Any ideas?

alban:
> I think this is a matter of comparing with `runtimeclient.*DefaultSocketPath` rather than with empty string. What do you think of it?

That would mostly work except when the user explicitely specify `--docker-socketpath=/run/docker.sock`, which happens to be the default path.

cobra has a function to check if the flag was explicitly set by the user or not:
https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed

But I don't know if it is easy to use and if it is worth it.

alban:
> Problem is we do not use cobra directly here but the Params:

I think that [RuntimeConfig struct](https://github.com/inspektor-gadget/inspektor-gadget/blob/4de8c0167d08a987da06914dfe45a08f77860a0d/pkg/container-utils/containerutils.go#L44) should contains a pointer to a string instead of the string directly so we can distinguish whether it is set or not (nil or not):
```
type RuntimeConfig struct {
	Name       string
	SocketPath *string
}
```

The `RuntimeConfig struct` can be initialized by calling [FlagSet.Changed](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed). Then [NewContainerRuntimeClient](https://github.com/inspektor-gadget/inspektor-gadget/blob/4de8c0167d08a987da06914dfe45a08f77860a0d/pkg/container-utils/containerutils.go#L49C6-L49C31) would test if SocketPath is nil.

mqasimsarfraz:
@eiffel-fl I see a overlap between this PR and https://github.com/inspektor-gadget/inspektor-gadget/pull/2423. Or is this fixing a broader issue compared to https://github.com/inspektor-gadget/inspektor-gadget/pull/2423 focused on `ig` only?

flyth:
> @flyth Any opinion on this, particularly with regard to #2512?

The refactoring doesn't change too much about how the [parameters are bound to cobra](https://github.com/inspektor-gadget/inspektor-gadget/blob/fd007532d77024293be1c5fd3b58e3f11709a95c/cmd/common/registry.go#L786) (so this would help anyway), but IMHO that would be where binding to env vars also should happen (instead of inside the operator for example - I think all cobra/viper/env related code should be inside `cmd`).

For ig-k8s it's a bit more tricky, because the global part of operators doesn't go through the `registry.go`, but maybe the `AddFlags()` can still be reused to work with the `params` pkg (I haven't looked deeply into this.)

Anyway, I think we should be using [spf13/viper](https://github.com/spf13/viper) to add support for configuration files - which [more or less automatically](https://github.com/spf13/viper?tab=readme-ov-file#working-with-environment-variables) brings support for the matching env variables.

#2423 seems to go a bit more in that direction, but seems also not yet complete.

PR #1850, opened 2023-07-13
container-utils: Generalize testutils pkg and test runtime-client interface

blanquicet:
# container-utils: Generalize testutils pkg and test runtime-client interface

The container-utils/testutils package was generalized (new `Container` interface defined) to be easily used directly for any tests involving the creation of containers using Docker or Containerd (Both use the API).

The new `Container` interface was used to create tests to verify the runtime-client interface methods:

```bash
sudo go test -v -run TestRuntimeClientInterface ./pkg/container-utils/runtime-client...
```

NOTE: The same `Container` interface could be used to test other modules like the container-hook (some experiments [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/43692ae9e69240d3dea663a67131b18c2cb04944)). cc @alban

### DIFF
--- a/integration/container_interface.go
+++ b/integration/container_interface.go
@@ -47,11 +47,13 @@ func NewContainerFactory(containerRuntime string) (ContainerFactory, error) {
 }
 
 type containerSpec struct {
-	name         string
-	cmd          string
+	// Options
 	options      []testutils.Option
 	cleanup      bool
 	startAndStop bool
+
+	// Internal
+	started bool
 }
 
 // containerOption is a function that modifies a ContainerSpec and exposes only
--- a/integration/container_interface.go
+++ b/integration/container_interface.go
@@ -46,40 +46,36 @@ func NewContainerFactory(containerRuntime string) (ContainerFactory, error) {
 	}
 }
 
-type containerSpec struct {
-	// Options
+type cOptions struct {
 	options      []testutils.Option
 	cleanup      bool
 	startAndStop bool
-
-	// Internal
-	started bool
 }
 
 // containerOption is a function that modifies a ContainerSpec and exposes only
 // few options from testutils.Option to the user.
-type containerOption func(specs *containerSpec)
+type containerOption func(opts *cOptions)
 
 func WithContainerImage(image string) containerOption {
-	return func(specs *containerSpec) {
-		specs.options = append(specs.options, testutils.WithImage(image))
+	return func(opts *cOptions) {
+		opts.options = append(opts.options, testutils.WithImage(image))
 	}
 }
 
 func WithContainerSeccompProfile(profile string) containerOption {
-	return func(specs *containerSpec) {
-		specs.options = append(specs.options, testutils.WithSeccompProfile(profile))
+	return func(opts *cOptions) {
+		opts.options = append(opts.options, testutils.WithSeccompProfile(profile))
 	}
 }
 
 func WithCleanup() containerOption {
-	return func(specs *containerSpec) {
-		specs.cleanup = true
+	return func(opts *cOptions) {
+		opts.cleanup = true
 	}
 }
 
 func WithStartAndStop() containerOption {
-	return func(specs *containerSpec) {
-		specs.startAndStop = true
+	return func(opts *cOptions) {
+		opts.startAndStop = true
 	}
 }
--- a/integration/container_interface.go
+++ b/integration/container_interface.go
@@ -23,10 +23,10 @@ import (
 )
 
 type ContainerFactory interface {
-	NewContainer(name, cmd string, opts ...containerOption) ContainerInterface
+	NewContainer(name, cmd string, opts ...containerOption) IntegrationTestsContainer
 }
 
-type ContainerInterface interface {
+type IntegrationTestsContainer interface {
 	Run(t *testing.T)
 	Start(t *testing.T)
 	Stop(t *testing.T)
--- a/integration/container_interface.go
+++ b/integration/container_interface.go
@@ -22,16 +22,8 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
-type ContainerSpec struct {
-	Name         string
-	Cmd          string
-	Options      []containerOption
-	Cleanup      bool
-	StartAndStop bool
-}
-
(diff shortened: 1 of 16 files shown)


pkg/container-utils/runtime-client/runtimeclient_test.go
```diff
@@ -0,0 +1,165 @@
...
+			// Test runtime client methods
+			t.Run("GetContainers", func(t *testing.T) {
+				t.Parallel()
+
+				containers, err := rc.GetContainers()
+				require.Nil(t, err)
+				require.NotNil(t, containers)
+
+				for _, eData := range expectedData {
+					var found bool
```
burak-ok, line 99: nit: default is `false`, but I prefer to have `found := false` instead

pkg/container-utils/testutils/containerd.go
```diff
@@ -34,34 +35,71 @@ const (
...
+	return c.id
+}
+
+func (c *ContainerdContainer) Pid() int {
+	return c.pid
+}
+
+func (c *ContainerdContainer) initClientAndCtx() error {
+	var err error
+	c.client, err = containerd.New("/run/containerd/containerd.sock",
```
burak-ok, line 74: Use the default value from the following file? https://github.com/inspektor-gadget/inspektor-gadget/blob/03c29258d42e7d13126b7a83da8f89ce4e8d030e/pkg/container-utils/runtime-client/interface.go#L29

pkg/container-utils/testutils/containerd.go
```diff
@@ -71,115 +109,170 @@ func RunContainerdContainer(ctx context.Context, t *testing.T, command string, o
...
+	// We need to wait until the task is killed before trying to delete it
+	if c.options.forceDelete {
+		// If we are forcing the deletion, we don't want to wait forever as the
+		// task might be already stopped
+		select {
+		case <-c.exitStatus:
+		case <-time.After(1 * time.Second):
+		}
+	} else {
+		<-c.exitStatus
```
burak-ok, line 229: Wouldn't the same issue you described 5 lines prior happen here too?

burak-ok (commented):
Did only a single pass reading the commits and wanted to post the initial comments

pkg/container-utils/testutils/containerd.go
blanquicet: You're right. It never happened to me, but it could happen

integration/container_interface.go
```diff
@@ -22,16 +22,8 @@ import (
...
-	Cleanup      bool
-	StartAndStop bool
-}
-
 type ContainerFactory interface {
-	NewContainer(ContainerSpec) ContainerInterface
+	NewContainer(name, cmd string, opts ...containerOption) ContainerInterface
 }
 
 type ContainerInterface interface {
```
burak-ok, line 29:
```suggestion
type ContainerInterface interface {
	testutils.Container
	IsCleanup() bool
	IsStartAndStop() bool
}
```
Do you think this makes sense? We don't need some methods of `testutils.Container` but they way we created them they already have it.
Or maybe even combining the interfaces alltogether and make them one

pkg/container-utils/runtime-client/runtimeclient_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+					require.True(t, cmp.Equal(cData, eData),
+						"unexpected container data:\n%s", cmp.Diff(cData, eData))
+				}
+			})
+		})
+	}
+}
+
+// TestDeleteContainers is useful to delete containers from
```
burak-ok, line 144: Is this a TODO or won't test

pkg/container-utils/testutils/containerd.go
burak-ok: I see.
Then I would additionally create a very simple and small unittest which checks if the default socket path is still the one you set here manually. If it changes we can see the failed test directly and know what happened.

If we don't have such a test and a hardcoded value all tests (might) fail and we have to "search" for the reason.

WDYT?

burak-ok (approved):
Looks good to me generally

pkg/container-utils/testutils/containerd.go
blanquicet: IMO, the output is clear enough: 

```
=== CONT  TestRuntimeClientInterface/containerd
    containerd.go:88: Failed to initialize client: creating a client: failed to dial "/invalid/path/to/containerd.sock": context deadline exceeded: connection error: desc = "transport: error while dialing: dial unix:///invalid/path/to/containerd.sock: timeout"
```

WDYT?

pkg/container-utils/runtime-client/runtimeclient_test.go
blanquicet: I added it because during the development of the previous test, depending on where it failed, sometimes the environment remained dirty, so I used this function to enforce a cleanup. However, it is not an actual test. Maybe I should remove it?

integration/container_interface.go
burak-ok: I don't know either what to do :D
I just got confused while reviewing `type ContainerInterface interface` and `type Container interface` and that was just a quick thought

pkg/container-utils/runtime-client/runtimeclient_test.go
burak-ok: Ah, I see. Maybe this could then be even part of a cleanup function if a container is not deleted correctly (and then the test fails)

pkg/container-utils/testutils/containerd.go
burak-ok: Seems good enough :+1:

pkg/container-utils/runtime-client/runtimeclient_test.go
blanquicet: It is already part of the cleanup function of the main test: `TestRuntimeClientInterface`, but during development, sometimes the test crashes and such function wasn't called. At that point, I executed this code ("test") to clean up all the remaining containers.

integration/container_interface.go
blanquicet: I renamed it to `IntegrationTestsContainer`.

Issue #1832, opened 2023-07-06
TestClose fails when other Docker containers run on the machine


alban:
## Description

The unit test `TestClose` fails when other Docker containers run on the machine

## Impact

File descriptor leak: `ig` and `gadgettracermanager` have two file descriptors never closed for each initial container.

## Environment and steps to reproduce

Start a docker container in the background:
```
docker run -d -p 5000:5000 --restart=always --name registry registry:2
```
Run the TestClose test:
```
go test -test.v -exec sudo -run TestClose ./pkg/ig-manager/...
```
See the error:
```
=== RUN   TestClose
    ig-manager_test.go:119: After 20 attempts, fd leaked:
        0: /dev/null
        1: pipe:[2658229]
        2: pipe:[2658229]
        4: anon_inode:[eventpoll]
        5: pipe:[2665117]
        6: pipe:[2665117]
        
        0: /dev/null
        1: pipe:[2658229]
        2: pipe:[2658229]
        24: mnt:[4026535441]
        25: net:[4026535445]
        26: mnt:[4026535441]
        27: net:[4026535445]
        28: mnt:[4026533836]
        29: net:[4026534233]
        30: mnt:[4026535441]
        31: net:[4026535445]
        32: mnt:[4026534944]
        33: net:[4026534948]
        34: mnt:[4026533836]
        35: net:[4026534233]
        36: mnt:[4026533274]
        37: net:[4026533278]
        38: mnt:[4026535441]
        39: net:[4026535445]
        4: anon_inode:[eventpoll]
        40: mnt:[4026534944]
        41: net:[4026534948]
        42: mnt:[4026533836]
        43: net:[4026534233]
        44: mnt:[4026533274]
        45: net:[4026533278]
        48: mnt:[4026534944]
        49: net:[4026534948]
        5: pipe:[2665117]
        50: mnt:[4026533836]
        51: net:[4026534233]
        52: mnt:[4026533274]
        53: net:[4026533278]
        56: mnt:[4026534944]
        57: net:[4026534948]
        6: pipe:[2665117]
        60: mnt:[4026533274]
        61: net:[4026533278]
--- FAIL: TestClose (6.30s)
FAIL
FAIL	github.com/inspektor-gadget/inspektor-gadget/pkg/ig-manager	6.327s
FAIL
```

## Expected behavior

The test should not fail even if the machine has some containers running unrelated to the tests.

## Additional information

[...]

PR #1854, opened 2023-07-17
container-collection: fix fd leak for initial containers

alban:
The enrichers were called twice for initial containers.

This was especially problematic for the WithTracerCollection enricher because it opens file descriptors and they were never closed.

This patch ensures the enricher were called only one time.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1832

## Testing done

```
go test -test.v -exec sudo -run TestClose ./pkg/ig-manager/...
```

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -819,6 +819,9 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
 					container.Runtime.ContainerID, err)
 				return false
 			}
+			if container.mntNsFd != 0 {
+				log.Warnf("WithTracerCollection: mntns reference already set for container %s", container.Runtime.ContainerID)
+			}
 			container.mntNsFd = mntNsFd
 
 			netNsPath := filepath.Join(host.HostProcFs, fmt.Sprint(container.Pid), "ns", "net")
@@ -827,6 +830,9 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
 				log.Warnf("WithTracerCollection: failed to open netns reference for container %s: %s",
 					container.Runtime.ContainerID, err)
 				return false
+			}
+			if container.netNsFd != 0 {
+				log.Warnf("WithTracerCollection: netns reference already set for container %s", container.Runtime.ContainerID)
 			}
 			container.netNsFd = netNsFd
 			return true
--- a/pkg/ig-manager/ig-manager_test.go
+++ b/pkg/ig-manager/ig-manager_test.go
@@ -127,6 +127,17 @@ func checkFdList(t *testing.T, initialFdList string, attempts int, sleep time.Du
 func TestClose(t *testing.T) {
 	utilstest.RequireRoot(t)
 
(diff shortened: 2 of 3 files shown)


pkg/ig-manager/ig-manager_test.go
```diff
@@ -127,6 +127,17 @@ func checkFdList(t *testing.T, initialFdList string, attempts int, sleep time.Du
...
+	opts := []testutils.Option{
+		testutils.WithName("test-ig-close"),
+		testutils.WithoutRemoval(),
+		testutils.WithoutWait(),
+		testutils.WithoutLogs(),
+	}
+	testutils.RunDockerContainer(context.Background(), t, "sleep inf", opts...)
+	t.Cleanup(func() {
+		testutils.RemoveDockerContainer(context.Background(), t, "test-ig-close")
+	})
```
blanquicet, line 139: With https://github.com/inspektor-gadget/inspektor-gadget/pull/1850, this will become:

```suggestion
	c := testutils.NewDockerContainer("test-ig-close", "sleep inf", testutils.WithoutLogs())
	c.Start(t)
	t.Cleanup(func() {
		c.Stop(t)
	})
```

pkg/container-collection/options.go
```diff
@@ -819,6 +819,9 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
 					container.Runtime.ContainerID, err)
 				return false
 			}
+			if container.mntNsFd != 0 {
+				log.Warnf("WithTracerCollection: mntns reference already set for container %s", container.Runtime.ContainerID)
+			}
 			container.mntNsFd = mntNsFd
```
blanquicet, line 825: Is the reassignment of `container.mntNsFd` enough to close (GC) the previous fd?

blanquicet (approved):
Just have one question but LGTM!

pkg/container-collection/options.go
alban: No, that would not close the previous fd if a previous fd exists.

If the previous `mntNsFd` is not zero, that's a bug, it is not supposed to happen. The bug could either be that there is a previous fd, or that due to some memory corruption, `mntNsFd` is assigned to a non-zero number that is not an open fd, or it could be another fd unrelated to mount namespaces that is used by other code. In that latter case, I thought it would be less harmful to potentially leak a fd rather than potentially close an unrelated fd.

What do you think?

pkg/ig-manager/ig-manager_test.go
alban: This looks cleaner :)

pkg/container-collection/options.go
blanquicet: Another case is if we introduce a regression and somehow we start calling the enrichers twice again. But it is supposed to be detected by the tests you added. I just tried and it does.

> In that latter case, I thought it would be less harmful to potentially leak a fd rather than potentially close an unrelated fd.
>
Yeah, but I think it safer to not close fds that we didn't open.

Issue #1681, opened 2023-05-30
[RFE] Adopt client/server setup also for ig


flyth:
## Current situation

Currently, `ig` only supports starting a single gadget at a time. To run multiple gadgets, you have to start several `ig` processes. Once an `ig` process is terminated, it will also no longer collect events from the gadgets.

## Impact

It's not easily possible to leave `ig` running in the background to collect events over a longer period of time.

## Ideal future situation

It would be great to have a long-running daemonized `ig` version running in the background that a client application can access to start/stop gadgets. This daemon should handle multiple gadgets at the same time and be accessible over a unix or tcp socket.

## Implementation options

We can re-use most of the parts we already have for the connection between `kubectl-gadget` and the `gadgettracermgr`. The `gadget-service` (pkg/gadget-service) can be used to listen for incoming requests and manage the gadget instantiation. The `grpc-runtime` of `kubectl-gadget` can be reused to connect to the daemon - for that we can use a direct connection instead of tunneling through the Kubernetes API server.

The only missing part is the persistence that yet has to be fully implemented based on the refactoring of #1281 and  #1396. This should be made reusable for this solution as well, once merged.

PR #1856, opened 2023-07-17
design: Add containerized gadgets design doc

mauriciovasquezbernal:
Add document with the design of the containerized gadget support in IG.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1722
Supersedes  https://github.com/inspektor-gadget/inspektor-gadget/pull/1716

### DIFF
--- a/docs/design/002-containerized-gadgets.md
+++ b/docs/design/002-containerized-gadgets.md
@@ -1,0 +1,697 @@
+# Containerized Gadgets Support Inspektor Gadget
+
+The gadgets we provide are heavily coupled to Inspektor Gadget as they are all developed on the same
+repository and maintained by our team. Adding new gadgets implies that they have to be reviewed and
+approved by us. This approach presents some disadvantages:
+- Users need to install new versions of Inspektor Gadget to get new gadgets (or updates to them).
+- Too specific gadgets can't be added to our code base.
+- Users can't have private gadgets.
+
+The [containerized gadgets idea](https://github.com/inspektor-gadget/inspektor-gadget/issues/1669)
+aims to make Inspektor Gadget a framework to run eBPF programs (gadgets), like Docker does for
+containers. This approach decouples the gadgets from Inspektor Gadget solving the issues mentioned
+above.
+
+Many ideas of this design document are based on [bumblebee](https://github.com/solo-io/bumblebee)
+and [ebpf_exporter](https://github.com/cloudflare/ebpf_exporter) projects.
+
+## User Experience
+
+The UX we provide should be very close to the existing container runtimes like Docker. This document
+considers the client-server setup for ig described in
+https://github.com/inspektor-gadget/inspektor-gadget/issues/1681. It's being implemented at the time
+of writing this document.
+
+### `run`
+
+The run command starts a gadget:
+
+```bash
+$ gadgetctl run mygadgetimage:latest
+# print output to terminal
+```
+
+`mygadgetimage:latest` is a gadget OCI image.
+This command should provide a flag to run on detached mode:
+
+```bash
+$ gadgetctl run mygadgetimage:latest --detach
+a3397355fc4b2ba49cdc1ab2d36f728cbda65f0ef979854186f2ea780e8659e1
+```
+
+This command automatically pulls the image if it is not available in the local image cache.
+
+### `list`
+
+List print the running gadgets
+
+```bash
+$ gadgetctl list
+...
+```
+
+### `attach`
+
+Attach to a running gadget
+
+```bash
+$ gadgetctl attach mygadget
+# gets terminal attached to the gadget
+```
+
+### `pull`
+
+Pulls a gadget image
+
+```bash
+$ gadgetctl pull mygadgetimage:latest
+```
+
+### `build`
+
+Builds and packages a gadget as an OCI image
+
+```bash
+$ gadgetctl build --prog foo.bpf.c --definition foo.yaml mygadgetimage:latest
+```
+
+Details of this command are still to be defined. Like other flags that are supported.
+
+### `tag`
+
+Tag an image
+
+```bash
+$ gadgetctl tag mygadgetimage:latest ghcr.io/foo/mygadgetimage:v1
+```
+
+### `login`
+
+Login into a container registry
+
+```bash
+$ gadgetctl login -u admin -p 1234
+```
+
+### `image`
+
+Different commands to handle images
+
+- `image list`
+- `image delete`
+etc...
+
+TODO: Should commands like `ig pull`, `ig build`, `ig push` be aliases of `ig image ...` ones?
+
+## Gadget Packaging
+
+Gadgets should be packed as [OCI](https://github.com/opencontainers/image-spec) images. Initially
+the following layers are proposed:
+- `application/vnd.definition.image.inspektor-gadget.io+yaml`: [Definition file](#definition-file)
+- `application/vnd.ebpf.image.inspektor-gadget.io+binary`: Compiled eBPF object.
+
+### OCI Annotations
+
+The OCI image should contain the following [OCI annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys):
+- `org.opencontainers.image.title`
+- `org.opencontainers.image.description`
+- `org.opencontainers.image.url`
+
+TODO: How the user could specify these annotations when using the build command?
+
+### Definition File
+
+The definition file contains the following information for a gadget:
+- `documentation`: Long text describing how to use the gadget. This should describe the different
+  parameters supported and can contain some practical examples about how to use it.
+- `columns`: Information about the event that the gadget provides. Each element represents a columns
+  attribute as defined in pkg/columns/columninfo.go.
+
+This definition file is mandatory in the early implementation stage, but later on it'll be
+optional.
+
+## Gadget Implementation
+
+A gadget is composed by eBPF programs and maps that are compiled into an eBPF ELF object. According
+to the names of the maps and programs, and the types used, Inspektor Gadget is able to understand
+how to parse, enrich and print information from the gadget.
+Inspektor Gadget requires the ELF file to have sections named according to the convention defined in
+libbpf and cilium/ebpf.
+
+Currently the built-in gadgets have a strong categorization, however in this new containerized
+gadgets approach that categorization is softer, each gadget defines how it behaves according to the
+maps / types they use and it's possible that the same gadget implements different behaviors, like
+tracer and topper, etc.
+
+### Maps
+
+#### `RingBuf` and `PerfEventArray` with `print_` Prefix (a.k.a tracers)
+
+These maps indicate that the gadget produces a stream of events. Inspektor Gadget uses the BTF
+information to format and print the events.
+
+```c
+// taken from gadgets/trace_open.bpf.c
+struct event {
+	__u64 timestamp;
+	__u32 pid;
+	__u32 uid;
+	__u32 gid;
+	mnt_ns_id_t mntns_id;
+	int ret;
+	int flags;
+	__u16 mode;
+	__u8 comm[TASK_COMM_LEN];
+	__u8 fname[NAME_MAX];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__type(value, struct event);
+} print_events SEC(".maps");
+```
+
+NOTE: In libbpf and cilium/ebpf, maps of type RingBuf and PerfEventArray normally have an explicit
+definition with:
+
+```c
+__uint(value_size, sizeof(u32));
+```
+
+Or an implicit one (when value_size is not defined, the bpf loader understands that RingBuf and
+PerfEventArray store a file descriptor, i.e. an integer of size 4).
+
+However, for containerized gadgets, Inspektor Gadget changes the semantic of the value field in the
+map definition to be able to carry the expected BTF type sent to the ring buffers. This idea
+initially comes from Bumblebee.
+
+As a consequence, BPF programs written in this way will not be loadable by generic bpf loaders but
+only by Inspektor Gadget.
+
+TODO: Revisit this appproach and check if there is a way to remove the limitation of being loaded by
+other projects.
+
+#### `HashMap` with `stats_` Prefix (a.k.a toppers)
+
+These maps are used to implement toppers, i.e. gadgets that print a list of elements sorted by
+specific parameters, like TCP connections by sent bytes.
+
+Inspektor Gadget automatically reads and sorts those maps according to the configuration provided by the user:
+- Interval: How often to print the map contents
+- Sort by: Field within the event type to sort by
+
+#### BPF Iterators (a.k.a snapshotters)
+
+Programs of type `iter/` are automatically loaded and attached by Inspektor Gadget, then they are
+triggered one time and its output is parsed according to the BTF information provided by it.
+
+In order to have BTF information available, the programs have to define an unused variable with the
+type they output in the iterator:
+
+```c
+const volatile event_t myiterator_output = {};
+
+const struct myevent *gadget_iter_type __attribute__((unused));
+
+SEC("iter/task_file")
+int myiterator(struct bpf_iter__task_file *ctx)
+```
+
+According to the type of iterator, it's run in different ways:
+- iter/task, iter/task_file: relative to the current pid namespace. Inspektor Gadget switches to the
+  host pid namespace as appropriate to get all processes.
+- iter/tcp, iter/udp: relative to the current network namespace. Inspektor Gadget iterates over all
+  network namespaces of interest and triggers the program in each of them. (containers selected with
+  the usual filter flags like --container)
+- iter/bpf_map_elem: relative to a map. Unsupported.
+
+#### `HashMap` with `hist_` Prefix (a.k.a profilers)
+
+These maps are used to implement profilers: gadgets that output a histogram. The histogram collection
+starts when the gadget is run, and it's printed when the gadgets stops.
+
+TODO1: It's possible to also print the histogram periodically, however it's not explored yet.
+
+TODO2: There is probably some overlap with histogram metrics support for Prometheus. It's very
+likely that the same gadget can be used for both purposes.
+
+#### Prometheus
+
+TODO: We can heavily base this on https://github.com/cloudflare/ebpf_exporter.
+
+##### Counters
+
+##### Histograms
+
+##### Gauges
+
+
+### Inspektor Gadget API for Gadgets Developers
+
+Inspektor Gadget provides some C header defining some types and functions to be used by gadgets.
+Those types and functions are used to enable some of the features described below.
+
+Please check the note in [Support Inspektor Gadget API
+changes](#support-inspektor-gadget-api-changes) to get further details about handling changes in
+this API.
+
+#### Data Types and Helper Headers
+
+Types provided in pkg/gadgets/common/types.h are used to format data in a specific way.
+
+- `endpoint_t`: Represent an L3 or L4 endpoint. Inspektor Gadget automatically enriches it
+  with the Kubernetes Pod and/or Service details corresponding to that IP address.
+- pkg/gadgets/common/mntns_filter.h: used to filter and enrich data by mount namespace
+
+### Types of eBPF programs
+
+Inspektor Gadget automatically loads and attaches the eBPF programs. The following table describes
+the current support and the future plans:
+
+Symbols:
+- ✅: implemented
+- 👷: work in progress
+- 📅: desired feature, we should plan it
+
+| ebpf program type     | Support | Difficulty |
+|-----------------------|---------|------------|
+| socket_filter         |   👷    |            |
+| sk_reuseport/migrate  |         |            |
+| sk_reuseport          |         |            |
+| kprobe/               |   ✅    |            |
+| uprobe/               |         |            |
+| kretprobe/            |   ✅    |            |
+| uretprobe/            |         |            |
+| tc                    |         |            |
+| classifier            |         |            |
+| action                |         |            |
+| tracepoint/           |   ✅    |            |
+| tp/                   |         |            |
+| raw_tracepoint/       |         |            |
+| raw_tp/               |         |            |
+| raw_tracepoint.w/     |         |            |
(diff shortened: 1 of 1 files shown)


docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,592 @@
...
+
+Programs of type `iter/` are automatically loaded and attached by Inspektor Gadget, then they are
+triggered and its output is parsed according to the BTF information provided by it.
+
+TODO: It's not clear how to provide that BTF information as there is not any place in the iter
+program to define it. One possibility is to define a global constant and manually associated that to
+the program, by using its name for instance.
+
+```c
+const volatile event_t myiterator_output = {};
```
mauriciovasquezbernal, line 217: I think it can work as done in https://github.com/inspektor-gadget/inspektor-gadget/pull/1866. 

It's still not clear how to handle other kind of iterators, like socket ones that should be executed in different network namespaces.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,603 @@
...
+### `run`
+
+The run command starts a gadget:
+
+```bash
+$ ig run mygadgetimage:latest
+# print output to terminal
+```
+
+`image` is gadget OCI image.
```
burak-ok, line 30: nit: In the command you used `mygadgetimage` here it says `image`

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,603 @@
...
+help gadget developers with this, like:
+- Integration testing:
+  https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/helpers.go
+- Unit tests: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/internal/test
+
+
+Inspektor Gadget will use some specific design gadgets for testing purposes to test that all the
+APIs and functionality it provides to gadgets is working fine.
+
+TODO: We need to double check this. IMO we shouldn't be testing external gadgets in Inspektor
```
burak-ok, line 314: I have the same opinion. We should only test the gadgets that we maintain and provide. Also making sure that we are testing all/most of the API and features Inspektor Gadget provides

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,603 @@
...
+
+func main() {
+	opts := runner.Opts{
+		// Authentcation options
+    // Gadget parameters
+	}
+	gadget := runner.Run("ghcr.io/foo/mygadget:v1", opts)
+	defer gadget.Close()
+
+	for gadget.Done() {
```
burak-ok, line 364:
```suggestion
	for !gadget.Done() {
```

burak-ok (approved):
LGTM.
Since I don't know all the gadgets other people should have a look at this too

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+- Users can't have private gadgets.
+
+The [containerized gadgets idea](https://github.com/inspektor-gadget/inspektor-gadget/issues/1669)
+aims to make Inspektor Gadget a framework to run eBPF programs (gadgets), like Docker does for
+containers. This approach decouples the gadgets from Inspektor Gadget solving the issues mentioned
+above.
+
+Many ideas of this design document are based on [bumblebee](https://github.com/solo-io/bumblebee)
+and [ebpf_exporter](https://github.com/cloudflare/ebpf_exporter) projects.
+## User Experience
```
alban, line 17: New line before header

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+### `run`
+
+The run command starts a gadget:
+
+```bash
+$ ig run mygadgetimage:latest
+# print output to terminal
+```
+
+`mygadgetimage:latest` is gadget OCI image.
```
alban, line 30:
```suggestion
`mygadgetimage:latest` is a gadget OCI image.
```

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+This command should provide a flag to run on detached mode:
+
+```bash
+$ ig run mygadgetimage:latest --detached
```
alban, line 34:
```suggestion
$ ig run --detach mygadgetimage:latest
```

Docker calls it `--detach` (and not `--detached`) and it has to be given before the image name.

```
  -d, --detach                         Run container in background and print container ID
```

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+Created gadget foo
```
alban, line 35:
```suggestion
a3397355fc4b2ba49cdc1ab2d36f728cbda65f0ef979854186f2ea780e8659e1
```

`docker run -d` returns an id without text, so scripts don't have to parse the string.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+```
+
+This command automatically pulls the image if it's available in the local image cache.
```
alban, line 38:
```suggestion
This command automatically pulls the image if it is not available in the local image cache.
```

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+the following layers are proposed:
+- `application/config.image.inspektor-gadget.io+yaml`: [Configuration file](#configuration-file)
+- `application/ebpf.image.inspektor-gadget.io+binary`: Compiled eBPF object.
+
+### Configuration File
+
+The configuration file contains the following information for a gadget:
+- `name`: Human readable name of the gadget, like `trace_exec`, `profile_ebpf`, etc.
+- `description`: Short description about the gadget.
+- `documentation`: Long text describing how to use the gadget. This should describe the different
```
alban, line 106: Some of those fields are redundant with the OCI annotations that are already defined for all OCI images in:
https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

So instead of duplicating them (and then we don't know which ones are the valid ones), should we give a recommendation that users should set the OCI annotations like documentation, url, etc.?

If we need to duplicate them for practical reasons, I would like to have a generic `annotations map[string]string` reusing the keys defined by OCI.

docs/design/002-containerized-gadgets.md
blixtra, line 38: Should this be "if it's _not_ in the local image cache". It should just run it if it is in the local image cache, right?

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+
+These maps are use to implement profilers: gadgets that output a histogram. The histogram collection
+starts when the gadget is run, and it's printed when the gadgets stops.
+
+TODO1: It's possible to also print the histogram periodically, however it's not explored yet.
+
+TODO2: There is probably some overlap with histogram metrics support for Prometheus. It's very
+likely that the same gadget can be used for both purposes.
+
+#### Prometheus
```
blixtra, line 186: I don't understand why this is a special category and not an export option for various other gadgets. What I'd imagined is that certian types of gadgets will be able to configure support for Prometheus. Then in the config file we could specify how that works and the user could then make use of that at runtime if that's what they wanted. Optionally, they could expose it as an API or imply have it output to standard out.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+### Running Gadgets Directly
+
+There are some cases where external applications want to run gadgets directly in their processes,
+i.e. by using gadgets as a library. Currently, the different builtin gadgets provide golang packages
+that can be consumed by external applications. Some examples are available in ../../examples. In the
+containerized gadgets the same approach can't be used as these gadgets don't provide a golang
+package.
+
+In order to support this case with containerized gadgets, Inspektor Gadget should provide a golang
+package that provides the primites for applications to run and consume data generated by gadgets.
```
blixtra, line 342: typo: primitives

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+
+Convert bitfield from an event to a human readable string. Example: bind's [optionsToString](https://github.com/inspektor-gadget/inspektor-gadget/blob/b57f2bae31a46b40d8e0204b85099ae37f15d21d/pkg/gadgets/trace/bind/tracer/tracer.go#L154).
+
+#### Iface index
+
+Convert network interface index to string. Example: bind's [BoundDevIf](https://github.com/inspektor-gadget/inspektor-gadget/blob/b57f2bae31a46b40d8e0204b85099ae37f15d21d/pkg/gadgets/trace/bind/tracer/tracer.go#L236-L250).
+
+#### Custom user space logic
+
+There are some gadgets that have a lot of logic in user space, to support this gadgets we'll need to
```
blixtra, line 435: I think we should tackle this soon. It's a very empowering feature and one that we already need.

docs/design/002-containerized-gadgets.md
blixtra, line 34: Is `--detach` just for preventing the output from going to standard out? What is the difference to `--deamon`

blixtra (commented):
Thanks for this. This focuses on the specifics of individual containers. For example, the configuration if for the gadget. It's not the user configuration file for how to run the container. Would it not be useful at this stage to also give some consideration for that user configuration and how the user, for example, might want to bundle gadgets or even deploy them as a bundle?

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+
+It could be implemneted by using the pkg/params package.
+
+
+#### Enum convert
+
+Convert an enum to a string. Examples: pktTypeNames, qTypeNames, rCodeNames, etc. Some could use BTF
+from the kernel, but not all.
+
+Mauricio: I think we can use the BTF from the program too.
```
alban, line 419: Good idea. When reading the BTF from the ELF object, we will need to distinguish the types defined internally and those defined externally (in the kernel) and then implement the "relocation" for the current kernel.

docs/design/002-containerized-gadgets.md
alban, line 435:
```suggestion
There are some gadgets that have a lot of logic in user space. To support those gadgets, we'll need to
```

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+
+#### Iface index
+
+Convert network interface index to string. Example: bind's [BoundDevIf](https://github.com/inspektor-gadget/inspektor-gadget/blob/b57f2bae31a46b40d8e0204b85099ae37f15d21d/pkg/gadgets/trace/bind/tracer/tracer.go#L236-L250).
+
+#### Custom user space logic
+
+There are some gadgets that have a lot of logic in user space, to support this gadgets we'll need to
+run custom code they provide. This presents big challenges (security, flexibility) and it's out of
+scope for now. We can revisit this later on if a needed arises.
```
alban, line 437:
```suggestion
scope for now. We can revisit this later on if the need arises.
```

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+#### snapshot process
+
+It supports a compatibility mode to get processes when iterators are not available ->
+https://github.com/inspektor-gadget/inspektor-gadget/blob/e91a93bb2c0a7d05ed6ac2689b89ec7465923dcf/pkg/gadgets/snapshot/process/tracer/tracer.go#L216.
+We'll lose that in the containerized version.
+
+#### snapshot socket
+
+Inspektor Gadget should automatically run the iter/ programs on the different network namespaces of
+the containers. It's still TBD how a gadget will define that it wants this behavior.
```
alban, line 479: It depends on the type of iterator:
* iter/task, iter/task_file: relative to the pid namespace. But since pid namespaces are hierarchical, to be executed on the host pid namespace. If running with hostPID=true, easy to do. If running without hostPID, it is done with a host systemd unit thanks to pkg/utils/bpf-iter-ns/
* iter/tcp, iter/udp: relative to the net namespace. To be run in all network namespaces of interest (containers selected with the usual filter flags like `--container`)
* iter/bpf_map_elem: relative to a map. Unsupported.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: I agree to name is as `--detach`. Regarding the position of the flag I think it doesn't matter in our case as `run` doesn't expect a command as in the docker case.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Yes!

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Yes, I agree we should use the OCI ones. Also because the configuration file should be optional later on.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: It's not about having another category, but about describing how a gadget should implement that support. All this section describes the different features gadgets can implement, Prometheus is just another of those features. In order to enable prometheus export, a given gadget will need to define some maps in a specific format, like `count_`, `histogram_`, et. (TBD), then Inspektor Gadget will expose those prometheus metrics automatically.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Right. I changed the wording to make it clear we'll do that later on.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Good point. `--detach` only makes sense for `gadgetctl` (https://github.com/inspektor-gadget/inspektor-gadget/issues/1681) or `kubectl-gadget`. It means to run the gadget and detach the terminal from it. `--daemon` is for `ig` and it means to run it as an agent on background. I updated the text to (hopefully) make it more clear.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+	__u8 fname[NAME_MAX];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(u32));
+	__type(value, struct event);
+} print_events SEC(".maps");
+```
+
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Added the note and a TODO to check this again. I really would like this to be removed so we don't have to handle it.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+### Inspektor Gadget API for Gadgets Developers
+
+Inspektor Gadget provides some C header defining some types and functions to be used by gadgets.
+Those types and functions are used to enable some of the features described below.
+
+#### Data Types and Helper Headers
+
+Types provided in pkg/gadgets/common/types.h are used to format data in a specific way.
+
+- `ip_addr`: Union representing either IPv4 or IPv6 address.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it can be useful when not enrichment is desired for a given IP address (because of performance reasons). However I think we can go ahead with only the endpoints and introduce other types if needed later on. I'll remove `ip_addr` then.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+- `endpoint_t`: Represent an L3 or L4 endpoint. Inspektor Gadget automatically enriches it
+  with the Kubernetes Pod and/or Service details corresponding to that IP address.
+- pkg/gadgets/common/mntns_filter.h: used to filter and enrich data by mount namespace
+
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Added a link to the already existing "Support Inspektor Gadget API changes" section.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+	"ig.io/runner"
+)
+
+func main() {
+	opts := runner.Opts{
+	// Authentcation options
+	// Gadget parameters
+	}
+	gadget := runner.Run("ghcr.io/foo/mygadget:v1", opts)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Makes sense. I added a new section about this. But this is something we need to explore further to define the exact API to use in both cases.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,602 @@
...
+#### Endpoint enrichment
+
+Networking gadgets that want to enrich IP addresses with Pod and Services name can use the `gadget_l3endpoint_t` and `gadget_l4endpoint_t` types provided by Inspektor Gadget.
+
+NOTE: It's under development at the time of writing this document in
+https://github.com/inspektor-gadget/inspektor-gadget/pull/1825.
+
+#### `socket_filter` programs
+
+Attach ebpf programs of type `BPF_PROG_TYPE_SOCKET_FILTER`. TODO: What's special about these?
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Thanks for the explanation. I added a sentence about this package.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Makes sense. The whole mechanism is not 100% clear to me, but it can be defined when implementing it.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Thanks, added this comment in the "BPF Iterators (a.k.a snapshotters)" section.

docs/design/002-containerized-gadgets.md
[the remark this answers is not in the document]
burak-ok: > I think it can be useful when not enrichment is desired for a given IP address (because of performance reasons).

I think this should be configurable by the user of the gadget and not the author. We need to implement "toggleable" operators anyway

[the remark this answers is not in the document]
mauriciovasquezbernal: Yeah, I agree. We can try to enrich all IPs, but providing an option for the user to disable it.

alban (approved):
LGTM, thanks!

I've made some phrasing suggestions, but you can merge after checking/addressing them.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,685 @@
...
+```
+
+Details of this command are still to be defined. Like other flags that are supported.
+
+### `tag`
+
+Tag an image
+
+```bash
+$ gadgetctl tag mygadgetimage:latest ghcr.io/foo/mygadgetimage:v1
```
mqasimsarfraz, line 85: Perphaps we should have a `login` command as well to write credentials to file like [docker login](https://docs.docker.com/engine/reference/commandline/login/). Since ready credentials from `~/.docker/config.json` is a common practice.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,685 @@
...
+etc...
+
+TODO: Should commands like `ig pull`, `ig build`, `ig push` be aliases of `ig image ...` ones?
+
+## Gadget Packaging
+
+Gadgets should be packed as [OCI](https://github.com/opencontainers/image-spec) images. Initially
+the following layers are proposed:
+- `application/config.image.inspektor-gadget.io+yaml`: [Configuration file](#configuration-file)
+- `application/ebpf.image.inspektor-gadget.io+binary`: Compiled eBPF object.
```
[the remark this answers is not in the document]
alban: I think you're right, we should use the [vendor tree](https://en.wikipedia.org/wiki/Media_type#Vendor_tree) because:

> Registrations in the standards tree must be either associated with IETF specifications approved directly by the IESG, or registered by an IANA recognized standards-related organization

And we are not registering our media-type with any standards-related organization. So it should be namespaced correctly.

https://datatracker.ietf.org/doc/html/rfc2048#section-2.1

>    Registrations in the vendor tree will be distinguished by the leading
   facet "vnd.".  That may be followed, at the discretion of the
   registration, by either a media type name from a well-known producer
   (e.g., "vnd.mudpie") or by an IANA-approved designation of the
   producer's name which is then followed by a media type or product
   designation (e.g., vnd.bigcompany.funnypictures).

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,685 @@
...
+type they output in the iterator:
+
+```c
+const volatile event_t myiterator_output = {};
+
+const struct myevent *gadget_iter_type __attribute__((unused));
+
+SEC("iter/task_file")
+int myiterator(struct bpf_iter__task_file *ctx)
+```
```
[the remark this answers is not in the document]
mauriciovasquezbernal: It's ignored in https://github.com/inspektor-gadget/inspektor-gadget/pull/1866.

docs/design/002-containerized-gadgets.md
```diff
@@ -0,0 +1,685 @@
...
+Attach ebpf programs of type `BPF_PROG_TYPE_SOCKET_FILTER`. These programs need to be attached in
+each network namespace of interest. Currently it's handled by the
+`pkg/gadgets/internal/networktracer` package. Somehow the same logic needs to be used for
+containerized gadgets.
+
+#### Socket enricher
+
+Use the socket enricher from the ebpf code. This can be done in a very similar way to the
+`gadget_mntns_filter_map`: Inspektor Gadget looks for this map, automatically replaces and keeps it
+updated with the right one.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Much better this way. Thanks!

docs/design/002-containerized-gadgets.md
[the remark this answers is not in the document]
mauriciovasquezbernal: Actually I have a question here. We discussed to use something different in https://github.com/inspektor-gadget/inspektor-gadget/pull/1863/files#r1273752216, which one should we use? It seems like there is not a standard one for ebpf ELF files.

docs/design/002-containerized-gadgets.md
mauriciovasquezbernal: Makes sense.

PR #1863, opened 2023-07-18
containerized gadgets: Add build/tag/push/pull/list command

burak-ok:
# containerized gadgets: Add image build/tag/push/pull/list/login command

- `ig image build --prog-amd64 ./gadgets/trace_tcpconnect_x86.bpf.o --prog-arm64 ./gadgets/trace_tcpconnect_arm64.bpf.o --definition ./gadgets/trace_tcpconnect.yaml ghcr.io/burak-ok/test:latest`
  This builds the given eBPF ELF files and the definition file into an oci image and saves it on the filesystem
- `ig image tag ghcr.io/burak-ok/test:latest ghcr.io/burak-ok/test2:latest`
  Adds the second tag to the image referenced by the first tag
- `ig image login`
- `ig image list`
  List all local tagged oci images on the filesystem
   ```shell
   REPOSITORY                     TAG                          DIGEST      
   ghcr.io/burak-ok/test2         latest                       18e0a9df8e44
   ghcr.io/burak-ok/test          latest                       18e0a9df8e44
   ```
- `ig image push ghcr.io/burak-ok/test:latest`
  Pushes the locally tagged image to the given remote repository
- `ig image pull ghcr.io/burak-ok/test:latest`
  Pulls the tagged image from the given remote repository to the local filesystem

The `viper` library was added for reading env variables/config. It has excellent support for `cobra` too as @flyth mentioned.
This is not not the best integration of the library but this PR is more about the `ig push...` commands than the `viper` library.

### Example 
```bash
$ go run ./cmd/ig/... image build --prog-amd64 ./gadgets/trace_tcpconnect_x86.bpf.o --prog-arm64 ./gadgets/trace_tcpconnect_arm64.bpf.o --definition ./gadgets/trace_tcpconnect.yaml ghcr.io/burak-ok/test:latest
$ go run ./cmd/ig/... image push ghcr.io/burak-ok/test:latest 
```
which results into https://github.com/users/burak-ok/packages/container/package/test
```json
$ crane manifest ghcr.io/burak-ok/test:latest | jq .
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:5719faf80e2a1d574aea3c643c648502e5e616b09f8b90a8b7fedd0caa748f2e",
      "size": 590,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:bfc3b708670972b9bd0f1d7e64b3f3772c677e0791d39dffa4b9cfa47031ff4f",
      "size": 590,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```
Example for `amd64`:
```json
$ crane manifest ghcr.io/burak-ok/test@sha256:5719faf80e2a1d574aea3c643c648502e5e616b09f8b90a8b7fedd0caa748f2e | jq .
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/config.image.inspektor-gadget.io+yaml",
    "digest": "sha256:eb71861c0a5d5ca9a706fffa20fead54f64a1624dfeef5d334a8a3c2462c4c37",
    "size": 628,
    "annotations": {
      "org.opencontainers.image.title": "config.json"
    }
  },
  "layers": [
    {
      "mediaType": "application/ebpf.oci.image.config.v1+binary",
      "digest": "sha256:059d8c8c3a726b24965cb454a4d7cef563b92c26ac7aa17a9aba6d1512f5bc24",
      "size": 866264,
      "annotations": {
        "org.opencontainers.image.authors": "TODO: authors",
        "org.opencontainers.image.description": "TODO: description",
        "org.opencontainers.image.title": "program.o"
      }
    }
  ]
}
```


mauriciovasquezbernal:
I think it's a good foundation for the work we want to do, however I think we should have a more docker-like experience. I was thinking about (related to https://github.com/inspektor-gadget/inspektor-gadget/pull/1856):

```
# builds an image stored locally (as docker build does)
$ ig build --prog myprog.bpf.c --definition foo.yaml myimage:latest

# tag with remote OCI registry url 
$ ig tag myimage:latest ghcr.io/foo/myimage:v1

# show existing images 
$ ig image list
# TODO: should we provide delete?, etc? 

# push to remote registry 
$ ig push ghcr.io/foo/myimage:v1

# pull image to local cache
$ ig pull ghcr.io/foo/myimage:v99

# run image (it'll be automatically pulled if not present)
$ ig run ghcr.io/foo/myimage:v99
```

I suppose we can use https://pkg.go.dev/oras.land/oras-go/v2@v2.2.1/content/file to handle the local image cache.

burak-ok:
# Please read the updated PR description

I added the following commands:
- `ig build`
- `ig tag`
- `ig list` (Without image prefix since we don't have any other image based operations. Can be easily moved later if needed)
- `ig pull`
- `ig push`

`ig-run` is not implemented yet

> Wouldn't it be possible to get a flag for the eBPF object file and one for the architecture?
> `$ ig push --prog ./foo/bar.io --arch arm64`

When building the image I need both ELF files and there a differentiation has to happen. Something with `--arch` would be needed twice then for a multiarch image

> I suppose we can use https://pkg.go.dev/oras.land/oras-go/v2@v2.2.1/content/file to handle the local image cache.

For locally stored images one has to use https://pkg.go.dev/oras.land/oras-go/v2@v2.2.1/content/oci instead. I also though that this could be a way to ship Inspektor Gadget with preloaded gadgets for environments without internet or connections to outside registries.
For now it is stored in `~/.ig/oci-store`. I thought about having a fallback to an in-memory store for cases where there is no home directory (some servers). So at least the gadget can be loaded and be ran. Of course this in-memory store makes only sense for `ig run`

blixtra:
I'm in agreement with @mauriciovasquezbernal [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/1863#issuecomment-1641119987). This should be a clean and familiar experience. People already know how to work with containers, let's build on that.

Additionally, I'd like to see us have conventions for building. For example, a user should be able to run `ig build` in a directory and `ig build` should do a couple things.

1. Check for a gadget.build file. If that file is found we should the use the information within to build the the gadget.
2. If that file is not found and there is a .c file (BPF code) and a definition with the same name, build that.
3. If none of those are found, exit with an explanation.

This allows gadget builders to easily have a gadget.build file along side the code checked into the repo with all the build instructions included.

We should of course offer arguments to override, but that should not be the default and preferred experience.

cmd/common/containerized-gadgets/build.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func pushDescriptorIfNotExists(target oras.Target, desc ocispec.Descriptor, contentReader io.Reader) error {
+	exists, err := target.Exists(context.Background(), desc)
+	if err != nil {
+		return fmt.Errorf("check descriptor exists: %w", err)
+	}
+	if exists {
+		return nil
+	}
```
mauriciovasquezbernal, line 95: One alternative is to always try to push and check if the return value is `ErrAlreadyExists` (https://pkg.go.dev/github.com/containerd/containerd/errdefs#ErrAlreadyExists)

cmd/common/containerized-gadgets/build.go
```diff
@@ -0,0 +1,231 @@
...
+	err = target.Push(context.Background(), desc, contentReader)
+	if err != nil {
+		return fmt.Errorf("push descriptor: %w", err)
+	}
+
+	return nil
+}
+
+func createEbpfDescAndPush(target oras.Target, progFilePath string) (*ocispec.Descriptor, error) {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Nit: Many functions are returning a pointer but it's really not useful because it seems all oras functions receive the object, so I think it's better to return the structure to avoid the use of `*` and `&` all around.

cmd/common/containerized-gadgets/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"strings"
+
+	"github.com/spf13/viper"
+	"oras.land/oras-go/v2/content/oci"
+	"oras.land/oras-go/v2/registry/remote/auth"
+)
+
+func getLocalOciStore() (*oci.Store, error) {
+	home, err := os.UserHomeDir()
```
mauriciovasquezbernal, line 28: What about using `/var/lib`?

cmd/common/containerized-gadgets/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("get user home dir: %w", err)
+	}
+
+	return oci.New(home + "/.ig/oci-store")
```
mauriciovasquezbernal, line 33: use `filepath.Join()`.

cmd/common/containerized-gadgets/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	// So we need everything before the last double, which includes also the port
+	splitted := strings.Split(tag, ":")
+	return strings.Join(splitted[:len(splitted)-1], ":")
+}
+
+func getHostString(repoUrlStr string) string {
+	// For now just assume its something like ghcr.io:443/inspektor-gadget/ig
+	// So we need everything before the first slash, which includes also the port
+	return strings.Split(repoUrlStr, "/")[0]
+}
```
mauriciovasquezbernal, line 47: Should we add some checks to this to be sure it doesn't crash if the output is wrong?

cmd/common/containerized-gadgets/list.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns/formatter/textcolumns"
+	"github.com/spf13/cobra"
```
mauriciovasquezbernal, line 25: nit: Please move this import above to group imports.

cmd/common/containerized-gadgets/build.go
```diff
@@ -0,0 +1,231 @@
...
+	definitionFilePath string
+	tag                string
+}
+
+func NewBuildCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "build tag",
+		Short: "build a containerized gadget image",
+		RunE:  runBuild,
+	}
```
mauriciovasquezbernal, line 44: I think all these commands should set `SilenceUsage: true` to avoid printing misleading (not useful) information when there is a failure.

cmd/common/containerized-gadgets/list.go
```diff
@@ -0,0 +1,88 @@
...
+			imageColumn := imageColumn{
+				Repository: repo,
+				Tag:        strings.TrimLeft(fullTag, repo+":"),
+			}
+
+			desc, err := ociStore.Resolve(context.Background(), fullTag)
+			if err == nil {
+				imageColumn.Digest = desc.Digest.String()
+			}
+			// TODO: else debug log?
```
mauriciovasquezbernal, line 63:
```
if err != nil {
 log....
 continue;
}

everythingelsethatshouldbedonewhengood()
````

File: cmd/common/containerized-gadgets/push.go [hunk 9]
```diff
@@ -0,0 +1,71 @@
...
+	"oras.land/oras-go/v2/registry/remote"
+)
+
+type pushOptions struct {
+	tag string
+}
+
+func NewPushCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "push tag",
```
mauriciovasquezbernal, line 32: nit: `tag` -> `TAG` to make it clear that it's an argument. (Same in other commands).

mauriciovasquezbernal (commented):
I like it. We are still missing the support for actually building the eBPF code, but I think we can add that in a second iteration. Having this support will help a lot to test and implement OCI support in the `run` command. I tested and everything is working fine. Some nits around.

mauriciovasquezbernal:
> Check for a gadget.build file. If that file is found we should the use the information within to build the the gadget.

How are you imagining this file? I think it can be a very simple version of Dockerfile, more like a build script that tells `ig build` the right commands to invoke to build the eBPF objects. I'm even wondering if we could simplify it and just use a Makefile.

cmd/common/containerized-gadgets/build.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("open eBPF program file: %w", err)
+	}
+	defer progFile.Close()
+	progBytes, err := io.ReadAll(progFile)
+	if err != nil {
+		return nil, fmt.Errorf("read eBPF program file: %w", err)
+	}
+
+	progDesc := content.NewDescriptorFromBytes("application/ebpf.oci.image.config.v1+binary", progBytes)
```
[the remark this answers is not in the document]
burak-ok: I actually don't know. It is the same as in bumblebee and it seemed generic enough.

And since the eBPF files that we have are also "just standard" ELF files i thought we should make that layer as generic as possible

burak-ok:
Should I remove `ig pull` completely? With #1875in mind we probably don't need (and want) OCI images on the "client" side.

What would be the use case for `ig pull` then?

mauriciovasquezbernal:
I think it can still be useful for ig when https://github.com/inspektor-gadget/inspektor-gadget/issues/1681 is implemented. Also, in the `kubectl-gadget` case it could be useful for pre-pulling the images, but we should change the logic to pull that in each node and not on the client.

cmd/common/containerized-gadgets/build.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Fair enough.

cmd/common/containerized-gadgets/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("get user home dir: %w", err)
+	}
+
+	return oci.New(home + "/.ig/oci-store")
+}
+
+func getRepositoryFromTag(tag string) string {
+	// For now just assume its something like ghcr.io:443/inspektor-gadget/ig:latest
+	// So we need everything before the last double, which includes also the port
```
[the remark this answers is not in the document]
burak-ok: My german language came out :D We call it "double point"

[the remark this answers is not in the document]
mauriciovasquezbernal: I feel you, we call it "two points" in Spanish and I have made a similar mistake multiple times :D.

mauriciovasquezbernal:
> How are you imagining this file? I think it can be a very simple version of Dockerfile, more like a build script that tells `ig build` the right commands to invoke to build the eBPF objects. I'm even wondering if we could simplify it and just use a Makefile.

I checked bumblebee's implementation and I think we can follow a very similar approach: It has a build command like `bee build INPUT_FILE REGISTRY_REF [flags]`. Some interesting things are:
- Build is done by using a build script. A [default one](https://github.com/solo-io/bumblebee/blob/main/builder/build.sh) is provided but can also be changed. (`--build-script`)
- It's possible to compile using the tool chain installed in the local machine or by using a builder docker image provided by the project. (`--local` option)
- It's possible to build the ebpf objects without generating the oci image. (`--binary-only`)

cmd/common/containerized-gadgets/helpers.go
burak-ok: I don't think that this will crash. `splitted[:len(splitted)-1]` will return an empty string if there is no `:` or if its an empty string. 
`strings.Split(repoUrlStr, "/")[0]` will also always return at least something (even if it is an empty string)

None the less we should have better warning/error messages regarding this. Should these function return `(string, error)` or print the error themselves?

cmd/common/containerized-gadgets/build.go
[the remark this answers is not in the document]
burak-ok: It was either returning `nil` + dealing with dereferencing or returning `ocispec.Descriptor{}`. I changed it, since I don't know which way is the better one :D

cmd/common/containerized-gadgets/helpers.go
mauriciovasquezbernal: IMO it's always better to have it completely under control and throw an error if the input is not what we expect. Specially in this case that it's coming from the user.

cmd/common/containerized-gadgets/helpers.go
burak-ok: I would argue that it depends. The currently added commands don't require being root.

Do we want to enforce that?

mauriciovasquezbernal: Let's keep using $HOME for now then. I think it's a parameter that we'll make configurable later on anyway.

mauriciovasquezbernal (commented):
It's looking very good!

cmd/common/containerized-gadgets/image/tag.go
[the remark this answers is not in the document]
burak-ok: Will be fixed with next update to this repo:
```
REPOSITORY                                                                     TAG                                                                            DIGEST      
ghcr.io/burak-ok/test                                                          latest                                                                         b448ce1c81c6
ghcr.io/mauriciovasquezbernal/mytraceexec                                      latest                                                                         b448ce1c81c6
ghcr.io/mauriciovasquezbernal/mytraceexec                                      latestbar                                                                      b448ce1c81c6
ghcr.io/mauriciovasquezbernal/mytraceexec                                      latestv5555                                                                    b448ce1c81c6
test                                                                           latest                                                                         b448ce1c81c6
```

burak-ok:
Update:
1. Files are moved out of the `containerized-gadgets` subfolder
2. Default path for `oci-store` and `config.json` (auth) is `/var/lib/ig`
3. Added `ig login ghcr.io` based on https://github.com/containers/common/blob/main/pkg/auth/auth.go#L87

cmd/common/login.go
```diff
@@ -0,0 +1,48 @@
...
+
+package common
+
+import (
+	"context"
+	"os"
+
+	"github.com/containers/common/pkg/auth"
+	"github.com/containers/image/v5/types"
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
```
mauriciovasquezbernal, line 23: nit: please move this to a different block.

cmd/common/utils/helpers.go
```diff
@@ -0,0 +1,191 @@
...
+
+	var cfg *configfile.ConfigFile
+	var err error
+
+	if authOptions.AuthFile == "" {
+		cfg, err = config.Load(authOptions.AuthFile)
+		if err != nil {
+			return nil, fmt.Errorf("load auth config: %w", err)
+		}
+	} else {
```
mauriciovasquezbernal, line 162: I see you removed the `use-docker` option, what's the plan there? By using something like 

```diff
diff --git cmd/common/utils/helpers.go cmd/common/utils/helpers.go
index ab7071fd..93394d79 100644
--- cmd/common/utils/helpers.go
+++ cmd/common/utils/helpers.go
@@ -153,9 +153,14 @@ func CreateAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.C
 
        var cfg *configfile.ConfigFile
        var err error
+       var found bool
 
-       if authOptions.AuthFile == "" {
-               cfg, err = config.Load(authOptions.AuthFile)
+       if _, err := os.Stat(authOptions.AuthFile); err == nil {
+               found = true
+       }
+
+       if !found {
+               cfg, err = config.Load("")
                if err != nil {
                        return nil, fmt.Errorf("load auth config: %w", err)
                }
```

we can make it default to use docker configuration if the ig one is not available.

mauriciovasquezbernal (approved):
Two minor comments. LGTM after those are handled. 

Btw, we have a login command, we should have a logout too. But it can be handled later on.

Good work!

cmd/common/utils/helpers.go
burak-ok: I misread a review comment from earlier :D

burak-ok:
> Btw, we have a login command, we should have a logout too. But it can be handled later on.

Added

cmd/common/utils/helpers.go
mauriciovasquezbernal: Looks good to me. I won't worry too much about the $HOME issue, probably that's gone be changed when https://github.com/inspektor-gadget/inspektor-gadget/issues/1681 gets in.

mauriciovasquezbernal (approved):
I tested the logout command and it works fine. LGTM again! 

Thanks for handling all the comments.

mqasimsarfraz (commented):
Looking good overall. Just small comments. I am having issues testing it via complied `ig` command. (`go run -exec sudo ./cmd/ig ...` works fine):

```
→ sudo ./ig image push ghcr.io/mqasimsarfraz/test
Pushing ghcr.io/mqasimsarfraz/test:latest...
fatal error: unexpected signal during runtime execution
[signal SIGFPE: floating-point exception code=0x1 addr=0x7f4e9cd75d49 pc=0x7f4e9cd75d49]

runtime stack:
runtime.throw({0x29df0a5?, 0x20186ba?})
	/usr/local/go/src/runtime/panic.go:1047 +0x5d fp=0x7f4e9e7fb120 sp=0x7f4e9e7fb0f0 pc=0x43ca9d
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:823 +0x369 fp=0x7f4e9e7fb170 sp=0x7f4e9e7fb120 pc=0x453069

goroutine 52 [syscall]:
→ sudo ./ig image push ghcr.io/mqasimsarfraz/test
Pushing ghcr.io/mqasimsarfraz/test:latest...
fatal error: unexpected signal during runtime execution
[signal SIGFPE: floating-point exception code=0x1 addr=0x7f4e9cd75d49 pc=0x7f4e9cd75d49]

runtime stack:
runtime.throw({0x29df0a5?, 0x20186ba?})
	/usr/local/go/src/runtime/panic.go:1047 +0x5d fp=0x7f4e9e7fb120 sp=0x7f4e9e7fb0f0 pc=0x43ca9d
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:823 +0x369 fp=0x7f4e9e7fb170 sp=0x7f4e9e7fb120 pc=0x453069

goroutine 52 [syscall]:
```

Any idea what might be wrong?

cmd/common/image/list.go
```diff
@@ -0,0 +1,97 @@
...
+func runList(cmd *cobra.Command, args []string) error {
+	ociStore, err := utils.GetLocalOciStore()
+	if err != nil {
+		return fmt.Errorf("get oci store: %w", err)
+	}
+
+	type imageColumn struct {
+		Repository string `column:"repository"`
+		Tag        string `column:"tag"`
+		Digest     string `column:"digest,width:12,fixed"`
```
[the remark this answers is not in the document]
mqasimsarfraz: Resolving this, this should be indeed `Digest` :see_no_evil: [imageid vs digest](https://stackoverflow.com/a/56391252/7274380) can get confusing :smile:

cmd/common/image/push.go
```diff
@@ -0,0 +1,83 @@
...
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/registry/remote"
+
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
+)
+
+type pushOptions struct {
+	image    string
+	authOpts utils.AuthOptions
+}
```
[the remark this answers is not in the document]
burak-ok: Added to `pull` and `push`

cmd/common/utils/helpers.go
```diff
@@ -0,0 +1,207 @@
...
+		return "", fmt.Errorf("parse url %q: %w", image, err)
+	}
+	if url.Host != "" && url.Scheme != "" {
+		return "", fmt.Errorf("no scheme allowed")
+	}
+
+	if len(image) == 0 {
+		return "", fmt.Errorf("empty image")
+	}
+	index := strings.LastIndexByte(image, ':')
```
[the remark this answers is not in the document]
burak-ok: Checked how docker handles it and it relies on the `"github.com/docker/distribution/reference"` package. So now I use it :+1:

cmd/common/image/list.go
```diff
@@ -0,0 +1,97 @@
...
+			imageColumns = append(imageColumns, &imageColumn)
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("listing all tags: %w", err)
+	}
+
+	cols := columns.MustCreateColumns[imageColumn]()
+	cols.MustSetExtractor("digest", func(i *imageColumn) string {
```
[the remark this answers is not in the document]
burak-ok: Added :)

burak-ok:
> Looking good overall. Just small comments. I am having issues testing it via complied `ig` command. (`go run -exec sudo ./cmd/ig ...` works fine):
> 

Running it with `GODEBUG=netdns=go ig image....` should work. Analysis still has to be done why.

cmd/common/image/pull.go
```diff
@@ -0,0 +1,86 @@
...
+	ociStore, err := utils.GetLocalOciStore()
+	if err != nil {
+		return fmt.Errorf("get oci store: %w", err)
+	}
+
+	repository, err := utils.GetRepositoryFromImage(o.image)
+	if err != nil {
+		return fmt.Errorf("get repository from image %q: %w", o.image, err)
+	}
+	client, err := utils.CreateAuthClient(repository, &o.authOpts)
```
mqasimsarfraz, line 63: nit: Thinking if we should skip auth when insecure used?

mqasimsarfraz (approved):
Great! Thanks for resolving all the comments. Once we take care of [this](https://github.com/inspektor-gadget/inspektor-gadget/pull/1863#pullrequestreview-1565632741) this is ready to go in!

burak-ok:
> Also, if I build with `-tags osusergo netgo static_build` ([reference](https://github.com/containerd/containerd/blob/main/Makefile#L95C27-L95C32)) it also works fine. But I am not sure if that is right solution.

The introducing commit seems to be https://github.com/containerd/containerd/commit/efc5df577382e4990c16348f4db4b55563d9177e

> When compiling containerd binaries statically, linker rightfully
> complains:
> 
> ```
> + make BUILDTAGS=static_build 'EXTRA_FLAGS=-buildmode pie'
> 'EXTRA_LDFLAGS=-extldflags "-fno-PIC -static"'
> 🇩 bin/ctr
> /tmp/go-link-343047789/000000.o: In function
> `_cgo_b0c710f30cfd_C2func_getaddrinfo':
> /tmp/go-build/net/_obj/cgo-gcc-prolog:46: warning: Using 'getaddrinfo'
> in statically linked applications requires at runtime the shared
> libraries from the glibc version used for linking
> ```
> 
> The same error appears for ctr, containerd, and containerd-stress
> binaries.
> 
> The fix is to use Go's own DNS resolver functions, rather than
> glibc's getaddrinfo() -- this option is turned on by `netgo` build
> tag.
> 
> See https://golang.org/pkg/net/ (look for "Name Resolution") for
> more details.
> 
> Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

burak-ok:
Compiling gives me the same messages:

```bash
➭ go build -ldflags "-extldflags '-static'" github.com/inspektor-gadget/inspektor-gadget/cmd/ig          
# github.com/inspektor-gadget/inspektor-gadget/cmd/ig
/usr/bin/ld: /tmp/go-link-774727284/000032.o: in function `pluginOpen':
/_/plugin/plugin_dlopen.go:19: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-774727284/000013.o: in function `mygetgrouplist':
/_/os/user/getgrouplist_unix.go:15: warning: Using 'getgrouplist' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-774727284/000012.o: in function `mygetgrgid_r':
/_/os/user/cgo_lookup_cgo.go:45: warning: Using 'getgrgid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-774727284/000012.o: in function `mygetgrnam_r':
/_/os/user/cgo_lookup_cgo.go:54: warning: Using 'getgrnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-774727284/000012.o: in function `mygetpwnam_r':
/_/os/user/cgo_lookup_cgo.go:36: warning: Using 'getpwnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-774727284/000012.o: in function `mygetpwuid_r':
/_/os/user/cgo_lookup_cgo.go:27: warning: Using 'getpwuid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-774727284/000009.o: in function `_cgo_cbcce81e6342_C2func_getaddrinfo':
/tmp/go-build/cgo-gcc-prolog:58: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

mqasimsarfraz:
Thanks for finding https://github.com/containerd/containerd/commit/efc5df577382e4990c16348f4db4b55563d9177e it. Looks good to me. Can you please create a commit with message like https://github.com/containerd/containerd/commit/efc5df577382e4990c16348f4db4b55563d9177e and then we should be good. We can always revisit it in case we have any issues.

Issue #1859, opened 2023-07-18
ig top ebpf shows empty container columns


alban:
## Description

The following columns are shown with the `ig top ebpf` command:
* RUNTIME.CONTAINERNAME
* RUNTIME.CONTAINERIMAGENAME

But they are always empty.

## Impact

Using too much space on the screen for nothing.

## Environment and steps to reproduce

```
$ sudo ig top ebpf
```

## Expected behavior

Either the columns contains useful information or they should be hidden.

## Additional information

The top ebpf gadget currently does not enrich the events with any container metadata.

cc @blanquicet @flyth

claudiamarcubina:
I see that RUNTIME.CONTAINERIMAGENAME and RUNTIME.CONTAINERNAME are shown with most of the gadgets with `ig`. E.g.
- ` ig snapshot *`
- `ig top *`
- `ig trace *`
- a.s.o.

While RUNTIME.CONTAINERNAME is populated, RUNTIME.CONTAINERIMAGENAME is always empty for now (work in progress). I am thinking to hide both fields for every gadget for `ig`. Is there currently a general way to do this, without modifying each of the gadgets?

PR #1865, opened 2023-07-18
top/ebpf: hide containerImageName column

claudiamarcubina:
# Hide containerImageName column 

- Addresses #1859

Before:
`$ ig top ebpf`

image

After:
`$ ig top ebpf`

image

### DIFF
--- a/cmd/ig/containers/containers.go
+++ b/cmd/ig/containers/containers.go
@@ -77,10 +77,12 @@ func NewListContainersCmd() *cobra.Command {
 			cols.SetExtractor("event", func(event *containercollection.PubSubEvent) string {
 				return event.Type.String()
 			})
-			// Display the runtime name and container ID when watching containers
+			// Display the runtime name, container ID and image name when watching containers
 			col, _ := cols.GetColumn("runtime.containerId")
 			col.Visible = true
 			col, _ = cols.GetColumn("runtime.runtimeName")
+			col.Visible = true
+			col, _ = cols.GetColumn("runtime.containerImageName")
 			col.Visible = true
 
 			parser, err := commonutils.NewGadgetParserWithRuntimeInfo(&commonFlags.OutputConfig, cols)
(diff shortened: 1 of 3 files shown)


mauriciovasquezbernal (approved):
LGTM from code inspection. Thanks!

mauriciovasquezbernal:
> * Addresses

nit: It's possible to automatically link PRs to close issues when they're merged -> https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

However, `addresses` is not on the keyword list.

blanquicet (changes requested):
I don't think this is the correct way to fix the issue as such a solution is only addressing the problem for the top ebpf gadget while we want the `ContainerImageName` to not be displayed in any gadget by default, see discussion [here](https://kubernetes.slack.com/archives/CSYL75LF6/p1689687086044749?thread_ts=1689686770.151899&cid=CSYL75LF6).

IMO, the `ContainerImageName` field should be managed as `ContainerID` and `RuntimeName`: 

- Hidden by default using the keyword `hide` in the [column tag](https://github.com/inspektor-gadget/inspektor-gadget/blob/df2743c1bcabbdaf18604e69b454153ab4bfcb18/pkg/types/types.go#L123).
- Manually enabled for the [`list-containers`](https://github.com/inspektor-gadget/inspektor-gadget/blob/df2743c1bcabbdaf18604e69b454153ab4bfcb18/pkg/container-collection/containers.go#L203-L208) and [`list-containers -w`](https://github.com/inspektor-gadget/inspektor-gadget/blob/df2743c1bcabbdaf18604e69b454153ab4bfcb18/cmd/ig/containers/containers.go#L80-L84) commands.

claudiamarcubina:
Moved to #1905 as it is addressing all gadgets, not only top ebpf.

Issue #1831, opened 2023-07-06
Deprecate and delete runc-fanotify in favour of container-hook


alban:
runc-fanotify and container-hook are similar and share some code but have subtle differences. When making a change in one, we often need to update the other one and we risk forgetting.

I suggest to wait until September and then delete runc-fanotify if we have not found any issue with container-hook in various environments.

alban:
Thanks @flyth for testing the ig container image (from this branch https://github.com/inspektor-gadget/inspektor-gadget/pull/1809#issuecomment-1633005357) on mac.

The container-hook package does not work on mac because of the lack of BTF: https://github.com/docker/for-mac/issues/6800.

The error message is as follow:

> Error: initializing manager: starting container fanotify: readPrivateDataFromFd: loading maps and programs: field IgFgetX: program ig_fget_x: apply CO-RE relocations: load kernel spec: no BTF found for kernel version 5.15.49-linuxkit: not supported

But runc-fanotify does not work either on mac, for different reasons (fanotify support in the kernel?), so that's not a regression.

alban:
`container-hook` requires either a kernel with `CONFIG_DEBUG_INFO_BTF`, or a kernel supported by btfhub. `runc-fanotify` does not use BTF. So it depends if we want IG to keep compatibility with kernels without BTF nor btfhub support. Since current gadgets require either BTF or btfhub support, I think it does not make sense to keep `runc-fanotify` for the sake of old kernels that can't run our current gadgets anyway.

blanquicet:
> Since current gadgets require either BTF or btfhub support, I think it does not make sense to keep runc-fanotify for the sake of old kernels that can't run our current gadgets anyway.
>
I agree.

blanquicet:
@alban, this still need to be done. I think the time we define to remove it has passed. Is there any blocker that prevents us from removing it?

alban:
No blockers from me :)

blanquicet:
Ok, so I change the name of this issue and reopened it.

PR #1867, opened 2023-07-19
Adding "/usr/local/bin/runc" to the list of potential runc paths

slashben:
# Adding /usr/local/bin/runc to the list of potential installation paths of runc

We have found an installation with contrainerd where runc was installed into the /usr/local/bin directory. This was not included in the paths where IG is looking for the runc binary though it looks like it should.

## Testing done

Verified the change on the given setup.

### DIFF
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -119,6 +119,7 @@ var runcPaths = []string{
 	"/bin/runc",
 	"/usr/bin/runc",
 	"/usr/sbin/runc",
+	"/usr/local/bin/runc",
 	"/usr/local/sbin/runc",
 	"/usr/lib/cri-o-runc/sbin/runc",
 	"/run/torcx/unpack/docker/bin/runc",
--- a/pkg/container-hook/tracer.go
+++ b/pkg/container-hook/tracer.go
@@ -142,6 +142,7 @@ var runtimePaths = []string{
 	"/bin/runc",
 	"/usr/bin/runc",
 	"/usr/sbin/runc",
+	"/usr/local/bin/runc",
 	"/usr/local/sbin/runc",
 	"/usr/lib/cri-o-runc/sbin/runc",
 	"/run/torcx/unpack/docker/bin/runc",


pkg/runcfanotify/runcfanotify.go
```diff
@@ -119,6 +119,7 @@ var runcPaths = []string{
 	"/bin/runc",
 	"/usr/bin/runc",
 	"/usr/sbin/runc",
+	"/usr/local/bin/runc",
```
mqasimsarfraz, line 122: Can you please sync the changes with [container-hook](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-hook/tracer.go#L141) as well? We will have to do it till https://github.com/inspektor-gadget/inspektor-gadget/issues/1831 is resolved.

slashben: Good catch!

alban (approved):
LGTM

/usr/local/bin is part of the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) so it is good to include it by default.

I think implementing a `find` is not practical because it could be very slow if the filesystem is large.

slashben:
> I think implementing a `find` is not practical because it could be very slow if the filesystem is large.

Agreed, what about `which runc` like functionality? (looking for `runc` in the `$PATH` directories)

alban:
> Agreed, what about `which runc` like functionality? (looking for `runc` in the `$PATH` directories)

I think using `$PATH` is not ideal: when running `ig` in a container, the `$PATH` variable is not defined by the Linux distro but by the container. And some of the paths are not in the `$PATH` variable, like `/usr/lib/cri-o-runc/sbin/`.

I think it is better to add paths as needed as you've done. We could also make it configurable if we want to support an environment where we can't hard code the path (such as a Kubernetes/containerd/runc stack recompiled and used without installing it).

slashben:
> > > I think implementing a `find` is not practical because it could be very slow if the filesystem is large.
> > 
> > 
> > Agreed, what about `which runc` like functionality? (looking for `runc` in the `$PATH` directories)
> 
> We search for `runc` under `/host`, so we would need additional tweaking to use `$PATH`.

@eiffel-fl @alban , if you think this is the direction I will work on this. Please confirm

PR #1868, opened 2023-07-19
pkg/runcfanotify: Search runc in some directories if not in runcPaths.

eiffel-fl:
If runc binary was not found in runcPaths, we will search it by walking some directories.
This is inefficient but permits Inspektor Gadget to work on distro where runc is stored in custom places.
Moreover, the idea is to then add the custom path to runcPaths by encouraging people to open a PR.

### DIFF
--- a/pkg/runcfanotify/runcfanotify.go
+++ b/pkg/runcfanotify/runcfanotify.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"math"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


pkg/runcfanotify/runcfanotify.go
```diff
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
```
blanquicet, line 22: If we decide to apply this, please sync the changes with [container-hook](https://github.com/inspektor-gadget/inspektor-gadget/blob/df2743c1bcabbdaf18604e69b454153ab4bfcb18/pkg/container-hook/tracer.go#L254) as that's the one we are currently using by default.

mauriciovasquezbernal:
Ref https://github.com/inspektor-gadget/inspektor-gadget/pull/1867

alban:
> Perhaps a middle ground if to implement an option to add paths to the list. Then this error message could be extended with:
>
> 1. Suggest to use such option to manually specify the path of the binary
> 2. Report an issue so we can add that path if it's a standard location

That seems good to me.

I wonder about using an environment variable instead of an CLI flag. That would be easier because the pkgs could be used in different programs:
1. ig
2. gadgettracermanager
3. third-party tools using pkg/runcfanotify or pkg/container-hook.

PR #1887, opened 2023-07-25
docs: add examples how to run ig in a custom pod

alban:
The ig documentation included explanations how to run it with Docker and how to run it with the "kubectl debug" command.

This patch adds another way to use ig in a custom pod. This shows an example of Dockerfile and a Kubernetes daemon set yaml. This is useful for third party software that wants to include ig.

### DIFF
--- a/docs/examples/ds-ig.yaml
+++ b/docs/examples/ds-ig.yaml
@@ -1,0 +1,51 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: ig
+  labels:
+    k8s-app: example-ig
+spec:
+  selector:
+    matchLabels:
+      name: example-ig
+  template:
+    metadata:
+      labels:
+        name: example-ig
+    spec:
+      containers:
+      - name: ig
+        # CHANGEME: use your own image
+        image: mycontainerimage:latest
+        securityContext:
+          # CHANGEME: you can also use a more restrictive securityContext
+          # See example in
+          # https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/resources/manifests/deploy.yaml
+          privileged: true
+        # CHANGEME: replace this with your own commands
+        command:
+        - "/bin/sh"
+        - "-c"
+        - "sleep inf"
+        env:
+        - name: HOST_ROOT
+          value: "/host"
+        volumeMounts:
+          - mountPath: /host
+            name: host
+          - mountPath: /run
+            name: run
+          - mountPath: /sys/kernel/debug
+            name: debugfs
+      nodeSelector:
+        kubernetes.io/os: linux
+      volumes:
+        - name: host
+          hostPath:
+            path: /
+        - name: run
+          hostPath:
+            path: /run
+        - name: debugfs
+          hostPath:
+            path: /sys/kernel/debug
--- a/docs/ig.md
+++ b/docs/ig.md
@@ -29,6 +29,8 @@ The architecture of `ig` is described in the main
   debug the containers of the cluster.
 - In some cases, you might have root SSH access to the Kubernetes nodes of a
   cluster, but not to the `kubeconfig`.
+- If you don't want to install `kubectl-gadget` on your machine, you can run
+  `ig` in a Kubernetes pod and read the output directly.
 - If you are implementing an application that needs to get insights from the
   Kubernetes node, you could include the `ig` binary in your container
   image, and your app simply execs it. In such a case, it is suggested to use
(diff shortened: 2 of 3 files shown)


docs/ig.md
```diff
@@ -222,3 +222,29 @@ List of flags:
   `--auto-mount-filesystems` to ig to automatically mount those filesystems.
 - `--pid=host` runs in the host PID namespace. Optional on Linux. This is necessary on Docker Desktop on Windows because
   /host/proc does not give access to the host processes.
+
+### Using ig in a custom Kubernetes pod
+
+If you want to include `ig` in your own container images, you take example on the following Dockerfile:
```
mauriciovasquezbernal, line 228: you take -> you can take ?

docs/ig.md
```diff
@@ -222,3 +222,29 @@ List of flags:
...
+
+```Dockerfile
+# In production, you should use a specific version of ig instead of latest:
+# --build-arg BASE_IMAGE=ghcr.io/inspektor-gadget/ig:v0.18.1
+ARG BASE_IMAGE=ghcr.io/inspektor-gadget/ig:latest
+FROM ${BASE_IMAGE} as ig
+
+# Your own image
```
mauriciovasquezbernal, line 236: What about mentioning that it should work fine in most cases but that the right libc version should be present for it to work?

docs/examples/ds-ig.yaml
```diff
@@ -0,0 +1,57 @@
...
+      name: example-ig
+  template:
+    metadata:
+      labels:
+        name: example-ig
+    spec:
+      containers:
+      - name: ig
+        # Replace this with your own image
+        image: alpine:3.17
```
mauriciovasquezbernal, line 19: Why not to use the ig image directly or mention to use the image prepared in the `Using ig in a custom Kubernetes pod` section?

docs/ig.md
```diff
@@ -222,3 +222,29 @@ List of flags:
...
+FROM alpine:3.17
+COPY --from=ig /usr/bin/ig /usr/bin/ig
+ENV HOST_ROOT=/host
+# The rest of your Dockerfile
+```
+
+If you want to run `ig` in a Kubernetes pod, you can take example on
```
mauriciovasquezbernal, line 243:
```suggestion

In order to `ig` in a Kubernetes pod use 
```

docs/examples/ds-ig.yaml
```diff
@@ -0,0 +1,57 @@
...
+          value: "/host"
+        volumeMounts:
+          - mountPath: /host
+            name: host
+          - mountPath: /run
+            name: run
+          - mountPath: /sys/kernel/debug
+            name: debugfs
+          - mountPath: /sys/fs/bpf
+            name: bpffs
```
mauriciovasquezbernal, line 42: is `/sys/fs/bpf` actually needed by `ig`? I know we need it for ig-k8s in order to [pin the mount ns map for standard tracers](https://github.com/inspektor-gadget/inspektor-gadget/blob/9d8f024001c38d5677308b8d67289c5bcaf90d43/pkg/standardgadgets/trace/standardtracerbase.go#L89), but I don't see we're pinning any object on ig.

docs/examples/ds-ig.yaml
```diff
@@ -0,0 +1,57 @@
...
+    metadata:
+      labels:
+        name: example-ig
+    spec:
+      containers:
+      - name: ig
+        # Replace this with your own image
+        image: alpine:3.17
+        securityContext:
+          privileged: true
```
mauriciovasquezbernal, line 21: Indicate it's possible to run without privileged and point to our deploy.yaml with details about capabilities needed?

docs/ig.md
```diff
@@ -222,3 +222,29 @@ List of flags:
...
+# --build-arg BASE_IMAGE=ghcr.io/inspektor-gadget/ig:v0.18.1
+ARG BASE_IMAGE=ghcr.io/inspektor-gadget/ig:latest
+FROM ${BASE_IMAGE} as ig
+
+# Your own image
+FROM alpine:3.17
+COPY --from=ig /usr/bin/ig /usr/bin/ig
+ENV HOST_ROOT=/host
+# The rest of your Dockerfile
+```
```
blanquicet, line 241: Isn't this also valid outside k8s?

docs/examples/ds-ig.yaml
blanquicet: Mmmmm I suppose it is because this example is about "how to add the capability of using `ig` from your own application", so you need to include the binary in an already existing app. Instead, using directly the ig image would be like "how to easily run `ig` on Kubernetes". It slightly different. We could have this example with the ig image by default but mention in the documentation to run [this](https://github.com/inspektor-gadget/inspektor-gadget/blob/a9c054e4a699890f75ad43b3add65cfcbf886c42/docs/examples/ds-ig.yaml#L24-L29) in case they want to use their own base image.

alban: > Why not to use the ig image directly or mention to use the image prepared in the Using ig in a custom Kubernetes pod section?

That image is not published and I wanted to show an example that an user can just use to try it out. We can't easily use the default `ig` image because it does not provide the `sleep inf` binary. We could write directly the command `ig trace exec` in the yaml, but that's a bit weird on a DaemonSet with multiple nodes because we can't directly select the node where we want to run the tracer.

To avoid confusion, I could write 2 different sections:
- how to use the existing ig container image in a Kubernetes pod. The gadget `ig trace exec` would be hard coded in the pod yaml because we don't have `sleep inf`.
- how to write your own container image with `ig`. I can give the Dockerfile and a DaemonSet with `sleep inf` and the kubectl-exec workflow.

alban:
Thanks for the reviews so far! I updated the branch.

docs/ig.md
mauriciovasquezbernal: Ah right, I usually use `go build .` on my machine and it doesn't produce a static binary, but `make ig` does. Sorry for the noise.

docs/ig.md
```diff
@@ -222,3 +222,40 @@ List of flags:
   `--auto-mount-filesystems` to ig to automatically mount those filesystems.
 - `--pid=host` runs in the host PID namespace. Optional on Linux. This is necessary on Docker Desktop on Windows because
   /host/proc does not give access to the host processes.
+
+### Using ig in a Kubernetes pod
```
mauriciovasquezbernal, line 226: I'm wondering if we can add more information on this case. We should describe what is the motivation to run `ig` and not `ig-k8s`. (I suppose it's because by using ig we can get the output directly without needing to execute kubectl-gadget, right?)

mqasimsarfraz (approved):
LGTM, just one comment!

docs/ig.md
```diff
@@ -222,3 +222,40 @@ List of flags:
...
+```bash
+$ kubectl apply -f docs/examples/pod-ig.yaml
+$ kubectl logs ig
+RUNTIME.CONTAINERNAME          RUNTIME.CONTAIN… PID              PPID             COMM             RET ARGS
+kube-proxy                     k8s.gcr.io/kube… 3985376          3024961          ip6tables        0   /usr/sbin/ip6tables -w 5 -W 100000 -S K…
+```
+
+### Adding ig in your own container image
+
+In order to add `ig` in your own container image, you can take example on the following Dockerfile:
```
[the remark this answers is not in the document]
alban: I added an explanation about multi-platform builds.

PR #1889, opened 2023-07-26
pkg/gadgets: Add unit test for trace oomkill.

eiffel-fl:


pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,161 @@
...
+	require.NoError(t, err, "writing %d to %s", childPid, procsPath)
+
+	proc, err := os.FindProcess(int(childPid))
+	require.NoError(t, err, "no process with PID %d", childPid)
+
+	// Unblock the child.
+	unix.Sendmsg(sock[0], []byte{}, []byte{}, nil, 0)
+
+	_, err = proc.Wait()
+	require.NoError(t, err, "waiting child with PID %d", childPid)
```
alban, line 136: We should do a wait with timeout:

https://github.com/golang/go/issues/9580#issuecomment-69717462

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Give some time for the tracer to capture the events
+	time.Sleep(5 * time.Second)
```
alban, line 139: We should not wait the full 5 seconds if the event arrives earlier.

The `eventCallback()` above can send a message in a channel and here we can `select{}` for either the timeout or the event.

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,161 @@
...
+	// The whole idea of this test is to spawn a child which will allocate so much
+	// memory that it get killed by the OOM killer, hence generating an event.
+	// We cannot do this in the root cgroup, as it could lead to misbehavior from
+	// other test as they would not be able to allocate memory.
+	// So, the whole idea is to create a memory cgroup, set the memory.max limit,
+	// spawn a process, subscribe this child to the memory cgroup and then
+	// allocate the memory until it get killed.
+	err = os.Mkdir(cgroupPath, 0o755)
+	require.NoError(t, err, "creating cgroup path %s", cgroupPath)
+	defer syscall.Rmdir(cgroupPath)
```
alban, line 79: I would also run a `syscall.Rmdir` at the beginning of the test to clean up previous executions of the test that were interrupted without running the defer.

alban (changes requested):

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,184 @@
...
+		case event := <-events:
+			killedPid := event.KilledPid
+			require.Equal(t, killedPid, uint32(childPid), "killedPid is %d, expected %d", killedPid, childPid)
+
+			killedComm := event.KilledComm
+			childComm := path.Base(os.Args[0])
+			require.Equal(t, killedComm, childComm, "killedComm is %s, expected %s", killedComm, childComm)
+		case <-time.After(10*time.Second):
+			t.Fatalf("waiting for OOM kill event from child %d: time out", childPid)
+	}
```
mauriciovasquezbernal, line 171: I was wondering if we can use a similar approach in other tests to avoid having the sleep, but I suppose this one works fine because there are not other OOM kill events on the host at the time of running this test. For more generic tests like open / exec this won't work. 

btw, have you thought about using `utilstest.NewRunnerWithTest` to check if the filtering logic is working fine? (Yes, I have no idea if that plays nice with this cgroup thing).

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+		case err := <-done:
+			require.NoError(t, err, "waiting child with PID %d", childPid)
+		case <-time.After(10 * time.Second):
+			t.Fatalf("waiting child with PID %d: time out", childPid)
+		}
+
+		select {
+		case event := <-events:
+			killedPid := event.KilledPid
+			require.Equal(t, killedPid, uint32(childPid), "killedPid is %d, expected %d", killedPid, childPid)
```
mauriciovasquezbernal, line 172: nit: It's not needed to print the expected and actual value. `.Equal` will print them automatically if they don't match.

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+			killedComm := event.KilledComm
+			childComm := path.Base(os.Args[0])
+			require.Equal(t, killedComm, childComm, "killedComm is %s, expected %s", killedComm, childComm)
```
mauriciovasquezbernal, line 176: We can't test any of the `Triggered` fields because it can be any process, right?

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+	utilstest.RequireRoot(t)
+
+	tracer := createTracer(t, &tracer.Config{}, func(*types.Event) {})
+
+	// Check that a double stop doesn't cause issues
+	tracer.Stop()
+	tracer.Stop()
+}
+
+func TestOomkillTracer(t *testing.T) {
```
mauriciovasquezbernal, line 58: nit: We use upper case `OOM` in the integration tests.

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+	t.Parallel()
+
+	utilstest.RequireRoot(t)
+
+	controllerPath := "/sys/fs/cgroup/cgroup.controllers"
```
mauriciovasquezbernal, line 63: nit: Perhaps these can be constants at the top of the file?

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		_, err = procsFile.Write([]byte(fmt.Sprintf("%d\n", childPid)))
+		procsFile.Close()
+		require.NoError(t, err, "writing %d to %s", childPid, procsPath)
+
+		proc, err := os.FindProcess(int(childPid))
+		require.NoError(t, err, "no process with PID %d", childPid)
+
+		// Unblock the child.
+		unix.Sendmsg(sock[0], []byte{}, []byte{}, nil, 0)
```
mauriciovasquezbernal, line 154: nit: check return value just in the remote case it's failing?

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+			require.NoError(t, err, "waiting child with PID %d", childPid)
+		case <-time.After(10 * time.Second):
+			t.Fatalf("waiting child with PID %d: time out", childPid)
+		}
+
+		select {
+		case event := <-events:
+			killedPid := event.KilledPid
+			require.Equal(t, killedPid, uint32(childPid), "killedPid is %d, expected %d", killedPid, childPid)
+
```
mauriciovasquezbernal, line 173: Check also the mount ns now that you're using NewRunnerWithTest

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,195 @@
...
+func TestOomkillTracer(t *testing.T) {
+	t.Parallel()
+
+	utilstest.RequireRoot(t)
+
+	controllerPath := "/sys/fs/cgroup/cgroup.controllers"
+	_, err := os.Stat(controllerPath)
+	if err != nil {
+		t.Skip("This test requires cgroup v2 to be mounted: %w", err)
```
mauriciovasquezbernal, line 66: Skipf and %s.

pkg/gadgets/trace/oomkill/tracer/tracer_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func TestOomkillTracer(t *testing.T) {
+	t.Parallel()
+
+	utilstest.RequireRoot(t)
+
+	_, err := os.Stat("/sys/fs/cgroup/cgroup.controllers")
+	if err != nil {
+		t.Skip("This test requires cgroup v2 to be mounted: %w", err)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Are github actions using cgroups v2?

mauriciovasquezbernal (approved):
Some nits. I couldn't test it because I don't have cgroups v2 on my machine. 

LGTM already. Perhaps you want to wait a bit for @alban's approval too.

mauriciovasquezbernal:
Unit tests are frozen for this. Should we also implement the timeout for proc.Wait() in trace signal test?

PR #1893, opened 2023-07-26
experimental: Fix detection logic

mauriciovasquezbernal:
The information from the deploy info only has to be taken into consideration when running in kubectl-gadget. When running ig, only the env variable is relevant.

This was causing an issue of enabling experimental features in ig when running on the same host where kubectl gadget deploy --experimental was used before.

```bash
$ ./kubectl-gadget deploy --experimental

$ cat $HOME/.ig/info.json  | jq '.Experimental'
true

 # IG_EXPERIMENTAL is not defined
$ sudo -E echo $IG_EXPERIMENTAL
```

Before this commit:

```
 # experimental features are wrongly enabled
$ sudo -E ./ig trace exec
...
```

After this commit:

```
 # not experimental features enabled by ig
$ sudo -E ./ig trace exec
RUNTIME.CONTAINERNAME                    RUNTIME.CONTAINERIMA… PID
PPID       COMM             RET ARGS

 # but can be enabled with IG_EXPERIMENTAL=true
$ sudo -E IG_EXPERIMENTAL=true ./ig trace exec
...
```

### DIFF
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -20,10 +20,12 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	// Import this early to set the enrivonment variable before any other package is imported
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/environment/local"
+
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/ig/containers"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/environment"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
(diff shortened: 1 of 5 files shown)


mauriciovasquezbernal:
@flyth the description of the env package says "Package environment is a temporary workaround", I know I'm somehow abusing it for this purpose, but I didn't find other solutions. What do you think about it?

mauriciovasquezbernal:
@eiffel-fl please let me know if you're able to test it, otherwise I'll merge it.

mauriciovasquezbernal:
```bash
# With this contribution
$ rm ~/.ig/info.json
$ ./kubectl-gadget deploy --experimental --image-pull-policy=Never
...
$ cat $HOME/.ig/info.json  | jq '.Experimental'
cat: /home/francis/.ig/info.json: Aucun fichier ou dossier de ce type
```

This is weird, the info.json file should be created by the deploy command. I tried here and it works as expected. Could you check this up?

mauriciovasquezbernal:
```bash
$ rm $HOME/.ig/info.json                                                             
$ ./kubectl-gadget trace exec  
```

We don't expect the user to remove the info file after deploying Inspektor Gadget. Then, I think it's fine not to print that message.

PR #1901, opened 2023-07-27
gadgets/run: Fix handling of --host parameter

mauriciovasquezbernal:
Avoid crashing when the --host parameter is passed as in that case the mount ns filter map is not created.

Fixes: 07ca0fb4e97a ("Initial support for containerized gadgets")


burak-ok (approved):
Thanks for the quick fix for this. LGTM

PR #1903, opened 2023-07-28
ci: Use secrets through env variables when accessed in run blocks.

eiffel-fl:
The GitHub actions documentation advices to store secrets in env variables when they are accessed in run block [1].
> To make a secret available to an action, you must set the secret as an input
> or environment variable in the workflow file.

Fixes: 702f651115cd ("ci: Run integration tests on ARO")
Fixes: f8117bcc669a ("ci: Run test on AKS clusters.")
Fixes: 320de7dca9d2 ("ci: Sign checksums file using cosign.")
[1]: https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets


burak-ok (approved):
Thanks for the changes, LGTM

PR #1905, opened 2023-07-31
Hide `ContainerImageName` column by default

claudiamarcubina:
# Hide `ContainerImageName` column by default

`ContainerImageName` column was displayed for most of the gadgets taking unnecessary space on the screen.

It is now hidden for all gadgets and only displayed for `ig list-containers` and `ig list-containers -w`.

Fixes: #1859


blanquicet (approved):
LGTM! Thanks for handling this.

PR #1906, opened 2023-07-31
pkg/gadgets: Add timeout waiting for child process in trace signal unit test

eiffel-fl:
Let's see if this avoid unit test to time out.
Otherwise, I will just revert the whole test.


mauriciovasquezbernal (approved):
LGTM. Let's merge it and see if the testing deadlock is fixed.

PR #1909, opened 2023-08-01
Revert trace signal unit test

eiffel-fl:
Hi.

I suspect this test to makes our CI timed out (even though they should be skipped, since the CI is not run as root).
Let's see if reverting everything helps there, otherwise further investigation would be needed.

Best regards.


mauriciovasquezbernal:
> Then, something is really wrong:

That's printed when running the unit-tests without root in https://github.com/inspektor-gadget/inspektor-gadget/blob/baaff2ae7ac51a1fb3e0b739ba3ae32b356a155f/.github/workflows/inspektor-gadget.yml#L685-L686

I agree, we should probably change that to run them only once.

mauriciovasquezbernal (approved):
I triggered the workflow multiple times and it appears to be working fine. Let's merge it and then we can investigate what's going on.

PR #1915, opened 2023-08-03
integration: Pass testing.T object to helper functions

mauriciovasquezbernal:
Pass the testing object directly to the helper functions to allow them to fail the tests. This avoids all the error check logic on the callers.

ref https://github.com/inspektor-gadget/inspektor-gadget/pull/1866#discussion_r1281952242

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -26,6 +26,7 @@ import (
 	"testing"
 
 	"github.com/kr/pretty"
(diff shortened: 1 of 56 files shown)


burak-ok (approved):
Thank you for this :100: This helps a lot

And I forgot to pass `testing.T` into all functions in `run_helper.go` :see_no_evil:

PR #1938, opened 2023-08-07
Decreasing the size of trace/open and trace/exec hash maps

slashben:
# Decreasing the size of the hash maps in trace/open and trace/exec to 10%

These two gadgets have been using hash maps in the size of 10240 for storing data in between enter and exit phases of the system's respective system calls. Due to the increase in record size (both of them store a file path which was increased to `PATH_MAX`-> 4096 bytes) and the aforementioned multiplier the gadget allocated ~40Mb of RAM. 

Neither of these system calls is blocking nor do they seem to take a long time to execute and due to this, it is not justified to allocate 10240 map entries for 10240 parallel system calls.

Based on our tests a heavily loaded machine with file-intensive load is not getting beyond 3(!) concurrent open system calls. Therefore lowering the allocation to 10% seems to be a safe decision.


alban (approved):
LGTM!

Given that both syscalls normally don't block, I would also be ok to reduce it further if necessary.

I can merge this PR once the commit message is fixed as suggested by @eiffel-fl

alban:
Side note: a shell script can easily fill the `start` map in the `trace open` gadget in this way:

```
#!/bin/bash
max=5000
for i in $(seq 1 $max) ; do
  rm -f ./pipe-$i
  mkfifo ./pipe-$i
  cat ./pipe-$i &
done
```
Leftover processes can be removed with `ps aux|grep [c]at.*pipe|awk '{print $2}'|xargs kill`.

Since nobody is writing to the pipe files, the reader side is blocked in the `openat` syscall.

But this is a general issue unrelated to this PR. If we want to isolate one container from another, we could use a limit of parallel syscall per container.

mauriciovasquezbernal (approved):
LGTM. Please fix the commit message in order to merge it

mauriciovasquezbernal:
Commit message is just a detail, I'll move forward and merge this to have this available for the next release. Thanks a lot for the contribution!

slashben:
Hey!

> Can you please share more insights regarding this test?

We were testing a Kubernetes node with 4 CPUs running a Elaticsearch data node with ~80% CPU average load for an hour. We were testing how many concurrent `open` system-calls are happening.

> One good middle ground would be to make this value configurable or potentially tied to the number of CPU the host has, something like `X * $(nproc)` where X is still to be determined. Indeed, I guess the higher the number of core, the higher the memory (I do not really imagine having a 96 CPU cores machine with only 7GB of DRAM).

My take on this. In principle like the idea of scaling these parameters with CPU count of the node, but... 

The only concern is the scaling is the higher layers. When defining a Kubernetes DaemonSet, there is no way to allocate CPU/Memory as function to the number of CPUs or size of Memory. It will require the users to define or adjust the CPU/Memory allocations at the workload level manually.

This makes me think, that it might be better to control both the IG and workload level allocations at the same level otherwise it will be really hard to tune the system.

slashben:
> Thank you for the precision! But how do you stress the Elasticsearch? With a load testing tool like like [gattling](https://gatling.io/) or [JMeter](https://jmeter.apache.org/)?

Neither 😞 It is the load on our dev system. 

> I am not sure that I understand totally. The idea would be to add some flags to make it possible to configure Inspektor Gadget memory usage? 

Yes

> But then, how would Inspektor Gadget memory relates to the workload? I mean, you deploy Inspektor Gadget and the workload independently.

I mean the following. In our use case, we have IG as part of our workload. I want to be able to control all allocations (both what IG allocates, what we allocate and the workload level requests/limits) in one place.

PR #1940, opened 2023-08-07
ci: Make AKS cluster name unique.

eiffel-fl:
This avoids issues where we try to create an already existing cluster.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1025,7 +1025,7 @@ jobs:
     - name: Craft cluster name
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


mqasimsarfraz:
The change is looking good. My only concern is how would we ensure there isn't any dangling clusters? Also, do we have a run against `citest` already?

mqasimsarfraz:
> > My only concern is how would we ensure there isn't any dangling clusters?
> 
> The CI step to remove it always executed. So, no dangling ones.
> 

Makes sense. I was also thinking for scenarios if the deletion job itself has issues (downtime from Azure/GH) but perhaps we can check CI resource group in future if that even happens.

PR #1941, opened 2023-08-07
Dockerfiles: Improve cross building.

eiffel-fl:
Previously to this commit, it was possible to cross build the images from amd64 to arm64.
With this commit, cross building can now be done in any direction:
1. From amd64 to arm64.
2. From arm64 to amd64.


burak-ok (approved):
Looks good.

One comment is that `ig-tests.Dockerfile` also has a condition based on `arm64`. So maybe there we should also support `arm64`->`amd64` compiling aswell to be aligned?

PR #1942, opened 2023-08-07
feat: Add namespace overrides to Helm chart

terramauthe:
# Add namespace overrides to Helm chart

Allows setting a `namespaceOverride` on the helm chart

## How to use

- Generate helm chart from this
- Deploy helm chart with `namespaceOverride` set in the values file
- Observe namespaces are overridden

## Testing done

- Followed the abovementioned steps


terramauthe:
It seems like the namespace and DaemonSet pod names are hardcoded - I'll close this for now. I likely won't have time soon to make this configurable. :)

PR #1944, opened 2023-08-08
pkg/gadgets: Add tracing of failed exec calls to trace exec.

eiffel-fl:
Hi.

In this PR, I added `--failed` which permits tracing failed `exec()` calls:

```bash
$ sudo ig trace exec --host
RUNTIME.CONTAINERNAME                                   PID        PPID       COMM             RET ARGS
                                                        75772      4318       zsh              -2  /usr/local/sbin/foobar                                                  
                                                        75772      4318       zsh              -2  /usr/local/bin/foobar                                                   
                                                        75772      4318       zsh              -2  /usr/sbin/foobar                                                        
                                                        75772      4318       zsh              -2  /usr/bin/foobar                                                         
                                                        75772      4318       zsh              -2  /sbin/foobar                                                            
                                                        75772      4318       zsh              -2  /bin/foobar                                                             
                                                        75772      4318       zsh              -2  /usr/games/foobar                                                       
                                                        75772      4318       zsh              -2  /usr/local/games/foobar                                                 
                                                        75772      4318       zsh              -2  /snap/bin/foobar
# The same with --ignore-errors
$ sudo ig trace exec --host --ignore-errors
RUNTIME.CONTAINERNAME                                   PID        PPID       COMM             RET ARGS
```

Best regards.

### DIFF
--- a/pkg/gadgets/trace/exec/tracer/tracer_test.go
+++ b/pkg/gadgets/trace/exec/tracer/tracer_test.go
@@ -195,6 +195,31 @@ func TestExecTracer(t *testing.T) {
 				require.Equal(t, events[0].Cwd, cwd, "Event has bad cwd")
 			},
 		},
+		"event_failed": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *tracer.Config {
+				return &tracer.Config{
+					MountnsMap: utilstest.CreateMntNsFilterMap(t, info.MountNsID),
+				}
+			},
+			generateEvent: generateFailedEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, _ int, events []types.Event) {
+				require.Len(t, events, 1, "One event expected")
+				require.Equal(t, []string{"/bin/foobar"}, events[0].Args, "Event has bad args")
+				require.NotEqual(t, int(0), events[0].Retval, "Event returns 0, while it should return an error code")
+			},
+		},
+		"event_failed_ignored": {
+			getTracerConfig: func(info *utilstest.RunnerInfo) *tracer.Config {
+				return &tracer.Config{
+					MountnsMap:   utilstest.CreateMntNsFilterMap(t, info.MountNsID),
+					IgnoreErrors: true,
+				}
+			},
+			generateEvent: generateFailedEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, _ int, events []types.Event) {
+				require.Len(t, events, 0, "Zero events expected")
+			},
+		},
 	} {
 		test := test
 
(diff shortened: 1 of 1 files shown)


pkg/gadgets/trace/exec/tracer/tracer_test.go
```diff
@@ -195,6 +195,25 @@ func TestExecTracer(t *testing.T) {
...
+				}
+			},
+			generateEvent: func() (int, error) {
+				cmd := exec.Command("/bin/foobar")
+				cmd.Run()
+
+				return 0, nil
+			},
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, _ int, events []types.Event) {
+				fmt.Printf("events: %v\n", events)
```
blanquicet, line 212: Leftover during development?

```suggestion

```

pkg/gadgets/trace/exec/tracer/tracer_test.go
```diff
@@ -195,6 +195,25 @@ func TestExecTracer(t *testing.T) {
...
+				require.Len(t, events, 1, "One event expected")
+				require.Equal(t, []string{"/bin/foobar"}, events[0].Args, "Event has bad args")
```
blanquicet, line 214: Shouldn't we test that the return value was actually set? For instance:

```go
				require.NotEqual(t, int(0), events[0].Retval, "Event has bad retval")
```

Or we could directly verify it is `-2` as we know that's the error we are triggering. I'm fine with both solutions, maybe checking `0` is more generic.

blanquicet (commented):
A few comments in the tests, but it looks good already.

pkg/gadgets/trace/exec/tracer/gadget.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	ParamCwd = "cwd"
+	ParamCwd    = "cwd"
+	ParamFailed = "failed"
```
mauriciovasquezbernal, line 27: We have a similar `IgnoreErrors` param in trace bind https://github.com/inspektor-gadget/inspektor-gadget/blob/9a0941298c639e4a52d1d98b9c69b1cb603f113a/pkg/gadgets/trace/bind/tracer/gadget.go#L28

Should we name both of them the same?

mauriciovasquezbernal: IMO it's better to stick to the same pattern all around Inspektor Gadget. I'm ok with changing `ignore-errors` to `failed` in trace bind, but those should be aligned, otherwise it's difficult for users.

Regarding the default value, should we ignore failed operations by default? IMO they are relevant and could help to diagnose an issue, so I think we should show them by default. If they're creating too much noise, the user can disable them.

mauriciovasquezbernal (approved):
Some minor nits around. LGTM otherwise.

PR #1945, opened 2023-08-08
ci: Remove workaround to test 'cri-o' in GH actions

mqasimsarfraz:
There was an issue running `minikube` with `--container-runtime=cri-o` in GitHub actions due to docker version older than `v0.23.0` https://github.com/kubernetes/minikube/issues/12928#issuecomment-1245458044. GitHub actions now use [newer version](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/5794995680/job/15705708009#step:4:18) so it should be fine to remove the workaround.


mqasimsarfraz:
Thanks for the review!

Issue #1784, opened 2023-06-26
[RFE] Use ig for systemd units


alban:
## Current situation

`ig` can be used to get events from containers and soon to get events from the host (https://github.com/inspektor-gadget/inspektor-gadget/pull/1741).

## Impact

When events are coming from the host, they don't specify from which systemd unit they come from. It is also not possible to filter events by systemd unit.

## Ideal future situation

```
$ sudo ig trace open --host --show-systemd --sdunit=sshd
SDUNIT     PID        COMM         FD    ERR PATH
sshd       2677903    sshd         4     0   /etc/nsswitch.conf                                           
sshd       2677903    sshd         4     0   /etc/ld.so.cache                                             
```

(CLI flags to be discussed)

## Implementation options

- Maintain a map of type BPF_MAP_TYPE_CGROUP_ARRAY with the cgroup-v2 of systemd units (Linux >= 4.8)
- Use helper bpf_current_task_under_cgroup (Linux >= 4.9)
- Add CLI flags to select systemd units

## Additional information

- bpftrace has a builtin function cgroup_path (https://github.com/iovisor/bpftrace/pull/2055) but it is executed in userspace.

Issue #1403, opened 2023-03-07
[RFE] Allow filtering by Kubernetes resources in `ig`


blanquicet:
## Current situation

After implementing https://github.com/inspektor-gadget/inspektor-gadget/issues/737, we will have the Kubernetes columns but we are not yet able to filter by Kubernetes pod, namespace or container name (only container name given by runtimes).

## Impact

If `ig` is used to debug a Kuberentes cluster, it is not possible to filter by those resources.

## Ideal future situation

To allow doing this:

```bash
$ sudo ig trace exec --k8s-namespace myPod
```

```bash
$ sudo ig trace exec --k8s-pod myPod
```

Not sure how this one should be as kubectl-gadget is `containername` but for `ig` we are using it for runtime container name:
```bash
$ sudo ig trace exec --k8s-containername myPod
```

mqasimsarfraz:
> After implementing https://github.com/inspektor-gadget/inspektor-gadget/issues/737, we will have the Kubernetes columns but we are not yet able to filter by Kubernetes pod, namespace or container name (only container name given by runtimes).

#737 has been completed which allows filtering by columns. For example, we can filter by Kubernetes namespace columns as:

```
$ ig trace open -F k8s.namespace:kube-system 
RUNTIME.CONTAINERNAME                                                              PID        COMM             FD       ERR PATH                                                                                   
k8s_coredns_coredns-6d4b75cb6d-hwt7m_kube-system_25f06dfc-fb45-4486-8cb8-292ab643… 13672      coredns          13       0   /etc/hosts
```

Having said that, we still need to support filtering by [`containercollection` ](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/localmanager/localmanager.go#L273) and only open topic is deciding how should we name the params for such filtering:

- Namespace: `--namespace (-n)` vs `--k8s-namespace`
- Pod: `--podname (-p)` vs `--k8s-podname`
- Container `--containername (-c)` vs `--k8s-containername`

In general, I like having flags like `-n`/`-p`/`-c` for such filtering but it can be confusing so the above option sound OK to me!

mqasimsarfraz:
@blanquicet @mauriciovasquezbernal thoughts?

alban:
[...]

PR #1960, opened 2023-08-16
ig: Add systemd filtering and enrichment

burak-ok:
# ig: Add systemd filtering and enrichment

When using `ig --host` we now can add additional an flag (`--show-systemd`) to only show events from `systemd` units. Additionally the column `SDNAME` appears and gets enriched by a new operator:
```bash
$ sudo ig  trace open --host --show-systemd   
RUNTIME.CONTAINERNAME          SDNAME                            PID        COMM             FD    ERR PATH                                                
                               systemd-oomd.service              1520       systemd-oomd     7     0   /sys/fs/cgroup/user.slice/user-1000.slice/user@1000…
                               systemd-oomd.service              1520       systemd-oomd     7     0   /sys/fs/cgroup/user.slice/user-1000.slice/user@1000…
                               systemd-oomd.service              1520       systemd-oomd     7     0   /proc/meminfo                                       
                               systemd-journald.service          539        systemd-journal  0     2   /run/log/journal/3af0e0c3288c4974a74422d795e028d4/s…
                               systemd-journald.service          539        systemd-journal  0     2   /run/log/journal/3af0e0c3288c4974a74422d795e028d4/s…
```

Another additional flag (`--sdunit sshd.service`) will only show events from that specific `systemd` unit:
```bash
$ sudo ig trace open --host --show-systemd --sdunit mdatp.service
RUNTIME.CONTAINERNAME          SDNAME                            PID        COMM             FD    ERR PATH                                                
                               mdatp.service                     2111       tvm_agent        17    0   /proc/self/smaps                                    
                               mdatp.service                     2079       telemetryd_v2    19    0   /etc/hosts                                          
                               mdatp.service                     1699       wdavdaemon       10    0   /proc/2082384/stat                                  
                               mdatp.service                     1699       wdavdaemon       10    0   /proc/2082384/cmdline 
```

Solves #1784

## Implementation

If `--show-systemd` is not set neither the enricher or filtering will be started and no overhead will incur

### eBPF cgroup filtering
There is a ebpf helper function called `bpf_current_task_under_cgroup`. The function checks if the current process is running under that cgroup or any of its childern. But it requires the usage of `BPF_MAP_TYPE_CGROUP_ARRAY` which is (unfortunatly) an array instead of a hash map. Therefore when trying to filter on specific cgroupids the eBPF program needs to iterate over the whole array. This performance hit is too large for us (not measured!)

This PR uses a normal `BPF_MAP_TYPE_HASH` and a simple check if the current tasks' cgroup id is present in the hash map. To support children of the cgroup the userspace has to put their own cgroup ids into said hash map. The advantage is the fast lookup for the filtering.

###  `SystemdCache`

For enriching the `SDNAME` column and keeping the filter hash map up to date the `SystemdCache` object was created. It is a Singleton and somewhat analogous to the `K8sInventoryCache`.

Every second (configurable) the cgroup ids of all active systemd units gets queried and updated in its internal map.
- For querying on the userspace the function `GetSystemdUnit(fd uint64) (string, bool)` is provided (used by the enricher)
- To always have an up to date view of all active systemd units a pub/sub mechanism was introduced. An object with two methods (`AddCgroup(string, uint64)` and `RemoveCgroup(string, uint64)` can subscribe on the cache. It will receive all Add/Remove events for cgroup ids with the systemd unit name. In order to have a complete view over all active systemd units the object can request that "all old entries" (So every systemd unit which got added before the object subscribed) will be sent to it. This can be done if `publishOldEntries` is set to true when subscribing using `Subscribe(sub SystemdSubInterface, publishOldEntries bool)`

## TODO
- [x] Add enable/disable flag
- [x] Implement setting the filter
- [x] Find right place for creating eBPF map and replacing it. Really done?
- [ ] Incoorperate into container-collection
- [ ] Optional: Set map size to 1 if cgroup filtering is not used to save memory


pkg/operators/systemdresolver/systemdresolver.go
```diff
@@ -0,0 +1,127 @@
...
+type SystemdResolver struct {
+	systemdCache *common.SystemdCache
+}
+
+func (k *SystemdResolver) Name() string {
+	return OperatorName
+}
+
+func (k *SystemdResolver) Description() string {
+	return "SystemdResolver resolves cgroup ids to systemd names"
```
alban, line 46:
```suggestion
	return "SystemdResolver resolves cgroup ids to systemd unit names"
```

alban (commented):
Thanks, that's great start

burak-ok:
I don't know if this is the right place and way to do this.

I have added a flag to `localmanager` called `--show-systemd`.
In order to use the columns feature `columnsTags:"params:show-systemd"` I had to modify `registry,go`
Furthermore the `systemdresolver` operator queries this flag too and doesn't run.

So with the latest commit nothing gets queried through `dbus`

pkg/operators/common/systemdcache.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+		fullpath, err := cgroups.CgroupPathV2AddMountpoint(path)
+		if err != nil {
+			continue
+		}
+		id, err := cgroups.GetCgroupID(fullpath)
+		if err != nil {
+			continue
+		}
+		// Check internal map first to lessen contention on `idToUnitname sync.Map`
```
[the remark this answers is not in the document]
burak-ok: No numbers - just gut feeling.
None the less that specific check is not there anymore in a further commit. I forgot to add the patch to that commit - sorry for that

pkg/operators/localmanager/localmanager.go
```diff
@@ -293,6 +301,11 @@ func (l *localManagerTrace) PreGadgetRun() error {
 	id := uuid.New()
 	l.subscriptionKey = id.String()
 	host := l.params.Get(Host).AsBool()
+	systemd := l.params.Get(Systemd).AsBool()
+
+	if !host && systemd {
+		return fmt.Errorf("flag --%s can only be used with --%s", Systemd, Host)
+	}
```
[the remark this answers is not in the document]
burak-ok: I also thought about making both implicit - But I wanted to present it here for discussion. Also like the implication more

pkg/gadgets/audit/seccomp/tracer/tracer.go
```diff
@@ -78,7 +78,7 @@ func (t *Tracer) install() error {
 		return fmt.Errorf("loading ebpf program: %w", err)
```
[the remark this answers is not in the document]
burak-ok: No, but I should definitely look into that :D

pkg/operators/localmanager/localmanager.go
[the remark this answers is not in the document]
burak-ok: `--show-systemd` now implies `--host`.
But I can't make it so that `--sdunit sshd.service` implies `--show-systemd` because of the automatic column filtering I do here. It is required that the param is set to true

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -62,6 +64,11 @@ static __always_inline bool trace_allowed(u32 tgid, u32 pid)
 	if (gadget_should_discard_mntns_id(mntns_id))
 		return false;
 
+	cgroup_id = bpf_get_current_cgroup_id();
+
+	if (gadget_should_discard_cgroup_id(cgroup_id))
+		return false;
+
```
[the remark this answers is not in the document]
burak-ok: (I don't exactly know so this answer is more of a question)
Do we want to have both filterings for all gadgets?
Shouldn't they be able to also put in a `cgroup_id` (I don't even know if this makes sense)

burak-ok:
I change the commits a little bit around and cleaned them up so they are more concise. Especially now we can see how we can add `systemd` enrichment and/or `systemd` filtering to another gadget.


> I think we can expect it regarding the code shape at the moment but it explodes with another gadget:

Seems that I didn't add a check - Just did that. Now someone would get an error if he wants to filter `systemd` events on a gadget that doesn't support it (yet).
This shows another shortcoming of this implementation: The flags will be shown for all gadgets even though they are not supported for some gadgets.
This could probably be solved if I move the filtering to another `operator`. But then how could I set the `--host` parameter of `localmanager` when I am in another operator.

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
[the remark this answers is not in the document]
burak-ok: I meant more like this:
You proposed:
```c
inline bool gadget_should_filter(void)
{
	return gadget_should_discard_mntns_id(gadget_get_mntns_id()) 
	       || gadget_should_discard_cgroup_id(bpf_get_current_cgroup_id()));
}
```

What if a gadget wants to check for another `cgroup_id` (For any reason - I don't know if that makes sense). Therefore we would need something like this
```c
inline bool gadget_should_filter(__u64 mntns_id, __u64 cgroup_id)
{
	return gadget_should_discard_mntns_id(mntns_id) 
	       || gadget_should_discard_cgroup_id(cgroup_id);
}
```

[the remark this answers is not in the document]
burak-ok: It is. But then this comes at the cost of having to update all gadgets again if we introduce a new filtering method.

With your proposal without any arguments we don't need to change the old gadgets at all if we introduce yet another filtering method.

That is why I don't know what to do :D

pkg/operators/common/systemdcache.go
```diff
@@ -0,0 +1,219 @@
...
+	systemdOnce.Do(func() {
+		systemdCache, err = newSystemdCache(1 * time.Second)
+	})
+	return systemdCache, err
+}
+
+func (cache *SystemdCache) Subscribe(sub SystemdSubInterface, publishOldEntries bool) {
+	logrus.Debugf("Subscribing to systemd cache")
+	cache.subMutex.Lock()
+	cache.subscribers = append(cache.subscribers, sub)
```
blanquicet, line 65: What about verifying that the new subscriber wasn't already subscribed, and in case, print an info message but simply go ahead? Maybe using a map instead of a slice will do the trick. Making that change will also simplify `Unsubscribe` implementation.

pkg/operators/common/systemdcache.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+func GetSystemdCache() (*SystemdCache, error) {
+	systemdOnce.Do(func() {
+		systemdCache, err = newSystemdCache(1 * time.Second)
+	})
+	return systemdCache, err
+}
+
+func (cache *SystemdCache) Subscribe(sub SystemdSubInterface, publishOldEntries bool) {
```
blanquicet, line 62: What about making `Subscribe` return the list of the current entries as we do for the [container-collection](https://github.com/inspektor-gadget/inspektor-gadget/blob/4ee22a5d28610cf664896b650936e4f87936d992/pkg/container-collection/container-collection.go#L478)?

pkg/operators/common/systemdcache.go
```diff
@@ -0,0 +1,219 @@
...
+func (cache *SystemdCache) publishRemove(unitName string, id uint64) {
+	cache.subMutex.Lock()
+	defer cache.subMutex.Unlock()
+	for _, updater := range cache.subscribers {
+		updater.RemoveCgroup(unitName, id)
+	}
+}
+
+func (cache *SystemdCache) update() {
+	units, err := cache.dbusConn.ListUnitsByPatternsContext(context.TODO(), []string{"active"}, []string{})
```
blanquicet, line 166: Is it really necessary to ask the whole list every time? Couldn't we subscribe to the changes and forward the updates instead of doing it manually?

pkg/operators/common/systemdcache.go
```diff
@@ -0,0 +1,219 @@
...
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cgroups"
+)
+
+type SystemdSubInterface interface {
+	AddCgroup(string, uint64)
+	RemoveCgroup(string, uint64)
+}
+
+type SystemdCache struct {
```
blanquicet, line 34: IIUC, this `SystemdCache` is a unique cache that keeps an up-to-date list of systemd units and is used by two operators for different purposes:

1. The systemd resolver: Enrich the events with the systemd unit names. The cache returns the systemd unit name from the CgroupID. 
2. The local manager: Keep the `CgroupMap` up-to-date with the existing systemd units. The `CgroupMap` is a map BPF programs use to filter by cgroup when `--sdunit` is passed.

Is my understanding correct?

For the `--sdunit` flag, we need to clarify that the very first events of the requested unit might not be captured as the notification of the new systemd unit is not synchronous as the container-hook does for the containers.

pkg/gadgets/trace/open/tracer/tracer_test.go
```diff
@@ -115,6 +115,11 @@ func TestOpenTracer(t *testing.T) {
 		validateEvent   func(*testing.T, *utilstest.RunnerInfo, int, []types.Event)
 	}
 
+	normalize_cgroupid := func(e *types.Event) {
+		e.CgroupID = 0
+		e.Sdname = ""
```
blanquicet, line 120: Shouldn't sdname be always empty? It is supposed to be enriched by the systemdresolver operator and here it is not being used. Did I miss something?

pkg/gadgets/trace/open/tracer/tracer_test.go
```diff
@@ -115,6 +115,11 @@ func TestOpenTracer(t *testing.T) {
 		validateEvent   func(*testing.T, *utilstest.RunnerInfo, int, []types.Event)
 	}
 
+	normalize_cgroupid := func(e *types.Event) {
+		e.CgroupID = 0
```
blanquicet, line 119: Out of curiosity, is adding the CgroupID to the `RunnerInfo` struct that complex?

cc @mauriciovasquezbernal @alban

blanquicet (commented):
I mainly reviewed the first commits and I focused on the cache. It took me more than I expected as I was missing some knowledge for cgroups and systemd, so I might have used wrong terminology in my comments, if so, please let me know 🙂

alban:
Thanks @burak-ok for the PR!

Some notes on the general architecture:

Instead of using the systemd D-Bus API to get the list of systemd units, we could use fanotify on the cgroup2 filesystem (in /sys/fs/cgroup or /sys/fs/cgroup/unified depending on the cgroup configuration).

Then, it will be possible to make the events synchronous (and to be able to handle early events) by using FAN_ACCESS_PERM (like we do in pkg/container-hook).

IG needs to get the cgroup id. Currently, this is done with `name_to_handle_at()`. In Linux 6.6 (unreleased yet), it will be possible to get the cgroup id from the fanotify event:
* https://github.com/torvalds/linux/commit/0ce7c12e88cfa3e4cb0800e8172e5befea5c199a
* https://lore.kernel.org/lkml/CABWYdi39+TJd1qV3nWs_eYc7XMC0RvxG22ihfq7rzuPaNvn1cQ@mail.gmail.com/T/

This kernel patch was implemented for the needs of https://github.com/cloudflare/ebpf_exporter in order to get the cgroup id for short-lived cgroups. In the case of Inspektor Gadget, we could continue to use FAN_ACCESS_PERM and that should be enough to support short-lived cgroups. So we don't need to wait for Linux 6.6.

Thanks @Werkov for pointing me to the kernel patches mentioned above.

alban:
PoC using fanotify to detect new cgroups synchronously:
https://github.com/alban/inspektor-gadget/commit/1c704c01bd08d738a1db7fa694c36f45a65bbc4c

pkg/operators/common/systemdcache.go
```diff
@@ -0,0 +1,219 @@
...
+
+func (cache *SystemdCache) Subscribe(sub SystemdSubInterface, publishOldEntries bool) {
+	logrus.Debugf("Subscribing to systemd cache")
+	cache.subMutex.Lock()
+	cache.subscribers = append(cache.subscribers, sub)
+	cache.subMutex.Unlock()
+
+	// Publish old entries after subscribing
+	// This ensures that the subscriber always has all entries that it needs
+	// It may recieve some entries twice because of that
```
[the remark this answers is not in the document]
burak-ok: 1. User subscribes to the systemdcache (now cgroup-hook) and wants the old entries aswell
2. Before it gets the old entries it gets interrupted
3. A new systemd service starts
3.1 User will be notified of this new systemd service
3.2 The systemd service will be entered into the currently known entries list
4. Now the interrupted process from 2. gets resumed
5. The code iterates through all currently known entries and publishes to the new subscriber (including the already published new systemd service

The same problem will happen when we return a list of the current systemd entries instead of this "republish" old entries

pkg/operators/common/systemdcache.go
burak-ok: I like this approach more since we only need to write this code once in the systemdcache (now cgroup-hook). Else we need to handle the list that we get back in every code place where we use the systemdcache/cgroup-hook

pkg/operators/common/systemdcache.go
burak-ok: Is now synchronous. Thanks to you and @alban :)

burak-ok:
> PoC using fanotify to detect new cgroups synchronously: [alban@1c704c0](https://github.com/alban/inspektor-gadget/commit/1c704c01bd08d738a1db7fa694c36f45a65bbc4c)

Thanks to your PoC we now synchronously get notified - and therefore miss no events at all :+1: 

For the deletion I didn't find any way to get notified synchronously. For this PR now I just implemented a simple mechanism which does a `stat` on the cgroup dir and removes the entry if the stat failed

pkg/gadgets/trace/open/tracer/tracer_test.go
alban: As opposed to cgroup-v1, cgroup-v2 are normally per process and not per thread (except with the "thread mode", see `man 7 cgroups`, but that's too complex for here). Since our testing framework is just using one thread for the workload, testing the cgroup-v2 id here would not work.

burak-ok:
> * [ ]  Add error message for `--sdunit=foo` when unsupported by the gadget. For now, only `--show-systemd` shows the error:

Done
> * [ ]  Add support for containerized gadgets

I think I would prefer it in a separate PR. WDYT?

> * Rename flag to `--sd-unit` (discussion in [[RFE] Allow filtering by Kubernetes resources in `ig` #1403 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/issues/1403#issuecomment-1757575985)).

Done

> * Move `--show-systemd` to `-o +systemd` group of columns (see [GetKubernetesColumns](https://github.com/inspektor-gadget/inspektor-gadget/blob/bd979c651eea8009b191c6aa7755448de3130180/cmd/common/utils/parser.go#L122)). See also [Filter events at ebpf level #2068](https://github.com/inspektor-gadget/inspektor-gadget/issues/2068)

I don't know about this one. I have no problem about using `-o +systemd` to enable the column, but then we need another flag to control if we want the systemd enrichment at all. Or do we want to enable it always?
Enabled or Disabled by default?

burak-ok:
I would add support for image based gadgets in another PR. Especially only after the big refactoring PR goes through.
WDYT?

Issue #1849, opened 2023-07-13
[RFE] ig: Support other containerd namespaces than "k8s.io" 


burak-ok:
## Current situation

Our containerd client currently defaults to the containerd namespace `k8s.io`

https://github.com/inspektor-gadget/inspektor-gadget/blob/dbc41928d25185aa87c26314c7a72420d32c1683/pkg/container-utils/containerd/containerd.go#L70

## Impact

Other containerd containers are not shown and not traced

## Ideal future situation

1. It should show all namespaces by default
2. Give an option to filter by that namespace

## Implementation options

We would need to query all namespaces and then query its containers -> More API calls

When enriching newly created containerd containers we might have to scan all namespaces again and search the newly created container in all of them

PR #1961, opened 2023-08-16
container-utils/containerd: Add support for custom namespace

mqasimsarfraz:
Currently we only support initial containers in `k8s.io` namespace for containerd. This change add supports for a `--containerd-namespace` flag to allow selecting custom namespace. The main use case of this change is to allow selecting "default" namespace, used by containers created using `nerdctl/ctr`.

Fixes: #1849 

### Testing Done

```
$ sudo ctr image pull docker.io/library/nginx:latest
$ sudo ctr c create docker.io/library/nginx:latest test-container
$ sudo ctr t start test-container -d
$ sudo ./ig list-containers --containerd-namespace default
RUNTIME.RUNTIMENAME RUNTIME.CONTAINERID RUNTIME.CONTAINERNAME RUNTIME.CONTAINERIMAGENAME    
containerd          test-container      test-container        docker.io/library/nginx:latest
$ sudo ./ig snapshot process --containerd-namespace default
RUNTIME.CONTAINE… COMM      PID       UID       GID      
test-container    nginx     791186    0         0        
test-container    nginx     791252    101       101      
test-container    nginx     791253    101       101      
test-container    nginx     791254    101       101      
test-container    nginx     791255    101       101      
test-container    nginx     791256    101       101      
test-container    nginx     791257    101       101      
test-container    nginx     791258    101       101      
test-container    nginx     791259    101       101
```

### DIFF
--- a/cmd/ig/utils/flags.go
+++ b/cmd/ig/utils/flags.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/containerd/containerd/pkg/cri/constants"
+
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
 	containerutilsTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/types"
@@ -51,6 +53,9 @@ type CommonFlags struct {
 
 	// Number of seconds that the gadget will run for
 	Timeout int
+
+	// ContainerdNamespace is the namespace used by containerd
+	ContainerdNamespace string
 }
 
 func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
@@ -62,12 +67,14 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		for _, p := range parts {
 			runtimeName := types.String2RuntimeName(strings.TrimSpace(p))
 			socketPath := ""
+			namespace := ""
 
 			switch runtimeName {
 			case types.RuntimeNameDocker:
 				socketPath = commonFlags.RuntimesSocketPathConfig.Docker
 			case types.RuntimeNameContainerd:
 				socketPath = commonFlags.RuntimesSocketPathConfig.Containerd
+				namespace = commonFlags.ContainerdNamespace
 			case types.RuntimeNameCrio:
 				socketPath = commonFlags.RuntimesSocketPathConfig.Crio
 			case types.RuntimeNamePodman:
@@ -85,10 +92,18 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 				}
 			}
 
-			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, &containerutilsTypes.RuntimeConfig{
+			r := &containerutilsTypes.RuntimeConfig{
 				Name:       runtimeName,
 				SocketPath: socketPath,
-			})
+			}
+
+			if namespace != "" {
+				r.Extra = &containerutilsTypes.ExtraConfig{
+					Namespace: namespace,
+				}
+			}
+
+			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, r)
 		}
 
 		// Output Mode
@@ -133,6 +148,13 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		"timeout",
 		0,
 		"Number of seconds that the gadget will run for",
+	)
+
+	command.PersistentFlags().StringVar(
+		&commonFlags.ContainerdNamespace,
+		"containerd-namespace",
+		constants.K8sContainerdNamespace,
+		"Namespace used by containerd",
 	)
 }
 
--- a/cmd/ig/utils/flags.go
+++ b/cmd/ig/utils/flags.go
@@ -20,6 +20,7 @@ import (
 
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
+	containerutilsTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/types"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 
 	log "github.com/sirupsen/logrus"
@@ -46,7 +47,7 @@ type CommonFlags struct {
 
 	// RuntimeConfigs contains the list of the container runtimes to be used
 	// with their specific socket path.
-	RuntimeConfigs []*containerutils.RuntimeConfig
+	RuntimeConfigs []*containerutilsTypes.RuntimeConfig
 
 	// Number of seconds that the gadget will run for
 	Timeout int
@@ -84,7 +85,7 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 				}
 			}
 
-			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, &containerutils.RuntimeConfig{
+			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, &containerutilsTypes.RuntimeConfig{
 				Name:       runtimeName,
 				SocketPath: socketPath,
 			})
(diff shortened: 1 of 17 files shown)


charts/gadget/templates/daemonset.yaml
```diff
@@ -94,6 +94,8 @@ spec:
               value: "/host"
             - name: IG_EXPERIMENTAL
               value: {{ .Values.config.experimental | quote }}
+            - name: CONTAINERD_NAMESPACE
+              value: "k8s.io"
```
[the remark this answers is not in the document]
mqasimsarfraz: containerd k8s namespace is [fixed to be "k8s.io"](https://github.com/containerd/containerd/blob/main/pkg/cri/constants/constants.go#L21) so doesn't makes sense to make it configurable when running in Kubererntes. Perhaps, it can be moved to entrypoint/Dockerfile to be clear?

pkg/container-utils/containerd/containerd.go
```diff
@@ -85,25 +79,30 @@ func (c *ContainerdClient) Close() error {
 }
 
 func (c *ContainerdClient) GetContainers() ([]*runtimeclient.ContainerData, error) {
-	containers, err := c.client.Containers(c.ctx)
+	ctx := namespaces.WithNamespace(context.TODO(), constants.K8sContainerdNamespace)
```
burak-ok: Thats a nice constant. How did you find that :D

pkg/container-utils/containerd/containerd.go
```diff
@@ -335,3 +366,10 @@ func getContainerName(container containerd.Container, labels map[string]string)
 
 	return container.ID()
 }
+
+func containerdNamespaces() []string {
+	if ns := os.Getenv(namespaces.NamespaceEnvVar); ns != "" {
```
burak-ok, line 371: For later: We should start thinking about using `viper` + `cobra` to support both environment variables and flags

burak-ok (approved):
LGTM. Thanks for the change and support for different containerd namespaces

mqasimsarfraz:
Umm. I am having second thoughts if it is good idea to loop over multiple namespaces at once e.g If we have containers with same ids in both "default" and "k8s.io" (such might be very rare), it can be problematic. I see following ways forward:

1. Introduce a `--containerd-namespace` flag and use that to select a specific namespace. Keep "k8s.io" as a default.
2. Introduce the idea of namespace in container-id e.g `k8s.io/test-container` or `default/test-container` to keep them unique.
3. Merge the changes as it is and document the case of containers having same container-id in different namespaces.

I personally prefer 1. Thoughts?

cmd/ig/utils/flags.go
```diff
@@ -90,6 +97,10 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 			})
 		}
 
+		if commonFlags.ContainerdNamespace != constants.K8sContainerdNamespace {
+			os.Setenv(namespaces.NamespaceEnvVar, commonFlags.ContainerdNamespace)
+		}
```
blanquicet, line 102: TBH, I'm unsure about passing this information in an env variable. IMO, it should go through the runtime config like the socket path. Did you explore that possibility?

pkg/container-utils/containerd/containerd.go
```diff
@@ -85,25 +80,30 @@ func (c *ContainerdClient) Close() error {
 }
 
 func (c *ContainerdClient) GetContainers() ([]*runtimeclient.ContainerData, error) {
-	containers, err := c.client.Containers(c.ctx)
+	ctx := namespaces.WithNamespace(context.TODO(), getContainerdNamespace())
+	return c.getContainersInNamespace(ctx)
+}
```
blanquicet, line 85: Could you please add some tests using https://github.com/inspektor-gadget/inspektor-gadget/pull/1850 to verify that a containerd runtime client can retrieve containers in different namespaces? It shouldn't be challenging to modify the [defaultNamespaec](https://github.com/inspektor-gadget/inspektor-gadget/blob/02d21cb3a9c45e1ed1b32ba804d2d606913dd070/pkg/container-utils/testutils/containerd.go#L35) by adding a new [option](https://github.com/inspektor-gadget/inspektor-gadget/blob/02d21cb3a9c45e1ed1b32ba804d2d606913dd070/pkg/container-utils/testutils/options.go#L24) like `WithContainerdNamespace`. WDYT?

BTW, you should manually set the `namespaces.NamespaceEnvVar` in the tests. That's another reason why I'm not convinced about this implementation.

blanquicet (commented):
Some comments. I have some doubts about the chosen implementation.

cmd/ig/utils/flags.go
mqasimsarfraz: > Did you explore that possibility?

I see the point in this, I felt [RuntimeConfig](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/containerutils.go#L45) was bit too generic for including a namespace field since it is only valid for containerd. What do you think about following?

```
type ContainerdConfig struct {
	Namespace string
}

type RuntimeConfig struct {
	Name             types.RuntimeName
	SocketPath       string
	ContainerdConfig *ContainerdConfig
}
```

blanquicet: I think it might work. I wouldn't call it `ContainerdConfig` but something more generic like `ExtraConfig` or similar; naming is always hard. In any case, I wouldn't use `Containerd` in it. WDYT?

mqasimsarfraz: I experimented with the approach and it is looking like [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/c686936caebba32ec66426b8749d52d7bfc51957). What do you say?

blanquicet: Just one comment, for the rest it looks good to me. I would like to have an opinion of @eiffel-fl who had already approved the previous approach.

mqasimsarfraz:
@eiffel-fl @blanquicet I update the PR to use the new approach. Also, added a test for the change. It would be great if I please can get another review!

mqasimsarfraz:
> I would rather have the second commit being first and the adding of `ExtraConfig` in the second (on top of the previous second which became the first 🤯).

Good point. Yeah, that would be much clear. Let me swap them!

pkg/container-utils/containerd/containerd_test.go
```diff
@@ -0,0 +1,49 @@
...
+		c2.Stop(t)
+	})
+	require.Nil(t, err)
+	require.NotNil(t, defaultClient)
+
+	container, err = defaultClient.GetContainer("test-default")
+	require.Nil(t, err)
+	require.NotNil(t, container)
+	require.Equal(t, "test-default", container.Runtime.ContainerID)
+	require.Equal(t, "test-default", container.Runtime.ContainerName)
```
blanquicet, line 48: nit: What about creating a third containerd client with namespace "empty-ns" and executing GetContainers to confirm it doesn't capture either "test-default" nor "test-k8s-io" but an empty list?

cmd/ig/utils/flags.go
```diff
@@ -88,6 +95,9 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, &containerutilsTypes.RuntimeConfig{
 				Name:       runtimeName,
 				SocketPath: socketPath,
+				Extra: &containerutilsTypes.ExtraConfig{
+					Namespace: namespace,
+				},
 			})
```
blanquicet: I wouldn't even pass the ExtraConfig struct if the namespace is nil. Something like:

```go
r := &containerutilsTypes.RuntimeConfig{
	Name:       runtimeName,
	SocketPath: socketPath,
}

if namespace != "" {
	r.Extra = &containerutilsTypes.ExtraConfig{
		Namespace: namespace,
	}
}

commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, r)
```

pkg/operators/localmanager/localmanager.go
```diff
@@ -187,6 +196,9 @@ partsLoop:
 		rc = append(rc, &containerutilsTypes.RuntimeConfig{
 			Name:       runtimeName,
 			SocketPath: socketPath,
+			Extra: &containerutilsTypes.ExtraConfig{
```
blanquicet, line 199: dito

blanquicet (approved):
LGTM!

cmd/ig/utils/flags.go
mqasimsarfraz: Nice. It should be this way, makes the right use of pointer for `ExtraConfig`.

pkg/container-utils/containerd/containerd_test.go
mqasimsarfraz: Done.

mqasimsarfraz:
Thanks for great reviews!

Issue #625, opened 2022-04-22
[EPIC]: Remove BCC dependency by default


mauriciovasquezbernal:
Compile Once - Run Everywhere (CO-RE) is now the preferred way to make eBPF programs portable instead of compiling them at the target machine as (standard) [BCC-Python](https://github.com/iovisor/bcc/tree/master/tools) tools do. Inspektor Gadget supports CO-RE and "BCC" modes for almost all our gadgets, we try to use CO-RE and we fallback to the BCC versions when there is an error. Even if this approach allows us to support a wider set of kernel versions, it presents some big challenges from a development / maintenance point of view: It's difficult to keep features on both types of tools aligned, we have to test twice the gadgets, we have to maintain a [BCC fork](https://github.com/kinvolk/bcc/tree/gadget). From a efficiency point of view, it makes the container image very big (~800MB), and BCC tools require higher resources to be used as them need to be compiled on the target machine. 

BTF is now enabled on almost all major Linux distributions, even Ubuntu enabled it for 18.04 in the kernel version 5.4.0.92 (https://github.com/aquasecurity/btfhub-archive/commit/334a3b230cb6f79a757b8dc8b011d462fb812cd2), Inspektor Gadget is uses [BTFGen](https://kinvolk.io/blog/2022/03/btfgen-one-step-closer-to-truly-portable-ebpf-programs/) and [BTFHub](https://github.com/aquasecurity/btfhub) to be able to run in some kernels without BTF, hence we're proposing to move BCC tools into a "legacy" mode, concentrating our development effort on the new CO-RE tools and making them the default. Specifically, the idea is: 

0. Update all our gadgets to [support CO-RE](https://github.com/kinvolk/inspektor-gadget/issues?q=is%3Aissue+is%3Aopen+label%3Aco-re). 
1. Create two different container images for Inspektor Gadget (BCC and CO-RE, ongoing work here https://github.com/kinvolk/inspektor-gadget/pull/608). 
2. Use the "CO-RE" container image by default when deploying IG with "kubectl gadget deploy".
3. Provide a flag to use the BCC image for scenarios where CO-RE tools can't work.

mauriciovasquezbernal:
I updated this issue to go step by step. Let's start by pushing the CO-RE image to the container registry and provide a flag to choose which one to use. In the future we can make the CO-RE the default.

mauriciovasquezbernal:
The time to deprecate and remove BCC tools has finally come, the reasons are:

[...]

PR #1962, opened 2023-08-17
cmd: Permit deploying CO-RE flavor.

eiffel-fl:
This commit adds the "--co-re" CLI flag to deploy the CO-RE flavor of Inspektor Gadget.
It cannot be used with "--image", as "--co-re" will use "latest-core". We also remove CAP_SYS_MODULE when using CO-RE flavor as this capability is not needed since we do not embed bcc gadgets.

### DIFF
--- a/docs/devel/CONTRIBUTING.md
+++ b/docs/devel/CONTRIBUTING.md
@@ -38,7 +38,7 @@ kubectl-gadget-linux-amd64` or `make kubectl-gadget-darwin-amd64`.
 
 Inspektor Gadget provides two different container images:
 - gadget-default: Contains CO-RE and BCC gadgets
-- gadget-core: Containes only CO-RE gadgets that are integrated with the gadget tracer manager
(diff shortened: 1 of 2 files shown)


Preisschild:
> @Preisschild @kim0 if you are interested in taking a look.

Hi, thanks. Would love to test.

Is there an image already available? I can't find it in the CI. I think it is expired.

Can I deploy it with `kubectl gadget deploy --image <image>`?

Preisschild:
Seems to work. Image used is `v0.19.0-core` and `SYS_MODULE` is not included.

Preisschild:
Yes

mauriciovasquezbernal (commented):
I think it's time to tackle this from the root: https://github.com/inspektor-gadget/inspektor-gadget/issues/625. I updated the description there with the steps to deprecate and remove BCC, I think we can use this PR to handle that. WDYT @eiffel-fl?

PR #1965, opened 2023-08-21
cri: Drop support for CRI v1alpha2

mqasimsarfraz:
Currently, we support CRI v1/v1alpha2 simultaneously. The approach we were using (via `unsafe.Pointer`) works fine till cri-api v1.27.x but since cri-api v1.28.x added a field ('user_specified_image' field to type 'ImageSpec') it breaks the [conversion](https://github.com/inspektor-gadget/inspektor-gadget/pull/1956).

The main reason to support both the CRI API version was to support older version of containerd (1.4.x), which has already been addressed since we migrated to container API directly. So this changes doesn't effect containerd runtimeclient in anyway. In terms of CRI-O, this means we will now only support CRI-O clusters newer than Kubernetes version: v1.20.0.


mqasimsarfraz:
> Just check the commit message "sincewe" -> "since we".

Done 
> Also, should the minimal supported version be documented somewhere?

Good idea. I created a [commit](https://github.com/inspektor-gadget/inspektor-gadget/pull/1965/commits/75514485ef583ba065e8925b01d1aa06e7014fd1) to document it. Please let me know if it looks good!

docs/ig.md
```diff
@@ -46,14 +46,16 @@ The instruction to install `ig` are available in the main
...
+we are using containerd API directly but only `k8s.io` namespace is supported,
+meaning only the ones created via Kubernetes. In addition, it can also use the CRI to
+trace containers managed by CRI-O, Support for non-Kubernetes containers with
 containerd is coming, see issue
-[#734](https://github.com/inspektor-gadget/inspektor-gadget/issues/734).
+[#1849](https://github.com/inspektor-gadget/inspektor-gadget/issues/1849).
 
-By default, `ig` will try to communicate with the Docker Engine
-API and the CRI API of containerd and CRI-O:
+**Note:** We only support CRI v1 meaning that only CRI-O v1.20+ (compatible with Kubernetes v1.20+) is supported.
```
alban, line 56: This note and the details of support for containerd should be moved away for this "Usage" section into docs/requirements.md. At the moment, docs/requirements.md only has one section about the kernel config. It could have another section about Kubernetes & container runtimes.

alban (approved):
LGTM

I added a comment about the documentation, but it was already problematic before your patch, so I think you could merge this PR and address the comment in a future PR if you prefer.

mqasimsarfraz:
> I added a comment about the documentation, but it was already problematic before your patch, so I think you could merge this PR and address the comment in a future PR if you prefer.

Thanks, I will merge this one and open a separate PR for documentation.

PR #1967, opened 2023-08-23
docs: Switch udp/tcp to upper case.

eiffel-fl:
ProtoString() was modified to return the prototype as upper case. This commit reflects this change on the documentation.

Fixes: b9e7897e78cb ("feat(DNS-gadget): Add more fields to DNS gadget")


alban (approved):
LGTM

PR #1968, opened 2023-08-23
doc/containerized-gadgets: Update Packaging chapter to reflect current status

burak-ok:
# doc/containerized-gadgets: Update Packaging chapter to reflect current status

Updates the documentation about the packaging of containerized gadgets.

### DIFF
--- a/docs/design/002-containerized-gadgets.md
+++ b/docs/design/002-containerized-gadgets.md
@@ -105,21 +105,73 @@ TODO: Should commands like `ig pull`, `ig build`, `ig push` be aliases of `ig im
 
 ## Gadget Packaging
 
-Gadgets should be packed as [OCI](https://github.com/opencontainers/image-spec) images. Initially
-the following layers are proposed:
-- `application/vnd.definition.image.inspektor-gadget.io+yaml`: [Definition file](#definition-file)
-- `application/vnd.ebpf.image.inspektor-gadget.io+binary`: Compiled eBPF object.
+Gadgets should be packed as [OCI](https://github.com/opencontainers/image-spec) images.
+It is expected that the first manifest is an image index `application/vnd.oci.image.index.v1+json`.
+All supported architectures has to be listed here
+```bash
+$ crane manifest ghcr.io/inspektor-gadget/tcp-connect:test | jq .
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.index.v1+json",
+  "manifests": [
+    {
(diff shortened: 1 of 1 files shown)


mqasimsarfraz (approved):
Thanks, LGTM!

docs/design/002-containerized-gadgets.md
```diff
@@ -105,21 +105,73 @@ TODO: Should commands like `ig pull`, `ig build`, `ig push` be aliases of `ig im
...
 
-The OCI image should contain the following [OCI annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys):
-- `org.opencontainers.image.title`
-- `org.opencontainers.image.description`
-- `org.opencontainers.image.url`
 
 TODO: How the user could specify these annotations when using the build command?
 
-### Definition File
+### Definition File <a name="def-file"></a>
```
blanquicet, line 174: nit: This change is not needed, you can create anchor links in md files just by doing `[Tittle of link](#name-of-section-to-link)`, in this case `[the definition file](#definition-file)`, see more [here](https://github.com/fefong/markdown_readme).

docs/design/002-containerized-gadgets.md
```diff
@@ -105,21 +105,73 @@ TODO: Should commands like `ig pull`, `ig build`, `ig push` be aliases of `ig im
...
+}
+
+```
 
-The OCI image should contain the following [OCI annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys):
-- `org.opencontainers.image.title`
-- `org.opencontainers.image.description`
-- `org.opencontainers.image.url`
 
 TODO: How the user could specify these annotations when using the build command?
```
blanquicet, line 172: I suggest to allow specifying them in the build command like [`fpm`](https://fpm.readthedocs.io/en/v1.15.1/cli-reference.html) that allows specifying `--license`, `--url`, `--description`, etc. 

However, I was wondering why not to move the author and description info into the definition file? Or maybe the convention is to have this info as annotations?

burak-ok: Nothing is set in stone and everything is still experimental - I also don't know the best place :D

I have chosen to put it in an annotation on the eBPF layer so the author is tied to the eBPF code. The definition file is "just" a file which tells Inspektor Gadget which fields of the struct should be shown -> Nothing "to author" here

blanquicet: Makes sense, so we can go with the same approach of `fpm`.

Issue #1875, opened 2023-07-20
run: Handling OCI images


burak-ok:
# Discussion about `ig run ghcr.io/foo/bar` and `kubectl-gadget run ghcr.io/foo/bar`

Should the client or the server pull the multi-arch image. Initial discussion with @alban 

## `ig`

Doesn't matter that much/Not that interesting. client == server

### Pulling is done client side (`kubectl-gadget`)

The multi-arch image is pulled by the client. It can read the config and the BTF if needed.
It then sends both/specific ELF binaries to every gadget pod.
They execute it and everything works.

For air-gapped servers the user has to push the needed images to a local registry first. Only the client needs access to the registry.
Alternativly we can ship a file based oci store. We then try to read from that first in `kubectl-gadget`

### Pulling is done server side (gadget pods)

Only the needed architecture of the image is pulled by the pods.
No large network transfers are needed between the client and all nodes (The more nodes - the larger the initial performance hit and network congestions)

However the client (`kubectl-gadget`) needs to know the config.
This means at least one gadget pod has to send this information back to the client.
If there is no config (which we want to support in the future) we need to make sure that all pods (multi-arch) would send the same information.

We additionally need the BTF at the client side (currently) aswell.
Gadget pods are sending `event struct`s back to the client. The struct has a `RawData []byte` buffer.
We need access to the BTF to decode the starting addresses and lengths of the kernel struct members. 

For air-gapped servers the user has to push the needed images to a local registry first. (Only) the pods needs access to the registry.
If we wanted to ship a file based oci store we would need to deploy alongside every gadget pod

## Multi-Arch problems

Since we would deploy the same ELF file on a cluster with varying kernel versions and architectures it could happen that the kernel member struct offsets and lengths are different.
So the decoding of the BTF and the `RawData []byte` buffer should happen at the pod?

mauriciovasquezbernal:
> Doesn't matter that much/Not that interesting. client == server

It's almost true except for https://github.com/inspektor-gadget/inspektor-gadget/issues/1681. Anyway, in this case multiarch support is not a challenge, so I agree there's not too much to discuss here.

> ### Pulling is done client side (`kubectl-gadget`)
> 
> The multi-arch image is pulled by the client. It can read the config and the BTF if needed. It then sends both/specific ELF binaries to every gadget pod. They execute it and everything works.
 
I think sending images to all nodes puts a lot of stress to the client. In many cases the client can be running in a machine with a slow connection. 

[...]

PR #1974, opened 2023-08-25
[WIP] gadget/run: Implement running from an oci image

burak-ok:
# Unfinished and experimental
## run: Implement running from an oci image

This PR is the current state for `./kubectl-gadget run ghcr.io/burak-ok/test`.
So this downloads the oci image on demand, unpacks it and runs it on the cluster.

It is currently split into 2 commits:

### gadget/run: Implement pulling and running oci images

As mentioned in the commit message this can be seen as an intermediate step.
The client and the server are pulling the image, and reading the BTF. So everything works the same as the "normal" built-in gadgets.
Which means that the `type Event struct` gets sent through gRPC from the Server (Cluster) to the Client (where the `kubectl-gadget` binary is) 
Since the `run` gadget deals with by design highly dynamic fields we don't know at compile time which columns we need and will have. This is all dependent on the oci-image we get at runtime.
For that we have a `RawData []byte` field in the `type Event struct` of the run gadget. Since the server sends the whole struct to the client it also gets that raw data.

The problem is that now the client has to decode the raw BTF data itself. Therefore it needs access to the BTF to lookup the starting address and length of the fields. Because of thoughts that the architecture and endianess and other stuff could affect the layout of the `RawData []byte` we want the client to not deal with the raw BTF data itself. See dicussion in issue #1875 

But this commit can be used for demonstrations when we know we have the same layout.

### gadget/run: client shouldn't pull image and parse BTF

This commit explores a way for the client to get and receive all necessary information from the servers. (In this commit the client still pulls and reads the BTF, but the information is not used -> That code can be deleted/deactivated later)

The client expects a new type of gRPC message from each server (`EventTypeColAttr`). That message contains the `columns.Attributes` for all dynamic fields.
For every event the server is not sending the JSON serialized `type Event struct`. We are using the "other" JSON serializer instead for the `run` gadget (See PR #1817 for more information regarding that). This unfortunately means that we can not use the normal `json.Unmarshal` call to deserialize the message on the client side.

The idea now is to populate all known and static fields (like `PodName` and `Node`) manually. For the dynamic fields from eBPF we store them in an (otherwise unused??) field in the `type Event struct`: `Data interface{}`.
After populating the static fields on the client side we create a `map[string]string`. We fill it with the dynamic eBPF fields which are present in the received JSON. We also know which of the JSON fields belong to the dynamic eBPF fields because of the `EventTypeColAttr` message.
The column name is used as the map key and the value is its map value. (Remember: The operations done on the `Data interface{}` field is done on the client)

When receiving the `EventTypeColAttr` event type we also created a new `columns.Columns[Event]`. We add a virtual column for each `columns.Attributes` we receive. The extractor for that column is implemented in the following way:
```golang
ebpfVals, ok := ev.Data.(map[string]string)
if !ok { /*...*/ }
return ebpfVals[colName]
```
So we are expecting the `Data interface{}` field to be a `map[string]string` - and that is exactly where and how we store our dynamic eBPF fields after we received them from the server.

### Issues

The parser and formatter are created very early in the program for `kubectl-gadget`. All the filtering and formatting is done before and the parser we create after receiving this event doesn't have that information anymore. So that part is not working.

We think that because of that that we will need a larger refactoring regarding that.

cc @mqasimsarfraz


burak-ok:
Feature is developed in #2015

PR #1984, opened 2023-08-29
Remove duplicated function

mauriciovasquezbernal:
We got two functions that do the same (ProtoString and ProtoToString), use only one of them.


blanquicet (approved):
LGTM!

Issue #1929, opened 2023-08-03
[EPIC] Implement Image-based Gadgets 


mauriciovasquezbernal:
Containerized gadgets were proposed in https://github.com/inspektor-gadget/inspektor-gadget/issues/1669 and its [design document](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md) was merged in https://github.com/inspektor-gadget/inspektor-gadget/pull/1856. This issue is an umbrella tracker of all implementation issues for it:

- [X] Initial Support (https://github.com/inspektor-gadget/inspektor-gadget/pull/1743)
- [X] https://github.com/inspektor-gadget/inspektor-gadget/issues/1827
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1917
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1918
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1919
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1920
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1921
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1922
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1923
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/issues/1924
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1925
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1926
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1927
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1928
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1875
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/1932
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/2006
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/2021
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/2032
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/2045
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/2046
- [x] https://github.com/inspektor-gadget/inspektor-gadget/issues/2096
[...]

PR #1985, opened 2023-08-30
[PoC] dns: implement parseLabelSequence in wasm

alban:
This patch removes the implementation of parseLabelSequence from the dns tracer package and instead implements it in an external wasm module. In this PoC, the wasm module was compiled from Golang with tinygo but it could be from other languages.

The wasm module could be moved to the containerized gadget.

Compile the wasm module:
```
sudo dnf install tinygo binaryen
make -C gadgets/dns-label-sequence/
```

```
$ go run -exec sudo ./cmd/ig/... trace dns
RUNTIME.CONTAINERNAME    PID    TID    COMM  QR TYPE      QTYPE NAME              RCODE  NUMANSW…
keen_stonebraker         32910  32910  wget  Q  OUTGOING  A     wikipedia.org.           0
```

Event generated with:
```
docker run -ti --rm busybox wget wikipedia.org
```

xref https://github.com/inspektor-gadget/inspektor-gadget/issues/1929

cc @mauriciovasquezbernal

### DIFF
--- a/gadgets/dns-label-sequence/main.go
+++ b/gadgets/dns-label-sequence/main.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"unsafe"
+)
+
+var bufIn [1024]byte
+var bufOut [1024]byte
+
+//go:export alloc
+func alloc(size uint32) *byte {
+	return &bufIn[0]
+}
+
+//go:export parseLabelSequence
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal:
@alban can we close it in favor of https://github.com/inspektor-gadget/inspektor-gadget/pull/2204?

alban:
Closing in favor of #2204

Issue #1920, opened 2023-08-03
run: Support socket enricher


mauriciovasquezbernal:
Support programs that use the socket enricher (pkg/gadgets/internal/socketenricher/tracer.go) as described in the design document https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md#socket-enricher

alban:
The gadgets currently using the socket enricher are:
1. network gadgets using the Attacher interface:
  a. trace dns
  b. trace network
  c. trace sni
2. gadgets using tracepoints:
  a. trace tcpdrop
  b. trace tcpretrans

The two categories use different socket enricher APIs:
1. network gadgets: `struct sockets_value *gadget_socket_lookup(const struct __sk_buff *skb)`
2. tracing gadgets: `struct sockets_value *gadget_socket_lookup(const struct sock *sk, __u32 netns)`

We don't support the Attacher interface in containerized gadgets yet (see #1919). So I can start with the support for the socket enricher in containerized gadgets using tracepoints.

The tcpdrop gadget requires the enum conversion (#1928) which is not supported yet. So I will start with adding tcpretrans in `gadgets/`.

PR #1997, opened 2023-09-04
Socket enricher for containerized gadgets

alban:
# Socket enricher for containerized gadgets

Before this patch, the socket enricher was only available for built-in gadgets. This patch adds support for containerized gadgets.

As example, it uses the tcpretrans gadget.

## How to use

```
go run -exec 'sudo -E' ./cmd/ig/... run --prog @./gadgets/trace_tcpretrans_x86.bpf.o --definition @./gadgets/trace_tcpretrans.yaml
```

## Testing done

Generate events with:
```
$ docker run -ti --rm --cap-add NET_ADMIN --name=netem wbitt/network-multitool -- /bin/bash
bash-5.1# tc qdisc add dev eth0 root netem drop 25% ^C
bash-5.1# wget 1.1.1.1
```
Observe events with:
```
$ go run -exec 'sudo -E' ./cmd/ig/... run --prog @./gadgets/trace_tcpretrans_x86.bpf.o --definition @./gadgets/trace_tcpretrans.yaml
RUNTIME.CONTAINERNAME     SRC              DST          TASK     PID     UID      GID
netem                     172.17.0.3:35536 1.1.1.1:80   wget     105692  0        0
netem                     172.17.0.3:56494 1.1.1.1:443  wget     105692  0        0
```
Or:
```
$ go run -exec 'sudo -E' ./cmd/ig/... run --prog @./gadgets/trace_tcpretrans_x86.bpf.o --definition @./gadgets/trace_tcpretrans.yaml -o yaml
container: ""
dst:
  addr: 1.1.1.1
  kind: ""
  name: ""
  namespace: ""
  port: 80
  proto: TCP
  v: 4
gid: 0
hostnetwork: false
mntns: 4026532835
mntns_id: 4026532835
namespace: ""
node: ""
pid: 105842
pod: ""
runtime:
  containerId: 3b45f227d11673f86ecff58fecb0e350f0e807935b9005a87c81dca1be5e3a89
  containerImageName: wbitt/network-multitool
  containerName: netem
  runtimeName: docker
src:
  addr: 172.17.0.3
  kind: ""
  name: ""
  namespace: ""
  port: 58712
  proto: TCP
  v: 4
task: wget
timestamp: "1970-01-01T01:00:00.000000000+01:00"
uid: 0
```

TODO:
- [x] Endianness issue with source port and destination port
- [x] Missing ip source and ip destination (waiting for #1825 before fixing this)
- [x] Duplicate 'mntns_id' and 'mntns' field

TODO unrelated to this PR (will be fixed in another PR):
- [ ] Wrong timestamp
- [ ] Duplicate 'container' field

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1920


alban:
I rebased after #1825. I fixed some TODOs but left some for future PRs (see updated PR description).

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -176,6 +180,21 @@ func (t *Tracer) installTracer() error {
 		}
 	}
 
+	// Only create socket enricher if this is used by the tracer
+	for _, m := range t.spec.Maps {
+		if m.Name == networktracer.SocketsMapName {
```
alban, line 185: At the moment, `networktracer.SocketsMapName` is defined as follows:
```
SocketsMapName = "sockets"
```

What do you think about renaming it to "gadget_sockets" so that it is namespaced correctly for third-party gadget authors?

mauriciovasquezbernal: Sure, I think we must rename that to avoid collisions.

gadgets/trace_tcpretrans.yaml
```diff
@@ -0,0 +1,33 @@
+name: tcpretrans
+description: trace TCP retransmissions
+columns:
+  - name: timestamp
+    template: timestamp
+    order: 900
```
mauriciovasquezbernal, line 6: I had to comment this out to make it work:

```bash
Error: calling custom parser: adding fields: duplicate column name "timestamp"
exit status 1
```

I'll open an issue for it as it's not related to this PR.

mauriciovasquezbernal: Ref https://github.com/inspektor-gadget/inspektor-gadget/issues/2000

pkg/gadgets/run/tracer/tracer.go
alban: I renamed it with the "ig_" prefix like some other maps and progs.

I also moved the constant from the `networktracer` package to the `socketenricher` package where it semantically belongs.

mauriciovasquezbernal: > I renamed it with the "ig_" prefix like some other maps and progs.

I'm trying to use `gadget_` everywhere as in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/common/mntns_filter.h for the API. I think we're using `ig_` in the programs because the length restriction. However I don't have a preference in any of them, but I'd prefer everything aligned.

mauriciovasquezbernal (approved):
Just one comment about the naming of the map, but it's not blocking. LGTM, thanks!

pkg/gadgets/run/tracer/tracer.go
alban: Ok, I renamed to gadget_sockets.

I will merge once the CI finishes.

PR #1998, opened 2023-09-05
Dockerfiles: Make CO-RE the default and set previous default as bcc.

eiffel-fl:
Hi.

In this PR, CO-RE becomes the default and old default becomes `-bcc` image.

Best regards.

### DIFF
--- a/Dockerfiles/gadget-default.Dockerfile
+++ b/Dockerfiles/gadget-default.Dockerfile
@@ -9,6 +9,10 @@
 ARG BUILDER_IMAGE=golang:1.19-bullseye
 ARG BASE_IMAGE=debian:bullseye-slim
 
+# bpftrace upstream image
+ARG BPFTRACE="ghcr.io/inspektor-gadget/bpftrace"
+
+FROM ${BPFTRACE} as bpftrace
 # Prepare and build gadget artifacts in a container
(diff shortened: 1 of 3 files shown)


Dockerfiles/gadget-default.Dockerfile
```diff
@@ -97,8 +102,6 @@ COPY gadget-container/hooks/nri/conf.json /opt/hooks/nri/
 # BTF files
 COPY hack/btfs /btfs/
 
-COPY --from=bpftrace /usr/bin/bpftrace /usr/bin/bpftrace
-
```
mauriciovasquezbernal: It is an issue. We should support bpftrace on the default image. When I added it, I decided to only add to the default image as it wasn't working on the CO-RE one because some issues with libraries. Do you know if we can make bpftrace work on the default image? Like having a statically compiled version of it.

mauriciovasquezbernal (approved):
Looks great overall. Some minor nits:
- Change this to bcc -> https://github.com/inspektor-gadget/inspektor-gadget/blob/5779db88ca11998134eaa9625dc4fbc56b72d492/Makefile#L296
- Update documentation around https://github.com/inspektor-gadget/inspektor-gadget/blob/6e7668c8d7ef2db1c5abc6089b8fce245bc82ce6/docs/devel/CONTRIBUTING.md#L37-L54

Approving now to save another review round.

PR #1999, opened 2023-09-05
Refactoring Kubernetes labels

nlamirault:
# [Support Kubernetes recommended labels and additional labels]

- Use [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
- Support to additional labels

## How to use

```
helm template .
```

## Testing done

Manifests seems valid. Ex:

```yaml
---
# Source: gadget/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    helm.sh/chart: gadget-1.0.0
    app.kubernetes.io/name: gadget
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/component: controller
    app.kubernetes.io/version: "1.0.0"
    app.kubernetes.io/managed-by: Helm
  name: release-name-gadget
  namespace: gadget
```

### DIFF
--- a/charts/gadget/values.yaml
+++ b/charts/gadget/values.yaml
@@ -2,6 +2,13 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# -- Skip Helm labels
+skipLabels: true
+
+# -- Labels to be added to all other resources.
+additionalLabels:
+  labels: {}
+
 config:
   # -- How to get containers start/stop notifications (auto, crio, podinformer, nri, fanotify, fanotify+ebpf")
   hookMode: auto
--- a/charts/gadget/templates/_helpers.tpl
+++ b/charts/gadget/templates/_helpers.tpl
@@ -38,6 +38,33 @@ gadget
 {{- end }}
 
 {{/*
+Common labels
+*/}}
+{{- define "gadget.labels" -}}
+helm.sh/chart: {{ include "gadget.chart" . }}
+{{ include "gadget.selectorLabels" . }}
+app.kubernetes.io/component: controller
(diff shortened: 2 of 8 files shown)


mqasimsarfraz:
@nlamirault Thanks for the contribution. The changes are looking great with a quick test. 

But I see that the CI is [failing](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/6086576341/job/16515124392?pr=1999) since it detect changes in [deploy.yaml](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/resources/manifests/deploy.yaml) (used by `kubectl gadget deploy`) since we don't want helm related labels in it. Perhaps we can guard these labels with `skipLabels` (default false) value and set it to true [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/charts/Makefile#L42)?

charts/gadget/values.yaml
```diff
@@ -2,6 +2,9 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# Labels to be added to all other resources.
```
mqasimsarfraz, line 5: nit: It is missing `--` prefix in order to auto generate documentation. (Also, I would like it below config object)

mqasimsarfraz:
Thanks for handling the comments. Can you please create a single commit for all the changes ? Afterwards, it is ready to go!

mqasimsarfraz (approved):
LGTM!

mqasimsarfraz:
Thanks for the contribution!

Issue #1919, opened 2023-08-03
run: Support gadgets with `Attacher` interface


mauriciovasquezbernal:
The initial support for containerized gadgets introduced in https://github.com/inspektor-gadget/inspektor-gadget/pull/1743 doesn't support gadgets that implement the [attacher interface](https://github.com/inspektor-gadget/inspektor-gadget/blob/0eeb7f19d53882e15dedf9247f6a331d4d1b0e14/pkg/operators/kubemanager/kubemanager.go#L47-L50) like trace dns and trace sni. 

Described in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md#socket_filter-programs

### Implementation options

The current implementation of those gadgets is based on the networktracer (pkg/gadgets/internal/networktracer/tracer.go) that implements such a interface and attaches the eBPF programs to the different network namespaces of the containers when they're created. 

I believe we'd need to reused (or reimplement) the networktracer in the run command. 

Note that such interface is also used by the iterators programs in https://github.com/inspektor-gadget/inspektor-gadget/pull/1866

PR #2003, opened 2023-09-06
run: Support gadgets with Attacher interface (networking gadgets)

alban:
# run: Support gadgets with Attacher interface (networking gadgets)

Difficulties:
- Networktracer's NewTracer() is called before we know the ebpf program, so it needed some refactoring
- AttachContainer/DetachContainer are called before we know the ebpf program. AttachContainer attaches the ebpf dispatcher program. Then the tail call map is updated once we know the ebpf program.

## How to use

```
$ go run -exec 'sudo -E' ./cmd/ig/... run --prog @./gadgets/trace_dns_x86.bpf.o --definition @./gadgets/trace_dns.yaml
RUNTIME.CONTAINERNAME                  PID     SRC                      DST                      UID      GID      NAME                                   QR PKT_TYPE   RCODE
netem                                  865482  172.17.0.3:40591         192.168.0.1:53           0        0        	wikipediaorg                         0  4          0
netem                                  865482  192.168.0.1:53           172.17.0.3:40591         0        0        	wikipediaorg                         1  0          0
```

```
$ go run -exec 'sudo -E' ./cmd/ig/... run --prog @./gadgets/trace_dns_x86.bpf.o --definition @./gadgets/trace_dns.yaml -o yaml
container: ""
dst:
  addr: 192.168.0.1
  kind: ""
  name: ""
  namespace: ""
  port: 53
  proto: UDP
  v: 4
gid: 0
hostnetwork: false
latency_ns: 0
mntns: 4026533092
mntns_id: 4026533092
name: "\twikipedia\x03org"
namespace: ""
node: ""
pid: 867049
pkt_type: 4
pod: ""
qr: 0
rcode: 0
runtime:
  containerId: 05f8906b9fc6002ffd757b2eadc13c9d1fed8214fc82197a22ec7512fd370c8d
  containerImageName: wbitt/network-multitool
  containerName: netem
  runtimeName: docker
src:
  addr: 172.17.0.3
  kind: ""
  name: ""
  namespace: ""
  port: 41280
  proto: UDP
  v: 4
timestamp: "1970-01-01T01:00:00.000000000+01:00"
uid: 0
```

## Testing done

TODO:
- [x] Rebase after #1997 (socket enricher)
  - [x] Check container enrichment
  - [x] Use proper columns: endpoints
- [x] Developer documentation

TODO for future PRs:
- [ ] Fix `MAX_ADDR_ANSWERS` (changed to 1 to avoid verifier error)
- [ ] Find a solution for dns event variable length
- [ ] Use bpf timers to clean up ebpf map

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1919


Makefile.btfgen
```diff
@@ -15,7 +15,10 @@ SOURCE_BTF_FILES = $(shell find $(BTFHUB_ARCHIVE)/ -iregex ".*$(subst x86,x86_64
 MIN_CORE_BTF_FILES = $(patsubst $(BTFHUB_ARCHIVE)/%.btf.tar.xz, $(OUTPUT)/%.btf, $(SOURCE_BTF_FILES))
 BPF_ALL_O_FILES = $(shell find pkg/gadgets -type f -regex ".*\($(ARCH)\|bpfel\).o")
 # Filter out BPF objects that only contain BPF maps without BPF programs
-BPF_PROGS_O_FILES = $(filter-out pkg/gadgets/trace/network/tracer/graphmap_bpfel%,$(BPF_ALL_O_FILES))
+BPF_PROGS_O_FILES = $(filter-out \
+	pkg/gadgets/trace/network/tracer/graphmap_bpfel% \
```
alban, line 19: graphmap_bpfel was deleted in commit 0a4227db582e3360283753c76b71dd81cec93aa2. I should remove this line.

gadgets/Makefile
```diff
@@ -21,5 +24,35 @@ all: $(GADGETS)
...
+#
+# Since Ubuntu does not install it in a standard path, add a compiler flag for
+# it.
+%_bpfel.bpf.o : %.bpf.c
+	CLANG_OS_FLAGS= ; \
+	if [ "$(shell grep -oP '^NAME="\K\w+(?=")' /etc/os-release)" == "Ubuntu" ]; then \
+		CLANG_OS_FLAGS="-I/usr/include/$(shell uname -m)-linux-gnu" ; \
+	fi ; \
+	docker run --rm \
+		--name ebpf-object-builder \
```
alban, line 46: It does not work for me when compiled without docker (on Fedora)... Hence using the docker builder.

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -117,18 +117,30 @@ func (t *Tracer[Event]) newAttachment(
 }
 
 func NewTracer[Event any](
-	spec *ebpf.CollectionSpec,
 	baseEvent func(ev types.Event) *Event,
 	processEvent func(rawSample []byte, netns uint64) (*Event, error),
```
mauriciovasquezbernal, line 140: I think these two should be moved to `Listen()` or `Run()` as they are not really needed when creating the network tracer.

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -255,57 +276,64 @@ func (t *Tracer[Event]) EventCallback(event any) {
...
 
-func (t *Tracer[Event]) listen(
-	rd *perf.Reader,
-	baseEvent func(ev types.Event) *Event,
-	processEvent func(rawSample []byte, netns uint64) (*Event, error),
-	eventCallback func(*Event),
-) {
+func (t *Tracer[Event]) Listen() {
+	go t.listen()
+}
```
mauriciovasquezbernal, line 319: Any reason not to combine this with `Run()`. I image builtin gadgets could have something like:

```golang
	err = networkTracer.Run(spec, types.Base, parseDNSEvent)
	if err != nil {
		return fmt.Errorf("setting network tracer spec: %w", err)
	}
```

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -75,12 +78,26 @@ type Tracer struct {
 	perfReader        *perf.Reader
 	printMapValueSize uint32
 
+	useAttacherInterface bool
```
mauriciovasquezbernal, line 81: This can be a local variable in the current implementation.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -62,6 +63,8 @@ type Config struct {
 }
 
 type Tracer struct {
+	*networktracer.Tracer[types.Event]
```
mauriciovasquezbernal, line 66: I think we don't should embed this but make it a normal member. I suppose you embedded it to make this tracer implement the `{Attach,Detach}Container` methods, however I think we'll need more control over those methods (like in https://github.com/inspektor-gadget/inspektor-gadget/pull/1866) hence we need to implement those directly in this type. The implementation on this could be a "proxy" and call those methods in `networktracer.Tracer`.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -236,6 +253,15 @@ func (t *Tracer) installTracer() error {
...
 			t.links = append(t.links, l)
+		} else if p.Type == ebpf.SocketFilter && strings.HasPrefix(p.SectionName, "socket") {
+			if t.useAttacherInterface {
+				return fmt.Errorf("several socket filters found, only one is supported")
+			}
+			t.useAttacherInterface = true
+			err := t.Tracer.AttachProg(t.collection.Programs[progName])
+			if err != nil {
+				return fmt.Errorf("attaching ebpf program to dispatcher: %w", err)
+			}
```
mauriciovasquezbernal, line 264: With my comment above making a networktracer.Tracer a member, do you think we could have multiple instances of it in order to support multiple socket programs? (We can handle this later on)

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -464,4 +490,8 @@ func (t *Tracer) SetEventHandler(handler any) {
 		panic("event handler invalid")
 	}
 	t.eventCallback = nh
+
+	if t.Tracer != nil {
+		t.Tracer.SetEventHandler(handler)
+	}
```
mauriciovasquezbernal, line 496: If the run gadget is pulling the events by itself, is this call really needed?

pkg/gadgets/internal/networktracer/bpf/dispatcher-map.bpf.c
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: (GPL-2.0 WITH Linux-syscall-note) OR Apache-2.0
+/* Copyright (c) 2023 The Inspektor Gadget authors */
+
+#include <linux/bpf.h>
+#include <bpf/bpf_helpers.h>
+
+#include "dispatcher-map.h"
+
+char _license[] SEC("license") = "GPL";
```
mauriciovasquezbernal, line 9: Does it really need to be GPL?

mauriciovasquezbernal (commented):
First pass. I'll continue tomorrow with another one.

gadgets/trace_dns.bpf.c
```diff
@@ -0,0 +1,403 @@
...
+			.id = event->id,
+		};
+		if (event->qr == DNS_QR_QUERY &&
+		    event->pkt_type == PACKET_OUTGOING) {
+			bpf_map_update_elem(&query_map, &query_key,
+					    &event->timestamp, BPF_NOEXIST);
+		} else if (event->qr == DNS_QR_RESP &&
+			   event->pkt_type == PACKET_HOST) {
+			__u64 *query_ts =
+				bpf_map_lookup_elem(&query_map, &query_key);
```
mauriciovasquezbernal, line 347: We need to find a way to clean this up

gadgets/trace_dns.bpf.c
```diff
@@ -0,0 +1,403 @@
...
+	if (load_half(skb, DNS_OFF + offsetof(struct dnshdr, qdcount)) != 1)
+		return 0;
+
+	__u16 ancount =
+		load_half(skb, DNS_OFF + offsetof(struct dnshdr, ancount));
+	__u16 nscount =
+		load_half(skb, DNS_OFF + offsetof(struct dnshdr, nscount));
+
+	// Skip DNS queries with answers
+	if ((flags.qr == 0) && (ancount + nscount != 0))
```
mauriciovasquezbernal, line 393: I know if was there before but I don't think the extra `()` are needed.

gadgets/Makefile
mauriciovasquezbernal: What error do you get?
Using the following worked fine for me on mint.

```Makefile
%_bpfel.bpf.o : %.bpf.c
	clang -target bpf -Wall -g -O2 -c $(filter %.c,$^) \
		-I ../pkg/gadgets/common/ \
		-I ../pkg/gadgets/internal/socketenricher/bpf/ \
		-o $@
```

I really think we should avoid using docker here now, it'll come later done in another layer.

pkg/gadgets/run/tracer/tracer.go
alban: In theory yes, but with the current way the code is organised, it is not easy:

`AttachContainer()`/`DetachContainer()` on the `run` gadget can be called by the runtime as soon as `NewInstance()` / `PreGadgetRun()` is called, that is before `Run()` is called. But the `run` gadget only know the ebpf code when `Run()` method is called. So `AttachContainer()` should be made to work without knowing how many socket filters the program has.

This also means that the networkTracer's dispatcher ebpf program is installed even if the gadget does not have any socket filter. This is unfortunate.

I was considering adding a container cache in the `run` gadget. But I think that would be at the wrong layer. Alternatively, the following initialization could be done after calling the `Run()` method:
* https://github.com/inspektor-gadget/inspektor-gadget/blob/3653cd35afc97383ed49cbef0b4b065989a5faaa/pkg/operators/localmanager/localmanager.go#L322-L345
* https://github.com/inspektor-gadget/inspektor-gadget/blob/3653cd35afc97383ed49cbef0b4b065989a5faaa/pkg/operators/kubemanager/kubemanager.go#L283-L300

I suggest to handle this issue later.

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -116,20 +136,30 @@ func (t *Tracer[Event]) newAttachment(
...
+	if err != nil {
+		return nil, err
+	}
+	var opts ebpf.CollectionOptions
+	if err = dispatcherTailCallMapSpec.LoadAndAssign(&t.dispatcherTailCallMapObjs, &opts); err != nil {
+		return nil, fmt.Errorf("loading ebpf program: %w", err)
+	}
+
+	return t, nil
+}
```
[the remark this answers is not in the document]
alban: It can work but it means the map parameter (`max_entries`) are defined at 2 different places and need to be kept in sync. However, that's just one parameter, and your solution reduces the amount of files. I'm ok either way. What do you prefer?

gadgets/trace_dns.bpf.c
alban: I suggest to use bpf timers from Linux 5.15:
https://github.com/torvalds/linux/commit/b00628b1c7d595ae5b544e059c27b1f5828314b4

That would change the [requirements](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/requirements.md) for the dns gadget from Linux 5.4 to Linux 5.15. But the builtin dns gadget will still be here for a while, until users switch to Linux >= 5.15.

pkg/gadgets/internal/networktracer/tracer.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's better to reduce the amount of files. We can add a comment in both places to keep the parameter in sync.

pkg/gadgets/run/tracer/tracer.go
mauriciovasquezbernal: I agree, let's do this later on. Thanks for the clarification.

mauriciovasquezbernal (commented):
It's looking very good! Another set of comments, I'll make the final review and approve once they're handled.

gadgets/Makefile
```diff
@@ -21,5 +24,13 @@ all: $(GADGETS)
...
 		-o $@
 
+# Networking gadgets
+%_bpfel.bpf.o : %.bpf.c
+	clang -target bpf -Wall -g -O2 -c $(filter %.c,$^) \
+		-I ../pkg/gadgets/common/ \
+		-I ../pkg/gadgets/internal/socketenricher/bpf/ \
+		-o $@
+
+
```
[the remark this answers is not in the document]
alban: I agree. That would make thing simpler. Also, a gadget can use both socket filters and tracing programs.

pkg/gadgets/internal/networktracer/bpf/dispatcher-map.bpf.c
alban: File removed, following discussion below.

pkg/gadgets/internal/networktracer/tracer.go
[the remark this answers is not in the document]
alban: Ok, I am removing the file and I added the 2 comments

Makefile.btfgen
```diff
@@ -15,7 +15,9 @@ SOURCE_BTF_FILES = $(shell find $(BTFHUB_ARCHIVE)/ -iregex ".*$(subst x86,x86_64
 MIN_CORE_BTF_FILES = $(patsubst $(BTFHUB_ARCHIVE)/%.btf.tar.xz, $(OUTPUT)/%.btf, $(SOURCE_BTF_FILES))
 BPF_ALL_O_FILES = $(shell find pkg/gadgets -type f -regex ".*\($(ARCH)\|bpfel\).o")
 # Filter out BPF objects that only contain BPF maps without BPF programs
-BPF_PROGS_O_FILES = $(filter-out pkg/gadgets/trace/network/tracer/graphmap_bpfel%,$(BPF_ALL_O_FILES))
+BPF_PROGS_O_FILES = $(filter-out \
+	pkg/gadgets/internal/networktracer/dispatchertailcallmap_bpfel%, \
```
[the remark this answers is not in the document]
alban: Correct. I will revert.

graphmap_bpfel should be removed too but since it is not related to this PR, it should be done in another PR.

alban:
> I suppose it'll be fixed with something similar to https://github.com/inspektor-gadget/inspektor-gadget/pull/1985 in the future.

Yes. I am preparing a [branch](https://github.com/alban/inspektor-gadget/tree/alban_asg2023) which gives a good output, but that will be for other PRs:

```
$ sudo -E ./ig run albantest.azurecr.io/trace-dns:v2
RUNTIME.CONTAINERNAME           SRC                      DST                      PID     TASK              UID      GID      NAME                            QR PKT_TYPE RCODE   
netem                           172.17.0.3:36213         192.168.0.1:53           1143195 wget              0        0        wikipedia.org.                  0  4        0
```

> Great job, LGTM!

Thanks. I will merge once the CI completes.

PR #2013, opened 2023-09-07
ci: Use federated credential to login to Azure.

eiffel-fl:


docs/ci.md
```diff
@@ -169,25 +169,51 @@ $ subscription=<mySubscription>
...
 {
-# The JSON object containing a secret should be stored as CI secret.
+# It replies with a JSON object which has name set to $federated_name.
 }
 ```
 
 After doing this, you will need to create several secrets:
 
-1. `AZURE_AKS_CREDS`: It stores the JSON object outputted by ``.
+1. `AZURE_AKS_CLIENT_ID`: The application ID as given by `az ad app list --display-name $app_name --query [0].id`.
```
mqasimsarfraz, line 213:
```suggestion
1. `AZURE_AKS_CLIENT_ID`: The application ID as given by `az ad app list --display-name $app_name --query [0].appId`.
```

mqasimsarfraz: > It seems both of them do not work.

This is weird. I used the exact steps in [different project](https://github.com/Azure/kubectl-aks/pull/46) to see if it works and everything is green. I wondering this might be linked to how app is registered. I will see if I find anything!

mqasimsarfraz (approved):
LGTM! The login step is [green](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/6260616182/job/17002159833#step:5:24) now! 

Great work! It should save us from the pain of credential renewal.

PR #2019, opened 2023-09-11
Create include folder with all header files exposed by IG

mauriciovasquezbernal:
As we're moving more towards containerized gadgets, we need to prepare everything needed to build gadgets outside the tree. The first step is to consolidate all headers files used by gadgets in the same place.

This commit has a lot of changed files, but there is not any behaviour change.

The only need thing, is the "make {install,remove}-headers" targets.


mauriciovasquezbernal:
I can't reopen this PR, hence I opened a new one in https://github.com/inspektor-gadget/inspektor-gadget/pull/2022.

Issue #2006, opened 2023-09-06
`ig image build` should compile the eBPF code 


mauriciovasquezbernal:
The `ig image build` expects the **compiled** program as argument (`--prog-amd64` and `--prog-arm64`). To be aligned with the docker UX, the build command should actually build the eBPF code and create the gadget image. 

### Implementation options

We should provide a builder container with all dependencies ready for users that don't want to install all of that in their machines, but we should also support the case when one wants to use the local installed tools. We can take inspiration from the bumblebee build command -> https://github.com/solo-io/bumblebee/blob/main/pkg/cli/internal/commands/build/build.go

mauriciovasquezbernal:
Fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/2015.

PR #2020, opened 2023-09-11
ig image build: build ebpf from source

mauriciovasquezbernal:
```    
    This commit updates the "image build" command to compile the eBPF code
    from source instead of taking an already compiled bpf object.
    
    The building process is controlled by a build file (build.yaml by
    default), if that's not found the command assumes some defaults, like
    the program.bpf.c and definition.yaml names.
```

### How to use 

0. Create a folder and put your `program.bpf.c` and `definition.yaml` files there (or go to any existing one under gadgets/)

```bash
$ cd mygadget
$ ls
definition.yaml  program.bpf.c
```

1. Build the gadget image 

#### Using local installed clang

a. Install headers in the system

```bash
$ sudo make install-headers
```

b. Build the image

```bash
$ sudo -E ig image build --local -t mygadget:latest .
Successfully built docker.io/library/mygadget:latest@sha256:3a23c1f08a8b722ca49ec00cf2a1ca1373a87ea7ac52c23992e55bafe1bc9542
```

#### Using builder container 

In this case you don't need to have any clang nor related dev tool installed on your machine, only docker is needed
Note: --builder is needed as the upstream builder will only be updated when this PR is merged

```
$ sudo -E ig image build --builder-image=ghcr.io/mauriciovasquezbernal/inspektor-gadget-ebpf-builder:latest -t mygadget:latest .
Successfully built docker.io/library/mygadget:latest@sha256:842e69c79177908b6998737b86fc691e8fc0b3e45e2030cafcb362cbfcb1c039
``` 

3. Run

```bash
$ sudo -E ig run mygadget:latest
....
```

### Note

This is a mega PR including changes from multiple PRs, the main purpose of this is to check that everything is in place and working together. The following PRs should be merged before:

- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/2015
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/2022
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/2031

### TODO
- [x] Provide builder container image
- [ ] Documentation
- [x] Update Github Actions to build and push gadgets
- [x] How to test it? (cc @mqasimsarfraz given that we're taking about this)

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/2006


gadgets/Makefile
```diff
@@ -1,24 +1,36 @@
...
-	clang -target bpf -Wall -g -O2 -D __TARGET_ARCH_x86 -c $(filter %.c,$^) \
-		-I ../include/ \
-		-I ../include/gadget/amd64/ \
-		-o $@
-%_arm64.bpf.o : %.bpf.c
-	clang -target bpf -Wall -g -O2 -D __TARGET_ARCH_arm64 -c $(filter %.c,$^) \
-		-I ../include/ \
-		-I ../include/gadget/arm64/ \
-		-o $@
+#TODO: How to create an independent target for each gadget do this can be run in parallel?
```
burak-ok, line 18: Suggestion: https://github.com/inspektor-gadget/inspektor-gadget/commit/797e13ad926acfd2a58c34d8df77da45112a818b

Also the TODOs from that commit:
1. We rely here on `ig`. We should build it first -> make dependency. But these are different makefiles
2. We need the path to the `ig` executable we built in 1.
3. The output when using `make build -j` (parallel build) is not that great

.github/actions/set-container-repo-and-determine-image-tag/action.yml
```diff
@@ -58,3 +64,30 @@ runs:
...
+        echo "gadgets-repo=${gadgets_repo}" >> $GITHUB_OUTPUT
+    - name: Output gadgets image tag
+      id: gadgets-tag
+      shell: bash
+      run: |
+        # If GITHUB_REF_NAME is 'foo/bar', image_tag will be 'foo-bar', we need
+        # this because it is not possible to have slash in image tag.
+        gadgets_tag=${GITHUB_REF_NAME//\//-}
+        if [ "$gadgets_tag" = "main" ]; then
+            gadgets_tag="latest"
```
burak-ok, line 90: I don't know the convention about the `latest` tag.
As an end user I always think its the latest release version. Maybe something like `head` or `dev` is better. On release we tag it with `latest`.

Another (maybe) issue:
What if I call my branch `latest` or some version number `v0.20.0`

cmd/common/image/build.go
```diff
@@ -17,80 +17,202 @@ package image
...
+
+	if err := os.Chdir(opts.path); err != nil {
+		return fmt.Errorf("changing directory: %w", err)
+	}
+
+	if opts.local {
+		buildCmd := exec.Command(
+			"/bin/sh", "-c",
+			string(buildScript),
+			"", // TODO: why is it needed?
```
burak-ok, line 153: According to the [man page of `sh`](https://linux.die.net/man/1/sh):

> If the -c option is present, then commands are read from string. If there are arguments after the string, they are assigned to the positional parameters, starting with $0.


Never saw embedding a whole script into a sh executing like that. Didn't knew this was possible :D

burak-ok (commented):
Looks great after the first pass.
I still have questions though:

1. Is the builder image update automatically (on release)?
2. I can't seem to get the local build to work (I executed `sudo make install-headers`):
```bash
~/codes/inspektor-gadget-3/gadgets/trace/tcpconnect:burak/cg/build_makefile ➭ go run --exec "sudo -E" ../../../cmd/ig/... image build testfoo:latest -l
Error: build script: exit status 1: + inputFile=program.bpf.c
+ outputDir=/tmp/gadget-build2525032023
+ cflags=
+ clang -target bpf -Wall -g -O2 -D __TARGET_ARCH_x86 -c program.bpf.c -I /usr/include/gadget/amd64/ -o /tmp/gadget-build2525032023/x86.bpf.o
program.bpf.c:7:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.

exit status 1
```

mauriciovasquezbernal:
> * Is the builder image update automatically (on release)?

It's not. I think it should be updated on release and also when merging to main, but IMO it can be done in a different PR.

> * I can't seem to get the local build to work (I executed `sudo make install-headers`):

You also need libbpf headers `git clone --branch ${LIBBPF_VERSION} --depth 1 https://github.com/libbpf/libbpf.git libbpf/src && make install_headers`

.github/actions/set-container-repo-and-determine-image-tag/action.yml
mauriciovasquezbernal: > As an end user I always think its the latest release version. Maybe something like head or dev is better. 

AFAUI `latest` is used when no tag is specified, I'm not sure there is a standard convention to indicate it's the latest tagged release. 

>On release we tag it with latest.


Using `latest` isn't recommended , users should stick to a fixed version, so I think we can continue using latest for development. 

> What if I call my branch latest or some version number v0.20.0

I think there are other cases where this can get broken, but IMO they're not so relevant. We're following the same approach for other Inspektor Gadget images.

blanquicet:
> You also need libbpf headers git clone --branch ${LIBBPF_VERSION} --depth 1 https://github.com/libbpf/libbpf.git libbpf/src && make install_headers
>
This needs to be documented. In general, we are still missing all the documentation regarding the building. Even if we still change it in the future (I know it is heavily under development), we will ship this in the next release, so we need at least a minimum of documentation, don't you think?

mauriciovasquezbernal:
> so we need at least a minimum of documentation, don't you think?

I agree, but this is only relevant when the `--local` parameter is set (which is false by default). Btw, this is the exact same case if you try `make ebpf-objects-outside-docker`.

gadgets/Makefile
mauriciovasquezbernal: Thanks for the suggestion. I addresses the TODOs by adding a rule on the top makefile.

mauriciovasquezbernal:
Closing in favor of https://github.com/inspektor-gadget/inspektor-gadget/pull/2015 as I was getting crazy handle these two PRs separately.

PR #2043, opened 2023-09-15
ci: Fix uploading ig container image as artifact

mauriciovasquezbernal:
Fixes: 463f5a88fddd ("ci: Publish ig container images.")

It was causing this step to fail with `Warning: No files were found with the provided path: /tmp/ig-container-image-linux-amd64.tar. No artifacts will be uploaded.`

Ref https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/6184453746/job/16790129503

It wasn't causing any other issue as these artifacts are not being used anywhere.


mauriciovasquezbernal:
> Hum... So you are basically telling me the `ig` image was never updated?

No, the artifact is never used, however it's still useful for manual debugging in some cases and testing in the future.

PR #2044, opened 2023-09-15
btfgen: Embed btf generated files into the binary

### LINKED ISSUE #752: local-gadget: Use BTFGen
Inspektor Gadget got support for BTFGen on https://github.com/kinvolk/inspektor-gadget/pull/387, however local-gadget doesn't use it. local-gadget is shipped as binary, hence we need to embed the generated BTF files there. I think we can follow a very similar approach than the one used in the BCC (https://github.com/iovisor/bcc/pull/3889):
- Generate the BTF files using a makefile (probably the existing [Makefile.btfgen](https://github.com/kinvolk/inspektor-gadget/blob/main/Makefile.btfgen) can be used)
- Compress the generated files on a tar.xz or similar file
- Use `go:embed` to include it within the binary 
- Use [`archive/tar`](https://pkg.go.dev/archive/tar) to uncompress it a runtime based on the running distribution / kernel version

mauriciovasquezbernal:
This commit changes the approach regarding btfgen, now the btf generated files are embeded in the golang binary, which means that now it's also available for ig.

This also updates the CI to execute btfgen before building ig or ig-k8s.

### Getting a host without BTF

Ironically it's difficult to get a machine without BTF support these days. I used a ubuntu focal64 through vagrant (`config.vm.box = "ubuntu/focal64"`). It ships a kernel version that already has btf support, hence I installed another kernel that doesn't include BTF and rebooted the machine:

```
$ uname -r
5.4.0-99-generic
$ sudo apt-get install linux-image-5.8.0-45-generic -y
...
```

After reboot

```
$ uname -r 
5.8.0-45-generic

$ cat /boot/config-$(uname -r) | grep -i btf
CONFIG_VIDEO_SONY_BTF_MPX=m
# CONFIG_DEBUG_INFO_BTF is not se
```

Install docker (it should not be required, I'll open an issue soon about this)

```bash
$ sudo apt-get install docker.io -y
```

### Before this PR 

```bash
$ make ig-linux-amd64
...

$ ls -lh ig-linux-amd64 
-rwxr-xr-x 1 mvb mvb **102M** Sep 15 11:13 ig-linux-amd64

$ sudo ./ig-linux-amd64 trace open --host
WARN[0000] Failed to initialize BTF: open btfs/x86/ubuntu/20.04/x86_64/5.8.0-45-generic.btf: file does not exist 
Error: initializing operators: initializing operator "LocalManager": initializing manager: starting container fanotify: readPrivateDataFromFd: loading maps and programs: field IgFgetX: program ig_fget_x: apply CO-RE relocations: load kernel spec: no BTF found for kernel version 5.8.0-45-generic: not supported
```

#### Option A: Locally compiling ig with btf enabled

```bash
$ make ENABLE_BTFGEN=true ig-linux-amd64 -j$(nproc)
#  some minutes after 

$ ls -lh ig-linux-amd64 
-rwxr-xr-x 1 mvb mvb **104M** Sep 15 11:09 ig-linux-amd64
                       ^^^ binary is only ~2MB bigger.

$ sudo ./ig-linux-amd64 trace open --host
RUNTIME.CONTAINERNAME                                      PID        COMM             FD    ERR PATH                                                         
                                                           3936       ig-linux-amd64   44    0   /proc                                                        
                                                           3936       ig-linux-amd64   44    0   /proc                                                        
                                                           3936       ig-linux-amd64   44    0   /proc                                                        
                                                           572        multipathd       9     0   /sys/devices/pci0000:00/0000:00:14.0/host2/target2:0:0/2:0:0…
                                                           572        multipathd       9     0   /sys/devices/pci0000:00/0000:00:14.0/host2/target2:0:0/2:0:0…
                                                           416        systemd-journal  37    0   /proc/572/status
```

#### Option B: Getting ig from a CI run 

ig is compiled in the CI with btfgen enabled, hence downloading the ig binary from a run of this PR should work too.

Fixes #752


mauriciovasquezbernal:
> I would advice to split the commit by having one which adds the `btfgen` package and the other using it.

I wanted to do that but given it involves changes to the CI, makefile and gadges, it's difficult to split in logic pieces that work alone, so I prefer to have a big commit with all changes in.

pkg/btfgen/btfgen.go
```diff
@@ -0,0 +1,139 @@
...
+	ID        string
+	VersionID string
+	Arch      string
+	Kernel    string
+}
+
+func getOSInfo() (*osInfo, error) {
+	osInfo := &osInfo{}
+
+	file, err := os.Open("/etc/os-release")
```
alban, line 100: Should this use /host?

```suggestion
	file, err := os.Open(filepath.Join(host.HostRoot, "/etc/os-release"))
```

pkg/btfgen/btfgen.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("opening file: %w", err)
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		parts := strings.Split(line, "=")
```
alban, line 109:
```suggestion
		parts := strings.SplitN(line, "=", 2)
```

Just in case some version strings contain the `=` symbol.

alban (commented):
Initial comments. I'll continue soon.

Makefile.btfgen
```diff
@@ -11,11 +11,18 @@ ifndef ARCH
...
+# - bpf2go uses x86 and arm64
+ARCH_CLEANED = $(subst amd64,x86,$(ARCH))
+OUTPUT ?= pkg/btfgen/btfs
+
+SOURCE_BTF_FILES = $(shell find $(BTFHUB_ARCHIVE)/ -iregex ".*$(subst x86,x86_64,$(ARCH_CLEANED)).*" -type f -name '*.btf.tar.xz')
+MIN_CORE_BTF_FILES = $(patsubst $(BTFHUB_ARCHIVE)/%.btf.tar.xz, $(OUTPUT)/$(ARCH_CLEANED)/%.btf, $(SOURCE_BTF_FILES))
+BPF_ALL_O_FILES = $(shell find pkg/ -type f -regex ".*\($(ARCH_CLEANED)\|bpfel\).o")
 # Filter out BPF objects that only contain BPF maps without BPF programs
-BPF_PROGS_O_FILES = $(filter-out pkg/gadgets/trace/network/tracer/graphmap_bpfel%,$(BPF_ALL_O_FILES))
+BPF_PROGS_O_FILES = $(filter-out pkg/gadgettracermanager/containers-map/containersmap_bpfel%,$(BPF_ALL_O_FILES))
```
alban, line 25: Similarly to the discussion in https://github.com/inspektor-gadget/inspektor-gadget/pull/2003#discussion_r1319814545, the map from the pkg/gadgettracermanager/containers-map package could be created programmatically.

But that can be in a different PR.

pkg/gadgets/internal/networktracer/tracer.go
```diff
@@ -116,6 +117,9 @@ func (t *Tracer[Event]) newAttachment(
 		MapReplacements: map[string]*ebpf.Map{
 			"tail_call": t.dispatcherMap,
 		},
+		Programs: ebpf.ProgramOptions{
+			KernelTypes: btfgen.GetBTFSpec(),
```
alban, line 121: The dispatcher program is a socket filter that only uses the stable Linux API, it should need BTF.

I am worried that setting `KernelTypes` would slow down the performances unnecessarily.

pkg/gadgets/internal/socketenricher/tracer.go
```diff
@@ -73,6 +74,12 @@ func (se *SocketEnricher) start() error {
 
 	kallsyms.SpecUpdateAddresses(specIter, []string{"socket_file_ops"})
 
+	opts := ebpf.CollectionOptions{
+		Programs: ebpf.ProgramOptions{
+			KernelTypes: btfgen.GetBTFSpec(),
```
alban, line 79: Happy to see that the socket enricher can use btfgen!

alban (commented):
One blocking comment about the network dispatcher.

The rest looks good to me.

I didn't review the CI part.

Makefile.btfgen
mauriciovasquezbernal: Right, the difference is that in this case providing BTF information manually for the map is not that easy, and in this case it could be appreciated to pretty print the contents of the map while debugging.

pkg/gadgets/internal/networktracer/tracer.go
mauriciovasquezbernal: Makes sense. I'll remove custom BTF for this program.

pkg/btfgen/btfgen.go
mauriciovasquezbernal: Good catch, thanks!

PR #2054, opened 2023-09-18
Replaced all libseccomp usage

amitschendel:
Removed libseccomp-go usage by creating a standalone go package for syscalls translation.

## How to use

## Testing done
Ran the tests locally and ran the gadget.

## TODO
Remove all the libseccomp-dev requirements from the code.
I didn't want to touch it because it involves some CI files and I don't want to do damage there.


matthyx:
@amitschendel few questions:

- why did you create your own repo instead of putting csv and parsers directly inside inspektor?
- native [csv](https://pkg.go.dev/encoding/csv) package isn't good enough?

amitschendel:
@matthyx Thanks for the feedback,
1. The idea behind creating a repo is to out source this functionality so others can contribute to it more easily and use it in other projects as well. Do you think it's better to put it directly in ig?
2. The native csv package is good but it's easier to work with csv files with the used package, I can change it if you think it's bad.

matthyx:
> Also, instead of using custom baked CSV, what do you think of using `libseccomp` [one](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.csv)?

I think @mauriciovasquezbernal fears the implication of LGPL code inside the project... I don't think any of us is a lawyer, but if someone can prove it's safe I would rather use this file too :)

matthyx:
> @matthyx Thanks for the feedback,
> 
> 1. The idea behind creating a repo is to out source this functionality so others can contribute to it more easily and use it in other projects as well. Do you think it's better to put it directly in ig?
> 2. The native csv package is good but it's easier to work with csv files with the used package, I can change it if you think it's bad.

let's see what @mauriciovasquezbernal says

amitschendel:
I have used this [source](https://github.com/hrw/syscalls-table) for creating the syscalls tables.
And it seems to be even more updated then libseccomp-golang table.
But if you want to switch lmk.

mauriciovasquezbernal:
Regarding LGPL, I also think that should not be an issue, but I prefer to be totally sure and not to use that from libseccomp.

> The idea behind creating a repo is to out source this functionality so others can contribute to it more easily and use it in other projects as well. Do you think it's better to put it directly in ig?

I'd rather have this on this repository. Later on we can move that to a different one if needed.
My reasons are:
- Probably it's easier to maintain inside this org.
- We don't have to worry a lot by breaking its API until we have a version we like.

> The native csv package is good but it's easier to work with csv files with the used package, I can change it if you think it's bad.

Actually I had something different in mind to implement it. What about hardcoding this directly in golang by using a map, so making this translation is only a matter of a map lookup instead of having to read a file?

There could be a file for each different architecture:

```
//syscalls_amd64.go
syscallsMap[int]string = 
....
}
```

Btw, the current solution doesn't work fine as the csv file is not embeded in the binary:

```bash
$ cat cosa.go 
package main

import (
	"fmt"

	"github.com/amitschendel/syscalls/pkg/syscalls"
)

func main() {
	fmt.Println(syscalls.GetNumberByName("", "openat"))
	fmt.Println(syscalls.GetNameByNumber("arm64", 56))
}

$ go run cosa.go
0 open artifacts/syscalls-x86_64.csv: no such file or directory
 open artifacts/syscalls-arm64.csv: no such file or directory
```

mauriciovasquezbernal:
I think https://github.com/hrw/syscalls-table/blob/master/templates/syscalls.html.j2 could be modified to directly generate the golang code, but perhaps it's too much and complicated for this PR?

matthyx:
> I think https://github.com/hrw/syscalls-table/blob/master/templates/syscalls.html.j2 could be modified to directly generate the golang code, but perhaps it's too much and complicated for this PR?

I suggest we do the first iteration by hand with the suggested files in https://github.com/inspektor-gadget/inspektor-gadget/pull/2054#issuecomment-1723907803 and then we can check how to automate this.

pkg/utils/syscalls/amd64/syscalls.go
```diff
@@ -0,0 +1,968 @@
+package amd64
```
mauriciovasquezbernal, line 1: I think it's better to have these two files (amd64/syscalls.go and arm64/syscalls.go) on the same folder, then we can use build tags to only compile the one that is relevant for the current architecture. Something like for this file:

```
//go:build linux amd64
// +build linux amd64
```

pkg/utils/syscalls/syscalls-helper.go
```diff
@@ -0,0 +1,46 @@
...
+	"runtime"
+
+	amd64syscalls "github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls/amd64"
+	arm64syscalls "github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls/arm64"
+)
+
+func GetSyscallNumberByName(arch string, name string) (int, bool) {
+	if arch == "" {
+		arch = runtime.GOARCH
+	}
```
mauriciovasquezbernal, line 13: I don't think we need this parameter. We're only interested in getting that information for the current architecture, so I'll remove this and use the conditional compilation logic described in my other comment.

pkg/utils/syscalls/syscalls-helper.go
```diff
@@ -0,0 +1,46 @@
+package syscalls
```
mauriciovasquezbernal, line 1: nit: please don't forget to add the license header as we have it on other files.

pkg/utils/syscalls/amd64/syscalls.go
```diff
@@ -0,0 +1,968 @@
...
+
+// This updated to kernel 6.6-rc2
+var SyscallsNameToNumber = map[string]int{
+	"_llseek":                      -1,
```
mauriciovasquezbernal, line 5: What does it mean to have `-1`, that the syscalls is not available on this arch? if yes, I think we could remove them.

pkg/utils/syscalls/amd64/syscalls.go
```diff
@@ -0,0 +1,968 @@
+package amd64
+
+// This updated to kernel 6.6-rc2
+var SyscallsNameToNumber = map[string]int{
```
mauriciovasquezbernal, line 4: nit: make this map unexported.

pkg/utils/syscalls/arm64/syscalls.go
```diff
@@ -0,0 +1,918 @@
...
+	"vm86old":                      -1,
+	"vmsplice":                     75,
+	"wait4":                        260,
+	"waitid":                       95,
+	"waitpid":                      -1,
+	"write":                        64,
+	"writev":                       66,
+}
+
+var SyscallsNumberToName = map[int]string{
```
mauriciovasquezbernal, line 610: nit: is it possible to get this ordered by number?

pkg/utils/syscalls/amd64/syscalls.go
```diff
@@ -0,0 +1,968 @@
...
+	132: "utime",
+	280: "utimensat",
+	235: "utimes",
+	58:  "vfork",
+	153: "vhangup",
+	278: "vmsplice",
+	61:  "wait4",
+	247: "waitid",
+	1:   "write",
+	20:  "writev"}
```
mauriciovasquezbernal, line 968: nit: put the `}` in a new line as done in other places.

mauriciovasquezbernal (commented):
It's looking very good. Some comments about using go build tags to only compile the code for the current architecture.

pkg/utils/syscalls/amd64/syscalls.go
amitschendel: Done

pkg/utils/syscalls/syscalls-helper.go
amitschendel: Done

pkg/utils/syscalls/syscalls-helper.go
amitschendel: Done

pkg/utils/syscalls/amd64/syscalls.go
amitschendel: Done

pkg/utils/syscalls/arm64/syscalls.go
amitschendel: Any special reason for this?

pkg/utils/syscalls/syscalls_amd64.go
```diff
@@ -0,0 +1,741 @@
+//go:build linux && amd64
```
matthyx, line 1: I think this build constraint is by default, as we saw in the official doc https://pkg.go.dev/cmd/go#hdr-Build_constraints

matthyx: yes please, since they bring no value

pkg/utils/syscalls/arm64/syscalls.go
mauriciovasquezbernal: I was thinking that it'll be easier to understand what is the latest syscall that was added, but `syscallsNameToNumber` is ordered by name so that's good enough.

pkg/utils/syscalls/syscalls_amd64.go
amitschendel: Done

pkg/gadgets/advise/seccomp/tracer/tracer.go
```diff
@@ -23,7 +23,7 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
-	libseccomp "github.com/seccomp/libseccomp-golang"
+	syscalls "github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls"
```
mauriciovasquezbernal, line 26: nit: (same in other places)

```suggestion
	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls"
```

pkg/utils/syscalls/syscalls-helper.go
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+func GetSyscallNumberByName(name string) (int, bool) {
+	var number int
+	var ok bool
+
+	number, ok = syscallsNameToNumber[name]
+
+	return number, ok
```
mauriciovasquezbernal, line 23: nit: (same in GetSyscallNameByNumber)

```suggestion
	return syscallsNameToNumber[name]
```

pkg/utils/syscalls/syscalls_amd64.go
```diff
@@ -0,0 +1,738 @@
...
+	132: "utime",
+	280: "utimensat",
+	235: "utimes",
+	58:  "vfork",
+	153: "vhangup",
+	278: "vmsplice",
+	61:  "wait4",
+	247: "waitid",
+	1:   "write",
+	20:  "writev"
```
mauriciovasquezbernal, line 737: missing `,` at the end of the line: "pkg/utils/syscalls/syscalls_amd64.go:737:15: syntax error: unexpected newline in composite literal; possibly missing comma or }"

mauriciovasquezbernal (commented):
Some more comments. I'll kick the CI and approve if everything is fine after.

pkg/utils/syscalls/syscalls-helper.go
mauriciovasquezbernal, line 1: I'll just name this file syscalls.go.

pkg/utils/syscalls/syscalls_amd64.go
matthyx: I have the same error

amitschendel: my linter missed it...

pkg/utils/syscalls/syscalls-helper.go
amitschendel: This doesn't work for some reason, it unpacks only the value but not the "ok" var unless you do it as I do it.

pkg/utils/syscalls/syscalls-helper.go
amitschendel: done

pkg/gadgets/trace/capabilities/tracer/tracer.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
 	"github.com/cilium/ebpf/perf"
-	libseccomp "github.com/seccomp/libseccomp-golang"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls"
```
mauriciovasquezbernal, line 28: I'm really sorry I forgot about this. We're trying to keep imports in blocks:
- stantard libraries
- external dependencies
- internal packages

Could you please move it to the block below where all the `github.com/inspektor-gadget/inspektor-gadget` imports are? (same for other places)

pkg/utils/syscalls/syscalls.go
```diff
@@ -0,0 +1,33 @@
+package syscalls
```
mauriciovasquezbernal, line 1: nit: move package name after license.

pkg/utils/syscalls/syscalls-helper.go
mauriciovasquezbernal: You're right. 

Perhaps 

```	
	number, ok := syscallsNameToNumber[name]
	return number, ok
``` 

is a bit more concise?

mauriciovasquezbernal (commented):
Great. Would you mind fixing the last nits and it'll be good to go once the CI passes? 

Thanks a lot for the contribution!

mauriciovasquezbernal:
@amitschendel  final thing: please squash all the commits into one in order to merge.

amitschendel:
OOPS I went too far with reset :(
I will try to fix it

mauriciovasquezbernal (approved):
I rebased to main and force pushed. LGTM, thanks a lot for the contribution!

PR #2056, opened 2023-09-18
cmd: Introduce runtime and k8s columns

mqasimsarfraz:
#1702 added support for k8s metadata in JSON/YAML but the support for columns was still missing. This PR allows the same information to be available as columns as well. 

Related: #737
Fixes: #2066 

## Idea

- Before this change both `ig` and `kubectl-gadget` would filter `k8s` and `runtime` columns respectively rather then hiding. So we will switch to marking them hidden as needed by each client.
- The hidden columns can be displayed using `-o columns` as needed by the user.
- `list-containers` needs an extra work but `k8s` metadata can also be displayed using `-o custom-columns`

## Tesing done

```
$ make minikube-start-containerd
$ make ig
$ docker cp ig minikube-containerd:/bin/ig
$ docker exec -it minikube-containerd bash
$ ig list-containers # same as before
RUNTIME.RUNTIMENAME RUNTIME.CONTAINERID                                              RUNTIME.CONTAINERNAME   RUNTIME.CONTAINERIMAGENAME                                 
containerd          0a9cbc23a8409cf015e389fa71f2422f28dbea1129e66a40e38685fb67424ef6 coredns                 k8s.gcr.io/coredns/coredns:v1.8.6                          
containerd          5e48d5e717d89f2237eb0b364d9c8b6265f949e5600f642d55420e1a4922e2e2 etcd                    k8s.gcr.io/etcd:3.5.3-0 
$ ig list-containers -o custom-columns=runtime.containerid,runtime.containername,k8s.pod,k8s.namespace
RUNTIME.CONTAINERID                                              RUNTIME.CONTAINERNAME   K8S.POD                                     K8S.NAMESPACE
0a9cbc23a8409cf015e389fa71f2422f28dbea1129e66a40e38685fb67424ef6 coredns                 coredns-6d4b75cb6d-94hkg                    kube-system  
5e48d5e717d89f2237eb0b364d9c8b6265f949e5600f642d55420e1a4922e2e2 etcd                    etcd-minikube-containerd                    kube-system

$ ig trace open -o columns=runtime.containername,k8s.pod,k8s.namespace,pid,comm
RUNTIME.CONTAINERNAME                                                K8S.POD                                                             K8S.NAMESPACE                                                       PID        COMM            
gadget                                                               gadget-p4kvb                                                        gadget                                                              448891     runc:[2:INIT]   
gadget                                                               gadget-p4kvb                                                        gadget                                                              448891     runc:[2:INIT]   
```

### DIFF
--- a/cmd/common/registry.go
+++ b/cmd/common/registry.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/frontends"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/frontends/console"
-	cols "github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
 	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
 	gadgetregistry "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-registry"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
@@ -47,7 +46,7 @@ const (
 )
 
 // AddCommandsFromRegistry adds all gadgets known by the registry as cobra commands as a subcommand to their categories
-func AddCommandsFromRegistry(rootCmd *cobra.Command, runtime runtime.Runtime, columnFilters []cols.ColumnFilter) {
+func AddCommandsFromRegistry(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumnTags []string) {
 	runtimeGlobalParams := runtime.GlobalParamDescs().ToParams()
 
 	// Build lookup
@@ -95,16 +94,16 @@ func AddCommandsFromRegistry(rootCmd *cobra.Command, runtime runtime.Runtime, co
 		}
 		categoryCmd.AddCommand(buildCommandFromGadget(
 			gadgetDesc,
-			columnFilters,
 			runtime,
 			runtimeGlobalParams,
 			operatorsGlobalParamsCollection,
 			gadgetInfo.OperatorParamsCollection.ToParams(),
+			hiddenColumnTags,
 		))
 	}
 }
 
-func buildColumnsOutputFormat(gadgetParams *params.Params, parser parser.Parser) gadgets.OutputFormats {
+func buildColumnsOutputFormat(gadgetParams *params.Params, parser parser.Parser, hiddenColumnTags []string) gadgets.OutputFormats {
 	paramTags := make(map[string]string)
 	if gadgetParams != nil {
 		for _, param := range *gadgetParams {
@@ -144,7 +143,7 @@ func buildColumnsOutputFormat(gadgetParams *params.Params, parser parser.Parser)
 		}
 		fmt.Fprintf(&out, "\n")
 	}
-	fmt.Fprintf(&out, "    Default columns: %s\n", strings.Join(parser.GetDefaultColumns(), ","))
+	fmt.Fprintf(&out, "    Default columns: %s\n", strings.Join(parser.GetDefaultColumns(hiddenColumnTags...), ","))
 
 	of.Description += out.String()
 
@@ -173,11 +172,11 @@ func buildOutputFormatsHelp(outputFormats gadgets.OutputFormats) []string {
 
 func buildCommandFromGadget(
 	gadgetDesc gadgets.GadgetDesc,
-	columnFilters []cols.ColumnFilter,
 	runtime runtime.Runtime,
 	runtimeGlobalParams *params.Params,
 	operatorsGlobalParamsCollection params.Collection,
 	operatorsParamsCollection params.Collection,
+	hiddenColumnTags []string,
 ) *cobra.Command {
 	var outputMode string
 	var filters []string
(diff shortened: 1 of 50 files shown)


pkg/container-collection/containers.go
```diff
@@ -211,6 +211,11 @@ func GetColumns() *columns.Columns[Container] {
 	col, _ = cols.GetColumn("runtime.containerImageName")
 	col.Visible = true
 
+	// hide the k8s column
+	for _, c := range cols.GetColumnMap(columns.WithTag("kubernetes")) {
+		c.Visible = false
+	}
+
```
mauriciovasquezbernal, line 218: Use similar logic to set to visible runtime columns above?

mauriciovasquezbernal (commented):
It looks great. I don't know if having the `K8s.` prefix in kubectl-gadget has been discussed, IMO that's good:

```bash
$ ./kubectl-gadget trace open
K8S.NODE                  K8S.NAMESPACE             K8S.POD                  K8S.CONTAINER            PID        COMM         FD ERR P
```

I didn't find anything to be fixed in the code, only some times that are missing:
- It's missing to update all gadgets guides (sorry, I now it's to annoying to do it, perhaps a good regexp to replace can help)
- Prometheus documentation should be updated as well 
https://github.com/inspektor-gadget/inspektor-gadget/blob/146bc5afd7669f68f665d977efac60940f33a4f9/docs/gadgets/prometheus.md#L95-L98
(other places in same file)

btw, this will also fix https://github.com/inspektor-gadget/inspektor-gadget/issues/2066.

cmd/common/registry.go
```diff
@@ -47,7 +47,7 @@ const (
 )
 
 // AddCommandsFromRegistry adds all gadgets known by the registry as cobra commands as a subcommand to their categories
-func AddCommandsFromRegistry(rootCmd *cobra.Command, runtime runtime.Runtime, columnFilters []cols.ColumnFilter) {
+func AddCommandsFromRegistry(rootCmd *cobra.Command, runtime runtime.Runtime, columnFilters []cols.ColumnFilter, hiddenColumnTags []string) {
```
blanquicet, line 50: The `columnFilters` here and in `buildCommandFromGadget` is not needed anymore.

cmd/common/registry.go
```diff
@@ -420,6 +422,10 @@ func buildCommandFromGadget(
 						}
 					}
 				}
+				// hide columns by tag (e.g. kubernetes, runtime) if requested by the caller
+				if len(hiddenColumnTags) > 0 {
+					hiddenTags = append(hiddenTags, hiddenColumnTags...)
+				}
 				requestedColumns = parser.GetDefaultColumns(hiddenTags...)
```
blanquicet, line 429: We should pass these `hiddenTags` also [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/0d5ed51e1f9e0d67d37b1d3e6469ebd328838e44/cmd/common/registry.go#L148). Otherwise, we are showing the incorrect default columns in the `help` command:

```bash
$ sudo ./ig trace open -h
Trace open system calls
...
  -o, --output string          Output format (columns, json, jsonpretty, yaml).

                               Columns (columns)
                                   Available columns:
                                     comm
                                     err
                                     fd
                                     flags
                                     fullPath (requires --full-path)
                                     gid
                                     k8s.container
                                     k8s.hostnetwork
                                     k8s.namespace
                                     k8s.node
                                     k8s.pod
                                     mntns
                                     mode
                                     path
                                     pid
                                     ret
                                     runtime.containerId
                                     runtime.containerImageName
                                     runtime.containerName
                                     runtime.runtimeName
                                     timestamp
                                     uid
                                   Default columns: runtime.containerName,k8s.node,k8s.namespace,k8s.pod,k8s.container,pid,comm,fd,err,path,fullPath
```

blanquicet (approved):
I like the approach of hiding the columns instead of filtering them. It should also make the implementation of the `--k8s` and `--runtime` flags easier. BTW, given that you will add that in a second PR, this one is not fully fixing https://github.com/inspektor-gadget/inspektor-gadget/issues/737.

LGTM after managing the comments.

mauriciovasquezbernal:
> It should also make the implementation of the `--k8s` and `--runtime` flags

I'm wondering about what's the idea there. Provide independent flags to enable the runtime and k8s columns? IMO we should avoid adding so many flags, what about having something like `-o columns=runtime`? (we can try to find a different syntax for making it obvious it's for the full "kind" of columns.)

mqasimsarfraz:
> It's missing to update all gadgets guides (sorry, I now it's to annoying to do it, perhaps a good regexp to replace can help)

Done. Agreed, it wasn't fun to do but I realized we were missing `RUNTIME` columns for `ig` in docs as well so thanks for catching it!

blanquicet:
> > It should also make the implementation of the `--k8s` and `--runtime` flags
> 
> I'm wondering about what's the idea there. Provide independent flags to enable the runtime and k8s columns? IMO we should avoid adding so many flags, what about having something like `-o columns=runtime`? (we can try to find a different syntax for making it obvious it's for the full "kind" of columns.)

My initial idea was to add `--k8s` only to `ig` and `--runtime` only to `kubectl-gadget`, but it was just my proposal, we can ask for opinions in the slack channel 🙂.

cmd/ig/containers/containers.go
```diff
@@ -86,6 +86,10 @@ func NewListContainersCmd() *cobra.Command {
 			col.Visible = true
 			col, _ = cols.GetColumn("runtime.containerImageName")
 			col.Visible = true
+			// hide the k8s column
+			for _, c := range cols.GetColumnMap(columns.WithTag("kubernetes")) {
+				c.Visible = false
+			}
```
blanquicet, line 92: Please synch this with what you did in `pkg/container-collection/containers.go`. Maybe you could even unify it in a function to avoid having to update it in both places each time.

mqasimsarfraz: Done

mqasimsarfraz:
Thanks for the reviews!

PR #2061, opened 2023-09-20
pkg: Remove vmlinux-cgo.h.

eiffel-fl:
This file was not mandatory as we can use __type instead, so let's remove it.


mauriciovasquezbernal (approved):
Nice! Less code to take care about.

PR #2064, opened 2023-09-20
Dockerfiles: Remove C cross compiler and C libraries.

eiffel-fl:
As we no more use Cgo, we can get rid of C cross compiler and corresponding libraries.


mauriciovasquezbernal (approved):
I like it. It helps to simplify the build process a lot.

Issue #1926, opened 2023-08-03
Design (and implement?) support for Prometheus metrics in image-based gadgets


mauriciovasquezbernal:
The design of the prometheus support is still missing in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md#prometheus. 

This should be probably done after https://github.com/inspektor-gadget/inspektor-gadget/issues/1598 to apply all the learning from there.

mauriciovasquezbernal:
Closing in favor of #1598.

PR #2065, opened 2023-09-20
run: expose prometheus metrics

alban:
Allow containerized gadgets to expose Prometheus metrics.

This PR makes the feature work and enable the configuration file to select the gadget. Hopefully the user interface will be improved later on.

## How to use

Prepare my-prometheus-config.yaml as follows:
```
metrics_name: guide
metrics:
  - name: dns_traffic
    type: counter
    category: ""
    gadget: run
    gadgetArgs:
      - albantest.azurecr.io/trace-dns:v3
    labels:
      - namespace
      - pod
      - container
```

```
$ sudo -E ig prometheus --config @my-prometheus-config.yaml
INFO[0000] Publishing metrics...                        
```

## Testing done

Tested in this branch:
https://github.com/alban/inspektor-gadget/tree/alban_asg2023

## TODO

I am not sure if this should be done in this PR or later on:

- [ ] Exposing the gadget name via `category=""`, `gadget=run` & `gadgetArgs` is awkward. The configuration file should be reworked.

### DIFF
--- a/docs/gadgets/prometheus.md
+++ b/docs/gadgets/prometheus.md
@@ -33,6 +33,21 @@ metrics:
       # defines the granularity of the labels to capture. See below.
 ```
 
+To use your own gadget, you can refer to the gadget as follows:
+
+```yaml
+metrics_name: dns_traffic
+metrics:
+  - name: dns_traffic
+    type: counter
+    category: ""
+    gadget: run
+    gadgetArgs:
+      - myregistry.azurecr.io/trace-dns:v1
+    labels:
+      # defines the granularity of the labels to capture. See below.
+```
+
 ### Filtering (aka Selectors)
 
 It's possible to configure Inspektor Gadget to only update metrics for some specific labels. This is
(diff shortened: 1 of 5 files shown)


blixtra:
I'd like us to take a moment to think about all the eventual export types. We are using the OTel library for the Prometheus export and we should assume that we'll support the other formats/destinations that the OTel library provides. My current understanding is that we will support the following...

- Prometheus
- OTel traces
- Logs
- gRPC API

Probably more but I'd think those cover a large part of what would be useful.

So, for me it's important that we make sure the metadata supports a way to define the various exports. It could be as easy as.

```md
...
export:
  -type: prometheus
     [details here]
  -type: logs
     [details here]
...
```

mauriciovasquezbernal:
I agree with Chris, I think we should hold this PR a bit until we handle https://github.com/inspektor-gadget/inspektor-gadget/issues/1926

mauriciovasquezbernal:
I propose to close this one as we need to think better the design of this feature.

PR #2067, opened 2023-09-21
pkg/gadget-service: migrate gRPC API of the gadget-service to a separate package

flyth:
This change splits the gRPC API we use for communication between kubectl-gadget and gadgettracermanager and the internal gadgettracermanager gRPC API (used for housekeeping jobs). As we plan to communicate the former to third parties and also use it for other environments (communication between ig and a new client (probably called `gadgetctl`)), it makes sense to have it in a separate package that's better documented and stable.

Part of #1681


mauriciovasquezbernal (approved):
LGTM!

PR #2070, opened 2023-09-21
Enrich events with image digest

claudiamarcubina:
# [cri-o and containerd] Enrich events with image digest

This PR adds `ContainerImageDigest` column to initial events of cri-o and containerd containers. 

New containers enrichment is currently blocked because of image digest missing from OCI annotations. Will be cared of in a future iteration.

## How to use

`RUNTIME.CONTAINERIMAGEDIGEST` column is by default hidden and can be manually added with `ig list-containers -o custom-columns=runtime.containerimagename,runtime.containerimagedigest`

Part of #1310

### DIFF
--- a/docs/crds/gadgets/ebpftop.md
+++ b/docs/crds/gadgets/ebpftop.md
@@ -8,7 +8,7 @@ ebpftop shows cpu time used by ebpf programs.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,runtime.containerImageName,k8s.node,k8s.namespace,k8s.pod,k8s.container,k8s.hostnetwork,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount,totalcpu,percpu). (default -runtime,-runcount)
+ - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,runtime.containerImageName,runtime.containerImageDigest,k8s.node,k8s.namespace,k8s.pod,k8s.container,k8s.hostnetwork,progid,type,name,runtime,runcount,cumulruntime,cumulruncount,totalruntime,totalRunCount,mapmemory,mapcount,totalcpu,percpu). (default -runtime,-runcount)
 
 ### Example CR
 
--- a/docs/crds/gadgets/filetop.md
+++ b/docs/crds/gadgets/filetop.md
@@ -8,7 +8,7 @@ filetop shows reads and writes by file, with container details.
 The following parameters are supported:
  - interval: Output interval, in seconds. (default 1)
  - max_rows: Maximum rows to print. (default 20)
- - sort_by: The field to sort the results by (runtime.runtimeName,runtime.containerId,runtime.containerName,runtime.containerImageName,k8s.node,k8s.namespace,k8s.pod,k8s.container,k8s.hostnetwork,mntns,pid,tid,comm,reads,writes,rbytes,wbytes,T,file). (default -reads,-writes,-rbytes,-wbytes)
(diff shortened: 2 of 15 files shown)


blanquicet (approved):
LGTM!

I tested it with `ig` and saw that the CRI-O and containerd initial containers are enriched with the digest. 

On the other hand, I ran some tests for not-covered cases and noticed the following:

- The CRI-O initial containers are not being enriched because by the time we call `GetContainer` from `containerRuntimeEnricher`, CRI-O doesn't know yet about that container. Hence, it returns nil (Note that we are not printing this log. I'll open a PR to improve the logging to ease the debugging). Is it possible to get this information from the container-hook? cc @alban 
- The containerd initial containers are not being enriched because `GetContainer` still need to be updated (probably you're leaving that for another PR): [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/f4bf32a01f9176a022ab22d07838b134f9e70ed5) my experiments. BTW, as I suggested before, conversion from `containerd.Container` to `runtimeclient.ContainerData` should be done in one single place for `GetContainer`, `GetContainers` and `GetContainerDetails`.

mqasimsarfraz:
> Is it possible to get this information from the container-hook?

I don't think we can get this information from container-hook but it is definitely available via CRI [image service](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto#L146). We have the image name from the annotations already so we can go from `image name` to `image digest` via image service even if CRI-O runtime doesn't know the container. WDYT?

blanquicet:
> > Is it possible to get this information from the container-hook?
> 
> I don't think we can get this information from container-hook but it is definitely available via CRI [image service](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto#L146). We have the image name from the annotations already so we can go from `image name` to `image digest` via [image status](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto#L152) even if CRI-O runtime doesn't know the container. WDYT?
>
Does it work even if the runtime haven't downloaded the OCI image yet?

> Edit: Also, we should upstream image digest annotation to cri-o as well.
>
Yeah, we should. Same for containerd to keep them aligned.

@claudiamarcubina please squash last commit to the commit where the typo was added. Once done, please go ahead and merge this PR as it is and let's manage the enrichment of initial containers in another PR.

alban:
> Is it possible to get this information from the container-hook? cc @alban

Unfortunately, no, because container-hook works with runc/crun and those programs don't know about images or digest.

> I don't think we can get this information from container-hook but it is definitely available via CRI [image service](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto#L146). We have the image name from the annotations already so we can go from image name to image digest via [image status](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto#L152) even if CRI-O runtime doesn't know the container. WDYT?

I wonder how reliable that is. What if a second pod with the same image tag is started, but the OCI registry has been updated? We would have 2 pods with the same image tag but with 2 different image digest.

> Does it work even if the runtime haven't downloaded the OCI image yet?

* Container *manager*: containerd / CRI-O
* Container *runtime*: runc / crun

Container-hook works detects the container runtime. When the container runtime is started, the container manager necessarily finished to download the OCI image, so that should not be a problem.

> Edit: Also, we should upstream image digest annotation to cri-o as well.

Yes. Ditto for Docker/moby.
I started to look around with https://github.com/alban/moby/commit/f3d970f9412aa785967bf770173f725ba56f2163 but I didn't try it.

mqasimsarfraz:
> I wonder how reliable that is. What if a second pod with the same image tag is started, but the OCI registry has been updated? We would have 2 pods with the same image tag but with 2 different image digest.

That is a good point. I know from experience that `docker` untags the older image and uses image id in `IMAGE` column to avoid any confusion. From [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/docker/docker.go#L260), I assume only show image id for untagged images so it should be same. But not sure about other container managers.

PR #2071, opened 2023-09-21
ig: Make container-hook optional

mauriciovasquezbernal:
72e7751f9c11 ("pkg: Allow to get events for all processes.") introduced
    support for showing events on the host, however, ig still requires
    runc to be installed to work. This requirement doesn't make sense anymore
    as one could want to get host events without having any container.
    
    This commit makes the container-hook initialization optional for the ig
    case, instead of error it prints out a warning message. 

## How to test

0. Get a system without runc installed (I used vagrant with "ubuntu/mantic64")

```bash
$ runc
Command 'runc' not found, but can be installed with:
apt install runc
Please ask your administrator.
```

### Before this commit 

```bash
$ sudo ./ig trace open --host
Error: initializing operators: initializing operator "LocalManager": initializing manager: starting container fanotify: no container runtime can be monitored with fanotify. The following paths were tested: /bin/runc,/usr/bin/runc,/usr/sbin/runc,/usr/local/bin/runc,/usr/local/sbin/runc,/usr/lib/cri-o-runc/sbin/runc,/run/torcx/unpack/docker/bin/runc,/usr/bin/crun,/usr/bin/conmon. You can use the RUNTIME_PATH env variable to specify a custom path. If you are successful doing so, please open a PR to add your custom path to runtimePaths
```

### After this commit 

```bash
$ sudo ./ig trace open --host
WARN[0001] failed to start container-hook: container enrichment and filtering won't work 
RUNTIME.CONTAINERNAME                                      PID        COMM             FD    ERR PATH
                                                           1          systemd          8     0   /proc/496/cgroup 
```

### DIFF
--- a/pkg/operators/localmanager/localmanager.go
+++ b/pkg/operators/localmanager/localmanager.go
@@ -210,14 +210,17 @@ partsLoop:
 
 	igManager, err := igmanager.NewManager(l.rc)
 	if err != nil {
-		return commonutils.WrapInErrManagerInit(err)
+		log.Warnf("Failed to create container-collection")
+		log.Debugf("Failed to create container-collection: %s", err)
 	}
 	l.igManager = igManager
 	return nil
 }
 
 func (l *LocalManager) Close() error {
-	l.igManager.Close()
+	if l.igManager != nil {
+		l.igManager.Close()
+	}
 	return nil
 }
 
@@ -233,6 +236,10 @@ func (l *LocalManager) Instantiate(gadgetContext operators.GadgetContext, gadget
 		params:             params,
 		gadgetInstance:     gadgetInstance,
 		gadgetCtx:          gadgetContext,
+	}
+
+	if l.igManager == nil {
+		traceInstance.enrichEvents = false
 	}
 
 	return traceInstance, nil
@@ -258,9 +265,7 @@ func (l *localManagerTrace) Name() string {
 
 func (l *localManagerTrace) PreGadgetRun() error {
 	log := l.gadgetCtx.Logger()
-
 	id := uuid.New()
-	l.subscriptionKey = id.String()
 	host := l.params.Get(Host).AsBool()
 
 	// TODO: Improve filtering, see further details in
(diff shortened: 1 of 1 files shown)


pkg/ig-manager/ig-manager.go
```diff
@@ -98,7 +98,7 @@ func NewManager(runtimes []*containerutilsTypes.RuntimeConfig) (*IGManager, erro
 		containercollection.WithCgroupEnrichment(),
 		containercollection.WithLinuxNamespaceEnrichment(),
 		containercollection.WithMultipleContainerRuntimesEnrichment(runtimes),
-		containercollection.WithContainerFanotifyEbpf(),
+		containercollection.WithContainerFanotifyEbpf(false),
```
blanquicet, line 101: Shouldn't this be false only if users pass `--host`? If such a flag is not passed, it means users want to debug only containers, so it doesn't make sense to go ahead.

mauriciovasquezbernal: Some thoughts:
- It doesn't seem to easy to me getting the value of such parameter there. I see it's called from LocalManager.Init(), and there the operator params are available, but `--host` is not an operator but a gadget param IIUC.

https://github.com/inspektor-gadget/inspektor-gadget/blob/5a6dd27ae890d493929423eb452914c9bb76f867/pkg/operators/localmanager/localmanager.go#L163

- In the current implementation we could know the value of `--host`, but once https://github.com/inspektor-gadget/inspektor-gadget/issues/1681 is implemented we won't know that value at the time the daemon is started. It's even possible that the user tries to run different gadgets with and without --host
- I'm not totally convinced that the default should be `--host=false`, so I rather prefer not to tie those two things there.

alban (approved):
LGTM.

I wonder if some users would want a way to fail when they pass flags to filter by containers but the filters don't work. But for now I think the warning is good enough.

pkg/ig-manager/ig-manager.go
blanquicet: With the client/server approach, I agree it should go ahead. However, we need to provide some feedback when the user starts the server, the container-hook fails to get initialized, and they start a gadget without `--host`. Otherwise, container enrichment and filtering won't work, and they will have no idea what's happening because the error `Error: initializing operators: initializing operator "LocalManager": initializing manager: starting` will be only on the server side, when it was launched.

Maybe a new `gadgetctl status` could provide the server's status, including which hooks were installed and their statuses? It could also provide other info like the one for BTF: it was downloaded, or it's using the one from the system, etc. That could be useful to request when folks report an issue.

mauriciovasquezbernal:
I decided to take another approach with this PR, now the logic is one level above and we're able to provide warnings when the gadget is run:

```
$ sudo ./ig trace exec
WARN[0001] Failed to create container-collection        
RUNTIME.CONTAINERNAME                             PID        PPID       COMM             RET ARGS                                                             
Error: running gadget: gadget prerun: pre gadget run on operator "LocalManagerTrace": container-collection isn't available

$ sudo ./ig trace exec --host
WARN[0001] Failed to create container-collection        
RUNTIME.CONTAINERNAME                             PID        PPID       COMM             RET ARGS                                                             
WARN[0001] container-collection isn't available no container filtering will be done 

$ sudo ./ig trace dns
WARN[0001] Failed to create container-collection        
RUNTIME.CONTAINERNAME             PID        TID        COMM             QR TYPE      QTYPE      NAME                              RCODE              NUMANSW…
Error: running gadget: gadget prerun: pre gadget run on operator "LocalManagerTrace": container-collection isn't available

$ sudo ./ig trace dns --host
WARN[0001] Failed to create container-collection        
RUNTIME.CONTAINERNAME             PID        TID        COMM             QR TYPE      QTYPE      NAME                              RCODE              NUMANSW…
WARN[0002] container-collection isn't available, no containers will be traced 
```

What do you think?

pkg/operators/localmanager/localmanager.go
```diff
@@ -273,21 +278,38 @@ func (l *localManagerTrace) PreGadgetRun() error {
...
 
-		log.Debugf("set mountnsmap for gadget")
-		setter.SetMountNsMap(mountnsmap)
+			log.Debugf("set mountnsmap for gadget")
+			setter.SetMountNsMap(mountnsmap)
 
-		l.mountnsmap = mountnsmap
+			l.mountnsmap = mountnsmap
+		} else if l.manager.igManager == nil {
+			log.Warn("container-collection isn't available no container filtering will be done")
```
blanquicet, line 298: Mention enrichment won't work either:

```suggestion
			log.Warn("container-collection isn't available: container enrichment and filtering won't work")
```

Here or in the `Instantiate` when you set `enrichEvents`, it's the same.

blanquicet (approved):
I prefer this approach. It covers all the cases and provides proper info to users.

alban (approved):

PR #2075, opened 2023-09-22
examples/dns: Add missing call to host.Init()

mauriciovasquezbernal:
It panics without this call.

panic: host.Init() must be called before calling isHostNamespace()

Fixes: a65dbbc6bcfd ("host package: workarounds for kubectl-debug and Windows")

@alban do you think it's ok to have this call there or should we move it to the tracer itself so users don't need to add this?


alban (approved):
LGTM. I think it's good like this.

PR #2076, opened 2023-09-22
Add eBPF filter to path prefix

eiffel-fl:
Hi.

This PR adds eBPF filter to path prefix:

```
RUNTIME.CONTAINERNAME          PID              COMM             FD  ERR PATH                            
                               412595           git              3   0   /home/francis/.config/git/config
                               412595           git              3   0   /home/francis/.gitconfig        
                               412595           git              3   0   /home/francis/Codes/kinvolk/ins…
                               412595           git              3   0   /home/francis/.config/git/config
                               412595           git              3   0   /home/francis/.gitconfig        
                               412596           git              3   0   /home/francis/.config/git/config
                               412596           git              3   0   /home/francis/.gitconfig        
                               412596           git              3   0   /home/francis/Codes/kinvolk/ins…
                               412596           git              3   0   /home/francis/.config/git/config
```

Best regards.

### DIFF
--- a/docs/gadgets/trace/open.md
+++ b/docs/gadgets/trace/open.md
@@ -73,3 +73,14 @@ CONTAINER                     PID        COMM             FD  ERR PATH          
 test-trace-open-fullpath      1330356    cat              3   0   /etc/passwd                     /etc/passwd
 test-trace-open-fullpath      1330401    cat              3   0   ../etc/mtab                     /proc/22/mounts
 ```
+
+You can also filter by prefix paths using `--prefixes`.
+It will then only report events where opened files matched one of the given prefixes.
+
+```bash
+RUNTIME.CONTAINERNAME                            PID        COMM             FD    ERR PATH
+test-trace-open-fullpath                         64069      touch            3     0   /tmp/foo/quux.txt
+test-trace-open-fullpath                         64034      bash             3     0   /tmp/bar.txt
+```
+
+Note that this filtering occurs in eBPF.


amitschendel:
Looks very cool!
Can you add support for multiple prefixes instead of just one?

pkg/gadgets/trace/open/tracer/bpf/opensnoop.bpf.c
```diff
@@ -33,14 +34,41 @@ struct {
 	__uint(value_size, sizeof(u32));
 } events SEC(".maps");
 
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
```
alban, line 38: This could be optimized with a BPF_MAP_TYPE_LPM_TRIE.

alban:
For the record, I had the following suggestion, but in the end, my suggestion is not working.

We already have DetectBpfKtimeGetBootNs & FixBpfKtimeGetBootNs to detect if we have bpf_ktime_get_boot_ns. And if not, it is replaced by a `Mov R0, 0` instruction.

In the case of bpf_for_each_map_elem, we can do something similar but with `Mov R0, -95`  (ENOTSUP 95 Operation not supported). Then, in the ebpf code, we can just check `if (ret == -95)`.

But we couldn't get it to work because the ebpf bytecode then contains a function that is never called.

amitschendel:
@eiffel-fl @alban Thank you both, highly appreciated.
This is very useful, I hope that maybe I will have time to continue your work soon!

amitschendel:
How about doing something like this? Instead of using `bpf_for_each_map_elem`
```c
static int matches_prefix(struct bpf_map *prefixes_map, char *filename)
{
    for (size_t i = 0; i < prefixes_nr; i++) {
        char *map_prefix = bpf_map_lookup_elem(prefixes_map, &i);
        if (!map_prefix) {
            // Handle the case where map_lookup_elem fails.
            return 0;
        }

        /* Filename does not match the prefix, we need to test others. */
        for (size_t j = 0; map_prefix[j] != '\0' && j < NAME_MAX - 1; j++) {
            if (filename[j] == '\0' || map_prefix[j] != filename[j])
                return 0;
        }

        /* Filename matches the prefix, we can skip others entries. */
        return 1;
    }

    return 0; // No match found.
}
```

mauriciovasquezbernal:
> But we couldn't get it to work because the ebpf bytecode then contains a function that is never called.

I think we could remove that function. But all of this seems too hacky to me. 

I see a suggestion about using a LPM, didn't it work? Another idea could be to unroll the loop, it'll increase complexity but I think this program is far away from the kernel limit.

amitschendel:
Awesome!
Couple of questions:
1. Is it supported from the api? can I create a tracer and pass a config or just from the cli? (I need it in the api).
2. Can the prefix be the name of the file? for example if I will access bla.txt and the prefix is bla.txt will it work? 

Thanks

burak-ok (approved):
LGTM (Maybe wait for another approval of a more experienced eBPF dev)
Also tested it locally and it works well :rocket:

PR #2078, opened 2023-09-25
ig: add daemon mode

### LINKED ISSUE #1681: [RFE] Adopt client/server setup also for ig

flyth:
This adds support to run `ig` in daemon mode as a service. A new binary called `gadgetctl` has been added to serve as a client for `ig` running in said daemon mode. By default, ig listens on a unix socket when started with `./ig daemon` and `gadgetctl` uses that same socket to connect to by default.

The `grpc-runtime` has been changed to also allow direct gRPC connections - still: most of the code is shared.

#### First terminal

```bash
./ig daemon
```

#### Second terminal
```bash
./gadgetctl trace open
```

### Daemon as root, gadgetctl as other user

Use the gid of a group that you want to have access to the service.

#### First terminal (as root)

```bash
./ig daemon --daemon-group 1000
```

#### Second terminal (as member of group 1000)
```bash
./gadgetctl trace open
```

## Todos:

* [x] documentation
* [x] CI

---

Resolves #1681
Requires #2067

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -160,13 +160,14 @@ jobs:
       - name: Check if added dependencies do not contain CVE.
         uses: actions/dependency-review-action@v3
 
-  build-kubectl-gadget:
-    name: kubectl-gadget
+  build-clients:
+    name: clients
     # level: 0
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
+        client: [kubectl-gadget, gadgetctl]
         os: [linux, darwin, windows]
         arch: [amd64, arm64]
         exclude:
@@ -187,7 +188,7 @@ jobs:
         registry: ${{ env.REGISTRY }}
         container-image: ${{ env.CONTAINER_REPO }}
         bcc: false
-    - name: Build kubectl-gadget-${{ matrix.os }}-${{ matrix.arch }}
+    - name: Build ${{ matrix.client }}-${{ matrix.os }}-${{ matrix.arch }}
       run: |
         git checkout
 
@@ -199,26 +200,26 @@ jobs:
           exit 1
         fi
 
-        kubectl_gadget=kubectl-gadget-${{ matrix.os }}-${{ matrix.arch }}
+        client=${{ matrix.client }}-${{ matrix.os }}-${{ matrix.arch }}
 
         CONTAINER_REPO=${{ steps.set-repo-determine-image-tag.outputs.container-repo }} \
         IMAGE_TAG=${{ steps.set-repo-determine-image-tag.outputs.image-tag }} \
-        make $kubectl_gadget
+        make $client
 
         # Prepare assets for release and actions artifacts
-        platform=$(echo ${kubectl_gadget} | cut -d- -f3-4)
+        platform='${{ matrix.os }}-${{ matrix.arch }}'
         mkdir $platform
-        cp $kubectl_gadget $platform/kubectl-gadget
+        cp $client $platform/${{ matrix.client }}
         cp LICENSE $platform/
         tar --sort=name --owner=root:0 --group=root:0 \
-          -czf ${kubectl_gadget}.tar.gz -C $platform \
-          kubectl-gadget LICENSE
+          -czf ${client}.tar.gz -C $platform \
+          ${{ matrix.client }} LICENSE
         rm -rf $platform
-    - name: Add kubectl-gadget-${{ matrix.os }}-${{ matrix.arch }}.tar.gz as artifact.
+    - name: Add ${{ matrix.client }}-${{ matrix.os }}-${{ matrix.arch }}.tar.gz as artifact.
       uses: actions/upload-artifact@master
       with:
-        name: kubectl-gadget-${{ matrix.os }}-${{ matrix.arch }}-tar-gz
-        path: /home/runner/work/inspektor-gadget/inspektor-gadget/kubectl-gadget-${{ matrix.os }}-${{ matrix.arch }}.tar.gz
+        name: ${{ matrix.client }}-${{ matrix.os }}-${{ matrix.arch }}-tar-gz
+        path: /home/runner/work/inspektor-gadget/inspektor-gadget/${{ matrix.client }}-${{ matrix.os }}-${{ matrix.arch }}.tar.gz
 
   btfgen:
     name: btfgen
@@ -1090,7 +1091,7 @@ jobs:
     needs:
       - check-secrets
       - test-unit
-      - build-kubectl-gadget
+      - build-clients
       - build-ig
       - build-gadget-container-images
       - publish-gadget-images-manifest
@@ -1189,7 +1190,7 @@ jobs:
     needs:
       - check-secrets
       - test-unit
-      - build-kubectl-gadget
+      - build-clients
       - build-ig
       - build-gadget-container-images
       - publish-gadget-images-manifest
@@ -1237,7 +1238,7 @@ jobs:
     # level: 3
     needs:
       - test-unit
-      - build-kubectl-gadget
+      - build-clients
       - build-ig
       - build-gadget-container-images
       - build-helper-images
@@ -1377,13 +1378,13 @@ jobs:
     - name: Rename all artifacts to *-${{ github.ref_name }}.tar.gz
       shell: bash
       run: |
-        for i in kubectl-gadget-*-*-tar-gz/kubectl-gadget-*-*.tar.gz ig-*-*-tar-gz/ig-*-*.tar.gz; do
+        for i in kubectl-gadget-*-*-tar-gz/kubectl-gadget-*-*.tar.gz ig-*-*-tar-gz/ig-*-*.tar.gz gadgetctl-*-*-tar-gz/gadgetctl-*-*.tar.gz; do
           mv $i $(dirname $i)/$(basename $i .tar.gz)-${{ github.ref_name }}.tar.gz
         done
     - name: Compute checksums for all artifacts
       shell: bash
       run: |
-        for i in kubectl-gadget-*-*-tar-gz/kubectl-gadget-*-*.tar.gz ig-*-*-tar-gz/ig-*-*.tar.gz inspektor-gadget-${{ github.ref_name }}.yaml ig_packages/*; do
+        for i in kubectl-gadget-*-*-tar-gz/kubectl-gadget-*-*.tar.gz ig-*-*-tar-gz/ig-*-*.tar.gz gadgetctl-*-*-tar-gz/gadgetctl-*-*.tar.gz inspektor-gadget-${{ github.ref_name }}.yaml ig_packages/*; do
           hash=$(sha256sum $i | cut -d' ' -f1)
           echo "${hash}  $(basename $i)" >> SHA256SUMS
         done
@@ -1412,6 +1413,12 @@ jobs:
       uses: csexton/release-asset-action@v2
       with:
         pattern: "ig-*-*-tar-gz/ig-*-*.tar.gz"
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        release-url: ${{ steps.create_release.outputs.upload_url }}
+    - name: Upload gadgetctl *.tar.gz binary
+      uses: csexton/release-asset-action@v2
+      with:
+        pattern: "gadgetctl-*-*-tar-gz/gadgetctl-*-*.tar.gz"
         github-token: ${{ secrets.GITHUB_TOKEN }}
         release-url: ${{ steps.create_release.outputs.upload_url }}
     - name: Upload YAML
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ include crd.mk
 include tests.mk
 include minikube.mk
 
-LDFLAGS := "-X main.version=$(VERSION) \
+LDFLAGS := "-X github.com/inspektor-gadget/inspektor-gadget/cmd/common.version=$(VERSION) \
 -X main.gadgetimage=$(CONTAINER_REPO):$(IMAGE_TAG) \
 -extldflags '-static'"
 
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,36 @@ install/kubectl-gadget: kubectl-gadget-$(GOHOSTOS)-$(GOHOSTARCH)
 	mkdir -p ~/.local/bin/
 	cp kubectl-gadget-$(GOHOSTOS)-$(GOHOSTARCH) ~/.local/bin/kubectl-gadget
 
+GADGETCTL_TARGETS = \
+	gadgetctl-linux-amd64 \
+	gadgetctl-linux-arm64 \
+	gadgetctl-darwin-amd64 \
+	gadgetctl-darwin-arm64 \
+	gadgetctl-windows-amd64
+
+.PHONY: list-gadgetctl-targets
+list-gadgetctl-targets:
+	@echo $(GADGETCTL_TARGETS)
+
+.PHONY: gadgetctl-all
+gadgetctl-all: $(GADGETCTL_TARGETS) gadgetctl
+
+gadgetctl: gadgetctl-$(GOHOSTOS)-$(GOHOSTARCH)
+	cp gadgetctl-$(GOHOSTOS)-$(GOHOSTARCH) gadgetctl
+
+gadgetctl-%: phony_explicit
+	export GO111MODULE=on CGO_ENABLED=0 && \
+	export GOOS=$(shell echo $* |cut -f1 -d-) GOARCH=$(shell echo $* |cut -f2 -d-) && \
+	go build -ldflags $(LDFLAGS) \
+		-tags withoutebpf \
+		-o gadgetctl-$${GOOS}-$${GOARCH} \
+		github.com/inspektor-gadget/inspektor-gadget/cmd/gadgetctl
+
+.PHONY: install/gadgetctl
+install/gadgetctl: gadgetctl-$(GOHOSTOS)-$(GOHOSTARCH)
+	mkdir -p ~/.local/bin/
+	cp gadgetctl-$(GOHOSTOS)-$(GOHOSTARCH) ~/.local/bin/gadgetctl
+
 GADGET_CONTAINERS = \
 	gadget-default-container \
 	gadget-bcc-container
--- a/cmd/common/sync.go
+++ b/cmd/common/sync.go
@@ -1,0 +1,31 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"github.com/spf13/cobra"
+
+	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
+)
+
+func NewSyncCommand(runtime *grpcruntime.Runtime) *cobra.Command {
+	return &cobra.Command{
+		Use:   "sync",
+		Short: "Synchronize gadget information with server",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runtime.UpdateDeployInfo()
+		},
+	}
+}
--- a/cmd/gadgetctl/main.go
+++ b/cmd/gadgetctl/main.go
@@ -1,0 +1,85 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/all-gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/environment"
+	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
+)
+
+var infoSkipCommands = []string{"version"}
+
+func main() {
+	if experimental.Enabled() {
+		log.Info("Experimental features enabled")
+	}
+
+	rootCmd := &cobra.Command{
+		Use:   filepath.Base(os.Args[0]),
+		Short: "Collection of gadgets for containers",
+	}
+	common.AddVerboseFlag(rootCmd)
+
+	skipInfo := false
+	for _, arg := range os.Args[1:] {
+		for _, skipCmd := range infoSkipCommands {
+			if strings.ToLower(arg) == skipCmd {
+				skipInfo = true
+			}
+		}
+	}
+
+	rootCmd.AddCommand(common.NewVersionCmd())
+
+	runtime := grpcruntime.New()
+
+	runtimeGlobalParams := runtime.GlobalParamDescs().ToParams()
+	common.AddFlags(rootCmd, runtimeGlobalParams, nil, runtime)
+	err := runtime.Init(runtimeGlobalParams)
+	if err != nil {
+		log.Fatalf("initializing runtime: %v", err)
+	}
+
+	if !skipInfo {
+		// evaluate flags early for runtimeGlobalFlags; this will make
+		// sure that --remote-address has already been parsed when calling
+		// InitDeployInfo(), so it can target the specified address
+		rootCmd.ParseFlags(os.Args[1:])
+		runtime.InitDeployInfo()
+	}
+
+	hiddenColumnTags := []string{"kubernetes"}
+	common.AddCommandsFromRegistry(rootCmd, runtime, hiddenColumnTags)
+
+	rootCmd.AddCommand(common.NewSyncCommand(runtime))
+
+	if err := rootCmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
+	environment.Environment = environment.Local
+}
--- a/cmd/ig/daemon.go
+++ b/cmd/ig/daemon.go
@@ -1,0 +1,102 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strconv"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	gadgetservice "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+)
+
+func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
+	daemonCmd := &cobra.Command{
+		Use:          "daemon",
+		Short:        "Run Inspektor Gadget as a daemon",
+		Args:         cobra.NoArgs,
+		SilenceUsage: true,
+	}
+
+	var socket string
+	var group string
+
+	daemonCmd.PersistentFlags().StringVarP(
+		&group,
+		"group",
+		"G",
+		"0",
+		"Group name or id that the unix socket should use (if daemon-socket is set to e.g. unix:///path/to.socket)")
+
+	daemonCmd.PersistentFlags().StringVarP(
+		&socket,
+		"host",
+		"H",
+		api.DefaultDaemonPath,
+		"The socket to listen on for new requests. Can be a unix socket"+
+			" (unix:///path/to.socket) or a tcp socket (tcp://127.0.0.1:1234)")
+
+	daemonCmd.RunE = func(cmd *cobra.Command, args []string) error {
+		if os.Geteuid() != 0 {
+			return fmt.Errorf("%s must be run as root to be able to run eBPF programs", filepath.Base(os.Args[0]))
+		}
+
+		socketURL, err := url.Parse(socket)
+		if err != nil {
+			return fmt.Errorf("invalid daemon-socket address %q: %w", socketURL, err)
+		}
+
+		gid := 0
+		if tmpGroup, err := user.LookupGroup(group); err == nil {
+			gid, err = strconv.Atoi(tmpGroup.Gid)
+			if err != nil {
+				return fmt.Errorf("unexpected non-numeric group id %q for group %q", tmpGroup.Gid, group)
+			}
+		} else if tmpGid, err := strconv.Atoi(group); err == nil {
+			gid = tmpGid
+		} else {
+			return fmt.Errorf("group %q not found", group)
+		}
+
+		var socketPath string
+		socketType := socketURL.Scheme
+		switch socketType {
+		default:
+			return fmt.Errorf("invalid type for daemon-socket %q; please use 'unix' or 'tcp'", socketType)
+		case "unix":
+			socketPath = socketURL.Path
+		case "tcp":
+			socketPath = socketURL.Host
+		}
+
+		log.Infof("starting Inspektor Gadget daemon at %q", socket)
+		service := gadgetservice.NewService(log.StandardLogger())
+		return service.Run(gadgetservice.RunConfig{
+			SocketType: socketType,
+			SocketPath: socketPath,
+			SocketGID:  gid,
+		})
+	}
+
+	return daemonCmd
+}
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -62,6 +62,7 @@ func main() {
 	hiddenColumnTags := []string{"kubernetes"}
 	common.AddCommandsFromRegistry(rootCmd, runtime, hiddenColumnTags)
 
+	rootCmd.AddCommand(newDaemonCommand(runtime))
 	if experimental.Enabled() {
 		rootCmd.AddCommand(image.NewImageCmd())
 		rootCmd.AddCommand(common.NewLoginCmd())
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -55,7 +55,7 @@ func main() {
 
 	rootCmd.AddCommand(
 		containers.NewListContainersCmd(),
-		newVersionCmd(),
+		common.NewVersionCmd(),
 	)
 
 	runtime := local.New()
--- a/docs/ig.md
+++ b/docs/ig.md
@@ -195,6 +195,110 @@ re-execute itself in a privileged systemd unit if necessary.
 This is possible because the "kubectl debug node" gives access to the systemd socket (`/run/systemd/private`) via the
 /host volume.
 
+### Using ig as a daemon
+
+`ig` can also be run as a daemon. You can then use `gadgetctl` to connect to it as an unprivileged user.
+We will be using a new group called `ig` to allow controlling `ig` in this example.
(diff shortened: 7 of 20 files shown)


cmd/ig/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+)
+
+func addDaemonCommand(rootCmd *cobra.Command, runtime runtime.Runtime) {
+	daemonCmd := &cobra.Command{
+		Use:   "daemon",
+		Short: "runs Inspektor Gadget as a daemon",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
```
alban, line 37: Could those 3 lines be deleted since you overwrite RunE on line 58?

cmd/ig/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+
+	daemonCmd.PersistentFlags().StringVarP(
+		&socket,
+		"daemon-socket",
+		"",
+		api.DefaultDaemonPath,
+		"The socket to listen on for new Inspektor Gadget requests. Can be a unix socket"+
+			" (unix:///path/to.socket) or a tcp socket (tcp://127.0.0.1:1234)")
+
+	daemonCmd.RunE = func(cmd *cobra.Command, args []string) error {
```
alban, line 58: if len(args) != 0, return an error?

alban (commented):
Patch 1 and 3 looks generally good, just small comments. I didn't review the gadgetctl patch yet.

alban:
How do I specify the unix socket to connect to?
```
$ ./gadgetctl trace mount
RUNTIME.CONTAINERNAME                                  COMM             PID        TID        CALL                                                                                                                                            
Error: running gadget: dialing gadget pod on node "": dialing node "": context deadline exceeded
```

It takes 30 seconds to print the error. The column line is printed immediately, leading the user to think there is no problem for 30 seconds.
```
$ time ./gadgetctl trace open 
RUNTIME.CONTAINERNAME                                                                           PID        COMM             FD        ERR PATH                                                                                                
Error: running gadget: dialing gadget pod on node "": dialing node "": context deadline exceeded

real	0m30.042s
user	0m0.026s
sys	0m0.023s
```

It needs a version command:
```
$ ./gadgetctl version
Error: unknown command "version" for "gadgetctl"
Run 'gadgetctl --help' for usage.
```

flyth:
> How do I specify the unix socket to connect to?

I expected it to be `--remote-address` but that didn't properly get added as a flag. Will check what I missed when splitting the code, thanks! :)

> It needs a version command

Thanks, will add it!

cmd/ig/daemon.go
flyth: Good point, probably should use `Args:  cobra.NoArgs` instead of manual testing, though (although I only saw that used once in our codebase), right?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -54,33 +70,44 @@ const (
...
+	for _, option := range options {
+		option(r)
 	}
+	return r
+}
 
+// InitInfo loads the locally stored gadget catalog. If no catalog is stored locally,
+// it will try to fetch one from one of the gadget nodes and store it locally. It will issue warnings on
+// failures.
+func (r *Runtime) InitInfo(params *params.Params) error {
```
mauriciovasquezbernal, line 99: "It will issue warnings on failures": Should the return `error` be removed then?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -110,35 +138,50 @@ func (r *Runtime) Close() error {
...
+	switch r.connectionMode {
+	case ConnectionModeDirect:
+		p.Add(params.ParamDescs{
+			{
+				Key:          ParamRemoteAddress,
+				Description:  "Comma-separated list of remote address (gRPC) to connect to",
+				DefaultValue: api.DefaultDaemonPath,
+			},
+		}...)
+		return p
```
mauriciovasquezbernal, line 151: nit: how about just returning the slice here without having to use `p.Add()`?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -184,17 +227,79 @@ nodesLoop:
...
-		return nil, fmt.Errorf("get gadget pods: Inspektor Gadget is not running on the requested node(s): %v", nodes) //nolint:all
+	if len(targets) == 0 {
+		return nil, fmt.Errorf("no valid targets")
 	}
+	conn, err := r.dialContext(ctx, targets[0])
+	if err != nil {
+		return nil, err
+	}
+	client := api.NewGadgetManagerClient(conn)
+	return client, nil
```
mauriciovasquezbernal, line 295: super minor nit: Avoid the temp variable and just `return api.NewGadgetManagerClient(conn), nil`?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -212,54 +317,57 @@ func (r *Runtime) RunGadget(gadgetCtx runtime.GadgetContext) (runtime.CombinedGa
...
+	}
+
+	// If we're in Kubernetes connection mode, we need a custom dialer
+	if r.connectionMode == ConnectionModeKubernetesProxy {
+		opts = append(opts, grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
+			return NewK8SExecConn(ctx, target, time.Second*ConnectTimeout)
+		}))
+	}
+
+	conn, err := grpc.DialContext(dialCtx, "passthrough:///"+target.name, opts...)
```
mauriciovasquezbernal, line 353: What's `passthrough:///` for here. I tried to look around in the grpc and I can't find documentation about it.

cmd/common/sync.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
+)
+
+func AddSyncCommand(rootCmd *cobra.Command, runtime *grpcruntime.Runtime) {
+	rootCmd.AddCommand(&cobra.Command{
+		Use:   "sync",
+		Short: "Synchronize gadget information with your cluster",
```
mauriciovasquezbernal, line 26: cluster doesn't apply to gadgetctl. Perhaps "Synchronize gadget information with remote" or "Synchronize gadget information with ig server"?

cmd/ig/daemon.go
mauriciovasquezbernal, line 37: use `SilenceUsage: true,`? Otherwise I get weird things like:

```bash
$ ./ig daemon 
Error: ig must be run as root to be able to run eBPF programs
Usage:
  ig daemon [flags]

Flags:
      --daemon-group-id int    Group id that the unix socket should use (if daemon-socket is set to e.g. unix:///path/to.socket)
      --daemon-socket string   The socket to listen on for new Inspektor Gadget requests. Can be a unix socket (unix:///path/to.socket) or a tcp socket (tcp://127.0.0.1:1234) (default "unix:///var/run/ig.socket")
  -h, --help                   help for daemon

Global Flags:
      --auto-mount-filesystems          Automatically mount bpffs, debugfs and tracefs if they are not already mounted
      --auto-sd-unit-restart            Automatically run in a privileged systemd unit if lacking enough capabilities
      --auto-wsl-workaround             Automatically find the host procfs when running in WSL2
      --containerd-namespace string     Containerd namespace to use (default "k8s.io")
      --containerd-socketpath string    Co......
```

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -89,11 +116,11 @@ func New(skipInfo bool) *Runtime {
 		log.Warnf("could not store gadget info: %v", err)
 	}
 
-	return r
+	return nil
 }
 
 func (r *Runtime) UpdateDeployInfo() error {
```
mauriciovasquezbernal, line 122: I know it was before, but shouldn't `UpdateDeployInfo` and `InitInfo` be moved to info.go?

cmd/ig/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/spf13/cobra"
+
+	gadgetservice "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+)
+
+func addDaemonCommand(rootCmd *cobra.Command, runtime runtime.Runtime) {
+	daemonCmd := &cobra.Command{
+		Use:   "daemon",
```
mauriciovasquezbernal, line 33: I'm wondering about the meaning of daemon here. From [wikipedia](https://en.wikipedia.org/wiki/Daemon_(computing)) it's defined as "[computer program](https://en.wikipedia.org/wiki/Computer_program) that runs as a [background process](https://en.wikipedia.org/wiki/Background_process), rather than being under the direct control of an interactive user", but in this PR is used more like a server. Should we call it server instead of daemon? /cc @alban

cmd/ig/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+
+	var socket string
+	var gid int
+
+	daemonCmd.PersistentFlags().IntVarP(
+		&gid,
+		"daemon-group-id",
+		"",
+		0,
+		"Group id that the unix socket should use (if daemon-socket is set to e.g. unix:///path/to.socket)")
```
mauriciovasquezbernal, line 48: Any thoughts between using the group id vs group name? Docker uses the name: `--group`. Perhaps the id is better because we don't have to worry about creating the group?

cmd/ig/daemon.go
mauriciovasquezbernal: There is one instance of each case:

```bash
$ git grep cobra.NoArgs
cmd/common/advise/seccomp-profile.go:           Args:         cobra.NoArgs,

$ git grep "len(args) != 0"
pkg/gadgets/run/tracer/run.go:  if len(args) != 0 {
```

 I think `cobra.NoArgs` is better.

pkg/runtime/grpc/info.go
```diff
@@ -18,41 +18,25 @@ import (
...
-
-	// Get a random gadget pod and get the info from there
-	pods, err := getGadgetPods(ctx, []string{})
-	if err != nil {
-		return nil, fmt.Errorf("get gadget pods: %w", err)
+	if params == nil {
+		// use defaults, if params are unset (e.g. automatic fetching
+		// of catalog after deployment)
+		params = r.ParamDescs().ToParams()
 	}
```
mauriciovasquezbernal, line 35: I can't understand it. nil is passed in all places where it's called, when a value other than nil will make sense?

mauriciovasquezbernal:
I also think there is something off with the timeout, `./gadgetctl -h` takes 30 seconds even if the socket file doesn't exist, shouldn't it return immediately in that case?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -110,35 +138,50 @@ func (r *Runtime) Close() error {
 }
 
 func (r *Runtime) ParamDescs() params.ParamDescs {
```
mauriciovasquezbernal, line 140: I think these should be `GlobalParamDescs`

blanquicet (commented):
IMO, it is looking good and you kept it still "simple". My comments are mostly nits, so it is fine for me to go ahead with this PR.

cmd/gadgetctl/main.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	err = runtime.InitInfo(nil)
+	if err != nil {
+		log.Fatalf("fetching info from remote: %v", err)
+	}
+	hiddenColumnTags := []string{"kubernetes"}
+	common.AddCommandsFromRegistry(rootCmd, runtime, hiddenColumnTags)
+	common.AddSyncCommand(rootCmd, runtime)
+
+	if err := rootCmd.Execute(); err != nil {
```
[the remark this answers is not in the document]
flyth: Cobra seems to print the error in `cmd.ExecuteC()` for most cases. This is also how we handle it in the other binaries (`ig`, `kubectl-gadget`), I _think_ it's fine.

pkg/runtime/grpc/grpc-runtime.go
flyth: When we add the persistency/long running gadgets, there will be a couple of shared parameters that p is initialized with. I wanted to keep the changes minimal, so I didn't clean this up just to move it again with the next PR. Maybe I should add notes/TODOs in such cases.

pkg/runtime/grpc/grpc-runtime.go
flyth: It's a scheme handler defined [here](https://pkg.go.dev/google.golang.org/grpc/internal/resolver/passthrough), the resolution of targets is partially documented [here](https://github.com/grpc/grpc/blob/master/doc/naming.md). Before, we only used a custom dialer (to tunnel through the Kubernetes API server) and left that target string empty, but in direct mode this is actually used to hand over the target. I'll need to re-check why I couldn't use `unix://...` or `tcp://` directly, though (too long ago).

cmd/ig/daemon.go
flyth: Hmm, I think in its current state "server" would indeed be a better choice, as it just serves requests and doesn't do any background running of things. With the long running gadgets that will however change again and things will indeed run in the background and you just control using what exactly that is using the socket (in addition to still be able to use it as a "server").

cmd/ig/daemon.go
flyth: We could rename it to `daemon-group` (or just `group` if you prefer that) and allow both; if it's numeric, use it as is, if it's a non-numeric, use `user.LookupGroupId()` to fetch the actual GID. WDYT?

pkg/runtime/grpc/grpc-runtime.go
flyth: I had that in GlobalParamDescs first, but if you run gadgetctl as a (web-)server (to be added later on™), it might make sense to choose your target after the server has been initialized.

cmd/ig/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	gadgetservice "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+)
+
+func addDaemonCommand(rootCmd *cobra.Command, runtime runtime.Runtime) {
```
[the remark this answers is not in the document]
flyth: You mean `ig` should fork itself and stay active in the background with the foreground process exiting? I _think_ we just need a proper service definition that starts `ig daemon` - that way logging is also routed through systemd (or whatever).

cmd/ig/main.go
```diff
@@ -59,8 +59,10 @@ func main() {
 	)
 
 	runtime := local.New()
+	// columnFilters for ig
```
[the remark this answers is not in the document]
flyth: Yes 😄

Makefile
```diff
@@ -149,6 +149,36 @@ install/kubectl-gadget: kubectl-gadget-$(GOHOSTOS)-$(GOHOSTARCH)
...
+	gadgetctl-darwin-amd64 \
+	gadgetctl-darwin-arm64 \
+	gadgetctl-windows-amd64
+
+.PHONY: list-gadgetctl-targets
+list-gadgetctl-targets:
+	@echo $(GADGETCTL_TARGETS)
+
+.PHONY: gadgetctl-all
+gadgetctl-all: $(GADGETCTL_TARGETS) gadgetctl
```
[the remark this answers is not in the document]
flyth: Seems to copy the version for your host OS to `./kubectl-gadget` / `./gadgetctl` (without OS and architecture in the name). Developer convenience, I guess, but I just copy/pasted it :)

Makefile
```diff
@@ -149,6 +149,36 @@ install/kubectl-gadget: kubectl-gadget-$(GOHOSTOS)-$(GOHOSTARCH)
 	mkdir -p ~/.local/bin/
 	cp kubectl-gadget-$(GOHOSTOS)-$(GOHOSTARCH) ~/.local/bin/kubectl-gadget
 
+GADGETCTL_TARGETS = \
```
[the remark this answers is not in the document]
flyth: Probably; afaik, we're not 100% sure if `gadgetctl` will stay or whether we combine it with `kubectl-gadget`, so I wanted to keep the changes easily reversible and not overcomplicate the code. For the CI workflow it was easier to add than it would be here (at least with my bash script knowledge), so I don't know if it's worth it right now.

cmd/gadgetctl/main.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func main() {
+	if experimental.Enabled() {
+		log.Info("Experimental features enabled")
+	}
+
+	rootCmd := &cobra.Command{
+		Use:   filepath.Base(os.Args[0]),
+		Short: "Collection of gadgets for containers",
```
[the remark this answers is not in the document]
flyth: Since this is presented prominently, /cc @blixtra - Any suggestions? This is the headline that's shown when you run `ig` / `gadgetctl`.

```
$ ./gadgetctl

Usage:
  gadgetctl [command]
...
```

pkg/gadget-service/service.go
```diff
@@ -253,7 +257,32 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
 	return nil
 }
 
-func (s *Service) Run(network, address string, serverOptions ...grpc.ServerOption) error {
+func (s *Service) newListener(network, address string, gid int) (net.Listener, error) {
+	if network == "unix" && gid != 0 {
+		if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
+			return nil, fmt.Errorf("remove existing unix socket at %q: %w", address, err)
+		}
```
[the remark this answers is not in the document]
flyth: No, you're right, should be done in any case.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -184,17 +227,79 @@ nodesLoop:
...
+		inTargets := params.Get(ParamRemoteAddress).AsStringSlice()
+		targets := make([]target, 0)
+		for _, t := range inTargets {
+			purl, err := url.Parse(t)
+			if err != nil {
+				return nil, fmt.Errorf("invalid remote address %q: %w", t, err)
+			}
+			targets = append(targets, target{
+				name: t,
+				node: purl.Hostname(),
```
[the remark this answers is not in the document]
flyth: Yeah, that's true for local unix socket connections (`unix:///path/to/socket` is missing a hostname); if you'd use tcp, this would be the IP or hostname. Should we use a hardcoded "localhost" when hostname is empty?

cmd/gadgetctl/main.go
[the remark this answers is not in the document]
mauriciovasquezbernal: OK, I was asking because I remember that I executed a command and I only got the usage without any error message. But I can't remember what it was, if we're doing the same in other binaries then it's fine.

pkg/runtime/grpc/grpc-runtime.go
mauriciovasquezbernal: Fair enough.

pkg/runtime/grpc/grpc-runtime.go
mauriciovasquezbernal: If this is still required, could you please add a comment in the code with the your github comment above?

cmd/ig/daemon.go
mauriciovasquezbernal: I think it's a good idea to support both, numeric and string. Just one concern, is it possible to name a group with a number?

flyth: I already added support for both; a group name cannot be all-numeric according to [this post](https://unix.stackexchange.com/a/754021) (linking to [this](https://github.com/shadow-maint/shadow/blob/master/lib/chkname.c#L28)).

alban: > I think it's a good idea to support both, numeric and string.

Docker does that with a single flag. Here the group can be the group name or the number:
```
  -u, --user string                    Username or UID (format: <name|uid>[:<group|gid>])
```

> Just one concern, is it possible to name a group with a number?

It seems it depends on the distro, but that seems to be a bad idea:
https://unix.stackexchange.com/questions/287077/why-cant-linux-usernames-begin-with-numbers

cmd/ig/daemon.go
mauriciovasquezbernal: What will be the daemon behaviour in the future? Will it fork and run in background? (Somehow related to Jose's comment in https://github.com/inspektor-gadget/inspektor-gadget/pull/2078#discussion_r1337571289)

Ideally the solution will be to have an `gadgetd` / `igd` and `gadgetctl` binaries, but we also want it to be able to work alone. I think the current solution is good enough, just be sure to document what daemon is about.

cmd/ig/daemon.go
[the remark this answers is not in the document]
alban: I agree, running ig as a server should not "daemonize" (fork/detach from terminal) by default, so it is easier to integrate with systemd.

pkg/gadget-service/service.go
```diff
@@ -253,7 +260,71 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
...
+		if "unix://"+address == api.DefaultDaemonPath {
+			fixDirPermissions := false
+			dir := filepath.Dir(address)
+			fi, err := os.Stat(dir)
+			if err != nil {
+				// Directory doesn't exist, create it
+				if err := os.MkdirAll(dir, 0o710); err != nil && !errors.Is(err, os.ErrExist) {
+					return nil, fmt.Errorf("creating directory %q: %w", dir, err)
+				}
+				fixDirPermissions = true
```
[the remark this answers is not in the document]
flyth: Are you sure? I specifically checked the docs (but didn't try it out) [here](https://linux.die.net/man/2/mkdir), and it's saying:

> It is modified by the process's umask in the usual way: in the absence of a default ACL, the mode of the created directory is (mode & ~umask & 0777).

.github/workflows/inspektor-gadget.yml
```diff
@@ -199,26 +200,26 @@ jobs:
...
+        client=${{ matrix.client }}-${{ matrix.os }}-${{ matrix.arch }}
 
         CONTAINER_REPO=${{ steps.set-repo-determine-image-tag.outputs.container-repo }} \
         IMAGE_TAG=${{ steps.set-repo-determine-image-tag.outputs.image-tag }} \
-        make $kubectl_gadget
+        make $client
 
         # Prepare assets for release and actions artifacts
-        platform=$(echo ${kubectl_gadget} | cut -d- -f3-4)
+        platform=$(echo ${client} | cut -d- -f3-4)
```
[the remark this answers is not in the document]
flyth: Ah, nice catch! - I just looked at the artifacts and they looked fine.

cmd/ig/daemon.go
```diff
@@ -0,0 +1,104 @@
...
+		Args:         cobra.NoArgs,
+		SilenceUsage: true,
+	}
+
+	var socket string
+	var group string
+
+	daemonCmd.PersistentFlags().StringVarP(
+		&group,
+		"daemon-group",
```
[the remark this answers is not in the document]
flyth: For Docker it seems to be `--group` (@mauriciovasquezbernal mentioned it [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2078#discussion_r1337639733), docs [here](https://docs.docker.com/engine/reference/commandline/dockerd/)), so we should probably go with that one as well. Docker uses `-H` / `--host` for the socket path / address - should we use that, too?

/cc @mauriciovasquezbernal @alban

cmd/ig/daemon.go
flyth: xref discussion [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2078#discussion_r1337571289)

mauriciovasquezbernal: Probably it can still be called server and the matter it's a daemon or not is controlled by systemd?, i.e. we don't provide an option to run as a daemon directly. WDYT?

/cc @alban @blanquicet pinging you folks to consolidate the discussion on this thread.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -184,17 +191,78 @@ nodesLoop:
...
+	switch r.connectionMode {
+	case ConnectionModeKubernetesProxy:
+		// Get nodes to run on
+		nodes := params.Get(ParamNode).AsStringSlice()
+		pods, err := getGadgetPods(ctx, nodes)
+		if err != nil {
+			return nil, fmt.Errorf("get gadget pods: %w", err)
+		}
+		if len(pods) == 0 {
+			return nil, fmt.Errorf("get gadget pods: Inspektor Gadget is not running on the requested node(s): %v", nodes) //nolint:all
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I removed that in https://github.com/inspektor-gadget/inspektor-gadget/pull/2015 and it didn't complain.

cmd/ig/daemon.go
alban: I agree: let's call the CLI command "ig server" and let's run this in the foreground.

If ever needed later, we could add an option "ig server -d" to run in the background, but that should not be necessary with systemd, so I prefer not to add that in this PR.

pkg/runtime/grpc/info.go
flyth: Hmm, I think for now we can use the default value for it, so I'll remove the parameter - later on it might make sense to fetch the deploy information from a specific host/node (e.g. if using a tcp socket), but we don't have a UI for that right now.

pkg/gadget-service/service.go
```diff
@@ -253,7 +260,68 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
...
+func newUnixListener(network, address string, gid int) (net.Listener, error) {
+	if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("removing existing unix socket at %q: %w", address, err)
+	}
+
+	mask := syscall.Umask(0o777)
+	defer syscall.Umask(mask)
+
+	// If the given path is the default, try to create it and change permissions; if it's not the default, it is
+	// up to the user to manage it
```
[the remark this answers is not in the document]
flyth: You mean for `listener, err := net.Listen(network, address)`? There could be several reasons for that to fail (permissions, missing dir, ...), but you're right, it's missing the info on what it tried to do, so changing it to:

```go
return nil, fmt.Errorf("creating listener of type %q at %q: %w", network, address, err)
```

pkg/runtime/grpc/info.go
```diff
@@ -18,41 +18,62 @@ import (
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 )
 
-func loadRemoteDeployInfo() (*deployinfo.DeployInfo, error) {
-	ctx, cancelDial := context.WithTimeout(context.Background(), time.Second*ConnectTimeout)
-	defer cancelDial()
+// InitInfo loads the locally stored gadget catalog. If no catalog is stored locally,
+// it will try to fetch one from one of the gadget nodes and store it locally. It will issue warnings on
+// failures.
+func (r *Runtime) InitInfo(params *params.Params) error {
```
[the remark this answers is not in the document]
flyth: Will rename it to `InitDeployInfo()` and change the description - we renamed it from catalog to deploy info some time ago.

pkg/runtime/grpc/info.go
```diff
@@ -18,41 +18,62 @@ import (
...
+	defer cancelDial()
+	if params == nil {
+		// use defaults, if params are unset (e.g. automatic fetching
+		// of catalog after deployment)
+		params = r.ParamDescs().ToParams()
+	}
+	client, err := r.getClientFromRandomTarget(ctx, params)
 	if err != nil {
-		return nil, fmt.Errorf("dialing gadget pod on node %q: %w", pod.node, err)
+		return nil, fmt.Errorf("dialing random target: %w", err)
```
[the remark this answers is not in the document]
flyth: I'll add a debug message so you can see using `-v` which remote was selected and extend the err returned to contain that info.

cmd/ig/daemon.go
```diff
@@ -0,0 +1,104 @@
...
+		} else {
+			tmpGroup, err := user.LookupGroup(group)
+			if err != nil {
+				return fmt.Errorf("looking up group %q: %w", group, err)
+			}
+			gid, err = strconv.Atoi(tmpGroup.Gid)
+			if err != nil {
+				return fmt.Errorf("unexpected non-numeric group id %q for group %q", tmpGroup.Gid, group)
+			}
+		}
```
mauriciovasquezbernal, line 81: nit: I'm wondering if we should follow [2.4 chown, chgrp, chroot, id: Disambiguating user names and IDs](https://www.gnu.org/software/coreutils/manual/coreutils.html#Disambiguating-names-and-IDs) (TL;DR: first check for the group, then fallback to the numeric id) just to be sure there is not any confusion. Again, it's only done once so I don't think looking the group before has a big performance penalty.

cmd/ig/daemon.go
```diff
@@ -0,0 +1,104 @@
...
+
+		var socketPath string
+		socketType := socketURL.Scheme
+		switch socketType {
+		default:
+			return fmt.Errorf("invalid type for daemon-socket %q; please use 'unix' or 'tcp'", socketType)
```
mauriciovasquezbernal, line 87: out of curiosity: Why do you prefer to have the default case first? I always read switches like, if X happen then, if Y happen then, otherwise do the default. Having the default at the beginning breaks my mental model of switch cases :D. 

(btw, it's just a personal note, use whatever you think is the best).

cmd/ig/daemon.go
flyth: I don't have a strong opinion on this, but it seems™ like right now it's behaving the same as [`dockerd`](https://github.com/moby/moby/blob/master/cmd/dockerd/daemon.go).

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -212,56 +280,59 @@ func (r *Runtime) RunGadget(gadgetCtx runtime.GadgetContext) (runtime.CombinedGa
...
+	// If we're in Kubernetes connection mode, we need a custom dialer
+	if r.connectionMode == ConnectionModeKubernetesProxy {
+		opts = append(opts, grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
+			return NewK8SExecConn(ctx, target, time.Second*ConnectTimeout)
+		}))
+	}
+
+	conn, err := grpc.DialContext(dialCtx, "passthrough:///"+target.name, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("dialing node %q: %w", target.node, err)
```
mauriciovasquezbernal, line 318: nit: this message is misleading when running without kubernetes as node is empty. 

```
WARN[0030] could not load gadget info from remote: dialing random target: dialing node "": context deadline exceeded
```

pkg/runtime/grpc/grpc-runtime.go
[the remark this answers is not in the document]
mauriciovasquezbernal: This target works very well for kubernetes, there you have a pod and node, but for connection mode direct it's confusing. I'll say to leave the node empty and to udpate all debug / error messages to only include it if it's not empty.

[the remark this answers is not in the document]
flyth: I've changed it so that for unix sockets it'll now say "local" and for tcp targets it will be the target hostname.

cmd/ig/daemon.go
flyth: Hehe :) I think the idea was just to error out "early" as nothing fancy happens in the default case. I think I normally also tend to write the default last, if there's other mechanics involved.

cmd/common/sync.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2023 The Inspektor Gadget authors
```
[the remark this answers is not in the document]
flyth: > But OK, I will not put my veto against if it stays like it.

I'll take that and will try to do better next time :)

pkg/gadget-service/service.go
```diff
@@ -253,7 +260,68 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
...
+				log.Debugf("fixing ownership/mode on dir %q (uid: %d => %d, gid: %d => %d, mode: %d => %d)",
+					dir,
+					ufi.Uid, 0,
+					ufi.Gid, gid,
+					ufi.Mode, 0o710,
+				)
+				fixDirPermissions = true
+			}
+		}
+		if fixDirPermissions {
```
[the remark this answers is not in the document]
flyth: Hmm... yeah, why not. Simplifies the code a bit.

pkg/gadget-service/service.go
```diff
@@ -264,11 +332,20 @@ func (s *Service) Run(network, address string, serverOptions ...grpc.ServerOptio
...
-	if err != nil {
-		return err
+	switch runConfig.SocketType {
+	case "unix":
+		listener, err := newUnixListener(runConfig.SocketType, runConfig.SocketPath, runConfig.SocketGID)
+		if err != nil {
+			return fmt.Errorf("creating unix listener: %w", err)
+		}
+		s.listener = listener
+	default:
```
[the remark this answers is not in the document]
flyth: Actually, I think it's better to have a "tcp" case here and make the default return an error. (Will leave the default case at the bottom ;))

flyth:
> I tested it and got this behavior:
> 
> ```
> francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./gadgetctl trace open --host                                 michael/ig-daemon % u=
> Error: unknown flag: --host
> francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ sudo ./gadgetctl trace open --host                            michael/ig-daemon % u=
> [sudo] Mot de passe de francis : 
> RUNTIME.CONTAINERNAME                            PID        COMM             FD    ERR PATH                                              
>                                                  3892       ksystemstats     11    0   /proc/meminfo                                     
> ```
> 
> It is OK, but it should rather explode yelling I should be root?

I'm not sure I fully understand the issue here. Did you specify a valid user group like this: `ig daemon --group GID`? You should not need to be root to use `gadgetctl` if you specified that (and of course are member of that group).

> Also, is there any way to easily test the TCP way? I would like to take a look at it.

That should now work again:

```bash
./ig daemon -H tcp://127.0.0.1:9999
```

```bash
./gadgetctl trace open --remote-address tcp://127.0.0.1:9999
```

cmd/gadgetctl/main.go
```diff
@@ -0,0 +1,64 @@
...
+	common.AddVerboseFlag(rootCmd)
+
+	runtime := grpcruntime.New()
+
+	runtimeGlobalParams := runtime.GlobalParamDescs().ToParams()
+	err := runtime.Init(runtimeGlobalParams)
+	if err != nil {
+		log.Fatalf("initializing runtime: %v", err)
+	}
+	err = runtime.InitInfo(nil)
```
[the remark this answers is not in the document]
flyth: Added skipping "version" for now.

docs/ig.md
```diff
@@ -195,6 +195,95 @@ re-execute itself in a privileged systemd unit if necessary.
 This is possible because the "kubectl debug node" gives access to the systemd socket (`/run/systemd/private`) via the
 /host volume.
 
+### Using ig as a daemon
```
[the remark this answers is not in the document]
flyth: We surely need to completely restructure the ig docs sooner or later as it has become more complex but also slightly outdated - then it'd be a good idea to split it into multiple files.

.github/workflows/inspektor-gadget.yml
mqasimsarfraz, line 1: This is missing steps of publishing the `gadgetctl` binary in the release artifact. I believe we need to add them from [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/workflows/inspektor-gadget.yml#L1377) onwards.

flyth: Ah, good point, thanks! I added it, hopefully it's correct this way :)

mqasimsarfraz: That was quick, thank!

mauriciovasquezbernal (approved):
I think it's ready to go! Last minor nits but it's looking fine! 

Good job!

flyth:
Thanks for the reviews!

pkg/runtime/grpc/info.go
```diff
@@ -18,41 +18,57 @@ import (
...
-	conn, err := grpc.DialContext(ctx, "", dialOpt, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	// use default params for now
+	params := r.ParamDescs().ToParams()
+	client, err := r.getClientFromRandomTarget(ctx, params)
 	if err != nil {
-		return nil, fmt.Errorf("dialing gadget pod on node %q: %w", pod.node, err)
+		return nil, fmt.Errorf("dialing random target: %w", err)
 	}
-	client := api.NewGadgetManagerClient(conn)
-	defer conn.Close()
```
mauriciovasquezbernal, line 55: When is `conn` closed? I think `getClientFromRandomTarget` should return the connection instead of the client, so the caller has the opportunity to call `defer conn.Close()`. (opened https://github.com/inspektor-gadget/inspektor-gadget/pull/2092 to fix it)

Issue #2077, opened 2023-09-25
Parameter title is not displayed when using `--help`


eiffel-fl:
`Parameter.Title` is not displayed when using `--help`:

```bash
$ ./ig trace open --help
...
                               
      --full-path              
  -h, --help                   help for open
```

The parameter is defined here:
https://github.com/inspektor-gadget/inspektor-gadget/blob/dbacb0139cb99bf5ddec716df14a70742f7f6334/pkg/gadgets/trace/open/tracer/gadget.go#L49-L54

mauriciovasquezbernal:
Was it fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/2079? @eiffel-fl

PR #2079, opened 2023-09-25
pkg: Add missing Description for some ParamDesc

eiffel-fl:


flyth:
> @flyth I think there a misunderstanding somewhere between `Title` and `Description` (hence #2077).
> Can you please shed some light on why do we need the two of them?

Key is used for the parameter on cobra (--key) (and also for our API), title is used as a more human readable form (for example, if key is `ip`, title could be `IP Address`) that defaults to the value of key (and is not used at all on the CLI, but for example in Headlamps Web UI) - and description is what is shown behind the parameter as explanation of what it does. So for the CLI, key & description is all that is needed - if you want to add a layer of extra fancy on top, define a title as well :)

flyth:
> > > @flyth I think there a misunderstanding somewhere between `Title` and `Description` (hence #2077).
> > > Can you please shed some light on why do we need the two of them?
> > 
> > 
> > Key is used for the parameter on cobra (--key) (and also for our API), title is used as a more human readable form (for example, if key is `ip`, title could be `IP Address`) that defaults to the value of key (and is not used at all on the CLI, but for example in Headlamps Web UI) - and description is what is shown behind the parameter as explanation of what it does. So for the CLI, key & description is all that is needed - if you want to add a layer of extra fancy on top, define a title as well :)
> 
> Makes sense! Thank you! So, if we want to be complete (both from CLI and other UI), we should define all of them? Should we then forbid them to be empty?

I don't think we need to forbid that; I think in 80% of the cases, the key is just fine if first character is uppercased (which the UI can easily do). So `--address` becomes `Address`, `--namespace` becomes `Namespace` and so on.

flyth:
> I added another commit to ensure the Description is not empty.

Thinking about this again, I'm not sure whether we should enforce this programmatically; containerized gadgets will also use this and IMHO we should leave it up to the user to provide a description or not (and maybe just warn at build time). If it were just for our own gadgets, I would be all for it.

flyth:
> > > I added another commit to ensure the Description is not empty.
> > 
> > 
> > Thinking about this again, I'm not sure whether we should enforce this programmatically; containerized gadgets will also use this and IMHO we should leave it up to the user to provide a description or not (and maybe just warn at build time). If it were just for our own gadgets, I would be all for it.
> 
> I can use a warning instead, but having the `Description` to indicate what does a parameter is the bare minimum. In case you use `--help`, you would just get the parameters list without any explanation, which seems amateur and/or odd.

Yeah, I totally agree - but looking forward to the containerized gadgets, your messages would be shown to the _user_ of a gadget, who probably can't do anything about it. So IMHO we should warn/error/enforce when building the gadget, so the actual developer can correct it.

If this PR makes sure that our internal gadgets are fine in regards to descriptions, I think we can skip warnings/errors, as we're probably not adding many gadgets and/or functionality to the old gadgets before switching to the containerized versions.

flyth (approved):
LGTM, thanks!

Sorry for the back and forth regarding warnings/errors!

PR #2083, opened 2023-09-27
pkg/helpers: Fix logic to detect if kernel supports bpf_ktime_get_boo…

eiffel-fl:
…t_ns().

The kernel defines many BPF helpers functions and all of them have an ID [1]. Sadly, we cannot compare the number of helper functions with the ID to know if the kernel supports the given function.
Indeed, the kernel defines two abstract helper functions:
* unspec which has ID 0 [2]
* and __BPF_FUNC_MAX_ID which indicates the last ID [3].

The solution is then to test that latest helper is __BPF_FUNC_MAX_ID and that bpf_ktime_get_boot_ns() ID is lower.

Fixes: f87486947273 ("Rewrite detectBpfKtimeGetBootNs using sync.Once")
Fixes: ffb7d9910f2d ("gadgets: get timestamp from userspace on Linux <5.7")
[1]: https://github.com/torvalds/linux/blob/0e945134b680040b8613e962f586d91b6d40292d/include/uapi/linux/bpf.h#L5646
[2]: https://github.com/torvalds/linux/blob/0e945134b680040b8613e962f586d91b6d40292d/include/uapi/linux/bpf.h#L5647
[3]: https://github.com/torvalds/linux/blob/0e945134b680040b8613e962f586d91b6d40292d/include/uapi/linux/bpf.h#L5873

### DIFF
--- a/pkg/gadgets/helpers.go
+++ b/pkg/gadgets/helpers.go
@@ -196,20 +196,26 @@ var (
 // in the current kernel. False negatives are possible if BTF is not available.
 func DetectBpfKtimeGetBootNs() bool {
 	bpfKtimeGetBootNsOnce.Do(func() {
+		bpfKtimeGetBootNsExists = false
+
 		btfSpec, err := btf.LoadKernelSpec()
 		if err != nil {
-			bpfKtimeGetBootNsExists = false
 			return
 		}
 
 		enum := &btf.Enum{}
 		err = btfSpec.TypeByName("bpf_func_id", &enum)
 		if err != nil {
-			bpfKtimeGetBootNsExists = false
 			return
 		}
 
-		bpfKtimeGetBootNsExists = len(enum.Values) >= BpfKtimeGetBootNsFuncID
+		for _, value := range enum.Values {
+			if value.Name == "BPF_FUNC_ktime_get_boot_ns" && value.Value == BpfKtimeGetBootNsFuncID {
+				bpfKtimeGetBootNsExists = true
+
+				return
+			}
+		}
 	})
 
 	return bpfKtimeGetBootNsExists


pkg/gadgets/helpers.go
```diff
@@ -209,7 +209,12 @@ func DetectBpfKtimeGetBootNs() bool {
 			return
 		}
 
-		bpfKtimeGetBootNsExists = len(enum.Values) >= BpfKtimeGetBootNsFuncID
+		last := enum.Values[len(enum.Values) - 1]
+		if last.Name != "__BPF_FUNC_MAX_ID" {
+			panic("Last BPF helper must be __BPF_FUNC_MAX_ID")
```
alban, line 214: Why panic?

```suggestion
			bpfKtimeGetBootNsExists = false
			return
```

mauriciovasquezbernal:
Why not to loop on the enum values looking for the one corresponding to this function:

```
		for _, v := range enum.Values {
			if v.Value == BpfKtimeGetBootNsFuncID && v.Name == "BPF_FUNC_ktime_get_boot_ns" {
				bpfKtimeGetBootNsExists = true
				return
			}
		}
```

It's slower but I think the code is easier to follow and we don't have to rely on the last one being `__BPF_FUNC_MAX_ID`. 
The check on `BpfKtimeGetBootNsFuncID` is not strictly needed, but doesn't hurt.

pkg/gadgets/helpers.go
```diff
@@ -196,20 +196,26 @@ var (
...
 		enum := &btf.Enum{}
 		err = btfSpec.TypeByName("bpf_func_id", &enum)
 		if err != nil {
-			bpfKtimeGetBootNsExists = false
 			return
 		}
 
-		bpfKtimeGetBootNsExists = len(enum.Values) >= BpfKtimeGetBootNsFuncID
+		for _, value := range enum.Values {
+			if value.Value == BpfKtimeGetBootNsFuncID {
```
mauriciovasquezbernal, line 213: We also need to check the string value, otherwise we'll have an issue similar to the one before. For instance, this check will fail on this specific commit https://github.com/torvalds/linux/blob/0a05861f80fe7d4dcfdabcc98d9854947573e072/include/uapi/linux/bpf.h#L3154, `sk_assign` is 124 and `__BPF_FUNC_MAX_ID` is 125, so this logic will indicate the helper is available when it's not. 

If we want to avoid looping through the array, we can take into consideration that the array is built as:
[0] -> unspec
[1...N-2]: helpers (N-2 being the last helper)
[N-1]: FUNC_MAX_ID

To check if the helper exists we can do `bpfKtimeGetBootNsExists = len(enum.Values) >= BpfKtimeGetBootNsFuncID + 2`
But TBH I think a simple linear search is a lot easier to understand (and to get it right).

mauriciovasquezbernal:
> once before loading the eBPF code

Just to be sure we're on the same page, the function is called once in the process life as it uses bpfKtimeGetBootNsOnce.

pkg/gadgets/helpers.go
mauriciovasquezbernal: > I think we can only check the name and ID is not really useful here.

Right, I added that for two (not really important) reasons:
- To speed up a bit the check
- I remembered somebody mentioned that a kernel got a backporting wrong and the ID of a helper was messed up.

mauriciovasquezbernal (approved):
LGTM. Thanks for the fix!

PR #2092, opened 2023-09-29
runtimes/grpc: Fix connection leakage

mauriciovasquezbernal:
Refactor the code a bit to allow the caller to close the connection.

Fixes: 2a15d06e4d73 ("pkg/runtime/grpc: add support for direct gRPC connections")


flyth (approved):
Nice catch! Much better this way, thanks!

PR #2097, opened 2023-10-02
Refactor pkg/oci

mauriciovasquezbernal:
This is a complete refactoring of the oci package. The whole idea is to implement all the OCI related functionalities we need in this package, avoiding other parts of the code having to interact with packages like "github.com/docker" or "oras.land/oras-go/v2".

The reasoning behind is:
- We can concentrate all OCI related code in the same place
- It's a lot easier to test if it's a separated package

This package now provides methods for pulling, pushing, tagging and building OCI images. Other methods that we previously exposed are now internal.

This commit also:
- Uses -ing in all error messages
- Small esthetic changes to cmd/*

NOTE: Ideally this should be multiple commits, but it's rather difficult to split a refactor like it into smaller pieces.

### DIFF
--- a/pkg/oci/build.go
+++ b/pkg/oci/build.go
@@ -1,0 +1,208 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package oci
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/docker/distribution/reference"
+	"github.com/opencontainers/image-spec/specs-go"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/errdef"
+)
+
+const (
+	ArchAmd64 = "amd64"
+	ArchArm64 = "arm64"
+)
+
+const (
+	eBPFObjectMediaType = "application/vnd.gadget.ebpf.program.v1+binary"
+	metadataMediaType   = "application/vnd.gadget.config.v1+yaml"
+)
+
+type BuildGadgetImageOpts struct {
+	// List of eBPF objects to include in the image. The key is the architecture and the value
+	// is the path to the eBPF object.
+	EBPFObjectPaths map[string]string
+	// Path to the metadata file.
+	MetadataPath string
+}
+
+// BuildGadgetImage creates an OCI image with the objects provided in opts. The image parameter in
+// the "name:tag" format is used to name and tag the created image. If it's empty the image is not
+// named.
+func BuildGadgetImage(ctx context.Context, opts *BuildGadgetImageOpts, image string) (*GadgetImageDesc, error) {
+	ociStore, err := getLocalOciStore()
+	if err != nil {
+		return nil, fmt.Errorf("getting oci store: %w", err)
+	}
+
+	indexDesc, err := createImageIndex(ctx, ociStore, opts)
+	if err != nil {
+		return nil, fmt.Errorf("creating image index: %w", err)
+	}
+
+	imageDesc := &GadgetImageDesc{
+		Digest: indexDesc.Digest.String(),
+	}
+
+	if image != "" {
+		targetImage, err := normalizeImageName(image)
+		if err != nil {
+			return nil, fmt.Errorf("normalizing image: %w", err)
+		}
+
+		err = ociStore.Tag(ctx, indexDesc, targetImage.String())
+		if err != nil {
+			return nil, fmt.Errorf("tagging manifest: %w", err)
+		}
+
+		imageDesc.Repository = targetImage.Name()
+		if ref, ok := targetImage.(reference.Tagged); ok {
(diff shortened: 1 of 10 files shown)


pkg/oci/oci.go
```diff
@@ -281,22 +399,126 @@ func NewAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
-			return nil, fmt.Errorf("create auth client: %w", err)
+			return nil, fmt.Errorf("creating auth client: %w", err)
 		}
 		repo.Client = client
 	}
 
 	return repo, nil
 }
+
+func getImageListDescriptor(imageStore oras.ReadOnlyTarget, image string) (ocispec.Index, error) {
```
burak-ok, line 422:
```suggestion
func getImageListDescriptor(imageStore oras.ReadOnlyTarget, normalizedImage string) (ocispec.Index, error) {
```

From the codepath this function only gets normalized image names, so this is not an issue. The issue is when this function in the future gets called with a non normalized one.
So I think the least we should do to "catch" and future issues is to name the parameter that way (Or another way - you know the problems with naming things :D)

Now that I've thought about it a little more. Maybe we should introduce a new type. So the compiler always checks our usages of it. I just don't want headaches in the future xD. But I think this would be another PR

burak-ok (approved):
LGTM besides that (larger) comment that I have. But that should be another PR.

Tested it locally but probably didn't test all corner cases

pkg/oci/oci.go
mauriciovasquezbernal: I see that parameter is passed to `imageStore.Resolve` and it's called there `reference`, so I think it's fine to call it reference. I'm not sure introducing a new type is a good idea, perhaps it adds a lot of boilerplate to cast from one to the other.

burak-ok: Yep, the boilerplate code is the disadvantage. And I also don't know how "annoying" or how much boilerplate would be needed.

pkg/oci/build.go
```diff
@@ -0,0 +1,202 @@
...
+const (
+	ARCH_AMD64 = "amd64"
+	ARCH_ARM64 = "arm64"
+)
+
+type BuildGadgetImageOpts struct {
+	// List of eBPF objects to include on the image. The key is the architecture and the value
+	// is the path to the eBPF object.
+	EBPFObjectPaths map[string]string
+	// Path to the medatada file
```
flyth, line 43:
```suggestion
	// Path to the metadata file
```

pkg/oci/build.go
```diff
@@ -0,0 +1,202 @@
...
+	"oras.land/oras-go/v2/errdef"
+)
+
+const (
+	ARCH_AMD64 = "amd64"
+	ARCH_ARM64 = "arm64"
+)
+
+type BuildGadgetImageOpts struct {
+	// List of eBPF objects to include on the image. The key is the architecture and the value
```
flyth, line 40:
```suggestion
	// List of eBPF objects to include in the image. The key is the architecture and the value
```

pkg/oci/build.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return nil
+}
+
+func createEbpfProgramDesc(target oras.Target, progFilePath string) (ocispec.Descriptor, error) {
+	progBytes, err := os.ReadFile(progFilePath)
+	if err != nil {
+		return ocispec.Descriptor{}, fmt.Errorf("reading eBPF program file: %w", err)
+	}
+	progDesc := content.NewDescriptorFromBytes("application/vnd.gadget.ebpf.program.v1+binary", progBytes)
```
flyth, line 97: Maybe use consts for `application/vnd.gadget.ebpf.program.v1+binary` and `application/vnd.gadget.config.v1+yaml` as well?

pkg/oci/build.go
```diff
@@ -0,0 +1,202 @@
...
+	// List of eBPF objects to include on the image. The key is the architecture and the value
+	// is the path to the eBPF object.
+	EBPFObjectPaths map[string]string
+	// Path to the medatada file
+	MetadataPath string
+}
+
+// BuildGadgetImage creates an OCI image with the objects provided in opts. The created image is
+// tagged if image is not empty.
+func BuildGadgetImage(opts *BuildGadgetImageOpts, image string) (*GadgetImageDesc, error) {
```
flyth, line 49: Could you clarify a bit what exactly "The created image is tagged if image is not empty" means (maybe add an example?).

pkg/oci/build.go
```diff
@@ -0,0 +1,202 @@
...
+func createEbpfProgramDesc(target oras.Target, progFilePath string) (ocispec.Descriptor, error) {
+	progBytes, err := os.ReadFile(progFilePath)
+	if err != nil {
+		return ocispec.Descriptor{}, fmt.Errorf("reading eBPF program file: %w", err)
+	}
+	progDesc := content.NewDescriptorFromBytes("application/vnd.gadget.ebpf.program.v1+binary", progBytes)
+	progDesc.Annotations = map[string]string{
+		ocispec.AnnotationTitle:       "program.o",
+		ocispec.AnnotationAuthors:     "TODO: authors",
+		ocispec.AnnotationDescription: "TODO: description",
```
flyth, line 101: Should these be exposed?

flyth (commented):
Nice, I really like it!

Do you think it makes sense to promote the contexts to parameters where applicable now that you're refactoring anyway (and use `context.Background()` for now in the callers)?

pkg/oci/build.go
mauriciovasquezbernal: Added the comment in a different way. Not sure if the result is better.

pkg/oci/build.go
mauriciovasquezbernal: Yes, they should. However we're still understanding the different annotations to be used, so I prefer to skip that for now.

mauriciovasquezbernal:
> Do you think it makes sense to promote the contexts to parameters where applicable now that you're refactoring anyway (and use `context.Background()` for now in the callers)?

Nice suggestion! I used context.TODO() in the caller to make it clear we still need to pass a valid context there later on.

pkg/oci/build.go
flyth: Yeah, that really helps, thanks!

flyth (approved):
LGTM, thanks!

PR #2099, opened 2023-10-03
pkg/oci: Use distribution/reference instead of docker/distribution/re…

eiffel-fl:
…ference.

The former should be used as the latter is deprecated [1].

[1]: https://pkg.go.dev/github.com/docker/distribution/reference


mauriciovasquezbernal (approved):
LGTM. Thanks for the fix.

mauriciovasquezbernal:
Merging to rebase https://github.com/inspektor-gadget/inspektor-gadget/pull/2097.

Issue #1420, opened 2023-03-10
Automatically build and push "helper" container images


mauriciovasquezbernal:
We use some helper container images in Inspektor Gadget that are pushed to https://github.com/orgs/inspektor-gadget/packages

- Dockerfiles/ebpf-builder.Dockerfile
- Dockerfiles/ig-builder.Dockerfile

The purpose of this issues is to follow the same approach as in https://github.com/inspektor-gadget/inspektor-gadget/pull/1414/commits/cccdf0296a5df8e4543427deeb89c67cd10752dc to build and push those images automatically when they are modified.

EDIT: Images should be published on released to and this comment should be fixed https://github.com/inspektor-gadget/inspektor-gadget/pull/2103#discussion_r1346181785.

mauriciovasquezbernal:
Fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/1472.

mauriciovasquezbernal:
Well, no actually. It's avoid different images.

PR #2103, opened 2023-10-04
Rename ebpf builder image

mauriciovasquezbernal:
Rename from ghcr.io/inspektor-gadget/inspektor-gadget-ebpf-builder to ghcr.io/inspektor-gadget/ebpf-builder as "inspektor-gadget" is redundant.

---

I pushed the image manually, however it should be automatically pushed once https://github.com/inspektor-gadget/inspektor-gadget/issues/1420 is fixed.

### DIFF
--- a/cmd/common/image/build.go
+++ b/cmd/common/image/build.go
@@ -49,7 +49,7 @@ import (
 var makefile []byte
 
 const (
-	DEFAULT_BUILDER_IMAGE = "ghcr.io/inspektor-gadget/inspektor-gadget-ebpf-builder:latest"
+	DEFAULT_BUILDER_IMAGE = "ghcr.io/inspektor-gadget/ebpf-builder:latest"
 	DEFAULT_EBPF_SOURCE   = "program.bpf.c"
 	DEFAULT_METADATA      = "gadget.yaml"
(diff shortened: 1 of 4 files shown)


cmd/common/image/build.go
```diff
@@ -49,7 +49,7 @@ import (
 var makefile []byte
 
 const (
-	DEFAULT_BUILDER_IMAGE = "ghcr.io/inspektor-gadget/inspektor-gadget-ebpf-builder:latest"
+	DEFAULT_BUILDER_IMAGE = "ghcr.io/inspektor-gadget/ebpf-builder:latest"
```
alban, line 52: Once https://github.com/inspektor-gadget/inspektor-gadget/issues/1420 is fixed, `DEFAULT_BUILDER_IMAGE` could be set to a tag such as "ghcr.io/inspektor-gadget/ebpf-builder:v0.21.0" if cmd/common/version.go defines the variable `version` as a release.

In that way, the old versions of `ig` would not break when a new version update the ebpf-builder image with some incompatible command line.

This problem existed before this PR, so I'll approve it.

alban (approved):
LGTM but please notice the comment on the use of ":latest".

PR #2105, opened 2023-10-04
ci: Use correct sources for 'Publish gadget img manifest'

mqasimsarfraz:
Use correct sources for publishing gadget image manifest to registry. Currently if fails with:

```
Publish the manifest list
> ERROR: no sources specified
```

Fixes: 77582cbbf31d ("treewide: Remove bcc flavor and rename default.")

## Testing done

- https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/6408490245/job/17397845995


mqasimsarfraz:
> I am nonetheless wondering why the CI did not catch these mistakes 🤔.

I found it via CI run but only after it was merged since 'Publish gadget img manifest' is only done for `main`. Maybe we pass such changes through `citest`, just to be sure.

Issue #1795, opened 2023-06-28
Version skew warning message is not printed


mauriciovasquezbernal:
We printed a warning in case the kubectl-gadget and the gadget pod were not the same version when we're using custom resources to run gadgets:

https://github.com/inspektor-gadget/inspektor-gadget/blob/83dde7d1ea4e16c14118531a1acdaf92123ff723/cmd/kubectl-gadget/utils/trace.go#L185

However, this code path is not used anymore after https://github.com/inspektor-gadget/inspektor-gadget/pull/1396. We should implement that again to be sure we inform users.

## Implementation options

We already have a mechanism to send the deploy info from the gadget pods to kubectl-gadget

https://github.com/inspektor-gadget/inspektor-gadget/blob/83dde7d1ea4e16c14118531a1acdaf92123ff723/pkg/runtime/grpc/grpc-runtime.go#L80

We could include the gadget pod version there and print the warning around these lines 

https://github.com/inspektor-gadget/inspektor-gadget/blob/83dde7d1ea4e16c14118531a1acdaf92123ff723/pkg/runtime/grpc/grpc-runtime.go#L73-L84

There's an Version field in the proto definition

https://github.com/inspektor-gadget/inspektor-gadget/blob/83dde7d1ea4e16c14118531a1acdaf92123ff723/pkg/gadgettracermanager/api/gadgettracermanager.proto#L133

We can use that field for this purpose. However, in the server side it's not filled yet 

https://github.com/inspektor-gadget/inspektor-gadget/blob/83dde7d1ea4e16c14118531a1acdaf92123ff723/pkg/gadget-service/service.go#L76

~We can use the [`INSPEKTOR_GADGET_VERSION`](https://github.com/inspektor-gadget/inspektor-gadget/blob/83dde7d1ea4e16c14118531a1acdaf92123ff723/pkg/resources/manifests/deploy.yaml#L163) env variable to get it.~

Edit: Using `INSPEKTOR_GADGET_VERSION` doesn't work as it's actually set to the kubectl-gadget version. New options are:
[...]

PR #2113, opened 2023-10-06
kubectl-gadget/ig-k8s: change connection to gadget service from exec to port-forward

flyth:
This PR changes the default connection method used by `kubectl-gadget` to connect to the gadget-service running on the `gadgettracermgr`.

Until now, we executed a `socat` binary (using the exec method of the Kubernetes API server) inside the gadget pod to be able to forward gRPC traffic to a unix socket, to be able to talk to the gRPC endpoint of the gadget-service.

After this PR, the gadget-service will listen on a tcp socket by default (127.0.0.1:8080) and `kubectl-gadget` will connect to it using the port-forward method of the Kubernetes API server.

The reason that we used a unix socket in the past, was to properly secure the socket. As the gadget pod was running in the host network namespace, running our service on a tcp port would have exposed it on the host, making it accessible from unknown sources. Now that we're running in our own network namespace, we're not exposing it anymore outside of our own pod.

Advantages are:
* granting the previously required privilege to `exec` things on our gadget pod can be much more dangerous than just allowing a port forwarding to the gadget service
* removes the requirement of having to use (and bundling) the `socat` binary to actually connect to the previously used unix socket

### DIFF
--- a/pkg/runtime/grpc/grpc-runtime.go
+++ b/pkg/runtime/grpc/grpc-runtime.go
@@ -52,24 +52,41 @@ const (
 	ConnectionModeDirect ConnectionMode = iota
 
 	// ConnectionModeKubernetesProxy will connect to a gRPC endpoint through a kubernetes API server by first looking
-	// up an appropriate target node using the kubernetes API, then using the exec endpoint of the kubernetes API to
-	// spawn a socat process that then will forward the gRPC connection to the unix socket that is listening inside
-	// the IG container (see gadgettracermgr).
+	// up an appropriate target node using the kubernetes API, then using either the exec (when used with
+	// KubernetesProxyConnectionMethodUnix) or the port forward (when used with KubernetesProxyConnectionMethodTCP)
+	// endpoint of the Kubernetes API to forward the gRPC connection to the service listener (see gadgettracermgr).
 	ConnectionModeKubernetesProxy
 )
 
 const (
-	ParamNode          = "node"
-	ParamRemoteAddress = "remote-address"
+	ParamNode              = "node"
+	ParamRemoteAddress     = "remote-address"
+	ParamConnectionMethod  = "connection-method"
+	ParamConnectionTimeout = "connection-timeout"
+
+	// ParamGadgetServiceTCPPort is only used in combination with KubernetesProxyConnectionMethodTCP
+	ParamGadgetServiceTCPPort = "tcp-port"
 
 	// ConnectTimeout is the time in seconds we wait for a connection to the remote to
 	// succeed
-	ConnectTimeout = 30
+	ConnectTimeout = 5
 
 	// ResultTimeout is the time in seconds we wait for a result to return from the gadget
 	// after sending a Stop command
 	ResultTimeout = 30
+
+	// KubernetesProxyConnectionMethodTCP uses the Kubernetes API Server using port-forwarding to connect to the
+	// gadget service
+	KubernetesProxyConnectionMethodTCP = "tcp"
+
+	// KubernetesProxyConnectionMethodUnix uses the Kubernetes API Server using exec to connect to the gadget
+	// service
+	// Deprecated: due to the complexity and security concerns regarding the required privilege to use this,
+	// KubernetesProxyConnectionMethodUnix is now the standard.
+	KubernetesProxyConnectionMethodUnix = "unix"
 )
+
+var kubernetesProxyConnectionMethods = []string{KubernetesProxyConnectionMethodTCP, KubernetesProxyConnectionMethodUnix}
 
 type Runtime struct {
 	info           *deployinfo.DeployInfo
@@ -141,7 +158,14 @@ func (r *Runtime) ParamDescs() params.ParamDescs {
 }
 
 func (r *Runtime) GlobalParamDescs() params.ParamDescs {
-	p := params.ParamDescs{}
+	p := params.ParamDescs{
+		{
+			Key:          ParamConnectionTimeout,
+			Description:  "Maximum time to establish a connection to remote target in seconds",
+			DefaultValue: fmt.Sprintf("%d", ConnectTimeout),
+			TypeHint:     params.TypeUint,
+		},
+	}
 	switch r.connectionMode {
 	case ConnectionModeDirect:
 		p.Add(params.ParamDescs{
@@ -154,6 +178,20 @@ func (r *Runtime) GlobalParamDescs() params.ParamDescs {
 		}...)
 		return p
 	case ConnectionModeKubernetesProxy:
+		p.Add(params.ParamDescs{
+			{
+				Key:            ParamConnectionMethod,
+				Description:    "Method used to connect to the gadget service; either tcp or unix (deprecated)",
+				DefaultValue:   KubernetesProxyConnectionMethodTCP,
+				PossibleValues: kubernetesProxyConnectionMethods,
+			},
+			{
+				Key:          ParamGadgetServiceTCPPort,
+				Description:  "Port used to connect to the gadget service",
+				DefaultValue: fmt.Sprintf("%d", api.GadgetServicePort),
+				TypeHint:     params.TypeUint16,
+			},
+		}...)
 		return p
 	}
 	panic("invalid connection mode set for grpc-runtime")
@@ -248,7 +286,7 @@ func (r *Runtime) getTargets(ctx context.Context, params *params.Params) ([]targ
 }
 
 func (r *Runtime) GetGadgetInfo(ctx context.Context, desc gadgets.GadgetDesc, gadgetParams *params.Params, args []string) (*runTypes.GadgetInfo, error) {
-	ctx, cancelDial := context.WithTimeout(ctx, time.Second*ConnectTimeout)
+	ctx, cancelDial := context.WithTimeout(ctx, time.Second*time.Duration(r.globalParams.Get(ParamConnectionTimeout).AsUint()))
 	defer cancelDial()
 
 	// use default params for now
@@ -311,7 +349,9 @@ func (r *Runtime) getConnToRandomTarget(ctx context.Context, runtimeParams *para
 	}
 	target := targets[0]
 	log.Debugf("using target %q (%q)", target.addressOrPod, target.node)
-	conn, err := r.dialContext(ctx, target)
+
+	timeout := time.Second * time.Duration(r.globalParams.Get(ParamConnectionTimeout).AsUint())
+	conn, err := r.dialContext(ctx, target, timeout)
 	if err != nil {
 		return nil, fmt.Errorf("dialing %q (%q): %w", target.addressOrPod, target.node, err)
 	}
@@ -360,7 +400,7 @@ func (r *Runtime) runGadgetOnTargets(
 	return results, results.Err()
 }
 
-func (r *Runtime) dialContext(dialCtx context.Context, target target) (*grpc.ClientConn, error) {
+func (r *Runtime) dialContext(dialCtx context.Context, target target, timeout time.Duration) (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithBlock(),
@@ -368,8 +408,22 @@ func (r *Runtime) dialContext(dialCtx context.Context, target target) (*grpc.Cli
 
 	// If we're in Kubernetes connection mode, we need a custom dialer
 	if r.connectionMode == ConnectionModeKubernetesProxy {
+		mode := r.globalParams.Get(ParamConnectionMethod).String()
 		opts = append(opts, grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
-			return NewK8SExecConn(ctx, target, time.Second*ConnectTimeout)
+			switch mode {
+			case KubernetesProxyConnectionMethodTCP:
+				port := r.globalParams.Get(ParamGadgetServiceTCPPort).AsUint16()
+				return NewK8SPortFwdConn(ctx, target, port, timeout)
+			case KubernetesProxyConnectionMethodUnix: // deprecated
+				log.Warnf("using deprecated connection mode KubernetesProxyConnectionMethodUnix")
+				return NewK8SExecConn(ctx, target, timeout)
+			default:
+				return nil, fmt.Errorf(
+					"invalid connection method %q (valid modes: %+v)",
+					mode,
+					kubernetesProxyConnectionMethods,
+				)
+			}
 		}))
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/runtime/grpc/k8s-portfwd-dialer.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("parsing restConfig.Host: %w", err)
+	}
+
+	targetURL.Path = path.Join(
+		"api", "v1",
+		"namespaces", "gadget",
+		"pods", conn.podName,
+		"portforward",
+	)
```
mauriciovasquezbernal, line 79: I'm wondering that's the advantage over a simple `fmt.Sprintf("/api/v1/....%s.....", conn.podName)`

pkg/runtime/grpc/k8s-portfwd-dialer.go
```diff
@@ -0,0 +1,157 @@
...
+	go func() {
+		message, err := io.ReadAll(errorStream)
+		switch {
+		case err != nil:
+			log.Errorf("k8sPortFwd connection: reading from error stream: %v", err)
+		case len(message) > 0:
+			log.Errorf("k8sPortFwd connection: forwarding port: %v", string(message))
+		}
+		close(errorChan)
+	}()
```
mauriciovasquezbernal, line 111: I can't understand this code. errorChan is never used below, perhaps instead of printing the error it should be sent to the channel to be read and handled below? (or just remove the channel?)

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -69,6 +70,14 @@ const (
 	// ResultTimeout is the time in seconds we wait for a result to return from the gadget
 	// after sending a Stop command
 	ResultTimeout = 30
+
+	// ConnectionMethodTCP uses the Kubernetes API Server using port-forwarding to connect to the
```
mauriciovasquezbernal, line 74: nit: perhaps `ConnectionProtocol` is better?

mauriciovasquezbernal (commented):
Looks very good, I added some comments but almost all of them are nits.

I tested it and I think the connection timeout is too high, it takes it 30 seconds to return an error:

```bash
$ time ./kubectl-gadget trace exec --connection-method=unix-socket
K8S.NODE               K8S.NAMESPACE          K8S.POD                K8S.CONTAINER          PID        PPID       COMM        RET ARGS                        
Error: running gadget: dialing target on node "ubuntu-hirsute": dialing "gadget-4wtpm" ("ubuntu-hirsute"): context deadline exceeded

real	0m30.699s
user	0m0.113s
sys	0m0.134s
```

Other tests worked fine.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -69,6 +70,14 @@ const (
...
+	// gadget service
+	ConnectionMethodTCP = "tcp"
+
+	// ConnectionMethodUnix uses the Kubernetes API Server using exec to connect to the gadget
+	// service; this method is deprecated
+	ConnectionMethodUnix = "unix-socket"
```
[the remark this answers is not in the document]
flyth: I used unix before, but it seemed too unclear; "tcp" implies that it's something about connections, "unix" just points to the OS. But I'm fine with either.

pkg/runtime/grpc/k8s-portfwd-dialer.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+)
+
+type k8sPortFwdDialer struct {
+	io.Writer
+	io.Reader
+	conn       httpstream.Connection
+	stream     httpstream.Stream
+	podName    string
+	cancelFunc func()
```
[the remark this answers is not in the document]
flyth: And it's telling the truth! :)

pkg/runtime/grpc/k8s-portfwd-dialer.go
flyth: I don't really see one in this case - I believe the code handling the port forwarding in k8s used this, so I went with it.

pkg/runtime/grpc/k8s-portfwd-dialer.go
flyth: Leftover code :) I removed the channel. There's no real place where we could send this error to - but the overlying protocol should also return an error, so I think we're fine.

pkg/runtime/grpc/grpc-runtime.go
flyth: Hmm, I don't know - sure, "tcp" is a protocol, but is "unix" as well?

mauriciovasquezbernal: Right, "method" is better then.

pkg/runtime/grpc/grpc-runtime.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I just wanted to keep it aligned with the network parameter in `net.Listen`: `The network must be "tcp", "tcp4", "tcp6", "unix" or "unixpacket".`. But I'm ok with either too, so I leave it up to you.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -368,8 +385,16 @@ func (r *Runtime) dialContext(dialCtx context.Context, target target) (*grpc.Cli
 
 	// If we're in Kubernetes connection mode, we need a custom dialer
 	if r.connectionMode == ConnectionModeKubernetesProxy {
+		mode := r.globalParams.Get(ParamConnectionMethod).String()
 		opts = append(opts, grpc.WithContextDialer(func(ctx context.Context, s string) (net.Conn, error) {
-			return NewK8SExecConn(ctx, target, time.Second*ConnectTimeout)
+			switch mode {
+			case ConnectionMethodTCP:
+				return NewK8SPortFwdConn(ctx, target, time.Second*ConnectTimeout)
+			case ConnectionMethodUnix: // deprecated
```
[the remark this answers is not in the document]
flyth: I don't have one :) This allows newer `kubectl-gadget` versions to connect to not yet updated nodes - even though we generally don't support using different versions between those. So I'd say one or two releases at most.

pkg/runtime/grpc/k8s-portfwd-dialer.go
flyth: > `Join()` is platform agnostic, in case we run this code on Windows node (I do not know if it is possible and am wondering about knowing).

That'd be `filepath`, then. Here we're building a URL with the `path` package, so it's only handling forward slashes.

pkg/gadget-service/api/consts.go
```diff
@@ -25,5 +25,6 @@ const (
 
 const (
 	GadgetServiceSocket = "/run/gadgetservice.socket"
+	GadgetServicePort   = 8080
```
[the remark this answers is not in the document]
flyth: The `api` package just defines the defaults and those are const.

But you're right, this should be configurable in both client and server.

pkg/runtime/grpc/k8s-portfwd-dialer.go
```diff
@@ -0,0 +1,154 @@
...
+	errorStream.Close()
+
+	go func() {
+		message, err := io.ReadAll(errorStream)
+		switch {
+		case err != nil:
+			log.Errorf("k8sPortFwd connection: reading from error stream: %v", err)
+		case len(message) > 0:
+			log.Errorf("k8sPortFwd connection: forwarding port: %v", string(message))
+		}
```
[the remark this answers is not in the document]
flyth: > Shouldn't this code run in a loop?

In theory we could, but we don't really need that information in realtime (the endpoint could handle several port forwarding connections over time, but we just use one), so there's no need to handle that protocol in more detail here. We still need to subscribe and handle that stream, though, otherwise forwarding does not work.

flyth:
> I tested it and I think the connection timeout is too high

I changed the default to 5 seconds and made it configurable with `--connection-timeout`.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -141,7 +158,14 @@ func (r *Runtime) ParamDescs() params.ParamDescs {
 }
 
 func (r *Runtime) GlobalParamDescs() params.ParamDescs {
-	p := params.ParamDescs{}
+	p := params.ParamDescs{
+		{
+			Key:          ParamConnectionTimeout,
+			Description:  "Maximum time to establish a connection to remote target in seconds",
+			DefaultValue: fmt.Sprintf("%d", ConnectTimeout),
+			TypeHint:     params.TypeInt,
```
mauriciovasquezbernal, line 166: nit: TypeUint?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -154,6 +178,20 @@ func (r *Runtime) GlobalParamDescs() params.ParamDescs {
 		}...)
 		return p
 	case ConnectionModeKubernetesProxy:
+		p.Add(params.ParamDescs{
+			{
+				Key:            ParamConnectionMethod,
+				Description:    "Method to use to connect to the gadget service; either tcp or unix-socket (deprecated)",
```
mauriciovasquezbernal, line 184: unix-socket -> unix?

Also, perhaps this is better to avoid to ... to ... to (ditto for ParamGadgetServiceTCPPort)

```suggestion
				Description:    "Method used to connect to the gadget service; either tcp or unix-socket (deprecated)",
```

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -387,7 +440,8 @@ func (r *Runtime) runGadget(gadgetCtx runtime.GadgetContext, target target, allP
 	connCtx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	dialCtx, cancelDial := context.WithTimeout(gadgetCtx.Context(), time.Second*ConnectTimeout)
+	timeout := time.Second * time.Duration(r.globalParams.Get(ParamConnectionTimeout).AsInt())
```
mauriciovasquezbernal, line 443: Can this variable be declared once before L411 and used in both places? (L415 and L444)

mauriciovasquezbernal (approved):
Some minor nits, LGTM. 

I tried to use it with an old IG installation and it was the result:

```
$ ./kubectl-gadget trace open 
K8S.NODE                  K8S.NAMESPACE             K8S.POD                  K8S.CONTAINER            PID        COMM         FD ERR PATH                     
ERRO[0000] k8sPortFwd connection: forwarding port: error forwarding port 8080 to pod 156d861a8f194c7aa67e3961de54a8ee9ddce27971aba126a7831dd687e039ca, uid : failed to execute portforward in network namespace "/var/run/netns/cni-0250c7fb-4b6e-d6be-efaa-b08b8a7a76fd": failed to connect to localhost:8080 inside namespace "156d861a8f194c7aa67e3961de54a8ee9ddce27971aba126a7831dd687e039ca", IPv4: dial tcp4 127.0.0.1:8080: connect: connection refused IPv6 dial tcp6: address localhost: no suitable address found  
ERRO[0001] k8sPortFwd connection: forwarding port: error forwarding port 8080 to pod 156d861a8f194c7aa67e3961de54a8ee9ddce27971aba126a7831dd687e039ca, uid : failed to execute portforward in network namespace "/var/run/netns/cni-0250c7fb-4b6e-d6be-efaa-b08b8a7a76fd": failed to connect to localhost:8080 inside namespace "156d861a8f194c7aa67e3961de54a8ee9ddce27971aba126a7831dd687e039ca", IPv4: dial tcp4 127.0.0.1:8080: connect: connection refused IPv6 dial tcp6: address localhost: no suitable address found  
Error: running gadget: dialing target on node "ubuntu-hirsute": dialing "gadget-vnq76" ("ubuntu-hirsute"): context deadline exceeded
```

Perhaps that error message could be extended with a hint to use `--connection-method=unix`? 
I also we need https://github.com/inspektor-gadget/inspektor-gadget/issues/1795 back, otherwise people could be confused when using different versions.

pkg/runtime/grpc/k8s-portfwd-dialer.go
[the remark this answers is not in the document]
flyth: The `Close()` just closes the sending direction of that stream, while a `Reset()` would close both directions afaiu. Naming is a bit confusing there.

pkg/runtime/grpc/grpc-runtime.go
flyth: `TypeDuration` parses a different format, it expects something like `3s` for 3 seconds (instead of just a `3` for an int). I think for such small durations having to use the `3s` confuses more than it helps, that's why I chose to stay with ints.

pkg/runtime/grpc/grpc-runtime.go
flyth: Hmm, L444 is another function - it could be defined in `runGadget` and then used as parameter to `dialContext` to be reused, but then the other callers of `dialContext` would have to get the timeout value. I'll change it to that.

pkg/runtime/grpc/k8s-portfwd-dialer.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+func (k *k8sPortFwdDialer) SetReadDeadline(t time.Time) error {
+	return nil
+}
+
+func (k *k8sPortFwdDialer) SetWriteDeadline(t time.Time) error {
+	return nil
+}
```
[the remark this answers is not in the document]
flyth: I'll change the `t time.Time` to `_ time.Time` - but I don't think documenting all functions is necessary, as the base type `k8sExecConn` is not exported and the generating func `NewK8SExecConn()` just returns the interface `net.Conn` (so documentation would always point to that anyway).

pkg/runtime/grpc/grpc-runtime.go
mauriciovasquezbernal: btw, I just noticed that we have a `--request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0"` parameter on kubectl-gadget (probably coming from the common kubernetes flags). Do you think these two parameters could be combined?

flyth: Hmm, I don't know. It would probably be okay, but I'd not be fine with the default value of `0`. But then again, I like to have it separated because both things are really only useful in edge cases where it might be actually important to control the values independently.

flyth:
Forgot to say: thanks for the reviews and great feedback! ❤️

PR #2114, opened 2023-10-06
ci: Only release image without CVEs.

eiffel-fl:
Hi.

I marked this PR as RFC as maybe we should be able to bypass this.
On the other hand, this is not a good thing to publish flawed images.

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1375,6 +1375,7 @@ jobs:
       - build-gadgets-examples
       - package-helm-charts
       - check-secrets
+      - scan-gadget-container-images
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (commented):
I think it'll become a bottleneck in our workflow not being able to push those images if there is a known CVE on the base image. I'm ok with having this for the release workflow, there we could wait until the base image is fixed, and if it's taking too long to change it or find another solution.

mauriciovasquezbernal (approved):
I think it's fine. @alban do you have any opinions on this one?

alban (approved):
LGTM

The patch ensures that the step `scan-gadget-container-images` is executed for the release.

I see it has `ignore-unfixed: true`. Hopefully with that setting, we will not get blocked too much for a release.

PR #2117, opened 2023-10-06
go.mod: Remove replace directive for bumblebee

mauriciovasquezbernal:
We don't depend on bumblebee anymore, remove this directive.


alban (approved):
LGTM.

The CI failure is not related to this change.

PR #2121, opened 2023-10-09
ci: Add SBOMs as release artifacts.

eiffel-fl:


burak-ok (approved):
LGTM from my side.

As a follow up we could attach the sbom to the oci image itself: https://oras.land/blog/oras-0.14-and-future/#attach-the-sbom-to-this-image (Didn't read it fully, don't know what we would need to do)

Issue #2115, opened 2023-10-06
Handle array of bytes when filtering columns in containerized gadgets


mqasimsarfraz:
## Description

Currently if we try to filter columns of the type `char []foo` in containerized gadgets we run into following issue:

```
→ sudo -E ./ig run ghcr.io/inspektor-gadget/gadget/trace_open:latest -F comm:kubelet
INFO[0000] Experimental features enabled                
Error: setting filters: invalid filter "comm:kubelet": tried to match "kubelet" on unsupported column "comm"

```

Possibly we need to handle array of bytes here: https://github.com/inspektor-gadget/inspektor-gadget/blob/f75ed8034eb958b83f854ba71d5ea0c7ca59989e/pkg/columns/filter/filter.go#L84

## Impact

Not able to filter columns like `comm`, `fname` etc

## Expected behavior

Be able to filter with every column.

PR #2122, opened 2023-10-09
filter: Allow filtering for '[]char' for containerized gadgets

mqasimsarfraz:
Resolves: #2115

## before the PR
````
$ go run -exec 'sudo -E' ./cmd/ig/ run ghcr.io/inspektor-gadget/gadget/trace_open -F comm:~cat
Error: setting filters: invalid filter "comm:kubelet": tried to match "kubelet" on unsupported column "comm"
````

## after the PR
```
$ go run -exec 'sudo -E' ./cmd/ig/ run ghcr.io/inspektor-gadget/gadget/trace_open -F comm:cat
RUNTIME.CONTAINERNAME                                                              PID     COMM             UID      GID      RET            FNAME                                                                                  
stoic_sutherland                                                                   65767   cat              0        0        -2             /etc/ld.so.cache                                                                       
stoic_sutherland                                                                   65767   cat              0        0        -2             /lib/x86_64-linux-gnu/glibc-hwcaps/x86-64-v3/libm.so.6   
$ go run -exec 'sudo -E' ./cmd/ig/ run ghcr.io/inspektor-gadget/gadget/trace_open -F fname:~/etc/.*
RUNTIME.CONTAINERNAME                                                              PID     COMM             UID      GID      RET            FNAME                                                                                  
stoic_sutherland                                                                   66084   cat              0        0        -2             /etc/ld.so.cache                                                                       
stoic_sutherland                                                                   66084   cat              0        0        3              /etc/resolv.conf 
```

### CODE REVIEW

**Comment by flyth** [comment]
Could you please add a couple of test cases that demonstrate how this is going to be used?
Also, what do you think about adding support for slices as well?

**Comment by mqasimsarfraz** [comment]
> Could you please add a couple of test cases that demonstrate how this is going to be used? Also, what do you think about adding support for slices as well?

Thanks for the suggestions. I have added some tests. Please feel free to have a look again!

File: pkg/columns/filter/filter.go [hunk 1]
```diff
@@ -263,6 +266,40 @@ func getComparisonFuncForComparisonType[OT constraints.Ordered, T any](ct compar
...
+		return func(a *T) bool {
+			return ff(a) <= refValue.(string) != negate
+		}
+	default:
+		return func(a *T) bool {
+			return false
+		}
+	}
+}
+
```
mauriciovasquezbernal, line 302: What about refactor `getComparisonFuncForComparisonType` to receive `ff`, then reuse it in `case reflect.Array, reflect.Slice:` instead of duplicating the logic on this new `getComparisonFuncAsStringField`. 

For sure, easier to understand with some code: https://github.com/inspektor-gadget/inspektor-gadget/commit/887d62873ed798ea4c8b5684b23258ad2fc7edf6

pkg/columns/filter/filter.go
```diff
@@ -143,7 +143,8 @@ func GetFilterFromString[T any](cols columns.ColumnMap[T], filter string) (*Filt
 		fs.value = filterRule
 	}
 
-	if fs.comparisonType == comparisonTypeRegex && column.Kind() != reflect.String {
+	if fs.comparisonType == comparisonTypeRegex &&
+		(column.Kind() != reflect.String && column.Kind() != reflect.Array && column.Kind() != reflect.Slice) {
 		return nil, fmt.Errorf("tried to apply regular expression on non-string column %q", fs.column.Name)
 	}
```
mauriciovasquezbernal, line 149: I know the logic was there before, but can't we lift this limitation and allow using regexp with non-string columns too? In that case, we'll need convert the field to string, using `GetFieldAsString`, before matching the regexp.

mauriciovasquezbernal (commented):
I tested it and works as expected. However I think we need some more discussion on this one, some questions I have:
1. What happens for fields like `u64 foo[4]`? 
2. Does it fix https://github.com/inspektor-gadget/inspektor-gadget/issues/2243?
3. Can we enable regexp support for all types of columns?

pkg/columns/filter/filter.go
mqasimsarfraz: Done. Although I am thinking if we should still have checks for `structs` ?

pkg/columns/columns.go
```diff
@@ -615,7 +615,7 @@ func GetFieldAsStringExt[T any](column ColumnInternals, floatFormat byte, floatP
 		return func(entry *T) string {
 			return "TODO"
 		}
-	case reflect.String:
+	case reflect.String, reflect.Slice:
```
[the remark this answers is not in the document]
mqasimsarfraz: Nice catch! I reworked the logic. Also, I was considering only to handle `[]byte` only but we can also handle `[]string` by adding following logic:

```
case reflect.Slice:
...
...
  if column.(*Column[T]).Type().Elem().Kind() == reflect.String {
	  ff := GetFieldFunc[[]string, T](column)
	  return func(entry *T) string {
		  return strings.Join(ff(entry), ", ")
	  }
  }
```

What do you think?

pkg/columns/filter/filter.go
mauriciovasquezbernal: Do you mean in case there is a struct within a struct in the event type? I think it makes sense, however I'll ping @flyth as he has the ideas more clear on this module.

mauriciovasquezbernal (approved):
LGTM. I tested it with the trace gadget as stated in the PR description and it works fine.

mqasimsarfraz:
Thanks for the reviews. I will merge this one and we can handle open conversations in follow up PR.

PR #2126, opened 2023-10-10
Dockerfiles: Upgrade packages on BASE_IMAGE.

eiffel-fl:
This avoids releasing an image with CVE.

### DIFF
--- a/Dockerfiles/gadget.Dockerfile
+++ b/Dockerfiles/gadget.Dockerfile
@@ -41,9 +41,7 @@ RUN set -ex; \
 		yum install -y $PACKAGES; \
 	elif command -v apt-get; then \
 		apt-get update && \
-		apt-get install -y $PACKAGES && \
-		apt-get clean && \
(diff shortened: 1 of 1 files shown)


mqasimsarfraz:
It will definitely help avoid CVEs in the base image. The only tricky bit I would like to understand is how [docker cache](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/workflows/inspektor-gadget.yml#L337) in the CI will behave. Essentially everything people experienced here: https://github.com/moby/moby/issues/3313.

mqasimsarfraz:
> I do not think we can have fresh (i.e. patched) and cache at the same time.

Agreed. I am not sure how much time are we getting out of cache anyways. Probably not a lot!

mqasimsarfraz (approved):
LGTM! Thanks for handling feedback. Please wait for another review before merge.

mauriciovasquezbernal (commented):
I'm not sure why it's something we should take care of. Shouldn't it be done by the base image publisher?

PR #2129, opened 2023-10-10
service: Reduce critical section size.

eiffel-fl:
The mutex should only be hold to write and read seq value, not to deal with writing to channel.


flyth:
It's important that it deals with writing to the channel inside that section as well, as otherwise the order (given by seq) is not guaranteed anymore. Btw, the writing to the channel is (due to the select/default statements) non-blocking.

PR #1472, opened 2023-03-24
integration: Allow using PR image for dnstester

mqasimsarfraz:
This PR introduces a way to test changes to dnstester in a PR and on a fork. The idea is to save the dnstester image tag if changes are detected and pass it as a flag to integration tests. The push is disabled by default on forks and for external contributors by using `PUSH_HELPERS` secret. 

## Testing done

I tested the changes on a fork after setting `PUSH_HELPERS = ENABLE_PUSH_HELPERS` secret as:

- [Without custom PR image](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/4511808818) - [logs](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/4511808818/jobs/7944477217#step:6:1741)
- [With custom PR image](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/4512232476) - [logs](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/4512232476/jobs/7945491138#step:6:132)

## Todo
- [ ] Create `PUSH_HELPERS` variable in the repository

PR #2135, opened 2023-10-11
ci: Publish ebpf builder helper image

mqasimsarfraz:
Fixes: #1420

It uses the same idea of #1472 and will only:
- Build images if changes are detected
- Publish images if variable `PUSH_HELPERS = ENABLE_PUSH_HELPERS` is defined (it helps avoid issues with external contributors PR)
- Publish images for each release

## Testing done

- Fork: A [test commit](https://github.com/mqasimsarfraz/inspektor-gadget/commit/33523afbb00c11be79102946f0d876bd033b56c8) result in helper [image being built](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/6482505619/job/17602162537) and used [here](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/6482505619/job/17602170043#step:4:1).
- Release: A test release (`v0.10.0`) publish the tags [here](https://github.com/mqasimsarfraz/inspektor-gadget/pkgs/container/ebpf-builder/versions).

## Todo

- [x] Claim ownership of https://github.com/orgs/inspektor-gadget/packages/container/package/ebpf-builder since it belongs to Inspektor Gadget `org` (or someone who created it) not `repo`. 

## Follow up
- [ ] Add support for arm64
- [ ]  Use the published release image [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/image/build.go#L45). (Should we? Separate PR?)

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -681,18 +681,23 @@ jobs:
         # https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-type-check-expression
         filters: |
           pattern: ${{ toJson(matrix.image.filter-patterns) }}
+    - name: Check if we should build helpers
+      id: check-build-helpers
+      if: steps.filter.outputs.pattern == 'true' || startsWith(github.ref_name, 'v')
+      run: |
+          echo "build=true" >> $GITHUB_OUTPUT
     - name: Set up Docker Buildx
-      if: steps.filter.outputs.pattern == 'true'
+      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/setup-buildx-action@v3
     - name: Login to Container Registry
-      if: steps.filter.outputs.pattern == 'true'
+      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/login-action@v3
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
     - name: Set container repository and determine image tag
-      if: steps.filter.outputs.pattern == 'true'
+      if: steps.check-build-helpers.outputs.build == 'true'
       id: set-repo-determine-image-tag
       uses: ./.github/actions/set-container-repo-and-determine-image-tag
       with:
@@ -700,7 +705,7 @@ jobs:
         container-image: ${{ github.repository_owner }}/${{ matrix.image.name }}
     - name: Build ${{ matrix.image.name }} image
       id: build-image
-      if: steps.filter.outputs.pattern == 'true'
+      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/build-push-action@v5
       with:
         context: ${{ matrix.image.context }}
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -5,6 +5,7 @@ env:
   GO_VERSION: 1.19.6
   AZURE_AKS_CLUSTER_PREFIX: ig-ci-aks-
   DEFAULT_DNSTESTER_IMAGE: ghcr.io/inspektor-gadget/dnstester:latest
+  DEFAULT_EBPF_BUILDER_IMAGE: ghcr.io/inspektor-gadget/ebpf-builder:latest
 concurrency:
   group: ${{ github.ref }}
   # We do not want to cancel job in progress on main to be sure to catch new
@@ -124,7 +125,8 @@ jobs:
 
   ebpf-objects-checks:
     name: eBPF Object checks
-    # level: 0
+    # level: 1
+    needs: build-helper-images
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -140,7 +142,7 @@ jobs:
           fi
       - name: Detect changes for ebpf objects
         run: |
-          make ebpf-objects
+          make EBPF_BUILDER=${{ needs.build-helper-images.outputs.ebpf_builder_image }} ebpf-objects
           changes="$(git status --porcelain)"
           if [ -n "$changes" ] ; then
             echo "$changes"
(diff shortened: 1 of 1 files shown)


.github/workflows/inspektor-gadget.yml
```diff
@@ -657,9 +660,15 @@ jobs:
...
           - name: "dnstester"
-            dockerfile-dir: "tools/dnstester"
+            context: "tools/dnstester"
+            dockerfile: "tools/dnstester/Dockerfile"
             platform: "linux/amd64,linux/arm64"
             filter-pattern: "tools/dnstester/*"
+          - name: "ebpf-builder"
+            context: "/home/runner/work/inspektor-gadget/inspektor-gadget"
+            dockerfile: "Dockerfiles/ebpf-builder.Dockerfile"
+            platform: "linux/amd64"
```
alban, line 670: Does it mean `ig image build` would not work on linux/arm64?

alban (commented):
> Claim ownership of https://github.com/orgs/inspektor-gadget/packages/container/package/ebpf-builder since it belongs to Inspektor Gadget org (or someone who created it) not repo.

I did it manually. But you can add the label in `Dockerfiles/ebpf-builder.Dockerfile`:

```
LABEL org.opencontainers.image.source https://github.com/inspektor-gadget/inspektor-gadget
```

(although the `LABEL` keyword in the Dockerfile might not work for multi-arch images)

> Use the published release image [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/image/build.go#L45). (Should we? Separate PR?)

I think we should with the following rule:
- If VERSION is a release, use that tag by default
- If not, use `latest` by default

What do you think?

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: I saw the [existing image](https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/ebpf-builder) was also only built for `linux/amd64` so I just used that but yes that would mean `ig image build` won't work on `linux/arm64` with builder image. Do you know if it fine to build it for `linux/arm64` ?

mauriciovasquezbernal:
> I think we should with the following rule:
> 
>     * If VERSION is a release, use that tag by default
> 
>     * If not, use `latest` by default

I think it if the image is modified by a PR, then the built image for that specific PR should be used, otherwise it'll be impossible to implement PRs that modify the image and the gadgets (as https://github.com/inspektor-gadget/inspektor-gadget/pull/2107) without having a "REMOVE BEFORE MERGE" commit pointing the image to something else.

.github/workflows/inspektor-gadget.yml
```diff
@@ -677,25 +677,25 @@ jobs:
         filters: |
           pattern: ${{ matrix.image.filter-pattern }}
     - name: Set up Docker Buildx
-      if: steps.filter.outputs.pattern == 'true'
+      if: steps.filter.outputs.pattern == 'true' || startsWith(github.ref, 'refs/tags/v')
```
mauriciovasquezbernal, line 680: Would it make sense to define a new `should-skip` step with this logic and then other steps will only check a single output from the `should-skip` step?

mauriciovasquezbernal (changes requested):
Looks good mostly, only blocking comment on missing rules on pattern.

IMO we can handle arm64 in another PR. 

Btw, I rebased https://github.com/inspektor-gadget/inspektor-gadget/pulls on top of it so we can check with a real example if it's working fine :D

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: I tried building it for `arm64` and running into following [error](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/6493130779/job/17633458645#step:8:1181):
```
#9 33.43 E: Package 'gcc-multilib' has no installation candidate
#9 ERROR: process "/dev/.buildkit_qemu_emulator /bin/sh -c apt-get update \t&& apt-get install -y gcc-multilib lsb-release wget software-properties-common gnupg" did not complete successfully: exit code: 100
```

.github/workflows/inspektor-gadget.yml
```diff
@@ -657,9 +660,15 @@ jobs:
...
-            dockerfile-dir: "tools/dnstester"
+            context: "tools/dnstester"
+            dockerfile: "tools/dnstester/Dockerfile"
             platform: "linux/amd64,linux/arm64"
             filter-pattern: "tools/dnstester/*"
+          - name: "ebpf-builder"
+            context: "/home/runner/work/inspektor-gadget/inspektor-gadget"
+            dockerfile: "Dockerfiles/ebpf-builder.Dockerfile"
+            platform: "linux/amd64"
+            filter-pattern: "Dockerfiles/ebpf-builder.Dockerfile"
```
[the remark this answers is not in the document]
mqasimsarfraz: Done.

mqasimsarfraz:
> Btw, I rebased https://github.com/inspektor-gadget/inspektor-gadget/pulls on top of it so we can check with a real example if it's working fine :D

I see it failed because of missing patterns, I updated the list and now it [looking better](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/6493511261) with [PR on fork](https://github.com/mqasimsarfraz/inspektor-gadget/pull/37)!

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: I see, I opened a separate PR (it will also test if image is built fine for PR) for this, https://github.com/inspektor-gadget/inspektor-gadget/pull/2139 so we can track `arm64` alongside.

mqasimsarfraz: It failed while building/pushing gadgets: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/6495111185/job/17642715987

mqasimsarfraz:
I merged this one and we can track `arm64` in https://github.com/inspektor-gadget/inspektor-gadget/pull/2139

PR #2141, opened 2023-10-12
pkg/columns/json: Implement FormatEntries()

mauriciovasquezbernal:
075f2b1e4fd9 ("pkg/columns: add json formatter") implemented a function
    to marshal a single entry to json using the columns information. This
    commit introduces a new function to marshal a slice of entries to a
    json array.
---

This was initially proposed in https://github.com/inspektor-gadget/inspektor-gadget/pull/1866, but since merging that PR is taking a bit long, I decided to extract some commits in a separated PR to help with the review process.

### DIFF
--- a/pkg/columns/formatter/json/json.go
+++ b/pkg/columns/formatter/json/json.go
@@ -35,7 +35,7 @@ type column[T any] struct {
 type Formatter[T any] struct {
 	options *Options
 	columns []*column[T]
-	printer func(entry *T, buf *encodeState)
+	printer func(buf *encodeState, entry *T, indent string)
 }
 
 // NewFormatter returns a Formatter that will turn entries of type T into JSON representation
@@ -131,12 +131,12 @@ func NewFormatter[T any](cols columns.ColumnMap[T], options ...Option) *Formatte
 	return tf
 }
 
-func (f *Formatter[T]) getPrinter(level int, cols []*column[T]) func(entry *T, buf *encodeState) {
-	indent := strings.Repeat("  ", level+1)
+func (f *Formatter[T]) getPrinter(level int, cols []*column[T]) func(buf *encodeState, entry *T, indent string) {
+	levelIndent := []byte("  ")
 	colsWithParent := map[string][]*column[T]{}
 	colsWithoutParent := []*column[T]{}
 
-	funcs := make([]func(entry *T, buf *encodeState), 0)
+	funcs := make([]func(buf *encodeState, entry *T, indent string), 0)
 
 	for _, col := range cols {
 		// Save columns which have a parent in its name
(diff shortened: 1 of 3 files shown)


pkg/columns/formatter/json/json.go
```diff
@@ -35,7 +35,7 @@ type column[T any] struct {
 type Formatter[T any] struct {
 	options *Options
 	columns []*column[T]
-	printer func(entry *T, buf *encodeState)
+	printer func(indent string, entry *T, buf *encodeState)
```
flyth, line 38: nit: `indent` feels like the least important parameter, so maybe it'd be better placed at the end. Seeing this, I'd also suggest to put `buf` first, as most times the target one writes to is the first parameter - but feel free to ignore.

pkg/columns/formatter/json/json.go
```diff
@@ -157,18 +156,18 @@ func (f *Formatter[T]) getPrinter(level int, cols []*column[T]) func(entry *T, b
 	// Better name?
 	handlePrefix := func() {
 		if !first {
-			funcs = append(funcs, func(entry *T, buf *encodeState) {
+			funcs = append(funcs, func(indent string, entry *T, buf *encodeState) {
 				buf.WriteByte(',')
 				if f.options.prettyPrint {
-					buf.WriteString("\n" + indent)
+					buf.WriteString("\n" + indent + "  ")
```
flyth, line 162: Fyi: in isolated tests, calling this separately (`WriteByte('\n')`, `WriteString(indent)`, `Write(LevelIndent)`, where `LevelIndent = []byte("  ")` performed 3x better than the string concatenation; however, in the full scenario this didn't really have weight.

pkg/columns/formatter/json/json.go
```diff
@@ -208,34 +207,73 @@ func (f *Formatter[T]) getPrinter(level int, cols []*column[T]) func(entry *T, b
...
 		for _, f := range funcs {
-			f(entry, buf)
+			f(indent, entry, buf)
 		}
 	}
 }
 
+func (f *Formatter[T]) formatEntry(indent string, entry *T, buf *encodeState) {
+	if entry == nil {
+		buf.WriteString(indent + "{}")
```
flyth, line 224: This is wrong, it should be `null` (without quotation marks); was also wrong for `FormatEntry()` before, where it was an empty string instead.

flyth (commented):
Looks good, I just found one issue regarding nil values.

pkg/columns/formatter/json/json.go
```diff
@@ -208,34 +207,79 @@ func (f *Formatter[T]) getPrinter(level int, cols []*column[T]) func(entry *T, b
...
+				buf.WriteString(", ")
+			}
+		}
+		buf.WriteByte(']')
+
+		return buf.String()
 	}
-	buf.WriteByte('}')
+
+	buf.WriteString("[\n")
```
burak-ok, line 275: Maybe instead of a return in the `if` we can have a
```golang
} else {
 ...
}
return buf.String()
```

Normally I am also the person for early returns but this just looks in my head perfect for a if-else :sweat_smile:

burak-ok (approved):
LGTM

pkg/columns/formatter/json/json.go
flyth: That didn't make any difference for me in any test I did. The difference was only in concatenating the string before writing.

flyth (approved):
LGTM, thanks!

pkg/columns/formatter/json/json.go
mauriciovasquezbernal: Right, makes sense on this case.

pkg/columns/formatter/json/json.go
mauriciovasquezbernal: I updated the code to avoid the string concatenation. It was not difficult so why not?

PR #2144, opened 2023-10-12
run/types: Remove unused field

mauriciovasquezbernal:
Remove this field to avoid confusion on the code.


burak-ok (approved):
Thanks for removing it

PR #2147, opened 2023-10-13
treewide: Update to golang 1.20.

eiffel-fl:
cilium/ebpf release v0.12.0 is only compatible with golang 1.20 [1]. Let's bump to be able to use this release.

[1]: https://github.com/cilium/ebpf/releases/tag/v0.12.0

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -2,7 +2,7 @@ name: Inspektor Gadget CI
 env:
   REGISTRY: ghcr.io
   CONTAINER_REPO: ${{ github.repository }}
-  GO_VERSION: 1.19.6
+  GO_VERSION: 1.20.10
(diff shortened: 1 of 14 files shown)


mauriciovasquezbernal (approved):
LGTM from code inspection. 

Regarding the linter, perhaps the version needs to be updated?

pkg/container-collection/container-collection_test.go
```diff
@@ -102,12 +102,12 @@ func BenchmarkLookupContainerByNetns(b *testing.B) {
 		})
 	}
 
-	rand.Seed(time.Now().UnixNano())
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
flyth, line 105: nit: define globally and reuse here and on L78 like in the other tests?

flyth (approved):
Looks good to me from code inspection as well. Thanks!

PR #2152, opened 2023-10-17
pkg/params: Implement some extra functions

mauriciovasquezbernal:
Implement some functionality needed in https://github.com/inspektor-gadget/inspektor-gadget/pull/2119

- pkg/params: Add IsSet() method
- pkg/params: Add AsAny()
- pkg/params: Implement float32 and float64 validators

### DIFF
--- a/pkg/params/params.go
+++ b/pkg/params/params.go
@@ -352,6 +352,49 @@ func (p *Param) Set(val string) error {
 	return nil
 }
 
+// AsAny returns the value of the parameter according to its type hint. If there is not any type
+// hint, it returns the value as string.
+func (p *Param) AsAny() any {
+	switch p.TypeHint {
+	case TypeBool:
+		return p.AsBool()
+	case TypeString:
+		return p.AsString()
+	case TypeBytes:
+		return p.AsBytes()
+	case TypeInt:
+		return p.AsInt()
+	case TypeInt8:
+		return p.AsInt8()
+	case TypeInt16:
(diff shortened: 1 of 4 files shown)


pkg/params/params.go
```diff
@@ -352,6 +352,53 @@ func (p *Param) Set(val string) error {
 	return nil
 }
 
+func (p *Param) IsSet() bool {
+	return p.DefaultValue != p.value
+}
```
flyth, line 357: How is `IsSet()` used? Maybe it's better to add a boolean that is set when `Set()` is called and check for that - and then add another func `IsDefault() bool` that does this comparison (negated).

flyth (commented):
Looks good from code inspection, only `IsSet()` sounds a bit misleading, IMHO.

pkg/params/params.go
mauriciovasquezbernal: TBH I don't remember what was the reason I chose this implementation. I think it was related to `Set()` being called in by the service (while decoding params from the client), and that causing some issues with the default value of some ebpf params. I tried to add a boolean to param, tested with https://github.com/inspektor-gadget/inspektor-gadget/pull/2119  and worked fine, so I'll stick to that solution. I also added the `IsDefault()` method as it could be useful in other situations.

flyth: Awesome, thanks for adding that!

flyth (approved):

PR #2156, opened 2023-10-18
Cleanup metadata file

mauriciovasquezbernal:
This PR contains some cleanups for https://github.com/inspektor-gadget/inspektor-gadget/pull/2107. Please check each individual commit to get more details.


pkg/gadgets/run/tracer/helpers.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+
+	"github.com/cilium/ebpf/btf"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/run/types"
+)
+
+// getAnyMapElem returns the first element of a map. If the map is empty, it returns nil, nil.
+func getAnyMapElem[K comparable, V any](m map[K]V) (*K, *V) {
```
burak-ok, line 26: I still feel uneasy about this function. AFAIU there is no guarantee that in the following example
```golang
k1, v1 = getAnyMapElem(myMap)
k2, v2 = getAnyMapElem(myMap)
```
`k1 == k2 && v1==v2` is true

I know we are checking somewhere in the code that there is only a single tracer map  - and that is what we currently support. So its okay

testdata/populate_metadata_tracer_add_missing_field.bpf.c
burak-ok, line 1: Is there something missing in this file? If its only the include isn't this an exact copy of that file?

include/gadget/macros.h
burak-ok, line 1: I had the same idea for `GADGET_PARAM` so that eBPF developers can just write something like 
 `GADGET_PARAM(param_name, c_type, init_value)`. That macro then also creates the variable correctly with `const volatile c_type param_name = init_value;` etc

burak-ok (commented):
Just did a static analysis - Didn't run it (yet)

testdata/populate_metadata_tracer_add_missing_field.bpf.c
mauriciovasquezbernal: I knew somebody was to comment on this. When changing the names of the files, I found out there was a 1:1 mapping for all cases but for this one, so I decided to use this "trick" to keep always that 1:1 mapping. I added a comment to hopefully make this more clear.

pkg/gadgets/run/tracer/helpers.go
mauriciovasquezbernal: I agree that's not totally clear, but TBH I don't know how to improve it right now. Hopefully the limitation of a having a single tracer will be removed soon and we don't need to use it anymore.

PR #2162, opened 2023-10-19
Add `ig image remove` command

eiffel-fl:
Hi.

This PR adds `ig image remove` command.
It depends on work in progress in oras-project/oras-go#614.

Best regards.

### DIFF
--- a/docs/core-concepts/images.md
+++ b/docs/core-concepts/images.md
@@ -216,6 +216,29 @@ ghcr.io/inspektor-gadget/trace_exec                   latest                    
 ghcr.io/inspektor-gadget/trace_open                   latest                                                3a23c1f08a8b
 ```
 
+#### `remove`
+
+Remove the given gadget image from the host.
+
+```bash
+$ sudo ig image remove -h
+INFO[0000] Experimental features enabled
+Remove local gadget image
+
+Usage:
+  ig image remove IMAGE [flags]
+
+Flags:
+  -h, --help   help for remove
+
+```
+
+```bash
+$ sudo ig image remove gadget
+INFO[0000] Experimental features enabled
+Successfully removed gadget
+```
+
 #### `pull`
 
 Pull the specified image from a remote registry.


mqasimsarfraz:
Thanks for looking into this, I have been doing `rm -rf  /var/lib/ig/oci-store` to clean up all images so this would be much better. I am wondering if it should be `image delete` vs `image rm` ?

mqasimsarfraz:
This is helpful for https://github.com/inspektor-gadget/inspektor-gadget/pull/2171 as well. Let me test it! 

> I will remove the hold tag.

Does that mean we can merge it or we still have to wait to till upstream PRs are merged?

pkg/oci/oci.go
```diff
@@ -305,6 +306,48 @@ func ListGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
...
+	for _, descriptor := range descriptors {
+		deleteTree(ctx, store, descriptor)
+	}
+
+	predecessors, err := store.Predecessors(ctx, root)
+	if err != nil {
+		return fmt.Errorf("getting image predecessors: %w", err)
+	}
+
+	if len(predecessors) > 1 {
```
burak-ok, line 324: I am not so sure how `.Predecessors()` work internally. But if it is like in a graph then I think that this check has to be the first thing in the function.

Example (All letters are descriptors here):
my-image:v0.3:     `A->B->C->D`
my-image:v0.2:     `Z->B->C->D`

Predecessors(B): 2
Predecessors(C): 1
Predecessors(D): 1

C and D would be deleted if we call `deleteTree(Z)`

Am I thinking right about this?

cmd/common/image/remove.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
+)
+
+func NewRemoveCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "remove IMAGE",
+		Short:        "Remove host gadget image",
```
burak-ok, line 29: `Remove local gadget image` maybe?

burak-ok:
> > > I will remove the hold tag.
> > 
> > 
> > Does that mean we can merge it or we still have to wait till upstream PRs are merged?
> 
> It depends. If we want to have the remove command in the next release, then better to merge it now. Otherwise, we can wait. By the way the upstream PR was rebased and should be merged soon. On another hand, merging it soon here we can still modify the `go.*` to use upstream rather than fork once it is merged.

There was no activity since 2 weeks on the upstream PR. Should we merge this now?

go.mod
mauriciovasquezbernal, line 1: Fix commit title, it's not for prototyping anymore.

pkg/oci/oci.go
```diff
@@ -294,6 +295,48 @@ func ListGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
...
+	}
+
+	for _, descriptor := range descriptors {
+		deleteTree(ctx, store, descriptor)
+	}
+
+	return store.Delete(ctx, root)
+}
+
+// DeleteGadgetImage remove the given image.
```
mauriciovasquezbernal, line 320: remove -> removes

pkg/oci/oci.go
```diff
@@ -294,6 +295,48 @@ func ListGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
...
+		return nil
+	}
+
+	descriptors, err := content.Successors(ctx, store, root)
+	if err != nil {
+		return fmt.Errorf("getting image successors: %w", err)
+	}
+
+	for _, descriptor := range descriptors {
+		deleteTree(ctx, store, descriptor)
```
mauriciovasquezbernal, line 314: Check returned error?

pkg/oci/oci.go
```diff
@@ -294,6 +295,48 @@ func ListGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
...
 
+func deleteTree(ctx context.Context, store *oci.Store, root ocispec.Descriptor) error {
+	predecessors, err := store.Predecessors(ctx, root)
+	if err != nil {
+		return fmt.Errorf("getting image predecessors: %w", err)
+	}
+
+	if len(predecessors) > 1 {
+		return nil
+	}
```
mauriciovasquezbernal, line 306: Why `> 1`? Because the current tree we're deleting is counted there?

mauriciovasquezbernal (changes requested):
I think there is something off, if there are two tags pointing to the same digest, deleting one of them causes the other one to be removed:

```bash
# there are not images at the beginning 
$ sudo -E ig image list
REPOSITORY                                                               TAG                                                                      DIGEST      

# build an image
$ sudo -E ig image build -t mygadget .  --local
Successfully built docker.io/library/mygadget:latest@sha256:0b34e37d953f64b800ccbb547481e47f5f517cf9bf0aa290ab4f4760c9bb92fa

$ sudo -E ig image list
REPOSITORY                                                               TAG                                                                      DIGEST      
docker.io/library/mygadget                                               latest                                                                   0b34e37d953f

# tag an image with the above 
$ sudo -E ig image tag mygadget foo
Successfully tagged with docker.io/library/foo:latest@sha256:0b34e37d953f64b800ccbb547481e47f5f517cf9bf0aa290ab4f4760c9bb92fa

$ sudo -E ig image list
REPOSITORY                                                               TAG                                                                      DIGEST      
docker.io/library/foo                                                    latest                                                                   0b34e37d953f
docker.io/library/mygadget                                               latest                                                                   0b34e37d953f

# remove the first image 
$ sudo -E ig image remove mygadget:latest

# second one is gone as well
$ sudo -E ig image list
REPOSITORY                                                               TAG                                                                      DIGEST    
```

go.mod
mauriciovasquezbernal: Isn't https://github.com/oras-project/oras-go/pull/614 already merged?

pkg/oci/oci.go
```diff
@@ -272,26 +267,98 @@ func ListGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
...
+	if err != nil {
+		return fmt.Errorf("listing images: %w", err)
+	}
+
+	digest := descriptor.Digest.String()
+	for _, img := range images {
+		imgFullName := fmt.Sprintf("%s:%s", img.Repository, img.Tag)
+		if img.Digest == digest && imgFullName != fullName {
+			return ociStore.Untag(ctx, fullName)
+		}
```
mauriciovasquezbernal, line 359: Would you mind adding a comment indicating what's the reason for this check here? I think it's not so easy to understand. Something like:

`if there are multiple images pointing to the same digest, just untag the image instead of removing the whole tree`

mauriciovasquezbernal (approved):
I tested and it works fine now. 

I'm wondering if we should follow an garbage-collection approach for this too as suggested in https://github.com/oras-project/oras-go/pull/645#issuecomment-1821155544?, deleting an image could be something like 


```go
untag()
collectGarbage()
```

Would you mind adding an entry to https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/images.md?

I'm approving it now as does its job. It's up to you to merge or wait for the upstream PRs to be ready. 

We're missing the tests for this (and other commands), we can do them in https://github.com/inspektor-gadget/inspektor-gadget/pull/2171.

cmd/common/image/remove.go
```diff
@@ -0,0 +1,44 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			image := args[0]
+
+			err := oci.DeleteGadgetImage(context.TODO(), image)
+			if err != nil {
+				return fmt.Errorf("removing gadget image: %w", err)
+			}
+			return nil
+		},
+	}
```
mauriciovasquezbernal, line 41: Please use `utils.MarkExperimental(cmd)` as done in https://github.com/inspektor-gadget/inspektor-gadget/pull/2248

cmd/common/image/remove.go
```diff
@@ -0,0 +1,44 @@
...
+		SilenceUsage: true,
+		Args:         cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			image := args[0]
+
+			err := oci.DeleteGadgetImage(context.TODO(), image)
+			if err != nil {
+				return fmt.Errorf("removing gadget image: %w", err)
+			}
+			return nil
```
mauriciovasquezbernal, line 39: What about printing a message when the command works fine as done in https://github.com/inspektor-gadget/inspektor-gadget/blob/17225b68c17ae58dcff76da5dc82bd38b379abbe/cmd/common/image/tag.go#L41 ?

I find it a bit weird to run it and not get any output:

```bash
$ sudo -E ig image remove docker.io/library/foo
```

mauriciovasquezbernal (approved):
Besides two minor comments, LGTM!

PR #2163, opened 2023-10-19
Extend GADGET_TRACE_MAP() and fix TODO

mauriciovasquezbernal:
This PR extends the GADGET_TRACE_MAP() macro to provide more information and by using it fix a TODO in the code.

- gadgets: Extend GADGET_TRACE_MAP()
- gadgets: Dot not rely on BTF map type


pkg/gadgets/run/types/metadata.go
```diff
@@ -291,37 +283,37 @@ func getColumnSize(typ btf.Type) uint {
 }
 
 func (m *GadgetMetadata) populateTracers(spec *ebpf.CollectionSpec) error {
-	traceMap := getTracerMapFromeBPF(spec)
-	if traceMap == nil {
-		log.Debug("No trace map found")
+	traceInfo, err := getTracerInfo(spec)
```
blanquicet, line 286: In most of this function it was used `trace` instead of `tracer` for the info, map, struct, etc:

```suggestion
	tracerInfo, err := getTracerInfo(spec)
```

testdata/populate_metadata_1_tracer_1_struct_from_scratch.bpf.c
```diff
@@ -13,15 +13,12 @@ struct event {
...
 };
 
-// we need this to make sure the compiler doesn't remove our struct
-const struct event *unusedevent __attribute__((unused));
-
 struct {
 	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
 	__uint(key_size, sizeof(u32));
-	__type(value, struct event);
+	__uint(value_size, sizeof(__u32));
```
blanquicet, line 19: Sometimes it is `__u32` and some others `u32`. Is there a specific reason for this?

blanquicet (commented):
I only did a code inspection of the use of this new gadget-tracer-info during the validation and population of tracers.

pkg/gadgets/run/types/metadata.go
```diff
@@ -348,11 +340,30 @@ func getGadgetIdentByPrefix(spec *ebpf.CollectionSpec, prefix string) string {
...
-	mapName := getGadgetIdentByPrefix(spec, traceMapPrefix)
-	return spec.Maps[mapName]
+type tracerInfo struct {
+	name      string
+	mapName   string
+	eventType string
+}
+
+// getTracerInfo returns the tracer info generated with GADGET_TRACER().
+// If there are multiple annotations only the first one is returned.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Which documentation do you mean?

pkg/gadgets/run/tracer/helpers.go
```diff
@@ -44,13 +44,9 @@ func getEventTypeBTF(progContent []byte, metadata *types.GadgetMetadata) (*btf.S
 			return nil, fmt.Errorf("BPF map %q not found", tracer.MapName)
 		}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Good catch. It's already done in `handleTracers`.

testdata/populate_metadata_1_tracer_1_struct_from_scratch.bpf.c
mauriciovasquezbernal: There is not any reason. We're already using a mix of them in other places and I copy & pasted the one with __u32. Changed it to be u32 to keep it uniform.

alban: In this context, it is fine to use either. But in networking programs, we don't include `vmlinux.h` so `u32` is not defined and we can only use `__u32`.

pkg/gadgets/run/types/metadata.go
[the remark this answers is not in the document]
blanquicet: `docs/devel/hello-world-gadget.md`?

pkg/gadgets/run/types/metadata.go
blanquicet: `traceMap` and `traceMapStruct` also need to be fixed.

blanquicet (approved):
LGTM!

mauriciovasquezbernal:
@alban @flyth do you have any comments about this PR?

flyth (approved):
I really liked the simplicity before, but I understand that replacing the map value type is too hacky - so this is fine for me.

Regarding `GADGET_TRACER()` - maybe we could name it `GADGET_REGISTER_TRACER()` to highlight that it does something to existing things rather than defining everything on its own. I'm fine either way, though.

testdata/validate_metadata1.bpf.c
```diff
@@ -57,7 +47,7 @@ struct {
 	.type = BPF_MAP_TYPE_PERF_EVENT_ARRAY,
 	.max_entries = 4,
 	.key_size = sizeof(int),
-	.value_size = sizeof(struct event),
+	.value_size = sizeof(int),
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I switched to `__32` in many places, but I still would like to take another look to understand which type we should be using in all the places.

Issue #2167, opened 2023-10-20
image-based gadgets: Operators are always enabled for all gadgets


mauriciovasquezbernal:
The current logic enables all operators without taking into account the actual gadget being run. This is not efficient at all as those operators consume resources that aren't useful.

For instance, running the DNS gadget causes the localmanager operator to create a mount ns set map that is not used

```bash
$ sudo -E ig run ghcr.io/inspektor-gadget/gadget/trace_dns:latest
...

# in another terminal

$ sudo bpftool map list | grep mntnsset -A 2
1838: hash  name mntnsset_a1f387  flags 0x0
	key 8B  value 4B  max_entries 1024  memlock 16384B
	pids ig(379769)
```

From a UX point of view it's also bad, considering the hello-world-gadget proposed in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/hello-world-gadget.md before adding the filtering by mount namespace:

The gadget prints some runtime columns that are always empty
```
$ sudo -E ig run mygadget:latest
INFO[0000] Experimental features enabled
RUNTIME.CONTAINERNAME                                                          PID
WARN[0000] The gadget doesn't provide metadata
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1113
                                                                               1219
                                                                               220121
```

and provides some filtering options like `-c, --containername string   Show only data from containers with that name (default "")` and `--host                   Show data from both the host and containers` that don't have any effect.

### Ideal solution

Only the relevant operators should be enabled their flags and columns should be shown to users.

Issue #2131, opened 2023-10-10
[RFE] Update GadgetInfo() rpc to avoid handling BTF/Wasm client-side


alban:
## Current situation

I would like to revisit the design decision in this commit:

https://github.com/inspektor-gadget/inspektor-gadget/pull/2015/commits/8b513e713340f35f9b301af6914542d3f6f8abf1

Currently, in the client/server architecture, the client needs to get the columns of a gadget. This is done with the `GetGadgetInfo()` rpc call. Then, the server returns the following struct:
```
type GadgetInfo struct {
	GadgetDefinition *GadgetDefinition
	ProgContent      []byte
}
```

## Impact

`ProgContent` includes the BTF information and the client needs to parse this in order to find the columns. Some special care needs to be taken client-side for the L3Endpoints and L4Endpoints columns.

## Ideal future situation

I think the client should not parse BTF at all.

## Implementation options

Instead, the client should receive `GadgetDefinition` as today + `ComputedColumnsAttrs []columns.Attributes`. The server would read gadget.yaml and the BTF to compute the columns.

## Additional information

When implementing support for Wasm modules, new virtual columns could be added programmatically. I don't want the client to have to execute Wasm code. But the server could execute Wasm code to compute the `ComputedColumnsAttrs` to return to the client.

In the PoC I did in https://github.com/alban/inspektor-gadget/tree/alban_asg2023, I needed to add some special type `typedef __u8 wasm_string_t[MAX_DNS_NAME];` so that the client could notice the type `wasm_string_t` and create a new column. But I would like to create new columns from Wasm server-side without having to add special types in BTF. Therefore, this issue would be a prerequisite for implementing virtual columns from Wasm.

alban:
I think the difficulty is that we are trying to make the json marshalling/unmarshalling work on generic types which are not known at build-time. I suggest the following alternative with the following characteristics:

* Can be marshalled & unmarshalled without knowing the underlying type
* The Go native json marshaller / unmarshaller can be used without implementing custom marshaller/unmarshaller
* Does not use the `reflect` package
* Does not produce the same json but the I think the alternative json I produce is actually useful for some client apps

```go
type ColumnValue struct {
	// Only one of the following field should be set
	I8   int8    `json:"i8,omitempty"`
	I16  int16   `json:"i16,omitempty"`
	I32  int32   `json:"i32,omitempty"`
	I64  int64   `json:"i64,omitempty"`
[...]

PR #2168, opened 2023-10-20
run: Only enable relevant operators according to the gadget being run

mauriciovasquezbernal:
This PR handles the issues described in #2167. 

### How to test

Considering the following simple gadget that doesn't provide any mount ns related info:

```c
// program.bpf.c

// Kernel types definitions
// Check https://blog.aquasec.com/vmlinux.h-ebpf-programs for more details
#include <vmlinux.h>

// eBPF helpers signatures
// Check https://man7.org/linux/man-pages/man7/bpf-helpers.7.html to learn
// more about different available helpers
#include <bpf/bpf_helpers.h>

// Inspektor Gadget macros
#include <gadget/macros.h>

struct event {
	__u32 pid;
};

// we need to "use" this struct for the compiler not to remove it in an optimization pass
const struct event *unusedevent __attribute__((unused));

struct {
	__type(value, struct event);
} events SEC(".maps");

GADGET_TRACE_MAP(events);

int enter_openat(struct trace_event_raw_sys_enter *ctx)
{
	struct event event = {};

	event.pid = bpf_get_current_pid_tgid() >> 32;

	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU,
			      &event, sizeof(event));

	return 0;
}

```

and its metadata file

```yaml
name: 'TODO: Fill the gadget name'
description: 'TODO: Fill the gadget description'
tracers:
  events:
    mapName: events
    structName: event
structs:
  event:
    fields:
    - name: pid
      description: 'TODO: Fill field description'
      attributes:
        width: 10
        alignment: left
        ellipsis: end
```

Build and tag is as `mygadget`

```bash
$ sudo -E ig image build .  -t mygadget
```

### Before this PR 

```bash
$ sudo -E ig run mygadget -h
Run a containerized gadget
...
```

The gadget provides some filtering options that don't do anything
```bash 
   -c, --containername string   Show only data from containers with that name (default ""
...
      --host                   Show data from both the host and containers
```

And a lot of possible columns that aren't really available

```bash
                                Available columns:
                                     k8s.container
                                     k8s.hostnetwork
                                     k8s.namespace
                                     k8s.node
                                     k8s.pod
                                     mntns                                     
                                     runtime.containerId
                                     runtime.containerImageDigest
                                     runtime.containerImageName
                                     runtime.containerName
                                     runtime.runtimeName

```

```bash
$ sudo -E ig run mygadget 
RUNTIME.CONTAINERNAME                                                                                                  PID                                    
                                                                                                                       8626                                   
                                                                                                                       3516                                   
                                                                                                                       1140                                   
                                                                                                                       1140                                   
                                                                                                                       1140                                   
                                                                                                                       1140                                   
                                                                                                                       1140                                   
                                                                                                                       1140                                   
                                                                                                                       1140                                   
                                                                                                                       1140    
```

A mount ns set map that is not used is created:

```bash
$ sudo bpftool map list | grep mntnsset -A 2
1996: hash  name mntnsset_21a9da  flags 0x0
	key 8B  value 4B  max_entries 1024  memlock 16384B
	pids ig(438531)
```

### After this PR 

Filtering options and columns aren't shown anymore:
(we still need to fix the timestamp one)

```bash
                                Available columns:
                                  pid
                                  timestamp
```

```bash
$ sudo -E ig run mygadget 
PID                                                                                                                                                           
3516                                                                                                                                                          
2915                                                                                                                                                          
2915                                                                                                                                                          
2915                                                                                                                                                          
2915                                                                                                                                                          
2915                                                                                                                                                          
2915                                                                                                                                                          
442243 
```

And the mount ns set map is not created:

```bash
$ sudo bpftool map list | grep mntnsset -A 2
```


blanquicet:
Are the not-available filtering (e.g., `-c` and `--host`) options also gone with this PR (I wasn't specified in the description)?

mauriciovasquezbernal:
> Are the non-available filtering (e.g., `-c` and `--host`) options also gone with this PR (I wasn't specified in the description)?

Right, those flags aren't shown anymore as the PR description says: "**Filtering options** and columns aren't shown anymore"

cmd/common/registry.go
```diff
@@ -375,6 +378,13 @@ func buildCommandFromGadget(
 				strings.Join(outputFormatsHelp, "\n")+"\n\n",
 			)
 
+			// TODO: why moving this inside PreRunE breaks the namespace parameter?
+			// It's always "default".
+			// Add operator flags
```
mauriciovasquezbernal, line 383: note to myself: Check if this TODO is still true, otherwise remove comment.

pkg/gadgets/run/tracer/helpers.go
```diff
@@ -56,3 +77,259 @@ func getEventTypeBTF(progContent []byte, metadata *types.GadgetMetadata) (*btf.S
...
+		case types.L3EndpointTypeName, types.L4EndpointTypeName:
+			features.HasEndpoints = true
+		}
+	}
+
+	for _, p := range spec.Programs {
+		if p.Type == ebpf.SocketFilter && strings.HasPrefix(p.SectionName, "socket") {
+			features.IsAttacher = true
+			// TODO: info.Features.HasNetNs = true
+			// However it creates an issue because the netns column is added twice!
```
mauriciovasquezbernal, line 175: Check if it's still true after https://github.com/inspektor-gadget/inspektor-gadget/pull/2191

pkg/gadgets/run/tracer/run.go
```diff
@@ -454,6 +231,27 @@ func (g *GadgetDesc) getColumns(info *types.GadgetInfo) (*columns.Columns[types.
...
+	// Remove columns according to the features that the gadget supports.
+	// TODO: what about attacher interface?
+	for name, col := range cols.ColumnMap {
+		if !info.Features.HasMountNs && name == "mntns" {
+			delete(cols.ColumnMap, name)
+		}
+
+		if !info.Features.HasNetNs && name == "netns" {
+			delete(cols.ColumnMap, name)
+		}
```
mauriciovasquezbernal, line 243: Check after https://github.com/inspektor-gadget/inspektor-gadget/pull/2191

mauriciovasquezbernal (commented):
Some auto review to avoid forgetting.

flyth:
To be honest, I'm not sure whether this "features" struct is the best approach - mainly because it defeats the advantage of the modularity that the operators have now. Another more modular approach could be to allow the operator to look at the programs' spec and then decide from there on if it should be enabled or not.

Combined with the generic und fully dynamic data structure discussed in #2131, the operator could then add the required fields and fill them out at runtime. E.g.: the bpf program uses a type like `gadget_mnt_ns`, the operator catches it, enables itself, adds all additional runtime fields and fills them out accordingly afterwards.

Keeping the operators modular allows us and also third parties to simply add functionality without touching a great amount of code, thus keeping it much easier to maintain. The operator concept could also be extended to wasm modules later on using the same/similar APIs.

Happy to provide a PoC for this if you're not entirely against it :)

mauriciovasquezbernal:
> To be honest, I'm not sure whether this "features" struct is the best approach - mainly because it defeats the advantage of the modularity that the operators have now. Another more modular approach could be to allow the operator to look at the programs' spec and then decide from there on if it should be enabled or not.
> 
> Combined with the generic und fully dynamic data structure discussed in #2131, the operator could then add the required fields and fill them out at runtime. E.g.: the bpf program uses a type like `gadget_mnt_ns`, the operator catches it, enables itself, adds all additional runtime fields and fills them out accordingly afterwards.

It is a very interesting idea. I'll explore it.

mauriciovasquezbernal:
Closing this one as I'll be fixed by some refactoring being done by @flyth.

PR #2174, opened 2023-10-23
cmd: Ignore unknown flags when parsing params early

burak-ok:
When connecting to a server we need to have informations about the connection itself. That is why we parse the flags as early as possible. Even before we have all of them available, because the flags are derived from the servers information.

This commit enableds the option to ignore unknown flags on this early flag parsing, so we can get information on the connection flags without erroring out.

## Replicating the issue before this patch

Run `ig` in daemon mode in a terminal:
```bash
$ sudo -E ./ig daemon -H tcp://127.0.0.1:9999
```
Open a new terminal and do the following:
```bash
$ rm ~/.ig/info.json
$ ./gadgetctl top file --all-files --remote-address tcp://127.0.0.1:9999
WARN[0005] could not load gadget info from remote: dialing random target: dialing "unix:///var/run/ig/ig.socket" ("local"): dialing "unix:///var/run/ig/ig.socket" ("local"): context deadline exceeded 
Error: unknown command "top" for "gadgetctl"
Run 'gadgetctl --help' for usage.
```

### DIFF
--- a/cmd/common/utils/flags.go
+++ b/cmd/common/utils/flags.go
@@ -17,6 +17,7 @@ package utils
 import (
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -151,4 +152,44 @@ func AddRegistryAuthVariablesAndFlags(cmd *cobra.Command, authOptions *oci.AuthO
 		false,
 		"Allow connections to HTTP only registries",
 	)
+}
+
+// removeSplitSortArgs removes the --sort flag with its arg, if it isn't in the
+// merged form of --sort=foo
+func removeSplitSortArgs(args []string) []string {
+	for i := 0; i < len(args); i++ {
+		if args[i] == "--sort" {
+			// Remove also the next element as it is the arg of --sort
+			return append(args[:i], args[i+2:]...)
+		}
+	}
+	return args
+}
+
+// removeHelpArg removes the --help flag
+func removeHelpArg(args []string) []string {
+	for i := 0; i < len(args); i++ {
+		if args[i] == "--help" || args[i] == "-h" {
+			return append(args[:i], args[i+1:]...)
+		}
+	}
+	return args
+}
+
+func ParseEarlyFlags(cmd *cobra.Command) error {
+	// Do not error out on unknown flags, but still validate currently
+	// known ones.
+	// Other flags will be validated in the `Execute()` call and unknown
+	// ones will be rejected
+
+	cmd.FParseErrWhitelist.UnknownFlags = true
+	defer func() {
+		cmd.FParseErrWhitelist.UnknownFlags = false
+	}()
+	// temporary workaround for https://github.com/inspektor-gadget/inspektor-gadget/pull/2174#issuecomment-1780923952
+	args := removeSplitSortArgs(os.Args[1:])
+	// --help should be handled after we registered all commands
+	args = removeHelpArg(args)
+	err := cmd.ParseFlags(args)
+	return err
 }
--- a/cmd/gadgetctl/main.go
+++ b/cmd/gadgetctl/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/all-gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/environment"
 	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
@@ -68,18 +69,12 @@ func main() {
 		// sure that --remote-address has already been parsed when calling
 		// InitDeployInfo(), so it can target the specified address
 
-		// Do not error out on unknown flags, but still validate currently
-		// known ones.
-		// Other flags will be validated in the `Execute()` call and unknown
-		// ones will be rejected
-		rootCmd.FParseErrWhitelist.UnknownFlags = true
-		err := rootCmd.ParseFlags(os.Args[1:])
+		err := commonutils.ParseEarlyFlags(rootCmd)
 		if err != nil {
 			// Analogous to cobra error message
 			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 			os.Exit(1)
 		}
-		rootCmd.FParseErrWhitelist.UnknownFlags = false
 		runtime.InitDeployInfo()
 	}
 
--- a/cmd/gadgetctl/main.go
+++ b/cmd/gadgetctl/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -66,7 +67,19 @@ func main() {
 		// evaluate flags early for runtimeGlobalFlags; this will make
 		// sure that --remote-address has already been parsed when calling
 		// InitDeployInfo(), so it can target the specified address
-		rootCmd.ParseFlags(os.Args[1:])
+
+		// Do not error out on unknown flags, but still validate currently
+		// known ones.
+		// Other flags will be validated in the `Execute()` call and unknown
+		// ones will be rejected
+		rootCmd.FParseErrWhitelist.UnknownFlags = true
+		err := rootCmd.ParseFlags(os.Args[1:])
+		if err != nil {
+			// Analogous to cobra error message
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
+		rootCmd.FParseErrWhitelist.UnknownFlags = false
 		runtime.InitDeployInfo()
 	}
 
--- a/cmd/kubectl-gadget/main.go
+++ b/cmd/kubectl-gadget/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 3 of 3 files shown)


cmd/gadgetctl/main.go
```diff
@@ -66,6 +67,17 @@ func main() {
...
+		// known ones.
+		rootCmd.FParseErrWhitelist.UnknownFlags = true
+		err := rootCmd.ParseFlags(os.Args[1:])
+		if err != nil {
+			// Analogous to cobra error message
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
+		rootCmd.FParseErrWhitelist.UnknownFlags = false
 		rootCmd.ParseFlags(os.Args[1:])
```
mauriciovasquezbernal, line 81: Is this line still required?

burak-ok: No, not at all

burak-ok:
> Also, it'd be nice to have a fixes tag in case we need to track this down. I suppose it's `Fixes: e89e6828c334 ("add gadgetctl as client for ig")` & `Fixes: cd5768eb8078 ("pkg/runtime/grpc: add option to connect to the gadget service using either tcp or unix socket")`

I added it. But just to know how this would help in the future.

cmd/gadgetctl/main.go
```diff
@@ -66,7 +67,19 @@ func main() {
 		// evaluate flags early for runtimeGlobalFlags; this will make
 		// sure that --remote-address has already been parsed when calling
 		// InitDeployInfo(), so it can target the specified address
-		rootCmd.ParseFlags(os.Args[1:])
+
+		// Do not error out on unknown flags, but still validate currently
+		// known ones.
+		// Other flags will be validated in the `Execut()` call and unknown
```
mauriciovasquezbernal, line 73:
```suggestion
		// Other flags will be validated in the `Execute()` call and unknown
```

cmd/kubectl-gadget/main.go
```diff
@@ -78,7 +79,19 @@ func main() {
 		// evaluate flags early for runtimeGlobalFlags; this will make
 		// sure that --connection-method has already been parsed when calling
 		// InitDeployInfo()
-		rootCmd.ParseFlags(os.Args[1:])
+
+		// Do not error out on unknown flags, but still validate currently
+		// known ones.
+		// Other flags will be validated in the `Execut()` call and unknown
```
mauriciovasquezbernal, line 85: ditto

mauriciovasquezbernal (approved):
LGTM once the typo is fixed. Thanks for handling this!

mauriciovasquezbernal:
> I added it. But just to know how this would help in the future.

In Linux it's used to understand if a given patch has to be backported to a stable version. We're not maintaining any stable version, hence we don't need it. However I think it helps documenting the git tree to understand when a given bug was introduced.

burak-ok:
# This bugfix opens a new bug which the integration tests thankfully revealed

## How did the early params parse work before this PR

We first add some flags to the `rootCmd`. These are mostly flags for k8s connection settings (`kubectl-gadget`) or grpc connection settings(`gadgetctl`)
The parsing call parses all flags from left to right. It skips arguments without dashes in front of them.
So everything worked if one used `./gadgetctl top file --remote-address tcp://127.0.0.1:9999 --all-files`. The parser would parse everything until it is at `--all-files`. Since that flag is not registered yet with cobra, it returns an error. The returned error is not checked.

Therefore executing `./gadgetctl top file --all-files --remote-address tcp://127.0.0.1:9999` will not work and we won't get any feedback why it failed. We just get an error that it can not connect to the socket even though we gave another remote address sepcifically

## How this Bugfix PR currently works

For the very first flag parse of `rootCmd` we set `UnknownFlags = true` and back to `UnknownFlags = false` afterwards. This means no errors will be returned for flags that are unknown.
So `./gadgetctl top file --all-files --remote-address tcp://127.0.0.1:9999` works even though `--all-files` is not yet recognized.
For this the `pflag` library has to do some guessing what arguments would be flags, the flag values, etc...
It works great for most of our usages

## The problem with the current Bugfix PR

`./kubectl-gadget top file -n test-top-file-352369229555705923 --sort -writes -o json --timeout 10` is the unit test which failed. Let's bring it down to the minimal failing example:
`./kubectl-gadget top file --sort -writes -o json`

The problem specifically is with `--sort -writes`. Remember that we currently do not have any flags added to the command than the ones for the K8s connection. So the `pflag` library sees `--sort` and guesses that this is a boolean flag since the next arg starts with a dash (`-`) `--sort -writes`.
Now it looks at `-writes` and interprets it as a group of shorthand flags: `-writes` -> `-w -r -i -t -e -s`. Lets just skip everything until the `-s` at the end.

`-s` is the shorthand for the `--server` flag for the K8s connection and we already have it added to the `rootCmd`. It now takes the next argument and uses that string as the address and port of the Kubernetes API server. The next arg in our example (`./kubectl-gadget top file --sort -writes -o json`) is `-o`.
Therefore it tries to connect to `http://-o/api/v1/namespaces/gadget/pods?labelSelector=k8s-app%3Dgadget` which doesn't make sense at all

burak-ok:
The newest commit add function for 
1. Remove `--sort foobar`
2. Remove `--help`

This is only done for the first early parsing of the flags before all commands and flags are added to cobra

cmd/common/utils/flags.go
```diff
@@ -152,3 +152,29 @@ func AddRegistryAuthVariablesAndFlags(cmd *cobra.Command, authOptions *oci.AuthO
...
+func RemoveHelpArg(args []string) []string {
+	var newArgs []string
+	for i := 0; i < len(args); i++ {
+		if args[i] == "--help" {
+			continue
+		}
+		newArgs = append(newArgs, args[i])
+	}
+	return newArgs
+}
```
blanquicet, line 180: nit: Once you find the index of the arg(s) you want to delete, you don't need to continue iterating:

```suggestion
// RemoveSplitSortArgs removes the --sort flag with its arg, if it isn't in the
// merged form of --sort=foo
func RemoveSplitSortArgs(args []string) []string {
	for i := 0; i < len(args); i++ {
		if args[i] == "--sort" {
			// Remove also the next element as it is the arg of --sort
			return append(args[:i], args[i+2:]...)
		}
	}
	return args
}

// RemoveHelpArg removes the --help flag
func RemoveHelpArg(args []string) []string {
	for i := 0; i < len(args); i++ {
		if args[i] == "--help" {
			return append(args[:i], args[i+1:]...)
		}
	}
	return args
}
```

cmd/gadgetctl/main.go
```diff
@@ -66,7 +68,23 @@ func main() {
...
+
+		// Do not error out on unknown flags, but still validate currently
+		// known ones.
+		// Other flags will be validated in the `Execute()` call and unknown
+		// ones will be rejected
+		rootCmd.FParseErrWhitelist.UnknownFlags = true
+		// temporary workaround for https://github.com/inspektor-gadget/inspektor-gadget/pull/2174#issuecomment-1780923952
+		args := commonutils.RemoveSplitSortArgs(os.Args[1:])
+		// --help should be handled after we registered all commands
+		args = commonutils.RemoveHelpArg(args)
```
blanquicet, line 80: Mmmmm you mean that if cobra parses here the `--help` flag it will compose immediately the help message even if `Execute` hasn't been called yet? Otherwise, I don't see why we should remove `--help`. What am I missing?

burak-ok: The `--help` is not defined (yet) and we get the following error: https://pkg.go.dev/github.com/spf13/pflag@v1.0.5#ErrHelp

cmd/common/utils/flags.go
mauriciovasquezbernal, line 180: What about creating a single `RemoveUnkownFlags` function?

cmd/kubectl-gadget/main.go
```diff
@@ -78,7 +80,23 @@ func main() {
...
+		args := commonutils.RemoveSplitSortArgs(os.Args[1:])
+		// --help should be handled after we registered all commands
+		args = commonutils.RemoveHelpArg(args)
+		err := rootCmd.ParseFlags(args)
+		if err != nil {
+			// Analogous to cobra error message
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
+		rootCmd.FParseErrWhitelist.UnknownFlags = false
```
mauriciovasquezbernal, line 99: Perhaps all of it could go to a helper in commonutils and be reused from kubectl-gadget and gadgetctl?

Something like: 

```golang
// ParseEarlyFlags evaluates flags that are needed to connect to the server.
func ParseEarlyFlags(cmd *cobra.Cmd, args []string) error
```

mauriciovasquezbernal (commented):
Some minor comments. I tested it in different situations and works fine (besides the missing logic to handle `-h`). I'll do a final pass and approve once current comments are handled.

cmd/common/utils/flags.go
```diff
@@ -152,3 +152,29 @@ func AddRegistryAuthVariablesAndFlags(cmd *cobra.Command, authOptions *oci.AuthO
...
+
+// RemoveHelpArg removes the --help flag
+func RemoveHelpArg(args []string) []string {
+	var newArgs []string
+	for i := 0; i < len(args); i++ {
+		if args[i] == "--help" {
```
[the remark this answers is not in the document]
burak-ok: I know that with manual flag parsing I will forget a case :D

cmd/common/utils/flags.go
```diff
@@ -152,3 +153,41 @@ func AddRegistryAuthVariablesAndFlags(cmd *cobra.Command, authOptions *oci.AuthO
 		"Allow connections to HTTP only registries",
 	)
 }
+
+// RemoveSplitSortArgs removes the --sort flag with its arg, if it isn't in the
+// merged form of --sort=foo
+func RemoveSplitSortArgs(args []string) []string {
```
mauriciovasquezbernal, line 159: nit: this and RemoveHelpArg can be private.

mauriciovasquezbernal (approved):
One last final nit. LGTM. thanks for handling it.

blanquicet (approved):
LGTM from code inspection.

cmd/common/utils/flags.go
```diff
@@ -152,3 +153,41 @@ func AddRegistryAuthVariablesAndFlags(cmd *cobra.Command, authOptions *oci.AuthO
...
+	return args
+}
+
+func ParseEarlyFlags(cmd *cobra.Command) error {
+	// Do not error out on unknown flags, but still validate currently
+	// known ones.
+	// Other flags will be validated in the `Execute()` call and unknown
+	// ones will be rejected
+
+	cmd.FParseErrWhitelist.UnknownFlags = true
```
blanquicet, line 185: nit: Looks like this places where a nice `defer` would make code clearer (even if it doesn't provide any value 🙂):

```suggestion
	cmd.FParseErrWhitelist.UnknownFlags = true
	defer func() {
		cmd.FParseErrWhitelist.UnknownFlags = false
	}()
```

burak-ok: I am always a fan of defer. Thanks

PR #2176, opened 2023-10-23
c-u/docker: add `buildContainerData`

### LINKED ISSUE #1911: Combine code in `taskAndContainerToContainerData` and `GetContainer`
I think we should combine code in `taskAndContainerToContainerData` and `GetContainer` to avoid having to update both of them every time we add a new field.

_Originally posted by @blanquicet in https://github.com/inspektor-gadget/inspektor-gadget/pull/1876#discussion_r1275974355_

claudiamarcubina:
refactor `GetContainerDetails` and `DockerContainerToContainerData` to simplify maintenance

resolve #1911

### DIFF
--- a/pkg/container-utils/docker/docker.go
+++ b/pkg/container-utils/docker/docker.go
@@ -159,20 +159,17 @@ func (c *DockerClient) GetContainerDetails(containerID string) (*runtimeclient.C
 		return nil, errors.New("container host config is nil")
 	}
 
+	containerData := buildContainerData(
+		containerJSON.ID,
+		containerJSON.Name,
+		containerJSON.Config.Image,
+		containerJSON.State.Status,
+		containerJSON.Config.Labels)
+
 	containerDetailsData := runtimeclient.ContainerDetailsData{
-		ContainerData: runtimeclient.ContainerData{
-			Runtime: runtimeclient.RuntimeContainerData{
-				BasicRuntimeMetadata: types.BasicRuntimeMetadata{
-					ContainerID:        containerJSON.ID,
-					ContainerName:      strings.TrimPrefix(containerJSON.Name, "/"),
-					RuntimeName:        types.RuntimeNameDocker,
-					ContainerImageName: getContainerImageNamefromImage(containerJSON.Config.Image, containerJSON.ID),
-				},
-				State: containerStatusStateToRuntimeClientState(containerJSON.State.Status),
-			},
-		},
-		Pid:         containerJSON.State.Pid,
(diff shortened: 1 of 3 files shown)


pkg/container-utils/docker/docker.go
```diff
@@ -278,3 +266,40 @@ func getContainerImageNamefromImage(image string, containerID string) string {
 	// Case 3 or 4
 	return image
 }
+
+// `getContainerData` takes in two pointers to `dockertypes.ContainerJSON` and `dockertypes.Container`
+// and returns a pointer to a `runtimeclient.ContainerData` and an error. It extracts basic metadata about
+// a Docker container and constructs a `runtimeclient.ContainerData` struct with this information.
+// The function also maps the container state to a runtime client state.
+// If either of the input parameters is `nil`, the function returns an error.
+func getContainerData(containerJSON *dockertypes.ContainerJSON, container *dockertypes.Container) (*runtimeclient.ContainerData, error) {
```
blanquicet, line 275: IMO, I would prefer this function to receive the ID, Name, Image and State.

blanquicet, line 275: As for the previous PR, please extend this function to receive the labels and enrich the ContainerData with the K8s metadata.

blanquicet (commented):
Thanks for working on this. I would make the code clear, avoiding duplicated logic.

pkg/container-utils/docker/docker.go
claudiamarcubina: Thank you for the feedback, Jose :) I added the changes in a new commit, as my first commit was turned into a merge commit and seemed a bit messy to alter that one.

blanquicet: What about moving the call of `strings.TrimPrefix`, `getContainerImageNamefromImage` and `containerStatusStateToRuntimeClientState` inside `buildContainerData`?

claudiamarcubina: That's indeed a very good idea! Done. Thanks!

burak-ok (commented):
Looks good and will help with maintance so much thanks :)

Just a small comment: Could you sqash the commits so it is a single one?

blanquicet (approved):
LGTM. Please squash commits.

claudiamarcubina:
> Looks good and will help with maintance so much thanks :)
> 
> Just a small comment: Could you sqash the commits so it is a single one?

Thanks for checking :) Squashing done.

burak-ok (approved):

PR #2177, opened 2023-10-24
image: Use empty descriptor if no metadata file is provided

mqasimsarfraz:
As mentioned in the artifact usage [1] of image spec, if we don't have config we should use an empty descriptor [2]. This should avoid running into issues when pushing the images to registry since most of them [don't support](https://github.com/opencontainers/image-spec/issues/1025) "no config" option.

[1] - https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage
[2] - https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidance-for-an-empty-descriptor

### running
```
→ sudo rm -rf /var/lib/ig
→ DIR=$(mktemp -d)
→ cp gadgets/trace_open/program.bpf.c $DIR
→ go run -exec "sudo -E" ./cmd/ig image build -t localhost:5000/trace_open:latest $DIR
Successfully built localhost:5000/trace_open:latest@sha256:a3088f32924f4a00f975c06310c0c125d39023d26fdff27844ca4cfadd6ac603
→ go run -exec "sudo -E" ./cmd/ig run localhost:5000/trace_open:latest 
...

```
### pushing
```
→ docker run --rm -d -p 5000:5000 --name registry registry:2
go run -exec "sudo -E" ./cmd/ig image push --insecure localhost:5000/trace_open:latest
```
### pushing (w/o empty descriptor)

```
→ go run -exec "sudo -E" ./cmd/ig image push --insecure localhost:5000/trace_open:latest 
Pushing localhost:5000/trace_open:latest...
Error: pushing gadget: copying to remote repository: invalid reference
exit status 1

```

### DIFF
--- a/pkg/oci/build.go
+++ b/pkg/oci/build.go
@@ -148,6 +148,16 @@ func createMetadataDesc(ctx context.Context, target oras.Target, metadataFilePat
 	return defDesc, nil
 }
 
+func createEmptyDesc(ctx context.Context, target oras.Target) (ocispec.Descriptor, error) {
+	emptyDesc := ocispec.DescriptorEmptyJSON
+	emptyDesc.ArtifactType = eBPFObjectMediaType
+	err := pushDescriptorIfNotExists(ctx, target, emptyDesc, bytes.NewReader(emptyDesc.Data))
+	if err != nil {
+		return ocispec.Descriptor{}, fmt.Errorf("pushing empty descriptor: %w", err)
+	}
+	return emptyDesc, nil
+}
+
 func createManifestForTarget(ctx context.Context, target oras.Target, metadataFilePath, progFilePath, arch string) (ocispec.Descriptor, error) {
 	progDesc, err := createEbpfProgramDesc(ctx, target, progFilePath)
 	if err != nil {
@@ -161,6 +171,12 @@ func createManifestForTarget(ctx context.Context, target oras.Target, metadataFi
 		defDesc, err = createMetadataDesc(ctx, target, metadataFilePath)
 		if err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf("creating metadata descriptor: %w", err)
+		}
+	} else {
+		// Create an empty descriptor
+		defDesc, err = createEmptyDesc(ctx, target)
+		if err != nil {
+			return ocispec.Descriptor{}, fmt.Errorf("creating empty descriptor: %w", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/oci/build.go
```diff
@@ -148,6 +148,16 @@ func createMetadataDesc(ctx context.Context, target oras.Target, metadataFilePat
 	return defDesc, nil
 }
 
+func createEmptyDesc(ctx context.Context, target oras.Target) (ocispec.Descriptor, error) {
+	emptyDesc := ocispec.DescriptorEmptyJSON
+	emptyDesc.ArtifactType = eBPFObjectMediaType
```
mauriciovasquezbernal, line 153: Is it required and if yes, shouldn't it be metadataMediaType?

mauriciovasquezbernal (commented):
Just a question. I tested and it works fine.

Thanks a lot for this fix.

pkg/oci/build.go
burak-ok: I think it should be even something different completly: `application/vnd.oci.empty.v1+json`

burak-ok (commented):
Thanks for fixing this. I guess I never thought about not giving a metadatafile while implementing this :D

pkg/oci/build.go
mqasimsarfraz: > Is it required

Yes it is, spec says "If the config.mediaType is set to the empty value, the artifactType MUST be defined."

P.S: Empty value != no config :D 

> shouldn't it be metadataMediaType?

Great question. I wasn't too sure about it. You are right `metadataMediaType` sound better. I looked at the example 2 [here](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage) and went with the type from the layer but it might not work well if we add more layers. What do you say?

mqasimsarfraz: > I think it should be even something different completly: `application/vnd.oci.empty.v1+json` (`MediaTypeEmptyJSON`)

The media type in descriptor already has this type. So it has to be something that tells about the artifact in the image. No?

burak-ok: > > I think it should be even something different completly: application/vnd.oci.empty.v1+json (MediaTypeEmptyJSON)
> 
> I think that's for the mediaType.

I got confused because of the name `metadataMediaType` :see_no_evil:

mqasimsarfraz: > I'm not totally sure, after reading the spec, I'd say `eBPFObjectMediaType` is better as it refers to the layer below.

I thought about this bit more and I think we should stick with `eBPFObjectMediaType`. I mean we can trim `+type` ( or use a different artifactType) but I am fine with what we have currently.

mqasimsarfraz:
Thanks for the review folks. I think it is ready to go in. All I am missing is an approval! :laughing:

burak-ok (approved):
LGTM

PR #2181, opened 2023-10-26
gadgettracermanager: Make gRPC buffer size configurable.

eiffel-fl:
This commit adds flags and environment values to set gRPC buffer size rather than having it hardcoded.
The default value size is set to 16384 to improve horizontal scaling.

### DIFF
--- a/cmd/ig/daemon.go
+++ b/cmd/ig/daemon.go
@@ -39,6 +39,7 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 
 	var socket string
 	var group string
+	var eventBufferLength uint64
 
 	daemonCmd.PersistentFlags().StringVarP(
 		&group,
@@ -54,6 +55,13 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 		api.DefaultDaemonPath,
 		"The socket to listen on for new requests. Can be a unix socket"+
 			" (unix:///path/to.socket) or a tcp socket (tcp://127.0.0.1:1234)")
+
+	daemonCmd.PersistentFlags().Uint64VarP(
+		&eventBufferLength,
+		"events-buffer-length",
+		"",
+		16384,
+		"The events buffer length. A low value could impact horizontal scaling.")
 
 	daemonCmd.RunE = func(cmd *cobra.Command, args []string) error {
 		if os.Geteuid() != 0 {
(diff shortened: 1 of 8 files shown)


cmd/ig/daemon.go
```diff
@@ -55,6 +56,13 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 		"The socket to listen on for new requests. Can be a unix socket"+
 			" (unix:///path/to.socket) or a tcp socket (tcp://127.0.0.1:1234)")
 
+	daemonCmd.PersistentFlags().Uint64VarP(
+		&size,
```
flyth, line 60: Maybe a more specific name like `eventBufferSize` or `eventBufferLen` is better here.

pkg/gadget-service/service.go
```diff
@@ -56,16 +56,18 @@ type RunConfig struct {
...
 	api.UnimplementedGadgetManagerServer
-	listener net.Listener
-	runtime  runtime.Runtime
-	logger   logger.Logger
-	servers  map[*grpc.Server]struct{}
+	listener   net.Listener
+	runtime    runtime.Runtime
+	logger     logger.Logger
+	servers    map[*grpc.Server]struct{}
+	bufferSize uint64
```
flyth, line 63: Also here - maybe `eventBufferSize` or `eventBufferLen`.

flyth (commented):
I think having this number configurable is a really good thing. At which point did it lead to issues? Can you please attach some data/benchmarks around it?

Also, this is about the buffer of events (in _number of events_, __not__ _number of bytes of event data_) that we're holding before marshaling them, so it's got nothing to do at that stage with gRPC, yet (gRPC uses buffers on its own, so let's not confuse by calling this gRPC buffer (naming again, I know... hehe) - even though it eventually will get sent over gRPC). So I think `eventBufferSize` or `eventBufferLen` would be a better name.

Just fyi, sending looks something like this:
```
Event -> __Event Buffer__ -> Marshaling -> gRPC -> TCP send buffer -> wire
```

The buffer is mainly used to throw away events if the client is slow in reading the events (and thus the tcp write on the server is blocking), and it especially handles bursts at the start of a gadget run - it also kicks in if marshaling is too slow to keep up (if the CPU is under heavy load for example). It would be interesting to know if increasing the TCP write buffer would also help.

burak-ok (approved):
From my side it looks good.

Issue #2000, opened 2023-09-05
containerized-gadgets: Timestamp column not working


mauriciovasquezbernal:
There are multiple problems with the timestamp column in containerized gadgets.

1. Adding that column to the definition file generates an error while running the gadget:

https://github.com/inspektor-gadget/inspektor-gadget/blob/5bf93b0a60704ee45d0a6c435de87be7769da037/gadgets/trace_open.yaml#L39-L43

```bash
$ go run -exec "sudo -E" ./cmd/ig/. run --prog @./gadgets/trace_open_x86.bpf.o --definition @./gadgets/trace_op
en.yaml
[sudo] password for mvb:            
INFO[0000] Experimental features enabled                
Error: calling custom parser: adding fields: duplicate column name "timestamp"
exit status 1
```

This happens because the event produced by the run gadget includes `eventtypes.Event` that already has a timestamp inside:

https://github.com/inspektor-gadget/inspektor-gadget/blob/a21a25751f0d3a8956838445173cecdcbcc5b28f/pkg/gadgets/run/types/types.go#L33

https://github.com/inspektor-gadget/inspektor-gadget/blob/a21a25751f0d3a8956838445173cecdcbcc5b28f/pkg/types/types.go#L297

This could be fixed by creating some logic to avoid adding the timestamp column as it's already present on the event.

3. Timestamp is not actually set 

```bash
$ go run -exec "sudo -E" ./cmd/ig/. run --prog @./gadgets/trace_open_x86.bpf.o --definition @./gadgets/trace_open.yaml -o columns=timestamp
INFO[0000] Experimental features enabled                
TIMESTAMP                          
1969-12-31T19:00:00.000000000-05:00
...
1969-12-31T19:00:00.000000000-05:00
```

I think we're missing a call to `WallTimeFromBootTime`  in the gadget to set this value: https://github.com/inspektor-gadget/inspektor-gadget/blob/a21a25751f0d3a8956838445173cecdcbcc5b28f/pkg/gadgets/run/tracer/tracer.go#L393-L401

ref https://github.com/inspektor-gadget/inspektor-gadget/pull/1997

PR #2188, opened 2023-10-27
gadgets: Fix trace_dns and trace_tcpretrans failing to run

mauriciovasquezbernal:
Gadgets were failing with the following error:

Error: calling custom parser: getting columns: adding fields: duplicate column name "netns"

This commit workaround that by disabling the netns column until a proper fix is implemented.

### DIFF
--- a/gadgets/trace_dns/gadget.yaml
+++ b/gadgets/trace_dns/gadget.yaml
@@ -51,13 +51,13 @@ structs:
     - name: anaddr
       attributes:
         width: 16
-    - name: netns
-      description: 'TODO: Fill field description'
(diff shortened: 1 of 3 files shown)


burak-ok (approved):
LGTM

Didn't we also have this issue with timestamp

mauriciovasquezbernal:
> Didn't we also have this issue with timestamp

You're right -> https://github.com/inspektor-gadget/inspektor-gadget/issues/2000

Actually I found this issue while working on fixing the timestamp one. I'll open a PR for both fixes later on.

Thanks for the review.

PR #2195, opened 2023-11-03
flags: Clone to avoid modifying os.Args

mqasimsarfraz:
If we try to shrink args (to drop '--sort ...') e.g from `top file --sort -writes -m 10 --timeout 10 --interval 10` we end up reusing the underlying array and hence the original os.Args end up getting modified resulting in `top file -m 10 --timeout 10 --interval 10 --interval 10`. The commit clones the slice to ensure os.Args stay untouched and evaluated correctly later on.

### before the PR

```
→ ./kubectl-gadget  snapshot process --sort pid  --all-namespaces
K8S.NODE                                   K8S.NAMESPACE                             K8S.POD                                   K8S.CONTAINER                             COMM             PID                  UID        GID       
minikube-docker                            gadget                                    gadget-76f66                              gadget                                    gadgettracerman  3591435              0          0         
minikube-docker                            kube-system                               storage-provisioner                       storage-provisioner                       storage-provisi  3269857              0          0         
minikube-docker                            kube-system                               coredns-6d4b75cb6d-9dsjr                  coredns                                   coredns          3269733              0          0         
minikube-docker                            kube-system                               kube-proxy-swnds                          kube-proxy                                kube-proxy       3269478              0          0         
minikube-docker                            kube-system                               kube-apiserver-minikube-docker            kube-apiserver                            kube-apiserver   3268122              0          0         
minikube-docker                            kube-system                               kube-controller-manager-minikube-docker   kube-controller-manager                   kube-controller  3268080              0          0         
minikube-docker                            kube-system                               etcd-minikube-docker                      etcd                                      etcd             3268070              0          0         
minikube-docker                            kube-system                               kube-scheduler-minikube-docker            kube-scheduler                            kube-scheduler   3268062              0          0   
```

### After the PR

```
→ ./kubectl-gadget  snapshot process --sort pid  --all-namespaces
K8S.NODE                                   K8S.NAMESPACE                             K8S.POD                                   K8S.CONTAINER                             COMM             PID                  UID        GID       
minikube-docker                            kube-system                               kube-scheduler-minikube-docker            kube-scheduler                            kube-scheduler   3268062              0          0         
minikube-docker                            kube-system                               etcd-minikube-docker                      etcd                                      etcd             3268070              0          0         
minikube-docker                            kube-system                               kube-controller-manager-minikube-docker   kube-controller-manager                   kube-controller  3268080              0          0         
minikube-docker                            kube-system                               kube-apiserver-minikube-docker            kube-apiserver                            kube-apiserver   3268122              0          0         
minikube-docker                            kube-system                               kube-proxy-swnds                          kube-proxy                                kube-proxy       3269478              0          0         
minikube-docker                            kube-system                               coredns-6d4b75cb6d-9dsjr                  coredns                                   coredns          3269733              0          0         
minikube-docker                            kube-system                               storage-provisioner                       storage-provisioner                       storage-provisi  3269857              0          0         
minikube-docker                            gadget                                    gadget-76f66                              gadget                                    gadgettracerman  3591435              0          0     
```
**Note:**  You need `sort` before `--all-namespace` to ensure while shrinking the args we modify the order of items in [`removeSplitSortArgs`](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/utils/flags.go#L163) to reproduce the issue.

### DIFF
--- a/cmd/common/utils/flags.go
+++ b/cmd/common/utils/flags.go
@@ -187,9 +187,19 @@ func ParseEarlyFlags(cmd *cobra.Command) error {
 		cmd.FParseErrWhitelist.UnknownFlags = false
 	}()
 	// temporary workaround for https://github.com/inspektor-gadget/inspektor-gadget/pull/2174#issuecomment-1780923952
-	args := removeSplitSortArgs(os.Args[1:])
+	args := cloneStringSlice(os.Args[1:]) // clone to avoid modifying the original os.Args
+	args = removeSplitSortArgs(args)
 	// --help should be handled after we registered all commands
 	args = removeHelpArg(args)
 	err := cmd.ParseFlags(args)
 	return err
+}
+
+// TODO: Use slices.Clone() when we upgrade to Go 1.21
+// See: https://cs.opensource.google/go/x/exp/+/7918f672:slices/slices.go;l=336
+func cloneStringSlice(s []string) []string {
+	if s == nil {
+		return nil
+	}
+	return append([]string{}, s...)
 }
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	log "github.com/sirupsen/logrus"
@@ -25,6 +26,7 @@ import (
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/image"
+	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/ig/containers"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
@@ -57,6 +59,15 @@ func main() {
 		containers.NewListContainersCmd(),
 		common.NewVersionCmd(),
 	)
+
+	// evaluate flags early; this will make sure that flags for host are evaluated before
+	// calling host.Init()
+	err := commonutils.ParseEarlyFlags(rootCmd)
+	if err != nil {
+		// Analogous to cobra error message
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
 
 	runtime := local.New()
 	hiddenColumnTags := []string{"kubernetes"}


.github/workflows/inspektor-gadget.yml
```diff
@@ -2,7 +2,7 @@ name: Inspektor Gadget CI
 env:
   REGISTRY: ghcr.io
   CONTAINER_REPO: ${{ github.repository }}
-  GO_VERSION: 1.20.10
```
blanquicet, line 5: Why is this upgrade necessary for this PR?

cmd/common/utils/flags.go
```diff
@@ -187,7 +188,8 @@ func ParseEarlyFlags(cmd *cobra.Command) error {
 		cmd.FParseErrWhitelist.UnknownFlags = false
 	}()
 	// temporary workaround for https://github.com/inspektor-gadget/inspektor-gadget/pull/2174#issuecomment-1780923952
-	args := removeSplitSortArgs(os.Args[1:])
+	args := slices.Clone(os.Args) // clone to avoid modifying the original os.Args
+	args = removeSplitSortArgs(args)
```
blanquicet, line 192: Good catch!

This is another solution but I think your proposal is cleaner:

```go
// removeSplitSortArgs removes the --sort flag with its arg, if it isn't in the
// merged form of --sort=foo
func removeSplitSortArgs(args []string) []string {
	for i := 0; i < len(args); i++ {
		if args[i] == "--sort" {
			// Remove also the next element as it is the arg of --sort
			ret := make([]string, 0, len(args)-2)
			ret = append(ret, args[:i]...)
			return append(ret, args[i+2:]...)
		}
	}
	return args
}

// removeHelpArg removes the --help flag
func removeHelpArg(args []string) []string {
	for i := 0; i < len(args); i++ {
		if args[i] == "--help" || args[i] == "-h" {
			ret := make([]string, 0, len(args)-1)
			ret = append(ret, args[:i]...)
			return append(ret, args[i+1:]...)
		}
	}
	return args
}
```

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: To have `slices.Clone()`, I mean looking at what `slices.Clone()` we can also do (using [append](https://stackoverflow.com/questions/71316774/is-it-equivalent-use-copy-or-append-slice-type-in-golang)):

```args := append([]string{}, os.Args[1:]...) // clone to avoid modifying the original os.Args```

But should be fine to go to `1.21`. What do you say?

File: cmd/common/utils/flags.go (hunk 2 above)
Comment (mqasimsarfraz) [reply cmd/common/utils/flags.go:192]: > This is another solution but I think your proposal is cleaner

Yeah, Also cloning one level above ensure it is done for all following function.

File: .github/workflows/inspektor-gadget.yml (hunk 1 above)
Comment (blanquicet) [reply .github/workflows/inspektor-gadget.yml:5]: IMO, I would avoid upgrading it unless it is strictly necessary or it provides a big advantage. And, I consider this doesn't seem the case.

However, let's see what other maintainers think.

Comment (flyth) [reply .github/workflows/inspektor-gadget.yml:5]: I think 1.21 with its `maps` and `slices` packages and the new `min()` and `max()` funcs alone are worth the upgrade as we use those things constantly. And this is timed just after a release, so there's plenty of time ahead of us to make sure it's running well. So my opinion is: go for it.

The release notes state that Go 1.21 no longer runs on Win 7 & 8 and older MacOS versions, but I don't know if the binaries it produces are affected as well - we should check for that and maybe add that to the release notes.

Comment (mqasimsarfraz) [reply .github/workflows/inspektor-gadget.yml:5]: > (But maybe this should get its own PR prior to this one)

Yeah, maybe it is a better idea to open a separate PR for golang 1.21 and gather more use-cases.

> `args := append([]string{}, os.Args[1:]...) // clone to avoid modifying the original os.Args`

If the [above](https://github.com/inspektor-gadget/inspektor-gadget/pull/2195/files#r1381893852) approach looks good, I can unblock this fix and we can continue with `golang 1.21` separately?

Comment (mauriciovasquezbernal) [reply .github/workflows/inspektor-gadget.yml:5]: I'll suggest not to tie these two things together. Unblock this PR by using that approach and add a comment saying it should be changed once we use go1.21, then in another PR we can discuss the go1.21 implications.

**Comment by mauriciovasquezbernal** [comment]
Perhaps you can also include the commit from https://github.com/inspektor-gadget/inspektor-gadget/pull/2192, so we get all the fixes in the same PR?

File: .github/workflows/inspektor-gadget.yml (hunk 1 above)
Comment (mqasimsarfraz) [reply .github/workflows/inspektor-gadget.yml]: Makes sense. I would open a separate PR for bumping `golang` version. For now I switched to different approach with a comment.

File: cmd/common/utils/flags.go [hunk 3]
```diff
@@ -187,7 +188,8 @@ func ParseEarlyFlags(cmd *cobra.Command) error {
 		cmd.FParseErrWhitelist.UnknownFlags = false
 	}()
 	// temporary workaround for https://github.com/inspektor-gadget/inspektor-gadget/pull/2174#issuecomment-1780923952
-	args := removeSplitSortArgs(os.Args[1:])
+	args := slices.Clone(os.Args) // clone to avoid modifying the original os.Args
```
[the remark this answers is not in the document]
mqasimsarfraz: Yep, that should also do the trick. I read [this](https://github.com/go101/go101/wiki/There-is-not-a-perfect-way-to-clone-slices-in-Go) to confuse myself more :D and then decided to go for whatever [slices](https://cs.opensource.google/go/x/exp/+/7918f672:slices/slices.go;l=336) will give us in future.

flyth (approved):
LGTM, thanks!

mqasimsarfraz:
Thanks for reviews!

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: I opened https://github.com/inspektor-gadget/inspektor-gadget/pull/2206 to update golang version.

PR #2207, opened 2023-11-07
Use distroless as base image

eiffel-fl:
Hi.

This PR does the following:
1. It translates shell scripts used in docker image to golang.
2. It removed the deprecated `exec/unix` way to connect to gadget service.
3. It uses distroless instead of debian slim as gadget `BASE_IMAGE`.

```bash
$ ../kubectl exec -ti -n gadget gadget-dwq27 -- bash       francis/go-entrypoint *% u=
OCI runtime exec failed: exec failed: unable to start container process: exec: "bash": executable file not found in $PATH: unknown
command terminated with exit code 126
$ ./kubectl-gadget trace exec -A | head -5                 francis/go-entrypoint *% u=
K8S.NODE                       K8S.NAMESPACE                  K8S.POD                        K8S.CONTAINER                  PID              PPID             COMM             RET ARGS                                    
minikube-docker                gadget                         gadget-dwq27                   gadget                         212278           212269           gadgettracerman  0   /bin/gadgettracermanager -liveness      
minikube-docker                gadget                         gadget-dwq27                   gadget                         212302           212288           gadgettracerman  0   /bin/gadgettracermanager -liveness      
minikube-docker                gadget                         gadget-dwq27                   gadget                         212353           212344           gadgettracerman  0   /bin/gadgettracermanager -liveness      
minikube-docker                gadget                         gadget-dwq27                   gadget                         212377           212363           gadgettracerman  0   /bin/gadgettracermanager -liveness
```

Best regards.

### DIFF
--- a/Dockerfiles/gadget.Dockerfile
+++ b/Dockerfiles/gadget.Dockerfile
@@ -1,7 +1,7 @@
 # Dockerfile for Inspektor Gadget.
 
 ARG BUILDER_IMAGE=golang:1.21-bullseye
-ARG BASE_IMAGE=debian:bullseye-slim
+ARG BASE_IMAGE=gcr.io/distroless/base-debian12
 
 # bpftrace upstream image
 ARG BPFTRACE="ghcr.io/inspektor-gadget/bpftrace"
@@ -31,25 +31,6 @@ LABEL org.opencontainers.image.description="Inspektor Gadget is a collection of 
 LABEL org.opencontainers.image.documentation="https://inspektor-gadget.io/docs"
 LABEL org.opencontainers.image.licenses=Apache-2.0
 
-# install runtime dependencies  according to the package manager
-# available on the base image
-RUN set -ex; \
-	PACKAGES='ca-certificates util-linux' && \
-	if command -v tdnf; then \
-		tdnf install -y $PACKAGES; \
-	elif command -v yum; then \
-		yum install -y $PACKAGES; \
-	elif command -v apt-get; then \
-		apt-get update && \
-		apt-get install -y $PACKAGES && \
-		apt-get clean && \
-		rm -rf /var/lib/apt/lists/*; \
-	elif command -v apk; then \
-		apk add gcompat $PACKAGES; \
-	fi && \
-	(rmdir /usr/src || true) && ln -sf /host/usr/src /usr/src && \
-	rm -f /etc/localtime && ln -sf /host/etc/localtime /etc/localtime
-
 COPY --from=builder /gadget/gadget-container/bin/entrypoint /
 COPY --from=builder /gadget/gadget-container/bin/cleanup /
 
@@ -71,6 +52,3 @@ COPY gadget-container/hooks/nri/conf.json /opt/hooks/nri/
 ## Hooks Ends
 
 COPY --from=bpftrace /usr/bin/bpftrace /usr/bin/bpftrace
-
-# Mitigate https://github.com/kubernetes/kubernetes/issues/106962.
-RUN rm -f /var/run


gadget-container/entrypoint.go
```diff
@@ -0,0 +1,290 @@
...
+		parts := strings.SplitN(line, "=", 2)
+		if len(parts) != 2 {
+			continue
+		}
+
+		if parts[0] == "PRETTY_NAME" {
+			return strings.Trim(parts[1], "\""), nil
+		}
+	}
+
```
mqasimsarfraz, line 56: Missing `scanner.Err()` check?

gadget-container/entrypoint.go
```diff
@@ -0,0 +1,290 @@
...
+			return fmt.Errorf("copying: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func main() {
+	if _, err := os.Stat(filepath.Join(host.HostRoot, "/bin")); os.IsNotExist(err) {
+		log.Fatalf("%s must be executed in a pod with access to the host via /host", os.Args[0])
```
mqasimsarfraz, line 196: nit: 
```suggestion
		log.Fatalf("%s must be executed in a pod with access to the host via %s", os.Args[0], host.HostRoot)
```

gadget-container/entrypoint.go
```diff
@@ -0,0 +1,290 @@
...
+	case "nri":
+		err := installNRIHooks()
+		if err != nil {
+			log.Fatalf("installing NRI hooks: %w", err)
+		}
+	}
+
+	gadgetTracerManagerHookMode := "auto"
+	switch hookMode {
+	case "crio", "nri":
```
mqasimsarfraz, line 257: nit: I wonder if it makes sense to use `const` for these since we already use them in multiple places [1] [2], not sure about the best place! 

1. https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgettracermanager/gadgettracermanager.go#L282
2. https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/kubectl-gadget/deploy.go#L89

pkg/resources/manifests/deploy.yaml
mqasimsarfraz, line 1: These changes will have to go to https://github.com/inspektor-gadget/inspektor-gadget/blob/main/charts/gadget/templates/daemonset.yaml and then `make generate-manifests` ?

mqasimsarfraz (commented):
Thanks for this, going in the right direction. I will test it later on!

flyth:
I like it!

What do you think about instead of running entrypoint and spawning gadgettracermgr, include what entrypoint does into gadgettracermgr and run it directly? Saves a bit of RAM that the entrypoint binary would take up otherwise.

gadget-container/entrypoint/entrypoint.go
```diff
@@ -0,0 +1,295 @@
...
+	err = copyFile(destinationPath, "/opt/hooks/nri/nrigadget", 0o640)
+	if err != nil {
+		return fmt.Errorf("copying: %w", err)
+	}
+
+	configPath := "/opt/hooks/nri/conf.json"
+
+	content, err := os.ReadFile(configPath)
+	if err == nil {
+		var object map[string]interface{}
```
mqasimsarfraz, line 153: see above comment

gadget-container/entrypoint/entrypoint.go
```diff
@@ -0,0 +1,295 @@
...
+	"strings"
+	"syscall"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
+	"golang.org/x/sys/unix"
+)
+
+var regex = regexp.MustCompile(`^1:name=systemd:.*/crio-[0-9a-f]*\.scope$`)
```
burak-ok, line 34: rename to  `crioRegex`?

gadget-container/entrypoint/entrypoint.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return fmt.Errorf("creating %s: %w", path, err)
+	}
+
+	for _, file := range []string{"ocihookgadget", "prestart.sh", "poststop.sh"} {
+		log.Infof("Installing %s", file)
+
+		path := filepath.Join("/opt/hooks/oci", file)
+		destinationPath := filepath.Join(host.HostRoot, path)
+		err := copyFile(destinationPath, path, 0o640)
```
alban, line 102: Not related to this PR, but I noticed the path on the host is hardcoded to `/host/opt/hooks/oci` and I think it will be problematic when we start supporting several deployments of Inspektor Gadget in parallel with #2208.

cc @burak-ok

burak-ok, line 102: Don't we need the execute bits set?

alban:
> What do you think about instead of running entrypoint and spawning gadgettracermgr, include what entrypoint does into gadgettracermgr and run it directly? Saves a bit of RAM that the entrypoint binary would take up otherwise.

I think it is a good idea to make the entrypoint a Go package, so it can potentially be used both by gadgettracermgr and ig. If it is available in ig, then we could make use of the crio hooks (and other hooks) with the [kubectl debug node](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/ig.md#using-ig-with-kubectl-debug-node) command with the `ghcr.io/inspektor-gadget/ig` container image. But that requires additional flags to the `ig` binary to configure the hooks (and probably go embed dependencies such as `prestart.sh`), so it does not have to be in this PR.

gadget-container/cleanup/cleanup.go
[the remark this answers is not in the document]
alban: They are not needed anymore:

- /host/etc/ld.so.preload has been removed in b8a5e22b02196e2c6278330488ee6b602a0aaed6
- /sys/fs/bpf/straceback/ was from the old traceloop implementation ([straceback-main-bpf.c](https://github.com/kinvolk/traceloop/blob/37aef5618b938896bcee7463551dabc4d6f73fab/bpf/straceback-main-bpf.c#L61))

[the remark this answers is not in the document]
alban: confirmed :)

gadget-container/entrypoint/entrypoint.go
```diff
@@ -0,0 +1,284 @@
...
+func installNRIHooks() error {
+	log.Info("Installing NRI hooks")
+
+	destinationPath := filepath.Join(host.HostRoot, "opt/nri/bin")
+	err := os.MkdirAll(destinationPath, 0o755)
+	if err != nil {
+		return fmt.Errorf("creating %s: %w", destinationPath, err)
+	}
+
+	err = copyFile(destinationPath, "/opt/hooks/nri/nrigadget", 0o640)
```
mqasimsarfraz, line 145: The gadget seems to be crashing when I am trying to use with `nri` hook (`kubectl set env -n gadget daemonset/gadget INSPEKTOR_GADGET_OPTION_HOOK_MODE=nri`): 

```
time="2023-11-08T14:38:52Z" level=info msg="INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER=true"
time="2023-11-08T14:38:52Z" level=info msg="Inspektor Gadget version: v0.22.0-15-g1d1e1a5d"
time="2023-11-08T14:38:52Z" level=info msg="Installing NRI hooks"
time="2023-11-08T14:38:52Z" level=fatal msg="installing NRI hooks: copying: writing /host/opt/nri/bin: open /host/opt/nri/bin: is a directory"
```

Probably `destinationPath` needs to contain file name?

gadget-container/entrypoint/entrypoint.go
```diff
@@ -0,0 +1,284 @@
...
+
+		path := filepath.Join("/opt/hooks/oci", file)
+		destinationPath := filepath.Join(host.HostRoot, path)
+		err := copyFile(destinationPath, path, 0o750)
+		if err != nil {
+			return fmt.Errorf("copying: %w", err)
+		}
+	}
+
+	for _, file := range []string{"etc/containers/oci/hooks.d", "usr/share/containers/oci/hooks.d/"} {
```
mqasimsarfraz, line 109: This is also failing for me on `crio` cluster:

```
time="2023-11-08T14:53:52Z" level=info msg="Installing OCI hooks configuration in /host/etc/containers/oci/hooks.d"
time="2023-11-08T14:53:52Z" level=warning msg="Couldn't install OCI hooks configuration"
time="2023-11-08T14:53:52Z" level=info msg="Installing OCI hooks configuration in /host/usr/share/containers/oci/hooks.d"
time="2023-11-08T14:53:52Z" level=warning msg="Couldn't install OCI hooks configuration"
```
I am assuming it similar to the issue below!

gadget-container/entrypoint/entrypoint.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	path := "/opt/hooks/nri/nrigadget"
+	destinationPath = filepath.Join(destinationPath, filepath.Base(path))
+	err = copyFile(destinationPath, path, 0o640)
+	if err != nil {
+		return fmt.Errorf("copying: %w", err)
+	}
+
+	configPath := "/opt/hooks/nri/conf.json"
```
mqasimsarfraz, line 153: I think this needs to be [/host/etc/nri/conf.json](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadget-container/entrypoint.sh#L96). I mean we should be appending our plugin only if host has existing config?

mqasimsarfraz (commented):
I tested and it works great. Thanks for handling all the feedback. Only one comment related to `nri` hook left and then it is ready to go in!

mqasimsarfraz:
I tested it and it works fine. Just needs to address small suggestion and afterwards we can merge it.

gadget-container/cleanup/cleanup.go
```diff
@@ -0,0 +1,94 @@
...
+
+	var configList nriv1.ConfigList
+	err = json.Unmarshal(content, &configList)
+	if err != nil {
+		return
+	}
+
+	length := len(configList.Plugins)
+	if length < 2 {
+		os.Remove(configPath)
```
mqasimsarfraz, line 62: nit: Not related to this PR but this check can be improved to ensure plugin type is `nrigadget`. While testing I found out if I switch from hook mode `auto` to `nri` the pod restart will remove the existing `conf.json` irrespective of the plugin.

mqasimsarfraz:
[ARO](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/6848457006/attempts/1#summary-18618861171) doesn't seem to be happy after this change. Perhaps we missed something related to `cri-o` hook?

PR #2210, opened 2023-11-08
Add SECURITY.md.

eiffel-fl:
It was inspired from that of moby.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -1,0 +1,21 @@
+---
+title: Reporting security issues
+weight: 90
+description: >
+  Instructions to report security issues found in Inspektor Gadget
+---
+
+# Reporting security issues
+
+As Inspektor Gadget maintainers, we take security seriously.
+If you discover a security issue, please bring it to our attention right away.
+
+## Reporting a Vulnerability
+
+You have two possibilities:
+1. Using the GitHub feature "Report a security vulnerability" when opening an issue.
+It will then create a non-public issue specific to security.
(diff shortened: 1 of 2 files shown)


mqasimsarfraz (approved):
LGTM! 

Do you think it would be a good idea to mention it in [contribution guidelines](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/CONTRIBUTING.md#contribution-guidelines) as well?

alban (commented):
The GitHub feature "Privately reporting a security vulnerability" is enabled on this repository.

Either we mention it in SECURITY.md, or we should disable that feature if we prefer to use emails.

image

alban:
There should be some changes in the `docs` directory, so that the documentation website gets updated:
https://inspektor-gadget.io/docs


Maybe just move SECURITY.md in the `docs` directory: according to the [GitHub documentation about security policies](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository), the `docs` directory is a valid location for GitHub to find SECURITY.md:
> you can add a SECURITY.md file to your repository's root, docs, or .github folder

docs/SECURITY.md
```diff
@@ -0,0 +1,14 @@
+# Reporting security issues
```
alban, line 1: You can add a [Hugo front matter](https://gohugo.io/content-management/front-matter/) so that Hugo orders the files in the desired order (with the `weight` parameter). See other markdown files.

PR #2211, opened 2023-11-08
Kubernetes resources: cleanup traceloop entries

alban:
The traceloop gadget no longer requires special Linux capabilities or special RBAC capabilities.

cc @flyth


alban:
Thanks! I'll wait a bit to see if @mqasimsarfraz or @flyth have concerns. If not, I'll merge it tomorrow.

flyth (approved):
Thanks for cleaning this up!

As said previously, we will probably need the Role/RoleBinding back when we move from the CRD to ConfigMaps, but we can reintroduce that when we get there.

PR #2215, opened 2023-11-09
go.mod: Use upstream cilium/ebpf.

eiffel-fl:
This fixes build errors when bumping golang.org/x/* packages [1].

[1]: https://github.com/cilium/ebpf/commit/36f6316fb169


alban (approved):
LGTM

Hopefully, `golang.org/x/sys` is released soon so we can stop referring to the commit.

Issue #2096, opened 2023-10-02
containerized-gadgets: Support running gadgets from private registries in Kubernetes


mauriciovasquezbernal:
We need to understand how to support pulling gadget images from private registries in the Kubernetes case. From a quick research, it seems we should use Kubernetes secrets and allow the run command to point to one of those. (as done in https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/ for pods)

```bash
$ kubectl create secret docker-registry regcred --docker-server=<your-registry-server> --docker-username=<your-name> --docker-password=<your-pword> --docker-email=<your-email>

... 

$ kubectl gadget run ghcr.io/foo/bar --image-pull-secrets=regcred
```

It's annoying for users to specify the secrets each time, so we also need to think about a "default" secret that is used when no secret is specified.

mqasimsarfraz:
How about handling it at deployment time:
```
$ kubectl gaget deploy --image-pull-secrets=regcred 
```
And when specified, we can just mount the credentials in the pod, [example](https://github.com/GoogleContainerTools/kaniko/tree/main#pushing-to-azure-container-registry).

mauriciovasquezbernal:
Makes sense to me. But I think we should also allow users to specify it in the `run` command in case they want to use a given credentials.

mqasimsarfraz:
Sounds good. I will pick this one up!

PR #2223, opened 2023-11-13
image-based-gadgets: Introduce gadget pull secret

mqasimsarfraz:
Currently there isn't a way to run gadgets from private registry in Kubernetes. This change mounts/symlinks the registry credentials to allow pulling gadgets from private registries. The flag `gadgetPullSecret` is inspired from [`imagePullSecrets`](https://kubernetes.io/docs/concepts/containers/images#specifying-imagepullsecrets-on-a-pod). 

Fixes: #2096 

### deploy

```bash
$ kubectl create ns gadget
# assuming you already did docker login qasimregistry.azurecr.io
$ kubectl create secret -n gadget docker-registry gadget-pull-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json
$ ./kubectl-gadget deploy --image-pull-policy Never
$ ./kubectl-gadget run qasimregistry.azurecr.io/gadget/trace_open:latest -A
```

### helm

```bash
$ kubectl create ns gadget
# assuming you already did docker login qasimregistry.azurecr.io
$ kubectl create secret -n gadget docker-registry gadget-pull-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json
$ make -C charts build
$ helm install -n gadget gadget --set config.mountPullSecret=true --set image.pullPolicy=Never --set config.experimental=true  ./charts/bin/gadget/
$ ./kubectl-gadget run qasimregistry.azurecr.io/gadget/trace_open:latest -A
```

### gadget

```bash
$ kubectl create secret -n gadget docker-registry gadget-pull-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json
$ ./kubectl-gadget run qasimregistry.azurecr.io/gadget/trace_open:latest  --pull-secret gadget-pull-secret -A
```

## Todo

- [x] Documentation
- [x] Allow specifying gadgetPullSecret at runtime. (Will be handled later on when https://github.com/inspektor-gadget/inspektor-gadget/issues/2225 is fixed)

### DIFF
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -58,6 +58,8 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
 )
 
+const gadgetPullSecret = "gadget-pull-secret"
+
 var deployCmd = &cobra.Command{
 	Use:          "deploy",
 	Short:        "Deploy Inspektor Gadget on the cluster",
@@ -395,6 +397,11 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 		return commonutils.WrapInErrSetupK8sClient(err)
 	}
 
+	var isPullSecretPresent bool
+	if _, err = k8sClient.CoreV1().Secrets(utils.GadgetNamespace).Get(context.TODO(), gadgetPullSecret, metav1.GetOptions{}); err == nil {
+		isPullSecretPresent = true
+	}
+
 	for _, object := range objects {
 		var currentGadgetDS *appsv1.DaemonSet
 
@@ -479,6 +486,29 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 			currentGadgetDS, _ = k8sClient.AppsV1().DaemonSets(utils.GadgetNamespace).Get(
 				context.TODO(), "gadget", metav1.GetOptions{},
 			)
+
+			// handle pull secret
+			if isPullSecretPresent {
+				daemonSet.Spec.Template.Spec.Volumes = append(daemonSet.Spec.Template.Spec.Volumes, v1.Volume{
+					Name: "pull-secret",
+					VolumeSource: v1.VolumeSource{
+						Secret: &v1.SecretVolumeSource{
+							SecretName: gadgetPullSecret,
+							Items: []v1.KeyToPath{
+								{
+									Key:  ".dockerconfigjson",
+									Path: "config.json",
+								},
+							},
+						},
+					},
+				})
+				gadgetContainer.VolumeMounts = append(gadgetContainer.VolumeMounts, v1.VolumeMount{
+					Name:      "pull-secret",
+					MountPath: "/var/run/secrets/gadget/pull-secret",
+					ReadOnly:  true,
+				})
+			}
 		}
 
 		if printOnly {
--- a/docs/gadgets/run.md
+++ b/docs/gadgets/run.md
@@ -57,6 +57,65 @@ ubuntu-hirsute         default                mypod2                 mypod2     
 ubuntu-hirsute         default                mypod2                 mypod2                 242164  cat         0        0        3   /dev/null
 ```
 
+### Private registries in Kubernetes
+
+In order to use private registries, you will need a [Kubernetes secret](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/) having credentials to access the registry.
+
+There are two different ways to use this support:
+
+#### Defining a default secret when deploying Inspektor Gadget
+
+This approach creates a secret that will be used by default when pulling the gadget images. It requires to have a `docker-registry` secret named `gadget-pull-secret` in the `gadget` namespace:
+
+Let's create the `gadget` namespace if it doesn't exist:
+
+```bash
+$ kubectl create namespace gadget
+```
+
+then create the secret:
+
+```bash
+$ kubectl create secret docker-registry gadget-pull-secret -n gadget --docker-server=MYSERVER --docker-username=MYUSERNAME --docker-password=MYPASSWORD
+```
+
+or you can create the secret from a file:
+
+```bash
+$ kubectl create secret docker-registry gadget-pull-secret -n gadget --from-file=.dockerconfigjson=$HOME/.docker/config.json
+```
+
+then, deploy Inspektor Gadget:
+
+```bash
+$ kubectl gadget deploy ...
+```
+
+this secret will be used by default when running a gadget:
+
+
+```bash
+$ kubectl gadget run myprivateregistry.io/trace_tcpconnect:latest
+```
+
+#### Specifying the secret when running a gadget
+
+It's possible to pass a secret each time a gadget is run, you'd need to follow a similar approach as above to create the secret:
+
+```bash
+# from credentials
+$ kubectl create secret docker-registry my-pull-secret -n gadget --docker-server=MYSERVER --docker-username=MYUSERNAME --docker-password=MYPASSWORD
+
+# from a file
+$ kubectl create secret docker-registry my-pull-secret -n gadget --from-file=.dockerconfigjson=$HOME/.docker/config.json
+```
+
+Then, it can be used each time a gadget is run:
+
+```bash
+$ kubectl gadget run myprivateregistry.io/trace_tcpconnect:latest --pull-secret my-pull-secret
+```
+
 ## With `ig`
 
 ``` bash
--- a/pkg/gadgets/run/tracer/run.go
+++ b/pkg/gadgets/run/tracer/run.go
@@ -47,6 +47,7 @@ const (
 	authfileParam         = "authfile"
 	insecureParam         = "insecure"
 	pullParam             = "pull"
+	pullSecret            = "pull-secret"
 )
 
 type GadgetDesc struct{}
@@ -104,6 +105,12 @@ func (g *GadgetDesc) ParamDescs() params.ParamDescs {
 			},
 			TypeHint: params.TypeString,
 		},
+		{
+			Key:         pullSecret,
+			Title:       "Pull secret",
+			Description: "Secret to use when pulling the gadget image",
+			TypeHint:    params.TypeString,
+		},
 	}
 }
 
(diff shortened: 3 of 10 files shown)


mauriciovasquezbernal:
I wondering about the UX of this:
1. Should we decouple the deploy operation from the secret creation? I mean, what if the secret is created after IG is deployed?

I think those should be decoupled, it can be confusing for users having to redeploying IG to take changes to the pull secret. It could be implemented by having a kubernetes watcher that updates the config.json file when the secret is created / updated.

2. Do we really need a `--gadget-pull-secret` flag? Can't we make IG look automatically for a secret with a hardcoded name in the gadget namespace?
3. I suppose the `kubectl gadget run foo --pull-secret=bar` feature will arrive later on, right?

mqasimsarfraz:
> I think those should be decoupled, it can be confusing for users having to redeploying IG to take changes to the pull secret. It could be implemented by having a kubernetes watcher that updates the config.json file when the secret is created / updated.

As per [secret documentation](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod), the volume should update automatically (unless it is a subPath, which is why I don't mount it directly at `/var/lib/ig/config.json`) if someone updates secret.

> When a volume contains data from a Secret, and that Secret is updated, Kubernetes tracks this and updates the data in the volume, using an eventually-consistent approach.

I think if I switch to symlink (`/var/run/secrets/gadget/gadget-pull-secret` -> `/var/lib/ig/config.json`) or having an env to override authfile path should do the trick. What do you say?

> Do we really need a --gadget-pull-secret flag? Can't we make IG look automatically for a secret with a hardcoded name in the gadget namespace?

Good idea. It would be better to go for a hardcoded name 

> I suppose the kubectl gadget run foo --pull-secret=bar feature will arrive later on, right?

Yes, Once we are done with: https://github.com/inspektor-gadget/inspektor-gadget/pull/2242

mauriciovasquezbernal:
> I think if I switch to symlink (`/var/run/secrets/gadget/gadget-pull-secret` -> `/var/lib/ig/config.json`) or having an env to override authfile path should do the trick. What do you say?

Right, symlink should work fine and keep the used secrets updated. 

How does it play with the `kubectl gadget run foo --pull-secret=bar` feature? I suppose in that case we need to contact the API server to retrieve the secret and pass it to the OCI registry somehow.

mqasimsarfraz:
> How does it play with the kubectl gadget run foo --pull-secret=bar feature? I suppose in that case we need to contact the API server to retrieve the secret and pass it to the OCI registry somehow.

Yep, that is my plan. I didn't complete the implementation yet but [experimented](https://github.com/inspektor-gadget/inspektor-gadget/compare/qasim/private-registry...mqasimsarfraz:inspektor-gadget:qasim/private-registry) with it a bit.

mqasimsarfraz:
@mauriciovasquezbernal thanks for the quick fix for #2242, I rebased/updated the PR to handle the case for `kubectl gadget run ...`. I switched to using `gadget-pull-secret` for all the cases and they seem to be working fine. What do you think about the UX now?

pkg/oci/oci.go
```diff
@@ -342,28 +350,48 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+			if err != nil {
+				return nil, fmt.Errorf("loading auth config: %w", err)
+			}
+		} else {
+			defer authFileReader.Close()
+			cfg, err = config.LoadFromReader(authFileReader)
+			if err != nil {
+				return nil, fmt.Errorf("loading auth config: %w", err)
+			}
 		}
```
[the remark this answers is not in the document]
mqasimsarfraz: I agree we could make it more readable but I was avoiding changing the way we check for falling back to docker auth like `en(authOptions.AuthFile) > 0` is always be true because of default path. How about following?

```
if authOptions.PullSecret {
	k8sClient, err := k8sutil.NewClientset("")
	if err != nil {
		return nil, fmt.Errorf("creating new k8s clientset: %w", err)
	}
	gps, err := k8sClient.CoreV1().Secrets("gadget").Get(context.TODO(), gadgetPullSecret, metav1.GetOptions{})
	if err != nil {
		return nil, fmt.Errorf("getting secret %q: %w", gadgetPullSecret, err)
	}
	if gps.Type != corev1.SecretTypeDockerConfigJson {
		return nil, fmt.Errorf("secret %q is not of type %q", gadgetPullSecret, corev1.SecretTypeDockerConfigJson)
	}

	cfg, err = config.LoadFromReader(bytes.NewReader(gps.Data[corev1.DockerConfigJsonKey]))
	if err != nil {
		return nil, fmt.Errorf("loading auth config: %w", err)
	}
} else if authFileReader, err := os.Open(authOptions.AuthFile); err != nil {
	// If the AuthFile was not set explicitly, we allow to fall back to the docker auth,
	// otherwise we fail to avoid masking an error from the user
	if !errors.Is(err, os.ErrNotExist) || authOptions.AuthFile != DefaultAuthFile {
		return nil, fmt.Errorf("opening auth file %q: %w", authOptions.AuthFile, err)
	}

	log.Debugf("Couldn't find default auth file %q...", authOptions.AuthFile)
	log.Debugf("Using default docker auth file instead")
	log.Debugf("$HOME: %q", os.Getenv("HOME"))

	cfg, err = config.Load("")
	if err != nil {
		return nil, fmt.Errorf("loading auth config: %w", err)
	}

} else {
	defer authFileReader.Close()
	cfg, err = config.LoadFromReader(authFileReader)
	if err != nil {
		return nil, fmt.Errorf("loading auth config: %w", err)
	}
}
```

burak-ok (commented):
Didn't look at all commits yet

cmd/kubectl-gadget/deploy.go
```diff
@@ -471,6 +478,29 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 			currentGadgetDS, _ = k8sClient.AppsV1().DaemonSets(utils.GadgetNamespace).Get(
 				context.TODO(), "gadget", metav1.GetOptions{},
 			)
+
+			// handle pull secret
+			if isPullSecretPresent {
```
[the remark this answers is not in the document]
mqasimsarfraz: > So adding a secret after deploying will fail?

What do we mean by 'fail' here? Like not able to use a secret after deploying gadget pods? As per [secret usage](https://kubernetes.io/docs/concepts/configuration/secret/#using-a-secret):

> Secret volume sources are validated to ensure that the specified object reference actually points to an object of type Secret. Therefore, a Secret needs to be created before any Pods that depend on it.

So we have to create it first or redeploy after creating the secret. Also, a user can just use `--pull-secret` with `run` to use the newly created secret!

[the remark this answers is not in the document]
burak-ok: Makes sense. I also didn't read the next commit yet where you wrote the documentation which statest that one has to redeploy :sweat_smile:

mqasimsarfraz:
> I suspect I am forgetting something.

It seems you found a bug, `az acr login` creates an entry with `identityToken` in `config.json`. We should be passing it down as [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/85d4bf05b58500c91eb11548e6b5850569cc0267). Let me update the PR.

Otherwise if you `docker login` or `ig login` with registry credentials it should work fine.

pkg/gadgets/run/tracer/run.go
```diff
@@ -104,6 +105,12 @@ func (g *GadgetDesc) ParamDescs() params.ParamDescs {
 			},
 			TypeHint: params.TypeString,
 		},
+		{
+			Key:         pullSecret,
+			Title:       "Pull secret",
+			Description: "Use the default (gadget-pull-secret) pull secret for the registry",
+			TypeHint:    params.TypeBool,
+		},
```
mauriciovasquezbernal, line 113: IMO it should be a string and let the user define any secret they want. The default secret is useful to avoid having to repeat it each time, but if users want to use a different secret for a specific gadget, we should support it too.

docs/gadgets/run.md
```diff
@@ -57,6 +57,45 @@ ubuntu-hirsute         default                mypod2                 mypod2
...
+```bash
+$ kubectl create secret docker-registry regcred -n gadget --from-file=.dockerconfigjson=$HOME/.docker/config.json
+```
+
+the secret can be used in following two ways while:
+
+#### running a gadget
+
+```bash
+$ kubectl gadget run ghcr.io/inspektor-gadget/gadget/trace_tcpconnect:latest --pull-secret
```
mauriciovasquezbernal, line 86: nit: it'll be better to use something like `myprivateregistry.io/trace_tcpconnect:latest` (just to be sure people understand it's about gadgets on private registries, but it's not needed for the gadgets we expose as they are public)

pkg/gadgets/run/tracer/run.go
mauriciovasquezbernal: btw, sorry if I'm asking too much, but I think we're almost there to have support for both modes, so why not to implement them now?

mqasimsarfraz: All good, I had my doubts about this already :D I agree we should allow using any custom secrets with `run` as it makes sense to specify any exiting secret at runtime. Remembering secret name across `deploy` can get tricky so makes sense to use a default secret.

mqasimsarfraz:
> I wanted to give it another try, but it will wait a bit more:
> 
> ```shell
> $ ./kubectl-gadget run                       (remotes/origin/qasim/private-registry) %
> INFO[0000] Experimental features enabled                
> Error: unknown command "run" for "kubectl-gadget"
> Run 'kubectl-gadget --help' for usage
> ```

Perhaps you need to sync the gadget information with `kubectl gadget sync`. Seems something to do with experimental features ?

mqasimsarfraz:
Thanks for the reviews!

pkg/gadgets/run/tracer/run.go
```diff
@@ -104,6 +105,12 @@ func (g *GadgetDesc) ParamDescs() params.ParamDescs {
 			},
 			TypeHint: params.TypeString,
 		},
+		{
+			Key:         pullSecret,
+			Title:       "Pull secret",
+			Description: "Secret to use when pulling the gadget image",
+			TypeHint:    params.TypeString,
+		},
```
mauriciovasquezbernal, line 113: I just realized this is also shown on the `ig` case, and it'll will cause ig trying to reach the api server. I think we need to fix that as in that case using a kubernetes secret doesn't make sense.

PR #2226, opened 2023-11-14
trace/capabilities: Modify cap_effective reading on kernel 6.3 and yo…

eiffel-fl:
…unger.

With kernel 6.3, kernel_cap_t was modified from a struct containing an array to a u64 [1].
We need to take this into account to be able to load the eBPF code.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f122a08b197d


matthyx (approved):

pkg/gadgets/trace/capabilities/tracer/bpf/capable.bpf.c
```diff
@@ -138,8 +138,12 @@ int BPF_KPROBE(ig_trace_cap_e, const struct cred *cred,
...
 		(u64)BPF_CORE_READ(task, real_cred, user_ns, ns.inum);
 	args.target_userns = (u64)BPF_CORE_READ(targ_ns, ns.inum);
-	BPF_CORE_READ_INTO(&args.cap_effective, task, real_cred,
-			   cap_effective.cap[0]);
+	if (LINUX_KERNEL_VERSION < KERNEL_VERSION(6, 3, 0))
+		BPF_CORE_READ_INTO(&args.cap_effective, task, real_cred,
+				   cap_effective.cap[0]);
+	else
+		BPF_CORE_READ_INTO(&args.cap_effective, task, real_cred,
+				   cap_effective);
```
mauriciovasquezbernal, line 146: Two ideas to avoid the kernel version check:
1. Use [`bpf_core_field_exists()`](https://nakryiko.com/posts/bpf-core-reference-guide/#bpf-core-field-exists) to check if there is a cap or val field in the `kernel_cap_t` struct. It'll probably require to create a ["CO-RE-relocatable type"](https://nakryiko.com/posts/bpf-core-reference-guide/#defining-own-co-re-relocatable-type-definitions), we already have few in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/include/gadget/core_fixes.bpf.h.
2. Since in both cases the struct is a wrapper around a 64bits integer, I think we can use `BPF_CORE_READ_INTO(&args.cap_effective, task, real_cred, cap_effective);` in all cases. 

I think the second solution is easier to implement, it'll work fine as far as `kernel_cap_t` continues to be a wrapper around an u64 integer, that according to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/diff/include/linux/capability.h?id=f122a08b197d should be the case, so I'll propose to go with this solution. Note: if you decide to go with this, please add a comment in the code with this reasoning.

mauriciovasquezbernal: Right, I was more concerned about this commit being backported making the kernel check useless.

mauriciovasquezbernal (approved):
LGTM from code inspection. Thanks for the quick fix.

PR #2228, opened 2023-11-15
pkg/deploy.yaml: Refine capabilities.

eiffel-fl:
Default capabilities are now dropped as they were forgotten in ed1c54db3e06 ("Remove privileged mode for running gadget container. "). We can also drop NET_ADMIN as we no more need it.

Sadly, we still need SYS_ADMIN for several operations, so we cannot replace it by using BPF and PERFMON on kernel where these capabilities are available.

### DIFF
--- a/charts/gadget/templates/daemonset.yaml
+++ b/charts/gadget/templates/daemonset.yaml
@@ -102,64 +102,57 @@ spec:
               type: "spc_t"
             capabilities:
               {{- if not .Values.capabilities }}
+              drop:
+                - ALL
               add:
-                # We need CAP_NET_ADMIN to be able to create BPF link.
-                # Indeed, link_create is called with prog->type which equals
-                # BPF_PROG_TYPE_CGROUP_SKB.
-                # This value is then checked in
-                # bpf_prog_attach_check_attach_type() which also checks if we have
-                # CAP_NET_ADMIN:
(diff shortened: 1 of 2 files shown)


charts/gadget/templates/daemonset.yaml
```diff
@@ -102,64 +102,57 @@ spec:
...
               add:
-                # We need CAP_NET_ADMIN to be able to create BPF link.
-                # Indeed, link_create is called with prog->type which equals
-                # BPF_PROG_TYPE_CGROUP_SKB.
-                # This value is then checked in
-                # bpf_prog_attach_check_attach_type() which also checks if we have
-                # CAP_NET_ADMIN:
-                # https://elixir.bootlin.com/linux/v5.14.14/source/kernel/bpf/syscall.c#L4099
-                # https://elixir.bootlin.com/linux/v5.14.14/source/kernel/bpf/syscall.c#L2967
-                - NET_ADMIN
```
mauriciovasquezbernal, line 114: Do you have any idea where is was needed before?

I suppose we'll need to add this back again once we support `BPF_PROG_TYPE_CGROUP_SKB` in containerized-gadgets.

mauriciovasquezbernal (approved):
Looks good to me once CI is green.

Issue #2225, opened 2023-11-14
cmd: flags for registry AuthOptions (--authfile/--insecure) doesn't seem to be working


mqasimsarfraz:
If you try to run a gadget that requires auth flags e.g running a gadget from an insecure registry it fails with the error:

```bash
→ go run -exec "sudo -E" ./cmd/ig run localhost:5000/gadget/trace_open:latest --insecure
INFO[0000] Experimental features enabled                
Error: getting gadget info: getting gadget image: pulling image "localhost:5000/gadget/trace_open:latest": downloading to local repository: failed to resolve localhost:5000/gadget/trace_open:latest: Get "https://localhost:5000/v2/gadget/trace_open/manifests/latest": http: server gave HTTP response to HTTPS client
```

It seems like the `--insecure` flag is not being respected. Also, I was trying to introduce a new flag (`--gadget-pull-secret`) in https://github.com/inspektor-gadget/inspektor-gadget/pull/2223 and it was getting ignored as well.

## Impact

Not able to use any registry auth options from CLI

## Environment and steps to reproduce

```bash
→ docker run --rm -d --name registry -p 5000:5000 registry:2
→ oras cp ghcr.io/inspektor-gadget/gadget/trace_open:latest localhost:5000/gadget/trace_open:latest
→ go run -exec "sudo -E" ./cmd/ig run localhost:5000/gadget/trace_open:latest --insecure
```

## Expected behavior

Be able to use registry auth options from CLI

## Additional information

From basic investigation it seems that the [gadgetParams](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/registry.go#L278) aren't ready when we try to pull image.

PR #2242, opened 2023-11-20
cmd: Fix --authfile/--insecure flags

mauriciovasquezbernal:
d1bccf6c124f ("gadgets/run: Support eBPF parameters") introduced a logic to remove the flags before calling GetGadgetInfo(), unfortunately it also broke the --authfile/--insecure flags as they were removed.

This commit changes the approach to use ParseEarlyFlags() to parse flags that are already known before contacting the server, like --insecure & --authfile, so they can be used by the server while it executes GetGadgetInfo().

Fixes: d1bccf6c124f ("gadgets/run: Support eBPF parameters")
Fix https://github.com/inspektor-gadget/inspektor-gadget/issues/2225

## How to test 

0. Push a gadget to a insecure registry. https://github.com/inspektor-gadget/inspektor-gadget/issues/2225 contains instructions for it. In my case I pushed it to 192.168.1.150:5000/foo:latest

I manually tested the following:

### ig 

```bash
sudo -E ig run 192.168.1.150:5000/foo:latest --insecure
RUNTIME.CONTAINERNAME                  PID                  PPID                 COMM             UID                  GID                 RETVAL
```

### ig + gadgectl using default unix socket

```bash
$ sudo -E ig daemon
INFO[0000] starting Inspektor Gadget daemon at "unix:///var/run/ig/ig.socket"
...

$ sudo -E ./gadgetctl run 192.168.1.150:5000/foo:latest 
RUNTIME.CONTAINERNAME                  PID                  PPID                 COMM             UID                  GID                 RETVAL
```

###  ig + gadgectl  using tcp 

```bash
$ sudo -E ig daemon -H tcp://127.0.0.1:1234
INFO[0000] starting Inspektor Gadget daemon at "tcp://127.0.0.1:1234"

...
$ ./gadgetctl run 192.168.1.150:5000/foo:latest --remote-address tcp://127.0.0.1:1234 --insecure
RUNTIME.CONTAINERNAME                  PID                  PPID                 COMM             UID                  GID                 RETVAL
```

### kubectl gadget 

```bash
$ ./kubectl-gadget run 192.168.1.150:5000/foo:latest --insecure -A 
K8S.NODE              K8S.NAMESPACE         K8S.POD               K8S.CONTAINER        PID         PPID        COMM        UID         GID         RETVAL
```

### TODO
- [X] Add integration tests for for --insecure

### DIFF
--- a/integration/inspektor-gadget/run_insecure_test.go
+++ b/integration/inspektor-gadget/run_insecure_test.go
@@ -1,0 +1,85 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/inspektor-gadget/inspektor-gadget/integration"
+)
+
+func TestRunInsecure(t *testing.T) {
+	ns := GenerateTestNamespaceName("test-run-insecure")
+
+	t.Parallel()
+
+	commandsPreTest := []*Command{
+		CreateTestNamespaceCommand(ns),
+		PodCommand("registry", "docker.io/library/registry:2", ns, "", ""),
+		WaitUntilPodReadyCommand(ns, "registry"),
+	}
+
+	RunTestSteps(commandsPreTest, t)
+
+	t.Cleanup(func() {
+		commands := []*Command{
+			DeleteTestNamespaceCommand(ns),
+		}
+		RunTestSteps(commands, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	registryIP := GetTestPodIP(t, ns, "registry")
+
+	// copy gadget image to insecure registry
+	orasCpCmds := []*Command{
+		{
+			Name: "Copy gadget image",
+			Cmd: fmt.Sprintf(`
+				kubectl apply -f - <<EOF
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: copier
+  namespace: %s
+spec:
+  template:
+    spec:
+      containers:
+      - name: copier
+        image: ghcr.io/oras-project/oras:v1.1.0
+        command:
+         - oras
+         - copy
+         - --to-plain-http
+         - %s/trace_open:%s
+         - %s:5000/trace_open:%s
+      restartPolicy: Never
+  backoffLimit: 4
+EOF`, ns, *gadgetRepository, *gadgetTag, registryIP, *gadgetTag),
+			ExpectedRegexp: "job.batch/copier created",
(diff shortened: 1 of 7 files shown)


mauriciovasquezbernal:
I added an integration test for the --insecure flag, unfortunately doing the same for the --authfile isn't so easy as it'll require to configure [native basic auth](https://distribution.github.io/distribution/about/deploying/#native-basic-auth) and TLS certificates in the registry. I'll open an issue to handle this later on.

mqasimsarfraz:
It seems that the `copier` pods keeps on restarting, perhaps we need pod restart policy `Never`. Maybe the failure is related to that? 🤔

mauriciovasquezbernal:
> It seems that the `copier` pods keeps on restarting, perhaps we need pod restart policy `Never`. Maybe the failure is related to that? 🤔

I made this change, kicked the CI few times and it seems to be working fine now. Thanks!

integration/inspektor-gadget/run_insecure_test.go
```diff
@@ -0,0 +1,65 @@
...
+		{
+			// Wait for the image to be available in the registry
+			// TODO: Ideally it shouldn't be a delay, but there is no an easy way to
+			// wait for a completed command using kubectl run
+			Cmd: "sleep 2",
+		},
+	}
+	RunTestSteps(orasCpCmds, t)
+
+	// TODO: Ideally it should not depend on a real gadget, but we don't have a "test gadget" available yet.
```
burak-ok, line 62: Created issue https://github.com/inspektor-gadget/inspektor-gadget/issues/2255 to track this.

integration/inspektor-gadget/run_insecure_test.go
```diff
@@ -0,0 +1,65 @@
...
+		{
+			Cmd: fmt.Sprintf("kubectl run -n %s copier --image ghcr.io/oras-project/oras:v1.1.0 --restart=Never -- copy --to-plain-http %s/trace_open:%s %s:5000/trace_open:%s",
+				ns, *gadgetRepository, *gadgetTag, registryIP, *gadgetTag),
+		},
+		WaitUntilPodReadyCommand(ns, "copier"),
+		{
+			// Wait for the image to be available in the registry
+			// TODO: Ideally it shouldn't be a delay, but there is no an easy way to
+			// wait for a completed command using kubectl run
+			Cmd: "sleep 2",
```
burak-ok, line 57: Couldn't we use something like `kubectl wait --for=Succeeded` ? (Ref for `Succeeded` https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase )

Or we add `--rm` to the `kubectl run` command and wait for the pod deletion`

burak-ok (approved):
LGTM
In other places we also have `sleep` commands but we could maybe change that

integration/inspektor-gadget/run_insecure_test.go
mauriciovasquezbernal: > Couldn't we use something like kubectl wait --for=Succeeded ?

Unfortunately it doesn't work. We can only wait on any of these conditions -> https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

I decided to change it and use a job, that's a better fit for this use case. Can you check it again please?

mqasimsarfraz (approved):
I tested it and it works fine, thanks!

integration/inspektor-gadget/run_insecure_test.go
burak-ok: See https://github.com/inspektor-gadget/inspektor-gadget/pull/2261

PR #2245, opened 2023-11-23
trace/open: Add missing extractor for flags

mauriciovasquezbernal:
TL;DR: We're missing the extractor for the flags and they were always
empty in the columns representation.

Before this commit:

```bash
$ sudo ig trace open -o columns=flags
FLAGS


// ^^^^ empty entries
```

After this commit:

```bash
$ sudo ig trace open -o columns=flags
FLAGS
O_RDONLY|O_CLOEXEC
O_RDONLY|O_CLOEXEC
O_RDONLY|O_CLOEXEC
O_RDONLY|O_CLOEXEC
O_RDONLY|O_CLOEXEC
```

Long story:

3c9916449a05 ("pkg: Add support for flags for trace open.") introduced
support for the flag column but it didn't define a custom extractor, at 
that time the column was formatted as:

```bash
sudo ig trace open -o columns=flags
FLAGS
[O_RDONLY O_CLOEXEC]
[O_RDONLY O_CLOEXEC]
[O_RDONLY O_CLOEXEC]
[O_RDONLY O_CLOEXEC]
[O_RDONLY O_CLOEXEC]
[O_RDONLY O_CLOEXEC]
```

Later on, 5ad73284af88 ("pkg/columns: add support for dynamic fields")
modified the columns library **not** to format slices to strings, hence
after this commit the column was printed as empty.

It's also possible to fix this on the columns library, but having a
custom extractor allows us to provide a better formatting to the user.

Fixes: 3c9916449a05 ("pkg: Add support for flags for trace open.")
Fixes: 5ad73284af88 ("pkg/columns: add support for dynamic fields")

### DIFF
--- a/pkg/gadgets/trace/open/types/types.go
+++ b/pkg/gadgets/trace/open/types/types.go
@@ -16,6 +16,7 @@ package types
 
 import (
 	"io/fs"
+	"strings"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
@@ -41,7 +42,13 @@ type Event struct {
 }
 
 func GetColumns() *columns.Columns[Event] {
-	return columns.MustCreateColumns[Event]()
+	openColumns := columns.MustCreateColumns[Event]()
+
+	openColumns.MustSetExtractor("flags", func(event *Event) any {
+		return strings.Join(event.Flags, "|")
+	})
+
+	return openColumns
 }
 
 func Base(ev eventtypes.Event) *Event {


burak-ok (approved):
LGTM
Thanks for adding this (back again?)

mauriciovasquezbernal:
> Do you know which commit introduced this bug?

It was an interesting investigation.

https://github.com/inspektor-gadget/inspektor-gadget/commit/3c9916449a050d98bbbc4e94c3d3705b59ccc488 introduced support for flags, but didn't define a custom extractor, then https://github.com/inspektor-gadget/inspektor-gadget/commit/5ad73284af88037f442c8081a90670d82a94108c changed the columns library in a way that complex types (slices, arrays) weren't formatted to strings anymore. 

I tried the following diff and it fixed the issue, but in my opinion the best solution is to have a custom extractor:

```diff
$ git diff
diff --git pkg/columns/columns.go pkg/columns/columns.go
index 7f759227..d1decec0 100644
--- pkg/columns/columns.go
+++ pkg/columns/columns.go
@@ -458,6 +458,7 @@ func (c *Columns[T]) MustSetExtractor(columnName string, extractor func(*T) stri
 type ColumnInternals interface {
        getOffset() uintptr
        getSubFields() []subField
+       GetRef(v reflect.Value) reflect.Value
        IsVirtual() bool
        HasCustomExtractor() bool
 }
@@ -606,7 +607,9 @@ func GetFieldAsStringExt[T any](column ColumnInternals, floatFormat byte, floatP
                }
        }
        return func(entry *T) string {
-               return ""
+               entryValue := reflect.ValueOf(entry)
+               field := column.GetRef(entryValue)
+               return fmt.Sprintf("%v", field.Interface())
        }
 }
```

@flyth do you think it's a good change to be able to format any type to string when a custom extractor isn't defined?

I updated the commit message and PR description.

PR #2257, opened 2023-11-28
gadgets/run: Print string representation of enums

eiffel-fl:
Hi.

This PR prints the string representation of enums.
Note that, it works only on named enum.

You can test it with the following patch:

```patch
diff --git a/gadgets/snapshot_socket/program.bpf.c b/gadgets/snapshot_socket/program.bpf.c
index cf790793..99ab7af9 100644
--- a/gadgets/snapshot_socket/program.bpf.c
+++ b/gadgets/snapshot_socket/program.bpf.c
@@ -49,7 +49,7 @@
 struct socket_entry {
        struct gadget_l4endpoint_t src;
        struct gadget_l4endpoint_t dst;
-       __u32 state;
+       enum tcp_conntrack state;
        __u32 ino;
 };
```

```bash
$ sudo -E ./ig run sockets -o columns=state,state_raw        francis/enum-string *% u=
STATE                                                                STATE_RAW                                                           
TCP_CONNTRACK_TIME_WAIT                                              7                                                                   
TCP_CONNTRACK_TIME_WAIT                                              7                                                                   
TCP_CONNTRACK_MAX                                                    10                                                                  
TCP_CONNTRACK_MAX                                                    10                                                                  
TCP_CONNTRACK_MAX                                                    10                                                                  
TCP_CONNTRACK_MAX                                                    10
```

Best regards.

### DIFF
--- a/pkg/gadgets/run/tracer/run.go
+++ b/pkg/gadgets/run/tracer/run.go
@@ -354,6 +354,30 @@ func getSimpleType(typ btf.Type) reflect.Type {
 	case *btf.Typedef:
 		typ, _ := getUnderlyingType(typedMember)
 		return getSimpleType(typ)
+	case *btf.Enum:
+		if typedMember.Signed {
+			switch typedMember.Size {
+			case 1:
+				return reflect.TypeOf(int8(0))
+			case 2:
+				return reflect.TypeOf(int16(0))
+			case 4:
+				return reflect.TypeOf(int32(0))
+			case 8:
+				return reflect.TypeOf(int64(0))
+			}
+		}
+
+		switch typedMember.Size {
+		case 1:
+			return reflect.TypeOf(uint8(0))
+		case 2:
+			return reflect.TypeOf(uint16(0))
+		case 4:
+			return reflect.TypeOf(uint32(0))
+		case 8:
+			return reflect.TypeOf(uint64(0))
+		}
 	}
 
 	return nil
@@ -492,6 +516,13 @@ func (g *GadgetDesc) getColumns(info *types.GadgetInfo) (*columns.Columns[types.
 	l4endpointCounter := 0
 	timestampsCounter := 0
 
+	type enumDef struct {
+		btfEnum *btf.Enum
+		attrs   columns.Attributes
+	}
+
+	enums := []enumDef{}
+
 	for i, field := range eventStruct.Fields {
 		member := members[field.Name]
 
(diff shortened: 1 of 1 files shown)


pkg/gadgets/run/tracer/run.go
```diff
@@ -527,6 +558,20 @@ func (g *GadgetDesc) getColumns(info *types.GadgetInfo) (*columns.Columns[types.
...
+			}
+
+			// Add raw columns for enum. Below a column that print the string
+			// representation is added as well.
+			enums = append(enums, enumDef{
+				btfEnum: enum,
+				attrs:   attrs,
+			})
+			attrs.Name = member.Name + "_raw"
+			attrs.Visible = false
```
mauriciovasquezbernal, line 574: I think it's better to move this outside this switch case. It seems hacky to me to use the default case above for this purpose.

pkg/gadgets/run/tracer/run.go
```diff
@@ -550,6 +595,49 @@ func (g *GadgetDesc) getColumns(info *types.GadgetInfo) (*columns.Columns[types.
...
+			if ev == nil || len(ev.RawData) == 0 {
+				return ""
+			}
+			val := getter(ev)
+
+			enum := &btf.Enum{}
+			err = btfSpec.TypeByName(e.btfEnum.Name, &enum)
+			if err != nil {
+				return "NOT_FOUND"
+			}
```
mauriciovasquezbernal, line 629: This works fine if the enum involved is defined in the kernel, but what about if this is an enum defined by the user?

mauriciovasquezbernal: Also, I think this logic can be moved outside this function to avoid calling it each time the conversion is done.

mauriciovasquezbernal: Finally, what about adding a comment explaining why we use the kernel type to get the string representation?

mauriciovasquezbernal: > For the moment, I kept this code in this function as I do not see where to put it.

I meant, to move the `btfSpec.TypeByName`  call outside the anonymous functions that is passed to `MustAddColumn`. That logic can be done just once instead of repeating it each time the extractor is called.

pkg/gadgets/run/tracer/run.go
```diff
@@ -577,6 +620,51 @@ func (g *GadgetDesc) getColumns(info *types.GadgetInfo) (*columns.Columns[types.
...
+		var getter func(*types.Event) uint64
+		if e.btfEnum.Signed {
+			getterT := columns.GetFieldAsNumberFunc[int64, types.Event](c)
+			getter = func(ev *types.Event) uint64 {
+				return uint64(getterT(ev))
+			}
+		} else {
+			getter = columns.GetFieldAsNumberFunc[uint64, types.Event](c)
+		}
+		cols.MustAddColumn(e.attrs, func(ev *types.Event) any {
```
mauriciovasquezbernal, line 644: I'd rather use `AddColumn` and handle the returned error.

mauriciovasquezbernal (approved):
LGTM. Thanks for working on this.

PR #2259, opened 2023-11-28
remove arm64 workaround in exec tracer

matthyx:

## How to use

## Testing done


matthyx:
> ```diff
> +//go:generate go run github.com/cilium/ebpf/cmd/
> ```

the extra architectures were not present before... I didn't know if they were safe to add
good catch for `tracer.go` I'll fix both

matthyx:
> ebpf-objects

fixed, sorry

PR #2204, opened 2023-11-06
Handling columns in wasm module

alban:
# Handling columns in wasm module

Part of https://github.com/inspektor-gadget/inspektor-gadget/issues/2202

## How to use

```
$ go run -exec 'sudo -E' ./cmd/ig/... image build -t ghcr.io/inspektor-gadget/gadget/trace_dns:latest gadgets/trace_dns/
INFO[0000] Experimental features enabled                
Successfully built ghcr.io/inspektor-gadget/gadget/trace_dns:latest@sha256:4513df2dfd78bde41fee7f46dc35db2f0923d5a266002ef2ab920ca486e246d3

$ go run -exec 'sudo -E' ./cmd/ig/... run ghcr.io/inspektor-gadget/gadget/trace_dns:latest
INFO[0000] Experimental features enabled                
RUNTIME.CONTAINERNAME  SRC               DST               PID     TASK  NAME            QR  PKT_TYPE RCODE    QTYPE
gifted_torvalds        172.17.0.4:48815  192.168.0.1:53    165520  ping  wikipedia.org.  Q   OUTGOING NoError  A
gifted_torvalds        172.17.0.4:48815  192.168.0.1:53    165520  ping  wikipedia.org.  Q   OUTGOING NoError  AAAA
gifted_torvalds        192.168.0.1:53    172.17.0.4:48815  165520  ping  wikipedia.org.  R   HOST     NoError  A
gifted_torvalds        192.168.0.1:53    172.17.0.4:48815  165520  ping  wikipedia.org.  R   HOST     NoError  AAAA
```

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

PR #2269, opened 2023-12-04
Avoid decoding BTF on the client

mauriciovasquezbernal:
This commit changes the logic to avoid the client having to decode BTF information. The whole idea consists on having the "columns descriptors" in the GadgetInfo structure. These descriptors are generated by the server and used by the client to decode the data it receives.

The data is sent as a binary blob [][]byte. The first array is the bpf event, the second one are fixed type fields and the other ones are used for strings. This commit adds some helper methods to add fields to the event.

This commit is an intermediate step in order to enable support for wasm. The next work is to start using it from operators to decouple them from the tracer implementation.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/2131

I rebased #2204 on top of this in https://github.com/inspektor-gadget/inspektor-gadget/tree/mauricio/experiments/alban_wasm2 and it seems to work fine.

/cc @alban @flyth

### DIFF
--- a/pkg/gadgets/run/tracer/run.go
+++ b/pkg/gadgets/run/tracer/run.go
@@ -188,6 +188,12 @@ func getGadgetInfo(params *params.Params, args []string, secretBytes []byte, log
 		return nil, err
 	}
 
+	ret.EventFactory = types.NewEventFactory()
+	ret.Columns, err = calculateColumnsForClient(ret.EventFactory, ret.GadgetMetadata, gadget.EbpfObject, logger)
+	if err != nil {
+		return nil, err
+	}
+
 	return ret, nil
 }
 
@@ -300,84 +306,47 @@ func loadSpec(progContent []byte) (*ebpf.CollectionSpec, error) {
 	return spec, err
 }
 
-func getType(typ btf.Type) reflect.Type {
-	switch typedMember := typ.(type) {
-	case *btf.Array:
-		arrType := getSimpleType(typedMember.Type)
+func reflectTypeFromType(typ types.Type) reflect.Type {
+	if typ.Kind == types.KindArray {
+		if typ.ArrayType == nil {
+			return nil
+		}
+		arrType := reflectTypeFromKind(typ.ArrayType.Kind)
 		if arrType == nil {
 			return nil
 		}
-		return reflect.ArrayOf(int(typedMember.Nelems), arrType)
-	default:
-		return getSimpleType(typ)
+		return reflect.ArrayOf(typ.ArrayNElements, arrType)
 	}
+
+	return reflectTypeFromKind(typ.Kind)
 }
 
-func getSimpleType(typ btf.Type) reflect.Type {
-	switch typedMember := typ.(type) {
-	case *btf.Int:
-		switch typedMember.Encoding {
-		case btf.Signed:
-			switch typedMember.Size {
-			case 1:
-				return reflect.TypeOf(int8(0))
-			case 2:
-				return reflect.TypeOf(int16(0))
-			case 4:
-				return reflect.TypeOf(int32(0))
-			case 8:
-				return reflect.TypeOf(int64(0))
-			}
-		case btf.Unsigned:
-			switch typedMember.Size {
-			case 1:
-				return reflect.TypeOf(uint8(0))
-			case 2:
-				return reflect.TypeOf(uint16(0))
-			case 4:
-				return reflect.TypeOf(uint32(0))
-			case 8:
-				return reflect.TypeOf(uint64(0))
-			}
-		case btf.Bool:
-			return reflect.TypeOf(bool(false))
-		case btf.Char:
-			return reflect.TypeOf(uint8(0))
-		}
-	case *btf.Float:
-		switch typedMember.Size {
-		case 4:
-			return reflect.TypeOf(float32(0))
-		case 8:
-			return reflect.TypeOf(float64(0))
-		}
-	case *btf.Typedef:
-		typ, _ := getUnderlyingType(typedMember)
-		return getSimpleType(typ)
-	case *btf.Enum:
-		if typedMember.Signed {
-			switch typedMember.Size {
-			case 1:
-				return reflect.TypeOf(int8(0))
-			case 2:
-				return reflect.TypeOf(int16(0))
-			case 4:
-				return reflect.TypeOf(int32(0))
-			case 8:
-				return reflect.TypeOf(int64(0))
-			}
-		}
-
-		switch typedMember.Size {
-		case 1:
-			return reflect.TypeOf(uint8(0))
-		case 2:
-			return reflect.TypeOf(uint16(0))
-		case 4:
-			return reflect.TypeOf(uint32(0))
-		case 8:
-			return reflect.TypeOf(uint64(0))
-		}
+func reflectTypeFromKind(kind types.Kind) reflect.Type {
+	switch kind {
+	case types.KindInt8:
+		return reflect.TypeOf(int8(0))
+	case types.KindInt16:
+		return reflect.TypeOf(int16(0))
+	case types.KindInt32:
+		return reflect.TypeOf(int32(0))
+	case types.KindInt64:
+		return reflect.TypeOf(int64(0))
+	case types.KindUint8:
+		return reflect.TypeOf(uint8(0))
+	case types.KindUint16:
+		return reflect.TypeOf(uint16(0))
+	case types.KindUint32:
+		return reflect.TypeOf(uint32(0))
+	case types.KindUint64:
+		return reflect.TypeOf(uint64(0))
+	case types.KindFloat32:
+		return reflect.TypeOf(float32(0))
+	case types.KindFloat64:
+		return reflect.TypeOf(float64(0))
+	case types.KindBool:
+		return reflect.TypeOf(false)
+	case types.KindString:
+		return reflect.TypeOf("")
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


pkg/gadgets/run/types/factory.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"reflect"
+	"unsafe"
+
+	"golang.org/x/exp/constraints"
+)
+
+type EventFactory struct {
+	lastOffset uintptr
+	lastIndex  int
```
burak-ok, line 26: The name is confusing me. Isn't `nextIndex`/`nextFreeIndex` here better? In the following code the usage seems to be that this indicates exactly that.

Or change the usage that this is actually the last valid index

pkg/gadgets/run/types/factory.go
```diff
@@ -0,0 +1,128 @@
...
+	ev.Blob[IndexFixed] = make([]byte, f.lastOffset)
+
+	return ev
+}
+
+type FieldType interface {
+	constraints.Integer | constraints.Float | bool
+}
+
+func AddField[T FieldType](f *EventFactory, name string) ColumnDesc {
```
burak-ok, line 51: I still hate that member methods can not be generic and the upstream discussion about this is still going with no end in sight :/

pkg/gadgets/run/types/factory.go
```diff
@@ -0,0 +1,128 @@
...
+	f.lastOffset += typ.Size()
+
+	f.setters[name] = func(ev *Event, v T) {
+		*(*T)(unsafe.Pointer(&ev.Blob[IndexFixed][offset])) = v
+	}
+
+	return col
+}
+
+func (f *EventFactory) AddString(name string) ColumnDesc {
```
burak-ok, line 75: I don't even know if we should make `AddString` a member method but `AddField` not (because of Go limitations).
IMO it should be the same to be streamlined

pkg/gadgets/run/tracer/run.go
```diff
@@ -754,3 +680,185 @@ func (g *GadgetDesc) EventPrototype() any {
...
+	}
+
+	columns := []types.ColumnDesc{}
+
+	for _, member := range eventType.Members {
+		member := member
+
+		_, ok := colNames[member.Name]
+		if !ok {
+			continue
```
burak-ok, line 800: Maybe add debug message that that member was skipped because it is not in the gadget metadata

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -401,13 +411,80 @@ func (t *Tracer) processEventFunc(gadgetCtx gadgets.GadgetContext) func(data []b
 			}
 			timestampsOffsets = append(timestampsOffsets, member.Offset.Bytes())
 		}
+
+		btfSpec, err := btf.LoadKernelSpec()
+		if err != nil {
+			logger.Warnf("Kernel BTF information not available. Enums won't be solved")
```
burak-ok, line 417:
```suggestion
			logger.Warnf("Kernel BTF information not available. Enums won't be resolved to strings")
```

burak-ok (commented):
Didn't run it yet. Only static analysis

pkg/gadgets/run/types/factory.go
mauriciovasquezbernal: Makes sense. Changed it to be a function.

pkg/gadgets/run/types/types.go
```diff
@@ -50,8 +56,11 @@ type Event struct {
 	MountNsID uint64 `json:"-"`
 	NetNsID   uint64 `json:"-"`
 
-	// Raw event sent by the ebpf program
-	RawData []byte `json:"raw_data,omitempty"`
+	// Blob is used to save data to be sent to the client.
+	// [0] is used for bpf event
+	// [1] is used for fixed-size members
+	// [1+] is used for variable size members
+	Blob [][]byte `json:"blob,omitempty"`
```
alban, line 63: I find the indexing system (IndexVirtual, IndexEBPF, IndexFixed) confusing. Would the following make the code clearer?

```suggestion
	// Blobs used to save data to be sent to the client.
	BpfEvent []byte `json:"bpfEvent,omitempty"`
	FixedSizedMembers []byte `json:"fixedSizeMembers,omitempty"`
	VariableSizedMembers [][]byte `json:"fixedSizeMembers,omitempty"`
```

pkg/gadgets/run/types/types.go
```diff
@@ -85,10 +94,56 @@ type Printer interface {
 	Logf(severity logger.Level, fmt string, params ...any)
 }
 
+// Type represents the type of a field.
+
```
alban, line 98:
```suggestion
// Kind represents the type of a field.
```
(without empty line)

Or, is this comment at the wrong place?

pkg/gadgets/run/types/types.go
```diff
@@ -85,10 +94,56 @@ type Printer interface {
...
+type Kind uint8
+
+// TODO: use reflect.Kind? (once we resolve TODO below)
+const (
+	KindNone = iota
```
alban, line 103:
```suggestion
	KindNone Kind = iota
```

pkg/gadgets/run/types/types.go
```diff
@@ -85,10 +94,56 @@ type Printer interface {
...
+	// Fields only for KindArray
+	ArrayNElements int // number of elements in the array
+	ArrayType      *Type
+}
+
+// ColumnDesc describes how a column is built. It's basically a serializable version of
+// columns.DynamicField
+type ColumnDesc struct {
+	Name      string
+	BlobIndex int // -1: virtual, 0: ebpf, 1: fixed length, 1+ strings
```
alban, line 136: Would it make it clearer to split this into two fields?

```suggestion
	BlobType BlobType // BlobType is an enum defined with iotas
	StringIndex int // only used when BlobType == BlobTypeString
```

pkg/gadgets/run/types/factory.go
```diff
@@ -0,0 +1,129 @@
...
+			Kind: reflectTypeToKind(typ),
+		},
+		Offset:    offset,
+		BlobIndex: IndexFixed,
+	}
+
+	f.nextOffset += typ.Size()
+
+	f.setters[name] = func(ev *Event, v T) {
+		*(*T)(unsafe.Pointer(&ev.Blob[IndexFixed][offset])) = v
```
alban, line 70: To support different endianness between the client and the server:
```
binary.BigEndian.PutUint64(ev.Blob[IndexFixed][offset:offset+8], (uint64) v)
```

And fix `f.nextOffset += 8`.

Alternatively, use the correct method PutUint16, PutUint32, PutUint64 etc. (even if it does not play as nice with generics)

I wonder if it would be easier to define `FixedSizedMembers [][]byte` (i.e. each fixed-sized member would have its own array)

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -329,6 +333,10 @@ func verifyGadgetUint64Typedef(t btf.Type) error {
 	return nil
 }
 
+func getAsInteger[OT constraints.Integer](data []byte, offset uint32) OT {
+	return *(*OT)(unsafe.Pointer(&data[offset]))
```
alban, line 337: Ditto to support different endianness:
```
return binary.BigEndian.Uint16(data[offset:xxx])
```

pkg/gadgets/run/types/factory.go
```diff
@@ -0,0 +1,129 @@
...
+
+	col := ColumnDesc{
+		Name: name,
+		Type: Type{
+			Kind: KindString,
+		},
+		BlobIndex: index,
+	}
+
+	f.nextIndex++
```
alban, line 87: These methods `FactoryAddField` and `FactoryAddString` are difficult to use because they need to be called in the exact same order between `calculateColumnsForClient` and `processEventFunc`. We have to be careful to iterate on the same object. In my previous code, I initially had problems because one was iterating over the btf fields, and the other was iterating over the types defined in the metadata yaml (and they were not in the right order).

So it makes me wonder if it's better to use hash maps keyed by the field name (`map[string][]byte`) instead of arrays (`[][]byte`).

alban (commented):
Thanks!

pkg/gadgets/run/types/types.go
```diff
@@ -85,10 +94,56 @@ type Printer interface {
...
+	Type      Type
+	Offset    uintptr
+}
+
 type GadgetInfo struct {
 	GadgetMetadata *GadgetMetadata
+	Columns        []ColumnDesc
 	ProgContent    []byte
 	GadgetType     gadgets.GadgetType
+	EventFactory   *EventFactory
```
alban, line 146: Since the `type GadgetInfo` is used in the [GetGadgetInfo gRPC method](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadget-service/api/api.proto#L94), I think there should be some documentation about the that in:

https://inspektor-gadget.io/docs/latest/reference/grpc/

Even if it is just a few sentences explaining how the json is encoded in the gRPC field. For now, the source says:
```
message GetGadgetInfoResponse {
  bytes info = 1; // encoded in json
}
```
But it does not say what is encoded in json and how to find the Golang struct that is marshalled into json.

pkg/gadgets/run/types/types.go
mauriciovasquezbernal: I can't remember, probably it was intended for `Type struct`. Anyway, this comment doesn't add so much information, will remove it.

pkg/gadgets/run/types/types.go
mauriciovasquezbernal: Added a commit with few sentences about it.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -401,13 +411,80 @@ func (t *Tracer) processEventFunc(gadgetCtx gadgets.GadgetContext) func(data []b
...
+				}
+			case types.KindInt64:
+				getter = func(data []byte) uint64 {
+					return uint64(getAsInteger[int64](data, member.Offset.Bytes()))
+				}
+			}
+
+			fieldSetter := types.GetSetter[string](t.eventFactory, member.Name)
+			enumSetter := func(ev *types.Event, data []byte) {
+				val := getter(data)
```
alban, line 468: What happens if `getter` is `nil`?

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -401,13 +411,80 @@ func (t *Tracer) processEventFunc(gadgetCtx gadgets.GadgetContext) func(data []b
...
+				kernelEnum := &btf.Enum{}
+				if err = btfSpec.TypeByName(enum.Name, &kernelEnum); err == nil {
+					// Use kernel enum if found
+					enum = kernelEnum
+				}
+			}
+
+			var getter func(data []byte) uint64
+			typ := simpleTypeFromBTF(member.Type)
+			switch typ.Kind {
```
alban, line 431: Can `simpleTypeFromBTF` return `nil`? In this case, reading `typ.Kind` would panic.

alban (approved):
Thanks @mauriciovasquezbernal.

I have 2 remaining questions about nil-dereferencement. After that, it can be merged. The rest of the comments can be handled later.

pkg/gadgets/run/tracer/tracer.go
mauriciovasquezbernal: I don't think it can return `nil`, it's called only for enums and all possibilities are handled in https://github.com/inspektor-gadget/inspektor-gadget/blob/2cf1a9e0cb47f8e627c6ab898e9b233922850bc0/pkg/gadgets/run/tracer/run.go#L746-L770

pkg/gadgets/run/tracer/tracer.go
mauriciovasquezbernal: Same as in the other comment, the switch statement is handling all possible values, so `getter` won't never be nil.

pkg/gadgets/run/tracer/tracer.go
alban: I think `typedMember.Size` is unmarshalled from the ELF file (from the BTF section) and I am not sure if the cilium/ebpf package checks that the size of an enum is one of those values 1,2,4,8. I tried to find that check but I didn't find it.

https://github.com/cilium/ebpf/blob/main/btf/btf_types.go#L165


It should be fine for valid ebpf/ELF files, but we should ensure IG does not panic for invalid or malicious content.

mauriciovasquezbernal: Makes sense. Added a check to verify this condition.

Issue #2268, opened 2023-12-04
Regression: 


matthyx:
## Description

DNS events are not enriched after upgrading to v0.23.0 on kernels without `socket_file_ops`.

## Impact

Business logic relying on `k8scontainerID` for DNS events don't work in my application.

## Environment and steps to reproduce

1. **Set-up**: use latest minikube v1.32.0
2. **Task**: use an application using the DNS tracer (such as https://github.com/kubescape/node-agent)
3. **Action(s)**:
  a. build it with v0.23.0
  b. see empty events
4. **Error**: `time="2023-12-04T16:58:46Z" level=error msg="creating socket enricher: updating socket_file_ops address with ksyms: file does not exist"`

## Expected behavior

No error and full events.

## Additional information

Looks like it was introduced by commit https://github.com/inspektor-gadget/inspektor-gadget/commit/f0884013d6d4ea75dc384ac285c06b195c19ba4a

alban:
Before commit https://github.com/inspektor-gadget/inspektor-gadget/commit/f0884013d6d4ea75dc384ac285c06b195c19ba4a, when the kernel does not provide the address of [socket_file_ops](https://github.com/torvalds/linux/blob/v6.5/net/socket.c#L152) in `/proc/kallsyms`, the socket enricher could still work in degraded mode (without enriching initial sockets but enriching subsequent sockets correctly) but after that patch, it fails completely.

In the case of the trace-dns gadget, we don't care so much about long-running sockets (we just care about single query/response UDP packets). So I think it is better to let the socket enricher work in degraded mode instead of disabling it completely.

I would like to know why `socket_file_ops` is in kallsyms in some kernels and not in others. This variable didn't change in Linux in the last 19 years. Is it because it is defined as a `static const`, so the compiler is free to include it or skip it from kallsyms?

To get the value of `socket_file_ops`  without using kallsyms, we could use something like [privatedata.go#L33-L34](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-hook/privatedata.go#L33-L34). This was done to read `(struct file *)->privata_data` but it can be adapted to read `(struct file *)->f_op`.

PR #2270, opened 2023-12-05
socketenricher: Print warning rather than error when problem with kal…

eiffel-fl:
…lsyms.

Fixes: f0884013d6d4 ("socketenricher: Catch SpecUpdateAddresses() error.")

### DIFF
--- a/pkg/gadgets/internal/socketenricher/tracer.go
+++ b/pkg/gadgets/internal/socketenricher/tracer.go
@@ -74,7 +74,9 @@ func (se *SocketEnricher) start() error {
 
 	err = kallsyms.SpecUpdateAddresses(specIter, []string{"socket_file_ops"})
 	if err != nil {
-		return fmt.Errorf("updating socket_file_ops address with ksyms: %w", err)
+		// Being unable to access to /proc/kallsyms can be caused by not having
+		// CAP_SYSLOG.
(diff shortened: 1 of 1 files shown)


pkg/gadgets/internal/socketenricher/tracer.go
```diff
@@ -74,7 +74,7 @@ func (se *SocketEnricher) start() error {
 
 	err = kallsyms.SpecUpdateAddresses(specIter, []string{"socket_file_ops"})
 	if err != nil {
-		return fmt.Errorf("updating socket_file_ops address with ksyms: %w", err)
+		log.Warnf("updating socket_file_ops address with ksyms: %v\nThis may result in network related information missing. Possible cause can be lack access of /proc/kallsyms due to CAP_SYSLOG not being set.", err)
```
alban, line 77: This can also be due to a kernel that does not provide the symbol `socket_file_ops` in `/proc/kallsyms` because the constant is declared as `static`.

That's a too long explanation to add in the `log.Warnf`. But you can add a comment before.

alban (approved):
Thanks.
1 suggested comment, LGTM after that.

alban (approved):
LGTM (once the CI is green)

alban:
Issue #2268

PR #2272, opened 2023-12-06
gadgets: Add trace mount.

eiffel-fl:


gadgets/trace_mount/gadget.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - name: mount_ns_id
+      description: 'Mount namespace inode id'
+      attributes:
+        template: ns
+    - name: ret
+      description: 'TODO: Fill field description'
+      attributes:
+        width: 16
+        alignment: left
+        ellipsis: end
```
mauriciovasquezbernal, line 28: I suppose 16 is too much for this return value. Perhaps 3 is a better width?

gadgets/trace_mount/gadget.yaml
```diff
@@ -0,0 +1,49 @@
...
+        template: comm
+    - name: src
+      attributes:
+        template: comm
+    - name: dest
+      attributes:
+        template: comm
+    - name: data
+      attributes:
+        template: comm
```
mauriciovasquezbernal, line 43: Why are these using `template:comm`?

gadgets/trace_mount/program.bpf.c
```diff
@@ -0,0 +1,194 @@
...
+	struct event *eventp;
+	int zero = 0;
+
+	argp = bpf_map_lookup_elem(&args, &tid);
+	if (!argp)
+		return 0;
+
+	eventp = bpf_map_lookup_elem(&heap, &zero);
+	if (!eventp)
+		return 0;
```
mauriciovasquezbernal, line 122: As commented on the other PR. this should cleanup the entry on the `args` map.

Another solution is to swap the lookups, first lookup on `heap` (it'll never fail and is very fast), then lookup on args.

gadgets/trace_mount/program.bpf.c
```diff
@@ -0,0 +1,194 @@
...
+	__type(key, int);
+	__type(value, struct event);
+} heap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} events SEC(".maps");
+
```
mauriciovasquezbernal, line 73: Please use `GADGET_TRACER` as suggested in https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.23.1/docs/devel/hello-world-gadget.md. It'll allow the `--update-metadata` flag to work fine.

gadgets/trace_mount/program.bpf.c
```diff
@@ -0,0 +1,197 @@
...
+	__u8 data[DATA_LEN];
+	enum op op;
+};
+
+const volatile pid_t target_pid = 0;
+
+GADGET_PARAM(target_pid);
+
+// we need this to make sure the compiler doesn't remove our struct
+const struct event *unusedevent __attribute__((unused));
```
mauriciovasquezbernal, line 52: This is provided by the `GADGET_TRACER` macro.

gadgets/trace_mount/program.bpf.c
```diff
@@ -0,0 +1,197 @@
...
+	const char *data;
+	enum op op;
+};
+
+struct event {
+	__u64 delta;
+	__u64 flags;
+	__u32 pid;
+	__u32 tid;
+	__u64 mount_ns_id;
```
mauriciovasquezbernal, line 36: This has to use the `gadget_mntns_id` type, otherwise the container enrichment won't work.

mauriciovasquezbernal (commented):
Just one blocking comment. LGTM after it's fixed.

PR #2274, opened 2023-12-06
Add code to use eBPF ring buffer

eiffel-fl:
Hi.

This PR adds code to use eBPF ring buffer when they are available and defaulting to perf one otherwise.
It is inspired by https://github.com/iovisor/bcc/pull/4262

I tested it and it indeed creates ring buffer:

```bash
$ sudo -E ./ig run mounts                                     francis/ringbuf *% u+2-2
RUNTIME.CONTAINERNAME   PID          TID          RET          COMM         FS           SRC          DEST        DATA        OP         
                        225253       225253       -2           mount        ext3         /foo         /bar                    MOUNT
^Z
$ sudo bpftool map list                                         francis/ringbuf *% u+2-2
...
198: ringbuf  name events  flags 0x0
        key 0B  value 0B  max_entries 262144  memlock 0B
        pids ig(225145)
```

It only works with the `trace mount` gadget which was tweaked to use the new functions.
At least, it shows having an agnostic approach is possible and we may adapt it to fit more gadgets.

Best regards.

### DIFF
--- a/gadgets/trace_mount/program.bpf.c
+++ b/gadgets/trace_mount/program.bpf.c
@@ -3,6 +3,7 @@
 #include <vmlinux.h>
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_core_read.h>
+#include <gadget/buffer.h>
 #include <gadget/macros.h>
 #include <gadget/mntns_filter.h>
 #include <gadget/types.h>
@@ -55,20 +56,10 @@ struct {
 	__type(value, struct arg);
 } args SEC(".maps");
 
-struct {
-	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
-	__uint(max_entries, 1);
-	__type(key, int);
-	__type(value, struct event);
-} heap SEC(".maps");
+// Roughly 30 events
+GADGET_TRACER_MAP(events, 1024 * 256);
 
-struct {
-	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
-	__uint(key_size, sizeof(__u32));
-	__uint(value_size, sizeof(__u32));
-} events SEC(".maps");
-
-GADGET_TRACER(exec, events, event);
+GADGET_TRACER(mount, events, event);
 
 // TODO: have to use "inline" to avoid this error:
 // bpf/mountsnoop.bpf.c:41:12: error: defined with too many args
@@ -116,7 +107,7 @@ static int probe_exit(void *ctx, int ret)
 	if (!argp)
 		return 0;
 
-	eventp = bpf_map_lookup_elem(&heap, &zero);
+	eventp = gadget_reserve_buf(&events, sizeof(*eventp));
 	if (!eventp)
 		goto cleanup;
 
@@ -150,8 +141,7 @@ static int probe_exit(void *ctx, int ret)
 	else
 		eventp->data[0] = '\0';
 
-	bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, eventp,
-			      sizeof(*eventp));
+	gadget_submit_buf(ctx, &events, eventp, sizeof(*eventp));
 
 cleanup:
 	bpf_map_delete_elem(&args, &tid);
--- a/gadgets/trace_mount/gadget.yaml
+++ b/gadgets/trace_mount/gadget.yaml
@@ -1,7 +1,7 @@
 name: trace mount
 description: trace mount syscalls
 tracers:
-  exec:
+  mount:
     mapName: events
     structName: event
 structs:


pkg/gadgets/run/tracer/tracer.go
```diff
@@ -177,6 +177,51 @@ func (t *Tracer) Close() {
...
+				return fmt.Errorf("no buffer map named %s", bufName)
+			}
+
+			bufMap.Type = ebpf.PerfEventArray
+			bufMap.KeySize = 4
+			bufMap.ValueSize = 4
+
+			t.spec.Maps[bufName] = bufMap
+		}
+	}
```
mauriciovasquezbernal, line 205: Print debug messages indicating ring_buf is not available and that logic will fallback to perf event array.

gadgets/trace_mount/program.bpf.c
```diff
@@ -117,9 +105,9 @@ static int probe_exit(void *ctx, int ret)
 	if (!argp)
 		return 0;
 
-	eventp = bpf_map_lookup_elem(&heap, &zero);
+	eventp = reserve_buf(sizeof(*eventp));
 	if (!eventp)
-		return 0;
+		goto cleanup;
```
mauriciovasquezbernal, line 110: Is this fixing an error in the previous implementation? I can't see how it's related to this PR.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -177,6 +177,51 @@ func (t *Tracer) Close() {
...
+func ringbufAvailable() bool {
+	_, err := ebpf.NewMap(&ebpf.MapSpec{
+		Type:       ebpf.RingBuf,
+		MaxEntries: uint32(os.Getpagesize()),
+	})
+
+	return err == nil
+}
+
+func (t *Tracer) loadeBPFObjects(opts ebpf.CollectionOptions, bufName string) error {
```
mauriciovasquezbernal, line 189: Keep the param name more specific?

```suggestion
func (t *Tracer) loadeBPFObjects(opts ebpf.CollectionOptions, tracerMapName string) error {
```

mauriciovasquezbernal (commented):
It looks very promising!

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -202,8 +202,6 @@ func (t *Tracer) loadeBPFObjects(opts ebpf.CollectionOptions, bufName string) er
 
 			t.spec.Maps[bufName] = bufMap
 		}
-	} else {
-		delete(t.spec.Maps, "heap")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Nice idea!

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -177,6 +179,78 @@ func (t *Tracer) Close() {
...
+		heapMapName := "gadget_heap"
+		// If we delete the gadget_heap map, the verifier will not load the code
+		// because it cannot find it despite the code using it being dead one:
+		// ...: create BPF collection: program ig_mount_x: missing map gadget_heap
+		// Rather than deleting it, we just set its size to 0 and use a hash to
+		// avoid having one struct per CPU, i.e. reducing as much as possible the
+		// memory footprint it uses.
+		heapMap, ok := t.spec.Maps[heapMapName]
+		if !ok {
+			return fmt.Errorf("no heap map named %s", heapMapName)
```
mauriciovasquezbernal, line 227: I don't think it has to be an error, if the map doesn't exist, just ignore it and keep going.

include/gadget/compat.bpf.h
```diff
@@ -0,0 +1,45 @@
...
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, MAX_EVENT_SIZE);
+} heap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, RINGBUF_SIZE);
+} events SEC(".maps");
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I was thinking to have something like https://github.com/inspektor-gadget/inspektor-gadget/commit/4fc4a79bc58ec36ba27c10ea94617f16d73362d6. 

IMO we should only apply this logic to maps that are created by using the macro. If the user explicitly creates the map of a given type, we shouldn't try to change it.

gadgets/trace_mount/program.bpf.c
```diff
@@ -55,20 +56,9 @@ struct {
...
-	__type(key, int);
-	__type(value, struct event);
-} heap SEC(".maps");
-
-struct {
-	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
-	__uint(key_size, sizeof(__u32));
-	__uint(value_size, sizeof(__u32));
-} events SEC(".maps");
+GADGET_TRACER_MAP(events, 1024 * 256);
```
mauriciovasquezbernal, line 59: Perhaps add a comment indicating how many events `1024 * 256` corresponds to?

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -177,6 +179,136 @@ func (t *Tracer) Close() {
...
+
+		return err
+	}
+
+	t.ringbufReader, err = ringbuf.NewReader(t.collection.Maps[tracerMapName])
+
+	return err
+}
+
+func (t *Tracer) loadeBPFObjects(opts ebpf.CollectionOptions, tracerMapName string) error {
```
mauriciovasquezbernal, line 271: I like having this small functions to make the code more modular. I suspect in the future we'll be adding more parameters to it, so what about using a structure as parameters from the beginning?

mauriciovasquezbernal (commented):
I really like it, IMO it's almost ready to go. My only blocking comment is about removing `createTracerMap`. I'll test it later on.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -271,22 +271,38 @@ func (t *Tracer) createTracerMap(tracerMapName string) error {
...
+			m := t.collection.Maps[tracerMapName]
+			switch m.Type() {
+			case ebpf.RingBuf:
+				t.ringbufReader, err = ringbuf.NewReader(t.collection.Maps[tracerMapName])
+			case ebpf.PerfEventArray:
+				t.perfReader, err = perf.NewReader(t.collection.Maps[tracerMapName], gadgets.PerfBufferPages*os.Getpagesize())
+			}
+			if err != nil {
+				return fmt.Errorf("create BPF map reader: %w", err)
+			}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I meant that following will work the same with less code:

```diff
diff --git pkg/gadgets/run/tracer/tracer.go pkg/gadgets/run/tracer/tracer.go
index 077ebe2e..69fcadde 100644
--- pkg/gadgets/run/tracer/tracer.go
+++ pkg/gadgets/run/tracer/tracer.go
@@ -254,20 +254,6 @@ func (t *Tracer) handleTracerMapDefinition(tracerMapName string) error {
        return nil
 }
 
-func (t *Tracer) createTracerMap(tracerMapName string) error {
-       var err error
-
-       if !isRingbufAvailable() {
-               t.perfReader, err = perf.NewReader(t.collection.Maps[tracerMapName], gadgets.PerfBufferPages*os.Getpagesize())
-
-               return err
-       }
-
-       t.ringbufReader, err = ringbuf.NewReader(t.collection.Maps[tracerMapName])
-
-       return err
-}
-
 func (t *Tracer) loadeBPFObjects(opts ebpf.CollectionOptions, tracerMapName string) error {
        var err error
 
@@ -285,24 +271,17 @@ func (t *Tracer) loadeBPFObjects(opts ebpf.CollectionOptions, tracerMapName stri
                return fmt.Errorf("create BPF collection: %w", err)
        }
 
-       if isTracerMapMacro {
-               err = t.createTracerMap(tracerMapName)
-               if err != nil {
-                       return fmt.Errorf("creating tracer map through GADGET_TRACER_MAP: %w", err)
+       // Some logic before loading the programs
+       if tracerMapName != "" {
+               m := t.collection.Maps[tracerMapName]
+               switch m.Type() {
+               case ebpf.RingBuf:
+                       t.ringbufReader, err = ringbuf.NewReader(t.collection.Maps[tracerMapName])
+               case ebpf.PerfEventArray:
+                       t.perfReader, err = perf.NewReader(t.collection.Maps[tracerMapName], gadgets.PerfBufferPages*os.Getpagesize())
                }
-       } else {
-               // Some logic before loading the programs
-               if tracerMapName != "" {
-                       m := t.collection.Maps[tracerMapName]
-                       switch m.Type() {
-                       case ebpf.RingBuf:
-                               t.ringbufReader, err = ringbuf.NewReader(t.collection.Maps[tracerMapName])
-                       case ebpf.PerfEventArray:
-                               t.perfReader, err = perf.NewReader(t.collection.Maps[tracerMapName], gadgets.PerfBufferPages*os.Getpagesize())
-                       }
-                       if err != nil {
-                               return fmt.Errorf("create BPF map reader: %w", err)
-                       }
+               if err != nil {
+                       return fmt.Errorf("create BPF map reader: %w", err)
                }
        }
 
```

gadgets/trace_mount/program.bpf.c
mauriciovasquezbernal: What's the issue with `sizeof(struct event)`? Is it that the size has to be a power of 2? Or is there another issue?

mauriciovasquezbernal: Right, it's because the sizeof(struct event) * N is not a power of two. The following makes it work but wastes a lot of space:


```diff
diff --git gadgets/trace_mount/program.bpf.c gadgets/trace_mount/program.bpf.c
index 6be76879..48210e7e 100644
--- gadgets/trace_mount/program.bpf.c
+++ gadgets/trace_mount/program.bpf.c
@@ -43,6 +43,8 @@ struct event {
        __u8 dest[PATH_MAX];
        __u8 data[DATA_LEN];
        enum op op;
+
+       __u8 pad[7608];
 };
 
 const volatile pid_t target_pid = 0;
@@ -57,7 +59,7 @@ struct {
 } args SEC(".maps");
 
 // Roughly 30 events
-GADGET_TRACER_MAP(events, 1024 * 256);
+GADGET_TRACER_MAP(events, sizeof(struct event)*32);
 
 GADGET_TRACER(mount, events, event);
```


 it'd be nice to have a macro to round to the next power of 2, but I couldn't find it -> https://stackoverflow.com/questions/22925016/rounding-up-to-powers-of-2-with-preprocessor-constants.

mauriciovasquezbernal (approved):
LGTM. Tested on `5.4.0-99-generic` and I can see how it prints the `DEBU[0002] Ring buffers are not available, defaulting to perf ones` message and the gadget works fine.

Thanks for working on this and handling all the comments.

flyth:
This looks nice!

I just read through the differences of ringbuf/perfbuf - using the ring buffer, we'd have to figure out dropped events on our own (afais it isn't handled by the current version of the run gadget for perf buffer as well, though). Maybe we should provide some helpers for that as well in the helper header file - maybe in a later PR?
Also, maybe we should add a helper for `bpf_ringbuf_discard` as `gadget_discard_buf` to make it even more consistent?

Issue #1213, opened 2022-12-20
`top/ebpf` does not work on `minikube` with `kvm` driver


eiffel-fl:
## Description

Nothing is printed when run `top/ebpf` when `minikube` was started with `--driver=kvm`.

## Impact

It is not possible to monitor eBPF programs on this platform.

## Environment and steps to reproduce
There is indeed an error when creating the PID tracer as the string `bpf_prog_fops` does not appear inside `/proc/kallsyms`:

```bash
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ minikube start --driver=docker                                            francis/local-gadget-top *% u+22-14
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ minikube version                                                                    francis/local-gadget-top *% u+22-14
minikube version: v1.28.0
commit: 986b1ebd987211ed16f8cc10aed7d2c42fc8392f
francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget top ebpf -o json                                                   francis/local-gadget-top *% u+22-14
^C%                                                                                                                                                                         francis@pwmachine:~/Codes/kinvolk/inspektor-gadget$ minikube ssh                                                                        francis/local-gadget-top *% u+22-14
...
$ uname -a
Linux minikube 5.10.57 #1 SMP Fri Oct 28 21:02:11 UTC 2022 x86_64 GNU/Linux
$ grep bpf_prog_fops /proc/kallsyms 
$
```

## Expected behavior

The eBPF programs statistics should be displayed.

## Additional information

This problems occurs both on `main` and even since `v0.10.0`.

PR #2280, opened 2023-12-10
kallsyms: use ebpf alternative when kallsyms not available

alban:
# kallsyms: use ebpf alternative when kallsyms not available

On the minikube kernel 5.10.57, /proc/kallsyms does not provide the following symbols:
- bpf_prog_fops (used by the top-ebpf gadget)
- socket_file_ops (used by the socket enricher)

This patch provides an alternative to get the address of those symbols using ebpf.

Most of the required ebpf code was already implemented for the container-hook package. The file is moved to make it available in a generic package:
pkg/container-hook/bpf/privatedata.bpf.c -> pkg/kfilefields/bpf/filefields.bpf.c

## How to use

No changes

## Testing done

```
minikube start -p minikube-kvm2 --driver=kvm2
kubectl run -ti --rm --image=busybox --restart=Never shell01 -- sh
nc -l -p 8080 &

minikube cp ig minikube-kvm2:/usr/bin/ig
minikube ssh
sudo chmod +x /usr/bin/ig
sudo ig top ebpf
sudo ig trace dns
sudo ig trace network
```

Tested by adding a `fmt.Printf`:
```
$ sudo ig trace dns
...
symbol socket_file_ops found at ffffffff94ea0660 using ebpf
...
$ sudo ig top ebpf
...
symbol bpf_prog_fops found at ffffffff94d99ea0 using ebpf
...
```

Tested by reverting 681c3701c31e0f760a81d603f5b2249c29a5b2be (#2270). It works too.

### GKE

integration tests run on GKE are passing now since symbol are being resolved:

- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8094106145/job/22119633956
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8094106145/job/22119634529

## Links

- Previous PR implementing most of the ebpf code for this: #1789
- Issue with top-ebpf #1213
- Issue with the socket enricher in trace-dns #2268

## TODO

- [ ] More automatic tests? Which ones?
- [ ] Should #2270 be reverted in this PR?


pkg/kallsyms/kallsyms.go
```diff
@@ -179,6 +179,12 @@ func specUpdateAddresses(
 		}
 	}
 
+	// List of symbols that we're able to find without using /proc/kallsyms
```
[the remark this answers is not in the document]
alban: kallsyms is also used in:
- profile-cpu gadget (`kAllSyms.LookupByInstructionPointer(ip)`)
- top-block-io gadget (`__blk_account_io_start`, `blk_account_io_start`, `__blk_account_io_done`, `blk_account_io_done`

In those cases, the symbol is not used in any field of a `struct file *`, so the same trick to pass a file descriptor in a unix socket (`SCM_RIGHTS`) cannot be used.

alban:
I've refactorised the code to make it possible to add some unit tests. It is now possible to test that the two methods to get the addresses of `bpf_prog_fops` and `socket_file_ops` return the same values.

> > Should https://github.com/inspektor-gadget/inspektor-gadget/pull/2270 be reverted in this PR?
> 
> Why not, but if everything failed (reading /proc/kallsyms and using eBPF), we should still print a warning.

After more thinking, I think both could be done in a future PR.

pkg/kallsyms/kallsyms.go
```diff
@@ -149,12 +151,54 @@ var (
...
+		}
+	}
+}
+
+func symbolResolverFromEbpf() symbolResolver {
+	// List of symbols that we're able to find in 'struct file *' using eBPF
+	symbolsBypass := map[string]fdType{
+		"socket_file_ops": fdTypeSocket,
+		"bpf_prog_fops":   fdTypeEbpfProgram,
+	}
```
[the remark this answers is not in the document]
alban: I am not sure, this map is only used in `func symbolResolverFromEbpf()`.

alban:
> > I've refactorised the code to make it possible to add some unit tests. It is now possible to test that the two methods to get the addresses of `bpf_prog_fops` and `socket_file_ops` return the same values.
> > > > Should #2270 be reverted in this PR?
> > > 
> > > 
> > > Why not, but if everything failed (reading /proc/kallsyms and using eBPF), we should still print a warning.
> > 
> > 
> > After more thinking, I think both could be done in a future PR.
> 
> What do you mean by both?

Sorry I was confused. At the moment, we print a warning when both methods (kallsyms and eBPF) fail, so I think it is fine as it is.

pkg/kallsyms/bpf/privatedata.bpf.c
```diff
@@ -9,13 +9,21 @@ const volatile u64 socket_ino = 0;
...
 	__type(value, u64); // private_data
 } ig_private_data SEC(".maps");
 
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int); // index: zero
+	__type(value, u64); // f_op
+} ig_f_op SEC(".maps");
+
```
mauriciovasquezbernal, line 26: What about using the same map with a structure containing private data and f_op? That'd make the handling easier as there is a single map, also readStructFileFields could return this structure.

pkg/kallsyms/kallsyms.go
```diff
@@ -149,12 +151,53 @@ var (
...
+		[]symbolResolver{
+			symbolResolverFromKallsyms(NewKAllSyms),
+			symbolResolverFromEbpf(),
+		},
+		spec,
+		symbols,
+	)
+}
+
+func symbolResolverFromKallsyms(kAllSymsFactory func() (*KAllSyms, error)) symbolResolver {
```
mauriciovasquezbernal, line 164: Why does it need a factory, can it receive `KAllSyms` directly?

pkg/kallsyms/kallsyms.go
```diff
@@ -149,12 +151,53 @@ var (
 //
 //	kallsyms.SpecUpdateAddresses(spec, []string{"socket_file_ops"})
 func SpecUpdateAddresses(spec *ebpf.CollectionSpec, symbols []string) error {
```
mauriciovasquezbernal, line 153: Move the `len(symbols) == 0` check on L204 here to avoid allocating things if nothing has to be done?

pkg/kallsyms/kallsyms.go
```diff
@@ -149,12 +151,53 @@ var (
...
+		return addr, nil
+	}
+}
+
+func symbolResolverFromEbpf() symbolResolver {
+	// List of symbols that we're able to find in 'struct file *' using eBPF
+	symbolsBypass := map[string]fdType{
+		"socket_file_ops": fdTypeSocket,
+		"bpf_prog_fops":   fdTypeEbpfProgram,
+	}
```
mauriciovasquezbernal, line 187: general golang question: is it better to use a map than a switch case?

pkg/kallsyms/kallsyms.go
```diff
@@ -137,6 +137,8 @@ var (
 	triedGetAddr = map[string]error{}
 )
 
+type symbolResolver func(symbol string) (uint64, error)
```
mauriciovasquezbernal, line 140: I'm wondering if defining it as an interface and implementing it on a Kallsyms and Bpf resolvers will make the code cleaner.

pkg/kallsyms/kallsyms.go
mauriciovasquezbernal, line 1: Some comments about previous implementation:

- Would merging `symbolsMap` and `triedGetAddr` maps in a single one containing a `type symbolResolve struct { addr uint64; err error }` as value make the code better?
- Does SymbolExists have to acquire the lock? Why it doesn't trigger the lookup logic and only returns from cache?

pkg/kallsyms/kallsyms.go
alban: Because the unit test uses a different factory that does not requires root but provides its own alternative to `/proc/kallsyms`.

mauriciovasquezbernal: What about something like:

```diff
diff --git pkg/kallsyms/kallsyms.go pkg/kallsyms/kallsyms.go
index 46522163..eb1d71b3 100644
--- pkg/kallsyms/kallsyms.go
+++ pkg/kallsyms/kallsyms.go
@@ -151,9 +151,14 @@ type symbolResolver func(symbol string) (uint64, error)
 //
 //     kallsyms.SpecUpdateAddresses(spec, []string{"socket_file_ops"})
 func SpecUpdateAddresses(spec *ebpf.CollectionSpec, symbols []string) error {
+       kAllSyms, err := NewKAllSyms()
+       if err != nil {
+               return err
+       }
+
        return specUpdateAddresses(
                []symbolResolver{
-                       symbolResolverFromKallsyms(NewKAllSyms),
+                       symbolResolverFromKallsyms(kAllSyms),
                        symbolResolverFromEbpf(),
                },
                spec,
@@ -161,16 +166,8 @@ func SpecUpdateAddresses(spec *ebpf.CollectionSpec, symbols []string) error {
        )
 }
 
-func symbolResolverFromKallsyms(kAllSymsFactory func() (*KAllSyms, error)) symbolResolver {
-       var kAllSyms *KAllSyms
+func symbolResolverFromKallsyms(kAllSyms *KAllSyms) symbolResolver {
        return func(symbol string) (uint64, error) {
-               if kAllSyms == nil {
-                       var err error
-                       kAllSyms, err = kAllSymsFactory()
-                       if err != nil {
-                               return 0, err
-                       }
-               }
                addr, ok := kAllSyms.symbolsMap[symbol]
                if !ok {
                        return 0, fmt.Errorf("symbol %q was not found in kallsyms", symbol)
diff --git pkg/kallsyms/kallsyms_test.go pkg/kallsyms/kallsyms_test.go
index 5274f334..d48f90c5 100644
--- pkg/kallsyms/kallsyms_test.go
+++ pkg/kallsyms/kallsyms_test.go
@@ -122,27 +122,27 @@ var specTest = &ebpf.CollectionSpec{
 func TestSpecRewriteConstants(t *testing.T) {
        t.Cleanup(resetState)
 
-       kAllSymsFactory := func() (*KAllSyms, error) {
-               kAllSymsStr := strings.Join([]string{
-                       "0000000000000000 A fixed_percpu_data",
-                       "ffffffffb4231f40 D bpf_prog_fops",
-                       "ffffffffb43723e0 d socket_file_ops",
-               }, "\n")
-
-               kAllSymsReader := strings.NewReader(kAllSymsStr)
-               kAllSyms, err := NewKAllSymsFromReader(kAllSymsReader)
-               require.Nil(t, err, "NewKAllSymsFromReader failed: %v", err)
-               return kAllSyms, nil
-       }
+       var err error
+
+       kAllSymsStr := strings.Join([]string{
+               "0000000000000000 A fixed_percpu_data",
+               "ffffffffb4231f40 D bpf_prog_fops",
+               "ffffffffb43723e0 d socket_file_ops",
+       }, "\n")
+
+       kAllSymsReader := strings.NewReader(kAllSymsStr)
+       kAllSyms, err := NewKAllSymsFromReader(kAllSymsReader)
+       require.NoError(t, err, "NewKAllSymsFromReader failed: %v", err)
+
        // Little endian representation of the addresses above:
        bpfProgFopsAddr := []byte{0x40, 0x1f, 0x23, 0xb4, 0xff, 0xff, 0xff, 0xff}
        socketFileOpsAddr := []byte{0xe0, 0x23, 0x37, 0xb4, 0xff, 0xff, 0xff, 0xff}
 
        spec := specTest.Copy()
 
-       err := specUpdateAddresses(
+       err = specUpdateAddresses(
                []symbolResolver{
-                       symbolResolverFromKallsyms(kAllSymsFactory),
+                       symbolResolverFromKallsyms(kAllSyms),
                },
                spec,
                []string{"abcde_bad_name"},
@@ -151,7 +151,7 @@ func TestSpecRewriteConstants(t *testing.T) {
 
        err = specUpdateAddresses(
                []symbolResolver{
-                       symbolResolverFromKallsyms(kAllSymsFactory),
+                       symbolResolverFromKallsyms(kAllSyms),
                },
                spec,
                []string{"bpf_prog_fops", "socket_file_ops"},
@@ -169,12 +169,15 @@ func TestSpecRewriteConstantsRoot(t *testing.T) {
        utilstest.RequireRoot(t)
        t.Cleanup(resetState)
 
+       kAllSyms, err := NewKAllSyms()
+       require.NoError(t, err, "NewKAllSyms failed: %v", err)
+
        resetState()
 
        spec1 := specTest.Copy()
-       err := specUpdateAddresses(
+       err = specUpdateAddresses(
                []symbolResolver{
-                       symbolResolverFromKallsyms(NewKAllSyms),
+                       symbolResolverFromKallsyms(kAllSyms),
                },
                spec1,
                []string{"bpf_prog_fops", "socket_file_ops"},
@@ -198,7 +201,7 @@ func TestSpecRewriteConstantsRoot(t *testing.T) {
        spec3 := specTest.Copy()
        err = specUpdateAddresses(
                []symbolResolver{
-                       symbolResolverFromKallsyms(NewKAllSyms),
+                       symbolResolverFromKallsyms(kAllSyms),
                        symbolResolverFromEbpf(),
                },
                spec3,

```

mqasimsarfraz:
It seems this is [affecting integrations runs](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7844617994/attempts/1#summary-21407476104) on GKE (OS: `Container-Optimized OS from Google`, Kernel: `5.15.133+`) since I see following warnings in the log:

```
time="2024-02-09T13:38:05Z" level=warning msg="updating socket_file_ops address with ksyms: file does not exist\nEither you cannot access /proc/kallsyms or this file does not contain socket_file_ops"
time="2024-02-09T13:38:17Z" level=warning msg="updating socket_file_ops address with ksyms: file does not exist\nEither you cannot access /proc/kallsyms or this file does not contain socket_file_ops"
```

I tested it with the [image from this PR](https://github.com/mqasimsarfraz/inspektor-gadget) and results are looking better.

pkg/kallsyms/kallsyms.go
```diff
@@ -149,20 +151,63 @@ var (
...
+func symbolResolverFromKallsyms(kAllSymsFactory func() (*KAllSyms, error)) symbolResolver {
+	var kAllSyms *KAllSyms
+	return func(symbol string) (uint64, error) {
+		if kAllSyms == nil {
+			var err error
+			kAllSyms, err = kAllSymsFactory()
+			if err != nil {
+				return 0, err
+			}
+		}
```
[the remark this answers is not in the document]
alban: Maybe `sync.Once()` but not `init()` since we want to avoid parsing kallsyms if the gadget does not use it.

pkg/kallsyms/kallsyms.go
```diff
@@ -149,20 +149,31 @@ var (
...
-	return specUpdateAddresses(kAllSymsFactory, spec, symbols)
+	if len(symbols) == 0 {
+		// Nothing to do
+		return nil
+	}
+
+	kAllSyms, err := NewKAllSyms()
+	if err != nil {
+		return err
+	}
```
alban, line 160: Before this change, `NewKAllSyms()` was only called if we have missing entries in the cache. With this change, we lose performance by reading and parsing /proc/kallsyms every time `SpecUpdateAddresses()` is called.

I suspect this causes the following performance issue in trace-dns and top-ebpf:
https://github.com/inspektor-gadget/inspektor-gadget/commit/661af6b62c392162aa18f5740645b784e8baa206#commitcomment-139171154

mqasimsarfraz: Great catch, you are right it seem to be related:

```
→ go test -exec sudo  -bench='BenchmarkAllGadgetsWithContainers/container0/top-ebpf'  -run=Benchmark ./internal/benchmarks/...
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
BenchmarkAllGadgetsWithContainers/container0/top-ebpf-8                        2         515444363 ns/op
PASS
ok      github.com/inspektor-gadget/inspektor-gadget/internal/benchmarks        1.716s

```

compared to using `sync.Once` as recommend [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2280#discussion_r1502323160):

```
→ go test -exec sudo  -bench='BenchmarkAllGadgetsWithContainers/container0/top-ebpf'  -run=Benchmark ./internal/benchmarks/...
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
BenchmarkAllGadgetsWithContainers/container0/top-ebpf-8                      100          11858637 ns/op
PASS
ok      github.com/inspektor-gadget/inspektor-gadget/internal/benchmarks        1.905s

```

pkg/kallsyms/kallsyms.go
mqasimsarfraz: I didn't touch these to avoid changing to much but happy to revisit if you think they should be part of this PR.

pkg/kfilefields/tracer.go
```diff
@@ -92,38 +122,46 @@ func readPrivateDataFromFd(fd int) (uint64, error) {
...
+	switch kind {
+	case FdTypeCustom:
+		// nothing to do
+	case FdTypeSocket:
+		fd = sock[0]
+	case FdTypeEbpfProgram:
+		fd = objs.IgFgetX.FD()
+	default:
+		return nil, fmt.Errorf("unknown fd type %d", kind)
+	}
```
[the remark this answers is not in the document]
mqasimsarfraz: I tried to make it more readable, let me know what do you think now?

pkg/kallsyms/kallsyms_test.go
```diff
@@ -77,78 +82,153 @@ func TestRealKAllSyms(t *testing.T) {
...
+		"ffffffffb43723e0 d socket_file_ops",
+	},
+		"\n")
+	kAllSymsReader := strings.NewReader(kAllSymsStr)
+	kAllSyms, err := NewKAllSymsFromReader(kAllSymsReader)
+	require.NoError(t, err, "NewKAllSymsFromReader failed: %v", err)
+
+	// Little endian representation of the addresses above:
+	bpfProgFopsAddr := []byte{0x40, 0x1f, 0x23, 0xb4, 0xff, 0xff, 0xff, 0xff}
+	socketFileOpsAddr := []byte{0xe0, 0x23, 0x37, 0xb4, 0xff, 0xff, 0xff, 0xff}
```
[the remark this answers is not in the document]
mqasimsarfraz: Good idea. Do you suggest to handle it in the same PR or in a separate PR? Also, regarding implementation I was wondering if we should have a separate package for [`NativeEndian`](https://github.com/inspektor-gadget/inspektor-gadget/blob/b2479d43f4de4e65ce16a17b76ee41f3cd046c70/pkg/datasource/endian_be.go) ?

pkg/kfilefields/kfilefields.go
```diff
@@ -0,0 +1,60 @@
...
+
+import "fmt"
+
+// ReadPrivateDataFromFd uses ebpf to read the private_data pointer from the
+// kernel "struct file" associated with the given fd.
+func ReadPrivateDataFromFd(fd int) (uint64, error) {
+	t, err := creatAndInstallTracer()
+	if err != nil {
+		return 0, fmt.Errorf("creating and installing tracer: %w", err)
+	}
```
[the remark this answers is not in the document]
mqasimsarfraz: You mean to avoid an extra call [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/f6569d97981748e999fb476da79937f72613df98/pkg/kfilefields/kfilefields.go#L46) ? Currently it isn't used often so should be fine as it is or am I missing something?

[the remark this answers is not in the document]
alban: We call `ReadPrivateDataFromFd` only once when the `container-hook` package is initialized, so I don't think it makes sense to optimize.

And about `ReadFOpForFdType`: it is used in `specUpdateAddresses()` but there is already a cache, so we won't install the tracer several times.

pkg/kallsyms/kallsyms.go
```diff
@@ -149,20 +154,34 @@ var (
...
+		// Nothing to do
+		return nil
+	}
+
+	var err error
+	once.Do(func() {
+		kallsyms, err = NewKAllSyms()
+	})
+	if err != nil {
+		return err
```
alban, line 167: I don't think the `err != nil` test would capture errors correctly for subsequent calls to SpecUpdateAddresses.
See https://go.dev/play/p/2RRA9mrM8Fq

The `err` variable needs to be a global variable next to once line 31.

pkg/kallsyms/kallsyms.go
```diff
@@ -149,20 +154,34 @@ var (
...
-	kAllSymsFactory := NewKAllSyms
-	return specUpdateAddresses(kAllSymsFactory, spec, symbols)
+	if len(symbols) == 0 {
+		// Nothing to do
+		return nil
+	}
+
+	var err error
+	once.Do(func() {
+		kallsyms, err = NewKAllSyms()
```
alban, line 164:
```
$ cat /proc/kallsyms |wc -c
15861623
```

This file is 16 MiB. In the previous implementation, the kallsyms was released from memory once the gadget using it (profile cpu or top block-io) is no longer using it. In this new implementation, it stays in memory permanently, even when the user stops a gadget such as profile-cpu.

Do you think it is the correct optimization between memory consumption and cpu consumption?

pkg/kfilefields/tracer.go
alban, line 1: Could you update CODEOWNERS about `pkg/kfilefields`?

mqasimsarfraz: Done.

pkg/kallsyms/kallsyms.go
mqasimsarfraz: Good catch! I fixed it.

pkg/kallsyms/kallsyms.go
mqasimsarfraz: I believe the caching logic we have in `specUpdateAddresses` means we can avoid having to keep the content of this file in memory. I wondering if it is better to delay reading this file till we try to resolve the symbols like before, something like this:

```
→ git diff
diff --git a/pkg/kallsyms/kallsyms.go b/pkg/kallsyms/kallsyms.go
index 0598f3a2..0381ff45 100644
--- a/pkg/kallsyms/kallsyms.go
+++ b/pkg/kallsyms/kallsyms.go
@@ -160,16 +160,9 @@ func SpecUpdateAddresses(spec *ebpf.CollectionSpec, symbols []string) error {
                return nil
        }
 
-       once.Do(func() {
-               kallsyms, initErr = NewKAllSyms()
-       })
-       if initErr != nil {
-               return fmt.Errorf("initializing kallsyms: %w", initErr)
-       }
-
        return specUpdateAddresses(
                []symbolResolver{
-                       newKAllSymsResolver(kallsyms),
+                       newKAllSymsResolver(),
                        newEbpfResolver(),
                },
                spec,
diff --git a/pkg/kallsyms/symbol-resolver.go b/pkg/kallsyms/symbol-resolver.go
index e2f74872..48c5dea8 100644
--- a/pkg/kallsyms/symbol-resolver.go
+++ b/pkg/kallsyms/symbol-resolver.go
@@ -32,15 +32,19 @@ type kAllSymsResolver struct {
 }
 
 // newKAllSymsResolver returns a new kAllSymsResolver.
-func newKAllSymsResolver(kAllSyms *KAllSyms) kAllSymsResolver {
-       if kAllSyms == nil {
-               panic("kAllSyms is nil")
-       }
-       return kAllSymsResolver{kAllSyms: kAllSyms}
+func newKAllSymsResolver() kAllSymsResolver {
+       return kAllSymsResolver{}
 }
 
 // Resolve resolves a kernel symbol using the KAllSyms.
 func (r kAllSymsResolver) Resolve(symbol string) (uint64, error) {
+       var err error
+       if r.kAllSyms == nil {
+               r.kAllSyms, err = NewKAllSyms()
+       }
+       if err != nil {
+               return 0, fmt.Errorf("failed to read kallsyms: %w", err)
+       }
        addr, ok := r.kAllSyms.symbolsMap[symbol]
        if !ok {
                return 0, fmt.Errorf("symbol %q was not found in kallsyms", symbol)

```

What do you say?

alban: The general idea looks good to me.

One small documentation issue:
```
// Resolve resolves a kernel symbol using the KAllSyms.
func (r kAllSymsResolver) Resolve(symbol string) (uint64, error) {
```

With the test `r.kAllSyms == nil`, this can not be called in parallel from multiple go routine. This is fine because the only user of this API is `specUpdateAddresses()` and it does not do parallel calls. But you can document it.

Btw, can the function `Resolve()` be unexported (renamed to `resolve()`)? Or does Golang impose methods from interfaces to be exported?

alban:
@mqasimsarfraz I see the GitHub Action ran for both `pull_request` and `push` even though the branch is not prefixed with `citest/`. How come? Did you start the tests manually? In any cases, it's a good idea to test this PR on various kernels since it is meant to add support for more kernels.

mqasimsarfraz:
> How come? Did you start the tests manually? 

The `push` run is also [done in the CI](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8391532394). I pushed the changes to [citest/alban_privatedata](https://github.com/inspektor-gadget/inspektor-gadget/tree/citest/alban_privatedata) as well and it seems GH automatically registers the checks for push here. 

I didn't want to open a new PR and lose review, so I pushed it to seperate branch!

pkg/kallsyms/kallsyms.go
mqasimsarfraz: I updated the code to ensure do don't keep the content of `/proc/kallsyms` in memory

pkg/kallsyms/kallsyms.go
```diff
@@ -181,17 +187,25 @@ func specUpdateAddresses(
...
 	// Add the symbols that are not in the cache to the cache
 	if !allFoundInCache {
-		k, err := kAllSymsFactory()
-		if err != nil {
-			triedGetAddr[symbols[0]] = err
-			return err
+		// initialize the resolvers
+		for _, resolver := range symbolResolvers {
+			if err := resolver.init(); err != nil {
+				return fmt.Errorf("initializing resolver: %w", err)
```
alban, line 193: This will initialize both the kallsyms resolver and the ebpf resolver. But it might be that the first one works fine so there is no need to initialize the second one.
In practice, the ebpf resolver initialization is a no-op and it is at the second place, so it does not impact performance, but it makes the code looks weird.

Also, if the kernel has `CONFIG_KALLSYMS=n`, `/proc/kallsyms` will not exist and kallsyms resolver's init() will return an error and it will not fallback to the ebpf resolver.

So maybe init() should not return an error:
```diff
 type symbolResolver interface {
-	init() error
+	init()
```
The error could be saved in the `kAllSymsStrResolver struct` and returned later at `resolve()`-time:
```diff
 type kAllSymsStrResolver struct {
 	kAllSymsResolver
 	KAllSymsStr string
+	initErr error
 }
```
What do you think?

pkg/kallsyms/kallsyms_test.go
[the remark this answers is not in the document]
alban: Since we use go 1.21 (see go.mod), we don't need to define NativeEndian ourselves anymore. We can use binary.NativeEndian:
https://pkg.go.dev/encoding/binary#pkg-variables

See [go 1.21 release notes](https://tip.golang.org/doc/go1.21):

> [encoding/binary](https://tip.golang.org/pkg/encoding/binary/)
> The new [NativeEndian](https://tip.golang.org/pkg/encoding/binary/#NativeEndian) variable may be used to convert between byte slices and integers using the current machine’s native endianness.

---

But since this PR didn't introduce the problem, please switch to NativeEndian in another PR once this one is merged.

pkg/kallsyms/kallsyms_test.go
```diff
@@ -77,78 +82,151 @@ func TestRealKAllSyms(t *testing.T) {
...
+	utilstest.RequireRoot(t)
+	t.Cleanup(resetState)
+
+	resetState()
+
+	spec1 := specTest.Copy()
+	err := specUpdateAddresses(
+		[]symbolResolver{
+			newKAllSymsResolver(),
+		},
```
alban, line 179: Won't this test will only works on some kernels?

pkg/kallsyms/kallsyms_test.go
```diff
@@ -77,78 +82,151 @@ func TestRealKAllSyms(t *testing.T) {
...
+	err = specUpdateAddresses(
+		[]symbolResolver{},
+		spec4,
+		[]string{"bpf_prog_fops", "socket_file_ops"},
+	)
+	require.ErrorIs(t, err, os.ErrNotExist, "specUpdateAddresses should have failed")
+
+	t.Logf("spec1: %v", spec1.Maps[".rodata"].Contents[0].Value)
+	t.Logf("spec2: %v", spec2.Maps[".rodata"].Contents[0].Value)
+	t.Logf("spec3: %v", spec3.Maps[".rodata"].Contents[0].Value)
```
alban, line 222: What's the idea behind your tests with spec1 and spec2? It seems to be the same test, and the cache is reset in between with `resetState()`.

pkg/kallsyms/kallsyms_test.go
```diff
@@ -77,78 +82,151 @@ func TestRealKAllSyms(t *testing.T) {
...
+	)
+	require.Nil(t, err, "specUpdateAddresses failed: %v", err)
+
+	resetState()
+
+	spec3 := specTest.Copy()
+	err = specUpdateAddresses(
+		[]symbolResolver{
+			newKAllSymsResolver(),
+			newEbpfResolver(),
```
alban, line 203: newEbpfResolver() might not be tested since it is in second position, and there is no tests where it is in the first position.

alban (commented):
I have some questions on the tests.

But otherwise, LGTM if the CI passes on GKE and other platforms!

I cannot officially mark the PR as approved since I started it.

pkg/kallsyms/kallsyms_test.go
mqasimsarfraz: Yes, not sure what can be done about a check to skip test?

mqasimsarfraz:
> I have some questions on the tests.

I fixed the test

> But otherwise, LGTM if the CI passes on GKE and other platforms!

Great, I will keep it open for a bit so @eiffel-fl or someone else can have another pass!

alban:
Are the current CI failure on AKS related to this PR?

pkg/kallsyms/kallsyms_test.go
alban: Yes. There are already helper functions RequireRoot, RequireKernelVersion. What about a new one:
```
utilstest.RequireFileContains(t, "/proc/kallsyms", "bpf_prog_fops")
utilstest.RequireFileContains(t, "/proc/kallsyms", "socket_file_ops")
```

mqasimsarfraz:
> Are the current CI failure on AKS related to this PR?

I am not sure, it seems CI is failing on [`main` as well.](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8448407680). I will try to run failing test manually against an AKS to confirm that!

pkg/kallsyms/kallsyms_test.go
```diff
@@ -77,78 +84,153 @@ func TestRealKAllSyms(t *testing.T) {
...
+
+	spec2 := specTest.Copy()
+	err = specUpdateAddresses(
+		[]symbolResolver{
+			newEbpfResolver(),
+		},
+		spec2,
+		[]string{"bpf_prog_fops", "socket_file_ops"},
+	)
+	require.Nil(t, err, "specUpdateAddresses failed: %v", err)
```
alban, line 199: This section of the test can work on kernels without `/proc/kallsyms`. Could you duplicate it in a new test `TestSpecRewriteConstantsWithEbpf` without the `utilstest.RequireFileContains`?
Of course, it won't check that the addresses have the correct value but it could at least check that they are non-zero and that `specUpdateAddresses()` does not return an error.

mqasimsarfraz: Done, I see the test are still failing on the PR hopefully https://github.com/inspektor-gadget/inspektor-gadget/pull/2649 will help with it.

pkg/kallsyms/kallsyms_test.go
```diff
@@ -77,78 +84,169 @@ func TestRealKAllSyms(t *testing.T) {
...
+	spec := specTest.Copy()
+	err := specUpdateAddresses(
+		[]symbolResolver{
+			newEbpfResolver(),
+		},
+		spec,
+		[]string{"bpf_prog_fops", "socket_file_ops"},
+	)
+	require.Nil(t, err, "SpecUpdateAddresses failed: %v", err)
+	require.NotZero(t, spec.Maps[".rodata"].Contents[0].Value, "Contents should not be zero")
```
alban, line 251: `spec.Maps[".rodata"].Contents[0]..Value` is a `[]byte` and its length should be 16 (2 pointers) with two non-zero values. I think `require.NotZero` does not test that.

https://github.com/stretchr/testify/blob/edd74b24a1751990eeb831ebc70902c36600c3af/assert/assertions.go#L1682

I think `reflect.Zero()` for a `[]byte` would be an empty slice, so the test `require.NotZero` would always pass, even if it is a `[]byte` of size 16 with only zeros.

mqasimsarfraz: How about this?

```
func TestSpecRewriteConstantsWithEbpf(t *testing.T) {
	utilstest.RequireRoot(t)
	t.Cleanup(resetState)

	spec := specTest.Copy()
	err := specUpdateAddresses(
		[]symbolResolver{
			newEbpfResolver(),
		},
		spec,
		[]string{"bpf_prog_fops", "socket_file_ops"},
	)
	require.Nil(t, err, "SpecUpdateAddresses failed: %v", err)
	require.Len(t, spec.Maps[".rodata"].Contents[0].Value, 16, "Contents should have 16 bytes")
	values, ok := spec.Maps[".rodata"].Contents[0].Value.([]byte)
	require.True(t, ok, "Contents should be a byte slice")
	emptyBytes := []byte{0, 0, 0, 0, 0, 0, 0, 0}
	require.NotEqual(t, emptyBytes, values[:8], "First byte should not be zero")
	require.NotEqual(t, emptyBytes, values[8:], "Last 8 bytes should be zero")
}
```

mqasimsarfraz:
Thanks for the reviews!

Issue #450, opened 2022-01-03
[RFE] Mount host paths inside /host prefix in pod


invidian:
## Current situation

To avoid possible interference with host OS like in #433 and as discussed in https://github.com/kinvolk/inspektor-gadget/pull/434#issuecomment-998840919, I think it's a good idea to mount all host paths under `/host` in the container, e.g. to avoid service account token being mounted in host mount namespace, like it's done right now.

## Impact

Better isolation when accessing host mountpoints.

## Ideal future situation

All hostPath mounts are prefixed with `/host`.

## Implementation options

Right now, container runtime paths are hardcoded https://github.com/kinvolk/inspektor-gadget/blob/b568952edf4d8f19fe104192f031fde63c330a69/pkg/gadgettracermanager/containerutils/docker/docker.go#L28 and that would have to be configurable to resolve this issue.

PR #348, opened 2021-10-18
Remove privileged for `inspektor-gadget` container.

eiffel-fl:
Hi.

This PR addresses #128.
Basically, we need to add several capabilities (in addition to the [default ones](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/container_security_guide/linux_capabilities_and_seccomp)):
1. `CAP_SYS_RESOURCE`: `gadgetracermanager` calls `setrlimit` with an infinite value.
2. `CAP_SYS_ADMIN`: To interact with BPF.
3. `CAP_SYSLOG`: To get kernel symbols addresses in `/proc/kallsyms`.
4. `CAP_NET_ADMIN`: To be able to create BPF link with `BPF_PROG_TYPE_CGROUP_SKB` program type.

I tested all the python gadgets (expect `bindsnoop` which was CORE tested).
I used three kernels: 5.10.70 (with `kind`), 5.10.37 (with `kvm2`) and 5.4.40 (with `kvm2`).

<details>
  <summary>Here are the output for all of them:</summary>

* `bindsnoop`:
```bash
francis@machine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget bindsnoop                                     francis/remove-privileged *% u=
NODE             NAMESPACE        POD              CONTAINER        PID     COMM             RET PROTO OPTS  IF   PORT  ADDR
# Jump to an other terminal.
francis@machine:~/Codes/kinvolk/inspektor-gadget$ kubectl run --restart=Never --image=debian mypod -- sleep inf  francis/remove-privileged *% u=
pod/mypod created
francis@machine:~/Codes/kinvolk/inspektor-gadget$ kubectl exec -ti mypod -- bash                                 francis/remove-privileged *% u=
root@mypod:/# apt update && apt install netcat
...
root@mypod:/# netcat -l 80
# Go back to first terminal
Tracing binds ... Hit Ctrl-C to end
NODE             NAMESPACE        POD              CONTAINER             PID COMM         PROT ADDR            PORT   OPTS IF
minikube         default          mypod            mypod               31811 netcat       UNKN 0.0.0.0            80 ...Rr  0
```
* `biotop`:
```bash
francis@machine:~/Codes/kinvolk/inspektor-gadget$ ./kubectl-gadget biotop --all-namespaces --node ip-10-0-30-247 francis/remove-privileged *% u=

FATA[0000] biotop only works with --all-namespaces       args="[]" command="kubectl-gadget biotop"
[...]

PR #2285, opened 2023-12-11
charts: Reduce /host coverage and remove /lib/modules.

eiffel-fl:
The second is no more needed as we do not need any modules since we drop BCC support.

The first reduce the /host coverage by mounting only what we really need. Note that, majority of host volumes are mounted read only. The few which are mounted as read/write are related to NRI and OCI hooks.

### DIFF
--- a/charts/gadget/templates/daemonset.yaml
+++ b/charts/gadget/templates/daemonset.yaml
@@ -162,12 +162,37 @@ spec:
               {{- toYaml .Values.capabilities | nindent 14 }}
               {{- end }}
           volumeMounts:
-            - mountPath: /host
-              name: host
+            - mountPath: /host/bin
+              name: bin
+              readOnly: true
+            # We need to have read/write as we write NRI and OCI config files
+            # here.
+            - mountPath: /host/etc
+              name: etc
+              readOnly: false
+            # We need to have read/write as we write NRI and OCI binaries here.
+            - mountPath: /host/opt
+              name: opt
+              readOnly: false
+            - mountPath: /host/usr
+              name: usr
+              readOnly: true
+            - mountPath: /host/run
+              name: run
+              readOnly: true
+            # WARNING Despite mounting host proc as readonly, it is possible to
+            # write host file system using symlinks under /host/proc. The
+            # following command, ran from gadget pod, will result in writing to
+            # the host filesystem:
+            # touch /host/proc/1/root/foobar
+            # This limitation comes from Inspektor Gadget needing to be run as
+            # unconfined with regard to AppArmor and having the SYS_PTRACE
+            # capability.
+            - mountPath: /host/proc
+              name: proc
+              readOnly: true
             - mountPath: /run
               name: run
-            - mountPath: /lib/modules
-              name: modules
             - mountPath: /sys/kernel/debug
               name: debugfs
             - mountPath: /sys/fs/cgroup
@@ -192,18 +217,36 @@ spec:
           {{- toYaml .Values.tolerations | nindent 8 }}
         {{ end }}
       volumes:
-        - name: host
+        # /bin is needed to find runc.
+        - name: bin
           hostPath:
-            path: /
+            path: /bin
+        # /etc is needed for several reasons:
+        # 1. entrypoint needs /etc/os-release to print information.
+        # 2. entrypoint needs /etc/nri to handle NRI hooks
+        # 3. entrypoint needs /etc/containers/oci to handle OCI hooks.
+        - name: etc
+          hostPath:
+            path: /etc
+        # /opt is needed for several reasons:
+        # 1. entrypoint needs /opt/nri to handle NRI hooks.
+        # 2. entrypoint needs /opt/hooks/oci to handle OCI hooks.
+        - name: opt
+          hostPath:
+            path: /opt
+        # /usr is needed to find runc.
+        - name: usr
+          hostPath:
+            path: /usr
+        - name: proc
+          hostPath:
+            path: /proc
         - name: run
           hostPath:
             path: /run
         - name: cgroup
           hostPath:
             path: /sys/fs/cgroup
-        - name: modules
-          hostPath:
-            path: /lib/modules
         - name: bpffs
           hostPath:
             path: /sys/fs/bpf
--- a/pkg/resources/manifests/deploy.yaml
+++ b/pkg/resources/manifests/deploy.yaml
@@ -239,12 +239,37 @@ spec:
                 # https://github.com/inspektor-gadget/inspektor-gadget/blob/3c51ff5e9f5b/gadgets/trace_dns/program.bpf.c#L365-L366
                 - NET_RAW
           volumeMounts:
-            - mountPath: /host
-              name: host
+            - mountPath: /host/bin
+              name: bin
+              readOnly: true
+            # We need to have read/write as we write NRI and OCI config files
+            # here.
+            - mountPath: /host/etc
+              name: etc
+              readOnly: false
+            # We need to have read/write as we write NRI and OCI binaries here.
+            - mountPath: /host/opt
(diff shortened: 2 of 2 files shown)


charts/gadget/templates/daemonset.yaml
```diff
@@ -160,12 +160,38 @@ spec:
...
+            - mountPath: /host/opt/nri
+              name: opt-nri
+            # We need to have read/write as we write OCI binaries here.
+            - mountPath: /host/opt/hooks/oci
+              name: opt-hooks-oci
+            - mountPath: /host/usr
+              name: usr
+              readOnly: true
+            # We need to have read/write as we write OCI config files here.
+            - mountPath: /host//usr/share/containers/oci/
```
alban, line 185:
```suggestion
            # We need to have read/write as we write OCI config files here (podman/cri-o).
            - mountPath: /host/usr/share/containers/oci/
```

charts/gadget/templates/daemonset.yaml
```diff
@@ -160,12 +160,38 @@ spec:
...
+              name: usr-share-containers-oci
+            - mountPath: /host/run
+              name: run
+              readOnly: true
```
alban, line 189: The `run` volume is mounted twice:
- /host/run (RO)
- /run (RW)

Is it on purpose? (It is fine, just asking.)

The mount in /run needs to be mounted in RW mode somewhere to be able to connect to sockets such as:
- /run/docker.sock
- /run/containerd/containerd.sock
- /run/systemd/private

charts/gadget/templates/daemonset.yaml
```diff
@@ -184,18 +210,50 @@ spec:
...
+            path: /bin
+        # /etc/os-release is needed by entrypoint to print information.
+        - name: etc-os-release
+          hostPath:
+            path: /etc/os-release
+        # /etc/nri is needed by entrypoint and cleanup to handle NRI hooks.
+        - name: etc-nri
           hostPath:
-            path: /
+            path: /etc/nri
```
alban, line 224: The gadget pod would fail to start if `/etc/nri` does not exist on the host. I think we unfortunately have to mount the full /etc to avoid this issue.

Ditto for some other paths:
- `/etc/containers/oci/`
-  `/opt/hooks/oci`
- `/usr/share/containers/oci/`
- etc.

alban (commented):
Thanks for working on this.

Making some of the mounts read-only is good.

charts/gadget/templates/daemonset.yaml
mauriciovasquezbernal: I think this discussion is somehow related to https://github.com/inspektor-gadget/inspektor-gadget/issues/450.

pkg/resources/manifests/deploy.yaml
```diff
@@ -239,12 +239,27 @@ spec:
...
+              name: opt
+            - mountPath: /host/usr
+              name: usr
+              readOnly: true
+            - mountPath: /host/run
+              name: run
+              readOnly: true
+            - mountPath: /host/proc
+              name: proc
+              readOnly: true
```
mauriciovasquezbernal, line 260: Does it actually force `/host/proc` to be readonly? I can still modify the **entire** host filesystem by using `/host/proc/1/cwd/`, like `rm /host/proc/1/cwd/bin/ping`. Can we prevent it on any way?

mauriciovasquezbernal (approved):
It looks good. Thanks!

I only have a small question, however this PR is a big improvement, so I'm approving it.

alban (approved):
LGTM

The CI failure does not seem related to this PR, so it should not block merging.

charts/gadget/templates/daemonset.yaml
```diff
@@ -162,12 +162,27 @@ spec:
...
-            - mountPath: /host
-              name: host
+            - mountPath: /host/bin
+              name: bin
+              readOnly: true
+            # We need to have read/write as we write NRI and OCI config files
+            # here.
+            - mountPath: /host/etc
+              name: etc
+            # We need to have read/write as we write NRI and OCI binaries here.
```
rata, line 172: nit: I'd add an explicit `readOnly: false` when you want to be rw

pkg/resources/manifests/deploy.yaml
rata: @mauriciovasquezbernal I think the proc mount is read only, but that path contains a symlink and the destination is not read-only.

E.g:
```
$ ls -l /proc/self/cwd
lrwxrwxrwx 1 rodrigo rodrigo 0 Jan 30 15:11 /proc/self/cwd -> /home/rodrigo
```

My guess is that cwd (or root, also in that same dir) point to a directory that is not readOnly, and that is all.

Is it possible @eiffel-fl ? Can you print the output of `ls -l /proc/1/` and `mount` from that pod?

Be aware that `/bin/` might be a symlink too (like in a "merged" /usr, it can be a symlink to /usr/bin), so try to look all the paths for symlinks. My guess is that what it ends up pointing to is a a path mounted rw, and that is why you can write via /proc too.

rata: @eiffel-fl interesting. So you see this foobar file on the host too, if you just ssh to the host, right? Can you verify that, just in case? I would be surprised if you don't, but let's make sure when you claim you write to the host fs, you actually do it.

IMHO, if you write to the host fs, the most likely scenario is that there is a bizarre reason for the kubelet to do this, as weird mount namespace whatever, magic links in /proc being really magic and blah. But there is also the chance that this can be improved in upstream k8s.

The most likely scenario, as in the link that you shared, is that to mount the hostPath /, basically you are privileged and there will be no CVE associated due to a privilege container escaping. However, _maybe_ we can do better to prevent this.

I'd reduce this to a minimal example: leave alone the daemonset, just create a pod, remove as many mounts as possible, etc. And share the minimal repro here too, if you can :)

rata: > The thing is that we are no more `privileged: true` in Inspektor Gadget since aaaaaages: #348.
> But, we sadly still need to use `CAP_SYS_ADMIN`: #2228.

This is confusing, but privileged can refer not just to that field, but to the `privileged` PSS. If you need hostPath or cap sys_admin, you are privileged in the PSS sense. And that is what will probably won't grant you a CVE.

When baking the small repro, please try removing capabilities. I wouldn't be shocked if some "magic-links" in proc are not so magic if you lose some capabilities or things like that.

mauriciovasquezbernal: I did a quick experiment, it's all about using `container.apparmor.security.beta.kubernetes.io/gadget: "unconfined"` with `SYS_PTRACE`, if I remove any of the two reading or writing /host/proc/1/root fails.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
  annotations:
    container.apparmor.security.beta.kubernetes.io/gadget: "unconfined"
spec:
  hostPID: false
  hostNetwork: false
  terminationGracePeriodSeconds: 0
  restartPolicy: Never
  containers:
    - name: gadget
      terminationMessagePolicy: FallbackToLogsOnError
      image: ubuntu:22.04
      command: [ "sleep", "inf" ]
      securityContext:
        capabilities:
          drop:
            - ALL
          add:
            - SYS_PTRACE
      volumeMounts:
        - mountPath: /host/proc
          name: proc
          readOnly: true
  volumes:
    - name: proc
      hostPath:
        path: /proc
```

```bash 
$ kubectl exec -it test-pod -- bash -c "touch /host/proc/1/root/foo"
$ ls /foo
/foo
```

rata: @mauriciovasquezbernal interesting! I wonder what SYS_PTRACE gives you to allow for that, but very nice. Today I learned!

PR #2298, opened 2023-12-14
Enum conversion: fix loop variable captured in closure

alban:
When iterating over the btf types, the loop variable was captured in the getter() closure. This caused enumSetter() to use the wrong btf type to get the enum value.

To avoid this problem, this patch makes sure that the closures use a copy of the loop variable.

Fixes: dcf979cbd68def8c793b6322bd3a77a5bd388d14

Problem discovered when trying to implement enums in the dns gadget: https://github.com/inspektor-gadget/inspektor-gadget/pull/2204#discussion_r1425822831


mauriciovasquezbernal (approved):
Good catch. Thanks for the quick fix.

burak-ok (approved):
LGTM, too. Thanks for spotting this

PR #2300, opened 2023-12-15
ci: Use v3 for upload-artifact.

eiffel-fl:
There is currently a bug with v4 of download-artifact and upload-artifact [1].
So, let's use v3 for both for the moment.


mauriciovasquezbernal (approved):
LGTM!

PR #2304, opened 2023-12-15
trace_exec: Handle args column with wasm

mauriciovasquezbernal:
Parse the args columns by using a wasm module.

TODO: It works fine for the columns output, however in the json mode
    the output shows a string while in the builtin gadget it's a json
    array.
    
TODO:
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/2204


mauriciovasquezbernal:
I'll take of this once we have wasm support available.

Issue #2190, opened 2023-10-30
[Discussion] Default registries for image-based gadgets


burak-ok:
## Current situation

Currently if we want to push/pull/run/.. we have to specify the complete url:

```bash
$ ig run ghcr.io/inspektor-gadget/trace_exec
```

## Future situation

We introduce a default registry, so we do not need to specify the full url

```bash
$ ig run trace_exec
```
or
```bash
$ ig run inspektor-gadget/trace_exec
```

PR #2307, opened 2023-12-18
oci: Use default registry for Inspektor Gadget

mauriciovasquezbernal:
See https://github.com/inspektor-gadget/inspektor-gadget/issues/2190

## Before

It's not possible to run a gadget by just specifying their name:

```bash
$ sudo -E ig image list
REPOSITORY                                  TAG                                         DIGEST 

$ sudo -E ig run trace_exec
Error: getting gadget info: getting gadget image: pulling image "trace_exec": creating remote repository: creating remote repository: invalid reference: missing repository
```

## After

Now it's possible

```bash 
$ sudo -E ig image list
REPOSITORY                                  TAG                                         DIGEST   

$ sudo -E ig run trace_exec
RUNTIME.CONTAINERNAME     PID            PPID           UID           GID           R… COMM         
```

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/2190


pkg/oci/oci.go
```diff
@@ -372,31 +387,29 @@ func DeleteGadgetImage(ctx context.Context, image string) error {
...
-func getRepositoryFromImage(image string) (string, error) {
-	repo, err := reference.Parse(image)
-	if err != nil {
-		return "", fmt.Errorf("parsing image %q: %w", image, err)
+// splitIGDomain splits a repository name to domain and remote-name.
+// If no valid domain is found, the default domain is used. Repository name
+// needs to be already validated before.
+func splitIGDomain(name string) (domain, remainder string) {
+	i := strings.IndexRune(name, '/')
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != localhost && strings.ToLower(name[:i]) == name[:i]) {
```
mqasimsarfraz, line 395: Just for my understanding, for `strings.ToLower(name[:i]) == name[:i])` we assume it it a domain name since it contains capital letters?

mqasimsarfraz (commented):
I tested it and works fine. It will simplify running official gadgets.

I have only one question regarding UX with `list` command. I am wondering if it makes sense to not show the default domain / default repo prefix when list the images. This should be similar to what docker does:

```
→ docker images
REPOSITORY   TAG       IMAGE ID   CREATED   SIZE
→ docker pull docker.io/library/alpine
Using default tag: latest
latest: Pulling from library/alpine
661ff4d9561e: Pull complete 
Digest: sha256:51b67269f354137895d43f3b3d810bfacd3945438e94dc5ac55fdac340352f48
Status: Downloaded newer image for alpine:latest
docker.io/library/alpine:latest
→ docker images
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
alpine       latest    f8c20f8bbcb6   5 weeks ago   7.38MB
```

Also, it will encourage to use names like `trace_open` for official gadget (after I just listed them using `list`). What do you think?

pkg/oci/oci.go
```diff
@@ -372,31 +387,29 @@ func DeleteGadgetImage(ctx context.Context, image string) error {
...
-}
-
-func getRepositoryFromImage(image string) (string, error) {
-	repo, err := reference.Parse(image)
-	if err != nil {
-		return "", fmt.Errorf("parsing image %q: %w", image, err)
+// splitIGDomain splits a repository name to domain and remote-name.
+// If no valid domain is found, the default domain is used. Repository name
+// needs to be already validated before.
+func splitIGDomain(name string) (domain, remainder string) {
```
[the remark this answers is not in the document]
burak-ok: Maybe something like `CreateImageUrl` and only return a single string. The only usage for this method just concats the return values anyway

burak-ok (commented):
Apart from the naming this looks good. I also added a unit test for the split function since it seems important :D
https://github.com/inspektor-gadget/inspektor-gadget/commit/90b0934c669c85b780b951734c02cc49f7c44768

mauriciovasquezbernal (commented):
> What do you think?

I agree. Updated the PR to trim the prefix when listing images.

pkg/oci/oci.go
mqasimsarfraz: >  domains are supported to be lower case, aren't they?

IIUC, domains are case insensitive, https://serverfault.com/a/261344

> It was introduced in distribution/reference@ ca2a562, the tests cases added parse the domain with upper case, not sure what this is for and the commit doesn't contain anymore
information.

I found more information here, https://github.com/distribution/distribution/pull/2979 and it gives more context why they did it.

pkg/oci/oci.go
```diff
@@ -259,16 +268,23 @@ func listGadgetImages(ctx context.Context, store *oci.Store) ([]*GadgetImageDesc
...
-			tag, err := getTagFromImage(fullTag)
-			if err != nil {
-				log.Debugf("getting tag from image %q: %s", fullTag, err)
-				continue
+
+			var repository string
+			if named, ok := parsed.(reference.Named); ok {
+				repository = named.Name()
+			}
+			repository = strings.TrimPrefix(repository, defaultDomain+"/"+officialRepoPrefix)
```
mqasimsarfraz, line 281: nit: Perhaps we can trim the `defaultDomain` for cases like `ghcr.io/foo/bar` to be `foo/bar`

mqasimsarfraz (approved):
LGTM! I tested it and it works as described in the PR description.

pkg/oci/oci.go
mauriciovasquezbernal: I prefer to trim it only when the full thing matches. I'm don't think we should be treating any other think than "ghcr.io/inspektor-gadget/gadgets" in a special way.

PR #2308, opened 2023-12-18
run: Add trace_signal.

eiffel-fl:
Hi.

In this PR, I added `trace signal` as image based gadget.
The `kill_only` parameter is the eBPF counterpart of [`kill-only`](https://github.com/inspektor-gadget/inspektor-gadget/blob/3184a4985cd85e72330e7403e98a3505b681b97d/pkg/gadgets/trace/signal/tracer/tracer.go#L136) as we cannot choose which tracepoint we want to trace with `run`.

Best regards.

### DIFF
--- a/integration/inspektor-gadget/run_trace_signal_test.go
+++ b/integration/inspektor-gadget/run_trace_signal_test.go
@@ -1,0 +1,101 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


gadgets/trace_signal/gadget.yaml
```diff
@@ -0,0 +1,54 @@
...
+  target_signal:
+    key: signal
+    defaultValue: ""
+    description: Show only events generated by this signal
+  failed_only:
+    key: failed
+    defaultValue: "false"
+    description: Show only failed events
+  kill_only:
+    key: kill
```
mauriciovasquezbernal, line 52: `kill-only`  and add `alias: k` to match trace signal?

gadgets/trace_signal/program.bpf.c
```diff
@@ -0,0 +1,214 @@
...
+	return 0;
+}
+
+SEC("tracepoint/syscalls/sys_enter_kill")
+int ig_sig_kill_e(struct trace_event_raw_sys_enter *ctx)
+{
+	pid_t tpid = (pid_t)ctx->args[0];
+	int sig = (int)ctx->args[1];
+
+	if (kill_only)
```
mauriciovasquezbernal, line 115: What about moving this check inside `probe_entry`, it seems it's done in all places. Same for probe_exit.

mauriciovasquezbernal (commented):
It looks very good. Please add this to the Makefile above. 

Are you planning to add integration tests as well?

mauriciovasquezbernal (approved):
LGTM.

@eiffel-fl would you mind please documenting (as github issues) the things this version is missing to have the same exact functionality as the built-in one? 

I can think of:
- Signal number to name transformation
- Way to attach programs conditionally based on param (it was fixed by having the conditionals on the programs, but I think this feature could be useful for other gadgets as well)

What else?

PR #2317, opened 2023-12-20
ci: Skip public key check when run by dependabot or from fork. 

eiffel-fl:
Jobs running from fork do not have access to secrets.
The same applies to dependabot, unless explicitly set in the secrets
configuration.
So, we skip this job when run from a fork or by dependabot.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/commit/7c004217391a4f9bd4a48bcee143fed80e3c50e0 ("Add Inspektor Gadget public key to source.")

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1080,6 +1080,13 @@ jobs:
     # level: 1
     needs: check-secrets
     runs-on: ubuntu-latest
+    # When this job is run by a PR coming from a fork, it will not have access
+    # to secrets.
+    # Also, unless specified in the secrets configuration, dependabot cannot access
+    # to secrets.
+    # So, we skip this job when run by dependabot and from a PR coming from a
+    # fork.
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
> Dependabot does not have access to secrets unless explicitly set in the secrets configuration.

What's happens with external contributions, isn't it the same case?

mauriciovasquezbernal:
> I guess it does not occur with them, as we (maintainers) decide to run the pipeline, so it should be able to access secrets like us.

It won't work as stated in https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow : 

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

I tried it out on https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7277762855/job/19830526937?pr=2321 and it fails. So I think we need to restrict it to only runs on PRs from this repository.

mauriciovasquezbernal (approved):

PR #2323, opened 2023-12-22
Convert existing gadgets to use buffer.h API

eiffel-fl:
Hi.

This PR converts existing gadgets to use buffer.h API.

Best regards.

PR #2331, opened 2023-12-26
Extend buffer.h

eiffel-fl:
Hi.

This PR extends `buffer.h` by adding `gadget_output_buf()` and proper documentation as well.
It also converts `trace_exec` to use this API with `gadget_output_buf()`.

Best regards.

### DIFF
--- a/docs/reference/gadget-helper-api.md
+++ b/docs/reference/gadget-helper-api.md
@@ -10,14 +10,15 @@ The gadget helper API is a set of features exposed by Inspektor Gadget that is a
 
 To make use of container enrichment, gadgets must include
 [gadget/mntns.h](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/include/gadget/mntns.h):
-```
+
+```C
 #include <gadget/mntns.h>
 ```
 
 Inspektor Gadget will automatically enrich events with container information
 when the events include the mount namespace in this way:
 
-```
+```C
 struct event {
         gadget_mntns_id mntns_id;
         /* other fields */
@@ -25,7 +26,8 @@ struct event {
 ```
 
 An eBPF program can look up the mount namespace with the following:
-```
+
+```C
 u64 mntns_id;
 mntns_id = gadget_get_mntns_id();
 ```
@@ -37,7 +39,8 @@ use socket enrichment to find the mount namespace.
 
 To make use of container filtering, gadgets must include
 [gadget/mntns_filter.h](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/include/gadget/mntns_filter.h):
-```
+
+```C
 #include <gadget/mntns_filter.h>
 ```
 
@@ -55,19 +58,22 @@ To make use of socket enrichment, gadgets must include
 [gadget/sockets-map.h](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/include/gadget/sockets-map.h).
 
 For eBPF programs of type socket filter:
-```
+
+```C
 #define GADGET_TYPE_NETWORKING
 #include <gadget/sockets-map.h>
 ```
 
 For eBPF programs of other kinds:
-```
+
+```C
 #define GADGET_TYPE_TRACING
 #include <gadget/sockets-map.h>
 ```
 
 This will define the following struct:
-```
+
+```C
 #define TASK_COMM_LEN 16
 struct sockets_value {
         __u64 mntns;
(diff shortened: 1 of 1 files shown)


include/gadget/buffer.h
```diff
@@ -18,6 +18,7 @@
 	} name SEC(".maps");				\
 	const void *gadget_map_tracer_##name __attribute__((unused));
 
+#ifndef GADGET_NO_RESERVE
```
mauriciovasquezbernal, line 21: ref https://github.com/inspektor-gadget/inspektor-gadget/pull/2323#discussion_r1435262889

I'm still not convinced by this one, I don't like that we're adding an additional knob when it's not fully needed.

I see two options: 

1. Use `MAX_EVENT_SIZE` to control it, if it's defined then the gadget_reserve_buf / gadget_submit_buf will be defined, otherwise them won't be available. 
1. Split buffer.h in two header files, buffer-output.h (or similar) will contain the map definition and `gadget_output_buf`, buffer.h will include buffer-output.h, define the gadget_heap and other functions.

docs/reference/gadget-helper-api.md
```diff
@@ -119,3 +125,90 @@ struct event {
...
+
+1. `void *gadget_reserve_buf(void *map, __u64 size)`: Reserves memory in the corresponding buffer.
+1. `long gadget_submit_buf(void *ctx, void *map, void *buf, __u64 size)`: Writes the previously reserved memory in the corresponding buffer.
+1. `void gadget_discard_buf(void *buf)`: Discards the previously reserved buffer. This is needed to avoid wasting memory.
+1. `long gadget_output_buf(void *ctx, void *map, void *buf, __u64 size)`: Reserves and writes the buffer in the corresponding map. This is equivalent to calling `gadget_reserve_buf()` and `gadget_submit_buf()`.
+
+The following snippet demonstrates how to use the code available in `<gadget/buffer.h>`, it is taken from `trace_open`:
+
+```C
+#include <gadget/buffer.h>
```
mauriciovasquezbernal, line 152: What about suggesting always setting `MAX_EVENT_SIZE`?

docs/reference/gadget-helper-api.md
```diff
@@ -119,3 +125,90 @@ struct event {
 * `struct gadget_l3endpoint_t` and `struct gadget_l4endpoint_t`: enrich with the Kubernetes endpoint. TODO: add details.
 * `typedef __u64 gadget_mntns_id`: container enrichment (see #container-enrichment)
 * `typedef __u64 gadget_timestamp`: add human-readable timestamp from `bpf_ktime_get_boot_ns()`.
+
+## Buffer API
+
+To share events with userspace, we rely on perf or eBPF ring buffers.
+Compare to perf ones, eBPF ring buffers are more recent, so they may not be available on your kernel.
+To ease using these buffers, we provide `<gadget/buffer.h>`.
+This header contains useful abstraction around the underlying type of buffer.
```
mauriciovasquezbernal, line 134: What about something like this?

```suggestion
There are two kind of eBPF maps used to send events to userspace: (a) perf ring buffer or (b) eBPF
ring buffer. The later is more efficient and flexible, however it's only available from kernel 5.8.
Check https://nakryiko.com/posts/bpf-ringbuf/ to get more details about the differences.
`<gadget/buffer.h>` provides an abstraction to automatically use the right buffer according to the
kernel features.
```

include/gadget/buffer.h
mauriciovasquezbernal: > If every time I would need to #include a library I would need to #define something this would definitely make me refrain using this library.

The second option doesn't require this.

> The thing we should do here is to have the gadget_heap, if used, size set to struct eventin golang code with some magic to round it to the nearest power of 2.

Why to involve golang if this can be done directly in c? Also the heap doesn't have to be a power of 2 (as opposed to the ring size)

mauriciovasquezbernal: > And then, we need to include two times...

The idea is to have buffer.h including buffer-ouput.h, hence from the user POV it's only needed to include one of the two files.

But I don't care too much about it TBH, if you still think the macro is the best approach let's go that way.

> Indeed for the power of pwo.
> By using golang no one has to compute offline the size of the struct as it would be done by go for users.

Makes sense.

mauriciovasquezbernal, line 21: Please rename to something more specific, like `GADGET_NO_BUF_RESERVE`.

mauriciovasquezbernal (approved):
One final nit. LGTM otherwise.

PR #2333, opened 2023-12-27
gadgets: Convert trace oomkill to image based gadget

eiffel-fl:
Hi.

This PR converts `trace oomkill` to image based gadget:

```bash
$ sudo -E ./ig run oomkills &                         francis/img-oom % u=
[1] 90799
RUNTIME.CONTAINERNAME          FPID             FUID             FGID             TPID             PAGES            FCOMM            TCOMM                                                           
$ docker run --name test-trace-oomkill -m 512M -it --rm busybox tail /dev/zeroo
test-trace-oomkill             90914            0                0                90914            262144           tail             tail
```

Best regards.

### DIFF
--- a/integration/inspektor-gadget/run_trace_oomkill_test.go
+++ b/integration/inspektor-gadget/run_trace_oomkill_test.go
@@ -1,0 +1,126 @@
+// Copyright 2023 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/run/types"
+
+	. "github.com/inspektor-gadget/inspektor-gadget/integration"
+)
+
+func runTraceOOMKill(t *testing.T, ns string, cmd string) {
+	// TODO: Handle it once we support getting container image name from docker
+	isDockerRuntime := IsDockerRuntime(t)
+
+	traceOOMKillCmd := &Command{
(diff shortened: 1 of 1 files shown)


gadgets/trace_oomkill/program.bpf.c
```diff
@@ -0,0 +1,73 @@
...
+#include <gadget/buffer.h>
+#include <gadget/macros.h>
+#include <gadget/mntns_filter.h>
+#include <gadget/types.h>
+
+#define TASK_RUNNING 0
+#define NAME_MAX 255
+#define INVALID_UID ((uid_t)-1)
+
+struct args_t {
```
burak-ok, line 18: unused?

integration/inspektor-gadget/run_trace_oomkill.go
```diff
@@ -0,0 +1,127 @@
...
+				m["mntns_id"] = 0
+				m["timestamp"] = ""
+				m["fcomm"] = ""
+			}
+
+			ExpectEntriesToMatchObj(t, output, normalize, expectedJsonObj)
+		},
+	}
+
+	limitPodYaml := fmt.Sprintf(`
```
burak-ok, line 73: Do you think we could generalize https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/command.go#L561 even more? So additional stuff we need (here `resources`) can be used?
I just don't like inline yaml :D

burak-ok: What I meant is something like https://github.com/inspektor-gadget/inspektor-gadget/commit/9abf9d9bca4ee8885d3ac068ab57aa0d59d1227d where we keep the old `PodCommand` but add another function.

integration/inspektor-gadget/run_trace_oomkill.go
burak-ok, line 1: The file has to have `_test.go` as a suffix. Currently it is not getting executed at all

gadgets/trace_oomkill/gadget.yaml
```diff
@@ -0,0 +1,35 @@
+name: oomkill
+description: trace OOM killer
+tracers:
+  oomkill:
```
burak-ok, line 4:
```suggestion
  events:
```

When running `ig build` with `--update-metadata` it does the following:
```yaml
+ events:
+   mapName: events
+   structName: event
  oomkill:
    mapName: events
    structName: event
```
Then the build throws the following error:
```bash
Error: validating metadata file: 1 error occurred:
	* only one tracer is allowed
```

burak-ok (approved):
Tested it and it works as advertised. LGTM

PR #2347, opened 2024-01-09
use client-go informers instead of active polling

matthyx:
# use client-go informers instead of active polling

Use https://pkg.go.dev/k8s.io/client-go/informers instead of active polling in the `K8sInventoryCache` component.

## How to use

## Testing done

I don't know exactly how to test this change, help needed here.

### DIFF
--- a/pkg/operators/common/kubeinventorycache.go
+++ b/pkg/operators/common/kubeinventorycache.go
@@ -15,16 +15,14 @@
 package common
 
 import (
-	"context"
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
-	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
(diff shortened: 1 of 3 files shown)


burak-ok (commented):
LGTM. Thanks for the improvement and PR.
Really appreciated :+1:

pkg/operators/common/kubeinventorycache.go
```diff
@@ -53,43 +49,32 @@ var (
...
 	return cache, err
 }
 
-func newCache(tickerDuration time.Duration) (*K8sInventoryCache, error) {
+func newCache(defaultResync time.Duration) (*K8sInventoryCache, error) {
 	clientset, err := k8sutil.NewClientset("")
 	if err != nil {
 		return nil, fmt.Errorf("creating new k8s clientset: %w", err)
 	}
+	factory := informers.NewSharedInformerFactory(clientset, defaultResync)
```
burak-ok, line 62: Do we need a resync period at all? (Really a question, don't have the experience with K8s informers)

matthyx: In most cases it's not needed, I can remove it.

burak-ok: Doesn't need to be changed - was just curious

burak-ok (approved):
PR tests ran successfully and also tried it locally.
Thanks again for the contribution

PR #2348, opened 2024-01-09
pkg/gadgets: Add cifs support to trace fsslower.

eiffel-fl:
Hi.

In this PR, I added support to `cifs` in `trace fsslower`:

```bash
root@vm-amd64:~# ./share/kinvolk/inspektor-gadget/ig trace fsslower --host -f cifs -m 0 > /tmp/cifs &
...
root@vm-amd64:~# more /tmp/cifs 
RUNTIME.CONTAINERNAME          PID              COMM             T      BYTES   
  OFFSET        LAT FILE                    
                               338              cat              O          0   
       0        962 perf.data               
                               339              cat              O          0   
       0       1467 perf.data               
                               340              cat              O          0   
       0        987 perf.data
```

Best regards.

### DIFF
--- a/pkg/gadgets/trace/fsslower/tracer/gadget.go
+++ b/pkg/gadgets/trace/fsslower/tracer/gadget.go
@@ -63,7 +63,7 @@ func (g *GadgetDesc) ParamDescs() params.ParamDescs {
 			Title:          "Filesystem",
 			DefaultValue:   "ext4",
 			Description:    "Filesystem to trace",
-			PossibleValues: []string{"btrfs", "ext4", "nfs", "xfs"},
+			PossibleValues: []string{"btrfs", "cifs-direct", "cifs-loose", "cifs-strict", "ext4", "nfs", "xfs"},
 		},
 	}
 }
--- a/pkg/gadgets/trace/fsslower/tracer/tracer.go
+++ b/pkg/gadgets/trace/fsslower/tracer/tracer.go
@@ -75,6 +75,27 @@ var fsConfMap = map[string]fsConf{
 		fsync:  "btrfs_sync_file",
 		statfs: "btrfs_statfs",
 	},
+	"cifs-direct": {
+		read:   "cifs_direct_readv",
+		write:  "cifs_direct_writev",
+		open:   "cifs_open",
+		fsync:  "cifs_fsync",
+		statfs: "cifs_statfs",
+	},
+	"cifs-loose": {
+		read:   "cifs_loose_read_iter",
+		write:  "cifs_file_write_iter",
+		open:   "cifs_open",
+		fsync:  "cifs_fsync",
+		statfs: "cifs_statfs",
+	},
+	"cifs-strict": {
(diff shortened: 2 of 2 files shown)


alban:
It seems cifs uses different functions depending on the mount options:
https://github.com/torvalds/linux/blob/9f8413c4a66f2fb776d3dc3c9ed20bf435eb305e/fs/smb/client/inode.c#L37-L51

Can you try with various options?
- `mount.cifs -o cache=none|loose|strict`
- `mount.cifs -o nobrl`

I think you have to either attach the ebpf programs to additional functions, or to document the restrictions in the [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/builtin-gadgets/trace/fsslower.md).

jepio:
> > It seems cifs uses different functions depending on the mount options: https://github.com/torvalds/linux/blob/9f8413c4a66f2fb776d3dc3c9ed20bf435eb305e/fs/smb/client/inode.c#L37-L51
> > Can you try with various options?
> > ```
> > * `mount.cifs -o cache=none|loose|strict`
> > 
> > * `mount.cifs -o nobrl`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think you have to either attach the ebpf programs to additional functions, or to document the restrictions in the [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/builtin-gadgets/trace/fsslower.md).
> 
> I saw this. Sadly, the only way to support this would be to have several definition for `cifs`, like `cifs-none`, `cifs-loose`, _etc._. I will add them, so we cover everything.

With this user experience I don't think this would be useful. A user shouldn't have to know what mount option was specified to trace vfs operations. This shows that a cifs specific tool is needed.

alban:
> With this user experience I don't think this would be useful. A user shouldn't have to know what mount option was specified to trace vfs operations. This shows that a cifs specific tool is needed.

I agree. This could be solved if the code defines a list of functions instead of a single set:
```diff
-fsConfMap = map[string]fsConf
+fsConfMap = map[string][]fsConf
```

pkg/gadgets/trace/fsslower/tracer/gadget.go
```diff
@@ -63,7 +63,7 @@ func (g *GadgetDesc) ParamDescs() params.ParamDescs {
 			Title:          "Filesystem",
 			DefaultValue:   "ext4",
 			Description:    "Filesystem to trace",
-			PossibleValues: []string{"btrfs", "ext4", "nfs", "xfs"},
+			PossibleValues: []string{"btrfs", "cifs-direct", "cifs-loose", "cifs-strict", "ext4", "nfs", "xfs"},
```
alban, line 66: This only covers parts of the possible mount options: it does not handle the CIFS_MOUNT_NO_BRL option (`-o nobrl` is orthogonal to the `-o cache=` options.

PR #2352, opened 2024-01-10
Add `watch` permission for `services`

burak-ok:
After https://github.com/inspektor-gadget/inspektor-gadget/pull/2347 changed the way we keep a up-to-date list of all pods and services (watchers instead of polling) everything seem to work.
IPs get resolved successfully to pod names and service names.

But in the gadget logs we see the following:
`1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.29.0/tools/cache/reflector.go:229: Failed to watch *v1.Service: unknown (get services)`

This PR adds the missing watch permissions for our ServiceAccount.

(I also added logging of the returned errors, but in this case they didn't help at all and logged nothing -> no error)

### DIFF
--- a/charts/gadget/templates/clusterrole.yaml
+++ b/charts/gadget/templates/clusterrole.yaml
@@ -12,8 +12,9 @@ rules:
     verbs: ["get", "watch", "list"]
   - apiGroups: [""]
     resources: ["services"]
-    # list services is needed by network-policy gadget.
-    verbs: ["list"]
+    # list is needed by network-policy gadget
+    # watch is needed by operators enriching with service informations
+    verbs: ["list", "watch"]
   - apiGroups: ["gadget.kinvolk.io"]
     resources: ["traces", "traces/status"]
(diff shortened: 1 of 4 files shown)


mauriciovasquezbernal (approved):
LGTM from code inspection.

burak-ok:
> Hi!
> 
> I would like to test it, any way to do so? Just `deploy` and I should check there is not the above mentioned message?
> 
> Best regards.

Correct. Best way would be check on main and then check with this PR.

alban (approved):
LGTM from code inspection too.

burak-ok:
> > Hi!
> > I would like to test it, any way to do so? Just `deploy` and I should check there is not the above mentioned message?
> > Best regards.
> 
> Correct. Best way would be check on main and then check with this PR.

Additionally you need to start a gadget which uses the Cache. So something like `trace tcp` or `trace network` should do the trick

PR #2354, opened 2024-01-10
K8sInventoryCache: Manage resources better + fix rare race condition

burak-ok:
## K8sInventoryCache

`Close()` should really only be called by `Stop()` and not by an operator directly.
If this would happen the enrichment for every other running gadget would fail (no new pods/services are getting registered)

The problem with the rare race condition was the following (This issue was before 0071cee0ffc63da02c862eea99ecf589cac1d266):
> In `Close()` the channel was closed and set to `nil`
> In the same time the loop had a chance to query the `exit` channel variable which was set to `nil` -> :boom: 

After this commit the race condition could only happen in the following situation:
> K8sInventoryCache.Start() is called and gets scheduled away after it created the channel exit
> K8sInventoryCache.Close() is called and exit gets assigned to nil
> K8sInventoryCache.Start() resumes and now gives exit as a parameter to factory.Start and factory.WaitForCacheSync

### DIFF
--- a/pkg/operators/common/kubeinventorycache.go
+++ b/pkg/operators/common/kubeinventorycache.go
@@ -22,6 +22,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
 	listersv1 "k8s.io/client-go/listers/core/v1"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
@@ -30,16 +31,21 @@ import (
 // K8sInventoryCache is a cache of Kubernetes resources such as pods and services
 // that can be used by operators to enrich events.
 type K8sInventoryCache struct {
-	factory informers.SharedInformerFactory
+	clientset *kubernetes.Clientset
 
-	pods listersv1.PodLister
-	svcs listersv1.ServiceLister
+	factory informers.SharedInformerFactory
+	pods    listersv1.PodLister
+	svcs    listersv1.ServiceLister
 
 	exit chan struct{}
 
 	useCount      int
 	useCountMutex sync.Mutex
 }
+
+const (
+	informerResync = 10 * time.Minute
+)
 
 var (
 	cache *K8sInventoryCache
@@ -49,24 +55,19 @@ var (
 
 func GetK8sInventoryCache() (*K8sInventoryCache, error) {
 	once.Do(func() {
-		cache, err = newCache(10 * time.Minute)
+		cache, err = newCache()
 	})
 	return cache, err
 }
 
-func newCache(defaultResync time.Duration) (*K8sInventoryCache, error) {
+func newCache() (*K8sInventoryCache, error) {
 	clientset, err := k8sutil.NewClientset("")
 	if err != nil {
 		return nil, fmt.Errorf("creating new k8s clientset: %w", err)
 	}
-	factory := informers.NewSharedInformerFactory(clientset, defaultResync)
-	pods := factory.Core().V1().Pods().Lister()
-	svcs := factory.Core().V1().Services().Lister()
 
 	return &K8sInventoryCache{
-		factory: factory,
-		pods:    pods,
-		svcs:    svcs,
+		clientset: clientset,
 	}, nil
 }
 
@@ -74,6 +75,10 @@ func (cache *K8sInventoryCache) Close() {
 	if cache.exit != nil {
 		close(cache.exit)
 		cache.exit = nil
+	}
+	if cache.factory != nil {
+		cache.factory.Shutdown()
+		cache.factory = nil
 	}
 }
 
@@ -83,6 +88,10 @@ func (cache *K8sInventoryCache) Start() {
 
 	// No uses before us, we are the first one
 	if cache.useCount == 0 {
+		cache.factory = informers.NewSharedInformerFactory(cache.clientset, informerResync)
+		cache.pods = cache.factory.Core().V1().Pods().Lister()
+		cache.svcs = cache.factory.Core().V1().Services().Lister()
+
 		cache.exit = make(chan struct{})
 		cache.factory.Start(cache.exit)
 		cache.factory.WaitForCacheSync(cache.exit)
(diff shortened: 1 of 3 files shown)


pkg/gadget-service/service.go
```diff
@@ -142,6 +142,7 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
 	if err != nil {
 		return fmt.Errorf("initialize operators: %w", err)
 	}
+	defer operators.GetAll().Close()
```
flyth, line 145: Operators are lazily loaded here, but they shouldn't be closed unless the service process is stopped (or we find a good way to schedule `.Close()` calls after a certain timeout to free resources) as other running gadgets might still use them.

burak-ok: I remembered so too, that `Init()` and `Close()` is shared between all of currently running gadgets.

But then I saw this `Init()` which is called for every gadget run and it confused me that I don't see the same amount of `Close()` for that.

burak-ok: I removed the commit containing the `close` since the other commit is one I want to merged as soon as possible

pkg/operators/common/kubeinventorycache.go
```diff
@@ -49,24 +55,19 @@ var (
...
 func GetK8sInventoryCache() (*K8sInventoryCache, error) {
 	once.Do(func() {
-		cache, err = newCache(10 * time.Minute)
+		cache, err = newCache()
 	})
 	return cache, err
 }
 
-func newCache(defaultResync time.Duration) (*K8sInventoryCache, error) {
+func newCache() (*K8sInventoryCache, error) {
```
mauriciovasquezbernal, line 63: I'm wondering if we really need this function anymore. At this point we can move the client set creation into the `cache.useCount == 0` conditional and just return `&K8sInventoryCache{}` on `GetK8sInventoryCache`.

mauriciovasquezbernal (commented):
> The problem with the rare race condition was the following:
In Close() the channel was closed and set to nil
In the same time the loop had a chance to query the exit channel variable which was set to nil -> 💥

I'm trying to understand how the current code fixes that. I suppose it's because it's also calling `cache.factory.Shutdown()`?

pkg/operators/common/kubeinventorycache.go
burak-ok: I left the client set creation here so it will be done only once, because it is really only needed once

burak-ok:
> > The problem with the rare race condition was the following:
> > In Close() the channel was closed and set to nil
> > In the same time the loop had a chance to query the exit channel variable which was set to nil -> 💥
> 
> I'm trying to understand how the current code fixes that. I suppose it's because it's also calling `cache.factory.Shutdown()`?

My newest edit didn't save. The situation I described was before commit 0071cee0ffc63da02c862eea99ecf589cac1d266

After this commit the race condition could only happen in the following situation:
1. `K8sInventoryCache.Start()` is called and gets scheduled away after it created the channel `exit`
2. `K8sInventoryCache.Close()` is called and `exit` gets assigned to nil
3. `K8sInventoryCache.Start()` resumes and now gives `exit` as a parameter to `factory.Start` and `factory.WaitForCacheSync`

Though I don't know if this would give us a panic or the 2 methods can handle a `nil` parameter

pkg/operators/common/kubeinventorycache.go
mauriciovasquezbernal: But the creation is not very expensive, right? Perhaps it's more expensive to keep it on memory than creating it each time it's needed. However it's not a big blocker for me.

mauriciovasquezbernal (approved):
Changes make sense to me from code inspection, so I'll approve.

pkg/operators/common/kubeinventorycache.go
```diff
@@ -83,6 +88,10 @@ func (cache *K8sInventoryCache) Start() {
 
 	// No uses before us, we are the first one
 	if cache.useCount == 0 {
+		cache.factory = informers.NewSharedInformerFactory(cache.clientset, informerResync)
+		cache.pods = cache.factory.Core().V1().Pods().Lister()
+		cache.svcs = cache.factory.Core().V1().Services().Lister()
+
 		cache.exit = make(chan struct{})
 		cache.factory.Start(cache.exit)
 		cache.factory.WaitForCacheSync(cache.exit)
```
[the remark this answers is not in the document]
burak-ok: If a gadget needs the K8sInventoryCache for enrichment this will get instantiated.
After it has finished running everything should be stopped since it is not needed anymore. Keep memory footprint down (I don't know what the factory stores internally. And if it still consumes some amount of cpi time)

pkg/operators/kubeipresolver/kubeipresolver.go
```diff
@@ -66,30 +64,29 @@ func (k *KubeIPResolver) CanOperateOn(gadget gadgets.GadgetDesc) bool {
 }
 
 func (k *KubeIPResolver) Init(params *params.Params) error {
```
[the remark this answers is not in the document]
burak-ok: Yes, the interface can be found at https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/operators.go#L41

mauriciovasquezbernal:
I think it also fixes the following issue: Try to run TestTraceNetwork twice:


```bash 
INTEGRATION_TESTS_PARAMS="-run TestTraceNetwork -no-deploy-ig -no-deploy-spo" make integration-tests -o kubectl-gadget
```

works fine 

```bash
INTEGRATION_TESTS_PARAMS="-run TestTraceNetwork -no-deploy-ig -no-deploy-spo" make integration-tests -o kubectl-gadget
```

Fails because the reported event doesn't have the IP address enriched (see "kind":"raw")

```
{"runtime":{"runtimeName":"containerd","containerId":"e2c0afd9c48ccc5e9e2ec40ea4548b7dcdab27c34668111e351f0de09ebacd85","containerImageName":"docker.io/library/busybox:latest"},"k8s":{"node":"ubuntu-hirsute","namespace":"test-trace-network-4805623087645882167","podName":"test-pod","containerName":"test-pod"},"timestamp":1705003113904336868,"type":"normal","mountnsid":4026532991,"netnsid":4026532924,"pid":1538427,"tid":1538427,"comm":"wget","uid":0,"gid":0,"pktType":"OUTGOING","proto":"TCP","port":80,"dst":{"addr":"172.16.118.184","version":4,"kind":"raw"}}
```

According to my research it was introduced by 0071cee0ffc63da02c862eea99ecf589cac1d266, so would you mind adding a "fixes" tag here too?

Issue #2202, opened 2023-11-06
[RFE] handling events in userspace with wasm


alban:
## Current situation

Containerized gadgets have a ebpf module but currently don't yet contain code in userspace to handle events.

## Impact

Some gadgets lose features after they're converted from the build-in gadget to a containerized gadget.

Example: the dns gadget has a "NAME" column with a domain name with dots:
```
$ sudo ig trace dns -c test-trace-dns
QR TYPE      QTYPE      NAME
Q  OUTGOING  A          inspektor-gadget.io.
```
This string is built up from userspace. The ebpf program only sends the encoded name from the DNS packet. The containerized dns gadget does not have this string with dots.

## Ideal future situation

Gadget authors can include a wasm module in their gadgets to implement custom changes in events in userspace.

## Implementation options

We can use [wazero](https://github.com/tetratelabs/wazero), a WebAssembly runtime in Go that does not require dependencies with cgo.

## Additional information

- [design/002-containerized-gadgets.md](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md)
- Initial PoC demo https://github.com/inspektor-gadget/inspektor-gadget/pull/1985
- PoC for the ASG conference https://github.com/alban/inspektor-gadget/tree/alban_asg2023
- Discussion how to pass columns from server to client https://github.com/inspektor-gadget/inspektor-gadget/issues/2131
- First PR adding wasm in the gadget images: https://github.com/inspektor-gadget/inspektor-gadget/pull/2127
- Second PR executing wasm modules: https://github.com/inspektor-gadget/inspektor-gadget/pull/2204

Issue #2045, opened 2023-09-15
image-based gadgets: Support BTFGen


mauriciovasquezbernal:
[BTFGen](https://kinvolk.io/blog/2022/03/btfgen-one-step-closer-to-truly-portable-ebpf-programs/) helps to support programs in kernels that don't expose BTF information. We support this for builtin gadgets and we should support it for containerized gadgets too. It seems to me the easiest way would be to provide all BTF information in a new layer on the OCI image, then this information will be provided to the cilium ebpf library before loading the programs into the kernel. All of this should be integrated with the [image build command](https://github.com/inspektor-gadget/inspektor-gadget/pull/2020) so the user doesn't have to generate these files manually.

alban:
> It seems to me the easiest way would be to provide all BTF information in a new layer on the OCI image

There are many BTF files: one for each supported kernel version.

For the builtin gadgets, they are all embedded with a single `go:embed` in a `embed.FS` variable.

Should the BTF files be in a compressed tarball inside an OCI layer of type `application/vnd.gadget.btf.v1+binary`?

mauriciovasquezbernal:
> Should the BTF files be in a compressed tarball inside an OCI layer of type application/vnd.gadget.btf.v1+binary?

Yes, I think it's the easiest way to implement it while reusing the logic in https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/btfgen. 

> Or can a single BTF file embed several BTF dataset along with the kernel version they refer to?

That's not supported by BTF. It's designed to contain the information for a single kernel.

alban:
The BTFGen step in the CI takes 12 minutes. If we want to use BTFGen in the `image build` command, it needs to be optimized to take a reasonable time.

The BTFHub database should be indexed by symbol instead of indexed by kernel version.

mauriciovasquezbernal:
> The BTFGen step in the CI takes 12 minutes. If we want to use BTFGen in the image build command, it needs to be optimized to take a reasonable time.

The approach we have in Inspektor Gadget is to disable it for developing purposes, and it's only done in the CI, at release or when pushing to main. Also, btfhub-archive is few GBs big right now, so probably most of the users don't want to use it when developing on their local machines.

> The BTFHub database should be indexed by symbol instead of indexed by kernel version.

Could you please elaborate more on this idea?

PR #2355, opened 2024-01-11
oci: wasm modules are no longer stored in the wasm architecture

alban:
The wasm program is now in an OCI layer.

Before this patch, the OCI image had 3 architectures with only one layer each:
1. amd64
2. arm64
3. wasm

After this patch, the OCI image has 2 architectures:
1. amd64
2. arm64

And each architecture has 1 or 2 layers with different media types:
- application/vnd.gadget.ebpf.program.v1+binary
- application/vnd.gadget.wasm.program.v1+binary

This patch initially comes from #2204. But I would like to get it merged early before the rest of #2204 is ready.

This is a potentially breaking change. However, the currently published image-based gadgets don't make use of wasm yet, so no existing gadget shoud break.

### DIFF
--- a/pkg/oci/build.go
+++ b/pkg/oci/build.go
@@ -117,27 +117,20 @@ func pushDescriptorIfNotExists(ctx context.Context, target oras.Target, desc oci
 	return nil
 }
 
-func createEbpfProgramDesc(ctx context.Context, target oras.Target, progFilePath string, arch string) (ocispec.Descriptor, error) {
+func createLayerDesc(ctx context.Context, target oras.Target, progFilePath, mediaType string) (ocispec.Descriptor, error) {
 	progBytes, err := os.ReadFile(progFilePath)
 	if err != nil {
 		return ocispec.Descriptor{}, fmt.Errorf("reading eBPF program file: %w", err)
 	}
-	mediaType := ""
-	switch arch {
-	case ArchWasm:
-		mediaType = wasmObjectMediaType
-	default:
-		mediaType = eBPFObjectMediaType
-	}
 	progDesc := content.NewDescriptorFromBytes(mediaType, progBytes)
 	progDesc.Annotations = map[string]string{
-		ocispec.AnnotationTitle:       "program.o",
+		ocispec.AnnotationTitle:       "TODO: title",
 		ocispec.AnnotationAuthors:     "TODO: authors",
 		ocispec.AnnotationDescription: "TODO: description",
 	}
 	err = pushDescriptorIfNotExists(ctx, target, progDesc, bytes.NewReader(progBytes))
 	if err != nil {
-		return ocispec.Descriptor{}, fmt.Errorf("pushing eBPF program: %w", err)
+		return ocispec.Descriptor{}, fmt.Errorf("pushing %q layer: %w", mediaType, err)
 	}
 
 	return progDesc, nil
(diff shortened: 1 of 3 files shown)


alban:
This PR is a prerequisite for the rest of the work on:
- #2202 (#2204)
- #2045

pkg/oci/oci.go
```diff
@@ -553,18 +560,26 @@ func getMetadataFromManifest(ctx context.Context, target oras.Target, manifest *
...
+	var layer ocispec.Descriptor
+	layerCount := 0
+	for _, l := range manifest.Layers {
+		if l.MediaType == mediaType {
+			layer = l
+			layerCount++
+		}
+	}
+	if layerCount != 1 {
+		return nil, fmt.Errorf("expected exactly one layer with media type %q, got %d", mediaType, layerCount)
```
burak-ok, line 573: This makes the `wasm` layer mandatory and that shouldn't be the case

burak-ok (commented):
Looks good. I like it that it is stored in its own layer now

pkg/oci/oci.go
alban: Thanks. I fixed it and also updated the reference documentation accordingly.

pkg/oci/oci.go
```diff
@@ -109,15 +110,23 @@ func GetGadgetImage(ctx context.Context, image string, authOpts *AuthOptions, pu
...
 	if err != nil {
 		return nil, fmt.Errorf("getting arch manifest: %w", err)
 	}
 
-	prog, err := getEbpfProgramFromManifest(ctx, imageStore, manifest)
+	prog, err := getLayerFromManifest(ctx, imageStore, manifest, eBPFObjectMediaType)
 	if err != nil {
 		return nil, fmt.Errorf("getting ebpf program: %w", err)
 	}
+	if len(prog) == 0 {
```
burak-ok, line 122: This works since `len(prog) == 0` if `prog = nil`. But this empty check is already done in `getLayerFromManifest`.

I prefer to write `prog == nil` with an error message "no ebpf program found" or "layer"

burak-ok (approved):
LGTM

pkg/oci/build.go
```diff
@@ -117,18 +117,11 @@ func pushDescriptorIfNotExists(ctx context.Context, target oras.Target, desc oci
...
-	mediaType := ""
-	switch arch {
-	case ArchWasm:
-		mediaType = wasmObjectMediaType
-	default:
-		mediaType = eBPFObjectMediaType
-	}
 	progDesc := content.NewDescriptorFromBytes(mediaType, progBytes)
 	progDesc.Annotations = map[string]string{
 		ocispec.AnnotationTitle:       "program.o",
```
mauriciovasquezbernal, line 127: nit: this annotation is not valid anymore as it could be wasm too.

mauriciovasquezbernal (approved):
I think it's a good cleanup. LGTM from code inspection.

alban:
The CI still fails on cri-o despite the PR being rebased after #2367.

burak-ok:
> The CI still fails on cri-o despite the PR being rebased after #2367.

Looking at the failed logs:
It appears that the expected event is there, but it is not enriched. It has no `runtime` information, nor `k8s` information. Other events like the one from the `nginx-pod` are correctly enriched

alban:
Thanks for checking. This seems unrelated to this PR, so could I merge it?

mauriciovasquezbernal:
> Thanks for checking. This seems unrelated to this PR, so could I merge it?

Go for it!

PR #2356, opened 2024-01-11
docs: Add documentation on doing bugfix release.

eiffel-fl:
Hi.

This PR added documentation about doing bugfix release.

Best regards.

### DIFF
--- a/docs/devel/release-process.md
+++ b/docs/devel/release-process.md
@@ -62,6 +62,13 @@ $ git push origin v0.x.0
 
 ## Post release tasks
 
+- Create a branch for this release named `release-v0.x`, this will be used as stable branch where fixes would be backported:
+
+```bash
+$ git checkout -b release-v0.x
+$ git push --set-upstream origin release-v0.x
+```
+
 - Check if the [milestone for the release](https://github.com/inspektor-gadget/inspektor-gadget/milestones) still
   contain open issues. If so, move them as appropriate. Close the milestone.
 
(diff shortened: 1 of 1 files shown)


docs/devel/release-process.md
```diff
@@ -62,6 +62,13 @@ $ git push origin v0.x.0
 
 ## Post release tasks
 
+- Create a branch for this release named `v0.x-stable`, this will be used as stable branch where fixes would be backported:
+
+```bash
+$ git checkout -b v0.x-stable
```
mauriciovasquezbernal, line 68: I'd suggest to use release instead of stable as it's more clear:

```suggestion
$ git checkout -b v0.x-release
```

As done by Cilium

or `release-0.x` as done by Kubernetes

docs/devel/release-process.md
```diff
@@ -92,3 +99,59 @@ are issues that need to be handled before.
...
+v0.x.1
+v0.x.0
+v0.24.0
+v0.23.1
+```
+
+You now need to backport the patch to the latest stable branch:
+
+```bash
+$ git checkout v0.x-stable
```
mauriciovasquezbernal, line 135: `git pull` after it to be sure it's updated?

docs/devel/release-process.md
```diff
@@ -92,3 +99,59 @@ are issues that need to be handled before.
...
+$ git checkout -b v0.x-stable/fix-something
+$ git cherry-pick 95fe7405738d58476ddad29856ebe30599644666
+# If the patch does not apply cleanly, you will need to adapt it.
+# Once done, you can just push your branch and open a PR:
```
mauriciovasquezbernal, line 139: Please indicate the PR should be opened against the release branch and not main.

docs/devel/release-process.md
```diff
@@ -92,3 +99,59 @@ are issues that need to be handled before.
...
+$ git push --set-upstream origin v0.x-stable/fix-something
+```
+
+Even if the patch applies cleanly, you need to open a PR for these two main reasons:
+
+1. It will run the integration test, ensuring backporting the patch does not create bug.
+2. You may get reviews from another reviewer who could catch different problems.
```
mauriciovasquezbernal, line 146: This is some discussion we had but IMO it's not relevant for this document.

docs/devel/release-process.md
```diff
@@ -92,3 +99,59 @@ are issues that need to be handled before.
...
+Once your PR is merged, you can now create the bugfix tag:
+
+```bash
+$ git checkout v0.x-stable
+$ git pull
+# Where z would be y + 1:
+$ git tag v0.x.z
+```
+
+Once done, follow the instructions listed [here](/docs/devel/release-process.md#making-a-new-release), to create a new release.
```
mauriciovasquezbernal, line 157: Can't you skip the name of the document and only  using the anchor for the link?

docs/devel/release-process.md
```diff
@@ -92,3 +99,59 @@ are issues that need to be handled before.
...
+Author: author <author@mail.com>
+Date:   Tue Jan 9 16:42:38 2024 +0100
+
+    check for empty record in capabilities tracer to avoid a panic
+
+    Fixes: 39aefb92dd1df2ff73647b17707984662f8718c0 ("pkg/gadgets: Add capabilities CO-RE tracer.")
+    Signed-off-by: author <author@mail.com>
+```
+
+To test if the buggy commit belong to the stable release, you can use the following:
```
mauriciovasquezbernal, line 120:
```suggestion
To test if the buggy commit belongs to the stable release, you can use the following:
```

docs/devel/release-process.md
```diff
@@ -92,3 +99,59 @@ are issues that need to be handled before.
 If there is a security issue or an important bug fix we make patch releases in between. We don't
 have a formal definition of what an "important bug fix" is, it's up the team to discuss and decide
 whether to make a new release or not.
+
+## Making a bugfix release
+
+All newly merged patches fixing previous bug have to be backported to the latest stable release.
+When you merge a PR containing a patch fixing a bug, you should test if the buggy commits belongs to the latest release.
+Let's take an example with this commit:
```
mauriciovasquezbernal, line 107: It's not clear to me, are we going to backport all fixes (even the very small ones) or are we going to check case by case if making a bug fix release for a specific commit is needed?

mauriciovasquezbernal: I kind of agree, we should backport all commits marked with `Fixes:`, but we shouldn't mark all bug fixes to be backported. We can discuss on each PR what to do.

Issue #2358, opened 2024-01-12
`TestTraceNetwork` failing on cri-o


mauriciovasquezbernal:
In https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7502984710/attempts/1#summary-20427218246 the TestTraceNetwork failed in cri-o, the most relevant part of the log is:

```
{"runtime":{"runtimeName":"cri-o","containerId":"10476a6f12c9fc96b56fc09463e11c38e5d86d7e556fb057503541ec66801441","containerName":"k8s_nginx-pod_nginx-pod_test-trace-network-6290233096354386822_6285d88f-ed26-40f1-9b62-7d86d2b34c5c_0","containerImageName":"docker.io/library/nginx:latest","containerImageDigest":"sha256:161ef4b1bf7effb350a2a9625cb2b59f69d54ec6059a8a155a1438d0439c593c"},"k8s":{"node":"minikube-cri-o-m02","namespace":"test-trace-network-6290233096354386822","podName":"nginx-pod","containerName":"nginx-pod"},"timestamp":1705068202909809098,"type":"normal","mountnsid":4026533659,"netnsid":4026533322,"pid":17948,"tid":17948,"comm":"nginx","uid":101,"gid":101,"pktType":"HOST","proto":"TCP","port":80,"podHostIP":"192.168.49.3","podIP":"10.244.1.18","podLabels":{"run":"nginx-pod"},"dst":{"addr":"10.244.0.8","version":4,"namespace":"test-trace-network-6290233096354386822","podname":"test-pod","kind":"pod","podlabels":{"run":"test-pod"}}}
....
[...]

PR #2367, opened 2024-01-15
integration: Normalize ContainerImageDigest in all tests

mauriciovasquezbernal:
This field is not available in all cases, hence normalize it.

Related https://github.com/inspektor-gadget/inspektor-gadget/issues/2358

### DIFF
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -127,7 +127,20 @@ out:
 				continue out
 			}
 		}
-		t.Fatalf("output doesn't contain the expected entry: %+v", expectedEntry)
+
+		var str strings.Builder
+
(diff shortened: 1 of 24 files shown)


alban (approved):
This looks good to me but I wonder why it was not noticed before.

cc @claudiamarcubina

mauriciovasquezbernal:
> This looks good to me but I wonder why it was not noticed before.

I have no idea. I spent quite a lot of time trying to understand this issue and I was really surprised by that behavior. I'll merge this PR to fix the CI and then we can continue the discussion in https://github.com/inspektor-gadget/inspektor-gadget/issues/2358.

PR #2368, opened 2024-01-16
Update DetectBpfKtimeGetBootNs

mauriciovasquezbernal:
- gadgets: Rename DetectBpfKtimeGetBootNs to HasBpfKtimeGetBootNs
- gadgets: Use ebpf-go features to test bpf helper existence

### DIFF
--- a/pkg/gadgets/helpers.go
+++ b/pkg/gadgets/helpers.go
@@ -21,13 +21,12 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
-	"sync"
 	"time"
 	"unsafe"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
-	"github.com/cilium/ebpf/btf"
+	"github.com/cilium/ebpf/features"
 	"github.com/cilium/ebpf/link"
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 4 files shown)


pkg/gadgets/helpers.go
```diff
@@ -187,38 +186,12 @@ func WallTimeFromBootTime(ts uint64) types.Time {
...
-			}
-		}
-	})
-
-	return bpfKtimeGetBootNsExists
+// HasBpfKtimeGetBootNs returns true if bpf_ktime_get_boot_ns is available
+func HasBpfKtimeGetBootNs() bool {
+	// We only care about the helper, hence test with ebpf.SocketFilter that exist in all
+	// kernels that support ebpf.
+	err := features.HaveProgramHelper(ebpf.SocketFilter, asm.FnKtimeGetBootNs)
```
alban, line 193: I wonder about the performance costs of creating a ebpf program just to test the availability of a function, compared to the previous check.

alban (approved):
LGTM

pkg/gadgets/helpers.go
mauriciovasquezbernal: I really don't know, previous logic was parsing BTF and looking for a type, that's not a very cheap operation as there are many types there. This approach is loading a program with 3 instructions and not BTF related things at all, it should be pretty fast.  

In any case, this approach is also cached, so the cost is only paid once. Even if this is slower than the previous one, it shouldn't impact the whole performance of Inspektor Gadget at all.

mauriciovasquezbernal: We can do that but I don't know if that really makes any difference as the result is already cached by the library.

PR #2398, opened 2024-01-22
CRI: Enrich with initial PodLabels

burak-ok:
# CRI: Enrich with initial PodLabels

CRI gives us the option to query the pod labels. This information is in the PodSandbox, but unfortunately this only gives us the initial pod labels. So the labels with which the pod was started. Updated labels are not possible with only reading PodSandbox

Since this information is only available through CRI, I introduced a switch (currently the flag is still todo), so the ContainerRuntimeEnrichment can be used with CRI only (instead of docker or containerd API).

This has the advantage that we are saving API calls, but would mean if one wants to see K8s Pod Labels with `ig` that non K8s containers will not be captured. This is an open question.

The other option is implemented in this branch: https://github.com/inspektor-gadget/inspektor-gadget/tree/burak/pod_label_2

Use with flag `--use-cri` for `ig`:
```bash
$ ig trace exec -o columns=k8s.pod,k8s.labels,runtime.containerName,comm --use-cri
K8S.POD                                                                K8S.LABELS                           RUNTIME.CONTAINERNAME                                                 COMM            
test-pod                                                               foo=bar,abc=xyz                      test-pod                                                              date            
gadget-9tvcn                                                           controller-revision-hash=74d5bc749,… gadget                                                                gadgettracerman 
gadget-9tvcn                                                           pod-template-generation=1,k8s-app=g… gadget                                                                gadgettracerman 
test-pod                                                               foo=bar,abc=xyz                      test-pod                                                              date            
test-pod                                                               foo=bar,abc=xyz                      test-pod                                                              date            
test-pod                                                               foo=bar,abc=xyz                      test-pod                                                              date            
test-pod                                                               foo=bar,abc=xyz                      test-pod                                                              date            
```

# TODO
- [ ] Use CRI only or additionally
- [x] Introduce flag for ContainerRuntimeEnrichment, to use CRI
- [ ] Introduce flag for specifying the path to the cri docker shim socket

### DIFF
--- a/cmd/ig/utils/flags.go
+++ b/cmd/ig/utils/flags.go
@@ -57,13 +57,24 @@ type CommonFlags struct {
 	// ContainerdNamespace is the namespace used by containerd
 	ContainerdNamespace string
 
-	// UseCri specifies whether to use the CRI API to talk to the runtime
-	UseCri bool
+	// RuntimeProtocol specifies whether to use the CRI API to talk to the runtime.
+	// Useful for docker and containerd.
+	// CRI-O is always using the CRI API. Podman is always using the internal API.
+	// Supported values: internal, cri.
+	RuntimeProtocol string
 }
 
 func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 	command.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
 		// Runtimes Configuration
+		switch commonFlags.RuntimeProtocol {
+		case containerutilsTypes.RuntimeProtocolInternal:
+		case containerutilsTypes.RuntimeProtocolCRI:
+		default:
+			return commonutils.WrapInErrInvalidArg("--runtime-protocol",
+				fmt.Errorf("runtime protocol %q is not supported", commonFlags.RuntimeProtocol))
+		}
+
 		parts := strings.Split(commonFlags.Runtimes, ",")
 
 	partsLoop:
@@ -96,15 +107,12 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 			}
 
 			r := &containerutilsTypes.RuntimeConfig{
-				Name:       runtimeName,
-				SocketPath: socketPath,
+				Name:            runtimeName,
+				SocketPath:      socketPath,
+				RuntimeProtocol: commonFlags.RuntimeProtocol,
 				Extra: containerutilsTypes.ExtraConfig{
-					UseCri: commonFlags.UseCri,
+					Namespace: namespace,
 				},
-			}
-
-			if namespace != "" {
-				r.Extra.Namespace = namespace
 			}
 
 			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, r)
@@ -161,11 +169,12 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		"Namespace used by containerd",
 	)
 
-	command.PersistentFlags().BoolVar(
-		&commonFlags.UseCri,
-		"use-cri",
-		false,
-		"Use CRI API to retrieve more K8s information, but ignore non K8s containers",
+	command.PersistentFlags().StringVar(
+		&commonFlags.RuntimeProtocol,
+		"runtime-protocol",
+		containerutilsTypes.RuntimeProtocolInternal,
+		fmt.Sprintf("Container runtime protocol (docker and containerd). Supported values are: %s",
+			strings.Join(containerutils.AvailableRuntimeProtocols, ", ")),
 	)
 }
 
--- a/cmd/ig/utils/flags.go
+++ b/cmd/ig/utils/flags.go
@@ -56,6 +56,9 @@ type CommonFlags struct {
 
 	// ContainerdNamespace is the namespace used by containerd
 	ContainerdNamespace string
+
+	// UseCri specifies whether to use the CRI API to talk to the runtime
+	UseCri bool
 }
 
 func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
@@ -95,12 +98,13 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 			r := &containerutilsTypes.RuntimeConfig{
 				Name:       runtimeName,
 				SocketPath: socketPath,
+				Extra: containerutilsTypes.ExtraConfig{
+					UseCri: commonFlags.UseCri,
+				},
 			}
 
 			if namespace != "" {
-				r.Extra = &containerutilsTypes.ExtraConfig{
-					Namespace: namespace,
-				}
+				r.Extra.Namespace = namespace
 			}
 
 			commonFlags.RuntimeConfigs = append(commonFlags.RuntimeConfigs, r)
@@ -155,6 +159,13 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		"containerd-namespace",
 		constants.K8sContainerdNamespace,
 		"Namespace used by containerd",
+	)
+
+	command.PersistentFlags().BoolVar(
+		&commonFlags.UseCri,
+		"use-cri",
+		false,
+		"Use CRI API to retrieve more K8s information, but ignore non K8s containers",
 	)
 }
 
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -187,6 +187,17 @@ func WithContainerImageName(imageName string, isDockerRuntime bool) CommonDataOp
 	}
 }
 
+// WithPodLabels sets the PodLabels to facilitate the tests
+func WithPodLabels(podName string, namespace string, enable bool) CommonDataOption {
+	return func(commonData *eventtypes.CommonData) {
+		if enable {
+			commonData.K8s.PodLabels = map[string]string{
+				"run": podName,
+			}
+		}
+	}
+}
+
 func BuildCommonData(namespace string, options ...CommonDataOption) eventtypes.CommonData {
 	e := eventtypes.CommonData{
 		K8s: eventtypes.K8sMetadata{
@@ -205,6 +216,12 @@ func BuildCommonData(namespace string, options ...CommonDataOption) eventtypes.C
 	return e
 }
 
+func BuildCommonDataK8s(namespace string, options ...CommonDataOption) eventtypes.CommonData {
+	e := BuildCommonData(namespace, options...)
+	WithPodLabels("test-pod", namespace, true)(&e)
+	return e
+}
+
 func BuildBaseEvent(namespace string, options ...CommonDataOption) eventtypes.Event {
 	e := eventtypes.Event{
 		Type:       eventtypes.NORMAL,
@@ -213,6 +230,12 @@ func BuildBaseEvent(namespace string, options ...CommonDataOption) eventtypes.Ev
 	for _, option := range options {
 		option(&e.CommonData)
 	}
+	return e
+}
+
+func BuildBaseEventK8s(namespace string, options ...CommonDataOption) eventtypes.Event {
+	e := BuildBaseEvent(namespace, options...)
+	WithPodLabels("test-pod", namespace, true)(&e.CommonData)
 	return e
 }
 
--- a/integration/ig/k8s/enrichment_pod_label_test.go
+++ b/integration/ig/k8s/enrichment_pod_label_test.go
@@ -56,7 +56,7 @@ func TestEnrichmentPodLabelExistingPod(t *testing.T) {
 
 	listContainersCmd := &Command{
 		Name: "RunListContainers",
-		Cmd:  fmt.Sprintf("ig list-containers -o json --runtimes=%s --use-cri", *containerRuntime),
+		Cmd:  fmt.Sprintf("ig list-containers -o json --runtimes=%s --runtime-protocol=cri", *containerRuntime),
 		ValidateOutput: func(t *testing.T, output string) {
 			expectedContainer := &containercollection.Container{
 				K8s: containercollection.K8sMetadata{
@@ -122,7 +122,7 @@ func TestEnrichmentPodLabelNewPod(t *testing.T) {
 
 	listContainersCmd := &Command{
 		Name:         "RunWatchContainers",
-		Cmd:          fmt.Sprintf("ig list-containers -o json --runtimes=%s --use-cri --watch", *containerRuntime),
+		Cmd:          fmt.Sprintf("ig list-containers -o json --runtimes=%s --runtime-protocol=cri --watch", *containerRuntime),
 		StartAndStop: true,
 		ValidateOutput: func(t *testing.T, output string) {
 			isDockerRuntime := *containerRuntime == ContainerRuntimeDocker
--- a/integration/ig/k8s/enrichment_pod_label_test.go
+++ b/integration/ig/k8s/enrichment_pod_label_test.go
@@ -1,0 +1,206 @@
+// Copyright 2022 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	. "github.com/inspektor-gadget/inspektor-gadget/integration"
+	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+)
+
+func TestEnrichmentPodLabelExistingPod(t *testing.T) {
+	t.Parallel()
+
+	cn := "test-enrichment-pod-label-existing-pod"
+	pod := cn
+	ns := GenerateTestNamespaceName(pod)
+
+	t.Cleanup(func() {
+		commandsPostTest := []*Command{
+			DeleteTestNamespaceCommand(ns),
+		}
+		RunTestSteps(commandsPostTest, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+	})
+
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		PodCommand(cn, "busybox", ns, `["sleep", "inf"]`, ""),
+		WaitUntilPodReadyCommand(ns, pod),
+	}
+	RunTestSteps(commands, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+
+	podUID := GetPodUID(t, ns, pod)
+
+	// Containerd and docker shim name the container with the Kubernetes container
+	// name, while CRI-O use a composed name.
+	runtimeContainerName := cn
+	if *containerRuntime == ContainerRuntimeCRIO {
+		// Test container shouldn't have been restarted, so append "0".
+		runtimeContainerName = "k8s_" + cn + "_" + pod + "_" + ns + "_" + podUID + "_" + "0"
+	}
+
+	listContainersCmd := &Command{
+		Name: "RunListContainers",
+		Cmd:  fmt.Sprintf("ig list-containers -o json --runtimes=%s --use-cri", *containerRuntime),
+		ValidateOutput: func(t *testing.T, output string) {
+			expectedContainer := &containercollection.Container{
+				K8s: containercollection.K8sMetadata{
+					BasicK8sMetadata: types.BasicK8sMetadata{
+						ContainerName: cn,
+						PodName:       pod,
+						Namespace:     ns,
+						PodLabels: map[string]string{
+							"run": cn,
+						},
+					},
+					PodUID: podUID,
+				},
+				Runtime: containercollection.RuntimeMetadata{
+					BasicRuntimeMetadata: types.BasicRuntimeMetadata{
+						RuntimeName:        types.String2RuntimeName(*containerRuntime),
+						ContainerName:      runtimeContainerName,
+						ContainerImageName: "docker.io/library/busybox:latest",
+					},
+				},
+			}
+
+			// Docker can provide different values for ContainerImageName. See `getContainerImageNamefromImage`
+			isDockerRuntime := *containerRuntime == ContainerRuntimeDocker
+			if isDockerRuntime {
+				expectedContainer.Runtime.ContainerImageName = ""
+			}
+
+			normalize := func(c *containercollection.Container) {
+				c.Pid = 0
+				c.OciConfig = nil
+				c.Bundle = ""
+				c.Mntns = 0
+				c.Netns = 0
+				c.CgroupPath = ""
+				c.CgroupID = 0
+				c.CgroupV1 = ""
+				c.CgroupV2 = ""
+
+				c.SandboxId = ""
+				c.Runtime.ContainerID = ""
+				c.Runtime.ContainerImageDigest = ""
+
+				// Docker can provide different values for ContainerImageName. See `getContainerImageNamefromImage`
+				if isDockerRuntime {
+					c.Runtime.ContainerImageName = ""
+				}
+			}
+
+			ExpectEntriesInArrayToMatch(t, output, normalize, expectedContainer)
+		},
+	}
+
+	RunTestSteps([]*Command{listContainersCmd}, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+}
+
+func TestEnrichmentPodLabelNewPod(t *testing.T) {
+	t.Parallel()
+
+	cn := "test-enrichment-pod-label-new-pod"
+	pod := cn
+	ns := GenerateTestNamespaceName(pod)
+
+	listContainersCmd := &Command{
+		Name:         "RunWatchContainers",
+		Cmd:          fmt.Sprintf("ig list-containers -o json --runtimes=%s --use-cri --watch", *containerRuntime),
+		StartAndStop: true,
+		ValidateOutput: func(t *testing.T, output string) {
+			isDockerRuntime := *containerRuntime == ContainerRuntimeDocker
+			expectedEvent := &containercollection.PubSubEvent{
+				Type: containercollection.EventTypeAddContainer,
+				Container: &containercollection.Container{
+					K8s: containercollection.K8sMetadata{
+						BasicK8sMetadata: types.BasicK8sMetadata{
+							ContainerName: cn,
+							PodName:       pod,
+							Namespace:     ns,
+							PodLabels: map[string]string{
+								"run": cn,
+							},
+						},
+					},
+					Runtime: containercollection.RuntimeMetadata{
+						BasicRuntimeMetadata: types.BasicRuntimeMetadata{
+							RuntimeName:        types.String2RuntimeName(*containerRuntime),
+							ContainerName:      cn,
+							ContainerImageName: "docker.io/library/busybox:latest",
+						},
+					},
+				},
+			}
+
+			// Docker can provide different values for ContainerImageName. See `getContainerImageNamefromImage`
+			if isDockerRuntime {
+				expectedEvent.Container.Runtime.ContainerImageName = ""
+			}
+
+			normalize := func(e *containercollection.PubSubEvent) {
+				e.Container.Pid = 0
+				e.Container.OciConfig = nil
+				e.Container.Bundle = ""
+				e.Container.Mntns = 0
+				e.Container.Netns = 0
+				e.Container.CgroupPath = ""
+				e.Container.CgroupID = 0
+				e.Container.CgroupV1 = ""
+				e.Container.CgroupV2 = ""
+				e.Timestamp = ""
+
+				e.Container.SandboxId = ""
+				e.Container.K8s.PodUID = ""
+				e.Container.Runtime.ContainerID = ""
+				e.Container.Runtime.ContainerImageDigest = ""
+
+				// CRI-O uses a custom container name composed, among
+				// other things, by the pod UID. We don't know the pod UID in
+				// advance, so we can't match the expected container name.
+				// TODO: Create a test for this once we support filtering by k8s
+				// container name. See
+				// https://github.com/inspektor-gadget/inspektor-gadget/issues/1403.
+				if e.Container.Runtime.RuntimeName == ContainerRuntimeCRIO {
+					e.Container.Runtime.ContainerName = cn
+				}
+
+				// Docker can provide different values for ContainerImageName. See `getContainerImageNamefromImage`
+				if isDockerRuntime {
+					e.Container.Runtime.ContainerImageName = ""
+				}
+			}
+
+			// Watching containers is a command that needs to be started before
+			// the container is created, so we can't filter by container name
+			// neither use ExpectAllInArrayToMatch here.
+			ExpectEntriesToMatch(t, output, normalize, expectedEvent)
+		},
+	}
+
+	commands := []*Command{
+		CreateTestNamespaceCommand(ns),
+		listContainersCmd,
+		SleepForSecondsCommand(2), // wait to ensure ig has started
+		PodCommand(pod, "busybox", ns, `["sleep", "inf"]`, ""),
+		WaitUntilPodReadyCommand(ns, pod),
+		DeleteTestNamespaceCommand(ns),
+	}
+	RunTestSteps(commands, t, WithCbBeforeCleanup(PrintLogsFn(ns)))
+}
--- a/integration/ig/k8s/top_file_test.go
+++ b/integration/ig/k8s/top_file_test.go
@@ -26,9 +26,11 @@ import (
 func newTopFileCmd(ns string, cmd string, startAndStop bool) *Command {
 	validateOutputFn := func(t *testing.T, output string) {
 		isDockerRuntime := *containerRuntime == ContainerRuntimeDocker
+		isCrioRuntime := *containerRuntime == ContainerRuntimeCRIO
 		expectedEntry := &types.Stats{
 			CommonData: BuildCommonData(ns, WithRuntimeMetadata(*containerRuntime),
 				WithContainerImageName("docker.io/library/busybox:latest", isDockerRuntime),
+				WithPodLabels("test-pod", ns, isCrioRuntime),
 			),
 			// echo is built-in
 			Comm:     "sh",
--- a/integration/inspektor-gadget/trace_tcpconnect_test.go
+++ b/integration/inspektor-gadget/trace_tcpconnect_test.go
@@ -37,7 +37,7 @@ func TestTraceTcpconnect(t *testing.T) {
 		StartAndStop: true,
 		ValidateOutput: func(t *testing.T, output string) {
 			expectedEntry := &tracetcpconnectTypes.Event{
-				Event:     BuildBaseEvent(ns, WithContainerImageName("docker.io/library/nginx:latest", isDockerRuntime)),
+				Event:     BuildBaseEventK8s(ns, WithContainerImageName("docker.io/library/nginx:latest", isDockerRuntime)),
 				Comm:      "curl",
 				IPVersion: 4,
 				SrcEndpoint: eventtypes.L4Endpoint{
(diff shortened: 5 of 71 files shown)


pkg/columns/columns.go
```diff
@@ -624,6 +624,28 @@ func GetFieldAsStringExt[T any](column ColumnInternals, floatFormat byte, floatP
...
+			ff := GetFieldFunc[map[string]string, T](column)
+			return func(entry *T) string {
+				result := strings.Builder{}
+				m := ff(entry)
+				for k, v := range m {
+					result.WriteString(k)
+					result.WriteString("=")
+					result.WriteString(v)
+					result.WriteString(",")
+				}
```
alban, line 644: Ordering so that the result is deterministic:

```suggestion
				m := ff(entry)
				kvPairs := []string
				for k, v := range m {
					kvPairs = append(kvPairs, fmt.Sprintf("%s=%s", k, v)
				}
				sort.Strings(kvPairs)
				for _, kvPair := range kvPairs {
					result.WriteString(kvPair)
				}

```

pkg/container-utils/cri/cri.go
```diff
@@ -132,7 +182,36 @@ func (c *CRIClient) GetContainerDetails(containerID string) (*runtimeclient.Cont
...
+
+func (c *CRIClient) getPodSandboxFromContainerID(containerID string) (*runtime.PodSandbox, error) {
+	containerID, err := runtimeclient.ParseContainerID(c.Name, containerID)
+	if err != nil {
+		return nil, err
+	}
+
+	containers, err := listContainers(c, &runtime.ContainerFilter{
+		Id: containerID,
+	})
```
alban, line 201: When called from `WithContainerRuntimeEnrichment()` or from `GetRunningContainers()`, it seems the gRPC method `ListContainers()` will be called for each container. Is it possible to optimize that?

integration/helpers.go
```diff
@@ -195,6 +195,9 @@ func BuildCommonData(namespace string, options ...CommonDataOption) eventtypes.C
 				// Pod and Container name are defined by BusyboxPodCommand.
 				PodName:       "test-pod",
 				ContainerName: "test-pod",
+				PodLabels: map[string]string{
+					"run": "test-pod",
```
alban, line 199: Can you explain in the commitmsg which tests make use of that, or where it is checked?

pkg/container-collection/options.go
```diff
@@ -47,17 +47,38 @@ import (
 
 func enrichContainerWithContainerData(containerData *runtimeclient.ContainerData, container *Container) {
 	// Runtime
-	container.Runtime.ContainerID = containerData.Runtime.ContainerID
-	container.Runtime.RuntimeName = containerData.Runtime.RuntimeName
-	container.Runtime.ContainerName = containerData.Runtime.ContainerName
-	container.Runtime.ContainerImageName = containerData.Runtime.ContainerImageName
-	container.Runtime.ContainerImageDigest = containerData.Runtime.ContainerImageDigest
+	if container.Runtime.ContainerID == "" {
+		container.Runtime.ContainerID = containerData.Runtime.ContainerID
```
alban, line 51: Can you explain in the commitmsg why we now need to check if each field is empty?
Is enrichContainerWithContainerData() called several times, and from where?

alban:
> This has the advantage that we are saving API calls, but would mean if one wants to see K8s Pod Labels with ig that non K8s containers will not be captured. This is an open question.

The labels on the pod or sandbox and the labels at the OCI level are not related. We only need the former. And the former only makes sense on Kubernetes, so I think it is fine.

pkg/container-utils/cri/cri.go
```diff
@@ -303,7 +398,7 @@ func CRIContainerToContainerData(runtimeName types.RuntimeName, container CRICon
 				ContainerName:        strings.TrimPrefix(containerMetadata.GetName(), "/"),
 				RuntimeName:          runtimeName,
 				ContainerImageName:   image.GetImage(),
-				ContainerImageDigest: strings.Split(imageRef, "@")[1],
+				ContainerImageDigest: digestFromRef(runtimeName, imageRef),
```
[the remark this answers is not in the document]
burak-ok: The CRI client was only used for cri-o before this change. Now that `docker` and `containerd` are used again, the digest panic'ed for me.
Docker sometimes (for me) saves only `sha256:4a165184c779c0a4f2d31d6676b7790589b977c3c8fbc0577dac2544fd69cade` in its imageRef

pkg/container-utils/cri/cri.go
burak-ok: I can look into that again.
Reason I have to do this is that `GetContainerDetails` querys the `ContainerStatus`. There we have no information about the `PodSandboxId` that we need.

burak-ok:
> > This has the advantage that we are saving API calls, but would mean if one wants to see K8s Pod Labels with ig that non K8s containers will not be captured. This is an open question.
> 
> The labels on the pod or sandbox and the labels at the OCI level are not related. We only need the former. And the former only makes sense on Kubernetes, so I think it is fine.

Okay, so its okay to loose that capability when wanting K8s pod labels (`foobar` won't be shown):
```bash
$ ig trace exec -o columns=k8s.pod,k8s.labels,runtime.containerName,comm --use-cri
K8S.POD                     K8S.LABELS                           RUNTIME.CONTAINERNAME                                                 COMM            
test-pod                    foo=bar,x=y                          k8s_test-pod_test-pod_default_0f42539f-a584-482c-a45d-0e9e41686fa7_0  date            
                                                                 foobar                                                                date           
```

pkg/container-collection/options.go
burak-ok: Added commit message:

> If the container is not fully enriched, this function will be called
multiple times. So if some RuntimeEnricher do not have all informations
about a given Container, it should not overwrite the current container
informations, only update it

pkg/container-utils/runtime-client/interface.go
```diff
@@ -28,6 +28,7 @@ const (
 	PodmanDefaultSocketPath     = "/run/podman/podman.sock"
 	ContainerdDefaultSocketPath = "/run/containerd/containerd.sock"
 	DockerDefaultSocketPath     = "/run/docker.sock"
+	CriDockerDefaultSocketPath  = "/run/cri-dockerd.sock"
```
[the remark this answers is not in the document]
burak-ok: Probably yes

integration/inspektor-gadget/trace_tcpconnect_test.go
```diff
@@ -100,7 +100,7 @@ func TestTraceTcpconnect_latency(t *testing.T) {
 		StartAndStop: true,
 		ValidateOutput: func(t *testing.T, output string) {
 			expectedEntry := &tracetcpconnectTypes.Event{
-				Event:     BuildBaseEvent(ns, WithContainerImageName("docker.io/library/nginx:latest", isDockerRuntime)),
+				Event:     BuildBaseEventWithExtraK8s(ns, WithContainerImageName("docker.io/library/nginx:latest", isDockerRuntime)),
```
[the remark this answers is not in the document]
burak-ok: For `ig` we can't add the K8s labels when we have non K8s containers

pkg/container-utils/cri/cri.go
```diff
@@ -82,16 +82,61 @@ func listContainers(c *CRIClient, filter *runtime.ContainerFilter) ([]*runtime.C
 	return res.GetContainers(), nil
 }
 
+func listPodSandboxes(c *CRIClient, filter *runtime.PodSandboxFilter) ([]*runtime.PodSandbox, error) {
+	podRequest := &runtime.ListPodSandboxRequest{}
+	if filter != nil {
+		podRequest.Filter = filter
+	}
```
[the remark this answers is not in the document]
burak-ok: Now that you mentioned this. It  just doesn't make sense. Thanks

pkg/container-utils/cri/cri.go
```diff
@@ -82,16 +82,61 @@ func listContainers(c *CRIClient, filter *runtime.ContainerFilter) ([]*runtime.C
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(podSandboxes) == 0 {
+		return nil, fmt.Errorf("pod sandbox %q not found", podSandboxID)
+	}
+	if len(podSandboxes) > 1 {
+		log.Warnf("CRIClient: found multiple pod sandboxes (%d) with ID %q. Taking the first one: %+v",
+			len(podSandboxID), podSandboxID, podSandboxes)
```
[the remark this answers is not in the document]
burak-ok: Imo it should be a fatal error. That should never happen.
I looked at `func (c *CRIClient) GetContainer` and thought there was a reason behind it.
Since you also think this is weird, I'll change it also for `GetContainer`

pkg/container-collection/options.go
```diff
@@ -47,17 +47,38 @@ import (
...
-	container.K8s.PodName = containerData.K8s.PodName
-	container.K8s.PodUID = containerData.K8s.PodUID
-	container.K8s.ContainerName = containerData.K8s.ContainerName
+	if container.K8s.Namespace == "" {
+		container.K8s.Namespace = containerData.K8s.Namespace
+	}
+	if container.K8s.PodName == "" {
+		container.K8s.PodName = containerData.K8s.PodName
+	}
+	if container.K8s.PodLabels == nil {
```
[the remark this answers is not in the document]
burak-ok: I'll switch back to using `nil`. This way we can differentiate between "labels not enriched" and "labels enriched, but it has none"

pkg/gadgettracermanager/gadgettracermanager.go
```diff
@@ -174,8 +174,8 @@ func (g *GadgetTracerManager) AddContainer(_ context.Context, containerDefinitio
 				Namespace:     containerDefinition.Namespace,
 				PodName:       containerDefinition.Podname,
 				ContainerName: containerDefinition.Name,
+				PodLabels:     map[string]string{},
```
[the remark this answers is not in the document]
alban: Replying to myself:

If defined, they are the labels of the sandbox (== pod labels).

https://github.com/containerd/nri/blob/main/types/v1/types.go#L94-L95

There is no such thing as OCI-level labels: in the OCI config, we have annotations but no labels:
https://github.com/opencontainers/runtime-spec/blob/68346ed538d50d73869d48edb009fb0c2d1dd66e/schema/config-schema.json#L32

pkg/container-utils/cri/cri.go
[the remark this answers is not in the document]
alban: Issue for the panic: https://github.com/inspektor-gadget/inspektor-gadget/issues/2413

pkg/container-utils/cri/cri.go
alban: Looking at this again, I see that the gRPC method `ListContainers` will not return all the containers because it is called with a filter. I was afraid that the complexity would be `n^2` with the number of containers. But that's not the case, so I think it is ok.

alban:
Testing with Minikube and Docker. My pod has 2 labels:
```
$ kubectl get pod -n kube-system storage-provisioner -o yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    addonmanager.kubernetes.io/mode: Reconcile
    integration-test: storage-provisioner
...
```
But `ig` finds more labels:
```
$ sudo -E /opt/ig list-containers -o json --use-cri
...
    "k8s": {
      "namespace": "kube-system",
      "podName": "storage-provisioner",
      "podLabels": {
        "addonmanager.kubernetes.io/mode": "Reconcile",
        "integration-test": "storage-provisioner",
        "io.kubernetes.pod.name": "storage-provisioner",
        "io.kubernetes.pod.namespace": "kube-system",
        "io.kubernetes.pod.uid": "c6f5db9b-f991-4314-9fcf-55860bd32ec0"
      },
      "containerName": "storage-provisioner",
      "podUID": "c6f5db9b-f991-4314-9fcf-55860bd32ec0"
    },
...
```
It seems that cri-dockerd merges the labels from the Kubernetes pod and other labels:
https://github.com/Mirantis/cri-dockerd/blob/519e39ceaa7f9e00319149b9d74b243466fa3963/config/constants.go#L27-L30
```
	KubernetesPodNameLabel       = "io.kubernetes.pod.name"
	KubernetesPodNamespaceLabel  = "io.kubernetes.pod.namespace"
	KubernetesPodUIDLabel        = "io.kubernetes.pod.uid"
	KubernetesContainerNameLabel = "io.kubernetes.container.name"
```
Note that only the first three labels (name, namespace, uid) are visible. The 4th one (name) is added too but it is not visible from the CRI API because it is filtered out:
https://github.com/Mirantis/cri-dockerd/blob/master/core/helpers.go#L85C18-L85C46

I think Inspektor Gadget should not report those synthetic labels that are not actually defined on the pod resource.

integration/ig/k8s/enrichment_pod_label_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+				// Docker and CRI-O use a custom container name composed, among
+				// other things, by the pod UID. We don't know the pod UID in
+				// advance, so we can't match the expected container name.
+				// TODO: Create a test for this once we support filtering by k8s
+				// container name. See
+				// https://github.com/inspektor-gadget/inspektor-gadget/issues/1403.
+				if e.Container.Runtime.RuntimeName == ContainerRuntimeDocker ||
+					e.Container.Runtime.RuntimeName == ContainerRuntimeCRIO {
+					e.Container.Runtime.ContainerName = cn
```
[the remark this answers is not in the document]
burak-ok: For the above test we can query the pod UID and then construct the container name from that. In this test `TestEnrichmentPodLabelNewPod` we start the `test-pod` after we start `ig`. So we don't have knowledge about the pod before and don't know what pod UID it will have

integration/ig/k8s/top_file_test.go
```diff
@@ -61,6 +63,9 @@ func newTopFileCmd(ns string, cmd string, startAndStop bool) *Command {
 			if isDockerRuntime {
 				e.Runtime.ContainerImageName = ""
 			}
+			if isCrioRuntime {
+				e.K8s.PodLabels["io.kubernetes.pod.uid"] = ""
+			}
```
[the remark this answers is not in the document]
burak-ok: That function thankfully won't be needed. That was not a "real" K8s label - just something the CRI pod sandbox gave us. Will be filtered out

integration/inspektor-gadget/trace_tcpconnect_test.go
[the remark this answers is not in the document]
burak-ok: I just also thought about it. Maybe for `integration/inspektor-gadget/` we introduce its own set of `BuildBaseEvent` and `BuildCommonData`.
In the long run we might add new K8s functionality and enrichment. Primarily this would only be supported by `kubectl-gadget`, since it by-default has connection to the K8s-API server

cmd/ig/utils/flags.go
```diff
@@ -156,6 +160,13 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		constants.K8sContainerdNamespace,
 		"Namespace used by containerd",
 	)
+
+	command.PersistentFlags().BoolVar(
+		&commonFlags.UseCri,
+		"use-cri",
+		false,
+		"Use CRI API to retrieve more K8s information, but ignore non K8s containers",
```
alban, line 168: @mauriciovasquezbernal Do you have an opinion on this flag?

In the case of containerd, `ig` connects to the same socket regardless of this flag but it would use a different gRPC protocol to get the information: CRI or internal containerd protocol.

In the case of docker, it is a different socket path for cri-dockerd than the main docker socket.

So the `--use-cri` interacts with the `--runtimes` flag:
```
  -r, --runtimes string                 Container runtimes to be used separated by comma. Supported values are: docker, containerd, cri-o, podman (default "docker,containerd,cri-o,podman")
```

`cri` cannot just be a possible choice in `--runtimes` because the runtime setting is used to know which socket to connect to. Unless we add `cri-docker`, `cri-containerd`, etc. but then it is more difficult to configure for the user.

I am not sure how to make it easier for the user, so I would keep the flag as it is for now, but it does not look ideal.

pkg/container-utils/docker/docker.go
```diff
@@ -46,10 +47,16 @@ type DockerClient struct {
...
+func NewDockerClient(socketPath string, useCri bool) (runtimeclient.ContainerRuntimeClient, error) {
 	if socketPath == "" {
 		socketPath = runtimeclient.DockerDefaultSocketPath
 	}
+	if useCri {
+		// TODO: Configurable
+		socketPath = runtimeclient.CriDockerDefaultSocketPath
+		criClient, err := cri.NewCRIClient(types.RuntimeNameDocker, socketPath, DefaultTimeout)
+		return &criClient, err
+	}
```
blanquicet, line 59: I'm wondering if this should be managed outside of the docker and containerd files. IMO, it's strange to have the `New` function here, but the rest of the implementation in the CRI file. It might be confusing for the developers in the future. Maybe it should be managed in `NewContainerRuntimeClient`. Something like this:

```go
	case types.RuntimeNameDocker:
		socketPath := runtime.SocketPath

		if runtime.Extra.RuntimeProtocol == types.RuntimeProtocolCRI {
			if envsp := os.Getenv("INSPEKTOR_GADGET_DOCKER_CRI_SOCKETPATH"); envsp != "" && socketPath == "" {
				socketPath = filepath.Join(host.HostRoot, envsp)
			}
			return runtimeclient.NewDockerCRIClient(socketPath)
		}

		if envsp := os.Getenv("INSPEKTOR_GADGET_DOCKER_INTERNAL_SOCKETPATH"); envsp != "" && socketPath == "" {
			socketPath = filepath.Join(host.HostRoot, envsp)
		}
		return docker.NewDockerClient(socketPath, runtime.Extra.UseCri)
	case types.RuntimeNameContainerd:
		socketPath := runtime.SocketPath
		if envsp := os.Getenv("INSPEKTOR_GADGET_CONTAINERD_SOCKETPATH"); envsp != "" && socketPath == "" {
			socketPath = filepath.Join(host.HostRoot, envsp)
		}

		if runtime.Extra.RuntimeProtocol == types.RuntimeProtocolCRI {
			return runtimeclient.NewContainerdCRIClient(socketPath)
		}

		return containerd.NewContainerdClient(socketPath, runtime.Extra)
```

Then, in `cri,go`:

```go
func NewCRIDockerClient(socketPath string) (runtimeclient.ContainerRuntimeClient, error) {
	if socketPath == "" {
		// Docker uses different socket paths for CRI and internal Docker protocol.
		socketPath = runtimeclient.CriDockerDefaultSocketPath
	}
	return NewCRIClient(types.RuntimeNameDocker, socketPath, DefaultTimeout)
}

func NewCRIContainerdClient(socketPath string) (runtimeclient.ContainerRuntimeClient, error) {
	if socketPath == "" {
		// Containerd uses the same socket for both CRI and internal containerd protocol.
		socketPath = runtimeclient.ContainerdDefaultSocketPath
	}
	return NewCRIClient(types.RuntimeNameContainerd, socketPath, DefaultTimeout)
}
```

BTW, probably all the `os.Getenv(...)` call should be done inside the `New` functions.

pkg/container-collection/options.go
```diff
@@ -82,6 +83,27 @@ func containerRuntimeEnricher(
 		return true
 	}
 
+	// Enrich labels when we have a sandbox ID. cri-o might not find the corresponding container yet
+	if container.SandboxId != "" {
+		criClient, ok := runtimeClient.(*cri.CRIClient)
+		if ok {
+			labels, err := criClient.GetPodLabels(container.SandboxId)
+			if err != nil {
+				log.Printf("Runtime enricher (%s): failed to GetPodLabels: %s",
```
blanquicet, line 92: Debugf?

pkg/container-collection/options.go
```diff
@@ -82,6 +83,27 @@ func containerRuntimeEnricher(
...
+					runtimeName, err)
+				// Container could be managed by another runtime, don't drop it.
+				return true
+			}
+			log.Printf("Runtime enricher (%s): pod labels: %v", runtimeName, labels)
```
blanquicet, line 97: Remember to remove this one or use debug log.

pkg/container-collection/containers.go
```diff
@@ -50,6 +50,9 @@ type Container struct {
 	// See https://github.com/opencontainers/runtime-spec/blob/main/bundle.md
 	Bundle string `json:"bundle,omitempty"`
 
+	// SandboxId is the sandbox id for the corresponding pod
+	SandboxId string `json:"sandboxId,omitempty"`
```
blanquicet, line 54: For now, this information will be only available for containers that are created when IG is already running and only for containerd and crio, right? Please update the comment. BTW, isn't this info provided by the runtime?

pkg/container-collection/options.go
```diff
@@ -82,6 +83,27 @@ func containerRuntimeEnricher(
...
+			container.K8s.PodLabels = labels
+
+			// If this was the only information needed, skip rest of this enricher
+			if runtimeclient.IsEnrichedWithK8sMetadata(container.K8s.BasicK8sMetadata) {
+				return true
+			}
+		}
+	}
```
blanquicet, line 105: Actually, I don't understand why is all this necessary if the pod labels will be available in the `ContainerData` struct returned by the `GetContainer` call below. In fact, it is breaking all the idea of using an abstract `runtimeClient` that doesn't know/care about the actual runtime or protocol used.

pkg/container-utils/cri/cri.go
```diff
@@ -288,12 +393,28 @@ type CRIContainer interface {
...
 
-func CRIContainerToContainerData(runtimeName types.RuntimeName, container CRIContainer) *runtimeclient.ContainerData {
+func digestFromRef(imageRef string) string {
+	splitted := strings.Split(imageRef, "@")
+	if len(splitted) == 1 {
+		return imageRef
+	} else {
+		return splitted[1]
+	}
+}
```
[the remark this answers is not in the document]
alban: Fix in PR https://github.com/inspektor-gadget/inspektor-gadget/pull/2455

alban:
I rebased on main and fixed the conflicts.

---

I see the CI has the following failures:

#### On `Test ig w k8s (containerd)`:

=== NAME  TestEnrichmentPodLabelNewPod

actual event:
```
{
   "container" : {
      "k8s" : {
         "containerName" : "test-enrichment-pod-label-new-pod",
         "namespace" : "test-enrichment-pod-label-new-pod-2280675184725679420",
         "podLabels" : {
            "run" : "test-enrichment-pod-label-new-pod"
         },
         "podName" : "test-enrichment-pod-label-new-pod"
      },
      "runtime" : {
         "containerImageName" : "docker.io/library/busybox:latest",
         "runtimeName" : "containerd"
      }
   },
   "event" : "CREATED"
}
```
expected event:
```
{
   "container" : {
      "k8s" : {
         "containerName" : "test-enrichment-pod-label-new-pod",
         "namespace" : "test-enrichment-pod-label-new-pod-2280675184725679420",
         "podLabels" : {
            "run" : "test-enrichment-pod-label-new-pod"
         },
         "podName" : "test-enrichment-pod-label-new-pod"
      },
      "runtime" : {
         "containerImageName" : "docker.io/library/busybox:latest",
         "containerName" : "test-enrichment-pod-label-new-pod",
         "runtimeName" : "containerd"
      }
   },
   "event" : "CREATED"
}
```
I don't know why `container.runtime.containerName` is missing.

#### On `Test ig w k8s (cri-o)`:

* FAIL: TestTraceTcpconnect_latency: we need to add the `podLabels` in the expected event
* FAIL: TestTraceNetwork: we need to add the `podLabels` in the expected event
* FAIL: TestTraceOOMKill: `podLabels` is expected but it is not found

blanquicet:
> I rebased on main and fixed the conflicts.
> 
> I see the CI has the following failures:
> #### On `Test ig w k8s (containerd)`:
> 
> === NAME TestEnrichmentPodLabelNewPod
> 
> [...]
> I don't know why `container.runtime.containerName` is missing.

The container name is not there because of this:

https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L100-L103

I don't know why it's returning only if the k8s metadata is enriched. It still needs to enrich the runtime metadata in the below code.

On the other side, I still don't understand the need of [this block](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L86-L105). Is it because CRI-O might get nothing in [this GetContainer call](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L107)? If so, why it should get [the PodSandbox](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L90) but nothing in the `GetContainer()`? Do we have a deadlock with CRI-O that I'm not aware of?

cmd/ig/utils/flags.go
blanquicet: pinging @mauriciovasquezbernal

blanquicet:
After several tests, things seem more clear. 

For new CRI-O containers:

1) The container-hook provides the PID, ContainerID and ContainerName when notifying the new CRI-O container. _NOTE: This could extended to provide the runtime name so that we avoid trying all the runtimes._
2) The OCIConfig enricher enriches the container with RuntimeName, ContainerImageName, SandboxId and all the k8s information except the PodLabers: Namespace, PodName, ContainerName and PodUID.
3) When the ContainerRuntimeEnricher is called, the container doesn't exist yet for CRI-O. So, [this GetContainer call](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L107) will fail. However, the PodSandbox indeed exist. That's why using the PodSanboxID, [this GetPodLabels call](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L90) will succeed. 

Said that, I got the answers to my questions:

> On the other side, I still don't understand the need of [this block](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L86-L105). Is it because CRI-O might get nothing in [this GetContainer call](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L107)? If so, why it should get [the PodSandbox](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L90) but nothing in the `GetContainer()`? Do we have a deadlock with CRI-O that I'm not aware of?

As said in (3), [this block](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L86-L105) is indeed useful. However, I suggest applying it only for CRI-O as for other runtimes [the GetContainer call](https://github.com/inspektor-gadget/inspektor-gadget/blob/692306499f2bfdfae49bb865a77e738d0c6423b7/pkg/container-collection/options.go#L107) will succeed and it will enrich the container with the PodLabels and other missing fields, e.g., ContainerImageDigest:

```golang
	// For new CRI-O containers, the next GetContainer call will always fail
	// because the container doesn't exist yet. So, Let's enrich the container
	// at least with the PodLabels as the PodSandbox do exist at this point.
	if container.SandboxId != "" && (container.Runtime.RuntimeName == "" || container.Runtime.RuntimeName == types.RuntimeNameCrio) {
```

> I don't know why it's returning only if the k8s metadata is enriched. It still needs to enrich the runtime metadata in the below code.
> 

This block needs to be indeed removed or modified. At most, it can return for CRI-O (we know the next GetContainer will fail) and let it go through for other runtimes.

pkg/container-collection/containers.go
blanquicet: > For now, this information will be only available for containers that are created when IG is already running and only for containerd and crio, right?
>
Correct.

> Please update the comment.
> 
We need to document who enriches what somewhere else, not here.

> BTW, isn't this info provided by the runtime?
Yes, it is. However, implementing this is not worth it because we only use 'SandboxID' to retrieve the PodLabels for new CRI-O containers from the runtime enricher. Why only in this case? [The `runtimeClient.GetContainer` call](https://github.com/inspektor-gadget/inspektor-gadget/blob/f02743d64d81a01f3212bc0a9e9387e67a3e9ec7/pkg/container-collection/options.go#L76) will fail because the container doesn't exist yet in CRI-O. However, the SandboxPods do exist. So, we can at least enrich the container with the pod labels.

cmd/ig/utils/flags.go
```diff
@@ -156,6 +160,13 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 		constants.K8sContainerdNamespace,
 		"Namespace used by containerd",
 	)
+
+	command.PersistentFlags().BoolVar(
```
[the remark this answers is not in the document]
blanquicet: Because these flags are for the `list-container` command.

pkg/container-collection/options.go
blanquicet: See comment https://github.com/inspektor-gadget/inspektor-gadget/pull/2398#issuecomment-1945366447.

pkg/gadgettracermanager/gadgettracermanager.go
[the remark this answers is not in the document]
blanquicet: Was this solved in https://github.com/inspektor-gadget/inspektor-gadget/pull/2418?

blanquicet:
Hi @alban, I fixed the CI and managed the comments. The only open discussions are:
- [The flag](https://github.com/inspektor-gadget/inspektor-gadget/pull/2398#discussion_r1471469049). This should be addressed in this PR.
- The dependency between the OCI Config and Runtime enrichers, see below. IMO, we could improve this in another PR.
- For the initial containers, the number of gRPC calls toward the runtime increased from 2 to 5 for each container when using the CRI. It means for CRI-O or for any runtime if `--use-cri` is specified. IMO, we could improve this in another PR.

I summarized in this table how SandboxID and PodLabels are now enriched. Notice that the SandboxID is used only internally in IG to compute the PodLabels in a special case, while the PodLabels field is now being added to the gadgets events within the k8s struct:

<table>
    <thead>
        <tr>
            <th rowspan=2></th>
            <th colspan=4>Initial Containers</th>
            <th colspan=4>New Containers</th>
        </tr>
        <tr>
            <th>CRI-O</th>
            <th>Containerd</th>
            <th>Docker</th>
            <th>Podman</th>
            <th>CRI-O</th>
            <th>Containerd</th>
            <th>Docker</th>
            <th>Podman</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>SandboxID</td>
            <td>X</td>
            <td>X</td>
            <td>X</td>
            <td>N/A</td>
            <td>OCI Config Annotations</td>
            <td>OCI Config Annotations</td>
            <td>X</td>
            <td>N/A</td>
        </tr>
        <tr>
            <td>k8s.PodLabels</td>
            <td>CRI-Runtime GetContainers() </td>
            <td>X</td>
            <td>X</td>
            <td>N/A</td>
            <td>CRI-Runtime GetPodLabels(k8s.SandboxID)* </td>
            <td>X</td>
            <td>X</td>
            <td>N/A</td>
        </tr>
        <tr>
            <td>k8s.PodLabels (--use-cri)</td>
            <td>CRI-Runtime GetContainers() </td>
            <td>CRI-Runtime GetContainers() </td>
            <td>CRI-Runtime GetContainers() </td>
            <td>N/A</td>
            <td>CRI-Runtime GetPodLabels(k8s.SandboxID)* </td>
            <td>CRI-Runtime GetContainer() </td>
            <td>CRI-Runtime GetContainer() </td>
            <td>N/A</td>
        </tr>
    </tbody>
</table>

`*`: New CRI-O containers are a special case for two reasons:

1. The ContainerRuntime enricher requires that the container be already enriched with the `SandboxID`, which is enriched by the OCI Config enricher. Currently, based on [the order the enrichers are added](https://github.com/inspektor-gadget/inspektor-gadget/blob/3802a66adc876b1c65848550320ac830105a5f97/pkg/ig-manager/ig-manager.go#L95-L103) to the container collection, it's known that the OCI Config enricher will be always called before the ContainerRuntime enricher. However, a better mechanism should be found to avoid such a dependency.
2. We can't use the normal `GetContainer` call because, by the time the enricher is called, the container doesn't exist yet in CRI-O. So, we implemented a [method](https://github.com/inspektor-gadget/inspektor-gadget/blob/3802a66adc876b1c65848550320ac830105a5f97/pkg/container-collection/options.go#L86-L105) to enrich the pod labels as the Sandbox already exists.

pkg/gadgettracermanager/gadgettracermanager.go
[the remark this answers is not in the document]
alban: > Was this solved in https://github.com/inspektor-gadget/inspektor-gadget/pull/2418?

Yes, I think it is handled correctly now.

alban:
Thanks @blanquicet !
I rebased on main again and I will try to implement `--runtime-protocol=internal|cri`.

alban:
```
# ./ig trace exec --runtime-protocol=cri -o columns=runtime.containerName,pid,comm,ret,k8s.labels
RUNTIME.CONTAINERNAME PID    COMM RET K8S.LABELS
shell8                766439 cat  0   run=shell8
```

cmd/ig/utils/flags.go
```diff
@@ -57,13 +57,22 @@ type CommonFlags struct {
 	// ContainerdNamespace is the namespace used by containerd
 	ContainerdNamespace string
 
-	// UseCri specifies whether to use the CRI API to talk to the runtime
-	UseCri bool
+	// RuntimeProtocol specifies whether to use the CRI API to talk to the runtime
+	// Supported values: internal, cri.
```
blanquicet, line 61: Should we specify here that it makes sense only for docker and containerd? For CRI-O and Podman there is no choice. CRI-O will only use CRI and Podman always its internal protocol.

cmd/ig/utils/flags.go
```diff
@@ -161,11 +170,12 @@ func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
...
-	command.PersistentFlags().BoolVar(
-		&commonFlags.UseCri,
-		"use-cri",
-		false,
-		"Use CRI API to retrieve more K8s information, but ignore non K8s containers",
+	command.PersistentFlags().StringVar(
+		&commonFlags.RuntimeProtocol,
+		"runtime-protocol",
+		"internal",
+		fmt.Sprintf("Container runtime protocol. Supported values are: %s",
```
blanquicet, line 177: dito

pkg/container-utils/types/types.go
```diff
@@ -16,9 +16,14 @@ package types
...
+)
+
 type ExtraConfig struct {
-	Namespace string
-	UseCri    bool
+	Namespace       string
+	RuntimeProtocol string
 }
 
 type RuntimeConfig struct {
```
blanquicet, line 29: IMO, it shouldn't go in `ExtraConfig` but directly in the `RuntimeConfig` and be mandatory (it will be automatically populated in most paths by the default flag value). The `ExtraConfig` struct should remain a pointer with the optional parameters as it was before. WDYT?

blanquicet (commented):
The change to `--runtime-protocol` LGTM.

pkg/container-utils/types/types.go
alban: I updated the last commit with that change.
Note that `ExtraConfig` was not a pointer.

pkg/container-utils/containerd/containerd.go
```diff
@@ -47,13 +48,26 @@ type ContainerdClient struct {
 	ctx    context.Context
 }
 
-func NewContainerdClient(socketPath string, config *containerutilsTypes.ExtraConfig) (runtimeclient.ContainerRuntimeClient, error) {
+func NewContainerdClient(socketPath string, config *containerutilsTypes.RuntimeConfig) (runtimeclient.ContainerRuntimeClient, error) {
```
blanquicet, line 51: nit: I think we should avoid passing the whole config because it contains also a socket path, and I would like to avoid creating confusions in the future. I suggest:

```suggestion
func NewContainerdClient(socketPath string, protocol string, config *containerutilsTypes.ExtraConfig) (runtimeclient.ContainerRuntimeClient, error) {
```

pkg/container-utils/types/types.go
blanquicet: Mmmm it was, but Burak modified it in a commit of this PR: 

https://github.com/inspektor-gadget/inspektor-gadget/blob/1072df20edbf07eed8269dd896494e4c1bc0afd8/pkg/container-utils/types/types.go#L23-L27

blanquicet (approved):
IMO, we are ready to go. I didn't test the new flag locally but the tests we have for it are covering most cases, so I trust them.

If you agree, I can create the issues to track the discussions we didn't resolve in this PR: https://github.com/inspektor-gadget/inspektor-gadget/pull/2398#issuecomment-1950260439

pkg/container-utils/containerd/containerd.go
alban: Updated

alban:
> IMO, we are ready to go. I didn't test the new flag locally but the tests we have for it are covering most cases, so I trust them.

Thanks. I updated the PR, and I will merge it once the CI passes.

> If you agree, I can create the issues to track the discussions we didn't resolve in this PR: https://github.com/inspektor-gadget/inspektor-gadget/pull/2398#issuecomment-1950260439

Yes, please.

blanquicet:
Issues to track missing tasks:

- https://github.com/inspektor-gadget/inspektor-gadget/issues/2523
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2524
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2525

PR #2400, opened 2024-01-22
trace exec upper_layer: fix BTF type size relocation

alban:
bpf_core_type_size() has to be used explicitly to ensure the correct BTF relocation will be used.

Using sizeof or other implicit methods don't work.

Fixes: 8c433c4b2939 ("image-based trace exec: add upper_layer field")
Fixes: a1d98b153cf1 ("built-in trace exec: add upper_layer field")

Fixes: #2323

## How to use

Check the upper layer field in the trace exec gadget on different kernels

## Testing done

Deploy this DaemonSet on AKS with 2 node pools:
- Mariner Linux
- Ubuntu

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: upperlayer
  labels:
    k8s-app: upperlayer
spec:
  selector:
    matchLabels:
      name: upperlayer
  template:
    metadata:
      labels:
        name: upperlayer
        role: demo
    spec:
      containers:
      - name: upperlayer
        image: ubuntu
        command: [ "sh", "-c", "cp /usr/bin/uname / ; while sleep 1 ; do /host/usr/bin/uname -a ; uname -a ; /uname -a ; done" ]
        volumeMounts:
        - name: host
          mountPath: /host
      volumes:
      - name: host
        hostPath:
          path: /
```

```
sudo -E ./ig image build -t myrepo.azurecr.io/trace-exec:debug ./gadgets/trace_exec
sudo -E ./ig image push myrepo.azurecr.io/trace-exec:debug
./kubectl-gadget run myrepo.azurecr.io/trace-exec:debug -o columns=k8s.node,comm,ret,args,upper_layer

K8S.NODE                            COMM   ARGS                 UPPER_LAYER
aks-userpool-39817355-vmss000000    uname  /host/usr/bin/uname  false
aks-userpool-39817355-vmss000000    uname  /usr/bin/uname       false
aks-userpool-39817355-vmss000000    uname  /uname               true
aks-userpool-39817355-vmss000000    sleep  /usr/bin/sleep       false
aks-azurepool-26141556-vmss000000   uname  /host/usr/bin/uname  false
aks-azurepool-26141556-vmss000000   uname  /usr/bin/uname       false
aks-azurepool-26141556-vmss000000   uname  /uname               true
aks-azurepool-26141556-vmss000000   sleep  /usr/bin/sleep       false
```


gadgets/trace_exec/program.bpf.c
```diff
@@ -148,17 +148,6 @@ int ig_execve_e(struct trace_event_raw_sys_enter *ctx)
...
-// struct ovl_inode defined in fs/overlayfs/ovl_entry.h
-// Unfortunately, not exported to vmlinux.h
-// and not available in /sys/kernel/btf/vmlinux
-// See https://github.com/cilium/ebpf/pull/1300
-// So we don't benefit from CO-RE's preserve_access_index
-// We only rely on vfs_inode and __upperdentry relative positions
-struct _ovl_inode {
-	struct inode vfs_inode;
-	struct dentry *__upperdentry;
-};
```
mauriciovasquezbernal, line 160: Perhaps it can be fixed by adding `__attribute__((preserve_access_index));` to this structure? (I think it'll force clang to emit a relocation when the container_of macro needs to calculate `offsetof(type,member)`)

mauriciovasquezbernal: Also, why is this structure so different from the upstream one? https://github.com/torvalds/linux/blob/v5.0/fs/overlayfs/ovl_entry.h#L91-L105

mauriciovasquezbernal: I think I now understand, we have the inode, and we suppose the dentry is the next member on the structure, so that's why we don't care about any other members.

gadgets/trace_exec/program.bpf.c
```diff
@@ -172,16 +161,17 @@ static __always_inline bool upper_layer()
...
-	if (!oi) {
-		return false;
-	}
-
 	struct dentry *upperdentry;
 
+	// struct ovl_inode defined in fs/overlayfs/ovl_entry.h
+	// Unfortunately, not exported to vmlinux.h
+	// and not available in /sys/kernel/btf/vmlinux
+	// See https://github.com/cilium/ebpf/pull/1300
```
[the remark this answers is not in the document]
alban: That would be good. But even if the cilium/ebpf PR is merged, we still have work to do to make use of it in Inspektor Gadget:

* How an image-based gadget can specify which kernel module to load, so that the `run` command uses the correct `KernelTypes`:
```
ebpf.CollectionOptions{
    MapReplacements: mapReplacements,
    Programs: ebpf.ProgramOptions{
        KernelTypes: /* BTF from the kernel and the modules */,
    },
}
```

* How does it interact with BTFGen? #2045

gadgets/trace_exec/program.bpf.c
alban: > Perhaps it can be fixed by adding __attribute__((preserve_access_index)); to this structure? (I think it'll force clang to emit a relocation when the container_of macro needs to calculate offsetof(type,member))

cilium/ebpf emits a bad relocation when I add this attribute because it does not find `struct _ovl_inode` in the kernel BTF.
```
const badRelo = 0xbad2310 
```

> I think I now understand, we have the inode, and we suppose the dentry is the next member on the structure, so that's why we don't care about any other members.

Correct. In that way, no need to define types for other fields.

alban:
* Renamed function to `has_upper_layer()`
* Rebased
* Pushed to `citest/alban_upperlayer_fix` too

alban:
The CI is green, on `citest/alban_upperlayer_fix` too.

gadgets/trace_exec/program.bpf.c
mauriciovasquezbernal: > cilium/ebpf emits a bad relocation when I add this attribute because it does not find struct _ovl_inode in the kernel BTF.

That's unfortunate.

mauriciovasquezbernal (approved):
LGTM. It's still "hacky" but the final solution requires a lot of work as mentioned by Alban.

gadgets/trace_exec/program.bpf.c
```diff
@@ -172,16 +161,17 @@ static __always_inline bool upper_layer()
...
+	// We only rely on vfs_inode and __upperdentry relative positions
 	bpf_probe_read_kernel(&upperdentry, sizeof upperdentry,
-			      &oi->__upperdentry);
+			      ((void *)inode) +
+				      bpf_core_type_size(struct inode));
+
```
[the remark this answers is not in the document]
alban: > We should definitely take inspiration from strong behavior and not weak assumption like this one.
> Can you open a follow-up PR to document this limitation?

Ok, I will.

> Also, is this field shown by default? It should be hidden by default.

It's hidden by default.

> I will review and test this PR, can you please do so?

I'll try as time allows.

> This is not the goal of Inspektor Gadget to load modules.
> But we can indeed, use the YAML to read which modules BTF under /sys/kerne/btf the gadget needs.

Yes, I meant loading the BTF file of a kernel module.

gadgets/trace_exec/program.bpf.c
lmb: You can use "flavour names" like `struct ovl_inode___gadget`. Note the three `_`. Anything after the underscore is ignored by the library when trying to find a target.

alban: @lmb Thanks for the link, I didn't know about struct flavors.

The following seems to work. Tested without https://github.com/cilium/ebpf/pull/1300, so `struct ovl_inode` is unknown to BTF and only the "else" branch is tested.

Is it the approach you recommend?

```
#include <vmlinux.h>

struct ovl_inode___gadget {
       struct inode vfs_inode;
       struct dentry *__upperdentry;
} __attribute__((preserve_access_index));
...

	if (bpf_core_type_exists(struct ovl_inode___gadget)) {
		struct ovl_inode___gadget *oi = container_of(
			inode, struct ovl_inode___gadget, vfs_inode);
		bpf_probe_read_kernel(&upperdentry, sizeof upperdentry,
				      &oi->__upperdentry);
	} else {
		bpf_probe_read_kernel(&upperdentry, sizeof upperdentry,
				      ((void *)inode) +
					      bpf_core_type_size(struct inode));
	}
```

lmb: I think something like checking for the existence of the field would be more reliable: https://github.com/libbpf/libbpf/blob/f81eef23b33c0dbf923e863a72ce51ea4d32e291/src/bpf_core_read.h#L151-L164

alban: I gave it a try: https://github.com/inspektor-gadget/inspektor-gadget/pull/2406

About `bpf_core_field_exists`: I'll give it a try. If the struct does not exist, I guess `bpf_core_field_exists` would just return false?

PR #2401, opened 2024-01-23
pkg/oci: Do not trim repository prefix if default.

eiffel-fl:
Otherwise, it impedes image removing by only untagging them.

Fixes: bd5578b86632 ("oci: Use default registry for Inspektor Gadget")

### DIFF
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -288,7 +288,6 @@ func listGadgetImages(ctx context.Context, store *oci.Store) ([]*GadgetImageDesc
 			if named, ok := parsed.(reference.Named); ok {
 				repository = named.Name()
 			}
-			repository = strings.TrimPrefix(repository, defaultDomain+"/"+officialRepoPrefix)
 
 			tag := "latest"
 			if tagged, ok := parsed.(reference.Tagged); ok {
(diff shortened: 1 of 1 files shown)


mqasimsarfraz:
> Otherwise, it impedes image removing by only untagging them.

 It was our intent to have [similar UX as docker](https://github.com/inspektor-gadget/inspektor-gadget/pull/2307#pullrequestreview-1817955510). Can't we just append the prefix to image remove if needed or the issue is coming from oras-go?

mauriciovasquezbernal:
> > Otherwise, it impedes image removing by only untagging them.
> 
> It was our intent to have [similar UX as docker](https://github.com/inspektor-gadget/inspektor-gadget/pull/2307#pullrequestreview-1817955510). Can't we just append the prefix to image remove if needed or the issue is coming from oras-go?

I also see these two options:

- Make `listGadgetImages` return the image with the full url, and trim the prefix in `ListGadgetImages`
- Add a `trimDefaultPrefix` parameter to `listGadgetImages`, it'll be true when called from `ListGadgetImages`, but false when called from `DeleteGadgetImage`

mauriciovasquezbernal:
Actually my proposal was to make this on [`ListGadgetImages`](https://github.com/inspektor-gadget/inspektor-gadget/blob/63d5794efd6f49f4917f1a1f40ca361247f48ed2/pkg/oci/oci.go#L318). I don't think we should expose the default registry details outside the pkg/oci package.

mauriciovasquezbernal:
That's why I proposed the second option:

> Add a trimDefaultPrefix parameter to listGadgetImages, it'll be true when called from ListGadgetImages, but false when called from DeleteGadgetImage

mauriciovasquezbernal:
As I mentioned before, I think we shouldn't expose the default prefix outside of this module. It'll fix this immediate issue, but then all users of this will have to be aware of the default domain, so I prefer to keep the logic only on this module for now.

mauriciovasquezbernal (approved):
LGTM. Thanks!

PR #2405, opened 2024-01-24
Fix CodeQL jobs in the CI

eiffel-fl:
Hi.

This PR fixes various problems in the current usage of CodeQL as pointed by:
https://github.com/github/codeql/issues/15405#issuecomment-1905608507

Best regards.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -311,9 +311,6 @@ jobs:
         # - latest commit of btfhub
         # - hash of all bpf objects
         key: btfgen-${{ matrix.platform }}-${{ steps.get-btfhub-head.outputs.head }}-${{ hashFiles('pkg/**/*.o') }}
-    - uses: github/codeql-action/init@v3
-      with:
-        languages: go
     - name: Build ig-linux-${{ matrix.platform }}
       run: |
         if [ "${{ matrix.platform }}" = 'arm64' ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/inspektor-gadget.yml
```diff
@@ -331,16 +331,64 @@ jobs:
           -czf ig-linux-${{ matrix.platform }}.tar.gz -C ${{ matrix.platform }} \
           ig LICENSE
         rm -rf ${{ matrix.platform }}
-    - name: CWE checks for ig-linux-${{ matrix.platform }}
-      uses: github/codeql-action/analyze@v3
-      with:
-        category: ig-linux-${{ matrix.platform }}
```
mauriciovasquezbernal: Should also `github/codeql-action/init` be removed from L314?

.github/workflows/inspektor-gadget.yml
```diff
@@ -331,16 +331,64 @@ jobs:
...
       uses: actions/upload-artifact@v3
       with:
         name: ig-linux-${{ matrix.platform }}-tar-gz
         path: /home/runner/work/inspektor-gadget/inspektor-gadget/ig-linux-${{ matrix.platform }}.tar.gz
 
+  build-ig-cwe:
+    name: Check ig CWE
+    # level: 1
+    needs:
+      - btfgen
```
mauriciovasquezbernal, line 344: I think we can omit btfgen altogether on this new job. It only generates some BTF files that are embed in the binary and it shouldn't produce any CWE alert.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1410,6 +1461,7 @@ jobs:
       - test-unit
       - build-clients
       - build-ig
+      - build-ig-cwe
       - build-gadget-container-images
       - build-helper-images
       - build-and-push-gadgets
```
mauriciovasquezbernal, line 1467: I'll propose to add `build-ig-cwe` only as a requirement for release. Otherwise it'll be very annoying to fix a CWE issue to later on find that other things as broken as well.

mauriciovasquezbernal (approved):
LGTM. Thanks for fixing it.

PR #2406, opened 2024-01-24
load btf from kernel modules

alban:
# load btf from kernel modules

With https://github.com/cilium/ebpf/pull/1300, cilium/ebpf is able to load the btf from kernel modules.

See https://github.com/inspektor-gadget/inspektor-gadget/pull/2400#discussion_r1465028972

## Testing done

```
$ sudo -E ./ig image build -t myregistry/trace-exec:debug ./gadgets/trace_exec
$ go run -exec 'sudo -E' ./cmd/ig/... run myregistry/trace-exec:debug -o columns=comm,ret,args,upper_layer
```
In another shell:
```
# cat /sys/kernel/tracing/trace_pipe
```
I tried both cases, when it succeeds or fails to load the overlay btf:
```
           <...>-155263  [001] ...21 122321.370091: bpf_trace_printk: ovl_inode exists, using BTF automatically
           <...>-155443  [002] ...21 122366.925450: bpf_trace_printk: ovl_inode does not exist, using bpf_core_type_size
```

### DIFF
--- a/gadgets/trace_exec/program.bpf.c
+++ b/gadgets/trace_exec/program.bpf.c
@@ -148,6 +148,11 @@ int ig_execve_e(struct trace_event_raw_sys_enter *ctx)
 	return 0;
 }
 
+struct ovl_inode___gadget {
+	struct inode vfs_inode;
+	struct dentry *__upperdentry;
+} __attribute__((preserve_access_index));
+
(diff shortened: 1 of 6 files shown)


alban:
The Cilium patch has been merged, so we can revisit this PR.

PR #2416, opened 2024-01-26
docs/examples: Improve documentation about privileged

mauriciovasquezbernal:


blanquicet (approved):
The best solution would be to keep them all updated but I'm aware it is difficult to do.

PR #2417, opened 2024-01-26
Add Volume Mount for /var/lib/ig/oci-store in Gadget DaemonSet

### LINKED ISSUE #2203: [RFE] Allow using volume for pulled gadgets in gadget pod
## Current situation

Currently if you run the gadget it is pulled and stored to `/var/lib/ig/oci-store` for each pod. I see following problems with it:

- If the gadget pod is restarted all the pulled gadgets are deleted.
- The stored data can be too much over time (shouldn't be a concern because of above)

## Impact

Not able to persist gadgets on restart or upgrades.

## Ideal future situation

Charts and `kubectl deploy` should provide a way to mount oci-store in gadget.

## Implementation options

Use the [volumes](https://kubernetes.io/docs/concepts/storage/volumes/) for `/var/lib/ig` in gadgets daemom-set.

moki1202:
This PR fixes #2203

### DIFF
--- a/charts/gadget/templates/daemonset.yaml
+++ b/charts/gadget/templates/daemonset.yaml
@@ -203,6 +203,10 @@ spec:
               name: cgroup
             - mountPath: /sys/fs/bpf
               name: bpffs
+            {{- if .Values.config.mountOCIStore }}
+            - mountPath: /var/lib/ig/oci-store
+              name: oci-store
+            {{- end }}
             {{- if .Values.config.mountPullSecret }}
             - mountPath: /var/run/secrets/gadget/pull-secret
               name: pull-secret
@@ -257,6 +261,11 @@ spec:
         - name: debugfs
           hostPath:
             path: /sys/kernel/debug
+        {{- if .Values.config.mountOCIStore }}
+        - name: oci-store
+          local:
+            path: /var/lib/ig/oci-store
+        {{- end }}
         {{- if .Values.config.mountPullSecret }}
         - name: pull-secret
           secret:
--- a/charts/gadget/values.yaml
+++ b/charts/gadget/values.yaml
@@ -25,6 +25,9 @@ config:
   # -- Daemon Log Level. Valid values are: "trace", "debug", "info", "warning", "error", "fatal", "panic"
   daemonLogLevel: "info"
 
+  # -- Whether  to mount oci-store volume in gadget.
+  mountOCIStore: false
+
   # -- Mount pull secret (gadget-pull-secret) to pull image-based gadgets from private registry
   mountPullSecret: false
 
@@ -58,3 +61,7 @@ skipLabels: true
 # -- Labels to be added to all other resources.
 additionalLabels:
   labels: {}
+
+persistentVolume:
+  # -- Storage capacity for the Persistent Volume
+  storageCapacity: 16Mi


alban:
Hi,

Could we make this optional and opt-out by default?
```
{{- if .Values.config.xxx
```

moki1202:
@alban 

> Hi,
> 
> Could we make this optional and opt-out by default?
> 
> ```
> {{- if .Values.config.xxx
> ```

@alban changes added.

charts/gadget/values.yaml
```diff
@@ -25,6 +25,9 @@ config:
   # -- Daemon Log Level. Valid values are: "trace", "debug", "info", "warning", "error", "fatal", "panic"
   daemonLogLevel: "info"
 
+  # -- Enable OCI Store to mount oci-store volume in gadget.
+  enableOCIStore: false
```
mqasimsarfraz, line 29: How about `mountOCIStore` to keep it consistent with `mountPullSecret` ?

moki1202: Resolved.

mqasimsarfraz: Great, thanks! I will trigger the CI run.

charts/gadget/templates/daemonset.yaml
```diff
@@ -204,6 +208,9 @@ spec:
         - name: debugfs
           hostPath:
             path: /sys/kernel/debug
+        - name: oci-store
```
mqasimsarfraz, line 211: We need `{{- if .Values.config.mountOCIStore }` here as well? CI seems to be [failing](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7695905877/job/20969885766#step:7:312) because of this.

moki1202: Resolved.

moki1202: Other than failing checks, is providing conditional statement necessary in the volume block?

mqasimsarfraz: > is providing conditional statement necessary in the volume block?

I mean it can be confusing to have it in volume block and not mount it in gadget container.

mqasimsarfraz (approved):
LGTM!

charts/gadget/templates/daemonset.yaml
```diff
@@ -210,6 +214,11 @@ spec:
         - name: debugfs
           hostPath:
             path: /sys/kernel/debug
+        {{- if .Values.config.mountOCIStore }}
+        - name: oci-store
+          hostPath:
```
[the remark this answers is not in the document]
moki1202: I'll update the PR soon.

charts/gadget/values.yaml
```diff
@@ -58,3 +61,7 @@ skipLabels: true
 # -- Labels to be added to all other resources.
 additionalLabels:
   labels: {}
+
+persistentVolume:
+  # -- Storage capacity for the Persistent Volume
+  storageCapacity: ""
```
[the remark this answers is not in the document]
moki1202: @eiffel-fl I won't be leaving this empty. I just wanted to confirm with you its default value.

mauriciovasquezbernal (commented):
How is using a hostPath persistent volume different than mounting a hostPath volume in the pod as proposed in the first version of this PR?

moki1202:
> > How is using a hostPath persistent volume different than mounting a hostPath volume in the pod as proposed in the first version of this PR?
> 
> Good catch! We should really avoid using [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) when not necessary.

What's the reason for avoiding hostpath?

moki1202:
> We should really avoid using [hostPath](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) when not necessary.

I'll change this to a local volume.

moki1202:
> Hi!
> 
> Thank you for switching to `local`! Can you please run `make generate-manifests` to propagate your changes to `pkg/resources/manifests/deploy.yaml`?
> 
> Best regards.

Already pushed with the last amend :)

charts/gadget/templates/ocistorepv.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
moki1202: @eiffel-fl I can't figure out why we'd use this since this persistent volume is anyway only mounted after checking this condition in daemonset.yaml.

moki1202:
@eiffel-fl I don't understand, do you mean using an empty directory volume here rather than defining a PV?

charts/gadget/templates/ocistorepv.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PersistentVolume
+metadata:
+  name: oci-store-pv
+spec:
+  capacity:
+    storage: {{- .Values.persistentVolume.storageCapacity }}  
+  accessModes:
+    - ReadWriteOnce  
+  local:
+    path: /var/lib/ig/oci-store
```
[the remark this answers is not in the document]
moki1202: 😅 I finally understood what you meant by using local directly. Apologies! Yes I think we can use it directly instead of hostPath. I'll push the changes.

moki1202:
> Thinking more about it, I am wondering if we should not also add a flag `kubectl gadget deploy`. This way, we have the same set of options in both the CLI and the helm chart.

what would this flag do exactly, can you ellaborate?

moki1202:
> > > Thinking more about it, I am wondering if we should not also add a flag `kubectl gadget deploy`. This way, we have the same set of options in both the CLI and the helm chart.
> > 
> > 
> > what would this flag do exactly, can you ellaborate?
> 
> Sure! The flag would basically use a volume mount instead of what we have currently. You can take a look at [3ddce30](https://github.com/inspektor-gadget/inspektor-gadget/commit/3ddce30451861f2c37e27532ca98a39e3c74011c), to see how we implemented the option in both the HELM charts and the CLI. It would need more code, as instead of setting an environment variable you would need to add the volume as well as do the wiring to actually use this volume, but it can give you an idea.

Thank you for the explanation, I can definitely give this a try.

PR #2418, opened 2024-01-26
WithKubernetesEnrichment: Enrich if PodLabels is nil

burak-ok:
## WithKubernetesEnrichment: Enrich if PodLabels is nil

ContainerSelector (Label enrichment) doesn't always work. There is a bug for some combinations of hookmode and container runtime.
This PR tries to solve this issue with checking if the `PodLabels` were set at all

I currently don't know why the tests for `cri-o` fail

## Issue

The problem is within the https://github.com/inspektor-gadget/inspektor-gadget/blob/54fe1e7d47d42a7c39cba5356c7d8da2b371a809/pkg/container-collection/options.go#L443 function

In its future container enricher it first checks if the `PodName` is already enriched. If it is, it skips itself and the PodLabels are never getting enriched. The `PodName` can be enriched in some cases by other previous enrichers, so the `KubernetesEnrichment` doesn't do anything
https://github.com/inspektor-gadget/inspektor-gadget/blob/54fe1e7d47d42a7c39cba5356c7d8da2b371a809/pkg/container-collection/options.go#L459-L461

## Reproduction
To test that it generally works in some cases:
```
$ make minikube-start-docker
$ make minikube-deploy

# In another shell start a pod:
$ kubectl run --rm -ti --image busybox test-pod --labels='foo=bar,abc=xyz'
$ while true; do date && sleep 3; done

# Back to the original shell:
$ ./kubectl-gadget trace exec -l "foo=bar"
K8S.NODE                      K8S.NAMESPACE                 K8S.POD                       K8S.CONTAINER                PID        PPID       COMM            RET ARGS                                 
minikube-docker               default                       test-pod                      test-pod                     1006139    1005893    date            0   /bin/date                            
```

But it doesn't work here:
```
$ minikube start -p minikube-cri-o --driver kvm --container-runtime=cri-o
$ ./kubectl-gadget deploy --image ghcr.io/inspektor-gadget/inspektor-gadget:v0.24.0 --image-pull-policy=Always --hook-mode crio

# In another shell start a pod:
$ kubectl run --rm -ti --image busybox test-pod --labels='foo=bar,abc=xyz'
$ while true; do date && sleep 3; done

# Back to the original shell:
$ ./kubectl-gadget trace exec -l "foo=bar"
K8S.NODE                      K8S.NAMESPACE                 K8S.POD                       K8S.CONTAINER                PID        PPID       COMM            RET ARGS                                 
# No events at all....

### CODE REVIEW

**Comment by alban** [comment]
Should we add a boolean in the `ContainerDefinition` struct to say whether the labels are defined? So we can distinguish a pod with known empty labels and a pod with unknown labels?

https://github.com/inspektor-gadget/inspektor-gadget/blob/6b316065bf04ac978a32ef43a99905792b27c37c/pkg/gadgettracermanager/api/gadgettracermanager.proto#L70

Then, the NRI and OCI hooks can set this boolean appropriately and you would not need to make the error-prone test `if len(containerDefinition.Labels) > 0`.

---

Some notes:

When using cri-o, the [entrypoint function](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadget-container/entrypoint/entrypoint.go#L102-L105) uses hooks from `/opt/hooks/oci`. The cri-o hook is configured in [gadget-container/hooks/crio/gadget-prestart.json](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadget-container/hooks/crio/gadget-prestart.json) and it will run [gadget-container/hooks/oci/prestart.sh](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadget-container/hooks/oci/prestart.sh), which will run [ocihookgadget](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadget-container/hooks/oci/main.go#L148-L151). In the end, it will call the AddContainer gRPC method:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadget-container/hooks/oci/main.go#L148-L151
```
	_, err = client.AddContainer(ctx, &pb.ContainerDefinition{
		Id:        ociStateID,
		Pid:       uint32(ociStatePid),
		OciConfig: string(bundleConfig),
	})
```
As you can see, the `Labels` are not set here.

This is different to the NRI hook, where the AddContainer methods define the labels:
https://github.com/inspektor-gadget/inspektor-gadget/blob/69e5e43b70df960ef6a3b0872efdabb4f99f3b90/gadget-container/hooks/nri/main.go#L139
```
	_, err = client.AddContainer(ctx, &pb.ContainerDefinition{
		Id:        r.ID,
		Labels:    labels,
		Namespace: namespace,
		Podname:   podName,
		Name:      containerName,
		Pid:       uint32(r.Pid),
	})
```

I think the labels here are [the labels of the sandbox](https://github.com/containerd/nri/blob/main/types/v1/types.go#L94-L95), so I guess they are the labels of the Kubernetes pod.

**Comment by burak-ok** [comment]
> Should we add a boolean in the `ContainerDefinition` struct to say whether the labels are defined? So we can distinguish a pod with known empty labels and a pod with unknown labels?

I am always in favor of getting rid of error prone stuff and make everything clearer. I'll add it

> I think the labels here are [the labels of the sandbox](https://github.com/containerd/nri/blob/main/types/v1/types.go#L94-L95), so I guess they are the labels of the Kubernetes pod.

Kind of, yes. The Kubernetes labels are there as well. But there are additional labels (Probably the same additional labels which I get through CRI, since this is also only checking for sandbox labels)

File: gadget-container/gadgettracermanager/main.go [hunk 1]
```diff
@@ -182,6 +182,7 @@ func main() {
 			Podname:   podname,
 			Name:      containername,
 			Labels:    labels,
+			LabelsSet: label != "",
```
alban, line 185: Ack.

For the record, this code path is only used for manual tests:
```
/bin/gadgettracermanager -call add-container -containerid xxx -containerpid xxx -namespace xxx ...
```

pkg/gadgettracermanager/api/gadgettracermanager.proto
```diff
@@ -68,6 +68,7 @@ message ContainerDefinition {
   string podname = 5;
   string name = 6;
   repeated Label labels = 7;
+  bool labels_set = 8;
```
alban, line 71: Maybe add a comment what it is for and that some hooks know the labels, some don't.

alban (approved):
LGTM

pkg/container-collection/options.go
```diff
@@ -456,50 +486,29 @@ func WithKubernetesEnrichment(nodeName string, kubeconfig *rest.Config) Containe
 
 		// Future containers
 		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
-			if container.K8s.PodName != "" {
+			if container.K8s.PodName != "" && container.K8s.PodLabels != nil {
```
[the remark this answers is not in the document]
burak-ok: > What about testing all the fields?

Right. I changed it to use `runtimeclient.IsEnrichedWithK8sMetadata(container.K8s.BasicK8sMetadata)`

pkg/container-collection/options.go
```diff
@@ -456,50 +486,32 @@ func WithKubernetesEnrichment(nodeName string, kubeconfig *rest.Config) Containe
 
 		// Future containers
 		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
-			if container.K8s.PodName != "" {
+			// Skip enriching if all k8s fields are already known.
+			// This is an optimization and to make sure to avoid erasing the fields in case of error.
+			if runtimeclient.IsEnrichedWithK8sMetadata(container.K8s.BasicK8sMetadata) &&
+				container.K8s.PodLabels != nil {
```
alban, line 492: IsEnrichedWithK8sMetadata() already checks PodLabels

burak-ok: In this PR, this is not the case.
Only in the other PR at this commit: https://github.com/inspektor-gadget/inspektor-gadget/pull/2398/commits/1b3feb537f7781133edd2ebb8b90dcd5df207fee#diff-8e49337a1417a39a62ebf427e84fb2faa566c96e37ae627ac484c05dbe65c483R149

PR #2423, opened 2024-01-29
Allow using env variables to set runtime configuration for `ig`

### LINKED ISSUE #1044: [RFE] Allow using env variables to set runtime configuration for `local-gadget`
## Description

Currently, it is necessary to set the runtimes to be used and their socket paths at each `local-gadget` call:

```bash
$ local-gadget list-containers --containerd-socketpath=/bar/foo --runtimes=containerd
$ local-gadget snapshot process --containerd-socketpath=/bar/foo --runtimes=containerd
$ local-gadget trace exec --containerd-socketpath=/bar/foo --runtimes=containerd
```

## Impact

It is cumbersome for users to have to specify it every time.

## Ideal future situation

`local-gadget` provides the possibility of using env variables to set them:

```bash
$ export RUNTIMES=containerd
$ export CONTAINERD_SOCKETPATH=/bar/foo
$ local-gadget list-containers
$ local-gadget snapshot process
$ local-gadget trace exec
```

It seems it's possible by using viper + cobra https://github.com/kubereboot/kured/pull/464/files.

pawarpranav83:
# Allows using env vars to set runtime config

Fixes #1044

Instead of specifying flag values every time, users can set env vars and viper would bind these vars to their appropriate flag values.

Note: When both env var and flag value are set, flag value takes precedence.

## How to use

Notation for Env variables - should attach the `IG` prefix with the flag name. And if the flag contains `-` use `_` instead.

Example - --containerd-socketpath : IG_CONTAINERD_SOCKETPATH

## Testing done

```
$ ig list-containers --containerd-socketpath=/bar/foo --runtimes=containerd
$ ig snapshot process --containerd-socketpath=/bar/foo --runtimes=containerd
$ ig trace exec --containerd-socketpath=/bar/foo --runtimes=containerd
```
and 
```
$ export RUNTIMES=containerd
$ export CONTAINERD_SOCKETPATH=/bar/foo
$ ig list-containers
$ ig snapshot process
$ ig trace exec
```
Should give similar results.

### DIFF
--- a/cmd/ig/main.go
+++ b/cmd/ig/main.go
@@ -17,9 +17,12 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
 
 	// Import this early to set the enrivonment variable before any other package is imported
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/environment/local"
@@ -43,14 +46,17 @@ import (
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/socketenricher"
 )
 
+const EnvPrefix = "IG"
+
 func main() {
 	if experimental.Enabled() {
 		log.Info("Experimental features enabled")
 	}
 
 	rootCmd := &cobra.Command{
-		Use:   "ig",
-		Short: "Collection of gadgets for containers",
+		Use:               "ig",
+		Short:             "Collection of gadgets for containers",
+		PersistentPreRunE: bindViper,
 	}
 	common.AddVerboseFlag(rootCmd)
 
@@ -61,15 +67,6 @@ func main() {
 		common.NewVersionCmd(),
 	)
 
-	// evaluate flags early; this will make sure that flags for host are evaluated before
-	// calling host.Init()
-	err := commonutils.ParseEarlyFlags(rootCmd, os.Args[1:])
-	if err != nil {
-		// Analogous to cobra error message
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-		os.Exit(1)
-	}
-
 	runtime := local.New()
 	hiddenColumnTags := []string{"kubernetes"}
 	common.AddCommandsFromRegistry(rootCmd, runtime, hiddenColumnTags)
@@ -79,7 +76,38 @@ func main() {
 	rootCmd.AddCommand(common.NewLoginCmd())
 	rootCmd.AddCommand(common.NewLogoutCmd())
 
+	// evaluate flags early; this will make sure that flags for host are evaluated before
+	// calling host.Init()
+	err := commonutils.ParseEarlyFlags(rootCmd, os.Args[1:])
+	if err != nil {
+		// Analogous to cobra error message
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(1)
 	}
+}
+
+// bindViper initializes viper and binds command flags with environment variables
+func bindViper(command *cobra.Command, args []string) error {
+	v := viper.New()
+
+	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	v.SetEnvPrefix(EnvPrefix)
+	v.AutomaticEnv()
+
+	command.Flags().VisitAll(func(f *pflag.Flag) {
+		// Apply the viper config value to the flag when the flag is not set and viper has a value
+		if !f.Changed && v.IsSet(f.Name) {
+			val := v.Get(f.Name)
+			log.Debugf("Binding %s command flag to environment variable: %s", f.Name, fmt.Sprintf("%v", val))
+			if err := command.Flags().Set(f.Name, fmt.Sprintf("%v", val)); err != nil {
+				log.Errorf("Error occurred while binding flags to env vars: %s", err)
+			}
+		}
+	})
+
+	return nil
 }
--- a/cmd/ig/utils/flags.go
+++ b/cmd/ig/utils/flags.go
@@ -59,7 +59,7 @@ type CommonFlags struct {
 }
 
 func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
-	command.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+	command.PreRunE = func(cmd *cobra.Command, args []string) error {
 		// Runtimes Configuration
 		parts := strings.Split(commonFlags.Runtimes, ",")
 
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/shopspring/decimal v1.3.1
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.0
+	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.18.2
 	github.com/stretchr/testify v1.8.4
 	github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635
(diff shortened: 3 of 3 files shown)


pawarpranav83:
cc @blanquicet

cmd/ig/main.go
```diff
@@ -79,7 +76,38 @@ func main() {
 	rootCmd.AddCommand(common.NewLoginCmd())
 	rootCmd.AddCommand(common.NewLogoutCmd())
 
+	// evaluate flags early; this will make sure that flags for host are evaluated before
+	// calling host.Init()
+	err := commonutils.ParseEarlyFlags(rootCmd, os.Args[1:])
```
pawarpranav83, line 81: `host.Init()` is called in `RunE` which would be executed in `rootCmd.Execute`, therefore, calling this function after adding all sub-commands so that flags are parsed (solves the above issue) but before `rootCmd.Execute` which satisfies the comment condition.

blanquicet: Is this necessary for the use of env variables or it is just something you want to add to make the code more readable/organised?

cmd/ig/main.go
```diff
@@ -100,6 +100,7 @@ func bindViper(command *cobra.Command, args []string) error {
 	v.AutomaticEnv()
 
 	command.Flags().VisitAll(func(f *pflag.Flag) {
+		fmt.Println(f.Name, f.Changed)
```
blanquicet, line 103: I suppose it was added for debugging purposes. Please clean it up.

cmd/ig/main.go
```diff
@@ -83,3 +89,24 @@ func main() {
...
+	v := viper.New()
+
+	replacer := strings.NewReplacer("-", "_")
+	v.SetEnvKeyReplacer(replacer)
+	v.SetEnvPrefix(EnvPrefix)
+	v.AutomaticEnv()
+
+	command.Flags().VisitAll(func(f *pflag.Flag) {
+		// Apply the viper config value to the flag when the flag is not set and viper has a value
+		if !f.Changed && v.IsSet(f.Name) {
```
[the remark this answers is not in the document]
blanquicet: @mqasimsarfraz added support for env variables for the [kubectl-aks](https://github.com/Azure/kubectl-aks) project, maybe he has some suggestions here.

cmd/ig/main.go
```diff
@@ -83,3 +89,24 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+// bindViper initializes viper and binds command flags with environment variables
+func bindViper(command *cobra.Command, args []string) error {
+	v := viper.New()
+
+	replacer := strings.NewReplacer("-", "_")
+	v.SetEnvKeyReplacer(replacer)
```
blanquicet, line 98: nit:

```suggestion
	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
```

cmd/ig/utils/flags.go
```diff
@@ -60,6 +60,8 @@ type CommonFlags struct {
 
 func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
 	command.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		command.Parent().PersistentPreRunE(command.Parent(), args)
+
```
blanquicet: Why do we need to do this?

cmd/ig/utils/flags.go
```diff
@@ -59,9 +59,7 @@ type CommonFlags struct {
 }
 
 func AddCommonFlags(command *cobra.Command, commonFlags *CommonFlags) {
-	command.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
-		command.Parent().PersistentPreRunE(command.Parent(), args)
-
+	command.PreRunE = func(cmd *cobra.Command, args []string) error {
```
blanquicet: This commit is changing something that was added in the previous commit. Sorry, but I'm not sure I got what you are trying to do.

blanquicet (commented):
Thanks @pawarpranav83 for your contribution! I still need to test it, but I already have some comments 🙂.

I think a @mqasimsarfraz review would be valuable here.

cmd/ig/main.go
pawarpranav83: It solves the issue of `f.Changed` being false even if we set the flag values of sub-commands. 
Therefore, parsing flags after adding all sub-commands.

cmd/ig/utils/flags.go
pawarpranav83: Here, I set it to `PreRunE`, so the `bindViper` which is set as `PersistentPreRunE` can execute.
When the parent command and child command both have `PersistentPreRunE`, the child command's function would be called instead, so `bindViper` was not being called with `list-containers` command.

cmd/ig/utils/flags.go
pawarpranav83: > When the parent command and child command both have PersistentPreRunE, the child command's function would be called instead, so bindViper was not being called with list-containers command.

We could call the parent's `PersistentPreRunE` explicitly when sub-command is overriding it, but this wasn't showing results for env var set value, so instead using `PreRunE`.

cmd/ig/main.go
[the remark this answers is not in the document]
pawarpranav83: This was resolved by parsing flags after adding all sub-commands.

cmd/ig/main.go
```diff
@@ -79,7 +76,36 @@ func main() {
...
+	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	v.SetEnvPrefix(EnvPrefix)
+	v.AutomaticEnv()
+
+	command.Flags().VisitAll(func(f *pflag.Flag) {
+		// Apply the viper config value to the flag when the flag is not set and viper has a value
+		if !f.Changed && v.IsSet(f.Name) {
+			val := v.Get(f.Name)
+			log.Infof("Binding %s command flag to environment variable: %s", f.Name, fmt.Sprintf("%v", val))
+			command.Flags().Set(f.Name, fmt.Sprintf("%v", val))
```
mqasimsarfraz, line 106: Should we handle the error here?

cmd/ig/main.go
```diff
@@ -79,7 +76,36 @@ func main() {
...
+
+// bindViper initializes viper and binds command flags with environment variables
+func bindViper(command *cobra.Command, args []string) error {
+	v := viper.New()
+
+	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	v.SetEnvPrefix(EnvPrefix)
+	v.AutomaticEnv()
+
+	command.Flags().VisitAll(func(f *pflag.Flag) {
```
mqasimsarfraz, line 101: Just for my understanding, the approach here is to bind only the environment variables (not flags) so we can `v.Get()` them and override the flag value if it wasn't set by flag?

mqasimsarfraz (commented):
It did an initial pass, will continue to test it!

cmd/ig/main.go
pawarpranav83: Yes.

mqasimsarfraz: Makes sense, thanks. Can you please rebase the PR so I can trigger a CI run?

pawarpranav83: Done.

alban:
Thanks! Will this work also when #2448 is merged?

pawarpranav83:
Are we going to keep socketPath flag? Like if the user manually sets the flag value, would this value override the auto-detect value?
In either case, it should be fine, since we have not hardcoded which flags viper should bind.

mqasimsarfraz:
> Thanks! Will this work also when #2448 is merged?

IIUC this change only impacts `ig` which doesn't overlap with any changes made in #2448 (`WithInitialKubernetesContainers` and `WithhPodInformer` aren't used by `ig`). Am I missing something?

mqasimsarfraz:
Also, just to align expectations it seems we are using `INSPEKTOR_GADGET_` env prefix for gadget pods (`IG_EXPERIMENTAL` being an exception) and here we will use `IG_` env prefix for `ig`. I assume it should be okay. @blanquicet any thoughts?

pawarpranav83:
@mqasimsarfraz Do we need to add support for configuration files as well?

Issue #2426, opened 2024-01-30
ig "--auto-mount-filesystems" logic invalid on Amazon Linux


alban:
## Description

On EKS with Linux `5.10.205-195.804.amzn2.x86_64`, running `ig` in a pod errors out with:
```
error: filesystems tracefs not mounted (did you try --auto-mount-filesystems?)
```
Even when `/sys/kernel` is bind mounted from the host.

While using `--auto-mount-filesystems` solves the issue, it is an unnecessarily strict check because `ig` checks for tracefs on `/sys/kernel/tracing` but on Amazon Linux, it is not mounted there. It is only mounted **on-demand** on `/sys/kernel/debug/tracing` (notice the **debug** directory).

## Impact

* Confusing check and error message
* `--auto-mount-filesystems` is a workaround but not the correct solution

## Environment and steps to reproduce

* `kubectl apply` this pod:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: ig
  labels:
    k8s-app: example-ig
spec:
  containers:
    - name: ig
      image: ghcr.io/inspektor-gadget/ig:latest
      securityContext:
        privileged: true
      command:
        - "ig"
        - "trace"
        - "exec"
        - "--host"
      env:
        - name: HOST_ROOT
          value: "/host"
      volumeMounts:
        - mountPath: /host
          name: host
        - mountPath: /run
          name: run
        - mountPath: /sys/kernel
          name: debugfs
  nodeName: CHANGEME.compute.internal
  volumes:
    - name: host
      hostPath:
        path: /
    - name: run
      hostPath:
        path: /run
    - name: debugfs
      hostPath:
        path: /sys/kernel
```

* Check logs with `kubectl logs ig`
* Notice the error

```
error: filesystems tracefs not mounted (did you try --auto-mount-filesystems?)
```

## Expected behavior

* ig should check both `/sys/kernel/tracing` and `/sys/kernel/debug/tracing`

According to my tests, `unix.Statfs()` on the tracing directory is enough to trigger the automount.

## Additional information

Using kubetl-debug-node and chroot to access the host:

```
$ kubectl debug node/$NODE -ti --image=ubuntu -- chroot /host /bin/bash
# uname -r
5.10.205-195.804.amzn2.x86_64
# cat /proc/self/mountinfo |grep ' /sys/kernel'
[...]

PR #2427, opened 2024-01-30
auto-mount-filesystems: add alternative path for tracefs

alban:
Amazon Linux does not use the new path "/sys/kernel/tracing" but only the old path "/sys/kernel/debug/tracing".

If tracefs is mounted at one of those paths, it is good enough for Inspektor Gadget. No need to error out or suggest anything.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/2426

## Testing done

- [x] This is pushed to a branch prefixed with `citest/`, so [more tests](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7711809052) are executed
- [x] Test on Amazon Linux

### DIFF
--- a/pkg/utils/host/workarounds.go
+++ b/pkg/utils/host/workarounds.go
@@ -218,39 +218,44 @@ func autoSdUnitRestart() (exit bool, err error) {
 func autoMountFilesystems(dryRun bool) (mountsSuggested []string, err error) {
 	fs := []struct {
 		name    string
-		path    string
+		paths   []string
 		magic   int64
 		suggest bool // suggest mounting this filesystem
 	}{
 		{
 			"bpf",
-			"/sys/fs/bpf",
+			[]string{"/sys/fs/bpf"},
 			unix.BPF_FS_MAGIC,
 			false, // do not make 'ig --auto-mount-filesystems=false' fail if bpffs is not mounted
 		},
 		{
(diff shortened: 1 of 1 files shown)


alban:
Tested on Amazon Linux 5.10.205-195.804.amzn2.x86_64 with:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: ig
  labels:
    k8s-app: example-ig
spec:
  containers:
    - name: ig
      image: $MYREPO/ig:tracefs
      #image: ghcr.io/inspektor-gadget/ig:latest
      securityContext:
        privileged: true
      command:
        - "ig"
        - "trace"
        - "exec"
        - "--host"
      env:
        - name: HOST_ROOT
          value: "/host"
      volumeMounts:
        - mountPath: /host
          name: host
        - mountPath: /run
          name: run
        - mountPath: /sys/kernel
          name: debugfs
  nodeName: $NODE
  volumes:
    - name: host
      hostPath:
        path: /
    - name: run
      hostPath:
        path: /run
    - name: debugfs
      hostPath:
        path: /sys/kernel
```

I changed $NODE to the correct node, and the image to both the upstream image and the new image built from this PR.
It fails with the upstream image and succeeds with the new image.

mauriciovasquezbernal (approved):
LGTM from code inspection

PR #2442, opened 2024-02-02
charts: Set container filesystem as read only.

eiffel-fl:


mauriciovasquezbernal (commented):
How does it play with https://github.com/inspektor-gadget/inspektor-gadget/pull/2417?

alban (approved):
LGTM from code inspection.

But please rebase and then check if the CI still green.

PR #2446, opened 2024-02-02
add version skew warning message

### LINKED ISSUE #1795: Version skew warning message is not printed

moki1202:
Fixes #1795 

add version skew warning message when kubectl-gadget and gadget pod versions are different when deploying a gadget.


moki1202:
@eiffel-fl apologies! I forgot to change this PR to a draft after a couple of suggestions/changes were mentioned on the related issue. I'll be making some significant changes and pushing it here today.

moki1202:
@eiffel-fl @mauriciovasquezbernal Instead of introducing a version var in gadgettracermanager I've used the PrintVersionSkewWarning function directly since we've now introduced a version package (client version) and to avoid repeating code blocks (image tag logic to get server gadget version).

cmd/kubectl-gadget/version.go
```diff
@@ -19,28 +19,26 @@ import (
...
-	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/utils"
+	"github.com/inspektor-gadget/inspektor-gadget/internal/version"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
 	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
 )
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
```
mauriciovasquezbernal, line 34: nit: remove this empty line.

internal/version/version.go
```diff
@@ -0,0 +1,23 @@
+package version
```
mauriciovasquezbernal, line 1: nit: add license header.

internal/version/version.go
```diff
@@ -0,0 +1,23 @@
...
+	var err error
+	parsedVersion, err = semver.ParseTolerant(version)
+	if err != nil {
+		panic(fmt.Sprintf("parsing version: %s", err))
+	}
+}
+
+func Version() semver.Version {
+	return parsedVersion
+}
```
mauriciovasquezbernal, line 23: nit: add line break.

mauriciovasquezbernal (changes requested):
> I've used the PrintVersionSkewWarning function directly since we've now introduced a version package (client version) and to avoid repeating code blocks (image tag logic to get server gadget version).

I think after introducing the internal version package it's better to go one step ahead and use this instead of relying on the image tag. Otherwise this solution won't work for ig as a daemon: https://www.inspektor-gadget.io/docs/v0.24.0/ig/#using-ig-as-a-daemon

Just to be clear, `printVersionSkewWarning` [uses](https://github.com/inspektor-gadget/inspektor-gadget/blob/01e52936e8c1d35f99b992a11b3205a9434229e0/cmd/kubectl-gadget/utils/trace.go#L196) the image tag, also this is part of trace.go logic that is somehow legacy, so I'd suggest to use the new approach in grpc runtime without taking it into account at all.

moki1202:
> > I've used the PrintVersionSkewWarning function directly since we've now introduced a version package (client version) and to avoid repeating code blocks (image tag logic to get server gadget version).
> 
> I think after introducing the internal version package it's better to go one step ahead and use this instead of relying on the image tag. Otherwise this solution won't work for ig as a daemon: https://www.inspektor-gadget.io/docs/v0.24.0/ig/#using-ig-as-a-daemon
> 
> Just to be clear, `printVersionSkewWarning` [uses](https://github.com/inspektor-gadget/inspektor-gadget/blob/01e52936e8c1d35f99b992a11b3205a9434229e0/cmd/kubectl-gadget/utils/trace.go#L196) the image tag, also this is part of trace.go logic that is somehow legacy, so I'd suggest to use the new approach in grpc runtime without taking it into account at all.

@mauriciovasquezbernal I thought that version.Version() here returns client-side version. I cannot think of a way to populate server-version if not for the image tag method. Sorry for all the unnecessary confusion from my side.

mauriciovasquezbernal:
No worries. `version.Version()` returns the version of the binary, I suppose for make it work on the gadgettracermanager you'll have to use a VERSION argument in the [Dockerfile](https://github.com/inspektor-gadget/inspektor-gadget/blob/5bba230d35f144758fcab911e8d2417e0f4a0308/Dockerfiles/gadget.Dockerfile) as mentioned before.

moki1202:
@mauriciovasquezbernal got it. Now that we get the binary version from version.Version(), how do I set the client version here (especially for gadgetctl).

internal/deployinfo/info.go
```diff
@@ -27,6 +27,7 @@ import (
 type DeployInfo struct {
 	Catalog      *runtime.Catalog
 	Experimental bool
+	GadgetTag    string
```
mauriciovasquezbernal, line 30: Nit: It's not the tag anymore. It's the `ServerVersion`.

pkg/gadget-service/api/api.pb.go
```diff
@@ -388,6 +388,7 @@ type InfoResponse struct {
 	Version      string `protobuf:"bytes,1,opt,name=version,proto3" json:"version,omitempty"`
 	Catalog      []byte `protobuf:"bytes,2,opt,name=catalog,proto3" json:"catalog,omitempty"`
 	Experimental bool   `protobuf:"varint,3,opt,name=experimental,proto3" json:"experimental,omitempty"`
+	ServerVersion string
```
mauriciovasquezbernal, line 391: This can't be edited by hand. In order to add a new field, edit the api.proto file around https://github.com/inspektor-gadget/inspektor-gadget/blob/50dd7158e1ba32066471996ff9f117dfb4fcb79b/pkg/gadget-service/api/api.proto#L74-L78
and then run `make generated-files` in the `pkg/gadget-service` folder.

Dockerfiles/gadget.Dockerfile
```diff
@@ -20,7 +22,8 @@ RUN cd /gadget && go mod download
 # This COPY is limited by .dockerignore
 COPY ./ /gadget
 RUN cd /gadget/gadget-container && \
-	make -j$(nproc) TARGET_ARCH=${TARGETARCH} gadget-container-deps
+	make -j$(nproc) TARGET_ARCH=${TARGETARCH} gadget-container-deps \
+	-ldflags"-X github.com/inspektor-gadget/inspektor-gadget/internal/version.version=${VERSION}"
```
mauriciovasquezbernal, line 26: -ldflags has to be passed to the `go build` command, not to the makefile. Something like 

```diff
$ git diff gadget-container/
diff --git gadget-container/Makefile gadget-container/Makefile
index b9a1a65e1..14763b8bd 100644
--- gadget-container/Makefile
+++ gadget-container/Makefile
@@ -3,6 +3,7 @@ gadget-container-deps: entrypoint cleanup ocihookgadget gadgettracermanager nrig
 
 
 TARGET_ARCH ?= $(shell go env GOHOSTARCH)
+VERSION ?= $(shell git describe --tags --always --dirty)
 
 # Entrypoint
 
@@ -26,6 +27,7 @@ cleanup:
 gadgettracermanager:
        mkdir -p bin
        GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=$(TARGET_ARCH) go build \
+               -ldflags "-X github.com/inspektor-gadget/inspektor-gadget/internal/version.version=$(VERSION)" \
                -o bin/gadgettracermanager \
                ./gadgettracermanager/
``` 

could work. (after removing this line)

mauriciovasquezbernal (commented):
> how do I set the client version here (especially for gadgetctl), or are we just checking it with the kubectl-gadget version?

Once you fix the comments I'm making in this review, the server and client versions could be accessed like:

```diff
diff --git cmd/kubectl-gadget/main.go cmd/kubectl-gadget/main.go
index 50bcea4a0..f8a8f4374 100644
--- cmd/kubectl-gadget/main.go
+++ cmd/kubectl-gadget/main.go
@@ -22,6 +22,8 @@ import (
        "github.com/spf13/cobra"
 
        // Import this early to set the enrivonment variable before any other package is imported
+       "github.com/inspektor-gadget/inspektor-gadget/internal/deployinfo"
+       "github.com/inspektor-gadget/inspektor-gadget/internal/version"
        _ "github.com/inspektor-gadget/inspektor-gadget/pkg/environment/k8s"
        paramsPkg "github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 
@@ -126,6 +128,11 @@ func main() {
 
        if !skipInfo {
                grpcRuntime.InitDeployInfo()
+               info, err := deployinfo.Load()
+               if err == nil {
+                       fmt.Printf("server version: %s\n", info.GadgetTag)
+                       fmt.Printf("client version: %s\n", version.Version())
+               }
        }
 
        namespace, _ := utils.GetNamespace()
```

(The same logic should work for gadgetctl in https://github.com/inspektor-gadget/inspektor-gadget/blob/c49480a4b7361e93f76a310b8c7ea1e95176eba2/cmd/gadgetctl/main.go#L78)

Dockerfiles/gadget.Dockerfile
moki1202: @mauriciovasquezbernal `VERSION ?= $(shell git describe --tags --always --dirty)` does this set version to the commit hash? Why would we do that?

mauriciovasquezbernal: `git describe` provides a human readable description of the state of the repo. See `git help describe`. We already use that in the main Makefile: 

https://github.com/inspektor-gadget/inspektor-gadget/blob/e38a90935da28cd56abca71a1fc01eb4e779e272/Makefile#L1

moki1202:
@mauriciovasquezbernal all lint errors fixed too.

cmd/kubectl-gadget/utils/trace.go
```diff
@@ -185,10 +184,11 @@ func getTraceClient() (*clientset.Clientset, error) {
 	return traceClient, err
 }
 
-func printVersionSkewWarning(pods *corev1.PodList) {
+func PrintVersionSkewWarning(pods *corev1.PodList) {
```
mauriciovasquezbernal, line 187: is it needed to expose this function?

mauriciovasquezbernal (commented):
Some more comments.

How did you test it?

cmd/gadgetctl/main.go
```diff
@@ -75,7 +78,20 @@ func main() {
...
 		}
+		gadgetctlVersion := version.Version()
 		runtime.InitDeployInfo()
+		info, err := deployinfo.Load()
+		if err != nil {
+			log.Warnf("failed to load deploy info: %s", err)
+		}
+		serverSemver, _ := semver.Make(info.ServerVersion)
+
+		if gadgetctlVersion.Equals(serverSemver) {
```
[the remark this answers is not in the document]
moki1202: This is very silly of me! Apologies for the oversight.

cmd/kubectl-gadget/utils/trace.go
moki1202: Forgot to roll this back. Exposing this is definitely not needed.

cmd/gadgetctl/main.go
```diff
@@ -75,7 +78,20 @@ func main() {
...
 		runtime.InitDeployInfo()
+		info, err := deployinfo.Load()
+		if err != nil {
+			log.Warnf("failed to load deploy info: %s", err)
+		}
+		serverSemver, _ := semver.Make(info.ServerVersion)
+
+		if !gadgetctlVersion.Equals(serverSemver) {
+			log.Warnf(
+				"WARNING: version skew detected (gadget pods v%s vs gadgetctl v%s), use 'gadgetctl' to fix it.\n",
```
mauriciovasquezbernal, line 91: Same about warning and line break. Additionally:

```suggestion
			log.Warnf(
				"Version skew detected (gadget daemon v%s vs gadgetctl v%s)",
```

mauriciovasquezbernal (commented):
How are you testing it? I wasn't able to make it work without adding the following lines:

```diff
diff --git Makefile Makefile
index 65a978b03..632df1b5b 100644
--- Makefile
+++ Makefile
@@ -191,6 +191,7 @@ gadget-container:
                        BTFHUB_ARCHIVE=$(HOME)/btfhub-archive/ -j$(nproc); \
        fi
        docker buildx build --load -t $(CONTAINER_REPO):$(IMAGE_TAG) \
+               --build-arg VERSION=$(VERSION) \
                -f Dockerfiles/gadget.Dockerfile .
 
 .PHONY: cross-gadget-container
@@ -203,7 +204,7 @@ cross-gadget-container:
                        ARCH=arm64 BTFHUB_ARCHIVE=$(HOME)/btfhub-archive/ -j$(nproc); \
        fi
        docker buildx build --platform=$(PLATFORMS) -t $(CONTAINER_REPO):$(IMAGE_TAG) \
-               --push \
+               --push --build-arg VERSION=$(VERSION) \
                -f Dockerfiles/gadget.Dockerfile .
```

moki1202:
> ```diff
> --build-arg VERSION=$(VERSION) \
> ```

I haven't tested the changes yet, but It'll be the first thing I do once I submit my LFX applications.

moki1202:
@mauriciovasquezbernal while running make minikube-deploy. I get a panic message `panic: parsing version: Short version cannot contain PreRelease/Build meta data`. Now I assume this is caused since version takes the value de42d32b-dirty (not semantic) from makefile. But I think this would only occur while in development so should I be concerned here? I would want to test my changes though.

mauriciovasquezbernal:
> But I think this would only occur while in development so should I be concerned here?

Yes, it should work fine on development too. Specially because `make minikube-deploy` is designed for that purpose.

moki1202:
> Hi!
> 
> I tested it, and nothing is printed in the logs despite:
> 
> ```
> $ ./kubectl-gadget version                                   (remotes/moki/fix/1795) %
> Client version: 0.25.0-52-gde42d32b-dirty
> Server version: HEAD
> ```
> 
> I think we are missing some pieces here. I am particularly wandering if you should not add `--build-arg VERSION=` when you build the gadget container.
> 
> > About the Makefile changes I don't know what gadget-container and cross-gadget container does. I'll read about them and maybe then this change would make more sense to me 😅.
> 
> `make gadget-container` builds the gadget container image, _i.e._ the part which is deployed to k8s cluster when you call `./kubectl-gadget deploy`. `make cross-gadget-container` does the same but it cross compiles the image for all supported architectures, _i.e._ `arm64` and `amd64`.
> 
> Best regards.

@eiffel-fl when I run ./kubectl-gadget trace ... or similarly gadgetctl trace ... I get the warning message that we intend to display, Do we want it only for version command? Thankyou for the explanation :).

moki1202:
@eiffel-fl Adding `--build-arg` to gadget-container sets our server version same as client in dev env (`0.25.0-52-gde42d32b-dirty`). Do we need this result? And I had another doubt. When I deploy gadgets throught `kubectl gadget deploy`, the `serverVersion` is being passed as an empty string, Shouldn't it come out to be the same as what we got from the image tag method?

moki1202:
> > @eiffel-fl Adding `--build-arg` to gadget-container sets our server version same as client in dev env (`0.25.0-52-gde42d32b-dirty`). Do we need this result? And I had another doubt. When I deploy gadgets throught `kubectl gadget deploy`, the `serverVersion` is being passed as an empty string, Shouldn't it come out to be the same as what we got from the image tag method?
> 
> I do not understand your first question. If you compile the CLI and container image on the same commit, you should get the same version. Can you please detail the second case?
> 
> In my case, I get `v0.0.0`, which is normal because this is the value it has in the env: [de42d32#diff-ee77dad7f16fa4de1cca48323a9f9cb3f055ff1f9b0d3ed608d21ad19992d177R15-R16](https://github.com/inspektor-gadget/inspektor-gadget/commit/de42d32b066e62a78c9a3eb14bc6039fbbc92734#diff-ee77dad7f16fa4de1cca48323a9f9cb3f055ff1f9b0d3ed608d21ad19992d177R15-R16) So, this part is never called: [de42d32#diff-5404b12decf531eab5ad4fe9888d2fd13003355ad2c839e881bf3cd40d34d07aR6](https://github.com/inspektor-gadget/inspektor-gadget/commit/de42d32b066e62a78c9a3eb14bc6039fbbc92734#diff-5404b12decf531eab5ad4fe9888d2fd13003355ad2c839e881bf3cd40d34d07aR6) Moreover, we do not have `.git` in the gadget image: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/Dockerfiles/gadget.Dockerfile.dockerignore#L1

once I deploy the gadgets and run ./kubectl-gadget version, the server version updates, wouldn't we want this exact version to be set in serverSemver, that updates the gadget version in the warning message?

cmd/kubectl-gadget/main.go
```diff
@@ -125,7 +128,22 @@ func main() {
 	}
 
 	if !skipInfo {
+		kubectlGadgetVersion := version.Version()
 		grpcRuntime.InitDeployInfo()
+		info, err := deployinfo.Load()
+		if err != nil {
+			log.Warnf("failed to load deploy info: %s", err)
```
mauriciovasquezbernal, line 135:
```suggestion
			log.Warnf("Failed to load deploy info: %s", err)
```

cmd/kubectl-gadget/version.go
mauriciovasquezbernal, line 1: Now that we have the version available on the server, we can update this file too to use that instead of the image tag. Something like this should work fine:

```go
var versionCmd = &cobra.Command{
	Use:          "version",
	Short:        "Show version",
	SilenceUsage: true,
		fmt.Println("Client version:", version.Version())

		// Get the server version from the deploy information
		info, err := deployinfo.Load()
		if err != nil {
			return fmt.Errorf("loading deploy info: %w", err)
		}

		if info.ServerVersion == "" {
			fmt.Println("Server version: unknown (not deployed?)")
			return nil
		}

		fmt.Printf("Server version: %s\n", info.ServerVersion)
		return nil
	},
}
```

cmd/kubectl-gadget/main.go
```diff
@@ -125,7 +128,22 @@ func main() {
...
+		}
+		serverSemver, err := semver.Make(info.ServerVersion)
+		if err != nil {
+			log.Errorf("Failed to create a semantic version object for the gadget pods version: %s", err)
+		}
+		if !kubectlGadgetVersion.Equals(serverSemver) {
```
mauriciovasquezbernal, line 141: There should also be a check on `len(kubectlGadgetVersion.Pre) == 0` to avoid printing this version when it's not a release. (See logic on cmd/kubectl-gadget/utils/trace.go)

cmd/kubectl-gadget/main.go
```diff
@@ -125,7 +128,22 @@ func main() {
...
+			log.Warnf(
+				"version skew detected (gadget pods v%s vs kubectl-gadget v%s), use 'kubectl gadget deploy' to fix it.\n",
+				serverSemver, kubectlGadgetVersion,
+			)
```
mauriciovasquezbernal, line 145: I think all of this can be factorized out on a new helper function that is called from kubetl-gadget and gadgetctl.

moki1202: Should I put the helper function in ./internal/version dir in a seperate file? I was thinking a name like "versionChecker" maybe.

mauriciovasquezbernal: I think it can go to `cmd/common/utils`.

cmd/kubectl-gadget/main.go
moki1202: @mauriciovasquezbernal can gadgetctl have a pre release?

internal/version/version.go
mauriciovasquezbernal: [SPDX](https://spdx.dev/) license headers are shorter than the standard ones we use (`// SPDX-License-Identifier: Apache-2.0` instead of the multi-line one). 

In any case, this is a discussion that doesn't belong to this PR, so please continue to use the same license header we're using in other files.

cmd/kubectl-gadget/main.go
mauriciovasquezbernal: Yes, the versioning logic is the same for all components.

cmd/kubectl-gadget/version.go
moki1202: I'll do this in a follow up PR as suggested by @eiffel-fl .

moki1202:
@mauriciovasquezbernal @eiffel-fl done with all the changes, hopefully this should be good to go. Thank you to both for the constant support and help, really appreciate it. I learned A LOT!

cmd/common/utils/version-checker.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/blang/semver"
+	log "github.com/sirupsen/logrus"
+)
+
+type VersionCheckParams struct {
+	ClientVersion         semver.Version
+	ServerVersion         string
+	SemanticVersionErrMsg string
+	VersionSkewWarningMsg string
+}
```
mauriciovasquezbernal, line 13: I don't think we need to pass a lot of parameters to this function. What about something like:

```diff
diff --git cmd/common/utils/version-checker.go cmd/common/utils/version-checker.go
index e20329c85..4e29045f6 100644
--- cmd/common/utils/version-checker.go
+++ cmd/common/utils/version-checker.go
@@ -1,30 +1,24 @@
 package utils
 
 import (
+	"fmt"
+
 	"github.com/blang/semver"
-	log "github.com/sirupsen/logrus"
+	"github.com/inspektor-gadget/inspektor-gadget/internal/version"
 )
 
-type VersionCheckParams struct {
-	ClientVersion         semver.Version
-	ServerVersion         string
-	SemanticVersionErrMsg string
-	VersionSkewWarningMsg string
-}
+func CheckServerVersionSkew(serverVersion string) error {
+	clientSemver := version.Version()
 
-func CheckVersionSkew(vc VersionCheckParams) {
-	if len(vc.ClientVersion.Pre) > 0 {
-		return
-	} else {
-		serverSemver, err := semver.Make(vc.ServerVersion)
-		if err != nil {
-			log.Errorf(vc.SemanticVersionErrMsg, err)
-		}
-		if !vc.ClientVersion.Equals(serverSemver) {
-			log.Warnf(
-				vc.VersionSkewWarningMsg,
-				serverSemver, vc.ClientVersion,
-			)
-		}
+	// Do not print any warning if this is a prerelease to avoid annoying developers.
+	if len(clientSemver.Pre) > 0 {
+		return nil
 	}
+
+	serverSemver, err := semver.Make(serverVersion)
+	if err != nil {
+		return fmt.Errorf("parsing semver: %w", err)
+	}
+
+	return fmt.Errorf("version skew detected: client (%s) vs server (%s)", clientSemver, serverSemver)
 }
diff --git cmd/gadgetctl/main.go cmd/gadgetctl/main.go
index 69e4cd9f8..581aaff6c 100644
--- cmd/gadgetctl/main.go
+++ cmd/gadgetctl/main.go
@@ -26,7 +26,6 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/internal/deployinfo"
-	"github.com/inspektor-gadget/inspektor-gadget/internal/version"
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/all-gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/environment"
 	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
@@ -81,14 +80,9 @@ func main() {
 		info, err := deployinfo.Load()
 		if err != nil {
 			log.Warnf("Failed to load deploy info: %s", err)
+		} else if err := commonutils.CheckServerVersionSkew(info.ServerVersion); err != nil {
+			log.Warnf(err.Error())
 		}
-		checkParams := commonutils.VersionCheckParams{
-			ClientVersion:         version.Version(),
-			ServerVersion:         info.ServerVersion,
-			SemanticVersionErrMsg: "Failed to create a semantic version object for the gadget daemon version: %s",
-			VersionSkewWarningMsg: "Version skew detected (gadget daemon v%s vs gadgetctl v%s)",
-		}
-		commonutils.CheckVersionSkew(checkParams)
 	}
 
 	hiddenColumnTags := []string{"kubernetes"}
diff --git cmd/kubectl-gadget/main.go cmd/kubectl-gadget/main.go
index 2f4406687..865272098 100644
--- cmd/kubectl-gadget/main.go
+++ cmd/kubectl-gadget/main.go
@@ -23,7 +23,6 @@ import (
 
 	// Import this early to set the enrivonment variable before any other package is imported
 	"github.com/inspektor-gadget/inspektor-gadget/internal/deployinfo"
-	"github.com/inspektor-gadget/inspektor-gadget/internal/version"
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/environment/k8s"
 	paramsPkg "github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 
@@ -131,14 +130,9 @@ func main() {
 		info, err := deployinfo.Load()
 		if err != nil {
 			log.Warnf("Failed to load deploy info: %s", err)
+		} else if err := commonutils.CheckServerVersionSkew(info.ServerVersion); err != nil {
+			log.Warnf(err.Error())
 		}
-		checkParams := commonutils.VersionCheckParams{
-			ClientVersion:         version.Version(),
-			ServerVersion:         info.ServerVersion,
-			SemanticVersionErrMsg: "Failed to create a semantic version object for the gadget pods version: %s",
-			VersionSkewWarningMsg: "Version skew detected (gadget pods v%s vs kubectl-gadget v%s), use 'kubectl gadget deploy' to fix it.\n",
-		}
-		commonutils.CheckVersionSkew(checkParams)
 	}
 
 	namespace, _ := utils.GetNamespace()

```

cmd/common/utils/version-checker.go
```diff
@@ -0,0 +1,30 @@
+package utils
```
mauriciovasquezbernal, line 1: nit: license header.

cmd/common/utils/version-checker.go
```diff
@@ -0,0 +1,30 @@
...
+	ClientVersion         semver.Version
+	ServerVersion         string
+	SemanticVersionErrMsg string
+	VersionSkewWarningMsg string
+}
+
+func CheckVersionSkew(vc VersionCheckParams) {
+	if len(vc.ClientVersion.Pre) > 0 {
+		return
+	} else {
```
mauriciovasquezbernal, line 18: nit: else not needed after this return.

mauriciovasquezbernal (commented):
Some more nits. Hopefully it'll be ready after it.

moki1202:
When I run `make lint` I get the following error-
```
docker run --rm --env XDG_CACHE_HOME=/tmp/xdg_home_cache \
        --env GOLANGCI_LINT_CACHE=/tmp/golangci_lint_cache \
        --user 1000:1000 -v /home/moki1202/go/src/github.com/moki1202/inspektor-gadget:/app -w /app \
        linter
Unable to find image 'linter:latest' locally
docker: Error response from daemon: pull access denied for linter, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```
I'm pretty sure this image doesn't exist on docker hub.

mauriciovasquezbernal:
Did you receive another error / warning before? `make lint` tries to build the linter image before:

https://github.com/inspektor-gadget/inspektor-gadget/blob/cf6328184d7eef5067f24c64047fd74411907f0f/Makefile#L291-L300

mauriciovasquezbernal (commented):
@moki1202 please squash the commits in a single one.

mauriciovasquezbernal:
>     * several tests failed with:
> 
> 
> ```
> command.go:517: Stop command(RunWatchContainers)
> command.go:519: Command returned(RunWatchContainers):
> panic: parsing version: Invalid character(s) found in major number "e57cd31"
> 
> goroutine 1 [running]:
> github.com/inspektor-gadget/inspektor-gadget/internal/version.init.0()
> 	/go/src/github.com/inspektor-gadget/inspektor-gadget/internal/version/version.go:33 +0x154
> ```
> 
> I guess this is the commit hash.
> 

You're right. The checkout action we're using doesn't fetch the tags by default, hence `git describe --tags --always --dirty` returns only the commit (which can't be parse to a semver):

```bash
$ mkdir /tmp/foo && cd /tmp/foo
$ git clone git@github.com:moki1202/inspektor-gadget.git --depth=1 -b "fix/1795"
...
$ cd inspektor-gadget/

$ git describe --tags --always --dirty
cf63281 

$ make ig
...

$ ./ig
panic: parsing version: Invalid character(s) found in major number "cf63281"

goroutine 1 [running]:
github.com/inspektor-gadget/inspektor-gadget/internal/version.init.0()
	/go/src/github.com/inspektor-gadget/inspektor-gadget/internal/version/version.go:33 +0x154
```

I think panicking when we're not able to parse the version is too much, I'll suggest logging a warning instead:

```diff
$ git diff
diff --git internal/version/version.go internal/version/version.go
index fc0835f..03a3c74 100644
--- internal/version/version.go
+++ internal/version/version.go
@@ -17,9 +17,8 @@
 package version
 
 import (
-       "fmt"
-
        "github.com/blang/semver"
+       log "github.com/sirupsen/logrus"
 )
 
 // version is filled out by the Makefile at build
@@ -30,7 +29,7 @@ func init() {
        var err error
        parsedVersion, err = semver.ParseTolerant(version)
        if err != nil {
-               panic(fmt.Sprintf("parsing version: %s", err))
+               log.Warnf("parsing version %q: %s", version, err)
        }
 }
```

mauriciovasquezbernal:
@moki1202 do you have plans to continue working on this?

moki1202:
@mauriciovasquezbernal I'll try completing this weekend.

mauriciovasquezbernal:
I continued the work on #2718. Thanks @moki1202 for starting working on this.

PR #2448, opened 2024-02-03
container-collection/k8s: auto detect container runtime socket path

### LINKED ISSUE #2447: [RFE] auto detect runtime's socket path in Kubernetes
## Current situation

Today, when starting IG in Kubernetes, the socket path is determined by an environment variables or falls back to a default based on the container runtime detected.

## Impact

The socket path needs to be pre-determined as part of the IG installation.

## Ideal future situation

In Kubernetes, retrieve the socket path from the node.

## Implementation options

In Kubernetes, we can query the API server to retrieve the kubelet configuration, which includes the container runtime endpoint

dorser:
# Auto detect container runtime socket path

The socket path of the container runtime can vary between different distributions. In order to find the socket path automatically, we query the kubelet configuration from the API Server.

Resolves #2447 

## How to use

Use container-collection with `WithInitialKubernetesContainers` or `WithhPodInformer` options.

## Testing done

* Executed ig-integration.test on a K3s deployment
* Used the example exec tracer program `WithInitialKubernetesContainers` on a K3s deployment


pkg/container-collection/k8s.go
```diff
@@ -53,6 +57,12 @@ func NewK8sClient(nodeName string) (*K8sClient, error) {
 
 	fieldSelector := fields.OneTermEqualSelector("spec.nodeName", nodeName).String()
 
+	kubeletConfig, err := getCurrentKubeletConfig(clientset, nodeName)
```
omriShneor, line 60: This method can return nil in kubeletConfig, currently you're only logging the error which will cause a runtime error in line 65 in that case when you do: `kubeletConfig.ContainerRuntimeEndpoint`
How do you want to handle that case?

dorser: Returning an error message upon failure

pkg/container-collection/k8s.go
```diff
@@ -181,3 +192,42 @@ func (k *K8sClient) ListContainers() (arr []Container, err error) {
 	}
 	return arr, nil
 }
+
+func getCurrentKubeletConfig(clientset *kubernetes.Clientset, nodeName string) (*kubeletconfig.KubeletConfiguration, error) {
```
galofir-ms, line 196: what are the required permissions for requesting the kubelet config?

go.mod
```diff
@@ -66,6 +66,11 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
+	k8s.io/kubelet v0.29.1
+	k8s.io/kubernetes v1.29.1
```
galofir-ms, line 71: from which k8s versions are these supported? is the go pkg version correlated with the k8s release version?

pkg/container-collection/k8s.go
```diff
@@ -53,6 +57,12 @@ func NewK8sClient(nodeName string) (*K8sClient, error) {
 
 	fieldSelector := fields.OneTermEqualSelector("spec.nodeName", nodeName).String()
 
+	kubeletConfig, err := getCurrentKubeletConfig(clientset, nodeName)
+	if err != nil {
+		return nil, fmt.Errorf("getting /configz %w", err)
+	}
+	socketPath, _ := strings.CutPrefix(kubeletConfig.ContainerRuntimeEndpoint, "unix:///")
```
galofir-ms, line 64: which mounts are required for socketPath discovery

pkg/container-collection/k8s.go
```diff
@@ -181,3 +192,42 @@ func (k *K8sClient) ListContainers() (arr []Container, err error) {
 	}
 	return arr, nil
 }
+
+func getCurrentKubeletConfig(clientset *kubernetes.Clientset, nodeName string) (*kubeletconfig.KubeletConfiguration, error) {
+	resp, err := clientset.CoreV1().RESTClient().Get().Resource("nodes").Name(nodeName).Suffix("proxy", "configz").DoRaw(context.TODO())
```
alban, line 197: How does this `Suffix("proxy", "configz")` work? Would the API server contact the Kubelet to fetch its config or is it cached in the node resource?

Do you know if this API is documented somewhere?

I wonder what are the pros and cons of this code compared to looking at the node resource
```
  daemonEndpoints:
    kubeletEndpoint:
      Port: 10250
```
and contact the kubelet directly. I guess your method will work even when `ig` runs in a pod in its own network namespace?

pkg/container-collection/k8s.go
dorser: Good call. We need to add `get` for `nodes/proxy` in the cluster role

pkg/container-collection/k8s.go
dorser: Discovering the socket endpoint doesn't require any new mounts as we find it by getting the kubelet config.
Once we find the socket, In theory, it can be anywhere on the host, but it's most likely to under `/run`.

pkg/container-collection/k8s.go
dorser: The kubelet API is properly documented but the /configz endpoint is mentioned in the documentation in a few places, like: https://k8s-docs.netlify.app/en/docs/tasks/administer-cluster/reconfigure-kubelet/#generate-the-configuration-file

Going through the API server solves problems around auth I stumbled upon when I tried to connect directly to kubelet.

What I'm doing here is equivalent to:
```
kubectl get --raw "/api/v1/nodes/dorserero-dev/proxy/configz"
```

And if we were to connect to kubelet directly it's:
```
curl --insecure -v https://localhost:10250/configz
```
Assuming the kubelet endpoint port is 10250.

Since this happens only once, proxying the request through the API server sounds reasonable to me.

pkg/container-collection/k8s.go
```diff
@@ -181,3 +192,42 @@ func (k *K8sClient) ListContainers() (arr []Container, err error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch /configz from %q", nodeName)
```
alban, line 199:
```suggestion
		return nil, fmt.Errorf("fetching /configz from %q: %w", nodeName, err)
```

pkg/container-collection/k8s.go
```diff
@@ -53,6 +57,12 @@ func NewK8sClient(nodeName string) (*K8sClient, error) {
 
 	fieldSelector := fields.OneTermEqualSelector("spec.nodeName", nodeName).String()
 
+	kubeletConfig, err := getCurrentKubeletConfig(clientset, nodeName)
+	if err != nil {
+		return nil, fmt.Errorf("getting /configz %w", err)
```
alban, line 62:
```suggestion
		return nil, fmt.Errorf("getting /configz: %w", err)
```

pkg/container-collection/k8s.go
alban: * Update charts/gadget/templates/clusterrole.yaml
* `make generate-manifests`

pkg/container-collection/k8s.go
alban: /run is already a host volume, so that should be fine.

pkg/container-collection/k8s.go
alban: I would like to know which version of Kubernetes introduced this `proxy/configz`.
If it is recent, we should document it on:
* docs/getting-started/requirements.md
* https://inspektor-gadget.io/docs/latest/getting-started/requirements/

alban (commented):
Thanks! I think that will make a good addition.

pkg/container-collection/k8s.go
dorser: I looked specifically at 1.26 as this is the oldest supported version at this time. 
The `/api/v1/nodes/{name}/proxy/{path}` endpoint exists in [1.26](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/).

The `/configz` endpoint is undocumented but there's this comment: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/configz/OWNERS

As well as this commit where it was first introduced:
https://github.com/kubernetes/kubernetes/pull/20528

mauriciovasquezbernal:
I think this is still relevant. @dorser do you have any plans to complete this PR? What's missing?

dorser:
Time :) 
I haven't been around for a while. I'll publish a new revision next week and hopefully we can close it.

PR #2451, opened 2024-02-05
pkg/columns: Do not elide output when redirected.

eiffel-fl:
Hi.

This PR removes ellipsis when the output is not a terminal:

```bash
$ sudo -E ./ig image list | awk '{ printf "%s:%s\n", $1, $2 }' | tail -2
blocks:latest
ghcr.io/inspektor-gadget/gadgets/blocks:latest
```

Best regards.

### DIFF
--- a/pkg/columns/formatter/textcolumns/options.go
+++ b/pkg/columns/formatter/textcolumns/options.go
@@ -37,6 +37,7 @@ type Options struct {
 	DefaultColumns []string    // defines which columns to show by default; will be set to all visible columns if nil
 	HeaderStyle    HeaderStyle // defines how column headers are decorated (e.g. uppercase/lowercase)
 	RowDivider     string      // defines the (to be repeated) string that should be used below the header
+	ShouldTruncate bool        // defines whether to truncate strings or not
 }
 
 func DefaultOptions() *Options {
@@ -46,6 +47,7 @@ func DefaultOptions() *Options {
 		DefaultColumns: nil,
 		HeaderStyle:    HeaderStyleUppercase,
 		RowDivider:     DividerNone,
+		ShouldTruncate: true,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


mqasimsarfraz:
Thanks for working on this, it will be nice to have this. 

Also, what do you think about combining it with [--no-trunc](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/image/list.go#L59) ?

flyth:
Maybe this should rather become an [`Option`](https://github.com/inspektor-gadget/inspektor-gadget/blob/5061f9e7333b4088d20dde98314537915f8fe0c2/pkg/columns/formatter/textcolumns/options.go#L32).

Otherwise we lose the possibility of formatting output for other targets (e.g. when used as a service and sending formatted output to another logserver or such), because only the local environment would be considered.

pkg/columns/formatter/textcolumns/options.go
```diff
@@ -37,6 +37,7 @@ type Options struct {
 	DefaultColumns []string    // defines which columns to show by default; will be set to all visible columns if nil
 	HeaderStyle    HeaderStyle // defines how column headers are decorated (e.g. uppercase/lowercase)
 	RowDivider     string      // defines the (to be repeated) string that should be used below the header
+	DoEllipsis     bool        // defines whether to add or not ellipsis.
```
flyth, line 40: Maybe `TruncateStrings` would be a better and more generic fit, because right now this defines whether or not to shorten the strings, not whether an ellipsis will be used.

flyth (commented):
Looks good, thanks! I'd just rename the field + option as it now just decides whether or not to shorten/truncate the string, not whether an ellipsis is used.

flyth (approved):
LGTM, thanks!

Issue #2038, opened 2023-09-14
CI: benchmarks step failing when multiple PRs are opened at the same time


mauriciovasquezbernal:
dependabot opens a lot of PRs at the same time to update dependencies, for instance (#2034, #2037, #2035), in all PRs the benchmarks step fails to push the results to the repo with the following error:

```
 ! [rejected]        gh-pages -> gh-pages (fetch first)
error: failed to push some refs to 'https://github.com/inspektor-gadget/ig-benchmarks.git'
hint: Updates were rejected because the remote contains work that you do not
hint: have locally. This is usually caused by another repository pushing to
hint: the same ref. If you want to integrate the remote changes, use
hint: 'git pull' before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
Warning: Auto-push failed because the remote gh-pages was updated after git pull
/usr/bin/git --work-tree=./benchmark-data-repository --git-dir=./benchmark-data-repository/.git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= reset --hard HEAD~1
HEAD is now at a03d40b add Gadget benchmarks (go) benchmark result for b3cf28c9a6a96de28e2ab66253ff4549049960e4
Warning: Retrying to generate a commit and push to remote gh-pages with retry count 10...
/usr/bin/git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= clone ***github.com/inspektor-gadget/ig-benchmarks.git ./benchmark-data-repository
fatal: destination path './benchmark-data-repository' already exists and is not an empty directory.
Error: Command 'git' failed with args '-c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= clone ***github.com/inspektor-gadget/ig-benchmarks.git ./benchmark-data-repository': fatal: destination path './benchmark-data-repository' already exists and is not an empty directory.
: Error: The process '/usr/bin/git' failed with exit code 128
```

(taken from https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/6180518847/job/16777093468?pr=2034)

It seems the benchmark-action/github-action-benchmark action has a logic to recover from that, but that logic is not failing, it seems to be related to https://github.com/benchmark-action/github-action-benchmark/issues/182.

ktrz:
[...]

PR #2453, opened 2024-02-05
integration: Publish aggregate test report summary

mqasimsarfraz:
Currently, getting overview of test failures in the CI can be challenging. This change addresses that by extending the work we did for summaries and enhances that by storing them to a GitHub page. 

It took inspiration from the `ig-benchmarks` and commits the reports on each change. [Similarly](https://github.com/benchmark-action/github-action-benchmark/pull/216/files), uses `retry` to avoid running into multiple `git push` issue.

## Testing Done

Reports: https://inspektor-gadget.github.io/ig-test-reports/

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1229,6 +1229,7 @@ jobs:
         with:
           test-log-file: integration.log
           test-step-conclusion: ${{ steps.integration-tests.conclusion }}
+          test-summary-suffix: ${{ matrix.runtime }}
 
   build-and-push-gadgets:
     name: Build and push gadgets
@@ -1316,6 +1317,7 @@ jobs:
         with:
           test-log-file: integration.log
           test-step-conclusion: ${{ steps.integration-tests.conclusion }}
+          test-summary-suffix: ${{ matrix.runtime }}
 
   test-integration-aks:
     name: Integration tests on AKS
@@ -1400,6 +1402,7 @@ jobs:
         image_tag: ${{ steps.set-repo-determine-image-tag.outputs.image-tag }}
         gadget_repository: ${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }}
         gadget_tag: ${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}
+        test_summary_suffix: "${{ matrix.os-sku }}-${{ matrix.arch }}"
     # This step ensures we get fresh credentials before removing the cluster
     - uses: azure/login@v1
       if: always()
@@ -1468,6 +1471,7 @@ jobs:
         image_tag: ${{ steps.set-repo-determine-image-tag.outputs.image-tag }}
         gadget_repository: ${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }}
         gadget_tag: ${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}
+        test_summary_suffix: "aro"
 
   test-integration-minikube:
     name: Integr. tests
@@ -1522,6 +1526,7 @@ jobs:
         dnstester_image: ${{ needs.build-helper-images.outputs.dnstester_image }}
         gadget_repository: ${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }}
         gadget_tag: ${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}
+        test_summary_suffix: ${{ matrix.runtime }}
     - if: ${{ matrix.runtime == 'docker' }}
       name: Check that README is up-to-date
       # We do this check here, as we need to run kubectl-gadget against a deployed installation to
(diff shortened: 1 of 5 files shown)


alban:
That looks great.

Will the web page display all previous tests or only  the last few ones? If it displays everything, that would be a lot of columns over time, and make it difficult to read.

mauriciovasquezbernal (commented):
It looks really great. I don't have so many comments, besides Alban's comment above I don't have anything to add. I think we should we the last N runs, N == 100?, and that's it.

mqasimsarfraz:
> If it displays everything, that would be a lot of columns over time, and make it difficult to read.

:100: 

> Will the web page display all previous tests or only the last few ones? 

I currently limit it to [last 10](https://github.com/inspektor-gadget/ig-test-reports/blob/main/index.html#L90) only. We can change this value in ig-test-reports later on as well.

mauriciovasquezbernal:
> I currently limit it to [last 10](https://github.com/inspektor-gadget/ig-test-reports/blob/main/index.html#L90) only. We can change this value in ig-test-reports later on as well.

Should we limit the size of the json somehow? I suppose it'll become to big and difficult to handle very soon.

mqasimsarfraz:
> > I currently limit it to [last 10](https://github.com/inspektor-gadget/ig-test-reports/blob/main/index.html#L90) only. We can change this value in ig-test-reports later on as well.
> 
> Should we limit the size of the json somehow? I suppose it'll become to big and difficult to handle very soon.

Great catch! I adjusted to only store last [100 results](https://github.com/inspektor-gadget/inspektor-gadget/pull/2453/files#diff-bda633e49a0d242b6842493c3896d4f766f361166fa4c08e1c2e01d827544cadR1595). I believe we should be fine with that but if that will be too much we can decrease this in future.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1531,6 +1536,84 @@ jobs:
...
+      (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == 'inspektor-gadget/inspektor-gadget'))
+    steps:
+    - uses: actions/checkout@v4
+    - name: Get all reports
+      uses: actions/download-artifact@v3
+      with:
+        name: "test-reports"
+    - name: Store test reports
+      shell: bash {0}
+      run: |
```
mauriciovasquezbernal, line 1562: Perhaps move this to something under https://github.com/inspektor-gadget/inspektor-gadget/tree/main/tools to avoid adding more lines to this file?

mauriciovasquezbernal:
> Also, with regard to the code itself, you make a parallel with `benchmarks` which has been really flaky recently...
> If possible, I would like to avoid adding other flaky tests in our CI as it just decreases the reliability over time.
> I understand this is indeed complicated to guess whether something will be flaky but the `retry` function does not really lead me to the path where I think this will not be flaky.

I think the logic introduced in the retry function is a good enough effort to avoid this job failing when other instances of it  run on parallel on different machines. Perhaps we can also use a concurrency group (https://docs.github.com/en/actions/using-jobs/using-concurrency) to avoid multiple instances of this job running on parallel and reduce the change of a collision?

mauriciovasquezbernal (commented):
It's looking very good. What do you think about moving the script to a different file? I think that would help to edit it easier (highlight syntax and other stuff will help too). 

Also, I can approve it now if you want to merge as it is. I saw the result and it's very nice, we can always improve it later on if needed. It's up to you how much you want to improve it before merging.

mqasimsarfraz:
> so use a concurrency group (https://docs.github.com/en/actions/using-jobs/using-concurrency) to avoid multiple instances of this job running on parallel and reduce the change of a collision?

I also thought about this but the reason I decided against it since it only allows single job in pending state?

> Any previously pending job or workflow in the concurrency group will be canceled. To also cancel any currently running job or workflow in the same concurrency group, specify cancel-in-progress: true.

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: Good idea! I will do it!

mauriciovasquezbernal:
So in that case we'll missing the entry for the skipped job in the results? 
I agree the current approach is better then, let's go for it.

.github/workflows/inspektor-gadget.yml
mqasimsarfraz: Done.

mauriciovasquezbernal (approved):

mqasimsarfraz:
> > The issue with benchmarks was [#2038 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/issues/2038#issuecomment-1910967021) so the idea was to see if we can handle the things in similar way to avoid issues in first place.
> 
> You made me doubt, but there is still an issue with benchmarks: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7810164205/job/21303171942?pr=2460 Maybe this is a different one, but this will change that I deem this to be flaky as hell.

The error does look different but I see the point. But I don't think we should block this change on it. I can keep a close on eye on next few builds after we merge this change and would be happy to switch to other approaches (e.g concurrency group) if they are more stable yet with less coverage. Flaky CI is the last thing I want :smile:

mqasimsarfraz:
I will merge this and keep a close eye on flakiness. Hopefully we will be able to identity failures faster!

Issue #2413, opened 2024-01-26
`ig` + `cri-o` panic on enriching some `imageRef` fields 


burak-ok:
## `ig` + `cri-o` panic on enriching some `imageRef` fields 

On my local cri-o minikube I get the following error:
```
root@minikube-cri-o:~# ig list-containers --watch -o json
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cri.CRIContainerToContainerData({0x27d9493, 0x5}, {0x2b58230, 0xc0002ce850})
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cri/cri.go:306 +0x845
github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cri.(*CRIClient).GetContainers(0xc00005d6c0)
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/cri/cri.go:94 +0x8f
github.com/inspektor-gadget/inspektor-gadget/pkg/ig-manager.NewManager.WithMultipleContainerRuntimesEnrichment.WithContainerRuntimeEnrichment.func10(0xc000549600)
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection/options.go:169 +0x379
github.com/inspektor-gadget/inspektor-gadget/pkg/ig-manager.NewManager.WithMultipleContainerRuntimesEnrichment.func6(0xc000549600?)
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection/options.go:116 +0x56
github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection.(*ContainerCollection).Initialize(0xc000549600, {0xc00058d4c0, 0x8, 0x0?})
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection/container-collection.go:103 +0x9b
github.com/inspektor-gadget/inspektor-gadget/pkg/ig-manager.NewManager({0xc00056e380, 0x4, 0x0?})
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/ig-manager/ig-manager.go:110 +0x565
github.com/inspektor-gadget/inspektor-gadget/cmd/ig/containers.NewListContainersCmd.func1(0xc00081ec00, {0x27d86ff, 0x4, 0x27d8703})
[...]

PR #2455, opened 2024-02-05
cri: check range before accessing imageRef

alban:
panic on enriching some imageRef fields:
panic: runtime error: index out of range [1] with length 1

Based on work from: @burak-ok 

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/2413

The patch originally comes from #2398


blanquicet (approved):
LGTM from code inspection.

PR #2456, opened 2024-02-06
pkg/columns: Add support for map[string]string

alban:
This patch is coming from #2398 by @burak-ok. Since it is independent from the rest, it could be merged separately.


flyth (approved):
Looks good from code inspection!

alban:
Thanks, I will merge once the CI is green.

Issue #2008, opened 2023-09-07
Dns Tracer - Uncompressed Messages Support


amitschendel:
## Description

The current dns tracer doesn't show the addresses of dns messages which aren't compressed as mentioned in the tracer docs.
@mauriciovasquezbernal told me to open an issue about it so you have track on it.

## Impact

Limited tracer output.

## Environment and steps to reproduce

Run the dns tracer in a local cluster such as minikube and try to resolve in cluster services.

## Expected behavior

The dns messages will probably be small and therefor coredns won't compress them so we won't see addresses output.

## Additional information

----

mauriciovasquezbernal:
Additional context: It's not supported right now because doing this on eBPF is very difficult. We're in the process of implementing our gadgets as "containerized gadgets" (https://github.com/inspektor-gadget/inspektor-gadget/issues/1929), I think we could support this by using wasm as done in this poc -> https://github.com/inspektor-gadget/inspektor-gadget/pull/1985

PR #2467, opened 2024-02-07
trace dns: support uncompressed messages

alban:
See:
https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.4

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/2008

## How to use

## Testing done

```patch
diff --git a/tools/dnstester/dnstester.go b/tools/dnstester/dnstester.go
index 3939fea5..c1c5eb27 100644
--- a/tools/dnstester/dnstester.go
+++ b/tools/dnstester/dnstester.go
@@ -23,7 +23,7 @@ import (
 func main() {
        dns.Handle(".", dns.HandlerFunc(func(w dns.ResponseWriter, r *dns.Msg) {
                m := new(dns.Msg)
-               m.Compress = true
+               m.Compress = false
                m.SetReply(r)
 
                var rr dns.RR
@@ -31,7 +31,7 @@ func main() {
                if r.Question[0].Name == "fake.test.com." {
                        switch r.Question[0].Qtype {
                        case dns.TypeA:
-                               rr, err = dns.NewRR("fake.test.com. A 127.0.0.1")
+                               rr, err = dns.NewRR("fake.test.com. A 127.0.0.16")
                        case dns.TypeAAAA:
                                rr, err = dns.NewRR("fake.test.com. AAAA ::1")
                        }
@@ -49,7 +49,7 @@ func main() {
 
        }))
 
-       server := &dns.Server{Addr: ":53", Net: "udp"}
+       server := &dns.Server{Addr: ":5454", Net: "udp"}
        if err := server.ListenAndServe(); err != nil {
                log.Fatalf("Failed to start dns server %s\n", err)
        }
```

```bash
$ go run dnstester.go
```

```
$ go run -exec 'sudo -E' ./cmd/ig/... trace dns --host -o columns=pid,tid,comm,qr,type,qtype,name,rcode,numAnswers,addresses
PID         TID         COMM        QR TYPE      QTYPE       NAME                RCODE      NUMA… ADDRESSES           
319113      319115      isc-net-00… Q  OUTGOING  A           fake.test.com.                 0                         
319105      319105      dnstester   Q  HOST      A           fake.test.com.                 0                         
319105      319108      dnstester   R  OUTGOING  A           fake.test.com.      NoError    1     127.0.0.16          
319113      319115      isc-net-00… R  HOST      A           fake.test.com.      NoError    1     127.0.0.16          
```
```
$ nslookup -type=a -port=5454 fake.test.com. 127.0.0.1
```


alban:
This patch works for direct `A` or `AAAA` replies but it does not work when there are `CNAME` redirections.

To simplify the parsing, I've made the assumption that the name being queried is the same used in `A` or `AAAA` replies but that assumption is incorrect with `CNAME` redirections. I've made this simplification because otherwise I had the following bpf verifier error:

```
BPF program is too large. Processed 1000001 insn
```

I wonder what's the best approach:
* Take the current patch, it works for some cases but not for CNAME... Quick partial solution but incomplete.
* Use bpf_loop() instead of unrolling loops at compilation (to avoid complexity explosion). But bpf_loop() requires Linux >= 5.17.
* Send the dns packet to userspace and use a proper dns library such as github.com/miekg/dns to parse the packet. It would take more time to rewrite the dns gadget...

mauriciovasquezbernal:
I suppose we don't need it anymore now that #2544 is merged. WDYT  @alban?

alban:
Correct, I will close this PR.

PR #2469, opened 2024-02-07
docs/devel/ci: Use only subscription ID to create AKS federated credentials

blanquicet:
# Use only subscription ID to create AKS federated credentials

We don't need the subscription name and ID. Everything can be done just with the ID.

BTW @eiffel-fl, great guide! It worked perfectly.


blanquicet:
Merging as for sure CI failure is not related with this PR.

PR #2475, opened 2024-02-08
image-based trace exec gadget: fix sys_exit_execve from thread

alban:
Same as 73acd49caedb ("trace exec: fix sys_exit_execve from thread") from #2454 but for the image-based gadget.

sys_enter_execve adds an entry in the "execs" map and sys_exit_execve removes it. However, sys_enter_execve and sys_exit_execve can be called from different thread id in case of a multithreaded application. This caused the map to not get cleaned up properly.

This patch fixes the tracking between sys_enter_execve and sys_exit_execve.

## How to use

No changes

## Testing done

Tested with:
```
make -C gadgets/ trace_exec
sudo -E ./ig run ghcr.io/inspektor-gadget/gadget/trace_exec:alban_fix_execve_threads2
```

Same scenarios as in #2454.

### DIFF
--- a/gadgets/trace_exec/program.bpf.c
+++ b/gadgets/trace_exec/program.bpf.c
@@ -59,6 +59,36 @@ GADGET_TRACER_MAP(events, 1024 * 256);
 
 GADGET_TRACER(exec, events, event);
 
+// man clone(2):
+//   If any of the threads in a thread group performs an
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #2480, opened 2024-02-08
ci: Enable running tests on release branch

mauriciovasquezbernal:
Now that we're backporting fixes, we need to run the CI in the release branch too.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -16,6 +16,7 @@ on:
   push:
     branches:
     - main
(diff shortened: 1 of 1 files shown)


flyth (approved):
LGTM

blanquicet:
But in the end, you will tag it as `v*`, so it should trigger the branch, isn't it?

mauriciovasquezbernal:
> But in the end, you will tag it as `v*`, so it should trigger the CI, isn't it?

That's a good point. However I think it's better to have test passing on the release branch before pushing the tag.

Issue #975, opened 2022-09-21
integration: Improve Seccomp Advisor tests


mauriciovasquezbernal:
The [integration tests for the Seccomp Advisor](https://github.com/kinvolk/inspektor-gadget/blob/6843c673b7f1d69e639529c852aff3e61903dbea/integration/integration_test.go#L839-L857) are only covering a simple case: Print a profile to the screen on-demand. There are other cases that we should cover too:

- Generate the profile upon pod termination
- Create the pod before / after starting the tracer
- Test different output modes ("Status", "Stream", "ExternalResource")

Ref https://github.com/kinvolk/inspektor-gadget/pull/948#discussion_r975069229

PR #2486, opened 2024-02-09
integration: Disable installing Security Profile Operator

mauriciovasquezbernal:
The security profile operator is only used by TestAuditSeccomp to install a simple seccomp profile in the cluster. This is an overkill.

This commit uses a simpler approach by copying the profile by using a DaemonSet to the node and disables installing this by default. The code to deploy/undeploy the SPO isn't removed as it could be useful to test the "advise seccomp" gadget in the future.

Some reasons to avoid deploying the SPO are:
- It needs quite a lot of resources
- It takes a lot of time to be deployed
- Some times it fails to be deployed, and also to be cleanup, leaving it running on the cluster

### DIFF
--- a/integration/inspektor-gadget/audit_seccomp_test.go
+++ b/integration/inspektor-gadget/audit_seccomp_test.go
@@ -25,8 +25,6 @@ import (
 
 func TestAuditSeccomp(t *testing.T) {
 	ns := GenerateTestNamespaceName("test-audit-seccomp")
-	spName := "log"
-
 	t.Parallel()
 
 	// TODO: Handle it once we support getting container image name from docker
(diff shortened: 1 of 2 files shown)


integration/inspektor-gadget/audit_seccomp_test.go
```diff
@@ -25,8 +25,6 @@ import (
 
 func TestAuditSeccomp(t *testing.T) {
```
mqasimsarfraz, line 26: Just wondering if we should also have a `TestAuditSeccompSPO` which allows testing with SPO and is guarded by:

```
if *doNotDeploySPO {
  t.Skip("Skipping test since SPO is not deployed")
}
```

Not sure if it makes sense or not.

mqasimsarfraz (commented):
I tested it on minikube and AKS is working fine. I agree that removing SPO as a dependency frees resources and simplify things.

Perhaps we should confirm with @blanquicet as well before removing SPO.

mqasimsarfraz (approved):

integration/inspektor-gadget/audit_seccomp_test.go
blanquicet: I don't think it's necessary as using this manual sp installation approach fully tests the audit-seccomp gadget.

integration/inspektor-gadget/integration_test.go
```diff
@@ -51,7 +51,7 @@ var (
 	dnsTesterImage = flag.String("dnstester-image", "ghcr.io/inspektor-gadget/dnstester:latest", "dnstester container image")
 
 	doNotDeployIG  = flag.Bool("no-deploy-ig", false, "don't deploy Inspektor Gadget")
-	doNotDeploySPO = flag.Bool("no-deploy-spo", false, "don't deploy the Security Profiles Operator (SPO)")
+	doNotDeploySPO = flag.Bool("no-deploy-spo", true, "don't deploy the Security Profiles Operator (SPO)")
```
blanquicet, line 54: I agree to keep the code because it will be helpful to test the advise seccomp-profile gadget, see https://github.com/inspektor-gadget/inspektor-gadget/issues/975.

blanquicet (approved):
I didn't test it, but I agree with the suggested changed.

mauriciovasquezbernal:
Merging now, I'll monitor this test to see if there are any issues later on.

Issue #2382, opened 2024-01-18
image-based: Create template repository for image base gadgets


mauriciovasquezbernal:
We should create [a template repository](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository) for creating a new gadget. The repository should include:
- A sample gadget source code (could be taken from https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.24.0/docs/devel/hello-world-gadget.md)
- Configuration to publish the gadget (similar to [existing](https://github.com/inspektor-gadget/inspektor-gadget/blob/122e8b6da0a9e9545e97aefbde94fec1c274c4ee/.github/workflows/inspektor-gadget.yml#L1153-L1193) logic in our github actions)
- Tests (once we sort https://github.com/inspektor-gadget/inspektor-gadget/issues/2046 out)
- Readme (describing how to use the gadget)

alban:
https://github.com/inspektor-gadget/gadget-template

PR #2487, opened 2024-02-11
docs: add gadget-template

alban:
The easiest way to create a new gadget is to start from the template repository. This patch adds a reference to it. We keep the documentation to start a gadget from scratch: it is still useful in case the developer already has a git repository for their project.

https://github.com/inspektor-gadget/gadget-template

https://github.com/inspektor-gadget/inspektor-gadget/issues/2382


blanquicet (approved):
LGTM!

PR #2503, opened 2024-02-14
ci: unpin docker buildx version

mqasimsarfraz:
We were using a specific version for docker buildx to have support for adding annotaions ( https://github.com/docker/buildx/pull/1965). But it has been released into stable already so removing workaround. You can confirm that `docker/setup-buildx-action@v3` uses [`v0.12.0+` by default now.](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/7903715108/job/21572292303#step:4:109)

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -695,9 +695,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-        with:
-          # TODO: Remove once buildx v0.12.0 is released with https://github.com/docker/buildx/pull/1965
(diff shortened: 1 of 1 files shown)


alban (approved):
It looks good!

Since that section of the GitHub action has `if: github.event_name != 'pull_request'`, we will only now if it works once it reaches the `main` branch, right?

mqasimsarfraz:
Thanks for quick review! 

> Since that section of the GitHub action has `if: github.event_name != 'pull_request'`, we will only now if it works once it reaches the `main` branch, right?

Yes but I pushed it to a fork and seem to be [working fine](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/7903642902/job/21572683466) there.

PR #2516, opened 2024-02-19
Fix auto-mount-filesystems

mauriciovasquezbernal:
If there is an error calling statfs on the path, for instance, it doesn't exist, the logic should continue without erroring out to suggest using "--auto-mount-filesystems"

Fixes: 073bcd016837 ("auto-mount-filesystems: add alternative path for tracefs")

## Testing

Create a privileged container without mounting tracefs nor debugfs from the host:

```bash
$ docker run --privileged -it --rm -v $(pwd):/ig ubuntu:22.04 bash
```

### Before this PR
 
```bash 
$ /ig/ig trace exec --host
error: statfs /sys/kernel/debug/tracing: no such file or directory
```

### After this PR 

```bash 
$ /ig/ig trace exec --host
error: filesystems debugfs, tracefs not mounted (did you try --auto-mount-filesystems?)
```

### DIFF
--- a/pkg/utils/host/workarounds.go
+++ b/pkg/utils/host/workarounds.go
@@ -215,7 +215,9 @@ func autoSdUnitRestart() (exit bool, err error) {
 // If dryRun is true, autoMount will only check if the filesystems need to be
 // mounted.
 // Returns the list of filesystems that need to be mounted.
-func autoMountFilesystems(dryRun bool) (mountsSuggested []string, err error) {
+func autoMountFilesystems(dryRun bool) ([]string, error) {
+	var mountsSuggested []string
+
 	fs := []struct {
 		name    string
 		paths   []string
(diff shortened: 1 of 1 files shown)


pkg/utils/host/workarounds.go
```diff
@@ -246,9 +246,9 @@ filesystemLoop:
 	for _, f := range fs {
 		var statfs unix.Statfs_t
 		for _, path := range f.paths {
-			err = unix.Statfs(path, &statfs)
-			if err != nil {
-				return mountsSuggested, fmt.Errorf("statfs %s: %w", path, err)
+			if err := unix.Statfs(path, &statfs); err != nil {
```
blanquicet, line 249: nit: This indeed solves the issue, but I would prefer to avoid using named return values so that we don't end up with this shadowing.

blanquicet (approved):
LGTM.

PR #2517, opened 2024-02-19
Mount filesystems by default when running in a container

mauriciovasquezbernal:
Inspektor Gadget requires some filesystems to be mounted in order to work (bpffs, tracefs and debugfs). a65dbbc6bcfd ("host package: workarounds for kubectl-debug and Windows") introduced a way to mount those filesystems automatically, however the logic wasn't enable by default on ig.

This commit enables that logic by default when running inside a container (HOST_ROOT is passed). This helps users to run ig inside a container without having to worry about passing all the mounts from the host.

After this commit, ig can be run as:

docker run --privileged -it --rm -v /:/host -e HOST_ROOT=/host -v /run:/run ig_image trace exec

The logic is still disabled by default when running in the host because we don't want to change the system configuration without making the user aware.

--- 

@alban  what do you think about this approach?

### DIFF
--- a/docs/examples/pod-ig.yaml
+++ b/docs/examples/pod-ig.yaml
@@ -18,7 +18,6 @@ spec:
       command:
         # CHANGEME: run the gadget of your choice
         - "ig"
-        - "--auto-mount-filesystems"
         - "trace"
         - "exec"
         - "--host"
@@ -30,8 +29,6 @@ spec:
           name: host
         - mountPath: /run
           name: run
-        - mountPath: /sys/kernel/debug
-          name: debugfs
   # CHANGEME: where do you want to run this pod?
   nodeName: minikube-containerd
   volumes:
@@ -41,6 +38,3 @@ spec:
     - name: run
       hostPath:
         path: /run
-    - name: debugfs
-      hostPath:
-        path: /sys/kernel/debug
--- a/pkg/utils/host/host.go
+++ b/pkg/utils/host/host.go
@@ -142,11 +142,17 @@ func AddFlags(command *cobra.Command) {
 		false,
 		"Automatically run in a privileged systemd unit if lacking enough capabilities",
 	)
+
+	// Enable the mount workaround by default when running inside a container.
+	automountFilesystemsDefault := false
+	if HostRoot != "" && HostRoot != "/" {
+		automountFilesystemsDefault = true
+	}
 	command.PersistentFlags().BoolVarP(
 		&autoMountFilesystemsFlag,
 		"auto-mount-filesystems",
 		"",
-		false,
+		automountFilesystemsDefault,
 		"Automatically mount bpffs, debugfs and tracefs if they are not already mounted",
 	)
 	command.PersistentFlags().BoolVarP(
--- a/docs/ig.md
+++ b/docs/ig.md
@@ -308,9 +308,6 @@ $ docker run -ti --rm \
     --privileged \
     -v /run:/run \
     -v /:/host \
-    -v /sys/kernel/debug:/sys/kernel/debug \
-    -v /sys/kernel/tracing:/sys/kernel/tracing \
-    -v /sys/fs/bpf:/sys/fs/bpf \
     --pid=host \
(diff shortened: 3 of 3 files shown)


pkg/utils/host/host.go
```diff
@@ -46,15 +46,6 @@ func init() {
 }
 
 type Config struct {
-	// AutoMountFilesystems will automatically mount bpffs, debugfs and
```
alban: Since the Config struct is now empty, should it be removed from the `func Init()` prototype?

pkg/utils/host/host.go
```diff
@@ -142,11 +131,17 @@ func AddFlags(command *cobra.Command) {
 		false,
 		"Automatically run in a privileged systemd unit if lacking enough capabilities",
 	)
+
+	// Enable the mount workaround by default when running inside a container.
+	automountFilesystemsDefault := false
+	if HostRoot != "" {
```
alban, line 136:
```suggestion
	if HostRoot != "" && HostRoot != "/" {
```

See line 43 for explanations.

pkg/utils/host/host.go
```diff
@@ -142,11 +131,17 @@ func AddFlags(command *cobra.Command) {
...
+		automountFilesystemsDefault = true
+	}
 	command.PersistentFlags().BoolVarP(
 		&autoMountFilesystemsFlag,
 		"auto-mount-filesystems",
 		"",
-		false,
+		automountFilesystemsDefault,
```
alban, line 143: I am wondering if it's ok to change the default value dynamically. Would `--help` display a different default value depending on `$HOST_ROOT`?

What do you think about changing the default value to true unconditionally? I am not sure what's the best, so you can choose either way.

alban (commented):
In general, I like the idea because it makes the `docker run` command line simpler.

To go further, I wonder if the `-e HOST_ROOT=` could be auto-detected somehow in a safe way.

I have some questions in the code but once they are addressed and the CI passes, it should be good to go.

pkg/utils/host/host.go
mauriciovasquezbernal: I prefer to change the default value. I can be confusing for the users to have a flag that doesn't have any effect when `$HOST_ROOT` is defined.

blanquicet: When one runs `ig --help`, it doesn't show the default value of bool variables:

```bash
$ ./ig --help
      [...]
      --auto-mount-filesystems          Automatically mount bpffs, debugfs and tracefs if they are not already mounted
      [...]
```

I suppose it is because it's expected to be always false and only true if I use such a flag. So, we have to manually add it within the description to make it visible.

Another option is to make it string and set it as "auto" by default:

```bash
      --mount-filesystems          Mount bpffs, debugfs and tracefs if they are not already mounted. Valid values are: true, false, auto (default "auto").
```

docs/examples/pod-ig.yaml
```diff
@@ -30,8 +29,6 @@ spec:
           name: host
         - mountPath: /run
           name: run
-        - mountPath: /sys/kernel/debug
```
blanquicet, line 33: Could you please update [this documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/ig.md#using-ig-in-a-container) too?

pkg/utils/host/host.go
mauriciovasquezbernal: Do we really need to change this more? I think just changing the default value is good enough for the purpose of this PR: To make it easier to run ig on containers.

blanquicet: IMO, at least the default value should be present when one runs `ig --help`.

blanquicet (approved):
LGTM!

mauriciovasquezbernal:
It turns out my previous commit was wrong. It was breaking the tests for kubernetes in minikube as the `AddFlags` isn't called from the gadget tracer manager. I changed the code and reintroduced the config structure again to always enable this for the gadget tracer manager.

PR #2498, opened 2024-02-13
docs/devel: Specify that `GADGET_*` macros are optional

blanquicet:
# docs/devel: Specify that `GADGET_*` macros are optional

It wasn't clear that the `GADGET_*` macro is optional for the gadget to work. However, it's necessary to enable IG to generate (from scratch) the metadata file from the eBPF code.

## TODO (For a future PR)

Add a list with the type of gadgets we support: `tracer` and `snapshotter` (probably in `gadget-helper-api.md`) and refer it from the [Implementing the eBPF program](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/hello-world-gadget.md#implementing-the-ebpf-program) section of `hello-world-gadget.md`.

PR #2518, opened 2024-02-19
pkg/run: Don't rely on GADGET_TRACER as it is optional

blanquicet:
# Don't rely on GADGET_TRACER as it is optional

```
GetTracerInfo gets the tracer information from what was defined by
the GADGET_TRACER macro. However, such macro is optional, so we
must not rely on it. Additionally, by the time
handleTracerMapDefinition() is called, the metadata was already
built (either from the metadata file or from the eBPF code), so we
can get the tracer information from there.
```

### DIFF
--- a/pkg/gadgets/run/tracer/tracer.go
+++ b/pkg/gadgets/run/tracer/tracer.go
@@ -274,8 +274,6 @@ func (t *Tracer) handleTracerMapDefinition(tracerMapName string) error {
 			bufMap.KeySize = 4
 			bufMap.ValueSize = 4
 
-			t.spec.Maps[tracerMapName] = bufMap
-
 			if heapMapPresent {
 				_, tracer := getAnyMapElem(t.config.Metadata.Tracers)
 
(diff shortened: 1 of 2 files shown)


blanquicet:
> Hi.
> 
Hola Francis 🙂 

> I would question the whole existence of this macro. If it is optional, we should just drop it and not add another level of complexity. Otherwise, we should document in which case it is mandatory to use it. 
>
The macro `GADGET_TRACER` is only useful to allow IG autogenerating the `tracers` and `structs` section of the metadata file. However, if developer created the metadata manually, the macro is not necessary anymore. That's why it is optional. That's also why the documentation says _"The `tracers` and `structs` sections will only be generated if the eBPF program defined a tracer using the `GADGET_TRACER` macro"_, so ...

> I suspect this documentation to no more be up to date:

... no, it is up to date. Actually, I recently updated it in [this PR](https://github.com/inspektor-gadget/inspektor-gadget/pull/2498) the documentation to make this optionality clearer:

https://github.com/inspektor-gadget/inspektor-gadget/blob/1072df20edbf07eed8269dd896494e4c1bc0afd8/docs/devel/hello-world-gadget.md?plain=1#L86-L92

If you consider it is not yet clear enough, would you mind opening a PR to improve the documentation?

mauriciovasquezbernal:
> If this metadata file is optional, then why do we need it? 

The metadata file allows to define extra attributes for the gadget as the format used to print the fields, and in the future documentation and all other stuff we can't put directly in the eBPF code. It's optional because the gadget can work without it, we don't want to force gadget developers creating this file if they don't want to define any of these extra attributes.

PR #2521, opened 2024-02-20
add btfhelpers package

flyth:
This pkg bundles a couple of helpers around btf type conversion. These versions will now also collect all visited types when going down the typedef tree so that they can later on be used as tags for fields.

Part of #2512

### DIFF
--- a/pkg/btfhelpers/btfhelpers.go
+++ b/pkg/btfhelpers/btfhelpers.go
@@ -1,0 +1,111 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package btfhelpers provides a couple of helper functions to bridge Go's reflection system with
+// types from BTF
+package btfhelpers
+
+import (
+	"reflect"
+
+	"github.com/cilium/ebpf/btf"
+)
+
+// GetType returns the reflect.Type for a given BTF type and the list of type names found while
+// resolving it.
+func GetType(typ btf.Type) (reflect.Type, []string) {
+	var refType reflect.Type
+	typeNames := []string{}
+
+	if typ.TypeName() != "" {
+		typeNames = append(typeNames, typ.TypeName())
+	}
+
+	switch typed := typ.(type) {
+	case *btf.Array:
+		arrType := getSimpleType(typed.Type)
+		if arrType == nil {
+			return nil, nil
+		}
+		if typed.Type.TypeName() != "" {
+			typeNames = append(typeNames, typed.Type.TypeName())
+		}
+		refType = reflect.ArrayOf(int(typed.Nelems), arrType)
+	case *btf.Typedef:
+		switch typed := typ.(type) {
+		case *btf.Typedef:
+			refType, typeNames2 := GetType(typed.Type)
+			typeNames = append(typeNames, typeNames2...)
+			return refType, typeNames
+		default:
+			return GetType(typed)
+		}
+	default:
+		refType = getSimpleType(typ)
+	}
+
+	return refType, typeNames
+}
+
+// GetUnderlyingType returns the underlying type of a typedef
+func GetUnderlyingType(tf *btf.Typedef) btf.Type {
+	switch typed := tf.Type.(type) {
+	case *btf.Typedef:
+		return GetUnderlyingType(typed)
+	default:
+		return typed
+	}
+}
+
+func getSimpleType(typ btf.Type) reflect.Type {
+	switch typed := typ.(type) {
+	case *btf.Int:
+		switch typed.Encoding {
+		case btf.Signed:
+			switch typed.Size {
+			case 1:
+				return reflect.TypeOf(int8(0))
+			case 2:
+				return reflect.TypeOf(int16(0))
+			case 4:
+				return reflect.TypeOf(int32(0))
+			case 8:
+				return reflect.TypeOf(int64(0))
+			}
+		case btf.Unsigned:
+			switch typed.Size {
+			case 1:
+				return reflect.TypeOf(uint8(0))
+			case 2:
+				return reflect.TypeOf(uint16(0))
+			case 4:
+				return reflect.TypeOf(uint32(0))
+			case 8:
+				return reflect.TypeOf(uint64(0))
+			}
+		case btf.Bool:
+			return reflect.TypeOf(false)
+		case btf.Char:
+			return reflect.TypeOf(uint8(0))
+		}
+	case *btf.Float:
+		switch typed.Size {
+		case 4:
+			return reflect.TypeOf(float32(0))
+		case 8:
+			return reflect.TypeOf(float64(0))
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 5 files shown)


pkg/btfhelpers/btfhelpers.go
```diff
@@ -0,0 +1,99 @@
...
+			return nil, nil
+		}
+		return reflect.ArrayOf(int(typedMember.Nelems), arrType), append(typeNames, typ.TypeName())
+	default:
+		arrType, typeNames := getSimpleType(typ)
+		return arrType, append(typeNames, typ.TypeName())
+	}
+}
+
+func GetUnderlyingType(tf *btf.Typedef) (btf.Type, []string, error) {
```
alban, line 39: Can you document the return values?

It seems it can never return an error. Could you remove `error`?

pkg/btfhelpers/btfhelpers.go
```diff
@@ -0,0 +1,99 @@
...
+// types from BTF
+package btfhelpers
+
+import (
+	"reflect"
+
+	"github.com/cilium/ebpf/btf"
+)
+
+func GetType(typ btf.Type) (reflect.Type, []string) {
```
alban, line 25: Can you document the return values?

pkg/gadgets/run/tracer/run.go should make use of it instead of `typeFromBTF`.

alban (commented):
Could you make use of it in pkg/gadgets/run/tracer/run.go and pkg/gadgets/run/types/metadata.go?

Ideally, this PR would delete more lines than it would add thanks to the factoring.

This could have some unit tests.

flyth:
> Could you make use of it in pkg/gadgets/run/tracer/run.go and pkg/gadgets/run/types/metadata.go?

In #2512 it is suggested to remove the run gadget (pkg/gadgets/run) in that PR - as this is only a precursor for that PR, I don't really see a point in updating that here - WDYT?

> This could have some unit tests.

Hmm, what would you suggest? Adding and parsing a real binary or mocking a btf.Type? The former seems to be a better idea but comes with quite a bit of overhead - and the integration tests would probably catch errors as well. The latter sounds rather hacky - need to look more into this.

pkg/btfhelpers/btfhelpers.go
mauriciovasquezbernal, line 39: I see this only resolves typedefs. Should it support resolving const, volatile, pointers, etc?

If not, I think it should be named `ResolveTypedef` to make it clear. 

(yes, I was the one introducing the function with that name originally)

pkg/btfhelpers/btfhelpers.go
```diff
@@ -0,0 +1,99 @@
...
+	switch typedMember := typ.(type) {
+	case *btf.Array:
+		arrType, typeNames := getSimpleType(typedMember.Type)
+		if arrType == nil {
+			return nil, nil
+		}
+		return reflect.ArrayOf(int(typedMember.Nelems), arrType), append(typeNames, typ.TypeName())
+	default:
+		arrType, typeNames := getSimpleType(typ)
+		return arrType, append(typeNames, typ.TypeName())
```
mauriciovasquezbernal, line 35: Why not just `return getSimpleType(typ)`?

pkg/btfhelpers/btfhelpers.go
```diff
@@ -0,0 +1,99 @@
...
+	case *btf.Typedef:
+		typ, typeNames, _ := GetUnderlyingType(typedMember)
+		refType, typeNames2 := getSimpleType(typ)
+		return refType, append(append(typeNames, typ.TypeName()), typeNames2...)
+	case *btf.Array: // TODO: this has to be supported here
+		arrType, typeNames := getSimpleType(typedMember.Type)
+		if arrType == nil {
+			return nil, nil
+		}
+		return reflect.ArrayOf(int(typedMember.Nelems), arrType), append(typeNames, typ.TypeName())
```
mauriciovasquezbernal, line 96: It should be here, or in GetType, but not in both places.

mauriciovasquezbernal (commented):
I implemented (i.e. copilot) some unit tests in https://github.com/inspektor-gadget/inspektor-gadget/commit/f9077f6bf736afc043ddb3971dfd1e34a3ad403c. Some are still failing, but I think they're good to discuss the returned values.

Do we need to support things like const, volatile, pointers, arrays of arrays and so on? 

Also, the original code in pkg/gadgets/run/tracer/run.go supported enums, aren't they needed here?

pkg/btfhelpers/btfhelpers.go
flyth: it appends the `typ.TypeName()` to the list as well

pkg/btfhelpers/btfhelpers.go
flyth: That's handled outside of this function for now - need to think more about whether we could need it at this level already.

pkg/btfhelpers/btfhelpers.go
mauriciovasquezbernal: `getSimpleType` does the same. So there are two instances of the type name in the return error. Is that expected?

mauriciovasquezbernal: Also, I see `getSimpleType` returns `typ.TypeName()` for almost all types, so perhaps we can remove it from there and handle it here?

mauriciovasquezbernal:
@flyth what do you think of an alternative implementation as done in https://github.com/inspektor-gadget/inspektor-gadget/commit/5d2c278af96c68e23ad647f4f2ab08d109949aa3? I did some quick tests on last version of the refactoring and it appears to be working fine.

flyth:
> @flyth what do you think of an alternative implementation as done in [5d2c278](https://github.com/inspektor-gadget/inspektor-gadget/commit/5d2c278af96c68e23ad647f4f2ab08d109949aa3)? I did some quick tests on last version of the refactoring and it appears to be working fine.

I'll try soon - at first glance it looks like it doesn't recurse, though, right? It seems to take a maximum of two type names.

mauriciovasquezbernal:
> I'll try soon - at first glance it looks like it doesn't recurse, though, right? It seems to take a maximum of two type names.

Ah right. So I think `GetUnderlyingType` should continue to return the type names. (Or implement this logic in GetType without relying on GetUnderlyingType)

mauriciovasquezbernal:
Updated it to recurse in https://github.com/inspektor-gadget/inspektor-gadget/commit/618d18cf0047c8babe7d52623ebe57cd19797228.

mauriciovasquezbernal:
> Updated it to recurse in [618d18c](https://github.com/inspektor-gadget/inspektor-gadget/commit/618d18cf0047c8babe7d52623ebe57cd19797228).

Seems like I forgot to attach the updated version of the tests. Done in https://github.com/inspektor-gadget/inspektor-gadget/commit/c6152baea85b1b81bc95ddc1e1077e0455028399

flyth:
> > Updated it to recurse in [618d18c](https://github.com/inspektor-gadget/inspektor-gadget/commit/618d18cf0047c8babe7d52623ebe57cd19797228).
> 
> Seems like I forgot to attach the updated version of the tests. Done in [c6152ba](https://github.com/inspektor-gadget/inspektor-gadget/commit/c6152baea85b1b81bc95ddc1e1077e0455028399)

Added the additional tests, thank you very much!

pkg/btfhelpers/btfhelpers.go
```diff
@@ -0,0 +1,111 @@
...
+	typeNames := []string{}
+
+	if typ.TypeName() != "" {
+		typeNames = append(typeNames, typ.TypeName())
+	}
+
+	switch typed := typ.(type) {
+	case *btf.Array:
+		arrType := getSimpleType(typed.Type)
+		if arrType == nil {
```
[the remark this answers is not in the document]
flyth: Arrays of non-numeric types (like other structs, etc.), I guess. Those wouldn't be supported for now.

pkg/btfhelpers/btfhelpers.go
flyth: Resolving your comments because you addressed them yourself :) Thanks for that!

flyth:
> Hi!
> 
> Can you use the functions you added in this commit to remove `reflectTypeFromType()` and associated functions?
> 
> Best regards.

I've used the `GetUnderlyingType()` from the new package, however, the rest of the functions are no easy replacement, as in the run gadget an interim type is used (`types.Type`), whereas the new lib only uses `reflect.Type`.

mauriciovasquezbernal (approved):
LGTM!

Issue #2175, opened 2023-10-23
[RFE] artifacthub: Add Gadget kind


mqasimsarfraz:
## Current situation

Currently all the gadgets are published as [packages](https://github.com/orgs/inspektor-gadget/packages?repo_name=inspektor-gadget) in container registry. In order to improve the discovery of the gadgets, we can publish them as a kind in [artifacthub](https://artifacthub.io/packages/search). We can also use the "Official" and "CNCF" badges to mark the gadgets maintained by us. 

## Impact

Improves the discovery of gadgets

## Implementation options

The [comment](https://github.com/artifacthub/hub/issues/2271#issuecomment-1235378557) gives an idea about the implementation, we need check if we need a [special tracker](https://github.com/artifacthub/hub/tree/master/internal/tracker/source) or a generic tracker can do a trick for us. A good starting point is opening an [issue](https://github.com/artifacthub/hub) to start conversion.  Some exmaples:

- https://github.com/artifacthub/hub/issues/2271
- https://github.com/artifacthub/hub/pull/2902
- https://github.com/artifacthub/hub/pull/514

alban:
Thanks for filing this issue. I agree it would be good to have a gadget kind.

---

Meanwhile, since artifacthub does not have a gadget kind yet, I was experimenting on [runc-vuln-gadget](https://github.com/alban/runc-vuln-gadget/) with the [container image kind](https://artifacthub.io/docs/topics/repositories/container-images/).

Steps:
* Ensure the gadget is public on the OCI registry (on GitHub, click on "Package settings")
* Ensure the OCI image index has the correct annotations on the **latest** tag. Example of Makefile target:
```Makefile
add-annotations:
        export VCS_DATE="$(shell date -d "@$(shell git log -1 --format=%at)" +%Y-%m-%dT%H:%M:%SZ --utc)" ; \
        export VCS_SOURCE="https://github.com/alban/runc-vuln-gadget/" ; \
        export VCS_SHA="$(shell git rev-list -1 HEAD)" ; \
        regctl image mod \
                $(CONTAINER_REPO):latest \
                --replace \
                --to-oci-referrers \
                --time-max "$${VCS_DATE}" \
                --annotation "org.opencontainers.image.description=A gadget to detect runc vulnerabilities" \
                --annotation "org.opencontainers.image.created=$${VCS_DATE}" \
                --annotation "oci.opencontainers.image.source=$${VCS_SOURCE}" \
[...]

PR #2522, opened 2024-02-21
image build: add annotations in OCI image

alban:
OCI images can have annotations at different levels:
- index
- manifest
- config
- layer

This patch adds annotations in the index, manifest and config levels.

artifacthub reads annotations at the manifest level. The 3 required annotations from artifacthub are defined:
- io.artifacthub.package.readme-url
- org.opencontainers.image.created
- org.opencontainers.image.description

Users can add custom images with the `annotations` field in gadget.yaml.

https://github.com/inspektor-gadget/inspektor-gadget/issues/2175

### DIFF
--- a/docs/reference/oci.md
+++ b/docs/reference/oci.md
@@ -49,6 +49,19 @@ Its content must be a valid ELF file.
 There must be at most one layer with the wasm media type. If present, it must
 not be empty and it must be a valid wasm file.
 
-## Image labels
+## Image annotations
 
-TODO
+OCI images can have annotations at different levels:
+- index
+- manifest
+- config
+- layer
+
+Inspektor Gadget automatically adds the following annotations at the manifest and config levels:
+- `org.opencontainers.image.*`: defined by [OCI Image Format](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys)
+  - title
+  - description
+  - url
+  - documentation
+  - source
+  - created
--- a/pkg/oci/build.go
+++ b/pkg/oci/build.go
@@ -22,13 +22,17 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"time"
 
 	"github.com/distribution/reference"
 	"github.com/opencontainers/image-spec/specs-go"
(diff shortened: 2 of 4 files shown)


docs/reference/oci.md
```diff
@@ -49,6 +49,14 @@ Its content must be a valid ELF file.
...
 
-TODO
+OCI images can have annotations at different levels:
+- index
+- manifest
+- config
+- layer
+
+Inspektor Gadget adds the following annotations at the manifest and config levels:
+- `io.artifacthub.package.readme-url`: defined by [ArtifactHUB](https://artifacthub.io/docs/topics/repositories/container-images/#image-metadata)
```
alban, line 61: Following the discussion on https://github.com/artifacthub/hub/issues/3665 (https://github.com/inspektor-gadget/inspektor-gadget/issues/2175):
New kinds of artifacts are encouraged to use the file [artifacthub-pkg.yml](https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml) file. So gadget authors would publish their gadgets on Artifact Hub by adding a this file in their git repository. Artifact Hub would not fetch the annotations from the OCI image.

So I will remove this artifacthub annotation for now.

Note that users can always add custom OCI annotations in the `annotations` field of `gadget.yaml`.

pkg/oci/build.go
```diff
@@ -136,15 +136,42 @@ func createLayerDesc(ctx context.Context, target oras.Target, progFilePath, medi
...
+		ocispec.AnnotationURL:           metadata.HomepageURL,
+		ocispec.AnnotationDocumentation: metadata.DocumentationURL,
+		ocispec.AnnotationSource:        metadata.SourceURL,
+	}
+
+	if metadata.Annotations != nil {
+		for k, v := range metadata.Annotations {
+			annotations[k] = v
+		}
+	}
```
mauriciovasquezbernal, line 160: You can avoid the nil check as range over nil is no-op in golang.

mauriciovasquezbernal (approved):
I haven't gone deep into the details but LGTM from code inspection.

PR #2528, opened 2024-02-22
docs: Update image build documentation

blanquicet:
# Update image build documentation

This PR makes two changes to the build documentation:
1. Update fields that can be specified in the `build.yaml` file.
2. Add reference to the build customization in the hello-word-gadget file.

### DIFF
--- a/docs/devel/hello-world-gadget.md
+++ b/docs/devel/hello-world-gadget.md
@@ -183,6 +183,10 @@ INFO[0000] Experimental features enabled
 Successfully built docker.io/library/mygadget:latest@sha256:dd3f5c357983bb863ef86942e36f4c851933eec4b32ba65ee375acb1c514f628
 ```
 
+Take into account that it is possible to customize the build process by defining a `build.yaml` file.
+Check the [Customizing your build](../../docs/core-concepts/images.md#customizing-your-build)
+section for more details.
+
 ## (Optional) Pushing the gadget image to a container registry
 
 You could push the gadget to a remote container registry. If you're using the same machine for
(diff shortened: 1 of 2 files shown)


docs/devel/hello-world-gadget.md
```diff
@@ -183,6 +183,29 @@ INFO[0000] Experimental features enabled
...
 
+### Customizing the build
+
+You can customize the build process by providing a `build.yaml` file. This file can be passed to the
+`image build` command using the `-f` flag. The `build.yaml` file can contain the following fields:
+
+- `ebpfsource`: The eBPF program file (default: "program.bpf.c")
+- `metadata`: The metadata file (default: "gadget.yaml")
+- `cflags`: The C flags used to compile the eBPF program (default: "")
+- `wasm`: The WASM file (default: "")
```
alban, line 194: This is already documented in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/core-concepts/images.md#customizing-your-build

I wonder if we should have a separate page building-images.md somewhere and reference it from both `devel/hello-world-gadget.md` and `core-concepts/images.md`.

blanquicet: > This is already documented in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/core-concepts/images.md#customizing-your-build
> 
🤦🏽 I didn't find it because I looked for the "cflags" string.

> I wonder if we should have a separate page building-images.md somewhere and reference it from both `devel/hello-world-gadget.md` and `core-concepts/images.md`.
>
I don't think the build information is big enough to require a separated document. I'll update this PR to:

- Include cflags in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/core-concepts/images.md#customizing-your-build
- Update `devel/hello-world-gadget.md` to point developer to `core-concepts/images.md` for build customization.

mauriciovasquezbernal (approved):
LGTM!

PR #2531, opened 2024-02-22
Use BTF to check block_rq_insert arguments

eiffel-fl:
Hi.

This PR uses BTF rather than `KERNEL_VERSION` to check the number of arguments of `block_rq_insert` tracepoint.

Best regards.

### DIFF
--- a/integration/inspektor-gadget/profile_blockio_test.go
+++ b/integration/inspektor-gadget/profile_blockio_test.go
@@ -26,10 +26,6 @@ import (
 func TestProfileBlockIO(t *testing.T) {
 	t.Parallel()
 
-	if *k8sDistro == K8sDistroEKSAmazonLinux {
-		t.Skip("profile block-io is not supported on Amazon Linux: See issue #2468")
-	}
-
 	commands := []*Command{
 		{
 			Name: "RunProfileBlockIOGadget",


pkg/gadgets/profile/block-io/tracer/tracer.go
```diff
@@ -93,12 +94,53 @@ func (t *Tracer) close() {
...
+
+	ptr, ok := typedef.Type.(*btf.Pointer)
+	if !ok {
+		return false, fmt.Errorf("expecting btf.Enum, got: %v", typedef.Type)
+	}
+
+	funcProto, ok := ptr.Target.(*btf.FuncProto)
+	if !ok {
+		return false, fmt.Errorf("expecting btf.FuncProto, got: %v", ptr.Target)
+	}
```
alban, line 119: Instead of checking in our Go code with `btfSpec.TypeByName`, we could test in the ebpf code with `bpf_core_type_matches`?

That would make it easier for the conversion to image-based gadget.

We need to update to the new `bpf_core_read.h` with `BPF_TYPE_MATCHES` and `bpf_core_type_matches`:
https://github.com/libbpf/libbpf/blob/v1.3.0/src/bpf_core_read.h#L194

Type match relocations was added in llvm/clang 15:
https://github.com/llvm/llvm-project/commit/d129ac27e872e189363e3d8cd6bc4e2152cd73ec

Example of usage:
https://github.com/torvalds/linux/blob/v6.6/tools/perf/util/bpf_skel/lock_contention.bpf.c#L215

This could look like this:

```
// 
// https://github.com/torvalds/linux/commit/a54895fa057c67700270777f7661d8d3c7fda88a
// -       TP_PROTO(struct request_queue *q, struct request *rq),
// +       TP_PROTO(struct request *rq),

typedef void (*btf_trace_block_rq_insert___new)(void *, struct request *);
typedef void (*btf_trace_block_rq_insert___old)(void *, struct request_queue *q, struct request *);

...

if (bpf_core_type_match(btf_trace_block_rq_issue___new)) {
  // After commit a54895fa (v5.11-rc1)
  return trace_rq_start((void *)ctx[0], true);
} else if (bpf_core_type_match(btf_trace_block_rq_issue___old)) {
  // Before commit a54895fa (v5.11-rc1)
  return trace_rq_start((void *)ctx[1], false);
} else {
  // Couldn't detect block/block_rq_insert tracepoint
  return 0;
}
```

However, cilium/ebpf does not yet support relocations of type `BPF_CORE_TYPE_MATCHES`
https://docs.kernel.org/bpf/llvm_reloc.html
```
enum bpf_core_relo_kind {
...
       BPF_CORE_ENUMVAL_VALUE     = 11, /* enum value integer value */
       BPF_CORE_TYPE_MATCHES      = 12, /* type match in target kernel */
};
```
It stops at `BPF_CORE_ENUMVAL_VALUE` / `reloEnumvalValue`:
https://github.com/cilium/ebpf/blob/v0.13.0/btf/core.go#L121

So we would need to implement it in cilium/ebpf first.

pkg/gadgets/profile/block-io/tracer/bpf/biolatency.bpf.c
```diff
@@ -15,11 +15,10 @@ const volatile bool targ_per_disk = false;
 const volatile bool targ_per_flag = false;
 const volatile bool targ_queued = false;
 const volatile bool targ_ms = false;
+const volatile bool targ_single = true;
```
alban, line 18: There are 2 tracepoints:
* btf_trace_block_rq_insert
* btf_trace_block_rq_issue

But you test only one of them. In practice it works because https://github.com/torvalds/linux/commit/a54895fa057c67700270777f7661d8d3c7fda88a changed both at the same time. But it would be cleaner to test both, define two `const volatile` for each tracepoint,and use `RewriteConstants` on both.

Or at least add a comment before `targ_single` to explain its semantics and list the tracepoints.

alban (approved):
LGTM from code inspection.
- One suggestion to improve code readability about the constant naming.
- Suggestion for improvement in cilium/ebpf (it's not so easy, so not to do in this PR, but later)

pkg/gadgets/profile/block-io/tracer/tracer.go
```diff
@@ -93,12 +94,53 @@ func (t *Tracer) close() {
...
+		return false, fmt.Errorf("loading kernel BTF information: %w", err)
+	}
+
+	if err := btfSpec.TypeByName("btf_trace_block_rq_issue", &typedef); err != nil {
+		return false, fmt.Errorf("searching btf_trace_block_rq_issue: %w", err)
+	}
+
+	ptr, ok := typedef.Type.(*btf.Pointer)
+	if !ok {
+		return false, fmt.Errorf("expecting btf.Enum, got: %v", typedef.Type)
```
mauriciovasquezbernal, line 113: btf.Enum -> btf.Pointer?

mauriciovasquezbernal (commented):
One minor nit. Agree with Alban idea of using `bpf_core_type_matches` (in a future PR).

pkg/gadgets/profile/block-io/tracer/tracer.go
mauriciovasquezbernal: > One big problem with this approach is that it will just silently failed.

Instead of `return 0` we could use a similar approach to "bad relo" (https://nakryiko.com/posts/bpf-core-reference-guide/#guarding-potentially-failing-relocations) by libbpf / cilium-ebpf to indicate an error.

> I would rather go toward having only one version for the most recent kernel as image based one.
> If people need to run it on dinosaur kernels, they can bake their own image, it is the whole point of image based gadgets.

I don't think it's right. We'd like a gadget to work on as many kernels as possible. Having the image-based version to only work in the most recent kernel is a big limitation we can't tolerate. I don't think the "users can bake their own image" is a valid argument as most users only want to run the gadget without getting dirt with ebpf development and so on. 

Btw, I won't call kernels < 5.11 "dinosaur kernels", 5.4 and 5.10 EOL is December 2026.

alban: > Instead of return 0 we could use a similar approach to "bad relo" (https://nakryiko.com/posts/bpf-core-reference-guide/#guarding-potentially-failing-relocations) by libbpf / cilium-ebpf to indicate an error.

Good idea!

I tried it and it works:
```
static void (*bpf_unreachable)(void) = (void *) 0xbad2310;
...
bpf_unreachable();
```

It causes the ebpf verifier to reject the program:
```
	; bpf_unreachable();
	142: (85) call unknown#195896080
	invalid func unknown#195896080
```

alban: Draft PR: https://github.com/inspektor-gadget/inspektor-gadget/pull/2542 (blocked on cilium/ebpf)

PR #2536, opened 2024-02-23
run: Print fields on same order as defined in the metadata

mauriciovasquezbernal:
In some cases it's not possible to order the fields in the event structure in a human-friendly way (padding and other reasons), hence it's possible to change the order of those fields in the metadata file. This commit updates the logic to use the same order defined in such file instaed of relying on the event structure layout.

### Before this PR

Fields are printed in the same order of the struct event

https://github.com/inspektor-gadget/inspektor-gadget/blob/432375356fac33c3ea32031f3adf06c827addb2f/gadgets/trace_open/program.bpf.c#L24-L36

```bash 
$ sudo -E ig run trace_open 
RUNTIME.CONTAINERNAME                PID                 UID                 GID                 RET    COMM             FNAME 
```

### After this PR 

Fields are printed in the same order as defined in the metadata

https://github.com/inspektor-gadget/inspektor-gadget/blob/0ac7b3d565b6f5775cccbe3595c1eea5a6b91871/gadgets/trace_open/gadget.yaml#L9-L46

```bash 
$ sudo -E ig run trace_open 
RUNTIME.CONTAINERNAME                PID                 COMM             UID                 GID                 RET    FNAME               
```

This code will be changed by #2512, but since it's a small change perhaps we can merge it first.

### DIFF
--- a/pkg/gadgets/run/tracer/run.go
+++ b/pkg/gadgets/run/tracer/run.go
@@ -792,28 +792,24 @@ func calculateColumnsForClient(
 		return nil, fmt.Errorf("getting value struct: %w", err)
 	}
 
-	colNames := map[string]struct{}{}
-
 	eventStruct, ok := gadgetMetadata.Structs[eventType.Name]
 	if !ok {
 		return nil, fmt.Errorf("struct %s not found in gadget metadata", eventType.Name)
 	}
 
-	for _, field := range eventStruct.Fields {
-		colNames[field.Name] = struct{}{}
+	members := map[string]btf.Member{}
+	for _, member := range eventType.Members {
+		members[member.Name] = member
 	}
 
 	columns := []types.ColumnDesc{}
 
-	for _, member := range eventType.Members {
-		member := member
-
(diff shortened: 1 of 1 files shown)


pkg/gadgets/run/tracer/run.go
```diff
@@ -792,28 +792,20 @@ func calculateColumnsForClient(
...
-		member := member
-
-		_, ok := colNames[member.Name]
-		if !ok {
-			logger.Debugf("field %q not present on metadata file, skipping", member.Name)
-			continue
-		}
-
+	for _, field := range eventStruct.Fields {
+		member := members[field.Name]
```
flyth, line 808: Can it happen that the field from the metadata doesn't exist as member of the btf type?

mauriciovasquezbernal: Good point. Yes, it could happen if both the gadget is built and run with `--validate-metadata=false`. I added an additional check for it.

flyth (approved):
LGTM

PR #2544, opened 2024-02-26
dns: Parse packet in user space

### LINKED ISSUE #2008: Dns Tracer - Uncompressed Messages Support

mauriciovasquezbernal:
dns: Parse DNS packet in user space
    
    The current logic tries to parse the full DNS packet in eBPF. This is
    diffcult to do a imposes some limitations, like not supporting:
    - DNS over TCP
    - Uncompressed messages
    - More than 8 answers
    
    This commit changes approach to parce the DNS packet on user space.

TODO: 
- [x] Allow configuring ports

Fixes #2008


flyth (commented):
This makes handling the packets much easier - excited to see how we solve that for the image based gadgets 😄

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -114,159 +102,136 @@ struct {
...
-	event->pkt_type = skb->pkt_type;
+	if (!is_dns_port(event.sport) && !is_dns_port(event.dport))
+		return 0;
 
-	// Read QTYPE right after the QNAME (name_len + the zero length octet)
-	// https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.2
-	event->qtype =
-		load_half(skb, DNS_OFF + sizeof(struct dnshdr) + name_len + 1);
+	event.netns = skb->cb[0]; // cb[0] initialized by dispatcher.bpf.c
+	event.timestamp = bpf_ktime_get_ns();
```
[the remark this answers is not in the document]
mauriciovasquezbernal: ohman, I'm trying to keep this PR with two independent commits but it's being hard. Thanks for catching this up!

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -114,159 +102,136 @@ struct {
...
-	bpf_skb_load_bytes(skb, DNS_OFF + sizeof(struct dnshdr), event->name,
-			   name_len);
+	event.netns = skb->cb[0]; // cb[0] initialized by dispatcher.bpf.c
+	event.timestamp = bpf_ktime_get_boot_ns();
+	event.proto = proto;
+	event.dns_off = dns_off;
 
-	event->pkt_type = skb->pkt_type;
+	if (!is_dns_port(event.sport) && !is_dns_port(event.dport))
+		return 0;
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Makes sense, updated the code. The only think I didn't change was the allocation of the event structure, it's being allocated on the stack, so I don't think "allocating" it later has a visible effect on performance.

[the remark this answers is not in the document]
flyth: I _think_ you can at least prevent the zeroing operation by using `struct event_t event;` and zero it manually later on, rather than using `struct event_t event = {};`, which implies it, right?

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -114,159 +104,63 @@ struct {
...
-					    union dnsflags flags,
-					    __u32 name_len, __u16 ancount)
-{
-	__u32 zero = 0;
-	struct event_t *event = bpf_map_lookup_elem(&tmp_event, &zero);
-	if (!event)
+	sport = load_half(skb, ETH_HLEN + sizeof(struct iphdr) +
+				       offsetof(struct udphdr, source));
+	dport = load_half(skb, ETH_HLEN + sizeof(struct iphdr) +
+				       offsetof(struct udphdr, dest));
```
blanquicet: I know this is the way we were doing it and you changed it in next commit but, I was wondering if it's safe to do this knowing that the IPv4 header doesn't have a fixed size. Shouldn't we always compute the layer 4 offset using the IHL field as you did in the next commit?

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -284,71 +178,40 @@ static __always_inline int output_dns_event(struct __sk_buff *skb,
...
 			}
 		}
 	}
 
-	// size of full structure - addresses + only used addresses
-	unsigned long long size =
-		sizeof(*event) - MAX_ADDR_ANSWERS * 16 + anaddrcount * 16;
-	bpf_perf_event_output(skb, &events, BPF_F_CURRENT_CPU, event, size);
+	__u64 skb_len = skb->len;
+	bpf_perf_event_output(skb, &events, skb_len << 32 | BPF_F_CURRENT_CPU,
```
blanquicet, line 289: Could you please describe why you passed from `BPF_F_CURRENT_CPU` to `skb_len << 32 | BPF_F_CURRENT_CPU`?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
...
 
-func (t *Tracer) run(ctx context.Context, logger logger.Logger, dnsTimeout time.Duration) error {
-	spec, err := loadDns()
+func (t *Tracer) parseDNSPacket(rawSample []byte, netns uint64) (*types.Event, error) {
+	// The sample received is a concatenation of the dnsEventT structure and the packet bytes.
+	bpfEvent := (*dnsEventT)(unsafe.Pointer(&rawSample[0]))
+	packetBytes := rawSample[unsafe.Sizeof(*bpfEvent):]
+
+	// The DNS layer starts after the ethernet, IP and UDP headers.
+	DNS_OFF := 14 + 20 + 8
```
blanquicet: Some comments for this line?

- dnsOffset would me more go-style?
- Isn't there defines like `ETH_HLEN` in gopacket pkg?
- Is it safe to consider the IP header with a fixed size (20)?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
...
+		event.Nameserver = event.SrcIP
+
+		event.Latency = time.Duration(bpfEvent.LatencyNs)
+		event.Rcode = rcodeToString(uint8(dnsLayer.ResponseCode))
+	} else {
+		event.Qr = types.DNSPktTypeQuery
+		event.Nameserver = event.DstIP
+	}
+
+	if len(dnsLayer.Questions) > 0 {
```
blanquicet, line 183: nit: But probably `len(dnsLayer.Questions)` is more readable:

```suggestion
	if dnsLayer.QDCount > 0 {
```

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
...
+		question := dnsLayer.Questions[0]
+		event.QType = question.Type.String()
+		event.DNSName = string(question.Name) + "."
+	}
```
blanquicet, line 187: We could execute this only if it's a query.

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
...
+
+	for _, answer := range dnsLayer.Answers {
```
blanquicet, line 189: Same this only for responses.

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
...
+		ID:         fmt.Sprintf("%.4x", dnsLayer.ID),
+		NumAnswers: int(dnsLayer.ANCount),
+	}
+
+	if dnsLayer.QR {
+		event.Qr = types.DNSPktTypeResponse
+		event.Nameserver = event.SrcIP
+
+		event.Latency = time.Duration(bpfEvent.LatencyNs)
+		event.Rcode = rcodeToString(uint8(dnsLayer.ResponseCode))
```
blanquicet, line 177:
```suggestion
		event.Rcode = dnsLayer.ResponseCode.String()
```
?

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
 	return nil
 }
 
-func (t *Tracer) run(ctx context.Context, logger logger.Logger, dnsTimeout time.Duration) error {
-	spec, err := loadDns()
+func (t *Tracer) parseDNSPacket(rawSample []byte, netns uint64) (*types.Event, error) {
+	// The sample received is a concatenation of the dnsEventT structure and the packet bytes.
```
blanquicet, line 136: I have to be honest. I didn't understand how we are sending the skb from the eBPF code 😅.

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -332,30 +132,80 @@ func (t *Tracer) install() error {
...
+	bpfEvent := (*dnsEventT)(unsafe.Pointer(&rawSample[0]))
+	packetBytes := rawSample[unsafe.Sizeof(*bpfEvent):]
```
blanquicet, line 138: We don't need to send the whole packet as we are only using the DNS layer. BTW, we already computed the IP header len in the eBPF code, so we don't need to re-compute the offsets here.

blanquicet (commented):
I only reviewed the first commit for now. I already have some comments and many questions 😅. Sorry if I commented things that are changed in the second commit, feel free to immediately resolve them.

anubhabMajumdar:
@mauriciovasquezbernal  Thanks for the change.

I tested this change in the following way:
- `make build-gadgets`
- `make minikube-start && make minikube-deploy`
- Exec into minikube and run `nslookup -type=a dc.services.visualstudio.com.` and `nslookup -type=aaaa dc.services.visualstudio.com.`

```bash
 ./kubectl-gadget trace dns -A                                                                               ✔
K8S.NODE                     K8S.NAMESPACE                K8S.POD                      PID        TID        COMM            QR TYPE      QTYPE      NAME                         RCODE          NUMANS…
minikube-docker                                                                        97501      57694      dockerd         Q  OUTGOING  A          dc.services.visualstudio.co…                0
ERRO[0004] decoding dns layer: resource record length exceeds data
minikube-docker                                                                        0          0                          R  OUTGOING  A          dc.services.visualstudio.co… NoError        23
minikube-docker                                                                        71848      71849      isc-worker0000  R  HOST      A          dc.services.visualstudio.co… NoError        23
minikube-docker                                                                        97501      57693      dockerd         Q  OUTGOING  AAAA       dc.services.visualstudio.co…                0
minikube-docker                                                                        97501      57693      dockerd         R  HOST      AAAA       dc.services.visualstudio.co… NoError        13
minikube-docker                                                                        0          0                          R  OUTGOING  AAAA       dc.services.visualstudio.co… NoError        13
minikube-docker                                                                        71910      71911      isc-worker0000  R  HOST      AAAA       dc.services.visualstudio.co… NoError        13
```

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -35,11 +29,26 @@
 #define DNS_QR_QUERY 0
 #define DNS_QR_RESP 1
 
+#define MAX_PORTS 16
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Something big enough to support all ports users could specify but small enough not to waste a lot of memory. I suppose nobody will need to specify more than 16 ports in this gadget. Another option is to use a hash map, but seems an overkill to me.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -35,11 +29,26 @@
...
+const volatile __u16 ports[MAX_PORTS] = { 53, 5353 };
+const volatile __u16 ports_len = 2;
```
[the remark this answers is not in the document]
mauriciovasquezbernal: It'll increase the complexity of the program (additional checks), so usually this is passed as constant too. (We'll need to think how to support this on the image base version later on)

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -81,30 +90,11 @@ struct dnshdr {
...
-	__uint(max_entries, 1);
-	__type(key, __u32);
-	__type(value, struct event_t);
-} tmp_event SEC(".maps");
-
 // Map of DNS query to timestamp so we can calculate latency from query sent to answer received.
 struct query_key_t {
 	__u64 pid_tgid;
 	__u16 id;
+	__u16 pad[3]; // this is needed, otherwise the verifier claims an invalid read from stack
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I have no idea to be honest and it took me a while to understand what was going on.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
mauriciovasquezbernal: As you mentioned, it's fixed by next commit. I wanted to keep the first commit as atomic as possible by only changing the things strictly needed. I'm still wondering if it's better to squash everything on a single commit and avoid the pain of having to divide it into two ones.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
mauriciovasquezbernal: This is the way to submit the skb and custom data (the struct). It doesn't seem to be documented in https://man7.org/linux/man-pages/man7/bpf-helpers.7.html, I had to check the implication of [perf_submit_skb()](https://github.com/iovisor/bcc/blob/master/docs/reference_guide.md#4-perf_submit_skb) to check how it worked.

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: It's all removed in next commit, I didn't want to expend a lot of time in something that will be removed one commit later. Again, I'll happy to squash both commits if preferred.

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: I suppose it doesn't change anything from a performance point of view, so I'll keep this way to avoid adding extra logic.

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: Same as above.

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: The string definitions of both places aren't exactly the same:

```golang
// DNS header RCODE (response code) field.
// https://datatracker.ietf.org/doc/rfc1035#section-4.1.1
func rcodeToString(rcode uint8) string {
	switch rcode {
	case 0:
		return "NoError"
	case 1:
		return "FormErr"
	case 2:
		return "ServFail"
	case 3:
		return "NXDomain"
	case 4:
		return "NotImp"
	case 5:
		return "Refused"
	default:
		return fmt.Sprintf("Unknown(%d)", rcode)
	}
}
```

vs 

```golang
func (drc DNSResponseCode) String() string {
	switch drc {
	default:
		return "Unknown"
	case DNSResponseCodeNoErr:
		return "No Error"
	case DNSResponseCodeFormErr:
		return "Format Error"
	case DNSResponseCodeServFail:
		return "Server Failure "
	case DNSResponseCodeNXDomain:
		return "Non-Existent Domain"
	case DNSResponseCodeNotImp:
		return "Not Implemented"
	case DNSResponseCodeRefused:
		return "Query Refused"
	case DNSResponseCodeYXDomain:
		return "Name Exists when it should not"
	case DNSResponseCodeYXRRSet:
		return "RR Set Exists when it should not"
	case DNSResponseCodeNXRRSet:
		return "RR Set that should exist does not"
	case DNSResponseCodeNotAuth:
		return "Server Not Authoritative for zone"
	case DNSResponseCodeNotZone:
		return "Name not contained in zone"
	case DNSResponseCodeBadVers:
		return "Bad OPT Version"
	case DNSResponseCodeBadKey:
		return "Key not recognized"
	case DNSResponseCodeBadTime:
		return "Signature out of time window"
	case DNSResponseCodeBadMode:
		return "Bad TKEY Mode"
	case DNSResponseCodeBadName:
		return "Duplicate key name"
	case DNSResponseCodeBadAlg:
		return "Algorithm not supported"
	case DNSResponseCodeBadTruc:
		return "Bad Truncation"
	case DNSResponseCodeBadCookie:
		return "Bad Cookie"
	}
}
```

I don't want to break users relying on this.

pkg/gadgets/trace/dns/tracer/tracer.go
mauriciovasquezbernal: > We don't need to send the whole packet as we are only using the DNS layer.

That's true. But I don't know if there is a way to only send **the last** part of a packet.

> BTW, we already computed the IP header len in the eBPF code, so we don't need to re-compute the offsets here.

I didn't get it. Could you extend a bit more please?

pkg/gadgets/trace/dns/tracer/bpf/dns.c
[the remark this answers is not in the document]
blanquicet: Yep, makes sense.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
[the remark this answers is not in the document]
blanquicet: > We'll need to think how to support this on the image base version later on
>
That's exactly why I'm asking 😅 .

pkg/gadgets/trace/dns/tracer/bpf/dns.c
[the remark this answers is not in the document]
blanquicet: @alban any idea what's happening here?

pkg/gadgets/trace/dns/tracer/bpf/dns.c
blanquicet: I understand your reasoning behind dividing it into two commits, but it makes the review and (managing the comments for you) more difficult. For your mental sanity 🤣 , I think you should either create two PRs, one for parsing in user-space and another for supporting IPv6; or squash the commits.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
blanquicet: Oh, that's what I was missing.

pkg/gadgets/trace/dns/tracer/tracer.go
blanquicet: `gopacket` is a well-known pkg, so we can consider it more "standard". I think that's a good justification to migrate and use it. We can alert users in the release notes and Slack about this breaking change.

pkg/gadgets/trace/dns/tracer/tracer.go
blanquicet: Answer [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2544#discussion_r1505015528).

pkg/gadgets/trace/dns/tracer/tracer.go
blanquicet: >> We don't need to send the whole packet as we are only using the DNS layer.
>>
> That's true. But I don't know if there is a way to only send **the last** part of a packet.
>
I checked that documentation and it says **the first** X bytes.

>> BTW, we already computed the IP header len in the eBPF code, so we don't need to re-compute the offsets here.
>>
>I didn't get it. Could you extend a bit more please?
>
I meant that in eBPF we are already extracting the IHL field and computing the correct offset of the DNS content (considering the IP header is not fixed). Therefore, if it were possible, we'd have been able to not send the whole packet but just from the DNS offset... Anyway, `bpf_perf_event_output` doesn't allow it, and I just saw you are sending the DNS offset in the event struct. So, it's fine.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -108,23 +106,87 @@ SEC("socket1")
...
+	// Do a first pass only to extract the port and drop the packet if it's not DNS
+	h_proto = load_half(skb, offsetof(struct ethhdr, h_proto));
+	switch (h_proto) {
+	case ETH_P_IP:
+		proto = load_byte(skb,
+				  ETH_HLEN + offsetof(struct iphdr, protocol));
+		// An IPv4 header doesn't have a fixed size. The IHL field of a packet
+		// represents the size of the IP header in 32-bit words, so we need to
+		// multiply this value by 4 to get the header size in bytes.
+		__u8 ihl_byte = load_byte(skb, ETH_HLEN);
```
blanquicet, line 124: How does this line deal with endianness and the [4 bits of version](https://github.com/torvalds/linux/blob/cf1182944c7cc9f1c21a8a44e0d29abe12527412/include/uapi/linux/ip.h#L88-L96)?

Why not use [the mechanism](https://github.com/inspektor-gadget/inspektor-gadget/blob/13cb07051c4617853983f2b1d3dc9c8296b8b626/pkg/gadgets/trace/network/tracer/bpf/network.c#L40-L49) the trace network gadget uses?

blanquicet (commented):
Interesting approach! Like Mike, looking forward to seeing this one implemented as image-based gadget.

The code looks good to me already from code inspection. I'll wait for the port params implementation and then I will test it.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
blanquicet: BTW, [this](https://medium.com/@nurkholish.halim/a-deep-dive-into-ebpf-writing-an-efficient-dns-monitoring-2c9dea92abdf) is an interesting blogpost using the same approach.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
mauriciovasquezbernal: I'll drop the second commit from this PR. I can't handle it anymore.

mauriciovasquezbernal: In any case, the original code wasn't handling it. I'll leave it as it was.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
mauriciovasquezbernal: This will be part of another PR.

docs/builtin-gadgets/trace/dns.md
```diff
@@ -83,11 +83,11 @@ The tool will list the different DNS queries that were made:
...
+test-trace-dns                          38797      38798      isc-net-0000     R  HOST      A          inspektor-gadget.io.                   No Error             2
 test-trace-dns                          38801      38802      isc-net-0000     Q  OUTGOING  AAAA       inspektor-gadget.io.                                        0
-test-trace-dns                          38801      38802      isc-net-0000     R  HOST      AAAA       inspektor-gadget.io.                   NoError              2
+test-trace-dns                          38801      38802      isc-net-0000     R  HOST      AAAA       inspektor-gadget.io.                   No Error             2
 test-trace-dns                          38807      38808      isc-net-0000     Q  OUTGOING  MX         inspektor-gadget.io.                                        0
-test-trace-dns                          38807      38808      isc-net-0000     R  HOST      MX         inspektor-gadget.io.                   NoError              3
+test-trace-dns                          38807      38808      isc-net-0000     R  HOST      MX         inspektor-gadget.io.                   No Error             3
 ```
 
 ### Limitations
```
blanquicet, line 93: Could you please update the limitations?

- The number of answers is no anymore limited to 8 and, afaiu, the gopacket pkg support uncompressed messages.
- The remaining limitations are the lack of support for DNS over TCP and IPv6. In addition, we won't parse the packet correctly if the IPv4 header has the Options field set.

blanquicet (approved):
Tested and working as expected. LGTM!

pkg/gadgets/trace/dns/tracer/bpf/dns.c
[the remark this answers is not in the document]
alban: I tried your branch without the `pad` field and I didn't get any verifier error.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -35,11 +29,26 @@
 #define DNS_QR_QUERY 0
 #define DNS_QR_RESP 1
 
+#define MAX_PORTS 16
+const volatile __u16 ports[MAX_PORTS] = { 53, 5353 };
```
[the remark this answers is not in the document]
mauriciovasquezbernal: If I don't put the size there I don't know how to write it from golang.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -35,11 +29,26 @@
...
+const volatile __u16 ports_len = 2;
+
+static __always_inline __u8 is_dns_port(__u16 port)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I didn't want to `#include <stdbool.h>` but I think it's fine. Will change it.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
```diff
@@ -114,159 +104,63 @@ struct {
...
-			int label_len = load_byte(
-				skb, DNS_OFF + sizeof(struct dnshdr) + i);
-			if (label_len == 0)
-				break;
-			// The simple solution "i += label_len" gives verifier
-			// errors, so work around with skip.
-			skip = label_len;
-		}
-	}
+	struct event_t event;
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Please see previous discussion on https://github.com/inspektor-gadget/inspektor-gadget/pull/2544#discussion_r1504086260, added a comment to make it more clear.

pkg/gadgets/trace/dns/tracer/bpf/dns.c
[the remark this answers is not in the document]
mauriciovasquezbernal: Which kernel version did you use? I get the same error on `5.15.0-94-generic` and `6.2.0-060200-generic`:

If I remove the padding it fails on my system with:

```bash 
	; struct query_key_t query_key = {
	379: (7b) *(u64 *)(r10 -120) = r1
	380: (6b) *(u16 *)(r10 -112) = r0
	; __u8 qr = flags.qr;
	381: (77) r9 >>= 15
	382: (57) r9 &= 1
	; if (qr == DNS_QR_QUERY && event.pkt_type == PACKET_OUTGOING) {
	383: (55) if r9 != 0x0 goto pc+12
	 R0=inv(id=0) R1=inv(id=0) R6=ctx(id=0,off=0,imm=0) R7=inv(id=0) R8=map_value(id=0,off=0,ks=12,vs=64,imm=0) R9=inv0 R10=fp0 fp-8=00000000 fp-16=mmmmmmmm fp-24=00000000 fp-32=0000mmmm fp-40=00000000 fp-48=0000mmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=inv fp-112=????mmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=inv fp-144=mmmmmmmm
	384: (79) r1 = *(u64 *)(r10 -136)
	385: (57) r1 &= 255
	386: (55) if r1 != 0x4 goto pc+9
	 R0=inv(id=0) R1_w=inv4 R6=ctx(id=0,off=0,imm=0) R7=inv(id=0) R8=map_value(id=0,off=0,ks=12,vs=64,imm=0) R9=inv0 R10=fp0 fp-8=00000000 fp-16=mmmmmmmm fp-24=00000000 fp-32=0000mmmm fp-40=00000000 fp-48=0000mmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=inv fp-112=????mmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=inv fp-144=mmmmmmmm
	; bpf_map_update_elem(&query_map, &query_key,
	387: (bf) r3 = r10
	388: (07) r3 += -96
	389: (bf) r2 = r10
	390: (07) r2 += -120
	391: (18) r1 = 0xffffa0188cace400
	393: (b7) r4 = 1
	394: (85) call bpf_map_update_elem#2
	invalid indirect read from stack R2 off -120+12 size 16
```

The map is created with 16 bytes as value size:

```
$ sudo strace -f -e trace=bpf ig trace dns
...
[pid 76279] bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_HASH, key_size=16, value_size=8, max_entries=1024, map_flags=0, inner_map_fd=0, map_name="query_map", map_ifindex=0, btf_fd=65, btf_key_type_id=1, btf_value_type_id=2}, 72) = 66
```

And the BTF information indicates that the struct is in fact 16 bytes:

```bash 
$ bpftool btf dump file dns_bpfel.o | grep query_key -A 3
[6] STRUCT 'query_key_t' size=16 vlen=2
	'pid_tgid' type_id=7 bits_offset=0
	'id' type_id=9 bits_offset=64
[7] TYPEDEF '__u64' type_id=8
```

@alban internally mentioned the compiler isn't filling the whole struct on initialization, this is causing the problem. But it's still not clear why the compiler is generating padding on this structure on this branch and not on main.

pkg/gadgets/trace/dns/tracer/tracer_test.go
```diff
@@ -1,69 +0,0 @@
-// Copyright 2019-2021 The Inspektor Gadget authors
```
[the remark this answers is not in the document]
mauriciovasquezbernal: How can I adapt this if the only tested function is now gone? I agree having unit test for this gadget will be great, but I feel it's out of scope of this PR.

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -97,265 +67,147 @@ func NewTracer() (*Tracer, error) {
...
+		return "OTHERHOST"
+	case 4:
+		return "OUTGOING"
+	case 5:
+		return "LOOPBACK"
+	case 6:
+		return "USER"
+	case 7:
+		return "KERNEL"
+	}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Ok, I'll undo this change to avoid discussing it. But this solution shouldn't be that bad as it's used in gopacket.

anubhabMajumdar (approved):

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -97,265 +67,147 @@ func NewTracer() (*Tracer, error) {
...
-		event.PktType = pktTypeNames[pktTypeUint]
-	}
-
-	qTypeUint := uint(bpfEvent.Qtype)
-	var ok bool
-	event.QType, ok = qTypeNames[qTypeUint]
-	if !ok {
-		event.QType = "UNASSIGNED"
+	if len(dnsLayer.Questions) > 0 {
+		question := dnsLayer.Questions[0]
```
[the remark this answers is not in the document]
blanquicet: IMO, it doesn't make sense to use gopacket parsing for some things and not for others. I agree gopacket is indeed a good source of trust, if we consider there is something missing, we should open a PR on gopacket instead of use a custom parsing. In this case in specific, honestly, I think people will mainly use the DNS gadget for tracing IPv4 and IPv6 request, and those are already included in the gopacket parsing. Said that, I suggest using the gopacket parsing also here.

integration/ig/k8s/trace_dns_test.go
```diff
@@ -175,7 +175,7 @@ func TestTraceDns(t *testing.T) {
 					PktType:    "HOST",
 					DNSName:    "nodomain.fake.test.com.",
 					QType:      "A",
-					Rcode:      "NXDomain",
+					Rcode:      "Non-Existent Domain",
```
alban, line 178: Should the width of the column be adapted?

pkg/gadgets/trace/dns/types/dns.go:
```
        Rcode      string        `json:"rcode,omitempty" column:"rcode,minWidth:8"`
```

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -37,40 +38,9 @@ import (
 
 const (
 	BPFQueryMapName = "query_map"
-	MaxAddrAnswers  = 8 // Keep aligned with MAX_ADDR_ANSWERS in bpf/dns-common.h
+	maxPorts        = 16
```
alban, line 41: Add comment both here and in dns.c:
// Keep in sync with dns.c MAX_PORTS
```
#define MAX_PORTS 16
```

pkg/gadgets/trace/dns/tracer/bpf/dns.c
alban: TIL!

I see bcc handles that automatically:
https://github.com/iovisor/bcc/blob/6cd272189771bbc7120d9d834e2588b93d8e9cfa/src/cc/frontends/clang/b_frontend_action.cc#L1017
```
            "((__u64)" + skb_len + " << 32) | BPF_F_CURRENT_CPU, " +
```

pkg/gadgets/trace/dns/tracer/tracer.go
```diff
@@ -108,254 +78,133 @@ var pktTypeNames = []string{
...
+	bpfEvent := (*dnsEventT)(unsafe.Pointer(&rawSample[0]))
+	packetBytes := rawSample[unsafe.Sizeof(*bpfEvent):]
+
+	// The DNS layer starts after the ethernet, IP and UDP headers.
+	// This is hardcoded for now as the datapath assumes the DNS packet is UDP over IPv4
+	// without any options. Next iterations will fix this.
+	// Ethernet + IP + UDP
+	const dnsOff = 14 + 20 + 8
+	dnsLayer := layers.DNS{}
+	err := dnsLayer.DecodeFromBytes(packetBytes[dnsOff:], gopacket.NilDecodeFeedback)
```
alban, line 127: To avoid `panic: runtime error: slice bounds out of range`:

You should check if `len(packetBytes) <= dnsOff`.

If the packet is a malformed UDP packet that has 7 bytes instead of 8 bytes, and without payload, the ebpf code currently does not check that.

integration/ig/k8s/trace_dns_test.go
mauriciovasquezbernal: I changed it to 12.

mauriciovasquezbernal:
Opened #2574 as a follow up work on this.,

PR #2545, opened 2024-02-27
pkg/gadgets: Fixes trace bind for kernel 6.6 and younger.

eiffel-fl:
In kernel 6.6, the bitfields in struct inet_sock where moved to a dedicated flag [1, 2, 3, 4].
To adapt to this change, a CO-RE solution was proposed in iovisor/bcc for bindsnoop [5].
This is a good solution, as it will work on kernel backporting the patches.

This commit adapts the bindsnoop solution to Inspektor Gadget.

[1]: https://lore.kernel.org/all/20230816081547.1272409-1-edumazet@google.com/
[2]: https://github.com/torvalds/linux/commit/3f7e753206bb
[3]: https://github.com/torvalds/linux/commit/ca571e2eb7eb
[4]: https://github.com/torvalds/linux/commit/4bd0623f04ee
[5]: https://github.com/iovisor/bcc/pull/4846/commits/a55b5a0c920f


pkg/gadgets/trace/bind/tracer/bpf/bindsnoop.h
```diff
@@ -32,4 +32,52 @@ union bind_options {
...
+	__u8 freebind: 1;
+	__u8 transparent: 1;
+	__u8 bind_address_no_port: 1;
+};
+
+enum {
+	INET_FLAGS_FREEBIND___x = 11,
+	INET_FLAGS_TRANSPARENT___x = 15,
+	INET_FLAGS_BIND_ADDRESS_NO_PORT___x = 18,
+};
```
mauriciovasquezbernal, line 50: I suppose we can't use `bpf_core_enum_value` with these ones as the enum is anonymous?

mauriciovasquezbernal (approved):
LGTM, thanks for the fix.

(Are you going to fix the u8 vs u16 issue upstream too?  I can do it otherwise if you want).

mauriciovasquezbernal:
> What do you mean by `u8` vs `u16`? This problem does not exist upstream as they do not use macros:
> [iovisor/bcc@a55b5a0](https://github.com/iovisor/bcc/commit/a55b5a0c920f3839fbb29b0e8837726f5fe49c72)

Ah, I thought they also have this issue. Nothing to be done then.

Issue #2444, opened 2024-02-02
trace exec: `cannot create bpf perf link: permission denied` on RHEL 9.3


matthyx:
## Description

Using the `trace exec` gadget fails on RHEL 9.3 with default kernel:
```
$ uname -a
Linux localhost.localdomain 5.14.0-362.18.1.el9_3.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Jan 3 15:54:45 EST 2024 x86_64 x86_64 x86_64 GNU/Linux
```

## Impact

Our customers cannot use our solution based on IG on RHEL 9 nodes.

## Environment and steps to reproduce

1. **Set-up**: download and install RHEL 9.3 server (https://developers.redhat.com/products/rhel/download - requires a FREE developer subscription).
2. **Task**: try to run `sudo ig trace exec`
3. **Action(s)**: it is also possible to simply run [exec.go](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/examples/gadgets/basic/trace/exec/exec.go)
  a. clone IG repo
  b. compile trace/exec
  c. run with `sudo`
4. **Error**:
```
[matthias@localhost ~]$ git clone https://github.com/inspektor-gadget/inspektor-gadget.git
Cloning into 'inspektor-gadget'...
remote: Enumerating objects: 40924, done.
remote: Counting objects: 100% (1875/1875), done.
remote: Compressing objects: 100% (939/939), done.
remote: Total 40924 (delta 1130), reused 1467 (delta 881), pack-reused 39049
Receiving objects: 100% (40924/40924), 83.34 MiB | 15.11 MiB/s, done.
Resolving deltas: 100% (25637/25637), done.
[matthias@localhost ~]$ cd inspektor-gadget/examples/gadgets/basic/trace/exec/
[matthias@localhost exec]$ ls
exec.go  README.md
[matthias@localhost exec]$ go build .
go: downloading github.com/cilium/ebpf v0.12.3
go: downloading golang.org/x/exp v0.0.0-20231108232855-2478ac86f678
go: downloading golang.org/x/text v0.14.0
go: downloading go.opentelemetry.io/otel v1.22.0
go: downloading golang.org/x/sys v0.16.0
go: downloading github.com/hashicorp/go-multierror v1.1.1
go: downloading github.com/sirupsen/logrus v1.9.3
go: downloading golang.org/x/term v0.16.0
go: downloading github.com/hashicorp/errwrap v1.1.0
[...]

PR #2546, opened 2024-02-27
gadgets: Replace context structs for tracepoint/syscall programs

mauriciovasquezbernal:
Tracepoint syscall programs should use `struct  syscall_trace_(enter|exit)` instead of `struct trace_event_raw_sys_(enter|exit)`.

See full discussion on https://github.com/inspektor-gadget/inspektor-gadget/issues/2444#issuecomment-1966440483.


mauriciovasquezbernal:
> Hi!
> 
> Can you please expand the commit message and add a references to the upstream Linux kernel commit?
> 
> Best regards.

I added the link to the kernel commit. What else would you want me to write there?

PR #2557, opened 2024-02-29
pkg/columns: add safety mechanism for dynamically added columns

flyth:
This prevents panics if `getStart()` returns a nil pointer. While this should not happen under normal circumstances, this adds another safety mechanism for handling untrusted data coming from image based gadgets.

Part of #2512


flyth:
Thanks for the review!

PR #2558, opened 2024-02-29
pkg/params: add nil checks to CopyToMap

flyth:
This helps preventing panics for operations on nil params and nil maps that could be coming from untrusted sources.


flyth:
> would it make sense to add error message?

This made me re-think. I removed the check again and made sure it is performed at a higher level.

Issue #1912, opened 2023-08-01
[RFE] Adding support for uprobes in gadgets


alban:
## Current situation

We have gadgets using different kinds of ebpf programs (kprobe, tracepoint, socket filter, etc.) but no gadgets using uprobes so far.

## Impact

Many use cases needs uprobe programs and they are not implemented:
- [ ] See encrypted network traffic by attaching to functions such as openssl's SSL_read() and SSL_write() (see [bcc's sslsnif](https://github.com/iovisor/bcc/blob/master/tools/sslsniff.py))
- [ ] memleak (see [bcc's memleak](https://github.com/iovisor/bcc/blob/41f8010e2e85616ad37303af20f77c1c717b9b11/libbpf-tools/memleak.bpf.c#L164))
- [ ] deadlock (see [bcc's deadlock](https://github.com/iovisor/bcc/blob/master/tools/deadlock.py#L501))

## Ideal future situation

Full support would include:

- [ ] Inspektor Gadget should support uprobes so that the aforementioned gadgets can be implemented as image-based gadgets.
- [ ] ~`ig script` support uprobe scripts~ (not the priority at the moment)
  - [ ] ~Support for dynamic container creation/termination~
- [ ] IG has the following gadgets, either built-in or containerized gadgets:
  - [ ] sslsniff
  - [ ] memleak
  - [ ] deadlock
- [ ] Support for user stacks
  - [ ] in uprobe gadgets
  - [ ] `ig profile cpu` (this is not using uprobes but the code for printing user stack should be similar)

Since it is quite complex, I suggest to start with [bcc's bashreadline](https://github.com/iovisor/bcc/blob/master/tools/bashreadline_example.txt) because it does not require support for user stack.

### Multiple attachment

Similarly to networking gadgets, uprobes ebpf programs need to be attached multiple times (one time for each container) because they are attached to the inode of interest in the rootfs of the container. So we will probably need something similar to pkg/gadgets/internal/networktracer to attach/detach containers but for uprobes instead of socket filters.

Since containers can be created and terminated at any time, the attach/detach should be done dynamically like it was done in pkg/gadgets/internal/networktracer. For bpftrace running as an external process, that will be very difficult since that would require dynamic communication between IG and bpftrace processes.

### Proper filtering

Note that the uprobe program can be executed outside of the selected container, so the uprobe program should check if it is called in a container of interest:
[...]

PR #2563, opened 2024-03-03
Introduce `uprobetracer` for auto attach/detach

i-Pear:
# Description
Part of #1912.
This patch introduces a new package `uprobetracer` for auto attaching/detaching.

# Testing
a. Two containers sharing the same image, make sure the target file is only probed once.
b. Starting/stopping/restarting the containers, and see if it works well.

### DIFF
--- a/gadgets/trace_malloc/program.bpf.c
+++ b/gadgets/trace_malloc/program.bpf.c
@@ -24,7 +24,7 @@ struct event {
 
 GADGET_TRACER_MAP(events, 1024 * 256);
 
-GADGET_TRACER(open, events, event);
+GADGET_TRACER(malloc, events, event);
 
 static __always_inline int submit_memop_event(struct pt_regs *ctx,
 					      enum memop operation, __u64 addr)
--- a/gadgets/trace_malloc/program.bpf.c
+++ b/gadgets/trace_malloc/program.bpf.c
@@ -49,13 +49,13 @@ static __always_inline int submit_memop_event(struct pt_regs *ctx,
 	return 0;
 }
 
-SEC("uretprobe//usr/lib/libc.so.6:malloc")
+SEC("uretprobe/libc:malloc")
 int trace_uprobe_malloc(struct pt_regs *ctx)
 {
 	return submit_memop_event(ctx, MALLOC, PT_REGS_RC(ctx));
 }
 
-SEC("uprobe//usr/lib/libc.so.6:free")
+SEC("uprobe/libc:free")
 int trace_uprobe_free(struct pt_regs *ctx)
 {
 	return submit_memop_event(ctx, FREE, PT_REGS_PARM1(ctx));
--- a/gadgets/trace_malloc/program.bpf.c
+++ b/gadgets/trace_malloc/program.bpf.c
@@ -4,8 +4,10 @@
 #include <vmlinux.h>
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_tracing.h>
+
 #include <gadget/buffer.h>
 #include <gadget/macros.h>
+#include <gadget/types.h>
 
 enum memop {
 	MALLOC,
@@ -13,6 +15,7 @@ enum memop {
 };
 
 struct event {
+	gadget_mntns_id mntns_id;
 	__u32 pid;
 	__u8 comm[TASK_COMM_LEN];
 	enum memop operation;
@@ -27,11 +30,15 @@ static __always_inline int submit_memop_event(struct pt_regs *ctx,
 					      enum memop operation, __u64 addr)
 {
 	struct event *event;
+	struct task_struct *task;
 
 	event = gadget_reserve_buf(&events, sizeof(*event));
 	if (!event)
 		return 0;
 
+	task = (struct task_struct *)bpf_get_current_task();
+
+	event->mntns_id = (u64)BPF_CORE_READ(task, nsproxy, mnt_ns, ns.inum);
 	event->pid = bpf_get_current_pid_tgid() >> 32;
 	bpf_get_current_comm(event->comm, sizeof(event->comm));
 	event->operation = operation;
--- a/pkg/gadgets/run/tracer/tracer.go
+++ b/pkg/gadgets/run/tracer/tracer.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"net"
 	"os"
-	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
@@ -48,9 +47,9 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/socketenricher"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/tchandler"
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/uprobetracer"
 	bpfiterns "github.com/inspektor-gadget/inspektor-gadget/pkg/utils/bpf-iter-ns"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
 )
 
 // keep aligned with pkg/gadgets/common/types.h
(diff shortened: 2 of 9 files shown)


pkg/gadgets/run/tracer/tracer.go
```diff
@@ -376,6 +376,22 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
 		case strings.HasPrefix(p.SectionName, "kretprobe/"):
 			logger.Debugf("Attaching kretprobe %q to %q", p.Name, p.AttachTo)
 			return link.Kretprobe(p.AttachTo, prog, nil)
+		case strings.HasPrefix(p.SectionName, "uprobe/"):
+			logger.Debugf("Attaching uprobe %q to %q", p.Name, p.AttachTo)
+			parts := strings.Split(p.AttachTo, ":")
+			ex, err := link.OpenExecutable(parts[0])
```
alban, line 382: To open executables from all containers, you can look at the following packages for inspiration:
* Socket filter ebpf programs:
  * https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/devel/network-gadget-dispatcher.png
  * https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/networktracer/tracer.go#L15-L31
* tc ebpf programs:
  * https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/tchandler/tracer.go#L15-L19

alban (commented):
Thanks for the quick draft PR!

I've added some initial comments.

Dockerfiles/ig.Dockerfile
```diff
@@ -11,7 +11,7 @@ ARG EBPF_BUILDER=ghcr.io/inspektor-gadget/ebpf-builder:latest
 ENV EBPF_BUILDER=${EBPF_BUILDER}
 
 COPY go.mod go.sum /cache/
-RUN cd /cache && go mod download
+RUN cd /cache && go env -w GOPROXY="https://goproxy.cn,direct" && go mod download
```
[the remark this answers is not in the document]
i-Pear: This will be dropped in the final version, it's just a workaround for my local dev environment.

alban:
TODOs:
- [ ] Documentation in [program-types.md](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/reference/program-types.md)
- [ ] Update `gadgets/Makefile` with `trace_malloc`
- [ ] Check CI results

pkg/uprobetracer/tracer.go
```diff
@@ -0,0 +1,85 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package uprobetracer
+
+import (
+	"fmt"
+	"github.com/cilium/ebpf"
+	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	"sync"
```
alban, line 21: The coding style is to group imports in 3 groups separated by an empty line:
- standard packages
- external dependencies
- internal dependencies

See example at https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/containerd/containerd.go#L17-L34

pkg/uprobetracer/tracer.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type inode_uuid struct {
+	fs_id int
```
alban, line 25:
```suggestion
	fsID unix.Fsid
```

Type of fsID:
https://pkg.go.dev/golang.org/x/sys@v0.18.0/unix#Fsid

gadgets/trace_malloc/program.bpf.c
alban, line 1: Each gadget should have a gadget.yaml file, see examples in other directories.

It can be automatically generated with `ig image build ... --update-metadata`.

alban (commented):
Thanks. Some additional comments.

i-Pear:
Thanks for the detailed comments, I will push an update after the implementation is finished.

i-Pear:
TODO

- [ ] support searching `ld` library path
- [ ] show `addr` in hex format
- [ ] avoid attacking with `../`

gadgets/trace_malloc/program.bpf.c
```diff
@@ -11,13 +11,14 @@ struct event {
 	__u32 pid;
 	__u64 addr;
 	bool is_malloc;
+	__u8 comm[TASK_COMM_LEN];
```
alban, line 14: gadget.yaml will need to be updated. This can be done automatically with `ig image build --update-metadata`

pkg/uprobetracer/tracer.go
```diff
@@ -0,0 +1,245 @@
...
+
+// We hold fd(s) of the executables, so we can use `/proc/self/fd/...` for attaching, it is used to avoid fd-reusing
+// We open the executables at attachContainer event, so if the file changed before the ebpf program loads,
+// we will still attach to the origin executable.
+
+// We are not maintaining ebpf.collection or perf-ring buffer by ourselves, it's hold by the parent tracer
+
+// All interfaces should hold lock, and inner functions do not
+
+package uprobetracer
```
alban, line 25: I think those comments could be added to the package documentation if you remove the empty lines.

https://go.dev/doc/comment

pkg/uprobetracer/tracer.go
```diff
@@ -0,0 +1,245 @@
...
+		t.attach(pid)
+	}
+	t.pendingCtrPids = nil
+
+	return nil
+}
+
+func searchForLibrary(containerPid uint32, path string) (string, error) {
+	if filepath.IsAbs(path) {
+		return filepath.Join(host.HostProcFs, fmt.Sprint(containerPid), "root", path), nil
```
alban, line 146: Consider using https://github.com/cyphar/filepath-securejoin

Credits to @rata for the link :)

```golang
SecureJoin(filepath.Join(host.HostProcFs, fmt.Sprint(containerPid), "root"), path)
```

pkg/uprobetracer/tracer.go
```diff
@@ -0,0 +1,245 @@
...
+	file, err := os.Open(attachFilePath)
+	if err != nil {
+		return
+	}
+
+	// get fsID of the executable
+	stat := unix.Statfs_t{}
+	err = unix.Fstatfs(int(file.Fd()), &stat)
+	if err != nil {
+		return
```
alban, line 172: `file.Close()` should be called explicitly in case of error.

The file descriptor will still be closed whenever the Go GC releases the variable `file`, but it is more difficult to debug when fds are closed at a nondeterministic time.

Ditto in other error cases.

pkg/uprobetracer/tracer.go
```diff
@@ -0,0 +1,245 @@
...
+
+type Tracer[Event any] struct {
+	progName       string
+	progType       string
+	attachFilePath string
+	attachSymbol   string
+	prog           *ebpf.Program
+
+	ctrPid2FsId    map[uint32]unix.Fsid
+	inodeRefCount  map[unix.Fsid]*inodeKeeper
```
alban, line 66: Could you add some comments explaining the purpose of this map and why you use `unix.Fsid` as a key?

Ditto for other maps

alban (commented):
Thanks. Some initial comments on the uprobetracer.

pkg/uprobetracer/tracer.go
i-Pear: Sorry, I misread the document. I thought the second value in the Fsid structure was the inodeID.

i-Pear:
using `--host` may be dangerous, because if ig is using the probed `libc.so`, it may become a dead loop.

i-Pear:
I have found that the `FsID` obtained using `statfs` does not correspond one-to-one with the actual inode. If two containers share the same image, the `FsID` is different, but the `inodeID` remains the same. Therefore, currently using `pair<FsID, inodeID>` as the key does not prevent the same file from being probed multiple times.

Does anyone have any ideas on this?

i-Pear:
TODO

- update upstream `cillum/ebpf` dependency: pass `inherit` flag to uprobe syscall

i-Pear:
BUGS

- [ ] pid filter not working with `uretprobe`
- [ ] `inherit` option not working with uprobe/uretprobe
- [ ] uprobe/uretprobe sometimes fail (not attaching successfully)
- [ ] uprobe/uretprobe sometimes disconnected (maybe caused by goalng gc? check it later) (low priority)

PR #2573, opened 2024-03-04
gha: Automatically mark and close stale issues / PRs

mauriciovasquezbernal:
Issues and PRs will be marked as stale after 60 days of inactivity, and will be closed 14 days after is there is not new activity.

This is used to avoid a lot of old issues / PRs piling up on the repository.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,37 @@
+name: Close stale issues and PRs
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 60
+          days-before-issue-close: 14
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had recent activity.
+            It will be closed in 14 days if no further activity occurs.
+          close-issue-message: |
+            This issue has been automatically closed due to inactivity.
+            If you believe this was closed in error, please feel free to reopen it.
+          stale-issue-label: lifecycle/stale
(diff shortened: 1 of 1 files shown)


mqasimsarfraz:
I assume we don't need any permission block ? I do see recommended permissions [here](https://github.com/actions/stale?tab=readme-ov-file#recommended-permissions) but perhaps that is optional ?

mauriciovasquezbernal:
Good catch, I don't think they're optional. Copied and paste from one of the examples below that don't have them.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,37 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 60
+          days-before-issue-close: 14
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had recent activity.
+            It will be closed if no further activity occurs.
```
mqasimsarfraz, line 20: nit: Perhaps we can mention it will be closed in 14 days, when we post stale issue message?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,37 @@
...
+            If you believe this was closed in error, please feel free to reopen it.
+          stale-issue-label: lifecycle/stale
+          exempt-issue-labels: good-first-issue,lifecycle/staleproof
+          exempt-all-issue-assignees: true # do not close issues with assignees
+
+          days-before-pr-stale: 60
+          days-before-pr-close: 14
+          stale-pr-message: |
+            This pull request has been automatically marked as stale because it has not had recent activity.
+            It will be closed if no further activity occurs. Thank you for your contributions.
```
mqasimsarfraz, line 32: same as above

mqasimsarfraz (approved):
LGTM, thanks for the change!

PR #2580, opened 2024-03-05
run: Support basic uprobe and uretprobe

i-Pear:
# run: Support basic uprobe and uretprobe
This patch introduces basic support for uprobe and uretprobe.
Filtering as well as auto attach/detach will be supported in the future.

Also, a simple gadget `trace_malloc` is added as demo.

## Testing done
`ig image build -t trace_malloc:latest .`
`ig run trace_malloc:latest --host`

### DIFF
--- a/gadgets/trace_malloc/gadget.yaml
+++ b/gadgets/trace_malloc/gadget.yaml
@@ -1,0 +1,28 @@
+name: trace malloc
+description: use uprobe to trace malloc and free in libc.so
+tracers:
+  events:
+    mapName: events
+    structName: event
+structs:
+  event:
+    fields:
+    - name: pid
+      attributes:
+        template: pid
+    - name: comm
+      description: command
+      attributes:
+        template: comm
+    - name: operation
+      description: memory operation type
+      attributes:
+        width: 16
+        alignment: left
+        ellipsis: end
+    - name: addr
+      description: address of malloc/free
+      attributes:
+        width: 20
+        alignment: left
+        ellipsis: end
--- a/gadgets/trace_malloc/program.bpf.c
+++ b/gadgets/trace_malloc/program.bpf.c
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2024 The Inspektor Gadget authors */
+
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <gadget/buffer.h>
+#include <gadget/macros.h>
+
+enum memop {
+	MALLOC,
+	FREE,
+};
+
+struct event {
+	__u32 pid;
+	__u8 comm[TASK_COMM_LEN];
+	enum memop operation;
+	__u64 addr;
+};
+
+GADGET_TRACER_MAP(events, 1024 * 256);
+
+GADGET_TRACER(open, events, event);
+
+static __always_inline int submit_memop_event(struct pt_regs *ctx,
+					      enum memop operation, __u64 addr)
+{
+	struct event *event;
+
+	event = gadget_reserve_buf(&events, sizeof(*event));
+	if (!event)
+		return 0;
+
+	event->pid = bpf_get_current_pid_tgid() >> 32;
+	bpf_get_current_comm(event->comm, sizeof(event->comm));
+	event->operation = operation;
+	event->addr = addr;
+
+	gadget_submit_buf(ctx, &events, event, sizeof(*event));
+
+	return 0;
+}
+
+SEC("uretprobe//usr/lib/libc.so.6:malloc")
+int trace_uprobe_malloc(struct pt_regs *ctx)
+{
+	return submit_memop_event(ctx, MALLOC, PT_REGS_RC(ctx));
+}
+
+SEC("uprobe//usr/lib/libc.so.6:free")
+int trace_uprobe_free(struct pt_regs *ctx)
+{
+	return submit_memop_event(ctx, FREE, PT_REGS_PARM1(ctx));
+}
+
+char LICENSE[] SEC("license") = "Dual BSD/GPL";
--- a/pkg/gadgets/run/tracer/tracer.go
+++ b/pkg/gadgets/run/tracer/tracer.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
@@ -49,6 +50,7 @@ import (
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 	bpfiterns "github.com/inspektor-gadget/inspektor-gadget/pkg/utils/bpf-iter-ns"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
 )
 
 // keep aligned with pkg/gadgets/common/types.h
@@ -376,6 +378,37 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
 		case strings.HasPrefix(p.SectionName, "kretprobe/"):
 			logger.Debugf("Attaching kretprobe %q to %q", p.Name, p.AttachTo)
 			return link.Kretprobe(p.AttachTo, prog, nil)
+		case strings.HasPrefix(p.SectionName, "uprobe/") || strings.HasPrefix(p.SectionName, "uretprobe/"):
+			captureHost := false
+			for _, container := range t.containers {
+				if container.Pid == 1 {
+					captureHost = true
+				}
+			}
+			if !captureHost {
+				return nil, fmt.Errorf("uprobe can only be used with --host at this moment")
+			}
+
+			parts := strings.Split(p.AttachTo, ":")
+			if len(parts) < 2 {
+				return nil, fmt.Errorf("invalid section name %q", p.AttachTo)
+			}
+			if !filepath.IsAbs(parts[0]) {
+				return nil, fmt.Errorf("section name is not an absolute path: %q", parts[0])
+			}
+			executablePath := filepath.Join(host.HostProcFs, "1/root", parts[0])
+			ex, err := link.OpenExecutable(executablePath)
+			if err != nil {
+				return nil, fmt.Errorf("opening executable: %q", executablePath)
+			}
+
+			logger.Debugf("Attaching uprobe %q to %q", p.Name, p.AttachTo)
+			switch strings.Split(p.SectionName, "/")[0] {
+			case "uprobe":
+				return ex.Uprobe(parts[1], prog, nil)
+			case "uretprobe":
+				return ex.Uretprobe(parts[1], prog, nil)
+			}
 		}
 		return nil, fmt.Errorf("unsupported section name %q for program %q", p.Name, p.SectionName)
 	case ebpf.TracePoint:
--- a/docs/reference/program-types.md
+++ b/docs/reference/program-types.md
@@ -52,3 +52,9 @@ Inspektor Gadget supports running multiple gadgets that use SchedCLS programs at
 Programs must return `TC_ACT_UNSPEC` in order to allow the packet to be processed by other gadgets.
 The order of execution of the programs is not deterministic, this is something we could visit later
 on.
+
+### Uprobes / Uretprobes (experimental)
+
+The section name must use the `uprobe/<file_path>:<symbol>` or `uretprobe/<file_path>:<symbol>` formats.
+`<file_path>` is the absolute path of an executable or a library, that the uprobe will be attached to.
+`<symbol>` is a debugging symbol that can be found in the file mentioned above.
--- a/gadgets/Makefile
+++ b/gadgets/Makefile
@@ -7,6 +7,7 @@ IG ?= ig
 GADGETS = \
 	trace_dns \
 	trace_exec \
+	trace_malloc \
 	trace_mount \
 	trace_oomkill \
 	trace_open \


i-Pear:
Sorry, forgot to add the gadget name into `gadgets/Makfile`

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -376,6 +376,46 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
...
+			}
+			if !captureHost {
+				return nil, fmt.Errorf("uprobe can only be used with --host at this moment")
+			}
+			logger.Debugf("Attaching uprobe %q to %q", p.Name, p.AttachTo)
+			parts := strings.Split(p.AttachTo, ":")
+			if len(parts) < 2 {
+				return nil, fmt.Errorf("invalid section name %q", p.AttachTo)
+			}
+			ex, err := link.OpenExecutable(parts[0])
```
alban, line 394: We should still use `filepath.Join(host.HostProcFs, "1/root", parts[0])` because even though this PR does not support tracing containers, it should support `ig` running in a container.

I think we should mandate an absolute path with `filepath.IsAbs()`

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -376,6 +376,46 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
 		case strings.HasPrefix(p.SectionName, "kretprobe/"):
 			logger.Debugf("Attaching kretprobe %q to %q", p.Name, p.AttachTo)
 			return link.Kretprobe(p.AttachTo, prog, nil)
+		case strings.HasPrefix(p.SectionName, "uprobe/"):
```
alban, line 379: Could the code be factorized a bit with something like:
```suggestion
		case strings.HasPrefix(p.SectionName, "uprobe/") || strings.HasPrefix(p.SectionName, "uretprobe/"):
```

Only the lines with `ex.Uprobe` or `ex.Uretprobe` differ.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -376,6 +376,46 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
...
+			captureHost := false
+			for _, container := range t.containers {
+				if container.Pid == 1 {
+					captureHost = true
+				}
+			}
+			if !captureHost {
+				return nil, fmt.Errorf("uprobe can only be used with --host at this moment")
```
alban, line 387: The test for `--host` is not ideal but this is ok for me since the next PR will add support for tracing containers.

alban (commented):
Thanks! A couple of changes requested. It should be good after that.

pkg/gadgets/run/tracer/tracer.go
i-Pear: Yes, the handling of `--host` here doesn't align with its original intent. According to the documentation of IG, when providing the `--host`, it's not only necessary to track processes on the host machine but also processes within containers.
However, at present, we're unable to implement tracing within containers. So we force users to specify `--host`, and we are only tracking processes outside of containers.

i-Pear: The next PR will come soon, so we can refactor it later.

gadgets/trace_malloc/program.bpf.c
```diff
@@ -0,0 +1,56 @@
...
+	__u64 addr;
+	bool is_malloc;
+};
+
+GADGET_TRACER_MAP(events, 1024 * 256);
+
+GADGET_TRACER(open, events, event);
+
+SEC("uretprobe//usr/lib/libc.so.6:malloc")
+int trace_uprobe_malloc(struct pt_regs *ctx)
```
alban, line 21: On Fedora, I had to use `/lib64/libc.so.6` instead of `/usr/lib/libc.so.6` to make it work. Do you know the difference between those two paths?

gadgets/trace_malloc/program.bpf.c
```diff
@@ -0,0 +1,56 @@
...
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include <gadget/buffer.h>
+#include <gadget/macros.h>
+
+struct event {
+	__u32 pid;
+	__u64 addr;
+	bool is_malloc;
```
alban, line 13: In the next PR, could you add the `comm` field?

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -376,6 +378,37 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
...
+			}
+
+			parts := strings.Split(p.AttachTo, ":")
+			if len(parts) < 2 {
+				return nil, fmt.Errorf("invalid section name %q", p.AttachTo)
+			}
+			if !filepath.IsAbs(parts[0]) {
+				return nil, fmt.Errorf("section name is not an absolute path: %q", parts[0])
+			}
+			executablePath := filepath.Join(host.HostProcFs, "1/root", parts[0])
```
alban, line 399: I was experimenting with paths such as:
```
SEC("uretprobe//../root/lib64/libc.so.6:malloc")
```
It gets translated to `/proc/1/root/../root/lib64/libc.so.6`.

I am not sure if we should use `filepath.Clean()` to remove the `..` part, or if we should care at all.

I suggest to keep it like that for now. Since we are tracing on the host with full privileges, and we implicitly trust the gadget image, there are no escaping concerns anyway.

For the next PR with support for tracing container workloads, we can think whether we want to prevent escaping, including symlink escaping.

cc @rata

alban (approved):
LGTM, thanks!

gadgets/trace_malloc/program.bpf.c
alban: Replying to myself: x86_64 vs i686.

```
$ file /lib64/libc.so.6
/lib64/libc.so.6: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=f888be5f5e7d58e04cabb8c675c7ab94e77dd68c, for GNU/Linux 3.2.0, not stripped

$ rpm -qf /lib64/libc.so.6
glibc-2.37-16.fc38.x86_64
```

```
$ file /usr/lib/libc.so.6
/usr/lib/libc.so.6: ELF 32-bit LSB shared object, Intel 80386, version 1 (GNU/Linux), dynamically linked, interpreter /lib/ld-linux.so.2, BuildID[sha1]=a116613533d396e2c906376b67b7d60b398fe10a, for GNU/Linux 3.2.0, not stripped
$ rpm -qf /usr/lib/libc.so.6
glibc-2.37-16.fc38.i686
```

I guess your distribution uses different paths.

Since different containers use different base Linux distributions, I wonder if we will need to search for libraries in different paths following the same algorithm as defined in [ld.so(8)](https://man7.org/linux/man-pages/man8/ld.so.8.html).

See also [prior art from bpftrace](https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#probes-uprobe):

> When tracing libraries, it is sufficient to specify the library name instead of a full path. The path will be then automatically resolved using /etc/ld.so.cache:
> 
> uprobe:libc:malloc { printf("Allocated %d bytes\n", arg0); }

In any cases, that's something for a future PR.

i-Pear: On my computer, the above two files share the same inode. But on ubuntu, it's `/usr/lib32/libc.so.6` and `/usr/lib/x86_64-linux-gnu/libc.so.6`. So it may depends on the default `ld` searching path of the distros.

gadgets/trace_malloc/program.bpf.c
i-Pear: btw, I'm looking for how to print the `addr` in hex format, but I didn't find any examples. This is a note that I should look into it later.

alban: cc @flyth

mauriciovasquezbernal: > I'm looking for how to print the addr in hex format, but I didn't find any examples. This is a note that I should look into it later.

We don't support it yet. Would you mind opening an issue? I suppose we can implement it by having a new attribute that indicates it must be printed using hex format.

gadgets/trace_malloc/program.bpf.c
```diff
@@ -0,0 +1,56 @@
...
+		return 0;
+
+	event->pid = bpf_get_current_pid_tgid() >> 32;
+	event->addr = PT_REGS_PARM1(ctx);
+	event->is_malloc = 0;
+
+	gadget_submit_buf(ctx, &events, event, sizeof(*event));
+
+	return 0;
+}
```
mauriciovasquezbernal, line 54: nit: perhaps you can create a helper function that does all the work and receives a `is_malloc` argument? Then these two programs will use the same function.

gadgets/trace_malloc/program.bpf.c
mauriciovasquezbernal, line 13: I'm wondering if it's better to use an enum:. 

```c
enum type {
	MALLOC,
	FREE,
}
```

The name of the operation will be printed on the IU. And it'll also help if we want to add more functions like `calloc`, `realloc` in the future to this program.

pkg/gadgets/run/tracer/tracer.go
```diff
@@ -376,6 +378,37 @@ func (t *Tracer) attachProgram(gadgetCtx gadgets.GadgetContext, p *ebpf.ProgramS
...
+			ex, err := link.OpenExecutable(executablePath)
+			if err != nil {
+				return nil, fmt.Errorf("opening executable: %q", executablePath)
+			}
+
+			logger.Debugf("Attaching uprobe %q to %q", p.Name, p.AttachTo)
+			switch strings.Split(p.SectionName, "/")[0] {
+			case "uprobe":
+				return ex.Uprobe(parts[1], prog, nil)
```
mauriciovasquezbernal, line 408: There is something I don't understand, the [documentation](https://pkg.go.dev/github.com/cilium/ebpf@v0.13.2/link#Executable.Uprobe) of this function states:

> Functions provided by shared libraries can currently not be traced and will result in an ErrNotSupported. 

But we're doing exactly that, aren't we? Is the documentation outdated?

mauriciovasquezbernal (commented):
Some minor comments. I understand this is a first PR on a series that will implement this support, so I'll be fine merging this without testing for the time being.

gadgets/trace_malloc/program.bpf.c
```diff
@@ -0,0 +1,56 @@
...
+
+char LICENSE[] SEC("license") = "GPL";
```
mauriciovasquezbernal, line 56: I'd suggest to use `Dual BSD/GPL` as done in other gadgets.

pkg/gadgets/run/tracer/tracer.go
alban: I think the documentation is incorrect. The uprobe feature is implemented by the kernel, and the kernel seems happy to support both executables and libraries:
https://docs.kernel.org/trace/uprobetracer.html

pkg/gadgets/run/tracer/tracer.go
rata: SGTM. I'd add a comment so we don't forget about this in the future.

rata: I'd mention the escape and that we should use securejoin in the future: https://github.com/cyphar/filepath-securejoin

I have no opinion regarding doing it now vs add a comment and do it later.

gadgets/trace_malloc/program.bpf.c
i-Pear: Opened issue #2589

pkg/gadgets/run/tracer/tracer.go
i-Pear: The next PR will come next week, and it will extend this part by supporting library searching.
I added a TODO in #2563

i-Pear:
Did a refactoring on the `trace_malloc` gadget. Now it also supports the `comm` field.

gadgets/trace_malloc/program.bpf.c
alban: The first line of the file can be updated too.

i-Pear: Changed to `(LGPL-2.1 OR BSD-2-Clause)`

gadgets/trace_malloc/gadget.yaml
```diff
@@ -0,0 +1,30 @@
...
+    fields:
+    - name: pid
+      attributes:
+        template: pid
+    - name: comm
+      description: command
+      attributes:
+        width: 16
+        alignment: left
+        ellipsis: end
```
mauriciovasquezbernal, line 18: nit: we have a template for it: `template: comm`

gadgets/trace_malloc/program.bpf.c
```diff
@@ -0,0 +1,56 @@
...
+	__u8 comm[TASK_COMM_LEN];
+	enum memop operation;
+	__u64 addr;
+};
+
+GADGET_TRACER_MAP(events, 1024 * 256);
+
+GADGET_TRACER(open, events, event);
+
+int submit_memop_event(struct pt_regs *ctx, enum memop operation, __u64 addr)
```
mauriciovasquezbernal, line 26: nit: use `__always_inline`?

mauriciovasquezbernal (approved):
LGTM. Thanks!

gadgets/trace_malloc/program.bpf.c
```diff
@@ -0,0 +1,57 @@
...
+struct event {
+	__u32 pid;
+	__u8 comm[TASK_COMM_LEN];
+	enum memop operation;
+	__u64 addr;
+};
+
+GADGET_TRACER_MAP(events, 1024 * 256);
+
+GADGET_TRACER(open, events, event);
```
alban, line 24: s/open/malloc

gadgets/trace_malloc/gadget.yaml
```diff
@@ -0,0 +1,28 @@
+name: trace malloc
+description: use uprobe to trace malloc and free in libc.so
+tracers:
+  events:
```
alban, line 4: The name of the tracer needs to match what's defined in GADGET_TRACER():
s/events/malloc

alban (commented):
One thing that I missed in the previous review.

gadgets/trace_malloc/gadget.yaml
i-Pear: Thanks, will fix in the next PR.

i-Pear: btw, if there a event field related to the `RUNTIME.CONTAINERNAME` column?

alban: The columns from container enrichment are automatically added when there is a `mntns_id` field.

PR #2584, opened 2024-03-06
gha/stale: Update operations-per-run to 500

mauriciovasquezbernal:
In [0] we hit the limit, hence better to increase it.

[0]: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8165699333


mqasimsarfraz (approved):
LGTM! I see that [rate limiting](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8165699333/job/22323281590#step:2:1720) won't be an issue here so all good!

PR #2585, opened 2024-03-06
pkg/oci: Few cleanups and improvements to make code clearer

blanquicet:
It doesn't change the integration with other modules, so rebasing refactoring on top of this is fast-forward. cc @flyth


flyth (approved):
Looks great, thanks!

PR #2603, opened 2024-03-11
trace exec: add pcomm field (parent command)

alban:
# trace exec: add pcomm field (parent command)

The trace exec gadgets shows new process executed. It is useful to know the parent processes that started the new processes. We already had the parent pid. This patch adds the parent command.

Similarly to the `comm` field, the `pcomm` field is limited to 16 characters by the kernel.

## How to use

Same as before

## Testing done

```
$ sh -c 'sh -c /bin/echo | /bin/cat'
```

```
$ sudo -E ./ig trace exec --host
RUNTIME.CONTAINERNAME       PID        PPID       COMM           PCOMM          RET ARGS
                            2192986    402902     sh             bash           0   /usr/bin/sh -c sh -c /bin/echo|/b…
                            2192987    2192986    sh             sh             0   /usr/bin/sh -c /bin/echo
                            2192988    2192986    cat            sh             0   /bin/cat
                            2192987    2192986    echo           sh             0   /bin/echo
```

## TODOs

- [ ] Update documentation
- [ ] Update image-based gadget trace_exec


alban:
Superseded by https://github.com/inspektor-gadget/inspektor-gadget/pull/2604

I want to prefix the branch with `citest-` but I cannot update the PR to use a different branch, hence closing.

Issue #2046, opened 2023-09-15
image-based gadgets: Gadgets testing considerations


mauriciovasquezbernal:
We have integration tests for the built-in and some image-based gadgets in https://github.com/inspektor-gadget/inspektor-gadget/tree/main/integration. We should provide a way for gadget-developers to easily implement integration tests for their gadgets too. In order to support this, we need:

1. Way to run ig: Wrapper around cmd.Exec (until we have a proper API #2021)
2. Way to generate events: Run commands in real containers: We already have some logic for that:
   - docker (https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/docker.go)
   - containerd (https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/containerd.go). 
3. Way to capture and compare events: We have some "match" functions in https://github.com/inspektor-gadget/inspektor-gadget/blob/main/integration/run_helpers.go. We need to improve them a lot to be easier to use, but they are a good starting point.

[LFX](https://mentorship.lfx.linuxfoundation.org/project/65719d46-b71f-4778-a29b-3bb878d6ec70)

ShivamAhir:
Hello, I am very interested in this project and would like to contribute in this season of LFX;)

ahsenkamal:
Hi @mauriciovasquezbernal I'd love to contribute to this during this LFX term. Are there any resources to go through and any pre-tests for applying?

PR #2450, opened 2024-02-05
Avoid repeated code on inspektor-gadget and kubectl-gadget integration tests

pawarpranav83:
## Avoid repeated code on inspektor-gadget and kubectl-gadget integration tests

Fixes #1271 
Removed code redundancy from common tests of `ig/k8s` and `kubectl-gadget` (integration/inspektor-gadget).

## Testing done

Integration tests were passed.
ig: `make test -C ./integration/k8s`
kubectl-gadget: `make integration-tests`

PR #2607, opened 2024-03-12
Testing package for image-based gadgets

pawarpranav83:
Provides a framework for testing image-based gadgets for internal and external users.
Partially solves #2046

## How to use

Functions in the ig module would be used in the gadget template where the users would define their custom gadgets.

## Testing done

Run `make gadgets-test -C ./gadgets`.
Currently, it tests trace_open image-based gadget.


pkg/test-fw/ig/command.go
```diff
@@ -0,0 +1,94 @@
...
+package ig
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type Command struct {
```
pawarpranav83, line 24: Instead of `IG`, the `Command` struct implements the interface `Teststeps`. Other methods are implemented on IG except `Run`, `Start` and `Stop`.
Since these methods require `*testing.T` parameter.

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+	// It corresponds to gadget like execsnoop which wait user to type Ctrl^C.
+	StartAndStop bool
+
+	// started indicates this command was started.
+	// It is only used by command which have StartAndStop set.
+	started bool
+
+	// Cleanup indicates this command is used to clean resource and should not be
+	// skipped even if previous commands failed.
+	Cleanup bool
```
mauriciovasquezbernal, line 49: nit: any reason to export this field?

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+	ig.createCmd()
+	if err := ig.cmd.Run(); err != nil {
+		switch e := err.(type) {
+		case *exec.Error:
+			return fmt.Errorf("command execution: %w", err)
+		case *exec.ExitError:
+			return fmt.Errorf("command exit code = %d", e.ExitCode())
+		default:
+			return err
+		}
```
mauriciovasquezbernal, line 98: Do you really need this switch statement here? Isn't it possible to return `err`?

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+	if err := ig.cmd.Start(); err != nil {
+		switch e := err.(type) {
+		case *exec.Error:
+			return fmt.Errorf("command execution: %w", err)
+		case *exec.ExitError:
+			return fmt.Errorf("command exit code = %d", e.ExitCode())
+		default:
+			return err
+		}
+	}
```
mauriciovasquezbernal, line 119: Same as above.

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func WithImage(image string) option {
+	return func(ig *IG) {
+		ig.image = image
+	}
+}
+
+// Flags should be in the form: "--flag_name=value" or "-shorthand=value"
```
mauriciovasquezbernal, line 203: nit: handle the lint comments.

pkg/test-fw/tests/trace_open_test.go
```diff
@@ -0,0 +1,122 @@
...
+	Type EventType `json:"type"`
+
+	// Message when Type is ERR, WARN, DEBUG or INFO
+	Message string `json:"message,omitempty"`
+}
+
+// A user would have to implement their own event type as well, therefore not using this from ig repo
+type traceOpenEvent struct {
+	Event
+	eventtypes.WithMountNsID
```
mauriciovasquezbernal, line 43: Image based gadgets don't base their events on these types. I think it's better to define the mount namespace, (and other interesting fields directly on this structure. As done for other fields below.

pkg/test-fw/tests/trace_open_test.go
```diff
@@ -0,0 +1,122 @@
...
+	Gid      uint32      `json:"gid" column:"gid,template:gid,hide"`
+	Comm     string      `json:"comm,omitempty" column:"comm,maxWidth:16"`
+	Ret      int         `json:"ret,omitempty" column:"ret,width:3,fixed,hide"`
+	Err      int         `json:"err,omitempty" column:"err,width:3,fixed"`
+	Flags    int         `json:"flags,omitempty" column:"flags,width:24,hide"`
+	FlagsRaw int32       `json:"flagsRaw,omitempty"`
+	Mode     int         `json:"mode,omitempty" column:"mode,width:10,hide"`
+	ModeRaw  fs.FileMode `json:"modeRaw,omitempty"`
+	FName    string      `json:"fname,omitempty" column:"path,minWidth:24,width:32"`
+	FullPath string      `json:"fullPath,omitempty" column:"fullPath,minWidth:24,width:32" columnTags:"param:full-path"`
```
mauriciovasquezbernal, line 56: All columns tags can be removed. (I think we can remove `omitempty` as well from all fields).

pkg/test-fw/tests/trace_open_test.go
```diff
@@ -0,0 +1,122 @@
...
+// limitations under the License.
+
+package tests
+
+import (
+	"io/fs"
+	"testing"
+
+	"github.com/inspektor-gadget/inspektor-gadget/integration"
+	IG "github.com/inspektor-gadget/inspektor-gadget/pkg/test-fw/ig"
```
mauriciovasquezbernal, line 22: nit: Usually package names are lower case. But `ig` seems too generic, we need to find a better naming.

pkg/test-fw/tests/trace_open_test.go
mauriciovasquezbernal, line 1: I'd love to have this file together with the gadget code. i.e. in `gadgets/trace_open/`.

mauriciovasquezbernal (commented):
Some initial comments.

pkg/test-fw/ig/command.go
```diff
@@ -0,0 +1,94 @@
...
+	IG
+
+	// Name of the command to be run, used to give information.
+	Name string
+
+	// ExpectedString contains the exact expected output of the command.
+	ExpectedString string
+
+	// ExpectedRegexp contains a regex used to match against the command output.
+	ExpectedRegexp string
```
mauriciovasquezbernal, line 35: I think we can get rid of these two things. `ValidateOutput` is generic enough and should be the only way we provide. If an user wants to validate against an exact string or regexp they can do something like:

```go
ValidateOutput: func(t *testing.T, output string) {
			require.Equal(t, "foo", output)
		},
```

We can even provide a helper like this:

```go
func MatchStringFunc(t *testing.T, expected string) func(t *testing.T, output string) {
	return func(t *testing.T, output string) {
		require.Equal(t, expected, output)
	}
}
```

to be used like:
```go
ValidateOutput: MatchStringFunc(t, "foo"),
```

(Same idea applies to the regexp case.

pkg/test-fw/ig/teststeps.go
```diff
@@ -0,0 +1,110 @@
...
+	// Start starts the step and immediately returns, it does wait until
+	// its completion, use Stop() for that.
+	Start(t *testing.T)
+
+	// Stop stops the step and waits its completion.
+	Stop(t *testing.T)
+
+	// IsCleanup returns true if the step is used to clean resource and
+	// should not be skipped even if previous commands failed.
+	IsCleanup() bool
```
mauriciovasquezbernal, line 39: I think when we implemented these tests, we weren't aware of [`t.Cleanup`](https://pkg.go.dev/testing#T.Cleanup), I think the can use it instead of this `IsCleanup()` logic.

(In any case, I think it's not used in this PR, so I think we can remove it).

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+
+// New creates a new IG configured with the options passed as parameters.
+// Supported parameters are:
+//
+//	WithImage(gadget_image)
+//	WithPath(string)
+//	WithEnv(...string)
+//	WithFlags(...string)
+//	WithStartAndStop()
+func New(opts ...option) (*IG, error) {
```
mauriciovasquezbernal, line 252: As this is only designed for testing purposes, I think this method should take a testing.T object and call t.Fail (or similar) instead of returning errors. (This will remove a lot of boilerplate error handling code around).

pkg/test-fw/tests/trace_open_test.go
pawarpranav83: igw? wrapper?

mauriciovasquezbernal: I'd go for `igrunner`.

pkg/test-fw/ig/command.go
pawarpranav83: @mauriciovasquezbernal Should we fail the test if ValidateOutput func is nil (not specified by the user), or simply return a log statement?

mauriciovasquezbernal: I think nothing special should happen in that case. It's possible that users only want to run a command without worrying about their output.

gadgets/trace_open/trace_open_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func TestTraceOpen(t *testing.T) {
+	cn := "test-trace-open"
+	containerFactory, err := integration.NewContainerFactory("docker")
+	require.NoError(t, err, "new container factory")
+
+	ig := igrunner.New(
+		igrunner.WithPath("ig"),
+		igrunner.WithImage("ghcr.io/inspektor-gadget/gadget/trace_open:latest"),
```
alban, line 64: I wonder if igrunner should also provide a `WithImageBuildFromPath($GADGET_DIR)`

Then, the test can first build the image with `ig image build -t $TEMP_NAME $GADGET_DIR` and then use the gadget previously built.

The test can get `$GADGET_DIR` with something like:
```
  _, filename, _, _ := runtime.Caller(0)
  gadgetDir := path.Dir(filename)
```

Maybe not in this PR.

alban (commented):
Thanks! Some initial comments.

pawarpranav83:
Regarding the --timeout flag, simply specifying it won't work since the control flow is blocked till the `ig.cmd` completes, so we can run the container command asynchronously in another go routine (not sure about the error handling of this).
Something like:
```
ig := igrunner.New(
		igrunner.WithPath("ig"),
		igrunner.WithImage("ghcr.io/inspektor-gadget/gadget/trace_open:latest"),
		igrunner.WithFlags("--runtimes=docker", "-o=json", "--timeout=5"),
)

go command.RunTestSteps([]command.TestStep{containerFactory.NewContainer(cn, "setuidgid 1000:1111 cat /dev/null")}, t)
command.RunTestSteps([]command.TestStep{traceOpenCmd}, t)
```
we can shift this logic inside `teststeps.go` and implement it with wait groups, but I think this would complicate it even more.

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,244 @@
...
+	"os/exec"
+	"syscall"
+
+	"github.com/blang/semver"
+)
+
+type IG struct {
+	path    string
+	image   string
+	version semver.Version
```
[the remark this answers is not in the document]
pawarpranav83: Not yet.

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// To be able to kill the process of ig.cmd,
+	// we need to send the termination signal to their process group ID
+	// (PGID). However, child processes get the same PGID as their parents by
+	// default, so in order to avoid killing also the integration tests process,
+	// we set the fields Setpgid and Pgid of syscall.SysProcAttr before
+	// executing ig.cmd. Doing so, the PGID of ig.cmd (and its children)
+	// will be set to its process ID, see:
+	// https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/syscall/exec_linux.go;l=32-34.
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: 0}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: @alban there is more context on this change on https://github.com/inspektor-gadget/inspektor-gadget/pull/554.
But I agree it's not relevant for this case. (This command is only executing  `ig` and there is not `sh` involved.

pkg/testing/command/command.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"testing"
+
+	igrunner "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/ig"
+	"github.com/stretchr/testify/require"
+)
+
+type Command struct {
+	// IG wrapper
+	igrunner.IG
```
mauriciovasquezbernal, line 26: I think we can have this relation in the other way around: Command to be a generic way or running commands, and `igrunner` a wrapper around it focused on ig.

Something like https://github.com/inspektor-gadget/inspektor-gadget/commit/8f44104dc2b2835c05fdaea763aaaac9560f2f81

pkg/testing/ig/ig.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's a good idea to keep it after my comment on https://github.com/inspektor-gadget/inspektor-gadget/pull/2607#discussion_r1534726092. If `Command` is something generic, we want to be sure `Kill()` kills all its children.

mauriciovasquezbernal:
> Regarding the --timeout flag, simply specifying it won't work since the control flow is blocked till the ig.cmd completes, so we can run the container command asynchronously in another go routine (not sure about the error handling of this).

I think it'll require to change the order of the commands:
- execute the container: It should generate a events in a loop (And use StartAndStop)
- execute the tracer


Something like this worked for me:

```bash
diff --git gadgets/trace_open/trace_open_test.go gadgets/trace_open/trace_open_test.go
index 8b30f5eb9..609b6b77c 100644
--- gadgets/trace_open/trace_open_test.go
+++ gadgets/trace_open/trace_open_test.go
@@ -63,8 +63,8 @@ func TestTraceOpen(t *testing.T) {
        traceOpenCmd := igrunner.New(
                igrunner.WithPath("ig"),
                igrunner.WithImage("ghcr.io/inspektor-gadget/gadget/trace_open:latest"),
-               igrunner.WithFlags("--runtimes=docker", "-o=json"),
-               igrunner.WithStartAndStop(),
+               igrunner.WithFlags("--runtimes=docker", "-o=json", "--timeout=5"),
+               //igrunner.WithStartAndStop(),
                igrunner.WithValidateOutput(func(t *testing.T, output string) {
                        expectedEntry := &traceOpenEvent{
                                Event: Event{
@@ -101,8 +101,8 @@ func TestTraceOpen(t *testing.T) {
        )

        testSteps := []command.TestStep{
+               containerFactory.NewContainer(cn, "while true; do setuidgid 1000:1111 cat /dev/null; sleep 0.1; done", integration.WithStartAndStop()),
                traceOpenCmd,
-               containerFactory.NewContainer(cn, "setuidgid 1000:1111 cat /dev/null"),
        }

        command.RunTestSteps(testSteps, t)
```

gadgets/trace_open/trace_open_test.go
```diff
@@ -0,0 +1,112 @@
...
+	FullPath  string      `json:"fullPath"`
+}
+
+func TestTraceOpen(t *testing.T) {
+	cn := "test-trace-open"
+	containerFactory, err := integration.NewContainerFactory("docker")
+	require.NoError(t, err, "new container factory")
+
+	ig := igrunner.New(
+		igrunner.WithPath("ig"),
```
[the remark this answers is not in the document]
pawarpranav83: Which CLI flag?

alban (commented):
Thanks. Some additional comments below.

TODO for this PR:
- [ ] Section in docs/devel/hello-world-gadget.md explaining how to add tests in an external gadget

TODO for later:
- [ ] PR in https://github.com/inspektor-gadget/gadget-template adding tests.

pkg/test-fw/ig/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+func parseMultiJSONOutput[T any](t *testing.T, output string, normalize func(*T)) []*T {
+	ret := []*T{}
+
+	decoder := json.NewDecoder(strings.NewReader(output))
+	for decoder.More() {
+		var entry T
+		if err := decoder.Decode(&entry); err != nil {
+			require.NoError(t, err, "decoding json")
```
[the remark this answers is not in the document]
alban: The error should already be printed by `require.NoError()`:
https://github.com/stretchr/testify/blob/edd74b24a1751990eeb831ebc70902c36600c3af/assert/assertions.go#L1549

.github/workflows/inspektor-gadget.yml
```diff
@@ -1282,6 +1282,34 @@ jobs:
...
+    - uses: actions/checkout@v4
+    - name: Get ig-linux-amd64.tar.gz from artifact.
+      uses: actions/download-artifact@v3
+      with:
+        name: ig-linux-amd64-tar-gz
+        path: /home/runner/work/inspektor-gadget/
+    - name: Unpack ig-linux-amd64.tar.gz
+      run: |
+        tar zxvf /home/runner/work/inspektor-gadget/ig-linux-amd64.tar.gz
+        sudo mv ig /usr/bin/ig
```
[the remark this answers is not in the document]
pawarpranav83: Build-ig builds ig and adds it as an artifact.
```
- name: Build ig-linux-${{ matrix.platform }}
      run: |
        if [ "${{ matrix.platform }}" = 'arm64' ]; then
          sudo apt-get update
          sudo apt-get install qemu-user-static
        fi

        make EBPF_BUILDER=${{ needs.build-helper-images.outputs.ebpf_builder_image }} ig-linux-${{ matrix.platform }}

        # Prepare assets for release and actions artifacts
        mkdir ${{ matrix.platform }}
        cp ig-linux-${{ matrix.platform }} ${{ matrix.platform }}/ig
        cp LICENSE ${{ matrix.platform }}/
        tar --sort=name --owner=root:0 --group=root:0 \
          -czf ig-linux-${{ matrix.platform }}.tar.gz -C ${{ matrix.platform }} \
          ig LICENSE
        rm -rf ${{ matrix.platform }}
    - name: Add ig-linux-${{ matrix.platform }}.tar.gz as artifact.
      uses: actions/upload-artifact@v3
      with:
        name: ig-linux-${{ matrix.platform }}-tar-gz
        path: /home/runner/work/inspektor-gadget/inspektor-gadget/ig-linux-${{ matrix.platform }}.tar.gz
```

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+	"os/exec"
+	"syscall"
+
+	"github.com/blang/semver"
+)
+
+type IG struct {
+	path    string
+	image   string
+	version semver.Version
```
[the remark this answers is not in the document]
pawarpranav83: These fields are only defined in this struct. Can u provide a reference for the other struct.

pkg/test-fw/ig/ig.go
```diff
@@ -0,0 +1,273 @@
...
+	cmd := exec.Command(path, "version")
+
+	var out bytes.Buffer
+	cmd.Stdout = &out
+
+	if err := cmd.Run(); err != nil {
+		return "", err
+	}
+	return out.String(), nil
+}
```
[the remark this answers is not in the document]
pawarpranav83: I don't understand. This function is showing diff add just because ig.go was moved, if u meant that this function is added again in this commit.

pkg/test-fw/ig/ig.go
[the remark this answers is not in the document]
pawarpranav83: That's the same struct, which is extended with additional fields in later commits.
Should I squash the commits so that there is no confusion?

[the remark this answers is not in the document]
pawarpranav83: Oh right, the same file was moved multiple times. Currently, it's in `pkg/testing/ig/ig.go`.

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,136 @@
...
+//	WithCommandName(name string)
+func New(opts ...option) *IG {
+	ig := &IG{
+		path: "ig",
+	}
+
+	for _, opt := range opts {
+		opt(ig)
+	}
+
```
burak-ok, line 129: We could check if `ig.Name` was ever set. If not we can set it to `ig.image` here maybe?

pkg/testing/match/match.go
```diff
@@ -0,0 +1,94 @@
...
+	"encoding/json"
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func parseMultiJSONOutput[T any](t *testing.T, output string, normalize func(*T)) []*T {
```
burak-ok, line 27: You could remove the functions that appear in this file from `integration/helpers.go` and include this package there. Would avoid code duplication.
Best would be a separate commit for this before the other changes

pkg/testing/command/command.go
```diff
@@ -0,0 +1,188 @@
...
+	c.Cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true, Pgid: 0}
+}
+
+// verifyOutput verifies if the stdout match with the expected regular expression and the expected
+// string. If it doesn't, verifyOutput makes the test fail.
+func (c *Command) verifyOutput(t *testing.T) {
+	output := c.stdout.String()
+
+	if c.ValidateOutput != nil {
+		c.ValidateOutput(t, output)
```
burak-ok, line 85: nit. Since after your last changes the `output` variable is not needed anymore outside the `if` condition 

```suggestion
		c.ValidateOutput(t, c.stdout.String())
```

pkg/testing/command/teststeps.go
```diff
@@ -0,0 +1,92 @@
...
+func WithCbBeforeCleanup(f func(t *testing.T)) func(opts *runTestStepsOpts) {
+	return func(ops *runTestStepsOpts) {
+		ops.cbBeforeCleanup = f
+	}
+}
+
+// RunTestSteps is used to run a list of test steps with stopping/clean up logic.
+// executeBeforeCleanup is executed before calling the cleanup functions, it can be use for instance
+// to print extra logs when the test fails.
+func RunTestSteps[S TestStep](steps []S, t *testing.T, options ...Option) {
```
blanquicet, line 60: While reviewing this, I realized that we don't need to use a type parameter here but interface type is enough. I opened https://github.com/inspektor-gadget/inspektor-gadget/pull/2646 to manage it.

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
blanquicet: > Are those steps necessary? Wouldn't the `ig` file already placed on the filesystem after the `build-ig` job? I see some other jobs follow the same pattern, so I am probably wrong, but I'd like to understand.

It's because each job runs in a different runner environment, see [documentation](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#overview). So, they don't share the filesystem.

gadgets/trace_open/trace_open_test.go
blanquicet: Yep, I would suggest to manage it in another PR.

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
blanquicet: > Are those steps necessary? Wouldn't the `ig` file already placed on the filesystem after the `build-ig` job? I see some other jobs follow the same pattern, so I am probably wrong, but I'd like to understand.

It's because each job runs in a different runner environment, see [documentation](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#overview). So, they don't share the filesystem.

pkg/test-fw/ig/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+		str.WriteString("captured:\n")
+		for _, entry := range entries {
+			entryJson, _ := json.Marshal(entry)
+			str.WriteString(string(entryJson))
+			str.WriteString("\n")
+		}
+		expectedEntryJson, _ := json.Marshal(expectedEntry)
+		str.WriteString("expected:\n")
+		str.WriteString(string(expectedEntryJson))
+		t.Fatal(str.String())
```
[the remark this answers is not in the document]
blanquicet: Mmmmm this is just to print the captured and expected output in the same format. The actual diff is done above with `reflect.DeepEqual`.

pkg/testing/ig/ig.go
blanquicet: Default values should go in the definition of the struct above.

On the other side, I'm not sure developers need to configure the command name. If we consider they might, so I suggest using a default like Burak suggested, something like `"Run"+"<ig.image>"`.

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,136 @@
...
+func WithPathFromEnvVar() option {
+	return func(ig *IG) {
+		ig.path = os.Getenv("IG")
+	}
+}
+
+func WithImage(image string) option {
+	return func(ig *IG) {
+		ig.image = fmt.Sprintf("%s/%s:%s", os.Getenv("GADGET_REPOSITORY"), image, os.Getenv("GADGET_TAG"))
```
blanquicet, line 61: IG will already add the default repo and tag if they are not present, don't need to do it here.

blanquicet (commented):
Great work also here. I can already see a follow-up PR from you further abstracting this for ig-k8s and kubectl-gadget as you're doing in https://github.com/inspektor-gadget/inspektor-gadget/pull/2450.

pkg/testing/ig/ig.go
burak-ok: > Default values should go in the definition of the struct above.

Indeed. But there we do not have the information about `ig.image` yet

pkg/testing/match/match.go
pawarpranav83: Maybe I should change these functions in a different PR? (mostly the built-in to image-based gadgets test conversion PR).

pkg/test-fw/ig/helpers.go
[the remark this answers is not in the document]
pawarpranav83: > I would like to have diff based output.

@eiffel-fl It only works when we have a single expected entry or string based output.
Since mostly we have a single expected entry we can add something like:
```
if len(entries) == 1 {
	t.Fatalf("Entry doesn't match:\n%s",
		cmp.Diff(expectedEntry, &entries[0]))
}
```

pkg/testing/ig/ig.go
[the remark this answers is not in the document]
pawarpranav83: Should I remove the version logic since we aren't using it (the ig features and flags used are already supported by the last three versions, so should we check for the version before using them?)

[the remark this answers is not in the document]
pawarpranav83: Should I remove the version logic? (since most of the features and flags are supported from the last five or six versions ig, and since we mostly install it via the repo itself, the version comes out to be 0.0.0, so it would give an error)

pkg/testing/command/command.go
```diff
@@ -0,0 +1,187 @@
...
+	"bytes"
+	"errors"
+	"os/exec"
+	"syscall"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type Command struct {
```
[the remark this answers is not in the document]
pawarpranav83: We will later replace integration logic with these files, while also converting the built-in gadgets test to image-based gadgets (that would be in another PR).

docs/devel/hello-world-gadget.md
```diff
@@ -550,6 +550,127 @@ RUNTIME.CONTAINERNAME  PID          COMM         FILENAME
...
+
+```golang
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+  // helper functions for creating and running commands in a container.
```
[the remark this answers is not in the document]
pawarpranav83: This is weird, the comments are aligned in my local editor but with the preview of the file and remote branch commit it's not :).

gadgets/trace_open/test/trace_open_test.go
```diff
@@ -0,0 +1,116 @@
...
+	Comm      string      `json:"comm"`
+	Ret       int         `json:"ret"`
+	Err       int         `json:"err"`
+	Flags     int         `json:"flags"`
+	FlagsRaw  int32       `json:"flagsRaw"`
+	Mode      int         `json:"mode"`
+	ModeRaw   fs.FileMode `json:"modeRaw"`
+	FName     string      `json:"fname"`
+	FullPath  string      `json:"fullPath"`
+}
```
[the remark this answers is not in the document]
pawarpranav83: I think @mauriciovasquezbernal suggested this would be a better approach, I can't recall the reasoning for the same.
The mapping approach is already used by `run*` tests in `integration/inspektor-gadget`.

pkg/testing/command/command.go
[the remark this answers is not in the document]
pawarpranav83: The testing package would be used by users with custom gadgets to implement their tests and we would also replace the current integration tests, where instead of using built-in gadgets (most tests use these) we would use the gadget images, some already exist: `run*` tests in `integration/inspektor-gadget`.

gadgets/trace_open/test/trace_open_test.go
blanquicet, line 57: This contains the fields built-in gadget provides, but the image-based gadget doesn't provide them all yet. Please use the image-based gadget as the reference:

https://github.com/inspektor-gadget/inspektor-gadget/blob/49c92a9303538b9304014188b789e5703c26e5dd/gadgets/trace_open/program.bpf.c#L24-L36

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,113 @@
...
+
+// New creates a new IG configured with the options passed as parameters.
+// Supported parameters are:
+//
+//	WithImage(gadget_image)
+//	WithPath(string)
+//	WithFlags(...string)
+//	WithStartAndStop()
+//	WithValidateOutput(validateOutput func(t *testing.T, output string))
+//	WithCommandName(name string)
```
blanquicet, line 98: `WithCommandName` doesn't exist anymore. I suggest removing this list as it's difficult to keep it updated.

gadgets/trace_open/test/trace_open_test.go
```diff
@@ -0,0 +1,116 @@
...
+package tests
+
+import (
+	"flag"
+	"io/fs"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/inspektor-gadget/inspektor-gadget/integration"
```
blanquicet, line 24: The idea is to use only the `testing` pkg and not import things from `integration`. So, we also need to move the [containers pkg](https://github.com/inspektor-gadget/inspektor-gadget/pull/2666) within the `testing` pkg (not actually move but copy for now but the same you did for `command` and `match`).

gadgets/trace_open/test/trace_open_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/command"
```
blanquicet, line 25: It's strange that the test imports the `command` pkg if this doesn't use commands directly but through the `igrunner`. In this case, it is because we have the `teststeps.go` in the same pkg. By moving it out as I suggested in a previous comment, this import will be gone.

gadgets/trace_open/test/trace_open_test.go
```diff
@@ -0,0 +1,116 @@
...
+	Err       int         `json:"err"`
+	Flags     int         `json:"flags"`
+	FlagsRaw  int32       `json:"flagsRaw"`
+	Mode      int         `json:"mode"`
+	ModeRaw   fs.FileMode `json:"modeRaw"`
+	FName     string      `json:"fname"`
+	FullPath  string      `json:"fullPath"`
+}
+
+var gadgetTest = flag.Bool("gadgets-test", false, "run gadget tests")
```
blanquicet, line 59: I'm wondering how we can avoid doing this for all gadgets 🤔 ... Probably creating a pkg in `gadgets/testing` that defines this flag and implements a `RunGadgetTests` function that checks the value. Something like this:

`gadgets/testing/testing.go` (Not in `pkg/testing` because this is something we don't want to expose to users but it's an internal requirement for the way we run our tests from the makefiles/ci):
```go
package testing

import "flag"

var gadgetTest = flag.Bool("gadgets-test", false, "run gadget tests")

func RunGadgetTests() bool {
	return *gadgetTest
}
```

`gadgets/trace_open/test/trace_open_test.go`:
```go
import (
	...
	. "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
	...
)

func TestTraceOpen(t *testing.T) {
	if !RunGadgetTests() {
		t.Skip()
	}
	...
}
```
 
Anyway, let's manage this in another PR where we add other tests.

@alban, any comment here?

blanquicet (commented):
I think it's almost ready to go. I still need to check the documentation but it looks already fine.

gadgets/trace_open/trace_open_test.go
blanquicet: Actually, now that I reviewed the CI part, I'm not sure about it anymore.

I realized the `image-based-gadgets-test` job is building all the gadgets, but `build-and-push-gadgets` job is also building and pushing them.

In order to verify the push and pull features, `image-based-gadgets-test` should use the images built and pushed by `build-and-push-gadgets`. However, it makes sense only for IG's CI because we want to test the push/pull features. On the contrary, in the [gadget-template](https://github.com/inspektor-gadget/gadget-template)'s CI, we should apply @alban's suggestion and build+test things locally.

blanquicet (commented):
Few comments about the CI.

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,244 @@
...
+	// No need to kill, command has not been executed yet or it already exited
+	if ig.cmd == nil || (ig.cmd.ProcessState != nil && ig.cmd.ProcessState.Exited()) {
+		return nil
+	}
+
+	// Given that we set Setpgid, here we just need to send the PID of ig.cmd
+	// (which is the same PGID) as a negative number to syscall.Kill(). As a
+	// result, the signal will be received by all the processes with such PGID,
+	// in our case, the process of ig.cmd.
+	err := syscall.Kill(-ig.cmd.Process.Pid, sig)
```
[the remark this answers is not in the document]
blanquicet: Things were updated to make `igrunner` a wrapper of `command` so that it reuses all the `command`'s logic for killing the process, which is generic for any command.

gadgets/trace_open/test/trace_open_test.go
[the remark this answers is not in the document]
burak-ok: > In the end, the goal of this type is to be translated to JSON to the be compared to the JSON output. I am wondering if we really need to define a type and if we should not rather use a map or JSON directly.

When I implemented the first test for the run gadget I had the following issue:
When decoding a JSON string into a `map[string]any` all numbers were of type `float32`. So in the test you would need to check the `Ret` variable with something like `jsonMap["Ret"] == 0.0f` or so.

gadgets/Makefile
```diff
@@ -1,9 +1,11 @@
 SHELL := /bin/bash
 
-GADGET_TAG ?= $(shell ../tools/image-tag branch)
-GADGET_REPOSITORY ?= ghcr.io/inspektor-gadget/gadget
 BUILDER_IMAGE ?= ghcr.io/inspektor-gadget/ebpf-builder:latest
-IG ?= ig
+
+export GADGET_TAG ?= $(shell ../tools/image-tag branch)
+export GADGET_REPOSITORY ?= ghcr.io/inspektor-gadget/gadget
+export IG ?= ig
```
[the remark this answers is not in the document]
blanquicet: > Do you mean the export keyword?
>
Yes, but I meant that they are not needed because we are not using those env variables inside the tests anymore. We are now passing the image name "trace_open" and letting `ig` add the default repo and tag (this way, we also test such a feature).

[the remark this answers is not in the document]
blanquicet: However, for `IG` it's still needed, so I also vote for the second option as we are doing for `KUBECTL_GADGET`:

https://github.com/inspektor-gadget/inspektor-gadget/blob/abd4dfda26fec0def8735bc2f1de21b9c51c84c2/Makefile#L276-L288

gadgets/trace_open/test/trace_open_test.go
blanquicet: > Why do we need this option -gadgets-test in the first place?
>
I guess Pranav get inspiration from what we do for `integration-tests`:

Call:

https://github.com/inspektor-gadget/inspektor-gadget/blob/abd4dfda26fec0def8735bc2f1de21b9c51c84c2/Makefile#L280

Usage:

https://github.com/inspektor-gadget/inspektor-gadget/blob/abd4dfda26fec0def8735bc2f1de21b9c51c84c2/integration/inspektor-gadget/integration_test.go#L102-L105

But, I'm fine with the `os.Getenv("IG")` approach. It confirmed in [this discussion](https://github.com/inspektor-gadget/inspektor-gadget/pull/2607#discussion_r1548093241) that we should avoid exporting things in the Makefile and only pass env variables to the commands that use them.

docs/devel/hello-world-gadget.md
```diff
@@ -550,6 +550,127 @@ RUNTIME.CONTAINERNAME  PID          COMM         FILENAME
...
+func TestMyGadget(t *testing.T) {
+	cn := "test-mygadget"
+
+  	// returns a container manager which implements an interface with methods for creating new container
+  	// and running commands within that container.
+	containerFactory, err := integration.NewContainerFactory("docker")
+	require.NoError(t, err, "new container factory")
+
+	mygadgetCmd := igrunner.New(
+    		// gadget repository and tag are added via env vars
```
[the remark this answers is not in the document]
blanquicet: Actually, this comment is not up-to-date anymore as mentioned in https://github.com/inspektor-gadget/inspektor-gadget/pull/2607#discussion_r1548088438.

blanquicet (commented):
I saw the new section in the documentation, and it'll be fine after it's updated with recent changes and you manage the comments.

So, in general, this PR seems ready to me once all comments are managed.

gadgets/trace_open/test/trace_open_test.go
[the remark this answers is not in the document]
blanquicet: Please also consider [this](https://github.com/inspektor-gadget/inspektor-gadget/pull/2607#discussion_r1547713294).

docs/devel/hello-world-gadget.md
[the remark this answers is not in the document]
pawarpranav83: Should I remove this whole comment and explain that gadget repo and tag are added automatically by IG on this [line](https://github.com/inspektor-gadget/inspektor-gadget/pull/2607/files#diff-5707bfb4f5107fcae74a123973315fe44d85c9b92215454ecfed0085ddc669f5R661)?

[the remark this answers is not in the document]
pawarpranav83: Since IG and not ig wrapper adds repo and tag to the gadget name, this difference does not affect the users so, we should still let them know that these variables need to be defined,

docs/devel/hello-world-gadget.md
```diff
@@ -550,6 +550,127 @@ RUNTIME.CONTAINERNAME  PID          COMM         FILENAME
...
+						},
+					},
+					Comm:     "cat",
+					Filename: "/dev/null",
+					Uid:      1000,
+					Gid:      1111,
+				}
+
+        			// used to "normalize" the output so that it only includes non-default values
+        			// for the fields we are able to verify.
```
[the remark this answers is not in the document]
pawarpranav83: Something like: used to "normalize" the output, sets random value fields to a default value so that it only includes non-default values for the fields we can verify?

pkg/testing/command/command.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2019-2024 The Inspektor Gadget authors
```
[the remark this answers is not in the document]
alban: @pawarpranav83 I think @eiffel-fl didn't mean to add a comment in this file but to edit the commit message (`git commit --amend` or `git rebase...`+'reword').  You can look at `git log` for example. Something like:
```
CI, docs, testing pkg fix

pkg/testing/command/command.go: code taken from integration/command.go at commit 18d52d8e057d ("Merge pull request #2605 from inspektor-gadget/citest/alban_pcomm")

Signed-off-by: ...
```
(please correct me if I'm wrong)

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,105 @@
...
+func WithPath(path string) option {
+	return func(ig *IGTestConfiguration) {
+		ig.path = path
+	}
+}
+
+// WithPathFromEnvVar used for reading custom path for ig executable from IG env var.
+func WithPathFromEnvVar() option {
+	return func(ig *IGTestConfiguration) {
+		ig.path = os.Getenv("IG")
```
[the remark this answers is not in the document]
pawarpranav83: This was already used in `gadgets/Makefile`, what do you suggest instead?

.github/workflows/inspektor-gadget.yml
```diff
@@ -1342,6 +1342,55 @@ jobs:
...
+    #level 2
+    needs:
+      - build-ig
+      - build-helper-images
+    runs-on: ubuntu-latest
+    permissions:
+      # allow publishing container image
+      # in case of public fork repo/packages permissions will always be read
+      contents: read
+      packages: write
```
pawarpranav83, line 1356: CI passes only when we provide the write permission, even though we are not publishing the images to the remote repository, we are building them locally, I would like to understand why this is required.

gadgets/trace_open/test/trace_open_test.go
pawarpranav83: It would be better to make this change in this PR itself, will try to update the test in test commit itself.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1342,6 +1342,55 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Login to Container Registry
```
mqasimsarfraz, line 1359: If we don't plan to push gadgets we shouldn't be logging in? CI might fail because of that.

.github/workflows/inspektor-gadget.yml
pawarpranav83: https://github.com/pawarpranav83/inspektor-gadget/actions/runs/8575164865/job/23503783080

.github/workflows/inspektor-gadget.yml
pawarpranav83: CI was failing till we provided this step: 
https://github.com/pawarpranav83/inspektor-gadget/actions/runs/8567763465/job/23480415618.
And after rebasing the read and write permissions were required to pass the test.

pawarpranav83:
Workflow approval @mqasimsarfraz.

.github/workflows/inspektor-gadget.yml
pawarpranav83: Yeah, you are right, the test doesn't work when directly specifying the value of the gadget repo and tag.
https://github.com/pawarpranav83/inspektor-gadget/actions/runs/8601351504/job/23568643301

I think the gadget-image doesn't exist, but still its running the test, and no output is captured.
Will try to test this.

pawarpranav83: Error: 
```
~/Desktop/Repos/inspektor-gadget/gadgets$ make test
CGO_ENABLED=0 \
IG_EXPERIMENTAL=true \
IG=ig \
go test -exec 'sudo -E' -v ./...
?       github.com/inspektor-gadget/inspektor-gadget/gadgets/testing    [no test files]
=== RUN   TestTraceOpen
    command.go:113: Start command(Run_trace_open)
    docker.go:117: Container "test-trace-open" output:
    command.go:130: Stop command(Run_trace_open)
    command.go:132: Command returned(Run_trace_open):
        time="2024-04-08T19:58:08+05:30" level=info msg="Experimental features enabled"
        
        
    match.go:71: output doesn't contain the expected entry
        captured:
        expected:
        {"runtime":{"runtimeName":"docker","containerName":"test-trace-open"},"k8s":{},"mountnsid":0,"pid":0,"uid":1000,"gid":1111,"comm":"cat","ret":3,"flags":0,"mode":0,"fname":"/dev/null"}
--- FAIL: TestTraceOpen (15.69s)
FAIL
FAIL    github.com/inspektor-gadget/inspektor-gadget/gadgets/trace_open/test    15.732s
FAIL
make: *** [Makefile:57: test] Error 1
```
Removed the image and skipped the build test, with no image the test is still running, and no output is displayed.

pawarpranav83: But if that image doesn't exist it does give an error, so its not about image not existing ig: 
```
~/Desktop/Repos/inspektor-gadget/gadgets$ make test
CGO_ENABLED=0 \
IG_EXPERIMENTAL=true \
IG=ig \
go test -exec 'sudo -E' -v ./...
?       github.com/inspektor-gadget/inspektor-gadget/gadgets/testing    [no test files]
=== RUN   TestTraceOpen
    command.go:113: Start command(Run_nice)
    docker.go:117: Container "test-trace-open" output:
    command.go:130: Stop command(Run_nice)
    command.go:132: Command returned(Run_nice):
        time="2024-04-08T20:18:25+05:30" level=info msg="Experimental features enabled"
        Error: getting gadget info: getting gadget image: ensuring image presence: pulling image (if missing) "nice": downloading to local repository: failed to resolve ghcr.io/inspektor-gadget/gadget/nice:latest: GET "https://ghcr.io/v2/inspektor-gadget/gadget/nice/manifests/latest": GET "https://ghcr.io/token?scope=repository%3Ainspektor-gadget%2Fgadget%2Fnice%3Apull&service=ghcr.io": response status code 403: denied: requested access to the resource is denied
        
        
    command.go:134: 
                Error Trace:    /home/jinx83/Desktop/Repos/inspektor-gadget/pkg/testing/command/command.go:134
                                                        /home/jinx83/Desktop/Repos/inspektor-gadget/pkg/testing/teststeps.go:78
                                                        /home/jinx83/Desktop/Repos/inspektor-gadget/pkg/testing/teststeps.go:92
                                                        /home/jinx83/Desktop/Repos/inspektor-gadget/gadgets/trace_open/test/trace_open_test.go:92
                Error:          Received unexpected error:
                                exit status 1
                Test:           TestTraceOpen
                Messages:       failed to kill command(Run_nice)
--- FAIL: TestTraceOpen (13.78s)
FAIL
FAIL    github.com/inspektor-gadget/inspektor-gadget/gadgets/trace_open/test    13.822s
FAIL
make: *** [Makefile:57: test] Error 1
```

pawarpranav83: The gadget is working fine, but since we are using StartAndStop, the command is started and when no gadget is found, it pulls the gadget from the remote repo, but meanwhile we run the container command till the gadget is run that command is missed so therefore, it shows no output.
But it still doesn't answer why it is pulling the image if we are building it locally.

pawarpranav83: Since the default value of `GADGET_REPOSITORY=ghcr.io/inspektor-gadget/gadget` and `GADGET_TAG=latest`, in case of local CI repo value was `ghcr.io/pawarpranav83/gadget` and tag value as `temp-fw` instead, which led to pulling the image from remote repository since we don't pass these env vars to make test (we rely on the feature of ig instead where it adds the prefix and suffix to gadget name automatically, and these are the default values).

Solution: We can manually set the image name with GADGET_REPOSITORY and GADGET_TAG environment variables, but since in the main branch CI, images are tagged with default values, it would use the local gadget-images.

blanquicet: Thanks for describing the issue with the `GADGET_REPOSITORY` and `GADGET_TAG`, it's clear now.

I agree we should pass the `GADGET_REPOSITORY` and `GADGET_TAG` values to the test and avoid letting IG add the default repo and tag (Sorry, I think I suggested doing that). 

Now, as I mentioned [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2607#discussion_r1545860499), I still consider the `test-gadgets` job shouldn't build the gadgets again but use the ones built and pushed by the `build-and-push-gadgets` job. 

On the other hand, you can solve the issue of IG not capturing the event because it's busy pulling the image by repeating the command many times. Notice that in the current tests, we use `BusyboxPodRepeatCommand` which executes the command periodically every `0.1` seconds. So, here you could do:

```go
containerFactory.NewContainer(cn, "while true; do setuidgid 1000:1111 cat /dev/null; sleep 0.1; done"),
```

Probably, we should implement the `WithRepeatedCmd(0.1)` option for `NewContainer`. But that's something we can do in another PR.

Makefile
```diff
@@ -380,6 +380,10 @@ build-gadgets: install/ig
 push-gadgets: install/ig
 	$(MAKE) -C gadgets/ push
 
+.PHONY: test-gadgets
+test-gadgets: install/ig
+	$(MAKE) -C gadgets/ test -o build
```
blanquicet, line 385: If you're not building anymore the gadgets, then whoever execute `test-gadgets` needs to ensure the gadgets were already built and pushed. Meaning that the `test-gadgets` job has to be in level 3 and have `build-and-push-gadgets` in its `needs`.

pawarpranav83: Right, I am going to revert this change and directly use the test command, thanks for pointing out the level 3 detail.

blanquicet: Mmmm this change is fine, just need to update the CI job dependecies.

gadgets/trace_open/test/trace_open_test.go
```diff
@@ -0,0 +1,93 @@
...
+					e.Runtime.ContainerImageDigest = ""
+				}
+
+				match.ExpectEntriesToMatch(t, output, normalize, expectedEntry)
+			}),
+	)
+
+	testSteps := []igtesting.TestStep{
+		traceOpenCmd,
+		containerFactory.NewContainer(cn, "setuidgid 1000:1111 cat /dev/null"),
```
blanquicet, line 89: Remember to ensure this command is executed multiple times with the `while` I suggested [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2607#discussion_r1556404269). Otherwise, you risk missing the event because ig is pulling the image.

Makefile
mauriciovasquezbernal: I'm not totally sure this rule should skip building the gadgets. `make tests-gadgets` should do all the work needed to run the tests (even building the gadgets). If we want to avoid rebuilding them in the CI, we can use something like `make -C gadgets/ test -o build` directly from the action workflow.

Makefile
```diff
@@ -412,6 +416,7 @@ help:
 	@echo  '  ig-tests			- Run ig manager unit tests'
 	@echo  '  gadgets-unit-tests		- Run gadget unit tests'
 	@echo  '  integration-tests		- Run integration tests'
+	@echo  '  test-gadgets			- Run image-based gadgets test'
```
mauriciovasquezbernal, line 419:
```suggestion
	@echo  '  test-gadgets			- Run gadgets tests'
```

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: I think such a helper could help. Btw, to use the while loop as you suggested, it's also needed to pass `containers.WithStartStop()`, otherwise the tests keeps running forever.

pkg/testing/match/match.go
mauriciovasquezbernal: I think we can do it right away on this PR.

mauriciovasquezbernal (commented):
It's looking really good! 

I have some minor comments. 

Also, all files on pkg/testing/containers/ are an exact copy of integration/containers/. I think it's better to avoid this specific duplication. We can create a commit that moves the files and updates the users of them. 

Same for match functions, if we can start avoiding duplication of them it'll be great. (For `Command` it's different because it was changed a lot, so making use of it in other places is rather difficult).

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,101 @@
...
+
+// WithValidateOutput used to compare the actual output with expected output.
+func WithValidateOutput(validateOutput func(t *testing.T, output string)) option {
+	return func(ig *IGTestConfiguration) {
+		ig.ValidateOutput = validateOutput
+	}
+}
+
+// New creates a new IG configured with the options passed as parameters.
+func New(image string, opts ...option) *IGTestConfiguration {
```
mauriciovasquezbernal, line 85: It can return `TestStep` and then `IGTestConfiguration` can be an internal type:


```diff
diff --git a/pkg/testing/ig/ig.go b/pkg/testing/ig/ig.go
index d51162030..441dc8de7 100644
--- a/pkg/testing/ig/ig.go
+++ b/pkg/testing/ig/ig.go
@@ -23,11 +23,12 @@ import (
        "os/exec"
        "testing"
 
+       igtesting "github.com/inspektor-gadget/inspektor-gadget/pkg/testing"
        "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/command"
 )
 
-// IGTestConfiguration is responsible for storing configuration of ig executable and provide methods to interact with.
-type IGTestConfiguration struct {
+// runner is responsible for storing configuration of ig executable and provide methods to interact with.
+type runner struct {
        path  string
        image string
 
@@ -36,7 +37,7 @@ type IGTestConfiguration struct {
        flags []string
 }
 
-func (ig *IGTestConfiguration) createCmd() {
+func (ig *runner) createCmd() {
        ig.flags = append(ig.flags, "-o=json")
        args := append([]string{"run", ig.image}, ig.flags...)
        cmd := exec.Command(ig.path, args...)
@@ -44,46 +45,46 @@ func (ig *IGTestConfiguration) createCmd() {
        ig.Cmd = cmd
 }
 
-type option func(*IGTestConfiguration)
+type option func(*runner)
 
 // WithPath used for providing custom path to ig executable.
 func WithPath(path string) option {
-       return func(ig *IGTestConfiguration) {
+       return func(ig *runner) {
                ig.path = path
        }
 }
 
 // WithPathFromEnvVar used for reading custom path for ig executable from IG env var.
 func WithPathFromEnvVar() option {
-       return func(ig *IGTestConfiguration) {
+       return func(ig *runner) {
                ig.path = os.Getenv("IG")
        }
 }
 
 // WithFlags args should be in form: "--flag_name=value" or "-shorthand=value".
 func WithFlags(flags ...string) option {
-       return func(ig *IGTestConfiguration) {
+       return func(ig *runner) {
                ig.flags = flags
        }
 }
 
 // WithStartAndStop used to set StartAndStop value to true.
 func WithStartAndStop() option {
-       return func(ig *IGTestConfiguration) {
+       return func(ig *runner) {
                ig.StartAndStop = true
        }
 }
 
 // WithValidateOutput used to compare the actual output with expected output.
 func WithValidateOutput(validateOutput func(t *testing.T, output string)) option {
-       return func(ig *IGTestConfiguration) {
+       return func(ig *runner) {
                ig.ValidateOutput = validateOutput
        }
 }
 
 // New creates a new IG configured with the options passed as parameters.
-func New(image string, opts ...option) *IGTestConfiguration {
-       ig := &IGTestConfiguration{
+func New(image string, opts ...option) igtesting.TestStep {
+       ig := &runner{
                path:  "ig",
                image: fmt.Sprintf("%s/%s:%s", os.Getenv("GADGET_REPOSITORY"), image, os.Getenv("GADGET_TAG")),
                Command: command.Command{

```

Makefile
pawarpranav83: CI Run: https://github.com/pawarpranav83/inspektor-gadget/actions/runs/8640186743/job/23687887510.
The test is passing locally, any reason why its giving unknown flag error only in CI?

pawarpranav83:
> Also, all files on pkg/testing/containers/ are an exact copy of integration/containers/. I think it's better to avoid this specific duplication. We can create a commit that moves the files and updates the users of them.

Will make these changes once the de-duplication PR is merged, to avoid the merge conflicts.

Makefile
mauriciovasquezbernal: mmm, I think it's because the timeout parameter is temporary gone as stated in https://github.com/inspektor-gadget/inspektor-gadget/issues/2653. @flyth do you know how difficult it's to add that parameter back?

flyth: Should be fairly easy, can take a look at it later today.

mauriciovasquezbernal: @pawarpranav83 perhaps you can switch to StartAndStop as it was before until we add the timeout parameter back?

gadgets/trace_open/test/trace_open_test.go
mauriciovasquezbernal, line 1: I think we can remove  integration/k8s/run_trace_open_test.go in this same commit.

pawarpranav83: We run the gadget using `$KUBECTL_GADGET`, so should we also cover the `gadgets/GADGET_NAME/test` with both `ig` and `kubectl-gadget`?

mauriciovasquezbernal: That'd be great to cover it in both scenarios.

pawarpranav83: Right, but we would need to set up the environment for testing it using kubectl-gadget (using `testMainInspektorGadget`) since we are going to replace the integration tests to use image-based gadgets (next PR), this would be covered there.
And the tests in `gadgets/GADGET_NAME/test` can be unit tests for the gadget.

pawarpranav83: I think tests in gadgets/GADGET_NAME/test would replace: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/trace/open/tracer/tracer_test.go, right?

mauriciovasquezbernal: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/trace/open/tracer/tracer_test.go is a unit test that doesn't use real containers. I think `gadgets/GADGET_NAME/test`  would replace `integration/k8s/run_GADGET_NAME_test.go`. We still need a way to run tests like `pkg/gadgets/trace/open/tracer/tracer_test.go` for image-based gadgets, but that's future work for now.

mauriciovasquezbernal:
Can you please integrate the last commit "removing duplication of containers and match pkg functions in integra…" into the first one?  

Ideally the commit structure could be something like:

- Move containers `integration/containers/` to `pkg/testing/containers/` (also updating the users of it)
- Same for things related to pkg/testing/match/match.go
- Introduce testing framework
- Test for trace open 
- Documentation

gadgets/trace_open/test/trace_open_test.go
mauriciovasquezbernal: Forget about my previous comments. I just realized this PR is running the events on local containers (by using the container factory) instead of using Kubernetes. `integration/k8s/run_trace_open_test.go` is using kubernetes, hence you shouldn't remove that file on this PR. Sorry for the noise and confusion. 

For now let's keep this PR as it's. I'll think about future steps, like extending the framework for running image-based gadgets on kubernetes too.

docs/devel/hello-world-gadget.md
```diff
@@ -589,6 +589,128 @@ RUNTIME.CONTAINERNAME  PID          COMM         FILENAME
...
+
+        // parses the output and matches it to expectedEntry.
+        match.ExpectEntriesToMatch(t, output, normalize, expectedEntry)
+      }
+    ),
+  )
+
+  testSteps := []igtesting.TestStep{
+    mygadgetCmd,
+    containerFactory.NewContainer(cn, "setuidgid 1000:1111 cat /dev/null"),
```
pawarpranav83, line 694: Need to update (if the CI works).

docs/devel/hello-world-gadget.md
```diff
@@ -589,6 +589,127 @@ RUNTIME.CONTAINERNAME  PID          COMM         FILENAME
...
+    // and stop later and verify the output. 
+    containerFactory.NewContainer(cn, "while true; do setuidgid 1000:1111 cat /dev/null; sleep 0.1; done", containers.WithStartAndStop()),
+    mygadgetCmd,
+  }
+
+  igtesting.RunTestSteps(testSteps, t)
+}
+```
+
+Before running the test we need to tag the image (skip if already tagged) and define environment variables for gadget repository and tag.
```
mauriciovasquezbernal, line 700: IMO it's not so friendly, please have a look at my comment above.

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,101 @@
...
+		ig.path = path
+	}
+}
+
+// WithPathFromEnvVar used for reading custom path for ig executable from IG env var.
+func WithPathFromEnvVar() option {
+	return func(ig *runner) {
+		ig.path = os.Getenv("IG")
+	}
+}
```
mauriciovasquezbernal, line 61: IMO we don't need this. We can hardcode it to always use the value on the `IG` if it's defined. By doing so we:
- Remove the needed use `WithPathFromEnvVar` in the test
- Remove the needed to set `IG` when running the test

I'm suggesting something like: (It's possible this diff doesn't apply because I edited it manually, but that's the idea)

```diff
diff --git a/gadgets/trace_open/test/trace_open_test.go b/gadgets/trace_open/test/trace_open_test.go
index 4a2367028..8ae964871 100644
--- a/gadgets/trace_open/test/trace_open_test.go
+++ b/gadgets/trace_open/test/trace_open_test.go
@@ -19,7 +19,6 @@ import (

        "github.com/stretchr/testify/require"

-       gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
        igtesting "github.com/inspektor-gadget/inspektor-gadget/pkg/testing"
        "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/containers"
        igrunner "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/ig"
@@ -44,15 +43,12 @@ type traceOpenEvent struct {
 }

 func TestTraceOpen(t *testing.T) {
-       gadgettesting.RequireEnvironmentVariables(t)
-
        cn := "test-trace-open"
        containerFactory, err := containers.NewContainerFactory("docker")
        require.NoError(t, err, "new container factory")

        traceOpenCmd := igrunner.New(
                "trace_open",
-               igrunner.WithPathFromEnvVar(),
                igrunner.WithFlags("--runtimes=docker", "--timeout=10"),
                igrunner.WithValidateOutput(
                        func(t *testing.T, output string) {
diff --git a/pkg/testing/ig/ig.go b/pkg/testing/ig/ig.go
index 5bf23a987..45aa0399b 100644
--- a/pkg/testing/ig/ig.go
+++ b/pkg/testing/ig/ig.go
@@ -53,13 +53,6 @@ func WithPath(path string) option {
        }
 }

-// WithPathFromEnvVar used for reading custom path for ig executable from IG env var.
-func WithPathFromEnvVar() option {
-       return func(ig *runner) {
-               ig.path = os.Getenv("IG")
-       }
-}
-
 // WithFlags args should be in form: "--flag_name=value" or "-shorthand=value".
 func WithFlags(flags ...string) option {
        return func(ig *runner) {
@@ -83,14 +76,24 @@ func WithValidateOutput(validateOutput func(t *testing.T, output string)) option

 // New creates a new IG configured with the options passed as parameters.
 func New(image string, opts ...option) igtesting.TestStep {
        ig := &runner{
                path:  "ig",
                Command: command.Command{
                        Name: "Run_" + image,
                },
        }

+       if path, ok := os.LookupEnv("IG"); ok {
+               ig.path = path
+       }
+
        for _, opt := range opts {
                opt(ig)
        }
```

pkg/testing/ig/ig.go
```diff
@@ -0,0 +1,101 @@
...
+	return func(ig *runner) {
+		ig.ValidateOutput = validateOutput
+	}
+}
+
+// New creates a new IG configured with the options passed as parameters.
+func New(image string, opts ...option) igtesting.TestStep {
+	ig := &runner{
+		path:  "ig",
+		image: fmt.Sprintf("%s/%s:%s", os.Getenv("GADGET_REPOSITORY"), image, os.Getenv("GADGET_TAG")),
```
mauriciovasquezbernal, line 88: IMO this should also be optional, there is not any reason why I should tag my gadget image if I want to run a test locally:

Something like the following should avoid it:


```diff
 func New(image string, opts ...option) igtesting.TestStep {
+       repository := os.Getenv("GADGET_REPOSITORY")
+       tag := os.Getenv("GADGET_TAG")
+       if repository != "" && tag != "" {
+               image = fmt.Sprintf("%s/%s:%s", repository, image, tag)
+       }
+
        ig := &runner{
                path:  "ig",
-               image: fmt.Sprintf("%s/%s:%s", os.Getenv("GADGET_REPOSITORY"), image, os.Getenv("GADGET_TAG")),
+               image: image,
                Command: command.Command{
                        Name: "Run_" + image,
                },
        }

        for _, opt := range opts {
                opt(ig)
        }
``

**Review by mauriciovasquezbernal — COMMENTED** [review]
I have two small comments about usability. Other than them, LGTM!

File: pkg/testing/ig/ig.go (hunk 48 above)
Comment (pawarpranav83) [reply pkg/testing/ig/ig.go:61]: Right, we can remove the first part, but the second part would still be required since we want to skip this test when we use `make test`, instead of using a separate flag we check whether IG is defined to determine whether we skip the test.

Comment (mauriciovasquezbernal) [reply pkg/testing/ig/ig.go:61]: Makes sense to me. Let's skip my comment for now then.

**Review by mauriciovasquezbernal — APPROVED** [review]
LGTM. It looks really good, thanks for working on it. 

I'd say next steps are to port the existing tests to use this:
```
integration/k8s/run_trace_mount_test.go
integration/k8s/run_trace_oomkill_test.go
integration/k8s/run_trace_signal_test.go
integration/k8s/run_trace_sni_test.go
integration/k8s/run_trace_tcp_test.go
```

Also non-trace tests should be converted to be sure the framework is flexible enough.

Issue #2591, opened 2024-03-07
trace/dns: Improve tests


mauriciovasquezbernal:
Implement tests for:
- ipv6 (se https://github.com/inspektor-gadget/inspektor-gadget/pull/2574#pullrequestreview-1915855066)
- uncompressed messages (only compressed messages are tested [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/284aaba7f80bcb8c4cab5db879babc6cf836c197/tools/dnstester/dnstester.go#L26))

mrhapile:
Hi maintainers 👋  
I’d like to work on this issue and add tests for IPv6 and uncompressed DNS messages, building on the changes from #2574.  
I’m planning to follow the existing test patterns in trace/dns and use a small DNS test server to validate both IPv4/IPv6 paths.  
Please let me know if that sounds good, and feel free to assign me. Thanks!

Issue #1271, opened 2023-01-16
[RFE] Avoid repeated code on inspektor-gadget and local-gadget integration tests


blanquicet:
As mentioned while [reviewing this PR](https://github.com/inspektor-gadget/inspektor-gadget/pull/1265#discussion_r1071008849), the code for some tests are almost identical for local-gadget k8s, non-k8s and inspektor-gadget. 

The idea of this issue is to avoid such redundancy. One possibility is to move the common code to a shared folder, maybe `integration/common`.

pawarpranav83:
Currently working on this. Will make a PR soon!

PR #2609, opened 2024-03-12
integration: Test uncompressed DNS messages 

mqasimsarfraz:
This change adds tests to support testing uncompressed DNS messages.

Related: #2591

### DIFF
--- a/integration/ig/k8s/trace_dns_test.go
+++ b/integration/ig/k8s/trace_dns_test.go
@@ -24,13 +24,10 @@ import (
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
-func TestTraceDns(t *testing.T) {
-	t.Parallel()
-	ns := GenerateTestNamespaceName("test-trace-dns")
-
+func newTraceDnsCmd(t *testing.T, ns string, dnsServerArgs string) *Command {
 	commandsPreTest := []*Command{
 		CreateTestNamespaceCommand(ns),
-		PodCommand("dnstester", *dnsTesterImage, ns, "", ""),
+		PodCommand("dnstester", *dnsTesterImage, ns, `["/dnstester"]`, dnsServerArgs),
 		WaitUntilPodReadyCommand(ns, "dnstester"),
 	}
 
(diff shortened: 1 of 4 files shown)


mqasimsarfraz:
> I am wondering if trace dns can detect if a DNS message was compressed and if it would make sense to add such a flag to the event.

I am not sure how much work is needed to do that, but I prefer to wait till we get a use-case for it. What do you say?

integration/ig/k8s/trace_dns_test.go
```diff
@@ -24,13 +24,15 @@ import (
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 )
 
-func TestTraceDns(t *testing.T) {
-	t.Parallel()
-	ns := GenerateTestNamespaceName("test-trace-dns")
+func newTraceDnsCmd(t *testing.T, ns string, uncompress bool) *Command {
+	cmdArgs := ""
```
mauriciovasquezbernal, line 28: perhaps `newTraceDnsCmd` can accept cmdArgs directly? So we can easily extend this in the future, for instance, to test ipv6.

mauriciovasquezbernal (approved):
LGTM. Just one minor comment.

We definitely need https://github.com/inspektor-gadget/inspektor-gadget/issues/1271, it's a mess to handle all the duplication of these tests.

Thanks for implementing tests!

integration/ig/non-k8s/trace_dns_test.go
```diff
@@ -152,7 +149,7 @@ func TestTraceDns(t *testing.T) {
 	}
 
 	dnsCmds := []string{
-		"/dnstester & sleep 2", // wait to ensure dns server is running
```
mqasimsarfraz, line 155: It seems we were suppose to use `&&` here so I fixed it, sharing to make sure it is caught in the review!

mqasimsarfraz: Ops, yes you are right. Not sure what was I thinking. :laughing: I will revert the change!

mqasimsarfraz:
> We definitely need https://github.com/inspektor-gadget/inspektor-gadget/issues/1271, it's a mess to handle all the duplication of these tests.

:100:

mqasimsarfraz:
Thanks for the reviews!

PR #2611, opened 2024-03-13
image list: add creation date

alban:
# image list: add creation date

The OCI image spec defines the creation date annotation: https://github.com/opencontainers/image-spec/blob/main/annotations.md

> `org.opencontainers.image.created` date and time on which the image was built, conforming to RFC 3339.

This patch ensures that the annotation is set with the same value for both amd64 and arm64.

Then, the date is printed in the image list command:

Note: the creation date will also be useful for the future Artifact Hub patch because the 'createdAt' field is mandatory in artifacthub-pkg.yml:

https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-pkg.yml
```
createdAt: The date this package was created (RFC3339 layout) (required)
```

## How to use

```
$ sudo -E ig image list
REPOSITORY  TAG     DIGEST       CREATED
trace_dns   latest  d98eea3ede9e 2024-03-06T16:50:41+01:00
```

## Testing done

```bash
go run -exec 'sudo -E' ./cmd/ig/... image build -t mytest:mytest2 gadgets/trace_dns/
go run -exec 'sudo -E' ./cmd/ig/... image list|grep mytest
```

### DIFF
--- a/pkg/oci/build.go
+++ b/pkg/oci/build.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"time"
 
 	"github.com/distribution/reference"
 	"github.com/opencontainers/image-spec/specs-go"
@@ -61,6 +60,8 @@ type BuildGadgetImageOpts struct {
 	UpdateMetadata bool
 	// If true, the metadata is validated before creating the image.
 	ValidateMetadata bool
+	// Date and time on which the image is built (date-time string as defined by RFC 3339).
+	CreatedDate string
 }
 
 // BuildGadgetImage creates an OCI image with the objects provided in opts. The image parameter in
@@ -147,7 +148,6 @@ func annotationsFromMetadata(metadataBytes []byte) (map[string]string, error) {
 	annotations := map[string]string{
 		ocispec.AnnotationTitle:         metadata.Name,
 		ocispec.AnnotationDescription:   metadata.Description,
-		ocispec.AnnotationCreated:       time.Now().Format(time.RFC3339),
 		ocispec.AnnotationURL:           metadata.HomepageURL,
 		ocispec.AnnotationDocumentation: metadata.DocumentationURL,
 		ocispec.AnnotationSource:        metadata.SourceURL,
@@ -186,7 +186,7 @@ func createEmptyDesc(ctx context.Context, target oras.Target) (ocispec.Descripto
 	return emptyDesc, nil
 }
 
-func createManifestForTarget(ctx context.Context, target oras.Target, metadataFilePath, progFilePath, wasmFilePath, arch string) (ocispec.Descriptor, error) {
+func createManifestForTarget(ctx context.Context, target oras.Target, metadataFilePath, progFilePath, wasmFilePath, arch, createdDate string) (ocispec.Descriptor, error) {
 	progDesc, err := createLayerDesc(ctx, target, progFilePath, eBPFObjectMediaType)
 	if err != nil {
 		return ocispec.Descriptor{}, fmt.Errorf("creating and pushing eBPF descriptor: %w", err)
@@ -204,6 +204,7 @@ func createManifestForTarget(ctx context.Context, target oras.Target, metadataFi
 		if err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf("creating metadata descriptor: %w", err)
 		}
+		defDesc.Annotations[ocispec.AnnotationCreated] = createdDate
 	} else {
 		// Create an empty descriptor
 		defDesc, err = createEmptyDesc(ctx, target)
@@ -211,6 +212,11 @@ func createManifestForTarget(ctx context.Context, target oras.Target, metadataFi
 			return ocispec.Descriptor{}, fmt.Errorf("creating empty descriptor: %w", err)
 		}
 		artifactType = eBPFObjectMediaType
+
+		// Even without metadata, we can still set some annotations
+		defDesc.Annotations = map[string]string{
+			ocispec.AnnotationCreated: createdDate,
+		}
 	}
 
 	// Create the manifest which combines everything and push it to the memory store
@@ -263,7 +269,7 @@ func createImageIndex(ctx context.Context, target oras.Target, o *BuildGadgetIma
 	layers := []ocispec.Descriptor{}
 
 	for arch, path := range o.EBPFObjectPaths {
-		manifestDesc, err := createManifestForTarget(ctx, target, o.MetadataPath, path, o.WasmObjectPath, arch)
+		manifestDesc, err := createManifestForTarget(ctx, target, o.MetadataPath, path, o.WasmObjectPath, arch, o.CreatedDate)
 		if err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf("creating %s manifest: %w", arch, err)
 		}
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -74,6 +74,7 @@ type GadgetImageDesc struct {
 	Repository string `column:"repository"`
 	Tag        string `column:"tag"`
 	Digest     string `column:"digest,width:12,fixed"`
+	Created    string `column:"created"`
 }
 
 func (d *GadgetImageDesc) String() string {
@@ -88,6 +89,11 @@ func getLocalOciStore() (*oci.Store, error) {
 		return nil, err
 	}
 	return oci.New(defaultOciStore)
+}
+
+func getTimeFromAnnotations(annotations map[string]string) string {
+	created, _ := annotations[ocispec.AnnotationCreated]
+	return created
 }
 
 // GetGadgetImage pulls the gadget image according to the pull policy and returns
@@ -162,7 +168,9 @@ func pullGadgetImageToStore(ctx context.Context, imageStore oras.Target, image s
 	imageDesc := &GadgetImageDesc{
 		Repository: targetImage.Name(),
 		Digest:     desc.Digest.String(),
+		Created:    "", // Unfortunately, oras.Copy does not return annotations
 	}
+
 	if ref, ok := targetImage.(reference.Tagged); ok {
 		imageDesc.Tag = ref.Tag()
 	}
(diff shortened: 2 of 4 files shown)


pkg/oci/build.go
```diff
@@ -263,7 +271,7 @@ func createImageIndex(ctx context.Context, target oras.Target, o *BuildGadgetIma
 	layers := []ocispec.Descriptor{}
 
 	for arch, path := range o.EBPFObjectPaths {
-		manifestDesc, err := createManifestForTarget(ctx, target, o.MetadataPath, path, o.WasmObjectPath, arch)
+		manifestDesc, err := createManifestForTarget(ctx, target, o.MetadataPath, path, o.WasmObjectPath, arch, o.CreatedDate)
```
blanquicet, line 274: > This patch ensures that the annotation is set with the same value for both amd64 and arm64.
>
I tested the `main` branch and the `org.opencontainers.image.created` annotation is already the same:

```bash
$ IG_EXPERIMENTAL=true go run --exec "sudo -E" ./cmd/ig/... image build -t joseregistry.azurecr.io/mytest:latest gadgets/trace
_dns/
Successfully built joseregistry.azurecr.io/mytest:latest@sha256:29500405310af1d442c6c6a282c88d401187d62d9381a168c5e0af1478eafed8

$ IG_EXPERIMENTAL=true go run --exec "sudo -E" ./cmd/ig/... image push joseregistry.azurecr.io/mytest:latest
Pushing joseregistry.azurecr.io/mytest:latest...
Successfully pushed joseregistry.azurecr.io/mytest:latest@sha256:29500405310af1d442c6c6a282c88d401187d62d9381a168c5e0af1478eafed8

# Create annotation for the index
$ crane manifest joseregistry.azurecr.io/mytest:latest | jq '.annotations."org.opencontainers.image.created"'
"2024-03-14T13:42:41-05:00"

# Create annotation for amd and arm manifests
$ crane manifest joseregistry.azurecr.io/mytest:latest | jq '.manifests[].annotations."org.opencontainers.image.created"'
"2024-03-14T13:42:41-05:00"
"2024-03-14T13:42:41-05:00"
```

Do you have an example of when they could be different?

pkg/oci/oci.go
```diff
@@ -90,6 +91,11 @@ func getLocalOciStore() (*oci.Store, error) {
 	return oci.New(defaultOciStore)
 }
 
+func getTimeFromAnnotations(annotations map[string]string) string {
```
mqasimsarfraz, line 94: nit: What do you think about showing the time since the image was created (e.g `14 minutes ago`) similar to `docker` ?

pkg/oci/build.go
```diff
@@ -204,13 +206,19 @@ func createManifestForTarget(ctx context.Context, target oras.Target, metadataFi
 		if err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf("creating metadata descriptor: %w", err)
 		}
+		defDesc.Annotations[ocispec.AnnotationCreated] = createdDate
```
mqasimsarfraz, line 209: Now that we are setting created annotation using `createdDate` do we still need https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/oci/build.go#L150 ?

pkg/oci/build.go
```diff
@@ -204,13 +206,19 @@ func createManifestForTarget(ctx context.Context, target oras.Target, metadataFi
...
 		// Create an empty descriptor
 		defDesc, err = createEmptyDesc(ctx, target)
 		if err != nil {
 			return ocispec.Descriptor{}, fmt.Errorf("creating empty descriptor: %w", err)
 		}
 		artifactType = eBPFObjectMediaType
+
+		// Even without metadata, we can still set some annotations
+		defDesc.Annotations = map[string]string{
+			ocispec.AnnotationCreated: time.Now().Format(time.RFC3339),
```
mqasimsarfraz, line 220: Why can't we use `createdDate` here?

mqasimsarfraz (commented):
Looking good, Just some minor comments!

pkg/oci/build.go
alban: If your build machine is fast enough, you won't see it.

I took the main branch and added this patch:
```patch
--- a/pkg/oci/build.go
+++ b/pkg/oci/build.go
@@ -141,7 +141,7 @@ func annotationsFromMetadata(metadataBytes []byte) (map[string]string, error) {
        if err := yaml.NewDecoder(bytes.NewReader(metadataBytes)).Decode(&metadata); err != nil {
                return nil, fmt.Errorf("decoding metadata file: %w", err)
        }
-
+       time.Sleep(2 * time.Second)
        // Suggested annotations for the OCI image
        // https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys
        annotations := map[string]string{
```

Then, tested with:
```sh
$ IG_EXPERIMENTAL=true go run --exec "sudo -E" ./cmd/ig/... image build -t albantest.azurecr.io/mytest:latest gadgets/trace_dns/
Successfully built albantest.azurecr.io/mytest:latest@sha256:ef0fad91c7ff0fb690667f46b0bd3c29a7bb1b6aad12fe81d3d0f3f15269b9b8
$ IG_EXPERIMENTAL=true go run --exec "sudo -E" ./cmd/ig/... image push albantest.azurecr.io/mytest:latest
Pushing albantest.azurecr.io/mytest:latest...
Successfully pushed albantest.azurecr.io/mytest:latest@sha256:ef0fad91c7ff0fb690667f46b0bd3c29a7bb1b6aad12fe81d3d0f3f15269b9b8
$ crane manifest albantest.azurecr.io/mytest:latest | jq '.annotations."org.opencontainers.image.created"'
"2024-03-19T13:11:04+01:00"
$ crane manifest albantest.azurecr.io/mytest:latest | jq '.manifests[].annotations."org.opencontainers.image.created"'
"2024-03-19T13:11:04+01:00"
"2024-03-19T13:11:06+01:00"
$
```

Here the two timestamps are different.

blanquicet: Got it!

pkg/oci/oci.go
alban: Done :)
Now it looks like this:
```
$ sudo -E ./ig image list
REPOSITORY                           TAG                 DIGEST       CREATED
albantest.azurecr.io/mytest          latest              30e8fd4c86ed 10 minutes ago
```
Or:
```
$ sudo -E ./ig image list | cat # or using --no-trunc
REPOSITORY TAG DIGEST CREATED
albantest.azurecr.io/mytest latest sha256:30e8fd4c86edb742969125f6ae71da065d3207498f5ce528d922c683904f8770 2024-03-19T13:59:31+01:00
```

pkg/oci/build.go
alban: Fixed :)

mqasimsarfraz (approved):
Looks good, I tested it and it seems to be working fine for locally built images.

I noticed if I pull the image and print it, I see an empty value:

```
REPOSITORY                                                              TAG                                                                     DIGEST       CREATED                                                                
test                                                                    latest                                                                  2772850011db 4 days ago                                                             
trace_open                                                              latest                                                                  42832ffe4547 
```

Is that expected?

alban:
> I noticed if I pull the image and print it, I see an empty value:
> Is that expected?

It depends if the image you pulled has the annotation with the creation date.

alban:
> > I noticed if I pull the image and print it, I see an empty value:
> > Is that expected?
> 
> It depends if the image you pulled has the annotation with the creation date.

Actually, there seems to be a bug there...

alban:
When pulling the image from the remote repository with `pullGadgetImageToStore()`, `oras.Copy()` does not return the annotations from the index but generate a new descriptor:
https://github.com/oras-project/oras-go/blob/main/registry/remote/repository.go#L1039-L1042

However, the annotations from the manifests are preserved. Two possible solutions:
1. When pulling, copy the annotations from the manifest to the index (`/var/lib/ig/oci-store/index.json`)
2. When listing the images (`ig image list`), read the annotations from the manifests (`/var/lib/ig/oci-store/blobs/sha256/`)

I tried to implement the first solution, but I didn't get it to work:

<details>

```patch
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -165,11 +165,31 @@ func pullGadgetImageToStore(ctx context.Context, imageStore oras.Target, image s
                return nil, fmt.Errorf("copying to remote repository: %w", err)
        }
 
+       // oras.Copy() ignores annotations from the index. Retrieve annotations from manifest.
+       manifest, err := getManifestForHost(ctx, imageStore, targetImage.String())
+       if err != nil {
+               return nil, fmt.Errorf("reading manifest: %w", err)
+       }
+
+       desc.Annotations = manifest.Annotations
+
+       var manifestDesc ocispec.Descriptor
+       manifestBytes, err := getContentBytesFromDescriptor(ctx, repo, manifestDesc)
+       if err != nil {
+               return nil, fmt.Errorf("getting manifest from descriptor: %w", err)
+       }
+       repo.Push(ctx, desc, bytes.NewReader(manifestBytes))
+       err = repo.Tag(ctx, desc, targetImage.String())
+       if err != nil {
+               return nil, fmt.Errorf("tagging image with annotations: %w", err)
+       }
+
        imageDesc := &GadgetImageDesc{
                Repository: targetImage.Name(),
                Digest:     desc.Digest.String(),
-               Created:    getTimeFromAnnotations(desc.Annotations),
+               Created:    getTimeFromAnnotations(manifest.Annotations),
        }
+       fmt.Printf("Image %s:%s@%s created at %+v\n", imageDesc.Repository, imageDesc.Tag, imageDesc.Digest, imageDesc.Created)
        if ref, ok := targetImage.(reference.Tagged); ok {
                imageDesc.Tag = ref.Tag()
        }
```

The Printf prints the correct date, but it is not saved in `/var/lib/ig/oci-store/index.json`... (to be debugged?)

</details>

See also this long discussion https://github.com/oras-project/oras-go/issues/22#issuecomment-1223664264 about annotations in descriptor versus annotations in the manifest itself. IIUC, the discussion suggests we should not rely on the remote registry to keep the annotations in the index, but instead we should use the annotations in the manifest.

I suggest to merge this PR, and find a way to fix it in a follow-up PR. Wdyt?

alban:
I fixed it with solution 2.

It should work with pulled images too now.

mqasimsarfraz (approved):
LGTM, works fine for pulled images as well!

PR #2612, opened 2024-03-14
trace_mount: add missing fields in gadget metadata

alban:
When rebuilding the trace_mount gadget with --update-metadata, it added missing fields. The integration tests are fixed accordingly.


mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #2620, opened 2024-03-18
docs/run: Use shorter form in the guide

mqasimsarfraz:


docs/guides/run.md
```diff
@@ -159,3 +165,9 @@ mycontainer3                                        122110  cat              0
 mycontainer3                                        122110  cat              0        0        3         /lib/libc.so.6
 mycontainer3                                        122110  cat              0        0        3         /dev/null
 ```
+
+By default, it will use the `ghcr.io` as the registry, `inspektor-gadget/gadget` as the repository prefix and `latest` as the tag. You can also run them in fully qualified form as:
+
```
mauriciovasquezbernal, line 170: Do we really need to duplicate it? Can't we have a common section with this information for both ig and kubectl-gadget?

mqasimsarfraz: Good point. I changed it now to only mention the default values.

docs/guides/run.md
```diff
@@ -57,6 +62,12 @@ ubuntu-hirsute         default                mypod2                 mypod2
 ubuntu-hirsute         default                mypod2                 mypod2                 242164  cat         0        0        3   /dev/null
 ```
 
+By default, it will use the `ghcr.io` as the registry, `inspektor-gadget/gadget` as the repository prefix and `latest` as the tag. You can also run them in fully qualified form as:
+
+```bash
+$ kubectl gadget run ghcr.io/inspektor-gadget/gadget/trace_tcpconnect:latest
+```
+
```
mauriciovasquezbernal, line 70: I suppose this part is not needed anymore?

mauriciovasquezbernal (approved):
LGTM once my last comment in handled. Thanks!

docs/guides/run.md
mqasimsarfraz: Oops, thanks for catching it. Removed

Issue #2316, opened 2023-12-20
[RFE] Add missing features to image based gadgets


eiffel-fl:
- [ ] `trace_mount`: The function call should be printed (`mount("/foo/, "/bar", ...).
- [x] `trace_signal`: Signal number should be translated to signal name (9 -> `SIGKILL`) (#2630)
- [ ] `trace_exec`: Arguments column is missing (https://github.com/inspektor-gadget/inspektor-gadget/pull/2304)
- [ ] `trace_tcpdrop`: once we support bitfield to string conversion (https://github.com/inspektor-gadget/inspektor-gadget/issues/2482), update `event->tcpflags` type (https://github.com/inspektor-gadget/inspektor-gadget/pull/2477#discussion_r1483555007)
- [ ] `trace_tcpdrop`: once we support sub-structures in `struct event`, convert fields of original `proc_ctx` to struct (https://github.com/inspektor-gadget/inspektor-gadget/pull/2477#discussion_r1484672309)
- [ ] `trace_open`: provide a human readable version of [flags](https://github.com/inspektor-gadget/inspektor-gadget/blob/5fec457b500ef91545b86f1f3a0a6038d46daf97/gadgets/trace_open/program.bpf.c#L32) once we support bitfield to string conversion 
- [x] `trace_dns`: correctly parse dns name field. (#2845)
- [ ] `trace_capabilities`: https://github.com/inspektor-gadget/inspektor-gadget/issues/2908#issue-2322586619

blanquicet:
Also, error numbers could be translated to error name, see https://github.com/inspektor-gadget/inspektor-gadget/pull/2687#pullrequestreview-1985985700.

mauriciovasquezbernal:
I opened different issues tracking all of this:
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2995
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2993
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2996
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2992

PR #2630, opened 2024-03-20
Add signal formatter

mauriciovasquezbernal:
Add logic to convert a signal number to its name.

Part of #2316
Supersedes #2325

### Testing

Build the trace_signal gadget. 

```bash 
# build a builder image that includes the changes on the headers
$ docker build -f Dockerfiles/ebpf-builder.Dockerfile . -t ebpf-builder

$ cd gadets/trace_signal
$ sudo -E ig image build  . -t trace_signal --builder-image ebpf-builder:latest
...
```

Run the gadget

```bash
$ sudo -E ig run trace_signal
RUNTIME.CONTAINERNAME     PID            TPID           MNTNS_ID  UID           GID           RET  COMM          SIG           TIMESTAMP
```

Run a container that generates a signal

```bash 
$ docker run --name c2 --rm -it busybox sh  -c 'sleep 3 & kill -kill $!'
```

The gadget shows the signal name:

```bash 
$ sudo -E ig run trace_signal
RUNTIME.CONTAINERNAME     PID            TPID           MNTNS_ID  UID           GID           RET  COMM          SIG           TIMESTAMP
c2                        54450          69685          402653419 0             0             0    sh            SIGKILL       2024-04-15T1
```

### TODO
- [ ] The new added xx_raw field is a string, but it should be a number. Extend the datasource to allow adding single static fields coming from enrichers?


include/gadget/types.h
```diff
@@ -34,4 +34,7 @@ typedef __u64 gadget_mntns_id;
 // time.
 typedef __u64 gadget_timestamp;
 
+// gadget_signal is used to represent a unix signal. It's automatically converted to the signal name.
+typedef __u32 gadget_signal;
```
flyth, line 38: nit: Maybe "a field is automatically added that contains the name as string" or something like it.

pkg/operators/formatters/formatters.go
```diff
@@ -123,6 +129,30 @@ type replacer struct {
 
 // careful: order and priority matter both!
 var replacers = []replacer{
+	{
+		name:      "signal",
+		selectors: []string{"type:" + SignalTypeName},
+		replace: func(ds datasource.DataSource, in datasource.FieldAccessor) (func(data datasource.Data) error, error) {
+			out, err := ds.AddField(in.Name() + "_str")
```
flyth, line 136: Are we sure about "_str" now? The other replacers don't use it anymore (they had easy exits, though, hehe.)

Also, maybe add the possibility to change the target name using an annotation like done with "formatters.timestamp.target". That served more as an example, but I think it adds value.

pkg/operators/formatters/formatters.go
```diff
@@ -18,8 +18,11 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"syscall"
 	"time"
 
+	"golang.org/x/sys/unix"
```
flyth, line 24: Just a note: we should later on think about splitting formatters into ones that could run on the client and ones that should run on the server; this dependency makes it OS dependent (I think), so would be suited for the server running operator.

flyth (approved):
Looks good, happy to see this being used 😄

pkg/operators/formatters/formatters.go
mauriciovasquezbernal: I changed the approach a bit. Now the original column is rewritten to contain the signal name, after all the string representation is the most interesting one. The raw value is still kept on a `foo_raw` column. 

Later on we can optimize it to avoid the value to string conversion in case the consumer is not interested in it.

pkg/operators/formatters/formatters.go
```diff
@@ -123,6 +129,34 @@ type replacer struct {
 
 // careful: order and priority matter both!
 var replacers = []replacer{
+	{
+		name:      "signal",
+		selectors: []string{"type:" + SignalTypeName},
+		replace: func(ds datasource.DataSource, in datasource.FieldAccessor) (func(data datasource.Data) error, error) {
+			oldName := in.Name()
+
+			in.Rename(oldName + "_raw")
```
flyth, line 138: This is missing an error check.

flyth (commented):
Still looks good from code inspection, there's just an error check missing that is important.

PR #2646, opened 2024-03-26
integration: Use interface type instead of type parameter

blanquicet:
# Use interface type instead of type parameter in integration tests

The `RunTestSteps()` can perfectly work using interfaces instead of type parameters. This change makes the code more readable and easier to understand. Additionally, as suggested by the official Go blog post [1]: "_Don’t use type parameters if method implementations differ_", and in this case, the implementation of `TestStep`'s methods is completely different between `integration.Command` and `testutils.ContainerdContainer` or `testutils.DockerContainer`. The blog post also states "_using a type parameter will generally not be faster than using an interface type_". So, another reason to not use type parameters in this case.

[1] https://blog.golang.org/generics-next-step

## Testing done

Integration tests continue working 🙂

### DIFF
--- a/integration/common/profile_tcprtt.go
+++ b/integration/common/profile_tcprtt.go
@@ -75,7 +75,7 @@ func RunTestProfileTCPRTT(t *testing.T) {
 	clientPodName := "test-pod"
 	ns := integration.GenerateTestNamespaceName("test-profile-tcprtt")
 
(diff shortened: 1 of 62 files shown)


burak-ok (approved):
Thank you. Finally `*Command{` and `&Command` are gone :D

PR #2647, opened 2024-03-26
integration: IntegrationTestsContainer interface is not necessary

blanquicet:
# IntegrationTestsContainer interface is not necessary

`IntegrationTestsContainer` defines what a container struct must implement so that it can be run by `RunTestSteps`. However, it's exactly what `TestStep` also does.


burak-ok (approved):
LGTM, thanks

PR #2649, opened 2024-03-27
integration: Wait until test-pod was OOMKilled for oomkill related tests

eiffel-fl:


alban (approved):
Not tested but the code looks good to me.
Feel free to merge if all the tests are green.

integration/inspektor-gadget/run_trace_oomkill_test.go
```diff
@@ -92,14 +92,21 @@ spec:
     - while true; do tail /dev/zero; done
 `, ns)
 
+	var waitCommand *Command
+	if *k8sDistro == K8sDistroAKSUbuntu {
+		waitCommand = WaitUntilTestPodOOMKilledCommand(ns)
+	} else {
+		waitCommand = WaitUntilTestPodReadyCommand(ns)
```
alban, line 99: Is the difference of behaviour really about the distribution?

If the container is slow to get oomkilled, we should be able to detect the pod be `condition=ready` and then oomkilled. If the container gets oomkilled fast, the pod status might go to oomkilled fast enough so we don't see the `condition=ready`.

It seems to me we should wait for the condition "ready or oomkilled" to avoid the race condition, if possible.

PR #2650, opened 2024-03-27
README.md: update artifacthub badges

alban:
Two changes:
1. Update the label to: "Helm charts"
2. Update the link to the URL of the Helm chart instead of the search webpage.

Note that the URL still contains "gadget". We are looking to change it to "inspektor-gadget" but at the moment it is linked to the chart name and renaming it would break the updates for users.

See related discussion in:
* https://github.com/inspektor-gadget/inspektor-gadget/pull/2586#discussion_r1526560343

Before this patch:

image

After this patch:

image

cc @mqasimsarfraz @eiffel-fl

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 [![Inspektor Gadget Test Reports](https://img.shields.io/badge/Link-Test%20Reports-blue)](https://inspektor-gadget.github.io/ig-test-reports)
 [![Inspektor Gadget Benchmarks](https://img.shields.io/badge/Link-Benchmarks-blue)](https://inspektor-gadget.github.io/ig-benchmarks/dev/bench)
 [![Release](https://img.shields.io/github/v/release/inspektor-gadget/inspektor-gadget)](https://github.com/inspektor-gadget/inspektor-gadget/releases)
-[![Artifact Hub: Gadgets](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/gadgets)](https://artifacthub.io/packages/search?repo=inspektor-gadgets)
+[![Artifact Hub: Gadgets](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/gadgets)](https://artifacthub.io/packages/search?repo=gadgets)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,7 +13,7 @@
 [![Inspektor Gadget Benchmarks](https://img.shields.io/badge/Link-Benchmarks-blue)](https://inspektor-gadget.github.io/ig-benchmarks/dev/bench)
 [![Release](https://img.shields.io/github/v/release/inspektor-gadget/inspektor-gadget)](https://github.com/inspektor-gadget/inspektor-gadget/releases)
 [![Artifact Hub: Gadgets](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/gadgets)](https://artifacthub.io/packages/search?repo=inspektor-gadgets)
-[![Artifact Hub](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/gadget)](https://artifacthub.io/packages/search?repo=gadget)
```
mqasimsarfraz, line 16: I am not able to open https://artifacthub.io/packages/search?repo=inspektor-gadgets. Does that work for you? 

I am thinking we should use the kind in the url something like: https://artifacthub.io/packages/search?kind=22 and in future maybe filter only the official/verified with: https://artifacthub.io/packages/search?kind=22&verified_publisher=true&official=true What do you say?

alban: > I am not able to open https://artifacthub.io/packages/search?repo=inspektor-gadgets. Does that work for you?

Oops, this comes from #2644. I added a patch to address this. Thanks for noticing.

> I am thinking we should use the kind in the url something like: https://artifacthub.io/packages/search?kind=22 and in future maybe filter only the official/verified with: https://artifacthub.io/packages/search?kind=22&verified_publisher=true&official=true What do you say?

I am not sure it is better to be honest. The repo URL is shorter. If we ever have several artifacthub repos for gadgets, we can update the badge at that time.

mqasimsarfraz: You are right. I thought `repo=gadgets` is searching for repo with gadgets in their name so I thought it is too generic and given we have specific kind (22), we can combine it with `verified_publisher=true&official=true` to ensure only gadgets published by us show up but actually `repo=gadgets` points to [our repository](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/artifacthub-repo.yml) and is indeed a shorter URL for what we want to do!

PR #2651, opened 2024-03-27
[CI] integration: Increase pod and job wait timeout to 45s

burak-ok:
# integration: Increase pod and job ready timeout to 45s

We get CI test failures because of timeouts and therefore we could maybe wait an extra 15s ([The default timeout is 30s](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_wait/))

- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8448407680/job/23142053593#step:11:2168
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8448407680/job/23142052066 (Log too long for direct link to line
   ```
   2024-03-27T08:52:39.6903877Z === NAME  TestTraceOOMKill
   2024-03-27T08:52:39.6904729Z     command.go:483: Command returned(WaitForTestPod):
   2024-03-27T08:52:39.6905683Z         error: timed out waiting for the condition on pods/test-pod
   ```
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8359069409/job/22881852532#step:7:2759

### DIFF
--- a/integration/command.go
+++ b/integration/command.go
@@ -665,7 +665,7 @@ func DeleteRemainingNamespacesCommand() *Command {
 func WaitUntilJobCompleteCommand(namespace string, jobname string) *Command {
 	return &Command{
 		Name:           fmt.Sprintf("WaitForJob: %s", jobname),
-		Cmd:            fmt.Sprintf("kubectl wait job --for condition=complete -n %s %s", namespace, jobname),
+		Cmd:            fmt.Sprintf("kubectl wait job --for condition=complete --timeout 45s -n %s %s", namespace, jobname),
 		ExpectedString: fmt.Sprintf("job.batch/%s condition met\n", jobname),
 	}
 }
@@ -675,7 +675,7 @@ func WaitUntilJobCompleteCommand(namespace string, jobname string) *Command {
 func WaitUntilPodReadyCommand(namespace string, podname string) *Command {
 	return &Command{
 		Name:           "WaitForTestPod",
-		Cmd:            fmt.Sprintf("kubectl wait pod --for condition=ready -n %s %s", namespace, podname),
+		Cmd:            fmt.Sprintf("kubectl wait pod --for condition=ready --timeout 45s -n %s %s", namespace, podname),
 		ExpectedString: fmt.Sprintf("pod/%s condition met\n", podname),
 	}
 }


burak-ok:
> Please, see: #2649

This is not only about OOMKill. Please see the tests also from the first and last link. The issues we have with the CI will not be solved by a single PR. Its not a single issue

> ```
> === NAME  TestRunInsecure
>   command.go:483: Command returned(WaitForJob: copier):
>        error: timed out waiting for the condition on jobs/copier
> ```

and 

> ```
> === NAME  TestProfileCpu
>     command.go:483: Command returned(WaitForTestPod):
>         error: timed out waiting for the condition on pods/test-pod
> ```

burak-ok:
> I stick to my guns, you can indeed increase the waiting time, but the best solution would be to find the why of this time out. For `oomkill`, we now know it, for the others, so far, I do not have any clue.

For the examples (besides the OOMKill) I have no other idea what it might be besides the network may be too slow and overloaded, so the images doesn't get pulled "fast" enough. And then of course the CPU constraints.

blanquicet:
> > I stick to my guns, you can indeed increase the waiting time, but the best solution would be to find the why of this time out. For `oomkill`, we now know it, for the others, so far, I do not have any clue.
> 
> For the examples (besides the OOMKill) I have no other idea what it might be besides the network may be too slow and overloaded, so the images doesn't get pulled "fast" enough. And then of course the CPU constraints.

We are running several tests in parallel so we might be overloading the cluster. I remember @mauriciovasquezbernal had similar issues when working on EKS. So, we should consider limiting the number of parallel tests.

blanquicet:
> > > > I stick to my guns, you can indeed increase the waiting time, but the best solution would be to find the why of this time out. For `oomkill`, we now know it, for the others, so far, I do not have any clue.
> > > 
> > > 
> > > For the examples (besides the OOMKill) I have no other idea what it might be besides the network may be too slow and overloaded, so the images doesn't get pulled "fast" enough. And then of course the CPU constraints.
> > 
> > 
> > We are running several tests in parallel so we might be overloading the cluster. I remember @mauriciovasquezbernal had similar issues when working on EKS. So, we should consider limiting the number of parallel tests.
> 
> 🤣 🤣 🤣 No, we just need to get bigger clusters.

Yep, that's the other option 😝

mauriciovasquezbernal:
I think we need to limit the number of parallel test. I remember I tried it out before when playing with EKS and it didn't work fine, but I just realized I could have confused the `-p` and `--parallel` flags (https://stackoverflow.com/a/72354980), so I wasn't limiting the number of parallel tests that can be run at all.

burak-ok:
> I think we need to limit the number of parallel test.

In the fail log https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8359069409/job/22881852532#step:7:2759 the failure happened on GKE where we use `t2a-standard-2`, which has 2 `vCPUs`.

`-parallel` defaults to `GOMAXPROCS` which means we ran with `-parallel 2` for GKE.

So the only thing we could do for GKE is disable parallel test completely and potentially double the CI time

PR #2669, opened 2024-03-31
pkg/uprobetracer: use `ReadFInodeFromFd` from kfilefields to identify inodes

i-Pear:
# pkg/uprobetracer: use `ReadFInodeFromFd` from kfilefields to identify inodes

This patch introduced `f_inode` field to pkg `kfilefields`,
and it's then used in `uprobetracer` to identify if two FDs share the same inode.

## Testing

Run two containers sharing the same image, and see if records are duplicate.


pkg/kfilefields/bpf/filefields.bpf.c
```diff
@@ -64,6 +65,7 @@ int BPF_KRETPROBE(ig_fget_x, struct file *ret)
 	if (!ff)
 		return 0;
 
+	ff->f_inode = (u64)BPF_CORE_READ(ret, f_inode);
```
alban, line 68: Does it work, do you get the `struct inode*` from the underlying filesystem? Is it the same inode pointer when checking two different containers using the same image?

Or do you have to access with this?
```
ff->private_data = (u64)BPF_CORE_READ(ret, private_data);
struct ovl_inode *inode = (struct ovl_inode *) ff->private_data;
ff->f_inode = (u64)BPF_CORE_READ(inode, vfs_inode);
```

Note that overlay is a kernel module, so it depends on https://github.com/cilium/ebpf/pull/1300 but we recently updated to cilium/ebpf v0.14.0 with it. We'd still need to use `ProgramOptions` with `KernelModuleTypes` I think.

pkg/uprobetracer/tracer.go
```diff
@@ -57,14 +58,14 @@ const (
...
-	token uint64
-}
-
-func getInodeUUID(file *os.File) (inodeUUID, error) {
-	// TODO: use `kfilefields` to get a unique token for each inode,
-	// here we just use `fd` as token, making each file has a distinct token.
-	return inodeUUID{uint64(file.Fd())}, nil
+func getInodeUUID(file *os.File) (uint64, error) {
+	token, err := kfilefields.ReadFInodeFromFd(int(file.Fd()))
+	priv, err := kfilefields.ReadPrivateDataFromFd(int(file.Fd()))
```
alban, line 63: Since `getInodeUUID()` can be called several times for a single gadget, it would make sense to optimize `kfilefields.ReadFInodeFromFd` to create the kfilefield ebpf program only once. Maybe with:
```
func ReadFInodesFromsFd(fd []int) ([]uint64, error)
```

(this could be done in a future PR)

pkg/kfilefields/bpf/filefields.bpf.c
i-Pear: With this method, we may need to provide the BTF for all potentially used file systems, or at least for the necessary structures.

i-Pear: For `--host`, there may have many diffrent FS, other than btrfs/overlayfs.

alban:
I use the following to test it:
```golang
package main

import (
	"fmt"
	"os"

	"github.com/inspektor-gadget/inspektor-gadget/pkg/kfilefields"
)

func main() {
	for _, filename := range os.Args[1:] {
		file, err := os.Open(filename)
		if err != nil {
			panic(err)
		}
		pd, err := kfilefields.ReadPrivateDataFromFd(int(file.Fd()))
		if err != nil {
			panic(err)
		}
		pi, err := kfilefields.ReadPrivateInodeFromFd(int(file.Fd()))
		if err != nil {
			panic(err)
		}
		fmt.Printf("%v\n", filename)
		fmt.Printf("PrivateData: %v\nPrivateInode: %v\n", pd, pi)
	}
}
```
Build it:
```
go build -o kfilefield ./main.go
```
Prepare 2 busybox containers:
```
docker run -ti --rm -v /:/host --name busybox1 busybox
docker run -ti --rm -v /:/host --name busybox2 busybox
```
Get the pids of the 2 containers:
```
export BUSYBOX1=$(docker inspect busybox1|jq '.[0].State.Pid')
export BUSYBOX2=$(docker inspect busybox2|jq '.[0].State.Pid')
```
A couple of tests on `/bin/sh`:
```
$ sudo ./kfilefield /proc/$BUSYBOX1/root/bin/sh /proc/$BUSYBOX2/root/bin/sh
/proc/2517303/root/bin/sh
PrivateData: 18446612684266798848
PrivateInode: 18446612699345928680
/proc/2517475/root/bin/sh
PrivateData: 18446612684261101312
PrivateInode: 18446612699345928680
$ sudo ./kfilefield /proc/$BUSYBOX1/root/bin/sh /proc/$BUSYBOX2/root/bin/sh
/proc/2517303/root/bin/sh
PrivateData: 18446612682768175872
PrivateInode: 12884901889
/proc/2517475/root/bin/sh
PrivateData: 18446612683637488896
PrivateInode: 18446612699345928680
```

The results are not consistent. So I don't think it works fine.

pkg/kfilefields/bpf/filefields.bpf.c
```diff
@@ -66,6 +68,8 @@ int BPF_KRETPROBE(ig_fget_x, struct file *ret)
 
 	ff->private_data = (u64)BPF_CORE_READ(ret, private_data);
 	ff->f_op = (u64)BPF_CORE_READ(ret, f_op);
+	private_file = (struct file *)ff->private_data;
+	ff->private_inode = (u64)BPF_CORE_READ(private_file, f_inode);
```
alban, line 72: `(struct file *)->private_data` is casted to a `struct file *` but this is only valid if the filesystem is overlay.

We can check if it is the case with `(struct file *)->i_sb->s_type->name == "overlay"`.

alban:
> > The results are not consistent. So I don't think it works fine.
> 
> Yes, I can reproduce. At most of the time it works, but sometimes the `PrivateInode` is zero, or to be a short value :(

The randomness was due to a bug in my test program main.go. I added `runtime.KeepAlive(file)` (I updated my previous comment) to ensure the fd is not closed and recycled to something else. That fixed the randomness.

i-Pear:
OK, I've learned once again that if a Golang program encounters sporadic randomness issues, the first thing to check is the garbage collector.

alban:
It needs `make clang-format` and `make ebpf-objects`.

i-Pear:
> It needs make clang-format and make ebpf-objects.

I did both of them

pkg/uprobetracer/tracer.go
```diff
@@ -57,14 +57,12 @@ const (
...
-func getInodeUUID(file *os.File) (inodeUUID, error) {
-	// TODO: use `kfilefields` to get a unique token for each inode,
-	// here we just use `fd` as token, making each file has a distinct token.
-	return inodeUUID{uint64(file.Fd())}, nil
+func getInodeUUID(file *os.File) (uint64, error) {
+	token, err := kfilefields.ReadRealInodeFromFd(int(file.Fd()))
+	if err != nil {
+		return 0, err
+	}
+	return token, nil
```
alban, line 65: Since this function is just a wrapper around `ReadRealInodeFromFd`, it could be removed and the caller can call `ReadRealInodeFromFd` directly.

pkg/uprobetracer/tracer.go
```diff
@@ -93,9 +91,9 @@ type Tracer[Event any] struct {
 	// when users write library names in ebpf section names, it's possible to
 	// find multiple libraries of different archs within the same container,
 	// making this a one-to-many mapping
-	containerPid2Inodes map[uint32][]inodeUUID
+	containerPid2Inodes map[uint32][]uint64
 	// keeps the fd and refCount for each inodeUUID
-	inodeRefCount map[inodeUUID]*inodeKeeper
+	inodeRefCount map[uint64]*inodeKeeper
```
alban, line 96: I suggest to rename `inodeUUID` to `realInodePtr` everywhere. What do you think?

And the comment before `inodeRefCount` should briefly explain why the real inode ptr is used instead of just the inode number, and why we need to care about getting the underlying file under overlayfs.

pkg/uprobetracer/tracer.go
```diff
@@ -218,7 +216,7 @@ func (t *Tracer[Event]) attach(containerPid uint32) {
 	}
 
 	for _, filePath := range attachFilePaths {
-		file, err := os.OpenFile(filePath, unix.O_PATH, 0)
+		file, err := os.Open(filePath)
```
alban, line 219: I remember you removed the `O_PATH` because `FOp` was `empty_fops` in that case:
https://github.com/torvalds/linux/blob/v6.2/fs/open.c#L839-L841

But `pkg/uprobetracer` does not use `FOp` but only `RealInode`. Does it work if you keep `O_PATH`?

pkg/kfilefields/kfilefields.go
```diff
@@ -58,3 +58,19 @@ func ReadFOpForFdType(ft FdType) (uint64, error) {
 	}
 	return ff.FOp, nil
 }
+
+// ReadRealInodeFromFd uses ebpf to read the f_inode pointer from the
+// kernel "struct file" associated with the given fd.
+// Specifically, if fd belongs to overlayFS, it will return the underlying, real inode.
+func ReadRealInodeFromFd(fd int) (uint64, error) {
```
alban, line 65: Could `ReadRealInodeFromFd` be tested in an unit test? We would not test whether the pointer has the correct value, but rather that the comparison between two realInodePtr is correct. The test would need to setup two temporary overlay filesystems with the same lower layer.

This could be in a separate PR, so this PR is not delayed too much.

pkg/uprobetracer/tracer.go
i-Pear: Do you mean using `realInodePtr` as a new type, and replace all `uint64`?

pkg/uprobetracer/tracer.go
i-Pear: If I use `O_PATH`, the `private_data` will also be `0`, then I'm not able to access the real file.

pkg/kfilefields/kfilefields.go
i-Pear: This should be a unittest for `kfilefields`, if the testing framework could support this, I'd like to do so.

pkg/uprobetracer/tracer.go
alban: No, uint64 is fine for me. I meant in the comment above and elsewhere in the file.

pkg/uprobetracer/tracer.go
alban: Ah, right. Could you add a comment above explaining why O_PATH cannot be used?

pkg/uprobetracer/tracer.go
```diff
@@ -94,9 +84,22 @@ type Tracer[Event any] struct {
...
-	inodeRefCount map[inodeUUID]*inodeKeeper
+	containerPid2Inodes map[uint32][]uint64
+	// keeps the fd and refCount for each realInodePtr
+	//
+	// we are using `realInodePtr` (the address of real inode in kernel) to identify a file
+	// instead of just the inode number.
+	// Since overlayFS overwrites the FsID of files and provides its own inode implementation,
+	// we cannot uniquely identify a file on disk using `<FsID, inode>` pairs.
+	//
+	// At the meanwhile, uprobe is using kernel function `d_real_inode` to get the underlying inode,
```
alban, line 95:
```suggestion
	// Meanwhile, uprobe is using kernel function `d_real_inode` to get the underlying inode,
```

alban (approved):
Thanks! LGTM from code review. I haven't tested it yet. Let's see the results of the CI first.

alban:
I prepared the following workload:
```
$ cat main.c 
#include <stdlib.h>
#include <unistd.h>

int main() {
  for (;;) {
    sleep(2);
    void *ptr = malloc(1);
    sleep(1);
    free(ptr);
  }
  return 0;
}
$ cat Dockerfile 
FROM ubuntu
COPY main /bin/main
$ cat Makefile 
all:
	gcc -Wall -o main ./main.c

docker:
	docker build -t malloc .
```

I started the workload on the host filesystem:
```
$ docker run -ti --rm -v /:/host --name c1 ubuntu
# /host/home/alban/test/malloc/main
$ docker run -ti --rm -v /:/host --name c2 ubuntu
# /host/home/alban/test/malloc/main
```
And traced with and without filtering options:
```
sudo -E ./ig run trace_malloc
sudo -E ./ig run trace_malloc -c c1
sudo -E ./ig run trace_malloc -c c2
```

The container filtering and enrichment worked fine and there are no duplicate events.

Then, I tried the workload again but on the overlay filesystem:
```
$ docker run -ti --rm -v /:/host --name c1 malloc /bin/main
$ docker run -ti --rm -v /:/host --name c2 malloc /bin/main
$ sudo -E ./ig run trace_malloc
$ sudo -E ./ig run trace_malloc -c c1
$ sudo -E ./ig run trace_malloc -c c2
```
It worked fined as well.

Issue #2553, opened 2024-02-28
[RFE] image-based gadgets: support stack map


alban:
## Current situation

While eBPF programs in image-based gadgets can use BPF_MAP_TYPE_STACK_TRACE maps, the content of those stack maps cannot be exposed to userspace in an event.

In image-based gadgets, ring buffer maps are typed: the BTF and the gadget metadata specify the struct and its list of fields. But it is self-contained and cannot currently reference an entry in a stack map.

## Impact

* The built-in gadget profile-cpu cannot be implemented as an image-based gadget.
* Image-based gadgets cannot meaningfully use BPF_MAP_TYPE_STACK_TRACE maps.

## Ideal future situation

References to entries of stack maps can be done from:
* The type of events send in a ring buffer maps
* The type a field in a hash map

## Implementation options

include/gadget/types.h has a new type:
```
struct gadget_stack_ref {
    int stack_map; // fd of the stack map
    int stack_id; // value returned by bpf_get_stackid()
}
```

So userspace can interpret a field of type `gadget_stack_id` as an index in a stack map.
The stack_map fd can be initialized with relocations in the same way the argument to bpf_map_lookup_elem is relocated to a map fd.

## Additional information

* An image-based version of the profile-cpu gadget would need to support stack map: #1925
* This is part of the epic #1929

This issue is about using stack maps from image-based gadgets.
Unwinding user stack will be for a future issue.

i-Pear:
@alban You could assign this to me. I think it would be better for me to design the framework once I become more acquainted with the project. I will implement this after completing the basic uprobe tasks, if it's not urgent.

mauriciovasquezbernal:
Fixed by #2671

Issue #173, opened 2021-05-26
New Gadget: Capability Advisor


marga-kinvolk:
Use case: help developers write the PodSecurityContext with the right capabilities

alban:
The trace capabilities gadget has a lot of non-interesting output. For example:
```
$ kubectl gadget trace capabilities
... CONTAINER        PID     COMM             UID     CAP  NAME             AUDIT  VERDICT 
... set-priority     7788    nice             0       21   CAP_SYS_ADMIN    0      Deny   
$ kubectl exec -ti -n gadget gadget-b5k8r -- bash
# /usr/share/bcc/tools/capable -v -K
09:52:57  0      19252  nice             21   CAP_SYS_ADMIN        0     
        cap_capable+0x1 [kernel]
        security_vm_enough_memory_mm+0x39 [kernel]
        insert_vm_struct+0x9d [kernel]
        __do_execve_file.isra.40+0x3a6 [kernel]
        __x64_sys_execve+0x39 [kernel]
        do_syscall_64+0x5e [kernel]
        entry_SYSCALL_64_after_hwframe+0x44 [kernel]
```

This is a non-audit capability checks. According to [bcc's documentation](https://github.com/iovisor/bcc/blob/e83019bdf6c400b589e69c7d18092e38088f89a8/man/man8/capable.8#L20-L22):
> These are those deemed not interesting and not necessary to audit, such as CAP_SYS_ADMIN checks on memory allocation to affect the behavior of overcommit.

bcc does not show them by default. I think we should do the same.

For the Capability Advisor, we need to skip them, otherwise users will think CAP_SYS_ADMIN is required for everything...

The flag `--print-stack` is still documented but the implementation has been removed in https://github.com/kinvolk/inspektor-gadget/pull/524:
```
$ git grep print-stack
docs/guides/trace/capabilities.md:You may include a kernel call stack for more context with `--print-stack`.  (If
```
I think it would be useful for the Capability Advisor.

alban:
> The trace capabilities gadget has a lot of non-interesting output.

Fixed by https://github.com/kinvolk/inspektor-gadget/pull/914

Possible remaining tasks:
- [ ] --print-stack (at least for the kernel stack; it's more difficult for the userspace stack)
- [ ] generate the PodSecurityContext with the right capabilities
- [ ] Test the UX from a user perspective and identify pain points

tmatens:
[...]

PR #2671, opened 2024-04-01
Support kernel stack map

i-Pear:
Solves #2553

It's with some hacking, and only for test purpose.


i-Pear:
Currently, I am unsure about who should hold the `stackIDMap` (perhaps the front end which prints the stack? Is it localManager or IgManager?), nor do I know how the stack should be printed. I need further understanding of the project's structure.

i-Pear:
The method mentioned in #2553 seems overly complicated; I am unsure why it requires mapping fd:
```
struct gadget_stack_ref {
    int stack_map; // fd of the stack map
    int stack_id; // value returned by bpf_get_stackid()
}
```

Furthermore, its purpose differs from that of `MntNsFilter`. I think we cannot simply copy the code from `MntNsFilter` because `MntNsFilter` needs to be accessed by facilities like `localManager` responsible for filtering container access, whereas `stackIdMap` is only used by the frontend and does not need to be accessed in various parts of the project.

i-Pear:
> containers with different versions of libc. So the address of a function on the stack has to be interpreted differently depending on the libc version. This can be resolved by using a different stack map for each uprobe attachment.

I may lack the necessary knowledge, but why can't stacks from different libraries & different gadgets be mixed within the same stack map? As I understand, the stack map simply associates a stackID with a stack (a set of addresses). In the example eBPF program attached to this PR, stackIDs are recorded alongside PIDs, and then the frontend uses the stackID to find the corresponding stack data, interpreting it using symbols associated with the PID. This doesn't seem to cause any confusion.

alban:
> > containers with different versions of libc. So the address of a function on the stack has to be interpreted differently depending on the libc version. This can be resolved by using a different stack map for each uprobe attachment.
> 
> I may lack the necessary knowledge, but why can't stacks from different libraries & different gadgets be mixed within the same stack map? As I understand, the stack map simply associates a stackID with a stack (a set of addresses).

Correct.

> In the example eBPF program attached to this PR, stackIDs are recorded alongside PIDs, and then the frontend uses the stackID to find the corresponding stack data, interpreting it using symbols associated with the PID. This doesn't seem to cause any confusion.

Do we have to use the PID? I am concerned that if the target process terminates before ig could open `/proc/$pid/maps`, it can't work. I thought that if we know that the event comes from `/bin/bash` and `/lib64/libc.so.6`, we can resolve the addresses to the symbols even when the process terminated. In this case, we need to know if a probe comes from `/lib64/libc.so.6` or another version of libc. But I guess I didn't account for memory relocations of dynamic libraries, so maybe my idea does not work.

flyth:
> @flyth I don't know where to add the code after the refactoring. I see field accessors have a method `Set()` which takes `[]byte` as input, but in the case of a stack passed in a ring buffer, the serialized bytes are not enough because we need to have access to the stack map and do a `bpf(BPF_MAP_LOOKUP_ELEM)`. Could you shed some light on this?

I think I need more info here - but it sounds like you would want to receive the stackID from the ring buffer, do the lookup in userspace and _then_ send whatever you receive through the DataSource (and not (just) the stackID).

alban:
> > @flyth I don't know where to add the code after the refactoring. I see field accessors have a method `Set()` which takes `[]byte` as input, but in the case of a stack passed in a ring buffer, the serialized bytes are not enough because we need to have access to the stack map and do a `bpf(BPF_MAP_LOOKUP_ELEM)`. Could you shed some light on this?
> 
> I think I need more info here - but it sounds like you would want to receive the stackID from the ring buffer, do the lookup in userspace and _then_ send whatever you receive through the DataSource (and not (just) the stackID).

Yes.
1. In the ring buffer, we get i.e. stack_id = 42.
2. We lookup stack_id = 42 in the stack map and we get the value `[]uintptr{0x123, 0x456, 0x789}` (if the stack has a depth of 3).
3. We look at the target process to resolve those 3 addresses. And return the stack `[]string{"getchar", "readline", "main"}` to the user.

flyth:
> > > @flyth I don't know where to add the code after the refactoring. I see field accessors have a method `Set()` which takes `[]byte` as input, but in the case of a stack passed in a ring buffer, the serialized bytes are not enough because we need to have access to the stack map and do a `bpf(BPF_MAP_LOOKUP_ELEM)`. Could you shed some light on this?
> > 
> > 
> > I think I need more info here - but it sounds like you would want to receive the stackID from the ring buffer, do the lookup in userspace and _then_ send whatever you receive through the DataSource (and not (just) the stackID).
> 
> Yes.
> 
> 1. In the ring buffer, we get i.e. stack_id = 42.
> 2. We lookup stack_id = 42 in the stack map and we get the value `[]uintptr{0x123, 0x456, 0x789}` (if the stack has a depth of 3).
> 3. We look at the target process to resolve those 3 addresses. And return the stack `[]string{"getchar", "readline", "main"}` to the user.

What should the UX be for now, then? Just a list of function names returned as string in a "stack" field - in JSON + columns?

Here's how I'd approach it (afaiu the PR):
* prep: [this](https://github.com/inspektor-gadget/inspektor-gadget/blob/abd4dfda26fec0def8735bc2f1de21b9c51c84c2/pkg/operators/ebpf/ebpf.go#L200) needs to be extended to support maps in a generic way (right now we do it for hardcoded names, but we should also populate maps with a given prefix, IMHO); for a quick test, just add another exception a couple of lines below
* create a new operator named "StackOperator" or something like it; make sure the operator registers itself on `init()` of the file like most other operators do and include it where other operators are included (for testing, could manually add in _both_ occurrences in `cmd/common/oci.go`)
* need to implement `DataOperator` interface and additionally the `DataOperatorInstance` interface on another type that can hold the stack map and is returned by the `InstantiateDataOperator()` func
* in the `InstantiateDataOperator()`, check for the stack-map reference (and create/set/cache it) by using GetVar("mapname") & SetVar(); if it's there, also check DataSources in `gadgetCtx` for `stackId` (`type:gadget_stack_id`); if found, add a new field called "stack" to the DataSource and cache the accessor.
* in PreStart(), subscribe to DataSources with the `stackId`; in the callbacks, extract `stackId` using the accessor, do the map lookup + extract info from the target process and set info using the accessor for the "stack" field (for initial version, I'd suggest just a concatenated, comma-separated string)
* in `Stop()`, destroy the map.

See other operators like `OciHandler`, `pkg/datasource/compat` (used by KubeManager + LocalManager) and `pkg/operators/formatters` for general info on how operators are used.

I just talked to @alban and he said it might be much more complex than what I proposed since there could be multiple stack maps per gadget run (per container + per target lib version...). So if you prefer, feel free to go ahead and do a PoC that just prints the results to stdout and we'll find a way to integrate it properly afterwards.

alban:
> What should the UX be for now, then? Just a list of function names returned as string in a "stack" field - in JSON + columns?

I think yes. There could be an option for a multiline output, such as bcc's tcpdrop tool:
https://github.com/iovisor/bcc/blob/6a5602cef2ebd97c351554d53a4f95532db6a568/tools/tcpdrop_example.txt#L7-L38

pkg/operators/ebpf/converters.go
```diff
@@ -105,10 +106,61 @@ func (i *ebpfInstance) initEnumConverter(gadgetCtx operators.GadgetContext) erro
...
+func (i *ebpfInstance) initStackConverter(gadgetCtx operators.GadgetContext) error {
+	kernelSymbolResolver, err := kallsyms.NewKAllSyms()
+	if err != nil {
+		return err
+	}
+	for _, ds := range gadgetCtx.GetDataSources() {
+		fmt.Println(ds)
+		// TODO: don't know how to assign tag to fields, hardcoding here
+		// for _, in := range ds.GetFieldsWithTag("stack") {
+		for _, in := range []datasource.FieldAccessor{ds.GetField("kernel_stack")} {
```
flyth, line 118: You can use `ds.GetFieldsWithTag("type:gadget_kernel_stack")`, since you already have a typedef for it and type names are automatically applied as tags by the ebpf operator.

pkg/datasource/compat/wrapper.go
```diff
@@ -25,9 +25,10 @@ import (
 )
 
 const (
-	MntNsIdType     = "type:gadget_mntns_id"
-	NetNsIdType     = "type:gadget_netns_id"
-	NetNsIdFallback = "name:netns"
+	MntNsIdType       = "type:gadget_mntns_id"
+	NetNsIdType       = "type:gadget_netns_id"
+	NetNsIdFallback   = "name:netns"
+	StackIdKernelType = "type:gadget_kernel_stack"
```
flyth, line 31: This should be moved to `pkg/operators/ebpf/converters.go`, I think.

i-Pear:
Also added `trace_capabilities` gadget, see #173 and #1319. But I don't know how to test it.

pkg/operators/ebpf/converters.go
```diff
@@ -106,10 +112,66 @@ func (i *ebpfInstance) initEnumConverter(gadgetCtx operators.GadgetContext) erro
...
+			in.SetHidden(true, false)
+			out, err := ds.AddField(in.Name()+"_symbolized", api.Kind_String)
+			if err != nil {
+				return err
+			}
+			converter := func(ds datasource.DataSource, data datasource.Data) error {
+				if i.stackIdMap == nil {
+					return errors.New("kernel stack map is not initialized but used. " +
+						"if you are using `gadget_kernel_stack` as event field, " +
+						"try to include <gadget/stack_map.h>")
```
alban, line 134: Should this error be returned in the `converter` function or in the parent function `initStackConverter`?

cc @flyth

pkg/operators/ebpf/ebpf.go
```diff
@@ -496,6 +501,22 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
 	constReplacements := make(map[string]any)
 
+	// create map for kernel stack
+	if _, ok := i.collectionSpec.Maps[KernelStackMapName]; ok {
+		stackIdMapSpec := ebpf.MapSpec{
+			Name:       KernelStackMapName,
+			Type:       ebpf.StackTrace,
+			KeySize:    4,
+			ValueSize:  8 * PerfMaxStackDepth,
+			MaxEntries: 10000,
```
alban, line 511: PerfMaxStackDepth and MAX_ENTRIES should be kept in sync between the Go source and the header file.

Add a comment `// Keep in sync with ...`.

I am also concerned with the growing API surface between ig and the gadget. Since third-party gadgets and ig are not to be released in lock-steps, we could have a gadget compiled with `stack_map.h` from an older version of ig, and then run it with a newer version of ig.

It might not matter for a map of type StackTrace, but for other kind of maps, this can cause problems. So the code pattern makes me uneasy.

I think we can accept it for now, but I am hoping we can use ebpf extensions later on.

cc @mauriciovasquezbernal

pkg/operators/ebpf/converters.go
i-Pear: I can move the check to `initStackConverter`. And it needs the kernel stack map to be initialized in `init` instead of `Start`.

pkg/operators/ebpf/ebpf.go
i-Pear: I agree, since I have no experience in ebpf extensions, I will try it with USDT arguments first.

i-Pear:
TODO: use ebpf extension to refactor this

i-Pear:
The failure in documentation checks could be ignored, once this got merged, the link will be available.

i-Pear:
As extension: https://github.com/i-Pear/inspektor-gadget/tree/stack_map_as_extension

i-Pear:
Changed kernel stack map name to `ig_kstack`.

alban (commented):
Thanks!

pkg/operators/ebpf/converters.go
```diff
@@ -106,10 +112,65 @@ func (i *ebpfInstance) initEnumConverter(gadgetCtx operators.GadgetContext) erro
...
+			out, err := ds.AddField(in.Name()+"_symbolized", api.Kind_String)
+			if err != nil {
+				return err
+			}
+			converter := func(ds datasource.DataSource, data datasource.Data) error {
+				inBytes := in.Get(data)
+				stackId := ds.ByteOrder().Uint32(inBytes)
+				if stackId == 0 {
+					out.Set(data, []byte{})
+					return nil
```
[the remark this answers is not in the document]
i-Pear: The `stack_id` field is in `unsigned` type, and `0` is indeed a valid stack.

gadgets/trace_capabilities/program.bpf.c
```diff
@@ -0,0 +1,334 @@
...
+	gadget_timestamp timestamp;
+	__u32 pid;
+	enum capability cap;
+	__u32 tgid;
+	__u32 uid;
+	__u32 gid;
+	int ret;
+	int audit;
+	int insetid;
+	__u64 syscall;
```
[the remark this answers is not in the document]
i-Pear: Done in #2913

gadgets/trace_capabilities/program.bpf.c
```diff
@@ -0,0 +1,334 @@
...
+	if (gadget_should_discard_mntns_id(mntns_id))
+		return 0;
+
+	u64 nr = ctx->args[1];
+	sc_ctx.nr = nr;
+
+	// The sys_exit tracepoint is not called for some syscalls.
+	if (!skip_exit_probe(nr))
+		bpf_map_update_elem(&current_syscall, &pid_tgid, &sc_ctx,
+				    BPF_ANY);
```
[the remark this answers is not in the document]
i-Pear: Could we put the code in #2475 in a common header? For example, we can add `fix_execve.h` and provide two helper functions `hook_execve_enter` and `hook_execve_exit`. Also, seems the `execveat` syscall needs the same fix.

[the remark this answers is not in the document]
alban: Unfortunately the implementation is slightly different for container-hook vs trace-exec. I didn't find a way to have common code... Maybe it's easier to fix it separately in this trace-capabilities gadget, and do the refactoring in a separate PR.

About execveat: it seems that the trace-exec gadgets (builtin and image-based) miss events from execveat, but that's a separate bug. The ebpf maps are not getting full in that case.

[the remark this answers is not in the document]
i-Pear: The implementations in `trace-exec` and `trace_capabilities` should be the same? `Container-hook` might have different requirements, but I just want to provide a common implementation for gadgets.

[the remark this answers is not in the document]
i-Pear: Opened #2965

gadgets/trace_capabilities/program.bpf.c
```diff
@@ -0,0 +1,334 @@
...
+	event->ret = PT_REGS_RC(ctx);
+	event->kstack = gadget_get_kernel_stack(ctx);
+	event->timestamp = bpf_ktime_get_boot_ns();
+
+	if (LINUX_KERNEL_VERSION >= KERNEL_VERSION(5, 1, 0)) {
+		event->audit = (ap->cap_opt & CAP_OPT_NOAUDIT) == 0;
+		event->insetid = (ap->cap_opt & CAP_OPT_INSETID) != 0;
+	} else {
+		event->audit = ap->cap_opt;
+		event->insetid = -1;
```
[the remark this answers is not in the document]
i-Pear: I don't recommend to use `bpf_core_type_matches` here, because `struct cred` also changed in Linux 6.7-rc6 [1].

Because the `vmlinux.h` in IG is version `6.6`, and my local environment is `6.10`, I spent a lot of time looking for why the `bpf_core_type_matches` returns false. In such cases, seems we need to trace all the nested structures, and provide headers for each version. What do you think?

[1] https://github.com/torvalds/linux/commit/f8fa5d76925991976b3e7076f9d1052515ec1fca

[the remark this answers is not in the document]
i-Pear: I guess your motivation for using `bpf_core_type_matches` is that some distributions may pick patches or do certain backports, making it better to judge the structure than the version number. 

However, we cannot predict which of the two patches (1: updating `struct cred`, 2: updating `union security_list_options`) is included in the user's Linux distribution. This leads to four scenarios, with potentially more combinations in the future.

alban:
@i-Pear #2965 is merged. Could you rebase this PR? Thanks!

i-Pear:
Also opened #3028 to fix a minor issue in builtin gadget.

gadgets/trace_capabilities/gadget.yaml
```diff
@@ -0,0 +1,83 @@
...
+      attributes:
+        template: ns
+    - name: target_userns
+      attributes:
+        template: ns
+    - name: cap_effective
+      attributes:
+        width: 20
+        alignment: left
+        ellipsis: end
```
alban, line 27: Once https://github.com/inspektor-gadget/inspektor-gadget/pull/3020 is merged, we can print cap_effective in a better way.

(not for this PR)

gadgets/trace_capabilities/program.bpf.c
```diff
@@ -0,0 +1,354 @@
...
+	gadget_syscall syscall;
+	gadget_kernel_stack kstack;
+	char task[TASK_COMM_LEN];
+};
+
+#define MAX_ENTRIES 10240
+
+const volatile pid_t my_pid = -1;
+const volatile pid_t targ_pid = -1;
+const volatile u32 linux_version_code = 0;
```
alban, line 92: Those are never set. They could be deleted.

gadgets/trace_capabilities/program.bpf.c
```diff
@@ -0,0 +1,354 @@
...
+const volatile bool audit_only = false;
+const volatile bool unique = false;
```
alban, line 94: You can use GADGET_PARAM().

gadgets/trace_tcpdrop/gadget.yaml
```diff
@@ -65,3 +65,9 @@ structs:
         width: 16
         alignment: left
         ellipsis: end
+    - name: kernel_stack
+      description: Kernel stack
+      attributes:
+        width: 20
+        alignment: left
+        ellipsis: end
```
alban, line 73: We should have a template for stacks.

pkg/operators/ebpf/converters.go
```diff
@@ -106,10 +112,63 @@ func (i *ebpfInstance) initEnumConverter(gadgetCtx operators.GadgetContext) erro
 	return nil
 }
 
+func (i *ebpfInstance) initStackConverter(gadgetCtx operators.GadgetContext) error {
+	kernelSymbolResolver, err := kallsyms.NewKAllSyms()
```
alban, line 116: For performance, could we skip the call to `kallsyms.NewKAllSyms()` if there are no fields with `StackIdKernelType`?

pkg/operators/ebpf/converters.go
```diff
@@ -106,10 +112,63 @@ func (i *ebpfInstance) initEnumConverter(gadgetCtx operators.GadgetContext) erro
...
+			if in == nil {
+				continue
+			}
+			if i.stackIdMap == nil {
+				return errors.New("kernel stack map is not initialized but used. " +
+					"if you are using `gadget_kernel_stack` as event field, " +
+					"try to include <gadget/kernel_stack_map.h>")
+			}
+			in.SetHidden(true, false)
+			out, err := ds.AddField(in.Name()+"_symbolized", api.Kind_String)
```
alban, line 131: @mauriciovasquezbernal @flyth Given the change in #3023, should we no longer add the "_symbolized" suffix?

(Not for this PR, since #3023 is not merged yet)

alban (commented):
Thanks! I think this PR is almost ready. Just some small changes requested.

gadgets/trace_tcpdrop/gadget.yaml
i-Pear: Opened #3041

i-Pear:
Looks strange... The `pkg/operators/ebpf: Support kernel stack map` commit will modify `pkg/gadgets/trace/tcp/tracer/tcptracer_arm64_bpfel.o`, but seems they have no relationship.

And the `amd64_bpfel.o` is not modified, making it more strange...

pkg/operators/ebpf/formatters.go
```diff
@@ -143,10 +150,75 @@ func (i *ebpfInstance) initEnumFormatter(gadgetCtx operators.GadgetContext) erro
...
+	for _, ds := range gadgetCtx.GetDataSources() {
+		for _, in := range ds.GetFieldsWithTag(StackIdKernelType) {
+			if in == nil {
+				continue
+			}
+			in.SetHidden(true, false)
+
+			if kernelSymbolResolver == nil {
+				var err error
+				kernelSymbolResolver, err = kallsyms.NewKAllSyms()
```
alban, line 164: I wonder if we can skip `kallsyms.NewKAllSyms()` if the `in` field is already hidden.

If the `print-stack` ebpfParam is false, then we don't need to spend CPU time parsing kallsyms.

But that would be an optimization for another PR.

alban (approved):
Thanks for the work, LGTM!

I'll wait to see the CI results. Then, I can merge this PR.

Issue #2560, opened 2024-03-01
[Discussion] image-based gadgets: How to mark the maps used by a gadget to generate information?


mauriciovasquezbernal:
## Introduction

The gadgets generate information using different kind of maps: tracers use perf event arrays or ring buffers, snapshotters use ebpf iterators and toppers and profilers use hashmaps. Inspektor Gadget needs to know what are the maps (and the types) that are used by the gadget to generate the information in order to be able to run the gadget and collect the information it generates. 

Currently Inspektor Gadget supports two ways to set this information. Any of the two methods needs to be used. If both methods are used, Inspektor Gadget validates that them contain the same information.

### Metadata file 

The metadata file, besides other things, allows to set the kind of information that is provided by the gadget:

- https://github.com/inspektor-gadget/inspektor-gadget/blob/8c433c4b29399cabd6ac233933c602f76c3c7c5f/gadgets/trace_exec/gadget.yaml#L3-L6
- https://github.com/inspektor-gadget/inspektor-gadget/blob/79eb65512419557898554ad063536b59bb38bf1c/gadgets/top_file/gadget.yaml#L3-L6
- https://github.com/inspektor-gadget/inspektor-gadget/blob/0ac7b3d565b6f5775cccbe3595c1eea5a6b91871/gadgets/snapshot_process/gadget.yaml#L3-L5

### `GADGET_` macros

There are some `GADGET_` macros defined in in [include/gadget/macros.h](https://github.com/inspektor-gadget/inspektor-gadget/blob/7cdf64eac0d1f943152b602abee2421e961f355e/include/gadget/macros.h) that allow to set this information directly in the ebpf code of the gadget: 

- https://github.com/inspektor-gadget/inspektor-gadget/blob/9949c29dfe45e9622d8e5f5e603c954ea25294e2/gadgets/trace_exec/program.bpf.c#L60
- https://github.com/inspektor-gadget/inspektor-gadget/blob/79eb65512419557898554ad063536b59bb38bf1c/gadgets/top_file/program.bpf.c#L68
- https://github.com/inspektor-gadget/inspektor-gadget/blob/f01ffa9821808fb7b40673c873110f83dc0de2f6/gadgets/snapshot_process/program.bpf.c#L32

The current implementation uses the metadata file as the single source of truth, these macros are used  to populate the metadata if this is missing when running the gadget

[...]

PR #2673, opened 2024-04-03
treewide: remove run gadget + types

flyth:
This removes the former run gadget and most of its types. The image build process needs to be updated to use the new methods after the refactoring, but #2560 probably needs to be discussed first.


blanquicet (approved):
LGTM!

Looking forward to merging https://github.com/inspektor-gadget/inspektor-gadget/pull/2607 so that we can start using that framework for testing and remove what we couldn't here.

flyth:
Thanks for the review!

PR #2675, opened 2024-04-03
docs: Document how to verify container image at runtime using policy-controller

eiffel-fl:


docs/getting-started/verify.md
```diff
@@ -32,6 +32,69 @@ The following checks were performed on each of these signatures:
 Getting the above output followed by a JSON array of payloads, ensures you the
 container image was signed using our private key.
 
+## Verify the container image at runtime
```
burak-ok, line 35:
```suggestion
## Verify the container image while deploying
```

burak-ok (approved):
Tested it locally and it works as described. Thanks :)

docs/getting-started/verify.md
```diff
@@ -32,6 +32,69 @@ The following checks were performed on each of these signatures:
...
+1/1 gadget pod(s) ready
+...
+Inspektor Gadget successfully deployed
+```
+
+As you can see, everything was successfully deployed.
+Now, let's undeploy Inspektor Gadget and try to deploy an old release which was not signed:
+
+```bash
+$ kubectl gadget undeploy
```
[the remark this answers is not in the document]
blanquicet: Good question. Of course, we should ensure every resource we deploy is indeed removed after `undeploy`.

docs/getting-started/verify.md
```diff
@@ -32,6 +32,69 @@ The following checks were performed on each of these signatures:
...
 container image was signed using our private key.
 
+## Verify the container image at runtime
+
+You can also verify the container image at runtime by using `policy-controller`.
+To do so, you first need to [install](https://docs.sigstore.dev/policy-controller/installation/) this component on your kubernetes cluster.
+We will then deploy an `admission-controller` using the following YAML file:
+
+```yaml
+apiVersion: policy.sigstore.dev/v1beta1
```
blanquicet, line 42: What about enabling this by default during deployment (like you're proposing for [gadget images](https://github.com/inspektor-gadget/inspektor-gadget/pull/2678#discussion_r1551061245))?

I mean, by default, we verify if the `policy-controller` is installed in the cluster:
- If it's, we deploy the `admission-controller` and verify the image. 
- If it isn't, we print something like "Warn: The policy-controller wasn't found: Impossible to verify image" or so.

Finally, if users don't want this verification, they use the flag `--verify-image=false`.

blanquicet (approved):
LGTM!

docs/getting-started/verify.md
blanquicet: > I just need to think how to articulate between `kubectl-gadget` and `ig`
>
What do you mean? This is only for verifying the container image for the gadget daemonset. How is this related to `ig`? Did I miss something (again 😅)?

docs/getting-started/verify.md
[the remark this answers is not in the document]
blanquicet: Maybe something you can ask opinions during our next sync dev meeting.

docs/getting-started/verify.md
blanquicet: All clear now. Thanks for working on this. I had to read about cosing and all signing stuff but it was fun 🙂

PR #2676, opened 2024-04-03
eks: Add timestamp tag

burak-ok:
# eks: Add timestamp tag

In a future PR  a periodic script to delete all cloudformation stacks and vpcs older than a day will be added. For that a timestamp is needed

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1603,7 +1603,7 @@ jobs:
         if [ ${{ matrix.arch }} = 'arm64' ]; then
           node_type='a1.xlarge'
         fi
-        eksctl create cluster --name ${{ env.CLUSTER_NAME }} --tags ig-ci=true --node-type $node_type
+        eksctl create cluster --name ${{ env.CLUSTER_NAME }} --tags ig-ci=true,ig-ci-timestamp=$(date +%s) --node-type $node_type
     - name: Run integration tests
       uses: ./.github/actions/run-integration-tests
       with:


.github/workflows/inspektor-gadget.yml
```diff
@@ -1603,7 +1603,7 @@ jobs:
         if [ ${{ matrix.arch }} = 'arm64' ]; then
           node_type='a1.xlarge'
         fi
-        eksctl create cluster --name ${{ env.CLUSTER_NAME }} --tags ig-ci=true --node-type $node_type
+        eksctl create cluster --name ${{ env.CLUSTER_NAME }} --tags ig-ci=true,ig-ci-timestamp=$(date +%s) --node-type $node_type
```
alban, line 1606: It would be more readable with something like:
```
$ date --iso-8601=seconds
2024-04-03T15:47:20+02:00
```

burak-ok: I chose this representation so I could make the condition easier:
`if [ $cf_stack_timestamp -gt $(date -d "3 hours ago" +%s) ]; then`

Of course I am no bash expert :D Is there a easy way with your more readable representation?

burak-ok:
> A test on `citest/` would be welcomed. This commit could be part of the bigger PR itself.

Yeah, totally forgot that we have `citest/` and its true, this should be part of the bigger PR

> I am not that convinced by the approach as this is more a workaround than tackling the root cause, but I guess this cannot harm.

This is not more of a workaround - This is a workaround :D But until the underlying (external) issue is solved this should remedy some of the negative effects

burak-ok:
Closing in favor of a bigger PR

PR #2678, opened 2024-04-03
Sign image based gadgets

eiffel-fl:
Hi.

In this PR, I added signing of image based gadgets.

Best regards.

### DIFF
--- a/docs/getting-started/verify.md
+++ b/docs/getting-started/verify.md
@@ -168,6 +168,22 @@ $ jq '' /tmp/gadget-container-image-linux-amd64/sbom_cyclonedx.json
 
 As the SBOM was signed with our private key, you can now inspect it to track down every dependencies we use to build our container image.
 
+## Verify image-based gadgets
+
+Like our container image, we sign all our image-based gadgets.
+The instructions to verify them are similar to the container image:
+
+```bash
+$ cosign verify --key https://raw.githubusercontent.com/inspektor-gadget/inspektor-gadget/main/pkg/resources/inspektor-gadget.pub ghcr.io/inspektor-gadget/gadget/trace_exec:latest
+The following checks were performed on each of these signatures:
+  - The cosign claims were validated
+  - Existence of the claims in the transparency log was verified offline
+  - The signatures were verified against the specified public key
+
+[{"critical":{"identity":{"docker-reference":"ghcr.io/inspektor-gadget/gadget/trace_exec"}, ...
+]
+```
+
 ## Verify an asset
 
 Rather than signing all the assets, we only sign the checksums file.


pkg/oci/oci.go
```diff
@@ -478,16 +490,85 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
 	}, nil
 }
 
+func prepareCheckOpts(ctx context.Context, publicKey string) (*cosign.CheckOpts, error) {
+	pubKey, err := signature.LoadPublicKeyRaw([]byte(publicKey), crypto.SHA256)
```
flyth, line 494: Is it always gonna be `crypto.SHA256` or do we have to make that customizable somehow?

pkg/oci/oci.go
```diff
@@ -478,16 +490,85 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+	ref, err = sign.GetAttachedImageRef(ref, "", ociremote.WithRemoteOptions(googleremote.WithContext(ctx)))
+	if err != nil {
+		return nil, fmt.Errorf("resolving attachment for image %s: %w", image, err)
+	}
+
+	return ref, nil
+}
+
+func verifyImage(image string, publicKey string) error {
+	ctx := context.TODO()
```
flyth, line 531: Maybe add context as parameter and forward from upper layers?

pkg/operators/oci-handler/oci.go
```diff
@@ -96,6 +99,20 @@ func (o *ociHandler) InstanceParams() api.Params {
...
+			DefaultValue: "true",
+			TypeHint:     api.TypeBool,
+		},
+		{
+			Key:          publicKey,
+			Title:        "Public key",
+			Description:  "Public key used to verify the image based gadget",
+			DefaultValue: resources.InspektorGadgetPublicKey,
+			TypeHint:     api.TypeString,
+		},
```
flyth, line 115: While I think it's okay to have these in here for testing (and also because it's not 100% cleanly implemented otherwise), those should eventually become `GlobalParams` that are initialized from values set in the deployment/config file instead of being transferred using gRPC from the client (which all these InstanceParams currently are) to ensure safety. So maybe add a TODO for now so we don't forget.

pkg/operators/oci-handler/oci.go
```diff
@@ -96,6 +99,20 @@ func (o *ociHandler) InstanceParams() api.Params {
...
+		{
+			Key:          verifyImage,
+			Title:        "Verify image",
+			Description:  "Verify image using the provided public key",
+			DefaultValue: "true",
+			TypeHint:     api.TypeBool,
+		},
+		{
+			Key:          publicKey,
+			Title:        "Public key",
```
flyth, line 111: I think later on we need to support multiple public keys and/or provide a tool to quickly re-sign our binaries.

pkg/operators/oci-handler/oci.go
```diff
@@ -96,6 +99,20 @@ func (o *ociHandler) InstanceParams() api.Params {
 			Description: "Secret to use when pulling the gadget image",
 			TypeHint:    api.TypeString,
 		},
+		{
+			Key:          verifyImage,
+			Title:        "Verify image",
+			Description:  "Verify image using the provided public key",
+			DefaultValue: "true",
```
flyth, line 106: I'm not sure whether we should enable this by default - if people want to try out stuff from artifact hub, this might confuse more than it helps.

pkg/oci/oci.go
```diff
@@ -478,16 +490,85 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
 	}, nil
 }
 
+func prepareCheckOpts(ctx context.Context, publicKey string) (*cosign.CheckOpts, error) {
+	pubKey, err := signature.LoadPublicKeyRaw([]byte(publicKey), crypto.SHA256)
+	if err != nil {
+		return nil, fmt.Errorf("loading public key: %w", err)
+	}
+
+	rekorPubKeys, err := cosign.GetRekorPubs(ctx)
```
flyth, line 499: Maybe add a comment on what `RekorPubs` are.

pkg/oci/oci.go
```diff
@@ -478,16 +490,85 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+	if err != nil {
+		return nil, fmt.Errorf("getting Rekor public keys: %w", err)
+	}
+
+	cTLogPubKeys, err := cosign.GetCTLogPubs(ctx)
```
flyth, line 504: Also maybe add a comment about what CTLogPubs are.

flyth (commented):
This greatly enhances security for critical environments, thanks! A couple of comments from a first pass.

pkg/operators/oci-handler/oci.go
flyth: Sure!

Operators have two config parts - a global and a per-gadget one (instance params). The global part should be initialized with configuration values from the host itself (e.g. params given to the server binary on the host or a config file on that machine), while all the per-gadget ones are forwarded to the client and can be filled out from there and sent back. So everything concerning security should be coming from the global part or at least be restricted from there (e.g. "a user may override this or that setting" - where user would refer to a client connection).

On `ig`, both params are treated similarly, as it's both client and server/host.

We sadly still don't have support for config files / generic params from the deployment, so the global params don't get used a lot right now. IMHO, this should definitely be handled before removing the experimental flag from image based gadgets.

pkg/operators/oci-handler/oci.go
flyth: Hmm, I think you're missing the client/server architecture in the picture. "Use `run` as many times as they have different keys" isn't the case there. You would deploy IG and configure it to "only trust gadgets signed by X" and can't change that with just running another process on your client (see above comment regarding global params vs instance params as well.)

So I think configuring it as "only trust IG official gadgets + our own" could be a thing that might be asked for. If we only want to support a single public key at a time, we should at least provide some help explaining how to quickly resign binaries (I'm not suggesting we build our own binary/app for that, but at least docs and maybe a script.)

go.mod
```diff
@@ -67,30 +67,90 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
-require github.com/google/gopacket v1.1.19
+require (
```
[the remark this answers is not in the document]
flyth: It's a crazy amount of stuff, tbh.

`github.com/go-chi/chi` is a HTTP router (used for HTTP _servers_).
`github.com/go-playground/universal-translator` is a i18n library.
`github.com/hashicorp/hcl` is a "structured configuration language"
`github.com/lestrrat-go/blackmagic` just says "Reflect-based black magic. YMMV, and use with caution"
`github.com/outcaste-io/ristretto` says "Ristretto is a fast, concurrent cache library built with a focus on performance and correctness."
`github.com/syndtr/goleveldb` "This is an implementation of the [LevelDB key/value database](https://github.com/google/leveldb) in the [Go programming language](https://go.dev/)"

Just picked a couple of random ones :)

pkg/operators/oci-handler/oci.go
flyth: Yeah, you're right, it's easy enough to disable it.

blanquicet: I agree it should be done by default.

pkg/resources/inspektor-gadget.pub
blanquicet, line 1: Please update [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/0d6558db5bb08723be1bb69beb4bde08d073b67a/docs/getting-started/verify.md?plain=1#L19).

.github/workflows/inspektor-gadget.yml
```diff
@@ -1344,6 +1347,20 @@ jobs:
...
+      - name: Sign the manifest list
+        if: needs.check-secrets.outputs.cosign == 'true' && ${{ github.event_name }} == 'push'
+        env:
+          COSIGN_PASSWORD: '${{ secrets.COSIGN_PASSWORD }}'
+          COSIGN_PRIVATE_KEY: '${{ secrets.COSIGN_PRIVATE_KEY }}'
+        run: |
+          for i in $(make -C gadgets --no-print-directory list_gadgets); do
+            gadget="${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }}/${i}:${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}"
+
+            cosign sign --key env://COSIGN_PRIVATE_KEY --yes --recursive $gadget
```
blanquicet, line 1362: The following warning message is displayed:

```
WARNING: Image reference ttl.sh/fb9571407fc55a27104917d3d3e7a36ee8735e0e/trace_dns:2678-merge uses a tag, not a digest, to identify the image to sign.
    This can lead you to sign a different image than the intended one. Please use a
    digest (example.com/ubuntu@sha256:abc123...) rather than tag
    (example.com/ubuntu:latest) for the input to cosign. The ability to refer to
    images by tag will be removed in a future release.
```

We can manage this in another PR, but please create an issue for it.

pkg/oci/oci.go
```diff
@@ -478,16 +490,85 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
 // newRepository creates a client to the remote repository identified by
 // image using the given auth options.
 func newRepository(image reference.Named, authOpts *AuthOptions) (*remote.Repository, error) {
-	repo, err := remote.NewRepository(image.Name())
+	img := image.Name()
+
+	if authOpts.CheckPublicKey {
+		publicKey := authOpts.PublicKey
+
+		err := verifyImage(img, publicKey)
```
blanquicet, line 559: Mmmm it's only executed for pull policy "always" and "missing". It means it won't be executed for images already present in the local storage. Is it fine? 🤔

pkg/oci/oci.go
```diff
@@ -37,12 +38,23 @@ import (
...
+	"github.com/google/go-containerregistry/pkg/name"
+	googleremote "github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
 type AuthOptions struct {
 	AuthFile    string
 	SecretBytes []byte
 	Insecure    bool
+
+	CheckPublicKey bool
```
blanquicet, line 56: nit: `VerifyImage` seems more appropriate than `CheckPublicKey`.

blanquicet (commented):
Awesome! 💯

pkg/oci/oci.go
blanquicet: Mmmm afaiu, rekor and ct logs are only used for keyless signing, which is not our case. So, we can omit them here.

pkg/oci/oci.go
```diff
@@ -478,16 +490,85 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+	if err != nil {
+		return fmt.Errorf("preparing check options: %w", err)
+	}
+
+	ref, err := prepareImageReference(ctx, image)
+	if err != nil {
+		return fmt.Errorf("preparing image reference: %w", err)
+	}
+
+	_, _, err = cosign.VerifyImageSignatures(ctx, ref, co)
```
blanquicet, line 543: If I'm using an insecure registry, don't we need to specify it here somehow? Same for any other auth options like the authentication file? I don't know, maybe I'm very probably missing something 😅

docs/getting-started/verify.md
```diff
@@ -107,6 +107,22 @@ $ jq '' /tmp/gadget-container-image-linux-amd64/sbom_cyclonedx.json
 
 As the SBOM was signed with our private key, you can now inspect it to track down every dependencies we use to build our container image.
 
+## Verify image based gadgets
+
+Like our container image, we sign all our image based gadgets.
+The instructions to verify them are similar to the container image:
+
+```bash
+$ cosign verify --key inspektor-gadget.pub ghcr.io/inspektor-gadget/gadget/trace_exec:latest
```
blanquicet, line 116: What about this?

```suggestion
$ cosign verify --key https://raw.githubusercontent.com/inspektor-gadget/inspektor-gadget/main/inspektor-gadget.pub ghcr.io/inspektor-gadget/gadget/trace_exec:latest
```

gadgets/Makefile
```diff
@@ -46,6 +46,9 @@ push: build
 		sudo -E IG_EXPERIMENTAL=true $(IG) image push $(GADGET_REPOSITORY)/$$GADGET:$(GADGET_TAG) || exit 1 ; \
 	done
 
+list_gadgets:
```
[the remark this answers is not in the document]
blanquicet: Those are the names of the gadgets themselves, `list-gadgets` is a Makefile target.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1344,6 +1348,20 @@ jobs:
...
+        if: needs.check-secrets.outputs.cosign == 'true' && github.event_name == 'push'
+        env:
+          COSIGN_PASSWORD: '${{ secrets.COSIGN_PASSWORD }}'
+          COSIGN_PRIVATE_KEY: '${{ secrets.COSIGN_PRIVATE_KEY }}'
+        run: |
+          for i in $(make -C gadgets --no-print-directory list_gadgets); do
+            gadget="${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }}/${i}:${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}"
+
+            cosign sign --key env://COSIGN_PRIVATE_KEY --yes --recursive $gadget
+          done
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Why would you need to store it on a Makefile variable? It's using `--key env://COSIGN_PRIVATE_KEY`, hence having it on env is enough.

pkg/operators/oci-handler/oci.go
alban: The documentation in docs/devel/hello-world-gadget.md should be updated to explain how to sign third-party gadgets (next to the section "Pushing the gadget image to a container registry").

(this can be in a separate PR if you prefer)

docs/getting-started/verify.md
```diff
@@ -107,6 +107,22 @@ $ jq '' /tmp/gadget-container-image-linux-amd64/sbom_cyclonedx.json
 
 As the SBOM was signed with our private key, you can now inspect it to track down every dependencies we use to build our container image.
 
+## Verify image based gadgets
```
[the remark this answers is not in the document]
alban: This comment seems to have been resolved by mistake? Ditto below.

alban (commented):
This looks good to me, apart of a couple of small comments I added above.

But since I'm not so familiar with the signing, I would like someone else to review as well.

docs/getting-started/verify.md
mauriciovasquezbernal: What about updating it now?

docs/devel/hello-world-gadget.md
```diff
@@ -207,6 +207,28 @@ Pushing ghcr.io/my-org/mygadget:latest...
 Successfully pushed ghcr.io/my-org/mygadget:latest@sha256:dd3f5c357983bb863ef86942e36f4c851933eec4b32ba65ee375acb1c514f628
 ```
```
mauriciovasquezbernal, line 209: (Optional) ... ?

pkg/operators/oci-handler/oci.go
mauriciovasquezbernal: > The documentation in docs/devel/hello-world-gadget.md should be updated to explain how to sign third-party gadgets (next to the section "Pushing the gadget image to a container registry").

Do we really want to continue adding more information to the hello-world guide? IMO it's becoming too complex. What about creating a new file under `docs/reference` with this information?

gadgets/Makefile
```diff
@@ -46,6 +47,13 @@ push: build
 		sudo -E IG_EXPERIMENTAL=true $(IG) image push $(GADGET_REPOSITORY)/$$GADGET:$(GADGET_TAG) || exit 1 ; \
 	done
 
+sign: push
+	@echo "Signing all gadgets"
+	for GADGET in $(GADGETS); do \
+		digest=$$(sudo -E IG_EXPERIMENTAL=true $(IG) image list --no-trunc | grep "$$GADGET " | awk '{ print $$3 }') ; \
+		cosign sign --key env://COSIGN_PRIVATE_KEY --yes --recursive $(GADGET_REPOSITORY)/$$GADGET@$$digest || exit 1 ; \
```
mauriciovasquezbernal, line 54: Is there any way to avoid having to introduce the key's password N times when running this command?

pkg/operators/oci-handler/oci.go
mauriciovasquezbernal: > I can drop the last commit if needed. 

I'd prefer to drop it.

mauriciovasquezbernal (approved):
I have two comments but they're not blocking. 

LGTM!

pkg/operators/oci-handler/oci.go
blanquicet: > There is definitely something I do not understand with the `GlobalParams()`, as their values seem to never change:
> 
Operators' global params will not work because we are always using default values:

https://github.com/inspektor-gadget/inspektor-gadget/blob/162b48b11b1f7bed5995a187bb9b9465f17184d6/pkg/gadget-context/run.go#L44-L51

So, I suggest you to keep the verification params as instance params for this PR, until we properly implement support for setting operators' global params.

blanquicet: > > The documentation in docs/devel/hello-world-gadget.md should be updated to explain how to sign third-party gadgets (next to the section "Pushing the gadget image to a container registry").
> 
> Do we really want to continue adding more information to the hello-world guide? IMO it's becoming too complex. What about creating a new file under `docs/reference` with this information?

IMO, we should add one sentence saying that we support signing and then a link to another document where we explain everything about it:

- By default, IG gadgets are signed.
- By default, IG will verify gadgets using the IG pub key.
- Verification ca be skipped using flag.
- People can use and create their own key-pair. Add an example of how to do it.
- And whatever I missed here 🙂

pkg/operators/oci-handler/oci.go
blanquicet: Let's keep it as TODO in an issue for another PR.

PR #2683, opened 2024-04-04
socket enricher: add new fields: ppid, pcomm, exepath, cwd

alban:
This patch adds new fields in the socket enricher. It is not used by any gadget yet, but it makes it possible to add the new fields in gadgets using the socket enricher.

## How to use

If you want to make use of the new fields in your own gadget, you could write something like:
```patch
--- a/pkg/gadgets/trace/dns/tracer/bpf/dns.c
+++ b/pkg/gadgets/trace/dns/tracer/bpf/dns.c
@@ -234,8 +234,11 @@ int ig_trace_dns(struct __sk_buff *skb)
                event.mount_ns_id = skb_val->mntns;
                event.pid = skb_val->pid_tgid >> 32;
                event.tid = (__u32)skb_val->pid_tgid;
+               event.ppid = skb_val->ppid;
                __builtin_memcpy(&event.task, skb_val->task,
                                 sizeof(event.task));
+               __builtin_memcpy(&event.pcomm, skb_val->pcomm,
+                                sizeof(event.pcomm));
                event.uid = (__u32)skb_val->uid_gid;
                event.gid = (__u32)(skb_val->uid_gid >> 32);
        }
```

See PR #2684 for an example in the dns gadget.

## Testing done

Let's check the CI.


alban (commented):

| Benchmark suite | Current: a38b673155e65cf1da697588f70b5ea903242f91 | Previous: ad488f391d1e1b0e72d28a6b4dcd5d12ad6b23a2 | Ratio |
|-|-|-|-|
| `BenchmarkAllGadgetsWithContainers/container0/trace-dns` | `26879440` ns/op | `8837005` ns/op | `3.04` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-network` | `19798211` ns/op | `2265452` ns/op | `8.74` |

CC: @alban

alban:
Branch updated.

@mauriciovasquezbernal I also tried ig from this branch with an old trace_dns gadget before this patch:
```
$ sudo -E ig image pull trace_dns:latest
$ sudo -E ig run trace_dns:latest
...
Error: starting operator "oci": starting operator "ebpf": creating eBPF collection: using replacement map gadget_sockets: ValueSize: 8272 changed to 64: map spec is incompatible with existing map
```

The `MapReplacements` in cilium/ebpf fails even when the new map is bigger than the old one.

When I remove the check in cilium/ebpf, the kernel seems to be fine with the larger map.

mauriciovasquezbernal (commented):
Do we have any idea of how expensive it's to get those new fields? I'm getting worried that we're adding so many fields and that'll impact the performance. There should be at least a way to opt-in for those "heavy to get" fields.

alban:
> Do we have any idea of how expensive it's to get those new fields? I'm getting worried that we're adding so many fields and that'll impact the performance. There should be at least a way to opt-in for those "heavy to get" fields.

I don't know how expensive it is. The problem with opt-in for heavy fields is that the socket enricher is a singleton: if several gadgets need it, it will be started only once. If it is started at first without heavy fields, the second gadget that needs it won't be able to change its configuration.

alban:
Closing. This patch will be developed as part of #2684.

PR #2685, opened 2024-04-04
kubernetes: cgroupfs: mount read-only

alban:
# kubernetes: cgroupfs: mount read-only

Inspektor Gadget needs access to the host volume /sys/fs/cgroup but does not need write access. Switch to readOnly=true.

## How to use

No changes.

## Testing done

None yet. Let's run the CI first.

Related PR: https://github.com/inspektor-gadget/inspektor-gadget/pull/2285


alban:
Various tests fail:
- AKS: TestWatchCreatedContainers
- ARO: TestAdviseNetworkpolicy
- GKE: TestRunTraceOOMKill

I am trying to run them again.

Issue #2493, opened 2024-02-13
Support a Configuration File


mauriciovasquezbernal:
Inspektor Gadget has a lot of parameters that can be configured at run time. Currently, all those parameters must be specified in the CLI, however this approach is not flexible, and it should be possible to define those parameters on a file to: 

- Avoid the user having to type them all the time it uses the CLI. It should be possible to override the values by using the CLI though.  
- Allow the user to configure Inspektor Gadget when running gadgets without the CLI as describe in the “Declarative way to run gadgets” case. 

These are the parameters we should support through the configuration file. Many of them are already exposed in the CLI (try running kubectl gadget or ig without any argument to see them) 

### Workarounds 

- Auto mount file systems: Automatically mount bpffs, debugfs and tracefs if they are not already mounted 
- Auto sd unit restart: Automatically run in a privileged systemd unit if lacking enough capabilities 
- Auto wsl workaround: Automatically find the host procfs when running in WSL2 

### Container runtimes (docker, cri-o, containerd, podman) 

- Namespace (only for containerd?) 
- Socket paths  
- Enable/disable: if IG should try to contact that container runtime 

### OCI registry  

- Credentials 
- Bool flag indicating if this is an insecure registry 
- Security policies (TODO add link to security issue) 

### Exporters  
- Prometheus 
  - Metrics listen address: Address to serve Prometheus metrics on 
  - Metrics path: Path to export Prometheus metrics on 

## kubectl-gadget 

- API server connection: How to connect to the k8s API server. (It seems some of them are inherited from the kubeconfig) 
  - connection-timeout 
  - request-timeout 
  - client-certificate 
  - client-key 
  - insecure-skip-tls-verify 
  - kubeconfig 
  - server 
  - tls-server-name 
   - token 
- Specific IG connection params 
  - tcp-port 
- gadget-namespace 
- verbose 

## gadgetctl 

- connection-timeout 
- remote-address: Comma-separated list of remote address (gRPC) to connect to (default "unix:///var/run/ig/ig.socket") 
- Verbose 

## Implementation options 

A lot of projects already use any sort of configuration file, so it’s a well-known approach in the industry. Projects like https://github.com/spf13/viper should contain all the support we need for this.

mqasimsarfraz:
Although https://github.com/inspektor-gadget/inspektor-gadget/pull/2831 was merged I will keep this issue for follow-up work:
- Support for ig-k8s
- Add unit/integration tests
[...]

PR #2707, opened 2024-04-10
Verify image based gadget

eiffel-fl:
Hi.

This PR adds code to verify previously signed image based gadgets:

```bash
$ sudo -E ./ig run ghcr.io/eiffel-fl/gadget/trace_open
ERRO[0001] instantiating operator "oci": ensuring image: verifying image "ghcr.io/eiffel-fl/gadget/trace_open": verifying signature: invalid signature when validating ASN.1 encoded signature 
ERRO[0003] instantiating operator "oci": ensuring image: verifying image "ghcr.io/eiffel-fl/gadget/trace_open": verifying signature: invalid signature when validating ASN.1 encoded signature 
^C%
$ sudo -E ./ig run --public-key="$(cat eiffel-fl.pub)" ghcr.io/eiffel-fl/gadget/trace_open
RUNTIME.CONTAINERNAME  PID          UID          GID          MNTNS_ID RET FL… MODE        COMM        FNAME                  TIMESTAMP
```

Best regards.

### DIFF
--- a/docs/getting-started/verify.md
+++ b/docs/getting-started/verify.md
@@ -164,6 +164,7 @@ $ jq '' /tmp/gadget-container-image-linux-amd64/sbom_cyclonedx.json
       "publisher": "Santiago Vila <sanvila@debian.org>",
       "name": "base-files",
 ...
+}
 ```
 
 As the SBOM was signed with our private key, you can now inspect it to track down every dependencies we use to build our container image.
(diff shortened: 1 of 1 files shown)


blanquicet (commented):
The logic seems clear to me, we get the cosign image signature for our index, verify it's the signature for our index (DockerManifestDigest) and extract the signature. And finally, we verify that signature with our public key.

Some comments and many questions 😅

pkg/oci/oci.go
```diff
@@ -478,16 +490,141 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+	if err != nil {
+		return fmt.Errorf("parsing public key: %w", err)
+	}
+
+	verifier, err := signature.LoadVerifier(pub, crypto.SHA256)
+	if err != nil {
+		return fmt.Errorf("loading verifier: %w", err)
+	}
+
+	err = verifier.VerifySignature(bytes.NewReader(decodedSignature), strings.NewReader(payload))
```
blanquicet, line 608: I don't understand why we need to pass the payload. The only useful info I see there is the `docker-manifest-digest` and we already verified it. Is it verified again?

```json
{
  "critical": {
    "identity": {
      "docker-reference": "172.28.128.1:5000/snapshot_socket"
    },
    "image": {
      "docker-manifest-digest": "sha256:4e0a8becac464f4cd37cc871a654639fba6b36c81c6841cede72140231447120"
    },
    "type": "cosign container image signature"
  },
  "optional": null
}
```

pkg/oci/oci.go
```diff
@@ -478,16 +490,141 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
 	}, nil
 }
 
+func craftSignatureTag(digest string) (string, error) {
+	// WARNING:
+	// https://sigstore.slack.com/archives/C0440BFT43H/p1712253122721879?thread_ts=1712238666.552719&cid=C0440BFT43H
+	// https://github.com/sigstore/cosign/pull/3622
```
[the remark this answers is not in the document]
blanquicet: Thanks for the explanation. What do you think about adding something like `WARNING: Cosign is considering changing the scheme for publishing/retrieving Sigstore bundles to/from an OCI registry` to the code and keeping the links?

pkg/oci/oci.go
blanquicet: Aahh with that payload we ask the verifier to verify with its pub key and the signature we are passing, the manifest with digest `docker-manifest-digest`. Makes sense. Thanks!

pkg/oci/oci.go
```diff
@@ -478,16 +490,141 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+	}
+
+	if img.Critical.Image.DockerManifestDigest != imageDigest{
+		return "", "", fmt.Errorf("payload digest does not correspond to image: expected %s, got %s", imageDigest, img.Critical.Image.DockerManifestDigest)
+	}
+
+	return signature, string(payloadBytes), nil
+}
+
+func verifyImage(ctx context.Context, image string, authOpts *AuthOptions) error {
```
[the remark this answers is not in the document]
blanquicet: Yeah, based on the way we are signing, it seems enough just to verify the index.

pkg/oci/oci.go
```diff
@@ -478,16 +490,141 @@ func newAuthClient(repository string, authOptions *AuthOptions) (*oras_auth.Clie
...
+
+	img := &payload.SimpleContainerImage{}
+	err = json.Unmarshal(payloadBytes, img)
+	if err != nil {
+		return "", "", fmt.Errorf("decoding payload: %w", err)
+	}
+
+	if img.Critical.Image.DockerManifestDigest != imageDigest{
+		return "", "", fmt.Errorf("payload digest does not correspond to image: expected %s, got %s", imageDigest, img.Critical.Image.DockerManifestDigest)
+	}
```
mauriciovasquezbernal, line 577: According to my understanding from https://github.com/containers/image/blob/main/docs/containers-signature.5.md#the-cryptographic-signature, this logic should only be executed after `VerifySignature` returns nil:

> Consumers of container signatures SHOULD verify the cryptographic signature against one or more trusted public keys (e.g. defined in a [policy.json signature verification policy file](https://github.com/containers/image/blob/main/docs/containers-policy.json.5.md)) before parsing or processing the JSON payload in any way, in particular they SHOULD stop processing the container signature if the cryptographic signature verification fails, without even starting to process the JSON payload.

mauriciovasquezbernal (commented):
Few quick comments before turning off my PC for the weekend. Will check it out with more care next week.

blanquicet (commented):
I still need to test it but looks great from code inspection.

mauriciovasquezbernal (commented):
I tested and it looks good.

This work creates the foundations for the image verification, but I think we still need to improve how the user can define the public keys. Something we can do once we have #2493.

PR #2708, opened 2024-04-10
pkg/datasource: Don't use sync.Pool for data

blanquicet:
# Don't use sync.Pool for data in data sources

The [sync.Pool](https://pkg.go.dev/sync#Pool) was useful in a previous refactoring iteration where we were creating fixed-sized buffers, which is not the case anymore.

## Testing done

CI is green 🙂.


flyth (approved):
LGTM, thanks!

PR #2652, opened 2024-03-27
re-add support for jsonpretty + yaml to image based gadgets

flyth:
This adds the missing `jsonpretty` and `yaml` output modes to the cli operator. Currently (as was with the previous
run gadget), YAML is handled by first marshaling to JSON and then converting, which is quite slow. We can later
on add an optimized YAML formatter that will handle marshaling directly.

This also sorts fields alphabetically.

```
$ ./ig run "ghcr.io/inspektor-gadget/gadget/trace_exec:latest" -o jsonpretty
INFO[0000] Experimental features enabled
{
  "args": "/opt/cni/bin/bridge",
  "args_count": 1,
  "args_size": 20,
  "comm": "bridge",
  "gid": 0,
  "k8s": {
    "container": "",
    "hostnetwork": false,
    "namespace": "",
    "node": "",
    "pod": ""
  },
  "loginuid": 4294967295,
  "mntns_id": 4026532263,
  "pid": 1475846,
  "ppid": 8590,
  "retval": 0,
  "runtime": {
    "containerId": "e127eea07a1ecd3d8a75515486c31fa6f685fd0ad7d98a598ac3ea70db0d4ed4",
    "containerImageDigest": "",
    "containerImageName": "gcr.io/k8s-minikube/kicbase:v0.0.40",
    "containerName": "minikube",
    "runtimeName": "docker"
  },
  "sessionid": 4294967295,
  "timestamp": "2024-03-27T23:41:35.264215823+01:00",
  "uid": 0,
  "upper_layer": false
}
...
```

PR #2709, opened 2024-04-11
snapshotters: Support sending and JSON-printing output as an array

blanquicet:
# Support sending and JSON-printing snapshotters' output as an array

This PR introduce the following changes:

- Add a new message to the gadget-service API to be able to send more than one data element at a time: `GadgetDataArray`. Notice that:
  - A single data element is represented in the code by the `Data` interface and it's implemented by `datasource.dataElement` (`api.DataElement`) for `api.GadgetDataArray` and `datasource.data` for `api.GadgetData`.
  - An array of data elements is represented in the code by the `DataArray` interface and by `[]*DataElement` in the API.
  - Along the sequence number and node name, a `GadgetDataArray` is composed by multiple data elements (`[]*DataElement`), while a `GadgetData` is composed by one single data element (`DataElement`).
  - Accessors works on a single data element, meaning that `fieldAccessor.Get` or `fieldAccessor.Set` receive the `Data` interface.
- datasources can be now of two types: `TypeSingle` (previously called `TypeEvent`) and `TypeArray` (introduced by this PR). Based on that type, the data source will emit one of the following `Packet` interfaces:
  - `PacketSingle`: The only type we had supported so far, which is used by tracers (and was temporary used by snapshotters until this PR). When serialized to be send through gRPC, it is sent as a `GadgetData` message.
  - `PacketArray`: The new added type that is now used by snapshotters. When serialized to be send through gRPC, it is sent as a `GadgetDataArray` message.
- Add support in datasource for emitting `PacketSingle` and `PacketArray`. In addition, datasource's subscribers has three subscription options:
   - `Subscribe`: Previously supported option which call the callback for each data element in the packet. It means that for array datasources, it will receive one element of the array at a time. This is the option most enrichers will use as they don't care about the way information is sent through gRPC: `PacketSingle` or `PacketArray`. They just need to enrich each data element: `Data`.
   - `SubscribeArray`: Option added by this PR for those subscribers that need to distinguish between single and array datasources. For instance, the clioperator needs to receive the datasources' output as array to be able to print marshall it into a JSON array.
   - `SubscribePacket`: Option added by this PR for those subscribers that only need to serialize what they receive to then send it through gRPC. For instance, the [svc](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadget-service/service-oci.go#L85) operator (created by the gadgetservice).
- For datasources of type `TypeArray`:
  - Add support in datasource's formatters to print the output a JSON array. So, the json package has now `MarshallArray` in addition to `Marshall` (that prints each data elements independently).
  - cli-operator will now use `json.MarshallArray` or `json.Marshall` based on the datasource type: `TypeSingle` or `TypeArray`.
- Register snapshotters' datasources as `TypeArray` and emit output as `PacketArray`. Doing so, its output can be printed as JSON array.

## How to use and testing

```
# Workload
$ docker run --rm --name shell01 nginx:1.21

# JSON output
$ IG_EXPERIMENTAL=true go run ./cmd/ig/... run snapshot_socket -o json
[{"ino":427112,"k8s":{"container":"","hostnetwork":false,"namespace":"","node":"","pod":""},"netns":4026532337,"runtime":{"containerId":"42ec02c853f6cedc841a89c32b5f90cd298229e8e31ba34c0e29fa7a9df39516","containerImageDigest":"","containerImageName":"nginx:1.21","containerName":"shell01","runtimeName":"docker"},"state":10},{"ino":427111,"k8s":{"container":"","hostnetwork":false,"namespace":"","node":"","pod":""},"netns":4026532337,"runtime":{"containerId":"42ec02c853f6cedc841a89c32b5f90cd298229e8e31ba34c0e29fa7a9df39516","containerImageDigest":"","containerImageName":"nginx:1.21","containerName":"shell01","runtimeName":"docker"},"state":10}]

# JSONPretty output
$ IG_EXPERIMENTAL=true go run ./cmd/ig/... run snapshot_socket -o jsonpretty
[
  {
    "ino": 427112,
    "k8s": {
      "container": "",
      "hostnetwork": false,
      "namespace": "",
      "node": "",
      "pod": ""
    },
    "netns": 4026532337,
    "runtime": {
      "containerId": "42ec02c853f6cedc841a89c32b5f90cd298229e8e31ba34c0e29fa7a9df39516",
      "containerImageDigest": "",
      "containerImageName": "nginx:1.21",
      "containerName": "shell01",
      "runtimeName": "docker"
    },
    "state": 10
  },
  {
    "ino": 427111,
    "k8s": {
      "container": "",
      "hostnetwork": false,
      "namespace": "",
      "node": "",
      "pod": ""
    },
    "netns": 4026532337,
    "runtime": {
      "containerId": "42ec02c853f6cedc841a89c32b5f90cd298229e8e31ba34c0e29fa7a9df39516",
      "containerImageDigest": "",
      "containerImageName": "nginx:1.21",
      "containerName": "shell01",
      "runtimeName": "docker"
    },
    "state": 10
  }
]
```

### TODO (Before merging)

- [x] Merge https://github.com/inspektor-gadget/inspektor-gadget/pull/2708
- [x] Merge https://github.com/inspektor-gadget/inspektor-gadget/pull/2637

### TODO (After merging)

- [ ] Re-enable [TestRunSnapshotProcess](https://github.com/inspektor-gadget/inspektor-gadget/blob/e375eb658ecddedaa03471be0f34d7d66be54abb/integration/k8s/run_snapshot_process_test.go#L27): We can't do it yet because we still need to merge output from all nodes into a single array.


blanquicet:
The enrichment is not working yet but the general logic can be reviewed now.

pkg/datasource/data.go
```diff
@@ -133,16 +201,33 @@ func (ds *dataSource) Type() Type {
...
 	d := &data{
-		Payload: make([][]byte, ds.payloadCount),
+		GadgetData: &api.GadgetData{
+			Data: &api.DataElement{
+				Payload: make([][]byte, ds.payloadCount),
+			},
+		},
 	}
+
+	d.dataElement = (*dataElement)(d.GadgetData.Data)
```
flyth, line 213: Can be moved to the definition above.

pkg/datasource/data.go
```diff
@@ -318,26 +403,61 @@ func (ds *dataSource) Subscribe(fn DataFunc, priority int) {
...
+				return err
+			}
+		case TypeArray:
+			iter := p.(PacketArray).Iterate()
+			for iter.Next() {
+				err := sub.fn(ds, iter.Data)
+				if err != nil {
+					return err
+				}
+			}
```
flyth, line 454: I'd prefer not to use `Iterate()` here to avoid the allocation in that function - we can iterate directly over the array, right? (I think in general that direct iteration using a `.Count()` and then `.GetElement(idx)` could be more efficient (avoiding the allocation as well), but I don't have concrete numbers on that.

pkg/datasource/data.go
```diff
@@ -26,15 +26,82 @@ import (
...
+func (d *dataArray) Raw() proto.Message {
+	return (*api.GadgetDataArray)(d.GadgetDataArray)
+}
+
+func (d *dataArray) Iterate() *DataIterator {
+	size := len(d.DataArray)
+	elems := make([]Data, 0, size)
+	for _, d := range d.DataArray {
+		elems = append(elems, (*dataElement)(d))
+	}
```
flyth, line 77: I'm not sure about this one - couldn't be just use a reference to `d.DataArray` here? Also, `DataIterator` could be returned as a non-pointer as it's small enough. Maybe it makes sense to do some benchmarks (also comparing it to using `.Count()` and `.GetElement(idx)` on the array directly instead of using the Iterator), but it might also be just premature optimization 😄

flyth (commented):
Did a first pass - already looks really good. Thanks!

I'd like to semantically separate `Packet` from `Data` a bit more (Operators shouldn't have to care about `Packets` if they only want to work on `Data`). I will play around with it a bit more soon.

pkg/datasource/data.go
blanquicet: It's a pointer to what is defined above, so we first need to define it to be able to assign it here, don't we?

flyth: 🙈 You're of course right, sorry.

pkg/datasource/data.go
blanquicet: Here in particular, I agree as it's within the datasource package, so the underlay struct (dataArray) is known.

pkg/datasource/data.go
blanquicet: > I'm not sure about this one - couldn't be just use a reference to `d.DataArray` here? 
>
I don't think so because `DataArrya` is a `[]*api.DataElement`. It means that elements don't implement `Data`. Considering that the iterator will only be used by frontend, we could make this allocation in `SubscribeArray` so that subscribers receive directly a `[]Data`. WDYT?

> Maybe it makes sense to do some benchmarks (also comparing it to using `.Count()` and `.GetElement(idx)` on the array directly instead of using the Iterator), but it might also be just premature optimization 😄
>
IMO, too premature.

flyth:
> > I'd like to semantically separate `Packet` from `Data` a bit more (Operators shouldn't have to care about `Packets` if they only want to work on `Data`). I will play around with it a bit more soon.
> 
> Makes sense. What about this?
> 
> ```go
> type DataFunc func(DataSource, Data) error
> type ArrayFunc func(DataSource, []Data) error
> type PacketFunc func(DataSource, Packet) error
> 
> Subscribe(dataFn DataFunc, priority int) # Should this be SubscribeData?
> SubscribeArray(arrayFn ArrayFunc, priority int)
> SubscribePacket(packetFn PacketFunc, priority int)
> ```

That looks great!
I think `Subscribe()` is still fine, as that's the default thing that all operators will do and it being on the `DataSource` implies that it's data already. But I'd also like to hear @mauriciovasquezbernal opinion on that.

pkg/datasource/data.go
blanquicet: BTW, I applied the same for `Dump`.

pkg/datasource/data.go
blanquicet: Let me run some experiments. I think I have an idea.

blanquicet: I found a way to avoid creating a new array but work directly with the underlay `DataArray`.

blanquicet:
I found a way to semantically separate Data and Packet. The only change in the API I proposed above is that I defined this:

```go
type ArrayFunc func(DataSource, DataArray) error
```

Instead of:
```go
type ArrayFunc func(DataSource, []Data) error
```

Doing so, operators can get an iterator from `DataArray` or do other operations like clone or whatever an operator needs to do to the original array of the packet. For now, it only does:

```go
type DataArray interface {
	// Len returns the number of elements in the array
	Len() uint32

	// Iterate returns an iterator over the data elements present in the array
	Iterate() *DataIterator
}
```

```

**Comment by blanquicet** [comment]
> The enrichment is not working yet but the general logic can be already reviewed.
>
Solved.

PR is now ready and CI is green.

File: pkg/operators/cli/clioperator.go [hunk 4]
```diff
@@ -246,30 +246,38 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+				ds.Subscribe(func(ds datasource.DataSource, data datasource.Data) error {
+					fmt.Println(string(jsonFormatter.Marshal(data)))
+					return nil
+				}, Priority)
+			case datasource.TypeArray:
+				ds.SubscribeArray(func(ds datasource.DataSource, dataArray datasource.DataArray) error {
+					fmt.Println(string(jsonFormatter.MarshalArray(dataArray)))
+					return nil
+				}, Priority)
 			}
```
flyth, line 280: This needs to be called only for `o.mode != ModeYAML` - right now it's for yaml returning both yaml and JSON.

flyth (approved):
This looks great, thanks! I think we should move forward with this as there are other PRs already touching similar areas in the DataSources.
Tested also by building and testing the sorter operator #2844 on top of this.

LGTM after fixing the yaml/JSON issue.

blanquicet:
Rebased on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/2823.

pkg/datasource/formatters/json/options.go
```diff
@@ -44,3 +44,9 @@ func WithPretty(val bool, indent string) Option {
 		formatter.indent = indent
 	}
 }
+
+func WithArray(val bool) Option {
+	return func(formatter *Formatter) {
+		formatter.array = val
+	}
+}
```
mauriciovasquezbernal, line 52: Is the `val` parameter really needed? I suppose `WithArray()` is clear enough to indicate the array is wanted.

pkg/datasource/datasource.go
```diff
@@ -65,35 +100,51 @@ type DataSource interface {
...
+	NewPacketSingle() PacketSingle
+	// NewPacketSingleFromRaw builds a new PacketSingle from a raw bytes slice coming from protobuf
+	NewPacketSingleFromRaw(b []byte) (PacketSingle, error)
 
-	// NewDataFromRaw builds a new data structure from a raw bytes slice coming from protobuf
-	NewDataFromRaw(b []byte) (Data, error)
+	// NewPacketArray and NewPacketArrayFromRaw are analogous to NewPacketSingle and NewPacketSingleFromRaw, but for
+	// PacketArray
+	NewPacketArray() PacketArray
+	NewPacketArrayFromRaw(b []byte) (PacketArray, error)
```
mauriciovasquezbernal, line 111: It feels a bit weird to have two set of functions to allocate Data, also that you can call NewPacketSingle in an array data source and viceversa. I was wondering if it was possible to split this interface by having DataSourceSingle and DataSourceArray, but it quickly became a mess. I think this solution is fine, but I'd like to have some more checks around to avoid the user calling the wrong function in a non-valid type, for instance, to call SubscribeArray() on a TypeSingle datasource, that should fail.

mauriciovasquezbernal (commented):
Besides my comments to add some more sanity checks, it looks good!

pkg/datasource/formatters/json/options.go
flyth: It allows doing [this](https://github.com/inspektor-gadget/inspektor-gadget/pull/2709/files#diff-8927e1963c71fd5f4eeb8fd5021767e8b96519013ba4357314d0807fd2532a04R249) instead of adding another if statement. See also [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/2652/files#r1550606356), haha 😄

pkg/datasource/data.go
```diff
@@ -150,13 +171,13 @@ func (ds *dataSource) NewData() Data {
...
+			d.Data.Payload[f.PayloadIndex] = make([]byte, 1)
 		case api.Kind_Int16, api.Kind_Uint16:
-			d.Payload[f.PayloadIndex] = make([]byte, 2)
+			d.Data.Payload[f.PayloadIndex] = make([]byte, 2)
 		case api.Kind_Int32, api.Kind_Uint32, api.Kind_Float32:
-			d.Payload[f.PayloadIndex] = make([]byte, 4)
+			d.Data.Payload[f.PayloadIndex] = make([]byte, 4)
 		case api.Kind_Int64, api.Kind_Uint64, api.Kind_Float64:
-			d.Payload[f.PayloadIndex] = make([]byte, 8)
+			d.Data.Payload[f.PayloadIndex] = make([]byte, 8)
```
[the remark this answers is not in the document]
blanquicet: No. Applied.

pkg/datasource/data.go
```diff
@@ -51,10 +51,52 @@ func (d *data) SetSeq(seq uint32) {
...
+
+func (d *dataArray) New() Data {
+	e := &dataElement{
+		Payload: make([][]byte, d.payloadCount),
+	}
+	for i := range e.Payload {
+		e.Payload[i] = make([]byte, 0)
+	}
+	return e
+}
```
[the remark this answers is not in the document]
blanquicet: Right. Applied.

pkg/datasource/data.go
```diff
@@ -372,27 +440,83 @@ func (ds *dataSource) Subscribe(fn DataFunc, priority int) {
...
+				}
+			}
+		}
+	}
+	for _, sub := range ds.subscriptionsArray {
+		err := sub.fn(ds, p.(*dataArray))
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
blanquicet: I modified the logic in `EmitAndRelease` to make sure this can't happen.

pkg/datasource/data.go
```diff
@@ -351,26 +440,83 @@ func (ds *dataSource) Subscribe(fn DataFunc, priority int) {
...
+		return ds.subscriptionsPacket[i].priority < ds.subscriptionsPacket[j].priority
+	})
+}
+
+func (ds *dataSource) EmitAndRelease(p Packet) error {
+	defer ds.Release(p)
+	for _, sub := range ds.subscriptionsData {
+		switch ds.dType {
+		case TypeSingle:
+			err := sub.fn(ds, p.(Data))
```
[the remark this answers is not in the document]
blanquicet: Yep, it works the same. I changed it.

pkg/datasource/formatters/json/options.go
blanquicet: I'm not a big fan of this either, but `WithPretty` behaves already like this, so it did the same for consistency.

mauriciovasquezbernal: Alright, let's keep using the same then, but seems weird to me.

pkg/datasource/datasource.go
blanquicet: I added the error and some test to verify the proper behaviour for creating new packets and subscribing to data, dataArray and packet.

PR #2712, opened 2024-04-11
fix gRPC timeout for image-based gadgets on GetGadgetInfo()

flyth:
The timeout could potentially happen if the server had to download a gadget image on a slow connection. This changes the timeout to be connection-related only and not a total time that the gRPC call may take (we don't set such a limit now.)

Discussed on [Slack](https://kubernetes.slack.com/archives/CSYL75LF6/p1712761766862949)

### DIFF
--- a/pkg/runtime/grpc/grpc-runtime.go
+++ b/pkg/runtime/grpc/grpc-runtime.go
@@ -375,6 +375,10 @@ func (r *Runtime) dialContext(dialCtx context.Context, target target, timeout ti
 			gadgetNamespace := r.globalParams.Get(ParamGadgetNamespace).AsString()
 			return NewK8SPortFwdConn(ctx, r.restConfig, gadgetNamespace, target, port, timeout)
 		}))
+	} else {
+		newCtx, cancel := context.WithTimeout(dialCtx, timeout)
+		defer cancel()
+		dialCtx = newCtx
 	}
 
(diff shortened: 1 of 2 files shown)


burak-ok (approved):
LGTM, thanks

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -375,6 +375,10 @@ func (r *Runtime) dialContext(dialCtx context.Context, target target, timeout ti
 			gadgetNamespace := r.globalParams.Get(ParamGadgetNamespace).AsString()
 			return NewK8SPortFwdConn(ctx, r.restConfig, gadgetNamespace, target, port, timeout)
 		}))
+	} else {
```
mauriciovasquezbernal, line 378: Why it doesn't apply to all cases?

flyth: Good question - I'd have to check again to give a proper answer - I know there were differences in how contexts were handled/used (e.g.: only for connection setup or for the actual outgoing connections), since the grpc connections have many underlying options/paths (like failovers, load balancing and such.)

mauriciovasquezbernal: Good thanks. Feel free to merge it if you're happy with it.

flyth: Yeah, I'll merge for now as this improves the situation for `ig` already.

flyth:
Thanks for the reviews!

Issue #2653, opened 2024-03-27
Open comments / change requests / discussions regarding the refactoring done in #2512


flyth:
This is a rough list of unaddressed / still open comments or discussions from #2512. I'm not sure if all of them are still valid, but I'll try to work my way through them 😄 

I'll add more details later on and maybe sort by priority - however, feel free to edit!

* [ ] [constants for egress/ingress](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1495338191)
* [ ] reevaluate yaml handling (viper)
* [ ] handle requested fields [#1](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1498976756), [#2](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1540383301)
* [ ] [rename methods by removing "OCI"](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1507961453)
* [ ] remove or use [EventTypeGadgetID](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1520743703)
* [ ] [uprobe support](https://github.com/inspektor-gadget/inspektor-gadget/commit/0c4652342fda6dcee07fb83746538781783c2bf6)
* [ ] dynamically sized events from eBPF [#1](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1525897363)
* [ ] [nested fields in metadata](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1523962099)
* [ ] unit tests, especially for DataSources [#1](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1529251319)
* [ ] [prevent duplicate map names](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1526499828)
* [ ] [return error on start/stop fail](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1526528464) (also check stop message)
* [ ] [truncated event handling](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1526581419) [#1](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1539338861), [#2](https://github.com/inspektor-gadget/inspektor-gadget/pull/2512#discussion_r1539342917)
[...]

PR #2713, opened 2024-04-11
fix: re-add timeout for image-based gadgets

flyth:
This adds the parameter `-t` / `--timeout` back to image-based gadgets.

Task from #2653

### DIFF
--- a/cmd/common/oci.go
+++ b/cmd/common/oci.go
@@ -17,6 +17,7 @@ package common
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


burak-ok (approved):
Tested it and LGTM

mauriciovasquezbernal (approved):
LGTM as well.

I see this flag is also shown for the snapshotters:

```bash 
$ sudo -E ig run snapshot_process -h | grep timeout
  -t, --timeout int                     Number of seconds that the gadget will run for, 0 to run indefinitely
```

We can fix it later on, I'll approve and merge to use it on #2607

PR #2738, opened 2024-04-18
image/build: Implement option to keep built file

mauriciovasquezbernal:
There are situations when keeping the built files is useful. Implement a -o/--output option for it

Taken from #2598

### Testing
```bash 
$ cd gadgets/trace_open
$ mkdir /tmp/build
$ sudo -E ig image build . -o /tmp/build/
Successfully built @sha256:8d3859d56a61db46c286c30e81417025b67c4668c9118b0c6f7c5f466e666f31
$ ls -la /tmp/build/
total 184
drwxrwxr-x  2 mauriciov mauriciov  4096 abr 18 12:51 .
drwxrwxrwt 27 root      root      12288 abr 18 12:51 ..
-rw-r--r--  1 mauriciov mauriciov 88784 abr 18 12:51 amd64.bpf.o
-rw-r--r--  1 mauriciov mauriciov 81704 abr 18 12:51 arm64.bpf.o
```

### DIFF
--- a/cmd/common/image/build.go
+++ b/cmd/common/image/build.go
@@ -66,6 +66,7 @@ type cmdOpts struct {
 	fileChanged      bool
 	image            string
 	local            bool
+	outputDir        string
 	builderImage     string
 	updateMetadata   bool
 	validateMetadata bool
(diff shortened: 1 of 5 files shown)


mauriciovasquezbernal:
> But for the first one, I am not sure we really need everything about WASM. Particularly as, for now, we do really support WASM.

This PR is not changing anything related to WASM, it's just moving the code around to make it easier to add more layers in the future. The wasm-related stuff was introduced in 4d61dac898bade77c1d773d48c8d0be4be1bb17a, so I think this discussion doesn't belong to this PR.

PR #2739, opened 2024-04-18
run: validate metadata

mauriciovasquezbernal:
Validate metadata before running the gadget

--- 

#2512 changed the way how the matadata was validated (all validation was almost removed). I think it's still a good idea to validate the whole metadata before processing the gadget and in the same place. This will allow to use the same logic to validate the metadata also when building the gadget. 

---

A deeper discussion will be the role of the metadata now that we agreed to make the macros to define tracers and other mandatory in #2560. Perhaps things like [MapName](https://github.com/inspektor-gadget/inspektor-gadget/blob/37ea34f58ad11c02f2d21281de490234b48c0bc3/pkg/metadata/v1/metadata.go#L23) and [StructName](https://github.com/inspektor-gadget/inspektor-gadget/blob/37ea34f58ad11c02f2d21281de490234b48c0bc3/pkg/metadata/v1/metadata.go#L25) aren't needed anymore now that those are contained in the eBPF program itself?. 

Any thoughts on this?


blanquicet:
> Perhaps things like [MapName](https://github.com/inspektor-gadget/inspektor-gadget/blob/37ea34f58ad11c02f2d21281de490234b48c0bc3/pkg/metadata/v1/metadata.go#L23) and [StructName](https://github.com/inspektor-gadget/inspektor-gadget/blob/37ea34f58ad11c02f2d21281de490234b48c0bc3/pkg/metadata/v1/metadata.go#L25) aren't needed anymore now that those are contained in the eBPF program itself?.
>
Yeah, the metadata file would have only the struct and eBPF params sections (BTW, we can now drop the params defaultValue field). We don't need to associate the structs to the data sources anymore, right?

In other words, with the macros we define the behaviour of the data sources and params. With the metadata, we configure only the "style" and documentation things.

PR #2740, opened 2024-04-18
run: Remove dead code

mauriciovasquezbernal:
Quick removal of dead code we don't need anymore.


burak-ok (approved):
Diffs with only lines deleted are my favorite ones
LGTM, thanks

PR #2745, opened 2024-04-22
gadgets: Add trace_ssl

i-Pear:
# gadgets: Add trace_ssl

Usage: capture SSL data/latency from OpenSSL, GnuTLS, NSS and libcrypto.

### DIFF
--- a/gadgets/trace_ssl/gadget.yaml
+++ b/gadgets/trace_ssl/gadget.yaml
@@ -1,0 +1,68 @@
+name: trace ssl
+description: Captures data on read/recv or write/send functions of OpenSSL, GnuTLS,
+  NSS and Libcrypto
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://inspektor-gadget.io/docs
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+tracers:
+  ssl:
+    mapName: events
+    structName: event
+structs:
+  event:
+    fields:
+    - name: mntns_id
+      description: mount namespace inode id
+      attributes:
+        template: ns
+    - name: operation
+      description: type of SSL operations
+      attributes:
+        width: 16
+        alignment: left
+        ellipsis: end
+    - name: timestamp
+      attributes:
+        template: timestamp
+    - name: latency_ns
+      attributes:
+        template: timestamp
+    - name: pid
+      attributes:
+        template: pid
+    - name: tid
+      attributes:
+        template: pid
+    - name: uid
+      attributes:
+        template: uid
+    - name: gid
+      attributes:
+        template: uid
+    - name: len
+      description: data length
+      attributes:
+        width: 10
+        alignment: left
+        ellipsis: end
+    - name: comm
+      description: command
+      attributes:
+        template: comm
+    - name: buf
+      description: unencrypted buffer
+      attributes:
+        width: 32
+        alignment: left
+        ellipsis: end
+    - name: retval
+      description: return value
+      attributes:
+        width: 20
+        alignment: left
+        ellipsis: end
+ebpfParams:
+  record_data:
+    key: record-data
+    defaultValue: "true"
+    description: controls whether the gadget will send data to userspace
--- a/gadgets/trace_ssl/program.bpf.c
+++ b/gadgets/trace_ssl/program.bpf.c
@@ -1,0 +1,412 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2024 The Inspektor Gadget authors */
+//
+// Prior art: sslsniff from BCC by Adrian Lopez
+// https://github.com/iovisor/bcc/blob/master/tools/sslsniff.py
+//
+// Prior art: sslsnoop from bpftrace by Tao Xu
+// https://github.com/bpftrace/bpftrace/blob/master/tools/sslsnoop.bt
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#include <gadget/buffer.h>
+#include <gadget/macros.h>
+#include <gadget/mntns_filter.h>
+
+#define MAX_BUF_SIZE 8192
+#define MAX_ENTRIES 8192
+
+enum operation {
+	/* libssl */
+	libssl_SSL_do_handshake,
+	libssl_SSL_read,
+	libssl_SSL_write,
+
+	/* libgnutls */
+	libgnutls_gnutls_record_send,
+	libgnutls_gnutls_record_recv,
+
+	/* libnss */
+	libnss_PR_Write,
+	libnss_PR_Send,
+	libnss_PR_Read,
+	libnss_PR_Recv,
+
+	/* libcrypto */
+	libcrypto_rsa_ossl_public_encrypt,
+	libcrypto_rsa_ossl_public_decrypt,
+	libcrypto_rsa_ossl_private_encrypt,
+	libcrypto_rsa_ossl_private_decrypt,
+	libcrypto_RSA_sign,
+	libcrypto_RSA_verify,
+	libcrypto_ossl_ecdsa_sign,
+	libcrypto_ossl_ecdsa_verify,
+	libcrypto_ossl_ecdh_compute_key,
+};
+
+struct event {
+	gadget_mntns_id mntns_id;
+	enum operation operation;
+	gadget_timestamp timestamp;
+	u64 latency_ns;
+	u32 pid;
+	u32 tid;
+	u32 uid;
+	u32 gid;
+	u32 len;
+	u64 retval;
+	char comm[TASK_COMM_LEN];
+	u8 buf[MAX_BUF_SIZE];
+};
+
+#define BASE_EVENT_SIZE ((size_t)(&((struct event *)0)->buf))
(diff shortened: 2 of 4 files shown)


i-Pear:
This pull request is updated according to the comments from the `sslsnoop` PR.

alban:
trace_sslsniff and trace_sslsnoop (#2746) are similar. To avoid user confusion, I think there should be only one. There could be a gadget parameter to enable or disable `buf`.

i-Pear:
This PR has been merged with `sslsnoop`. To avoid sending data to userspace, use `ig run trace_ssl:latest --record_data=false`.

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,406 @@
...
+};
+
+struct event {
+	gadget_mntns_id mntns_id;
+	enum operation operation;
+	gadget_timestamp timestamp;
+	u64 latency_ns;
+	u32 pid;
+	u32 tid;
+	u32 uid;
```
alban, line 55: Could add gid

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,406 @@
...
+		return 0;
+
+	event->mntns_id = mntns_id;
+	event->operation = libssl_SSL_DO_HANDSHAKE;
+	event->timestamp = ts;
+	event->latency_ns = ts - *tsp;
+	event->pid = pid_tgid >> 32;
+	event->tid = tid;
+	event->uid = bpf_get_current_uid_gid();
+	event->len = ret;
```
alban, line 272: This seems wrong, len should be left to 0 in the context of SSL_do_handshake.

Because of L258, ret is 1, see https://www.openssl.org/docs/manmaster/man3/SSL_do_handshake.html

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,406 @@
...
+		return 0;
+
+	event->mntns_id = mntns_id;
+	event->operation = op;
+	event->timestamp = ts;
+	event->latency_ns = ts - *tsp;
+	event->pid = pid_tgid >> 32;
+	event->tid = tid;
+	event->uid = bpf_get_current_uid_gid();
+	event->len = ret;
```
alban, line 356: `ret` seems to be an uninitialized variable on the stack.

alban (changes requested):
Thanks!

gadgets/trace_ssl/gadget.yaml
```diff
@@ -0,0 +1,68 @@
...
+      attributes:
+        width: 10
+        alignment: left
+        ellipsis: end
+    - name: comm
+      description: command
+      attributes:
+        template: comm
+    - name: buf
+      description: binary buffer
```
alban, line 53:
```suggestion
      description: unencrypted buffer
```

alban (commented):
Thanks!

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,413 @@
...
+	struct event *event;
+	u64 pid_tgid;
+	u64 uid_gid;
+	u32 pid;
+	u32 tid;
+	u64 ts;
+
+	mntns_id = gadget_get_mntns_id();
+	if (gadget_should_discard_mntns_id(mntns_id))
+		return 0;
```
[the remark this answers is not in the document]
i-Pear: I can understand your concern about security. I'll remove all `gadget_should_discard_mntns_id` in exit probes, and replace `return 0;` by `goto out;`.

Some other ideas:

- Is this used for writing the correct mntns in events?
> And event->mntns_id could be filled with the mntns_id from the "enter" probe kept in the context map keyed by tid.

- If a process changed its mntns, then all its events will not be traced? Maybe not strict enough in some security-related situations. If we attach to some kprobes which can change processes' mntns, can we still trace those processes with new mntns?

- What will happen if a process calls `clone` and changed its tid? For most use cases it doesn't matter, but developers may also think about it.

- Maybe I can manage to skip the ret-probes? For example, I can write (or patch) some assembly and bypass the "return" symbol. More easily, I think c++'s exception can unwind the stack directly, and bypass the ret-probe. If I do this, I can easily make the ebpf map full.

- If I patch libssl (it's indeed possible, maybe just `mprotect`+rw?), will the inode be the same? And will the trace point still be called? In other words, uprobe is "patching" the binary, and if the user is also patching the binary, who will win?

gadgets/trace_ssl/gadget.yaml
```diff
@@ -0,0 +1,68 @@
...
+        template: timestamp
+    - name: pid
+      attributes:
+        template: pid
+    - name: tid
+      attributes:
+        template: pid
+    - name: uid
+      attributes:
+        template: pid
```
alban, line 38:
```suggestion
        template: uid
```

gadgets/trace_ssl/gadget.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - name: gid
+      attributes:
+        template: pid
```
alban, line 41:
```suggestion
        template: uid
```

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,432 @@
...
+SEC("tracepoint/sched/sched_process_exit")
+int trace_sched_process_exit(void *ctx)
+{
+	u32 tid;
+
+	tid = (u32)bpf_get_current_pid_tgid();
+	bpf_map_delete_elem(&mntns_context, &tid);
+	bpf_map_delete_elem(&ssl_starttime, &tid);
+	bpf_map_delete_elem(&ssl_bufs, &tid);
+	bpf_map_delete_elem(&crypto_starttime, &tid);
```
alban, line 117: Optional suggestion: we now have several maps indexed by the tid. I think it would be cleaner to have a single map with a struct type as value.

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,432 @@
...
+	ts = bpf_ktime_get_boot_ns();
+
+	len = PT_REGS_RC(ctx);
+	if (len <= 0) // no data
+		goto clean;
+
+	bufp = bpf_map_lookup_elem(&ssl_bufs, &tid);
+	if (bufp == 0)
+		goto clean;
+	bpf_map_delete_elem(&ssl_bufs, &tid);
```
alban, line 168: Why calling `bpf_map_delete_elem` here? It will be done at the end of the function anyway.

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,432 @@
...
+	pid_tgid = bpf_get_current_pid_tgid();
+	uid_gid = bpf_get_current_uid_gid();
+	tid = (u32)pid_tgid;
+
+	tsp = bpf_map_lookup_elem(&crypto_starttime, &tid);
+	if (tsp == 0)
+		goto clean;
+
+	mntns_ptr = bpf_map_lookup_elem(&mntns_context, &tid);
+	if (mntns_ptr == 0)
```
alban, line 364: Since it is a pointer, we would usually write `NULL` instead of `0`, or just use:
```suggestion
	if (!mntns_ptr)
```

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,432 @@
...
+	u64 ts;
+	u64 *tsp;
+
+	ts = bpf_ktime_get_boot_ns();
+	pid_tgid = bpf_get_current_pid_tgid();
+	uid_gid = bpf_get_current_uid_gid();
+	tid = (u32)pid_tgid;
+
+	tsp = bpf_map_lookup_elem(&crypto_starttime, &tid);
+	if (tsp == 0)
```
alban, line 360: Since it is a pointer, we would usually write `NULL` instead of `0`, or just use:
```suggestion
	if (!tsp)
```

gadgets/trace_ssl/program.bpf.c
```diff
@@ -0,0 +1,432 @@
...
+	event->tid = tid;
+	event->uid = uid_gid;
+	event->gid = uid_gid >> 32;
+	event->len = len;
+	event->retval = PT_REGS_RC(ctx);
+
+	bpf_get_current_comm(&event->comm, sizeof(event->comm));
+
+	if (!record_data ||
+	    bpf_probe_read_user(&event->buf, buf_copy_size, (char *)*bufp))
```
alban, line 199: I get the following:

> Error: starting operator "oci": starting operator "ebpf": creating eBPF collection: program trace_uretprobe_libssl_SSL_write: load program: permission denied: 102: (85) call bpf_probe_read_user#112: R2 unbounded memory access, use 'var &= const' or 'if (var < const)' (133 line(s) omitted)

alban (commented):
Thanks!

gadgets/trace_ssl/program.bpf.c
i-Pear: Got the same err message in 6.6, seems it's because the low version of verifier.

i-Pear: Although `buf_copy_size` can be up to `MAX_BUF_SIZE`, and using `MAX_BUF_SIZE - 1` will waste 1 byte, I haven't found a better way to make the verifier happy. Let's go with this.

alban (approved):
LGTM, thanks!

PR #2746, opened 2024-04-22
[deprecated] gadgets: Add trace_sslsnoop

i-Pear:
# gadgets: Add trace_sslsnoop

Usage: show handshake latency and retval

### DIFF
--- a/gadgets/trace_sslsnoop/artifacthub-pkg.yml
+++ b/gadgets/trace_sslsnoop/artifacthub-pkg.yml
@@ -1,0 +1,28 @@
+# Artifact Hub package metadata file
+version: 0.27.0
+name: "trace sslsnoop"
+category: monitoring-logging
+displayName: "trace sslsnoop"
+createdAt: "2024-04-20T14:09:00+08:00"
+description: "use uprobe to show SSL handshake latency and retval"
+logoURL: "https://inspektor-gadget.io/media/brand-icon.svg"
+license: ""
+homeURL: "https://inspektor-gadget.io/"
+containersImages:
+    - name: gadget
+      image: "ghcr.io/inspektor-gadget/gadget/trace_sslsnoop:latest"
+      platforms:
+        - linux/amd64
+        - linux/arm64
+keywords:
+    - gadget
+links:
+    - name: source
+      url: "https://github.com/inspektor-gadget/inspektor-gadget/"
+install: |
+    # Run
+    ```bash
+    sudo IG_EXPERIMENTAL=true ig run ghcr.io/inspektor-gadget/gadget/trace_sslsnoop:latest
+    ```
+provider:
+    name: Inspektor Gadget
--- a/gadgets/trace_sslsnoop/gadget.yaml
+++ b/gadgets/trace_sslsnoop/gadget.yaml
@@ -1,0 +1,47 @@
+name: trace sslsnoop
+description: use uprobe to show SSL handshake latency and retval
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://inspektor-gadget.io/docs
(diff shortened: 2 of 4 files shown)


gadgets/trace_sslsnoop/artifacthub-pkg.yml
```diff
@@ -0,0 +1,28 @@
+# Artifact Hub package metadata file
+version: 0.27.0
+name: "trace sslsnoop"
+category: monitoring-logging
+displayName: "trace sslsnoop"
+createdAt: "2024-04-20T14:09:00+08:00"
+description: "use uprobe to show handshake latency and retval"
```
alban, line 7:
```suggestion
description: "use uprobe to show SSL handshake latency and retval"
```

gadgets/trace_sslsnoop/gadget.yaml
```diff
@@ -0,0 +1,47 @@
+name: trace sslsnoop
+description: use uprobe to show handshake latency and retval
```
alban, line 2:
```suggestion
description: use uprobe to show SSL handshake latency and retval
```

gadgets/trace_sslsnoop/program.bpf.c
```diff
@@ -0,0 +1,210 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)  TODO: Apache2?
+/* Copyright (c) 2024 The Inspektor Gadget authors */
+//
+// Based on sslsnoop from bpftrace by Tao Xu
```
alban, line 4: Could you add a link to the source?

gadgets/trace_sslsnoop/program.bpf.c
```diff
@@ -0,0 +1,210 @@
...
+		case SSL_write:
+		case SSL_do_handshake: {
+			u64 *start_time_ptr = bpf_map_lookup_elem(&start_ssl, &tid);
+			if (!start_time_ptr)
+			{
+				gadget_discard_buf(event);
+				return 0;
+			}
+			start_time = *start_time_ptr;
+			bpf_map_delete_elem(&start_ssl, &tid);
```
alban, line 130: This could be done before `gadget_reserve_buf`, so if `gadget_reserve_buf` fails, we would not leak the entry in `start_ssl`. Also, that would remove the need for `gadget_discard_buf`, making the code simpler.

Ditto for `start_crypto` below.

If the process is killed between the "record_start" and the "submit_event", the maps would leak entries too. This could be prevented by checking in the `sched_process_exit` tracepoint for threads to clean up.

i-Pear:
TODO: update Makefile

gadgets/trace_sslsnoop/program.bpf.c
i-Pear: > If the process is killed between the "record_start" and the "submit_event", the maps would leak entries too. This could be prevented by checking in the sched_process_exit tracepoint for threads to clean up.

Here we are using the `thread ID` instead of the `process ID`, because each thread may call a SSL function separately. Since the `sched_process_exit` is for process exiting, we can not use that here.

i-Pear: You are right, it will be called when a thread exits.

alban: Credits to Falco for the sched_process_exit idea:
https://github.com/falcosecurity/libs/blob/master/driver/modern_bpf/programs/attached/events/sched_process_exit.bpf.c#L151

i-Pear: This has been applied to all open pull requests.

i-Pear:
This PR will be closed once `sslsnoop` be merged into `sslsniff`.

PR #2747, opened 2024-04-22
pkg/uprobetracer: support LSM trace points

i-Pear:
# pkg/uprobetracer: support LSM trace points

## How to use

Use `lsm/<hook>` as section names.

### DIFF
--- a/docs/reference/program-types.md
+++ b/docs/reference/program-types.md
@@ -65,3 +65,7 @@ For common libraries, `<file_path>` can also be the library's name, such as `lib
 The section name must use the `usdt/<file_path>:<providerName>:<probeName>` format.
 `<file_path>` can be either an absolute path or a library name, same as the field in Uprobe.
 `<providerName>` and `<probeName>` are two fields that can jointly identify a USDT trace point.
+
+### Tracing with Linux Security Modules (LSM)
+The section name must use the `lsm/<hook>` format.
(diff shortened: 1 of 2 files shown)


alban:
Is there a way to use a gadget enforcing different LSM policies depending on the container?

E.g. `lsm/mmap_file` on `"/usr/lib/x86_64-linux-gnu/liblzma.so.5.6.1"` would be allowed in a Kubernetes namespace but not in others?

Like kprobes, lsm programs are global but they could have access to a ebpf hash map containing a list of mount namespaces for containers that match specific criteria (`--namespace`, `--label`...).

alban:
Example of project using eBPF LSM with different policies depending on the label `security.kubernetes.io/enforce` of the Kubernetes namespace:
https://lockc-project.github.io/policies/index.html
https://github.com/lockc-project/lockc/blob/main/lockc-ebpf/src/policy.rs#L7-L12

i-Pear:
> Is there a way to use a gadget enforcing different LSM policies depending on the container?

It could use `gadget_get_mntns_id` and `gadget_should_discard_mntns_id`?

i-Pear:
I remembered that one limitation of LSM eBPF programs is the lack of the 'ctx' parameter. This prevents functions like `gadget_submit_buf` from being used. Currently, I'm bypassing this by using `bpf_ringbuf_output`.

alban (approved):

PR #2751, opened 2024-04-22
trace exec: Add user and group field

### LINKED ISSUE #2706: [RFE] [trace exec] add 'user' and 'group' fields
## Current situation

We would like to have additional context to our trace exec events, specifically the User and Group fields

## Impact

We need this data in order to create better security value for our customers :)

## Additional information

GID and UID are of no use to us

burak-ok:
# trace exec: Add user and group field

This PR adds 2 additional fields to `trace exec`: `user` and `group` which are the **host** user name and group names.

In order not to read `/etc/passwd` and `/etc/group` constantly for every event a new caching struct was introduced (yes, another cache). It uses `fanotify` to get notified for modifications of these 2 files.

**_Remember_**: If you use `uid 1000` inside the container, the event will show `uid 1000` and `user burak` (On my host) instead of the "real" username inside the container. This PR has no support to get the username inside the container and it would be complicated to support every corner case then.

resolves #2706


alban:
Can this be used by image-based gadgets?

I think the limitations have to be documented somewhere in `docs/`:
- Resolve user name and group name on the host only.
- Does not follow [nsswitch.conf](https://man7.org/linux/man-pages/man5/nsswitch.conf.5.html) but hard code `/etc/passwd` and `/etc/group`, meaning that [ldap](https://wiki.debian.org/LDAP/NSS) or [nss-systemd](https://www.freedesktop.org/software/systemd/man/latest/nss-systemd.html) won't be supported. Systemd units with `DynamicUser=yes` will not be resolved.

I think we need to have documentation for each operator. This PR could start with the UidGidResolver operator.

pkg/operators/uidgidresolver/uidgidresolver.go
```diff
@@ -0,0 +1,126 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package uidgidresolver provides an operator that enriches events by looking
+// up uid and gid resolving them to the corresponding username and groupname.
+package uidgidresolver
```
alban, line 17: Document the limitations here too: https://github.com/inspektor-gadget/inspektor-gadget/pull/2751#issuecomment-2071838165

pkg/operators/common/usergroupcache.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+		defer func() {
+			// Only close the file if we are not going to use it
+			if notify != nil {
+				notify.File.Close()
+			}
+		}()
+
+		passwdPath := filepath.Join(host.HostRoot, "/etc/passwd")
+		err = notify.Mark(unix.FAN_MARK_ADD, unix.FAN_MODIFY, -1, passwdPath)
```
alban, line 104: Do you get the `FAN_MODIFY` event correctly after several changes? Does `notify.Mark` monitor the inode or the path? I noticed that `useradd` writes a new passwd file and atomically rename it to the correct path.
```
$ sudo strace -f useradd janetest
...
openat(AT_FDCWD, "/etc/passwd+", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 5
umask(022)                              = 0777
fchown(5, 0, 0)                         = 0
fchmod(5, 0644)                         = 0
newfstatat(5, "", {st_mode=S_IFREG|0644, st_size=0, ...}, AT_EMPTY_PATH) = 0
write(5, "root:x:0:0:root:/root:/bin/bash\n"..., 3705) = 3705
fsync(5)                                = 0
close(5)                                = 0
newfstatat(AT_FDCWD, "/etc/passwd", {st_mode=S_IFREG|0644, st_size=3656, ...}, AT_SYMLINK_NOFOLLOW) = 0
rename("/etc/passwd+", "/etc/passwd")   = 0
...
```

burak-ok: Doesn't seem to. I interpreted the documentation wrong
>  If pathname is absolute, it defines the filesystem object to be marked, and dirfd is ignored.


So I would need to mark `/etc` i guess

pkg/operators/uidgidresolver/uidgidresolver.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/common"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+const (
+	OperatorName = "UidGidResolver"
+)
+
+type UidResolverInterface interface {
+	ResolveUids() map[uint32]*string
```
mauriciovasquezbernal, line 35: Do we envision a case where a gadget will provide multiple uids? Personally I'd try to simplify this interface to something like:

```golang
GetUid() uint32
SetUserName(string)
```

It seems to me the map is an overkill right now as almost all gadgets won't use it. Also, builtin gadgets should be deprecated soon, so let's keep these interfaces as simple as possible.

pkg/operators/common/usergroupcache.go
mauriciovasquezbernal, line 1: Do we really need this under common/? I'd say we can put it on the same folder of the operator and move it later on if needed.

pkg/operators/common/resourcecache.go
```diff
@@ -0,0 +1,137 @@
...
+	for key, oldObj := range c.old {
+		if _, ok := c.current[key]; !ok {
+			objs = append(objs, oldObj.obj)
+		}
+	}
+	return objs
+}
+
+func (c *ResourceCache[T, Key]) Get(key Key) *T {
+	c.PruneOldObjects()
```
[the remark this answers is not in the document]
burak-ok: Could also be ran async on a timer of course. Will do that.

(Shouldn't be too heavy though, since old Resources are quite rare in our use cases maybe)

pkg/operators/common/resourcecache.go
```diff
@@ -0,0 +1,137 @@
...
+
+type oldResource[T any] struct {
+	deletionTimestamp time.Time
+	obj               T
+}
+
+type ResourceCache[T any, Key comparable] struct {
+	sync.Mutex
+	current map[Key]*T
+	old     map[Key]oldResource[*T]
```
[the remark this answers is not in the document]
burak-ok: Just for efficieny reasons I want to keep the list of active resources separate from the "old" resources.

Active resources always have a higher precedence over old resources, the key would need to be different, iterating over all resources in `PruneOldObjects` is way longer

pkg/operators/common/resourcecache.go
```diff
@@ -0,0 +1,137 @@
...
+	toKey   func(*T) (Key, error)
+}
+
+func NewResourceCache[T any, Key comparable](toKey func(*T) (Key, error)) *ResourceCache[T, Key] {
+	return &ResourceCache[T, Key]{
+		current: make(map[Key]*T),
+		old:     make(map[Key]oldResource[*T]),
+		toKey:   toKey,
```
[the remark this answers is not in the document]
burak-ok: Well this cache grew from the `KubeInvCache`. And there I wanted to avoid explicitly getting the key from a Pod or service.
But yes. This is somehow uncommon what I did here

pkg/operators/common/usergroupcache.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (cache *userGroupCache) Start() {
+	cache.useCountMutex.Lock()
+	defer cache.useCountMutex.Unlock()
+
+	// No uses before us, we are the first one
+	if cache.useCount == 0 {
+		fanotifyFlags := uint(unix.FAN_CLOEXEC | unix.FAN_NONBLOCK)
+		openFlags := os.O_RDONLY | unix.O_LARGEFILE | unix.O_CLOEXEC
+		notify, err := fanotify.Initialize(fanotifyFlags, openFlags)
```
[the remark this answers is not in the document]
burak-ok: Only reason was that we already used `fanotify` in our code and this would avoid pulling in more dependencies

pkg/operators/uidgidresolver/uidgidresolver.go
burak-ok: Sure, we can do that.
I had the idea from `trace exec`, since it also provides a `loginuid` field

pkg/operators/common/resourcecache.go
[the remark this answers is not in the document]
burak-ok: Moved to its own package. Any naming suggestions are welcome :)

burak-ok:
Pushed a new version.
It still can't monitor updated `/etc/passwd` files because of https://github.com/inspektor-gadget/inspektor-gadget/pull/2751#discussion_r1577765949

That will be handled tomorrow

burak-ok:
New version relies on `fsnotify` and monitors `/etc` so we will always get all the notifications that we need for `passwd` and `group`:

```bash
$ adduser foobar # another terminal
{...,"uid":1000,"user":"burak","gid":1001,"group":"foobar","upperlayer":false,"loginuid":4294967295,"sessionid":4294967295}
{...,"uid":1000,"user":"burak","gid":1002,"group":"","upperlayer":false,"loginuid":4294967295,"sessionid":4294967295}

$ adduser foobar2 # another terminal
{...,"uid":1000,"user":"burak","gid":1002,"group":"foobar2","upperlayer":false,"loginuid":4294967295,"sessionid":4294967295}
````

File: pkg/cachedmap/cachedmap.go [hunk 8]
```diff
@@ -31,6 +31,7 @@ type CachedMap[Key comparable, T any] interface {
 	Clear()
 	Add(key Key, obj *T)
 	Remove(key Key)
+	Keys() []Key
 	ToSlice() []*T
```
mauriciovasquezbernal, line 35: Name it `Values` then? (It'll be similar to the Keys and Values methods from the exp package https://pkg.go.dev/golang.org/x/exp/maps#Values)

docs/devel/operator/UidGidResolver.md
```diff
@@ -0,0 +1,37 @@
...
+	    ResolveUids() map[uint32]*string
+    }
+    ```
+2. Implement the GidResolverInterface for the `event struct` to resolve GIDs.
+   The return type is a map which maps the requested `gid` to string pointers, where the groupname should be written to.
+    ```go
+    type GidResolverInterface interface {
+	    ResolveGids() map[uint32]*string
+    }
+    ```
```
mauriciovasquezbernal, line 33: The interface was changed, right? I'd suggest not to document this part as we're trying to remove these builtin gadgets soon.

pkg/cachedmap/cachedmap.go
```diff
@@ -0,0 +1,171 @@
...
+	c.RLock()
+	defer c.RUnlock()
+
+	keys := make([]Key, 0, len(c.current)+len(c.old))
+	for key := range c.current {
+		keys = append(keys, key)
+	}
+	for oldKey := range c.old {
+		keys = append(keys, oldKey)
+	}
```
mauriciovasquezbernal, line 123: Is it fine to return duplicated keys?

pkg/operators/uidgidresolver/usergroupcache.go
```diff
@@ -0,0 +1,261 @@
...
+	for id := range oldEntries {
+		resourceCache.Remove(id)
+	}
+}
+
+func (cache *userGroupCache) GetUsername(uid uint32) string {
+	name := cache.userCache.Get(uid)
+	if name != nil {
+		return *name
+	}
```
mauriciovasquezbernal, line 251: This seems weird. What about making the `cachedmap` return `T` instead of `*T` on the `Get()` (and other) methods, then it's up to the user whether to store pointers or values. In this case, we could store `string` directly, removing this logic here.

pkg/operators/uidgidresolver/usergroupcache.go
```diff
@@ -0,0 +1,261 @@
...
+		userGroupCacheSingleton, userGroupErr = newUserGroupCache()
+	})
+	return userGroupCacheSingleton, userGroupErr
+}
+
+func newUserGroupCache() (*userGroupCache, error) {
+	return &userGroupCache{
+		userCache:  cachedmap.NewCachedMap[uint32, string](2 * time.Second),
+		groupCache: cachedmap.NewCachedMap[uint32, string](2 * time.Second),
+	}, nil
```
mauriciovasquezbernal, line 78: Can't we allocate these structs on `Start()`, and then `Close()` will call `{user,group}Cache.Stop()`? Otherwise the cleaning loop of the cached map will be running even when there are not users of this cache.

pkg/operators/uidgidresolver/usergroupcache.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (cache *userGroupCache) Start() {
+	cache.useCountMutex.Lock()
+	defer cache.useCountMutex.Unlock()
+
+	// No uses before us, we are the first one
+	if cache.useCount == 0 {
+		watcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			log.Warnf("UserGroupCache: create watcher: %v", err)
```
mauriciovasquezbernal, line 89: Why doesn't it return an error? I'd rather return an error here and handle it on `PreGadgetRun() ` (by printing a warning if we can to ignore it)

mauriciovasquezbernal (commented):
It's looking much better. I still have few comments.

pkg/cachedmap/cachedmap.go
burak-ok: I don't think so. In my mind I thought i was adding it to a map and therefore it didn't matter

pkg/operators/uidgidresolver/usergroupcache.go
burak-ok: TIL: I thought you can't instantiate a new object of a generic type parameter.

This won't compile. That is why I went with pointers
```
func (c *cachedMap[Key, T]) Get2(key Key) (T, error) {
    ....
    // Didn't find key
    return T{}
}
```

But there seems to be the following way, which compiles:
```
func (c *cachedMap[Key, T]) Get2(key Key) (T, error) {
    ....
    // Didn't find key
    var zeroValue T
    return zeroValue
}
```

burak-ok:
The CodeQL scan fails because I print out the full passwd path. Is this an issue or can we flag that as a false positive?

pkg/cachedmap/cachedmap.go
```diff
@@ -0,0 +1,178 @@
...
+		select {
+		case <-ticker.C:
+			c.PruneOldObjects()
+		case <-c.exit:
+			return
+		}
+	}
+}
+
+func (c *cachedMap[Key, T]) PruneOldObjects() {
```
mauriciovasquezbernal, line 104: nit: do we want to export this? (it's not part of the interface you defined above)

pkg/cachedmap/cachedmap_test.go
```diff
@@ -0,0 +1,210 @@
...
+)
+
+const defaultTestTTL = 2 * time.Second
+
+func TestNewCachedMap(t *testing.T) {
+	t.Parallel()
+
+	cache := NewCachedMap[int, string](defaultTestTTL).(*cachedMap[int, string])
+	require.NotNil(t, cache)
+	defer cache.Close()
```
mauriciovasquezbernal, line 33: nit: AFAIU it's a better practice to use test.Cleanup() instead of defers in tests. (https://stackoverflow.com/questions/61609085/what-is-useful-for-t-cleanup)

pkg/cachedmap/cachedmap_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	assert.Equal(t, 0, len(cache.current))
+	assert.Equal(t, 0, len(cache.old))
```
mauriciovasquezbernal, line 36: nit: https://pkg.go.dev/github.com/stretchr/testify/assert#Len (yes, the assert package has a function for each possible case you can think about)

pkg/cachedmap/cachedmap_test.go
```diff
@@ -0,0 +1,210 @@
...
+	locked := cache.TryLock()
```
mauriciovasquezbernal, line 37: mmm, I'm not sure about this one. Why would we want to Lock the cache? (IMO the lock shouldn't be exposed to users)

pkg/cachedmap/cachedmap_test.go
mauriciovasquezbernal, line 1: IMO this could (and should) be implemented as a black box test. In the end, we want to test the behavior we expose to the user, if we change the internal implementation these tests should continue to work the same.

docs/devel/operator/UidGidResolver.md
mauriciovasquezbernal, line 1: nit: id -> ID and gid -> GID all over the document?

pkg/gadgets/trace/exec/types/types.go
```diff
@@ -32,14 +32,32 @@ type Event struct {
 	Retval     int      `json:"ret,omitempty" column:"ret,width:3,fixed"`
 	Args       []string `json:"args,omitempty" column:"args,width:40"`
 	Uid        uint32   `json:"uid" column:"uid,template:uid,hide"`
+	Username   string   `json:"user" column:"user,hide"`
```
mauriciovasquezbernal, line 35: nit: probably omitempty makes sense for these two new members.

pkg/operators/uidgidresolver/usergroupcache.go
```diff
@@ -0,0 +1,261 @@
...
+
+const (
+	passwdFileName = "passwd"
+	groupFileName  = "group"
+	baseDirPath    = "/etc"
+)
+
+var (
+	userGroupCacheSingleton *userGroupCache
+	userGroupOnce           sync.Once
```
mauriciovasquezbernal, line 61: nit: you can use [OnceValue](https://cs.opensource.google/go/go/+/go1.22.2:src/sync/oncefunc.go;l=43) in go 1.21 to replace these two.

pkg/operators/uidgidresolver/usergroupcache.go
```diff
@@ -0,0 +1,261 @@
...
+		resourceCache.Remove(id)
+	}
+}
+
+func (cache *userGroupCache) GetUsername(uid uint32) string {
+	name, found := cache.userCache.Get(uid)
+	if !found {
+		return ""
+	}
+	return name
```
mauriciovasquezbernal, line 252: does 

```golang
name, _ := cache.userChache.Get(uid)
return name
````

work as well?

mauriciovasquezbernal (approved):
> The CodeQL scan fails because I print out the full passwd path. Is this an issue or can we flag that as a false positive?

I don't see why it can be an issue. The thing printed is the path of the password file but not any of this contents. I'd say it's a false positive. 

I made some more comments on small things, but IMO it's ready to go. Feel free to fix the comments you think are relevant. 


Good work! thanks!

pkg/cachedmap/cachedmap_test.go
burak-ok: On the one hand I understand the motivation behind implementing tests as a black box. Maybe the mutex lock for example was too much.
But on the other hand when I want to specifically want to test `Add` or `Remove` I want to be sure what happened in the implementation and directly touch the private instance maps.

I would say that a black box test is a different category of unit tests. There is probably a name for that or for the kind of test I wrote :D

pkg/cachedmap/cachedmap_test.go
burak-ok: Removed. Was maybe too much internal state testing

pkg/operators/uidgidresolver/usergroupcache.go
burak-ok: Works, thanks for the reminder

Issue #2021, opened 2023-09-11
image-based gadgets: Implement golang API


mauriciovasquezbernal:
There are cases where a user might want to integrate Inspektor Gadget on their application directly. It was done before by directly using the gadgets (https://github.com/inspektor-gadget/inspektor-gadget/tree/main/examples/gadgets/basic/trace/exec), but this approach won't be possible anymore now that we're implementing the gadgets as containerized gadgets.

We should hence provide a golang API to make this possible as stated in the design doc -> https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md#running-gadgets-directly

flyth:
Current high-level API state with custom exporter: https://gist.github.com/flyth/e686d44235d8d7145c6f25aea0bf296a

mauriciovasquezbernal:
This was mostly fixed by #2512. #2753 introduced some examples. 

There are still a couple of things to improve, but we can mark this as done. 

- #2893
- #2891
- #2892

Issue #1924, opened 2023-08-03
image-based gadgets: Implement support for toppers gadgets


mauriciovasquezbernal:
## Issue Tracker
- [x] top blockio #2989
- [ ] top ebpf #3443

See more details on https://github.com/inspektor-gadget/inspektor-gadget/issues/1924#issuecomment-2127209677

mauriciovasquezbernal:
Fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/2132.

blanquicet:
Reopening as this support was removed after https://github.com/inspektor-gadget/inspektor-gadget/pull/2512.

blanquicet:
TODOs:

- [ ] #2637: As mentioned in its description, in this PR I prepare snapshotters to easily collect the outcome of all iterators so that we can emit the whole output of a snapshotter in one single batch. Currently, we are sending one entry at a time.
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/2709: Support sending arrays of data in a single batch and apply it to snapshotters. 
- [ ] Create new operator that combines output from all nodes on client side.
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/2844: Create new operator for sorting
- [ ] Create new `cutter` operator (Or maybe inside the sorter operator) to be able to specify a fixed amount of entries per update on toppers (basically implement `--max-rows` flag on client side).
- [ ] Finally, combine everything to support running toppers using arrays and the cutter, sorter and combiner operators.

mauriciovasquezbernal:
I don't think it was fixed by #2844, reopening.

mauriciovasquezbernal:
I think we can close this. top_file, top_tcp and top_blockio had been converted. We'll be handling top ebpf (or related functionality) on #3759.

PR #2753, opened 2024-04-22
Implement image-based gadgets examples

mauriciovasquezbernal:
This PR provides some examples of running image-based gadgets through the [Golang API](https://github.com/inspektor-gadget/inspektor-gadget/issues/2021#issuecomment-2042702705)

Examples included: (readme added on this PR)

### Gadgets Examples

This folder contains different examples of how gadgets can be run from a Golang
application.

> ⚠️ These examples are work in progress. Be sure to check the release notes to
> understand relevant changes on the API.

TODO: link to API and concepts documentation before

### Simple Gadgets

Examples showing how to use some gadgets in their simplest configuration.

- [trace_open](./simple/trace_open/): Run the `trace_open` gadget and print the
  events to the terminal in json format.
- [trace_dns](./simple/trace_dns/): Run the `trace_dns` gadget with some of the
  operators it requires.

### Operators

Examples showing how to use some of the operators.

- [local_manager](./operators/local_manager/): Use the local manager operator
to filter and enrich events with container data.
- [cli](./operators/cli/): Use the CLI operator to print data to the terminal in
  a pretty way.

### Datasource

Examples showing how to use `datasource`

- [fields](./datasource/fields/): Show how to access specific fields from the datasource.
- [mutate](./datasource/mutate/): Mutate and add fields to a datasource.

### Executing Gadgets on remove Inspektor Gadget instances

Examples showing how to execute gadgets in remove instances of Inspektor Gadget
(either `ig` or `iìg-k8s`) by using the grpc runtime.

- [custom_operator](./grpc/custom_operator/): Run a remote gadget and print its
  output to the terminal in json format by using a custom operator.

--- 

The idea of this PR is to discuss some of the pain points I found while implementing the examples and possible solutions to them. However I shouldn't be blocked on the solutions to all of them. I think the examples are valuable in their current status.

---
### TODO
- [x] Configure CI to automatically compile them
- [ ] Open issues to further discuss some "pain points"

### Future TODOs
- [ ] Link to document with API  and documentation about concepts in examples/gadgets/README.md (https://github.com/inspektor-gadget/inspektor-gadget/pull/2753#discussion_r1594153736)
- [ ] Initialize operators outside the gadget context (https://github.com/inspektor-gadget/inspektor-gadget/pull/2753#discussion_r1594256847)
- [ ] How to handle SocketFilter and SchedCLS programs when localmanager is not available? (https://github.com/inspektor-gadget/inspektor-gadget/pull/2753#discussion_r1594240726)
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/2753#discussion_r1601757947


examples/gadgets/simple_trace_open/main.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := runtime.Init(nil); err != nil {
+		return fmt.Errorf("runtime init: %w", err)
+	}
+
+	// Before running the gadget, let's set some parameters.
+	// TODO: link to documentation with the parameters supported by this gadgert
+	params := map[string]string{
+		// Filter only events from the root user
+		"operator.oci.ebpf.uid": "0",
+	}
```
mauriciovasquezbernal: I found it very painful to get it right. What about the `ParamKey` idea I implemented in one of the commits of this PR?

mauriciovasquezbernal (commented):
Some questions I found while implementing the examples.

examples/gadgets/simple_trace_open/main.go
flyth: I think it's okay to have these - they're similar to how annotations in Kubernetes look like ([example](https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/#annotations)). But we should of course provide a good documentation around those params and the actual `annotations` we support, though.

The `ParamKey` looks okay, but the downsides are
* you add a hard dependency on the operator (which in these examples is okay, but still)
* you would need to implement them for all we offer to not be more confusing than helpful, so we might as well add that to the API itself
* the key actual depends on how the operator is used; we could later on have operators as layer operators as well as data operators (think wasm, for example - we can have them as part of the gadget image as well as globally installed wasm programs) - in the former case it would be `operator.oci.wasm`, in the latter it would only be `operator.wasm`

By the way, I thought maybe dropping the starting `operator.` might make sense now that we don't have `gadget` params anymore (right now it's only operator + runtime.)

examples/gadgets/fields/main.go
```diff
@@ -0,0 +1,85 @@
...
+
+	myOperator := simple.New("myHandler", simple.OnInit(func(gadgetCtx operators.GadgetContext) error {
+		for _, d := range gadgetCtx.GetDataSources() {
+			// Access some specific fields
+			pidF := d.GetField("pid")
+			commF := d.GetField("comm")
+			fnameF := d.GetField("fname")
+
+			d.Subscribe(func(source datasource.DataSource, data datasource.Data) error {
+				pid := pidF.Uint32(data)
```
flyth, line 43: Maybe we should add a generic `.Interface()` or `.Any()` (although I find that more confusing) to cast to whatever the underlying type is and return it as `any` in case you don't care about the type (e.g. when printing with `%v`).

examples/gadgets/simple_trace_open/main.go
mauriciovasquezbernal: I think the issue is not only about the prefix, but in general about the full key. I find it very difficult for a use of the API to understand what are the parameters (and their type) that are available and how this key should be composed. 

I'm wondering if we can offer a more structured way to set parameters, like a Golang structure.

flyth: How about additionally defining all params as either `ParamPARAMNAME` or `GlobalParamPARAMNAME` with a proper description that includes the type (we mostly have done it already anyway). Combined with the above that would lead to

```
operators.AddParam(params, ocihandler.OciHandler, ocihandler.ParamFoobar, "value")
```

(changed order of paramMap here). Usually the IDE would autocomplete after typing `ocihandler.Param...` and help with what is available.

mauriciovasquezbernal: I think we need more discussion on the parameters thing. IMO we can try to implement basic examples so folks can start playing with the API, in the meanwhile we can continue the parameters discussion.

mauriciovasquezbernal:
I updated the PR by reorganizing the examples and adding a few more. My proposal is to start the discussion on these examples, and create a wish-list of other examples that we can add later on. Any thoughts?

examples/gadgets/grpc/custom_operator/main.go
mauriciovasquezbernal, line 1: I think this example is totally screwed up because the gadget is being run on the client. I'll investigate more.

flyth: Removing `ocihandler.OciHandler` from operators used should fix it.

mauriciovasquezbernal: Got it! I wasn't using `operator.LocalManager.host": "true"`, hence I wasn't seeing any events because I didn't have any containers on the machine.

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+				fmt.Printf("%s\n", jsonOutput)
+				return nil
+			}, opPriority)
+		}
+		return nil
+	}))
+
+	// There are some gadgets, like trace_dns, trace_sni, trace_network that
+	// require the local manager (or kube manager) operators to work.
+	// TODO: How to mark them and why is it a hard requirement?
```
blanquicet, line 51: **Why a hard requirement?**

The local/kube manager operator is required for all the socket filter programs because they need to be container-aware to run in each container's netns. All this is done using the AttachContainer/DetachContainer interfaces implemented by the networktracer and called by the local/kube manager operator. Similar reasoning for sched cls programs. They require the local/kube manager operator.

**How to mark?**

Currently, based on the program types (socket filter and sched cls), the ebpf operator request the local/kube manager operator to get activated using the `NeedContainerEvents` context variable (I'll open a PR on this soon) in the `Prepare` phase. Maybe in the `Start` phase, before actually running the gadget, we could check if the local/kube manager operator got actually activated or not. In case it wasn't return an error and request for it: `SocketFilter and SchedCLS programs require an operator that generate container creation/deletion events`, or so.

To document that, maybe the required operators can be automatically added to metadata file?

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Create the local manager operator
+	localManagerOp := &localmanager.LocalManager{}
+	localmanagerParams := localManagerOp.GlobalParamDescs().ToParams()
```
blanquicet, line 55: nit: 

```suggestion
	localManagerParams := localManagerOp.GlobalParamDescs().ToParams()
```

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+	// The socker enricher operator is used to provide information about the
+	// process performing the DNS request.
+	socketEnricherOp := &socketenricher.SocketEnricher{}
+	socketEnricherOp.Init(nil)
+
+	gadgetCtx := gadgetcontext.New(
+		context.Background(),
+		"ghcr.io/inspektor-gadget/gadget/trace_dns:latest",
+		gadgetcontext.WithDataOperators(
+			ocihandler.OciHandler,
```
blanquicet, line 68: Comment about this operator that doesn't need to be initialized but it's a singleton?

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/localmanager"
+	ocihandler "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/oci-handler"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/simple"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/socketenricher"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local"
+)
+
+func do() error {
+	const opPriority = 50000
```
blanquicet, line 35: How do people know the priority they need to set so that their operator is called after the other operators?

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+			socketEnricherOp,
+			myOperator,
+		),
+	)
+
+	runtime := local.New()
+	if err := runtime.Init(nil); err != nil {
+		return fmt.Errorf("runtime init: %w", err)
+	}
+
```
blanquicet, line 79:
```suggestion
defer runtime.Close()

```

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+	// There are some gadgets, like trace_dns, trace_sni, trace_network that
+	// require the local manager (or kube manager) operators to work.
+	// TODO: How to mark them and why is it a hard requirement?
+
+	// Create the local manager operator
+	localManagerOp := &localmanager.LocalManager{}
+	localmanagerParams := localManagerOp.GlobalParamDescs().ToParams()
+	localmanagerParams.Get(localmanager.Runtimes).Set("docker")
+	localManagerOp.Init(localmanagerParams)
```
blanquicet, line 57: Mmmmmm I'm not sure we should manually init operators. IMO, we should pass the global params to gadget context, and it should init the operator for us (if operator is needed). BTW, what you're doing here is currently overwritten here, isn't it?

https://github.com/inspektor-gadget/inspektor-gadget/blob/e375eb658ecddedaa03471be0f34d7d66be54abb/pkg/gadget-context/run.go#L44-L51

On the other hand, how to manger the global params used for initialization? Should we use the ones coming from a config file or so and overwrite them if something is pass here?

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+
```
blanquicet, line 58:
```suggestion
defer localManagerOp.Close()

```

examples/gadgets/simple/trace_dns/main.go
```diff
@@ -0,0 +1,92 @@
...
+	// The socker enricher operator is used to provide information about the
+	// process performing the DNS request.
+	socketEnricherOp := &socketenricher.SocketEnricher{}
+	socketEnricherOp.Init(nil)
+
```
blanquicet, line 63:
```suggestion
	defer socketEnricherOp.Close()

```

blanquicet (commented):
Awesome! 🚀 

I started reviewing the documentation and the simple/trace_dns. I'll continue tomorrow with the rest. Some discussions are long, so maybe they can become a separate issue?

examples/gadgets/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Examples showing how to use `datasource`
+
+- [fields](./datasource/fields/): Show how to access specific fields from the datasource.
+- [mutate](./datasource/mutate/): Mutate and add fields to a datasource.
+
+### Executing Gadgets on remove Inspektor Gadget instances
+
+Examples showing how to execute gadgets in remove instances of Inspektor Gadget
+(either `ig` or `iìg-k8s`) by using the grpc runtime.
```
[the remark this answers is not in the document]
flyth: Also `iìg-k8s` 😄

examples/gadgets/simple/trace_dns/main.go
flyth: We're currently not doing it correctly. IMHO, the caller should always make sure operators are initialized properly instead of the `gadgetCtx` (which should only do the instance initialization).

Basically all our entry points (ig/main.go, kubectl-gadget/main.go, gadgetctl/main.go and gadgettracermgr) should initialize all operators globally from a config file (or params or env, basically the cobra/viper combo). It could be done by a wrapper so that it works the same way for all of them.

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: IMO we shouldn't try to couple gadgets and operators and this coupling we have is technical debt. SocketFilter gadgets should capture events on the host if the local/kube manager aren't used, SchedCLS ones should be able to be attached to host's interfaces too (by using the --iface param).

I'd suggest to open an issue about it and link it on the example. WDYT?

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: I totally second @flyth, operators should be initialized outside the gadget-context (it's possible to use the same operator for running multiple gadgets). I'll give it a try later on (otherwise I'll open an issue).

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: Good question. What I did was to check the priority used by other operators.

Currently it's the operator that tells that's the priority for the `Subscribe`, it requires operators to know the priorities of other operators (the CLI needs to be sure it has higher priority than the localmanager, formatter, etc). I'm really wondering if this is the right approach, perhaps we can move this responsibility to the caller? It could have a way to define the order of the operators and it should be sure they're called in the right order satisfying the dependencies. 

I know @flyth put a lot of thinking into this. Do you have strong opinions for one or another approach?

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: I added a comment on the trace open example that has more documentation than this.

examples/gadgets/simple/trace_dns/main.go
flyth: As #2709 introduces the difference between Packet and Data (with it's own priority handling), I suggest we define only three priorities as constants for now:

`PriorityDefault` as 0, `PriorityHigh` as something like -50000 and `PriorityLow` as 50000 (or -PriorityHigh). If a dev/user needs something in between, this allows a lot of room. But for normal cases the default priority should work fine. Same can be used for Packets, as they're always handled after the Data subscriptions.

examples/gadgets/simple/trace_dns/README.md
```diff
@@ -0,0 +1,54 @@
...
+Non-authoritative answer:
+Name:   inspektor-gadget.io
+Address: 2606:4700:3030::6815:ba0
+Name:   inspektor-gadget.io
+Address: 2606:4700:3037::ac43:a669
+```
+
+Those will printed in the gadget's terminal:
+
+TODO: check why this gadget is lacking so much information compared to the built-in one.
```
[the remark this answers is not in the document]
blanquicet: IP/Ports issue could be related to https://github.com/inspektor-gadget/inspektor-gadget/issues/2813

examples/gadgets/simple/trace_dns/main.go
blanquicet: > SocketFilter gadgets should capture events on the host if the local/kube manager aren't used
>
But we should show host events only if user request for them with the `--host` flag, which is also managed by the local manager operator 😅. So, if developer doesn't add the local manager operator, thus they can't specify the want the host events. In this situation, what should we do? Simply output nothing or return an error because we know the gadget won't produce anything?

At this point, we should detach `--host` implementation from the local manager.

> SchedCLS ones should be able to be attached to host's interfaces too (by using the --iface param)
>
The advantage here is that the `--iface` flag is manged by the ebpf operator. But what to do if user doesn't specify `--iface`? Return an error?

> I'd suggest to open an issue about it and link it on the example. WDYT?
>
Yep, agree.

examples/gadgets/simple/trace_open/main.go
```diff
@@ -0,0 +1,114 @@
...
+	// TODO: link to documentation with the parameters supported by this gadget
+	params := map[string]string{
+		// Filter only events from the root user
+		"operator.oci.ebpf.uid": "0",
+	}
+
+	// Finally, let's run our gadget
+	if err := runtime.RunGadget(gadgetCtx, nil, params); err != nil {
+		return fmt.Errorf("running gadget: %w", err)
+	}
```
blanquicet, line 104: Don't you think it's a little bit strange that we have a context but we don't wait for it to be done? When using context, I always expect the following pattern:

```go
ctx := newContext()
...
go runGadget(cxt)
...
<-ctx.Done()
```

examples/gadgets/simple/trace_open/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```bash
+$ sudo cat /dev/null
+```
+
+Those will printed in the gadget's terminal:
+
+```bash
+$ sudo ./trace_open
+{"args":"/usr/bin/cat","args_count":2,"args_size":23,"comm":"cat","gid":0,"loginuid":1001,"mntns_id":4026531841,"pid":9999,"ppid":98133,"retval":0,"sessionid":3,"timestamp":17145168898503,"uid":0,"upper_layer":false}
```
blanquicet, line 30: Mmmmm what am I doing wrong? I have different fields:

```bash
{
  "comm": "iio-sensor-prox",
  "err": 0,
  "fd": 8,
  "flags": 2048,
  "fname": "/dev/iio:device0",
  "gid": 0,
  "mntns_id": 4026532919,
  "mode": 0,
  "pid": 1358,
  "timestamp": 55999871673151,
  "uid": 0
}
```

examples/gadgets/datasource/fields/main.go
```diff
@@ -0,0 +1,75 @@
...
+	const opPriority = 50000
+	myOperator := simple.New("myHandler", simple.OnInit(func(gadgetCtx operators.GadgetContext) error {
+		for _, d := range gadgetCtx.GetDataSources() {
+			// Access some specific fields
+			pidF := d.GetField("pid")
+			commF := d.GetField("comm")
+			fnameF := d.GetField("fname")
+
+			d.Subscribe(func(source datasource.DataSource, data datasource.Data) error {
+				pid := pidF.Uint32(data)
```
blanquicet, line 41: How does developer know it's an `uint32`? I think we are missing a helper here to be able to make it more generic:

```go
	pid := new(pidF.Type())
	pidF.Unmarshal(data, pid)
```

Does it make sense to you? Do you think that would be possible?

cc @flyth

examples/gadgets/simple/trace_dns/main.go
blanquicet: BTW, what about the simple operator? Should we call also `myOperator.Init(nil)`? Or we want to skip it as we know it does nothing?

examples/gadgets/datasource/mutate/main.go
```diff
@@ -0,0 +1,92 @@
...
+			// Get accessors to some specific fields
+			uidF := d.GetField("uid")
+
+			d.Subscribe(func(source datasource.DataSource, data datasource.Data) error {
+				uid := uidF.Uint32(data)
+				// if uid is zero, set is_root to true
+				if uid == 0 {
+					// TODO: yes, we know, it has to be improved.
+					isRootF.Set(data, make([]byte, 1))
+					isRootF.PutUint8(data, 1)
```
blanquicet, line 59: This can already be updated now that we merged https://github.com/inspektor-gadget/inspektor-gadget/pull/2823/files

examples/gadgets/operators/cli/main.go
```diff
@@ -0,0 +1,58 @@
...
+	)
+
+	runtime := local.New()
+	if err := runtime.Init(nil); err != nil {
+		return fmt.Errorf("runtime init: %w", err)
+	}
+	defer runtime.Close()
+
+	params := map[string]string{
+		// columns, json, jsonpretty and yaml are supported
```
blanquicet, line 43: TODO: Link to operator documentation

examples/gadgets/operators/local_manager/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```bash
+$ docker run --name mycontainer --rm -it busybox sh -c "cat /dev/null"
+$ docker run --name foocontainer --rm -it busybox sh -c "cat /dev/null"
+```
+
+The gadget only captured the events from `mycontainer`, and you can see how
+they're enriched with the container name:
+
+TODO: why all runtime things aren't displayed by default?
```
blanquicet, line 31: It's because `igjson.WithShowAll(true)`

examples/gadgets/operators/local_manager/main.go
```diff
@@ -0,0 +1,95 @@
...
+				return nil
+			}, opPriority)
+		}
+		return nil
+	}))
+
+	// Create the local manager operator
+	localManagerOp := &localmanager.LocalManager{}
+	localManagerParams := localManagerOp.GlobalParamDescs().ToParams()
+	localManagerParams.Get(localmanager.Runtimes).Set("docker")
```
blanquicet, line 53: We need documentation explaining why some parms are passed here and some other in the `RunGadget`.

examples/gadgets/grpc/custom_operator/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+`ig` needs to be running in daemon mode in the same machine as this example:
+
+```bash
+$ sudo ig daemon --host tcp://127.0.0.1:8888
+INFO[0000] starting Inspektor Gadget daemon at "tcp://127.0.0.1:8888"
+```
+
+Run the binary. In this root is not needed because the all privileged operations
+are done by the ig process.
```
blanquicet, line 22:
```suggestion
are done by the `ig` process.
```

examples/gadgets/grpc/custom_operator/main.go
```diff
@@ -0,0 +1,85 @@
...
+				jsonOutput := jsonFormatter.Marshal(data)
+				fmt.Printf("%s\n", jsonOutput)
+				return nil
+			}, opPriority)
+		}
+		return nil
+	}))
+
+	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT)
+	defer stop()
```
blanquicet, line 51: Why this is needed here and not in other examples?

blanquicet (approved):
I added some comments but LGTM already. Thanks for working on this.

examples/gadgets/simple/trace_dns/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```bash
+$ sudo ./trace_dns
+{"anaddrcount":0,"ancount":0,"gid":0,"id":38647,"latency_ns":0,"mntns_id":4026533917,"name":"\u0010inspektor-gadget\u0002io","netns":4026534560,"pid":52845,"pkt_type":4,"qr":0,"qtype":1,"rcode":0,"runtime":{"containerName":"c3"},"task":"nslookup","tid":52845,"timestamp":10931213449822,"uid":0}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Alright, a lot of changes since I opened this. Now it looks like:

```bash 
{"anaddr":"00000000000000000000000000000000","anaddrcount":0,"ancount":0,"gid":0,"id":892,"latency_ns":0,"mntns_id":4026534842,"name":"inspektor-gadget.io","netns":4026534846,"pid":124823,"pkt_type":4,"qr":0,"qtype":1,"rcode":0,"runtime":{"containerName":"c3"},"task":"nslookup","tid":124823,"timestamp":8559269084592,"uid":0}
{"anaddr":"00000000000000000000000000000000","anaddrcount":0,"ancount":0,"gid":0,"id":8525,"latency_ns":0,"mntns_id":4026534842,"name":"inspektor-gadget.io","netns":4026534846,"pid":124823,"pkt_type":4,"qr":0,"qtype":28,"rcode":0,"runtime":{"containerName":"c3"},"task":"nslookup","tid":124823,"timestamp":8559269134837,"uid":0}
{"anaddr":"00000000000000000000ffffac43a669","anaddrcount":1,"ancount":2,"gid":0,"id":892,"latency_ns":2351945661,"mntns_id":4026534842,"name":"inspektor-gadget.io","netns":4026534846,"pid":124823,"pkt_type":0,"qr":1,"qtype":1,"rcode":0,"runtime":{"containerName":"c3"},"task":"nslookup","tid":124823,"timestamp":8561621030253,"uid":0}
```

after #2845

examples/gadgets/simple/trace_open/README.md
mauriciovasquezbernal: I updated it with the latest output. Probably something changed in between.

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: I think we can skip it.

examples/gadgets/datasource/mutate/main.go
mauriciovasquezbernal: Good one, it was even failing to compile.

examples/gadgets/operators/local_manager/main.go
mauriciovasquezbernal: We need all documentation regarding the params that the operators support. I'll open an issue about it later.

examples/gadgets/simple/trace_open/main.go
mauriciovasquezbernal: In this case we're controlling the context, so why to wait for it to be done? Who will mark it as done if it's not the caller? 

Perhaps we can add a signal handler and so on, but I think it's too much for this simple example.

examples/gadgets/datasource/fields/main.go
mauriciovasquezbernal: @flyth suggested to have a `Any()` or similar accessor method in https://github.com/inspektor-gadget/inspektor-gadget/pull/2753#discussion_r1575290682. In this case I think we're missing documentation, there should be a place that mentions all the fields that are provided by a gadget, their meaning and type. Perhaps they can be extracted from the metadata file and exposed in a markdown?, or let the user check the metadata file directly.

examples/gadgets/grpc/custom_operator/main.go
mauriciovasquezbernal: It's needed to close the connection in a clean way, otherwise the server will print something like:

`WARN[0040] error on connection: rpc error: code = Canceled desc = context canceled`

I added a comment.

examples/gadgets/operators/local_manager/README.md
```diff
@@ -0,0 +1,102 @@
...
+{"comm":"cat","err":0,"fd":3,"flags":524288,"fname":"/lib/libc.so.6","gid":0,"k8s":{"container":"","hostnetwork":false,"namespace":"","node":"","pod":""},"mntns_id":4026535202,"mode":0,"pid":129758,"runtime":{"containerId":"ca95d6d4fe64d4b2bc73a0d609c7fc80968e269e2660cca540a767840b03a58d","containerImageDigest":"","containerImageName":"busybox","containerName":"mycontainer","runtimeName":"docker"},"timestamp":9701345683679,"uid":0}
+{"comm":"cat","err":0,"fd":3,"flags":0,"fname":"/dev/null","gid":0,"k8s":{"container":"","hostnetwork":false,"namespace":"","node":"","pod":""},"mntns_id":4026535202,"mode":0,"pid":129758,"runtime":{"containerId":"ca95d6d4fe64d4b2bc73a0d609c7fc80968e269e2660cca540a767840b03a58d","containerImageDigest":"","containerImageName":"busybox","containerName":"mycontainer","runtimeName":"docker"},"timestamp":9701346391074,"uid":0}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Makes sense. Also previous text was wrong because I messed up copying and pasting.

examples/gadgets/grpc/custom_operator/main.go
```diff
@@ -0,0 +1,86 @@
...
+	gadgetCtx := gadgetcontext.New(
+		ctx,
+		"ghcr.io/inspektor-gadget/gadget/trace_open:latest",
+		gadgetcontext.WithDataOperators(myOperator),
+	)
+
+	runtime := grpcruntime.New()
+	runtimeGlobalParams := runtime.GlobalParamDescs().ToParams()
+	runtimeGlobalParams.Get(grpcruntime.ParamRemoteAddress).Set("tcp://127.0.0.1:8888")
+	//runtimeGlobalParams.Get(grpcruntime.ParamRemoteAddress).Set("unix:///var/run/ig/ig.socket")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I wanted to show examples of both, tcp and unix. I added a comment.

mauriciovasquezbernal:
> I am wondering nonetheless if we should not put them in the existing folder.

Where to you mean?

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: Opened #2891

examples/gadgets/simple/trace_dns/main.go
mauriciovasquezbernal: Opened #2892

examples/gadgets/simple_trace_open/main.go
mauriciovasquezbernal: Opened #2893 with some thoughts. I haven't have the change to explore this further.

examples/gadgets/simple/trace_open/main.go
```diff
@@ -0,0 +1,120 @@
...
+			)
+
+			d.Subscribe(func(source datasource.DataSource, data datasource.Data) error {
+				jsonOutput := jsonFormatter.Marshal(data)
+				fmt.Printf("%s\n", jsonOutput)
+				return nil
+			}, opPriority)
+		}
+		return nil
+	}))
```
blanquicet, line 76: Now that I'm working on [the combiner operator](https://github.com/inspektor-gadget/inspektor-gadget/issues/1924#issuecomment-2127209677), I noticed that by subscribing to a data source in the `OnInit` phase there's a risk that a new operator registers a new data source afte us and we miss it. Of course, you can rely on the priority at the moment you add your operator, but what about new operator created later on by other people?

One solution is to never call `GetDataSources` on the `OnInit` phase as other data sources could be registered in operators with lower priority and instead, do it in the `OnStart` phase where all operators are supposed to have already registered their data sources. But again, there's no guarantee.

We need to describe in the documentation to use high priority for operators registering datasource and maybe also tell that it's a good practice to use the `OnInit` phase is for datasource registration and the `OnStart` phase for subscribing to datasources. WDYT?

cc @flyth @mauriciovasquezbernal

flyth: I drafted [this](https://gist.github.com/flyth/d98befcdd7a1c6558af9af88012ee89a#operators) earlier which contains some information on the lifecycle hooks.

Basically:
* OnInit() / Instantiate(): Add & manipulate DataSources - priority of operator handles initialization order; this is what GetGadgetInfo() returns
* PreStart() (optional): Subscribe to DataSources (to prevent missing some data) - priority determines handling order
* Start(): Emit data, Subscriptions will be evaluated
* Stop(): Stop everything

Regarding subscription priorities: I think filtering should be done before sorting on the server side. Combining should be done (on the client side only) _before_ sorting but after filtering. So sort/filter would run on both client+server.

I hope this will work for 90% of the use cases 😄

mauriciovasquezbernal: I think I messed up the examples because there are two priorities:
- operator priority: defines the order on which operators are initialized, started and stopped. 
- subscription priority: order in which subscriptions are called

Those two are independent, an operator with a low priority (is initialized last), could call subscribe with a high priority (its callback will be called first. 

@flyth what's the design decision to have a priority for the operator and not to rely on the order used in WithDataOperators()?

flyth: > what's the design decision to have a priority for the operator and not to rely on the order used in WithDataOperators()?

Explicitly adding operators using `WithDataOperators()` was an option added later on; most (default) operators register themselves with the registry, so at least for that I think it makes sense to have the priority mentioned inside the operator (although we could make it overridable later on - e.g. the gadget.yaml could specify the order, or we could have a `WithDataOperatorsOrdered()` that would honor the order in which operators are given, overriding their preset priorities).

Issue #1433, opened 2023-03-13
Implement extended support for bpftrace


mauriciovasquezbernal:
From #1342 

This is the continuation of the work initiated on #1432. The purpose of this phase is to implement the ability to filter and enrich the data based on Kubernetes concepts like namespace/pods/container etc. 

```bash
# Filter events by container name 
$ kubectl gadget bpftrace -containername foo -e `tracepoint...` 
# or 
$ kubectl gadget bpftrace -e `tracepoint .. /containername=="foo"/ ...` 

# Print k8s related info  
$ kubectl gadget bpftrace -e 'tracepoint:syscalls:sys_enter_read { printf(“%s\n”, containername)}'
``` 

@alban did some experimentation using bpf extensions https://github.com/inspektor-gadget/inspektor-gadget/issues/1342#issuecomment-1455989538.

PR #2757, opened 2024-04-23
Remove `bpftrace`

eiffel-fl:
Hi.

We seldomly use the `script` gadget and it has several drawbacks:
1. The version of `bpftrace` we pack is outdated it is based on work done some time ago.
2. `bpftrace` needs `libc` which causes our scan job to fail time to time due to CVE in this library: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/8802366953/job/24158260570#step:4:43

Best regards.

### DIFF
--- a/Dockerfiles/gadget.Dockerfile
+++ b/Dockerfiles/gadget.Dockerfile
@@ -1,7 +1,7 @@
 # Dockerfile for Inspektor Gadget.
 
 ARG BUILDER_IMAGE=golang:1.22-bullseye
-ARG BASE_IMAGE=gcr.io/distroless/base-debian12
+ARG BASE_IMAGE=gcr.io/distroless/static-debian12
 
 # Prepare and build gadget artifacts in a container
 FROM --platform=${BUILDPLATFORM} ${BUILDER_IMAGE} as builder


charts/gadget/templates/daemonset.yaml
```diff
@@ -146,10 +146,6 @@ spec:
                 # https://github.com/torvalds/linux/blob/c42d9eeef8e5/fs/proc/namespaces.c#L58
                 - SYS_PTRACE
 
-                # Needed by setrlimit in bpftrace:
-                # https://github.com/iovisor/bpftrace/blob/ab863ef93ac1/src/main.cpp#L154
-                - SYS_RESOURCE
```
alban, line 151: We also need `SYS_RESOURCE` without bpftrace:
```
$ git grep rlimit.RemoveMemlock -- pkg/
pkg/gadgettracermanager/gadgettracermanager.go:258:             if err := rlimit.RemoveMemlock(); err != nil {
pkg/ig-manager/ig-manager.go:86:        if err := rlimit.RemoveMemlock(); err != nil {
```

alban:
I am also in favour of removing bpftrace as well for now.

Right now, the `script` gadget does not bring much value:
- With ig: it does not bring much value compared to using bpftrace directly
- With kubectl-gadget: there is already kubectl-trace for this

The `script` gadget can always be added later on when we have better integration and enrichment. But so far we didn't have time to do this additional work and meanwhile it increases the amount of work to maintain it. Moreover, I think that when we add the `script` gadget back, it will likely be a different architecture, so we don't save time by keeping it in the inspektor-gadget repository as it is today.

---

About the commit "charts: Remove SYS_RESOURCE": I think it does not work because IG uses rlimit too. In general, I would like any PR changing the capabilities to be tested on ARO and other Kubernetes distros (with "citest").

alban:
@kranurag7: are you currently using the script gadget?

mauriciovasquezbernal (approved):
Maintaining this is being a PITA, hence LGTM. A proper and more powerful implementation will be done in #1433.

Issue #2365, opened 2024-01-15
Enrich docker events with `ContainerImageDigest`


claudiamarcubina:
## Current situation

#2070 , #2143 and #2273 added `ContainerImageDigest` enrichment as follows:

| ig     | Initial containers | New containers |
|------------|--------------------|----------------|
| containerd | done                  | done               |
| cri-o      | done                  | blocked (#2273)  |
| docker     | TODO                  | TODO              |

| kubectl-gadget     | Initial containers | New containers |
|------------|--------------------|----------------|
| containerd | done                  | TODO (#2359)|
| cri-o      | done                  | blocked (#2273)  |
| docker     | TODO                  | TODO              |

This task is tracking the enrichment of docker events with `ContainerImageDigest` for both `ig` and `kubectl-gadget`.

## Implementation options

Implementation options still need to be researched.

See topic history in #1310

mmat11:
hi @claudiamarcubina I opened https://github.com/inspektor-gadget/inspektor-gadget/pull/2798 but I am not sure I covered both initial and new containers; can you have a look when you have time and perhaps start a CI run?

PR #2759, opened 2024-04-23
gadgets/trace_open/test: Verify container name and ID

blanquicet:
# Verify container name and ID in trace_open gadget test

- Leverage the [containers.TestContainer](https://github.com/inspektor-gadget/inspektor-gadget/blob/97eaf2fc80da992a6ed1b54cebc861bbf91844bd/pkg/testing/containers/factory.go#L45-L48) methods to retrieve the container ID and verify it in the expected output.
- Specify the image we want to use and verify it in the expected output.
- NOTE: The image digest is not yet enriched for Docker. See https://github.com/inspektor-gadget/inspektor-gadget/issues/2365.

## Testing done

CI is green 🙂 

## TODO (Future PR)

The [containers.TestContainer](https://github.com/inspektor-gadget/inspektor-gadget/blob/97eaf2fc80da992a6ed1b54cebc861bbf91844bd/pkg/testing/containers/factory.go#L45-L48) also provides the `Pid()`. Using that value, we can retrieve the mount ns inode as we do in [getNsInode](https://github.com/inspektor-gadget/inspektor-gadget/blob/97eaf2fc80da992a6ed1b54cebc861bbf91844bd/internal/test/runner.go#L205-L218) and verify also the `MountNsID` field in the test.

### DIFF
--- a/gadgets/trace_open/test/trace_open_test.go
+++ b/gadgets/trace_open/test/trace_open_test.go
@@ -46,14 +46,18 @@ type traceOpenEvent struct {
 func TestTraceOpen(t *testing.T) {
 	gadgettesting.RequireEnvironmentVariables(t)
 
-	cn := "test-trace-open"
 	containerFactory, err := containers.NewContainerFactory("docker")
 	require.NoError(t, err, "new container factory")
 
+	containerName := "test-trace-open"
+	containerImage := "docker.io/library/busybox"
+
 	testContainer := containerFactory.NewContainer(
-		cn,
+		containerName,
 		"while true; do setuidgid 1000:1111 cat /dev/null; sleep 0.1; done",
+		containers.WithContainerImage(containerImage),
 	)
+
 	testContainer.Start(t)
 	t.Cleanup(func() {
 		testContainer.Stop(t)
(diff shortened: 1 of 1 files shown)


gadgets/trace_open/test/trace_open_test.go
```diff
@@ -46,14 +46,18 @@ type traceOpenEvent struct {
 func TestTraceOpen(t *testing.T) {
 	gadgettesting.RequireEnvironmentVariables(t)
 
-	cn := "test-trace-open"
 	containerFactory, err := containers.NewContainerFactory("docker")
 	require.NoError(t, err, "new container factory")
 
+	cn := "test-trace-open"
```
mauriciovasquezbernal, line 54: what about using containerName? I always think of Cartoon Network when reading cn...

gadgets/trace_open/test/trace_open_test.go
```diff
@@ -46,20 +47,36 @@ type traceOpenEvent struct {
 func TestTraceOpen(t *testing.T) {
 	gadgettesting.RequireEnvironmentVariables(t)
 
-	cn := "test-trace-open"
-	containerFactory, err := containers.NewContainerFactory("docker")
+	runtime := "docker"
```
mauriciovasquezbernal, line 50: I suppose it's fairly simple to allow this to be configurable (in another PR for sure). Using an env variable seems the easiest way, right?

mauriciovasquezbernal (approved):
Looks great from code inspection. I really like where the testing is going.

gadgets/trace_open/test/trace_open_test.go
blanquicet: Yep, that's why I made that change. To prepare everything to make it easily configurable.

blanquicet: I'm still not sure if I prefer using env variable to pass this kind of configuration of flags, like [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/d54838472c7a2be0894f90369bb3068e630415d9/integration/ig/non-k8s/integration_test.go#L32).

mauriciovasquezbernal: I think the flag doesn't work well as we'd need to have a `TestMain` function to parse the flag, then running the test alone (in the gadget folder) won't work.

PR #2766, opened 2024-04-24
pkg/{runcfanotify, container-hook}: Refactor code to use notify package.

eiffel-fl:
Hi.

This PR refactors the common code between `runcfanotify` and `container-hook`.
I also unified the runtime paths array, so they now count the same number for both hooks and it is now easier to add a new hook.

Best regards.

### DIFF
--- a/pkg/container-hook/tracer.go
+++ b/pkg/container-hook/tracer.go
@@ -47,13 +47,13 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
-	securejoin "github.com/cyphar/filepath-securejoin"
 	ocispec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/s3rj1k/go-fanotify/fanotify"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/btfgen"
+	runtimefinder "github.com/inspektor-gadget/inspektor-gadget/pkg/container-hook/runtime-finder"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/kfilefields"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
@@ -137,24 +137,10 @@ type ContainerNotifier struct {
 	wg sync.WaitGroup
 }
 
-// runtimePaths is the list of paths where the container runtime runc or crun
-// could be installed. Depending on the Linux distribution, it could be in
-// different locations.
-//
-// When this package is executed in a container, it prepends the
-// HOST_ROOT env variable to the path.
-var runtimePaths = []string{
-	"/bin/runc",
-	"/usr/bin/runc",
-	"/usr/sbin/runc",
-	"/usr/local/bin/runc",
-	"/usr/local/sbin/runc",
-	"/usr/lib/cri-o-runc/sbin/runc",
-	"/run/torcx/unpack/docker/bin/runc",
-	"/usr/bin/crun",
+var runtimePaths []string = append(
+	runtimefinder.RuntimePaths,
 	"/usr/bin/conmon",
-	"/var/lib/rancher/k3s/data/current/bin/runc",
-}
+)
 
 // initFanotify initializes the fanotify API with the flags we need
 func initFanotify() (*fanotify.NotifyFD, error) {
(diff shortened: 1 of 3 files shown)


pkg/container-hook/tracer.go
```diff
@@ -47,7 +47,6 @@ import (
```
blanquicet, line 47: I don't think we should this. Instead, we should deprecate and delete runc-fanotify in favour of container-hook https://github.com/inspektor-gadget/inspektor-gadget/issues/1831.

pkg/utils/notify/notify.go
```diff
@@ -0,0 +1,70 @@
...
+var RuntimePaths = []string{
+	"/bin/runc",
+	"/usr/bin/runc",
+	"/usr/sbin/runc",
+	"/usr/local/bin/runc",
+	"/usr/local/sbin/runc",
+	"/usr/lib/cri-o-runc/sbin/runc",
+	"/run/torcx/unpack/docker/bin/runc", // Used in TODO
+	"/usr/bin/crun",
+	"/usr/bin/conmon",
```
alban, line 43: "conmon" should be in the list for `container-hook` but not for `runcfanotify`, see https://github.com/inspektor-gadget/inspektor-gadget/pull/2764#discussion_r1583272674

alban:
Supersedes: #2764

pkg/utils/notify/notify.go
```diff
@@ -0,0 +1,70 @@
...
+// When this package is executed in a container, it prepends the HOST_ROOT env
+// variable to the path.
+var RuntimePaths = []string{
+	"/bin/runc",
+	"/usr/bin/runc",
+	"/usr/sbin/runc",
+	"/usr/local/bin/runc",
+	"/usr/local/sbin/runc",
+	"/usr/lib/cri-o-runc/sbin/runc",
+	"/run/torcx/unpack/docker/bin/runc", // Used in TODO
```
alban, line 41:
```suggestion
	"/run/torcx/unpack/docker/bin/runc", // Used in Flatcar Container Linux
```

pkg/container-hook/tracer.go
```diff
@@ -57,6 +56,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/kfilefields"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/notify"
```
alban, line 59: What about moving `pkg/utils/notify/notify.go` to `pkg/container-hook/runtime-finder/finder.go`?

"utils" is a bit of an anti-pattern: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

And when we get rid of `runcfanotify`, it will already be in the right place, in a subdirectory of `container-hook`.

pkg/utils/notify/notify.go
alban: Actually, https://www.flatcar.org/blog/2024/04/os-innovation-with-systemd-sysext/ says:

> Therefore, we recently removed Torcx and recommend using systemd-sysext for deploying custom Docker/containerd versions.

But we can keep the torcx path for a while for compatibility with older Flatcar versions.

@pothos When using systemd-sysext to install `runc`, `crun` or `conmon`, will it be in `/usr/bin/` or do we need to add another path here?

pkg/container-hook/tracer.go
alban: IMO it's ok for `runcfanotify` to use a pkg in a subdirectory of `container-hook`.

pkg/utils/notify/notify.go
pothos: Should be in `/usr/bin/`: https://stable.release.flatcar-linux.net/amd64-usr/current/rootfs-included-sysexts/containerd-flatcar_contents.txt https://stable.release.flatcar-linux.net/amd64-usr/current/rootfs-included-sysexts/docker-flatcar_contents.txt

pkg/runcfanotify/runcfanotify.go
```diff
@@ -180,38 +165,34 @@ func NewRuncNotifier(callback RuncNotifyFunc) (*RuncNotifier, error) {
...
+			if err != nil {
+				log.Debugf("runc-fanotify: notifying %s: %v", r, err)
 				continue
 			}
 
-			if err := runcBinaryNotify.Mark(unix.FAN_MARK_ADD, unix.FAN_OPEN_EXEC_PERM, unix.AT_FDCWD, runcPath); err != nil {
-				log.Warnf("Runcfanotify: failed to fanotify mark: %s", err)
-				continue
-			}
+			log.Debugf("runc-fanotify: monitoring runtime at %s (originally %s)", r, notifiedPath)
```
alban, line 187: Order of %s parameters

pkg/container-hook/tracer.go
```diff
@@ -57,6 +56,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/kfilefields"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
+	runtimefinder "github.com/inspektor-gadget/inspektor-gadget/pkg/container-hook/runtime-finder"
```
alban, line 59: Please use `gofumpt -w` on both tracer.go and runcfanotify.go.

alban (approved):
3 small things to fix, LGTM after that (if CI is green), from code inspection.

I would like @blanquicet to say whether his concern is addressed. In my opinion, this is fine to merge this PR even if we remove runcfanotify after.

PR #2768, opened 2024-04-24
trace/{dns,sni}: Split NewTracer() logic

amirmalka:
# Split NewTracer() logic in dns and sni tracers

Before this change NewTracer() installed and run the tracer.
Now calling NewTracer() only installs the tracer, and Run() or RunWorkaround() runs it.
This change is needed so that the socket enricher could be set before loading the eBPF programs.

A similar solution was implemented in the network tracer.

Discussion: https://kubernetes.slack.com/archives/CSYL75LF6/p1713707118485529

## How to use

## Testing done

Before this change, `MountNsID` was empty even when setting the socket enricher. I have run the DNS tracer with this change implemented and I could see that mount namespace was filled successfully and so enrichment of events is working.


mauriciovasquezbernal (approved):
LGTM.

Merging as the CI failure is not related to this PR.

PR #2780, opened 2024-04-26
gadget-service: Improve address validation

mauriciovasquezbernal:
Make logic fail on socket paths like tcp://foo:9000

---

The following command doesn't show any error: (notice the missing `//` after `tcp`)

```bash 
$ sudo ig daemon --host tcp:127.0.0.1:8888
INFO[0000] starting Inspektor Gadget daemon at "tcp:127.0.0.1:8888"
```

But also doesn't work as ig doesn't listen on that address. Make it print an explicit error in that case.


flyth (approved):
Looks good from code inspection. Should we add an example in the error message?

PR #2785, opened 2024-04-29
TestRunTraceOOMKill: Add sleep after starting ig/kubectl-gadget

burak-ok:
This waits for 10 seconds for ig or kubectl-gadget to finished downloading the gadget from the OCI registry and starting it

Reference: For the OOMKill test from the builtin legacy gadget we even have a 2 second sleep there. With image based gadget we need a little bit more time to additionally download the image

### DIFF
--- a/integration/k8s/run_trace_oomkill_test.go
+++ b/integration/k8s/run_trace_oomkill_test.go
@@ -91,6 +91,9 @@ spec:
 
 	commands := []TestStep{
 		traceOOMKillCmd,
+		// Wait to ensure ig or kubectl-gadget has downloaded the image and started it
+		// before we start the test pod, which only gives us 1 event to trace (oneshot)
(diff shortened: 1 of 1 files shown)


alban:
This looks fine, but shouldn't we do the same for other image-based gadgets in `integration/k8s/run_*_test.go`?

burak-ok:
> This looks fine, but shouldn't we do the same for other image-based gadgets in `integration/k8s/run_*_test.go`?

It depends. Most other tests have a `while true` loop in the test pod. This OOMKill is like a oneshot. If you got the single event, good. If not then it would fail

alban (approved):
LGTM

Optionally, you could add a comment explaining that it is a one shot test and not a loop test, so it is necessary to wait for the ig/kubectl-gadget command to be ready.

PR #2786, opened 2024-04-29
tools: Rewrite check-readme in go

mauriciovasquezbernal:
Trying to keep the languages and tools we use as
uniform as possible.

---

I  was trying to fix the CI on #2774, I found out I have not idea how this perl thing works. Asked internally and other folks were confused as well, so I decided to switch to something our team knows better.


tools/check-readme/main.go
```diff
@@ -0,0 +1,79 @@
...
+	endIndex := slices.Index(readmeLines[startIndex:], "...")
+	if endIndex == -1 {
+		return fmt.Errorf("finding '...'")
+	}
+
+	// Remove unuseful things
+	readmeLines = readmeLines[startIndex+1 : startIndex+endIndex]
+	kubectlGadgetLines = kubectlGadgetLines[:len(readmeLines)-1]
+
+	// Compare line by line. TODO: compare whole thing and print preffy diff
```
alban, line 59:
```suggestion
	// Compare line by line. TODO: compare whole thing and print pretty diff
```

tools/check-readme/main.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Read the whole readme and find where the $ kubectl gadget --help part is
+	readmeLines := strings.Split(string(file), "\n")
+	startIndex := slices.Index(readmeLines, "$ kubectl gadget --help")
+	if startIndex == -1 {
+		return fmt.Errorf("finding '$ kubectl gadget --help'")
+	}
+	endIndex := slices.Index(readmeLines[startIndex:], "...")
+	if endIndex == -1 {
+		return fmt.Errorf("finding '...'")
```
alban, line 52: `$ kubectl gadget --help` and `...` can be constants.

tools/check-readme/main.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	// Remove unuseful things
+	readmeLines = readmeLines[startIndex+1 : startIndex+endIndex]
+	kubectlGadgetLines = kubectlGadgetLines[:len(readmeLines)-1]
+
+	// Compare line by line. TODO: compare whole thing and print preffy diff
+	for i := 0; i < len(kubectlGadgetLines); i++ {
+		if kubectlGadgetLines[i] != readmeLines[i] {
```
alban, line 61: Check `len(readmeLines)` before the for loop to ensure that `readmeLines[i]` does not panic.

tools/check-readme/main.go
```diff
@@ -0,0 +1,79 @@
...
+			return fmt.Errorf("lines don't match: \n%q\n%q", kubectlGadgetLines[i], readmeLines[i])
```
alban, line 62:
```suggestion
			return fmt.Errorf("line %d doesn't match: \n%q\n%q", i, kubectlGadgetLines[i], readmeLines[i])
```

tools/check-readme/main.go
```diff
@@ -0,0 +1,79 @@
...
+		return fmt.Errorf("finding '$ kubectl gadget --help'")
+	}
+	endIndex := slices.Index(readmeLines[startIndex:], "...")
+	if endIndex == -1 {
+		return fmt.Errorf("finding '...'")
+	}
+
+	// Remove unuseful things
+	readmeLines = readmeLines[startIndex+1 : startIndex+endIndex]
+	kubectlGadgetLines = kubectlGadgetLines[:len(readmeLines)-1]
```
alban, line 57: Check `len(kubectlGadgetLines)` for out of range panic.

alban (commented):
I understand Go code better too.

tools/check-readme/main.go
mauriciovasquezbernal: I added this single check. The one in the loop shouldn't be needed after it.

tools/check-readme/main.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2019-2024 The Inspektor Gadget authors
```
burak-ok, line 1:
```suggestion
// Copyright 2024 The Inspektor Gadget authors
```

tools/check-readme/main.go
```diff
@@ -0,0 +1,87 @@
...
+	if endIndex == -1 {
+		return fmt.Errorf("finding %q", threeDots)
+	}
+
+	// Remove unuseful things
+	readmeLines = readmeLines[startIndex+1 : startIndex+endIndex]
+	if len(kubectlGadgetLines) < len(readmeLines) {
+		return fmt.Errorf("kubectl-gadget output is shorter than README.md")
+	}
+	kubectlGadgetLines = kubectlGadgetLines[:len(readmeLines)]
```
burak-ok, line 65: You don't need this line if `len(readmeLines)` is used in the for loop condition below

tools/check-readme/main.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Compare line by line. TODO: compare whole thing and print pretty diff
```
burak-ok, line 67: Maybe https://github.com/sergi/go-diff

burak-ok (commented):
Yep. Go seems better :D

tools/check-readme/main.go
mauriciovasquezbernal: That's right, however I'd prefer to have both slices of the same exact length to make it easier to use something like go-diff later on.

burak-ok (approved):

PR #2790, opened 2024-04-30
integration: Logging timestamp before calling step.Stop

burak-ok:
# integration: Logging timestamp before calling step.Stop

When `step.Stop` uses `t.Fatalf` we will never get the timestamp. This way we know when we are initiating the stop at least


mqasimsarfraz (approved):
LGTM!

PR #2795, opened 2024-05-02
New gadget: trace dns with uprobes

alban:
# New gadget: trace dns with uprobes

As an alternative to the _trace dns_ gadget, this new gadget is using uprobe on `getaddrinfo` instead of socket filter.

## How to use

```
export IG_EXPERIMENTAL=true
cd gadgets/trace_udns
sudo -E ig image build -t trace_udns --update-metadata .
sudo -E ig run trace_udns --verify-image=false
```

Example of output:
```
RUNTIME.CONTAINERNAME                     MNTNS_ID          PID                    TID           COMM             NAME                  OPERATION_STR        
mycontainer                               4026534326        2604799                2604799       wget             www.wikipedia.org.    getaddrinfo          
```

## Testing done

Just the basic

Inspired from:
* https://github.com/libbpf/libbpf-bootstrap/blob/master/examples/rust/tracecon/src/bpf/tracecon.bpf.c
* https://www.brendangregg.com/blog/2016-02-08/linux-ebpf-bcc-uprobes.html

TODO:
- [ ] Add the IPs from the DNS answers
- [ ] Support other functions such as gethostbyname and gethostbyname2

### DIFF
--- a/gadgets/trace_udns/program.bpf.c
+++ b/gadgets/trace_udns/program.bpf.c
@@ -1,0 +1,154 @@
+// SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+/* Copyright (c) 2024 The Inspektor Gadget authors */
+
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#include <gadget/buffer.h>
+#include <gadget/macros.h>
+#include <gadget/mntns_filter.h>
+#include <gadget/types.h>
+
+#define AF_INET 2
+#define AF_INET6 10
+
+enum op {
+	getaddrinfo,
+	gethostbyname,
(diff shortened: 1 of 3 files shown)


gadgets/trace_udns/program.bpf.c
```diff
@@ -0,0 +1,154 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, u32); // tid
+	__type(value, struct lookup);
+} lookups SEC(".maps");
+
+SEC("uprobe/libc:getaddrinfo")
+int BPF_KPROBE(getaddrinfo_e, const char *node, const char *service,
```
i-Pear, line 70: weak suggestion: you can also use `BPF_UPROBE`, and the two macros are exactly the same.

PR #2796, opened 2024-05-02
datasources: Add Bool accessors

mauriciovasquezbernal:
Handling bool fields in useful in some cases, for instance one of the examples in #2753 needs this.

### DIFF
--- a/pkg/datasource/accessors.go
+++ b/pkg/datasource/accessors.go
@@ -88,6 +88,7 @@ type FieldAccessor interface {
 	Float64(Data) float64
 	String(Data) string
 	Bytes(Data) []byte
+	Bool(Data) bool
 
 	PutUint8(Data, uint8)
 	PutUint16(Data, uint16)
@@ -98,6 +99,7 @@ type FieldAccessor interface {
 	PutInt32(Data, int32)
 	PutInt64(Data, int64)
 	PutBytes(Data, []byte)
+	PutBool(Data, bool)
 }
 
 type fieldAccessor struct {
@@ -400,6 +402,14 @@ func (a *fieldAccessor) Bytes(data Data) []byte {
 	return a.Get(data)
 }
 
+func (a *fieldAccessor) Bool(data Data) bool {
+	val := a.Get(data)
+	if len(val) < 1 {
+		return false
+	}
+	return val[0] == 1
+}
+
 func (a *fieldAccessor) PutUint8(data Data, val uint8) {
 	a.Get(data)[0] = val
 }
@@ -434,4 +444,12 @@ func (a *fieldAccessor) PutInt64(data Data, val int64) {
 
 func (a *fieldAccessor) PutBytes(data Data, val []byte) {
 	a.Set(data, val)
+}
+
+func (a *fieldAccessor) PutBool(data Data, val bool) {
+	if val {
+		a.Get(data)[0] = 1
+	} else {
+		a.Get(data)[0] = 0
+	}
 }
--- a/pkg/datasource/formatters/json/json.go
+++ b/pkg/datasource/formatters/json/json.go
@@ -251,14 +251,11 @@ func (f *Formatter) addSubFields(accessors []datasource.FieldAccessor, prefix st
 			}
 		case api.Kind_Bool:
 			fn = func(e *encodeState, data datasource.Data) {
-				// handle arbitrary length bools
-				for b := range accessor.Get(data) {
-					if b != 0 {
-						e.WriteString("true")
-						return
-					}
+				if accessor.Bool(data) {
+					e.WriteString("true")
+				} else {
+					e.WriteString("false")
 				}
-				e.WriteString("false")
 			}
 		default:
 			fn = func(e *encodeState, data datasource.Data) {


burak-ok (approved):
LGTM

mauriciovasquezbernal:
I added another commit. Would you mind checking this as well @burak-ok ?

@flyth any comments on this?

pkg/datasource/accessors.go
```diff
@@ -382,6 +384,14 @@ func (a *fieldAccessor) Int64(data Data) int64 {
 	return int64(a.ds.byteOrder.Uint64(val))
 }
 
+func (a *fieldAccessor) Bool(data Data) bool {
+	val := a.Get(data)
+	if len(val) < 1 {
```
burak-ok, line 389: Now that I think about it, when could this happen?

pkg/datasource/formatters/json/json.go
```diff
@@ -250,14 +250,11 @@ func (f *Formatter) addSubFields(accessors []datasource.FieldAccessor, prefix st
...
-				// handle arbitrary length bools
-				for b := range accessor.Get(data) {
-					if b != 0 {
-						e.WriteString("true")
-						return
-					}
+				if accessor.Bool(data) {
+					e.WriteString("true")
+				} else {
+					e.WriteString("false")
```
flyth, line 256: This changes the behavior as it doesn't handle arbitrary length booleans anymore. The way it was implemented before made sure that any non-zero value as part of the byte slice was considered truthy (so was working also with longer ints and both endiannesses.)

pkg/datasource/accessors.go
flyth: Depends on the origin of the field; if it's a member of a eBPF struct, it usually won't happen as there are other checks in place. If you add a new field manually and didn't reserve space for it, it will be of length 0.

There's room for improvement on how we could internally handle that (e.g. let the DataSource handle memory allocation in the latter case), but that's outside of the scope of this PR :)

pkg/datasource/formatters/json/json.go
mauriciovasquezbernal: I understand your point but I'm not sure what's the right way to handle it. I think this should be aligned with the `Bool()` implementation (that right now only checks the first byte of the byte array). We could modify this to check all bytes of the array, but then something like the following could happen:

```go
f.PutUint16(256) // 1<< 8 (last byte is 0)

f.PutBool(false)

f.Bool() // returns true 
```

(that could be fixed by making PutBool clear all the bytes, but it seems an overkill to me)

IMHO it's users' responsibility to use the right method depending on the type stored in the variable. Perhaps we can check the type to enforce this? i.e. not Bool() can be called on integer fields and so on. WDYT?

flyth: > (that could be fixed by making PutBool clear all the bytes, but it seems an overkill to me)

Hmm, isn't that exactly how C would treat it? So I'd also set the field to `1` using the PutInt() method that works for the length.

> IMHO it's users' responsibility to use the right method depending on the type stored in the variable. Perhaps we can check the type to enforce this? i.e. not Bool() can be called on integer fields and so on. WDYT?

I'm not sure - IMHO if you flag something as bool in eBPF and it comes out as an int, it should still be treatable as bool on our side.

mauriciovasquezbernal: > isn't that exactly how C would treat it?

I don't think so, doesn't it depends on the casting you perform? Something like the following

```c
#include <stdio.h>
#include <stdlib.h>
#include <stdbool.h>
#include <stdint.h>

int main() {
    uint8_t data[8] = {0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8};

    uint64_t *u64 = (uint64_t*)data;
    printf("as uint64: %lu\n", *u64);

    bool *b = (bool*)data;
    printf("as bool: %d\n", *b);

    *b = false;
    printf("as bool: %d\n", *b);
    printf("as uint64: %lu\n", *u64);

    return 0;
}
```

prints 

```bash 
as uint64: 578437695752307201
as bool: 1
as bool: 0
as uint64: 578437695752307200
```

> So I'd also set the field to 1 using the PutInt() method that works for the length.

I don't follow this one. What do you mean?

> I'm not sure - IMHO if you flag something as bool in eBPF and it comes out as an int, it should still be treatable as bool on our side.

Why could it come out as an int? (not following this one as well)

I think the whole discussion goes into the definition we want to give to the bool field:

1. Bool() == false means all bytes are set to 0 (previous json implementation)
2. Bool() == false means the first byte is set to 0 (proposal in this PR)

I'm more for the second one as I don't see why we should make Bool() dependent on the values of all bytes. Same happens for other types. Mixing calls with different types will provide undefined output:

```go
f.Set([]byte("test-string-yes"))
f.Uint64() -> ??
f.Float32() -> ?? 

f.PutFloat32(3.1416) // I just realized we don't have this method, should we?
f.String() -> ??
```

mauriciovasquezbernal:
@alban I think we need another opinion on this topic, would you mind sharing your thoughts on this one?

flyth (approved):
I'm fine with merging this as it is, as it is blocking or at least complicating other PRs already. We can do a closer look into how different implementations might handle bools in a follow-up issue.

mauriciovasquezbernal:
Created #2826 to further discuss.

PR #2801, opened 2024-05-05
Minor change in create pod 

aritraghosh:
changing the run command to use the API version. The current version doesn't create the right pod name
`kubectl run test-pod --image=busybox:latest -- sleep inf
kubectl run --generator=deployment/apps.v1 is DEPRECATED and will be removed in a future version. Use kubectl run --generator=run-pod/v1 or kubectl create instead.`

## How to use

## Testing done

### DIFF
--- a/docs/builtin-gadgets/top/block-io.md
+++ b/docs/builtin-gadgets/top/block-io.md
@@ -13,7 +13,7 @@ the most block device input/output.
 First, we need to create one pod for us to play with:
 
 ```bash
-$ kubectl run test-pod --image busybox:latest sleep inf
+$ kubectl run  --generator=run-pod/v1 test-pod --image busybox:latest sleep inf
 ```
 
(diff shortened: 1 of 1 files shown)


blanquicet:
Hi @aritraghosh, glad to see you here. Welcome! 🙂

What `kubectl` version are you using? I think you are using an old version because since version [1.18.0](https://github.com/kubernetes/kubernetes/pull/87077), it's guaranteed that only pods would be created by `kubectl run`, thus the `--generator` flag has no effect. In fact, it was fully removed in version [1.21.0](https://github.com/kubernetes/kubernetes/pull/99732).

Now, considering the latest stable version is `1.30.0`, I think this would break most environments. WDYT?

aritraghosh:
Good catch, I was on 1.16. I upgraded to 1.3 now and it works. Closing the PR!

PR #2809, opened 2024-05-06
UserGroupCache: Move OnceValue to global scope

burak-ok:
Creating the OnceValue inside the function and then calling it doesn't create the desired effect. It just creates new caches all the time.

Therefore move the creation of the `OnceValue` to the global scope.


mauriciovasquezbernal (approved):
Good catch! What about adding a "Fixes" tag so we can backport it in our next bug fix release?

PR #2814, opened 2024-05-07
trace/network: Close Operators after closing the tracer

burak-ok:
# trace/network: Close Operators after closing the tracer

The `eventCallback` is still called multiple times after (some) operators were closed.

It might finally solve our flaky `TestAdviseNetworkpolicy` test in our CI.

Testing was done through calling the command 15-25 times manually :) I reproduced the issue on an AKS cluster. With minikube it always seemed to work before this PR. Now everything looks like it works


mauriciovasquezbernal (approved):
LGTM from code inspection. Let's hope it fixes the flaky test!

PR #2815, opened 2024-05-07
kubeinvcache: Create the cachedmaps on demand

burak-ok:
# kubeinvcache: Create the cachedmaps on demand

https://github.com/inspektor-gadget/inspektor-gadget/pull/2751 made the resource cache more general and modified it so that pruning old objects is done periodically in the background with the help of a timer.
That means if the `CachedMap` is allocated, there is a goroutine running in the background for it.

We do not need that if the `KubeInventoryCache` is not running. Therefore this PR creates these maps on demand and closes them accordingly

### DIFF
--- a/pkg/cachedmap/cachedmap.go
+++ b/pkg/cachedmap/cachedmap.go
@@ -62,6 +62,7 @@ func NewCachedMap[Key comparable, T any](oldEntryTTL time.Duration) CachedMap[Ke
 
 func (c *cachedMap[Key, T]) Close() {
 	close(c.exit)
+	c.Clear()
 }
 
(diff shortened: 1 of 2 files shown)


alban (approved):
LGTM.

I wonder if `cachedMap.pruneLoop()` should cleanup the `current` and `old` maps when  it receives the `c.exit` signal, but that's unrelated to this PR.

burak-ok:
> I wonder if `cachedMap.pruneLoop()` should cleanup the `current` and `old` maps when it receives the `c.exit` signal, but that's unrelated to this PR.

That is a quick win. Instead of doing it in the `pruneLoop()` I cleared the maps in the `Close()` function

PR #2823, opened 2024-05-08
datasources: Add missing setters, allocate memory and add tests

mauriciovasquezbernal:
See individual commits for more details.

### DIFF
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -138,9 +138,28 @@ func (ds *dataSource) NewData() Data {
 	d := &data{
 		Payload: make([][]byte, ds.payloadCount),
 	}
-	for i := range d.Payload {
-		d.Payload[i] = make([]byte, 0)
+
+	// Allocate memory for fixed size fields added with Add{Sub}Field
+	for _, f := range ds.fields {
+		// Skip all fields that don't need memory allocated: empty, static
+		// members and containers
+		if FieldFlagEmpty.In(f.Flags) || FieldFlagStaticMember.In(f.Flags) ||
+			FieldFlagContainer.In(f.Flags) {
+			continue
+		}
+
+		switch f.Kind {
+		case api.Kind_Bool, api.Kind_Int8, api.Kind_Uint8:
+			d.Payload[f.PayloadIndex] = make([]byte, 1)
+		case api.Kind_Int16, api.Kind_Uint16:
+			d.Payload[f.PayloadIndex] = make([]byte, 2)
+		case api.Kind_Int32, api.Kind_Uint32, api.Kind_Float32:
+			d.Payload[f.PayloadIndex] = make([]byte, 4)
+		case api.Kind_Int64, api.Kind_Uint64, api.Kind_Float64:
+			d.Payload[f.PayloadIndex] = make([]byte, 8)
+		}
 	}
+
 	return d
 }
 
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"sync"
 
+	"google.golang.org/protobuf/proto"
+
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 )
 
@@ -140,6 +142,19 @@ func (ds *dataSource) NewData() Data {
 		d.Payload[i] = make([]byte, 0)
 	}
 	return d
+}
+
+func (ds *dataSource) NewDataFromRaw(b []byte) (Data, error) {
+	data := &data{}
+	err := proto.Unmarshal(b, data.Raw())
+	if err != nil {
+		return nil, fmt.Errorf("unmarshaling payload: %w", err)
+	}
+
+	// TODO: check that size fields matches the size of the fields in the
+	// DataSource
+
+	return data, nil
 }
 
 func (ds *dataSource) ByteOrder() binary.ByteOrder {
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -69,7 +69,6 @@ func (f *field) ReflectType() reflect.Type {
 
 type dataSource struct {
 	name string
-	id   uint32
 
 	dType Type
 
--- a/pkg/datasource/accessors.go
+++ b/pkg/datasource/accessors.go
@@ -98,6 +98,9 @@ type FieldAccessor interface {
 	PutInt16(Data, int16)
 	PutInt32(Data, int32)
 	PutInt64(Data, int64)
+	PutFloat32(Data, float32)
+	PutFloat64(Data, float64)
+	PutString(Data, string)
 	PutBytes(Data, []byte)
 	PutBool(Data, bool)
 }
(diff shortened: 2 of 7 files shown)


pkg/datasource/data.go
```diff
@@ -137,9 +136,25 @@ func (ds *dataSource) NewData() Data {
 	d := &data{
 		Payload: make([][]byte, ds.payloadCount),
 	}
-	for i := range d.Payload {
-		d.Payload[i] = make([]byte, 0)
+
+	// Allocate memory for fixed size fields added with Add{Sub}Field
```
flyth, line 140: Maybe it's time to re-add the `sync.Pool` and then just reset the contents on the individual fields after getting an object from the pool. Can be done in another PR, though.

Also, we should at least check the sizes when receiving data from gRPC to give the same write guarantees on the client - or drop the packet (can also be done on a later PR).

flyth (approved):
Looks good from code inspection!

Thanks also for adding all those shiny tests!

pkg/datasource/data.go
mauriciovasquezbernal: It turns out that this PR broke the gadgectl / kubectl-gadget cases, an invalid memory access panic happens on `NewData()` because `PayloadIndex` is not set on the client, (it's not carried on the protobuf message), then the `d.Payload[f.PayloadIndex] = make([]byte, foo)` line crashes. 

I propose a solution to this in the last commit I pushed. WDYT?

mauriciovasquezbernal:
I also added some additional logic to avoid the setters crashing. Related to https://github.com/inspektor-gadget/inspektor-gadget/pull/2818#discussion_r1594012448.

pkg/datasource/data.go
flyth: > because PayloadIndex is not set on the client, (it's not carried on the protobuf message)

Hmm, I fail to see how/why it's not there as it's set and also part of the protobuf message. It should always be present unless `FieldFlagEmpty` is set, in which case the field doesn't have its own payload. I think it's important to fix it if that's not the case. I'll take a closer look later on again.

What I can see missing from the protobuf message is the `payloadCount` - when deserializing the DataSource info, it should be set to the expected payloadCount - as that is also necessary for adding new fields in operators on the client.

(I'm fine with having separate functions, though.)

mauriciovasquezbernal: > Hmm, I fail to see how/why it's not there as it's set and also part of the protobuf message. It should always be present unless FieldFlagEmpty is set, in which case the field doesn't have its own payload. I think it's important to fix it if that's not the case. I'll take a closer look later on again.

> What I can see missing from the protobuf message is the payloadCount - when deserializing the DataSource info, it should be set to the expected payloadCount - as that is also necessary for adding new fields in operators on the client.

I wrote this too fast. It's the `payloadCount` member on the datasource that is not set, hence the `Payload: make([][]byte, ds.payloadCount)` produces an empty payload, then the `d.Payload[f.PayloadIndex] = make([]byte, foo)` crashes as `d.Payload[f.PayloadIndex]` is an invalid memory access.

I agree we could carry this field over, but I think we need to use a different function to unmarshal from protobuf, otherwise all those allocations will be useless.

blanquicet: > I propose a solution to this in the last commit I pushed. WDYT?
> 
It looks fine to me.

> I think we need to use a different function to unmarshal from protobuf, otherwise all those allocations will be useless.
>
I agree with the dedicated function you created for unmarshal but why do you say that the allocation will be useless if we don't pre-allocate things? While I do see the advantage of pre-allocating things on the emitter (we make the API cleaner than [this](https://github.com/inspektor-gadget/inspektor-gadget/blob/c35c13c7dbfaa0f0b2781cf38d4f5c8a4af17a09/pkg/operators/ebpf/snapshotter.go#L194-L196)), I don't understand what's the advantage of pre-allocating things on the receiver if `proto.Unmarshal` will do that for us. What am I missing?

Regarding the `payloadCount`, I think that sending it from the server gives us a way to ensure we received the expected number of rows in the payload. Don't you think?

blanquicet (commented):
LGTM from code inspection.

blanquicet (approved):
😝

pkg/datasource/data.go
flyth: > Regarding the payloadCount, I think that sending it from the server gives us a way to ensure we received the expected number of rows in the payload. Don't you think?

This wouldn't ensure it (remote end could just lie about it) - we always have to check anyway. The DataSource info from the GadgetInfo is basically the contract that we have to check everything against. The `payloadCount` can be derived from that, so if we have to check for that anyway, we don't need to send it.


> I agree with the dedicated function you created for unmarshal but why do you say that the allocation will be useless if we don't pre-allocate things? While I do see the advantage of pre-allocating things on the emitter (we make the API cleaner than [this](https://github.com/inspektor-gadget/inspektor-gadget/blob/c35c13c7dbfaa0f0b2781cf38d4f5c8a4af17a09/pkg/operators/ebpf/snapshotter.go#L194-L196)), I don't understand what's the advantage of pre-allocating things on the receiver if proto.Unmarshal will do that for us. What am I missing?

You're probably talking about the same: when the event is coming from gRPC, it makes no sense to pre-allocate as gRPC will allocate anyway - hence the two different functions to initialize. This second deserialize function still has to check lengths against the "contract" to allow that data to pass through, otherwise things could crash.

pkg/datasource/data.go
```diff
@@ -137,12 +138,44 @@ func (ds *dataSource) NewData() Data {
...
+	for _, f := range ds.fields {
+		// Skip all fields that don't need memory allocated: empty, static
+		// members and containers
+		if FieldFlagEmpty.In(f.Flags) || FieldFlagStaticMember.In(f.Flags) ||
+			FieldFlagContainer.In(f.Flags) {
```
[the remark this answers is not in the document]
flyth: Also here: the remote end could lie about it and fake a `FIeldFlagContainer` with a non-Kind_Invalid field, which would then crash. So we should never assume, we always have to check here to be sure. (Not sure if there are already other checks in place that _could_ catch it, but still, this needs to be very clean)

Three of those exceptions still look confusing, maybe we can simplify that after we're done with the array support and these changes.

pkg/datasource/data.go
```diff
@@ -137,12 +138,44 @@ func (ds *dataSource) NewData() Data {
 	d := &data{
 		Payload: make([][]byte, ds.payloadCount),
 	}
-	for i := range d.Payload {
-		d.Payload[i] = make([]byte, 0)
+
+	// Allocate memory for fixed size fields added with Add{Sub}Field
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I prefer to be as explicit (and over cautious) as possible here.

pkg/datasource/data.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I prefer to be as explicit (and over cautious) as possible here.

pkg/datasource/data.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Fields added with that function don't need memory to be preallocated as the setter will set the entire buffer in https://github.com/inspektor-gadget/inspektor-gadget/blob/78a0153a6786d839bbf02cc047fd64ed4c7e5f9b/pkg/datasource/accessors.go#L190 to avoid copying the memory. 

I suppose calling `Set` in a member of a "static field container" without first calling `Set()` on the parent is a bad behavior and something we don't need to protect against. 

All those members have the FieldFlagStaticMember flag set 

https://github.com/inspektor-gadget/inspektor-gadget/blob/309f14ea56a07ccacbde833f9eb988bd23333501/pkg/datasource/data.go#L190

and the parent as FieldFlagContainer

https://github.com/inspektor-gadget/inspektor-gadget/blob/309f14ea56a07ccacbde833f9eb988bd23333501/pkg/datasource/data.go#L253

PR #2827, opened 2024-05-09
gadgets: Add trace_lsm

i-Pear:
# gadgets: Add trace_lsm

This gadget is used as a "strace" for LSM tracepoints.

## How to use

```
ig run --verify-image=false trace_lsm:latest --help  # list all supported LSM tracepoints
ig run --verify-image=false trace_lsm:latest --trace_inode_mkdir  # trace "inode_mkdir" tracepoint
ig run --verify-image=false trace_lsm:latest --trace_all  # trace all supported LSM tracepoints
```

### DIFF
--- a/gadgets/trace_lsm/artifacthub-pkg.yml
+++ b/gadgets/trace_lsm/artifacthub-pkg.yml
@@ -1,0 +1,28 @@
+# Artifact Hub package metadata file
+version: 0.28.1
+name: "trace lsm"
+category: monitoring-logging
+displayName: "trace lsm"
+createdAt: "2024-03-25T14:32:05+01:00"
+description: "a strace for LSM tracepoints"
+logoURL: "https://inspektor-gadget.io/media/brand-icon.svg"
+license: ""
+homeURL: "https://inspektor-gadget.io/"
+containersImages:
+    - name: gadget
+      image: "ghcr.io/inspektor-gadget/gadget/trace_lsm:latest"
+      platforms:
+        - linux/amd64
+        - linux/arm64
+keywords:
+    - gadget
+links:
+    - name: source
+      url: "https://github.com/inspektor-gadget/inspektor-gadget/"
+install: |
+    # Run
+    ```bash
+    sudo IG_EXPERIMENTAL=true ig run ghcr.io/inspektor-gadget/gadget/trace_lsm:latest
+    ```
+provider:
+    name: Inspektor Gadget
--- a/gadgets/trace_lsm/gadget.yaml
+++ b/gadgets/trace_lsm/gadget.yaml
@@ -1,0 +1,634 @@
+name: trace lsm
+description: a strace for LSM tracepoints
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://inspektor-gadget.io/docs
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+tracers:
+  lsm:
+    mapName: events
+    structName: event
+structs:
+  event:
+    fields:
+    - name: mntns_id
+      description: Mount namespace inode id
+      attributes:
+        template: ns
+    - name: pid
+      attributes:
+        template: pid
+    - name: tid
+      attributes:
+        template: pid
+    - name: uid
+      attributes:
+        template: uid
+    - name: gid
+      attributes:
+        template: uid
+    - name: tracepoint
+      description: lsm tracepoint enum
+      attributes:
+        width: 16
+        alignment: left
+        ellipsis: end
+    - name: comm
+      description: command
+      attributes:
+        template: comm
+    - name: timestamp
+      attributes:
+        template: timestamp
+ebpfParams:
+  trace_all:
+    key: trace-all
+    defaultValue: "false"
+    description: trace all LSM tracepoints
+  trace_binder_set_context_mgr:
+    key: trace-binder_set_context_mgr
+    defaultValue: "false"
+    description: Trace the binder_set_context_mgr LSM hook
+  trace_binder_transaction:
+    key: trace-binder_transaction
+    defaultValue: "false"
+    description: Trace the binder_transaction LSM hook
+  trace_binder_transfer_binder:
+    key: trace-binder_transfer_binder
+    defaultValue: "false"
+    description: Trace the binder_transfer_binder LSM hook
+  trace_binder_transfer_file:
+    key: trace-binder_transfer_file
+    defaultValue: "false"
+    description: Trace the binder_transfer_file LSM hook
+  trace_bprm_check_security:
+    key: trace-bprm_check_security
+    defaultValue: "false"
+    description: Trace the bprm_check_security LSM hook
+  trace_bprm_committed_creds:
+    key: trace-bprm_committed_creds
+    defaultValue: "false"
+    description: Trace the bprm_committed_creds LSM hook
+  trace_bprm_committing_creds:
+    key: trace-bprm_committing_creds
+    defaultValue: "false"
+    description: Trace the bprm_committing_creds LSM hook
+  trace_bprm_creds_for_exec:
+    key: trace-bprm_creds_for_exec
+    defaultValue: "false"
+    description: Trace the bprm_creds_for_exec LSM hook
+  trace_bprm_creds_from_file:
+    key: trace-bprm_creds_from_file
+    defaultValue: "false"
+    description: Trace the bprm_creds_from_file LSM hook
+  trace_capable:
+    key: trace-capable
+    defaultValue: "false"
+    description: Trace the capable LSM hook
+  trace_capget:
+    key: trace-capget
+    defaultValue: "false"
+    description: Trace the capget LSM hook
+  trace_capset:
+    key: trace-capset
+    defaultValue: "false"
+    description: Trace the capset LSM hook
+  trace_cred_alloc_blank:
+    key: trace-cred_alloc_blank
+    defaultValue: "false"
+    description: Trace the cred_alloc_blank LSM hook
+  trace_cred_free:
+    key: trace-cred_free
+    defaultValue: "false"
+    description: Trace the cred_free LSM hook
+  trace_cred_getsecid:
+    key: trace-cred_getsecid
+    defaultValue: "false"
+    description: Trace the cred_getsecid LSM hook
+  trace_cred_prepare:
+    key: trace-cred_prepare
+    defaultValue: "false"
+    description: Trace the cred_prepare LSM hook
+  trace_cred_transfer:
+    key: trace-cred_transfer
+    defaultValue: "false"
+    description: Trace the cred_transfer LSM hook
+  trace_d_instantiate:
+    key: trace-d_instantiate
+    defaultValue: "false"
+    description: Trace the d_instantiate LSM hook
+  trace_dentry_create_files_as:
+    key: trace-dentry_create_files_as
+    defaultValue: "false"
+    description: Trace the dentry_create_files_as LSM hook
+  trace_dentry_init_security:
+    key: trace-dentry_init_security
+    defaultValue: "false"
+    description: Trace the dentry_init_security LSM hook
+  trace_file_alloc_security:
+    key: trace-file_alloc_security
+    defaultValue: "false"
+    description: Trace the file_alloc_security LSM hook
+  trace_file_fcntl:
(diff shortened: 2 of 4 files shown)


gadgets/trace_lsm/program.bpf.c
```diff
@@ -0,0 +1,222 @@
...
+	F(shm_shmctl)               \
+	F(shm_shmat)                \
+	F(sem_alloc_security)       \
+	F(sem_free_security)        \
+	F(sem_associate)            \
+	F(sem_semctl)               \
+	F(sem_semop)                \
+	F(netlink_send)             \
+	F(d_instantiate)            \
+	F(getselfattr)              \
```
alban, line 159: On Linux 6.6.13-100.fc38.x86_64 I get:
```
program trace_lsm_getselfattr: attach LSM/LSMMac: getselfattr LSM hook not supported
```

This LSM hook was added in Linux 6.8:
https://github.com/torvalds/linux/commit/a04a1198088a1378d0389c250cc684f649bcc91e

For this PR, could you remove the hooks that were added recently, so this gadget works on most kernels?

---

I wonder how to make the user experience better (in another PR). What about a flag "besteffort":
```
SEC("lsm/getselfattr/besteffort")
```
And when defined, ig would ignore the error when the LSM hook is not supported in the current kernel.
In this way, the gadget author would decide if it's better to fail or not. Or maybe that should be something defined in gadget.yaml, so the ELF conventions don't diverge between libbpf, cilium/ebpf or ig. Alternatively, it could be something that the WASM module decide based on the BTF data, e.g. in `union security_list_options` (cc @mauriciovasquezbernal ). What do you think?

gadgets/trace_lsm/program.bpf.c
```diff
@@ -0,0 +1,222 @@
...
+	GADGET_PARAM(trace_##name);
+
+const volatile bool trace_all = false;
+GADGET_PARAM(trace_all);
+
+FOR_EACH_LSM_HOOK(DECLARE_LSM_PARAMETER)
+
+#define TRACE_LSM(name)                                                \
+	SEC("lsm/" #name)                                              \
+	int trace_lsm_##name()                                         \
```
alban, line 199: To be really useful, this gadget should access parameters.

And when it is a well-known type such as `struct path *`  or `struct file *`, print the full path using `include/gadget/filesystem.h`.

I wonder if type matches relocations can help (https://github.com/inspektor-gadget/inspektor-gadget/pull/2542).

But that's for another PR.

alban (changes requested):
Thanks. I commented with some small things to change, and some ideas for future PRs.

gadgets/trace_lsm/program.bpf.c
i-Pear: The `besteffort` flag seems to be a general option. Maybe we can create a new macro `SEC_BESTEFFORT`, and it will add strings `[besteffort]` in the end of section names. We will identify and erase it while attaching.

i-Pear: uprobe/libc:malloc?failstrategy=ignore&option2=foo2

i-Pear: I'm wrong, the `cillum/ebpf` library will verify the section name while creating ebpf collections ([NewCollectionWithOptions](https://github.com/cilium/ebpf/blob/main/collection.go#L366)), by finding the name in kernel [findProgramTargetInKernel](https://github.com/cilium/ebpf/blob/main/prog.go#L1011). So if we add `?para&para2` in the end of section names, the ebpf collection will fail to load.

i-Pear: For older kernels, the `attach LSM/LSMMac: getselfattr LSM hook not supported` error message also comes from `NewCollectionWithOptions`, not our attaching process. So to support the `besteffort` flag, we may also need to hack `cillum/ebpf`.

i-Pear: Issue in cillium/ebpf: [#https://github.com/cilium/ebpf/discussions/1470](https://github.com/cilium/ebpf/discussions/1470)

gadgets/trace_lsm/program.bpf.c
i-Pear: > I wonder if type matches relocations can help (https://github.com/inspektor-gadget/inspektor-gadget/pull/2542).

Do you mean we can use `bpf_core_type_matches` to find `struct path *` types automatically?

gadgets/trace_lsm/artifacthub-pkg.yml
alban, line 1: It would be good to have a README.md to improve the artifacthub page.

This could be done in another PR.

gadgets/trace_lsm/gadget.yaml
```diff
@@ -0,0 +1,638 @@
...
+    - name: comm
+      description: command
+      attributes:
+        template: comm
+    - name: timestamp
+      attributes:
+        template: timestamp
+ebpfParams:
+  trace_all:
+    key: trace_all
```
alban, line 44:
```suggestion
    key: trace-all
```

Other gadgets use a `-` for the CLI flag:
```bash
$ git grep 'key: .*[-_].*' -- gadgets/
gadgets/top_file/gadget.yaml:58:    key: all-files
gadgets/trace_exec/gadget.yaml:84:    key: ignore-failed
gadgets/trace_signal/gadget.yaml:51:    key: kill-only
```

gadgets/trace_lsm/gadget.yaml
```diff
@@ -0,0 +1,638 @@
...
+    defaultValue: ""
+    description: trace all LSM tracepoints
+  trace_binder_set_context_mgr:
+    key: trace_binder_set_context_mgr
```
alban, line 48: I suggest to use `trace-%s` for the CLI flags:
```suggestion
    key: trace-binder_set_context_mgr
```

Ditto below.

gadgets/trace_lsm/gadget.yaml
```diff
@@ -0,0 +1,638 @@
...
+    defaultValue: ""
+    description: ""
```
alban, line 50:
```suggestion
    defaultValue: "false"
    description: "Trace the binder_set_context_mgr LSM hook"
```

gadgets/trace_lsm/program.bpf.c
alban: I was thinking of something like this:

```c
#define TRACE_LSM(name)                                                        \
	union security_list_options___##name##arg1 {                           \
	    int (* #name)(const struct path *);                                \
	};                                                                     \
	union security_list_options___##name##arg2 {                           \
	    int (* #name)(const struct path *, const struct path *);           \
	};                                                                     \
	                                                                       \
	/* ignore attaching failures for compatibility */                      \
	SEC("lsm/" #name)                                                      \
	int trace_lsm_##name(void *arg1, void *arg2, void *arg3)               \
            struct event *event; /*TODO: alloc event in a map */               \
            char *path_str = NULL;                                             \
            struct path *path = NULL;                                          \
            if (bpf_core_type_matches(security_list_options___##name##arg1)) { \
                path = (struct path *) arg1;                                   \
            }                                                                  \
            if (bpf_core_type_matches(security_list_options___##name##arg2)) { \
                path = (struct path *) arg1;                                   \
            }                                                                  \
            if (path) {                                                        \
                path_str = get_path_str(&fs->pwd);                             \
            }                                                                  \
            if (path_str) {                                                    \
                bpf_probe_read_kernel_str(event->path, MAX_STRING_SIZE, path_str); \
            }                                                                  \
            ...
```

But that would work only for hooks with a single `struct path *` argument (for the arg1 test):
* path_truncate
* path_chroot
* inode_getattr
Or with 2 `struct path *` arguments:
* sb_pivotroot
* move_mount

But others would be missed, so I don't know if it is worth it.

In any cases, it is not for this PR.

alban (commented):
Thanks!

gadgets/trace_lsm/artifacthub-pkg.yml
i-Pear: Currently I don't have a k8s environment, will provide readme files for all my gadgets later.

gadgets/trace_lsm/program.bpf.c
i-Pear: I haven't much understood the code here, are there structs like `security_list_options___*` in kernel?
We can discuss about this tomorrow.

i-Pear: Looking into this: https://lwn.net/Articles/967893/

i-Pear: I finally gave up after trying for several hours, I couldn't find any CORE API that can be used to match parameter types.

gadgets/trace_lsm/program.bpf.c
```diff
@@ -0,0 +1,228 @@
...
+	F(inode_listsecurity)       \
+	F(inode_getsecid)           \
+	F(inode_copy_up)            \
+	F(inode_copy_up_xattr)      \
+	F(kernfs_init_security)     \
+	F(file_permission)          \
+	F(file_alloc_security)      \
+	F(file_free_security)       \
+	F(file_ioctl)               \
+	F(file_ioctl_compat)        \
```
alban, line 92: With `6.6.13-100.fc38.x86_64`, I get this error:
> Error: starting operator "oci": starting operator "ebpf": creating eBPF collection: program trace_lsm_file_ioctl_compat: attach LSM/LSMMac: file_ioctl_compat LSM hook not supported


After removing it, it works for me:
```diff
diff --git a/gadgets/trace_lsm/gadget.yaml b/gadgets/trace_lsm/gadget.yaml
index aac627a0..cde8c174 100644
--- a/gadgets/trace_lsm/gadget.yaml
+++ b/gadgets/trace_lsm/gadget.yaml
@@ -140,10 +140,6 @@ ebpfParams:
     key: trace-file_ioctl
     defaultValue: "false"
     description: Trace the file_ioctl LSM hook
-  trace_file_ioctl_compat:
-    key: trace-file_ioctl_compat
-    defaultValue: "false"
-    description: Trace the file_ioctl_compat LSM hook
   trace_file_lock:
     key: trace-file_lock
     defaultValue: "false"
diff --git a/gadgets/trace_lsm/program.bpf.c b/gadgets/trace_lsm/program.bpf.c
index 84fef316..771110f5 100644
--- a/gadgets/trace_lsm/program.bpf.c
+++ b/gadgets/trace_lsm/program.bpf.c
@@ -89,7 +89,6 @@
        F(file_alloc_security)      \
        F(file_free_security)       \
        F(file_ioctl)               \
-       F(file_ioctl_compat)        \
        F(mmap_addr)                \
        F(mmap_file)                \
        F(file_mprotect)            \
```

gadgets/trace_lsm/program.bpf.c
```diff
@@ -0,0 +1,228 @@
...
+GADGET_TRACER_MAP(events, 1024 * 256);
+
+GADGET_TRACER(lsm, events, event);
+
+#define DECLARE_LSM_PARAMETER(name)               \
+	const volatile bool trace_##name = false; \
+	GADGET_PARAM(trace_##name);
+
+const volatile bool trace_all = false;
+GADGET_PARAM(trace_all);
```
alban, line 190: I wonder if this should be true by default. Otherwise, when running the gadget without CLI flags, it will do nothing.

Maybe that will just be something to explain in the README.md later on.

gadgets/trace_lsm/gadget.yaml
```diff
@@ -0,0 +1,638 @@
...
+        template: pid
+    - name: tid
+      attributes:
+        template: pid
+    - name: uid
+      attributes:
+        template: pid
+    - name: gid
+      attributes:
+        template: pid
```
alban, line 28:
```suggestion
        template: uid
```

gadgets/trace_lsm/gadget.yaml
```diff
@@ -0,0 +1,638 @@
...
+        template: ns
+    - name: pid
+      attributes:
+        template: pid
+    - name: tid
+      attributes:
+        template: pid
+    - name: uid
+      attributes:
+        template: pid
```
alban, line 25:
```suggestion
        template: uid
```

alban (commented):
Some small changes remaining to do. It should be good to go after that. Thanks!

alban (approved):

PR #2828, opened 2024-05-09
oci: Fail is start image operator fails

mauriciovasquezbernal:
Make the whole operation fail is starting an image operator fails.

### Testing

Create a gadget that doesn't pass the verifier:

```c
#include <vmlinux.h>
#include <bpf/bpf_helpers.h>
#include <gadget/buffer.h>
#include <gadget/macros.h>
#include <gadget/mntns_filter.h>
#include <gadget/types.h>

#define NAME_MAX 255

struct event {
	gadget_mntns_id mntns_id;
	__u32 pid;
	char comm[TASK_COMM_LEN];
	char filename[NAME_MAX];
};

// events is the name of the buffer map and 1024 * 256 is its size.

// [Optional] Define a tracer
GADGET_TRACER(open, events, event);

int enter_openat(struct syscall_trace_enter *ctx)
{
	struct event *event;
	u64 mntns_id;

	mntns_id = gadget_get_mntns_id();
	if (gadget_should_discard_mntns_id(mntns_id))
		return 0;

	event = gadget_reserve_buf(&events, sizeof(*event));
	if (!event)
		return 0;

	event->mntns_id = mntns_id;

	bpf_get_current_comm(event->comm, sizeof(event->comm));
	bpf_probe_read_user_str(event->filename, sizeof(event->filename), (const char *)ctx->args[1]);

	//gadget_submit_buf(ctx, &events, event, sizeof(*event));

	return 0;
}

```

```bash 
$ sudo -E ig image build -t badgadget 
```

### Before this PR

Error is printed but program keeps running until Ctrl+C is hit. 

```bash
$ sudo -E ig run badgadget --verify-image=false
WARN[0000] you set --verify-image=false, image will not be verified
WARN[0000] you set --verify-image=false, image will not be verified
RUNTIME.CONTAINERNAME        MNTNS_ID            PID             COMM            FILENAME
ERRO[0000] starting operator "ebpf": creating eBPF collection: program enter_openat: load program: invalid argument: Unreleased reference id=4 alloc_insn=43 (85 line(s) omitted)


^C$ echo $?
0
```

### After this PR

Error is printed and ig process exits with an error 

```bash 
$ sudo -E ig run badgadget --verify-image=false
WARN[0000] you set --verify-image=false, image will not be verified
WARN[0000] you set --verify-image=false, image will not be verified
RUNTIME.CONTAINERNAME        MNTNS_ID            PID             COMM            FILENAME
Error: starting operator "oci": starting operator "ebpf": creating eBPF collection: program enter_openat: load program: invalid argument: Unreleased reference id=4 alloc_insn=43 (85 line(s) omitted)

$ echo $?
1
```

### DIFF
--- a/pkg/operators/oci-handler/oci.go
+++ b/pkg/operators/oci-handler/oci.go
@@ -266,11 +266,20 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
 }
 
 func (o *OciHandlerInstance) Start(gadgetCtx operators.GadgetContext) error {
+	started := []operators.ImageOperatorInstance{}
+
 	for _, opInst := range o.imageOperatorInstances {
 		err := opInst.Start(o.gadgetCtx)
 		if err != nil {
-			o.gadgetCtx.Logger().Errorf("starting operator %q: %v", opInst.Name(), err)
+			// Stop all started operators
+			for _, startedOp := range started {
+				startedOp.Stop(o.gadgetCtx)
+			}
+
+			return fmt.Errorf("starting operator %q: %w", opInst.Name(), err)
 		}
+
+		started = append(started, opInst)
 	}
 	return nil
 }


pkg/operators/oci-handler/oci.go
```diff
@@ -266,11 +266,20 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
 }
 
 func (o *OciHandlerInstance) Start(gadgetCtx operators.GadgetContext) error {
+	started := []operators.DataOperatorInstance{}
```
blanquicet, line 269:
```suggestion
	started := []operators.ImageOperatorInstance{}
```

pkg/operators/oci-handler/oci.go
```diff
@@ -266,11 +266,20 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
...
+
 	for _, opInst := range o.imageOperatorInstances {
 		err := opInst.Start(o.gadgetCtx)
 		if err != nil {
-			o.gadgetCtx.Logger().Errorf("starting operator %q: %v", opInst.Name(), err)
+			// Stop all started operators
+			for _, startedOp := range started {
+				startedOp.Stop(o.gadgetCtx)
```
blanquicet, line 276: Actually, why not to call `o.Stop()`. The `ImageOperatorInstance.Stop` implementations should be robust enough to do nothing if they haven't been initialized, don't you think?

pkg/operators/oci-handler/oci.go
mauriciovasquezbernal: Good catch! The compiler didn't say anything!

pkg/operators/oci-handler/oci.go
mauriciovasquezbernal: I know we have been doing this for the gadgets and so on but IMO there is not any reason why Stop() should be called on something that hasn't been started. It doesn't make sense and just complicates the implementation of those operators.

pkg/operators/oci-handler/oci.go
blanquicet: It's because the DataOperatorInstance implements a subset of ImageOperatorInstance's methods. The other way around wouldn't have worked 😅

pkg/operators/oci-handler/oci.go
blanquicet: Calling `Stop` complicates the logic of operators (They have to check nil before calling Close, Stop, etc ... Already done by the ebpf operator and most operators). Stopping only the started operators complicate the logic of the caller (You had the created a temporary slice to know which operators were started). I think it's a matter of decide where we want to put the "complexity" 😅. I'm ok either way, we just need to be consistent and use the same approach everywhere from now on.

blanquicet (approved):
LGTM!

pkg/operators/oci-handler/oci.go
mauriciovasquezbernal: It's better to complicate the logic in one place (oci operator) than in many (all image operator), specially given that users of the API could implement their own image operators.

PR #2832, opened 2024-05-10
datasource: Allow setting smaller static strings

mauriciovasquezbernal:
There are cases when an operator needs to change the value of a static string (coming from ebpf for instance). This commit ipt improves the logic to allow the operator calling the PutString() method with a string that is smaller than the buffer size.

---

I found this issue / limitation when trying to update the name field in wasm in #2823. That PR implements a workaround for it, but IMO it needs to be handled directly in the accessor.

@flyth any opinions on this one?

### DIFF
--- a/pkg/datasource/accessors.go
+++ b/pkg/datasource/accessors.go
@@ -193,6 +193,16 @@ func (a *fieldAccessor) Set(d Data, b []byte) error {
 		return errors.New("field cannot contain a value")
 	}
 	if FieldFlagStaticMember.In(a.f.Flags) {
+		// If it's a short string copy it and clean the rest with 0s
+		if (a.f.Kind == api.Kind_String || a.f.Kind == api.Kind_CString) && uint32(len(b)) < a.f.Size {
+			copy(d.payload()[a.f.PayloadIndex][a.f.Offs:a.f.Offs+a.f.Size], b)
+
+			for i := uint32(len(b)); i < a.f.Size; i++ {
+				d.payload()[a.f.PayloadIndex][a.f.Offs+i] = 0
+			}
+			return nil
+		}
+
 		if uint32(len(b)) != a.f.Size {
 			return invalidFieldLengthErr(len(b), int(a.f.Size))
 		}
--- a/pkg/datasource/accessors.go
+++ b/pkg/datasource/accessors.go
@@ -24,6 +24,22 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 )
 
+type InvalidFieldLengthErr struct {
+	Expected int
+	Actual   int
+}
+
+func (e *InvalidFieldLengthErr) Error() string {
+	return fmt.Sprintf("invalid field length, expected %d, got %d", e.Expected, e.Actual)
+}
+
+func invalidFieldLengthErr(size, expected int) error {
+	return &InvalidFieldLengthErr{
+		Expected: expected,
+		Actual:   size,
+	}
+}
+
 // FieldAccessor grants access to the underlying buffer of a field
 type FieldAccessor interface {
 	Name() string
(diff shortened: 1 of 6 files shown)


flyth:
Hmm...

I think we need to add errors from the `Put*` functions to do these kind of things, otherwise it could lead to (from the developer) unexpected behavior (e.g. some Puts would work, some wouldn't (if the string is too long)). To be fair, it's already the case if you'd try to write a string with the exact length of the field vs. another length right now.

IMHO this isn't best practice in the first place (editing a statically sized string), but if we allow it, we should at least make sure it's reported if it didn't succeed.

mauriciovasquezbernal:
> I think we need to add errors to the Put* functions to do these kind of things, otherwise it could lead to (from the developer) unexpected behavior (e.g. some Puts would work, some wouldn't (if the string is too long)). To be fair, it's already the case if you'd try to write a string with the exact length of the field vs. another length right now.

I totally agree. What do you think about the `"Get()"` methods? Should we return errors instead of default value when something goes wrong?

> IMHO this isn't best practice in the first place (editing a statically sized string), but if we allow it, we should at least make sure it's reported if it didn't succeed.

I don't see any reason why we should block this. I feel there are use cases where this is valuable.

flyth:
> > I think we need to add errors to the Put* functions to do these kind of things, otherwise it could lead to (from the developer) unexpected behavior (e.g. some Puts would work, some wouldn't (if the string is too long)). To be fair, it's already the case if you'd try to write a string with the exact length of the field vs. another length right now.
> 
> I totally agree. What do you think about the `"Get()"` methods? Should we return errors instead of default value when something goes wrong?

Yeah, let's do it - better now than later. I was a bit worried about returning two values, because it might not be compatible with all languages that would compile to WASM - but I think it's better to have a clean implementation on the host and let the guest APIs worry about getting it right there.

> > IMHO this isn't best practice in the first place (editing a statically sized string), but if we allow it, we should at least make sure it's reported if it didn't succeed.
> 
> I don't see any reason why we should block this. I feel there are use cases where this is valuable.

I agree, we shouldn't block this.

mauriciovasquezbernal:
Updated the PR to return errors. The code is not as clean as before, but we need it.

pkg/datasource/accessors.go
```diff
@@ -322,186 +334,203 @@ func (a *fieldAccessor) Annotations() map[string]string {
...
+func (a *fieldAccessor) Float64(data Data) (float64, error) {
+	i, err := a.Uint64(data)
+	if err != nil {
+		return 0.0, err
+	}
+	return math.Float64frombits(i), nil
 }
 
-func (a *fieldAccessor) String(data Data) string {
+func (a *fieldAccessor) String(data Data) (string, error) {
```
flyth, line 417: Can this (and `.Bytes()`) ever fail? But yeah, keeping this consistent is probably better than making exceptions.

pkg/datasource/formatters/json/json.go
```diff
@@ -199,59 +199,71 @@ func (f *Formatter) addSubFields(accessors []datasource.FieldAccessor, prefix st
 		switch accessor.Type() {
 		case api.Kind_Int8:
 			fn = func(e *encodeState, data datasource.Data) {
-				b := strconv.AppendInt(e.scratch[:0], int64(accessor.Int8(data)), 10)
+				v, _ := accessor.Int8(data)
+				b := strconv.AppendInt(e.scratch[:0], int64(v), 10)
```
flyth, line 203: 😢 This was beautiful once.

pkg/datasource/accessors.go
```diff
@@ -322,186 +334,203 @@ func (a *fieldAccessor) Annotations() map[string]string {
...
-	return val[0] == 1
+	return val[0] == 1, nil
 }
 
-func (a *fieldAccessor) PutUint8(data Data, val uint8) {
+func (a *fieldAccessor) PutUint8(data Data, val uint8) error {
 	b := a.Get(data)
 	if len(b) < 1 {
-		return
+		return ErrFieldTooShort
```
flyth, line 439: Do you think there are cases when it makes sense to read a Uint8 from something that's larger than 1 byte without specifying an additional offset? Otherwise maybe it's cleaner to just return `ErrInvalidFieldLength` and check for `len(b) != 1`. Also for the other types.

(Usually I'm not into restricting users from doing weird stuff, but here I don't know...)

flyth (commented):
Thanks for addressing the errors - it's indeed not as clean looking as before, but - well - I guess cleaner working. 🤷

pkg/datasource/accessors.go
mauriciovasquezbernal: So far they'll never fail, but I think having them consistent is better.

pkg/datasource/accessors.go
mauriciovasquezbernal: Added that check as well. At this point, I'm wondering if we should just check the type of the field instead... (but I'll host that discussion for a bit).

flyth (approved):
LGTM, thanks for cleaning this up!

pkg/datasource/accessors.go
```diff
@@ -24,6 +24,22 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 )
 
+type InvalidFieldLengthErr struct {
+	Expected int
+	Actual   int
+}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I took me quite few iterations to reach this! Glad to see you liked it!

PR #2833, opened 2024-05-12
ci: Shifting deploy and undeploy logic for IG to actions

pawarpranav83:
Shifting deploy and undeploy logic from `integration_test_kubectl_gadget.go` to `.github/actions/run-integration-tests/action.yml`.

This change is required since tests for image-based gadgets don't have a TestMain file.

## Testing done
Passing of Integration Tests in the CI.

### DIFF
--- a/.github/actions/run-integration-tests/action.yml
+++ b/.github/actions/run-integration-tests/action.yml
@@ -35,12 +35,20 @@ runs:
       with:
         name: kubectl-gadget-linux-amd64-tar-gz
         path: /home/runner/work/inspektor-gadget/inspektor-gadget/
+    - name: Deploy Inspektor Gadget
+      id: deploy-ig
+      shell: bash
+      run: |
+        tar zxvf /home/runner/work/inspektor-gadget/inspektor-gadget/kubectl-gadget-linux-amd64.tar.gz
+        if [[ ${{ inputs.kubernetes_distribution }} == "minikube-github" ]]; then
+          # Inspektor-Gadget container image is loaded into the cluster using "minikube image load" instead of pushing the container image to a registry.
+          EXTRA_FLAGS='--image-pull-policy=Never'
+        fi
+        ./kubectl-gadget deploy $EXTRA_FLAGS --debug --experimental --image=${{ inputs.container_repo }}:${{ inputs.image_tag }}
     - name: Integration tests
       id: integration-tests
       shell: bash
       run: |
-        tar zxvf /home/runner/work/inspektor-gadget/inspektor-gadget/kubectl-gadget-linux-amd64.tar.gz
-
         cleanup() {
             echo "IntegrationTestsJob: Workflow run is being cancelled: $1 was received"
             trap - $1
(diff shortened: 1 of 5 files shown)


pawarpranav83:
We would also need to change `contributing/testing/integration tests for ig/kubernetes` section, where instead of `minikube-start` we can specify `minikube-deploy` but this only works for docker, otherwise it gives the following error:
```
Exiting due to GUEST_IMAGE_LOAD: save to dir: caching images: caching image "/home/jinx83/.minikube/cache/images/amd64/ghcr.io/inspektor-gadget/inspektor-gadget_latest": write: unable to calculate manifest: blob sha256:b2ce0e0660777651a7a188ae1128acc61d01aca10a035a8b1faa2cdd8bbf0785 not found
```

mauriciovasquezbernal (commented):
Just a minor nit. LGTM otherwise.

mauriciovasquezbernal:
> I am nonetheless not sure about the Makefile part, as we need to have Inspektor Gadget deployed before running the integration tests.

I think we can require Inspektor Gadget to be deployed before running `integration-tests`. We could handle it in the Makefile but IMHO it becomes a mess because we'll need to take care of the undeploy step too. One idea is to have to different targets:

- integration-tests-without-deploy: Just run the tests
- integration-tests: deploy, run integration-tests-without-deploy and then undeploy. 

But again, IMO it's not worth it.

mqasimsarfraz:
> > I did some investigation and it seems this is coming from minikube and isn't related to changes in the PR.
> 
> I'm a bit lost. Is this only related to the "make minikube-deploy" target or also to the CI? I see we use in the CI [`minikube image load`](https://github.com/inspektor-gadget/inspektor-gadget/blob/9e647b4d855eed27fab67c72dc913f6af43c3372/.github/workflows/inspektor-gadget.yml#L1816) for the different runtimes and AFAIU it works fine. What am I missing?

Good old version mismatch! It seems the issue only happens on [Docker Version: 25](https://github.com/kubernetes/minikube/issues/18021#issuecomment-1904869897) and CI seem to be running [Docker Version: 24](https://github.com/actions/runner-images/blob/ubuntu22/20240422.1/images/ubuntu/Ubuntu2204-Readme.md#tools). I also confirmed I am running Docker Version: 25+ on my machine!

mauriciovasquezbernal:
> I do not see the plus-value.

We need this in order to implement tests for image-based gadgets like gadgets/trace_open/test/trace_open_test.go. In those ones we don't have a TestMain where we can put the logic to deploy Inspektor Gadget. 

> Before, we just ran the Makefile and everything was done, now we need to do it in two steps.

Are you really using this target in that way? I think the 99% of the times I use it I pass the `no-deploy-ig` flag.

mauriciovasquezbernal (approved):
LGTM. I'll merge once the CI is green. Thanks for handling this.

mauriciovasquezbernal:
Failing test is another instance of #2358, merging this.

Issue #2821, opened 2024-05-08
trace exec --paths uses too much memory


alban:
## Description

The ebpf map used by the trace exec gadget is defined as follow:
```
struct {
	__uint(type, BPF_MAP_TYPE_HASH);
#ifdef WITH_CWD
	__uint(max_entries, 1024);
#else /* !WITH_CWD */
	__uint(max_entries, 10240);
#endif /* !WITH_CWD */
	__type(key, pid_t);
	__type(value, struct event);
} execs SEC(".maps");
```

It has a different size depending on the define `WITH_CWD` thanks to #2146.

But `WITH_CWD` was removed in favour of `WITH_LONG_PATHS` in #1959. And it seems I forgot some occurrences of `WITH_CWD`.

## Impact

This causes the ebpf map to use a lot of memory.

When ig is used in a pod with the following memory limits, it fails:
```
    resources:
      limits:
        cpu: 150m
        memory: 232Mi
      requests:
        cpu: 50m
        memory: 128Mi
```
with the following error:
```
Error: running gadget: running gadget: installing tracer: loading ebpf spec: loading maps and programs: field IgExecveE: program ig_execve_e: map execs: map create: cannot allocate memory
```

When increasing the pod memory limit to 1232Mi, it works.

## Environment and steps to reproduce

See above.

## Expected behavior

- No dangling `WITH_CWD` references in the source code.
- Reasonable memory usage of the trace exec gadget.

## Additional information

We currently don't set any memory limits in charts/gadget/templates/daemonset.yaml. This bug was found using ig in a custom daemonset (not using the one defined upstream).

alban:
Fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/2840

PR #2146, opened 2023-10-13
trace {open, exec}: Revert map sizes to their previous value.

eiffel-fl:
Hi.

This PR revert map sizes to their previous values of 10240 entries in case the memory hungry options are not used.
Indeed, having only 1024 entries caused vertical scaling problem when run on big machine:

```
francis-ig-vertical@francis-ig-verticalvm:~$ sudo ./ig-bad trace exec -o json -c test-container > /tmp/ig-bad &[1] 2060496
francis-ig-vertical@francis-ig-verticalvm:~$ docker run --user 1000 -ti --rm --name test-container --entrypoint '/bin/sh' -w /tmp ghcr.io/colinianking/stress-ng:master -c 'stress-ng --exec $(nproc) --exec-fork-method clone --exec-method execve --exec-no-pthread --timeout 1s --metrics-brief'
stress-ng: info:  [1] setting to a 1 second run per stressor
stress-ng: info:  [1] dispatching hogs: 96 exec
stress-ng: info:  [1] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s
stress-ng: info:  [1]                           (secs)    (secs)    (secs)   (real time) (usr+sys time)
stress-ng: info:  [1] exec              42213      2.90     34.04    214.31     14567.68         169.98
stress-ng: info:  [1] successful run completed in 2.99s
francis-ig-vertical@francis-ig-verticalvm:~$ wc -l /tmp/ig-bad 
11604 /tmp/ig-bad
francis-ig-vertical@francis-ig-verticalvm:~$ fg
sudo ./ig-bad trace exec -o json -c test-container > /tmp/ig-bad
francis-ig-vertical@francis-ig-verticalvm:~$ sudo ./ig trace exec -o json -c test-container > /tmp/ig &[1] 2103140
francis-ig-vertical@francis-ig-verticalvm:~$ docker run --user 1000 -ti --rm --name test-container --entrypoint '/bin/sh' -w /tmp ghcr.io/colinianking/stress-ng:master -c 'stress-ng --exec $(nproc) --exec-fork-method clone --exec-method execve --exec-no-pthread --timeout 1s --metrics-brief'
stress-ng: info:  [1] setting to a 1 second run per stressor
stress-ng: info:  [1] dispatching hogs: 96 exec
stress-ng: info:  [1] stressor       bogo ops real time  usr time  sys time   bogo ops/s     bogo ops/s
[...]

PR #2840, opened 2024-05-13
trace exec builtin gadget: fix memory usage

alban:
The define WITH_CWD was renamed to WITH_LONG_PATHS. But some occurrences were forgotten.

Fixes: 4f471903d5c9 ("trace exec: add exe path") #1959

See also: 8a4461a5738a ("trace exec: Revert map size to its previous value") #2146

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/2821

## How to use

```
sudo ig trace exec
sudo ig trace exec --paths
```

## Testing done

Before the patch:
```
$ sudo bpftool map show name execs
276219: hash  name execs  flags 0x0
	key 4B  value 15960B  max_entries 10240  memlock 164331040B
	btf_id 28314
	pids ig(1295874)
```
=> 164 MB, whether `--paths` is used or not

After the patch:

```
$ sudo bpftool map show name execs
276266: hash  name execs  flags 0x0
	key 4B  value 7768B  max_entries 10240  memlock 80412192B
	btf_id 28390
	pids ig(1307407)
```
=> 80 MB without `--paths`

```
$ sudo bpftool map show name execs
276287: hash  name execs  flags 0x0
	key 4B  value 15960B  max_entries 1024  memlock 16481824B
	btf_id 28423
	pids ig(1308144)
```
=> 16 MB with `--paths`


alban:
> I am wondering if we cannot increase the number of elements when --paths is used to the point where both flavor have the same memory footprint.

Why not, I don't have strong opinions on this. It still seems huge to me. In comparison with bcc's execsnoop:

https://github.com/iovisor/bcc/blob/master/libbpf-tools/execsnoop.c#L19
```
#define PERF_BUFFER_PAGES   64
```
With 4 cpus, that's `64 * 4096 * 4` = 1 MiB, a lot smaller than our gadget. Do we do something wrong?

> Also, should this be ported to the containerized gadget flavor?

So far, the cwd and exepath columns are not implemented in the image-based gadget and there is no equivalent of `WITH_LONG_PATHS`.

Issue #2371, opened 2024-01-17
[RFE] Socket enricher for gadgets mixing socket filters and tracing ebpf programs


alban:
## Current situation

To use the socket enricher, a gadget has to include:
```
#define GADGET_TYPE_NETWORKING
#include <gadget/sockets-map.h>
```
Or:
```
#define GADGET_TYPE_TRACING
#include <gadget/sockets-map.h>
```
Depending on the gadget type.

See https://inspektor-gadget.io/docs/v0.24.0/reference/gadget-helper-api/#socket-enrichment

## Impact

It's not possible to write a gadget that would use both socket filters and kprobes.

## Ideal future situation

Users can write gadgets containing both kinds of ebpf programs.

## Implementation options

We could have 2 different headers and it should be possible to include them both.

## Additional information

The tcprst gadget might benefit from that: https://github.com/inspektor-gadget/inspektor-gadget/issues/1546

cc @blanquicet

moki1202:
@mauriciovasquezbernal Would simply changing the function names at both these places to `gadget_socket_lookup_networking()` and `gadget_socket_lookup_tracing()` work? I'll update their occurrences accordingly.
```
#ifdef GADGET_TYPE_NETWORKING
static __always_inline struct sockets_value *
gadget_socket_lookup_networking(const struct __sk_buff *skb) 
```

```
#ifdef GADGET_TYPE_TRACING
static __always_inline struct sockets_value *
gadget_socket_lookup_tracing(const struct sock *sk, __u32 netns)
```

alban:
It is fine for me to change the function names to have 2 different names.

But I think that would not "just" compile so easily because the two kinds of gadgets use different kinds of headers:
* tracing programs: use vmlinux.h and kernel internal headers
* networking programs: use uapi kernel headers

I don't think we can use both. Also, I don't know if both set of programs can use the same set of BTF data.

I initially added the "good first issue" label on this issue but now I think that was a bad decision because I don't know how to fix this issue. But you are welcome to give it a try if you want and see if you find solutions.

moki1202:
I'll surf more about it and I'll let you know if I come up with something useful.

alban:
PoC of gadget using both packet capture and uprobes:
https://github.com/alban/trace-udns/pull/1

I couldn't include `<gadget/sockets-map.h>` directly because of duplicate declarations with vmlinux.h but I copied some content over manually.

alban:
I will give this a try.

[...]

PR #2846, opened 2024-05-14
sockets-map.h: remove most defines from socket enricher header

alban:
sockets-map.h can be included in both tracer gadgets (kprobes, etc.) and network gadgets (packet capture). Move most #define to avoid polluting the 'define' namespace of gadgets.

This is a first step to enable gadgets containing both tracers and network programs.

See https://github.com/inspektor-gadget/inspektor-gadget/issues/2371

This is to avoid the manual copy of sockets-map.h as it was done in the PoC https://github.com/alban/trace-udns/pull/1


include/gadget/sockets-map.h
```diff
@@ -3,63 +3,36 @@
...
-			     unsigned long long off) asm("llvm.bpf.load.word");
-
-#endif
+// This file is shared between the networking and tracing programs.
+// Therefore, avoid includes that are specific to one of these types of programs.
+// For example, don't include <linux/ip.h> nor <vmlinux.h> here.
+// Redefine the constants we need but namespaced (SE_) so we don't pollute gadgets.
+
+#define SE_PACKET_HOST 0
+#define SE_PACKET_OUTGOING 4
```
mauriciovasquezbernal, line 12: not used?

include/gadget/sockets-map.h
```diff
@@ -110,7 +82,7 @@ gadget_socket_lookup(const struct __sk_buff *skb)
 
 	switch (h_proto) {
 	case bpf_htons(ETH_P_IP):
```
mauriciovasquezbernal, line 84: SE_ETH_P_IP

include/gadget/sockets-map.h
```diff
@@ -131,13 +103,13 @@ gadget_socket_lookup(const struct __sk_buff *skb)
 		break;
 
 	case bpf_htons(ETH_P_IPV6):
```
mauriciovasquezbernal, line 105: missing namespaced version of this one.

include/gadget/sockets-map.h
```diff
@@ -199,16 +171,16 @@ gadget_socket_lookup(const struct __sk_buff *skb)
...
 	case IPPROTO_TCP:
-		if (skb->pkt_type == PACKET_HOST)
-			off += offsetof(struct tcphdr, dest);
+		if (skb->pkt_type == SE_PACKET_HOST)
+			off += SE_TCPHDR_DEST_OFFSET;
 		else
-			off += offsetof(struct tcphdr, source);
+			off += SE_TCPHDR_SOURCE_OFFSET;
 		break;
 	case IPPROTO_UDP:
```
mauriciovasquezbernal, line 179: Do we need namespaced versions of this and IPPROTO_TCP too?

include/gadget/sockets-map.h
mauriciovasquezbernal, line 1: Should this file include <linux/bpf.h> to have definition of `BPF_MAP_TYPE_HASH`? Also the header that provides `SEC("`?

(It's not related to this PR, feel free to ignore if you think it's not relevant).

include/gadget/sockets-map.h
alban: No, those are defined in an anonymous enum in vmlinux.h.

include/gadget/sockets-map.h
alban: BPF_MAP_TYPE_HASH is defined in `vmlinux.h` too. So if we include `<linux/bpf.h>`, it will conflict.

SEC is defined in libbpf's `bpf_helpers.h`. I now included it. `bpf_helpers.h` includes the following:

```c
/*
 * Note that bpf programs need to include either
 * vmlinux.h (auto-generated from BTF) or linux/types.h
 * in advance since bpf_helper_defs.h uses such types
 * as __u64.
 */
#include "bpf_helper_defs.h"
```

So the gadget must include either `vmlinux.h` or `linux/types.h` before including `sockets-map.h`. I added some `#if` and comments at the top to explain it.

mauriciovasquezbernal (approved):
LGTM. Thanks for cleaning this up!

PR #2847, opened 2024-05-14
ci: improve handling of public-key-check

mqasimsarfraz:
Currently, in order to publish ig-k8s image / gadget images on a fork users will have to:

- Set secrets for cosign (`COSIGN_PRIVATE_KEY/COSIGN_PASSWORD`)
- Update the public key in source (pkg/resources/inspektor-gadget.pub) 

Otherwise `public-key-check` job [will fail](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/9061503098/job/24893441499) and ig-k8s image /gadgets images won't be published. The idea of this PR is to improve this workflow by not signing images by default on forks. `public-key-check` is removed from the `needs` of both the image publishing steps, allowing images to be published but not signed on forks. Having said that, `public-key-check` is moved to release to ensure if incorrect keys are configured we won't be able to release. So it isn't a fatal job for CI runs on main but only for releases.

Also tests need to be adapted to allow running without verification on a fork! 

## Testing Done

- A CI run in fork with Actions secrets configured (not Dependabot secrets) checks the public key: https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/9595327823/job/26459837791
- A CI run in fork without Actions secrets configured skips the public key checking: https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/9594669084

### DIFF
--- a/.github/actions/run-integration-tests/action.yml
+++ b/.github/actions/run-integration-tests/action.yml
@@ -23,6 +23,9 @@ inputs:
   gadget_tag:
     description: 'Tag used for gadgets OCI images'
     required: true
+  gadget_verify_image:
+    description: 'Whether to verify the gadget image.'
+    default: 'true'
   test_summary_suffix:
     description: 'Suffix of the test summary file'
     required: true
@@ -84,6 +87,7 @@ runs:
           DNSTESTER_IMAGE=${{ inputs.dnstester_image }} \
           GADGET_REPOSITORY=${{ inputs.gadget_repository }} \
           GADGET_TAG=${{ inputs.gadget_tag }} \
+          GADGET_VERIFY_IMAGE=${{ inputs.gadget_verify_image }} \
           -o kubectl-gadget integration-tests |& tee integration.log & wait $!
         echo "IntegrationTestsJob: Done"
     - name: Undeploy Inspektor Gadget
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -719,7 +719,6 @@ jobs:
     if: github.event_name != 'pull_request'
     needs:
       - build-gadget-container-images
-      - public-key-check
       - check-secrets
     runs-on: ubuntu-latest
     permissions:
@@ -786,7 +785,6 @@ jobs:
     needs:
       - build-ig-container-images
       - check-secrets
-      - public-key-check
     runs-on: ubuntu-latest
     permissions:
       # allow publishing container image
@@ -847,7 +845,6 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - check-secrets
-      - public-key-check
     outputs:
       dnstester_image: ${{ steps.image-tag.outputs.dnstester }}
       ebpf_builder_image: ${{ steps.image-tag.outputs.ebpf-builder }}
@@ -1261,15 +1258,11 @@ jobs:
     # level: 1
     needs: check-secrets
     runs-on: ubuntu-latest
-    # When this job is run by a PR coming from a fork, it will not have access
-    # to secrets.
-    # Also, unless specified in the secrets configuration, dependabot cannot access
-    # to secrets.
-    # So, we skip this job when run by dependabot and from a PR coming from a
-    # fork.
-    # We still need to run this job on push, otherwise it will not be run when
-    # we merge something on main.
-    if: github.event_name == 'push' || (github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]') && needs.check-secrets.outputs.cosign == 'true'
+    # The first check means that job will only run if the secrets are present. So it will be skipped
+    # for forks, PRs from forks and dependabot PRs (if no dependabot secret are configured).
+    # The second check ensures if we forgot configuring the secrets in main repo
+    # (inspektor-gadget/inspektor-gadget) so we can catch it early.
+    if: needs.check-secrets.outputs.cosign == 'true' || (github.event_name == 'push' && github.repository == 'inspektor-gadget/inspektor-gadget')
     steps:
       - uses: actions/checkout@v4
       - name: Install Cosign
@@ -1339,7 +1332,6 @@ jobs:
       - build-ig
       - build-helper-images
       - check-secrets
-      - public-key-check
     runs-on: ubuntu-latest
     permissions:
       # allow publishing container image
@@ -1982,6 +1974,7 @@ jobs:
       - check-secrets
       - scan-gadget-container-images
       - publish-gadget-images-manifest
+      - public-key-check
     runs-on: ubuntu-latest
     permissions:
       contents: write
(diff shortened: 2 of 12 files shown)


.github/workflows/inspektor-gadget.yml
```diff
@@ -1229,11 +1228,12 @@ jobs:
...
-    # So, we skip this job when run by dependabot and from a PR coming from a
-    # fork.
+    # So, this job will be skipped when run by dependabot (if no secrets configured) and
+    # from a PR coming from a fork.
     # We still need to run this job on push, otherwise it will not be run when
-    # we merge something on main.
-    if: github.event_name == 'push' || (github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]') && needs.check-secrets.outputs.cosign == 'true'
+    # we merge something on main. Also, we ensure this job will always run on
+    # for 'inspektor-gadget/inspektor-gadget' respository
+    if: needs.check-secrets.outputs.cosign == 'true' || (github.event_name == 'push' && github.repository == 'inspektor-gadget/inspektor-gadget')
```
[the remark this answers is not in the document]
mqasimsarfraz: I should have explained the approach bit better :see_no_evil: 
> The comment refers to dependabot, but it was removed from the if, I suspect it will fail when run by the bot.

Actually the first check `needs.check-secrets.outputs.cosign == 'true'` ensures that the job will only run if secrets are configured so if no secrets are configured the CI won't run! 

> Moreover, this totally blocks forks from signing their image without tweaking the CI.

Do you mean they need to modify `(github.event_name == 'push' && github.repository == 'inspektor-gadget/inspektor-gadget')` ? 

I added this explicit check for our main repository so if somehow we delete cosign secrets the job will start to fail but for forks this check is ignored. I just [tried it on a fork ](https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/9131341274/job/25110631396) and configured the secrets and it seem to be working fine.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1303,7 +1303,6 @@ jobs:
       - build-ig
       - build-helper-images
       - check-secrets
-      - public-key-check
```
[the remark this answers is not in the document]
mqasimsarfraz: I know it isn't a nice part but the idea was to remove this dependency from here so the entire job won't be skipped and gadgets image will be published without signing if secrets aren't configured. The other approach I have in mind is to split this job into two. 1) gadget pushing 2) gadget signing. The only tricky bit is how to share the gadgets digests between the jobs (I know how to do it but would require more changes) but in that case we can move the need of `public-key-check` to gadget-signing job. What do you say?

gadgets/Makefile
```diff
@@ -2,6 +2,7 @@ SHELL := /bin/bash
 
 GADGET_TAG ?= $(shell ../tools/image-tag branch)
 GADGET_REPOSITORY ?= ghcr.io/inspektor-gadget/gadget
+GADGET_VERIFY_IMAGE ?= true
```
mauriciovasquezbernal, line 5: I'm wondering if we want to have a specific flag or a generic one as I proposed on https://github.com/inspektor-gadget/inspektor-gadget/pull/3025/commits/9c60218bb1b9b8b7eab36f642e5717e83d5ca5f7

.github/actions/run-integration-tests/action.yml
```diff
@@ -23,6 +23,9 @@ inputs:
   gadget_tag:
     description: 'Tag used for gadgets OCI images'
     required: true
+  gadget_verify_image:
+    description: 'Whether to verify the gadget image.'
+    default: 'true'
```
mauriciovasquezbernal, line 28: Why is it needed? Do we want forks to be able to customize it?

mqasimsarfraz: Yeah, like if they configure their own secrets they are still able to verify images!

mauriciovasquezbernal: Can't we hardcode it to always be true? (If they're really into it they could change the hardcoded value from the makefile).

mauriciovasquezbernal: We chatted internally, I've totally missed that this is set to `gadget_verify_image: ${{ needs.check-secrets.outputs.cosign }}` on the workflow, so my idea of hardcoding it doesn't make sense.

.github/workflows/inspektor-gadget.yml
```diff
@@ -719,7 +719,6 @@ jobs:
     if: github.event_name != 'pull_request'
     needs:
       - build-gadget-container-images
-      - public-key-check
```
mauriciovasquezbernal, line 722: Should it also be removed from the `publish-ig-images-manifest` job on L789?

mauriciovasquezbernal (approved):
LGTM. I know the current situation is not perfect (unsigned images could be pushed on main if the configuration of the secrets is wrong), but I think it's more important to allow the CI to be run from forks to avoid things like https://github.com/inspektor-gadget/inspektor-gadget/pull/3028#issuecomment-2176449389.

Please open an issue indicating further steps to improve this PR.

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
mqasimsarfraz: I just did a test commit in fork to see if the job runs when secret is configured (only Action Secrets not Dependabot secrets): https://github.com/mqasimsarfraz/inspektor-gadget/commit/897d0407b0ce238a4478d55ec6b65f68a644532d and also another run when I remove them: https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/9595327823/job/26459837791 and it seem to be working as expected. 

Also, since we don't configure Dependabot secrets then job will be skipped for it.

mqasimsarfraz:
> Please open an issue indicating further steps to improve this PR.

I opened https://github.com/inspektor-gadget/inspektor-gadget/issues/3045. Please feel free to add if I missed something.

Also, I will keep the PR open for now so we can get more reviews (cc: @eiffel-fl)

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
mqasimsarfraz: > You only shared the job using your public key, not the one without.
Can you please share the second?

Oops sorry. This is the link to the job without secrets: https://github.com/mqasimsarfraz/inspektor-gadget/actions/runs/9594669084

mqasimsarfraz:
> If I understand correctly, now we only check the correspondence between the in-source public key and secret private key before releasing.

Yes, but the check will run for all the changes to main (even if we remove the secrets) and CI will be marked red if there is a mismatch. I know it isn't ideal but should help us catch it early! 

> Is my understanding of all the consequences correct?

Yeah, sounds complete to me. 

(not related to this PR but I feel we need to think/document a way in case we want to rotate our private key)

mqasimsarfraz:
> Please, open an issue for this, I would need to check how this is normally handled before writing anything.

Done, https://github.com/inspektor-gadget/inspektor-gadget/issues/3048

mqasimsarfraz:
Thanks for the reviews!

PR #2848, opened 2024-05-15
ig/gadgetctl: add TLS support

flyth:
This PR adds support for TLS encryption & authentication for `ig` running as daemon and `gadgetctl` to connect to it. It supports mutual verification by checking certificates against certificate authorities.

Here's how to test it (this short guide to create a CA and server/client keys is just an example!!!):

```
# Create CA and self-sign
openssl ecparam -name prime256v1 -genkey -noout -out ca.key
openssl req -new -x509 -subj "/C=DE/CN=CA" -sha256 -key ca.key -out ca.crt

# Create server key, csr & cert (note that you _have_ to use SAN (subjectAltName) for verification of server names later on)
openssl req -new -sha256 -subj "/C=DE/CN=local" -addext "subjectAltName = DNS:local" -key server.key -out server.csr
openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 1000 -sha256 -copy_extensions copy

# Create client key, csr & cert
openssl ecparam -name prime256v1 -genkey -noout -out client1.key
openssl req -new -sha256 -subj "/C=DE/CN=local" -addext "subjectAltName = DNS:local" -key client1.key -out client1.csr
openssl x509 -req -in client1.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out client1.crt -days 1000 -sha256
```

In this example both client and server are using the same CA - this is however not necessary.

Copy files ca.crt, server.key & server.cert to `ig` and another copy of ca.crt, client1.key & client1.cert to `gadgetctl` and run as follows:

```
sudo ./ig daemon --group 1000 -v --tls-cert server.crt --tls-key server.key --tls-client-ca ca.crt
```

(Replace 1000 as `--group` with your preferred group id to be able to connect from a `gadgetctl` not running as root.)

```
./gadgetctl run "trace_exec:latest" --host --tls-cert client1.crt --tls-key client1.key --tls-server-ca ca.crt --verify-image false --tls-server-name local
```

Since we're connecting locally, we have to specify the `--tls-server-name` we used as SAN when creating the key pairs above.

The server log will now contain information like this:

```
...
INFO[0002] [CN=local,C=DE] GetGadgetInfo("trace_exec:latest")
INFO[0002] [CN=local,C=DE] RunGadget("trace_exec:latest")
...

### CODE REVIEW

File: pkg/gadget-service/service-oci.go [hunk 1]
```diff
@@ -35,6 +37,15 @@ func (s *Service) GetGadgetInfo(ctx context.Context, req *api.GetGadgetInfoReque
 		return nil, fmt.Errorf("expected version to be %d, got %d", api.VersionGadgetInfo, req.Version)
 	}
 
+	p, ok := peer.FromContext(ctx)
+	if ok {
+		tlsInfo, ok := p.AuthInfo.(credentials.TLSInfo)
+		if ok {
+			subject := tlsInfo.State.VerifiedChains[0][0].Subject
+			s.logger.Infof("[%s] GetGadgetInfo(%q)", subject, req.ImageName)
+		}
```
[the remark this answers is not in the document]
flyth: No - it'll always be ok for TLS connections, as they're always verified. This will only be !ok if the connection is insecure (which is a valid and right now the default use case.)

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -61,6 +65,11 @@ const (
 	ParamConnectionMethod  = "connection-method"
 	ParamConnectionTimeout = "connection-timeout"
 
+	ParamTLSKey        = "tls-key"
+	ParamTLSCert       = "tls-cert"
+	ParamTLSServerCA   = "tls-server-ca"
+	ParamTLSServerName = "tls-server-name"
```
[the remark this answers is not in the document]
flyth: Regarding keeping them local but exported:
Those Params can also be set when using the IG Go API for connecting to a remote - in that case IMHO it already helps if you're typing `grpc.Param` (when using that runtime) and your IDE can show you the available options. We've done the same with params for other operators (and should sooner or later do it for the remaining ones - but this is still in progress and discussion (also in regards to where to put their documentation)), so you can more easily find and use them.

Reusing them when being local to that package, though, would mean to introde a dependency between the server component and the client - which I think isn't a good option.

Also, even if the strings match, the meaning is different - in one case it's the client's keypair, in the other it's the servers'. Their meaning is implied from the context they're used in.

But I'll add dedicated consts to the `cmd/ig` package as well to keep it a bit cleaner at least.

/cc @mauriciovasquezbernal regarding the handling of Params

pkg/gadget-service/service-oci.go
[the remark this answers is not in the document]
flyth: Hmm... It's pretty much a rabbit hole with the question "To what degree do we trust the imported fundamental packages like gRPC to work"? But if there's ever a good time to be paranoid, this might be it - hehe. I'll add that check.

cmd/ig/daemon.go
```diff
@@ -63,6 +71,27 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+		"",
+		"Path to tls cert file")
+
+	daemonCmd.PersistentFlags().StringVarP(
+		&clientCA,
+		"tls-client-ca",
+		"",
+		"",
+		"Path to CA certificate for client validation")
+
```
mauriciovasquezbernal, line 94: Use `StringVar` to avoid passing an extra empty string.

cmd/ig/daemon.go
```diff
@@ -87,11 +116,41 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 
 		log.Infof("starting Inspektor Gadget daemon at %q", socket)
 		service := gadgetservice.NewService(log.StandardLogger(), eventBufferLength)
+
+		var options []grpc.ServerOption
+
+		if serverKey != "" || serverCert != "" || clientCA != "" {
```
mauriciovasquezbernal, line 122: I think we should we add a more specific validation logic to fail if "one but not all" flags are defined instead of relying on the functions below failing. It'll also print a better error message to the user.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -364,8 +397,44 @@ func (r *Runtime) runBuiltInGadgetOnTargets(
 
 func (r *Runtime) dialContext(dialCtx context.Context, target target, timeout time.Duration) (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithBlock(),
+		grpc.WithReturnConnectionError(),
+	}
+
+	if tlsKey, tlsCert, tlsCA := r.globalParams.Get(ParamTLSKey).String(), r.globalParams.Get(ParamTLSCert).String(), r.globalParams.Get(ParamTLSServerCA).String(); tlsKey != "" || tlsCert != "" || tlsCA != "" {
```
mauriciovasquezbernal, line 404: I think this sets the new longest line record.

cmd/ig/daemon.go
```diff
@@ -87,11 +116,41 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+
+			tlsConfig := &tls.Config{
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+				Certificates: []tls.Certificate{cert},
+				ClientCAs:    ca,
+			}
+
+			options = append(options, grpc.Creds(credentials.NewTLS(tlsConfig)))
+		} else {
+			log.Warnf("no TLS configuration provided, running in insecure mode")
```
mauriciovasquezbernal, line 146: Do we need to print this message even when unix sockets are used? I'd argue it's not needed as in that case the client needs access to the unix socket file to contact the server.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -364,8 +397,44 @@ func (r *Runtime) runBuiltInGadgetOnTargets(
...
+		ca := x509.NewCertPool()
+		caBytes, err := os.ReadFile(tlsCA)
+		if err != nil {
+			return nil, fmt.Errorf("loading client CA certificate: %w", err)
+		}
+		if ok := ca.AppendCertsFromPEM(caBytes); !ok {
+			return nil, errors.New("failed to parse client CA certificate")
+		}
+
+		purl, _ := url.Parse(target.addressOrPod)
```
mauriciovasquezbernal, line 419: I think we need to check the return code. Running the following makes it explode:

```bash 
$ gadgetctl run trace_open:latest  --remote-address tcp://127.0.0.1:1234 --tls-cert client1.crt --tls-key client1.key --tls-server-ca ca.crt --tls-server-name local
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2c373d0]

goroutine 1 [running]:
net/url.(*URL).Hostname(...)
        /usr/local/go/src/net/url/url.go:1151
github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc.(*Runtime).dialContext(0xc0002a2b70, {0x4018548, 0xc000276b90}, {{0x7ffc9f23228b?, 0xc0006c3850?}, {0x7ffc9f23228b?, 0x14a1f9b?}}, 0x12a05f200)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/runtime/grpc/grpc-runtime.go:422 +0x550
github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc.(*Runtime).getConnToRandomTarget(0xc0002a2b70, {0x4018548, 0xc000276b90}, 0x17?)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/runtime/grpc/grpc-runtime.go:347 +0x216
github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc.(*Runtime).GetGadgetInfo(0xc0002a2b70, {0x4065ce0, 0xc0005fd500}, 0xd?, 0xc0005618c0)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/runtime/grpc/oci.go:39 +0x17d
github.com/inspektor-gadget/inspektor-gadget/cmd/common.NewRunCommand.func1(0xc0004f4f08, {0xc00045fd90, 0xb, 0xb})
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/cmd/common/oci.go:107 +0x4e3
github.com/spf13/cobra.(*Command).execute(0xc0004f4f08, {0xc00045fd90, 0xb, 0xb})
        /home/mauriciov/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:968 +0xa44
github.com/spf13/cobra.(*Command).ExecuteC(0xc000160f08)
        /home/mauriciov/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1115 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        /home/mauriciov/go/pkg/mod/github.com/spf13/cobra@v1.8.0/command.go:1039
main.main()
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/cmd/gadgetctl/main.go:92 +0x745
$
``

File: pkg/runtime/grpc/grpc-runtime.go [hunk 8]
```diff
@@ -162,6 +171,30 @@ func (r *Runtime) GlobalParamDescs() params.ParamDescs {
...
+			Description:  "TLS server CA certificate",
+			DefaultValue: "",
+			TypeHint:     params.TypeString,
+		},
+		{
+			Key:          ParamTLSServerName,
+			Description:  "override TLS server name (if omitted, using target server name)",
+			DefaultValue: "",
+			TypeHint:     params.TypeString,
+		},
```
mauriciovasquezbernal, line 197: Shouldn't they be moved under `case ConnectionModeDirect` to avoid showing them in the kubernetes case?

mauriciovasquezbernal (commented):
It's looking very good. Some initial comments.

I think there is missing a `openssl ecparam -name prime256v1 -genkey -noout -out server.key` step in the description.

cmd/ig/daemon.go
```diff
@@ -87,11 +116,41 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+			}
+
+			ca := x509.NewCertPool()
+			caBytes, err := os.ReadFile(clientCA)
+			if err != nil {
+				return fmt.Errorf("loading client CA certificate: %w", err)
+			}
+
+			if ok := ca.AppendCertsFromPEM(caBytes); !ok {
+				return errors.New("failed to parse client CA certificate")
```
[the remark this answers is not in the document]
flyth: I wondered the same: it only complains when wrapping errors - this is the last message in the chain, so it's fine. (e.g. "doing this: doing that: doing another thing: failed to blah")

pkg/runtime/grpc/grpc-runtime.go
flyth: So where is my trophy, then? 😆

pkg/runtime/grpc/grpc-runtime.go
flyth: Hmm yeah, I should definitely add that - it shouldn't fail for the given address, nevertheless.

pkg/runtime/grpc/grpc-runtime.go
flyth: At least until we also support it for ig-k8s, yes.

cmd/ig/daemon.go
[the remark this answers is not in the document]
flyth: Hmm, just ran into a similar situation on another PR without wrapping and it complained. 🤷

pkg/gadget-service/service-oci.go
```diff
@@ -50,6 +52,17 @@ func (s *Service) GetGadgetInfo(ctx context.Context, req *api.GetGadgetInfoReque
 		return nil, fmt.Errorf("expected version to be %d, got %d", api.VersionGadgetInfo, req.Version)
 	}
 
+	p, ok := peer.FromContext(ctx)
+	if ok {
```
mauriciovasquezbernal, line 56:
```suggestion
	if ok && p.AuthInfo != nil {
```

Otherwise it fails when with `Error: fetching gadget information: getting gadget info: rpc error: code = Unknown desc = authentication information are not TLS: <nil>` when no TLS information is provided.

(same below)

pkg/gadget-service/service-oci.go
```diff
@@ -50,6 +52,17 @@ func (s *Service) GetGadgetInfo(ctx context.Context, req *api.GetGadgetInfoReque
...
+		tlsInfo, ok := p.AuthInfo.(credentials.TLSInfo)
+		if !ok {
+			return nil, fmt.Errorf("authentication information are not TLS: %v", p.AuthInfo)
```
mauriciovasquezbernal, line 59:
```suggestion
			return nil, fmt.Errorf("authentication information is not TLS: %v", p.AuthInfo)
```

cmd/ig/daemon.go
```diff
@@ -100,11 +127,57 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+	* tls-cert: %q
+	* tls-client-ca: %q
+All these options should be set at the same time to enable TLS connection`,
+				serverKey, serverCert, clientCA)
+		}
+
+		var options []grpc.ServerOption
+
+		if serverKey != "" && serverCert != "" && clientCA != "" {
+			cert, err := tls.LoadX509KeyPair(serverCert, serverKey)
```
mauriciovasquezbernal, line 150: nit: emit some debugging messages indicating that TLS is enabled and which key, cert and clientca will be used.

cmd/ig/daemon.go
```diff
@@ -100,11 +127,57 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+
+		if tlsOptionsSet > 1 && tlsOptionsSet < 3 {
+			return fmt.Errorf(`
+missing at least one the TLS related options:
+	* tls-key: %q
+	* tls-cert: %q
+	* tls-client-ca: %q
+All these options should be set at the same time to enable TLS connection`,
+				serverKey, serverCert, clientCA)
+		}
```
mauriciovasquezbernal, line 145: It's fine now. Later on we can improve it to allow only client or server verification if we want to

cmd/ig/daemon.go
```diff
@@ -65,6 +74,24 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 		16384,
 		"The events buffer length. A low value could impact horizontal scaling.")
 
+	daemonCmd.PersistentFlags().StringVar(
+		&serverKey,
+		"tls-key",
+		"",
+		"Path to tls key file")
```
mauriciovasquezbernal, line 81: tls -> TLS in other places as well

mauriciovasquezbernal (commented):
It looks very good. Just one blocking comment on missing nil check.

cmd/ig/daemon.go
flyth, line 81: Maybe now that we're also adding other keys directly to the params/flags or config file, we should rather name this `tls-key-file` _or_ accept the key contents directly. Same for the other params.

cmd/ig/daemon.go
```diff
@@ -100,11 +127,57 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+	* tls-key: %q
+	* tls-cert: %q
+	* tls-client-ca: %q
+All these options should be set at the same time to enable TLS connection`,
+				serverKey, serverCert, clientCA)
+		}
+
+		var options []grpc.ServerOption
+
+		if serverKey != "" && serverCert != "" && clientCA != "" {
```
flyth, line 149: `tlsOptionsSet == 3`?

cmd/ig/daemon.go
```diff
@@ -100,11 +127,57 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
...
+
+			tlsConfig := &tls.Config{
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+				Certificates: []tls.Certificate{cert},
+				ClientCAs:    ca,
+			}
+
+			options = append(options, grpc.Creds(credentials.NewTLS(tlsConfig)))
+		} else if !strings.HasPrefix(socket, "unix") {
+			log.Warnf("no TLS configuration provided, communication between daemon and CLI will not be ciphered")
```
flyth, line 173:
```suggestion
			log.Warnf("no TLS configuration provided, communication between daemon and CLI will not be encrypted")
```
Easier to understand IMHO. Maybe even replace `daemon and CLI` with `client and server`.

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -364,9 +393,49 @@ func (r *Runtime) runBuiltInGadgetOnTargets(
...
 func (r *Runtime) dialContext(dialCtx context.Context, target target, timeout time.Duration) (*grpc.ClientConn, error) {
 	opts := []grpc.DialOption{
-		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		//nolint:staticcheck
 		grpc.WithBlock(),
+		//nolint:staticcheck
+		grpc.WithReturnConnectionError(),
+	}
+
+	if tlsKey, tlsCert, tlsCA := r.globalParams.Get(ParamTLSKey).String(), r.globalParams.Get(ParamTLSCert).String(), r.globalParams.Get(ParamTLSServerCA).String(); tlsKey != "" || tlsCert != "" || tlsCA != "" {
```
flyth, line 402: Do you want to have an error message similar to that in the daemon, if one or two params are _not_ set?

flyth (commented):
Thanks for adopting this!

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -364,9 +392,68 @@ func (r *Runtime) runBuiltInGadgetOnTargets(
...
+			Certificates: []tls.Certificate{cert},
+			RootCAs:      ca,
+		}
+
+		if serverName := r.globalParams.Get(ParamTLSServerName).String(); serverName != "" {
+			tlsConfig.ServerName = serverName
+		}
+
+		if tlsConfig.ServerName == "" {
+			return nil, errors.New("invalid hostname, use --tls-server-name to override")
```
mauriciovasquezbernal, line 451: nit: tls-server-name -> ParamTLSServerName

pkg/utils/tls/tls.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"os"
+)
+
+func CreateTLSCert(serverCert, serverKey string) (tls.Certificate, error) {
```
mauriciovasquezbernal, line 25:
```suggestion
func LoadTLSCert(serverCert, serverKey string) (tls.Certificate, error) {
```

pkg/utils/tls/tls.go
```diff
@@ -0,0 +1,46 @@
...
+	cert, err := tls.LoadX509KeyPair(serverCert, serverKey)
+	if err != nil {
+		return tls.Certificate{}, fmt.Errorf("loading TLS keypair: %w", err)
+	}
+
+	return cert, nil
+}
+
+func CreateTLSCA(clientCA string) (*x509.CertPool, error) {
```
mauriciovasquezbernal, line 34:
```suggestion
func LoadTLSCA(clientCA string) (*x509.CertPool, error) {
```

mauriciovasquezbernal (approved):
Some minor nits, but LGTM alrady.

PR #2850, opened 2024-05-15
pkg/datasource: remove dependency on pkg/gadgets

flyth:
This removes pkg/datasource's dependency on pkg/gadgets by copying the `FromCString()` function. Main reason behind it is that `pkg/gadgets` itself imports other things that are partially OS specific (pkg/btfgen / pkg/host) and prevent compilation on non-linux hosts.


mauriciovasquezbernal (approved):
LGTM!

PR #2851, opened 2024-05-15
pkg/operators/simple: fix onStop() callback being called if only onStart is set

flyth:
A wrong check in `Stop()` could lead to a crash if `onStart` had been registered but no `onStop`


blanquicet (approved):
LGTM!

PR #2860, opened 2024-05-16
trace_dns: fix padding

alban:
The dns gadget uses '#pragma pack(2)' to define some structs. Using pragma push/pod is preferred to limit the scope of the alignment setting to the struct definition that needs it.

There is no concrete issue at the moment, so there is no need to backport this patch. I was experimenting with some changes in the dns gadget, and the struct alignment was not as I expected.

This patch should make the code more robust for future changes.

Fixes: 4f7150e96093 ("Add NumAnswers and Addresses to DNS trace events")


mauriciovasquezbernal (approved):
LGTM!

PR #2877, opened 2024-05-21
local/kube manager operators: Print warning for multiple net/mnt ns fields 

blanquicet:
# Print warning for multiple net/mnt ns fields in local/kube manager operators

Let users know that we are not using all the net/mnt ns fields for enrichment but just one.

## How to use

A gadget defining more than one mntns or netns fileds. For instance:

```c
struct process_entry {
	gadget_mntns_id mntns_id;
	gadget_mntns_id jose;
	__u32 pid;
	__u32 tid;
	__u32 ppid;
	__u32 uid;
	__u32 gid;
	char comm[TASK_COMM_LEN];
};
```

IG will print a warning error like this:

```bash
$ IG_EXPERIMENTAL=true go run --exec "sudo -E" ./cmd/ig/... run $LOCAL_REPO/snapshot_process --insecure --verify-image=false -o jsonpretty -v
...
DEBU[0000] using mntns enrichment
WARN[0000] multiple fields with "type:gadget_mntns_id" found in DataSource "processes", only "mntns_id" will be used
...
```

### DIFF
--- a/gadgets/snapshot_socket/program.bpf.c
+++ b/gadgets/snapshot_socket/program.bpf.c
@@ -52,7 +52,7 @@ struct socket_entry {
 	struct gadget_l4endpoint_t dst;
 	__u32 state;
 	__u32 ino;
-	__u32 netns;
+	gadget_netns_id netns;
 };
 
 GADGET_SNAPSHOTTER(sockets, socket_entry, ig_snap_tcp, ig_snap_udp);
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -39,7 +39,7 @@ struct event_t {
 	struct gadget_l4endpoint_t dst;
 
 	gadget_mntns_id mntns_id;
-	__u32 netns;
+	gadget_netns_id netns;
 	__u32 pid;
 	__u32 tid;
 	__u32 uid;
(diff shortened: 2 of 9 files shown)


mauriciovasquezbernal (commented):
This LGTM from code inspection. I'm wondering if this message should also be printed why building the gadget, in the end it's up to the gadget developer and there is nothing the user can make about it.

Perhaps we could even fail the build process if two fields are found.

blanquicet:
> This LGTM from code inspection. I'm wondering if this message should also be printed why building the gadget, in the end it's up to the gadget developer and there is nothing the user can make about it.
> 
> Perhaps we could even fail the build process if two fields are found.
>
The current building implementation doesn't use the same code but still the `gadgets/run/types` pkg. There, we don't know yet if the local or kube manager operators were enabled. At most we could print this:

```bash
$ IG_EXPERIMENTAL=true go run --exec "sudo -E" ./cmd/ig/... image build -t $LOCAL_REPO/snapshot_process gadgets/snapshot_process/
WARN[0001] Using multiple fields of "type:gadget_mntns_id" may cause unpredictable behavior during enrichment
Successfully built 127.0.0.1:5000/snapshot_process:latest@sha256:834a915fe796ee07114790435ecc9ee688e60c35854ee3ee7feae7c47bdd5ae4
```

WDYT?

mauriciovasquezbernal (approved):

Issue #2872, opened 2024-05-20
Running an image-based gadget with `--pull=never` option, but image is still getting pulled


Sefi4:
## Description
I was following the hello world gadget tutorial (https://www.inspektor-gadget.io/docs/latest/devel/hello-world-gadget/) and when trying to run the image using `sudo -E ig run mygadget:latest` I got the following error (which is expected):

Error: fetching gadget information: initializing and preparing operators: instantiating operator "oci": ensuring image: verifying image "mygadget:latest": getting signing information: getting signature: getting signature bytes: GET "https://ghcr.io/v2/inspektor-gadget/gadget/mygadget/manifests/sha256-f85471cc2b20d7febd69e2af95b91888fbf4b0b597504f575f6b2f315ac5f5e3.sig": GET "https://ghcr.io/token?scope=repository%3Ainspektor-gadget%2Fgadget%2Fmygadget%3Apull&service=ghcr.io": response status code 403: denied: requested access to the resource is denied

However, when running `sudo -E ig run mygadget:latest --pull-secret=false` I still got the same exact error, meaning the image is still getting pulled.

## Impact
Unable to run image-based gadgets.

## Environment and steps to reproduce

1. **Set-up**: I'm using WSL with Docker Desktop v4.30.
2. **Task**: Following the Hello World Gadget tutorial (https://www.inspektor-gadget.io/docs/latest/devel/hello-world-gadget/)
3. **Action(s)**: Copy the program.pbf.c from the tutorial, build the image with `sudo -E ig image build -t mygadget:latest .' and run the gadget with `sudo -E ig run mygadget:latest --pull-secret=false`
5. **Error**: Error: fetching gadget information: initializing and preparing operators: instantiating operator "oci": ensuring image: verifying image "mygadget:latest": getting signing information: getting signature: getting signature bytes: GET "https://ghcr.io/v2/inspektor-gadget/gadget/mygadget/manifests/sha256-f85471cc2b20d7febd69e2af95b91888fbf4b0b597504f575f6b2f315ac5f5e3.sig": GET "https://ghcr.io/token?scope=repository%3Ainspektor-gadget%2Fgadget%2Fmygadget%3Apull&service=ghcr.io": response status code 403: denied: requested access to the resource is denied

## Expected behavior
I would expect the hello world gadget to run.

## Additional information

Please add any information here that does not fit the above format.

alban:
@eiffel-fl Do you know about this issue?

[...]

PR #2887, opened 2024-05-22
docs/hello-world-gadget: Add missing --verify-image=false

mauriciovasquezbernal:
Avoid image verification, otherwise the commands will fail.

Related  #2872

--- 

@eiffel-fl do we have any documentation to sign gadget images? I couldn't find it.


mauriciovasquezbernal:
I opened #2890 to improve documentation on the image signing.

PR #2901, opened 2024-05-27
clean up operator initialization code

### LINKED ISSUE #2891: Initialize operators outside the gadget context
> We're currently not doing it correctly. IMHO, the caller should always make sure operators are initialized properly instead of the gadgetCtx (which should only do the instance initialization).
> 
> Basically all our entry points (ig/main.go, kubectl-gadget/main.go, gadgetctl/main.go and gadgettracermgr) should initialize all operators globally from a config file (or params or env, basically the cobra/viper combo). It could be done by a wrapper so that it works the same way for all of them.
> 

See https://github.com/inspektor-gadget/inspektor-gadget/pull/2753#discussion_r1594293183

Related #2831

flyth:
Previously, for image-based gadgets, operators would lazily be initialized by the gadget context. That made it difficult to forward the actual global configuration of operators. Since we didn't really use those in the past, it was not a big issue.

However, now that we're for example using image verification that can be enabled/disabled using parameters, we need to clean this up. Currently, the client decides whether image validation is done by sending the appropriate flag to the server. This is however insecure, as the client shouldn't have authority over that decision. We may later on allow (using another param/setting that can be enabled on the server) clients to override certain behavior, but in general the server should decide about the security features.

This PR moves the initialization to all our entry points for image-based gadgets and exports the parameters to be set there, if applicable.

This means:
* cmd/common/oci.go exposes global operator params to client (or ig used in standalone mode)
* pkg/gadget-service collects and exposes params using a function
* cmd/ig/daemon.go uses the params exposed by the gadget-service
* gadgettracermgr should also expose those params exposed by the gadget-service, this is however currently not handled in this PR (but maybe should also be added; using env variables on deployment maybe?)

Fixes #2891

### DIFF
--- a/cmd/common/oci.go
+++ b/cmd/common/oci.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/frontends/console"
@@ -41,7 +42,10 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 	ociParams := apihelpers.ToParamDescs(ocihandler.OciHandler.InstanceParams()).ToParams()
 
 	// Add operator global flags
-	operatorsGlobalParamsCollection := operators.GlobalParamsCollection()
+	opGlobalParams := make(map[string]*params.Params)
+	for _, op := range operators.GetDataOperators() {
+		opGlobalParams[op.Name()] = apihelpers.ToParamDescs(op.GlobalParams()).ToParams()
+	}
 
 	// gadget parameters are only available after contacting the server
 	gadgetParams := make(params.Params, 0)
@@ -89,6 +93,12 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 
 			ops := make([]operators.DataOperator, 0)
 			for _, op := range operators.GetDataOperators() {
+				// Initialize operator
+				err := op.Init(opGlobalParams[op.Name()])
+				if err != nil {
+					log.Warnf("error initializing operator %s: %v", op.Name(), err)
+					continue
+				}
 				ops = append(ops, op)
 			}
 			ops = append(ops, clioperator.CLIOperator)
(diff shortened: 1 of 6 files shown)


pkg/gadget-service/service.go
```diff
@@ -62,16 +64,27 @@ type Service struct {
...
+		servers:   map[*grpc.Server]struct{}{},
+		logger:    defaultLogger,
+		operators: ops,
 	}
 }
 
+func (s *Service) SetEventBufferLength(val uint64) {
+	s.eventBufferLength = val
+}
+
```
mauriciovasquezbernal, line 87: nit: I suppose this change really belongs to the "cmd/ig: expose operator global params to cobra" commit.

I think this should also be together with commits that use it, otherwise the compilation is broken.

cmd/common/oci.go
```diff
@@ -89,6 +93,12 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 
 			ops := make([]operators.DataOperator, 0)
 			for _, op := range operators.GetDataOperators() {
```
mauriciovasquezbernal, line 95: Don't we have global parameters for layer operators? I suppose we'll need to add them later on.

pkg/gadget-service/service.go
```diff
@@ -62,16 +64,27 @@ type Service struct {
 	logger            logger.Logger
 	servers           map[*grpc.Server]struct{}
 	eventBufferLength uint64
+
+	// operatorGlobalParams stores all global parameters for DataOperators (non-legacy)
```
mauriciovasquezbernal, line 68: Outdated comment?

mauriciovasquezbernal (approved):
LGTM. Thanks for handling it. 

> gadgettracermgr should also expose those params exposed by the gadget-service, this is however currently not handled in this PR (but maybe should also be added; using env variables on deployment maybe?)

I think it can be handled once we have https://github.com/inspektor-gadget/inspektor-gadget/pull/2831 (cc @mqasimsarfraz)

I noticed that the order of the commits breaks the compilation for some of them, can it be fixed?

pkg/gadget-service/service.go
flyth: I really need to find a better way to manage these kind of things - took way too long 😞

cmd/common/oci.go
flyth: Not yet, but yeah, we should add them sooner or later.

flyth:
Thanks for the review!

PR #2902, opened 2024-05-27
oci-handler: handle parameters globally

flyth:
This moves the parameters of the oci-handler from the instances to the global part. The reason is that most of the parameters are security sensitive and thereby should be set on a server level and not be modified by the client (when in a client/server environment like ig as daemon or `ig-k8s`).

We might need to think about additional setting that partially allow overriding these settings from the client, but the default configuration should prevent such behavior.

Also, some parameters will probably have to be added to the deployment on `ig-k8s` and set from the `gadgettracermgr`.

This requires #2901

### DIFF
--- a/pkg/operators/oci-handler/oci.go
+++ b/pkg/operators/oci-handler/oci.go
@@ -25,7 +25,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
-	apihelpers "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api-helpers"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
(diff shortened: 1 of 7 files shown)


pkg/operators/oci-handler/oci.go
```diff
@@ -43,21 +42,20 @@ const (
...
 	return nil
 }
 
 func (o *ociHandler) GlobalParams() api.Params {
-	return nil
-}
-
-func (o *ociHandler) InstanceParams() api.Params {
 	return api.Params{
 		// Hardcoded for now
```
mauriciovasquezbernal, line 60: Not true anymore?

mauriciovasquezbernal (commented):
It makes sense to me. Is validating the metadata something we want to configure on the server side? (I think it makes sense, an administrator could like to avoid running gadgets without a valid metadata)

Before merging this we need to support this on gadget tracer manager, otherwise we'll block users from run their custom gadgets on k8s. (As they can't disable the image verification nor provide their own public key)

pkg/operators/oci-handler/oci.go
flyth: I don't know what it addressed originally 😄 But I guess it can be removed.

PR #2906, opened 2024-05-29
tests: Add tests for image export and import

mauriciovasquezbernal:
Add missing tests from #2779

### DIFF
--- a/integration/components/image/image_test.go
+++ b/integration/components/image/image_test.go
@@ -51,6 +51,13 @@ func TestImage(t *testing.T) {
 
 	tmpFolder := t.TempDir()
 	exportPath := path.Join(tmpFolder, "export.tar")
+	gadgetSrcFolder := path.Join(tmpFolder, "gadget")
+	err := os.MkdirAll(gadgetSrcFolder, 0o755)
+	require.NoError(t, err)
+
+	// create an empty eBPF program for the test as it compiles fine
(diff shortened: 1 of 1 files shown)


integration/components/image/image_test.go
```diff
@@ -182,6 +186,22 @@ func TestImage(t *testing.T) {
...
+		{
+			name: "export",
+			cmd:  commonImage.NewExportCmd(),
+			args: []string{testRegistryImage, exportPath},
+			expectedStdout: []string{
+				fmt.Sprintf("Successfully exported images to %s", exportPath),
+			},
+		},
+		{
+			name: "import",
```
burak-ok, line 205: nit: Add comment that the order of these 2 test steps are important

integration/components/image/image_test.go
```diff
@@ -182,6 +186,22 @@ func TestImage(t *testing.T) {
...
+			cmd:  commonImage.NewImportCmd(),
+			args: []string{exportPath},
+			expectedStdout: []string{
+				fmt.Sprintf("Successfully imported images:\n  %s", testRegistryImage),
```
burak-ok, line 209: Is there an easy way to check if the imported image is the same as the one we exported?

burak-ok (approved):
LGTM, just nits

integration/components/image/image_test.go
mauriciovasquezbernal: I don't think so. We could check the image digest but with the current structure of the test it's difficult to do.

PR #2911, opened 2024-05-29
ci/linter: bump golangci-lint version to handle Go 1.22 features

flyth:
e.g. ranging over ints


flyth:
Thanks for the review!

Issue #2908, opened 2024-05-29
[RFE] Add syscall formatter


i-Pear:
## Current situation

In `gadgets/trace_capabilities`, syscalls will be reported as numbers.

## Ideal future situation

Syscalls should be showed in strings, such as `SYS_OPENAT`.

## Implementation options

Similar to #2630, formatter is a good way to achieve this.

PR #2913, opened 2024-05-29
pkg/operators/formatters: Add syscall formatter

i-Pear:
# pkg/operators/formatters: Add syscall formatter

Syscalls will be formatted to strings, such as `SYS_OPENAT`.

Solves #2908

## How to use

Use `gadget_syscall` as the type of event field.

### DIFF
--- a/pkg/operators/formatters/formatters.go
+++ b/pkg/operators/formatters/formatters.go
@@ -18,6 +18,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"strings"
 	"syscall"
 	"time"
 
@@ -28,6 +29,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls"
 )
 
 // Keep this aligned with include/gadget/types.h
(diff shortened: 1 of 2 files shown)


pkg/operators/formatters/formatters.go
```diff
@@ -159,6 +163,39 @@ var replacers = []replacer{
...
 	},
+	{
+		name:      "syscall",
+		selectors: []string{"type:" + SyscallTypeName},
+		replace: func(ds datasource.DataSource, in datasource.FieldAccessor) (func(data datasource.Data) error, error) {
+			oldName := in.Name()
+
+			if err := in.Rename(oldName + "_raw"); err != nil {
+				return nil, fmt.Errorf("renaming field: %w", err)
+			}
```
flyth, line 174: I know we've used this behavior for backwards compatibility, but also had some discussions around how to improve naming. The issue is that the field type would change depending on whether this operator is present or not.

/cc @mauriciovasquezbernal @alban

flyth (commented):
Already looks great, thanks!

pkg/operators/formatters/formatters.go
```diff
@@ -159,6 +164,41 @@ var replacers = []replacer{
 		},
 		priority: 0,
 	},
+	{
+		name:      "syscall",
+		selectors: []string{"type:" + SyscallTypeName},
+		replace: func(ds datasource.DataSource, in datasource.FieldAccessor) (func(data datasource.Data) error, error) {
+			if in.Type() != api.Kind_Uint64 {
+				return nil, fmt.Errorf("checking field %q: should be uint64", in.Name())
```
flyth, line 172: nit:
```suggestion
				return nil, fmt.Errorf("checking field %q: expected uint64", in.Name())
```

flyth (approved):
LGTM, thanks!

flyth:
Thanks for working on this!

PR #2910, opened 2024-05-29
tools/ekscleanup: Use go.work.

eiffel-fl:
Hi.

The recent addition of `go.work` broke the CI:
https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/9281060516/job/25536388441#step:5:12

Adding this package to `go.work` solves this issue:

```bash
$ go run . -dryrun                          francis/fix-ekscleanup %
error: describing stacks: describing stacks: NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors
exit status 1
$ git checkout main                         francis/fix-ekscleanup %
Basculement sur la branche 'main'
Votre branche est à jour avec 'origin/main'.
$ go run .                                                 main % u=
main module (github.com/inspektor-gadget/inspektor-gadget) does not contain package github.com/inspektor-gadget/inspektor-gadget/tools/eks-cleanup
```

Best regards.

PR #2914, opened 2024-05-29
Remove usage of go workspaces

mauriciovasquezbernal:
71f3d26c3ea1 ("examples: Create go module") introduced a go workspace to guarantee that the examples are compiled with the current version of Inspetkor Gadget, however it created some other problems around:

- https://github.com/inspektor-gadget/inspektor-gadget/pull/2910
- https://github.com/inspektor-gadget/inspektor-gadget/pull/2909


mauriciovasquezbernal:
I think I forced push too many times and CI became crazy.

PR #2919, opened 2024-05-30
Verify ebpf builder image

eiffel-fl:
Hi.

This PR adds verification of eBPF builder image when building an image-based gadgets:

```bash
# OK
$ sudo -E ./ig image build --builder-image ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74 --public-key="$(cat eiffel-fl.pub)" -t trace_exec:verified -f gadgets/trace_exec/gadget.yaml gadgets/trace_exec
Pulling builder image ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74
ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74: Pulling from eiffel-fl/ebpf-builder
Digest: sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74
Status: Image is up to date for ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74
Successfully built ghcr.io/inspektor-gadget/gadget/trace_exec:verified@sha256:ba9d9d6d97e651d48aa54744bd1eb30d5e3fe11a994647eed3e9d520c11decf7
# Signed image, wrong key
$ sudo -E ./ig image build --builder-image ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74 -t trace_exec:verified -f gadgets/trace_exec/gadget.yaml gadgets/trace_exec
Pulling builder image ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74
ghcr.io/eiffel-fl/ebpf-builder@sha256:106d231ea9c78b9c3f24a4a63314637b9c58c4085580b4889dcd801cb29ade74: Pulling from eiffel-fl/ebpf-builder
Digest: sha256:106d231ea9c78b9c3f24a4a63314637b9
# Unsigned image, good key
$ sudo -E ./ig image build -t trace_exec:verified -f gadgets/trace_exec/gadget.yaml gadgets/trace_exec                                                                 
Pulling builder image ghcr.io/inspektor-gadget/ebpf-builder:latest
latest: Pulling from inspektor-gadget/ebpf-builder
c6cf28de8a06: Already exists 
891494355808: Already exists 
6582c62583ef: Already exists 
6d74711819e4: Already exists 
5ca436926747: Already exists 
a583c4d6eaf8: Already exists 
4f4fb700ef54: Pull complete 
6748707bb0b9: Pull complete 
b922e4ffae44: Pull complete 
63b4e20a56b7: Pull complete 
50e25b6234ef: Pull complete 
69ce03b9b098: Pull complete 
1cadb653954f: Pull complete 
b282bbca931d: Pull complete 
53e61673e932: Pull complete 
b122f5e56bff: Pull complete 
Digest: sha256:81254ecba6efc14ce92d425383ec555abd0be1954484bbda4e0687030eb177f2
Status: Downloaded newer image for ghcr.io/inspektor-gadget/ebpf-builder:latest
Error: verifying builder image: getting signing information: getting signature: getting signature bytes: ghcr.io/inspektor-gadget/ebpf-builder:sha256-81254ecba6efc14ce92d425383ec555abd0be1954484bbda4e0687030eb177f2.sig: not found
```

Best regards.

### DIFF
--- a/cmd/common/image/build.go
+++ b/cmd/common/image/build.go
@@ -40,6 +40,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/resources"
 )
 
 //go:embed Makefile.build
@@ -76,6 +77,8 @@ type cmdOpts struct {
 	validateMetadata bool
 	btfgen           bool
 	btfhubarchive    string
+	verifyImage      bool
+	publicKey        string
 }
 
 func NewBuildCmd() *cobra.Command {
@@ -110,6 +113,15 @@ func NewBuildCmd() *cobra.Command {
 
 	cmd.Flags().BoolVar(&opts.btfgen, "btfgen", false, "Enable btfgen")
 	cmd.Flags().StringVar(&opts.btfhubarchive, "btfhub-archive", "", "Path to the location of the btfhub-archive files")
+
+	cmd.Flags().BoolVar(
+		&opts.verifyImage,
+		"verify-image",
+		true,
+		"Verify the builder image using the provided public key")
+	cmd.Flags().StringVar(
+		&opts.publicKey,
+		"public-key", resources.InspektorGadgetPublicKey, "Public key used to verify the builder image")
 
 	return utils.MarkExperimental(cmd)
 }
@@ -303,6 +315,48 @@ func ensureBuilderImage(ctx context.Context, cli *client.Client, builderImage st
 	return jsonmessage.DisplayJSONMessagesStream(reader, out, outFd, isTTY, nil)
 }
 
+func getBuilderImageDigest(ctx context.Context, cli *client.Client, builderImage string) (string, error) {
+	builderImageInspect, _, err := cli.ImageInspectWithRaw(ctx, builderImage)
+	if err != nil {
+		return "", fmt.Errorf("inspecting builder image: %w", err)
+	}
+
+	repoDigests := len(builderImageInspect.RepoDigests)
+	expectedDigest := 1
+	if repoDigests != expectedDigest {
+		return "", fmt.Errorf("bad number of repository digest: expected %d, got %d", expectedDigest, repoDigests)
+	}
+
+	repoDigest := builderImageInspect.RepoDigests[0]
+	parts := strings.Split(repoDigest, "@")
+
+	if len(parts) != 2 {
+		return "", fmt.Errorf("repository digest has wrong format: got %s, expected image@hash", repoDigest)
+	}
+
+	return parts[1], nil
+}
+
+func verifyBuilderImage(ctx context.Context, cli *client.Client, builderImage string, imgOpts *oci.ImageOptions) error {
+	if !imgOpts.VerifyPublicKey {
+		fmt.Fprintf(os.Stderr, "you set --verify-image=false, builder image will not be verified\n")
+
+		return nil
+	}
+
+	builderImageRef, err := oci.NormalizeImageName(builderImage)
+	if err != nil {
+		return fmt.Errorf("normalizing builder image name: %w", err)
+	}
+
+	builderImageDigest, err := getBuilderImageDigest(ctx, cli, builderImage)
+	if err != nil {
+		return fmt.Errorf("getting builder image digest: %w", err)
+	}
+
+	return oci.VerifyImage(ctx, builderImageRef, builderImageDigest, imgOpts)
+}
+
 func buildInContainer(opts *cmdOpts, conf *buildFile) error {
 	cwd, err := os.Getwd()
 	if err != nil {
@@ -318,6 +372,22 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
 
 	if err := ensureBuilderImage(ctx, cli, opts.builderImage); err != nil {
 		return err
+	}
+
+	imgOpts := &oci.ImageOptions{
+		VerifyOptions: oci.VerifyOptions{
+			VerifyPublicKey: opts.verifyImage,
+			PublicKey:       opts.publicKey,
+		},
+		AuthOptions: oci.AuthOptions{
+			// We need to set this to connect to the remote repository.
+			AuthFile: oci.DefaultAuthFile,
+		},
+	}
+
+	err = verifyBuilderImage(ctx, cli, opts.builderImage, imgOpts)
+	if err != nil {
+		return fmt.Errorf("verifying builder image: %w", err)
 	}
 
 	wasmFullPath := ""


pkg/oci/oci.go
```diff
@@ -706,6 +706,14 @@ func verifyImage(ctx context.Context, image string, imgOpts *ImageOptions) error
...
 	}
 
+	return VerifyImage(ctx, imageRef, imageDigest, imgOpts)
+}
+
+// VerifyImage is used to verify the image corresponding to the reference and
+// digest with the given public key.
+// If the image was signed with the given public key, nothing is returned,
+// otherwise an error is returned.
+func VerifyImage(ctx context.Context, imageRef reference.Named, imageDigest string, imgOpts *ImageOptions) error {
```
mauriciovasquezbernal, line 716: I find it very confusing to have two functions with almost the same name and such a different signatures. Perhaps the contents of `verifyImage` can be inlined in `ensureImage`.

cmd/common/image/build.go
```diff
@@ -111,6 +114,15 @@ func NewBuildCmd() *cobra.Command {
 	cmd.Flags().BoolVar(&opts.btfgen, "btfgen", false, "Enable btfgen")
 	cmd.Flags().StringVar(&opts.btfhubarchive, "btfhub-archive", "", "Path to the location of the btfhub-archive files")
 
+	cmd.Flags().BoolVarP(
```
mauriciovasquezbernal, line 117: BoolVar to avoid passing the empty string. (same below)

cmd/common/image/build.go
```diff
@@ -111,6 +114,15 @@ func NewBuildCmd() *cobra.Command {
...
+		&opts.verifyImage,
+		"verify-image", "",
+		true,
+		"verify container image if policy-controller is installed on the cluster")
```
mauriciovasquezbernal, line 121: This comment needs to be updated. I got shocked when I read "policy controller" in this context :scream:

cmd/common/image/build.go
```diff
@@ -320,6 +374,21 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
...
+		AuthOptions: oci.AuthOptions{
+			AuthFile: oci.DefaultAuthFile,
+		},
+	}
+
+	err = verifyBuilderImage(ctx, cli, opts.builderImage, imgOpts)
+	if err != nil {
+		return fmt.Errorf("verifying builder image: %w", err)
+	}
+
```
mauriciovasquezbernal, line 391: Can this be moved to `ensureBuilderImage` to verify only once when the image is pulled by the first time?

cmd/common/image/build.go
```diff
@@ -320,6 +374,21 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
...
 	}
 
+	imgOpts := &oci.ImageOptions{
+		VerifyOptions: oci.VerifyOptions{
+			VerifyPublicKey: opts.verifyImage,
+			PublicKey:       opts.publicKey,
+		},
+		AuthOptions: oci.AuthOptions{
+			AuthFile: oci.DefaultAuthFile,
+		},
```
mauriciovasquezbernal, line 384: Is is this part needed?

mauriciovasquezbernal (commented):
Some initial comments.

pkg/oci/oci.go
mauriciovasquezbernal: Just to be sure we're on the same page, I was suggesting something like this:

```diff
diff --git a/pkg/oci/oci.go b/pkg/oci/oci.go
index 9ae4d6f24..0d17cd58e 100644
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -690,25 +690,6 @@ func checkPayloadImage(payloadBytes []byte, imageDigest string) error {
 	return nil
 }
 
-func verifyImage(ctx context.Context, image string, imgOpts *ImageOptions) error {
-	imageStore, err := GetLocalOciStore()
-	if err != nil {
-		return fmt.Errorf("getting local oci store: %w", err)
-	}
-
-	imageRef, err := NormalizeImageName(image)
-	if err != nil {
-		return fmt.Errorf("normalizing image name: %w", err)
-	}
-
-	imageDigest, err := getImageDigest(ctx, imageStore, imageRef.String())
-	if err != nil {
-		return fmt.Errorf("getting image digest: %w", err)
-	}
-
-	return VerifyImage(ctx, imageRef, imageDigest, imgOpts)
-}
-
 // VerifyImage is used to verify the image corresponding to the reference and
 // digest with the given public key.
 // If the image was signed with the given public key, nothing is returned,
@@ -827,12 +808,17 @@ func ensureImage(ctx context.Context, imageStore oras.Target, image string, imgO
 		return nil
 	}
 
-	err := verifyImage(ctx, image, imgOpts)
+	imageRef, err := NormalizeImageName(image)
+	if err != nil {
+		return fmt.Errorf("normalizing image name: %w", err)
+	}
+
+	imageDigest, err := getImageDigest(ctx, imageStore.(*oci.Store), imageRef.String())
 	if err != nil {
-		return fmt.Errorf("verifying image %q: %w", image, err)
+		return fmt.Errorf("getting image digest: %w", err)
 	}
 
-	return nil
+	return VerifyImage(ctx, imageRef, imageDigest, imgOpts)
 }
 
 // EnsureImage ensures the image is present in the local store
```

> I can rename the verifyImage() to verifyOCIImage() or verifyGadgetImage().
This way, we have the generic VerifyImage() and the specialized verifyBuilderImage() and verifyGadgetImage() which rely on the first one.
What do you think?

That can work as well, but TBH I don't see a lot of value on having `verifyImage`, it's used in a single place and just handle a couple of operations that can be done directly be the caller.

cmd/common/image/build.go
mauriciovasquezbernal: I understand it, however it's a bit weird to verify this image each time it's run, I think the normal workflow is to verify the image only when it's pulled, after that, we can trust it not gonna be changed on the system. 
The issue here is that the user could run "docker pull ..." by their own, and then our verification won't work...

I suppose there are not other alternatives than verifying it each time.

cmd/common/image/build.go
mauriciovasquezbernal: I don't think `oci.DefaultAuthFile` is right, these are the credentials to pull the builder image, and `oci.DefaultAuthFile` should store credentials to pull gadgets, which could be totally different. What happens if it's left empty to fallback to the docker credentials?

https://github.com/inspektor-gadget/inspektor-gadget/blob/8f7cb770488817dc5e51d56720761f30e95ea5a4/pkg/oci/oci.go#L523-L534

mauriciovasquezbernal: Alright, the logic there checks if AuthFile is DefaultAuthFile:

```golang
		// If the AuthFile was not set explicitly, we allow to fall back to the docker auth,
		// otherwise we fail to avoid masking an error from the user
		if !errors.Is(err, os.ErrNotExist) || authOptions.AuthFile != DefaultAuthFile {
			return nil, fmt.Errorf("opening auth file %q: %w", authOptions.AuthFile, err)
		}
```

IMO there should also be some logic to fallback to the default auth file if it's not set, but I don't want to deviate this PR, so using `AuthFile: oci.DefaultAuthFile,` is fine for me. Perhaps just add a comment saying it's needed by the library in order to work or something like that.

cmd/common/image/build.go
mauriciovasquezbernal: If I read this correctly, verifying the builder image costs more than building the gadget. IMO it's too much. I'm wondering if we can implement this in a different way:
- Pull and verify the image once on the CI job 
- Document that the builder image is signed and that users could verify it manually if they want 

@alban any comments on this?

mauriciovasquezbernal: > I am not sure what you meant by this, as the image is already signed in the CI.

I mean to verify the image when building the gadgets on the CI job. It's true that it's signed by one job, but others jobs using it should verify it as well.

> Despite the cost being high, I still think we should go for it, or at least verifying the image when pulled and having users use the image though digest.
Indeed, using a verified builder image ensures the gadget are built correctly and without "troubles" added to them.

I pretty much agree with all of this, but still seems off to me to have to verify the builder image each time a gadget  is built, spending a lot of cpu cycles for nothing :(. IMO we should try to verify the image only at pull time: If the users pull the image by themselves, then it's their responsibility to verify it. I'm not sure however if using digest is a mess and difficult to implement.

> Do you have numbers on time taken to build an image with WASM?

I've noticed it's much slower but I don't have numbers.

mauriciovasquezbernal: > Digest corresponding to latest.

Currently we hardcode `latest` in 

https://github.com/inspektor-gadget/inspektor-gadget/blob/fd01fbcd4d7b95eb02caacbd4dfa421369845c16/cmd/common/image/build.go#L52

how would it work with digests? 

> Numbers would be welcomed, so we can take a better decision knowing all the elements.

I don't think the wasm stuff should be taken into consideration here, those are independent things and one shouldn't affect the other. Even if building the wasm module is expensive, it's not a reason to have an expensive step to verify the container image.

mauriciovasquezbernal: Take into consideration that many (possibly the majority of) gadgets won't use wasm.

alban: > If users do not have the requirements to build them using --local and they do not already have the image, they would need to pull it, so connection would be required.

My understanding of the previous discussion was that the verification process required connectivity even if the container image was already in the Docker store. Was it incorrect?

> I do not understand:
> 
>  1. You can perfectly give the digest to --builder-image.
> 2. In this case, the verification should occur with ig?

You can pass a digest to `--builder-image` but at the moment, the semantic is to designate the image to use. It is not meant to say "I trust this digest, no need to verify the cryptographic signatures".

Issue #2956, opened 2024-05-31
'k8s.io/foo@v0.0.0: invalid version: unknown revision v0.0.0' when running "go list -u -m -json all"


mqasimsarfraz:
It seems after https://github.com/inspektor-gadget/inspektor-gadget/pull/2794 if we run `go list -u -m -json all` (my IDE is running this) it fails with error:

```
go: k8s.io/cluster-bootstrap@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/endpointslice@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kube-aggregator@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kube-controller-manager@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kube-proxy@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kubectl@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/legacy-cloud-providers@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/metrics@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/pod-security-admission@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/sample-apiserver@v0.0.0: invalid version: unknown revision v0.0.0
```

This is mainly because the change introduced `k8s.io/kubernetes` as a dependency (indirect) and as [mentioned](https://github.com/kubernetes/kubernetes?tab=readme-ov-file#to-start-using-k8s) it isn't recommented:
> To use Kubernetes code as a library in other applications, see the [list of published components](https://git.k8s.io/kubernetes/staging/README.md). Use of the k8s.io/kubernetes module or k8s.io/kubernetes/... packages as libraries is not supported.

But we can workaround this by https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505627280 (i.e. we should use `replace` to avoid such issue). Having said that following fixes the issue for me:

```
replace (
  k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.30.1
  k8s.io/endpointslice => k8s.io/endpointslice v0.30.1
  k8s.io/kube-proxy => k8s.io/kube-proxy v0.30.1
  k8s.io/kubectl => k8s.io/kubectl v0.30.1
[...]

PR #2959, opened 2024-06-03
container-collection: Don't rely on k8s.io/kubernetes/pkg/...

mqasimsarfraz:
As mentioned in Kubernetes readme [1], use of the k8s.io/kubernetes module or k8s.io/kubernetes/... packages as libraries is not supported and can lead to werid issues like `go list -u -m -json all` failures. So, this change removes it.

[1] - https://github.com/kubernetes/kubernetes?tab=readme-ov-file#to-start-using-k8s

Fixes: #2956


dorser:
I just saw the related issue and looked into it. Seems to be working properly for me.
Thanks!

mqasimsarfraz:
Thanks for the reviews!

Issue #2424, opened 2024-01-29
trace exec: map gets full and it stops working


alban:
## Description

The "execs" ebpf map gets full under a specific workload and then the gadget stops working.

## Impact

The gadget stops giving events.

## Environment and steps to reproduce

goexecve.go:
```go
package main

import (
	//"fmt"
	"os"
	"runtime"
	"syscall"
)

func main() {
	runtime.LockOSThread()
	//fmt.Printf("main: pid=%d tid=%d\n", syscall.Getpid(), syscall.Gettid())
	go func() {
		runtime.LockOSThread()
		//fmt.Printf("go: pid=%d tid=%d\n", syscall.Getpid(), syscall.Gettid())

		args := []string{"cat", "/dev/null"}
		env := os.Environ()

		binary := "/bin/cat"
		err := syscall.Exec(binary, args, env)
		if err != nil {
			panic(err)
		}
	}()
	select {}
}
```
Makefile:
```Makefile
all: build

build:
	go build -o goexecve goexecve.go

run: build
	docker run -ti --rm --name goexecve -v $$PWD:/opt ubuntu \
		sh -c 'for i in $$(seq 10240); do /opt/goexecve; done'

ig:
	sudo -E ./ig trace exec -o columns=pid,ppid,comm,ret,args -c goexecve
```

In two different terminals:
* `make ig`
* `make run`

## Expected behavior

Map entries get clean up in `tracepoint/syscalls/sys_exit_execve`.

## Additional information

Counting entries in the map
```
sudo bpftool map dump name execs|grep key|wc -l
```

alban:
The workload triggering the issue is a multi-threaded process and the thread performing the execve is not the thread group leader.

According to [man clone(2)](https://man7.org/linux/man-pages/man2/clone.2.html):

> If any of the threads in a thread group performs an
> [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html), then all threads other than the thread group
> leader are terminated, and the new program is executed in
> the thread group leader.

In the current implementation, the execs map is keyed by the thread id without checking the thread group leader. When the execve is successful, the `tracepoint/syscalls/sys_enter_execve` and the `tracepoint/syscalls/sys_exit_execve` are performed by threads with a different id. When the execve fails, the two tracepoints are performed by the same thread.

PR #2965, opened 2024-06-04
gadgets/trace_exec: extract fixup for execve as a common header

i-Pear:
# gadgets/trace_exec: extract fixup for execve as a common header

`execve` is being traced in many gadgets such as `trace_exec` and `trace_capabilities`. Due to the `execve` syscall may change pid of threads, we need some fixup to get the caller's pid, so that we can share data between syscall enter and exit. This patch extracts the fixup in `trace_exec` as a common header.


alban (approved):
LGTM!

Could you use `exec_fixes.h` in `pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c` too?

And for `pkg/container-hook/bpf/execruntime.bpf.c`, this is fine to keep as is for now because the logic is slightly different.

alban:
`gadget_enter_exec();` is missing in the new commit.

gadgets/trace_exec/program.bpf.c
```diff
@@ -218,36 +181,20 @@ static __always_inline bool has_upper_layer()
...
 		return 0;
-	id = bpf_get_current_pid_tgid();
-	pid = (pid_t)id;
-	tgid = id >> 32;
-	ret = ctx->ret;
 
-	pid_sum = bpf_map_lookup_elem(&pid_by_tgid, &tgid);
-	if (!pid_sum)
+	ret = ctx->ret;
+	execs_lookup_key = gadget_get_exec_caller_pid(ret);
```
alban, line 193: I had a look at the uid filter:
```c
if (valid_uid(targ_uid) && targ_uid != uid)
		return 0;
```
We use this filter both in `tracepoint/syscalls/sys_enter_execve` and `tracepoint/syscalls/sys_exit_execve`. If the uid changes in between, `gadget_get_exec_caller_pid` will not clean the map properly.

I tried with:
```
sudo -E ig run ghcr.io/inspektor-gadget/gadget/trace_exec:latest --verify-image=false --host --uid 1000
```
And run `/bin/mount` as regular user 1000. `/bin/mount` is a setuid executable, so I was expecting to trigger a bug.
However `bpf_get_current_uid_gid` is returning the _real_ credentials and not the _effective_ credentials. Setuid executables only change the _effective_ credentials, so no bug were triggered. Applications have to call `setreuid` to change the _real_ credentials. _Real_ credentials cannot change in the middle of an execve because on Linux credentials are a property of threads and not processes.

The reason why it works is complicated, so I'm thinking the filter should be removed in the `tracepoint/syscalls/sys_exit_execve`, or at least move `gadget_get_exec_caller_pid` before the filter.

alban (approved):
LGTM based on code review, but I have not tested it yet.

alban:
Tested using test from #2424
- [x] builtin trace exec
- [x] image-based trace exec

gadgets/trace_exec/program.bpf.c
i-Pear: Opened #3029

PR #2966, opened 2024-06-04
container-collection: add process metadata enricher

### LINKED ISSUE #2961: [RFE] K8s Enrichment: Container Start Time
We would like to have an additional field in k8s enrichment: Container start time. AFAIK we can use the proc/x file created time.

@alban

dorser:
# Add Process Metadata Enricher
Enrich events with the container started at time retrieved from `/proc/<PID>/stat`. This needs to be done as the runtime clients do not provide the started at data for APIs other than inspect.
This resolves #2961 

## How to use
I added this works to the local manager and the made the data available for any gadget.
`ig trace exec -o json`

## Testing done
`ig trace exec -o json`

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -910,3 +910,26 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
 		return nil
 	}
 }
+
+// WithProcEnrichment enables an enricher to add process metadata
+func WithProcEnrichment() ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
+			pid := int(container.Pid)
+			if pid == 0 {
+				log.Errorf("proc enricher: failed to enrich container %s with pid zero", container.Runtime.ContainerID)
+				return false
+			}
+
+			procStat, err := host.GetProcStat(pid)
+			if err != nil {
+				log.Errorf("proc enricher: failed to read /proc/%d/stat for container %s: %v", pid, container.Runtime.ContainerID, err)
+				return false
+			}
+
+			container.Runtime.ContainerStartedAt = procStat.StartedAt
+			return true
+		})
+		return nil
+	}
+}
--- a/pkg/ig-manager/ig-manager.go
+++ b/pkg/ig-manager/ig-manager.go
@@ -100,6 +100,7 @@ func NewManager(runtimes []*containerutilsTypes.RuntimeConfig) (*IGManager, erro
 		containercollection.WithMultipleContainerRuntimesEnrichment(runtimes),
 		containercollection.WithContainerFanotifyEbpf(),
 		containercollection.WithTracerCollection(l.tracerCollection),
+		containercollection.WithProcEnrichment(),
 	}
 
 	if !log.IsLevelEnabled(log.DebugLevel) && isDefaultContainerRuntimeConfig(runtimes) {
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -35,6 +35,7 @@ func init() {
 	columns.MustRegisterTemplate("container", "width:30")
 	columns.MustRegisterTemplate("containerImageName", "width:30")
 	columns.MustRegisterTemplate("containerImageDigest", "width:30")
+	columns.MustRegisterTemplate("containerStartedAt", "width:35,hide")
 	columns.MustRegisterTemplate("comm", "maxWidth:16")
 	columns.MustRegisterTemplate("pid", "minWidth:7")
 	columns.MustRegisterTemplate("uid", "minWidth:8")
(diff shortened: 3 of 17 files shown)


pkg/container-collection/options.go
```diff
@@ -910,3 +911,32 @@ func WithTracerCollection(tc TracerCollection) ContainerCollectionOption {
...
+func WithProcEnrichment() ContainerCollectionOption {
+	return func(cc *ContainerCollection) error {
+		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
+			pid := int(container.Pid)
+			if pid == 0 {
+				log.Errorf("proc enricher: failed to enrich container %s with pid zero", container.Runtime.ContainerID)
+				return true
+			}
+
+			p, err := process.NewProcess(int32(pid))
```
alban, line 925: This package uses unix signals to check if the process exists:
https://github.com/shirou/gopsutil/blob/master/process/process_posix.go#L121

And unix signals only work if the process is in the same pid namespace as ig.

This also uses `/proc`. If ig is running in a container, we would need to define the `$HOST_PROC` environment variable to make it use `/host/proc`:
https://github.com/shirou/gopsutil/blob/ac15d685454b02de6fbfddd95fecaf3e50fd45f7/internal/common/common.go#L392

I wonder if it's better to avoid another dependency and just do a `stat` syscall: 
```
stat --printf=%y /proc/$$
```

dorser: Makes sense. It was too late when I fixed that yesterday, I'll publish a new version using the stat syscall shortly.

pkg/ig-manager/ig-manager.go
```diff
@@ -100,6 +100,7 @@ func NewManager(runtimes []*containerutilsTypes.RuntimeConfig) (*IGManager, erro
 		containercollection.WithMultipleContainerRuntimesEnrichment(runtimes),
 		containercollection.WithContainerFanotifyEbpf(),
 		containercollection.WithTracerCollection(l.tracerCollection),
+		containercollection.WithProcEnrichment(),
```
alban, line 103: I think it should also be added in:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgettracermanager/gadgettracermanager.go#L270

So it works with kubectl-gadget.

pkg/types/types.go
```diff
@@ -138,6 +138,9 @@ type BasicRuntimeMetadata struct {
 	// containerd: events from both initial and new containers are enriched
 	// crio: events from initial containers are enriched
 	ContainerImageDigest string `json:"containerImageDigest,omitempty" column:"containerImageDigest,hide"`
+
+	// ContainerStartedAt is the unix timestamp at which the container was started at
+	ContainerStartedAt int64 `json:"containerStartedAt,omitempty" column:"containerStartedAt,hide"`
```
alban, line 143: It should use the `types.Time` type so that the columns output shows human-readable time.

See also:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/types/types.go#L64-L70

pkg/utils/host/host.go
```diff
@@ -57,6 +58,10 @@ type Config struct {
 	AutoMountFilesystems bool
 }
 
+type ProcStat struct {
+	StartedAt int64
+}
```
alban, line 63: Do you foresee other fields we might want from procfs?
I wonder if it's necessary to have this struct since it is only one field. But if there are more later, then it is fine.

alban (commented):
Thanks!

pkg/types/types.go
dorser: No problem. Although, I'm wondering if maybe we want to convert this to something human-readable only when we output columns. WDYT?

pkg/utils/host/host.go
dorser: I believe stuff like permissions and owner can be of interest but let me take that back to the team and verify.

pkg/types/types.go
alban: I don't know but I think it should be consistent with other timestamps.

Any other opinions?

pkg/utils/host/host.go
```diff
@@ -175,3 +182,17 @@ func GetProcCmdline(pid int) []string {
 	cmdlineBytes, _ := os.ReadFile(filepath.Join(HostProcFs, pidStr, "cmdline"))
 	return strings.Split(string(cmdlineBytes), "\x00")
 }
+
+func GetProcStat(pid int) (ProcStat, error) {
+	pidStr := fmt.Sprint(pid)
+	var stat syscall.Stat_t
+	if err := syscall.Stat(filepath.Join(HostProcFs, pidStr), &stat); err != nil {
+		return ProcStat{}, fmt.Errorf("reading /proc/%d/stat: %w", pid, err)
```
alban, line 190: The path in the error message does not match the path used in the `Stat()` call (`/host` prefix and `stat` suffix)
```suggestion
	path := filepath.Join(HostProcFs, fmt.Sprint(pid))
	if err := syscall.Stat(path, &stat); err != nil {
		return ProcStat{}, fmt.Errorf("reading %s: %w", path, err)
```

alban (approved):
One error message to fix, LGTM after that (if the CI is green after rebase).

PR #2980, opened 2024-06-10
uprobetracer: use secureopen to avoid TOCTOU issues

alban:
uprobetracer needs to open files in the container rootfs. The file paths in the container could change, so special care should be taken to avoid symlink escapes.

Requires Linux 5.6 for openat2 (https://github.com/torvalds/linux/commit/fddb5d430ad9fa91b49b1d34d0202ffe2fa0e179).

cc @rata @i-Pear 

## How to use

No changes

## Testing done

- [x] Test if the trace_ssl gadget still works.

### DIFF
--- a/pkg/uprobetracer/ldcache_parser.go
+++ b/pkg/uprobetracer/ldcache_parser.go
@@ -18,9 +18,10 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 	"unsafe"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/secureopen"
 )
 
 type ldCache1Entry struct {
@@ -134,17 +135,10 @@ func readCacheFormat2(data []byte) []ldEntry {
 // simulate the loader's behaviour, find library path in containers' `/etc/ld.so.cache`.
 // see https://github.com/bminor/glibc/blob/master/elf/cache.c#L292, the `print_cache` func
 // see https://github.com/iovisor/bcc/blob/master/src/cc/bcc_proc.c#L508, the `bcc_procutils_which_so` func
-func parseLdCache(ldCachePath string, libraryName string) ([]string, error) {
-	fileInfo, err := os.Stat(ldCachePath)
+func parseLdCache(containerPid uint32, ldCachePath string, libraryName string) ([]string, error) {
+	ldCacheFile, err := secureopen.ReadFileInContainer(containerPid, ldCachePath)
 	if err != nil {
-		return nil, fmt.Errorf("stat file %q: %w", ldCachePath, err)
-	}
-	if !fileInfo.Mode().IsRegular() {
-		return nil, fmt.Errorf("ldCache file %q is not regular", ldCachePath)
-	}
-	ldCacheFile, err := os.ReadFile(ldCachePath)
-	if err != nil {
-		return nil, fmt.Errorf("reading file %q: %w", ldCachePath, err)
+		return nil, fmt.Errorf("reading file %q in container pid %d: %w", ldCachePath, containerPid, err)
 	}
 	ldCacheFileSize := uint32(len(ldCacheFile))
 
--- a/pkg/uprobetracer/tracer.go
+++ b/pkg/uprobetracer/tracer.go
@@ -42,12 +42,12 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
-	securejoin "github.com/cyphar/filepath-securejoin"
 
 	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/kfilefields"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/secureopen"
 )
 
 type ProgType uint32
@@ -167,32 +167,20 @@ func (t *Tracer[Event]) AttachProg(progName string, progType ProgType, attachTo 
 }
 
 func (t *Tracer[Event]) searchForLibrary(containerPid uint32) ([]string, error) {
-	var targetPaths []string
-	var securedTargetPaths []string
+	var unsecurerTargetPaths []string
 
 	filePath := t.attachFilePath
 	if !filepath.IsAbs(filePath) {
-		containerLdCachePath, err := securejoin.SecureJoin(filepath.Join(host.HostProcFs, fmt.Sprint(containerPid), "root"), "etc/ld.so.cache")
-		if err != nil {
-			return nil, fmt.Errorf("path %q: %w", filePath, err)
-		}
-		ldCachePaths, err := parseLdCache(containerLdCachePath, filePath)
+		ldCachePath := "/etc/ld.so.cache"
+		ldCachePaths, err := parseLdCache(containerPid, ldCachePath, filePath)
 		if err != nil {
 			return nil, fmt.Errorf("parsing ld cache: %w", err)
 		}
-		targetPaths = ldCachePaths
+		unsecurerTargetPaths = ldCachePaths
 	} else {
-		targetPaths = append(targetPaths, filePath)
+		unsecurerTargetPaths = append(unsecurerTargetPaths, filePath)
 	}
-	for _, targetPath := range targetPaths {
-		securedTargetPath, err := securejoin.SecureJoin(filepath.Join(host.HostProcFs, fmt.Sprint(containerPid), "root"), targetPath)
-		if err != nil {
-			t.logger.Debugf("path %q in ld cache is not available: %s", filePath, err.Error())
-			continue
-		}
-		securedTargetPaths = append(securedTargetPaths, securedTargetPath)
-	}
-	return securedTargetPaths, nil
+	return unsecurerTargetPaths, nil
 }
 
 // attach uprobe program to the inode of the file passed in parameter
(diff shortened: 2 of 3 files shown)


pkg/utils/secureopen/secureopen.go
```diff
@@ -0,0 +1,87 @@
...
+	err = unix.Fstat(fd, &stat)
+	if err != nil {
+		return nil, fmt.Errorf("fstat %s in %s: %w", unsafePath, root, err)
+	}
+	if stat.Mode&unix.S_IFMT != unix.S_IFREG {
+		return nil, fmt.Errorf("procfd stat: not a regular file")
+	}
+
+	// Re-open in read-only mode (without O_PATH)
+	procfd := filepath.Join("/proc/self/fd", strconv.Itoa(int(fd)))
```
i-Pear, line 72: Use `host.HostProcFs` instead of `/proc` ?

alban: For `/proc/self`, it is better to use `/proc` directly because:
- `self` is normally available in the normal `/proc` (the current ig process lives in the same (or a child) pid namespace as the pid namespace for the `/proc` mount.
- `/host/proc/self` might not be available in some configurations. In particular on WSL2, when ig is executed directly from a shell on the `ubuntu` WSL2 Linux distribution but the containers are running in Docker Desktop (in the `docker-desktop` WSL2 Linux distribution). Because in that case, `/host/proc` is referring to a pid namespace which is not the current nor a parent pid namespace, but a sibling. See drawing on #2875.

pkg/utils/secureopen/secureopen.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Open with O_PATH first
+	howOPath := unix.OpenHow{
+		Flags:   unix.O_PATH,
+		Mode:    0,
+		Resolve: unix.RESOLVE_IN_ROOT | unix.RESOLVE_NO_MAGICLINKS,
+	}
+	fd, err := unix.Openat2(int(rootDir.Fd()), unsafePath, &howOPath)
+	if err != nil {
+		return nil, fmt.Errorf("openat2 %s in %s: %w", unsafePath, root, err)
```
rata, line 57: nit (and if you disagree, just ignore it). But for this and all the other file printing in errors, I'd use `%q` that print the string between quotes.

```suggestion
		return nil, fmt.Errorf("openat2 %q in %q: %w", unsafePath, root, err)
```

pkg/utils/secureopen/secureopen.go
```diff
@@ -0,0 +1,87 @@
...
+// https://github.com/torvalds/linux/commit/fddb5d430ad9fa91b49b1d34d0202ffe2fa0e179
+func OpenInContainer(containerPid uint32, unsafePath string) (*os.File, error) {
+	root := filepath.Join(host.HostProcFs, fmt.Sprint(containerPid), "root")
+	rootDir, err := os.OpenFile(root, unix.O_PATH, 0)
+	if err != nil {
+		return nil, fmt.Errorf("open o_path %s: %w", root, err)
+	}
+	defer rootDir.Close()
+
+	// Open with O_PATH first
```
rata, line 49: IMHO we should either explain why we do this (or remove the comment, but it doesn't seem to add much value like this).

pkg/utils/secureopen/secureopen.go
```diff
@@ -0,0 +1,87 @@
...
+		Mode:    0,
+		Resolve: unix.RESOLVE_IN_ROOT | unix.RESOLVE_NO_MAGICLINKS,
+	}
+	fd, err := unix.Openat2(int(rootDir.Fd()), unsafePath, &howOPath)
+	if err != nil {
+		return nil, fmt.Errorf("openat2 %s in %s: %w", unsafePath, root, err)
+	}
+	defer unix.Close(fd)
+
+	// Check if the path is a regular file
```
rata, line 61: This, with the func documentation and such straight-forward code, feels a little repetitive for me.

pkg/utils/secureopen/secureopen.go
```diff
@@ -0,0 +1,87 @@
...
+	"path/filepath"
+	"strconv"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
+)
+
+// OpenInContainer opens the given path in the given container in read-only
+// mode.
```
rata, line 32:
```suggestion
// OpenInContainer opens the given path in the given container referenced by containerPid in read-only
// mode.
```

btw, this is more generic than containers (this works for any process), but I think leaving it for containers is okay here.

pkg/uprobetracer/tracer.go
```diff
@@ -167,32 +167,20 @@ func (t *Tracer[Event]) AttachProg(progName string, progType ProgType, attachTo
 }
 
 func (t *Tracer[Event]) searchForLibrary(containerPid uint32) ([]string, error) {
-	var targetPaths []string
-	var securedTargetPaths []string
+	var unsecurerTargetPaths []string
```
rata, line 170: in the function attach you use `unsecured` as prefix insted of `insecurer`, as you use here. Is this a typo or on purpose? The former seems nicer to me.

rata (approved):
LGTM. 

From a safe POV, the new package seems to handle things safely (and it's quite elegant doing so, awesome!) and usage of the added functions seems safe too.

I added some VERY nit comments, mostly on comments :D. I was doubting if submitting them, but as all the important stuff is fine, it doesn't seem to add noise. Feel free to ignore them too :)

alban:
Thanks for the reviews! I updated the branch.

rata (approved):
LGTM! Great improvement :)

pkg/uprobetracer/tracer.go
```diff
@@ -225,20 +213,20 @@ func (t *Tracer[Event]) attachUprobe(file *os.File) (link.Link, error) {
...
-	if len(attachFilePaths) == 0 {
+	if len(unsecuredAttachFilePaths) == 0 {
 		t.logger.Debugf("cannot find file to attach in container %d for symbol %q", containerPid, t.attachSymbol)
 	}
 
-	for _, filePath := range attachFilePaths {
+	for _, filePath := range unsecuredAttachFilePaths {
 		// Do not use `O_PATH` flag here, because `ReadRealInodeFromFd` needs the `private_data` field
 		// in kernel "struct file", to access the underlying inode through overlayFS.
 		// Using `O_PATH` flag will cause the `private_data` field to be zero.
```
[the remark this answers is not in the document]
alban: I think it is still valid.
- `secureopen.OpenInContainer` internally uses `O_PATH` as an implementation detail but returns a file descriptor opened without `O_PATH`.
- `kfilefields.ReadRealInodeFromFd` requires the fd to be opened without `O_PATH`.

I could change the comment to this?

```suggestion
		// Thankfully, OpenInContainer returns a fd opened without `O_PATH`. This
		// is necessary because `ReadRealInodeFromFd` needs the `private_data` field
		// in kernel "struct file", to access the underlying inode through overlayFS.
		// Using `O_PATH` flag will cause the `private_data` field to be zero.
```

pkg/uprobetracer/ldcache_parser.go
```diff
@@ -134,17 +135,10 @@ func readCacheFormat2(data []byte) []ldEntry {
 // simulate the loader's behaviour, find library path in containers' `/etc/ld.so.cache`.
 // see https://github.com/bminor/glibc/blob/master/elf/cache.c#L292, the `print_cache` func
 // see https://github.com/iovisor/bcc/blob/master/src/cc/bcc_proc.c#L508, the `bcc_procutils_which_so` func
-func parseLdCache(ldCachePath string, libraryName string) ([]string, error) {
-	fileInfo, err := os.Stat(ldCachePath)
-	if err != nil {
-		return nil, fmt.Errorf("stat file %q: %w", ldCachePath, err)
-	}
-	if !fileInfo.Mode().IsRegular() {
```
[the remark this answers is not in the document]
alban: I didn't find it possible to use the `fileInfo.Mode().IsRegular()` API because it cannot operate on a file descriptor like `lstat(2)` or `statx(2)`.

alban:
Thanks. I will merge once the CI completes.

alban:
I don't understand why the CI fails on the integration tests `TestRunTraceTcp` for docker/containerd/cri-o. The tests are not related to uprobes.

pkg/utils/secureopen/secureopen.go
```diff
@@ -0,0 +1,88 @@
...
+// OpenInContainer opens the given path in the given container referenced by
+// containerPid in read-only mode.
+//
+// The resulting open file is guaranteed to be:
+// - inside the provided container
+// - a regular file
+// - without following magic links from procfs
+//
+// Requires Linux 5.6 for openat2:
+// https://github.com/torvalds/linux/commit/fddb5d430ad9fa91b49b1d34d0202ffe2fa0e179
```
mauriciovasquezbernal, line 40: Perhaps add a small description indicating how it works? (i.e. mentioning it relies on openat2 and unix.RESOLVE_IN_ROOT | unix.RESOLVE_NO_MAGICLINKS`)

pkg/uprobetracer/tracer.go
```diff
@@ -167,32 +167,20 @@ func (t *Tracer[Event]) AttachProg(progName string, progType ProgType, attachTo
 }
 
 func (t *Tracer[Event]) searchForLibrary(containerPid uint32) ([]string, error) {
```
mauriciovasquezbernal, line 169: I'm wondering if this implementation is easier to follow:

```golang
func (t *Tracer[Event]) searchForLibrary(containerPid uint32) ([]string, error) {
	filePath := t.attachFilePath
	if filepath.IsAbs(filePath) {
		return []string{filePath}, nil
	}

	ldCachePath := "/etc/ld.so.cache"
	ldCachePaths, err := parseLdCache(containerPid, ldCachePath, filePath)
	if err != nil {
		return nil, fmt.Errorf("parsing ld cache: %w", err)
	}
	return ldCachePaths, nil
}
```

mauriciovasquezbernal (approved):
Only two minor nits, but LGTM already.

PR #2983, opened 2024-06-10
execsnoop: add parent upper layer

### LINKED ISSUE #2962: [RFE] [trace exec] Parent_UpperLayer
Currently trace exec events emit `UpperLayer: bool` which is great. We would like to further augment the event by also adding the processes' parent upperLayer (i.e is the **parent** process part of the original image layer/overlayfs)

@alban

dorser:
# execsnoop: add parent upper layer

Adding parent upper layer to indicate whether the executable's parent is on the upper layer of overlayfs.

## How to use

```
$ sudo -E ./ig run --verify-image=false trace_exec:parent-upper-layer -o json
{"args":"/usr/bin/sh",REDACTED,"pupper_layer":false,"upper_layer":false}
{"args":"/usr/bin/cp",REDACTED,"pupper_layer":false,"upper_layer":false}
{"args":"/usr/bin/sh2",REDACTED,"pupper_layer":false,"upper_layer":true}
{"args":"/bin/echo",REDACTED,"pupper_layer":true,"upper_layer":false}
```

```
$ docker run -ti --rm --name=test ubuntu \                               ✔ │ cloudbox │ 18:14:48
    sh -c 'cp /bin/sh /bin/sh2; sh2 -c "/bin/echo pupper"'
pupper
```

Looking at the two last events, we see that sh2 is in the upper layer and for the echo command, we can see its parent, sh2, is in the upper layer.

## Testing done
See above

## Limitations
Similar to `upper_layer`. See: #2353

## Misc
Resolves #2962


gadgets/Makefile
```diff
@@ -37,6 +37,7 @@ build: $(GADGETS)
 $(GADGETS):
 	@echo "Building $@"
 	@sudo -E IG_EXPERIMENTAL=true $(IG) image build \
+		--update-metadata \
```
alban, line 40: It does not belong to this PR.

I am fine with adding it, I tried to add it in another PR, but AFAIR @mauriciovasquezbernal preferred to leave it out. Meanwhile, you can use the command from line  35.

gadgets/trace_exec/gadget.yaml
```diff
@@ -79,6 +79,12 @@ structs:
         width: 5
         alignment: left
         ellipsis: end
+    - name: pupper_layer
+      description: Whether the executable's parent is in the upper layer of the overlay filesystem
```
alban, line 83: Suggestion:
```suggestion
      description: Whether the executable's parent in the process hierarchy is in the upper layer of the overlay filesystem. Only initialized when the execution succeeded.
```
Ditto for upper_layer
Ditto in the documentation.

alban (commented):
It looks good in general, just a couple of small changes.

integration/k8s/trace_exec_test.go
```diff
@@ -29,9 +29,10 @@ func TestTraceExec(t *testing.T) {
 	t.Parallel()
 	ns := GenerateTestNamespaceName("test-trace-exec")
 
-	cmd := "cp /bin/date /date ; setuidgid 1000:1111 sh -c 'while true; do /date ; /bin/sleep 0.1; done'"
+	innerCmd := "while true ; do /bin/sleep 0.1 ; done"
+	cmd := fmt.Sprintf("cp /bin/sh /usr/bin/sh ; setuidgid 1000:1111 /usr/bin/sh -c '%s'", innerCmd)
```
dorser, line 33: @alban changed the test as suggested to make sure we test both upper layer and parent upper layer fields.

alban (approved):
LGTM, thanks!

Could you rebase and fix the conflicts? I will run the CI after that.

dorser:
> LGTM, thanks!
> 
> Could you rebase and fix the conflicts? I will run the CI after that.

Done! Thanks for the review!

Issue #2967, opened 2024-06-04
[trace exec] fanotify: too many open files / 


galofir-ms:
## Description

After running a process executing IG for more than a few hours, fanotify errors start to appear.

We've seen evidence of multiple errors:

![image](https://github.com/inspektor-gadget/inspektor-gadget/assets/146319678/76e16c03-3bab-4c70-9ab0-4dab4e76e66f)

## Impact
K8s enrichment stops working

@alban

alban:
I can reproduce this bug by running the following on the host:

```bash
mkdir -p  /tmp/test-runc/userdata
touch /tmp/test-runc/userdata/config.json
for i in `seq 128` ; do
    echo -n .
    runc create --bundle /tmp/test-runc --pid-file /tmp/test-runc/mypid &> /dev/null
done
echo
```

Then, ig outputs:
```
ERRO[0345] error monitoring runtime instance: fanotify: init error, too many open files 
```


I think `container-hook/tracer.go`/`parseOCIRuntime()` should have the following improvements:
- [ ] `n.monitorRuntimeInstance` should have a timeout: if runc didn't create the pid file in 2 seconds, don't monitor the directory forever. Or at least, when the runc process terminates without creating the pid file, cancel the monitoring.
- [ ] Check that the pid file is created by the correct pid of runc
- [ ] Different call instances of `parseOCIRuntime` could reuse the same fanotify group to avoid the limit [FANOTIFY_DEFAULT_MAX_GROUPS=128](https://github.com/torvalds/linux/blob/v6.6/fs/notify/fanotify/fanotify_user.c#L32)
- [ ] Ignore invalid `runc` commands (the call to `runc create` in my test script is not valid syntax since it does not have a container-id)

alban:
Instead of running  `runc create` manually, I can also trigger the bug by attempting to start a container with a bad configuration:
```
$ docker run --rm -it --name busybox -v /dev/null/null:/tmp/null busybox
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/dev/null/null" to rootfs at "/tmp/null": stat /dev/null/null: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type.
```

In this case, docker/containerd executes:
[...]

PR #2998, opened 2024-06-11
pkg/container-hook: add new test

alban:
This test runs a Docker container and checks it is correctly detected by the container-hook pkg.

This also runs Docker containers with invalid configuration to check if it makes IG misbehave.

There is a TODO in one test: it can only be removed after #2967 is fixed.

## How to use

No changes.

## Testing done

```
go test -exec 'sudo -E' ./pkg/container-hook/...
```

### DIFF
--- a/pkg/container-hook/tracer_test.go
+++ b/pkg/container-hook/tracer_test.go
@@ -1,0 +1,123 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build linux
+// +build linux
+
+package containerhook_test
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	containerhook "github.com/inspektor-gadget/inspektor-gadget/pkg/container-hook"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/testutils"
+)
+
+func TestContainerHook(t *testing.T) {
+	utilstest.RequireRoot(t)
+
+	type testDefinition struct {
+		generateEvent func(t *testing.T) string
+		validateEvent func(t *testing.T, info *utilstest.RunnerInfo, containerID string, events []containerhook.ContainerEvent)
+	}
+
+	for name, test := range map[string]testDefinition{
+		"one_container": {
+			generateEvent: generateEvent(0),
+			validateEvent: utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, containerID string) *containerhook.ContainerEvent {
+				return &containerhook.ContainerEvent{
+					Type:        containerhook.EventTypeAddContainer,
+					ContainerID: containerID,
+				}
+			}),
+		},
+		"one_container_after_some_failed_containers": {
+			generateEvent: generateEvent(2),
+			validateEvent: utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, containerID string) *containerhook.ContainerEvent {
+				return &containerhook.ContainerEvent{
+					Type:        containerhook.EventTypeAddContainer,
+					ContainerID: containerID,
+				}
+			}),
+		},
+		"one_container_after_many_failed_containers": {
+			// Test with a number bigger than FANOTIFY_DEFAULT_MAX_GROUPS
+			// https://github.com/torvalds/linux/blob/v6.9/fs/notify/fanotify/fanotify_user.c#L32
+			// #define FANOTIFY_DEFAULT_MAX_GROUPS	128
+			generateEvent: generateEvent(3), // TODO: change this number to 130
+			validateEvent: utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, containerID string) *containerhook.ContainerEvent {
+				return &containerhook.ContainerEvent{
+					Type:        containerhook.EventTypeAddContainer,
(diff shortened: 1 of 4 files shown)


pkg/container-hook/tracer_test.go
```diff
@@ -0,0 +1,130 @@
...
+		cmd.Stdout = &stdout
+		cmd.Stderr = &stderr
+		err := cmd.Run() // Avoid CombinedOutput because stderr might contain busybox pulling messages
+		if err != nil {
+			return "", fmt.Errorf("running container: %w\nStdout:\n%s\nStderr:\n%s",
+				err, string(stdout.Bytes()), string(stderr.Bytes()))
+		}
+		trimedOutput := strings.TrimSuffix(string(stdout.Bytes()), "\n")
+		return trimedOutput, nil
+	}
```
mauriciovasquezbernal, line 129: Perhaps we can use a call to the API instead of invoking the CLI?

We already have a library for it, I gave it a try:

```diff
diff --git a/pkg/container-hook/tracer_test.go b/pkg/container-hook/tracer_test.go
index b15b025d4..317c82067 100644
--- a/pkg/container-hook/tracer_test.go
+++ b/pkg/container-hook/tracer_test.go
@@ -18,11 +18,8 @@
 package containerhook_test

 import (
-       "bytes"
-       "errors"
        "fmt"
-       "os/exec"
-       "strings"
+       "math/rand"
        "testing"
        "time"

@@ -30,13 +27,14 @@ import (

        utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
        containerhook "github.com/inspektor-gadget/inspektor-gadget/pkg/container-hook"
+       "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/testutils"
 )

 func TestContainerHook(t *testing.T) {
        utilstest.RequireRoot(t)

        type testDefinition struct {
-               generateEvent func() (string, error)
+               generateEvent func(t *testing.T) string
                validateEvent func(t *testing.T, info *utilstest.RunnerInfo, stdoutStderr string, events []containerhook.ContainerEvent)
        }

@@ -63,7 +61,7 @@ func TestContainerHook(t *testing.T) {
                        // Test with a number bigger than FANOTIFY_DEFAULT_MAX_GROUPS
                        // https://github.com/torvalds/linux/blob/v6.9/fs/notify/fanotify/fanotify_user.c#L32
                        // #define FANOTIFY_DEFAULT_MAX_GROUPS  128
-                       generateEvent: generateEvent(3), // TODO: change this number to 130
+                       generateEvent: generateEvent(130), // TODO: change this number to 130
                        validateEvent: utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, stdoutStderr string) *containerhook.ContainerEvent {
                                return &containerhook.ContainerEvent{
                                        Type:        containerhook.EventTypeAddContainer,
@@ -91,40 +89,31 @@ func TestContainerHook(t *testing.T) {
                        require.Nil(t, err, "Returned error was not nil")
                        require.NotNil(t, notifier, "Returned notifier was nil")

-                       stdoutStderr, err := test.generateEvent()
-                       require.Nil(t, err, "Returned error was not nil")
+                       containerID := test.generateEvent(t)

                        // Give some time for the tracer to capture the events
                        time.Sleep(100 * time.Millisecond)

-                       test.validateEvent(t, nil, stdoutStderr, events)
+                       test.validateEvent(t, nil, containerID, events)
                })
        }
 }

-func generateEvent(nfailure int) func() (string, error) {
-       return func() (string, error) {
+func generateEvent(nfailure int) func(t *testing.T) string {
+       return func(t *testing.T) string {
                // Failed container with bad configuration
                for i := 0; i < nfailure; i++ {
-                       cmd := exec.Command("docker", "run", "--rm", "-v", "/dev/null:/dev/null/invalid", "busybox", "id")
-                       stdoutStderr, err := cmd.CombinedOutput()
-                       trimedOutput := strings.TrimSuffix(string(stdoutStderr), "\n")
-                       if err == nil {
-                               return trimedOutput, errors.New("container with invalid configuration should have failed")
-                       }
+                       fmt.Printf("running %d\n", i)
+                       name := fmt.Sprintf("ig-%d", rand.Uint32())
+                       // foo is not a valid command inside busybox, hence it fails
+                       container := testutils.NewDockerContainer(name, "foo", testutils.WithoutLogs())
+                       container.Run(t)
                }

                // Good container with valid configuration
-               var stdout, stderr bytes.Buffer
-               cmd := exec.Command("docker", "run", "--rm", "-d", "busybox", "id")
-               cmd.Stdout = &stdout
-               cmd.Stderr = &stderr
-               err := cmd.Run() // Avoid CombinedOutput because stderr might contain busybox pulling messages
-               if err != nil {
-                       return "", fmt.Errorf("running container: %w\nStdout:\n%s\nStderr:\n%s",
-                               err, string(stdout.Bytes()), string(stderr.Bytes()))
-               }
-               trimedOutput := strings.TrimSuffix(string(stdout.Bytes()), "\n")
-               return trimedOutput, nil
+               name := fmt.Sprintf("ig-%d", rand.Uint32())
+               container := testutils.NewDockerContainer(name, "id", testutils.WithoutLogs())
+               container.Run(t)
+               return container.ID()
        }
 }
```

However it doesn't work fine because the test with 130 failed containers passes.

I also tried with RunDockerFailedContainer, but it didn't work.

mauriciovasquezbernal (commented):
I think having this kind of tests is great, thanks!

Do you think it'll be possible to use the API instead of invoking the CLI? (to avoid depending on it and having to parse the output)

pkg/container-utils/testutils/options.go
```diff
@@ -60,12 +62,24 @@ func WithContext(ctx context.Context) Option {
...
+	}
+}
+
 func WithImage(image string) Option {
 	return func(opts *containerOptions) {
 		opts.image = image
 	}
 }
 
+func WithBindMounts(mounts []string) Option {
```
mauriciovasquezbernal, line 77: nit: Perhaps of passing a string we can define a new struct that contains source and target fields?

mauriciovasquezbernal (approved):
LGTM, thanks for handling the comments.

pkg/container-utils/testutils/options.go
alban: I don't know if it is worth it.
There is also a `Mount struct`:
https://github.com/moby/moby/blob/v26.1.4/api/types/mount/mount.go#L24-L31
But it is Docker specific and this `containerOptions struct` is used for both containerd and docker.

alban:
I added a test for `c.options.mounts` and `c.options.expectStartError` in `pkg/container-utils/testutils/containerd.go` to fail like it was done for other unsupported options.

PR #3008, opened 2024-06-12
Align k8s json and columns name

mauriciovasquezbernal:
Use podName (instead of pod) and containerName (instead of container) for both, json and columns output for the k8s fields.

This commit changes the header printed for the built-in gadgets from

$ ./kubectl-gadget trace exec
K8S.NODE          K8S.NAMESPACE     K8S.POD           K8S.CONTAINER     PID       PPID      COMM     PCOMM    RET ARGS

to

$ ./kubectl-gadget trace exec
K8S.NODE          K8S.NAMESPACE     K8S.PODNAME       K8S.CONTAINERNAME PID       PPID      COMM     PCOMM    RET ARGS

It's not a breaking change as users should be parsing the json output instead of the columns output.

### DIFF
--- a/cmd/common/utils/parser.go
+++ b/cmd/common/utils/parser.go
@@ -123,8 +123,8 @@ func GetKubernetesColumns() []string {
 	return []string{
 		"k8s.node",
 		"k8s.namespace",
-		"k8s.pod",
-		"k8s.container",
+		"k8s.podName",
+		"k8s.containerName",
 	}
 }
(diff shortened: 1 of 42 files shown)


alban (approved):
I think it would be nicer to use "pod" and "container" both for columns and json, but since that would be a breaking change, that would require more discussion.

LGTM for now, so you can unblock the CI issues.

PR #2912, opened 2024-05-29
Sign ig and helpers images

eiffel-fl:
Hi.

This PR sign `ig` and helpers images.

Best regards.

PR #3018, opened 2024-06-14
ci: Use jq to retrieve image digest instead of grep and awk.

eiffel-fl:
This is more reliable to use jq than bash piping to retrieve this information [1].

[1]: https://github.com/inspektor-gadget/inspektor-gadget/pull/2912#discussion_r1635496823


mauriciovasquezbernal (approved):
LGTM!

mauriciovasquezbernal (approved):
LGTM !

Issue #2996, opened 2024-06-11
gadgets: Finish implementing trace tcpdrop as image-based gadget


mauriciovasquezbernal:
- convert fields of original proc_ctx to struct (https://github.com/inspektor-gadget/inspektor-gadget/pull/2477#discussion_r1484672309)
- once we support bitfield to string conversion (https://github.com/inspektor-gadget/inspektor-gadget/issues/2482), update event->tcpflags type (https://github.com/inspektor-gadget/inspektor-gadget/pull/2477#discussion_r1483555007)

burak-ok:
> * convert fields of original proc_ctx to struct ([Add trace_tcpdrop image based gadget #2477 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/pull/2477#discussion_r1484672309))

Do I remember right that we still can't have substructs in eBPF.
Additionally we would have the `mount_ns_id` in the substruct. And that woud also mean that we have it twice. (1 per substruct). So enrichment for this is kinda hard?

mauriciovasquezbernal:
> Do I remember right that we still can't have substructs in eBPF.

We already have support for substructs. 

> Additionally we would have the mount_ns_id in the substruct. And that woud also mean that we have it twice. (1 per substruct). So enrichment for this is kinda hard?

Yeah, that's kind of an open question. @blanquicet added some warnings about this on #2877. Perhaps we should have a single field of this type and some logic to fill it from the source that is available.

burak-ok:
> > Do I remember right that we still can't have substructs in eBPF.
> 
> We already have support for substructs.

When I played around I didn't get (at least the json) output  for the substructs. But I didn't dive deep into it

mauriciovasquezbernal:
> convert fields of original proc_ctx to struct (https://github.com/inspektor-gadget/inspektor-gadget/pull/2477#discussion_r1484672309)

No other gadget is using a struct for common fields. I'll mark this as done and discuss it for all gadgets in a follow up issue.

PR #3020, opened 2024-06-14
Implement bitfield to string conversion

### LINKED ISSUE #2482: [RFE] run: Implement bitfield to string conversion
From the design document https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md#bitfield

> Convert bitfield from an event to a human readable string. Example: bind's [optionsToString](https://github.com/inspektor-gadget/inspektor-gadget/blob/b57f2bae31a46b40d8e0204b85099ae37f15d21d/pkg/gadgets/trace/bind/tracer/tracer.go#L154).

mauriciovasquezbernal:
This commit adds support for providing a human readable version of a bitfield.
This is done by using the same logic for enums, in this case, the gadget author has to define an enum with all values of the flags and that its name has the "_set" suffix, with this information the formatter will recognize and format it on the right way. Developers can use the "ebpf.formatter.bitfield.separator" annotation to set the separator to be used when joining the flags, by default "|" is used.

### Previous Idea

Before I was trying to use annotations in the metadata file, but after talking to @flyth I realized it was not the best solution and suggested by @flyth decided to implement it using 

### Testing 

Run the trace_tcp{drop,retrans} gadgets and see how they behave. 

### TODO

- [ ] Use arrays once #3032 is implemented (this PR won't block on that)

Fixes #2482
Related #2996

### DIFF
--- a/pkg/operators/ebpf/formatters.go
+++ b/pkg/operators/ebpf/formatters.go
@@ -16,6 +16,7 @@ package ebpfoperator
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/cilium/ebpf/btf"
 
@@ -26,7 +27,8 @@ import (
 )
 
 const (
-	enumTargetNameAnnotation = "ebpf.formatter.enum"
+	enumTargetNameAnnotation        = "ebpf.formatter.enum"
+	enumBitfieldSeparatorAnnotation = "ebpf.formatter.bitfield.separator"
 )
 
 func byteSliceAsUint64(in []byte, signed bool, ds datasource.DataSource) uint64 {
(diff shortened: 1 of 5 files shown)


pkg/operators/ebpf/formatters.go
```diff
@@ -92,18 +94,42 @@ func (i *ebpfInstance) initEnumFormatter(gadgetCtx operators.GadgetContext) erro
...
+					separator = "|"
+				}
+
+				formatter = func(ds datasource.DataSource, data datasource.Data) error {
+					inBytes := in.Get(data)
+					val := byteSliceAsUint64(inBytes, enum.Signed, ds)
+
+					var arr []string
+					for _, v := range enum.Values {
+						if val&v.Value != 0 {
```
flyth, line 112:
```suggestion
						if val&v.Value == v.Value {
```

This change is important for combinations. If for example you've got an enum entry with a value of 0x03, it otherwise would be `!= 0` for 0x01 and 0x02.

flyth (approved):
LGTM from code inspection! Thanks for following up on the enum idea!

I've just got one small change request.

pkg/operators/ebpf/formatters.go
mauriciovasquezbernal: I was wondering about this since there was a similar comment in https://github.com/inspektor-gadget/inspektor-gadget/pull/3025#discussion_r1648749330. Thanks for catching this up.

Issue #2816, opened 2024-05-07
[RFE] Add alternative enrichment naming option for l3address/l4address (and similar) formatters


flyth:
## Current situation

Currently, when using specific known variable types, the formatter operator will add a string representation as alternative into the eBPF structure. That means that a `source` field defined as l4address will have a `string` field added as `source.string`, even though all other subfields of `source` are hidden after adding the string representation. We do this to prevent a change of the underlying type (field with subfields) caused by an operator.

## Impact

`source.string` is unnecessarily long and causes trouble in the columns output.

## Ideal future situation

`source` would actually be enough.

## Implementation options

In order to keep the behavior of not changing the data layout depending on whether or not a specific operator is running, @alban proposed to add a `_raw` suffix to the eBPF field. If the formatter operator sees such a suffix, it could just remove it and use the remainder as the field name for the actual string representation and hide the `_raw` field.

PR #3023, opened 2024-06-17
formatters: Avoid changing name of the field

mauriciovasquezbernal:
The current logic of the formatters changes the name of the original field (adding a "_raw" suffix) and creates a new field with the original name.

This behavior is not right as the name and type of the fields depends on the operators being run, hence, it could cause issues when parsing the output of Inspektor Gadget and enabling / disabling the formatters operator.

This commit fixes it by keeping the original name in all the cases. The new name is derived from an annotation on the metadata file.

## Testing 

Columns output prints the pretty version of the fields (sig and timestamp in this example)

```bash 
$ sudo -E ig run trace_signal:latest --verify-image=false
...
RUNTIME.CONTAINERNAME  PID          TPID         MNTNS_ID  UID          GID         RET COMM        SIG         TIMESTAMP
c3                     16770        85763        402653492 0            0           0   sh          SIGTERM     2024-06-17T09:40:55.4905
```

Raw fields are also available in the json output: 

```bash 
$ sudo -E ig run trace_signal:latest --verify-image=false -o json | jq
...
{
  "comm": "sh",
  "gid": 0,
  "k8s": {
    "containerName": "",
    "hostnetwork": false,
    "namespace": "",
    "node": "",
    "podName": ""
  },
  "mntns_id": 4026534922,
  "pid": 16770,
  "ret": 0,
  "runtime": {
    "containerId": "67df57854c8dd52b766560725f3daec0a2bc423f00875f58f1197ca3b34a68f5",
    "containerImageDigest": "",
    "containerImageName": "busybox",
    "containerName": "c3",
    "runtimeName": "docker"
  },
  "sig": "SIGTERM",
  "sig_raw": 15,
  "timestamp": "2024-06-17T09:41:44.057569398-05:00",
  "timestamp_raw": 1718635304057569300,
  "tpid": 85846,
  "uid": 0
}
```

### `_raw` suffix 

@alban suggested to use a `_raw` suffix and automatically trim it from the operators. It's good because avoids having to specify the annotation, but I don't like having this "magic" behavior, so this PR always use the annotation. Should I implement the raw suffic support too?

## TODO

- [ ] Future PR: Carry on annotations, attributes, description to created field

- Ref #3020
- Ref #2816 (Fixes part of it, however {l3,l4}endpoints are still TBD in another PR)

### DIFF
--- a/pkg/utils/annotations/annotations.go
+++ b/pkg/utils/annotations/annotations.go
@@ -1,0 +1,44 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package annotations
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
+)
+
+func GetTargetNameFromAnnotation(
+	logger logger.Logger,
+	logPrefix string,
+	in datasource.FieldAccessor,
+	targetAnnotation string,
+) (string, error) {
+	annotations := in.Annotations()
+
+	if outName, ok := annotations[targetAnnotation]; ok {
+		logger.Debugf("%s: using custom target field %q for field %q", logPrefix, outName, in.Name())
+		return outName, nil
+	}
+
+	if outName, ok := strings.CutSuffix(in.Name(), "_raw"); ok {
+		logger.Debugf("%s: using %q as target field for %q", logPrefix, outName, in.Name())
+		return outName, nil
+	}
+
+	return "", fmt.Errorf("neither %q annotation nor '_raw' suffix found", targetAnnotation)
+}
--- a/gadgets/trace_dns/gadget.yaml
+++ b/gadgets/trace_dns/gadget.yaml
@@ -10,9 +10,7 @@ tracers:
 structs:
   event_t:
     fields:
-    - name: timestamp
-      attributes:
-        template: timestamp
+    - name: timestamp_raw
     - name: src
       description: Source endpoint
       attributes:
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -33,7 +33,7 @@
 #define MAX_ADDR_ANSWERS 1
 
 struct event_t {
-	gadget_timestamp timestamp;
+	gadget_timestamp timestamp_raw;
 
 	struct gadget_l4endpoint_t src;
 	struct gadget_l4endpoint_t dst;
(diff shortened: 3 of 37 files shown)


alban:
@mauriciovasquezbernal I like the logic you have implemented :+1:

flyth:
I think `field.targetname` is not specific enough for two reasons:

* there could be multiple targets for different converters/formatters/operators
* it's hard to find out what exactly reformatted the output by just looking at it and the metadata

I'd suggest something that is specific for the formatter/converter (in these cases) like `formatters.timestamp.target`, so you know instantly what operator and which functionality it affects.

> It's good because avoids having to specify the annotation, but I don't like having this "magic" behavior

I'm torn here: I also don't think too much magic is good, but this feels like something that is going to be used a LOT (timestamps especially) and I also think that we should try to avoid for the user to write a metadata file if not really necessary as it's the boring part of writing gadgets - in most cases the icing on the cake. And IMHO it also steepens the learning curve.

What about applying the "magic" only if additional circumstances are met, then? Like type AND name have to match - and maybe then issue a debug logging message saying "auto-generating 'timestamp' field based on name and type"? I know this adds a bit of logic in our code, but that way users can just copy/paste our gadget code and it would work ootb - and by seeing the message they'd know why it is there (should that question come up).

Difference in dev experience/learning curve for a first simple gadget:
* "oh nice, it shows the time human-readable automatically although it's an int"
* "Okay, that's an int, but I need the time in a readable format.
  * Let's debug: 'field.target annotation not found' - what is that? annotations? Have to read up on that. Oh, I need a metadata file. Oh, I need to describe my fields in yaml....".

But as said: I'm torn. Too much magic is not good. 🤷

mauriciovasquezbernal:
I updated the logic to also use some "magic" when the field has the `raw_` suffix. Please let me know what you think.

pkg/utils/annotations/annotations.go
```diff
@@ -0,0 +1,44 @@
...
+		logger.Debugf("%s: using custom target field %q for field %q", logPrefix, outName, in.Name())
+		return outName, nil
+	}
+
+	if outName, ok := strings.CutSuffix(in.Name(), "_raw"); ok {
+		logger.Debugf("%s: using %q as target field for %q", logPrefix, outName, in.Name())
+		return outName, nil
+	}
+
+	return "", fmt.Errorf("%q annotation nor '_raw' prefix found", targetAnnotation)
```
flyth, line 43:
```suggestion
	return "", fmt.Errorf("neither %q annotation nor '_raw' suffix found", targetAnnotation)
```

pkg/utils/annotations/annotations.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
+)
+
+func GetTargetNameFromAnnotation(
+	logger logger.Logger,
+	logPrefix string,
+	in datasource.FieldAccessor,
+	targetAnnotation string,
```
flyth, line 29: targetAnnotation could be replaced by logPrefix + `.target` (at least for the formatters pkg currently), but I guess you separated it to have the full annotation key (e.g. `timestampTargetAnnotation`) mentioned in code and thus searchable, right?

pkg/operators/ebpf/converters.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/annotations"
+)
+
+const (
+	enumTargetNameAnnotation = "ebpf.formatter.enum"
```
flyth, line 29: nit:
```suggestion

const (
	enumTargetNameAnnotation = "ebpf.converter.enum"
```

Or maybe rename the file.

flyth (commented):
Yeah, I think that's a good compromise now, thanks!

pkg/utils/annotations/annotations.go
mauriciovasquezbernal: Right, I prefer to keep them separated. I know it's more verbose but also more flexible.

pkg/operators/ebpf/converters.go
mauriciovasquezbernal: Renamed the file to have a single formatter concept.

flyth (approved):
LGTM! Thanks!

PR #3026, opened 2024-06-17
tests: Remove duplicated test

mauriciovasquezbernal:
This test is already implemented in gadgets/trace_open/test/trace_open_test.go.


mauriciovasquezbernal:
Ah, integration/k8s/run_insecure_test.go depends on this file. Will think about a solution for this.

mauriciovasquezbernal:
> Ah, integration/k8s/run_insecure_test.go depends on this file. Will think about a solution for this.

Fixed.

integration/k8s/run_insecure_test.go
```diff
@@ -63,5 +63,12 @@ func TestRunInsecure(t *testing.T) {
 	// TODO: Ideally it should not depend on a real gadget, but we don't have a "test gadget" available yet.
 	// As the image was not signed, we need to set --verify-image=false.
 	cmd := fmt.Sprintf("$KUBECTL_GADGET run --verify-image=false %s:5000/trace_open:%s -n %s -o json --insecure", registryIP, *gadgetTag, ns)
-	runTraceOpen(t, ns, cmd)
+
+	// run the gadget without verifying its output as we only need to check if it runs
```
burak-ok, line 67: But we are not checking if the gadget even ran anywhere AFAIU

mauriciovasquezbernal: It's checkd in https://github.com/inspektor-gadget/inspektor-gadget/blob/a2c1d2e4302e34f2c4a19c34efb4eca0833940b4/integration/command.go#L498-L502

I did a quick test by removing the `--insecure` flag and the test failed, maybe you can try it by yourself to be sure 100% it's working fine.

burak-ok: I just assumend that `c.kill()` would only return an error when "the killing" failed. Also the message in `require.NoError` is super misleading. I saw that error multiple times and thought - wow why did `kill` fail

Looking into the `kill` function it also reports the returned exitcode.

mauriciovasquezbernal: I changed the test not to use the startAndStop but timeout instead. The result is the same but it's easier to follow.

burak-ok: Oh my comment was nothing specific about this test. It was good.

I'll create a PR later/tomorrow to rename that `.kill` maybe or smth else

mauriciovasquezbernal:
@burak-ok can you please approve the PR? (Or is there anything else I'm missing?)

burak-ok (approved):

PR #1842, opened 2023-07-11
Remove `systemd` workaround

eiffel-fl:
Hi.

This PR removes the `systemd` workaround to use `ig` in `kubectl debug` with https://github.com/kubernetes/kubernetes/pull/119200:

```
$ kubectl debug node/minikube-docker --profile=sysadmin -ti --image-pull-policy=Never --image=ig-linux-amd64-workaround -- ig trace exec
Creating debugging pod node-debugger-minikube-docker-jlfzj with container debugger on node minikube-docker.
If you don't see a command prompt, try pressing enter.
k8s_quux_quux_default_a14a6fc… 149502           79539            cat              0   /bin/cat /etc/hostname                  
k8s_quux_quux_default_a14a6fc… 149503           79539            cat              0   /bin/cat /etc/hostname                  
k8s_quux_quux_default_a14a6fc… 149504           79539            cat              0   /bin/cat /etc/hostname
```

I removed a lot of code so maybe I broke few things there, but let's review it to avoid this kind of troubles.

Best regards.

PR #3030, opened 2024-06-18
Use `host.HostRoot` when listing containers

eiffel-fl:
Hi.

This PR fixes the specific case of `ig list-containers` where `HOST_ROOT` would be ignored:

```bash
# Before this commit:
$ docker run -ti --rm --privileged -v /:/host --pid=host ghcr.io/inspektor-gadget/ig list-containers        
RUNTIME.RUNTIMENAME RUNTIME.CONTAINERID RUNTIME.CONTAINERNAME RUNTIME.CONTAINERIMAGENAME RUNTIME.CONTAINERIMAGEDIGEST
# With this commit:
$ docker run -ti --rm --privileged -v /:/host --pid=host ig-linux-amd64 list-containers
WARN[0000] failed to get cgroups info of container cb5399ac31fac336a84c3d894c9547fdf1ce4fc4ae45893d625cfa219480168d from /proc/145942/cgroup: cgroup path not found in /proc/PID/cgroup 
WARN[0000] Runtime enricher (cri-o): couldn't get current containers: listing containers with request &ListContainersRequest{Filter:nil,}: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /host/run/crio/crio.sock: connect: no such file or directory" 
WARN[0000] Runtime enricher (podman): couldn't get current containers: listing containers: Get "http://d/v4.0.0/libpod/containers/json?all=true": dial unix /host/run/podman/podman.sock: connect: no such file or directory 
ERRO[0000] cgroup enricher: failed to get cgroup paths on container cb5399ac31fac336a84c3d894c9547fdf1ce4fc4ae45893d625cfa219480168d: cgroup path not found in /proc/PID/cgroup 
RUNTIME.RUNTIMENAME         RUNTIME.CONTAINERID RUNTIME.CONTAINERNAME                       RUNTIME.CONTAINERIMAG… RUNTIME.CONTAINERIMAG…
docker                      439f91158223fdc6e9… buildx_buildkit_magical_sammet0             moby/buildkit:buildx-…                                                     
docker                      cb5399ac31fac336a8… great_mendeleev                             ig-linux-amd64                               
docker                      690330e44400bffc9d… minikube-docker                             gcr.io/k8s-minikube/k…
```

Best regards.

### DIFF
--- a/docs/examples/ds-ig.yaml
+++ b/docs/examples/ds-ig.yaml
@@ -35,8 +35,6 @@ spec:
         volumeMounts:
           - mountPath: /host
             name: host
-          - mountPath: /run
-            name: run
           - mountPath: /sys/kernel/debug
             name: debugfs
       nodeSelector:
@@ -45,9 +43,6 @@ spec:
         - name: host
           hostPath:
             path: /
-        - name: run
-          hostPath:
-            path: /run
         - name: debugfs
           hostPath:
             path: /sys/kernel/debug
--- a/docs/examples/pod-ig.yaml
+++ b/docs/examples/pod-ig.yaml
@@ -27,14 +27,9 @@ spec:
       volumeMounts:
         - mountPath: /host
           name: host
-        - mountPath: /run
-          name: run
   # CHANGEME: where do you want to run this pod?
   nodeName: minikube-containerd
   volumes:
     - name: host
       hostPath:
         path: /
-    - name: run
-      hostPath:
-        path: /run
--- a/docs/getting-started/quick-start.md
+++ b/docs/getting-started/quick-start.md
@@ -34,7 +34,6 @@ We can use `docker run` to run `ig` on a Linux host:
 ```bash
 $ docker run -ti --rm \
     --privileged \
-    -v /run:/run \
     -v /:/host \
     --pid=host \
     ghcr.io/inspektor-gadget/ig \
--- a/docs/ig.md
+++ b/docs/ig.md
@@ -306,7 +306,6 @@ Example of command:
 ```bash
 $ docker run -ti --rm \
     --privileged \
-    -v /run:/run \
     -v /:/host \
     --pid=host \
     ghcr.io/inspektor-gadget/ig \
@@ -317,7 +316,6 @@ heuristic_yonath         3329233    3329211    ls               0   /bin/ls
 
 List of flags:
 - `--privileged` gives all capabilities such as `CAP_SYS_ADMIN`. It is required to run eBPF programs.
-- `-v /run:/run` gives access to the container runtimes sockets (docker, containerd, CRI-O).
 - `-v /:/host` gives access to the host filesystem. This is used to access the host processes via /host/proc, and access
   container runtime hooks (rootfs and config.json).
 - `--pid=host` runs in the host PID namespace. Optional on Linux. This is necessary on Docker Desktop on Windows because


alban:
Could you use SecureJoin?

alban:
> I was wondering about, but I would like to know when we have to use Securejoin() over Join().

The semantics for following symlinks is different. `Securejoin()` interprets symlinks relative to the `root` given as parameter. In contrast, `Join()` does not interpret symlinks and leave that task to users of the resulting path. In general, when the first parameter to join is the root of a container or of the host, use `SecureJoin()`. If not, use `Join()`.

Example where it gives different results:
On modern Linux distributions, `/var/run` is a symlink to `../run` (relative path). But on some old Linux distributions, `/var/run` was a symlink to `/run` (absolute path). Now imagine the scenario where we have this old Linux distribution with an absolute symlink, and a ig container where the users wants to use `/host/var/run/docker.sock`. With `Join()`, that would resolve to `/run/docker.sock` (inside the ig container, which is not what we want). With `SecureJoin()`, that would resolve to `/host/run/docker.sock` (pointing to the host's `docker.sock`, which is what we want).
This example is quite contrived because modern Linux distributions don't use absolute paths for the `/var/run` symlink (to avoid this kind of problems) and users tend to use `/run/docker.sock` instead of `/var/run/docker.sock` anyway.
But we should still use the semantically correct way to join paths in case users have a custom installation of docker with custom paths.

Additionally, using `Securejoin()` is sometimes not secure enough if the content of the `root` parameter is not trusted, depending on what is done to the resulting path in the code below the `SecureJoin`. In this PR, it is fine to use `SecureJoin` because the `root` parameter is the host root, so it is trusted (if malicious actors could modify files on the host root, we would have already lost). But when the `root` parameter is an untrusted container rootfs, it could be subject to TOCTOU issues. For this, IG now has helpers `OpenInContainer` and `ReadFileInContainer` (`pkg/utils/secureopen/secureopen.go`).

alban:
> Should we audit all the places where we use filepath.Join() to check if we need to replace them with securejoin.SecureJoin()?

That would be good. For the most part, it should be in a separate PR.

But I see some similar code that could be updated here: pkg/container-utils/containerutils.go:
```
socketPath = filepath.Join(host.HostRoot, envsp)
```

---

There are some more `/run` that can be removed:

```diff
--- a/charts/gadget/templates/daemonset.yaml
+++ b/charts/gadget/templates/daemonset.yaml
@@ -192,8 +192,6 @@ spec:
             - mountPath: /host/proc
               name: proc
               readOnly: true
-            - mountPath: /run
-              name: run
             - mountPath: /sys/kernel/debug
               name: debugfs
             - mountPath: /sys/fs/cgroup
```
And also:
```
$ git grep 'mountPath: /run'
charts/gadget/templates/daemonset.yaml:195:            - mountPath: /run
docs/examples/ds-ig.yaml:38:          - mountPath: /run
docs/examples/pod-ig.yaml:30:        - mountPath: /run
examples/builtin-gadgets/withfilter/trace/network/deploy.yaml:138:              mountPath: /run
examples/kube-container-collection/deploy.yaml:87:          mountPath: /run
examples/runc-hook/deploy.yaml:90:          mountPath: /run
pkg/resources/manifests/deploy.yaml:275:            - mountPath: /run
```

alban (approved):
LGTM from code review.

It would be good to test the [following](https://inspektor-gadget.io/docs/v0.29.0/ig/):
- [ ] CI tests on the citest/branch
- [ ] ig on the host
- [ ] ig with `kubectl debug node` (will be handled separately in #1842)
- [ ] ig as a systemd service
- [ ] ig in a docker container
- [ ] ig in a Kubernetes pod

And with [kubectl-gadget](https://inspektor-gadget.io/docs/v0.29.0/getting-started/install-kubernetes/):
- [ ] Helm chart
- [ ] kubectl-gadget deploy

TODO for other setups:
- [ ] [Addon for Minikube](https://github.com/kubernetes/minikube/blob/master/deploy/addons/inspektor-gadget/ig-daemonset.yaml.tmpl#L166)

alban:
Alternatively, you could add the following in `Dockerfiles/gadget.Dockerfile`:
```Dockerfile
RUN rmdir /run ; ln -s /host/run /run
```
However, `mkdir` and `ln` don't exist in the distroless image, so it would have to be done in the builder image, and then use `COPY --from=builder`.

Then, `-v /run:/run` can be removed without patching coreos/go-systemd.

I'll review your new version of this PR nonetheless if you decide to continue with the `-v /run:/run` in a separate PR.

alban (approved):
LGTM from code review.

But I couldn't check the CI results.

PR #3028, opened 2024-06-18
pkg/gadgets/capabilities: fix pid and tgid fields

i-Pear:
# pkg/gadgets/capabilities: fix pid and tgid fields

The `bpf_get_current_pid_tgid` helper returns `tgid << 32 | pid`, so:
```
event.pid = (u32)pid_tgid;
event.tgid = pid_tgid >> 32;
```

git blame: fbc8f659aaf215d5ffb9

PR #3031, opened 2024-06-18
 pkg/gadgets/capabilities: Fix pid field

mauriciovasquezbernal:
Fixes: 6609cad1f96e ("pkg/gadgets/capabilities: fix pid and tgid fields")

See #3028


alban (approved):

alban:
LGTM. Sometimes, two bugs cancel each other out, but it is better to fix them :)

PR #3038, opened 2024-06-19
gadgets/signal: Add test using the new framework.

eiffel-fl:
Hi.

I also wanted to add a test for `oomkill`, but it seems the new framework is missing the capability to run container with a memory limit.
Once https://github.com/inspektor-gadget/inspektor-gadget/pull/3036 is merged, I think this would be doable as what is currently done in https://github.com/inspektor-gadget/inspektor-gadget/pull/3036/commits/c0e01e99d249a373b0cc1dd65f01ece433f43100.
I am nonetheless still wondering if this is possible to run the container once `ig` is run, as this would be needed for `trace_oomkill`.

Best regards.


gadgets/trace_signal/test/trace_signal_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+				// Check only the existence of these fields
+				Pid:       utils.NormalizedInt,
+				Tid:       utils.NormalizedInt,
+				Timestamp: utils.NormalizedStr,
+				MountNsID: utils.NormalizedInt,
+			}
+
+			normalize := func(e *traceSignalEvent) {
+				utils.NormalizeInt(&e.Pid)
```
mauriciovasquezbernal, line 108: Add utils.NormalizeCommonData(&e.CommonData) to make it pass on k8s.

mauriciovasquezbernal (approved):
I tested it locally and after the change I suggested it works on both, k8s and ig. 
I kicked the CI again as it was failing due to an unrelated issue.

Approving now to about the RTT of another review cycle.

PR #3039, opened 2024-06-19
pkg/oci: Remove undeeded authOpts for getSigningInformation().

eiffel-fl:
This parameter was never used.

Fixes: 9f17b006486e ("pkg/oci: Add new options to check image was signed with given public key.")


burak-ok (approved):
LGTM. Would have thought a linter would catch this

PR #3040, opened 2024-06-19
integration/Command: Enhance error messages for kill

burak-ok:
The error message we got sometimes in the CI log was quite confusing. It mentioned that there was an error killing the process. Rather it was killing the process (if it still ran) and then also returning potential errors while running the `Command` or the exit code.

xref: https://github.com/inspektor-gadget/inspektor-gadget/pull/3026#discussion_r1644516943


mauriciovasquezbernal (approved):
LGTM, thanks!

PR #3044, opened 2024-06-20
Verify `ebpf-builder` in CI before building gadgets

eiffel-fl:
Hi.

Following the discussion in #2919, this PR verifies the `ebpf-builder` right before the build our gadgets.
It also modifies the referencing of this image while building to use digest, this way we should be sure to use the image we just verified.
It also adds some documentation on how to do so locally, so users can verify the image before building any gadgets.

Best regards.

### DIFF
--- a/docs/getting-started/verify.md
+++ b/docs/getting-started/verify.md
@@ -212,6 +212,40 @@ WARN[0000] you set --verify-image=false, image will not be verified
 RUNTIME.CONTAINERNAME  PID          UID          GID          MNTNS_ID RET FL… MODE        COMM        FNAME                  TIMESTAMP
 ```
 
+## Verify the ebpf-builder image
+
+We also sign the `ebpf-builder` image which is used to build gadgets.
+You can verify it using the following command:
+
+```bash
+$ cosign verify --key inspektor-gadget.pub ghcr.io/inspektor-gadget/ebpf-builder:latest
+```
+
+We highly recommend you to verify by digest and specify the digest when building.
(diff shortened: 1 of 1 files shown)


docs/getting-started/verify.md
```diff
@@ -212,6 +212,39 @@ WARN[0000] you set --verify-image=false, image will not be verified
 RUNTIME.CONTAINERNAME  PID          UID          GID          MNTNS_ID RET FL… MODE        COMM        FNAME                  TIMESTAMP
 ```
 
+## Verify the ebpf-builder image
+
+We also sign the `ebpf-builder` image which is used to build gadgets.
+You can verify it using the following:
```
burak-ok, line 218:
```suggestion
You can verify it using the following command:
```

docs/getting-started/verify.md
```diff
@@ -212,6 +212,39 @@ WARN[0000] you set --verify-image=false, image will not be verified
...
+
+```bash
+$ cosign verify --key inspektor-gadget.pub ghcr.io/inspektor-gadget/ebpf-builder:latest
+```
+
+We highly recommend you to verify by digest and specify the digest when building:
```
burak-ok, line 224:
```suggestion
We highly recommend you to verify by digest and specify the digest when building. This helps to protect against the TOCTOU problem. The image tag that got verified could point to another malicious image after the check.
```

I don't know if my explanation is worded good. But I think further information why verifying by digest is helpful

burak-ok (approved):
LGTM, thanks for making the CI safer.

Just 2 little nits

mauriciovasquezbernal (approved):
LGTM. What do you think about opening an issue to replace the usage of the `latest` tag in build.go:

https://github.com/inspektor-gadget/inspektor-gadget/blob/44e1e506998b28056791d4769da75f2cb650f94a/cmd/common/image/build.go#L51-L52

?

PR #3047, opened 2024-06-20
ci: Use westeurope for AKS.

eiffel-fl:
The VM are not avalaible in eastus:

Let's use westeurope for the moment.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1602,10 +1602,7 @@ jobs:
         # GitHub Action.
         az extension add --name aks-preview
 
-        # Let's keep thing in the US to avoid data crossing the Atlantic as
-        # GitHub data centers are in the US:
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
LGTM from code inspection. ARO test is failing but doesn't seem to be related to this PR.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1602,10 +1602,7 @@ jobs:
         # GitHub Action.
         az extension add --name aks-preview
 
-        # Let's keep thing in the US to avoid data crossing the Atlantic as
-        # GitHub data centers are in the US:
```
burak-ok, line 1606: Would this mean the latency is now higher while running the tests. So like pulling the gadget images. We already sometimes have issue because of timeouts, would this make our CI even more flaky?

mauriciovasquezbernal: Don't they use any kind of geo replication for the container registry?

Issue #3035, opened 2024-06-18
image-based gadgets: Port tests to use new testing framework 


mauriciovasquezbernal:
Now that #2867 is merged, we can use it to implement the tests for image based gadgets

The goal of this issue is to port the existing tests from integration/k8s to gadgets/. Inspiration can be taken from the existing test for trace_open: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_open/test/trace_open_test.go

- [x] run_insecure_test.go
- [ ] run_schedcls_test.go 
- [X] run_snapshot_process_test.go (#3053)
- [x] run_top_file_test.go (ee78caa8070712cd142760080c081758c58786ad & 44dc8b82f4dfeeda2279ae864e224ecfe453bf31)
- [x] run_trace_mount_test.go (#3036)
- [x] run_trace_oomkill_test.go (https://github.com/inspektor-gadget/inspektor-gadget/pull/3084)
- [x] run_trace_open_test.go (#3026)
- [x] run_trace_signal_test.go (https://github.com/inspektor-gadget/inspektor-gadget/pull/3038)
- [x] run_trace_sni_test.go (#3049)
- [x] run_trace_tcp_test.go (#3042)

unsuman:
I want to work on this issue, but where should I port these files? Should I create a new directory and place them separately?

mauriciovasquezbernal:
Hi @unsuman, thanks a lot for the interest on the project.

I think the only missing test to port is integration/k8s/run_schedcls_test.go. The idea is to move this under gadgets/ci/sched_cls_drop/test/integration as done for other gadgets in the PRs linked above. You can follow this https://www.inspektor-gadget.io/blog/2024/06/introducing-the-new-testing-framework-for-image-based-gadgets to understand more, but it can be a bit outdated.

PR #3049, opened 2024-06-20
gadgets/trace_sni: Port test to use new testing framework

mauriciovasquezbernal:
See #3035


burak-ok (approved):
Same as other PR:
LGTM, but not having deep insight yet into the new test harness

PR #3053, opened 2024-06-20
gadgets/snapshot_process: Port test to use new testing framework

mauriciovasquezbernal:
This PR contains two commits:
- gadgets/snapshot_process: Port test to use new testing framework
-  testing: Consolidate match logic in pkg/testing/match

The first commit is a rework / cleanup of the testing:

1. Move logic from helpers.go to match.go in order to have all the matching logic in the same place.
2. Consolidate the testing functions into two: MatchEntries() and MatchAllEntries() by creating a
new OutputMode type that defines how to parse the output. It (hopefully) makes the code more clear
as there aren't a lot of different functions around. This commit also adds some documentation explaning what those different modes are for.
3. Update all tests to use the new functions.


gadgets/snapshot_process/test/snapshot_process_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+			normalize := func(e *snapshotProcessEntry) {
+				utils.NormalizeCommonData(&e.CommonData)
+				utils.NormalizeInt(&e.MountNsID)
+				utils.NormalizeInt(&e.Pid)
+				utils.NormalizeInt(&e.Tid)
+				utils.NormalizeInt(&e.Ppid)
+			}
+
+			match.MatchEntries(t, match.PerLineJSONArrayMode, output, normalize, expectedEntry)
```
blanquicet, line 109: Mmmm using `PerLineJSONArrayMode` for snapshotters prevent us from verifying they are printing output in one single array. It means that if a snapshotter prints outcomes from each node separately, one per line, then this test will pass, and we don't want that. 

So, for now that we're not merging outcome from all nodes into one array it's fine, but probably we should use `JSONArrayMode` for snapshotters. WDYT? If you agree, please add a TODO here to change once the combiner operator is merged.

blanquicet (approved):
LGTM!

mauriciovasquezbernal:
I thought more about this. I think the main source of confusion is that we were trying to differentiate between the output being in a single line or multiple lines. Actually we don't need to differentiate those cases as it's possible to decode both cases with a json decoder. I updated the PR to use this new idea, the OutputMode thing is still present, but now it's about the kind of object to decode, and not how it's formatted. 

What do you think?

pkg/testing/match/match.go
burak-ok, line 1: I've added some simple Unittests: https://github.com/inspektor-gadget/inspektor-gadget/commit/54529dee4a370b7c35a48d2ce086b559628675e0 Feel free to cherry-pick

burak-ok (approved):
LGTM, thanks

docs/devel/hello-world-gadget.md
```diff
@@ -690,7 +690,7 @@ func TestMyGadget(t *testing.T) {
         }
 
         // parses the output and matches it to expectedEntry.
-        match.ExpectEntriesToMatch(t, output, normalize, expectedEntry)
+        match.MatchEntries(t, match.JSONMultiObjectMode, output, normalize, expectedEntry)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: With all the file and functions renaming it's complicated to separate this on different commits.

pkg/testing/match/match.go
```diff
@@ -72,28 +177,97 @@ out:
...
 }
+
+func decodeJSONOutput[T any](t *testing.T, outputMode OutputMode, output string, normalize func(*T)) []*T {
+	var entries []*T
+
+	switch outputMode {
+	case JSONSingleObjectMode:
+		entries = decodeJSONObjects[T](t, output, true)
+	case JSONMultiObjectMode:
+		entries = decodeJSONObjects[T](t, output, false)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'll ignore this one as IMO it doesn't improve the code.

pkg/testing/match/match.go
mauriciovasquezbernal: We reached the point where we need tests for our test helpers :scream: . Thanks for the tests!!

gadgets/snapshot_process/test/snapshot_process_test.go
mauriciovasquezbernal: You're right, this was working because we were running the integration tests in a cluster with a single node. I added a commit to this PR to use multiple nodes.

pkg/testing/match/match.go
burak-ok: But how do we test our tests that test our tests?

PR #3057, opened 2024-06-21
pkg/datasource: improve subscription handling

flyth:
Before this patch, subscriptions were dealt by type and then priority (so first single entries, then whole arrays and at the end the packet). The priority however is more important than the type, so this changes it to call all subscribers in order of their priority, regardless of type.


pkg/datasource/subscription.go
```diff
@@ -14,16 +14,6 @@
...
-	priority int
-	fn       DataFunc
-}
-
-type subscriptionArray struct {
-	priority int
-	fn       ArrayFunc
-}
-
 type subscriptionPacket struct {
```
mauriciovasquezbernal, line 17: Perhaps just call it `subscription`?

mauriciovasquezbernal (approved):
Makes sense, just one minor comment.

PR #3058, opened 2024-06-21
ig/build: Always pull builder-image if latest tag is used

burak-ok:
# ig/build: Always pull builder-image if latest tag is used

I had an older builder image on my system and then tried to compile new gadgets. It failed to compile since the builder image was too old.

I think when specifying `:latest` as a tag we should always check upstream to get the newest version.
WDYT?


mauriciovasquezbernal (approved):
Totally makes sense! Thanks for fixing it. 

We could switch to use the digest later on as mentioned in https://github.com/inspektor-gadget/inspektor-gadget/pull/3044#pullrequestreview-2132936998, however this fix is good for now.

PR #3059, opened 2024-06-21
tests: Disable TestTraceNetwork on cri-o

mauriciovasquezbernal:
This test is failing a lot (see issue #2358), creating a lot of noise and forcing us to rerun the CI multiple times.

It's not ideal to disable a failing test, but it's better than wasting a lot of time rerunning the CI. We'll keep the issue for fixing it once we have time.

https://inspektor-gadget.github.io/ig-test-reports/ shows how many times this test is failing, look for test-integration-k8s-ig_cri-o and then for TestTraceNetwork.

### DIFF
--- a/integration/k8s/trace_network_test.go
+++ b/integration/k8s/trace_network_test.go
@@ -26,6 +26,10 @@ import (
 )
 
 func TestTraceNetwork(t *testing.T) {
+	if containerRuntime == ContainerRuntimeCRIO {
+		t.Skipf("Skip running trace network test for %q runtime. See issue #2358", containerRuntime)
+	}
+
 	t.Parallel()
 	ns := GenerateTestNamespaceName("test-trace-network")
 


burak-ok:
I also feel the pain of this maybe flaky test

> It's not ideal to disable a failing test

True

> We'll keep the issue for fixing it once we have time

This test already fails since months. How are we going to assure that we are fixing it. It indeed might uncover a race condition (for cri-o) in Inspektor Gadget

mauriciovasquezbernal:
> How are we going to assure that we are fixing it. 

By giving more priority to the issue and ensuring somebody works on that. 

> It indeed might uncover a race condition (for cri-o) in Inspektor Gadget

Yes, there's something going on there was it's not normal that a test fails too much. 

As I mentioned above, I don't like skipping flaky tests, ideally we should fix them, but we're busy with other stuff and this is just creating a lot of overhead for us as we need to trigger the CI many times.

burak-ok (approved):
> By giving more priority to the issue and ensuring somebody works on that.

Lets make the linked issue P0 then

PR #3060, opened 2024-06-21
trace_mount: Add call column and flag decoding

### LINKED ISSUE #2995: gadgets: Finish implementing trace mount as image-based gadget
The `call` column is missing. It can be implemented with #2818

burak-ok:
Add a wasm layer which decodes the flags. Additionally the call column indicates how mount/umount was called in a summarized way.

resolves #2995


gadgets/trace_mount/program.go
```diff
@@ -0,0 +1,174 @@
...
+
+	for i, val := range flagNames {
+		if (1<<i)&flags == 0 {
+			continue
+		}
+		flagsStr = append(flagsStr, val)
+	}
+
+	return flagsStr
+}
```
mauriciovasquezbernal, line 70: I think you can handle all of this with #3020. I didn't use it on #3025 because the flags in trace_open aren't technically a bitfield.

gadgets/trace_mount/program.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Fields needed by this layer
+	flagsRawField, err := ds.GetField("flags_raw")
+	if err != nil {
+		api.Warnf("failed to get field: %s", err)
+		return 1
+	}
+
+	opField, err := ds.GetField("op")
```
mauriciovasquezbernal, line 100: What about using `op_raw` to (1) avoid depending on the enum formatter (2) comparing integers instead of strings?

gadgets/trace_mount/go.mod
mauriciovasquezbernal, line 1: `go mod tidy` is missing, otherwise tests won't run:

```bash 
$ cd gadgets/trace_mount/test/
$ IG_FLAGS="--verify-image=false" IG_PATH=ig IG_RUNTIME=docker go test -exec 'sudo -E' ./... -v
go: updates to go.mod needed; to update it:
        go mod tidy
```

mauriciovasquezbernal (commented):
Some comments. It's looking very good.

gadgets/trace_mount/program.go
burak-ok: I was just about to file a PR for a fix for #3020 with multiple bits :D Then I saw the other PR merged 4 hours ago

burak-ok:
`clang-format` fails because it doesn't like how I lined up the bitfield assignment

mauriciovasquezbernal (approved):
Works perfectly, thanks! 

> clang-format fails because it doesn't like how I lined up the bitfield assignment

I see two options:

1. Format the code according to the clang-format rules
2. Play around with its configuration until you get something good. ChatGPT suggested:
```
AlignConsecutiveAssignments: true
AlignConsecutiveDeclarations: true
```

It's up to you which one to choose.

burak-ok:
> 2. Play around with its configuration until you get something good. ChatGPT suggested:
> 
> ```
> AlignConsecutiveAssignments: true
> AlignConsecutiveDeclarations: true
> ```
> 
> It's up to you which one to choose.

Chosing 2 would lead to modifying at least 50 other files. I'll just reformat my code

PR #3064, opened 2024-06-24
gadgets/trace_exec: Update metadata.

eiffel-fl:
Fixes: 58f2f68f4284 ("execsnoop: add parent upper layer")


alban (approved):
LGTM. Thanks for fixing it!

cc @dorser

PR #3065, opened 2024-06-24
Revert "container-collection: add process metadata enricher"

eiffel-fl:
Hi!

I suppose this commit is faulty, let's test it and we can iterate to find a proper fix.

Best regards.


alban:
What is faulty about that commit?

alban (approved):
I tried to fix the tests in #3066 but it seems there are more tests to fix...

Ok for this revert (if the CI is green).

Then I can try to re-add #2966 but with all CI tests enabled (`citest/`).

PR #3068, opened 2024-06-24
pkg/operators: Add combiner

blanquicet:
# Adding combiner operator

This PR introduces a new operator named "combiner". Only for `TypeArray` datasources, this operator will be in charge of merging outcomes from all targets into a single array on the client side.

## Implementation details

- The combiner operator will activate itself only if:
  - Gadget is run against more than one single target. For instance, it'll never get activated for `ig` used as standalone (without `gadgetctl`).
  - There is at least one `TypeArray` datasources. **NOTE**: From now on, this documentation will assume that the datasources are `TypeArray`.
- The combiner operator requires the datasource to be annotated with `FetchIntervalAnnotation`, which is expected to be `time.Duration` type and is used as following:
  - If `FetchIntervalAnnotation` is 0, the combiner operator will emit a packet once it has received packets from all the targets (A maximum waiting time of 5 seconds has be hardcoded (Make it configurable?). 
  - Instead, if `FetchIntervalAnnotation` is greater than 0, the combiner operator will emit a packet once the timer is triggered, no matter it has received data from all targets or not.
  - The combiner operator is not currently using the `FetchCountAnnotation` and it relies on the server to close the datasource once done (Still to be implemented cc @flyth).
- The combiner operator is configured to have the highest priority on the client side. It is because, during the `PreStart` phase, the combiner operator will modify the datasources to force other operators on client side to subscribe to the datasource created by the combiner operator (From now on, called "combined-datasource") and not to the original ones coming from the server. The idea with this modification is that the combiner operator will subscribe to the original datasources and receive packets from each target separately, while it will emit one single packet on the combined-datasource only once it has received packets from all targets or timeout has been reached (This can be configured using `FetchIntervalAnnotation` as described above). Doing so, the rest of operators running on the server side will receive a packet where data from all nodes has already been combined, so they can sort, print and do any operation on the aggregated data arrays.

### DIFF
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -700,6 +700,35 @@ func (ds *dataSource) Fields() []*api.Field {
 	return res
 }
 
+func (ds *dataSource) CopyFieldsTo(out DataSource) error {
+	ds.lock.RLock()
+	defer ds.lock.RUnlock()
+
+	outDs, ok := out.(*dataSource)
+	if !ok {
+		return errors.New("invalid destination DataSource")
+	}
+
+	outDs.lock.Lock()
+	defer outDs.lock.Unlock()
+
+	if len(outDs.fields) == 0 {
+		outDs.fields = make([]*field, 0, len(ds.fields))
+		outDs.fieldMap = make(map[string]*field)
+	}
+
+	for _, f := range ds.fields {
+		msg := proto.Clone((*api.Field)(f))
+		nf, _ := msg.(*api.Field)
+		outDs.fields = append(outDs.fields, (*field)(nf))
+		if !FieldFlagUnreferenced.In(f.Flags) {
+			outDs.fieldMap[nf.Name] = (*field)(nf)
+		}
+	}
+
+	return nil
+}
+
 func (ds *dataSource) Accessors(rootOnly bool) []FieldAccessor {
 	ds.lock.RLock()
 	defer ds.lock.RUnlock()
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -218,6 +218,8 @@ func NewFromAPI(in *api.DataSource) (DataSource, error) {
 	} else {
 		ds.byteOrder = binary.LittleEndian
 	}
+	ds.tags = in.Tags
+	ds.annotations = in.Annotations
 	// TODO: add more checks / validation
 	return ds, nil
 }
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -176,6 +176,7 @@ func newDataSource(t Type, name string, options ...DataSourceOption) (*dataSourc
 		name:            name,
 		dType:           t,
 		requestedFields: make(map[string]bool),
+		requested:       true,
 		fieldMap:        make(map[string]*field),
 		byteOrder:       binary.NativeEndian,
 		tags:            make([]string, 0),
@@ -714,6 +715,12 @@ func (ds *dataSource) Accessors(rootOnly bool) []FieldAccessor {
 		})
 	}
 	return res
+}
+
+func (ds *dataSource) SetRequested(v bool) {
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+	ds.requested = v
 }
 
 func (ds *dataSource) IsRequested() bool {
--- a/pkg/datasource/datasource.go
+++ b/pkg/datasource/datasource.go
@@ -166,6 +166,7 @@ type DataSource interface {
 
 	Accessors(rootOnly bool) []FieldAccessor
 
+	SetRequested(bool)
 	IsRequested() bool
 
 	// ByteOrder returns a binary accessor using the byte order of the creator of the DataSource
--- a/pkg/datasource/datasource.go
+++ b/pkg/datasource/datasource.go
@@ -177,6 +177,8 @@ type DataSource interface {
 
 	Annotations() map[string]string
 	Tags() []string
+
+	CopyFieldsTo(DataSource) error
 }
 
 type DataSourceOption func(*dataSource)
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -317,7 +317,18 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
 	}
 
 	if run {
-		go c.run(localOperators)
+		if err := c.start(localOperators); err != nil {
+			return fmt.Errorf("starting local operators: %w", err)
+		}
+
+		c.Logger().Debugf("running...")
+
+		go func() {
+			// TODO: Client shouldn't need to wait for the timeout. It should be
+			// managed only on the server side.
+			WaitForTimeoutOrDone(c)
+			c.stop(localOperators)
+		}()
 	}
 
 	return nil
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -212,10 +212,26 @@ func (c *GadgetContext) RegisterDataSource(t datasource.Type, name string) (data
 	return ds, nil
 }
 
-func (c *GadgetContext) GetDataSources() map[string]datasource.DataSource {
+func (c *GadgetContext) getDataSources(all bool) map[string]datasource.DataSource {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	return maps.Clone(c.dataSources)
+
+	ret := maps.Clone(c.dataSources)
+	for name, ds := range ret {
+		// Don't forward unrequested data sources if all is false
+		if !all && !ds.IsRequested() {
+			delete(ret, name)
+		}
+	}
+	return ret
+}
+
+func (c *GadgetContext) GetDataSources() map[string]datasource.DataSource {
+	return c.getDataSources(false)
+}
+
+func (c *GadgetContext) GetAllDataSources() map[string]datasource.DataSource {
+	return c.getDataSources(true)
 }
 
 func (c *GadgetContext) SetVar(varName string, value any) {
--- a/pkg/operators/combiner/combiner.go
+++ b/pkg/operators/combiner/combiner.go
@@ -1,0 +1,316 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package combiner is a data operator that combines data from a same data
+// source coming from different targets into a single data source. This is
+// useful when we run the same gadget in a distributed environment like
+// Kubernetes and we want to combine the data from all the nodes on the client
+// side so that we can perform further operations on the combined data, e.g.
+// sorting. Notice that this operator is useful only when we have data sources
+// of type array.
+package combiner
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+)
+
+const (
+	// We need it to be the first operator to run in client side. Only operators
+	// registering data sources should run before the combiner operator.
+	Priority         = -500
+	DataSourcePrefix = "combined"
+	OperatorName     = "Combiner"
+)
+
+type combinerOperator struct{}
+
+func (o *combinerOperator) Name() string {
+	return OperatorName
+}
+
+func (o *combinerOperator) Init(params *params.Params) error {
+	return nil
+}
+
+func (o *combinerOperator) GlobalParams() api.Params {
+	return nil
+}
+
+func (o *combinerOperator) InstanceParams() api.Params {
+	return nil
+}
+
+func getFetchAnnotation(ds datasource.DataSource) (time.Duration, error) {
+	intervalAnn, ok := ds.Annotations()[api.FetchIntervalAnnotation]
+	if !ok {
+		return 0, errors.New("missing fetch interval annotation")
+	}
+	fetchInterval, err := time.ParseDuration(intervalAnn)
+	if err != nil {
+		return 0, fmt.Errorf("parsing fetch interval annotation to duration: %w", err)
+	}
+	return fetchInterval, nil
+}
+
+func (o *combinerOperator) InstantiateDataOperator(gadgetCtx operators.GadgetContext, paramValues api.ParamValues) (operators.DataOperatorInstance, error) {
+	targetsVar, ok := gadgetCtx.GetVar(runtime.NumRunTargets)
+	if !ok {
+		gadgetCtx.Logger().Debugf("combiner: No targets found. Skipping combiner operator")
+		return nil, nil
+	}
+
+	targets, ok := targetsVar.(int)
+	if !ok {
+		return nil, fmt.Errorf("invalid type for number of targets: %T, expected \"int\"", targetsVar)
+	}
+
+	if targets == 0 {
+		return nil, nil
+	}
+
+	configs := make(map[datasource.DataSource]*combinerConfig)
+	for _, ds := range gadgetCtx.GetDataSources() {
+		if ds.Type() == datasource.TypeArray {
+			interval, err := getFetchAnnotation(ds)
+			if err != nil {
+				return nil, fmt.Errorf("getting fetch annotation for ds %s: %w", ds.Name(), err)
+			}
+
+			configs[ds] = &combinerConfig{
+				// TODO: What happen if we receive more than one packet for the same
+				// target? We should probably have a way to handle this case.
+				packetBuf: make(chan datasource.PacketArray, targets),
+				interval:  interval,
+			}
+		}
+	}
+
+	if len(configs) != 0 {
+		gadgetCtx.Logger().Debugf("combiner: array data sources found (%d). Activating combiner operator", len(configs))
+		return &combinerOperatorInstance{
+			targets: targets,
+			configs: configs,
+		}, nil
+	}
+
+	return nil, nil
+}
+
+func (o *combinerOperator) Priority() int {
+	return Priority
+}
+
+type combinerConfig struct {
+	// Interval to wait for data before emitting the combined data
+	interval time.Duration
+
+	// Buffer to send data to the combiner data source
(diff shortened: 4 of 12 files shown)


pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,304 @@
...
+		ds.SubscribePacket(func(source datasource.DataSource, packet datasource.Packet) error {
+			// Use Marshal/Unmarshal to create a deep copy
+			b, _ := proto.Marshal(packet.Raw())
+			pArray, err := combinedDs.NewPacketArrayFromRaw(b)
+			if err != nil {
+				return fmt.Errorf("creating packet array from raw: %w", err)
+			}
+
+			// Send the packet to the combiner data source
+			o.pktBuffers[ds.Name()] <- pArray
```
mauriciovasquezbernal, line 280: nit: Move this outside the subscribe function to avoid the lookup for each packet:

```diff
$ git diff
diff --git a/pkg/operators/combiner/combiner.go b/pkg/operators/combiner/combiner.go
index fe124056c..6f0f115e4 100644
--- a/pkg/operators/combiner/combiner.go
+++ b/pkg/operators/combiner/combiner.go
@@ -266,6 +266,8 @@ func (o *combinerOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) e

                gadgetCtx.Logger().Debugf("%s: subscribing to %s", o.Name(), ds.Name())

+               pktBuffers := o.pktBuffers[ds.Name()]
+
                ds.SubscribePacket(func(source datasource.DataSource, packet datasource.Packet) error {
                        gadgetCtx.Logger().Debug("Received data array ... Sending to combiner")

@@ -277,7 +279,7 @@ func (o *combinerOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) e
                        }

                        // Send the packet to the combiner data source
-                       o.pktBuffers[ds.Name()] <- pArray
+                       pktBuffers <- pArray

                        return nil
                }, Priority)
```

mauriciovasquezbernal (commented):
Some minor comments. I noticed that the headers column is printed multiple times:

```bash
$ sudo -E ./gadgetctl --config gadgetctl.yaml run snapshot_process:latest
WARN[0000] version skew detected: client (v0.0.0) vs server (v0.30.0-282-geb3ddad9b-dirty)
RUNTIME.CONTAINERNAME           COMM                            PID               TID              UID              GID             PPID
RUNTIME.CONTAINERNAME           COMM                            PID               TID              UID              GID             PPID
```

But it seems not to be related to this PR, right?

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+)
+
+const (
+	// We need it to be the first operator to run in client side. Only operators
+	// registering data sources should run before the combiner operator.
+	Priority             = -500
+	DataSourcePrefix     = "combined"
+	OperatorName         = "Combiner"
+	OperatorInstanceName = "CombinerInstance"
```
flyth, line 44: Please just use one name for Operator + Instance (both "Combiner" or "combiner") - this is partially used for param creation, so it's less confusing if they're the same.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+		return nil, nil
+	}
+
+	targets, ok := targetsVar.(int)
+	if !ok {
+		return nil, fmt.Errorf("invalid type for number of targets: %T, expected \"int\"", targetsVar)
+	}
+	if targets > 0 {
+		for _, ds := range gadgetCtx.GetDataSources() {
+			if ds.Type() == datasource.TypeArray {
```
flyth, line 80: Should this be more explicit by using an annotation on the datasource?

`combiner.enable: true` on the data source; or if we stick to enabling by default allowing to do a `combiner.disable: true`.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+func (o *combinerOperator) Priority() int {
+	return Priority
+}
+
+type combinerOperatorInstance struct {
+	// Number of targets to (ideally) wait for before emitting the combined data
+	targets int
+
+	// Map of packet buffers to send data to the combiner data source
+	pktBuffers map[string]chan datasource.PacketArray
```
flyth, line 107: Maybe use the `datasource.DataSource` as key to the map. Maybe at some point we need to add the ability to rename data sources and that could mess things up, so I think it's better to move forward by using the actual data source in most cases.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+
+	for {
+		select {
+		case <-o.done:
+			gadgetCtx.Logger().Debug("combiner: done")
+			return
+		case inPacket := <-packetBuf:
+			if inPacket == nil {
+				continue
+			}
```
flyth, line 199: Why would this happen?

pkg/operators/combiner/combiner.go
flyth: Also, in here you could already collect the datasources and pre-fill a struct (if transformed to `map[datasource.DataSource]combinerConfig`, where combinerConfig also contains the `chan datasource.PacketArray`, replacing the `pktBuffers`), so in `PreStart()` you could iterate over that instead of getting all datasources again and applying filters (is array, maybe annotations valid, etc).

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+			return fmt.Errorf("getting fetch annotation for ds %s: %w", ds.Name(), err)
+		}
+
+		// Disable original data source to avoid other operators subscribing to it
+		ds.SetRequested(false)
+
+		// Register a new data source that will emit the combined data
+		combinedDs, err := gadgetCtx.RegisterDataSource(
+			datasource.TypeArray,
+			fmt.Sprintf("%s-%s", DataSourcePrefix, ds.Name()),
```
flyth, line 250: Add an annotation that can change the prefix or the whole name and only fall back if not present?

`combiner.target: combinedstuff`

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+		)
+		if err != nil {
+			return fmt.Errorf("registering combiner data source for %s: %w", combinedDs.Name(), err)
```
flyth, line 253:
```suggestion
			return fmt.Errorf("registering combiner data source for %s: %w", ds.Name(), err)
```
Otherwise this would probably crash on err != nil.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+		o.pktBuffers[ds.Name()] = make(chan datasource.PacketArray, o.targets)
+		go o.forwardData(gadgetCtx, ds.Name(), interval, combinedDs)
+
+		gadgetCtx.Logger().Debugf("%s: subscribing to %s", o.Name(), ds.Name())
+
+		ds.SubscribePacket(func(source datasource.DataSource, packet datasource.Packet) error {
+			gadgetCtx.Logger().Debug("Received data array ... Sending to combiner")
+
+			// Use Marshal/Unmarshal to create a deep copy
+			b, _ := proto.Marshal(packet.Raw())
```
flyth, line 273: Is `proto.Clone()` an option here? If it can't be used directly, maybe it makes sense to make that a member of `datasource.Packet`.

flyth (commented):
Looks great! I will need to play around with it a bit more.

The `requested` feature is being used slightly different to what I imagined initially (although it never got implemented, this was intended to be used as means to tell the server that the client is not interested in a particular datasource (or even fields) to save network bandwidth), maybe we should make it rather similar to the "unreferencing" we do for fields as well. But we can address that in a follow-up PR with more discussion.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,304 @@
...
+	if o.pktBuffers != nil {
+		for _, buf := range o.pktBuffers {
+			if buf != nil {
+				close(buf)
+			}
+		}
+		o.pktBuffers = nil
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
blanquicet: I don't think it'll be called more than once, but I don't know the approach we want to have here. If for any reason we call it twice or we stop a instance that hasn't been started, it will crash. Do we want to protect operators against that or we prefer to rely that the framework is working properly?

pkg/operators/combiner/combiner.go
blanquicet: Mmmm the hash table lookup is O(1), but yeah, for sure more expensive than a don't do it.

blanquicet:
> Looks great! I will need to play around with it a bit more.
> 
> The `requested` feature is being used slightly different to what I imagined initially (although it never got implemented, this was intended to be used as means to tell the server that the client is not interested in a particular datasource (or even fields) to save network bandwidth), maybe we should make it rather similar to the "unreferencing" we do for fields as well. But we can address that in a follow-up PR with more discussion.

Yeah. TBH, I used the `requested` feature because it was already there and wasn't used, but "requested" doesn't seem to be the proper name. Yeah, probably "unreferencing" or "enabling" would make clearer. Given that it's more about naming and not the feature itself, I also suggest managing in another PR.

pkg/datasource/data.go
```diff
@@ -699,6 +702,30 @@ func (ds *dataSource) Fields() []*api.Field {
 	return res
 }
 
+func (ds *dataSource) CopyFieldsTo(out DataSource) error {
```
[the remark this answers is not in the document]
blanquicet: For now, I updated it to not resetting them but appending. But yeah, I also would like `RegisterDataSource` to accept such an option.

pkg/datasource/data.go
```diff
@@ -699,6 +702,30 @@ func (ds *dataSource) Fields() []*api.Field {
...
+	outDs.lock.Lock()
+	defer outDs.lock.Unlock()
+
+	outDs.fields = make([]*field, 0, len(ds.fields))
+	outDs.fieldMap = make(map[string]*field)
+	for _, f := range ds.fields {
+		msg := proto.Clone((*api.Field)(f))
+		nf, _ := msg.(*api.Field)
+		outDs.fields = append(outDs.fields, (*field)(nf))
+		outDs.fieldMap[nf.Name] = (*field)(nf)
```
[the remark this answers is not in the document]
blanquicet: Completely missed it. Thanks!

pkg/datasource/datasource.go
```diff
@@ -33,6 +33,11 @@ const (
 	TypeArray
 )
 
+const (
+	FetchCountAnnotation    = "fetch-count"
+	FetchIntervalAnnotation = "fetch-interval"
```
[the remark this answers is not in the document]
blanquicet: Me neither. For now, I moved it to the api pkg.

pkg/gadget-context/gadget-context.go
```diff
@@ -317,6 +317,9 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
 	}
 
 	if run {
+		if err := c.start(localOperators); err != nil {
+			return fmt.Errorf("starting local operators: %w", err)
+		}
```
flyth, line 322: Maybe move the `start()` contents back into `run()` and create the go routine for the waiting/stopping inside of there as well (instead of below). Otherwise it's confusing that there's a `start()` and `run()` but no `stop()` or similar.

pkg/gadget-context/gadget-context.go
```diff
@@ -212,10 +212,26 @@ func (c *GadgetContext) RegisterDataSource(t datasource.Type, name string) (data
...
+	return ret
+}
+
+func (c *GadgetContext) GetDataSources() map[string]datasource.DataSource {
+	return c.getDataSources(false)
+}
+
+func (c *GadgetContext) GetDataSourcesAll() map[string]datasource.DataSource {
+	return c.getDataSources(true)
 }
```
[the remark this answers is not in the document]
blanquicet: Mmmm I want the clean `GetDataSources` to return the requested datasources (which is the most common case). Instead, return also the non-requested datasources only if developer intentionally want to include them, so I want that function to be called differently. 

I also thought about adding options to the GetDataSources because maybe in the future we want also to get datasources with specific annotations or tags. But for now, it seems like overkilling the problem. WDYT?

pkg/operators/combiner/combiner.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Good point, I think we need to clarify it. IMHO Stop() should never be called twice, I think that easy to do. I'm worried about the case when Stop() is called because Start() failed and some of the things in the instance could have not been initialized, in that case the test will make sense. 

Let's keep it as it's and make a pass on all operators later on.

pkg/operators/combiner/combiner.go
blanquicet: We need to do this in other operators then.

pkg/operators/combiner/combiner.go
blanquicet: You're right. It doens't make sense.

pkg/operators/combiner/combiner.go
blanquicet: TBH, I don't see the advantage.

pkg/operators/combiner/combiner.go
blanquicet: To make it be:

```go
			// Deep copy and send it to the combiner data source
			clonedPkt := packet.Clone()
			pktBuffers <- clonedPkt
```

pkg/operators/combiner/combiner.go
flyth: > > Should this be more explicit by using an annotation on the datasource?
> 
> I don't see a use case for a datasource that generates arrays that doesn't want to combine the output from all targets, especially due to the way we print them in the CLI (clean up screen each time a new batch arrives). Do you?

If you're just digesting the arrays as they come in from your nodes to do post-processing in your own third-party app (consuming just JSON for example), you might want to have separated events per node or so. I would even say that the default should be not to combine them as combining could be seen as "just" an optimization for certain use cases.

> > Also, in here you could already collect the datasources and pre-fill a struct
> 
> I didn't want to do that because `InstantiateDataOperator` is also called in `GetGadgetInfo`. Thus, it's a waste of resources just to then be discarded. I think we need to improve that first. WDYT?

I wouldn't worry too much about those resources tbh - it's not really much compared to everything else that is happening for `GetGadgetInfo()` anyway (parsing eBPF etc.). Also, since `GetGadgetInfo()` is only called on one node and can potentially be cached, it's called much less than `RunGadget()`, where the data isn't thrown away.

I think given that, I would prefer to have the logic of filtering/error checking only once and at the earliest possible time.

blanquicet:
I managed most comments. Ready for another round of reviews.

pkg/operators/combiner/combiner.go
flyth: Yeah - but I don't see why we shouldn't give the user the opportunity to change the name :)

pkg/operators/combiner/combiner.go
flyth: Yeah, looks good to me. I don't have numbers, but I would guess this is also faster.

pkg/gadget-context/gadget-context.go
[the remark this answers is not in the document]
flyth: Agree, if we see it from the `unreferenced` perspective (like discussed in another comment), it makes more sense to be the default use case to have them filtered. Maybe still `GetDataSourcesAll()` => `GetAllDataSources()` to sound more natural? But I don't mind if you want to keep it the way it is right now.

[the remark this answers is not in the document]
blanquicet: Applied.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,301 @@
...
+	targets int
+
+	// Map of packet buffers to send data to the combiner data source
+	pktBuffers map[datasource.DataSource]chan datasource.PacketArray
+
+	done chan struct{}
+}
+
+func (o *combinerOperatorInstance) Name() string {
+	return OperatorName
```
burak-ok, line 142: Do we want the same name for the instance or another one (e.g. _instance is added)

pkg/operators/combiner/combiner.go
blanquicet: Fair. I'll keep it enabled by default for now and will work on the possibility of disabling it through annotations in a follow-up PR.

pkg/operators/combiner/combiner.go
blanquicet: OK, I'll add it to the TODO list.

pkg/operators/combiner/combiner.go
blanquicet: I'll add it to the TODO list for follow-up optimizations.

pkg/operators/combiner/combiner.go
```diff
@@ -0,0 +1,309 @@
...
+			if config.interval == 0 && targetCount == o.targets {
+				if err := emitAndAllocate(); err != nil {
+					gadgetCtx.Logger().Errorf("Failed emitting and allocating combined data: %s", err)
+				}
+				return
+			}
+		case <-c:
+			if config.interval == 0 {
+				gadgetCtx.Logger().Warnf("Data is incomplete: timeout waiting for data from all targets (%d/%d)",
+					targetCount, o.targets)
```
burak-ok, line 228: Don't we need an emitAndRelease() here + return?

scenario with the current code:
1. We have 3 nodes
2. We set interval=0
3. We received packages from 2 nodes -> `targetCount==2`
4. 5 second timer runs out, because third node was a little late, and we are in the `<-c` case
5. We print the warning and call `emitAndAllocate()` -> `targetCount` is set to 0
6. We are still in the `for` loop and receive now the last package, so we are in the `inPacket := <-config.packetBuf:` case
7. The last package is now grouped with the second intervals from the first 2 nodes
8. ...

If I understood correctly `interval == 0` seems to be a one shot, nothing periodic, right? The `return` in the `if config.interval == 0 && targetCount == o.targets` indicates that at least.

pkg/datasource/data.go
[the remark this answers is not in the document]
blanquicet: https://github.com/inspektor-gadget/inspektor-gadget/issues/3268

pkg/operators/combiner/combiner.go
blanquicet: Yep: https://github.com/inspektor-gadget/inspektor-gadget/pull/3068#discussion_r1697657792

pkg/operators/combiner/combiner.go
blanquicet: > If I understood correctly interval == 0 seems to be a one shot, nothing periodic, right? The return in the if config.interval == 0 && targetCount == o.targets indicates that at least.
>
Correct.

> 7. The last package is now grouped with the second intervals from the first 2 nodes
>
There's no second interval because `interval` is 0. So, yes, the packet with be stored to be emitted, but it never gets actually emitted. However, you're right. For cleanness, it should return but the outcome will not change.

pkg/operators/combiner/combiner.go
flyth: Most of them already do it that way, but yeah, there are a couple leftovers that need fixing.

pkg/operators/combiner/combiner.go
flyth: > This combined ds will not exposed anywhere, not even in the metadata file.

You can still add annotations and such for it, it's being looked up 😄 

```
...
DEBU[0000] datasource combined-file not found in gadget info
...
```

flyth (approved):
Did some more tests - looks good!

LGTM, great job - thanks!

PR #3069, opened 2024-06-24
trace_tcpconnect: Add retcode

### LINKED ISSUE #2880: [RFE] Support tracing failed TCP connection attempts
## Current situation

Some users have highlighted that with current gadgets they can't detect whether a TCP connection attempt succeeded or failed. For instance, they would like to use IG to trace when a connection attempt to the Kubernetes API server fails.

The `trace tcp` gadget traces when a TCP connection is established or closed in the following situation:

- The TCP connection was established after a connect() system call.
- The TCP connection was established after an accept() system call.
- The TCP connection was closed following the close() system call.
- The TCP connection was either established or closed following an unknown reason.

On the other hand, the `trace tcpconnect` traces all the tcp connection attempts but it doesn't show whether it succeeded or not. 

## Impact

Users can't identify when a TCP connection attempt fails using Inspektor Gadget.

## Implementation options

Questions:

- Should we support this use case in trace tcp or trace tcpconnect? Or should we consolidate them into one single gadget that also shows the failed connection attempts.
- What to trace to capture this events? Brainstorming:
  - Check return value of `tcp_v4_connect` and `tcp_v6_connect`?
  - Check the tcp state changes (passing from `TCP_SYN_SENT` to any other state diff than `TCP_ESTABLISHED` means a failure)?

## Additional information

- TCP tracepoints: https://www.brendangregg.com/blog/2018-03-22/tcp-tracepoints.html
- Check what BCC offers to cover such a use case

burak-ok:
kprobe/tcp_v._connect is only sending the SYN package and also only returns errors regarding that package.

This commit uses the kprobe/inet_stream_connect kernel function. It encapsulates the entire tcp connect mechanism and therefore the retcodes can now be captured

closes #2880

### DIFF
--- a/gadgets/trace_tcpconnect/program.bpf.c
+++ b/gadgets/trace_tcpconnect/program.bpf.c
@@ -44,6 +44,7 @@ struct event {
 	__u32 gid;
 	gadget_mntns_id mntns_id;
 	__u64 latency;
+	int retcode;
 };
 
 const volatile int filter_ports[MAX_PORTS];
@@ -191,7 +192,7 @@ static __always_inline void count_v6(struct sock *sk, __u16 dport)
 
 static __always_inline void trace_v4(struct pt_regs *ctx, pid_t pid,
 				     struct sock *sk, __u16 dport,
-				     __u64 mntns_id)
+				     __u64 mntns_id, int ret)
 {
 	struct event *event;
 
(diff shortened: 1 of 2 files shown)


mauriciovasquezbernal (approved):
LGTM. I tried it and works fine, thanks!

gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -44,6 +44,7 @@ struct event {
 	__u32 gid;
 	gadget_mntns_id mntns_id;
 	__u64 latency;
+	int retcode;
```
mauriciovasquezbernal, line 47: I opened #3110 to provide a human readable version of this later on.

PR #3082, opened 2024-06-26
ci: Remove ARO from release needs.

eiffel-fl:
At the moment, ARO is broken, so let's remove it to avoid it impeding next release.


alban (approved):
Instead of removing it, can you comment it with a comment explaining why it is disabled? Presumably, it will be re-added later once it works again.

Otherwise, LGTM.

PR #3090, opened 2024-06-27
ci: Use eastus2 for AKS

burak-ok:
We seem to have capacity issues with the VMs we request in `westeurope`: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/9693321973/job/26752261184#step:9:30


burak-ok:
citest: https://github.com/inspektor-gadget/inspektor-gadget/commits/citest/burak/ci/region/east_us_2/

PR #3076, opened 2024-06-25
Improve endpoints formatter

mauriciovasquezbernal:
This PR improves the way endpoints are formatted in the columns and json formats:

#### Before 

The columns output printed a lot of fields related to the endpoints:

```bash 
$ sudo -E ig run trace_tcpconnect:latest --verify-image=false
INFO[0000] Experimental features enabled
RUNTIME.CONTAINERNA… SRC… SRC… DST… DS… TASK       PID        UID        GID        MNTNS_… TIMESTAMP              SRC.ADDRE… DST.ADDRE…
netem                341… 6    80   6   wget       137202     0          0          4026534 2024-06-25T15:42:05.24 172.17.0.2 1.1.1.1:80
netem                454… 6    443  6   wget       137202     0          0          4026534 2024-06-25T15:42:05.36 172.17.0.2 1.1.1.1:44
netem                606… 6    443  6   wget       137202     0          0          4026534 2024-06-25T15:42:06.44 172.17.0.2 1.0.0.1:4
```

It wasn't possible to choose an endpoint with `--fields`:

```bash
$ sudo -E ig run trace_tcpconnect:latest --verify-image=false --fields=src,dst
INFO[0000] Experimental features enabled
Error: pre-starting operator "cli": setting fields: column "src" is invalid
```

and the endpoints weren't present on the json output:

```bash
$ sudo -E ig run trace_tcpconnect:latest --verify-image=false -o json | jq
INFO[0000] Experimental features enabled
{
  "gid": 0,
  "k8s": {
    "containerName": "",
    "hostnetwork": false,
    "namespace": "",
    "node": "",
    "podName": ""
  },
  "latency": 0,
  "mntns_id": 4026534293,
  "pid": 137803,
  "runtime": {
    "containerId": "31fc2dd570cef84e2f9131e662833b6318cf5dac4d753c0854767e41419743c1",
    "containerImageDigest": "",
    "containerImageName": "wbitt/network-multitool",
    "containerName": "netem",
    "runtimeName": "docker"
  },
  "task": "wget",
  "timestamp": "2024-06-25T15:56:52.668658281-05:00",
  "timestamp_raw": 1719349012668658200,
  "uid": 0
}
```

#### After 

The columns output shows the "summary" version for dst and src columns by default, all other subfields are hidden:

(It's still to improve the annotations / attributes to avoid trimming these fields)

```bash 
$ sudo -E ig run trace_tcpconnect:latest --verify-image=false
INFO[0000] Experimental features enabled
[...]

PR #3093, opened 2024-06-27
Finish porting trace dns to image-based gadget

### LINKED ISSUE #2999: gadgets: Finish porting trace dns as an image-based gadget
The (image-based) [trace dns](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/trace_dns) gadget isn't synchronized with the [built-in version](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/trace/dns), for instance #2544 was only implemented in the built-in version. The goal of this issue is to identify and align the differences between both gadgets.

Missing things (non exhaustive list):
- There's not a human readable version of Qtype (can be fixed by using enums)
- Same as above for Pktype 
- Same for QR 
- Addresses returned in the reply aren't printed
- Mount and net ns fields aren't hidden by default

mauriciovasquezbernal:
This commit is very similar to d772f2ef04c7 ("dns: Parse DNS packet in user space"),
    but in this case the parsing is done in wasm because it's an image-based
    gadget.

    Unfortunately it seems that using gopacket with tinygo isn't possible,
    so I decided to use golang.org/x/net/dns/dnsmessage that provides
    the functionality we need.

--- 

Opening as draft until #3076 is merged. Feedback is already welcome. 

 Fixes #2999

### TODO 

- [x] Provide raw values for qtype and rcode

### DIFF
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -1,36 +1,76 @@
 // SPDX-License-Identifier: GPL-2.0
-/* Copyright (c) 2021 The Inspektor Gadget authors */
+/* Copyright (c) 2024 The Inspektor Gadget authors */
 
 #include <linux/bpf.h>
 #include <linux/if_ether.h>
 #include <linux/ip.h>
+#include <linux/ipv6.h>
 #include <linux/in.h>
-#include <linux/in6.h>
 #include <linux/tcp.h>
 #include <linux/types.h>
 #include <linux/udp.h>
 #include <sys/socket.h>
+#include <stdbool.h>
 
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_endian.h>
 
-#include <gadget/buffer.h>
 #include <gadget/macros.h>
 #include <gadget/types.h>
 
 #define GADGET_TYPE_NETWORKING
 #include <gadget/sockets-map.h>
 
-// Max DNS name length: 255
-// https://datatracker.ietf.org/doc/html/rfc1034#section-3.1
-#define MAX_DNS_NAME 255
+unsigned long long load_byte(const void *skb,
+			     unsigned long long off) asm("llvm.bpf.load.byte");
+unsigned long long load_half(const void *skb,
+			     unsigned long long off) asm("llvm.bpf.load.half");
+unsigned long long load_word(const void *skb,
+			     unsigned long long off) asm("llvm.bpf.load.word");
+
+#ifndef NEXTHDR_HOP
+#define NEXTHDR_HOP 0 /* Hop-by-hop option header. */
+#define NEXTHDR_TCP 6 /* TCP segment. */
+#define NEXTHDR_UDP 17 /* UDP message. */
+#define NEXTHDR_ROUTING 43 /* Routing header. */
+#define NEXTHDR_FRAGMENT 44 /* Fragmentation/reassembly header. */
+#define NEXTHDR_AUTH 51 /* Authentication header. */
+#define NEXTHDR_NONE 59 /* No next header */
+#define NEXTHDR_DEST 60 /* Destination options header. */
+#endif
+
+#define DNS_QR_QUERY 0
+#define DNS_QR_RESP 1
+
+#define MAX_PORTS 16
+const volatile __u16 ports[MAX_PORTS] = { 53, 5353 };
+const volatile __u16 ports_len = 2;
+
+static __always_inline bool is_dns_port(__u16 port)
+{
+	for (int i = 0; i < ports_len; i++) {
+		if (ports[i] == port)
+			return true;
+	}
+	return false;
+}
+
+enum pkt_type_t : __u8 {
+	HOST,
+	BROADCAST,
+	MULTICAST,
+	OTHERHOST,
+	OUTGOING,
+	LOOPBACK,
+	USER,
+	KERNEL,
+};
+
+// TODO: what's a reasonable value for this?
+// Or can we remove this altogether?
+#define MAX_PACKET (1024 * 9) // 9KB
 
 #define TASK_COMM_LEN 16
-
-// Maximum number of A or AAAA answers to include in the DNS event.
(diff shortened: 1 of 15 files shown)


pkg/container-utils/containerutils.go
```diff
@@ -181,3 +181,41 @@ func GetIfacePeers(pid int) ([]*net.Interface, error) {
 
 	return ifacesHost, err
 }
+
+// GetIfaces returns the interfaces inside the container
+// It returns only veth and up devices
+func GetIfaces(pid int) ([]*net.Interface, error) {
```
alban, line 187: Is it only for tests?

If so, it could be moved to testutils.

If not, what about moving this in `pkg/utils/netns/netns.go` to solve the import cycle?

Also, this is only used to get the IP so it could return a `[]net.IP`.

pkg/container-utils/containerutils.go
```diff
@@ -181,3 +181,41 @@ func GetIfacePeers(pid int) ([]*net.Interface, error) {
...
+		links, err := netlink.LinkList()
+		if err != nil {
+			return fmt.Errorf("getting links: %w", err)
+		}
+
+		for _, link := range links {
+			veth, ok := link.(*netlink.Veth)
+			if !ok {
+				continue
+			}
```
alban, line 200: It will not work with `--host`. But at the moment, there are no tests for `--host`, so that's ok?

pkg/container-utils/testutils/docker.go
```diff
@@ -140,6 +140,15 @@ func (d *DockerContainer) Run(t *testing.T) {
 		t.Fatalf("Failed to inspect container: %s", err)
 	}
 	d.pid = containerJSON.State.Pid
+
+	for _, m := range containerJSON.NetworkSettings.Networks {
+		if d.ip != "" {
+			t.Fatal("Multiple networks are not supported")
+		}
+		d.ip = m.IPAddress
```
alban, line 148: Why not checking with `len(containerJSON.NetworkSettings.Networks)`?

gadgets/trace_dns/program.bpf.c
```diff
@@ -153,253 +188,146 @@ struct {
 	__uint(max_entries, 1024);
 } query_map SEC(".maps");
```
alban, line 189: I am wondering whether to implement the latency field in Wasm so the `query_map` bpf map would become a Go LRU map and it could be cleanup automatically.

At the moment, it seems easy to generate traffic in a way where `query_map` becomes full and the gadget stops working.

mauriciovasquezbernal: I think I agree with you. Is it fine if we make this change in another PR? (We'll need to think if we want to make the same change on the built-in version)

mauriciovasquezbernal:
Regarding the double structure definition, I played around with zero length arrays and got something working on https://github.com/inspektor-gadget/inspektor-gadget/tree/mauricio/experiments/zero-length-arrays. I think this approach could be easier to understand and to maintain than having macros or other stuff. What do you think about it? cc @flyth

gadgets/trace_dns/program.bpf.c
```diff
@@ -48,44 +89,50 @@ struct event_t {
...
+	// Only on this structure
+	__u8 data[MAX_PACKET];
+};
 
-	char name[MAX_DNS_NAME];
+// TODO: We need this header structure as the packet itself is appended by
+// bpf_perf_event_output(). Hence the even we send over the perf ring buffer is
+// only the header without the packet. We can use the full structure above as
+// it exceeds the stack limit of 512 bytes in bpf.
+// TODO: We'll need to find a clearer way to implement this
```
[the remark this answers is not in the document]
mauriciovasquezbernal: That's a possibility. I tried to create something but I wasn't too happy with the result as the syntax isn't easy to follow. I also tried to look for other solutions, I was only able to find this https://stackoverflow.com/questions/51585625/c-preprocessor-add-elements-to-struct-at-compile-time, but yes, the answer states the solution will be a mess.

Do you have a concrete idea of how to implement it?

gadgets/trace_dns/program.bpf.c
```diff
@@ -48,44 +89,50 @@ struct event_t {
...
-	// qr says if the dns message is a query (0), or a response (1)
-	unsigned char qr;
-	unsigned char pkt_type;
-	unsigned char rcode;
+	__u16 dns_off; // DNS offset in the packet
+	__u32 data_len;
 
-	__u64 latency_ns; // Set only if qr is 1 (response) and pkt_type is 0 (Host).
+	// Only on this structure
+	__u8 data[MAX_PACKET];
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I don't think it's the same situation:
- trace exec: We want the last argument to have a variable size, so we define it with a big size in the structure and then we dynamically change the number of bytes we send over the ring buffer. 
- trace dns: We want Inspektor Gadget to add a field called `data`. This field is never filled in the ebpf program, but the skb is sent through the perf ring buffer by using `skb_len << 32 | BPF_F_CURRENT_CPU` as argument to the `bpf_perf_event_output` function.

flyth:
> Regarding the double structure definition, I played around with zero length arrays and got something working on https://github.com/inspektor-gadget/inspektor-gadget/tree/mauricio/experiments/zero-length-arrays. I think this approach could be easier to understand and to maintain than having macros or other stuff. What do you think about it? cc @flyth

Yeah, why not - that's at least completely one in eBPF.

Another solution would be to add an annotation to the datasource like `ebpf.overflow=targetField`, that would - whenever the event from eBPF is larger than the defined struct - write all the overflow data to  that targetField. That could be dynamically sized and then contain the packet data. But this solution would require the metadata file and not be all-ebpf.

mauriciovasquezbernal:
I'd propose to move ahead with this PR and then fix the "double structure definition issue" as it needs more time.

mauriciovasquezbernal:
Thanks for the review. Failing test is not related to this PR, so I'll merge it.

PR #3029, opened 2024-06-18
trace_exec: drop uid filter on `sys_exit_execve`

i-Pear:
# trace_exec: drop uid filter on `sys_exit_execve`

As discussed in [PR 2965](https://github.com/inspektor-gadget/inspektor-gadget/pull/2965#discussion_r1636751077), if the uid changes between `sys_enter_execve` and `sys_exit_execve`, maps will not be cleaned properly.
This PR removed the uid check in `sys_exit_execve`, for both image-based and builtin gadgets.

PR #3100, opened 2024-06-28
trace_exec: remove unused variable `uid`

i-Pear:
# trace_exec: remove unused variable `uid`

In #3029, we removed the uid check in `ig_execve_x`, but some related variables were not removed completely.

Seems many gadgets have some unused variables, but if `ig` could compile them successfully, no log will be printed, and we will not see the compiler warnings.
Could we print the stderr from the compiler?


mauriciovasquezbernal:
> Seems many gadgets have some unused variables, but if `ig` could compile them successfully, no log will be printed, and we will not see the compiler warnings.
> Could we print the stderr from the compiler?

You're right, there are a lot of warnings. The output of the compiler can be enabled with:

```bash 
$ make -C gadgets/ GADGET_BUILD_PARAMS=-v build
```

I think it'll be a good change for the CI.

alban:
The CI didn't start for some unknown reason. I pushed your branch in `citest/i-Pear-exec_uid` to see if the CI says something.

alban:
All the tests passed except on ARO, but the failure is not related to this PR, so I will merge it.

PR #3108, opened 2024-07-02
Makefile: Fix how ig-debug version is set.

eiffel-fl:
Hi!

We forgot to update this line when we changed how the version is computed.

Best regards.


burak-ok:
> We forgot to update this line when we changed how the version is computed.

Could you also include a `Fixes` tag in the commit message if the "bug" can be pointed to a single commit?

mauriciovasquezbernal (approved):

burak-ok (approved):
Thanks

PR #3109, opened 2024-07-02
prometheus: Use different name for metrics config

mqasimsarfraz:
Use different name for metric flag for prometheus to avoid conflict with: https://github.com/inspektor-gadget/inspektor-gadget/pull/2831

cc: @eiffel-fl


mqasimsarfraz:
> I am not sure if this makes sense, but the file under docs/design/ was not updated:

I decided to skip it on purpose probably because the design docs should reflect the initial version of the implementation (because it will hard to keep them in sync) but I don't have strong opinion on this. Do you have any recommendation?

PR #3115, opened 2024-07-03
gadgets/trace_capabilities: Add test

### LINKED ISSUE #2992: gadgets: Implement trace capabilities as image based one
Port [trace capabilities](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/trace/capabilities) to be an image-based gadget.

burak-ok:
closes #2992

### DIFF
--- a/gadgets/trace_capabilities/test/trace_capabilities_test.go
+++ b/gadgets/trace_capabilities/test/trace_capabilities_test.go
@@ -1,0 +1,154 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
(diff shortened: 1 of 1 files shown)


gadgets/trace_capabilities/test/trace_capabilities_test.go
```diff
@@ -0,0 +1,150 @@
...
+	testContainer := containerFactory.NewContainer(containerName, "nice -n -20 echo", containerOpts...)
+
+	var runnerOpts []igrunner.Option
+	var testingOpts []igtesting.Option
+	commonDataOpts := []utils.CommonDataOption{
+		utils.WithContainerImageName(containerImage),
+		// TODO: We need to start the container after the tracer, so we don't
+		// have the container ID available here. It could be possible to split
+		// the logic to have a container create + container start to be able to
+		// get the ID before starting it.
```
mauriciovasquezbernal, line 83: I think this test can be implemented similar to https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_open/test/trace_open_test.go. (Running the container before the tracer and the test command within a loop).

gadgets/trace_capabilities/test/trace_capabilities_test.go
```diff
@@ -0,0 +1,154 @@
...
+				utils.NormalizeString(&e.Kstack)
+				utils.NormalizeString(&e.CapEffective)
+
+				// Manually normalize fields that might contain 0
+				e.Uid = utils.NormalizedInt
+				e.Gid = utils.NormalizedInt
+				e.Tgid = utils.NormalizedInt
+				e.CurrentUserNs = utils.NormalizedInt
+				e.TargetUserNs = utils.NormalizedInt
+				e.Insetid = utils.NormalizedInt
```
mauriciovasquezbernal, line 138: I don't think it makes sense to set these fields to `NormalizedInt`. The idea of this is to be able to check that a given field was present on the json, by assigning it here, that check is not possible. i.e. if the field wasn't present on the json output, the test will pass the same.

I see the following options:
1. Set all these fields to 0 and remove them from the expected entry
2. Remove them from `traceCapabilitiesEvent` so they aren't considered at all.

I think the first approach is better, it'll be able to catch if the fields exist on the json but are the wrong type.

mauriciovasquezbernal (approved):
This test is passing, so I'll approve it now.

PR #3116, opened 2024-07-03
gadgets/Makefile: Report failing tests again

burak-ok:
c6cd71ffe4765ccffba635f9501ef922a20486c0 modified running the tests through `find`, which does not report back failing subtasks we created with `--execdir`.

This PR moves the go source files for wasm into its own subfolder and reverts that commit


gadgets/Makefile
```diff
@@ -88,22 +88,36 @@ clean:
...
-	IG_FLAGS=$(IG_FLAGS) \
-	find . -name '*_test.go' -execdir go test -exec 'sudo -E' -v ./... \;
+	IG_FLAGS=$(IG_FLAGS); \
+	for dir in $$(find . -name '*_test.go' -exec dirname {} \;); do \
+	  (cd $$dir && go test -exec 'sudo -E' -v ./...); \
+	  if [ $$? -ne 0 ]; then \
+	    overall_success=1; \
+	  fi; \
+	done; \
+	exit $$overall_success
```
mauriciovasquezbernal, line 123: What about trying to factorize it now?

mauriciovasquezbernal (approved):
LGTM. Thanks for fixing it.

gadgets/Makefile
```diff
@@ -88,22 +88,36 @@ clean:
...
 	GADGET_TAG=$(GADGET_TAG) \
-	IG_FLAGS=$(IG_FLAGS) \
-	find . -name '*_test.go' -execdir go test -exec 'sudo -E' -v ./... \;
+	IG_FLAGS=$(IG_FLAGS); \
+	for dir in $$(find . -name '*_test.go' -exec dirname {} \;); do \
+	  (cd $$dir && go test -exec 'sudo -E' -v ./...); \
+	  if [ $$? -ne 0 ]; then \
+	    overall_success=1; \
+	  fi; \
+	done; \
```
mauriciovasquezbernal, line 104: Also, is there a way to run this on parallel? I added t.Parallel() to all tests but it didn't do anything as they're run sequentially here.

burak-ok: I actually don't know. But if we parallize it we also need to limit the number of concurrent jobs.

burak-ok: We could also move the wasm layer into a subfolder, so the go file and the godot files are not hindering our testruns we did with `go test`.

WDYT?

burak-ok:
@mauriciovasquezbernal I had to update the second commit which fixes the `oomkill` tests. @eiffel-fl introduced a `waitUntilPodReadyOrOOMKilledCommand` for the "old" test to fix the flakyness. I reused that

burak-ok:
I get failures in the OOMKill test because of: https://github.com/inspektor-gadget/inspektor-gadget/pull/3036/commits/a414c4d41045e607a78debdd1f11ecadfb492584

The field is not getting normalized because its value is 0

mauriciovasquezbernal:
> The field is not getting normalized because its value is 0

If the field is 0, then you don't need to normalize it, right?

gadgets/Makefile
mauriciovasquezbernal: It makes a lot of sense. I agree, it'd be better to have a wasm module with all wasm related files (go.mod, go.sum, program.go), and then we won't have this annoying issue of the test runs.

burak-ok:
The whole PR was reworked to include the idea mentioned in https://github.com/inspektor-gadget/inspektor-gadget/pull/3116#discussion_r1665461924 (Move `.go` files into its own subfolder)

gadgets/trace_exec/go/go.mod
mauriciovasquezbernal, line 1: The replace directive needs to be updated to :

`replace github.com/inspektor-gadget/inspektor-gadget => ../../../`

And go mod tidy should be run too. (it's nice that after it all dependencies from the test file are gone)

docs/devel/hello-world-gadget-wasm.md
```diff
@@ -18,9 +18,9 @@ you are familiar with that guide before continuing with this.
 
 ### Creating our first wasm program
 
-Create a `program.go` file in the same folder of the `program.bpf.c` one. As a
-first step, let's define the `init`, `start` and `stop` functions and emit some
-log messages from them:
+Create a folder named `go` next to the to the `program.bpf.c` file. In there
```
mauriciovasquezbernal, line 21: Duplicated "to the"

gadgets/trace_exec/go/go.mod
burak-ok: Well, the `go.sum` files are totally empty now :D

gadgets/Makefile
```diff
@@ -10,6 +10,7 @@ KUBECTL_GADGET ?= kubectl-gadget
 IG_FLAGS ?=
 COSIGN ?= cosign
 GADGETS = \
+	trace_bind \
```
mauriciovasquezbernal, line 13: ops. I opened #3129 fixing other issues with this gadget.

mauriciovasquezbernal (approved):
LGTM, however I think #3129 needs to go first for it to work properly. 

Thanks!

PR #3130, opened 2024-07-04
gadgets: Test endpoints too

mauriciovasquezbernal:
The trace_sni gadget doesn't provide any information about endpoints, so the previous TODO was wrong.

### DIFF
--- a/gadgets/trace_tcp/test/trace_tcp_test.go
+++ b/gadgets/trace_tcp/test/trace_tcp_test.go
@@ -32,16 +32,16 @@ import (
 type traceTCPEvent struct {
 	eventtypes.CommonData
 
-	// TODO: endpoints, see https://github.com/inspektor-gadget/inspektor-gadget/issues/3034
-
-	Task      string `json:"task"`
-	MountNsID uint64 `json:"mntns_id"`
-	Timestamp string `json:"timestamp"`
-	Pid       uint32 `json:"pid"`
-	Uid       uint32 `json:"uid"`
-	Gid       uint32 `json:"gid"`
-	NetNsID   uint64 `json:"netns"`
-	Type      string `json:"type"`
+	Src       utils.L4Endpoint `json:"src"`
+	Dst       utils.L4Endpoint `json:"dst"`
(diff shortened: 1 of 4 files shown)


pkg/testing/utils/types.go
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+// Define some types that are only used for testing purposes
+
+type L4Endpoint struct {
```
burak-ok, line 19: You could embed `L3Endpoint` here. The JSON would still stay all in one object and should work

gadgets/trace_tcp/test/trace_tcp_test.go
```diff
@@ -139,6 +175,11 @@ func TestTraceTCP(t *testing.T) {
 				utils.NormalizeString(&e.Timestamp)
 				utils.NormalizeInt(&e.Pid)
 				utils.NormalizeInt(&e.NetNsID)
+				// Checking the ports is a little bit complicated as successive
+				// calls to curl with --local-port fail because of
+				// https://github.com/curl/curl/issues/6288
```
burak-ok, line 180: :(

burak-ok (approved):
LGTM, thanks

pkg/testing/utils/types.go
mauriciovasquezbernal: That's true, but the initialization would need something like:

L4Enpodint {
L3Endpoint {
...
...
}
}

Since it's only for testing purposes, I prefer to keep it that way.

gadgets/trace_tcp/test/trace_tcp_test.go
mauriciovasquezbernal: I've been thinking about this. Perhaps we should be using something else and not existing binaries/images for our tests. One idea is to use python to generate the very specific events we need.

PR #3138, opened 2024-07-08
container-hook: simplify sys_exit_execve logic

alban:
# tp/sys_exit_execve: simplify logic

Don't rely on complex logic in tracepoint/syscalls/sys_exit_execve to
figure out the correct pid.
Don't rely on ctx->ret because it is not always reliable.

Instead, use two tracepoints in two different scenarios:

- tracepoint/sched/sched_process_exec for successful execve
- tracepoint/syscalls/sys_exit_execve for failed execve

Updated:
- [x] container-hook
- [x] builtin trace-exec gadget
- [x] builtin trace-capabilities gadget
- [x] image-based trace-exec gadget
- [x] image-based trace-capabilities gadget

## How to use

No changes

## Testing done

TBD


pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
```diff
@@ -220,8 +207,42 @@ int ig_execve_x(struct syscall_trace_exit *ctx)
 	if (len <= sizeof(*event))
 		bpf_perf_event_output(ctx, &events, BPF_F_CURRENT_CPU, event,
 				      len);
+
+	return 0;
+}
+
+// We use syscalls/sys_exit_execve only to trace failed execve
```
mauriciovasquezbernal, line 214: nit: comment that this program is needed in all cases, not only when ignore-errors == false. (It was my first impression at the beginning when looking at the PR)

pkg/gadgets/trace/exec/tracer/tracer.go
```diff
@@ -156,6 +156,11 @@ func (t *Tracer) install() error {
 		return fmt.Errorf("attaching enter tracepoint: %w", err)
 	}
 
+	t.enterLink, err = link.Tracepoint("sched", "sched_process_exec", t.objs.IgSchedExec, nil)
+	if err != nil {
+		return fmt.Errorf("attaching sched_process_exec tracepoint: %w", err)
+	}
+
```
mauriciovasquezbernal, line 163: This one took me a while. The tracer stopped working for me after ~2 minutes. After some investigation (blaming every single component you can imagine, from the terminal to the kernel), I found that this line is replacing the link on L154, causing the link for sys_enter_execve to be garbage collected. There should be a field for this one.

pkg/container-hook/bpf/execruntime.bpf.c
```diff
@@ -35,40 +35,6 @@ struct {
...
-//
-// We don't want to use bpf_for_each_map_elem() because it requires Linux 5.13.
-//
-// If several execve(2) are performed in parallel from different threads, only
-// one can succeed. The kernel will run the tracepoint syscalls/sys_exit_execve
-// for the failing execve(2) first and then for the successful one last.
-//
-// So we can insert a tgid->pid mapping in the same hash entry by adding
-// the pid in value and removing it by subtracting. By the time we need to
-// lookup the pid by the tgid, there will be only one pid left in the hash entry.
```
mauriciovasquezbernal, line 60: I think this comment is valuable. Why not to adapt indicating why two different tracepoints need to be used for the exit?

mauriciovasquezbernal (commented):
It looks very good and I like the cleanup, the new code is easier to follow.

What about splitting it in different commits? If something goes wrong, it'll be difficult to understand what it was with a this commit changing many things. I also think it's fairly easy to split on:
- changes to container hook 
- changes to trace exec (built-in and image based can go in the same commit IMO)
- changes to trace capabilities

WDYT?

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
```diff
@@ -166,44 +163,34 @@ static __always_inline bool has_upper_layer(struct inode *inode)
...
-		if (inode) {
-			event->upper_layer = has_upper_layer(inode);
-		}
-
-		struct inode *pinode =
-			BPF_CORE_READ(parent, mm, exe_file, f_inode);
-		if (pinode) {
-			event->pupper_layer = has_upper_layer(pinode);
-		}
+	bpf_map_delete_elem(&execs, &execs_lookup_key);
```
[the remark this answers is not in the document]
alban: I thought it was safe because some other ebpf programs in the Linux tree have this pattern with the comment that the value are still accessible under rcu:
https://github.com/torvalds/linux/blob/933069701c1b507825b514317d4edd5d3fd9d417/tools/testing/selftests/bpf/progs/bpf_iter_bpf_hash_map.c#L66-L71

There are similar explanations at:
https://github.com/xdp-project/xdp-tutorial/issues/203

However, this thread seems to say something else (I am not sure to understand it):
https://lkml.org/lkml/2018/10/3/744

I find the code easier to read when bpf_map_delete_elem is done early, so we don't have to remember to delete the entry for each `return;`. What do you think?

pkg/gadgets/trace/exec/tracer/tracer.go
alban: Well spotted. Sorry for the mistake!

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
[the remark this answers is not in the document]
mauriciovasquezbernal: I personally prefer to have the delete operation at the end, it's a bit difficult to understand that you can continue to use a value after is has been deleted. 

In any case, the PR LGTM, I'll approve it and it's up to you which approach to use.

mauriciovasquezbernal (approved):
LGTM. Thanks for this improvement!

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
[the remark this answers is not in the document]
alban: I moved the bpf_map_delete_elem at the end. I will merge once the CI completes.

Issue #1567, opened 2023-04-24
gadgets: Standardize common info gadgets provide 


mauriciovasquezbernal:
We should be sure that all gadgets provide the same common information (when technically possible for sure)

This is the list of common fields all gadgets should provide: ( ❓ indicates that it needs more discussion)

- Kubernetes 
  - ✅ node
  - ✅ namespace
  - ✅ pod 
  - ❓pod uuid
  - ✅ container name
  - ✅ containerID
  - ✅ container image (tag and digest according to https://github.com/inspektor-gadget/inspektor-gadget/issues/1310)
  - ❓pid1 of the container
- container runtime
  - ✅ runtime name 
  - ✅ container name
  - ✅ containerID
  - ✅ container image (same as above in k8s)
-  process related
   - ppid (#2645)
   - pid  
   - tid 
   - comm 
   - pcomm
   - user related
     - uid https://github.com/inspektor-gadget/inspektor-gadget/pull/1666
     - gid https://github.com/inspektor-gadget/inspektor-gadget/pull/1666

### Implementation options 

We could create some common data structures to reuse in the different gadgets (https://github.com/inspektor-gadget/inspektor-gadget/issues/250). Something like could work:
- K8sInfo
- ContainerInfo
- ProcessInfo
- UserInfo

blixtra:
I think we should consider prioritizing this. We are getting adhoc requests from users to add a field here and there. It would be better to do this as part of a core set of data that we could add across the board.

alban:
I updated the issue description above with ✅ (when implemented already) and ❓(when more discussion required).

There are two tasks in this issue:
1. Adding missing fields in some gadgets
2. Making group of fields in generic structs

Categories of fields:

* ✅ **Kubernetes** and **container runtime**: the implementation is already generic (and not per-gadget)
* 🚧 **process related**: this is currently per-gadget. fsnotify added a `struct process` (#3438) but the struct definition is so far in the fsnotify gadget and not for all gadgets. Before making it more generic, we need to fix the following:
  * #3459 
  * #3460
  * #2684

Regardless that the struct process is not generic, here is the current status of some fields for some gadgets:
-  ✅ implemented
- 🚧 work in progress
- ❌ not implemented
- ❓need to check current status

| gadget | ppid | pid | tid | pcomm | comm | uid/gid | cwd | exepath |
|---|---|---|---|---|---|---|---|---|
[...]

PR #3149, opened 2024-07-10
gadgets: Forward config to datasources

### LINKED ISSUE #3052: formatters: Carry annotations and attributes
Some fields are converted to a human readable representation by the formatters operator. The gadgets can define some attributes for those fields on the metadata file:

https://github.com/inspektor-gadget/inspektor-gadget/blob/a199d26f59cb0b70a83a8989b3182f891aeb8169/gadgets/trace_mount/gadget.yaml#L50-L53

but those attributes aren't being applied to the field created by the formatter:

https://github.com/inspektor-gadget/inspektor-gadget/blob/a199d26f59cb0b70a83a8989b3182f891aeb8169/pkg/operators/ebpf/formatters.go#L90-L93

Those attributes should be carried on to avoid formatting issues.

Ref #3023

mauriciovasquezbernal:
This PR changes the way how the configuration of the fields is specified in the metadata file. It introduces a new datasources section that allows to (1) set annotations for the data source (will be used later for toppers and other support), (2) specify the 
field configuration. 

Check individual commits for more information. 

### Testing 

Run the gadgets and check that the output makes sense. Also, try changing the annotations in the metadata and check that they have an effect in the way fields are printed.

### Next 
There are some things we could do later on:
- [ ] Automatically apply the annotations based on the field type, for instance gadget_mntns_id gadget_timestamp, etc.
- [ ] Fix the logic that updates the metadata file.

Fixes #3052

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1378,7 +1378,7 @@ jobs:
           export GADGET_TAG=${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}
           export BUILDER_IMAGE=${{ needs.build-helper-images.outputs.ebpf_builder_image }}
 
-          make GADGET_BUILD_PARAMS="--update-metadata" build-gadgets -o install/ig
+          make build-gadgets -o install/ig
 
           # Check that metadata files are updated
           git diff --exit-code HEAD --
--- a/gadgets/snapshot_process/gadget.yaml
+++ b/gadgets/snapshot_process/gadget.yaml
@@ -3,40 +3,42 @@ description: Show running processes
 homepageURL: https://inspektor-gadget.io/
 documentationURL: https://inspektor-gadget.io/docs
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+datasources:
+  processes:
+    fields:
+      comm:
+        annotations:
+          template: comm
+      pid:
+        annotations:
+          description: Process ID
+          template: pid
+      tid:
+        annotations:
+          description: Thread ID
+          template: pid
+      ppid:
+        annotations:
+          description: Parent process ID
+          template: pid
+      uid:
+        annotations:
+          description: User ID
+          template: uid
+      gid:
+        annotations:
+          description: Group ID
+          template: uid
+      mntns_id:
+        annotations:
+          description: Mount namespace inode id
+          template: ns
 snapshotters:
   processes:
     structName: process_entry
 structs:
   process_entry:
     fields:
-    - name: comm
-      description: Process name
-      attributes:
-        template: comm
-    - name: pid
-      description: Process ID
-      attributes:
-        template: pid
-    - name: tid
-      description: Thread ID
-      attributes:
-        template: pid
-    - name: ppid
-      description: Parent process ID
-      attributes:
-        template: pid
-    - name: uid
-      description: User ID
-      attributes:
-        template: uid
-    - name: gid
-      description: Group ID
-      attributes:
-        template: uid
-    - name: mntns_id
-      description: Mount namespace inode id
-      attributes:
-        template: ns
 ebpfParams:
   show_threads:
     key: threads
--- a/gadgets/snapshot_socket/gadget.yaml
+++ b/gadgets/snapshot_socket/gadget.yaml
@@ -3,36 +3,32 @@ description: Show TCP and UDP sockets
 homepageURL: https://inspektor-gadget.io/
 documentationURL: https://inspektor-gadget.io/docs
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+datasources:
+  sockets:
+    fields:
+      src:
+        annotations:
+          description: Source address
+          template: l4endpoint
+      dst:
+        annotations:
+          description: Destination address
+          template: l4endpoint
+      state:
+        annotations:
+          columns.width: 10
+      ino:
+        annotations:
+          description: Inode number
+          columns.width: 10
+          columns.hidden: true
+      netns:
+        annotations:
+          description: Network namespace inode id
+          template: ns
 snapshotters:
   sockets:
     structName: socket_entry
 structs:
   socket_entry:
     fields:
-    - name: src
-      description: Source address
-      attributes:
-        minWidth: 24
-        maxWidth: 50
-    - name: dst
-      description: Destination address
-      attributes:
-        minWidth: 24
-        maxWidth: 50
-    - name: state
-      attributes:
-        width: 10
-        alignment: left
-        hidden: true
-        ellipsis: end
-    - name: ino
-      description: Inode number
-      attributes:
-        width: 10
-        alignment: left
-        hidden: true
-        ellipsis: end
-    - name: netns
-      description: Network namespace inode id
-      attributes:
-        template: ns
--- a/gadgets/trace_bind/gadget.yaml
+++ b/gadgets/trace_bind/gadget.yaml
@@ -3,6 +3,50 @@ description: Trace socket bindings
 homepageURL: https://inspektor-gadget.io/
 documentationURL: https://inspektor-gadget.io/docs
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+datasources:
+  bind:
+    fields:
+      timestamp_raw:
+        annotations:
+          columns.hidden: true
+      timestamp:
+        annotations:
+          template: timestamp
+      mount_ns_id:
+        annotations:
+          template: ns
+      addr:
+        annotations:
+          description: Address the socket is bound to
+          columns.width: 32
+      pid:
+        annotations:
+          template: pid
+      uid:
+        annotations:
+          template: uid
+      gid:
+        annotations:
+          template: uid
+      ret:
+        annotations:
+          description: Return value of the bind() syscall
+          columns.width: 5
+          columns.alignment: right
+      opts_raw:
+        annotations:
+          columns.hidden: true
+      opts:
+        annotations:
+          description: Socket options
+          columns.hidden: true
+      comm:
+        annotations:
+          template: comm
+      bound_dev_if:
+        annotations:
+          description: Index of the network interface the socket is bound to
+          columns.width: 10
 tracers:
   bind:
     mapName: events
@@ -10,46 +54,6 @@ tracers:
 structs:
   event:
     fields:
-    - name: timestamp_raw
-    - name: mount_ns_id
-      description: Mount namespace inode id
-      attributes:
-        template: ns
-    - name: addr
-      description: 'TODO: Fill field description'
-      attributes:
-        width: 16
-        alignment: left
-        ellipsis: end
-    - name: pid
-      attributes:
-        template: pid
-    - name: uid
-      attributes:
-        template: uid
-    - name: gid
-      attributes:
-        template: uid
-    - name: ret
-      description: Return value of the bind() syscall
-      attributes:
-        width: 3
-        alignment: left
-    - name: opts_raw
-      description: Socket options
-      attributes:
-        width: 16
-        alignment: left
-        ellipsis: end
-    - name: comm
-      attributes:
-        template: comm
-    - name: bound_dev_if
-      description: Index of the network interface the socket is bound to
-      attributes:
-        width: 10
-        alignment: left
-        ellipsis: end
 ebpfParams:
   ignore_errors:
     key: ignore_errors
--- a/gadgets/trace_capabilities/gadget.yaml
+++ b/gadgets/trace_capabilities/gadget.yaml
@@ -3,6 +3,77 @@ description: trace security capabilitiy checks
 homepageURL: https://inspektor-gadget.io/
 documentationURL: https://inspektor-gadget.io/docs
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+datasources:
+  capabilities:
+    fields:
+      timestamp_raw:
+        annotations:
+          columns.hidden: true
+      timestamp:
+        annotations:
+          template: timestamp
+      mntnsid:
+        annotations:
+          template: ns
+      current_userns:
+        annotations:
+          template: ns
+      target_userns:
+        annotations:
+          template: ns
+      cap_effective_raw:
+        annotations:
+          columns.hidden: true
+      cap_effective:
+        annotations:
+          columns.width: 20
+          columns.hidden: true
+      pid:
+        annotations:
+          template: pid
+      cap_raw:
+        annotations:
+          columns.hidden: true
+      cap:
+        annotations:
+          columns.hidden: true
+      tgid:
+        annotations:
+          template: pid
+      uid:
+        annotations:
+          template: uid
+      gid:
+        annotations:
+          template: uid
+      audit:
+        annotations:
+          columns.width: 11
+      insetid:
+        annotations:
+          columns.width: 11
+      syscall_raw:
+        annotations:
+          columns.hidden: true
+      syscall:
+        annotations:
+          columns.width: 20
+      task:
+        annotations:
+          template: comm
+      kstack_raw:
+        annotations:
+          columns.hidden: true
+      kstack:
+        annotations:
+          description: kernel stack
+          columns.width: 10
+          columns.hidden: true
+      capable:
+        annotations:
+          description: if the process has the requested capability
+          columns.width: 10
+          columns.hidden: true
 tracers:
   capabilities:
     mapName: events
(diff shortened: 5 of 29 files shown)


gadgets/trace_capabilities/gadget.yaml
```diff
@@ -3,6 +3,89 @@ description: trace security capabilitiy checks
...
+      cap_effective:
+        annotations:
+          columns.width: 20
+          columns.alignment: left
+          columns.ellipsis: end
+          columns.hidden: true
+      pid:
+        annotations:
+          template: pid
+      cap_raw:
```
flyth, line 36: Missing cap?

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
 homepageURL: https://inspektor-gadget.io/
 documentationURL: https://inspektor-gadget.io/docs/latest/
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget
+datasources:
+  dns:
+    fields:
+      timestamp_raw:
+        annotations:
+          columns.hidden:true
```
flyth, line 11: Missing space

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
...
+          template: timestamp
+      src:
+        annotations:
+          template: l4endpoint
+      dst:
+        annotations:
+          template: l4endpoint
+      pid:
+        annotations:
+          template: pid
```
flyth, line 23: Missing description (`PID of the process that sent the request`)

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
...
+          columns.hidden:true
+      timestamp:
+        annotations:
+          template: timestamp
+      src:
+        annotations:
+          template: l4endpoint
+      dst:
+        annotations:
+          template: l4endpoint
```
flyth, line 20: Missing description (`Destination endpoint`)

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
...
+    fields:
+      timestamp_raw:
+        annotations:
+          columns.hidden:true
+      timestamp:
+        annotations:
+          template: timestamp
+      src:
+        annotations:
+          template: l4endpoint
```
flyth, line 17: Missing description (`Source endpoint`)

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
...
+          columns.width: 8
+      latency_ns:
+        annotations:
+          columns.width: 8
+          columns.hidden: true
+      anaddr:
+        annotations:
+          columns.width: 16
+      netns:
+        annotations:
```
flyth, line 55: Missing description (`Network namespace inode id`)

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
...
+          template: ns
+      mntns_id:
+        annotations:
+          template: ns
```
flyth, line 59: Missing description (`Mount namespace inode id`)

gadgets/trace_dns/gadget.yaml
```diff
@@ -3,97 +3,93 @@ description: trace dns requests and responses
...
+      tid:
+        annotations:
+          columns.width: 16
+          columns.alignment: left
+          columns.hidden: true
+          columns.ellipsis: end
```
flyth, line 65: These look like the default values - should we just drop them everywhere?

gadgets/trace_exec/gadget.yaml
```diff
@@ -3,87 +3,78 @@ description: trace process executions
...
+        annotations:
+          template: comm
+      uid:
+        annotations:
+          template: uid
+          columns.hidden: true
+      gid:
+        annotations:
+          template: uid
+          columns.hidden: true
```
flyth, line 31: Weren't hidden before

gadgets/trace_lsm/gadget.yaml
```diff
@@ -3,40 +3,49 @@ description: a strace for LSM tracepoints
...
+  lsm:
+    fields:
+      timestamp_raw:
+        annotations:
+          columns.hidden: true
+      timestamp:
+        annotations:
+          template: timestamp
+      mntns_id:
+        annotations:
```
flyth, line 16: Missing description (`Mount namespace inode id`)

flyth (commented):
Started reviewing, but maybe it makes sense to directly fix and push? WDYT?

gadgets/snapshot_process/gadget.yaml
```diff
@@ -3,40 +3,42 @@ description: Show running processes
 homepageURL: https://inspektor-gadget.io/
 documentationURL: https://inspektor-gadget.io/docs
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
+datasources:
+  processes:
+    fields:
+      comm:
+        annotations:
+          template: comm
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Added it directly on the template so we don't need to repeat it for all gadgets.

We can extend it for pid, gid, tid, uid, etc, but I think we can handle it altogether with #1567, I'll investigate more on the following weeks.

gadgets/snapshot_socket/gadget.yaml
```diff
@@ -3,36 +3,36 @@ description: Show TCP and UDP sockets
...
+          description: Source address
+          template: l4endpoint
+      dst:
+        annotations:
+          description: Destination address
+          template: l4endpoint
+      state:
+        annotations:
+          columns.width: 10
+          columns.alignment: left
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'll ignore as this field requires more work: #3154

gadgets/trace_exec/gadget.yaml
mauriciovasquezbernal: I tuned it because this gadget is printing too many things. (In any case, I think we'll need to make a pass later on tuning all gadgets)

gadgets/trace_lsm/gadget.yaml
mauriciovasquezbernal: I'll ignore it, IMO this should be automatically added by Inspektor Gadget giving that we know the type of the field. (I tried to make it, but it was too much for this PR).

gadgets/trace_dns/gadget.yaml
mauriciovasquezbernal: I defined a new variable with default annotations and dropped all defaults from all gadget.yaml files.

gadgets/trace_bind/gadget.yaml
```diff
@@ -3,53 +3,63 @@ description: Trace socket bindings
...
+      opts_raw:
+        annotations:
+          columns.hidden: true
+      opts:
+        annotations:
+          description: Socket options
+          columns.width: 16
+          columns.alignment: left
+          columns.ellipsis: end
+          columns.hidden: true
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Yes, it's too much information to fit on the screen.

gadgets/snapshot_process/gadget.yaml
```diff
@@ -3,40 +3,42 @@ description: Show running processes
...
+        description: Process ID
+        annotations:
+          template: pid
+      tid:
+        description: Thread ID
+        annotations:
+          template: pid
+      ppid:
+        annotations:
+          description: Parent process ID
```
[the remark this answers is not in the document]
mauriciovasquezbernal: And in many other places. I did a pass, I hope I didn't miss anything this time.

gadgets/trace_bind/gadget.yaml
```diff
@@ -3,53 +3,63 @@ description: Trace socket bindings
...
+      timestamp:
+        annotations:
+          template: timestamp
+      mount_ns_id:
+        annotations:
+          template: ns
+      addr:
+        annotations:
+          description: Address the socket is bound to
+          columns.width: 32
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Yeah, but I think we should handle it with a template later on too. (I won't handle in this PR because the endpoints behave a little bit different and the template wasn't working for me).

pkg/datasource/data.go
```diff
@@ -427,25 +429,68 @@ func (ds *dataSource) AddStaticFields(size uint32, fields []StaticField) (FieldA
...
+	applyTemplate := func(f *field, annotations map[string]string) {
+		templateAnn := annotations[TemplateAnnotation]
+		if templateAnn == "" {
+			return
+		}
+
+		template := annotationsTemplates[templateAnn]
+		if template == nil {
+			log.Warnf("template %q not found for field %q", templateAnn, f.Name)
+			return
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's fine not to return an error. Even without a template, the gadget will be able to provide the information. Also, the user can't do anything about this error, given that it's a mistake from the gadget developer.

pkg/datasource/data.go
```diff
@@ -427,25 +429,68 @@ func (ds *dataSource) AddStaticFields(size uint32, fields []StaticField) (FieldA
...
+		}
+
+		log.Debugf("applying template %q to field %q", templateAnn, f.Name)
+		for k, v := range template {
+			f.Annotations[k] = v
+		}
 	}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Not really, only advantage that I see is that we don't pollute the module with this function name. I'll move it to be a function as I think it's clearer.

pkg/datasource/data.go
```diff
@@ -427,25 +429,68 @@ func (ds *dataSource) AddStaticFields(size uint32, fields []StaticField) (FieldA
...
+	// first pass applying configuration from metadata
 	for _, field := range newFields {
-		sub := fields.Sub(field.Name)
-		if sub == nil {
+		// TODO: set defaults
+
+		// apply template if setted
+		applyTemplate(field, field.Annotations)
+
+		// apply configuration from metadata
```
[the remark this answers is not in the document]
mauriciovasquezbernal: mmm, I think it's right. Why do you think it's misplaced?

pkg/gadget-context/gadget-context.go
```diff
@@ -191,10 +192,22 @@ func (c *GadgetContext) DataOperators() []operators.DataOperator {
...
+
+	options := make([]datasource.DataSourceOption, 0)
+	if cfg, ok := c.GetVar("config"); ok {
+		if v, ok := cfg.(*viper.Viper); ok {
+			sub := v.Sub("datasources." + name)
+			if sub != nil {
+				options = append(options, datasource.WithConfig(sub))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I don't see a big difference between both approaches. I'll ignore this.

.github/workflows/inspektor-gadget.yml
```diff
@@ -1426,12 +1426,21 @@ jobs:
         cache: true
       id: go
     - name: Gadgets tests
+      id: gadgets-tests
```
blanquicet, line 1429: If you put this commit in a separated PR, we can merge it straight away.

pkg/datasource/data.go
```diff
@@ -159,28 +160,43 @@ type dataSource struct {
...
+	return ds, nil
 }
 
-func New(t Type, name string) (DataSource, error) {
-	return newDataSource(t, name)
+func New(t Type, name string, options ...DataSourceOption) (DataSource, error) {
+	return newDataSource(t, name, options...)
 }
 
 func NewFromAPI(in *api.DataSource) (DataSource, error) {
```
blanquicet, line 203: Load in `NewFromAPI` also the annotations (also the tags), otherwise they will be sent from server to client, but operators in the client side will not see them.

pkg/datasource/data.go
```diff
@@ -159,28 +160,43 @@ type dataSource struct {
...
-	}, nil
+	}
+
+	for _, option := range options {
+		option(ds)
+	}
+
+	if ds.config != nil {
+		// Apply configuration
+		annotations := ds.config.GetStringMapString("annotations")
```
blanquicet, line 191: What about tags?

pkg/datasource/data.go
```diff
@@ -426,26 +428,67 @@ func (ds *dataSource) AddStaticFields(size uint32, fields []StaticField) (FieldA
...
+	if ds.config != nil {
+		fields = ds.config.Sub("fields")
+	}
+
+	// first pass applying configuration from metadata
 	for _, field := range newFields {
-		sub := fields.Sub(field.Name)
-		if sub == nil {
+		// apply template if setted
+		applyFieldTemplate(field, field.Annotations)
```
blanquicet, line 459: This is when, for instance, an operator adding a new field, adds also a template annotation for it?

pkg/datasource/data.go
```diff
@@ -178,7 +179,7 @@ func newDataSource(t Type, name string, options ...DataSourceOption) (*dataSourc
 		fieldMap:        make(map[string]*field),
 		byteOrder:       binary.NativeEndian,
 		tags:            make([]string, 0),
-		annotations:     map[string]string{},
+		annotations:     maps.Clone(defaultFieldAnnotations),
```
blanquicet, line 182: Mmmmm why does the ds need the default field annotations? I think it's wrong.

pkg/datasource/columns.go
```diff
@@ -197,3 +205,67 @@ func (ds *dataSource) Columns() (*columns.Columns[DataTuple], error) {
...
 	return cols, nil
 }
+
+var defaultFieldAnnotations = map[string]string{
+	ColumnsWidthAnnotation:     "16",
+	ColumnsEllipsisAnnotation:  string(metadatav1.EllipsisEnd),
+	ColumnsAlignmentAnnotation: string(metadatav1.AlignmentLeft),
+}
+
+var annotationsTemplates = map[string]map[string]string{
```
blanquicet, line 215: What about [syscall template](https://github.com/inspektor-gadget/inspektor-gadget/blob/2b4e5619a0cdb68e1ef2df0f21ea641169e2df94/pkg/types/types.go#L58)?

blanquicet (commented):
I reviewed the general approach and LGTM. I didn't review the changes in the metadata files. Next week, I'll run some gadgets to check the output and make some changes in the annotations to verify it's applied. We should implement a quick test for it. Maybe a metadata file that hides all columns and another with all visible, or something like that. It's just to check that annotations are being considered. WDYT?

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: Already done in #3152.

pkg/datasource/data.go
mauriciovasquezbernal: Isn't it already done? The `(*field)(f)` refers to field that contains annotations and tags.

pkg/datasource/data.go
mauriciovasquezbernal: Exactly. I added a comment that hopefully makes it more clear.

pkg/datasource/columns.go
mauriciovasquezbernal: good catch! There are still some templates for ipaddr, etc. that I won't add for now as I think with the new endpoint structure they aren't needed.

mauriciovasquezbernal:
> We should implement a quick test for it. Maybe a metadata file that hides all columns and another with all visible, or something like that. It's just to check that annotations are being considered. 

I added a test about it, but I'm not sure it's good. It seems to me it's testing a lot of things not related to the data sources. @flyth any opinion on this one?

flyth:
> > We should implement a quick test for it. Maybe a metadata file that hides all columns and another with all visible, or something like that. It's just to check that annotations are being considered.
> 
> I added a test about it, but I'm not sure it's good. It seems to me it's testing a lot of things not related to the data sources. @flyth any opinion on this one?

I also think it's doing too much (and the dependency on Parser/Columns is unnecessary IMHO). I think just checking the ds/field with `GetAnnotations()` should be enough.

mauriciovasquezbernal:
> I also think it's doing too much (and the dependency on Parser/Columns is unnecessary IMHO). I think just checking the ds/field with GetAnnotations() should be enough.

Makes sense. Updated it to only test the annotations and the flags.

pkg/datasource/data.go
blanquicet: You're right. These are annotations on the fields, not in the ds itself.

gadgets/snapshot_socket/gadget.yaml
```diff
@@ -3,36 +3,32 @@ description: Show TCP and UDP sockets
...
+      netns:
+        annotations:
+          description: Network namespace inode id
+          template: ns
 snapshotters:
   sockets:
     structName: socket_entry
 structs:
   socket_entry:
     fields:
```
blanquicet, line 34: Do you think we'll use this `structs` field in the future? If you don't, why we're not removing it?

pkg/datasource/columns.go
```diff
@@ -197,3 +205,71 @@ func (ds *dataSource) Columns() (*columns.Columns[DataTuple], error) {
...
+		ColumnsHiddenAnnotation: "true",
+		ColumnsWidthAnnotation:  "35",
+	},
+	"comm": {
+		DescriptionAnnotation:     "Process name",
+		ColumnsMaxWidthAnnotation: "16",
+	},
+	"pid": {
+		ColumnsMinWidthAnnotation:  "7",
+		ColumnsAlignmentAnnotation: string(metadatav1.AlignmentRight),
```
blanquicet, line 251: nit: Why alignment right for these fields?

blanquicet (commented):
Tests are enough also for me. I tested it by adding some annotations to change format and verifying it was applied on top of template. I just have one question about the `structs` field in the yaml but otherwise LGTM.

gadgets/snapshot_socket/gadget.yaml
mauriciovasquezbernal: I want to get this PR as soon as possible in, further cleanups will be done later on, like #3164.

pkg/datasource/columns.go
mauriciovasquezbernal: I find it easier to read integer values when they are aligned to the right.

blanquicet: Ok.

blanquicet (approved):
LGTM!

PR #3150, opened 2024-07-10
config: Add support for ig-k8s

mqasimsarfraz:
This PR allows specifying configuration related to ig-k8s using a configmap. It is mainly split into handling couple of different type of config:

1. config related to gadgettracermanager e.g (`hookMode`, `fallbackPodInformer`, etc) 
2. global params related to operators (currently we don't have any but will allow specify oci params: https://github.com/inspektor-gadget/inspektor-gadget/pull/2902)

Fixes: #2493 

## Testing Done

Since I haven't removed the environment variable, we need to delete the env from the deployment and validate if config is working fine:

```
$ kubectl set env daemonsets -n gadget gadget INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER-
# edit (kubectl edit) the config to disable the pod-informer
$ kubectl get -n gadget cm gadget  -o yaml
apiVersion: v1
data:
  config.yaml: |-
    hook-mode: auto
    fallback-pod-informer: false
    events-buffer-length: 16384
$ kubectl rollout restart -n gadget daemonset
$ ps aux | grep gadgettracermanager
root     1394179  1.6  0.2 5574072 146304 ?      Ssl  18:34   0:00 gadgettracermanager -serve -hook-mode=auto -controller -fallback-podinformer=false
```

Edit: Config file was more priority!

### DIFF
--- a/charts/gadget/templates/configmap.yaml
+++ b/charts/gadget/templates/configmap.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  {{- if not .Values.skipLabels }}
+  labels:
+    {{- include "gadget.labels" . | nindent 4 }}
+  {{- end }}
+  name: {{ include "gadget.fullname" . }}
+  namespace: {{ include "gadget.namespace" . }}
+data:
+    config.yaml: |-
+      hook-mode: {{ .Values.config.hookMode }}
+      fallback-pod-informer: {{ .Values.config.fallbackPodInformer }}
+      events-buffer-length: {{ .Values.config.eventsBufferLength }}
+      containerd-socketpath: {{ .Values.config.containerdSocketPath }}
+      crio-socketpath: {{ .Values.config.crioSocketPath }}
+      docker-socketpath: {{ .Values.config.dockerSocketPath }}
+      podman-socketpath: {{ .Values.config.podmanSocketPath }}
--- a/cmd/ig/daemon.go
+++ b/cmd/ig/daemon.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/config"
 	gadgetservice "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
@@ -94,6 +95,10 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 
 		log.Infof("starting Inspektor Gadget daemon at %q", socket)
 		service.SetEventBufferLength(eventBufferLength)
+
+		if err = config.Config.ReadInConfig(); err != nil {
+			log.Warnf("reading config: %v", err)
+		}
 
 		return service.Run(gadgetservice.RunConfig{
 			SocketType: socketType,
--- a/gadget-container/gadgettracermanager/main.go
+++ b/gadget-container/gadgettracermanager/main.go
@@ -42,6 +42,8 @@ import (
 
 	// This is a blank include that actually imports all gadgets
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/all-gadgets"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/config"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/config/gadgettracermanagerconfig"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	ocihandler "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/oci-handler"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/experimental"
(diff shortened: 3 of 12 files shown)


charts/gadget/templates/configmap.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    {{- include "gadget.labels" . | nindent 4 }}
+  {{- end }}
+  name: {{ include "gadget.fullname" . }}
+  namespace: {{ include "gadget.namespace" . }}
+data:
+    config.yaml: |-
+      hook-mode: {{ .Values.config.hookMode }}
+      fallback-pod-informer: {{ .Values.config.fallbackPodInformer }}
+      events-buffer-length: {{ .Values.config.eventsBufferLength }}
```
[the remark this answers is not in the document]
mqasimsarfraz: Good point, I will update the PR to include them!

mqasimsarfraz:
> I also do not understand how you can modify the values of configmap.yaml from kubectl-gadget deploy.
Particularly, I do not see any code in deploy.go which would propagate the flag value to the config.

We are indeed missing that, currently it only allows modifying configmap using `helm` but the idea would be same as env like we will use the flags to set the values in data section of configmap. I will push those changes soon!

Different topic, do you think we need to env like `INSPEKTOR_GADGET_OPTION_FALLBACK_POD_INFORMER` around? I was wondering if we should switch to config directly?

cmd/ig/daemon.go
```diff
@@ -95,6 +96,10 @@ func newDaemonCommand(runtime runtime.Runtime) *cobra.Command {
 		log.Infof("starting Inspektor Gadget daemon at %q", socket)
 		service.SetEventBufferLength(eventBufferLength)
 
+		if err = config.Config.ReadInConfig(); err != nil {
+			log.Debugf("reading config: %v", err)
```
mauriciovasquezbernal, line 100: Warn?

gadget-container/gadgettracermanager/main.go
```diff
@@ -340,7 +347,10 @@ func main() {
 
 		stringBufferLength := os.Getenv("EVENTS_BUFFER_LENGTH")
 		if stringBufferLength == "" {
-			log.Fatalf("Environment variable EVENTS_BUFFER_LENGTH not set")
+			stringBufferLength = config.Config.GetString(gadgettracermanagerconfig.EventsBufferLengthKey)
+		}
+		if stringBufferLength == "" {
+			log.Fatalf("Environment variable EVENTS_BUFFER_LENGTH or config not set")
```
mauriciovasquezbernal, line 353: What's the reason to keep the env variables? Compatibility? 

I think it's fine to keep them, but:
1. The configuration file should have precedence 
2. We should print warnings indicating those variables are deprecated, and remove them later on.

mauriciovasquezbernal (commented):
I was reviewing it without first checking the comments, and you already asked all questions  I have. 

Just want to clarify how do you envision the deploy flags working with all of this?

Are we keeping flags like `--hook-mode`, `--fallback-podinformer`, etc? What if I have all those things in a configuration file? Can I use something like `kubectl gadget deploy --gadget-tracer-manager-config myconfig.yaml`? Or should the user create that config map manually before deploying Inspektor Gadget?

I don't have supper strong opinions, but I'd propose:
- Keep the flags in kubectl gadget deploy (later on we can deprecate them if we want)
- Provide an option to specify the config file to use for the gadgettracermanager at deploy time
- Keep the env variables but deprecate them (printing warning if they're set in the gadget container)

WDYT?

mqasimsarfraz:
Thanks for quick reviews folks! 

> I was reviewing it without first checking the comments, and you already asked all questions I have.
> 
> Just want to clarify how do you envision the deploy flags working with all of this?
> 
> Are we keeping flags like `--hook-mode`, `--fallback-podinformer`, etc? What if I have all those things in a configuration file? Can I use something like `kubectl gadget deploy --gadget-tracer-manager-config myconfig.yaml`? Or should the user create that config map manually before deploying Inspektor Gadget?

I was thinking of keeping the flags and using them to manipulate the generated configmap (I just pushed the changes). Similarly it will be handled by [values.yaml](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/charts/gadget/values.yaml) in case for helm! 

> I don't have supper strong opinions, but I'd propose:
> 
> * Keep the flags in kubectl gadget deploy (later on we can deprecate them if we want)
> * Provide an option to specify the config file to use for the gadgettracermanager at deploy time
> * Keep the env variables but deprecate them (printing warning if they're set in the gadget container)
> 

I have more or less similar opinions except I haven't introduced the config file `--gadget-tracer-manager-config`. I was thinking we keep the flags since it allow quick way to configure installation and if users want to put their config in version control they can use `values.yaml` from helm!

mauriciovasquezbernal:
> I have more or less similar opinions except I haven't introduced the config file --gadget-tracer-manager-config

What do you think about it? I'm not sure we can keep adding more flags to the deploy command as done in #3158, but it's also true that helm is always an alternative...

mqasimsarfraz:
> I have some comments, particularly with regard to `podman`. 

Done

> Also, we should test this on `citest/` to ensure the new way of deploying is OK.

I pushed it here: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/9906559650 I see some tests are failing but can't see how it is related but I will investigate! 


> > I have more or less similar opinions except I haven't introduced the config file --gadget-tracer-manager-config

> What do you think about it? I'm not sure we can keep adding more flags to the deploy command as done in #3158, but it's also true that helm is always an alternative...

I think it is a valid concern. We should support a configuration file to allow storing deployment flags but probably makes sense to handle it in follow up PR?

mqasimsarfraz:
> Particularly, I am wondering if we would not also need a ~/.kubectl-gadget-deploy/config.yaml like what we have for ig.
This way, giving the config file on the CLI would not be needed.

This sounds better!

mauriciovasquezbernal (approved):
LGTM. I did a quick test and it worked perfectly. Thanks!

mqasimsarfraz:
Thanks for the reviews!

PR #3152, opened 2024-07-10
ci: Report summaries for gadgets tests

mauriciovasquezbernal:


mauriciovasquezbernal:
> I am not convinced by the last commit, but let's go with it for now.

Neither do I but I don't see other options. In this specific test we want to capture events generated when the container starts, and we don't have any mechanism to wait until the tracer it ready.

Issue #3032, opened 2024-06-18
datasource: Support array fields


mauriciovasquezbernal:
There are gadgets that have array field in the event structure:

- https://github.com/inspektor-gadget/inspektor-gadget/blob/a46a2f0121ea68d34b96c0b139de87d50fea5174/pkg/gadgets/trace/open/types/types.go#L35
- https://github.com/inspektor-gadget/inspektor-gadget/blob/dab012336fad94c38e420956ca7b6bcad1b406ff/pkg/gadgets/trace/exec/types/types.go#L33
- https://github.com/inspektor-gadget/inspektor-gadget/blob/1943a5088b5c4f761caac434b123b44ac4add1a8/pkg/gadgets/trace/mount/types/types.go#L39

The datasource currently doesn't support this kind of fields, so as a workaround we're joining them in a single string, as done in #3025. The datasource should support this kind of fields to avoid such workarounds.

Note: #2709 implement support for an array of events, but this case is different, it's an array inside an event.

matthyx:
@mauriciovasquezbernal is it something that could work https://github.com/inspektor-gadget/inspektor-gadget/pull/5165 ?

PR #3157, opened 2024-07-11
Adding cwd field to trace_exec image-based gadget

Sefi4:
Adding cwd field to trace_exec image-based gadget

## Testing done
* Unit Tests
* Local Testing

### DIFF
--- a/gadgets/trace_exec/gadget.yaml
+++ b/gadgets/trace_exec/gadget.yaml
@@ -20,6 +20,12 @@ datasources:
         annotations:
           description: Process name
           template: comm
+      cwd:
+        annotations:
+          description: The current working directory of the process (require --paths flag)
+          columns.width: 64
+          columns.hidden: "true"
+          columns.alignment: left
       pcomm:
         annotations:
           template: pcomm
@@ -84,6 +90,10 @@ ebpfParams:
     key: ignore-failed
     defaultValue: "true"
     description: Ignore failed events
+  paths:
+    key: paths
+    defaultValue: "false"
+    description: Show the cwd of the process.
   targ_uid:
     key: uid
     defaultValue: ""
--- a/gadgets/trace_exec/program.bpf.c
+++ b/gadgets/trace_exec/program.bpf.c
@@ -9,6 +9,7 @@
 #include <gadget/macros.h>
 #include <gadget/mntns_filter.h>
 #include <gadget/types.h>
+#include <gadget/filesystem.h>
 
 // Defined in include/uapi/linux/magic.h
 #define OVERLAYFS_SUPER_MAGIC 0x794c7630
@@ -42,15 +43,18 @@ struct event {
 	bool upper_layer;
 	bool pupper_layer;
 	unsigned int args_size;
+	char cwd[MAX_STRING_SIZE];
 	char args[FULL_MAX_ARGS_ARR];
 };
 
 const volatile bool ignore_failed = true;
 const volatile uid_t targ_uid = INVALID_UID;
 const volatile int max_args = DEFAULT_MAXARGS;
+const volatile bool paths = false;
 
 GADGET_PARAM(ignore_failed);
 GADGET_PARAM(targ_uid);
+GADGET_PARAM(paths);
 
 static const struct event empty_event = {};
 
@@ -132,6 +136,12 @@ int ig_execve_e(struct syscall_trace_enter *ctx)
 	event->args_count = 0;
 	event->args_size = 0;
 	event->mntns_id = mntns_id;
+
+	if (paths) {
+		struct fs_struct *fs = BPF_CORE_READ(task, fs);
+		char *cwd = get_path_str(&fs->pwd);
+		bpf_probe_read_kernel_str(event->cwd, MAX_STRING_SIZE, cwd);
+	}
 
 	ret = bpf_probe_read_user_str(event->args, ARGSIZE,
 				      (const char *)ctx->args[0]);
--- a/gadgets/trace_exec/test/trace_exec_test.go
+++ b/gadgets/trace_exec/test/trace_exec_test.go
@@ -50,6 +50,7 @@ type traceExecEvent struct {
 	Error       string `json:"error"`
 	UpperLayer  bool   `json:"upper_layer"`
 	PupperLayer bool   `json:"pupper_layer"`
+	Cwd         string `json:"cwd"`
 	Args        string `json:"args"`
 }
 
@@ -74,7 +75,7 @@ func TestTraceExec(t *testing.T) {
 	}
 
 	sleepArgs := []string{"/bin/sleep", "1"}
-	innerCmd := fmt.Sprintf("while true ; do %s; done", strings.Join(sleepArgs, " "))
+	innerCmd := fmt.Sprintf("cd tmp ; while true ; do %s; done", strings.Join(sleepArgs, " "))
 	// copies /bin/sh to /usr/bin/sh to check that the upper_layer is true when executing /usr/bin/sh
 	cmd := fmt.Sprintf("cp /bin/sh /usr/bin/sh ; setuidgid 1000:1111 /usr/bin/sh -c '%s'", innerCmd)
 	shArgs := []string{"/bin/sh", "-c", cmd}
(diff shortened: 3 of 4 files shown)


gadgets/trace_exec/program.bpf.c
```diff
@@ -38,9 +39,13 @@ struct event {
 	bool pupper_layer;
 	unsigned int args_size;
 	char comm[TASK_COMM_LEN];
+	char cwd[MAX_STRING_SIZE];
```
Sefi4, line 42: @mauriciovasquezbernal, @blanquicet, any suggestions on how to avoid redundant declaration of `cwd` for when --paths=false?

gadgets/trace_exec/test/trace_exec_test.go
```diff
@@ -98,13 +99,16 @@ func TestTraceExec(t *testing.T) {
 		commonDataOpts = append(commonDataOpts, utils.WithK8sNamespace(ns))
 	}
 
+	runnerOpts = append(runnerOpts, igrunner.WithFlags("--paths", "--verify-image=false"))
```
Sefi4, line 102: The --verify-image=false is required for me, since otherwise tests fail with:
Error: fetching gadget information: initializing and preparing operators: instantiating operator "oci": ensuring image: verifying image "trace_exec:sefitufan-addingPaths": getting signing information: getting signature: getting signature bytes: ghcr.io/inspektor-gadget/gadget/trace_exec:sha256-9b2b5ff0a0387979d8ee3ebc18f69eaf5a2c60d2ecbdcd5a96e8c03096ced541.sig: not found

Should I leave this filter?

gadgets/trace_exec/gadget.yaml
```diff
@@ -93,3 +97,7 @@ ebpfParams:
     key: uid
     defaultValue: ""
     description: Show only events generated by processes with this uid
+  paths:
+    key: paths
+    defaultValue: "false"
+    description: Add the path of the executable
```
burak-ok, line 103: For the CI to pass the params need to be in lexicographical order - So you need to put the `paths` before `targ_uid`

gadgets/trace_exec/test/trace_exec_test.go
burak-ok: For your own local tests you need that flag, since the gadget is not signed by the IG private key. In the PR CI we sign the image and also want to verify it.

So PR: no flag - local with flag

gadgets/trace_exec/test/trace_exec_test.go
```diff
@@ -169,7 +175,7 @@ func TestTraceExec(t *testing.T) {
 	))
 
 	runnerOpts = append(runnerOpts, igrunner.WithStartAndStop())
-	traceExecCmd := igrunner.New("trace_exec", runnerOpts...)
+	traceExecCmd := igrunner.New("trace_exec:sefitufan-addingPaths", runnerOpts...)
```
burak-ok, line 178:
```suggestion
	traceExecCmd := igrunner.New("trace_exec", runnerOpts...)
```

blanquicet (commented):
Thanks @Sefi4 for working on this!

gadgets/trace_exec/program.bpf.c
alban: In my opinion, the semantics of `macro: WITH_PATHS` are not good: your yaml snippet is presumably gadget.yaml, which is run by `ig run` to load the gadget correctly. But C macros are used at compiled time and I don't think that compilation details should leak in gadget.yaml. The gadget might be built from other languages than C and it should not matter for `ig run`. Maybe it would work better in build.yaml?

But I am still worried about the combinatorial explosion if several fields use such macros.

I think we should find a way to pass the type in the ring buffer at runtime. Either with our own event header with a field type identifying the btf type id, or with some already existing field from perf ring buffers.

gadgets/trace_exec/gadget.yaml
```diff
@@ -21,6 +21,12 @@ datasources:
       comm:
         annotations:
           template: comm
+      cwd:
+        annotations:
+          description: The current working directory of the executable
+          columns.width: 64
+          columns.hidden: "true"
```
Sefi4, line 28: Instead of having a literal like "true"/"false" for `columns.hidden` it could be nice to have it based on ebpf param paths. Something like `columns.hidden: "param:paths"`.

I think it's not supported right now [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/datasource/columns.go#L81) but could be nice to have, in addition to the discussion about optional fields

gadgets/trace_exec/test/trace_exec_test.go
mauriciovasquezbernal: It's also possible to use `IG_FLAGS="--verify-image=false"` when running the tests locally.

gadgets/trace_exec/test/trace_exec_test.go
```diff
@@ -98,75 +99,80 @@ func TestTraceExec(t *testing.T) {
...
-				utils.NormalizeInt(&e.Loginuid)
-				utils.NormalizeInt(&e.Sessionid)
-			}
-
-			match.MatchEntries(t, match.JSONMultiObjectMode, output, normalize, expectedEntries...)
-		},
-	))
+	runnerOpts = append(runnerOpts,
+		igrunner.WithFlags("--paths"),
+		igrunner.WithValidateOutput(
```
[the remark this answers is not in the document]
blanquicet: I think that should be managed in a unit-test instead.

gadgets/trace_exec/gadget.yaml
```diff
@@ -21,6 +21,12 @@ datasources:
       comm:
         annotations:
           template: comm
+      cwd:
+        annotations:
+          description: The current working directory of the executable
```
alban, line 26:
```suggestion
          description: The current working directory of the process
```

gadgets/trace_exec/gadget.yaml
```diff
@@ -79,6 +85,10 @@ ebpfParams:
     key: ignore-failed
     defaultValue: "true"
     description: Ignore failed events
+  paths:
+    key: paths
+    defaultValue: "false"
+    description: Show the cwd of the executable.
```
alban, line 91:
```suggestion
    description: Show the cwd of the process.
```

gadgets/trace_exec/program.bpf.c
blanquicet: > In my opinion, the semantics of `macro: WITH_PATHS` are not good: your yaml snippet is presumably gadget.yaml, which is run by `ig run` to load the gadget correctly. But C macros are used at compiled time and I don't think that compilation details should leak in gadget.yaml. The gadget might be built from other languages than C and it should not matter for `ig run`. Maybe it would work better in build.yaml?
> 
You're right. The build.yaml is probably more appropriated for specifying the defines. Using that, we could create two versions of the gadget, but then we are missing the association between each version and a given flag. How can you do that? Any idea?

> But I am still worried about the combinatorial explosion if several fields use such macros.
> 
IMO, that's an issue the gadget developer must take care of. We provide the tools, they are responsible to use them properly.

> I think we should find a way to pass the type in the ring buffer at runtime. Either with our own event header with a field type identifying the btf type id, or with some already existing field from perf ring buffers.
>
Could you please elaborate this idea?

alban: I don't have more precise idea, and I don't know if it could work. I just see in the manpage of perf_event_open that it has a lot of options. I wonder if metadata could be useful.

alban (approved):
LGTM

Let's wait for the CI results.

gadgets/trace_exec/test/trace_exec_test.go
```diff
@@ -98,75 +101,97 @@ func TestTraceExec(t *testing.T) {
...
+						Ppid:      utils.NormalizedInt,
+						Loginuid:  utils.NormalizedInt,
+						Sessionid: utils.NormalizedInt,
+					},
+					// test cwd
+					{
+						CommonData:  utils.BuildCommonData(cwdTestContainerName, commonDataOpts...),
+						Comm:        "sleep",
+						Cwd:         "/tmp",
+						Args:        "/bin/sleep 1",
```
blanquicet, line 167:
```suggestion
						Args:        strings.Join(sleepArgs, " "),
```

BTW, we use `sleepArgs` always with `strings.Join`. Probably, we should define it already as a string `"/bin/sleep 1"`.

blanquicet (approved):
LGTM from code inspection.

gadgets/trace_exec/test/trace_exec_test.go
mauriciovasquezbernal: The thing here is that `Args` should be a string array instead of a string (see #3032), we took this test for the builtin one, so using strings.Joins was a quick way to adapt the test.

gadgets/trace_exec/program.bpf.c
mauriciovasquezbernal: I made some experiments to check if we can support optional fields in a more efficient way (without wasting a lot of memory and sending a lot of bytes that aren't used on the ring buffer). I haven't reached any conclusion yet, but want to mention it so you're aware.

TBH I don't think the previous "macro-based" approach will work with image based gadgets, some limitations I see are:
- There is a macro for each optional parameter, the combination explodes (already mentioned by Alban before)
- We need to compile and store two versions of the program in the OCI image (we don't have anything similar yet)
  - We need to understand which version to load based on the parameter
- Testing is more complicated (in the end, those are two different ebpf objects we need to test)

### Idea 

I explored using a Type-length-value encoding approach. Thea idea is to make the gadget send an event that has a variable length, the first bytes will be the fields that are always present (pid, tid, uid, etc), and after it, to have optional fields encoded using TLV

```c
struct event {
 u32 pid;
 u32 tid;
 char optional_fields[0];
}
```

The optional fields will encoded as

```
+---------------+-----------+--------------+---------------+-----------+--------------+
| field0 length | field0 id | field0 value | field1 length | field1 id | field1 value | ...
+---------------+-----------+--------------+---------------+-----------+--------------+
```

The gadget would mark optional fields by using something like:

```c
// execs: name of the tracer 
// cwd: name of the field 
// 1: id of the field 

// TODO: define type of the field (string vs bytes[])

GADGET_TRACER_OPTIONAL_FIELD(execs, cwd, 1)
```

The ebpf operator will then use these macros to add and set the fields when reading the value from the ring buffer. 

Something I haven't explored is how to enable / disable those fields. We already have a [IsRequested()](https://github.com/inspektor-gadget/inspektor-gadget/blob/84f57b00c0e1f5c00e1ed75432b0f468e85da3ed/pkg/datasource/datasource.go#L169) method on the field interface, I think we can wire it to set the value of a constant in the eBPF code when a given field is enabled / disabled.

As mentioned before, it's still some quick experimentation I did, but I think I can make it work. 

Btw, I don't expect to block this PR, please keep it rolling without waiting for this support as it'll take some more time.

alban: I think TLV encoding is good. That would also help reducing the size for strings that typically are shorter than MAX_STRING_SIZE (4096). The `cwd` field is typically 2 bytes ("/" with a NULL).

GADGET_TRACER_OPTIONAL_FIELD should include the BTF type too, so that types like `gadget_l3endpoint_t` can be enriched later on.

gadgets/trace_exec/test/trace_exec_test.go
blanquicet: This is back 😅

Sefi4: Oops, I have to get used to it 😅

gadgets/trace_exec/program.bpf.c
mauriciovasquezbernal: > GADGET_TRACER_OPTIONAL_FIELD should include the BTF type too, so that types like gadget_l3endpoint_t can be enriched later on.

I think we can initially limit it to be only strings and byte arrays. (it seems to me they are the most common use cases). I fear including BTF information here will be complicated.

gadgets/trace_exec/program.bpf.c
```diff
@@ -43,15 +44,20 @@ struct event {
 	bool upper_layer;
 	bool pupper_layer;
 	unsigned int args_size;
+	char comm[TASK_COMM_LEN];
```
alban, line 47: Duplicate field `comm`: lines 31 and 47.

gadgets/trace_exec/test/trace_exec_test.go
```diff
@@ -79,8 +81,10 @@ func TestTraceExec(t *testing.T) {
 	cmd := fmt.Sprintf("cp /bin/sh /usr/bin/sh ; setuidgid 1000:1111 /usr/bin/sh -c '%s'", innerCmd)
 	shArgs := []string{"/bin/sh", "-c", cmd}
 	innerShArgs := []string{"/usr/bin/sh", "-c", innerCmd}
+	cwdTestCmd := fmt.Sprintf("cd /tmp ; %s", strings.Join(sleepArgs, " "))
 
 	testContainer := containerFactory.NewContainer(containerName, cmd, containerOpts...)
+	testCwdContainer := containerFactory.NewContainer(cwdTestContainerName, cwdTestCmd, containerOpts...)
```
alban, line 87: The tests fail in kubernetes because it creates 2 different pods and for each of them it creates a Kubernetes namespace with the same name. See `containers.WithContainerNamespace` a few lines above.

So currently the sequence is as follows:
- namespace/test-trace-exec-402149015484310388 created
- pod/test-trace-exec condition met
- namespace/test-trace-exec-402149015484310388 unchanged
- pod/test-trace-exec-cwd condition met
- pod "test-trace-exec-cwd" deleted
- namespace "test-trace-exec-402149015484310388" deleted
- Error from server (NotFound): pods "test-trace-exec" not found

When deleting the Kubernetes namespace for the first time, it deletes the pod that remained inside. Then, at the next step, it fails to delete the pod.

This can be fixed by using two different Kubernetes namespaces for each pod.

alban:
Could you please rebase? I would like to check in the CI that it interacts well with the changes from #3138.

gadgets/trace_exec/test/trace_exec_test.go
Sefi4: Wasn't sure how it would work with another namespace for kubectl-gadget. So changed the already exist container to test cwd too

pkg/testing/ig/ig.go
```diff
@@ -57,7 +57,7 @@ func WithPath(path string) Option {
 // WithFlags args should be in form: "--flag_name=value" or "-shorthand=value".
 func WithFlags(flags ...string) Option {
 	return func(ig *runner) {
-		ig.flags = flags
+		ig.flags = append(ig.flags, flags...)
```
Sefi4, line 60: I believe it was a bug or at least confusing for me. In the test I called withFlags twice which the second call override the previous exists flags.

PR #3159, opened 2024-07-11
Enabling mntns filters on DNS trace gadget

slashben:
# Enabling the mount namespace filter on the trace/dns gadget

The DNS gadget needs to be filterable by mount namespace (container), otherwise some special containers like CoreDNS will overload the reader side.

This code enables the standard filtering method common with other tracers.


alban:
Hi,
Have you found more information why the CPU is overloaded on nodes with coredns?

[Slack thead](https://kubernetes.slack.com/archives/CSYL75LF6/p1720705023758379)

PR #3161, opened 2024-07-12
tests: Remove run top test and helpers for this

mauriciovasquezbernal:
We can port this test to the new testing framework once support for toppers is back.


blanquicet (approved):
LGTM!

Issue #2997, opened 2024-06-11
[EPIC] Mark image-based gadgets as non-experimental


mauriciovasquezbernal:
A lot of work has been done in the image-based gadgets feature #1929. We can now start planning to mark this support as non-experimental. This issue contains the list of tasks we need to complete to unmark it.

## Features 
- [x] #1598
- [x] #2482
- [x] #2202 
- [x] #2493
- [ ] #2495

## Port built-in gadgets 

We aim to port all gadgets we few exceptions:

- traceloop (see #2822)
- top bpf: to be done later on once we have a more mature wasm support
- advise: There are some things to discuss first (#3003) so it's out of scope for now

Gadgets to port:

- audit
	- [x] #2988 
- profile (#1925)
	- [x] #3000
	- [ ] #3001
	- [ ] #3002
- top (#1924)
	- [ ] #2989 
	- [x] #2990
- trace
	- [x] #2991
	- [x] #2992
	- [x] #2999
	- [ ] #2993
	- [ ] #2994
	- [x] #2995	
	- [ ] #2996
	- [x] #3024

### Documentation 

We need to have a image-based gadgets oriented documentation, the following issue tracks this, however for removing the flag we don't need to have it totally done.

- [ ] #1184

PR #3163, opened 2024-07-12
Prepopulate gadgets and restrain adding new ones

eiffel-fl:
Hi.

In this PR, I added a mechanism to prepopulate gadgets and avoid adding new ones:

```bash
$ docker build -t ghcr.io/inspektor-gadget/prepopulator tools/prepopulator/.
...
$ ./bin/minikube/minikube-v1.33.1 image load ghcr.io/inspektor-gadget/prepopulator
$ make minikube-start
...
$ ./bin/minikube/minikube-v1.33.1 image load ghcr.io/inspektor-gadget/inspektor-gadget:francis-volume-ig-oci
$ ./kubectl-gadget deploy --image-pull-policy=Never --prepopulate-gadgets=trace_exec,trace_open,ghcr.io/eiffel-fl/gadget/trace_exec
...
$ ./kubectl-gadget run trace_exec -A                          francis/volume-ig-oci *%
K8S.NAMESPACE  K8S.PODNAME    K8S.CONTAINER… MNTNS… PID      PPID     UID      GID      R… U… P… COMM    ARGS    K8S.NO… TIMESTAMP       
gadget         gadget-6hhzt   gadget         402653 59193    59175    0        0        0  f… f… gadgett /bin/g… minikub 2024-07-12T12:59
gadget         gadget-6hhzt   gadget         402653 59194    59176    0        0        0  f… f… gadgett /bin/g… minikub 2024-07-12T12:59
^C%
$ ./kubectl-gadget run --verify-image=false ghcr.io/eiffel-fl/gadget/trace_exec -A
K8S.NAMESPACE    K8S.PODNAME      K8S.CONTAINERNA… MNTNS_… TIMESTAMP           PID      PPID     UID      GID      … U… COMM     K8S.NODE
gadget           gadget-6hhzt     gadget           4026537 7015298274488       59795    59776    0        0        0 f… gadgettr minikube
gadget           gadget-6hhzt     gadget           4026537 7015298274604       59794    59775    0        0        0 f… gadgettr minikube
^C%
$ ./kubectl-gadget run trace_oomkill -A                       francis/volume-ig-oci *%
Error: fetching gadget information: getting gadget info: rpc error: code = Unknown desc = getting gadget info: initializing and preparing operators: instantiating operator "oci": ensuring image: pulling image (if missing) "trace_oomkill": downloading to local repository: failed to create ingest file: open /var/lib/ig/oci-store/ingest/4032486abf82cf8e2e5d4e2ec332a693f10eb690fb3b5d51b5583f457340bb5d_359242559: read-only file system
```

Best regards.

### DIFF
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -105,6 +105,7 @@ var (
 	verifyImage         bool
 	publicKey           string
 	strLevels           []string
+	prepopulateGadgets  []string
 )
 
 var supportedHooks = []string{"auto", "crio", "podinformer", "nri", "fanotify", "fanotify+ebpf"}
@@ -231,6 +232,10 @@ func init() {
 	deployCmd.PersistentFlags().StringVarP(
 		&publicKey,
 		"public-key", "", resources.InspektorGadgetPublicKey, "Public key used to verify the container image")
+	rootCmd.AddCommand(deployCmd)
+	deployCmd.PersistentFlags().StringSliceVar(
+		&prepopulateGadgets,
+		"prepopulate-gadgets", []string{}, "List of gadgets to prepopulate")
 	rootCmd.AddCommand(deployCmd)
 }
 
@@ -630,6 +635,32 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 						return fmt.Errorf("invalid log level %q, valid levels are: %v", daemonLogLevel, strings.Join(strLevels, ", "))
 					}
 					gadgetContainer.Env[i].Value = daemonLogLevel
+				}
+			}
+
+			if len(prepopulateGadgets) > 0 {
+				daemonSet.Spec.Template.Spec.InitContainers = append(daemonSet.Spec.Template.Spec.InitContainers,
+					v1.Container{
+						Name:            "prepopulator",
+						Image:           "ghcr.io/inspektor-gadget/prepopulator",
+						Command:         []string{"/prepopulator"},
+						Args:            prepopulateGadgets,
+						ImagePullPolicy: policy,
+						VolumeMounts: []v1.VolumeMount{
+							{
+								MountPath: "/var/lib/ig",
+								Name:      "oci",
+							},
+						},
+					},
+				)
+
+				for i, volumeMount := range gadgetContainer.VolumeMounts {
+					if volumeMount.Name != "oci" {
+						continue
+					}
+
+					gadgetContainer.VolumeMounts[i].ReadOnly = true
 				}
 			}
 


mauriciovasquezbernal (commented):
I'm not sure about the prepolator tool, it's another tool we'll have to maintain. Why can't we make it directly on the gadget tracer manager?

We can define something like "prepulled" images in the configuration file:

```yaml
operator:
    oci:
        prepulled-images:
        - ghcr.io/foo/bar:latest
        - ghcr.io/foo/baz:latest
        disable-pulling: true 
```

and another flag to disable pulling images.

Then it'll be handled by the operator and available to both, ig and the gadgettracermanager.

Another option is to use "image export" and "image import", but it's not clear to me how the user could distribute the generated tarball to the different nodes.

flyth:
This is an interesting approach and I like quite a couple of advantages it brings (read-only), however, I wanted to add another idea that we have (briefly) discussed at the get together and that is connected to this: the idea of "installed gadgets".

"Installed gadgets" could be pretty much the same as "pre-populated gadgets" - they'd however be pre-populated by the gadget pods themselves by listening to a new custom resource called "installed-gadget" or something like that. So whenever a new resource like that gets created, all pods would immediately pull the specified image.

Advantages:
* RBAC control over what is pre-populated (can restrict who is able to deploy the CR)
* pre-population can be done without re-deploying IG by an admin at any time (just create a new CR)
* IG process is aware of state (all installed CRs) and can re-try downloading images (what if the Internet is down when prepopulating for example on some nodes)
* additional restrictions (like parameter restrictions) can be placed in the CR
* a "catalog" of available gadgets can quickly be compiled (we need something like this for other interfaces)

Disadvantages:
* no read-only volume (checks are done in the IG process using the CR, though - so even if an image would be placed by a third party, it could not be run)
* complexer logic when using default configuration of IG (e.g. without restrictions; in that case, installing the gadget should happen implicitly - probably using RBAC credentials of the client)

WDYT?

mauriciovasquezbernal:
I'd propose to hold it until we are finished with #2997.

mauriciovasquezbernal:
I'm thinking again about this and I'm not sure we need this support, at least not as a security mechanism. We're now able to control the gadgets that can be run by using [`allowed-gadgets`](https://www.inspektor-gadget.io/docs/latest/reference/restricting-gadgets). I see the mechanism provided in this PR more like a way to prepopulate the gadgets (to avoid having to reach the network), but that seems orthogonal to the security mechanism. In order to prepopulate the gadgets, I think something based on `image import` and `image export` should work. (this is however not clear how it'll work on the k8s case and if it should be a responsibility of Inspektor Gadget to handle it).

mauriciovasquezbernal:
I think if somebody is able to bypass the allowed-gadgets then we're in a really bad position. However I agree that a read-only volume provides extra security, but we need to think more how to implement it. I think this "prepolator" is too hacky and only covers the k8s case. I'm imagining a way provide a file with a set of "prepopulated" gadgets it to ig with a `--static-gadgets` or similar flag. 

I also want to get other folks opinions on this one. cc @alban

alban:
I don't really want to link those two features together:
1. Prepopulate gadgets
2. Restrict downloads

I think those two features should be configurable independently.

I don't think restricting downloads solely with a read-only volume is a good idea. If we make the path `/var/lib/ig/oci-store` configurable in the configuration file, then the read-only volume setting in the Helm chart will stop having an effect.

I also don't like that the download restriction is configured differently in the ig and the kubectl-gadget cases. And even in the kubectl-gadget case, there are 2 different code paths: configuration with Helm or with kubectl-gadget-deploy. This makes it easy to have regression bugs: for example, the deploy command checks for `volumeMount.Name != "oci"` to set the read-only flag, but if we rename the volume to oci-store at some point, it will silently stop being read-only.

I prefer the `disable-pulling: true` solution too because it would offer a single implementation in a single file for all use cases:
1. ig
2. kubectl-gadget with Helm
3. kubectl-gadget with the deploy command
4. custom software importing IG Golang packages

Next week, I plan to get user feedback for the IG Golang packages use case. I would like the two features (prepopulate & download restriction) to be available for that use case.

mqasimsarfraz:
Came across this in morning: https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/ and was thinking we can have an image with all the gadgets and mount it as mentioned in the blog. The feature is in Alpha so perhaps something for future.

mauriciovasquezbernal:
> (not sure 100% relevant)

It's very relevant! I checked the k8s documentation looking for options but there was nothing, I think this new volume is perfect for our use case, thanks a lot for sharing it!

PR #3164, opened 2024-07-12
Update metadata validation and generation

### LINKED ISSUE #2560: [Discussion] image-based gadgets: How to mark the maps used by a gadget to generate information?

mauriciovasquezbernal:
This PR cleans up the logic to update and validate the metadata file.

It's mostly a cleanup after #2512 based  on the discussion of #2560 and the changes on #3149 

Fixes #2560

--- 

We're thinking different ways to generate the metadata for a gadget, perhaps we can use GetGadgetInfo(), for this reason this PR only updates the logic to follow the changes on #3149 but doesn't introduce tests as it could be changed very soon again.


mauriciovasquezbernal:
Before going on the full motivation of the new approach, let me provide a bit of history on the different steps we've taken. The metadata file was initially conceived to fully describe a gadget, it contained which tracers, snapshotters, toppers, structures and parameters were exposed by it. It was the source of truth and the implementation fully relied on the information contained on this file to work. Then #2512 changed it by relying more on the information contained in the BPF object, this leaded to discuss the role of the metadata file in #2560, where we agreed to make the `GADGET_` macros mandatory. Then #3149 happened, making many parts (tracers, structures) of the metadata file useless. 

Now, the motivation for having this new approach (where we rely less on the metadata file is). This is mostly taken from #2560 plus some additional thoughts I have:
- There were two sources of information: (1) the ebpf object and (2) the metadata file. There was some complex logic to validate that both of them were aligned, also the fallback mechanisms when the metadata file was not present were confusing.
- Having two ways to define the gadget (`GADGET_` macros vs metadata) created confusion. 
- Forcing the user to create a metadata file adds more complexity to the gadget creation process. This file should be optional and only provide the "sugar" to make the gadget look better. For instance, if the user only wants to get json output, all column formatting options aren't needed.

pkg/gadgets/run/types/metadata.go
```diff
@@ -24,9 +24,6 @@ import (
 	"github.com/hashicorp/go-multierror"
 	log "github.com/sirupsen/logrus"
 
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/btfhelpers"
```
blanquicet, line 27: After these changes and a couple of changes in the metadata files, I think can revert https://github.com/inspektor-gadget/inspektor-gadget/commit/eb6de7b7a77c7353d9c24fca1cfc05d255deee23. Do you want to do it in another PR?

blanquicet (approved):
LGTM! Thanks for this cleanup!

docs/devel/hello-world-gadget.md
```diff
@@ -311,31 +311,18 @@ description: 'TODO: Fill the gadget description'
...
-        width: 16
-        alignment: left
-        ellipsis: end
-    - name: filename
-      description: 'TODO: Fill field description'
-      attributes:
-        width: 16
-        alignment: left
-        ellipsis: end
+      comm:
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Right, it's coming from the metadata package. I also don't know if we should populate it with all defaults.

pkg/gadgets/run/types/metadata.go
mauriciovasquezbernal: I'd prefer to wait a bit until we understand how to implement it again.

docs/devel/hello-world-gadget.md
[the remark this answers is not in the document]
blanquicet: We either do that or provide a scheme where people can see the default values we are using. Otherwise, users will see columns displayed with a certain format and they'll not understand where such a configuration is coming from.

pkg/gadgets/run/types/metadata.go
blanquicet: I saw you created the function to populate the datasource yaml leaf. I tested it with `--update-metadata` and it worked. I ensures now the alphabetical order so some yaml files need to be updated, but other than that, I don't see what is missing. What am I not considering?

docs/devel/hello-world-gadget.md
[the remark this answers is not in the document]
mauriciovasquezbernal: We can document the default annotations we use. The idea is to keep the metadata files clean without duplicating all default values.

pkg/gadgets/run/types/metadata.go
mauriciovasquezbernal: > I ensures now the alphabetical order so some yaml files need to be updated

This is what I was afraid of. Are you sure it's always the alphabetical order? If it's the case, we can re-enable that.

mauriciovasquezbernal: I'm merging it right now to unblock other changes. I'll think about this later on.

PR #3170, opened 2024-07-15
gadgets: Add audit seccomp

### LINKED ISSUE #2988: gadgets: Implement audit seccomp as image-based gadget
The [audit seccomp](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/audit/seccomp) uses the same mechanism to print events as the tracers, we have all functionality we need to port it now.

mauriciovasquezbernal:
Fixes #2988


gadgets/audit_seccomp/program.bpf.c
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2021 The Inspektor Gadget authors */
```
burak-ok, line 2: Should the year be updated?

burak-ok (approved):
LGTM, thanks.
I still think that it should be something like `trace_audit_seccomp` or just `trace_seccomp`

PR #3172, opened 2024-07-16
gadgets: Test on containerd too

mauriciovasquezbernal:
Enable running the gadgets tests on containerd.

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1393,8 +1393,8 @@ jobs:
         run: |
           GADGET_REPOSITORY=${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }} make -C gadgets/ sign -o push
 
-  test-gadgets:
-    name: Test gadgets
+  test-gadgets-local:
+    name: Test gadgets locally
     #level 3
     needs:
       - build-ig
@@ -1402,6 +1402,10 @@ jobs:
       - build-and-push-gadgets
       - check-secrets
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        runtime: [docker, containerd]
     steps:
     - uses: actions/checkout@v4
     - name: Set container repository and determine image tag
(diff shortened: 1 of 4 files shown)


gadgets/Makefile
```diff
@@ -99,12 +100,9 @@ test-ig-local: build
...
-test-k8s: build
-	IG_PATH=$(KUBECTL_GADGET) \
-	IG_RUNTIME=kubernetes \
-	CGO_ENABLED=0 \
-	IG_EXPERIMENTAL=true \
-	GADGET_REPOSITORY=$(GADGET_REPOSITORY) \
-	GADGET_TAG=$(GADGET_TAG) \
-	IG_FLAGS=$(IG_FLAGS) \
-	go test -exec 'sudo -E' -v ./...
+test-local: IG_PATH=$(IG)
```
burak-ok, line 103: Not your changes, but a question: Why do we have `IG_PATH` and `IG` env variables in this Makefile? Can't we just have one?

gadgets/Makefile
```diff
@@ -99,12 +100,9 @@ test-ig-local: build
...
+test-local: test
```
burak-ok, line 104: Probably because I never saw this - This confuses me having the same target under each other

.github/workflows/inspektor-gadget.yml
```diff
@@ -1393,15 +1393,19 @@ jobs:
         run: |
           GADGET_REPOSITORY=${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }} make -C gadgets/ sign -o push
 
-  test-gadgets:
-    name: Test gadgets
+  test-gadgets-local:
+    name: Test gadgets local
```
burak-ok, line 1397:
```suggestion
    name: Test gadgets on ig
```
or
```suggestion
    name: Test gadgets locally
```

burak-ok (approved):
LGTM, thanks

gadgets/Makefile
mauriciovasquezbernal: I didn't fin another way to do it, it's covered in https://stackoverflow.com/a/15230658 and https://www.gnu.org/software/make/manual/make.html#Target_002dspecific

gadgets/Makefile
mauriciovasquezbernal: `IG_PATH` is the path to the binary executed for the test. `IG` is the default location of the ig binary, and `KUBECTL-GADGET` is the default location of the kubectl-gadget binary. Probably we can remove the IG and KUBECTL-GADGET ones.

pkg/testing/utils/utils.go
```diff
@@ -93,9 +93,10 @@ func BuildCommonData(containerName string, options ...CommonDataOption) eventtyp
 func NormalizeCommonData(e *eventtypes.CommonData) {
 	// The container image digest is not currently enriched for Docker containers:
 	// https://github.com/inspektor-gadget/inspektor-gadget/issues/2365
-	if e.Runtime.RuntimeName == eventtypes.RuntimeNameDocker {
-		e.Runtime.ContainerImageDigest = ""
-	}
+	// It's enriched for other runtimes like containerd and cri-o, but our
+	// testing framework doesn't provide any way to get this information, so
+	// ignore this field for now.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Ref #3176

PR #3174, opened 2024-07-17
Add support to specify several public keys

eiffel-fl:
Hi!

This PR adds support to specify several public keys:

```bash
$ ./kubectl-gadget deploy --image-pull-policy=Never --gadgets-public-keys="$(cat eiffel-fl.pub)","$(cat eiffel-fl.pub)"
...
$ ./kubectl-gadget run ghcr.io/eiffel-fl/gadget/trace_exec   francis/public-keys *% u=
K8S.NAMESP… K8S.PODNAME K8S.CONTAI… MNTNS… TIMES… PID    PPID   UID    GID    LOGIN… SESSI… RETVAL ARGS_… UPPER… ARGS_… COMM  ARGS  K8S.…
^C%
# Inspektor Gadget key is always added.
$ ./kubectl-gadget run trace_exec                            francis/public-keys *% u=
K8S.NAMESPACE       K8S.PODNAME         K8S.CONTAINERNAME          PID       PPID RE… COMM      ARGS      K8S.NODE  TIMESTAMP            
^C%
```

Best regards.

### DIFF
--- a/cmd/kubectl-gadget/deploy.go
+++ b/cmd/kubectl-gadget/deploy.go
@@ -107,7 +107,7 @@ var (
 	publicKey           string
 	strLevels           []string
 	verifyGadgets       bool
-	gadgetsPublicKey    string
+	gadgetsPublicKeys   string
 )
 
 var supportedHooks = []string{"auto", "crio", "podinformer", "nri", "fanotify", "fanotify+ebpf"}
@@ -239,9 +239,12 @@ func init() {
 		"verify-gadgets", "",
 		true,
 		"verify gadgets using the provided public key")
-	deployCmd.PersistentFlags().StringVarP(
-		&gadgetsPublicKey,
-		"gadgets-public-key", "", resources.InspektorGadgetPublicKey, "Public key used to verify the gadgets")
+	// WARNING For now, use StringVar() instead of StringSliceVar() as only the
+	// first line of the file will be taken when used with
+	// --gadgets-public-keys="$(cat inspektor-gadget.pub),$(cat your-key.pub)"
+	deployCmd.PersistentFlags().StringVar(
+		&gadgetsPublicKeys,
+		"gadgets-public-keys", resources.InspektorGadgetPublicKey, "Public keys used to verify the gadgets")
 	rootCmd.AddCommand(deployCmd)
 }
 
@@ -739,8 +742,7 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 				return fmt.Errorf("%s.%s not found in config.yaml", gadgettracermanagerconfig.Operator, gadgettracermanagerconfig.Oci)
 			}
 
-			opOciCfg[gadgettracermanagerconfig.VerifyImage] = verifyGadgets
-			opOciCfg[gadgettracermanagerconfig.PublicKey] = gadgetsPublicKey
+			opOciCfg[gadgettracermanagerconfig.PublicKeys] = strings.Split(gadgetsPublicKeys, ",")
 
 			data, err := yaml.Marshal(cfg)
 			if err != nil {
--- a/integration/k8s/run_insecure_test.go
+++ b/integration/k8s/run_insecure_test.go
@@ -67,8 +67,9 @@ func TestRunInsecure(t *testing.T) {
 	}
 
 	// TODO: Ideally it should not depend on a real gadget, but we don't have a "test gadget" available yet.
-	// As the image was not signed, we need to set --verify-image=false.
-	cmd := fmt.Sprintf("ig run --verify-image=false %s:5000/trace_open:%s -o json --insecure --timeout 2", registryIP, *gadgetTag)
+	// As the image was not signed, we need to set --public-keys="" to deactivate
+	// image signature verification.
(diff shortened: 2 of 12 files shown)


integration/k8s/run_insecure_test.go
```diff
@@ -60,9 +61,19 @@ func TestRunInsecure(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("setting IG_EXPERIMENTAL: %v\n", err)
+	}
+
 	// TODO: Ideally it should not depend on a real gadget, but we don't have a "test gadget" available yet.
 	// As the image was not signed, we need to set --verify-image=false.
-	cmd := fmt.Sprintf("$KUBECTL_GADGET run --verify-image=false %s:5000/trace_open:%s -n %s -o json --insecure --timeout 2", registryIP, *gadgetTag, ns)
+	cmd := fmt.Sprintf("ig run --verify-image=false %s:5000/trace_open:%s -n %s -o json --insecure --timeout 2", registryIP, *gadgetTag, ns)
+
+	err = os.Setenv("IG_EXPERIMENTAL", "false")
```
burak-ok, line 73: Setting the env variable to `true` and then `false` here makes no difference. In line 71 only the string is built and not run.

So it is needed that it will back set back (best to its previous value) after running the `cmd`

pkg/oci/oci.go
```diff
@@ -788,9 +784,26 @@ func verifyImage(ctx context.Context, image string, imgOpts *ImageOptions) error
...
+			return fmt.Errorf("creating verifier for %s: %w", publicKey, err)
+		}
+
+		err = verifier.VerifySignature(bytes.NewReader(signatureBytes), bytes.NewReader(payloadBytes))
+		if err == nil {
+			verified = true
+
+			break
+		}
+
```
burak-ok, line 801:
```suggestion
```

cmd/kubectl-gadget/deploy.go
```diff
@@ -737,7 +737,8 @@ func runDeploy(cmd *cobra.Command, args []string) error {
 			}
 
 			opOciCfg[gadgettracermanagerconfig.VerifyImage] = verifyGadgets
-			opOciCfg[gadgettracermanagerconfig.PublicKeys] = gadgetsPublicKeys
+			fmt.Printf("gadgetsPublicKeys: %v\n", gadgetsPublicKeys)
```
burak-ok, line 740: Was probably only needed during development

integration/k8s/run_insecure_test.go
burak-ok: Totally missed that its not the `main` base branch

pkg/oci/oci.go
```diff
@@ -927,8 +939,8 @@ func ensureImage(ctx context.Context, imageStore oras.Target, image string, imgO
 		}
 	}
 
-	if !imgOpts.VerifyPublicKey {
-		imgOpts.Logger.Warnf("image signature verification is disabled due to using corresponding CLI options")
+	if len(imgOpts.PublicKeys) == 0 {
+		imgOpts.Logger.Warnf("image signature verification is disabled because no public keys were given")
```
mauriciovasquezbernal, line 943:
```suggestion
		imgOpts.Logger.Warnf("image signature verification is disabled because no public keys were provided")
```

?

mauriciovasquezbernal (commented):
The code looks good, I tested it with ig and ig-k8s and gadgetctl and it worked fine. 

The only issue I hit was when trying to use a configuration file with ig. I have the following config.yaml:

```yaml
operator:
    oci:
        allowed-digests: ""
        authfile: /var/lib/ig/config.json
        insecure: "false"
        public-keys:
        - |
          -----BEGIN PUBLIC KEY-----
          MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEoDOC0gYSxZTopenGmX3ZFvQ1DSfh
          Ir4EKRt5jC+mXaJ7c7J+oREskYMn/SfZdRHNSOjLTZUMDm60zpXGhkFecg==
          -----END PUBLIC KEY-----
        pull: missing
        pull-secret: ""
        validate-metadata: "true"
verbose: "true"
```

It seems the public-key isn't taken into consideration 

```bash 
WARN[0000] image signature verification is disabled because no public keys were given
``` 

/cc @mqasimsarfraz 

There is a lot of --verify-image left overs on the code, I'll approve once they're removed.

mqasimsarfraz:
> The only issue I hit was when trying to use a configuration file with ig.

Good catch! 

> It seems the config code needs to be adapted.

Yeah, it seems we are overwriting the global params here: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/config.go#L158 so this line can be removed. Apart from it it seems we need to handle slices as well in this code.

mauriciovasquezbernal:
> Can you please point a specific one?

All the instances on docs/devel/hello-world-gadget.md and docs/devel/hello-world-gadget-wasm.md, and some instances on the "Verify image-based gadgets" section of docs/getting-started/verify.md.

flyth:
> @flyth Any idea?

Just for reference: implemented [this](https://github.com/inspektor-gadget/inspektor-gadget/commit/4d903bc99641565a09aff66292dd9304e92ba603) to address string slice support.

mauriciovasquezbernal (approved):
I tested and it works fine for me, even with multiple keys on the conf file for ig.

PR #3186, opened 2024-07-18
gadget-context: make sure cancel() is called on gadgetContext

flyth:
Without calling cancel on the gadgetContext itself, we're leaking memory as many goroutines won't exit. Although we should probably also optimize and use local synchronization in operators (and waiting for cleanup on Stop()), this makes sure that at least when exiting gadgetcontext.run(), everything gets cancelled.

### DIFF
--- a/pkg/gadget-context/run.go
+++ b/pkg/gadget-context/run.go
@@ -89,6 +89,7 @@ func (c *GadgetContext) initAndPrepareOperators(paramValues api.ParamValues) ([]
 }
 
 func (c *GadgetContext) run(dataOperatorInstances []operators.DataOperatorInstance) error {
+	defer c.cancel()
 	log := c.Logger()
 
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
LGTM. It'll be nice to get some comparisons of the memory usage before and after this change.

flyth:
> LGTM. It'll be nice to get some comparisons of the memory usage before and after this change.

Thanks for the review!

I tested with trace_open and each additional gadget run would add roughly 0.8 MB - mostly coming from the parsed ebpf program data held in cilium structs. But keeping the readers open forever would in the end also hit other limits.

After the fix everything is freed up again.

Biggest chunk when running that specific gadget is coming from the KernelSpec - that is loaded only _once_ and takes around 37 MB. But it'll never be freed, because we keep it in the cache.

PR #3200, opened 2024-07-22
UidGidResolver: Make it usable in image based gadgets

burak-ok:
## UidGidResolver: Make it usable in image based gadgets

This PR introduces the UidGidResolver to the image based gadgets world. The second commit then showcases how it can be used in `trace_exec`

### DIFF
--- a/gadgets/trace_exec/gadget.yaml
+++ b/gadgets/trace_exec/gadget.yaml
@@ -36,11 +36,13 @@ datasources:
           description: User ID
           template: uid
           columns.hidden: true
+          uidgidresolver.target: user
       gid:
         annotations:
           description: Group ID
           template: uid
           columns.hidden: true
+          uidgidresolver.target: group
       ppid:
         annotations:
           template: pid
@@ -48,6 +50,7 @@ datasources:
         annotations:
           template: uid
           columns.hidden: true
+          uidgidresolver.target: loginuser
       sessionid:
         annotations:
           template: uid
--- a/pkg/operators/uidgidresolver/uidgidresolver.go
+++ b/pkg/operators/uidgidresolver/uidgidresolver.go
@@ -19,9 +19,14 @@
 package uidgidresolver
 
 import (
+	"fmt"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/annotations"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


gadgets/trace_exec/gadget.yaml
```diff
@@ -31,20 +31,20 @@ datasources:
         annotations:
           description: Thread ID
           template: pid
-      uid:
+      user_raw:
```
flyth, line 34: Maybe we should still use uid_raw/gid_raw/loginuid_raw?

pkg/operators/uidgidresolver/uidgidresolver.go
```diff
@@ -117,11 +202,52 @@ func (m *UidGidResolverInstance) enrich(ev any) {
 	}
 }
 
+func (m *UidGidResolverInstance) PreStart(gadgetCtx operators.GadgetContext) error {
+	for ds, fieldAccPairs := range m.fieldsUid {
+		for _, fieldAccPair := range fieldAccPairs {
+			ds.Subscribe(func(ds datasource.DataSource, data datasource.Data) error {
+				uid, _ := fieldAccPair.srcFieldAcc.Uint32(data)
```
flyth, line 209: You could return early on error here - same in the loop below.

flyth (commented):
Thanks for making this compatible!

gadgets/trace_exec/gadget.yaml
burak-ok: I changed the name to have the column name be `user` instead of `uid` when we are actually displaying usernames. Would be probably best if the enriched name is `user` and the raw name is `uid(_raw)` as you mentioned.

Is there a way to easily achieve that?

flyth: Hmm, makes sense - but then we also could use `uid` without the `_raw` suffix. So maybe:

* look for a target annotation
* if none is given, replace string `uid` with `user` (`gid` with `group`) and try using that for the target field name - so that it also works for `loginuid` to `loginuser`

pkg/operators/uidgidresolver/uidgidresolver.go
```diff
@@ -84,10 +101,78 @@ func (k *UidGidResolver) Instantiate(gadgetCtx operators.GadgetContext, gadgetIn
...
 
+type fieldAccPair struct {
+	srcFieldAcc datasource.FieldAccessor
+	dstFieldAcc datasource.FieldAccessor
+}
+
+func (k *UidGidResolver) InstantiateDataOperator(gadgetCtx operators.GadgetContext, paramValues api.ParamValues) (operators.DataOperatorInstance, error) {
+	logger := gadgetCtx.Logger()
+	fields_uid := make(map[datasource.DataSource][]fieldAccPair)
+	fields_gid := make(map[datasource.DataSource][]fieldAccPair)
```
mauriciovasquezbernal, line 112: fieldsUid and fieldsGid to keep it uniform with all the code?

mauriciovasquezbernal (approved):
LGTM, I tested and it works fine. 

Some questions:
- Why not to add this to all gadgets?
- Should it be enabled by default? What's the overhead it has?

burak-ok:
> * Can it be tested?

We would need the username of the corresponding uids on the host system before the test runs. Should be possible, right?

> * Why not to add this to all gadgets?

I always try to port only a single gadget when adding a feature. So if the PR has to be reworked several times there won't be too much work to modify every gadget everytime.
After it gets merged I'll post a follow up PR

> * Should it be enabled by default? What's the overhead it has?

There is an fnotify handler on the dir container `passwd` and `group`. Since these files shouldn't change that often reading them will be very rare.

Apart from that there is a go routine checking the `uid<->usermap` for deleted objects in the background every 2 seconds. Enriching is just a matter of looking the `uid` up in a map. So it shouldn't have huge performance implications.

PR #3205, opened 2024-07-23
ebpf-builder.Dockerfile: Use uppercase keyword

burak-ok:
There was a single `as` keyword that was lowercase. While building the image we get a warning:
```
 1 warning found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 9)
```


burak-ok:
> Sadly, we have other occurrences:

Thanks! I forgot to check other occurences. I just got the ebpf builder one because I built the image locally :D

PR #3216, opened 2024-07-24
treewide: Remove unused verify-gadgets flag and adapt CI jobs to dependabot

eiffel-fl:
The verify-gadgets was not used, so this commit removes it. Also, dependabot does not have access to signing related secrets, so this commit introduces a way to skip verification if these secrets cannot be accessed.

Fixes: f51ff0eb2ec8 ("treewide: Allow to have several public keys.")


mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #3217, opened 2024-07-24
ci: Verify eBPF builder image only if cosign secrets were set.

eiffel-fl:
In the case where dependabot bumps a dependency in the eBPF builder image, it will then build it.
But it cannot sign it, so the verification occurring later will fail.


mauriciovasquezbernal (approved):
I rebased #3210 on top of this and worked perfectly. So go for it! Thanks for the fix.

Issue #2903, opened 2024-05-27
[RFC] New gadget: fsnotify-enricher


alban:
RFC: should fsnotify-enricher remain an external gadget or moved to the main Inspektor Gadget repository?

## Current situation

There are no in-tree gadgets tracing the usage of the inotify/fanotify Linux API.

There is one but it is on an external repository:
https://github.com/alban/fsnotify-enricher

## Impact

- Changes to the fsnotify-enricher gadget are not reviewed by the Inspektor Gadget team
- Maintenance is more difficult.
  - For example, IG recently required to use char for strings (see 2607fcdc62ec82cee49d9115a0b7bb186f7e2a34)
```diff
-  __u8 tracer_comm[TASK_COMM_LEN];
+  char tracer_comm[TASK_COMM_LEN];
```

## Ideal future situation

There are a rule of thumb saying when it makes sense to add a gadget in-tree or develop it out-of-tree.

## Implementation options

1. Add fsnotify-enricher to the main repository
2. Archive the external repository

## Additional information

The fsnotify-enricher gadget is tracing the fsnotify APIs (both inotify and fanotify) in a generic way. The ebpf code was originally inspired from the [container-hook pkg](https://github.com/inspektor-gadget/inspektor-gadget/blob/370de192e1bd4326acb85db5ff09834e0f353e94/pkg/container-hook/bpf/execruntime.bpf.c#L89-L90) but it was made generic.

It is useful as a debugging tool for developers of applications using fsnotify APIs because it reveals more information than strace can provide (e.g. both tracer and tracee process metadata).

It is also useful to use as a Go package in applications using fsnotify APIs. Developers would then use the IG Go API. In this scenario, the application would typically filter events from one specific inotify or fanotify file descriptor. The `tracer-group` ebpf param is designed for this. The benefit is the same: it can enrich events you get from the kernel API with additional metadata such as inode information, tracer/tracee process metadata including when running in a different pid namespace (in a container or a Kubernetes pod).

flyth:
IMHO it should become a part of the main repo until we move all image-based gadgets out again (and then it'll still stay in our org). That way we can "guarantee" compatibility with all the changes that are happening right now.

alban:
@flyth Thanks. I'll make a PR to add the gadget to the main repo.

Issue #3132, opened 2024-07-04
wasm: Implement parameters handling 


mauriciovasquezbernal:
A wasm module should be able to

- Define new parameters
- Access value of parameters

Ref https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.30.0/docs/design/003-wasm-support.md#parameters

PR #3218, opened 2024-07-24
New gadget: fsnotify

alban:
# New gadget: fsnotify

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/2903

## How to use

See gadgets/fsnotify/README.md

## Testing done

- Only compiled the new gadget


gadgets/fsnotify/artifacthub-pkg.yml
```diff
@@ -0,0 +1,29 @@
+# Artifact Hub package metadata file
+version: 0.1.0
```
burak-ok, line 2:
```suggestion
version: 0.31.0
```

burak-ok (commented):
A lot have changed in the last week in regarding to the docs. I didn't comment everything in these files

For `artifacthub-pkg.yml`, `gadget.yaml`, `README.md` and `README.mdx` could you check / copy from (for example) `trace_exec` these files and change them to your needs.

Also a symbolic link from `docs/gadgets/fsnotify.mdx` to `gadgets/fsnotify/README.mdx` should be made

alban:
`--update-metadata` lost the metadata for all fields (documentation, attributes, templates) because the fields were moved from `structs` to `datasources`. If such a change in gadget.yaml happens again, we should provide tooling for users to update their gadgets because it is a lot of manual work to update.

gadgets/fsnotify/README.md
```diff
@@ -0,0 +1,5 @@
+# fsnotify
+
+The `fsnotify` gadget detects applications using inotify or fanotify.
+
+Check the full documentation on https://inspektor-gadget.io/docs/latest/gadgets
```
alban, line 5: I removed the /fsnotify suffix because the CI fails.
How to resolve this chicken egg problem?

gadgets/fsnotify/artifacthub-pkg.yml
```diff
@@ -0,0 +1,29 @@
...
+version: 0.31.0
+name: "fsnotify"
+category: monitoring-logging
+displayName: "fsnotify"
+createdAt: "2024-08-12T14:51:00Z"
+digest: "2024-08-12T14:51:00Z"
+description: "enrich inotify or fanotify events"
+logoURL: "https://inspektor-gadget.io/media/brand-icon.svg"
+license: ""
+homeURL: "https://inspektor-gadget.io/docs/latest/gadgets"
```
alban, line 11: I removed the /fsnotify suffix because the link is not yet valid.
How to resolve this chicken egg problem?

gadgets/fsnotify/README.md
mauriciovasquezbernal: I don't see we have any solution there. I'd propose to ignore the CI and use the right link there.

gadgets/fsnotify/artifacthub-pkg.yml
mauriciovasquezbernal: I think it's better to use the right link. We should have an automatic way to update the website when changes are done to the docs folder here. I'll think about it.

alban:
Updated. The only broken link revealed by the CI is https://inspektor-gadget.io/docs/latest/gadgets/fsnotify

burak-ok:
I thought I had posted the comment in my previous review, but I can't find it anymore. Nonetheless I have the following question:

Do we want to call this gadget just `fsnotify`. Isn't this more of a `trace_fsnotify`? This would also align it with all other gadgets

mauriciovasquezbernal (commented):
Some comments. I haven't gone through the whole ebpf code, but please don't block this on me.

gadgets/fsnotify/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+### `--tracer-group`
+Show only fsnotify events generated by a struct fsnotify_group with this address. You can find this address in Golang with kfilefields.ReadPrivateDataFromFd().
+
+Default value: ""
+
+### `--tracer-pid`
+Show only fsnotify events generated for processes with this pid
+
+Default value: ""
```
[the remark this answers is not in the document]
alban: I can add the spaces. But it comes from the template, so ideally we should fix it in the template (#3357).

gadgets/fsnotify/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+```
+
+You can generate events in another terminal with:
+```
+touch /tmp/ABCDE
+```
+
+The fsnotify-enricher gadget can observe and enrich the inotify events in the following way:
+```
+$ sudo -E ig run ghcr.io/inspektor-gadget/gadget/fsnotify:latest --verify-image=false --inotify-only --fields=tracer_comm,tracee_comm,i_wd,i_mask,i_cookie,name
```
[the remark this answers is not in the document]
alban: > -E is not needed anymore :)

Ok, I will remove it.

But in general it is needed to get `$HOME` to get the default docker auth file. In this case, `ghcr.io/inspektor-gadget/gadget/fsnotify:latest` is a public image, but without `-E` won't work for private images.

```bash
$ sudo sh -c 'echo $HOME'
/root
$ sudo -E sh -c 'echo $HOME'
/home/alban
```

alban:
Thanks for the reviews. Updated.

About the naming: since this gadget does not come from a built-in gadget, we don't have the need to keep the category, so "fsnotify" is fine.

alban:
gadget-fsnotify-architecture

alban:
The diagram above explains most of the architecture but some concepts couldn't be represented on it:
- merged events
- filters, particularly how the gadget can be used in combination with `kfilefields.ReadPrivateDataFromFd`
- how fanotify permission responses are sent
- kprobe/fsnotify_destroy_event is called from the tracee in case of merged event or fanotify permission response event, but from the tracer in other cases.

gadgets/fsnotify/artifacthub-pkg.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: gadget
+      image: "ghcr.io/inspektor-gadget/gadget/fsnotify:latest"
+      platforms:
+        - linux/amd64
+        - linux/arm64
+keywords:
+    - gadget
+links:
+    - name: source
+      url: "https://github.com/inspektor-gadget/inspektor-gadget/"
```
mauriciovasquezbernal, line 22: point this to the exact location of the gadget?

gadgets/fsnotify/gadget.yaml
```diff
@@ -0,0 +1,117 @@
+name: fsnotify
+description: Gadget enriching inotify or fanotify events
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://inspektor-gadget.io/docs
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/
```
mauriciovasquezbernal, line 5: ah, not related to this PR, but should probably update all gadgets to point to the specific gadget docs and source.

gadgets/fsnotify/gadget.yaml
```diff
@@ -0,0 +1,117 @@
...
+      fa_flags:
+        annotations:
+          description: flags from fanotify_init
+      fa_mask:
+        annotations:
+          description: Mask describing fanotify event (FAN_ACCESS, FAN_MODIFY, etc.)
+      fa_pid:
+        annotations:
+          description: pid from fanotify_event_metadata. Process ID, or thread ID
+            if flag FAN_REPORT_TID was set in fanotify_init
```
mauriciovasquezbernal, line 21: template: pid?

gadgets/fsnotify/gadget.yaml
```diff
@@ -0,0 +1,117 @@
...
+      fa_response:
+        annotations:
+          description: 'response to a fanotify permission request: allow or deny'
```
mauriciovasquezbernal, line 24: nit: Why `'` are used in some cases and no in others?

gadgets/fsnotify/program.bpf.c
```diff
@@ -0,0 +1,523 @@
...
+const volatile pid_t tracee_pid = 0;
+const volatile bool inotify_only = false;
+const volatile bool fanotify_only = false;
+GADGET_PARAM(tracer_group);
+GADGET_PARAM(tracer_pid);
+GADGET_PARAM(tracee_pid);
+GADGET_PARAM(inotify_only);
+GADGET_PARAM(fanotify_only);
+
+struct gadget_event {
```
mauriciovasquezbernal, line 105: Which one from mntns_id or tracee_mntns_id should be used for enrichment. Currently we only support one of them per struct. 

```bash 
WARN[0000] multiple fields of "type:gadget_mntns_id" found in DataSource "fsnotify", only "mntns_id" will be used
```

mauriciovasquezbernal, line 105: Please either add `_raw` suffix or use the `ebpf.formatter.enum` annotation for enumerations to work: https://www.inspektor-gadget.io/docs/latest/gadget-devel/gadget-ebpf-api#enumerations

```bash
WARN[0000] Failed to get target name for enum field "type": neither "ebpf.formatter.enum" annotation nor '_raw' suffix found
WARN[0000] Failed to get target name for enum field "fa_type": neither "ebpf.formatter.enum" annotation nor '_raw' suffix found
WARN[0000] Failed to get target name for enum field "fa_response": neither "ebpf.formatter.enum" annotation nor '_raw' suffix found
```

gadgets/fsnotify/artifacthub-pkg.yml
mauriciovasquezbernal: For reference, it was done in #3342.

gadgets/fsnotify/README.mdx
```diff
@@ -0,0 +1,117 @@
...
+touch /tmp/ABCDE
+```
+
+The fsnotify-enricher gadget can observe and enrich the inotify events in the following way:
+```bash
+$ sudo ig run ghcr.io/inspektor-gadget/gadget/fsnotify:latest --inotify-only --fields=tracer_comm,tracee_comm,i_wd,i_mask,i_cookie,name
+TRACER_COMM  TRACEE_COMM I_WD I_MASK    I_COOKIE NAME
+inotifywatch touch       1    134217760 0        ABCDE
+inotifywatch touch       1    134217732 0        ABCDE
+inotifywatch touch       1    134217736 0        ABCDE
```
mauriciovasquezbernal, line 79: Is is expected for NAME not to include the whole path?

gadgets/fsnotify/README.mdx
mauriciovasquezbernal, line 1: Please add the link to docs/gadgets too.

mauriciovasquezbernal (commented):
Some general comments. I tested as indicated in the guide and it works fine. Unfortunately I don't have a deep understanding of the fanotify/inotify mechanisms to provide a better review, however what I reviewed in the code makes sense. 

I'll try to make another pass once the these nits are fixed.

gadgets/fsnotify/gadget.yaml
alban: When the string contains a `:`, it needs to be quoted, otherwise the yaml parser gets confused.
I think that `--update-metadata` removes the quotes when not needed.
About `'` vs `"`, I don't know.

gadgets/fsnotify/README.mdx
alban: It's not implemented yet, see: https://github.com/alban/fsnotify-enricher/issues/1

It is not that easy to implement because at the time the kprobes are called, the kernel data structures only contain the name string and the inode number (in the inotify case) but no `struct file` or `struct inode`.

gadgets/fsnotify/gadget.yaml
mauriciovasquezbernal: You got a point!, I didn't realize there were too many `:` in this file.

gadgets/fsnotify/gadget.yaml
alban: Done for other gadgets in #3376

alban:
I think I addressed all the review comments.

After #3284 is completed, I'd like to update the guide to make use of it.

After #3132 is completed, I would like parameters (`--inotify-only`, `--fanotify-only`) to automatically hide/delete some columns (hide fanotify specific columns when `--inotify-only` is specified).

alban:
Merged by squashing commits 1 and 2 as suggested.
CI tests pass (except the URL check on the documentation as expected).

PR #3249, opened 2024-07-29
docs: Update docs format to docusaurus

sniok:
# Update docs format to docusaurus
Part of the migration of the website to docusaurus

## How to use
There will be an instruction how to preview latest doc in the website repository

## Testing done
Preview docs on the new website


docs/core-concepts/custom-resources.md
```diff
@@ -1,14 +1,17 @@
...
-> We are currently removing the Trace resource and using a [gRPC
-> API](../reference/grpc.md) instead. Once done, the Trace resource will be
-> deprecated.
+:::warning
+
+We are currently removing the Trace resource and using a [gRPC
+API](../reference/grpc.md) instead. Once done, the Trace resource will be
+deprecated.
+
+:::
```
mauriciovasquezbernal, line 14: Isn't the previous formatted in a good way? From https://polite-mushroom-00d947803-53.westeurope.1.azurestaticapps.net/docs/latest/core-concepts/custom-resources it seems the format is fine, so if possible I'd like to stick with it so it's also formatted on Github.

sniok: This new format is required for docusaurus. The reason it works in that deployment is because the admonitions are converted (using this [script](https://github.com/inspektor-gadget/website/blob/4533143dfbcd979c6e30905e535f613ca78a1923/tools/docs-fetcher.py#L79)). But when linking the latest docs from the source for the preview we want it to be fully in docusaurus format

docs/crds/gadgets/_index.md
mauriciovasquezbernal, line 1: Can we keep these files? Even if we are not rendering them?

mauriciovasquezbernal (approved):
LGTM. Just one comment about keeping the crd related files on the tree.

docs/crds/gadgets/_index.md
sniok: Yes, we definitely should keep them, good catch. While they are going to appear during preview, in the final build they'll be hidden.

mauriciovasquezbernal:
I created a new-docs branch and merged it there. The idea is to consolidate all docs changes in a branch before merging them to main.

PR #3271, opened 2024-08-01
pkg/datasource: Use FullName to add field.

eiffel-fl:
We mark the field as being replace with its FullName, so we need to use it here too.

```bash
$ ./kubectl-gadget run trace_bind           francis/fix-missing-endpoints-kubelct % u=
K8S.NAMESPACE  K8S.PODNAME    K8S.CONTAINER… ADDR            COMM        PID     TID      UID      GID BOU… K8S.NO… ERROR TIMESTAMP      
default        test-pod       test-pod       :::4242         nc       167030  167030        0        0 0    miniku…       2024-08-01T14:…
^C%
```

### DIFF
--- a/pkg/datasource/data.go
+++ b/pkg/datasource/data.go
@@ -209,7 +209,7 @@ func NewFromAPI(in *api.DataSource) (DataSource, error) {
 	for _, f := range in.Fields {
 		ds.fields = append(ds.fields, (*field)(f))
 		if !FieldFlagUnreferenced.In(f.Flags) {
-			ds.fieldMap[f.Name] = (*field)(f)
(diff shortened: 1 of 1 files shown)


flyth:
> I am wondering if we need the same here:
> 
> https://github.com/inspektor-gadget/inspektor-gadget/blob/87381edcb4f17f6a03270e4c218f62cdf96e734d/pkg/datasource/data.go#L419
> 
> This whole thing about `FullName` vs `Name` is confusing.

Agree - I'll check this.

mauriciovasquezbernal (approved):
This fixes the linked issue, so LGTM. Thanks for the quick fix.

PR #3272, opened 2024-08-01
pkg/gadgets: Fix tracepoint for top block-io.

eiffel-fl:
Fixes: df44b47b082a ("pkg/gadgets: Use tp_btf for top block-io.")


burak-ok (approved):
Thanks for the fast fix

PR #3281, opened 2024-08-04
Columns header is printed twice Fixed

### LINKED ISSUE #3264: Columns header is printed twice for snapshotters
The columns header is being printed twice for snapshotter gadgets. The behavior happens with all binaries:

```bash 
$ ./kubectl-gadget run snapshot_process:latest
K8S.NAMESPACE     K8S.PODNAME       K8S.CONTAINERNAME COMM            PID      TID      UID      GID     PPID K8S.NODE
K8S.NAMESPACE     K8S.PODNAME       K8S.CONTAINERNAME COMM            PID      TID      UID      GID     PPID K8S.NODE
 
$ sudo -E gadgetctl run snapshot_process:latest
RUNTIME.CONTAINERNAME       COMM                       PID            TID            UID            GID           PPID
RUNTIME.CONTAINERNAME       COMM                       PID            TID            UID            GID           PPI

$ sudo -E ig run snapshot_process:latest
RUNTIME.CONTAINERNAME       COMM                       PID            TID            UID            GID           PPID
RUNTIME.CONTAINERNAME       COMM                       PID            TID            UID            GID           PPI
```

andoriyaprashant:
## describe the change in one sentence

The update fixes an issue where column headers were printed twice. A printedHeader variable was added to ensure headers are only printed once by tracking if the header has already been output.

## How to use

Reviewers should run the affected commands to confirm that column headers are now printed correctly and only once.

## Testing done

Commands tested:

- $ ./kubectl-gadget run snapshot_process:latest
- $ sudo -E gadgetctl run snapshot_process:latest
- $ sudo -E ig run snapshot_process:latest

### DIFF
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -259,10 +259,14 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error 
 				}
 			}
 			headerFuncs = append(headerFuncs, printHeader)
-
+			
+			printedHeader := false
 			for _, headerFunc := range headerFuncs {
-				headerFunc()
-			}
+				if !printedHeader {
+					headerFunc()
+					printedHeader = true
+				}
+			}			
 
 			switch ds.Type() {
 			case datasource.TypeSingle:


pkg/operators/cli/clioperator.go
```diff
@@ -259,10 +259,14 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+			
+			printedHeader := false
 			for _, headerFunc := range headerFuncs {
-				headerFunc()
-			}
+				if !printedHeader {
+					headerFunc()
+					printedHeader = true
+				}
+			}
```
blanquicet, line 269: I see two issues with this approach:

1. For `snapshot_socket` and `snapshot_process` gadgets, which don't require the terminal to be cleaned up (datasources don't have `AnnotationClearScreenBefore` annotation), the header is still being printed twice.
2. For `top_*` gadgets, if I set `--map-fetch-interval 10s`, the terminal will be cleaned up, but nothing will be printed for 10 seconds (of course, not even the header). I'm wondering if this may make users think IG is blocked or so.

blanquicet (commented):
Hi @andoriyaprashant, welcome!

Thanks for your contribution. I noticed there are still a couple of situations where we still have issues with the header. Could you please check if we can address them also in this PR?

andoriyaprashant:
> Hi @andoriyaprashant, welcome to the community!
> 
> 
> 
> Thanks for your contribution. I noticed there are still a couple of situations where we still have issues with the header. Could you please check if we can address them also in this PR?

Yes sir, I will try to make the changes you mentioned 👍🏻

mauriciovasquezbernal:
@andoriyaprashant do you have any plans to complete it?

mauriciovasquezbernal:
Superseded by #3641.

Issue #3267, opened 2024-07-31
`--fields` flag doesn't support relative columns (with `+`or `-` prefixes)


mauriciovasquezbernal:
Before [it was possible](https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.30.0/docs/guides/common-features.md#custom-columns) to use the `-o columns=+foo` syntax to show all default columns + foo, but now it's not possible with the fields flag:

```bash 
$ sudo -E ig run trace_exec:latest --fields=-timestamp
[sudo] password for mauriciov:
INFO[0000] Experimental features enabled
WARN[0000] Ignoring runtime "cri-o" with non-existent socketPath "/run/crio/crio.sock"
WARN[0000] Ignoring runtime "podman" with non-existent socketPath "/run/podman/podman.sock"
Error: pre-starting operator "cli": setting fields: column "-timestamp" is invalid
```

PR #3284, opened 2024-08-04
clioperator: Add support for adding / removing fields with --fields

h4l0gen:
# clioperator: Enhance field management with '--fields' tag. Refactored the CLI operator to allow adding and removing fields using the '--fields' tag.

Now able to remove fields,  when specified like this `sudo -E ./ig run trace_exec:latest --fields=-timestamp`

## Testing done
Yes, I've recreated ig binary and tested it locally.
fixes: #3267

### DIFF
--- a/docs/reference/run.mdx
+++ b/docs/reference/run.mdx
@@ -469,6 +469,34 @@ mycontainer                      wget           477678            44591
 </TabItem>
 </Tabs>
 
+It's possible to hide and show some fields by prefixing them with +/-, for instance
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+
+```bash
+$ kubectl gadget run trace_exec:latest --fields=-error,-timestamp
+COMM  PID  TID  PCOMM  PPID  ARGS  K8S.NODE  K8S.NAMESPACE  K8S.PODNAME  K8S.CONTAINERNAME  USER  LOGINUSER  GROUP
+
+$ kubectl gadget run trace_exec:latest --fields=+uid,+pid
+COMM  PID  TID  PCOMM  PPID  ARGS  K8S.NODE  K8S.NAMESPACE  K8S.PODNAME  K8S.CONTAINERNAME  ERROR  USER  LOGINUSER  GROUP  UID
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+
+```bash
+$ sudo ig run trace_exec:latest --fields=-error,-timestamp
+RUNTIME.CONTAINERNAME  COMM  PID  TID  PCOMM  PPID  ARGS  USER  LOGINUSER  GROUP
+
+$ sudo ig run trace_exec:latest --fields=+uid,+pid
+RUNTIME.CONTAINERNAME  COMM  TID  PCOMM  PPID  ARGS  TIMESTAMP  USER  LOGINUSER  GROUP  UID  PID
+```
+
+</TabItem>
+</Tabs>
+
 ## Run for a specific amount of time
 
 Many gadgets will run forever, printing the gathered output until we press
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -17,6 +17,7 @@ package clioperator
 import (
 	"fmt"
 	"os"
+	"slices"
 	"sort"
 	"strings"
 
@@ -179,6 +180,47 @@ func (o *cliOperatorInstance) ExtraParams(gadgetCtx operators.GadgetContext) api
 	return api.Params{fields, mode}
 }
 
+func parseFields(fieldsString string, defaultFields []string) []string {
+	fields := strings.Split(fieldsString, ",")
+
+	addedFields := make([]string, 0, len(fields))
+	deletedFields := make([]string, 0, len(fields))
+	explicitFields := make([]string, 0, len(fields))
+
+	for _, field := range fields {
+		field = strings.TrimSpace(field)
+		if field == "" {
+			continue
+		}
+		switch field[0] {
+		case '+':
+			addedFields = append(addedFields, field[1:])
+		case '-':
+			deletedFields = append(deletedFields, field[1:])
+		default:
+			if !slices.Contains(explicitFields, field) {
+				explicitFields = append(explicitFields, field)
+			}
+		}
+	}
+
+	result := defaultFields
+	if len(explicitFields) > 0 {
+		result = explicitFields
+	}
+
+	for _, field := range addedFields {
+		if !slices.Contains(result, field) {
+			result = append(result, field)
+		}
+	}
+
+	for _, field := range deletedFields {
+		result = slices.DeleteFunc(result, func(s string) bool { return s == field })
+	}
+	return result
+}
+
 func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error {
 	params := apihelpers.ToParamDescs(o.ExtraParams(gadgetCtx)).ToParams()
 	params.CopyFromMap(o.paramValues, "")
@@ -237,7 +279,8 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error 
 			formatter := p.GetTextColumnsFormatter()
 
 			if hasFields {
-				err := formatter.SetShowColumns(strings.Split(fields, ","))
+				parsedFields := parseFields(fields, defCols)
+				err = formatter.SetShowColumns(parsedFields)
 				if err != nil {
 					return fmt.Errorf("setting fields: %w", err)
 				}


h4l0gen:
Here, how we can use it after these changes:

when no flag is provided, comes with default headers
```
sudo -E ./ig run trace_exec:latest 
RUNTIME.CONTAINERNAME         COMM                        PID             TID PCOMM                      PPID ARGS            ERROR TIMESTAMP                         USER            LOGINUSER      GROUP 
```
default headers after removing `timestamp` field
```
$ sudo -E ./ig run trace_exec:latest --fields=-timestamp
COMM                             PID                 TID PCOMM                           PPID ARGS                RUNTIME.CONTAINERNAME              ERROR USER               LOGINUSER          GROUP
```
default headers along with `uid` header
```
$ sudo -E ./ig run trace_exec:latest --fields=+uid
COMM                      PID            TID PCOMM                    PPID ARGS           RUNTIME.CONTAINERNAME      ERROR TIMESTAMP                       USER           LOGINUSER     GROUP                   UID
```
default headers with `timestamp`, `error` is removed and `uid`, `gid` added
```
$ sudo -E ./ig run trace_exec:latest --fields=-timestamp,-error,+uid,+gid
COMM                          PID              TID PCOMM                        PPID ARGS             RUNTIME.CONTAINERNAME         USER            LOGINUSER       GROUP                       UID             GID
```
only given headers will be there
```
$ sudo -E ./ig run trace_exec:latest --fields=comm,pid
COMM                      PID
```

pkg/operators/cli/clioperator.go
```diff
@@ -237,7 +276,11 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 			formatter := p.GetTextColumnsFormatter()
 
 			if hasFields {
-				err := formatter.SetShowColumns(strings.Split(fields, ","))
+				parseFields, err := parseFields(fields, defCols)
```
mauriciovasquezbernal, line 279: nit: 
```suggestion
				parsedFields, err := parseFields(fields, defCols)
```

(to avoid having a function and a variable with the same name)

pkg/operators/cli/clioperator.go
```diff
@@ -179,6 +179,45 @@ func (o *cliOperatorInstance) ExtraParams(gadgetCtx operators.GadgetContext) api
 	return api.Params{fields, mode}
 }
 
+func parseFields(fieldsString string, defaultFields []string) ([]string, error) {
+    fields := strings.Split(fieldsString, ",")
+    result := make([]string, len(defaultFields))
```
mauriciovasquezbernal, line 184: What about using two maps, one for "hide" and one for "show" fields to be able to check things like the same field is hidden but shown, etc. 

Something similar to the implementation in https://github.com/inspektor-gadget/inspektor-gadget/blob/7ec66979122fb7e78bc34bd7cedf2e3db8eed24d/pkg/datasource/formatters/json/json.go#L94-L113. 

(One advantage of using map is that removing elements from it is very easy, also we don't need to worry about duplicates)

h4l0gen: Hi @mauriciovasquezbernal now using, separate maps for removing and adding fields. And also checking duplicates at any stage. And finally show combined result. Please review and provide your feedback

mauriciovasquezbernal: Unfortunately I think the maps wasn't a good idea as it messes up with the order of the fields. I think we need to switch back to slices and use helpers like slices.Contains to avoid doing a lot of manual work.

pkg/operators/cli/clioperator.go
```diff
@@ -179,6 +179,59 @@ func (o *cliOperatorInstance) ExtraParams(gadgetCtx operators.GadgetContext) api
...
+            addedFields[fieldName] = struct{}{}
+        case '-':
+            fieldName := field[1:]
+            if _, ok := addedFields[fieldName]; ok {
+                return nil, fmt.Errorf("field %q both added (+) and removed (-)", fieldName)
+            }
+            hideFields[fieldName] = struct{}{}
+        default:
+            // If no prefix, return only these fields
+            return strings.Split(fieldsString, ","), nil
```
mauriciovasquezbernal, line 213: This would fail if something like `--fields=timestamp,+error` is passed. I think in that case the timestamp and error columns should be shown.

h4l0gen: Hi @mauriciovasquezbernal, sorry for late changes, involved in creating LFX proposal. I've made changes, 
It uses, `slices.DeleteFunc` for remove fields. Also added `fields=timestamp,+error` case.

alban:
Thanks for working on this.

Could you update the documentation to specify how to use the syntax with + and -?

In the same file line 118/119:
> "Available data sources / fields"

In the reference documentation:
https://www.inspektor-gadget.io/docs/latest/reference/run#selecting-specific-fields

pkg/operators/cli/clioperator.go
```diff
@@ -237,7 +268,8 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 			formatter := p.GetTextColumnsFormatter()
 
 			if hasFields {
-				err := formatter.SetShowColumns(strings.Split(fields, ","))
+				parseFields := parseFields(fields, defCols)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Also be sure to remove any trailing white space from this line.

pkg/operators/cli/clioperator.go
```diff
@@ -179,6 +180,36 @@ func (o *cliOperatorInstance) ExtraParams(gadgetCtx operators.GadgetContext) api
 	return api.Params{fields, mode}
 }
 
+func parseFields(fieldsString string, defaultFields []string) []string {
+    fields := strings.Split(fieldsString, ",")
+    result := make([]string, len(defaultFields))
```
mauriciovasquezbernal, line 185:
```suggestion
    result := make([]string, 0, len(defaultFields))
```

(otherwise the result will contain a lot of empty elements and fail)

pkg/operators/cli/clioperator.go
```diff
@@ -179,6 +180,36 @@ func (o *cliOperatorInstance) ExtraParams(gadgetCtx operators.GadgetContext) api
 	return api.Params{fields, mode}
 }
 
+func parseFields(fieldsString string, defaultFields []string) []string {
```
mauriciovasquezbernal, line 183: There are some corner cases that aren't handled by this implementation:

```bash 
# It should not fail:
$ sudo ig run trace_exec:latest --fields=-timestamp
Error: starting operators: pre-starting operator "cli": setting fields: column "" is invalid

# it should print timestamp and error 
$ sudo ig run trace_exec:latest --fields=+timestamp,error
                                                                                                                                   ERROR
```

What do you think about this alternative to handle them :

```go
func parseFields(fieldsString string, defaultFields []string) []string {
	fields := strings.Split(fieldsString, ",")

	addedFields := []string{}
	deletedFields := []string{}
	explicitFields := []string{}

	for _, field := range fields {
		field = strings.TrimSpace(field)
		if field == "" {
			continue
		}
		switch field[0] {
		case '+':
			addedFields = append(addedFields, field[1:])
		case '-':
			deletedFields = append(deletedFields, field[1:])
		default:
			if !slices.Contains(explicitFields, field) {
				explicitFields = append(explicitFields, field)
			}
		}
	}

	result := defaultFields
	if len(explicitFields) > 0 {
		result = explicitFields
	}

	for _, field := range addedFields {
		if !slices.Contains(result, field) {
			result = append(result, field)
		}
	}

	for _, field := range deletedFields {
		result = slices.DeleteFunc(result, func(s string) bool { return s == field })
	}

	return result
}
```

mauriciovasquezbernal (changes requested):
Some additional comments. Please be sure to run `make lint` to format your code.

Please also squash your commits.

pkg/operators/cli/clioperator.go
h4l0gen: Yes @mauriciovasquezbernal this is little different approach and looks really good to me.  
```
addedFields := make([]string, 0, len(fields))
	deletedFields := make([]string, 0, len(fields))
	explicitFields := make([]string, 0, len(fields))
```
just doing this little change `Preallocate slices` as this can improve performance slightly by reducing the number of memory allocations.

I was also thinking of using trimming space, but its not that much required. 
its working fine now.
```bash
$ sudo -E ./ig run trace_exec:latest --fields=+timestamp,error 
[sudo] password for h4logen: 
ERROR                              TIMESTAMP 

```
Please tell me is this looks good to you? then I will change documentation regarding this. Thank you

mauriciovasquezbernal: > just doing this little change Preallocate slices as this can improve performance slightly by reducing the number of memory allocations.

I'm not sure about this since you're preallocating all three slices to have the same size of the input fields, i.e. you're preallocating more space than will be ever used. 
Also, this code is not part of a hot path, it's only run before running the gadget and the number of fields are rather small  (<100), so I don't think performance here is too critical. 

In any case, I don't have a strong opinion on the approach, any is good to me.

mauriciovasquezbernal: > Please tell me is this looks good to you? then I will change documentation regarding this. Thank you

Yes, it LGTM. thanks! 

btw. please format your commit message:
- title: add a "clioperator:" prefix 
- body: add a description of **what** is being done, something similar to the PR description

h4l0gen: @mauriciovasquezbernal I've changed commit message.

mauriciovasquezbernal: I still see the commit message as 

```
Signed-off-by: h4l0gen <ks3913688@gmail.com>

using addedfield and hidefield map logic

Signed-off-by: h4l0gen <ks3913688@gmail.com>

changing back to slices instead of map

Signed-off-by: h4l0gen <ks3913688@gmail.com>

including corner cases left

Signed-off-by: h4l0gen <ks3913688@gmail.com>

including corner cases left

Signed-off-by: h4l0gen <ks3913688@gmail.com>
```

Would you mind fixing it to have:
- a short description of the change in this commit
- your signoff 

(you can use `git commit --amend` to edit it)

mauriciovasquezbernal: Hi, you put all the content on the title. Please split that into title + body. 

Also, please don't forget [this](https://github.com/inspektor-gadget/inspektor-gadget/pull/3284#issuecomment-2296303301) comment about adding documentation.

h4l0gen: @mauriciovasquezbernal added documentation, please review it.

mauriciovasquezbernal: looks good. Can you also add the kubectl-gadget one to keep it uniform?

h4l0gen: @mauriciovasquezbernal made changes, we are good to go now, Please review.

docs/reference/run.mdx
```diff
@@ -510,6 +510,34 @@ mycontainer                      wget           477678            44591
...
+It's possible to hide and show some fields by prefixing them with +/-, for instance
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+
+```bash
+$ kubectl gadget run trace_exec:latest --fields=-error,-timestamp
+COMM  PID  TID  PCOMM  PPID  ARGS  K8S.NODE  K8S.NAMESPACE  K8S.PODNAME  K8S.CONTAINERNAME  USER  LOGINUSER  GROUP
+
+$ ./kubectl-gadget run trace_exec:latest --fields=+uid,+pid
```
mauriciovasquezbernal, line 522: kubectl gadget

mauriciovasquezbernal (commented):
Last final comment. Other than that, LGTM!

docs/reference/run.mdx
h4l0gen: @mauriciovasquezbernal please check. Thank you 🙏

mauriciovasquezbernal: I still see ./kubect-gadget.

mauriciovasquezbernal (approved):
LGTM. Will merge once CI is green.

PR #3291, opened 2024-08-05
guides: Add profile_blockio

mqasimsarfraz:
Add guide for profile_blockio

### DIFF
--- a/gadgets/profile_blockio/README.mdx
+++ b/gadgets/profile_blockio/README.mdx
@@ -1,0 +1,247 @@
+---
+title: profile_blockio
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# profile_blockio
+
+The `profile_blockio` gadget gathers information about the usage of the
+block device I/O (disk I/O), generating a histogram distribution of I/O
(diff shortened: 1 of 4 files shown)


gadgets/profile_blockio/README.mdx
```diff
@@ -0,0 +1,210 @@
...
+        ```
+
+        Now, let's increase the I/O operations using the stress tool:
+
+        ```bash
+        # Start by creating our testing namespace
+        $ kubectl create ns test-biolatency
+
+        # Run stress with 1 worker that will generate I/O operations
+        $ kubectl run --restart=Never --image=polinux/stress stress-io -n test-biolatency -- stress --io 1
```
burak-ok, line 102: Shouln't we use `-hdd` instead?
```suggestion
        $ kubectl run --restart=Never --image=polinux/stress stress-io -n test-biolatency -- stress --hdd 1
```

mqasimsarfraz: This is coming from: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/builtin-gadgets/profile/block-io.md . I will prefer to keep it as it is for now and can improve it follow up PR?

Issue #3241, opened 2024-07-26
gadgets/trace_tcpretrans: Missing tests


mauriciovasquezbernal:
The [trace_tcpretrans](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/trace_tcpretrans) gadget doesn't have any test implemented. There is not a built-in version of this gadget, hence these tests need to be implemented from scratch.

mastersans:
I'll like to work on this one.

PR #3294, opened 2024-08-05
feat(test): added test for gadgets/trace_tcpretrans 

mastersans:
fixes: #3241
- added test for trace tcp retransmission gadget.

cc @burak-ok @mauriciovasquezbernal 
### Testing done: Yes
I ran test locally and heres output :
Screenshot from 2024-08-05 17-42-24

### DIFF
--- a/gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
+++ b/gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
@@ -1,0 +1,140 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+	igtesting "github.com/inspektor-gadget/inspektor-gadget/pkg/testing"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/containers"
+	igrunner "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/ig"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/match"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+)
+
+type traceTCPretransEvent struct {
+	eventtypes.CommonData
+
+	Timestamp string `json:"timestamp"`
+	MntNsID   uint64 `json:"mntns_id"`
+	NetNs     uint64 `json:"netns_id"`
+
+	Pid uint32 `json:"pid"`
+	Tid uint32 `json:"tid"`
+	Uid uint32 `json:"uid"`
+	Gid uint32 `json:"gid"`
+
+	Comm string           `json:"comm"`
+	Type string           `json:"type"`
+	Src  utils.L4Endpoint `json:"src"`
+	Dst  utils.L4Endpoint `json:"dst"`
+}
+
+func TestTraceTCPretrans(t *testing.T) {
+	gadgettesting.RequireEnvironmentVariables(t)
+	utils.InitTest(t)
+
+	if utils.CurrentTestComponent == utils.KubectlGadgetTestComponent {
+		t.Skip("Skipping test as K8s doesn't support privileged containers")
+	} else if utils.CurrentTestComponent == utils.IgLocalTestComponent && utils.Runtime == "containerd" {
+		t.Skip("Skipping test as containerd test utils can't use the network")
+	}
+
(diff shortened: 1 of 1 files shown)


gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestTraceTCPretrans(t *testing.T) {
+	gadgettesting.RequireEnvironmentVariables(t)
+	utils.InitTest(t)
+
+	containerFactory, err := containers.NewContainerFactory(utils.Runtime)
+	require.NoError(t, err, "new container factory")
+	containerName := "test-trace-tcpretrans"
+	containerImage := "wbitt/network-multitool:latest"
```
mauriciovasquezbernal, line 56: Please use the full name of the image: `docker.io/wbitt/network-multitool:latest` (we've seen situations when not using the registry causes issues to run the tests)

gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	var ns string
+	//run the container with privileged mode to be able to use tc command
```
mauriciovasquezbernal, line 59: nit: add space after `// ` in comments.

gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
```diff
@@ -0,0 +1,126 @@
...
+	containerOpts := []containers.ContainerOption{containers.WithContainerImage(containerImage), containers.WithPrivileged()}
+
+	if utils.CurrentTestComponent == utils.KubectlGadgetTestComponent {
+		ns = utils.GenerateTestNamespaceName(t, "test-trace-tcpretrans")
+		containerOpts = append(containerOpts, containers.WithContainerNamespace(ns))
+	}
+	//for testing purpose drop 25% of packets and wget 1.1.1.1 every 0.1 seconds
+	cmds := "tc qdisc add dev eth0 root netem drop 30% && while true; do wget --no-check-certificate -q -O /dev/null 1.1.1.1; sleep 0.1; done"
```
mauriciovasquezbernal, line 67: use https://1.1.1.1 to be sure the destination port is always 443.

gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
```diff
@@ -0,0 +1,126 @@
...
+	NetNs     uint64 `json:"netns_id"`
+
+	Pid uint32 `json:"pid"`
+	Tid uint32 `json:"tid"`
+	Uid uint32 `json:"uid"`
+	Gid uint32 `json:"gid"`
+
+	Comm string           `json:"comm"`
+	Type string           `json:"type"`
+	Dst  utils.L4Endpoint `json:"dst"`
```
mauriciovasquezbernal, line 46: Why is src missing?

mauriciovasquezbernal (commented):
Looks very good, thanks! Two minor comments and a question.

gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
mastersans: Hi @mauriciovasquezbernal,
I wasn't sure about the IP for the container, and I also wanted to ask about using tc to emulate packet drop, which requires NET_ADMIN capabilities. With my understanding of IG, there isn't a way to add individual capabilities to test containers. As a result, I am running the container with privileges, which may cause errors with K8s since it doesn't support privileged containers. Is there a workaround for this? Should we add something like container.AddCapabilities() to testutils? I hope I didn't miss something obvious.
Thanks!

mauriciovasquezbernal: I think we can at least check that the field is present. (Using the normalize logic)

Regarding capabilities and privileged, I think for testing purposes it's just fine to be able to run the pod / container as privileged. Regarding the k8s thing, I'd suggest to skip this test on k8s, you can use something like:

https://github.com/inspektor-gadget/inspektor-gadget/blob/84d2519362b494848d31d4ddf9cfa2fcbf87cf37/gadgets/trace_dns/test/trace_dns_test.go#L71-L73

(checking for utils.CurrentTestComponent == utils.KubectlGadgetTestComponent should work)

mastersans: @mauriciovasquezbernal I have made the desired changes and squashed the commits for this one, Thanks!

mauriciovasquezbernal (approved):
LGTM. I'll merge once the CI passes. Thanks!

mastersans:
> LGTM. I'll merge once the CI passes. Thanks!

@mauriciovasquezbernal not sure why retrans test is not capturing and failing, I'll find solution and will let you know.

gadgets/trace_tcpretrans/test/trace_tcpretrans_test.go
```diff
@@ -0,0 +1,140 @@
...
+	var ns string
+	// run the container with privileged mode to be able to use tc command
+	containerOpts := []containers.ContainerOption{containers.WithContainerImage(containerImage), containers.WithPrivileged()}
+
+	if utils.CurrentTestComponent == utils.KubectlGadgetTestComponent {
+		ns = utils.GenerateTestNamespaceName(t, "test-trace-tcpretrans")
+		containerOpts = append(containerOpts, containers.WithContainerNamespace(ns))
+	}
+	// for testing purpose drop 30% of packets and wget 1.1.1.1 every 0.1 seconds
+	cmds := "tc qdisc add dev eth0 root netem drop 30% && while true; do wget --no-check-certificate -q -O /dev/null https://1.1.1.1; sleep 0.1; done"
```
mauriciovasquezbernal, line 74: Perhaps increase the drop probability so we're sure packets are dropped while running the test.

mauriciovasquezbernal: You're right. I don't think that's the issue.

Issue #2173, opened 2023-10-23
`--max-rows` doesn't work for snapshot gadgets


mauriciovasquezbernal:
Snapshot gadgets provide a `--max-rows` flag:

```bash
$ sudo ig snapshot process -h | grep max-rows
  -m, --max-rows uint32        Maximum number of rows to return (default 50)

$ sudo ig snapshot socket -h | grep max-rows
  -m, --max-rows uint32        Maximum number of rows to return (default 50)
```

However the flag doesn't have any effect:

```bash
$ sudo ig snapshot process --threads --max-rows 5 | wc -l 
27

$ sudo ig snapshot process --threads --max-rows 10 | wc -l 
27

$ sudo ig snapshot socket --host --max-rows 5 | wc -l 
126

$ sudo ig snapshot socket --host --max-rows 10 | wc -l 
127
```

### Implementation options

- Make snapshot gadgets use this flag
- Remove this flag for snapshot gadgets

mauriciovasquezbernal:
This was fixed by #3327 for image-based gadgets. For built-in gadgets we won't provide a fix.

PR #3303, opened 2024-08-06
 Remove --max-rows Flag from Snapshot Gadgets

kunal-511:
#2173   Remove --max-rows Flag from Snapshot Gadgets 

This PR addresses the issue of the --max-rows flag not having any effect on the snapshot process and snapshot socket commands. The flag has been removed to prevent confusion and ensure the commands operate as expected without unused options.

## How to use

To validate this PR, reviewers should build the inspektor-gadget executable and run the snapshot process and snapshot socket commands to ensure they work correctly without the --max-rows flag.

## Testing done

1. Built the inspektor-gadget executable.
2. Ran the snapshot process and snapshot socket commands to ensure they function without the --max-rows flag.
3. Confirmed the commands produce the expected output without any errors.

### DIFF
--- a/pkg/gadgets/params.go
+++ b/pkg/gadgets/params.go
@@ -25,7 +25,8 @@ import (
 const (
 	ParamInterval = "interval"
 	ParamSortBy   = "sort"
-	ParamMaxRows  = "max-rows"
+
+// ParamMaxRows  = "max-rows"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


blanquicet (commented):
Hi @kunal-511. Thanks for this contribution.

With your proposal, you're removing the flag also from toppers, which indeed need it.

mauriciovasquezbernal:
This was fixed by #3327. Now the flag is called `--max-entries`. Thanks for your contribution!

PR #3307, opened 2024-08-06
columns: add support for displaying event data in Hex format

Snehil-Shah:
# Adds support for displaying event data (like addresses) in Hexadecimal format

image

- Adds `columns.hex` annotation field for datasources in `gadget.yaml` configuration.
- Adds `address` template with `columns.hex: true`. I wasn't sure, what other annotations to include for *address* columns for better formatting so suggestions are welcome
- Enabling this formats integer type entries to hexadecimal in the format `0x{hex_addr}` (as shown in the image above).

## How to use

- Add the `columns.hex` annotation to any integer type datasource in any gadget's yaml configuration.
- Build and run the gadget.
- Validate the column entries.

## Testing done

The above steps were used to validate the PR.

### DIFF
--- a/pkg/datasource/columns.go
+++ b/pkg/datasource/columns.go
@@ -39,6 +39,7 @@ const (
 	ColumnsEllipsisAnnotation  = "columns.ellipsis"
 	ColumnsHiddenAnnotation    = "columns.hidden"
 	ColumnsFixedAnnotation     = "columns.fixed"
+	ColumnsHexAnnotation       = "columns.hex"
 
 	DescriptionAnnotation = "description"
 	TemplateAnnotation    = "template"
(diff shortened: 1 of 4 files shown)


pkg/datasource/columns.go
```diff
@@ -262,6 +267,9 @@ var annotationsTemplates = map[string]map[string]string{
 		ColumnsMinWidthAnnotation:  "8",
 		ColumnsAlignmentAnnotation: string(metadatav1.AlignmentRight),
 	},
+	"address": {
+		ColumnsHexAnnotation: "true",
+	},
```
mauriciovasquezbernal, line 272: I think it's too generic. It can be a memory address, network address, etc. IMO we don't need a template for this, we can set it each field that wants to use this feature, like the address field of the trace_malloc gadget: 

https://github.com/inspektor-gadget/inspektor-gadget/blob/f443360a9f8261801c4444a8b45e7ac94289f25f/gadgets/trace_malloc/gadget.yaml#L45-L48

mauriciovasquezbernal (commented):
It looks very good, some minor comments.

Snehil-Shah:
@mauriciovasquezbernal Done, sorry for the oversight.

mauriciovasquezbernal (approved):
This looks great. Thanks!

PR #3315, opened 2024-08-07
pkg/gadgets: Fix tree output for snapshot process.

eiffel-fl:
With the changes added in d61e5ef270f1
("cc: Move BasicK8sMetadata to pkg/types and use it in CommonData") this output mode no more worked in non k8s context as there would be an empty string for K8S.ContainerName.
Let's use the mount namespace ID to group processes in a container instead.

Fixes: d61e5ef270f1 ("cc: Move BasicK8sMetadata to pkg/types and use it in CommonData")


burak-ok (approved):
Makes sense. LGTM and thanks

Issue #3184, opened 2024-07-18
testing: Containerd containers can't access the network


mauriciovasquezbernal:
Some of our tests require network connectivity to work, but the containers we create with `NewContainerdContainer` can't access the network. For instance, the trace_sni test shows the following (after removing the [logic](https://github.com/inspektor-gadget/inspektor-gadget/blob/0140bf3f9a9f5431956eafb658a797e4ca6c0743/gadgets/trace_sni/test/trace_sni_test.go#L50-L52) to skip the test on containerd)

Run the test

```bash 
$ cd gadgets/trace_sni/test
$ IG_FLAGS="--verify-image=false" IG_PATH=ig IG_RUNTIME=containerd go test -exec 'sudo -E' ./... -
```

Check the container output in another terminal

```bash 
$ sudo nerdctl --namespace k8s.io attach test-trace-sni
wget: bad address 'inspektor-gadget.io'
wget: bad address 'inspektor-gadget.io'
wget: bad address 'inspektor-gadget.io'
wget: bad address 'inspektor-gadget.io'
...
```

The created containers should have network connectivity to allow those test to run.

Ref #3172

h4l0gen:
Hi, I am looking into this issue, please assign it to me.

mastersans:
Hi @h4l0gen this one also affect one of the test https://github.com/inspektor-gadget/inspektor-gadget/pull/3294 I wrote So incase if you decide to work on some other issue do let me know I would be more than happy to take this one.

h4l0gen:
sure @mastersans changes are done, I am about to create a PR, please check once if that resolve your problem. Thank you!

PR #3318, opened 2024-08-07
containerd containers can use host network now

h4l0gen:
# containerd containers can use network now

containerd container can use host network connection for trace_sni test.

## How to use

Run test, with containerd, and containers can connect to network now. 

## Testing done

YES
fixes: #3184


h4l0gen:
Hi @mauriciovasquezbernal, main changes:
- Here I decided to always use the host network for containerd containers in the test environment.
- in pkg/container-utils/testutils/containerd.go:
  Modified the Run method to always use the host network
  
  I was also thinking of using `Bridge Network` or `CNI` approach instead of using host's network directly. Please @mauriciovasquezbernal provide your thoughts on this. Thank you!

h4l0gen:
image

mauriciovasquezbernal:
> I was also thinking of using Bridge Network or CNI approach instead of using host's network directly. Please @mauriciovasquezbernal provide your thoughts on this. Thank you!

It'd be much better if it uses a different namespace. Do you think it's easy to implement?

gadgets/trace_sni/test/trace_sni_test.go
```diff
@@ -69,7 +65,7 @@ func TestTraceSNI(t *testing.T) {
 	testContainer := containerFactory.NewContainer(
 		containerName,
 		"while true; do setuidgid 1000:1111 wget --no-check-certificate -T 2 -q -O /dev/null https://inspektor-gadget.io; sleep 0.1; done",
-		containerOpts...,
+		containers.WithContainerImage(containerImage),
```
mauriciovasquezbernal, line 68: Are all changes related to image really needed?

h4l0gen: Hi @mauriciovasquezbernal, yes you are right, actually they don't require, I just remembered I was doing some debugging with that and forgot to change it back.

h4l0gen:
Hi @mauriciovasquezbernal , I've tried a lot with `netns` package. But not able to capture event from container. Its something that, namespace is not setting up correctly for container. Its really getting complicated. Additionally, I found this [blog](https://medium.com/@teddyking/namespaces-in-go-network-fdcf63e76100)  
To use this:
- Compile the netsetgo.go file and place the binary in /usr/local/bin/.
- Set the correct permissions: sudo chown root:root /usr/local/bin/netsetgo && sudo chmod 4755 /usr/local/bin/netsetgo

I don't know, it would suit with IG tool or not. What you think about it?

Also 
> It'd be much better if it uses a different namespace. Do you think it's easy to implement?

Yes, its little complicated, but if you really want using different namespace approach, I will try more to solve it.

mauriciovasquezbernal:
> Hi @mauriciovasquezbernal , I've tried a lot with `netns` package. But not able to capture event from container. Its something that, namespace is not setting up correctly for container. Its really getting complicated. Additionally, I found this [blog](https://medium.com/@teddyking/namespaces-in-go-network-fdcf63e76100) To use this:
> 
>     * Compile the netsetgo.go file and place the binary in /usr/local/bin/.
> 
>     * Set the correct permissions: sudo chown root:root /usr/local/bin/netsetgo && sudo chmod 4755 /usr/local/bin/netsetgo
> 
> 
> I don't know, it would suit with IG tool or not. What you think about it?

I'm sorry, I can't follow your message. Could you please try to explain (1) what you are trying to do and (2) the problems you're facing?

Issue #3260, opened 2024-07-30
[RFE] K8s Context Enrichment - Controller Kind & Name


galofir-ms:
Hello

We would like to be able to correlate ebpf events with the originating controller, in addition to existing enrichments such as k8s and container runtime.

I understand this might require a query to API server and a non trivial request

Lets discuss

alban:
Done for kubectl-gadget in #3319.
For `ig`, it is not yet done.

PR #3319, opened 2024-08-07
container-collection: Add OwnerKind and OwnerName to K8sEnrichment

burak-ok:
When using the `KubernetesEnrichment` resolve also the Owner Kind and Owner Name. Partially solves https://github.com/inspektor-gadget/inspektor-gadget/issues/3260

For `ig w/ K8s` we need to talk to the API server. This is still TODO, needs to be discussed and should be optional (either enrichment failing siliently or actived through flag)

## image based gadgets
```
./kubectl-gadget run trace_exec -n gadget --fields k8s.podname,k8s.owner.kind,k8s.owner.name,comm
K8S.PODNAME                        K8S.OWNER.KIND     K8S.OWNER.NAME     COMM            
gadget-mdwnx                       DaemonSet          gadget             gadgettracerman 
gadget-mdwnx                       DaemonSet          gadget             gadgettracerman 
```

## Built-in gadgets
```
./kubectl-gadget trace exec -n gadget -o columns=k8s.podname,k8s.owner.kind,k8s.owner.name,comm 
K8S.PODNAME                        K8S.OWNER.KIND     K8S.OWNER.NAME     COMM            
gadget-mdwnx                       DaemonSet          gadget             gadgettracerman 
gadget-mdwnx                       DaemonSet          gadget             gadgettracerman
```

### DIFF
--- a/pkg/container-collection/options.go
+++ b/pkg/container-collection/options.go
@@ -517,64 +517,69 @@ func WithKubernetesEnrichment(nodeName string, kubeconfig *rest.Config) Containe
 
 		// Future containers
 		cc.containerEnrichers = append(cc.containerEnrichers, func(container *Container) bool {
-			// Skip enriching if all k8s fields are already known.
+			// Skip enriching if basic k8s fields are already known.
 			// This is an optimization and to make sure to avoid erasing the fields in case of error.
-			if runtimeclient.IsEnrichedWithK8sMetadata(container.K8s.BasicK8sMetadata) {
-				return true
-			}
-
-			var pod *corev1.Pod
-			var err error
-			if container.K8s.PodName == "" || container.K8s.Namespace == "" {
-				pod, err = getPodByCgroups(clientset, nodeName, container)
-				if err != nil {
-					log.Errorf("kubernetes enricher (from UID): cannot find pod for container %s: %s", container.Runtime.ContainerName, err)
-					return false
-				}
-			} else {
-				pod, err = clientset.CoreV1().Pods(container.K8s.Namespace).Get(context.TODO(), container.K8s.PodName, metav1.GetOptions{})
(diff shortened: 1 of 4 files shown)


pkg/container-collection/options.go
```diff
@@ -517,64 +517,69 @@ func WithKubernetesEnrichment(nodeName string, kubeconfig *rest.Config) Containe
...
+				if container.K8s.PodName != "" && container.K8s.ContainerName == "" {
+					return false
+				}
 			}
 
+			if container.K8s.ownerReference == nil {
+				_, err = container.GetOwnerReference()
+				if err != nil {
+					log.Errorf("kubernetes enricher: failed to get owner reference for container %s: %s", container.Runtime.ContainerID, err)
+					return false
```
alban, line 580: `return false`: Should the container really be dropped if fetching the owner from the api server fails?

alban (approved):
One question about the `return false`, LGTM otherwise.

pkg/container-collection/options.go
burak-ok: That is a good question. I think `GetOwnerReference` only failys if there are probablems with the API server. So I would say that is an error worth reporting.

But is dropping then the right choice? Idk

alban: I would keep the `log.Errorf` but remove the `return false`.

If users run `kubectl delete replicaset foo --force`, then the replica set will be deleted immediately, so the call to the api server will fail. But the processes in the container will not be terminated immediately, so we still need to keep the container in the container collection to enrich ebpf events.

mauriciovasquezbernal:
I think this breaks the CI -> https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10355549023/job/28663453554

PR #3136, opened 2024-07-05
gadgettracermanager: Set environment to Kubernetes

mqasimsarfraz:
Fixes: #3014 

### Before this change
```
$ make minikube-deploy
$ ./kubectl-gadget run trace_open:latest -A
INFO[0000] Experimental features enabled
RUNTIME.CONTAINERNAME   PID          UID          GID          MNTNS_ID  ERR FD          COMM        FNAME                   TIMESTAMP
```

### After this change

```
$ make minikube-deploy
$ ./kubectl-gadget run trace_open:latest -A
→ ./kubectl-gadget run trace_open -A
INFO[0000] Experimental features enabled                
K8S.NAMESPACE        K8S.PODNAME          K8S.CONTAINERNAME    PID         UID         GID         MNTNS_ID E… FD          COMM       FNAME                 MODE       FLAGS      K8S.NODE   TIMESTAMP             
gadget               gadget-lsknt         gadget               844834      0           0           40265331  0 66          gadgettrac /sys/kernel/debug/tra ---------- O_RDONLY|O minikube-d 2024-07-05T12:45:15.66
gadget               gadget-lsknt         gadget               844834      0           0           40265331  0 68          gadgettrac /host/proc            ---------- O_RDONLY|O minikube-d 2024-07-05T12:45:16.25
gadget               gadget-lsknt         gadget               845840      0           0           40265331  0 5           runc:[2:IN /sys/kernel/mm/transp ---------- O_RDONLY   minikube-d 2024-07-05T12:45:16.55

```

Issue #3014, opened 2024-06-13
kubectl-gadget: Image-based gadgets show the wrong columns


mauriciovasquezbernal:
Image based gadgets when run on Kubernetes show the runtime columns by default and not the Kubernetes ones:

```bash
$ ./kubectl-gadget run trace_open:latest
INFO[0000] Experimental features enabled
RUNTIME.CONTAINERNAME   PID          UID          GID          MNTNS_ID  ERR FD          COMM        FNAME                   TIMESTAMP
```

The ig command shows the right (runtime) ones:

```bash
$ sudo -E ig run trace_open:latest --pull always
INFO[0000] Experimental features enabled
WARN[0000] Ignoring runtime "cri-o" with non-existent socketPath "/run/crio/crio.sock"
WARN[0000] Ignoring runtime "podman" with non-existent socketPath "/run/podman/podman.sock"
RUNTIME.CONTAINERNAME   PID          UID          GID          MNTNS_ID  ERR FD          COMM        FNAME                   TIMESTAMP
``

PR #3323, opened 2024-08-08
datasource: Fix order for K8S.NODE

mqasimsarfraz:
The change fixes order for `K8S.NODE` column. 

### before the PR

```
$ ./kubectl gadget run trace_open
K8S.NAMESPACE           K8S.PODNAME             K8S.CONTAINERNAME       COMM                  PID          TID          UID          GID  FD FNAME                    MODE         K8S.NODE     ERROR     TIMESTAMP
```

### after the PR 

```
$ ./kubectl gadget run trace_open
K8S.NODE               K8S.NAMESPACE          K8S.PODNAME            K8S.CONTAINERNAME      COMM                 PID         TID         UID         GID  FD FNAME                    MODE        ERROR    TIMESTAMP
```


burak-ok (approved):
LGTM, thanks

pkg/datasource/compat/wrapper.go
```diff
@@ -134,7 +134,14 @@ func WrapAccessors(source datasource.DataSource, mntnsidAccessor datasource.Fiel
...
 	}
 
-	ev.nodeAccessor, err = k8s.AddSubField("node", api.Kind_String, datasource.WithTags("kubernetes"))
+	ev.nodeAccessor, err = k8s.AddSubField("node",
+		api.Kind_String,
+		datasource.WithTags("kubernetes"),
+		datasource.WithAnnotations(map[string]string{
+			datasource.TemplateAnnotation: "node",
+		}),
+		datasource.WithOrder(-31),
```
blanquicet, line 143: How was it working before?

mauriciovasquezbernal (approved):
LGTM. 

(We need to think how to fix the whole ordering thing, as we're using the order of the fields in the eBPF structure).

pkg/datasource/compat/wrapper.go
mqasimsarfraz: Which part? The `k8s.node` fields was already there, I am just defining an order so it comes to the front or am I missing something?

blanquicet: I mean, IIRC, `k8s.node` was printed by first in the past. So, what did break it that made necessary to specify the order for it?

mqasimsarfraz: IIUC it was like that since refactoring, initially `k8s` [columns were hidden ](https://github.com/inspektor-gadget/inspektor-gadget/issues/3014)so we didn't notice it? They were enabled with: https://github.com/inspektor-gadget/inspektor-gadget/pull/3136#issue-2392550863 and PR output confirms it was the case.

mqasimsarfraz:
Thanks for the reviews!

PR #3324, opened 2024-08-08
docs/op: Add UidGidResolver documentation

burak-ok:
# docs/op: Add UidGidResolver documentation

Is this the way we want to document the dataoperators?

### DIFF
--- a/docs/reference/operators/uidgidresolver.mdx
+++ b/docs/reference/operators/uidgidresolver.mdx
@@ -1,0 +1,11 @@
+---
+title: UidGidResolver
+sidebar_position: 10
+---
+
+# UidGidResolver
+
+The `UidGidResolver` data operator resolves `uid` and `gid` (user and group identifiers) values to the corresponding user and group names.
+It does this by monitoring the `/etc/passwd` and `/etc/group` files on the host.
+This leads to the restrictions that user and groupnames inside the guest might not be correctly resolved.
+It means the events are enriched with the user and group names from the host, so the ones for the container are ignored.
--- a/docs/gadget-devel/gadget-ebpf-api.md
+++ b/docs/gadget-devel/gadget-ebpf-api.md
@@ -285,6 +285,25 @@ Symbolize the kernel stack from `gadget_get_kernel_stack(ctx)` (see [kernel-stac
 
 - `ebpf.formatter.kstack`: Name of the new field. If the annotation is not set and the source field name has a `_raw` suffix, the target name will be set to the source name without that suffix.
 
+### `gadget_uid` and `gadget_gid`
+
+The `uid` and `gid` saved to these types will be resolved to the corresponding username and groupname on the host system:
+
+```c
+struct event {
+	gadget_uid user_raw;
+};
+```
+
+```json
+  "user": "root",
+  "user_raw": 0,
+```
+
+#### Annotations
+
+- `uidgidresolver.target`: Name of the new field. If the annotation is not set and the source field name has a `_raw` suffix, the target name will be set to the source name without that suffix.
+
 ### Enumerations
 
 Inspektor Gadget supports enums already defined on the kernel or enums defined


docs/reference/operators/uidgidresolver.mdx
```diff
@@ -0,0 +1,58 @@
...
+# UidGidResolver
+
+The `UidGidResolver` data operator resolves `uid` and `gid` values to the corresponding username and groupname.
+It does this by monitoring the `/etc/passwd` and `/etc/group` files on the host.
+This leads to the restrictions that user and groupnames inside the guest might not be correctly resolved.
+
+## Usage
+
+To use the `UidGidResolver` one has to use the `gadget_uid` and `gadget_gid` typedefs in the eBPF code and write the `uid` and `gid` values into it.
+Finally either of the following requirements has to be met:
```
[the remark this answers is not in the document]
burak-ok: Makes sense, thanks!

docs/reference/operators/uidgidresolver.mdx
```diff
@@ -0,0 +1,58 @@
...
+            my_gadget:
+                fields:
+                    uid:
+                        annotations:
+                            uidgidresolver.target: user
+                    gid:
+                        annotations:
+                            uidgidresolver.target: group
+        ```
+    </TabItem>
```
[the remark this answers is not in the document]
burak-ok: What do you mean with group everything?

Regarding tabs or two headers under each other. For me either of these options is fine. We should just discuss this now, since this differentiation `raw/target` is used in other operators aswell

mauriciovasquezbernal (commented):
My idea regarding https://www.inspektor-gadget.io/docs/latest/reference/operators/ is to provide information for the gadget users. The documentation on this PR is great, but I see most of them is focused in gadget developers. I think it could go to https://www.inspektor-gadget.io/docs/latest/gadget-devel/gadget-ebpf-api#enriched-types, WDYT?

burak-ok:
> My idea regarding https://www.inspektor-gadget.io/docs/latest/reference/operators/ is to provide information for the gadget users.

Would you say the that roughly the first paragraph would be targeted for gadget users? It gives information what to expect from the enriched fields from the operator?

docs/reference/operators/uidgidresolver.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: UidGidResolver
+sidebar_position: 10
+---
+
+# UidGidResolver
+
+The `UidGidResolver` data operator resolves `uid` and `gid` values to the corresponding username and groupname.
```
mauriciovasquezbernal, line 8:
```suggestion
The `UidGidResolver` data operator resolves `uid` and `gid` (user and group identifiers) values to the corresponding user and group names.
```

?

docs/reference/operators/uidgidresolver.mdx
```diff
@@ -0,0 +1,10 @@
...
+It does this by monitoring the `/etc/passwd` and `/etc/group` files on the host.
+This leads to the restrictions that user and groupnames inside the guest might not be correctly resolved.
```
mauriciovasquezbernal, line 10: What are those restrictions? When doesn't it work? I think it's important to document it.

mauriciovasquezbernal (approved):
> Would you say the that roughly the first paragraph would be targeted for gadget users? It gives information what to expect from the enriched fields from the operator?

Yes, I think that's good enough for them as it explains what's the information that is provided by the operator and briefly describes how it works. 

I added a comment for more information on the limitations, but I'd say this already LGTM!

docs/reference/operators/uidgidresolver.mdx
burak-ok: I added an example. WDYT?

mauriciovasquezbernal: Perhaps say that the event is always enriched with the user name in the host and that the container one isn't taken into consideration?

I think it can be something like:

> It does this by monitoring the `/etc/passwd` and `/etc/group` files on the host. It means the events are enriched with the user and group names from the host, so the ones for the container are ignored.

PR #3327, opened 2024-08-09
pkg/operators: Add limiter

blanquicet:
# Add limiter operator

This PR introduces the `Limiter` data operator which operates on array data sources by limiting the number of entries in each batch of data. A great scenario for this operator is when you are already sorting data within an array of data and you want to filter out the top `X` entries. 

This operator has lower priority than combiner (if it's enabled) and sorted (to limit after things are sorted).

Notice this functionality was known as `max-rows` in built-in gadgets. However, it was changed to `max-entries` as it also limit the number of entries in the JSON or any other kind of output.

Refer https://github.com/inspektor-gadget/inspektor-gadget/issues/1924

## TODOs for follow-up PRs

- [x] Limit also on server side to avoid wasting resources sending things that will be dropped on client.

### DIFF
--- a/docs/reference/operators/limiter.md
+++ b/docs/reference/operators/limiter.md
@@ -1,0 +1,21 @@
+---
+title: Limiter
+---
+
+The Limiter operator limits the number of entries in each batch of data. This
+operator is only enabled for data sources of type array. A great scenario for
+this operator is when you are already sorting data within an array of data and
+you want to filter out the top `max-entries` entries.
+
+## Instance Parameters
+
+### `--max-entries`
+
+The maximum number of entries for each batch of data. If using multiple array
+data sources, prefix the value with 'datasourcename:' and separate with ','. If
+no array data source is specified, the value will be applied to all array data
+sources. Use -1 to disable the limiter.
+
+Fully qualified name: `operators.limiter.max-entries`
+
+Default value: `-1`
--- a/pkg/datasource/datasource.go
+++ b/pkg/datasource/datasource.go
@@ -68,6 +68,12 @@ type DataArray interface {
 
 	// Swap swaps two elements of the array by their index
 	Swap(i, j int)
+
+	// Resize resizes the array to the given size. If the new size is smaller
+	// than the current size, the array will be truncated and the elements
+	// beyond the new size will be released. Currently, it does not support
+	// resizing to a larger size.
+	Resize(int) error
 }
 
 type Packet interface {
--- a/pkg/gadget-service/api-helpers/aphelpers_test.go
+++ b/pkg/gadget-service/api-helpers/aphelpers_test.go
@@ -1,0 +1,215 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package apihelpers provides some helper functions for the API package; these were extracted into this package
+// to avoid having additional dependencies on the API package itself
+package apihelpers
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetStringValuesPerDataSource(t *testing.T) {
+	tests := []struct {
+		name        string
+		input       string
+		expected    map[string]string
+		expectedErr bool
+	}{
+		{
+			name:     "empty input",
+			input:    "",
+			expected: map[string]string{},
+		},
+		{
+			name:     "valid without datasource",
+			input:    "50",
+			expected: map[string]string{"": "50"},
+		},
+		{
+			name:     "valid with datasource",
+			input:    "datasource1:10",
+			expected: map[string]string{"datasource1": "10"},
+		},
+		{
+			name:     "valid with multiple datasources",
+			input:    "datasource1:10,datasource2:20",
+			expected: map[string]string{"datasource1": "10", "datasource2": "20"},
+		},
+		{
+			name:     "valid with empty element",
+			input:    "datasource1:10,",
+			expected: map[string]string{"datasource1": "10"},
+		},
+		{
+			name:     "valid with empty string value",
+			input:    "datasource1:10,datasource2:",
+			expected: map[string]string{"datasource1": "10", "datasource2": ""},
+		},
+		{
+			name:     "valid with non-number value",
+			input:    "datasource1:foo",
+			expected: map[string]string{"datasource1": "foo"},
+		},
+		{
+			name:        "invalid mixing datasource and no datasource",
+			input:       "10,datasource1:20",
+			expectedErr: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, err := GetStringValuesPerDataSource(test.input)
+			if test.expectedErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+			require.Equal(t, test.expected, got)
+		})
+	}
+}
+
+func TestGetIntValuesPerDataSource(t *testing.T) {
+	tests := []struct {
+		name        string
+		input       string
+		expected    map[string]int
+		expectedErr bool
+	}{
+		{
+			name:     "empty input",
+			input:    "",
+			expected: map[string]int{},
+		},
+		{
+			name:     "valid without datasource",
+			input:    "50",
+			expected: map[string]int{"": 50},
+		},
+		{
+			name:     "valid with datasource",
+			input:    "datasource1:10",
+			expected: map[string]int{"datasource1": 10},
+		},
+		{
+			name:     "valid with multiple datasources",
+			input:    "datasource1:10,datasource2:20",
+			expected: map[string]int{"datasource1": 10, "datasource2": 20},
+		},
+		{
+			name:     "valid with empty element",
+			input:    "datasource1:10,",
+			expected: map[string]int{"datasource1": 10},
+		},
+		{
+			name:        "invalid with empty string value",
+			input:       "datasource1:10,datasource2:",
+			expectedErr: true,
+		},
+		{
+			name:        "invalid with invalid number",
+			input:       "datasource1:foo",
+			expectedErr: true,
+		},
+		{
+			name:        "invalid mixing datasource and no datasource",
+			input:       "10,datasource1:20",
+			expectedErr: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, err := GetIntValuesPerDataSource(test.input)
+			if test.expectedErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+			require.Equal(t, test.expected, got)
+		})
+	}
+}
+
+func TestGetDurationValuesPerDataSource(t *testing.T) {
+	tests := []struct {
+		name        string
+		input       string
+		expected    map[string]time.Duration
+		expectedErr bool
+	}{
+		{
+			name:     "empty input",
+			input:    "",
+			expected: map[string]time.Duration{},
+		},
+		{
+			name:     "valid without datasource",
+			input:    "50s",
+			expected: map[string]time.Duration{"": 50 * time.Second},
+		},
+		{
+			name:     "valid with datasource",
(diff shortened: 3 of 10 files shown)


pkg/operators/cutter/cutter.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (s *cutterOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error {
+	for ds, maxRows := range s.cutters {
+		ds.SubscribeArray(func(ds datasource.DataSource, data datasource.DataArray) error {
+			if data.Len() <= maxRows {
+				return nil
+			}
+
+			// TODO: Cut the data
```
blanquicet, line 130: TODO: Should `DataArray` implement a `Drop` method?

flyth:
How about naming this one "LimitOperator" or "LimiterOperator" instead? Coming from data handling things like SQL & Co. that feels more suited.

blanquicet:
> How about naming this one "LimitOperator" or "LimiterOperator" instead? Coming from data handling things like SQL & Co. that feels more suited.
>
Limiter is also fine for me.

pkg/operators/cutter/cutter.go
flyth: Maybe rather a `.Resize()` (as that's more generic) that can grow and shrink later on - for now, we only need to implement shrinking and returning an error if trying to grow. It needs to release the removed elements, then.

blanquicet: Makes sense as idea. Let me give it a try.

blanquicet:
This PR is now ready for review.

docs/reference/operators/limiter.mdx
```diff
@@ -0,0 +1,20 @@
...
+The `Limiter` data operator operates on array data sources by limiting the
+number of entries in each batch of data. A great scenario for this operator is
+when you are already sorting data within an array of data and you want to filter
+out the top `X` entries.
+
+## Flags
+
+### `--max-entries`
+
+The maximal number of entries for each batch of data. If using multiple data
```
flyth, line 17:
```suggestion
The maximum number of entries for each batch of data. If using multiple data
```

pkg/datasource/datasource.go
```diff
@@ -68,6 +68,11 @@ type DataArray interface {
 
 	// Swap swaps two elements of the array by their index
 	Swap(i, j int)
+
+	// Resize resizes the array to the given size. If the new size is smaller
+	// than the current size, the array will be truncated. Currently, it does
+	// not support resizing to a larger size.
```
flyth, line 74: Maybe add a note that truncated entries are also being released.

pkg/operators/limiter/limiter.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+const (
+	name            = "limiter"
+	ParamMaxEntries = "max-entries"
```
flyth, line 34: Maybe we can use `--limit` now as well? This would again move things more into an SQL direction that people might be more familiar with. But I don't have a super strong opinion on it.

pkg/operators/limiter/limiter.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+	}
+}
+
+func (l *limiterOperator) InstantiateDataOperator(gadgetCtx operators.GadgetContext, instanceParamValues api.ParamValues) (operators.DataOperatorInstance, error) {
+	for _, ds := range gadgetCtx.GetDataSources() {
+		if ds.Type() == datasource.TypeArray {
+			return &limiterOperatorInstance{
+				maxEntries: instanceParamValues[ParamMaxEntries],
+			}, nil
```
flyth, line 70: This creates a `limiterOperatorInstance` even if there's no param value set. Maybe add a check? (I know you're providing a default above, so it should be filled in from the clients using that value, but using the API or other ways, it might be empty which would lead to no returned value (as default would be 0, right?))

pkg/operators/limiter/limiter.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (l *limiterOperator) InstanceParams() api.Params {
+	return api.Params{
+		{
+			Key:   ParamMaxEntries,
+			Title: "Max Entries",
+			Description: "The maximal number of entries for each batch of data. " +
+				"If using multiple data sources, prefix the value with 'datasourcename:' and separate with ';'.",
+			DefaultValue: "50",
```
flyth, line 59: I'm not sure about having this default value - it might be rather surprising for gadget authors if they're not aware of this operator/functionality to have their data limited. I'd suggest keeping it empty (unlimited) or use `-1` explicitly for "unlimited".

pkg/operators/limiter/limiter.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (l *limiterOperatorInstance) Name() string {
+	return name
+}
+
+// parseMaxEntries parses the max-entries parameter and returns a map of data
+// source names to the maximum number of entries to keep. If no data source is
+// specified, the value is used for all data sources.
+func parseMaxEntries(maxEntries string) (map[string]int, error) {
```
flyth, line 91: We have some [helpers](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadget-service/api-helpers/apihelpers.go#L111) now. But we still need to discuss the syntax of those (outside this PR).

pkg/operators/limiter/limiter.go
```diff
@@ -0,0 +1,164 @@
...
+		if len(dsLimiters) > 1 {
+			return nil, fmt.Errorf("mixing %s rules with and without specifying data source", ParamMaxEntries)
+		}
+	}
+
+	return dsLimiters, nil
+}
+
+func (l *limiterOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error {
+	dsLimiters, err := parseMaxEntries(l.maxEntries)
```
flyth, line 122: Maybe move the validation / parsing to InstantiateDataOperator, so that errors can be caught early.

pkg/operators/limiter/limiter.go
```diff
@@ -0,0 +1,164 @@
...
+		var maxEntries int
+		if val, ok := dsLimiters[""]; ok {
+			maxEntries = val
+		} else {
+			maxEntries = dsLimiters[ds.Name()]
+		}
+
+		if ds.Type() != datasource.TypeArray {
+			return fmt.Errorf("%s can only be used on array data sources", ParamMaxEntries)
+		}
```
flyth, line 137: Currently, this would happen for gadgets that have both array as well as single entry datasources. Maybe just continue in that case? (We need to make sure to validate "max entries" so that none are set for single entry datasources)

pkg/operators/limiter/limiter_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+package limiter
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseMaxEntries(t *testing.T) {
```
flyth, line 23: Ah, those tests might be nice for the [helpers](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadget-service/api-helpers/apihelpers.go#L111) I mentioned 😄

flyth (commented):
Did a first pass, looking good already!

pkg/operators/limiter/limiter.go
blanquicet: I tried to parse `max-entries` in `InstantiateDataOperator` and instantiate it only if `max-entries` has a valid value, but I found an issue doing so. 

`InstantiateDataOperator` is also called during the `GetGadgetInfo` and, in that moment, the `instanceParamValues` are not yet available. So, we'd wrongly conclude "we don't need the limiter, so don't instantiate it", resulting in not including the operator's params in the `GadgetInfo`.

Sometimes I think that the instantiation for `GetGadgetInfo` should be different than the one for `Run`.

pkg/operators/limiter/limiter.go
blanquicet: Let's see what others think about it.

pkg/operators/limiter/limiter.go
blanquicet: I also thought about not limiting but I wanted to initially keep the same behaviour we had in built-in gadgets. Let's see what other think also here.

pkg/operators/limiter/limiter.go
blanquicet: Let's continue the discussion in https://github.com/inspektor-gadget/inspektor-gadget/pull/3327#discussion_r1726835433

pkg/operators/limiter/limiter.go
blanquicet: I was thinking that we need that helper. Great! I'll use it and move the tests there.

pkg/operators/limiter/limiter.go
blanquicet: I added one more check to cover single entry datasources:

```go
	for _, ds := range gadgetCtx.GetDataSources() {
		var maxEntries int
		if val, ok := dsLimiters[""]; ok {
			maxEntries = val
		} else {
			if val, ok := dsLimiters[ds.Name()]; ok {
				maxEntries = val
			} else {
				continue
			}
		}

		if ds.Type() != datasource.TypeArray {
			return fmt.Errorf("%s can only be used on array data sources", ParamMaxEntries)
		}

		...
	}
```

WDYT?

flyth: Yeah, that should do it.

pkg/operators/limiter/limiter.go
flyth: Ah, right - that happened for other operators as well, already. We could still check for errors and in any case return at least an empty operator if there's an array ds to make sure params are transported.

> Sometimes I think that the instantiation for GetGadgetInfo should be different than the one for Run.

Would make it even more complicated IMHO. But we should build a list of things to do / not to do in initialization & so on :)

blanquicet: > We could still check for errors and in any case return at least an empty operator if there's an array ds to make sure params are transported.
>
I followed this suggestion to detect wrong flag values earlier.

pkg/operators/limiter/limiter.go
blanquicet: I set no limit by default.

blanquicet:
I managed all the comments and it's ready for a second round of reviews.

pkg/gadget-service/api-helpers/aphelpers_test.go
```diff
@@ -0,0 +1,215 @@
...
+			input:    "datasource1:10",
+			expected: map[string]string{"datasource1": "10"},
+		},
+		{
+			name:     "valid with multiple datasources",
+			input:    "datasource1:10,datasource2:20",
+			expected: map[string]string{"datasource1": "10", "datasource2": "20"},
+		},
+		{
+			name:     "valid with empty element",
```
flyth, line 54: I'm not sure whether this should really be valid - but we can discuss on it later.

pkg/operators/limiter/limiter.go
flyth: /cc @mauriciovasquezbernal @alban

flyth (approved):
Looks great from code inspection. Thanks for the additional tests!

pkg/operators/limiter/limiter.go
mauriciovasquezbernal: I feel "max-entries" is easier to understand, but not a hard opinions as well.

blanquicet:
Rebased on top of main and updated the documentation following https://github.com/inspektor-gadget/inspektor-gadget/pull/3406/files approach.

docs/reference/operators/limiter.md
```diff
@@ -0,0 +1,21 @@
...
+### `--max-entries`
+
+The maximum number of entries for each batch of data. If using multiple array
+data sources, prefix the value with 'datasourcename:' and separate with ','. If
+no array data source is specified, the value will be applied to all array data
+sources. Use -1 to disable the limiter.
+
+Fully qualified name: `operators.limiter.max-entries`
+
+Default value: "-1"
```
flyth, line 21:
```suggestion
Default value: `-1`
```

docs/reference/operators/limiter.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Limiter
+---
+
+The Limiter operator limits the number of entries in each batch of data. This
+operator is only enabled for data sources of type array. A great scenario for
+this operator is when you are already sorting data within an array of data and
+you want to filter out the top `X` entries.
```
flyth, line 8:
```suggestion
you want to filter out the top `max-entries` entries.
```
Maybe?

flyth (approved):
One nit, one suggestion regarding the updated docs 😄

PR #3329, opened 2024-08-10
formatters: enrich proto field to include protocol name in L4 endpoints

### LINKED ISSUE #3233: formatters: Enrich proto field too
The `gadget_l4endpoint_t` type is enriched with the human readable version of the IP address, however the proto is still printed in its numeric value:

```
$ sudo -E ig run trace_tcpconnect:latest --verify-image=false -o json | jq
{
  "dst": {
    "addr": "1.1.1.1",
    "port": 80,
    "proto": 6,
    "version": 4
  }, 
  "src": {
    "addr": "172.17.0.2",
    "port": 38566,
    "proto": 6,
    "version": 4
  },
}
```

We should provide a human readable version of the proto according to https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml. 

Ref #3076

Snehil-Shah:
# Enrich proto field in `gadet_l4endpoint_t`

Resolves #3233.

- Changes existing `proto` field to `proto_raw`.
- Adds a new enriched subfield `proto` which is the protocol name in string format.

```
  "dst": {
    "addr": "1.1.1.1",
    "port": 80,
    "proto_raw": 6,
    "proto": "TCP",
    "version": 4
  }
```

## How to use

- Run any gadget using the `proto_raw` field like `trace_tcpconnect`.

## Testing done

Ran the above steps

## Questions

I gave a generic one-word description to protocols with no keyword. For instance, protocol 61 (according to [iana](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml)) denotes any host internal protocol. So it just prints `"internal"` as the protocol. Is this behavior welcome?


pkg/utils/protocols/protocols.go
```diff
@@ -0,0 +1,174 @@
...
+	145: "NSH",
+	253: "experimentation",
+	254: "experimentation",
+	255: "reserved",
+}
+
+func GetProtocolNameByNumber(number int) (string, bool) {
+	name, ok := protoNumberToName[number]
+
+	return name, ok
```
mauriciovasquezbernal, line 173:
```suggestion
	return protoNumberToName[number]
```

pkg/operators/formatters/formatters.go
```diff
@@ -412,6 +427,14 @@ var replacers = []replacer{
 
 				port, _ := ports[0].Uint16(entry)
 				endpointF.PutString(entry, fmt.Sprintf("%s:%d", addrStr, port))
+
+				protoNumber, _ := protos[0].Uint16(entry)
+				protoName, exist := protocols.GetProtocolNameByNumber(int(protoNumber))
+				if !exist {
+					protoName = "unassigned"
```
mauriciovasquezbernal, line 434:
```suggestion
					protoName = fmt.Sprintf("proto#%u", protoNumber)
```

pkg/operators/formatters/formatters.go
```diff
@@ -392,6 +393,20 @@ var replacers = []replacer{
...
+			if protos[0].Size() != 2 {
+				return nil, fmt.Errorf("proto size expected to be 2 bytes")
+			}
+
+			protoFieldName := strings.TrimSuffix(protos[0].Name(), "_raw")
+			protoField, err := in.AddSubField(protoFieldName, api.Kind_String)
+			if err != nil {
+				return nil, err
+			}
+
```
mauriciovasquezbernal, line 409: After #2997 we need to be more conservative. This code should check if there's a proto_raw field, if there is not, it should continue to work as before (perhaps printing a warning), if the proto_raw field is present, it should enrich it. 

(this is important to be able to run gadgets that were compiled with an old version of the headers)

mauriciovasquezbernal (changes requested):
One blocking comment about being able to run gadgets that don't have the proto_raw field.

mauriciovasquezbernal:
@Snehil-Shah  do you have any plans to complete it?

Snehil-Shah:
@mauriciovasquezbernal Hey, sorry I got stuck with some stuff. I will get back to it after a few days if that's okay..?

mauriciovasquezbernal:
> @mauriciovasquezbernal Hey, sorry I got stuck with some stuff. I will get back to it after a few days if that's okay..?

For sure, I only wanted to ensure you don't forget about it.

pkg/utils/protocols/protocols.go
Snehil-Shah: But it's possible the protocol number might not have an enriched name in the map. Isn't it better to explicitly handle that?

mauriciovasquezbernal: What I meant is to avoid declaring the temporal variables, the result should be the same. Unless I'm missing something.

mauriciovasquezbernal (commented):
I trigger the CI. However I think you'll have to update the integration tests:

The proto field

https://github.com/inspektor-gadget/inspektor-gadget/blob/5944bcc9f5eda9ec0b821e36c05e1711c0e3071e/pkg/testing/utils/types.go#L23

is now a string

mauriciovasquezbernal:
> So currently, it just prints "internal" as the protocol. Is this behavior welcome?

I think that's fine.

Snehil-Shah:
> What I meant is to avoid declaring the temporal variables, the result should be the same. Unless I'm missing something.

Ahh right, makes sense.

pkg/utils/protocols/protocols.go
Snehil-Shah: I get the following compiler error

```
have (string)
want (string, bool)
```

I guess we need to be explicit here with the return types, currently returning it directly either returns the found value or an empty string, not an acknowledgment variable with it.

mauriciovasquezbernal: TIL: Thanks for trying it out!

pkg/testing/utils/types.go
```diff
@@ -20,7 +20,7 @@ type L4Endpoint struct {
 	Addr    string `json:"addr"`
 	Version uint8  `json:"version"`
 	Port    uint16 `json:"port"`
-	Proto   uint16 `json:"proto"`
```
mauriciovasquezbernal, line 23: Please also update the different tests with udp/tcp accordingly.

mauriciovasquezbernal:
We need the last commit of #3572 in order to fix the tests on this PR. The issue is that the unit test job is trying to run the image-based gadgets unit tests but using the `latest` version of those gadgets, hence the test fails as `latest` is using a number instead of a string for the proto.

mauriciovasquezbernal:
@Snehil-Shah please rebase this on top of main. It should make tests pass.

Snehil-Shah:
@mauriciovasquezbernal Done!

mauriciovasquezbernal (approved):
Tests are now passing. Thanks a lot!

Issue #3304, opened 2024-08-06
[EPIC]: Package Inspektor Gadget


mauriciovasquezbernal:
Inspektor Gadget provides three binaries (ig, gadgetctl and kubectl-gadget) that are available to download for different architectures and operating systems on the release page. This approach is not very user friendly, as they need to download the binary manually to get the updates, etc. 

The goal of of this issue is to create packages for:
- [ ] Windows 
   - [ ] Winget
   - and/or 
   - [ ] Chocolatey
- [ ] MacOS
   - [ ] Brew
- [ ] Linux
   - [ ] Ubuntu
   - [ ] Fedora
   - [ ] What else?

### Discussion

We can package the binaries in different ways

- option 1: Put all binaries in the single inspektor-gadget package
- option2: Put the ig + gadgetctl binaries in on package, and put kubectl-gadget in another 

I'd say option is easier for us and for the users.

(ig only will be package for Linux as it doesn't support Windows or MacOS)

rd-michel:
can you share any updates?

ive also downloaded gadgetctl recently (manually) ... and ive asked myself "where is the corresponding brew package for this tool :) "

mauriciovasquezbernal:
> can you share any updates?

We've packaged it for Fedora https://repology.org/project/inspektor-gadget/versions, other package managers are on the roadmap but we don't have an estimate time for them.

PR #3336, opened 2024-08-12
GHA: create post-release PR to update README.md

alban:
See discussion in #3300.


mauriciovasquezbernal:
I second Francis here. I think it creates extra complexity we don't need. We can link to https://github.com/inspektor-gadget/inspektor-gadget/releases/latest/ and say something like "download the gadgetctl binary for your operating system..". In any case, this will be improved once we have #3304.

PR #3338, opened 2024-08-12
snapshot_process: Sleep before starting the gadget

burak-ok:
This makes sure that our test container has already spawned and is running the command we want to see in the snapshot

We have many flaky CI failures on `snapshot_process`. This should fix it

### DIFF
--- a/gadgets/snapshot_process/test/snapshot_process_test.go
+++ b/gadgets/snapshot_process/test/snapshot_process_test.go
@@ -17,6 +17,7 @@ package tests
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
Thanks for the fix!

mauriciovasquezbernal:
Unfortunately I don't think it fixes the issue:
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10386743239/attempts/1#summary-28759182888
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10380957419/attempts/1#summary-28742131702
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10393961200/attempts/1#summary-28783086720

PR #3345, opened 2024-08-13
docs/legacy: Update the sort_by list

burak-ok:
Two new fields were added in https://github.com/inspektor-gadget/inspektor-gadget/pull/3319. They have to be update in the legacy docs aswell, to satisfy the CI


alban (approved):
Please add DCO. LGTM after that.

PR #3346, opened 2024-08-13
Use JSON args to prevent unintended behavior related to OS signals

burak-ok:
We had 2 warnings shown inside Github regarding this:
image


mauriciovasquezbernal (approved):
Nice! Thanks for getting rid of those annoying warnings.

Issue #3357, opened 2024-08-14
gadgets' READMEs are out of sync with gadgets/README.template


alban:
## Description

We have a Makefile rule to automatically update gadgets' READMEs with the list of flags defined in gadget.yaml.
```
make -C gadgets trace_exec/README.md
```

But commit c1acdfeed6936a957f2e4e4eef153da677d0297f changed the documentation without using the template, so it is out of sync.

cc @mauriciovasquezbernal @sniok

## Impact

We can no longer use `make` to update the documentation when flags change.

PR #3358, opened 2024-08-14
gadgets/readmes: update all gadgets documentation

alban:
# gadgets/readmes: update all gadgets documentation

The templates are updated. When a new ebpf param is added in a gadget, the readmes can be updated **in place** with:
```
make -C gadgets readmes
```

4 gadgets didn't have a README.mdx (lsm, malloc, signal, ssl). This is now fixed.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/3357

## How to use

```
make -C gadgets readmes
```

## Testing done

None


gadgets/README.template
```diff
@@ -1,27 +1,38 @@
...
-### `--{{ index $flag "key" }}`
-{{ index $flag "description" }}
-
-Default value: "{{ index $flag "defaultValue" }}"
-{{ end -}}
-{{- else -}}
-No flags.
-{{- end -}}
+
+TODO
```
mauriciovasquezbernal, line 34: Can't it be generated too?

gadgets/trace_bind/gadget.yaml
```diff
@@ -56,9 +56,9 @@ datasources:
           columns.width: 10
 ebpfParams:
   ignore_errors:
-    key: ignore_errors
+    key: ignore-errors
     alias: i
-    defaultValue: ""
+    defaultValue: "true"
```
mauriciovasquezbernal, line 61: note: It's actually not needed anymore as we're getting these from the ebpf object. (I'll open an issue / PR to clean it up).

gadgets/snapshot_socket/README.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: snapshot_socket
-sidebar_position: 35
+sidebar_position: 40
```
mauriciovasquezbernal, line 3: The idea was to use multiples of 10 for the existing gadgets and then start filling the spaces with the new ones. But I suppose this approach of changing the position for all of them works as well. I'm wondering if the script you included in the commit message should be a makefile target.

mauriciovasquezbernal (approved):
LGTM from a quick inspection. I'm wondering if the scripts you used to update the files (change the index and the digest) should be part of the makefile.

gadgets/README.template
alban: The flags section are generated by `README-flags.template` and it does in-place update.


`README.template` is only used when the gadget's `README.mdx` does not exist yet.

alban:
> I'm wondering if the scripts you used to update the files (change the index and the digest) should be part of the makefile.

I am adding the script for renumbering in the Makefile.

But I don't think the digest update can be done in the Makefile anymore: the templates are updating README.mdx and not README.md anymore. And artifacthub-pkg.yml is using README.md, which is no longer automatically generated. So the Makefile does not have a way to detect when README.md is modified. In the script, I specified the list of gadgets manually because of this:
```
DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
for i in trace_lsm trace_malloc trace_signal trace_ssl trace_tcpconnect  ; do sed -i "s/^digest:.*$/digest: $DATE/g" gadgets/$i/artifacthub-pkg.yml ; done
```

mauriciovasquezbernal:
There was still a missing piece: #3364

PR #3359, opened 2024-08-14
docs: Use GitHub instead of elixir

mauriciovasquezbernal:
The broken link job fails often with elixir:

https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10391104393/job/28773219330


alban (approved):

mauriciovasquezbernal:
The job is failing because there is a docker broken link, I'll check it later on.

PR #3370, opened 2024-08-16
ig: Add optional kubeconfig parameter

burak-ok:
ig: Add optional kubeconfig parameter


blanquicet:
This PR is needed because running `ig` with the `enrich-with-k8s-apiserver` flag in an AKS node is failing.

PR #3378, opened 2024-08-20
pkg/datasource: Hide enriched timestamp column by default

blanquicet:
# Hide enriched timestamp column by default

Hide the timestamp columns as it takes a lot of space and clutter the screen.


pkg/datasource/columns.go
```diff
@@ -222,6 +222,8 @@ var annotationsTemplates = map[string]map[string]string{
 		ColumnsWidthAnnotation:    "35",
 		ColumnsMaxWidthAnnotation: "35",
 		ColumnsEllipsisAnnotation: "end",
+		ColumnsHiddenAnnotation:   "true",
```
burak-ok, line 225: I would prefer to add this annotation in the `formatters.go` where the timestamp field is created

alban: Can you please clarify what difference does it make? The purpose of this patch is to hide timestamp.

burak-ok: When a gadget developer uses the template in the `gadget.yaml` for his field he wants and gets the correct formatting of the timestamp field itself.

1. `ColumnsHiddenAnnotation:   "true",`
    His field gets hidden by default even if he didn't use the `gadget_timestamp` type

2. Doing it in `formatters.go`
   The field doesn't get hidden if he doesn't use `gadget_timestamp`. If he uses the type, then his field gets hidden.

After a little bit of further thinking I don't know if the use case in 2) makes sense

mauriciovasquezbernal: I think it's fine to add it in the template, it'll allow to keep all field configuration on the same place, i.e using "template: timestamp" from the metadata of from the API when adding the field will have the same result. The user can always explicit mark the field as non hidden by using "columns.hidden: false" in the metadata.

flyth (approved):
LGTM from inspection.

PR #3389, opened 2024-08-22
charts: Fix config map volume

mauriciovasquezbernal:
There was an indentation problem and it wasn't possible to apply the manifest directly with kubectl:

```bash
$ kubectl apply -f pkg/resources/manifests/deploy.yaml
namespace/gadget unchanged
serviceaccount/gadget unchanged
configmap/gadget unchanged
clusterrole.rbac.authorization.k8s.io/gadget-cluster-role unchanged
clusterrolebinding.rbac.authorization.k8s.io/gadget-cluster-role-binding unchanged
role.rbac.authorization.k8s.io/gadget-role unchanged
rolebinding.rbac.authorization.k8s.io/gadget-role-binding unchanged
Error from server (BadRequest): error when creating "pkg/resources/manifests/deploy.yaml": DaemonSet in version "v1" cannot be handled as a DaemonSet: strict decoding error: unknown field "spec.template.spec.volumes[11].defaultMode", unknown field "spec.template.spec.volumes[11].items"
```


charts/gadget/templates/daemonset.yaml
```diff
@@ -279,10 +279,10 @@ spec:
...
-        - configMap:
+        - name: config
+          configMap:
             name: {{ include "gadget.fullname" . }}
-          defaultMode: 400
-          items:
-            - key: config.yaml
-              path: config.yaml
-          name: config
+            defaultMode: 400
```
rata, line 285: We were talking privately about this, so I got curious and had a look at the PR. This seems it might not be what you intended.

I understand this was not added in this PR, but here: https://github.com/inspektor-gadget/inspektor-gadget/commit/b2cacd70a7105b20a887da9b8972bbfcb4f4824d

However, 400 is a decimal number and therefore, when used in octal for the permissions, it is: 620: rw for owner, w for the group. The config probably wants a 400 in octal (read for the owner only).

It sucks, I know :(. But yaml didn't have any way to enforce a number to be in octal, or interpret it in octal by default. Or at least I didn't find any reasonable way to do it in k8s.

It seems you want 400 in octal, not decimal. JSON doesn't support octal (or didn't 8 years ago when I wrote the defaultMode for k8s), but it seems yaml does support it (it was not common back then either, IIRC). I think here you want `0o400` (or 256 in decimal), IOW read only for the owner.

Am I missing something?

charts/gadget/templates/daemonset.yaml
```diff
@@ -279,10 +279,10 @@ spec:
...
+            items:
+              - key: config.yaml
+                path: config.yaml
```
rata, line 288: Having a quick look at the configmap in git, it seems it only has this file. The idea of the item fields is to only mount in the pod some files/keys, and you use the items to specify which ones. But if you want all, no need to add this.

rata (commented):
@mauriciovasquezbernal I've got curious about the PR you shared with me over chat and had a look :-)

I think I've found some things are not as intended (k8s is confusing here, but AFAIK we can't do any better in k8s :-/). Let me know if it makes sense or maybe not

mauriciovasquezbernal:
@rata thansk for the review, I think your comments make sense, I updated the PR. However I'd like to have a review from @mqasimsarfraz on this.

rata:
@mauriciovasquezbernal please do test that the octal notation works as expected, I haven't used that in a while, I might miss-remember :-)

charts/gadget/templates/daemonset.yaml
mqasimsarfraz: > I think here you want 0o400 (or 256 in decimal), IOW read only for the owner.

Yes, we wanted to use `400` octal :see_no_evil:. I think we can use `0400` to use the octal value (or the decimal value you mentioned) https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#configmapvolumesource-v1-core:

> defaultMode is optional: mode bits used to set permissions on created files by default. Must be an octal value between 0000 and 0777 or a decimal value between 0 and 511. YAML accepts both octal and decimal values, JSON requires decimal values for mode bits. Defaults to 0644. Directories within the path are not affected by this setting. This might be in conflict with other options that affect the file mode, like fsGroup, and the result can be other mode bits set.

rata: @mqasimsarfraz The only truth is reality. Let me know if you test it what happens :)

I think you need `0o400` for the yaml parser to understand this number should be read in octal instead of decimal. But let's see... :)

mqasimsarfraz:
```
→ kubectl get pods -n gadget -o yaml | grep configMap: -A 5
    - configMap:
        defaultMode: 256
        name: gadget
      name: config
    - name: kube-api-access-m56kh
      projected:
...

```

Looking better!

mqasimsarfraz (approved):
Thanks for the fix! I tested the changes and they work as expected!

Issue #3310, opened 2024-08-06
docs: Document different APIS


mauriciovasquezbernal:
The apis (https://www.inspektor-gadget.io/docs/latest/category/apis) aren't documented yet 

- [ ] [Golang API](https://www.inspektor-gadget.io/docs/latest/api/golang)
  - [ ] Considering moving the readmes from [examples](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/examples/gadgets/datasource/mutate) to be available on the website
- [ ] [gRPC API](https://www.inspektor-gadget.io/docs/latest/api/grpc)

836hardik-agrawal:
hey @mauriciovasquezbernal  ,gadgets directory contain 4 sub directory and a readme file. That readme file contain information about each subdirectory.So we have to add that readme on website and link for their sub directory ,or all readme files

andoriyaprashant:
Hi @mauriciovasquezbernal 
I would like to work on this. Please assign this issue to me.
Thank you

836hardik-agrawal:
Hi @mauriciovasquezbernal  I have raised a PR for updating the changes regarding gRPC API . PR :https://github.com/inspektor-gadget/inspektor-gadget/pull/3405 .
For the Golang API I have little doubt which i asked on Slack . As soon as they get cleared I will raise for it also,

blanquicet:
Regarding gRPC API:

- **api.proto**: @836hardik-agrawal is working on this (https://github.com/inspektor-gadget/inspektor-gadget/pull/3405) and, following our suggestions, he is dividing it into two big sections (so that we can easily drop legacy one once we deprecate it):
  - gRPC API
  - Legacy gRPC API for Built-In Gadgets
- gadgettracermanager.proto: Where do we want to put it? See [this](https://github.com/inspektor-gadget/inspektor-gadget/pull/3405#discussion_r1734463203) discussion.

Issue #2100, opened 2023-10-03
Rename and consolidate binaries and container images


mauriciovasquezbernal:
TL;DR: We have too many binaries and container images, it's difficult to understand for the users and to handle for us.

Inspektor Gadget currently provides two main tools:

- Inspektor Gadget on Kubernetes, a.k.a [gadgettracermanager](https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.21.0/gadget-container/gadgettracermanager/main.go): A tool to trace and inspect applications running in Kubernetes. It was the first tool we created as the initial scope of the project was only about Kubernetes.
- [`ig`](https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.21.0/docs/ig.md): A tool to debug processes and applications on the local host. 

Each of these tools have a client part, [kubectl-gadget](https://github.com/inspektor-gadget/inspektor-gadget/tree/v0.21.0/cmd/kubectl-gadget) and [gadgetctl](https://github.com/inspektor-gadget/inspektor-gadget/tree/v0.21.0/cmd/gadgetctl) and there is a container image published for each of them [inspektor-gadget](https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/inspektor-gadget) & [ig](https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/ig).

The gadgettracermanager and ig binaries were too different at the beginning, but after much refactoring these two two binaries are becoming more and more similar on its implementation and use cases and we're considering adding features present in one to the other, for instance, #737, #1861 , #1862. 

The main differences between the two binaries right now are:
- gadgettracermanager has some kubernetes specific logic:
  - contacts the API server to 
    - get notifications about pod creation
    - list pods and services
    - implement the tracer controller 
  - use specific operators for Kubernetes (KubeIPResolver)

Regarding the container images, the inspektor-gadget one has some additional things like:
- entrypoint.sh and cleanup.sh
- hooks 

The differences are not that big, so the proposal of this issue is to consolidate them in a single binary and container image. In order to enable the Kubernetes specific features, we could have a `gadget mode` env variable or command line argument. One important thing is to be sure the UX is clean, for instance a user using `ig` on a local node shouldn't have all Kubernetes related flags available, unless explicitly requested. 

Regarding the naming, it should be called `ig` and the container image [inspektor-gadget](https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/inspektor-gadget) .

[...]

PR #3406, opened 2024-08-27
docs: add operator documentation

### LINKED ISSUE #3305: docs: Document Operators
We need to create the documentation for operators in https://www.inspektor-gadget.io/docs/latest/reference/operators/. 

The documentation should introduce the operator concept and explain the differences between image and data operators. Then, for each operator it should cover:
- [ ] Description (what does it handle? What information does it add?)
- [ ] Options (things that can be configured)

This should cover all available operators:

### Data Operators

- [ ] Kubemanager
- [ ] Localmanager
- [ ] CLI 
- [ ] Combiner 
- [ ] Filter 
- [ ] Formatters
- [ ] Kubeipresolver
- [ ] Otel-metrics
- [ ] Prometheus
- [ ] Simple?
- [ ] SocketEnricher
- [ ] Sort
- [x] [UidGidResolver](https://github.com/inspektor-gadget/inspektor-gadget/pull/3324)

### Image Operators
- [ ] Oci handler?
- [ ] bpf
- [ ] wasm
- [ ] btfgen

mauriciovasquezbernal:
Fixes #3305

## Future Work

- [ ] Document when operators are enabled/disabled
- [ ] Add annotations 
- [ ] Document operator implementation 
- [ ] Document where operators are used: client, server, both.
- [x] Add priority
- [ ] Check parameter descriptions


docs/reference/operators/otel-metrics.md
mauriciovasquezbernal, line 1: @flyth any chance you can provide a quick document for this and for the prometheus one?

flyth: I'm not sure whether we should add the `prometheus` one in here. It's only valid for the legacy gadgets and might confuse more than it helps.

flyth:
Thanks for working on this!

IMHO we should differentiate between global & instance parameters in the docs as well. Maybe we should even add the fully qualified name of the instance parameters so that they're easy to copy/paste - like `operators.sort.sort`.

docs/reference/operators/socketenricher.md
```diff
@@ -0,0 +1,14 @@
+---
+title: SocketEnricher
+---
+
+The Socket enricher operator provides process information for sockets. Gadgets
```
blanquicet, line 5:
```suggestion
The Socket enricher operator provides process information of the socket owner. Gadgets
```

docs/reference/operators/index.mdx
```diff
@@ -6,7 +6,32 @@ description: Documentation for Operators
...
+operator is a component that performs a specific action, for example, loading
+eBPF programs, sorting the output, etc. Operators are modular and they all share
+the same interface. Currently all operators are shipped within Inspektor Gadget,
+but we plan to support user-provided operators in the future.
+
+Operators are configured in a pipeline, the data is generated on the layer
+operators and then passed through the differnt data operators until it's
+exported by one of them.
+
+### Layer Operators
```
blanquicet, line 19: The code uses "image" and here we are talking about "layer" operators.

docs/reference/operators/ebpf.md
```diff
@@ -0,0 +1,34 @@
+---
+title: ebpf
+sidebar_position: 10
+---
+
+The eBPF operator handles layer of type
+`application/vnd.gadget.ebpf.program.v1+binary`, i.e. layers that contain an
+ELF file with the eBPF programs. This operator loads the programs into the
+kernel and attaches them to the different hooks as specified by the gadget
+developer.
```
blanquicet, line 10: What about formatters: enum, bitfield, kstack?

IMO, adding a TODO is fine for this PR. It's impossible to cover everything in one single PR.

blanquicet (commented):
Awesome! Thanks for working on this!

docs/reference/operators/otel-metrics.md
mauriciovasquezbernal: Good point. This should only cover operator for image-based gadgets.

docs/reference/operators/index.mdx
```diff
@@ -6,7 +6,32 @@ description: Documentation for Operators
...
-- parameters they take
+Operators are a very important part of the Inspetkor Gadget architecture. An
+operator is a component that performs a specific action, for example, loading
+eBPF programs, sorting the output, etc. Operators are modular and they all share
+the same interface. Currently all operators are shipped within Inspektor Gadget,
+but we plan to support user-provided operators in the future.
+
+Operators are configured in a pipeline, the data is generated on the layer
+operators and then passed through the differnt data operators until it's
+exported by one of them.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I wanted to create a mental mode where there is a source (layer operators) and a sink (last data operator). But I will remove it for now.

docs/reference/operators/index.mdx
```diff
@@ -6,7 +6,32 @@ description: Documentation for Operators
...
+
+### Layer Operators
+
+Layer operators handle the different layers that compone a Gadget. These
+operators access the information contained on the layer and perform the actions
+needed for the gadget to work, like loading the eBPF programs, running the WASM
+module, etc.
+
+These operators produce data that is then processed by data operators.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > I'm not sure about this distinction you're doing. For instance, the wasm operator is an image layer but it doesn't produce any data, it just processes the data generated by the ebpf operator. It goes in conflict with what you're saying, isn't it?

You're right. I was thinking only on the ebpf as an example of the layer operator. Currently wasm can process and produce data as well and operators like btfgen don't produce any data. I'll change the description to be more generic.

docs/reference/operators/formatters.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Formatters
+---
+
+The Formatters operator provides a human readable representation of things like
+IP addresses, signal names, error names, etc. Gadgets provide a "machine
+readable" version and then the operator adds the human readable counterpart:
+
+## Timestamp
```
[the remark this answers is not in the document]
mauriciovasquezbernal: This reference documentation is user focused. I added those examples to show the information the formatter operator adds. I don't think we should be adding types and annotations here as they're already documented in the gadget developer part. Perhaps we can split that gadget developer to be "per-operator", but that's another discussion.

docs/reference/operators/kubeipresolver.md
```diff
@@ -0,0 +1,26 @@
+---
+title: KubeIPResolver
+---
+
+The KubeIPResolver operator enriches IP addresses with pod and service
+information. For instance, in this example the `10.96.0.10` IP is enriched with
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's implicit the pod and service information is only added to IPs from pods and services.

docs/reference/operators/kubemanager.md
```diff
@@ -0,0 +1,52 @@
+---
+title: KubeManager
+---
+
+The KubeManager operator is used when Inspektor Gadget runs on Kubernetes. This
+operator keeps track of the pods running on the node by talking to the
+kube-apiserver and to the different container runtimes. This operator uses the
+pod informer and/or fanotify mechanisms to detect when containers are created.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Do we need to be that specific in this document? I added "and/or" to keep it generic. We can go deeper on this, but it's future work :tm: and PRs are welcome :stuck_out_tongue:

docs/reference/operators/kubemanager.md
```diff
@@ -0,0 +1,52 @@
...
+This information is used to:
+- Update the eBPF maps used to filter events in eBPF by container name.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Right, copied and pasted from the local manager :D

docs/reference/operators/btfgen.md
```diff
@@ -0,0 +1,16 @@
+---
+title: btfgen
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'm not sure we want to be that explicit again. I wanted to explain layer vs data operators at the beginning to provide more context, but after that it gets complicated.
- wasm: Layer operator because handles a layer, but it's also a data operator because handles data
- bpf: pretty much the same, it also enriches some information 

I'd prefer to omit it by now. In the end, an operator being layer or data doesn't change the user experience.

docs/reference/operators/index.mdx
flyth: No strong preference - I've used both over the course of the refactorings 😄 And both have pros and cons, IMHO.

mauriciovasquezbernal:
> Maybe we should even add the fully qualified name of the instance parameters so that they're easy to copy/paste - like operators.sort.sort.

Added them. I hope I didn't messed around with some of them.

mauriciovasquezbernal:
> Not right now, but I think we should add that as an option. 

ok, let's document that when we implement this support.

> Having the complete operator doc together on one page is another advantage of that, then 😄

That's a good discussion. I think combining all the documentation on the same page is good for maintaining purposes and for avoiding duplication, however I think it'll become too complex for folks that only want to understand what an operator does and don't care about its implementation. In any case, I'd like not to block this PR on that discussion.

docs/reference/operators/kubemanager.md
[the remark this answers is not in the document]
blanquicet: Hahaha I wasn't suggested to specify it in this document. I was just highlighting the fact the lack of such a documentation.

docs/reference/operators/index.mdx
mauriciovasquezbernal: Let's use "layer" here then.

blanquicet (commented):
> Can the gadget users change this when running a gadget? If not, I think annotations should go to gadget developer specific docs.
>

I don't agree. IMO, the gadget developer documents should introduce the developer to the possibilities they have, maybe mention some examples, but then point them to a more detailed documentation they can visit if they think they need it. Following an example:

In the [Metadata file documentation](https://inspektor-gadget.io/docs/latest/gadget-devel/metadata), we state developers can use annotations to control different aspects of operators for a given datasource. I don't expect we put [here](https://inspektor-gadget.io/docs/latest/gadget-devel/metadata#datasource) a collection of all the possible annotations from all the operators. Instead, as we are currently doing, we point developers to operators' documentation. But then, at this point, we must provide the whole list of annotations per operators in the operators' documentation, don't you think?

Otherwise, how can developers know that they can use the `cli.clear-screen-before` annotation to clean up the screen right before the data is printed in the terminal? And that's an annotation we are already using in `top_*` gadgets: [top_file](https://github.com/inspektor-gadget/inspektor-gadget/blob/9a797dc046f8bc1f45e85f15db7e99dd4e5cb6e5/gadgets/top_file/gadget.yaml#L9).

I'm OK if we want to add them in another PR, I'm just saying that we should add them in the operators' documentation. WDYT?

mauriciovasquezbernal:
My concern about providing that documentation here is that it'll be too much information for users. Going a step back, I've envisioned the documentation to be focused in three different targets:
- gadget users: Folks that want to run gadgets 
- gadget developers: Folks that want to create gadgets
- Inspektor Gadget developers: Folks that develop the framework of Inspektor Gadget

The [reference](https://inspektor-gadget.io/docs/latest/reference/) documentation is only oriented for "gadget users", and the [gadget development](https://inspektor-gadget.io/docs/latest/gadget-devel/) one for "gadget developers". In order to document operators we have two options:


### 1. Split the documentation for different targets (Idea on this PR) 

This PR is only covering the operators documentation for "gadget users". For gadget developers we have the information spread around on https://inspektor-gadget.io/docs/latest/gadget-devel/gadget-ebpf-api and https://inspektor-gadget.io/docs/latest/gadget-devel/metrics and we don't have anything for "Inspektor Gadget developers". Probably we should create a new operators page under the gadget developer to provide developer-specific details, like annotations, types, etc. 


### 2. Provide the full documentation of the operator on the same page

We could divide that on introduction, parameters, and developer specific things like annotations, types, etc and then we could also have architecture and more advance topics for "Inspektor Gadget developers".

I think keeping the documentation separated will help readers, even if that will complicate the life for us as things will be spread around. Any opinions on this one?

blanquicet:
> The [reference](https://inspektor-gadget.io/docs/latest/reference/) documentation is only oriented for "gadget users", and the [gadget development](https://inspektor-gadget.io/docs/latest/gadget-devel/) one for "gadget developers". 
> 
Mmmmm if your idea is that [reference](https://inspektor-gadget.io/docs/latest/reference/) documentation is only oriented for "gadget users", then I think we shouldn't put operators' documentation there. A "gadget users" doesn't need to know what an operator to be able to run a gadget. IMO, the [Running Gatgets](https://www.inspektor-gadget.io/docs/latest/reference/run) is already enough: it describes the filtering, output modes + fields and other special flags like timeout (Maybe here we are missing here the `--map-fetch-*`).

> My concern about providing that documentation here is that it'll be too much information for users. 
>
Now that you said that [reference](https://inspektor-gadget.io/docs/latest/reference/) documentation is only oriented for "gadget users", I understand why you're saying that. It's indeed too much info. But, IMO, as I said above, describing operators here is already too much. I just checked, and there is no one reference to operators in the whole folder `docs/reference`.

BTW, we need to specify somewhere (maybe in the README) that:

- Documentation for "Gadget users" is in docs/reference
- Documentation for "Gadget developers" is in docs/gadget-devel
- Documentation for "Inspektor Gadget developers" is in docs/development (?)

> In order to document operators we have two options:
>

My proposal:

- "Gadget users" (docs/reference): No specific reference or info of operators, just what is written in the README.
- "Gadget developers" (docs/gadget-devel): The content of this PR + annotations.
- "Inspektor Gadget developers" (docs/development or wherever it will go): Description of methods implemented by operators (`PreStart`, `Start`, etc.), what should be done on each stage, which operators runs on server and/or client, etc. Then, all this info is referenced from an architecture doc where we have a diagram showing which methods are called for gRPC calls, e.g., for `GetGadgetInfo` we instantiate operators and call `ExtraParams`, while for `Run` we do the same + call `PreStart` and `Start`.

mauriciovasquezbernal:
> Mmmmm if your idea is that [reference](https://inspektor-gadget.io/docs/latest/reference/) documentation is only oriented for "gadget users", then I think we shouldn't put operators' documentation there.

I think it's not bad to document the operators here. They are a key component of the whole architecture and users interact with them by using the flags. If we don't document them here, where can we put the flags that operators provide?

> BTW, we need to specify somewhere (maybe in the README) that:

Let's discuss this documentation separation with the team to be sure we're all aligned on that.

blanquicet:
> > Mmmmm if your idea is that [reference](https://inspektor-gadget.io/docs/latest/reference/) documentation is only oriented for "gadget users", then I think we shouldn't put operators' documentation there.
> 
> I think it's not bad to document the operators here. They are a key component of the whole architecture and users interact with them by using the flags.
> 
Do the users need to know they are interacting with the operators? Isn't it enough to know what flags they can use with an example?

> If we don't document them here, where can we put the flags that operators provide?
>
We are already describing them in the different guides: 

- [Running Gadgets](https://www.inspektor-gadget.io/docs/latest/reference/run) describes how to use the flags provided by cli (output mode and fields) and localmanager/kubemanager (for filtering) operators.
- [Insecure Registries](https://www.inspektor-gadget.io/docs/latest/reference/insecure-registries), [Restricting Gadgets](https://www.inspektor-gadget.io/docs/latest/reference/restricting-gadgets), [Verifying Assets](https://www.inspektor-gadget.io/docs/latest/reference/verify-assets), etc., describe the flags provided by the oci operator.
- [ig](https://www.inspektor-gadget.io/docs/latest/reference/ig) describes the flags provided by the localmanager operator specifically for `ig`, like [enrich-with-k8s-apiserver](https://www.inspektor-gadget.io/docs/latest/reference/ig#usage-with-flag---enrich-with-k8s-apiserver). BTW, I noticed we are missing the ones related to the runtime: `runtime`, `runtimeprotocol`, the ones for the socket path, etc. But as we have discussed several times, that doc needs to be refactored.

In general, I prefer presenting flags to "gadget users" in a feature-oriented way, instead of listing the flags per operator. Instead, for "gadget developers" and "IG developers", I definitely want to have documentation for each operator and all the flags, annotations, etc., it offers.

Of course, that's my preference. Let's try to understand together what's the most clear way we can present all this for different users 🙂.

docs/reference/operators/oci.md
```diff
@@ -0,0 +1,82 @@
...
+Default: `false`
+
+## Instance Parameters
+
+### `authfile`
+
+TODO: is this really a instance param?
+
+Path of the authentication file. This overrides the `REGISTRY_AUTH_FILE`
+environment variable.
```
alban, line 52: When using client/server (gadgetctl or kubectl-gadget), is this the environment variable on the client or server process?

Is the path referring to a file on the client machine or the server machine?

docs/reference/operators/oci.md
```diff
@@ -0,0 +1,82 @@
...
+
+Fully qualified name: `operators.oci.authfile`
+
+Default: `/var/lib/ig/config.json`
```
alban, line 56: What about `$HOME/.docker/config.json`?

docs/reference/operators/oci.md
```diff
@@ -0,0 +1,82 @@
...
+
+### `validate-metadata`
+
+Validate the gadget metadata before running the gadget
```
alban, line 60: Since it is an instance parameter, does it mean a client can disable validation against the wishes of the server sysadmin?

docs/reference/operators/sort.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Sort
+---
+
+The Sort operator sorts the output. This operator is only enabled for data
+sources of type array.
```
alban, line 6: Does it operate on the client or the server?

docs/reference/operators/cli.md
```diff
@@ -0,0 +1,27 @@
+---
+title: CLI
+---
+
+The CLI operator prints information to the terminal.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > When using ig+gadgetctl or kubectl-gadget+k8s, is the CLI operator running on the server or the client?

I changed the text to be:

>  The CLI operator prints information to the terminal. When using `gadgetctl` or
> `kubectl-gadget`, this opereator takes the data coming from the server and
> prints it to the terminal.

>  Is it enabled/disabled automatically, or can the user do that on the command line?

All operators are automatically enabled. Any reason why you're asking about this specific one?

docs/reference/operators/ebpf.md
```diff
@@ -0,0 +1,41 @@
...
+
+Network interface to attach networking programs to.
+
+Fully qualified name: `operators.oci.ebpf.iface`
+
+### `trace-pipe`
+
+Print to the terminal the `/sys/kernel/debug/tracing/trace_pipe` file, i.e.
+where the `bpf_printk()` function prints to. This is useful for debugging
+gadgets.
```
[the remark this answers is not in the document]
mauriciovasquezbernal: It uses the gadget logging mechanism. Not CLI operator nor dataresource are involved IIUC.

I updated the text to be:

> Print debug information generated by eBPF with `bpf_printk()` to the terminal.

Hopefully that's clearer.

docs/reference/operators/kubemanager.md
```diff
@@ -0,0 +1,64 @@
+---
+title: KubeManager
+---
+
+The KubeManager operator is used when Inspektor Gadget runs on Kubernetes. This
```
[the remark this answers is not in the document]
mauriciovasquezbernal: One more reason for #2100

docs/reference/operators/oci.md
mauriciovasquezbernal: It's the path on the server. According to the TODO above, this shouldn't the exposed when running gadgetctl or kubectl-gadget. I'll handle that in another PR.

docs/reference/operators/oci.md
mauriciovasquezbernal: That's a fallback

docs/reference/operators/oci.md
mauriciovasquezbernal: That's right. Do you think it should be moved to be a global param?

docs/reference/operators/sort.md
mauriciovasquezbernal: Only on the server AFAIU. cc @flyth and @blanquicet

docs/reference/operators/btfgen.md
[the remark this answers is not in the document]
mauriciovasquezbernal: I decided to split it in two different folders. I'm still not sure was an operator can be both, but I don't have better ideas...

docs/reference/operators/cli.md
```diff
@@ -0,0 +1,27 @@
...
+
+Determines the output format:
+- `json`: Outputs in JSON format.
+- `jsonpretty`: Outputs in pretty-printed JSON format.
+- `columns`: Outputs in a columnar format.
+- `yaml`: Outputs in YAML format.
+- `none`: No output.
+- `raw`: Outputs raw data.
+
+Fully qualified name: `operators.cli.mode`
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Added a quick comment about it. Will handle the links and examples in #3310

docs/reference/operators/kubeipresolver.md
```diff
@@ -0,0 +1,26 @@
...
+      "k8s-app": "kube-dns",
+      "kubernetes.io/cluster-service": "true",
+      "kubernetes.io/name": "CoreDNS"
+    }
+  }
+```
+
+## Parameters
+
+None
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > When is this operator enabled or disabled? I think each operator should document that.

I added a general comment indicating how operators are enabled or disabled. It's still missing to document that for each operator. I added a follow up on the PR description. 

> How does it relate to the flags --enrich-with-k8s-apiserver?

Good point. TBH I'm not sure, pinging @burak-ok since he has played with both things. 

> And when using the Golang API?

I suppose it'll work fine if there is a kubeconfig available, however I'm not sure how a user of the API could pass the clientset directly, I think users should be able to do that (later on).

Ref #3370

docs/reference/operators/layer/ebpf.md
```diff
@@ -0,0 +1,40 @@
...
+Interval in which to iterate over "iter maps", i.e. hashmaps marked with
+`GADGET_MAPITER()`
+
+Fully qualified name: `operators.oci.ebpf.map-fetch-interval`
+
+Default: `1000ms`
+
+### `map-fetch-count`
+
+Number of map fetch cycles - use 0 for unlimited.
```
blanquicet, line 36: Also this one is only available for `GADGET_MAPITER` data sources.

docs/reference/operators/data/filter.md
```diff
@@ -0,0 +1,31 @@
...
+### `filter`
+
+This parameter allows you to filter events based on specific field values
+provided by the gadget. This is particularly useful for narrowing down the
+output to entries that meet certain criteria.
+
+The filter syntax supports the following operations:
+
+```bash
+- `columnName==value`: Matches if the content of `columnName` equals exactly `value`.
```
blanquicet, line 22: Please apply changes in https://github.com/inspektor-gadget/inspektor-gadget/pull/3416.

docs/reference/operators/btfgen.md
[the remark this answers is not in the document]
blanquicet: I think the confusion of an operator being both layer and data operators comes from the wrong definition we were using for each type of operator:

- Layer: Handles layers and produce data but doesn't handle data (wrong because ebpf and wasm actually handle data).
- Data: Handles data but doesn't produce data (wrong because we want to produce data from some data op in the future).

Instead, to clarify things, I would suggest the following definition: Both type of operators **can** produce and handle data. The only difference is that layer operators does it by using additional data coming from a layer (and that's what makes them a layer operator):

- Wasm: Handle data using the WASM module present in the wasm layer.
- eBPF: Produce and handle ([formatters](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/ebpf/formatters.go)) data using the eBPF program present in the ebpf layer.
- btfgen: Neither produce nor handle data (it's fine, we say they can, not they must). Instead, it uses files present in the bpfgen layer to allow Gadgets run in machines without BTF.

WDYT?

docs/reference/operators/localmanager.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `container-namespace`
+
+Containerd namespace to use
+
+Default: `k8s.io`
+
+### `runtime-protocol`
+
+Container runtime protocol. Supported values are: internal, cri
```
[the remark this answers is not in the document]
blanquicet: I would say it's implicit it only works with Docker, Containerd and CRI runtimes and we don't need to specify that, right?

docs/reference/operators/formatters.md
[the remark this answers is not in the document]
blanquicet: Mmmm what about now that we agreed this is not anymore user-focused but developer-focused (gadget and IG developer) documentation? Should we add the annotations (in a follow-up PR)?

blanquicet (commented):
I think we'll be ready to go after moving the folder to a gadget developer-focused folder.

docs/reference/operators/layer/ebpf.md
mauriciovasquezbernal: Added it.

docs/reference/operators/data/filter.md
mauriciovasquezbernal: ops, thanks !

docs/reference/operators/formatters.md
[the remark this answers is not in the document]
mauriciovasquezbernal: Yes! Added a TODO in the PR description.

docs/reference/operators/ebpf.md
mauriciovasquezbernal: Added a "Document operator implementation" TODO in the description as well.

docs/reference/operators/sort.md
mauriciovasquezbernal: I added a TODO to document where all operators are used.

mauriciovasquezbernal:
After thinking more about it, I decided to move the whole operators documentation to spec/. I'm not sure we should be exposing this concept in that a direct way to users, instead we can mention the functionality operators provide without explicitly mentioning the operator (as we're already doing and as it was mentioned in https://github.com/inspektor-gadget/inspektor-gadget/pull/3406#issuecomment-2317107437). 

Hence, I think it's fine to add much more details to the documentation of operators, like annotations, implementation, etc. I added a bunch to TODOs to be handled later on.

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,53 @@
...
+### Instance
+
+These parameters can be set each time a gadget is run and only have impact on
+the gadget being run.
+
+Parameters are passed as flags when [running](../../reference/run.mdx) a gadget using the CLI
+tools. When using the [APIs](../../apis), it's required to use the full
+qualified name.
+
+<DocCardList />
```
blanquicet, line 53: Should we put a tittle between the `Instance` section and the list of operators. It looks strange (like there were part of the instance section):

image

blanquicet (approved):
LGTM! Thanks for working on this!

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,53 @@
...
+([1](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.31.0/pkg/operators#DataOperator),
+[2](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.31.0/pkg/operators#ImageOperator)).
+They are responsible for specific tasks, such as loading eBPF programs,
+enriching events, sorting output, etc. All operators can generate, drop or
+manipulate data and few of them handle the
+[layers](../../oci.md#image-layers-and-media-types) of gadget OCI image.
+
+Operators are arranged in a pipeline according to their priority:
+
+![Opeators Pipeline](./operators-pipeline.svg)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > The localmanager and kubemanager operators should be after the oci (ebpf, wasm, btf) operators, together with the filter, sorter, etc. Why did you put them at the beginning of the pipeline?

Too fast copy and paste from a previous image created by @flyth. 

> BTW, (not related with this PR), I just realized the filter should have higher priority so that we avoid enriching (localmanager/kubemanger, formatter, uidgidresolver, etc.) events that will be discarded by the filter operator.

What about if we want to filter by enriched data?

Thanks for the priority data. I added it to each operator, perhaps we can put a document with your comment too.

docs/reference/operators/btfgen.md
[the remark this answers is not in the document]
mauriciovasquezbernal: I updated the documentation again not to have a strong differentiation of data and layer operators. I only mentioned that all operators can act over data and few of them handle layers of the OCI image. I think that's more clear.

docs/spec/operators/index.mdx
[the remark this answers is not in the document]
blanquicet: > What about if we want to filter by enriched data?
>
Good point. That's why we put it there hahaha.

> Thanks for the priority data. I added it to each operator, perhaps we can put a document with your comment too.
>
Yep. A central point where to see all prios in orde would be very useful. Maybe somehow auto-generated and verified in the CI?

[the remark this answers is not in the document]
blanquicet: Now that I think better. Maybe we should start using always mermaid graph as they can be easily updated by anyone. WDYT?

docs/spec/operators/cli.md
```diff
@@ -0,0 +1,35 @@
+---
+title: CLI
+---
+
+The CLI operator prints information to the terminal. When using `gadgetctl` or
+`kubectl-gadget`, this opereator takes the data coming from the server and
+prints it to the terminal.
```
flyth, line 7: Maybe simpler:
The CLI operator is the default output operator - it prints all information and data from a gadget to the standard output (usually your terminal). It supports different output formats, depending on the gadget and the use case.

docs/spec/operators/ebpf.md
```diff
@@ -0,0 +1,41 @@
...
+The eBPF operator handles the [ebpf layer](../oci.md#the-ebpf-layer)
+that contains an ELF file with the eBPF programs. This operator loads the
+programs into the kernel and attaches them to the different hooks as specified
+by the gadget developer.
+
+## Instance Parameters
+
+### `iface`
+
+Network interface to attach networking programs to.
```
flyth, line 15:
```suggestion
Network interface to attach networking programs to. Only available if the gadget uses networking programs.
```

Not sure if it's worth mentioning that.

docs/spec/operators/ebpf.md
```diff
@@ -0,0 +1,41 @@
...
+### `trace-pipe`
+
+Print debug information generated by eBPF with `bpf_printk()` to the terminal.
+
+Fully qualified name: `operators.oci.ebpf.trace-pipe`
+
+### `map-fetch-interval`
+
+Interval in which to iterate over "iter maps", i.e. hashmaps marked with
+`GADGET_MAPITER()`
```
flyth, line 28:
```suggestion
Interval in which to iterate over eBPF maps that have been marked with
`GADGET_MAPITER()`
```

docs/spec/operators/ebpf.md
```diff
@@ -0,0 +1,41 @@
...
+
+Fully qualified name: `operators.oci.ebpf.map-fetch-interval`
+
+Default: `1000ms`
+
+### `map-fetch-count`
+
+Number fetch cycles (use 0 for unlimited) for "iter maps", i.e. hashmaps marked
+with `GADGET_MAPITER()`.
```
flyth, line 37:
```suggestion
Number of fetch cycles (use 0 for unlimited) for eBPF maps that have been marked with `GADGET_MAPITER()`.
```

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: 'Operators'
+sidebar_position: 0
+description: Documentation for Operators
+---
+
+import DocCardList from '@theme/DocCardList';
+
+Operators are the cornerstone of the Inspetkor Gadget architecture. They are
```
flyth, line 9:
```suggestion
Operators are the cornerstones of the Inspektor Gadget architecture. They are
```

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,55 @@
...
+enriching events, sorting output, etc. All operators can generate, drop or
+manipulate data and few of them handle the
+[layers](../../oci.md#image-layers-and-media-types) of gadget OCI image.
+
+Operators are arranged in a pipeline according to their priority:
+
+![Operators Pipeline](./operators-pipeline.svg)
+
+Operators are enabled automatically based on the gadget that's being run,
+according to the layers present on the OCI image, to the types used by the eBPF
```
flyth, line 23:
```suggestion
according to the layers present on the OCI image, by the types used by the eBPF
```

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,55 @@
...
+programs, etc. It's currently [not
+possible](https://github.com/inspektor-gadget/inspektor-gadget/issues/1992) to
+explicitly disable an operator. Also, all operators are currently built-in, [we
+plan](https://github.com/inspektor-gadget/inspektor-gadget/issues/2497) to
```
flyth, line 27:
```suggestion
however plan](https://github.com/inspektor-gadget/inspektor-gadget/issues/2497) to
```

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,55 @@
...
+control sensitive things like security configuration that can only be set by the
+system administrator.
+
+### Instance
+
+These parameters can be set each time a gadget is run and only have impact on
+the gadget being run.
+
+Parameters are passed as flags when [running](../../reference/run.mdx) a gadget using the CLI
+tools. When using the [APIs](../../apis), it's required to use the full
```
flyth, line 50:
```suggestion
Parameters are passed as flags when [running](../../reference/run.mdx) a gadget using the clients (kubectl-gadget / gadgetctl). When using the [APIs](../../apis), it's required to use the full
```

docs/spec/operators/kubemanager.md
```diff
@@ -0,0 +1,68 @@
+---
+title: KubeManager
+---
+
+The KubeManager operator is used by the gadget Pod, i.e. then Inspektor Gadget
+is deployed to Kubernetes as a DaemonSet. This operator keeps track of the pods
```
flyth, line 6:
```suggestion
The KubeManager operator is used when Inspektor Gadget
is deployed to Kubernetes as a DaemonSet. This operator keeps track of the pods
```
Maybe easier to understand.

docs/spec/operators/kubemanager.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Priority
+
+-1
+
+## Instance Parameters
+
+### `containername`
+
+Show only data from containers with that name
```
flyth, line 42:
```suggestion
Show only data from containers with the specified name
```

docs/spec/operators/kubemanager.md
```diff
@@ -0,0 +1,68 @@
...
+
+Fully qualified name: `operators.KubeManager.containername`
+
+### `podname`
+
+Show only data from pods with that name
```
flyth, line 48:
```suggestion
Show only data from pods with the specified name
```

docs/spec/operators/localmanager.md
```diff
@@ -0,0 +1,107 @@
...
+- runtime
+  - runtime name
+  - container name
+  - container ID
+  - container image name
+  - container image digest
+  - container started time
+
+It also adds some Kubernetes information gathered from the container runtimes,
+i.e. this operator doesn't talk to the kube-apiserver unless the
```
flyth, line 26:
```suggestion
but it doesn't talk to the kube-apiserver unless the
```

docs/spec/operators/localmanager.md
```diff
@@ -0,0 +1,107 @@
...
+Connect to the K8s API server to get further K8s enrichment, like the [owner
+reference](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).
+
+Default: `false`
+
+## Instance Parameters
+
+### `containername`
+
+Show only data from containers with that name
```
flyth, line 97:
```suggestion
Show only data from containers with the specified name
```

I know we also use it that way on the `--help`, but it sounds weird.

docs/spec/operators/oci.md
```diff
@@ -0,0 +1,83 @@
...
+
+The OCI operator handles Gadgets images. It pulls images from the OCI registries
+and invokes the different layer operators according to the
+[layers](../oci.md#image-layers-and-media-types) present on the image.
+
+## Global Parameters
+
+### `verify-image`
+
+Verify image using the provided public key. Check [Verify image-based
```
flyth, line 14:
```suggestion
Verify the gadget's OCI image using the provided public key. Check [Verify image-based
```

docs/spec/operators/oci.md
```diff
@@ -0,0 +1,83 @@
...
+
+Public keys used to verify the gadgets. Check [Verify image-based
+gadgets](../../reference/verify-assets.mdx#verify-image-based-gadgets) to learn more.
+
+Default: [Inspektor Gadget public
+key](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/resources/inspektor-gadget.pub).
+
+### `allowed-gadgets`
+
+List of allowed gadgets. If a gadget is not part of it, execution will be
```
flyth, line 29:
```suggestion
List of digests (checksums) that determine allowed gadgets. If a gadget's digest is not part of it, execution will be
```

docs/spec/operators/oci.md
```diff
@@ -0,0 +1,83 @@
...
+Disallow pulling gadgets from registries. Check [Disallow pulling
+Gadgets](../../reference/disallow-pulling.mdx) to learn more.
+
+Default: `false`
+
+## Instance Parameters
+
+### `authfile`
+
+TODO: is this really a instance param?
```
flyth, line 49: it shouldn't be

docs/spec/operators/oci.md
```diff
@@ -0,0 +1,83 @@
...
+- `missing`: Pull the gadget image only if missing.
+- `never`: Never pull the gadget image.
+
+Fully qualified name: `operators.oci.pull`
+
+Default: `missing`
+
+### `pull-secret`
+
+Secret to use when pulling the gadget image
```
flyth, line 81: IMHO this should also be global.

docs/spec/operators/socketenricher.md
```diff
@@ -0,0 +1,21 @@
+---
+title: SocketEnricher
+---
+
+The Socket enricher operator provides information about the socket owner, i.e.
+the process that created the socket. This is used to correlate sockets and
+process information in gadgets that only have information about the socket, like
+networking ones. The following information can be get from a socket object:
```
flyth, line 8:
```suggestion
networking ones. The following information can be gotten from a socket object:
```

docs/spec/operators/sort.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Priority
+
+9500
+
+## Instance Parameters
+
+### `sort`
+
+Sort by fields. Join multiple fields with ','. Prefix a field with '-' to sort
```
flyth, line 16:
```suggestion
Sort a gadget's output by fields. Join multiple field names with ','. Prefix a field with '-' to sort
```

flyth (approved):
Looks great!

Just some comments - none of them are blockers. Feel free to take & leave as you see fit.

docs/spec/operators/index.mdx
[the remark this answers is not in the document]
alban:
```mermaid
block-beta
  columns 8
  block:group1:2
    columns 3
    space oci space ebpf wasm btfgen
    style oci fill:#ffffff00,stroke:#ffffff00
  end
  block:group2:2
    columns 1
    kubemanager localmanager
  end
  filter
  sorter
  ...
  block:group3:2
    columns 1
    otelmetrics["otel metrics"]
    cli
  end

```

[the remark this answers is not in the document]
alban:
```mermaid
flowchart LR
    oci-->manager
    manager-->filter
    filter-->sorter
    sorter-->dotdotdot["..."]
    dotdotdot-->output
    subgraph oci
        ebpf
        wasm
        btfgen
    end
    subgraph manager[" "]
        direction TB
        kubemanager~~~localmanager
    end
    subgraph output[" "]
        otelmetrics["otel metrics"]~~~cli
    end
```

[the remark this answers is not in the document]
alban: With mermaid, we can make boxes clickable to the right section in the documentation.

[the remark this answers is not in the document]
mauriciovasquezbernal: I took the version without the clickable box. I wanted them to point to the operator documentation but I wasn't able.

docs/spec/operators/index.mdx
mauriciovasquezbernal: I added the list of different tools as suggested. I don't think clients is totally right as it also applies to `ig`.

docs/spec/operators/socketenricher.md
mauriciovasquezbernal: I know that's the right form but it sounds very weird to me.

mauriciovasquezbernal:
Thanks a lot all for the reviews. Since I got a lot of help, I added your signoffs to the commit with the documentation. 

There are still some TODOs, but I'd rather merge it now instead of keep iterating forever.

docs/spec/operators/index.mdx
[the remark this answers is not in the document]
alban: @mauriciovasquezbernal I think you mistakenly merged the version with the links.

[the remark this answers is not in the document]
mauriciovasquezbernal: ops, opened #3437

docs/spec/operators/index.mdx
```diff
@@ -0,0 +1,76 @@
...
+sidebar_position: 0
+description: Documentation for Operators
+---
+
+import DocCardList from '@theme/DocCardList';
+
+Operators are the cornerstones of the Inspektor Gadget architecture. They are
+modular components implementing the same interfaces
+([1](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.31.0/pkg/operators#DataOperator),
+[2](https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget@v0.31.0/pkg/operators#ImageOperator)).
```
blanquicet, line 12: @mauriciovasquezbernal why to point to a specific version and not to point to the latest:

https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget/pkg/operators#DataOperator
https://pkg.go.dev/github.com/inspektor-gadget/inspektor-gadget/pkg/operators#ImageOperator

?

mauriciovasquezbernal: I didn't find a way to point to latest, but you're right, latest it better.

blanquicet: https://github.com/inspektor-gadget/inspektor-gadget/pull/3440

docs/spec/operators/localmanager.md
```diff
@@ -0,0 +1,107 @@
...
+  - pod name
+  - namespace
+  - pod labels
+  - owner (only when using `enrich-with-k8s-apiserver`)
+
+## Priority
+
+-1
+
+## Global Parameters
```
blanquicet, line 39: Why global params don't have the fully qualified name? That's what how we are stating params can be passed through [environment variables and config file](https://www.inspektor-gadget.io/docs/latest/reference/configuration):

> You can use environment variables to configure the behavior of the run command. The environment variables use fully qualified names (as in the [configuration file](https://www.inspektor-gadget.io/docs/latest/reference/configuration#configuration-file)) with the prefix `INSPEKTOR_GADGET_`.
>

However, I think I found an inconsistency. Let's use runtimes param as example:

- Env variable: `INSPEKTOR_GADGET_OPERATOR_LOCALMANAGER_RUNTIMES`
- Config file: `operator.localmanager.runtimes`
- But if it were an instance param: `operator**s**.localmanager.runtimes`

cc @flyth

docs/reference/operators/kubeipresolver.md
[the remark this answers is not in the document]
burak-ok: The IP enricher currently is not working for image based gadgets. That seems to be an oversight. Issue https://github.com/inspektor-gadget/inspektor-gadget/issues/3483

The `enrich-with-k8s-apiserver` flag only adds `WithKubernetesEnrichment` and does nothing else. Enabling these operators (and passing the correct clientset) is still TODO. Issue https://github.com/inspektor-gadget/inspektor-gadget/issues/3484

PR #3410, opened 2024-08-28
docs: add otel-metrics operator docs

flyth:
Add docs for the otel-metrics operator.

Q: Should we add info about annotations in here as well?


docs/reference/operators/otel-metrics.md
```diff
@@ -2,8 +2,38 @@
...
+### `otel-metrics-name`
+
+Overrides the name of a datasource and explicitly sets it as name for the export. This is mandatory if you
+want to export the metrics. Use a name that is unique to the gadget+params combination to avoid collision of metrics.
+
+Default: ``
+
+### `otel-metrics-print-interval`
+
+Interval in which metrics should be emitted as human-readable text. This only has effect for data sources that are
```
blanquicet, line 35: Why we started talking about human-readable? Is there other kid of output, is it configurable?

docs/reference/operators/otel-metrics.md
```diff
@@ -2,8 +2,38 @@
...
+
+The listen address that should be serving Prometheus requests.
+
+Default: `0.0.0.0:2224`
+
+## Instance Parameters
+
+### `otel-metrics-name`
+
+Overrides the name of a datasource and explicitly sets it as name for the export. This is mandatory if you
```
blanquicet, line 28: I didn't understand why it's mandatory in such a case, but I haven't tried the operator yet. Let me try it and come back here.

docs/reference/operators/otel-metrics.md
flyth: This was added to mimic the behavior of profile gadgets without having full rendering support of metrics to the CLI in a better way (I think it only makes sense with a proper UI). So this converts metrics to the histogram output that we've shown in the built-in versions and annotate it as "text" so the CLI knows to just print it (and not consider it as columns or such.)
You can also access metrics using JSON.

docs/reference/operators/otel-metrics.md
flyth: Mandatory to make sure _you mean it_. Otherwise gadget runs (same gadget, different parameters for example) would create metrics with the same name, export it and then mess up existing data. This way you _explicitly_ tell it to export the metrics with a certain (stable) name.

docs/reference/operators/otel-metrics.md
```diff
@@ -2,8 +2,38 @@
...
+want to export the metrics. Use a name that is unique to the gadget+params combination to avoid collision of metrics.
+
+Default: ``
+
+### `otel-metrics-print-interval`
+
+Interval in which metrics should be emitted as human-readable text. This only has effect for data sources that are
+annotated using `metrics.print=true`. This is also limited to print histograms for now. This functionality might be
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Ref https://github.com/inspektor-gadget/inspektor-gadget/pull/3406#issuecomment-2315533666 to keep all discussion there.

mauriciovasquezbernal (approved):
LGTM!

PR #3413, opened 2024-08-28
pkg/operators/cli: Improve log level for errors and avoid blocking other ds when one fails

blanquicet:
# Improve log level for errors and avoid blocking other ds when one fails

See individual commits.

### DIFF
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -329,7 +329,7 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error 
 					fmt.Print(string(yml))
 					return nil
 				}, Priority)
-				return nil
+				continue
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/operators/cli/clioperator.go
```diff
@@ -327,7 +329,7 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 					fmt.Print(string(yml))
 					return nil
 				}, Priority)
-				return nil
+				continue
```
flyth, line 332: Nice catch!

flyth (approved):
Looks good from code inspection, thanks for the quick fix!

Two suggestions:
* maybe warn-level is better suited here - I think it makes sense to reserve error as level for when we completely error out and not continue running the gadget at all.
* maybe add information like "; skipping datasource %q" to let the user know about the consequence of the warning/error

PR #3414, opened 2024-08-28
 pkg/datasource/columns: Default MaxWith attribute based on field type (For int and bool)

blanquicet:
# Default MaxWith attribute based on field type (For int and bool)

After https://github.com/inspektor-gadget/inspektor-gadget/pull/3401, I propose to always set the maximum width to the maximum number of characters an integer and bool can use. 

This PR also adds support for "type" special value in the field annotations for the Width, MaxWidth and MinWidth field attributes.

## Testing done

Before and after this PR, the output for the following command is still the same:

```bash
$ go run --exec "sudo" ./cmd/ig/... run $MY_REPO/trace_exec --verify-image=false --insecure-registries=$LOCAL_REPO --pull always --fields=pcomm,ppid,comm,pid
PCOMM                  PPID COMM                    PID
```

### DIFF
--- a/pkg/datasource/columns.go
+++ b/pkg/datasource/columns.go
@@ -213,8 +213,16 @@ func (ds *dataSource) Columns() (*columns.Columns[DataTuple], error) {
 			continue
 		}
 
-		if attributes.Width == 0 {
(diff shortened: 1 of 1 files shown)


flyth (approved):
LGTM from code inspection, thanks!

PR #3433, opened 2024-09-02
docs: Add tip for trying w/o installation

mqasimsarfraz:
Add a tip for quick-start without installation!

### DIFF
--- a/gadgets/audit_seccomp/README.mdx
+++ b/gadgets/audit_seccomp/README.mdx
@@ -23,12 +23,18 @@ one of these two conditions:
         ```bash
         $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/audit_seccomp:latest [flags]
         ```
+        :::tip
+        If you want to try the gadget without installation you can check [kubectl node debug](https://inspektor-gadget.io/docs/latest/quick-start#kubectl-node-debug).
+        :::
     </TabItem>
 
     <TabItem value="ig" label="ig">
         ```bash
         $ sudo ig run ghcr.io/inspektor-gadget/gadget/audit_seccomp:latest [flags]
         ```
+        :::tip
+        If you want to try the gadget without installation you can check [run in a container](https://inspektor-gadget.io/docs/latest/quick-start#run-in-a-container).
+        :::
     </TabItem>
 </Tabs>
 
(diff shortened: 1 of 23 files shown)


alban (approved):
LGTM.

But let's wait a bit to see if there is another opinion.

mauriciovasquezbernal (commented):
What about something more generic like:

"Please check the [quick start](https://inspektor-gadget.io/docs/latest/quick-start) guide to learn about the different ways to use Inspektor Gadget"

OR 

add a lot of tabs with all different combinations
gadgetctl 
kubectl node debug
ig on a container

mqasimsarfraz:
> What about something more generic like:
> 
> "Please check the [quick start](https://inspektor-gadget.io/docs/latest/quick-start) guide to learn about the different ways to use Inspektor Gadget"

I fear it might be too generic and it should be fine to show them the quick-start related to kubernetes or linux depending on the tab they select. What do you say?

> OR
> 
> add a lot of tabs with all different combinations gadgetctl kubectl node debug ig on a container

I had a similar idea. But that would mean repeating the guide for each case! :thinking:

blanquicet:
I'm not a big fan of this because, in a Kubernetes environment, users could also copy the `ig` binary into the node and run it from there, use `ig` from a pod, etc. What I mean is that mentioning only those 4 options in the gadgets guide oversimplies the variety of options people have to run Inspektor Gadget. Maybe it's fine for the `Quick Start` just because it's indeed a quick start, but we shouldn't promote those 4 options around the documentation. Probably, we should convert the `Installing on Kubernetes` to `Using on Kubernetes` and explain there all the options I mentioned above, but that's another discussion (out of scope for this PR).

Said that, I'd prefer to show how to run gadgets with `kubectl-gadget` and `ig` in a generic way in the gadgets' documentation, and that's it.

blanquicet:
> One possibility could be to provide the gadget url and then point to the [quick start](https://inspektor-gadget.io/docs/latest/quick-start) in case the users don't have IG available yet. This will be useful for users that start by reading the gadget documentation without first checking the quick-start guide.
>
Could makes sense. Maybe something like "If you’d like to try this gadget but don’t have Inspektor Gadget installed yet, check out the [quick start documentation](https://inspektor-gadget.io/docs/latest/quick-start) for easy setup instructions."

PR #3435, opened 2024-09-02
docs/gadget-devel/building: Clarify that sudo needs to be run with -E

mauriciovasquezbernal:


alban (approved):
This change looks good to me, but this other section needs the same change too:

https://inspektor-gadget.io/docs/latest/reference/images#authentication

`sudo ig image pull` and `sudo ig image push` needs `-E` for private registries in case it needs the credentials at `~/.docker/config.json` because the `~` symbol is resolved with `$HOME`.

```
$ sudo sh -c 'echo $HOME'
/root
$ sudo -E sh -c 'echo $HOME'
/home/alban
```

mauriciovasquezbernal:
@alban would you mind opening a PR with that fix? I think it'll need some discussion.

PR #3446, opened 2024-09-05
Add a second map to get pods and svcs by IP without GetCmp

matthyx:
# Add a second map to get pods and svcs by IP without GetCmp

Checking pprof on big clusters (45k Pods) led to discover that a lot of CPU was wasted iterating over the map to run a func, which is the current method to match Pod IP addresses in the cache.

This PR proposes a second map indexed directly on IP addresses to fix that.

## Testing done

Was tried in kubescape's node-agent https://github.com/kubescape/node-agent/pull/362 and deployed on that huge cluster where we checked the better performance of the tracers.

### DIFF
--- a/pkg/operators/common/kubeinventorycache.go
+++ b/pkg/operators/common/kubeinventorycache.go
@@ -49,8 +49,10 @@ type inventoryCache struct {
 
 	factory informers.SharedInformerFactory
 
-	pods cachedmap.CachedMap[string, *v1.Pod]
-	svcs cachedmap.CachedMap[string, *v1.Service]
+	pods     cachedmap.CachedMap[string, *v1.Pod]
+	podsByIp cachedmap.CachedMap[string, *v1.Pod]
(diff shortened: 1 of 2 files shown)


pkg/operators/common/kubeinventorycache.go
```diff
@@ -189,13 +197,15 @@ func (cache *inventoryCache) OnAdd(obj any, _ bool) {
 			return
 		}
 		cache.pods.Add(key, o)
+		cache.podsByIp.Add(o.Status.PodIP, o)
```
alban, line 200: What is the value of PodIP when the pod has `hostNetwork=true`?

During initialization, before the CNI assigns an IP to the pod, is `o.Status.PodIP` initialized to `""`?

Would it causes problems if the pod (or several pods) has `o.Status.PodIP=""`?

amitschendel (approved):

alban (approved):
Thanks!

PR #3447, opened 2024-09-05
pkg/operators/otel-metrics: don't automatically apply metrics type based on role tag

flyth:
Before this change, the operator would automatically flag every member of a map key as metrics.type=key. That however also affected types that were not suited for being a metrics key and thus led to an error. After this change, keys have to be explicitly set.

### DIFF
--- a/pkg/operators/otel-metrics/otel-metrics.go
+++ b/pkg/operators/otel-metrics/otel-metrics.go
@@ -618,9 +618,6 @@ func (m *otelMetricsOperatorInstance) PreStart(gadgetCtx operators.GadgetContext
 					metricsType = MetricTypeCounter
 				} else if f.HasAnyTagsOf("type:gadget_gauge__u32", "type:gadget_gauge__u64") {
 					metricsType = MetricTypeGauge
-				} else if f.HasAnyTagsOf("role:key") {
-					// This is probably coming from the key of an eBPF map
-					metricsType = MetricTypeKey
 				}
 			}
 


alban:
Does `docs/gadget-devel/metrics.md` explain how to explicitly set a key?

blanquicet (approved):
LGTM!

I tested it by trying to enable metrics collection on `top_tcp` like this:

```bash
diff --git a/gadgets/top_tcp/gadget.yaml b/gadgets/top_tcp/gadget.yaml
index e2ab551a1b4f..4f56919c5e5e 100644
--- a/gadgets/top_tcp/gadget.yaml
+++ b/gadgets/top_tcp/gadget.yaml
@@ -7,6 +7,7 @@ datasources:
   tcp:
     annotations:
+      metrics.collect: "true"
     fields:
       mntns_id:
         annotations:
@@ -14,6 +15,7 @@ datasources:
           template: ns
       comm:
         annotations:
+          metrics.type: "key"
           description: Process name
           template: comm
       runtime.containerName:
@@ -33,9 +35,11 @@ datasources:
           columns.hidden: true
       received:
         annotations:
+          metrics.type: "counter"
           description: Received bytes
       sent:
         annotations:
+          metrics.type: "counter"
           description: Sent bytes
       src:
         annotations:
```

Before this PR: It fails.
After this PR: It works and I can see metrics exported in Prometheus.


> Does `docs/gadget-devel/metrics.md` explain how to explicitly set a key?
>
@alban yes, it does 🙂

mauriciovasquezbernal:
The documentation wasn't totally clear: I opened #4365

Issue #2994, opened 2024-06-11
gadgets: Implement trace fsslower as image based one


mauriciovasquezbernal:
Port [trace fsslower](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/trace/fsslower) to be an image-based gadget.

burak-ok:
Needs dynamic binding of eBPF programs to tracepoints/kprobes. Depending on the user provided runtime parameter

mauriciovasquezbernal:
> Needs dynamic binding of eBPF programs to tracepoints/kprobes. Depending on the user provided runtime parameter

Exactly! This will handled by #3132 and #3131.

Issue #3131, opened 2024-07-04
wasm: Allow handling of eBPF objects


mauriciovasquezbernal:
There are cases when it's useful to allow the Wasm module to directly interact with eBPF objects.

- Manipulate maps: Get, Delete, Update, etc. (many possible operations allowed on a map)
- Manipulate the eBPF objects before loading
  - Disable or enable programs based on kernel features
  - BPFProgramDisableAutoAttach(progName string)
  - BPFProgramManualAttach(progName string, attachTo string)
- Discover host features
  - KAllSyms (from pkg/kallsyms/kallsyms.go)
    - KAllSymsSymbolExists(name string) bool
    - KAllSymsLookupByInstructionPointer(ip uint64) string
    - KAllSymsSpecUpdateAddresses(funcName string) error
  - btf
    - BTFKernelSpecTypeByName(name string) (btfID u32) (TODO: how to marshall btf.Type?)

Ref https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.30.0/docs/design/003-wasm-support.md#handle-ebpf-objects

PR #3449, opened 2024-09-05
 wasm: Add parameters handling to wasm

### LINKED ISSUE #3132: wasm: Implement parameters handling

mauriciovasquezbernal:
Allow gadgets developers to interact with parameters
    from the wasm module:
    - add new parameters
    - get value of parameters

Fixes #3132

It's needed by #2994


docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -280,3 +280,58 @@ Parameters:
 
 Return value:
 - None
+
+### Parameters
+
+#### `newParam(key string, description string, defaultValue string, typeHint string, title string, alias string, isMandatory u32) u32`
```
alban, line 286: Do we need `newParam`?

Wouldn't it be enough to define a parameter in gadget.yaml and add `setParam` in the wasm api?

mauriciovasquezbernal: Some comments on this:

- Currently it's not that easy to do as the params on the metadata file are only handled by the ebpf operator [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/d28c76d0ef853f40fa5594b24e3febf044b757d6/pkg/operators/ebpf/params.go#L105), it means those are only visible to the ebpf operator. 
- Even if we could define the parameters on the metadata file, I still think it's valuable to be able to create them from wasm:
  - The user doesn't have to bother defining the parameter in one place and reading it in another 
  - The parameter creation could use some complex logic that can't be specified in yaml

mauriciovasquezbernal: > On the other hand, it means the --help documentation can only be printed by executing the wasm module, or at least the gadgetInit() function of the wasm module.

This is done similarly in other operators. For instance, the ebpf operator parses the eBPF spec to add parameters like [iface](https://github.com/inspektor-gadget/inspektor-gadget/blob/fbb6b2443e3abc2080d5c202cb4200ef932829c8/pkg/operators/ebpf/ebpf.go#L505-L510). IMHO it's ok to call these functions to generate the help, in the end that's the source of truth regarding the available flags according to the env where the gadget is running.

> Also, the template gadgets/README-flags.template and the make -C gadgets readme would need to run the wasm module too. I would prefer to documentation directly available in gadget.yaml, so it can easily be read by third-party tools.

I have some concerns with auto generating the list of flags from the metadata file:
- The documentation don't include any flag provided by the different operators (--filter, --containername, --iface, etc). 
- What about flags that could be added dynamically based on a specific condition? Or that it's default value can be changed based on something?

In general I think third party tools (like UIs) will have to rely more in the `GetGagetInfo()` to be able to provide all parameters (flags) that are supported by a gadget in a specific environment. Perhaps we can use this also to generate the flags instead of only relying only on the metadata?

> With
https://github.com/inspektor-gadget/inspektor-gadget/pull/3475, both api.SetAnnotation and api.NewParam are called from the same gadgetInit() function.
With
https://github.com/inspektor-gadget/inspektor-gadget/issues/3131, there will be more functions doing low-level calls like reading kallsyms, using bpf() syscall. If there are done in the same gadgetInit() function, that means the documentation can only be generated on a Linux machine.

I actually have to introduce two new PreStart() and PostStop() functions in #3463. We can (and should) document that the Init() one is called in preparation (also when running `--help`) stating that the gadget developer has to be careful not to perform heavy operations there. 

> If you still want to have it defined in the wasm module, I think it should be in a dedicated function gadgetAddParams, so that --help does not need to call gadgetInit() doing low-level calls.

Please see the comment above above PreStart / PostStop. Those heavy operations could be done on PreStart().

alban: For parameters defined in the eBPF code, we have `--update-metadata` to read them from the eBPF module and copy them in gadget.yaml. Do you think we can have something similar for the parameters defined in the WASM module?

I realise it does not work for parameters dynamically generated from the current environment (e.g. kallsyms, list of LSM hooks, etc.). But in my opinion, that's something to avoid. In a Kubernetes cluster with several nodes potentially using different kernel versions, it could potentially generate different parameters. When using `kubectl-gadget run` to get the help, it could give a different answer depending on the node that executed the WASM module.

> What about flags that could be added dynamically based on a specific condition? Or that it's default value can be changed based on something?

In #3463 (trace_fsslower), the parameter "filesystem" is not dynamic: it is just a string with the default value "ext4" (a string). So I don't know if we have a real use case where we need dynamic parameters that are defined differently depending on the environment.

It could also have the value "all" where it attaches the kprobes to all supported filesystems on the current node. So an heterogeneous cluster with some nodes with xfs support and some without, it would be possible to use `--filesystem=all` to attach it to whichever filesystems are supported on each node. No need for dynamic parameters for this use case.

mauriciovasquezbernal:
I updated the PR to define the parameter on the yaml file. I'm still missing:
- How to uniform it with ebpf parameters. It seems it touches al of things to get this right
- How to generate the gadget readme from it (related to the point above too)

mauriciovasquezbernal:
> How to uniform it with ebpf parameters. It seems it touches al of things to get this right
> How to generate the gadget readme from it (related to the point above too)

Fixed those two.

pkg/operators/ebpf/params.go
```diff
@@ -102,7 +102,7 @@ func (i *ebpfInstance) populateParam(t btf.Type, varName string) error {
 	}
 
 	// Fill additional information from metadata
-	paramInfo := i.config.Sub("params." + varName)
+	paramInfo := i.config.Sub("params.ebpf." + varName)
```
flyth, line 105: Hmm, I was gonna ask whether we should keep supporting the old format for a bit longer, but looking at the metadata, it seems we never used `params.` but `ebpfParams.` for the gadgets... 🙈

mauriciovasquezbernal:
Moved the third commit to #3490 to make the review easier.

pkg/operators/ebpf/params.go
mauriciovasquezbernal: Yeah, there were also some questions about the compatibility in #3490.

docs/gadget-devel/gadget-wasm-api-raw.md
mauriciovasquezbernal: I removed the newParam function and switched to use the metadata. We can implement it back if needed later on.

pkg/operators/wasm/wasm.go
```diff
@@ -65,16 +66,37 @@ func (w *wasmOperator) InstantiateImageOperator(
...
 
+	cfg, ok := gadgetCtx.GetVar("config")
+	if !ok {
+		return nil, fmt.Errorf("missing configuration")
+	}
+	v, ok := cfg.(*viper.Viper)
+	if !ok {
+		return nil, fmt.Errorf("invalid configuration format")
+	}
+	instance.config = v
```
flyth, line 88: Should we just ignore a non-existent config here? Otherwise we're introducing basically a hard dependency on the viper instance that is currently only served by the oci operator.

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -280,3 +280,15 @@ Parameters:
 
 Return value:
 - None
+
+### Parameters
```
flyth, line 284: Maybe add a paragraph of how to define those parameters in the gadget.yaml?

flyth (commented):
Looks good - just one question around the config being mandatory.

pkg/operators/wasm/wasm.go
mauriciovasquezbernal: You're right. Let's ignore it for now. Even if IMO the config is going to be more and more informant and perhaps we can make it a first class citizen later on.

flyth (approved):
LGTM from code inspection! Thanks!

PR #3450, opened 2024-09-05
operators/ebpf: Fix crash when spec doesn't contain types

mauriciovasquezbernal:
Compiling an empty progra.bpf.c file will produce an eBPF layer with a collection spec without types, this causes a crash when trying to run such a gadget.

This commit add an additional check to ensure the collection spec contains types, and also adds a test case to cover this scenario.

### Testing 

Option 1: Revert the fix and run the test in this PR
Option 2: Create a gadget with an empty program.bpf.c and try to run it.


pkg/operators/ebpf/ebpf_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	orasoci "oras.land/oras-go/v2/content/oci"
+
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf"
+	ocihandler "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/oci-handler"
```
blanquicet, line 28: Not related to this PR: We need to uniform this and avoid some operators are enabled by blank import and other explicitly.

pkg/operators/ebpf/ebpf_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2024 The Inspektor Gadget authors
```
blanquicet, line 1: Thanks for creating the tests!

blanquicet (approved):
LGTM

PR #3458, opened 2024-09-09
treewide: Bump go to 1.22.7.

eiffel-fl:
encoding/gob is subject of CVE-2024-34156, which was fixed in go 1.22.7 [1, 2].

[1]: https://www.cve.org/CVERecord?id=CVE-2024-34156
[2]: https://pkg.go.dev/vuln/GO-2024-3106


alban (commented):
Should this file be updated too?
```
.github/workflows/cloud-cleanup.yml:3:  GO_VERSION: 1.22.0
```

And others:
```
$ git grep '1\.22\.'
```

alban:
> > Should this file be updated too?
> > ```
> > .github/workflows/cloud-cleanup.yml:3:  GO_VERSION: 1.22.0
> > ```
> 
> I will update it, but I do not really understand which version is used here, as:
> 
> https://github.com/inspektor-gadget/inspektor-gadget/blob/16caa8264a2fc374291cb1eb5ab4c5d6081cf068/tools/eks-cleanup/go.mod#L3

The go versions in `go.mod` don't specify which version of go will run but the minimum version of go acceptable by the project. There are more:`find -name go.mod|xargs grep 'go 1\.'`

I don't know if we should necessarily update them. This has an impact on third-party projects that import Inspektor Gadget packages.

alban (approved):
LGTM, thanks!

PR #3448, opened 2024-09-05
pkg/datasource: Use FullName to look for fields in config

blanquicet:
# Use FullName to look for fields in config

Similar to https://github.com/inspektor-gadget/inspektor-gadget/pull/3293 and https://github.com/inspektor-gadget/inspektor-gadget/pull/3271, we are not using the `FullName` to look for fields. This prevents us from correctly annotating fields.

## Testing done

Try to annotate fields with parent like this:

```bash
diff --git a/gadgets/top_tcp/gadget.yaml b/gadgets/top_tcp/gadget.yaml
index d942a1aa51c6..dcd89c74c7bd 100644
--- a/gadgets/top_tcp/gadget.yaml
+++ b/gadgets/top_tcp/gadget.yaml
@@ -8,6 +8,9 @@ datasources:
     annotations:
       cli.clear-screen-before: "true"
     fields:
+      runtime.containername:
+        annotations:
+          columns.hidden: true
       mntns_id:
         annotations:
           description: Mount namespace inode id
```

Then, build and push the gadget:

```bash
$ go run ./cmd/ig/... image build -t $LOCAL_REPO/top_tcp:no-run-container gadgets/top_tcp/
Pulling builder image ghcr.io/inspektor-gadget/ebpf-builder:latest
latest: Pulling from inspektor-gadget/ebpf-builder
Digest: sha256:993dd2ee641faaf7a038df4534c893023a0a94edd24b9ba127c2eb70a00ea33a
Status: Image is up to date for ghcr.io/inspektor-gadget/ebpf-builder:latest
Successfully built ghcr.io/blanquicet/top_tcp:no-run-container@sha256:ac0e605d9600f9220249b91dec3f314fa087923b5c6302d9b6a401d2d105dbf7
```

```bash
$ go run ./cmd/ig/... image push $LOCAL_REPO/top_tcp:no-run-container
Pushing ghcr.io/blanquicet/top_tcp:no-run-container...
Successfully pushed ghcr.io/blanquicet/top_tcp:no-run-container@sha256:ac0e605d9600f9220249b91dec3f314fa087923b5c6302d9b6a401d2d105dbf7
```

Then, run the same gadget with and without this PR:

- **Before this PR:** Annotation is not applied:

```bash
$ go run --exec "sudo -E" ./cmd/ig/... run $LOCAL_REPO/top_tcp:no-run-container --verify-image=false
RUNTIME.CONTAINERNAME         PID SRC                         DST                         COMM        SENT       RECEIVED
[...]

PR #3468, opened 2024-09-10
pkg/operators/oci-handler: Support annotate nested fields

blanquicet:
# Support annotate nested fields

After this PR, it's now possible to annotate nested fields like `runtime.containername` using the `--annotate` flag:

```bash
--annotate=tcp.runtime.containername:metrics.type=key
```
## Testing done

I tested it on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/2960, but after rebasing it onto `main` so that https://github.com/inspektor-gadget/inspektor-gadget/pull/3448 was taken:

Server:

```bash
sudo ig daemon -H tcp://$MY_IP:9999 \
			--otel-metrics-listen=true
```

Client:

```bash
gadgetctl --remote-address tcp://$IG_TARGET:9999 \
			run ghcr.io/inspektor-gadget/gadget/top_tcp:latest \
			--annotate=tcp:metrics.collect=true,tcp.runtime.containername:metrics.type=key,tcp.received:metrics.type=counter,tcp.sent:metrics.type=counter \
			--otel-metrics-name tcp:demotcp -v \
			--detach
```

Then, I generated some traffic in a container on the server side and checked the exported metrics, I saw the values exported with containerName as key:

```bash
$ curl http://$IG_TARGET:2224/metrics -s | grep received
# HELP received_total
# TYPE received_total counter
received_total{containerName="mycontainer",otel_scope_name="demotcp",otel_scope_version=""} 7.9485798e+07
```


pkg/operators/oci-handler/oci.go
```diff
@@ -291,6 +293,12 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
 			current[annotation[0]] = annotation[1]
 			viper.Set("datasources."+subject[0]+".fields."+subject[1]+".annotations", current)
 			log.Debugf("field annotation %q added", ann)
+		default:
+			// field in a nested struct
+			current := viper.GetStringMapString("datasources." + subject[0] + ".fields." + strings.Join(subject[1:], ".") + ".annotations")
```
flyth, line 298: Instead of joining, we could also use SplitN(..., 2) above.

pkg/operators/oci-handler/oci.go
```diff
@@ -279,6 +279,8 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
 
 		subject := strings.Split(annInfo[0], ".")
 		switch len(subject) {
+		case 0:
```
flyth, line 282: 0 will never be the case, as the remainder (even if empty) will be returned as first item. The error should only happen `if len(annInfo[])` is 0

flyth (commented):
Thanks!

Can you please adopt the initial support for `--annotate` in this PR? I'd like to remove it from the GadgetInstanceManager PR, as it's not really needed there.

blanquicet:
@flyth I took the initial support here and manage the comments.

docs/spec/operators/oci.md
```diff
@@ -85,3 +86,9 @@ Default: `missing`
 Secret to use when pulling the gadget image
 
 Fully qualified name: `operators.oci.pull-secret`
+
+### `annotate`
+
+Add annotations to datasources or fields.
```
flyth, line 92: Maybe add the format of the parameter here as well:

```suggestion
Add annotations to datasources or fields. Use a comma-separated list in the formats:

`datasource:annotation=value` to add an annotation to a datasource
`datasource.field:annotation=value` to add an annotation to the field of a datasource
```

PR #3472, opened 2024-09-11
pkg: Add safeset package.

eiffel-fl:
Before this commit, we used the private variable safeSet available in encoding/json through linkname.
But, since golang 1.23, linkname is not authorized by default [1].

So, this commit introduces the safeset package which defines the SafeSet map. This map has the exact same value as encoding/json.safeSet [2]. It is also safe to add it in our sources, as this variable was not modified upstream since its introduction in 2016 [3].

This permits the whole project to be compiled using golang 1.23.

[1]: https://github.com/golang/go/issues/67401
[2]: https://cs.opensource.google/go/go/+/release-branch.go1.23:src/encoding/json/tables.go;l=15-112
[3]: https://cs.opensource.google/go/go/+/ed8f207940c8787d344664a43071b235e2ce5c68

### DIFF
--- a/pkg/columns/formatter/json/json.go
+++ b/pkg/columns/formatter/json/json.go
@@ -25,6 +25,7 @@ import (
 	_ "unsafe"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/columns"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/safeset"
 )
 
 type column[T any] struct {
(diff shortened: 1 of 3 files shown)


flyth (approved):
Thanks for addressing this - looks good!

I'd prefer a copy (roughly 100 lines) for each of the two packages that requires it, to keep dependencies down, but I'm okay with not doing it at this time.

Issue #3274, opened 2024-08-01
Add support for dynamically linking eBPF programs in image based gadgets


burak-ok:
Some builtin gadgets like `trace fsslower` allow the user to decide which tracepoint to attach to (The user choses which filesystem format and we attach to the corresponding kprobe)

Other builtin gadgets like `top blockio` checks for the existence of tracepoints or kprobes and only then attaches to them according to what is on the system.

It would be nice if we can dynamically attach to tracepoints

h4l0gen:
Hey @burak-ok I want to work on this issue. Can you please assign it to me. Thank you!

burak-ok:
Correct, most gadgets are fine with hardcoded associations between the eBPF programs and tracepoints/kprobes.

As you said for some gadgets we need to make that association/linking dynamically. So at runtime and according to either user input, kernel versions or other stuff

mauriciovasquezbernal:
Fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/3449

PR #3475, opened 2024-09-11
wasm: Add support for setting config

mauriciovasquezbernal:
This will be used by the trace_fsslower gadget to attach programs to the right hook according to the user settings.

Needed by #3463


pkg/operators/wasm/misc.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	val, err := stringFromStack(m, valuePtr)
+	if err != nil {
+		i.logger.Warnf("setAnnotation: reading string from stack: %v", err)
+		stack[0] = 1
+		return
+	}
+
+	i.config.Set(fmt.Sprintf("%s.annotations.%s", objName, name), val)
```
flyth, line 69: This has to be differentiated between setting an annotation on a datasource and a field:

* datasource:
   datasources.%s.annotations.%s
* fields:
   datasources.%s.fields.%s.annotations.%s

Depending on the number of dots in objName.
* no dot: datasource annotation
* 1+ dots: field annotation

See also #3468

mauriciovasquezbernal: The idea is to give full control to the users and allow them to set the annotation on anything:

- datasource: `setObjAnnotation("datasources.myds", "fooannotation", "fooval")`
- fields: `setObjAnnotation("datasources.myds.fields.myfield", "fooannotation", "fooval")`
- programs: `setObjAnnotation("programs.ig_fssl_read_e", "attach_to", config.read)` from #3463

Otherwise we'll need to implement few different specific `set{Program,Annotation,Field}Annotation` helpers.

flyth: But shouldn't id be called setConfig() with just key and value, then?

mauriciovasquezbernal: That can be an option, but the user will have to pass the "annotations" suffix. Do we want to be that generic?

flyth: I'm not sure - that could easily create a hard-to-debug mess 😄

blanquicet:
Does this solve https://github.com/inspektor-gadget/inspektor-gadget/issues/3274?

cc @burak-ok

mauriciovasquezbernal:
> Does this solve #3274?
> 
> cc @burak-ok

Kind of, this and #3463 and #3449 or #3485 will fix it. (I'm working on it)

burak-ok:
> > Does this solve #3274?
> > cc @burak-ok
> 
> Kind of, this and #3463 and #3449 or #3485 will fix it. (I'm working on it)

Yes, the PRs combined will fix it

pkg/operators/wasm/misc.go
alban: > setObjAnnotation("programs.ig_fssl_read_e", "attach_to", config.read)

Can we attach a single bpf program to several targets? This can be useful for: 
- kprobes, e.g. the pwru use case
- uprobes for custom SSL gadgets

With kprobes, the wasm module can read kallsyms to decide how to annotate the bpf program.

But in the case of uprobes, we don't know in advance what are the container images and target programs that are going to run. It would be good to have a way to parse a container filesystem before deciding where to attach the uprobe. This could be at a different place depending on the container image.

This doesn't have to be supported in this PR but it's good to keep that in mind. I'm wondering if this API with annotations is enough for the advanced use cases.

alban: The current gadgets using uprobes are attached to libraries:
- libc (`trace_malloc`)
- libssl, libcrypto (`trace_ssl`)

And the path to the libraries (e.g. `/lib64/libc.so.6`) is determined by parsing `/etc/ld.so.cache` in the container. So adding an annotation like `uprobe/libc:malloc` with a similar syntax as the section name (`SEC("uprobe/libc:malloc")`) is good enough.

But for uprobes on executables (such as Golang programs because they don't use dynamic libraries) might need something different. I'd like to attach the uprobe on `crypto/tls.(*Conn).Write` but ideally the executable path (e.g. `/usr/bin/etcd` in one container, `/usr/bin/opa` in another container) would not need to be specified, neither in the section name nor in the annotation. The wasm module might want to have a way to look for Golang programs in the container filesystem.

flyth: I still think this API is confusing (especially when compared to #3468 - that and this PR should probably be aligned when we name this "set annotation").

I'm also not sure about introducing annotations to metadata sections other than datasources - do you really think it's necessary and we can't just use the keys/values directly? If we don't need them, then I'd really rather go with `setConfig()` (or similarly named) to give the gadget dev all the possibilities, and add a dedicated `setAnnotation()` for manipulating annotations on the datasources more easily - aligned with the `--annotate` param.

The PR itself looks good to me.

/cc @blanquicet

mauriciovasquezbernal: Updated it to be SetConfig instead. The annotations one is still TBD but as you mentioned privately, it can be done by calling `SetConfig("foo.bar.annotations.zas", "yesno")`

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -310,3 +310,17 @@ Parameters:
 
 Return value:
 - The value of the parameter.
+
+### Config
+
+#### `setConfig(key string, u64 val, u32 kind) uint32`
+
+Sets the value of key within the gadget's configuration.
```
flyth, line 318:
```suggestion
Sets the value of key within the gadget's configuration (gadget.yaml).
```

flyth (approved):
LGTM from code inspection.

pkg/operators/wasm/wasm_test.go
```diff
@@ -380,3 +381,40 @@ func TestWasmParams(t *testing.T) {
...
+
+func TestConfig(t *testing.T) {
+	utilstest.RequireRoot(t)
+
+	t.Parallel()
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	t.Cleanup(cancel)
+
+	ociStore, err := orasoci.NewFromTar(ctx, "testdata/config.tar")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I don't know really. Having these generated will make it much harder to run the unit tests, but I agree, let's discuss it on another issue.

wasmapi/go/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+	keyPtr := uint64(stringToBufPtr(key))
+
+	switch t := val.(type) {
+	case string:
+		valPtr := uint64(stringToBufPtr(t))
+		result = setConfig(keyPtr, valPtr, uint32(Kind_String))
+	default:
+		return fmt.Errorf("unsupported type: %T", val)
+	}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: To avoid having to change it back to a switch once we need add support for more types.

pkg/operators/wasm/wasm_test.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Opened #3513

PR #3485, opened 2024-09-13
operators/oci: allow adding custom parameters from gadget.yaml

flyth:
This adds support to create custom parameters in the manifest (`gadget.yaml`) that can then apply parameters, annotations and such to the base metadata. This allows easy creation of presets of configurations by the gadget author.

It uses [template/text](https://pkg.go.dev/text/template) for template logic and has access to the configuration (gadget params) and the metadata itself.

If the gadget developer needs even more logic, the fallback should be to use a WASM module for it (#3449, #3475), which will of course require the gadget author to be familiar with Go (or whatever language is used to compile to WASM in that case).

```
$ ig run mygadget --help
Run a gadget

Usage:
  ig run [flags]

Flags:
...
      --mycustomparam string                   Description for the param [option1] (default "option1")
      --mysecondparam string                    (default "")
```

When running a gadget, you can now set:
```sh
$ ig run mygadget --myCustomParam=option1 --mysecondparam foobar
```

Which would lead to these annotations being set:
```
myDataSource:columns.output=none
myDataSource:my.annotation=A template driven value, now foobar.
myDataSource:my.second.annotation=We can also access other configuration values, like the gadget name: mygadget
```

Basically, everything under `patch` will be used to patch the gadget metadata.

### DIFF
--- a/docs/gadget-devel/parameters.md
+++ b/docs/gadget-devel/parameters.md
@@ -4,6 +4,8 @@ sidebar_position: 610
 description: 'Gadget Parameters'
 ---
 
+## Parameters from eBPF programs
+
 A Gadget can expose parameters to the client from the eBPF program. Inspektor
 Gadget provides the mechanism to expose the parameters as CLI flags to the user
 and allow to set them from the configuration file.
@@ -30,3 +32,59 @@ params:
       defaultValue: "false"
       description: Description for the param
 ```
+
+## Customizable parameters
+
+Much of Inspektor Gadget's functionality is controlled by parameters and
+annotations. To give gadget authors more freedom, one can define new parameters
+that affect a group of other parameters, datasource and field annotations, operator's
+configuration or any other field in the metadata file. This works by patching the
+gadget metadata itself, depending on those custom parameters' values.
+
+In the `gadget.yaml`, create a new section called `custom` to `params` like this:
+
+```yaml
+params:
+  custom:
+    myCustomParam:
+      description: A parameter that will apply a patch based on one of the possible values.
+      defaultValue: "option1"
+      values:
+        option1:
+          patch:
+            datasources:
+              myDataSource:
+                annotations:
+                  columns.output: none
+    mySecondParam:
+      description: Another parameter. This will apply a patch globally, regardless of the actual value.
+      defaultValue: "default"
+      patch:
+        datasources:
+          myDataSource:
+            annotations:
+              my.annotation: >-
+                A template driven value, now {{call .getParamValue "custom.mySecondParam"}}.
+              my.second.annotation: >-
+                We can also access other configuration values, like the gadget name: {{call .getConfig "name"}}
+```
+
+All keys below `custom` will become new parameters that can be set by the user
+of the gadget. Depending on the parameters' value, a `patch` node will be applied
+to the gadget metadata itself. (See `myCustomParam` example above.)
+
+If no specific values are configured, a global `patch` section will be applied. (See
+`mySecondParam` example above.)
+
+Strings used can use templates (explained [here](https://pkg.go.dev/text/template)).
+Functions available to the templating engine are:
+
+### getParamValue(paramName)
+
+Can be used to get the value of any set parameter; paramName needs to be fully qualified, e.g.
+`custom.mySecondParam` or `operator.oci.ebpf.target_family`.
+
+### getConfig(key)
+
+Can be used to get the value of any metadata key like `name`,
+`params.custom.myCustomParam.description`, etc.
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -1,4 +1,4 @@
-// Copyright 2022-2024 The Inspektor Gadget authors
+// Copyright 2022-2025 The Inspektor Gadget authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"text/template"
 	"time"
 
 	"github.com/spf13/viper"
@@ -42,6 +43,8 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/parser"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
 )
+
+const customParamsKey = "params.custom"
 
 // GadgetContext handles running gadgets by the gadget interface; it orchestrates the whole lifecycle of the gadget
 // instance and communicates with gadget and runtime.
@@ -69,6 +72,7 @@ type GadgetContext struct {
 	localOperators []operators.DataOperatorInstance
 	vars           map[string]any
 	params         []*api.Param
+	paramValues    api.ParamValues
 	loaded         bool
 	imageName      string
 	metadata       []byte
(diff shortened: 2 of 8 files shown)


pkg/operators/oci-handler/oci.go
```diff
@@ -264,6 +266,91 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
...
+					v1 := valSub.Get(k1)
+					if s, ok := v1.(string); ok {
+						tpl, err := template.New(k1).Parse(s)
+						if err != nil {
+							return fmt.Errorf("parsing custom param %q value %q: %q cannot be parsed as template: %w", k, val, k1, err)
+						}
+						out := bytes.NewBuffer(nil)
+						err = tpl.Execute(out, map[string]any{
+							"paramValues": o.paramValues,
+							"getConfig": func(key string) string {
```
blanquicet, line 322: We have to document that people can't use `{{call .getConfig "params.X.Y"}}`.

pkg/operators/oci-handler/oci.go
```diff
@@ -264,6 +266,91 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
 		return fmt.Errorf("unmarshalling metadata: %w", err)
 	}
 
+	// Extract custom params
+	customParams := viper.Sub("params.custom")
+	if customParams != nil {
```
blanquicet, line 271: Please move all this parsing to a function and write some unit tests for it.

pkg/operators/oci-handler/oci.go
```diff
@@ -264,6 +266,91 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
...
+				Description:  paramSub.GetString("description"),
+				Prefix:       "custom.",
+				DefaultValue: paramSub.GetString("defaultValue"),
+			}
+			for value := range paramSub.GetStringMap("values") {
+				// Skip wildcard param
+				if value == "*" {
+					continue
+				}
+				p.PossibleValues = append(p.PossibleValues, value)
```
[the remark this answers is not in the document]
flyth: I don't know if we'll add more features in the future, but for now you're right, we can just skip those.

docs/gadget-devel/parameters.md
```diff
@@ -4,6 +4,8 @@ sidebar_position: 610
 description: 'Gadget Parameters'
 ---
 
+## Parameters from eBPF programs
```
alban, line 7: Before this section, you could have an introduction explaining that the parameters are generally grouped by operator. The group name is either the operator name (e.g. `ebpf`) or "custom" for parameters that don't belong to an operator.

Without this explanation, it is not clear that the group name cannot be arbitrarily chosen by the gadget author.

```
params:
  ebpf:
    myparam:
```

If the parameter group name is neither "custom" or an operator name, it will be ignored by IG.

alban: Alternatively, instead of grouping parameters by operator, could parameters be generic and have a field saying it belong to an operator?
```diff
 type GadgetMetadata struct {
 ...
         // Params exposed by this gadget. It includes params for different operators
-        Params map[string]map[string]params.ParamDesc `yaml:"params,omitempty"`
+        Params map[string]params.ParamDesc `yaml:"params,omitempty"`

 }
```
```diff
 type ParamDesc struct {
 ...
+    // Owner tell the operator that own this parameter. Can be empty if none.
+    Owner string
 }
```

docs/gadget-devel/parameters.md
```diff
@@ -29,3 +31,39 @@ ebpfParams:
...
+Much of Inspektor Gadget's functionality is controlled by parameters and
+annotations. To give gadget authors more freedom, you can define new parameters
+that affect a group of other parameters or annotations.
+
+In the `gadget.yaml`, create a new section called `custom` to `params` like this:
+
+```yaml
+params:
+  custom:
+    myCustomParam:
```
alban, line 46: Instead of abstract names, the documentation could give a real use case.

I suggest taking example on the `trace_exec` gadget with the following parameters:
```yaml
datasources:
  exec:
    fields:
      cwd:
        annotations:
          description: The current working directory of the process (require --paths flag)
          columns.width: 64
          columns.hidden: "true"
          columns.alignment: left
      exepath:
        annotations:
          description: The executable binary file of the process (require --paths flag)
          columns.width: 64
          columns.hidden: "true"

params:
  ebpf:
    collect_paths:
      key: collect-paths
      defaultValue: "false"
      description: Collect current working directory and executable path.
  custom:
    paths:
      key: paths
      defaultValue: "false"
      description: Print current working directory and executable path (implies collect-paths).
      values:
        true:
          applyConfig:
              // TODO: something to switch "columns.hidden" of the two fields above and also change collect_paths
        false:

```

pkg/operators/oci-handler/oci.go
```diff
@@ -264,6 +266,91 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
...
+		for k := range viper.GetStringMap("params.custom") {
+			log.Debugf("evaluating custom param %q", k)
+			paramSub := customParams.Sub(k)
+			valuesSub := paramSub.Sub("values")
+			if valuesSub == nil {
+				log.Debugf("custom param %q has no values", k)
+				continue
+			}
+			p := &api.Param{
+				Key:          k,
```
alban, line 281: The Key field might be different than the key of the map.

Notice `-` != `_` in the trace_exec gadget:
```
params:
  ebpf:
    ignore_failed:
      key: ignore-failed 
      defaultValue: "true"
      description: Ignore failed events
```

pkg/operators/oci-handler/oci.go
```diff
@@ -264,6 +266,91 @@ func (o *OciHandlerInstance) init(gadgetCtx operators.GadgetContext) error {
...
+			if val, ok := o.paramValues["custom."+k]; ok {
+				valName := val
+				if !paramSub.IsSet("values."+valName+".applyConfig") && paramSub.IsSet("values.*.applyConfig") {
+					// Use wildcard fallback
+					valName = "*"
+				}
+
+				log.Debugf("applying custom param %q value %q", k, valName)
+
+				valSub := paramSub.Sub("values." + valName + ".applyConfig")
```
alban, line 305: With the current code, the `applyConfig` section only works when the parameter is in the "custom" group (and not when it is in the "ebpf" group). Shouldn't it be made more generic so that parameters belonging to the ebpf operator can also apply some config changes? (It would require something like the Owner field mentioned above).

pkg/operators/oci-handler/oci.go
flyth: `key` is not being used, instead the actual object key is, right? I don't think we should consider key as a valid field to avoid this ambiguity.

pkg/operators/oci-handler/oci.go
flyth: I'm not sure about allowing this behavior also in other sections. I'd rather have the option to hide ebpf (or other) params and then still set their values using the custom params (`--print-ustack` (custom param) setting `--collect-ustack` (ebpf param), even if hidden - I know this example doesn't make sense 😄).

Otherwise, I think this would become even more complex than it already is.

pkg/operators/oci-handler/oci.go
blanquicet: It looks redundant to have two keys; can't we get rid of one of them? For instance, remove the `key` from the yaml and store the yaml object key as the [`Param.Key` value](https://github.com/inspektor-gadget/inspektor-gadget/blob/michael/oci/decl-params/pkg/gadget-service/api/api.pb.go#L982).

docs/gadget-devel/parameters.md
```diff
@@ -30,3 +32,53 @@ params:
...
+of the gadget. Depending on the parameters' value, a `patch` node will be applied
+to the gadget metadata itself.
+
+Strings used can use templates (explained [here](https://pkg.go.dev/text/template)).
+Functions available to the templating engine are:
+
+#### getParamValue(paramName)
+
+Can be used to get the value of any set parameter; paramName needs to be fully qualified, e.g.
+`custom.mySecondParam`.
```
blanquicet: Please add another example for other params like `params.ebpf.target_family`.

docs/gadget-devel/parameters.md
```diff
@@ -30,3 +32,53 @@ params:
...
+            annotations:
+              my.annotation: >-
+                A template driven value, now {{call .getParamValue "custom.mySecondParam"}}.
+              my.second.annotation: >-
+                We can also access other configuration values, like the gadget name: {{call .getConfig "name"}}
+```
+
+All keys below `custom` will become new parameters that can be set by the user
+of the gadget. Depending on the parameters' value, a `patch` node will be applied
+to the gadget metadata itself.
```
blanquicet: I think we should explain here the [two mechanism](https://github.com/inspektor-gadget/inspektor-gadget/blob/6b632e6d41eb39966b29028fb90b335817beeec6/pkg/gadget-context/gadget-context.go#L265) to use it:

1) Apply a different patch based on parameter's value. Example with `myCustomParam`
2) Apply the same patch and just change the value. Example with `mySecondParam`

docs/gadget-devel/parameters.md
blanquicet: Remember to please apply this.

docs/gadget-devel/parameters.md
blanquicet: Currently, we only have `ebpf` and `custom`. Said that, I prefer grouping them into those two groups.

docs/gadget-devel/parameters.md
```diff
@@ -30,3 +32,53 @@ params:
       defaultValue: "false"
       description: Description for the param
 ```
+
+## Parameters from metadata (gadget.yaml)
```
blanquicet:
```suggestion
## Customizable parameters
```

ebpf and custom params are defined in the `gadget.yaml` so I don't know if "Parameters from metadata (gadget.yaml)" is a good naming. Maybe Dynamic or Customizable is a better name.

docs/gadget-devel/parameters.md
```diff
@@ -30,3 +32,53 @@ params:
...
+```
+
+All keys below `custom` will become new parameters that can be set by the user
+of the gadget. Depending on the parameters' value, a `patch` node will be applied
+to the gadget metadata itself.
+
+Strings used can use templates (explained [here](https://pkg.go.dev/text/template)).
+Functions available to the templating engine are:
+
+#### getParamValue(paramName)
```
blanquicet:
```suggestion
### getParamValue(paramName)
```

Same for `getConfig`.

docs/gadget-devel/parameters.md
```diff
@@ -30,3 +32,53 @@ params:
...
+      defaultValue: "option1"
+      values:
+        option1:
+          patch:
+            datasources:
+              myDataSource:
+                annotations:
+                  columns.output: none
+    mySecondParam:
+      patch:
```
blanquicet, line 62: Please add `description` (and `defaultValue`?) fields. It's a good practice and we want to promote that in the documentation

docs/gadget-devel/parameters.md
alban: In https://github.com/inspektor-gadget/inspektor-gadget/pull/4314#discussion_r2298241264, we have an example of **operator** parameter (`--symbolizers=debuginfod-client-cache` or `--symbolizers=debuginfod-client-cache-on-ig-server` or both) that would need to update two **ebpf** parameters (`--collect-build-id=true` and `--ig-build-id-max-entries=1024` (if not already set to a non-zero value explicitly separately by the user)).

But none of those parameters are specific to a gadget: the **operator** parameter would not be described in gadget.yaml, and the **ebpf** parameters are defined in `include/gadget/user_stack_map.h` and hopefully we won't need to include them in all gadgets' gadget.yaml that include this header file (like it is done with  [wellKnownParams](https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.43.0/pkg/operators/ebpf/params.go#L121)).

Would custom parameters be able to help in this case?

flyth: Currently, I don't think it can help with that specific scenario.

a) currently this _requires_ defining a custom parameter as trigger for patching
b) (other) parameters are especially tricky to patch, because they've been consumed already at the time custom params are being handled; with additional functionality for the gadget.yaml we could potentially work around that by just defining default values in there.

I already tried extending this PR in a couple of more generic directions, but that requires much more work and thought.

It's good to know another use case like yours, though, so we can consider those later on.

pkg/gadget-context/gadget-context.go
```diff
@@ -232,18 +236,116 @@ func (c *GadgetContext) SetParams(params []*api.Param) {
...
+	params := make([]*api.Param, 0)
+
+	// Safeguard
+	if c.paramValues == nil {
+		c.paramValues = make(api.ParamValues)
+	}
+
+	paramValues := c.paramValues
+	customParams := v.Sub(customParamsKey)
+	if customParams != nil {
```
blanquicet: I separated this logic and created some tests for it in https://github.com/inspektor-gadget/inspektor-gadget/commit/a5c0e22fc5353d2cc8426db74fd29b96f825bbee. There are a couple of tests that I think should fail, but, tbh, most of those errors should happen at build time. How can we do that?

blanquicet (commented):
Looks good. I tested it and it's working as expected. Please check the tests I suggested.

Question: Is there any use case you think we're missing with the current implementation? I didn't find any from my tests.

docs/gadget-devel/parameters.md
```diff
@@ -30,3 +32,53 @@ params:
       defaultValue: "false"
       description: Description for the param
 ```
+
+## Parameters from metadata (gadget.yaml)
+
+Much of Inspektor Gadget's functionality is controlled by parameters and
+annotations. To give gadget authors more freedom, one can define new parameters
+that affect a group of other parameters or annotations by patching the
```
blanquicet: I suggest providing more examples of what people can do with this:

affect a group of other parameters, data sources and field annotations, operator's configuration or any other field in the metadata file ...

blanquicet (approved):
LGTM!

pkg/gadget-context/gadget-context.go
flyth: Thanks! We could add some plausibility checks at build time, but I'd like to postpone that a bit until we've seen how this feature will be adopted and extended.

flyth:
> Question: Is there any use case you think we're missing with the current implementation? I didn't find any from my tests.

* [this one](https://github.com/inspektor-gadget/inspektor-gadget/pull/3485#discussion_r2300366027) from @alban could be relevant (basically patching based on the values of _other_/external parameters, not specifically the custom ones)
* we might think of a more generic templating solution that could always be applied
* default values for params is a thing that would be nice to be able to set in the gadget.yaml and then overridden using other params

I'm sure there will be even more once we start using it 😄

flyth:
Thanks for the reviews and keeping this alive 😄 !

PR #3493, opened 2024-09-16
docs: Improve trace_oomkill guide

mqasimsarfraz:
Improve trace_oomkill guide. Upstream References:
- https://github.com/kubernetes/kubernetes/pull/126096
- https://kubernetes.io/docs/concepts/architecture/cgroups/#using-cgroupv2

### DIFF
--- a/gadgets/trace_oomkill/README.mdx
+++ b/gadgets/trace_oomkill/README.mdx
@@ -32,7 +32,16 @@ Running the gadget:
 
 <Tabs groupId="env">
     <TabItem value="kubectl-gadget" label="kubectl gadget">
-        Start by creating a namespace and a deployment:
+        Kubernetes allows you to set memory limits on containers. When a container exceeds its memory limit, the kernel's out-of-memory (OOM) killer is invoked, which kills the process requesting more memory than allowed in such a container.
+        Before Kubernetes version `1.28.0`, the OOM kill events were only generated if the main container process (PID 1) was OOM killed. If any other process in the container was OOM killed, it would remain "invisible" to Kubernetes, meaning users wouldn't be notified.
+        Even after version `1.28.0`, OOM kill events will not be generated if [cgroup v2 isn't enabled](https://kubernetes.io/docs/concepts/architecture/cgroups/#using-cgroupv2) or if [singleProcessOOMKill](https://github.com/kubernetes/kubernetes/pull/126096) is being used.
+
+        To ensure you are notified whenever any process within a container is OOM killed, regardless of whether it is PID 1 or not, you can use the `trace_oomkill` gadget.
+        To demonstrate it, let's start by creating a deployment in a namespace:
+
+        :::note
+        This guide uses Kubernetes version `1.27.0` to ensure that OOM kill events (other than main process) are not reported by Kubernetes.
+        :::
 
         ```bash
         $ kubectl create namespace oomkill-demo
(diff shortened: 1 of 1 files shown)


gadgets/trace_oomkill/README.mdx
```diff
@@ -32,7 +32,14 @@ Running the gadget:
...
-        Start by creating a namespace and a deployment:
+        Kubernetes allows you to set memory limits on containers. When a container exceeds its memory limit, the kernel's out-of-memory (OOM) killer is invoked, which selects a process to kill to free up memory.
+        Normally Kubernetes allows to check for OOM events only at the pod level so if a process is killed by the OOM killer, it is not possible to know which process was killed especially if the container has multiple processes running.
+        This gadget allows you to trace the OOM kill events at process level. To demonstrate this, start by creating a deployment in a namespace:
+
+        :::note
+        The Kubernetes version used in this guide is `1.27.0` since `1.28.0` [introduced cgroup aware OOM killing](https://github.com/kubernetes/kubernetes/pull/117793) when possible.
+        But even with `1.28.0+`, the OOM kill event might be missed if [cgroup v2 isn't enabled](https://kubernetes.io/docs/concepts/architecture/cgroups/#using-cgroupv2) or perhaps [singleProcessOOMKill](https://github.com/kubernetes/kubernetes/pull/126096) is used.
+        :::
```
blanquicet, line 42: I suggest putting this info together. Something like:

```
Before Kubernetes version `1.28.0`, the OOM kill events were only generated if the main container process (PID 1) was OOM killed. 
If any other process in the container was OOM killed, it would remain "invisible" to Kubernetes, meaning users wouldn't be notified. 
However, even after version `1.28.0`, OOM kill events will not be generated if [cgroup v2 isn't enabled](https://kubernetes.io/docs/concepts/architecture/cgroups/#using-cgroupv2)
or if [singleProcessOOMKill](https://github.com/kubernetes/kubernetes/pull/126096) is being used.

To ensure you are notified whenever any process within a container is OOM killed, regardless of whether it is PID 1 or not, you can use the `trace_oomkill` gadget. 
To demonstrate it, let's start by creating a deployment in a namespace:

:::note
This guide uses Kubernetes version `1.27.0` as an example of scenario were Kubernetes doesn't generate OOM kill events for processes differente than PID 1.
:::
```

gadgets/trace_oomkill/README.mdx
```diff
@@ -65,7 +72,29 @@ Running the gadget:
...
+        ```
+
+        or check the `lastState` of the pod:
+
+        ```bash
+        $ kubectl get pod -l app=oomkill-demo -n oomkill-demo -o jsonpath="{.items[*].status.containerStatuses[*].lastState}"
+        {}
+        ```
+
+        Since the OOM kill event is not reported by Kubernetes, we can use the gadget to trace the OOM kill event at process level. So, go back to *the first terminal* and see:
```
blanquicet, line 96: I suggest no to use the terminology "pod level" and "process level", but something simpler:

```suggestion
        Even if the OOM kill event is not reported by Kubernetes, it's still notified by the `trace_oomkill` gadget. So, go back to *the first terminal* and see:
```

blanquicet (approved):
Great improvement! I made a couple of suggestions trying to make it clearer, but feel free to drop them or take only the ones you agree with as it LGTM already.

blanquicet (approved):
LGTM!

PR #3495, opened 2024-09-17
docs: Fix operators instance params

mqasimsarfraz:
As seen [in the example](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/examples/gadgets/operators/local_manager/main.go#L76) the fully qualified names should be `operator.` instead of `operators.`


mauriciovasquezbernal (approved):
Thanks for catching this up!

Issue #3467, opened 2024-09-10
[RFE] Reporting container pid for trace exec gadget


Sefi4:
Add for built-in trace exec and image-based trace-exec gadgets the ability to report the container pid for process events which originated from that container

alban:
I started #3501 but I will not have time to complete it this month. If someone wants to pick it up, feel free to do it.

PR #3501, opened 2024-09-19
Container enrichment: add container PID

alban:
# Container enrichment: add container PID

- [X] Image-based gadgets :+1:
- [x] Builtin gadgets :+1:

## How to use

Read the `runtime.ContainerPid` field.

## Testing done

Tested with command:
```
$ docker run -ti --rm --name test ubuntu sh -c 'cat /dev/null'
```

Output for image-based gadgets:
```
$ sudo ig run ghcr.io/inspektor-gadget/gadget/trace_exec:latest --fields runtime.ContainerName,runtime.ContainerPid,pid,comm,args -c test
RUNTIME.CONTAINERNAME     RUNTIME.CONTAINER…        PID COMM         ARGS
test                                 2911217    2911217 sh           /usr/bin/sh -c cat /dev/null
test                                 2911217    2911269 cat          /usr/bin/cat /dev/null
```

Output for builtin gadgets:
```
$ sudo ig trace exec -o columns=runtime.ContainerName,runtime.ContainerPid,pid,comm,args -c test
RUNTIME.CONTAINERNAME                                                                 RUNTIME.C… PID        COMM             ARGS
test                                                                                  33748      33748      sh               /usr/bin/sh -c cat /dev/null
test                                                                                  33748      33770      cat              /usr/bin/cat /dev/null
```

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/3467

### DIFF
--- a/pkg/container-collection/k8s.go
+++ b/pkg/container-collection/k8s.go
@@ -184,9 +184,16 @@ func (k *K8sClient) GetRunningContainers(pod *v1.Pod) []Container {
 
 		containerDef := Container{
 			Runtime: RuntimeMetadata{
-				BasicRuntimeMetadata: containerData.Runtime.BasicRuntimeMetadata,
+				BasicRuntimeMetadata: types.BasicRuntimeMetadata{
+					RuntimeName:          containerData.Runtime.RuntimeName,
+					ContainerID:          containerData.Runtime.ContainerID,
+					ContainerName:        containerData.Runtime.ContainerName,
+					ContainerPID:         uint32(pid),
+					ContainerImageName:   containerData.Runtime.ContainerImageName,
+					ContainerImageDigest: containerData.Runtime.ContainerImageDigest,
+					ContainerStartedAt:   containerData.Runtime.ContainerStartedAt,
+				},
 			},
-			Pid: uint32(pid),
 			K8s: K8sMetadata{
 				BasicK8sMetadata: types.BasicK8sMetadata{
 					Namespace:     pod.GetNamespace(),
--- a/pkg/datasource/compat/wrapper.go
+++ b/pkg/datasource/compat/wrapper.go
@@ -46,6 +46,7 @@ type EventWrapperBase struct {
 	containernameAccessor        datasource.FieldAccessor
 	runtimenameAccessor          datasource.FieldAccessor
 	containeridAccessor          datasource.FieldAccessor
+	containerpidAccessor         datasource.FieldAccessor
 	containerimagenameAccessor   datasource.FieldAccessor
 	containerimagedigestAccessor datasource.FieldAccessor
 	hostNetworkAccessor          datasource.FieldAccessor
@@ -238,7 +239,7 @@ func WrapAccessors(source datasource.DataSource, mntnsidAccessor datasource.Fiel
 		datasource.WithAnnotations(map[string]string{
 			metadatav1.TemplateAnnotation: "container",
 		}),
-		datasource.WithOrder(-26),
+		datasource.WithOrder(-27),
 	)
 	if err != nil {
 		return nil, err
@@ -251,7 +252,7 @@ func WrapAccessors(source datasource.DataSource, mntnsidAccessor datasource.Fiel
 			metadatav1.ColumnsFixedAnnotation: "true",
 		}),
 		datasource.WithFlags(datasource.FieldFlagHidden),
-		datasource.WithOrder(-25),
+		datasource.WithOrder(-26),
 	)
 	if err != nil {
 		return nil, err
@@ -264,6 +265,17 @@ func WrapAccessors(source datasource.DataSource, mntnsidAccessor datasource.Fiel
 			metadatav1.ColumnsMaxWidthAnnotation: "64",
 		}),
 		datasource.WithFlags(datasource.FieldFlagHidden),
+		datasource.WithOrder(-25),
+	)
+	if err != nil {
+		return nil, err
+	}
+	ev.containerpidAccessor, err = runtime.AddSubField(
+		"containerPid",
+		api.Kind_Uint32,
+		datasource.WithAnnotations(map[string]string{
+			metadatav1.TemplateAnnotation: "containerPid",
+		}),
 		datasource.WithOrder(-24),
 	)
 	if err != nil {
@@ -368,6 +380,10 @@ func (ev *EventWrapper) SetPodMetadata(container types.Container) {
 }
 
 func (ev *EventWrapper) SetContainerMetadata(container types.Container) {
+	if ev.containerpidAccessor.IsRequested() {
+		ev.containerpidAccessor.PutUint32(ev.Data, container.ContainerPid())
+	}
+
 	ev.SetPodMetadata(container)
 }
 
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -33,6 +33,7 @@ func init() {
 	columns.MustRegisterTemplate("namespace", "width:30")
 	columns.MustRegisterTemplate("pod", "width:30,ellipsis:middle")
 	columns.MustRegisterTemplate("container", "width:30")
+	columns.MustRegisterTemplate("containerPid", "width:6,hide")
 	columns.MustRegisterTemplate("containerImageName", "width:30")
 	columns.MustRegisterTemplate("containerImageDigest", "width:30")
 	columns.MustRegisterTemplate("containerStartedAt", "width:35,hide")
@@ -111,6 +112,7 @@ type Container interface {
 	RuntimeMetadata() *BasicRuntimeMetadata
 	UsesHostNetwork() bool
 	K8sOwnerReference() *K8sOwnerReference
+	ContainerPid() uint32
 }
 
 type BasicRuntimeMetadata struct {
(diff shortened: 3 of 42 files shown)


pkg/datasource/compat/wrapper.go
```diff
@@ -259,6 +260,20 @@ func WrapAccessors(source datasource.DataSource, mntnsidAccessor datasource.Fiel
...
+	if err != nil {
+		return nil, err
+	}
+	ev.containerpidAccessor, err = runtime.AddSubField(
+		"containerPid",
+		api.Kind_Uint32,
+		datasource.WithAnnotations(map[string]string{
+			datasource.ColumnsWidthAnnotation:     "6",
+			datasource.ColumnsAlignmentAnnotation: "right",
+			datasource.ColumnsMaxWidthAnnotation:  "64",
```
mauriciovasquezbernal, line 274: Does `"template": "containerPid"` work to avoid repeating it again?

pkg/types/types.go
```diff
@@ -127,6 +129,9 @@ type BasicRuntimeMetadata struct {
 	// response with multiple containers, ContainerName contains only the first element.
 	ContainerName string `json:"containerName,omitempty" column:"containerName,template:container"`
 
+	// ContainerPID is the process ID of the first process in the container.
+	ContainerPID uint32 `json:"containerPid,omitempty" column:"containerPid,template:pid,hide"`
+
```
mauriciovasquezbernal, line 134: We already have a PID field in the container structure. I think we need to remove that to avoid duplication. This is the reason why this is not working as well for built-in gadgets, this field is never been set:

```diff
$ git diff
diff --git a/pkg/container-collection/container-collection.go b/pkg/container-collection/container-collection.go
index 2d9e87fa4..b4b7c2f59 100644
--- a/pkg/container-collection/container-collection.go
+++ b/pkg/container-collection/container-collection.go
@@ -210,6 +210,7 @@ func (cc *ContainerCollection) AddContainer(container *Container) {
                }
        }

+       container.Runtime.ContainerPID = container.Pid
        _, loaded := cc.containers.LoadOrStore(container.Runtime.ContainerID, container)
        if loaded {
                return
```

alban: Thanks, it works with your patch. But I don't see any duplication in the yaml output:

```bash
$ sudo ig trace exec -o yaml -c test
```
```yaml
---
args:
- /usr/bin/sh
- -c
- cat /dev/null
comm: sh
gid: 0
group: root
k8s:
  owner: {}
loginuid: 4294967295
mountnsid: 4026532891
pcomm: containerd-shim
pid: 45296
ppid: 45273
pupperlayer: false
runtime:
  containerId: 7ba96e85c2df96108036c73379b513d2797b2ea43f2fbdc0888981cf55708353
  containerImageName: ubuntu
  containerName: test
  containerPid: 45296
  containerStartedAt: 1729592047995532764
  runtimeName: docker
sessionid: 4294967295
timestamp: 1729592048136816332
type: normal
uid: 0
upperlayer: false
user: root

---
args:
- /usr/bin/cat
- /dev/null
comm: cat
gid: 0
group: root
k8s:
  owner: {}
loginuid: 4294967295
mountnsid: 4026532891
pcomm: sh
pid: 45319
ppid: 45296
pupperlayer: false
runtime:
  containerId: 7ba96e85c2df96108036c73379b513d2797b2ea43f2fbdc0888981cf55708353
  containerImageName: ubuntu
  containerName: test
  containerPid: 45296
  containerStartedAt: 1729592047995532764
  runtimeName: docker
sessionid: 4294967295
timestamp: 1729592048138417372
type: normal
uid: 0
upperlayer: false
user: root
```

pkg/datasource/compat/wrapper.go
alban: It seems to work. So I can remove the hidden flag because it automatically comes from the template.

pkg/types/types.go
mauriciovasquezbernal: I wasn't clear enough, sorry. I meant that the container.Pid field

https://github.com/inspektor-gadget/inspektor-gadget/blob/67f2288c9d774037ece7bcb25eacb7a9ec7ed618/pkg/container-collection/containers.go#L41-L42

and the container.Runtime.ContainerPID field represents the same information:

https://github.com/inspektor-gadget/inspektor-gadget/blob/67f2288c9d774037ece7bcb25eacb7a9ec7ed618/pkg/types/types.go#L132-L133

So probably we want to remove the former and leave only the later.

mauriciovasquezbernal (approved):
Besides the duplicated pid field it LGTM!

pkg/types/types.go
```diff
@@ -216,6 +221,7 @@ func (c *CommonData) SetContainerMetadata(container Container) {
 	c.Runtime.RuntimeName = runtime.RuntimeName
 	c.Runtime.ContainerName = runtime.ContainerName
 	c.Runtime.ContainerID = runtime.ContainerID
+	c.Runtime.ContainerPID = runtime.ContainerPID
 	c.Runtime.ContainerImageName = runtime.ContainerImageName
 	c.Runtime.ContainerImageDigest = runtime.ContainerImageDigest
 	c.Runtime.ContainerStartedAt = runtime.ContainerStartedAt
```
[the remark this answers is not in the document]
alban: No idea. I will try that.

pkg/types/types.go
alban: Thanks for the clarification. I updated the PR.

pkg/types/types.go
```diff
@@ -127,6 +129,9 @@ type BasicRuntimeMetadata struct {
 	// response with multiple containers, ContainerName contains only the first element.
 	ContainerName string `json:"containerName,omitempty" column:"containerName,template:container"`
 
+	// ContainerPID is the process ID of the first process in the container.
+	ContainerPID uint32 `json:"containerPid,omitempty" column:"containerPid,template:pid,hide"`
```
blanquicet, line 133:
```suggestion
	ContainerPID uint32 `json:"containerPid,omitempty" column:"containerPid,template:containerPid"`
```

pkg/datasource/columns.go
```diff
@@ -272,6 +272,10 @@ var annotationsTemplates = map[string]map[string]string{
 	"containerImageName": {
 		ColumnsWidthAnnotation: "30",
 	},
+	"containerPid": {
+		ColumnsWidthAnnotation:  "6",
+		ColumnsHiddenAnnotation: "true",
```
blanquicet, line 277:
```suggestion
		TemplateAnnotation:  "pid",
		ColumnsHiddenAnnotation: "true",
```

If not, please add `ColumnsAlignmentAnnotation: string(metadatav1.AlignmentRight)` like other numeric fields.

blanquicet (approved):
LGTM from code inspection.

alban:
@blanquicet Thanks for the review! I struggle to make the tests in `pkg/container-utils/runtime-client/runtimeclient_test.go` pass and I see you wrote those tests initially. Following Mauricio's [request](https://github.com/inspektor-gadget/inspektor-gadget/pull/3501#discussion_r1810928069) to remove the duplicate field, I tried to initialize the ContainerPid field in:
- pkg/container-utils/containerd/containerd.go
- pkg/container-utils/docker/docker.go
- pkg/container-utils/podman/podman.go
- pkg/container-utils/runtime-client/runtimeclient_test.go

But I am not sure if it should be done for GetContainer, GetContainerDetails and GetContainers, or only some of them, or none of them. Do you know?

blanquicet:
> But I am not sure if it should be done for GetContainer, GetContainerDetails and GetContainers, or only some of them, or none of them. Do you know?
>
Oh, I see. `GetContainer` and `GetContainers` return `ContainerData` which doesn't contain the container PID. It's because getting that PID, in some of the runtime implementations (e.g., CRI and Docker) requires an extra gRPC call that we didn't want to do because PID is not always required. For cases where PID is required, `ContainerRuntimeClient`'s consumers use `GetContainerDetails`, which makes that extra gRPC call to retrieve the PID from the runtime.

Following that logic, `ContainerPid` should only be set on `GetContainerDetails`. However, now that you're moving the PID into `BasicRuntimeMetadata`, it implicitly goes into `ContainerData`, and it's not correct that consumers calling `GetContainer` and `GetContainers` will now have access to `ContainerData.Runtime.ContainerPID` but it'll be always empty.

Only one simple solution came to my mind is to prevent `ContainerData` from re-using `BasicRuntimeMetadata`. In this way, your change on `BasicRuntimeMetadata` will not affect `ContainerData` thus, `GetContainer` and `GetContainers` will continue not returning the PID. However, it also implies that:

1. Every time we add something to `BasicRuntimeMetadata`, we will need to manually add it into `ContainerData`... But that's fine, we should only add things to `ContainerData` once the `ContainerRuntimeClient` supports it.
2. We'll have to copy the fields one by one when enriching containers with `ContainerData`, but [we are already doing that](https://github.com/inspektor-gadget/inspektor-gadget/blob/e3350a50135e092569c654040baea750b8ba75d1/pkg/container-collection/options.go#L55-L74) to ensure we don't set empty fields. 

So, it seems fine to decouple `BasicRuntimeMetadata` from `ContainerData`. WDYT?

pkg/container-collection/k8s.go
```diff
@@ -176,17 +175,18 @@
...
-		}
-
 		containerDef := Container{
 			Runtime: RuntimeMetadata{
-				BasicRuntimeMetadata: containerData.Runtime.BasicRuntimeMetadata,
+				BasicRuntimeMetadata: types.BasicRuntimeMetadata{
+					RuntimeName:          containerData.Runtime.RuntimeName,
+					ContainerID:          containerData.Runtime.ContainerID,
+					ContainerName:        containerData.Runtime.ContainerName,
+					ContainerPID:         uint32(containerData.Pid),
```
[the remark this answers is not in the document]
alban: Thanks, it should be fixed now.

alban:
Thanks @blanquicet for the help!

The tests finally pass.

@mauriciovasquezbernal Do we need something for the FOSSA hooks, or could I merge this?

alban:
I resolved the conflict with #3600.

mauriciovasquezbernal:
> @mauriciovasquezbernal Do we need something for the FOSSA hooks, or could I merge this?

I see they're passing, probably it was a temporary false positive.

Issue #3195, opened 2024-07-22
image-based gadgets: Implement unit tests


mauriciovasquezbernal:
Currently we are running integration tests for image-based gadgets (thanks to #2046) and unit tests for built-in gadgets (see #817), however we don't run unit tests for image-based gadgets yet. 

### Motivation 

Integration tests are a great way to check that everything is working fine, however it's difficult to check a lot of specific details in those tests. The unit tests are ideal to check that:
- The gadget provides the correct information for all fields 
- The filtering mechanism works fine (by mount namespace or other things like uid, ports, etc) 
- The gadget flags change the gadget's behavior 
- The gadget works on different kernel versions (see #1343) 

### Implementation Options 

For this support we need the following components:

- A mechanism to generate events: The existing one on https://github.com/inspektor-gadget/inspektor-gadget/tree/v0.30.0/internal/test can be used 
- A mechanism to match the result: It's already done in https://github.com/inspektor-gadget/inspektor-gadget/tree/v0.30.0/pkg/testing/match 
- A mechanism to run the gadget: It's possible to run a gadget by using the Golang API as done in https://github.com/inspektor-gadget/inspektor-gadget/tree/v0.30.0/examples/gadgets/simple/trace_open. It should be possible to create a wrapper (or even an operator) around the Golang API for testing purposes so users don't have to repeat all the same code in all tests.

btwshivam:
Hi @mauriciovasquezbernal 
I hope this message finds you well. My name is Shivam, and I am excited to contribute to the implementation of unit tests for image-based gadgets under the LFX mentorship program.

I am eager to enhance the reliability of these gadgets and would appreciate your guidance on where to start. Could you please suggest some issues to begin with?

HitanshuPrasad:
hi @mauriciovasquezbernal @alban ,  Interested in this project for fall term LFX. excited to solve such a problem. Thanks!

mauriciovasquezbernal:
> Could you please suggest some issues to begin with?

Sure, I opened some new issues about missing **integration** tests we have. I think they're a good starting point to get familiar with this tasks
- https://github.com/inspektor-gadget/inspektor-gadget/issues/3235
- https://github.com/inspektor-gadget/inspektor-gadget/issues/3240

Ali-Alnosairi:
Hey @alban @mauriciovasquezbernal,
[...]

PR #3502, opened 2024-09-19
Unit testing Image Based Gadgets

mastersans:
# Unit Testing for Image based gadgets.

fixes: #3195

This PR introduces new changes in which unit-testing capability for image based gadgets are added which leverages newly added `gadgetrunner` which is a wrapper on top of golang api (simple operator used) which can generate event before and in between the gadget running process and captured events can be validated against the expected output using the pre existing matching function in utilstest package and runner being used for event generation.
## How to use

To try out the newly added unit test for image based gadget you can use following  cmd: `go test -exec 'sudo -E' -v` 
Inside any `unit` directory that is present in the `test` directory for some gadgets, Below is an example.
```bash
mastersans@sanskar-vivobook:~/inspektor-gadget/gadgets/top_file/test/unit$ go test -exec 'sudo -E' -v
```

more unit test for different other gadgets will be added in future. Currently added are:

- Trace_Open
- Trace_Tcp
- Top_File
- Snapshot_Process

Works needs to be done on PR:
Older TODO:
- [x] Fix problems while running Parallel test.
- [x] Improve gadgetapi to support running of different gadgets
- [ ] Work on respective Makefile and Yaml file. 
- [x] Port different buildin gadgets unittest to image based once
- [ ] Documentation.

## Testing done

Testing is done locally with same procedure as mentioned before and for CI respective Makefile target and Workflow is added and tested on Fork.


pkg/testing/gadgetapi/gadgetapi.go
```diff
@@ -0,0 +1,121 @@
...
+	gadgetOperator := simple.New("gadget", simple.OnInit(func(gadgetCtx operators.GadgetContext) error {
+		fmt.Println("Initializing gadget operator...")
+
+		for _, d := range gadgetCtx.GetDataSources() {
+			jsonFormatter, _ := igjson.New(d)
+
+			d.Subscribe(func(source datasource.DataSource, data datasource.Data) error {
+				jsonOutput := jsonFormatter.Marshal(data)
+				outputCh <- string(jsonOutput)
+				fmt.Printf("Captured data: %s\n", jsonOutput)
```
mauriciovasquezbernal, line 57: I think we should be providing the raw `source datasource.DataSource, data datasource.Data` to the gadget. 

Perhaps `NewGadget` could have a `source datasource.DataSource, data datasource.Data` callback argument, then the each gadget can decide how to process the data, for instance, to marhsall to json.

However I'm wondering if we should go a step back, perhaps this wrapper is creating some distraction, what about moving all this logic to unit/trace_open_test.go, once we have test working, try to implement tests for few more gadgets and then see how we can create this common wrapper?

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			api.StartGadget(params)
+			runner := utilstest.NewRunnerWithTest(t, testCase.runnerConfig)
+			utilstest.RunWithRunner(t, runner, func() error {
+				testCase.generateEvent()
+				// Wait for the event to be processed
+				time.Sleep(1 * time.Second)
+				output := api.Output()
+				match.MatchEntries(t, match.JSONSingleArrayMode, string(output), testCase.normalize, testCase.expectedEntry)
```
mauriciovasquezbernal, line 118: In the issue #3195 I proposed to use the [match](https://github.com/inspektor-gadget/inspektor-gadget/tree/v0.30.0/pkg/testing/match) package to validate events, however I'm not sure it's a good fit for unit tests. As I mentioned above, one big different between unit and integration tests is that in the former we want to be as exhaustive as possible, it means to check fields that change at runtime, like pid, tid, fd, etc. The internal/test/helpers.go package implements functions that are more flexible because they receive some callbacks, so perhaps they're better for this purpose.

pkg/testing/gadgetapi/gadgetapi.go
mastersans: Yeah its best to do it that way, lets get it done in future commits.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	ocihandler "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/oci-handler"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/simple"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/local"
+
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf"
+
+	igjson "github.com/inspektor-gadget/inspektor-gadget/pkg/datasource/formatters/json"
```
mauriciovasquezbernal, line 38: nit: All of these can go to the same group.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+			<-subscribed
+			time.Sleep(1 * time.Second)
+			close(subscribed)
+			var fd int
+			runner := utilstest.NewRunnerWithTest(t, testCase.runnerConfig)
+			utilstest.RunWithRunner(t, runner, func() error {
+				fd, _ = testCase.generateEvent()
+				return nil
+			})
```
mauriciovasquezbernal, line 142: The simple operator supports to be called onStart() too. What do you think about generating the events on the onStart() hook and remove some synchronization from this test?

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,168 @@
...
+					}
+				})(t, info, fd, events)
+				return nil
+			},
+		},
+	}
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			var wg sync.WaitGroup
+			events := make(chan ExpectedTraceOpenEvent, 100)
```
mauriciovasquezbernal, line 82: nit: You can use a normal slice and use append inside the subcribed function.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,168 @@
...
+				if err := runtime.Init(nil); err != nil {
+					fmt.Errorf("runtime init: %w", err)
+				}
+				params := map[string]string{
+					// Filter only events from the root user
+					"operator.oci.ebpf.uid": "0",
+				}
+
+				if err := runtime.RunGadget(gadgetCtx, nil, params); err != nil {
+					fmt.Errorf("running gadget: %w", err)
```
mauriciovasquezbernal, line 128: nit: We can use directly `require.NotError` to check that a function returned nil.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,143 @@
...
+				processId = pid
+				return nil
+			})
+			timeout := 5 * time.Second
+			const opPriority = 50000
+
+			gadgetOperator := simple.New("gadget",
+				simple.OnInit(func(gadgetCtx operators.GadgetContext) error {
+					for _, d := range gadgetCtx.GetDataSources() {
+						jsonFormatter, _ := igjson.New(d)
```
mauriciovasquezbernal, line 92: I think it's better to check the error to be totally sure it's nil.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,271 @@
...
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, fd int, events []ExpectedTraceOpenEvent) error {
+				utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, fd int) *ExpectedTraceOpenEvent {
+					return &ExpectedTraceOpenEvent{
+						Comm:  "unit.test",
+						Pid:   info.Pid,
+						Tid:   info.Tid,
+						Uid:   uint32(info.Uid),
+						Gid:   uint32(info.Gid),
+						Fd:    uint32(fd),
+						FName: "../../../../../../../tmp/test_relative_path",
```
mauriciovasquezbernal, line 135: this one fails for me. Should it use `generateRelativePathForAbsolutePath` instead of hardcoding the path?

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,143 @@
...
+			generateEvent: generateEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, fd int, events []ExpectedSnapshotProcessEvent) error {
+				utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, sleepPid int) *ExpectedSnapshotProcessEvent {
+					return &ExpectedSnapshotProcessEvent{
+						Comm:      "sleep",
+						Pid:       sleepPid,
+						Tid:       sleepPid,
+						Uid:       0,
+						Gid:       0,
+						ParentPid: info.Pid,
```
mauriciovasquezbernal, line 68: It's confusing but we need to use `info.Tid` here.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,143 @@
...
+					gadgetOperator,
+				),
+				gadgetcontext.WithTimeout(timeout),
+			)
+			runtime := local.New()
+			err := runtime.Init(nil)
+			require.NoError(t, err, "initializing runtime")
+			params := map[string]string{
+				"operator.oci.ebpf.uid": "0",
+			}
```
mauriciovasquezbernal, line 126: nit: This gadget doesn't support this parameter. Check supported parameters in https://www.inspektor-gadget.io/docs/latest/gadgets/snapshot_process#flags. Or by looking at GADGET_PARAM in the code.

gadgets/trace_open/test/unit/trace_open_test.go
mastersans: @mauriciovasquezbernal I am not sure it is best way or not the older build in were implemented this way, also again not sure the test seems to pass locally when i tested locally but I'll look into it.

mauriciovasquezbernal: I think we need to use `generateRelativePathForAbsolutePath` otherwise it'll be depending on the location of the test on your machine, that's probably why it works for you and not for me.

pkg/operators/formatters/formatters.go
mauriciovasquezbernal, line 1: You can do something like

https://github.com/inspektor-gadget/inspektor-gadget/blob/4f2bbc7c954cdd0a45f77466631420998bb9956b/pkg/operators/cli/clioperator.go#L403

to avoid having to expose the structure type.

mastersans: I'll get it done while porting trace_tcp test to use gadgetapi :+1:

pkg/operators/formatters/formatters.go
```diff
@@ -66,26 +66,26 @@ const (
 	errnoTargetAnnotation     = "formatters.errno.target"
 )
 
-type formattersOperator struct{}
+type FormattersOperator struct{}
```
mauriciovasquezbernal, line 69: Please undo this as indicated in https://github.com/inspektor-gadget/inspektor-gadget/pull/3502#discussion_r1783062319

pkg/operators/simple/options.go
mauriciovasquezbernal, line 1: These changes should go into a separated commit.

internal/test/gadgetapi.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, fmt.Errorf("image name is required")
+	}
+	if imageVersion == "" {
+		imageVersion = "latest"
+	}
+	if timeout < 0 {
+		return nil, fmt.Errorf("invalid timeout value")
+	}
+
+	imageURL := fmt.Sprintf("ghcr.io/inspektor-gadget/gadget/%s:%s", imageName, imageVersion)
```
mauriciovasquezbernal, line 73: do not append the specific inspektor gadget repo, this should be usable by external gadgets as well.

internal/test/gadgetapi.go
```diff
@@ -0,0 +1,178 @@
...
+	apiParams     api.ParamValues
+	runtimeParams *params.Params
+
+	gadgetCtx      *gadgetcontext.GadgetContext
+	gadgetOperator operators.DataOperator
+	DataFunc       datasource.DataFunc
+	DataOperator   []operators.DataOperator
+	JsonFormatter  *igjson.Formatter
+
+	CapturedEvent   []T
```
mauriciovasquezbernal, line 55: CapturedEvents? (with an "s")

internal/test/gadgetapi.go
```diff
@@ -0,0 +1,178 @@
...
+	OnGadgetRun     func(gadgetCtx operators.GadgetContext) error
+	BeforeGadgetRun func() error
+	NormalizeField  func(event *T)
```
mauriciovasquezbernal, line 58: NormalizeEvent?

internal/test/gadgetapi.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// NewGadget initializes a new GadgetApi
+func NewGadget[T any](imageName, imageVersion string, timeout time.Duration, apiParams api.ParamValues, eventtype T, mnsFilterMap *ebpf.Map) (*GadgetApi[T], error) {
```
mauriciovasquezbernal, line 62: Since this is specifically designed for testing purposes, I think it's better to take the testing.T object in this constructor instead of having to pass it to the StartGadget() and other functions. Also this constructor doesn't have to return an error if it takes the testing object.

mauriciovasquezbernal: Also, what's `eventtype` for? can it be removed?

mauriciovasquezbernal: Finally, this function is receiving many parameters, some of them optional, what about creating a options struct:

type NewGadgetOpts struct {
    imageName 
    imageVersion
    ...
}

and only receiving this on NewGadget. It'll make it easier to add more parameters if needed later on.

internal/test/gadgetapi.go
```diff
@@ -0,0 +1,178 @@
...
+func (g *GadgetApi[T]) StartGadget(t *testing.T) {
+	var mu sync.Mutex
+	if g.DataFunc == nil {
+		// Use default data function if none is provided
+		g.DataFunc = func(source datasource.DataSource, data datasource.Data) error {
+			event := new(T)
+			jsonOutput := g.JsonFormatter.Marshal(data)
+			err := json.Unmarshal(jsonOutput, event)
+			require.NoError(t, err, "unmarshalling event")
+			// print the captured event for debugging
```
mauriciovasquezbernal, line 94: Please remove outdated comments.

internal/test/gadgetapi.go
```diff
@@ -0,0 +1,178 @@
...
+			mu.Unlock()
+			return nil
+		}
+	}
+	gadgetOperatorOpts := []simple.Option{
+		simple.OnInit(func(gadgetCtx operators.GadgetContext) error {
+			for _, d := range gadgetCtx.GetDataSources() {
+				jsonFormatter, err := igjson.New(d,
+					igjson.WithShowAll(true),
+					igjson.WithPretty(true, "  "),
```
mauriciovasquezbernal, line 110: Do we need pretty for testing purposes?

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,297 @@
...
+			apiParams := map[string]string{
+				"operator.oci.ebpf.uid": "0",
+			}
+			var mntsFilterMap *ebpf.Map
+			if testCase.mntsFilterMap != nil {
+				mntsFilterMap = testCase.mntsFilterMap(runner.Info)
+			}
+			gdgt, err := gadgetapi.NewGadget(
+				"trace_open",
+				"latest",
```
mauriciovasquezbernal, line 252: Please make this configurable. You can take some inspiration from pkg/testing/ig/ig.go.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,297 @@
...
+			generateEvent: func(t *testing.T) int {
+				err := os.Mkdir("/tmp/foo", 0o750)
+				if err != nil {
+					require.NoError(t, err, "mkdir for /tmp/foo")
+					return 0
+				}
+				defer os.RemoveAll("/tmp/foo")
+
+				fd, err := unix.Open("/tmp/foo/bar.test", unix.O_RDONLY|unix.O_CREAT, 0)
+				if err != nil {
```
mauriciovasquezbernal, line 207: This check is not needed when require.NoError is used. (same in other places)

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,297 @@
...
+	Gid   uint32 `json:"gid"`
+	Fd    uint32 `json:"fd"`
+	FName string `json:"fname"`
+}
+
+type testDef struct {
+	runnerConfig  *utilstest.RunnerConfig
+	mntsFilterMap func(info *utilstest.RunnerInfo) *ebpf.Map
+	generateEvent func(t *testing.T) int
+	validateEvent func(t *testing.T, info *utilstest.RunnerInfo, fd int, events []ExpectedTraceOpenEvent) error
```
mauriciovasquezbernal, line 47: Does it need to return `error`? I think we can remove it.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,297 @@
...
+						Uid:   uint32(info.Uid),
+						Gid:   uint32(info.Gid),
+						Fd:    uint32(fd),
+						FName: "/dev/null",
+					}
+				})(t, info, fd, events)
+				return nil
+			},
+		},
+		"test_flags_and_mode": {
```
mauriciovasquezbernal, line 103: Flags nor mode are being tested. Are they missing from the event struct definition and so on?

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,165 @@
...
+	Uid       uint32 `json:"uid"`
+	Gid       uint32 `json:"gid"`
+	ParentPid int    `json:"ppid"`
+	MountNsID uint64 `json:"mntns_id"`
+}
+
+type testDef struct {
+	runnerConfig   *utilstest.RunnerConfig
+	generateEvent  func() (int, error)
+	validateEvent  func(t *testing.T, info *utilstest.RunnerInfo, fd int, events []ExpectedSnapshotProcessEvent) error
```
mauriciovasquezbernal, line 46: rename fd to the `sleelPid` (or the correct one) for this case.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,165 @@
...
+			runnerConfig:  runnerConfig,
+			generateEvent: generateEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, fd int, events []ExpectedSnapshotProcessEvent) error {
+				if len(events) == 0 {
+					t.Fatalf("no events were captured")
+				}
+				for _, event := range events {
+					if event.Pid != event.Tid {
+						t.Fatalf("thread %d was captured", event.Tid)
+					}
```
mauriciovasquezbernal, line 115: nit: we can use `require.Equal` in this case.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,165 @@
...
+		})
+	}
+}
+func generateEvent() (int, error) {
+	cmd := exec.Command("/bin/sleep", "5")
+	if err := cmd.Start(); err != nil {
+		return 0, fmt.Errorf("running command: %w", err)
+	}
+	return cmd.Process.Pid, nil
+}
```
mauriciovasquezbernal, line 165: Same here, please take testing.T as argument and avoid returning an error to simplify the logic.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,165 @@
...
+				utilstest.RunWithRunner(t, runner, func() error {
+					pid, _ := testCase.generateEvent()
+					processId = pid
+					return nil
+				})
+				return nil
+			}
+
+			gdgt.StartGadget(t)
+			fd := processId
```
mauriciovasquezbernal, line 154: nit: pass this variable directly to the next line to avoid an extra temp variable around.

gadgets/trace_tcp/test/unit/trace_tcp_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+type testDef struct {
+	addr          string
+	port          int
+	runnerConfig  *utilstest.RunnerConfig
+	generateEvent func(t *testing.T, addr string, port int)
+	validateEvent func(t *testing.T, info *utilstest.RunnerInfo, fd int, events []ExpectedTraceTcpEvent) error
+}
+
+func TestTraceOpenGadget(t *testing.T) {
```
mauriciovasquezbernal, line 67: open -> tcp

gadgets/trace_tcp/test/unit/trace_tcp_test.go
mauriciovasquezbernal, line 1: Please update this file to use the latest framework.

mauriciovasquezbernal (changes requested):
It's looking very good. Some comments around. 

Please update the commit structure:
1st commit: introduce "gadget api"
2nd commit: trace open tests
3rd commit: snapshot proces 
and so on.

I need to make another pass on the "gadget api"

mauriciovasquezbernal (changes requested):
Some more comments I forgot to submit.

gadgets/top_file/test/unit/top_file_test.go
```diff
@@ -0,0 +1,162 @@
...
+	file, err := os.Create(filepath)
+	require.NoError(t, err, "creating file")
+
+	defer func() {
+		err := file.Close()
+		require.NoError(t, err, "closing file")
+	}()
+
+	// Set up the command to continuously write to the file with a 5 seconds timeout
+	cmd := exec.Command("bash", "-c", "timeout 5 bash -c 'while true; do echo -n foo > bar; sleep 0.3; done'")
```
mauriciovasquezbernal, line 152: What about writing the file directly with golang? We could have a very presize control of how many bytes we write and we could check this on the test.

mauriciovasquezbernal (commented):
Looks very good. A lot of minor comments to make it even better. 

Please run `make lint` to fix some style issues.

gadgets/trace_tcp/test/unit/trace_tcp_test.go
```diff
@@ -0,0 +1,160 @@
...
+							Version: 4,
+							Port:    9070,
+							Proto:   6,
+						},
+					}
+				})(t, info, fd, events)
+				return nil
+			},
+		},
+		"captures_all_events_2": {
```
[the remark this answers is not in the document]
mastersans: @mauriciovasquezbernal It has a different port number it was mostly to check the test running perfectly with parallel test.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/cilium/ebpf"
+	"github.com/stretchr/testify/require"
+
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	containerutils "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils"
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf"
```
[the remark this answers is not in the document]
mastersans: @mauriciovasquezbernal removing this import seems to throw an error with all the gadgets regarding oci. 
```bash
Error:          Received unexpected error:
                                initializing and preparing operators: instantiating operator "oci": image doesn't contain valid gadget layers
                Test:           TestSnapshotProcessGadget/captures_events_with_no_filter
                Messages:       running gadget error```

File: pkg/testing/gadgetrunner/gadgetrunner.go [hunk 30]
```diff
@@ -0,0 +1,201 @@
...
+			return nil
+		}),
+	}
+	if g.mnsFilterMap != nil {
+		gadgetOperatorOpts = append(gadgetOperatorOpts,
+			// On PreStart set the mount ns filter map
+			simple.OnPreStart(func(gadgetCtx operators.GadgetContext) error {
+				gadgetCtx.SetVar(gadgets.MntNsFilterMapName, g.mnsFilterMap)
+				gadgetCtx.SetVar(gadgets.FilterByMntNsName, true)
+
```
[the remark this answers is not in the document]
mastersans: @mauriciovasquezbernal yeah its to be done in this way, normal test where unset mount namespaces seems to fail, with a nil filter.

gadgets/trace_open/test/unit/trace_open_test.go
```diff
@@ -0,0 +1,298 @@
...
+						ErrRaw:   0,
+					}
+				})(t, info, fd, events)
+			},
+		},
+		"test_relative_path": {
+			runnerConfig: &utilstest.RunnerConfig{},
+			generateEvent: func(t *testing.T) int {
+				relPath := generateRelativePathForAbsolutePath(t, "/tmp/test_relative_path")
+				fd, err := unix.Open(relPath, unix.O_CREAT|unix.O_RDWR, unix.S_IRWXU|unix.S_IRGRP|unix.S_IWGRP|unix.S_IXOTH)
```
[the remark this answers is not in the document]
mastersans: I was done in the similar manner in the previous buildin gadget unit test, unsure on weather its best to do it this way or normally.

gadgets/top_file/test/unit/top_file_test.go
mastersans: I actually tried that out during initially implementing the test, but in my findings the event generated though that approach sometimes gets captured and sometimes act a little flaky, also using it lot of extra events are also generated, bash approach gets captured better, should I shift to using golang?

gadgets/trace_tcp/test/unit/trace_tcp_test.go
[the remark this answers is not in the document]
mauriciovasquezbernal: ok, but at this point I think we can remove it as it's not testing anything different.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Do you still get this error if you put this import in pkg/testing/gadgetrunner/gadgetrunner.go?

gadgets/trace_open/test/unit/trace_open_test.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it should be as "normal" as possible, we're testing the mode and other stuff in a specific test below.

gadgets/snapshot_process/test/unit/snapshot_process_test.go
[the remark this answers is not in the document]
mastersans: it passes now, this import totally went off my notice. :+1:

alban:
Hi! I know updating the documentation is still in your TODO, but could you please update the explanations in the PR description?

* The PR description says "unit testing" but the PR contains both unit tests and integration tests
* Before running the command "go test -exec 'sudo -E' -v":
  * cd gadgets/snapshot_process/test/unit
  * Build the gadget: which command did you use? I got the error "getting signing information" and I didn't see if I can pass parameters like "--verify-image"
  * Alternatively, did you use `export GADGET_REPOSITORY=ghcr.io/inspektor-gadget/gadget/snapshot_process GADGET_TAG=latest`? I still couldn't pass the signature error.

About the import `utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"`: is it planned to move it outside of `internal`? Can it be used for external gadgets? Could you add "update github.com/inspektor-gadget/gadget-template" in your TODO?

mastersans:
Hi @alban as image verification is by default set to true, using the gadgetrunner  will check for verification in contrary to that if you want to turn off the image verification you can set `operator.oci.verify-image: false` in the parameter pass in gadgetrunner but that logs a warning regarding the disabled verification, Yes i checked the image verification part you shared locally, It seems to pass for me so really unsure how to handle it   ; )
 regarding other comments I'll get it done ASAP:
image

mauriciovasquezbernal (commented):
It's looking very good. Some more nits. 

Regarding the image verification, we need to introduce this support directly on the gadget runner. I think we can follow a very similar approach to https://github.com/inspektor-gadget/inspektor-gadget/blob/bd91aa6cf76676f53275ae0ff6ad3d1b6c04dd81/.github/workflows/inspektor-gadget.yml#L1482

In this case, add a specific `IG_VERIFY_IMAGE` env variable that we check on the test to set the `operator.oci.verify-image` when running the gadget.

gadgets/top_file/test/unit/top_file_test.go
```diff
@@ -0,0 +1,159 @@
...
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var filepath string
+			runner := utilstest.NewRunnerWithTest(t, testCase.runnerConfig)
+			params := map[string]string{
+				"operator.oci.ebpf.map-fetch-interval": "1000ms",
+				"operator.LocalManager.host":           "true",
+			}
```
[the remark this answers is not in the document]
mastersans: @mauriciovasquezbernal when i was initailly running the the gadgets using the golang api and also using IG only the container related events were logged so i was using it, but now that you mentioned it I ran the test without these params and a little strange to me that they pass now, so yeah I'll remove them. :+1:

mastersans:
I think this should be a good start for the Makefile and Workflow. Also will fix commit history soon.

gadgets/Makefile
```diff
@@ -101,19 +104,43 @@ clean:
...
+.PHONY: 
+test-unit: build
+	IG_VERIFY_IMAGE=$(IG_VERIFY_IMAGE) \
+	GADGET_TAG=$(GADGET_TAG) \
+	GADGET_REPOSITORY=$(GADGET_REPOSITORY) \
+	go test -v -exec 'sudo -E' ./.../$(UNIT_TEST_DIR)/...
+
+
+.PHONY: 
+%/test-unit: build
```
mauriciovasquezbernal, line 116: %/test-unit: %

To avoid compiling all gadgets when running unit tests for a single one.

(same in other similar cases)

mauriciovasquezbernal (commented):
One quick comment. I'll enable the CI to run, if tests pass it's good to go after you reorganize the commit history.

mauriciovasquezbernal (approved):
I'm approving and merging this. Good work, thanks!

alban:
@mastersans This PR breaks `make test-gadgets` using the Makefile from the top-level directory:
```
$ make test-gadgets
echo Building ig-linux-amd64
Building ig-linux-amd64
...
sudo cp ig /usr/local/bin/ig
make -C gadgets/ test
make[1]: Entering directory '/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/gadgets'
make[1]: *** No rule to make target 'test'.  Stop.
make[1]: Leaving directory '/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/gadgets'
make: *** [Makefile:385: test-gadgets] Error 2
```

Could you fix it? The test-gadgets target could be replaced by two targets `test-unit-gadgets` and `test-integration-gadgets`.

mastersans:
Hi @alban  I'll get the appropiate changes to done to https://github.com/inspektor-gadget/inspektor-gadget/pull/3631 to get it fixed :+1:

PR #3515, opened 2024-09-24
cmd/common: allow to specify gadget runtime configuration using "-f"

flyth:
This allows specifying a manifest file using the "-f" flag.

Gadget instance specs can look like this:

```yaml
version: 1
kind: instance-spec
imageName: trace_exec
name: yoyo
tags:
  - foo
  - bar
paramValues:
---
version: 1
kind: instance-spec
imageName: trace_exec
name: gagaga
tags:
  - second
paramValues:
  operator.oci.wasm.gadget_a: 1212
  operator.oci.ebpf.gadget_a: 1212
```

When defining multiple specs inside the file like that, this functionality can only be used in combination with `--detach`.

### DIFF
--- a/cmd/common/oci.go
+++ b/cmd/common/oci.go
@@ -16,7 +16,10 @@ package common
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"maps"
+	"os"
 	"slices"
 	"strings"
 	"time"
@@ -28,6 +31,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/config"
 	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
+	gadgetmanifest "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-manifest"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	apihelpers "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api-helpers"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
@@ -108,10 +112,14 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 	var timeoutSeconds int
 	var gadgetInstanceID string
 
+	var inFile string
+
 	var skipParams []string
 	if commandMode == CommandModeAttach {
 		skipParams = append(skipParams, "!attach")
 	}
+
+	initializedOperators := false
 
 	preRun := func(cmd *cobra.Command, args []string) error {
 		err := runtime.Init(runtimeGlobalParams)
@@ -150,6 +158,10 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 			return cmd.ParseFlags(args)
 		}
 
+		if inFile != "" {
+			return fmt.Errorf("please only specify an image OR manifest")
+		}
+
 		ops := make([]operators.DataOperator, 0)
 		for _, op := range operators.GetDataOperators() {
 			// Initialize operator
@@ -161,6 +173,7 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 			ops = append(ops, op)
 		}
 		ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator)
+		initializedOperators = true
 
 		imageName := actualArgs[0]
 		if grpcrt, ok := runtime.(*grpcruntime.Runtime); ok && commandMode == CommandModeAttach {
@@ -226,7 +239,7 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 
 		showHelp, _ := cmd.Flags().GetBool("help")
 
-		if len(args) == 0 {
+		if len(args) == 0 && inFile == "" {
 			if showHelp {
 				additionalMessage := "Specify the gadget image to get more information about it"
 				cmd.Long = fmt.Sprintf("%s\n\n%s", cmd.Short, additionalMessage)
@@ -250,31 +263,89 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 
 		ops := make([]operators.DataOperator, 0)
 		for _, op := range operators.GetDataOperators() {
+			if !initializedOperators {
+				// initialize operators if not yet done in PreRun (e.g. when -f was specified)
+				err := op.Init(opGlobalParams[op.Name()])
+				if err != nil {
+					log.Warnf("error initializing operator %s: %v", op.Name(), err)
+					continue
+				}
+			}
 			ops = append(ops, op)
 		}
 		ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator)
 
 		timeoutDuration := time.Duration(timeoutSeconds) * time.Second
 
-		imageName := args[0]
+		var image string
+		if len(args) > 0 {
+			image = args[0]
+		}
+
+		paramValueMap := make(map[string]string)
+
+		if inFile != "" {
+			f, err := os.Open(inFile)
+			if err != nil {
+				return fmt.Errorf("opening gadget runtime manifest file %s: %w", inFile, err)
+			}
+			specs, err := gadgetmanifest.InstanceSpecsFromReader(f)
+			f.Close()
+			if err != nil {
+				return fmt.Errorf("reading gadget runtime manifest file %s: %w", inFile, err)
+			}
+
+			detachedParam := runtimeParams.Get("detach")
+			isDetach := detachedParam != nil && detachedParam.AsBool()
+
+			if isDetach {
+				// Copy special oci params
+				ociParams.CopyToMap(paramValueMap, "operator.oci.")
+				return runInstanceSpecsDetached(ctx, runtime, specs, runtimeParams, paramValueMap,
+					gadgetcontext.WithDataOperators(ops...),
+					gadgetcontext.WithTimeout(timeoutDuration),
+					gadgetcontext.WithUseInstance(false),
+				)
+			}
+
+			if len(specs) > 1 {
+				// output different error messages, depending on whether `--detach` is available
+				if detachedParam == nil {
+					return fmt.Errorf("multiple gadget instance specs found in manifest %s", inFile)
+				} else {
+					return fmt.Errorf("multiple gadget instance specs found in manifest %s: this is only supported in combination with --detach", inFile)
+				}
+			}
+
+			spec := specs[0]
+			image = spec.Image
+			runtimeParams.Set("id", spec.ID)
+			runtimeParams.Set("name", spec.Name)
+			runtimeParams.Set("tags", strings.Join(spec.Tags, ","))
+
+			tempParams := spec.ParamValues
+			maps.Copy(tempParams, paramValueMap)
+			paramValueMap = tempParams
+		}
+
 		if commandMode == CommandModeAttach {
 			// the gadgetID should be present from GetGadgetInfo above
-			imageName = gadgetInstanceID
+			image = gadgetInstanceID
 		}
 
 		gadgetCtx := gadgetcontext.New(
 			ctx,
-			imageName,
+			image,
 			gadgetcontext.WithDataOperators(ops...),
 			gadgetcontext.WithTimeout(timeoutDuration),
 			gadgetcontext.WithUseInstance(commandMode == CommandModeAttach),
 		)
 
-		paramValueMap := make(map[string]string)
-
 		// Write back param values
-		for _, p := range info.Params {
-			paramValueMap[p.Prefix+p.Key] = paramLookup[p.Prefix+p.Key].String()
+		if info != nil {
+			for _, p := range info.Params {
+				paramValueMap[p.Prefix+p.Key] = paramLookup[p.Prefix+p.Key].String()
+			}
 		}
 
 		// Also copy special oci params
@@ -318,6 +389,7 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 
 	if commandMode != CommandModeAttach {
 		AddOCIFlags(cmd, ociParams, skipParams, runtime)
+		cmd.PersistentFlags().StringVarP(&inFile, "file", "f", "", "path to gadget runtime manifest to apply")
 	}
 
 	AddOCIFlags(cmd, runtimeGlobalParams, skipParams, runtime)
@@ -328,6 +400,35 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 	}
 
 	return cmd
+}
+
+func runInstanceSpecsDetached(
+	ctx context.Context,
+	runtime runtime.Runtime,
+	specs []*gadgetmanifest.InstanceSpec,
+	runtimeParams *params.Params,
+	paramValueMap map[string]string,
+	runOptions ...gadgetcontext.Option,
+) error {
+	var merr error
+	for _, spec := range specs {
+		image := spec.Image
+
+		// Set some well-known params
+		runtimeParams.Set("id", spec.ID)
+		runtimeParams.Set("name", spec.Name)
+		runtimeParams.Set("tags", strings.Join(spec.Tags, ","))
+
+		maps.Copy(spec.ParamValues, paramValueMap)
+
+		gadgetCtx := gadgetcontext.New(ctx, image, runOptions...)
+
+		err := runtime.RunGadget(gadgetCtx, runtimeParams, spec.ParamValues)
+		if err != nil {
+			merr = errors.Join(merr, fmt.Errorf("running gadget from manifest file: %w", err))
+		}
+	}
+	return merr
 }
 
 func AddOCIFlags(cmd *cobra.Command, params *params.Params, skipParams []string, runtime runtime.Runtime) {
--- a/cmd/common/runwithspec_test.go
+++ b/cmd/common/runwithspec_test.go
@@ -1,0 +1,258 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/environment"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
+	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
+)
+
+type testMode int
+
+const (
+	testModeBoth testMode = iota
+	testModeInteractiveOnly
+	testModeDetachOnly
+)
+
+type testSpec struct {
+	Name           string
+	Manifest       string
+	Mode           testMode
+	AdditionalArgs []string
+	ExpectedParams api.ParamValues
+	ExpectedName   string
+	ExpectedID     string
+	ExpectedTags   []string
+	ExpectError    bool
+}
+
+type testRuntime struct {
+	grpcruntime.Runtime
+	t        *testing.T
+	testSpec *testSpec
+}
+
+func (r *testRuntime) Init(*params.Params) error {
+	return nil
+}
+
+func (r *testRuntime) Close() error {
+	return nil
+}
+
+func (r *testRuntime) GetGadgetInfo(gadgetCtx runtime.GadgetContext, runtimeParams *params.Params, paramValueMap api.ParamValues) (*api.GadgetInfo, error) {
+	panic("unimplemented")
+}
+
+func (r *testRuntime) RunGadget(gadgetCtx runtime.GadgetContext, runtimeParams *params.Params, paramValueMap api.ParamValues) error {
+	if r.testSpec.ExpectedID != "" {
+		assert.Equal(r.t, r.testSpec.ExpectedID, runtimeParams.Get(grpcruntime.ParamID).AsString())
+	}
+	if r.testSpec.ExpectedName != "" {
+		assert.Equal(r.t, r.testSpec.ExpectedName, runtimeParams.Get(grpcruntime.ParamName).AsString())
+	}
+	if r.testSpec.ExpectedTags != nil {
+		assert.Equal(r.t, r.testSpec.ExpectedTags, runtimeParams.Get(grpcruntime.ParamTags).AsStringSlice())
+	}
+	for k, v := range r.testSpec.ExpectedParams {
+		val, ok := paramValueMap[k]
+		require.True(r.t, ok, "expected paramValue exists for %q", k)
+		require.Equal(r.t, v, val)
+	}
+	return nil
+}
+
+func createTempManifest(t *testing.T, manifest string) (string, error) {
+	f, err := os.CreateTemp("", "manifest")
+	if err != nil {
+		return "", fmt.Errorf("creating temp manifest file: %w", err)
+	}
+	t.Cleanup(func() {
+		os.Remove(f.Name())
+	})
+	_, err = f.WriteString(manifest)
+	if err != nil {
+		return "", fmt.Errorf("writing temp manifest: %w", err)
+	}
+	return f.Name(), nil
+}
+
+func TestRunWithSpec(t *testing.T) {
+	tests := []*testSpec{
+		{
+			Name:           "specified image on command line",
+			ExpectError:    true,
+			AdditionalArgs: []string{"trace_exec"},
+			Manifest: `
+apiVersion: 1
+kind: instance-spec
+image: trace_exec
+`,
+		},
+		{
+			Name: "invalid version",
+			Manifest: `
+apiVersion: a`,
+			ExpectError: true,
+		},
+		{
+			Name: "invalid kind",
+			Manifest: `
+apiVersion: 1
+kind: yo`,
+			ExpectError: true,
+		},
+		{
+			Name: "one spec with id, name, tags and params",
+			Manifest: `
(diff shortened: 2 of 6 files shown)


blanquicet:
> Also, documentation would be welcomed! But let's first agree on the naming before proceeding with it.
>
IMO, it's a _must_ for the release. But yeah, let's first agree on the naming ... naming ... naming 🤣

pkg/gadget-config/config.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func GetFromReader(r io.Reader) ([]*Config, error) {
+	cfgBlob, err := io.ReadAll(r)
+	if err != nil {
+		return nil, fmt.Errorf("reading gadget config from reader: %w", err)
+	}
+
+	res := make([]*Config, 0)
+	cfgEntries := multiYamlRegex.Split(string(cfgBlob), -1)
```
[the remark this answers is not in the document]
flyth: I used what we had since I thought there was no other solution - however, I found a way and applied it now.

cmd/common/oci.go
```diff
@@ -256,12 +260,77 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
 
-		imageName := args[0]
+		var imageName string
+		if len(args) > 0 {
+			imageName = args[0]
+		}
+
+		paramValueMap := make(map[string]string)
+
+		if inFile != "" {
```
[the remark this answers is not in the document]
flyth: I extracted parts to a function to increase readability.

cmd/common/oci.go
```diff
@@ -256,12 +260,77 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+			isDetach := false
+			if detachedParam := runtimeParams.Get("detach"); detachedParam != nil && detachedParam.AsBool() {
+				isDetach = true
+			}
+
+			if len(cfgs) > 1 && !isDetach {
+				return fmt.Errorf("multiple gadget runtime configurations found in file %s: this is only supported in combination with --detach", inFile)
+			}
+
+			if isDetach {
```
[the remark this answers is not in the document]
flyth: See above, I extracted parts to a new func to increase readability. Now both branches are similar in length.

flyth:
Changed:

File is now called `manifest`. A manifest can host multiple things of which we only evaluate those of kind `instance-spec`(maybe `runtime-spec` is better?) - the kind field is mandatory. Also added a mandatory version field to future-proof.

WDYT?

cmd/common/oci.go
```diff
@@ -327,9 +377,42 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 		AddOCIFlags(cmd, operatorParams, skipParams, runtime)
 	}
 
+	cmd.PersistentFlags().StringVarP(&inFile, "file", "f", "", "path to gadget runtime configuration to apply")
```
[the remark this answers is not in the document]
flyth: Probably something like "path to manifest file containing the instance specs".

mauriciovasquezbernal:
> Changed:
> 
> File is now called `manifest`. A manifest can host multiple things of which we only evaluate those of kind `instance-spec`(maybe `runtime-spec` is better?) - the kind field is mandatory. Also added a mandatory version field to future-proof. This was done to align more with the kubernetes naming as discussed.
> 
> WDYT?

I like it. I think the items inside the manifest should be called "instance-specs".

flyth:
> > Changed:
> > File is now called `manifest`. A manifest can host multiple things of which we only evaluate those of kind `instance-spec`(maybe `runtime-spec` is better?) - the kind field is mandatory. Also added a mandatory version field to future-proof. This was done to align more with the kubernetes naming as discussed.
> > WDYT?
> 
> I like it. I think the items inside the manifest should be called "instance-specs".

So `instance-specs` as kind vs `instance-spec`? Kubernetes is only using `spec`.

cmd/common/oci.go
```diff
@@ -256,12 +261,57 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 
 		timeoutDuration := time.Duration(timeoutSeconds) * time.Second
 
-		imageName := args[0]
+		var imageName string
+		if len(args) > 0 {
+			imageName = args[0]
+		}
+
```
mauriciovasquezbernal, line 268: What about if the user runs `... run -f foo.yaml myimage:latest`? I think it should fail.

cmd/common/oci.go
```diff
@@ -256,12 +261,57 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+			if err != nil {
+				return fmt.Errorf("reading gadget runtime configuration file %s: %w", inFile, err)
+			}
+
+			detachedParam := runtimeParams.Get("detach")
+			isDetach := detachedParam != nil && detachedParam.AsBool()
+
+			if len(specs) > 1 && !isDetach {
+				return fmt.Errorf("multiple gadget runtime configurations found in file %s: this is only supported in combination with --detach", inFile)
+			}
```
mauriciovasquezbernal, line 287: nit: move this after the `if isDetach` conditional? (It seems to me it breaks a little bit the flow of the logic in the current position).

If it's moved below, then it could just be `if len(specs) > 0` as the isDetach branch was handled before.

pkg/gadget-manifest/instance-spec.go
```diff
@@ -0,0 +1,66 @@
...
+
+type InstanceSpec struct {
+	Version     int               `json:"version" yaml:"version"`
+	Kind        string            `json:"kind" yaml:"kind"`
+	ImageName   string            `json:"imageName" yaml:"imageName"`
+	ID          string            `json:"id" yaml:"id"`
+	Name        string            `json:"name" yaml:"name"`
+	Tags        []string          `json:"tags" yaml:"tags"`
+	ParamValues map[string]string `json:"paramValues" yaml:"paramValues"`
+}
```
mauriciovasquezbernal, line 38: How should we validate this struct? Should we define a yaml schema?

cmd/common/oci.go
```diff
@@ -256,12 +261,57 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+
+			if isDetach {
+				// Copy special oci params
+				ociParams.CopyToMap(paramValueMap, "operator.oci.")
```
mauriciovasquezbernal, line 291: I don't understand this logic. It's copying some parameters to the map, but `ParamValues` on the spec is totally ignored. 

Perhaps something like can help:

```bash 
diff --git a/cmd/common/oci.go b/cmd/common/oci.go
index eb407ce0c..e428af186 100644
--- a/cmd/common/oci.go
+++ b/cmd/common/oci.go
@@ -288,8 +288,8 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn

                        if isDetach {
                                // Copy special oci params
-                               ociParams.CopyToMap(paramValueMap, "operator.oci.")
-                               return runInstanceSpecsDetached(ctx, runtime, specs, runtimeParams, paramValueMap,
+                               //ociParams.CopyToMap(paramValueMap, "operator.oci.")
+                               return runInstanceSpecsDetached(ctx, runtime, specs, runtimeParams, ociParams,
                                        gadgetcontext.WithTimeout(timeoutDuration),
                                        gadgetcontext.WithUseInstance(commandMode == CommandModeAttach),
@@ -387,11 +387,15 @@ func runInstanceSpecsDetached(
        runtime runtime.Runtime,
        specs []*gadgetconfig.InstanceSpec,
        runtimeParams *params.Params,
-       paramValueMap map[string]string,
+       ociParams *params.Params,
        runOptions ...gadgetcontext.Option,
 ) error {
        var merr error
        for _, spec := range specs {
+               // TODO: Clone map?
+               paramValueMap := spec.ParamValues
+               ociParams.CopyToMap(paramValueMap, "operator.oci.")
+
                imageName := spec.ImageName

                // Set some well-known params
```

cmd/common/oci.go
```diff
@@ -256,12 +261,57 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+					gadgetcontext.WithUseInstance(commandMode == CommandModeAttach),
+				)
+			}
+
+			cfg := specs[0]
+			imageName = cfg.ImageName
+			runtimeParams.Set("id", cfg.ID)
+			runtimeParams.Set("name", cfg.Name)
+			runtimeParams.Set("tags", strings.Join(cfg.Tags, ","))
+		}
```
mauriciovasquezbernal, line 304: Add `paramValueMap = cfg.ParamValues`, otherwise parameters on the file will be ignored when running `... run -f foo.yaml` (without --detach)

cmd/common/oci.go
```diff
@@ -256,12 +261,57 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+
 		if commandMode == CommandModeAttach {
 			// the gadgetID should be present from GetGadgetInfo above
 			imageName = gadgetInstanceID
 		}
 
+		if imageName == "" {
+			return fmt.Errorf("no image name provided")
+		}
```
mauriciovasquezbernal, line 313: When can this happen? only when using `-f` and having the yaml template without the image? IMHO this validation should be in `InstanceSpecsFromReader`.

mauriciovasquezbernal:
> So instance-specs as kind vs instance-spec? Kubernetes is only using spec.

Nope, I meant `instance-spec`s. Kind should be `instance-spec` as it refers to a single one.

mauriciovasquezbernal (commented):
It looks very good! 


Few general questions:
- Are planning to provide a way to delete the whole manifest? Like `gadgetctl delete -f manifest.yaml` (I suppose it'll only work if the manifests have the name and/or the ID defined)
- Should we print a warning in case the user defines a non existing parameter? If there is a typo in a parameter, and we don't print any warning, it could take al of time for the user to understand what's wrong. (I have been in that position multiple times). (this can be done later on as an improvement as it's not strictly related to this PR)

cmd/common/oci.go
```diff
@@ -256,12 +261,57 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+				return fmt.Errorf("multiple gadget runtime configurations found in file %s: this is only supported in combination with --detach", inFile)
+			}
+
+			if isDetach {
+				// Copy special oci params
+				ociParams.CopyToMap(paramValueMap, "operator.oci.")
+				return runInstanceSpecsDetached(ctx, runtime, specs, runtimeParams, paramValueMap,
+					gadgetcontext.WithDataOperators(ops...),
+					gadgetcontext.WithTimeout(timeoutDuration),
+					gadgetcontext.WithUseInstance(commandMode == CommandModeAttach),
```
blanquicet, line 295: I don't think this line is necessary. It'll never happen.

pkg/gadget-manifest/instance-spec.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type InstanceSpec struct {
+	Version     int               `json:"version" yaml:"version"`
+	Kind        string            `json:"kind" yaml:"kind"`
+	ImageName   string            `json:"imageName" yaml:"imageName"`
+	ID          string            `json:"id" yaml:"id"`
+	Name        string            `json:"name" yaml:"name"`
+	Tags        []string          `json:"tags" yaml:"tags"`
+	ParamValues map[string]string `json:"paramValues" yaml:"paramValues"`
```
blanquicet, line 37: You're never setting it. It's also missing in the run command to take this value and send it in the `RunGadget`.

cmd/common/oci.go
blanquicet, line 380: This flag is available for `ig`, but it's not working. Mmmm in general, should it be there?

pkg/gadget-manifest/instance-spec.go
```diff
@@ -0,0 +1,66 @@
...
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("parsing gadget spec (entry %d): %w", c, err)
+		}
+		if cfg == nil {
+			continue
+		}
+		if cfg.Kind != KindInstanceSpec {
+			return nil, fmt.Errorf("expected kind %q in manifest, got kind %q", KindInstanceSpec, cfg.Kind)
```
blanquicet, line 58:
```suggestion
			return nil, fmt.Errorf("expected kind %q for entry %d in manifest, got kind %q", KindInstanceSpec, c, cfg.Kind)
```

blanquicet (commented):
> > I like it. I think the items inside the manifest should be called "instance-specs".
> 
> So `instance-specs` as kind vs `instance-spec`? Kubernetes is only using `spec`.
>
IMO, it should be `instance-spec`.

Looks great! I have some general comments:

- `-f` is available in `ig`, should it be there? If yes, it's not working.
- paramValues are not working
- Do you plan to provide a way to delete them all at once?

cmd/common/oci.go
flyth: I don't know - should it?

I could imagine users having setup filters for namespaces / pods etc in a manifest and then want to run different gadgets using it.

cmd/common/oci.go
flyth: Yeah, I broke it with the recent push, sorry :/

Added some tests to make sure it's valid.

pkg/gadget-manifest/instance-spec.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+type InstanceSpec struct {
+	Version     int               `json:"version" yaml:"version"`
+	Kind        string            `json:"kind" yaml:"kind"`
+	Image       string            `json:"image" yaml:"image"`
+	ID          string            `json:"id" yaml:"id"`
+	Name        string            `json:"name" yaml:"name"`
+	Tags        []string          `json:"tags" yaml:"tags"`
+	ParamValues map[string]string `json:"paramValues" yaml:"paramValues"`
```
mqasimsarfraz, line 37: Just to confirm: `map[string]string `  means we don't support:
```
paramValues:
  operator:
    foo: baz
    bar:  baz
```

mqasimsarfraz (commented):
Looking goo!

cmd/common/oci.go
mauriciovasquezbernal: IMO it's too confusing to have this mixed operation mode. What about if there are multiple instances on the manifest? What about if the manifest has an image defined but it doesn't match with the one on the CLI?

blanquicet: IMO, it breaks the balance between flexibility and complexity, so it should fail ... "Keep it simple" 🙂

cmd/common/runwithspec_test.go
```diff
@@ -0,0 +1,240 @@
...
+				}
+				require.NoError(t, err)
+			}
+		}
+		if tt.Mode == testModeInteractiveOnly || tt.Mode == testModeBoth {
+			t.Run(tt.Name, runTest(false))
+		}
+		if tt.Mode == testModeDetachOnly || tt.Mode == testModeBoth {
+			t.Run(tt.Name, runTest(true))
+		}
```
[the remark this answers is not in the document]
flyth: `testModeBoth` (testing for detached and non-detached mode) is used by default if no `*Only` mode is specified (and thus being used by most tests now).

pkg/gadget-manifest/instance-spec.go
flyth: Yeah, we don't. Do you think it's not okay to not support that?

pkg/gadget-manifest/instance-spec.go
flyth: What exactly do you mean? Applying stricter rules to catch extra fields for example? Or to provide things like auto-completion in IDEs?

Currently, we check for `version` and `kind` when parsing, and for `image` before running.

cmd/common/oci.go
flyth: I don't know. I don't like artificially preventing users from doing things that would work just fine by intuition (well, at least my intuition 😄)

Currently, everything specified on the command-line would take precedence over what is written in the manifest. Even for multiple instances that would be true.

But I seem to be in the minority seeing your comments + upvotes, hehe.

So, just so that we're on the same page, please confirm:
* using `-f` should prevent specifying an image on CLI
* parameters specified on the CLI should be ignored (or warned about? tricky.) when using `-f`

Careful, though: this would (without introducing exceptions) also apply to connection related settings (like setting the `--remote-address`, for example).

pkg/gadget-manifest/instance-spec.go
mqasimsarfraz: We support it in [configuration file](https://www.inspektor-gadget.io/docs/latest/reference/configuration#configuration-file) so was wondering if we should keep it same here but I don't have too strong opinion on it!

cmd/common/oci.go
mauriciovasquezbernal: Perhaps I'm misunderstanding something. Is it possible to specify things like `--remote-address`, `--connection-timeout` (and other global parameters) on the manifest file? I would expect this not to be possible as global params **aren't** related to a specific gadget instance. 

For "instance parameters", I think they won't be available as not gadget image is specified.  Hence, in my understanding there is not any parameter that can be overwritten:
- global params: only available on the CLI 
- instance params: only available on the manifest

pkg/gadget-manifest/instance-spec.go
mauriciovasquezbernal: I meant to check things like:
- Name can't contain white spaces 
- Tags can't contain `,`
- etc.

pkg/gadget-manifest/instance-spec.go
mauriciovasquezbernal: I think supporting only the "dot" notation is better. It's the same notation used by the Golang API and very similar to the CLI (in the CLI we can use shorter names).

cmd/common/oci.go
flyth: You're totally right about the global params, I mixed those up.

Instance params _are_ available, but only if a gadget image is specified - but I will change it to the behavior you described.

pkg/gadget-manifest/instance-spec.go
flyth: You're right, it makes sense to add those.

cmd/common/oci.go
flyth: It was not correctly initializing the operators in that case, I fixed it.

cmd/common/oci.go
flyth: Added validation to the spec reader.

cmd/common/oci.go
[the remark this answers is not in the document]
flyth: The plan was to also allow attaching if only one spec is present - but we can (re-)add that properly later and focus on just starting things from the file for now.

docs/reference/manifests.mdx
```diff
@@ -0,0 +1,41 @@
...
+```
+
+This example contains two gadget instance specifications that are divided by three dashes (`---`).
+
+The `version` and `kind` fields define the specification used. They are mandatory and currently we only support
+this version and kind as part of the manifest.
+
+Specifying an `image` is also mandatory, while the rest of the fields are optional.
+
+When specifying `paramValues`, please use the fully qualified parameter names provided with their respective
```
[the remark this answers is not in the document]
flyth: Provided a link to the operators docs and an example reference.

docs/reference/manifests.mdx
```diff
@@ -0,0 +1,123 @@
...
+image: trace_open
+name: my-gadget-2
+```
+
+This example contains two gadget instance specifications that are divided by three dashes (`---`).
+
+The `apiVersion` and `kind` fields define the specification used. They are mandatory and currently we only support
+this version and kind as part of the manifest.
+
+Specifying an `image` is also mandatory, while the rest of the fields are optional.
```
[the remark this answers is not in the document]
flyth: I'm fine with either way. But we need to come to a conclusion :)

/cc @mauriciovasquezbernal

cmd/common/oci.go
```diff
@@ -250,31 +263,93 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+				return fmt.Errorf("reading gadget runtime manifest file %s: %w", inFile, err)
+			}
+
+			detachedParam := runtimeParams.Get("detach")
+			isDetach := detachedParam != nil && detachedParam.AsBool()
+
+			if isDetach {
+				// Copy special oci params
+				ociParams.CopyToMap(paramValueMap, "operator.oci.")
+				return runInstanceSpecsDetached(ctx, runtime, specs, runtimeParams, paramValueMap,
```
mauriciovasquezbernal, line 304: super minor nit: `paramValueMap` is empty at this point, what about passing `ociParams` to `runInstanceSpecsDetached` and moving `ociParams.CopyToMap(paramValueMap, "operator.oci.")` into it?

cmd/common/oci.go
```diff
@@ -250,31 +263,93 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+
+			spec := specs[0]
+			image = spec.Image
+			runtimeParams.Set("id", spec.ID)
+			runtimeParams.Set("name", spec.Name)
+			runtimeParams.Set("tags", strings.Join(spec.Tags, ","))
+
+			tempParams := spec.ParamValues
+			maps.Copy(tempParams, paramValueMap)
+			paramValueMap = tempParams
```
mauriciovasquezbernal, line 328: I have read these lines multiple times and I can't understand what they do. 
Can't it just be `maps.Copy(paramValueMap, spec.ParamValues)`?

cmd/common/oci.go
```diff
@@ -330,6 +406,39 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
...
+		runtimeParams.Set("name", spec.Name)
+		runtimeParams.Set("tags", strings.Join(spec.Tags, ","))
+
+		maps.Copy(spec.ParamValues, paramValueMap)
+
+		gadgetCtx := gadgetcontext.New(
+			ctx,
+			image,
+			runOptions...,
+		)
```
mauriciovasquezbernal, line 432: `gadgetCtx := gadgetcontext.New(ctx, image, runOptions...)`?

(I kind of feel bad commenting these things, but on the other hand I want to keep it as clean as possible)

docs/reference/manifests.mdx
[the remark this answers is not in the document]
mauriciovasquezbernal: _I think_ `image` is better than `gadget`. In kubernetes, docker, etc. image is used instead of container.

docs/reference/manifests.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+When specifying `paramValues`, please use the fully qualified parameter names provided with their respective
+documentations in the [operators section](../spec/operators) ([example](../spec/operators/filter#filter)).
+
+:::warning
+
+If a manifest contains multiple instance specs, it can only be run by additionally using `--detach`. This is however
+only available when using `gadgetctl`. The run command will then try to create instances for all given specs and return
+either their IDs.
```
mauriciovasquezbernal, line 46:
```suggestion
their IDs.
```

Or what's the role of "either" in his sentence?

docs/reference/manifests.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+:::
+
+## Running a single gadget interactively
+
+When running interactively, you are limited to one gadget instance spec per manifest. This applies to `ig` as well as
+`gadgetctl` without the `--detach` flag.
+
+<Tabs groupId="env">
```
mauriciovasquezbernal, line 55: The idea of using tabs is to differentiate different ways of performing the same task, for instance, to compare how to run the same gadget with ig, gadgectl, kubectl-gadge, etc.. I think in this case the gadget.yaml should be shown before and  then there can be tabs for ig and gadgetctl.

cmd/common/runwithspec_test.go
```diff
@@ -0,0 +1,260 @@
...
+	if r.testSpec.ExpectedID != "" {
+		assert.Equal(r.t, r.testSpec.ExpectedID, runtimeParams.Get(grpcruntime.ParamID).AsString())
+	}
+	if r.testSpec.ExpectedName != "" {
+		assert.Equal(r.t, r.testSpec.ExpectedName, runtimeParams.Get(grpcruntime.ParamName).AsString())
+	}
+	if r.testSpec.ExpectedTags != nil {
+		assert.Equal(r.t, r.testSpec.ExpectedTags, runtimeParams.Get(grpcruntime.ParamTags).AsStringSlice())
+	}
+	if r.testSpec.ExpectedParams != nil {
```
mauriciovasquezbernal, line 81: nit: this check is not needed as range over nil works fine.

cmd/common/runwithspec_test.go
```diff
@@ -0,0 +1,260 @@
...
+		for k, v := range r.testSpec.ExpectedParams {
+			val, ok := paramValueMap[k]
+			require.True(r.t, ok, "expected paramValue exists for %q", k)
+			require.True(r.t, ok, "expected paramValue to be %q, got %q", v, val)
```
mauriciovasquezbernal, line 85: `require.Equal(r.t, v, val)` ? 

Also the require package already provides the "expected .. got" message, so we don't need to add it.

cmd/common/runwithspec_test.go
```diff
@@ -0,0 +1,260 @@
...
+`,
+			ExpectError: true,
+		},
+	}
+
+	environment.Environment = environment.Local
+
+	for _, tt := range tests {
+		runTest := func(detach bool) func(t *testing.T) {
+			return func(t *testing.T) {
```
mauriciovasquezbernal, line 229: I added `t.Parallel()` and everything exploded :exploding_head:

cmd/common/oci.go
mauriciovasquezbernal: So we can remove this check here, right?

- if not using "-f" help will be printed
- if using "-f" the spec reader will complain

mauriciovasquezbernal (approved):
Some nits but LGTM already. Thanks for the work on this!

cmd/common/runwithspec_test.go
```diff
@@ -0,0 +1,260 @@
...
+func TestRunWithSpec(t *testing.T) {
+	tests := []*testSpec{
+		{
+			Name:           "specified image on command line",
+			ExpectError:    true,
+			AdditionalArgs: []string{"trace_exec"},
+			Manifest: `
+apiVersion: 1
+kind: instance-spec
+image: trace_exec
```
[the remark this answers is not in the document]
flyth: No, it's supposed to throw an error - e.g. the image from the CLI should _not_ be used if a file is specified.

docs/reference/manifests.mdx
flyth: It was "either an error or..." at one point, but I decided against using it. Didn't catch that when re-reading 😄

cmd/common/oci.go
flyth: Don't feel bad about it.
I think I left it that way because it felt "more important" that way (the gadgetcontext being the central point of things), but I agree this should be shortened.

docs/reference/manifests.mdx
flyth: Hmm, I've also seen tabs used to show the contents of different files, for example when showing HTML/JS/CSS. But this is kind of mixed. I still think it's useful this way because it shows that the example consists of both the command and the file, without taking more space (we already got an example at the top.)

cmd/common/runwithspec_test.go
flyth: It seems that `setFlagsFromConfig()` has global side-effects that cause that (concurrent map read/writes). I think it's fine for the tests.

flyth:
Thanks for the reviews!

PR #3517, opened 2024-09-24
trace_tcpconnect: Fix missing src and dst ip and port on error

burak-ok:
When an error happens in the connect call the src ip addr + port and the dst port is cleared in the kernel before returning. Therefore we need to save all the information at a point when it is available.

Edit: Actual source line is: https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/af_inet.c#L738
~~See https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L350~~

## Before
```bash
RUNTIME.CONTAINERNAME        SRC                                   DST                                   COMM                  PID        TID        UID        GID ERROR     
quizzical_cannon             172.17.0.2:34286                      142.250.185.131:80                    wget                41798      41798          0          0           
quizzical_cannon             172.17.0.2:35018                      142.250.186.99:80                     wget                41798      41798          0          0           
quizzical_cannon             0.0.0.0:0                             192.168.2.1:0                         wget                41882      41882          0          0 ECONNREFUS
quizzical_cannon             0.0.0.0:0                             192.168.1.50:0                        wget                42026      42026          0          0 EHOSTUNREA
```

## After
```bash
RUNTIME.CONTAINERNAME        SRC                                   DST                                   COMM                  PID        TID        UID        GID ERROR     
quizzical_cannon             172.17.0.2:53208                      142.250.186.99:80                     wget                40138      40138          0          0           
quizzical_cannon             172.17.0.2:56386                      216.58.206.67:80                      wget                40138      40138          0          0           
quizzical_cannon             172.17.0.2:47944                      192.168.2.1:80                        wget                40195      40195          0          0 ECONNREFUS
quizzical_cannon             172.17.0.2:32952                      192.168.1.50:80                       wget                40736      40736          0          0 EHOSTUNREA
```

### DIFF
--- a/gadgets/trace_tcpconnect/README.mdx
+++ b/gadgets/trace_tcpconnect/README.mdx
@@ -6,9 +6,11 @@ sidebar_position: 0
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-# trace tcpconnect
+# trace_tcpconnect
 
-trace tcp connections
+This gadget traces outgoing TCP connect attempts initiated through the `connect` system call.
+For non-blocking sockets ([O_NONBLOCK](https://man7.org/linux/man-pages/man7/socket.7.html)), the connection attempt is handled asynchronously.
+In these cases, the error field may not reflect the final result of the connection attempt but may instead show `115`/`EINPROGRESS`, indicating that the attempt has been initiated but is not yet complete.
 
 ## Getting started
 
@@ -34,4 +36,132 @@ No flags.
 
 ## Guide
 
-TODO
+Create an interactive test workload in another terminal:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl run --rm -it --image busybox test-trace-tcpconnect
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ docker run --rm -it --name test-trace-tcpconnect busybox'
+        ```
+    </TabItem>
+</Tabs>
+
+Then, let's run the gadget and trace some events:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl-gadget run trace_tcpconnect:%IG_TAG% --containername test-trace-tcpconnect
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run trace_tcpconnect:%IG_TAG% --containername test-trace-tcpconnect
+        ```
+    </TabItem>
+</Tabs>
+
+To generate some event we go back to our test workload, which we created earlier, and use `wget` on some hosts:
+
+```bash
+$ wget www.inspektor-gadget.io
+$ wget 192.168.1.1
+```
+
+Now we should see some events in the terminal where the gadget is running:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        K8S.NODE          K8S.NAMESPACE     K8S.PODNAME       K8S.CONTAINERNAME SRC                     DST                     COMM            PID       TID      UID      GID ERROR 
+        minikube-docker   default           test-tra…pconnect test-trace-tcpcon 10.244.0.5:33152        188.114.96.3:80         wget         104183    104183        0        0       
+        minikube-docker   default           test-tra…pconnect test-trace-tcpcon 10.244.0.5:44394        188.114.96.3:443        wget         104183    104183        0        0       
+        minikube-docker   default           test-tra…pconnect test-trace-tcpcon 10.244.0.5:46028        192.168.1.1:80          wget         104997    104997        0        0 ECONNR
+        ^C
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        RUNTIME.CONTAINERNAME        SRC                                   DST                                   COMM                  PID        TID        UID        GID ERROR     
+        test-trace-tcpconnect        172.17.0.2:53208                      188.114.96.3:80                       wget                40138      40138          0          0           
+        test-trace-tcpconnect        172.17.0.2:56386                      188.114.96.3:443                      wget                40138      40138          0          0           
+        test-trace-tcpconnect        172.17.0.2:50946                      192.168.1.1:80                        wget                40697      40697          0          0 ECONNREFUS
+        ^C
+        ```
+    </TabItem>
+</Tabs>
+
+Finally, to clean the system we switch back to the test workload and exit out of it:
+```bash
+$ exit
+```
+
+## Architecture
+
+```mermaid
+sequenceDiagram
+    participant Tracee
+
+    box Kernel functions
+        participant sys_connect
+        participant inet_stream_connect
+        participant tcp_v4_connect/tcp_v6_connect
+    end
+    box eBPF maps
+        participant sockets_per_process
+        participant src_dst_per_process
+        participant events
+    end
+
+    activate Tracee
+    Tracee ->> sys_connect: connect()
+    activate sys_connect
+
+    sys_connect ->> inet_stream_connect: 
+    activate inet_stream_connect
+
+    rect rgba(0, 0, 255, .1)
+        Note right of inet_stream_connect: Save sock in kprobe.<br/>We need it in the kretprobe,<br/>but don't have access to it
+        inet_stream_connect -->> sockets_per_process: 
+        activate sockets_per_process
+    end
+
+    inet_stream_connect ->> tcp_v4_connect/tcp_v6_connect: 
+    activate tcp_v4_connect/tcp_v6_connect
+
+    rect rgba(0, 0, 255, .1)
+        Note right of tcp_v4_connect/tcp_v6_connect: Save src/dst ip/port in kretprobe.<br/>At the end of inet_stream_connect<br/>they might get reset
+        tcp_v4_connect/tcp_v6_connect -->> src_dst_per_process: 
+        activate src_dst_per_process
+    end
+
+    tcp_v4_connect/tcp_v6_connect ->> inet_stream_connect: 
+    deactivate tcp_v4_connect/tcp_v6_connect
+
+
+    rect rgba(0, 0, 255, .1)
+        Note right of inet_stream_connect: Gather all information in the kretprobe<br/>and submit the event
+        sockets_per_process -->> inet_stream_connect: 
+        deactivate sockets_per_process
+        src_dst_per_process -->> inet_stream_connect: 
+        deactivate src_dst_per_process
+        inet_stream_connect -->> events: 
+        activate events
+    end
+
+    inet_stream_connect ->> sys_connect: 
+    deactivate inet_stream_connect
+    sys_connect ->> Tracee: 
+    deactivate sys_connect
+
+    deactivate Tracee
+    deactivate events
+```
--- a/gadgets/trace_tcpconnect/program.bpf.c
+++ b/gadgets/trace_tcpconnect/program.bpf.c
@@ -33,6 +33,11 @@ struct ipv6_flow_key {
 	__u16 dport;
 };
 
+struct src_dst {
+	struct gadget_l4endpoint_t src;
+	struct gadget_l4endpoint_t dst;
+};
+
 struct event {
 	gadget_timestamp timestamp_raw;
 	gadget_mntns_id mntns_id;
@@ -64,14 +69,25 @@ const volatile __u64 targ_min_latency_ns = 0;
 #define AF_INET6 10
 
 // sockets_per_process keeps track of the sockets between:
-// - kprobe enter_tcp_connect
-// - kretprobe exit_tcp_connect
+// - kprobe inet_stream_connect
+// - kretprobe inet_stream_connect
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, MAX_ENTRIES);
 	__type(key, u32); // tid
 	__type(value, struct sock *);
 } sockets_per_process SEC(".maps");
+
+// src_dst_per_process keeps track of the src and dst information
+// between tcp_v4/6_connect and the return of inet_stream_connect
+// Since tcp_v4/6_connect is directly inside of inet_stream_connect,
+// we do not need to keep a huge number of entries.
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, MAX_ENTRIES);
+	__type(key, u32); // tid
+	__type(value, struct src_dst);
+} src_dst_per_process SEC(".maps");
 
 struct piddata {
 	gadget_comm comm[TASK_COMM_LEN];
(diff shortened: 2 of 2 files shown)


gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -372,6 +422,62 @@ int BPF_KRETPROBE(ig_inet_stream_connect_ret, int ret)
 	return exit_tcp_connect(ctx, ret);
 }
 
+SEC("kretprobe/tcp_v4_connect")
```
blanquicet, line 425: Why not to use `kprobe/tcp_v{4|6}_connect`? It seems to me that the src and dst information is already present in the args:

- `sk` already contains the source IP and port: https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L218
- `uaddr` contains the dst IP and port: https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L216

Using `kprobe/tcp_v{4|6}_connect` will allow us to still get the [src IP/port](https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L350) and [dst port](https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L352) when there's an error sending the SYN (which is not covered with this PR).

gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -372,6 +422,62 @@ int BPF_KRETPROBE(ig_inet_stream_connect_ret, int ret)
...
+	} else {
+		read_l4endpoints_from_sock_v6(&src_dst_entry.src,
+					      &src_dst_entry.dst, sk);
+	}
+
+	bpf_map_update_elem(&src_dst_per_process, &tid, &src_dst_entry, 0);
+	return 0;
+}
+
+SEC("kretprobe/tcp_v6_connect")
```
blanquicet, line 453: IMO, the logic became quite complex. I suggest creating a mermaid diagram as we did for [fsnotify](https://www.inspektor-gadget.io/docs/latest/gadgets/fsnotify#architecture) where we describe what we do at each hook. In particular, how programs interact with maps. WDYT?

gadgets/trace_tcpconnect/README.mdx
```diff
@@ -28,10 +30,120 @@ Running the gadget:
     </TabItem>
 </Tabs>
 
-## Flags
+## Guide
 
-No flags.
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
```
blanquicet, line 36: Improvement for a follow-up PR:

Use here the example we had in the built-in version here: https://www.inspektor-gadget.io/docs/latest/gadgets/builtin/trace/tcpconnect#on-kubernetes (After updating it with new error field 🙂)

blanquicet (commented):
Thanks for fixing it and creating an initial guide!

gadgets/trace_tcpconnect/README.mdx
```diff
@@ -28,10 +30,120 @@ Running the gadget:
...
+        minikube-docker     default                     myapp1-pod                  myapp1-pod                  date                    2957148        2957148 sh                    2589510 /bin/date           root           uid:4294967295 root
+        minikube-docker     default                     myapp1-pod                  myapp1-pod                  cat                     2957149        2957149 sh                    2589510 /bin/cat /pro…      root           uid:4294967295 root
+        minikube-docker     default                     myapp1-pod                  myapp1-pod                  sleep                   2957150        2957150 sh                    2589510 /bin/sleep 1        root           uid:4294967295 root
+        ^C
+        ```
+
+        Processes of both pods are spawned: `myapp1-pod` spawns `cat /proc/version` and `/bin/sleep 1`,
+        `myapp2-pod` spawns `/bin/echo sleep-10` and `/bin/sleep 10`, both spawn `true` and `date`.
+        We can stop to trace again by hitting Ctrl-C.
```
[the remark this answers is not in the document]
burak-ok: Even better: I didn't save the file before commiting

gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -210,13 +253,11 @@ static __always_inline void trace_v4(struct pt_regs *ctx, __u64 pid_tgid,
...
-	BPF_CORE_READ_INTO(&event->src.addr_raw.v4, sk,
-			   __sk_common.skc_rcv_saddr);
-	BPF_CORE_READ_INTO(&event->dst.addr_raw.v4, sk, __sk_common.skc_daddr);
-	event->dst.port = dport;
-	event->src.port = BPF_CORE_READ(sk, __sk_common.skc_num);
+	if (src_dst_entry) {
+		event->src = src_dst_entry->src;
+		event->dst = src_dst_entry->dst;
+	} else
+		read_l4endpoints_from_sock_v4(&event->src, &event->dst, sk);
```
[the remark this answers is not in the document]
burak-ok: It can happen that src/dst is already there. For example when calling `connect` on an already connected socket. I don't know for `SS_CONNECTING` though.

In general any other path/case than https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/af_inet.c#L654

gadgets/trace_tcpconnect/program.bpf.c
burak-ok: For example the source port is still unknown at the start of the function. It is set in https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L299 There `inet_hash_connect` is called which sets the port. See comment https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/inet_hashtables.c#L1164

gadgets/trace_tcpconnect/program.bpf.c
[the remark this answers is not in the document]
blanquicet: You're right. Could you please add a comment to clarify it? Something like:

```suggestion
	} else {
		// src_dst_entry is not set when tcp_v{4|6}_connect is not called, e.g.,
		// when calling inet_stream_connect on an already connected socket. So,
		// try to read the endpoints from the socket here.
		read_l4endpoints_from_sock_v4(&event->src, &event->dst, sk);
	}
```

Same in `trace_v6` and `exit_inet_stream_connect`.

gadgets/trace_tcpconnect/program.bpf.c
blanquicet: It'd be highly appreciated 🙂

gadgets/trace_tcpconnect/program.bpf.c
burak-ok: I don't understand what you are suggesting, maybe I missed something.

This `kretprobe/tcp_v4_connect` was introduced because we don't have the port and ip in `exit_inet_stream_connect` when an error happened

blanquicet: I'm suggesting anticipating the collection of the src/dst ip and port from `kretprobe/tcp_v{4|6}_connect` to `kprobe/tcp_v{4|6}_connect` because by using `kretprobe/tcp_v{4|6}_connect`, we'll not provide the src IP/port and dst port when there's an error in `tcp_v{4|6}_connect` (while sending the SYN) because they are cleaned up: [here](https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L350) and [here](https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L353), respectively.

The problem with applying my suggestion is that we will not provide the src port if there is an error in `inet_stream_connect` because, as you said, the src port is still unknown in `kprobe/tcp_v{4|6}_connect`. So, let's continue with your proposal, but please document what is still missing.

burak-ok: We could maybe trace `inet_hash_connect` instead then: https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L299
But I am currently not so sure what this function does right after the hash connect: https://elixir.bootlin.com/linux/v6.11/source/net/ipv4/tcp_ipv4.c#L305

gadgets/trace_tcpconnect/program.bpf.c
burak-ok: Could you please take a look. Since I am currently too familiar with the topic I want to make sure its understandable for everyone

gadgets/trace_tcpconnect/README.mdx
```diff
@@ -28,10 +30,134 @@ Running the gadget:
...
+        tcp_v4_connect/tcp_v6_connect -->> src_dst_per_process: 
+        activate src_dst_per_process
+    end
+
+    tcp_v4_connect/tcp_v6_connect ->> inet_stream_connect: 
+    deactivate tcp_v4_connect/tcp_v6_connect
+
+
+    rect rgba(0, 0, 255, .1)
+        Note right of inet_stream_connect: Gather all information in the kprobe<br/>and submit the event
```
alban, line 147: Do you mean in the kretprobe? Or from the kprobe?

```suggestion
        Note right of inet_stream_connect: Gather all information in the kretprobe<br/>and submit the event
```

gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -64,15 +69,26 @@ const volatile __u64 targ_min_latency_ns = 0;
...
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, MAX_ENTRIES);
 	__type(key, u32); // tid
 	__type(value, struct sock *);
 } sockets_per_process SEC(".maps");
 
+// src_dst_per_process keeps track of the src and dst information
+// between tcp_v4/6_connect and the return of inet_stream_connect
+// Since tcp_v4/6_connect is directly inside of inet_stream_connect,
+// we do not need to keep a large number of entries.
```
alban, line 84: Is it true with synchronous sockets? Could the thread block in the `tcp_v4_connect` function, and if we have more than max_entries=32 threads on the system blocked on that function, the map could get full.

burak-ok: That is also true. What size would you suggest?

burak-ok: For now I increased it from `32` to `128`

gadgets/trace_tcpconnect/README.mdx
```diff
@@ -28,10 +30,134 @@ Running the gadget:
     </TabItem>
 </Tabs>
 
-## Flags
+## Guide
 
-No flags.
```
alban, line 33: Note that by removing the flags sections, it will not be readded automatically when new flags are added in the gadget because the sed command would not find it:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/Makefile#L79

In my opinion, it is fine to keep the flags section. Or the Makefile could be updated.

gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -64,15 +69,26 @@ const volatile __u64 targ_min_latency_ns = 0;
...
 	__type(value, struct sock *);
 } sockets_per_process SEC(".maps");
 
+// src_dst_per_process keeps track of the src and dst information
+// between tcp_v4/6_connect and the return of inet_stream_connect
+// Since tcp_v4/6_connect is directly inside of inet_stream_connect,
+// we do not need to keep a huge number of entries.
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 128);
```
alban, line 87: I don't understand why max_entries is different between the two maps `sockets_per_process` and `src_dst_per_process`.

burak-ok: It might be a (premature) optimization.

`sockets_per_process` stores a pointer, whereas `src_dst_per_process` is at least 42 bytes large. `src_dst_per_process` gets written into after the `SYN` package gets sent, but with synchronous sockets the waiting is still done afterwards.
So we might need to have the larger buffer?

gadgets/trace_tcpconnect/README.mdx
```diff
@@ -34,4 +36,132 @@ No flags.
...
+
+    activate Tracee
+    Tracee ->> sys_connect: connect()
+    activate sys_connect
+
+    sys_connect ->> inet_stream_connect: 
+    activate inet_stream_connect
+
+    rect rgba(0, 0, 255, .1)
+        Note right of inet_stream_connect: Save skb in kprobe.<br/>We need it in the kretprobe,<br/>but don't have access to it
```
alban, line 132: s/skb/sock
It is not a socket buffer (skb) but a socket.

gadgets/trace_tcpconnect/program.bpf.c
```diff
@@ -210,13 +253,15 @@ static __always_inline void trace_v4(struct pt_regs *ctx, __u64 pid_tgid,
...
-	event->dst.port = dport;
-	event->src.port = BPF_CORE_READ(sk, __sk_common.skc_num);
+	if (src_dst_entry) {
+		event->src = src_dst_entry->src;
+		event->dst = src_dst_entry->dst;
+	} else {
+		// src_dst_entry is not set when tcp_v{4|6}_connect is not called, e.g.,
+		// when calling inet_stream_connect on an already connected socket. So,
+		// try to read the endpoints from the socket here.
+		read_l4endpoints_from_sock_v4(&event->src, &event->dst, sk);
```
alban, line 263: The logic does not work in some edge-cases. But it was already not working before.

Steps to reproduce:

* Start the gadget:
```
$ sudo -E ig run ghcr.io/inspektor-gadget/gadget/trace_tcpconnect:burak-trace_tcpconnect-fix --verify-image=false
RUNTIME.CONTAINERNAME                 SRC                                               DST                                              COMM                    PID        TID ERROR         
```

* Start a container. To make it reproduce faster, change the tcp timeout of the container to approximately 7 seconds:
```
$ docker run -ti --rm --privileged busybox
/ # sysctl net.ipv4.tcp_syn_retries=2
net.ipv4.tcp_syn_retries = 2
```

* Start a tcp connection to nowhere and quickly stop the process with ctrl-z (SIGSTOP), and resume it with `fg`:
```
/ # wget 1.2.3.4
Connecting to 1.2.3.4 (1.2.3.4:80)
^Z[1]+  Stopped                    wget 1.2.3.4
/ # fg
```

* Notice the gadget output after waiting a few seconds more:
```
RUNTIME.CONTAINERNAME  SRC               DST         COMM  PID     TID ERROR
boring_edison          172.17.0.3:32798  1.2.3.4:80  wget  434444  434444 error#512
boring_edison          0.0.0.0:0         1.2.3.4:0   wget  434444  434444 ETIMEDOUT
```

There is two events (one emitted at the ctrl-z, and one emitted at the end of the tcp timeout).

alban (approved):
LGTM, thanks.
Please fix s/skb/sock before merging.
IMO the edge case problem  https://github.com/inspektor-gadget/inspektor-gadget/pull/3517#discussion_r1822990039 can be fixed in a later PR.

alban:
I think `tracepoint/sched/sched_process_exit` should clean up `src_dst_per_process` and `sockets_per_process` in case `kretprobe/inet_stream_connect` does not get called due to maxactive being too low (example of [kretprobe-maxactive issue](https://github.com/iovisor/bcc/issues/1072)).

Issue #3454, opened 2024-09-06
Support for output mode for profile_blockio


mqasimsarfraz:
It seems `--output` or `-o` is ignored for `profile_blockio`:

```
./kubectl-gadget run profile_blockio -o json
latency
        µs               : count    distribution
         0 -> 1          : 0        |                                        |
         1 -> 2          : 0        |                                        |
         2 -> 4          : 0        |                                        |
         4 -> 8          : 2        |                                        |
         8 -> 16         : 8        |***                                     |
        16 -> 32         : 8        |***                                     |
        32 -> 64         : 8        |***                                     |
        64 -> 128        : 0        |                                        |
       128 -> 256        : 0        |                                        |
       256 -> 512        : 1        |                                        |
       512 -> 1024       : 8        |***                                     |
      1024 -> 2048       : 2        |                                        |
      2048 -> 4096       : 98       |****************************************|
      4096 -> 8192       : 0        |                                        |
      8192 -> 16384      : 0        |                                        |
     16384 -> 32768      : 0        |                                        |
     32768 -> 65536      : 0        |                                        |
     65536 -> 131072     : 0        |                                        |
    131072 -> 262144     : 0        |                                        |
    262144 -> 524288     : 0        |                                        |
    524288 -> 1048576    : 0        |                                        |
   1048576 -> 2097152    : 0        |                                        |
   2097152 -> 4194304    : 0        |                                        |
   4194304 -> 8388608    : 0        |                                        |
   8388608 -> 16777216   : 0        |                                        |
  16777216 -> 33554432   : 0        |                                        |
  33554432 -> 67108864   : 0        |                                        |
```

## Expected output:

Gadget run should respect the output mode requested e.g `-o json` should output json or a warning if it isn't supported.

blanquicet:
https://github.com/inspektor-gadget/inspektor-gadget/pull/3566 doesn't solve but mitigate this issue.

PR #3518, opened 2024-09-25
RFC: gadgets: Add grouping by remote and local addresses for profile_tcprtt

eiffel-fl:
Hi,

This PR adds grouping feature for `profile_tcprtt`:

```bash
$ sudo ./ig run -o json --verify-image=false profile_tcprtt:francis-profile-tcprtt --group-by-remote-addresses         
WARN[0000] No digest found for image sha256:5ec7527ba87826d36c6ff8a8fbcf7b6e34ebac32c2f63fb3f23f5875a0ec59dc 
[]
[{"addr":{"addr":"::1","version":6},"latency":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,2,0,0,0,0,0,0,0,0,0,0,0,0]},{"addr":{"addr":"::2","version":6},"latency":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,0,0,0,0,0,0,0,0,0,0,0,0]}]
```

For JSON to work, we need to remove the following from the gadget YAML:

```patch
diff --git a/gadgets/profile_tcprtt/gadget.yaml b/gadgets/profile_tcprtt/gadget.yaml
index c1fb413d2..994e1e474 100644
--- a/gadgets/profile_tcprtt/gadget.yaml
+++ b/gadgets/profile_tcprtt/gadget.yaml
@@ -5,9 +5,6 @@ documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/profile_tc
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/profile_tcprtt
 dataSources:
   tcprtt:
-    annotations:
-      cli.output: "none"
-      metrics.print: "true"
     fields:
       latency:
         annotations
```

This is related to https://github.com/inspektor-gadget/inspektor-gadget/issues/3454.

Moreover, histogram output needs to be tweaked to print the address of the histogram like what we do for builtins.
I thought to use the metric description for this purpose, but not whether this is correct.

Best regards.

### DIFF
--- a/gadgets/profile_tcprtt/gadget.yaml
+++ b/gadgets/profile_tcprtt/gadget.yaml
@@ -5,9 +5,6 @@ documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/profile_tc
 sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/profile_tcprtt
 dataSources:
   tcprtt:
-    annotations:
-      cli.output: "none"
-      metrics.print: "true"
     fields:
       latency:
         annotations:
@@ -29,3 +26,11 @@ params:
       key: ms
       defaultValue: "false"
       description: Convert latency to milliseconds, by default it uses microseconds.
+    targ_laddr_hist:
+      key: group-by-local-addresses
+      defaultValue: "false"
+      description: Show an histogram for each local addresses.
+    targ_raddr_hist:
+      key: group-by-remote-addresses
+      defaultValue: "false"
+      description: Show an histogram for each remote addresses.
--- a/gadgets/profile_tcprtt/program.bpf.c
+++ b/gadgets/profile_tcprtt/program.bpf.c
@@ -21,6 +21,8 @@
 #define AF_INET 2 /* IP version 4			*/
 #define AF_INET6 10 /* IP version 6			*/
 
+const volatile bool targ_laddr_hist = false;
+const volatile bool targ_raddr_hist = false;
 const volatile __u16 targ_sport = 0;
 const volatile __u16 targ_dport = 0;
 const volatile __u32 targ_saddr = 0;
@@ -29,6 +31,8 @@ const volatile __u8 targ_saddr_v6[IPV6_LEN] = {};
 const volatile __u8 targ_daddr_v6[IPV6_LEN] = {};
 const volatile bool targ_ms = false;
 
+GADGET_PARAM(targ_laddr_hist);
+GADGET_PARAM(targ_raddr_hist);
 GADGET_PARAM(targ_sport);
 GADGET_PARAM(targ_dport);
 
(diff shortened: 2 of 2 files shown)


blanquicet:
From what I can see:

1. This PR breaks current histogram output.
2. The JSON output provided by this PR does not contain information about intervals (I'm not sure how useful it's without such information). Instead, I believe the JSON output should be similar to what we provide with the built-in flavor:

    ```json
    {"histograms":[{"unit":"µs","intervals":[{"count":0,"start":0,"end":1},{"count":0,"start":2,"end":3},{"count":0,"start":4,"end":7},{"count":0,"start":8,"end":15},{"count":0,"start":16,"end":31},{"count":0,"start":32,"end":63},{"count":0,"start":64,"end":127},{"count":0,"start":128,"end":255},{"count":0,"start":256,"end":511},{"count":0,"start":512,"end":1023},{"count":0,"start":1024,"end":2047},{"count":0,"start":2048,"end":4095},{"count":2,"start":4096,"end":8191},{"count":7,"start":8192,"end":16383},{"count":10,"start":16384,"end":32767},{"count":8,"start":32768,"end":65535},{"count":7,"start":65536,"end":131071},{"count":0,"start":131072,"end":262143},{"count":4,"start":262144,"end":524287}],"address":"******","addressType":"All Addresses","average":61299.68421052631}]}
    ```

Therefore, IMO, we shouldn't go ahead with this feature until we support a proper JSON output for histograms. To do that, I think we have two options:

1. Make otel-metrics operator aware of the requested output mode. I don't like it because we'd create a dependency between operators.
2. Make cli operator aware of histograms as output mode. I think it could receive an array of bytes that can be unmarshalled into an array of structs like this:

    ```go
	type Histogram struct {
		Unit      Unit       `json:"unit,omitempty"`
		Intervals []Interval `json:"intervals,omitempty"`
		Labels    []string   `json:"labels,omitempty"`
	}
    ```

    Which would be basically an extended [Histogram](https://github.com/inspektor-gadget/inspektor-gadget/blob/8fed16feb150fcbd6e6e6ec2d7460d22c7ccb050/pkg/histogram/histogram.go#L39-L44) to carry also `Labels` where we can store a set of `key=values` pairs for information related to the histogram, e.g., `"address":"******","addressType":"All Addresses","average":61299.68421052631`. Then the cli opeartor can use [Histogram.String()](https://github.com/inspektor-gadget/inspektor-gadget/blob/8fed16feb150fcbd6e6e6ec2d7460d22c7ccb050/pkg/histogram/histogram.go#L82) to render it (if `histogram` output was requested) or print it as JSON, jsonpretty or YAML. 

WDYT?

blanquicet:
FYI, I created https://github.com/inspektor-gadget/inspektor-gadget/pull/3566 which is kind of related to this PR.

PR #3519, opened 2024-09-25
gadgets: Fix headless test

mauriciovasquezbernal:
The cosign key is not available when running the tests on the CI for external contributors or dependabot, it means that this test will fail because the signature can't be found.

Introduce logic to pass IG_FLAGS to "ig daemon" to disable the image verification when needed.

Fixes: c837450e4e4b ("add tests for ig --daemon in headless mode")

See #3514

cc @flyth @burak-ok


flyth (approved):
LGTM. Thanks for the fix, I wasn't aware of that!

PR #3528, opened 2024-09-30
operators/ebpf: Do not add ebpf objects to context

mauriciovasquezbernal:
Adding ebpf objects like maps and variables to the context is not needed. This commit removes that to allow users of the API to pass their own mount ns map, for instance, for testing purposes.

```
	filterMap := utilstest.CreateMntNsFilterMap(t, runner.Info.MountNsID)
	gadgetCtx.SetVar(gadgets.MntNsFilterMapName, filterMap)
	gadgetCtx.SetVar(gadgets.FilterByMntNsName, true)
```

Will be used in #3502 later on.

### DIFF
--- a/pkg/operators/ebpf/maps.go
+++ b/pkg/operators/ebpf/maps.go
@@ -61,9 +61,6 @@ func (i *ebpfInstance) populateMap(t btf.Type, varName string) error {
 
 	i.vars[varName] = newVar
 
-	// Set variable to nil pointer to map, so it's present
-	var nilVal *ebpf.Map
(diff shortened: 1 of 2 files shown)


mauriciovasquezbernal:
This idea doesn't work at the socket enricher operator needs to check if the sockets map is used to active itself:

https://github.com/inspektor-gadget/inspektor-gadget/blob/4f88e74d9225bf483d48f0cee3e658ad178c56a0/pkg/operators/socketenricher/socketenricher.go#L168

Will think about another solution for the tests.

PR #3535, opened 2024-10-02
charts: Set allowPrivilegeEscalation to false.

eiffel-fl:
This is a not a silver bullet, as we already have CAP_SYS_ADMIN, but it does no harm.


mqasimsarfraz (commented):
It seems `allowPrivilegeEscalation` will always be set `true` with `SYS_ADMIN` and only reason it works is because of a [bug](https://github.com/kubernetes/kubernetes/issues/119568) so perhaps we shouldn't make this change. What do you think?

References:
- https://github.com/kubernetes/kubernetes/issues/119568
- https://github.com/kubernetes/kubernetes/issues/119569
- https://github.com/istio/istio/pull/52540

Issue #3193, opened 2024-07-22
[RFE] Develop event generation capabilities


burak-ok:
Inspektor Gadget enables users to inspect Linux containers and Kubernetes workloads, offering the powerful capability to monitor traffic for currently running applications. Nevertheless, the ability to generate specific events would be a great addition to the existing features. For example, if users could trigger a DNS request from a specific pod or make an HTTP request to a particular endpoint, Inspektor Gadget could check if the request succeeded or not. Users would then be notified if there are errors/outages/problems in the cluster.

The issue could be solved in this order:
1. Explore the best way to implement event generation, external program vs implementing it within Inspektor Gadget. 
2. Develop DNS/HTTP (for example) event generator and see how it works with gadgets
3. Use the event generation in our integration testing
4. Create event generators for all other gadgets

HarshitNagpal29:
Hi @burak-ok,

I hope you're doing well. I'm really excited about this project. I have a few questions to ensure I align with the project's goals:

1. Should the event generation be an external program or integrated within Inspektor Gadget?
2. Beyond DNS and HTTP, are there other high-priority events to consider?

Any further guidance or resources you could provide would be greatly appreciated. I feel confident that I have the necessary skills to work on this project and look forward to contributing.

Thank you!

rahulrana701:
@burak-ok the project sounds pretty cool and interesting would love to dive deep into this project with further guidance.

burak-ok:
Nice to see that this issue is interesting for people :)

> 1. Should the event generation be an external program or integrated within Inspektor Gadget?

We are also not sure about this one. We think it might be better to have it integrated into the program itself. So it is easier for users to use. But we are open about discussing other options too

> 2. Beyond DNS and HTTP, are there other high-priority events to consider?

Other events might be tcp traffic, mounts, file creation and in general mostly everything we have a gadget for. But we do not have priorities for them

anshjain16:
> Nice to see that this issue is interesting for people :)
> 
> > 1. Should the event generation be an external program or integrated within Inspektor Gadget?
> 
> We are also not sure about this one. We think it might be better to have it integrated into the program itself. So it is easier for users to use. But we are open about discussing other options too
> 
> > 2. Beyond DNS and HTTP, are there other high-priority events to consider?
> 
> Other events might be tcp traffic, mounts, file creation and in general mostly everything we have a gadget for. But we do not have priorities for them

[...]

PR #3538, opened 2024-10-02
discussion on approaches to implement event-generator

h4l0gen:
# Discussion on approaches to implement event-generation utility

## Testing done
Yes

### DIFF
--- a/cmd/common/oci.go
+++ b/cmd/common/oci.go
@@ -33,6 +33,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	clioperator "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/cli"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/combiner"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/eventgen"
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/limiter"
 	ocihandler "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/oci-handler"
 	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/otel-metrics"
@@ -160,8 +161,8 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 			}
 			ops = append(ops, op)
 		}
-		ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator)
-
+		// ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator)
+		ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator, eventgen.EventGen)
 		imageName := actualArgs[0]
 		if grpcrt, ok := runtime.(*grpcruntime.Runtime); ok && commandMode == CommandModeAttach {
 			instances, ambiguous, notfound, err := findGadgetInstances(grpcrt, runtimeParams, []string{imageName})
@@ -252,8 +253,8 @@ func NewRunCommand(rootCmd *cobra.Command, runtime runtime.Runtime, hiddenColumn
 		for _, op := range operators.GetDataOperators() {
 			ops = append(ops, op)
 		}
-		ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator)
-
+		// ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator)
+		ops = append(ops, clioperator.CLIOperator, combiner.CombinerOperator, eventgen.EventGen)
 		timeoutDuration := time.Duration(timeoutSeconds) * time.Second
 
 		imageName := args[0]
--- a/pkg/operators/eventgen/dns.go
+++ b/pkg/operators/eventgen/dns.go
@@ -1,0 +1,119 @@
+package eventgen
+
+import (
+    "bytes"
+    "context"
+    "fmt"
+    "time"
+    corev1 "k8s.io/api/core/v1"
+    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+    "k8s.io/client-go/kubernetes"
+    "k8s.io/client-go/kubernetes/scheme"
+    "k8s.io/client-go/rest"
+    "k8s.io/client-go/tools/remotecommand"
+)
+
+type DNSGenerator struct {
+    clientset *kubernetes.Clientset
+    config    *rest.Config
+}
+
+func NewDNSGenerator(config *rest.Config) (Generator, error) {
+    clientset, err := kubernetes.NewForConfig(config)
+    if err != nil {
+        return nil, fmt.Errorf("failed to create Kubernetes client: %v", err)
+    }
+    return &DNSGenerator{
+        clientset: clientset,
+        config:    config,
+    }, nil
+}
+
+func (d *DNSGenerator) Generate(domain string) (string, error) {
+    namespace := "default"
+    podName := fmt.Sprintf("dns-test-pod-%d", time.Now().Unix())
+
+    if err := d.createAndWaitForPod(namespace, podName); err != nil {
+        return "", err
+    }
+
+    if err := d.executeDNSLookup(namespace, podName, domain); err != nil {
+        return "", err
+    }
+
+    return podName, nil
+}
+
+func (d *DNSGenerator) Cleanup(podName string) error {
+    namespace := "default"
+    err := d.clientset.CoreV1().Pods(namespace).Delete(context.Background(), podName, metav1.DeleteOptions{})
+    if err != nil {
+        return fmt.Errorf("failed to delete pod %s: %v", podName, err)
+    }
+    fmt.Printf("Successfully deleted pod %s in namespace %s\n", podName, namespace)
+    return nil
+}
+
+func (d *DNSGenerator) createAndWaitForPod(namespace, name string) error {
+    pod := &corev1.Pod{
+        ObjectMeta: metav1.ObjectMeta{Name: name},
+        Spec: corev1.PodSpec{
+            Containers: []corev1.Container{{
+                Name:    "dns-test",
+                Image:   "busybox",
+                Command: []string{"sh", "-c", "while true; do sleep 3600; done"},
+            }},
+        },
+    }
+
+    _, err := d.clientset.CoreV1().Pods(namespace).Create(context.Background(), pod, metav1.CreateOptions{})
+    if err != nil {
+        return fmt.Errorf("failed to create pod: %v", err)
+    }
+
+    for i := 0; i < 60; i++ {
+        pod, err := d.clientset.CoreV1().Pods(namespace).Get(context.Background(), name, metav1.GetOptions{})
+        if err != nil {
+            return err
+        }
+        if pod.Status.Phase == corev1.PodRunning {
+            return nil
+        }
+        time.Sleep(time.Second)
+    }
+    return fmt.Errorf("pod did not become ready within 60 seconds")
+}
+
+func (d *DNSGenerator) executeDNSLookup(namespace, podName, domain string) error {
+    req := d.clientset.CoreV1().RESTClient().Post().
+        Resource("pods").
+        Name(podName).
+        Namespace(namespace).
+        SubResource("exec").
+        VersionedParams(&corev1.PodExecOptions{
+            Command: []string{"sh", "-c", fmt.Sprintf("echo 'Performing DNS lookup for %s'; nslookup %s; echo 'DNS lookup completed'", domain, domain)},
+            Stdout:  true,
+            Stderr:  true,
+        }, scheme.ParameterCodec)
+
+    exec, err := remotecommand.NewSPDYExecutor(d.config, "POST", req.URL())
+    if err != nil {
+        return fmt.Errorf("failed to create executor: %v", err)
+    }
+
+    var stdout, stderr bytes.Buffer
+    err = exec.Stream(remotecommand.StreamOptions{
+        Stdout: &stdout,
+        Stderr: &stderr,
+    })
+    if err != nil {
+        return fmt.Errorf("failed to execute command: %v", err)
+    }
+
+    fmt.Printf("DNS lookup output:\n%s\n", stdout.String())
+    if stderr.Len() > 0 {
+        fmt.Printf("DNS lookup error:\n%s\n", stderr.String())
+    }
+
+    return nil
+}
--- a/pkg/operators/eventgen/eventgen.go
+++ b/pkg/operators/eventgen/eventgen.go
@@ -1,0 +1,157 @@
+package eventgen
+
+import (
+	"fmt"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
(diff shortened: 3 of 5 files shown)


h4l0gen:
- flag approach: dns event-generator output
```bash
h4logen@kapil:~/new-ig-arch/inspektor-gadget$ go run --exec "sudo -E" ./cmd/ig run trace_dns --filter name==example.com. --eventgen-enable=true --eventgen-type=dns --eventgen-target=example.com
RUNTIME.CONTAINERNAME          SRC                                     DST                                     COMM                   PID QR QTYPE           NAME                          RCODE    ADDRESSES      
WARN[0014] Replacement map "gadget_sockets" incompatible: ValueSize: 8272 changed to 64: map spec is incompatible with existing map 
WARN[0014] Is trace_dns compatible with ig 0.0.0?       
INFO[0014] Starting EventGen with type: dns, target: example.com 
minikube                       10.244.0.62:48145                       10.96.0.10:53                                                    0 Q  A               example.com.                                          
minikube                       10.244.0.62:48145                       10.244.0.26:53                                                   0 Q  A               example.com.                                          
DNS lookup output:
Performing DNS lookup for example.com
Server:		10.96.0.10
Address:	10.96.0.10:53

Non-authoritative answer:
Name:	example.com
Address: 93.184.215.14

Non-authoritative answer:
Name:	example.com
Address: 2606:2800:21f:cb07:6820:80da:af6b:8b2c

DNS lookup completed

INFO[0019] Generated dns event using pod dns-test-pod-1727700258 
INFO[0019] Starting EventGen with type: dns, target: example.com 
minikube                       10.244.0.62:48145                       10.96.0.10:53                                                    0 Q  AAAA            example.com.                                          
minikube                       10.244.0.62:48145                       10.244.0.26:53                                                   0 Q  AAAA            example.com.                                          
minikube                       10.244.0.26:40171                       192.168.49.1:53                                                  0 Q  AAAA            example.com.                                          
minikube                       10.244.0.26:48557                       192.168.49.1:53                                                  0 Q  A               example.com.                                          
minikube                       192.168.49.1:53                         10.244.0.26:48557                                                0 R  A               example.com.                  Success  93.184.215.14  
minikube                       192.168.49.1:53                         10.244.0.26:48557                                                0 R  A               example.com.                  Success  93.184.215.14  
minikube                       192.168.49.1:53                         10.244.0.26:40171                                                0 R  AAAA            example.com.                  Success  2606:2800:21f:…
minikube                       192.168.49.1:53                         10.244.0.26:40171                                                0 R  AAAA            example.com.                  Success  2606:2800:21f:…
minikube                       10.244.0.26:53                          10.244.0.62:48145                                                0 R  A               example.com.                  Success  93.184.215.14  
minikube                       10.96.0.10:53                           10.244.0.62:48145                                                0 R  A               example.com.                  Success  93.184.215.14  
minikube                       10.244.0.26:53                          10.244.0.62:48145                                                0 R  AAAA            example.com.                  Success  2606:2800:21f:…
minikube                       10.96.0.10:53                           10.244.0.62:48145                                                0 R  AAAA            example.com.                  Success  2606:2800:21f:…
minikube                       10.244.0.63:60827                       10.96.0.10:53                                                    0 Q  A               example.com.                                          
minikube                       10.244.0.63:60827                       10.244.0.26:53                                                   0 Q  A               example.com.                                          
minikube                       10.244.0.63:60827                       10.96.0.10:53                                                    0 Q  AAAA            example.com.                                          
DNS lookup output:
Performing DNS lookup for example.com
Server:		10.96.0.10
Address:	10.96.0.10:53

Name:	example.com
Address: 93.184.215.14

Name:	example.com
Address: 2606:2800:21f:cb07:6820:80da:af6b:8b2c

DNS lookup completed

INFO[0023] Generated dns event using pod dns-test-pod-1727700263 
minikube                       10.244.0.63:60827                       10.244.0.26:53                                                   0 Q  AAAA            example.com.                                          
minikube                       10.244.0.26:53                          10.244.0.63:60827                                                0 R  A               example.com.                  Success  93.184.215.14  
minikube                       10.96.0.10:53                           10.244.0.63:60827                                                0 R  A               example.com.                  Success  93.184.215.14  
minikube                       10.244.0.26:53                          10.244.0.63:60827                                                0 R  AAAA            example.com.                  Success  2606:2800:21f:…
minikube                       10.96.0.10:53                           10.244.0.63:60827                                                0 R  AAAA            example.com.                  Success  2606:2800:21f:…
^CSuccessfully deleted pod dns-test-pod-1727700263 in namespace default
INFO[0041] Cleaned up event pod dns-test-pod-1727700263 
Successfully deleted pod dns-test-pod-1727700258 in namespace default
INFO[0041] Cleaned up event pod dns-test-pod-1727700258`
```


I still need  to polish this output, but this is overall how it will work.

h4l0gen:
Hi folks, I am Kapil Sharma, working on [event-generation](https://github.com/inspektor-gadget/inspektor-gadget/issues/3193) with [@Burak Ok](https://kubernetes.slack.com/team/U042GTWLDH8) and [@Qasim Sarfraz](https://kubernetes.slack.com/team/U01S39E4D3J). I have came up with two approaches to implement event-generation within Inspektor-gadget. These are the most promising approaches I found (please feel free to suggest any other approach you may have):
- Flag approach: using eventgen as a flag that can be used while running gadgets. For example, using the DNS event generator:

`$ go run --exec "sudo -E" ./cmd/ig run trace_open --eventgen-enable=true --eventgen-type=dns --eventgen-target=example.com`

Its a concise approach simply can run event-generator along with gadgets. Here we can run gadget and enable event-generator, so event-generator will create events and gadget will show that. I have included the complete output of this approach in a [draft PR](https://github.com/inspektor-gadget/inspektor-gadget/pull/3538#issuecomment-2389136897), please take a look.

- Sub-command Approach: I have previously implemented this with the July version (with kubectl-gadget).

using eventgen  as a Sub-command allows it to run separately from the gadget, meaning both processes are independent and don't rely on each other.
we can run gadget like this

`$ ./kubectl-gadget trace dns`

In another terminal, we can run the DNS event generator to generate events:

`$ ./kubectl-gadget eventgen dns example.com`

The gadget will catch and display events generated by the DNS event generator.I would like your suggestions on which approach would be more suitable.I have created a [draft PR](https://github.com/inspektor-gadget/inspektor-gadget/pull/3538) for the first approach to start discussion. Thank you!

blanquicet:
Thanks for creating this space to discuss. I have some comments already.

Regarding flags vs sub-command:

- I like this feature being implemented as flags because it allows to synchronize the event generation with the gadget lifecycle, i.e., start generating events only after the gadget was actually started and (in a follow-up PR) validate output once gadget is stopped. I'm a little bit worry about the number of flags, but after merging [#3515](https://github.com/inspektor-gadget/inspektor-gadget/pull/3515), users will be able to change this:

    ```bash
    ig run trace_dns --filter name==example.com. --eventgen-enable=true --eventgen-type=dns --eventgen-target=example.com
    ```

  into this:

    ```yaml
    version: 1
    kind: instance-spec
    imageName: trace_dns
    name: capture_dns_events
    paramValues:
      operator.filter.filter: "name==example.com."
      operator.eventgen.eventgen-enable: true
      operator.eventgen.eventgen-type: dns
      operator.eventgen.eventgen-target: example.com
    ``` 

- On the other side, I like the idea of sub-command because it allows to use one single event generation for more than one gadget. For instance, HTTPS events with trace_dns, trace_tcpconnect, or downloading/uploading test files with trace_tcpdrop, trace_tcpretrans. However, we will lost the ability to sync with gadget lifecycle.
- In the end, I think I prefer the flag implementation.

Regarding operator implementation:

- The generation of events must happen on server side.
- The environment where events will be generated shouldn't be tight to Kubernetes (probably what you did was just an example). I suggest re-using the [Container interface](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/testutils/container_interface.go#L39) we created for testing. With few changes, it'll allow us to run commands on docker/containerd containers (even though in this case we don't care about the runtime), but also in Kubernetes containers by keeping a common interface. In addition, we can extend it to allow creating container with specific linux namespaces. For instance, to create container with the host network linux namespace for generating DNS or HTTPS events for `ig` running as standalone with the `--host` flag.

In general, I think we can start by providing a subset of eventgen-type: dns and https. But at a certain point, as we did for gadgets, we should allow people running custom things like bash scripts or even WASM. Let's keep it in the back of our mind.

mqasimsarfraz:
> The generation of events must happen on server side.

This is a great discussion point. One of the main concerns with server side generation would mean to allow `gadget` pod (via RBAC) to be able to `create/delete` pods in all the namespace which I am not sure we should be doing. Having said that we do need server side generation for existing pods/containers (switching to particular namespace and generating events from there) so would be nice if we can find a way to do it in the same way for both. One idea to avoid touching RBAC permission would be to skip talking to API server for pods but rather use CRI API in Kubernetes and Container Runtime APIs for runtimes using [Container interface](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/testutils/container_interface.go#L39) like you mentioned. Any thoughts on this?

blanquicet:
> > The generation of events must happen on server side.
> 
> This is a great discussion point. One of the main concerns with server side generation would mean to allow `gadget` pod (via RBAC) to be able to `create/delete` pods in all the namespace which I am not sure we should be doing. Having said that we do need server side generation for existing pods/containers (switching to particular namespace and generating events from there) so would be nice if we can find a way to do it in the same way for both. One idea to avoid touching RBAC permission would be to skip talking to API server for pods but rather use CRI API in Kubernetes and Container Runtime APIs for runtimes using [Container interface](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/testutils/container_interface.go#L39) like you mentioned. Any thoughts on this?
>

Oh, raised a great point. Yeah, I would avoid modifying the RBAC for the ig-k8s pod. So, at this point, we will need to differentiate the implementation:
- ig and ig+gadgetctl: On server side
- kubectl-gadget: On client side

WDYT?

blanquicet:
> Having said that we do need server side generation for existing pods/containers (switching to particular namespace and generating events from there)
> 
Why do we need this?

mqasimsarfraz:
> > Having said that we do need server side generation for existing pods/containers (switching to particular namespace and generating events from there)
> 
> Why do we need this?

Imagine if you want to troubleshoot DNS for existing pod, in this case it would be helper if we can generate the request from the given pod and see it works fine. Does that make sense?

burak-ok:
> On the other side, I like the idea of sub-command because it allows to use one single event generation for more than one gadget. For instance, HTTPS events with trace_dns, trace_tcpconnect, or downloading/uploading test files with trace_tcpdrop, trace_tcpretrans. However, we will lost the ability to sync with gadget lifecycle.

I think that there would be no problem to implement such features for the flag approach too. The event-gen should not be directly tied to the gadget.
So soemthing like `kubectl gadget eventgen http` or `kubectl gadget run trace_tcp --eventgen http` should be equivalent

The only problem we might have (with both approaches) comes when we want to verify the gadget output. The verification is gadget specific, but not the event generation

blanquicet:
> > On the other side, I like the idea of sub-command because it allows to use one single event generation for more than one gadget. For instance, HTTPS events with trace_dns, trace_tcpconnect, or downloading/uploading test files with trace_tcpdrop, trace_tcpretrans. However, we will lost the ability to sync with gadget lifecycle.
> 
> I think that there would be no problem to implement such features for the flag approach too. The event-gen should not be directly tied to the gadget. 
> 
If it's an operator, by default, it's tied to the gadget's lifecycle, i.e., it'll start generating events when gadget starts and stop doing it when gadget stops. In fact, the `eventgen-enable` would be an instance parameter, so users can only ask to start generating events when running a given gadget. How to do that if you have more than one?

> So soemthing like `kubectl gadget eventgen http` or `kubectl gadget run trace_tcp --eventgen http` should be equivalent
> 
If I didn't miss anything on my above comment, they are not.

> The only problem we might have (with both approaches) comes when we want to verify the gadget output. The verification is gadget specific, but not the event generation. (Of course generating file events for network gadgets makes no sense, etc..)
>
Nope, for same reason I described above. If `eventgen-enable`, `eventgen-type`, etc., are instance parameters, each eventgen operator instance is associated to a specific gadget, so the verification.

burak-ok:
I thought you meant using (for example) an http event generator for different gadgets in different instances.

Did you mean starting the event generator and then in parallel use multiple gadgets to capture the events?

blanquicet:
> I thought you meant using (for example) an http event generator for different gadgets in different instances.
> 
> Did you mean starting the event generator and then in parallel use multiple gadgets to capture the events?
>
Yes. For instance, one may want to know if a host is able to reach a specific endpoint and with what latency. So, they'd generate HTTP events and then run, in parallel, `trace_dns` and `trace_tcpconnect` (both specs defined in a [manifest file](https://github.com/inspektor-gadget/inspektor-gadget/pull/3485)) to know:

- Whether the URL was resolved or not and, in case, the error
- Whether the TCP connection got established or not and, in case, the error
- The TCP and DNS latency

burak-ok:
> That's why I concluded saying it prefer the flag implementation, at least we will have a perfect synchronization for one single gadget.

(Thinking out loud). Maybe we could control with a flag how many gadgets should run before the event gen actually starts generating. Our operator framework allows for that. But it gets complicated and in the way of other things maybe

blanquicet:
> > That's why I concluded saying it prefer the flag implementation, at least we will have a perfect synchronization for one single gadget.
> 
> (Thinking out loud). Maybe we could control with a flag how many gadgets should run before the event gen actually starts generating. Our operator framework allows for that. But it gets complicated and in the way of other things maybe
>
I think it'd imply "synchronization" between operator's instances, which doesn't exist right now. I don't know how complex it would be. However, let's keep it in mind. I suggest adding it to future implements:

> let's track all the improvements we are planning in https://github.com/inspektor-gadget/inspektor-gadget/issues/3193:
>
> - The eventgen operator could also validate gadget output and disable normal gadget output. It could just say, "all the generated events were captured" or "X was missing" (and print the gadget output).
> - Allow generating events on existing pods/containers.
>
- Support using one single event generator for more than one gadget running in parallel and capturing the events. An example of use case is described [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/3538#issuecomment-2394183007).

blanquicet:
To summarize, I think we kind of agreed on the following initial proposal:

- Event generator is associated to one single gadget using operator's parameters, i.e., CLI flags or `paramValues` in [manifest file](https://github.com/inspektor-gadget/inspektor-gadget/pull/3515)).
- Events are generated on server side, except for Kubernetes.
- Events are generated on pods/containers created by the event generator itself and not on existing pods/containers (For now).
- Event generator doesn't validate output (For now).

What do you all think?

mqasimsarfraz:
> To summarize, I think we kind of agreed on the following initial proposal:
> 
> * Event generator is associated to one single gadget using operator's parameters, i.e., CLI flags or `paramValues` in [manifest file](https://github.com/inspektor-gadget/inspektor-gadget/pull/3515).
> * Events are generated on server side, except for Kubernetes.
> * Events are generated on pods/containers created by the event generator itself and not on existing pods/containers (For now).
> * Event generator doesn't validate output (For now).
> 
> What do you all think?

+1, sounds good to me. If we all agree I will update https://github.com/inspektor-gadget/inspektor-gadget/issues/3193 for future improvements!

blanquicet:
@h4l0gen what do you think?

h4l0gen:
Hi, I have been following entire discussion, I also liked the synchronization idea so +1 from my side to flag approach. Just two concerns/doubts I have in my mind:
> Events are generated on server side

`server side` concept still not much clear to me, It would be great help if you guys can guide me in this. 

secondly, How we are planning about validation, are we going to compare output with expected output. Something like [this](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_dns/test/trace_dns_test.go#L134).

PR #3540, opened 2024-10-04
operators/localmanager: Do not print warning if the runtime is not found

mauriciovasquezbernal:
In most of the cases users don't have all runtimes available, these warnings are very noisy. This commit updates the logic to avoid printing any warning if the runtime path is not set, if it's set then an error is returned.

### Before 

```bash 
# warnings printed 
$ sudo ig run trace_open:latest --pull always
RUNTIME.CONTAINERNAME     COMM                 PID        TID        UID        GID  FD FNAME                    MODE          ERROR

# even more warnings printed when explicitly setting the path 
$ sudo ig run trace_open:latest --pull always --docker-socketpath=nonexistingpath
WARN[0000] Ignoring runtime "docker" with non-existent socketPath "nonexistingpath"
RUNTIME.CONTAINERNAME     COMM                 PID        TID        UID        GID  FD FNAME                    MODE          ERROR
```

### After

```bash 
# not warning is printed 
$ sudo ig run trace_open:latest --pull always
RUNTIME.CONTAINERNAME     COMM                 PID        TID        UID        GID  FD FNAME                    MODE          ERROR

# error out if explicitly provided path doesn't exist
$ sudo ig run trace_open:latest --pull always --docker-socketpath=nonexistingpath
WARN[0000] error initializing operator LocalManager: runtime "docker" with non-existent socketPath "nonexistingpath"
Error: starting operators: pre-starting operator "LocalManager": container-collection isn't available
```

cc @blixtra

### DIFF
--- a/pkg/operators/localmanager/localmanager.go
+++ b/pkg/operators/localmanager/localmanager.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/cilium/ebpf"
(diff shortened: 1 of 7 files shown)


pkg/operators/localmanager/localmanager.go
```diff
@@ -182,45 +183,45 @@ func (l *LocalManager) CanOperateOn(gadget gadgets.GadgetDesc) bool {
...
+		socketPathIsSet := socketPathParam.IsSet()
 
 		cleanSocketPath, err := securejoin.SecureJoin(host.HostRoot, socketPath)
 		if err != nil {
 			log.Debugf("securejoin failed: %s", err)
+			continue
 		}
 
 		if _, err := os.Stat(cleanSocketPath); err != nil {
-			log.Warnf("Ignoring runtime %q with non-existent socketPath %q", runtimeName, socketPath)
```
mqasimsarfraz, line 222: Perhaps we should keep a debug log?

pkg/operators/localmanager/localmanager.go
```diff
@@ -182,45 +183,45 @@ func (l *LocalManager) CanOperateOn(gadget gadgets.GadgetDesc) bool {
...
-			continue partsLoop
+			if socketPathIsSet {
```
mqasimsarfraz, line 221: Shouldn't we check for if the user changes the runtime param? like I don't see any error/warning for:

```
$ go run -exec "sudo -E" ./cmd/ig run trace_open -r cri-o
...
```

mqasimsarfraz (commented):
Looking good! Just couple of comments!

pkg/operators/localmanager/localmanager.go
mauriciovasquezbernal: Makes sense, updated it. Thanks for the idea!

alban (commented):

| Benchmark suite | Current: eb6d8582c0356381b3033e52af4a3df6673f43a1 | Previous: fe9501f4d10704eb76efe5b698821395bf7905cc | Ratio |
|-|-|-|-|
| `BenchmarkAllGadgetsWithContainers/container0/advise-seccomp` | `29399257` ns/op | `12267658` ns/op | `2.40` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-bind` | `200425268` ns/op | `100122452` ns/op | `2.00` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-fsslower` | `482106776` ns/op | `234323104` ns/op | `2.06` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-mount` | `132633654` ns/op | `65051583` ns/op | `2.04` |

CC: @alban

mqasimsarfraz (approved):
LGTM!

Issue #3003, opened 2024-06-11
gadgets: Implement support for advise gadgets as image-based ones


mauriciovasquezbernal:
The advise gadgets capture events and suggest different configurations. We should also support them as image-based gadgets. 

This support requires two different things:
- access contents from ebpf maps: this be done by using wasm: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/003-wasm-support.md#handle-ebpf-objects
- print custom output to the terminal (or to a file): https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/003-wasm-support.md#define-new-output-modes.

Some of the gadgets, like advise seccomp-profile, have integrations to directly apply the profile to the cluster: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/builtin-gadgets/advise/seccomp-profile.md#integration-with-kubernetes-security-profiles-operator. Implement this support is difficult and its out of scope for now, the profiles can be printed to the terminal (or to a file) and then the user will apply them.

TODO: We should think about the UX of these gadgets before porting them. Currently they require a `start`, `stop` operations, however I'm not sure we want to continue using this approach.

mauriciovasquezbernal:
Closing in favor of two more granular issues: #3891, #3890

PR #3566, opened 2024-10-08
pkg/operators/otel-metrics: Hide render datasource from gadget users

blanquicet:
# Hide render datasource from gadget users

Making the render datasource (used by the otel-metrics to print histograms on CLI side) visible to users provides some issues and unnecessary complexity:

- Gadget developer needs to apply some annotations to the datasource they created and some others to a datasource named `renderedmetrics` (which they didn't created). For instance, for the `trace_tcprtt` gadget:

    ```yaml
    dataSources:
      tcprtt:
        annotations:
          cli.output: "none"
          metrics.print: "true"
        fields:
          latency:
            annotations:
              metrics.unit: µs
      renderedmetrics:
        annotations:
    ```

- Available fields on help message is confusing:

    ```bash
                                                 "tcprtt" (data source):
                                                   latency
                                                   unused

                                                 "renderedmetrics" (data source):
                                                   text
                                                (default "tcprtt:unused,latency;renderedmetrics:text")
    ```

- [#3454](https://github.com/inspektor-gadget/inspektor-gadget/issues/3454) Additionally, any output mode different than raw is ignored and it always print the raw output (histogram)

   ```bash
   $ go run --exec "sudo" ./cmd/ig/... run profile_tcprtt -o json
   latency
           µs               : count    distribution
            0 -> 1          : 0        |                                        |
   [...]
   ```
This PR proposes to hide the `renderedmetrics` datasource from users by disabling the original data source and applying its annotations on the one for render the output. Doing so, we simplify the yaml by setting all the annotations to one datasource and preventing users from setting `cli.output: "none"` (which was quite strange because it's not even an annotation):

```yaml
dataSources:
  tcprtt:
    annotations:
      metrics.print: "true"
    fields:
      latency:
        annotations:
          metrics.unit: µs
```

The available flags now look cleaner:

```bash
                                             "tcprtt-rendered" (data source):
                                               text
                                            (default "text")
```

Yeah, I know there's still some magic with this `-rendered` suffixed data source, but I tried to document it in a new gadget-devel documentation: [Displaying Output](https://github.com/inspektor-gadget/inspektor-gadget/blob/jose/render-datasource/docs/gadget-devel/output.md).

In addition, any output different than `raw` fails, mitigating [#3454](https://github.com/inspektor-gadget/inspektor-gadget/issues/3454):

```bash
$ go run --exec "sudo" ./cmd/ig/... run $LOCAL_REPO/profile_tcprtt:outputmode --insecure-registries=$LOCAL_REPO --verify-image=false --help
  -o, --output string                          Specifies output mode for all ("mode1") or per data source ("datasource:mode1,datasource2:mode2")
                                                 Supported data sources / output modes:
                                                   "tcprtt-rendered" (data source):
                                                     histogram
                                                (default "histogram")

$ go run --exec "sudo" ./cmd/ig/... run $LOCAL_REPO/profile_tcprtt:outputmode --insecure-registries=$LOCAL_REPO --verify-image=false -o json
[...]
WARN[0000] output mode "json" for data source "tcprtt-rendered" is not supported; skipping data source
```

### TODOs to be managed on other PRs (I'll create issues for them)

- [ ] Uniform flag format of fields supporting values per datasource. Some of them use "," (e.g., `map-fetch-interval` and `map-fetch-count`) and some other ";" (e.g., `fields`) as separator.
- [ ] Support more human readable manifest format for fields supporting values per datasouce. [Here](https://github.com/inspektor-gadget/inspektor-gadget/pull/3566#discussion_r1802630951) an option.

### DIFF
--- a/docs/gadget-devel/output.md
+++ b/docs/gadget-devel/output.md
@@ -1,0 +1,33 @@
+---
+title: 'Displaying Output'
+sidebar_position: 620
+---
+
+By default, Inspektor Gadget allows you to display the output of each data
+source in all the output modes supported by the [CLI
+operator](../spec/operators/cli.md#output). Unless you have specific
+requirements, we recommend using the CLI operator's default configuration.
+However, for [Map Iterators](./gadget-intro.md#map-iterators) with a map value
+type of `gadget_histogram_slot__u32` or `gadget_histogram_slot__u64` (TODO: Add
+links), the output should be displayed as a chart, which the CLI operator
+doesn't support. In such cases, you can use the [Otel
+Metrics](../spec/operators/otel-metrics.md) operator to render the output as a
+histogram by annotating the data source with [metrics.print:
+"true"](../spec/operators/otel-metrics.md#metricsprint). The
+[profile_blockio](../gadgets/profile_blockio.mdx) gadget is an example of this.
+
+## Fields
+
+The fields to be displayed in the output when using the `columns`, `json`,
+`jsonpretty` or `yaml` output modes depend on the data source type:
+
+- Tracers: The fields are all the elements of the event `struct` specified when
+  defining the data source.
+- Map Iterators: The fields are all the elements of the key and value `struct`s
+  used to define the eBPF map to be iterated. As mentioned above, there is a
+  special case for the map iterators with a map value type of
+  `gadget_histogram_slot__u32` or `gadget_histogram_slot__u64`. In such cases,
+  the data source created by the Otel Metrics operator will contain just one
+  field called `text` which will carry the rendered histogram as a plain text.
+- Snapshotters: The fields are all the elements of the snapshot entry `struct`
+  specified when defining the data source.
--- a/docs/spec/operators/cli.md
+++ b/docs/spec/operators/cli.md
@@ -10,9 +10,11 @@ different output formats, depending on the gadget and the use case.
 
 10000
 
-## Instance Parameters
+## Parameters
 
-### `fields`
+### Instance Parameters
+
+#### `fields`
 
 Set the fields to be printed. Multiple data sources and fields can be specified
 using the format
@@ -20,16 +22,64 @@ using the format
 
 Fully qualified name: `operator.cli.fields`
 
-### `mode`
+#### `output`
 
-Determines the output format:
-- `json`: Outputs in JSON format.
-- `jsonpretty`: Outputs in pretty-printed JSON format.
-- `columns`: Outputs in a columnar format.
-- `yaml`: Outputs in YAML format.
-- `none`: No output.
-- `raw`: Outputs raw data.
+Determines the output format of the data source. It can be either a single output
+mode for all data sources or an output mode for each data source individually in
+the format `datasource:mode,datasource2:mode`.
 
-Fully qualified name: `operator.cli.mode`
+The following modes are supported:
+
+- `columns`: This is the default output mode. It displays the output in a
+  tabular format with a column for each field in the data source, unless the
+  field is annotated with `columns.hidden: true`. In addition to hiding columns,
+  you can also customize the columns format by using several annotations like
+  `columns.width`, `columns.alignment`, etc., as explained in TODO.
+- `json`: This mode displays the output in JSON format. Note that the output
+  contains all the fields of the data source, even if they annotated with
+  `columns.hidden: true` as this annotation applies only to the `columns` mode.
+  Depending on the data source type, it may be an array of objects or multiple
+  objects separated by newlines.
+- `jsonpretty`: As the `json` mode, but the output is formatted in a more
+  human-readable way.
+- `yaml`: This mode displays the output in YAML format. Like the `json` mode, it
+  contains all the fields of the data source. YAML entries will be separated by
+  `---` to make it easier to read.
+
+By default, the CLI operator allows setting the output of each data source in
+all the supported modes. However, this can be customized by annotating the data
+source with the [supported-output-modes](#clisupported-output-modes) annotation.
+
+Also, you can change the default output mode for each data source by setting the
+[default-output-mode](#clidefault-output-mode) annotation.
+
+Fully qualified name: `operator.cli.output`
 
 Default: `columns`
+
+## Annotations
+
+### Data Source Annotations
+
+#### `cli.supported-output-modes`
+
+Comma-separated list of output modes supported for the data source. This is
+useful when you want to customize the output modes supported by the data source.
+
+Note that if you include any output mode different from the ones supported by
+the CLI operator, it will treat as a custom output mode and will be printed as a
+string.
+
+#### `cli.default-output-mode`
+
+Set the default output mode for the data source. This is useful when you want to
+set a different default output mode for a data source.
+
+#### `cli.clear-screen-before`
+
+Clear the screen before printing the output of the data source. This is useful
+for data sources that emit periodically the batch of data and you want to clear
+the screen before printing the new data.
+
+This annotation is only applicable to the `columns` and custom output modes
+added by annotating the data source with `cli.supported-output-modes`.
--- a/docs/spec/operators/otel-metrics.md
+++ b/docs/spec/operators/otel-metrics.md
@@ -11,31 +11,99 @@ The otel-metrics operator handles collecting and exporting metrics using the
 
 9995
 
-## Global Parameters
+## Parameters
 
-### `otel-metrics-listen`
+### Global Parameters
+
+#### `otel-metrics-listen`
 
 Enables the Prometheus exporter on the address given by `otel-metrics-listen-address` if set to `true`.
 
 Default: `false`
 
-## Instance Parameters
+#### `otel-metrics-listen-address`
 
 The listen address that should be serving Prometheus requests.
 
 Default: `0.0.0.0:2224`
 
-## Instance Parameters
+### Instance Parameters
 
-### `otel-metrics-name`
+#### `otel-metrics-name`
 
 Overrides the name of a datasource and explicitly sets it as name for the export. This is mandatory if you
 want to export the metrics. Use a name that is unique to the gadget+params combination to avoid collision of metrics.
 
-### `otel-metrics-print-interval`
+Fully qualified name: `operator.otel-metrics.otel-metrics-name`
+
+#### `otel-metrics-print-interval`
 
 Interval in which metrics should be emitted as human-readable text. This only has effect for data sources that are
 annotated using `metrics.print=true`. This is also limited to print histograms for now. This functionality might be
 removed in the future.
+The minimum interval is 25ms.
+
+Fully qualified name: `operator.otel-metrics.otel-metrics-print-interval`
 
 Default: `1000ms`
+
+## Annotations
+
+### Data Source Annotations
+
+#### `metrics.collect`
+
+Together with the `otel-metrics-listen=true` and `otel-metrics-name=<name>`
+flags, this annotation is used to enable the Otel Metrics operator to export the
+data source's output as Prometheus metrics.
+
+#### `metrics.print`
+
+If set to `"true"`, the Otel Metrics operator will render the data source's
+output in more human-friendly formats.
+
+Currently, this feature only supports rendering as histograms the output of [Map
+Iterators](../../gadget-devel/gadget-intro.md#map-iterators) with the map’s
+value type of `gadget_histogram_slot__u32` or `gadget_histogram_slot__u64`
+(TODO: Add link). To achieve this, the Otel Metrics operator disables the
+original data source and creates a new one, suffixed with `-rendered`, which
+will emit the original data source's output as a rendered histogram.
+Additionally, the Otel Metrics operator will configure the CLI operator for this
+new data source as follows:
+
+- Set both the [cli.supported-output-modes](./cli.md#clisupported-output-modes)
+  and [cli.default-output-mode](./cli.md#clidefault-output-mode) annotations to
+  `histogram`. This will create a custom output mode that will make the CLI
+  operator print the output as it is received (i.e., a rendered histogram), and
+  use this mode as the default.
+- Set the [cli.clear-screen-before](./cli.md#cliclear-screen-before) annotation
+  to `true` to make the CLI operator clear the screen before printing each
+  histogram.
+
+### Field Annotations
+
+#### `metrics.type`
+
+Defines the type of the field. If not set, the operator will try to infer it
+from the field type.
+
+Possible values: `counter`, `gauge`, `histogram`, `key`.
+
+#### `metrics.unit`
+
+This annotation is used to set the [OpenTelemetry instrument
+unit](https://pkg.go.dev/go.opentelemetry.io/otel/metric@v1.30.0#WithUnit). It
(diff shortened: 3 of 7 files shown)


alban (commented):

| Benchmark suite | Current: 0e110eadce75639570734d31d06949a327c76a82 | Previous: 3471c2925dc8cca92f579775800d72b8fbb3ef47 | Ratio |
|-|-|-|-|
| `BenchmarkAllGadgetsWithContainers/container0/advise-seccomp` | `28961198` ns/op | `13213925` ns/op | `2.19` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-bind` | `199583751` ns/op | `99129311` ns/op | `2.01` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-mount` | `134193739` ns/op | `64942160` ns/op | `2.07` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-open` | `149597498` ns/op | `73282647` ns/op | `2.04` |

CC: @alban

pkg/operators/otel-metrics/otel-metrics.go
```diff
@@ -530,7 +529,37 @@ func (m *otelMetricsOperatorInstance) init(gadgetCtx operators.GadgetContext) er
...
+				for k, v := range ds.Annotations() {
+					ods.AddAnnotation(k, v)
+				}
+
+				if ods.Annotations()["cli.supported-output-modes"] == "" {
+					ods.AddAnnotation("cli.supported-output-modes", "raw")
+				}
+				if ods.Annotations()["cli.clear-screen-before"] == "" {
+					ods.AddAnnotation("cli.clear-screen-before", "true")
+				}
```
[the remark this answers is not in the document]
blanquicet: No, if annotations are not present, it'll be just an empty string. Annotations don't have default values as parameters.

pkg/operators/cli/clioperator.go
```diff
@@ -252,6 +281,12 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 			mode = annotatedMode
 		}
 
+		if !slices.Contains(o.supportedOutputModes[ds.Name()], mode) {
+			gadgetCtx.Logger().Warnf("output mode %q for data source %q is not supported; skipping data source",
+				mode, ds.Name())
```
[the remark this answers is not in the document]
blanquicet: Yeah, but stopping everything in these cases is something we still need to implement.

pkg/operators/cli/clioperator.go
```diff
@@ -276,9 +310,15 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 			fields, hasFields = fieldLookup[""] // fall back to default
 		}
 
-		mode := o.mode
-		if annotatedMode := ds.Annotations()["cli.output"]; annotatedMode != "" {
-			mode = annotatedMode
+		mode, ok := modes[ds.Name()]
+		if !ok {
+			mode, ok = modes[""]
```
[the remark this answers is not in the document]
blanquicet: It's because output can be now specified either as `columns` (Stored in the map with key `""`) or `datasource:json,datasource2:columns`. This is the same format we use for params like `map-fetch-interval` and `map-fetch-count` and helper `GetStringValuesPerDataSource` to parse it.

This support will have no impact on users because we don't have gadgets with more than one datasource. However, this code prepares things for that.

BTW, I improved the flag description to make it clearer: `Specifies output mode for all ("mode1") or per data source ("datasource:mode1,datasource2:mode2")`

pkg/operators/otel-metrics/otel-metrics.go
```diff
@@ -513,7 +513,6 @@ func (m *otelMetricsOperatorInstance) shutdown() {
 }
 
 func (m *otelMetricsOperatorInstance) init(gadgetCtx operators.GadgetContext) error {
```
mauriciovasquezbernal, line 515: I know the goal of this PR is different, but what about this alternative implementation of this function: Is it equivalent?

```go
func (m *otelMetricsOperatorInstance) init(gadgetCtx operators.GadgetContext) error {
	for _, ds := range gadgetCtx.GetDataSources() {
		annotations := ds.Annotations()
		metricsCollect := annotations[AnnotationMetricsCollect] == "true"
		// We only allow printing if the gadget is _not_ running remotely
		metricsPrint := annotations[AnnotationMetricsPrint] == "true" && !gadgetCtx.IsRemoteCall()

		// Not collecting AND not printing, do nothign for this data source
		if !metricsCollect && !metricsPrint {
			continue
		}

		// we only use the global instance if there's an explicit name mapping available; otherwise we'll fallback
		// to using a dedicated local registry/collector instance
		useGlobal := false
		if metricsPrint {
			gadgetCtx.Logger().Debugf("enabling print for %s", ds.Name())

			// Disable original data source to avoid other operators subscribing to it
			ds.SetRequested(false)

			// Create a new data source for the output with a single field
			odsName := fmt.Sprintf("%s-%s", ds.Name(), PrintDataSourceSuffix)
			ods, err := gadgetCtx.RegisterDataSource(datasource.TypeSingle, odsName)
			if err != nil {
				return fmt.Errorf("registering %q: %w", odsName, err)
			}

			// set default annotations
			ods.AddAnnotation("cli.supported-output-modes", "raw")
			ods.AddAnnotation("cli.default-output-mode", "raw")
			ods.AddAnnotation("cli.clear-screen-before", "true")

			// Use annotations from the original data source
			for k, v := range ds.Annotations() {
				ods.AddAnnotation(k, v)
			}

			f, err := ods.AddField(PrintFieldName, api.Kind_String, datasource.WithAnnotations(map[string]string{"content-type": "text/plain"}))
			if err != nil {
				return fmt.Errorf("adding field %q: %w", PrintFieldName, err)
			}

			// TODO: Store these fields per datasource to support multiple datasources
			m.outputDS = ods
			m.outputField = f
		}

		// if there's an explicit mapping set for the datasource, we will export it using the global exporter
		// (if that is available)
		mappedName, ok := m.nameMappings[ds.Name()]
		if ok && m.op.exporter != nil {
			useGlobal = true
		} else if ok {
			gadgetCtx.Logger().Warnf("global exporter not configured, using local metric instance")
		}

		// If mapped name is empty, it hasn't been explicitly set (or set to empty), so we will use the data source name
		if mappedName == "" {
			mappedName = ds.Name()
		}

		gadgetCtx.Logger().Debugf("collecting metrics for data source %q as %q", ds.Name(), mappedName)

		m.collectors[ds] = &metricsCollector{
			output:            metricsPrint,
			mappedName:        mappedName,
			useGlobalProvider: useGlobal,
		}
	}
	return nil
}
```

docs/gadget-devel/output.md
```diff
@@ -0,0 +1,40 @@
...
+
+The fields to be displayed in the output when using the `columns`, `json`,
+`jsonpretty` or `yaml` output modes depend on the data source type:
+
+- Tracers: The fields are all the elements of the event `struct` specified when
+  defining the data source.
+- Map Iterators: The fields are all the elements of the key and value `struct`s
+  used to define the eBPF map to be iterated. As mentioned above, there is a
+  special case for the map iterators with a map value type of
+  `gadget_histogram_slot__u32` or `gadget_histogram_slot__u64`. In such cases,
```
mauriciovasquezbernal, line 36: I think histograms are a special case because those are the only ones we support printing to the terminal, but I don't see any reason why not to support printing counters and gauges too. i.e. all cases where are metrics involved will be "special".

docs/spec/operators/cli.md
```diff
@@ -10,26 +10,76 @@ different output formats, depending on the gadget and the use case.
...
+for data sources that emit periodically the bath of data and you want to clear
+the screen before printing the new data.
+
+This annotation is only applicable to the `columns` and `raw` output modes.
+
+#### `cli.supported-output-modes`
+
+Comma-separated list of output modes supported for the data source. This is
+useful when you want to restrict the output modes for a data source.
+
```
mauriciovasquezbernal, line 81: I have mixed feelings. I think it's useful internally (an operator could register a new output mode), but I don't see why the gadget developer should worry about it.

mauriciovasquezbernal (commented):
Most of it makes sense to me. By using this PR I think we can move away from the "raw" output mode that was too generic:

Regarding to this topic, I find the "raw" term too generic, what about if we use this mechanism to allow operators registering custom output modes, in this case, they're restricted to having a single string field called text for now:

```diff
diff --git a/pkg/operators/cli/clioperator.go b/pkg/operators/cli/clioperator.go
index 8e151bbfb..4a951379d 100644
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -45,7 +45,6 @@ const (
        ModeColumns    = "columns"
        ModeYAML       = "yaml"
        ModeNone       = "none"
-       ModeRaw        = "raw"

        // AnnotationClearScreenBefore can be used to clear the screen before printing a new event; usually used for
        // array events
@@ -330,7 +329,7 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
                }

                switch mode {
-               case ModeRaw:
+               default:
                        before := func() {}
                        if ds.Annotations()[AnnotationClearScreenBefore] == "true" && term.IsTerminal(int(os.Stdout.Fd())) {
                                before = clearScreen
diff --git a/pkg/operators/otel-metrics/otel-metrics.go b/pkg/operators/otel-metrics/otel-metrics.go
index 25a03cc0e..5803cd83c 100644
--- a/pkg/operators/otel-metrics/otel-metrics.go
+++ b/pkg/operators/otel-metrics/otel-metrics.go
@@ -546,10 +546,10 @@ func (m *otelMetricsOperatorInstance) init(gadgetCtx operators.GadgetContext) er
                                }

                                if ods.Annotations()["cli.supported-output-modes"] == "" {
-                                       ods.AddAnnotation("cli.supported-output-modes", "raw")
+                                       ods.AddAnnotation("cli.supported-output-modes", "histogram")
                                }
                                if ods.Annotations()["cli.default-output-mode"] == "" {
-                                       ods.AddAnnotation("cli.default-output-mode", "raw")
+                                       ods.AddAnnotation("cli.default-output-mode", "histogram")
                                }
                                if ods.Annotations()["cli.clear-screen-before"] == "" {
                                        ods.AddAnnotation("cli.clear-screen-before", "true")
```

This could be useful also for #3003 when gadgets should be able to define new output modes from wasm.

I'll give it another pass next week.

pkg/operators/cli/clioperator.go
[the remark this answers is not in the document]
blanquicet: It's a map, so `""` is the key and `"columns"` is the value. When multiple datasource are specified, `"<datasource>"` is the key and `"columns"`, or `"json"` or whatever is the value. I'm sorry but don't get your point.

pkg/operators/otel-metrics/otel-metrics.go
blanquicet: It's and it's clearer. So, using it.

docs/gadget-devel/output.md
blanquicet: Mmmm but counters and gauges can simply be printed as columns, JSON, or any other output mode supported by the CLI operator, they don't require a special treatment.

docs/spec/operators/cli.md
blanquicet: Makes sense. I removed the references to these annotations in the gadget-devel documentation.

docs/spec/operators/otel-metrics.md
```diff
@@ -11,31 +11,98 @@ The otel-metrics operator handles collecting and exporting metrics using the
...
 
+Fully qualified name: `operator.otel-metrics.otel-metrics-print-interval`
+
 Default: `1000ms`
+
+## Annotations
+
+### Data Source Annotations
+
+#### `metrics.collect`
```
blanquicet, line 53: IMO, this should be called `metrics.export` as the Otel Metrics operator, if enabled, will always collect, but it will only export if this annotation is set.

docs/spec/operators/cli.md
```diff
@@ -10,26 +10,76 @@ different output formats, depending on the gadget and the use case.
...
 `datasource:comma,separated,fields;datasource2:comma,separated,fields`.
 
 Fully qualified name: `operator.cli.fields`
 
-### `mode`
+#### `output`
+
+Determines the output format of the data source. It can be either a single output
+mode for all data sources or an output mode for each data source individually in
+the format `datasource:mode,datasource2:mode`.
```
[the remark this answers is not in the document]
blanquicet: Do you mean have a flag for each datasource? Or what do you mean?

The current implementation is same format we use for params like `map-fetch-interval` and `map-fetch-count` and helper `GetStringValuesPerDataSource` to parse it:

- `--output columns`: Sets one mode for all data sources
- `--output datasource:columns,datasource2:json: Sets one output for each data source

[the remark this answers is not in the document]
blanquicet: In the YAML, you mean in the [manifest](https://github.com/inspektor-gadget/inspektor-gadget/pull/3515), right?

Yeah, we can think about supporting something like:

```
version: 1
kind: instance-spec
imageName: trace_exec
name: myTraceExec
paramValues:
  operator.cli.output:
    datasource0: columns
    datasource1: json
  operator.oci.wasm.gadget_a: 1212
  operator.oci.ebpf.gadget_a: 1212
```

blanquicet:
> Regarding to this topic, I find the "raw" term too generic, what about if we use this mechanism to allow operators registering custom output modes
> 
It makes sense to me. I applied your suggestion.

> in this case, they're restricted to having a single string field called text for now:
>
Not really. They can have multiple fields, but all of them will be printed as a string (whatever comes from doing [string(fieldData)](https://github.com/inspektor-gadget/inspektor-gadget/blob/jose/render-datasource/pkg/datasource/accessors.go#L577-L579))

docs/spec/operators/cli.md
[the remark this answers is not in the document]
blanquicet: Added everything as "Future TODOs" on PR's description.

docs/gadget-devel/output.md
mauriciovasquezbernal: Makes sense, we can print other kind of metrics as columns.

docs/spec/operators/otel-metrics.md
```diff
@@ -11,31 +11,99 @@ The otel-metrics operator handles collecting and exporting metrics using the
...
+#### `metrics.print`
+
+If set to `"true"`, the Otel Metrics operator will render the data source's
+output in more human-friendly formats.
+
+Currently, this feature only supports rendering as histograms the output of [Map
+Iterators](../../gadget-devel/gadget-intro.md#map-iterators) with the map’s
+value type of `gadget_histogram_slot__u32` or `gadget_histogram_slot__u64`
+(TODO: Add link). To achieve this, the Otel Metrics operator disables the
+original data source and create a new one, suffixed with `-rendered`, which will
```
mauriciovasquezbernal, line 69:
```suggestion
original data source and creates a new one, suffixed with `-rendered`, which will
```

mauriciovasquezbernal (approved):
LGTM. IMO it's a good clean up and it's ready to go!

PR #3568, opened 2024-10-10
new_gadget: chaos_packet_drop : Gadget to randomly drop TCP/UDP packets

MeherRushi:
# Adding a new gadget to simulate TCP/UDP packet drops in the network

- This gadget is used to simulate the random packet drops in the network based on independent (Bernoulli) probability model. The program runs at the tc hook point and can be configured to run at ingress/egress and for TCP/UDP packets and the drop probability percentage can also be configured. We can also drop packets of a specific IP/port number.

## Features 
- Works at both ingress and egress based on the input parameters
- Works for both TCP/UDP packets based on the input parameters
- Randomly drops packets in the network based on independent (Bernoulli) probability model which can be configured using the loss_percentage parameter
- Can filter the packet drops per specific IP and port number

## Issues faced currently.

There are 2 main issues:
- drop_cnt
- comm 

#### drop_cnt

The drop count value that is being displayed in the map seems to be a little inaccurate. I am not really sure if the  logic of the program is the issue or is it because of GADGET_MAPITER.

The main issue being, the drop_cnt value for a particular IP address,port pair that is stored a key in the table should always be incremented by 1. But I see a lot of repeating constant 1 values. I suspect that the map is being cleared from time to time, but I am not really sure. Any help with regards to this would be appreciated.

Eg:
```RUNTIME.CONTAINERNAME        DST                                   SRC                                   FILTER_IPV4                          DROP_CNT   INGR… EGRE… COMM                  PID
test-pkt-drop                80.249.99.148:50318                   172.17.0.2:50318                      0.0.0.0:0                            1          false true  wget              2653142
test-pkt-drop                80.249.99.148:50318                   172.17.0.2:50318                      0.0.0.0:0                            2          false true  ����           4294942639
test-pkt-drop                80.249.99.148:50318                   172.17.0.2:50318                      0.0.0.0:0                            75         false true  wget              2653142
test-pkt-drop                80.249.99.148:50318                   172.17.0.2:50318                      0.0.0.0:0                            1          false true  wget              2653142
```

#### comm field

I have enriched the comm field using the socket enricher but sometimes the value seems to be unreadable strings.

Eg:
```RUNTIME.CONTAINERNAME        DST                                   SRC                                   FILTER_IPV4                          DROP_CNT   INGR… EGRE… COMM                  PID
test-pkt-drop                172.17.0.2:56510                      80.249.99.148:80                      0.0.0.0:0                            1          true  false ������v�����^… 4294938030
test-pkt-drop                172.17.0.2:56510                      80.249.99.148:80                      0.0.0.0:0                            11         true  false ����#                   0
test-pkt-drop                172.17.0.2:56510                      80.249.99.148:80                      0.0.0.0:0                            10         true  false                4294967295
test-pkt-drop                172.17.0.2:56510                      80.249.99.148:80                      0.0.0.0:0                            6          true  false ����           4294942639
test-pkt-drop                172.17.0.2:56510                      80.249.99.148:80                      0.0.0.0:0                            43         true  false wget              2612075
```

## TODO :
- Currently the IP parameter is taken in octets, need to change that to be more user friendly. This can be also used by many other gadgets - it is also a TODO in `inspektor-gadget/gadgets/profile_tcprtt/program.bpf.c`
- This would also extend the filtering to ipv6 address
- Testing
- Fixing drop_cnt and comm fields

## How to use

Described in the Readme.mdx file

## Testing done

Tested Locally. Test Cases are to be written.


flyth (commented):
Thanks for the contribution, good job!

This looks much more performant now with the early returns. I think we can still improve readability a bit by using switch statements instead of if/elses.

Could you also please lint the `.c` code and streamline it a bit, especially:

* indentation
* comments
* line-lengths
* white-spaces

I will run some tests later on.

gadgets/chaos_packet_drop/README.mdx
```diff
@@ -0,0 +1,215 @@
...
+WARN[0001] image signature verification is disabled due to using corresponding option 
+WARN[0001] image signature verification is disabled due to using corresponding option 
+RUNTIME.CONTAINERNAME        DST                                   SRC                                   FILTER_IPV4                          DROP_CNT   INGR… EGRE… COMM                  PID
+RUNTIME.CONTAINERNAME        DST                                   SRC                                   FILTER_IPV4                          DROP_CNT   INGR… EGRE… COMM                  PID
+test-pkt-drop                80.249.99.148:53000                   172.17.0.2:53000                      0.0.0.0:0                            6          false true  wget              2645279
+RUNTIME.CONTAINERNAME        DST                                   SRC                                   FILTER_IPV4                          DROP_CNT   INGR… EGRE… COMM                  PID
+test-pkt-drop                80.249.99.148:53000                   172.17.0.2:53000                      0.0.0.0:0                            1          false true  ����`�        1955046299
```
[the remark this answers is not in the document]
MeherRushi: Yup, It has been resolved now, after initializing it to 0

gadgets/chaos_packet_drop/gadget.yaml
```diff
@@ -0,0 +1,115 @@
...
+documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/chaos_packet_drop
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/chaos_packet_drop
+datasources:
+  events_map:
+    annotations:
+      events_map.collect: "true"
+    fields: {}
+  events_map_iter:
+    fields:
+      comm:
```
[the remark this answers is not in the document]
MeherRushi: Hey, this seem to be a great idea. But I realized that we need to add the same types to the test support as well right , for now I have declared the structs of Proc in the test file but I would like to shift those structs to types.go of the utils package #3600 ?

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+			    port != bpf_ntohs(udph->dest))
+				return TC_ACT_OK;
+
+			event.src.port = bpf_ntohs(udph->source);
+			key.dst.port = skb->pkt_type == SE_PACKET_HOST ?
+					       bpf_ntohs(udph->dest) :
+					       bpf_ntohs(udph->source);
+			event.src.proto_raw = key.dst.proto_raw = IPPROTO_UDP;
+			sockets_key_for_md.proto = IPPROTO_UDP;
+			break;
```
mauriciovasquezbernal, line 344: This logic is repeated on the IPv4 case. Do you think it's possible to move it to a common place?

One idea will be to move it below this, something like:
- pointer to L4 = null
- parse L3 proto (Ipv4 and IPv6). Update pointer to L4 to right position within the packet 
- parse L4 proto: based on the L4 pointer defined above.

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+	}
+	struct event *event_map_val = bpf_map_lookup_elem(
+		&events_map,
+		key); /* The events which are stored in the events_map */
+
+	if (!event)
+		return TC_ACT_OK;
+
+	if (rand_num <=
+	    (u32)threshold) // Run the code only if the random number is less than the threshold
```
mauriciovasquezbernal, line 121: Use an early return here. 

```
if (rand_num > ...)
   return TC_ACT_OK
```

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+
+	if (ingress == true) {
+		swap_src_dst(event, key);
+	}
+	struct event *event_map_val = bpf_map_lookup_elem(
+		&events_map,
+		key); /* The events which are stored in the events_map */
+
+	if (!event)
+		return TC_ACT_OK;
```
mauriciovasquezbernal, line 118: Is it possible?

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+				 (__u64)0xFFFFFFFF) /
+		100; // loss_percentage% of UINT32_MAX
+
+	if (ingress == true) {
+		swap_src_dst(event, key);
+	}
+	struct event *event_map_val = bpf_map_lookup_elem(
+		&events_map,
+		key); /* The events which are stored in the events_map */
+
```
mauriciovasquezbernal, line 116: this should be executed later on: only when we're going to drop the packet, otherwise this lookup is done for nothing.

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+/* The maximum number of items in maps */
+#define MAX_ENTRIES 8192
+#define TASK_COMM_LEN 16
+
+struct events_map_key {
+	struct gadget_l4endpoint_t dst;
+};
+
+struct event {
+	gadget_timestamp timestamp_raw;
```
mauriciovasquezbernal, line 47: What's the purpose of having a timestamp here? This gadget provides a "summary" of how many events have been dropped, I can't see how a timestamp is useful.

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+			event->timestamp_raw = bpf_ktime_get_boot_ns();
+			bpf_map_update_elem(&events_map, key, event,
+					    BPF_NOEXIST);
+		} else {
+			// Increment the the value of drop count by 1.
+			// We use sync fetch and add which is an atomic addition operation
+			__sync_fetch_and_add(&event_map_val->drop_cnt, 1);
+			event_map_val->timestamp_raw = bpf_ktime_get_boot_ns();
+			bpf_map_update_elem(&events_map, key, event_map_val,
+					    BPF_EXIST);
```
mauriciovasquezbernal, line 157: I don't think it's necessary. `event_map_val` is already a pointer to the value in the map, so the operation above will update the value saved in the map.

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+						 skb_val->task,
+						 sizeof(event->proc.comm));
+				__builtin_memcpy(
+					&event->proc.parent.comm,
+					skb_val->ptask,
+					sizeof(event->proc.parent.comm));
+				event->proc.creds.uid = (__u32)skb_val->uid_gid;
+				event->proc.creds.gid =
+					(__u32)(skb_val->uid_gid >> 32);
+			}
```
mauriciovasquezbernal, line 147: The `gadget_process_populate_from_socket`  could replace all of it. Please check out it's used in the trace_dns gadget.

gadgets/chaos_packet_drop/program.bpf.c
mauriciovasquezbernal: Also, what about moving this much early on the whole processing so we don't expend a lot of time checking other things if in the end we're not going to drop the packet? I think it could be almost the first thing done the in the gadget.

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+} events_map SEC(".maps");
+
+GADGET_MAPITER(events_map_iter, events_map);
+
+// we use the following variables as parameters
+const volatile struct gadget_l3endpoint_t filter_ip = { 0 };
+const volatile __u16 port = 0;
+const volatile __u32 loss_percentage = 100;
+const volatile bool filter_tcp =
+	true; /* This is a boolean flag to enable filtering of TCP packets */
```
mauriciovasquezbernal, line 72: nit: What about something like 

```
/* This is a boolean flag to enable filtering of UDP packets */
const volatile bool filter_udp = true;
```

to avoid this weird formatted caused by the linter?

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+				      ip6h->saddr.in6_u.u6_addr8, 16)))
+			return TC_ACT_OK;
+		if (egress && filter_ip.version == 6 &&
+		    (__builtin_memcmp((const void *)filter_ip.addr_raw.v6,
+				      ip6h->daddr.in6_u.u6_addr8, 16)))
+			return TC_ACT_OK;
+
+		event.src.version = key.dst.version = 6;
+		sockets_key_for_md.family = SE_AF_INET6;
+		if (ip6h->nexthdr != NULL)
```
mauriciovasquezbernal, line 298: nexthdr is a u8, why is it compared against null?

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+		case IPPROTO_TCP:
+			if (!filter_tcp)
+				return TC_ACT_OK;
+			struct tcphdr *tcph = (struct tcphdr *)(ip6h + 1);
+			if ((void *)(tcph + 1) > data_end)
+				return TC_ACT_OK;
+
+			if (ingress && port != 0 &&
+			    port != bpf_ntohs(tcph->source))
+				return TC_ACT_OK;
```
mauriciovasquezbernal, line 311: nit: I'm not 100% sure, but perhaps it's better to apply the `bpf_htons` operation on the constant to avoid this bering performed for all packets. (Same in other places)

gadgets/chaos_packet_drop/program.bpf.c
```diff
@@ -0,0 +1,385 @@
...
+
+#define GADGET_TYPE_NETWORKING
+#include <gadget/sockets-map.h>
+
+/* The maximum number of items in maps */
+#define MAX_ENTRIES 8192
+#define TASK_COMM_LEN 16
+
+struct events_map_key {
+	struct gadget_l4endpoint_t dst;
```
mauriciovasquezbernal, line 43: AFAIU it's dst or src according to the value of ingress and egress, right? if yes, can we name this something else than dst?

mauriciovasquezbernal (commented):
Some initial comments on the structure. I'll make another pass once these ones are handled.

MeherRushi:
Hey, yes, will resolve the pointers soon

PR #3574, opened 2024-10-11
docs: Remove increasing sidebar_position from docs/gadgets

burak-ok:
Docasaurus sorts the sidebar items alphabetically by default if the position is the same. Therefore we can just set the same position for image based gadgets (In this PR `0`) and the folder for builtin gadgets to a higher value (In this PR `1`). This ordering of the builtin gadget folder is the same as before, since they are not the future.

This PR then also removes the `make readmes` target and the corresponding github workflow check


mauriciovasquezbernal (approved):
Great, I considered so many complicated solutions but this one is simpler and works perfectly! Thanks!

PR #3575, opened 2024-10-12
feat: Accepting IPv4 and IPv6 parameters as command line input arguments

MeherRushi:
- Made changes to params.go to label all parameters with prefix `ipaddr` as type TypeString
- Made changes to ebpf.go operator to convert all parameters with prefix `ipaddr_v4` from string to uint32 and
 all parameters with prefix `ipaddr_v6` from string to struct uint128
- Defined uint128 as a struct with 2 members of uint64 and wrote a helper function `IPStringToUint128()` in helpers.go

# Accepting IPv4 and IPv6 arguments as a input parameter

This commit allows us to now specify IPv4 addresses as an input parameter in the cli. 

For example:
Let's assume we wanted to have a parameter of ipv4 and ipv6 address. We declare it as a u32 int and u128 in the kernel space as shown below:
```C
volatile const __u32 ipaddr_v4 = 3232235777;
volatile const u128 ipaddr_v6 =  {0x123456789ABCDEF0, 0x0FEDCBA987654321}; 

GADGET_PARAM(ipaddr_v4);
GADGET_PARAM(ipaddr_v6);
```
Then after building the gadget we run it using the following

```bash
sudo ig run bit_flip:exp --verify-image=false --ipaddr_v4="127.0.0.1"  --ipaddr_v6="2001:db8::1" -v
```

Here the ipaddr_v4 is be converted into a 32 bit integer using the pre-defined helper function `gadgets.IPStringToUint32()`
And the ipaddr_v6 is converted into uint128 using a custom defined helper function `gadgets.IPStringToU128()`

There values are logged when we use `-v` flag:
```bash
DEBU[0007] setting param value "ipaddr_v4" = 16777343   
DEBU[0007] setting param value "ipaddr_v6" = {2306139568115548160 1} 
```

## How to use

The parameters to be accepted as IPv4 must begin with `prefix ipaddr_v4`
So it can be `ipaddr_v4_1` or `ipaddr_v4_anything` but should have prefix ipaddr_v4

Similar for IPv6 parameters, they need to have prefix `ipaddr_v6`

## Testing done

Locally tested it.

### DIFF
--- a/pkg/operators/ebpf/ebpf.go
+++ b/pkg/operators/ebpf/ebpf.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"reflect"
 	"slices"
@@ -45,6 +46,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/networktracer"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/socketenricher"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/tchandler"
@@ -581,7 +583,25 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 			continue
 		}
 		constReplacements[name] = paramMap[name].AsAny()
-		i.logger.Debugf("setting param value %q = %v", name, paramMap[name].AsAny())
+
+		switch p.TypeHint {
+		case api.TypeIP:
+			i.logger.Debugf("handling IP param: %q", name)
+			ipAddr := ebpftypes.L3Endpoint{}
+			ipParam := constReplacements[name].(net.IP)
+			if ip := ipParam.To4(); ip != nil {
+				ipAddr.Version = 4
+				copy(ipAddr.V6[:4], ip)
+			} else if ip := ipParam.To16(); ip != nil {
+				copy(ipAddr.V6[:], ip)
+				ipAddr.Version = 6
+			} else {
+				return fmt.Errorf("invalid IP address: %v", ipParam)
+			}
+			constReplacements[name] = ipAddr
+		}
+
+		i.logger.Debugf("setting param value %q = %v", name, constReplacements[name])
 	}
 
 	for _, v := range i.vars {
--- a/pkg/operators/ebpf/params.go
+++ b/pkg/operators/ebpf/params.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/btfhelpers"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 )
 
@@ -70,6 +71,11 @@ func getTypeHint(typ btf.Type) params.TypeHint {
 		return getTypeHint(typ)
 	case *btf.Volatile:
 		return getTypeHint(typedMember.Type)
+	case *btf.Struct:
+		switch typedMember.Name {
+		case ebpftypes.L3EndpointTypeName:
+			return params.TypeIP
+		}
 	}
 
 	return params.TypeUnknown
--- a/pkg/operators/ebpf/params_defaults.go
+++ b/pkg/operators/ebpf/params_defaults.go
@@ -24,6 +24,7 @@ import (
 	"github.com/cilium/ebpf/btf"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/btfhelpers"
+	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
 )
 
 var errMapNoBTFValue = errors.New("map spec does not contain a BTF Value")
@@ -111,6 +112,11 @@ func (i *ebpfInstance) fillParamDefaults() error {
 					} else {
 						defaultValue = "true"
 					}
+				}
+			case *btf.Struct:
+				switch t.Name {
+				case ebpftypes.L3EndpointTypeName:
+					defaultValue = "0.0.0.0"
 				}
 			}
 
--- a/pkg/operators/ebpf/types/types.go
+++ b/pkg/operators/ebpf/types/types.go
@@ -12,9 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package types contains the names and Golang representation of the types used
+// by the eBPF programs. Keep this aligned with include/gadget/types.h
 package types
 
-// Keep this aligned with include/gadget/types.h
 const (
 	IPAddrTypeName      = "gadget_ip_addr_t"
 	L3EndpointTypeName  = "gadget_l3endpoint_t"
@@ -42,3 +43,10 @@ const (
 	HistogramSlotU32TypeName = "gadget_histogram_slot__u32"
 	HistogramSlotU64TypeName = "gadget_histogram_slot__u64"
 )
+
+// L3Endpoint is the Golang representation of struct gadget_l3endpoint_t
+type L3Endpoint struct {
+	V6      [16]byte
+	Version uint8
+	_       [3]byte
+}


mauriciovasquezbernal (commented):
I think we already have almost all infrastructure needed for this change:

1. We shouldn't rely on the param name to check it's type, we can use BTF as done in https://github.com/inspektor-gadget/inspektor-gadget/blob/b3201a152510b715f30fc9080f5e1fbb4dd60efd/pkg/operators/ebpf/params.go#L95
2. We already have logic to parse IP params

https://github.com/inspektor-gadget/inspektor-gadget/blob/3e16bf5cd7829a6960dd1f68f0a7d2599a32ea3a/pkg/params/params.go#L543-L545

https://github.com/inspektor-gadget/inspektor-gadget/blob/3e16bf5cd7829a6960dd1f68f0a7d2599a32ea3a/pkg/params/validators.go#L45

3. I think we can reuse existing types representing an IP address

https://github.com/inspektor-gadget/inspektor-gadget/blob/d5e0f1198bd74c9ef66dea1c53a84ea94b40cecc/include/gadget/types.h#L15-L18


I drafted something using those things above:

```diff 
diff --git a/pkg/operators/ebpf/ebpf.go b/pkg/operators/ebpf/ebpf.go
index c48a78072..9ac425b0f 100644
--- a/pkg/operators/ebpf/ebpf.go
+++ b/pkg/operators/ebpf/ebpf.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"reflect"
 	"slices"
@@ -543,6 +544,13 @@ func (i *ebpfInstance) tracePipe(gadgetCtx operators.GadgetContext) error {
 	return nil
 }
 
+// Golang representation of struct gadget_l3endpoint_t in include/gadget/types.h
+type ipAddr struct {
+	v6      [16]byte
+	version uint8
+	pad     [3]byte
+}
+
 func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 	i.logger.Debugf("starting ebpfInstance")
 
@@ -580,7 +588,30 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 		if !p.fromEbpf {
 			continue
 		}
-		constReplacements[name] = paramMap[name].AsAny()
+
+		paramValue := paramMap[name].AsAny()
+
+		switch p.TypeHint {
+		case api.TypeIP:
+			i.logger.Infof("handling IP param: %q", name)
+			ipAddr := ipAddr{}
+			ipParam := paramValue.(net.IP)
+			if ip := ipParam.To4(); ip != nil {
+				ipAddr.version = 4
+				copy(ipAddr.v6[:4], ip)
+			} else if ip := ipParam.To16(); ip != nil {
+				copy(ipAddr.v6[:], ip)
+				ipAddr.version = 6
+			} else {
+				// TODO: is it possible?
+				return fmt.Errorf("invalid IP address: %v", ipParam)
+			}
+			paramValue = ipAddr
+
+			fmt.Printf("IP: %+v\n", ipAddr)
+		}
+
+		constReplacements[name] = paramValue
 		i.logger.Debugf("setting param value %q = %v", name, paramMap[name].AsAny())
 	}
 
diff --git a/pkg/operators/ebpf/params.go b/pkg/operators/ebpf/params.go
index b02ab1d89..976f10fd9 100644
--- a/pkg/operators/ebpf/params.go
+++ b/pkg/operators/ebpf/params.go
@@ -70,6 +70,12 @@ func getTypeHint(typ btf.Type) params.TypeHint {
 		return getTypeHint(typ)
 	case *btf.Volatile:
 		return getTypeHint(typedMember.Type)
+	case *btf.Struct:
+		switch typedMember.Name {
+		// TODO: define contant for it
+		case "gadget_l3endpoint_t":
+			return params.TypeIP
+		}
 	}
 
 	return params.TypeUnknown
```

And tested it with this gadget:

```c 
#include <vmlinux.h>
#include <bpf/bpf_helpers.h>
#include <gadget/buffer.h>
#include <gadget/macros.h>
#include <gadget/types.h>

//volatile const int p1 = 4;
//GADGET_PARAM(p1);

volatile const struct gadget_l3endpoint_t myparam;
GADGET_PARAM(myparam);

struct event {
	__u32 pid;
};

GADGET_TRACER_MAP(open_events, 1024 * 256);
GADGET_TRACER(open, open_events, event);

int enter_openat(struct trace_event_raw_sys_enter *ctx)
{
	if (myparam.version == 4) {
		bpf_printk("my param is a v4 address: %pi4", &myparam.addr_raw.v4);
	} else if (myparam.version == 6) {
		bpf_printk("my param is a v6 address: %pi6", myparam.addr_raw.v6);
	}

	// copy and pasted from old gadget
	struct event *event;
	event = gadget_reserve_buf(&open_events, sizeof(*event));
	if (!event)
		return 0;

	gadget_submit_buf(ctx, &open_events, event, sizeof(*event));

	return 0;
}

```

(more information about printf on https://docs.ebpf.io/linux/helper-function/bpf_trace_printk/)

I think that can be the base for this support. 

PD: I didn't take considerations about endianess at all, please check everything is working fine in that regard.

mauriciovasquezbernal:
If we really need, we can extend [gadget_ip_addr_t](https://github.com/inspektor-gadget/inspektor-gadget/blob/d5e0f1198bd74c9ef66dea1c53a84ea94b40cecc/include/gadget/types.h#L9-L12) with alternative definitions for the addresses, like u64[2], u32[4], etc. for the v6 one and u8[4] for the v4 one.

MeherRushi:
Thank you so much for the help provided with the approach.

The suggested method of taking IP parameters does seem better. I have made the changes as suggested .
I have also added the default value of the parameter to be "0.0.0.0" in case the parameter is not specified else by default if would be set to `nil` resulting in the invalid IP error.

I did test this with basic filtering. The endianess is by default the big endian (network endian)I believe as we don't make any casting anywhere and it works while filtering packets. So we can directly read the address without changing it from network endian to host endian etc...

pkg/operators/ebpf/ebpf.go
```diff
@@ -580,7 +588,29 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 		if !p.fromEbpf {
 			continue
 		}
-		constReplacements[name] = paramMap[name].AsAny()
+		paramValue := paramMap[name].AsAny()
+
+		switch p.TypeHint {
+		case api.TypeIP:
+			i.logger.Infof("handling IP param: %q", name)
```
mauriciovasquezbernal, line 595: Debugd?

pkg/operators/ebpf/ebpf.go
```diff
@@ -580,7 +588,29 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+			} else if ip := ipParam.To16(); ip != nil {
+				copy(ipAddr.v6[:], ip)
+				ipAddr.version = 6
+			} else {
+				// TODO: is it possible?
+				return fmt.Errorf("invalid IP address: %v", ipParam)
+			}
+			paramValue = ipAddr
+
+			fmt.Printf("IP: %+v\n", ipAddr)
```
mauriciovasquezbernal, line 610: remove this

pkg/operators/ebpf/ebpf.go
```diff
@@ -580,7 +588,29 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+			ipAddr := ipAddr{}
+			ipParam := paramValue.(net.IP)
+			if ip := ipParam.To4(); ip != nil {
+				ipAddr.version = 4
+				copy(ipAddr.v6[:4], ip)
+			} else if ip := ipParam.To16(); ip != nil {
+				copy(ipAddr.v6[:], ip)
+				ipAddr.version = 6
+			} else {
+				// TODO: is it possible?
```
mauriciovasquezbernal, line 605: please handle this TODO

pkg/operators/ebpf/ebpf.go
```diff
@@ -543,6 +544,13 @@ func (i *ebpfInstance) tracePipe(gadgetCtx operators.GadgetContext) error {
 	return nil
 }
 
+// Golang representation of struct gadget_l3endpoint_t in include/gadget/types.h
+type ipAddr struct {
+	v6      [16]byte
+	version uint8
+	pad     [3]byte
```
mauriciovasquezbernal, line 551: I think it's possible to use `_ [3]byte`

pkg/operators/ebpf/types.go
```diff
@@ -33,4 +33,7 @@ const (
 
 	// Prefix used to mark variables used by operators
 	varPrefix = "gadget_var_"
+
+	// Constant defining gadget_l3endpoint_t
```
mauriciovasquezbernal, line 37: Do we need this comment?

mauriciovasquezbernal (commented):
Some minor comments but looks good.

Since I did the suggestion, I'd like somebody else to review it before merging.

pkg/operators/ebpf/types.go
MeherRushi: We can remove it, 
I just thought since every other prefix had a comment, I'll add one here.
or maybe I can make it more descriptive , something like :
`Constant defining the name of a struct type for filtering IP addresses`

pkg/operators/ebpf/ebpf.go
MeherRushi: ok, got it

pkg/operators/ebpf/ebpf.go
MeherRushi: Yes, will change it

pkg/operators/ebpf/ebpf.go
MeherRushi: Yup, it is possible, this case will run when `ipParam.To4()` or `ipParam.To6()` throws an error. So I believe we just have to remove the comment

MeherRushi:
yup, I have updated the code, please do check again

MeherRushi:
yup
I'll do that

mauriciovasquezbernal:
@MeherRushi I updated this PR to align it with 3c90ec83ae4b017c94eb8e37657547dfbfa15a52. I'll merge it once the CI is green.

PR #3583, opened 2024-10-14
cilium/ebpf memory optimizations

alban:
# btfgen: optimization: return BTF when kernel supports it

This makes use of:
- https://github.com/cilium/ebpf/pull/1584
  - See also https://github.com/cilium/ebpf/pull/1588
- https://github.com/cilium/ebpf/pull/1585
  - See also https://github.com/cilium/ebpf/pull/1590

cilium/ebpf has two loading options:
- opts.KernelTypes: BTF for the kernel
- opts.KernelModuleTypes: map of BTFs for kernel modules

They are both optional: cilium/ebpf tries to figure out where the BTF is located if not provided by the caller.

Inspektor Gadget has different categories of ebpf programs:
1. Internal ebpf programs:
   - socket enricher
   - kfilefields
   - container-hook
2. Programs from builtin gadgets
3. Programs from image-based gadgets

In categories 1 and 2, ebpf programs are not attached to functions from kernel modules so we don't need cilium/ebpf support for that. Unfortunately, even without needing kernel module support, the cilium/ebpf implementation is costly in memory and cpu: it parses /proc/kallsyms to find out if a ebpf program should be attached to a function from a kernel module.

With a coming cilium/ebpf patch, we can avoid the cost of parsing /proc/kallsyms when providing opts.KernelTypes without providing opts.KernelModuleTypes. This patch does this to improve performances.

This patch does not affect programs from image-based gadgets (category 3), because opts.KernelTypes and opts.KernelModuleTypes are set from different code.

## How to use

No changes

## Testing done

With the patches:
```
export PPROF_ADDR=localhost:6061 ; go run -exec "sudo -E" ./cmd/ig/... -r docker trace exec
go tool pprof -http=:8081 http://localhost:6061/debug/pprof/heap
```

* With the patches, alloc_space: 44MB
image

* With the patches, inuse_space: 37MB
image

Without the patches:
```
export PPROF_ADDR=localhost:6062 ; go run -exec "sudo -E" ./cmd/ig/... -r docker trace exec
go tool pprof -http=:8082 http://localhost:6062/debug/pprof/heap
```

* Without the patches, alloc_space: 110MB
image

* Without the patches, inuse_space: 49MB
image


alban:
Replaced by https://github.com/inspektor-gadget/inspektor-gadget/pull/3616

PR #3571, opened 2024-10-10
Flush memory allocations when no longer needed

alban:
Flush btf kernel data from cache when no longer needed.

Also remove unnecessary regexp from the cri test.

```
sudo -E ./ig trace exec
go tool pprof -http=:8081 http://localhost:6060/debug/pprof/heap
```

Before: 45 MB

pprof-ig-1

After: 3 MB

pprof-ig-4

## How to use

No changes

## Testing done

- I have not checked how slower it becomes.
- I have not checked with pmap or smem, only pprof.
- The numbers above are with the builtin trace-exec gadget. With the image-based trace_exec gadget, there are still some btf allocations that are not flushed, so pprof shows 38MB in initEnumFormatters:

cc @burak-ok 

Fixes: #3541

PR #3585, opened 2024-10-15
cri: move tests to separate package

alban:
The tests depends on github.com/google/go-cmp/cmp which is not necessary to include in ig. The dependency compiles regexp which cause 0.5MB to be permanently used by the ig process. This is visible in the pprof flamegraph as 'regexp.MustCompile'.

This patch was originally part of #3571 but that PR might not be merged, so I am submitting the patch separately.

cc @burak-ok

### DIFF
--- a/pkg/container-utils/cri/cri.go
+++ b/pkg/container-utils/cri/cri.go
@@ -363,6 +363,16 @@ func parseExtraInfo(extraInfo map[string]string,
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


burak-ok (approved):
Looked up in the logs if `TestParseExtraInfo` is still running at all. Yes it is. We should probably do that for every unit test that imports big stuff.

Even though in theory the go compiler could see that we are not using anything from `go-cmp` in `ig`

Thanks :)

PR #3592, opened 2024-10-16
pkg/operators/otel-metrics: Add mechanism for graceful close server

blanquicet:
# Add mechanism for graceful close otel-metrics server

While writing some tests, I found the otel-metrics server doesn't have a mechanism to gracefully close the server for exporting data. This PR adds that support.

## How to use

No changes from users perspectives.

## Testing done

This PR adds some unit tests for it.

### DIFF
--- a/pkg/operators/otel-metrics/otel-metrics.go
+++ b/pkg/operators/otel-metrics/otel-metrics.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/operators/otel-metrics/otel-metrics.go
```diff
@@ -87,6 +89,8 @@ func (m *otelMetricsOperator) Init(globalParams *params.Params) error {
 		return nil
 	}
 
+	// TODO: Validate the listen address to immediately return an error
```
blanquicet, line 92: Any suggestion how to do it quickly?

pkg/operators/otel-metrics/otel-metrics.go
```diff
@@ -100,16 +104,37 @@ func (m *otelMetricsOperator) Init(globalParams *params.Params) error {
...
 				log.Errorf("serving otel metrics on: %s", err)
+				m.server = nil
 				return
 			}
 		}()
 	}
 	return nil
 }
 
+func (m *otelMetricsOperator) Close() {
```
blanquicet, line 125: Shouldn't we call these `Close` functions on [ops](https://github.com/inspektor-gadget/inspektor-gadget/blob/jose/test-otel-init/cmd/common/oci.go#L276)?

pkg/operators/otel-metrics/otel-metrics.go
blanquicet: [ResolveTCPAddr](https://pkg.go.dev/net#ResolveTCPAddr)?

flyth (commented):
Thanks for addressing this!

pkg/operators/otel-metrics/otel-metrics.go
mauriciovasquezbernal: But it should be handled by the gadget-context, shouldn't it be?

flyth: No - this is regarding the global operator part, not the operator instance (which is handled by the gadget-context).

pkg/operators/otel-metrics/otel-metrics.go
blanquicet: I did it manually because `ResolveTCPAddr` parses `<any-string>:1234` as `0.0.0.0:1234` instead of an error.

PR #3596, opened 2024-10-17
Add maps functions to WASM

eiffel-fl:
This commit adds the following functions to the WASM API:
* GetMap(name string): return a handle to an existing eBPF map.
* Map.Lookup(key any, value any): get the value corresponding to key in the map.
* Map.Put(key any, value any): store the value with the given key in the map.
* Map.Update(key any, value any, flags ebpf.MapUpdateFlags): update the value
corresponding to key in the map. The flags can be used to customize this
behavior.
* Map.Delete(key any): remove the key from the map.

They are basically wrappers around corresponding functions in cilium/ebpf.

### DIFF
--- a/docs/gadget-devel/gadget-wasm-api-raw.md
+++ b/docs/gadget-devel/gadget-wasm-api-raw.md
@@ -15,9 +15,9 @@ Data types passed to the API are encoded using 64 bits. Scalar types like
 integers, booleans and floats are casted directly to 64 bit integers and passed
 using the stack.
 
-Strings and byte arrays are stored in the wasm module's memory. A 64 bit
-integer is used to represent a pointer to them, the higher 32 bits contains the
-length and the lower 32 the memory address.
+Strings and byte arrays are stored in the wasm module's memory as `bufPtr`. A 64
+bit integer is used to represent a pointer to them, the higher 32 bits contains
+the length and the lower 32 the memory address.
 
 ## Wasm Module Exported Functions
 
@@ -313,7 +313,7 @@ Parameters:
 - `field` (u32): Field handle (as returned by `dataSourceGetField` or `dataSourceAddField`)
 - `data` (u32): Data handle
 - `kind` (u32): Kind of access: How to write the field
-  `value` (u64): Value to store in the field
+- `value` (u64): Value to store in the field
 
 Return value:
 - None
@@ -369,6 +369,54 @@ Parameters:
 - `key` (string): Key to set.
 - `val` (u64): Value to set.
 - `kind` (u32): Kind of `val`.
+
+Return value:
+- 0 in case of success, 1 otherwise.
+
+### eBPF Maps
+
+#### `getMap(name string) uint32`
+
+Get a handle to an existing eBPF map.
+
+Parameters:
+- `name` (string): Map's name
+
+Return value:
+- (u32) Handle to map on success, 0 on error.
+
+#### `mapLookup(m uint32, keyptr uint64, valueptr uint64) uint32`
+
+Lookup the map for a value corresponding to given key.
+
+Parameters:
+- `m` (u32): Map handle (as returned by `getMap`)
+- `keyptr` (u64): A `bufPtr` to key.
+- `valueptr` (u64): A `bufPtr` pointer to store the value of the map.
+
+Return value:
+- 0 in case of success, 1 otherwise.
+
+#### `mapUpdate(m uint32, keyptr uint64, valueptr uint64, flags uint64) uint32`
+
+Update the value corresponding to key in the given map.
+
+Parameters:
+- `m` (u32): Map handle (as returned by `getMap`)
+- `keyptr` (u64): A `bufPtr` to data corresponding to key.
+- `valueptr` (u64): A `bufPtr` to data corresponding to value.
+- `flags` (u64): A set of flags used to modify the update behavior. Correct values are documented [here](https://github.com/cilium/ebpf/blob/061e86d8f5e9/map.go#L790-L801).
+
+Return value:
+- 0 in case of success, 1 otherwise.
+
+#### `mapDelete(m uint32, keyptr uint64) uint32`
+
+Delete the value corresponding to key in the given map.
+
+Parameters:
+- `m` (u32): Map handle (as returned by `getMap`)
+- `keyptr` (u64): A `bufPtr` to data corresponding to key.
 
 Return value:
 - 0 in case of success, 1 otherwise.


pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,216 @@
+// Copyright 2024 The Inspektor Gadget authors
```
blanquicet, line 1: Once agreed on implementation, please remember to update [the API documentation](https://www.inspektor-gadget.io/docs/latest/gadget-devel/gadget-wasm-api-raw#api). Thanks!

wasmapi/go/map.go
```diff
@@ -0,0 +1,114 @@
...
+
+package api
+
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
```
alban, line 23: Why do we have this dependency in wasmapi?

pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,216 @@
...
+
+	"github.com/cilium/ebpf"
+	"github.com/tetratelabs/wazero"
+	wapi "github.com/tetratelabs/wazero/api"
+)
+
+func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
+	exportFunction(env, "getMap", i.getMap,
+		[]wapi.ValueType{wapi.ValueTypeI64}, // MapName
+		[]wapi.ValueType{wapi.ValueTypeI32}, // Map
```
alban, line 28: That sounds good to me.

Later we can add another function `getMapByID(uint32) uint32` so custom gadgets could pick up maps that were created externally (e.g. by bpftrace or another third-party tool). This was a use case from discussions at the All Systems Go conference.

pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		i.logger.Warnf("mapLookup: writing back value to stack: %v", err)
+		stack[0] = 0
+		return
+	}
+
+	stack[0] = 1
+}
+
+// mapUpdate update a value.
```
mauriciovasquezbernal, line 142: update -> updates

pkg/operators/wasm/testdata/map/program.go
```diff
@@ -0,0 +1,137 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This program tries as hard as it can to break the host by calling functions
+// with wrong arguments. It uses the low level functions directly as the goal is
+// to test the host and not the wrapper API. Tests under dataarray and fields
+// test also the higher level API.
```
mauriciovasquezbernal, line 18: Remove this.

pkg/operators/wasm/testdata/map/program.go
mauriciovasquezbernal, line 1: Please also provide tests in badguest.

mauriciovasquezbernal (commented):
Some initial comments. I'll do deeper later on.

pkg/operators/ebpf/ebpf.go
```diff
@@ -641,6 +641,10 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 	}
 	i.collection = collection
 
+	for name, m := range i.collection.Maps {
+		gadgetCtx.SetVar(name, m)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: When I said namespace, I meant something simple like using `"map/'` + mapName` to avoid maps clashing with other objects. Maps clashing with maps is something more difficult to support now. related to supporting multiple ebpf instances as well.

wasmapi/go/helpers.go
```diff
@@ -77,28 +77,37 @@ func anytoBufPtr(a any) (bufPtr, error) {
...
+		// Then, we need to take a pointer over the data instead of the data
+		// directly.
+		unsafePtr = unsafe.Pointer(&((*iface)(unsafe.Pointer(&a))).data)
+	default:
+		unsafePtr = ((*iface)(unsafe.Pointer(&a))).data
+	}
 
-	return buf
+	return bufPtr(uint64(size)<<32 | uint64(uintptr(unsafePtr))), nil
 }
```
mauriciovasquezbernal, line 111: I'd rather copy the data or implement a solution like https://github.com/cilium/ebpf//blob/e8bdb762e15776a8c37c2a96cc0a438b6ce1612b/internal/sysenc/marshal.go that avoids copying the data in some safe cases.

wasmapi/go/helpers.go
```diff
@@ -49,6 +53,54 @@ func stringToBufPtr(s string) bufPtr {
...
+	err := binary.Write(buffer, binary.NativeEndian, a)
+	if err != nil {
+		return bufPtr(0), fmt.Errorf("converting %T to []byte: %w", a, err)
+	}
+
+	// TODO Remove this when using binary.Encode().
+	copy(buf, buffer.Bytes())
+
+	return bytesToBufPtr(buf), nil
+}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > No, because the length of the buffer will only be the length of what is written while we want to have the length of the structure, even to have only final 0.

I see the binary package uses a [different way](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/encoding/binary/binary.go;l=680) than `reflect.TypeOf(a).Size()`, isn't it better to rely on the size returned by `binary.Write` than trying to calculate the size by ourselves?

Also, in what case do you think the written size could be smaller than the size of the structure?

wasmapi/go/helpers.go
```diff
@@ -49,6 +53,54 @@ func stringToBufPtr(s string) bufPtr {
...
+
+func readBufPtr(ptr bufPtr) []byte {
+	length := ptr >> 32
+	buf := make([]byte, length)
+
+	// My kingdom for a memcpy()...
+	copy(buf, unsafe.Slice((*byte)(unsafe.Pointer(uintptr(uint32(ptr)))), length))
+
+	return buf
+}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Why is the capacity relevant here? Because the copy logic on Map.Lookup?

I think ptr.bytes() could be implemented as readBufPtr.

mauriciovasquezbernal (commented):
It's looking very good. My only blocking comment for now is the one about using ptr.bytes() in bytesToBufPtr.

wasmapi/go/helpers.go
```diff
@@ -49,6 +52,26 @@ func stringToBufPtr(s string) bufPtr {
 	return bufPtr(uint64(len(s))<<32 | uint64(uintptr(unsafePtr)))
 }
 
+// anyToBufPtr returns a bufPtr that encodes the pointer and length of the
+// input.
+// The input is first encoded to binary buffer, which is then used as the
+// returned bufPtr.
+// WARNING the binary encoding will only works on with fixed size data, *i.e.*
```
mauriciovasquezbernal, line 59:
```suggestion
// WARNING the binary encoding will only work on with fixed size data, *i.e.*
```

pkg/operators/wasm/testdata/badguest/program.go
mauriciovasquezbernal, line 1: The idea of the badguest test is to try to break the host by passing wrong arguments to the functions it exposes. For the maps I can see:
- getMap: wrong pointer to map name
- mapLookup: bad map handle, bad key pointer, bad value pointer
- etc.

mauriciovasquezbernal (commented):
The implementation looks good to me. Looking forward for the docs.

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -341,3 +341,51 @@ Parameters:
...
+
+#### `getMap(name string) uint32`
+
+Get a handle to an existing eBPF map.
+
+Parameters:
+- `name` (string): Map's name
+
+Return value:
+- (u32) Handle to the data source on success, 0 on error.
```
mauriciovasquezbernal, line 355:
```suggestion
- (u32) Handle to map on success, 0 on error.
```

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -341,3 +341,51 @@ Parameters:
...
+
+#### `mapLookup(m uint32, keyptr uint64, valueptr uint64) uint32`
+
+Lookup the map for a value corresponding to given key.
+
+Parameters:
+- `m` (u32): Map handle (as returned by `getMap`)
+- `keyptr` (u64): A `bufPtr` to data corresponding to key.
+- `valueptr` (u64): A `bufPtr` to a pointer of data corresponding to value. The pointed data will be updated to reflect the content of the map.
```
mauriciovasquezbernal, line 364:
```suggestion
- `valueptr` (u64): A `bufPtr` pointer to store the value of the map.
```

?

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -341,3 +341,51 @@ Parameters:
...
+Return value:
+- (u32) Handle to the data source on success, 0 on error.
+
+#### `mapLookup(m uint32, keyptr uint64, valueptr uint64) uint32`
+
+Lookup the map for a value corresponding to given key.
+
+Parameters:
+- `m` (u32): Map handle (as returned by `getMap`)
+- `keyptr` (u64): A `bufPtr` to data corresponding to key.
```
mauriciovasquezbernal, line 363: Should we describe what a bufPtr is on "Data types"?

mauriciovasquezbernal, line 363:
```suggestion
- `keyptr` (u64): A `bufPtr` to key.
```

?

pkg/operators/wasm/testdata/badguest/program.go
```diff
@@ -256,6 +286,38 @@ func gadgetInit() int {
 	assertNonZero(setConfig(invalidStrPtr, stringToBufPtr("value"), uint32(api.Kind_String)), "setConfig: bad key ptr")
 	assertNonZero(setConfig(stringToBufPtr("key"), invalidStrPtr, uint32(api.Kind_String)), "setConfig: bad value ptr")
 
+	/* Map */
+	assertZero(getMap(0), "getMap: bad map pointer")
```
mauriciovasquezbernal, line 290: nit: use invalidStrPtr instead of 0 as 0 is handled in a special way: https://github.com/inspektor-gadget/inspektor-gadget/blob/4e50955197ed82a2a47ac67cbfbe844c6d56b351/pkg/operators/wasm/helpers.go#L61-L64

Same for pointers.

mauriciovasquezbernal (approved):
I'll try to use this next week but I'm approving it now. Perhaps wait a little bit in case somebody else wants to review.

docs/gadget-devel/gadget-wasm-api-raw.md
mauriciovasquezbernal: The bufPtr concept is no described on this file before. I think we need to describe a bufPtr is a uint64 encoding length and size. It's somehow explained on the "Data types" section of this document, but btfPtr is not explicitly mentioned.

pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	value, err := bufFromStack(m, valuePtr)
+	if err != nil {
+		i.logger.Warnf("mapLookup: getting a buf for value pointer: %v", err)
+		stack[0] = 1
+		return
+	}
+
+	err = ebpfMap.Update(key, value, ebpf.MapUpdateFlags(flags))
```
alban, line 169: I think we have to verify the type of ebpf map. For maps of type array or hash, that's fine. But for maps of type tail_call or other maps that accepts a file descriptor as value, there is a risk that the wasm module uses an integer and the kernel interprets that as a fd.

pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	key, err := bufFromStack(m, keyPtr)
+	if err != nil {
+		i.logger.Warnf("mapLookup: getting a buf for key pointer: %v", err)
+		stack[0] = 1
+		return
+	}
+
+	value, err := ebpfMap.LookupBytes(key)
```
alban, line 118: Similarly to `mapUpdate`, we should check the type of the map. For map types that would return a file descriptor, this would cause the kernel to add a file descriptor in the ig process. For now we can support arrays and hash tables. Other map types can be added later if needed.

pkg/operators/wasm/maps.go
alban: For future compatibility, we should verify the content of the flags is among the following values. As of today, the kernel supports:
- BPF_ANY
- BPF_NOEXIST
- BPF_EXIST
- BPF_F_LOCK

And none of those values would cause any problem when used from wasm. And the kernel correctly returns `-EINVAL` when using an unknown flag.

But if the kernel adds a new flag, we should assess whether it could cause a problem for Inspektor Gadget before automatically supporting it. For example, if a future flag causes the addition of a new file descriptor in the process, we would not want to allow it from wasm without more thinking about the correct way to do it.

pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	stack[0] = wapi.EncodeU32(i.addHandle(ebpfMap))
+}
+
+// mapLookup searches the map for a value corresponding to the given key.
+// Params:
+// - stack[0]: Map handle
+// - stack[1]: Key pointer
+// - stack[2]: Value pointer
```
alban, line 97: I am pondering about adding a flags parameter and use `ebpfMap.LookupWithFlags(...)`.

But that could be for a future PR. As of today, the only use for this would be for maps using spinlocks (`BPF_F_LOCK`).

pkg/operators/wasm/maps.go
```diff
@@ -0,0 +1,208 @@
...
+// Return value:
+// - 0 on success, 1 on error
+func (i *wasmOperatorInstance) mapLookup(ctx context.Context, m wapi.Module, stack []uint64) {
+	mapHandle := wapi.DecodeU32(stack[0])
+	keyPtr := stack[1]
+	valuePtr := stack[2]
+
+	ebpfMap, ok := getHandle[*ebpf.Map](i, mapHandle)
```
alban, line 105: Thinking aloud about race conditions:

I am looking at `pkg/operators/ebpf/ebpf.go` `ebpfInstance.Close()`: it calls `i.collection.Close()`, which will call the `close()` syscall on the map fd.

And below, the call to `ebpfMap.LookupBytes()` cannot check if the map fd has been closed (or even worse, the fd number reallocated to another ebpf map).

Do we have any guarantee that `ebpfInstance.Close()` is only called after the wasm module is stopped?

cc @mauriciovasquezbernal : similar kind of topic as in #3542.

mauriciovasquezbernal: > And below, the call to ebpfMap.LookupBytes() cannot check if the map fd has been closed (or even worse, the fd number reallocated to another ebpf map).

The cilium/ebpf library sets the FD of the map to -1 on Close() and returns `math.MaxUint32`, hence the kernel will return an error there. It's still an issue, but there is not a risk of using the wrong map.

> Do we have any guarantee that ebpfInstance.Close() is only called after the wasm module is stopped?

Unfortunately we don't have any. @burak-ok was suggesting the image layer operators should be stopped in reverse order, that will call wasm.Stop() before ebpf.Stop(). I think that's a good idea. However the order of the operators is [determined by the layers' order](https://github.com/inspektor-gadget/inspektor-gadget/blob/146c28a290bb1a1d14f1aff1acdcfed88681a80a/pkg/operators/oci-handler/oci.go#L310), so we'll need to sort those operators based on a priority to avoid it depending on the image.

pkg/operators/wasm/maps.go
alban: My issue was that the wasm operator would not know that a fd is added to the ig process, so it would never close it. So the process would leak file descriptors, even after the gadget is stopped.

I tried `eventsMap.Lookup` on a perf event array map (simulating an old kernel to avoid using the bpf ring buffer). It fails with ENOTSUPP:
```
[pid 76443] bpf(BPF_MAP_LOOKUP_ELEM, {map_fd=33, key=0xc003ad41e0, value=0xc005275f90, flags=BPF_ANY}, 32) = -1 ENOTSUPP (Unknown error 524)
```
See https://github.com/torvalds/linux/blob/v6.8/kernel/bpf/arraymap.c#L818

So it is not too bad for this map kind. But for maps of maps, this is still a problem.

I also tried to call Update:
```
		eventsMap, err := api.GetMap("events")
		if err != nil {
			api.Warnf("failed to get events map: %s", err)
			return
		}
		var zero [4]byte
		fd := uint32(44)
		err = eventsMap.Update(&zero, &fd, 0)
		if err != nil {
			api.Warnf("failed to lookup events map: %s", err)
			return
		}
```

Strace shows it fails with EPERM:
```
bpf(BPF_MAP_UPDATE_ELEM, {map_fd=13, key=0xc00394f620, value=0xc00394f670, flags=BPF_ANY}, 32) = -1 EPERM (Operation not permitted)
```

This is because the map was frozen in #3151.

But for maps of maps, that should work.

I don't think it is appropriate for the wasm program to poke around ig file descriptor by knowing the fd integer, rather than using a handle.

alban:
Is there a way to release a Map after calling `GetMap()`?

I get `WARN[0032] too many open handles` after calling `GetMap()` in the ds.Subscribe callback function.

burak-ok:
That is the same issue as in https://github.com/inspektor-gadget/inspektor-gadget/pull/3532#discussion_r1840507173 . I'm coming up with a PR soon to fix that

burak-ok:
> Is there a way to release a Map after calling `GetMap()`?

Opened PR for that in https://github.com/inspektor-gadget/inspektor-gadget/pull/3702

PR #3613, opened 2024-10-22
operators/wasm: add test for emitting packets from new datasource

Snehil-Shah:
# adds a test for creating a new data source and emitting single packets from wasm.

1. Wrote a test gadget that creates a new data source and emits data.
2. Exported gadget to a tar file.
3. Wrote wasm test for the above gadget.
4. Updated makefile artifacts list.

## How to use

Run `wasm_test.go`.

## Testing done

Ran the above-mentioned test.


pkg/operators/wasm/testdata/dataemit/gadget.yaml
```diff
@@ -0,0 +1,13 @@
+name: dataEmit
```
blanquicet, line 1: This file in not necessary. I'd remove it.

blanquicet (approved):
Apart from removing metadata file I don't have any other comment. LGTM!

pkg/operators/wasm/testdata/dataemit/gadget.yaml
Snehil-Shah: @blanquicet Done!

PR #3539, opened 2024-10-02
Add gadget user documentation for exporting metrics to Prometheus

flyth:
Ref: https://github.com/inspektor-gadget/inspektor-gadget/issues/3306

PR #3614, opened 2024-10-23
operators/otel-metrics: add support for implicit counter

flyth:
This adds support for an implicit counter that can be activated if the annotation `metrics.implicit-counter.name` contains a value. A description can optionally be added using `metrics.implicit-counter.description`. The added Int64Counter is increased whenever the targeted datasource emits an event.

Example of how to quickly convert an existing trace gadget into a metrics collector using this (Note though that events are collected in userspace, so this might not be viable for gadgets with high throughput. In those cases using a map iterator is preferable):

```sh
./ig run trace_open:latest --host --verify-image=false --otel-metrics-name open:open --annotate open:metrics.collect=true,open:metrics.implicit-counter.name=ctr,open.runtime.containername:metrics.type=key
```

### DIFF
--- a/docs/gadget-devel/metrics.md
+++ b/docs/gadget-devel/metrics.md
@@ -74,6 +74,23 @@ values.
 
 > Note: adding multiple metric.key annotations will increase cardinality a lot and take up more space and storage.
 
+#### Adding an implicit counter
+
+If you just want to count the number of events happening for a given set of keys, you can add another annotation to your
+datasource that will increase a counter whenever a new event has been emitted.
+
+```yaml
+datasources:
+  events:
+    annotations:
+      metrics.implicit-counter.name: eventctr
+      matrics.implicit-counter.description: number of events
+...
+```
+
+This is especially useful if you just want to quickly convert an existing event-based gadget to a metrics collector, as
+you can easily set annotations when running a gadget using the `--annotate` flag.
+
 ### Using well-known types in the eBPF code
 
 You can also edit your eBPF source code and use well-known types (TODO links to well known types) instead of annotating
--- a/pkg/operators/otel-metrics/otel-metrics.go
+++ b/pkg/operators/otel-metrics/otel-metrics.go
@@ -62,6 +62,9 @@ const (
 	AnnotationMetricsUnit        = "metrics.unit"
 	AnnotationMetricsBoundaries  = "metrics.boundaries"
 
+	AnnotationImplicitCounterName        = "metrics.implicit-counter.name"
+	AnnotationImplicitCounterDescription = "metrics.implicit-counter.description"
+
 	PrintDataSourceSuffix = "rendered"
 	PrintFieldName        = "text"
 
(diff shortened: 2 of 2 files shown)


docs/gadget-devel/metrics.md
```diff
@@ -74,6 +74,23 @@ values.
 
 > Note: adding multiple metric.key annotations will increase cardinality a lot and take up more space and storage.
 
+#### Adding an implicit counter
```
blanquicet, line 77: We need to document this also for gadget users on top of https://github.com/inspektor-gadget/inspektor-gadget/pull/3539, but it's fine if you want to do it on a follow-up PR, once both are merged.

pkg/operators/otel-metrics/otel-metrics.go
```diff
@@ -612,6 +615,22 @@ func (m *otelMetricsOperatorInstance) PreStart(gadgetCtx operators.GadgetContext
 
 		hasValueFields := false
 
+		// Support an implicit counter
+		if implicitCounter := ds.Annotations()[AnnotationImplicitCounterName]; implicitCounter != "" {
+			tOptions := make([]metric.Int64CounterOption, 0)
+			if implicitCounterDescription := ds.Annotations()[AnnotationImplicitCounterDescription]; implicitCounterDescription != "" {
+				tOptions = append(tOptions, metric.WithDescription(implicitCounterDescription))
```
blanquicet, line 622: It'd be awesome if we could provide more context in this description about when the counter is increased. For instance, if users is running this:

```bash
ig run trace_open:latest --verify-image=false --otel-metrics-name open:open --annotate open:metrics.collect=true,open:metrics.implicit-counter.name=NskubesysAndErr,open.runtime.containername:metrics.type=key -F k8s.namespace==kube-system,err!=0
```

We know that there will be a `ctrKubeSysAndErr` counter for each diff `containername` within the `kube-system` namespace and it'll increase every time a process fails to open a file. How to put that filtering information into the description? At least the `k8s.namespace==kube-system,err!=0`? It'll prevent users (like me 🤣) from naming metrics like `NskubesysAndErr` because the description will already contain that context.

I don't have that much experience with Prometheus/Grafana ... How usually people do that?
--- 

Not related with this PR: 

Issue with my idea: We have some filters with `--containername` and some others with `-F`. Another reason to merge them all into `-F`.

flyth: > It'd be awesome if we could provide more context in this description about when the counter is increased.

IIUC, that sounds like too much magic. Even if we could do things like that (`-F` and `--containername` aren't parameters that the otel-metrics operator knows about), it might result in something that most of the users wouldn't want in the first place.
Also, IMHO if you want the metrics in the way you mentioned, I would make `err` a `metrics.type=key` as well. Then you could simply filter by that value as well in the metrics frontend (Grafana) later on (at the cost of storage, ofc).

blanquicet: Fair

mauriciovasquezbernal (approved):
It looks useful to me. It's a bit weird that events are still printed to the terminal, but I suppose you'll handle that later on, right?

I'm approving it anyway.

flyth:
Thanks for the reviews!

> It looks useful to me. It's a bit weird that events are still printed to the terminal, but I suppose you'll handle that later on, right?

It doesn't have anything to do with the output and IMHO we shouldn't assume that when setting this the only output mode should be otel-metrics. That should be handled by the gadget author or user.

blanquicet:
> It looks useful to me. It's a bit weird that events are still printed to the terminal, but I suppose you'll handle that later on, right?
>
I would say that if users are only interested in metrics, they will use ig-daemon + gadgetctl-detach. Or, if they are running just a quick test to see what is exported, I don't think they mind it also prints things to the terminal.

PR #3616, opened 2024-10-23
go.mod: update cilium/ebpf with more memory optimization

alban:
This uses an unreleased version of cilium/ebpf.

The specific patches from cilium/ebpf are listed in a comment in go.mod.

This includes https://github.com/cilium/ebpf/pull/1590.

https://github.com/cilium/ebpf/pull/1588 was already picked up via [#3595](https://github.com/inspektor-gadget/inspektor-gadget/pull/3595).

cc @burak-ok

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -218,8 +218,10 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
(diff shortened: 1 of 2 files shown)


burak-ok (approved):
LGTM. And thanks again for kicking this off at all

PR #3620, opened 2024-10-24
build: fix dockerignore file

alban:
The dockerignore file must be named after the Dockerfile with the ".dockerignore" suffix. It was ignored by docker-build because it didn't have the correct name.

The content was wrong. Add missing directory.

Fixes: d8e9d0f2774c ("ig/local-gadget: Rename local-gadget in docker images and containers")

cc @burak-ok


burak-ok (approved):
TIL, thanks

PR #3621, opened 2024-10-24
Makefile: make the 'docker build' command configurable

alban:
This allows me to optionally switch to another builder like depot: https://depot.dev/

```bash
DOCKER_BUILD='depot build' time make ig
```
=> 41 seconds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ GOPROXY ?= $(shell go env GOPROXY)
 
 DLV ?= dlv
 
+BUILD_COMMAND ?= docker buildx build
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,6 +17,8 @@ GOPROXY ?= $(shell go env GOPROXY)
 
 DLV ?= dlv
 
+DOCKER_BUILD ?= docker buildx build
```
mauriciovasquezbernal, line 20: BUILD_COMMAND? (since it's not specific to docker?

mauriciovasquezbernal (approved):

PR #3622, opened 2024-10-24
Move authfile and pull-secret to GlobalParams and update documentation

Priyansurout:
# moving the `authfile` and `pull-secret` parameters from InstanceParams to GlobalParams.

- Moved `authfile` and `pull-secret` to GlobalParams in `pkg/operators/oci-handler/oci.go`.
- Updated the corresponding documentation to reflect this change.

Closes [#3444](https://github.com/inspektor-gadget/inspektor-gadget/issues/3444).

PR #3628, opened 2024-10-24
oci: Expose pull-secret param only when running on k8s

mauriciovasquezbernal:
This param is k8s-specific, hence only expose it on this case.

### Before 

Using `--pull-secret` with ig will cause it to try to reach the api server, this is wrong as ig shouldn't try to use this secret.

```bash 
$ sudo ig run trace_open:latest --pull-secret=foo
Error: fetching gadget information: initializing and preparing operators: instantiating operator "oci": getting secret "foo": Get "https://192.168.49.2:8443/api/v1/namespaces/gadget/secrets/foo": dial tcp 192.168.49.2:8443: connect: no route to host
```

### After 

`--pull-secret` flag isn't available when running ig

```bash 
$ sudo ig run trace_open:latest --pull-secret=foo
Error: unknown flag: --pull-secret
```

Found while testing #3622

### DIFF
--- a/pkg/operators/oci-handler/oci.go
+++ b/pkg/operators/oci-handler/oci.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


mqasimsarfraz (approved):
LGTM! I tested it and it works fine!

PR #3641, opened 2024-10-28
operators/cli: Improve logic to print header

### LINKED ISSUE #3264: Columns header is printed twice for snapshotters

mauriciovasquezbernal:
Printing the header when using the columns mode is a simple but complicated task as there are multiple things to take into account.

1. It should only be printed once for gadgets that dump all the information at once. Like snapshotters.
2. It should be printed right away for gadgets that collect and print the information later on (like toppers).

This commit is a "best effort" to fix that, probably there are corner cases not covered yet.

Supersedes #3281
Fixes #3264

### DIFF
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -321,10 +321,13 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error 
 			continue
 		}
 
+		clearScreenBefore := ds.Annotations()[AnnotationClearScreenBefore] == "true"
+		isTerminal := term.IsTerminal(int(os.Stdout.Fd()))
+
 		switch mode {
 		default:
 			before := func() {}
-			if ds.Annotations()[AnnotationClearScreenBefore] == "true" && term.IsTerminal(int(os.Stdout.Fd())) {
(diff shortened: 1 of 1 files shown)


flyth (approved):
LGTM from code inspection.

Hard to wrap your head around all use-cases here - I'll test some more later on.

PR #3651, opened 2024-10-31
feat: enhance version command output to json format

Priyansurout:
# enhance version command output to give json format

git_pr

## How to use

 kubectl-gadget version -o json

### DIFF
--- a/cmd/kubectl-gadget/version.go
+++ b/cmd/kubectl-gadget/version.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -24,7 +25,21 @@ import (
 	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
 )
 
+// VersionInfo represents the structure for version information output
+type VersionInfo struct {
+	ClientVersion *Version `json:"clientVersion,omitempty"`
+	ServerVersion *Version `json:"serverVersion,omitempty"`
+}
+
+// Version contains detailed version information
+type Version struct {
+	Version string `json:"version"`
+}
+
+var outputFormat string
+
 func init() {
+	versionCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format. One of: json|''")
 	rootCmd.AddCommand(versionCmd)
 }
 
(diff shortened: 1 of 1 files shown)


cmd/kubectl-gadget/version.go
```diff
@@ -15,16 +15,29 @@
...
+}
+
+// Version contains detailed version information
+type Version struct {
+	Version string `json:"version"`
+	Status  string `json:"status,omitempty"`
+}
+
 func init() {
+	versionCmd.Flags().StringP("output", "o", "", "Output format. One of: json|''")
```
mauriciovasquezbernal, line 40: nit: I think it's better to use StringVarP, so you have a global variable with the value and don't need to call `cmd.Flags().GetString("output")` below.

cmd/kubectl-gadget/version.go
```diff
@@ -33,32 +46,70 @@ var versionCmd = &cobra.Command{
 	Short:        "Show version",
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Printf("Client version: v%s\n", version.Version())
+		outputFormat, err := cmd.Flags().GetString("output")
+		if err != nil {
+			return fmt.Errorf("error getting output format: %w", err)
```
mauriciovasquezbernal, line 51: nit: do not use the "error ..." prefix on error messages.

```suggestion
			return fmt.Errorf("getting output format: %w", err)
```

cmd/kubectl-gadget/version.go
```diff
@@ -33,32 +46,70 @@ var versionCmd = &cobra.Command{
...
-		return nil
-	},
+func outputDefault(info *VersionInfo) error {
+	fmt.Printf("Client version: v%s\n", info.ClientVersion.Version)
+	if info.ServerVersion.Status != "" {
+		fmt.Printf("Server version: %s\n", info.ServerVersion.Status)
+	} else {
+		fmt.Printf("Server version: v%s\n", info.ServerVersion.Version)
+	}
+	return nil
```
mauriciovasquezbernal, line 114: Does it really make sense to have them as functions or can we put the value directly on the switch statement above?

cmd/kubectl-gadget/version.go
```diff
@@ -15,16 +15,29 @@
...
+// VersionInfo represents the structure for version information output
+type VersionInfo struct {
+	ClientVersion *Version `json:"clientVersion,omitempty"`
+	ServerVersion *Version `json:"serverVersion,omitempty"`
+}
+
+// Version contains detailed version information
+type Version struct {
+	Version string `json:"version"`
+	Status  string `json:"status,omitempty"`
```
mauriciovasquezbernal, line 36: I'm not sure about this status field. I think we don't need it. (see comment below)

cmd/kubectl-gadget/version.go
```diff
@@ -33,32 +46,70 @@ var versionCmd = &cobra.Command{
...
 
-		return nil
-	},
+func outputDefault(info *VersionInfo) error {
+	fmt.Printf("Client version: v%s\n", info.ClientVersion.Version)
+	if info.ServerVersion.Status != "" {
+		fmt.Printf("Server version: %s\n", info.ServerVersion.Status)
+	} else {
+		fmt.Printf("Server version: v%s\n", info.ServerVersion.Version)
+	}
```
mauriciovasquezbernal, line 113: What about the following logic:

If server version is not empty, print it. Otherwise print a generic "not-available" message when it's empty?

mauriciovasquezbernal (commented):
Some comments. Thanks for working on this.

cmd/kubectl-gadget/version.go
Priyansurout: I agree StringVarP would make the code cleaner

cmd/kubectl-gadget/version.go
Priyansurout: yes good idea we can do it directly in the switch statement

cmd/kubectl-gadget/version.go
Priyansurout: yes we can do it

mauriciovasquezbernal:
Can you please squash all commits into one?

cmd/kubectl-gadget/version.go
```diff
@@ -15,16 +15,30 @@
...
 import (
+	"encoding/json"
 	"fmt"
 
-	"github.com/spf13/cobra"
-
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/internal/version"
 	grpcruntime "github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc"
+	"github.com/spf13/cobra"
```
mauriciovasquezbernal, line 24: Please run `make lint` to fix the order of these imports. LGTM otherwise.

Priyansurout: done

mauriciovasquezbernal (approved):

PR #3652, opened 2024-10-31
field: add `AddTags` method to fieldAccessor and corresponding WASM API

Snehil-Shah:
# adds methods to add tags to fields

- Adds `AddTags` method to pkg/datasource/accessors
- Adds corresponding methods to `Field` in WASM API (and go implementation)

## How to use

You can add tags to a field through WASM now. Example:

```go
f = ds.AddField("mntns_id", api.Kind_Uint64)
f.AddTag("type:gadget_mntns_id") // enrich mntns id
```

## Testing done

Observed implications of the above code example.


alban:
The code looks good to me.

Could you add a test in pkg/operators/wasm/wasm_test.go, using `AddTag` and `HasAllTagsOf`?

Snehil-Shah:
@alban Done!

alban (approved):
LGTM

pkg/datasource/accessors.go
```diff
@@ -392,6 +395,10 @@ func (a *fieldAccessor) Tags() []string {
 	return slices.Clone(a.f.Tags)
 }
 
+func (a *fieldAccessor) AddTag(tag string) {
+	a.f.Tags = append(a.f.Tags, tag)
```
flyth, line 399: Please lock this section like done with the other public functions:

```golang
	a.ds.lock.Lock()
	defer a.ds.lock.Unlock()
```

It seems like it's missing in some other functions as well - we need to address that as well, but shouldn't introduce more functions lacking it, IMHO.

flyth (commented):
Thanks for adding this!

Maybe it'd be good to have the internal API use a variadic `AddTags()` any only limit the WASM API to one tag.

Snehil-Shah:
@flyth I didn't do that because `datasource` also has an `AddTag` method with a plural `Tags()` GET method, so a singular `AddTag` would be more consistent with the rest of the APIs, given it also has a plural `Tags()` method (I think?).

flyth:
> @flyth I didn't do that because `datasource` also has an `AddTag` method with a plural `Tags()` GET method, so a singular `AddTag` would be more consistent with the rest of the APIs, given it also has a plural `Tags()` method (I think?).

Hmm, it seems the `AddTag()` of `datasource` is never used for now. So maybe we can instead rename that one to `AddTags()` and make it  `AddTags()`, then? WDYT?

Snehil-Shah:
@flyth makes sense, but it will also introduce inconsistency with the wasm api's singular `AddTag` method😬.. your call.

Snehil-Shah:
We can have `AddTags` on wasm as well, but would make the implementation a bit complicated ie. grouping bytes into groups of 8 (string enc uint64).

alban:
The tests fail with this error message:
```
time="2024-11-04T08:55:24Z" level=warning msg="error from map iterator: bad file descriptor"
```

burak-ok (approved):
LGTM, thanks.
Also checked the contents of the `.tar`

PR #3654, opened 2024-10-31
operators/ebpf: Fix map iterators waiting logic

mauriciovasquezbernal:
The previous logic was relying on the gadget context, but it was never being closed. This caused the map iterator to keep reading the map forever, producing the following error after a timeout:

```bash
 sudo ig run mygadget:latest --verify-image=false -o json --timeout 2WARN[0001] image signature verification is disabled due to using corresponding option
[]
[]
WARN[0004] error from map iterator: bad file descriptor
[]
WARN[0005] error from map iterator: bad file descriptor
[]
```
This logic switches to only rely on the Stop() operation and an internal channel.

This bug only appeared after 85b756c881d3 ("operators/ebpf: Fix race condition with wasm operator") as it introduced a waiting group, but the bug was present before.

Fixes: f6aedf67c485 ("pkg/operators/ebpf: introduce map iterators")

### DIFF
--- a/examples/gadgets/operators/otel_metrics/main.go
+++ b/examples/gadgets/operators/otel_metrics/main.go
@@ -20,6 +20,7 @@ import (
 	"math/rand"
 	"os"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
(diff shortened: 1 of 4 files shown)


flyth (approved):
LGTM from code inspection.

Since you already touched the example, maybe it'd be good to add the same WaitGroup logic there as a "best practice" example.

mauriciovasquezbernal:
I added the waitgroup, however the example is not working. The following diff fixes it: 

```diff
 git diff
diff --git a/examples/gadgets/operators/otel_metrics/main.go b/examples/gadgets/operators/otel_metrics/main.go
index 17bf872c7..270a4f2c9 100644
--- a/examples/gadgets/operators/otel_metrics/main.go
+++ b/examples/gadgets/operators/otel_metrics/main.go
@@ -173,7 +173,7 @@ func do() error {
        defer runtime.Close()

        params := map[string]string{
-               "operator.otel-metrics.otel-metrics-name": "metrics",
+               "operator.otel-metrics.otel-metrics-name": "metrics:metrics",
        }

        // Run the gadget
```

As it's not related to this PR, I'll move forward and handle it on a different one.

PR #3659, opened 2024-11-03
gadgets: add qdisc_latency tracer

### LINKED ISSUE #3118: [RFE] New gadget for qdisc latency
## Current situation

We don't have a gadget telling how much time a network packet is waiting in the network scheduler.

## Impact

This is one piece missing to investigate network problems.

## Ideal future situation

Histogram showing latency in network scheduler, similar to top-block-io gadget.

## Implementation options

We can use the following tracepoints:
```
/sys/kernel/tracing/events/qdisc/qdisc_enqueue/format
/sys/kernel/tracing/events/qdisc/qdisc_dequeue/format
/sys/kernel/tracing/events/skb/consume_skb/format
/sys/kernel/tracing/events/skb/kfree_skb/format
```

We can use a hash table keyed by the skb and measure the time between the enqueue and the dequeue. Note that the dequeue could dequeue several skbs in the same call, so we also need consume_skb and kfree_skb to clean up the map.

## Additional information

cc @blanquicet

patrickpichler:
# Introduce qdisc_latency profiler

In order to analyse network related issues, the qdisc_latency profiler tool has been added. It will visualise the amount a network packet spend in the network scheduler in an histogram. 

**Note:** The profiler requires at least kernel 5.14, as the tracepoints used were introduced in that kernel version. 

## How to use

Run the new profiler via `ig profile qdisc-latency`. 

## Testing done

I tried running the gadget while performing various network tasks. One strange thing I could not get figured out though was, that I never managed to get a packet queued in my qdisc. Maybe somebody has an idea here? 

Fix #3118

### DIFF
--- a/gadgets/profile_qdisc_latency/README.mdx
+++ b/gadgets/profile_qdisc_latency/README.mdx
@@ -1,0 +1,351 @@
+---
+title: profile_qdisc_latency
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# profile_qdisc_latency
+
+The `profile_qdisc_latency` gadget gathers information about the usage of the
+network interfaces, generating a histogram distribution of latency caused by
+the network scheduler when consuming packets off qdiscs, when the gadget is stopped.
+
+The histogram shows the number of packets enqueued to qdiscs (`count` column) that lie in the
+latency range `interval-start` -> `interval-end` (`µs` column). By default the latency is
+measured in microseconds. If the `--ms` flag is passed, it will be shown in milliseconds.
+
+This guide will use the [netem qdisc](https://man7.org/linux/man-pages/man8/tc-netem.8.html) to
+emulate delay in sending packets. To configure it, the [tc](https://man7.org/linux/man-pages/man8/tc.8.html)
+program of the `iproute2` is used.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_qdisc_latency:%IG_TAG% [flags]
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/profile_qdisc_latency:%IG_TAG% [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Guide
+
+Run the gadget in a terminal:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run profile_qdisc_latency:%IG_TAG% --node minikube-docker
+        ```
+
+        It will start to display the qdisc latency distribution as follows:
+        ```bash
+        latency
+        µs               : count    distribution
+               0 -> 1          : 2        |****************************************|
+               1 -> 2          : 0        |                                        |
+               2 -> 4          : 0        |                                        |
+               4 -> 8          : 0        |                                        |
+               8 -> 16         : 0        |                                        |
+              16 -> 32         : 0        |                                        |
+              32 -> 64         : 0        |                                        |
+              64 -> 128        : 0        |                                        |
+             128 -> 256        : 0        |                                        |
+             256 -> 512        : 0        |                                        |
+             512 -> 1024       : 0        |                                        |
+            1024 -> 2048       : 0        |                                        |
+            2048 -> 4096       : 0        |                                        |
+            4096 -> 8192       : 0        |                                        |
+            8192 -> 16384      : 0        |                                        |
+           16384 -> 32768      : 0        |                                        |
+           32768 -> 65536      : 0        |                                        |
+           65536 -> 131072     : 0        |                                        |
+          131072 -> 262144     : 0        |                                        |
+          262144 -> 524288     : 0        |                                        |
+          524288 -> 1048576    : 0        |                                        |
+         1048576 -> 2097152    : 0        |                                        |
+         2097152 -> 4194304    : 0        |                                        |
+         4194304 -> 8388608    : 0        |                                        |
+         8388608 -> 16777216   : 0        |                                        |
+        16777216 -> 33554432   : 0        |                                        |
+        33554432 -> 67108864   : 0        |                                        |
+        ```
+
+        Now to introduce some more latency, let's add a `netem` qdisc with some latency and jitter.
+
+        ```bash
+        # Start by creating our testing namespace
+        $ kubectl create ns qdisc-latency-test
+
+        # Run a pod that we will emulate network latency in
+        $ kubectl run -n qdisc-latency-test --rm -it netem-test \
+            --image=alpine --restart=Never \
+            --privileged
+
+        # Inside the container run the following commands
+        $ apk update
+        $ apk add iproute2
+        # This will introduce a latency of 100ms with 100ms jitter
+        $ tc qdisc add dev eth0 root netem delay 100ms 100ms
+
+        # Now let's ping some remote host
+        $ ping google.com
+        ```
+
+        Using the profile qdisc-latency gadget, we can generate another histogram to analyse the
+        latency of scheduled network packets:
+
+        ```bash
+        # Run the gadget again
+        $ kubectl gadget run profile_qdisc_latency:%IG_TAG% --node minikube-docker
+        latency
+              µs               : count    distribution
+               0 -> 1          : 2        |****                                    |
+               1 -> 2          : 0        |                                        |
+               2 -> 4          : 0        |                                        |
+               4 -> 8          : 1        |**                                      |
+               8 -> 16         : 0        |                                        |
+              16 -> 32         : 5        |***********                             |
+              32 -> 64         : 14       |********************************        |
+              64 -> 128        : 17       |****************************************|
+             128 -> 256        : 0        |                                        |
+             256 -> 512        : 0        |                                        |
+             512 -> 1024       : 0        |                                        |
+            1024 -> 2048       : 0        |                                        |
+            2048 -> 4096       : 0        |                                        |
+            4096 -> 8192       : 0        |                                        |
+            8192 -> 16384      : 0        |                                        |
+           16384 -> 32768      : 0        |                                        |
+           32768 -> 65536      : 0        |                                        |
+           65536 -> 131072     : 0        |                                        |
+          131072 -> 262144     : 0        |                                        |
+          262144 -> 524288     : 0        |                                        |
+          524288 -> 1048576    : 0        |                                        |
+         1048576 -> 2097152    : 0        |                                        |
+         2097152 -> 4194304    : 0        |                                        |
+         4194304 -> 8388608    : 0        |                                        |
+         8388608 -> 16777216   : 0        |                                        |
+        16777216 -> 33554432   : 0        |                                        |
+        33554432 -> 67108864   : 0        |                                        |
+        ```
+
+        The new histogram shows how the latency numbers increased.
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run profile_qdisc_latency:%IG_TAG%
+        ```
+
+        It will start to display the qdisc latency distribution as follows:
+
+        ```bash
+        latency
+        µs               : count    distribution
+               0 -> 1          : 2        |****************************************|
+               1 -> 2          : 0        |                                        |
+               2 -> 4          : 0        |                                        |
+               4 -> 8          : 0        |                                        |
+               8 -> 16         : 0        |                                        |
+              16 -> 32         : 0        |                                        |
+              32 -> 64         : 0        |                                        |
+              64 -> 128        : 0        |                                        |
+             128 -> 256        : 0        |                                        |
+             256 -> 512        : 0        |                                        |
+             512 -> 1024       : 0        |                                        |
+            1024 -> 2048       : 0        |                                        |
+            2048 -> 4096       : 0        |                                        |
+            4096 -> 8192       : 0        |                                        |
+            8192 -> 16384      : 0        |                                        |
+           16384 -> 32768      : 0        |                                        |
+           32768 -> 65536      : 0        |                                        |
+           65536 -> 131072     : 0        |                                        |
+          131072 -> 262144     : 0        |                                        |
+          262144 -> 524288     : 0        |                                        |
+          524288 -> 1048576    : 0        |                                        |
+         1048576 -> 2097152    : 0        |                                        |
+         2097152 -> 4194304    : 0        |                                        |
+         4194304 -> 8388608    : 0        |                                        |
+         8388608 -> 16777216   : 0        |                                        |
+        16777216 -> 33554432   : 0        |                                        |
+        33554432 -> 67108864   : 0        |                                        |
+        ```
+
+        Now to introduce some more latency, let's add a `netem` qdisc with some latency and jitter.
+
+        ```bash
+        # We spawn a container where we will add network latency to
+        $ docker run --rm --privileged --name qdisc-latency-test -it alpine
+
+        # Inside the container run the following commands
+        apk update
+        apk add iproute2
+        # This will introduce a latency of 100ms with 100ms jitter
+        tc qdisc add dev eth0 root netem delay 100ms 100ms
+
+        # Now let's ping some remote host
+        ping google.com
+        ```
+
+        Using the profile qdisc-latency gadget, we can generate another histogram to analyse the
+        latency of scheduled network packets:
+
+        ```bash
+        $ sudo ig run profile_qdisc_latency:%IG_TAG%
+        latency
+              µs               : count    distribution
+               0 -> 1          : 2        |****                                    |
+               1 -> 2          : 0        |                                        |
+               2 -> 4          : 0        |                                        |
+               4 -> 8          : 1        |**                                      |
+               8 -> 16         : 0        |                                        |
+              16 -> 32         : 5        |***********                             |
+              32 -> 64         : 14       |********************************        |
+              64 -> 128        : 17       |****************************************|
+             128 -> 256        : 0        |                                        |
+             256 -> 512        : 0        |                                        |
+             512 -> 1024       : 0        |                                        |
+            1024 -> 2048       : 0        |                                        |
+            2048 -> 4096       : 0        |                                        |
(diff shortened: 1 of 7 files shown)


alban:
Thanks for the contribution!

We are trying to migrate from built-in gadgets (source in `pkg/gadgets/`) to image-based gadgets (source `gadgets/`). Would it be possible to implement this new gadget as image-based gadget? As a profiler, you could take example on `gadgets/profile_blockio/`.

> One strange thing I could not get figured out though was, that I never managed to get a packet queued in my qdisc. Maybe somebody has an idea here?

Did you test with the default qdisc? Could you try with the [netem](https://man7.org/linux/man-pages/man8/tc-netem.8.html) qdisc?

```bash
tc qdisc add dev eth0 root netem delay 100ms
```

patrickpichler:
> We are trying to migrate from built-in gadgets
Ahh sorry I missed this. I will reimplement it as a image based profiler then 👌 

> Did you test with the default qdisc
I did test it with the default qdisc, as well as netem and even tried to use something like hsfc. I am not sure why (I am by no means an expert in qdisc/linux networking though), but the tracepoint was never hit. I also tried hooking it with `bpftrace` and had the same issues. I think [this](https://elixir.bootlin.com/linux/v5.15/source/net/core/dev.c#L4212) if never evaluates to `true`, but I didn't have time to trace when it should be set to true.

alban:
> > We are trying to migrate from built-in gadgets
>
> Ahh sorry I missed this. I will reimplement it as a image based profiler then 👌

Thanks :)
There is some background on this [blog post](https://inspektor-gadget.io/blog/2024/08/empowering-observability_the_advent_of_image_based_gadgets) and some documentation (like [Hello world gadget](https://inspektor-gadget.io/docs/latest/gadget-devel/hello-world-gadget)).

> > Did you test with the default qdisc
>
> I did test it with the default qdisc, as well as netem and even tried to use something like hsfc. I am not sure why (I am by no means an expert in qdisc/linux networking though), but the tracepoint was never hit. I also tried hooking it with `bpftrace` and had the same issues. I think [this](https://elixir.bootlin.com/linux/v5.15/source/net/core/dev.c#L4212) if never evaluates to `true`, but I didn't have time to trace when it should be set to true.

Does it mean the gadget has no output at all? Are the other tracepoints working (qdisc_dequeue, consume_skb, kfree_skb)?

patrickpichler:
Weird, I tried again with the `netem` qdisc and I can now trace it with `bpftrace`. I need to look into this again, something weird is happening. 

I will update the PR once I ported the program and figured out more :)

patrickpichler:
Alright, I ported the program to be a gadget. Finally my local qdisc are working again (still weird that the tracepoint never fired before, but whatever). 

I tried to base the gadget on `profile_tcprtt`, as it also uses parameters.

gadgets/profile_qdisc_latency/program.bpf.c
```diff
@@ -0,0 +1,125 @@
...
+	if (slot >= PROFILER_MAX_SLOTS)
+		slot = PROFILER_MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->latency[slot], 1);
+
+cleanup:
+	bpf_map_delete_elem(&start, &skb);
+}
+
+SEC("raw_tp/qdisc_enqueue")
+int BPF_PROG(qdisc_enqueue, struct Qdisc *qdisc, const struct netdev_queue *txq,
```
alban, line 93: Would it be possible to filter on a specific network interface?

We have the network interface name and index at:
* txq->dev.name
* txq->dev.ifindex

alban (commented):
Thanks!

gadgets/profile_qdisc_latency/program.bpf.c
patrickpichler: I have no idea how to get the name in as a parameter 😅 I looked at the other gadgets and I guess arrays are currently not supported, or am I missing something?

Another thing though, are `ifindex` number unique across network namespaces though? From what I've gathered by searching the internet, it hints that indexes might get reused. Also a local experiment with running a container shows, that the loopback interface has the same `ifnumber` as the host.

patrickpichler: @alban what do you think?

alban: > I have no idea how to get the name in as a parameter 😅 I looked at the other gadgets and I guess arrays are currently not supported, or am I missing something?

I am not sure how to do it. We can leave that for a future PR.

> Another thing though, are ifindex number unique across network namespaces though? From what I've gathered by searching the internet, it hints that indexes might get reused. Also a local experiment with running a container shows, that the loopback interface has the same ifnumber as the host.

The `lo` interface is special and AFAIK the same number is reused across network namespaces.

ifindex is namespaced since Linux 3.7 (see https://github.com/torvalds/linux/commit/aa79e66eee5d525e2fcbd2a5fcb87ae3dd4aa9e9 and [details about link-netnsid](https://unix.stackexchange.com/questions/441876/how-to-find-the-network-namespace-of-a-veth-peer-ifindex)), so we cannot assume that `ifindex` is globally uniquely identifying the network interface & network namespace. However, in the case of veth pairs, they are _very likely_ to be unique because they are created by the CNI plugin on the host netns before being moved to the container netns.

mayasingh17: Hi @patrickpichler is there anything else someone from our team might be able to help with?

patrickpichler: Hey, 
Sorry for the long radio silence. I was busy during the last week. The code is pretty much done, I just need to test it.

patrickpichler: I am done with testing. @alban from what i see locally, the `ifindex` of the `eth0` device in the network namespace for my local kind cluster is always `2`. I am not 100% sure how useful this is to be completely honest.

alban: I don't know how kind is configuring the network. It might be correct that it is always 2.

There is ongoing work with mapping the ifindex with the device name in #3913, but it's best to keep that for another PR.

patrickpichler: I looked into it a bit more. Kindnet (the default CNI of kind) simply creates a veth pair in the network namespace via netlink. Since in the containers network namespace only the loopback interface exists, it will always be 2. 

Anyway, do you think this PR is ready to merge, or is there any blocking issue? :)

gadgets/profile_qdisc_latency/program.bpf.c
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if (targ_ms)
+		delta /= 1000000U;
+	else
+		delta /= 1000U;
+	slot = log2l(delta);
+	if (slot >= PROFILER_MAX_SLOTS)
+		slot = PROFILER_MAX_SLOTS - 1;
+	__sync_fetch_and_add(&histp->latency[slot], 1);
```
alban, line 81: I added a debug:
```
bpf_printk("delta: %lld, slot: %llu", delta, slot);
```

I run the experiment with 300ms:
```
$ docker run -ti --rm --cap-add NET_ADMIN --name=netem wbitt/network-multitool -- /bin/bash
# tc qdisc add dev eth0 root netem delay 300ms 1ms 25%
# wget -O /dev/null https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.13.tar.xz
```

And I run the gadget with trace_pipe enabled:
```
sudo ./ig run ghcr.io/inspektor-gadget/gadget/profile_qdisc_latency:HEAD --verify-image=false --trace-pipe -v 2>&1 |strings
```
And I get:
```
delta: 300417, slot: 18
```
I get a delta of 300ms + 1ms jitter as expected, but in the wrong slot.


I tested the following program:

<details>

```c
#include <stdio.h>
#include <stdint.h>

static __always_inline uint64_t custom_log2(uint32_t v)
{
        uint32_t shift, r;

        r = (v > 0xFFFF) << 4;
        v >>= r;
        shift = (v > 0xFF) << 3;
        v >>= shift;
        r |= shift;
        shift = (v > 0xF) << 2;
        v >>= shift;
        r |= shift;
        shift = (v > 0x3) << 1;
        v >>= shift;
        r |= shift;
        r |= (v >> 1);

        return r;
}

static __always_inline uint64_t custom_log2l(uint64_t v)
{
        uint32_t hi = v >> 32;

        if (hi)
                return custom_log2(hi) + 32;
        else
                return custom_log2(v);
}

int main() {
    for (int i = 0; i <= 10; i++) {
        printf("custom_log2l(%d) = %lu\n", i, custom_log2l(i));
    }

    return 0;
}
```

</details>

It gives the output:

```
$ ./log2l 
custom_log2l(0) = 0
custom_log2l(1) = 0
custom_log2l(2) = 1
custom_log2l(3) = 1
custom_log2l(4) = 2
custom_log2l(5) = 2
custom_log2l(6) = 2
custom_log2l(7) = 2
custom_log2l(8) = 3
custom_log2l(9) = 3
custom_log2l(10) = 3
```

Which is the wrong slot for input 3:
```
     0           0 -> 1          : 0        |                                        |
     1           1 -> 2          : 0        |                                        |
     2           2 -> 4          : 0        |                                        |
     3           4 -> 8          : 0        |                                        |
     4           8 -> 16         : 0        |                                        |
     5          16 -> 32         : 0        |                                        |
     6          32 -> 64         : 0        |                                        |
     7          64 -> 128        : 0        |                                        |
     8         128 -> 256        : 0        |                                        |
     9         256 -> 512        : 0        |                                        |
    10         512 -> 1024       : 0        |                                        |
    11        1024 -> 2048       : 0        |                                        |
    12        2048 -> 4096       : 0        |                                        |
    13        4096 -> 8192       : 0        |                                        |
    14        8192 -> 16384      : 0        |                                        |
    15       16384 -> 32768      : 0        |                                        |
    16       32768 -> 65536      : 0        |                                        |
    17       65536 -> 131072     : 0        |                                        |
    18      131072 -> 262144     : 10562    |****************************************|
    19      262144 -> 524288     : 0        |                                        |
    20      524288 -> 1048576    : 0        |                                        |
    21     1048576 -> 2097152    : 0        |                                        |
    22     2097152 -> 4194304    : 0        |                                        |
    23     4194304 -> 8388608    : 0        |                                        |
    24     8388608 -> 16777216   : 0        |                                        |
    25    16777216 -> 33554432   : 0        |                                        |
    26    33554432 -> 67108864   : 0        |                                        |
```

It seems there is a bug in log2l or its usage (possibly in other gadgets too).

alban (commented):
Thanks! I added some comments with things to fix, but I think it is pretty close to ready.

gadgets/profile_qdisc_latency/program.bpf.c
patrickpichler: Interesting! I copied the code from another histogram gadget. I will have a look!

gadgets/profile_qdisc_latency/README.mdx
```diff
@@ -0,0 +1,320 @@
...
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        gadgetctl run ghcr.io/inspektor-gadget/gadget/profile_qdisc_latency:%IG_TAG% \
```
patrickpichler, line 272: Is there anything else i need to do in order to get prometheus metrics working? I tried running it via 
```bash
ig run profile_qdisc_latency:latest --verify-image=false --ms --annotate=qdisc_latency:metrics.collect=true --otel-metrics-name=qdisc_latency:qdisc-latency-metrics --otel-metrics-listen
```
but a curl to the metrics endpoint is not returning any results.

mayasingh17: @flyth do you have any ideas?

flyth: The name of the datasource seems to be `qdisc` - so the mapping using `--otel-metrics-name` should use that. Could you please try with `--otel-metrics-name=qdisc:qdisc-latency-metrics`? The rest looks good.

gadgets/profile_qdisc_latency/program.bpf.c
patrickpichler: Is it the wrong output for `3` though (I am no expert in this area though, so I could be totally wrong!)? From what I understand the buckets should be as follows
`0` is a broken value, that is why BCC is doing `+1`, as Brendan Gregg wanted to distinguish `0` and `1` (related [ticket](https://github.com/iovisor/bcc/issues/1649)). 
`[1] = Bucket 0`
`[2, 4[ = Bucket 1`
`[4, 8[ = Bucket 2`
`[8, 16[ = Bucket 3` 
and so on. 

I guess the problem comes from the fact that a `0` bucket is visualised in the output. To fix this, either the `0` bucket needs to be dropped, or we need to add `+1` to the result of the `log2l` output. 

Just from looking at the eBPF code, the issue is for sure also present in all other gadgets using the `log2l` function.

alban: We can have a valid 0 value because the delta precision is not infinite. The unit is either millisecond or microsecond. If we have 0.4ms or 0.4μs, it will rounded to the value 0.

I filed https://github.com/inspektor-gadget/inspektor-gadget/issues/3996. This can be fixed separately (it is not a blocker for this PR).

gadgets/profile_qdisc_latency/README.mdx
```diff
@@ -0,0 +1,351 @@
...
+          524288 -> 1048576    : 0        |                                        |
+         1048576 -> 2097152    : 0        |                                        |
+         2097152 -> 4194304    : 0        |                                        |
+         4194304 -> 8388608    : 0        |                                        |
+         8388608 -> 16777216   : 0        |                                        |
+        16777216 -> 33554432   : 0        |                                        |
+        33554432 -> 67108864   : 0        |                                        |
+        ```
+
+        Now to introduce some more latenyc, let's add a `netem` qdisc with some latency and jitter.
```
alban, line 184:
```suggestion
        Now to introduce some more latency, let's add a `netem` qdisc with some latency and jitter.
```

gadgets/profile_qdisc_latency/README.mdx
```diff
@@ -0,0 +1,351 @@
...
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ gadgetctl list
+        ID           NAME                                                      TAGS                                                      GADGET
+        3e68634c4c28 amazing_payne                                                                                                       ghcr.io/inspektor-gadget/gadget/profile_qdisc_latency:latest
+        ```
+
+        ```bash
+        $ gadgetctl/... delete 3e68634c4c28
```
alban, line 347:
```suggestion
        $ gadgetctl delete 3e68634c4c28
```

alban (commented):
Thanks for spending the time to write the documentation!

I added a few comments. I think it will be ready when they are handled.

Once this PR is merged, there could be a few follow-ups (not a blocker for this initial PR):
- [ ] Fix the log2l issue #3996
- [ ] Add tests for this new gadget

gadgets/profile_qdisc_latency/program.bpf.c
alban: The filter is not a blocker for this PR. I think when the other current comments are handled, I think it will be ready to be merged.

gadgets/profile_qdisc_latency/program.bpf.c
patrickpichler: I guess a potential fix could look something like the following
```c
if (delta > 0)
    slot = log2l(delta) + 1;
```

So `0` values would go into their own bucket and the rest is shifted by one, which should fix the histogram output.

alban:
Thanks for the update.

A symlink in the directory `docs/gadgets/` needs to be added.

gadgets/profile_qdisc_latency/README.mdx
```diff
@@ -0,0 +1,351 @@
...
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        gadgetctl run ghcr.io/inspektor-gadget/gadget/profile_qdisc_latency:%IG_TAG% \
+                    --annotate=qdisc_latency:metrics.collect=true \
+                    --otel-metrics-name=qdisc_latency:qdisc-latency-metrics \
```
alban, line 274: I have not tried it, but from the discussion above it seems to need this patch:

```suggestion
                    --otel-metrics-name=qdisc:qdisc-latency-metrics \
```

Unless I misunderstood the discussion.

gadgets/profile_qdisc_latency/README.mdx
```diff
@@ -0,0 +1,351 @@
...
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        Unless you configured the metrics listener to do differently, the
+        metrics will be available at `http://localhost:2224/metrics` on the
+        server side. And for the `profile_qdisc_latency` gadget we ran above, the
+        metrics will be available under the `qdisc-latency` scope:
```
alban, line 291:
```suggestion
        metrics will be available under the `qdisc-latency-metrics` scope:
```

patrickpichler:
@alban all comments have been addressed. I guess the build is failing, as the link will be added as part of this PR.

alban (approved):
LGTM, thank you!

mayasingh17:
[celebrate]     Maya Singh reacted to your message:
________________________________

alban:
The documentation website did not deploy for an unknown reason:

https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/13431170818/job/37523195271

> Run gh workflow run -R inspektor-gadget/website azure-static-web-apps-polite-mushroom-00d947803.yml
  gh workflow run -R inspektor-gadget/website azure-static-web-apps-polite-mushroom-00d947803.yml
  shell: /usr/bin/bash -e {0}
  env:
    GH_TOKEN: ***
could not create workflow dispatch event: HTTP 422: Workflow does not have 'workflow_dispatch' trigger (https://api.github.com/repos/inspektor-gadget/website/actions/workflows/22380062/dispatches)
Error: Process completed with exit code 1.

alban:
Artifact Hub failed to process the file:

> error getting package metadata (path: /tmp/artifact-hub1557883500/gadgets/profile_qdisc_latency): error validating package metadata file: 1 error occurred:
> * invalid metadata: description not provided

mauriciovasquezbernal:
> The documentation website did not deploy for an unknown reason:

The inspektor_gadget repo triggers a workflow on the website repo by using workflow_distach. However it was removed in https://github.com/inspektor-gadget/website/pull/151. I can investigate this tomorrow or early next week.

Issue #3643, opened 2024-10-29
TestTraceNetwork failing on containerd


burak-ok:
Since some time the `TestTraceNetwork` keeps failing often for `containerd`. For `cri-o` is was already disabled in https://github.com/inspektor-gadget/inspektor-gadget/issues/2358

```
2024-10-25T06:34:12.6699123Z         {"runtime":{},"k8s":{"owner":{}},"type":"normal","uid":0,"gid":0,"pktType":"OUTGOING","proto":"TCP","port":80,"dst":{"addr":"10.244.0.13","version":4}}
2024-10-25T06:34:12.6701766Z         {"runtime":{"runtimeName":"containerd","containerName":"nginx-pod","containerImageName":"docker.io/library/nginx:latest"},"k8s":{"namespace":"test-trace-network-5702393919195608251","podName":"nginx-pod","containerName":"nginx-pod","owner":{}},"type":"normal","comm":"nginx","uid":0,"gid":0,"pktType":"HOST","proto":"TCP","port":80,"dst":{"addr":"10.244.0.14","version":4}}
2024-10-25T06:34:12.6703275Z         expected:
2024-10-25T06:34:12.6705510Z         {"runtime":{"runtimeName":"containerd","containerName":"test-pod","containerImageName":"docker.io/library/busybox:latest"},"k8s":{"namespace":"test-trace-network-5702393919195608251","podName":"test-pod","containerName":"test-pod","owner":{}},"type":"normal","comm":"wget","uid":0,"gid":0,"pktType":"OUTGOING","proto":"TCP","port":80,"dst":{"addr":"10.244.0.13","version":4}}
```

The `test-pod` doesn't get enriched at all. This is the same issue I personally saw with some flaky runs in `cri-o`.

burak-ok:
Already commented about this in June in https://github.com/inspektor-gadget/inspektor-gadget/pull/3059#issuecomment-2188470452

burak-ok:
This gadget+test was removed in https://github.com/inspektor-gadget/inspektor-gadget/pull/4573

PR #3669, opened 2024-11-04
tests: Disable TestTraceNetwork on containerd

burak-ok:
This test is currently failing a lot of times and hinders us from merging and releasing without multiple attempts of re-runs


burak-ok:
See https://github.com/inspektor-gadget/inspektor-gadget/issues/3643 and https://inspektor-gadget.github.io/ig-test-reports/

alban (approved):

PR #3674, opened 2024-11-07
Use `ContainerPid` method in example

ninadingole:
# Use `ContainerPid` in example

There was a change done to access container PID and the field is no longer directly accessible from `Container` field. We need to access it through the new method `ContainerPid()`

## How to use

This change is done as a part of examples in logging so review would be straightforward

### DIFF
--- a/examples/container-collection/container-collection.go
+++ b/examples/container-collection/container-collection.go
@@ -33,10 +33,10 @@ func main() {
 		switch event.Type {
 		case containercollection.EventTypeAddContainer:
 			fmt.Printf("Container added: %q pid %d\n",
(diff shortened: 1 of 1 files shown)


alban (approved):

alban:
LGTM.

The CI (.github/workflows/inspektor-gadget.yml) tries to build some of the examples but not this one. Ideally, a test should be added for this example too.

alban:
The patch still looks good to me, but could you rebase your branch on main instead of having the merge commit in the PR? Thanks!

ninadingole:
@alban the branch is rebased

PR #3686, opened 2024-11-10
Readme: Update section to only show recent videos

blixtra:
I've added some recent videos that were missing and removed to old videos which don't really describe how IG works now. There are probably some others to add which we should do in follow up commits.

I think we should have a dedicated file with all/old talks but I don't think they need to be in the Readme.


mauriciovasquezbernal (approved):
To do later: Create a separated file with old presentations.

Issue #3675, opened 2024-11-07
run: `--annotate` replaces whole datasource configuration


mauriciovasquezbernal:
The `--annotate`  flag overrides all config file, leading to different issues when using this flag:

```bash
$ sudo ig run trace_open:latest --containername=foo --annotate open:foo=bar
RUNTIME.CONTAINERNAME   COMM                PID        TID FD         FLAGS_RAW    MODE… FNAME        MODE         FLAGS        ERROR
```

See how columns that should be hidden are shown. 

```bash 
$ sudo ig run trace_open:latest --containername=foo
RUNTIME.CONTAINERNAME          COMM                    PID        TID  FD FNAME                            MODE             ERROR
```

The seems to be around https://github.com/inspektor-gadget/inspektor-gadget/blob/2a4fce47ed011d3e6de0ad65594a23b90f775ecd/pkg/operators/oci-handler/oci.go#L287-L316

A possible solution is proposed in https://github.com/inspektor-gadget/inspektor-gadget/commit/a70bbcc5c97737da0c391e32c378526b9d2bc31e by @flyth

flyth:
Fixed by #3696

PR #3696, opened 2024-11-12
oci-handler: fix annotate param merge with prior config

flyth:
This fixes an issue that could occur when merging `--annotate` flags into an existing `gadget.yaml` viper config. With the previous method it could happen that viper would place the key in an ambiguous way so that the older values got lost.


mauriciovasquezbernal (approved):
Fixes #3675

Thanks for the fix.

PR #3699, opened 2024-11-13
Remove runcfanotify package

### LINKED ISSUE #1831: Deprecate and delete runc-fanotify in favour of container-hook

alban:
It has been replaced by the container-hook package.

runcfanotify only works when Inspektor Gadget runs in the host pid namespace. But this is not the case since June 2023.

This also removes the option "kubectl-gadget deploy --legacy-host-pid".

See also #1709

Fixes #1831


examples/runc-hook/README.md
flyth, line 1: Maybe rename `examples/runc-hook` to `examples/container-hook`?

flyth (approved):
LGTM from code inspection. Maybe just rename the example as well.

examples/runc-hook/README.md
alban: examples/container-hook already exists though... but it is less complete.
Should I delete examples/container-hook and then do the rename?

flyth: Ah - I'd choose to keep the more complete solution, then. But I'm fine with whatever you decide.

alban:
Examples updated. I also pushed to branch `citest/alban_delete_runc_fanotify`.

alban:
Tests with `citest/alban_delete_runc_fanotify` passed fine, after 1 retry:
https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/12029998050

alban:
I checked that the runcfanotify removal is fine for Kubescape. Kubescape does not use runcfanotify since December 2023, see:
- https://github.com/kubescape/node-agent/blob/v0.2.185/pkg/containerwatcher/v1/container_watcher_private.go#L117-L118
- https://github.com/kubescape/node-agent/pull/153 ("The container collection options have been updated to use containercollection.WithContainerFanotifyEbpf instead of containercollection.WithRuncFanotify")

Issue #2346, opened 2024-01-08
[RFE] [trace exec] memfd field


galofir-ms:
Add the `memfd` to specify whether the process was run on the filesystem or on a memfd

alban:
Reopen.

See also #1913 for some context.

alban:
Assigned to @dorser

(I tried to use the "Assignees" field on the issue but github does not let me, so I leave a comment instead)

dorser:
Let's try and assign to me again

PR #3706, opened 2024-11-15
execsnoop: add from_memfd field

dorser:
# Add `from_memfd` field to execsnoop
Resolves: https://github.com/inspektor-gadget/inspektor-gadget/issues/2346

## How to use

## Testing done

### DIFF
--- a/pkg/factory/setk8sdefaults_test.go
+++ b/pkg/factory/setk8sdefaults_test.go
@@ -1,0 +1,70 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


pkg/gadgets/trace/exec/tracer/tracer.go
```diff
@@ -49,6 +49,7 @@ type execsnoopEventAbbrev struct {
 	ArgsCount   int32
 	UpperLayer  bool
 	PupperLayer bool
+	FromMemfd   bool
 	_           [2]byte
```
mauriciovasquezbernal, line 53: This struct needs to be copied from pkg/gadgets/trace/exec/tracer/execsnoop_bpfel.go

```suggestion
	_           [1]byte
```

pkg/gadgets/trace/exec/tracer/tracer.go
```diff
@@ -69,6 +70,7 @@ type execsnoopWithLongPathsEventAbbrev struct {
...
```
mauriciovasquezbernal, line 74:
```suggestion
	_           [1]byte
```

mauriciovasquezbernal (commented):
How do you test it? I wasn't able to make it return from_memfd true while testing. Which code did you use to execute from memory?

PR #3713, opened 2024-11-19
operators/wasm: Add WASM function to create and close maps.

eiffel-fl:


pkg/operators/wasm/maps.go
```diff
@@ -213,3 +270,26 @@ func (i *wasmOperatorInstance) mapDelete(ctx context.Context, m wapi.Module, sta
...
+func (i *wasmOperatorInstance) mapClose(ctx context.Context, m wapi.Module, stack []uint64) {
+	mapHandle := wapi.DecodeU32(stack[0])
+
+	ebpfMap, ok := getHandle[*ebpf.Map](i, mapHandle)
+	if !ok {
+		stack[0] = 1
+		return
+	}
+
+	err := ebpfMap.Close()
```
alban, line 288: * I understand closing a map created by `newMap`, but how is it safe to close a map from `getMap`? For example, what if the wasm module closes the map "events" or the map "sockets" from the socket enricher? It seems like it will close the map without checking if it can be used internally by IG.
* How do you prevent a wasm program from calling Lookup after calling Close? It seems IG will call the bpf syscall with fd -1. Shouldn't we detect that?

pkg/operators/wasm/maps.go
```diff
@@ -56,6 +67,52 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
...
+		[]wapi.ValueType{
+			wapi.ValueTypeI32, // Map
+		},
+		[]wapi.ValueType{wapi.ValueTypeI32}, // Error
+	)
+}
+
+// newMap creates a new map.
+// Params:
+// - stack[0] is the map type
```
alban, line 81:
```suggestion
// - stack[0] is the map name
```

pkg/operators/wasm/testdata/map/program.go
```diff
@@ -128,6 +129,48 @@ func gadgetStart() int {
...
+		KeySize:    uint32(4),
+		ValueSize:  uint32(4),
+		MaxEntries: 1,
+	}
+
+	newMap, err := api.NewMap(mapSpec)
+	if err != nil {
+		api.Errorf("creating map %s", mapSpec.Name)
+	}
+	defer newMap.Close()
```
alban, line 144: Could you add tests:
- Running Close several times
- Trying Lookup/Put after Close and checking the error code

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -375,6 +375,20 @@ Return value:
 
 ### eBPF Maps
 
+#### `newMap(name string, typ uint32, keySize uint32, valueSize uint32, maxEntries uint32) uint32`
+
+Create a new eBPF map.
```
alban, line 380: What is the main use case for this? Will you need to add types (BTF) to the maps, or will we only need untyped maps?

pkg/operators/wasm/maps.go
alban: > Why do you mean by prevent?

I mean that IG should detect when mapClose has been called on a handle. Then, further calls to mapClose would return an error without relying on the kernel returning EBADF.

> Otherwise, you can call the BPF syscall with fd -1 and you will get an error as reported by the kernel.

True, but this looks weird to me. It is exposing low-level details to the wasm module instead of wrapping it.

pkg/operators/wasm/maps.go
```diff
@@ -56,6 +68,61 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
...
+	exportFunction(env, "mapClose", i.mapClose,
+		[]wapi.ValueType{
+			wapi.ValueTypeI32, // Map
+		},
+		[]wapi.ValueType{wapi.ValueTypeI32}, // Error
+	)
+}
+
+var createdMapMutex sync.RWMutex
+var createdMap = make(map[uint32]struct{})
```
alban, line 81: It should be part of `wasmOperatorInstance` to avoid collisions when running several gadgets at the same time.

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -420,3 +434,13 @@ Parameters:
 
 Return value:
 - 0 in case of success, 1 otherwise.
+
+#### `mapClose(m uint32) uint32`
+
+Close the map created by `newMap()`.
```
alban, line 440:
```suggestion
Close the map created by `newMap()`. The map handle is released and can no longer be used.
```

I wonder, in that case, if it should be called `mapRelease`, like other functions `dataSourceRelease`, `dataArrayRelease`...

pkg/operators/wasm/maps.go
```diff
@@ -213,3 +276,43 @@ func (i *wasmOperatorInstance) mapDelete(ctx context.Context, m wapi.Module, sta
...
+	if !ok {
+		i.logger.Warnf("mapRelease: map %d was not created by newMap() or was already closed", mapHandle)
+		stack[0] = 1
+		return
+	}
+
+	ebpfMap, ok := getHandle[*ebpf.Map](i, mapHandle)
+	if !ok {
+		stack[0] = 1
+		return
```
alban, line 301: Shouldn't we add a warning with `i.logger.Warnf` here?

pkg/operators/wasm/maps.go
```diff
@@ -213,3 +276,43 @@ func (i *wasmOperatorInstance) mapDelete(ctx context.Context, m wapi.Module, sta
...
+// - 0 on success, 1 on error
+func (i *wasmOperatorInstance) mapRelease(ctx context.Context, m wapi.Module, stack []uint64) {
+	mapHandle := wapi.DecodeU32(stack[0])
+
+	i.createdMapMutex.RLock()
+	_, ok := i.createdMap[mapHandle]
+	i.createdMapMutex.RUnlock()
+
+	if !ok {
+		i.logger.Warnf("mapRelease: map %d was not created by newMap() or was already closed", mapHandle)
```
alban, line 293: Shouldn't this test be done after `getHandle[*ebpf.Map]()`? So if the wasm code calls mapRelease() on a handle of different type, it does not say "map %d".

pkg/operators/wasm/maps.go
```diff
@@ -213,3 +276,43 @@ func (i *wasmOperatorInstance) mapDelete(ctx context.Context, m wapi.Module, sta
...
+	ebpfMap, ok := getHandle[*ebpf.Map](i, mapHandle)
+	if !ok {
+		stack[0] = 1
+		return
+	}
+
+	err := ebpfMap.Close()
+	if err != nil {
+		i.logger.Warnf("mapRelease: closing map: %v", err)
+		stack[0] = 1
```
alban, line 307: I wonder if `delHandle` and `i.createdMap` should still be cleaned up when the `close()` syscall returns an error. This would prevent a wasm module from calling `mapRelease` twice, and then `close` twice on the same file descriptor. And the manpage of close says it is a bug to call close twice in case of error, because the fd can be closed even when returning an error, so the fd number can be reallocated to something else.

Current versions of Linux seems to never return an error when calling `close()` on fds of type bpf map, so it is not a problem in practice. But if that changes, that would be a problem.

pkg/operators/wasm/testdata/badguest/program.go
```diff
@@ -112,6 +115,9 @@ func mapUpdate(m uint32, keyptr uint64, valueptr uint64, flags uint64) uint32
 //go:wasmimport env mapDelete
 func mapDelete(m uint32, keyptr uint64) uint32
 
+//go:wasmimport env mapClose
+func mapRelease(m uint32) uint32
```
alban, line 119: The name does not match the comment: mapClose vs mapRelease.

alban (approved):
I'm adding 4 comments to address, but LGTM after that.
Thanks.

docs/gadget-devel/gadget-wasm-api-raw.md
mauriciovasquezbernal: I think it should be called `mapClose`. `dataSourceRelease` and `dataArrayRelease` are related to the [datasource.Release](https://github.com/inspektor-gadget/inspektor-gadget/blob/13440171913d6d48f22e2cc034eb7a330e13a990/pkg/datasource/datasource.go#L142-L143)

PR #3714, opened 2024-11-21
operators/wasm: Add functions to interact with perf buffer.

eiffel-fl:
TODO: Documentation once we agree on the API.


pkg/operators/wasm/testdata/perf/program.go
```diff
@@ -0,0 +1,102 @@
...
+	type event struct {
+		a      int32
+		b      int32
+		c      int8
+		unused [247]int8
+	}
+
+	// Generate some events to fill the buffer.
+	for range 21 {
+		syscall.Open("/tmp/test-perf-array", 0, 0)
```
alban, line 41: What does `syscall.Open` do in a wasm module?

pkg/operators/wasm/perf.go
```diff
@@ -0,0 +1,242 @@
...
+
+	stack[0] = 0
+}
+
+// perfReaderSetDeadline() set the perf reader deadline
+// Params:
+// - stack[0]: Perf reader handle
+// - stack[1]: Time in nanoseconds
+// Return value:
+// - 0 on success, 1 on error
```
alban, line 160: Do we need to expose that to the wasm module?

AFAIK, so far, the functions in the wasm module don't sleep. With `perfReaderRead` and `perfReaderSetDeadline`, you are changing that. This has the potential to introduce tricky bugs. For example, if a gadget sets up a long deadline, and the user is trying to stop the gadget. If the wasm runner is blocked, IG would not be able to stop the gadget.

It seems to me that the event loop should be implemented in IG core instead of in the wasm guest. And for traceloop, we don't need blocking reads anyway, so it seems we are introducing complex API without a use case for it.

PR #3717, opened 2024-11-21
gadgets/deadlock: add param for `pid` based filtering

Snehil-Shah:
# Adds `--pid` flag for filtering

Adds `--pid` flag for targeted tracing based on process ID (tgid).

## How to use

Just run the gadget with the `--pid	` flag set:

```shell
sudo ig run deadlock:latest --pid=53240
```

## Testing done

- Ran my test container that has a binary that emits deadlock events.
- Ran the binary inside the test container and found it's pid using `trace_exec`.
- Ran the deadlock gadget with `--pid` flag set.
- Confirmed that the events were showing up. Ran the binary again (this time a diff pid), and the events don't show up as expected.


Snehil-Shah:
Some questions:

- Currently, we are sending PID after every process ends, to userspace. Can this clog things up? Should we maintain another map to track only the PIDs of interest?
- We could optimize memory further by also cleaning up the `edges` bpf map. This will also allow the gadget to handle cases when PID is reused by a new process. But currently, the key of the map has `pid`, `mutex1`, and `mutex2`. To clean all entries of a PID, we might have to traverse every entry to find what to clean up. Is there a better solution?

alban (approved):
LGTM, thanks!

PR #3718, opened 2024-11-22
port advise_seccomp to image-based gadget

flyth:
Currently WIP

This implements a couple of things:
* adds a PreStop() lifecycle hook to do things "last minute"
* fixes lifecycle of WASM instances
* adds "flush-on-stop" annotation for map fetching in PreStop()
* adds an option to create initial map values for datasources (this is used as scratch map because the syscall bitmap is too large to fit on the stack)

How it currently works is:
* WASM module creates and self-subscribes to a new DataSource which is just used to emit every possible syscall IDs; this will be enriched by the formatter operator to contain the actual syscall names, with which the WASM module will create a lookup map (this is somewhat hacky but works well)
* WASM creates a new text DataSource for the actual advise output
* Once the gadget is stopped, it will flush the map contents (due to the annotation) of the eBPF part and convert it to a proper output

Currently needs to be run with some map flags:
```
ig run advise_seccomp --verify-image=false --map-fetch-count=0 --map-fetch-interval=0
```

Output looks like this:

```
// otel2-grafana-1
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    "epoll_pwait",
    "futex",
    "nanosleep",
    "read",
    "sched_yield",
    "write"
  ]
}
// otel2-backend-1
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    "accept4",
    "close",
    "connect",
    "epoll_ctl",
    "epoll_pwait",
    "futex",
    "getpeername",
    "getsockname",
    "getsockopt",
    "nanosleep",
    "read",
    "sched_yield",
    "setsockopt",
    "socket",
    "write"
  ]
}
// otel2-tempo-1
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    "epoll_pwait",
    "futex",
    "nanosleep",
    "read",
    "write"
  ]
}
```

This will only work in interactive mode, as we don't have a "stopped" state in detached mode for now.

Part of #3003

### DIFF
--- a/gadgets/advise_seccomp/go/program.go
+++ b/gadgets/advise_seccomp/go/program.go
@@ -1,0 +1,179 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"slices"
+	"strings"
+
+	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
+)
+
+var (
+	sc             api.DataSource
+	syscallds      api.DataSource
+	textds         api.DataSource
+	textField      api.Field
+	syscallMapping map[uint64]string
+)
+
+//export gadgetPreStart
+func gadgetPreStart() int {
+	syscallMapping = make(map[uint64]string)
+
+	scStringField, err := sc.GetField("syscall")
+	if err != nil {
+		api.Warnf("getting syscall text field (enricher inactive?)")
+		return 1
+	}
+
+	scIntField, err := sc.GetField("syscall_raw")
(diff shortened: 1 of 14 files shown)


gadgets/advise_seccomp/go/program.go
```diff
@@ -0,0 +1,179 @@
...
+					syscallStrings = append(syscallStrings, syscallName)
+				}
+			}
+
+			slices.Sort(syscallStrings)
+
+			if containerName == "" {
+				switch mntnsid {
+				case 1:
+					containerName = "host"
```
mauriciovasquezbernal, line 115: the mount namespace id for the host isn't zero.  We have some logic to get it on 

https://github.com/inspektor-gadget/inspektor-gadget/blob/8912ea06de042187f4a125402694dd17b2ed7edd/pkg/operators/localmanager/localmanager.go#L245-L249

however it isn't exposed to wasm.

mauriciovasquezbernal (commented):
I like it a lot. Looking forward so see it built on top of #3756 to avoid the datasources hack. I also think the annotation initialization thing isn't needed if we store the initial syscall value on a different map.

mauriciovasquezbernal:
I'm working on this one now, will update it shortly.

mauriciovasquezbernal:
Opened a new PR on #4089.

mauriciovasquezbernal:
Closing in favor of #4089

PR #3725, opened 2024-11-26
operators/wasm: Add functions to interact with perf buffer.

eiffel-fl:
TODO: Add documentation once agreed on the API.
Previous PR #3714 closed by mistake with merge target.


mauriciovasquezbernal (commented):
I'm a bit concerned about having those too granular functions. What about having a single 
`perfReaderOp(op, arg1, arg2, arg3...)` function that is multiplexed. (similar to the bpf() syscall on the kernel. I think that would be easier to implement, maintain and extend in the future.

mauriciovasquezbernal:
Let's ask other folks opinions on this one: @alban @flyth @blanquicet

pkg/operators/wasm/perf.go
```diff
@@ -0,0 +1,242 @@
...
+	perfReaderHandle := wapi.DecodeU32(stack[0])
+	addrBufPtr := wapi.DecodeU32(stack[1])
+
+	perfReader, ok := getHandle[*perf.Reader](i, perfReaderHandle)
+	if !ok {
+		stack[0] = 1
+		return
+	}
+
+	record, err := perfReader.Read()
```
alban, line 192: I still have concerns about exposing blocking functions to the wasm module.

Even if traceloop uses `perfReaderRead` in a non-blocking way (the ring buffer is configured as backward and overwriteable. AFAIK, that makes it non-blocking), this new API enables blocking methods.

Instead of mirroring the cilium/ebpf/perf API into wasm, I would prefer to have a higher-level API without the blocking behaviour. In practice, gadgets/traceloop/go/program.go would not call Pause+SetDeadline+Read+Resume but a single call to the wasm host doing the 4 function calls in one step. In this way, the wasm module would not be able to block.

Previous discussion in https://github.com/inspektor-gadget/inspektor-gadget/pull/3714#discussion_r1856381535

wasmapi/go/perf.go
```diff
@@ -0,0 +1,109 @@
...
+	return nil
+}
+
+func (p PerfReader) Read() ([]byte, error) {
+	var buf bufPtr
+
+	ret := perfReaderRead(uint32(p), uint32(uintptr(unsafe.Pointer(&buf))))
+	switch ret {
+	case 0:
+		return buf.bytes(), nil
```
mauriciovasquezbernal, line 92: it's missing a call to `buf.free()` before returning.

pkg/operators/wasm/perf.go
```diff
@@ -0,0 +1,249 @@
...
+		return
+	}
+
+	err := perfReader.Close()
+	if err != nil {
+		i.logger.Warnf("perfReaderClose: resuming perf reader: %v", err)
+		stack[0] = 1
+		return
+	}
+
```
mauriciovasquezbernal, line 247: release the handle? ref #3702

pkg/operators/wasm/testdata/perf/program.go
```diff
@@ -0,0 +1,102 @@
...
+		api.Errorf("creating perf reader")
+		return 1
+	}
+	defer perfReader.Close()
+
+	_, err = perfReader.Read()
+	if err == nil {
+		api.Errorf("perf reader must be paused before reading")
+		return 1
+	}
```
mauriciovasquezbernal, line 56: Is this true for all perf event array buffers? I don't see we pause it on the different tracers we have.

pkg/operators/wasm/testdata/perf/program.go
```diff
@@ -0,0 +1,102 @@
...
+
+	err = perfReader.SetDeadline(api.Now())
+	if err != nil {
+		api.Errorf("setting perf reader deadline to now")
+		return 1
+	}
+
+	buf, err := perfReader.Read()
+	if err != nil {
+		api.Errorf("reading perf record")
```
mauriciovasquezbernal, line 78: missing return 1? Or do you want to rely on the nil check below?

pkg/operators/wasm/wasm_test.go
```diff
@@ -613,3 +614,48 @@ func TestSyscall(t *testing.T) {
...
+	t.Cleanup(func() { runtime.Close() })
+
+	stop := make(chan bool)
+	go func() {
+		for {
+			select {
+				case <-stop:
+					return
+				default:
+					fd, err := syscall.Creat("/tmp/test-perf-array", 0)
```
mauriciovasquezbernal, line 648: Is this for generating an event that triggers the ebpf program to submit info on the perf event array? If that's the case, a comment will help to clarify it.  
Also, should it use a timer to avoid creating a lot of events?

mauriciovasquezbernal (commented):
It looks good to me. I think we can avoid the deadline (blocking read) issue by always passing a deadline on the past to the perf event reader and hide this from wasm users. If we want that functionality, we can create it later on.

mauriciovasquezbernal (commented):
Implementation LGTM! I haven't tested it but the unit test on this PR and the WIP traceloop should be enough. 

Looking forward for the documentation to approve this. 

Thanks!

mauriciovasquezbernal (approved):
Two quick comments on the docs, but LGTM already. Thanks for working on this!

Issue #3582, opened 2024-10-14
[RFE] Support ARGSIZE from user input in trace_exec


michaelkatch:
## Current situation

Currently trace_exec is truncating arguments with length exceeding 128 bytes by default
![image](https://github.com/user-attachments/assets/61ede3eb-ebe6-4d64-9aea-2baf195bdf21)

## Impact
Some process executions are passed with longer parameters, specifically the first parameters (i.e. base64 encoded payloads) and we are unable to see them correctly

## Ideal future situation
I suggest accepting a parameter from the user indicating the maximum argument size we are willing to trace 
Ideally this argument will impact all of the arguments (The default maximum is 60) but increasing the length even for the first 10 arguments is reasonable

alban:
Related issues:
- https://github.com/inspektor-gadget/inspektor-gadget/issues/2993
- https://github.com/inspektor-gadget/inspektor-gadget/issues/3491

I am not sure we can do it fully dynamic. Do your base64 payloads have a maximum size?

I see the maximum on Linux is 128KiB:
https://github.com/torvalds/linux/blob/v6.8/include/uapi/linux/binfmts.h#L9-L16

michaelkatch:
@alban missed your comment , sorry about that.
After we discussed today, I suggested increasing the argument size to 256 bytes for the 5 arguments (that should be enough for us) and hopefully not take too much space

alban:
tl;dr: we can even increase to 256 bytes for 20 arguments **and** reduce the size of the buffer.

---

There are 3 ebpf modules that trace exec arguments:
1. builtin trace-exec gadget
2. image-based trace_exec gadget
3. container-hook

Those 3 modules support up either 128 or 256 bytes for each argument:

```
$ git grep '#define ARGSIZE'
gadgets/trace_exec/program.bpf.c:18:#define ARGSIZE 128
pkg/container-hook/bpf/execruntime.h:5:#define ARGSIZE 256
pkg/gadgets/trace/exec/tracer/bpf/execsnoop.h:5:#define ARGSIZE 128
```

The container-hook package switched from 128 to 256 bytes in commit 99c369ef3554879053a669fb64b9173c3e01648b in order to support some setups with long paths (Docker on WSL2).

[...]

Issue #3491, opened 2024-09-16
gadgets: Optimize support for optional big fields


mauriciovasquezbernal:
Some Gadgets support optional big fields, like [cwd](https://github.com/inspektor-gadget/inspektor-gadget/blob/2c89326ed126b5af4c6588691064cc5297ea89bb/gadgets/trace_exec/program.bpf.c#L46) on the trace_exec gadget which is 4096 bytes long. The current support for these kind of fields is sub-optimal as we always send the full event when the field is disabled, it means, for each event we transfer a lot of bytes that aren't used for anything. 

The scope of this issue is to implement a way to support these optional big fields in an optimal way.

### type-length-value (TLV)

The [TLV encoding](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value) is very common in networking applications. We could support encode the optional fields by using this:

```c 
struct event {
	// all fixed fields
	gadget_timestamp timestamp_raw;
	gadget_mntns_id mntns_id;
};

// Add tlv fields
// TODO: support type: string and byte
GADGET_TLV_FIELD(event, cwd, 1) // 1 is the id
GADGET_TLV_FIELD(event, args, 2) // 2 is the id

int foo() {
	int tlv_size = 50; // make some calculations here for this
	int total_size = sizeof(struct event) + tlv_size;

	struct event *event = gadget_allocate_event(total_size);

	char *cwd = get_cwd();
	gadget_set_encode_tlv(event, cwd, strlen(cwd), 1);

	char *args = get_args();
	gadget_set_encode_tlv(event, args, strlen(args), 2);

	gadget_submit_event(event, total_size);

	return 0;
}
```

### Multiple events 

Another option is send optional fields in different "packets" from the eBPF program:

```c
struct event {
	// all fixed fields
	gadget_timestamp timestamp_raw;
	gadget_mntns_id mntns_id;
};

int foo() {
	// Allocate and send event
	struct event *event = gadget_allocate_event(total_size);
	gadget_submit_event(event, sizeof (*event));

	// send optional fields
	char *cwd = get_cwd();
	gadget_submit_event(cwd, strlen(cwd));

	char *args = get_args();
	gadget_submit_event(args, strlen(args));

	return 0;
}
```

I have the gut feeling that the TLV idea is easier to implement; however I wanted to show both of them for the discussion.

alban:
[...]

PR #3726, opened 2024-11-26
exec args: support up to 256 bytes per arg

alban:
There are 3 ebpf modules that trace exec arguments:
1. builtin trace-exec gadget
2. image-based trace_exec gadget
3. container-hook

Those 3 modules support up either 128 or 256 bytes for each argument:

$ git grep '#define ARGSIZE'
gadgets/trace_exec/program.bpf.c:18:#define ARGSIZE 128 pkg/container-hook/bpf/execruntime.h:5:#define ARGSIZE 256 pkg/gadgets/trace/exec/tracer/bpf/execsnoop.h:5:#define ARGSIZE 128

The container-hook package switched from 128 to 256 bytes in order to support some setups with long paths (Docker on WSL2). See https://github.com/inspektor-gadget/inspektor-gadget/pull/1767.

The 3 modules support up to 20 arguments. There is some left over code (coming from [bcc's execsnoop](https://github.com/iovisor/bcc/blob/8d85dcfac86bb7402a20bea5ceba373e5e019b6c/libbpf-tools/execsnoop.c#L75)) aiming to support 60 arguments but it has not been implemented in Inspektor Gadget. Nonetheless, the struct event is needlessly big enough to support 60 arguments.

This patch changes the argument size to 256 bytes for the 20 arguments we support today. This reduce the size of the args field in the struct event for the 20 arguments to `256*20 = 5 KiB` (instead of `128*60 = 7.5 KiB`).

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/3582

cc @michaelkatch


alban:
I tested that it captures 256 bytes correctly:
```bash
#!/bin/bash
for i in {250..259} ; do
  arg=$(head -c $i < /dev/zero | tr '\0' '/')
  echo $i
  cat $arg 2>/dev/null
done
```

It gives the following output:
```bash
$ sudo ./ig trace exec -o json --host -F comm:cat|jq --unbuffered -r '.args[1]'|awk '{ print length }'
250
251
252
253
254
255
255
255
255
255
```

This seems to correctly limit each argument to 256 bytes.

Also, I tested that it can returns 20 arguments:

```bash
arg=$(head -c 300 < /dev/zero | tr '\0' '/')
cat $(for i in {1..25} ; do echo $i$arg ; done) 2>/dev/null
```

It seems to work fine.

burak-ok (approved):
LGTM.

Either 
1. https://github.com/inspektor-gadget/inspektor-gadget/issues/3491 
2. Users being able to set the length of fields at gadget start

should be implemented

alban:
Thanks. I rebased and used `make ebpf-objects` to fix the conflicts. I'll merge once the CI completes.

PR #3728, opened 2024-11-28
operators/wasm: Add FD() to get eBPF map file descriptor.

eiffel-fl:


alban (changes requested):
I think the fd number is internal to IG core and it should not be exposed to the wasm module.

I guess you need it for maps of maps. In that case, I suggest to implement `mapUpdateWithHandle` (in addition to `mapUpdate`).

```go
func mapUpdate(m uint32, keyptr uint64, valueptr uint64, flags uint64) uint32
func mapUpdateWithHandle(m uint32, keyptr uint64, valuehandle uint32, flags uint64) uint32
```

In that way, the wasm guest would not be able to give a fd number that it does not own. And the wasm host can convert the handle to a fd and check that the handle is still valid.

alban:
> As a middle ground, we can have FD() (int, error) which returns the FD only on maps created with newMap() and an error for those generated through getMap(). What do you think?

I still think it is wrong for the wasm module to have to use file descriptors. I would still like the wasm module not to be able to call `mapUpdate` with fds that might belong to other gadgets (maybe generated by getMap but from other gadgets).

What do other people think?

mauriciovasquezbernal:
> Giving the handle as value with a flag indicate we should take the FD on the handle?

Yes. 

> Otherwise, cannot we, in all cases try to cast the value to a handle and if it works we take the FD?

I'd prefer to be more explicit on this. I think there are cases where we can't automatically distinguish  a handle, for instance, the user could try to update a map with the integer 42, but it can happen that this integer corresponds to a map handle too.

alban:
> > > Giving the handle as value with a flag indicate we should take the FD on the handle?
> > 
> > Yes.
> 
> I like this idea, @alban do you agree?

Yes, that sounds good.

Issue #3627, opened 2024-10-24
gadgets/fsnotify: Missing tests


mauriciovasquezbernal:
The [fsnotify](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/fsnotify) gadget doesn't have any tests implemented. 

Edit 1:
Unit tests were implemented on #3729

burak-ok:
Since unit tests are implemented, should this be closed or renamed to `fsnotify` missing integration test?

mauriciovasquezbernal:
It was about having at least any tests, so I'll close it. We can create new issues if we decide each gadget should have an integration test available.

PR #3729, opened 2024-11-28
tests: added unit test to fsnotify gadget

4rivappa:
Unit test to validate `inotify` event capture through `fsnotify` gadget.

Related to #3627

### About test

The test simulates event generation based on the example in the gadget’s [documentation](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/fsnotify/README.mdx). 
It uses the [fsnotify](https://github.com/fsnotify/fsnotify) Go package to watch the `/tmp` path and capture events.

Dependencies: [fsnotify package](https://github.com/fsnotify/fsnotify)

### Testing done

Tested with `gadgets-unittest` step of inspektor-gadget.yml workflow. [[successful run](https://github.com/4rivappa/inspektor-gadget/actions/runs/12074081422)]

[21st Dec 2024] Updated PR test [successful run](https://github.com/4rivappa/inspektor-gadget/actions/runs/12443836290/job/34743485124)

### DIFF
--- a/gadgets/fsnotify/test/unit/fsnotify_test.go
+++ b/gadgets/fsnotify/test/unit/fsnotify_test.go
@@ -1,0 +1,227 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/gadgetrunner"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+)
+
+type Process struct {
+	PPid  uint32 `json:"ppid"`
+	Pid   uint32 `json:"pid"`
+	Tid   uint32 `json:"tid"`
+	Comm  string `json:"comm"`
+	PComm string `json:"pcomm"`
+}
+
+type EventDetails struct {
+	FileName string
+	Ino      uint32
+	InoDir   uint32
+}
+
+type ExpectedFsnotifyEvent struct {
+	Timestamp string `json:"timestamp"`
+
+	Type string `json:"type"`
+
+	TraceeProc Process `json:"tracee_proc"`
+	TracerProc Process `json:"tracer_proc"`
+
+	TraceeMntnsId uint64 `json:"tracee_mntns_id"`
+	TracerMntnsId uint64 `json:"tracer_mntns_id"`
+
+	TraceeUId uint32 `json:"tracee_uid"`
+	TraceeGId uint32 `json:"tracee_gid"`
+	TracerUId uint32 `json:"tracer_uid"`
+	TracerGId uint32 `json:"tracer_gid"`
+
+	Prio uint32 `json:"prio"`
+
+	FaMask string `json:"fa_mask"`
+	IMask  string `json:"i_mask"`
+
+	FaType     string `json:"fa_type"`
+	FaPId      uint32 `json:"fa_pid"`
+	FaFlags    uint32 `json:"fa_flags"`
+	FaFFlags   uint32 `json:"fa_f_flags"`
+	FaResponse string `json:"fa_response"`
+
+	IWd     int32  `json:"i_wd"`
+	ICookie uint32 `json:"i_cookie"`
+	IIno    uint32 `json:"i_ino"`
+	IInoDir uint32 `json:"i_ino_dir"`
+
+	Name string `json:"name"`
+}
+
+type testDef struct {
+	runnerConfig  *utilstest.RunnerConfig
+	generateEvent func() (EventDetails, error)
+	validateEvent func(t *testing.T, info *utilstest.RunnerInfo, eventDetails EventDetails, events []ExpectedFsnotifyEvent)
+}
+
+func TestFsnotifyGadget(t *testing.T) {
+	gadgettesting.MinimumKernelVersion(t, "5.10")
+	gadgettesting.InitUnitTest(t)
+	runnerConfig := &utilstest.RunnerConfig{}
+
+	testCases := map[string]testDef{
+		"captures_inotify_event": {
+			runnerConfig:  runnerConfig,
+			generateEvent: generateEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, eventDetails EventDetails, events []ExpectedFsnotifyEvent) {
+				utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, pid int) *ExpectedFsnotifyEvent {
+					return &ExpectedFsnotifyEvent{
+						Timestamp: utils.NormalizedStr,
+
+						Type:  "inotify",
+						IMask: "IN_MODIFY",
+
+						TraceeMntnsId: info.MountNsID,
+						TracerMntnsId: utils.NormalizedInt,
+
+						FaType:     utils.NormalizedStr,
+						FaResponse: utils.NormalizedStr,
+
+						IWd:     utils.NormalizedInt,
+						IIno:    eventDetails.Ino,
+						IInoDir: eventDetails.InoDir,
+
+						Name: eventDetails.FileName,
+					}
+				})(t, info, 0, events)
+			},
+		},
+	}
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			var eventDetails EventDetails
+			runner := utilstest.NewRunnerWithTest(t, testCase.runnerConfig)
+
+			normalizeEvent := func(event *ExpectedFsnotifyEvent) {
+				utils.NormalizeString(&event.Timestamp)
+				utils.NormalizeInt(&event.TracerMntnsId)
+
+				utils.NormalizeString(&event.FaType)
+				utils.NormalizeString(&event.FaResponse)
+			}
+			onGadgetRun := func(gadgetCtx operators.GadgetContext) error {
+				utilstest.RunWithRunner(t, runner, func() error {
+					var err error
+					eventDetails, err = testCase.generateEvent()
+					if err != nil {
+						return err
+					}
+					return nil
+				})
+				return nil
+			}
+			opts := gadgetrunner.GadgetRunnerOpts[ExpectedFsnotifyEvent]{
+				Image:          "fsnotify",
+				Timeout:        5 * time.Second,
+				OnGadgetRun:    onGadgetRun,
+				NormalizeEvent: normalizeEvent,
+			}
+
+			gadgetRunner := gadgetrunner.NewGadgetRunner(t, opts)
+
+			gadgetRunner.RunGadget()
+
+			testCase.validateEvent(t, runner.Info, eventDetails, gadgetRunner.CapturedEvents)
+		})
+	}
+}
+
+func generateEvent() (EventDetails, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return EventDetails{}, err
+	}
+	defer watcher.Close()
+
+	err = watcher.Add(os.TempDir())
+	if err != nil {
+		return EventDetails{}, err
+	}
+
+	// event 1
+	newFile, err := os.CreateTemp(os.TempDir(), "test-*.txt")
+	if err != nil {
+		return EventDetails{}, err
+	}
+	defer newFile.Close()
+
+	// event 2
+	_, err = newFile.WriteString("Hello, fsnotify!")
+	if err != nil {
+		return EventDetails{}, err
+	}
+
+	// Receiving events.
+forLoop:
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return EventDetails{}, errors.New("channel closed")
+			}
+			if event.Has(fsnotify.Write) {
+				if event.Name != newFile.Name() {
+					return EventDetails{}, fmt.Errorf("watcher: unexpected event: %q, expected %q", event.Name, newFile.Name())
+				}
+				break forLoop
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return EventDetails{}, errors.New("channel closed")
+			}
+			return EventDetails{}, fmt.Errorf("watcher: %w", err)
+		}
+	}
+
+	// Get inode values of test file and its parent directory
+	fileInode, err := utils.GetInode(newFile.Name())
+	if err != nil {
+		return EventDetails{}, err
+	}
+	dirInode, err := utils.GetInode(path.Dir(newFile.Name()))
+	if err != nil {
+		return EventDetails{}, err
+	}
+
+	fileName := path.Base(newFile.Name())
+	eventDetails := EventDetails{
+		FileName: fileName,
+		Ino:      uint32(fileInode),
+		InoDir:   uint32(dirInode),
+	}
+	return eventDetails, nil
+}
--- a/pkg/testing/utils/utils.go
+++ b/pkg/testing/utils/utils.go
@@ -18,8 +18,10 @@ import (
 	"bytes"
 	"fmt"
 	"math/rand"
+	"os"
 	"os/exec"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
@@ -213,6 +215,21 @@ func GetContainerRuntime(t *testing.T) string {
 	parts := strings.Split(ret, ":")
 	require.GreaterOrEqual(t, len(parts), 1, "unexpected container runtime version")
 	return parts[0]
+}
+
+// GetInode extracts the inode of a given path
+func GetInode(path string) (uint64, error) {
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return 0, fmt.Errorf("attempting to stat path %s: %w", path, err)
+	}
+
+	// Extract the inode value from the system information
+	sysInfo := fileInfo.Sys()
+	if stat, ok := sysInfo.(*syscall.Stat_t); ok {
+		return uint64(stat.Ino), nil
+	}
+	return 0, fmt.Errorf("encountering issues when asserting system info as *syscall.Stat_t for path: %s", path)
 }
 
 // GenerateTestNamespaceName returns a string which can be used as unique


gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+)
+
+type ExpectedFsnotifyEvent struct {
+	Timestamp string `json:"timestamp"`
+
+	Type string `json:"type"`
+
+	TraceeProc ebpftypes.Process `json:"tracee_proc"`
+	TracerProc ebpftypes.Process `json:"tracer_proc"`
```
mauriciovasquezbernal, line 38: The process struct used in fsnotify is a bit different from `ebpftypes.Process`. I think it's better to define it here instead of using `ebpftypes.Process`

gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,185 @@
...
+	testCases := map[string]testDef{
+		"captures_inotify_event": {
+			runnerConfig:  runnerConfig,
+			generateEvent: generateEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, filename string, events []ExpectedFsnotifyEvent) {
+				utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, pid int) *ExpectedFsnotifyEvent {
+					return &ExpectedFsnotifyEvent{
+						Type: "inotify",
+
+						IMask: 134217732, // 134217732 = 0x08000004 = FS_ATTRIB | FS_EVENT_ON_CHILD
```
mauriciovasquezbernal, line 86:
```suggestion
						IMask: 0x08000004, // FS_ATTRIB | FS_EVENT_ON_CHILD
```

gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+				utils.NormalizeProc(&event.TraceeProc)
+				utils.NormalizeProc(&event.TracerProc)
+				utils.NormalizeInt(&event.TraceeMntnsId)
+				utils.NormalizeInt(&event.TracerMntnsId)
+
+				utils.NormalizeInt(&event.TraceeUId)
+				utils.NormalizeInt(&event.TraceeGId)
+				utils.NormalizeInt(&event.TracerUId)
+				utils.NormalizeInt(&event.TracerGId)
```
mauriciovasquezbernal, line 123: Can't we test then too? The runner allows to use a different uid / gid. See fields on utilstest.RunnerConfig.

gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,185 @@
...
+	err = watcher.Add("/tmp/")
+	if err != nil {
+		return "", err
+	}
+
+	touchCmd := exec.Command("touch", "/tmp/ABCDE")
+	err = touchCmd.Run()
+	if err != nil {
+		return "", err
+	}
```
mauriciovasquezbernal, line 182: What about doing it directly from this executable instead of relying on an external binary?

Also, what about using a true temporal file (by using [os.CreateTemp](https://pkg.go.dev/os#CreateTemp)) to avoid any possible collision.

gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,185 @@
...
+					return &ExpectedFsnotifyEvent{
+						Type: "inotify",
+
+						IMask: 134217732, // 134217732 = 0x08000004 = FS_ATTRIB | FS_EVENT_ON_CHILD
+						Name:  filename,
+
+						Timestamp: utils.NormalizedStr,
+
+						TraceeMntnsId: utils.NormalizedInt,
+						TracerMntnsId: utils.NormalizedInt,
```
mauriciovasquezbernal, line 92: We can also test these two, they should be utilstest.RunnerInfo.MountNsID.

mauriciovasquezbernal (commented):
It looks good. I wondering if we can test more fields instead of normalizing them, like Prio, FaMask, FaType, flags, etc.

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: Thank you so much @mauriciovasquezbernal , for the review!
I faced some challenges while debugging, so it took me a bit more time.

I didn't notice the difference at first, which might be why I wasn't getting any values in the process.
I have updated it with the defined struct.

However, I'm still not able to get values for `comm` and `pcomm`.
Let me know if I'm missing anything here.

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: Yeah, that makes more sense for better understanding of the values.

With the new `generateEvent` method, this is updated to
`IMask: 0x08000002, // FS_MODIFY | FS_EVENT_ON_CHILD`

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: Yes, that works well
I’ve updated the `generateEvent` method to use `os.CreateTemp`

Thanks :)

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: Yeah, `TraceeMntnsId` is the same as `utilstest.RunnerInfo.MountNsID`.

However, the Tracer's MntnsId is different. For now, I've normalized it. 
We should look at the MntnsId of the fsnotify watcher, as that will be our `TracerMntnsId`.

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: Yeah, I tried using different uid/gid values, but neither the tracer nor the tracee has these values. Only runnerInfo has the new, different uid/gid.

Need to debug why we’re unable to retrieve those values in tracer/tracee.

gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+const TASK_COMM_LEN = 16
+
+type Process struct {
+	PPid  uint32    `json:"ppid"`
+	Pid   uint32    `json:"pid"`
+	Tid   uint32    `json:"tid"`
+	Comm  [TASK_COMM_LEN]byte `json:"comm"`
+	PComm [TASK_COMM_LEN]byte `json:"pcomm"`
```
mauriciovasquezbernal, line 41: Can't it be `string`?

gadgets/fsnotify/test/unit/fsnotify_test.go
```diff
@@ -0,0 +1,230 @@
...
+	dirSys := dirInfo.Sys()
+	var dirInode uint64
+	if dirStat, ok := dirSys.(*syscall.Stat_t); ok {
+		dirInode = uint64(dirStat.Ino)
+	}
+	fmt.Printf("Inode of Directory: %d\n", dirInode)
+
+	return inode, dirInode, nil
+}
+
```
mauriciovasquezbernal, line 230: nit: I think implementing a helper to provide the inode of a file will be very useful here (and in other places as well). If you are for it, it'll great to implement this helper on `pkg/utils/inode`.

mauriciovasquezbernal (commented):
Two small nits, but it looks great already. I tested it locally and the test works fine, I'll hit the CI once you handle those two nits. 

Thanks!

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: yeah done, let me know if we need anything else tweaked!

gadgets/fsnotify/test/unit/fsnotify_test.go
4rivappa: I didn't notice any difference, as I couldn't retrieve any values from them.
updated it with `string` just in case.

pkg/utils/inode/inode.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"syscall"
+)
+
+// ExtractFileAndDirInodes extracts the inode values for a given file and its parent directory
+func ExtractFileAndDirInodes(fileName string) (uint64, uint64, error) {
```
mauriciovasquezbernal, line 25: This function is too specific. Create a generic `GetInone(fileName string) (uint64, error)` function, then the caller can invoke it twice: one for the file and another one for the parent.

4rivappa: yeah, updated!

pkg/utils/inode/inode.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"syscall"
+)
+
+// GetInode extracts the inode of a given path
+func GetInode(path string) (uint64, error) {
+	// Get information about the given path (file or directory)
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return 0, fmt.Errorf("failed to stat path %s: %w", path, err)
```
mauriciovasquezbernal, line 28: nit: remove these "failed to" prefixes on the error messages. Otherwise the CI will complain.

mauriciovasquezbernal (commented):
Looks very good, just one minor nit, I'll kick the CI and merge if everything goes fine after it.

pkg/utils/inode/inode.go
4rivappa: yeah, done !

pkg/utils/inode/inode.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"syscall"
+)
+
+// GetInode extracts the inode of a given path
+func GetInode(path string) (uint64, error) {
+	// Get information about the given path (file or directory)
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return 0, fmt.Errorf("could not stat path %s: %w", path, err)
```
mauriciovasquezbernal, line 28: The idea is to avoid all these "failed to", "could not", etc. prefixes to avoid having error messages like:
"failed to X: could not Y: unable to Z". 

Unfortunately it's not well-documented, but we're following these rules: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.semgrep_rules.yaml

4rivappa: yeah, ok .. now I get it.
updated with verbs in continuous form.

mauriciovasquezbernal:
Tests on other kernel versions are failing. One of the reasons is that the testing kernel doesn't have inotify enabled. I enabled and pushed a version of it, but it isn't able to capture the event:

```bash 
$ cd gadgets
$ make IG_VERIFY_IMAGE=false GADGET_TAG=latest KERNEL_VERSION=6.8-with-fsnotify fsnotify/test-unit -o fsnotify
IG_VERIFY_IMAGE=false \
GADGET_TAG=latest \
GADGET_REPOSITORY=ghcr.io/inspektor-gadget/gadget \
VIMTO=vimto \
VIMTO_VM_MEMORY=4096M \
 sudo -E vimto -kernel ghcr.io/mauriciovasquezbernal/ci-kernels:6.8-with-fsnotify -memory 4096M -- go test -v ./fsnotify/test/unit/...
=== RUN   TestFsnotifyGadget
=== RUN   TestFsnotifyGadget/captures_inotify_event
=== PAUSE TestFsnotifyGadget/captures_inotify_event
=== CONT  TestFsnotifyGadget/captures_inotify_event
time="2024-12-27T15:36:19-05:00" level=warning msg="image signature verification is disabled due to using corresponding option"
time="2024-12-27T15:36:19-05:00" level=warning msg="template \"pid\" not found for field \"fa_pid\""
time="2024-12-27T15:36:19-05:00" level=warning msg="template \"timestamp\" not found for field \"timestamp\""
    helpers.go:151: Event wasn't captured
--- FAIL: TestFsnotifyGadget (0.00s)
    --- FAIL: TestFsnotifyGadget/captures_inotify_event (7.24s)
FAIL
FAIL    github.com/inspektor-gadget/inspektor-gadget/gadgets/fsnotify/test/unit 8.316s
FAIL
Error: go: exit status 1
make: *** [Makefile:138: fsnotify/test-unit] Error 128
```

I don't know what's wrong. Perhaps @alban has an idea here.

alban:
The fsnotify gadget does not report when the filesystem activity happens but when the process using the fsnotify kernel API receives the events. The function `generateEvent` in the tests starts a watcher with `fsnotify.NewWatcher()` and then quickly close it with `defer watcher.Close()` without taking the time to receive the events. On Linux, fsnotify uses inotify which is asynchronous, so the filesystem activity is not paused for the tracer to receive the events. So there is a race condition.

The following patch fixes the test:

<details>

```diff
diff --git a/gadgets/fsnotify/test/unit/fsnotify_test.go b/gadgets/fsnotify/test/unit/fsnotify_test.go
index 3450b7993..b5845bf28 100644
--- a/gadgets/fsnotify/test/unit/fsnotify_test.go
+++ b/gadgets/fsnotify/test/unit/fsnotify_test.go
@@ -15,6 +15,8 @@
 package tests
 
 import (
+	"errors"
+	"fmt"
 	"os"
 	"path"
 	"testing"
@@ -36,7 +38,7 @@ type Process struct {
 	Tid   uint32 `json:"tid"`
 	Comm  string `json:"comm"`
 	PComm string `json:"pcomm"`
-};
+}
 
 type EventDetails struct {
 	FileName string
@@ -111,7 +113,7 @@ func TestFsnotifyGadget(t *testing.T) {
 						IIno:    eventDetails.Ino,
 						IInoDir: eventDetails.InoDir,
 
-						Name:  eventDetails.FileName,
+						Name: eventDetails.FileName,
 					}
 				})(t, info, 0, events)
 			},
@@ -183,6 +185,28 @@ func generateEvent() (EventDetails, error) {
 		return EventDetails{}, err
 	}
 
+	// Receiving events.
+forLoop:
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return EventDetails{}, errors.New("channel closed")
+			}
+			if event.Has(fsnotify.Write) {
+				if event.Name != newFile.Name() {
+					return EventDetails{}, fmt.Errorf("watcher: unexpected event: %q, expected %q", event.Name, newFile.Name())
+				}
+				break forLoop
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return EventDetails{}, errors.New("channel closed")
+			}
+			return EventDetails{}, fmt.Errorf("watcher: %w", err)
+		}
+	}
+
 	// Get inode values of test file and its parent directory
 	fileInode, err := inode.GetInode(newFile.Name())
 	if err != nil {
@@ -201,4 +225,3 @@ func generateEvent() (EventDetails, error) {
 	}
 	return eventDetails, nil
 }
-
```

</details>

Output:

```
$ make -C gadgets/ VIMTO=/home/alban/go/bin/vimto IG_VERIFY_IMAGE=false GADGET_TAG=latest KERNEL_VERSION=6.8-with-fsnotify fsnotify/test-unit -o fsnotify
make: Entering directory '/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/gadgets'
IG_VERIFY_IMAGE=false \
GADGET_TAG=latest \
GADGET_REPOSITORY=ghcr.io/inspektor-gadget/gadget \
VIMTO=/home/alban/go/bin/vimto \
VIMTO_VM_MEMORY=4096M \
 sudo -E /home/alban/go/bin/vimto -kernel ghcr.io/mauriciovasquezbernal/ci-kernels:6.8-with-fsnotify -memory 4096M -- go test -v ./fsnotify/test/unit/...
=== RUN   TestFsnotifyGadget
=== RUN   TestFsnotifyGadget/captures_inotify_event
=== PAUSE TestFsnotifyGadget/captures_inotify_event
=== CONT  TestFsnotifyGadget/captures_inotify_event
time="2025-01-30T17:09:07+01:00" level=warning msg="image signature verification is disabled due to using corresponding option"
time="2025-01-30T17:09:07+01:00" level=warning msg="template \"pid\" not found for field \"fa_pid\""
time="2025-01-30T17:09:08+01:00" level=warning msg="template \"timestamp\" not found for field \"timestamp\""
--- PASS: TestFsnotifyGadget (0.00s)
    --- PASS: TestFsnotifyGadget/captures_inotify_event (6.35s)
PASS
ok  	github.com/inspektor-gadget/inspektor-gadget/gadgets/fsnotify/test/unit	7.968s
make: Leaving directory '/home/alban/go/src/github.com/inspektor-gadget/inspektor-gadget/gadgets'
```

mauriciovasquezbernal:
Thanks @alban for the help. I enabled the fsnotify for the testing kernels on https://github.com/mauriciovasquezbernal/ci-kernels/commit/3e1a378c13d8ca53b40ed0188dd747f454a31edd, and new images are already on my repo. However #4005 needs to be merged first to use the new versions.

alban:
Thanks @mauriciovasquezbernal, I restarted all the CI jobs and hopefully it will pick up the new kernels.

4rivappa:
> Actually the branch needs to be rebased on origin/main to pick the new kernels. @4rivappa Could you do that? (or let me know if it is ok for me to click on the "Update branch" button)

Yeah, I was testing CI with updated kernels in other branch. Thought to update this, after successful execution.
We faced issues for these kernel versions -`6.11`, `6.10` and `5.4`.

Looking into it, and also have some issue with `i_mask_set` struct.
Sharing more info on slack ! Thankyou :)

mauriciovasquezbernal:
Tests are failing for kernel 6.10 and 6.11 and we're not currently sure why. As we discussed internally, please skip the test on those versions and open an issue to fix it later on. (This PR was about implementing the tests, I'm totally fine with not handling the fix right now)

alban:
> Tests are failing for kernel 6.10 and 6.11 and we're not currently sure why.

Do those kernels have the following configs?
- CONFIG_FSNOTIFY
- CONFIG_INOTIFY_USER
- CONFIG_FANOTIFY
- CONFIG_FANOTIFY_ACCESS_PERMISSIONS

If it is not that, maybe it is a change in the kernel.

The error from the CI is:

> starting operators: starting operator "oci": starting operator "ebpf": creating eBPF collection: program fsnotify_insert_event_e: load program: bad CO-RE relocation: invalid func unknown#195896080 (135 line(s) omitted)

From reading the source, the function [fsnotify_insert_event_e](https://github.com/inspektor-gadget/inspektor-gadget/blob/8ca73cfd05fe5cfc9843275d02ea6232cdda3d58/gadgets/fsnotify/program.bpf.c#L296) does the following relocations:
* struct fsnotify_group -> priority
* struct inotify_event_info -> wd, mask, sync_cookie, name_len, name, fse
* struct fanotify_event -> mask, type, pid, fse
* struct fanotify_path_event -> fae
* struct fanotify_perm_event -> fae

The only changed I notice in kernel 6.10 compared to IG's vmlinux.h is `struct fsnotify_group -> priority` (type changed from `unsigned int` to `enum fsnotify_group_prio` in Linux v6.9 with [477cf917dd028](https://github.com/torvalds/linux/commit/477cf917dd02853ba78a73cdeb6548889e5f8cd7)) but this type change should not change the [relocation accessor](https://github.com/torvalds/linux/blob/09fbf3d502050282bf47ab3babe1d4ed54dd1fd8/include/uapi/linux/bpf.h#L7453-L7498) so I don't know what is going on.

Still, you can try this patch in a new branch to check if it is related to the type change:
```patch
-ee->prio = BPF_CORE_READ(group, priority);
+ee->prio = 0;
```

4rivappa:
> The only changed I notice in kernel 6.10 compared to IG's vmlinux.h is struct fsnotify_group -> priority

Ha, exactly. I just checked the changelog !

> Still, you can try this patch in a new branch to check if it is related to the type change:

Even removing the `BPF_CORE_READ(group, priority)` method, didn't help.
Here is the [workflow](https://github.com/4rivappa/inspektor-gadget/actions/runs/13286574895), where I have triggered the kernel-matrix unit-tests.

mauriciovasquezbernal:
> Do those kernels have the following configs?

    CONFIG_FSNOTIFY
    CONFIG_INOTIFY_USER
    CONFIG_FANOTIFY
    CONFIG_FANOTIFY_ACCESS_PERMISSIONS

The current ci-kernels only have CONFIG_FANOTIFY. I tried with those flags as well and the result was the same.

mauriciovasquezbernal:
I can confirm it's not an issue with the testing images. I tried this on an Ubuntu 24.10 machine with kernel 6.11 and it prints the same error.

alban:
I added some `fmt.Printf` in `AssignMetadataToInstructions` (`vendor/github.com/cilium/ebpf/btf/ext_info.go`) to print the ebpf instructions and relocations before they are applied (and before the code is poisoned) and I get the following:

```
Instruction: StXMemW dst: r9 src: r1 off: 8 imm: 0 ; source 	gadget_event->type_raw = fa_resp; ; relo 
Instruction: MovImm dst: r1 imm: 64 ; source  ; relo CORERelocation(byte_off, Struct:"fsnotify_group"[0:7], local_id=965)
Instruction: StXMemDW dst: rfp src: r6 off: -120 imm: 0 ; source  ; relo 
Instruction: AddReg dst: r6 src: r1 ; source  ; relo 
Instruction: MovReg dst: r1 src: rfp ; source  ; relo 
Instruction: AddImm dst: r1 imm: -16 ; source  ; relo 
Instruction: MovImm dst: r2 imm: 4 ; source 	gadget_event->prio = BPF_CORE_READ(group, priority); ; relo 
```

The failing relocation is `CORERelocation(byte_off, Struct:"fsnotify_group"[0:7], local_id=965)`.
Accessor `[0:7]` means 8th field of `struct fsnotify_group` and according to [our current vmlinux.h](https://raw.githubusercontent.com/inspektor-gadget/inspektor-gadget/refs/heads/main/include/gadget/amd64/vmlinux.h), that's `priority`.

alban:
> Even removing the BPF_CORE_READ(group, priority) method, didn't help.
> Here is the [workflow](https://github.com/4rivappa/inspektor-gadget/actions/runs/13286574895), where I have triggered the kernel-matrix unit-tests.

Could it be that it is a different error on a different relocation?

mauriciovasquezbernal:
@4rivappa please rebase this PR once #4046 is merged and add the logic to skip the test on kernel 5.4.

pkg/utils/inode/inode.go
```diff
@@ -0,0 +1,37 @@
...
+package inode
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// GetInode extracts the inode of a given path
+func GetInode(path string) (uint64, error) {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Is it that bad? I think it could be useful for other tests later on.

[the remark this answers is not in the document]
4rivappa: Initially it was an unexported function inside fsnotify_test.go, 
But after receiving this [comment](https://github.com/inspektor-gadget/inspektor-gadget/pull/3729#discussion_r1897953489) from mauriciovasquezbernal, I moved it !

Should we undo this , ?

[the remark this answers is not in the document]
alban: > Is it that bad? I think it could be useful for other tests later on.

If it is for the tests, maybe it could be in `/pkg/testing/utils`.

I don't think it would be useful outside of tests because it does not allow to avoid symlink container escape issues when getting the inode of a file in a container. For for tests, this is fine.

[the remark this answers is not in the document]
mauriciovasquezbernal: Right, it makes sense to move this to the testing package then!

mauriciovasquezbernal (approved):
Everything is now green. Thanks a lot for your perseverance to get this done and helping us to find a fix a bug on this gadget.

PR #3730, opened 2024-11-29
expose SocketPath in k8s.go

matthyx:
# expose SocketPath in k8s.go

In our node-agent I need to create and use a runtime.NewImageServiceClient for which it's very handy to make sure I have the same socket path as IG.

## How to use

N/A

## Testing done

Works in my PR https://github.com/kubescape/node-agent/pull/384/

### DIFF
--- a/pkg/container-collection/k8s.go
+++ b/pkg/container-collection/k8s.go
@@ -41,10 +41,11 @@ import (
 )
 
 type K8sClient struct {
-	clientset     *kubernetes.Clientset
-	nodeName      string
-	fieldSelector string
-	runtimeClient runtimeclient.ContainerRuntimeClient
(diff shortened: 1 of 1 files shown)


pkg/container-collection/k8s.go
```diff
@@ -45,6 +45,7 @@ type K8sClient struct {
 	nodeName      string
 	fieldSelector string
 	runtimeClient runtimeclient.ContainerRuntimeClient
+	SocketPath    string
```
mauriciovasquezbernal, line 48:
```suggestion
	RuntimeSocketPath    string
```

For a second I thought it was a socket path to talk to the API server.

mauriciovasquezbernal (approved):
LGTM. I'll merge once the CI is green.

PR #3741, opened 2024-12-02
gadgets/deadlock: fix dead `tid` clean-up

Snehil-Shah:
# Fix incorrect dead `tid` clean-up

- We were previously clearing all held mutexes by a thread as well when a thread died. This might not necessarily be the case in actuality. It's better to stay conservative and wait for the process to die before cleaning up dead tids.
- We were previously sending each dead PID to userspace when tracing `sched/process_exit`. This can potentially clog things up with PIDs that we are not interested in. Added a new map for tracking traced PIDs so we only send PIDs we want to delete.

## How to use

Just run the gadget. There shouldn't be much difference for most use cases.

## Testing done

Ran the gadget.


gadgets/deadlock/go/program.go
```diff
@@ -264,13 +267,33 @@ func gadgetInit() int {
...
+				deadTids[tid] = struct{}{}
+			}
+			for deadTid := range deadTids {
+				err := threadToHeldMutexesBPFMap.Delete(deadTid)
+				if err != nil {
+					api.Warnf("failed to delete key from threads: %s", err)
+					continue
+				}
+			}
+		}
```
Snehil-Shah, line 294: This can be further optimized by maintaining a separate map in Go that maps PIDs to its threads so as to avoid traversing all edges of the graph, but maybe I'm just overthinking and this would work just fine without having to worry about ANOTHER buffer. WDYT?

PR #3752, opened 2024-12-03
ig-k8s: Fix fix server version message

mauriciovasquezbernal:
Print the version only when running with -serve, otherwise it messes up some json output used by the advise network-policy gadget.

It turns out adding a new info log on the gadget tracer manager breaks something. I was too confident a simple message won't break anything and merged #3743 without waiting for the CI to finish.

Fixes: a46f5c2efc55 ("ig-k8s: Fix server version message")

--- 

Untitled

### DIFF
--- a/gadget-container/gadgettracermanager/main.go
+++ b/gadget-container/gadgettracermanager/main.go
@@ -120,8 +120,6 @@ func init() {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
CI failure is not related to this change, moving ahead.

PR #3756, opened 2024-12-04
operators/wasm: Add functions to get syscalls names and declarations.

eiffel-fl:
TODO:
- [ ] Documentation once agreed on the API.


mauriciovasquezbernal (commented):
I think the syscall declaration related API is too granular, i.e. a lot of calls need to be done to get details about a syscall. What about defining a type with the full syscall implementation and enabling `getSyscallDeclaration` to fill a buffer passed by the user:

```c
struct gadget_syscall_param {
  char name[64]; // is 64 enough?
  u8 isPointer;
  ...
};

struct gadget_syscall_syscall {
  char name[64]; // same
  gadget_syscall_param param[6];
  int nparams; // relying on name == "" for the last param
};
```

The the caller can use something like:

```c
gadget_syscall mysyscall;
getSyscallDeclaration(&mysyscall)
```

For sure, a more friendly wrapper should be provided for golang.

pkg/gadgets/traceloop/syscall-helpers/syscall_helpers.go
```diff
@@ -219,27 +193,27 @@ func gatherSyscallsDeclarations() (map[string]syscallDeclaration, error) {
...
-func (s syscallDeclaration) paramIsPointer(paramNumber uint8) (bool, error) {
+func (s SyscallDeclaration) ParamIsPointer(paramNumber uint8) (bool, error) {
 	if int(paramNumber) >= len(s.params) {
 		return false, fmt.Errorf("param number %d out of bounds for syscall %q", paramNumber, s.name)
 	}
 	return s.params[paramNumber].isPointer, nil
 }
 
-func (s syscallDeclaration) getParameterName(paramNumber uint8) (string, error) {
+func (s SyscallDeclaration) GetParameterName(paramNumber uint8) (string, error) {
```
mauriciovasquezbernal, line 216: Related to my previous comment. I think we can make `param` and all its fields public, then this function won't be needed (specially important regarding the error handling around). The caller code could be something like:

```golang 
			syscallDeclaration, err := syscallhelpers.GetSyscallDeclaration(syscallsDeclarations, event.Syscall)
			if err != nil {
				return nil, fmt.Errorf("getting syscall definition")
			}

			for i, param := range syscallDeclaration.Params {
				fmt.Printf("param in position %d is named %s and is a pointer: %t\n", i, param.Name, param.IsPointer)
			}
```

WDYT?

mauriciovasquezbernal (commented):
The new syscalls API seems good to me. Some comments about padding and about the syscall helpers.

pkg/gadgets/traceloop/syscall-helpers/syscall_helpers.go
```diff
@@ -35,12 +35,12 @@ type param struct {
 	isPointer bool
 }
 
-type syscallDeclaration struct {
+type SyscallDeclaration struct {
 	name   string
 	params []param
 }
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Why not to make this good now and avoid changing it later on when removing traceloop? It seems like double work to review something that will be changed again in few months.

wasmapi/go/syscall.go
```diff
@@ -0,0 +1,101 @@
...
+	Params []SyscallParam
+}
+
+func GetSyscallName(id uint16) (string, error) {
+	ptr := bufPtr(getSyscallName(uint32(id)))
+	if ptr == 0 {
+		return "", fmt.Errorf("getting syscall name for syscall id %d", id)
+	}
+
+	return ptr.string(), nil
```
mauriciovasquezbernal, line 66: It's missing a call to `ptr.free()`

wasmapi/go/syscall.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func GetSyscallDeclaration(name string) (SyscallDeclaration, error) {
+	var decl syscallDeclaration
+
+	bufPtr, err := anyToBufPtr(&decl)
+	if err != nil {
+		return SyscallDeclaration{}, err
+	}
```
mauriciovasquezbernal, line 75: Why to copy an empty value of decl to the buffer if it'll be rewritten by the host?

mauriciovasquezbernal: I'm wondering if something like `	bufPtr := bytesToBufPtr((*byte)(reflect.ValueOf(&decl).UnsafePointer()))` does the trick? (also remove the copy below)

pkg/operators/wasm/testdata/badguest/program.go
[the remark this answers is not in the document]
mauriciovasquezbernal: The goal of this test is to emulate a bad (malicious?) guest trying to break the host by calling different functions with wrong arguments. Perhaps a better idea would be to implement it using a fuzzer, but I never tried it, so I sticked to testing some bad combinations for all functions.

wasmapi/go/syscall.go
mauriciovasquezbernal: I don't think we need to pass through `[]bytes`, it should be possible to get a pointer to syscallDeclaration directly. The following appears to work for me:

```golang
func GetSyscallDeclaration(name string) (SyscallDeclaration, error) {
	var decl syscallDeclaration

	// create pointer to decl that will be filled by the host
	unsafePtr := unsafe.Pointer(&decl)
	size := unsafe.Sizeof(decl)
	bufPtr := bufPtr(uint64(size)<<32 | uint64(uintptr(unsafePtr)))

	ret := getSyscallDeclaration(uint64(stringToBufPtr(name)), uint64(bufPtr))
	runtime.KeepAlive(name)
	runtime.KeepAlive(decl)
	if ret == 1 {
		return SyscallDeclaration{}, fmt.Errorf("syscall declaration %s not found", name)
	}

	declaration := SyscallDeclaration{
		Name:   fromCString(decl.name[:]),
		Params: make([]SyscallParam, decl.nrParams),
	}
	for i := range decl.nrParams {
		declaration.Params[i] = SyscallParam{
			Name:      fromCString(decl.params[i].name[:]),
			IsPointer: (decl.params[i].flags & isPointerFlag) != 0,
		}
	}

	return declaration, nil
}
```

I tried creating a generic helper for it:

```golang 
func anyToBufPtr2(b any) bufPtr {
	unsafePtr := unsafe.Pointer(&b)
	size := unsafe.Sizeof(b)
	return bufPtr(uint64(size)<<32 | uint64(uintptr(unsafePtr)))
}
``` 

but it didn't work as the size of the pointer was always 8. (It seems `unsafe.Sizeof` on `any` is messing up, but I didn't investigate more)

pkg/operators/wasm/syscalls.go
```diff
@@ -0,0 +1,168 @@
...
+		return syscallhelpers.GatherSyscallsDeclarations()
+	})()
+	if err != nil {
+		i.logger.Warnf("getSyscallDeclaration: gathering syscall declarations: %v", err)
+		stack[0] = 1
+		return
+	}
+
+	// This can only occur if above sync.Once failed, so let's return as the stack
+	// return value was correctly set.
```
mauriciovasquezbernal, line 114: Is it really needed? I suppose the return on L110 will handle the failure case, and we can trust the function to return something different than nil when err == nil.

mauriciovasquezbernal (commented):
Quick comment while checking it for a last time. I think the implementation is good now, time to focus on docs!

pkg/operators/wasm/testdata/syscall/program.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"fmt"
+
+	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
+)
+
+//export gadgetInit
+func gadgetInit() int {
+	syscallID := uint16(1337)
```
burak-ok, line 25: Make `1337` a constant with a descriptive name or add a comment that we expect that 1337 isnt a syscall number with a name (currently)

pkg/operators/wasm/testdata/syscall/program.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	// Check behavior is conform with strace.
+	expectedSyscallName := fmt.Sprintf("syscall_%x", syscallID)
+	if syscallName != expectedSyscallName {
+		api.Errorf("mismatch for syscall %d: expected %q, got %q", syscallID, expectedSyscallName, syscallName)
+		return 1
+	}
+
+	syscallID = uint16(428)
```
burak-ok, line 39: Here also, `openTreeSyscallNum := 428`

pkg/operators/wasm/testdata/badguest/program.go
```diff
@@ -118,6 +118,18 @@ func mapDelete(m uint32, keyptr uint64) uint32
...
+
+//go:wasmimport env syscallDeclarationGetParameterCount
+func syscallDeclarationGetParameterCount(s uint32) int32
+
+//go:wasmimport env syscallDeclarationParamIsPointer
+func syscallDeclarationParamIsPointer(s uint32, param uint32) uint32
+
+//go:wasmimport env syscallDeclarationGetParameterName
+func syscallDeclarationGetParameterName(s uint32, param uint32) uint64
+
```
mauriciovasquezbernal, line 132: Is it old stuff that needs to be removed?

pkg/operators/wasm/testdata/syscall/go.mod
```diff
@@ -0,0 +1,8 @@
+module main
+
+go 1.22.8
```
mauriciovasquezbernal, line 3: Please update to 1.23.0

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -445,3 +445,26 @@ Parameters:
...
+Return value:
+- (u64) A string containing the name of the syscall, in case the syscall is unknown, it returns "syscall_ID" with ID displayed as hexadecimal like strace.
+
+#### `getSyscallDeclaration(name uint64, pointer uint64) uint32`
+
+Get the syscall declaration for this syscall name.
+
+Parameters:
+- `name` (u64): Syscall name.
+- `pointer` (u64): A pointer to a syscallDeclaration structure, it is used to store the data instead of returning them.
```
mauriciovasquezbernal, line 467: Can you add a link to this structure definition?

mauriciovasquezbernal (approved):
LGTM once these small final nits are fixed. Thanks for handling all the feedback on this one.

docs/gadget-devel/gadget-wasm-api-raw.md
mauriciovasquezbernal: A user reading this documentation won't have any clue of what "syscallDeclaration structure" is. I'm suggesting to add a link to the Github code defining it (it doesn't matter if the link is broken until we merge this PR), or copy and paste the definition here. The goal is to show the layout of the structure, different fields and their sizes.

PR #3757, opened 2024-12-04
wasm/testdata: Use in-tree ig to build the images

burak-ok:
This ensures the most up to date is used for building and exporting the images

### DIFF
--- a/pkg/operators/ebpf/testdata/Makefile
+++ b/pkg/operators/ebpf/testdata/Makefile
@@ -1,4 +1,5 @@
 ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+IG ?= ig
 
 TEST_ARTIFACTS = \
 	empty \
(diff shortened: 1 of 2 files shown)


mauriciovasquezbernal:
I don't understand this change. Why there are two different binaries on the system? 

What I do on my system is to create a symlink from /usr/bin/ig to the compiled version of the binary to avoid having two different ones around:

```bash
$ ls -l $(which ig)
lrwxrwxrwx 1 root root 48 abr 10  2024 /usr/bin/ig -> /home/mauriciov/kinvolk/ebpf/inspektor-gadget/ig
```

Will this help on this case?

burak-ok:
> I don't understand this change. Why there are two different binaries on the system?

I never installed `ig` from upstream repo on my system. Never needed that use case.
Also in almost any other Makefiles we have a ENV variable for the `ig` path - so the use case seems to be there that there might be different binaries

> What I do on my system is to create a symlink from /usr/bin/ig to the compiled version of the binary to avoid having two different ones around:
> Will this help on this case?

I also have multiple folders for `Inspektor Gadget`. Mostly that is done when I work on parallel stuff. Of course I could always commit and then checkout another branch, but then I might forget about the other branch :sweat_smile: 
But I don't want to force a change because of the way how I work

mauriciovasquezbernal:
> Also in almost any other Makefiles we have a ENV variable for the ig path - so the use case seems to be there that there might be different binaries

What about adding this also to this makefile so each developer can customize the binary to use?

burak-ok:
> > Also in almost any other Makefiles we have a ENV variable for the ig path - so the use case seems to be there that there might be different binaries
> 
> What about adding this also to this makefile so each developer can customize the binary to use?

Just changed it to the env variable way.
But after looking at the changes again and seeing `IG_SOURCE_PATH=$(realpath $(ROOT_DIR)/../../../..)` I think in this situation it might actually be better for consistency to always refer to `ig` through `$(realpath $(ROOT_DIR)/../../../../ig`

mauriciovasquezbernal (approved):
Approving as it is. If you still think having it harcoded to `../ ... /ig` is better, that's fine for me as well.

PR #3760, opened 2024-12-04
tests: Increase timeout for snapshotter tests

### LINKED ISSUE #3033: snapshotters: Execution doesn't automatically end
The snapshotters gadgets collect the status of the system and dump it. The built-in snapshotters print the results and the command returns immediately, however, for the image-based ones, the user has to hit Ctrl + C to make the program finish. 

This is because the default timeout 0 means to keep running forever, causing 

https://github.com/inspektor-gadget/inspektor-gadget/blob/6e250d5258598d56f206ddb8d99e911666f81e19/pkg/gadget-context/run.go#L119

to block until the context is canceled when pressing Ctrl + C.

The program should return after the information is printed. 

The scope of this issue is to discuss the implications and challenges of this, I know @flyth has already thought a lot about this one. Perhaps you can share your opinions and possible ways to fix it?

### LINKED ISSUE #3361: ci: snapshot process test is too flaky
This test is failing a lot:

- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10386743239/attempts/1#summary-28759182888
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10380957419/attempts/1#summary-28742131702
- https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/10393961200/attempts/1#summary-28783086720

#3338 tried to fix it but it's still failing.

mauriciovasquezbernal:
Snapshotter tests are failing a lot (see issue #3361). It seems to be that 1 second is too low for the message pump to send the information to the client, hence it's being dropped.

This commit increases the timeout to 5 seconds, however the final solution should be to fix #3033.

Fixes #3361

### DIFF
--- a/gadgets/snapshot_process/test/integration/snapshot_process_test.go
+++ b/gadgets/snapshot_process/test/integration/snapshot_process_test.go
@@ -68,12 +68,14 @@ func TestSnapshotProcess(t *testing.T) {
 	var testingOpts []igtesting.Option
 	commonDataOpts := []utils.CommonDataOption{utils.WithContainerImageName(containerImage), utils.WithContainerID(testContainer.ID())}
 
-	// TODO: timeout shouldn't be required
+	// TODO: timeout shouldn't be required. We need to use something big like 5
+	// seconds to avoid the message being lost.
+	const timeoutParam = "--timeout=5"
 	switch utils.CurrentTestComponent {
 	case utils.IgLocalTestComponent:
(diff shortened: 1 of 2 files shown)


mauriciovasquezbernal:
I've tested it multiple times on https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/12166918916 and it seems to be working fine. I'll rerun it a couple of times here to see if they work fine.

PR #3762, opened 2024-12-05
operators/wasm: Add WASM functions to get running containers

eiffel-fl:
TODO:
- [ ] Add documentation once agreed on the API.
- [ ] Add unit tests for containers API once agreed on the API (idea: create a container in go host, get the list in guest and ensure len(list) > 1).

### DIFF
--- a/pkg/operators/wasm/containers.go
+++ b/pkg/operators/wasm/containers.go
@@ -1,0 +1,109 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package wasm
+
+import (
+	"context"
+
+	"github.com/tetratelabs/wazero"
+	wapi "github.com/tetratelabs/wazero/api"
+
+	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
+	igmanager "github.com/inspektor-gadget/inspektor-gadget/pkg/ig-manager"
+)
+
+func (i *wasmOperatorInstance) addContainersFuncs(env wazero.HostModuleBuilder) {
+	exportFunction(env, "getContainers", i.getContainers,
+		[]wapi.ValueType{},
+		[]wapi.ValueType{wapi.ValueTypeI32}, // Containers array
+	)
+
+	exportFunction(env, "containerGetCgroupID", i.containerGetCgroupID,
+		[]wapi.ValueType{
+			wapi.ValueTypeI32, // Container
+		},
+		[]wapi.ValueType{wapi.ValueTypeI64}, // CgroupID or error
+	)
+
+	exportFunction(env, "containerGetMntns", i.containerGetMntns,
+		[]wapi.ValueType{
+			wapi.ValueTypeI32, // Container
+		},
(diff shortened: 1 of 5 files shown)


pkg/operators/wasm/containers.go
```diff
@@ -0,0 +1,109 @@
...
+			wapi.ValueTypeI32, // Container
+		},
+		[]wapi.ValueType{wapi.ValueTypeI64}, // MntNs ID or error
+	)
+}
+
+// getContainers gets a list of all running containers.
+// Return value:
+// - Array of container handle on success, 0 on error
+func (i *wasmOperatorInstance) getContainers(ctx context.Context, m wapi.Module, stack []uint64) {
```
blanquicet, line 51: IMO, we shouldn't create a dedicated API for exposing containers information to WASM, but we should create a new data source for it. For instance, the container-collection could be an operator who register a new data source and emits events for CREATED and DELETED containers. Basically, what we have in [pubsub.go](https://github.com/inspektor-gadget/inspektor-gadget/blob/michael/oci/decl-params/pkg/container-collection/pubsub.go#L107). Then, the localmanager, kubemanager and the WASM module (when the gadget requires it) subscribe to that data source. We could even, optionally, emit those container events through gRPC if user requests for it.

mauriciovasquezbernal: Why do we need to wait for "stage 3" for it? I think Jose's suggestion makes sense and will simplify all of it by avoiding this dedicated API.

mauriciovasquezbernal: Yes, I'm aware of that. But I think in the long term it'll pay off to implement it on the "right way" instead of creating a dedicated throw-away API. All the datasource integration with wasm is implemented, it seems to me extending the local and kube manager operators to emit container events won't be that hard.

wasmapi/go/map.go
```diff
@@ -60,7 +60,7 @@ type MapType uint32
 const (
 	UnspecifiedMap MapType = iota
 	Hash
-	Array
+	ArrayMap
```
burak-ok, line 63: In this commit and even while PR there is no other mention of `ArrayMap`. Is this used anywhere?

pkg/operators/wasm/containers.go
burak-ok: With having a WASM interface for this we have the advantage of deduplicating all the container information.
With PubSub we will have many copies of the same information about the containers. We tried in other PRs to minimize the footprint of caching the containers and this could potentially enlarge it again

Container Collection will have a copy of all containers
LocalManager will have a copy of all containers
+ every and operator layer that has interest in all the currently running containers

And every layer has duplicated logic to cache all containers concurrently, too.


I am currently not sure about this proposal

wasmapi/go/containers.go
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+package api
+
+import (
+	"fmt"
+)
+
+//go:wasmimport env getContainers
+func getContainers() uint32
```
burak-ok, line 22: Somewhere it should be documented that this function returns the containers on the CURRENT host.

In a K8s scenario the WASM developer might think it could return all containers from the whole Cluster

wasmapi/go/map.go
burak-ok: Was just wondering why a const was changed and no uses of it -> It looks like it was never used in the past

pkg/operators/wasm/containers.go
mauriciovasquezbernal: Usually operators won't care about the full container object, so they could create a slim version and store it. Also, concurrency could be handled by the operator emitting the container events, so different consumers don't have care about synchronization of concurrent events.

mauriciovasquezbernal: One issue with that approach is that the data source is registered and the CLI operator tries to print it to the user. I think we'll need a way to mark this datasource as "internal". @flyth WDYT?

mauriciovasquezbernal: mmm, does the call to `ds, err := api.GetDataSource("containers")` return successfully but then the subscription isn't registered? That's weird.

Issue #3563, opened 2024-10-08
gadgets: Assess implementing traceloop as an image-based gadget


mauriciovasquezbernal:
The [traceloop](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/traceloop) gadget is unique. It uses an over-writable ring buffer to store the last syscalls executed by a process / container, they are only dumped when requested by the user. We usually describe it as a flight recorder as it saves the last syscalls the process / container executed before crashing / exiting.

The purpose of this issue is to assess the effort needed to implement this gadget as an image-based one.

PR #3763, opened 2024-12-06
Convert traceloop to image-based gadget

eiffel-fl:
This PR converts traceloop to image-based gadget:

```bash
$ sudo ../ig run --verify-image=false traceloop:francis-traceloop-img-based   
WARN[0000] No digest found for image sha256:5ec7527ba87826d36c6ff8a8fbcf7b6e34ebac32c2f63fb3f23f5875a0ec59dc 
^CWARN[0002] perfReaderRead: reading perf buffer: epoll wait: i/o timeout 
INFO[0002] &{{{{   0   1970-01-01T00:00:00.000000000Z} { {  map[] } false { }}} 1970-01-01T03:12:42.522003038Z normal } {4026533517} 0 3506 mysqld futex [{uaddr 0x5654d0ad90a0 <nil>} {op 129 <nil>} {val 1 <nil>} {utime 0x0 <nil>} {uaddr2 0x0 <nil>} {val3 1 <nil>}] 0}
...
INFO[0002] &{{{{   0   1970-01-01T00:00:00.000000000Z} { {  map[] } false { }}} 1970-01-01T03:12:42.550589450Z normal } {4026533877} 4 5018 apache2 select [] unfinished}
```

Requisites PR:
* https://github.com/inspektor-gadget/inspektor-gadget/pull/3762
* https://github.com/inspektor-gadget/inspektor-gadget/pull/3756
* https://github.com/inspektor-gadget/inspektor-gadget/pull/3815
* https://github.com/inspektor-gadget/inspektor-gadget/pull/3725
* https://github.com/inspektor-gadget/inspektor-gadget/pull/3881
*  #3957
* https://github.com/inspektor-gadget/inspektor-gadget/pull/4129

THIS PR FOCUSES ONLY ON CONVERTING TRACELOOP TO IMAGE-BASED GADGET.
The goal is not to refactor existing traceloop code (both golang and eBPF), this can be done later on.

TODO:
- [x] Use mapUpdate() to fill the syscallDefs map in eBPF to have better enrichment of syscall having char* as argument. 
- [x] Use datasource to have container enrichment and better printing.
- [x] Documentation once agreed on the printing.
- [x] Integration test once agree on the printing.

Future work (will be converted to issues once merged, see https://github.com/inspektor-gadget/inspektor-gadget/issues/3563#issuecomment-2401746552):
- [x] Traceloop stage 3: Automatically attach/detach to containers when created/deleted.
- [x] Traceloop stage 4: Improve the UX to mimic the built-in one: not needed, attach/detach can already mimic it.
- [x] Bump available map type to reflect cilium/ebpf 0.17.1: https://github.com/inspektor-gadget/inspektor-gadget/pull/3889
- [x] Refactor wasm_test.go: https://github.com/inspektor-gadget/inspektor-gadget/pull/4107.
- [ ] Refactor containers datasource to use a structure instead of accessing fields directly: https://github.com/inspektor-gadget/inspektor-gadget/issues/4288
- [ ] Add containers datasource to k8s manager: https://github.com/inspektor-gadget/inspektor-gadget/issues/4285
- [ ] Add syscalls filtering: https://github.com/inspektor-gadget/inspektor-gadget/issues/4286
- [ ] Add support for array of struct fields: https://github.com/inspektor-gadget/inspektor-gadget/issues/3032
- [ ] Refine using timestamp to SYS_EVENT_CONT, as it is not enough: https://github.com/inspektor-gadget/inspektor-gadget/pull/3763#discussion_r2005520442: https://github.com/inspektor-gadget/inspektor-gadget/issues/4287
- [ ] Use bytes.IndexByte() in \*romCString\*(): https://github.com/inspektor-gadget/inspektor-gadget/pull/3763#discussion_r2027073916: https://github.com/inspektor-gadget/inspektor-gadget/issues/4284

### DIFF
--- a/gadgets/traceloop/README.mdx
+++ b/gadgets/traceloop/README.mdx
@@ -1,0 +1,120 @@
+---
+title: traceloop
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# traceloop
+
+The traceloop gadget is a syscalls flight recorder.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Unsupported
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/traceloop:%IG_TAG% [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Guide
+
+First, we need to run an application that generates some events.
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Unsupported
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ docker run -it --rm --name test-traceloop busybox /bin/sh
+        ```
+    </TabItem>
+</Tabs>
+
+Then, let's run the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Unsupported
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run traceloop:%IG_TAG% --containername test-traceloop
+        RUNTIME.CONTAINERNAME                        CPU         PID COMM             SYSCALL                     PARAMETERS                  RET
+        ```
+    </TabItem>
+</Tabs>
+
+Now, let's generate some events:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Unsupported
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        Run a command inside the container:
+
+        ```bash
+        / # ls
+        ```
+    </TabItem>
+</Tabs>
+
+Let's collect the syscalls:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Unsupported
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        Press Ctrl+C to collect the syscalls:
+
+        ```bash
+        $ sudo ig run traceloop:%IG_TAG% --containername test-traceloop
+        RUNTIME.CONTAINERNAME                        CPU         PID COMM             SYSCALL                     PARAMETERS                  RET
+        ...
+        test-traceloop                            5         58054 sh               execve                    filename="/bin/ls", a…             0
+        test-traceloop                            5         58054 ls               brk                       brk=0                  102559763509…
+        test-traceloop                            5         58054 ls               mmap                      addr=0, len=8192, pro… 123786398932…
+        test-traceloop                            5         58054 ls               access                    filename="/etc/ld.so.… -1 (Permissi…
+        ...
+        test-traceloop                            5         58054 ls               write                     fd=1, buf="\x1b[1;34m…           201
+        test-traceloop                            5         58054 ls               exit_group                error_code=0                       X
+        ...
+        ```
+    </TabItem>
+</Tabs>
+
+Finally, clean the system:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Unsupported
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ docker rm -f test-traceloop
+        ```
+    </TabItem>
+</Tabs>
+
+## Limitations
+
+This gadget has the following limitations:
+1. It cannot be run in kubernetes context.
+2. Timestamps are not filled on kernel older than 5.7.
--- a/gadgets/traceloop/artifacthub-pkg.yml
+++ b/gadgets/traceloop/artifacthub-pkg.yml
@@ -1,0 +1,29 @@
+# Artifact Hub package metadata file
+version: 0.39.0
+name: "traceloop"
+category: monitoring-logging
+displayName: "traceloop"
+createdAt: "2024-10-08T09:57:54Z"
+digest: "2024-10-08T09:57:54Z"
+description: "syscalls flight recorder"
+logoURL: "https://inspektor-gadget.io/media/brand-icon.svg"
+license: ""
+homeURL: "https://inspektor-gadget.io/docs/latest/gadgets/traceloop"
+containersImages:
+    - name: gadget
+      image: "ghcr.io/inspektor-gadget/gadget/traceloop:latest"
+      platforms:
+        - linux/amd64
+        - linux/arm64
+keywords:
+    - gadget
+links:
+    - name: source
+      url: "https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/traceloop"
+install: |
+    # Run
+    ```bash
+    sudo ig run ghcr.io/inspektor-gadget/gadget/traceloop:latest
+    ```
+provider:
+    name: Inspektor Gadget
--- a/gadgets/traceloop/build.yaml
+++ b/gadgets/traceloop/build.yaml
@@ -1,0 +1,5 @@
+wasm: go/program.go
+# To print debug messages:
+# cflags: '-DSHOW_DEBUG'
+# To print error messages:
+# cflags: '-DSHOW_ERROR'
--- a/gadgets/traceloop/gadget.yaml
+++ b/gadgets/traceloop/gadget.yaml
@@ -1,0 +1,37 @@
+name: traceloop
+description: syscalls flight recorder
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/traceloop
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/traceloop
+annotations:
+  enable-containers-datasource: true
+datasources:
+  traceloop:
+    annotations:
+      description: Syscalls
+    fields:
+      mntns_id:
+        annotations:
+          template: mntns_id
+      cpu:
+        annotations:
+          description: CPU number where the syscall was executed
+          columns.maxwidth: 4
+      pid:
+        annotations:
+          template: pid
+      comm:
+        annotations:
+          template: comm
+      syscall:
+        annotations:
+          template: syscall
+      parameters:
+        annotations:
+          description: Syscall's parameters
+          columns.width: 20
+      ret:
+        annotations:
+          description: Syscall's return value
+          columns.width: 10
+          columns.alignment: right
--- a/gadgets/traceloop/go/go.mod
+++ b/gadgets/traceloop/go/go.mod
@@ -1,0 +1,11 @@
+module traceloop
+
+go 1.23.0
+
+toolchain go1.23.4
+
+// Version doesn't matter because of the replace directive below.
+require github.com/inspektor-gadget/inspektor-gadget v0.0.0
+
+// Only needed by in-tree gadgets
+replace github.com/inspektor-gadget/inspektor-gadget => ../../../
--- a/gadgets/traceloop/go/program.go
+++ b/gadgets/traceloop/go/program.go
@@ -1,0 +1,821 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+	"unsafe"
+
+	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
+)
+
+type eventType uint32
+
+const (
+	syscallEventTypeEnter eventType = 0
+	syscallEventTypeExit            = 1
+	syscallEventTypeCont            = 2
+)
+
+// These consts must match the content of program.bpf.c.
+const (
+	useNullByteLength        uint64 = 0x0fffffffffffffff
+	useRetAsParamLength      uint64 = 0x0ffffffffffffffe
+	useArgIndexAsParamLength uint64 = 0x0ffffffffffffff0
+	paramProbeAtExitMask     uint64 = 0xf000000000000000
+
+	syscallArgs uint8 = 6
+
+	// os.Getpagesize() in wasm will return 65536:
+	// https://cs.opensource.google/go/go/+/master:src/runtime/os_wasm.go;l=13-14?q=physPageSize&ss=go%2Fgo&start=11
+	// https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances
+	linuxPageSize uint32 = 4096
+)
+
+// TODO Find all syscalls which take a char * as argument and add them there.
+var syscallDefs = map[string][6]uint64{
+	"execve":      {useNullByteLength, 0, 0, 0, 0, 0},
+	"access":      {useNullByteLength, 0, 0, 0, 0, 0},
+	"open":        {useNullByteLength, 0, 0, 0, 0, 0},
+	"openat":      {0, useNullByteLength, 0, 0, 0, 0},
+	"mkdir":       {useNullByteLength, 0, 0, 0, 0, 0},
+	"chdir":       {useNullByteLength, 0, 0, 0, 0, 0},
+	"pivot_root":  {useNullByteLength, useNullByteLength, 0, 0, 0, 0},
+	"mount":       {useNullByteLength, useNullByteLength, useNullByteLength, 0, 0, 0},
+	"umount2":     {useNullByteLength, 0, 0, 0, 0, 0},
+	"sethostname": {useNullByteLength, 0, 0, 0, 0, 0},
+	"statfs":      {useNullByteLength, 0, 0, 0, 0, 0},
+	"stat":        {useNullByteLength, 0, 0, 0, 0, 0},
+	"statx":       {0, useNullByteLength, 0, 0, 0, 0},
+	"lstat":       {useNullByteLength, 0, 0, 0, 0, 0},
+	"fgetxattr":   {0, useNullByteLength, 0, 0, 0, 0},
+	"lgetxattr":   {useNullByteLength, useNullByteLength, 0, 0, 0, 0},
+	"getxattr":    {useNullByteLength, useNullByteLength, 0, 0, 0, 0},
+	"newfstatat":  {0, useNullByteLength, 0, 0, 0, 0},
+	"read":        {0, useRetAsParamLength | paramProbeAtExitMask, 0, 0, 0, 0},
+	"write":       {0, useArgIndexAsParamLength + 2, 0, 0, 0, 0},
+	"getcwd":      {useNullByteLength | paramProbeAtExitMask, 0, 0, 0, 0, 0},
+	"pread64":     {0, useRetAsParamLength | paramProbeAtExitMask, 0, 0, 0, 0},
+}
+
+// Used as cache for getSyscallDeclaration().
+var sysDeclarationCache map[string]api.SyscallDeclaration
+
+type eventFields struct {
+	mntnsID    api.Field
+	cpu        api.Field
+	pid        api.Field
+	comm       api.Field
+	syscall    api.Field
+	parameters api.Field
+	ret        api.Field
+}
+
+var (
+	dsOutput api.DataSource
+	fields   eventFields
+)
+
+type containerRingReader struct {
+	innerBuffer api.Map
+	perfReader  api.PerfReader
+}
+
+type tracelooper struct {
+	mapOfPerfBuffers api.Map
+
+	// key:   mntnsID
+	// value: *containerRingReader
+	readers sync.Map
+}
+
+var t tracelooper
+
+// Keep in sync with type in program.bpf.c.
+type traceloopSyscallEventContT struct {
+	EventType          eventType
+	Param              [128]uint8
+	MonotonicTimestamp uint64
+	Length             uint64
+	Index              uint8
+	Failed             uint8
+	_                  [5]byte
+}
+
+// Keep in sync with type in program.bpf.c.
+type traceloopSyscallEventT struct {
+	EventType          eventType
+	Args               [6]uint64
+	MonotonicTimestamp uint64
+	BootTimestamp      uint64
+	Pid                uint32
+	Cpu                uint16
+	Id                 uint16
+	Comm               [16]uint8
+	ContNr             uint8
+	_                  [62]byte
+}
+
+type syscallEvent struct {
+	bootTimestamp      uint64
+	monotonicTimestamp uint64
+	typ                eventType
+	contNr             uint8
+	cpu                uint16
+	id                 uint16
+	pid                uint32
+	comm               string
+	args               []uint64
+	mountNsID          uint64
+	retval             uint64
+}
+
+type syscallEventContinued struct {
+	monotonicTimestamp uint64
+	index              uint8
+	param              string
+}
+
+type syscallParam struct {
+	name    string
+	value   string
+	content *string
+}
+
+type event struct {
+	timestamp  int64
+	mountNsID  uint64
+	cpu        uint16
+	pid        uint32
+	comm       string
+	syscall    string
+	parameters []syscallParam
+	retval     string
+}
+
+func paramsToString(parameters []syscallParam) string {
+	var sb strings.Builder
+
+	for idx, p := range parameters {
+		value := p.value
+		if p.content != nil {
+			value = *p.content
+		}
+
+		fmt.Fprintf(&sb, "%s=%s", p.name, value)
+
+		if idx < len(parameters)-1 {
+			sb.WriteString(", ")
+		}
+	}
+
+	return sb.String()
+}
+
+func (t *tracelooper) attach(mntnsID uint64) error {
+	perfBufferName := fmt.Sprintf("perf_buffer_%d", mntnsID)
+
+	// 1. Create inner Map as perf buffer.
+	// Keep the spec in sync with program.bpf.c.
+	innerBuffer, err := api.NewMap(api.MapSpec{
+		Name:      perfBufferName,
+		Type:      api.PerfEventArray,
+		KeySize:   uint32(4),
+		ValueSize: uint32(4),
+	})
+	if err != nil {
+		return fmt.Errorf("creating map %s", fmt.Sprintf("perf_buffer_%d", mntnsID))
+	}
+
+	// 2. Use this inner Map to create the perf reader.
+	perfReader, err := api.NewPerfReader(innerBuffer, 64*linuxPageSize, true)
+	if err != nil {
+		innerBuffer.Close()
+
+		return fmt.Errorf("creating perf ring buffer: %w", err)
+	}
+
+	// 3. Add the inner map's file descriptor to outer map.
+	err = t.mapOfPerfBuffers.Update(mntnsID, innerBuffer, api.UpdateNoExist)
+	if err != nil {
+		innerBuffer.Close()
+		perfReader.Close()
+
+		return fmt.Errorf("adding perf buffer to map with mntnsID %d: %w", mntnsID, err)
+	}
+
+	t.readers.Store(mntnsID, &containerRingReader{
+		innerBuffer: innerBuffer,
+		perfReader:  perfReader,
+	})
+
+	return nil
+}
+
+func (t *tracelooper) detach(mntnsID uint64) error {
+	err := t.mapOfPerfBuffers.Delete(mntnsID)
+	if err != nil {
+		return fmt.Errorf("removing perf buffer from map with mntnsID %d: %w", mntnsID, err)
+	}
+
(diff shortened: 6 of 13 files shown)


gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,573 @@
...
+	 * is converted to the wall-clock time in userspace. It only works
+	 * from Linux 5.7. On older kernels, the BPF bytecode for
+	 * bpf_ktime_get_boot_ns is automatically removed by the BPF loader,
+	 * see FixBpfKtimeGetBootNs. In this way, this BPF program can still be
+	 * loaded on older kernels. */
+	u64 boot_ts = bpf_ktime_get_boot_ns();
+
+	/* The monotonic timestamp is used by traceloop to match the sys_enter
+	 * event with the cont and sys_exit events. This is an internal
+	 * implementation detail not exposed to the user. */
```
mauriciovasquezbernal, line 229: Is it possible that two syscalls get executed at the exact same time on different cpus? Should the cpu number be added to this key as well?

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,649 @@
...
+		for _, enterEvent := range enterTimestampEvents {
+			syscallName, err := api.GetSyscallName(enterEvent.id)
+			if err != nil {
+				return nil, fmt.Errorf("getting syscall name: %w", err)
+			}
+
+			event := &tracelooptypes.Event{
+				Event: eventtypes.Event{
+					Type:      eventtypes.NORMAL,
+					Timestamp: timestampFromEvent(enterEvent),
```
mauriciovasquezbernal, line 379: Any way to use the timestamp formatter instead of having to handle it again here?

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,649 @@
...
+			event.param = strconv.Quote(event.param)
+
+			_, ok := syscallContinuedEventsMap[event.monotonicTimestamp]
+			if !ok {
+				// Just create a 0 elements slice for the moment, the ContNr will be
+				// checked later.
+				syscallContinuedEventsMap[event.monotonicTimestamp] = make([]*syscallEventContinued, 0)
+			}
+
+			syscallContinuedEventsMap[event.monotonicTimestamp] = append(syscallContinuedEventsMap[event.monotonicTimestamp], event)
```
mauriciovasquezbernal, line 362:
```suggestion
			syscallContinuedEventsMap[event.monotonicTimestamp] = append(syscallContinuedEventsMap[event.monotonicTimestamp], event)
```

IIUC this should work fine. Same in similar places.

mauriciovasquezbernal (commented):
I did a quick pass. I'll go deeper while reviewing other PRs too.

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,649 @@
...
+	if errNo >= -4095 && errNo <= -1 {
+		return fmt.Sprintf("-1 (%s)", syscall.Errno(-errNo).Error())
+	}
+	return fmt.Sprintf("%d", ret)
+}
+
+func (t *tracelooper) read(reader *containerRingReader) ([]*tracelooptypes.Event, error) {
+	syscallContinuedEventsMap := make(map[uint64][]*syscallEventContinued)
+	syscallEnterEventsMap := make(map[uint64][]*syscallEvent)
+	syscallExitEventsMap := make(map[uint64][]*syscallEvent)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: What I meant is, why to have `syscallEnterEventsMap := make(map[uint64][]*syscallEvent)` instead of `syscallEnterEventsMap := make(map[uint64]*syscallEvent)`? there should be a single enter event for each timestmap, right?

gadgets/traceloop/go/program.go
mauriciovasquezbernal: If [`gadget_timestamp`](https://www.inspektor-gadget.io/docs/v0.35.0/gadget-devel/gadget-ebpf-api#gadget_timestamp) is used Inspektor Gadget will perform the uint to string conversion automatically. I this case the field isn't coming from ebpf, but I think it should work the same.

alban: `bpf_ktime_get_boot_ns()` was added in Linux 5.8 (released 4 years ago). Given that you are re-writing traceloop, I think it is fine to drop the compatibility with Linux < 5.8, and assume `bpf_ktime_get_boot_ns()` is available.

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,638 @@
...
+		if in[i] == 0 {
+			return string(in[:i])
+		}
+	}
+	return string(in[:l])
+}
+
+func wallTimeFromBootTime(ts uint64) eventtypes.Time {
+	if ts == 0 {
+		return eventtypes.Time(time.Now().UnixNano())
```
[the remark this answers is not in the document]
alban: In general, it is fine for me to add a function like `timeNow` if needed.

But maybe it is not needed if we agree to require Linux >= 5.8 for the tracelop gadget (https://github.com/inspektor-gadget/inspektor-gadget/pull/3763#discussion_r1889998903).

alban:
A symlink should be added in `docs/gadgets/` like other gadgets.

gadgets/traceloop/README.mdx
```diff
@@ -0,0 +1,219 @@
+---
+title: traceloop
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# trace_dns
```
alban, line 9: The documentation should be updated.
```suggestion
# traceloop
```

gadgets/traceloop/artifacthub-pkg.yml
```diff
@@ -0,0 +1,29 @@
+# Artifact Hub package metadata file
+version: 0.39.0
+name: "traceloop"
+category: monitoring-logging
+displayName: "traceloop"
+createdAt: "2024-10-08T09:57:54Z"
+digest: "2024-10-08T09:57:54Z"
+description: "syscalls flight recvorder"
```
alban, line 8:
```suggestion
description: "syscalls flight recorder"
```

gadgets/traceloop/gadget.yaml
```diff
@@ -0,0 +1,36 @@
...
+          description: Process ID
+          template: pid
+      comm:
+        annotations:
+          description: Process name
+          template: comm
+      syscall:
+        annotations:
+          description: Syscall
+          template: comm
```
alban, line 32: wrong template
Use something similar to audit_seccomp or trace_capabilities:
```suggestion
          columns.width: 20
```

gadgets/traceloop/gadget.yaml
```diff
@@ -0,0 +1,36 @@
...
+      ret:
+        annotations:
+          description: Syscall's return value
+          template: comm
```
alban, line 36:
```suggestion
          columns.width: 5
          columns.alignment: right
```

gadgets/traceloop/go/go.mod
```diff
@@ -0,0 +1,13 @@
+module traceloop
+
+go 1.23.0
+
+toolchain go1.23.4
+
+// Version doesn't matter because of the replace directive below.
+require github.com/inspektor-gadget/inspektor-gadget v0.0.0
+
+require golang.org/x/exp v0.0.0-20250103183323-7d7fa50e5329 // indirect
```
alban, line 10: Where is it used? Does it need `go mod tidy` to remove it?

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,788 @@
...
+	"sort"
+	"strconv"
+	"sync"
+	"syscall"
+	"time"
+	"unsafe"
+
+	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
+	tracelooptypes "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/traceloop/types"
+	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
```
alban, line 30: We should not import anything else than `wasmapi/go/` in a gadget.

* tracelooptypes is the builtin gadget. If you need something from there, it should be moved to this file directly.
* if you need something from eventtypes, it should be added to wasmapi/go/. But it seems the current usage should not be used like that: we can do without Time, and the other things belong to builtin gadgets.

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,586 @@
...
+#define __NR_exit 93
+#elif defined(__TARGET_ARCH_x86)
+#define __NR_rt_sigreturn 15
+#define __NR_exit_group 231
+#define __NR_exit 60
+#else
+#error "Traceloop is not supported on your architecture."
+#endif
+
+/* Compile with -DSHOW_DEBUG to print debug messages. */
```
alban, line 99:
```suggestion
/* Add the following in build.yaml to print debug messages:
 *  cflag: '-DSHOW_DEBUG'
  */
```

Does it still work?

Ditto below for SHOW_ERROR.

You can include a build.yaml with some comments
```yaml
## To print debug messages:
# cflag: '-DSHOW_DEBUG'
## To print error messages:
# cflag: '-DSHOW_ERROR'
```

gadgets/traceloop/program.bpf.c
alban: > Is it possible that two syscalls get executed at the exact same time on different cpus? Should the cpu number be added to this key as well?

Possible but unlikely with the nano second precision.

We cannot add the cpu in the key because of the task scheduler as @eiffel-fl said. We should add the tid in the key but this makes a bit more complicated with syscalls that can change the tid like execve on multithreaded processes.

Matching sys_enter, continuation and sys_exit events is best effort only: there are cases where it cannot be matched:
- traceloop started after a syscall or stopped before the end of a syscall.
- sys_enter overwritten (because the ring buffer is configured as overwriteable). Note that it is not necessarily the oldest event that is erased first due to each cpu having its own ring buffer, and one syscall can spread events over several ring buffers (with task migration).

So I suggest to leave it like this for now.

gadgets/traceloop/program.bpf.c
alban: This should be in `build.yaml` (and not gadget.yaml`): the `cflag:` field, once uncommented, only works in `build.yaml`.

alban (commented):
Some small changes requested, but I think the code is almost good to go. Once they are fixed and the remaining documentation is done, 

- [x] Write gadget documentation
- [ ] Update symlink in docs

gadgets/traceloop/build.yaml
```diff
@@ -0,0 +1,5 @@
+wasm: go/program.go
+# To print debug messages:
+# cflags: '-DSHOW_DEBUG'
+# To print error messages:
+# cflags: '-DSHOW_ERROR'
```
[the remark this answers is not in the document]
alban: I guess this is partially due to:
```c
	struct syscall_event_t sc = { .event_type = SYSCALL_EVENT_NORMAL };
```
with the `__u8 padding[56];`.
It should be moved away from the stack.
The trace_exec gadget uses a `static const struct event empty_event` to avoid allocating the large event on the stack.

mauriciovasquezbernal:
I'm reviewing it but before pushing all my comment I'm wondering if the current state is broken:

```bash 
$ sudo ig run traceloop --verify-image=false
[sudo] password for mauriciov:
RUNTIME.CONTAINERNAME                     CPU         PID COMM             SYSCALL                   PARAMETERS                      RET
^CWARN[0007] perfReaderRead: reading perf buffer: epoll wait: i/o timeout
```

Tried on 7b63b0048a0d32e934fa893c98bf0852d7abc5c

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,824 @@
...
+			event := &event{
+				timestamp: timestampFromEvent(enterEvent),
+				mountNsID: enterEvent.mountNsID,
+				cpu:           enterEvent.cpu,
+				pid:           enterEvent.pid,
+				comm:          enterEvent.comm,
+				syscall:       syscallName,
+			}
+
+			syscallDeclaration, err := api.GetSyscallDeclaration(event.syscall)
```
mauriciovasquezbernal, line 440: This function has a **HUGE** overhead. I tried to use a cache for it:

```diff
+var sysDeclarationCache map[string]api.SyscallDeclaration
+
+func getSyscallDeclaration(name string) (api.SyscallDeclaration, error) {
+       if sysDeclarationCache == nil {
+               sysDeclarationCache = make(map[string]api.SyscallDeclaration)
+       }
+
+       if declaration, ok := sysDeclarationCache[name]; ok {
+               return declaration, nil
+       }
+
+       declaration, err := api.GetSyscallDeclaration(name)
+       if err != nil {
+               return api.SyscallDeclaration{}, fmt.Errorf("getting syscall definition: %w", err)
+       }
+
+       sysDeclarationCache[name] = declaration
+
+       return declaration, nil
+}
+
 func (t *tracelooper) read(reader *containerRingReader) ([]*event, error) {
        syscallContinuedEventsMap := make(map[uint64][]*syscallEventContinued)
        syscallEnterEventsMap := make(map[uint64][]*syscallEvent)
@@ -431,13 +452,14 @@ func (t *tracelooper) read(reader *containerRingReader) ([]*event, error) {
-                       syscallDeclaration, err := api.GetSyscallDeclaration(event.syscall)
+                       //syscallDeclaration, err := api.GetSyscallDeclaration(event.syscall)
+                       syscallDeclaration, err := getSyscallDeclaration(event.syscall)
```

As workload I'm running `while true; do cat /dev/null; sleep 1; done` inside a container. The results are:

Without cache:

```bash 
$ time sudo ig run traceloop:latest --verify-image=false --timeout 5
real    1m0,030s
user    0m0,015s
sys     0m0,035s
```

With the cache

```bash 
$ time sudo ig run traceloop:latest --verify-image=false --timeout 5
real    0m10,117s
user    0m0,011s
sys     0m0,036s
```

(I didn't go further but ideas on #3992 could help to profile it)

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,824 @@
...
+
+func toString(parameters []syscallParam) string {
+	var sb strings.Builder
+
+	for idx, p := range parameters {
+		value := p.value
+		if p.content != nil {
+			value = *p.content
+		}
+		sb.WriteString(fmt.Sprintf("%s=%s", p.name, value))
```
mauriciovasquezbernal, line 170: perhaps this is a bit more efficient:

```suggestion
	        sb.WriteString(p.name)
                sb.WriteRune('=')
                sb.WriteString(p.value)
```

gadgets/traceloop/go/program.go
mauriciovasquezbernal, line 1: This file is not formatted. (Visual studio performs a lot of changes when saving it)

gadgets/traceloop/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+The traceloop gadget is a syscalls flight recorder.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        TODO
```
mauriciovasquezbernal, line 19:
```suggestion
        Unsupported
```

mauriciovasquezbernal (commented):
It's looking very good. Some comments here and there.

pkg/operators/wasm/wasm.go
```diff
@@ -321,7 +321,7 @@ func (i *wasmOperatorInstance) Stop(gadgetCtx operators.GadgetContext) error {
 	}()
 
 	// We need a new context in here, as gadgetCtx has already been cancelled
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1500)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Perhaps it can be changed once https://github.com/inspektor-gadget/inspektor-gadget/pull/3763#discussion_r2017196994 is fixed.

mauriciovasquezbernal:
> I'm reviewing it but before pushing all my comment I'm wondering if the current state is broken:

Sorry for the noise, I was too impatient and I was killing it as it was stuck, but then I realized after a while the results are printed (see https://github.com/inspektor-gadget/inspektor-gadget/pull/3763#discussion_r2017196994)

Why is the `perfReaderRead: reading perf buffer: epoll wait: i/o timeout` warning printed? Can we avoid it? I think it's misleading.

gadgets/traceloop/go/program.go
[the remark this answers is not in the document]
mauriciovasquezbernal: Perhaps we can use `WithWalltime` from https://pkg.go.dev/github.com/tetratelabs/wazero#section-readme instead of implementing a custom timeNow function. However we need to pay attention as the wazero documentation states that "This does not default to time.Now as that violates sandboxing.".

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,685 @@
...
+ * 0xfe (ret as param. length).
+ */
+const __u64 USE_ARG_INDEX_AS_PARAM_LENGTH = 0x0ffffffffffffff0ULL;
+const __u64 USE_ARG_INDEX_AS_PARAM_LENGTH_MASK = 0xfULL;
+
+const __u8 SYSCALL_EVENT_TYPE_ENTER = 0;
+const __u8 SYSCALL_EVENT_TYPE_EXIT = 1;
+
+const __u8 SYSCALL_EVENT_NORMAL = 1;
+const __u8 SYSCALL_EVENT_CONT = 2;
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Why can't we differentiate them with a single enum with these values?

```
SYSCALL_EVENT_ENTER
SYSCALL_EVENT_CONT
SYSCALL_EVENT_EXIT
```

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,685 @@
...
+/*
+ * syscall_event_t and syscall_event_cont_t are big structures, let store them
+ * here instead of using the stack.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(key_size, sizeof(struct fake_stack_key));
+	__uint(value_size, sizeof(struct syscall_event_t));
+	__uint(max_entries, 1024);
+} fake_stack SEC(".maps");
```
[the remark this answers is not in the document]
mauriciovasquezbernal: This is being used to hold data between the execution of the same program, not to hold data between multiple executions of programs, hence the task being migrated is not relevant.

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,685 @@
...
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(key_size, sizeof(u64));
+	__uint(value_size, sizeof(struct syscall_def_t));
+	/*
+	 * We have around 300 syscalls, let's use the immediate greater power of
+	 * 2.
+	 */
+	__uint(max_entries, 512);
```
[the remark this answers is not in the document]
mauriciovasquezbernal: hash maps are preallocated:

> Hash maps are pre-allocated by default, this means that even a completely empty hash map will use the same amount of kernel memory as a full map.

(https://github.com/isovalent/ebpf-docs/blob/master/docs/linux/map-type/BPF_MAP_TYPE_HASH.md#bpf_f_no_prealloc)

There is a way to disable it by using BPF_F_NO_PREALLOC but it can create some other issues as described in https://github.com/torvalds/linux/commit/6c90598174322b8888029e40dd84a4eb01f56afe.  (I don't know if the situation has changed on the newest kernel versions)

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,824 @@
...
+		record := make([]byte, sysEventSize)
+		if err := reader.perfReader.Read(record); err != nil {
+			if errors.Is(err, os.ErrDeadlineExceeded) {
+				break
+			} else {
+				return nil, err
+			}
+		}
+		records = append(records, record)
+	}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I meant to store the events as `syscallEvent` on the first loop directly instead of saving them as raw bytes to save some memory. The loop under the "// Let's try to publish the events we gathered." comment won't change.

gadgets/traceloop/program.bpf.c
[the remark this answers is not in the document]
alban: A task can be migrated between the sys_enter tracepoint and the sys_exit tracepoint but not in a middle of a single ebpf program. And the map "fake_stack" is used to keep data within a single ebpf program only, so there is no task migration.

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,824 @@
...
+	// 3. Add the inner map's file descriptor to outer map.
+	err = t.mapOfPerfBuffers.Update(mntnsID, innerBuffer, api.UpdateNoExist)
+	if err != nil {
+		innerBuffer.Close()
+		perfReader.Close()
+
+		return fmt.Errorf("adding perf buffer to map with mntnsID %d: %w", mntnsID, err)
+	}
+
+	t.readers.Store(cgroupID, &containerRingReader{
```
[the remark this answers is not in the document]
alban: I think we can do everything with the mntns. There is no reason why traceloop does it differently than other gadgets.

Traceloop was initially using the cgroupid because at the time, we didn't have support for CO-RE and we couldn't get the mntns from the task without assuming a specific kernel. We don't have this problem with the cgroupid because there is a helper function for it.

The other reason was to support systemd units (which use a separate cgroup but might not use a separate mntns). However, other gadgets don't support systemd units, and the current state of the traceloop PR does not support it either, so I think it is better to use mntns.

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,802 @@
...
+		if in[i] == 0 {
+			return string(in[:i])
+		}
+	}
+	return string(in[:l])
+}
+
+func wallTimeFromBootTime(ts uint64) int64 {
+	if ts == 0 {
+		return time.Now().UnixNano()
```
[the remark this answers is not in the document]
alban: But since `time.Now()` does not work in WASM, it is better to delete that code. Printing a timestamp of 1 January 1970 (Epoch) on Linux < 5.7 is better than printing 1 January 2022. The "Unix Epoch" of 1970 is more well known than the "WASM Epoch" of 2022.

Then, the "## Limitation" section in README.mdx could just say that traceloop timestamps works only on Linux >= 5.7.

blanquicet (commented):
I didn't went very in details into the code, but I tested it in different situations and it always worked as expected. My only concerns is the `perfReaderRead: reading perf buffer: epoll wait: i/o timeout` warning. Why is it always happening?

alban:
> > > I'm reviewing it but before pushing all my comment I'm wondering if the current state is broken:
> > 
> > 
> > Sorry for the noise, I was too impatient and I was killing it as it was stuck, but then I realized after a while the results are printed (see [#3763 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/pull/3763#discussion_r2017196994))
> > Why is the `perfReaderRead: reading perf buffer: epoll wait: i/o timeout` warning printed? Can we avoid it? I think it's misleading.
> 
> This comes from the following:
> 
> https://github.com/inspektor-gadget/inspektor-gadget/blob/48296c35caee65c8304ab13d7692c71e9f28c538/pkg/operators/wasm/perf.go#L176
> 
> 
> We should keep it, otherwise, we would not be able to know if something wrong occurred while reading the perf buffer.

I don't understand what this means. I just tried traceloop and I get this warning each time. What was wrong while reading the perf buffer? And what is the user supposed to do to fix the problem?

gadgets/traceloop/go/program.go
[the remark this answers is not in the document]
alban: I think the 3 lines could just be deleted
```go
if ts == 0 {
	return 0
}
```
because if `ts==0`, the line `return time.Unix...` will return the same result. And then, `wallTimeFromBootTime` can be inlined in `timestampFromEvent`.

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,835 @@
...
+	if err != nil {
+		return fmt.Errorf("removing perf buffer from map with mntnsID %d: %w", mntnsID, err)
+	}
+
+	return nil
+}
+
+func fromCString(in []byte) string {
+	for i := 0; i < len(in); i++ {
+		if in[i] == 0 {
```
alban, line 248: Could this use `idx := bytes.IndexByte(in, 0)` instead?
Ditto in `fromCStringN()`.
https://pkg.go.dev/bytes#IndexByte

(Not a blocker for this PR)

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,835 @@
...
+			return string(in[:i])
+		}
+	}
+	return string(in[:l])
+}
+
+func wallTimeFromBootTime(ts uint64) int64 {
+	if ts == 0 {
+		return 0
+	}
```
alban, line 272: Could be deleted since the line `return time.Unix...` will do the same thing.
Then, wallTimeFromBootTime could be merged in timestampFromEvent

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,835 @@
...
+// system calls indicate a failure by returning a negative error
+// number to the caller on architectures without a separate error
+// register/flag, as noted in syscall(2); when this happens, the
+// wrapper function negates the returned error number (to make it
+// positive), copies it to errno, and returns -1 to the caller of
+// the wrapper.
+func retToStr(ret uint64) string {
+	errNo := int64(ret)
+	if errNo >= -4095 && errNo <= -1 {
+		return fmt.Sprintf("-1 (%s)", syscall.Errno(-errNo).Error())
```
alban, line 337: Does this work on arm64?

(Not a blocker for this PR)

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,835 @@
...
+			api.Debugf("unknown event type: got %d, expected %d or %d", sysEvent.EventType, syscallEventTypeEnter, syscallEventTypeCont)
+		}
+	}
+
+	err = reader.perfReader.Resume()
+	if err != nil {
+		return nil, err
+	}
+
+	// Let's try to publish the events we gathered.
```
alban, line 449:
```suggestion
	// Publish the events we gathered.
```

(Not a blocker)

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,835 @@
...
+	sort.Slice(events, func(i, j int) bool {
+		return events[i].timestamp < events[j].timestamp
+	})
+
+	// Remove timestamps if we couldn't get reliable ones
+	// 	if hasBpfKtimeGetBootNs() {
+	// 		for i := range events {
+	// 			events[i].Timestamp = 0
+	// 		}
+	// 	}
```
alban, line 614: Delete this comment

gadgets/traceloop/go/program.go
```diff
@@ -0,0 +1,835 @@
...
+	}
+
+	nameField, err := ds.GetField("name")
+	if err != nil {
+		api.Errorf("Failed to get field: %v", err)
+		return 1
+	}
+
+	ds.Subscribe(func(ds api.DataSource, data api.Data) {
+		eventType, err := eventTypeField.String(data, 7)
```
alban, line 756: add a comment to explain 7.

gadgets/traceloop/program.bpf.c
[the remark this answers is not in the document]
alban: Yes, with a size of 1.

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,680 @@
...
+	if (!args) {
+		bpf_error_printk("exit: there should be a pt_regs for key %lu",
+				 pid);
+
+		goto end;
+	}
+
+	bpf_probe_read_kernel(args, sizeof(*args), (void *)ctx->args[0]);
+	nr = syscall_get_nr(args);
+	/* TODO Why this can occur? */
```
alban, line 538: Just curious, did you see it happening?

alban (approved):
I tested it and it worked for me! Thanks.

I still have a few comments, but LGTM after that (if the test passes).

gadgets/traceloop/program.bpf.c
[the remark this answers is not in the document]
alban: Since the bpf program is not preempted and there is no reentrance, each cpu can only have one execution at a time. So the typical code pattern is to use a per-cpu array of size 1. We would only access the cell at index zero. Example:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_dns/program.bpf.c#L134

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,685 @@
...
+	 */
+	if (syscall_def == NULL)
+		syscall_def = &default_definition;
+
+	task = (struct task_struct *)bpf_get_current_task();
+	mntns_id = (u64)BPF_CORE_READ(task, nsproxy, mnt_ns, ns.inum);
+
+	perf_buffer = bpf_map_lookup_elem(&map_of_perf_buffers, &mntns_id);
+	if (!perf_buffer)
+		goto clean;
```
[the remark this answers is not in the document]
mauriciovasquezbernal: The 

```c
	task = (struct task_struct *)bpf_get_current_task(); 
	mntns_id = (u64)BPF_CORE_READ(task, nsproxy, mnt_ns, ns.inum);
```

should be moved before too, otherwise mntns_id will be always 0 for the map lookup.

gadgets/traceloop/program.bpf.c
```diff
@@ -0,0 +1,685 @@
...
+				ret);
+		}
+
+		bpf_map_delete_elem(&fake_stack, &event_cont_key);
+	}
+
+end:
+	bpf_map_delete_elem(&regs_map, &pid);
+clean:
+	bpf_map_delete_elem(&fake_stack, &event_key);
```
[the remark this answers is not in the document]
mauriciovasquezbernal: The whole idea of this comment was to say that there is not any reason to use a hashmap for the "fake stack", it can be a percpu array with 1 element. Some reorganization of the code could be required, but I don't remember the details.

PR #3797, opened 2024-12-13
ebpf: Do not use deprecated RewriteConstants function

mauriciovasquezbernal:


alban:
More `//nolint:staticcheck` need to be reverted from #3595.

pkg/container-hook/tracer.go
```diff
@@ -257,12 +257,13 @@ func (n *ContainerNotifier) installEbpf(fanotifyFd int) error {
...
-		"tracer_group": fanotifyPrivateData,
+	execSpec := &execruntimeSpecs{}
+	if err := spec.Assign(&execSpec); err != nil {
+		return err
 	}
-	//nolint:staticcheck
-	if err := spec.RewriteConstants(consts); err != nil {
-		return fmt.Errorf("RewriteConstants: %w", err)
+
+	if err := execSpec.TracerGroup.Set(fanotifyPrivateData); err != nil {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: `execruntimeSpecs` is just a structure containing the specs, it doesn't provide methods like LoadAndAssign. TBH I'm not sure how it's designed in that way, IMO it would make more sense for `loadExecruntime` to return `execruntimeSpecs` instead of `*ebpf.CollectionSpec`, but there is probably a strong good reason not to follow that approach.

pkg/utils/ebpf/ebpf.go
```diff
@@ -0,0 +1,43 @@
...
+package ebpfutils
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/cilium/ebpf"
+)
+
+func SpecSetVars(spec *ebpf.CollectionSpec, values map[string]interface{}) error {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it's more robust to rely on the code generated by bpf2go instead of passing the names of the constants as arguments to a function.

pkg/gadgets/profile/tcprtt/tracer/tracer.go
```diff
@@ -236,22 +236,27 @@ func (t *Tracer) install() error {
...
+		tcpRttSpec.TargDaddrV6:   t.config.filterRemoteAddressV6,
 	}
 
-	//nolint:staticcheck
-	if err := spec.RewriteConstants(consts); err != nil {
-		return fmt.Errorf("rewriting constants: %w", err)
+	for varSpec, val := range consts {
+		if err := varSpec.Set(val); err != nil {
+			return fmt.Errorf("setting variable %s: %w", varSpec, err)
+		}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: `SpecSetVars` should be used when the caller doesn't have a struct generated with bpf2go (in the ebpf operator for image based gadgets for instance). It's not possible to use that helper since it relies on the const name, instead the logic here relies on the ebpf.VariableSpec fields on the collection spec.

To extend a bit more on this, I tried to use the new `VariableSpec` introduced in bpf2go, but I can revert it back to only relying on the const name too.

mauriciovasquezbernal:
Can I get another review (and hopefully approval) on this please?

pkg/operators/ebpf/ebpf.go
```diff
@@ -639,13 +644,11 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
 			}
 			i.logger.Debugf("setting var %q to %v", v.name, t)
-			constReplacements[v.name] = res
-		}
-	}
 
-	//nolint:staticcheck
-	if err := i.collectionSpec.RewriteConstants(constReplacements); err != nil {
-		return fmt.Errorf("rewriting constants: %w", err)
+			if err := ebpfutils.SpecSetVar(i.collectionSpec, v.name, res); err != nil {
```
alban, line 648: `ebpfutils.SpecSetVar` is called twice in this file: for gadget parameters and for ebpf instance variables. Could it be called twice for the same cilium/ebpf variable? If so, a comment explaining the precedence we've chosen would help.

pkg/utils/ebpf/ebpf.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2024 The Inspektor Gadget authors
```
alban, line 1:
```suggestion
// Copyright 2025 The Inspektor Gadget authors
```

alban (approved):
LGTM, thanks!

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: The second call is done to set some internal variables like enabling filtering by mount ns:

 https://github.com/inspektor-gadget/inspektor-gadget/blob/8912ea06de042187f4a125402694dd17b2ed7edd/pkg/operators/localmanager/localmanager.go#L610

Unless there is a mistake, SpecSetVar won't be called twice for the same variable, since params are defined by users with the GADGET_PARAM macro, it shouldn't overlap with "internal variables".

pkg/utils/ebpf/ebpf.go
mauriciovasquezbernal: I'll skip it to avoid a whole CI run again.

PR #3806, opened 2024-12-16
dns: don't send full packet to userspace

alban:
# dns: don't send full packet to userspace

The code was meant to only send 9kB of the packet to userspace, but it was sending the full packet instead.

## How to use

No changes

## Testing done

- [x] builtin gadget
- [x] image-based gadget


mauriciovasquezbernal (approved):
LGTM. Thanks for the fix!

alban:
The CI failure is unreleated. See #3809.

alban:
Related issue: https://github.com/cilium/ebpf/issues/1633

PR #3807, opened 2024-12-16
oci: remote manifest file

mqasimsarfraz:
This change allows getting manifest file remotely. It should help us host sample manifest files and share them with different use-cases:

```
$ go run ./cmd/kubectl-gadget run -f https://gist.githubusercontent.com/mqasimsarfraz/dcc1521afebdcc5708fb16efb9de2677/raw/f5b5a3d2737b44142c9f52792b170c8415a3e057/failed-dns-requests.yaml
K8S.PODNAME                                           SRC                                                  DST                                                  NAME                                                  RCODE
```

### DIFF
--- a/cmd/common/utils/http.go
+++ b/cmd/common/utils/http.go
@@ -1,0 +1,48 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+func DownloadFile(rawURL string) (io.ReadCloser, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing URL: %w", err)
+	}
+
+	client := http.Client{
+		Timeout: 10 * time.Second,
+	}
+
+	switch u.Scheme {
+	case "http", "https":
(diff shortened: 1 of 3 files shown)


cmd/common/utils/http.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func DownloadFile(rawURL string) (io.ReadCloser, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing URL: %w", err)
+	}
+
+	client := http.Client{
+		Timeout: 2 * time.Second,
```
flyth, line 32: I think 2 is too low here for slow connections, maybe more like 10-15 seconds?

cmd/common/utils/http.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	switch u.Scheme {
+	case "http", "https":
+		resp, err := client.Get(rawURL)
+		if err != nil {
+			return nil, fmt.Errorf("downloading file: %w", err)
```
flyth, line 39: I think returning `err` is enough, since the message is already good enough on the consumer and this func is always about downloading.

cmd/common/utils/http.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("downloading file: %s", resp.Status)
```
flyth, line 42: Maybe something like "unexpected status code: %s"?

flyth (commented):
I like this a lot for the UX!

Could you please adjust the help information for `-f` to include that it also can download files?

mauriciovasquezbernal (commented):
Can we implement a test for it?

cmd/common/utils/http.go
mqasimsarfraz: Done

mqasimsarfraz:
> Can we implement a test for it?

Good idea! I updated the PR to test with remote manifest file as well!

mauriciovasquezbernal (approved):
LGTM. Thanks for this nice UI improvement. 

btw, some documentation on https://www.inspektor-gadget.io/docs/latest/reference/manifests will be appreciated.

mqasimsarfraz:
> btw, some documentation on https://www.inspektor-gadget.io/docs/latest/reference/manifests will be appreciated.

I want to include certain sample manifest files before reworking the documentation. I will open a separate PR for this!

PR #3809, opened 2024-12-17
treewide: Format eBPF code using clang-format-18.

eiffel-fl:
CI uses clang-format-18 to check that eBPF code is correctly formatted [1]. Let's use the same to fix errors discovered by CI.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/12349927639/job/34461748991?pr=3803#step:3:151
[1]: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/12349927639/job/34461748991?pr=3803#step:3:91

### DIFF
--- a/gadgets/top_file/stat.h
+++ b/gadgets/top_file/stat.h
@@ -16,12 +16,12 @@
 #define S_ISGID 0002000
 #define S_ISVTX 0001000
 
-#define S_ISLNK(m) (((m)&S_IFMT) == S_IFLNK)
-#define S_ISREG(m) (((m)&S_IFMT) == S_IFREG)
-#define S_ISDIR(m) (((m)&S_IFMT) == S_IFDIR)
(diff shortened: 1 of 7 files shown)


burak-ok:
> CI uses clang-format-18 to check that eBPF code is correctly formatted [1]

They are currently switching `ubuntu:latest` from 22 to 24. So in the meanwhile we sometimes get the old clang version (when running on 22) and sometimes the new clang version (when running on 24).
The rollout will be complete on January 17th according to https://github.com/actions/runner-images/issues/10636

To always have the same version on local and CI we could run `clang-format` in a docker container like we do for `make lint`

alban:
We can use `runs-on: ubuntu-24.04` in the yaml files to stick to that specific version.
See:
```bash
$ git grep runs-on .github/
```

mauriciovasquezbernal (approved):
Approving it to unblock the CI for now. But we need to guarantee this version is stable by implementing any of the idea discussed above.

PR #3815, opened 2024-12-18
operators/wasm: Add support for array/hash of maps.

eiffel-fl:


pkg/operators/wasm/maps.go
```diff
@@ -76,6 +78,10 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
 	)
 }
 
+// Keep in sync with wasmapi/go/map.go.
+// Get the next free flag after UpdateLock
+var getFdFlag = uint64(1 << (bits.TrailingZeros64(uint64(ebpf.UpdateLock)) + 1))
```
alban, line 83: What about keeping space for potential new upstream flags after `BPF_F_LOCK`?
```suggestion
var getFdFlag = uint64(1 << 32)
```

pkg/operators/wasm/maps.go
```diff
@@ -236,7 +245,19 @@ func (i *wasmOperatorInstance) mapUpdate(ctx context.Context, m wapi.Module, sta
...
-	err = ebpfMap.Update(key, value, ebpfFlags)
+	if getFD {
+		targetMapHandle := binary.NativeEndian.Uint32(value)
+		targetMap, ok := getHandle[*ebpf.Map](i, targetMapHandle)
+		if !ok {
+			stack[0] = 1
+			return
+		}
+
+		err = ebpfMap.Update(key, uint32(targetMap.FD()), ebpfFlags)
```
alban, line 256: Could you check if `ebpfMap` is a map that accepts a fd? (`BPF_MAP_TYPE_PROG_ARRAY`, `BPF_MAP_TYPE_ARRAY_OF_MAPS`, `BPF_MAP_TYPE_HASH_OF_MAPS`, ...)

Otherwise, if the `ebpfMap` is a regular `BPF_MAP_TYPE_HASH`, this would provide a way for the wasm module to discover the fd of maps.

Additionally, the check will be helpful for gadget author to debug problems.

Similar in the "else" case, please check that `ebpfMap` does not accept a fd.

pkg/operators/wasm/maps.go
```diff
@@ -209,6 +225,9 @@ func (i *wasmOperatorInstance) mapUpdate(ctx context.Context, m wapi.Module, sta
 	valuePtr := stack[2]
 	flags := stack[3]
 
+	getFD := (flags & getFdFlag) != 0
+	flags &= ^getFdFlag
```
burak-ok, line 229: I really got confused by the `^`. TIL in golang bitwise not is `^`.
Apparently you can even do
```suggestion
	flags &^= getFdFlag
```

pkg/operators/wasm/testdata/mapofmap/program.go
```diff
@@ -0,0 +1,109 @@
...
+		ValueSize:  4,
+		MaxEntries: 1,
+	})
+	if err != nil {
+		api.Errorf("creating map %s", mapName)
+		return 1
+	}
+	defer hashMap.Close()
+
+	err = mapOfMap.Update(key, hashMap, api.GetFdFlag)
```
burak-ok, line 63: The naming of `GetFdFlag` seems confusing in this sitation here. We are updating an entry in a map, but the naming suggests we are getting a FD.

TBH I also don't know what to name that flag. Something more generic like `UseFdFlag` or so

pkg/operators/wasm/maps.go
```diff
@@ -195,6 +201,16 @@ func (i *wasmOperatorInstance) mapLookup(ctx context.Context, m wapi.Module, sta
...
 }
 
+func mapAcceptsFD(m *ebpf.Map) bool {
+	mapType := m.Type()
+	for _, typ := range mapTypesAcceptingFD {
+		if mapType == typ {
+			return true
+		}
+	}
+	return false
```
mauriciovasquezbernal, line 211:
```suggestion
	return slices.Contains(mapTypesAcceptingFD, m.Type())
```

pkg/operators/wasm/maps.go
```diff
@@ -76,6 +77,11 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
 	)
 }
 
+// Keep in sync with wasmapi/go/map.go.
+var getFdFlag = uint64(1 << 63)
```
mauriciovasquezbernal, line 81:
```suggestion
const getFdFlag = uint64(1 << 63)
```

pkg/operators/wasm/maps.go
```diff
@@ -209,6 +225,9 @@ func (i *wasmOperatorInstance) mapUpdate(ctx context.Context, m wapi.Module, sta
 	valuePtr := stack[2]
 	flags := stack[3]
 
+	getFD := (flags & getFdFlag) != 0
+	flags &^= getFdFlag
```
mauriciovasquezbernal, line 229: What about defining a mask? first 32 bits are bpf() syscall flags, upper 32 bits are used by Inspektor Gadget.

pkg/operators/wasm/maps.go
mauriciovasquezbernal: Given that it's heavily used (one invocation for each map update), I'm wondering if it makes sense to implement by using a switch statement instead.

wasmapi/go/map.go
```diff
@@ -53,6 +53,10 @@ const (
 	UpdateLock
 )
 
+// nolint:all
+// Keep in sync with pkg/operators/wasm/maps.go
+var UseFdFlag = MapUpdateFlags(1 << 63)
```
mauriciovasquezbernal, line 58: - `const`
- Put a quick comment indicating what it's about

pkg/operators/wasm/maps.go
```diff
@@ -236,7 +255,31 @@ func (i *wasmOperatorInstance) mapUpdate(ctx context.Context, m wapi.Module, sta
...
+		err = ebpfMap.Update(key, uint32(targetMap.FD()), ebpfFlags)
+	} else {
+		if mapAcceptsFD(ebpfMap) {
+			i.logger.Warnf("mapUpdate: map type %v accepts FD, expecting any map except one in: %v", ebpfMap.Type(), mapTypesAcceptingFD)
+			stack[0] = 1
+			return
+		}
+
+		err = ebpfMap.Update(key, value, ebpfFlags)
+	}
```
mauriciovasquezbernal, line 281: At this point I'm really wondering if we need the getFD flag. Can't we just imply it from the map type? If trying to update a "mapAcceptsFD" map, then treat the value as a handle. I know I suggested the flag before, but if we're going to implement it in this way, I think we can avoid it. 

Is there anything I am missing?

pkg/operators/wasm/testdata/mapofmap/program.go
```diff
@@ -0,0 +1,100 @@
...
+		_ [3]int8
+	}
+
+	mapOfMapName := "map_of_map"
+	key := map_test_struct{a: 42, b: 42, c: 43}
+	var val int32
+
+	mapOfMap, err := api.GetMap(mapOfMapName)
+	if err != nil {
+		api.Errorf("%s map exists", mapOfMapName)
```
mauriciovasquezbernal, line 36:
```suggestion
		api.Errorf("%s map doesn't exist", mapOfMapName)
```

pkg/operators/wasm/testdata/mapofmap/program.go
```diff
@@ -0,0 +1,100 @@
...
+	err = mapOfMap.Lookup(key, &val)
+	if err != nil {
+		api.Errorf("no value found for key %v in %s", key, mapOfMapName)
+		return 1
+	}
+
+	if val == 0 {
+		api.Errorf("expected FD value to be different than 0")
+		return 1
+	}
```
mauriciovasquezbernal, line 75: Should it return the **handle** for that map instead of an FD?

mauriciovasquezbernal (commented):
My main comment is about the need of the getFD map. I'll make a further pass once we make a decision on that.

pkg/operators/wasm/maps.go
```diff
@@ -76,6 +77,9 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
 	)
 }
 
+// Keep in sync with wasmapi/go/map.go.
+const getFdFlag = uint64(1 << 63)
```
burak-ok, line 81:
```suggestion
const useFdFlag = uint64(1 << 63)
```

pkg/operators/wasm/maps.go
```diff
@@ -236,7 +254,33 @@ func (i *wasmOperatorInstance) mapUpdate(ctx context.Context, m wapi.Module, sta
 		return
 	}
 
-	err = ebpfMap.Update(key, value, ebpfFlags)
+	mapTypesAcceptingFD := []ebpf.MapType{ebpf.ArrayOfMaps, ebpf.HashOfMaps}
+
+	if getFD {
```
burak-ok, line 259: General question: If we already know when a map expects a FD for the value and return an error when the user doesn't set the flag right:
Why do we have the flag at all? Can't we make this distinction automatic?

pkg/operators/wasm/testdata/mapofmap/program.go
burak-ok: I am also in favor of abstracting as much as possible for the WASM guest. Also for the lookup after the lookup as @mauriciovasquezbernal mentioned.

The WASM user shouldn't have to think about if he now has a handle or a fd

pkg/operators/wasm/maps.go
alban: Removing the getFD flag is fine for me.

There are other map kinds that take a fd, such as BPF_MAP_TYPE_PERF_EVENT_ARRAY, BPF_MAP_TYPE_PROG_ARRAY, BPF_MAP_TYPE_CGROUP_ARRAY, etc. The name of the function `mapAcceptsFD` does not reflect that.

But those don't accept the fd of a bpf map, but a fd of different types.

We don't have any wasm function to get a fd from a cgroup, so we don't support BPF_MAP_TYPE_CGROUP_ARRAY. We need to return an error if a wasm program attemps to access it, regardless of the getFD flag.

The function `mapAcceptsFD` needs to be replaced by something that tells if the map:
- Is unsupported (e.g. BPF_MAP_TYPE_CGROUP_ARRAY), so it can't be used from wasm
- Is supported and takes the fd of a bpf map
- Is supported and takes a regular buffer

It seems to me it cannot just be a function returning a `bool`.

Maps of unknown types (from newer Linux releases) should be unsupported by default.

pkg/operators/wasm/maps.go
```diff
@@ -76,6 +78,35 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
...
+	// A newer kernel can introduce a new type of map taking an FD as a value,
+	// this way user could use this new map to leak FD in guest WASM.
+	if !isMapTypeKnown(typ) {
+		return fmt.Errorf("unknown map type %v", typ)
+	}
+
+	// We do not support map taking FD different than map FD to avoid leaking them
+	// in guest wasm.
+	if isMapTypeFDArray(typ) {
+		return fmt.Errorf("FD array maps like %v are not supported", typ)
```
alban, line 104: With the current implementation, it will not error out for CgroupArray (it is commented out).

The name is confusing. Why not the following:
```go
	// Reject unknown maps
	if !isMapTypeKnown(typ) {
		return fmt.Errorf("unknown map type %v", typ)
	}
	// Maps containing maps are supported
	if isMapTypeFDMap(typ) {
		return nil
	}
	// Maps containing a fd of a different types are not supported
	if isMapTypeFD(typ) {
		return fmt.Errorf("map type %v unsupported", typ)
	}
	// Other regular maps are supported
	return nil
}
```

In this way, `isMapTypeFD` could be written like this:
```go
// isMapTypeFD indicates if the map contains a fd, as listed by the macro IS_FD_MAP:
// https://github.com/torvalds/linux/blob/v6.12/kernel/bpf/syscall.c#L43-L49
func isMapTypeFD(typ ebpf.MapType) bool {
	return typ == ebpf.ProgramArray ||
		typ == ebpf.PerfEventArray/ ||
		typ == ebpf.CgroupArray ||
		typ == ebpf.ArrayOfMaps ||
		typ == ebpf.HashOfMaps
}
```

pkg/operators/wasm/testdata/mapofmap/program.go
mauriciovasquezbernal: It's not only about "hiding" the map FD, but allowing users to act on those inner maps. For instance, how to update an element on a map of a maps. This requires to perform the outer lookup to get the inner map and then call update on it. With the current approach that's not possible as the return value of the first lookup is the inner map ID, and not its handle. (And we don't provide a helper to get a map by its ID)

I see two options:

1. Automatically cast the innerMap ID to a handle on the lookup operation, returning a handle to the inner 

```go
var innerMap api.Map 

// some logic to get outerMap
....

// get inner map
outerMap.Lookup(outerKey, &innerMap)

// update element on inner map
innerMap.Update(innerKey, foo)
```

2. Provide a way to get a map from its ID. As it's done on https://github.com/cilium/ebpf//blob/88883cf26fbcd11381b26c5b6ad1dfec225de3bf/examples/map_in_map/main.go#L100-L102

I think since we already have the handle abstraction, we can avoid the extra work for the user an implement the first option. WDYT?

alban: TIL: a lookup from userspace does not install the fd in the fd table of the process, but merely returns the id.

@mauriciovasquezbernal I think either option is fine. If we choose the first option, we might have to implement the wasm helper from the second option later on anyway for other use cases, but we don't have to do it now.

pkg/operators/wasm/testdata/mapofmap/program.go
```diff
@@ -0,0 +1,111 @@
...
+	})
+	if err != nil {
+		api.Errorf("creating map %s", hashMapName)
+		return 1
+	}
+	defer hashMap.Close()
+
+	err = mapOfMap.Put(key, hashMap)
+	if err != nil {
+		api.Errorf("setting %s FD value for key %v in %s: %v", hashMapName, key, mapOfMapName, err)
```
mauriciovasquezbernal, line 56:
```suggestion
		api.Errorf("setting %s inner map value for key %v in %s: %v", hashMapName, key, mapOfMapName, err)
```

or 

```suggestion
		api.Errorf("setting %s value for key %v in %s: %v", hashMapName, key, mapOfMapName, err)
```

pkg/operators/wasm/testdata/mapofmap/program.go
mauriciovasquezbernal, line 1: What about the following test:

```diff
t$ git diff
diff --git a/pkg/operators/wasm/testdata/mapofmap.tar b/pkg/operators/wasm/testdata/mapofmap.tar
index fb8fbeaee..111d36f96 100644
Binary files a/pkg/operators/wasm/testdata/mapofmap.tar and b/pkg/operators/wasm/testdata/mapofmap.tar differ
diff --git a/pkg/operators/wasm/testdata/mapofmap/program.go b/pkg/operators/wasm/testdata/mapofmap/program.go
index ddd88e60d..5041c3caf 100644
--- a/pkg/operators/wasm/testdata/mapofmap/program.go
+++ b/pkg/operators/wasm/testdata/mapofmap/program.go
@@ -79,6 +79,20 @@ func gadgetStart() int {
                return 1
        }

+       // test that getting an inner map and using the returned handle works
+       var innerMap2 api.Map
+       err = mapOfMap.Lookup(key, &innerMap2)
+       if err != nil {
+               api.Errorf("no value found for key %v in %s", key, mapOfMapName)
+               return 1
+       }
+
+       err = innerMap2.Put(uint32(42), uint32(43))
+       if err != nil {
+               api.Errorf("putting value in inner map %s: %v", hashMapName, err)
+               return 1
+       }
+
        arrayMapName := "test_array"
        arrayMap, err := api.NewMap(api.MapSpec{
                Name:       arrayMapName,
```

pkg/operators/wasm/testdata/mapofmap/program.go
```diff
@@ -0,0 +1,111 @@
...
+
+	if uint32(innerMap) == 0 {
+		api.Errorf("expected handle to be different than 0")
+		return 1
+	}
+
+	if uint32(innerMap) == uint32(hashMap) {
+		api.Errorf("expected handle to be different than hashMap")
+		return 1
+	}
```
mauriciovasquezbernal, line 74: What's the purpose of this check?

pkg/operators/wasm/testdata/mapofmap/program.go
```diff
@@ -0,0 +1,111 @@
...
+		api.Errorf("creating map %s", arrayMapName)
+		return 1
+	}
+	defer arrayMap.Close()
+
+	err = mapOfMap.Put(key, arrayMap)
+	if err == nil {
+		api.Errorf("should not be able setting %s FD value for key %v in %s", arrayMapName, key, mapOfMapName)
+		return 1
+	}
```
mauriciovasquezbernal, line 100: I'm not sure about this. In the end it's not testing our implementation but the kernel.

pkg/operators/wasm/maps.go
```diff
@@ -76,6 +78,35 @@ func (i *wasmOperatorInstance) addMapFuncs(env wazero.HostModuleBuilder) {
 	)
 }
 
+func isMapTypeKnown(typ ebpf.MapType) bool {
+	return typ <= ebpf.TaskStorage
+}
```
mauriciovasquezbernal, line 83: nit: add comment to keep it aligned with wasmapi/go/map.go? 

I was checking why only ebpf.TaskStorage given that new types maps were added in https://github.com/cilium/ebpf/blob/v0.17.1/types.go, I found out it's because on wasmapi/go/map.go we only have up to TaskStorage.

mauriciovasquezbernal (commented):
The implementation looks good to me besides some small nits. 

> I went with 1., caller has to release the map handle obtained from mapLookup().
I will update the documentation if we agree.

Please be sure to include a note indicating a lookup on a map of maps returns a handle to the inner map and that this handle needs to be closed. (In oposition to the normal behaviour in user space that returns the inner map ID).

pkg/operators/wasm/maps.go
```diff
@@ -185,6 +230,19 @@ func (i *wasmOperatorInstance) mapLookup(ctx context.Context, m wapi.Module, sta
...
+		mapID := ebpf.MapID(binary.NativeEndian.Uint32(value))
+
+		innerMap, err := ebpf.NewMapFromID(mapID)
+		if err != nil {
+			i.logger.Warnf("mapLookup: getting inner map from ID %d: %v", mapID, err)
+			stack[0] = 1
+			return
+		}
+
+		binary.NativeEndian.PutUint32(value, i.addHandle(innerMap))
```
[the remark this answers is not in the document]
mauriciovasquezbernal: @eiffel-fl I have no idea how you generate these links but they don't work for me.

image

pkg/operators/wasm/testdata/mapofmap/program.go
mauriciovasquezbernal: You're right, I thought you were using the same handle returned by `api.NewMap`, sorry about the noise.

pkg/operators/wasm/testdata/mapofmap/program.go
mauriciovasquezbernal: Right, I was confused as above.

pkg/operators/wasm/maps.go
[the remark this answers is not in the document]
mauriciovasquezbernal: In any case, you're right, I tried to refactor it to avoid using doing the conversion but it's a mess as we need []bytes to return the value to the caller in any case.

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -419,7 +421,7 @@ Parameters:
 - `m` (u32): Map handle (as returned by `getMap`)
 - `keyptr` (u64): A `bufPtr` to data corresponding to key.
 - `valueptr` (u64): A `bufPtr` to data corresponding to value.
-- `flags` (u64): A set of flags used to modify the update behavior. Correct values are documented [here](https://github.com/cilium/ebpf/blob/061e86d8f5e9/map.go#L790-L801).
+- `flags` (u64): A set of flags used to modify the update behavior. Correct values are documented [here](https://github.com/cilium/ebpf/blob/061e86d8f5e9/map.go#L790-L801). The additional value UseFdFlag can be used to store the FD of the map handle given as `valueptr`.
```
mauriciovasquezbernal, line 424: remove the reference to UseFdFlag

mauriciovasquezbernal (approved):
One final minor nit, LGTM already.

Issue #3767, opened 2024-12-08
Exepath field of trace exec (both built-in and image-based) gadget may be empty


Sefi4:
## Description

There are circumstances where exepath may be empty. One scenario (suggested by @alban) is where a file without a Shebang is being executed, thus `execve` isn't triggered and exepath populated with an empty string. I thought it may be possible to make sure exepath is always populated, by using args[0] only in the case where exepath is empty. To overcome the possibility of args[0] being relative, I thought to use `absPath := filepath.Clean(filepath.Join(cwd, args[0]))`. And to solve args[0] being symlink I though using `secureExe, err := securejoin.SecureJoin("/proc/<containerPid>/root/", absPath)` where containerPid is the pid of the container (which was added recently). Then, exepath is being populated from secureExe (after trimming the container root fs path).

Do you guys think this logic is valid? If so, do you think it's something worth merging to the trace exec gadgets (both image and built-in) or exepath should be exactly `task->mm->exe_file->f_path`?

alban:
For more context:
```
$ echo 'cat /etc/hostname' > script.sh
$ chmod +x script.sh
$ strace -f -e execve,openat sh -c ./script.sh
execve("/usr/bin/sh", ["sh", "-c", "./script.sh"], 0x7fff8b5e4698 /* 76 vars */) = 0
...
execve("./script.sh", ["./script.sh"], 0x562082406800 /* 76 vars */) = -1 ENOEXEC (Exec format error)
openat(AT_FDCWD, "./script.sh", O_RDONLY) = 3
...
[pid 94206] execve("/usr/bin/cat", ["cat", "/etc/hostname"], 0x562082406800 /* 76 vars */) = 0
```

Notice that `execve` returns `ENOEXEC` because the script does not have a shebang. Then, the shell fallbacks on opening the script with `openat` and interpret it without `execve`.

The failed execve is captured by Inspektor Gadget with ret=-8:

```
$ sudo ig trace exec --host --paths -o columns=comm,pcomm,ret,args,exepath
COMM             PCOMM            RET ARGS                                     EXEPATH
strace           bash             0   /usr/bin/strace -f -e execve,openat sh … /usr/bin/strace
sh               strace           0   /usr/bin/sh -c ./script.sh               /usr/bin/bash
sh               strace           -8  ./script.sh
cat              sh               0   /usr/bin/cat /etc/hostname               /usr/bin/cat
```

This issue is about adding the exepath when execve fails.

[...]

PR #3818, opened 2024-12-18
trace exec: add 'file' field for scripts

alban:
For scripts, the 'exepath' and 'file' fields are different. For example:
- exepath: "/bin/sh"
- file: "/tmp/script.sh"

'exepath' and 'file' are always set when the execution succeeds. When the execution fails, it is best effort, depending on the error. For now, they are set when the execution of a script fails because the schebang is missing.

PR filed in discussion with @Sefi4.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/3767

## How to use

```
$ cd /tmp
$ cat ./script.sh 
cat /etc/hostname
$ cat ./script2.sh 
#!/bin/bash
cat /etc/hostname
$ strace -f -e execve,openat sh -c ./script.sh
$ strace -f -e execve,openat sh -c ./script2.sh
```

```
$ sudo ./ig trace exec --host --paths -o columns=comm,pcomm,ret,args,exepath,file -F 'file:~/tmp'
COMM       PCOMM  RET ARGS         EXEPATH        FILE                                                              
sh         strace -8  ./script.sh                 /tmp/script.sh
script2.sh strace 0   ./script2.sh /usr/bin/bash  /tmp/script2.sh
```
 
## Testing done

TODO


gadgets/trace_exec/program.bpf.c
```diff
@@ -252,4 +253,31 @@ int ig_execve_x(struct syscall_trace_exit *ctx)
...
+	struct path f_path;
+
+	if (!paths)
+		return 0;
+
+	event = bpf_map_lookup_elem(&execs, &pid);
+	if (!event)
+		return 0;
+
+	// security_bprm_check is called repeatively following the shebang
```
Sefi4, line 271: A small typpo
```suggestion
	// security_bprm_check is called repeatedly following the shebang
```

gadgets/trace_exec/program.bpf.c
```diff
@@ -252,4 +253,31 @@ int ig_execve_x(struct syscall_trace_exit *ctx)
...
 
+SEC("kprobe/security_bprm_check")
+int BPF_KPROBE(security_bprm_check, struct linux_binprm *bprm)
+{
+	u32 pid = (u32)bpf_get_current_pid_tgid();
+	struct event *event;
+	char *file;
+	struct path f_path;
+
+	if (!paths)
```
Sefi4, line 264: Could be moved to be before variable declarations I guess

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
```diff
@@ -261,4 +261,30 @@ int ig_execve_x(struct syscall_trace_exit *ctx)
 	return 0;
 }
 
+SEC("kprobe/security_bprm_check")
+int BPF_KPROBE(security_bprm_check, struct linux_binprm *bprm)
+{
+#ifdef WITH_LONG_PATHS
```
Sefi4, line 267: Does it make sense to wrap the whole function with the WITH_LONG_PATHS ifdef? I think it would be more efficient

Sefi4: Though I'm not sure how we can handle the link part in tracer.go if we'll wrap the whole function

alban: Unfortunately, tracer.go relies on `execsnoopObjects` and `execsnoopWithLongPathsObjects` having the same bpf programs and maps. It would require more refactoring if we want to do that, so I would prefer to postpone that to another PR.

gadgets/trace_exec/program.bpf.c
```diff
@@ -252,4 +253,31 @@ int ig_execve_x(struct syscall_trace_exit *ctx)
...
+		return 0;
+
+	event = bpf_map_lookup_elem(&execs, &pid);
+	if (!event)
+		return 0;
+
+	// security_bprm_check is called repeatively following the shebang
+	// Only get the first call.
+	if (event->file[0] != '\0')
```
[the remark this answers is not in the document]
alban: This is for chaining shebangs. With the following example:
* `/tmp/script1.sh`:
```bash
#!/tmp/script2.sh
```
* `/tmp/script2.sh`
```bash
#!/bin/sh
```
The kprobe will be called 3 times:
1. `/tmp/script1.sh`
2. `/tmp/script2.sh`
3. `/bin/sh`

If we only take the last run, we will pick `/bin/sh`, which is the same as exepath, so that would not be useful.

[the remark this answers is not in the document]
Sefi4: Ah I see, thanks!

gadgets/trace_exec/program.bpf.c
```diff
@@ -252,4 +253,31 @@ int ig_execve_x(struct syscall_trace_exit *ctx)
...
+	event = bpf_map_lookup_elem(&execs, &pid);
+	if (!event)
+		return 0;
+
+	// security_bprm_check is called repeatedly following the shebang
+	// Only get the first call.
+	if (event->file[0] != '\0')
+		return 0;
+
+	f_path = BPF_CORE_READ(bprm, file, f_path);
```
Sefi4, line 276: Can we get `f_inode` from `bprm->file` and use the `has_upper_layer` function to improve the limitations around upper layer in the scenario of a script?

Doesn't have to be in this PR, just a thought

Sefi4: Opened this for tracking [[RFE] Improve trace exec upper layer limitation in case of a script · Issue #3838 · inspektor-gadget/inspektor-gadget](https://github.com/inspektor-gadget/inspektor-gadget/issues/3838)

mauriciovasquezbernal (approved):
Approving and merging it. We can handle further improvements in other PRs.

PR #3819, opened 2024-12-18
build: Make ebpf layer optional

mauriciovasquezbernal:
This commit changes the logic to make the eBPF layer optional. It can be
used for different things like:
- A gadget that only includes wasm: The wasm module could collect some
information from the system using wasm.
- A gadget that only defines the metadata: There could be gadgets that
only require to define a metadata file to trigger specific operators on
IG, like the stats gadgets introduced in PR 3820.
- Gadgets like "top cpu" or "top memory" that could be implemented
without any eBPF code.


pkg/oci/fix_owner_linux.go
```diff
@@ -21,42 +21,27 @@ import (
...
-			return err
-		}
-	}
 
-	return nil
-}
-
-func fixGeneratedFilesOwner(opts *BuildGadgetImageOpts) error {
-	info, err := os.Stat(opts.EBPFSourcePath)
+	info, err := os.Stat(cwd)
```
alban, line 30: Could this just be `os.Stat(".")`?

pkg/oci/fix_owner_linux.go
```diff
@@ -21,42 +21,27 @@ import (
 	"syscall"
 )
 
-func fixOwner(targetFile, modelFile string) error {
-	info, err := os.Stat(modelFile)
+func fixOwner(targetFiles ...string) error {
+	cwd, err := os.Getwd()
```
alban, line 25: This relies on the chdir done previously. This is fine but it is done in a different package. Could the function be named `fixOwnerWithCwd` to document this requirement?

cmd/common/image/helpers/Makefile.build
```diff
@@ -12,7 +12,10 @@ ARCHS = amd64 arm64
 TARGETS = $(foreach ARCH,$(ARCHS),$(OUTPUTDIR)/$(ARCH).bpf.o)
 
 .PHONY: all
-all: $(TARGETS) wasm
+all:
```
alban, line 15: The `all` target is no longer used. Should it just be removed?

pkg/oci/fix_owner_linux.go
mauriciovasquezbernal: I changed the approach not to rely on this. Now the function receives a "model file" which is used to change the owner on all other files. It's a callers responsibility to pass this file.

cmd/common/image/helpers/Makefile.build
mauriciovasquezbernal: It was needed for gadgets with only the metadata file (like the one used in #3820), otherwise make tries to run the ebpf target. However, I fixed it by removing this and adding some extra logic to avoid invoking make if there's nothing to build.

pkg/oci/fix_owner_linux.go
```diff
@@ -21,42 +21,22 @@ import (
 	"syscall"
 )
 
-func fixOwner(targetFile, modelFile string) error {
+func fixOwner(modelFile string, targetFiles ...string) error {
```
flyth, line 24: Maybe call this func `copyFileOwner()` or something like that?

flyth (approved):
LGTM from code inspection. Will run a couple of manual tests later on.

PR #3820, opened 2024-12-18
gadgets: Introduce bpfstats gadget

### LINKED ISSUE #3759: Export Gadgets' statistics
Inspektor Gadget should provide information about the CPU and memory used by the gadgets being run. It's similar to the existing "[top ebpf](https://www.inspektor-gadget.io/docs/v0.35.0/gadgets/builtin/top/ebpf)" command, but data should be aggregated per gadget too.

There should be two ways to gather this information:

### Terminal

A new `top` command should be implemented in the different binaries:

```bash 
$ ig top 
GADGET_NAME     IMAGE               PROG_NAME           TOTAL_RUNTIME        TOTAL_RUN_COUNT      AVG_RUNTIME
my-trace-open   trace_open:latest                       432086               124                  3484
                                    ig_openat_x         326709               63                   5185
                                    ig_open_e           0                    0                    0
                                    ig_openat_e         105377               61                   1727
                                    ig_open_x           0                    0                    0
                                    
--- 
... more gadgets
```

We'd need to decide whether per program information will be printed by default or if there should be a flag like `ig top --per-program` to enable it.

### API Metrics 

This information should be available on the gRPC API and exported as metrics (see #2981) 

### Implementation Options 

This support has to be part of the framework and should not be a gadget, i.e. not to implement it as proposed in #3443.

mauriciovasquezbernal:
This PR introduces support for exposing statistics (metrics) about the ebpf programs and gadgets running on the system. 
Check the individual commits to get more details on the implementation.

## Testing 

1. build the gadget

```bash 
$ sudo ig image build gadgets/bpfstats -t bpfstats
```

2. start ig on daemon mode 

```bash 
$ sudo ig daemon --verify-image=false --enable-bpfstats
```

3. Run some gadgets

```bash 
$ sudo gadgetctl run --name mytraceopen trace_open:latest --detach
$ sudo gadgetctl run --name mytraceexec trace_exec:latest --detach
```

### Using the CLI 

4. Run the bpfstats gadget 

```bash 
$ sudo gadgetctl run bpfstats --sort -runtime
WARN[0000] local                | image signature verification is disabled due to using corresponding option
GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
a9b10cf8fd5 mytraceopen          trace_open:latest         0                               100497             258         1710240      6
9da0de911e9 mytraceexec          trace_exec:latest         0                                    0               0       183170336      7
```

By default, the gadget shows the stats aggregated by gadget. Using `--programs` makes it return per-program stats:

```bash
$ sudo gadgetctl run bpfstats --sort -runtime --programs
GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
                                                          51 __x64_sys_socke             55538314           10478         6467272      7
                                                          50 __x64_sys_socke             18184594           10474         1083840      4
                                                          53 __x64_sys_bind_             15679318            4407         6467272      7
                                                          47 __x64_sys_unlin             12154534            1278         6467272      7
                                                          55 __x64_sys_conne             11895276            3375         6467272      7
                                                          52 __x64_sys_bind_              5629158            4407         1083840      4
                                                          39 sched_process_e              4845671            2760         4391808      4
                                                          38 task_newtask                 4841680            2475         4391808      4
                                                          54 __x64_sys_conne              3883108            3341         1083840      4
                                                          46 __x64_sys_unlin              3576277            1278         1083840      4
                                                          41 __x64_sys_renam              3163412             286         6467272      7
                                                          37 sched_process_e              1597635             256         4391808      4
                                                          49 __x64_sys_unlin              1490552             225         6467272      7
a9b10cf8fd5 mytraceopen          trace_open:latest       643 ig_openat_e                  1185322            3154         1098240      3
                                                          59 __x64_sys_accep              1106122             242         6467272      
...
```

### Exporting to Prometheus

Enable the metrics on the daemon

```bash
$ sudo ig daemon --verify-image=false --enable-bpfstats --otel-metrics-listen
```

now run the gadget with the annotations to enable the metrics

```bash 
$ sudo gadgetctl run bpfstats --otel-metrics-name=bpfstats:bpfstats --annotate=bpfstats:metrics.collect=true --detach
INFO[0000] installed as "8423cd8e53339c8d4501ec7cdff436bc"
```

check the available metrics:

```bash 
$ curl http://localhost:2224/metrics -s  | grep bpfstats
mapCount{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 7
mapCount{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 6
mapMemory{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 1.83170336e+08
mapMemory{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 1.71024e+06
otel_scope_info{otel_scope_name="bpfstats",otel_scope_version=""} 1
runcount_total{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 636
runcount_total{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 219513
runtime_total{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 445114
runtime_total{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 7.3383934e+07
```

### Discussion

- Should it be really a gadget? in this case, almost all the work is done by the operator, the gadget is just a template enabling it. I was able to implement a [POC](https://github.com/inspektor-gadget/inspektor-gadget/tree/mauricio/bpfstats-without-gadget) to have this as a first-class citizen in IG:

```bash 
$ sudo gadgetctl stats
preRun
GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
a9b10cf8fd5 mytraceopen          trace_open:latest         0                              2194531            7884         1710240      6
9da0de911e9 mytraceexec          trace_exec:latest         0                                23245              28       183170336      7
```

### TODO

- [x] Documentation 
- [x] Testing

### Issues to open
- [ ] bpf_metadata (https://github.com/inspektor-gadget/inspektor-gadget/pull/3820#discussion_r1969477020)
- [ ] How to hide columns based on params (https://github.com/inspektor-gadget/inspektor-gadget/pull/3820#discussion_r1951757640)
- [ ] eBPF stats are never disabled 
- [ ] Gadget name and ID missing for attached gadgets (https://github.com/inspektor-gadget/inspektor-gadget/pull/3820#discussion_r2035553258)
- [ ] Provide human version of runtime and map size
- [ ] Test external programs too

Fixes #3759

### DIFF
--- a/gadgets/bpfstats/README.mdx
+++ b/gadgets/bpfstats/README.mdx
@@ -1,0 +1,317 @@
+---
+title: bpfstats
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# bpfstats
+
+The bpfstats Gadget provides CPU and memory usage for Gadgets and eBPF programs.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run ghcr.io/inspektor-gadget/gadget/bpfstats:%IG_TAG% [flags]
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run ghcr.io/inspektor-gadget/gadget/bpfstats:%IG_TAG% [flags]
+```
+  </TabItem>
+</Tabs>
+
+## Flags
+
+### `--all`
+
+Collect statistics for all eBPF programs
+
+Default value: "false"
+
+## Guide
+
+In order for this gadget to work, the server needs to be running with the
+`--enable-bpfstats` flag:
+
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget deploy --enable-bpfstats
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo ig daemon --enable-bpfstats
+```
+  </TabItem>
+</Tabs>
+
+Then, let's run the gadget:
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+
+```bash
+$ kubectl gadget run bpfstats:%IG_TAG%
+NODENAME           GADGETID  GADGETNAME         GADGETIMAGE        PROGID PROGNAME               RUNTIME     RUNCOUNT    MAPMEMORY MAPC…
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run bpfstats:%IG_TAG%
+GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
+```
+  </TabItem>
+</Tabs>
+
+You can run the gadget with `--help` to get the meaning of the different
+columns:
+
+- gadgetID
+  Unique ID assigned to each Gadget instance
+- gadgetImage
+  Name of the Gadget image (like trace_open, trace_exec, etc.)
+- gadgetName
+  Name of the Gadget instance
+- mapCount
+  Number of maps used by the eBPF program or Gadget
+- mapMemory
+  Memory used by maps in bytes
+- progID
+  eBPF program ID assigned by the Linux kernel
+- progName
+  Name of the eBPF program
+- runcount
+  Number of times the eBPF program or Gadget has run
+- runtime
+  Time that the eBPF program or Gadget has run in nanoseconds
+
+It won't print anything as there are not gadgets running on the system. Let's
+create a couple of gadgets:
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run --name mytraceopen trace_open:%IG_TAG% --detach
+$ kubectl gadget run --name mytraceexec trace_exec:%IG_TAG% --detach
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run --name mytraceopen trace_open:%IG_TAG% --detach
+$ sudo gadgetctl run --name mytraceexec trace_exec:%IG_TAG% --detach
+```
+  </TabItem>
+</Tabs>
+
+The bpfstats will print the statistics for these two gadgets:
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run bpfstats:%IG_TAG%
+NODENAME           GADGETID  GADGETNAME         GADGETIMAGE        PROGID PROGNAME               RUNTIME     RUNCOUNT    MAPMEMORY MAPC…
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%     0                        2537846        10798      1710240     6
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%     0                           8265           12      8742176     7
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run bpfstats:%IG_TAG%
+GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%       0                                    0               0         8742176      7
+bb95fa2c8ab mytraceopen          trace_open:%IG_TAG%       0                                95233             242         1710240      6
+```
+  </TabItem>
+</Tabs>
+
+By default the gadget prints consolidated information for the Gadgets. In this
+case PROGID is 0 and PROGNAME is empty, as a Gadget can have multiple programs.
+It's possible to get per-program statistics by using the `--all` flag. In
+this mode, the statistics for eBPF programs not handled by Inspektor Gadget are
+shown as well.
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run bpfstats:%IG_TAG% --all
+NODENAME           GADGETID  GADGETNAME         GADGETIMAGE        PROGID PROGNAME               RUNTIME     RUNCOUNT    MAPMEMORY MAPC…
+minikube                                                                2 hid_tail_call                0            0         8512     1
+minikube                                                               38 sched_process_e       89770034        15277      4391808     4
+minikube                                                               39 task_newtask         172413186        91790      4391808     4
+minikube                                                               40 sched_process_e      144593647        91635      4391808     4
+minikube                                                               41 __x64_sys_renam        7185786         2173      1083840     4
+minikube                                                               42 __x64_sys_renam       23368056         2184      6467272     7
+minikube                                                               43 __x64_sys_renam        4649957         2069      1083840     4
+minikube                                                               44 __x64_sys_renam       18957942         2072      6467272     7
+minikube                                                               45 __x64_sys_renam              0            0      1083840     4
+minikube                                                               46 __x64_sys_renam              0            0      6467272     7
+minikube                                                               47 __x64_sys_unlin       37799602        14147      1083840     4
+minikube                                                               48 __x64_sys_unlin      120719371        14370      6467272     7
+minikube                                                               49 __x64_sys_unlin       17037065         9561      1083840     4
+minikube                                                               50 __x64_sys_unlin       56861068         9561      6467272     7
+minikube                                                             1244 ig_execve_e           15430709         3239       849792     2
+minikube                                                             1245 ig_execve_x            1792662         3655       837504     1
+minikube                                                             1246 ig_fa_pick_e           2899018         4581        21632     2
+minikube                                                             1247 ig_fa_pick_x           4478788         4581      1194336     4
+minikube                                                             1248 ig_sched_exec          2759885         2728       837504     1
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%  1259 ig_open_x                    0            0      1615904     4
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%  1260 ig_openat_x           14944051        83160      1615904     4
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%  1261 ig_open_e                    0            0      1098240     3
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%  1262 ig_openat_e           33097520        83005      1098240     3
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1263 ig_execve_x             200186          570      8647840     5
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1264 ig_execveat_x                0            0      8647840     5
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1265 security_bprm_c         225868          362      8372040     4
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1266 ig_execve_e             340822          518      8372040     4
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1267 ig_execveat_e            37108           52      8372040     4
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1268 ig_sched_exec           226692          362      8647840     1
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run bpfstats --all
+GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
+                                                           2 hid_tail_call                      0               0            8512      1
+                                                          38 sched_process_e              1212539             180         4391808      4
+                                                         198                                    0               0           48672      1
+                                                         736 ig_execve_e                  1103998             317          849792      2
+                                                         737 ig_execve_x                   160239             317          837504      1
+                                                         738 ig_fa_pick_e                  373093             509           21632      2
+                                                         739 ig_fa_pick_x                  378858             509         1194336      4
+                                                         740 ig_sched_exec                 112317              67          837504      1
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%     743 security_bprm_c                49435              67         8372040      4
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%     744 ig_execve_e                   140296             317         8372040      4
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%     745 ig_execveat_e                      0               0         8372040      4
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%     746 ig_execve_x                    89667             317         8647840      5
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%     747 ig_sched_exec                  76171              67         8647840      5
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%     748 ig_execveat_x                      0               0         8647840      5
+bb95fa2c8ab mytraceopen          trace_open:%IG_TAG%     749 ig_open_e                          0               0         1098240      3
+bb95fa2c8ab mytraceopen          trace_open:%IG_TAG%     750 ig_openat_e                 21538317           48300         1098240      3
+bb95fa2c8ab mytraceopen          trace_open:%IG_TAG%     751 ig_open_x                          0               0         1615904      4
+bb95fa2c8ab mytraceopen          trace_open:%IG_TAG%     752 ig_openat_x                  9076187           48503         1615904      4
+```
+  </TabItem>
+</Tabs>
+
+Finally, clean the system:
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget remove mytraceexec mytraceopen
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl remove mytraceexec mytraceopen
+```
+  </TabItem>
+</Tabs>
+
+### Exporting metrics
+
+The `bpfstats` Gadget provides the following metrics for `mapCount`,
+`mapMemory`, `runcount` and `runtime`. To enable the metrics listener, check the
+[Exporting Metrics](../reference/export-metrics.mdx) documentation. To enable
+the collector for this gadget, run the following command:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo gadgetctl run bpfstats --name mystats --otel-metrics-name=bpfstats:bpfstats --annotate=bpfstats:metrics.collect=true --detach
+        INFO[0000] installed as "8423cd8e53339c8d4501ec7cdff436bc"
+        ```
+    </TabItem>
+</Tabs>
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        Unless you configured the metrics listener to do differently, the
+        metrics will be available at `http://localhost:2224/metrics` on the
+        server side. For the `bpfstats` gadget we ran above, you
+        can find the metrics under the `bpfstats` scope:
+
+        ```bash
+        $ curl http://localhost:2224/metrics -s  | grep bpfstats
+        mapCount{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 7
+        mapCount{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 6
+        mapMemory{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 1.83170336e+08
+        mapMemory{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 1.71024e+06
+        otel_scope_info{otel_scope_name="bpfstats",otel_scope_version=""} 1
+        runcount_total{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 636
+        runcount_total{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 219513
+        runtime_total{gadgetImage="trace_exec:latest",gadgetName="mytraceexec",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 445114
+        runtime_total{gadgetImage="trace_open:latest",gadgetName="mytraceopen",otel_scope_name="bpfstats",otel_scope_version="",progName=""} 7.3383934e+07
+        ```
+    </TabItem>
+</Tabs>
+
+Finally, stop metrics collection:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo gadgetctl delete mystats
+        8423cd8e53339c8d4501ec7cdff436bc
+        ```
+    </TabItem>
+</Tabs>
+
+### Limitations
+
+#### Memory usage of maps
+
+The shown value for MapMemory is read from `/proc/<pid>/fdinfo/<map_id>`. This
+is the maximum size the map can have, but it doesn't necessarily reflect its
+current memory allocation. Additionally, maps can be used by more than one
+program and would account towards the MapMemory of all those programs.
+
+Also note:
+* BPF_MAP_TYPE_PERF_EVENT_ARRAY: value_size is not counting the ring buffers,
+  but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+* BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS: value_size is not counting the inner maps,
+  but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+
+#### bpfstats not disabled for Kernels < 5.8
+
+The bpfstats Gadgets enables collection of eBPF program stats by trying to use
+`BPF_ENABLE_STATS` first (which requires Linux >= 5.8). If that fails, it will
+fall back to trying to enable via sysctl (/proc/sys/kernel/bpf_stats_enabled),
+on this case, the collection of stats will remain active even after stopping
+Inspektor Gadget.
+
+It is a temporary limitation that we're working to fix, in the meanwhile you can
+use `sysctl kernel.bpf_stats_enabled=0` to disable it.
--- a/gadgets/bpfstats/gadget.yaml
+++ b/gadgets/bpfstats/gadget.yaml
@@ -1,0 +1,8 @@
+name: bpfstats
+description: Get memory and CPU usage of eBPF programs and Gadgets
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/bpfstats
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/bpfstats
+operator:
+  ebpf:
+    emitstats: true
--- a/gadgets/bpfstats/test/unit/bpfstats_test.go
+++ b/gadgets/bpfstats/test/unit/bpfstats_test.go
@@ -1,0 +1,151 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/gadgetrunner"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+)
+
+type ExpectedBpfstatsEvent struct {
+	GadgetID    string `json:"gadgetID"`
+	GadgetImage string `json:"gadgetImage"`
+	GadgetName  string `json:"gadgetName"`
+	MapCount    int    `json:"mapCount"`
+	MapMemory   int    `json:"mapMemory"`
+	ProgID      int    `json:"progID"`
+	ProgName    string `json:"progName"`
+	Runcount    int    `json:"runcount"`
+	Runtime     int    `json:"runtime"`
+}
+
+type testDef struct {
+	runnerConfig   *utilstest.RunnerConfig
+	allPrograms    bool
+	generateEvent  func(t *testing.T)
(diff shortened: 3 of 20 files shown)


mauriciovasquezbernal:
> Should it be really a gadget?

I started the implementation this way to reuse all the datasources, operators, etc. logic to implement this functionality. However, it seems weird to have an almost empty gadget to have this. I'd prefer to have it implemented as a `gadgetctl stats/top` native command instead of having to use `gadgetctl run bpfstats` and pull the image from a container registry.

Any more opinions on this?

pkg/operators/ebpf/ebpf.go
```diff
@@ -676,6 +686,34 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 	}
 	i.collection = collection
 
+	if i.bpfOperator.gadgetObjs != nil {
```
blanquicet, line 689: We don't need to collect program and map IDs for this gadget if the emitstats annotation wasn't set. I know it will do nothing anyway, but this operation is anyway useless in that case.

pkg/operators/ebpf/ebpf.go
```diff
@@ -676,6 +686,34 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+		// collect program IDs and map IDs for this gadget
+		i.bpfOperator.mu.Lock()
+		gadgetObjs := gadgetObjs{}
+
+		for _, p := range i.collection.Programs {
+			info, err := p.Info()
+			if err != nil {
+				continue
```
blanquicet, line 697: Log a warning this failed? I think it's an important log because (1) user enabled stats on daemon and (2) user requested to collect stats for this gadget (see previous comment), and we will not provide info about that specific program. I think at least a warn log is worth it.

pkg/operators/ebpf/ebpf.go
```diff
@@ -676,6 +686,34 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+			}
+
+			id, _ := info.ID()
+			gadgetObjs.programIDs = append(gadgetObjs.programIDs, id)
+		}
+
+		for _, m := range i.collection.Maps {
+			info, err := m.Info()
+			if err != nil {
+				continue
```
blanquicet, line 707: dito

pkg/operators/ebpf/ebpf.go
```diff
@@ -676,6 +686,34 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+			}
+
+			id, _ := info.ID()
+			gadgetObjs.mapIDs = append(gadgetObjs.mapIDs, id)
+		}
+		i.bpfOperator.gadgetObjs[gadgetCtx] = gadgetObjs
```
blanquicet, line 713: `i.bpfOperator.mu.Lock()` can be moved right before this line.

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,535 @@
...
+
+func (i *ebpfOperator) Init(params *params.Params) error {
+	if !params.Get(EnableBPFStatsParam).AsBool() {
+		return nil
+	}
+
+	i.gadgetObjs = make(map[operators.GadgetContext]gadgetObjs)
+
+	// Enable stats collection
+	// TODO: when to disable it?
```
blanquicet, line 64: I think it should be enabled when starting the data operator instance and disabled when stopping it. Then, the internal `refCnt` of bpfstats package should do the trick.

BTW, I think the logic is missing an increasement when it has been already enabled:

```go
func EnableBPFStats() error {
	mutex.Lock()
	defer mutex.Unlock()

	if refCnt != 0 {
		refCnt++ // This is missing in the original code
		return nil
	}
```

Or, did I miss something?

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,535 @@
...
+	if err := bpfstats.EnableBPFStats(); err != nil {
+		return fmt.Errorf("enabling bpf stats: %w", err)
+	}
+
+	return nil
+}
+
+func (i *ebpfOperator) InstantiateDataOperator(
```
blanquicet, line 72: I think the idea of making the ebpf op also a data operator is brilliant! Bravo!

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	instance.ds, err = gadgetCtx.RegisterDataSource(datasource.TypeArray, StatsDSName)
+	if err != nil {
+		return nil, err
+	}
+
+	intervalAnn := paramValues[ParamGadgetStatisticsInterval]
+	if intervalAnn == "" {
+		intervalAnn = "1000ms"
```
blanquicet, line 108: This should be already there, isn't it?

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+	instance.interval, err = time.ParseDuration(intervalAnn)
+	if err != nil {
+		return nil, fmt.Errorf("parsing duration: %w", err)
+	}
+
+	countAnn := paramValues[ParamGadgetStatisticsCount]
+	if countAnn == "" {
+		countAnn = "0"
```
blanquicet, line 117: dito

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,535 @@
...
+			return nil, err
+		}
+		instance.nodeName = nodeName
+	}
+
+	// gadget-specific fields
+	instance.gadgetIDField, err = instance.ds.AddField("gadgetID", api.Kind_String,
+		datasource.WithAnnotations(map[string]string{
+			metadatav1.ColumnsWidthAnnotation: "8",
+			// NOT REALLY SURE
```
blanquicet, line 145: Why should we need this? The gadgetName is already unique and 1:1 with the ID, so having the ID would just increase the cardinality of the metrics. Or what am I missing?

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,535 @@
...
+			metadatav1.ColumnsWidthAnnotation: "16",
+			"metrics.type":                    "key",
+		}),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// low-level fields
+	// TODO: how to disable it based on the programs param?
```
blanquicet, line 172: IIRC, we instantiate twice the operator: one for the `GetGadgetInfo` and another for the `RunGadget`. For `GetGadgetInfo` programStats will be always false, but for the `RunGadget`, it should carry the proper value. Mmmm why can't we use `programStats`? I'm sure I'm missing something.

gadgets/bpfstats/gadget.yaml
```diff
@@ -0,0 +1,7 @@
+name: bpf stats
```
blanquicet, line 1: I don't think it should be a gadget, but a separated command like in the PoC you implemented.

blanquicet (commented):
I did an initial review, and I really like it!

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: > I think it should be enabled when starting the data operator instance and disabled when stopping it.

If the option is enabled globally (--enable-bpfstats), we need to enable it when the daemon starts, otherwise we'll miss the stats for gadgets run before the stats command/gadget.

The issue here is that operators don't have a Close() or similar method to perform cleanup tasks.

mauriciovasquezbernal: > BTW, I think the logic is missing an increasement when it has been already enabled:
> 
> ```go
> func EnableBPFStats() error {
> 	mutex.Lock()
> 	defer mutex.Unlock()
> 
> 	if refCnt != 0 {
> 		refCnt++ // This is missing in the original code
> 		return nil
> 	}
> ```
> 
> Or, did I miss something?

Yes, that seems like a bug. It's fixed in #3976

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: Yes, that's right, the ID shouldn't be a key.

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: The first call doesn't get any parameters and it's the result of this first call which is used by the CLI operator to set the default columns, so I haven't found a way to change the list of columns (or their visibility) based on a param.

pkg/operators/ebpf/stats.go
blanquicet: Mmm you can simply don't validate them if they are not present. That's what I did in the [limiter](https://github.com/inspektor-gadget/inspektor-gadget/blob/4d0d32998539152e1156ce3a2cf5fafa15c61cea/pkg/operators/limiter/limiter.go#L67-L80).

pkg/operators/ebpf/stats.go
blanquicet: What about a [custom param](https://github.com/inspektor-gadget/inspektor-gadget/pull/3485) that adds an annotation to the `emitstats`?

pkg/operators/ebpf/stats.go
blanquicet: > If the option is enabled globally (--enable-bpfstats), we need to enable it when the daemon starts, otherwise we'll miss the stats for gadgets run before the stats command/gadget.
>
Aaahh because we want to collect stats from the moment a gadget starts, not from the moment the stats command/gadget is executed. Good point.

> The issue here is that operators don't have a Close() or similar method to perform cleanup tasks.
>
Yeah, then I agree.

pkg/operators/ebpf/ebpf.go
blanquicet: Ahh ok, because the `emitstats` is not an annotation we add to the gadget we want to collect stats from, but to the "empty" gadget we are running to gather the collected stats. Ok, ok. Understood. Thanks for the explanation.

pkg/operators/ebpf/stats.go
blanquicet: I found an issue with my approach in the limiter. There are some cases, like for unit tests, where the operator is called without `paramValues`. If I don't set a default value, then the operator will run with `0`, which in the case of the limiter is to truncate all the output 😅.

alban:
Could we have metrics about individual ebpf maps too? Or would that be for a future PR?
In https://github.com/inspektor-gadget/inspektor-gadget/pull/4058, I couldn't use the top-ebpf gadget, but I needed to run bpftool & jq manually:

```
```

mauriciovasquezbernal:
> Could we have metrics about individual ebpf maps too? Or would that be for a future PR? In #4058, I couldn't use the top-ebpf gadget, but I needed to run bpftool & jq manually:
> 
> ```
> $ sudo bpftool map show -j | jq -r '.[] | {id: .id, type: .type, name: .name, bytes_memlock: .bytes_memlock, comms: (if .pids then (.pids | map(.comm) | unique | join(",")) else "" end)} | [.id, .type, .name, .bytes_memlock, .comms] | @tsv' | sort -k4n | column -t -s $'\t'|grep ig
> ```

I haven't planned to implement that functionality on this PR, but once we agree on the current proposal, I could try it. I suppose the goal is to print the maps information and enrich that with gadget information if possible, i.e. to say which gadget is using a given map. Do you have something else on mind?

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: I'd like to avoid this discussion for now. Once we agree on the whole idea, we can merge it and I can try to improve it later on.

alban:
> I haven't planned to implement that functionality on this PR, but once we agree on the current proposal, I could try it.

That's fine.

> I suppose the goal is to print the maps information and enrich that with gadget information if possible, i.e. to say which gadget is using a given map. Do you have something else on mind?

Yes, that's correct. Note:
- a ebpf map might belong to several gadgets (e.g. the gadget_sockets map from the socket enricher)
- a ebpf map might not belong to any gadgets (e.g. maps not created by ig)

Then, I wanted to sort the maps by size to see which ones to optimize first.

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: Updated following #4050

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,527 @@
...
+
+func (i *ebpfOperatorDataInstance) getStats() ([]stat, error) {
+	stats := make([]stat, 0)
+
+	mapSizes, err := bpfstats.GetMapsMemUsage()
+	if err != nil {
+		return nil, fmt.Errorf("getting map memory usage: %w", err)
+	}
+
+	// TODO: reduce lock contention
```
alban, line 380: Is it necessary? `getStats` is called from a Ticker, so no parallel calls. And other usages of `mu.Lock()` are not frequent.

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,527 @@
...
+		stat.runcount -= oldProgStat.runcount
+
+		mapIDs, _ := pi.MapIDs()
+		for _, mapID := range mapIDs {
+			stat.mapMemory += mapSizes[mapID]
+			stat.mapCount += 1
+		}
+
+		// enrich with gadget information if they're part of a gadget
+		if ctx, ok := programToGadget[ebpf.ProgramID(id)]; ok {
```
alban, line 479: This only works if the gadget is executed by the same ig process as the one running the stats.

To make it work when running several ig processes, we have to use bpf_metadata:
https://qmonnet.github.io/whirl-offload/2021/09/23/bpftool-features-thread/#program-metadata

Could you file an issue for a future PR?

gadgets/bpfstats/gadget.yaml
flyth: > IMO, the fact that it's reusing the gadget lifecycle to run an "empty" gadget doesn't make it a gadget. It's an internal detail that doesn't need to be exposed to users.

I disagree with this. IMHO, the fact that it uses the lifecycle is exactly what makes it a gadget. It's similar to the discussion we had about mem/cpu usage. An annotation like the one used here should be used to enable internal datasources that emit events regarding mem/cpu usage. A basic gadget might just take that data and print it, while another gadget could combine that info with other data from ebpf/wasm/whatever.

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,527 @@
...
+				if err != nil {
+					gadgetCtx.Logger().Errorf("Failed to get stats: %v", err)
+					continue
+				}
+				if err := i.sendStats(stats); err != nil {
+					gadgetCtx.Logger().Errorf("Failed to emit stats: %v", err)
+					continue
+				}
+				ctr++
+				// TODO: client remians connected even if the gadget is stopped
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Running `sudo gadgetctl run bpfstats --statistics-count 5` doesn't work because the gadget ends emitting the data but the client keeps connected (i.e. it doesn't return automatically and the user has to press Ctrl +C). I fixed it by removing this interval for now. --timeout X is always a possibility to run for a given amount of time.

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: You're right. I was a bit bothered at the beginning because it was covering a much bigger region, but I think it's fine now.

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: TIL about bpf_metadat, I'll open the issue when merging this PR. Thanks!

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,527 @@
...
+		i.mapCountField.PutUint64(d, stat.mapCount)
+
+		arr.Append(d)
+	}
+
+	return i.ds.EmitAndRelease(arr)
+}
+
+func getProgramStats(cache map[ebpf.ProgramID]progStat, id ebpf.ProgramID) (progStat, error) {
+	if stat, ok := cache[id]; ok {
```
flyth, line 348: `cache` is never filled.

flyth (commented):
I like this a lot! Will play around with it a bit and report back.

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,527 @@
...
+		if err != nil {
+			return nil, err
+		}
+		instance.nodeName = nodeName
+	}
+
+	// gadget-specific fields
+	instance.gadgetIDField, err = instance.ds.AddField("gadgetID", api.Kind_String,
+		datasource.WithAnnotations(map[string]string{
+			metadatav1.ColumnsWidthAnnotation: "8",
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'm not sure. Wouldn't it increase the cardinality at lot? We already have name (which should be unique) for this purpose. 

ref https://github.com/inspektor-gadget/inspektor-gadget/pull/3820#discussion_r1951756581

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,527 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// low-level fields
+	// TODO: how to disable it based on the programs param?
+	instance.progIDField, err = instance.ds.AddField("progID", api.Kind_Uint32,
+		datasource.WithAnnotations(map[string]string{
+			metadatav1.ColumnsAlignmentAnnotation: string(metadatav1.AlignmentRight),
+			metadatav1.ColumnsWidthAnnotation:     "5",
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I think it shouldn't be there. The progID isn't stable between different executions of the same gadget, so it'll be difficult to correlate it.

pkg/operators/ebpf/stats.go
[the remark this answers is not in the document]
flyth: If both `id` and `name` are grouped, they wouldn't increase cardinality but only the key size. But if one of these is re-used (while the other is not), then in one case the runs would be grouped and in the other case they wouldn't. I don't think it matters a lot as this probably won't happen, but I think it's cleaner.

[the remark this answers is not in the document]
flyth: Actually, it's probably more common to set the name and not the id and expect everything to be grouped under whatever you called it, so you're right.

pkg/operators/ebpf/stats.go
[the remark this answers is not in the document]
flyth: This is similar to the comment above, it depends on how granular we want to be (identifying single gadget runs or not). But yeah, let's not do it for now - I guess in most cases having the `node` keyed will sort it out anyway.

gadgets/bpfstats/gadget.yaml
mauriciovasquezbernal: After some discussion we agreed it's better to have it as a gadget.

gadgets/bpfstats/gadget.yaml
```diff
@@ -0,0 +1,7 @@
+name: bpfstats
```
blanquicet, line 1: WDYT about adding the datasource with the description of fields. It is useful when one runs `--help` to get the fields' description:

```bash
$ go run -exec "sudo -E" ./cmd/gadgetctl run bpfstats --help
[...]
                                                 "bpfstats" (data source):
                                                   gadgetID
                                                   gadgetImage
                                                   gadgetName
                                                   mapCount
                                                   mapMemory
                                                   progID
                                                   progName
                                                   runcount
                                                   runtime
```

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,505 @@
...
+	instance.interval, err = time.ParseDuration(intervalAnn)
+	if err != nil {
+		return nil, fmt.Errorf("parsing duration: %w", err)
+	}
+
+	instance.ds.AddAnnotation(api.FetchIntervalAnnotation, intervalAnn)
+	instance.ds.AddAnnotation("cli.clear-screen-before", "true")
+
+	if nodeName, ok := os.LookupEnv("NODE_NAME"); ok {
+		instance.nodeNameField, err = instance.ds.AddField("nodeName", api.Kind_String,
```
blanquicet, line 114: Or should the description be included here when adding the fields?

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+Collect per-program statistics
+
+Default value: "false"
+
+## Guide
+
+In order for this gadget to work, the server needs to be running with the
+`--enable-bpfstats` flag:
+
+First, we need to run an application that generates some events.
```
blanquicet, line 44: Wrong copy & paste? 😅

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%     0                           8265           12      8742176     7
+```
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run bpfstats:%IG_TAG%
+GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
+a624bdbf27f mytraceexec          trace_exec:%IG_TAG%       0                                    0               0         8742176      7
+bb95fa2c8ab mytraceopen          trace_open:%IG_TAG%       0                                95233             242         1710240
```
blanquicet, line 115: MapCount is missing in this line

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+```
+  </TabItem>
+</Tabs>
+
+By default the gadget prints consolidated information for the Gadgets. It's
```
blanquicet, line 120: Should we add a comment here explaining that the programid will be 0 when running it per gadget?

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+possible to get per-program statistics by using the `--programs` flag. In this
+mode, the statistics for eBPF programs not handled by Inspektor Gadget are shown
+as well.
```
blanquicet, line 123: I'm wondering if we should use two flags for the two functionalities this flag provides:

- `per-program` to not summarize stats per gadget but per program
- `all` to show also programs that don't belong to any gadget

WDYT?

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+<Tabs groupId="env">
+  <TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run bpfstats:%IG_TAG% --programs
+NODENAME           GADGETID  GADGETNAME         GADGETIMAGE        PROGID PROGNAME               RUNTIME     RUNCOUNT    MAPMEMORY MAPC…
+WARN[0000] minikube             | image signature verification is disabled due to using corresponding option
```
blanquicet, line 130: Please remove this line. It just creates noise.

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%  1261 ig_open_e                    0            0      1098240     3
+minikube           6bcfcca7a mytraceopen        trace_open:%IG_TAG%  1262 ig_openat_e           33097520        83005      1098240     3
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1263 ig_execve_x             200186          570      8647840     5
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1264 ig_execveat_x                0            0      8647840     5
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1265 security_bprm_c         225868          362      8372040     4
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1266 ig_execve_e             340822          518      8372040     4
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1267 ig_execveat_e            37108           52      8372040     4
+minikube           040e3fc1d mytraceexec        trace_exec:%IG_TAG%  1268 ig_sched_exec           226692          362      8647840
```
blanquicet, line 161: Also here the map count is missing.

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+  </TabItem>
+
+  <TabItem value="ig-daemon" label="ig-daemon">
+```bash
+$ sudo gadgetctl run bpfstats:%IG_TAG%
+GADGETID    GADGETNAME           GADGETIMAGE          PROGID PROGNAME                     RUNTIME        RUNCOUNT       MAPMEMORY MAPCO…
+```
+  </TabItem>
+</Tabs>
+
```
blanquicet, line 78: Here we should explain the meaning of each column.

pkg/gadget-service/instance-manager/instance.go
```diff
@@ -205,6 +205,8 @@ func (p *GadgetInstance) Run(
 		gadgetcontext.WithLogger(logger),
 		gadgetcontext.WithDataOperators(ops...),
 		gadgetcontext.WithAsRemoteCall(true),
+		gadgetcontext.WithName(p.name),
+		gadgetcontext.WithID(p.id),
```
blanquicet, line 209: Probably out of scope of this PR but it shows the issue. I think that we should assign the gadget name and id to all gadgets, not only for headless mode. It's because, right now, if I run these two gadgets:

```bash
gadgetctl run trace_open:v0.39.0 --name name1
```

```bash
gadgetctl run trace_open:v0.39.0 --name name2
```

IG will completely ignore the names and there will no way to distinguish them in the bpfstats gadget:

```bash
GADGETID   GADGETNAME          GADGETIMAGE         PROG… PROGNAME                 RUNTIME      RUNCOUNT     MAPMEMORY MAPC…
                               trace_open:v0.39.0      0                            33683           204      11952832     7
                               trace_open:v0.39.0      0                            37666           204      11952832     7
```

blanquicet (commented):
I tested it again and it's working as expected. I also reviewed the documentation and have some comments. IMO, once the question about the `programs` flag is solved, it's ready to go.

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
+---
+title: bpfstats
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# bpfstats
```
blanquicet, line 9: I think we should mention the mechanism we are using to enable stats: BPF_ENABLE_STATS and fallback to `/proc/sys/kernel/bpf_stats_enabled`. 

In addition, we should highlight that for kernels < 5.8, where `BPF_ENABLE_STATS` is not available, we are not setting `kernel.bpf_stats_enabled` back to `0`.

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: I think it need to go here where fields are created.

gadgets/bpfstats/README.mdx
mauriciovasquezbernal: I added them from the output of --help but I'm not sure we should replicate it here.

gadgets/bpfstats/README.mdx
mauriciovasquezbernal: I added a comment. The column is printed there because currently there is not a way to disable it based on the arguments. Once that's done, the column will be hidden by default avoiding any confusion.

gadgets/bpfstats/README.mdx
mauriciovasquezbernal: I'm not totally convinced. I think it's too much. Perhaps having only `--all` will be enough (it prints all programs on the system, and enriches with gadget information the ones from Inspektor Gadget).

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+The bpfstats Gadget provides CPU and memory usage for Gadgets and eBPF programs.
+
+## Getting started
```
blanquicet, line 13: Please add the "Exporting metrics" section (like [this](https://inspektor-gadget.io/docs/latest/gadgets/profile_blockio#exporting-metrics)) describing the metrics this gadget is able to export with one simple example.

gadgets/bpfstats/README.mdx
blanquicet: I'm ok with just `all`.

gadgets/bpfstats/README.mdx
blanquicet: > In addition, we should highlight that for kernels < 5.8, where BPF_ENABLE_STATS is not available, we are not setting kernel.bpf_stats_enabled back to 0.
>
And probably create an issue for this.

pkg/operators/ebpf/stats.go
```diff
@@ -0,0 +1,514 @@
...
+			datasource.WithAnnotations(map[string]string{
```
blanquicet, line 115: "node" template?

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,314 @@
...
+  but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+* BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS: value_size is not counting the inner maps,
+  but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+
+#### bpfstats not disabled for Kernels < 5.8
+
+The bpfstats Gadgets enables collection of eBPF program stats by trying to use
+`BPF_ENABLE_STATS` first (which requires Linux >= 5.8). If that fails, it will
+fall back to trying to enable via sysctl (/proc/sys/kernel/bpf_stats_enabled),
+on this case,  the collection of stats will remain active even after stopping
```
blanquicet, line 313: nit:

```suggestion
on this case, the collection of stats will remain active even after stopping
```

WDYT about specifying:
1. This is a temporary limitation. Just to let people know we are working on it.
2. That they can stop it by running: `sysctl kernel.bpf_stats_enabled=0`.

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,314 @@
...
+The shown value for MapMemory is read from `/proc/<pid>/fdinfo/<map_id>`. This
+is the maximum size the map can have, but it doesn't necessarily reflect its
+current memory allocation. Additionally, maps can be used by more than one
+program and would account towards the MapMemory of all those programs.
+
+Also note:
+* BPF_MAP_TYPE_PERF_EVENT_ARRAY: value_size is not counting the ring buffers,
+  but only their file descriptors (i.e. sizeof(int) = 4 bytes)
+* BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS: value_size is not counting the inner maps,
+  but only their file descriptors (i.e. sizeof(int) = 4 bytes)
```
blanquicet, line 306: Is it something IG-specific or it's coming from the kenel functionality itself? If it's the second, it would be great if we can add a link to the kernel documentation (if it ever exists).

gadgets/bpfstats/README.mdx
```diff
@@ -0,0 +1,314 @@
...
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        WIP: Headless mode for kubectl gadget is under development
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo gadgetctl run bpfstats --otel-metrics-name=bpfstats:bpfstats --annotate=bpfstats:metrics.collect=true --detach
```
blanquicet, line 242: Tested and worked properly. I suggest specifying a name to avoid using the id when deleting it.

gadgets/bpfstats/test/unit/bpfstats_test.go
```diff
@@ -0,0 +1,151 @@
...
+				require.Contains(t, gadgetRunner.CapturedEvents, expectedEvent)
+			}
+		})
+	}
+}
+
+func generateEvent(t *testing.T) {
+	// Run a gadget so it's captured by the bpstats gadget
+	opts := gadgetrunner.GadgetRunnerOpts[ExpectedBpfstatsEvent]{
+		Image:   testGadgetImage,
```
blanquicet, line 146: Mmmm even if we pass the name here, given that we are using the local runtime, it will be ignored, right?

TODO: Validate it once we support setting name for "local" instances. Related to https://github.com/inspektor-gadget/inspektor-gadget/pull/3820#discussion_r2035553258.

blanquicet (approved):
Thanks for working on this! LGTM!

gadgets/bpfstats/README.mdx
mauriciovasquezbernal: This is coming from the kernel AFAIU. TBH I [took](https://github.com/inspektor-gadget/inspektor-gadget/blob/c1acdfeed6936a957f2e4e4eef153da677d0297f/docs/gadgets/builtin/top/ebpf.md#L98-L106) this from the top ebpf which this implementation is based on. I don't think there is any kernel documentation that explains it.

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: Ref #4317

pkg/operators/ebpf/stats.go
mauriciovasquezbernal: Ref #4318

gadgets/bpfstats/README.mdx
mauriciovasquezbernal: #4319

pkg/gadget-service/instance-manager/instance.go
mauriciovasquezbernal: Ref #4320

gadgets/bpfstats/README.mdx
blanquicet: @flyth do you remember where did you take [that](https://github.com/inspektor-gadget/inspektor-gadget/pull/876) info from?

PR #3825, opened 2024-12-19
ebpf-builder: Update to clang-18 and provide clang-format

### LINKED ISSUE #3602: [RFC] clang-format version
Newest update: The Rollout has been put on hold. Nonetheless we should discuss this

In our CI we are using `clang-format` on `ubuntu-latest`. That ubuntu version was pointing to `Ubuntu 22.04`, which has `clang-format` up to version `14.0`

There is currently a rollout to change `ubuntu-latest` to `Ubuntu 24.04` https://github.com/actions/runner-images/issues/10636 where `clang-format` is available up to version `18.0`.
The problem here is that the rules are slightly changed and we would need to reformat the code to make it compliant again: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/11347257386/job/31558242289#step:3:79
```
diff --git a/gadgets/trace_open/program.bpf.c b/gadgets/trace_open/program.bpf.c
index 9a72f99..5605657 100644
--- a/gadgets/trace_open/program.bpf.c
+++ b/gadgets/trace_open/program.bpf.c
@@ -13,7 +13,7 @@
 
 #define TASK_RUNNING 0
 #define NAME_MAX 255
-#define INVALID_UID ((uid_t)-1)
+#define INVALID_UID ((uid_t) - 1)
 
 struct args_t {
 	const char *fname;
```

- Stay on `Ubuntu 22.04`
- Update to `Ubuntu 24..04`, but install earlier version of clang-tidy
- Update to `Ubuntu 24.04` and use newest version of clang-tidy. We would need the newest version of it too locally

mauriciovasquezbernal:
I'm having some issues to format the code properly on #3648 since I don't have clang-format-18 available locally, so fix this to avoid the same pain for other users. 

- Update to clang-18
- Provide clang-format

Fixes #3602


burak-ok:
These are some nice improvements on the object files percentage wise. Nice

mauriciovasquezbernal:
> These are some nice improvements on the object files percentage wise. Nice

I noticed the same! I'm wondering if we have also some performance improvements by using a more recent compiler :thinking:

PR #3828, opened 2024-12-20
gadgets: trace_oomkill: Use gadget_process structure.

eiffel-fl:
Will update documentation once agreed on the UX as I am not sure about having two process struct in the event one.

### DIFF
--- a/gadgets/trace_oomkill/program.bpf.c
+++ b/gadgets/trace_oomkill/program.bpf.c
@@ -7,6 +7,7 @@
 #include <bpf/bpf_core_read.h>
 
 #include <gadget/buffer.h>
+#include <gadget/common.h>
 #include <gadget/macros.h>
 #include <gadget/mntns_filter.h>
 #include <gadget/types.h>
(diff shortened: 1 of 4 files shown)


gadgets/trace_oomkill/program.bpf.c
```diff
@@ -15,14 +16,9 @@
 
 struct event {
 	gadget_timestamp timestamp_raw;
-	gadget_mntns_id mntns_id;
-	__u32 fpid;
-	__u32 fuid;
-	__u32 fgid;
-	__u32 tpid;
+	struct gadget_process fprocess;
+	struct gadget_process tprocess;
```
blanquicet, line 20: I wonder if it is worth to use the `gadget_process` for the tprocess which will only carry the pid, comm and mntsns. I'm not sure, also because having two mntnsid generate this warn:

```
multiple fields of "type:gadget_mntns_id" found in DataSource "oomkill", only "mntns_id" will be used
```

Additionally, please update the gadget.yml and documentation to show the columns that need to be shown.

gadgets/trace_oomkill/program.bpf.c
mauriciovasquezbernal, line 1: TASK_COMM_LEN isn't used anymore (unless you switch back to an own definition for tprocess according to https://github.com/inspektor-gadget/inspektor-gadget/pull/3828/files#r1894490219)

gadgets/trace_oomkill/program.bpf.c
mauriciovasquezbernal: What about just using a u64 to hold the target mount namespace? (until we clarify what's the meaning of multiple mount namespaces on a structure)

mauriciovasquezbernal: With the current PR the enrichment is done with `fprocess`, right? 

What I'm proposing is to use u64 for one of the namespaces to avoid the confusion.

mauriciovasquezbernal: ok, now I get it. I think we should keep the same enrichment as on main. Let's ignore the warning for now, we can fix it later on.

mauriciovasquezbernal (approved):
LGMT from code inspection. Just one nit about the metadata.

PR #3830, opened 2024-12-20
(built-in)/trace/dns: Fix logic for percpu array map

mauriciovasquezbernal:
The BPF_NOTEXIST flag fails when used on a percpu array map as the entry for index zero is always present. The same applies to the bpf_map_delete operation.

This commit removes those two invocations and improves the logic to only clear fields when really needed.

Fixes: b7d6c1bec756 ("builtin trace-dns: use new fields from socket enricher")


blanquicet (approved):
LGTM from code inspection.

PR #3839, opened 2024-12-24
feat(kubectl-gadget): disallow instances with same name

### LINKED ISSUE #3679: kubectl-gadget: `attach` prints misleading error when there are multiple instances with the same name
```bash
$ kubectl-gadget run trace_exec:latest --name foo --detach
INFO[0002] installed as "9fb56b25bb4454fd520862a66d03ae30"

$ kubectl-gadget run trace_exec:latest --name foo --detach
INFO[0000] installed as "a8cd20255d43c86a25a669fec94e0868"

$ kubectl-gadget list
ID           NAME                                      TAGS                                     GADGET
9fb56b25bb44 foo                                                                                trace_exec:latest
a8cd20255d43 foo                                                                                trace_exec:latest

$ kubectl-gadget attach foo
Error: gadget instance not found
```

It should print ambiguous. 

Also, do we really want to allow multiple instances with the same name? It doesn't seem to be allowed on `ig`

KapilSareen:
### Fixes  #3679 

- Updated the `createGadgetInstance` function to validate if an instance with the same name already exists.
- If a duplicate is detected, the function will now return an error instead of proceeding with the creation.

### DIFF
--- a/pkg/gadget-service/store/k8s-configmap-store/configmapstore.go
+++ b/pkg/gadget-service/store/k8s-configmap-store/configmapstore.go
@@ -234,6 +234,17 @@ func (s *Store) handleErr(err error, key string) {
 func (s *Store) CreateGadgetInstance(ctx context.Context, req *api.CreateGadgetInstanceRequest) (*api.CreateGadgetInstanceResponse, error) {
 	log.Debugf("create gadget instance: %+v", req.GadgetInstance.GadgetConfig)
 
+	// Check whether the gadget instance with the same name already exists
+	instances, err := s.ListGadgetInstances(ctx, &api.ListGadgetInstancesRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("listing gadget instances: %w", err)
+	}
+
+	for _, instance := range instances.GadgetInstances {
+		if instance.Name == req.GadgetInstance.Name {
+			return nil, fmt.Errorf("gadget instance with name '%s' already exists", req.GadgetInstance.Name)
+		}
+	}
 	tmpTrue := true
 	cmap := &corev1.ConfigMap{
 		TypeMeta: v1.TypeMeta{
@@ -260,7 +271,7 @@ func (s *Store) CreateGadgetInstance(ctx context.Context, req *api.CreateGadgetI
 		BinaryData: nil,
 	}
 
-	_, err := s.clientset.CoreV1().ConfigMaps(gadgetNamespace).Create(ctx, cmap, v1.CreateOptions{})
+	_, err = s.clientset.CoreV1().ConfigMaps(gadgetNamespace).Create(ctx, cmap, v1.CreateOptions{})
 	if err != nil {
 		return nil, err
 	}


pkg/runtime/grpc/instances.go
```diff
@@ -118,7 +118,16 @@ func (r *Runtime) createGadgetInstance(gadgetCtx runtime.GadgetContext, runtimeP
...
+	// checks if the instance name already exists
+	instances, err := r.GetGadgetInstances(gadgetCtx.Context(), runtimeParams)
+	if err != nil {
+		return fmt.Errorf("getting gadget instances: %w", err)
+	}
+	for _, instance := range instances {
+		if instance.Name == instanceName {
+			return fmt.Errorf("gadget instance with name %q already exists", instanceName)
+		}
+	}
```
mauriciovasquezbernal, line 130: I think it should be implemented server side on https://github.com/inspektor-gadget/inspektor-gadget/blob/7eafb8e0b244bfe6fecf1bfffb139e447bf3a4a5/pkg/gadget-service/store/k8s-configmap-store/configmapstore.go#L234

Look how it's done for the filestore implementation: 

https://github.com/inspektor-gadget/inspektor-gadget/blob/06f618de976fb77420ac70dde0d07d1a2996cd70/pkg/gadget-service/store/file-store/filestore.go#L136-L143

KapilSareen: Done! @mauriciovasquezbernal, let me know your thoughts on the changes now.

pkg/gadget-service/store/k8s-configmap-store/configmapstore.go
```diff
@@ -234,6 +234,17 @@ func (s *Store) handleErr(err error, key string) {
 func (s *Store) CreateGadgetInstance(ctx context.Context, req *api.CreateGadgetInstanceRequest) (*api.CreateGadgetInstanceResponse, error) {
 	log.Debugf("create gadget instance: %+v", req.GadgetInstance.GadgetConfig)
 
+	// Check whether the gadget instance with the same name already exists
+	instances, err := s.ListGadgetInstances(ctx, &api.ListGadgetInstancesRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list gadget instances: %w", err)
```
mauriciovasquezbernal, line 240:
```suggestion
		return nil, fmt.Errorf("listing gadget instances: %w", err)
```

(to make the semgrep on the CI happy)

mauriciovasquezbernal (commented):
Just a minor nit. I'll approve and merge once the CI is green.

pkg/gadget-service/store/k8s-configmap-store/configmapstore.go
KapilSareen: Done!

mauriciovasquezbernal (approved):
Thanks for the contribution. 

@flyth feel free to chime in if you have thoughts about this later on.

PR #3841, opened 2024-12-26
tree-wide: Remove usage of multierr

mauriciovasquezbernal:
Use the standard lib errors.Join() instead.


burak-ok (approved):
Nice find that Go can do that without and libs

mauriciovasquezbernal:
> Nice find that Go can do that without and libs

All credit goes to @flyth https://github.com/inspektor-gadget/inspektor-gadget/pull/3778#discussion_r1894094620 :)

PR #3842, opened 2024-12-27
pkg/oci: Use a file lock to avoid concurrency issues

### LINKED ISSUE #3554: `ig image build` doesn't work well when executed on parallel
It's not possible to run multiple instances of `ig image build` in parallel as some of them fail.

### How to reproduce 

Two different ways to reproduce it (they are the same issue in the end, but want to highlight it's not related to our makefile but to the ig image build command itself)

#### Using parallel 

```bash 
# change directory to any gadget 
$ cd mygadget

$  parallel -j 16 -i sh -c 'sudo ig image build . -t myfoo{}' -- {1..128}
```

After some tries, errors like "Error: getting oci store: invalid OCI Image Index: failed to decode index file: unexpected EOF" appear

#### Building gadgets

```bash 
$ cd gadgets
$ make -j build 
``` 

Same errors like " Error: getting oci store: invalid OCI Image Index: failed to decode index file: invalid character '}' after array element" happen after a while.

## Impact 

After 9f4cc8109141ef2d54734c0c8a4147034561321f the CI is failing a lot with this kind of errors:
- "Error: getting oci store: invalid OCI Image Index: failed to decode index file: EOF" https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/11220516775/job/31192782632?pr=3551
- "Error: pushing gadget: copying to remote repository: failed to resolve ttl.sh/cada60c5ce98aad5ff6168e93d33f6b1de31028c/top_tcp:3553-merge: not found" https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/11214976414/job/31171311087

mauriciovasquezbernal:
Invoking multiple instances of the ig binary on parallel causes issues
as the oras library doesn't provide any multi process syncronization
guarantees on the index.json file. This commit fixes that by using a
file lock and some logic to detect when this file has been updated by
another process.

Specifically, this commit:
- Disables AutoSaveIndex to have control over when the index.json file
  is written.
- Uses a file lock to be sure only one process is writing to the
  index.json file at a time.
- Adds some logic to retry an operation when the index.json file has been
  updated by another process.

Different alternatives were considered:
- big lock: It caused issues as the critical section was too big and
  different instances of ig timeout while running gadgets.
- merge index: Instead of retry the whole operation when the index.json
  file was written by another process, we could merge it with the changes.
  However it's difficult to implement as it'll need the oras library to
  provide a list of changes on the index to apply.
- use a db to store the index: It probably requires a lot of changes to
  the oras library.

## Testing 

All the cases described here work fine with this PR, but fail before it

### Multiple ig instances run on parallel 

Try executing `ig image` commands at the same time, like building multiple gadgets and listing all gadgets:

#### ig build on parallel

terminal 1:

```bash 
$ while true; do sudo ig image list > /dev/null; done
```

terminal 2

```bash 
$ make -C gadgets/ -j 8 build
...
```

#### Running multiple gadgets at the same time with the oci store is empty

```bash 
$ sudo rm -rf /var/lib/ig
$ sudo ig run trace_open & sudo ig run trace_exec
[1] 139772
RUNTIME.CONTAINERNAME          COMM                          PID              TID               FD FNAME                            MODE             ERROR
RUNTIME.CONTAINERNAME                                    COMM                    PID        TID ARGS                               ERRO
```

## ig-k8s/ig daemon and ig run at the same time 

Let's start with an empty oci store

```bash 
$ sudo rm -rf /var/lib/ig
$ sudo ig image list
REPOSITORY                                TAG                                      DIGEST       CREATED
```

Then, run ig daemon

```bash 
$ sudo ig daemon
INFO[0000] starting Inspektor Gadget daemon at "unix:///var/run/ig/ig.socket"
WARN[0000] reading config: open /root/.ig/config.yaml: no such file or directory
WARN[0000] no TLS configuration provided, communication between daemon and CLI will not be encrypted
```

On another terminal, try to run a gadget without pulling the image:

```bash 
$ sudo gadgetctl run trace_open --pull never
Error: fetching gadget information: getting gadget info: rpc error: code = Unknown desc = getting gadget info: initializing and preparing operators: instantiating operator "oci": ensuring image: resolving image "ghcr.io/inspektor-gadget/gadget/trace_open:latest" on local registry: not found
```

Pull the image with ig 

```bash 
$ sudo ig image pull trace_open
Pulling trace_open...
Successfully pulled ghcr.io/inspektor-gadget/gadget/trace_open:latest@sha256:b4deed0f11ddc4b365aa492d2b1c73f9e7d3cb3ad2138a83024c6414ea68a120
```

Running the gadget again works:

```bash
$ sudo gadgetctl run trace_open --pull never
RUNTIME.CONTAINERNAME          COMM                    PID        TID  FD FNAME                            MODE             ERROR
```

It means `ig daemon` has visibility on the changes make by ig to the store without having to restart it.

Fixes #3554

Ref https://github.com/oras-project/oras-go/issues/472#issuecomment-1876251180
Ref https://github.com/oras-project/oras-go/issues/286
Ref https://github.com/ratify-project/ratify/issues/1110
Ref https://github.com/opencontainers/umoci/issues/216#issuecomment-343332733
Ref https://github.com/opencontainers/umoci/issues/146

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/fsnotify/fsnotify v1.8.0
 	github.com/giantswarm/crd-docs-generator v0.11.2
 	github.com/godbus/dbus/v5 v5.1.0
+	github.com/gofrs/flock v0.12.1
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/gopacket/gopacket v1.3.1
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -33,7 +33,6 @@ import (
 	"runtime"
 	"slices"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/distribution/reference"
@@ -96,15 +95,6 @@ const (
 	localhost = "localhost"
 )
 
-// getLocalOciStore returns a single local oci store. oci.Store is concurrently safe only
-// against its own instance inside the same go program
-var getLocalOciStore = sync.OnceValues(func() (*oci.Store, error) {
-	if err := os.MkdirAll(filepath.Dir(defaultOciStore), 0o700); err != nil {
-		return nil, err
-	}
-	return oci.New(defaultOciStore)
-})
-
 // GadgetImageDesc is the description of a gadget image.
 type GadgetImageDesc struct {
 	Repository string `column:"repository"`
@@ -126,12 +116,32 @@ func getTimeFromAnnotations(annotations map[string]string) string {
 
 // PullGadgetImage pulls the gadget image into the local oci store and returns its descriptor.
 func PullGadgetImage(ctx context.Context, image string, authOpts *AuthOptions) (*GadgetImageDesc, error) {
-	ociStore, err := getLocalOciStore()
+	var desc *GadgetImageDesc
+	err := retry("PullGadgetImage", func() error {
+		var err error
+		desc, err = pullGadgetImage(ctx, image, authOpts)
+		return err
+	})
+
+	return desc, err
+}
+
+func pullGadgetImage(ctx context.Context, image string, authOpts *AuthOptions) (*GadgetImageDesc, error) {
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
 
-	return pullGadgetImageToStore(ctx, ociStore, image, authOpts)
+	desc, err := pullGadgetImageToStore(ctx, ociStore.Store, image, authOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := ociStore.saveIndexWithLock(); err != nil {
+		return nil, err
+	}
+
+	return desc, nil
 }
 
 // pullGadgetImageToStore pulls the gadget image into the given store and returns its descriptor.
@@ -201,7 +211,7 @@ func pullImage(ctx context.Context, targetImage reference.Named, imageStore oras
 
 // PushGadgetImage pushes the gadget image and returns its descriptor.
 func PushGadgetImage(ctx context.Context, image string, authOpts *AuthOptions) (*GadgetImageDesc, error) {
-	ociStore, err := getLocalOciStore()
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
@@ -233,6 +243,17 @@ func PushGadgetImage(ctx context.Context, image string, authOpts *AuthOptions) (
 
 // TagGadgetImage tags the src image with the dst image.
 func TagGadgetImage(ctx context.Context, srcImage, dstImage string) (*GadgetImageDesc, error) {
+	var desc *GadgetImageDesc
+	err := retry("TagGadgetImage", func() error {
+		var err error
+		desc, err = tagGadgetImage(ctx, srcImage, dstImage)
+		return err
+	})
+
+	return desc, err
+}
+
+func tagGadgetImage(ctx context.Context, srcImage, dstImage string) (*GadgetImageDesc, error) {
 	src, err := normalizeImageName(srcImage)
 	if err != nil {
 		return nil, fmt.Errorf("normalizing src image: %w", err)
@@ -242,7 +263,7 @@ func TagGadgetImage(ctx context.Context, srcImage, dstImage string) (*GadgetImag
 		return nil, fmt.Errorf("normalizing dst image: %w", err)
 	}
 
-	ociStore, err := getLocalOciStore()
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
@@ -252,7 +273,14 @@ func TagGadgetImage(ctx context.Context, srcImage, dstImage string) (*GadgetImag
 		// Error message not that helpful
 		return nil, fmt.Errorf("resolving src: %w", err)
 	}
-	ociStore.Tag(context.TODO(), targetDescriptor, dst.String())
+
+	if err := ociStore.Tag(context.TODO(), targetDescriptor, dst.String()); err != nil {
+		return nil, fmt.Errorf("tagging image: %w", err)
+	}
+
+	if err := ociStore.saveIndexWithLock(); err != nil {
+		return nil, err
+	}
 
 	imageDesc := &GadgetImageDesc{
 		Repository: dst.Name(),
@@ -296,7 +324,7 @@ func sortIndex(indexPath string) (*ocispec.Index, error) {
 }
 
 func ExportGadgetImages(ctx context.Context, dstFile string, images ...string) error {
-	ociStore, err := getLocalOciStore()
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return fmt.Errorf("getting oci store: %w", err)
 	}
@@ -353,7 +381,7 @@ func ImportGadgetImages(ctx context.Context, srcFile string) ([]string, error) {
 		return nil, fmt.Errorf("loading src bundle: %w", err)
 	}
 
-	ociStore, err := getLocalOciStore()
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
@@ -482,12 +510,12 @@ func getGadgetImages(ctx context.Context, store *oci.Store) ([]*GadgetImageDesc,
 
 // GetGadgetImages gets all the gadget images.
 func GetGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
-	ociStore, err := getLocalOciStore()
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
 
-	images, err := getGadgetImages(ctx, ociStore)
+	images, err := getGadgetImages(ctx, ociStore.Store)
 	if err != nil {
 		return nil, fmt.Errorf("listing all tags: %w", err)
 	}
@@ -500,7 +528,7 @@ func GetGadgetImages(ctx context.Context) ([]*GadgetImageDesc, error) {
 }
 
 func GetGadgetImageDesc(ctx context.Context, image string) (*GadgetImageDesc, error) {
-	ociStore, err := getLocalOciStore()
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
@@ -511,7 +539,7 @@ func GetGadgetImageDesc(ctx context.Context, image string) (*GadgetImageDesc, er
 	}
 
 	fullName := targetImage.String()
-	desc, err := getGadgetImageDescriptor(ctx, ociStore, fullName)
+	desc, err := getGadgetImageDescriptor(ctx, ociStore.Store, fullName)
 	if err != nil {
 		return nil, fmt.Errorf("getting gadget image descriptor: %w", err)
 	}
@@ -521,7 +549,13 @@ func GetGadgetImageDesc(ctx context.Context, image string) (*GadgetImageDesc, er
 
 // DeleteGadgetImage removes the given image.
 func DeleteGadgetImage(ctx context.Context, image string) error {
-	ociStore, err := getLocalOciStore()
+	return retry("DeleteGadgetImage", func() error {
+		return deleteGadgetImage(ctx, image)
+	})
+}
+
+func deleteGadgetImage(ctx context.Context, image string) error {
+	ociStore, err := newLocalOciStore()
 	if err != nil {
 		return fmt.Errorf("getting oci store: %w", err)
 	}
@@ -537,7 +571,7 @@ func DeleteGadgetImage(ctx context.Context, image string) error {
 		return fmt.Errorf("resolving image: %w", err)
 	}
 
-	images, err := getGadgetImages(ctx, ociStore)
+	images, err := getGadgetImages(ctx, ociStore.Store)
 	if err != nil {
 		return fmt.Errorf("listing images: %w", err)
 	}
@@ -555,15 +589,23 @@ func DeleteGadgetImage(ctx context.Context, image string) error {
 			// Where foo and bar are different names referencing the same image, as
 			// the digest shows.
 			// In this case, we just untag the image name given by the user.
-			return ociStore.Untag(ctx, fullName)
+			if err := ociStore.Untag(ctx, fullName); err != nil {
+				return err
+			}
+
+			return ociStore.saveIndexWithLock()
 		}
 	}
 
-	err = ociStore.Delete(ctx, descriptor)
-	if err != nil {
+	if err = ociStore.Delete(ctx, descriptor); err != nil {
 		return err
 	}
 
+	if err := ociStore.saveIndexWithLock(); err != nil {
+		return err
+	}
+
+	// TODO: GC() could race with other processes calling it a the same time.
 	return ociStore.GC(ctx)
 }
 
@@ -976,12 +1018,18 @@ func ensureImage(ctx context.Context, imageStore oras.Target, image string, imgO
 
 // EnsureImage ensures the image is present in the local store
 func EnsureImage(ctx context.Context, image string, imgOpts *ImageOptions, pullPolicy string) error {
-	imageStore, err := getLocalOciStore()
-	if err != nil {
-		return fmt.Errorf("getting local oci store: %w", err)
-	}
+	return retry("EnsureImage", func() error {
+		imageStore, err := newLocalOciStore()
+		if err != nil {
+			return fmt.Errorf("getting local oci store: %w", err)
+		}
 
-	return ensureImage(ctx, imageStore, image, imgOpts, pullPolicy)
+		if err := ensureImage(ctx, imageStore, image, imgOpts, pullPolicy); err != nil {
+			return err
+		}
+
+		return imageStore.saveIndexWithLock()
+	})
 }
 
 func getManifestForHost(ctx context.Context, target oras.ReadOnlyTarget, image string) (*ocispec.Manifest, error) {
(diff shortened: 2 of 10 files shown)


burak-ok (commented):
There is another exported function from the `oci.go` file: `GetContentFromDescriptor`

I think it would also need the locking, right?

burak-ok:
While thinking about it: Do we want to have a timeout while we wait on the lock?
If the file somehow didn't get unlocked we should notify the user instead of hanging indefinitly and giving no feedback

alban:
I wonder if the 5-second timeout is too short: pulling an OCI image on a slow network could take more than 5 seconds. Another instance of ig might want to wait for longer before giving up.

Alternatively, is it possible to download a large image in a temporary location before saving it into the OCI store? That might minimize the lock time.

mauriciovasquezbernal:
Good point about pulling the image. The main issue is to have concurrent writes/reads to the index.json file. I pushed a different version that uses `AutoSaveIndex: false` and manually saves the index while holding the lock, hence it's held by a very short period of time avoiding timeout issues.

pkg/oci/oci.go
```diff
@@ -102,7 +106,25 @@ var getLocalOciStore = sync.OnceValues(func() (*oci.Store, error) {
 	if err := os.MkdirAll(filepath.Dir(defaultOciStore), 0o700); err != nil {
 		return nil, err
 	}
-	return oci.New(defaultOciStore)
+
+	// calling oci.New causes it to read the index.json file, lock to avoid it
+	// messing up with other instances writing this
```
burak-ok, line 111: This comment makes me unsure if this is truly concurrent. According to that comment `oci.New` is the point where the index is read. With the following codepath we get in troublt:

t1=thread 1, t2=thread2
1. local oci store is empty
2. t1 calls getLocalOciStore(), it thinks there are no images
3. t2 calls getLocalOciStore(), it thinks there are no images
4. t1 downloads image 1 and saves it to the index
5. t2 downloads image 2 and saves it to the index. Since it read the index before 4, it will overwrite the index and save only image 2

mauriciovasquezbernal: Ah, you're totally right. I switched back to an approach that locks the operation from the beginning to the end (includes the time needed to pull the image). I was checking around other options but I don't see any **easy** solution. Pulling the image to a temporary store seems also difficult to handle. IMHO this solution is good enough for now.

pkg/oci/oci.go
```diff
@@ -96,18 +103,42 @@ const (
...
+	*oci.Store
+	lock *flock.Flock
+}
+
+func (s *localOciStore) Unlock() error {
+	return s.lock.Unlock()
+}
+
+// getLocalOciStore returns a single local oci store. oci.Store is concurrently
+// safe only against its own instance inside the same go program. The caller
```
burak-ok, line 116:
```suggestion
// getLocalOciStore returns a single local oci store. oci.Store is also concurrently
// safe against other ig processes through a file lock. The caller
```

pkg/oci/oci.go
```diff
@@ -96,18 +103,42 @@ const (
...
+// musth call Unlock() when done with the store.
+var getLocalOciStore = sync.OnceValues(func() (*localOciStore, error) {
```
burak-ok, line 118: The function inside the `sync.OnceValues` is only called once.
It doesn't make sense to call `l.TryLockContext` or any locking here.

We need to also have a `Lock()` method for `localOciStore` it seems. The `sync.OnceValues` was introduced to have the directory creation only once and only have a single instance of `oci.Store`

pkg/oci/oci.go
burak-ok: Now after I have written https://github.com/inspektor-gadget/inspektor-gadget/pull/3842#discussion_r1908496461 the above scenario only comes up with multiple processes, not threads

mauriciovasquezbernal: I assumed you're referring to multiple processes in your comment above. You're right, the difficult part is to handle multiple processes, the sync.Once and the internal mutex used by oci.New() protects it on the same process.

pkg/oci/oci.go
burak-ok: I meant we need to have a `.Lock()` method on the `localOciStore` because the contents of the `sync.Once` is only run once.

So if you have multiple `getLocalOciStore()` currently this PR only locks the file when the function is called the first time in the programs lifetime.
Therefore the problem happens when `getLocalOciStore()` is called multiple times in a single instance

mauriciovasquezbernal: Ah, I got you! The difficulty is that, as described above, the lock needs to be taken before calling `oci.New()`. Now that I think better, the `sync.Once` approach causes the following issue:

```bash 
# image store is empty 
$ sudo ig image list
REPOSITORY                                TAG                                      DIGEST       CREATED

# on another terminal run ig as daemon
$ sudo ig daemon --verify-image=false
INFO[0000] starting Inspektor Gadget daemon at "unix:///var/run/ig/ig.socket"
WARN[0000] reading config: open /root/.ig/config.yaml: no such file or directory
WARN[0000] no TLS configuration provided, communication between daemon and CLI will not be encrypted

# run a gadget

$ sudo gadgetctl run trace_open
WARN[0002] local                | image signature verification is disabled due to using corresponding option
RUNTIME.CONTAINERNAME          COMM                    PID        TID  FD FNAME                            MODE             ERROR

# pull another gadget to the store 
$ sudo ig image pull trace_exec
Pulling trace_exec...
Successfully pulled ghcr.io/inspektor-gadget/gadget/trace_exec:latest@sha256:5f4e9408e8d15b32b736a02473f523e6cc4c459dd203489f4653156c0cf44a29
$ sudo ig image list
REPOSITORY                                TAG                                      DIGEST       CREATED
trace_exec                                latest                                   5f4e9408e8d1 52 minutes ago
trace_open                                latest                                   6064950f49aa 52 minutes ago

# Trying to run the second image will fail
$ sudo gadgetctl run trace_exec --pull never
Error: fetching gadget information: getting gadget info: rpc error: code = Unknown desc = getting gadget info: initializing and preparing operators: instantiating operator "oci": ensuring image: resolving image "ghcr.io/inspektor-gadget/gadget/trace_exec:latest" on local registry: not found
```

I can think of other cases as well: (untested)

```bash 
# ig running as daemon
$ sudo ig daemon

# other instances of ig modify the store
$ ig image pull 
$ ig image build 
$ ig run ...


# run a gadget with ig daemon that requires pulling it 
$ sudo gadgetctl run foo --pull always

# ig daemon stops 
# all work done by ig above is lost as the index is overwriten by ig daemon
```

So I think we need to get rid of this sync.Once, it'll be more inefficient but I don't see other options with the current model we have.

mauriciovasquezbernal:
I had to change this again. I've switched to using different instances of `oci.Store` in the same process and protecting them with a mutex, it turned to be a bottleneck and the tests were failing with timeouts. Now I'm using a middle ground implementation: for the same process we use the same `oci.Store` instance and rely on it for the synchronization, for different processes we use a file lock. 

I'm aware of the limitations of this implementation: (1) ig can't be run when ig-k8s or ig daemon is running and (2) multiple instances of ig running at the same time will have big contention. Unfortunately, I couldn't find a better solution, and according to https://github.com/oras-project/oras-go/issues/472#issuecomment-1876251180 "Multi-process support is also another non-trivial work and it is not on the roadmap of oras-go as it requires an atomicity source such as databases, file systems with O_EXCL support, and etc.", so this PR is a kind of best-effort approach for now.

cmd/ig/daemon.go
```diff
@@ -171,6 +173,11 @@ All these options should be set at the same time to enable TLS connection`,
 			log.Warnf("no TLS configuration provided, communication between daemon and CLI will not be encrypted")
 		}
 
+		if err := oci.Lock(context.Background()); err != nil {
+			return fmt.Errorf("locking OCI storage: %w", err)
+		}
+		defer oci.Unlock()
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Please see https://github.com/inspektor-gadget/inspektor-gadget/pull/3842#issuecomment-2583234332 with some reasoning about it. In other words, what you're describing is the ideal case, but it's very difficult to implement (at least, I haven't found a solution for it).

pkg/oci/oci.go
```diff
@@ -242,6 +291,11 @@ func TagGadgetImage(ctx context.Context, srcImage, dstImage string) (*GadgetImag
 		return nil, fmt.Errorf("normalizing dst image: %w", err)
 	}
 
+	if err := Lock(ctx); err != nil {
+		return nil, err
+	}
+	defer Unlock()
+
 	ociStore, err := getLocalOciStore()
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I tried something similar on https://github.com/inspektor-gadget/inspektor-gadget/commit/5bd114573b0bb8250b93ebc6246ca8bc0306bfc0, but I hit some issues described on https://github.com/inspektor-gadget/inspektor-gadget/pull/3842#issuecomment-2583234332 and I have to switch it have a separate Lock() function to be called from outside.

cmd/ig/daemon.go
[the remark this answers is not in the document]
mauriciovasquezbernal: > I do not like "big kernel lock"-like solution.

I agree. However it doesn't seem to be trivial and this PR is just preventing the user from running something racy, like ig-k8s and ig on the same machine. 

> We already contributed to oras-go, I would reach them and ask whether they see big blockers toward this.
Having a good solution upstream would be better for everyone.

I'll reach to them to see if they have any ideas on how to implement it.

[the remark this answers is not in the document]
mauriciovasquezbernal: I did some research (ref links on PR description) and updated this PR to use a retry approach. I've been testing it and it seems to be working fine. There is one test failing, but I'm confident it can be fixed.

pkg/oci/helpers.go
```diff
@@ -0,0 +1,125 @@
...
+	"reflect"
+	"time"
+
+	"github.com/gofrs/flock"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content/oci"
+)
+
+const (
+	retryLimit = 20
```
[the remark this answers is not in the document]
mauriciovasquezbernal: There is not any guarantee that this retry mechanism will always work. I chose those numbers following my "gut feeling". 20 tries each 500 milliseconds is 10 seconds, a reasonable timeout the user could wait for IMHO.

I was wondering if an exponential backoff algorithm will work better, but I'm not sure it's worth it. Also, please consider that this PR is trying to fix a race, so I'd rather have an imperfect mechanism than something that is racy. If this retry fails, the user can try to run the operation (pull, build, tag, etc...) again.

pkg/oci/oci.go
```diff
@@ -126,12 +122,32 @@ func getTimeFromAnnotations(annotations map[string]string) string {
...
 
-	return pullGadgetImageToStore(ctx, ociStore, image, authOpts)
+	desc, err := pullGadgetImageToStore(ctx, ociStore.Store, image, authOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := ociStore.SaveIndexWithLock(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I don't think there is a big difference in both cases, even when using `AutoSaveIndex` it should happen that the computer crashes after pulling the image and before saving the index, like [before this line](https://github.com/oras-project/oras-go/blob/6fce64e956e850a3af7ea16e85959c23e3eddc4b/copy.go#L469)

Btw, if there is a crash before saving the index, it shouldn't be big issue, the user could run the command again and it'll work, i.e. the store won't be on a corrupted state. This because the oras-go library is smart enough to detect if a blob is already present on the system, so it'll avoid pulling it again and will just update the index.

cmd/ig/daemon.go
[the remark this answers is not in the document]
mauriciovasquezbernal: > There is one test failing, but I'm confident it can be fixed.

It should be fixed by the latest push.

go.mod
```diff
@@ -80,6 +80,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require github.com/gofrs/flock v0.12.1
```
burak-ok, line 83: Put this in the above `require` block?

pkg/oci/oci.go
```diff
@@ -98,12 +97,9 @@ const (
 
 // getLocalOciStore returns a single local oci store. oci.Store is concurrently safe only
```
burak-ok, line 98: This is not `a single local oci` store anymore. We are creating new ones every time in this PR

pkg/oci/oci.go
```diff
@@ -126,12 +122,32 @@ func getTimeFromAnnotations(annotations map[string]string) string {
...
 		return nil, fmt.Errorf("getting oci store: %w", err)
 	}
 
-	return pullGadgetImageToStore(ctx, ociStore, image, authOpts)
+	desc, err := pullGadgetImageToStore(ctx, ociStore.Store, image, authOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := ociStore.SaveIndexWithLock(); err != nil {
```
burak-ok, line 146: I don't see how this PR solves https://github.com/inspektor-gadget/inspektor-gadget/pull/3842#discussion_r1906998862

`getLocalOciStore` reads the oci store locally in a syncronous manner itself. But after the call another process can read the store too. If now both pull different images and save their indexes wouldn't we miss an image in the store index? (See linked comment)

mauriciovasquezbernal: > If now both pull different images and save their indexes wouldn't we miss an image in the store index?

The `SaveIndexWithLock` function checks if the index was changed by another process before writing to it, if it was changed, it returns `ErrRetry` forcing the operation to be retried again.

burak-ok: Somehow missed the DeepEqual +1

pkg/oci/oci.go
```diff
@@ -537,11 +577,15 @@ func DeleteGadgetImage(ctx context.Context, image string) error {
...
 	}
 
-	images, err := getGadgetImages(ctx, ociStore)
+	images, err := getGadgetImages(ctx, ociStore.Store)
 	if err != nil {
 		return fmt.Errorf("listing images: %w", err)
 	}
 
+	defer func() {
+		err = ociStore.SaveIndexWithLock()
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Ah, good catch Alban. I'm wondering if this is a good patter for the defer call, or just better to call it multiple times inside the function. I personally find this difficult to follow.

pkg/oci/helpers.go
```diff
@@ -0,0 +1,127 @@
...
+	log "github.com/sirupsen/logrus"
+	"oras.land/oras-go/v2/content/oci"
+)
+
+const (
+	retryLimit = 20
+	retryDelay = 500 * time.Millisecond
+)
+
+var ErrRetry = errors.New("retry")
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > should it be exported?

The initial idea was that the caller outside the package should handle it, but now it's only used internally. Updated it and added the documentation. Thanks!

pkg/oci/retry.go
```diff
@@ -0,0 +1,77 @@
...
+const (
+	retryLimit = 20
+	retryDelay = 500 * time.Millisecond
+)
+
+// errRetry hints that the function can safely be tried again after a small
+// time.
+var errRetry = errors.New("retry")
+
+// ErrRetryLimitExceeded indicates the function was retry too many times.
```
alban, line 37:
```suggestion
// ErrRetryLimitExceeded indicates the function was retried too many times.
```

pkg/oci/retry.go
```diff
@@ -0,0 +1,77 @@
...
+var ErrRetryLimitExceeded = errors.New("retry limit exceeded")
```
alban, line 38:
```suggestion
var errRetryLimitExceeded = errors.New("retry limit exceeded")
```

pkg/oci/oci.go
```diff
@@ -252,7 +273,14 @@ func TagGadgetImage(ctx context.Context, srcImage, dstImage string) (*GadgetImag
 		// Error message not that helpful
 		return nil, fmt.Errorf("resolving src: %w", err)
 	}
-	ociStore.Tag(context.TODO(), targetDescriptor, dst.String())
+
+	if err := ociStore.Tag(context.TODO(), targetDescriptor, dst.String()); err != nil {
+		return nil, fmt.Errorf("tagging image: %w", err)
```
alban, line 278: Does `Tag` fail if it is already tagged by a previous attempt?

pkg/oci/oci.go
```diff
@@ -537,11 +571,18 @@ func DeleteGadgetImage(ctx context.Context, image string) error {
...
+	images, err := getGadgetImages(ctx, ociStore.Store)
 	if err != nil {
 		return fmt.Errorf("listing images: %w", err)
 	}
 
+	defer func() {
+		if err != nil {
+			return
+		}
+		err = ociStore.saveIndexWithLock()
```
alban, line 583: This function calls `ociStore.GC(ctx)` line 608.

But the documentation of `GC` says:
```go
// GC removes garbage from Store. Unsaved index will be lost. To prevent unexpected
// loss, call SaveIndex() before GC or set AutoSaveIndex to true.
```
With this PR, SaveIndex() is no longer called **before** GC().

pkg/oci/retry.go
mauriciovasquezbernal: I think it's fine to have this public. Callers could use it.

pkg/oci/oci.go
mauriciovasquezbernal: It appears to be idempotent:

```bash 
$ sudo ig image tag trace_open foo
$ sudo ig image tag trace_open foo
$ sudo ig image tag trace_open foo

File: pkg/oci/oci.go (hunk 16 above)
Comment (mauriciovasquezbernal) [reply pkg/oci/oci.go:583]: Good point. I've read this comment but I didn't get the idea before. I updated it to call `saveIndexWithLock` before GC(). I think GC() can still be racy at this point, but I'll explicitly ignore this for now.

File: pkg/oci/oci.go (hunk 11 above)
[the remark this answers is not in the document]
Comment (mauriciovasquezbernal) [reply pkg/oci/oci.go]: Removed after handling the comment on https://github.com/inspektor-gadget/inspektor-gadget/pull/3842#discussion_r1934395478

File: pkg/oci/retry.go [hunk 17]
```diff
@@ -0,0 +1,77 @@
...
+
+	defer indexFile.Close()
+
+	var index ocispec.Index
+	if err := json.NewDecoder(indexFile).Decode(&index); err != nil {
+		return nil, fmt.Errorf("decoding index file: %w", err)
+	}
+
+	return &index, nil
+}
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Absolutely!

mauriciovasquezbernal:
I've seen a snapshot process fail twice here, let's merge it and monitor to see if it continues happening.

PR #3843, opened 2024-12-28
Refactor LocalManger to be used from package level variable only

Sefi4:
LocalManger operator caches k8s/container runtime information. This PR un-export LocalManger struct and instead introduces a package level variable, to be used by Golang users of IG platform. This will make sure, Golang users of localManager operator will share the localManger cache for all running gadgets.

## How to use
Examples refactored to demonstrate usage

### DIFF
--- a/examples/gadgets/grpc/custom_operator/main.go
+++ b/examples/gadgets/grpc/custom_operator/main.go
@@ -68,11 +68,7 @@ func do() error {
 	}
 	defer runtime.Close()
 
-	params := map[string]string{
-		// Capture events coming from the host too
-		"operator.LocalManager.host": "true",
-	}
-	if err := runtime.RunGadget(gadgetCtx, nil, params); err != nil {
+	if err := runtime.RunGadget(gadgetCtx, nil, nil); err != nil {
 		return fmt.Errorf("running gadget: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


examples/gadgets/grpc/custom_operator/main.go
```diff
@@ -68,11 +68,7 @@ func do() error {
 	}
 	defer runtime.Close()
 
-	params := map[string]string{
-		// Capture events coming from the host too
-		"operator.LocalManager.host": "true",
-	}
-	if err := runtime.RunGadget(gadgetCtx, nil, params); err != nil {
+	if err := runtime.RunGadget(gadgetCtx, nil, nil); err != nil {
```
Sefi4, line 71: localmanager isn't used in this example, so I thought to remove params to avoid confusion

mauriciovasquezbernal:
It makes sense to me. I don't see why a user would like to have more than one instance of the local manager in the same process, if that needed arises before we can always introduce a New() function to allocate and return it.

Sefi4:
> It makes sense to me.

Awesome, I rebased. Can you please run the CI?

mauriciovasquezbernal (approved):
LGTM, thanks!

PR #3847, opened 2025-01-02
pkg: Rework error message to cope with code scanning alerts.

eiffel-fl:
Code scanning alerted about logging of sensitive information but this is actually a false positive as we do not print the content of /etc/passwd in host file system but only its path, so something like /host/etc/passwd.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/security/code-scanning/277


burak-ok (approved):
I thought I marked this as a false positive some time ago.
I guess its fine to hardcode the path. LGTM

PR #3852, opened 2025-01-06
gadgetctl/kubectl-gadget: fixes crash when running attach right after run 

### LINKED ISSUE #3678: gadgetctl/kubectl-gadget: Crash when running `attach` right after `run`
Consider the following manifest:

```yaml
# manifest.yaml
apiVersion: 1
kind: instance-spec
image: trace_exec
name: mygadgetinstance
tags:
  - foo
  - bar
paramValues:
```

```bash
$ sudo gadgetctl run -f manifest.yaml --detach && sudo gadgetctl attach mygadgetinstance
INFO[0000] installed as "729e15bb1655c84211a742c8ff7479a7"
[signal SIGSEGV: segmentation violation code=0x1 addr=0xa0 pc=0x307ccc8]

goroutine 1 [running]:
github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context.(*GadgetContext).LoadGadgetInfo(0xc00052c820, 0x0, 0xc000130f00, 0x0)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/gadget-context/gadget-context.go:320 +0xc8
github.com/inspektor-gadget/inspektor-gadget/pkg/runtime/grpc.(*Runtime).GetGadgetInfo(0xc0008b4870, {0x46fe7e0, 0xc00052c820}, 0xd?, 0xc000130f00)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/runtime/grpc/oci.go:62 +0x32c
github.com/inspektor-gadget/inspektor-gadget/cmd/common.NewRunCommand.func1(0xc000754908, {0xc0003104c0, 0x1, 0x1})
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/cmd/common/oci.go:223 +0xc65
github.com/spf13/cobra.(*Command).execute(0xc000754908, {0xc0003104c0, 0x1, 0x1})
        /home/mauriciov/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:970 +0xa44
github.com/spf13/cobra.(*Command).ExecuteC(0xc0004be908)
        /home/mauriciov/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        /home/mauriciov/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
main.main()
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/cmd/gadgetctl/main.go:108 +0x905
````

Adding a sleep between the commands fixes it:

```bash
$ sudo gadgetctl run -f manifest.yaml --detach && sleep 2 && sudo gadgetctl attach mygadgetinstance
INFO[0000] installed as "77922556b8944cff8b2ace4ed3075e62"
RUNTIME.CONTAINERNAME                                    COMM                    PID        TID ARGS                               ERRO
```

KapilSareen:
### Fixes #3678
Uses a channel to signal completion of `Run` function in `instance.go` before returning gadgetinfo. Earlier this returned null, leading to a null pointer dereferencing error.

### Testing done

image

### DIFF
--- a/pkg/gadget-service/instance-manager/instance.go
+++ b/pkg/gadget-service/instance-manager/instance.go
@@ -59,9 +59,11 @@ type GadgetInstance struct {
 	cancel               func()
 	state                gadgetState
 	error                error
+	ready                chan struct{}
 }
 
 func (p *GadgetInstance) GadgetInfo() (*api.GadgetInfo, error) {
+	<-p.ready
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	return p.gadgetInfo, p.error
@@ -186,6 +188,7 @@ func (p *GadgetInstance) Run(
 				Payload: d,
 			}
 			p.gadgetInfo = gi
+			close(p.ready)
 			return nil
 		}),
 	)
--- a/pkg/gadget-service/instance-manager/manager.go
+++ b/pkg/gadget-service/instance-manager/manager.go
@@ -101,6 +101,7 @@ func (m *Manager) RunGadget(instance *api.GadgetInstance) {
 		eventBufferOffs: 0,
 		cancel:          cancel,
 		clients:         map[*GadgetInstanceClient]struct{}{},
+		ready:           make(chan struct{}),
 	}
 	m.mu.Lock()
 	m.gadgetInstances[gi.id] = gi


alban:
I don't know what is the correct solution, but I don't think the sleep workaround is a good solution.

KapilSareen:
Can you suggest something?

alban:
@KapilSareen I am not too familiar with that part of the code, sorry.

@flyth Do you know who could help? `cmd/gadgetctl` is not listed in CODEOWNERS.

KapilSareen:
Sure! I'll look into it very soon.

KapilSareen:
Hey @mauriciovasquezbernal , I debugged the flow and here's what I found:
- The server side is handling it all very well , there is no error 
    >  func (s *Service) GetGadgetInfo(ctx context.Context, req *api.GetGadgetInfoRequest) (*api.GetGadgetInfoResponse, error) { 
 

- However, on the client side when `LoadGadgetInfo` is invoked first time , the `out.GadgetInfo` is set to null  , which leads to the error `panic: runtime error: invalid memory address or nil pointer dereference`.
    >  err = gadgetCtx.LoadGadgetInfo(out.GadgetInfo, paramValues, false) 

And here is how I solved it, I tried running `client.GetGadgetInfo` in a for loop until `out.GadgetInfo` is not null, It works currently as intended , here is the specific code:
```
// pkg/runtime/grpc/oci.go

out := &api.GetGadgetInfoResponse{}
for out.GadgetInfo == nil {
	out, err = client.GetGadgetInfo(gadgetCtx.Context(), in)
	if err != nil {
		return nil, fmt.Errorf("getting gadget info: %w", err)
	}
}
err = gadgetCtx.LoadGadgetInfo(out.GadgetInfo, paramValues, false)

``` 

Do let me know if this solution is good enough.

cc @alban

mauriciovasquezbernal:
I don't think looping is the right solution. If the server is working fine, why sees a nil object? Where is it lost? We need to investigate and find why it's happening.

KapilSareen:
@mauriciovasquezbernal 
I did a bit more research around this and found that the nil is origining around here :

https://github.com/inspektor-gadget/inspektor-gadget/blob/39da0258c782023ab7c5d36dea85bbe16907f7d0/pkg/gadget-service/service-oci.go#L71-L75

The `gi` term is correctly populated; however, when the `run` command has not yet completed, the `gi.GadgetInfo()` function `((*api.GadgetInfo))` fails to populate due to mutex locks.

At the moment, I do not have a solid solution for this issue. Would you recommend closing this PR, or do you have any suggestions to address this problem?

KapilSareen:
Hey @mauriciovasquezbernal , this works for me. Can you confirm?

pkg/gadget-service/instance-manager/instance.go
```diff
@@ -127,6 +129,9 @@ func (p *GadgetInstance) Run(
 	runtime runtime.Runtime,
 	logger logger.Logger,
 ) error {
+	if p.ready == nil {
+		p.ready = make(chan struct{})
+	}
```
mauriciovasquezbernal, line 134: I think it's better to allocate this when the GadgetInstance is created.

KapilSareen: sure

KapilSareen: can you check again?

mauriciovasquezbernal: I meant this: https://github.com/inspektor-gadget/inspektor-gadget/blob/60e33adecb444587c2584420a52215e6d92bf8e6/pkg/gadget-service/instance-manager/manager.go#L95-L104

KapilSareen: hope this one does the job

pkg/gadget-service/instance-manager/instance.go
```diff
@@ -186,6 +188,9 @@ func (p *GadgetInstance) Run(
 				Payload: d,
 			}
 			p.gadgetInfo = gi
+			if p.ready != nil {
```
mauriciovasquezbernal, line 191: can it ever be nil?

mauriciovasquezbernal (commented):
Please also format the commit message with something like

area: what's done

for instance

gadget-service: Fix crash ...

pkg/gadget-service/instance-manager/instance.go
KapilSareen: ahh my bad! earlier it could be, but now doesn't makes sense, let me fix it

mauriciovasquezbernal (approved):
It fixes the issue for me, thanks!

I'll merge this one once CI is green.

PR #3859, opened 2025-01-07
tree-wide: Bump golang to 1.23

mauriciovasquezbernal:
Bump the Golang version in the whole repository, including tools and gadgets.

This is needed because there are a few dependencies that need this Golang version:
- #3814 
- #3789


burak-ok (approved):
I was just about to open a PR for the Go 1.23 update too :D

We got the same number of files and same hashes (Didn't test all)

mauriciovasquezbernal:
I was hitting https://github.com/tetratelabs/wazero/issues/2357, so I needed to downgrade to tinygo 0.34.0 instead of 0.35.0

PR #3863, opened 2025-01-08
Fix hanging on invalid filter, refine multiple filter handling in tra…

Priyansurout:
# Improve error handling for filter operator

Improves error handling for invalid -F filter values in trace_tcp. Instead of hanging indefinitely, the gadget now fails gracefully with a clear error message. Multiple filters also produce consistent results as expected.

## Testing done
`sudo ./ig run trace_tcp trace_tcp -F comm:coredns -F type:connect --verify-image=false
`
output : `"type:connect": invalid filter format: use == instead of :`

Issue #3453, opened 2024-09-06
Improve error handling for filter operator


mqasimsarfraz:
## Current situation

Currently if you provide a broken filter value to `-F` flag the gadget run hangs forever:

```
$ ./kubectl-gadget run trace_tcp -A -F comm:coredns
```

Also, on providing multiple filters it seems I am getting unexpected events e.g (`kube-apis…` events in following output) so what sure what to expect when using multiple filters:

```
$ ./kubectl-gadget run trace_tcp -A -F comm==coredns -F type==connect 
K8S.NODE              K8S.NAMESPACE         K8S.PODNAME           K8S.CONTAINERNAME     SRC                          DST                          COMM              PID        TID        UID        GID TYPE      
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:45496              127.0.0.1:8080               coredns       1108489    1108517      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:45498              127.0.0.1:8080               coredns       1108489    1108518      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:45502              127.0.0.1:8080               coredns       1108489    1108513      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:32908              127.0.0.1:8080               coredns       1108489    1108517      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:32920              127.0.0.1:8080               coredns       1108489    1108518      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:32936              127.0.0.1:8080               coredns       1108489    1108513      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:32950              127.0.0.1:8080               coredns       1108489    1108518      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:32960              127.0.0.1:8080               coredns       1108489    1108515      65532      65532 connect   
minikube-docker       kube-system           coredns-7d…ff4d-m7bws coredns               127.0.0.1:32970              127.0.0.1:8080               coredns       1108489    1108513      65532      65532 connect   
[...]

PR #3865, opened 2025-01-08
oci-service: Fix closing logic

mauriciovasquezbernal:
The previous logic was causing issues when running the gadget failed because the client was handing forever. This is because if `s.runtime.RunGadget` fails early the simple operator won't be initialized and hence the done <- true operation will block as there is not any goroutine reading from the channel.

This commit fixes that by closing the channel instead of writing to it.

Fixes first part of #3453.
Replaces #3863

### DIFF
--- a/pkg/gadget-service/service-oci.go
+++ b/pkg/gadget-service/service-oci.go
@@ -149,7 +149,7 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
 
 	done := make(chan bool)
 	defer func() {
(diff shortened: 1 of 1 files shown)


flyth (approved):
Nice catch!

mauriciovasquezbernal:
Tests passed on the [previous run](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/12677907381?pr=3865). I just pushed a fix to a typo and added the `Fixes` tag, so I'll merge without waiting the CI again.

PR #3799, opened 2024-12-15
Add deadlock integration test

AdvH039:
# Deadlock tests
Refer #3626 

## How to use
#set environment variables first
go test -exec 'sudo -E' gadgets/deadlock/tests/integration/deadlock_test.go -v

## Testing done
ig works but kubectl-gadget is not working currently

PR #3868, opened 2025-01-09
trace exec gadgets: add support for execveat

alban:
Both builtin and image-based trace-exec gadgets were supporting the execve syscall but not execveat. This patch adds support for execveat.

## How to use

No changes.

## Testing done

Using execveat.c as follows:

<details>

```c
#define _GNU_SOURCE

#include <stdio.h>
#include <stdlib.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <string.h>

int main(int argc, char *argv[], char **envp) {
    if (argc < 2) {
        fprintf(stderr, "Usage: %s <program_to_execute>\n", argv[0]);
        exit(1);
    }

    char *method = getenv("METHOD");
    if (method == NULL) {
        fprintf(stderr, "Environment variable METHOD not set.\n");
        exit(1);
    }

    char *program = argv[1];

    if (strcmp(method, "execve") == 0) {
        execve(program, argv+1, envp);
    } else if (strcmp(method, "execveat") == 0) {
        int fd = open(".", O_RDONLY | O_DIRECTORY);
        if (fd == -1) {
            perror("open");
            exit(1);
        }

        if (execveat(fd, program, argv+1, envp, 0) == -1) {
            perror("execveat");
            exit(1);
        }

        close(fd); 
    } else {
        fprintf(stderr, "Invalid value for METHOD environment variable.\n");
        exit(1);
    }

    // execve and execveat should not return unless an error occurs.
    fprintf(stderr, "Error: execve or execveat failed to execute the program.\n");
    exit(1);
}
```

</details>

Compile with:
```bash 
$ gcc -Wall -o execveat execveat.c
```

Generate events with:

```bash
$ METHOD=execve ./execveat /bin/echo "using execve" ; METHOD=execveat ./execveat /bin/echo "using execveat"
using execve
using execveat
```

Check the builtin gadget:

```bash
$ go run -exec sudo ./cmd/ig/... trace exec --host -o columns=comm,pcomm,ret,args
COMM             PCOMM            RET ARGS
execveat         bash             0   ./execveat /bin/echo using execve
echo             bash             0   /bin/echo using execve
execveat         bash             0   ./execveat /bin/echo using execveat
echo             bash             0   /bin/echo using execveat
```

Check the image-based gadget:

```bash
$ sudo ig run ghcr.io/inspektor-gadget/gadget/trace_exec:alban_execveat --host --verify-image=false --fields=comm,pid,args
COMM                    PID ARGS
execveat             158206 ./execveat /bin/echo using execve
echo                 158206 /bin/echo using execve
execveat             158207 ./execveat /bin/echo using execveat
echo                 158207 /bin/echo using execveat
```


mauriciovasquezbernal (approved):
Besides the small nit LGTM from code inspection.

Do you think we can use your test program to extend the integration test. We can compile it on a container as done in gadgets/deadlock/test/integration/deadlock_test.go. WDYT?

alban:
> Do you think we can use your test program to extend the integration test. We can compile it on a container as done in gadgets/deadlock/test/integration/deadlock_test.go. WDYT?

Good idea. I will try that in a follow-up PR.

Note: gadgets/deadlock/test/integration/deadlock_test.go is available in #3799

Issue #3838, opened 2024-12-24
[RFE] Improve trace exec upper layer limitation in case of a script


Sefi4:
## Current situation
Both [built-in](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadgets/builtin/trace/exec.md#:~:text=In%20case%20of%20a%20shell%20script%2C%20the%20upper%20layer%20field%20will%20refer%20to%20the%20location%20of%20the%20shell%20program%20(e.g.%20/bin/sh)%20and%20not%20the%20script%20file.) and [image-based](https://github.com/inspektor-gadget/inspektor-gadget/pull/3818/files#diff-76752672a5e70bb1c1c9ce971f6455abc7e16c5e792c990e653fd01cabd4c565) trace exec gadgets has a limitation in case of a script. Where upper layer refers to the shell instead of the script.

As @alban introduced a hook to the `security_bprm_check` kernel function in this [PR](https://github.com/inspektor-gadget/inspektor-gadget/pull/3818/files), I thought it may be possible to improve upper layer to refer to the actual script instead of the shell.

## Ideal future situation
With below proposed implementation, upper layer may still point to an interpreter instead of the script - depends on how the script ran. If running `/bin/bash script.sh`, then I believe f_node would still point to bash and not to the script's inode, as opposed to running `./script.sh`. Ideally, upper_layer would always refer to the underlying script, though I'm not sure how to achieve this.

## Implementation options
As we already have access to bprm->file, I think we could get access to `f_inode = BPF_CORE_READ(bprm, file, f_inode);` Then we could pass the `f_inode` to the [has_upper_layer](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_exec/program.bpf.c#L164) function, and set upper layer accordingly. It would also require changing `has_upper_layer` function to return if upper_layer is already set to true, so it won't be overridden by a call to `sched_process_exec`

## Additional details:
Since `security_bprm_check` is being called before the executable would have been executed, upper layer can be set to true even in unsuccessful executions

Sefi4:
In a second thought, it might be better to have another field, e.g. fupper_layer which would be the upper layer of the filepath and keep upper layer to be upper layer of exepath (maybe rename to eupper_layer)

PR #3869, opened 2025-01-09
Adding fupper layer field for trace_exec image-based gadget

Sefi4:
# Adding fupper layer field for trace_exec image-based gadget
fupper_laer field provides an indication for whether the executable file of the process is in the upper layer of the overlay filesystem. This can be useful information when exepath != filepath, such as in a case of a script.

In response to https://github.com/inspektor-gadget/inspektor-gadget/issues/3838

## Testing done
* Unit tests
image

### DIFF
--- a/gadgets/trace_exec/gadget.yaml
+++ b/gadgets/trace_exec/gadget.yaml
@@ -68,14 +68,17 @@ datasources:
           columns.width: 16
       upper_layer:
         annotations:
-          description: Whether the executable is in the upper layer of the overlay filesystem
+          description: Whether the executable file of the process (interpreter in case of script) is in the upper layer of the overlay filesystem. Only initialized when the execution succeeded.
           columns.width: 5
           columns.hidden: true
       pupper_layer:
         annotations:
-          description: Whether the executable's parent in the process hierarchy
-            is in the upper layer of the overlay filesystem. Only initialized when
-            the execution succeeded.
+          description: Whether the executable's parent in the process hierarchy (interpreter in case of script) is in the upper layer of the overlay filesystem. Only initialized when the execution succeeded.
+          columns.width: 5
+          columns.hidden: true
+      fupper_layer:
+        annotations:
+          description: Whether the executable file of the process (script in case of script) is in the upper layer of the overlay filesystem. Initialized when the execution succeeded or failed with the -ENOEXEC error ("Exec format error"). Notice that to get failed executions, you need to set ignore_failed to false.
           columns.width: 5
           columns.hidden: true
 params:
--- a/gadgets/trace_exec/program.bpf.c
+++ b/gadgets/trace_exec/program.bpf.c
@@ -38,6 +38,7 @@ struct event {
 	gadget_errno error_raw;
 	int args_count;
 	bool upper_layer;
+	bool fupper_layer;
 	bool pupper_layer;
 	unsigned int args_size;
 	char cwd[MAX_STRING_SIZE];
@@ -79,6 +80,13 @@ struct {
 	__type(key, pid_t);
 	__type(value, struct event);
 } execs SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 10240);
+	__type(key, pid_t);
+	__type(value, __u8);
+} security_bprm_hit_map SEC(".maps");
 
 GADGET_TRACER_MAP(events, 1024 * 256);
 
@@ -198,12 +206,12 @@ static __always_inline bool has_upper_layer(struct inode *inode)
 SEC("tracepoint/sched/sched_process_exec")
 int ig_sched_exec(struct trace_event_raw_sched_process_exec *ctx)
 {
-	u32 execs_lookup_key = ctx->old_pid;
+	u32 pre_sched_pid = ctx->old_pid;
 	struct event *event;
 	struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 	struct task_struct *parent = BPF_CORE_READ(task, real_parent);
 
-	event = bpf_map_lookup_elem(&execs, &execs_lookup_key);
+	event = bpf_map_lookup_elem(&execs, &pre_sched_pid);
 	if (!event)
 		return 0;
 
@@ -229,7 +237,8 @@ int ig_sched_exec(struct trace_event_raw_sched_process_exec *ctx)
 	if (len <= sizeof(*event))
 		gadget_output_buf(ctx, &events, event, len);
 
-	bpf_map_delete_elem(&execs, &execs_lookup_key);
+	bpf_map_delete_elem(&execs, &pre_sched_pid);
+	bpf_map_delete_elem(&security_bprm_hit_map, &pre_sched_pid);
 
 	return 0;
 }
@@ -266,6 +275,7 @@ static __always_inline int exit_execve(void *ctx, int retval)
 		gadget_output_buf(ctx, &events, event, len);
 cleanup:
 	bpf_map_delete_elem(&execs, &pid);
+	bpf_map_delete_elem(&security_bprm_hit_map, &pid);
 	return 0;
 }
 
@@ -286,14 +296,10 @@ int ig_execveat_x(struct syscall_trace_exit *ctx)
 SEC("kprobe/security_bprm_check")
 int BPF_KPROBE(security_bprm_check, struct linux_binprm *bprm)
 {
-	if (!paths)
-		return 0;
-
 	u32 pid = (u32)bpf_get_current_pid_tgid();
 	struct event *event;
 	char *file;
 	dev_t dev_no;
-	unsigned long inode_no;
 	struct path f_path;
 
 	event = bpf_map_lookup_elem(&execs, &pid);
@@ -302,18 +308,32 @@ int BPF_KPROBE(security_bprm_check, struct linux_binprm *bprm)
 
 	// security_bprm_check is called repeatedly following the shebang
 	// Only get the first call.
-	if (event->file[0] != '\0')
+	__u8 *exists = bpf_map_lookup_elem(&security_bprm_hit_map, &pid);
+	if (exists) {
 		return 0;
+	}
 
-	f_path = BPF_CORE_READ(bprm, file, f_path);
-	file = get_path_str(&f_path);
-	struct inode *inode = BPF_CORE_READ(f_path.dentry, d_inode);
-	dev_no = BPF_CORE_READ(inode, i_sb, s_dev);
-	inode_no = BPF_CORE_READ(inode, i_ino);
-	bpf_probe_read_kernel_str(event->file, MAX_STRING_SIZE, file);
-	event->dev_major = MAJOR(dev_no);
-	event->dev_minor = MINOR(dev_no);
-	event->inode = BPF_CORE_READ(inode, i_ino);
+	__u8 hit = 1;
+	if (bpf_map_update_elem(&security_bprm_hit_map, &pid, &hit,
+				BPF_NOEXIST)) {
+		return 0;
+	}
+
+	struct inode *inode = BPF_CORE_READ(bprm, file, f_inode);
+	if (inode)
+		event->fupper_layer = has_upper_layer(inode);
+
+	if (paths) {
+		f_path = BPF_CORE_READ(bprm, file, f_path);
+		file = get_path_str(&f_path);
+		bpf_probe_read_kernel_str(event->file, MAX_STRING_SIZE, file);
+
+		dev_no = BPF_CORE_READ(inode, i_sb, s_dev);
+		event->dev_major = MAJOR(dev_no);
+		event->dev_minor = MINOR(dev_no);
+		event->inode = BPF_CORE_READ(inode, i_ino);
+	}
+
 	return 0;
 }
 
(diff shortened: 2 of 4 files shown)


gadgets/trace_exec/program.bpf.c
```diff
@@ -269,6 +271,10 @@ int BPF_KPROBE(security_bprm_check, struct linux_binprm *bprm)
 	if (event->file[0] != '\0')
 		return 0;
 
+	inode = BPF_CORE_READ(bprm, file, f_inode);
+	if (inode)
+		event->fupper_layer = has_upper_layer(inode);
```
alban, line 276: Since `fupper_layer` is not a path, it is not logical to enable it only with `--paths`. The test `if (!paths) return 0;` could be moved so that `fupper_layer` would always be set regardless of `--paths`. This requires some refactoring because we wouldn't be able to rely on `event->file[0] != '\0'` to check if `security_bprm_check` is executed for the first time. I suggest to either use a separate map to count `security_bprm_check` calls, or to still use `event->file` as a marker but clean it up before `gadget_output_buf`.

gadgets/trace_exec/gadget.yaml
```diff
@@ -60,6 +60,11 @@ datasources:
             the execution succeeded.
           columns.width: 5
           columns.hidden: true
+      fpupper_layer:
+        annotations:
+          description: Whether the executable file of the process is in the upper layer of the overlay filesystem (require --paths flag).
```
alban, line 65: I suggest to remove "require --paths flag" (see other comment below).

The 3 `{,f,p}upper_layer` fields should document the following:
1. Whether it refers to the script or the interpreter
2. When it is initialized

Phrasing suggestions:
* upper_layer: Whether the executable file of the process (interpreter in case of script) is in the upper layer of the overlay filesystem.  Only initialized when the execution succeeded.
* pupper_layer: Whether the executable's parent in the process hierarchy (interpreter in case of script) is in the upper layer of the overlay filesystem. Only initialized when the execution succeeded.
* fpupper_layer: Whether the executable file of the process (script in case of script) is in the upper layer of the overlay filesystem. Initialized when the execution succeeded or failed with the -ENOEXEC error ("Exec format error").

gadgets/trace_exec/test/integration/trace_exec_test.go
```diff
@@ -41,6 +41,7 @@ type traceExecEvent struct {
 	Error       string `json:"error"`
 	UpperLayer  bool   `json:"upper_layer"`
 	PupperLayer bool   `json:"pupper_layer"`
+	FupperLayer bool   `json:"fupper_layer"`
```
alban, line 44: This unit test does not test the distinction between UpperLayer and FupperLayer when running a script. This is not easy to test the two cases because we don't have a container image with an unmodified script.

But you could run a test by creating a script in `/dev/myscript.sh` and in `/bin/myscript.sh`. Since `/dev` is a tmpfs, it will never be on an overlay upper layer.

```
printf '#!/bin/sh\ncat /dev/null' > /dev/myscript.sh ; chmod +x /dev/myscript.sh ; /dev/myscript.sh
printf '#!/bin/sh\ncat /dev/null' > /bin/myscript.sh ; chmod +x /bin/myscript.sh ; /bin/myscript.sh
```
The first command should have FupperLayer=false, and the second command should have FupperLayer=true.

alban (commented):
Thanks!

gadgets/trace_exec/gadget.yaml
```diff
@@ -60,6 +60,11 @@ datasources:
             the execution succeeded.
           columns.width: 5
           columns.hidden: true
+      fpupper_layer:
```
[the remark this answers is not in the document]
Sefi4: Oops, thanks :)

gadgets/trace_exec/program.bpf.c
Sefi4: Yes, I think it make more sense too.
I chose using event->file as a marker and cleaning it up before sending it, because it requires less overhead than map does. Though, using a map would be clearer to what's happening. I guess it's more of a philosophical question, but should I favor explicitness (using a map) or efficiency, what do you think?

alban: I don't have a strong opinion, as long as the code is clear (with a comment if necessary).

gadgets/trace_exec/program.bpf.c
```diff
@@ -269,8 +269,12 @@ int BPF_KPROBE(security_bprm_check, struct linux_binprm *bprm)
...
 
-	f_path = BPF_CORE_READ(bprm, file, f_path);
-	file = get_path_str(&f_path);
+	struct inode *inode = BPF_CORE_READ(bprm, file, f_inode);
+	if (inode)
+		event->fupper_layer = has_upper_layer(inode);
+
+	struct path f_path = BPF_CORE_READ(bprm, file, f_path);
+	char *file = get_path_str(&f_path);
 	bpf_probe_read_kernel_str(event->file, MAX_STRING_SIZE, file);
```
alban, line 278: Part of the reason to have `--paths` configurable is to avoid the performance penalty of `get_path_str` if not requested by the user. But this changes defeat this.

You could try something like this:
```c
if (paths) {
	struct path f_path = BPF_CORE_READ(bprm, file, f_path);
	char *file = get_path_str(&f_path);
	bpf_probe_read_kernel_str(event->file, MAX_STRING_SIZE, file);
} else {
	// Add a mark in the event so if security_bprm_check is executed
	// several times for the same event, we will know about it and
	// event->fupper_layer will not be overwritten.
	// The mark will be cleaned just before gadget_output_buf if necessary.
	event->file[0] = 'X';
}
```

But now I wonder if `get_path_str` can return NULL or an empty string. That would prevent the mark from working as intended.

Sefi4: Ah I see, then let's use map instead

gadgets/trace_exec/test/integration/trace_exec_test.go
```diff
@@ -117,13 +118,14 @@ int main(int argc, char *argv[], char **envp) {
 	buildCmd := fmt.Sprintf("echo %s | base64 -d > exec.c && gcc -Wall -o /bin/exec-syscall exec.c", progBase64)
 	sleep1Args := []string{"/usr/bin/sleep", "0.4"}
 	sleep2Args := []string{"/usr/bin/sleep", "0.6"}
+	scriptsCmd := "echo '#!/bin/sh\nls' > /bin/with_shabeng.sh ; /dev/script.sh echo ls > /dev/script.sh ; echo ls > /bin/script.sh ; chmod +x /dev/script.sh /bin/script.sh /bin/with_shabeng.sh ; /dev/script.sh ; /bin/script.sh ; /bin/with_shabeng.sh;"
```
alban, line 121: - s/shabeng/shebeng/g
- spurious `/dev/script.sh` . Or did I miss something?

```suggestion
	scriptsCmd := "echo '#!/bin/sh\nls' > /bin/with_shebeng.sh ; echo ls > /dev/script.sh ; echo ls > /bin/script.sh ; chmod +x /dev/script.sh /bin/script.sh /bin/with_shebeng.sh ; /dev/script.sh ; /bin/script.sh ; /bin/with_shebeng.sh;"
```

gadgets/trace_exec/test/integration/trace_exec_test.go
```diff
@@ -117,13 +118,14 @@ int main(int argc, char *argv[], char **envp) {
...
 	innerCmd := fmt.Sprintf(
 		"cd tmp ; while true ; do METHOD=execve exec-syscall %s; METHOD=execveat exec-syscall %s; done",
 		strings.Join(sleep1Args, " "),
 		strings.Join(sleep2Args, " "),
 	)
 	// copies /usr/bin/sh to /usr/bin/sh2 to check that the upper_layer is true when executing /usr/bin/sh2
-	cmd := fmt.Sprintf("%s && cp /usr/bin/sh /usr/bin/sh2 && nsenter --setuid 1000 --setgid 1111 /usr/bin/sh2 -c '%s'", buildCmd, innerCmd)
+	cmd := fmt.Sprintf("%s %s && cp /usr/bin/sh /usr/bin/sh2 && nsenter --setuid 1000 --setgid 1111 /usr/bin/sh2 -c '%s'", scriptsCmd, buildCmd, innerCmd)
```
alban, line 128: In your current code, `scriptsCmd` is executed only once at the beginning. It is better to split it into two parts:
1. prepareScriptsCmd: executed only once, prepanded to the `cmd` variable
2. execScriptsCmd: executed in the "while" loop, inserted in the `innerCmd`

This is because the test do the following:
1. starts the ig process in the background
2. start the container running the script

If ig is slow to start, it might miss the first events, but since the execs are in a "while" loop, the events will be observed at some point after it gets initialized.

gadgets/trace_exec/test/integration/trace_exec_test.go
```diff
@@ -149,22 +151,85 @@ int main(int argc, char *argv[], char **envp) {
...
+					// inner script.sh
+					{
+						CommonData:  utils.BuildCommonData(containerName, commonDataOpts...),
+						Proc:        utils.BuildProc("sh", 0, 0),
+						Cwd:         "/",
+						Args:        "/dev/script.sh",
+						PupperLayer: false,
+						UpperLayer:  false,
+						FupperLayer: false,
+						Exepath:     "/usr/bin/dash",
```
alban, line 167: Why dash?

gadgets/trace_exec/test/integration/trace_exec_test.go
```diff
@@ -149,22 +151,85 @@ int main(int argc, char *argv[], char **envp) {
...
+						Inode:       utils.NormalizedInt,
+
+						// Check the existence of the following fields
+						Timestamp: utils.NormalizedStr,
+						Loginuid:  utils.NormalizedInt,
+						Sessionid: utils.NormalizedInt,
+					},
+					{
+						CommonData:  utils.BuildCommonData(containerName, commonDataOpts...),
+						Proc:        utils.BuildProc("with_shabeng.sh", 0, 0),
```
alban, line 201:
```suggestion
						Proc:        utils.BuildProc("with_shebeng.sh", 0, 0),
```

gadgets/trace_exec/test/integration/trace_exec_test.go
Sefi4: Ok, got it

gadgets/trace_exec/test/integration/trace_exec_test.go
Sefi4: Discussed offline, the /bin/sh Shebeng is expanded to dash

Sefi4:
> should be removed

Do you prefer to have the limitation removed completely or update it to refer to the fupper_layer? Because if we want to be super precise here, it's still true that upper_layer would refer to the interpreter location
image

And if we already touching the limitations part, do you think it worth to note that `fupper_layer` wouldn't always refer to the script?
e.g. If we run `python3 myscript.py`. Both `file` and `fupper_layer` would refer to python and not to myscript.py

alban:
The text is still true after this PR, but it is no longer a limitation, so it should not be in the "Limitations" section.

If we need to document it further, you can do that in a separate section. But I don't think it is necessary to document it further because it is already explained in gadget.yaml, so I think it is easier to just delete that text.

gadgets/trace_exec/test/integration/trace_exec_test.go
```diff
@@ -115,15 +116,18 @@ int main(int argc, char *argv[], char **envp) {
 	}
 
 	buildCmd := fmt.Sprintf("echo %s | base64 -d > exec.c && gcc -Wall -o /bin/exec-syscall exec.c", progBase64)
+	prepareScriptsCmd := "echo '#!/bin/sh\nls' > /bin/with_shebeng.sh ; echo ls > /dev/script.sh ; echo ls > /bin/script.sh ; chmod +x /dev/script.sh /bin/script.sh /bin/with_shebeng.sh ; "
```
alban, line 119: The echo builtin command in the dash shell interpret "\n" but the echo builtin command from bash or the GNU echo program does not, unless option "-e" is used. But then, dash's echo does not support "-e".

I think it is better to use "printf" instead of echo because it has a consistent behaviour for "\n" in all 3 versions.

```
$ dash -c 'echo "1\n2"'
1
2
$ dash -c '/bin/echo "1\n2"'
1\n2
$ bash -c 'echo "1\n2"'
1\n2
$ bash -c 'echo -e "1\n2"'
1
2
$ dash -c 'echo -e "1\n2"'
-e 1
2
$ dash -c 'printf "1\n2"'
1
2
```

Sefi4: Ah, I wasn't aware of this inconsistency. That's good to know. Changed it for the shebeng script

alban (approved):
LGTM!

I can merge it if the CI passes.

Sefi4:
Documentation checks seems to fail, not sure what it means, should I run make generate-documentation?

Sefi4:
Ah I see now:
image

This link in building.md is broken: https://reproducible-builds.org/docs/source-date-epoch/)/docs/source-date-epoch/

alban:
The website (https://reproducible-builds.org) seems to have been fixed. The CI is fine now.

Issue #3854, opened 2025-01-06
[RFE] fsnotify gadget: add support for Linux < 5.16


alban:
## Current situation

This issue was initially reported on:
https://github.com/alban/fsnotify-enricher/issues/4

Linux 5.16 renamed `fsnotify_add_event` to `fsnotify_insert_event`.
https://github.com/torvalds/linux/commit/1ad03c3a326a86e259389592117252c851873395#diff-0b91dd4559c511ec5e12f88b5152d07124a9e6dff3e518ebd799b55b50953ba0

On < 5.16, this code won't work:
```c
SEC("kprobe/fsnotify_insert_event")
int BPF_KPROBE(fsnotify_insert_event_e, struct fsnotify_group *group,
	       struct fsnotify_event *event)
```

## Impact

The fsnotify gadget does not work on Linux < 5.16:

> running gadget: starting operators: starting operator "oci": starting operator "ebpf": attaching eBPF program "fsnotify_insert_event_e": creating perf_kprobe PMU (arch-specific fallback for "fsnotify_insert_event"): token __x64_fsnotify_insert_event: not found: no such file or directory

## Ideal future situation

The fsnotify gadget works both on Linux < 5.16 and Linux >= 5.16.

## Implementation options

Use a wasm module to test the availability of `fsnotify_add_event` and `fsnotify_insert_event` in kallsyms, and attach the kprobe to the right function.

## Additional information

The issue has been observed on Linux 5.10, 5.14, 5.15.

PR #3871, opened 2025-01-09
wasm: add kallsymsSymbolExists

alban:
# wasm: add kallsymsSymbolExists

This allows to fix the fsnotify gadget for Linux < 5.16.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/3854

This is part of https://github.com/inspektor-gadget/inspektor-gadget/issues/3131

## How to use

No changes

## Testing done

None. I'd like to check the CI for the fsnotify tests under different kernel versions.


pkg/operators/wasm/kallsyms.go
```diff
@@ -0,0 +1,66 @@
...
+	symbolName, err := stringFromStack(m, symbolNamePtr)
+	if err != nil {
+		i.logger.Warnf("kallsymsSymbolExists: reading string from stack: %v", err)
+		stack[0] = 0
+		return
+	}
+
+	if i.kernelSymbolResolver == nil {
+		var err error
+		i.kernelSymbolResolver, err = kallsyms.NewKAllSyms()
```
alban, line 52: I am not sure where to release `i.kernelSymbolResolver` to free memory.

I'll try to add it just after `i.callGuestFunction(gadgetCtx.Context(), "gadgetPreStart")`.

I don't want to negate the efforts from #3616 and #3595.

alban: I think the PR in the current version is not good enough:

### CPU performance

Reading `/proc/kallsyms` is slow. In my tests, it takes 0.19s for the kernel to generate the file dynamically. When a file of the same size is read from a normal filesystem, it takes less than 0.01s. In practice, the symbols we want are from the builtin kernel and not modules, so towards the beginning of the file. So we waste time parsing the full file. When stopping at the symbol we want (`fsnotify_insert_event`), it takes 0.01s.

```bash
$ sudo time -f %S cat /proc/kallsyms > /dev/null
0.19

$ sudo time -f %S grep -m 1 -w fsnotify_insert_event /proc/kallsyms > /dev/null
0.01
```

### Memory performance

Previous work (https://github.com/inspektor-gadget/inspektor-gadget/pull/3571#issuecomment-2410489279) has shown that parsing `/proc/kallsyms` with `strings.Fields` is memory intensive:
> pprof shows the kallsyms takes 8MB of allocations in use, but 38 MB allocations in total (some of which will be released by the GC after initialisation).

cilium/ebpf fixed the same problem with a custom reader: https://github.com/cilium/ebpf/pull/1588

### Solution

https://github.com/cilium/ebpf/pull/1588 is implemented in an internal package, so it cannot be imported.

The gadget should be able to tell ig which symbols it cares about and only keep those in memory.

This has an impact on the API: unfortunately, it cannot be a simple function anymore
(`func KallsymsSymbolExists(symbol string) bool`).

mauriciovasquezbernal: > The gadget should be able to tell ig which symbols it cares about and only keep those in memory.

> This has an impact on the API: unfortunately, it cannot be a simple function anymore
(func KallsymsSymbolExists(symbol string) bool).

I'm not sure I follow this limitation. Could you please extend a bit more?

alban: > I'm not sure I follow this limitation. Could you please extend a bit more?

I changed my mind: I think it is fine to implement it as a single simple function.

I'd like to improve the pkg/kallsyms package first via #3932. Then I will update this PR to make use of it.

Marking as draft for now.

alban: #3932 is merged. Rebased.

pkg/operators/wasm/testdata/kallsyms/go.mod
```diff
@@ -0,0 +1,8 @@
+module main
+
+go 1.23.0
+
+require github.com/inspektor-gadget/inspektor-gadget v0.27.0
```
mauriciovasquezbernal, line 5: v0.27.0 > v0.0.0

pkg/operators/wasm/testdata/kallsyms/program.bpf.c
mauriciovasquezbernal, line 1: This is not true anymore after #3819

pkg/operators/wasm/testdata/kallsyms/program.go
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This program tries as hard as it can to break the host by calling functions
+// with wrong arguments. It uses the low level functions directly as the goal is
+// to test the host and not the wrapper API. Tests under dataarray and fields
+// test also the higher level API.
```
mauriciovasquezbernal, line 18: remove?

wasmapi/go/kallsyms.go
```diff
@@ -0,0 +1,31 @@
...
+//go:wasmimport env kallsymsSymbolExists
+func kallsymsSymbolExists(sumbol uint64) uint32
+
+func KallsymsSymbolExists(symbol string) bool {
+	ret := kallsymsSymbolExists(uint64(stringToBufPtr(symbol)))
+	runtime.KeepAlive(symbol)
+	if ret != 0 {
+		return true
+	}
+	return false
```
mauriciovasquezbernal, line 30: `return ret != 0` ?

wasmapi/go/kallsyms.go
mauriciovasquezbernal, line 1: please add the corresponding testing to pkg/operators/wasm/testdata/badguest/program.go (i.e. calling kallsymsSymbolExists with a wrong string)

gadgets/fsnotify/go/program.go
```diff
@@ -0,0 +1,39 @@
...
+		fallbackExists := api.KallsymsSymbolExists("fsnotify_add_event")
+		if !fallbackExists {
+			api.Errorf("kernel symbol not found: fsnotify_add_event or fsnotify_insert_event")
+			return 1
+		}
+		api.SetConfig("programs.fsnotify_insert_event_e.attach_to", "fsnotify_add_event")
+		return 0
+	}
+
+	return 0
```
mauriciovasquezbernal, line 36: nit: perhaps an early return makes it a bit more readable:

```golang
	// Linux 5.16 renamed fsnotify_add_event to fsnotify_insert_event
	// https://github.com/torvalds/linux/commit/1ad03c3a326a86e259389592117252c851873395
	if api.KallsymsSymbolExists("fsnotify_insert_event") {
		return 0
	}

	if api.KallsymsSymbolExists("fsnotify_add_event") {
		api.SetConfig("programs.fsnotify_insert_event_e.attach_to", "fsnotify_add_event")
		return 0
	}

	api.Errorf("kernel symbol not found: fsnotify_add_event or fsnotify_insert_event")
	return 0
```

wasmapi/go/kallsyms.go
alban: Updated.

mauriciovasquezbernal (approved):
LGTM! (edit: after the conflicting file has been fixed :( )

Issue #3513, opened 2024-09-23
testing: Consider avoid storing bin artifacts on the repository


mauriciovasquezbernal:
We store binary artifacts needed for unit tests in different paths on the repository:

- https://github.com/inspektor-gadget/inspektor-gadget/tree/main/testdata
- https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/operators/ebpf
- https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/operators/wasm/testdata

The purpose of this issue is to discuss if we want to keep storing them or if we want to use a different approach 

See https://github.com/inspektor-gadget/inspektor-gadget/pull/3475#discussion_r1771502833

PR #3881, opened 2025-01-13
operators/wasm: Check map type is supported in operation functions. 

eiffel-fl:
We want to allow creating all types of maps, as thy can be used in eBPF, but
we want to restrain operations on maps manipulating FD.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/commit/2e3e77a573d57cdd7cf6881e3b18ab0218b22abb ("operators/wasm: Add support for map of maps.")

### DIFF
--- a/pkg/operators/wasm/maps.go
+++ b/pkg/operators/wasm/maps.go
@@ -91,7 +91,7 @@ func isMapOfMaps(typ ebpf.MapType) bool {
 	return typ == ebpf.ArrayOfMaps || typ == ebpf.HashOfMaps
 }
 
-func isMapTypeSupported(typ ebpf.MapType) error {
+func isMapTypeSupportedForAccess(typ ebpf.MapType) error {
 	// We do not permit operation on map unknown to us.
 	// A newer kernel can introduce a new type of map taking an FD as a value,
 	// this way user could use this new map to leak FD in guest WASM.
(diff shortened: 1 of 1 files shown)


alban (changes requested):
PerfEventArray takes a file descriptor as value for `mapUpdate`.

[NewReaderWithOptions](https://github.com/cilium/ebpf/blob/v0.17.1/perf/reader.go#L188) is still adding the fd in the map using [Map.Put](https://github.com/cilium/ebpf/blob/v0.17.1/perf/reader.go#L444):
```
if err := pr.array.Put(uint32(i), fd.Uint()); err != nil {
	return fmt.Errorf("couldn't put event fd %d for CPU %d: %w", fd, i, err)
}
```

So this PR would allow a wasm module to call `mapUpdate` on an integer, and the kernel would interpret it as the fd number.

alban (commented):
What about renaming the function `isMapTypeSupported` to `isMapTypeSupportedForAccess`? Some maps are supported for `newMap` and for `getMap` but not for `mapLookup` and `mapUpdate`. So the naming would make it clearer.

alban (approved):
The file pkg/operators/wasm/testdata/mapofmap.tar should not be part of that PR.

Otherwise, LGTM!

mauriciovasquezbernal:
> None of these .Tar files should be part of the repo, I do not see why we persist in having binaries in the repo.
> We should just build them at runtime before the tests.

The main issue about removing these files is that users will have to build them before running unit tests. Anyway, happy to discuss it on #3513

PR #3122, opened 2024-07-03
Update google.golang.org/grpc to v1.65.0

mauriciovasquezbernal:
Ignoring the linter until we migrate to NewClient().

See https://github.com/grpc/grpc-go/issues/7090#issuecomment-2043623605

Supersedes #2849

PR #3874, opened 2025-01-10
Add additional dns information

AdvH039:
# [Add additional DNS information to trace_dns gadget]

Refer #2610 

## How to use

Test with the tools/dnstester container and display the respective columns  ttl_values,is_compressed, TC, RA, RD

## Testing done

Screenshot from 2025-01-10 18-30-06

PR #3888, opened 2025-01-14
Use grpc NewClient instead of deprecated Dial

AdvH039:
# Replace NewClient with Dial
Can be done after merging of  #3122 

## Testing done
Passes linter and integration tests. Keeping it as draft because I am not sure if more testing or discussion is required.

### DIFF
--- a/pkg/runtime/grpc/grpc-runtime.go
+++ b/pkg/runtime/grpc/grpc-runtime.go
@@ -381,7 +381,7 @@ func (r *Runtime) getConnToRandomTarget(ctx context.Context, runtimeParams *para
 	log.Debugf("using target %q (%q)", target.addressOrPod, target.node)
 
 	timeout := time.Second * time.Duration(r.globalParams.Get(ParamConnectionTimeout).AsUint16())
-	conn, err := r.dialContext(ctx, target, timeout)
+	conn, err := r.dialContext(target, timeout)
 	if err != nil {
 		return nil, fmt.Errorf("dialing %q (%q): %w", target.addressOrPod, target.node, err)
 	}
@@ -392,7 +392,7 @@ func (r *Runtime) getConnFromTarget(ctx context.Context, runtimeParams *params.P
 	log.Debugf("using target %q (%q)", target.addressOrPod, target.node)
 
 	timeout := time.Second * time.Duration(r.globalParams.Get(ParamConnectionTimeout).AsUint16())
-	conn, err := r.dialContext(ctx, target, timeout)
+	conn, err := r.dialContext(target, timeout)
 	if err != nil {
 		return nil, fmt.Errorf("dialing %q (%q): %w", target.addressOrPod, target.node, err)
 	}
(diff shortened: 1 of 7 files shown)


flyth:
IIRC it wasn't that easy to update to the non-deprecated versions, but I'd need to check again. Did you run tests on the changes? Does it still work correctly for `ig --daemon` with `gadgetctl` and the Kubernetes deployment (with `kubectl-gadget`)?

AdvH039:
I am receiving these errors when running in headless mode 
```
RUNTIME.CONTAINERNAME                                                             PID        COMM             PROTO ADDR                                        PORT  OPTS  IF                              
Error: running gadget: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /var/run/ig/ig.socket: connect: permission denied"
```
It is stil not ready as you said. 
Could you also review my other PR #3874? I would appreciate it a lot. Thanks!!

flyth:
> I am receiving these errors when running in headless mode
> 
> ```
> RUNTIME.CONTAINERNAME                                                             PID        COMM             PROTO ADDR                                        PORT  OPTS  IF                              
> Error: running gadget: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /var/run/ig/ig.socket: connect: permission denied"
> ```

That actually looks like a misconfiguration rather than the gRPC stack not working; the socket file needs to be accessible from the client (`gadgetctl`) by either starting `ig --daemon` with a proper `--group` or using `gadgetctl` from root as well.

AdvH039:
Sorry I missed that. Thanks!! I tried `gadgetctl `(in root mode) and `kubectl-gadget` on minikube with a few gadgets and it seems to be working fine. I can only find the integration tests for ig in the docs and tested so. Do I have to adjust the environment variables used for different clients and runtimes?

pkg/runtime/grpc/grpc-runtime.go
```diff
@@ -516,14 +511,9 @@ All these options should be set at the same time to enable TLS connection`,
 			gadgetNamespace := r.globalParams.Get(ParamGadgetNamespace).AsString()
 			return NewK8SPortFwdConn(ctx, r.restConfig, gadgetNamespace, target, port, timeout)
 		}))
-	} else {
-		newCtx, cancel := context.WithTimeout(dialCtx, timeout)
```
burak-ok, line 520: It seems like for the `else` case we are losing the timeout.
We would need to keep that even with the new `NextClient`

burak-ok (commented):
We would also need to discuss if we need to establish the connection immediatly.

According to https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#the-wrong-way-grpcdial `grpc.Dial` and `grpc.DialContext` established the connection immediatly, whereas `grpc.NewClient` doesn't do that

mauriciovasquezbernal:
@AdvH039 are you planning to update this one?

AdvH039:
Yes give me this week please. Thank you!!

AdvH039:
image

Running into this error while running gadgetctl

PR #3889, opened 2025-01-15
wasmapi: Bump available map to reflect cilium/ebpf 0.17.1.

eiffel-fl:
New map types were added in the latest cilium/ebpf release [1]. Let's add them to our WASM API.

[1]: https://github.com/cilium/ebpf/commit/d340f32a007b

### DIFF
--- a/wasmapi/go/map.go
+++ b/wasmapi/go/map.go
@@ -56,7 +56,7 @@ const (
 type MapType uint32
 
 // Taken from:
-// https://github.com/cilium/ebpf/blob/061e86d8f5e9/types.go#L15-L98
+// https://github.com/cilium/ebpf/blob/6d6c5e3225732525434b0c192c97950488a4051b/types.go#L13-L105
 const (
 	UnspecifiedMap MapType = iota
 	Hash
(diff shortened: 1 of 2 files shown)


wasmapi/go/map.go
```diff
@@ -88,6 +88,10 @@ const (
 	RingBuf
 	InodeStorage
 	TaskStorage
+	BloomFilter
+	UserRingbuf
+	CgroupStorage
+	Arena
 )
```
alban, line 95: LGTM

- [BloomFilter](https://github.com/torvalds/linux/commit/9330986c03006ab1d33d243b7cfe598a7a3c1baa) takes a NULL pointer as key, so it won't work but the kernel will just return an error and it is safe.
- [UserRingbuf](https://github.com/torvalds/linux/commit/583c1f420173f7d84413a1a1fbf5109d798b4faa) requires separate bpf helpers. The current lookup/update/delete helpers will safely return an error.
- [CgroupStorage](https://github.com/torvalds/linux/commit/de9cbbaadba5adf88a19e46df61f7054000838f6) takes the cgroup inode id as key but not its fd, so the marshalling will work fine.
- [Arena](https://github.com/torvalds/linux/commit/317460317a02a1af512697e6e964298dedd8a163)'s current lookup/update/delete helpers will safely return an error (-EOPNOTSUPP or -EINVAL).

alban (approved):
LGTM

PR #3899, opened 2025-01-16
docs: multiple filters

mqasimsarfraz:
Document how to specify multiple filters in the filter operator.

Fixes: https://github.com/inspektor-gadget/inspektor-gadget/issues/3453


pkg/operators/filter/filter.go
```diff
@@ -73,6 +73,7 @@ func (f *filterOperator) InstanceParams() api.Params {
     field<value      - matches, if the content of field is less than the value
     field~value      - matches, if the content of field matches the regular expression 'value'
                  see [https://github.com/google/re2/wiki/Syntax] for more information on the syntax
+  Multiple filters can be combined using a comma: field1==value1,field2==value2
```
alban, line 76: Does [cmd/common/registry.go#L288-L296](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/registry.go#L288-L296) need similar documentation too?

docs/spec/operators/filter.md
```diff
@@ -31,5 +31,8 @@ The filter syntax supports the following operations:
 - `field<value`: Matches if the content of `field` is less than `value`.
 - `field~value`: Matches if the content of `field` matches the regular expression `value`. See [RE2 Syntax](https://github.com/google/re2/wiki/Syntax) for more details.
 ```
+### `multiple filters`
+
+You can specify multiple filters by separating them with a comma. The filter `field1==value1,field2==value2` will match only events where `field1` equals `value1` and `field2` equals `value2`.
```
alban, line 36: This needs to tell how to escape commas if the "value" string contains a comma.

pkg/operators/filter/filter.go
mqasimsarfraz: > Does [cmd/common/registry.go#L288-L296](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/cmd/common/registry.go#L288-L296) need similar documentation too?

That is related to builtin gagets but good idea to include the comment. 

> And docs/reference/run.mdx too?

It seems it is already documented here: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/reference/run.mdx#examples

mauriciovasquezbernal (commented):
LGTM besides Alban's comments. Do you think we can also this opportunity to fix the bad rendering on the website:

image

I think we need to remove the code spinet that covers all of this.

mqasimsarfraz:
Updated the format: 

image

docs/spec/operators/filter.md
mqasimsarfraz: Great point. Looking at this: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/filter/filter.go#L88 I assume we don't have a way to escape comma in the value. Let me open a separate PR for it!

mauriciovasquezbernal (approved):
LGTM. Thanks!

mqasimsarfraz:
Thanks for the reviews!

docs/spec/operators/filter.md
mqasimsarfraz: I opened the PR: https://github.com/inspektor-gadget/inspektor-gadget/pull/3907

PR #3907, opened 2025-01-17
filter: Allow escaping commas

mqasimsarfraz:
As discussed in the review: https://github.com/inspektor-gadget/inspektor-gadget/pull/3899#discussion_r1918914740 we need a way to escape commas in the filter values. This PR adds support for escaping commas in the filter values by using a backslash (`\`).

### Testing Done

```bash
# cat /tmp/a,b
$ go run -exec sudo ./cmd/ig run trace_open --host -F 'fname==/tmp/a\,b'
RUNTIME.CONTAINERNAME                                                                COMM                    PID        TID  FD FNAME                                                                                    MODE                                         ERROR                            
                                                                                     cat                   13519      13519   3 /tmp/a,b                                                                                 ----------                                                         
```

Also, `filter_test.go` has been updated to handle the new case!

### DIFF
--- a/pkg/operators/filter/filter.go
+++ b/pkg/operators/filter/filter.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
-	"strings"
 
 	"golang.org/x/exp/constraints"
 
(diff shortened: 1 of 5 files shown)


pkg/operators/filter/filter.go
```diff
@@ -101,6 +100,32 @@ func (f *filterOperator) InstantiateDataOperator(gadgetCtx operators.GadgetConte
 	return fop, nil
 }
 
+func splitFilter(filterCfg string) []string {
```
flyth, line 103: I think this func should be moved to the api package as it can potentially be reused a lot.

mqasimsarfraz: Done. Let me know how it looks now!

mauriciovasquezbernal (approved):
I tested it and works as expected!

alban:
You could add a test for escaping backslash (`backslash=\\,slash=/,comma=\,`)

mqasimsarfraz:
> You could add a test for escaping backslash (`backslash=\\,slash=/,comma=\,`)

done

Issue #3121, opened 2024-07-03
gadgets: Provide mechanism to show the name of a network interface based on its index


mauriciovasquezbernal:
The trace bind gadget should provide the name of the interface the socket was bound to:

https://github.com/inspektor-gadget/inspektor-gadget/blob/5bf9b54f0237e048a986fa11ad5c73cbadaf08cf/gadgets/trace_bind/program.bpf.c#L46-L47

The built-in version uses the following logic to get the name of the device:

https://github.com/inspektor-gadget/inspektor-gadget/blob/13cb07051c4617853983f2b1d3dc9c8296b8b626/pkg/gadgets/trace/bind/tracer/tracer.go#L235-L250

By the way, the current logic is wrong as it doesn't take into consideration the network namespace, hence the conversion always tries to look the interface from the host POV.

This logic is missing from the image based gadget. There are few options to implement this:

1.  Use wasm: Unfortunately it doesn't seem a viable option as wasm can't access netlink, and if it could we probably won't expose it to gadgets because of safety reasons. 
2. Provide a formatter: Define a `gadget_net_index` type and logic to provide an interface name from it 
3. Get the interface name directly in eBPF: Perhaps this is an option and we won't need any extra support

schnrj:
@mauriciovasquezbernal can you please assign this issue to me?

mauriciovasquezbernal:
For sure, but before I'd like to discuss the plan. How are you planning to implement this issue?

PR #3913, opened 2025-01-20
Find interface name wrt to namespace

AdvH039:
Resolving interface number to 

Refer #3121

## How to use

I only added namespace switching to ensure correct interface name while resolving from an interface index in the in-built gadget. Normally it works fine but when I test it on a custom namespace using netns, the interface index cannot be resolved.

Commands I used (Where it's failing) :

sudo ip link add veth0 type veth peer name veth1
sudo ip link set veth0 netns custom_ns
sudo ip netns exec custom_ns ip addr show veth0
sudo ip netns exec custom_ns ip addr add 192.168.1.1/24 dev veth0
sudo ip netns exec custom_ns ip link set veth0 up
sudo ip netns exec custom_ns python3 -c "import socket
import os
def bind_to_custom_interface(interface, port):
    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, interface.encode())
        sock.bind(('192.168.1.1', port))
        print(f'Socket bound to interface {interface} on port {port}')
    except Exception as e:
        print(f'Error binding to interface: {e}')
    return sock
if __name__ == '__main__':
    interface = 'veth0'
    port = 12345
    if os.geteuid() != 0:
        print('This script requires root privileges. Run it as root.')
    else:
        bind_to_custom_interface(interface, port)
"
(I traced this using the --host flag)

I tried exploring the implementation of the logic in the image based gadget but got stuck:

I tried acessing the inode number of the namespace that was present in sock structure but that gives the inode number in the kenel space which is different from inode number in user space and not very usable.
I do not see any direct ebpf functions that help find interfaces and resolve them in a particular namespace  so I am confused on how to proceed with option 4 .
Finally the only  data I could obtain from the trace and was usable in the user space for namespace switching was the pid but requires netlink for resolving index.
I have not explored option 2 and 3 and was wondering how would we implement them without netlink.
( I have kept the following PR as a draft to further explore the issue. Thanks !!)

### DIFF
--- a/pkg/gadgets/trace/bind/tracer/tracer.go
+++ b/pkg/gadgets/trace/bind/tracer/tracer.go
@@ -31,6 +31,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/types"
 	eventtypes "github.com/inspektor-gadget/inspektor-gadget/pkg/types"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/nsenter"
 )
 
 //go:generate go run github.com/cilium/ebpf/cmd/bpf2go -target $TARGET -cc clang -cflags ${CFLAGS} -type bind_event bindsnoop ./bpf/bindsnoop.bpf.c -- -I./bpf/
@@ -236,23 +237,6 @@ func (t *Tracer) run() {
 
 		bpfEvent := (*bindsnoopBindEvent)(unsafe.Pointer(&record.RawSample[0]))
 
-		interfaceString := ""
-		interfaceNum := int(bpfEvent.BoundDevIf)
-		if interfaceNum != 0 {
-			// It does exist a net link which index is 0.
-			// But eBPF bindsnoop code often gives 0 as interface number:
-			// https://github.com/iovisor/bcc/blob/63618552f81a2631990eff59fd7460802c58c30b/tools/bindsnoop_example.txt#L16
-			// So, we only use this function if interface number is different than 0.
-			interf, err := netlink.LinkByIndex(interfaceNum)
-			if err != nil {
-				msg := fmt.Sprintf("Cannot get net interface for %d : %s", interfaceNum, err)
-				t.eventCallback(types.Base(eventtypes.Err(msg)))
-				return
-			}
-
-			interfaceString = interf.Attrs().Name
-		}
-
 		addr := gadgets.IPStringFromBytes(bpfEvent.Addr, int(bpfEvent.Ver))
 
 		event := types.Event{
@@ -265,7 +249,6 @@ func (t *Tracer) run() {
 			Addr:          addr,
 			Port:          bpfEvent.Port,
 			Options:       optionsToString(bpfEvent.Opts),
-			Interface:     interfaceString,
 			Comm:          gadgets.FromCString(bpfEvent.Task[:]),
 			WithMountNsID: eventtypes.WithMountNsID{MountNsID: bpfEvent.MountNsId},
 			Uid:           bpfEvent.Uid,
@@ -275,6 +258,44 @@ func (t *Tracer) run() {
 		if t.enricher != nil {
 			t.enricher.EnrichByMntNs(&event.CommonData, event.MountNsID)
 		}
+
+		interfaceString := ""
+
+		interfaceNum := int(bpfEvent.BoundDevIf)
+		if interfaceNum != 0 {
+			// It does exist a net link which index is 0.
+			// But eBPF bindsnoop code often gives 0 as interface number:
+			// https://github.com/iovisor/bcc/blob/63618552f81a2631990eff59fd7460802c58c30b/tools/bindsnoop_example.txt#L16
+			// So, we only use this function if interface number is different than 0.
+
+			var interf netlink.Link
+			pid := int(event.Pid)
+
+			// Current limitations:
+			// - Short-lived processed might have terminated already, so we cannot enter
+			//   its netns.
+			// - 'pid' is from the top-level pid namespace, which might differ from IG's
+			//   /host/proc/ on some setups such as Minikube with the docker driver.
+
+			err := nsenter.NetnsEnter(pid, func() error {
+				interf, err = netlink.LinkByIndex(interfaceNum)
+				if err != nil {
+					return err
+				}
+				return nil
+			})
+			if err != nil {
+				msg := fmt.Sprintf("Cannot find interface name for index : %d (PID : %d) : %s", interfaceNum, pid, err)
+				t.eventCallback(types.Base(eventtypes.Err(msg)))
+				continue
+			}
+
+			if interf != nil {
+				interfaceString = interf.Attrs().Name
+			}
+
+		}
+		event.Interface = interfaceString
 
 		t.eventCallback(&event)
 	}


alban:
> I do not see any direct ebpf functions that help find interfaces and resolve them in a particular namespace so I am confused on how to proceed with option 4 .

There is no direct bpf helper but it might be possible to find it manually. Inspired from [__dev_get_by_index](https://github.com/torvalds/linux/blob/v6.12/net/core/dev.c#L847) (not tested):
```c
#define NETDEV_HASHBITS    8
#define NETDEV_HASHENTRIES (1 << NETDEV_HASHBITS)
#define IFNAMSIZ        16

struct event {
  ...
  char name[IFNAMSIZ];
}

struct sock *sock = BPF_CORE_READ(socket, sk);
struct net *net = BPF_CORE_READ(sk, __sk_common.skc_net.net);

struct hlist_head *head = &net->dev_index_head[ifindex & (NETDEV_HASHENTRIES - 1)];

// Get dev from the linked list
// When there is a small number of interfaces in the network
// namespace (less than NETDEV_HASHENTRIES), it should just
// be the first device in the linked list.
struct net_device *dev = ...;
if (dev != NULL && dev->ifindex == ifindex) {
  bpf_probe_read_kernel(event.name, IFNAMSIZ, dev->name);
}
```

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -235,14 +236,36 @@ func (t *Tracer) run() {
...
 			// So, we only use this function if interface number is different than 0.
+			originalNs, err := netns.Get()
+			if err != nil {
+				msg := fmt.Sprintf("Failed to get original namespace: %s", err)
+				t.eventCallback(types.Base(eventtypes.Err(msg)))
+				return
+			}
+			defer originalNs.Close()
+
+			handle, err := netns.GetFromPid(int(bpfEvent.Pid))
```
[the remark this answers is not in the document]
AdvH039: @alban I have implemented the logic that you have mentioned in the above comment. I tried testing by binding a socket to a non zero indexed interface (`eth0` or `docker0`) within the container using a python script. The built-in gadget was only able to trace what I was doing when I used the` --host` flag (without the flag it doesn't trace this.). It was working fine when I was running the container on the host network but when I ran the container on a non-host network it gave the following error:
```
ERRO[0031] Cannot find interface for index %!!(MISSING)d(<nil>) : Link not found 
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1fc3ac8]

goroutine 67 [running]:
github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/tracer.(*Tracer).run(0xc000696d00)
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/tracer/tracer.go:291 +0x868
created by github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/tracer.(*Tracer).Run in goroutine 1
	/go/src/github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/bind/tracer/tracer.go:315 +0x325
```

> ContainerPID is the pid1 of the container, and should be in the same network namespace as the target process.

Does the above get violated when I run the container in a non host network?

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -276,6 +263,36 @@ func (t *Tracer) run() {
...
+			// It does exist a net link which index is 0.
+			// But eBPF bindsnoop code often gives 0 as interface number:
+			// https://github.com/iovisor/bcc/blob/63618552f81a2631990eff59fd7460802c58c30b/tools/bindsnoop_example.txt#L16
+			// So, we only use this function if interface number is different than 0.
+			var interf netlink.Link
+			CRPID := int(event.CommonData.Runtime.ContainerPID)
+			err := nsenter.NetnsEnter(CRPID, func() error {
+				interf, err = netlink.LinkByIndex(interfaceNum)
+				if err != nil {
+					msg := fmt.Sprintf("Cannot find interface for index %d : %s", interf, err)
```
alban, line 280:
```suggestion
					msg := fmt.Sprintf("Cannot find interface for index %v : %s", interf, err)
```
interf is an interface, so it cannot be printed with %d

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -276,6 +263,36 @@ func (t *Tracer) run() {
...
+					return nil
+				}
+				return nil
+			})
+			if err != nil {
+				msg := fmt.Sprintf("Cannot enter container process having pid %d : %s", CRPID, err)
+				t.eventCallback(types.Base(eventtypes.Err(msg)))
+				return
+			}
+			interfaceString = interf.Attrs().Name
```
alban, line 291:
```suggestion
			if interf != nil {
				interfaceString = interf.Attrs().Name
			}
```

pkg/gadgets/trace/bind/tracer/tracer.go
alban:
```suggestion
					msg := fmt.Sprintf("Cannot find interface for index %d (ContainerPID: %d) : %s", interfaceNum, CRPID, err)
```

alban:
I see that the implementation of `netlink.LinkByIndex` changed: it is reusing the netlink socket that was created before changing the network namespace, so entering the network namespace does not have any effect on `netlink.LinkByIndex`.

I think the new code pattern within the `NetnsEnter` callback should be:
```go
h, err := netlink.NewHandle()
if err != nil {
    ...
}
defer h.Close()
interf, err = h.LinkByIndex(...)
if err != nil {
    ...
}
```
So that `LinkByIndex` acts on a netlink socket created by `NewHandle` while being in the correct network namespace.

pkg/gadgets/trace/bind/tracer/tracer.go
mauriciovasquezbernal, line 280: You want to print the index here, not interf. 

```suggestion
					msg := fmt.Sprintf("Cannot find interface for index %d : %s", interfaceNum, err)
```

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -276,6 +263,36 @@ func (t *Tracer) run() {
...
+			// https://github.com/iovisor/bcc/blob/63618552f81a2631990eff59fd7460802c58c30b/tools/bindsnoop_example.txt#L16
+			// So, we only use this function if interface number is different than 0.
+			var interf netlink.Link
+			CRPID := int(event.CommonData.Runtime.ContainerPID)
+			err := nsenter.NetnsEnter(CRPID, func() error {
+				interf, err = netlink.LinkByIndex(interfaceNum)
+				if err != nil {
+					msg := fmt.Sprintf("Cannot find interface for index %d : %s", interf, err)
+					t.eventCallback(types.Base(eventtypes.Err(msg)))
+					return nil
```
mauriciovasquezbernal, line 282: All this logic is running inside a function, so instead of calling t.eventCallback, it should return an error here, then the conditional on L286 will catch it and trigger the t.eventCallback.

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -276,6 +263,36 @@ func (t *Tracer) run() {
...
+
+		interfaceNum := int(bpfEvent.BoundDevIf)
+		event.Index = interfaceNum // for debugging purposes. will be removed.
+		if interfaceNum != 0 {
+			// It does exist a net link which index is 0.
+			// But eBPF bindsnoop code often gives 0 as interface number:
+			// https://github.com/iovisor/bcc/blob/63618552f81a2631990eff59fd7460802c58c30b/tools/bindsnoop_example.txt#L16
+			// So, we only use this function if interface number is different than 0.
+			var interf netlink.Link
+			CRPID := int(event.CommonData.Runtime.ContainerPID)
```
mauriciovasquezbernal, line 276: Unfortunately `event.CommonData.Runtime.ContainerPID` isn't available at this point. We enrich the events after calling `t.eventCallback` at the bottom. Few options:

- use the event.Pid: it'll be racy as Alban mentioned in the other comment
- extend the gadget to provide the network namespace inode number and use it instead of the pid to enter the namespace. 
- Is there a way to enter the network namespace from the mount namespace? (I don't think so)

@alban what do you think about this? Since we're deprecating this gadget on the next release, I think using event.Pid is good enough, it was broken in any case.

AdvH039: Yes, while I was testing the gadget I was obtaining a zero value for `event.CommonData.Runtime.ContainerPID` and it was leading to issues while resolving namespace.
@mauriciovasquezbernal I think we can try out the solution @alban  mentioned [here](https://github.com/inspektor-gadget/inspektor-gadget/pull/3913#issuecomment-2602033969). 

> extend the gadget to provide the network namespace inode number and use it instead of the pid to enter the namespace.

IMO Inode numbers found in kernel space don't seem useful in obtaining the namespace and you can only find them in userspace through a PID. I don't know how you would directly obtain a user space inode number through any other means.

AdvH039: @alban I tried reading the node of the linked list by using bpf_probe_read_kernel like this
`struct net_device *dev;
bpf_probe_read_kernel(dev, sizeof(*dev), head->first);
`
But I get this error:

`Error: starting operators: starting operator "oci": starting operator "ebpf": creating eBPF collection: program ig_bind_ipv4_x: load program: permission denied: 66: (79) r2 = *(u64 *)(r7 +304): R7 invalid mem access 'scalar' (98 line(s) omitted)
`

Should I just replace `event.CommonData.Runtime.ContainerPID` with `event.Pid` and complete the PR and explore the in kernel implementation in a different PR?

alban: >  I get this error

You should get more context by using `-v`.

> Should I just replace `event.CommonData.Runtime.ContainerPID` with `event.Pid` and complete the PR and explore the in kernel implementation in a different PR?

Ok!

Also, Mauricio told me that my comment in https://github.com/inspektor-gadget/inspektor-gadget/pull/3913#issuecomment-2607153028 could be wrong and that using `netlink.LinkByIndex` in the callback should be fine. Sorry for the wrong analysis.

mauriciovasquezbernal: IMO for the built-in gadget we should try to fix the bug without invoking eBPF to keep it as simple as possible.

AdvH039: I wanted to implement it in the kernel for the image-based gadget only as it seems like a viable option.

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -26,11 +26,12 @@ import (
 	"github.com/cilium/ebpf/link"
 	"github.com/cilium/ebpf/perf"
 	"github.com/vishvananda/netlink"
-
+
```
mauriciovasquezbernal, line 29: Please be sure to run `make lint` for format this file. (remove trailing while spaces on this case)

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -276,6 +259,38 @@ func (t *Tracer) run() {
...
+		interfaceNum := int(bpfEvent.BoundDevIf)
+		if interfaceNum != 0 {
+			// It does exist a net link which index is 0.
+			// But eBPF bindsnoop code often gives 0 as interface number:
+			// https://github.com/iovisor/bcc/blob/63618552f81a2631990eff59fd7460802c58c30b/tools/bindsnoop_example.txt#L16
+			// So, we only use this function if interface number is different than 0.
+			
+			var interf netlink.Link
+			PID := int(event.Pid)
```
mauriciovasquezbernal, line 272:
```suggestion
			pid := int(event.Pid)
```

mauriciovasquezbernal (commented):
Please fix the formatting. It should be to go after that.

mauriciovasquezbernal (approved):
LGTM. Thanks. 

@alban do you want to review this again?

alban (changes requested):
2 small changes to do. It should be good after that. Thank you :)

pkg/gadgets/trace/bind/tracer/tracer.go
```diff
@@ -276,6 +259,38 @@ func (t *Tracer) run() {
...
+				interf, err = netlink.LinkByIndex(interfaceNum)
+				if err != nil {
+					return err
+				}
+				return nil
+			})
+			if err != nil {
+				msg := fmt.Sprintf("Cannot find interface name for index : %d (PID : %d) : %s", interfaceNum, pid, err)
+				t.eventCallback(types.Base(eventtypes.Err(msg)))
+				return
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Good catch Alban! I remember I had this issue with the old version while testing it.

AdvH039:
Made the changes. @alban I just used `continue `as `interfaceString` defaults to `""` when the if statement is continued.

AdvH039:
@alban Done :)

alban (approved):
LGTM!
Could you please squash the two commits into one, and rebase on top of origin/main?

AdvH039:
@alban Done !!

PR #3914, opened 2025-01-20
operators: align operator priorities

### LINKED ISSUE #3731: [RFE] Refactor / align operator priorities
## Current situation

Operators use mostly hardcoded values between 0 and 10000.

## Impact

Every time we add a new operator, we have to think about which priority to give and it might be confusing to find a proper value.

## Ideal future situation

We provide a couple of consts like `PrioritySource`, `PrioritySink` and `PriorityPostprocess` for guidance. Maybe later on, we should reintroduce a dependency solver as well.

## Implementation options

Add mentioned consts to the operator package.

KapilSareen:
### Fixes #3731

added consts `PrioritySource`, `PrioritySink` and `PriorityPostprocess`

### DIFF
--- a/pkg/operators/operators.go
+++ b/pkg/operators/operators.go
@@ -58,6 +58,11 @@ type (
 // MapPrefix is used to avoid clash with maps and other eBPF objects when added
 // to gadget context.
 const MapPrefix string = "map/"
+const (
+	PrioritySource      = -1000
+	PriorityPostprocess = 0
+	PrioritySink        = 1000
+)
 
 type Operator interface {
 	// Name must return a unique name for the operator


KapilSareen:
@flyth Could you confirm if the assigned values align with your expectations, or would you like me to adjust them? Please let me know.

KapilSareen:
> I think the values should be -1000 (Source), 0 (Default / Postprocess / Manipulate) and 1000 (Sink).

Updated the commit according to this.



> They then need to be adopted by all operators at the same time. We're currently looking into how to rework the operators by using composition of interfaces / types, which might address the adoption part.

Regarding this, do you want this to be implemented in this PR itself or in future?
Also if you want this to be implemented now, can you give more details how you plan to implement this?

Issue #3902, opened 2025-01-16
[RFE] CI to report memory usage of ig with pprof


alban:
## Current situation

ig seems to have RAM usage of around 100-120 MB (#3541). We don't have a easy way to know where the memory allocations come from. It is possible to use see the flamegraph of memory usage with pprof, but it is not enabled by default

## Impact

- We don't know if the memory allocations are reasonable.
- When trying to reduce the memory usage, we don't know where to start.

## Ideal future situation

The CI includes a job executed for each PR giving us a flamegraph for:
- pprof inuse_space
- pprof alloc_space

We would need to run a gadget and wait a little while (~10s) after it is initialized.

## Implementation options

Using pprof:
- see patches from #3541.
- See commit ff19514a17d681762d0e7e7d32ee7464a7772cdf

## Additional information

This issue is part of #3541.

This issue is similar to #3900 but for dynamic memory usage instead of static analysis of the ig binary.

KapilSareen:
Hi @alban, I am working on this issue. I'm having a few doubts regarding the flamegraph implementation

KapilSareen:
Hi @alban, I am working on this issue. 
I'm having a few doubts regarding the flamegraph implementation. I have implemented `inuse_space` and `alloc_space` [here](https://github.com/KapilSareen/inspektor-gadget/actions/runs/12909586819/job/35997615184), but i'm not much clear about how you wish to implement flamegraphs, can you give more details?

alban:
I am not sure how to do the framegraph...
But your text result is already very good: I can see that btf is using the most of inuse_space, and kallsyms has been the most allocated in alloc_space.
Let's get the text done first, and we can see if we still need the flamegraph later.

alban:
> Would you like this in the same format as the previous one, such as in summary steps, or should I leave it as is?

Yes, same as before.

KapilSareen:
Opened PR. You can view the final example output [here](https://github.com/KapilSareen/inspektor-gadget/actions/runs/12916995099).

Issue #3541, opened 2024-10-04
[Research] `ig` memory usage investigation


burak-ok:
## Simple measurement using `htop`
`ig` seems to have RAM usage of around 100-120 MB. To measure this correctly in our experiments lets add the following loop to our program in a separate go routine.
```golang
go func() {
  for {
    runtime.GC()
    debug.FreeOSMemory()
    time.Sleep(1 * time.Second)
  }
}()
```
This manually runs the garbage collector and asks it to release its free memory back to the system.
Let's now run a gadget and look at the memory consumption:
```bash
$ ig run trace_dns
```
In `htop` I get a `RES` size of `111M` and a `SHR` of `52M`

## Using `pprof`
To dig deeper we can use `pprof` which is a golang profiler.
While building it is also advisable to use `CGO_ENABLED=0`
```golang
import (
  _ "net/http/pprof"
  ...
)

func main() {
  go func() {
    http.ListenAndServe("localhost:6060", nil)
  }
  ...
}
```
Now if we run `ig run trace_dns` we can connect to the pprof server in another terminal:
```bash
$ go tool pprof -http=:8081 http://localhost:6060/debug/pprof/heap
```
A browser will start up and shows us the heap usage at that timestamp. So wait some seconds until the `ig` process fully started. We want to capture the memory usage while the gadget is running.
When switching to `View->Flamegraph` we can see which parts of our code allocated the most memory:
![image](https://github.com/user-attachments/assets/321437e7-c00e-41ac-a1fb-d9e31f58c982)
So `pprof` shows a total golang memory usage of `50M`. Of that `50M` our `FanotifyEbpf` notifier uses `46M`. Almost all of that is in the BTF Spec.

But from our statistics from `htop`, `/proc/<PID>/smaps` and are reporting a usage of more than `100M`. So where is the rest of the usage that is not reported by `pprof`

## The missing memory not discovered by `pprof`

I am using the `smem` tool which gives easy access to `USS` and `RSS`.
`USS` is the unique memory size that a process has. So if no other program is using the shared part it will equal to the `RSS`.

Running only a single instance of `ig run trace_dns` gives me the following output:
```bash
$ sudo smem -r -k -P "./ig"
    PID User     Command                         Swap      USS      PSS      RSS 
1989621 root     ./ig run trace_dns                 0   111.7M   111.7M   111.7M 
```
[...]

PR #3915, opened 2025-01-20
CI: static analysis of ig binary size

### LINKED ISSUE #3900: [RFE] CI to report static analysis of ig binary size
## Current situation

The ig binary from [release v0.36.0](https://github.com/inspektor-gadget/inspektor-gadget/releases/tag/v0.36.0) is 110MB:
```
$ ls -lh ig
-rwxr-xr-x 1 alban alban 110M Jan  6 09:23 ig
```

But we don't know which components are taking most of the space.

## Impact

- We don't know if the size is reasonable.
- When trying to optimize the size, we don't know where to start.

## Ideal future situation

The CI includes a job executed for each PR telling us:
* The size used by each Go package
* The size used by each ELF section

## Implementation options

* https://github.com/Zxilly/go-size-analyzer (see [example](https://stackoverflow.com/questions/70764915/how-to-check-the-size-of-packages-linked-into-my-go-code))
* https://gsa.zxilly.dev/
* https://stackoverflow.com/questions/70764915/how-to-check-the-size-of-packages-linked-into-my-go-code (see [example](https://github.com/inspektor-gadget/inspektor-gadget/issues/3541#issuecomment-2503917968))

## Additional information

This issue is part of #3541.

KapilSareen:
## Fixes #3900

Adds a CI workflow that reports static analysis of ig binary size

## Testing done

Yes, tested on local fork. Attaching screenshot for reference:

image

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -711,6 +711,47 @@ jobs:
           rm -rf /tmp/.buildx-cache
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
+  analyse-ig-binary-size:
+    name: IG binary size analysis
+    runs-on: ubuntu-latest
+    needs:
+      - build-ig
+    strategy:
+      matrix:
+        platform: [amd64, arm64]
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: true
+        id: go
+      - name: Install go-size-analyzer
+        run: |
+          go install github.com/Zxilly/go-size-analyzer/cmd/gsa@e47aa2561e70b937f2be7441b2eeca984ff60e8b # v1.7.6
+      - name: Get ig-linux-${{matrix.platform}}.tar.gz from artifact
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: ig-linux-${{matrix.platform}}-tar-gz
+          path: /home/runner/work/inspektor-gadget/
+      - name: Unpack ig-linux-${{matrix.platform}}.tar.gz
+        run: |
+          tar zxvf /home/runner/work/inspektor-gadget/ig-linux-${{matrix.platform}}.tar.gz
+          gsa ./ig -f "json" --compact | tee out.json
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size, type: .value.type}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
+        shell: bash
+      - name: Generate Markdown Table
+        uses: buildingcash/json-to-markdown-table-action@b442169239ef35f1dc4e5c8c3d47686c081a7e65  #v1.1.0
+        id: table
+        with:
+          json_file_path: ig-packages.json
+      - name: Add Markdown Table
+        env:
+          TABLE_OUTPUT: ${{ steps.table.outputs.table }}
+        run: |
+          echo '### Package Size Summary' >> $GITHUB_STEP_SUMMARY
+          echo "$TABLE_OUTPUT" >> $GITHUB_STEP_SUMMARY
+
   scan-gadget-container-images:
     name: Scan gadget img
     # level: 2


.github/workflows/ig-analysis.yml
```diff
@@ -0,0 +1,34 @@
...
+          cache: true
+        id: go
+
+      - name: Install dependencies
+        run: |
+          go mod tidy
+
+      - name: Install go-size-analyzer
+        run: |
+          go get github.com/Zxilly/go-size-analyzer/cmd/gsa@latest
```
alban, line 26: Since this has a recent release [v1.7.6](https://github.com/Zxilly/go-size-analyzer/releases/tag/v1.7.6), should we use it instead of the latest tag?

.github/workflows/ig-analysis.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Run make ig
+        run: |
+          make ig
```
alban, line 30: ig is also built in .github/workflows/inspektor-gadget.yml with job "Build ig ${{ matrix.os }} ${{ matrix.platform }} container image as artifacts". Should we reuse that build or is it better to do this independent build in parallel?

alban (commented):
Thanks for working on this.

I added 2 questions, which I don't know the answer to. Maybe others have an opinion?

This does not give the details of the main package (github.com/inspektor-gadget/inspektor-gadget), but that's a good start.

alban:
> > This does not give the details of the main package (github.com/inspektor-gadget/inspektor-gadget)
>
> wdym @alban ? I can see it at the top with 15.79%.

We don't have granularity for the different packages in the `pkg/*` directories. Currently we can't see that `pkg/gadgets` is taking most of the space, followed by `pkg/columns`, etc.

It is possible to do that with a custom tool (see https://github.com/inspektor-gadget/inspektor-gadget/issues/3541#issuecomment-2503917968) but it is more work... so maybe it is better to start with your PR as is, and improve it later.

.github/workflows/ig-analysis.yml
KapilSareen: @eiffel-fl @alban 

I tried implementing your suggestion using the pre-existing artifact, but I encountered a complication. It seems that this workflow needs to wait for the artifact to be uploaded first, and there doesn't appear to be a straightforward solution for this.

However, I came across a [GitHub Action](https://github.com/marketplace/actions/wait-on-check) that could help with this. Would you like me to try using this, or do you have another approach in mind?

Just to clarify, I'm relatively new to writing CI/CD workflows, so I might be missing something simple.

KapilSareen: oh this seems like a good idea. Ignore my commit for now , I'll implement this.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,39 @@ jobs:
...
+        with:
+          go-version: '1.23.4'
+          cache: true
+        id: go
+      - name: Install dependencies
+        run: |
+          go mod tidy
+      - name: Install go-size-analyzer
+        run: |
+          go get github.com/Zxilly/go-size-analyzer/cmd/gsa@v1.7.6
```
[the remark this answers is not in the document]
KapilSareen: I updated it to `latest` for now. Please let me know if you need me to revert the change.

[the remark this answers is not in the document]
KapilSareen: sorry @eiffel-fl , I guess the `go mod tidy` was also not required. Removed it.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,34 @@ jobs:
...
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
+        with:
+          go-version: '1.23.4'
+          cache: true
+        id: go
+      - name: Install go-size-analyzer
+        run: |
+          go get github.com/Zxilly/go-size-analyzer/cmd/gsa@latest
```
alban, line 714: The CI says:

>	'go get' is no longer supported outside a module.
>	To build and install a command, use 'go install' with a version,
>	like 'go install example.com/cmd@latest'
>	For more information, see https://golang.org/doc/go-get-install-deprecation
>	or run 'go help get' or 'go help install'.

```suggestion
          go install github.com/Zxilly/go-size-analyzer/cmd/gsa@latest
```

alban:
go-size-analyzer can output json, which can be converted to a markdown table with [json-to-markdown-table-action](https://github.com/marketplace/actions/json-to-markdown-table-action) and then displayed nicely in the job output with the [Job Summaries feature](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/#create-summaries).

That would be easier to find out the data in the CI user interface.

KapilSareen:
fixed a typo

KapilSareen:
>  Is it possible to use the hash as well? Like what we do for actions and docker images.

LGTM. It would also be more consistant with the other jobs too, updated it.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,46 @@ jobs:
...
+        shell: bash
+      - name: Generate Markdown Table
+        uses: buildingcash/json-to-markdown-table-action@v1.1.0
+        id: table
+        with:
+          json_file_path: packages.json
+      - name: Add Markdown Table
+        run: |
+          echo '### Package Size Summary' >> $GITHUB_STEP_SUMMARY
+          echo "${{ steps.table.outputs.table }}" >> $GITHUB_STEP_SUMMARY
```
alban, line 738: The expressions inside of `${{ }}` are evaluated and substituted with the resulting values before the shell script is run, so if the output mistakenly contains a payload like `a"; command`, then it would do something unexpected.

Github recommends the following code pattern to avoid this issue:

```suggestion
      - name: Add Markdown Table
        env:
          TABLE_OUTPUT=${{ steps.table.outputs.table }}
        run: |
          echo '### Package Size Summary' >> $GITHUB_STEP_SUMMARY
          echo "$TABLE_OUTPUT" >> $GITHUB_STEP_SUMMARY
```

cc @mauriciovasquezbernal

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,46 @@ jobs:
...
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: ig-linux-amd64-tar-gz
+          path: /home/runner/work/inspektor-gadget/
+      - name: Unpack ig-linux-amd64.tar.gz
+        run: |
+          tar zxvf /home/runner/work/inspektor-gadget/ig-linux-amd64.tar.gz
+      - name: Generate JSON
+        run: |
+          gsa ./ig -f "json" --compact > out.json
```
alban, line 725: Could you also print the json in the logs?
```suggestion
          gsa ./ig -f "json" --compact | tee out.json
```
In this way, if there is a bug in the gsa output or in the `jq` query below, it will be possible to debug it.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,46 @@ jobs:
...
+      - name: Extract Packages as JSON Array
+        run: |
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size}] | sort_by(.size) | reverse' out.json > packages.json
```
alban, line 728: The filename `packages.json` is often used in Javascript projects, so I think it is better to name it differently because we don't use the same syntax as in those JS projects. I suggest the `ig-` prefix, in case we later analyse the binary size of other binaries (e.g. gadgetctl, kubectl-gadget).

Also, I suggest to use `tee` to add the output in the job logs, to be able to debug it later.

```suggestion
          jq '[.packages | to_entries[] | {package: .key, size: .value.size}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
```

alban (commented):
One line has an empty key:

image

I suggest adding debug logs with the `tee` command to be able to debug this.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+          gsa ./ig -f "json" --compact | tee out.json
+      - name: Extract Packages as JSON Array
+        run: |
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
+        shell: bash
+      - name: Generate Markdown Table
+        uses: buildingcash/json-to-markdown-table-action@v1.1.0
+        id: table
+        with:
+          json_file_path: packages.json
```
alban, line 734:
```suggestion
          json_file_path: ig-packages.json
```

KapilSareen:
Getting this error:
> 737 |           TABLE_OUTPUT=${{ steps.table.outputs.table }}
    |           ^~~~~~~~~~~~~~~~
Error: Process completed with exit code 1.

I think it should have been like:
```
env:
  TABLE_OUTPUT: ${{ steps.table.outputs.
```
what do you think @alban ?

.github/workflows/inspektor-gadget.yml
KapilSareen: Getting this error:
> 737 |           TABLE_OUTPUT=${{ steps.table.outputs.table }}
    |           ^~~~~~~~~~~~~~~~
Error: Process completed with exit code 1.

I think it should have been like:
```
env:
  TABLE_OUTPUT: ${{ steps.table.outputs.
```
what do you think @alban ?

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+          path: /home/runner/work/inspektor-gadget/
+      - name: Unpack ig-linux-amd64.tar.gz
+        run: |
+          tar zxvf /home/runner/work/inspektor-gadget/ig-linux-amd64.tar.gz
+      - name: Generate JSON
+        run: |
+          gsa ./ig -f "json" --compact | tee out.json
+      - name: Extract Packages as JSON Array
+        run: |
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
```
alban, line 728: I see what's causing the empty entry:
```
┌──────────────────────────────────────────────────────────────────────────────────────────────┐
│ ig                                                                                           │
├─────────┬───────────────────────────────────────────────────────────────┬────────┬───────────┤
│ PERCENT │ NAME                                                          │ SIZE   │ TYPE      │
├─────────┼───────────────────────────────────────────────────────────────┼────────┼───────────┤
│ 1.01%   │                                                               │ 1.2 MB │ generated │
```

Or, in json:
```json
{
   "analyzers" : [
      "disasm",
      "dwarf",
      "pclntab"
   ],
   "name" : "ig",
   "packages" : {
      "" : {
         "name" : "",
         "size" : 1163344,
         "subPackages" : {},
         "symbols" : [],
         "type" : "generated"
      },
   "size" : 117402575
}
```

I think we can just add the column "type" for now. Something like:
```suggestion
          jq '[.packages | to_entries[] | {package: .key, size: .value.size, type: .value.type}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
```

alban:
> Also, would it be okay if I take on https://github.com/inspektor-gadget/inspektor-gadget/issues/3902 as well?

Yes, it is fine :) Please comment on #3902 to say it if you take it.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+  analyze-and-build:
+    name: IG binary size analysis
+    runs-on: ubuntu-latest
+    needs: 
+      - build-ig 
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
+        with:
+          go-version: '1.23.4'
```
mauriciovasquezbernal, line 709:
```suggestion
          go-version: ${{ env.GO_VERSION }}
```

To make it easier to update the golang version.

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+          cache: true
+        id: go
+      - name: Install go-size-analyzer
+        run: |
+          go install github.com/Zxilly/go-size-analyzer/cmd/gsa@e47aa2561e70b937f2be7441b2eeca984ff60e8b
```
mauriciovasquezbernal, line 714: nit: please add a comment indicating which version it points to as done in #3199

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+          tar zxvf /home/runner/work/inspektor-gadget/ig-linux-amd64.tar.gz
+      - name: Generate JSON
+        run: |
+          gsa ./ig -f "json" --compact | tee out.json
+      - name: Extract Packages as JSON Array
+        run: |
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size, type: .value.type}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
+        shell: bash
+      - name: Generate Markdown Table
+        uses: buildingcash/json-to-markdown-table-action@v1.1.0
```
mauriciovasquezbernal, line 731: pin this by hash as well

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
           rm -rf /tmp/.buildx-cache
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
+  analyze-and-build:
```
mauriciovasquezbernal, line 700:
```suggestion
  analyse-ig-binary-size:
```

?

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+      - name: Unpack ig-linux-amd64.tar.gz
+        run: |
+          tar zxvf /home/runner/work/inspektor-gadget/ig-linux-amd64.tar.gz
+      - name: Generate JSON
+        run: |
+          gsa ./ig -f "json" --compact | tee out.json
+      - name: Extract Packages as JSON Array
+        run: |
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size, type: .value.type}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
+        shell: bash
```
mauriciovasquezbernal, line 729: IMO these three can be in a single step. Any reason to have these too granular steps?

mauriciovasquezbernal (changes requested):
Sorry for chiming in a little bit late. Two blocking comments about using the GO_VERSION env and pinning dependencies by hash.

.github/workflows/inspektor-gadget.yml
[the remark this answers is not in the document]
mauriciovasquezbernal: The code scanning will trigger an alert if we use the tag: https://github.com/inspektor-gadget/inspektor-gadget/security/code-scanning/254

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,48 @@ jobs:
...
+          cache: true
+        id: go
+      - name: Install go-size-analyzer
+        run: |
+          go install github.com/Zxilly/go-size-analyzer/cmd/gsa@e47aa2561e70b937f2be7441b2eeca984ff60e8b
+      - name: Get ig-linux-amd64.tar.gz from artifact.
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: ig-linux-amd64-tar-gz
+          path: /home/runner/work/inspektor-gadget/
```
mauriciovasquezbernal, line 719: Finally, we're are only analyzing the amd64 binary, should we do the same for the arm64 one? (I think it could be done by using a [matrix strategy](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy) without adding a lot of lines of code).

KapilSareen:
How does it look to you now @mauriciovasquezbernal ?

.github/workflows/inspektor-gadget.yml
```diff
@@ -697,6 +697,47 @@ jobs:
...
+      - name: Get ig-linux-${{matrix.platform}}.tar.gz from artifact
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: ig-linux-${{matrix.platform}}-tar-gz
+          path: /home/runner/work/inspektor-gadget/
+      - name: Unpack ig-linux-${{matrix.platform}}.tar.gz
+        run: |
+          tar zxvf /home/runner/work/inspektor-gadget/ig-linux-${{matrix.platform}}.tar.gz
+          gsa ./ig -f "json" --compact | tee out.json
+          jq '[.packages | to_entries[] | {package: .key, size: .value.size, type: .value.type}] | sort_by(.size) | reverse' out.json | tee ig-packages.json
```
alban, line 727: In a future PR, we could get the sizes for ig, gadgetctl and kubectl-gadget.

alban (approved):
LGTM, thanks!

The CI failure is not related to this PR:
```
ERROR: 1 dead links found!
[✖] https://retina.sh/docs/Contributing/ → Status: 404
```

.github/workflows/inspektor-gadget.yml
KapilSareen: Yes sure! If you decide to create an issue for that I'll be happy to work on that.

PR #3917, opened 2025-01-21
pkg/oci: Log errors as warning instead of debug when listing images.

eiffel-fl:


alban:
When would this error happen? Could you give an example of output when it happens?

mauriciovasquezbernal:
I'm not sure about this change as well. IMO it's more difficult to handle it on the caller side than just printing warnings on the library itself. In any case, if we want to change this, then also callers should be updated on this PR.

mauriciovasquezbernal (approved):
LGTM. Thanks!

PR #3928, opened 2025-01-23
trace_tcp: add option to see only failed connections

alban:
# trace_tcp: add option to see only failed connections

When using trace_tcp to monitor connection issues, we don't want to see successful connections. Currently, users can use a filter like `--filter ret!=0` but this implements a filter in userspace, which is less efficient than in ebpf. This patch adds a filter in ebpf to save cpu cycles.

## How to use

Use `--failure-only` on the command line.

## Testing done

Tested with the following:
```
echo|nc 1.1.1.1 80     # successful connect
echo|nc 127.0.0.1 8181 # failed connect
```

Results without and with `--failure-only`:
```
$ sudo ./ig run ghcr.io/inspektor-gadget/gadget/trace_tcp:latest --verify-image=false -F proc.comm==nc --host --connect-only
RUNTIME.CONTAINER… COMM              PID        TID SRC                      DST                     TYPE      ERROR
                   nc             601382     601382 192.168.0.22:40316       1.1.1.1:80              connect
                   nc             601391     601391 127.0.0.1:51788          127.0.0.1:8181          connect   ECONNRE

$ sudo ./ig run ghcr.io/inspektor-gadget/gadget/trace_tcp:latest --verify-image=false -F proc.comm==nc --host --connect-only --failure-only
RUNTIME.CONTAINER… COMM              PID        TID SRC                      DST                     TYPE      ERROR
                   nc             601438     601438 127.0.0.1:38594          127.0.0.1:8181          connect   ECONNRE
```

cc @johananl

### DIFF
--- a/gadgets/trace_tcp/ebpf/accept.h
+++ b/gadgets/trace_tcp/ebpf/accept.h
@@ -28,8 +28,12 @@ static __always_inline int handle_sys_accept_x(struct syscall_trace_exit *ctx)
 	skpp = bpf_map_lookup_elem(&tcp_tid_sock, &tid);
 	if (!skpp)
 		return 0;
-	sk = *skpp;
 
+	// User does not want to see successful events
+	if (failure_only && ctx->ret >= 0)
+		goto end;
+
+	sk = *skpp;
 	family = BPF_CORE_READ(sk, __sk_common.skc_family);
 
 	fill_tuple(&t, sk, family);
--- a/gadgets/trace_tcp/ebpf/close.h
+++ b/gadgets/trace_tcp/ebpf/close.h
@@ -16,6 +16,11 @@ static __always_inline void handle_tcp_close(struct pt_regs *ctx,
 	struct event *event;
 	u16 family;
 
+	// Close events don't have errors.
+	// User does not want to see successful events.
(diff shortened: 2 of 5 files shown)


mauriciovasquezbernal (commented):
Should this new variable be used around https://github.com/inspektor-gadget/inspektor-gadget/blob/a7bd421594f44a2b8475e1cd93fb25538781824f/gadgets/trace_tcp/ebpf/accept.h#L48-L49 too?

I see a user could want to see the failures for both, connect and accept events. 

Also, since we don't support getting the failure in the close event, should we discard those events if failure-only is true?

alban:
> I see a user could want to see the failures for both, connect and accept events.

It makes sense. I updated accordingly. However, failed accept would be rare: errors such as `EBADF` would not appear since we don't emit an event when we can't find the tuple.

> Also, since we don't support getting the failure in the close event, should we discard those events if failure-only is true?

It makes sense, I updated accordingly.

Unrelated to this PR: I just realised that close events are not displayed unless the user uses `--host` because we don't have the process data, so we don't enrich by container.

mauriciovasquezbernal (approved):
LGTM and works as expected. Thanks!

PR #3925, opened 2025-01-22
ci: report memory usage of ig with pprof

KapilSareen:

PR #3932, opened 2025-01-23
kallsyms: implement just in time lookups without allocations

alban:
# kallsyms: implement just in time lookups without allocations

This implements kernel symbol lookups in /proc/kallsyms without unnecessary dynamic memory allocations.

This is 2x faster and almost completely removes dynamic memory allocations.

```
$ go test -exec=sudo -run=Benchmark -bench=. ./pkg/kallsyms/... -count 1
goos: linux
goarch: amd64
pkg: github.com/inspektor-gadget/inspektor-gadget/pkg/kallsyms
BenchmarkKernelSymbolAddressJITParsing-4               5         208193754 ns/op            4688 B/op         19 allocs/op
BenchmarkKernelSymbolAddressPreLoading-4               3         425194186 ns/op        105551938 B/op    639608 allocs/op
PASS
ok      github.com/inspektor-gadget/inspektor-gadget/pkg/kallsyms       4.746s
```

Once this PR is merged, I will make use of it in #3871.

## How to use

No changes.

## Testing done

Let's run the CI.

### DIFF
--- a/pkg/kallsyms/kallsyms.go
+++ b/pkg/kallsyms/kallsyms.go
@@ -135,11 +135,6 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
 }
 
 // SymbolExists returns true if the given symbol exists in the kernel.
-func (k *KAllSyms) SymbolExists(symbol string) bool {
-	_, ok := k.symbolsMap[symbol]
-	return ok
-}
-
 func SymbolExists(symbol string) bool {
 	_, _, err := KernelSymbolAddress(symbol)
 	return err == nil
--- a/pkg/kallsyms/kallsyms.go
+++ b/pkg/kallsyms/kallsyms.go
@@ -17,6 +17,8 @@ package kallsyms
 
 import (
 	"bufio"
+	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -28,6 +30,8 @@ import (
 
 	ebpfutils "github.com/inspektor-gadget/inspektor-gadget/pkg/utils/ebpf"
 )
+
+var ErrAmbiguousKsym = errors.New("multiple kernel symbols with the same name")
 
 type KAllSyms struct {
 	// symbols is a slice of kernel symbols. Order is preserved.
@@ -43,6 +47,10 @@ type kernelSymbol struct {
 }
 
 // NewKAllSyms reads /proc/kallsyms and returns a KAllSyms.
+//
+// It is meant to be used when all the symbols need to be permanently loaded.
+// It comes with a big resource penalty. For looking up only a few symbols, it
+// is more efficient to use SymbolExists or KernelSymbolAddress.
 func NewKAllSyms() (*KAllSyms, error) {
 	file, err := os.Open("/proc/kallsyms")
 	if err != nil {
@@ -130,6 +138,99 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
 func (k *KAllSyms) SymbolExists(symbol string) bool {
 	_, ok := k.symbolsMap[symbol]
 	return ok
+}
+
+func SymbolExists(symbol string) bool {
+	_, _, err := KernelSymbolAddress(symbol)
+	return err == nil
+}
+
+// KernelSymbolAddress looks up a symbol from /proc/kallsyms.
+//
+// symbol: the symbol to lookup
+//
+// Returns the address of the symbol and its module if any
+func KernelSymbolAddress(symbol string) (uint64, string, error) {
+	file, err := os.Open("/proc/kallsyms")
+	if err != nil {
+		return 0, "", err
+	}
+	defer file.Close()
+
+	return kernelSymbolAddressFromReader(file, symbol)
+}
+
+// kernelSymbolAddressFromReader looks up a symbol from the reader.
+//
+// reader: source; it should be in the same file format as /proc/kallsyms
+// symbol: the symbol to lookup
+//
+// Returns the address of the symbol and its module if any
+func kernelSymbolAddressFromReader(r io.Reader, symbol string) (uint64, string, error) {
+	var (
+		count int
+		addr  uint64
+		mod   string
+		err   error
+	)
+
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		// Lines looks like:
+		// ffffffff906e97e0 T security_bprm_check
+		// ffffffffc1cb5010 T netem_module_init    [sch_netem]
+
+		line := scanner.Bytes()
+		typeIdx := bytes.IndexByte(line, byte(' '))
+		if typeIdx == -1 || typeIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no type", line)
+		}
+		nameIdx := bytes.IndexByte(line[typeIdx+1:], byte(' '))
+		if nameIdx == -1 || typeIdx+1+nameIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no symbol", line)
+		}
+		modIdx := bytes.IndexByte(line[typeIdx+1+nameIdx+1:], byte('\t'))
+		var nameBytes []byte
+		if modIdx == -1 {
+			// no module
+			nameBytes = line[typeIdx+1+nameIdx+1:]
+		} else {
+			// module found
+			nameBytes = line[typeIdx+1+nameIdx+1 : typeIdx+1+nameIdx+1+modIdx]
+		}
+		if !bytes.Equal([]byte(symbol), nameBytes) {
+			continue
+		}
+
+		addr, err = strconv.ParseUint(string(line[:typeIdx]), 16, 64)
+		if err != nil {
+			return 0, "", fmt.Errorf("parsing line %q: invalid address: %w", line, err)
+		}
+		if modIdx != -1 {
+			mod = string(line[typeIdx+1+nameIdx+1+modIdx+1:])
+			mod = strings.Trim(mod, "[]")
+		}
+		count++
+		if count > 1 {
+			break
+		}
+	}
+
+	if err = scanner.Err(); err != nil {
+		return 0, "", fmt.Errorf("reading kallsyms: %w", err)
+	}
+
+	switch count {
+	case 0:
+		return 0, "", os.ErrNotExist
+	case 1:
+		return addr, mod, nil
+	default:
+		// Multiple addresses for a symbol have been found. Like libbpf
+		// and cilium/ebpf, reject referring to ambiguous symbols.
+		return 0, "", fmt.Errorf("symbol %s: duplicate found at address 0x%x: %w",
+			symbol, addr, ErrAmbiguousKsym)
+	}
 }
 
 var (
--- a/pkg/kallsyms/kallsyms_test.go
+++ b/pkg/kallsyms/kallsyms_test.go
@@ -80,3 +80,30 @@ func TestRealKAllSyms(t *testing.T) {
 	// Since we use RequireRoot, it should not happen.
 	require.NotEqual(t, 0, addr, "bpf_prog_fops has a zero address")
 }
+
+// BenchmarkKernelSymbolAddressJITParsing searches for the symbol while
+// reading the kallsyms file
+func BenchmarkKernelSymbolAddressJITParsing(b *testing.B) {
+	utilstest.RequireRoot(b)
+	b.ReportAllocs()
+
+	for n := 0; n < b.N; n++ {
+		exists := SymbolExists("security_bprm_check")
+		require.Equal(b, true, exists, "exists")
+	}
+}
+
+// BenchmarkKernelSymbolAddressPreLoading loads the full kallsyms before
+// looking up the symbol
+func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
+	utilstest.RequireRoot(b)
+	b.ReportAllocs()
+
+	for n := 0; n < b.N; n++ {
+		k, err := NewKAllSyms()
+		require.NoError(b, err)
+		require.NotNil(b, k)
+		exists := k.SymbolExists("security_bprm_check")
+		require.Equal(b, true, exists)
+	}
+}
--- a/pkg/kallsyms/kallsyms_test.go
+++ b/pkg/kallsyms/kallsyms_test.go
@@ -23,11 +23,6 @@ import (
 	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
 )
 
-func resetState() {
-	symbolsMap = map[string]uint64{}
-	triedGetAddr = map[string]error{}
-}
-
 func TestCustomKAllSyms(t *testing.T) {
 	kAllSymsStr := strings.Join([]string{
 		"0000000000000000 A fixed_percpu_data",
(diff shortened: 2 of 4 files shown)


pkg/kallsyms/kallsyms.go
```diff
@@ -127,9 +131,97 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
...
+	file, err := os.Open("/proc/kallsyms")
+	if err != nil {
+		return 0, "", err
+	}
+	defer file.Close()
+
+	return kernelSymbolAddressFromReader(file, symbol)
+}
+
+// kernelSymbolAddressFromReader
```
mauriciovasquezbernal, line 154: nit: missing docs?

pkg/kallsyms/kallsyms.go
```diff
@@ -127,9 +131,97 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
...
+			return 0, "", fmt.Errorf("parsing line %q: no symbol", line)
+		}
+		modIdx := bytes.IndexByte(line[typeIdx+1+nameIdx+1:], byte(' '))
+		var nameBytes []byte
+		if modIdx == -1 {
+			// no module
+			nameBytes = line[typeIdx+1+nameIdx+1:]
+			if !bytes.Equal([]byte(symbol), nameBytes) {
+				continue
+			}
```
mauriciovasquezbernal, line 186: Factor out this and the one on L189 to be before L195?:

```golang
		var nameBytes []byte
		if modIdx == -1 {
			// no module
			nameBytes = line[typeIdx+1+nameIdx+1:]
		} else {
			// module found
			nameBytes = line[typeIdx+1+nameIdx+1 : typeIdx+1+nameIdx+1+modIdx]
		}

		if !bytes.Equal([]byte(symbol), nameBytes) {
			continue
		}

```

pkg/kallsyms/kallsyms.go
```diff
@@ -127,9 +131,97 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
...
+
+		text := scanner.Text()
+		_, err := fmt.Sscanf(text, "%x %c %s\t%s", &addr, &typ, &name, &mod)
+		if err == nil {
+			// a module was found
+			mod = strings.Trim(mod, "[]")
+		} else {
+			// no module was found
+			_, err = fmt.Sscanf(text, "%x %c %s", &addr, &typ, &name)
+		}
```
mauriciovasquezbernal, line 203: What about using the `modIdx == -1` conditional here? We already know if the module is there or not, so we don't need to check that again.

pkg/kallsyms/kallsyms.go
```diff
@@ -127,9 +131,97 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
...
+		if err != nil {
+			return 0, "", fmt.Errorf("parsing line %q: %w", text, err)
+		}
+		count++
```
mauriciovasquezbernal, line 207: I don't know if this is a worth optimization, but what about testing if count > 1 here to avoid doing uneeded work?

pkg/kallsyms/kallsyms.go
```diff
@@ -127,9 +131,97 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
...
+		line := scanner.Bytes()
+		typeIdx := bytes.IndexByte(line, byte(' '))
+		if typeIdx == -1 || typeIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no type", line)
+		}
+		nameIdx := bytes.IndexByte(line[typeIdx+1:], byte(' '))
+		if nameIdx == -1 || nameIdx+1+nameIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no symbol", line)
+		}
+		modIdx := bytes.IndexByte(line[typeIdx+1+nameIdx+1:], byte(' '))
```
mauriciovasquezbernal, line 179: Can't we rely on the last byte being `]` to check if the module is present?

pkg/kallsyms/kallsyms.go
```diff
@@ -127,9 +131,97 @@ func (k *KAllSyms) LookupByInstructionPointer(ip uint64) string {
...
+		// ffffffff906e97e0 T security_bprm_check
+		// ffffffffc1cb5010 T netem_module_init    [sch_netem]
+
+		line := scanner.Bytes()
+		typeIdx := bytes.IndexByte(line, byte(' '))
+		if typeIdx == -1 || typeIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no type", line)
+		}
+		nameIdx := bytes.IndexByte(line[typeIdx+1:], byte(' '))
+		if nameIdx == -1 || nameIdx+1+nameIdx+1 >= len(line) {
```
mauriciovasquezbernal, line 176:
```suggestion
		if nameIdx == -1 || typeIdx+1+nameIdx+1 >= len(line) {
```
?

pkg/kallsyms/kallsyms.go
mauriciovasquezbernal, line 1: I think it's needed to document when to use `NewKAllSyms` or `KernelSymbolAddress`. It's not immediately clear to the readers that one of the versions has a very big resource penalty when looking a single (or few symbols).

pkg/kallsyms/kallsyms_test.go
```diff
@@ -62,21 +53,83 @@ func TestCustomKAllSyms(t *testing.T) {
...
+	require.NotEqual(t, 0, addr, "addr")
+	require.Equal(t, "", kmod, "kmod")
+	addr, kmod, err = KernelSymbolAddress("socket_file_ops")
+	require.NoError(t, err)
+	require.NotEqual(t, 0, addr, "addr")
+	require.Equal(t, "", kmod, "kmod")
+	addr, kmod, err = KernelSymbolAddress("abcde_bad_name")
+	require.Error(t, err)
+	require.NotEqual(t, 0, addr, "addr")
+	require.Equal(t, "", kmod, "kmod")
```
mauriciovasquezbernal, line 107: Perhaps a table based test could reduce a bit the number of lines.

pkg/kallsyms/kallsyms_test.go
```diff
@@ -62,21 +53,83 @@ func TestCustomKAllSyms(t *testing.T) {
...
+	addr, kmod, err = kernelSymbolAddressFromReader(r(), "socket_file_ops")
+	require.NoError(t, err)
+	require.NotEqual(t, 0, addr, "addr")
+	require.Equal(t, "", kmod, "kmod")
+	addr, kmod, err = kernelSymbolAddressFromReader(r(), "abcde_bad_name")
+	require.Error(t, err)
+	require.NotEqual(t, 0, addr, "addr")
+	require.Equal(t, "", kmod, "kmod")
```
mauriciovasquezbernal, line 81: Same here.

mauriciovasquezbernal (commented):
It seems like another example of "Ask a programmer to review 10 lines of code, he'll find 10 issues. Ask him to do 500 lines and he'll say it looks good."

It looks very good, almost all my comments (but the one about documentation) are nits that can be ignored if you want.

pkg/kallsyms/kallsyms.go
```diff
@@ -132,6 +136,99 @@ func (k *KAllSyms) SymbolExists(symbol string) bool {
...
+		typ   rune
+		mod   string
+		name  string
+	)
+
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		// Lines looks like:
+		// ffffffff906e97e0 T security_bprm_check
+		// ffffffffc1cb5010 T netem_module_init    [sch_netem]
```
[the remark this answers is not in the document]
alban: There is a cilium/ebpf package but:
1. it is internal so it cannot be imported. It is only used for internal use cases.
2. it does not implement LookupByInstructionPointer or SpecUpdateAddresses.

See also https://github.com/cilium/ebpf/pull/1588#issuecomment-2602479675

pkg/kallsyms/kallsyms_test.go
```diff
@@ -80,3 +80,30 @@ func TestRealKAllSyms(t *testing.T) {
...
+}
+
+// BenchmarkKernelSymbolAddressPreLoading loads the full kallsyms before
+// looking up the symbol
+func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
+	utilstest.RequireRoot(b)
+	b.ReportAllocs()
+
+	for n := 0; n < b.N; n++ {
+		k, err := NewKAllSyms()
```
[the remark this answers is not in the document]
alban: It is true that so far NewKallSyms() was optimized for parsing once and reusing, but it is mostly not the right angle for optimizing our use cases.

Our package kallsyms is used in two categories of use cases:
1. Continuous usage (using LookupByInstructionPointer)
  a. Builtin profile/cpu
  b. Image-based gadgets making use of kernel stack (trace_capabilities and trace_tcpdrop)
2. Single use at startup (using SymbolExists and/or SpecUpdateAddresses)
  a. Builtin top/block-io (but only on old kernels)
  b. Builtin top/ebpf
  c. Gadgets using the socket enricher: trace-dns, trace-sni, trace-network, tcpdrop, tcpretrans (both builtin and image-based)

For use case 1, we need to load all the symbols and keep them in memory. This PR does not change that. It was using KAllSyms.LookupByInstructionPointer before and it still does that.

But most gadgets don't fit in that use case, and we had users complaining about memory usage. For use case 2, this PR makes it faster (because symbols were looked up only once or twice). And it also use less memory.

The benchmarks are benchmarking `SymbolExists`, which is used only for use case 2. So I think the benchmarks are correct to look up a symbol only once is good because it matches how it is used by our gadgets.

alban:
> Where do you plan to use this function?

Do you mean kallsyms.SymbolExists? This PR makes use of it wherever KallSyms.SymbolExists was used before, that is in the builtin top/block-io (but only on old kernels). In addition, I will make use of it in WASM via #3871 once merged.

pkg/kallsyms/kallsyms_test.go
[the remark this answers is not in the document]
alban: The problem with doing it the way you suggest is that this will still do a lot of dynamic allocations. Even if they are temporary, they cause more work on the garbage collector, and it is difficult to reclaim the free pages from the OS when it is a large amount of small allocations because they end up being allocated on the heap with brk rather than mmap like for big allocations. Additionally, it requires users to increase the memory limit in the pod spec more than they would do if dynamic allocations are just avoided (Kubernetes cannot use a different memory limit for initialization and for after).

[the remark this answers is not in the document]
alban: #3925 tells us that kallsyms allocated 59.08MB (Alloc Space). Then, the memory was released (0MB in Inuse Space). That's a huge amount to add in the memory limit in pod specs for our users.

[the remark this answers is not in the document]
alban: I am not sure to follow. Both benchmarks are opening the file /proc/kallsyms, reading and parsing it at each iteration of the loop. So they both account for the cost of the syscalls.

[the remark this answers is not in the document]
alban: > With regard to this, can you rebase, this way we have this analysis done here?

Done. In the new [pprof summary](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/13011989412?pr=3932), kallsyms does not appear anymore (neither in Alloc Space nor in Inuse Space)

pkg/kallsyms/kallsyms.go
alban: It actually saves a lot of time (according to my tests on https://github.com/inspektor-gadget/inspektor-gadget/pull/3871#discussion_r1913325535), specially for symbols that are not in a module because they are located at the beginning of the file `/proc/kallsyms`.

However, cilium/ebpf chose to return an error on duplicate symbols to avoid attaching to the wrong symbol:
https://github.com/cilium/ebpf/blob/9f201154004369b0bc1b99a06ee578aa46c6aab4/internal/kallsyms/kallsyms.go#L121-L123

We are in a different context here since we don't use our kallsyms package to attach to a kernel symbol. For the current usage of `kallsyms.SymbolExists`, it would be a good optimization. But since I am planning to offer the feature to WASM modules where I don't know in advance if it is correct to silently accept ambiguous symbols, I thought it is better to be careful. We can revisit if we need further optimizations.

pkg/kallsyms/kallsyms.go
alban: Yes on current kernel versions but I don't know if future kernels could add more fields.

pkg/kallsyms/kallsyms.go
```diff
@@ -132,6 +140,102 @@ func (k *KAllSyms) SymbolExists(symbol string) bool {
...
+		line := scanner.Bytes()
+		typeIdx := bytes.IndexByte(line, byte(' '))
+		if typeIdx == -1 || typeIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no type", line)
+		}
+		nameIdx := bytes.IndexByte(line[typeIdx+1:], byte(' '))
+		if nameIdx == -1 || typeIdx+1+nameIdx+1 >= len(line) {
+			return 0, "", fmt.Errorf("parsing line %q: no symbol", line)
+		}
+		modIdx := bytes.IndexByte(line[typeIdx+1+nameIdx+1:], byte(' '))
```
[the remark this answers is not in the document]
alban: It would be easier but that brings back the dynamic allocations and performance penalty.

pkg/kallsyms/kallsyms.go
```diff
@@ -132,6 +140,102 @@ func (k *KAllSyms) SymbolExists(symbol string) bool {
...
+		text := scanner.Text()
+		var err error
+		if modIdx == -1 {
+			// no module
+			_, err = fmt.Sscanf(text, "%x %c %s", &addr, &typ, &name)
+		} else {
+			// module found
+			_, err = fmt.Sscanf(text, "%x %c %s\t%s", &addr, &typ, &name, &mod)
+			mod = strings.Trim(mod, "[]")
+		}
```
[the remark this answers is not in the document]
alban: You're right. I thought it would be less lines of code that way, but after giving it a try, it is better your way.

pkg/kallsyms/kallsyms.go
alban: It makes sense! Updated.

pkg/kallsyms/kallsyms_test.go
```diff
@@ -62,21 +54,178 @@ func TestCustomKAllSyms(t *testing.T) {
...
+				!strings.Contains(string(kallsymsBytes), test.expectedKmod) {
+				t.Skipf("Test requires kernel module %q", test.expectedKmod)
+			}
+
+			if SymbolExists(test.symbol) != test.expectedExists {
+				t.Errorf("SymbolExists(%s) = %v, want %v",
+					test.symbol,
+					!test.expectedExists,
+					test.expectedExists)
+			}
```
mauriciovasquezbernal, line 191: It's also possible to use require.Equal for this one, If you don't want the test to fail immediately, you can use the assert package: https://pkg.go.dev/github.com/stretchr/testify/assert#Equal

mauriciovasquezbernal (approved):
Just a minor comment on a test, but LGTM already!

PR #3940, opened 2025-01-25
Explicitly set ETIMEDOUT for interrupted connection attempts fix  #3929

### LINKED ISSUE #3929: trace_tcp: error reporting on edge cases is not correct
## Description

1. When an connection is interrupted by process termination, the connect event is reported without errors.
2. When an connection is interrupted by a SIGSTOP, the connect event can be reported twice with different errors.

Initially reported partially on https://github.com/inspektor-gadget/inspektor-gadget/pull/3517#discussion_r1822990039

## Impact

1. Successful connections cannot be distinguished from connections that fail by timeout.
2. Duplicate events are confusing

### Reproduce the first case

Attempt a connection on an non-existent endpoint and terminate the process before the connection completes:

```bash
$ echo|timeout 0.5 nc 1.1.1.1 81
```
```bash
$ sudo ig run ghcr.io/inspektor-gadget/gadget/trace_tcp:latest -F proc.comm==nc --host --connect-only
RUNTIME.CONTAINER… COMM              PID        TID SRC                      DST                     TYPE      ERROR  
                   nc             602120     602120 192.168.0.22:35576       1.1.1.1:81              connect          
```

### Reproduce the second case

* Start the gadget:
```
$ sudo -E ig run ghcr.io/inspektor-gadget/gadget/trace_tcpconnect:burak-trace_tcpconnect-fix --verify-image=false
RUNTIME.CONTAINERNAME                 SRC                                               DST                                              COMM                    PID        TID ERROR         
```

* Start a container. To make it reproduce faster, change the tcp timeout of the container to approximately 7 seconds:
```
$ docker run -ti --rm --privileged busybox
/ # sysctl net.ipv4.tcp_syn_retries=2
net.ipv4.tcp_syn_retries = 2
```

* Start a tcp connection to nowhere and quickly stop the process with ctrl-z (SIGSTOP), and resume it with `fg`:
```
/ # wget 1.2.3.4
Connecting to 1.2.3.4 (1.2.3.4:80)
^Z[1]+  Stopped                    wget 1.2.3.4
/ # fg
```

* Notice the gadget output after waiting a few seconds more:
```
RUNTIME.CONTAINERNAME  SRC               DST         COMM  PID     TID ERROR
boring_edison          172.17.0.3:32798  1.2.3.4:80  wget  434444  434444 error#512
boring_edison          0.0.0.0:0         1.2.3.4:0   wget  434444  434444 ETIMEDOUT
```

There is two events (one emitted at the ctrl-z, and one emitted at the end of the tcp timeout).

## Expected behavior

1. Successful connections can be distinguished from connections interrupted before they completed.
2. No duplicate events

## Additional information

cc @burak-ok

Priyansurout:
# Improve TCP Connection Interruption Error Tracking in trace_tcp Gadget

The changes focus on modifying the handle_tcp_set_state() function to explicitly handle and report connection interruptions, particularly for scenarios involving quick timeouts or process terminations

## Testing done

input 1. echo|timeout 0.5 nc 1.1.1.1 81
           2.  nc -v google.com 443

trace_tcp


alban:
It needs `make clang-format`.

Could you add a test in `gadgets/trace_tcp/test/unit/trace_tcp_test.go` for this case?

gadgets/trace_tcp/ebpf/connect.h
```diff
@@ -100,14 +100,15 @@ static __always_inline void handle_tcp_set_state(struct pt_regs *ctx,
...
+	int err = 0;
 
 	// It would be nice if we can add a handler for TCP_SYN_SENT instead
 	// of using kprobes on tcp_v4_connect and tcp_v6_connect
 	// But when the state is set to TCP_SYN_SENT we may not have
 	// the correct source port yet
 	// https://elixir.bootlin.com/linux/v6.11.8/source/net/ipv4/tcp_ipv4.c#L301
-	if (state != TCP_ESTABLISHED && state != TCP_CLOSE)
+	if (state != TCP_ESTABLISHED && state != TCP_CLOSE &&
+	    state != TCP_SYN_SENT)
```
alban, line 111: Why is `TCP_SYN_SENT` necessary? Is the comment above still correct?

gadgets/trace_tcp/test/unit/trace_tcp_test.go
```diff
@@ -187,6 +187,39 @@ func TestTraceTcpGadget(t *testing.T) {
...
+			runnerConfig:  &utilstest.RunnerConfig{},
+			generateEvent: generateConnectionInterruptionEvent,
+			validateEvent: func(t *testing.T, info *utilstest.RunnerInfo, fd int, _ int, events []ExpectedTraceTcpEvent) error {
+				utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, pid int) *ExpectedTraceTcpEvent {
+					return &ExpectedTraceTcpEvent{
+						Proc:    info.Proc,
+						Type:    "connect",
+						NetNsId: int(info.NetworkNsID),
+						Src: utils.L4Endpoint{
+							Addr:    "0.0.0.0",
```
alban, line 204: This address does not look correct.

gadgets/trace_tcp/ebpf/connect.h
Priyansurout: Thank you for catching this issue. You're correct - including TCP_SYN_SENT was incorrect here.. I'll remove the TCP_SYN_SENT check

gadgets/trace_tcp/test/unit/trace_tcp_test.go
Priyansurout: You're right, using "0.0.0.0" as the source address isn't correct. so can i use 127.0.0.1 [localhost] ip?

alban: I am not sure. Shouldn't it be the IP of the interface used for routing to 1.2.3.4, which cannot be known statically?
In that case, you would have to use `unix.Getsockinfo` in `rawDialWithInterruption` after the connect.
But I don't know if `unix.Getsockinfo` works when the connection is not completed yet. If the kernel does not give that info, then maybe "0.0.0.0" is correct.

Priyansurout: You're right, I'll add code to get the real IP address that's being used
If we can't get the real IP, we'll use "0.0.0.0" as a backup plan

Priyansurout: Hi @alban , I'm thinking of an approach like this:

- **First**: Try to get the real source IP using `unix.Getsockname.`
- **Fallback**: If that fails (for example, if the connection isn't established), fall back to "0.0.0.0".
- **Test validation:** Use "0.0.0.0" in the test validation since that's what we'll get in most cases.

This should satisfy both requirements:
1. Try to get the real source IP when possible.
2. Use "0.0.0.0" as a fallback when we can't get the real IP.

Could you please let me know if this approach is correct, or if I should change something?

Priyansurout: Hi @alban , I'm thinking of an approach like this:

- **First**: Try to get the real source IP using `unix.Getsockname.`
- **Fallback**: If that fails (for example, if the connection isn't established), fall back to "0.0.0.0".
- **Test validation:** Use "0.0.0.0" in the test validation since that's what we'll get in most cases.

This should satisfy both requirements:
1. Try to get the real source IP when possible.
2. Use "0.0.0.0" as a fallback when we can't get the real IP.

Could you please let me know if this approach is correct, or if I should change something ?

Priyansurout: Hi @Alban, I am thinking about using` utils.NormalizedId ` to handle any source IP address

PR #3947, opened 2025-01-27
ADOPTERS.md: Fixes contributing link for retina.sh.

eiffel-fl:


flyth (approved):
LGTM as the old url is 404 - but I'm not sure if that page is a proper target for "Microsoft".

Issue #3705, opened 2024-11-15
DNS tracer is increasing CPU load on CoreDNS pods


matthyx:
## Description

DNS tracer doesn't filter on container namespace and therefore parses all DNS packets happening on a Kubernetes node.

## Impact

Our agent and colocated CoreDNS pods see an increased CPU consumption when using DNS tracers. Excluding CoreDNS containers doesn't change the situation.

## Environment and steps to reproduce

1. **Set-up**:
  a. start a kubernetes cluster (kind is fine)
  b. helm repo add coredns https://coredns.github.io/helm
  c. clone https://github.com/northflank-guides/dns-performance-tests.git
  d. add label `coredns-test: coredns` to a node
  e. helm --namespace=coredns-test install coredns coredns/coredns --create-namespace -f dns-performance-tests/coredns-values.yaml
  f. cd dns-performance-tests/kubernetes-client-script/
  g. npm install
  h. node index.js
2. **Task**:
  a. start dnsperf following https://github.com/northflank-guides/dns-performance-tests/tree/main?tab=readme-ov-file#dnsperf
  b. fix image in deployment using `guessi/dnsperf`
  c. reduce dnsperf `MAX_QPS` to make sure container isn't CPU throttled
4. **Action(s)**:
  a. check coredns CPU
  b. start DNS tracer and limit to a NS `kubectl gadget trace dns -n default`
5. **Error**:
  a. check core CPU increase by 10-20%

## Expected behavior

Excluding CoreDNS containers should be working and reducing CPU load.

## Additional information

Slack discussion: https://kubernetes.slack.com/archives/CSYL75LF6/p1720705023758379
WIP PR: https://github.com/inspektor-gadget/inspektor-gadget/pull/3159

mqasimsarfraz:
@matthyx I am trying to reproduce the issue and wanted to clarify if you are running any workloads in the `default` namespace? The reason I am asking is, if I have a pod with `HostNetwork=true` in the namespace we want to trace I see much more spike in CoreDNS cpu (`700m` -> `950m`) compared to (`700m` -> `760m`) if we don't have anything installed.

matthyx:
I didn't have anything installed for the reproducer, but in real clusters there's and indeed the spikes are higher

matthyx:
@mauriciovasquezbernal did this slide out of the v0.36.0 milestone?

mqasimsarfraz:
[...]

PR #3951, opened 2025-01-27
gadgets: Uniform filtering for networking gadgets

### LINKED ISSUE #706: gadgets: Uniform filtering feature
We can filter even in our gadgets by using different parameters, however it's not uniform and not all gadgets support all filtering options. The goal of this issue is to uniform all gadgets to be able to filter by the following parameters:

- mount ns id: filter events by container 
- pid: filter by process
- uid: filter by user 

Additionally, we should consider if network gadgets should allow to filter by:
- family: ipv4, ipv6
- protocol: udp, tcp
- destination/source ips?
- destination/source ports?

ref https://github.com/kinvolk/inspektor-gadget/pull/692#discussion_r899146245

mauriciovasquezbernal:
Add a new gadget_should_discard_data_by_skb() helper and use it from different neworking gadgets.

## Testing 

Try the different filtering options on the trace_dns and trace_sni gadgets:

```bash 
$ docker run -it --rm busybox sh -c "while true; do nslookup google.com; sleep 1; wget google.com; done"
```

```bash
$ sudo ig run trace_dns --verify-image=false --comm wget
RUNTIME.CONTA… SRC                    DST                    COMM         PID     TID QR QTYPE   NAME          RCODE    ADDRES… LATENCY…
stoic_shamir   172.17.0.2:53950       1.1.1.1:53             wget      195991  195991 Q  A       google.com.                         0ns
stoic_shamir   172.17.0.2:53950       1.1.1.1:53             wget      195991  195991 Q  AAAA    google.com.                         0ns
stoic_shamir   1.1.1.1:53             172.17.0.2:53950       wget      195991  195991 R  AAAA    google.com.   Success  2607:f…  63.99ms
stoic_shamir   1.1.1.1:53             172.17.0.2:53950       wget      195991  195991 R  A       google.com.   Success  142.25…  64.09ms
stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 Q  PTR     _nfs._tcp.lo…                       0ns
stoic_shamir   172.17.0.2:52925       1.1.1.1:53             wget      195991  195991 Q  A       www.google.c…                       0ns
stoic_shamir   172.17.0.2:52925       1.1.1.1:53             wget      195991  195991 Q  AAAA    www.google.c…                       0ns
stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 Q  ALL     5.0.f.e.3.3.…                       0ns
stoic_shamir   1.1.1.1:53             172.17.0.2:52925       wget      195991  195991 R  A       www.google.c… Success  142.25…  68.08ms
stoic_shamir   1.1.1.1:53             172.17.0.2:52925       wget      195991  195991 R  AAAA    www.google.c… Success  2607:f…  68.04ms
stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 Q  ALL     5.0.f.e.3.3.…                       0ns
stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 Q  ALL     5.0.f.e.3.3.…                       0ns
stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 R                        Success  fe80::…      0ns
stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 Q  PTR     _nfs._tcp.lo…                       0ns
stoic_shamir   172.17.0.2:57324       1.1.1.1:53             wget      195997  195997 Q  A       google.com.                         0n
```

Fixes #706


include/gadget/mntns_filter.h
```diff
@@ -3,7 +3,14 @@
 #ifndef MNTNS_FILTER_H
 #define MNTNS_FILTER_H
 
+#ifndef GADGET_TYPE_NETWORKING
+// gadget/mntns.h expects the gadget to include vmlinux.h since it uses struct
+// task_struct, this is an issue for gadgets that don't include vmlinux.h, like
+// networking gadgets. Ideally this file should not be included here, but
+// removing it could break a lot of gadgets, so let's use this hack to allow
+// this working for networking gadgets.
 #include <gadget/mntns.h>
```
alban, line 12: Could the `#ifndef GADGET_TYPE_NETWORKING` test be moved to `gadget/mntns.h` instead?

gadgets/trace_dns/program.bpf.c
```diff
@@ -328,6 +329,8 @@ int ig_trace_dns(struct __sk_buff *skb)
 
 	// Enrich event with process metadata
 	struct sockets_value *skb_val = gadget_socket_lookup(skb);
+	if (gadget_should_discard_data_by_skb(skb_val))
+		return 0;
```
alban, line 333: A DNS query could be accepted and the DNS answer could be discarded if the UDP socket was closed before receiving the answer and re-created with a different pid or different comm.

Then, the entries in query_map might not get cleaned up.

But this problem existed before your PR and the image-based dns gadget should do the garbage collection in the WASM module, like it was done for the builtin gadget in pkg/gadgets/trace/dns/tracer/gc.go.

include/gadget/mntns_filter.h
mauriciovasquezbernal: I could do that, by IMO it's confusing to import a file has all its contents under a  `#ifndef`. My initial thought when implementing this file some time ago was: if anybody is using mntns_filter.h, then they'll need to use mntns.h to get the mount ns, so let's include one in another to avoid the user having to include both of them. But it turns out that on networking gadgets it's not true, as the mount ns is coming from the socket enricher.

gadgets/trace_dns/program.bpf.c
mauriciovasquezbernal: I see, so there is not any action on this PR, right? I'll create an issue about that after this PR is done.

include/gadget/mntns_filter.h
alban: Ok.

But we don't have a lot of out-of-tree gadgets, so IMO it would be fine as well to update our in-tree gadgets and the documentation to include mntns.h.

gadgets/trace_dns/program.bpf.c
alban: Correct. I just wrote it down not to forget.

(I have not finished reviewing the PR)

include/gadget/mntns_filter.h
mauriciovasquezbernal: Handled in the last push.

gadgets/trace_sni/program.bpf.c
```diff
@@ -220,9 +220,11 @@ int ig_trace_sni(struct __sk_buff *skb)
 	}
 	event.timestamp_raw = bpf_ktime_get_boot_ns();
 
+	struct sockets_value *skb_val = gadget_socket_lookup(skb);
 	// Enrich event with process metadata
```
burak-ok, line 224: `Comment should stay right above `gadget_process_populate_from_socket`

gadgets/trace_sni/program.bpf.c
```diff
@@ -220,9 +220,11 @@ int ig_trace_sni(struct __sk_buff *skb)
...
-	gadget_process_populate_from_socket(gadget_socket_lookup(skb),
-					    &event.proc);
+	if (gadget_should_discard_data_by_skb(skb_val))
```
burak-ok, line 225: Move this check right to the beginning of this function to filter packets early

burak-ok (commented):
The 2 review comments are also the same for `trace_dns`.

burak-ok:
> ```shell
> $ sudo ig run trace_dns --verify-image=false --comm wget
> WARN[0000] image signature verification is disabled due to using corresponding option
> WARN[0001] image signature verification is disabled due to using corresponding option
> RUNTIME.CONTA… SRC                    DST                    COMM         PID     TID QR QTYPE   NAME          RCODE    ADDRES… LATENCY…
> ....
> stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 R                        Success  fe80::…      0ns
> stoic_shamir   fe80::6c7d:9dff:fe33:… ff02::fb:5353                         0       0 Q  PTR     _nfs._tcp.lo…                       0ns
> stoic_shamir   172.17.0.2:57324       1.1.1.1:53             wget      195997  195997 Q  A       google.com.                         0n
> ```

So this filtered by `--comm wget`, but in the output are still empty comm lines

gadgets/trace_sni/program.bpf.c
mauriciovasquezbernal: I'm not totally sure this should go to the very beginning as this is a heavy operation and other filtering logic on these gadgets are faster. I moved it a bit above on this gadget and left it on the same position on trace_dns.

mauriciovasquezbernal:
> So this filtered by `--comm wget`, but in the output are still empty comm lines

That's a good question. The current implementation avoid to filter out packets without process information. I don't have a strong opinion, do you think we should skip them as well?

alban (commented):

| Benchmark suite | Current: 1dda95d8540abea75a542094f85d84d02b94bf1c | Previous: 8d38183ed2dd3424d100881e5a4e8d60f9ab31f4 | Ratio |
|-|-|-|-|
| `BenchmarkAllGadgetsWithContainers/container0/trace-dns` | `21166317` ns/op | `6310476` ns/op | `3.35` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-network` | `19462668` ns/op | `4617489` ns/op | `4.21` |

CC: @alban

include/gadget/types.h
```diff
@@ -3,6 +3,12 @@
 #ifndef __TYPES_H
 #define __TYPES_H
 
+// Networking gadgets don't include vmlinux.h, so we need to define few types
+// needed by other headers.
+#ifdef GADGET_TYPE_NETWORKING
+#include <stdbool.h>
+#endif
```
alban, line 10: Why do you need `stdbool.h`?

In order to fix #2371, I would like to be possible for gadget authors to write networking gadgets that include `vmlinux.h`. So I prefer not to rely on `#ifdef GADGET_TYPE_NETWORKING` to detect the presence of `vmlinux.h`. Instead, I suggest to use the same code pattern as in `sockets-map.h`:

```c
// Types such as __u64 below requires to include either <linux/types.h> or
// <vmlinux.h> before. We can't include both because they are incompatible. Let
// the gadget choose which one to include.
#if !defined(__VMLINUX_H__) && !defined(_LINUX_TYPES_H)
#error "Include <linux/types.h> or <vmlinux.h> before including this file."
#endif
```

mauriciovasquezbernal: > Why do you need stdbool.h?

To get a definition of bool that we're using in a lot of places. 

I changed the approach, is it fine now?

mauriciovasquezbernal:
> Apart from my comment about `#ifdef GADGET_TYPE_NETWORKING`, this looks good!

Thanks Alban. Any comments about https://github.com/inspektor-gadget/inspektor-gadget/pull/3951#issuecomment-2665647574?

alban:
> > Apart from my comment about `#ifdef GADGET_TYPE_NETWORKING`, this looks good!
> 
> Thanks Alban. Any comments about https://github.com/inspektor-gadget/inspektor-gadget/pull/3951#issuecomment-2665647574?

I think they should be filtered out.

burak-ok:
> > So this filtered by `--comm wget`, but in the output are still empty comm lines
> 
> That's a good question. The current implementation avoids to filter out packets without process information. I don't have a strong opinion, do you think we should skip them as well?

IMO when I want to filter something, I really only want to get these specific lines. But this seams like an issue for another PR

mauriciovasquezbernal:
I updated the logic to drop the events when any filter is configured but not socket enricher information is available.

include/gadget/filter.h
```diff
@@ -119,5 +123,24 @@ static __always_inline bool gadget_should_discard_data_current()
...
+#endif
+
+// gadget_should_discard_data_by_skb returns true if the gadget should skip this
+// event based on the fields of the skb_val struct.
+static __always_inline bool
+gadget_should_discard_data_by_skb(struct sockets_value *skb_val)
+{
+	if (!skb_val) {
+		// drop the packet if any filtering is specified
+		if (gadget_filter_by_mntns || targ_pid != 0 || targ_tid != 0 ||
```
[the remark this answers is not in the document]
mauriciovasquezbernal: > what happen to request at **host** level

> ./kubectl-gadget run ghcr.io/mqasimsarfraz/gadget/trace_dns:mauricio-uniform-filtering-networking **-A**

hmmm, I thought `-A` was for showing the event in all namespaces, and that we don't support showing events on the host when using kubectl-gadget. I remember there was a quite long discussion about it some time ago but I can't find any github link with it. 

IMO the behavior that you're seeing above is expected and the previous one was a bug you're using as a feature :rofl:. 

I could revert this not to drop events when skb_val is nil to avoid affecting that behavior.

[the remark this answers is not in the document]
mqasimsarfraz: > IMO the behavior that you're seeing above is expected and the previous one was a bug you're using as a feature 🤣.

It also felt like using a bug as a feature in first place as well! :smile: 

> I could revert this not to drop events when skb_val is nil to avoid affecting that behavior.

I actually agree that default behavior should be to drop events if skb_val is nil. We are using too much CPU cycles to trace host packets if CoreDNS is on the same node ([3rd Optimization here](https://github.com/inspektor-gadget/inspektor-gadget/issues/3705#issuecomment-2607603349)). My initial idea was to have some way e.g `--host/--trace-all` param to include additional events if needed but I am fine delaying it till we have use-case?

include/gadget/filter.h
```diff
@@ -119,5 +123,24 @@ static __always_inline bool gadget_should_discard_data_current()
...
+		    targ_uid != GADGET_INVALID_ID ||
+		    targ_gid != GADGET_INVALID_ID || targ_comm[0] != '\0')
+			return true;
+	}
+
+	return gadget_should_discard_data(
+		skb_val->mntns, skb_val->pid_tgid >> 32, skb_val->pid_tgid,
+		skb_val->task, skb_val->uid_gid, skb_val->uid_gid >> 32);
```
burak-ok, line 143: Comment not for this PR:
We should maybe create another function where we could just give `pid_tgid` and `uid_gid`. These shifts are easy to mess up imo

burak-ok (approved):
Everything looks good now from my side. Thanks :)

include/gadget/types.h
alban: Your change looks good to me.

alban (approved):

include/gadget/filter.h
mauriciovasquezbernal: Perhaps create some helpers to extract those fields is better? I suppose there is not any overhead on doing something like `gadget_should_discard_data(..., gadget_get_pid(pid_tdig), gadget_get_tid(pid_tgid)`, probably they can be just macros too.

include/gadget/filter.h
```diff
@@ -119,5 +123,24 @@ static __always_inline bool gadget_should_discard_data_current()
...
+	if (!skb_val) {
+		// drop the packet if any filtering is specified
+		if (gadget_filter_by_mntns || targ_pid != 0 || targ_tid != 0 ||
+		    targ_uid != GADGET_INVALID_ID ||
+		    targ_gid != GADGET_INVALID_ID || targ_comm[0] != '\0')
+			return true;
+	}
+
+	return gadget_should_discard_data(
+		skb_val->mntns, skb_val->pid_tgid >> 32, skb_val->pid_tgid,
```
alban, line 142: `skb_val` can be NULL here. This causes the dns and sni gadgets to fail:

```
$ sudo ./ig run ghcr.io/inspektor-gadget/gadget/trace_dns:main --host -v --verify-image=false
...
	; skb_val->mntns, skb_val->pid_tgid >> 32, skb_val->pid_tgid,
	470: (79) r7 = *(u64 *)(r8 +8)
	R8 invalid mem access 'scalar'
	processed 367 insns (limit 1000000) max_states_per_insn 1 total_states 30 peak_states 30 mark_read 13 
```

PR #3953, opened 2025-01-28
Added node name to slim pod

afek854:
# [Title: Added node name to slim pod]


alban:
The patch looks fine but could you explain the use case for this addition?

afek854:
> The patch looks fine but could you explain the use case for this addition?

Kubescape uses this cache type to enrich IP to pod in the cluster but we also need node name for the enrichment.

alban (approved):
LGTM

@burak-ok good for you as well?

burak-ok (approved):
LGTM too. Thanks :)

PR #3955, opened 2025-01-28
perf/ring buffer: Continue reading in case of errors

mqasimsarfraz:
As discussed in https://github.com/cilium/ebpf/issues/1633 we should continue reading the perf ring buffer in case of errors. Some examples from cilium/ebpf repository:
- https://github.com/cilium/ebpf/blob/main/examples/uretprobe/main.go#L96-L97
- https://github.com/cilium/ebpf/blob/9f201154004369b0bc1b99a06ee578aa46c6aab4/examples/fentry/main.go#L89-L90

### DIFF
--- a/pkg/operators/ebpf/tracer.go
+++ b/pkg/operators/ebpf/tracer.go
@@ -15,6 +15,7 @@
 package ebpfoperator
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -130,7 +131,11 @@ func (t *Tracer) receiveEventsFromRingReader(gadgetCtx operators.GadgetContext) 
 	for {
 		rec, err := t.ringbufReader.Read()
 		if err != nil {
-			return err
+			if errors.Is(err, ringbuf.ErrClosed) {
+				return err
+			}
+			gadgetCtx.Logger().Warnf("error reading ringbuf event: %v", err)
+			continue
 		}
 		pSingle, err := t.ds.NewPacketSingle()
 		if err != nil {
@@ -170,7 +175,11 @@ func (t *Tracer) receiveEventsFromPerfReader(gadgetCtx operators.GadgetContext) 
 	for {
 		rec, err := t.perfReader.Read()
 		if err != nil {
-			return err
+			if errors.Is(err, perf.ErrClosed) {
+				return nil
+			}
+			gadgetCtx.Logger().Warnf("error reading perf event: %v", err)
+			continue
 		}
 		pSingle, err := t.ds.NewPacketSingle()
 		if err != nil {
--- a/pkg/gadgets/audit/seccomp/tracer/tracer.go
+++ b/pkg/gadgets/audit/seccomp/tracer/tracer.go
@@ -108,9 +108,9 @@ func (t *Tracer) run() {
 				return
 			}
 
-			msg := fmt.Sprintf("Error reading perf ring buffer: %s", err)
-			t.eventCallback(types.Base(eventtypes.Err(msg)))
-			return
+			msg := fmt.Sprintf("reading perf ring buffer: %s", err)
+			t.eventCallback(types.Base(eventtypes.Warn(msg)))
+			continue
 		}
 
 		if record.LostSamples > 0 {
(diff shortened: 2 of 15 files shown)


alban (approved):
LGTM from code review.
Thanks!

Have you tried to reproduce the issue with the C program from https://github.com/cilium/ebpf/issues/1633 and reverting https://github.com/inspektor-gadget/inspektor-gadget/pull/3806?

pkg/operators/ebpf/tracer.go
```diff
@@ -170,7 +175,11 @@ func (t *Tracer) receiveEventsFromPerfReader(gadgetCtx operators.GadgetContext)
 	for {
 		rec, err := t.perfReader.Read()
 		if err != nil {
-			return err
+			if errors.Is(err, perf.ErrClosed) {
+				return nil
+			}
+			gadgetCtx.Logger().Warnf("error reading perf event: %v", err)
```
flyth, line 181: How often/when does this actually happen? Maybe we should just increase the drop counter instead of sending a warning?

mqasimsarfraz:
> Have you tried to reproduce the issue with the C program from https://github.com/cilium/ebpf/issues/1633 and reverting https://github.com/inspektor-gadget/inspektor-gadget/pull/3806?

I just tested it using C program and we continue to get new events after read failures:

### reverted https://github.com/inspektor-gadget/inspektor-gadget/pull/3806 and with PR changes
```
$ go run -exec sudo ./cmd/ig trace dns
# ./maxudp
# nslookup google.com.
RUNTIME.CONTAINERNAME                                                                                                         PID        TID        PPID       COMM             PCOMM            QR TYPE      QTYPE      NAME                                                                                                                          RCODE                                                              NUMANSW…
WARN[0020] reading perf ring buffer: read sample: unexpected EOF 
d0f5d334e329c7bd7ed0c250ecd37b81845b3080278d546c3940369605bdaffc                                                              0          0          0                                            Q  MULTICAST PTR        _ipps._tcp.local.                                                                                                                                                                                0       
d0f5d334e329c7bd7ed0c250ecd37b81845b3080278d546c3940369605bdaffc                                                              49150      49150      38106      nslookup         sh               Q  OUTGOING  A          google.com.                                                                                                                                                                                      0       
d0f5d334e329c7bd7ed0c250ecd37b81845b3080278d546c3940369605bdaffc                                                              49150      49150      38106      nslookup         sh               Q  OUTGOING  AAAA       google.com.                                                                                                                                                                                      0       
d0f5d334e329c7bd7ed0c250ecd37b81845b3080278d546c3940369605bdaffc                                                              49150      49150      38106      nslookup         sh               R  HOST      A          google.com.                                                                                                                   No Error                                                           1       
d0f5d334e329c7bd7ed0c250ecd37b81845b3080278d546c3940369605bdaffc                                                              49150      49150      38106      nslookup         sh               R  HOST      AAAA       google.com.                                                                                                                   No Error                                                           1  
```

pkg/operators/ebpf/tracer.go
```diff
@@ -130,7 +131,11 @@ func (t *Tracer) receiveEventsFromRingReader(gadgetCtx operators.GadgetContext)
 	for {
 		rec, err := t.ringbufReader.Read()
 		if err != nil {
-			return err
+			if errors.Is(err, ringbuf.ErrClosed) {
+				return err
+			}
+			gadgetCtx.Logger().Warnf("error reading ringbuf event: %v", err)
+			continue
```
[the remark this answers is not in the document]
mqasimsarfraz: It seems to be the case for both. I took this code from here: https://github.com/cilium/ebpf/blob/9f201154004369b0bc1b99a06ee578aa46c6aab4/examples/fentry/main.go#L85-L90

pkg/operators/ebpf/tracer.go
mqasimsarfraz: I don't expect it to happen to too often otherwise we would have cases where gadgets should have stopped working. But I might be wrong. @alban do you have any thoughts?

alban: This specific PR is not about lost events because ig can't keep up, but about a bug when the gadget generates bad events (see details in https://github.com/cilium/ebpf/issues/1633). So I think the patch is good.

Issue #3936, opened 2025-01-24
trace_dns: Getting empty events under high load


mqasimsarfraz:
`trace_dns` seem to be getting empty events under high load with an error message like:

```
$ ./kubectl-gadget run trace_dns -p mypod
...
WARN[0009] minikube-docker      | error emitting data: getting IP address: getting IP: unknown IP version: 0 
..
```

## Impact

It seems we are losing certain events in high load.

### ig-k8s
- Install Inspektor Gadget in a minikube cluster.
- Start DNS gadget:
```
kubectl gadget run trace_dns -p mypod
```
- Create a pod:
```
kubectl -n default run mypod -it --image=wbitt/network-multitool -- /bin/sh
```
- Perform parallel requests:
```
seq 1 50 | xargs -n1 -P50 -I{} nslookup kubernetes.default.svc.cluster.local.
```
- The trace should print warnings `WARN[0009] minikube-docker      | error emitting data: getting IP address: getting IP: unknown IP version: 0 `

### ig

```
$ go run -exec sudo ./cmd/ig run trace_dns --verify-image=false --host -F "name==example.com."
# running "seq 1 50 | xargs -n1 -P50 -I{} nslookup -type=a example.com." in another terminal
...
      127.0.0.53:53          127.0.0.1:45189        is…   18281   18329 R  A  exa… Success  2… 15
      127.0.0.1:52769        127.0.0.53:53          is…   18363   18418 Q  A  exa…             0n
WARN[0035] error emitting data: getting IP address: getting IP: unknown IP version: 0 
      127.0.0.1:46447        127.0.0.53:53          is…   18406   18470 Q  A  exa…             0n
      127.0.0.1:46447        127.0.0.53:53          sy…     987     987 Q  A  exa…             0n
      127.0.0.53:53          127.0.0.1:46447        sy…     987     987 R  A  exa… Success  2… 0n
      127.0.0.53:53          127.0.0.1:46447        is…   18406   18470 R  A  exa… Success  2… 19
      127.0.0.53:53          127.0.0.1:33564        sy…     987     987 R  A  exa… Success  9… 0n
WARN[0039] error emitting data: getting IP address: getting IP: unknown IP version: 0 
WARN[0039] error emitting data: getting IP address: getting IP: unknown IP version: 0 
WARN[0087] error emitting data: getting IP address: getting IP: unknown IP version: 0 
...
```

## Additional Information:
[...]

PR #3956, opened 2025-01-28
ebpf/tracer: Report lost samples

mqasimsarfraz:
It seems we aren't handling lost samples correctly and under high load, we are getting empty samples: https://github.com/cilium/ebpf/blob/9f201154004369b0bc1b99a06ee578aa46c6aab4/perf/reader.go#L78 resulting in error messages like https://github.com/inspektor-gadget/inspektor-gadget/issues/3936. This change reports the lost samples and continue.

Related: https://github.com/inspektor-gadget/inspektor-gadget/issues/3936

### Follow-up

During testing I realized if we have slow subscriber call backs (e.g `wasm`) we can easily end up in a situation where we don't read perf buffer fast enough, resulting in samples being dropped. Is there anything can be done about it?

### DIFF
--- a/pkg/operators/ebpf/tracer.go
+++ b/pkg/operators/ebpf/tracer.go
@@ -172,6 +172,14 @@ func (t *Tracer) receiveEventsFromPerfReader(gadgetCtx operators.GadgetContext) 
 		if err != nil {
 			return err
 		}
+		if rec.LostSamples > 0 {
+			gadgetCtx.Logger().Warnf("reading perf ring buffer: lost %d samples", rec.LostSamples)
+			t.ds.ReportLostData(rec.LostSamples)
+			// if we lose samples, we don't get the raw sample,
+			// so we shouldn't emit anything
+			continue
+		}
+
 		pSingle, err := t.ds.NewPacketSingle()
 		if err != nil {
 			gadgetCtx.Logger().Warnf("error creating new packet: %v", err)
@@ -204,9 +212,6 @@ func (t *Tracer) receiveEventsFromPerfReader(gadgetCtx operators.GadgetContext) 
 		err = t.ds.EmitAndRelease(pSingle)
 		if err != nil {
 			gadgetCtx.Logger().Warnf("error emitting data: %v", err)
-		}
-		if rec.LostSamples > 0 {
-			t.ds.ReportLostData(rec.LostSamples)
 		}
 	}
 }


alban (approved):
LGTM, thanks!

This is the same code pattern as in:
https://github.com/cilium/ebpf/blob/9f201154004369b0bc1b99a06ee578aa46c6aab4/examples/uretprobe/main.go#L100
It reads `LostSamples` before reading `RawSample`.

It will conflict with #3955. I guess you want to merge #3955 first?

pkg/operators/ebpf/tracer.go
```diff
@@ -172,6 +172,14 @@ func (t *Tracer) receiveEventsFromPerfReader(gadgetCtx operators.GadgetContext)
 		if err != nil {
 			return err
 		}
+		if rec.LostSamples > 0 {
+			gadgetCtx.Logger().Warnf("reading perf ring buffer: lost %d samples", rec.LostSamples)
+			t.ds.ReportLostData(rec.LostSamples)
+			// if we lose samples, we don't get the raw sample,
+			// so we shouldn't emit anything
+			continue
+		}
```
flyth, line 181: Is this really the case? I was under the impression that we get the lost samples with the next valid record - but I might be wrong.

Additionally, doing a `gadgetCtx.Logger().Warnf()` is potentially dangerous here: if the sample loss is caused by IG not keeping up with the load, logging will just worsen the situation (it will serialize and send a log packet).

alban: > Is this really the case? I was under the impression that we get the lost sample count with the next valid record - but I might be wrong.

From the cilium/ebpf link posted by Qasim, when `LostSamples > 0`, `RawSample` is always empty.

> Additionally, doing a `gadgetCtx.Logger().Warnf()` is potentially dangerous here: if the sample loss is caused by IG not keeping up with the load, logging will just worsen the situation

The possibility of sample loss cannot be removed completely, so I think it is better to warn the user.

`ReportLostData()` currently does nothing.

>  (it will serialize and send a log packet).

Is it done synchronously?

flyth: > From the cilium/ebpf link posted by Qasim, when LostSamples > 0, RawSample is always empty.

👍🏻 

> The possibility of sample loss cannot be removed completely, so I think it is better to warn the user. ReportLostData() currently does nothing.

I don't fear sample loss in general, I fear that it's a situation that is caused by IG (not being able to keep up with the load) - and that by increasing the load (sending a static warning as string over the network for each loss notice), it would worsen. We could instead increase the packet seq by the loss amount (that would trigger a message on the client iirc, but would make it indistinguishable whether the loss is cause by ebpf or the network buffer) or at least limit how often the warning gets sent.

> Is it done synchronously?

Serializing yes, sending no (with buffer).

mqasimsarfraz:
> It will conflict with https://github.com/inspektor-gadget/inspektor-gadget/pull/3955. I guess you want to merge https://github.com/inspektor-gadget/inspektor-gadget/pull/3955 first?

Yes, I plan to merge https://github.com/inspektor-gadget/inspektor-gadget/pull/3955 and then will update this one! 

> Is this really the case? I was under the impression that we get the lost sample count with the next valid record - but I might be wrong.

Yes, we don't get any raw samples with lost samples. I tested it with:

```
		if rec.LostSamples > 0 {
			gadgetCtx.Logger().Warnf("reading perf ring buffer: lost %d (%d) samples", rec.LostSamples, len(rec.RawSample))
			t.ds.ReportLostData(rec.LostSamples)
			// if we lose samples, we don't get the raw sample,
			// so we shouldn't emit anything
			//continue
		}
```

and got:

```
WARN[0011] reading perf ring buffer: lost 28 (0) samples 
WARN[0011] error emitting data: getting IP address: getting IP: unknown IP version: 0 
```

pkg/operators/ebpf/tracer.go
mauriciovasquezbernal: > least limit how often the warning gets sent.

Time for rate limited logging support?

mauriciovasquezbernal: Yeah, I agree. I see this PR on its current state if fixing a bug, so I won't block on the logging issue. If we see a lot of logs later on, we can limit it.

mauriciovasquezbernal (approved):
IMO it's good to go. Thanks for fixing this!

mqasimsarfraz:
Thanks for the reviews!

mqasimsarfraz:
> ## Follow-up
> During testing I realized if we have slow subscriber call backs (e.g wasm) we can easily end up in a situation where we don't read > perf buffer fast enough, resulting in samples being dropped. Is there anything can be done about it?

I wanted to start a conversation around it it because dropping events around 50 QPS isn't great. I was doing some manual testing I see following number [with wasm patch](https://gist.github.com/mqasimsarfraz/891dfdfdb96abca53d51093e526701e9) and [w/o wasm patch](https://gist.github.com/mqasimsarfraz/61eb70e39f9edac66e2746d8421640c1): 


### with wasm
```
RUNTIME.CONTAINERNAME                                                                                                  SRC                                                  DST                                                  COMM                    PID        TID                                                                     LATENCY_NS
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981 172.17.0.2:60583                                     192.168.0.1:53                                       nslookup              29085      29085 Q    A                          google.com.                                                                                                       0ns
INFO[0039] perf event processing time: 962.727µs        
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981 192.168.0.1:53                                       172.17.0.2:60583                                     nslookup              29085      29085 R    A                          google.com.                                    Success      142.250.181.206                                    5.12ms
INFO[0039] perf event processing time: 45.23344ms       
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981 172.17.0.2:44877                                     192.168.0.1:53                                       nslookup              29086      29086 Q    A                          google.com.                                                                                                       0ns
INFO[0040] perf event processing time: 1.255857ms       
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981 192.168.0.1:53                                       172.17.0.2:44877                                     nslookup              29086      29086 R    A                          google.com.                                    Success      142.250.181.206                                    3.06ms
INFO[0040] perf event processing time: 1.326021ms       
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981 172.17.0.2:45715                                     192.168.0.1:53                                       nslookup              29087      29087 Q    A                          google.com.                                                                                                       0ns

```

### w/o wasm
```
RUNTIME.CONTAINERNAME                                                                                                  SRC                                                  DST                                                  COMM                    PID        TID                                                                     LATENCY_NS
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981924598bcbfc38d251                                                       172.17.0.2:45993                                     192.168.0.1:53                                       nslookup              29411      29411                                                                            0ns
INFO[0013] perf event processing time: 242.792µs        
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981924598bcbfc38d251                                                       192.168.0.1:53                                       172.17.0.2:45993                                     nslookup              29411      29411                                                                         4.31ms
INFO[0013] perf event processing time: 141.404µs        
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981924598bcbfc38d251                                                       172.17.0.2:57836                                     192.168.0.1:53                                       nslookup              29414      29414                                                                            0ns
INFO[0015] perf event processing time: 208.833µs        
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981924598bcbfc38d251                                                       192.168.0.1:53                                       172.17.0.2:57836                                     nslookup              29414      29414                                                                         5.17ms
INFO[0015] perf event processing time: 144.559µs        
4615162ee5d036c1d73327e3f3992d152a0c0b385e0e981924598bcbfc38d251                                                       172.17.0.2:60216                                     192.168.0.1:53                                       nslookup              29415      29415                                                                            0ns

```

Perhaps we need to understand why it takes `45ms` to `EmitAndRelease` sometime. Any thoughts? Am I missing something?

alban:
> Perhaps we need to understand why it takes 45ms to EmitAndRelease sometime. Any thoughts? Am I missing something?

Could you create an issue about this? It could be similar to #3541 but for cpu performance.

I also see that EmitAndRelease takes the most time:
image

Steps to reproduce:

<details>

```
sudo ./ig run trace_dns --host --pprof-addr localhost:6060
docker run -ti --rm --name=dnstester ghcr.io/inspektor-gadget/dnstester
docker exec -ti dnstester /bin/sh -c 'while true ; do nslookup example.com 127.0.0.1 >/dev/null ; done'
go tool pprof -http=:8080 http://127.0.0.1:6060/debug/pprof/profile
```

</details>

Issue #3560, opened 2024-10-08
[RFE] Implement event generation


mqasimsarfraz:
The issue describes the user experience and implementation details for event generation:

## User experience (UX)

The event generation will be tied to a single gadget run and controlled using operator's parameters i.e., CLI flags or paramValues in https://github.com/inspektor-gadget/inspektor-gadget/pull/3515. Following are the examples:

### flags

 ```bash
$ ig run trace_dns --filter name==example.com. --eventgen-enable=true --eventgen-type=dns --eventgen-target=example.com
```
or
```
$ kubectl gadget run --filter name==example.com. --eventgen-enable=true --eventgen-type=dns --eventgen-target=example.com
```

or using instance manifest via:

```yaml
version: 1
kind: instance-spec
image: trace_dns
name: capture_dns_events
paramValues:
    operator.filter.filter: "name==example.com."
    operator.eventgen.eventgen-enable: true
    operator.eventgen.eventgen-type: dns
    operator.eventgen.eventgen-target: example.com
```
### Implementation details:

In terms of implementation it will be implemented as an [operator](https://www.inspektor-gadget.io/docs/latest/spec/operators/) and:
- The generation of events must happen on server side except for Kubernetes (see below).
- The environment where events will be generated shouldn't be tight to Kubernetes (probably what you did was just an example). I suggest re-using the [Container interface](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/container-utils/testutils/container_interface.go#L39) we created for testing. With few changes, it'll allow us to run commands on docker/containerd containers (even though in this case we don't care about the runtime), but also in Kubernetes containers by keeping a common interface. In addition, we can extend it to allow creating container with specific linux namespaces. For instance, to create container with the host network linux namespace for generating DNS or HTTPS events for ig running as standalone with the --host flag.

#### Params:
Currently I think following params should be there:
- [Global] eventgen-enable - To allow admins to disable/enable event generation.
- [Instance] eventgen-type - If specified should start event generation based on the type specified e.g DNS,HTTP,etc.
- [Instance] eventgen-target - Domain name, file name, etc
- [Instance] eventgen-count - Number of events to generate, default "-1" to keep generating events in loop. 
- [Instance] eventgen-extraparams - Extra params to control different event generators e.g DNS record type, Number of retries, etc (maybe this could be part of target?)

mqasimsarfraz:
[...]

PR #3957, opened 2025-01-28
operators: localmanager: Use datasource to attach and detach con…

eiffel-fl:
…tainers.

Based on:
https://github.com/inspektor-gadget/inspektor-gadget/commit/4a571255ceec9723c91bc59ad29533b7931b8d5a

Tested in: 
https://github.com/inspektor-gadget/inspektor-gadget/commit/8799d8fc938843efa1ca8d7d2c277ad90c0b8f40

### DIFF
--- a/pkg/operators/localmanager/localmanager.go
+++ b/pkg/operators/localmanager/localmanager.go
@@ -26,6 +26,7 @@ import (
 	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
 
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
@@ -319,10 +320,37 @@ type localManagerTrace struct {
 	gadgetCtx          operators.GadgetContext
 
 	eventWrappers map[datasource.DataSource]*compat.EventWrapperBase
+
+	containersDs              datasource.DataSource
+	eventTypeField            datasource.FieldAccessor
+	cgroupIDField             datasource.FieldAccessor
+	mountNsIDField            datasource.FieldAccessor
+	nameField                 datasource.FieldAccessor
+	containersSubscriptionKey string
 }
 
 func (l *localManagerTrace) Name() string {
 	return OperatorName
+}
+
+func (l *localManagerTrace) emitContainersDatasourceEvent(eventType containercollection.EventType, container *containercollection.Container) error {
+	ev, err := l.containersDs.NewPacketSingle()
+	if err != nil {
+		return fmt.Errorf("creating new containers datasource packet: %w", err)
+	}
+
+	l.eventTypeField.PutString(ev, eventType.String())
+	l.cgroupIDField.PutUint64(ev, container.CgroupID)
+	l.mountNsIDField.PutUint64(ev, container.Mntns)
+	l.nameField.PutString(ev, container.Runtime.ContainerName)
+
+	err = l.containersDs.EmitAndRelease(ev)
+	if err != nil {
+		log.Errorf("%v", err)
+		return fmt.Errorf("emitting containers datasource event: %w", err)
+	}
+
+	return nil
 }
 
 func (l *localManagerTrace) PreGadgetRun() error {
@@ -416,6 +444,9 @@ func (l *localManagerTrace) PreGadgetRun() error {
 						attachContainerFunc(event.Container)
 					case containercollection.EventTypeRemoveContainer:
 						detachContainerFunc(event.Container)
+					default:
+						log.Errorf("unknown event type, expected either %s or %s, got %s", containercollection.EventTypeAddContainer, containercollection.EventTypeRemoveContainer, event.Type)
+						return
 					}
 				},
 			)
@@ -454,6 +485,9 @@ func (l *localManagerTrace) PostGadgetRun() error {
 				l.attacher.DetachContainer(l.manager.fakeContainer)
 			}
 		}
+	}
+	if l.containersSubscriptionKey != "" {
+		l.manager.igManager.Unsubscribe(l.containersSubscriptionKey)
 	}
 	return nil
 }
@@ -497,6 +531,51 @@ func (l *localManager) InstantiateDataOperator(gadgetCtx operators.GadgetContext
 		return nil, err
 	}
 
+	cfg, ok := gadgetCtx.GetVar("config")
+	if !ok {
+		return nil, fmt.Errorf("missing configuration")
+	}
+	v, ok := cfg.(*viper.Viper)
+	if !ok {
+		return nil, fmt.Errorf("invalid configuration format")
+	}
+
+	enableContainersDs := v.GetBool("annotations.enable-containers-datasource")
+
+	var containersDs datasource.DataSource
+	var eventTypeField datasource.FieldAccessor
+	var cgroupIDField datasource.FieldAccessor
+	var mountNsIDField datasource.FieldAccessor
+	var nameField datasource.FieldAccessor
+
+	if enableContainersDs {
+		containersDs, err = gadgetCtx.RegisterDataSource(datasource.TypeSingle, "containers")
+		if err != nil {
+			return nil, fmt.Errorf("creating datasource: %w", err)
+		}
+		containersDs.AddAnnotation("cli.default-output-mode", "none")
+
+		eventTypeField, err = containersDs.AddField("event_type", api.Kind_String)
+		if err != nil {
+			return nil, fmt.Errorf("adding field event_type: %w", err)
+		}
+
+		cgroupIDField, err = containersDs.AddField("cgroup_id", api.Kind_Uint64)
+		if err != nil {
+			return nil, fmt.Errorf("adding field cgroup_id: %w", err)
+		}
+
+		mountNsIDField, err = containersDs.AddField("mntns_id", api.Kind_Uint64)
+		if err != nil {
+			return nil, fmt.Errorf("adding field mntns_id: %w", err)
+		}
+
+		nameField, err = containersDs.AddField("name", api.Kind_String)
+		if err != nil {
+			return nil, fmt.Errorf("adding field name: %w", err)
+		}
+	}
+
 	// Wrapper is used to have ParamDescs() with the new signature
 	traceInstance := &localManagerTraceWrapper{
 		localManagerTrace: localManagerTrace{
@@ -506,6 +585,12 @@ func (l *localManager) InstantiateDataOperator(gadgetCtx operators.GadgetContext
 			params:             params,
 			gadgetCtx:          gadgetCtx,
 			eventWrappers:      make(map[datasource.DataSource]*compat.EventWrapperBase),
+
+			containersDs:   containersDs,
+			eventTypeField: eventTypeField,
+			cgroupIDField:  cgroupIDField,
+			mountNsIDField: mountNsIDField,
+			nameField:      nameField,
 		},
 	}
 
@@ -618,6 +703,46 @@ func (l *localManagerTraceWrapper) PreStart(gadgetCtx operators.GadgetContext) e
 }
 
 func (l *localManagerTraceWrapper) Start(gadgetCtx operators.GadgetContext) error {
+	host := l.params.Get(Host).AsBool()
+	var containers []*containercollection.Container
+
+	if l.containersDs == nil {
+		return nil
+	}
+
+	if l.manager.igManager != nil {
+		containerSelector := containercollection.ContainerSelector{
+			Runtime: containercollection.RuntimeSelector{
+				ContainerName: l.params.Get(ContainerName).AsString(),
+			},
+		}
+
+		l.containersSubscriptionKey = uuid.New().String()
+
+		log.Debugf("add datasource containers subscription to igManager")
+		containers = l.manager.igManager.Subscribe(
+			l.containersSubscriptionKey,
+			containerSelector,
+			func(event containercollection.PubSubEvent) {
+				err := l.emitContainersDatasourceEvent(event.Type, event.Container)
+				if err != nil {
+					log.Errorf("emitting containers datasource event: %v", err)
+				}
+			},
+		)
+	}
+
+	if host {
+		containers = append(containers, l.manager.fakeContainer)
+	}
+
+	for _, container := range containers {
+		err := l.emitContainersDatasourceEvent(containercollection.EventTypeAddContainer, container)
+		if err != nil {
+			return fmt.Errorf("emitting containers datasource event: %w", err)
+		}
+	}
+
 	return nil
 }
 
--- a/docs/spec/operators/localmanager.md
+++ b/docs/spec/operators/localmanager.md
@@ -105,3 +105,19 @@ Show data from both the host and containers
 Fully qualified name: `operator.LocalManager.host`
 
 Default: `false`
+
+## Containers datasource
+
+This operator creates the `containers` datasource which publishes events about containers being created or deleted.
+It is only activated if the gadget has the following annotations:
+
+```yaml
+annotations:
+  enable-containers-datasource: true
+```
+
+So far, the event published contains the following fields:
+- `event_type`: Either "CREATED" or "DELETED".
+- `name`: The container name.
+- `mntns_id`: The container mount namespace ID.
+- `cgroup_id`: The container cgroup ID.
--- a/gadgets/Makefile
+++ b/gadgets/Makefile
@@ -47,6 +47,7 @@ GADGETS = \
 	top_tcp \
 	snapshot_process \
 	snapshot_socket \
+	ci/datasource-containers \
 	ci/inner_fields \
 	ci/sched_cls_drop \
 	#
--- a/gadgets/ci/datasource-containers/build.yaml
+++ b/gadgets/ci/datasource-containers/build.yaml
@@ -1,0 +1,1 @@
+wasm: go/program.go
--- a/gadgets/ci/datasource-containers/gadget.yaml
+++ b/gadgets/ci/datasource-containers/gadget.yaml
@@ -1,0 +1,19 @@
+name: mygadget
+description: Example gadget
+homepageURL: http://mygadget.com
+documentationURL: https://mygadget.com/docs
+sourceURL: https://github.com/my-org/mygadget/
+annotations:
+  enable-containers-datasource: true
+datasources:
+  open:
+    fields:
+      pid:
+        annotations:
+          columns.hidden: "true"
+      tid:
+        annotations:
+          columns.hidden: "true"
+      comm:
+        annotations:
+          columns.hidden: "true"
--- a/gadgets/ci/datasource-containers/go/go.mod
+++ b/gadgets/ci/datasource-containers/go/go.mod
@@ -1,0 +1,9 @@
+module main
+
+go 1.23.0
+
+// Version doesn't matter because of the replace directive below.
+require github.com/inspektor-gadget/inspektor-gadget v0.0.0
+
+// Only needed by in-tree gadgets
+replace github.com/inspektor-gadget/inspektor-gadget => ../../../../
--- a/gadgets/ci/datasource-containers/go/program.go
+++ b/gadgets/ci/datasource-containers/go/program.go
@@ -1,0 +1,81 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package main
+
+import (
+	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
+)
+
+//export gadgetPreStart
+func gadgetPreStart() int {
+	ds, err := api.GetDataSource("containers")
+	if err != nil {
+		api.Errorf("Failed to get data source: %v", err)
+		return 1
+	}
+
+	eventTypeField, err := ds.GetField("event_type")
+	if err != nil {
+		api.Errorf("Failed to get field: %v", err)
+		return 1
+	}
+
+	cgroupIDField, err := ds.GetField("cgroup_id")
+	if err != nil {
+		api.Errorf("Failed to get field: %v", err)
+		return 1
+	}
+
+	mntnsIDField, err := ds.GetField("mntns_id")
+	if err != nil {
+		api.Errorf("Failed to get field: %v", err)
+		return 1
+	}
+
+	nameField, err := ds.GetField("name")
+	if err != nil {
+		api.Errorf("Failed to get field: %v", err)
+		return 1
+	}
+
+	ds.Subscribe(func(ds api.DataSource, data api.Data) {
+		_, err := eventTypeField.String(data, 7)
+		if err != nil {
+			api.Errorf("getting event_type from corresponding field: %v", err)
+			return
+		}
+
+		_, err = cgroupIDField.Uint64(data)
+		if err != nil {
+			api.Errorf("getting cgroup_id from corresponding field: %v", err)
+			return
+		}
+
+		_, err = mntnsIDField.Uint64(data)
+		if err != nil {
+			api.Errorf("getting mntns_id from corresponding field: %v", err)
+			return
+		}
+
+		_, err = nameField.String(data, 64)
+		if err != nil {
+			api.Errorf("getting name from corresponding field: %v", err)
+			return
+		}
+	}, 0)
+
+	return 0
+}
+
+func main() {}
--- a/gadgets/ci/datasource-containers/program.bpf.c
+++ b/gadgets/ci/datasource-containers/program.bpf.c
@@ -1,0 +1,27 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+/* Copyright (c) 2025 The Inspektor Gadget authors */
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <gadget/buffer.h>
+#include <gadget/common.h>
+#include <gadget/macros.h>
+#include <gadget/types.h>
+
+#define NAME_MAX 255
+
+struct event {
+	struct gadget_process proc;
+};
+
+GADGET_TRACER_MAP(events, 1024 * 256);
+
+GADGET_TRACER(open, events, event);
+
+/* Needed to trigger the localmanager operator. */
+SEC("tracepoint/syscalls/sys_enter_openat")
+int enter_openat(struct syscall_trace_enter *ctx)
+{
+	return 0;
+}
+
+char LICENSE[] SEC("license") = "GPL";
--- a/gadgets/ci/datasource-containers/test/integration/datasource_test.go
+++ b/gadgets/ci/datasource-containers/test/integration/datasource_test.go
@@ -1,0 +1,111 @@
+// Copyright 2024 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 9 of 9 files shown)


mauriciovasquezbernal:
> Since there is no subscription, because the guest does not exist at this moment, the function is not called.
> So, the function is only called when the guest exists, hence we are missing containers existing before the guest was started.
> 
> I am wondering if we do not need a PostGadgetStart() hook where we would actually handle this.
> WDYT @flyth @mauriciovasquezbernal?

Unfortunately I don't think a `PostGadgetStart()` will work, in the ig-k8s or ig daemon cases the {Local,Kube}Manager operator would have been running before the gadget is started, so it'll also miss the existing containers. 

I see the following options now:
1. Implement a `subscribe_for_containers` API for wasm, but instead of copying the full container structure with all fields, we can use datasource logic to pass an api.Data object instead and let the user get only the fields they are interested on. This could also help to decouple the API, as adding new fields (or changing the existing ones) won't affect the layout of any copied structure.
2. Implement a `subscribe_for_containers` API for wasm. In this case, defining a container structure that is fully copied. It seems more inefficient and less flexible, but also easier to implement. 
3. Follow the approach on #3762, by returning a container handle and then having specific functions to access each of its fields. IMHO this is a too granular API. 

Perhaps option 2 can be implemented so we unblock the traceloop work?

Happy to get more opinions on this one: 
cc @blanquicet @flyth @alban @burak-ok

flyth:
[This](https://github.com/inspektor-gadget/inspektor-gadget/commit/82622f546890140c395b9582d95c2e6e4346535d) tries to add a buffering mechanism for existing containers.

What it does:
* creates a new array and stores existing container information in there
* on subscription events, it will check whether the gadget is already running. If not, it will add container information to the array, if yes, it will just pass the info to the datasource
* in Start(), the container information from the array is drained

I haven't tested it yet, but it should be conform to the lifecycle promises (emit data only AFTER Start() is called).

mauriciovasquezbernal:
> [This](https://github.com/inspektor-gadget/inspektor-gadget/commit/82622f546890140c395b9582d95c2e6e4346535d) tries to add a buffering mechanism for existing containers.
> 
> What it does:
> 
>     * creates a new array and stores existing container information in there
> 
>     * on subscription events, it will check whether the gadget is already running. If not, it will add container information to the array, if yes, it will just pass the info to the datasource
> 
>     * in Start(), the container information from the array is drained
> 
> 
> I haven't tested it yet, but it should be conform to the lifecycle promises (emit data only AFTER Start() is called).

This is interesting. Now I realized `localManagerTrace` is a per gadget object. At this point I'm wondering if we need the buffer, perhaps we can follow this approach:
1. Create the data source on PreStart(): **do not emit any events yet**
2. The gadget subscribes to the data source on PreStart() too. (local manager should run before that the wasm operator, so the datasource should be available at this point)
3. Create a new subcription to the container collection on Start(), emitting all the existing containers. At this point the gadget has already subscribed, so it doesn't miss the initial containers

I think all of this should be optional to avoid a performance overhead, perhaps a new annotation on the metadata to enable it?

burak-ok:
> So we should wrap all calls to `Function.Call()` with a mutex.

As long as `Function.Call()` is used for different exported functions concurrently there is no issue though. The problem arises when the same exported function is called with the same `Function` wazero object at the same time

blanquicet:
> > 2. How do we supress the output of the container created events at the output? This is normally done through deactivating the Datasource in WASM.
> >    This would break our easy way of stacking multiple layers of WASM in a gadget. Now one has to make sure only the last WASM program that needs the containers deactivates it.
> > ```
> 
> I was thinking to add a "hidden" boolean to datasource.
>

I don't think we should disable (`Unreference`) it because it prevents other WASM layers, as Burak said, or even other operators (for instance the [event-generator](https://github.com/inspektor-gadget/inspektor-gadget/issues/3560)) from leveraging it. Instead, IMO, we finally found a use case for the `none` output mode I removed [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/00d70daa910a02476f5ac0b6d7664bd6c3a408c2). Using it, the localmanager (and kubemanager) should set the default output mode:

```bash
        containersDs, err := gadgetCtx.RegisterDataSource(datasource.TypeSingle, "containers")
	if err != nil {
		return nil, fmt.Errorf("creating datasource: %w", err)
	}
	containersDs.AddAnnotation("cli.default-output-mode", "none")
```

Doing so, any IG consumers (e.g., kubectl-gadget, gadgectl or an UX like Headlamp) have the possibility to display it whenever they want, in the format they want:

```bash
$ gadgetctl run trace_exec --output exec:columns,containers:json
RUNTIME.CONTAINERNAME                      COMM                    PID        TID       PPID ARGS                     ERROR
{"cgroup_id":20279,"event_type":"CREATED","mntns_id":4026534465}
{"cgroup_id":9724,"event_type":"CREATED","mntns_id":4026533157}
mycontainer                                ls                   120724     120724     114685 /usr/bin/ls --color=au…
```

But, by default, it will be not displayed:

```bash
$ gadgetctl run trace_exec 
RUNTIME.CONTAINERNAME                      COMM                    PID        TID       PPID ARGS                     ERROR
mycontainer                                ls                   120724     120724     114685 /usr/bin/ls --color=au…
```

I gave it a try [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/719e82f2ca34a587f4663e31ab0a7b3b092b867e) and it worked.

In the future, we could optimize things and instead of sending data to client and not displaying it, we should provide a way to not send it at all. However, we don't have anything like that currently, and I don't see a quick way to implement it. So, to unblock this PR, I suggest going with the cli annotation.

alban:
> > So we should wrap all calls to `Function.Call()` with a mutex.
> 
> As long as `Function.Call()` is used for different exported functions concurrently there is no issue though. The problem arises when the same exported function is called with the same `Function` wazero object at the same time

Are you sure? There might be no problem from the wazero point of view but depending on the gadget, these wasm functions might use some data structures in the wasm module that the gadget author didn't plan to make thread-safe. Even if the gadget author wants to write their wasm module in a thread-safe way, it is not really possible: [tinygo is meant to run on a single core](https://tinygo.org/docs/guides/tips-n-tricks/). Since running in a single core is assumed, the tinygo implementation of mutexes does not use atomic operations since it is not possible to have other threads running in parallel:
- [sync/Mutex.Lock in tinygo](https://github.com/tinygo-org/tinygo/blob/v0.35.0/src/sync/mutex.go#L16) (trivial implementation assuming single core with cooperative scheduling)
- [sync/Mutex.Lock in golang](https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/sync/mutex.go;l=83) (using atomic operations)

If you run several `Function.Call()` in parallel from the same wasm module, even from 2 different functions, the tinygo assumptions would break.

pkg/operators/localmanager/localmanager.go
```diff
@@ -497,6 +530,27 @@ func (l *localManager) InstantiateDataOperator(gadgetCtx operators.GadgetContext
 		return nil, err
 	}
 
+	containersDs, err := gadgetCtx.RegisterDataSource(datasource.TypeSingle, "containers")
```
blanquicet, line 533: This ds should be registered only if gadget requests for it. I suggest using a localmanager operator's annotation.

cc @mauriciovasquezbernal

pkg/operators/localmanager/localmanager.go
```diff
@@ -497,6 +530,27 @@ func (l *localManager) InstantiateDataOperator(gadgetCtx operators.GadgetContext
...
+
+	cgroupIDField, err := containersDs.AddField("cgroup_id", api.Kind_Uint64)
+	if err != nil {
+		return nil, fmt.Errorf("adding field cgroup_id: %w", err)
+	}
+
+	mountNsIDField, err := containersDs.AddField("mntns_id", api.Kind_Uint64)
+	if err != nil {
+		return nil, fmt.Errorf("adding field mntns_id: %w", err)
+	}
```
blanquicet, line 552: Not specifically related with this PR, but I think we should have a way to register a struct as a datasource to avoid having to manually add each field like this. Same for setting values.

blanquicet (commented):
It's just missing some details, but I think this approach is the way to go.

pkg/operators/wasm/datasource.go
mauriciovasquezbernal, line 1: nit: perhaps you can create a function that calls the callback with the lock and reuse it:

```diff
diff --git a/pkg/operators/wasm/datasource.go b/pkg/operators/wasm/datasource.go
index 0b79fcd3..fbc55ec3 100644
--- a/pkg/operators/wasm/datasource.go
+++ b/pkg/operators/wasm/datasource.go
@@ -266,6 +266,13 @@ const (
        subscriptionTypePacket subscriptionType = 3
 )

+func (i *wasmOperatorInstance) callDsCallback(ctx context.Context, cbID uint64, dsHandle uint64, dataHandle uint64) error {
+       i.dataSourceCallbackLock.Lock()
+       _, err := i.dataSourceCallback.Call(ctx, cbID, dsHandle, dataHandle)
+       i.dataSourceCallbackLock.Unlock()
+       return err
+}
+
 // dataSourceSubscribe subscribes to the datasource.
 // Params:
 // - stack[0]: DataSource handle
@@ -297,27 +304,22 @@ func (i *wasmOperatorInstance) dataSourceSubscribe(ctx context.Context, m wapi.M
        case subscriptionTypeData:
                err = ds.Subscribe(func(source datasource.DataSource, data datasource.Data) error {
                        tmpData := i.addHandle(data)
-                       i.dataSourceCallbackLock.Lock()
-                       _, err := i.dataSourceCallback.Call(ctx, cbID, stack[0], wapi.EncodeU32(tmpData))
-                       i.dataSourceCallbackLock.Unlock()
+                       i.callDsCallback(ctx, cbID, stack[0], wapi.EncodeU32(tmpData))
                        i.delHandle(tmpData)
                        return err
                }, int(prio))
        case subscriptionTypeArray:
                err = ds.SubscribeArray(func(source datasource.DataSource, data datasource.DataArray) error {
                        tmpData := i.addHandle(data)
-                       i.dataSourceCallbackLock.Lock()
-                       _, err := i.dataSourceCallback.Call(ctx, cbID, stack[0], wapi.EncodeU32(tmpData))
-                       i.dataSourceCallbackLock.Unlock()
+                       i.callDsCallback(ctx, cbID, stack[0], wapi.EncodeU32(tmpData))
                        i.delHandle(tmpData)
                        return err
                }, int(prio))
        case subscriptionTypePacket:
                err = ds.SubscribePacket(func(source datasource.DataSource, data datasource.Packet) error {
                        tmpData := i.addHandle(data)
+                       i.callDsCallback(ctx, cbID, stack[0], wapi.EncodeU32(tmpData))
                        i.dataSourceCallbackLock.Lock()
-                       _, err := i.dataSourceCallback.Call(ctx, cbID, stack[0], wapi.EncodeU32(tmpData))
-                       i.dataSourceCallbackLock.Unlock()
                        i.delHandle(tmpData)
                        return err
                }, int(prio))
```

pkg/operators/localmanager/localmanager.go
```diff
@@ -618,6 +677,35 @@ func (l *localManagerTraceWrapper) PreStart(gadgetCtx operators.GadgetContext) e
...
+			Runtime: containercollection.RuntimeSelector{
+				ContainerName: l.params.Get(ContainerName).AsString(),
+			},
+		}
+
+		log.Debugf("add subscription to igManager")
+		containers = l.manager.igManager.Subscribe(
+			uuid.New().String(),
+			containerSelector,
+			func(event containercollection.PubSubEvent) {},
```
mauriciovasquezbernal, line 694: I think we need to call emitContainersDatasourceEvent here instead of in the other subscription, otherwise an event for the same container could be generated twice (or a container missed) as both subscriptions aren't guaranteed to be synchronized.

mauriciovasquezbernal (commented):
Some things:
- I like the mutex idea for the dscallback. The ebpf operator uses the same goroutine to generate events, but there is not any guarantee it'll keep doing so in the future. Also, other operators generating events could not have this guarantee, so I think it's better to fix it in the wasm operator.
- Please don't forget make this an opt-in feature only for gadgets interested on it.
- Do you plan to support it on kubemanager as well?
- Do you plan to add all fields of the container? (it's totally fine for me to only have cgroup id and mnt ns for now)

pkg/operators/localmanager/localmanager.go
mauriciovasquezbernal: The entry on gadget.yml doesn't create the datasource but just provides its config. IMO this annotation can be part of the gadget itself:

```diff
diff --git a/gadgets/datasource/gadget.yaml b/gadgets/datasource/gadget.yaml
index d4627baa..627498f4 100644
--- a/gadgets/datasource/gadget.yaml
+++ b/gadgets/datasource/gadget.yaml
@@ -3,6 +3,8 @@ description: Example gadget
 homepageURL: http://mygadget.com
 documentationURL: https://mygadget.com/docs
 sourceURL: https://github.com/my-org/mygadget/
+annotations:
+  enable-containers-datasource: true
 datasources:
   open:
     fields:
diff --git a/pkg/operators/localmanager/localmanager.go b/pkg/operators/localmanager/localmanager.go
index 97434111..129392dc 100644
--- a/pkg/operators/localmanager/localmanager.go
+++ b/pkg/operators/localmanager/localmanager.go
@@ -26,6 +26,7 @@ import (
        securejoin "github.com/cyphar/filepath-securejoin"
        "github.com/google/uuid"
        log "github.com/sirupsen/logrus"
+       "github.com/spf13/viper"

        commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
        containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
@@ -530,25 +531,43 @@ func (l *localManager) InstantiateDataOperator(gadgetCtx operators.GadgetContext
                return nil, err
        }

-       containersDs, err := gadgetCtx.RegisterDataSource(datasource.TypeSingle, "containers")
-       if err != nil {
-               return nil, fmt.Errorf("creating datasource: %w", err)
+       cfg, ok := gadgetCtx.GetVar("config")
+       if !ok {
+               return nil, fmt.Errorf("missing configuration")
        }
-       containersDs.AddAnnotation("cli.default-output-mode", "none")
-
-       eventTypeField, err := containersDs.AddField("event_type", api.Kind_String)
-       if err != nil {
-               return nil, fmt.Errorf("adding field event_type: %w", err)
+       v, ok := cfg.(*viper.Viper)
+       if !ok {
+               return nil, fmt.Errorf("invalid configuration format")
        }

-       cgroupIDField, err := containersDs.AddField("cgroup_id", api.Kind_Uint64)
-       if err != nil {
-               return nil, fmt.Errorf("adding field cgroup_id: %w", err)
-       }
+       enableContainersDs := v.GetBool("annotations.enable-containers-datasource")

-       mountNsIDField, err := containersDs.AddField("mntns_id", api.Kind_Uint64)
-       if err != nil {
-               return nil, fmt.Errorf("adding field mntns_id: %w", err)
+       var containersDs datasource.DataSource
+       var eventTypeField datasource.FieldAccessor
+       var cgroupIDField datasource.FieldAccessor
+       var mountNsIDField datasource.FieldAccessor
+
+       if enableContainersDs {
+               containersDs, err = gadgetCtx.RegisterDataSource(datasource.TypeSingle, "containers")
+               if err != nil {
+                       return nil, fmt.Errorf("creating datasource: %w", err)
+               }
+               containersDs.AddAnnotation("cli.default-output-mode", "none")
+
+               eventTypeField, err = containersDs.AddField("event_type", api.Kind_String)
+               if err != nil {
+                       return nil, fmt.Errorf("adding field event_type: %w", err)
+               }
+
+               cgroupIDField, err = containersDs.AddField("cgroup_id", api.Kind_Uint64)
+               if err != nil {
+                       return nil, fmt.Errorf("adding field cgroup_id: %w", err)
+               }
+
+               mountNsIDField, err = containersDs.AddField("mntns_id", api.Kind_Uint64)
+               if err != nil {
+                       return nil, fmt.Errorf("adding field mntns_id: %w", err)
+               }
        }

        // Wrapper is used to have ParamDescs() with the new signature
```

mauriciovasquezbernal:
> This would be needed to have traceloop behave the same between being run by ig and kubectl-gadget?
In this case, yes, but I will handle this in a later PR.

Yes, this would be needed for running traceloop on k8s.

pkg/operators/localmanager/localmanager.go
blanquicet: I agree the annotation should be part of the gadget and the implementation looks like what Mauro suggested. I just propose to use the notation: `annotations.operator.LocalManager.enable-containers-datasource`. As it's not a generic annotation, we are requesting the localmanager operator to do that.

mauriciovasquezbernal: And what about when this is going to be implemented on the kubemanager? Should the gadget add a second annotation?

flyth:
> I don't think we should disable (`Unreference`) it because it prevents other WASM layers, as Burak said, or even other operators (for instance the [event-generator](https://github.com/inspektor-gadget/inspektor-gadget/issues/3560)) from leveraging it. Instead, IMO, we finally found a use case for the `none` output mode I removed [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/00d70daa910a02476f5ac0b6d7664bd6c3a408c2). Using it, the localmanager (and kubemanager) should set the default output mode:

IMHO we should take it a step further. We could introduce a new annotation "internal" that is set on this datasource and makes sure it doesn't get sent over the network when in client/server mode. And those "internal" datasources should be skipped by the sink operators as well. (I still would like to see the `none` output mode re-introduced 😄)

blanquicet:
> > I don't think we should disable (`Unreference`) it because it prevents other WASM layers, as Burak said, or even other operators (for instance the [event-generator](https://github.com/inspektor-gadget/inspektor-gadget/issues/3560)) from leveraging it. Instead, IMO, we finally found a use case for the `none` output mode I removed [here](https://github.com/inspektor-gadget/inspektor-gadget/commit/00d70daa910a02476f5ac0b6d7664bd6c3a408c2). Using it, the localmanager (and kubemanager) should set the default output mode:
> 
> IMHO we should take it a step further. We could introduce a new annotation "internal" that is set on this datasource and makes sure it doesn't get sent over the network when in client/server mode. And those "internal" datasources should be skipped by the sink operators as well. (I still would like to see the `none` output mode re-introduced 😄)

Yeah, that's the proper solution but the `none` output is quicker and I don't want to block this PR more. We can work on that in a follow-up PR as such a "waste" of network will only happen for the traceloop and not for all gadgets (thanks to the `enable-containers-datasource` annotation).

pkg/operators/localmanager/localmanager.go
blanquicet: 🤦🏽  ... Good point!

mauriciovasquezbernal:
> I wanted to tweak the datasource gadget to use it as a test, but it seems the output datasource never prints data. Anyone has any better idea on how to test it?

Two options:
1. Implement it as an integration test (running the ci/datasource gadget with -o json prints the events to the terminal)
2. Use the following diff to make it work with the unit test you implemented:

```diff
diff --git a/gadgets/ci/datasource/test/unit/datasource_test.go b/gadgets/ci/datasource/test/unit/datasource_test.go
index ea033579..13bb4e84 100644
--- a/gadgets/ci/datasource/test/unit/datasource_test.go
+++ b/gadgets/ci/datasource/test/unit/datasource_test.go
@@ -17,18 +17,15 @@ package tests
 import (
        "fmt"
        "os"
-       "os/user"
        "testing"
        "time"

-       "golang.org/x/sys/unix"
-
-       "github.com/stretchr/testify/require"
-
        gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
        utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
        "github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+       "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/localmanager"
        "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/gadgetrunner"
+       "github.com/stretchr/testify/require"
 )

 type ExpectedEvent struct {
@@ -39,6 +36,13 @@ func TestDatasource(t *testing.T) {
        gadgettesting.InitUnitTest(t)
        t.Parallel()

+       localManagerOp := localmanager.LocalManagerOperator
+       localManagerParams := localManagerOp.GlobalParamDescs().ToParams()
+       localManagerParams.Get(localmanager.Runtimes).Set("docker")
+       err := localManagerOp.Init(localManagerParams)
+       require.NoError(t, err, "init local manager")
+       defer localManagerOp.Close()
+
        runnerConfig := &utilstest.RunnerConfig{}
        runner := utilstest.NewRunnerWithTest(t, runnerConfig)

@@ -52,7 +56,7 @@ func TestDatasource(t *testing.T) {
        if _, ok := os.LookupEnv("GADGET_REPOSITORY"); !ok {
                image = fmt.Sprintf("ghcr.io/inspektor-gadget/gadget/%s", image)
        }
-       opts := gadgetrunner.GadgetRunnerOpts[ExpectedInnerEvent]{
+       opts := gadgetrunner.GadgetRunnerOpts[ExpectedEvent]{
                Image:       image,
                Timeout:     5 * time.Second,
                OnGadgetRun: onGadgetRun,
@@ -60,8 +64,8 @@ func TestDatasource(t *testing.T) {
        gadgetRunner := gadgetrunner.NewGadgetRunner(t, opts)
        gadgetRunner.RunGadget()

-       utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, _ int) *ExpectedInnerEvent {
-               return &ExpectedEvent{ Name: "test-container-datasource" }
+       utilstest.ExpectAtLeastOneEvent(func(info *utilstest.RunnerInfo, _ int) *ExpectedEvent {
+               return &ExpectedEvent{Name: "test-container-datasource"}
        })(t, runner.Info, 0, gadgetRunner.CapturedEvents)
 }

diff --git a/pkg/operators/localmanager/localmanager.go b/pkg/operators/localmanager/localmanager.go
index d14a6a2f..2e10c844 100644
--- a/pkg/operators/localmanager/localmanager.go
+++ b/pkg/operators/localmanager/localmanager.go
@@ -26,7 +26,7 @@ import (
        securejoin "github.com/cyphar/filepath-securejoin"
        "github.com/google/uuid"
        log "github.com/sirupsen/logrus"
-       "github.com/spf13/viper"
+       "github.com/spf13/viper"

        commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
        containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
@@ -749,7 +749,7 @@ func (l *localManagerTraceWrapper) Stop(gadgetCtx operators.GadgetContext) error
 }

 func init() {
-       lm := &localManager{}
+       lm := LocalManagerOperator
        operators.Register(lm)
        operators.RegisterDataOperator(lm)
 }
```

I'm not sure about the docker dependency for the unit tests, it seems too much for me. Perhaps it's better to implement it as an integration test?

pkg/operators/localmanager/localmanager.go
```diff
@@ -618,6 +699,48 @@ func (l *localManagerTraceWrapper) PreStart(gadgetCtx operators.GadgetContext) e
...
+
+	if l.manager.igManager != nil {
+		containerSelector := containercollection.ContainerSelector{
+			Runtime: containercollection.RuntimeSelector{
+				ContainerName: l.params.Get(ContainerName).AsString(),
+			},
+		}
+
+		log.Debugf("add subscription to igManager")
+		containers = l.manager.igManager.Subscribe(
```
mauriciovasquezbernal, line 713: I think we need to save the key and call unscribe later on.

mauriciovasquezbernal (commented):
Looks very good. Some final nits and it should ready to go after them.

mqasimsarfraz (approved):
Looking great, I am able to test this:

```
$ go run -exec sudo ./cmd/ig run ghcr.io/inspektor-gadget/gadget/ci/datasource-containers:francis-datasource-containers --verify-image=false -o json
WARN[0001] This gadget was built with ig v0.37.0-171-g717f5fcc6 and it's being run with v0.0.0. Gadget could be incompatible 
WARN[0001] This gadget was built with ig v0.37.0-171-g717f5fcc6 and it's being run with v0.0.0. Gadget could be incompatible 
{"cgroup_id":23944,"event_type":"CREATED","mntns_id":4026533092,"name":"1de9b668ac18d64b8bee6dc1afd14c92576b03c7ed97c421b57a61e13f09c9f8"}
{"cgroup_id":16676,"event_type":"CREATED","mntns_id":4026533029,"name":"8c9d85b2c2554ae20cd58402aab8294473d8757ab7fa1f00dd0fef37cb0eaab8"}
{"cgroup_id":23944,"event_type":"DELETED","mntns_id":4026533092,"name":"1de9b668ac18d64b8bee6dc1afd14c92576b03c7ed97c421b57a61e13f09c9f8"}
{"cgroup_id":24260,"event_type":"CREATED","mntns_id":4026533093,"name":"cf84779c4fcd7bb63643ac9e302ed105e8ef1adca22cce923105ffa206aabd14"}

```

I don't like the output when it is combined with an existing gadgets because then we get different events:

```
 → go run -exec sudo ./cmd/ig run ghcr.io/inspektor-gadget/gadget/trace_dns:francis-datasource-containers --verify-image=false  -o json
WARN[0001] This gadget was built with ig v0.37.0-171-g717f5fcc6 and it's being run with v0.0.0. Gadget could be incompatible 
WARN[0002] This gadget was built with ig v0.37.0-171-g717f5fcc6 and it's being run with v0.0.0. Gadget could be incompatible 
{"cgroup_id":24618,"event_type":"CREATED","mntns_id":4026533093,"name":"9c10c0d59c9689180ba5a3bd59e40a231f9814736f4edbe5b984c03c1ba0526f"}
{"cgroup_id":16676,"event_type":"CREATED","mntns_id":4026533029,"name":"8c9d85b2c2554ae20cd58402aab8294473d8757ab7fa1f00dd0fef37cb0eaab8"}
{"addresses":"","cwd":"","dst":{"addr":"192.168.0.1","port":53,"proto":"UDP","proto_raw":17,"version":4},"exepath":"","id":"5d1b","k8s":{"containerName":"","hostnetwork":false,"namespace":"","node":"","owner":{"kind":"","name":""},"podLabels":"","podName":""},"latency_ns":"0ns","latency_ns_raw":0,"name":"google.com.","netns_id":4026533157,"num_answers":0,"pkt_type":"OUTGOING","pkt_type_raw":4,"proc":{"comm":"nslookup","creds":{"gid":0,"group":"root","uid":0,"user":"root"},"mntns_id":4026533093,"parent":{"comm":"sh","pid":107921},"pid":109120,"tid":109120},"qr":"Q","qr_raw":false,"qtype":"A","qtype_raw":1,"ra":false,"rcode":"","rcode_raw":0,"rd":true,"runtime":{"containerId":"9c10c0d59c9689180ba5a3bd59e40a231f9814736f4edbe5b984c03c1ba0526f","containerImageDigest":"sha256:a5d0ce49aa801d475da48f8cb163c354ab95cab073cd3c138bd458fc8257fbf1","containerImageName":"docker.io/library/busybox:latest","containerName":"9c10c0d59c9689180ba5a3bd59e40a231f9814736f4edbe5b984c03c1ba0526f","containerPid":107921,"containerStartedAt":1740497414811625064,"runtimeName":"containerd"},"src":{"addr":"172.17.0.3","port":50421,"proto":"UDP","proto_raw":17,"version":4},"tc":false,"timestamp":"2025-02-25T16:43:41.751719519+01:00","timestamp_raw":1740498221751719519}
{"addresses":"","cwd":"","dst":{"addr":"192.168.0.1","port":53,"proto":"UDP","proto_raw":17,"version":4},"exepath":"","id":"5d1c","k8s":{"containerName":"","hostnetwork":false,"namespace":"","node":"","owner":{"kind":"","name":""},"podLabels":"","podName":""},"latency_ns":"0ns","latency_ns_raw":0,"name":"google.com.","netns_id":4026533157,"num_answers":0,"pkt_type":"OUTGOING","pkt_type_raw":4,"proc":{"comm":"nslookup","creds":{"gid":0,"group":"root","uid":0,"user":"root"},"mntns_id":4026533093,"parent":{"comm":"sh","pid":107921},"pid":109120,"tid":109120},"qr":"Q","qr_raw":false,"qtype":"AAAA","qtype_raw":28,"ra":false,"rcode":"","rcode_raw":0,"rd":true,"runtime":{"containerId":"9c10c0d59c9689180ba5a3bd59e40a231f9814736f4edbe5b984c03c1ba0526f","containerImageDigest":"sha256:a5d0ce49aa801d475da48f8cb163c354ab95cab073cd3c138bd458fc8257fbf1","containerImageName":"docker.io/library/busybox:latest","containerName":"9c10c0d59c9689180ba5a3bd59e40a231f9814736f4edbe5b984c03c1ba0526f","containerPid":107921,"containerStartedAt":1740497414811625064,"runtimeName":"containerd"},"src":{"addr":"172.17.0.3","port":50421,"proto":"UDP","proto_raw":17,"version":4},"tc":false,"timestamp":"2025-02-25T16:43:41.751788722+01:00","timestamp_raw":1740498221751788722}

```

But I understand it will be fixed by https://github.com/inspektor-gadget/inspektor-gadget/pull/3957#issuecomment-2654797037 so please open an issue for it!

PR #1770, opened 2023-06-21
trace/exec: Fix reading task->loginuid in kernels without CONFIG_AUDIT

mauriciovasquezbernal:
Use bpf_core_field_exists() to only read this value when it's available.

Fixes https://github.com/inspektor-gadget/inspektor-gadget/issues/1769

@alban do you have an environment ready to test it out?

PR #3961, opened 2025-01-29
Fix reading loginuid/sessionid in kernels without CONFIG_AUDIT

Sefi4:
Use bpf_core_field_exists() to only read this value when it's available


alban (approved):
One suggestion; LGTM afterwards.

This is the same fix as #1770 for the image-based trace_exec.

This should help for GKE kernel 6.6.56.

PR #3967, opened 2025-01-30
gadgets/trace_dns: Use same namespace for client and server

AdvH039:
# Use existing namespaces in test

Allow the use of existing namespaces by creating an option for the same rather than creating a namespace for each respective container when it is ran.
Tested the use of it in the trace_dns gadget test.

### DIFF
--- a/gadgets/trace_dns/test/integration/trace_dns_test.go
+++ b/gadgets/trace_dns/test/integration/trace_dns_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/testutils"
 	igtesting "github.com/inspektor-gadget/inspektor-gadget/pkg/testing"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/containers"
 	igrunner "github.com/inspektor-gadget/inspektor-gadget/pkg/testing/ig"
@@ -85,18 +86,18 @@ func TestTraceDNS(t *testing.T) {
 		clientImage = DefaultClientImage
 	}
 
-	// TODO: The current logic creates the namespace when running the pod, hence
-	// we need a namespace for each pod
-	var nsClient, nsServer string
+	var nsTest string
 	serverContainerOpts := []containers.ContainerOption{containers.WithContainerImage(serverImage)}
 	clientContainerOpts := []containers.ContainerOption{containers.WithContainerImage(clientImage)}
 
 	if utils.CurrentTestComponent == utils.KubectlGadgetTestComponent {
-		nsClient = utils.GenerateTestNamespaceName(t, "test-trace-dns-client")
-		clientContainerOpts = append(clientContainerOpts, containers.WithContainerNamespace(nsClient))
+		nsTest = utils.GenerateTestNamespaceName(t, "test-trace-dns")
+		testutils.CreateK8sNamespace(t, nsTest)
 
-		nsServer = utils.GenerateTestNamespaceName(t, "test-trace-dns-server")
-		serverContainerOpts = append(serverContainerOpts, containers.WithContainerNamespace(nsServer))
+		clientContainerOpts = append(clientContainerOpts, containers.WithContainerNamespace(nsTest))
+		clientContainerOpts = append(clientContainerOpts, containers.WithUseExistingNamespace())
+		serverContainerOpts = append(serverContainerOpts, containers.WithContainerNamespace(nsTest))
+		serverContainerOpts = append(serverContainerOpts, containers.WithUseExistingNamespace())
 	}
 
 	serverContainer := containerFactory.NewContainer(serverContainerName, "/dnstester", serverContainerOpts...)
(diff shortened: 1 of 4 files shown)


AdvH039:
@mqasimsarfraz could you review this PR? Thanks!!

gadgets/trace_dns/test/integration/trace_dns_test.go
```diff
@@ -72,25 +73,25 @@ func TestTraceDNS(t *testing.T) {
...
+		serverContainerOpts = append(serverContainerOpts, containers.WithUseExistingNamespace())
+		testutils.ExecuteCreateTestNamespaceCommand(t, nsTest)
 	}
 
 	serverContainer := containerFactory.NewContainer(serverContainerName, "/dnstester", serverContainerOpts...)
 	serverContainer.Start(t)
 	t.Cleanup(func() {
 		serverContainer.Stop(t)
 	})
+	fmt.Print(nsTest)
```
alban, line 94:
```suggestion
	t.Logf("Using namespace %q", nsTest)
```

alban (commented):
Could you explain why it is useful to run the client and server in the same Kubernetes namespace?

gadgets/trace_dns/test/integration/trace_dns_test.go
AdvH039: The todo comment mentioned there was no option of using a pre-existing namespace so I implemented it. It would probably save us resources used by creating only one namespace.

AdvH039: It would probably be useful in another situation.

AdvH039: @alban Do you want me to remove the modifications in trace dns test and keep the rest as it is?

AdvH039: @alban could you review this again please?

AdvH039: @mauriciovasquezbernal could you take a look please?

gadgets/trace_dns/test/integration/trace_dns_test.go
```diff
@@ -72,25 +73,25 @@ func TestTraceDNS(t *testing.T) {
...
-		clientContainerOpts = append(clientContainerOpts, containers.WithContainerNamespace(nsClient))
-
-		nsServer = utils.GenerateTestNamespaceName(t, "test-trace-dns-server")
-		serverContainerOpts = append(serverContainerOpts, containers.WithContainerNamespace(nsServer))
+		nsTest = utils.GenerateTestNamespaceName(t, "test-trace-dns")
+		clientContainerOpts = append(clientContainerOpts, containers.WithContainerNamespace(nsTest))
+		clientContainerOpts = append(clientContainerOpts, containers.WithUseExistingNamespace())
+		serverContainerOpts = append(serverContainerOpts, containers.WithContainerNamespace(nsTest))
+		serverContainerOpts = append(serverContainerOpts, containers.WithUseExistingNamespace())
+		testutils.ExecuteCreateTestNamespaceCommand(t, nsTest)
```
mauriciovasquezbernal, line 86: - It could be named `CreateK8sNamespace()`
- Internally it could register the deletion of the namespace using `t.Cleanup()`, so the caller doesn't have to worry about it.
- It should be executed right after `nsTest = utils.GenerateTestNamespaceName(t, "test-trace-dns")`

mauriciovasquezbernal:
That's trying to print the logs from a previous container in case it has restarted. I don't see any relation to your change.

gadgets/trace_dns/test/integration/trace_dns_test.go
```diff
@@ -85,24 +86,27 @@ func TestTraceDNS(t *testing.T) {
...
+		serverContainerOpts = append(serverContainerOpts, containers.WithUseExistingNamespace())
 	}
 
 	serverContainer := containerFactory.NewContainer(serverContainerName, "/dnstester", serverContainerOpts...)
 	serverContainer.Start(t)
 	t.Cleanup(func() {
 		serverContainer.Stop(t)
+		if utils.CurrentTestComponent == utils.KubectlGadgetTestComponent {
+			testutils.DeleteK8sNamespace(t, nsTest)
+		}
```
mauriciovasquezbernal, line 109: I meant to make it inside CreateK8sNamespace:

```golang
func CreateK8sNamespace(t *testing.T, namespace string) {
	igtesting.RunTestSteps([]igtesting.TestStep{createTestNamespaceCommand(namespace)}, t)
	t.Cleanup(func() {
		DeleteK8sNamespace(t, namespace)
	})
}
```

AdvH039:
> That's trying to print the logs from a previous container in case it has restarted. I don't see any relation to your change.

Ok but it shows those errors when I made the changes, it wasn't there before. I will try your change and see if it improves.

gadgets/trace_dns/test/integration/trace_dns_test.go
AdvH039: Then I don't really need to set DeleteK8sNamespace to be global right? Or should I in case the user wants it?

mauriciovasquezbernal: I think it's better to leave it unexported for now.

mauriciovasquezbernal:
Can you please fix the commit (and PR title) to something like

gadgets/trace_dns: Use same namespace for client and server

And then put something in the commit body if you need extra space to give more details

mauriciovasquezbernal (approved):
LGTM. Thanks!

PR #3382, opened 2024-08-21
Add inspect subcommand to image

eiffel-fl:
Hi.

This PR adds `image inspect`:

```bash
$ sudo ./ig image inspect trace_exec -o jsonpretty          francis/image-inspect % u=
{
  "Repository": "ghcr.io/inspektor-gadget/gadget/trace_exec",
  "Tag": "latest",
  "Digest": "sha256:e13e3859be5ed8cef676a720274480d2748f66fd98cf8d963af6c4c05121526f",
  "Created": "2024-07-24T21:07:26Z"
}%
$ sudo ./ig image inspect trace_quux -o jsonpretty          francis/image-inspect % u=
Error: inspecting image: getting gadget image descriptor: Found tag "ghcr.io/inspektor-gadget/gadget/trace_quux:latest" but couldn't get a descriptor for it: not found
```

TODO:
- [ ]: Add documentation once we agree on the CLI.

Best regards.

PR #3985, opened 2025-02-03
feat: ig inspect subcommand for the wasm module

Gmin2:
# inspect subcommand for the wasm submodule

inspect subcommand for the wasm submodule that gets the `gadgetAPIVersion` and `upcalls` array from the wasm gadget as mentioned in https://github.com/inspektor-gadget/inspektor-gadget/pull/3382#pullrequestreview-2250296366

So my approach may be wrong i could have move the logic for extracting the wasm module inside the `pkg/operators/wasm/wasm.go` but in that case i was not able to pass it down to the oci struct, is there any way i can do that?
or any better design solution to this ?

## Testing done
Screenshot from 2025-02-03 16-55-14


Gmin2:
can i get a review @alban ?

mauriciovasquezbernal:
This (or part of this) was implemented as part of the mentorship, so closing it.

Issue #3992, opened 2025-02-05
[Research] profiling wasm performances


alban:
## Current situation

Gadgets using WASM have a performance penalty (cpu and memory) and we don't have tools to measure it. This research issue is for analysing if the performance penalty is excessive and find ways to improve it.

## Impact

While we notice performance problems, we don't know the root causes and we don't know how to fix them.

## Ideal future situation

- [ ] We have benchmarks with gadgets using wasm and we can monitor improvements and regressions.
- [ ] We can generate cpu flamegraphs showing the stack traces inside the WASM modules
- [ ] We have fixed some of the low-hanging fruits

## Implementation options

- Go benchmarks in `pkg/operators/wasm/` with specific gadgets in `pkg/operators/wasm/testdata/`
- https://github.com/dispatchrun/wzprof (I have a WIP experiment in https://github.com/alban/inspektor-gadget/tree/alban_wzprof)

## Additional information

Previous discussions:
- Profiling the dns gadget for 30 seconds shows it can spend >90% of cpu time in WASM execution in some scenarios (see https://github.com/inspektor-gadget/inspektor-gadget/pull/3956#issuecomment-2633695112)
- The socket enricher can cause performance issues (see #3705 and some fixes in #3923)
- We have added pprof support in Inspektor Gadget, this can be used for identifying the cause of performance issues (see #3902)

cc @mqasimsarfraz

alban:
Using wzprof & wazero, I got the following profile (see my branch [alban_wzprof](https://github.com/alban/inspektor-gadget/tree/alban_wzprof)).

It seems to be spending 97% of its time in runtime.sliceGrow via the dynamic memory allocator in [Field.Bytes()](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/wasmapi/go/fields.go#L180-L185).


![Image](https://github.com/user-attachments/assets/e7005f34-bdd0-406c-9fc4-d05c4a935f79)

alban:
I implemented `Field.BytesToSlice()` (https://github.com/alban/inspektor-gadget/commit/aad4fedc06ab738a88b26835ce78f653ffc40364) as a replacement to `Field.Bytes()` without dynamic allocations. Instead of using malloc/free, the gadget pre-allocate a single buffer. The WASM module tells the host where to write the field.

Copying the fields is no longer the bottleneck. Instead, the profiling now sees the dns parsing as the task taking the cpu.

[...]

PR #3995, opened 2025-02-06
trace_exec, trace_dns: remove dynamic allocations from wasm module

alban:
# wasmapi: implement Field.BytesToSlice without dynamic allocations

Profiling of the wasm modules in the trace_exec and trace_dns gadgets revealed that more than 95% of the cpu time is spent in dynamic allocations done by the `Field.Bytes()` helper from the wasm API. See profiling in #3992.

This patch introduces an alternative way to get the bytes of a field without dynamic allocations for fields of type `string` or `[]byte`:
- `Field.BytesToSlice()` (wasm go API)
- `fieldGetToBuffer()` (underlying wasm helper)

## How to use

No changes

## Testing done

See profiling in #3992.

```
sudo ig run ghcr.io/inspektor-gadget/gadget/trace_exec:alban_fieldGet --verify-image=false
sudo ig run ghcr.io/inspektor-gadget/gadget/trace_dns:alban_fieldGet --verify-image=false 
```

cc @mqasimsarfraz

### DIFF
--- a/docs/gadget-devel/gadget-wasm-api-raw.md
+++ b/docs/gadget-devel/gadget-wasm-api-raw.md
@@ -284,7 +284,7 @@ Return value:
 
 #### `fieldGet(u32 field, u32 data, u32 kind) u64`
 
-Get the value of a field.
+Get the value of a field into a newly allocated buffer.
 
 Parameters:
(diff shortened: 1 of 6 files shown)


mqasimsarfraz (approved):
I tested the changes and less events are being dropped!

PR #4002, opened 2025-02-07
go/wasm: Update to Tinygo 0.35.0

burak-ok:
# go/wasm: Update to Tinygo 0.35.0

This uses the newest info from https://github.com/tetratelabs/wazero/pull/2368


cmd/common/image/helpers/Makefile.build
```diff
@@ -26,7 +26,7 @@ wasm:
 else ifeq (go,$(patsubst %.go,go,$(WASM)))
 wasm: $(WASM)
 	cd $(dir $(WASM)) && \
-	$(TINYGO) build -o $(OUTPUTDIR)/program.wasm -target=wasi --no-debug $(notdir $(WASM))
+	$(TINYGO) build -o $(OUTPUTDIR)/program.wasm -target=wasip1 --no-debug -buildmode=c-shared $(notdir $(WASM))
```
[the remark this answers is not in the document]
burak-ok: This intructs the compiler that your WASM binary is not a "run to completion" binary with a `main`, rather a library where the hosts calls to regulary. That concept is called `reactor` in WASM

alban:
I am wondering about the UX: does it give a specific error message if trying to run an old version of a gadget with this new version of `ig`?

burak-ok:
1. The `go:wasmexport` there should be no problem. This is just the new directive to export the function.
2. The `.WithStartFunctions("_initialize")` shouldn't also be a problem AFAIK. I don't know how it behaved in the previous versions of tinygo but this just annotates that the start function of the module is `_initialize` instead of `_start` (This is the requirement of the reactor concept)

```bash
$ burak/wasm/tinygo.0.35.0 ✓ ➭ go run --exec "sudo -E" ./cmd/ig run trace_exec:v0.37.0
RUNTIME.CONTAINERNAME                 COMM                    PID        TID ARGS                   ERROR
reverent_elbakyan                     sh                    13932      13932 /bin/sh                  
reverent_elbakyan                     date                  13956      13956 /bin/date foo bar             
```

burak-ok:
> So, for now, tinygo gives us compatibility with both the old and new way of exporting function.
> Do they state until when they will keep this?

There is no compatibility AFAIK. If you try to compile old gadgets with `1.25` it will fail (I think at least). The simple test output I showed was running a compiled `1.24` WASM layer with the modifications in our wazero host settings.

So the compiliation might fail or not work. But already compiled gadgets with `tinygo 1.24` should still work

burak-ok:
> > > So, for now, tinygo gives us compatibility with both the old and new way of exporting function.
> > > Do they state until when they will keep this?
> > 
> > 
> > There is no compatibility AFAIK. If you try to compile old gadgets with `1.25` it will fail (I think at least). The simple test output I showed was running a compiled `1.24` WASM layer with the modifications in our wazero host settings.
> 
> What do you mean by 1.24 and 1.25? 

Sorry, my brain was at Go 1.24 :/
I meant tinygo 0.34 and 0.35

> So, if an user has an old gadget still using the old syntax, they would be able to run it with a newer ig (using tinygo 0.35), but if they try to compile it again they would get an error message?

I think no compilation error, there might be an error when the gadget runs, that it can't find the exported functions that it expects

docs/gadget-devel/hello-world-gadget-wasm.md
```diff
@@ -21,20 +21,20 @@ import (
 	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
 )
 
-//export gadgetInit
-func gadgetInit() int {
+//go:wasmexport gadgetInit
+func gadgetInit() int32{
```
mauriciovasquezbernal, line 25:
```suggestion
func gadgetInit() int32 {
```

docs/gadget-devel/hello-world-gadget-wasm.md
```diff
@@ -175,8 +175,8 @@ add a field that contains a human readable representation of the event.
 The `gadgetInit` functions now looks like:
 
 ```go
-//export gadgetInit
-func gadgetInit() int {
+//go:wasmexport gadgetInit
+func gadgetInit() int32{
```
mauriciovasquezbernal, line 179:
```suggestion
func gadgetInit() int32 {
```

mauriciovasquezbernal (approved):
LGTM. I'm aware of the IG vs gadget compatibility issues, I'll create an issue soon to go deeper on that, but it doesn't block this issue. Thanks Burak for handling this!

PR #4003, opened 2025-02-07
service-oci: Introduce "sendMutex" to synchronize access to runGadget.Send

### LINKED ISSUE #3997: Performance: kubectl-gadget hangs under high load
While tracing in a high load we end up in a situation where `kubectl-gadget` seems to be hanged and no new events are reported:

- Start DNS gadget:
```
kubectl gadget run trace_dns -p mypod
```
- Create a pod:
```
kubectl -n default run mypod -it --image=wbitt/network-multitool -- /bin/sh
```
- Perform parallel requests:
```
seq 1 1000 | xargs -n1 -P50 -I{} nslookup kubernetes.default.svc.cluster.local.
```
- After a certain point we should stop seeing new events and only warnings:

```
minikube-docker                kube-system                    coredns-7db6d8ff4d-7jr7c       coredns                        p/kube-system/coredns-7db6d8ff4d-7jr7c:53 p/default/mypod:57224                    coredns             1125570    1125570 R  AAAA             kubernetes.default.svc.clust… Success                                    0ns
minikube-docker                                                                                                             p/kube-system/coredns-7db6d8ff4d-7jr7c:53 p/default/mypod:57224                                              0          0 R  AAAA             kubernetes.default.svc.clust… Success                                    0ns
minikube-docker                                                                                                             s/kube-system/kube-dns:53                 p/default/mypod:57224                                              0          0 R  AAAA             kubernetes.default.svc.clust… Success                                    0ns
minikube-docker                default                        mypod                          mypod                          s/kube-system/kube-dns:53                 p/default/mypod:57224                    isc-net-0000        1128577    1128583 R  AAAA             kubernetes.default.svc.clust… Success                               234.52µs
WARN[0024] minikube-docker      | reading perf ring buffer: lost 58 samples 
WARN[0024] minikube-docker      | reading perf ring buffer: lost 2 samples 
WARN[0024] minikube-docker      | reading perf ring buffer: lost 6 samples 

```

## Impact

Not able to trace events anymore

## Expected behavior

gadget should continue tracing even if there are some error for previous events

## Additional information

- Since this is happening only for `kubectl-gadget` perhaps something to do it `grpc`? 
- Also, I hit it while testing changes for https://github.com/inspektor-gadget/inspektor-gadget/pull/3995 but was able to reproduce on previous versions as well. It seems since we are able to handle more load the issue become prominent after the performance fix.

mqasimsarfraz:
`runGadget.Send` uses SendMsg to send messages which isn't safe to use accross multiple goroutines. This change fixes the issue by introducing a lock.

> It is safe to have a goroutine calling SendMsg and another goroutine calling RecvMsg on the same stream at the same time, but it is not safe to call SendMsg on the same stream in different goroutines.
ref: https://github.com/grpc/grpc-go/blob/master/stream.go#L1550

Fixes #3997 

## Testing done

Not able to reproduce #3997


flyth:
Good find!

We already have "locking" in place by using a channel in the first place. So rather than adding a separate lock, I'd suggest re-using that channel (by defining it a bit earlier and re-using it for the logger's send() func).

mqasimsarfraz:
> We already have "locking" in place by using a channel in the first place. So rather than adding a separate lock, I'd suggest re-using that channel (by defining it a bit earlier and re-using it for the logger's send() func).

thanks, done. Can you please have another look?

pkg/gadget-service/service-oci.go
```diff
@@ -136,9 +136,15 @@ func (s *Service) RunGadget(runGadget api.GadgetManager_RunGadgetServer) error {
...
 	}
 
+	// Create payload buffer
+	outputBuffer := make(chan *api.GadgetEvent, s.eventBufferLength)
+
 	// Create a new logger that logs to gRPC and falls back to the standard logger when it failed to send the message
 	logger := logger.NewFromGenericLogger(&Logger{
-		send:           runGadget.Send,
+		send: func(event *api.GadgetEvent) error {
+			outputBuffer <- event
```
flyth, line 145: This is again a bit dangerous - I wonder if we should add the severity to the callback to decide on what to do.

As it is right now, that line is blocking if the buffer is full - hence it might cause the caller to block and so on. Down in the message pump we have the `select`/`default` construct that makes it non-blocking; if the buffer is full, the default will be used. I think the "best" approach would be to do the select/default and then in the default log locally on the server as fallback.

mauriciovasquezbernal (approved):
The heavy lifting was already done by @flyth while reviewing it, but I think it's good to go.

Thanks @mqasimsarfraz for investigating those nasty issues.

PR #4005, opened 2025-02-07
ci: Update unit test kernel versions

mauriciovasquezbernal:
Follow versions used in https://github.com/cilium/ci-kernels/

These are the new versions pushed on https://github.com/mauriciovasquezbernal/ci-kernels/actions/runs/13205395012.


burak-ok (approved):
Makes sense and LGTM

Issue #4008, opened 2025-02-07
tests: TestPerf is flaky


mauriciovasquezbernal:
[TestPerf](https://github.com/inspektor-gadget/inspektor-gadget/blob/6aa45d4c7094603b103641670e33a9ac3c4aa651/pkg/operators/wasm/wasm_test.go#L618) introduced in #3725 is failing in some ci runs:

ref https://github.com/inspektor-gadget/inspektor-gadget/pull/3916#issuecomment-2608109663

https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/13197378701/job/36841561987

Okabe-Rintarou-0:
Hello~I've checked the code and logs, it was due to perfReader.Read() will sometime return os.ErrDeadlineExceeded:
```go
func (p PerfReader) Read() ([]byte, error) {
	var buf bufPtr

	ret := perfReaderRead(uint32(p), uint32(uintptr(unsafe.Pointer(&buf))))
	switch ret {
	case 0:
		bytes := buf.bytes()
		buf.free()
		return bytes, nil
	case 1:
		return nil, errors.New("reading perf reader record")
	case 2:
		return nil, os.ErrDeadlineExceeded
	default:
		return nil, fmt.Errorf("bad return value: expected 0, 1 or 2, got %d", ret)
	}
}
```

So that the assertion here will fail:
```go
buf, err := perfReader.Read()
if err != nil {
	api.Errorf("reading perf record")
	return 1
}
```

What about detect these kind of error and then retry? I mean check if `errors.Is(err, os.ErrDeadlineExceeded)`

Okabe-Rintarou-0:
BTW, i've signed up for the lfx mentorship project. I'd like to assign myself for some good first issues. I think these one would be a good start for me :).

burak-ok:
[...]

PR #4014, opened 2025-02-09
fix: fix flaky TestPerf(#4008)

Okabe-Rintarou-0:
# fix flaky TestPerf(#4008)

I've added some retries for buf, err = perfReader.Read(), in case that os.ErrDeadlineExceeded occurs.

## How to use

## Testing done


mauriciovasquezbernal (approved):
This solution makes sense, thanks a lot. I'll merge once the CI is green.

PR #4025, opened 2025-02-12
fix: fix flaky TestPerf

Okabe-Rintarou-0:
# Fix flaky TestPerf

trigger open syscall hooks manually to completely fix the flaky test


mauriciovasquezbernal:
Events are already generated on https://github.com/inspektor-gadget/inspektor-gadget/blob/e8b65625670e93fb39f1ae38db4299e0212518bb/pkg/operators/wasm/wasm_test.go#L648-L651

Please check #4027

PR #4028, opened 2025-02-12
pkg/operators: Simplify perf buffer testing.

eiffel-fl:
We now generate the events in guest instead of generating them in host.
    This avoid having timing issues and flakyness.
    
    Fixes: 7386530c3378 ("operators/wasm: Add functions to interact with perf buffer.")

### DIFF
--- a/pkg/operators/wasm/testdata/perf/program.go
+++ b/pkg/operators/wasm/testdata/perf/program.go
@@ -54,6 +54,11 @@ func gadgetStart() int32 {
 		return 1
 	}
 
+	for range 10 {
+		// Let's generate some events by calling indirectly the write() syscall.
+		api.Infof("testing perf array")
+	}
+
 	err = perfReader.Pause()
 	if err != nil {
 		api.Errorf("pausing perf reader")
--- a/pkg/operators/wasm/testdata/perf/program.bpf.c
+++ b/pkg/operators/wasm/testdata/perf/program.bpf.c
@@ -14,7 +14,8 @@ struct {
 	__uint(value_size, sizeof(__u32));
 } events SEC(".maps");
 
-static __always_inline int trace_enter(struct syscall_trace_enter *ctx)
+SEC("tracepoint/syscalls/sys_enter_write")
+int test_write_e(struct syscall_trace_enter *ctx)
 {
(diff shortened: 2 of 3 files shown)


mauriciovasquezbernal:
It still fails with:

```
2025-02-12T15:53:16.1825780Z === NAME  TestPerf
2025-02-12T15:53:16.1826005Z     wasm_test.go:644: 
2025-02-12T15:53:16.1826998Z         	Error Trace:	/home/runner/work/inspektor-gadget/inspektor-gadget/pkg/operators/wasm/wasm_test.go:644
2025-02-12T15:53:16.1827351Z         	Error:      	Received unexpected error:
2025-02-12T15:53:16.1828711Z         	            	starting operators: starting operator "oci": starting operator "wasm": gadgetStart failed
2025-02-12T15:53:16.1828955Z         	Test:       	TestPerf
2025-02-12T15:53:16.1829247Z         	Messages:   	running gadget
2025-02-12T15:53:16.1829401Z --- FAIL: TestPerf (0.93s)
```

pkg/operators/wasm/testdata/perf/program.go
```diff
@@ -56,25 +55,25 @@ func gadgetStart() int32 {
 		return 1
 	}
 
+	for range 10 {
+		// Generate some events.
+		// While testing, this syscall returns an errno but the events are
+		// generated.
+		syscall.Open("/tmp/test-perf-array", syscall.O_CREAT|syscall.O_WRONLY|syscall.O_TRUNC, 0)
```
alban, line 62: I don't think it generates any event. The ebpf program is hooked on `tracepoint/syscalls/sys_enter_open` and `tracepoint/syscalls/sys_enter_openat` but this `syscall.Open` is running in the wasm module, so it can't run syscalls on the host kernel.

I see the tinygo `syscall.Open()` function is implemented with `libc_open()`:
https://github.com/tinygo-org/tinygo/blob/v0.25.0/src/syscall/syscall_libc.go#L68

I am not sure what it does, but still, it is running in the wasm sandbox, and I don't see where it would be connected to the host to perform syscalls on the host.

mauriciovasquezbernal (approved):
I think it's a good idea to use write instead of open(). 

LGTM, thanks!

PR #4035, opened 2025-02-12
Add warning if gadget is run with a different IG version

mauriciovasquezbernal:
Ideally Inspektor Gagget should allow running gadgets built with different ig versions. However, we haven't been able to implement this feature. To avoid user confusion, print a warning if the gadget is run with a different version of IG than the one it was built with.

### Testing 

1. Compile ig with a fake 0.98.0 version

```bash 
$ make VERSION=v0.98.0 
```

2. Build any gadget using this verison

```bash 
$ sudo ig image build . -t hello-world
```

3. Run the gadget

```bash 
$ sudo ig run hello-world --verify-image=false
RUNTIME.CONTAINERNAME                      PID COMM             FILENAME
```

No warning is printed on this case 

4. Compile a fake 0.99.0 version of ig 

```bash 
$ make VERSION=v0.99.0 ig
```

5. Run the gadget again 

```bash 
$ sudo ig run hello-world --verify-image=false
WARN[0001] This gadget was built with ig v0.98.0 and it's being run with v0.99.0. Gadget could be incompatible
WARN[0001] This gadget was built with ig v0.98.0 and it's being run with v0.99.0. Gadget could be incompatible
RUNTIME.CONTAINERNAME                      PID COMM             FILENAME
```

The warning is now printed.

### DIFF
--- a/pkg/operators/oci-handler/oci.go
+++ b/pkg/operators/oci-handler/oci.go
@@ -21,14 +21,18 @@ import (
 	"io"
 	"strings"
 
+	"github.com/blang/semver"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/inspektor-gadget/inspektor-gadget/internal/version"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/environment"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	apihelpers "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api-helpers"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
@@ -46,6 +50,10 @@ const (
 	verifyImage             = "verify-image"
 	publicKeys              = "public-keys"
 	allowedGadgets          = "allowed-gadgets"
+)
+
+const (
+	builderVersionAnnotation = "io.inspektor-gadget.builder.version"
 )
 
 type ociHandler struct {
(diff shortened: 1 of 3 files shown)


pkg/operators/oci-handler/oci.go
```diff
@@ -201,6 +209,33 @@ func (o *OciHandlerInstance) ExtraParams(gadgetCtx operators.GadgetContext) api.
 	return o.extraParams
 }
 
+func warnBuilderVersion(manifest *v1.Manifest, logger logger.Logger) {
```
flyth, line 212:
```suggestion
func checkBuilderVersion(manifest *v1.Manifest, logger logger.Logger) {
```

"warn*" sounds like it should warn all the time (e.g. check already done, now warn).

flyth (approved):
Great change!

It would be great if we could only show it only once, but that's already true for a couple of messages. (The user isn't aware that we're first pulling gadget information and then run it).

Also, I was thinking that maybe when using shortcuts (only "trace_exec" etc.) and our official repo in the url without a tag we should automatically append the current IG version as tag. But that might clash with untagged gadgets that are "in development" on the local machine.

mauriciovasquezbernal:
> It would be great if we could only show it only once, but that's already true for a couple of messages. (The user isn't aware that we're first pulling gadget information and then run it).

Unfortunately I don't see an easy way to fix it. Perhaps the `InstantiateDataOperator` should have a way to know if it's being called from GetGadgetInfo or from RunGadget. 

> Also, I was thinking that maybe when using shortcuts (only "trace_exec" etc.) and our official repo in the url without a tag we should automatically append the current IG version as tag. But that might clash with untagged gadgets that are "in development" on the local machine.

Perhaps it's not that a bad idea. Currently it defaults to `latest`, but with the version incompatibilities we have, it is becoming confusing for users.

mauriciovasquezbernal:
> * Old gadgets can be used with newer ig?
> 
> * New gadgets can be used with older ig?
> 
> * Or only a 1on1 compatibility?

Opened #4047 to further discuss it.

mauriciovasquezbernal:
> > Also, I was thinking that maybe when using shortcuts (only "trace_exec" etc.) and our official repo in the url without a tag we should automatically append the current IG version as tag. But that might clash with untagged gadgets that are "in development" on the local machine.
> 
> Perhaps it's not that a bad idea. Currently it defaults to `latest`, but with the version incompatibilities we have, it is becoming confusing for users.

I tried it out but IMO it creates further confusion. For instance, `ig image build foo` will produce foo:v0.3.4, but the version of the gadget could be totally different to the one from ig (specially for external gadgets). It's true we could have some logic to apply it only for official gadgets, or only when running a gadget, but it seems too complex to me. I think we can tackle this issue in a different way: #4049

Issue #3835, opened 2024-12-23
[Feat:Improve Unit Test Coverage wherever required ]


yashpandey06:
https://mentorship.lfx.linuxfoundation.org/project/d3a1a899-1ca0-4e10-a402-01ef6fde26f2

---

# Improve Unit Test Coverage for Inspektor Gadget

## Current Situation

Inspektor Gadget currently has limited unit test coverage. Some critical code paths and edge cases remain untested, which could lead to undetected bugs and reduced code reliability.

## Impact

The lack of comprehensive unit tests in Inspektor Gadget can result in:  
1. Increased risk of regressions during future code changes.  
2. Difficulty in identifying and fixing bugs efficiently.  
3. Reduced confidence in the stability and reliability of the codebase.  

## Ideal Future Situation

Inspektor Gadget should have robust and comprehensive unit tests that cover all critical code paths and edge cases. This would:  
1. Increase the test coverage as much as possible.  
2. Ensure that the codebase is stable and resilient to changes.  
3. Facilitate faster development and debugging by identifying issues early.  

## Implementation Options

1. **Audit Existing Tests**: Identify gaps in the current test coverage and prioritize adding tests for untested or under-tested code.  
2. **Add Unit Tests**: Write detailed unit tests for Inspektor Gadget, covering:  
   - Edge cases.  
   - Common use cases.  
   - Potential failure scenarios.  
3. **Use Coverage Tools**: Utilize tools like `go test -cover` to measure coverage and identify areas that need improvement.  

## Additional Information

- Testing should adhere to the existing project standards and use consistent naming conventions.  
- Mocking and stubbing may be used where necessary to isolate unit tests from external dependencies.  
- Ensure that the added tests are maintainable and easy to understand.

### Priority 1
- [ ] [gadgets/trace_capabilities](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/trace_capabilities)
- [ ] [gadgets/trace_bind](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/trace_bind)
- [ ] [gadgets/snapshot_socket](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/snapshot_socket)
- [ ] [gadgets/trace_exec](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/trace_exec)
[...]

PR #4042, opened 2025-02-13
[test]:improve coverage for kallsysms.go in pkg/kallsyms

Dhiren-Mhatre:
# [Add test cases to improve kallsyms package coverage from 40% to 73%]

This PR adds several new test cases to improve code coverage for the kallsyms package. 

## How to use
Reviewers can validate this PR by:
1. Running the full test suite: `go test ./pkg/kallsyms/`
2. Verifying the increased coverage: `go tool cover -func=coverage.out`

## Testing done
```bash
$ go test -coverprofile=coverage.out ./pkg/kallsyms/
ok  	github.com/inspektor-gadget/inspektor-gadget/pkg/kallsyms	0.482s
```
image

Part of [Feat:Improve Unit Test Coverage wherever required] #3835

### DIFF
--- a/pkg/kallsyms/kallsyms_test.go
+++ b/pkg/kallsyms/kallsyms_test.go
@@ -15,10 +15,14 @@
 package kallsyms
 
 import (
+	"errors"
+	"fmt"
 	"os"
 	"strings"
 	"testing"
 
+	"github.com/cilium/ebpf"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
@@ -36,21 +40,52 @@ func TestCustomKAllSymsInstructionPointer(t *testing.T) {
 	require.Nil(t, err, "NewKAllSymsFromReader failed: %v", err)
 
 	lookupByInstructionPointerTests := []struct {
+		name               string
 		instructionPointer uint64
 		expectedSymbol     string
 	}{
-		{0, "fixed_percpu_data"},
-		{0xffffffffb4231f39, "fixed_percpu_data"},
-		{0xffffffffb4231f40, "bpf_prog_fops"},
-		// TODO: is it correct? should it be bpf_prog_fops?
-		{0xffffffffb4231f41, "bpf_prog_fops"},
-		{0xffffffffb43723df, "bpf_prog_fops"},
-		{0xffffffffb43723e0, "socket_file_ops"},
-		{0xffffffffb43723e1, "socket_file_ops"},
+		{
+			name:               "exact_match_start",
+			instructionPointer: 0,
+			expectedSymbol:     "fixed_percpu_data",
+		},
+		{
+			name:               "middle_of_first_range",
+			instructionPointer: 0xffffffffb4231f39,
+			expectedSymbol:     "fixed_percpu_data",
+		},
+		{
+			name:               "exact_match_middle",
+			instructionPointer: 0xffffffffb4231f40,
+			expectedSymbol:     "bpf_prog_fops",
+		},
+		{
+			name:               "middle_of_second_range",
+			instructionPointer: 0xffffffffb4231f41,
+			expectedSymbol:     "bpf_prog_fops",
+		},
+		{
+			name:               "end_of_second_range",
+			instructionPointer: 0xffffffffb43723df,
+			expectedSymbol:     "bpf_prog_fops",
+		},
+		{
+			name:               "start_of_last_range",
+			instructionPointer: 0xffffffffb43723e0,
+			expectedSymbol:     "socket_file_ops",
+		},
+		{
+			name:               "after_last_symbol",
+			instructionPointer: 0xffffffffb43723e1,
+			expectedSymbol:     "socket_file_ops",
+		},
 	}
 	for _, tt := range lookupByInstructionPointerTests {
-		require.Equal(t, tt.expectedSymbol, kAllSyms.LookupByInstructionPointer(tt.instructionPointer),
-			"LookupByInstructionPointer(0x%x)", tt.instructionPointer)
+		t.Run(tt.name, func(t *testing.T) {
+			result := kAllSyms.LookupByInstructionPointer(tt.instructionPointer)
+			assert.Equal(t, tt.expectedSymbol, result,
+				"LookupByInstructionPointer(0x%x)", tt.instructionPointer)
+		})
 	}
 }
 
@@ -66,36 +101,42 @@ func TestCustomKAllSymsParsing(t *testing.T) {
 	}, "\n")
 
 	tests := []struct {
+		name       string
 		symbol     string
 		expectErr  bool
 		expectAddr uint64
 		expectKmod string
 	}{
 		{
+			name:       "core_kernel_symbol",
 			symbol:     "bpf_prog_fops",
 			expectErr:  false,
 			expectAddr: 0xffffffffb4231f40,
 			expectKmod: "",
 		},
 		{
+			name:       "another_core_symbol",
 			symbol:     "socket_file_ops",
 			expectErr:  false,
 			expectAddr: 0xffffffffb43723e0,
 			expectKmod: "",
 		},
 		{
+			name:       "veth_module_symbol",
 			symbol:     "veth_init",
 			expectErr:  false,
 			expectAddr: 0xffffffffc1b26010,
 			expectKmod: "veth",
 		},
 		{
+			name:       "netem_module_symbol",
 			symbol:     "netem_module_init",
 			expectErr:  false,
 			expectAddr: 0xffffffffc1cb5010,
 			expectKmod: "sch_netem",
 		},
 		{
+			name:       "nonexistent_symbol",
 			symbol:     "abcde_bad_name",
 			expectErr:  true,
 			expectAddr: 0,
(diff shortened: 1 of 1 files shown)


Dhiren-Mhatre:
@burak-ok  I noticed that my PR checks failed, but I didn’t see any issues in the changes I made. Could you help me confirm if the failure is due to my changes in the test file or if there’s another issue?

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+		},
+		{
+			name:        "invalid_address",
+			input:       "INVALID_ADDR A fixed_percpu_data",
+			expectedErr: "strconv.ParseUint",
+		},
+		{
+			name:        "malformed_line",
+			input:       "0000000000000000",
+			expectedErr: "line \"0000000000000000\" has less than 3 fields",
```
burak-ok, line 249: The error message is the same as the first test case. Is this a different codepath?

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+		ip       uint64
+		expected string
+	}{
+		{
+			name:     "exact_match_start",
+			ip:       0x0,
+			expected: "first_symbol",
+		},
+		{
+			name:     "before_first_symbol",
```
burak-ok, line 286: This is the exact same `ip` as `exact_match_start` and this test is indeed not before the first symbol

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			reader := strings.NewReader(tt.input)
+			_, err := NewKAllSymsFromReader(reader)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), tt.expectedErr)
+		})
+	}
+}
+
+func TestLookupByInstructionPointerEdgeCases(t *testing.T) {
```
burak-ok, line 263: The are some edge cases that are already tested in `TestCustomKAllSymsInstructionPointer`. Could you instead of writing a new function update that?
Especially with the helpful test names you did in this function

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			reader := strings.NewReader(tt.input)
+			_, _, err := kernelSymbolAddressFromReader(reader, tt.symbol)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), tt.expectedErr)
+		})
+	}
+}
+
+func TestKernelSymbolAddressFromReaderModuleSuccess(t *testing.T) {
```
burak-ok, line 364: This kind of test is already done in `TestCustomKAllSymsParsing`

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+	require.Equal(t, "test_module", mod)
+}
+
+func TestSymbolExistsWithEmptySymbol(t *testing.T) {
+	exists := SymbolExists("")
+	require.False(t, exists)
+}
+
+// Test for parsing various module formats
+func TestKernelSymbolAddressFromReaderModuleFormats(t *testing.T) {
```
burak-ok, line 379: These kind of tests are already done in other functions. But IMO I think it is good to have descriptive tests targeting something specific. So I think this is okay

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+	}{
+		{
+			name:     "no_module",
+			input:    "ffffffffc1234567 T test_symbol",
+			symbol:   "test_symbol",
+			wantMod:  "",
+			wantAddr: 0xffffffffc1234567,
+		},
+		{
+			name:     "with_module",
```
burak-ok, line 395: This test is kind of the same as `module_with_with_underscore` since the modile name has an underscore here too

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "mock resolver error")
+
+	err = specUpdateAddresses(
+		[]symbolResolver{failResolver},
+		spec,
+		[]string{"symbol1"},
+	)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "mock resolver error")
```
burak-ok, line 482: Since you already mocked the symbol resolver: Could you add a test where the resolving works?

pkg/kallsyms/kallsyms_test.go
```diff
@@ -224,3 +226,258 @@ func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
...
+
+	err := specUpdateAddresses(
+		[]symbolResolver{failResolver},
+		spec,
+		[]string{"symbol1"},
+	)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "mock resolver error")
+
+	err = specUpdateAddresses(
```
burak-ok, line 476: Please add a description that this is specifically designed to test if the error cache works. Also would be good to somehow differentiate between the cached error and the non cached one

pkg/kallsyms/kallsyms_test.go
```diff
@@ -198,29 +237,147 @@ func TestRealKAllSymsParsing(t *testing.T) {
 	}
 }
 
-// BenchmarkKernelSymbolAddressJITParsing searches for the symbol while
-// reading the kallsyms file
-func BenchmarkKernelSymbolAddressJITParsing(b *testing.B) {
```
burak-ok, line 203: This function is used for benchmarking (see `b *testing.B`). This function shouldn't be removed

pkg/kallsyms/kallsyms_test.go
```diff
@@ -198,29 +237,147 @@ func TestRealKAllSymsParsing(t *testing.T) {
...
+			_, err := NewKAllSymsFromReader(reader)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), tt.expectedErr)
+		})
 	}
 }
 
-// BenchmarkKernelSymbolAddressPreLoading loads the full kallsyms before
-// looking up the symbol
-func BenchmarkKernelSymbolAddressPreLoading(b *testing.B) {
```
burak-ok, line 215: This function is used for benchmarking (see `b *testing.B`). This function shouldn't be removed

pkg/kallsyms/kallsyms_test.go
burak-ok: Seems like we would somehow need to create instances `ebpf.VariableSpec`. Let's leave this for later

pkg/kallsyms/kallsyms_test.go
```diff
@@ -198,29 +237,147 @@ func TestRealKAllSymsParsing(t *testing.T) {
...
+
+	t.Run("all_resolvers_fail", func(t *testing.T) {
+		resolver1 := &mockSymbolResolver{
+			resolveFunc: func(symbol string) (uint64, error) {
+				return 0, fmt.Errorf("first resolver failed")
+			},
+		}
+		resolver2 := &mockSymbolResolver{
+			resolveFunc: func(symbol string) (uint64, error) {
+				return 0, fmt.Errorf("second resolver failed")
```
burak-ok, line 320: This error message could be defined inside the `t.Run("all_resolvers_fail", func(t *testing.T) {` function as a const. Then the variable can be used in the `.Contains` check below

pkg/kallsyms/kallsyms_test.go
```diff
@@ -198,29 +237,147 @@ func TestRealKAllSymsParsing(t *testing.T) {
...
+			},
+		}
+
+		err := specUpdateAddresses(
+			[]symbolResolver{resolver1, resolver2},
+			spec,
+			[]string{"missing_symbol"},
+		)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "second resolver failed")
```
burak-ok, line 336: Descriptive test case messages would be nice here. Something like `Only the error of the last resolve should be returned`

pkg/kallsyms/kallsyms_test.go
```diff
@@ -198,29 +237,147 @@ func TestRealKAllSymsParsing(t *testing.T) {
...
+
+		// Verify error is cached
+		cachedErr, exists := triedGetAddr["missing_symbol"]
+		require.True(t, exists)
```
burak-ok, line 340: Another descriptive test case message here

burak-ok (commented):
The tests itself look good. Could you review if you need `.require` instead of `.assert` everywhere? Also the test case messages can be more descriptive

Dhiren-Mhatre:
I've converted require to assert where appropriate and added descriptive messages throughout the test file. Thanks for the review!

burak-ok (approved):
LGTM. Thanks for the contribution :)

Issue #4057, opened 2025-02-14
[RFE] Reduce maps size in trace exec


dorser:
## Current situation

The trace_exec gadget allocates about 170MBs for its maps. Given the natures of the syscalls traced in trace_exec, we can reduce drastically the number of allocated entries (currently set to 10K).

## Implementation options
Number of entries can be reduced to per CPU (with a given factor although I think not necessary). 

In addition, we can revisit the size of fields like cwd to also reduce the event size which is currently at 4k

mauriciovasquezbernal:
Fixed by #4058

PR #4058, opened 2025-02-14
trace_exec: reduce maps size

dorser:
# Reduce Memory Usage in `trace_exec` Gadget

This resolves issue #4057.

This PR changes the eBPF maps in the `trace_exec` gadget from `BPF_MAP_TYPE_HASH` to `BPF_MAP_TYPE_PERCPU_HASH` and reduces their max_entries from `10240` to `1`. These modifications aim to lower memory usage by drastically shrinking the size of each map, which can help in resource-constrained environments or when the data stored in these maps is minimal.

## How to use
```
 sudo -E ig image build $IG_SOURCE_PATH/gadgets/trace_exec -t trace_exec
 sudo -E ig run trace_exec:latest --host --verify-image=false
```

## Testing done
Built `ig` cli and the gadget program locally on a Linux environment with kernel version 6.8. I was able to successfully star the gadget using `ig run` (as described above):
image

I wrote a "benchmarking" program that calls `execve("/bin/true", args, envp);` to test exec contention. The benchmarking program executed 10K calls from multiple threads from all available CPUs. I observed contention at 8 threads - thus setting the `max_entries` to 8 which resulted at no loss.

In addition, observing the memory using `ig top ebpf` showing the map memory is at 3MB~.

### DIFF
--- a/gadgets/trace_exec/program.bpf.c
+++ b/gadgets/trace_exec/program.bpf.c
@@ -79,14 +79,14 @@ static const struct event empty_event = {};
 // - tracepoint/syscalls/sys_exit_execve is always called
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__uint(max_entries, 10240);
+	__uint(max_entries, 1024);
 	__type(key, pid_t);
 	__type(value, struct event);
 } execs SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__uint(max_entries, 10240);
+	__uint(max_entries, 1024);
 	__type(key, pid_t);
 	__type(value, __u8);
 } security_bprm_hit_map SEC(".maps");


alban:
I don't think we can use per-cpu ebpf maps here because the scheduler could decide to switch to another task between the `enter_execve` and `exit_execve`. Then the task could be scheduled on another cpu.

dorser:
Totally makes sense. The idea was to "dynamically" allocate memory based on the number of CPUs.
I'm wondering if we can achieve something similar by setting the map sizes from the eBPF operator (Something like [this](https://github.com/inspektor-gadget/inspektor-gadget/blob/85925c77a41a0806f43e00d1c7fa579cfb5da16a/pkg/operators/ebpf/ebpf.go#L293C36-L293C54).

Maybe a "gadget parameter" to sepcify map sizes?

Can you elaborate on why the ring buffer is set to 1024 * 256? I personally feel comfortable with starting with 1024 for the max entires (using 64 cores * 16 threads as a rule of thumb to the "average" node size) knowning we might lose some samples on larger and very busy nodes.

gadgets/trace_exec/program.bpf.c
```diff
@@ -79,14 +79,14 @@ static const struct event empty_event = {};
 // - tracepoint/syscalls/sys_exit_execve is always called
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__uint(max_entries, 10240);
+	__uint(max_entries, 1024);
 	__type(key, pid_t);
 	__type(value, struct event);
 } execs SEC(".maps");
```
Sefi4, line 85: In addition to reducing map size, should we use BPF_F_NO_PREALLOC flag to disable pre allocation?
I guess it would be more effective on large maps, but if we're interested in highly efficient memory utilization, is there a reason not to use this flag?

Sefi4:
Should we also change events map to be 1024 * sizeof(struct event)?

dorser:
AFAIU BPF_F_NO_PREALLOC introduces data race conditions we are likely to hit in this gadget.

regarding the buffer size, I haven’t observed any loss (with WASM) when I benchmarked this so I’m good with its current size.

alban (approved):
LGTM.

Before this patch, sorted memory usage of individual maps:

```
```

### Image-based trace_exec:
```
19432  hash          ig_fa_pick_ctx   6368       ig
19437  array         .bss             8192       ig
19438  array         .rodata          24576      ig
19435  hash          mntnsset_e1ef9d  83168      ig
19436  percpu_array  bufs             131400     ig
19440  ringbuf       events           275800     ig
19433  queue         ig_fa_records    335200     ig
19430  hash          exec_args        690496     ig
19441  hash          security_bprm_h  918752     ig
19425  hash          containers       1127616    ig
19439  hash          execs            180311680  ig
```
### Builtin trace-exec:
```
19517  perf_event_array  events           352       ig
19512  hash              ig_fa_pick_ctx   6368      ig
19515  hash              gadget_mntns_fi  83168     ig
19513  queue             ig_fa_records    335200    ig
19510  hash              exec_args        690496    ig
19505  hash              containers       1127616   ig
19516  hash              execs            54269504  ig
```
### Builtin trace-exec with --paths:
```
19538  perf_event_array  events           352       ig
19532  hash              ig_fa_pick_ctx   6368      ig
19535  hash              gadget_mntns_fi  83168     ig
19533  queue             ig_fa_records    335200    ig
19530  hash              exec_args        690496    ig
19525  hash              containers       1127616   ig
19536  hash              execs            18069056  ig
```

dorser:
Should we do the change for built-in as well (Not sure given it's getting deprecated?)

Used your command to measure memory after this patch for the image-based gadget:
```
$ sudo bpftool map show -j | jq -r '.[] | {id: .id, type: .type, name: .name, bytes_memlock: .bytes_memlock, 
```
Results:
```
comms: (if .pids then (.pids | map(.comm) | unique | join(",")) else "" end)} | [.id, .type, .name, .bytes_memlock, .comms] | @tsv' | sort -k4n | column -t -s $'\t'
14  hash                           46880
3   hash                           46880
13  array         ringbuf_loss_co  328
69  hash          ig_fa_pick_ctx   7680
78  array         .bss             8192
2   prog_array    hid_jmp_table    8512
68  array         .rodata          12288
77  array         .rodata          24576
74  hash          security_bprm_h  84480
75  hash          gadget_mntns_fi  84480
12  ringbuf       ringbuf          275800
73  ringbuf       events           275800
11  hash          AT_skpid         330240
70  queue         ig_fa_records    335200
10  hash          AT_skcomm        363136
76  percpu_array  bufs             524616
67  hash          exec_args        753536
62  hash          containers       1141120
72  hash          execs            18296960
```

alban:
> Should we do the change for built-in as well (Not sure given it's getting deprecated?)

Yes, given it is not a complex patch, please do if you can. Thanks!

PR #4061, opened 2025-02-14
Improve gadget version skew warning

mauriciovasquezbernal:
Do not print anything if parsing the version fails. It means the the gadget could have been built with a development version of Inspektor Gadget, hence it'll create a lot of noise when using the :latest version.

--- 

Annoying message is:

```bash 
$ sudo ig run trace_exec:latest  --pull always
WARN[0002] parsing builder version: Invalid character(s) found in major number "19bcf49"
WARN[0004] parsing builder version: Invalid character(s) found in major number "19bcf49"
```

Fixes: c04bab7e12e2 ("Add warning if gadget is run with a different IG version")


alban:
The `Warnf` could be a `Debugf` with a comment explaining it can happen in development versions.

alban (approved):

PR #4062, opened 2025-02-14
operators/ebpf: Do not create kstack map before

mauriciovasquezbernal:
Avoid creating the kstack map before and reuse the one existing on the gadget.

cc @alban


alban (approved):
LGTM

The previous implementation was done in:
https://github.com/inspektor-gadget/inspektor-gadget/pull/2671

I can't find a reason why the map was created separately.

Issue #4069, opened 2025-02-16
[RFE] Reduce "gadget_sockets" and "enriched_fsnotify_events" maps size


yaeltz:
## Current situation

There is currently an allocation of about 130MBs for gadget_sockets map, and of 42MB for enriched_fsnotify_events. the number of allocated entries (currently set to 16384  and 10240).

ig top ebpf -o columns=NAME,MAPMEMORY --sort -MAPMEMORY  

![Image](https://github.com/user-attachments/assets/2749d76f-4981-4dd0-ba96-4a0fbae3f8df)

![Image](https://github.com/user-attachments/assets/00bbc921-8dfa-449e-b98a-247a84cbd4e3)

## Impact

We see a high memory consumption of gadgets and we would like to reduce memory footprint as much as possible

## Ideal future situation

Reduce memory consumption as much as possible

## Implementation options

Is it possible to reduce the number of records (16384 and 10240) ?
Is it possible to reduce the size of fields, for example SE_PATH_MAX?

alban:
#4076 and #4083 are both merged

Issue #1882, opened 2023-07-24
gadgets: Implement mechanism for logging in eBPF programs


mauriciovasquezbernal:
Currently the only alternative for logging in eBPF programs that we have available is `bpf_printk`. It's far from ideal as the logging is printed to a location not accessible by users. We can implement a more robust logging engine by sending logs to user space, where Inspektor Gadget can parse and relay them to the user. 

### Implementation options.

We can use a ring buffer to send logs to user space, the event we send should contain the format message and the arguments:

```c
struct gadget_log_event_t {
   u64 args[4]; // TODO: max number of args to support
   u64 gadget_id; // used by user space to proxy the message to the right client
   unsigned char level; // Enum?
  char fmt[MAX_FMT]; // define max
};
```

The perf event array (or ring buffer) is unique for the whole Inspektor Gadget and shared by all gadgets, each gadget defines a gadget id constant that is used by user space to forward messages to the right client.  

All of this could be self contained in a helper file, that will provide:
- gadget id constant 
- Perf array buffer definition
- A `gadget_log(int level, char *msg, u64 args...)` like helpers

alban:
I think we have to be careful for ⁠future [support for API changes](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/002-containerized-gadgets.md#support-inspektor-gadget-api-changes).

I wonder if IG could automatically replace calls to `bpf_printk` by an implementation of your `gadget_log` with a perf ring buffer map in a similar way we patch `bpf_ktime_get_boot_ns`. That would reduce the IG-specific API surface that we would have to maintain. But it is probably more complex than `bpf_ktime_get_boot_ns` because we have to handle different arguments.

We could also use a bpf extension: at least the IG-specific API would be defined by the function without exposing the internal detail of the perf ring buffer. But we would need to be able to pass the argument to the bpf extension. I don't know if it is possible, but there was some research at https://github.com/inspektor-gadget/inspektor-gadget/pull/1097#issuecomment-1594595733.

mauriciovasquezbernal:
> I wonder if IG could automatically replace calls to bpf_printk by an implementation of your gadget_log with a perf ring buffer map in a similar way we patch bpf_ktime_get_boot_ns. That would reduce the IG-specific API surface that we would have to maintain. But it is probably more complex than bpf_ktime_get_boot_ns because we have to handle different arguments.

[...]

PR #4076, opened 2025-02-17
fsnotify: reduce map size

alban:
Reduce max_entries of enriched_fsnotify_events map from 10240 to 1024. This reduces the size from 42 MiB to 4 MiB.

Part of #4069

## How to use

No changes.

## Testing done

Before:
```
19598  hash          ig_fa_pick_ctx   6368      ig
19605  hash          fsnotify_remove  6368      ig
19602  hash          fsnotify_insert  6912      ig
19608  array         .bss             8192      ig
19607  array         .rodata          12288     ig
19609  percpu_array  gadget_heap      41288     ig
19601  hash          mntnsset_f6e864  83168     ig
19604  percpu_array  bufs             131400    ig
19606  ringbuf       events           275800    ig
19599  queue         ig_fa_records    335200    ig
19596  hash          exec_args        690496    ig
19591  hash          containers       1127616   ig
19603  hash          enriched_fsnoti  44599168  ig
```

After:
```
19624  hash          ig_fa_pick_ctx   6368     ig
19631  hash          fsnotify_remove  6368     ig
19630  hash          fsnotify_insert  6912     ig
19635  array         .bss             8192     ig
19634  array         .rodata          12288    ig
19628  percpu_array  gadget_heap      41288    ig
19627  hash          mntnsset_1bb8c0  83168    ig
19629  percpu_array  bufs             131400   ig
19633  ringbuf       events           275800   ig
19625  queue         ig_fa_records    335200   ig
19622  hash          exec_args        690496   ig
19617  hash          containers       1127616  ig
19632  hash          enriched_fsnoti  4466560  ig
```

cc @yaeltz

### DIFF
--- a/gadgets/fsnotify/program.bpf.c
+++ b/gadgets/fsnotify/program.bpf.c
@@ -154,7 +154,7 @@ struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, void *); // struct fsnotify_event *event
 	__type(value, struct enriched_event);
-	__uint(max_entries, 10240);
+	__uint(max_entries, 1024);
 } enriched_fsnotify_events SEC(".maps");
 
 const volatile u64 tracer_group = 0;


gadgets/fsnotify/program.bpf.c
```diff
@@ -154,7 +154,7 @@ struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, void *); // struct fsnotify_event *event
 	__type(value, struct enriched_event);
-	__uint(max_entries, 10240);
+	__uint(max_entries, 1024);
```
blanquicet, line 157: I don't know but, isn't there any way we can profile the gadget with different workloads and verify that the new value is enough in most cases and the old one was too big?

alban: @dorser tried with a limit of 128 and could not fill the buffer by looping on `/bin/true`.

Of course it is not a guarantee if it's enough on a different workload or with a slower process processing the inotify events, but 1024 seems fine to me.

dorser: `fsnotify_insert_event()` uses a [lock](https://github.com/torvalds/linux/blob/6537cfb395f352782918d8ee7b7f10ba2cc3cbf2/fs/notify/notification.c#L93) as well. However, I don't think my benchmark for `exec` applies here as the lock contention is probably different. Specfically for our use case, I'm pretty sure this is enough as we set a small number of watchers on files we expect to see little to no usage.

mauriciovasquezbernal: Checking the return value of `bpf_map_update_elem` and using #1882 would be a good way to tell user this map size is too small.

mauriciovasquezbernal:
A final thought on this (doesn't have to be implemented): Is it a good idea, or technically possible, to have the size of this map as an argument of the gadget? We could define a good default, but each user could tune it for their specific use case.

alban:
> A final thought on this (doesn't have to be implemented): Is it a good idea, or technically possible, to have the size of this map as an argument of the gadget? We could define a good default, but each user could tune it for their specific use case.

Why not. The ebpf operator already has a few instance parameters. We could add one like `--map-replace-max-entries enriched_fsnotify_events=2048`. AFAIR, max_entries is not used by the verifier to reject a program, so that should be technically doable to patch the CollectionSpec in the ebpf operator before instantiating it.

dorser:
Is it feasible to set the map size as a gadget parameter? That was my initial thought but I couldn’t get it to work.

dorser:
Oh patching the collectionspec. Yeah that’ll work

PR #4086, opened 2025-02-18
ustack operator: add goresym param

alban:
# ustack operator: add goresym param

Go symbol recovery tool: https://github.com/mandiant/GoReSym

stripped Go binaries don't have a symbol table but it can be recovered with GoReSym.

Example: the official kubelet binary is distributed without symbol table:
```
$ nm /tmp/kubelet
nm: /tmp/kubelet: no symbols
```

Note: on my computer, the `GoReSym` tool can take one minute to run on the `kubelet` binary, and use several GB of memory. So it cannot be run on the fly when we receive a stack trace from ebpf. Instead, this PR allows the user to prepare the symbol table in advance and pass it to the ustack operator.

The kubelet is 116MB and its goresym json file is 38MB.

## How to use

Use the ustack operator's parameters " --goresym=". As other parameters, you can use the `@` symbol to pass a file.

The file could be a single goresym json file or a array of them. Inspektor Gadget will use the "Build ID" to find the correct symbol table to use.

## Testing done

```
minikube start
minikube ssh sudo chmod +x /ig
minikube cp minikube:/var/lib/minikube/binaries/v1.26.1/kubelet /tmp/kubelet
wget https://github.com/mandiant/GoReSym/releases/download/v3.0.2/GoReSym-linux.zip
unzip GoReSym-linux.zip
chmod +x GoReSym
./GoReSym -d /tmp/kubelet > /tmp/kubelet.goresym
minikube cp ./ig minikube:/ig
minikube cp /tmp/kubelet.goresym minikube:/tmp/kubelet.goresym
minikube ssh "sudo /ig run albantest.azurecr.io/gadget/trace_capabilities:alban_user_stack_goresym2 --verify-image=false --host --print-ustack=true --goresym=@/tmp/kubelet.goresym -o yaml"
```

Output:
```yaml
proc:
  comm: kubelet
kstack: '[0]security_capable; [1]capable_wrt_inode_uidgid; [2]generic_permission;
  [3]proc_fd_permission; [4]inode_permission; [5]may_open; [6]path_openat; [7]do_filp_open;
  [8]do_sys_openat2; [9]__x64_sys_openat; [10]do_syscall_64; [11]entry_SYSCALL_64_after_hwframe; '
ustack: '[0]runtime/internal/syscall.Syscall6; [1]syscall.Syscall6; [2]syscall.openat;
  [3]os.openFileNolog; [4]os.OpenFile; [5]io/ioutil.ReadDir; [6]github.com/google/cadvisor/container/libcontainer.processStatsFromProcs;
  [7]github.com/google/cadvisor/container/libcontainer.(*Handler).GetStats; [8]github.com/google/cadvisor/container/raw.(*rawContainerHandler).GetStats;
  [9]github.com/google/cadvisor/manager.(*containerData).updateStats; [10]github.com/google/cadvisor/manager.(*containerData).housekeepingTick;
  [11]github.com/google/cadvisor/manager.(*containerData).housekeeping; [12]github.com/google/cadvisor/manager.(*containerData).Start.func1;
  [13]runtime.goexit; '
```


alban (commented):

| Benchmark suite | Current: b5e7216f42024ad271cc9fcdf20f409044b669c4 | Previous: bc18339c67db4982e02a64d8645876980b68784e | Ratio |
|-|-|-|-|
| `BenchmarkAllGadgetsWithContainers/container0/trace-dns` | `21631076` ns/op | `6447926` ns/op | `3.35` |
| `BenchmarkAllGadgetsWithContainers/container0/trace-network` | `20359946` ns/op | `4722844` ns/op | `4.31` |

CC: @alban

pkg/operators/ustack/ustack.go
```diff
@@ -53,12 +56,25 @@ func (o *Operator) GlobalParams() api.Params {
...
+func (o *Operator) ParamDescs() params.ParamDescs {
+	return params.ParamDescs{
+		{
+			Key:          goReSymParam,
+			Description:  "Symbol tables for stripped Go executables in GoReSym json format",
+			DefaultValue: "",
+		},
+	}
+}
+
```
mauriciovasquezbernal, line 68: Implement this directly in InstanceParams(). The `ParamDescs` method is legacy and shouldn't be needed by new operators.

pkg/symbolizer/symbolizer.go
```diff
@@ -80,7 +83,11 @@ func (k exeKey) String() string {
 	return fmt.Sprintf("ino=%d mtime=%d.%d", k.ino, k.mtimeSec, k.mtimeNsec)
 }
 
-func NewSymbolizer() (*Symbolizer, error) {
+type SymbolizerOptions struct {
+	GoReSymSpec string
```
mauriciovasquezbernal, line 87: Please document it. 
- Link to GoReSym
- Specify the format expected. json? is it only the path or the full thing?

pkg/symbolizer/symbolizer.go
```diff
@@ -89,13 +96,19 @@ func NewSymbolizer() (*Symbolizer, error) {
...
+	}
 
 	s := &Symbolizer{
 		symbolTables:    make(map[exeKey]*symbolTable),
 		hostProcFsPidNs: uint32(pid1PidNsStat.Ino),
 		pruneTickerTime: pruneTickerTime,
 		symbolTableTTL:  symbolTableTTL,
 		exit:            make(chan struct{}),
+
+		goReSyms: goReSyms,
```
mauriciovasquezbernal, line 111: What about making this logic more modular? Like having a common interface (basically `Resolve(task Task, addresses []uint64) ([]string, error)`) and then implementing it in different resolvers (existing resolver, GoReSym added on this PR, etc).

mauriciovasquezbernal (commented):
I did a quick check. I think it's great to explore different mechanisms for resolving the symbols.

> Use the ustack operator's parameters " --goresym=". As other parameters, you can use the @ symbol to pass a file.

I'm not sure we should use the `@` logic for these big files as these are sent [copied](https://github.com/inspektor-gadget/inspektor-gadget/blob/f360b474375d9c153f1c98e720516a87e706205b/cmd/common/params.go#L32-L42) on the request. IMO it should be the path, the question is, how can the user make this available to the gadget pod without having to hack a lot?

flyth:
This will not be compatible with detached mode on Kubernetes, because the file would be stored in the ConfigMap.

In general, I like supporting multiple symbolizers. But we really should discuss the architecture beforehand. If you already know the binaries in advance, then it might be better to do the resolving on the client (which could also pull info from external servers, then).

Looking forward to the modular approach!

alban:
> This will not be compatible with detached mode on Kubernetes, because the file would be stored in the ConfigMap.
>
> In general, I like supporting multiple symbolizers. But we really should discuss the architecture beforehand. If you already know the binaries in advance, then it might be better to do the resolving on the client (which could also pull info from external servers, then).

Good idea.

Since #4178, the addresses and the symbols are in different fields. And it is possible to disable the symbolization. To enable symbolization client-side, the server has to provide the build-id. Then, the client could use the build-id to fetch the symbol table (with debuginfod, goresym or other methods).

I'll convert this PR to draft and work on getting the build-id first.

PR #4087, opened 2025-02-18
ustack: don't warn about processes on the host

alban:
Processes in containers have several pids from different pid namespace point of view. But processes on the host have only one pid. This is expected, so the warnings were unwarranted.

This patch remove the warning in case of host process. This also makes the warnings clearer.


mauriciovasquezbernal (approved):
LGTM from code inspection!

Issue #1416, opened 2023-03-10
trace/dns: Support DNS over TCP


mauriciovasquezbernal:
From https://serverfault.com/a/404843:

> DNS uses TCP when the size of the request or the response is greater than a single packet such as with responses that have many records or many IPv6 responses or most DNSSEC responses.

The trace dns gadget should also support DNS over TCP to give a complete view of all DNS requests.

PR #4088, opened 2025-02-18
trace_dns: TCP support w/o reassembly 

mqasimsarfraz:
This change add support for TCP with per packet approach. It doesn't support to reassemble the segments but would be helpful since  nodelocaldns always uses TCP to resolve DNS on K8s nodes.

Ref: https://datatracker.ietf.org/doc/html/rfc7766#section-8

Related: https://github.com/inspektor-gadget/inspektor-gadget/issues/1416

## Testing done

```
/ # dig +tcp +short microsoft.com.
20.231.239.246
20.76.201.171
20.70.246.20
20.236.44.162
20.112.250.133
```

```
→ go run -exec sudo ./cmd/ig run trace_dns:qasim-dns-over-tcp --verify-image=false
WARN[0001] This gadget was built with ig v0.37.0-93-gf451011af-dirty and it's being run with v0.0.0. Gadget could be incompatible 
WARN[0002] This gadget was built with ig v0.37.0-93-gf451011af-dirty and it's being run with v0.0.0. Gadget could be incompatible 
RUNTIME.CONTAINERNAME                           SRC                                                  DST                                                  COMM                    PID        TID QR   QTYPE                      NAME                                           RCODE        ADDRESSES                                      LATENCY_NS
e66800e2296c59440687a10239b9fea7ab9552b088a240c 172.17.0.2:34535                                     192.168.0.1:53                                       isc-net-0000         191321     191322 Q    A                          microsoft.com.                                                                                                    0ns
e66800e2296c59440687a10239b9fea7ab9552b088a240c 192.168.0.1:53                                       172.17.0.2:34535                                     isc-net-0000         191321     191322 R    A                          microsoft.com.                                 Success      20.231.239.246,20.76.201…                         36.09ms

```

Also, added integration tests! 

## Note

I tested this change with a large payload i.e `~500` record to see how it behaves if DNS payload is fragmented and I am getting `WARN[0034] failed to unpack dns message: Name: segment prefix is reserved` warning for second segment. Perhaps fine to keep things like this? Thoughts?

```
RUNTIME.CONTAINERN… SRC                       DST                       COMM              PID        TID QR QTYPE      NAME                RCODE    ADDRESSES   LATENCY_NS
81d0328dcf2d3be4cb8 127.0.0.1:45021           127.0.0.1:53              isc-net-0…      87764      87765 Q  A          fake.test.com.                                  0ns
81d0328dcf2d3be4cb8 127.0.0.1:45021           127.0.0.1:53              dnstester       87212      87212 Q  A          fake.test.com.                                  0ns
81d0328dcf2d3be4cb8 127.0.0.1:53              127.0.0.1:45021           dnstester       87212      87212 R  A          fake.test.com.      Success  127.0.0.…          0ns
81d0328dcf2d3be4cb8 127.0.0.1:53              127.0.0.1:45021           isc-net-0…      87764      87765 R  A          fake.test.com.      Success  127.0.0.…       4.36ms
WARN[0034] failed to unpack dns message: Name: segment prefix is reserved 
81d0328dcf2d3be4cb8 127.0.0.1:53              127.0.0.1:45021           dnstester       87212      87212                                                               0ns
WARN[0034] failed to unpack dns message: Name: segment prefix is reserved 
81d0328dcf2d3be4cb8 127.0.0.1:53              127.0.0.1:45021           isc-net-0…      87764      87765                                                               0ns

```

### DIFF
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -262,20 +262,50 @@ int ig_trace_dns(struct __sk_buff *skb)
 		return 0;
 	}
 
+	// Since we have the same offset for source and destination ports for both TCP and UDP,
+	// we can use the same code to extract them.
+	// - offsetof(struct udphdr, source) == offsetof(struct tcphdr, source)
+	// - offsetof(struct udphdr, dest) == offsetof(struct tcphdr, dest)
 	switch (proto) {
 	case IPPROTO_UDP:
+	case IPPROTO_TCP:
 		sport = load_half(skb,
 				  l4_off + offsetof(struct udphdr, source));
 		dport = load_half(skb, l4_off + offsetof(struct udphdr, dest));
-		dns_off = l4_off + sizeof(struct udphdr);
 		break;
-	// TODO: support TCP
 	default:
 		return 0;
 	}
 
 	if (!is_dns_port(sport) && !is_dns_port(dport))
 		return 0;
+
+	// Calculate the DNS offset in the packet
+	switch (proto) {
+	case IPPROTO_UDP:
+		dns_off = l4_off + sizeof(struct udphdr);
+		break;
+	case IPPROTO_TCP:
+		// This is best effort, since we don't reassemble TCP segments.
+		struct tcphdr tcph;
+		if (bpf_skb_load_bytes(skb, l4_off, &tcph, sizeof tcph))
+			return 0;
+
+		// The data offset field in the header is specified in 32-bit words. We
+		// have to multiply this value by 4 to get the TCP header length in bytes.
+		__u8 tcp_header_len = tcph.doff * 4;
+
+		// Skip if we don't have any data to avoid handling control segments
+		dns_off = l4_off + tcp_header_len;
+		if (skb->len <= dns_off)
+			return 0;
+
+		// DNS is after the TCP header and the 2 bytes of the length of the DNS packet
+		dns_off += 2;
+		break;
+	default:
+		return 0;
+	}
 
 	event = bpf_map_lookup_elem(&tmp_events, &zero);
 	if (!event)
--- a/gadgets/trace_dns/test/integration/trace_dns_test.go
+++ b/gadgets/trace_dns/test/integration/trace_dns_test.go
@@ -63,7 +63,21 @@ const (
 	DefaultClientImage = "docker.io/library/busybox:latest"
 )
 
-func TestTraceDNS(t *testing.T) {
+type testCase struct {
+	name     string
+	protocol string
+
+	serverImage string
+	serverCmd   string
+
+	clientImage       string
+	clientUID         uint32
+	clientGID         uint32
+	clientExpectedCmd string
+	clientCmds        func(string, uint32, uint32) []string
+}
+
+func newTraceDNSStep(t *testing.T, tc testCase) (igtesting.TestStep, []igtesting.Option) {
 	gadgettesting.RequireEnvironmentVariables(t)
 	utils.InitTest(t)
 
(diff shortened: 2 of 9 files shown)


gadgets/trace_dns/program.bpf.c
```diff
@@ -269,7 +269,27 @@ int ig_trace_dns(struct __sk_buff *skb)
 		dport = load_half(skb, l4_off + offsetof(struct udphdr, dest));
 		dns_off = l4_off + sizeof(struct udphdr);
 		break;
-	// TODO: support TCP
+	case IPPROTO_TCP:
+		// This is best effort, since we don't reassemble segments.
+		struct tcphdr tcph;
+		if (bpf_skb_load_bytes(skb, l4_off, &tcph, sizeof tcph))
+			return 0;
```
alban, line 276: Here we copy 20 bytes instead of just the ports. This means a bit more work to do on all tcp traffic, including on traffic unrelated to DNS.

Also, by luck, we have:
- `offsetof(struct udphdr, source)` == `offsetof(struct tcphdr, source)`
- `offsetof(struct udphdr, dest)` == `offsetof(struct tcphdr, dest)`

So we could do the `is_dns_port` test before the `switch (proto)`.

I don't know if the optimization is worth it, wdty?

mqasimsarfraz: Do mean something following?

```
    // Since we have the same offset for both UDP and TCP, we can use the same code to extract the ports for both.
    // - offsetof(struct udphdr, source) == offsetof(struct tcphdr, source)
    // - offsetof(struct udphdr, dest) == offsetof(struct tcphdr, dest)
	switch (proto) {
	case IPPROTO_UDP:
	case IPPROTO_TCP:
		sport = load_half(skb,
				  l4_off + offsetof(struct udphdr, source));
		dport = load_half(skb, l4_off + offsetof(struct udphdr, dest));
		break;
	default:
		return 0;
	}

	if (!is_dns_port(sport) && !is_dns_port(dport))
		return 0;
```

It works fine. Also, I feel it always better to return early.

alban (approved):
A couple of comments but LGTM already. Thanks!

gadgets/trace_dns/test/integration/trace_dns_test.go
```diff
@@ -126,15 +127,17 @@ func TestTraceDNS(t *testing.T) {
 
 	var runnerOpts []igrunner.Option
 	var testingOpts []igtesting.Option
-	commonDataOpts := []utils.CommonDataOption{utils.WithContainerImageName(clientImage), utils.WithContainerID(clientContainer.ID())}
+	commonDataOpts := []utils.CommonDataOption{utils.WithContainerID(clientContainer.ID())}
 
 	switch utils.CurrentTestComponent {
 	case utils.IgLocalTestComponent:
+		// TODO: investigate why the ContainerImageName with IMAGE@sha256:DIGEST doesn't work as expected for ig
```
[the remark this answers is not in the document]
mqasimsarfraz: Open this: https://github.com/inspektor-gadget/inspektor-gadget/issues/4104

mqasimsarfraz:
Thanks for the review!

gadgets/trace_dns/program.bpf.c
```diff
@@ -262,21 +262,51 @@ int ig_trace_dns(struct __sk_buff *skb)
...
+		// The data offset field in the header is specified in 32-bit words. We
+		// have to multiply this value by 4 to get the TCP header length in bytes.
+		__u8 tcp_header_len = tcph.doff * 4;
+
+		// Skip if we don't have any data to avoid handling control segments
+		dns_off = l4_off + tcp_header_len;
+		if (skb->len <= dns_off)
+			return 0;
+
+		// DNS is after the TCP header and the 2 bytes of the length of the DNS packet
```
blanquicet, line 303:
```suggestion
		// DNS messages over TCP are prefixed with a two byte length field which
		// gives the message length, excluding the two byte length field.
		// https://www.rfc-editor.org/rfc/rfc1035#section-4.2.2
```

gadgets/trace_dns/program.bpf.c
```diff
@@ -262,21 +262,51 @@ int ig_trace_dns(struct __sk_buff *skb)
...
+		dns_off += 2;
```
blanquicet, line 304: I'm wondering if we can discard here the fragmented DNS payload by comparing that length field with the data length (`skb->len`). I'd expect that for the fragmented payloads, the length present in this 2-bytes field will be larger than the actual payload in the packet we are parsing: `dns_off + 2 + dns_tcp_lenght` > `skb->len`. WDYT?

My goal is to avoid sending it to user space if it will be anyway discarded by the wasm parsing.

blanquicet (approved):
Just one question, otherwise LGTM from code inspection!

gadgets/trace_dns/program.bpf.c
mqasimsarfraz: Good point. I did some testing and second fragment doesn't contain any length so the above check will only work for first packet. Having said that, having a check to skip if there is mismatch between calculated payload length and actual payload length can help:
```
		// If the calculated length of the DNS packet is not the same as the length of the TCP payload, skip
		// the packet. This is a best effort check, as we don't reassemble TCP segments.
		__u16 dns_payload_len = load_half(skb, dns_off);
		if (dns_payload_len != skb->len - dns_off - 2)
			return 0;
```

Do you think it is worth it?

blanquicet: Oh, I see. With this check, we reduce the amount of "_WARN[0034] failed to unpack dns message: Name: segment prefix is reserved_", but didn't remove them completely. To do that, we will need to keep track of the 5-tuple, and I don't know if it is worth it.

@alban any comment?

> Because it might be better if we inform user if certain things aren't being traced
>
It's documented that we don't support it, so, imo, it's fine to silently skip them. Now that I think ... It could be a good counter to export as metric: the number of TCP segments belonging to a DNS query/answer that were discarded by the gadget because they were fragmented over the network. It's more an "internal" information but indicates to users what they are not seeing on the trace.

cc @mauriciovasquezbernal

PR #4089, opened 2025-02-18
gadgets: Implement advise_seccomp as image-based

### LINKED ISSUE #3890: gadgets: Implement seccomp profile advise as image-based gadget
Port the [seccomp advise](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/advise/seccomp) gadget to be image-based.

This requires:
- access contents from ebpf maps. It can be implemented in two different ways:
  - by using wasm: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/003-wasm-support.md#handle-
ebpf-objects
  - by using a datasource: As proposed in #3718
- print custom output to the terminal (or to a file): https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/003-wasm-support.md#define-new-output-modes.

WIP in #3718
x-ref #3003

mauriciovasquezbernal:
Implement the advise_seccomp as image-based

This is heavily based on #3718.  
- adds a PreStop() lifecycle hook to do things "last minute"
- fixes lifecycle of WASM instances
- adds "flush-on-stop" annotation for map fetching in PreStop()

### How it works
- The eBPF code collects syscalls in a per mount namespace map. It's registered as a map-iter. 
- The wasm module subscribes to that map, and also creates a new text datasource with the seccomp policy output.
- When the gadget is stopped, the ebpf map is read and emitted, then the wasm module generates the seccomp policy

### Testing

The gadget needs to be run with some specific params, when it's stopped, prints a seccomp policy for each container on the system that was allowed by the filtering options:

```bash 
$ sudo ig run advise_seccomp --verify-image=false --map-fetch-count=0 --map-fetch-interval=0
^C// mycontainer2
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    "access",
    "arch_prctl",
    "brk",
    "clone",
    "close",
    "dup2",
    "execve",
    "fcntl",
    "getcwd",
    "geteuid",
    "getpgrp",
    "getpid",
    "getppid",
    "getrandom",
    "getuid",
    "ioctl",
    "lseek",
    "mmap",
    "mprotect",
    "newfstatat",
    "openat",
    "poll",
    "pread64",
    "prlimit64",
    "read",
    "recvfrom",
    "rseq",
    "rt_sigaction",
    "rt_sigreturn",
    "sendto",
    "set_robust_list",
    "set_tid_address",
    "setitimer",
    "setpgid",
    "setsockopt",
    "socket",
    "wait4",
    "write"
  ]
}
// mycontainer
{
  "architectures": [
    "SCMP_ARCH_X86_64",
    "SCMP_ARCH_X86",
    "SCMP_ARCH_X32"
  ],
  "syscalls": [
    "access",
    "arch_prctl",
    "brk",
    "clone",
    "close",
    "execve",
    "exit_group",
    "geteuid",
    "getpid",
    "getrandom",
    "getuid",
    "ioctl",
    "mmap",
    "mprotect",
    "newfstatat",
    "openat",
    "poll",
    "pread64",
    "prlimit64",
    "read",
    "rseq",
    "rt_sigaction",
    "rt_sigreturn",
    "sendfile",
    "set_robust_list",
    "set_tid_address",
    "setpgid",
    "wait4",
    "write"
  ]
}
```

#### params needed

In order to work properly, the gadget requires `--map-fetch-count=0` and `--map-fetch-interval=0`, ideally these params shouldn't be required.

#### output format

The gadget prints a seccomp policy for each container. I'm not so sure about this, perhaps it should print a single seccomp policy with all syscalls captured by the filter configuration. This will help to generate policies for deamonsets and replicas, but will require to run the gadget multiple times to generate the policies for multiple containers.

### TODO 
- [x] Documentation

#### Requisites 
- [x] #4159
- [x] #4160
- [x] #4161

--- 

Replaces #3718 
Fixes #3890

### DIFF
--- a/docs/gadget-devel/gadget-wasm-api-raw.md
+++ b/docs/gadget-devel/gadget-wasm-api-raw.md
@@ -569,3 +569,17 @@ Parameters:
 
 Return value:
 - (u32) 1 if the symbol exists, 0 otherwise.
+
+
+### Filtering
+
+#### `shouldDiscardMntnsID(mntnsID uint64) uint32`
+
+Check if a mount namespace ID should be discarded. It must only be used in
+`gadgetStart` or after.
+
+Parameters:
+- `mntnsID` (u64): Mount namespace ID
+
+Return value:
+- (u32) 1 if the mount namespace ID should be discarded, 0 otherwise.
--- a/gadgets/advise_seccomp/README.mdx
+++ b/gadgets/advise_seccomp/README.mdx
@@ -1,0 +1,507 @@
+---
+title: advise_seccomp
+sidebar_position: 0
+---
+
+import CodeBlock from '@theme/CodeBlock';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+The seccomp profile advisor gadget records syscalls that are issued in a
+specified container, and then uses this information to generate the corresponding
+seccomp profile.
+
+## Getting started
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/advise_seccomp:%IG_TAG% [flags]
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/advise_seccomp:%IG_TAG% [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Flags
+
+No Flags.
+
+## Guide
+
+We need to start the advise_seccomp gadget before running our workload, so it's
+able to capture all the syscalls the container uses.
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+
+```bash
+$ kubectl gadget run advise_seccomp:%IG_TAG% --map-fetch-count=0 --map-fetch-interval=0 --podname default-pod
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+
+```bash
+$ sudo ig run advise_seccomp:%IG_TAG% --map-fetch-count=0 --map-fetch-interval=0 --containername mycontainer
+```
+
+</TabItem>
+</Tabs>
+
+The --map-fetch-count=0 --map-fetch-interval=0 parameters indicate the policy
+will be generated once and only when the gadget is stopped.
+
+Then, start our application and interact with it to be sure it generates all the
+syscalls it needs to work.
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+
+import unconfined from '!!raw-loader!./files/seccomp-unconfined.yaml';
+
+<CodeBlock language="yaml">{unconfined}</CodeBlock>
+
+```bash
+$ kubectl apply -f https://raw.githubusercontent.com/inspektor-gadget/inspektor-gadget/refs/heads/%IG_BRANCH%/docs/gadgets/files/seccomp-unconfined.yaml
+```
+
+Use port-forward to access the server from the client machine
+
+```bash
+$ kubectl -n default port-forward default-pod 3000:80 &
+
+$ curl localhost:3000
+<!DOCTYPE html>
+<html>
+<head>
+<title>Welcome to nginx!</title>
+<style>
+html { color-scheme: light dark; }
+body { width: 35em; margin: 0 auto;
+font-family: Tahoma, Verdana, Arial, sans-serif; }
+</style>
+</head>
+<body>
+<h1>Welcome to nginx!</h1>
+<p>If you see this page, the nginx web server is successfully installed and
+working. Further configuration is required.</p>
+
+<p>For online documentation and support please refer to
+<a href="http://nginx.org/">nginx.org</a>.<br/>
+Commercial support is available at
+<a href="http://nginx.com/">nginx.com</a>.</p>
+
+<p><em>Thank you for using nginx.</em></p>
+</body>
+</html>
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+
+```bash
+$ docker run --name mycontainer --rm -d docker.io/library/nginx
+e6bc6e02989054f4984c04b7213d304767faffa295b277e8f36a5b2422409e18
+
+$ curl $(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' mycontainer)
+<!DOCTYPE html>
+<html>
+<head>
+<title>Welcome to nginx!</title>
+<style>
+html { color-scheme: light dark; }
+body { width: 35em; margin: 0 auto;
+font-family: Tahoma, Verdana, Arial, sans-serif; }
+</style>
+</head>
+<body>
+<h1>Welcome to nginx!</h1>
+<p>If you see this page, the nginx web server is successfully installed and
+working. Further configuration is required.</p>
+
+<p>For online documentation and support please refer to
+<a href="http://nginx.org/">nginx.org</a>.<br/>
+Commercial support is available at
+<a href="http://nginx.com/">nginx.com</a>.</p>
+
+<p><em>Thank you for using nginx.</em></p>
+</body>
+</html>
+```
+
+</TabItem>
+</Tabs>
+
+Now, go back and stop the gadget. It'll print to the terminal the seccomp policy
+for all container running on the system:
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+
+```bash
+$ kubectl gadget run advise_seccomp:%IG_TAG% --map-fetch-count=0 --map-fetch-interval=0 --podname default-pod
+^C
+// test-container
+{
+  "defaultAction": "SCMP_ACT_ERRNO",
+  "architectures": [
+    "SCMP_ARCH_X86_64",
+    "SCMP_ARCH_X86",
+    "SCMP_ARCH_X32"
+  ],
+  "syscalls": [
+    {
+      "names": [
+        "accept4",
+        "access",
+        "arch_prctl",
+        "bind",
+        "brk",
+        "capget",
+        "capset",
+        "chdir",
+        "chown",
+        "clone",
+        "close",
+        "connect",
+        "dup2",
+        "epoll_create",
+        "epoll_ctl",
+        "epoll_pwait",
+        "epoll_wait",
+        "eventfd2",
+        "execve",
+        "exit_group",
+        "faccessat2",
+        "fadvise64",
+        "fchdir",
+        "fchown",
+        "fcntl",
+        "fgetxattr",
+        "fsetxattr",
+        "fstat",
+        "fstatfs",
+        "futex",
+        "getcwd",
+        "getdents64",
+        "getegid",
+        "geteuid",
+        "getgid",
+        "getpid",
+        "getppid",
+        "getrandom",
+        "gettid",
+        "getuid",
+        "io_setup",
+        "ioctl",
+        "listen",
+        "lseek",
+        "mkdir",
+        "mmap",
+        "mprotect",
+        "munmap",
+        "nanosleep",
+        "newfstatat",
+        "openat",
+        "pipe2",
+        "prctl",
+        "pread64",
+        "prlimit64",
+        "pwrite64",
+        "read",
+        "recvfrom",
+        "recvmsg",
+        "rename",
+        "rseq",
+        "rt_sigaction",
+        "rt_sigprocmask",
+        "rt_sigreturn",
+        "rt_sigsuspend",
+        "sched_getaffinity",
+        "sendfile",
+        "sendmsg",
+        "set_robust_list",
+        "set_tid_address",
+        "setgid",
+        "setgroups",
+        "setsockopt",
+        "setuid",
+        "sigaltstack",
+        "socket",
+        "socketpair",
+        "statfs",
+        "syscall_1f4",
+        "sysinfo",
+        "tgkill",
+        "umask",
+        "uname",
+        "utimensat",
+        "vfork",
+        "wait4",
+        "write",
+        "writev"
+      ],
+      "action": "SCMP_ACT_ALLOW"
+    }
+  ]
+}
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+
+```bash
+$ sudo ig run advise_seccomp:%IG_TAG% --map-fetch-count=0 --map-fetch-interval=0 --containername mycontainer
+// mycontainer
+{
+  "defaultAction": "SCMP_ACT_ERRNO",
+  "architectures": [
+    "SCMP_ARCH_X86_64",
+    "SCMP_ARCH_X86",
+    "SCMP_ARCH_X32"
+  ],
+  "syscalls": [
+    {
+      "names": [
+        "accept4",
+        "access",
+        "arch_prctl",
+        "bind",
+        "brk",
+        "capget",
+        "capset",
+        "chdir",
+        "chown",
+        "clone",
+        "close",
+        "connect",
+        "dup2",
+        "epoll_create",
+        "epoll_ctl",
+        "epoll_pwait",
+        "epoll_wait",
+        "eventfd2",
+        "execve",
+        "exit_group",
+        "faccessat2",
+        "fadvise64",
+        "fchdir",
+        "fchown",
+        "fcntl",
+        "fgetxattr",
+        "fsetxattr",
+        "fstat",
+        "fstatfs",
+        "futex",
+        "getcwd",
+        "getdents64",
+        "getegid",
+        "geteuid",
+        "getgid",
+        "getpid",
+        "getppid",
+        "getrandom",
+        "gettid",
+        "getuid",
+        "io_setup",
+        "ioctl",
+        "listen",
+        "lseek",
+        "mkdir",
+        "mmap",
+        "mprotect",
+        "munmap",
+        "nanosleep",
+        "newfstatat",
+        "openat",
+        "pipe2",
+        "prctl",
+        "pread64",
+        "prlimit64",
+        "pwrite64",
+        "read",
+        "recvfrom",
+        "recvmsg",
+        "rename",
+        "rseq",
+        "rt_sigaction",
+        "rt_sigprocmask",
+        "rt_sigreturn",
+        "rt_sigsuspend",
+        "sched_getaffinity",
+        "sendfile",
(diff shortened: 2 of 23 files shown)


gadgets/advise_seccomp/gadget.yaml
```diff
@@ -0,0 +1,14 @@
...
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/advise_seccomp
+datasources:
+  syscalls:
+    annotations:
+      cli.output-mode: none
+      ebpf.map.flush-on-stop: true
+  advise:
+    annotations:
+      cli.supported-output-modes: advise
+      cli.default-output-mode: advise
```
alban, line 14: `advise` is not listed as possible output-mode in docs/spec/operators/cli.md

pkg/operators/cli/clioperator.go
```diff
@@ -137,6 +137,10 @@ func (o *cliOperatorInstance) ExtraParams(gadgetCtx operators.GadgetContext) api
   Supported data sources / output modes:`)
 	outputDefaultValues := make([]string, 0, len(dataSources))
 	for _, ds := range dataSources {
+		if ds.Annotations()["cli.output-mode"] == "none" {
```
alban, line 140: Should it use a constant?
Ditto below.

pkg/operators/wasm/wasm.go
```diff
@@ -101,6 +101,8 @@ func (w *wasmOperator) InstantiateImageOperator(
 }
 
 type wasmOperatorInstance struct {
+	ctx       context.Context
+	cancel    func()
```
alban, line 105: Could it be named preStartCtx and preStartCancel?

I am not sure I understand why it is needed, despite the explanation below. What does `callGuestFunction` do with the context?

gadgets/advise_seccomp/test/unit/advise_seccomp_test.go
```diff
@@ -0,0 +1,178 @@
...
+				require.Contains(t, policy.Syscalls, "getpid")
+				require.Contains(t, policy.Syscalls, "getppid")
+				require.Contains(t, policy.Syscalls, "getuid")
+				require.Contains(t, policy.Syscalls, "geteuid")
+				require.Contains(t, policy.Syscalls, "openat")
+				require.Contains(t, policy.Syscalls, "close")
+				require.Contains(t, policy.Syscalls, "sysinfo")
+				require.Contains(t, policy.Syscalls, "chdir")
+				require.Contains(t, policy.Syscalls, "mmap")
+				require.Contains(t, policy.Syscalls, "munmap")
```
alban, line 88: Could it use NotContains for some syscalls that were not used like `reboot`, `unshare` or `setns`?

alban (commented):
This looks generally good. Thanks!

The operator framework changes should have another reviewer.

About the params and output format: I agree with your comments. If it is mentioned in the documentation, they don't need to be resolved in this initial PR.

gadgets/advise_seccomp/go/program.go
```diff
@@ -0,0 +1,127 @@
...
+			scs, _ := syscalls.Bytes(data)
+			var containerName string
+			if container != 0 {
+				containerName, _ = container.String(data)
+			}
+
+			syscallStrings := make([]string, 0)
+			for i := 0; i < len(scs); i++ {
+				if scs[i] > 0 {
+					syscallName, err := api.GetSyscallName(uint16(i))
```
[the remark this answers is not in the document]
mauriciovasquezbernal: The callback is executed for each container that was recorded by the gadget. How do you think it can be improved? by having a cache inside the module with all syscall names?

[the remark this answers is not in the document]
alban: Yes, that was what I have thinking. But it's probably a too early optimization. Maybe `api.GetSyscallName` is actually fast.

gadgets/advise_seccomp/gadget.yaml
mauriciovasquezbernal: This output mode is specific for this gadget, so it shouldn't be documented on the operator.

alban: I didn't know a gadget could add a custom output mode. Should the operator documentation say that a gadget can implement its own?

pkg/operators/wasm/wasm.go
mauriciovasquezbernal: The issue is the following: gadgetCtx.Context() was passed to PreStart() where many gadgets subscribed to a data source, then this context was [passed](https://github.com/inspektor-gadget/inspektor-gadget/blob/8e1aecb422d023757650ec0bb61ecc7d91b8c29b/pkg/operators/wasm/datasource.go#L295-L317) to the callback. Then, when gadgetCtx.Context()  was cancelled, the subscriptions didn't get invoked because ctx in `_, err := i.dataSourceCallback.Call(ctx`  was cancelled.

gadgets/advise_seccomp/gadget.yaml
mauriciovasquezbernal: Done in the last commit.

gadgets/advise_seccomp/go/program.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			var out strings.Builder
+			out.WriteString(fmt.Sprintf("// %s\n", containerName))
+			jsonText, _ := json.MarshalIndent(map[string]any{
+				"defaultAction": "SCMP_ACT_ERRNO",
+				"architectures": []string{
+					"SCMP_ARCH_X86_64",
+					"SCMP_ARCH_X86",
+					"SCMP_ARCH_X32",
```
alban, line 110: TODO for a future PR: don't hardcode this but add support for arm64. The ebpf code already supports it.

flyth (approved):
Took a look mostly at the changes from the initial PR. LGTM from code inspection. Much better without the map init in user-space.

Regarding the need to set `--map-fetch-count=0` and `--map-fetch-interval=0`: I thought we already supported overriding param defaults from the gadget.yaml 🙈 
We should address that soon - maybe as part of a revived #3485 😄

burak-ok (dismissed):
Same for `gadgetPreStart`

mauriciovasquezbernal:
It turns out this gadget was missing more work than anticipated. I found some issues while writing documentation:
- The format of output was wrong, `syscalls` is not a list of strings but a different object with name and action
- The filtering by mount namespace can **not** be used on the ebpf code as the container collection isn't ready, i.e. some boostrap syscalls needed by runc were missing. This creates a further complication, now the gadget emits a profile for all container running on the system. 
  - Should we include some parameters on the wasm program to generate profiles for some pods / containers?
- The current approach doesn't seem to work on k8s (I think we shouldn't use runtime.containerName to name the profile)

For these reasons I'm marking it as draft again and moving the target milestone for the next release instead of rushing it in.

mauriciovasquezbernal:
I split this PR into smaller ones to ease the review. 

Regarding the UX:

> The filtering by mount namespace can not be used on the ebpf code as the container collection isn't ready, i.e. some boostrap syscalls needed by runc were missing. This creates a further complication, now the gadget emits a profile for all container running on the system.
>     Should we include some parameters on the wasm program to generate profiles for some pods / containers?


What should be the ideal UX for this gadget and when (and how) to implement it?
Different options:
1. (Current implementation) Print a profile for each container on the system. 
   - I think it makes this gadget very difficult to use on k8s as so many different profiles will be printed
2. Provide a parameter to filter my container name
   - It can be very simple, only the container name is taken into consideration. Its not possible to filter by pod name, namespace, labels.
3. Allow filtering by all k8s concepts, pod name,  namespace, labels, etc. 
   - I think this requires to expose some of the container collection logic to wasm, so a module can check if a mount namespace matches a given filter specified by the user

IMO option 2 is a good tradeoff between simplicity and functionality, Any thoughts?

alban:
> 2. Provide a parameter to filter my container name

What would you do if several containers have the same name (e.g. pods from a ReplicaSet with n>1 running on the same node)? Is it the k8s container name or the name known by the containerd/docker container runtime?

mauriciovasquezbernal:
> What would you do if several containers have the same name (e.g. pods from a ReplicaSet with n>1 running on the same node)?

It prints one profile for each container.

Actually I'm trying to implement it, but I'm facing some issues, so I think it'll have to explore another alternatives (like option 3).

mauriciovasquezbernal:
This is ready for review again. I ended up implementing a `ShouldDiscardMntNsID` to be used by the gadget to discard profiles the user is not interested in. Please check the documentation (in the sense, test it) while reviewing it.

pkg/operators/wasm/wasm_test.go
```diff
@@ -466,3 +467,13 @@ func TestConfig(t *testing.T) {
 
 	require.Equal(t, "myvalue", v.GetString("foo.bar.zas"))
 }
+
+func TestFiltering(t *testing.T) {
+	gadgetCtx := createGadgetCtx(t, "filtering")
+
+	mntNsMap := utilstest.CreateMntNsFilterMap(t, 777)
```
alban, line 474: Add a comment
```
// Value to keep in sync with testdata/filtering/program.go
```

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -567,3 +567,16 @@ Parameters:
 
 Return value:
 - (u32) 1 if the symbol exists, 0 otherwise.
+
+
+### Filtering
+
+#### `shouldDiscardMntnsID(mntnsID uint64) uint32`
+
+Check if a mount namespace ID should be discarded.
```
alban, line 576: Should it say that it can only be called at gadgetStart or after. If called at gadgetInit, the result might be undefined?

wasmapi/go/filter.go
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	_ "unsafe"
```
alban, line 18: Why unsafe?

gadgets/advise_seccomp/README.mdx
```diff
@@ -0,0 +1,485 @@
...
+```bash
+$ kubectl gadget run advise_seccomp:%IG_TAG% --map-fetch-count=0 --map-fetch-interval=0 --podname default-pod
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+
+```bash
+$ sudo ig run advise_seccomp:%IG_TAG% --map-fetch-count=0 --map-fetch-interval=0 --containername mycontainer
```
alban, line 51: Could you explain why `--map-fetch-count=0 --map-fetch-interval=0`?

gadgets/advise_seccomp/README.mdx
```diff
@@ -0,0 +1,485 @@
...
+</body>
+</html>
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+
+```bash
+$ docker run --name mycontainer --rm -d docker.io/library/nginx
```
alban, line 107: Shouldn't you run it unconfined?

For this specific workload, this is fine. But if you run an unprivileged container, by default docker will not give CAP_SYS_ADMIN and it block the mount syscall. If the container tries to call `mount`, the advise_seccomp gadget will not capture it because the seccomp filter will block it before it can be captured in `raw_tracepoint/sys_enter`.

gadgets/advise_seccomp/README.mdx
```diff
@@ -0,0 +1,485 @@
...
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/advise_seccomp:%IG_TAG% [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Flags
+
+No Flags.
+
+## Guide
```
alban, line 34: Additional sections: `## Limitations`:
* Does it start recording at the right point in the runc startup?
* Does it work with crun?
* General warning that it is difficult to exercise all code paths in your workload
* Run workload without seccomp filter during the learning phase to be able to capture all syscalls (see comment below).
* Merging profiles with ReplicaSet

Additional sections: `## Related project`:
* SPO's Profile Recording: https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/installation-usage.md#record-seccomp-profile
* https://github.com/containers/oci-seccomp-bpf-hook (podman only)

burak-ok:
I get the following output running on K8s:
```
$ ./kubectl-gadget run ghcr.io/burak-ok/test --pull always  --map-fetch-count=0 --map-fetch-interval=0
WARN[0000] version skew detected: client (v0.0.0) vs server (v0.38.0-63-g3b0723c66) 
WARN[0003] minikube-docker      | image signature verification is disabled due to using corresponding option 
WARN[0008] Data is incomplete: timeout waiting for data from all targets (0/1) 
^C% 
```

This error is from the `combiner` operator. Did we miss the K8s side in https://github.com/inspektor-gadget/inspektor-gadget/pull/4159 ?

wasmapi/go/filter.go
mauriciovasquezbernal: It's needed by `//go:wasmimport` or `//go:linkname`, I can't remember which one.

gadgets/advise_seccomp/README.mdx
mauriciovasquezbernal: > If the container tries to call mount, the advise_seccomp gadget will not capture it because the seccomp filter will block it before 

I think it's fine not to record syscalls that are blocked, in any case the application will fail, so it's up to the user to guarantee the application is working fine while recording the syscalls.

mauriciovasquezbernal:
> I get the following output running on K8s:
> 
> ```
> $ ./kubectl-gadget run ghcr.io/burak-ok/test --pull always  --map-fetch-count=0 --map-fetch-interval=0
> WARN[0000] version skew detected: client (v0.0.0) vs server (v0.38.0-63-g3b0723c66) 
> WARN[0003] minikube-docker      | image signature verification is disabled due to using corresponding option 
> WARN[0008] Data is incomplete: timeout waiting for data from all targets (0/1) 
> ^C% 
> ```
> 
> This error is from the `combiner` operator. Did we miss the K8s side in #4159 ?

It doesn't seem to be related to this or that PR, it happens with other gadgets as well:

```bash 
t$ kubectl gadget run top_file:main --map-fetch-count=0 --map-fetch-interval=0 --podname mypod -o json
WARN[0000] minikube             | image signature verification is disabled due to using corresponding option
WARN[0005] Data is incomplete: timeout waiting for data from all targets (0/1)
[]
```

pkg/operators/wasm/filter.go
```diff
@@ -0,0 +1,55 @@
...
+	// all data
+	if i.mntNsIDMap == nil {
+		stack[0] = 0
+		return
+	}
+
+	ret := uint32(0)
+
+	mntnsID := uint64(stack[0])
+	if err := i.mntNsIDMap.Lookup(mntnsID, &ret); err != nil {
```
alban, line 49: It is a bit weird to ask the kernel to lookup the ebpf map via the bpf syscall instead of querying the container-collection or the tracer-collection. We should have the information in the ig process.

But the current architecture does not make it easy to access the information in the ig process.

pkg/operators/wasm/filter.go
```diff
@@ -0,0 +1,55 @@
...
+// Params:
+// - stack[0]: mount ns ID
+// Return value:
+// - 1 if the mount ns ID should be filter out, 0 otherwise
+func (i *wasmOperatorInstance) shouldDiscardMntnsID(ctx context.Context, m wapi.Module, stack []uint64) {
+	// if the filtering map is not configured, we assume the gadget wants to get
+	// all data
+	if i.mntNsIDMap == nil {
+		stack[0] = 0
+		return
```
alban, line 43: Even if the advise seccomp gadget does not include "include/gadget/mntns_filter.h", the MntNsFilterMapName map is created by localmanager or kubemanager, so it works. But this seems fragile to me, it could stop working if the localmanager or kubemanager was implemented differently.

alban (approved):
A few changes requested, LGTM after that!

I have some concerns about the access to the ebpf map from wasm, but it would require larger changes to address it.

gadgets/advise_seccomp/artifacthub-pkg.yml
```diff
@@ -0,0 +1,29 @@
+# Artifact Hub package metadata file
+version: 0.37.0
+name: "advise seccomp"
+category: monitoring-logging
+displayName: "advise seccomp"
+createdAt: "2025-02-03T11:23:41Z"
+digest: "2025-02-03T11:23:41Z"
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Do we have a makefile to update it? I usually just copy and paste an existing one.

gadgets/advise_seccomp/README.mdx
```diff
@@ -0,0 +1,506 @@
...
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+
+For Kubernetes, we can follow [Restrict a Container's Syscalls with
+seccomp](https://kubernetes.io/docs/tutorials/security/seccomp/).
+
+Copy the profile to the node:
+
+```bash
+$ minikube cp profile.json /var/lib/kubelet/seccomp/profile.json
```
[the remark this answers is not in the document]
mauriciovasquezbernal: It worked for me without the profile. I suppose it defaults to the default profile. I'd prefer to avoid having it on the documentation as it adds noise.

gadgets/advise_seccomp/artifacthub-pkg.yml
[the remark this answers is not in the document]
alban: We don't have a Makefile but we have a GitHub Action to update it at each release:
https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.39.0/.github/workflows/release.yml#L34-L45
See for example #4298.

For a PR introducing a new gadget, you can just update it manually to make it look reasonable.

pkg/operators/wasm/filter.go
mauriciovasquezbernal: You're right. I explicitly imported the file on the gadget to avoid it relying on {local,kube}manager implementations.

pkg/operators/wasm/filter.go
mauriciovasquezbernal: You're right, unfortunately it seems difficult to fix with the current architecture as you mentioned. I added a TODO in the code and hopefully we can fix it once we remove built-in gadgets.

mauriciovasquezbernal:
CI failures are not related to this PR or they are a chicken and egg situation.

PR #4091, opened 2025-02-19
builtin trace exec: reduce map size

dorser:
# Reduce Memory Usage in builtin trace_exec Gadget

This is a follow up to PR #4058 and related to #4057.

This PR changes max entries of the eBPF map in built-in the `trace_exec` gadget from `10240` to `1024` (when using without --paths option). This modification aim to lower memory usage by shrinking the pre-allocated memory for the execs map, which can help in resource-constrained environments or when the data stored in these maps is minimal.

## How to use

```
make ebpf-objects
make ig
sudo ./ig trace exec # without cwd, exepath and and
```

## Testing done
```
sudo bpftool map show -j | jq -r '.[] | {id: .id, type: .type, name: .name, bytes_memlock: .bytes_memlock, comms: (if .pids then (.pids | map(.comm) | unique | join(",")) else "" end)} | [.id, .type, .name, .bytes_memlock, .comms] | @tsv' | sort -k4n | column -t -s $'\t' | grep execs
```
### Before the change:
*without paths*
```
231  hash              execs            54333312
```

### After the change 
*without paths*
```
267  hash              execs            5500800
```


pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
```diff
@@ -41,7 +41,7 @@ struct {
 #ifdef WITH_LONG_PATHS
 	__uint(max_entries, 1024);
 #else /* !WITH_LONG_PATHS */
-	__uint(max_entries, 10240);
+	__uint(max_entries, 1024);
```
alban, line 44: The `#ifdef WITH_LONG_PATHS` line 41 could be removed since both branches are now the same.

alban:
Apart from the `#ifdef` comment, the patch looks good to me.

Could you rebase your branch on `origin/master` and squash the commits into one?

pkg/gadgets/trace/exec/tracer/bpf/execsnoop.bpf.c
dorser: Oh wow lol :D

alban (approved):
LGTM if the CI is green.

Issue #4000, opened 2025-02-07
[Research] Use go 1.24 for WASM instead of tinygo


burak-ok:
Go 1.24 will have finally support for [exporting functions](https://github.com/golang/go/issues/65199). This means in theory we can use Go 1.24 instead of tinygo.

My test branch with go 1.24: https://github.com/inspektor-gadget/inspektor-gadget/tree/burak/wasm/go1.24/test

Both versions worked (with some code changes) and testing resulted in the following:

|                                     | Go 1.24                  | Tinygo 0.35       |
|-------------------------------------|--------------------------|-------------------|
| Less dependencies in general        | Yes                      | No                |
| Fully working "standard library"       | Yes                      | No                |
| CGO                                 | No                       | Yes               |
| malloc/free                         | No                       | Yes               |
| Fast adoption of new WASI standards | No                       | Yes               |
| Binary size                         | 2472 KB                  | 521 KB            |
| Stripped binary size                | -ldflags "-w -s": 2424KB | -no-debug: 162 KB |

Malloc/Free is not available, nor is CGO. We would need to change some of our code to adopt to this standard. While we don't use the exported `free`, we use `malloc`. https://github.com/inspektor-gadget/inspektor-gadget/pull/3995 for example uses other mechanisms to transfer bytes to the wasm program.

CGO is currently used to free the bytes allocated by `malloc` from the above step

burak-ok:
[Go `1.24` is released](https://go.dev/blog/go1.24)

blanquicet:
As agreed on dev sync meeting, we will go for Go 1.24.

burak-ok:
> As agreed on dev sync meeting, we will go for Go 1.24. Please [@burak-ok](https://github.com/burak-ok) create a list of sub-tasks required to achieve that so that different people can start working on that.

Tbh, its only the `malloc`. With that also the `free` disappears and everything works. I linked my test branch: https://github.com/inspektor-gadget/inspektor-gadget/tree/burak/wasm/go1.24/test

mqasimsarfraz:
After profiling tinygo compiled gadgets with https://github.com/dispatchrun/wzprof, I saw a lot of time spent in allocations so wanted to see how things are with golang1.24 and following are the results [using this patch](https://gist.github.com/mqasimsarfraz/fb50643a18f9e427c2c14ecb04a31279), sorting makes it worse but that isn't the point:

## 50 parallel requests

### tinygo
```
[...]

PR #4097, opened 2025-02-20
 wasm: Remove malloc requirement for guest

### LINKED ISSUE #3073: wasm: Check `malloc` export considerations
The current implementation requires the guest to expose the `malloc` function. 

https://github.com/inspektor-gadget/inspektor-gadget/blob/7a3e62698bbc6ed763ea0b448197de2e093ed510/pkg/operators/wasm/wasm.go#L239-L243

It's used by the host to allocate memory on the guest to save the value (like string or bytes of a field). The malloc function is automatically exported when using tinygo, but it could be not exported when using another language. We should check if we need to switch to a more generic approach to allow implementing gadgets in other programming languages.

Ref https://github.com/inspektor-gadget/inspektor-gadget/pull/2818#discussion_r1594318011

mauriciovasquezbernal:
Malloc() was required to be able to allocate and write data into the
    guest memory. This requirement was a bit strict as:
    - malloc() is exposed when using tinygo by a obscure undocumented way
    - This isn't exposed when using go 1.24
    - It can be challenging to expose on another programming languages

    This commit removes this requirement. The main change is that the guest
    now has to pass a preallocated buffer where the host will write the data.
    It can be suboptimal for some cases when the guest doesn't know the
    size, but it can help to reduce the allocations if the guest passes a
    static buffer in multiple calls. (As it was done in the trace_exec and
    trace_dns gadgets).

After applying this change, we can use go 1.24 to compile our gadgets as experimented on https://github.com/inspektor-gadget/inspektor-gadget/issues/4000#issuecomment-2656096773 by @burak-ok 

Fixes #3073

### DIFF
--- a/docs/gadget-devel/gadget-wasm-api-raw.md
+++ b/docs/gadget-devel/gadget-wasm-api-raw.md
@@ -15,9 +15,8 @@ Data types passed to the API are encoded using 64 bits. Scalar types like
 integers, booleans and floats are casted directly to 64 bit integers and passed
 using the stack.
 
-Strings and byte arrays are stored in the wasm module's memory as `bufPtr`. A 64
-bit integer is used to represent a pointer to them, the higher 32 bits contains
-the length and the lower 32 the memory address.
+Strings and byte arrays are represented by a 64 bits integer. The higher 32 bits
+contains the length and the lower 32 the memory address.
 
 ## Wasm Module Exported Functions
 
@@ -53,15 +52,6 @@ things. This function is optional.
 ### `gadgetPostStop`
 
 This function is called after the the gadget is stopped. This function is optional.
-
-#### `malloc`
-
-The gadget should expose a `malloc` function that allocates memory on the heap
-of the wasm module. This is needed by the host to allocate memory to pass
-strings and byte arrays around. This function is automatically exported when
-using tinygo https://github.com/tinygo-org/tinygo/issues/2788.
-
-TODO: We'll check this requirement later on.
 
 #### `dataSourceCallback`
 
@@ -282,9 +272,9 @@ Return value:
 
 ### Fields
 
-#### `fieldGet(u32 field, u32 data, u32 kind) u64`
+#### `fieldGetScalar(u32 field, u32 data, u32 kind) u64`
 
-Get the value of a field into a newly allocated buffer.
+Get the value of a scalar field.
 
 Parameters:
 - `field` (u32): Field handle (as returned by `dataSourceGetField` or `dataSourceAddField`)
@@ -292,20 +282,10 @@ Parameters:
 - `kind` (u32): Kind of access: How to read the field.
 
 Return value:
-- Value of the field:
-  - If the returned value is of type String or Bytes, it will
-  be allocated inside the wasm guest memory by calling the function malloc. The
-  Wasm module must either provide its own implementation of malloc or be
-  compiled against a library which provides it such as libc. It is the
-  responsibility of the caller to free the allocation.
-  - The reference Wasm guest Go library
-  "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
-  automatically frees the memory as appropriate so if your Wasm module uses that
-  reference implementation, you don't have to call free.
-  - The function returns 0 in case of errors (ambiguous with scalar types like u32).
-  TODO: Find a way to report errors!
+- Value of the field: The value of the field or 0 in case of errors
+  - TODO: 0 is ambiguous, find a way to report errors!
 
-#### `fieldGetToBuffer(u32 field, u32 data, u32 kind, u64 dst) u32`
+#### `fieldGetBuffer(u32 field, u32 data, u32 kind, u64 dst) i32`
 
 Get the value of a field of type String or Bytes into an existing buffer.
 
@@ -313,9 +293,10 @@ Parameters:
 - `field` (u32): Field handle (as returned by `dataSourceGetField` or `dataSourceAddField`)
 - `data` (u32): Data handle
 - `kind` (u32): Kind of access: How to read the field.
+- `dst` (u64): A pointer to a buffer where the value will be stored.
 
 Return value:
-- Value of the field: the number of bytes copied or 0 in case of errors.
+- Value of the field: the number of bytes copied or -1 in case of errors.
 
 #### `fieldSet(u32 field, u32 data, u32 kind, u64 value)`
 
@@ -361,15 +342,16 @@ params:
      ...
 ```
 
-#### `getParamValue(key string) string`
+#### `getParamValue(key string, dst uint64) uint32`
 
 Return the value of a parameter.
 
 Parameters:
 - `key` (string): Key of the parameter.
+- `dst` (u64): A pointer to a buffer where the value will be stored.
 
 Return value:
-- The value of the parameter.
+- 0 in case of success, 1 otherwise.
 
 ### Config
 
@@ -475,15 +457,18 @@ Return value:
 
 ### Syscalls
 
-#### `getSyscallName(id uint32) uint64`
+#### `getSyscallName(id uint32, dst uint64) uint32`
 
-Get the syscall name for this syscall ID.
+Get the syscall name for this syscall ID. In case the syscall is unknown, this
+function resolves it as "syscall_ID" with ID displayed as hexadecimal like
+  strace.
 
 Parameters:
 - `id` (u32): Syscall ID
+- `dst` (u64): A pointer to a buffer where the name will be stored.
 
 Return value:
-- (u64) A string containing the name of the syscall, in case the syscall is unknown, it returns "syscall_ID" with ID displayed as hexadecimal like strace.
+- (u32) 0 in case of success, 1 otherwise.
 
 #### `getSyscallDeclaration(name uint64, pointer uint64) uint32`
 
@@ -539,13 +524,13 @@ Parameters:
 Return value:
 - (u32) 0 on success, 1 on error.
 
-#### `func perfReaderRead(perfMapHandle uint32, addrBufPtr uint32) uint32`
+#### `func perfReaderRead(perfMapHandle uint32, dst uint64) uint32`
 
 Read the perf buffer.
 
 Parameters:
 - `perfMapHandle` (u32): Handle to a perf buffer.
-- `addrBufPtr` (u32): Address to a bufptr where the record will be written. The bufptr will be allocated by the function and must be freed by the caller.
+- `dst` (u64): A pointer to a buffer where the data will be stored.
 
 Return value:
 - (u32) 0 on success, 1 on error, 2 on deadline exceeded.
--- a/pkg/operators/wasm/fields.go
+++ b/pkg/operators/wasm/fields.go
@@ -25,7 +25,7 @@ import (
 )
 
 func (i *wasmOperatorInstance) addFieldFuncs(env wazero.HostModuleBuilder) {
-	exportFunction(env, "fieldGet", i.fieldGet,
+	exportFunction(env, "fieldGetScalar", i.fieldGetScalar,
 		[]wapi.ValueType{
 			wapi.ValueTypeI32, // Accessor
 			wapi.ValueTypeI32, // Data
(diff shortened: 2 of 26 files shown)


pkg/operators/wasm/fields.go
```diff
@@ -206,55 +187,29 @@ func (i *wasmOperatorInstance) fieldGetToBuffer(ctx context.Context, m wapi.Modu
...
-			val = handleBytes([]byte(str))
-		}
-	case api.Kind_Bytes:
-		bytes, err := field.Bytes(data)
-		if err == nil {
-			val = handleBytes(bytes)
+	case api.Kind_String, api.Kind_Bytes:
+		bytes := field.Get(data)
+		if err := i.writeToDstBuffer(bytes, fieldDst); err != nil {
+			i.logger.Warnf("fieldGetToBuffer: writing bytes to guest memory: %v", err)
```
alban, line 203:
```suggestion
			i.logger.Warnf("fieldGetToBuffer: %v", err)
```

pkg/operators/wasm/helpers.go
```diff
@@ -81,17 +81,14 @@ func exportFunction(
 		Export(name)
 }
 
-func (i *wasmOperatorInstance) writeToGuestMemory(ctx context.Context, buf []byte) (uint64, error) {
-	res, err := i.guestMalloc.Call(ctx, uint64(len(buf)))
-	if err != nil {
-		return 0, fmt.Errorf("malloc failed: %w", err)
+func (i *wasmOperatorInstance) writeToDstBuffer(src []byte, dstBuf uint64) error {
+	if getLength(dstBuf) < uint32(len(src)) {
+		return fmt.Errorf("fieldGet: writing %d bytes to guest memory buffer of %d bytes", len(src), getLength(dstBuf))
```
alban, line 86:
```suggestion
		return fmt.Errorf("writing %d bytes to guest memory buffer of %d bytes", len(src), getLength(dstBuf))
```

pkg/operators/wasm/helpers.go
```diff
@@ -81,17 +81,14 @@ func exportFunction(
...
 	}
-
-	if !i.mod.Memory().Write(uint32(res[0]), buf) {
-		return 0, fmt.Errorf("out of memory write")
+	if !i.mod.Memory().Write(getAddress(dstBuf), src) {
+		return fmt.Errorf("fieldGet: writing bytes to guest memory: out of memory write")
```
alban, line 89:
```suggestion
		return fmt.Errorf("writing bytes to guest memory: out of memory write")
```

pkg/operators/wasm/helpers.go
```diff
@@ -101,3 +98,14 @@ func isDataArrayHandle(handle uint32) bool {
 func getIndexFromDataArrayHandle(dataHandle uint32) int {
 	return int(dataHandle &^ dataArrayHandleFlag >> 16)
 }
+
+func (i *wasmOperatorInstance) writeToDstBuf(src []byte, dst uint64) (uint64, error) {
```
alban, line 102: This function is not necessary: it is the same as `writeToDstBuffer`, except that it returns the uint64. But the return value is never used. So you could keep `writeToDstBuffer`.

wasmapi/go/syscall.go
```diff
@@ -56,15 +56,13 @@ type SyscallDeclaration struct {
 }
 
 func GetSyscallName(id uint16) (string, error) {
-	ptr := bufPtr(getSyscallName(uint32(id)))
-	if ptr == 0 {
+	dst := make([]byte, 64)
```
alban, line 59: Define a constant `maxSyscallLength = 64` somewhere.

From [syscalls.csv](https://github.com/seccomp/libseccomp/blob/main/src/syscalls.csv), the max length is 28 so using 64 is fine.
```
cut -d, -f1 syscalls.csv  |awk '{print length($0)" "$0}'|sort -n
```

pkg/operators/wasm/fields.go
```diff
@@ -78,14 +78,13 @@ func (i *wasmOperatorInstance) getDataFromDatasourceHandle(dataHandle uint32) (d
 	return data, data != nil
 }
 
-// fieldGet returns the field's value.
+// fieldGet returns the field's value for scalar fields.
```
alban, line 81: What about naming the function `fieldGetScalar`?

pkg/operators/wasm/fields.go
```diff
@@ -190,7 +171,7 @@ func (i *wasmOperatorInstance) fieldGet(ctx context.Context, m wapi.Module, stac
 	stack[0] = val
 }
 
-// fieldGetToBuffer returns the field's value.
+// fieldGetToBuffer returns the field's value for string and bytes fields.
```
alban, line 174: What about naming the function `fieldGetBuffer`?

pkg/operators/wasm/syscalls.go
```diff
@@ -28,8 +28,11 @@ import (
 
 func (i *wasmOperatorInstance) addSyscallsDeclarationsFuncs(env wazero.HostModuleBuilder) {
 	exportFunction(env, "getSyscallName", i.getSyscallName,
-		[]wapi.ValueType{wapi.ValueTypeI32}, // Syscall ID
-		[]wapi.ValueType{wapi.ValueTypeI64}, // Syscall Name
+		[]wapi.ValueType{
+			wapi.ValueTypeI32, // Syscall ID
+			wapi.ValueTypeI64, // Buffer to save syscall name
+		},
+		[]wapi.ValueType{wapi.ValueTypeI64}, // Error
```
alban, line 35: Why not return value of `ValueTypeI32`? Other functions that return "0 in case of success, 1 otherwise" are `ValueTypeI32` so it is not consistent.

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -361,15 +361,16 @@ params:
      ...
 ```
 
-#### `getParamValue(key string) string`
+#### `getParamValue(key string, dst uint64) uint64`
```
alban, line 364: Why not returning uint32? Some other functions that return "0 in case of success, 1 otherwise" use a uint32.

pkg/operators/wasm/helpers.go
mauriciovasquezbernal: ohman! It was me trying to split the logic into two different commits. I removed `writeToDstBuf`

pkg/operators/wasm/syscalls.go
mauriciovasquezbernal: My bad. It was used to return the name too, so I only changed the description without checking the type. Thanks for catching this up!

docs/gadget-devel/gadget-wasm-api-raw.md
mauriciovasquezbernal: Same as above.

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -282,7 +282,7 @@ Return value:
 
 ### Fields
 
-#### `fieldGet(u32 field, u32 data, u32 kind) u64`
+#### `fieldGetScalar(u32 field, u32 data, u32 kind) u64`
 
 Get the value of a field into a newly allocated buffer.
```
burak-ok, line 287: And the `Return value` section also needs editing
```suggestion
Get the value of a field.
```

wasmapi/go/config.go
```diff
@@ -17,9 +17,11 @@ package api
 import (
 	"fmt"
 	"runtime"
+	_ "unsafe"
 )
 
 //go:wasmimport env setConfig
+//go:linkname setConfig setConfig
```
burak-ok, line 24: Could you explain why this is now needed?

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -539,13 +543,13 @@ Parameters:
...
 
-#### `func perfReaderRead(perfMapHandle uint32, addrBufPtr uint32) uint32`
+#### `func perfReaderRead(perfMapHandle uint32, dst uint64) uint32`
 
 Read the perf buffer.
 
 Parameters:
 - `perfMapHandle` (u32): Handle to a perf buffer.
-- `addrBufPtr` (u32): Address to a bufptr where the record will be written. The bufptr will be allocated by the function and must be freed by the caller.
+- `dst` (u32): A pointer to a buffer where the data will be stored.
```
mqasimsarfraz, line 552:
```suggestion
- `dst` (u64): A pointer to a buffer where the data will be stored.
```

pkg/operators/wasm/helpers.go
```diff
@@ -81,17 +81,14 @@ func exportFunction(
 		Export(name)
 }
 
-func (i *wasmOperatorInstance) writeToGuestMemory(ctx context.Context, buf []byte) (uint64, error) {
-	res, err := i.guestMalloc.Call(ctx, uint64(len(buf)))
-	if err != nil {
-		return 0, fmt.Errorf("malloc failed: %w", err)
+func (i *wasmOperatorInstance) writeToDstBuffer(src []byte, dstBuf uint64) error {
+	if getLength(dstBuf) < uint32(len(src)) {
+		return fmt.Errorf("writing %d bytes to guest memory buffer of %d bytes", len(src), getLength(dstBuf))
```
mqasimsarfraz, line 86:
```suggestion
		return fmt.Errorf("writing %d bytes to guest memory buffer of %d bytes: not enough memory", len(src), getLength(dstBuf))
```

wasmapi/go/config.go
mauriciovasquezbernal: It's needed to be able to run unit tests when cgo is removed from  wasmapi/go/helpers.go:

```bash 
$ go test -exec sudo .
# github.com/inspektor-gadget/inspektor-gadget/wasmapi/go
../../../wasmapi/go/datasource.go:30:6: missing function body
../../../wasmapi/go/datasource.go:39:6: missing function body
../../../wasmapi/go/datasource.go:42:6: missing function body
../../../wasmapi/go/datasource.go:45:6: missing function body
../../../wasmapi/go/datasource.go:48:6: missing function body
../../../wasmapi/go/datasource.go:51:6: missing function body
../../../wasmapi/go/datasource.go:54:6: missing function body
../../../wasmapi/go/fields.go:24:6: missing function body
../../../wasmapi/go/fields.go:27:6: missing function body
../../../wasmapi/go/fields.go:30:6: missing function body
../../../wasmapi/go/fields.go:27:6: too many errors
FAIL    github.com/inspektor-gadget/inspektor-gadget/pkg/operators/wasm [build failed]
FAIL
```

mqasimsarfraz (approved):
LGTM! 

Please feel free to wait for other reviews if needed!

wasmapi/go/config.go
mauriciovasquezbernal: In Inspektor Gadget we don't have documentation about it. Do you think we should document it? If yes, where and what should we write?

wasmapi/go/fields.go
```diff
@@ -164,11 +169,13 @@ func (f Field) SetFloat64(data Data, value float64) error {
 	return nil
 }
 
-func (f Field) String(data Data) (string, error) {
-	str := bufPtr(fieldGet(uint32(f), uint32(data), uint32(Kind_String)))
-	ret := str.string()
-	str.free()
-	return ret, nil
+func (f Field) String(data Data, maxSize uint32) (string, error) {
+	dst := make([]byte, maxSize)
```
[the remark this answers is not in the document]
mauriciovasquezbernal: We have a Size() method:

```go
	// Size returns the expected size of the underlying field or zero, if the field has a dynamic size
	Size() uint32
```

but it doesn't work for strings. I don't see any other option that getting the full string and applying len() on it.

wasmapi/go/perf.go
```diff
@@ -70,21 +75,17 @@ func (p PerfReader) Resume() error {
 	return nil
 }
 
-func (p PerfReader) Read() ([]byte, error) {
-	var buf bufPtr
-
-	ret := perfReaderRead(uint32(p), uint32(uintptr(unsafe.Pointer(&buf))))
+func (p PerfReader) Read(dst []byte) error {
```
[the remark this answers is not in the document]
mauriciovasquezbernal: Happy to do that in a following PR. By the way, we just allow to read the raw sample, should we provide all the fields like CPU, LostSamples, Remaining too?

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -15,9 +15,8 @@ Data types passed to the API are encoded using 64 bits. Scalar types like
 integers, booleans and floats are casted directly to 64 bit integers and passed
 using the stack.
 
-Strings and byte arrays are stored in the wasm module's memory as `bufPtr`. A 64
-bit integer is used to represent a pointer to them, the higher 32 bits contains
-the length and the lower 32 the memory address.
+Strings and byte arrays are represented by a 64 bits integer. The higher 32 bits
```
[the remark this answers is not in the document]
mauriciovasquezbernal: I'm not sure it can be easily split into different commits as all changes are related and to be honest I think it's too much work for little benefit.

mauriciovasquezbernal:
The TestRunInsecure is failing with:

```
Error: fetching gadget information: initializing and preparing operators: instantiating operator "oci": instantiating operator "wasm": initializing wasm: instantiating wasm: "fieldGet" is not exported in module "env"
```

This is an issue we have on the test setup: it's trying to run the latest version of the gadget instead of the one built on this PR, hence it fails because this PR introduces breaking changes and the logic now fails if the image has a wrong layer (#4108). I'll move ahead and merge it and open and issue / PR to handle the test case. (It'll work fine once this CI runs on main and pushes the updated image to the registry)

PR #4111, opened 2025-02-23
Fix use of latest tag

### LINKED ISSUE #4049: tree-wide: `latest` tag meaning
We've been pushing container and gadget images with the `latest` tag that follows the main branch:

https://github.com/inspektor-gadget/inspektor-gadget/blob/77582cbbf31d20b3f62e0ed00d3aea621637395f/.github/actions/set-container-repo-and-determine-image-tag/action.yml#L82-L84

After some thinking from https://github.com/inspektor-gadget/inspektor-gadget/pull/4035#pullrequestreview-2613119741, I feel we're using it the wrong way. Somebody using the latest tag doesn't expect to run a development version, but the latest released version.

I checked a couple of projects and the latest tag follows the latest release:
- cilium (uses stable instead of latest): https://hub.docker.com/layers/cilium/cilium/v1.17.1/images/sha256-a731981fef38429551dabdbc347b0f9af66a9f935bcc861117d63bc6681b3ec0 is the same as https://hub.docker.com/layers/cilium/cilium/stable/images/sha256-a731981fef38429551dabdbc347b0f9af66a9f935bcc861117d63bc6681b3ec0
- ubuntu: https://hub.docker.com/layers/library/ubuntu/noble/images/sha256-3afff29dffbc200d202546dc6c4f614edc3b109691e7ab4aa23d02b42ba86790 is the same as https://hub.docker.com/layers/library/ubuntu/latest/images/sha256-3afff29dffbc200d202546dc6c4f614edc3b109691e7ab4aa23d02b42ba86790 

## Website 

I think we have the same issue on the website, latest is pointing to the dev version instead of latest release.

## Implementation Options

- Use a `main` tag that follows the main branch on Github. (Useful for folks that want to try new features without waiting for a release)
- Update `latest` when we make a release to point to it

AdvH039:
Fixes #4049

### DIFF
--- a/.github/actions/set-container-repo-and-determine-image-tag/action.yml
+++ b/.github/actions/set-container-repo-and-determine-image-tag/action.yml
@@ -52,7 +52,13 @@ runs:
         # If GITHUB_REF_NAME is 'foo/bar', image_tag will be 'foo-bar', we need
         # this because it is not possible to have slash in image tag.
         image_tag=${GITHUB_REF_NAME//\//-}
+        image_event=${GITHUB_EVENT_NAME}
+        
         if [ "$image_tag" = "main" ]; then
+            image_tag="main"
+        fi
+
+        if [ "$image_event" = "release" ]; then
             image_tag="latest"
         fi
 
@@ -79,7 +85,13 @@ runs:
         # If GITHUB_REF_NAME is 'foo/bar', image_tag will be 'foo-bar', we need
         # this because it is not possible to have slash in image tag.
         gadget_tag=${GITHUB_REF_NAME//\//-}
+        gadget_event=${GITHUB_EVENT_NAME}
+        
         if [ "$gadget_tag" = "main" ]; then
+            gadget_tag="main"
+        fi
+
+        if [ "$gadget_event" = "release" ]; then
             gadget_tag="latest"
         fi
 


AdvH039:
Should there be some changes in the workflows also?

mauriciovasquezbernal:
@AdvH039 I don't think the current logic works fine. It uses GITHUB_EVENT_NAME to detect when a release is being done, but that actions is never call on the release workflow (.github/workflows/release.yml) as images aren't built there.

I see two options to implement this:

1. Add extra steps to tag and push latest for the different images around

https://github.com/inspektor-gadget/inspektor-gadget/blob/211d205f8b8c93be26c43d9cf88614b810ed5f51/.github/workflows/inspektor-gadget.yml#L478-L489

https://github.com/inspektor-gadget/inspektor-gadget/blob/211d205f8b8c93be26c43d9cf88614b810ed5f51/.github/workflows/inspektor-gadget.yml#L511-L522

etc.

2. Add a new step that combines all of that in the same place (i.e. it pulls all vx.y.z images, tags them to latest and pushes them to the registry)

I think way 1 is cleaner, but I'm not totally sure

mauriciovasquezbernal:
Replaced by #4179

PR #4118, opened 2025-02-24
ebpf-builder.Dockerfile: remove workaround on software-properties-common

alban:
The workaround was introduced in a64e9228d3c6 ("ebpf-builder.Dockerfile: workaround to install software-properties-common") #4084.

---

Let's check the CI to see if the error described in #4084 is fixed.


alban:
It seems to work without the workaround now. However, I can't find recent changes in Ubuntu packages that would explain it:
https://launchpad.net/ubuntu/+source/pyjwt
https://launchpad.net/ubuntu/+source/python-oauthlib

alban:
I don't know the root cause of the initial problem in Ubuntu packages. I merged the workaround in #4084 because it was blocking us in several PRs unrelated to the ebpf-builder.

This PR is just testing the CI without the workaround to see if the problem in Ubuntu packages is resolved. It seems that the CI passes now, so I think we can remove the workaround, although I still don't know the root cause.

PR #4129, opened 2025-02-25
operators/wasm: Add function to get syscall ID from syscall name.

eiffel-fl:


alban:
Could you share how it is going to be used? Is it for letting users filtering syscalls by name in traceloop?

pkg/operators/wasm/syscalls.go
```diff
@@ -60,6 +66,31 @@ func (i *wasmOperatorInstance) getSyscallName(ctx context.Context, m wapi.Module
...
+// - stack[0]
+// Return value:
+// - Syscall ID on success, 0xffffffffffffffff on error
+func (i *wasmOperatorInstance) getSyscallID(ctx context.Context, m wapi.Module, stack []uint64) {
+	syscallNamePtr := stack[0]
+
+	syscallName, err := stringFromStack(m, syscallNamePtr)
+	if err != nil {
+		i.logger.Warnf("getSyscallID: reading string from stack: %v", err)
+		stack[0] = 0xffffffffffffffff
```
alban, line 80: It is difficult to review if it has the correct amount of `f`...
Could you use `math.MaxUint64`?
Ditto in other places.

mauriciovasquezbernal: can't we use -1 and return an i32? It seems to be more aligned with other APIs. (I'd would expect the number of syscalls to be lower than 2^31 for quite some time, so an overflow shouldn't be an issue).

mauriciovasquezbernal:
Please don't forget to add the docs for this one.

pkg/operators/wasm/syscalls.go
mauriciovasquezbernal: This was marked as resolved but the code is still returning MaxUint..., any reason not to use -1?

PR #4122, opened 2025-02-24
update dockerhub reference

mauriciovasquezbernal:
Use images from own fork now that #4121 is in.
Fixes #4096

PR #4133, opened 2025-02-25
ci: Fix location of ci images

mauriciovasquezbernal:
Location should be ghcr.io/inspektor-gadget/ci/foo instead of ghcr.io/inspektor-gadget/inspektor-gadget/ci/foo.

Fixes: 3175b3db4bd6 ("ci: Mirror dockerhub images to our repo")

Ref https://github.com/inspektor-gadget/inspektor-gadget/pull/4122#discussion_r1969867592

### DIFF
--- a/.github/workflows/dockerhub-mirror.yml
+++ b/.github/workflows/dockerhub-mirror.yml
@@ -1,8 +1,12 @@
+# This workflow mirrors few container images we use on the CI from Dockerhub to
+# ghcr.io to avoid pull rate limit issues. Images are mirrored to
+# ghcr.io/inspektor-gadget/ci/<image-name>.
+#
 # based on https://github.com/rblaine95/dockerhub-mirror/
 name: Mirror Dockerhub
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dockerhub-mirror.yml
```diff
@@ -2,7 +2,7 @@
 name: Mirror Dockerhub
 env:
   REGISTRY: ghcr.io
-  CONTAINER_REPO: ${{ github.repository }}
+  CONTAINER_REPO: ${{ github.repository_owner }}
```
alban, line 5: You can add a comment giving an example of full image name for busybox.

alban (approved):
LGTM!

.github/workflows/dockerhub-mirror.yml
mauriciovasquezbernal: Added a top comment documenting the workflow and indicating where images are pushed to.

PR #4134, opened 2025-02-25
operators/env: introduce operator to create new fields from environment variables

flyth:
This operator lets you add new fields to datasources and fill it with contents of environment variables. For security reasons, environment variables that should be read by the operator need to be configured explicitly using the `--env-vars` parameter.

Annotations need to look like this:
```yaml
datasources:
  myds:
    annotations:
      env.fields.my-new-field: ENV_VAR_1
```

### DIFF
--- a/pkg/operators/env/env.go
+++ b/pkg/operators/env/env.go
@@ -1,0 +1,147 @@
+// Copyright 2024-2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package env
+
+import (
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+const (
+	Name     = "env"
+	Priority = 1
+
+	AnnotationPrefix = "env.fields."
+
+	ParamEnvVars = "env-vars"
+)
+
+type Config struct {
+	AllowedVars []string
+}
+
+type envOperator struct {
+	config Config
+}
+
+func (e *envOperator) Name() string {
+	return Name
+}
+
+func (e *envOperator) Init(params *params.Params) error {
+	e.config.AllowedVars = strings.Split(params.Get(ParamEnvVars).AsString(), ",")
+	return nil
+}
+
+func (e *envOperator) GlobalParams() api.Params {
+	return api.Params{
+		{
+			Key:         "env-vars",
+			Description: "Comma-separated list of environment variables to allow to be included",
+			Title:       "Allowed Environment Variables",
+		},
+	}
+}
+
+func (e *envOperator) InstanceParams() api.Params {
+	return nil
+}
+
+func (e *envOperator) InstantiateDataOperator(gadgetCtx operators.GadgetContext, instanceParamValues api.ParamValues) (operators.DataOperatorInstance, error) {
+	instance := &envOperatorInstance{
+		op:     e,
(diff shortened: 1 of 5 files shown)


pkg/operators/env/env.go
```diff
@@ -0,0 +1,147 @@
...
+	fields map[datasource.DataSource][]datasource.DataFunc
+}
+
+func (e *envOperatorInstance) Name() string {
+	return Name
+}
+
+func (e *envOperatorInstance) init(gadgetCtx operators.GadgetContext) error {
+	e.fields = make(map[datasource.DataSource][]datasource.DataFunc)
+	for _, ds := range gadgetCtx.GetDataSources() {
```
blanquicet, line 96: I think this should be done on the `PreStart` to ensure we are registering to all data sources as operators with lower prio might register data sources during instantiate phase.

pkg/operators/env/env.go
```diff
@@ -0,0 +1,147 @@
+// Copyright 2024-2025 The Inspektor Gadget authors
```
blanquicet, line 1: Could you please add some tests. Thanks!

blanquicet (commented):
LGTM already, just missing some testing.

pkg/operators/env/env.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, ds := range gadgetCtx.GetDataSources() {
+		annotations := ds.Annotations()
+		for k, v := range annotations {
+			fieldName, ok := strings.CutPrefix(k, AnnotationPrefix)
+			if !ok {
+				continue
+			}
+
+			// check, if the environment variable is allowed
+			if !slices.Contains(e.op.config.AllowedVars, v) {
```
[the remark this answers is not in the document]
flyth: Since environment variables are case sensitive, I think this should be too. Otherwise it's creating ambiguities.

pkg/operators/env/env.go
```diff
@@ -0,0 +1,147 @@
...
+			})
+		}
+	}
+	return nil
+}
+
+func (e *envOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error {
+	for ds, funcs := range e.fields {
+		for _, f := range funcs {
+			err := ds.Subscribe(f, 0)
```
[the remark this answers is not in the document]
flyth: Op's priority is used to decide when the operator is instantiated - subscription is deciding the order of event handling.

In this case the idea behind those numbers are:
* this operator will always only add fields to existing datasources (which usually are created on instantiation); so op priority should be grater than the default priority for operators that are creating datasources usually. But I also wanted to keep it as low as possible so that other enriching operators can pick it up - this made me set it to 1
* since it's static, we should write the data as soon as possible, which made me pick 0

I agree that we could probably use 1 for both, but 0 says "asap" and makes it a bit clearer.
I don't like the current priority system, it can become confusing quickly :/

(btw, now that I wrote this: actually, this operator later on could be an emitter for _all_ allowed environment variables through a dedicated datasource. But let's not go there now 😄)

pkg/operators/env/env.go
flyth: PreStart is too late. Datasourced and fields should be initialized on instantiation - otherwise it won't get picked up by `GetGadgetInfo`, which means it would be missing from column lists, as options for sorting etcpp.

pkg/operators/env/env.go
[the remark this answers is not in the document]
blanquicet: You're right.

pkg/operators/env/env.go
blanquicet: That's what I'm saying. Data sources will be registered in instantiation, but this op might lost those data sources registered on operators with lower priority.

blanquicet (approved):
LGTM!

pkg/operators/env/env.go
flyth: It still needs to be done on initialization - we could set priority to a higher number if you think it might conflict, but it still should be done on instantiation so the field ends up in the gadget info.

blanquicet: I think we should standardize that. All op register datasources on instantiate and all op get data source son PreStart.

blanquicet: > PreStart is too late. 
> 
Why?

flyth: > > PreStart is too late.
> 
> Why?

Frontends (e.g. cobra, now IG Desktop) rely on the info from `GetGadgetInfo` in order to show which fields there are - in order to sort, filter, etcpp. And `GetGadgetInfo` only instantiates and never calls other lifecycle operations. So with only the instantiation, datasources should have been initialized AND enriched (well, fields added etc.)

flyth: > I think we should standardize that. All op register datasources on instantiate and all op get data source son PreStart.

Still in progress in #3778 😄 

General rule of thumb currently is: instantiate for DS creation / manipulation, PreStart for subscriptions to not miss events and Start for emitting events.

flyth:
Thanks for the review!

Issue #4127, opened 2025-02-25
Bug: Runtime Image Enrichment (and ig filters) transiently not working


galofir-ms:
From the cluster we've seen evidence of some enrichers failing.

IG stderr exception: time="2025-02-17T18:02:05Z" level=error msg="cgroup enricher: failed to get cgroup paths on container 3876b8642d43dba2e2d4fee85bed83d5dfecbaa2c2780e6cda2ed971fcbb7cba: parsing cgroup: open /host/proc/10269/cgroup: no such file or directory"
IG stderr: time="2025-02-17T18:02:05Z" level=error msg="namespace enricher: failed to get mnt namespace on container 3876b8642d43dba2e2d4fee85bed83d5dfecbaa2c2780e6cda2ed971fcbb7cba: stat /host/proc/10269/ns/mnt: no such file or directory"
IG stderr: time="2025-02-17T18:02:05Z" level=warning msg="WithTracerCollection: failed to open mntns reference for container 3876b8642d43dba2e2d4fee85bed83d5dfecbaa2c2780e6cda2ed971fcbb7cba: no such file or directory"
IG stderr: time="2025-02-19T02:13:13Z" level=error msg="kubernetes enricher: failed to get owner reference for container ed3de14527ac9279ed3a1b3e6b49d717adeb29dc80a2444c739631b4bc92e4df: enriching owner reference: getting kube-system/pods/v1/aks-secrets-store-csi-driver-p7wjr owner reference: fetching pods/v1 kube-system/aks-secrets-store-csi-driver-p7wjr: Get \"https://esmp-c0j531rk.privatelink.southeastasia.azmk8s.io:443/api/v1/namespaces/kube-system/pods/aks-secrets-store-csi-driver-p7wjr\": dial tcp: lookup esmp-c0j531rk.privatelink.southeastasia.azmk8s.io: i/o timeout"
[...]

PR #4137, opened 2025-02-26
container-collection: initial containers: better logs for inexistent pid

alban:
Detect early when the container runtime gives an initial container with a pid that does not exist. This could happen if the container terminates while querying the container runtime, or if there is a misconfiguration and we can't query the container details in /host/proc/pid.

It could help with debugging #4127.

## Testing done

Start a test container:
```
docker run --name busyboxtest -ti --rm -v /run:/run -v /usr/bin/runc:/usr/bin/runc busybox
```

Start ig with the wrong configuration:

```
$ sudo HOST_ROOT=/proc/$(docker inspect busyboxtest | jq '.[0].State.Pid')/root/ ./ig list-containers -w
```
Output before the patch:
```
ERRO[0001] cgroup enricher: failed to get cgroup paths on container 7bc5bdded36f820e29bdad5f20f0141e9768919cad4b844b95ec7fa448ae673b: parsing cgroup: open /proc/101286/root/proc/101286/cgroup: no such file or directory 
ERRO[0001] namespace enricher: failed to get mnt namespace on container 7bc5bdded36f820e29bdad5f20f0141e9768919cad4b844b95ec7fa448ae673b: stat /proc/101286/root/proc/101286/ns/mnt: no such file or directory 
WARN[0001] WithTracerCollection: failed to open mntns reference for container 7bc5bdded36f820e29bdad5f20f0141e9768919cad4b844b95ec7fa448ae673b: no such file or directory 
```
Output after the patch:
```
WARN[0000] Runtime enricher (docker): Skip container "busyboxtest" (ID: 7bc5bdded36f820e29bdad5f20f0141e9768919cad4b844b95ec7fa448ae673b, image: busybox): PID 101286 doesn't exist 
```

### DIFF
--- a/pkg/container-collection/k8s.go
+++ b/pkg/container-collection/k8s.go
@@ -19,6 +19,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"os"
+	"path/filepath"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


blanquicet (approved):
LGTM!

Notice those two logs will be printed only for issues with already existing containers at start-up of IG. Does https://github.com/inspektor-gadget/inspektor-gadget/issues/4127 happen with new or already running containers?

alban:
> Notice those two logs will be printed only for issues with already existing containers at start-up of IG. Does #4127 happen with new or already running containers?

I am not sure.

For new containers, the `container-hook` package already checks /host/proc/pid [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/6cfd9d3a1a6b9a04d5efc48228cadbf1274868f4/pkg/container-hook/tracer.go#L551) since IG v0.35.0 (which is the version used in #4127).

PR #4138, opened 2025-02-26
snapshot_process: Increase timeout to fix test

mauriciovasquezbernal:
The TestSnapshotProcessGadget/captures_events_with_no_filter is failing a lot because it's not able to capture the event. Increase the timeout to be sure sleep lasts long enough to be captured by the gadget.


burak-ok:
Since it doesn't fail always: Maybe we should set the timeout to 10 or 15. Not 30?

mauriciovasquezbernal:
I rerrun the unit test job around 6 times and it always worked. I'd say it's fixing the flaky one.

burak-ok (approved):

PR #4142, opened 2025-02-27
gadgets: port profile_cpu to image based gadget

### LINKED ISSUE #3001: gadgets: Implement profile cpu as image-based one
TImplement the [profile cpu](https://github.com/inspektor-gadget/inspektor-gadget/tree/main/pkg/gadgets/profile/cpu) gadget as image-based one. 

TODO: How feasible is to port it? Does it need https://github.com/inspektor-gadget/inspektor-gadget/issues/2202?

flyth:
While converting the gadget was pretty straightforward, the actual program type allows many different settings to be applied. We need to discuss what and how we want to support those. I think it's wrong to just hardcode the values for our specific use-case (but I'm not familiar enough with the actual capabilities to decide).

The biggest question is: should we support just forwarding raw values (integers) from the configuration or abstract it with `count_sw_cpu_clock` etc. to always use the values defined in the `unix` package (for compatibility reasons)?

~~Also, this will make more sense after the addition of `flush-on-stop` in #4089.~~

The gadget now uses `flush-on-stop`, so it currently also needs `--map-fetch-interval 0s` set.

Fixes #3001

### DIFF
--- a/docs/gadget-devel/program-types.md
+++ b/docs/gadget-devel/program-types.md
@@ -47,8 +47,28 @@ You can find the list of iterator types supported by Linux with:
 
 #### Fentry / Fexit
 
-The section name must use the `fentry/<function_name` or `fexit/<function_name>`. As in kprobes,
-`<function_name>` is the kernel function that the kprobe will be attached to.
+The section name must use the `fentry/<function_name>` or `fexit/<function_name>`. As in kprobes,
+`<function_name>` is the kernel function that the program will be attached to.
+
+### PerfEvents
+
+The section name must be `perf_event/<name>`, where `<name>` is used to apply parameters to the
+program using the `gadget.yaml` file.
+
+Currently, we only support the following settings (`<name>` is `myPerfEvent` in this case):
+
+```yaml
+programs:
+  myPerfEvent:
+    perf:
+      type: software
+      config: count_sw_cpu_clock
+      sampleType: sample_raw
+    sampler:
+      frequency: 49
+```
+
+All parameters are mandatory for now.
 
 ### Raw Tracepoints
 
--- a/docs/gadgets/profile_cpu.mdx
+++ b/docs/gadgets/profile_cpu.mdx
@@ -1,0 +1,125 @@
+---
+title: profile_cpu
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# profile_cpu
+
+The profile cpu gadget takes samples of stack traces.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_cpu:%IG_TAG% --map-fetch-interval 0 [flags]
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/profile_cpu:%IG_TAG% --map-fetch-interval 0 [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Guide
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        Here we deploy a small demo pod "random":
+
+        ```bash
+        $ kubectl run --restart=Never --image=busybox random -- sh -c 'cat /dev/urandom > /dev/null'
+        pod/random created
+        ```
+
+        Using the profile cpu gadget, we can see the list of stack traces.
+        The following command filters only for pods named "random", execute the command
+        and interrupt it after ~30 seconds. The `--kernel-stacks-only` option is passed to show only the
+        kernel stack traces.
+
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_cpu:%IG_TAG% --podname random --kernel-stacks-only --map-fetch-interval 0
+        ```
+
+        After a while press with Ctrl-C to stop trace collection
+
+        ```
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_cpu:%IG_TAG% --timeout 5 --podname random --kernel-stacks-only
+        K8S.NODE                 K8S.NAMESPACE            K8S.PODNAME             K8S.CONTAINERNAME              PID COMM             SAMPLES              KERN_STACK
+        minikube-docker          default                  random                  random                       38130 containerd-shim  5                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  2                    [0]chacha_block_generic;…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  2                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  2                    [0]chacha_permute; [1]ge…
+        ```
+
+        From the traces above, you can see that the pod is spending CPU time in the
+        Linux function `urandom_read`.
+
+        Instead of waiting, you can use the `--timeout` argument:
+
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_cpu:%IG_TAG% --timeout 5 --podname random --kernel-stacks-only --map-fetch-interval 0
+        K8S.NODE                 K8S.NAMESPACE            K8S.PODNAME             K8S.CONTAINERNAME              PID COMM             SAMPLES              KERN_STACK
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]_copy_to_iter; [1]get…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  2                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  5                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]chacha_permute; [1]ge…
+        minikube-docker          default                  random                  random                       38130 containerd-shim  1                    [0]get_random_bytes_user…
+        ```
+
+        Finally, we need to clean up our pod:
+
+        ```bash
+        $ kubectl delete pod random
+        ```
+    </TabItem>
+    <TabItem value="ig" label="ig">
+        * Generate some kernel load:
+
+        ```bash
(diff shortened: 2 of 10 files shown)


pkg/operators/ebpf/attach.go
```diff
@@ -145,6 +149,91 @@ func (i *ebpfInstance) attachProgram(gadgetCtx operators.GadgetContext, p *ebpf.
...
+			}
+		}
+		for cpu := 0; cpu < runtime.NumCPU(); cpu++ {
+			fd, err := unix.PerfEventOpen(
+				&unix.PerfEventAttr{
+					Type:        perfType,
+					Config:      perfConfig,
+					Sample_type: perfSampleType,
+					Sample:      frequency,
+					Bits:        1 << 10,
```
alban, line 214: Define as a constant, as it is done in pkg/gadgets/profile/cpu/tracer/tracer.go:
```
const (
        // In C, struct perf_event_attr has a freq field which is a bit in a
        // 64-length bitfield.
        // In Golang, there is a Bits field which 64 bits long.
        // From C, we can deduce freq (which permits using frequency not period)
        // is the 10th bit.
        frequencyBit = 1 << 10
)
```

pkg/operators/ebpf/attach.go
```diff
@@ -145,6 +149,91 @@ func (i *ebpfInstance) attachProgram(gadgetCtx operators.GadgetContext, p *ebpf.
...
+					perfConfig = val
+				}
+			}
+			if tmp := i.config.GetString("programs." + name + ".perf.sampleType"); tmp != "" {
+				switch tmp {
+				case "sample_raw":
+					perfSampleType = unix.PERF_SAMPLE_RAW
+				default:
+					// Try to get raw value // TODO: should we support it in this way as well?
+					val, err := strconv.ParseUint(tmp, 10, 64)
```
alban, line 192: I am not sure if we should support config and type in a generic way. I don't know if all config & type combinations are meant to be used with ebpf.

I would add those support only if we tested that it works. Otherwise, we can start with supporting just the basic PERF_SAMPLE_RAW + PERF_COUNT_SW_CPU_CLOCK.

pkg/operators/ebpf/attach.go
```diff
@@ -145,6 +149,91 @@ func (i *ebpfInstance) attachProgram(gadgetCtx operators.GadgetContext, p *ebpf.
...
+			}
+			if tmpFrequency := i.config.GetString("programs." + name + ".sampler.frequency"); tmpFrequency != "" {
+				var err error
+				frequency, err = strconv.ParseUint(tmpFrequency, 10, 64)
+				if err != nil {
+					return nil, fmt.Errorf("parsing frequency %q for program %q: %w", tmpFrequency, name, err)
+				}
+			}
+		}
+		for cpu := 0; cpu < runtime.NumCPU(); cpu++ {
```
alban, line 207: Iterating on cpu is fine for PERF_COUNT_SW_CPU_CLOCK.

But for PERF_COUNT_SW_PAGE_FAULTS and others, do we need a perf fd for each cpu?

If we don't know the answer, I would rather start with basic support instead of supporting more things if we are not sure it is working.

pkg/operators/ebpf/attach.go
flyth: Yeah, I was making it configurable as well, but then stopped to discuss here 😄 If we keep it fixed, I'll revert it to using the constant w/ explanation ofc.

pkg/operators/ebpf/attach.go
flyth: Hmm. Tbh, if it doesn't crash (or is harmful in other ways), I'd rather keep it than having to make potential users of those other combinations wait for at least another release cycle of IG. Seems we should invest some time to investigate other use cases.

pkg/operators/ebpf/attach.go
flyth: Good point. As said: feels like we need to investigate a bit to see how this is being used in other scenarios.

gadgets/profile_cpu/program.bpf.c
```diff
@@ -0,0 +1,115 @@
...
+#endif /* __TARGET_ARCH_arm64 || __TARGET_ARCH_x86 */
+
+SEC("perf_event/profiler")
+int ig_prof_cpu(struct bpf_perf_event_data *ctx)
+{
+	u64 id = bpf_get_current_pid_tgid();
+	u32 pid = id >> 32;
+	u32 tid = id;
+	u64 *valp;
+	static const u64 zero = 0;
```
alban, line 71:
```suggestion
	struct values *valp;
	static const struct values zero = {0,};
```

gadgets/profile_cpu/program.bpf.c
```diff
@@ -0,0 +1,115 @@
...
+		u64 ip = PT_REGS_IP(&ctx->regs);
+
+		if (is_kernel_addr(ip)) {
+			key.kernel_ip = ip;
+		}
+	}
+
+	valp = bpf_map_lookup_or_try_init(&counts, &key, &zero);
+	if (valp)
+		__sync_fetch_and_add(valp, 1);
```
alban, line 110:
```suggestion
		__sync_fetch_and_add(&valp->samples, 1);
```

docs/gadget-devel/program-types.md
```diff
@@ -47,9 +47,29 @@ You can find the list of iterator types supported by Linux with:
 
 #### Fentry / Fexit
 
-The section name must use the `fentry/<function_name` or `fexit/<function_name>`. As in kprobes,
+The section name must use the `fentry/<function_name>` or `fexit/<function_name>`. As in kprobes,
 `<function_name>` is the kernel function that the kprobe will be attached to.
```
alban, line 51:
```suggestion
The section name must use the `fentry/<function_name>` or `fexit/<function_name>`. As in kprobes,
`<function_name>` is the kernel function that the program will be attached to.
```

alban:
I started adding missing bits:
https://github.com/alban/inspektor-gadget/commits/alban_profile_cpu/

gadgets/profile_cpu/gadget.yaml
```diff
@@ -0,0 +1,36 @@
...
+    perf:
+      type: software
+      config: count_sw_cpu_clock
+      sampleType: sample_raw
+    sampler:
+      frequency: 49
+datasources:
+  samples:
+    annotations:
+      views.modes.flamegraph: true
```
mauriciovasquezbernal, line 17: Are these specific for ig desktop?

mauriciovasquezbernal (commented):
Some comments. I think it'll be better to split the commit into one introducing the support in the framework for perf_event programs and another one introducing the gadget.

gadgets/profile_cpu/gadget.yaml
flyth: Yes :)

gadgets/profile_cpu/program.bpf.c
```diff
@@ -0,0 +1,114 @@
...
+SEC("perf_event/profiler")
+int ig_prof_cpu(struct bpf_perf_event_data *ctx)
+{
+	if (gadget_should_discard_data_current())
+		return 0;
+
+	u64 mntns_id;
+	mntns_id = gadget_get_current_mntns_id();
+	if (gadget_should_discard_mntns_id(mntns_id))
+		return 0;
```
mauriciovasquezbernal, line 75: It shouldn't be needed as `gadget_should_discard_data_current` includes it.

pkg/operators/ebpf/attach.go
```diff
@@ -145,6 +153,88 @@ func (i *ebpfInstance) attachProgram(gadgetCtx operators.GadgetContext, p *ebpf.
...
+			frequency, err = strconv.ParseUint(tmpFrequency, 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("parsing frequency %q for program %q: %w", tmpFrequency, name, err)
+			}
+		} else {
+			return nil, fmt.Errorf("sampler.frequency not specified for program %q", name)
+		}
+		if frequency == 0 {
+			return nil, fmt.Errorf("sampler.frequency is zero for program %q", name)
+		}
```
mauriciovasquezbernal, line 207: What's the plan regarding this? To remove all the configuration options and leave hardcoded the case we know that works or to keep it configurable?

flyth: The plan is to have the gadget author specify __all__ parameters as needed (not relying on defaults that could change later on) in order to keep forward compatibility (e.g.: if we choose to add other options later on, existing gadget (hopefullly) won't need an update).

gadgets/profile_cpu/README.md
```diff
@@ -0,0 +1,5 @@
+# profile_cpu
+
+The `profile_cpu` gadget that profiles the CPU.
```
mauriciovasquezbernal, line 3:
```suggestion
The `profile_cpu` gadget profiles the CPU.
```

pkg/operators/ebpf/attach.go
```diff
@@ -145,6 +153,88 @@ func (i *ebpfInstance) attachProgram(gadgetCtx operators.GadgetContext, p *ebpf.
...
+		if tmp := i.config.GetString("programs." + name + ".perf.type"); tmp != "" {
+			switch tmp {
+			case "software":
+				perfType = unix.PERF_TYPE_SOFTWARE
+			default:
+				return nil, fmt.Errorf("unsupported perf.type %q", tmp)
+			}
+		} else {
+			return nil, fmt.Errorf("perf.type not specified for program %q", name)
+		}
```
mauriciovasquezbernal, line 175: nit: can you consolidate these two on the same switch case?

(same in all places below)

```
case "":
    return nil, fmt.Errorf("perf.type not specified for program %q", name)
```

docs/gadgets/profile_cpu.mdx
mauriciovasquezbernal, line 1: Do you think we can provide a guide like in other gadgets?

mauriciovasquezbernal (approved):
LGTM besides some minor nits.

pkg/operators/ebpf/attach.go
alban: IMO it is not necessary to force gadget authors to specify all parameters for forward compatibility.
For example: ".perf.type=software" could be hardcoded. And if we later want to make it configurable, we could have ".perf.type" default to "software". So older gadgets without this configuration would still work.

That would allow to simplify the gadget author job, and those lines could be deleted:
```
programs:
  profiler:
    perf:
      type: software
      config: count_sw_cpu_clock
      sampleType: sample_raw
    sampler:
      frequency: 49
```

pkg/operators/ebpf/attach.go
```diff
@@ -145,6 +153,82 @@ func (i *ebpfInstance) attachProgram(gadgetCtx operators.GadgetContext, p *ebpf.
...
+			return nil, fmt.Errorf("sampler.frequency is zero for program %q", name)
+		}
+		for cpu := 0; cpu < runtime.NumCPU(); cpu++ {
+			fd, err := unix.PerfEventOpen(
+				&unix.PerfEventAttr{
+					Type:        perfType,
+					Config:      perfConfig,
+					Sample_type: perfSampleType,
+					Sample:      frequency,
+					Bits:        frequencyBit,
```
alban, line 209:
```suggestion
					Bits:        unix.PerfBitFreq,
```

See:
https://pkg.go.dev/golang.org/x/sys/unix
```go
const PerfBitFreq uint64 = CBitFieldMaskBit10
```

Then, we could delete our constant `frequencyBit`.

pkg/operators/ebpf/attach.go
alban: I also don't have a super strong opinion on this.

About the setting "frequency=49", I wonder if this could be a parameter on the ebpf operator (chosen at runtime) instead of a configuration option in the gadget (chosen at build-time).

flyth:
Thanks for the reviews!

PR #4146, opened 2025-02-28
otel-logs + otel-metrics: add support for static attributes/keys

flyth:
With this, static attributes can be added as annotations to datasources like this:
logs.attribute.ATTRIBUTENAME.value=content
metrics.key.KEYNAME.value=content

By default, values are evaluated as strings, but we also support int64 and float64. To use those, add another annotation like this:

logs.attribute.ATTRIBUTENAME.type=int64
logs.attribute.ATTRIBUTENAME.value=100
metrics.key.KEYNAME.type=int64
metrics.key.KEYNAME.value=100


pkg/operators/otel-metrics/otel-metrics.go
```diff
@@ -675,6 +682,26 @@ func (m *otelMetricsOperatorInstance) PreStart(gadgetCtx operators.GadgetContext
...
+			AnnotationKeysTypeSuffix,
+			attribute.Int64Value,
+			attribute.Float64Value,
+			attribute.StringValue,
+		)
+		if err != nil {
+			return fmt.Errorf("preparing static keys: %w", err)
+		}
+		for _, fn := range staticValFns {
+			k, v := fn()
```
mauriciovasquezbernal, line 699: Why does `GetKeyValueFuncsFromAnnotations` need to return lambda functions if it's evaluated right after on this loop?

PR #4147, opened 2025-02-28
otel: allow overriding key/name

flyth:
This allows properly overriding the keys of metrics/log fields with for example a name coming from an annotation.

This is only used for otel-logs currently, but might be added to otel-metrics in the future.


burak-ok (approved):
LGTM

PR #4148, opened 2025-02-28
trace_dns: Ignore fragmented packets

mqasimsarfraz:
Ignore fragmented packets early. Also, couple of improvements.

### Testing Done:

Make a request with large message
```
$ dig -4 @localhost +tcp fake.test.com.

; <<>> DiG 9.18.34 <<>> -4 @localhost +tcp fake.test.com.
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 19742
;; flags: qr rd; QUERY: 1, ANSWER: 500, AUTHORITY: 0, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;fake.test.com.                 IN      A
...
...
...
;; ANSWER SECTION:
;; Query time: 2 msec
;; SERVER: 127.0.0.1#53(localhost) (TCP)
;; WHEN: Thu Feb 27 10:06:27 UTC 2025
;; MSG SIZE  rcvd: 8031
```

No errors are printed by WASM layer as here: https://github.com/inspektor-gadget/inspektor-gadget/pull/4088

```
$  go run -exec sudo ./cmd/ig run ghcr.io/inspektor-gadget/gadget/trace_dns:qasim-check-dns-tcp --verify-image=false
5394f06315dbf73368946aab7084512529690aff67bca636bdcc0860619d04ed   127.0.0.1:40207                                      127.0.0.1:53                                         isc-net-0000          55246      55247 Q     A                                    fake.test.com.                                                                                                                                                  0ns
5394f06315dbf73368946aab7084512529690aff67bca636bdcc0860619d04ed   127.0.0.1:40207                                      127.0.0.1:53                                         dnstester             54584      54584 Q     A                                    fake.test.com.                                                                                                                                                  0ns
```

cc: @blanquicet

### DIFF
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -281,13 +281,13 @@ int ig_trace_dns(struct __sk_buff *skb)
 		return 0;
 
 	// Calculate the DNS offset in the packet
+	struct tcphdr tcph;
 	switch (proto) {
 	case IPPROTO_UDP:
 		dns_off = l4_off + sizeof(struct udphdr);
 		break;
 	case IPPROTO_TCP:
 		// This is best effort, since we don't reassemble TCP segments.
-		struct tcphdr tcph;
 		if (bpf_skb_load_bytes(skb, l4_off, &tcph, sizeof tcph))
 			return 0;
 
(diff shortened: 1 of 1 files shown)


gadgets/trace_dns/program.bpf.c
```diff
@@ -300,7 +300,16 @@ int ig_trace_dns(struct __sk_buff *skb)
 		if (skb->len <= dns_off)
 			return 0;
 
-		// DNS is after the TCP header and the 2 bytes of the length of the DNS packet
+		// DNS messages over TCP are prefixed with a two byte length field which
```
blanquicet, line 303: Mmmm I analyzed it more and I think we can do something better. Given that TCP is a stream-based protocol, if there is a big DNS query/response, the fragmentation will not happen at the TCP level, but at the IP level. So, I was wondering if we can check the _Fragment Offset Field_ in the IP header to identify whether the packet is a fragment, i.e., if this field is non-zero, it indicates that the packet is a fragment. WDYT?

mauriciovasquezbernal:
@mqasimsarfraz what should we do with this PR?

mqasimsarfraz:
> @mqasimsarfraz what should we do with this PR?

I am not sure if we need this change so I will propose to close this? Given tcp support has been there for a bit and nobody requested for this! 

Also, we should rather look into supporting reassembling TCP packets (if needed)

mauriciovasquezbernal:
> I will propose to close this? 

Agree.

Issue #3177, opened 2024-07-17
ci: Test summaries are broken after upgrating to upload-artifact@v4


mauriciovasquezbernal:
Our test summary report logic is broken:

> Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

(from https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/9979389630/job/27578945097)

It happens because upload-artifact@v3 [allowed](https://github.com/actions/upload-artifact/tree/v3?tab=readme-ov-file#uploading-to-the-same-artifact) to use the same artifact name to combine files coming from multiple jobs, however it's not allowed on v4 (https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes). 

Ref #2306.

PR #4157, opened 2025-03-03
ci: Use different name for test-report artifacts

mqasimsarfraz:
As mentioned here https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes we shouldn't be using same artifact name for a workflow run. This change uses the suffix to ensure we don't use the same artifact name for all the reports.

Fixes: #3177


mqasimsarfraz:
Thanks for the review! The failing annotations are gone!

PR #4161, opened 2025-03-03
operators/cli: allow setting output-mode to "none" from annotation

mauriciovasquezbernal:
Extracted from #4089

--- 

@flyth you're the original author of this commit. Could you please clarify why you think it should be discussed and what the alternatives are?

### DIFF
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -58,8 +58,15 @@ const (
 	// output modes for a DataSource in a comma-separated list.
 	AnnotationSupportedOutputModes = "cli.supported-output-modes"
 
+	// AnnotationOutputMode can be used to specify the output mode for a
+	// DataSource.
+	AnnotationOutputMode = "cli.output-mode"
+
 	// AnnotationDefaultOutputMode can be used to specify the default output mode for a DataSource.
 	AnnotationDefaultOutputMode = "cli.default-output-mode"
+
+	// OutputModeNone can be used to disable output for a DataSource.
+	OutputModeNone = "none"
 )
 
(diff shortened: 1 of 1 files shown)


flyth:
> Extracted from #4089
> 
> @flyth you're the original author of this commit. Could you please clarify why you think it should be discussed and what the alternatives are?

I think I added that note because this feature (although using params instead of an annotation) had been removed from the initial refactoring PR or shortly after. I still think it's something we really need 😄 

We just need to make sure it behaves as expected when used in combination with `-o`. The way it is implemented right now, it prevents the data source from ever showing up as an option for `-o`, right? I think it's okay that way.

mauriciovasquezbernal:
Right, the data source is ignored by the cli operator, using `-o json` doesn't print any reason. I'd say that's the expected result, right?

I'm wondering if `cli.output-mode` and `cli.default-output-mode` can be confusing. Perhaps `cli.output-mode`  could be `cli.ignore` or `cli.disable`

pkg/operators/cli/clioperator.go
```diff
@@ -58,8 +58,15 @@ const (
 	// output modes for a DataSource in a comma-separated list.
 	AnnotationSupportedOutputModes = "cli.supported-output-modes"
 
+	// AnnotationOutputMode can be used to specify the output mode for a
+	// DataSource.
+	AnnotationOutputMode = "cli.output-mode"
```
blanquicet, line 63: I'm unsure this annotation is necessary. Maybe we can simply set the `cli.supported-output-modes` as `ModeNone` when we want the clioperator to complete ignore a data source.

mauriciovasquezbernal:
After all it was a good idea to extract some commits from #4089. I don't think we need this PR, `cli.supported-output-modes: none` works fine. The implementation is a bit different from 2a828a9921a6f5f50e261dbfb55b35911d867695, but I think the final result for the user is (almost) the same.

PR #4165, opened 2025-03-04
wasm: Use cache to speed-up things

mqasimsarfraz:
After switching to golang in guest (previously tinygo) the compile
time (InstantiateWithConfig) for wasm modules has noticeably increased.
Wazero allows using a cache to avoid compiling these modules everytime.
The cache is key is based of SHA of wasm program and cache directory uses
wazero version to ensure no surprises if IG changes wazero version. The change uses the cache across different
gadgets runs to speed-up starting a gadget. 

### v0.38.0

```
$ time sudo -E ig run trace_dns:v0.38.0 --pull always --verify-image=false --timeout 1
RUNTIME.CONTAINERNAME                                            SRC                                                  DST                                                  COMM                    PID        TID QR    QTYPE                              NAME                                                             RCODE            ADDRESSES                                                         LATENCY_NS

real    0m4,274s
user    0m0,003s
sys     0m0,008s
```

### main

```
$ time sudo -E ./ig run trace_dns:main --pull always --verify-image=false --timeout 1
RUNTIME.CONTAINERNAME                                            SRC                                                  DST                                                  COMM                    PID        TID QR    QTYPE                              NAME                                                             RCODE            ADDRESSES                                                         LATENCY_NS

real    0m5,320s
user    0m0,004s
sys     0m0,007s
```

### PR

```bash
# ensure to remove cache directory
$ sudo rm -rf /var/run/ig
$ time sudo -E ./ig run trace_dns:main --pull always --verify-image=false --timeout 1
RUNTIME.CONTAINERNAME                                            SRC                                                  DST                                                  COMM                    PID        TID QR    QTYPE                              NAME                                                             RCODE            ADDRESSES                                                         LATENCY_NS

real    0m4,125s
user    0m0,004s
sys     0m0,009s

# second time things should be little more faster
$ time sudo -E ./ig run trace_dns:main --pull always --verify-image=false --timeout 1
RUNTIME.CONTAINERNAME                                            SRC                                                  DST                                                  COMM                    PID        TID QR    QTYPE                              NAME                                                             RCODE            ADDRESSES                                                         LATENCY_NS

real    0m3,448s
user    0m0,002s
sys     0m0,009s

```

### DIFF
--- a/pkg/operators/wasm/wasm.go
+++ b/pkg/operators/wasm/wasm.go
@@ -48,9 +48,25 @@ const (
 
 	// Indicates the handle encodes a member of a data array as index << 16 | arrayHandle
 	dataArrayHandleFlag = uint32(1 << 31)
+
+	// cache path for the wasm compilation
+	cacheDir = "/var/run/ig/wasm-cache"
 )
 
-type wasmOperator struct{}
+type wasmOperator struct {
+	cache wazero.CompilationCache
+}
+
+func newWasmOperator() *wasmOperator {
+	cache, err := wazero.NewCompilationCacheWithDir(cacheDir)
+	if err != nil {
+		logger.DefaultLogger().Debugf("failed to create wasm compilation cache: %v", err)
+		return &wasmOperator{}
+	}
+	return &wasmOperator{
+		cache: cache,
+	}
+}
 
 func (w *wasmOperator) Name() string {
 	return "wasm"
@@ -76,7 +92,7 @@ func (w *wasmOperator) InstantiateImageOperator(
 		createdMap:  map[uint32]struct{}{},
 	}
 
-	if err := instance.init(gadgetCtx, target, desc); err != nil {
+	if err := instance.init(gadgetCtx, target, desc, w.cache); err != nil {
 		instance.close(gadgetCtx)
 		return nil, fmt.Errorf("initializing wasm: %w", err)
 	}
@@ -212,11 +228,13 @@ func (i *wasmOperatorInstance) init(
 	gadgetCtx operators.GadgetContext,
 	target oras.ReadOnlyTarget,
 	desc ocispec.Descriptor,
+	cache wazero.CompilationCache,
 ) error {
 	ctx := gadgetCtx.Context()
 	rtConfig := wazero.NewRuntimeConfig().
 		WithCloseOnContextDone(true).
-		WithMemoryLimitPages(256) // 16MB (64KB per page)
+		WithMemoryLimitPages(256). // 16MB (64KB per page)
+		WithCompilationCache(cache)
 	i.rt = wazero.NewRuntimeWithConfig(ctx, rtConfig)
 
 	igModuleBuilder := i.rt.NewHostModuleBuilder("ig")
@@ -354,5 +372,5 @@ func (i *wasmOperatorInstance) close(gadgetCtx operators.GadgetContext) error {
 }
 
 func init() {
-	operators.RegisterOperatorForMediaType(wasmObjectMediaType, &wasmOperator{})
+	operators.RegisterOperatorForMediaType(wasmObjectMediaType, newWasmOperator())
 }


pkg/operators/wasm/wasm.go
```diff
@@ -50,7 +50,9 @@ const (
 	dataArrayHandleFlag = uint32(1 << 31)
 )
 
-type wasmOperator struct{}
+type wasmOperator struct {
+	cache wazero.CompilationCache
```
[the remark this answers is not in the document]
mqasimsarfraz: >  Does it keep growing when ig is used in daemon mode and users try various gadgets?

IIUC yes it will keep growing for different gadgets. Some more information on cache: https://github.com/tetratelabs/wazero/blob/cc40233f33197942e79997c3e6fccbe453ddb535/config.go#L129

> If so, would it be possible to have a limit on memory usage, or a limit in time? And make it configurable via some parameters on the wasm operator?

I am not able to find an option to configure it for the runtime but as per [wazero slack](https://gophers.slack.com/archives/C040AKTNTE0/p1740885450483399) following are our options:

> how to auto clean wam compilation cache:
> > There's no provision to clean a [CompilationCache](https://pkg.go.dev/github.com/tetratelabs/wazero#CompilationCache).
If you mean the in memory one, just create a NewCompilationCache, and create new Runtimes with it.
If you mean the file caches (NewCompilationCacheWithDir), just delete the directory (os.RemoveAll).

Based on this I am thinking if we should refresh the cache after intervals rather keeping a global cache for operator. Do you have any suggestions?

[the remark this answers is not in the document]
burak-ok: > Does it keep growing when ig is used in daemon mode and users try various gadgets?

I couldn't really measure out precisely the amount of memory it would still hold. I would say around 2-3MB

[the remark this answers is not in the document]
mqasimsarfraz: After testing the changes bit more it seems the modules are removed from the cache, once [close is called](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/wasm/wasm.go#L333), at gadget stop: :
- Reading the module from Cache: https://github.com/tetratelabs/wazero/blob/main/internal/engine/wazevo/engine_cache.go#L100
- Deleting the module from Cache: https://github.com/tetratelabs/wazero/blob/main/internal/engine/wazevo/engine.go#L514
- Deletion triggered once module Close: https://github.com/tetratelabs/wazero/blob/main/config.go#L382 

The reason cache was growing from me was, we never call Stop (hence Close) for the first time we initialize runtime/module. Need to think of a way to do it and then we should be good!

pkg/operators/wasm/cache.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func (c *compilationCache) refreshCache() {
+	c.refreshMu.Lock()
+	defer c.refreshMu.Unlock()
+
+	if time.Since(c.refreshTime) > time.Hour {
+		if c.cache != nil {
+			c.cache.Close(context.Background())
```
burak-ok, line 45: Can it happen that a gadget that is already running now has problems when calling guest functions? Or is the cache only used while initializing the module as a whole?

mqasimsarfraz: > Or is the cache only used while initializing the module as a whole?

Yes, it is only used during initializing: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/wasm/wasm.go#L253

pkg/operators/wasm/cache.go
```diff
@@ -0,0 +1,50 @@
...
+		cache:       wazero.NewCompilationCache(),
+		refreshTime: time.Now(),
+	}
+}
+
+func (c *compilationCache) refreshCache() {
+	c.refreshMu.Lock()
+	defer c.refreshMu.Unlock()
+
+	if time.Since(c.refreshTime) > time.Hour {
```
alban, line 43: 1 hour seems long... Would 1 minute be enough?

Initially we wanted to avoid the double compilation between the GetGadgetInfo and the RunGadget methods.

It can be a constant.

mqasimsarfraz: Done!

pkg/operators/wasm/wasm.go
```diff
@@ -76,7 +78,10 @@ func (w *wasmOperator) InstantiateImageOperator(
 		createdMap:  map[uint32]struct{}{},
 	}
 
-	if err := instance.init(gadgetCtx, target, desc); err != nil {
+	// Refresh the cache to avoid using stale data
```
mauriciovasquezbernal, line 81: Could you please describe a bit more when stale data could be used?

Also, better to move this to init()?

pkg/operators/wasm/wasm.go
```diff
@@ -230,6 +237,7 @@ func (i *wasmOperatorInstance) init(
 	i.addPerfFuncs(env)
 	i.addKallsymsFuncs(env)
 
+	cc.refreshMu.Lock()
```
mauriciovasquezbernal, line 240: 1. What is lock protecting from? (cache.Close() being called before this module is initialized?)
2. It's never released if any of the returns below are executed 
3. It seems it's covering a big zone. Is it possible to use a rw lock to avoid multiple "wasm" modules being compiled in parallel?

mqasimsarfraz:
I changed the approach of the PR, I wasn't too happy on how I was cleaning the cache especially in future we will [`Close`](https://github.com/inspektor-gadget/inspektor-gadget/pull/3778) to better handle things. Also, I realized `wazero` has file system based (I am using `tmpfs` backed location) cache which is better suited to help with gadget startup including caching it across multiple runs. 

@alban @burak-ok can you please have another look?

pkg/operators/wasm/wasm.go
```diff
@@ -212,11 +228,15 @@ func (i *wasmOperatorInstance) init(
...
 	desc ocispec.Descriptor,
+	cache wazero.CompilationCache,
 ) error {
 	ctx := gadgetCtx.Context()
 	rtConfig := wazero.NewRuntimeConfig().
 		WithCloseOnContextDone(true).
 		WithMemoryLimitPages(256) // 16MB (64KB per page)
+	if cache != nil {
+		rtConfig = rtConfig.WithCompilationCache(cache)
+	}
```
mauriciovasquezbernal, line 239: nit: I think you can safely pass nil to `WithCompilationCache`, so the conditional shouldn't be needed.

mauriciovasquezbernal (approved):
I new approach is much cleaner, thanks!

pkg/operators/wasm/wasm.go
mqasimsarfraz: Good catch! It can indeed handle it: https://github.com/tetratelabs/wazero/blob/main/runtime.go#L172!

mqasimsarfraz:
Thanks for the reviews!

Issue #3387, opened 2024-08-22
[RFE] Inspect a gadget in details


alban:
https://mentorship.lfx.linuxfoundation.org/project/a6d66c40-3d12-4fa4-88bf-18574f6b4ec0

---

## Current situation

A gadget can be inspected with the `ig image inspect` command (implemented in #3382). But it does not give all possible details.

See https://github.com/inspektor-gadget/inspektor-gadget/pull/3382#pullrequestreview-2250296366

## Impact

When gadget developers are trying to debug their gadgets, they cannot easily inspect the different layers, decompile, list the ebpf programs and ebpf maps etc.

## Ideal future situation

It should be possible to get the following information from a gadget:

- list of architectures supported by the gadget
- list of layers
- information from gadget.yaml:
  - list of ebpf params
  - list of data sources with the list of fields
- information from the ebpf module
  - list of sections (so I can see kprobe/sys_mount)
  - list of ebpf maps with their types (so I don't have to use readelf and objdump)
  - show bpf bytecode + source (when available in ELF sections) of requested ebpf programs
- information from the wasm module
  - what `gadgetAPIVersion` does it use?
  - which wasm upcalls does it make?

## Implementation options

Additionally, it would be nice to have an implementation both in `ig` and in a wasm module to be reusable on websites. For the second case, this probably means we can't use the cilium/ebpf package.

## Additional information

I gave it a try with wasm and came up with the following website:

https://github.com/alban/bpfanalyser
https://alban.github.io/bpfanalyser/

![Image](https://github.com/user-attachments/assets/d1546bfc-eb64-497f-b759-3c9a4c78eb37)

![Image](https://github.com/user-attachments/assets/a6265836-2f23-47bc-864e-5f427022b3e5)

Useful changes in cilium/ebpf: https://github.com/cilium/ebpf/pull/1552

sukhman-sukh:
Hello,
My name is Sukhman Singh. I am a low-level software engineer and open-source contributor at @sdslabs. I was also a Google Summer of Code '24 fellow at Qaul.net.
I have extensive experience with Rust, Golang, and C and hands-on experience in eBPF development using both C and Rust. I am highly interested in the project's vision and the above issue. I also have some ideas on how `ig image inspect` can be used for more verbose and detailed information and layers-by-layer analysis
I am planning to apply for LFX'25 Spring and would love to contribute to this issue. Are there any pre-tests or prerequisite contributions required to qualify for LFX for this project?
[...]

PR #4167, opened 2025-03-04
inspect: adds ebpf and wasm extra info

KapilSareen:
This PR enhances the `inspect` subcommand by adding extra information and ensuring compatibility with both `ig-only` and client-server (`gadgetctl`) setups.  

### **Changes**  
- Extracts additional information from `wasm` and `ebpf` operators and stores it in `gadgetctx` for later use.  
- In a client-server setup, this information is transmitted via gRPC to the front-end `gadgetctx`.  
- In an `ig-only` setup, the extracted data is displayed directly.  
- Adds support for four output modes: `json`, `jsonpretty`, `yaml`, and `custom`.  
    - **Custom mode example:**  
    ```bash
    $ sudo ig image inspect myimage -o custom --extra-info=ebpf.maps
    ```  

Part of #3387.


cmd/common/image/inspect.go
```diff
@@ -31,6 +31,10 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
```
blanquicet, line 31: Please run `make lint` to fix formatting.

pkg/gadget-service/api/api.proto
```diff
@@ -249,6 +249,8 @@ message GetGadgetInfoRequest {
   uint32 version = 3;
 
   uint32 flags = 4;
+
+  bool getExtraInfo = 5;
```
blanquicet, line 253: I think we can drop the `get` and `extraInfo` is fine. Also please add a comment describing it, something like:

cmd/common/image/inspect.go
```diff
@@ -49,6 +53,21 @@ func NewInspectCmd() *cobra.Command {
...
+			// empty for now, we'll progressively add operators
+		
+			gadgetCtx := gadgetcontext.New(
+				context.Background(),
+				image.String(),
+				gadgetcontext.WithDataOperators(ops...),
+				gadgetcontext.WithUseInstance(false),
+				)
+		
+			gadgetCtx.SetExtraInfo(true)
```
blanquicet, line 66: I suggest making this an option so that we can do:

```go
			gadgetCtx := gadgetcontext.New(
				context.Background(),
				image.String(),
				gadgetcontext.WithGadgetExtraInfo(true),
		)
```

And then add only one new method to gadget-context:

```
func (c *GadgetContext) IncludeExtraInfo() bool
```

Similar to `useInstance` with `WithUseInstance` and `UseInstance`.

blanquicet (commented):
Few comments, it looks good already.

pkg/gadget-service/api/api.proto
```diff
@@ -158,6 +158,24 @@ message GadgetInfo {
   bytes metadata = 6;
   repeated Param params = 7;
   string id = 8;
+  ExtraEbpfInfo extraEbpfInfo = 9;
+}
+
+message ExtraEbpfInfo{
+  repeated string sections = 1;
+  repeated Map maps = 2;
+  repeated Program programs = 3;
```
alban, line 167: I am pondering about adding a generic map:
```proto
message ExtraEbpfInfo {
  map<string, GadgetInspectAddendum> addenda = 1;
}

message GadgetInspectAddendum {
  // mediaType could be application/json, text/plain, text/mermaid, text/markdown...
  string mediaType = 1;
  bytes content = 2;
}
```
So that operators would not have to modify the grpc API when adding new content (graph, text, ...).

It could contain things like:
* ebpf.maps // list of maps
* ebpf.programs // list of programs
* ebpf.mapsGraph // mermaid graph showing relations between maps and programs
* wasm.foo

```json
{
  "ebpf.maps": {
    "mediaType": "application/json",
    "content": "[{name: \"events\", type: \"BPF_MAP_TYPE_PERF_EVENT_ARRAY\"}, {name: \"execs\", type: \"BPF_MAP_TYPE_HASH\"}]",
  }
}
```

Then, [Inspektor Gadget Desktop](https://github.com/inspektor-gadget/ig-desktop) will be able to display additional information without recompiling. It would just recognise the mediaType and render it appropriately.

pkg/operators/ebpf/ebpf.go
```diff
@@ -323,6 +333,45 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
 }
 
+func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
+	ef, err := elf.NewFile(bytes.NewReader(i.program))
+	if err != nil {
+		return fmt.Errorf("parsing elf file: %w", err)
+	}
+	var sections []string
+	var maps []*api.Map
+	var programs []*api.Program
```
alban, line 343: In addition to programs and maps, we could add variables as understood by cilium/ebpf:
https://github.com/cilium/ebpf/blob/v0.17.3/collection.go#L46

alban (commented):
It is going in a good direction. Thanks!

cc @flyth

cmd/common/image/inspect.go
```diff
@@ -49,6 +54,28 @@ func NewInspectCmd() *cobra.Command {
 				return fmt.Errorf("inspecting image: %w", err)
 			}
 
+			// add the required operators (TODO)
+			gadgetCtx := gadgetcontext.New(
+				context.Background(),
+				image.String(),
+				// gadgetcontext.WithDataOperators(ops...),
```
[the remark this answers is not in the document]
KapilSareen: We don't for the gadgetctl-ig setup, but the ig only setup needs it, otherwise the ebpf operator is not initialised. Let me know what you think.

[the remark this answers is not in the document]
alban: It makes sense. Your change is fine then.

pkg/gadget-service/api/api.proto
flyth: Yes, agree - please keep it as generic as possible.

Some name change proposals:
`extraEbpfInfo` => `extraInfo`
`addendum/addenda` => `attachments/attachment` (or just `data`)
`mediaType` => `contentType`

pkg/gadget-service/api/api.proto
```diff
@@ -249,6 +267,9 @@ message GetGadgetInfoRequest {
   uint32 version = 3;
 
   uint32 flags = 4;
+
+  // can be used to include more info for debugging (e.g., in ig image inspect)
+  bool extraInfo = 5;
```
flyth, line 272:
```suggestion
  bool requestExtraInfo = 5;
```

This could also be made a flag (using one bit) of the currently unused `flags` field above.

KapilSareen: Renamed the bool. Not sure what you meant by this
> This could also be made a flag

KapilSareen:
@flyth @alban I've addressed the requested changes from my side. Please review them again, and if everything looks good, I'll proceed with adding the WASM data in the same manner.

cmd/common/image/inspect.go
```diff
@@ -49,6 +59,62 @@ func NewInspectCmd() *cobra.Command {
...
+
+			info, err = runtime.GetGadgetInfo(gadgetCtx, runtimeParams, paramValueMap)
+			if err != nil {
+				return fmt.Errorf("getting gadget info: %w", err)
+			}
+			fmt.Println(info.ExtraInfo.Data["ebpf.sections"])
+			fmt.Println(info.ExtraInfo.Data["ebpf.maps"])
+			fmt.Println(info.ExtraInfo.Data["ebpf.programs"])
+			fmt.Println(info.ExtraInfo.Data["wasm.upcalls"])
+			fmt.Println(info.ExtraInfo.Data["wasm.gadgetAPIVersion"])
```
alban, line 116: Can you iterate over `Data` in a generic way and insert it in the json/jsonpretty/yaml?

For the column output, I am thinking we would not print the information since it would not fit.

We could also have a custom output for showing one section. For example:
```
$ sudo ig image inspect myimage -o custom --extrainfo=ebpf.programs
```

pkg/operators/ebpf/ebpf.go
```diff
@@ -323,6 +352,75 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
+	for _, p := range i.collectionSpec.Programs {
+		programs = append(programs, &Program{
+			Section:  p.SectionName,
+			Bytecode: []byte(fmt.Sprintf("%v", p.Instructions.String())),
+		})
+	}
+	programsJson, _ := json.Marshal(programs)
+
+	// Add variables
+	for name := range i.collectionSpec.Variables {
```
alban, line 392: Unfortunately, the `name` field in cilium/ebpf.VariableSpec is not exported. See [discussion about exporting those fields](https://github.com/cilium/ebpf/pull/1564#issuecomment-2426347390).

https://github.com/cilium/ebpf/blob/v0.17.3/variable.go#L17

Instead, you could use the String() method:
```go
for _, v := range i.collectionSpec.Variables {
		variables = append(variables, &VariableSpec{
			name: v.String(),
		})
```

That will give you everything in a string in [this format](https://github.com/cilium/ebpf/blob/v0.17.3/variable.go#L104). It is not ideal, but that's what we can do now.

pkg/operators/ebpf/ebpf.go
```diff
@@ -323,6 +352,75 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
+			Type: m.Type.String(),
+		})
+	}
+	mapsJson, _ := json.Marshal(maps)
+
+	// Add programs
+	for _, p := range i.collectionSpec.Programs {
+		programs = append(programs, &Program{
+			Section:  p.SectionName,
+			Bytecode: []byte(fmt.Sprintf("%v", p.Instructions.String())),
```
alban, line 386: I tested with:
```bash
$ sudo ./ig image inspect fsnotify:latest|less
```
And I see base64 code:
```json
{
  "ebpf.maps": {
    "content": "[{\"Name\":\"events\",\"Type\":\"RingBuf\"},{\"Name\":\"gadget_heap\",\"Type\":\"PerCPUArray\"},{\"Name\":\"bufs\",\"Type\":\"PerCPUArray\"},{\"Name\":\"gadget_mntns_filter_map\",\"Type\":\"Hash\"},{\"Name\":\"enriched_fsnotify_events\",\"Type\":\"Hash\"},{\"Name\":\"fsnotify_insert_event_ctx\",\"Type\":\"Hash\"},{\"Name\":\"fsnotify_remove_first_event_ctx\",\"Type\":\"Hash\"}]",
    "contentType": "application/json"
  },
  "ebpf.programs": {
    "content": "[{\"Section\":\"kretprobe/fsnotify_remove_first_event\",\"Bytecode\":\"aWdfZmFfcGlja194OgoJICAgOyBpbnQgQlBGX0tSRVRQUk9CRShpZ19mYV9waWNrX3gsIHN"
  }
}
```

Any idea why?

Does it work better with this?
```diff
-Bytecode: []byte(fmt.Sprintf("%v", p.Instructions.String())),
+Bytecode: []byte(fmt.Sprintf("%v", p.Instructions)),
```
Instead of `Bytecode`, I would name the field `Source` since it should contain human readable code.

Ideally, the content field should contain something like this:
```
	    ; int ig_trace_dns(struct __sk_buff *skb)
	   0: MovReg dst: r6 src: r1
	    ; h_proto = load_half(skb, offsetof(struct ethhdr, h_proto));
	   1: LdAbsH imm: 12
	    ; switch (h_proto) {
	   2: StXMemDW dst: rfp src: r0 off: -152 imm: 0
	   3: JEqImm dst: r0 off: 34 imm: 34525
	   4: JNEImm dst: r0 off: 311 imm: 2048
	    ; __u8 ihl_byte = load_byte(skb, ETH_HLEN);
	   5: LdAbsB imm: 14
```

cmd/common/image/inspect.go
```diff
@@ -18,85 +18,154 @@ import (
...
+			gadgetcontext.WithUseInstance(false),
+			gadgetcontext.IncludeExtraInfo(true),
+		)
 
-				formatter := textcolumns.NewFormatter(cols.GetColumnMap(), textcolumns.WithShouldTruncate(isTerm))
-				formatter.WriteTable(cmd.OutOrStdout(), []*oci.GadgetImageDesc{image})
-			default:
-				return fmt.Errorf("invalid output mode %q, valid values are: %s", outputMode, strings.Join(outputModes, ", "))
+		paramValueMap := make(map[string]string)
+		ociParams.CopyToMap(paramValueMap, "operator.oci.")
```
alban, line 106: Could you use the parameter pull=never?
https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.38.0/pkg/operators/oci-handler/oci.go#L138-L149


The current documentation of the inspect command says:
```
inspect     Inspect the local gadget image
```
But for me, the inspect command downloads things from ghcr.io. I think it should not download anything.

alban (commented):
Thanks! Some comments.

The documentation in `docs/reference/images.md` should be updated:
https://inspektor-gadget.io/docs/latest/reference/images#inspect

The documentation in `docs/api/grpc.md` should be updated too:
https://inspektor-gadget.io/docs/latest/api/grpc
But since it is currently empty, I think it should be done in a separate PR.

I think you can remove the "WIP" from the PR description.

pkg/gadget-context/gadget-context.go
```diff
@@ -306,10 +313,28 @@ func (c *GadgetContext) SerializeGadgetInfo() (*api.GadgetInfo, error) {
...
+		ei, _ := c.GetVar("extraInfo.ebpf")
+		for k, v := range ei.(*api.ExtraInfo).Data {
+			gi.ExtraInfo.Data[k] = v
+		}
+
+		// serialize wasm extraInfo
+		wi, _ := c.GetVar("extraInfo.wasm")
+		for k, v := range wi.(*api.ExtraInfo).Data {
+			gi.ExtraInfo.Data[k] = v
+		}
```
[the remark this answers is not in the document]
KapilSareen: here, v is an `*api.ExtraInfo` itself (it has a Data field that needs to be iterated upon too before assigning to gi) so this won't work, but I got the point. Will make it generic, Thanks!

pkg/operators/ebpf/ebpf.go
KapilSareen: weirdly enough, it was due to typecasting to `[]byte`. Fixed the type to String now, works fine.

pkg/operators/ebpf/ebpf.go
KapilSareen: Fixed this and added the other variable related info too, let me know if anything is irrelevant.

cmd/common/image/inspect.go
```diff
@@ -18,85 +18,144 @@ import (
...
+	cmd.RunE = func(cmd *cobra.Command, args []string) error {
+		image, err := oci.GetGadgetImageDesc(context.TODO(), args[0])
+		if err != nil {
+			return fmt.Errorf("inspecting image: %w", err)
+		}
+
+		runtimeGlobalParams := runtime.GlobalParamDescs().ToParams()
+		runtimeParams := runtime.ParamDescs().ToParams()
+		ociParams := apihelpers.ToParamDescs(ocihandler.OciHandler.InstanceParams()).ToParams()
+		ociParams.Set("pull", oci.PullImageNever)
```
KapilSareen, line 63: @alban do you believe this is correct?

alban: I tested with:
```diff
--- a/pkg/oci/oci.go
+++ b/pkg/oci/oci.go
@@ -944,6 +944,7 @@ func getContentBytesFromDescriptor(ctx context.Context, fetcher content.Fetcher,
 }
 
 func ensureImage(ctx context.Context, imageStore oras.Target, image string, imgOpts *ImageOptions, pullPolicy string) error {
+       fmt.Printf("ensureImage(ctx, imageStore, %q, imgOpts, %q\n", image, pullPolicy)
        switch pullPolicy {
        case PullImageAlways:
                _, err := pullGadgetImageToStore(ctx, imageStore, image, &imgOpts.AuthOptions)
```

And running:
```bash
sudo ./ig image inspect fsnotify:latest 2>&1|grep ensureImage
```
I get the result "never", so it seems to work.

However, it still generate network traffic to ghcr.io (verified by running the dns gadget in another ig instance), probably to verify the image signature because the param "verify-image" is set to true by default... I think we should set "verify-image=false" because we are not running any gadget and we are not downloading anything. I don't think it would make sense to expose the `--verify-image` flag on the CLI for the inspect command. But I would like a second opinion, cc @flyth

pkg/operators/ebpf/ebpf.go
KapilSareen: I couldn't find a clean, generic way to extract the nested type from v.Type(), other than manually converting it to a map and checking for errors to verify if the field exists. Moreover, v.Type().TypeName() only returns the variable's name, which isn’t helpful in this case. I'm not a fan of this approach. What are your thoughts on dropping this information altogether, or is it essential for your use case?

cmd/common/image/inspect.go
KapilSareen: Works now with param "verify-image" hardcoded to false

docs/reference/images.md
```diff
@@ -284,24 +284,35 @@ $ sudo ig image inspect -h
...
+    "contentType": "application/json"
+  },
+  ...
 }
+
+# Retrieve specific info using custom output
+$ sudo ig image inspect -o custom --extra-info=ebpf.sections ghcr.io/inspektor-gadget/gadget/trace_tcp
+["",".strtab",".text","kprobe/tcp_v4_connect",".relkprobe/tcp_v4_connect","kretprobe/tcp_v4_connect",".relkretprobe/tcp_v4_connect","kprobe/tcp_v6_connect",".relkprobe/tcp_v6_connect","kretprobe/tcp_v6_connect",".relkretprobe/tcp_v6_connect","kprobe/tcp_close",".relkprobe/tcp_close","kprobe/tcp_set_state",".relkprobe/tcp_set_state","kretprobe/inet_csk_accept",".relkretprobe/inet_csk_accept",".rodata",".bss","license",".maps",".BTF",".rel.BTF",".BTF.ext",".rel.BTF.ext",".llvm_addrsig",".symtab"]
```
alban, line 317: I suggest to add more examples with jq.

Listing programs.

```
$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs|jq -r '.[].Section'
kprobe/fanotify_handle_event
kprobe/fsnotify_remove_first_event
kretprobe/fsnotify_remove_first_event
kprobe/inotify_handle_inode_event
kretprobe/fanotify_handle_event
kprobe/fsnotify_insert_event
kprobe/fsnotify_destroy_event
kretprobe/inotify_handle_inode_event
```

Printing the source of a program.

```
$ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs|jq -r '.[] | select(.Section=="kprobe/fsnotify_remove_first_event") | .Source'
ig_fa_pick_e:
	  ; int BPF_KPROBE(ig_fa_pick_e, struct fsnotify_group *group)
	 0: LdXMemDW dst: r6 src: r1 off: 112 imm: 0
	 1: StXMemDW dst: rfp src: r6 off: -8 imm: 0
	  ; pid_tgid = bpf_get_current_pid_tgid();
	 2: Call FnGetCurrentPidTgid
...
	25: MovImm dst: r0 imm: 0
	26: Exit
```

cmd/common/image/inspect.go
alban: It works but it still print the warning message:
```
```

In my opinion, this message should not be printed in this context. But this can be changed later in another PR once we agree on the correct behaviour.

cmd/common/image/inspect.go
```diff
@@ -18,85 +18,145 @@ import (
...
+			bytes, err := yaml.Marshal(extraInfoMap)
+			if err != nil {
+				return fmt.Errorf("marshalling image and extra info to YAML: %w", err)
+			}
+			fmt.Fprint(cmd.OutOrStdout(), string(bytes))
+		case utils.OutputModeCustom:
+			extraInfo, _ := cmd.PersistentFlags().GetString("extra-info")
+			customInfo := ""
+			for k, v := range info.ExtraInfo.Data {
+				if extraInfo == k {
```
alban, line 143: I think there should be an error message when:
* extraInfo==""
* extraInfo is not found in `info.ExtraInfo.Data`

docs/reference/images.md
```diff
@@ -284,24 +284,35 @@ $ sudo ig image inspect -h
...
 Usage:
-  ig image inspect [flags]
+  ig image inspect IMAGE [flags]
 
 Flags:
   -h, --help            help for inspect
-  -o, --output string   Output mode, possible values are, columns, json, jsonpretty (default "columns")
+  -o, --output string   Output mode: json, jsonpretty, yaml, or custom (default "jsonpretty")
+  --extra-info string   In custom mode, specify particular info required
+
```
alban, line 293: extra empty line

pkg/operators/wasm/wasm.go
```diff
@@ -274,11 +276,47 @@ func (i *wasmOperatorInstance) init(
...
+func (i *wasmOperatorInstance) addExtraInfo(gadgetcontext operators.GadgetContext, version uint64, wasmProgram []byte) error {
+	module, err := i.rt.CompileModule(gadgetcontext.Context(), wasmProgram)
+	if err != nil {
+		return err
+	}
+	imports := module.ImportedFunctions()
+	upcalls := []string{}
+	for _, imp := range imports {
+		parts := strings.Split(imp.Name(), ".")
+		upcalls = append(upcalls, parts[len(parts)-1])
```
alban, line 301: This gets both upcalls ("setConfig","kallsymsSymbolExists","gadgetLog") and other functions that are not provided by IG such as `runtime.fd_write` or `__imported_wasi_snaps`.

You can filter those functions with:

```go
moduleName, name, isImport := imp.Import()
```
If `isImport` is true and `moduleName` is "ig", then it should be an upcall. But it needs to be rebased on #4209.

docs/reference/images.md
```diff
@@ -284,24 +284,35 @@ $ sudo ig image inspect -h
...
+    "contentType": "application/json"
+  },
+  "ebpf.sections": {
+    "content": "[\"\",\".strtab\",\".text\",\"kprobe/tcp_v4_connect\",\".relkprobe/tcp_v4_connect\",\"kretprobe/tcp_v4_connect\",\".relkretprobe/tcp_v4_connect\",\"kprobe/tcp_v6_connect\",\".relkprobe/tcp_v6_connect\",\"kretprobe/tcp_v6_connect\",\".relkretprobe/tcp_v6_connect\",\"kprobe/tcp_close\",\".relkprobe/tcp_close\",\"kprobe/tcp_set_state\",\".relkprobe/tcp_set_state\",\"kretprobe/inet_csk_accept\",\".relkretprobe/inet_csk_accept\",\".rodata\",\".bss\",\"license\",\".maps\",\".BTF\",\".rel.BTF\",\".BTF.ext\",\".rel.BTF.ext\",\".llvm_addrsig\",\".symtab\"]",
+    "contentType": "application/json"
+  },
+  ...
 }
+
+# Retrieve specific info using custom output
```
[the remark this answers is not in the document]
alban: Before this PR, the following information were displayed by the inspect command:
```
$ sudo ./ig image inspect fsnotify:main -o jsonpretty ; echo
{
  "Repository": "ghcr.io/inspektor-gadget/gadget/fsnotify",
  "Tag": "main",
  "Digest": "sha256:120614ca3b37da10504efd8697ab720e0da05f1b1a316a1a25343fcc662384e7",
  "Created": "2025-03-25T08:56:13Z"
}
```
The new inspect command does not give this information anymore.
I think it could be added back in the "oci" operator, in `OciHandlerInstance.init()`.
What do you think?

The information is still available in the `ig image list -o json` command, so I think it could ok to do it in a follow-up PR.

cmd/common/image/inspect.go
```diff
@@ -18,85 +18,148 @@ import (
...
+			if extraInfo == "" || info.ExtraInfo.Data[extraInfo] == nil {
+				return fmt.Errorf("extra info %q not found", extraInfo)
+			}
+			customInfo := ""
+			for k, v := range info.ExtraInfo.Data {
+				if extraInfo == k {
+					customInfo = string(v.Content)
+					break
+				}
 			}
```
alban, line 150: Could it just be `customInfo = info.ExtraInfo.Data[extraInfo].Content`?

pkg/gadget-context/gadget-context.go
```diff
@@ -371,6 +392,19 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
 		}()
 	}
 
+	if c.ExtraInfo() && extraInfo != nil {
+		for k, v := range extraInfo.Data {
+			prefix := strings.Split(k, ".")[0]
+			ei, ok := c.GetVar("extraInfo." + prefix)
```
alban, line 398: Why do we only take the prefix? Could you add a comment saying what is the syntax of `k`?

pkg/gadget-service/api/api.proto
```diff
@@ -158,6 +158,17 @@ message GadgetInfo {
   bytes metadata = 6;
   repeated Param params = 7;
   string id = 8;
+  ExtraInfo extraInfo = 9;
+}
+
+message ExtraInfo {
+  map<string, GadgetInspectAddendum> data = 1;
```
alban, line 165: Could you add a comment to say what could be the key of the map?

pkg/operators/ebpf/ebpf.go
```diff
@@ -191,6 +193,23 @@ type ebpfInstance struct {
...
+	Section string
+	Source  string
+}
+
+type VariableSpec struct {
+	Name   string
+	Offset uint64
+	Size   uint64
+	Map    string
+}
```
alban, line 211: Those 3 types don't need to be exported, they could be named `extraInfoMap`, `extraInfoProgram` and `extraInfoVariable`.

pkg/operators/ebpf/ebpf.go
```diff
@@ -325,6 +353,78 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
 	return nil
 }
 
+func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
```
alban, line 356: I would move this function and the 3 types to a new file `pkg/operators/ebpf/extrainfo.go` since `ebpf.go` is already big. That will also be better for future PRs adding more extra-infos.

pkg/operators/wasm/wasm.go
```diff
@@ -276,11 +277,49 @@ func (i *wasmOperatorInstance) init(
...
+			return fmt.Errorf("adding extra info: %w", err)
+		}
+	}
+
 	i.dataSourceCallback = mod.ExportedFunction("dataSourceCallback")
 
 	return err
 }
 
+func (i *wasmOperatorInstance) addExtraInfo(gadgetcontext operators.GadgetContext, version uint64, wasmProgram []byte) error {
```
alban, line 293: This could also be in a new file `pkg/operators/wasm/extrainfo.go`

alban (commented):
Thanks! Some small comments to fix, and it should be good to merge afterwards.

TODO for future PRs:
- [ ] Add "image inspect" command in kubectl-gadget
- [ ] Tests
- [ ] More extra-info from operators
  - [ ] Image info from oci operator (Repository, Tag, Digest, Created, gadget.yaml metadata, annotations from OCI image)
  - [ ] Relation between ebpf programs and maps (flowchart Mermaid graph)
  - [ ] Sequence Mermaid graph
  - [ ] Relocation added in source code

docs/reference/images.md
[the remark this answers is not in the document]
KapilSareen: Sure let's just add it back in the next one

cmd/common/image/inspect.go
```diff
@@ -67,7 +67,9 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 		for _, op := range operators.GetDataOperators() {
 			opGlobalParams[op.Name()] = apihelpers.ToParamDescs(op.GlobalParams()).ToParams()
 		}
-		opGlobalParams["oci"].Set("verify-image", "false")
+		if ociParams, exists := opGlobalParams["oci"]; exists {
+			ociParams.Set("verify-image", "false")
```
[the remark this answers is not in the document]
KapilSareen: yes, gadgetctl was crashing before this, it fixes that error.

alban (approved):
LGTM, thanks a lot!

I'll merge it later today, unless @flyth has feedback on https://github.com/inspektor-gadget/inspektor-gadget/pull/4167#discussion_r2015865013. But IMO this can be fixed in a follow-up PR.

cmd/common/image/inspect.go
flyth: > I think we should set "verify-image=false" because we are not running any gadget and we are not downloading anything.

I actually think we should use that parameter for the inspect command as well. While we're not running any eBPF code, we're still executing some WASM (if present) with `GetGadgetInfo()`. If we have the verification in place, I'd also be okay with inspecting gadgets that need to be downloaded first (so pull policy doesn't need to be hard-coded as well).

cmd/common/image/inspect.go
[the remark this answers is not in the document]
flyth: No, I don't think we should add more dependencies on the oci operator to our clients - currently, it's already the case in other places because of the params, but I think it'd be better to remove it sooner rather than later.

Also, `verify-image` is a global param (not per-instance), so it shouldn't matter on the clients at all (i.e. `gadgetctl image inspect` will only work if `ig daemon` is run with `--verify-image=false` for unsigned/foreign images, right?).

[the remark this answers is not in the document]
KapilSareen: Thank @flyth, I was already discussing this with @alban and I've already removed this in https://github.com/inspektor-gadget/inspektor-gadget/pull/4259

Issue #1861, opened 2023-07-18
[RFE] Use KubeIPResolver in ig


burak-ok:
## `KubeIPResolver` can only used in `kubectl-gadget`

`KubeIPResolver` is only used in `kubectl-gadget`

## Impact

Enriching the ip addresses for gadgets is not working in `ig`. With #1702 we now have K8s information for events in `ig` so we could enrich the columns for this case

## Ideal future situation

The `KubeIPResolver` and `KubeNameResolver` should work in `ig` too. If we can't get any K8s information the operator shouldn't fail.

burak-ok:
In order for the `KubeIPResolver` to work we need to have a `kubeconfig`. In my test minikube cluster I don't have one on the node itself. So I have to copy it into the node first before I can enrich the network traffic with `KubeIPResolver`

Is this the target use-case for this issue?

cc @blanquicet

mqasimsarfraz:
> In my test minikube cluster I don't have one on the node itself

Perhaps we can use `/etc/kubernetes/kubelet.conf`, `kubelet` uses the same to talk to API server.

alban:
I think it is a valid use case.
This is similar to #3319: we need ig to be able to talk to the api server.

In some cases, when `ig` is deployed in a pod, using `rest.InClusterConfig()` is enough. For example, in `kubectl debug` or in custom deployments.

PR #4169, opened 2025-03-04
KubeNameResolver: Add support for image based gadgets

burak-ok:
This will be needed for `advise networkpolicy`

This Operator will now be always active for K8s deployments.
Is this okay? Should it be configurable instead? Or are we still waiting for the config to selectivly disable operators in general

### DIFF
--- a/docs/spec/operators/kubenameresolver.md
+++ b/docs/spec/operators/kubenameresolver.md
@@ -1,0 +1,74 @@
+---
+title: KubeNameResolver
+---
+
+The KubeNameResolver operator uses the `K8s.PodName` and `K8s.Namespace` fields to enrich the event with `K8s.PodIp` and `K8s.HostIp` fields. 
+This operator is disabled by default (See [annotation](#annotation) for how to enable it).
+
+The example below shows a request from `test-pod` pod in json format:
+
+Without `KubeNameResolver`:
+```json
+{
+  ...
+  "k8s": {
+    "containerName": "test-pod",
+    "hostnetwork": false,
+    "namespace": "default",
+    "node": "minikube-docker",
+    "owner": {
+      "kind": "",
+      "name": ""
+    },
+    "podLabels": "run=test-pod",
+    "podName": "test-pod"
+  },
+  "proc": {
+    "comm": "wget",
+    ...
+  },
+  ...
+}
+```
+
+With `KubeNameResolver`:
+```json
+{
+  ...
+  "k8s": {
+    "containerName": "test-pod",
+    "hostIP": "192.168.58.2",
+    "hostnetwork": false,
+    "namespace": "default",
+    "node": "minikube-docker",
+    "owner": {
+      "kind": "",
+      "name": ""
+    },
+    "podIP": "10.244.0.29",
+    "podLabels": "run=test-pod",
+    "podName": "test-pod"
+  },
+  "proc": {
+    "comm": "wget",
+    ...
+  },
+  ...
+}
+```
+
+## Priority
+
+11
+
+## Parameters
+
+None
+
+## Annotation
+
+This operator is disabled by default and is only activated for a datasource if it has the following annotations:
+datasource_name:
+  annotations:
+    kubenameresolver.enable: true    
+```
--- a/pkg/operators/kubenameresolver/kubenameresolver.go
+++ b/pkg/operators/kubenameresolver/kubenameresolver.go
@@ -1,4 +1,4 @@
-// Copyright 2023 The Inspektor Gadget authors
+// Copyright 2023-2025 The Inspektor Gadget authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/common"
@@ -31,6 +33,7 @@ import (
 
 const (
 	OperatorName = "KubeNameResolver"
+	priority     = 11
 )
 
 type KubeNameResolverInterface interface {
@@ -92,6 +95,7 @@ func (k *KubeNameResolver) Instantiate(gadgetCtx operators.GadgetContext, gadget
 type KubeNameResolverInstance struct {
 	gadgetCtx      operators.GadgetContext
 	k8sInventory   common.K8sInventoryCache
+	accessors      map[datasource.DataSource]k8sAccesors
 	gadgetInstance any
 }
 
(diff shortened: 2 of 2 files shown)


pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,154 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+			logger.Debugf("> no k8s fields found")
+			continue
+		}
+
+		k8sAccesors := k8sAccesors{}
+		var err error
+
+		k8sAccesors.PodName = ds.GetField("k8s.podName")
+		if k8sAccesors.PodName == nil {
+			continue
```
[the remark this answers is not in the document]
burak-ok: Makes sense, I added warnings for every field under `K8s`

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,152 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+		accessors:    accessors,
+	}, nil
+}
+
+func (m *KubeNameResolverInstance) enrichSingle(data datasource.Data, accessor k8sAccesors) {
+	podName, _ := accessor.PodName.String(data)
+	namespace, _ := accessor.Namespace.String(data)
+
+	pod := m.k8sInventory.GetPodByName(namespace, podName)
+	if pod != nil {
```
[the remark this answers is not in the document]
burak-ok: I added the errors for `.String()`

The `pod != nil` check still needs to be there, since there is a (highly unlikly scenario) that we don't have informations about the pod. In that case we don't want an error

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,164 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+			if numDashes >= 2 {
+				currentDashIndex := -1 // The search starts at index + 1
+				// Skip the last 2 text segments
+				for range numDashes - 1 {
+					// Current index + 1 (the dash) + the next "local" dash index
+					currentDashIndex += 1 + strings.Index(pod.Name[currentDashIndex+1:], "-")
+				}
+				owner := pod.Name[:currentDashIndex]
+				accessor.Owner.PutString(data, owner)
+			}
```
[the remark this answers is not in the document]
burak-ok: I don't want to modify the previous behavior. So if we get `my-nginx-something-3352425-v413` we want `my-nginx-something`.

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,176 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+			numDashes++
+			if numDashes == 2 {
+				break
+			}
+		}
+
+		if numDashes == 2 {
+			accessor.Owner.PutString(data, pod.Name[:lastIndex])
+		}
+	}
```
mauriciovasquezbernal, line 271: Shouldn't this be part of the gadget itself and not of this operator?

mauriciovasquezbernal (commented):
I don't remember very well why we have an independent operator for this functionality. Would it be possible to integrate this into the kubemanager? It's the one enriching events with podname, etc. so probably it could also add the PodIP and HostIP if needed. Or is there any reason to keep it separated?

burak-ok:
> I don't remember very well why we have an independent operator for this functionality. Would it be possible to integrate this into the kubemanager? It's the one enriching events with podname, etc. so probably it could also add the PodIP and HostIP if needed. Or is there any reason to keep it separated?

I can't remember it as well. I am just extending this functionality because it is needed to port the `advise networkpolicy` to image based gadget.

One reason (in the future) might be to disable this behaviour of `KubeNameResolver`. Because it keeps always keeps (another) list of all K8s Pods and Services cached, one might want to reduce the memory consumption.
If its a separate operator we could disable it quite easily

burak-ok:
> If we integrate this into KubeManager, wouldn't it be more efficient as only one list of pods/services will be kept?

AFAIK the `KubeManager` doesn't keep a list of all K8s Pods and Svcs at all

mauriciovasquezbernal:
It uses the podinformer and somehow it keeps track of a list of pods (however only for the single node). If it's easier I'm for going with a new operator.

burak-ok:
> If it's easier I'm for going with a new operator.

I started porting it to `KubeManager` but then an old issue came into my mind: https://github.com/inspektor-gadget/inspektor-gadget/issues/1861 
We might want to do this kind of enrichment also for `ig` - and with a separate Operator it is possible. Merging it into `KubeManager` might make it more difficult to use.

mauriciovasquezbernal:
> This Operator will now be always active for K8s deployments.
Is this okay? Should it be configurable instead? Or are we still waiting for the config to selectivly disable operators in general

I'm a bit worried about the performance (specially because we have received feedback on that matter). IIUC, it means it'll be enabled for all gadgets that support kubernetes enrichment (trace_open, trace_exec, etc., right? If that's the case, I'll propose to use a special annotation and only enable this operator for specific gadgets for now. I acknowledge it breaks the gadget and operator independence, but I don't want to affect the performance of many other gadgets that don't need this operator.

pkg/operators/kubenameresolver/kubenameresolver.go
burak-ok: Makes sense, this is something too specific (Also the comment :D)

burak-ok:
> > This Operator will now be always active for K8s deployments.
> > Is this okay? Should it be configurable instead? Or are we still waiting for the config to selectivly disable operators in general
> 
> I'm a bit worried about the performance (specially because we have received feedback on that matter). IIUC, it means it'll be enabled for all gadgets that support kubernetes enrichment (trace_open, trace_exec, etc., right? If that's the case, I'll propose to use a special annotation and only enable this operator for specific gadgets for now. I acknowledge it breaks the gadget and operator independence, but I don't want to affect the performance of many other gadgets that don't need this operator.

I mean we could have a table of operators that are enabled by default and disabled by default maybe?

> Also, any suggestion how to test this change?

If you mean some kind of unit or intergration tests: I didn't plan any for this PR. Its too reliant on the K8s API for a unit tests and imo too small for an integration test.
This operator would be tested when `advise_networkpolicy` is implemented and has its own integration test.

burak-ok:
I also added the missing documentation for this operator

docs/spec/operators/kubenameresolver.md
```diff
@@ -0,0 +1,77 @@
+---
+title: KubeNameResolver
+---
+
+The KubeNameResolver enriches the `K8s.PodIp`,`K8s.HostIp` and `K8s.PodLabels` fields by reading and resolving the `K8s.PodName` and `K8s.Namespace` fields.
```
mauriciovasquezbernal, line 5: I feel "enriching a field" is wrong, it should be "enriching an event"

What about?

```suggestion
The KubeNameResolver operator uses the `K8s.PodName` and `K8s.Namespace` fields to enrich the event with `K8s.PodIp`,`K8s.HostIp` and `K8s.PodLabels` fields. 
```

Or something along the line.

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,150 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+	PodLabels datasource.FieldAccessor
+}
+
+func (k *KubeNameResolver) InstantiateDataOperator(gadgetCtx operators.GadgetContext, instanceParamValues api.ParamValues) (operators.DataOperatorInstance, error) {
+	logger := gadgetCtx.Logger()
+	accessors := make(map[datasource.DataSource]k8sAccesors)
+	for _, ds := range gadgetCtx.GetDataSources() {
+		logger.Debugf("KubeNameResolverOperator inspecting datasource %q", ds.Name())
+
+		if enable, ok := ds.Annotations()["kubenameresolver.enable"]; !ok || enable != "true" {
```
mauriciovasquezbernal, line 167: super minor nit: you can compare directly with the string:

```suggestion
		if ds.Annotations()["kubenameresolver.enable"] != "true" {
```

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,150 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+		if k8sAccesors.Namespace == nil {
+			logger.Warnf("No namespace field found in datasource %q", ds.Name())
+			continue
+		}
+
+		k8sAccesors.PodLabels = ds.GetField("k8s.podLabels")
+		if k8sAccesors.PodLabels == nil {
+			logger.Warnf("No podLabels field found in datasource %q", ds.Name())
+			continue
+		}
```
mauriciovasquezbernal, line 197: I can't understand how PodLabels is used. I know PodName and Namespace are used to lookup the pod, but this field is not needed there. So what's the relationship of this operator and this field? I mean,  it's created and set on pkg/datasource/compat/wrapper.go, so why is this operator touching it?

pkg/operators/kubenameresolver/kubenameresolver.go
```diff
@@ -134,6 +138,150 @@ func (m *KubeNameResolverInstance) EnrichEvent(ev any) error {
...
+			logger.Debugf("> no k8s fields found")
+			continue
+		}
+
+		k8sAccesors := k8sAccesors{}
+		var err error
+
+		k8sAccesors.PodName = ds.GetField("k8s.podName")
+		if k8sAccesors.PodName == nil {
+			logger.Warnf("No podName field found in datasource %q", ds.Name())
```
mauriciovasquezbernal, line 183: Do we need to print all these messages as warnings? Is it really and issue for the datasource not to have this field? I mean, is this information important for the user?

mauriciovasquezbernal (commented):
> If you mean some kind of unit or intergration tests: I didn't plan any for this PR. Its too reliant on the K8s API for a unit tests and imo too small for an integration test.
This operator would be tested when advise_networkpolicy is implemented and has its own integration test.

I meant like a sample gadget or something similar to check that it's working. Otherwise I can only base the review on code inspection. (But it's fine now, I trust you tested it :D) 

Besides the comment on podLabels, it LGTM.

pkg/operators/kubenameresolver/kubenameresolver.go
burak-ok: I am not sure. This was suggested in https://github.com/inspektor-gadget/inspektor-gadget/pull/4169#discussion_r1981147925 but I see both arguments and I can't decide between the two

pkg/operators/kubenameresolver/kubenameresolver.go
burak-ok: I guess this operator preceeded the changes that led to `PodLabels` be enriched by default in `wrapper.go`.

You are right, this is not needed anymore. Removing it

burak-ok:
> I meant like a sample gadget or something similar to check that it's working. Otherwise I can only base the review on code inspection. (But it's fine now, I trust you tested it :D)

I tested it locally with 
```bash
$ go run ./cmd/kubectl-gadget run trace_tcp -o json --annotate tracetcp:kubenameresolver.enable=true
```

pkg/operators/kubenameresolver/kubenameresolver.go
mauriciovasquezbernal: I don't have a strong opinion as well, so let's leave them.

mauriciovasquezbernal (approved):
LGTM. Thanks. I didn't test it myself but I trust you! :)

PR #4175, opened 2025-03-05
gadget trace_open: add user stacks

alban:
Example of output for a simple Go program opening a file:

```
sudo ig run ghcr.io/inspektor-gadget/gadget/trace_open:alban_trace_open_ustack \
        --pull=never --verify-image=false \
        --host --comm=gocat --collect-ustack -o yaml
...
ustack: '[0]internal/runtime/syscall.Syscall6; [1]syscall.Syscall6; [2]syscall.openat;
  [3]os.open; [4]os.openFileNolog; [5]os.OpenFile; [6]main.main; [7]runtime.main;
  [8]runtime.goexit.abi0; '
```

This is the same implementation as:
- trace_malloc in commit d33e3325b2c30ca0884b1d30028213f56a379b32
- trace_capabilities in commit 0b76243b4c6ae4cd1270324fdf3eea944d8e1397

### DIFF
--- a/gadgets/trace_open/README.mdx
+++ b/gadgets/trace_open/README.mdx
@@ -29,6 +29,12 @@ $ sudo ig run ghcr.io/inspektor-gadget/gadget/trace_open:%IG_TAG% [flags]
 </Tabs>
 ## Flags
 
+### `--collect-ustack`
+
+Collect user stack traces
+
+Default value: "false"
+
 ### `--failed`
 
 Show only failed events
--- a/gadgets/trace_open/gadget.yaml
+++ b/gadgets/trace_open/gadget.yaml
@@ -32,8 +32,20 @@ datasources:
         annotations:
           columns.width: 32
           columns.minwidth: 24
+      ustack_raw:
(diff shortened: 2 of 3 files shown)


alban:
Feedback from OOB discussion:
`--print-ustack` should be called `--collect-ustack` since it does not print anything by default with the default column output.

alban:
Should `--print-kstack` be changed to `--collect-kstack` too? (e.g. in trace_capabilities)

alban:
Ok, I let's do it with https://github.com/inspektor-gadget/inspektor-gadget/pull/4248 first. Once merged, I will rework on this PR.

alban:
#4248 is merged and this PR is rebased. Ready for review.

mauriciovasquezbernal (approved):
LGTM. Are we planning to do the same on all gadgets?

alban:
> LGTM. Are we planning to do the same on all gadgets?

I'm not sure yet. For trace_open, I think there is a clear use case to know why the syscall was called.

PR #4176, opened 2025-03-05
ci: Compute checksums for SBOMs

eiffel-fl:


.github/workflows/inspektor-gadget.yml
```diff
@@ -2306,6 +2297,15 @@ jobs:
           rm -fr "$temp_dir"
         done
     - name: Install Cosign
+    - name: Compute checksums for all artifacts
```
burak-ok, line 2300: Move one line above the `- name: Install Cosign`

burak-ok (commented):
LGTM, thanks for increasing the security score in CLOMonitor

burak-ok (approved):
> Was this behavior already met on contributors' forks?

I guess it was always this way and nobody complained?

PR #4177, opened 2025-03-05
operators/wasm: Fix return code for fieldSet

mauriciovasquezbernal:
Stack[0] was never assigned to zero, it caused the different set functions to return an error when the operation was successful.

Fixes: 7a3e62698bbc ("operators: Add wasm operator")

Ref https://github.com/inspektor-gadget/inspektor-gadget/pull/4160#discussion_r1979586309

### DIFF
--- a/pkg/operators/wasm/testdata/badguest/program.go
+++ b/pkg/operators/wasm/testdata/badguest/program.go
@@ -294,7 +294,7 @@ func gadgetInit() int32 {
 	dataHandle := dataSourceNewPacketSingle(dsSingleHandle)
 	assertNonZero(dataHandle, "dataSourceNewPacketSingle: creating new")
 
-	assertNonZero(fieldSet(fieldHandle, dataHandle, uint32(api.Kind_Uint32), 1234), "fieldSet: ok")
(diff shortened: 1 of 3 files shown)


pkg/operators/wasm/testdata/badguest/program.go
```diff
@@ -294,7 +294,7 @@ func gadgetInit() int32 {
 	dataHandle := dataSourceNewPacketSingle(dsSingleHandle)
 	assertNonZero(dataHandle, "dataSourceNewPacketSingle: creating new")
 
-	assertNonZero(fieldSet(fieldHandle, dataHandle, uint32(api.Kind_Uint32), 1234), "fieldSet: ok")
+	assertZero(fieldSet(fieldHandle, dataHandle, uint32(api.Kind_Uint32), 1234), "fieldSet: ok")
```
burak-ok, line 297: I think this is the most fun part that we apparently expected an error :D

burak-ok (approved):
LGTM, thanks for finding this

pkg/operators/wasm/testdata/badguest/program.go
mauriciovasquezbernal: Probably generated by copilot and nobody caught it.

PR #4178, opened 2025-03-05
ustack: add --symbolizers parameter

alban:
For now, we have only one symbolizer: symtab. I am planning to add more symbolizers in future PRs for goresym and debuginfod.

The symtab symbolizer fills the new sub-field symbols:
```yaml
ustack:
  addresses: '[0]0x000000000040332e; [1]0x00000000004771e6; [2]0x000000000047752e;
    [3]0x000000000048250b; [4]0x00000000004829ae; [5]0x0000000000482a7e; [6]0x0000000000482afe;
    [7]0x0000000000482b7e; [8]0x0000000000435a9d; [9]0x00000000004625a1; '
  symbols: '[0]runtime/internal/syscall.Syscall6; [1]syscall.Syscall; [2]syscall.Syscall.abi0;
    [3]golang.org/x/sys/unix.Chroot; [4]main.level3; [5]main.level2; [6]main.level1;
    [7]main.main; [8]runtime.main; [9]runtime.goexit.abi0; '
  ...
```
Using --symbolizers=none means the ustack operator will not run the symbolizer, so the symbols sub-field will be empty:
```yaml
ustack:
  addresses: '[0]0x000000000040332e; [1]0x00000000004771e6; [2]0x000000000047752e;
    [3]0x000000000048250b; [4]0x00000000004829ae; [5]0x0000000000482a7e; [6]0x0000000000482afe;
    [7]0x0000000000482b7e; [8]0x0000000000435a9d; [9]0x00000000004625a1; '
  symbols: ""
  ...
```
Disabling all symbolizers can be useful because the addresses are still returned and users might want to do the symbolization themselves out of band for performance reasons.

Note that gadget still have a --print-ustack parameter. When not used, the addresses are not returned:
```yaml
ustack:
  addresses: ""
  symbols: ""
```
The column view can be enabled with:
```bash
$ sudo ./ig run ghcr.io/inspektor-gadget/gadget/trace_capabilities:latest \
		--collect-ustack --comm=gochr-dyn --host \
		--symbolizers=auto --fields=proc.comm,cap,ustack.addresses,ustack.symbols
COMM       CAP               USTACK.ADDRESSES                                                          USTACK.SYMBOLS
gochr-dyn  CAP_SYS_CHROOT    [0]0x000000000040332e; [1]0x00000000004771e6; [2]0x000000000047752e; [3]… [0]runtime/internal/syscall.Syscall6; [1]syscall.Syscall; [2]syscall.Sys…
```

### DIFF
--- a/docs/gadget-devel/gadget-ebpf-api.md
+++ b/docs/gadget-devel/gadget-ebpf-api.md
@@ -341,11 +341,7 @@ Symbolize the kernel stack from `gadget_get_kernel_stack(ctx)` (see [kernel-stac
 
 ### `gadget_user_stack`
 
-Symbolize the user stack from `gadget_get_user_stack(ctx, &event->ustack_raw, collect_ustack)` (see [user-stack-traces](#user-stack-traces)).
-
-#### Annotations
-
-- `ebpf.formatter.ustack`: Name of the new field. If the annotation is not set and the source field name has a `_raw` suffix, the target name will be set to the source name without that suffix.
+Symbolize the user stack from `gadget_get_user_stack(ctx, &event->ustack, collect_ustack)` (see [user-stack-traces](#user-stack-traces)).
 
 ### `gadget_uid` and `gadget_gid`
 
@@ -611,14 +607,14 @@ struct {
 Then, add a field in the event structure with the type of `gadget_user_stack`,
 designated for storing the stack id along with identifiers for the executable
 so that the stack can be symbolised in userspace.
-`gadget_get_user_stack(ctx, &event->ustack_raw, collect_ustack)` could be used
+`gadget_get_user_stack(ctx, &event->ustack, collect_ustack)` could be used
 to populate this field, this helper function will store the kernel stack into
-`ig_ustack` and fill the field passed as parameter. When collect_ustack is
-false, ustack_raw is initialized to zero and ig will ignore the stack trace.
+`ig_ustack` and fill the field passed as parameter. When `collect_ustack` is
+false, `ustack` is initialized to zero and ig will ignore the stack trace.
 
 ```C
 struct event {
-	struct gadget_user_stack ustack_raw;
+	struct gadget_user_stack ustack;
 	/* other fields */
 };
 
@@ -631,7 +627,7 @@ GADGET_PARAM(collect_ustack);
 	if (!event)
 		return 0;
 
-	gadget_get_user_stack(ctx, &event->ustack_raw, collect_ustack);
+	gadget_get_user_stack(ctx, &event->ustack, collect_ustack);
 ```
 
 ## Common information
--- a/gadgets/trace_open/program.bpf.c
+++ b/gadgets/trace_open/program.bpf.c
@@ -30,7 +30,7 @@ struct event {
 	__u32 fd;
 	int flags_raw;
 	__u16 mode_raw;
-	struct gadget_user_stack ustack_raw;
+	struct gadget_user_stack ustack;
 	char fname[NAME_MAX];
 };
 
@@ -117,7 +117,7 @@ static __always_inline int trace_exit(struct syscall_trace_exit *ctx)
 
 	/* event data */
 	gadget_process_populate(&event->proc);
-	gadget_get_user_stack(ctx, &event->ustack_raw, collect_ustack);
+	gadget_get_user_stack(ctx, &event->ustack, collect_ustack);
 
 	bpf_probe_read_user_str(&event->fname, sizeof(event->fname), ap->fname);
 	event->flags_raw = ap->flags;
--- a/pkg/operators/ustack/ustack.go
+++ b/pkg/operators/ustack/ustack.go
@@ -18,6 +18,7 @@ package ustack
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/cilium/ebpf"
 
@@ -27,7 +28,6 @@ import (
 	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/symbolizer"
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/annotations"
 )
 
 const (
@@ -36,7 +36,8 @@ const (
 )
 
 const (
-	userStackTargetNameAnnotation = "ebpf.formatter.ustack"
+	// Params
+	symbolizersParam = "symbolizers"
 )
 
 type Operator struct{}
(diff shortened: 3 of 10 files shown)


docs/gadget-devel/gadget-ebpf-api.md
```diff
@@ -341,7 +341,7 @@ Symbolize the kernel stack from `gadget_get_kernel_stack(ctx)` (see [kernel-stac
 
 ### `gadget_user_stack`
 
-Symbolize the user stack from `gadget_get_user_stack(ctx, &event->ustack_raw)` (see [user-stack-traces](#user-stack-traces)).
+Symbolize the user stack from `gadget_get_user_stack(ctx, &event->ustack)` (see [user-stack-traces](#user-stack-traces)).
```
flyth, line 344: Why did you remove the `_raw` here? The main reason we opted for fields with a `_raw` suffix was to not have overlapping field definitions that vary in their types depending on whether an operator was active or not. In this case, without the operator the type will be a blob of addresses, but with the operator it will become an `object` (which doesn't have that blob anymore in the currently supported output modes) with two new attributes `addresses` and `symbols`, right?

pkg/operators/ustack/ustack.go
```diff
@@ -36,7 +36,11 @@ const (
 )
 
 const (
+	// Annotations
 	userStackTargetNameAnnotation = "ebpf.formatter.ustack"
```
flyth, line 40: Not used currently, but I'd suggest to bring it back (and rename it to fit this operator - sth. like `ustack.target`).

flyth (commented):
Good idea!

Maybe later on we should allow to register custom symbolizers on init, so we don't have hard dependencies on the actual symbolizers in the operator (e.g. make `Resolve(task Task, addresses []uint64) ([]string, error)` part of an interface).

docs/gadget-devel/gadget-ebpf-api.md
alban: In both cases (with or without the operator), ustack is a struct. The operator adds some subfields in the struct but does not overwrite the type of any subfield.
I can reintroduce separate ustack and ustack_raw but it is a bit more complicated to create ustack because I then need to create both the ustack outer struct and the inner subfields.

flyth: Ah, sorry, makes sense. Let's keep it without the `_raw`, then.

pkg/operators/ustack/ustack.go
alban: Since I removed the _raw thing, I will remove this annotation too

pkg/operators/ustack/ustack.go
```diff
@@ -54,19 +55,44 @@ func (o *Operator) GlobalParams() api.Params {
...
+
+	opts := symbolizer.SymbolizerOptions{}
+	symbolizers := instanceParamValues[symbolizersParam]
+	switch symbolizers {
+	case "none":
+		opts.UseSymtab = false
+	case "auto":
+		opts.UseSymtab = true
+	default:
+		list := strings.Split(symbolizers, ",")
```
flyth, line 78: an empty list would result in an error, right? Should we check for empty list and assume it's "none"?

pkg/operators/ustack/ustack.go
```diff
@@ -54,19 +55,44 @@ func (o *Operator) GlobalParams() api.Params {
 }
 
 func (o *Operator) InstanceParams() api.Params {
-	return nil
+	return api.Params{&api.Param{
+		Key:          symbolizersParam,
+		Description:  `Symbolizers to use. Possible values are: "none", "auto", or comma-separated list among: "symtab"`,
+		DefaultValue: "auto",
```
flyth, line 61: How do we define "auto"? Would it be the same as "all" or would we differentiate? If "auto" would mean all, we could remove "none"/"auto" and just have the default value reflect all options (comma-separated). That would also make it more accessible to other frontends (e.g. we could provide "PossibleValues" and allow a set of them, so the frontend could show items you can just check/uncheck).

flyth (commented):
Looks good - I've got just two questions regarding the parameter itself.

pkg/operators/ustack/ustack.go
alban: At the moment, "auto" means "all", but as we add more options, I was thinking that "auto" should only include the options which are fast enough to be a reasonable default. For example, "goresym" would not be included in "auto".

pkg/operators/ustack/ustack.go
alban: Ok. I added "" in the `case "none"` above.

pkg/operators/ustack/ustack.go
```diff
@@ -54,19 +55,44 @@ func (o *Operator) GlobalParams() api.Params {
...
+			case "symtab":
+				opts.UseSymtab = true
+			default:
+				return nil, fmt.Errorf("invalid symbolizer: %s", s)
+			}
+		}
+	}
+
+	var err error
+	if symbolizers != "none" {
```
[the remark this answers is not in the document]
alban: The intent was for the symbolizer `symbolizer.NewSymbolizer()` to support several methods for symbolization, and using Symtab is only one of them.

pkg/symbolizer/symbolizer.go
```diff
@@ -181,6 +188,13 @@ type Task struct {
 }
 
 func (s *Symbolizer) Resolve(task Task, addresses []uint64) ([]string, error) {
+	if s.options.UseSymtab {
+		return s.resolveWithSymtab(task, addresses)
+	}
+	return make([]string, len(addresses)), nil
```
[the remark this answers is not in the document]
alban: The idea in general was that symbolization can succeed for some addresses and fail for other addresses. In this case, no symbolization is done. But the caller can still print the depth of the stack.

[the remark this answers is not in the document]
alban: I think it is too complicated for what it worth. I can switch to your earlier solution and `return nil, nil`.

alban:
Updated

gadgets/trace_open/program.bpf.c
```diff
@@ -117,7 +117,7 @@ static __always_inline int trace_exit(struct syscall_trace_exit *ctx)
 
 	/* event data */
 	gadget_process_populate(&event->proc);
-	gadget_get_user_stack(ctx, &event->ustack_raw, collect_ustack);
+	gadget_get_user_stack(ctx, &event->ustack, collect_ustack);
```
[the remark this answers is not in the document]
alban: Possibly but not in this PR.

There is a test for ustack in the `trace_capabilities` gadget. I'll think of a way to test ustack more generally (statically vs dynamically compiled programs, stripped or not, dynamic libraries, short-lived programs, etc.) and not only for a specific gadget.

PR #4179, opened 2025-03-05
ci: Use main instead of latest for container tags

mauriciovasquezbernal:
We're using latest on the wrong way, use main to indicate this is about the development version.

Part of #4049

### DIFF
--- a/.github/actions/run-integration-tests/action.yml
+++ b/.github/actions/run-integration-tests/action.yml
@@ -16,7 +16,7 @@ inputs:
     required: true
   dnstester_image:
     description: 'The image used for the dnstester.'
-    default: 'ghcr.io/inspektor-gadget/dnstester:latest'
+    default: 'ghcr.io/inspektor-gadget/dnstester:main'
   gadget_repository:
     description: 'Repository where gadget images are stored.'
     required: true
--- a/.github/actions/set-container-repo-and-determine-image-tag/action.yml
+++ b/.github/actions/set-container-repo-and-determine-image-tag/action.yml
@@ -1,6 +1,6 @@
 name: "Output the container repository and image tag to be used"
 description: "Container repository is inputs.registry/inputs.container-image.
-  Image tag is latest if branch is main, otherwise it is the tag or branch name with slashes replaced by hyphens."
+  Image tag is the tag or branch name with slashes replaced by hyphens."
 
 inputs:
   registry:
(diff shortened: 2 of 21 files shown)


alban (approved):
Good to see a patch that removes more lines than it adds.

burak-ok (commented):
Don't we also need to update all other occurences of `latest` in the CI pipeline and tests.

The CI checks if it needs to build the `ebpf-builder` image for example and would just pull from `latest` (which is not the latest anymore)

burak-ok:
Also here:
https://github.com/inspektor-gadget/inspektor-gadget/blob/d373a8aea2480dcd52112c63bfa86f01d155a2d3/.github/workflows/inspektor-gadget.yml#L9-L10

mauriciovasquezbernal:
> https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/workflows/inspektor-gadget.yml#L771

Updated it, however :main seems wrong as well. It should use the tag for this specific PR instead. 

> https://github.com/inspektor-gadget/inspektor-gadget/blob/main/.github/workflows/inspektor-gadget.yml#L2230

This is replacing :latest on the manifests to the version, so it should be kept as `:latest`. btw @eiffel-fl, is this broken, I see the image on https://github.com/inspektor-gadget/inspektor-gadget/releases/download/v0.38.0/inspektor-gadget-v0.38.0.yaml pointing to latest, shouldn't it point to v0.38.0?

mauriciovasquezbernal:
There are several chicken and egg issues on this PR, so the CI won't pass. It passed on my fork (https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/13683390170), I'll merge it tomorrow morning to have time to fix any issues it might cause.

alban:
We need to update `gadgets/Makefile` too:
```diff
--- a/gadgets/Makefile
+++ b/gadgets/Makefile
@@ -4,7 +4,7 @@ ROOT_DIR:=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 
 GADGET_TAG ?= $(shell ../tools/image-tag branch)
 GADGET_REPOSITORY ?= ghcr.io/inspektor-gadget/gadget
-BUILDER_IMAGE ?= ghcr.io/inspektor-gadget/ebpf-builder:latest
+BUILDER_IMAGE ?= ghcr.io/inspektor-gadget/ebpf-builder:main
 KERNEL_REPOSITORY ?= ghcr.io/inspektor-gadget/ci-kernels
 IG ?= ig
 KUBECTL_GADGET ?= kubectl-gadget
```

So that I can build a gadget with:
```bash
$ make -C gadgets fsnotify
```

mauriciovasquezbernal:
Opened #4183 to fix the missing refs

PR #4181, opened 2025-03-06
ci: fix permissions

eiffel-fl:
We also have this one:
https://github.com/inspektor-gadget/inspektor-gadget/security/code-scanning/403
But I do not see a proper way to fix it, as we indeed need it.

### DIFF
--- a/.github/workflows/dockerhub-mirror.yml
+++ b/.github/workflows/dockerhub-mirror.yml
@@ -22,8 +22,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - image: alpine:latest
-            repo: docker.io/library
           - image: busybox:latest
             repo: docker.io/library
           - image: gcc:latest


mqasimsarfraz (commented):
Ready to go once CI is green! 

Please delete the package after merge: https://github.com/inspektor-gadget/inspektor-gadget/pkgs/container/ci%2Falpine to avoid confusion in future.

mqasimsarfraz (approved):

mqasimsarfraz:
> We also have this one: [Token-Permissions](/inspektor-gadget/inspektor-gadget/security/code-scanning/403) But I do not see a proper way to fix it, as we indeed need it.

I don't understand this as well but after merge "Dangerous-Workflow" is back to 10/10 (before it was 0/10) https://scorecard.dev/viewer/?uri=github.com/inspektor-gadget/inspektor-gadget so we should be fine.

PR #4198, opened 2025-03-10
pkg/gadget-context: sort operators by priority and name to be deterministic

flyth:
Just taking the priority of an operator is not enough to get a deterministic order of the operators. In case two operators have the same priority, the name should also be considered for the run order.

Non-deterministic ordering can mix up for example fields in data sources, when multiple nodes are running the same operators but in a different oder.

/cc @burak-ok

### DIFF
--- a/pkg/gadget-context/context_test.go
+++ b/pkg/gadget-context/context_test.go
@@ -1,0 +1,85 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


burak-ok (approved):
LGTM, thanks for fixing this.

Could you add a small unit test for this (If its easily doable currently without too much mocking)

flyth:
Agree that it needs to be done sooner rather than later, but this doesn't have anything to do with the deterministic issue that we're fixing here 😃

burak-ok:
WHen I implemented the `KubeNameResolver` I knew that I put the same priority as `KubeIPResolver`. I just thought it would be deterministic

alban (approved):

flyth:
Thanks for the reviews!

PR #4203, opened 2025-03-11
Bug fix - use skb load bytes instead of bpf read

afek854:
# [Title: Bug fix for kernel version 5.4.0]

The following error occur when trying to load dns tracer
creating BPF collection: program ig_trace_dns: load program: invalid argument: unknown func bpf_probe_read_str#4 (575 line(s) omitted)

It is because bpf_probe_read_kernel_str does not supported in 5.4 &  bpf_probe_read_str does not supported in 5.4 as in socket filter.

Thanks


alban:
The same thing needs to be fixed in image-based gadget: gadgets/trace_dns/program.bpf.c

Could we use conditional:

```
        if (bpf_core_enum_value_exists(enum bpf_func_id,
                                       BPF_FUNC_probe_read_kernel_str)) {
                bpf_probe_read_kernel_str(...);
        } else {
                bpf_skb_load_bytes(...);
        }
```

afek854:
Hey, i do not think those errors are related to my PR. 
Am i wrong?

burak-ok (approved):
LGTM. Thanks

gadgets/trace_dns/program.bpf.c
```diff
@@ -364,11 +364,34 @@ int ig_trace_dns(struct __sk_buff *skb)
...
+			    sizeof(event->exepath)) {
+				int exepath_len = sizeof(skb_val->exepath);
+				if (bpf_skb_load_bytes(
+					    skb,
+					    (unsigned long)skb_val->exepath,
+					    event->exepath, exepath_len) < 0) {
+					return 0;
+				}
+			}
+		}
```
mauriciovasquezbernal, line 394: How is it supposed to work? The documentation of [bpf_skb_load_bytes](https://docs.ebpf.io/linux/helper-function/bpf_skb_load_bytes/) says:

> This helper was provided as an easy way to load data from a packet. It can be used to load len bytes from offset from the packet associated to skb, into the buffer pointed by to.

Here it's being used to copy data from one structure (not the packet) to another. If I try to enforce this to be used:

```diff
$ git diff
diff --git a/gadgets/trace_dns/program.bpf.c b/gadgets/trace_dns/program.bpf.c
index 8bfcbf5c4..1000ba864 100644
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -365,19 +365,20 @@ int ig_trace_dns(struct __sk_buff *skb)
        gadget_process_populate_from_socket(skb_val, &event->proc);

        if (paths && skb_val != NULL) {
-               if (bpf_core_enum_value_exists(
-                           enum bpf_func_id, BPF_FUNC_probe_read_kernel_str)) {
-                       bpf_probe_read_kernel_str(
-                               &event->cwd, sizeof(event->cwd), skb_val->cwd);
-                       bpf_probe_read_kernel_str(&event->exepath,
-                                                 sizeof(event->exepath),
-                                                 skb_val->exepath);
-               } else {
+               //if (bpf_core_enum_value_exists(
+               //          enum bpf_func_id, BPF_FUNC_probe_read_kernel_str)) {
+               //      bpf_probe_read_kernel_str(
+               //              &event->cwd, sizeof(event->cwd), skb_val->cwd);
+               //      bpf_probe_read_kernel_str(&event->exepath,
+               //                                sizeof(event->exepath),
+               //                                skb_val->exepath);
+               //} else {
                        if (sizeof(skb_val->cwd) <= sizeof(event->cwd)) {
                                int cwd_len = sizeof(skb_val->cwd);
                                if (bpf_skb_load_bytes(
                                            skb, (unsigned long)skb_val->cwd,
                                            event->cwd, cwd_len) < 0) {
+                                       bpf_printk("returning here");
                                        return 0;
                                }
                        }
@@ -389,10 +390,11 @@ int ig_trace_dns(struct __sk_buff *skb)
                                            skb,
                                            (unsigned long)skb_val->exepath,
                                            event->exepath, exepath_len) < 0) {
+                                       bpf_printk("returning here2");
                                        return 0;
                                }
                        }
-               }
+               //}
        }

        // Handle nameserver
```

I don't see any event printed:

```bash
$ sudo ig run trace_dns --verify-image=false --paths --fields cwd,exepath,proc.comm,proc.creds.uid
[sudo] password for mauriciov:
CWD                                               EXEPATH                                           COMM                UID

```

and the following is printed to the kernel logs:

```bash
$ cat /sys/kernel/debug/tracing/trace_pipe
           <...>-146197  [015] b..11 19417.077303: bpf_trace_printk: returning here
        nslookup-146197  [015] b..11 19417.077369: bpf_trace_printk: returning here
          <idle>-0       [013] ..s21 19417.090046: bpf_trace_printk: returning here
          <idle>-0       [013] ..s21 19417.090586: bpf_trace_printk: returning here
           <...>-146199  [013] b..11 19418.094505: bpf_trace_printk: returning here
```

cc @burak-ok @alban 

(found while working on #4543)

alban: @mauriciovasquezbernal Good finding... I guess it does not work.

The easiest solution would be to leave the fields `cwd` and `exepath` empty when `BPF_FUNC_probe_read_kernel_str` does not exist, and add this limitation in the documentation.

> bpf_probe_read_str does not supported in 5.4 in socket filter ebpf's

@afek854 Do you know more about this limitation? In [kernel-versions.md](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md), I see probe_read_str is in Linux 4.11. Was the support for probe_read_str in socket filter programs added later on or not at all?

mauriciovasquezbernal: According to https://docs.ebpf.io/linux/helper-function/bpf_probe_read_str/ it's not supported. Perhaps a memory copy can be a good fallback when BPF_FUNC_probe_read_kernel_str isn't present?

alban: How would you do a memory copy? With `__builtin_memcpy`? AFAIR, the compiler converts `__builtin_memcpy` to a loop when the size is too big (or unknown at compile-time). So it could fail the verification.

Could you open an issue for this discussion?

mauriciovasquezbernal: Done in #4598

PR #4209, opened 2025-03-14
wasm: Change module name from env to ig

burak-ok:
We should name our module `ig` instead of `env`

### DIFF
--- a/docs/gadget-devel/gadget-wasm-api-raw.md
+++ b/docs/gadget-devel/gadget-wasm-api-raw.md
@@ -20,7 +20,7 @@ contains the length and the lower 32 the memory address.
 
 ## Wasm Module Exported Functions
 
-The Wasm module implemented by the gadget also needs to export some functions to
+The Wasm program implemented by the gadget also needs to export some functions to
 be invoked by the host.
 
 ### `gadgetAPIVersion`
@@ -58,6 +58,8 @@ This function is called after the the gadget is stopped. This function is option
 See description in dataSourceSubscribe below.
 
 ## API
+
+The Wasm API provided to the gadget resides in the `ig` module.
 
 ### Log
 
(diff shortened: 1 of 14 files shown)


alban:
Does it break compatibility if a gadget was built before this change and run with ig compiled after this change?

If the module name is part of the API, then it should be documented in docs/gadget-devel/gadget-wasm-api-raw.md

burak-ok:
> Does it break compatibility if a gadget was built before this change and run with ig compiled after this change?

Yes, this is (again) a breaking change.
> What is the error message if that does not work?

I don't know for sure but it will be along the lines of `function can not be found in module "env"` or so

> If the module name is part of the API, then it should be documented in docs/gadget-devel/gadget-wasm-api-raw.md

We didn't name the module and since WASM programs can provide multiple modules I change the doc to specify that we need the listed functions in the `ig` module

mqasimsarfraz:
> We should name our module ig instead of env

I am trying to understand why do we need this change? Are there any best practices / conventions?

burak-ok:
> > We should name our module ig instead of env
> 
> I am trying to understand why do we need this change? Are there any best practices / conventions?

There is no need, just a "cleanup".

docs/gadget-devel/gadget-wasm-api-raw.md
```diff
@@ -20,7 +20,7 @@ contains the length and the lower 32 the memory address.
 
 ## Wasm Module Exported Functions
 
-The Wasm module implemented by the gadget also needs to export some functions to
+The `ig` wasm module implemented by the gadget also needs to export some functions to
 be invoked by the host.
```
alban, line 24: Functions like `newDataSource` are **imported** in the gadget wasm module, so the "ig" name is implemented in the wasm host and not by the gadget.

Could this be rephrased accordingly?

burak-ok: Seems like I totally changed the wrong place in the documentation. Fixed it

alban (approved):
LGTM, but could you find out the exact error message for this:

> > What is the error message if that does not work?
>
> I don't know for sure but it will be along the lines of function can not be found in module "env" or so

burak-ok:
> LGTM, but could you find out the exact error message for this:

Running "old" gadgets with this PR results in:
`Error: fetching gadget information: initializing and preparing operators: instantiating operator "oci": instantiating operator "wasm": initializing wasm: instantiating wasm: module[env] not instantiated`

PR #4221, opened 2025-03-18
golangci-lint: Update to v1.64.8

burak-ok:
The lint step in https://github.com/inspektor-gadget/inspektor-gadget/pull/4214 fails. While running `make lint` locally on that branch, the RAM usage just keeps rising (over 30GB) and then it gets OOM killed.

Updating `golangci-lint` resolves this issue.


alban (approved):
LGTM

burak-ok:
`Scan gadget img` step doesn't fail because of this PR. Will merge

PR #4241, opened 2025-03-23
Cli Operator Unit Tests

Shaheer-Ahmd:
# Add CLI Operator Tests for Various Field Formats

This PR introduces a set of tests for the CLI operator functionality, covering various input field scenarios. The tests evaluate how the system handles different input values, such as single fields, multiple fields, empty fields, numbers as strings, and cases with a large number of fields. The tests ensure that the expected outputs are generated correctly in default, YAML, and JSON formats. (an effort towards #3835 )

The changes include the addition of two major test functions: `TestCliOperators` and `TestJsonArray`. Each test checks the correct output for different combinations of input values. This includes checking if the program formats single and multiple fields correctly in different serialization formats (default, YAML, JSON), and ensures proper handling of edge cases like empty values and large sets of fields.

## How to use

Reviewers can validate this PR by running the tests in the `pkg/operators/cli` package. To run the tests, use the following command:

```
go test -v ./pkg/operators/cli
```

Ensure that all tests pass and confirm that the outputs are correct for each input scenario.

## Testing done

I ran the tests for the CLI operator functionality using the following command:

```
go test -v ./pkg/operators/cli
```

All tests passed successfully. The output for each test case was checked against the expected results, and it correctly handled various scenarios including:

- Single and multiple fields with values.
- Empty values and fields.
- Handling of numbers as strings.
- Large number of fields.

Each output format (default, YAML, and JSON) was validated, ensuring the system's behavior aligns with the expectations defined in the test cases.

### DIFF
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -1,4 +1,4 @@
-// Copyright 2024 The Inspektor Gadget authors
+// Copyright 2024-2025 The Inspektor Gadget authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@ package clioperator
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"slices"
 	"sort"
@@ -334,11 +335,7 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error 
 			}
 			ds.Subscribe(func(ds datasource.DataSource, data datasource.Data) error {
 				before()
-				for _, f := range ds.Accessors(false) {
-					if s, err := f.String(data); err == nil {
-						fmt.Print(s)
-					}
-				}
+				defaultDataFn(ds, data, os.Stdout)
 				return nil
 			}, Priority)
 		case ModeNone:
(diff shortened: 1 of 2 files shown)


pkg/operators/cli/clioperator.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 The Inspektor Gadget authors
+// Copyright 2024-25 The Inspektor Gadget authors
```
burak-ok, line 1: I think we use the full year everywhere else
```suggestion
// Copyright 2024-2025 The Inspektor Gadget authors
```

pkg/operators/cli/clioperator_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource/formatters/json"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCliOperators(t *testing.T) {
```
burak-ok, line 30: Could you add another test function (can be a single test) with nested values?

burak-ok (commented):
Looks good, thanks

burak-ok (approved):

PR #4249, opened 2025-03-24
Added unit tests for formmaters in ebpf operator

Shaheer-Ahmd:
# Add parameterized test cases for stack trace and enum formatting

This PR introduces parameterized test cases for validating the `fetchAndFormatStackTrace` function and the `getFormattersForEnums` function. These tests ensure correct functionality for various stack trace scenarios and enum formatting conditions. (an effort towards #3835 )

For `fetchAndFormatStackTrace`, the test cases verify correct stack trace formatting for:
- A valid stack trace with multiple entries.
- An empty stack trace.
- A single valid entry.

For `getFormattersForEnums`, the test cases validate:
- Correct mapping of non-bitfield enums.
- Proper handling of bitfield enums with different separators.
- Handling of unknown enum values.
- Cases where multiple enums are present.

## How to use

To validate this PR, reviewers should:
1. Run the tests using:
   ```sh
   go test -run TestFetch -v ./pkg/operators/ebpf/; go test -run TestGet -v ./pkg/operators/ebpf/

   ```
2. Verify that all test cases pass without errors.
3. Review the added parameterized tests to confirm they cover edge cases and expected scenarios.

## Testing done

The following command was run to execute the test cases:
```sh
go test -run TestFetch -v ./pkg/operators/ebpf/; go test -run TestGet -v ./pkg/operators/ebpf/

```
Expected output:
```
=== RUN   TestFetchAndFormatStackTrace
=== RUN   TestFetchAndFormatStackTrace/Valid_stack_trace
=== RUN   TestFetchAndFormatStackTrace/Empty_stack_trace
=== RUN   TestFetchAndFormatStackTrace/Single_valid_entry
--- PASS: TestFetchAndFormatStackTrace (0.00s)
    --- PASS: TestFetchAndFormatStackTrace/Valid_stack_trace (0.00s)
    --- PASS: TestFetchAndFormatStackTrace/Empty_stack_trace (0.00s)
    --- PASS: TestFetchAndFormatStackTrace/Single_valid_entry (0.00s)
PASS
ok  	github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf	(cached)
=== RUN   TestGetFormattersForEnums
=== RUN   TestGetFormattersForEnums/Single_non-bitfield_enum_valid
=== RUN   TestGetFormattersForEnums/Single_non-bitfield_enum_valid_no_target_annotation
=== RUN   TestGetFormattersForEnums/Single_bitfield_enum_multiple_flags
=== RUN   TestGetFormattersForEnums/Non-bitfield_enum_unknown
=== RUN   TestGetFormattersForEnums/Non-bitfield_enum_with_multiple_similar_values
=== RUN   TestGetFormattersForEnums/Bitfield_enum_no_flags_set
=== RUN   TestGetFormattersForEnums/Bitfield_enum_custom_separator
=== RUN   TestGetFormattersForEnums/Multiple_enums,_first_non-match_then_match
=== RUN   TestGetFormattersForEnums/Multiple_non-bitfield_enums_both_matching
=== RUN   TestGetFormattersForEnums/Non-bitfield_then_bitfield_enum,_bitfield_wins
=== RUN   TestGetFormattersForEnums/Multiple_bitfield_enums,_last_wins
--- PASS: TestGetFormattersForEnums (0.00s)
    --- PASS: TestGetFormattersForEnums/Single_non-bitfield_enum_valid (0.00s)
    --- PASS: TestGetFormattersForEnums/Single_non-bitfield_enum_valid_no_target_annotation (0.00s)
    --- PASS: TestGetFormattersForEnums/Single_bitfield_enum_multiple_flags (0.00s)
    --- PASS: TestGetFormattersForEnums/Non-bitfield_enum_unknown (0.00s)
    --- PASS: TestGetFormattersForEnums/Non-bitfield_enum_with_multiple_similar_values (0.00s)
    --- PASS: TestGetFormattersForEnums/Bitfield_enum_no_flags_set (0.00s)
    --- PASS: TestGetFormattersForEnums/Bitfield_enum_custom_separator (0.00s)
    --- PASS: TestGetFormattersForEnums/Multiple_enums,_first_non-match_then_match (0.00s)
    --- PASS: TestGetFormattersForEnums/Multiple_non-bitfield_enums_both_matching (0.00s)
    --- PASS: TestGetFormattersForEnums/Non-bitfield_then_bitfield_enum,_bitfield_wins (0.00s)
    --- PASS: TestGetFormattersForEnums/Multiple_bitfield_enums,_last_wins (0.00s)
PASS
ok  	github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf	0.177s
```

All test cases passed successfully, ensuring correctness of stack trace formatting and enum conversion logic.

### DIFF
--- a/pkg/operators/ebpf/formatters.go
+++ b/pkg/operators/ebpf/formatters.go
@@ -1,4 +1,4 @@
-// Copyright 2024 The Inspektor Gadget authors
+// Copyright 2024-2025 The Inspektor Gadget authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/kallsyms"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/annotations"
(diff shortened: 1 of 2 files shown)


pkg/operators/ebpf/formatters.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 The Inspektor Gadget authors
+// Copyright 2024-25 The Inspektor Gadget authors
```
burak-ok, line 1:
```suggestion
// Copyright 2024-2025 The Inspektor Gadget authors
```

pkg/operators/ebpf/formatters_test.go
```diff
@@ -0,0 +1,499 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
+	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
+)
+
+func TestGetFormattersForEnums(t *testing.T) {
+	type testCaseDatum struct {
+		enum             *enum
+		inputValue       uint64
+		expectedOutput   string
+		fieldName        string
```
burak-ok, line 34: A little bit more descriptive what that member actually describes. I hope I understood that correctly
```suggestion
		expectedTargetFieldName        string
```

pkg/operators/ebpf/formatters_test.go
```diff
@@ -0,0 +1,499 @@
...
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/logger"
+	ebpftypes "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf/types"
+)
+
+func TestGetFormattersForEnums(t *testing.T) {
+	type testCaseDatum struct {
+		enum             *enum
```
burak-ok, line 31: I think we could define a set of enums outside of the `testCases` and then just use them in the test cases. Therefore we wouldn't need to have that many repetitive definitions of the `enum` struct

AFAIU the `isMatch` is also only used to add enum definitions inside a test case. And we wouldn't need the variable after the above change

Something like https://gist.github.com/burak-ok/14eb0eb1c13f1e3ffa7753a70c497116 . What do you think? Or would this make it make complex to understand?

pkg/operators/ebpf/formatters_test.go
Shaheer-Ahmd: yeah that would be more descriptive, thanks

pkg/operators/ebpf/formatters_test.go
Shaheer-Ahmd: yes, that would definitely be more cleaner and readable, thanks. However, in some testcases, I need to alter the memberName (for no targetAnnotation case) and a single variable can't be used for every testcase. Therefore, now I have implemented a hybrid approach, a single variable for repetetive cases and defined a function for the other cases.

burak-ok: Nice idea to also make a generator function for the enums.

I am still not sure about this whole `isMatch`. When I modify the expected values in a test case where `isMatch: false`, the test is still successfull: https://gist.github.com/burak-ok/f4aadc3b2994ff884aedeea370a08ba7

If the intention is to have multiple enum structs for a test I suggest to move the `enum` field from `testCaseDatum` to `testCase` and make it `enums []*enum` where you are able to put in multiple enums

Shaheer-Ahmd: the `getFormattersForEnums` function returns a list of formatter functions for all the matching enums in a list and we do not have a way to see which formatter function corresponds to which enum. This is why I felt the need to introduce a boolean variable that could tell us which enums we can skip while calling the formatter functions.
e.g we have 5 enums and 3 matches. The `getFormattersForEnums` function would return an array of 3 formatter functions that correspond to the 3 matched enums. Now, we don't know which enums correspond to the 3 formatter functions that we got which is exactly why I added the `isMatch` field.

 If isMatch is false, the expectedValue is unused which is why the testcase is still passing in your case. I could remove the unused fields for such scenarios.

burak-ok: > Now, we don't know which enums correspond to the 3 formatter functions that we got which is exactly why I added the isMatch field.

That makes sense, thanks for the explanation.

> If isMatch is false, the expectedValue is unused which is why the testcase is still passing in your case. I could remove the unused fields for such scenarios.

Since they are not needed it would be nice to remove them. Also helps with understanding the testcode

Shaheer-Ahmd: done

PR #4271, opened 2025-03-31
inspect: makes pull policy and verify-image configurable

KapilSareen:
## Changes

adds flags oci flags to `inspect` command, mainly making pull policy and image verification configurable

## Testing done

```
$ sudo ./ig image inspect trace_dns:main --pull never --verify-image=false
{
  "ebpf.maps": {
    "content": "[{\"Name\":\"gadget_mntns_filter_map\",\"Type\":\"Hash\"},{\"Name\":\"tmp_events\",\"Type\":\"PerCPUArray\"},{\"Name\":\"query_map\",\"Type\":\"Hash\"},{\"Name\":\"events\",\"Type\":\"PerfEventArray\"},{\"Name\":\"gadget_sockets\",\"Type\":\"Hash\"}]",
    "contentType": "application/json"
  },
  "ebpf.programs": {
...
  ```

Related to #4270

### DIFF
--- a/cmd/common/image/inspect.go
+++ b/cmd/common/image/inspect.go
@@ -26,12 +26,10 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/config"
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/oci"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 
 	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	apihelpers "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api-helpers"
 	ocihandler "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/oci-handler"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/common/image/inspect.go [hunk 1]
```diff
@@ -162,5 +156,15 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 		fmt.Sprintf("Output mode, possible values are, %s", strings.Join(outputModes, ", ")),
 	)
 
+	var skipParams []string
+	if commandMode == common.CommandModeAttach {
+		skipParams = append(skipParams, "!attach")
+	}
```
blanquicet, line 162: We don't want to add the headless-related flags to the inspect command as it will never run in headless mode. So, we don't need to pass the `commandMode` around. We can simply define the `skipParams` like this:

```golang
        // We don't want to add the headless-related flags to the inspect command
	skipParams := []string{"!attach"}
```

blanquicet (commented):
Some comments but it looks good.

cmd/common/image/inspect.go
```diff
@@ -37,9 +36,11 @@ import (
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
 )
 
-func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
+func NewInspectCmd(runtime runtime.Runtime, commandMode common.CommandMode) *cobra.Command {
```
blanquicet, line 39: This is not needed anymore

KapilSareen: sorry for that, missed it.

cmd/common/image/inspect.go
```diff
@@ -40,6 +39,8 @@ import (
 func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 	var outputMode string
 
+	opGlobalParams := make(map[string]*params.Params)
+
 	outputModes := []string{utils.OutputModeYAML, utils.OutputModeJSON, utils.OutputModeJSONPretty, utils.OutputModeCustom}
 
 	cmd := &cobra.Command{
```
blanquicet, line 46: With this change, by default, the pull-policy will be `missing`. So, people don't need to pull the image in advance. So, I suggest updating the description to `Inspect a gadget image`. WDYT?

blanquicet (approved):
LGTM!

cmd/common/image/inspect.go
KapilSareen: yeah, makes sense to me

cmd/common/image/inspect.go
```diff
@@ -40,47 +38,37 @@ import (
...
+	runtimeGlobalParams := runtime.GlobalParamDescs().ToParams()
+	runtimeParams := runtime.ParamDescs().ToParams()
 
-		var info *api.GadgetInfo
-		err = runtime.Init(runtimeGlobalParams)
-		if err != nil {
-			return fmt.Errorf("initializing runtime: %w", err)
-		}
-		defer runtime.Close()
+	runtime.Init(runtimeGlobalParams)
```
alban, line 63: Calling `runtime.Init(runtimeGlobalParams)` in `NewInspectCmd()` means it will always be executed, even when the user does not run the `image inspect` command. It should be moved below in the `cmd.RunE` callback, so it is only executed when the user uses the `image inspect` command.

This is causing #4301.

PR #4282, opened 2025-04-03
Unit Tests for Inventory Cache

Shaheer-Ahmd:
# Add comprehensive unit tests for inventoryCache lifecycle and CRUD operations

This PR introduces a full suite of unit tests for the `inventoryCache` component in `pkg/operators/common/`. The tests cover the complete lifecycle of the cache, including `Start()`/`Stop()` reference counting, and all CRUD handlers (`OnAdd`, `OnUpdate`, and `OnDelete`) for both Pods and Services. Additionally, various edge cases are validated, such as handling invalid inputs, unknown types, and missing metadata that lead to expected logging of errors.

The new test cases improve confidence in the correctness of the cache implementation and its interactions with underlying Kubernetes objects. These tests validate behavior for lookup by name and IP, and also ensure correct log output when expected errors occur. Each test is isolated and uses the `cachedmap` with short timeouts to simulate expiration behavior.
(an effort towards #3835)
## How to use

To validate this PR:
1. Run the tests with the following command:
   ```
   go test -v ./pkg/operators/common/
   ```
2. Verify that all test cases pass and produce the expected output.
3. Optionally review log assertions in failure cases to ensure meaningful messages are emitted for unexpected input.

## Testing done

All tests were run locally using:

```
go test -v ./pkg/operators/common/
```

All test cases passed successfully, including:
- Reference counting in `Start()`/`Stop()`
- Adding valid and invalid Pods/Services
- Updating objects and verifying new IP-based lookups
- Deleting objects and ensuring they are removed from all relevant maps
- Verifying error logs when invalid inputs are passed

These tests provide robust coverage and ensure the `inventoryCache` behaves as expected under both typical and edge-case scenarios.

### DIFF
--- a/pkg/operators/common/common_test.go
+++ b/pkg/operators/common/common_test.go
@@ -1,0 +1,362 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"bytes"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/cachedmap"
+)
+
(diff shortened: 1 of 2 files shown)


Shaheer-Ahmd:
Questions:
1- Currently, adding an element with an empty key is allowed in the pods and svcs maps. Is this the correct behavior? Since an empty key might not have a real meaning associated with it.
2- Updating an object in the cache only introduces a new object without removing the old object. Is this the correct behavior?

burak-ok (approved):
LGTM.

Just a general question: Do you think we should add another test for the following case:
1. Create Cache with 1s TTL
2. Add and delete a Pod
3. Check you can still get the Pod in a 1s window after deletion?

When I wrote the tests for `CachedMap` I added such tests - but the `KubeInvCache` is "just using" the `CachedMap`. So I think it is not neccessary to add such a test again, right?
If we would add such tests that would lead to copy pasting every test and having multiples of them.

Shaheer-Ahmd:
> Do you think we should add another test for the following case.

I would argue against it aswell since the primary purpose of adding the delete testcase was to cover the 3 branches in the `OnDelete`(pods, svc and invalid case). Checking for different TTLs should not be a concern of the tests at this level, instead, they should be in the `cachedMap` package as you have already implemented. Writing unit tests for a functionality of 1 package in another package makes you loose control over the unexposed functions of the original package which makes it hard to sub away any dependencies.

PR #4290, opened 2025-04-06
inspect: adds command to kubectl-gadget

KapilSareen:
## Changes
- extends the `NewImageCmd` to receive the new 'addCommands' param which can be used to add image commands optionally in `gadgetctl/kubectl-gadget` (only commands that are grpc-supported)
- adds inspect command in `kubectl-gadget`

Part of #3387

### DIFF
--- a/cmd/kubectl-gadget/main.go
+++ b/cmd/kubectl-gadget/main.go
@@ -26,6 +26,7 @@ import (
 	paramsPkg "github.com/inspektor-gadget/inspektor-gadget/pkg/params"
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
+	img "github.com/inspektor-gadget/inspektor-gadget/cmd/common/image"
 	commonutils "github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/advise"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/kubectl-gadget/utils"
@@ -153,6 +154,11 @@ func main() {
 		log.Warnf(err.Error())
 	}
 
+	// add image subcommands to be added, for now only inspect is supported
+	imgCommands := []*cobra.Command{
+		img.NewInspectCmd(grpcRuntime),
+	}
+
 	gadgetNamespace := runtimeGlobalParams.Get(grpcruntime.ParamGadgetNamespace).AsString()
 
 	hiddenColumnTags := []string{"runtime"}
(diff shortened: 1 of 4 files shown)


cmd/kubectl-gadget/main.go
```diff
@@ -167,6 +168,7 @@ func main() {
 	rootCmd.AddCommand(common.NewRunCommand(rootCmd, grpcRuntime, hiddenColumnTags, common.CommandModeRun))
 	rootCmd.AddCommand(common.NewRunCommand(rootCmd, grpcRuntime, hiddenColumnTags, common.CommandModeAttach))
 	rootCmd.AddCommand(common.NewConfigCmd(grpcRuntime, rootFlags))
+	rootCmd.AddCommand(img.NewImageCmd(grpcRuntime))
```
alban, line 171: It seems to work for the `image inspect` command.

However, the other  `image` commands try to use the local oci-store instead of the remote one:
```
$ ./kubectl-gadget image list
Error: list gadgets: getting oci store: mkdir /var/lib/ig/oci-store: permission denied
```

```
$ ./kubectl-gadget image pull trace_exec:main
Pulling trace_exec:main...
Error: pulling gadget: getting oci store: mkdir /var/lib/ig/oci-store: permission denied
```

I don't think the grpc API has something for all `image` commands.

I suggest to start with only the `image inspect` command in this PR, and open a separate issue for the other `image` commands.

KapilSareen: Sure!

KapilSareen: Now it'll be used as:
```
$ ./kubectl-gadget inspect trace_dns:main
```

Is it fine or would you like it to be consistant with gadgetctl?

alban: I think it is better to be consistent with gadgetctl, so later on when adding the other image commands, we don't have to change the UX.

alban: I think it does not exist at the moment... I am not sure how to do it.

I wonder about adding all the commands but `kubectl-gadget image pull` would just print "Not implemented yet" to the user. So adding the commands but detecting it is not in `ig` but in `gadgetctl` or `kubectl-gadget`.

By the way, the `main` branch has the same problem:
```
$ ./gadgetctl image pull trace_exec:foo
Pulling trace_exec:foo...
Error: pulling gadget: getting oci store: mkdir /var/lib/ig/oci-store: permission denied
```

KapilSareen: @alban have you tried using sudo? It works for me, but I haven't checked for kubectl-gadgetctl
```
$ ./gadgetctl image list
Error: list gadgets: getting oci store: mkdir /var/lib/ig/oci-store: permission denied
$ sudo ./gadgetctl image list
REPOSITORY                                            TAG                                                   DIGEST       CREATED                                             
trace_dns                                             main                                                  87de05648f4a 5 days ago                                          
trace_dns                                             main                                                  
```

alban: I have not tried with sudo, but using sudo is not a correct solution: the client (`gadgetctl --remote-address=tcp://10.0.0.42:1234`) and the server (`ig daemon --host=tcp://10.0.0.42:1234`) might run on different computers and communicate via TCP. Using sudo on the client would access `/var/lib/ig/oci-store` on the wrong computer.

alban (approved):
LGTM, thanks!

PR #4302, opened 2025-04-08
Use bytes.IndexBytes() in *romCString*()

eiffel-fl:


flyth:
Just out of curiosity: did you run benchmarks for this? Afair these specific for loops get highly optimized by the compiler.

burak-ok (approved):
LGTM

Benchmark:
```
goos: linux
goarch: amd64
pkg: traceloop
cpu: 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz
              │   old.txt   │            new.txt            │
              │   sec/op    │   sec/op     vs base          │
FromCString-8   17.63n ± 3%   17.25n ± 4%  ~ (p=0.075 n=10)

              │  old.txt   │            new.txt             │
              │    B/op    │    B/op     vs base            │
FromCString-8   16.00 ± 0%   16.00 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

              │  old.txt   │            new.txt             │
              │ allocs/op  │ allocs/op   vs base            │
FromCString-8   1.000 ± 0%   1.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

PR #4304, opened 2025-04-08
fixes arbitrary Runtime.Init call from inspect command

### LINKED ISSUE #4301: --auto-mount-filesystems is required when pulling a gadget artifact using 'ig image pull'
## Description

When using `ig image pull` to pull a gadget artifact it fails with this error: "error: filesystems debugfs, tracefs not mounted (did you try --auto-mount-filesystems?)".

This error is new to version v0.39.0 (didn't occur in v0.38.1 nor v0.38.0).

## Impact
--auto-mount-filesystems flag has to be provided in order to pull a gadget image.

## Expected behavior

The flag shouldn't be required for pulling gadget images.

KapilSareen:
Fixes #4301


alban (approved):
It seems to work for me. The `image pull` command no longer calls `host.Init()`. And the `image inspect` command still works in my tests.

PR #4309, opened 2025-04-09
feat(helm): missing Kubernetes recommended labels

nlamirault:
# Add missing Kubernetes recommended labels

Support all [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)


mqasimsarfraz (approved):
LGTM!

As a follow-up I will handle changing [`skipLabels`](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/charts/gadget/values.yaml#L81) to false so we can make use of these labels by default! Something like this: https://github.com/inspektor-gadget/inspektor-gadget/commit/44b646c1f994f2563760eff0b9ae64cb4d9f183e

Issue #4060, opened 2025-02-14
[RFE] Add support for ebpf fields of type "array of struct"


alban:
## Current situation

There are restrictions on events sent in a bpf ring buffer: they currently cannot be of type array of struct.

## Impact

Example of unsupported type:
```c
struct pidnumber {
	__u32 pid;
	__u32 pidns;
}
struct gadget_user_stack {
    struct pidnumber pidnumbers[3];
}
```
Output:
```
DEBU[0002]  skipping field "ustack_raw.pidnumbers" (*btf.Array)
```

## Ideal future situation

Support for array of structs.

## Implementation options

Unknown

## Additional information

It would be useful for #4004. But it is not a blocker: I can manage without arrays for now.

_Originally posted by @eiffel-fl in https://github.com/inspektor-gadget/inspektor-gadget/pull/4004#discussion_r1955694973_

mauriciovasquezbernal:
It seems like a duplicated of #3032

PR #4314, opened 2025-04-10
ustack: add buildid field for client-side symbolization

alban:
# ustack: add buildid field for client-side symbolization

Executables in the ubuntu container image don't have symbol tables but they have a buildid that can be resolved on [debuginfod](https://sourceware.org/elfutils/Debuginfod.html) servers.

## How to use

```
sudo ig run ghcr.io/inspektor-gadget/gadget/trace_capabilities --collect-ustack --comm=chroot -o yaml

ustack:
  addresses: '[0]0x00007f46ba58e4fb; [1]0x00007f46ba4913b8; [2]0x00007f46ba49147b;
    [3]0x000055b3b6b0e2d5; '
  buildid: '[0]91f01b4ad171c80b6303d08d1f08cba8b990413d +1274fb; [1]91f01b4ad171c80b6303d08d1f08cba8b990413d
    +2a3b8; [2]91f01b4ad171c80b6303d08d1f08cba8b990413d +2a47b; [3]ae03c7096b4806aa173c0fd1861b019daf6a057a
    +32d5; '
  symbols: ""
```

Test workload:
```
docker run -ti --rm ubuntu:24.10 chroot /
```

The symbols can be resolved client-side with:

```
$ sudo ig run ghcr.io/inspektor-gadget/gadget/trace_capabilities \
	--collect-ustack --comm=chroot -o json | \
	jq --unbuffered -r .ustack.buildid | \
	./symbolizer.sh
__GI_chroot
__libc_start_call_main
call_init
_start
```

symbolizer.sh contains the following:
```bash
#!/bin/bash
export DEBUGINFOD_URLS="https://debuginfod.elfutils.org"
while IFS= read -r line; do
    stack="$(echo "${line%;}" | tr ';' '\n')"
    while IFS= read -r entry; do
        if [[ "$entry" =~ \[.*\]([0-9a-fA-F]+)\ \+(.*) ]]; then
            buildid="${BASH_REMATCH[1]}"
            offset="${BASH_REMATCH[2]}"
            sudo --preserve-env=DEBUGINFOD_URLS DEBUGINFOD_VERBOSE=false \
                debuginfod-find debuginfo $buildid &> /dev/null
            sudo addr2line  -e /root/.cache/debuginfod_client/$buildid/debuginfo -f -C $offset 2>/dev/null | head -1
        fi
    done <<< "$stack"
done
```

## TODO

- [x] Decide how to enable/disable the buildid field from the ebpf program
- [x] Client-side operator for symbolization with debuginfod.

### DIFF
--- a/docs/reference/stack-traces.mdx
+++ b/docs/reference/stack-traces.mdx
@@ -1,0 +1,251 @@
+---
+title: Stack traces
+sidebar_position: 1300
+description: Kernel and user stack traces
+---
+
+Inspektor Gadget supports collecting kernel and user space stack traces through
+eBPF. Stack traces provide valuable context about what code paths led to an
+event, helping with debugging, performance analysis (including generating
+flamegraphs), and security investigations.
+
+## Usage
+
+Gadgets making use of stack trace functionality typically disable it by default
+due to the additional overhead. When enabled, stack traces can be collected
+from:
+
+- **Kernel space**: Shows the kernel call stack when an event occurred. Enabled
+  with `--collect-kstack`.
+- **User space**: Shows the application call stack when an event occurred.
+  Enabled with `--collect-ustack`.
+
+The stack traces contain the addresses of the functions in the call stack. To
+convert these addresses into human-readable function names, symbolization is
+performed. Inspektor Gadget supports different symbolization methods, including
+client-side and server-side symbolization.
+
+- **Kallsyms**: Uses the kernel's kallsyms mechanism to resolve kernel addresses
+  (server-side only). This is always enabled for kernel stack traces.
+- **Symtab**: Uses the user space application's symbol table to resolve user
+  addresses (server-side only). This is enabled by default. You can use
+  `--symbolizers=symtab` to be explicit.
+- **Debuginfod cache**: Uses the debuginfod cache to resolve user addresses
+  (server-side or client-side) from build ids and offsets. This requires
+  `--collect-build-id`. This is disabled by default for performance reasons. Use
+  `--symbolizers=debuginfod-cache` to enable it on the client or on the
+  standalone ig. Use `--symbolizers=debuginfod-cache-on-ig-server` to enable it
+  on the server only.
+
+### Examples
+
+
+```bash
+kubectl gadget run trace_capabilities:%IG_TAG% \
+    --collect-kstack \
+    --collect-ustack \
+    --collect-build-id \
+    --symbolizers symtab,debuginfod-cache \
+    -o yaml
+```
+
+### Output Format
+
+When stack traces are enabled, the output includes additional columns or fields showing the stack traces. For example:
+
+```
+ustack:
+  addresses: '[0]0x00007ff9dd7734fb; [1]0x00007ff9dd6763b8; [2]0x00007ff9dd67647b;
+    [3]0x00005617712fa2d5; '
+  buildid: '[0]91f01b4ad171c80b6303d08d1f08cba8b990413d +1274fb; [1]91f01b4ad171c80b6303d08d1f08cba8b990413d
+    +2a3b8; [2]91f01b4ad171c80b6303d08d1f08cba8b990413d +2a47b; [3]ae03c7096b4806aa173c0fd1861b019daf6a057a
+    +32d5; '
+  symbols: '[0]chroot; [1]__libc_start_call_main; [2]__libc_start_main_alias_1; [3]_start; '
+```
+
+The raw stack trace output consists of memory addresses. Symbolization converts
+these addresses into human-readable function names.
+
+## Architecture
+
+Stack trace collection in Inspektor Gadget works as follows:
+
+1. **Collection**: The eBPF program uses BPF helpers like `bpf_get_stack()` to
+   collect stack traces at specific trace points. Inspektor Gadget provides a
+   helper API at `gadget/kernel_stack_map.h` and `gadget/user_stack_map.h` to
+   help gadget authors. See [Stack maps in Gadget eBPF
+   API](../gadget-devel/gadget-ebpf-api.md#Stack-maps) for more information.
+2. **Transfer**: Stack traces are sent to userspace as part of the event data.
+3. **Processing**: The ebpf operator and the ustack operator process and
+   optionally symbolize the stack traces.
+
+
+## Performance Considerations
+
+Enabling stack traces increases overhead in several ways:
+
+- Additional CPU cycles for the eBPF program to collect the stack trace
+- Increased memory usage for storing stack frames
+- Greater data transfer between kernel and user space
+- CPU usage for symbolization
+
+The most impactful is the symbolization because it needs to parse the ELF file
+and load the symbol table.
+
+## Limitations
+
+### Stack depth
+
+Inspektor Gadget collects up to 127 frames for kernel stack traces and 127
+frames for user stack traces.
+
+It can be a problem with Java because Java applications often have deep call
+stacks, and hitting the frame limit may result in incomplete stack traces that
+miss important context.
+
+### Missing symbols
+
+- The symtab symbolization method only works if the executable has debug symbols
+  available.
+- The debuginfod symbolization method does not download the debuginfo packages
+  itself but leaves it to the user.
+- The debuginfod symbolization method relies on the build id being available in
+  an ELF note such as `.note.gnu.build-id` or `.note.go.buildid`. When building
+  with gcc, this can be added with `-Wl,--build-id`.
+  ```bash
+  $ readelf -n /bin/cat|grep 'Build ID:'
+    Build ID: 54d277fac459da46ee1a054d1ef6a5d02a3d9346
+  ```
+
+## Guide
+
+Prepare a simple program:
+```c
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+
+__attribute__((noinline)) void level3() {
+    chroot("/");
+}
+__attribute__((noinline)) void level2() {
+    level3();
+}
+__attribute__((noinline)) void level1() {
+    level2();
+}
+
+int main() {
+    level1();
+    sleep(1);
+    return 0;
+}
+```
+
+Create a container for compiling and running the example application:
+
+```bash
+$ docker run -ti --rm --name test ghcr.io/inspektor-gadget/ci/gcc:latest
+cat > chroot.c
+: copy/paste the program, then ctrl-d
+gcc -Wall -o basic chroot.c
+gcc -Wall -static -o static chroot.c
+cp basic stripped ; strip stripped
+cp static static-stripped ; strip static-stripped
+gcc -Wall -Wl,--build-id -o stripped-buildid chroot.c ; strip stripped-buildid
+gcc -Wall -fPIE -pie -o pie chroot.c
+```
+
+Use the trace_capabilities gadget to collect the stack traces:
+
+```bash
+sudo ig run \
+    ghcr.io/inspektor-gadget/gadget/trace_capabilities:%IG_TAG% \
+    --collect-ustack \
+    --collect-build-id \
+    --symbolizers symtab,debuginfod-cache \
+    -c test --fields proc.comm,ustack.symbols
+```
+
+We can now run the various versions of the program. Observe the output:
+```
+COMM             USTACK.SYMBOLS
+static           [0]chroot; [1]level2; [2]level1; [3]main; [4]__libc_start_call_main;
+static-stripped  [0][unknown]; [1][unknown]; [2][unknown]; [3][unknown]; [4][unknown];
+```
+
+When the binary is not stripped of symbol table, the symbtab symbolizer works
+correctly. The stripped binary does not have symbols, so the symtab symbolizer
+cannot resolve the addresses.
+
+Some Linux distributions strip the binaries by default, but provide ways to
+retrieve the symbols, either with debuginfo packages or by using [public
+debuginfod database](https://sourceware.org/elfutils/Debuginfod.html). The
+debuginfod database is indexed by build id, which is a unique identifier for the
+binary generated at compile time with `gcc -Wl,--build-id`. This is what is
+suggested by the warnings:
+
+```
+WARN[0084] Debuginfo def5460e3cee00bfee25b429c97bcc4853e5b3a8 for test/stripped-buildi not found in /root/.cache/debuginfod_client/def5460e3cee00bfee25b429c97bcc4853e5b3a8/debuginfo. Suggested remedial: "DEBUGINFOD_CACHE_PATH=/root/.cache/debuginfod_client DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo def5460e3cee00bfee25b429c97bcc4853e5b3a8"
+WARN[0084] Debuginfo 448b87dfff9599434fba6b1086ee18447460151f for test/stripped-buildi not found in /root/.cache/debuginfod_client/448b87dfff9599434fba6b1086ee18447460151f/debuginfo. Suggested remedial: "DEBUGINFOD_CACHE_PATH=/root/.cache/debuginfod_client DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo 448b87dfff9599434fba6b1086ee18447460151f"
+stripped-buildi  [0][unknown]; [1][unknown]; [2][unknown]; [3][unknown]; [4][unknown];
+```
+
+Let's use another container to demonstrate this:
+```bash
+$ docker run -ti --rm --name test ubuntu:24.10
+# chroot / chroot / id
+```
+
+At first, Inspektor Gadget will not be able to resolve the symbols:
+```
+WARN[0282] Debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d for test/chroot not found in /root/.cache/debuginfod_client/91f01b4ad171c80b6303d08d1f08cba8b990413d/debuginfo. Suggested remedial: "DEBUGINFOD_CACHE_PATH=/root/.cache/debuginfod_client DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d"
+WARN[0282] Debuginfo ae03c7096b4806aa173c0fd1861b019daf6a057a for test/chroot not found in /root/.cache/debuginfod_client/ae03c7096b4806aa173c0fd1861b019daf6a057a/debuginfo. Suggested remedial: "DEBUGINFOD_CACHE_PATH=/root/.cache/debuginfod_client DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo ae03c7096b4806aa173c0fd1861b019daf6a057a"
+chroot           [0][unknown]; [1][unknown]; [2][unknown]; [3][unknown];
+chroot           [0][unknown]; [1][unknown]; [2][unknown]; [3][unknown];
+```
+
+Ubuntu 24.10 provides debuginfod packages on the federated server
+https://debuginfod.elfutils.org, so we can run the suggested commands as root:
+```bash
+sudo DEBUGINFOD_CACHE_PATH=/root/.cache/debuginfod_client DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d
+sudo DEBUGINFOD_CACHE_PATH=/root/.cache/debuginfod_client DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo ae03c7096b4806aa173c0fd1861b019daf6a057a
+```
+
+Now, Inspektor Gadget can resolve the symbols even though the running executable is stripped:
+```
+COMM             USTACK.SYMBOLS
+chroot           [0]chroot; [1]__libc_start_call_main; [2]__libc_start_main_alias_1; [3]_start;
+chroot           [0]chroot; [1]__libc_start_call_main; [2]__libc_start_main_alias_1; [3]_start;
+```
+
+The same can apply for dynamic libraries. Even though `basic` was built without
+build id, it uses libc which was built with build id. So after using the
+debuginfod-find command to retrieve the missing debuginfo files, we can see the
+libc symbols for `basic`:
+
+```
+COMM             USTACK.SYMBOLS
+basic            [0]__GI_chroot; [1]level2; [2]level1; [3]main; [4]__libc_start_call_main;
+```
+
+Inspektor Gadget also supports Position-Independent Executables (PIE), so you
+can also see symbols for `pie`:
+
+```
+COMM             USTACK.SYMBOLS
+pie              [0]__GI_chroot; [1]level2; [2]level1; [3]main; [4]__libc_start_call_main;
+```
+
+We can also collect the kernel stack with `--collect-kstack`:
+```bash
+sudo ig run \
+    ghcr.io/inspektor-gadget/gadget/trace_capabilities:%IG_TAG% \
+    --collect-kstack \
+    -c test --fields proc.comm,kstack
+```
+Here is the output:
+```
+COMM             KSTACK
+chroot           [0]security_capable; [1]ns_capable; [2]__x64_sys_chroot; [3]do_syscall_64; [4]entry_SYSCALL_64_after_hwframe;
+```


mauriciovasquezbernal:
This looks amazing, now the thing is how to implement the client part. Have you envisioned how it should work? Do you think this operator will have the time to process the data online (i.e. resolving the symbols when they arrive) or should we enable a way to record all the information to a file and them enrich it offline?

alban:
> This looks amazing, now the thing is how to implement the client part. Have you envisioned how it should work? Do you think this operator will have the time to process the data online (i.e. resolving the symbols when they arrive) or should we enable a way to record all the information to a file and them enrich it offline?
> 
> Or what other ideas do you have to implement it client side?

I don't know... Maybe it is easier for Inspektor Gadget Desktop because it can download the debuginfo asynchronously and update the flamegraph when the data becomes available.

It is more difficult for ig/gadgetctl in a terminal because it can't update previous events it has already printed.

It might be ok for some gadgets like trace_mount to do it online because it does not generate a lot of events. But gadgets like cpu profilers should do the symbolization at least asynchronously or offline.

alban:
I pushed a new commit that let the symbolizer reuse the cache from debuginfod cache when available.

Example of command:
```
sudo ig run ghcr.io/inspektor-gadget/gadget/trace_capabilities:alban_build_id --verify-image=false --collect-ustack --collect-build-id --symbolizers symtab,debuginfod-client-cache --comm=chroot --debuginfod-client-cache-path ""
```

Notice that `--symbolizers debuginfod-client-cache` is not enabled by default.

Exampe of workload:
```
docker run -ti --rm ubuntu:24.10
chroot / chroot / chroot / id
```
Initial output:
```
WARN[0005] Debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d for suspicious_brattain/chroot not found in /root/.cache/debuginfod_client/91f01b4ad171c80b6303d08d1f08cba8b990413d/debuginfo. Suggested remedial: "sudo DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d"
```
Run the suggested command to populate the cache. You can also use other
servers as listed on https://sourceware.org/elfutils/Debuginfod.html.
After that, ig will report:
```yaml
ustack:
  addresses: '[0]0x00007efd894424fb; [1]0x00007efd893453b8; [2]0x00007efd8934547b;
    [3]0x00005649dec0e2d5; '
  buildid: '[0]91f01b4ad171c80b6303d08d1f08cba8b990413d +1274fb; [1]91f01b4ad171c80b6303d08d1f08cba8b990413d
    +2a3b8; [2]91f01b4ad171c80b6303d08d1f08cba8b990413d +2a47b; [3]ae03c7096b4806aa173c0fd1861b019daf6a057a
    +32d5; '
  symbols: '[0]__GI_chroot; [1]__libc_start_call_main; [2]call_init; [3]_start; '
```
If you want to remove the cache, run:
```
sudo rm -rf /root/.cache/debuginfod_client/91f01b4ad171c80b6303d08d1f08cba8b990413d
```

alban:
I pushed another commit, and now it can be used on Kubernetes with kubectl-gadget. The ustack operator can enrich the stack client-side.

Users can use one or several of the following options:

* `--symbolizers debuginfod-client-cache`: use debuginfod client cache (`~/.cache/debuginfod_client`) on the client (machine where `kubectl-gadget` or `ig run...` is running)
* `--symbolizers debuginfod-client-cache-on-ig-server` use debuginfod client cache (`/root/.cache/debuginfod_client`) on the server (machine where `ig daemon` or `gadgettracermanager` is running)
* `--symbolizers symtab`: use the symbol table in the executable on the server (machine where `ig daemon` or `gadgettracermanager` is running)

Example of command:
```
kubectl-gadget run albantest.azurecr.io/trace_capabilities:alban_build_id \
        --collect-ustack --collect-build-id \
        --symbolizers debuginfod-client-cache \
        --comm=chroot \
        --debuginfod-client-cache-path "/home/alban/.cache/debuginfod_client" \
        -o yaml
```
Workload on Kubernetes:
```
kubectl run -ti --rm --image ubuntu:24.10 --restart=Never shell01 -- bash
chroot / chroot / chroot / id
```

At first, kubectl-gadget gives the following output:
```
WARN[0004] Debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d for shell01/chroot not found in /home/alban/.cache/debuginfod_client/91f01b4ad171c80b6303d08d1f08cba8b990413d/debuginfo. Suggested remedial: "sudo DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d"
WARN[0004] Debuginfo ae03c7096b4806aa173c0fd1861b019daf6a057a for shell01/chroot not found in /home/alban/.cache/debuginfod_client/ae03c7096b4806aa173c0fd1861b019daf6a057a/debuginfo. Suggested remedial: "sudo DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo ae03c7096b4806aa173c0fd1861b019daf6a057a"
```

As regular user, on the client, run:
```
DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo ae03c7096b4806aa173c0fd1861b019daf6a057a
DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo 91f01b4ad171c80b6303d08d1f08cba8b990413d
```

Now, kubectl-gadget can give the correct output:
```
ustack:
  addresses: '[0]0x00007ff9dd7734fb; [1]0x00007ff9dd6763b8; [2]0x00007ff9dd67647b;
    [3]0x00005617712fa2d5; '
  buildid: '[0]91f01b4ad171c80b6303d08d1f08cba8b990413d +1274fb; [1]91f01b4ad171c80b6303d08d1f08cba8b990413d
    +2a3b8; [2]91f01b4ad171c80b6303d08d1f08cba8b990413d +2a47b; [3]ae03c7096b4806aa173c0fd1861b019daf6a057a
    +32d5; '
  symbols: '[0]chroot; [1]__libc_start_call_main; [2]__libc_start_main_alias_1; [3]_start; '
```

pkg/operators/ustack/ustack.go
```diff
@@ -281,11 +407,23 @@ func (o *OperatorInstance) Name() string {
...
+		// TODO: detect if we're client-side without the ebpf operator
+		// gadgetCtx.IsRemoteCall() is not sufficient
+		o.clientSide = true
+		for ds, funcs := range o.subscriptions {
+			for _, f := range funcs {
+				ds.Subscribe(f, Priority)
+			}
+		}
+
+		return nil
```
alban: TODO: Once https://github.com/inspektor-gadget/inspektor-gadget/pull/4816 is merged, we can rewrite this change in a cleaner way.

pkg/operators/ebpf/ebpf.go
```diff
@@ -700,6 +700,24 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
 		}
 	}
 
+	for _, m := range i.collectionSpec.Maps {
+		if m.MaxEntries == 0 {
```
mauriciovasquezbernal: Do we only want to consider maps with 0 as max entries? What about if we do the other way around, checking for the existing of the `maxEntriesStr` variable and from it set MaxEntries regardless its value.

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal, line 1: We need to document, and hopefully test it. (referring to first commit)

include/gadget/user_stack_map.h
mauriciovasquezbernal, line 1: nit: I think changes to this file on commit "symbolizer: use debuginfod cache when available" should go to "ustack: add buildid field for client-side symbolization".

pkg/symbolizer/symbolizer.go
```diff
@@ -339,3 +425,70 @@ func (s *Symbolizer) newSymbolTable(pid uint32, expectedExeKey exeKey) (*symbolT
...
+		if err != nil {
+			return err
+		}
+
+		s.lockSymbolTablesFromBuildID.Lock()
+
+		s.symbolTablesFromBuildID[buildIDStr] = table
+
+		table.timestamp = time.Now()
+		symbol := table.lookupByAddr(stackItems[i].Offset)
```
mauriciovasquezbernal: stackItems[i] -> query? (same in other places)

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,252 @@
...
+  reasons. Use `--symbolizers=debuginfod-client-cache` to enable it on the
+  client or on the standalone ig. Use
+  `--symbolizers=debuginfod-client-cache-on-ig-server` to enable it on the
+  server only.
+
+### Examples
+
+
+```bash
+kubectl-gadget run trace_capabilities \
```
mauriciovasquezbernal: kubectl-gadget -> kubectl gadget

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,252 @@
...
+
+The raw stack trace output consists of memory addresses. Symbolization converts
+these addresses into human-readable function names.
+
+## Architecture
+
+Stack trace collection in Inspektor Gadget works as follows:
+
+1. **Collection**: The eBPF program uses BPF helpers like `bpf_get_stack()` to
+   collect stack traces at specific trace points. Inspektor Gadget provides an
```
mauriciovasquezbernal: an -> a

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,252 @@
...
+cp static static-stripped ; strip static-stripped
+gcc -Wall -Wl,--build-id -o stripped-buildid chroot.c ; strip stripped-buildid
+gcc -Wall -fPIE -pie -o pie chroot.c
+```
+
+Use the trace_capabilities gadget to collect the stack traces:
+
+```bash
+sudo ig run \
+    ghcr.io/inspektor-gadget/gadget/trace_capabilities:alban_build_id \
```
mauriciovasquezbernal: alban_build_id > %IG_TAG%

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,252 @@
...
+    --verify-image=false \
```
mauriciovasquezbernal: remove

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,252 @@
...
+}
+
+int main() {
+    level1();
+    sleep(1);
+    return 0;
+}
+```
+
+Compile it in a container with various options:
```
mauriciovasquezbernal: I was thinking the container was used only to compile it, perhaps say something like: "Create a container for compiling and running the example application".

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,252 @@
...
+```
+COMM             USTACK.SYMBOLS
+pie              [0]__GI_chroot; [1]level2; [2]level1; [3]main; [4]__libc_start_call_main;
+```
+
+We can also collect the kernel stack with `--collect-kstack`:
+```bash
+sudo ig run \
+    ghcr.io/inspektor-gadget/gadget/trace_capabilities:alban_build_id \
+    --verify-image=false \
```
mauriciovasquezbernal: ditto.

include/gadget/user_stack_map.h
```diff
@@ -24,6 +24,23 @@ struct {
...
+const volatile int ig_build_id_max_entries = 1024;
+GADGET_PARAM(ig_build_id_max_entries);
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(key_size, sizeof(u32));
+	__uint(value_size,
+	       GADGET_USER_MAX_STACK_DEPTH * sizeof(struct bpf_stack_build_id));
+	__uint(max_entries, 0); // To be replaced at runtime
+} ig_build_id SEC(".maps");
```
mauriciovasquezbernal, line 40: Shouldn't it be called `gadget_`...?

pkg/operators/ustack/ustack.go
```diff
@@ -88,11 +109,17 @@ func (o *Operator) InstantiateDataOperator(gadgetCtx operators.GadgetContext, in
...
 	// When the Symbolizer implements more options, they can be added here
-	if opts.UseSymtab {
-		instance.symbolizer, err = symbolizer.NewSymbolizer(opts)
-		if err != nil {
-			return nil, err
-		}
+	if opts.UseSymtab || opts.UseDebugInfodClientCache {
+		// Use a sync.OnceValue to delay the creation of the Symbolizer because:
+		// - otherwise it is needlessly created during GetGadgetInfo
+		instance.symbolizer = sync.OnceValue(func() *symbolizer.Symbolizer {
```
mauriciovasquezbernal: nit: Can we use OnceValues to also return the error to be able to log it on the callers?

pkg/operators/ustack/ustack.go
```diff
@@ -55,31 +59,48 @@ func (o *Operator) GlobalParams() api.Params {
...
 	}
 
-	opts := symbolizer.SymbolizerOptions{}
+	opts := symbolizer.SymbolizerOptions{
+		DebuginfodCachePath: instanceParamValues[debuginfodCachePathParam],
+	}
 	symbolizers := instanceParamValues[symbolizersParam]
 	switch symbolizers {
 	case "", "none":
 		opts.UseSymtab = false
```
mauriciovasquezbernal: It's not part of this commit but I suppose we can remove it.

pkg/operators/ustack/ustack.go
```diff
@@ -173,13 +201,36 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
 
-			addressesField, err := in.AddSubField("addresses", api.Kind_String, datasource.WithFlags(datasource.FieldFlagHidden))
-			if err != nil {
-				return err
+			// The ustack operator can run both on the client and on the server side.
+			// If it runs on the client side, the server might have already added the subfields.
+			var addressesField, buildIDField, symbolsField datasource.FieldAccessor
+			var err error
+
+			if addressesFieldAll := in.GetSubFieldsWithTag("name:addresses"); len(addressesFieldAll) == 0 {
```
mauriciovasquezbernal: Should we somehow check this field is the specific one added by the ustack operator and not any other `addresses` field a gadget author could add?

pkg/operators/ustack/ustack.go
```diff
@@ -229,25 +280,91 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+					if addressesStr != "" {
+						addressesField.PutString(data, addressesStr)
+					}
+					if buildIDStr != "" {
+						buildIDField.PutString(data, buildIDStr)
+					}
+					stackItems = append(stackItems, moreStackItems...)
+					alreadyKnownSymbols = make([]string, len(stackItems))
+				} else {
+					// The symbolizer might be used client-side where we don't
```
mauriciovasquezbernal, line 333: Do we need to execute this else branch if the symbolizer is nil?

pkg/operators/ustack/ustack.go
```diff
@@ -229,25 +280,91 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+						_, err = fmt.Sscanf(buildidList[i], "[%d]%s +%x", &idx, &buildidStr, &offset)
+						if err != nil {
+							_, _ = fmt.Sscanf(buildidList[i], "[%d]%x", &idx, &ip)
+							// It's ok if we don't have a build ID.
+						} else {
+							validBuildID = true
+						}
+						buildidSlice, err := hex.DecodeString(buildidStr)
+						if err != nil {
+							break
```
mauriciovasquezbernal, line 381: warn?

pkg/operators/ustack/ustack.go
```diff
@@ -229,25 +280,91 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+						}
+						// It's ok if we don't have a build ID. But if we have one, it should be valid.
+						if len(buildidSlice) != 20 && len(buildidSlice) != 0 {
+							break
```
mauriciovasquezbernal, line 386: same here.

pkg/operators/ustack/ustack.go
```diff
@@ -229,25 +280,91 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
 				mtimeNsec, _ := mtimeNsecField[0].Uint32(data)
 
-				stack := [ebpftypes.UserPerfMaxStackDepth]uint64{}
-				err := o.userStackMap.Lookup(stackId, &stack)
-				if err != nil {
-					logger.Warnf("stack with ID %d is lost: %s", stackId, err.Error())
-					return nil
-				}
+				stackItems := make([]symbolizer.StackItemQuery, 0, ebpftypes.UserPerfMaxStackDepth)
+				var alreadyKnownSymbols []string
```
mauriciovasquezbernal: IIUC these are symbols resolved server-side? if that's the case, can you add a comment? At the beginning I was thinking it was a sort of symbols cache.

pkg/operators/ustack/ustack.go
```diff
@@ -229,25 +280,91 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+				} else {
+					// The symbolizer might be used client-side where we don't
+					// have access to BPF maps. Access data from the data source
+					// instead.
+					addressesStr, _ := addressesField.String(data)
+					addressesList := strings.Split(addressesStr, "; ")
+					buildIDStr, _ := buildIDField.String(data)
+					buildidList := strings.Split(buildIDStr, "; ")
+					alreadyKnownSymbolsStr, _ := symbolsField.String(data)
+					alreadyKnownSymbols = strings.Split(alreadyKnownSymbolsStr, "; ")
```
mauriciovasquezbernal, line 341: I find all this string manipulation a bit heavy. Is there a way to avoid it? Like carrying the raw data in a different field to avoid this back and forth conversion.

pkg/symbolizer/symbolizer.go
```diff
@@ -187,109 +230,76 @@ type Task struct {
...
-func (s *Symbolizer) Resolve(task Task, addresses []uint64) ([]string, error) {
-	if s.options.UseSymtab {
-		return s.resolveWithSymtab(task, addresses)
-	}
-	// At the moment, we don't support other symbolization methods than symtab
-	// and NewSymbolizer() is only called with UseSymtab. So this code path is
-	// not taken. When implementing more symbolization methods, this can be
-	// added here.
-	return nil, nil
+// StackItemQuery is one item of the stack. It contain the data found from BPF:
```
mauriciovasquezbernal:
```suggestion
// StackItemQuery is one item of the stack. It contains the data found from BPF:
```

pkg/symbolizer/symbolizer.go
```diff
@@ -187,109 +230,76 @@ type Task struct {
...
-	if newKey != expectedExeKey {
-		newComm, _ := os.Readlink(fmt.Sprintf("/proc/self/fd/%d", file.Fd()))
-		newComm = filepath.Base(newComm)
-		return nil, fmt.Errorf("opening executable: got %q inode %d, mtime %d.%d (expected %s)",
-			newComm, ino, stat.Mtim.Sec, stat.Mtim.Nsec,
-			expectedExeKey)
-	}
-	if fs.Size() > maxExecutableSize {
-		return nil, fmt.Errorf("executable is too large (%d bytes)", fs.Size())
+func (s *Symbolizer) resolveStackItemsWithTable(table *symbolTable, baseAddress uint64, stackItems []StackItemQuery, res []StackItemResponse) {
```
mauriciovasquezbernal: Move to symtab_linux?

mauriciovasquezbernal (commented):
I did a first pass to get familiar with the code. I'll make a second one checking the UX.

pkg/operators/ustack/ustack.go
alban: I'll think about it.
Otherwise, another solution would be arrays of structs but they are not yet supported in our datasources: #4060.

pkg/operators/ebpf/ebpf.go
alban: Good point. Also, some maps legitimately have max_entries=0:

```
struct {
	__uint(key_size, sizeof(__u32));
	__uint(value_size, sizeof(__u32));
} events SEC(".maps");
```
```
$ sudo ig run trace_dns
WARN[0003] Replacing map "events" max_entries: variable "events_max_entries" not found
```

pkg/operators/ebpf/ebpf.go
alban: I am pondering removing this commit 2bcb1c50ba916e63a284e2e4bc5047befc2609df for now, or move it to another PR.

This was added for reducing memory consumption when the map `ig_build_id` is not used (by default we have `--collect-build-id=false`).

```
struct {
	__uint(value_size,
	       GADGET_USER_MAX_STACK_DEPTH * sizeof(struct bpf_stack_build_id));
	__uint(max_entries, 0); // To be replaced at runtime
} ig_build_id SEC(".maps");
```

memory = max_entries * (key_size + value_size) = 1024 * (4 + 127 * 32) = 4165632 = 3.9 MiB

We're wasting 3.9 MiB by default because the map `ig_build_id` is not used by default.

But the PR is not enough because setting `--collect-build-id=true` does not automatically change the other parameter to `--ig-build-id-max-entries=1024`, so we need additional ideas.

#3485 might help.

include/gadget/user_stack_map.h
alban: Why? Other maps in include files are named ig_ustack and ig_kstack.

I see we also have `gadget_mntns_filter_map` and `gadget_sockets`...

I prefer shorter names because `bpftool map` will print truncated names.

pkg/operators/ustack/ustack.go
alban: Alternatively, we could implement some method on the gadgetContext to figure out if we're in a GetGadgetInfo context. Similar idea to #4816. That would enable earlier errors.

pkg/operators/ustack/ustack.go
alban: I can check for that with a second tag "operator:ustack" and warn & skip when the tag is not here.

pkg/operators/ustack/ustack.go
alban: I think not.

pkg/symbolizer/symbolizer.go
alban: This method was added in a first commit and modified in a second commit. Do you know a tool to move it easily and keep the PR decomposed in logical commits? Otherwise, I'd need to run an interactive rebase, edit both commits, and resolve the conflicts.

alban: I did it manually with these commands:
```
git rebase -i b710bb9f9200cf36222dbe978ca8e039c66816c4
# Add "edit" for the relevant commits

Iterate over each commit to edit:
    If conflict:
        git restore -s CURRENT_COMMIT_ID .

    sed -i 's/stackItems/stackQueries/g' $(git ls-files '*.go')

    If no conflict:
        git commit -a --amend --no-edit
    If conflict:
        git commit -a --no-edit

    git rebase --continue
```

pkg/symbolizer/symbolizer.go
alban: I did it manually with these commands:
```
git rebase -i b710bb9f9200cf36222dbe978ca8e039c66816c4
# Add "edit" for the relevant commits

Iterate over each commit to edit:
    If conflict:
        git restore -s CURRENT_COMMIT_ID .

    sed -i 's/stackItems/stackQueries/g' $(git ls-files '*.go')

    If no conflict:
        git commit -a --amend --no-edit
    If conflict:
        git commit -a --no-edit

    git rebase --continue
```

include/gadget/user_stack_map.h
mauriciovasquezbernal: It's a mess right now:
- [gadget_sockets](https://github.com/inspektor-gadget/inspektor-gadget/blob/1eb27fd2bea98f7cc5e68e316e3f46e12ecdc311/include/gadget/sockets-map.h#L61)
- [gadget_mntns_filter_map](https://github.com/inspektor-gadget/inspektor-gadget/blob/04c60a7421824cded611598b25f714a3ca63e845/include/gadget/mntns_filter.h#L19)
- [gadget_heap](https://github.com/inspektor-gadget/inspektor-gadget/blob/1eb27fd2bea98f7cc5e68e316e3f46e12ecdc311/include/gadget/buffer.h#L27)
- etc.

> I prefer shorter names because bpftool map will print truncated names.

That's a good point.

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: I like that idea. It could help us in a lot of situations where we have specific logic to check for it. @flyth any opinions on this?

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: I'm wondering if wasm can help. But perhaps something on top of #3485 is simpler.

include/gadget/user_stack_map.h
```diff
@@ -24,6 +24,23 @@ struct {
...
+
+static const struct bpf_stack_build_id
+	empty_build_id[GADGET_USER_MAX_STACK_DEPTH] = {};
```
mauriciovasquezbernal: Prefix it with either `ig` or `gadget` to avoid name clashes.

pkg/operators/ustack/ustack.go
```diff
@@ -247,6 +254,48 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+						var v bpfStackBuildID
+						_ = x[unsafe.Sizeof(v)-sizeOfBpfStackBuildID]
+					}
+					buildIDBuf := [ebpftypes.UserPerfMaxStackDepth * sizeOfBpfStackBuildID]byte{}
+					err = o.buildIDMap.Lookup(stackId, &buildIDBuf)
+					if err != nil {
+						logger.Warnf("build id for stack ID %d is lost: %s", stackId, err.Error())
+						return nil
+					}
+					buildid := (*[ebpftypes.UserPerfMaxStackDepth]bpfStackBuildID)(unsafe.Pointer(&buildIDBuf[0]))
```
mauriciovasquezbernal: What about something like:

```go
					buildid := [ebpftypes.UserPerfMaxStackDepth]bpfStackBuildID{}
					err := o.buildIDMap.Lookup(stackId, &buildid)
					if err != nil {
						logger.Warnf("build id for stack ID %d is lost: %s", stackId, err.Error())
						return nil
					}
```

(requires to make the members of bpfStackBuildID public)

Or use `LookupBytes` if you really want to stick to the unsafe casting.

pkg/operators/ustack/ustack.go
```diff
@@ -247,6 +254,48 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+						buildID    [unix.BPF_BUILD_ID_SIZE]uint8
+						offsetOrIP uint64 // Union of offset and ip
+					}
+					const sizeOfBpfStackBuildID = 32
+					// Static assert that the size of bpfStackBuildID is correct
+					_ = func() {
+						var x [1]struct{}
+						var v bpfStackBuildID
+						_ = x[unsafe.Sizeof(v)-sizeOfBpfStackBuildID]
+					}
```
mauriciovasquezbernal: Why do we need this assertion? Shouldn't it be compared against the size on the bpf map instead against a constant?

pkg/operators/ustack/ustack.go
```diff
@@ -270,16 +288,22 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
 
 					var buildIDsBuilder strings.Builder
 				buildid_iter:
-					for i, b := range buildid {
+					for i := 0; i < ebpftypes.UserPerfMaxStackDepth; i++ {
+						if errLookup != nil {
+							// The gadget didn't collect build ids
+							// Gadgets can use --collect-build-id to enable collecting build ids
+							break
+						}
```
mauriciovasquezbernal: why to break inside the loop if errLookup is never updated in it?

pkg/operators/ebpf/params.go
```diff
@@ -122,6 +122,10 @@ var wellKnownParams = map[wellKnownParamKey]wellKnownParamVal{
 		Key:         "collect-ustack",
 		Description: "Collect user stack traces",
 	},
+	{TypeName: "bool", VarName: "collect_build_id"}: {
+		Key:         "collect-build-id",
+		Description: "Collect build IDs",
+	},
```
mauriciovasquezbernal, line 128: Ideally this should go to the "ustack: add buildid field for client-side symbolization" commit.

pkg/symbolizer/symbolizer.go
```diff
@@ -339,3 +425,70 @@ func (s *Symbolizer) newSymbolTable(pid uint32, expectedExeKey exeKey) (*symbolT
...
+
+		s.lockSymbolTablesFromBuildID.Lock()
+
+		s.symbolTablesFromBuildID[buildIDStr] = table
+
+		table.timestamp = time.Now()
+		symbol := table.lookupByAddr(stackQueries[i].Offset)
+		if symbol != "" {
+			res[i].Found = true
+			res[i].Symbol = symbol
```
mauriciovasquezbernal: What about if the symbol was already resolved?

pkg/symbolizer/symbolizer.go
mauriciovasquezbernal, line 1: I'm wondering if we can make this file more modular, like each symbolizer type (symtab, debufinfodclientcache, etc) being a different type that implements `Resolve` and then this file being a small one that iterates over all registered symbolizers. I think this will be difficult to maintain if it continues growing, for instance, right now it was two different mutexes and symbol tables.

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: A possible solution is to send the raw bytes over the wire and decode it client side. (It'll have to take endianess into account, but perhaps that's better than parsing this string)

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: ok, so add a check to avoid it?

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: What about using `IsClient` here to be more explicit:
- IsClient = true: the field must exist 
- IsClient = false: base it only on the ebpf map and add the fields for the client to consume them


Another idea I have is: perhaps we can split this operator into two, a server side reads the buildinfo and creates a field with it, the client side looks for that field and resolves the symbols. It's similar to what's is already done but having to operators could clarify where different parts of the code are executed. 

@flyth thoughts on this?

pkg/operators/ustack/ustack.go
```diff
@@ -55,31 +59,48 @@ func (o *Operator) GlobalParams() api.Params {
...
 		for _, s := range list {
 			switch s {
 			case "symtab":
-				opts.UseSymtab = true
+				opts.UseSymtab = !gadgetCtx.IsClient()
+			case "debuginfod-client-cache":
+				if !gadgetCtx.IsRemoteCall() {
+					opts.UseDebugInfodClientCache = true
+				}
+			case "debuginfod-client-cache-on-ig-server":
```
mauriciovasquezbernal: I find this confusing, both client and server are mentioned on the same name. Could we use `debuginfod-client-cache` and `debuginfod-server-cache` instead?

alban: "debuginfod-client-cache-on-ig-server" means that the IG server is looking up the files from the cache prepared by e.g. the debuginfod-find binary, aka the debuginfod client.

In the current PR, IG is not communicating with the debuginfod server, so naming debuginfod-server-cache could be confusing.

It is confusing but I am not sure what to suggest instead.

debuginfod-client

mauriciovasquezbernal: > "debuginfod-client-cache-on-ig-server" means that the IG server is looking up the files from the cache prepared by e.g. the debuginfod-find binary, aka the debuginfod client.

What confuses me is the "client" part. I think the fact that debuginfod-find is the client is not relevant for the user IMO. `debuginfod-cache-on-ig-server` sounds better to me, even if not perfect.

alban:
I started preliminary smaller PRs:
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/4919
- [x] https://github.com/inspektor-gadget/inspektor-gadget/pull/4920
- [ ] https://github.com/inspektor-gadget/inspektor-gadget/pull/4921

pkg/symbolizer/symbolizer.go
alban: If both symtab and debuginfod resolver worked, it probably does not matter which one we pick.

It does not usually happen: if the debuginfod method works, it means the binary comes from a Linux distribution which ships its binaries stripped and send the dbg pkg to the debuginfod database. There is no point to populate the debuginfod database if the binary is not stripped in the first place.

pkg/operators/ustack/ustack.go
alban: To avoid one level of indentation :D 
Should I rewrite it to avoid the errLookup check in the loop?

alban:
I rebased this PR on the other PRs listed in the previous comment, and addressed the reviews.

pkg/symbolizer/symbolizer.go
mauriciovasquezbernal: I think you replied to yourself on https://github.com/inspektor-gadget/inspektor-gadget/pull/4314#discussion_r2291930602.

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: Isn't it possible to add 

```go
		if errLookup != nil {
			return addressesStr, "", stackQueries, nil
		}
```

after the conditional?

pkg/operators/ustack/ustack.go
```diff
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"unsafe"
 
 	"github.com/cilium/ebpf"
+	"golang.org/x/sys/unix"
```
mauriciovasquezbernal: This should to to ustack_linux.go

pkg/operators/ustack/ustack.go
```diff
@@ -247,6 +254,46 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
...
+							status     int32
+							buildID    [unix.BPF_BUILD_ID_SIZE]uint8
+							offsetOrIP uint64 // Union of offset and ip
+						}
+						buildIDMapValueSize := o.buildIDMap().ValueSize()
+						buildIDBuf := [ebpftypes.UserPerfMaxStackDepth * unsafe.Sizeof(bpfStackBuildID{})]byte{}
+						if int(buildIDMapValueSize) != len(buildIDBuf) {
+							logger.Warnf("build id map has unexpected value size %d instead of %d", buildIDMapValueSize, len(buildIDBuf))
+							return nil
+						}
```
mauriciovasquezbernal: This is changed on the "ustack operator: add client-side support (gadgetctl, kubectl-gadget)" commit. Why not to squash "ustack: add buildid field for client-side symbolization", "symbolizer: move common functions to new file" & "ustack operator: add client-side support (gadgetctl, kubectl-gadget)"?

pkg/operators/ustack/ustack_linux.go
```diff
@@ -47,6 +49,59 @@ func readUserStackMap(gadgetCtx operators.GadgetContext, userStackMap *ebpf.Map,
...
 
-	return addressesStr, stackQueries, nil
+	// The buildIDMap is optional. Older gadgets won't have it.
+	if buildIDMap != nil && buildIDMap.MaxEntries() > 0 {
+		// struct bpf_stack_build_id is part of Linux UAPI:
+		// https://github.com/torvalds/linux/blob/v6.14/include/uapi/linux/bpf.h#L1451
+		type bpfStackBuildID struct {
+			status     int32
+			buildID    [unix.BPF_BUILD_ID_SIZE]uint8
+			offsetOrIP uint64 // Union of offset and ip
```
mauriciovasquezbernal: These fields need to be exported, otherwise it panicks:


```bash
panic: reflect: reflect.Value.SetInt using value obtained using unexported field

goroutine 21387 [running]:
reflect.flag.mustBeAssignableSlow(0xc0003bcd60?)
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/reflect/value.go:254 +0xb4
reflect.flag.mustBeAssignable(...)
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/reflect/value.go:244
reflect.Value.SetInt({0x3a422a0?, 0xc002b22000?, 0x4635080?}, 0x1)
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/reflect/value.go:2132 +0x45
encoding/binary.(*decoder).value(0xc0003bd238, {0x3a422a0?, 0xc002b22000?, 0x41?})
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/encoding/binary/binary.go:883 +0x13f
encoding/binary.(*decoder).value(0xc0003bd238, {0x3e2faa0?, 0xc002b22000?, 0xc0003bd1b0?})
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/encoding/binary/binary.go:863 +0x7e5
encoding/binary.(*decoder).value(0xc0003bd238, {0x3ace7e0?, 0xc002b22000?, 0xc0003bd268?})
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/encoding/binary/binary.go:850 +0x8cc
encoding/binary.Decode({0xc002b23000, 0xfe0, 0xfe0}, {0x45f8190, 0x63c0940}, {0x39ddb80, 0xc002b22000})
        /home/mauriciov/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.0.linux-amd64/src/encoding/binary/binary.go:326 +0x345
github.com/cilium/ebpf/internal/sysenc.Unmarshal({0x39ddb80, 0xc002b22000}, {0xc002b23000, 0xfe0, 0xfe0})
        /home/mauriciov/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250729164112-d994daa25101/internal/sysenc/marshal.go:91 +0x1cf
github.com/cilium/ebpf/internal/sysenc.Buffer.Unmarshal({0xc002b23000?, 0x39ddb80?}, {0x39ddb80?, 0xc002b22000?})
        /home/mauriciov/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250729164112-d994daa25101/internal/sysenc/buffer.go:75 +0x4a
github.com/cilium/ebpf.(*Map).unmarshalValue(0xc0023648c0?, {0x39ddb80?, 0xc002b22000?}, {0xc002b23000?, 0x0?})
        /home/mauriciov/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250729164112-d994daa25101/map.go:1631 +0x365
github.com/cilium/ebpf.(*Map).LookupWithFlags(0xc0023648c0, {0x3a422e0, 0xc002b00160}, {0x39ddb80, 0xc002b22000}, 0x0)
        /home/mauriciov/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250729164112-d994daa25101/map.go:770 +0xe5
github.com/cilium/ebpf.(*Map).Lookup(...)
        /home/mauriciov/go/pkg/mod/github.com/cilium/ebpf@v0.19.1-0.20250729164112-d994daa25101/map.go:747
github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ustack.readUserStackMap({0x4616ce8?, 0xc000884000?}, 0xc002364870, 0xc0023648c0, 0x3317)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/operators/ustack/ustack_linux.go:70 +0x458
github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ustack.(*OperatorInstance).init.func1({0x4629668?, 0xc0004cc9c0?}, {0x45cc020, 0xc002b12050})
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/operators/ustack/ustack.go:301 +0x795
github.com/inspektor-gadget/inspektor-gadget/pkg/datasource.(*dataSource).Subscribe.func1({0x0?, 0x0?}, {0x45cc048?, 0xc002b12050?})
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/datasource/data.go:595 +0x42
github.com/inspektor-gadget/inspektor-gadget/pkg/datasource.(*dataSource).EmitAndRelease(0xc0004cc9c0, {0x45cc048, 0xc002b12050})
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/datasource/data.go:664 +0xdf
github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf.(*Tracer).processEvent(0xc00201f170, {0x45ec280?, 0xc001af11d0?}, {0xc002c23c30, 0xa8, 0xa8})
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/operators/ebpf/tracer.go:209 +0x252
github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf.(*Tracer).receiveEvents(0xc00201f170, {0x4616ce8, 0xc000884000}, 0xc003567f98?)
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/operators/ebpf/tracer.go:156 +0x25c
created by github.com/inspektor-gadget/inspektor-gadget/pkg/operators/ebpf.(*ebpfInstance).runTracer in goroutine 1
        /home/mauriciov/kinvolk/ebpf/inspektor-gadget/pkg/operators/ebpf/tracer.go:260 +0x35f
```

pkg/operators/ustack/ustack_linux.go
```diff
@@ -47,6 +49,59 @@ func readUserStackMap(gadgetCtx operators.GadgetContext, userStackMap *ebpf.Map,
...
+		}
+		buildid := [ebpftypes.UserPerfMaxStackDepth]bpfStackBuildID{}
+		if buildIDMap.ValueSize() != uint32(unsafe.Sizeof(buildid)) {
+			logger.Warnf("build id map has unexpected value size %d instead of %d",
+				buildIDMap.ValueSize(), unsafe.Sizeof(buildid))
+			return "", "", nil, nil
+		}
```
mauriciovasquezbernal: I think we can move this check to the init phase to avoid having to perform it on the hot path.

mauriciovasquezbernal (commented):
@alban what do you think of https://github.com/inspektor-gadget/inspektor-gadget/pull/4314/files#r2344519335?

pkg/symbolizer/symbolizer.go
alban: This makes sense. And also if done in different packages, it would allow clients like gadgetctl and kubectl to skip the compilation of symtab_linux.go, even if they are compiled for Linux.

alban:
I updated the branch and addressed most of the comments, but I still need to address https://github.com/inspektor-gadget/inspektor-gadget/pull/4314/files#r2344519335 and test again.

I'll do that after #4921 is merged.

alban:
Updated with interfaces.

pkg/operators/ustack/ustack.go
```diff
@@ -189,13 +210,46 @@ func (o *OperatorInstance) init(gadgetCtx operators.GadgetContext) error {
 				continue
 			}
 
-			addressesField, err := in.AddSubField("addresses", api.Kind_String, datasource.WithFlags(datasource.FieldFlagHidden))
-			if err != nil {
-				return err
+			// The ustack operator can run both on the client and on the server side.
```
mauriciovasquezbernal, line 216: nit: perhaps a small helper could reduce code duplication:

```go
			// The ustack operator can run both on the client and on the server side.
			// If it runs on the client side, the server might have already added the subfields.
			addField := func(name string) (datasource.FieldAccessor, error) {
				field := in.GetSubFieldsWithTag("name:" + name)
				if len(field) == 0 {
					return in.AddSubField(name, api.Kind_String,
						datasource.WithFlags(datasource.FieldFlagHidden),
						datasource.WithTags("name:"+name, "operator:ustack"))
				}
				if !field[0].HasAllTagsOf("operator:ustack") {
					logger.Warn("field " + name + " exists but does not belong to the ustack operator")
					return nil, fmt.Errorf("field %q exists but does not belong to the ustack operator", name)
				}

				return field[0], nil
			}

			addressesField, err := addField("addresses")
			if err != nil {
				return err
			}
			buildIDField, err := addField("buildid")
			if err != nil {
				return err
			}
			symbolsField, err := addField("symbols")
			if err != nil {
				return err
			}
```

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: This is still unhandled.

pkg/operators/ustack/ustack.go
mauriciovasquezbernal: ditto.

pkg/symbolizer/debuginfod/debuginfod.go
```diff
@@ -0,0 +1,168 @@
...
+			buildIDRemovedCount++
+			buildIDSymbolRemovedCount += len(table.Symbols)
+		}
+	}
+	if buildIDRemovedCount > 0 {
+		log.Debugf("symbol tables from build ID pruned: %d symbol tables with %d symbols removed (remaining: %d symbol tables)",
+			buildIDRemovedCount, buildIDSymbolRemovedCount,
+			len(d.symbolTablesFromBuildID))
+	}
+	d.lockSymbolTablesFromBuildID.Unlock()
```
mauriciovasquezbernal: nit: better to use defer to avoid deadlocks later on

pkg/symbolizer/debuginfod/debuginfod.go
```diff
@@ -0,0 +1,168 @@
...
+					suggestedCmd := fmt.Sprintf("DEBUGINFOD_CACHE_PATH=%s DEBUGINFOD_URLS=https://debuginfod.elfutils.org debuginfod-find debuginfo %s",
+						d.options.DebuginfodCachePath, buildIDStr)
+					log.Warnf("Debuginfo %s for %s not found in %s. Suggested remedial: %q", buildIDStr, task.Name, debuginfoPath, suggestedCmd)
+				}
+				continue
+			}
+			log.Warnf("Failed to open debuginfo file %s for %s: %v", debuginfoPath, task.Name, err)
+			continue
+		}
+		defer file.Close()
```
mauriciovasquezbernal: is ok to hold a lot of open files inside the loop until the function returns?

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,251 @@
...
+  `--collect-build-id`. This is disabled by default for performance reasons. Use
+  `--symbolizers=debuginfod-cache` to enable it on the client or on the
+  standalone ig. Use `--symbolizers=debuginfod-cache-on-ig-server` to enable it
+  on the server only.
+
+### Examples
+
+
+```bash
+kubectl gadget run trace_capabilities \
```
mauriciovasquezbernal:
```suggestion
kubectl gadget run trace_capabilities:%IG_TAG% \
```

docs/reference/stack-traces.mdx
```diff
@@ -0,0 +1,251 @@
...
+The raw stack trace output consists of memory addresses. Symbolization converts
+these addresses into human-readable function names.
+
+## Architecture
+
+Stack trace collection in Inspektor Gadget works as follows:
+
+1. **Collection**: The eBPF program uses BPF helpers like `bpf_get_stack()` to
+   collect stack traces at specific trace points. Inspektor Gadget provides a
+   helper API at `gadget/kernel_stack_map.h` and `<gadget/user_stack_map.h>` to
```
mauriciovasquezbernal:
```suggestion
   helper API at `gadget/kernel_stack_map.h` and `gadget/user_stack_map.h` to
```

(or they other way around to keep them aligned)

pkg/symbolizer/debuginfod/debuginfod.go
mauriciovasquezbernal: Perhaps all of the logic above can be moved to a wrapper around NewSymbolTableFromFile that accepts the path of the file.

mauriciovasquezbernal (approved):
Some minor nits but LGTM already, great job @alban!

PR #4329, opened 2025-04-14
ustack: generic collect-ustack param and templates

alban:
* Move GADGET_PARAM(collect_ustack) in the header file so it does not have to be repeated in all gadgets' program.bpf.c.
* Make it a well-known parameter so it does not have to be specified in gadget.yaml.
* Make use of templates so gadget.yaml does not have to repeat the descriptions of the ustack fields.

## How to use

* Gadget authors no longer have to explicitly add `GADGET_PARAM(collect_ustack)` in their source code.
* Gadget authors no longer have to add boilerplate in gadget.yaml for collect_ustack param and ustack fields

## Testing done

```
$ sudo ./ig run ghcr.io/inspektor-gadget/gadget/trace_capabilities:alban_collect_ustack_param --verify-image=false --help
...
      --collect-ustack                         Collect user stack traces
...
...
                                                   ustack
                                                     User stack trace
                                                   ustack.addresses
                                                   ustack.exe_inode
                                                   ustack.mtime_nsec
                                                   ustack.mtime_sec
                                                   ustack.pid_level0
                                                   ustack.pid_level1
                                                   ustack.pidns_level0
                                                   ustack.pidns_level1
                                                   ustack.stack_id
                                                   ustack.symbols
```

## TODO

This PR does not add a template for the subfields `ustack.addresses` and `ustack.symbols` but that's already cleaner this way.


pkg/operators/ebpf/types/types.go
```diff
@@ -40,6 +40,7 @@ const (
 	ProcessTypeName     = "gadget_process"
 	CredsTypeName       = "gadget_creds"
 	ParentTypeName      = "gadget_parent"
+	UstackTypeName      = "gadget_user_stack"
```
mauriciovasquezbernal, line 43: We already have `UserStackTypeName`.

mauriciovasquezbernal (approved):
Besides a duplicated type name it LGTM!  Approving to same another review iteration.

I got some questions not related to this PR:
1. Any specific reason to set all fields to 0 when collect_ustack = false:
https://github.com/inspektor-gadget/inspektor-gadget/blob/20eb2d9a18946995fa467a92e7b8bbcdf60d8a3e/include/gadget/user_stack_map.h#L127-L135

I see only exe_inode being used to detect if ustack is enabled:

https://github.com/inspektor-gadget/inspektor-gadget/blob/4c2576e6cf086223a3be4d6fe92f357d0db5f0a0/pkg/operators/ustack/ustack.go#L186-L190

2. Related to the above: Is it possible to use the value of the parameter to enable / disable ustack enrichment in pkg/operators/ustack/ustack.go instead on relying on the presence of a `gadget_user_stack` member. My question is because I'd like to avoid as much overhead as possible when this is disabled.

alban:
> Any specific reason to set all fields to 0 when collect_ustack = false:

That should work too.
That makes a difference in one scenario: if the ustack operator is not used (e.g. a Go application using IG packages but without linking to the ustack operator), then those fields will be displayed in json/yaml output without initialization, so it makes contains garbage values from previous events. This possibly leaks unwanted information; this might or might not matter, depending on the application.

> Related to the above: Is it possible to use the value of the parameter to enable / disable ustack enrichment in pkg/operators/ustack/ustack.go instead on relying on the presence of a gadget_user_stack member. My question is because I'd like to avoid as much overhead as possible when this is disabled.

That should be possible but this change a couple of things:
- Gadgets ebpf code can no longer decide to selectively enable the stack for some events only. I currently don't see a use case for this though.
- The "collect_ustack" parameter is now automatically added when including "user_stack_map.h" but it is not part of the Gadget API. A gadget might choose to not rely on "user_stack_map.h" and implement the ebpf code itself. But if the operator relies on the "collect_ustack" parameter, that becomes part of the Gadget API. So if we do that, it should be documented in the Gadget API.

In follow-up PRs, I wanted to experiment reducing the size of the `struct event` when collect_ustack is not enabled. I'll keep the discussion above in mind to avoid as much overhead as possible when ustack is disabled.

PR #4331, opened 2025-04-14
Introduce `value.one-of` annotation.

mqasimsarfraz:
Currently, we don't have a way to specify a list of possible values for a field in the metadata annotations. This makes filtering `--filter` harder since user can't know the format/possible values for the fields. This PR introduces a new annotation `value.one-of` which can be used to specify a list of possible values for a field.

This change is also useful for other frontends like Headlamp and ig-desktop which can use this information to provide better filtering experience to the users.

## Testing done

```
$ ig run trace_dns --verify-image=false -h
...
                                           "dns" (data source):
                                             addresses
                                             cwd
                                               The current working directory of the process (require --paths flag)
                                             ...
                                             qr
                                               Specifies whether this message is a query (Q), or a response (R).
                                               One of: Q, R
                                             qr_raw
                                             qtype
                                               Query type
                                               One of: A, NS, CNAME, SOA, PTR, MX, TXT, AAAA, SRV, OPT, WKS, HINFO, MINFO, AXFR, ALL
                                             qtype_raw
                                             ra
                                               Recursion Available Flag
                                             rcode
                                               Response code
                                               One of: Success, FormatError, ServerFailure, NameError, NotImplemented, Refused
                                             rcode_raw

...
```


blanquicet (approved):
LGTM from code inspection

mqasimsarfraz:
Thanks for the review!

PR #4333, opened 2025-04-15
inspect: adds ebpf relocations in extraInfo

KapilSareen:
## Changes

This PR aims to add ebpf relocations info in the `image inspect` subcommand.

Part of https://github.com/inspektor-gadget/inspektor-gadget/issues/3387

### DIFF
--- a/pkg/operators/ebpf/extrainfo.go
+++ b/pkg/operators/ebpf/extrainfo.go
@@ -19,6 +19,10 @@ import (
 	"debug/elf"
 	"encoding/json"
 	"fmt"
+	"strings"
+
+	"github.com/cilium/ebpf/asm"
+	"github.com/cilium/ebpf/btf"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
@@ -30,8 +34,9 @@ type extraInfoMap struct {
 }
 
 type extraInfoProgram struct {
-	Section string
-	Source  string
+	Section              string
+	Source               string
+	SourceWithRelocation string
 }
 
 type extraInfoVariable struct {
@@ -72,8 +77,9 @@ func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
 	// Add programs
 	for _, p := range i.collectionSpec.Programs {
 		programs = append(programs, &extraInfoProgram{
-			Section: p.SectionName,
-			Source:  p.Instructions.String(),
+			Section:              p.SectionName,
+			Source:               p.Instructions.String(),
+			SourceWithRelocation: getSourceWithRelocations(p.Instructions),
 		})
 	}
 	programsJson, _ := json.Marshal(programs)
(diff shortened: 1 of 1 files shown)


pkg/operators/ebpf/extrainfo.go
```diff
@@ -108,8 +118,22 @@ func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
...
 	return nil
 }
+
+func GetRelocations(insns asm.Instructions) {
+	iter := insns.Iterate()
+	for iter.Next() {
+		relo := btf.CORERelocationMetadata(iter.Ins)
+		if relo != nil {
+			fmt.Println("Relocation:", relo, "; Line: ", iter.Ins.Source())
+		}
```
alban, line 137: We should print lines without relocation too, for context. Something like:

```suggestion
		source := iter.Ins.Source()
		if source != "" {
			fmt.Printf("; %s\n", source)
		}
		relo := btf.CORERelocationMetadata(iter.Ins)
		if relo != nil {
			fmt.Println("; %v\n", relo)
		}
		fmt.Printf("%v\n", iter.Ins)
```

pkg/operators/ebpf/extrainfo.go
```diff
@@ -108,8 +123,30 @@ func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
...
+		ContentType: "application/json",
+		Content:     []byte(relocations),
+	}
 
 	gadgetCtx.SetVar("extraInfo.ebpf", ebpfInfo)
 
 	return nil
 }
+
+func GetRelocations(insns asm.Instructions) (relocations string) {
```
alban, line 136: * No need to export this function
* Maybe it could be renamed to `getSourceWithRelocations()` since it gets more than just the relocations

pkg/operators/ebpf/extrainfo.go
```diff
@@ -41,6 +45,13 @@ type extraInfoVariable struct {
 	Map    string
 }
 
+type relocation struct {
+	Offset  uint64
+	Type    uint32
+	Symbol  string
+	Section string
```
alban, line 52: This type does not seem used anymore

alban (commented):
This seems good. I will try it soon. Meanwhile, here are some comments.

pkg/operators/ebpf/extrainfo.go
KapilSareen: Yes, sorry my bad.

pkg/operators/ebpf/extrainfo.go
```diff
@@ -69,12 +73,16 @@ func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
...
-	// Add programs
+	// Add programs and relocations
+	var relocations string
 	for _, p := range i.collectionSpec.Programs {
 		programs = append(programs, &extraInfoProgram{
 			Section: p.SectionName,
 			Source:  p.Instructions.String(),
 		})
+		r := getSourceWithRelocations(p.Instructions)
+		relocations += r
```
alban, line 84: I wonder if relocations should be part of the existing "ebpf.programs". The json struct "extraInfoProgram" could have a 3rd field "SourceWithRelocation string".

KapilSareen: hmm, makes sense to me, if we're ignoring the UX part.

pkg/operators/ebpf/extrainfo.go
```diff
@@ -113,3 +119,21 @@ func (i *ebpfInstance) addExtraInfo(gadgetCtx operators.GadgetContext) error {
...
+	iter := insns.Iterate()
+	for iter.Next() {
+		source := iter.Ins.Source()
+		if source != nil {
+			relocs.WriteString(fmt.Sprintf("; %s\n", source))
+		}
+		relo := btf.CORERelocationMetadata(iter.Ins)
+		if relo != nil {
+			relocs.WriteString(fmt.Sprintf("; %v\n", relo))
+		}
```
blanquicet, line 135: nit:

```suggestion
		if source := iter.Ins.Source(); source != nil {
			relocs.WriteString(fmt.Sprintf("; %s\n", source))
		}
		if relo := btf.CORERelocationMetadata(iter.Ins); relo != nil {
			relocs.WriteString(fmt.Sprintf("; %v\n", relo))
		}
```

blanquicet (approved):
Let's see what Alban thinks, but LGTM!

alban (approved):
LGTM!

PR #4325, opened 2025-04-11
docs: Fix website build

mqasimsarfraz:
`{}` in BPF_MAP_TYPE_{HASH,ARRAY}_OF_MAPS was making the website build to fail. I put backticks around it to fix the build.

Was hitting this issue on my PR: https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/14404543472/job/40397739857

## Testing Done

CI is [green](https://github.com/inspektor-gadget/inspektor-gadget/actions/runs/14405204745/job/40399967243?pr=4325).

PR #4334, opened 2025-04-15
ci: Move the website build check higher

mqasimsarfraz:
Currently if we introduce a new documentation page we don't check if it will break the webiste build because the link check step fails with 404 for that page. This change reorders the checks to ensure we always check if a page will break website build to avoid issues like: https://github.com/inspektor-gadget/inspektor-gadget/pull/4325

Ref: https://github.com/inspektor-gadget/inspektor-gadget/pull/4325#issuecomment-2797328660


mqasimsarfraz:
I am putting this PR on hold till we merge: https://github.com/inspektor-gadget/inspektor-gadget/pull/4338

PR #4337, opened 2025-04-15
wasmapi: improve error messages

alban:
Before this patch, a gadget could fail with this error:
```
ERRO[0003] failed to get param value: error getting param value
Error running application: running gadget: starting operators: starting operator "oci": pre-starting operator "wasm": gadgetPreStart failed
```

The wasmapi has a lot of error messages that don't give much context. This patch adds a bit more context in the error messages.


mauriciovasquezbernal:
@alban  are you still interested on this PR? If yes, would you mind rebasing and marking as ready for review?

PR #4338, opened 2025-04-16
ci: Use UmbrellaDocs/action-linkspector instead of deprecated gaurav-…

eiffel-fl:
…nelson/github-action-markdown-link-check.

The latter was deprecated in April 2025 and author recommends using UmbrellaDocs/action-linkspector instead [1].

[1]: https://github.com/gaurav-nelson/github-action-markdown-link-check/commit/0524e79d8d7d

### DIFF
--- a/docs/design/001-prometheus.md
+++ b/docs/design/001-prometheus.md
@@ -433,7 +433,7 @@ and filling them in their eBPF programs
 This gadget creates the eBPF maps on behalf of users when they specify they want a counter. We will
 need to be sure those maps stick to the contract defined above. The syntax to defining counters,
 histograms, etc by using the DSL can be very similar to what we already have in bpftrace
-[https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#2-count-count](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#2-count-count),
+[https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#2-count-count](https://github.com/iovisor/bpftrace/blob/master/do/reference_guide.md#2-count-count),
 [https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#8-hist-log2-histogram](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#8-hist-log2-histogram).
 
 # Out of Scope
--- a/docs/reference/verify-assets.mdx
+++ b/docs/reference/verify-assets.mdx
@@ -10,7 +10,7 @@ import TabItem from '@theme/TabItem';
 Inspektor Gadget container image and release assets are signed using
 [`cosign`](https://github.com/sigstore/cosign).
 In this guide, we will see how you can verify them with this tool.
-Note that, You would need to have `cosign` [v2.0](https://github.com/sigstore/cosign/blob/main/README.md#developer-installation) installed.
+Note that, You would need to have `cosign` [v2.0](https://github.com/sigstre/cosign/blob/main/README.md#developer-installation) installed.
 
 ## Verify the container image manually
 


mauriciovasquezbernal:
> For now, there is a problem with using several extensions, which we need as we have md and mdx files: [UmbrellaDocs/linkspector#124](https://github.com/UmbrellaDocs/linkspector/pull/124)

AFAIU it's only an issue when dirs is also used. Is it an issue for us? if so, what should we do with this PR?

mauriciovasquezbernal:
> We can either wait https://github.com/UmbrellaDocs/linkspector/pull/124 to be merged, or we can use https://github.com/tcort/github-action-markdown-link-check.
Though, the second is claimed to be maintained, but it has the same latest commit as upstream.

Thanks for the clarification. IMO we can wait for that PR to be merged.

burak-ok:
> ... or we can use https://github.com/tcort/github-action-markdown-link-check. Though, the second is claimed to be maintained, but it has the same latest commit as upstream.

It seems like the repo has started to become alive. First changes which just mentions that its currently just the same as upstream version, but they'll pick up/transfer issues there: https://github.com/tcort/github-action-markdown-link-check/pull/1

.github/workflows/inspektor-gadget.yml
```diff
@@ -54,12 +54,11 @@ jobs:
...
     - name: Check that there are not broken links
-      uses: gaurav-nelson/github-action-markdown-link-check@3c3b66f1f7d0900e37b71eca45b63ea9eedfce31 # v1.0.17
+      uses: umbrelladocs/action-linkspector@v1.3.5
       with:
-        config-file: .github/workflows/mlc_config.json
-        use-quiet-mode: 'yes'
-        check-modified-files-only: 'yes'
-        base-branch: 'main'
+        config_file: .github/workflows/linkspector.yml
+        reporter: github-pr-review
```
mauriciovasquezbernal, line 60: Why is it different from the one on the config file?

.github/workflows/inspektor-gadget.yml
```diff
@@ -54,12 +54,11 @@ jobs:
         make generate-documentation
         git diff --exit-code HEAD --
     - name: Check that there are not broken links
-      uses: gaurav-nelson/github-action-markdown-link-check@3c3b66f1f7d0900e37b71eca45b63ea9eedfce31 # v1.0.17
+      uses: umbrelladocs/action-linkspector@v1.3.5
```
mauriciovasquezbernal, line 57: Please stick it to the hash.

.github/workflows/inspektor-gadget.yml
eiffel-fl: We need to add .mdx.

.github/workflows/inspektor-gadget.yml
eiffel-fl: I am still prototyping, hence the fixed version and no hash.

eiffel-fl:
> I see it prints a weird error message:
> 
> ```
> message:"Cannot reach https://github.com/iovisor/bpftrace/blob/master/do/reference_guide.md#2-count-count Status: 404" location:{path:"docs/design/001-prometheus.md" range:{start:{line:436 column:1} end:{line:436 column:173}}} severity:ERROR source:{name:"linkspector" url:"https://github.com/UmbrellaDocs/linkspector"}
>   time=2025-06-13T06:57:37.327Z level=INFO msg="reviewdog: failed to post a review comment: POST https://api.github.com/repos/inspektor-gadget/inspektor-gadget/pulls/4338/reviews: 403 Resource not accessible by integration []"
>   reviewdog: This GitHub Token doesn't have write permission of Review API [1],
>   so reviewdog will report results via logging command [2] and create annotations similar to
>   github-pr-check reporter as a fallback.
>   [1]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
>   [2]: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions
>   reviewdog: found at least one issue with severity greater than or equal to the given level: any
> ```
> 
> Do you know how to remove it? It think it's a bit misleading and doesn't immediately tell that something is going on with the link.

This error is normal behavior.
Basically, it would like to write comments such as us in the PR to highlight the link errors, but it lacks the necessary authorization to do so.
I am not sure if we should give this authorization, I just would like the tool to write the error in the CI output, remaining seems over-engineered.

Anyway, for now, I still cannot have it detect the error in the .mdx file, which is a real problem for us.

PR #4366, opened 2025-04-23
docs: Add documentation for exposing metrics on K8s

### LINKED ISSUE #4007: docs: Document exporting metrics when using kubectl-gadget
The docs for kubectl-gadget are missing on https://www.inspektor-gadget.io/docs/latest/reference/export-metrics. Headless mode was merged on #3587.

mauriciovasquezbernal:
Fixes #4007

### Discussion 

Should we have deeper documentation on how to install and configure prometheus to scrap the metrics?

### DIFF
--- a/docs/reference/export-metrics.mdx
+++ b/docs/reference/export-metrics.mdx
@@ -21,7 +21,7 @@ In order to enable the metrics listener, you need to set the
 <Tabs groupId="env">
     <TabItem value="kubectl-gadget" label="kubectl gadget">
         ```bash
-        WIP: Headless mode for kubectl gadget is under development
+        kubectl gadget deploy --otel-metrics-listen=true
         ```
     </TabItem>
 
@@ -69,13 +69,18 @@ Following is an example of how to enable metrics export for the
 <Tabs groupId="env">
     <TabItem value="kubectl-gadget" label="kubectl gadget">
         ```bash
-        WIP: Headless mode for kubectl gadget is under development
+        kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_blockio:%IG_TAG% \
+                    --name profileblockio \
+                    --annotate=blockio:metrics.collect=true \
+                    --otel-metrics-name=blockio:blockio-metrics \
+                    --detach
         ```
     </TabItem>
 
     <TabItem value="ig" label="ig">
         ```bash
         gadgetctl run ghcr.io/inspektor-gadget/gadget/profile_blockio:%IG_TAG% \
+                    --name profileblockio \
                     --annotate=blockio:metrics.collect=true \
                     --otel-metrics-name=blockio:blockio-metrics \
                     --detach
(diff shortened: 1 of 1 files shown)


docs/reference/export-metrics.mdx
```diff
@@ -21,7 +21,7 @@ In order to enable the metrics listener, you need to set the
 <Tabs groupId="env">
     <TabItem value="kubectl-gadget" label="kubectl gadget">
         ```bash
-        WIP: Headless mode for kubectl gadget is under development
+        kubectl gadget deploy --otel-metrics-listen=true
```
blanquicet, line 24: Don't we need to specify the address and port as the default conflicts with old Prometheus implementation?

blanquicet: I just tried and it worked, even if when running the gadget it prints the following error:

```
$ kubectl gadget run ghcr.io/inspektor-gadget/gadget/profile_blockio:v0.39.0 \
                    --name profileblockio \
                    --annotate=blockio:metrics.collect=true \
                    --otel-metrics-name=blockio:blockio-metrics \
                    --detach
ERRO[0000] serving otel metrics on: listen tcp 0.0.0.0:2224: bind: address already in use
INFO[0005] installed as "a97ab8a80c2e31bb1d8d0b347b4a42ed"
```

blanquicet: This is not related with this PR, so let's go ahead.

blanquicet (approved):
LGMT!

docs/reference/export-metrics.mdx
mauriciovasquezbernal: > Don't we need to specify the address and port as the default conflicts with old Prometheus implementation?

The old Prometheus implementation [uses 2223](https://github.com/inspektor-gadget/inspektor-gadget/blob/1b1c95331753a32a340583f8accdd79380669f29/pkg/operators/prometheus/prometheus.go#L43) by default, and it's only enabled when running `ig prometheus` so there is not any collision there. 

One issue is that we're hardcoding the [prometheus annotations](https://github.com/inspektor-gadget/inspektor-gadget/blob/8d78571d8b12d79ac606b3c82418dbff18475d34/pkg/resources/manifests/deploy.yaml#L149-L151) on the deployment manifest, so prometheus will try to scrape metrics on port 2223 instead of the value passed with `otel-metrics-listen-address`. I'll open an issue or PR for it.

> ERRO[0000] serving otel metrics on: listen tcp 0.0.0.0:2224: bind: address already in use

I don't get that error when running the gadget.

blanquicet: > One issue is that we're hardcoding the [prometheus annotations](https://github.com/inspektor-gadget/inspektor-gadget/blob/8d78571d8b12d79ac606b3c82418dbff18475d34/pkg/resources/manifests/deploy.yaml#L149-L151) on the deployment manifest, so prometheus will try to scrape metrics on port 2223 instead of the value passed with `otel-metrics-listen-address`. I'll open an issue or PR for it.
> 
It means that following that guide, the metrics will not be available on Prometheus. Am I right?

mauriciovasquezbernal: Ref #4371

mauriciovasquezbernal:
> Should we have deeper documentation on how to install and configure prometheus to scrap the metrics?

Opened #4372.

PR #4373, opened 2025-04-24
inspect: improves utility and command ux

KapilSareen:
## Changes

Currently, the output of `image inspect` command is very large and difficult to directly comprehend.
This PR adds improved formatting options to the command and improves the UX.

## How to use

```bash
$ sudo ./ig image inspect trace_dns:latest --extra-info="ebpf.programs" --jsonpath="[0].Section" 
[
  "socket1"
]

$ sudo ./ig image inspect trace_dns:latest --extra-info="ebpf.sections" --jsonpath="[2]" 
[
  ".text"
]

$ sudo ./ig image inspect trace_dns:latest --extra-info="ebpf.flowchart" --jsonpath="[1]"
Error: extra info "ebpf.flowchart" is not JSON
```

Part of #3387

### DIFF
--- a/cmd/common/image/inspect.go
+++ b/cmd/common/image/inspect.go
@@ -20,8 +20,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/PaesslerAG/jsonpath"
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v2"
+	"sigs.k8s.io/yaml"
 
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common"
 	"github.com/inspektor-gadget/inspektor-gadget/cmd/common/utils"
@@ -40,7 +41,7 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 
 	opGlobalParams := make(map[string]*params.Params)
 
-	outputModes := []string{utils.OutputModeYAML, utils.OutputModeJSON, utils.OutputModeJSONPretty, utils.OutputModeCustom}
+	outputModes := []string{utils.OutputModeYAML, utils.OutputModeJSON, utils.OutputModeJSONPretty}
 
 	cmd := &cobra.Command{
 		Use:          "inspect",
@@ -50,6 +51,7 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 	}
 
 	cmd.PersistentFlags().String("extra-info", "", "Custom info type to display")
+	cmd.PersistentFlags().String("jsonpath", "", "JSONPath to extract from the extra info")
 
 	ociParams := apihelpers.ToParamDescs(ocihandler.OciHandler.InstanceParams()).ToParams()
 
@@ -107,35 +109,110 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 				"content":     string(v.Content),
 			}
 		}
-		switch outputMode {
-		case utils.OutputModeJSON:
-			bytes, err := json.Marshal(extraInfoMap)
-			if err != nil {
-				return fmt.Errorf("marshalling image and extra info to JSON: %w", err)
-			}
-			fmt.Fprint(cmd.OutOrStdout(), string(bytes))
-		case utils.OutputModeJSONPretty:
-			bytes, err := json.MarshalIndent(extraInfoMap, "", "  ")
-			if err != nil {
-				return fmt.Errorf("marshalling image and extra info to JSON: %w", err)
-			}
-			fmt.Fprint(cmd.OutOrStdout(), string(bytes))
-		case utils.OutputModeYAML:
-			bytes, err := yaml.Marshal(extraInfoMap)
-			if err != nil {
-				return fmt.Errorf("marshalling image and extra info to YAML: %w", err)
-			}
-			fmt.Fprint(cmd.OutOrStdout(), string(bytes))
-		case utils.OutputModeCustom:
-			extraInfo, _ := cmd.PersistentFlags().GetString("extra-info")
-			if extraInfo == "" {
-				return fmt.Errorf("extra info not specified (see --extra-info)")
-			}
-			if info.ExtraInfo.Data[extraInfo] == nil {
+
+		var customResult interface{}
+		extraInfo, _ := cmd.PersistentFlags().GetString("extra-info")
+		jsonPath, _ := cmd.PersistentFlags().GetString("jsonpath")
+		if extraInfo != "" {
+			dataEntry, ok := info.ExtraInfo.Data[extraInfo]
+			if !ok {
 				return fmt.Errorf("extra info %q not found", extraInfo)
 			}
-			customInfo := string(info.ExtraInfo.Data[extraInfo].Content)
-			fmt.Fprint(cmd.OutOrStdout(), customInfo)
+
+			switch dataEntry.ContentType {
+			case "application/json":
+				if err := json.Unmarshal(dataEntry.Content, &customResult); err != nil {
+					return fmt.Errorf("unmarshalling JSON content: %w", err)
+				}
+			case "text/yaml":
+				var jsonCompatible map[string]interface{}
+
+				jsonBytes, err := yaml.YAMLToJSON(dataEntry.Content)
+				if err != nil {
+					return fmt.Errorf("converting YAML to JSON: %w", err)
+				}
+
+				if err := json.Unmarshal(jsonBytes, &jsonCompatible); err != nil {
+					return fmt.Errorf("unmarshalling JSON content: %w", err)
+				}
+
+				customResult = jsonCompatible
+			default:
+				customResult = string(dataEntry.Content)
+			}
+			if jsonPath != "" {
+				if dataEntry.ContentType != "application/json" && dataEntry.ContentType != "text/yaml" {
+					return fmt.Errorf("jsonpath %q can only be used with JSON or YAML content", jsonPath)
+				}
+
+				customResult, err = jsonpath.Get(fmt.Sprintf("$%s", jsonPath), customResult)
+				if err != nil {
+					return fmt.Errorf("resolving path %q: %w", jsonPath, err)
+				}
+				if customResult == nil {
+					return fmt.Errorf("path %q not found in extra info %q", jsonPath, extraInfo)
+				}
+			}
+		}
+		if jsonPath != "" && extraInfo == "" {
+			return fmt.Errorf("jsonpath %q can only be used with extra info", jsonPath)
+		}
+
+		switch outputMode {
+		case utils.OutputModeJSON:
+			if customResult != nil {
+				if _, ok := customResult.(string); !ok {
+					bytes, err := json.Marshal(customResult)
+					if err != nil {
+						return fmt.Errorf("marshalling image and extra info to JSON: %w", err)
+					}
+					fmt.Fprint(cmd.OutOrStdout(), string(bytes), "\n")
+				} else {
+					fmt.Fprint(cmd.OutOrStdout(), customResult, "\n")
+				}
+			} else {
+				bytes, err := json.Marshal(extraInfoMap)
+				if err != nil {
+					return fmt.Errorf("marshalling image and extra info to JSON: %w", err)
+				}
+				fmt.Fprint(cmd.OutOrStdout(), string(bytes), "\n")
+			}
+		case utils.OutputModeJSONPretty:
+			if customResult != nil {
+				if _, ok := customResult.(string); !ok {
+					bytes, err := json.MarshalIndent(customResult, "", "  ")
+					if err != nil {
+						return fmt.Errorf("marshalling image and extra info to JSON (pretty): %w", err)
+					}
+					fmt.Fprint(cmd.OutOrStdout(), string(bytes), "\n")
+				} else {
+					fmt.Fprint(cmd.OutOrStdout(), customResult, "\n")
+				}
+			} else {
+				bytes, err := json.MarshalIndent(extraInfoMap, "", "  ")
+				if err != nil {
+					return fmt.Errorf("marshalling image and extra info to JSON (pretty): %w", err)
+				}
+				fmt.Fprint(cmd.OutOrStdout(), string(bytes), "\n")
+			}
+		case utils.OutputModeYAML:
+			if customResult != nil {
+				if _, ok := customResult.(string); !ok {
+					bytes, err := yaml.Marshal(customResult)
+					if err != nil {
+						return fmt.Errorf("marshalling image and extra info to YAML: %w", err)
+					}
+					fmt.Fprint(cmd.OutOrStdout(), string(bytes))
+				} else {
+					fmt.Fprint(cmd.OutOrStdout(), customResult, "\n")
+				}
+			} else {
+				bytes, err := yaml.Marshal(extraInfoMap)
+				if err != nil {
+					return fmt.Errorf("marshalling image and extra info to YAML: %w", err)
+				}
+				fmt.Fprint(cmd.OutOrStdout(), string(bytes))
+			}
 		default:
 			return fmt.Errorf("invalid output mode %q, valid values are: %s", outputMode, strings.Join(outputModes, ", "))
 		}
--- a/docs/reference/images.md
+++ b/docs/reference/images.md
@@ -288,8 +288,9 @@ Usage:
 
 Flags:
   -h, --help            help for inspect
-  -o, --output string   Output mode: json, jsonpretty, yaml, or custom (default "jsonpretty")
-  --extra-info string   In custom mode, specify particular info required
+  -o, --output string   Output mode: json, jsonpretty, yaml (default "jsonpretty")
+  --extra-info string   specify particular info required
+  --jsonpath   string   JSONPath to extract from the extra info
 ```
 
 ```bash
@@ -313,13 +314,12 @@ $ sudo ig image inspect ghcr.io/inspektor-gadget/gadget/trace_tcp
 
 # List possible extra-info
 $ sudo ig image inspect fsnotify:main |jq keys
-WARN[0001] image signature verification is disabled due to using corresponding option 
 [
+  "ebpf.flowchart",
   "ebpf.maps",
-  "ebpf.mermaid.flowchart",
-  "ebpf.mermaid.sequence",
   "ebpf.programs",
   "ebpf.sections",
+  "ebpf.sequence",
   "ebpf.variables",
   "oci.created",
   "oci.digest",
@@ -331,25 +331,27 @@ WARN[0001] image signature verification is disabled due to using corresponding o
   "wasm.upcalls"
 ]
 
-# Retrieve specific info using custom output
-$ sudo ig image inspect -o custom --extra-info=ebpf.sections ghcr.io/inspektor-gadget/gadget/trace_tcp
+# Retrieve specific info using --extra-info
+$ sudo ig image inspect --extra-info=ebpf.sections ghcr.io/inspektor-gadget/gadget/trace_tcp -o json
 ["",".strtab",".text","kprobe/tcp_v4_connect",".relkprobe/tcp_v4_connect","kretprobe/tcp_v4_connect",".relkretprobe/tcp_v4_connect","kprobe/tcp_v6_connect",".relkprobe/tcp_v6_connect","kretprobe/tcp_v6_connect",".relkretprobe/tcp_v6_connect","kprobe/tcp_close",".relkprobe/tcp_close","kprobe/tcp_set_state",".relkprobe/tcp_set_state","kretprobe/inet_csk_accept",".relkretprobe/inet_csk_accept",".rodata",".bss","license",".maps",".BTF",".rel.BTF",".BTF.ext",".rel.BTF.ext",".llvm_addrsig",".symtab"]
 
 # Listing programs
-$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs|jq -r '.[].Section'
-WARN[0001] image signature verification is disabled due to using corresponding option 
-kprobe/fanotify_handle_event
-kprobe/fsnotify_remove_first_event
-kretprobe/fsnotify_remove_first_event
-kprobe/inotify_handle_inode_event
-kretprobe/fanotify_handle_event
-kprobe/fsnotify_insert_event
-kprobe/fsnotify_destroy_event
-kretprobe/inotify_handle_inode_event
+$ sudo ig image inspect fsnotify --extra-info ebpf.programs --jsonpath='[*].Section'
+[
+  "kprobe/fsnotify_insert_event",
+  "kretprobe/inotify_handle_inode_event",
+  "kretprobe/fanotify_handle_event",
+  "kretprobe/fsnotify_remove_first_event",
+  "kretprobe/inotify_handle_event",
+  "kprobe/fanotify_handle_event",
+  "kprobe/fsnotify_destroy_event",
+  "kprobe/fsnotify_remove_first_event",
+  "kprobe/inotify_handle_inode_event",
+  "kprobe/inotify_handle_event"
+]
 
 # Printing the source of a program
-$ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs|jq -r '.[] | select(.Section=="kprobe/fsnotify_remove_first_event") | .Source'
-WARN[0001] image signature verification is disabled due to using corresponding option 
+$ sudo ig image inspect fsnotify:latest --extra-info ebpf.programs|jq -r '.[] | select(.Section=="kprobe/fsnotify_remove_first_event") | .Source'
 ig_fa_pick_e:
 	  ; int BPF_KPROBE(ig_fa_pick_e, struct fsnotify_group *group)
 	 0: LdXMemDW dst: r6 src: r1 off: 112 imm: 0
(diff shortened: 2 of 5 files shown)


cmd/common/image/inspect.go
```diff
@@ -127,14 +128,38 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
...
+			// The regex matches a string that starts with a valid field name followed by a dot and another valid field name
+			re := regexp.MustCompile(`^([a-zA-Z_][a-zA-Z0-9_]*\.[a-zA-Z_][a-zA-Z0-9_]*)(.*)$`)
+
 			extraInfo, _ := cmd.PersistentFlags().GetString("extra-info")
 			if extraInfo == "" {
 				return fmt.Errorf("extra info not specified (see --extra-info)")
 			}
-			if info.ExtraInfo.Data[extraInfo] == nil {
+			matches := re.FindStringSubmatch(extraInfo)
+			if info.ExtraInfo.Data[matches[1]] == nil {
```
alban, line 139: FindStringSubmatch can be `nil` if extraInfo does not match the regex. Then, `matches[1]` would panic.

alban:
Instead of reimplementing a JSONPath filter, we should find a Go library that can do that.

We could use different parameters for clarity:
```diff
---extra-info="ebpf.programs[0].Section"
+--extra-info="ebpf.programs" --jsonpath=".[0].Section"
```

Then, in a future PR, we can also implement go-templates as an alternative to JSONPath:

```diff
---extra-info="ebpf.programs[0].Section"
+--extra-info="ebpf.programs" --go-template="{{range .}}{{printf "%s\n" .Section}}{{end}}"
```

Depending on the desired output, users would be able to choose JSONPath or go-templates.

kubectl supports both:
1. [kubectl JSONPath Support](https://kubernetes.io/docs/reference/kubectl/jsonpath/)
2. [kubectl go-templates](https://tty4.dev/development/kubernetes-go-templates/)

See also:
* [RFC 9535 JSONPath: Query Expressions for JSON](https://www.rfc-editor.org/rfc/rfc9535.html)
* [JSONPath Online Evaluator](https://jsonpath.com/)

go.mod
```diff
@@ -224,6 +225,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.5.0 // indirect
 )
 
+require github.com/PaesslerAG/gval v1.0.0 // indirect
```
alban, line 228: This require could be moved with the other ones above:
```diff
--- a/go.mod
+++ b/go.mod
@@ -92,6 +92,7 @@ require (
        github.com/Masterminds/sprig/v3 v3.3.0 // indirect
        github.com/Microsoft/go-winio v0.6.2 // indirect
        github.com/Microsoft/hcsshim v0.12.9 // indirect
+       github.com/PaesslerAG/gval v1.0.0 // indirect
        github.com/PaesslerAG/jsonpath v0.1.1
        github.com/beorn7/perks v1.0.1 // indirect
        github.com/blang/semver/v4 v4.0.0 // indirect
@@ -225,8 +226,6 @@ require (
        sigs.k8s.io/structured-merge-diff/v4 v4.5.0 // indirect
 )
 
-require github.com/PaesslerAG/gval v1.0.0 // indirect
-
 // gopacket/gopacket with:
 // - https://github.com/gopacket/gopacket/pull/101
 replace github.com/gopacket/gopacket => github.com/gopacket/gopacket v1.3.2-0.20241202175635-b43272ae1eb8
```

cmd/common/image/inspect.go
```diff
@@ -127,12 +129,41 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 			}
 			fmt.Fprint(cmd.OutOrStdout(), string(bytes))
 		case utils.OutputModeCustom:
+
```
alban, line 132: extra line

cmd/common/image/inspect.go
```diff
@@ -127,12 +129,41 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
...
+				dataEntry, ok := info.ExtraInfo.Data[extraInfo]
+				if !ok {
+					return fmt.Errorf("extra info %q not found", extraInfo)
+				}
+				if dataEntry.ContentType != "application/json" {
+					return fmt.Errorf("extra info %q is not JSON", extraInfo)
+				}
+
+				// unmarshal the JSON content
+				var dataEntryContent []interface{}
```
alban, line 150: This works for json arrays but not for json maps.

With `--extra-info=oci.manifest` I get:
```
Error: unmarshalling JSON content: json: cannot unmarshal object into Go value of type []interface {}
```

I suggest to try with `[]interface{}` and if that does not work, to fallback on `map[string]interface{}`.

alban (commented):
Thanks! A few things to fix, and it should be good after.

alban (commented):
Could you update https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/reference/images.md#inspect?

I tried to find the correct `--jsonpath=` flag for the following command from the documentation but I couldn't find it:
```bash
$ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs|jq -r '.[] | select(.Section=="kprobe/fsnotify_remove_first_event") | .Source'
```

I'd like the documentation to use `--jsonpath` instead of `jq`, and to include examples both for array json documents and map json documents.

KapilSareen:
Now, the output matches the one given by jq. Also I updated the docs but this one I suppose cannot be done entirely without jq as of now because our indexing is number based instead of indexing by section names. Do we want to change that?

> I tried to find the correct `--jsonpath=` flag for the following command from the documentation but I couldn't find it:
> 
> ```shell
> $ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs|jq -r '.[] | select(.Section=="kprobe/fsnotify_remove_first_event") | .Source'
> ```
> 
> I'd like the documentation to use `--jsonpath` instead of `jq`, and to include examples both for array json documents and map json documents.

docs/reference/images.md
```diff
@@ -336,7 +337,7 @@ $ sudo ig image inspect -o custom --extra-info=ebpf.sections ghcr.io/inspektor-g
 ["",".strtab",".text","kprobe/tcp_v4_connect",".relkprobe/tcp_v4_connect","kretprobe/tcp_v4_connect",".relkretprobe/tcp_v4_connect","kprobe/tcp_v6_connect",".relkprobe/tcp_v6_connect","kretprobe/tcp_v6_connect",".relkretprobe/tcp_v6_connect","kprobe/tcp_close",".relkprobe/tcp_close","kprobe/tcp_set_state",".relkprobe/tcp_set_state","kretprobe/inet_csk_accept",".relkretprobe/inet_csk_accept",".rodata",".bss","license",".maps",".BTF",".rel.BTF",".BTF.ext",".rel.BTF.ext",".llvm_addrsig",".symtab"]
 
 # Listing programs
-$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs|jq -r '.[].Section'
+$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs --jsonpath='[*].Section'
 WARN[0001] image signature verification is disabled due to using corresponding option 
 kprobe/fanotify_handle_event
 kprobe/fsnotify_remove_first_event
```
alban, line 343: The formatting in the docs does not match what I get:
```
$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs --jsonpath='[*].Section'
["kretprobe/inotify_handle_inode_event","kprobe/fsnotify_insert_event","kretprobe/fsnotify_remove_first_event","kprobe/inotify_handle_event","kretprobe/inotify_handle_event","kprobe/fanotify_handle_event","kprobe/fsnotify_destroy_event","kprobe/inotify_handle_inode_event","kretprobe/fanotify_handle_event","kprobe/fsnotify_remove_first_event"]
```

I suggest to use `json.MarshalIndent` and then to make sure the output in the docs matches the real output.

alban (commented):
I have doubts on the `-o custom` mode: at the moment, `--extra-info` is ignored unless `-o custom` is given. And when using `--extra-info`, it is impossible to specify `json`, `jsonpretty` or `yaml` output.

Wouldn't it work better to delete the "custom" output mode and only check `extraInfo != ""` to decide whether to select a specific extra-info?
Then, it would be possible for the user to select an extra-info and a jsonpath with the output of their choice (yaml, json, jsonpretty).
```
$ sudo ig image inspect audit_seccomp:main --extra-info=oci.manifest -o jsonpretty --jsonpath='.annotations["org.opencontainers.image.description"]'
```

alban:
> > I tried to find the correct `--jsonpath=` flag for the following command from the documentation but I couldn't find it:
> > ```shell
> > $ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs|jq -r '.[] | select(.Section=="kprobe/fsnotify_remove_first_event") | .Source'
> > ```

I found the command for this (in the [doc](https://goessner.net/articles/JsonPath/)):

```
$ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs --jsonpath='[?@.Section=="kprobe/fsnotify_remove_first_event"].Source'
["ig_fa_pick_e:\n\t  ; .......26: Exit\n"]
```

But I don't know why it gives the result in an array.

So I still need `jq` to get rid of the array and to print the string without the quotes:
```
$ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs --jsonpath='[?@.Section=="kprobe/fsnotify_remove_first_event"].Source' | jq -r '.[0]'
ig_fa_pick_e:
	  ; int BPF_KPROBE(ig_fa_pick_e, struct fsnotify_group *group)
	...
	25: MovImm dst: r0 imm: 0
	26: Exit
```

KapilSareen:
> Wouldn't it work better to delete the "custom" output mode and only check `extraInfo != ""` to decide whether to select a specific extra-info?
> Then, it would be possible for the user to select an extra-info and a jsonpath with the output of their choice (yaml, json, jsonpretty).

Makes sense, let's refactor.

alban:
> @alban would it'll be alright if I add a commit in this PR itself refactoring this output mode stuff?

Sure, that seems like a good split: 1 commit for the jsonpath, 1 commit for refactoring the custom output.

> JSONPath does not allow to get an [index] out of a filtered expression:

I could find a way out of this limitation with a `--raw` parameter that removes the quotes and the array, similarly to `jq -r`:

<details>

```diff
--- a/cmd/common/image/inspect.go
+++ b/cmd/common/image/inspect.go
@@ -52,6 +52,7 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
 
        cmd.PersistentFlags().String("extra-info", "", "Custom info type to display")
        cmd.PersistentFlags().String("jsonpath", "", "JSONPath to extract from the extra info")
+       cmd.PersistentFlags().Bool("raw", false, "raw")
 
        ociParams := apihelpers.ToParamDescs(ocihandler.OciHandler.InstanceParams()).ToParams()
 
@@ -138,6 +139,7 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
                        }
 
                        jsonPath, _ := cmd.PersistentFlags().GetString("jsonpath")
+                       raw, _ := cmd.PersistentFlags().GetBool("raw")
 
                        if jsonPath != "" {
                                dataEntry := info.ExtraInfo.Data[extraInfo]
@@ -155,9 +157,20 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
                                if err != nil {
                                        return fmt.Errorf("resolving path %q: %w", jsonPath, err)
                                }
-                               bytes, err := json.Marshal(result)
-                               if err != nil {
-                                       return fmt.Errorf("marshalling JSONPath result to JSON: %w", err)
+                               var bytes []byte
+                               resultArray, okArray := result.([]interface{})
+                               resultString, okString := result.(string)
+                               if raw && okArray {
+                                       for _, entry := range resultArray {
+                                               bytes = append(bytes, []byte(fmt.Sprint(entry))...)
+                                       }
+                               } else if raw && okString {
+                                       bytes = []byte(resultString)
+                               } else {
+                                       bytes, err = json.Marshal(result)
+                                       if err != nil {
+                                               return fmt.Errorf("marshalling JSONPath result to JSON: %w", err)
+                                       }
                                }
                                fmt.Fprint(cmd.OutOrStdout(), string(bytes))
                                return nil
```

```
$ sudo ig image inspect -o custom fsnotify:latest --extra-info ebpf.programs --jsonpath='[?(@.Section=="kprobe/fsnotify_remove_first_event")].Source' --raw
ig_fa_pick_e:
	  ; int BPF_KPROBE(ig_fa_pick_e, struct fsnotify_group *group)
	 0: LdXMemDW dst: r6 src: r1 off: 112 imm: 0
	...
```

</details>

KapilSareen:
There seems to be an issue. I refactored and removed the custom output mode, but in doing so our output format broke too, here's an example:
```
$ sudo ./ig image inspect fsnotify --extra-info ebpf.programs --jsonpath='[*].Section' -o jsonpretty
{
  "ebpf.programs[*].Section": [
    "kprobe/inotify_handle_event",
    "kretprobe/fanotify_handle_event",
    "kprobe/fsnotify_destroy_event",
    "kprobe/fsnotify_insert_event",
    "kretprobe/fsnotify_remove_first_event",
    "kretprobe/inotify_handle_event",
    "kretprobe/inotify_handle_inode_event",
    "kprobe/fanotify_handle_event",
    "kprobe/fsnotify_remove_first_event",
    "kprobe/inotify_handle_inode_event"
  ]
}
```
It seems here, we'll have to wrap the output in a json object so as to be consistent with the output format. But as you mentioned earlier, we want the above command's output to be something like:
```
$ sudo ./ig image inspect fsnotify --extra-info ebpf.programs --jsonpath='[*].Section' -o jsonpretty
[
    "kprobe/inotify_handle_event",
    "kretprobe/fanotify_handle_event",
    "kprobe/fsnotify_destroy_event",
    "kprobe/fsnotify_insert_event",
    "kretprobe/fsnotify_remove_first_event",
    "kretprobe/inotify_handle_event",
    "kretprobe/inotify_handle_inode_event",
    "kprobe/fanotify_handle_event",
    "kprobe/fsnotify_remove_first_event",
    "kprobe/inotify_handle_inode_event"
  ]
```
(which is not json, same case with other fields like mermaids, metadata,etc)
So IMO the custom output command was serving these edge cases better, should we revert back?

Or we could also just abstarct the custom output mode from UI by making it default in case `--extra-info` is not null.

Thoughts? @alban @blanquicet

cmd/common/image/inspect.go
```diff
@@ -107,35 +109,69 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
...
+					return fmt.Errorf("resolving path %q: %w", jsonPath, err)
+				}
+				if result == nil {
+					return fmt.Errorf("path %q not found in extra info %q", jsonPath, extraInfo)
+				}
+
+				// overwrite extrainfo map
+				customMap := make(map[string]interface{})
+				customMap[fmt.Sprintf("%s%s", extraInfo, jsonPath)] = result
+				extraInfoMap = customMap
```
alban, line 152: > I refactored and removed the custom output mode, but in doing so our output format broke too, here's an example:
> 
> ```
> $ sudo ./ig image inspect fsnotify --extra-info ebpf.programs --jsonpath='[*].Section' -o jsonpretty
> {
>   "ebpf.programs[*].Section": [
>     "kprobe/inotify_handle_event",
> ```
> [...]

This is the code that makes the key like "ebpf.programs[*].Section". I think you can keep the variable "result", and use it in the `switch outputMode` below.

docs/reference/images.md
```diff
@@ -336,7 +337,7 @@ $ sudo ig image inspect -o custom --extra-info=ebpf.sections ghcr.io/inspektor-g
 ["",".strtab",".text","kprobe/tcp_v4_connect",".relkprobe/tcp_v4_connect","kretprobe/tcp_v4_connect",".relkretprobe/tcp_v4_connect","kprobe/tcp_v6_connect",".relkprobe/tcp_v6_connect","kretprobe/tcp_v6_connect",".relkretprobe/tcp_v6_connect","kprobe/tcp_close",".relkprobe/tcp_close","kprobe/tcp_set_state",".relkprobe/tcp_set_state","kretprobe/inet_csk_accept",".relkretprobe/inet_csk_accept",".rodata",".bss","license",".maps",".BTF",".rel.BTF",".BTF.ext",".rel.BTF.ext",".llvm_addrsig",".symtab"]
 
 # Listing programs
-$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs|jq -r '.[].Section'
+$ sudo ig image inspect -o custom fsnotify --extra-info ebpf.programs --jsonpath='[*].Section'
 WARN[0001] image signature verification is disabled due to using corresponding option 
 kprobe/fanotify_handle_event
```
alban, line 342: - `-o custom` should be removed
- The output in the documentation should be updated.

Ditto in other commands

cmd/common/image/inspect.go
```diff
@@ -107,35 +109,96 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
...
+				if err := json.Unmarshal(dataEntry.Content, &customResult); err != nil {
+					return fmt.Errorf("unmarshalling JSON content: %w", err)
+				}
+			} else {
+				customResult = string(dataEntry.Content)
+			}
+			if jsonPath != "" {
+				// unmarshal the JSON content
+				var dataEntryContent interface{}
+				if err := json.Unmarshal(dataEntry.Content, &dataEntryContent); err != nil {
```
alban, line 133: `json.Unmarshal(dataEntry.Content, ...)` is potentially done twice: line 124 and 133. The previous unmarshalling should be reused if already done.

To simplify the code, `--jsonpath` could be forbidden if the content is not json in the first place. At the moment, it errors out if I try to use `--jsonpath` on something that is not json:
```
$ sudo ig image inspect ghcr.io/inspektor-gadget/gadget/trace_tcp --extra-info 'ebpf.sequence' -o yaml --jsonpath='[*]'
Error: unmarshalling JSON content: invalid character 's' looking for beginning of value
```

cmd/common/image/inspect.go
```diff
@@ -107,35 +109,96 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
...
+			}
+		}
+
 		switch outputMode {
 		case utils.OutputModeJSON:
-			bytes, err := json.Marshal(extraInfoMap)
-			if err != nil {
-				return fmt.Errorf("marshalling image and extra info to JSON: %w", err)
+			if customResult != nil {
+				if _, ok := customResult.(string); !ok {
```
alban, line 150: Instead of testing for `string`, I wonder about checking if we have a JSON document:
```go
isJSON := dataEntry.ContentType == "application/json"
...
if isJSON {
    bytes, err := json.Marshal(customResult)
    ...
    fmt.Fprint(cmd.OutOrStdout(), string(bytes), "\n")
} else {
    fmt.Fprint(cmd.OutOrStdout(), customResult, "\n")
}
```

That would make a difference if the JSON document is a single string. But I think it is better for consistency.

alban (commented):
A few comments to address and then the patch should be ready to go!

Please use `gofumpt -w`.

cmd/common/image/inspect.go
KapilSareen: I did this because of this-

ebpf.program is of type "application/json". But when we want a specific Source of a program, we don't want it to be json.Marshalled because this is the output format :
```
$ sudo ./ig image inspect ghcr.io/inspektor-gadget/gadget/trace_tcp --extra-info 'ebpf.programs' --jsonpath='[0].Source'

"sys_accept4_x:\n\t   ; int sys_accept4_x(struct syscall_trace_exit *ctx)\n\t  0: MovReg dst: r9 src: r1\n\t   ; __u32 tid = bpf_get_current_pid_tgid();\n\t  1: Call FnGetCurrentPidTgid\n\t   ; __u32 tid = bpf_get_current_pid_tgid();\n\t  2: StXMemW dst: rfp src: r0 off: -20 imm: 0\n\t  3: MovImm dst: r1 imm: 0\n\t   ; struct tuple_key_t t = {};\n\t  4: StXMemW dst: rfp src: r1 off: -24 imm: 0\n\t  5: StXMemDW dst: rfp src: r1 off: -32 imm: 0\n\t  6: StXMemDW dst: rfp src: r1 off: -40 imm: 0\n\t  7: StXMemDW dst: rfp src: r1 off: -48 imm: 0\n\t  8: StXMemDW dst: rfp src: r1 off: -56 imm: 0\n\t
```

and we want it to be like this:
```
$ sudo ./ig image inspect ghcr.io/inspektor-gadget/gadget/trace_tcp --extra-info 'ebpf.programs' --jsonpath='[0].Source'

ig_tcp_v4_co_x:
           ; int BPF_KRETPROBE(ig_tcp_v4_co_x, int ret)
          0: LdXMemDW dst: r1 src: r1 off: 80 imm: 0
          1: MovImm dst: r2 imm: 0
           ; struct tuple_key_t tuple = {};
          2: StXMemW dst: rfp src: r2 off: -24 imm: 0
          3: StXMemDW dst: rfp src: r2 off: -32 imm: 
          ...
          
  ```
  
  LMK your thoughts on this, I keeping it as it is for now.

File: docs/reference/images.md [hunk 10]
```diff
@@ -313,7 +314,6 @@ $ sudo ig image inspect ghcr.io/inspektor-gadget/gadget/trace_tcp
 
 # List possible extra-info
 $ sudo ig image inspect fsnotify:main |jq keys
-WARN[0001] image signature verification is disabled due to using corresponding option
```
alban, line 316: A few lines below, the doc says:
```
  "ebpf.mermaid.flowchart",
  "ebpf.mermaid.sequence",
```
But these have been renamed. The doc should be updated accordingly.

docs/reference/images.md
```diff
@@ -331,25 +331,27 @@ WARN[0001] image signature verification is disabled due to using corresponding o
   "wasm.upcalls"
 ]
 
-# Retrieve specific info using custom output
-$ sudo ig image inspect -o custom --extra-info=ebpf.sections ghcr.io/inspektor-gadget/gadget/trace_tcp
+# Retrieve specific info using --extra-info
+$ sudo ig image inspect --extra-info=ebpf.sections ghcr.io/inspektor-gadget/gadget/trace_tcp
 ["",".strtab",".text","kprobe/tcp_v4_connect",".relkprobe/tcp_v4_connect","kretprobe/tcp_v4_connect",".relkretprobe/tcp_v4_connect","kprobe/tcp_v6_connect",".relkprobe/tcp_v6_connect","kretprobe/tcp_v6_connect",".relkretprobe/tcp_v6_connect","kprobe/tcp_close",".relkprobe/tcp_close","kprobe/tcp_set_state",".relkprobe/tcp_set_state","kretprobe/inet_csk_accept",".relkretprobe/inet_csk_accept",".rodata",".bss","license",".maps",".BTF",".rel.BTF",".BTF.ext",".rel.BTF.ext",".llvm_addrsig",".symtab"]
```
alban, line 336: When I test it, it prints "jsonpretty" by default.
Either update the output, or change the command line with `-o json` so that it matches the output.

docs/reference/images.md
```diff
@@ -372,8 +374,16 @@ $ sudo ig image inspect audit_seccomp:main --extra-info=oci.manifest -o custom|j
   "org.opencontainers.image.url": "https://inspektor-gadget.io/"
 }
 
+# Indexing ebpf programs using the jsonpath flag
+$ sudo ig image inspect advise_seccomp:latest --verify-image=false --extra-info=ebpf.programs --jsonpath="[0].Section"
```
alban, line 378:
```suggestion
$ sudo ig image inspect advise_seccomp:latest --extra-info=ebpf.programs --jsonpath="[0].Section"
```

docs/reference/images.md
```diff
@@ -372,8 +374,16 @@ $ sudo ig image inspect audit_seccomp:main --extra-info=oci.manifest -o custom|j
...
+"raw_tracepoint/sys_enter"
+
+# Indexing oci manifest fields using jsonpath flag
+$ sudo ig image inspect advise_seccomp:latest --verify-image=false --extra-info=oci.manifest --jsonpath=".layers[0].digest"
```
alban, line 382:
```suggestion
$ sudo ig image inspect advise_seccomp:latest --extra-info=oci.manifest --jsonpath=".layers[0].digest"
```

cmd/common/image/inspect.go
```diff
@@ -107,35 +109,94 @@ func NewInspectCmd(runtime runtime.Runtime) *cobra.Command {
...
+				// unmarshal the JSON content
+				if err := json.Unmarshal(dataEntry.Content, &customResult); err != nil {
+					return fmt.Errorf("unmarshalling JSON content: %w", err)
+				}
+			} else {
+				customResult = string(dataEntry.Content)
+			}
+			if jsonPath != "" {
+				if dataEntry.ContentType != "application/json" {
+					return fmt.Errorf("jsonpath %q can only be used with JSON content", jsonPath)
```
alban, line 132: Would it be possible to use --jsonpath on yaml documents?
```
$ sudo ig image inspect audit_seccomp:main --extra-info=oci.metadata --jsonpath='.datasources.seccomp.fields.ustack'
Error: jsonpath ".datasources.seccomp.fields.ustack" can only be used with JSON content
```

alban (commented):
I think it should return an error when giving a jsonpath without extrainfo (instead of ignoring it).
```
sudo ig image inspect fsnotify:latest --jsonpath=']]]]'
```

cmd/common/image/inspect.go
KapilSareen: I suppose not, we're using `jsonpath.Get` to index.

alban: `jsonpath.Get()` takes aa `interface{}` as argument so I was wondering if it would accept an `interface{}` created by a yaml unmarshaller.

alban (commented):
Thanks! Last few comments.
You can rebase and squash the commits into 1.

alban (approved):
LGTM, thanks!

Issue #4335, opened 2025-04-15
[RFE] Improve messaging of `ig image build`


mqasimsarfraz:
Currently we are able to run `ig build image` in any directory and it produces a gadget:

```
$ mktemp -d
/tmp/tmp.mJR3WT0cjO
$ cd /tmp/tmp.mJR3WT0cjO
$ sudo ig image build -t foo:bar .
Successfully built ghcr.io/inspektor-gadget/gadget/foo:bar@sha256:c34ea0149ba4f31b286e81ab5c077acbfba62c94efcf250f35cb00dcba795a47
```

Which might be expected because ebpf layer is optional but this can confusing for user that are just getting started and might run `ig image build` in incorrect directory. We should consider having a info/warning message telling users that they have built a gadget w/o ebpf layer or an empty gadget. 

cc: @burak-ok

AdvH039:
@mqasimsarfraz Can I be assigned this issue? Are there more example cases where we should log warnings for the users?

mqasimsarfraz:
@AdvH039 After a discussion with @blanquicet on this issue, we decided the gadget should be built successfully only if user either specifies:
- program.bpf.c or
- metadata file 
So rather then log message we should just print the error and exit if none of them is provided by the user. I would want to get more thoughts before we move towards implementation. @burak-ok / @mauriciovasquezbernal  thoughts?

burak-ok:
Yeah, any valid inputfile should be available before building.

Bet it the `program.bpf.c` file, the `metadata` file or probably only a single WASM/golang file

AdvH039:
So I have to just check if the other files like `metadata, wasm`  exist like similar to what we did [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/4d2c7dc08013879219dd62a6be5b0c23643c7b58/cmd/common/image/build.go#L206) for program.bpf.c and if none of the files exist then return an error?

mauriciovasquezbernal:
> * program.bpf.c or
> 
> * metadata file
>   So rather then log message we should just print the error and exit if none of them is provided by the user. I would want to get more thoughts before we move towards implementation. [@burak-ok](https://github.com/burak-ok) / [@mauriciovasquezbernal](https://github.com/mauriciovasquezbernal)  thoughts?

That makes sense. Also a single golang module should be valid as mentioned by Burak.

I think the implementation should check if the generated image doesn't have any layer and print an error / warning.

[...]

PR #4376, opened 2025-04-25
Give error when incorrect build

AdvH039:
Refer #4335 

Screenshot from 2025-04-25 22-55-50

The error is visible while building empty directories, but I run into a `pulling builder image` problem whenever I use an
ig executable local to the directory with the new changes or not.

image

Also it shows an error for directory containing standalone go files. I think this is because they don't form a layer in the final built image.

### DIFF
--- a/cmd/common/image/build.go
+++ b/cmd/common/image/build.go
@@ -210,6 +210,42 @@ func runBuild(cmd *cobra.Command, opts *cmdOpts) error {
 		conf.EBPFSource = ""
 	}
 
+	var hasEBPFSource, hasMetadata, hasWasm, hasGo bool
+	if conf.EBPFSource != "" {
+		if _, err := os.Stat(conf.EBPFSource); err == nil {
+			hasEBPFSource = true
+		}
+	}
+	if conf.Metadata != "" {
+		if _, err := os.Stat(conf.Metadata); err == nil {
+			hasMetadata = true
+		}
+	}
+	if conf.Wasm != "" {
+		if _, err := os.Stat(conf.Wasm); err == nil {
+			hasWasm = true
+		}
+	}
+
+	goFolder := "./go"
+
+	if _, err := os.Stat(goFolder); err == nil {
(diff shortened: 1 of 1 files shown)


mqasimsarfraz:
I think we should perform this check before building the gadget. I did some [testing](https://github.com/mqasimsarfraz/inspektor-gadget/commit/d797ca2acd4a40af679fac9cdc9b0ad2950dd36c) (please feel free to use it/polish it). Also, the error message needs to include more information that at least one of the input (ebpf source, metadata, wasm) is needed.

AdvH039:
Cool. I'll get it done by Friday. I have exams right now. Thank you !!

AdvH039:
Should I just add a check for existence of go files to your code and make a pr? I think most of it is covered with your changes.

cmd/common/image/build.go
```diff
@@ -210,6 +210,42 @@ func runBuild(cmd *cobra.Command, opts *cmdOpts) error {
...
+			hasMetadata = true
+		}
+	}
+	if conf.Wasm != "" {
+		if _, err := os.Stat(conf.Wasm); err == nil {
+			hasWasm = true
+		}
+	}
+
+	goFolder := "./go"
```
mqasimsarfraz, line 230: @mauriciovasquezbernal @eiffel-fl Just wanted to you to be aware of this change since we might need something for rust as well.

mqasimsarfraz (approved):
LGTM, I will keep it open for a bit to get more reviews but then I will merge it. Thanks!

cmd/common/image/build.go
mauriciovasquezbernal: @Daksh-10 FYI!

Daksh-10: I also tried to build with empty folders and it built it successfully, even though I didn't provide any required files. I'll implement it for rust as well.

mqasimsarfraz:
Thanks for the contribution!

PR #4396, opened 2025-04-29
GadgetContext: Close local operators after remote operators are all closed

burak-ok:
Before this PR both code paths listened on the same channel. So when a Ctrl+C comes in we do the two things in parallel:
1. Close all operators on the client side
2. Notify server side to close all operators

As an example problems always occur when one is using the `ebpf.map.flush-on-stop` annotation:
1. We are closing all operators on the client side. This is almost guaranteed to execute first and faster (no grpc communication, etc..)
2. We are notifying the server side that the gadget should stop
3. `ebpf.map.flush-on-stop` is seen and the map and events are flushed now to the client
4. Events are received by the client, but all operators are already stopped -> No further processing is done
5. The local CLI operator is still printing these though, since it doesn't do anything in its `.Stop()`

For the use case of `advise networkpolicy` I rely on the local operator `Combiner` and if that is Stopped, it doesn't process any events anymore (unlike the local CLI operator)

This PR changes the behavior to be the following way:
1. Notify server side to close all operators
2. Wait till all servers seized their operations
3. Close the local operators

### DIFF
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -69,6 +69,7 @@ type GadgetContext struct {
 	lock             sync.Mutex
 	dataSources      map[string]datasource.DataSource
 	dataOperators    []operators.DataOperator
+	localOperators   []operators.DataOperatorInstance
 	vars             map[string]any
 	params           []*api.Param
 	prepareCallbacks []func()
@@ -374,25 +375,18 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
 		c.SetVar("config", v)
 	}
 
-	// After loading gadget info, start local operators as well
-	localOperators, err := c.initAndPrepareOperators(paramValues)
+	var err error
+	// After loading gadget info, get local operators params as well
+	c.localOperators, err = c.initAndPrepareOperators(paramValues)
 	if err != nil {
 		return fmt.Errorf("initializing local operators: %w", err)
 	}
 
 	if run {
-		if err := c.start(localOperators); err != nil {
+		if err := c.start(c.localOperators); err != nil {
 			return fmt.Errorf("starting local operators: %w", err)
 		}
-
 		c.Logger().Debugf("running...")
-
-		go func() {
-			// TODO: Client shouldn't need to wait for the timeout. It should be
-			// managed only on the server side.
-			WaitForTimeoutOrDone(c)
-			c.stop(localOperators)
-		}()
 	}
 
 	if c.ExtraInfo() && extraInfo != nil {
@@ -410,6 +404,14 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
 		}
 	}
 	return nil
+}
+
+func (c *GadgetContext) StopLocalOperators() {
+	if c.localOperators == nil {
+		return
+	}
+	c.stop(c.localOperators)
+	c.localOperators = nil
 }
 
 func (c *GadgetContext) OrasTarget() oras.ReadOnlyTarget {
--- a/pkg/runtime/grpc/oci.go
+++ b/pkg/runtime/grpc/oci.go
@@ -123,6 +123,9 @@ func (r *Runtime) runGadgetOnTargets(
 	}
 
 	wg.Wait()
+	// Stop local operators after all remote targets
+	// have stopped their operators and "returned"
+	gadgetCtx.StopLocalOperators()
 	return results, results.Err()
 }
 
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -59,6 +59,7 @@ type GadgetContext interface {
 	GetVar(string) (any, bool)
 	SerializeGadgetInfo(requestExtraInfo bool) (*api.GadgetInfo, error)
 	LoadGadgetInfo(info *api.GadgetInfo, paramValues api.ParamValues, run bool, extraInfo *api.ExtraInfo) error
+	StopLocalOperators()
 	Params() []*api.Param
 	SetMetadata([]byte)
 	SetParams([]*api.Param)


pkg/gadget-context/gadget-context.go
```diff
@@ -414,6 +401,49 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
...
+
+	go func() {
+		// TODO: Client shouldn't need to wait for the timeout. It should be
+		// managed only on the server side.
+		WaitForTimeoutOrDone(c)
+
+		select {
+		case <-remoteDone:
+		// TODO Do we need a timeout here? I think not, since a second Ctrl+C does an exit
+		case <-time.After(3 * time.Second):
```
mauriciovasquezbernal, line 438: The issue is that the client will remain stuck is the server never ends, however I think it's fine and I'd prefer not to introduce another arbitrary-defined timeout in our code.

mauriciovasquezbernal (commented):
It looks very good @burak-ok. I modified the CLI operator not to print after it has been closed, and I can see how the issue you mentioned on the description is reproducible in the main branch but not on this PR.

pkg/gadget-context/gadget-context.go
burak-ok: Thats what I thought. Also as the comment indicates, even if we have no timeout here, a second `Ctrl +C` should terminate the process anyway

pkg/gadget-context/gadget-context.go
```diff
@@ -414,6 +401,49 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
...
+	//
+	// The naming suggests that this is only to check if the info was alreadty loaded
+	// but it is also used to check if the gadget was already started?
+	c.lock.Lock()
+	if c.loaded {
+		// TODO: verify that info matches what we previously loaded
+		c.lock.Unlock()
+		return nil
+	}
+	c.lock.Unlock()
```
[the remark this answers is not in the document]
burak-ok: We need something like this though.
The `LoadGadgetInfoAndRun` would break the whole logic for concurrent runs. Or we can say that we don't support and expect multiple calls to `LoadGadgetInfoAndRun`

pkg/gadget-context/gadget-context.go
mauriciovasquezbernal: I changed by mind. I think we need to introduce this timeout to be totally sure this logic ends even if the server hangs. When running `kubectl-gadget` or `gadgetctl` pressing Ctrl + C twice does the trick, but it won't work if those binaries are executed by another binary without the logic to handle that case or if this code is used directly from Go.

cc @alban @flyth thoughts on this?

pkg/gadget-context/gadget-context.go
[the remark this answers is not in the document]
mauriciovasquezbernal: I think we don't support multiple calls to `Run()` either, so IMO we're fine without this logic.

pkg/gadget-context/gadget-context.go
```diff
@@ -412,6 +399,35 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
...
+
+	go func() {
+		// TODO: Client shouldn't need to wait for the timeout. It should be
+		// managed only on the server side.
+		WaitForTimeoutOrDone(c)
+
+		select {
+		case <-remoteDone:
+		// TODO Do we need a timeout here? I think not, since a second Ctrl+C does an exit
+		case <-time.After(3 * time.Second):
```
mauriciovasquezbernal, line 422: I think 3 seconds is too short. IMO it should be > 10 seconds so the server has time to send the information to the client.

mauriciovasquezbernal (approved):
LGTM from code inspection, however I'd like to get another opinion on this.

pkg/gadget-context/gadget-context.go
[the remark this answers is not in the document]
burak-ok: It appears that we need the logic.
When we have N nodes `LoadGadgetInfoAndRun` is run N times and we don't want to start the local operators multiple times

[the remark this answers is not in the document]
burak-ok: No extra logic needed anymore, the new function is gone

pkg/gadget-context/gadget-context.go
burak-ok: No extra timeout needed, since we already have a default wait time for the server side to close in 
https://github.com/inspektor-gadget/inspektor-gadget/blob/e0f2eb7fa91346bb4a1e0bfc93e73f1da3143d9a/pkg/runtime/grpc/oci.go#L292

pkg/gadget-context/gadget-context.go
burak-ok: See https://github.com/inspektor-gadget/inspektor-gadget/pull/4396#discussion_r2100732546

pkg/runtime/grpc/oci.go
```diff
@@ -123,6 +123,11 @@ func (r *Runtime) runGadgetOnTargets(
 	}
 
 	wg.Wait()
+	// If we are in interactive mode, stop local operators after all remote targets
+	// have stopped their operators and "returned"
+	if !gadgetCtx.UseInstance() {
+		gadgetCtx.StopLocalOperators()
+	}
```
mauriciovasquezbernal, line 130: Why not to call it in all cases?

mauriciovasquezbernal:
I think there are still some issues, like "local operators" not being stopped when calling LoadGadgetInfo(), but this is being handled in #3778. I have a question about `UseInstance`, otherwise LGTM!

pkg/runtime/grpc/oci.go
burak-ok: Now that you asked that, it might make sense to just run it everytime.

I though in the daemon or headless variant we don't want to stop the local operators - but they are only on the client anyway...
Maybe I missed something now, but is something like the `sort` or `combiner` operator running in a detached mode. No, right?

mauriciovasquezbernal: Ah, good point. We're running those operators even in detached mode. I think we need to run them to get their parameters, but since these are operators on the client, probably those parameters should only be set on `attach`.

@flyth any thoughts on this?

burak-ok: > Ah, good point. We're running those operators even in detached mode. I think we need to run them to get their parameters, but since these are operators on the client, probably those parameters should only be set on attach.

We could also keep the behavior the same as before this PR. We always stopped the local operators when the context is done

mauriciovasquezbernal: I agree, feel free then to remove the `if gadgetCtx.UseInstance() {` conditional and merge if you think it's ready to go.

burak-ok: Done

PR #4400, opened 2025-04-29
Document filter quoting

johananl:
# Document filter quoting

When specifying filters which contain special characters either on a shell or in a YAML document, it's easy to run into quoting issues. For example, specifying proc.comm~'^bash$' in a YAML document results in the single quotes being passed down to IG which in turn causes unexpected filtering behavior in user space.

## How to use

Ensure the content makes sense and the docs are rendered correctly.

## Testing done

I ran

```
make docs
npm install docusaurus
npm run dev
```

under https://github.com/inspektor-gadget/website and ensured the documents render fine.

### DIFF
--- a/docs/reference/run.mdx
+++ b/docs/reference/run.mdx
@@ -142,6 +142,16 @@ The filter syntax supports the following operations:
 - `field~value`: Matches if the content of `field` matches the regular expression `value`. See [RE2 Syntax](https://github.com/google/re2/wiki/Syntax) for more details.
 ```
 
+:::info
+
+It's recommended to wrap the **entire** filter expression with single quotes when using filters containing special characters to avoid unexpected behavior:
+
+```bash
+--filter 'proc.comm~^ba.*$'
+```
+
+:::
+
 #### Examples
 
 **Equal filter**: To filter events where the `comm` field equals `cat`, use:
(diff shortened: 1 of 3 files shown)


docs/spec/operators/filter.md
```diff
@@ -30,6 +30,25 @@ The filter syntax supports the following operations:
...
+[CLI](../../reference/run.mdx) example:
+
+```bash
+--filter 'proc.comm~^ba.*$'
+```
+
+[Gadget instance manifest](../../reference/manifests.mdx) example:
+
+```yaml
+operator.filter.filter: 'proc.comm~^ba.*$'
```
alban, line 47: What about also adding this example in `docs/reference/manifests.mdx`?

docs/reference/run.mdx
```diff
@@ -142,6 +142,17 @@ The filter syntax supports the following operations:
 - `field~value`: Matches if the content of `field` matches the regular expression `value`. See [RE2 Syntax](https://github.com/google/re2/wiki/Syntax) for more details.
 ```
 
+:::info
+
+If facing unexpected behavior with filters containing special characters, try wrapping the
+**entire** filter expression with single quotes (rather than just the right hand of the operator):
```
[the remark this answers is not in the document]
johananl: Thanks, I've taken your rephrase and added "with single quotes" for clarity.

docs/spec/operators/filter.md
johananl: Happy to do that. Which part of the page specifically did you have in mind @alban?

alban: I was thinking this part:

https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/reference/manifests.mdx#running-a-single-gadget-interactively
https://inspektor-gadget.io/docs/latest/reference/manifests#running-a-single-gadget-interactively

where you have:
```yaml
paramValues:
```

alban (approved):

PR #4401, opened 2025-04-29
Update fields in filter examples

johananl:
# Update fields in filter examples

Some filter examples we have in the docs use outdated fields.

## How to use

Check that the new filters are correct.

## Testing done

I used the modified filters locally using `ig` and ensured they work.

### DIFF
--- a/docs/reference/run.mdx
+++ b/docs/reference/run.mdx
@@ -147,7 +147,7 @@ The filter syntax supports the following operations:
 **Equal filter**: To filter events where the `comm` field equals `cat`, use:
 
 ```bash
---filter comm==cat
+--filter 'proc.comm==cat'
 ```
 
 This filters for events related to the `cat` command.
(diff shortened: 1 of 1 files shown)


docs/reference/run.mdx
```diff
@@ -147,29 +147,29 @@ The filter syntax supports the following operations:
...
 ```bash
---filter uid!=0
+--filter proc.creds.uid!=0
 ```
 
 **Regular expression filter**: To match a regular expression in a specific column, use:
 
 ```bash
---filter comm~'^c.*'
+--filter proc.comm~'^c.*'
```
alban, line 164:
```suggestion
--filter 'proc.comm~^c.*'
```

The other examples could have simple quotes too.

blanquicet: Yeah, quotes are necessary when using negative regex like `-F 'proc.comm!~^c.*'`. In any case, I think it's a good practice to suggest using quotes. WDYT? If you agree, I suggest adding it to the documentation.

johananl: I agree. Fixed.

johananl: @blanquicet I'm not clear on your other suggestion. Could you clarify what I should change compared to the current state of the PR?

alban (approved):

alban:
It seems the CI is unreliable, but it is not related to this PR, so I'll merge.

PR #4402, opened 2025-04-29
helm: Add parameter to enable bpfstats

mauriciovasquezbernal:
Leftover from #3820

### DIFF
--- a/charts/gadget/templates/configmap.yaml
+++ b/charts/gadget/templates/configmap.yaml
@@ -18,6 +18,8 @@ data:
       docker-socketpath: {{ .Values.config.dockerSocketPath }}
       podman-socketpath: {{ .Values.config.podmanSocketPath }}
       operator:
+        ebpf:
+          enable-bpfstats: {{ .Values.config.enableBpfStats }}
         oci:
           verify-image: {{ .Values.config.verifyGadgets }}
           public-keys:{{ toYaml $.Values.config.gadgetsPublicKeys | nindent 12 }}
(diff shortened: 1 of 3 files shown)


alban (approved):
LGTM, that seems inline with #3820.

Why is enableBpfStats not enabled by default? Does it mean the gadget won't work unless IG was deployed with this option?

mqasimsarfraz (approved):
Looking good! 

I am asking myself if it would make sense to allow user to specify custom params under `operator` via helm values so we don't limit the charts.

mauriciovasquezbernal:
> Why is enableBpfStats not enabled by default? 

Because of performance reasons. As described on https://docs.ebpf.io/linux/syscall/BPF_ENABLE_STATS/, enabling the collection of BPF stats has a performance on the system.

> Does it mean the gadget won't work unless IG was deployed with this option?

That's right. Now that you raised this topic, I think it's not the best UX as statistics should be enabled only when the bpfstats gadget is run. I'll think about it more.

> Same question as Alban, is this performance related? And is this documented?

That's not documented. I'll handle it while checking the UX too.

> I am asking myself if it would make sense to allow user to specify custom params under operator via helm values so we don't limit the charts.

Yeah, it took also me a bit of time to understand what we were missing. I don't know what's the right solution, but it seems this duplication (`kubectl gadget deploy`, helm chart and config map) is creating some confusion. 
Perhaps a solution based on the parameters library (as mentioned by @flyth) could make more sense.

mauriciovasquezbernal:
@alban @eiffel-fl I opened #4438 to improve the UX for that gadget.

PR #4081, opened 2025-02-18
[test]:create test for trace_capabilities in gadgets

Dhiren-Mhatre:
# Add unit tests for trace_capabilities gadget

This PR adds unit tests for the trace_capabilities gadget following the same pattern as trace_open_test.go. The tests cover basic functionality, filter validation, capability checking, audit mode, and proper UID/GID tracking. This contributes to improving unit test coverage as requested in #3835.

## How to use
To validate this PR, please run the unit tests for the trace_capabilities gadget:
```bash
go test -v ./gadgets/trace_capabilities/test/unit/
```

## Testing done
I have run the unit tests locally with the following command and verified they pass:
```bash
$ go test -v ./gadgets/trace_capabilities/test/unit/
ok      github.com/inspektor-gadget/inspektor-gadget/gadgets/trace_capabilities/test/unit     3.525s
```

The tests verify:
- Basic event capture with no filters
- Filter behavior for both matching and non-matching cases
- Different capability checks
- Audit mode functionality
- Proper UID/GID tracking in events

Part of  #3835 (as it adds coverage for trace_capabilities).

PR #4403, opened 2025-04-29
Unit Tests for Trace Capabilities Gadget

Shaheer-Ahmd:
# Add unit tests for trace_capabilities gadget

This PR introduces a new set of unit tests for the `trace_capabilities` gadget to validate its ability to detect capability-related syscalls. These tests simulate various syscalls (`unshare`, `socket`, `chroot`, `nice`) that require specific Linux capabilities (`CAP_SYS_ADMIN`, `CAP_NET_RAW`, `CAP_SYS_CHROOT`, `CAP_SYS_NICE`) and assert that the gadget captures the corresponding events with the correct capability.

Each test case uses a minimal syscall invocation to trigger the gadget and then checks that exactly one event is captured, with the expected `Cap` field value. The tests are written using the standard Go `testing` package, along with helper utilities from `inspektor-gadget` for setting up the test environment and gadget runner.

an effort towards #3835 

## How to use

To validate this PR, reviewers can run the following command to execute the unit tests:

```bash
GADGET_TAG=trace_cap_unit_tests IG_VERIFY_IMAGE=false go test -v -exec 'sudo -E' ./gadgets/trace_capabilities/test/unit/...
```

Ensure that the tests pass and that each one captures a single event with the correct capability name.

## Testing done

I manually ran the following command to execute the unit tests:

```bash
GADGET_TAG=trace_cap_unit_tests IG_VERIFY_IMAGE=false go test -v -exec 'sudo -E' ./gadgets/trace_capabilities/test/unit/...
```

All test cases passed successfully, and the gadget accurately captured the capability events for each syscall. The output confirmed that the expected capabilities matched the captured `Cap` values in each case.

### DIFF
--- a/gadgets/trace_capabilities/test/unit/trace_capabilities_test.go
+++ b/gadgets/trace_capabilities/test/unit/trace_capabilities_test.go
@@ -1,0 +1,189 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"os"
+	"os/exec"
+	"syscall"
+	"testing"
+	"time"
+	"unsafe"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/gadgetrunner"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+)
+
+type ExpectedTraceCapabilitiesEvent struct {
+	Proc          utils.Process
+	Cap           string `json:"cap"`
+	Audit         uint32 `json:"audit"`
+	Syscall       string `json:"syscall"`
+	CurrentUserNs uint64 `json:"current_userns"`
+	TargetUserNs  uint64 `json:"target_userns"`
+	Insetid       uint32 `json:"insetid"`
+}
+
+type testDef struct {
+	name                string
+	generateEvent       func()
+	requestedPermission string
+	syscall             string
+	insetId             uint32
+}
+
+func TestTraceCapabilitiesGadget(t *testing.T) {
+	gadgettesting.InitUnitTest(t)
+	testCases := []testDef{
+		{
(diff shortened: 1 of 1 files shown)


gadgets/trace_capabilities/test/unit/trace_capabilities_test.go
```diff
@@ -0,0 +1,98 @@
...
+			name: "unshare",
+			generateEvent: func() error {
+				return syscall.Unshare(syscall.CLONE_NEWUTS)
+			},
+			requestedPermission: "CAP_SYS_ADMIN",
+		},
+		{
+			name: "raw_socket",
+			generateEvent: func() error {
+				_, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_RAW, syscall.IPPROTO_ICMP)
```
burak-ok, line 53: If the syscall was successfull, the socket should be closed again

gadgets/trace_capabilities/test/unit/trace_capabilities_test.go
```diff
@@ -0,0 +1,98 @@
...
+	requestedPermission string
+}
+
+func TestTraceCapabilitiesGadget(t *testing.T) {
+	gadgettesting.InitUnitTest(t)
+	testCases := []testDef{
+		{
+			name: "unshare",
+			generateEvent: func() error {
+				return syscall.Unshare(syscall.CLONE_NEWUTS)
```
burak-ok, line 46: I would ignore the returned error in the `genereateEvent` functions.
We have a test to capture if that permission was requested here - So we check if we requested it at all - but we do not mind if the request itself was successfull or blocked

gadgets/trace_capabilities/test/unit/trace_capabilities_test.go
mauriciovasquezbernal: Not only for this, all successful changes should be rolled back to leave the process on the same estate as before.

mauriciovasquezbernal (commented):
The actions performed in the tests are shared by all other tests running on parallel. I think we can't do things like chroot, or unshare on the tests. I think it's fine to remove those two and test with the remaining ones you have on this PR.

mauriciovasquezbernal (commented):
btw, #4081 was also implementing these tests. Perhaps you can take some tests from that one too and merge into a single PR. (Please be sure to give attribution to the original author of the PR)

mauriciovasquezbernal:
> Do you think I should take a lock via runtime.LockOSThread for the entirety of the generateEvent function? Or should I just remove these tests?

The runner already takes the lock [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/119ef757d09e1009511ab35303ce8bf965bf2510/internal/test/runner.go#L113-L118), the issue is that some changes affect the whole process, hence other threads will be affected too affecting parallel tests. We need to be sure the changes only affect the current thread and then we're all fine.

mauriciovasquezbernal (commented):
The tests look good, however some fail on my machine and CI. Could you please check them?

Shaheer-Ahmd:
> The tests look good, however some fail on my machine and CI. Could you please check them?

I replaced the tests that had side effects on the whole process with those that have no such side effects. Now, the tests are passing.

Shaheer-Ahmd:
> The tests look good, however some fail on my machine and CI. Could you please check them?

one of the tests requested different permissions based on the kernel version which is why some tests were failing - added conditional branching to fix that. Also, I replaced the process-affecting tests with ones that do not have a side effect on other threads.

mauriciovasquezbernal (approved):
LGTM. Will not merge right now to see if somebody else wants to have a look.

burak-ok (approved):
LGTM

PR #4413, opened 2025-05-01
Add Trace Bind Unit Tetsts

Shaheer-Ahmd:
# Add unit tests for trace_bind gadget covering multiple protocol and port combinations

This PR introduces a new set of unit tests for the `trace_bind` gadget in the Inspektor Gadget project. The tests are designed to validate the correct event capturing behavior of the gadget when various socket binding operations are performed. Specifically, the tests cover combinations of IPv4 and IPv6 with both TCP and UDP protocols, using both fixed and ephemeral ports.

Each test verifies that the captured bind event contains the expected address, protocol, and port (non-zero for ephemeral ports). The tests utilize the netcat (`nc`) utility to simulate socket bindings and ensure that the gadget emits one event per bind operation, with accurate protocol and address information.

an effort towards #3835 

## How to use

To validate this PR, reviewers can run the unit tests using the following command from the project root:

```bash
GADGET_TAG=trace-bind-unit-tests IG_VERIFY_IMAGE=false go test -v -exec 'sudo -E' ./gadgets/trace_bind/test/unit/...
````

## Testing done

The following command was executed to run the new unit tests:

```bash
GADGET_TAG=trace-bind-unit-tests IG_VERIFY_IMAGE=false go test -v -exec 'sudo -E' ./gadgets/trace_bind/test/unit/...
```

All test cases passed successfully. Each case verified that:

* The bind operation was detected and emitted exactly one event.
* The emitted event had the expected IP address, protocol, and port (or a non-zero port for ephemeral tests).

```

### DIFF
--- a/gadgets/trace_bind/test/unit/trace_bind_test.go
+++ b/gadgets/trace_bind/test/unit/trace_bind_test.go
@@ -1,0 +1,171 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"net"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/gadgetrunner"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+)
+
+type ExpectedTraceBindEvent struct {
+	Addr  utils.L4Endpoint `json:"addr"`
+	Error string           `json:"error"`
+	Proc  utils.Process
+}
+
+type testCase struct {
+	name    string
+	port    uint16
+	network string
+	version uint8
+	addr    string
+}
+
+func TestTraceBind(t *testing.T) {
+	gadgettesting.InitUnitTest(t)
+	testCases := []testCase{
+		{
+			name:    "TCP4 loopback on fixed port",
+			addr:    "127.0.0.1",
+			network: "TCP",
+			version: 4,
+			port:    12345,
+		},
+		{
+			name:    "TCP4 loopback on ephemeral port",
+			addr:    "127.0.0.1",
+			network: "TCP",
+			version: 4,
+			port:    0,
+		},
+		{
+			name:    "TCP6 loopback on fixed port",
+			addr:    "::1",
+			network: "TCP",
+			port:    2154,
+			version: 6,
+		},
+		{
+			name:    "UDP4 loopback on high port",
+			addr:    "127.0.0.1",
+			network: "UDP",
+			port:    12345,
+			version: 4,
+		},
+		{
+			name:    "UDP6 loopback on ephemeral port",
+			addr:    "::1",
+			network: "UDP",
+			port:    0,
+			version: 6,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			runner := utilstest.NewRunnerWithTest(t, &utilstest.RunnerConfig{})
+			onGadgetRun := func(gadgetCtx operators.GadgetContext) error {
+				utilstest.RunWithRunner(t, runner, func() error {
+					// pick socket parameters based on tc.network
+					var domain, socktype, proto int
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: gadgets/trace_bind/test/unit/trace_bind_test.go [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+
+	utilstest "github.com/inspektor-gadget/inspektor-gadget/internal/test"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/gadgetrunner"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/testing/utils"
+)
+
+type ExpectedTraceBindEvent struct {
+	Addr  utils.L4Endpoint `json:"addr"`
```
mauriciovasquezbernal, line 34: Please also add `Proc utils.Process `json:"proc"`` to test it.

gadgets/trace_bind/test/unit/trace_bind_test.go
```diff
@@ -0,0 +1,121 @@
...
+			onGadgetRun := func(gadgetCtx operators.GadgetContext) error {
+				utilstest.RunWithRunner(t, runner, func() error {
+					argv := []string{"/usr/bin/nc", "-l", "-s", tc.addr, "-p", fmt.Sprintf("%d", tc.port), "-w", "1"}
+					if tc.network == "udp4" || tc.network == "udp6" {
+						argv = append(argv, "-u")
+					}
+					_, err := os.StartProcess("/usr/bin/nc", argv, &os.ProcAttr{})
+					require.NoError(t, err)
+					return nil
+				})
```
mauriciovasquezbernal, line 92: Is it possible to implement this without relying on an external binary? I think you can open bind the socket directly from Go.

gadgets/trace_bind/test/unit/trace_bind_test.go
```diff
@@ -0,0 +1,121 @@
...
+			require.Len(t, gadgetRunner.CapturedEvents, 1, "Expected 1 event, got %d", len(gadgetRunner.CapturedEvents))
+
+			require.Equal(t, tc.addr, gadgetRunner.CapturedEvents[0].Addr.Addr, "Expected address %s, got %s", tc.addr, gadgetRunner.CapturedEvents[0].Addr.Addr)
+			require.Contains(t, tc.network, strings.ToLower(gadgetRunner.CapturedEvents[0].Addr.Proto), "Expected protocol %s, got %s", tc.network[:3], gadgetRunner.CapturedEvents[0].Addr.Proto)
+
+			if tc.port == 0 {
+				require.NotEqual(t, 0, gadgetRunner.CapturedEvents[0].Addr.Port, "Expected non-zero port, got %d", gadgetRunner.CapturedEvents[0].Addr.Port)
+			} else {
+				require.Equal(t, tc.port, gadgetRunner.CapturedEvents[0].Addr.Port, "Expected port %d, got %d", tc.port, gadgetRunner.CapturedEvents[0].Addr.Port)
+			}
```
mauriciovasquezbernal, line 118: I think it's better to match the full event as done in gadgets/trace_open/test/unit/trace_open_test.go by using `utilstest.ExpectOneEvent` and similar helpers.

Shaheer-Ahmd: oh thanks, that would be very helpful (I was unaware of `utilstest.ExpectOneEvent` previously). 
Incase of ephemeral port, I would have to manually see which port is selected by the os though and then would have to pass it in the event.

gadgets/trace_bind/test/unit/trace_bind_test.go
Shaheer-Ahmd: Initially, I was going with `net.Listen` but due to go's internal logic there were multiple bind events being emitted for some reason. Therefore, I went with the `netcat` command.
Now, as you suggested, I have changed it to `syscall.Socket` to open a socket and `syscall.Bind` to bind to it. My only concern now is that the code has become cluttered to handle the boilerplate requirements whereas `nc` was much cleaner. What do you think?

gadgets/trace_bind/test/unit/trace_bind_test.go
mauriciovasquezbernal: You're right. Btw, you can base these tests on pkg/gadgets/trace/bind/tracer/tracer_test.go, there we have almost all logic needed here.

gadgets/trace_bind/test/unit/trace_bind_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+type testCase struct {
+	name    string
+	port    uint16
+	network string
+	version uint8
+	addr    string
+}
+
+func TestTraceBind(t *testing.T) {
```
burak-ok, line 46: A `gadgettesting.InitUnitTest(t)` is missing remove the memlock on some kernels

gadgets/trace_bind/test/unit/trace_bind_test.go
mauriciovasquezbernal: I think it's better not to rely on external binaries, even if it means to keep some boiler plate code.

mauriciovasquezbernal (approved):
LGTM from code inspection.

gadgets/trace_bind/test/unit/trace_bind_test.go
burak-ok: The import for this seems to be missing too:
```gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"```

mauriciovasquezbernal:
@Shaheer-Ahmd can you please update the commit message and use something more explicit than "add diverse tests"? something like "gadgets/trace_bind: add unit tests" will work perfectly.

Shaheer-Ahmd:
thanks for pointing the naming issue. I will try to rename the existing commits aswell.

burak-ok (approved):
LGTM, thanks

PR #4415, opened 2025-05-02
Remove the targ_pid from gadgets

burak-ok:
The target is already there for every gadget and filtering is done through gadget_should_discard_data_current


mauriciovasquezbernal (approved):
Good catch. I tested the top_tcp one and worked fine. The deadlock looks also fine from code inspection.

Issue #4367, opened 2025-04-23
ci: TestRealKAllSymsParsing/symbol_from_veth_kmod is failing


mauriciovasquezbernal:
FAIL: TestRealKAllSymsParsing/symbol_from_veth_kmod is failing a lot:

``` 
2025-04-23T20:37:38.4522945Z     kallsyms_test.go:227: 
2025-04-23T20:37:38.4523481Z         	Error Trace:	/home/runner/work/inspektor-gadget/inspektor-gadget/pkg/kallsyms/kallsyms_test.go:227
2025-04-23T20:37:38.4523621Z         	Error:      	Not equal: 
2025-04-23T20:37:38.4523826Z         	            	expected: true
2025-04-23T20:37:38.4524029Z         	            	actual  : false
2025-04-23T20:37:38.4524288Z         	Test:       	TestRealKAllSymsParsing/symbol_from_veth_kmod
2025-04-23T20:37:38.4524623Z         	Messages:   	Symbol veth_init existence check should match expected value
2025-04-23T20:37:38.4524739Z     kallsyms_test.go:232: 
2025-04-23T20:37:38.4525243Z         	Error Trace:	/home/runner/work/inspektor-gadget/inspektor-gadget/pkg/kallsyms/kallsyms_test.go:232
2025-04-23T20:37:38.4525434Z         	Error:      	Received unexpected error:
2025-04-23T20:37:38.4525664Z         	            	file does not exist
2025-04-23T20:37:38.4525921Z         	Test:       	TestRealKAllSymsParsing/symbol_from_veth_kmod
2025-04-23T20:37:38.4526221Z         	Messages:   	Should successfully resolve existing symbol veth_init
2025-04-23T20:37:38.4526337Z     kallsyms_test.go:233: 
2025-04-23T20:37:38.4526838Z         	Error Trace:	/home/runner/work/inspektor-gadget/inspektor-gadget/pkg/kallsyms/kallsyms_test.go:233
2025-04-23T20:37:38.4526995Z         	Error:      	Should not be: 0x0
2025-04-23T20:37:38.4527246Z         	Test:       	TestRealKAllSymsParsing/symbol_from_veth_kmod
2025-04-23T20:37:38.4527562Z         	Messages:   	Address for existing symbol veth_init should not be zero
2025-04-23T20:37:38.4527674Z     kallsyms_test.go:234: 
[...]

PR #4429, opened 2025-05-06
pkg/kallsyms: re-enable test for kernel module symbol

alban:
Kallsyms test on symbol veth_init failed a lot, so it was disabled.

I believe this is due to using a function declared with "__init" because the memory holding the code can be lazily reclaimed when the kernel needs more memory,

This patch re-enables the test but uses a different symbol that does not use "__init".

Fixes: d25e2295d1e6 ("pkg/kallsyms: disable failing test")

Fixes: #4367

## Testing done

```
$ go test -exec sudo -v ./pkg/kallsyms/...
...
--- PASS: TestRealKAllSymsParsing (3.67s)
    --- PASS: TestRealKAllSymsParsing/simple_symbol1 (0.46s)
    --- PASS: TestRealKAllSymsParsing/simple_symbol2 (0.47s)
    --- PASS: TestRealKAllSymsParsing/symbol_from_veth_kmod (0.49s)
    --- PASS: TestRealKAllSymsParsing/symbol_from_netem_kmod (0.47s)
    --- PASS: TestRealKAllSymsParsing/symbol_from_overlay_kmod (0.47s)
    --- PASS: TestRealKAllSymsParsing/nonexistent_symbol (0.46s)
```

### DIFF
--- a/pkg/kallsyms/kallsyms_test.go
+++ b/pkg/kallsyms/kallsyms_test.go
@@ -187,14 +187,12 @@ func TestRealKAllSymsParsing(t *testing.T) {
 			expectedKmod:   "",
 			expectedExists: true,
 		},
-		// This test is failing a lot, disable it until we understand why.
-		// See https://github.com/inspektor-gadget/inspektor-gadget/issues/4367
-		//{
-		//	name:           "symbol_from_veth_kmod",
-		//	symbol:         "veth_init",
-		//	expectedKmod:   "veth",
-		//	expectedExists: true,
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal (approved):
LGTM. Thanks Alban for handling this!

burak-ok (approved):
Thanks for fixing and renabling this.
Just as a precaution, could you also restart _only_ the unit test step of the CI to check if it still flakes

PR #4430, opened 2025-05-06
ci: Remove unit tests from integration test needs.

eiffel-fl:
The reasoning behind this first was that unit tests were short, so no need to run the longer integration tests if the unit tests already failed. But now, unit tests are long too, so let's have everything run in parallel.

Suggested-by: Mauricio Vásquez <>


mauriciovasquezbernal (approved):
Thanks for handling it. LGTM once the CI is green.

Issue #3938, opened 2025-01-24
[RFE] Allow "doesn't start with foo|bar|baz" filter in trace gadgets


johananl:
## Current situation

When specifying filters such as `-F 'runtime.containerImageName:~^(?!(foo|bar|baz)).*$'` (in trace_exec but likely in other gadgets too), the following error is given:

```
Error: setting filters: invalid filter "runtime.containerImageName:~^(?!(foo|bar|baz)).*$": compiling regular expression "^(?!(foo|bar|baz)).*$": error parsing regexp: invalid or unsupported Perl syntax: `(?!`
```

That's because the Go regex engine doesn't support negative lookahead (`?!`). It would be nice to have some way of specifying "field X doesn't start with any of the following".

It's very likely that I can work around this problem with a complex regex, however I wonder if there is a better way to get the same result.

## Impact

I want to filter out events from containers whose image is in a predefined set. I can't.

## Ideal future situation

I want to filter out events from containers whose image is in a predefined set without writing very complex regex.

## Implementation options

I'm not sure what would be a good way to go about this. Some thoughts though:

- Maybe new operators such as "startswith" and "endswith" could help (But then how would we negate? A generic "not" operator?).
- Maybe list operators such as "in" and "not in" can help (maybe an "in" operator is enough if we add a generic "not" operator).

We should evaluate carefully how complex we want the filters to get. At some point this will start to look like a DSL...

## Additional information

The Falco project has a flexible rules mechanism which might provide some ideas: https://falco.org/docs/concepts/rules/conditions/#operators

blanquicet:
Hi @johananl, I think this is what you're looking for:

```
ig run trace_exec:latest -F 'runtime.containerImageName!~^(debian|ubuntu).*$'
```

johananl:
Yes, I've found out about negative regex in the meantime. Thanks.

PR #4433, opened 2025-05-06
pkg/operators/filter: Update documentation to include negative regex

blanquicet:
# Update filter op's documentation

Include negative regex in the `filter`'s param documentation.

Related #3938

### DIFF
--- a/pkg/operators/filter/filter.go
+++ b/pkg/operators/filter/filter.go
@@ -71,8 +71,11 @@ func (f *filterOperator) InstanceParams() api.Params {
     field<=value     - matches, if the content of field is less than or equal to the value
     field<value      - matches, if the content of field is less than the value
     field~value      - matches, if the content of field matches the regular expression 'value'
+    field!~value     - matches, if the content of field does not match the regular expression 'value'
                  see [https://github.com/google/re2/wiki/Syntax] for more information on the syntax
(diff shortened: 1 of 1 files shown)


pkg/operators/filter/filter.go
```diff
@@ -71,8 +71,11 @@ func (f *filterOperator) InstanceParams() api.Params {
     field<=value     - matches, if the content of field is less than or equal to the value
     field<value      - matches, if the content of field is less than the value
     field~value      - matches, if the content of field matches the regular expression 'value'
+    field!~value     - matches, if the content of field does not match the regular expression 'value'
                  see [https://github.com/google/re2/wiki/Syntax] for more information on the syntax
   Multiple filters can be combined using a comma: field1==value1,field2==value2
+  It is recommended to use single quotes to escape the filter string, specifically if using regular expressions.
```
mauriciovasquezbernal, line 77:
```suggestion
  It is recommended to use single quotes to escape the filter string, especially if using regular expressions.
```

?

mauriciovasquezbernal (approved):
LGTM. Please also update docs/gadgets/switching_to_image_based_gadgets.mdx with this change.

pkg/operators/filter/filter.go
blanquicet: I thought "especially" but I wrote "specifically" 😅

PR #4438, opened 2025-05-07
gadgets/bpfstats: Do not require --enable-bpfstats

mauriciovasquezbernal:
The current implementation of the bpfstats gadget requires the daemon to be deployed with the --enable-bpfstats parameter to enable eBPF stats collection in the kernel. This UX is bad as users will need to enable it from the beginning, having a performance issue for all gadgets that are run, and by not being able to use this functionality on an existing IG deployment without having to restart it.

This PR changes the approach to enable the stats only when the bpfstats gadget is run, which is more practical as users don't need to change any configuration at deploy time and the stats collection is only enabled while the bpfstats gadget runs.

By changing this, we also fix an issue where stats collection wasn't disabled after running Inspektor Gadget.

Ref #4402

### DIFF
--- a/pkg/operators/ebpf/ebpf.go
+++ b/pkg/operators/ebpf/ebpf.go
@@ -67,7 +67,7 @@ const (
 	AnnotationFlushOnStop = "ebpf.map.flush-on-stop"
 )
 
-type gadgetObjs struct {
+type gadgetObjects struct {
 	programIDs []ebpf.ProgramID
 	mapIDs     []ebpf.MapID
 }
@@ -75,7 +75,7 @@ type gadgetObjs struct {
 // ebpfOperator reads ebpf programs from OCI images and runs them
 type ebpfOperator struct {
 	mu         sync.Mutex
-	gadgetObjs map[operators.GadgetContext]gadgetObjs
+	gadgetObjs map[operators.GadgetContext]gadgetObjects
 }
 
 func (o *ebpfOperator) Name() string {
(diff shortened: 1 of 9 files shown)


pkg/operators/ebpf/ebpf.go
```diff
@@ -679,35 +679,36 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
-		i.bpfOperator.gadgetObjs[gadgetCtx] = gadgetObjs
-		i.bpfOperator.mu.Unlock()
+
+		id, _ := info.ID()
+		gadgetObjs.mapIDs = append(gadgetObjs.mapIDs, id)
+	}
+	i.bpfOperator.mu.Lock()
+	if i.bpfOperator.gadgetObjs == nil {
+		i.bpfOperator.gadgetObjs = make(map[operators.GadgetContext]gadgetObjects)
 	}
```
flyth, line 709: Should we move this to the init code where the other maps are initialized as well?

flyth (approved):
LGTM from code inspection - this is a much better solution, thanks!

mqasimsarfraz (approved):

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: If I move this to `func (o *ebpfOperator) Init(params *params.Params)` then a lot of tests and applications using the Go API will crash as they don't call that function. I could move it to `InstantiateImageOperator()` but I don't see any different with having it on `Start()`.

flyth: Makes sense - nevermind, then.

mauriciovasquezbernal:
Failing test is not related to this PR.

Issue #4321, opened 2025-04-10
gadgets/bpfstats: Provide human readable field for runtime and mapmemory


mauriciovasquezbernal:
Inspektor Gadget supports printing a human readable version of bytes (#3392) and time (#3393) and many gadgets use it already:

- https://github.com/inspektor-gadget/inspektor-gadget/blob/e75442de689c3844d69a9f0bd9c1a32577667ab2/gadgets/top_file/program.bpf.c#L38
- https://github.com/inspektor-gadget/inspektor-gadget/blob/509e06dbf21af24e9c62ecf42cd01cb9cb1480b2/gadgets/top_tcp/program.bpf.c#L33
- https://github.com/inspektor-gadget/inspektor-gadget/blob/6a2c5e949cbfc464b40d7b61196c9f2e1cffffaf/gadgets/trace_dns/program.bpf.c#L88
- https://github.com/inspektor-gadget/inspektor-gadget/blob/ed78d40e26626b17c41be8dc38a416bf4958704d/gadgets/trace_ssl/program.bpf.c#L54

This doesn't work for the bpfstats gadget as those values are coming from the ebpf operator and not from eBPF code:

https://github.com/inspektor-gadget/inspektor-gadget/blob/d263b7467faf268900d5e3059d7a11ed3ded4f0d/pkg/operators/ebpf/stats.go#L182-L190

https://github.com/inspektor-gadget/inspektor-gadget/blob/d263b7467faf268900d5e3059d7a11ed3ded4f0d/pkg/operators/ebpf/stats.go#L205-L213

Tagging those fields with `"type:gadget_bytes"` and `"type:gadget_duration"` should make the trick, however it didn't work for me when implementing the gadget.

Issue #4288, opened 2025-04-04
[RFE] Register structure for container datasource


eiffel-fl:
## Current situation

The container datasource only publishes few information about container, like cgroup ID and mount ns ID.
To add more, one should add another field:
https://github.com/inspektor-gadget/inspektor-gadget/pull/3957/commits/0b836c714ce3963e74ead4299ca67c9821a1a765#diff-5e4395e2fe0662f76e05b22ad45362677942ed833f9d00029827ac16b13c20acR324-R328
https://github.com/inspektor-gadget/inspektor-gadget/pull/3957/commits/0b836c714ce3963e74ead4299ca67c9821a1a765#diff-5e4395e2fe0662f76e05b22ad45362677942ed833f9d00029827ac16b13c20acR342-R345

## Impact

Adding other field is not flexible and hard to maintain.

## Ideal future situation

A structure should be used.

## Implementation options

https://github.com/inspektor-gadget/inspektor-gadget/pull/3957#discussion_r1951473473

PR #4446, opened 2025-05-09
introduce process operator and top_process gadget

flyth:
# Introduce process operator and top_process gadget

This pull request introduces a new "process" operator, which emits events about running processes, including CPU and memory usage.

A gadget leveraging the internal datasource created by the process operator has also been added.

```shell
$ ig run top_process:mine --verify-image=false --sort -memoryRSS

RUNTIME.CONTAINERNAME                   PID         PPID        COMMAND               CPUUSAGE              MEMORYRSS             MEMORYVIRTUAL         THREADCOUNT STATE                USERNAME             STARTTIME            STARTTIMESTR         MOUNTNSID
                                        29766       29765       ig                    0                     132177920             2132725760            10          S                    0                    272548               2025-05-09T12:26:43… 4026531841
                                        744         1           ig                    0                     126717952             1903644672            7           S                    0                    911                  2025-05-09T11:41:27… 4026531841
                                        10215       1           fwupd                 0                     72523776              451932160             5           S                    0                    231518               2025-05-09T12:19:53… 4026532335
                                        1144        1           dockerd               0                     59121664              2366349312            14          S                    0                    942                  2025-05-09T11:41:27… 4026531841
                                        766         1           containerd            0                     27246592              1906266112            9           S                    0                    913                  2025-05-09T11:41:27… 4026531841
                                        9784        9371        docker                0                     26632192              1966112768            10          S                    1000                 188454               2025-05-09T12:12:42… 4026531841
...
test-trace-tcpretrans                   29760       9845        openssl               10001                 4939776               6651904               1           R                    0                    271102               2025-05-09T12:26:29… 4026532270
...
```


pkg/operators/process/process.go
```diff
@@ -0,0 +1,649 @@
...
+
+	// Read /proc directory to get all processes
+	procDir, err := os.Open("/proc")
+	if err != nil {
+		return fmt.Errorf("opening /proc directory: %w", err)
+	}
+	defer procDir.Close()
+
+	// Get all process directories (named by PID)
+	entries, err := procDir.Readdir(-1)
```
alban, line 381: You can use helper function:
```
dirEntries, err := os.ReadDir(host.HostProcFs)
```

This is the code pattern in the project:
```bash
$ git grep os.ReadDir.host.HostProcFs
pkg/container-hook/tracer.go:413:                       dirEntries, err := os.ReadDir(host.HostProcFs)
pkg/gadgets/snapshot/process/tracer/tracer.go:200:      items, err := os.ReadDir(host.HostProcFs)
pkg/gadgets/top/ebpf/tracer/tracer.go:167:      processes, err := os.ReadDir(host.HostProcFs)
```

pkg/operators/process/process.go
```diff
@@ -0,0 +1,649 @@
...
+	needStatus := p.enabledFields[fieldPPID] || p.enabledFields[fieldCommand] ||
+		p.enabledFields[fieldState] || p.enabledFields[fieldUid] ||
+		p.enabledFields[fieldMemoryRSS] || p.enabledFields[fieldMemoryVirtual] ||
+		p.enabledFields[fieldThreadCount]
+
+	needStat := p.enabledFields[fieldCPUUsage] || p.enabledFields[fieldStartTime]
+
+	// Read process status if needed
+	if needStatus {
+		statusFile := fmt.Sprintf("/proc/%d/status", pid)
```
alban, line 510: `host.HostProcFs`

pkg/operators/process/process.go
```diff
@@ -0,0 +1,649 @@
...
+				continue
+			}
+
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+
+			switch key {
+			case "Name":
+				if p.enabledFields[fieldCommand] {
+					procInfo.Command = value
```
alban, line 531: Spaces are valid in the command name:
```
$ printf '  XX  ' > /proc/$$/comm
$ grep ^Name: /proc/$$/status
Name:	  XX  
$ grep ^Name: /proc/$$/status|sed 's/ /_/g'
Name:	__XX__
```

Note that the comm in /proc/pid/status is escaped with `ESCAPE_SPACE | ESCAPE_SPECIAL`:
https://github.com/torvalds/linux/blob/v6.12/fs/proc/array.c#L445
See escape rules:
https://github.com/torvalds/linux/blob/v6.12/lib/string_helpers.c#L528-L538

pkg/operators/process/process.go
```diff
@@ -0,0 +1,649 @@
...
+		statusData, err := os.ReadFile(statusFile)
+		if err != nil {
+			return procInfo, fmt.Errorf("reading status file: %w", err)
+		}
+
+		// Parse status file
+		scanner := bufio.NewScanner(strings.NewReader(string(statusData)))
+		for scanner.Scan() {
+			line := scanner.Text()
+			parts := strings.SplitN(line, ":", 2)
```
alban, line 520:
```suggestion
			parts := strings.SplitN(line, ":\t", 2)
```

`\t` is always added: see for example:
https://github.com/torvalds/linux/blob/v6.12/fs/proc/array.c#L444-L445
https://github.com/torvalds/linux/blob/v6.12/fs/proc/array.c#L184-L197

pkg/operators/process/process.go
```diff
@@ -0,0 +1,649 @@
...
+			if len(parts) != 2 {
+				continue
+			}
+
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
```
alban, line 526: strings.TrimSpace should be removed:
* Key: the kernel does not add spaces
* Value: spaces can be valid part of the field, see below

pkg/operators/process/process.go
```diff
@@ -0,0 +1,778 @@
...
+	restData := statData[commEndIdx+1:]
+	fields := strings.Fields(restData)
+
+	// We need at least 21 fields after the comm field to access field 22
+	if len(fields) < 21 {
+		return nil, fmt.Errorf("invalid stat file format: not enough fields")
+	}
+
+	// Create a result array with only the fields we need
+	// This is more efficient than creating a full array with all fields
```
alban, line 762: Why is it more efficient? I would imagine the CPU usage to parse numbers is not too bad, but the dynamic allocations and the branches are more of concern.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,778 @@
...
+			builder.WriteByte(cmd[i])
+		}
+	}
+
+	return builder.String()
+}
+
+// parseStatFile parses the /proc/[pid]/stat file, handling the comm field correctly
+// The comm field is enclosed in parentheses and may contain spaces
+func parseStatFile(statData string) ([]string, error) {
```
alban, line 722: Since `os.ReadFile` gives a `[]byte`,this function could work on a `[]byte` to avoid memory copies:
```suggestion
func parseStatFile(statData []byte) ([]string, error) {
```
The [bytes package](https://pkg.go.dev/bytes) has similar functions for Index, LastIndex, TrimLeft, etc.

alban (commented):
Some questions about performance. It would be good to have benchmark on parsing procfs.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,778 @@
...
+			}
+
+			key := parts[0]
+			value := parts[1]
+
+			switch key {
+			case "Name":
+				if p.enabledFields[fieldCommand] {
+					procInfo.Command = unescapeCommand(value)
+				}
```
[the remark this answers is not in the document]
flyth: I'm not super sure - I can imagine us extending this a lot. But if a gadget is only using 1-2 fields of it (unlike the top_process one), I think it can make a difference.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,778 @@
...
+	pidStr := strings.TrimSpace(statData[:commStartIdx])
+
+	// Extract the comm (field 2)
+	comm := statData[commStartIdx+1 : commEndIdx]
+
+	// Extract the rest of the fields
+	// We only need a few specific fields, so we'll parse them directly
+	// without creating a full array of all fields
+	restData := statData[commEndIdx+1:]
+	fields := strings.Fields(restData)
```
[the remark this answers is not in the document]
flyth: Yeah, I think it makes sense. However, changing it to that makes the code a lot harder to read/understand, which is why I didn't do it now. I _think_ another optimization would be to multithread reading the proc info, but that also increases complexity a bit.

If you think we should have those optimizations at day one, I'm happy to implement them. Since this is a high priority item, I don't want to block it longer than necessary.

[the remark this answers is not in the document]
alban: It is not necessary to optimize in this PR because this feature is not used or not enabled by default on existing gadgets.

pkg/operators/process/process.go
flyth: More efficient than filling out all of them. I think changing everything to use []byte instead of string should have already improved the situation.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+		}
+
+		// Parse stat file - we need to handle the comm field specially
+		// The comm field is enclosed in parentheses and may contain spaces
+		statFields := make([][]byte, 22)
+		err = parseStatFile(statData, statFields)
+		if err != nil {
+			return procInfo, fmt.Errorf("parsing stat file: %w", err)
+		}
+		if len(statFields) >= 22 {
```
mauriciovasquezbernal, line 624: Why do we need this check? Can't it be less than 22 is we're using 22 on L619?

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+	// Extract the rest of the fields
+	// We only need a few specific fields, so we'll parse them directly
+	// without creating a full array of all fields
+	restData := statData[commEndIdx+1:]
+	fields := bytes.Fields(restData)
+
+	// We need at least 21 fields after the comm field to access field 22
+	if len(fields) < 21 {
+		return fmt.Errorf("invalid stat file format: not enough fields")
+	}
```
mauriciovasquezbernal, line 761: Also, do we really need this check? Is it too bad to make it crash if the passed array isn't big enough?

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+	configKeyInterval = "internalDataSources.processes.interval"
+	configKeyFields   = "internalDataSources.processes.fields"
+
+	// Default values
+	defaultInterval = 60 * time.Second
+
+	// Field names
+	fieldPID           = "pid"
+	fieldPPID          = "ppid"
+	fieldCommand       = "command"
```
mauriciovasquezbernal, line 58: We're using `comm` everywhere else.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+		if err != nil {
+			return procInfo, fmt.Errorf("reading stat file for PID %d: %w", pid, err)
+		}
+		if len(statData) == 0 {
+			return procInfo, fmt.Errorf("empty stat file for PID %d", pid)
+		}
+
+		// Parse stat file - we need to handle the comm field specially
+		// The comm field is enclosed in parentheses and may contain spaces
+		statFields := make([][]byte, 22)
```
mauriciovasquezbernal, line 619: Some questions about this and the whole parseStatFile logic:
- Why 22? 
- Can't `parseStatFile` return something more structured?

pkg/operators/localmanager/localmanager.go
```diff
@@ -342,99 +342,103 @@ func (l *localManagerTrace) PreGadgetRun() error {
...
 	}
 
-	// If --host is set, we do not want to create the below map because we do not
-	// want any filtering.
-	if setter, ok := l.gadgetInstance.(MountNsMapSetter); ok {
-		if !host {
-			if l.manager.igManager == nil {
-				return fmt.Errorf("container-collection isn't available")
-			}
+	if l.gadgetInstance != nil {
```
mauriciovasquezbernal, line 345: nit: use an early return? I also think the operations above (to the the contianerSelector and so on), should be avoided if this is nil, they aren't used in that case.

pkg/operators/kubemanager/kubemanager.go
```diff
@@ -244,81 +244,85 @@ func (m *KubeManagerInstance) PreGadgetRun() error {
 		m.params.Get(ParamAllNamespaces).AsBool(),
 	)
 
-	if setter, ok := m.gadgetInstance.(MountNsMapSetter); ok {
-		err := m.manager.gadgetTracerManager.AddTracer(m.id, containerSelector)
-		if err != nil {
-			return fmt.Errorf("adding tracer: %w", err)
-		}
+	if m.gadgetInstance != 0 {
```
mauriciovasquezbernal, line 247: Why 0? The localmanager operator is using `nil`. Also same comment about using an early return.

pkg/datasource/columns.go
```diff
@@ -152,6 +152,11 @@ func (ds *dataSource) Columns() (*columns.Columns[DataTuple], error) {
 				}
 			case metadatav1.ColumnsAliasAnnotation:
 				attributes.Alias = v
+			case metadatav1.ColumnsPrecisionAnnotation:
+				attributes.Precision, err = strconv.Atoi(v)
+				if err != nil {
+					return nil, fmt.Errorf("reading precision cor column %q: %w", f.Name, err)
```
mauriciovasquezbernal, line 158: cor -> for

gadgets/top_process/gadget.yaml
```diff
@@ -0,0 +1,30 @@
...
+          columns.alignment: right
+      memoryVirtual:
+        annotations:
+          columns.alignment: right
+      threadCount:
+        annotations:
+          columns.alignment: right
+      startTime:
+        annotations:
+          columns.alignment: right
```
mauriciovasquezbernal, line 30: Shouldn't it be configured by the operator itself?

pkg/operators/process/process.go
mauriciovasquezbernal, line 1: For the future: This will benefit a lot from #4288

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+	}
+
+	instance := &processOperatorInstance{
+		interval:      interval,
+		enabledFields: enabledFields,
+		done:          make(chan struct{}),
+		dataSource:    ds,
+	}
+
+	// Add fields to the data source based on enabled fields
```
mauriciovasquezbernal, line 179: nit: Add descriptions and annotations to all the fields.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+
+	// Get the interval from config or use default
+	interval := viperConfig.GetDuration(configKeyInterval)
+	if interval <= 0 {
+		interval = defaultInterval
+		gadgetCtx.Logger().Debugf("Using default interval: %s", interval)
+	}
+
+	// Get the enabled fields from config
+	enabledFields := make(map[string]bool)
```
mauriciovasquezbernal, line 136: What about this alternative approach?
1. Do not use this map, base all the logic on the fields variable below
2. The logic to set the default could be handled by something like

```go
if len(fields) == 0 {
  fields = []string{fieldPid, fieldPPID, etc..}
```

3. Add some logic to ensure pid and mountns is always present
4. The logic to create the datasources could be something like

```go
for _, fieldName := range fields {
select {
case fieldPid:
 instance.pidField := ...
case fieldPpid:
...
}
}
```

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("adding startTime field: %w", err)
+		}
+
+		// Add the formatted start time field
+		instance.startTimeStrField, err = ds.AddField(fieldStartTimeStr, api.Kind_String)
+		if err != nil {
+			return nil, fmt.Errorf("adding startTimeStr field: %w", err)
+		}
+	}
```
mauriciovasquezbernal, line 254: I'm wondering if this can be automatically handled by the formatters operator. I tried something similar on #4321 but it didn't work, perhaps you can give it a try.

mauriciovasquezbernal (commented):
The general idea looks good to me! I did a pass looking at implementation details and so on. I'll make another later on trying user experience, corners cases and so on.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-service/api"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/params"
+)
+
+const (
+	Name = "process"
```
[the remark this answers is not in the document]
flyth: I actually think it's better to have it exported to have access to it when using the Go API. I know that after instantiating you've got the `Name()` as well, but there might be situations where you need it beforehand. Opinion about it is not too strong, though :)

pkg/operators/process/process.go
```diff
@@ -0,0 +1,777 @@
...
+					if prevTime, ok := p.lastCPUTimes[pid]; ok && timeDelta > 0 {
+						// CPU usage is the difference in CPU time (in jiffies) divided by the elapsed time (in seconds)
+						// We need to convert jiffies to seconds by dividing by clockTicksPerSecond
+						cpuTimeDiff := float64(totalTime - prevTime)
+						if cpuTimeDiff < 0 {
+							// Handle case where process was restarted with same PID
+							cpuTimeDiff = float64(totalTime)
+						}
+						cpuUsage := cpuTimeDiff / clockTicksPerSecond / timeDelta * 100.0
+						procInfo.CPUUsage = cpuUsage
```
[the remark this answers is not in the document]
flyth: I think it's fine without, but I'll add some for readability.

pkg/operators/process/process.go
flyth: We don't anymore - I changed the logic a bit and this is a leftover.

pkg/operators/process/process.go
flyth: Yeah, I think it's bad to crash on a gadget level (which this context would run in) at some random point on runtime. Controlled crashes should IMHO only happen on startup or if we know the system is really in a bad state anyways (this could happen due to an update that changed the file layout, although I'm sure that this API is pretty stable).

pkg/operators/process/process.go
flyth: 22 (so actually field 21) is just the last field we currently need.

a) I tried to optimize a bit without losing readability.
b) I saw multiple documentations just saying "you'll find X in the stat file as 16th field", for example. So this tries to help a bit in adding fields later on.

pkg/operators/localmanager/localmanager.go
flyth: I'll extract that to a new function - an early return might confuse if stuff is added at the end of the function otherwise.

pkg/operators/process/process.go
flyth: I tried it and it kind of worked, but I'm not sure if we shouldn't keep the numeric value as is (e.g. time since boot) and just add the date as usual.

pkg/operators/process/process.go
flyth: Added description + templates.

flyth:
I have implemented some optimizations to reduce CPU time / allocations for the hot paths at the cost of some readability. I hope it's still somewhat clear.

The aggregation of process information will now be multithreaded as well.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,778 @@
...
+		return fmt.Errorf("emitting packet array: %w", err)
+	}
+	// Set to nil so it's not released again in the defer function
+	packetArray = nil
+
+	return nil
+}
+
+// getProcessInfo reads process information from /proc/{pid}
+func (p *processOperatorInstance) getProcessInfo(pid int, timeDelta float64) (ProcessInfo, error) {
```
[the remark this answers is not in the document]
flyth: I extracted the functions into an own package, but for now made it a sub-package. If we should use things in other contexts, we can move it out later on.

pkg/operators/process/helpers/processhelpers.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		fields := bytes.SplitN(scanner.Bytes(), []byte(":"), 2)
+		if len(fields) < 2 {
+			continue
+		}
+		if !bytes.HasPrefix(fields[0], []byte("MemTotal")) {
+			continue
+		}
```
mauriciovasquezbernal, line 158: Can't we just use 

```go
memBytes, ok := bytes.CutPrefix(scanner.Bytes(), []byte("MemTotal"))`
if !ok {
    continue
}
```

pkg/operators/process/helpers/processhelpers.go
```diff
@@ -0,0 +1,393 @@
...
+				break
+			}
+			if e != nil {
+				s.Close()
+				bufPool.Put(xbuf)
+				return pi, fmt.Errorf("read stat %d: %w", pid, e)
+			}
+		}
+		s.Close()
+		data := (*xbuf)[:n]
```
mauriciovasquezbernal, line 287: nit: can we move this to a helper to avoid code duplication?

pkg/operators/process/process.go
```diff
@@ -0,0 +1,642 @@
...
+	p.bootTime = bootTime
+
+	// Start the process monitoring goroutine
+	p.wg.Add(1)
+	go p.monitorProcesses(gadgetCtx)
+
+	return nil
+}
+
+func getBootTime(gadgetCtx operators.GadgetContext) time.Time {
```
mauriciovasquezbernal, line 365: nit: perhaps better to return an error too and log in the caller.

gadgets/top_process/README.mdx
```diff
@@ -0,0 +1,77 @@
...
+- Mount namespace ID
+
+It also adds environment enrichments (container/k8s information), if possible.
+
+You can sort the output by any field, for example by CPU usage:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/top_process:%IG_TAG% --sort -cpuUsage --max-entries 12
```
mauriciovasquezbernal, line 53: any idea why the mountnsid is shown? Shouldn't it be hidden by default? Smae for starttime and starttimestr, can we hide one of them?

pkg/operators/process/helpers/processhelpers_test.go
```diff
@@ -0,0 +1,182 @@
...
+	pid := os.Getpid()
+	proc, err := processhelpers.GetProcessInfo(pid, 0, allOptions{
+		cpuCount: 1,
+	})
+	require.NoError(t, err)
+
+	assert.Equal(t, pid, proc.PID)
+	assert.Greater(t, proc.PPID, 0)
+	assert.NotEmpty(t, proc.Command)
+	assert.Zero(t, proc.CPUUsage)
```
mauriciovasquezbernal, line 106: Why CPU usage is 0 for this process?

pkg/operators/process/helpers/processhelpers_test.go
```diff
@@ -0,0 +1,182 @@
...
+func BenchmarkSingle(b *testing.B) {
+	pid := os.Getpid()
+	opts := allOptions{cpuCount: 1}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		processhelpers.GetProcessInfo(pid, 0, opts)
+	}
+}
+
+func BenchmarkAllProcessesSeq(b *testing.B) {
```
mauriciovasquezbernal, line 124: Do we want to merge these?

mauriciovasquezbernal (commented):
I did an additional pass. I need to test is a bit before giving final approval.

pkg/operators/process/helpers/processhelpers_test.go
flyth: It can only get the CPU usage on second run (where it has the delta and can calculate it from that).

docs/spec/operators/process.md
```diff
@@ -0,0 +1,179 @@
...
+- `ppid` - Parent Process ID
+- `command` - Command name
+- `cpuUsage` - CPU usage percentage
+- `memoryRSS` - Resident Set Size
+- `memoryVirtual` - Virtual memory size
+- `threadCount` - Number of threads
+- `state` - Process state
+- `uid` - Process owner UID
+- `startTime` - Process start time (clock ticks since system boot)
+- `startTimeStr` - Process start time as a formatted date-time string (automatically included when startTime is enabled)
```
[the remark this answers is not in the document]
flyth: I think I answered to a similar comment already (somewhere): I think it's useful to provide both the relative time to boot as well as the absolute start time. If we align it with how we usually handle timestamps, the time relative to boot would be lost.

pkg/operators/process/process.go
flyth: I changed the implementation to your suggested one, thanks!

pkg/operators/process/process.go
```diff
@@ -0,0 +1,610 @@
...
+				"columns.alignment":              "right",
+				metadatav1.DescriptionAnnotation: "The CPU usage of the process as a percentage.",
+			}))
+			if err != nil {
+				return nil, fmt.Errorf("adding cpuUsage field: %w", err)
+			}
+			requireCPUInfo = true
+		case fieldCPUUsageRelative:
+			instance.cpuRelativeField, err = ds.AddField(fieldCPUUsageRelative, api.Kind_Float64, datasource.WithAnnotations(map[string]string{
+				"columns.precision":              "1",
```
mauriciovasquezbernal, line 191: nit: used constants on metadatav1 too for these ones

mauriciovasquezbernal (approved):
just a final minor nit, but already LGTM!

pkg/operators/process/helpers/processhelpers.go
```diff
@@ -0,0 +1,387 @@
...
+		}
+		xbuf, n, err := readBytes(s)
+		if err != nil {
+			return pi, fmt.Errorf("reading stat file for pid %d: %w", pid, err)
+		}
+		s.Close()
+		data := (*xbuf)[:n]
+
+		o := bytes.IndexByte(data, '(')
+		c := bytes.LastIndexByte(data, ')')
```
alban, line 284: JFYI, I referenced this code in:
https://github.com/inspektor-gadget/inspektor-gadget/pull/4520#discussion_r2111569105

flyth: That was faster than expected ;) Moved the helpers out of the operators dir, directly into /pkg.

/cc @mqasimsarfraz

mqasimsarfraz:
It is missing entry in: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/Makefile#L23

blanquicet (commented):
In general, it looks great. Most of my comments are to improve documentation and UX. I have these main concerns I added as comments:

1. Why we introduced the `internalDataSources` mechanism instead of using annotations? -> We can merge this and continue the discussion. In any case, if we change it, it will not impact the gadget users.
2. Why is it printing host processes by default
3. `--containername` and k8s filters are not working.

2 and 3 are related and I think we can manage it in a follow-up PR and release an initial support now.

docs/spec/operators/process.md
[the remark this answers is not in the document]
flyth: Hmm, any suggestions? I think the descriptions are pretty on point, but am unsure about the names.

The difference is just that one is relative (to system boot) and one is absolute (date/time). Whenever we use `gadget_timestamp` it's actually the same: `bpf_ktime_get_boot_ns()` also returns a time relative to system boot, and we `magically` convert that to an absolute value without telling. However, by doing that, we're losing the boot time info, which is IMHO (especially here) valuable as well. Like said, you can easily see that a process was restarted / started later on, if the time since boot is > 5min, for example.

How about adding one more field, then? `timeSinceBoot` and then handle the timestamp like a `gadget_timestamp` - `startTime_raw` + `startTime`.

docs/spec/operators/process.md
```diff
@@ -0,0 +1,195 @@
...
+- `pid` - Process ID (always included)
+- `ppid` - Parent Process ID
+- `comm` - Command name
+- `cpuUsage` - CPU usage percentage
+- `cpuUsageRelative` - CPU usage percentage relative to number of CPUs available
+- `memoryRSS` - Resident Set Size
+- `memoryVirtual` - Virtual memory size
+- `memoryRelative` - Total memory used relative to available memory
+- `threadCount` - Number of threads
+- `state` - Process state
```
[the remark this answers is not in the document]
flyth: Added it in the details and a note here that it can be found below.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,610 @@
...
+		case fieldThreadCount:
+			instance.threadCountField, err = ds.AddField(fieldThreadCount, api.Kind_Int32, datasource.WithAnnotations(map[string]string{
+				metadatav1.ColumnsAlignmentAnnotation: "right",
+				metadatav1.DescriptionAnnotation:      "The number of threads used by the process.",
+			}))
+			if err != nil {
+				return nil, fmt.Errorf("adding threadCount field: %w", err)
+			}
+		case fieldState:
+			instance.stateField, err = ds.AddField(fieldState, api.Kind_String, datasource.WithAnnotations(map[string]string{
```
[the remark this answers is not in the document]
flyth: I'll adjust it to these values, but now thinking about it, I don't like it. `State` should be 1, not 5. Whether the interface wants to prefer showing the whole header should be up to the interface itself. The annotation should only represent the actual data (think about localization of the headers, for example). But let's not open that box now.

gadgets/top_process/README.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run top_process --sort -cpuUsage --max-entries 12
+        RUNTIME.CONTAINERNAME            PID COMM               CPUUSAGE CPUUSAGERELA…     MEMORYRSS MEMORYVIRTUAL MEMORYRELATI… THREADCOUNT STATE               UID STARTTIMESTR
+        test-container                   747 ig                      0.3           0.1      94101504    1979633664           2.3           8 S                     0 2025-05-20T2
+        test-container                 40192 ig                      0.3           0.1     134131712    2056871936           3.3           9 S                     0 2025-05-27T1
+        test-container                 32493 ig                      0.3           0.1     161894400    2061258752           3.9           8 S                     0 2025-05-24T1
+                                          98 ata_sff                 0.0           0.0             0             0           0.0           1 I                     0 2025-05-20T2
+                                          99 md                      0.0           0.0             0             0           0.0           1 I                     0 2025-05-20T2
```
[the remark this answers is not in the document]
flyth: Yes, filtering is currently not possible using those parameters, even though the params are introduced by the KubeManager/LocalManager operators. We should totally add support for it, but in a later PR.

gadgets/top_process/README.mdx
```diff
@@ -0,0 +1,77 @@
...
+                                                                                               99 md                      0.0           0.0             0             0           0.0           1 I                     0 2025-05-20T2
+                                                                                              817 sshd                    0.0           0.0       6160384      15507456           0.2           1 S                     0 2025-05-20T2
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run top_process --sort -cpuUsage --max-entries 12
+        RUNTIME.CONTAINERNAME            PID COMM               CPUUSAGE CPUUSAGERELA…     MEMORYRSS MEMORYVIRTUAL MEMORYRELATI… THREADCOUNT STATE               UID STARTTIMESTR
+        test-container                   747 ig                      0.3           0.1      94101504    1979633664           2.3           8 S                     0 2025-05-20T2
```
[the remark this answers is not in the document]
blanquicet: Isn't it done by default?

gadgets/top_process/README.mdx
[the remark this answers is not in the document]
blanquicet: Yes, sure. Work for a follow-up PR.

docs/spec/operators/process.md
```diff
@@ -0,0 +1,179 @@
...
+    - command
+    - pid
+    - ppid
+    - cpuUsage
+    - memoryRSS
+    - memoryVirtual
+    - threadCount
+    - state
+    - uid
+    - startTime
```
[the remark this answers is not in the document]
flyth: You're right - it makes more sense as you otherwise can't see exactly _where_ that data is coming from.

docs/spec/operators/process.md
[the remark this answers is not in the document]
flyth: The `startTime_raw` we usually have is a UNIX timestamp, so still absolute / not the same. So I would find it more confusing naming it the same way, when it's not the same.

pkg/operators/process/process.go
```diff
@@ -0,0 +1,617 @@
...
+				metadatav1.DescriptionAnnotation:     "The state of the process (e.g., \"R\" for running, \"S\" for sleeping, \"Z\" for zombie, etc.).",
+				metadatav1.ValueOneOfAnnotation:      "R,S,D,Z,T,t,W,X,x,K,W,P,I",
+				metadatav1.ColumnsMaxWidthAnnotation: "5",
+			}))
+			if err != nil {
+				return nil, fmt.Errorf("adding state field: %w", err)
+			}
+		case fieldUid:
+			instance.uidField, err = ds.AddField(fieldUid, api.Kind_Uint32, datasource.WithAnnotations(map[string]string{
```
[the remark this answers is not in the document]
flyth: Hmm, it should work like that, yes. At one point I did the same for the timestamp (before reverting to the clock ticks) and it worked. In that case I would hardcode those strings though instead of adding a dependency.

docs/spec/operators/process.md
```diff
@@ -0,0 +1,195 @@
...
+
+#### `threadCount`
+
+The number of threads in the process.
+
+Type: `int32`
+
+#### `state`
+
+The state of the process (e.g., "R" for running, "S" for sleeping, "Z" for zombie, etc.).
```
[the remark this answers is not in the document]
flyth: ffs 😄 copy/paste fail :/

pkg/operators/process/process.go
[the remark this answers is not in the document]
blanquicet: But, the new added field `user` is hidden and I wasn't able to show it by default even if I updated the gadget.yaml with:

```yaml
datasources:
  processes:
    annotations:
    fields:
      user:
        annotations:
          columns.hidden: false
```

docs/spec/operators/process.md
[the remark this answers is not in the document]
blanquicet: But `raw` stands only for the fact that it's not parsed but exactly how the system provides it. It is not necessarily associated to a specific format. Anyway, let's stay with what you have now, the description already mentions it is in "clock ticks" format.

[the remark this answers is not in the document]
flyth: Yeah, but it's not raw anymore after the operator accessed it, since it [recalculates even the `_raw` field](https://github.com/inspektor-gadget/inspektor-gadget/blob/20e682851c37c0528ac73beef81344c8d47cea25/pkg/operators/formatters/formatters.go#L340).

flyth:
Thanks for the reviews! Skipping the failing documentation check.

PR #4447, opened 2025-05-09
ci: Enable cache for building helper images

### LINKED ISSUE #4345: [RFE] Compiling a gadget is slow: ebpf-builder image is too big
## Description

To compile a gadget, ig pulls the ebpf-builder IG image. Usually only once a day, but during the next day or 2, I need to pull it again. Even if the only changes I did were in the `gadgets/my_gadget` directory.

The image currently is 2.5GB. I've just done a pull with docker to easily see the size when reporting, and:

```
$ docker images
REPOSITORY                              TAG       IMAGE ID       CREATED         SIZE
ghcr.io/inspektor-gadget/ebpf-builder   main      21d90d9cf9ac   9 days ago      2.5GB
```

This image has a lot of layers and is big, which makes it slow. This is an example of how much time it takes to pull from my office:

```
$ time IG="$PWD/../ig" make check_mount
Building check_mount
Pulling builder image ghcr.io/inspektor-gadget/ebpf-builder:main
main: Pulling from inspektor-gadget/ebpf-builder
23b7d26ef1d2: Pull complete 
07d1b5af933d: Pull complete 
1eb98adba0eb: Pull complete 
6eb08228d7a5: Pull complete 
f1d296901bdc: Pull complete 
cc508665364b: Pull complete 
4f4fb700ef54: Pull complete 
522358ac4428: Pull complete 
a5ae286bea94: Pull complete 
12478ecdd8ea: Pull complete 
c3d5af7476da: Pull complete 
b36df9a13ce4: Pull complete 
95538aced7d4: Pull complete 
4126f1676295: Pull complete 
Digest: sha256:849546bcee32e8329eb028dd25db1d3987a1c07da2a770451b5f2047ea901e4b
Status: Downloaded newer image for ghcr.io/inspektor-gadget/ebpf-builder:main
Successfully built ghcr.io/inspektor-gadget/gadget/check_mount:main@sha256:40e945b5456fc4d9c0aedb81dd1aa9d3d3fc2c63121b18aa6ffce23e4ec19f23

real	9m19.168s
user	0m0.068s
sys	0m0.141s
```

## Impact

Compiling is slow when you need to pull the ebpf-builder image.

## Expected behavior

Compiling when you have to pull the image takes less than 9 minutes, ideally less than one minute too.

mauriciovasquezbernal:
We aren't caching the Docker layers for helper images like the gadget-builder. It causes the following:
1. The building takes very long (~5mins)
2. A new built image has a totally different set of digest for its layers compared to one generated in a previous builder as our images don't have "reproducible builds" support.

The second point affects users compiling gadgets, as each time the image was changed upstream, they have to pull all the layers (hundreds of MBs).

This commit enables the cache fixing the issues described above. Now, also the logic to avoid building the image based on some file patterns is removed as we can always try to build it with a very minimum overhead.

--- 

The issue with #4345 was not only having a big image (which has been improved a lot in #4407), but producing one with different hash each time even if there were not any changes

---

### Testing 

I pushed this commit to the main branch of my fork, the image was built and pushed and the cache was created:
(run log https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/14938958457/job/41972649853)

image

Then I pulled it from my machine:

```bash
$ docker pull ghcr.io/mauriciovasquezbernal/gadget-builder:main
main: Pulling from mauriciovasquezbernal/gadget-builder
254e724d7786: Already exists
54a73374204c: Pull complete
6b6fa017d695: Pull complete
305b1a7efd5b: Pull complete
51f865fbab7f: Pull complete
c30234827b1e: Pull complete
3491132245d9: Pull complete
4f4fb700ef54: Pull complete
d5f27c6711b6: Pull complete
Digest: sha256:1b0644f6ec703871b4567123a0f95d23eb58116eed1e21dca9cdba3d38f2aed9
Status: Downloaded newer image for ghcr.io/mauriciovasquezbernal/gadget-builder:main
ghcr.io/mauriciovasquezbernal/gadget-builder:main
```

As expected, it pulled a lot of layers

Then I pushed a commit simulating a small change to a header file, now the compilation only took 30 seconds:
(run https://github.com/mauriciovasquezbernal/inspektor-gadget/actions/runs/14938958457/job/41972649852)

image

When trying to pull it, only a small layer was pulled:

```bash
$ docker pull ghcr.io/mauriciovasquezbernal/gadget-builder:main
main: Pulling from mauriciovasquezbernal/gadget-builder
254e724d7786: Already exists
54a73374204c: Already exists
6b6fa017d695: Already exists
305b1a7efd5b: Already exists
51f865fbab7f: Already exists
c30234827b1e: Already exists
3491132245d9: Already exists
4f4fb700ef54: Already exists
33adc89c2121: Pull complete
Digest: sha256:e1ea1634fb927b303c04bec25d0994674b996a0466e9917a61e44c75783ca81a
Status: Downloaded newer image for ghcr.io/mauriciovasquezbernal/gadget-builder:main
ghcr.io/mauriciovasquezbernal/gadget-builder:main
```

Fixes #4345

cc @mqasimsarfraz @rata

### DIFF
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -988,42 +988,30 @@ jobs:
             context: "tools/dnstester"
             dockerfile: "tools/dnstester/Dockerfile"
             platform: "linux/amd64,linux/arm64"
-            filter-patterns:
-              - "tools/dnstester/*"
+            key-files: "tools/dnstester/*"
           - name: "gadget-builder"
             context: "/home/runner/work/inspektor-gadget/inspektor-gadget"
             dockerfile: "Dockerfiles/gadget-builder.Dockerfile"
             platform: "linux/amd64,linux/arm64"
-            filter-patterns:
-              - "include/**"
-              - "Dockerfiles/gadget-builder.Dockerfile"
-              - "cmd/common/image/Makefile.build"
+            key-files: "'include/**','Dockerfiles/gadget-builder.Dockerfile','cmd/common/image/Makefile.build'"
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-    - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
-      id: filter
-      with:
-        # https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-type-check-expression
-        filters: |
-          pattern: ${{ toJson(matrix.image.filter-patterns) }}
-    - name: Check if we should build helpers
-      id: check-build-helpers
-      # always build the images on release, merge to main or to a release branch
-      if: steps.filter.outputs.pattern == 'true' || startsWith(github.ref_name, 'v') || startsWith(github.ref_name, 'release-') || github.ref_name == 'main'
-      run: |
-          echo "build=true" >> $GITHUB_OUTPUT
     - name: Set up Docker Buildx
-      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+    - name: Cache Docker layers
+      uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-docker-${{ matrix.image.name }}-${{ hashFiles(matrix.image.key-files) }}
+        restore-keys: |
+          ${{ runner.os }}-docker-${{ matrix.image.name}}-
     - name: Login to Container Registry
-      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
     - name: Set container repository and determine image tag
-      if: steps.check-build-helpers.outputs.build == 'true'
       id: set-repo-determine-image-tag
       uses: ./.github/actions/set-container-repo-and-determine-image-tag
       with:
@@ -1031,7 +1019,6 @@ jobs:
         container-image: ${{ github.repository_owner }}/${{ matrix.image.name }}
     - name: Build ${{ matrix.image.name }} image
       id: build-image
-      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1 # v6.16.0
       with:
         context: ${{ matrix.image.context }}
@@ -1041,6 +1028,8 @@ jobs:
         # https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository
         push: ${{ vars.PUSH_HELPERS == 'ENABLE_PUSH_HELPERS' }}
         tags: ${{ steps.set-repo-determine-image-tag.outputs.container-repo }}:${{ steps.set-repo-determine-image-tag.outputs.image-tag }}
+        cache-from: type=local,src=/tmp/.buildx-cache
+        cache-to: type=local,dest=/tmp/.buildx-cache-new
         platforms: ${{ matrix.image.platform }}
     - name: Save ${{ matrix.image.name }} image tag output
       id: image-tag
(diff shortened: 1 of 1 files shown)


rata:
> This commit enables the cache fixing the issues described above

Awesome! What is the change providing reproducible images now? Or we are not but just using the cache action in a better way for this image?

.github/workflows/inspektor-gadget.yml
```diff
@@ -988,50 +988,37 @@ jobs:
...
-      run: |
-          echo "build=true" >> $GITHUB_OUTPUT
     - name: Set up Docker Buildx
-      if: steps.check-build-helpers.outputs.build == 'true'
       uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+    - name: Cache Docker layers
+      uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-docker-${{ matrix.image.name }}-${{ hashFiles(matrix.image.key-files) }}
```
mqasimsarfraz, line 1005: Caching layers would mean things like `apt-get update` won't be using the latest changes so perhaps we should expire the cache? Having a date in the key should do the tick: https://github.com/marketplace/actions/cache#creating-a-cache-key

mqasimsarfraz (commented):
I like this change, will make things faster! 

I know we are relying on cache to have the same layers in the final image but perhaps we should open issue to see what is the root cause that the image isn't reproducable?

.github/workflows/inspektor-gadget.yml
```diff
@@ -988,50 +988,37 @@ jobs:
...
-      with:
-        # https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-type-check-expression
-        filters: |
-          pattern: ${{ toJson(matrix.image.filter-patterns) }}
-    - name: Check if we should build helpers
-      id: check-build-helpers
-      # always build the images on release, merge to main or to a release branch
-      if: steps.filter.outputs.pattern == 'true' || startsWith(github.ref_name, 'v') || startsWith(github.ref_name, 'release-') || github.ref_name == 'main'
-      run: |
-          echo "build=true" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
mauriciovasquezbernal: That's right, the logic was:
- Only build it on PRs when the relevant files were changed
- Always build it on main and for releases

As the cache speeds it a lot (only few seconds are needed when not relevant file is changed), it makes more sense to rely directly on the build cache than on this complicated mechanism.

mauriciovasquezbernal:
> Awesome! What is the change providing reproducible images now? Or we are not but just using the cache action in a better way for this image?

The build is not reproducible yet. (building without the cache produces different images). But using the cache produces the same image as long as it's valid.

mauriciovasquezbernal:
> I know we are relying on cache to have the same layers in the final image but perhaps we should open issue to see what is the root cause of why the image build isn't reproducible (Although not sure if that is worth it)?

I think it's a good idea to have reproducible builds in the long term, but IMO it's a low priority task for now. Perhaps you can open the issue indicating it's a "good to have" but not super urgent.

mqasimsarfraz:
> > I know we are relying on cache to have the same layers in the final image but perhaps we should open issue to see what is the root cause of why the image build isn't reproducible (Although not sure if that is worth it)?
> 
> I think it's a good idea to have reproducible builds in the long term, but IMO it's a low priority task for now. Perhaps you can open the issue indicating it's a "good to have" but not super urgent.



> > I know we are relying on cache to have the same layers in the final image but perhaps we should open issue to see what is the root cause of why the image build isn't reproducible (Although not sure if that is worth it)?
> 
> I think it's a good idea to have reproducible builds in the long term, but IMO it's a low priority task for now. Perhaps you can open the issue indicating it's a "good to have" but not super urgent.

Sounds good, I opened the issue: https://github.com/inspektor-gadget/inspektor-gadget/issues/4456. Please feel free to edit it if I missed something!

.github/workflows/inspektor-gadget.yml
mauriciovasquezbernal: > Caching layers would mean things like apt-get update won't be using the latest changes so perhaps we should expire the cache? 

What's the issue of not using the latest changes? Having old code with CVEs and that stuff?

> Having a date in the key should do the tick: https://github.com/marketplace/actions/cache#creating-a-cache-key

If we invalidate the cache too often, then what's the advantage of this PR?
Perhaps we can manually delete the cache and rerun this job when needed?

mqasimsarfraz (approved):
LGTM! 

Edit: I tried re-triggering subset of jobs but they failed because of expired artifacts so will need to trigger the whole CI again! :see_no_evil:

PR #4466, opened 2025-05-14
operators: ebpf: Use Info() for --trace-pipe.

eiffel-fl:
This way, using --trace-pipe actually prints the data without needing -v.

Fixes: d28c76d0ef85 ("pkg/operators/ebpf: ebpf layer operator")


flyth (approved):
Makes sense to me, thanks!

PR #4470, opened 2025-05-15
cmd: Specify run and attach short help for the needed gadget.

eiffel-fl:
cc: @rata

### DIFF
--- a/cmd/common/oci.go
+++ b/cmd/common/oci.go
@@ -50,8 +50,8 @@ import (
 type CommandMode string
 
 const (
-	CommandModeRun    CommandMode = "run"
-	CommandModeAttach CommandMode = "attach"
+	CommandModeRun    CommandMode = "run GADGET"
+	CommandModeAttach CommandMode = "attach GADGET_INSTANCE"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/common/oci.go
```diff
@@ -50,8 +50,8 @@ import (
 type CommandMode string
 
 const (
-	CommandModeRun    CommandMode = "run"
-	CommandModeAttach CommandMode = "attach"
+	CommandModeRun    CommandMode = "run GADGET"
+	CommandModeAttach CommandMode = "attach GADGET_INSTANCE"
```
rata, line 54: Usually required params go like `<gadget>`. Maybe use that style here too?

rata:
I guess it is unrelated to this, but ig run also takes another set of flags before the gardget, right?

cmd/common/oci.go
rata: Usually optional args go like `[optional arg]` but it seems the required ones are not so established. LGTM, then :)

rata (approved):
LGTM, thanks!

alban (approved):

PR #4482, opened 2025-05-20
container-collection: publish oci config as a string

alban:
This PR is a first step for #4458.

The container-collection package detects when the container runtime starts a new container and reads the oci config (config.json from the container runtime spec). This config is used internally in the container-collection package for two purposes:
1. Look up the mount for the Kubernetes ServiceAccount token volume to distinguish several containers from the same pod.
2. Look up the annotations for Kubernetes enrichment (pod name, etc.)

This patch makes the config available in the containers data source as a string. It is now unmarshalled partially only when needed.

Before this patch, the config was completely unmarshalled using the opencontainers/runtime-spec package v1.2.1 (version specified in go.mod):
https://github.com/opencontainers/runtime-spec/blob/v1.2.1/specs-go/config.go#L6

```
$ sudo ig list-containers -o json -w
{
  "timestamp": "2025-05-20T11:04:34+02:00",
  "event": "CREATED",
  "container": {
    "runtime": {
      "runtimeName": "docker",
      "containerId": "a5c7f2341346455531e3b5c4ca9fa96ed598d2c8d157ff617bfce1f5a948be57",
      "containerName": "ecstatic_tesla",
      "containerPid": 49978,
      "containerImageName": "busybox",
      "containerStartedAt": 1747731874644255522
    },
    "k8s": {},
    "ociConfig": {
      "ociVersion": "1.2.0",
      "process": {
        "terminal": true,
```

This means that if users run runc/crun with a newer version of runtime-spec with new fields, they will be lost.

After this patch, the config is displayed as a string without changes:
```
$ sudo ig list-containers -o json -w
{
  "timestamp": "2025-05-20T11:02:36+02:00",
  "event": "CREATED",
  "container": {
    "runtime": {
      "runtimeName": "docker",
      "containerId": "e63be34d020f4b3f44cd7cba885511f751dc901ab605324551d7e18ede482ca1",
      "containerName": "ecstatic_solomon",
      "containerPid": 48248,
      "containerImageName": "busybox",
      "containerStartedAt": 1747731756812735022
    },
    "k8s": {},
    "ociConfig": "{\"ociVersion\":\"1.2.0\",\"process\":{\"terminal\":true
```

This will be useful in the next patches when implementing the possibility to modify the config from the "containers" data source:
- We will avoid unnecessary unmarshalling/re-marshalling
- This will help with compatibility when additional fields are added in the container-runtime spec.

## Testing done

- [x] Listing containers
  - [x] Manual test: `sudo ig list-containers -o json -w`
  - [x] Automatic tests: `integration/**/list_containers_test.go`
- [x] OCI config in the "containers" data source: `gadgets/ci/datasource-containers/test/integration/datasource_test.go`

### DIFF
--- a/examples/container-hook/main.go
+++ b/examples/container-hook/main.go
@@ -32,7 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/inspektor-gadget/inspektor-gadget/pkg/container-hook"
+	containerhook "github.com/inspektor-gadget/inspektor-gadget/pkg/container-hook"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/k8sutil"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/host"
 )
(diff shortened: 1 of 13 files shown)


mauriciovasquezbernal (approved):
LGTM from code inspection.

PR #4492, opened 2025-05-22
operator/combiner: Move DS creation to Instantiate

burak-ok:
This moves the DS unreference and new DS creation to the InstantiateDataOperator call. This is needed in order for other Operators to reference the new DS instead of the old one that would have been replaced in PreStart


flyth (approved):
LGTM from code inspection.

PR #4493, opened 2025-05-22
combiner: Disable timeout if interval and count is zero

burak-ok:
This combination only happens for the advise_networkpolicy gadget. There we have an array datasource and only want to get a packet at the end of the gadget, when it gets cancled/stopped.

This is done through setting count and interval to 0, while having flush-on-stop set to true

### DIFF
--- a/pkg/operators/combiner/combiner.go
+++ b/pkg/operators/combiner/combiner.go
@@ -24,6 +24,7 @@ package combiner
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"time"
 
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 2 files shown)


pkg/operators/combiner/combiner.go
```diff
@@ -173,7 +186,13 @@ func (o *combinerOperatorInstance) forwardData(
...
 		// Define a maximum waiting time for data from all targets
 		// TODO: Make it configurable?
-		timeout := time.NewTimer(5 * time.Second)
+		defaultTimeout := 5 * time.Second
+		if config.count == 0 {
+			// We are only handling array data sources here
+			// If interval and count are both 0, wait till the user stops the gadget
+			defaultTimeout = 24 * 365 * time.Hour
+		}
+		timeout := time.NewTimer(defaultTimeout)
```
mauriciovasquezbernal, line 195: What about letting `c` be nil instead of using one year? (waiting on nil blocks forever, so the result is the same in practice)

mauriciovasquezbernal (commented):
Thanks for working on this! We have the same issue on the advise_seccomp gadget.

pkg/operators/combiner/combiner.go
burak-ok: Even better!

pkg/operators/combiner/combiner.go
```diff
@@ -171,11 +184,17 @@ func (o *combinerOperatorInstance) forwardData(
...
+			// We are only handling array data sources here
+			// If interval and count are both 0, wait till the user stops the gadget
+			c = nil
+		} else {
+			// Define a maximum waiting time for data from all targets
+			// TODO: Make it configurable?
+			timeout := time.NewTimer(5 * time.Second)
+			defer timeout.Stop()
+			c = timeout.C
+		}
```
mauriciovasquezbernal, line 197: The first conditional isn't needed. (even if I agree the comments are easier to write with both branches)

```suggestion
		// If count is 0, wait until the user stops the gadget, otherwise wait for
		// 5 seconds
		// TODO: Make it configurable?
		if config.count != 0 {
			timeout := time.NewTimer(5 * time.Second)
			c = timeout.C
			defer timeout.Stop()
		}
```

mauriciovasquezbernal (approved):
Just a small nit, feel free to take or ignore it. LGTM, thanks!

PR #4498, opened 2025-05-22
gadgets/trace_exec: Add parent_exepath

### LINKED ISSUE #4483: [RFE] Add parent exepath to gadget_process_populate
## Current situation

Parent process is being populated with comm and pid

## Ideal future situation

In addition to current parent process fields, exe field will be populated as well

mauriciovasquezbernal:
Add path of parent executable.

Fixes #4483

cc @alban


burak-ok (approved):
LGTM after the clang-format error is resolved

alban (approved):

PR #4501, opened 2025-05-23
pkg/params: allow escaping comma characters for string slices

flyth:
This allows strings to contain escaped commas (e.g. `abc\,def,ghi`) that are not considered delimiters (leading to `["abc,def", "ghi"` instead of `["abc\", "def", ghi"]`). Escape characters are only removed if used with a legal character (comma and backslash), and otherwise kept as is.

I also modified the already existing code in the api package to behave the same way (before it would always remove escape characters). I deliberately chose to duplicate code here to keep both packages fully self-contained.

This was reported as bug on [slack](https://kubernetes.slack.com/archives/CSYL75LF6/p1747680905544569).

### DIFF
--- a/pkg/gadget-service/api/helpers.go
+++ b/pkg/gadget-service/api/helpers.go
@@ -1,4 +1,4 @@
-// Copyright 2023-2024 The Inspektor Gadget authors
+// Copyright 2023-2025 The Inspektor Gadget authors
 //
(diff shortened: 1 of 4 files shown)


mqasimsarfraz (approved):
I tested the changes and they work as expected! 

Also, I agree it should be fine to keep backslash if we don't have special characters following.

flyth:
Thanks for the review!

PR #4505, opened 2025-05-23
metadata: Continue validation w/o ebpf layer

mqasimsarfraz:
Currently if we don't have any ebpf program we skip certain validation around metadata file e.g. validateEbpfParams. This change will continue to validate even if we don't have any ebpf programs. Also, it also removes gadget name as the mandatory field since we don't really rely on it.


burak-ok:
> Currently if we don't have any ebpf program we skip certain validation around metadata file e.g. validateEbpfParams. 

Looking at the naming `validateEbpfParams` I think that specific should be skipped when there are no `ebpf` programs. Should we rename the validateEbpfParams to validateGadgetParams if we are using it for WASM and other future things too?

mqasimsarfraz:
> Looking at the naming validateEbpfParams I think that specific should be skipped when there are no ebpf programs

The current validation only validates ebpf param if they exists so we should be fine here.

> Should we rename the validateEbpfParams to validateGadgetParams if we are using it for WASM and other future things too?

I don't have strong opinion on this, like current logic is specific to [ebpf params](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/run/types/metadata.go#L64) so we can keep it as it is. Once we change the scope we can rename as needed?

burak-ok (approved):
LGTM

mqasimsarfraz:
Thanks for the review

PR #4509, opened 2025-05-23
bpfstats/unittest: Set values to normalized values

burak-ok:
Manually set the values to the normalized ones because the function doesn't modify the value when it is 0.
I saw these values to be 0 and I think that might be normal, right?
What about the other normalized values?

Fixes: 0ef927570639c85d4a16efa2991cbe24ad0228e7


mauriciovasquezbernal (approved):
Makes sense, these values can be 0. 

off-topic: The idea of the normalization logic is to check if a field is present in the json but without checking its value. The current implementation treats 0 as a special case, as it means the value wasn't present or it's actually zero. I think we need to switch to use pointers around to distinguish when the field is not present (`nil`) or when it's zero. 

(I'll think about it an open an issue later on)

Issue #2526, opened 2024-02-22
[RFE] Provide helpers to fill `struct gadget_l4endpoint_t` up


blanquicet:
## Current situation

All the gadgets that provide the src/dst IP/port and the protocol using the `gadget_l4endpoint_t` has to implement the same logic to fill it:

https://github.com/inspektor-gadget/inspektor-gadget/blob/2340dd4c608f8f2716f3521ee502cf2fe5a0bcdc/gadgets/trace_tcpconnect/program.bpf.c#L331-L349

## Impact

A lot of duplicated code.

## Ideal future situation

IG provides helpers to fill the `gadget_l4endpoint_t` struct from `struct sock *`.

Priyansurout:
Hello @blanquicet  I would like to work on this issue

blanquicet:
Hi @Priyansurout, sorry for the long delay. Yes, it was merged but it's still a helper useful for other gadgets like top_tcp, trace_bind, trace_tcp, etc.

PR #4511, opened 2025-05-23
gadgets: Compile with '-Werror' in the CI

burak-ok:
gadgets: Compile with '-Werror' in the CI

### DIFF
--- a/cmd/common/image/build.go
+++ b/cmd/common/image/build.go
@@ -454,6 +454,10 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
 		ebpfFullPath = filepath.Join(gadgetSourcePath, conf.EBPFSource)
 	}
 
+	if cflags, set := os.LookupEnv("CFLAGS"); set {
+		conf.CFlags += " " + cflags
+	}
+
 	cmd := buildCmd("/out", ebpfFullPath, wasmFullPath, conf.CFlags, "/btfhub-archive", opts.btfgen)
 
 	mounts := []mount.Mount{
--- a/.github/workflows/inspektor-gadget.yml
+++ b/.github/workflows/inspektor-gadget.yml
@@ -1558,6 +1558,7 @@ jobs:
           BUILDER_IMAGE: ${{ needs.build-helper-images.outputs.gadget_builder_image }}
           GADGET_REPOSITORY: ${{ steps.set-repo-determine-image-tag.outputs.gadget-repository }}
           GADGET_TAG: ${{ steps.set-repo-determine-image-tag.outputs.gadget-tag }}
+          CFLAGS: "-Werror"
         run: |
           make build-gadgets -o install/ig -j$(nproc)
 


gadgets/trace_tcpdrop/program.bpf.c
```diff
@@ -156,12 +156,8 @@ static __always_inline int __trace_tcp_drop(void *ctx, struct sock *sk,
 		BPF_CORE_READ_INTO(
 			&event->src.addr_raw.v6, sk,
 			__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr32);
-		if (event->src.addr_raw.v6 == 0)
-			goto cleanup;
```
burak-ok, line 160: I wasn't so sure about this one. The warning is:

> /work/gadgets/trace_tcpdrop/program.bpf.c:159:27: warning: comparison of array 'event->src.addr_raw.v6' equal to a null pointer is always false [-Wtautological-pointer-compare]

So can the address be 0 after reading something into it? If yes we need to make sure that the compiler knows that that can be the case.

Because right now it thinks that never can be true and optimize that out

mauriciovasquezbernal: > So can the address be 0 after reading something into it? 

It can't, the compiler is right. The `addr_raw` member is a union defined as:

https://github.com/inspektor-gadget/inspektor-gadget/blob/04c60a7421824cded611598b25f714a3ca63e845/include/gadget/types.h#L13-L17

Checking if `v4 == 0` is fine as it's a scalar, but checking `v6 == 0` is wrong as it's an array, so that check is checking the array's address and not its contents. I see two options:

1. Extending the `gadget_ip_addr_t` union to contain a 128bits integer representing the IPv6 as done in the built-in tcpdrop:
 
https://github.com/inspektor-gadget/inspektor-gadget/blob/5a9ec7fe720e51c7a39ac16148e6e13e1a0555b5/pkg/gadgets/trace/tcpdrop/tracer/bpf/tcpdrop.h#L23-L27

2. Create a helper to check if an gadget_ip_addr_t is zero. (It'd need to loop through all bytes of it or use some casting to compare against. Like done in profile_tcprtt:

https://github.com/inspektor-gadget/inspektor-gadget/blob/4795e94479f14ce571b19f6131c633345e93aff8/gadgets/profile_tcprtt/program.bpf.c#L89-L95

I think option 1 is easier to handle. TBH I don't remember why I chose `__u8 v6[16];` over `unsigned __int128` when implementing it.

mauriciovasquezbernal: It's related to #2526 too.

burak-ok: My C really got rusty :/

Implemented option 1

mauriciovasquezbernal:
If you move the changes to the ebpf code to a new PR we could merge it right away to fix those warnings.

cmd/common/image/helpers/Makefile.build
```diff
@@ -13,7 +13,7 @@ LLVM_STRIP ?= llvm-strip
 #
 # Since Ubuntu does not install it in a standard path, add a compiler flag for
 # it.
-BASECFLAGS = -target bpf -Wall -g -O2 -I/usr/include/$(shell uname -m)-linux-gnu
+BASECFLAGS = -target bpf -Wall -Werror -g -O2 -I/usr/include/$(shell uname -m)-linux-gnu
```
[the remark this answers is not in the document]
eiffel-fl: I agree with Mauricio here and would rather not have this as default.
Though, we should aim towards tackling ALL warnings as individual responsibility and warnings we cannot cope with should be documented.

In this case, someone could decide to use CFLAGS to specify some flags, so one could definitely do CLAFGS='-Werror' make trace_exec.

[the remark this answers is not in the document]
burak-ok: So disabled by default in the `Makefile`.

Do we want to enable it in our CI though?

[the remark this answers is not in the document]
mauriciovasquezbernal: I'd say yes, if we find this too annoying we always can change it back.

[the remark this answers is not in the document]
burak-ok: I updates this PR.

First commit introduces a new flag to `ig image build` and the second commit uses it in our workflow

cmd/common/image/build.go
```diff
@@ -110,6 +111,7 @@ func NewBuildCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.outputDir, "output", "o", "", "Path to a folder to store generated files while building")
 	cmd.Flags().StringVarP(&opts.image, "tag", "t", "", "Name for the built image (format name:tag)")
 	cmd.Flags().StringVar(&opts.builderImage, "builder-image", builderImage, "Builder image to use")
+	cmd.Flags().StringVar(&opts.cflags, "cflags", "", "Additional flags to pass to the ebpf compiler")
```
mauriciovasquezbernal: How does it play with cflags defined in build.yaml? Are they overwritten? If that's the case, we'd need to document it.

burak-ok: Good catch.
I updated it so the `cflags` flags from the `ig image build --cflags` command get appended to the one from the `build.yaml`

cmd/common/image/build.go
```diff
@@ -75,6 +75,7 @@ type cmdOpts struct {
 	validateMetadata bool
 	btfgen           bool
 	btfhubarchive    string
+	cflags           string
```
eiffel-fl: Is this commit really needed?
To have the -Werror passed, you can use CFLAGS='-Werror' when building the gadgets in the CI.

burak-ok: The ENV variable doesn't get passed through to the docker container:

(-O6 doesnt exists, it should error out):
```bash
$ CFLAGS="-O6" IG=`pwd`/ig make -C gadgets trace_exec
...
Successfully built ghcr.io/inspektor-gadget/gadget/trace_exec:burak-gadgets-werror@sha256:b89624fe9c0c8e7dfdfa180ca040db3cbb44575104275870b489c7d611528364
make: Leaving directory '/home/burak/codes/inspektor-gadget/gadgets'
```

eiffel-fl: Would it make better sense to fix this rather than introducing the new flag? Or is the new flag the only way to fix this?

burak-ok: That is up to how we want to achieve this.
1. This Flag
2. Pass only the `CFLAGS` environment varialbe through
3. Pass all environment variables through
4. Something else

mauriciovasquezbernal: > Pass only the CFLAGS environment varialbe through

IMO we should go with this one to avoid the flag.

burak-ok: Changed the code to that

Issue #1774, opened 2023-06-22
[RFE] New gadget: snapshot file


alban:
## Current situation

We have two snapshot gadgets:
- snapshot process
- snapshot socket

But no "snapshot file" to see all the files opened by all processes

## Impact

Lack of visibility on open files, except some tcp/udp sockets.

## Ideal future situation

Show all the open files (without filtering out the non-socket files). It would know about a few types of files (inet/unix sockets, regular files...) and would display that information. Only show files attached to processes, so sockets in TIME_WAIT would not be shown.

Examples of file types:
```
--type socket
--type socket/inet
--type socket/inet6
--type socket/unix
--type regular
--type pipe
--type bpf
--type bpf/prog...
```

## Implementation options

This was initially implemented in https://github.com/inspektor-gadget/inspektor-gadget/pull/1597
Current work in progress: https://github.com/inspektor-gadget/inspektor-gadget/pull/1725

mani1911:
So there is a need for gadget that shows files opened by all processes right?
Can I be assigned this task?

mauriciovasquezbernal:
@mani1911 for sure, are you still interested?

mani1911:
> @mani1911 for sure, are you still interested?

yes i would love to take up this issue

PR #4514, opened 2025-05-24
New Gadget: Snapshot File

Shaheer-Ahmd:
From the root directory, run this to build the gadget
`./ig image build -t snapshot_file:latest ./gadgets/snapshot_file`
then run the gadget using
`sudo ./ig run snapshot_file:latest --verify-image=false`
you may pass optional  `--type-mask` parameter for filtering

an effort towards #1774 

sample output:
image

### DIFF
--- a/gadgets/snapshot_file/README.mdx
+++ b/gadgets/snapshot_file/README.mdx
@@ -1,0 +1,112 @@
+---
+title: snapshot_file
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+The `snapshot_file` gadget shows open files for running processes.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/snapshot_file:%IG_TAG% [flags]
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/snapshot_file:%IG_TAG% [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Flags
+
+### `--paths`
+
+Show full file paths (by default, only dentry names are shown).
+
+Default value: `"false"`
+
+### `--file-type-mask`
+
+Filter which file types to display using a bitmask.
+By default, only **regular files** and **directories** are shown (bitmask value 17).
+
+Supported file types:
+
+- REGULAR
+- SOCKET
+- PIPE
+- BPF_MAP
+- DIRECTORY
+- CHAR_DEV
+- BLOCK_DEV
+- SYMLINK
+- OTHER
+
+The bitmask order, from left to right, corresponds to the file types listed above.
+
+## Guide
+
+Run a pod / container:
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl run --restart=Never --image=busybox test-snapshot-file -- sh -c 'echo hello > /tmp/demo.txt && tail -f /tmp/demo.txt'
+pod/test-snapshot-file created
+```
+
+</TabItem>
+
+<TabItem value="ig" label="ig">
+```bash
+$ docker run --name test-snapshot-file -d busybox /bin/sh -c 'echo hello > /tmp/demo.txt && tail -f /tmp/demo.txt'
+...
+```
+
+Now, run the gadget and see how it shows the open file:
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run snapshot_file:%IG_TAG%
+K8S.NODE         K8S.NAMESPACE   K8S.PODNAME         K8S.CONTAINERNAME   COMM   PID   TID   FD   PATH      TYPE       FLAGS    
+minikube-docker  default         test-snapshot-file  test-snapshot-file  tail   5432  5432  3    demo.txt  REGULAR    O_RDONLY  
+^C
+```
+</TabItem>
+
+<TabItem value="ig" label="ig">
+```bash
+$ sudo ig run snapshot_file:%IG_TAG% -c test-snapshot-file
+RUNTIME.CONTAINERNAME             COMM               PID        TID   FD         PATH            TYPE            FLAGS              
+test-snapshot-file                tail               5321       5321  3          demo.txt        REGULAR         O_RDONLY
+^C
+```
+</TabItem>
+</Tabs>
+
+Finally, clean the system:
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl delete pod test-snapshot-file
+```
+</TabItem>
+
+<TabItem value="ig" label="ig">
+```bash
+$ docker rm -f test-snapshot-file
+```
+</TabItem>
+</Tabs>
+
--- a/gadgets/snapshot_file/gadget.yaml
+++ b/gadgets/snapshot_file/gadget.yaml
@@ -1,0 +1,30 @@
+name: snapshot file
+description: Show open files
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/snapshot_file
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/snapshot_file
+datasources:
+  files:
+    fields:
+      mntns_id:
+        annotations:
+          columns.hidden: true
+      inode:
+        annotations:
+          columns.hidden: true
+      flags_raw:
+        annotations:
+          columns.hidden: true
+      flags:
+        annotations:
+          columns.hidden: false
+params:
+  ebpf:
+    paths:
+      key: paths
+      defaultValue: "false"
+      description: Enable to get the complete file paths in the output. If disabled, only the file names will be shown.
+    file_type_mask:
+      key: type-mask
+      defaultValue: "17"
+      description: File type mask to filter files by type (default includes regular files and directories => 2^0 + 2^4 = 17)
--- a/gadgets/snapshot_file/program.bpf.c
+++ b/gadgets/snapshot_file/program.bpf.c
@@ -1,0 +1,148 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2025 Shaheer Ahmad */
+
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+
+#include <gadget/macros.h>
+#include <gadget/filter.h>
+#include <gadget/types.h>
+#include <gadget/filesystem.h>
+
+const volatile bool paths = false;
+
+#define MAX_PATH_LEN 256
+
+enum file_type {
+	REGULAR = 0,
+	SOCKET,
+	PIPE,
+	BPF_MAP,
+	DIRECTORY,
+	CHAR_DEV,
+	BLOCK_DEV,
+	SYMLINK,
+	OTHER
+};
+
+// Regular and Directory by default
+const volatile u32 file_type_mask = (1U << REGULAR) | (1U << DIRECTORY);
+
+struct gadget_file {
+	gadget_mntns_id mntns_id;
+	char comm[TASK_COMM_LEN];
+	gadget_pid pid;
+	gadget_tid tid;
+	__u32 fd;
+	enum file_type type_raw;
+	char path[MAX_PATH_LEN];
+	__u64 inode;
+	gadget_file_flags flags_raw;
+};
+
+#define S_IFMT 00170000 /* bitmask for the file type bitfields */
+#define S_IFSOCK 0140000 /* socket */
+#define S_IFLNK 0120000 /* symbolic link */
+#define S_IFREG 0100000 /* regular file */
+#define S_IFBLK 0060000 /* block device */
+#define S_IFDIR 0040000 /* directory */
+#define S_IFCHR 0020000 /* character device */
+#define S_IFIFO 0010000 /* FIFO */
+
+GADGET_PARAM(file_type_mask);
+GADGET_PARAM(paths);
+
+GADGET_SNAPSHOTTER(files, gadget_file, ig_snap_file);
+
+static __always_inline u32 classify_file_type(struct file *file)
+{
+	u32 mode = BPF_CORE_READ(file, f_inode, i_mode) & S_IFMT;
+
+	switch (mode) {
+	case S_IFREG:
+		return REGULAR;
+	case S_IFDIR:
+		return DIRECTORY;
+	case S_IFIFO:
+		return PIPE;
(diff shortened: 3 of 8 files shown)


Shaheer-Ahmd:
Discussion points:
1) Should we display all the file types by default or only the `REGULAR` and `DIRECTORY` types should be shown? Furthermore, what is the best format for the type mask param UX wise?
2) We would like the FLAGS and the MODE column to be more human readable using `flags_set` and `mode_set` bitfield enums. But, that requires us to define our own enum (since the name should have a `_set` suffix) with all the possible linux FLAGS and MODES values. What should we do here?

Shaheer-Ahmd:
@alban @mauriciovasquezbernal thoughts on the discussion points?

mauriciovasquezbernal:
> Should we display all the file types by default or only the REGULAR and DIRECTORY types should be shown? 

I think regular and directory by default is good enough

> Furthermore, what is the best format for the type mask param UX wise?

> We would like the FLAGS and the MODE column to be more human readable using flags_set and mode_set bitfield enums. But, that requires us to define our own enum (since the name should have a _set suffix) with all the possible linux FLAGS and MODES values. What should we do here?


I'd suggest we handle it **as** on gadgets/trace_open/go/program.go

burak-ok:
> > Furthermore, what is the best format for the type mask param UX wise?
> > We would like the FLAGS and the MODE column to be more human readable using flags_set and mode_set bitfield enums. But, that requires us to define our own enum (since the name should have a _set suffix) with all the possible linux FLAGS and MODES values. What should we do here?
> 
> I'd suggest we handle it on gadgets/trace_open/go/program.go

Do you mean like the logic we had for `trace_open`? So doing it in the user space portion of the gadget.
Should we create a shared package for the user space portion, so gadgets wouldn't need to duplicate this logic?

mauriciovasquezbernal:
> Do you mean like the logic we had for `trace_open`? So doing it in the user space portion of the gadget.

Oh right, I edited my comment.

> Should we create a shared package for the user space portion, so gadgets wouldn't need to duplicate this logic?

Ideally yes, but I don't know if it complicates compilation of gadgets.

burak-ok:
> > Should we create a shared package for the user space portion, so gadgets wouldn't need to duplicate this logic?
> 
> Ideally yes, but I don't know if it complicates compilation of gadgets.

Then let's copy paste for now I guess.

We shouldn't block this PR until we figure out in which repo/package we want to provide a shared library for userspace golang code. Then we might also need to think about feature parity to userspace rust code

mauriciovasquezbernal:
And also, given that it's common functionality, I'll also think to include that on the framework directly to avoid the wasm overhead. (but let's continue with this PR as it's)

Shaheer-Ahmd:
@burak-ok @mauriciovasquezbernal 
Let me know if you would like me to change/add anything in the code that I added. If everything looks good, we can move towards merging this.

gadgets/snapshot_file/go/go.mod
mauriciovasquezbernal, line 1: Please use the same structure as in other gadgets, for instance gadgets/trace_dns/go/go.mod

gadgets/snapshot_file/build.yaml
```diff
@@ -0,0 +1 @@
+wasm: go/program.go
```
mauriciovasquezbernal: nit: missing line break at the end

gadgets/snapshot_file/gadget.yaml
```diff
@@ -0,0 +1,39 @@
+name: snapshot file
+description: Show files
```
mauriciovasquezbernal:
```suggestion
description: Show open files
```

gadgets/snapshot_file/program.bpf.c
```diff
@@ -0,0 +1,155 @@
+// SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note
+/* Copyright (c) 2025 Your Name
+ *
+ * Snapshot File Gadget
+ *
+ * Iterates over all open files of each task and emits one record per file.
+ */
```
mauriciovasquezbernal: Add a proper license header.

gadgets/snapshot_file/gadget.yaml
```diff
@@ -0,0 +1,39 @@
...
+    fields:
+      mntns_id:
+        annotations:
+          columns.hidden: true
+      tid:
+        annotations:
+          columns.hidden: true
+      pid:
+        annotations:
+          columns.hidden: true
```
mauriciovasquezbernal: If gadget_pid and gadget_tid types are used in the bpf code, these members on the metadata could be removed as IG will be able to apply the proper template for those fields.

gadgets/snapshot_file/go/program.go
```diff
@@ -0,0 +1,131 @@
...
+			flagsStr = append(flagsStr, val)
+		}
+	}
+
+	return flagsStr
+}
+
+//go:wasmexport gadgetInit
+func gadgetInit() int32 {
+
```
mauriciovasquezbernal: remove this extra line

gadgets/snapshot_file/go/program.go
```diff
@@ -0,0 +1,131 @@
...
+		modeStr := fs.FileMode(modeRaw)
+		modeF.SetString(data, modeStr.String())
+
+		// flags
+		flagsRaw, err := flagsRawF.Int32(data)
+		if err != nil {
+			api.Warnf("failed to get flags: %s", err)
+			return
+		}
+		fmt.Printf("Subscribing to data source, flagsRaw: \n", flagsRaw)
```
mauriciovasquezbernal: remove

gadgets/snapshot_file/gadget.yaml
```diff
@@ -0,0 +1,34 @@
+name: snapshot_file
+image: snapshot_file:latest
```
burak-ok:
```suggestion
```
Since this YAML file is part of the image, it shouldn't have anything specific to the URL or tag specified in it.
Also no other gadget.yaml has an `image` key

gadgets/snapshot_file/gadget.yaml
```diff
@@ -0,0 +1,34 @@
+name: snapshot_file
```
burak-ok:
```suggestion
name: snapshot file
```
To keep consistent with other gadget.yaml files

gadgets/snapshot_file/go/program.go
```diff
@@ -0,0 +1,128 @@
...
+package main
+
+import (
+	"io/fs"
+	"strings"
+
+	api "github.com/inspektor-gadget/inspektor-gadget/wasmapi/go"
+)
+
+var flagNames = []string{
```
burak-ok: @mauriciovasquezbernal created a PR (https://github.com/inspektor-gadget/inspektor-gadget/pull/4792) where the logic for mode and flags are shifted from WASM to inside IG core functionality. So after this PR decoding these two fields won't be needed anymore.

I think until the PR is merged, you can either rebase your code on top of that branch, or cherry-pick the commit in the meanwhile

gadgets/snapshot_file/README.mdx
```diff
@@ -0,0 +1,128 @@
...
+minikube-docker default         test-snapshot-file  test-snapshot-file  sh     5432  5432  1    REGULAR   1234567    demo.txt
+^C
+```
+</TabItem>
+
+<TabItem value="ig" label="ig">
+```bash
+$ sudo ig run snapshot_file:%IG_TAG% -c test-snapshot-file
+RUNTIME.CONTAINERNAME   COMM   PID   TID   FD   TYPE      INODE      PATH
+test-snapshot-file      sh     5321  5321  1    REGULAR   1234567    demo.txt
```
burak-ok: The `MODE` and `FLAGS` fields are not present here in the output (Same for the K8s output)

gadgets/snapshot_file/README.mdx
```diff
@@ -0,0 +1,128 @@
...
+```
+</TabItem>
+</Tabs>
+
+Then, open a file in the container:
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl exec -it test-snapshot-file -- sh -c 'echo hello > /tmp/demo.txt && sleep inf'
```
burak-ok: The file handle should be directly closed after the echo exits? (Maybe not?) So its hard to run the snapshot gadget in that exact millisecond.

You could use `tail -f <some textfile` to keep an open FD. Sleep won't be needed then too

gadgets/snapshot_file/go/program.go
Shaheer-Ahmd: @burak-ok 
~~I updated the flags and mode types to the ones defined in the mentioned PR (gadget_file_[flags/mode]) and also set `columns.hidden=false` in the YAML file, but I still can't see the processed flags and mode in the gadget output. Am I missing something here?~~ (Also sorry for a late update, I was very busy with my university)
nvm, my `ig` binary was outdated, remaking it fixed the issue.

Shaheer-Ahmd: will sign and squash once the PR is finalized

gadgets/snapshot_file/test/unit/snapshot_file_test.go
```diff
@@ -0,0 +1,183 @@
...
+				// 		Type: "REGULAR",
+				// 		Path: "demo.txt",
+				// 		// Pid:  uint32(pid),
+				// 		// Tid:  uint32(pid),
+				// 		MntNsID: info.MountNsID,
+				// 	}
+				// })(t, info, pid, events)
+
+				//atleast 1 event should have comm == tail and pid == pid
+				for _, event := range events {
```
Shaheer-Ahmd, line 99: @burak-ok wanted to clear some confusion before I write more unit tests:
The PIDs that I get in the `generateEvent` function are different from what the gadget captures, maybe because the eBPF program is using the host PID namespace, and the testcase is using the container's PID namespace. As a result, I am unable to compare the PID of the captured event. Do you think it is fine if I do not compare the pids?

burak-ok: The unit tests are normally running directly on the host directly where ig runs. For me locally your test seems to work:
```
=== CONT  TestSnapshotFileGadget/captures_events_with_no_filter
    snapshot_file_test.go:173: Started echo command with PID 132955
    snapshot_file_test.go:174: Started tail command with PID 132956
time="2025-10-21T17:42:09+02:00" level=warning msg="gadget signature verification is disabled due to using corresponding option"
    snapshot_file_test.go:149: Events received with comm: tail
    snapshot_file_test.go:152: Found matching event: {MntNsID:4026532424 Comm:tail Pid:132956 Tid:132956 Type:REGULAR Path:demo.txt}
    snapshot_file_test.go:70: Found matching event and pids match: {MntNsID:4026532424 Comm:tail Pid:132956 Tid:132956 Type:REGULAR Path:demo.txt}
--- PASS: TestSnapshotFileGadget (0.00s)
    --- PASS: TestSnapshotFileGadget/captures_events_with_no_filter (5.10s)
```

gadgets/snapshot_file/test/unit/snapshot_file_test.go
```diff
@@ -0,0 +1,183 @@
...
+				}
+			}
+
+			testCase.validateEvent(t, runner.Info, processId, gadgetRunner.CapturedEvents)
+		})
+	}
+}
+
+func generateEvent(t *testing.T) (int, error) {
+	const filePath = "/tmp/demo.txt"
```
burak-ok: With https://pkg.go.dev/testing#T.TempDir you can generate a random tempdir which will be cleaned up automatically

Shaheer-Ahmd:
@burak-ok 
I have updated the tests and am using unique filepaths to filter the event corresponding to each testcase because my pids are not matching for some reason (maybe because I am using WSL and the testcase uses WSL pid namespace) . Nevertheless, could you please uncomment the `PID, TID` attributes in the `ExpectedSnapshotFileEvent` structs and check on your end if the tests are passing. Thanks!

burak-ok:
> @burak-ok I have updated the tests and am using unique filepaths to filter the event corresponding to each testcase because my pids are not matching for some reason (maybe because I am using WSL and the testcase uses WSL pid namespace) . Nevertheless, could you please uncomment the `PID, TID` attributes in the `ExpectedSnapshotFileEvent` structs and check on your end if the tests are passing. Thanks!

Thanks for still working on this. The tests passes for me with PID and TID enabled. Could you check if the snapshot_process tests works on WSL? Just to check if its only because of WSL or not

Apart from the WSL issues only these fast things remain from my side:
1. For the unit tests to succeed in the CI you need to add the new gadget to
https://github.com/inspektor-gadget/inspektor-gadget/blob/f5a2de2fc2610e6b3bd9e2a3fde0de12f5a3e3cb/gadgets/Makefile#L23
2. run `make lint`
3. update the timestamps/versions in the artifacthub files, etc..
4. For the README page to appear on the website you need to create a symbolic link:
  ```
  cd docs/gadgets && ln -s ../../gadgets/snapshot_file/README.mdx snapshot_file.mdx
  ```

Shaheer-Ahmd:
> Could you check if the snapshot_process tests works on WSL? Just to check if its only because of WSL or not

Those are failing too.

Shaheer-Ahmd:
@burak-ok made the final changes. Please let me know if this needs anything else.

gadgets/snapshot_file/program.bpf.c
```diff
@@ -0,0 +1,150 @@
...
+		return 0;
+
+	info.mntns_id = task->nsproxy->mnt_ns->ns.inum;
+	__builtin_memcpy(info.comm, task->comm, TASK_COMM_LEN);
+	info.pid = task->tgid;
+	info.tid = task->pid;
+	info.fd = ctx->fd;
+	info.inode = BPF_CORE_READ(file, f_inode, i_ino);
+	info.flags_raw = BPF_CORE_READ(file, f_flags);
+	info.mode_raw = BPF_CORE_READ(file, f_mode);
```
burak-ok: `f_mode` is of type `fmode_t` (https://elixir.bootlin.com/linux/v6.17.8/source/include/linux/fs.h#L1102).
It might look like it is the mode that we know in the user space, but this is something different. See https://elixir.bootlin.com/linux/v6.17.8/source/include/linux/fs.h#L130 and also other macros around that line. 

Looking around `struct file` I think there is no possibility to get the file mode from this iter point. Lets remove it form the snapshot_file gadget

gadgets/snapshot_file/README.mdx
```diff
@@ -0,0 +1,112 @@
...
+```
+
+Now, run the gadget and see how it shows the open file:
+
+<Tabs groupId="env">
+<TabItem value="kubectl-gadget" label="kubectl gadget">
+```bash
+$ kubectl gadget run snapshot_file:%IG_TAG%
+K8S.NODE        K8S.NAMESPACE   K8S.PODNAME         K8S.CONTAINERNAME   COMM   PID   TID   FD   PATH    TYPE    FLAGS   MODE      
+minikube-docker default         test-snapshot-file  test-snapshot-file  tail     5432  5432  3  demo.txt REGULAR    O_RDONLY    -rw-------
```
burak-ok: This is not aligned properly

Shaheer-Ahmd:
@burak-ok thanks a bunch for pointing the file mode issue out. I have removed it from the program and the documentation.

burak-ok (approved):
LGTM

Thanks for working on this :)

PR #4516, opened 2025-05-26
ig: add --kubeconfig option for --enrich-with-k8s-apiserver

alban:
Before this patch, --enrich-with-k8s-apiserver could only be used when ig is running in a Kubernetes pod.

With this patch, ig can be executed outside of a Kubernetes pod by specifying the kubeconfig path. When the kubeconfig path is not specified, ig will try the in-cluster config, and then it will try the default path /root/.kube/config if it does not work.

Example of command in minikube:
```
sudo ig run trace_exec:main --enrich-with-k8s-apiserver=true --kubeconfig=/etc/kubernetes/admin.conf
```

## Testing done

TODO


alban:
> Do you think we could add a test to make sure we never break this feature?

I don't know how to add such a test. We have tests with:
- IgLocalTestComponent
- KubectlGadgetTestComponent

But do we have tests with IgRemoteK8sTestComponent?

PR #4517, opened 2025-05-26
wasmapi: go: datasource: Remove uneeded subscription type.

eiffel-fl:
We can handle callback directly and check their type to call the corresponding function as done in [1].

[1]: https://github.com/inspektor-gadget/inspektor-gadget/pull/4275/files#diff-40f44fa6b449c7f56a32c72861ba5c1d280effe0de863ecc64e6c4fef0846779R336-R344

### DIFF
--- a/wasmapi/go/datasource.go
+++ b/wasmapi/go/datasource.go
@@ -119,30 +119,25 @@ const (
 	DataSourceTypeArray     DataSourceType = 2
 )
 
-type subscription struct {
-	typ subscriptionType
-	cb  any
-}
-
 var (
 	dsSubscriptionCtr = uint64(0)
-	dsSubcriptions    = map[uint64]subscription{}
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
> From performance point of view, both are equivalent:
> 
> ```
> goos: linux
> goarch: amd64
> pkg: github.com/inspektor-gadget/inspektor-gadget/pkg/operators/wasm
> cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
>                │  old.txt   │           new.txt            │
>                │   sec/op   │   sec/op    vs base          │
> WasmDataEmit-8   10.01 ± 0%   10.01 ± 0%  ~ (p=0.247 n=10)
> ```

Thanks for checking this up! LGTM then!

mauriciovasquezbernal (approved):

PR #4518, opened 2025-05-26
ig, ig-k8s: ensure that $NODE_NAME is set

alban:
"ig run --enrich-with-k8s-apiserver" requires that $NODE_NAME is set. This patch ensures that nodeName is initialized correctly.

Without this patch, the function getPodByCgroups (which allows to distinguish different containers in a single pod) will not be able to fetch only the pods for the current node. The kube-apiserver will receive requests with this URL:

    /api/v1/pods?fieldSelector=spec.nodeName%3D

On clusters with many nodes, this was a waste of resources.


mauriciovasquezbernal (approved):
LGTM. Good catch Alban!

PR #4520, opened 2025-05-27
container-collection: Handle OciConfig for initial containers

mqasimsarfraz:
This change add support for reading OciConfig for initial containers as well. 

Fixes: #3884 

### docker

```
$ make ig
$ make minikube-start-docker
$ docker cp ig minikube-docker:/bin/ig
$ minikube ssh 
$ sudo ig list-containers -o json
...
"ociConfig": "{\"ociVersion\":\"1.2.0\",\"process\":{\"user\":{\"uid\":0,\"gid\":0,\"additionalGids\":[0]},\"args\":[\"/storage-provisioner\"],\"env\":[\"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin\",\"HOSTNAME=minikube-docker\",\"KUBERNETES_PORT_443_TCP=tcp://10.96.0.1:443\",\"KUBE_DNS_SERVICE_HOST=10.96.0.10\",\"KUBE_DNS_SERVICE_PORT_DNS=53\",\"KUBE_DNS_PORT_53_UDP=udp://10.96.0.10:53\",\"KUBE_DNS_PORT_9153_TCP_PROTO=tcp\",\"KUBERNETES_SERVICE_HOST=10.96.0.1\",\"KUBERN
...
```

### containerd

```
$ make ig
$ make minikube-start-containerd
$ docker cp ig minikube-containerd:/bin/ig
$ minikube ssh 
$ sudo ig list-containers -o json
...
[...]

PR #4522, opened 2025-05-27
container-hook: introduce PreCreateContainer events

alban:
PreCreateContainer events happen before the container is created and include the OCI config. The event is included in the "containers" datasource, so gadgets have an opportunity to inspect the container id and container config.

The pid, namespaces and other fields are not available since they will be known only after the container is created.

---

This is part of #4458.

This patch does not allow gadgets to modify the container config yet. I intend to introduce that in a future PR.

## How to use

```
$ sudo ig list-containers -w
...
2025-05-27T17:30:22… PRECREATE                      3888fa4ac0652b4…                                       0                                               1970-01-01T01:00:00.000000000+01:00
2025-05-27T17:30:22… CREATED    docker              3888fa4ac0652b4… keen_cartwright                       3357032 busybox                                 2025-05-27T17:30:22.432075058+02:00
2025-05-27T17:30:27… DELETED    docker              3888fa4ac0652b4… keen_cartwright                       3357032 busybox                                 2025-05-27T17:30:22.432075058+02:00
```

## Testing done

```
$ sudo ig list-containers -w
$ sudo ig list-containers -w -o json
```

## TODO

- [x] #4531
- [x] add test in gadgets/ci/datasource-containers/test/integration/datasource_test.go (done after #4520)

### DIFF
--- a/gadgets/ci/datasource-containers/test/integration/datasource_test.go
+++ b/gadgets/ci/datasource-containers/test/integration/datasource_test.go
@@ -120,3 +120,76 @@ func TestDatasourceContainers(t *testing.T) {
 	}
 	igtesting.RunTestSteps(steps, t, testingOpts...)
 }
+
+func TestDatasourceContainersPreCreate(t *testing.T) {
+	gadgettesting.RequireEnvironmentVariables(t)
+	utils.InitTest(t)
+
+	containerFactory, err := containers.NewContainerFactory(utils.Runtime)
+	require.NoError(t, err, "new container factory")
+	containerName := "test-datasource-containers-precreate"
+	containerImage := gadgettesting.BusyBoxImage
+
+	var ns string
+	containerOpts := []containers.ContainerOption{
+		containers.WithContainerImage(containerImage),
+		containers.WithStartAndStop(),
+	}
+
+	if utils.CurrentTestComponent == utils.KubectlGadgetTestComponent {
+		ns = utils.GenerateTestNamespaceName(t, "test-datasource-containers")
+		containerOpts = append(containerOpts, containers.WithContainerNamespace(ns))
+	}
+
+	testContainer := containerFactory.NewContainer(
+		containerName,
+		"sleep inf",
+		containerOpts...,
+	)
+
+	var runnerOpts []igrunner.Option
+	var testingOpts []igtesting.Option
+
+	switch utils.CurrentTestComponent {
+	case utils.IgLocalTestComponent:
+		runnerOpts = append(runnerOpts, igrunner.WithFlags(fmt.Sprintf("-r=%s", utils.Runtime)))
+	case utils.KubectlGadgetTestComponent:
+		runnerOpts = append(runnerOpts, igrunner.WithFlags(fmt.Sprintf("-n=%s", ns)))
+		testingOpts = append(testingOpts, igtesting.WithCbBeforeCleanup(utils.PrintLogsFn(ns)))
+	}
+
+	runnerOpts = append(runnerOpts, igrunner.WithValidateOutput(
+		func(t *testing.T, output string) {
+			expectedEntry := &DatasourceContainersEvent{
+				EventType:       "PRECREATE",
+				ContainerID:     utils.NormalizedStr,
+				ContainerConfig: utils.NormalizedStr,
+			}
+
+			normalize := func(e *DatasourceContainersEvent) {
+				utils.NormalizeString(&e.ContainerID)
+
+				// only validate the OCI config is not empty before normalizing it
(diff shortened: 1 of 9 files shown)


pkg/operators/localmanager/localmanager.go
mauriciovasquezbernal, line 1: nit: perhaps we should update pkg/operators/kubemanager/kubemanager.go to be aligned with this.

gadgets/ci/datasource-containers/test/integration/datasource_test.go
```diff
@@ -120,3 +120,76 @@ func TestDatasourceContainers(t *testing.T) {
...
+		runnerOpts = append(runnerOpts, igrunner.WithFlags(fmt.Sprintf("-r=%s", utils.Runtime)))
+	case utils.KubectlGadgetTestComponent:
+		runnerOpts = append(runnerOpts, igrunner.WithFlags(fmt.Sprintf("-n=%s", ns)))
+		testingOpts = append(testingOpts, igtesting.WithCbBeforeCleanup(utils.PrintLogsFn(ns)))
+	}
+
+	runnerOpts = append(runnerOpts, igrunner.WithValidateOutput(
+		func(t *testing.T, output string) {
+			expectedEntry := &DatasourceContainersEvent{
+				EventType:       "PRECREATE",
```
mauriciovasquezbernal, line 164: Can we extend the `TestDatasourceContainers` test to look for this event too instead of adding a whole new test for it.

mauriciovasquezbernal (commented):
The implementation looks good. A question on the testing part. Please also check that the test is failing in k8s.

gadgets/ci/datasource-containers/test/integration/datasource_test.go
alban: The first test starts the container before IG. The new test starts the container after IG. I think it's good to test both scenarios. I'm not sure how to factorise the code though.

mauriciovasquezbernal: ok, it's fine then to keep both tests independently if factorizing them is complex.

alban:
> Please also check that the test is failing in k8s.

~~The issue is that [k8s.go:NewContainer()](https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.41.0/pkg/testing/containers/k8s.go#L24) is immediately starting a new container instead of returning an object with Start/Stop/Run methods to be started at the correct time, like it is done for docker/containerd. I'll add those methods and see if it helps.~~
Edit: The Start/Stop/Run methods are in [pkg/container-utils/testutils/k8s.go](https://github.com/inspektor-gadget/inspektor-gadget/blob/v0.41.0/pkg/container-utils/testutils/k8s.go).

mauriciovasquezbernal (approved):
LGTM once the tests pass!

alban:
The CI still fails. I can reproduce the failure locally on minikube with:

```
VERIFY_GADGETS=false make install/ig  kubectl-gadget minikube-deploy
GADGET_REPOSITORY=albantest.azurecr.io/gadget make -C gadgets ci/datasource-containers
sudo -E ig image push albantest.azurecr.io/gadget/ci/datasource-containers:alban_PreCreateContainer
GADGET_REPOSITORY=albantest.azurecr.io/gadget IG_RUNTIME=kubernetes IG_PATH=$PWD/kubectl-gadget make -C gadgets ci/datasource-containers/test-integration
```

I'll investigate more.

pkg/container-collection/container-collection.go
```diff
@@ -490,7 +490,8 @@ func (cc *ContainerCollection) Subscribe(key interface{}, selector ContainerSele
 	}
 	ret := []*Container{}
 	cc.pubsub.Subscribe(key, func(event PubSubEvent) {
-		if ContainerSelectorMatches(&selector, event.Container) {
+		// PRECREATE events cannot be filtered since they are not enriched yet.
+		if event.Type == EventTypePreCreateContainer || ContainerSelectorMatches(&selector, event.Container) {
 			f(event)
```
alban, line 495: This was the missing change that was required for the feature to work on kubernetes.

By default, `kubectl-gadget` only considers the containers in the "default" namespace. But PRECREATE events don't have k8s enrichment yet so they were filtered out. ig didn't have this problem because it does not have a filter on the "default" namespace.

With this, the tests pass locally. Let's see if the CI agrees.

alban:
The tests pass now. 
@mauriciovasquezbernal do you want to have another look or is it still good?

mauriciovasquezbernal:
> The tests pass now. @mauriciovasquezbernal do you want to have another look or is it still good?

Go for it!

PR #4523, opened 2025-05-27
container-collection/k8s: Set PodLabels with the method

burak-ok:
Initial K8s containers set the PodLabels directly. This changed in https://github.com/inspektor-gadget/inspektor-gadget/pull/4237 where the PodLabels ***has to be set*** through the Method `SetPodLabels`, so it updates the `podLabelsAsString` which actually gets written to the events

Fixes: 8647f0e04eb51d15673325791de22f8324a436d9

### DIFF
--- a/pkg/container-collection/k8s.go
+++ b/pkg/container-collection/k8s.go
@@ -212,10 +212,12 @@ func (k *K8sClient) GetRunningContainers(pod *v1.Pod) []Container {
 					Namespace:     pod.GetNamespace(),
 					PodName:       pod.GetName(),
 					ContainerName: s.Name,
-					PodLabels:     labels,
 				},
 			},
 		}
+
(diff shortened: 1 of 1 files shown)


pkg/container-collection/k8s.go
```diff
@@ -155,6 +155,7 @@ func (k *K8sClient) GetNonRunningContainers(pod *v1.Pod) []string {
 func (k *K8sClient) GetRunningContainers(pod *v1.Pod) []Container {
 	containers := []Container{}
 
+	fmt.Printf("Labels for pod %s/%s: %v\n", pod.GetNamespace(), pod.GetName(), pod.Labels)
```
alban, line 158: leftover debug

alban (commented):
LGTM once the leftover debug is removed.

PR #4534, opened 2025-05-28
gha:Remove update website trigger

burak-ok:
Website was moved away from Azure Static Web App in commit https://github.com/inspektor-gadget/website/commit/824c418a8be8b7b53b0bdcae1a2255eba55e7ca3

Since then this Action is failing for us in our CI


burak-ok:
> I agree with removing it, but can you please create an issue to investigate how to add it back?

What did this trigger do? It only runs on `main`.
Adding new versions to the website is done through a scheduled action in the `inspektor-gadget/website` repo. So was this for a latest commit kind of documentation?

burak-ok:
> That's right, it was for building latest which is related to the main branch

Ahh, we changed the meaning and forgot it maybe here, too?

Issue #3901, opened 2025-01-16
[RFE] Use less memory when loading kernel BTF


alban:
## Current situation

In my previous tests, cilium/ebpf's btf.LoadKernelSpec() allocate 27MB. See flamegraphs in https://github.com/cilium/ebpf/pull/1589. But most of the kernel BTF is not used, so it is a waste of memory.

## Impact

ig seems to have RAM usage of around 100-120 MB (#3541). 

Users are trying to run ig in a pod with memory limits in the pod spec but are discourage by the big limitations to write.

## Ideal future situation

ig only loads the kernel BTF necessary for the gadget.

## Implementation options

PoC https://github.com/cilium/ebpf/pull/1589

## Additional information

This issue is part of #3541.

burak-ok:
Does https://github.com/inspektor-gadget/inspektor-gadget/pull/4541 solve this issue?

mauriciovasquezbernal:
I think it does, but I'd like to get @alban thoughts.

PR #4541, opened 2025-05-30
deps: Update to go-ebpf version with lower memory usage

mauriciovasquezbernal:
I noticed the go-ebpf folks introduced an optimization for the memory usage on https://github.com/cilium/ebpf/pull/1778. 

I tried with while running the trace_dns gadget and using --pprof-addr, the results are:

### main 

Screenshot from 2025-05-30 10-49-00

### this PR 

Screenshot from 2025-05-30 10-49-36

It represents a reduction of around 30MB! 

Ref #3901 #3541


burak-ok:
Might also be because of 
- https://github.com/cilium/ebpf/pull/1763
- https://github.com/cilium/ebpf/pull/1772

burak-ok (approved):
LGTM, after the eBPF update.
Is it known when the library gets a new release?

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.24.0
 
 require (
 	github.com/blang/semver v3.5.1+incompatible
-	github.com/cilium/ebpf v0.18.0
+	github.com/cilium/ebpf v0.18.1-0.20250521101936-dd4d949f2f7b
```
alban: Why change the version here instead of using the `replace` command?

mauriciovasquezbernal: No specific reason, I though it was more common to do that than using replace, but it seems we use replace for those cases.

mauriciovasquezbernal:
> Is it known when the library gets a new release?

They don't have a fixed release schedule.

mauriciovasquezbernal:
Test failure is not related to this PR, moving ahead.

PR #4078, opened 2025-02-17
Proof of concept: gadget supporting several versions of socketenriche…

alban:
Proof of concept: gadget supporting several versions of socketenricher with CORE relocations

The socket enricher ebpf module defines some type:
```c
	// socket-enricher.bpf.c
	struct socket_enricher_foo {
		char foo[8];
	};
```
Then, the btf type is extracted from the socket enricher ebpf module:
```c
	// socketenricher/tracer.go
	reader := bytes.NewReader(_SocketenricherBytes)
	btfSpec, err := btf.LoadSpecFromReader(reader)
	se.btfSpec = btfSpec
```
And passed to the gadget ebpf module:
```c
	// operators/ebpf/ebpf.go
	opts.Programs.KernelModuleTypes = map[string]*btf.Spec{
		"ig": seBTFSpec,
	}
	collection, err := ebpf.NewCollectionWithOptions(i.collectionSpec, opts)
```
Then, cilium/ebpf is using both the kernel btf and the socketenricher btf as btf target:
```diff
--- /home/alban/go/src/github.com/cilium/ebpf/prog.go	2025-02-17 19:54:11.388743187 +0100
+++ vendor/github.com/cilium/ebpf/prog.go	2025-02-17 18:40:55.810021713 +0100
@@ -296,6 +296,8 @@
 	copy(insns, spec.Instructions)
 
 	kmodName, err := spec.kernelModule()
+	kmodName = "ig"
+	fmt.Printf("module: %s\n", kmodName)
 	if err != nil {
 		return nil, fmt.Errorf("kernel module search: %w", err)
 	}
```
This allows a gadget to use CORE relocations to identify the btf types from the socket enricher.
```c
	// gadgets/trace_dns/program.bpf.c
	struct socket_enricher_foo {
		char foo[1];
	};
	struct socket_enricher_foo___siz2 {
		char foo[2];
	};
	struct socket_enricher_foo___siz4 {
		char foo[4];
	};
	struct socket_enricher_foo___siz8 {
		char foo[8];
	};

	    if (bpf_core_type_exists(struct socket_enricher_foo)) {
	        bpf_printk("type exists: socket_enricher_foo");
	        bpf_printk("sizeof(socket_enricher_foo) = %d", bpf_core_field_size(struct socket_enricher_foo, foo));
	    }
	    if (bpf_core_type_matches(struct socket_enricher_foo___siz2)) {
	        bpf_printk("type match: socket_enricher_foo___siz2");
	    }
[...]

Issue #4047, opened 2025-02-13
[EPIC] gadgets: Compatibility guarantees


mauriciovasquezbernal:
As we approach v1.0, we need to guarantee that gadgets compiled with a version of Inspektor Gadget work with a different one. This is a key functionality as we don't want to couple the version of the gadgets and the IG version. (as it's currently done). 

## Version Guarantees

- Gadgets compiled with ig v1.X must work with ig >= v1.X
- Gadgets compiled with ig v1.X could not work with ig < 1.X

This policy guarantees that users can upgrade ig without breaking gadgets that were currently working, and also allows gadget developers to use new features introduced in ig without worrying to support older versions.

## Implementation Options 

This is a raw list of things that could cause compatibility issues

- [BPF API](https://inspektor-gadget.io/docs/latest/gadget-devel/gadget-ebpf-api)
- [Wasm API](https://inspektor-gadget.io/docs/latest/gadget-devel/gadget-wasm-api-raw)
- [Metadata file](https://inspektor-gadget.io/docs/latest/gadget-devel/metadata)

## Implementation Options

TBD

PR #4543, opened 2025-05-30
socketenricher: Use CO-RE to make it resilient against changes

mauriciovasquezbernal:
## Description 

The socketenricher operator is used  to enrich events with process information from a socket. This is implemented as an ebpf map that stores a list of tuples (IPs+ports) and the information of the process that created the socket. This map is an API contract between the gadgets and Inspektor Gadget, any change to the layout of the data on it will require the gadgets to be recompiled. Of course, this is a very big limitation.

This commit uses the existing CO-RE technology to make it possible to update the layout of this map without breaking the gadgets. This is done by generating a BTF data that represents the structure stored on this map, then this data is fed to the ebpf-go library when loading the program, ensuring it relocates the access to that structure.

This in practice allows to:
- Introduce new information to the socket enricher
- Change the type or position of existing fields on it
- Dynamically disable some expensive fields (like paths).
  - Reduce the size of the map
  - Avoid spending CPU cycles on disabled fields

## Testing

0. Let's compile the trace_dns gadget with this version of IG:

```bash
$ sudo -E ig image build ./gadgets/trace_dns/ -t trace_dns --builder-image-pull=never
```

1. Run the gadget to ensure it's working fine:

```bash
$ sudo ig run trace_dns --verify-image=false --paths --fields cwd,exepath,proc.comm,proc.creds.uid
CWD                                               EXEPATH                                           COMM                UID
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
```

I'm using `setuidgid 1000:1000 nslookup google.com` inside a busybox container to generate events.

### Adding Information to the sockets_val structure

Let's say we need to add a new field to the sockets_val structure:

```bash
$ git diff include/
diff --git a/include/gadget/types.h b/include/gadget/types.h
index 52ec5a225..c7ef0a26b 100644
--- a/include/gadget/types.h
+++ b/include/gadget/types.h
@@ -147,6 +147,7 @@ struct sockets_value {
        __u64 deletion_timestamp;
        __u32 ppid;
        char ipv6only;
+       __u32 mynewfield;

        // These fields are optional and can be disabled in the socket enricher
        // operator. Gadgets using these fields MUST check if they exist by using
```

Recompile Inspektor Gadget (DON'T RECOMPILE THE GADGET!):

```bash
$ make make ebpf-objects
$ make ig
```

And run the  gadget again:

```bash
$ sudo ig run trace_dns --verify-image=false --paths --fields cwd,exepath,proc.comm,proc.creds.uid
CWD                                               EXEPATH                                           COMM                UID
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
/                                                 /bin/nslookup                                     nslookup           1000
```

It works! What is means is that we're able to extend the sockets_val structure and a gadget compiled with an older version of IG is still working fine.

### Disabling Expensive Fields

Another feature of this PR is that it's now possible to disable expensive fields on the socket enricher, and the gadgets will continue to work fine, of course, they won't have those fields available.

Run the gadget with the default fields enabled:

```bash
$ sudo ig run trace_dns --verify-image=false --paths --fields cwd,exepath,proc.comm,proc.creds.uid
CWD                                               EXEPATH                                           COMM                UID
```

```bash
$ sudo bpftool map list name gadget_sockets
1632: hash  name gadget_sockets  flags 0x0
        key 12B  value 1104B  max_entries 16384  memlock 19438272B
        pids ig(338730)
```

The map has a a key size of 1104 bytes and uses 19438272B ~(18MB) of memory.
Now, let's disable the expensive fields:

```bash
$ sudo ig run trace_dns --verify-image=false --paths --fields cwd,exepath,proc.comm,proc.creds.uid --socket-enricher-fields=""
CWD                                               EXEPATH                                           COMM                UID
```

```bash
$ sudo bpftool map list name gadget_sockets
1666: hash  name gadget_sockets  flags 0x0
        key 12B  value 80B  max_entries 16384  memlock 2628288B
        pids ig(339607)
```

The map now has a key size of 80 bytes and uses 2628288B ~(2.5MB) of memory. Of course, the most important part is that it still works:

```bash
$ sudo ig run trace_dns --verify-image=false --paths --fields cwd,exepath,proc.comm,proc.creds.uid --socket-enricher-fields=""
CWD                                               EXEPATH                                           COMM                UID
                                                                                                    nslookup           1000
                                                                                                    nslookup           1000
                                                                                                    nslookup           1000
                                                                                                    nslookup           1000
```

Notice how the cwd and exepath fields are empty. This feature is useful in environments where memory is limited, or when users don't care about those expensive fields.

### TODO

- [x] The btf information for the sockets map is wrong, so bpftool map dump doesn't work properly.
- [x] Experiment to check if it's possible to change the size of the expensive fields instead of just disable, enable them, i.e. support #4083 in a dynamic way
- [x] Document how gadget developers should consume optional fields from the socket enricher
- [x] Tests 
- [x] Investigate if the way used to merge the BTFs could be problematic on some cases

### FAQ

- Q: How does it actually work?
- A: Magic. Jokes aside, it's based on all exiting BTF infrastructure, we're generating BTF information for the sockets_val structure and feeding it to the loader library, then all relocations are done as when the program tries to access a kernel field.

- Q: Does it depend on the kernel version?
- A: I tested it and worked fine up to 5.4, probably it doesn't work with older versions. I hit an issue when trying to compile the gadget with clang-16, but it worked fine with clang-18 that is currently used in the gadget-builder image
 
This is a evolved version of #4078 (thanks @alban for the initial idea), and one step towards #4047

### DIFF
--- a/gadgets/trace_dns/program.bpf.c
+++ b/gadgets/trace_dns/program.bpf.c
@@ -190,6 +190,11 @@ struct {
 	__uint(max_entries, 1024);
 } query_map SEC(".maps");
 
+static __always_inline unsigned int min(unsigned int a, unsigned int b)
+{
+	return a < b ? a : b;
+}
+
 SEC("socket1")
 int ig_trace_dns(struct __sk_buff *skb)
 {
@@ -365,32 +370,25 @@ int ig_trace_dns(struct __sk_buff *skb)
 	gadget_process_populate_from_socket(skb_val, &event->proc);
 
 	if (paths && skb_val != NULL) {
-		if (bpf_core_enum_value_exists(
-			    enum bpf_func_id, BPF_FUNC_probe_read_kernel_str)) {
-			bpf_probe_read_kernel_str(
-				&event->cwd, sizeof(event->cwd), skb_val->cwd);
-			bpf_probe_read_kernel_str(&event->exepath,
-						  sizeof(event->exepath),
-						  skb_val->exepath);
-		} else {
-			if (sizeof(skb_val->cwd) <= sizeof(event->cwd)) {
-				int cwd_len = sizeof(skb_val->cwd);
-				if (bpf_skb_load_bytes(
-					    skb, (unsigned long)skb_val->cwd,
-					    event->cwd, cwd_len) < 0) {
-					return 0;
-				}
+		unsigned int cwd_len =
+			min(bpf_core_field_size(skb_val->cwd), GADGET_PATH_MAX);
+		unsigned int exepath_len = min(
+			bpf_core_field_size(skb_val->exepath), GADGET_PATH_MAX);
+		bool cwd_exists = bpf_core_field_exists(skb_val->cwd);
+		bool exepath_exists = bpf_core_field_exists(skb_val->exepath);
+		bool probe_read_kernel_str_exists = bpf_core_enum_value_exists(
+			enum bpf_func_id, BPF_FUNC_probe_read_kernel_str);
+
+		if (probe_read_kernel_str_exists) {
+			if (cwd_exists) {
+				bpf_probe_read_kernel_str(&event->cwd, cwd_len,
+							  skb_val->cwd);
 			}
 
-			if (sizeof(skb_val->exepath) <=
-			    sizeof(event->exepath)) {
-				int exepath_len = sizeof(skb_val->exepath);
-				if (bpf_skb_load_bytes(
-					    skb,
-					    (unsigned long)skb_val->exepath,
-					    event->exepath, exepath_len) < 0) {
-					return 0;
-				}
+			if (exepath_exists) {
+				bpf_probe_read_kernel_str(&event->exepath,
+							  exepath_len,
+							  skb_val->exepath);
 			}
 		}
 	}
--- a/pkg/socketenricher/bpf/socket-enricher.bpf.c
+++ b/pkg/socketenricher/bpf/socket-enricher.bpf.c
@@ -94,8 +94,7 @@ static __always_inline void insert_current_socket(struct sock *sock)
 	}
 	if (bpf_core_field_exists(socket_value->exepath)) {
 		int exepath_size = bpf_core_field_size(socket_value->exepath);
-		char *exepath =
-			get_path_str(&exe_file->f_path);
+		char *exepath = get_path_str(&exe_file->f_path);
 		bpf_probe_read_kernel_str(socket_value->exepath, exepath_size,
 					  exepath);
 	}


pkg/btfhelpers/btfhelpers.go
```diff
@@ -150,3 +152,36 @@ func getSimpleType(typ btf.Type) reflect.Type {
...
+	}
+
+	iterator = spec2.Iterate()
+	for iterator.Next() {
+		if _, err := builder.Add(iterator.Type); err != nil {
+			return nil, fmt.Errorf("adding types: %w", err)
+		}
+	}
+
+	buf := make([]byte, 0, 10*1024*1024) // 10MB buffer
```
alban: Why is it necessary to allocate before?

pkg/socketenricher/tracer.go
```diff
@@ -66,6 +80,133 @@ func NewSocketEnricher() (*SocketEnricher, error) {
...
+	}
+
+	btfStruct := &btf.Struct{
+		Name:    SocketsValueName,
+		Size:    currentOffset,
+		Members: members,
+	}
+	types = append(types, btfStruct)
+
+	btfSpec, err := btfhelpers.BuildBTFSpec(types)
```
alban: This converts `types` into `btfSpec` but `btfSpec` is only used in `MergeBtfs` function. It seems the merge could be implemented in a simpler way by taking the socket enricher `[]btf.Type` directly instead of converting back and forth between btf types and btf spec.

pkg/socketenricher/bpf/socket-enricher.bpf.c
```diff
@@ -81,18 +81,24 @@ static __always_inline void insert_current_socket(struct sock *sock)
...
 
+	if (bpf_core_field_exists(socket_value->cwd)) {
+		char *cwd = get_path_str(&fs->pwd);
+		bpf_probe_read_kernel_str(socket_value->cwd,
+					  sizeof(socket_value->cwd), cwd);
+	}
+	if (bpf_core_field_exists(socket_value->exepath)) {
+		char *exepath = get_path_str(&exe_file->f_path);
+		bpf_probe_read_kernel_str(socket_value->exepath,
+					  sizeof(socket_value->exepath),
```
alban: Instead of `sizeof`, use `bpf_core_field_size` so that we can later add another parameter to the socket enricher to make the size configurable at run-time.

Ditto in `socket_value->cwd` and in `sockets-iter.bpf.c`.

Note that `include/gadget/filesystem.h` currently uses:
```
#define GADGET_PATH_MAX    512
```
So this assumes `GADGET_PATH_MAX` == `SE_PATH_MAX`. I think `filesystem.h` could be made more flexible to adapt to the size chosen at run-time.
Also, `filesystem.h` assumes that it is a power of two, so if we make it a parameter, it needs to be verified.

pkg/socketenricher/tracer.go
```diff
@@ -66,6 +80,133 @@ func NewSocketEnricher() (*SocketEnricher, error) {
...
+
+	types := []btf.Type{
+		uint8T, uint16T, uint32T, uint64T,
+		int8T, int16T, int32T, int64T,
+		charT, cString16, cString512,
+	}
+
+	// fixed fields. This needs to be aligned with the definition of sockets_val
+	// in include/gadget/types.h
+	members := []btf.Member{
```
alban: Couldn't this be parsed from `_SocketenricherBytes` or some other data from `pkg/socketenricher/socketenricher_x86_bpfel.go`?

alban (commented):
I like this. Some initial comments.

pkg/btfhelpers/btfhelpers.go
mauriciovasquezbernal: It's not, I removed it.

pkg/socketenricher/tracer.go
mauriciovasquezbernal: Good suggestion Alban, the code is much cleaner now!

pkg/socketenricher/tracer.go
mauriciovasquezbernal: Good suggestion, it's present on spec.Types.

pkg/socketenricher/bpf/socket-enricher.bpf.c
mauriciovasquezbernal: I explored it a bit and it worked fine, however I'm still missing to update the filesystem.h part.

mauriciovasquezbernal: I also updated the filesystem part, however I think we need more work on that file, but it can be done in another PR.

pkg/socketenricher/tracer.go
```diff
@@ -66,6 +101,67 @@ func NewSocketEnricher() (*SocketEnricher, error) {
...
+	srcBtfStruct := &btf.Struct{}
+	if err := se.spec.Types.TypeByName(SocketsValueName, &srcBtfStruct); err != nil {
+		return fmt.Errorf("getting BTF struct %q: %w", SocketsValueName, err)
+	}
+
+	members := []btf.Member{}
+	currentOffset := uint32(0)
+	for _, member := range srcBtfStruct.Members {
+		// stop when we find the first optional field, cwd.
+		if member.Name == "cwd" {
```
alban: Instead of hardcoding "cwd" as the first optional field, the struct could have a marker:

```c
struct sockets_value {
  //...mandatory fields...
  ...

  char optional_fields_start[0]; // zero-sized field

  //...optional fields...
  char cwd[123];
  ...
}
```

0-sized fields seems to work. Tested with:
```c
#include <stdio.h>
#include <stddef.h>

struct foo {
	int a;
	char start[0];
	char z;
};

int main() {
	printf("%ld\n", offsetof(struct foo, a));
	printf("%ld\n", offsetof(struct foo, start));
	printf("%ld\n", offsetof(struct foo, z));
	return 0;
}
```

pkg/operators/socketenricher/socketenricher.go
```diff
@@ -157,7 +193,15 @@ func (i *SocketEnricherInstance) EnrichEvent(ev any) error {
 }
 
 func (s *SocketEnricher) GlobalParams() api.Params {
-	return nil
+	return api.Params{
+		{
+			Key:          fieldsParam,
+			Title:        "Socket enricher fields",
+			Description:  "List of fields and their sizes to the enabled on the socket enricher. It uses the field0=size,field1=size,... format",
+			DefaultValue: "cwd=512,exepath=512",
```
alban: That's a breaking change in the CLI flags since gadget param `--paths` will no longer be enough to enable the feature.  Are you planning to explain this in the documentation?

alban (commented):
On a high level, this is going in the right direction. Thanks!

pkg/socketenricher/tracer.go
mauriciovasquezbernal: Good idea, also worked for me!

pkg/operators/socketenricher/socketenricher.go
mauriciovasquezbernal: I wasn't planning to explain this. Do you think we should?

docs/spec/operators/socketenricher.md
```diff
@@ -18,4 +18,13 @@ networking ones. The following information can be gotten from a socket object:
 
 ## Parameters
 
-None
+### `socket-enricher-fields`
+
+List of optional fields and their sizes to be enabled on the socket enricher
+using the field0=size,field1=size,... format. Disabling or reducing the size of
+the optional fields can reduce the memory and CPU usage of Inspektor Gadget.
+Passing 0 as the size will use the default size of 512 bytes.
```
alban: > Disabling

How can an user disable it? I guess with an empty string, but that should be more explicit.

pkg/operators/socketenricher/socketenricher.go
alban: Actually, wasn't my comment incorrect? With the current default value, cwd and exepath will be enabled by default. Before the PR, the socket enricher was an implementation detail hidden to the users. I was afraid that this PR will force users to understand the socket enricher to be able to use gadgets like trace_dns. But if the default value does not change the current behaviour, then it is fine.

mauriciovasquezbernal: > With the current default value, cwd and exepath will be enabled by default.

That's right. The default value of the `socket-enricher-fields` params enables the cwd and exepath fields with a length of 512 bytes, it's the same behavior before this PR.

> I was afraid that this PR will force users to understand the socket enricher to be able to use gadgets like trace_dns. 

If the system admin (the ones that can set the `socket-enricher-fields` param) disables those fields, them will be empty even if the user passes --paths. I think we can mention it somehow on the docs of each gadget. I'll do it once this PR is more mature.

include/gadget/filesystem.h
```diff
@@ -76,7 +76,9 @@ static __always_inline struct qstr get_d_name_from_dentry(struct dentry *dentry)
 	return BPF_CORE_READ(dentry, d_name);
 }
 
-static __always_inline void *get_path_str(struct path *path)
+// get_path_str_with_size() returns a pointer to a string representation of the
+// path. Size has to be a power of two.
```
alban: It could be tested with:

```c
if (size <= 0 || (size & (size - 1)) == 0)
    bpf_core_unreachable()
```

include/gadget/filesystem.h
```diff
@@ -162,6 +164,10 @@ static __always_inline void *get_path_str(struct path *path)
 	return &string_p->buf[buf_off];
 }
 
+static __always_inline void *get_path_str(struct path *path) {
+	return get_path_str_with_size(path, GADGET_PATH_MAX);
```
alban: Line 206 in read_full_path_of_open_file_fd:
`get_path_str(&f_path)` should be replaced with `get_path_str_with_size(&f_path, buf_len)`

pkg/socketenricher/tracer.go
```diff
@@ -43,20 +48,63 @@ const (
...
+}
+
+type Config struct {
+	Cwd     FieldConfig
+	Exepath FieldConfig
+}
+
+func isPowerOfTwo(n uint32) bool {
+	if n == 0 {
+		return true
```
alban: 0 is not a power of two. Why accept size 0?

pkg/btfhelpers/btfhelpers.go
```diff
@@ -150,3 +152,52 @@ func getSimpleType(typ btf.Type) reflect.Type {
...
+	}
+
+	builder, err := btf.NewBuilder(allTypes)
+	if err != nil {
+		return nil, fmt.Errorf("creating BTF builder: %w", err)
+	}
+
+	for _, typ := range types {
+		if _, err := builder.Add(typ); err != nil {
+			return nil, fmt.Errorf("adding types: %w", err)
```
alban: Before adding the new type, should it check for colliding names?
```go
name := typ.TypeName()
if name != "" {
    _, err := spec.AnyTypesByName(typ.TypeName())
    if err != btf.ErrNotFound {
        return nil, fmt.Errorf("type %q collide with kernel type", name)
    }
}
```

pkg/operators/ebpf/ebpf.go
```diff
@@ -628,6 +629,47 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+		// Load the programs with the spec from the socket enricher and the
+		// kernel or btfgen specs
+		seTypes, ok := seTypesI.([]btf.Type)
+		if !ok {
+			return fmt.Errorf("invalid socket enricher BTF spec: expected []btf.Type, got %T", seTypesI)
+		}
+
+		kernelSpec := opts.Programs.KernelTypes
+		if kernelSpec == nil {
+			kernelSpec, err = btf.LoadKernelSpec()
```
alban: Doesn't it defeat the optimization from https://github.com/cilium/ebpf/pull/1763?

We should check the memory allocation & usage with pprof before & after this PR.

pkg/operators/ebpf/ebpf.go
```diff
@@ -628,6 +629,47 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+				return fmt.Errorf("loading kernel BTF spec: %w", err)
+			}
+		}
+
+		mergedBtf, err := btfhelpers.AppendTypesToSpec(kernelSpec, seTypes)
+		if err != nil {
+			return fmt.Errorf("merging BTF specs: %w", err)
+		}
+
+		opts.Programs.KernelTypes = mergedBtf
```
alban: This prevents gadgets from attaching a kprobe to a function from a kernel module. I guess it might be ok, we don't have any such gadgets so far.

pkg/operators/ebpf/ebpf.go
```diff
@@ -628,6 +629,47 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
+
+		structSizeI, ok := gadgetCtx.GetVar("socketEnricherStruct")
```
alban: Shouldn't this variable be named `btfStructI`?

pkg/operators/socketenricher/socketenricher.go
```diff
@@ -83,6 +89,36 @@ func (s *SocketEnricher) CanOperateOn(gadget gadgets.GadgetDesc) bool {
...
 
 func (s *SocketEnricher) Init(params *params.Params) error {
+	fields := params.Get(fieldsParam).AsStringSlice()
+
+	s.seConfig = &tracer.Config{}
+
+	for _, field := range fields {
+		sizeStr := "512"
+
+		parts := strings.Split(field, "=")
```
alban: Why not using `SplitN`?

pkg/socketenricher/trace_optional_fields_test.go
```diff
@@ -0,0 +1,262 @@
+//go:build linux
+// +build linux
+
+// Copyright 2023 The Inspektor Gadget authors
```
alban: 2025

pkg/socketenricher/trace_optional_fields_test.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+type offsets struct {
+	cwd     fieldOffset
+	exepath fieldOffset
+}
+
+func firstN(str string, n int) string {
+	v := []rune(str)
+	if n >= len(v) {
```
alban: Why use `rune`? AFAIK, the bpf code does not care about UTF-8 strings; it just handle arrays of strings. `len(str)` seems more correct in this context.

pkg/socketenricher/tracer.go
```diff
@@ -79,26 +188,53 @@ func (se *SocketEnricher) start() error {
...
+		return fmt.Errorf("loading socket enricher asset: %w", err)
+	}
+
+	if err := se.generateTypes(); err != nil {
+		return fmt.Errorf("getting BTF spec: %w", err)
+	}
+
+	kernelSpec := btfgen.GetBTFSpec()
+	if kernelSpec == nil {
+		kernelSpec, err = btf.LoadKernelSpec()
```
alban: Ditto: is it defeating the optimization from https://github.com/cilium/ebpf/pull/1763?

pkg/socketenricher/tracer_test.go
```diff
@@ -49,14 +49,66 @@ func TestSocketEnricherStopIdempotent(t *testing.T) {
...
+
+	utilstest.RequireRoot(t)
+	utilstest.HostInit(t)
+
+	type testDefinition struct {
+		config Config
+	}
+
+	for name, test := range map[string]testDefinition{
+		"invalid_cwd_size": {
```
alban, line 71: You can add a test for `{Enabled: true, Size: 0}`.

gadgets/trace_dns/program.bpf.c
```diff
@@ -21,7 +21,7 @@
 #include <gadget/filter.h>
 
 // Don't include <gadget/filesystem.h> in networking gadgets
-#define GADGET_PATH_MAX 512
+#define GADGET_PATH_MAX 4096
```
alban: `struct event_t` and `struct event_header_t` are unfortunately not made flexible, so `cwd` and `exepath` will unnecessarily use more space by default in the perf ring buffer.

Could the ebpf operator recognise that those fields come from the socket-enricher and change their size dynamically too?

alban (commented):
Thanks! It seems to be almost ready. Most of my comments should be easy to fix, except:
- Memory footprint question with BTF loading
- Size of trace_dns events in the perf ring buffer

include/gadget/filesystem.h
mauriciovasquezbernal: Good catch, thanks!

pkg/socketenricher/tracer.go
mauriciovasquezbernal: You're right, the logic was there to allow Size == 0, but that's being handled in the socket enricher operator and no on this package.

pkg/socketenricher/tracer_test.go
mauriciovasquezbernal: Added a case for it on invalid config.

pkg/btfhelpers/btfhelpers.go
mauriciovasquezbernal: I'm not sure, perhaps we can avoid this check and let the ebpf library catch any issue when doing the core relocations? Do you see any risk on this approach?

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: This affects the memory:

before:

Screenshot from 2025-06-19 11-35-23

after:

image

I probably don't understand #1763 well enough, but I thought calling `bpf.LoadKernelSpec` won't have a negative impact as the same instance should be returned because of the cache, but it seems I'm wrong. I'll investigate more how this can be improved.

mauriciovasquezbernal: I got it! After all, appending the socket enricher types to the kernel spec wasn't a good idea, as it generated a new larger spec. I decided to push for a different approach: I extended the ebpf library to receive a list of additional targets for core relocations (https://github.com/inspektor-gadget/ebpf/commit/0e8505ba230fd498a027bc18d5684fac8f257063), and used it here. Once we decided this is a good approach, I'll propose that change to cilium/ebpf folks.

With this change, the memory usage is:

image

pkg/socketenricher/tracer.go
mauriciovasquezbernal: Let's keep the conversation on one single thread: https://github.com/inspektor-gadget/inspektor-gadget/pull/4543#discussion_r2154469845

gadgets/trace_dns/program.bpf.c
mauriciovasquezbernal: Yes, that's doable. A couple of weeks ago I was experimenting this approach in builtin trace_exec (https://github.com/inspektor-gadget/inspektor-gadget/commit/863ebbc3210f9d76fd9942d7f0de44f00c14b926), I almost got it to work. Is it fine if we explore this on another PR?

pkg/btfhelpers/btfhelpers.go
mauriciovasquezbernal: It's not relevant anymore after my latest change.

include/gadget/filesystem.h
mauriciovasquezbernal: This breaks the test on kernel 5.4. I replaced it by:

```c
	if (!is_pow_of_two(size))
		return NULL;
```

gadgets/trace_dns/program.bpf.c
alban: I am concerned that if this PR is merged in the next release but the other change in trace_dns is not, then we will have a release with a trace_dns gadget which is less efficient memory-wise, and more likely to drop events when the ring buffer gets full.

I suggest to keep `GADGET_PATH_MAX=512` in this PR, and calculate the `min(cwd_len, GADGET_PATH_MAX)` before `bpf_probe_read_kernel_str`.

In this way, the infrastructure in IG will be ready and gadgets like trace_dns can make use of it in a future release. Meanwhile, when using the default parameters, there would be no drawbacks with memory efficiency.

go.mod
```diff
@@ -209,4 +209,4 @@ replace github.com/gopacket/gopacket => github.com/gopacket/gopacket v1.3.2-0.20
 // - https://github.com/cilium/ebpf/pull/1763
 // - https://github.com/cilium/ebpf/pull/1772
 // - https://github.com/cilium/ebpf/pull/1778
-replace github.com/cilium/ebpf => github.com/cilium/ebpf v0.18.1-0.20250521101936-dd4d949f2f7b
+replace github.com/cilium/ebpf => github.com/inspektor-gadget/ebpf v0.0.0-20250619201013-0e8505ba230f
```
alban: Needs rebase & conflict fixing.

The cilium/ebpf mauricio/extra-targets branch should be rebased on top of origin/main to include https://github.com/cilium/ebpf/commit/9958a4f58fb4e533201e0a76d6d82eeeaa91434b and other optimizations.

Ditto for examples/go.mod.

docs/spec/operators/socketenricher.md
```diff
@@ -18,4 +18,14 @@ networking ones. The following information can be gotten from a socket object:
...
 ## Parameters
 
-None
+### `socket-enricher-fields`
+
+List of optional fields and their sizes to be enabled on the socket enricher
+using the field0=size,field1=size,... format. Disabling or reducing the size of
+the optional fields can reduce the memory and CPU usage of Inspektor Gadget.
+Passing 0 as the size will use the default size of 512 bytes. If a field is not
+present on the list, then it's disabled.
```
alban: It could be documented that the maximum value is 4096 and that it must be a power of 2.

include/gadget/types.h
```diff
@@ -127,20 +127,39 @@ struct gadget_process {
...
 	char task[TASK_COMM_LEN];
 	char ptask[TASK_COMM_LEN];
 	__u64 sock;
 	__u64 deletion_timestamp;
-	char cwd[SE_PATH_MAX];
-	char exepath[SE_PATH_MAX];
 	__u32 ppid;
 	char ipv6only;
+
+	char optional_fields_start[0] __attribute__((aligned(8)));
```
alban, line 152: Just curious, why does it need to be aligned?

pkg/operators/ebpf/ebpf.go
```diff
@@ -628,6 +629,38 @@ func (i *ebpfInstance) Start(gadgetCtx operators.GadgetContext) error {
...
 
+	opts := ebpf.CollectionOptions{
+		MapReplacements: mapReplacements,
+	}
+
+	if seTypesI, ok := gadgetCtx.GetVar("socketEnricherbtf"); ok {
+		gadgetCtx.Logger().Debugf("using socket enricher BTF spec from context")
+		// Load the programs with the spec from the socket enricher and the
+		// kernel or btfgen specs
+		seTypes, ok := seTypesI.([]btf.Type)
```
alban: Since you now only use `seTypes` to build the `seSpec`, the socket enricher could store the spec in the gadgetCtx. That would make less conversions when running several gadgets using a single socket enricher.

Sorry for the change since https://github.com/inspektor-gadget/inspektor-gadget/pull/4543#discussion_r2123433675

Then, btfhelpers.BuildSpec would only be used one time by the socket enricher. I am not sure the new functions in btfhelpers are worth it; they could be inlined.

pkg/socketenricher/trace_optional_fields_test.go
```diff
@@ -0,0 +1,261 @@
...
+					start: optionalFieldsStart + 512,
+					end:   optionalFieldsStart + 512 + 512,
+				},
+			},
+			expectedEvent: &expectedEvent{
+				cwd:     firstN(cwd, 511),
+				exepath: firstN(exepath, 511),
+			},
+		},
+		"cwd_32_exepath_32": {
```
alban: Could you try a shorter length such as 4?

The current exepath probably fits on 32 characters.

alban (approved):
LGTM once the remaining comments are adressed, particularly https://github.com/inspektor-gadget/inspektor-gadget/pull/4543#discussion_r2166454034

Thanks!

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: This approach isn't used anymore, so this limitation is not present.

include/gadget/types.h
mauriciovasquezbernal: There is not a strong motivation, but I think it's better to have the optional fields aligned to avoid any kind of surprises later on. (Even if CO-RE should take care of it).

gadgets/trace_dns/program.bpf.c
mauriciovasquezbernal: >I suggest to keep GADGET_PATH_MAX=512 in this PR, and calculate the min(cwd_len, GADGET_PATH_MAX) before bpf_probe_read_kernel_str.

Unfortunately it causes a verifier error. I'd say to hold this PR until I found a solution for it. I don't want to limit it to 512 bytes everywhere and then extend it to 4096, that'd be confusing.

pkg/socketenricher/bpf/socket-enricher.bpf.c
mauriciovasquezbernal: I think the change on the filesytem.h was wrong, if size is set to small values `get_path_str` starts to behave in a weird way, like only returning the last component of the path, making a path that is a mix of the first components and the last one, removing the first component, etc. I think it's safer to keep this value as 4096 and let the caller copy only what's needed.

pkg/socketenricher/trace_optional_fields_test.go
mauriciovasquezbernal: I tried with 4 and I found https://github.com/inspektor-gadget/inspektor-gadget/pull/4543#discussion_r2167387407. After fixing it, it now works with 4.

gadgets/trace_dns/program.bpf.c
mauriciovasquezbernal: Nevermind, it seems with the latest changes it works!

gadgets/trace_dns/program.bpf.c
```diff
@@ -190,6 +190,11 @@ struct {
 	__uint(max_entries, 1024);
 } query_map SEC(".maps");
 
+static __always_inline int min(int a, int b)
```
alban: `bpf_core_field_size` seems to return an unsigned int:

```
#define bpf_core_field_size(field)                                          \
        __builtin_preserve_field_info(field, BPF_FIELD_BYTE_SIZE)
```

https://github.com/llvm/llvm-project/blob/40cc4379cda6e0d6efe72c55d1968f9cf427a16a/clang/include/clang/Basic/BuiltinsBPF.td#L15

Should the `min()` signature be changed to unsigned `(int a, unsigned int b)`?

include/gadget/filesystem.h
```diff
@@ -6,7 +6,7 @@
 
 // clang-format off
 #define MAX_PERCPU_BUFSIZE (1 << 15)  // set by the kernel as an upper bound
-#define GADGET_PATH_MAX    512        // smaller than PATH_MAX to reduce memory consumption
+#define GADGET_PATH_MAX     4096
```
alban: This constant is also used for `struct event` in gadgets trace_exec, trace_mount, fdpass, fsnotify, top_file. So this is increasing the size of events in the ring buffer.

I think we need a different define for gadgets and for filesystem.h's internals.

gadgets/trace_dns/program.bpf.c
mauriciovasquezbernal: I updated it to be `unsigned int min(unsigned int a, unsigned int b)` as `bpf_probe_read_kernel_str` also takes an `unsigned int` as size.

include/gadget/filesystem.h
mauriciovasquezbernal: Good catch! You're right, the memory usage was being increased a lot:

```bash 
$ sudo gadgetctl run bpfstats
GADGE… GADGETNAME   GADGETIMAGE  PRO… PROGNAME     PROGTYPE            RUNTIME RUNCOUNT MAPMEM… MA… COMMS       PIDS
                    trace_exec      0                                      0ns        0   24 MB   7
```

After your suggestion it's back to the same size as main:

```bash 
$ sudo gadgetctl run bpfstats
GADGE… GADGETNAME   GADGETIMAGE  PRO… PROGNAME     PROGTYPE            RUNTIME RUNCOUNT MAPMEM… MA… COMMS       PIDS
                    trace_exec      0                                      0ns        0  9.3 MB   7
```

Issue #4416, opened 2025-05-02
operators: Fix missing default values for image layer operators


mauriciovasquezbernal:
Implement a more generic solution for the issue fixed in #4409

PR #4544, opened 2025-05-30
gadgets: Add dummy unit tests for all gadgets

mauriciovasquezbernal:
Add dummy unit tests to verify that at least those gadgets are running fine on different kernel versions. For sure, we need to extend them later on to verify the data provided by the gadget.

### DIFF
--- a/gadgets/profile_blockio/test/unit/profile_blockio_test.go
+++ b/gadgets/profile_blockio/test/unit/profile_blockio_test.go
@@ -1,0 +1,33 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"testing"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+)
+
+func TestProfileBlockio(t *testing.T) {
+	// TODO: This is a dummy test to check that the gadget runs without errors.
+	// It should be extended to check that the gadget produces correct data.
+
+	// TODO: This is the minimum kernel version the gadget works on among the
+	// ones we test. We need to check if it works with other versions that we
+	// don't test, like 5.9
+	gadgettesting.MinimumKernelVersion(t, "5.10")
+
+	gadgettesting.DummyGadgetTest(t, "profile_blockio")
+}
--- a/gadgets/profile_cpu/test/unit/profile_cpu_test.go
+++ b/gadgets/profile_cpu/test/unit/profile_cpu_test.go
@@ -1,0 +1,27 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"testing"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+)
+
+func TestProfileCpu(t *testing.T) {
+	// TODO: This is a dummy test to check that the gadget runs without errors.
+	// It should be extended to check that the gadget produces correct data.
+	gadgettesting.DummyGadgetTest(t, "profile_cpu")
+}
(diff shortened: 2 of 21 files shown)


alban (commented):

| Benchmark suite | Current: 2e57d8bea2157f238b5fde1b398dd7be8962f8c1 | Previous: dbd8e44f2d302831465782563cc2067bbdf301c8 | Ratio |
|-|-|-|-|
| `BenchmarkAllGadgetsWithContainers/container0/trace-sni` | `50462508` ns/op | `16445470` ns/op | `3.07` |

CC: @alban

gadgets/testing/testing.go
```diff
@@ -124,3 +126,19 @@ func InitUnitTest(t testing.TB) {
 	utilstest.RequireRoot(t)
 	RemoveMemlock(t)
 }
+
+// DummyGadgetTest runs a dummy gadget test that only checks if the gadget
+// can be started without errors.
+func DummyGadgetTest(t *testing.T, gadgetName string) {
```
eiffel-fl: Should this function takes a kernel version as argument and skips in case the running kernel is older?
Or should this be part of each gadget unit tests?

mauriciovasquezbernal: I added them as part of each unit test, so it's easier to understand the version needed. I also pushed some other fixes to hopefully make this pass on all kernels, let's see.

gadgets/traceloop/test/unit/traceloop_test.go
```diff
@@ -0,0 +1,32 @@
...
+	_ "github.com/inspektor-gadget/inspektor-gadget/pkg/operators/localmanager"
+)
+
+func TestTraceloop(t *testing.T) {
+	// TODO: This is a dummy test to check that the gadget runs without errors.
+	// It should be extended to check that the gadget produces correct data.
+
+	paramValues := map[string]string{
+		"operator.oci.wasm.syscall-filters": "",
+	}
```
eiffel-fl: Isn't it empty by default?

mauriciovasquezbernal: Default values aren't passed when using the Golang API (#4416), so we need to set this explicitly to avoid an error on the program.go logic that tries to get this parameter.

gadgets/profile_blockio/test/unit/profile_blockio_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+)
+
+func TestProfileBlockio(t *testing.T) {
+	// TODO: This is a dummy test to check that the gadget runs without errors.
+	// It should be extended to check that the gadget produces correct data.
+
+	// TODO: This is the minimum kernel version among the ones we test. We need
+	// to confirm what is the actual minimum kernel version for this gadget.
```
eiffel-fl: This comment is not true, or this is wrong copy/pasting, as we test also for 5.4:
https://github.com/inspektor-gadget/inspektor-gadget/blob/968b26f465982cf4798753ce19a23eab7d63ef2e/.github/workflows/inspektor-gadget.yml#L1639

gadgets/trace_dns/test/unit/trace_dns_test.go
```diff
@@ -0,0 +1,44 @@
...
+		Image:          "trace_dns",
+		Timeout:        5 * time.Second,
+		MntnsFilterMap: utilstest.CreateMntNsFilterMap(t, runner.Info.MountNsID),
+		ParamValues: map[string]string{
+			"operator.oci.ebpf.paths": "true",
+		},
+	}
+
+	gadgetRunner := gadgetrunner.NewGadgetRunner(t, opts)
+	gadgetRunner.RunGadget()
```
eiffel-fl: Why cannot you use DummyGadgetTest() here?

mauriciovasquezbernal: Good catch. I implemented DummyGadgetTest() after this and I forgot to update it.

gadgets/profile_blockio/test/unit/profile_blockio_test.go
mauriciovasquezbernal: What I wanted to say is that we aren't testing all kernel versions, so we don't know if this gadget works with 5.9 for instance, as we only test 5.4, 5.10, etc...

I updated the comment to (hopefully) be more clear

gadgets/profile_cpu/test/unit/profile_cpu_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"testing"
+
+	gadgettesting "github.com/inspektor-gadget/inspektor-gadget/gadgets/testing"
+)
+
+func TestProfileCpu(t *testing.T) {
+	// TODO: This is a dummy test to check that the gadget runs without errors.
+	// It should be extended to check that the gadget produces correct data.
+	gadgettesting.DummyGadgetTest(t, "profile_cpu")
```
alban, line 26: This test file is duplicated many times with small variations. It could be deduplicated by using a symlink:
```
gadgets/profile_cpu/test/unit/profile_cpu_test.go -> ../../../unit_test.go.template
```

The gadget name could be found with something like:
```go
	_, file, _, _ := runtime.Caller(0)
	paths := strings.Split(file, string(filepath.Separator))
	gadget := paths[len(paths)-4]
```

And the minimum kernel version could be fetched with:
```go
minimumKernelVersion, err := os.ReadFile("testdata/minimum-kernel-version.txt")
if err != nil {
	gadgettesting.MinimumKernelVersion(t, strings.TrimSpace(string(minimumKernelVersion)))
}
```

mauriciovasquezbernal: The goal is to extend these files later on implementing real tests, so I think it doesn't hurt to keep them as independent files.

alban: Ok, you can go ahead then :)

PR #4545, opened 2025-05-31
Set visibility of formatted columns

### LINKED ISSUE #4459: [RFE] operator/formatters:  Allow controlling visibility of fields
Currently all the fields added by [formatters operator](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/operators/formatters/formatters.go) have hard coded visibility resulting in gadget developer not able to control the visibility of certain fields. For example:

| Type | Hidden |
|--------|--------|
| l4endpoint | False |
| l3endpoint | False |
| duration | True |
| bytes | True |
| timestamp | True |
| errno | True |  

## Impact

Gadget developer not able to control visibility of certain fields

## Ideal future situation

Gadget developer should be able to control the visibility of all the fields

## Implementation options

We can look at the annotations of the field being formatted and if it contains `columns.hidden` we can toggle the visibility if needed. Having said that, should we respect the current hard coded fields? I think visibility should only be controlled via metadata file.

AdvH039:
Fixes #4459

### DIFF
--- a/docs/gadget-devel/gadget-ebpf-api.md
+++ b/docs/gadget-devel/gadget-ebpf-api.md
@@ -219,6 +219,8 @@ The IP field will be translated to its string representation:
   },
 ```
 
+By default this field is shown but you can use `columns.hidden` to the control visibility.
+
 ### `struct gadget_l4endpoint_t`
 
 Represents a layer 4 endpoint (IP address + port) for UDP and TCP. It's defined as:
@@ -245,6 +247,7 @@ It'll produce an output like:
 
 or `172.17.0.2:46076` depending on the output mode used.
 
+By default this field is shown but you can use `columns.hidden` to the control visibility.
 
 ### `gadget_timestamp`
 
@@ -267,7 +270,7 @@ It will produce the following output when using the `json` output mode:
   "timestamp_raw": 1721943247136974800,
 ```
 
-Notice both fields will be hidden by default when using the `columns` output mode.
+By default this field is hidden but you can use `columns.hidden` to the control visibility.
 
 #### Annotations
 
@@ -289,6 +292,8 @@ struct event {
   "sig_raw": 23,
 ```
 
+By default this field is hidden but you can use `columns.hidden` to the control visibility.
+
 #### Annotations
 
 - `formatters.signal.target`: Name of the new field. If the annotation is not set and the source field name has a _raw suffix, the target name will be set to the source name without that suffix.
@@ -308,6 +313,8 @@ struct event {
   "error_raw": 2,
 ```
 
+By default this field is hidden but you can use `columns.hidden` to the control visibility.
+
 #### Annotations
 
 - `formatters.errno.target`: Name of the new field. If the annotation is not set and the source field name has a `_raw` suffix, the target name will be set to the source name without that suffix.
@@ -326,6 +333,8 @@ struct event {
   "syscall": "SYS_SOCKET",
   "syscall_raw": 41,
 ```
+
+By default this field is hidden but you can use `columns.hidden` to the control visibility.
 
 #### Annotations
 
--- a/pkg/operators/formatters/formatters.go
+++ b/pkg/operators/formatters/formatters.go
@@ -154,10 +154,7 @@ func handleL3Endpoint(in datasource.FieldAccessor) (func(entry datasource.Data) 
 
 	in.AddAnnotation(datasource.ColumnsReplaceAnnotation, addrF.FullName())
 
-	// Hide all subfields
-	in.SetHidden(true, true)
-	// Show only root field that will contain the pretty address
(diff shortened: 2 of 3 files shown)


pkg/utils/annotations/annotations.go
```diff
@@ -42,3 +42,25 @@ func GetTargetNameFromAnnotation(
...
+
+func SetColumnsHiddenAnnotation(
+	defaultValue bool,
+	in datasource.FieldAccessor,
+) {
+	annotations := in.Annotations()
+
+	in.SetHidden(true, true)
+
+	outName, ok := annotations["columns.hidden"]
```
mqasimsarfraz: Use the const from here: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/metadata/v1/annotations.go#L23

pkg/utils/annotations/annotations.go
```diff
@@ -42,3 +42,25 @@ func GetTargetNameFromAnnotation(
 
 	return "", fmt.Errorf("neither %q annotation nor '_raw' suffix found", targetAnnotation)
 }
+
```
mqasimsarfraz, line 46: Perhaps you can have a comment saying for fields with subfields, it will only effects root field or something similar.

pkg/operators/formatters/formatters.go
```diff
@@ -154,10 +154,7 @@ func handleL3Endpoint(in datasource.FieldAccessor) (func(entry datasource.Data)
 
 	in.AddAnnotation(datasource.ColumnsReplaceAnnotation, addrF.FullName())
 
-	// Hide all subfields
-	in.SetHidden(true, true)
-	// Show only root field that will contain the pretty address
-	in.SetHidden(false, false)
+	annotations.SetColumnsHiddenAnnotation(true, in)
```
mqasimsarfraz:
```suggestion
	annotations.SetColumnsHiddenAnnotation(false, in)
```

mqasimsarfraz (commented):
Looking good, some comments! 

Can you please ensure the default values for fields is documented here: https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/gadget-devel/gadget-ebpf-api.md

AdvH039:
@mqasimsarfraz I noticed that `bytes` and `duration` fields are not documented. Should I create an issue for the same?

docs/gadget-devel/gadget-ebpf-api.md
```diff
@@ -245,6 +249,10 @@ It'll produce an output like:
 
 or `172.17.0.2:46076` depending on the output mode used.
 
+#### Annotations
```
mqasimsarfraz: I would rather add a sentence because this annotation doesn't exist by default:

By default this field is hidden but you can use `columns.hidden` to the control visibility.

pkg/utils/annotations/annotations.go
```diff
@@ -42,3 +43,26 @@ func GetTargetNameFromAnnotation(
...
+// SetColumnsHiddenAnnotation controls only the visibility of the root field for fields having subfields.
+func SetColumnsHiddenAnnotation(
```
mqasimsarfraz:
```suggestion
// SetFieldVisibility sets the visibility of a field using metadatav1.ColumnsHiddenAnnotation
// For a field with subfields, it will only set the visibility of the root field
func SetFieldVisibility(
```

mqasimsarfraz:
> @mqasimsarfraz I noticed that `bytes` and `duration` fields are not documented. Should I create an issue for the same?

Yes please, thanks!

docs/gadget-devel/gadget-ebpf-api.md
```diff
@@ -651,4 +660,4 @@ common information.
 - `void gadget_process_populate(struct gadget_process *p)`: Fill `p` with
   the current process information
 - `void gadget_process_populate_from_socket(const struct sockets_value *skb_val, struct gadget_process *p)`:
-  Fill `p` with the information on `skb_val` returned by `gadget_socket_lookup()`.
+  Fill `p` with the information on `skb_val` returned by `gadget_socket_lookup()`.
```
mqasimsarfraz: nit: Please keep the newline here.

pkg/utils/annotations/annotations.go
```diff
@@ -42,3 +43,27 @@ func GetTargetNameFromAnnotation(
...
+	if !ok {
+		in.SetHidden(defaultValue, false)
+	} else {
+		hide := false
+
+		if outName == "true" {
+			hide = true
+		}
+
+		in.SetHidden(hide, false)
```
mauriciovasquezbernal: very minor nit while looking around. (I won't fully review it)

```suggestion
		in.SetHidden(outName == "true", false)
```

PR #4552, opened 2025-06-03
gadget: Introduce categories

mqasimsarfraz:
## Problem Statement:
Currently the focus of the gadgets have been to debug/inspect systems and workloads, allowing users to use gadgets with a guarantee it won't add/change anything on their system. While this is valuable it stops us from covering following use-cases where:

- User want to generate traffic to complete the troubleshooting workflow.
- User want to drop certain packets for chaos engineering.
- User want to modify the packets for testing
- User wants to block certain syscalls for security reasons.

Considering the above use-cases are valuable we should consider introducing a concept to be able to categorize the gadgets, allowing users/admins to control which type of gadgets are allowed to run on their system.

## Solution:
To solve the above problem we will introduce categories for gadgets. The idea here is to ensure each gadget has a category, allowing admins to control which type category should be allowed. Current proposed categories are:

- `observer` -  A gadget that only observes the system.
- `generator` - A gadget that generate events on the system. 
- `mutator` - A gadget that alter/modifies the system.

by default only `observer` will be allowed.

## Goals
- A gadget **must** be put into a category at build time.
- An Inspektor Gadget admin **must** have a control over which category of gadgets are allowed via global params.
- Inspektor Gadget **must** reject gadgets other the observer category by default.
- A gadget **must** be rejected if it doesn't have a category or isn't allowed by the admin.
- A gadget developer **should** be easily able to identify the category of their gadget.
- An Inspektor Gadget admin **should** be easily able to control which category of gadgets are allowed.

### Out of Scope

The current PR doesn't plan to include any enforcement. Having said that, given we have image verification in place, ensuring correct category will be enforced for official gadgets.

## Implementation Details

Currently, there are three ways to set gadget category:

- `gadget-category` CLI flag (for wasm)
-  Using `category` field in gadget.yaml
-  Using macro in program.bpf.c

Gadgets w/o the category will be rejected at build time. At runtime ocihandler will ensure only allowed gadgets are executed as configured by global params.

## Alternative Implementation:

An alternative implementation can be to use the idea of tags as suggested by @alban / @flyth. Meaning we don't have any fixed categories but allow users to have custom tags in their gadgets and these tags can be used to control which gadgets are allowed to run. For official gadgets, we can have a some well-defined tags to correctly categorize the gadgets but I am afraid the flexibility here might create a confusion both for users and admins?

## Todo

- [ ] Include/Fix tests
- [ ] Deployment/Charts
- [ ] Documentation - Include conventions to encourage users to name their gadgets in a way that category is clear.

This will be a **breaking change**, meaning we won't be able to run gadgets build without there changes.


alban:
Will the category stay declarative (gadget authors decide how to categorize their gadgets; ig trusts the category), or will there be some enforcement (observers cannot use some kind of bpf programs like tc, xdp, and they cannot use some helper functions like `bpf_send_signal()`)?

mayasingh17:
this is awesome, thanks for providing this proposal

mqasimsarfraz:
> Will the category stay declarative (gadget authors decide how to categorize their gadgets; ig trusts the category), or will there be some enforcement (observers cannot use some kind of bpf programs like tc, xdp, and they cannot use some helper functions like `bpf_send_signal()`)?

In the current scope of the PR I don't plan to implement the enforcement mainly because it will be hard to make any guarantees but in future we should definitely consider this.  Having said that, in this PR I will include all the documentation that should aid in categorizing the gadget and include the recommendation you mentioned.

alban:
I'd like if this was done with tags rather than categories.
In #4458 (see also #4522), I am adding the ability to change the configuration of containers (just-in-time patching OCI configs). The gadgets making use of it should have the "containers-edit" tag, but it is still an observer otherwise. It would not drop packet or generate any events.

flyth:
+1 to tags vs. categories from me. That would offer a lot more flexibility.

We can use tags for both preventing certain gadgets to be run (as configured by the admin, like with categories) as well as prevent a gadget from using features it hasn't declared using tags.

PR #4556, opened 2025-06-04
gadgets/Makefile: Make GADGETS variable configurable

burak-ok:
When I want to build, push, run tests, ... a subset of gadgets I currently need to do the following:

```bash
$ make -C gadgets trace_exec
$ make -C gadgets trace_open
$ ...

$ make -C gadgets trace_exec/test_unit
$ make -C gadgets trace_open/test_unit
```

With the commit in this PR I can easily do all the steps in one:
```bash
$ GADGETS="trace_exec trace_open" make -C gadgets
$ GADGETS="trace_exec trace_open" make -C gadgets test-unit
```

This has the additional benefit for `test-unit` and `test-integration`, where it only runs the tests for gadgets that actually have the corresponding tests

### DIFF
--- a/gadgets/Makefile
+++ b/gadgets/Makefile
@@ -20,7 +20,7 @@ DOCKER ?= docker
 UNIT_TEST_DIR = test/unit
 INTEGRATION_TEST_DIR = test/integration
 
-GADGETS = \
+GADGETS ?= \
 	advise_networkpolicy \
 	advise_seccomp \
 	audit_seccomp \


mauriciovasquezbernal (approved):
I have nothing against this change, so I'm approving it, but I want to discuss a couple of things:

> When I want to build, push, run tests, ... a subset of gadgets I currently need to do the following:
> 
> ```shell
> $ make -C gadgets trace_exec
> $ make -C gadgets trace_open
> $ ...
> 
> $ make -C gadgets trace_exec/test_unit
> $ make -C gadgets trace_open/test_unit
> ```

Why not to run it in a single command with:

```bash
$ make -C gadgets trace_exec/test-unit trace_open/test-unit
```

> This has the additional benefit for `test-unit` and `test-integration`, where it only runs the tests for gadgets that actually have the corresponding tests

I'm not sure I followed this.

burak-ok:
> > This has the additional benefit for `test-unit` and `test-integration`, where it only runs the tests for gadgets that actually have the corresponding tests
> 
> I'm not sure I followed this.

For example I am writing a script or pipeline and only want to deal with the gadgets `deadlock` (it has no unittest, no integration test), `trace_exec` (unittest and integration test are available), `advise_networkpolicy` (it has no unittest, but an integration test).

With this change I can do
```bash
$ GADGETS="deadlock trace_exec advise_networkpolicy" make -C gadgets test-unit
$ GADGETS="deadlock trace_exec advise_networkpolicy" make -C gadgets test-integration
```
and the Makefile will only run the gadget tests, when they are available, thanks to this wildcard:
https://github.com/inspektor-gadget/inspektor-gadget/blob/371e9531a8f728173b6e227df83f3e04b6dfac53/gadgets/Makefile#L186

But when I try to run for example unit test for deadlock (which has none) I would get the error:
```bash
$ make -C gadgets deadlock/test-unit
...
pattern ./deadlock/test/unit/...: lstat ./deadlock/test/unit/: no such file or directory
FAIL	./deadlock/test/unit/... [setup failed]
...
```

mauriciovasquezbernal:
That makes sense, thanks for clarifying.

PR #4560, opened 2025-06-04
Document kernel version

### LINKED ISSUE #4527: [RFE] Add kernel requirements for image-based gadgets
## Current situation

The kernel requirements are only documented for built-in gadgets and image-based gadgets doesn't have any info about it.

## Impact

If I navigate the image-based documentation, I have no idea what kernel version a given gadget needs.

## Ideal future situation

Kernel requirements are available in image-based gadgets documentation.

## Implementation options

Move the https://inspektor-gadget.io/docs/latest/gadgets/builtin/requirements/ to the image-based docs and add the kernel every gadget requires in its own documentation.

AdvH039:
Fixes #4527

### DIFF
--- a/gadgets/advise_networkpolicy/README.mdx
+++ b/gadgets/advise_networkpolicy/README.mdx
@@ -10,6 +10,12 @@ The advise networkpolicy gadget monitors the network activity in the specified n
 and records a summary of TCP and UDP traffic. This is then used to generate Kubernetes
 network policies.
 
+## Requirements
+
+- Minimum Kernel Version : *5.4
+
+*This is the minimal kernel version we have tried for this Gadget, however it's possible that it works with earlier versions.
+
 ## Getting started
 
 
--- a/gadgets/top_blockio/README.mdx
+++ b/gadgets/top_blockio/README.mdx
@@ -13,6 +13,10 @@ This gadget requires Linux Kernel Version 6.5+.
 
 Note: If you have an encrypted disk some block IO operations will not be shown, since device mappers are in between these operations to operate transparently. See https://docs.kernel.org/admin-guide/device-mapper/dm-crypt.html
 
+## Requirements
+
+- Minimum Kernel Version : 6.6
+
 ## Getting started
 
 Running the gadget:
--- a/gadgets/trace_tcpdrop/README.mdx
+++ b/gadgets/trace_tcpdrop/README.mdx
@@ -10,6 +10,10 @@ import TabItem from '@theme/TabItem';
 
 The trace_tcpdrop gadget tracks TCP kernel-dropped packets/segments.
 
+## Requirements
+
+- Minimum Kernel Version : 6.1
+
 ## Getting started
 
 Running the gadget:
--- a/gadgets/trace_tcpretrans/README.mdx
+++ b/gadgets/trace_tcpretrans/README.mdx
@@ -10,6 +10,12 @@ import TabItem from '@theme/TabItem';
 
 The trace_tcpretrans gadget tracks TCP retransmissions.
 
+## Requirements
+
+- Minimum Kernel Version : *5.4
+
+*This is the minimal kernel version we have tried for this Gadget, however it's possible that it works with earlier versions.
+
 ## Getting started
 
 Running the gadget:
--- a/gadgets/README.template
+++ b/gadgets/README.template
@@ -10,6 +10,10 @@ import TabItem from '@theme/TabItem';
 
 {{ (datasource "gadget").description }}
 
+## Requirements
+
+TODO
+
 ## Getting started
 
 Running the gadget:
--- a/gadgets/advise_seccomp/README.mdx
+++ b/gadgets/advise_seccomp/README.mdx
@@ -11,6 +11,10 @@ The seccomp profile advisor gadget records syscalls that are issued in a
 specified container, and then uses this information to generate the corresponding
 seccomp profile.
 
+## Requirements
+
+- Minimum Kernel Version : 5.6
+
 ## Getting started
 
 <Tabs groupId="env">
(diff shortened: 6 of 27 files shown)


alban:
- [ ] Could you please rebase? #4599 and #4601 fixed the symlink for the profile_cpu gadget.
- [ ] Update gadgets/README.template

> Do we really want to document the _CONFIG? IHMO it's difficult to check what those requirements are and to maintain them, so I'd rather avoid it. (Unless a Gadget uses a not very common option)

I agree, I think it is better to remove it. Specially for the common ones like `KPROBES`, `KRETPROBES`. The current PR does not seem to be correct (e.g. capabilities gadget uses kretprobes).

eiffel-fl:
> * [ ]  Could you please rebase? [docs: profile_cpu: fix symlink #4599](https://github.com/inspektor-gadget/inspektor-gadget/pull/4599) and [docs: Fix symlink for profile_cpu #4601](https://github.com/inspektor-gadget/inspektor-gadget/pull/4601) fixed the symlink for the profile_cpu gadget.
> 
>     * [ ]  Update gadgets/README.template
> 
> 
> > Do we really want to document the _CONFIG? IHMO it's difficult to check what those requirements are and to maintain them, so I'd rather avoid it. (Unless a Gadget uses a not very common option)
> 
> I agree, I think it is better to remove it. Specially for the common ones like `KPROBES`, `KRETPROBES`. The current PR does not seem to be correct (e.g. capabilities gadget uses kretprobes).

I do not think there were really useful for users.
Moreover, these options were documented something like 2 or 3 years ago, when they were not that common.
Right now, I guess a majority of distro kernels have them set, so let's not add them.

One problem though with kernel version concerns kernel with backported patches, but in this case I would let the user handles their own kernel.

gadgets/advise_networkpolicy/README.mdx
```diff
@@ -10,6 +10,14 @@ The advise networkpolicy gadget monitors the network activity in the specified n
 and records a summary of TCP and UDP traffic. This is then used to generate Kubernetes
 network policies.
 
+## Requirements
+
+| Minimum Kernel |
+|----------------|
+| *5.4           |
```
eiffel-fl: Do we really need a table here?
It made sense before because this was a table for all gadgets, but now what we write this in each file, I am wondering.

mauriciovasquezbernal: I was wondering the same! I think we can just use a list:

```
- Minimum kernel version: foo
 ```

File: gadgets/advise_networkpolicy/README.mdx [hunk 2]
```diff
@@ -10,6 +10,12 @@ The advise networkpolicy gadget monitors the network activity in the specified n
 and records a summary of TCP and UDP traffic. This is then used to generate Kubernetes
 network policies.
 
+## Requirements
+
+Minimum Kernel Version : *5.4
```
mauriciovasquezbernal:
```suggestion
- Minimum Kernel Version : *5.4
```

gadgets/top_blockio/README.mdx
```diff
@@ -13,6 +13,10 @@ This gadget requires Linux Kernel Version 6.5+.
 
 Note: If you have an encrypted disk some block IO operations will not be shown, since device mappers are in between these operations to operate transparently. See https://docs.kernel.org/admin-guide/device-mapper/dm-crypt.html
 
+## Requirements
+
+- Minimum Kernel Version : 6.5
```
mauriciovasquezbernal: Where is this value coming from? I see the test declares 5.6: https://github.com/inspektor-gadget/inspektor-gadget/blob/5c743c64a6d959ed7a618d70062d1bee8cb8f7cb/gadgets/top_blockio/test/unit/top_blockio_test.go#L30

gadgets/trace_tcpdrop/README.mdx
```diff
@@ -10,6 +10,10 @@ import TabItem from '@theme/TabItem';
 
 The trace_tcpdrop gadget tracks TCP kernel-dropped packets/segments.
 
+## Requirements
+
+- Minimum Kernel Version : 6.1
```
mauriciovasquezbernal, line 15: Where is this coming from?

gadgets/trace_tcpretrans/README.mdx
```diff
@@ -10,6 +10,10 @@ import TabItem from '@theme/TabItem';
 
 The trace_tcpretrans gadget tracks TCP retransmissions.
 
+## Requirements
+
+- Minimum Kernel Version : 4.15
```
mauriciovasquezbernal: ditto.

mauriciovasquezbernal (commented):
Just to be sure we're on the same page, all the ones declaring 5.4 is because that's the minimal kernel version we test, or am I missing something?

gadgets/trace_tcpretrans/README.mdx
AdvH039: From https://inspektor-gadget.io/docs/latest/gadgets/builtin/requirements/ here

gadgets/trace_tcpdrop/README.mdx
AdvH039: [here](https://github.com/inspektor-gadget/inspektor-gadget/blob/8c6eae1947a1d6eea84fe0259fe1aaebbf4261c4/gadgets/trace_tcpdrop/test/unit/trace_tcpdrop_test.go#L30)

gadgets/top_blockio/README.mdx
AdvH039: Typo, mb sry. You mean *declares 6.6 right?

mauriciovasquezbernal: Yes, 6.6 (another typo)

mauriciovasquezbernal (approved):
LGTM, thanks for working on this!

PR #4562, opened 2025-06-04
update: allow ig to have rust support

Daksh-10:
This will allow CI to get the updated changes and execute rust tests correctly.

### DIFF
--- a/Dockerfiles/gadget-builder.Dockerfile
+++ b/Dockerfiles/gadget-builder.Dockerfile
@@ -2,6 +2,7 @@ ARG CLANG_LLVM_VERSION=18
 ARG BPFTOOL_VERSION=v7.3.0
 ARG LIBBPF_VERSION=v1.3.0
 ARG GOLANG_VERSION=1.24.2
+ARG RUST_VERSION=1.87.0
 
 # Args need to be redefined on each stage
 # https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
@@ -29,6 +30,7 @@ RUN \
 FROM debian:bookworm-slim@sha256:90522eeb7e5923ee2b871c639059537b30521272f10ca86fdbbbb2b75a8c40cd
 ARG CLANG_LLVM_VERSION
 ARG GOLANG_VERSION
+ARG RUST_VERSION
 # libc-dev is needed for various headers, among others
 # /usr/include/arch-linux-gnu/asm/types.h.
 # We make clang aware of these files through CFLAGS in Makefile.
@@ -59,6 +61,19 @@ RUN ARCH=$(dpkg --print-architecture) \
 	rm go${GOLANG_VERSION}.linux-${ARCH}.tar.gz && \
 	ln -s /usr/local/go/bin/go /usr/local/bin/go && \
 	chmod +x /usr/local/go/bin/go
+
+# Install rust, this was taken from:
+# https://github.com/rust-lang/rustup/blob/ece5ff09f126/rustup-init.sh#L162
+# We also remove components we do not need for building gadgets.
+RUN ARCH=$(dpkg --print-architecture | perl -pi -e 's/amd64/x86_64/' | perl -pi -e 's/arm64/aarch64/') \
+	&& wget --quiet https://static.rust-lang.org/rustup/dist/${ARCH}-unknown-linux-gnu/rustup-init && \
+	chmod +x rustup-init && \
+	./rustup-init -y --default-toolchain ${RUST_VERSION} --target wasm32-wasip1 --no-update-default-toolchain --no-modify-path && \
+	/root/.cargo/bin/rustup component remove clippy && \
+	/root/.cargo/bin/rustup component remove rust-docs && \
+	/root/.cargo/bin/rustup component remove rustfmt && \
+	rm rustup-init && \
+	ln -s /root/.cargo/bin/cargo /usr/local/bin/cargo
 
 COPY --from=builder /usr/include/bpf /usr/include/bpf
 COPY --from=builder /usr/local/bin/bpftool /usr/local/bin/bpftool
--- a/cmd/common/image/build.go
+++ b/cmd/common/image/build.go
@@ -456,12 +456,13 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
 
 	cmd := buildCmd("/out", ebpfFullPath, wasmFullPath, conf.CFlags, "/btfhub-archive", opts.btfgen)
 
+	// The work mount ReadOnly field is updated as false, to allow Cargo.lock to compiled in /work folder for rust source code.
 	mounts := []mount.Mount{
 		{
 			Type:     mount.TypeBind,
 			Target:   "/work",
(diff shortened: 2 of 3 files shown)


Dockerfiles/gadget-builder.Dockerfile
```diff
@@ -77,6 +88,7 @@ ENV GOTOOLCHAIN=local
 # Create a directory which can be read, written and executed by everyone, this
 # avoid trouble when running as non root.
 RUN mkdir -m 777 /work
+
```
eiffel-fl: Remove this empty line.

Dockerfiles/gadget-builder.Dockerfile
```diff
@@ -62,6 +64,15 @@ RUN ARCH=$(dpkg --print-architecture) \
 
 COPY --from=builder /usr/include/bpf /usr/include/bpf
 COPY --from=builder /usr/local/bin/bpftool /usr/local/bin/bpftool
+COPY --from=rustbuilder /usr/local/cargo /usr/local/.cargo
+COPY --from=rustbuilder /usr/local/rustup /usr/local/.rustup
+
+# Add Rust to PATH
+ENV PATH="/usr/local/.cargo/bin:/usr/local/.rustup:$PATH"
```
eiffel-fl:
```suggestion
COPY --from=rustbuilder /usr/local/cargo /usr/local/bin/cargo
COPY --from=rustbuilder /usr/local/rustup /usr/local/bin/rustup
```

cmd/common/image/build.go
```diff
@@ -461,7 +461,7 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
 			Type:     mount.TypeBind,
 			Target:   "/work",
 			Source:   pathHost,
-			ReadOnly: true,
+			ReadOnly: false,
```
eiffel-fl, line 465: Port the comment indicating why we need this...

cmd/common/image/helpers/Makefile.build
```diff
@@ -38,6 +38,13 @@ wasm: $(WASM)
 	# -buildmode=c-shared to build the wasm as a reactor module. See https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md#current-unstable-abi
 	cd $(dir $(WASM)) && \
 	CGO_ENABLED=0 GOOS=wasip1 GOARCH=wasm go build -o $(OUTPUTDIR)/program.wasm -buildmode=c-shared -ldflags "-w -s" $(notdir $(WASM))
+else ifeq (rs, $(patsubst %.rs,rs,$(WASM)))
+wasm: $(WASM)
+	cd $(dir $(WASM)) && \
+	cargo build --target wasm32-wasip1 --release && \
```
eiffel-fl: Use [CARGO_TARGET_DIR](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads) to specify the directory, this way you may not need the previous commit:

```suggestion
	CARGO_TARGET_DIR=$(OUTPUTDIR) cargo build --target wasm32-wasip1 --release && \
```

Dockerfiles/gadget-builder.Dockerfile
Daksh-10: If I remove the path env variable then I need to provide rustup binary absolute path.

cmd/common/image/build.go
Daksh-10: Done!

cmd/common/image/helpers/Makefile.build
Daksh-10: Cargo doesn't allow to create specific artifacts. So, the CARGO_TARGET_DIR copies full ./target folder to output dir(/out). And we only need program.wasm file inside output folder otherwise, we need to change the logic a bit.

1. Create a new mount 'build' with write permissions.
2. Copy the final program.wasm file to /out mount for further processing.

I do think that making '/work' mount as writable is not a problem. Do you think there is an issue with it?

Dockerfiles/gadget-builder.Dockerfile
Daksh-10: Removing the PATH variable made it messy (it is still manageable). Do you think is there a better approach to it?

cmd/common/image/helpers/Makefile.build
eiffel-fl: Let's keep it like this for now, but open an issue, once merged, to make /work read-only again.

Dockerfiles/gadget-builder.Dockerfile
eiffel-fl: Let's keep it like this for now.

cmd/common/image/build.go
mauriciovasquezbernal: I see these changes disappeared from the PR. Is it intended? @Daksh-10

Daksh-10: Yes, the compilation now is done in "/out" folder, so readonly can be maintained on "/work"

mauriciovasquezbernal: mmm, I got confused by your comment on https://github.com/inspektor-gadget/inspektor-gadget/pull/4562#discussion_r2130530421 as it states it wasn't possible.

Daksh-10: Yes, that was wrong. It is now updated.

cmd/common/image/build.go
```diff
@@ -456,12 +456,13 @@ func buildInContainer(opts *cmdOpts, conf *buildFile) error {
...
 	cmd := buildCmd("/out", ebpfFullPath, wasmFullPath, conf.CFlags, "/btfhub-archive", opts.btfgen)
 
+	// The work mount ReadOnly field is updated as false, to allow Cargo.lock to compiled in /work folder for rust source code.
 	mounts := []mount.Mount{
 		{
 			Type:     mount.TypeBind,
 			Target:   "/work",
 			Source:   pathHost,
-			ReadOnly: true,
+			ReadOnly: false,
```
eiffel-fl, line 465: Why did you add back this commit?
This is no more needed now what we have all the cargo build artifacts stored in OUTPUT_DIR.

Daksh-10: It is needed, if we don't have Cargo.lock file then we can't build .wasm artifacts. Try to delete the Cargo.lock file then try to build the image.

eiffel-fl: Indeed, I used the following commit and it has Cargo.lock committed, hence the fact it worked.
Let's keep it as is for now, but open an issue, so we can investigate how to remove this workaround.

Daksh-10: Actually, I suggested a way, where we can copy the whole source code to /out directory and then we can cargo built there and it won't be an issue. But, Mauricio said it won't be an issue with current approach.

mauriciovasquezbernal: Let's discuss it further on an issue, but I don't see any problem with having `ReadOnly: false`

mauriciovasquezbernal (approved):
LGTM as well. Thanks!

PR #4565, opened 2025-06-05
trace_tcp: add accept-only parameter

alban:
The trace_tcp already had a connect-only parameter to get only connect events. This patch adds the accept-only parameter, working in the same way, to get only accept events.

## How to use

```
sudo ig run ghcr.io/inspektor-gadget/gadget/trace_tcp:alban_tcp_accept_only --verify-image=false --host --accept-only
```

## Testing done

I tested the above command.

@johananl


gadgets/trace_tcp/README.mdx
```diff
@@ -29,23 +29,23 @@ Running the gadget:
 </Tabs>
 ## Flags
 
-### `--pid`
+### `--accept-only`
```
blanquicet, line 32: I suggest mentioning that it's recommended to use `--accept-only` and `--connect-only` rather than `--filter='type==connect'` and `--filter='type==accept'` because the former is done in eBPF directly while for the latter, the events will be sent to user space and then discarded there (Most resource consuming).

Issue #4564, opened 2025-06-05
[RFE] Add tcpdump / generic packet capture gadget


flyth:
## Current situation

We currently don't support capturing and exporting packets in a way that can be easily consumed by third party tools like tcpdump or Wireshark. We also don't support libpcap-style filtering in eBPF programs.

## Ideal future situation

People will be able to use IG similar to `tcpdump` and capture traffic filtered by both libpcap-style filters as well as our own container-aware filters.

## Implementation options

libpcap can be used to compile filters to cbpf, which then can be converted and injected into eBPF. We should keep it as generic as possible, so that filters can not only be used to capture packets, but also for other filter-based operations (e.g. filter traffic).
Additionally, IG should get support for the pcap-ng file format so that the collected traffic can be viewed in for example Wireshark. Enrichments (context information) can be added to both the interface information present in pcap-ng as well as comments on each packet.

## Additional information

A (later) optimization would be to create an asynchronous caching mode for our enriching operators:
Instead of enriching every packet with the same information, only "new" information should be posted to a dedicated new datasource, whose data can be used to build lookup tables on the client (or in a pcap-ng file).

mauriciovasquezbernal:
Fixed by #4564. @flyth feel free to reopen if I am wrong.

PR #4567, opened 2025-06-05
introduce generic packet filters and new gadget tcpdump

flyth:
This adds support for generic packet filters compiled using libpcap (needs to present on the system).
Also introduces a new gadget `tcpdump` that allows capturing (filtered) packets on the system that can be written to pcap files.

```
$ cd gadgets/tcpdump
$ ig image build -t tcpdump:mine .
$ ig run --verify-image=false tcpdump:mine --pf "port 80" -o pcap-ng | tcpdump -nvr -
reading from file -, link-type EN10MB (Ethernet), snapshot length 65535
2025/06/06 13:00:45 failed to upload metrics: context deadline exceeded: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 172.16.37.20:4317: connect: connection refused"
13:00:58.045870 IP (tos 0x0, ttl 64, id 9615, offset 0, flags [DF], proto TCP (6), length 60)
    172.17.0.2.57098 > 13.107.253.44.80: Flags [S], cksum 0xb6d9 (incorrect -> 0x52b2), seq 2293594051, win 64240, options [mss 1460,sackOK,TS val 3150948366 ecr 0,nop,wscale 7], length 0
13:00:58.066287 IP (tos 0x0, ttl 127, id 57611, offset 0, flags [none], proto TCP (6), length 48)
    13.107.253.44.80 > 172.17.0.2.57098: Flags [S.], cksum 0x1332 (correct), seq 273959432, ack 2293594052, win 32768, options [mss 1460,wscale 1,nop], length 0
13:00:58.066333 IP (tos 0x0, ttl 64, id 9616, offset 0, flags [DF], proto TCP (6), length 40)
    172.17.0.2.57098 > 13.107.253.44.80: Flags [.], cksum 0xb6c5 (incorrect -> 0xbd01), ack 1, win 502, length 0
13:00:58.066337 IP (tos 0x0, ttl 64, id 9617, offset 0, flags [DF], proto TCP (6), length 116)
    172.17.0.2.57098 > 13.107.253.44.80: Flags [P.], cksum 0xb711 (incorrect -> 0xbcad), seq 1:77, ack 1, win 502, length 76: HTTP
13:00:58.066396 IP (tos 0x0, ttl 127, id 57612, offset 0, flags [none], proto TCP (6), length 40)
    13.107.253.44.80 > 172.17.0.2.57098: Flags [.], cksum 0x7eab (correct), ack 77, win 16384, length 0
13:00:58.091520 IP (tos 0x0, ttl 127, id 57613, offset 0, flags [none], proto TCP (6), length 319)
    13.107.253.44.80 > 172.17.0.2.57098: Flags [P.], cksum 0xa8d7 (correct), seq 1:280, ack 77, win 16384, length 279: HTTP, length: 279
	HTTP/1.1 307 Temporary Redirect
	Date: Fri, 06 Jun 2025 11:00:58 GMT
	Content-Type: text/html
	Content-Length: 0
	Connection: keep-alive
	Location: https://microsoft.com/
	x-azure-ref: 20250606T110058Z-174fcd5bb4dl6jrhhC1FRAh8s0000000085g00000000k4w6
	X-Cache: CONFIG_NOCACHE

13:00:58.091541 IP (tos 0x0, ttl 64, id 9618, offset 0, flags [DF], proto TCP (6), length 40)
    172.17.0.2.57098 > 13.107.253.44.80: Flags [.], cksum 0xb6c5 (incorrect -> 0xbb9f), ack 280, win 501, length 0
13:00:58.091639 IP (tos 0x0, ttl 64, id 9619, offset 0, flags [DF], proto TCP (6), length 40)
    172.17.0.2.57098 > 13.107.253.44.80: Flags [F.], cksum 0xb6c5 (incorrect -> 0xbb9e), seq 77, ack 280, win 501, length 0
13:00:58.091974 IP (tos 0x0, ttl 127, id 57614, offset 0, flags [none], proto TCP (6), length 40)
    13.107.253.44.80 > 172.17.0.2.57098: Flags [.], cksum 0x7d93 (correct), ack 78, win 16384, length 0
13:00:58.114661 IP (tos 0x0, ttl 127, id 57615, offset 0, flags [none], proto TCP (6), length 40)
    13.107.253.44.80 > 172.17.0.2.57098: Flags [F.], cksum 0x7d92 (correct), seq 280, ack 78, win 16384, length 0
13:00:58.114679 IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 40)
    172.17.0.2.57098 > 13.107.253.44.80: Flags [.], cksum 0xbb9d (correct), ack 281, win 501, length 0
^Ctcpdump: pcap_loop: error reading dump file: Interrupted system call
```

## Remarks

* Metadata currently only can be added to interfaces (using annotations, see tcpdump example)
* Metadata transport is currently very inefficient, as we're sending all enrichments with every packet

## Todos

* [x] Tests
* [ ] Provide WireShark plugin to extract metadata ([WIP](https://gist.github.com/flyth/82d4993cbd2a306d7dfff59a4e5f1603))

Implements #4564

### DIFF
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/cilium/ebpf v0.19.1-0.20250729164112-d994daa25101 // indirect
+	github.com/cloudflare/cbpfc v0.0.0-20240920015331-ff978e94500b // indirect
 	github.com/containerd/cgroups/v3 v3.0.5 // indirect
 	github.com/containerd/containerd v1.7.28 // indirect
 	github.com/containerd/containerd/api v1.8.0 // indirect
@@ -65,6 +66,7 @@ require (
 	github.com/google/go-cmp v0.7.0 // indirect
 	github.com/google/go-containerregistry v0.20.3 // indirect
 	github.com/google/uuid v1.6.0 // indirect
+	github.com/gopacket/gopacket v1.4.0 // indirect
 	github.com/grafana/regexp v0.0.0-20240518133315-a468a5bfb3bc // indirect
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.2 // indirect
@@ -95,6 +97,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.1.1 // indirect
 	github.com/opencontainers/selinux v1.12.0 // indirect
+	github.com/packetcap/go-pcap v0.0.0-20250723190045-d00b185f30b7 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.4 // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pkg/errors v0.9.1 // indirect
--- a/gadgets/tcpdump/README.mdx
+++ b/gadgets/tcpdump/README.mdx
@@ -1,0 +1,80 @@
+---
+title: tcpdump
+sidebar_position: 0
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# tcpdump
+
+The tcpdump gadget captures packets in container contexts and allows applying pcap-compatible filters. This is usually
+combined with the `pcap-ng` output mode and piped to the `tcpdump` command or to a file - see the guide below.
+
+## Getting started
+
+Running the gadget:
+
+<Tabs groupId="env">
+    <TabItem value="kubectl-gadget" label="kubectl gadget">
+        ```bash
+        $ kubectl gadget run ghcr.io/inspektor-gadget/gadget/tcpdump:%IG_TAG% [flags]
+        ```
+    </TabItem>
+
+    <TabItem value="ig" label="ig">
+        ```bash
+        $ sudo ig run ghcr.io/inspektor-gadget/gadget/tcpdump:%IG_TAG% [flags]
+        ```
+    </TabItem>
+</Tabs>
+
+## Flags
+
+### `--snaplen`
+
+Sets the maximum number of bytes to capture from a packet.
+
+Default value: 0
+
+## Guide
+
+### Piping to tcpdump
+
+If you want to let tcpdump analyze the captured traffic directly, you can use the `pcap-ng` output mode and pipe IG's
+output directly to tcpdump like so:
+
+```bash
+$ ig run tcpdump:%IG_TAG% --host --pf "port 80" -o pcap-ng | tcpdump -nvr -
+14:18:23.386163 IP (tos 0x0, ttl 64, id 59190, offset 0, flags [DF], proto TCP (6), length 60)
+    172.17.0.2.53618 > 13.107.253.67.80: Flags [S], cksum 0xb6f0 (incorrect -> 0x68f7), seq 1211089028, win 64240, options [mss 1460,sackOK,TS val 3404957675 ecr 0,nop,wscale 7], length 0
+14:18:23.405108 IP (tos 0x0, ttl 127, id 3783, offset 0, flags [none], proto TCP (6), length 48)
+    13.107.253.67.80 > 172.17.0.2.53618: Flags [S.], cksum 0x988a (correct), seq 2080464456, ack 1211089029, win 32768, options [mss 1460,wscale 1,nop], length 0
+14:18:23.405212 IP (tos 0x0, ttl 64, id 59191, offset 0, flags [DF], proto TCP (6), length 40)
+    172.17.0.2.53618 > 13.107.253.67.80: Flags [.], cksum 0xb6dc (incorrect -> 0x425a), ack 1, win 502, length 0
+14:18:23.405869 IP (tos 0x0, ttl 64, id 59192, offset 0, flags [DF], proto TCP (6), length 116)
+    172.17.0.2.53618 > 13.107.253.67.80: Flags [P.], cksum 0xb728 (incorrect -> 0x4206), seq 1:77, ack 1, win 502, length 76: HTTP
+14:18:23.406531 IP (tos 0x0, ttl 127, id 3784, offset 0, flags [none], proto TCP (6), length 40)
+    13.107.253.67.80 > 172.17.0.2.53618: Flags [.], cksum 0x0404 (correct), ack 77, win 16384, length 0
+14:18:23.433402 IP (tos 0x0, ttl 127, id 3785, offset 0, flags [none], proto TCP (6), length 319)
+    13.107.253.67.80 > 172.17.0.2.53618: Flags [P.], cksum 0x98ef (correct), seq 1:280, ack 77, win 16384, length 279: HTTP, length: 279
+	HTTP/1.1 307 Temporary Redirect
+	Date: Wed, 30 Jul 2025 12:18:23 GMT
+	Content-Type: text/html
+	Content-Length: 0
+	Connection: keep-alive
+	Location: https://microsoft.com/
+	X-Cache: CONFIG_NOCACHE
+
+14:18:23.433435 IP (tos 0x0, ttl 64, id 59193, offset 0, flags [DF], proto TCP (6), length 40)
+    172.17.0.2.53618 > 13.107.253.67.80: Flags [.], cksum 0xb6dc (incorrect -> 0x40f8), ack 280, win 501, length 0
+14:18:23.434372 IP (tos 0x0, ttl 64, id 59194, offset 0, flags [DF], proto TCP (6), length 40)
+    172.17.0.2.53618 > 13.107.253.67.80: Flags [F.], cksum 0xb6dc (incorrect -> 0x40f7), seq 77, ack 280, win 501, length 0
+14:18:23.436706 IP (tos 0x0, ttl 127, id 3786, offset 0, flags [none], proto TCP (6), length 40)
+    13.107.253.67.80 > 172.17.0.2.53618: Flags [.], cksum 0x02ec (correct), ack 78, win 16384, length 0
+14:18:23.466609 IP (tos 0x0, ttl 127, id 3787, offset 0, flags [none], proto TCP (6), length 40)
+    13.107.253.67.80 > 172.17.0.2.53618: Flags [F.], cksum 0x02eb (correct), seq 280, ack 78, win 16384, length 0
+14:18:23.466629 IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 40)
+    172.17.0.2.53618 > 13.107.253.67.80: Flags [.], cksum 0x40f6 (correct), ack 281, win 501, length 0
+```
+
--- a/gadgets/tcpdump/gadget.yaml
+++ b/gadgets/tcpdump/gadget.yaml
@@ -1,0 +1,44 @@
+name: tcpdump
+description: Capture network packets
+homepageURL: https://inspektor-gadget.io/
+documentationURL: https://www.inspektor-gadget.io/docs/latest/gadgets/tcpdump
+sourceURL: https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/tcpdump
+params:
+  ebpf:
+    snaplen:
+      key: snaplen
+      description: 'Maximum number of bytes to capture from a packet.'
+datasources:
+  packets:
+    annotations:
+      ebpf.rest.len: payload_len
+      ebpf.rest.name: payload
+      cli.supported-output-modes: pcap-ng,columns,json,yaml
+      cli.default-output-mode: columns
+      cli.raw.fields: payload
+      pcap.payload: payload
+      pcap.timestamp: timestamp
+      pcap.packet-length: payload_len
+    fields:
+      ifindex:
+        annotations:
+          pcap.interface-key: true
+      payload:
+        annotations:
+          content-type: application/x-raw-packet
+      k8s.node:
+        annotations:
+          pcap.interface-var: k8s.node
+          pcap.interface-key: true
+      k8s.containerName:
+        annotations:
+          pcap.interface-var: k8s.containerName
+      k8s.podName:
+        annotations:
+          pcap.interface-var: k8s.podName
+      k8s.podLabels:
+        annotations:
+          pcap.interface-var: k8s.podLabels
+      runtime.containerName:
+        annotations:
+          pcap.interface-var: runtime.containerName
--- a/gadgets/tcpdump/program.bpf.c
+++ b/gadgets/tcpdump/program.bpf.c
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2025 The Inspektor Gadget authors */
+
+#include <vmlinux.h>
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_endian.h>
+#include <bpf/bpf_core_read.h>
+
+#include <gadget/macros.h>
+#include <gadget/types.h>
+#include <gadget/buffer.h>
+#include <gadget/filter.h>
+#include <gadget/packetfilter.h>
+
+#define GADGET_TYPE_NETWORKING
+#include <gadget/sockets-map.h>
+
+#define MAX_PKT_LEN 1500
+
+#define PACKET_TYPE_EGRESS 0
+#define PACKET_TYPE_INGRESS 1
+
+struct packet_event_t {
+	gadget_timestamp timestamp_raw;
+	u8 packet_type;
+	u32 ifindex;
+	u32 payload_len;
+	u32 packet_size;
+	struct gadget_process proc;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} packets SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(struct packet_event_t));
+} tmp_packets SEC(".maps");
+
+GADGET_TRACER(packets, packets, packet_event_t);
+
+const volatile __u16 snaplen = MAX_PKT_LEN;
+GADGET_PARAM(snaplen);
+
+GADGET_PF();
+
+static __always_inline int handle(struct __sk_buff *skb, __u8 packet_type)
+{
+	if (!gadget_pf_matches((void *)skb, (void *)(long)skb->data,
+			       (void *)(long)skb->data_end))
+		return 0;
+
+	struct gadget_socket_value *skb_val = gadget_socket_lookup(skb);
+	if (gadget_should_discard_data_by_skb(skb_val))
+		return 0;
+
+	struct packet_event_t *event;
+
+	int zero = 0;
+	event = bpf_map_lookup_elem(&tmp_packets, &zero);
+	if (!event)
+		return 0; // it never happens
+
+	__builtin_memset(event, 0, sizeof(*event));
+
+	__u64 len = skb->len;
+	event->packet_size = len;
+
+	if (len > snaplen)
+		len = snaplen;
+
+	event->payload_len = len;
+	event->timestamp_raw = bpf_ktime_get_ns();
+	event->ifindex = skb->ifindex;
+	event->packet_type = packet_type;
+
+	// Enrich event with process metadata
+	gadget_process_populate_from_socket(skb_val, &event->proc);
+
+	bpf_perf_event_output(skb, &packets, len << 32 | BPF_F_CURRENT_CPU,
+			      event, sizeof(*event));
+	return 0;
+}
+
+SEC("classifier/ingress/main")
+int ingress_main(struct __sk_buff *skb)
+{
+	bpf_skb_pull_data(skb, 0);
+	return handle(skb, PACKET_TYPE_INGRESS);
+}
+
+SEC("classifier/egress/main")
+int egress_main(struct __sk_buff *skb)
+{
+	bpf_skb_pull_data(skb, 0);
+	return handle(skb, PACKET_TYPE_EGRESS);
+}
+
+char _license[] SEC("license") = "GPL";
--- a/go.mod
+++ b/go.mod
@@ -27,13 +27,13 @@ require (
 	github.com/gofrs/flock v0.12.1
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
+	github.com/gopacket/gopacket v1.4.0
 	github.com/kr/pretty v0.3.1
 	github.com/moby/moby v28.4.0+incompatible
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/opencontainers/runtime-spec v1.2.1
 	github.com/packetcap/go-pcap v0.0.0-20250723190045-d00b185f30b7
 	github.com/prometheus/client_golang v1.23.2
-	github.com/prometheus/client_golang v1.23.0
 	github.com/s3rj1k/go-fanotify/fanotify v0.0.0-20210917134616-9c00a300bb7a
 	github.com/seccomp/libseccomp-golang v0.11.0 // indirect
 	github.com/sigstore/sigstore v1.9.5
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.24.0
 require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/cilium/ebpf v0.19.1-0.20250729164112-d994daa25101
+	github.com/cloudflare/cbpfc v0.0.0-20240920015331-ff978e94500b
 	github.com/containerd/containerd v1.7.28
 	github.com/containerd/errdefs v1.0.0
 	github.com/containerd/nri v0.9.0
@@ -30,7 +31,9 @@ require (
 	github.com/moby/moby v28.4.0+incompatible
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/opencontainers/runtime-spec v1.2.1
+	github.com/packetcap/go-pcap v0.0.0-20250723190045-d00b185f30b7
 	github.com/prometheus/client_golang v1.23.2
+	github.com/prometheus/client_golang v1.23.0
 	github.com/s3rj1k/go-fanotify/fanotify v0.0.0-20210917134616-9c00a300bb7a
 	github.com/seccomp/libseccomp-golang v0.11.0 // indirect
 	github.com/sigstore/sigstore v1.9.5
@@ -54,7 +57,7 @@ require (
 	go.opentelemetry.io/otel/sdk/log v0.14.0
 	go.opentelemetry.io/otel/sdk/metric v1.38.0
 	golang.org/x/exp v0.0.0-20250408133849-7e4ce0ab07d0
-	golang.org/x/net v0.44.0 // indirect
+	golang.org/x/net v0.44.0
 	golang.org/x/sync v0.17.0 // indirect
 	golang.org/x/sys v0.36.0
 	golang.org/x/term v0.35.0
--- a/pkg/libpcap-compiler/ebpf.go
+++ b/pkg/libpcap-compiler/ebpf.go
@@ -1,0 +1,77 @@
+// Copyright 2025 The Inspektor Gadget authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package libpcap_compiler compiles libpcap filter expressions into eBPF
+// code. This package contains parts and/or was inspired from:
+// https://github.com/jschwinger233/elibpcap/
+// https://github.com/mozillazg/ptcpdump/
+package libpcap_compiler
+
+import (
+	"github.com/cilium/ebpf/asm"
+	"github.com/cloudflare/cbpfc"
+)
+
+const (
+	RejectAllExpr = "__reject_all__"
+)
+
+type StackOffset int
+
+const (
+	BpfReadKernelOffset StackOffset = -8 * (iota + 1)
+	R1Offset
+	R2Offset
+	R3Offset
+	R4Offset
+	R5Offset
+	AvailableOffset
+)
+
+// CompileEbpf compiles a libpcap filter expression to eBPF instructions
+func CompileEbpf(expr string, fname string) (insts asm.Instructions, err error) {
+	if expr == RejectAllExpr {
+		return asm.Instructions{
+			asm.Mov.Reg(asm.R4, asm.R5), // r4 = r5 (data = data_end)
+		}, nil
+	}
+	cbpfInsts, err := CompileCbpf(expr)
+	if err != nil {
+		return
+	}
+
+	ebpfInsts, err := cbpfc.ToEBPF(cbpfInsts, cbpfc.EBPFOpts{
+		// skb->data is at r4, skb->data_end is at r5.
+		PacketStart: asm.R4,
+		PacketEnd:   asm.R5,
+		Result:      asm.R0,
+		ResultLabel: "_result_" + fname,
+		// result is at R0, _skb is at R1, __skb is at R2, ___skb is at R3.
+		// these are reserved and will be used internally
+		Working:     [4]asm.Register{asm.R0, asm.R1, asm.R2, asm.R3},
+		LabelPrefix: "_prefix_" + fname,
+		StackOffset: -int(AvailableOffset),
+	})
+	if err != nil {
+		return
+	}
+
+	return append(ebpfInsts,
+		asm.Mov.Imm(asm.R1, 0).WithSymbol("_result_"+fname), // r1 = 0 (_skb)
+		asm.Mov.Imm(asm.R2, 0),                              // r2 = 0 (__skb)
+		asm.Mov.Imm(asm.R3, 0),                              // r3 = 0 (___skb)
+		asm.Mov.Reg(asm.R4, asm.R0),                         // r4 = $result (data)
+		asm.Mov.Imm(asm.R5, 0),                              // r5 = 0 (data_end)
+	), nil
+}
--- a/pkg/operators/cli/clioperator.go
+++ b/pkg/operators/cli/clioperator.go
@@ -17,12 +17,18 @@ package clioperator
 import (
 	"fmt"
 	"io"
+	"net/url"
 	"os"
+	"runtime"
 	"slices"
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
+	"github.com/gopacket/gopacket"
+	"github.com/gopacket/gopacket/layers"
+	"github.com/gopacket/gopacket/pcapgo"
 	"golang.org/x/term"
 	"sigs.k8s.io/yaml"
 
@@ -49,6 +55,7 @@ const (
 	ModeYAML       = "yaml"
 	ModeNone       = "none"
 	ModeRaw        = "raw"
+	ModePCAPNG     = "pcap-ng"
 
 	DefaultOutputMode = ModeColumns
 
@@ -62,6 +69,23 @@ const (
 
 	// AnnotationDefaultOutputMode can be used to specify the default output mode for a DataSource.
 	AnnotationDefaultOutputMode = "cli.default-output-mode"
+
+	// AnnotationPCAPPayload must point to the field carrying the binary payload of a packet
+	AnnotationPCAPPayload = "pcap.payload"
+
+	// AnnotationPCAPTimestamp must point to the field containing the capture timestamp of a packet
+	AnnotationPCAPTimestamp = "pcap.timestamp"
+
+	// AnnotationPCAPPacketLen must point to the field containing the original length of the packet
+	AnnotationPCAPPacketLen = "pcap.packet-length"
+
+	// AnnotationPCAPInterfaceKey (if set to true) defines for a field to become part of the interface key
+	// (a composite of host info and interface info that makes it unique)
+	AnnotationPCAPInterfaceKey = "pcap.interface-key"
+
+	// AnnotationPCAPInterfaceVar registers a field to become part of the interface comment variables
+	// with the given name
+	AnnotationPCAPInterfaceVar = "pcap.interface-var"
 )
 
 var (
@@ -460,6 +484,128 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error 
 					return nil
 				}, Priority)
 			}
+		case ModePCAPNG:
+			// Check ds for compatiblity
+			payloadField := ds.GetField(ds.Annotations()[AnnotationPCAPPayload])
+			if payloadField == nil {
+				return fmt.Errorf("%s annotation not found", AnnotationPCAPPayload)
+			}
+			timestampField := ds.GetField(ds.Annotations()[AnnotationPCAPTimestamp])
+			if timestampField == nil {
+				return fmt.Errorf("timestamp field not found")
+			}
+			lengthField := ds.GetField(ds.Annotations()[AnnotationPCAPPacketLen])
+			if lengthField == nil {
+				return fmt.Errorf("length field not found")
+			}
+
+			// TODO:
+			// Currently, each data packet carries all interface/enriched information with the payload;
+			// optimally, we would create a special datasource for it and cache the sent information both
+			// on sender and receiver to save bandwidth.
+
+			// Find fields to be used for the interface key
+			var interfaceKeyFields []datasource.FieldAccessor
+			for _, f := range ds.Fields() {
+				if v, ok := f.Annotations[AnnotationPCAPInterfaceKey]; ok && v == "true" {
+					acc := ds.GetField(f.FullName)
+					if acc == nil {
+						return fmt.Errorf("field %q not found", f.FullName)
+					}
+					interfaceKeyFields = append(interfaceKeyFields, acc)
+				}
+			}
+
+			// Find fields to be used for the interface variables; these variables will become
+			// part of the interface description and can be decoded using scripts in WireShark, for
+			// example. TODO: also add this functionality on a packet level.
+			var interfaceVarFns []func(datasource.Data, *strings.Builder)
+			for _, f := range ds.Fields() {
+				if v, ok := f.Annotations[AnnotationPCAPInterfaceVar]; ok {
+					acc := ds.GetField(f.FullName)
+					if acc == nil {
+						return fmt.Errorf("field %q not found", f.FullName)
+					}
+					if acc.Type() != api.Kind_String && acc.Type() != api.Kind_CString {
+						return fmt.Errorf("field %q cannot be used as interface variable; currently, only string fields are supported", f.FullName)
+					}
+					interfaceVarFns = append(interfaceVarFns, func(data datasource.Data, sb *strings.Builder) {
+						str, err := acc.String(data)
+						if err != nil {
+							return
+						}
+						sb.WriteString(v)
+						sb.WriteByte('=')
+						sb.WriteString(url.QueryEscape(str))
+					})
+					interfaceKeyFields = append(interfaceKeyFields, acc)
+				}
+			}
+
+			var mu sync.Mutex
+
+			wr, err := pcapgo.NewNgWriter(os.Stdout, layers.LinkTypeEthernet)
+			if err != nil {
+				return fmt.Errorf("creating pcapgo.NgWriter: %w", err)
+			}
+
+			interfaces := make(map[string]int)
+
+			ds.Subscribe(func(ds datasource.DataSource, data datasource.Data) error {
+				mu.Lock()
+				defer mu.Unlock()
+
+				var key strings.Builder
+				for _, f := range interfaceKeyFields {
+					key.Write(f.Get(data))
+					key.WriteByte(':') // Delimiter
+				}
+				ifIndex := key.String()
+
+				ts, _ := timestampField.Uint64(data)
+				payload, _ := payloadField.Bytes(data)
+				length, _ := lengthField.Uint32(data)
+
+				pcapInterface, ok := interfaces[ifIndex]
+				if !ok {
+					// Build description for interface
+					var description strings.Builder
+					for i, f := range interfaceVarFns {
+						if i > 0 {
+							description.WriteByte(';')
+						}
+						f(data, &description)
+					}
+
+					newInterface := pcapgo.NgInterface{
+						Name:        "virtual", // TODO: get from host
+						Description: description.String(),
+						OS:          runtime.GOOS, // TODO: get from host
+						LinkType:    layers.LinkTypeEthernet,
+					}
+					newIndex, err := wr.AddInterface(newInterface)
+					if err != nil {
+						return err
+					}
+					interfaces[ifIndex] = newIndex
+					pcapInterface = newIndex
+				}
+
+				err = wr.WritePacket(gopacket.CaptureInfo{
+					Timestamp:      time.Unix(0, int64(time.Duration(ts)*time.Microsecond)),
+					InterfaceIndex: pcapInterface,
+					CaptureLength:  len(payload),
+					Length:         int(length),
+				}, payload)
+				if err != nil {
+					gadgetCtx.Logger().Warnf("failed to write packet: %v", err)
+					return nil
+				}
+
+				// need to make sure this gets written before returning buffers
+				wr.Flush()
+				return nil
+			}, Priority)
 		}
 
 	}
--- a/pkg/operators/ebpf/ebpf.go
+++ b/pkg/operators/ebpf/ebpf.go
@@ -30,6 +30,7 @@ import (
 	"sync"
 
 	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/ebpf/btf"
 	"github.com/cilium/ebpf/link"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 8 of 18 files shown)


gadgets/tcpdump/gadget.yaml
```diff
@@ -0,0 +1,22 @@
...
+      ebpf.rest.len: payload_len
+      ebpf.rest.name: payload
+      cli.supported-output-modes: pcap-ng,columns,json,yaml
+      cli.default-output-mode: columns
+      cli.raw.fields: payload
+      cli.pcap-ng.payload: payload
+      cli.pcap-ng.interface: ifindex
+      cli.pcap-ng.timestamp: timestamp
+      cli.pcap-ng.capture-length: payload_len
+      cli.pcap-ng.length: payload_len
```
mauriciovasquezbernal: What's better? To add annotations here indicating what are the fields, or to annotate the fields themselves?

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,110 @@
...
+GADGET_TRACER(packets, packets, packet_event_t);
+
+const volatile __u16 snaplen = MAX_PKT_LEN;
+GADGET_PARAM(snaplen);
+
+static __noinline bool gadget_pf_main(void *_skb, void *__skb, void *___skb,
+				      void *data, void *data_end)
+{
+	return data != data_end && _skb == __skb && __skb == ___skb;
+}
```
mauriciovasquezbernal: Perhaps it can be defined as extern to avoid having to implement it here.

gadgets/tcpdump/gadget.yaml
flyth: Yeah, thought about the same. However, this way it's a 1:1 relation - if you annotate the fields wrongly (multiple fields with the same annotation), it could create ambiguities - this way it can't.

gadgets/tcpdump/program.bpf.c
flyth: That doesn't work, at least not directly. I'll experiment a bit with it to see if there's any other way.

pkg/libpcap-compiler/ebpf.go
```diff
@@ -0,0 +1,78 @@
...
+		StackOffset: -int(AvailableOffset),
+	})
+	if err != nil {
+		return
+	}
+
+	return append(ebpfInsts,
+		asm.Mov.Imm(asm.R1, 0).WithSymbol("_result_"+fname), // r1 = 0 (_skb)
+		asm.Mov.Imm(asm.R2, 0),                              // r2 = 0 (__skb)
+		asm.Mov.Imm(asm.R3, 0),                              // r3 = 0 (___skb)
```
mauriciovasquezbernal, line 73: What's the meaning of the `_skb`, `__skb`,... comments?

examples/go.mod
mauriciovasquezbernal, line 1: Please update the commit message.

go.mod
```diff
@@ -76,6 +76,11 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require (
+	github.com/cloudflare/cbpfc v0.0.0-20240920015331-ff978e94500b
+	github.com/packetcap/go-pcap v0.0.0-20250723190045-d00b185f30b7
+)
```
mauriciovasquezbernal: nit: move it to the existing require block.

pkg/operators/cli/clioperator.go
```diff
@@ -459,6 +462,55 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+			}
+			lengthField := ds.GetField(ds.Annotations()["cli.pcap-ng.length"])
+			if lengthField == nil {
+				return fmt.Errorf("length field not found")
+			}
+
+			var mu sync.Mutex
+
+			// Output using pcap-ng
+			// wr, err := pcapgo.NewNgWriter(os.Stdout, layers.LinkTypeEthernet)
```
mauriciovasquezbernal: nit: remove?

pkg/operators/cli/clioperator.go
```diff
@@ -459,6 +462,55 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+			wr, err := pcapng.New(os.Stdout)
+			if err != nil {
+				return fmt.Errorf("creating pcapgo.NgWriter: %w", err)
+			}
+			ds.Subscribe(func(ds datasource.DataSource, data datasource.Data) error {
+				mu.Lock()
+				defer mu.Unlock()
+
+				// ifIndex, _ := interfaceField.Uint32(data)
+				// ts, _ := timestampField.Uint64(data)
```
mauriciovasquezbernal: Remove? Or use below?

pkg/operators/cli/clioperator.go
```diff
@@ -459,6 +462,55 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+				payload, _ := payloadField.Bytes(data)
+				length, _ := lengthField.Uint32(data)
+
+				wr.WritePacket(payload, &pcapng.PacketOptions{
+					Timestamp:      time.Now(),
+					InterfaceIndex: 0,
+					CaptureLength:  int(length),
+					Length:         int(length),
+					AncillaryData:  nil,
+					Comment:        "fooby",
```
mauriciovasquezbernal: :eyes:

gadgets/tcpdump/README.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+Sets the maximum number of bytes to capture from a packet.
+
+Default value: 0
+
+## Guide
+
+### Piping to tcpdump
+
+If you want to let tcpdump analyze the captured traffic directly, you can use the `pcap-ng` output mode and pipe IGs
```
mauriciovasquezbernal: IGs -> IG's

gadgets/tcpdump/program.bpf.c
mauriciovasquezbernal, line 1: nit: please add the license header.

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+static __always_inline int handle(struct __sk_buff *skb, __u8 packet_type)
+{
+	struct gadget_socket_value *skb_val = gadget_socket_lookup(skb);
+	if (gadget_should_discard_data_by_skb(skb_val))
+		return 0;
+	if (!gadget_pf_matches((void *)skb, (void *)(long)skb->data,
+			       (void *)(long)skb->data_end))
+		return 0;
+
+	void *packet = NULL;
```
mauriciovasquezbernal: I think you can use directly `struct packet_event_t` for this one. Would it require another cast below?

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+	struct packet_event_t *event;
+
+	packet = gadget_reserve_buf(&packets, sizeof(*event) + snaplen);
+	if (!packet)
+		return 0;
+	event = (struct packet_event_t *)packet;
+	if (!event)
+		return 0;
```
mauriciovasquezbernal: I don't think this extra check is required, at this point we're sure packet is not NULL and c casting doesn't do any extra check, so the result will be always not NULL.

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+	__u64 len = skb->len;
+	event->packet_size = len;
+
+	if (len > snaplen)
+		len = snaplen;
+
+	if (len > 0) {
+		u8 *data = (u8 *)(event + 1);
+		bpf_probe_read_kernel(data, len, (void *)(long)skb->data);
+	}
```
mauriciovasquezbernal: Any preference by doing this or using the perf event output as in trace_dns?

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+
+	event->payload_len = len;
+	event->timestamp_raw = bpf_ktime_get_ns();
+	event->ifindex = skb->ifindex;
+	event->packet_type = packet_type;
+
+	if (skb_val != NULL) {
+		// Enrich event with process metadata
+		gadget_process_populate_from_socket(skb_val, &event->proc);
+	}
```
mauriciovasquezbernal: nit: gadget_process_populate_from_socket already checks for NULL, so you can call it without the check.

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+
+GADGET_PF();
+
+static __always_inline int handle(struct __sk_buff *skb, __u8 packet_type)
+{
+	struct gadget_socket_value *skb_val = gadget_socket_lookup(skb);
+	if (gadget_should_discard_data_by_skb(skb_val))
+		return 0;
+	if (!gadget_pf_matches((void *)skb, (void *)(long)skb->data,
+			       (void *)(long)skb->data_end))
```
mauriciovasquezbernal, line 56: I'd say it's better to perform this check first as it's faster. The socket enricher lookup is a bit expensive.

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+		return 0;
+
+	void *packet = NULL;
+	struct packet_event_t *event;
+
+	packet = gadget_reserve_buf(&packets, sizeof(*event) + snaplen);
```
mauriciovasquezbernal: nit: If we decide to go with ringbuf, can we reserve it according to the real length `skb->len`?

gadgets/tcpdump/program.bpf.c
```diff
@@ -0,0 +1,102 @@
...
+	event->timestamp_raw = bpf_ktime_get_ns();
+	event->ifindex = skb->ifindex;
+	event->packet_type = packet_type;
+
+	if (skb_val != NULL) {
+		// Enrich event with process metadata
+		gadget_process_populate_from_socket(skb_val, &event->proc);
+	}
+
+	gadget_submit_buf(NULL, NULL, packet, 0);
```
mauriciovasquezbernal: Don't we need to pass the context and the `packets` map? 
Or better, would this logic work if it fallbacks to use perf_ring_buffers?

gadgets/tcpdump/README.mdx
mauriciovasquezbernal, line 1: Please add README.md and create the symlink in docs to README.mdx.

gadgets/tcpdump/README.mdx
```diff
@@ -0,0 +1,80 @@
...
+output directly to tcpdump like so:
+
+```bash
+$ ig run --verify-image=false tcpdump:mine --host --pf-main "port 80" -o pcap-ng | tcpdump -nvr -
```
mauriciovasquezbernal: sudo ig run tcpdump:%IG_TAG% ...

mauriciovasquezbernal: `--pf-main` doesn't work for me, should it be `--pf`? Also, is `--pf` good enough?

gadgets/tcpdump/gadget.yaml
```diff
@@ -0,0 +1,27 @@
+name: tcpdump
+description: Receive packets from hosts
```
mauriciovasquezbernal:
```suggestion
description: Capture network packets
```

?

gadgets/tcpdump/README.mdx
mauriciovasquezbernal: Same for host, I got a `WARN[0001] start tracing container "": getting network namespace of pid 1: getting interface by index: route ip+net: no such network interface` warning and them not packets from the host are captured. Perhaps it's specific to my system?

pkg/operators/ebpf/ebpf.go
```diff
@@ -376,27 +378,67 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
 				return fmt.Errorf("invalid section name %q", p.SectionName)
 			}
-			if parts[0] != "classifier" {
-				return fmt.Errorf("invalid section name %q", p.SectionName)
-			}
 
-			var direction tchandler.AttachmentDirection
+			lastIdx := -1
+		scanPacketFilter:
+			for idx, inst := range p.Instructions {
```
mauriciovasquezbernal: What about using `for idx := lastIdx; idx < len(p.Instructions); idx++` to avoid the strange conditional below?

pkg/operators/ebpf/ebpf.go
```diff
@@ -376,27 +378,67 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
+					insns[injectIdx] = insns[injectIdx].WithMetadata(asm.Metadata{})
+					p.Instructions = append(insns[:injectIdx], append(filterInsns, insns[injectIdx:]...)...)
 
-			handler, err := tchandler.NewHandler(direction)
-			if err != nil {
-				return fmt.Errorf("creating tc network tracer: %w", err)
+					lastIdx = injectIdx
+
+					// Scan from the top
+					goto scanPacketFilter
```
mauriciovasquezbernal: Why do we need to scan from the beginning?

pkg/operators/ebpf/ebpf.go
```diff
@@ -376,27 +378,67 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
+				}
+
+				handler, err := tchandler.NewHandler(direction)
+				if err != nil {
+					i.Close(gadgetCtx)
+					return fmt.Errorf("creating tc network tracer: %w", err)
+				}
+
+				i.tcHandlers[p.Name] = handler
+			case "tcx":
```
mauriciovasquezbernal: 🤔

pkg/operators/cli/clioperator.go
```diff
@@ -459,6 +462,55 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 					return nil
 				}, Priority)
 			}
+		case ModePCAPNG:
```
mauriciovasquezbernal, line 487: Should it check if the output is a terminal and avoid throwing a lot of garbage there?

pkg/pcap-ng/pcapng.go
mauriciovasquezbernal, line 1: Do we really want to write this from scratch? What about using https://pkg.go.dev/github.com/gopacket/gopacket/pcapgo?

If we use that one, I think we can also provide metadata about the process like https://github.com/mozillazg/ptcpdump does:

Image

Which is produced by a code like:

```go
			var commField datasource.FieldAccessor

			procField := ds.GetField("proc")
			if procField != nil {
				commField = ds.GetField("proc.comm")

			}

			pcapw, err := pcapgo.NewNgWriter(os.Stdout, layers.LinkTypeEthernet)
			if err != nil {
				return fmt.Errorf("creating pcapgo.NgWriter: %w", err)
			}

			ds.Subscribe(func(ds datasource.DataSource, data datasource.Data) error {
				mu.Lock()
				defer mu.Unlock()

				//ifIndex, _ := interfaceField.Uint32(data)
				// ts, _ := timestampField.Uint64(data)
				payload, _ := payloadField.Bytes(data)
				length, _ := lengthField.Uint32(data)

				captureInfo := gopacket.CaptureInfo{
					Timestamp:     time.Now(),
					CaptureLength: int(length),
					Length:        int(length),
					//InterfaceIndex: int(ifIndex), // ifIndex,
					//InterfaceIndex: idx, // use the index we added above
				}

				opts := pcapgo.NgPacketOptions{}

				if commField != nil {
					comm, _ := commField.String(data)
					opts.Comments = append(opts.Comments, fmt.Sprintf("comm: %s", comm))
				}

				pcapw.WritePacketWithOptions(captureInfo, payload, opts)

				// need to make sure this gets written before returning buffers
				pcapw.Flush()
				return nil
			}, Priority)
```

Of course, we'll need to understand how to get that in a clean way, perhaps an annotation to indicate who is the proc field?


Edit: I see the implementation here somehow has support for `sectionComment`, but I still think we shouldn't create it from scratch.

pkg/operators/cli/clioperator.go
```diff
@@ -459,6 +462,55 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+				// ifIndex, _ := interfaceField.Uint32(data)
+				// ts, _ := timestampField.Uint64(data)
+				payload, _ := payloadField.Bytes(data)
+				length, _ := lengthField.Uint32(data)
+
+				wr.WritePacket(payload, &pcapng.PacketOptions{
+					Timestamp:      time.Now(),
+					InterfaceIndex: 0,
+					CaptureLength:  int(length),
+					Length:         int(length),
```
mauriciovasquezbernal: These two are different: `cli.pcap-ng.capture-length` vs `cli.pcap-ng.capture-lengt`

gadgets/tcpdump/gadget.yaml
```diff
@@ -0,0 +1,27 @@
...
+      ebpf.rest.len: payload_len
+      ebpf.rest.name: payload
+      cli.supported-output-modes: pcap-ng,columns,json,yaml
+      cli.default-output-mode: columns
+      cli.raw.fields: payload
+      cli.pcap-ng.payload: payload
+      cli.pcap-ng.interface: ifindex
+      cli.pcap-ng.timestamp: timestamp
+      cli.pcap-ng.capture-length: payload_len
+      cli.pcap-ng.length: payload_len
```
mauriciovasquezbernal: `packet_size`?

mauriciovasquezbernal (commented):
It's looking really good, I think it's a great addition to the project! 

Some things that should be done before merging it:
- Can we create some developer documentation?
- We should have some tests for the gadget at least. 

Also, when I run it using the columns mode I get this:

```bash
$ sudo ig run tcpdump --verify-image=false --pf "port 53"
RUNTIME.CONTAINERNAME          PA… IFINDEX    PAYLOAD_L… PACKET_SI… COMM                    PID        TID PAYLOAD
mycontainer                    0   11         74         74         wget                  19119      19119 <74 bytes>
mycontainer                    0   11         74         74         wget                  19119      19119 <74 bytes>
mycontainer                    1   11         90         90         wget                  19119      19119 <90 bytes>
mycontainer                    1   11         102        102        wget                  19119      19119 <102 bytes>
mycontainer                    0   11         74         74         wget                  19121      19121 <74 bytes>
mycontainer                    0   11         74         74         wget                  19121      19121 <74 bytes>
mycontainer                    1   11         90         90         wget                  19121      19121 <90 bytes>
mycontainer                    1   11         102        102        wget                  19121      19121 <102 bytes>
```

Do you think there's an easy way to give a basic string representation of the payload?

mauriciovasquezbernal:
Ah I almost forgot, the support is only available for TC programs, should it be extended for packet filters one too?

gadgets/tcpdump/program.bpf.c
flyth: Could you please clarify this?

pkg/operators/cli/clioperator.go
flyth: I'm fine printing binary here. Other tools also don't care and I more fear there might be a situation where a user actually wants the binary output and that check would prevent it somehow.

pkg/pcap-ng/pcapng.go
flyth: The existing libraries were all limited in ways and/or not written in a performant way. The actual spec isn't too complicated, so I think it's fine to use our own implementations, as it also lets us add more functionality in the future more easily.

pkg/libpcap-compiler/ebpf.go
flyth: It refers to the three parameters used on the original function (that is being created by the macro).

gadgets/tcpdump/README.mdx
flyth: > --pf-main doesn't work for me, should it be --pf? Also, is --pf good enough?

Yes, I renamed that; you can use multiple packet filters if you give them names; without a name you can only use one that is then fed by the `--pf` parameter alone. I'll clean that up.

pkg/libpcap-compiler/ebpf.go
mauriciovasquezbernal: Can you please clarify why do we need all these three parameters on that function?

gadgets/tcpdump/program.bpf.c
mauriciovasquezbernal: Sure! `snaplen` is a param set by the user. In this case, we only need to reserve space enough for holding the packet, so using skb->len instead of snaplen should work. (Unless am I missing something and the verifier is not happy about it)

pkg/operators/cli/clioperator.go
mauriciovasquezbernal: I remember once I tried a tool (wget?) and it printed a message to prevent this situation. I'm fine with keeping it if you think it's safe.

pkg/pcap-ng/pcapng.go
mauriciovasquezbernal: Do you have specific examples of features they are missing? Also, are they really a bottleneck for us? I think committing to maintain this code in our repo could cause a big overhead later on, so I'd rather stick to existing libraries unless there is a very powerful reason not to.

pkg/libpcap-compiler/ebpf.go
flyth: The additional skb parameters just reserve some registers for the conversion from cbpf to ebpf, afaiu.

gadgets/tcpdump/program.bpf.c
flyth: Ah, you mean we should use `min(snaplen, skb->len)`, right? I'll try.

pkg/pcap-ng/pcapng.go
flyth: Name resolution data (outside of actual DNS packets), TLS secrets and such things can all go into PCAP-ng files. Interfaces can also include extended information / statistics.

So, if coupled with uprobes for for example openssl, we could make it possible to decipher encrypted data. Using name resolution, we could let the reader resolve containers & services.

flyth: There's also "Custom Blocks" that we might integrate later on to feed arbitrary data into the pcap-ng file as well.

See also [the pcap-ng spec](https://datatracker.ietf.org/doc/draft-ietf-opsawg-pcapng/03/).

flyth:
> Do you think there's an easy way to give a basic string representation of the payload?

We should think about the UX for that a bit, but currently, I think the focus is on providing the capture functionality initially (and with that the "tcpdump" addon).

> Ah I almost forgot, the support is only available for TC programs, should it be extended for packet filters one too?

Yes, IMHO, we should support it for everything that comes with an skb (also XDP, which we currently don't support at all, etc). But to get things rolling, I'd like to keep this PR to a minimum.

gadgets/tcpdump/program.bpf.c
mauriciovasquezbernal: Yeah, or move up the `len` calculation done in L70 and use it.

pkg/pcap-ng/pcapng.go
mauriciovasquezbernal: I'd like to get @alban's opinion on this one.

alban: I wonder about the performance of this custom pkg/pcap-ng/pcapng.go versus github.com/gopacket/gopacket/pcapgo.

Also, it should be possible to both use github.com/gopacket/gopacket/pcapgo but also write directly to the underlying writer when writing custom blocks (unsupported by gopacket), as long as `NgWriter.Flush()` is called before.

That would reduce the amount of code we have to maintain.

alban: Also, the current PR does not make use of extra features (name resolution, TLS secrets, custom blocks). We can start by using gopacket, and switch to something custom when we actually need it. It seems the only API we use is `WritePacketWithOptions`, so it should be easy to migrate when actually needed.

mauriciovasquezbernal: I like Alban's idea! Let's start with this library and check later on if we actually need to migrate.

gadgets/tcpdump/program.bpf.c
flyth: I was under the impression that ring buffer is generally faster (which is true in general) and I also wanted to use the helpers we provide. But it looks like perf events are actually faster in the special situation where you forward packets. I changed it for now to do it the same way trace_dns does.

pkg/pcap-ng/pcapng.go
flyth: Seems like there was a very recent update (1.4.0) that added a lot of functionality that we could use. I removed the lib for now, let's see how far we can get with it.

examples/go.mod
mauriciovasquezbernal: > libpcap (must be installed on the system

That's not true anymore.

pkg/operators/cli/clioperator.go
```diff
@@ -460,6 +480,88 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+				}
+
+				err = wr.WritePacket(gopacket.CaptureInfo{
+					Timestamp:      time.Unix(0, int64(time.Duration(ts)*time.Microsecond)),
+					InterfaceIndex: pcapInterface,
+					CaptureLength:  len(payload),
+					Length:         int(length),
+				}, payload)
+				if err != nil {
+					gadgetCtx.Logger().Warnf("failed to write packet: %v", err)
```
mauriciovasquezbernal, line 601: nit: should it return here?

pkg/operators/cli/clioperator.go
```diff
@@ -460,6 +480,88 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
...
+				ts, _ := timestampField.Uint64(data)
+				payload, _ := payloadField.Bytes(data)
+				length, _ := lengthField.Uint32(data)
+
+				pcapInterface, ok := interfaces[ifIndex]
+				if !ok {
+					var description []string
+					if containerNameField != nil {
+						if tmp, err := containerNameField.String(data); err == nil {
+							description = append(description, fmt.Sprintf("container="+url.QueryEscape(tmp)))
```
mauriciovasquezbernal: nit: sprintf not needed?

mauriciovasquezbernal (commented):
It's looking great! Some minor comments around. I'll be nice to have some testing.

pkg/operators/cli/clioperator.go
flyth: Good question - I think eventually this should return an error that should then remove the subscription for clioperator (e.g. an error from a subscription callback should only happen once).

go.mod
```diff
@@ -30,7 +31,9 @@ require (
 	github.com/moby/moby v28.4.0+incompatible
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/opencontainers/runtime-spec v1.2.1
+	github.com/packetcap/go-pcap v0.0.0-20250723190045-d00b185f30b7
 	github.com/prometheus/client_golang v1.23.2
+	github.com/prometheus/client_golang v1.23.0
```
mauriciovasquezbernal: nit: This commit should remove the line above.

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: This is still pending.

pkg/operators/ebpf/ebpf.go
mauriciovasquezbernal: Same

pkg/operators/ebpf/ebpf.go
```diff
@@ -376,27 +378,66 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
+					direction = tchandler.AttachmentDirectionIngress
+				case "egress":
+					direction = tchandler.AttachmentDirectionEgress
+				default:
+					return fmt.Errorf("unsupported hook type %q", parts[1])
+				}
+
+				handler, err := tchandler.NewHandler(direction)
+				if err != nil {
+					i.Close(gadgetCtx)
```
mauriciovasquezbernal: This call isn't needed anymore. (probably a leftover after some rebasing)

pkg/operators/ebpf/ebpf.go
```diff
@@ -541,6 +582,33 @@ func (i *ebpfInstance) ExtraParams(gadgetCtx operators.GadgetContext) api.Params
 
 	// MapIter params
 	res = append(res, i.mapParams()...)
+
+	// Iterate over programs
+	filters := make(map[string]struct{})
+	for programName, program := range i.collectionSpec.Programs {
+		// check for packet filters
+		for _, inst := range program.Instructions {
+			if name, ok := strings.CutPrefix(inst.Symbol(), packetFilterPrefix); ok {
```
mauriciovasquezbernal: perhaps an "early continue" can help to reduce the indentation a bit, but it's totally a style thing, so up to you.

```go
			name, ok := strings.CutPrefix(inst.Symbol(), packetFilterPrefix)
			if !ok {
				continue
			}
			if _, ok := filters[name]; ok {
				continue
			}
```

pkg/operators/ebpf/ebpf.go
```diff
@@ -376,27 +378,66 @@ func (i *ebpfInstance) init(gadgetCtx operators.GadgetContext) error {
...
 
-			var direction tchandler.AttachmentDirection
+			lastIdx := -1
+		scanPacketFilter:
+			for idx, inst := range p.Instructions {
+				if idx <= lastIdx {
+					continue
+				}
+				if name, ok := strings.CutPrefix(inst.Symbol(), packetFilterPrefix); ok {
+					filter := i.paramValues[packetFilterParam+"-"+name]
```
mauriciovasquezbernal: perhaps a small `getFilterParam(name string)` helper would reduce duplication of this logic around.

pkg/operators/cli/clioperator.go
```diff
@@ -460,6 +484,128 @@ func (o *cliOperatorInstance) PreStart(gadgetCtx operators.GadgetContext) error
 					return nil
 				}, Priority)
 			}
+		case ModePCAPNG:
+			// Check ds for compatiblity
+			payloadField := ds.GetField(ds.Annotations()[AnnotationPCAPPayload])
+			if payloadField == nil {
+				return fmt.Errorf("payload field not found")
```
mauriciovasquezbernal: nit: should we add the value of the annotation to the error message?

mqasimsarfraz:
I am having issues tracing container started after gadget run:

```
$ sudo -E ig run tcpdump:michael-gadgets-tcpdump-new
WARN[0000] gadget signature verification is disabled due to using corresponding option 
WARN[0000] gadget signature verification is disabled due to using corresponding option 
RUNTIME.CONTAINERNAME                                                                                                                                                                                                  PA… IFINDEX    PAYLOAD_L… PACKET_SI… COMM                    PID        TID PAYLOAD                                                                                                           
WARN[0009] start tracing container "": getting network namespace of pid 267845: no interface found 
```

I get the "no interface found" warning when starting a container in another terminal and no traffic is captured.

Everything else works great, awesome job! 

```
 ./kubectl-gadget run ghcr.io/mqasimsarfraz/tcpdump:latest -n kube-system -o pcap-ng | tcpdump -nvr -
reading from file -, link-type EN10MB (Ethernet), snapshot length 262144
01:00:00.000000 IP (tos 0x0, ttl 64, id 6590, offset 0, flags [DF], proto TCP (6), length 60)
    10.244.0.1.50848 > 10.244.0.8.8181: Flags [S], cksum 0x161f (incorrect -> 0xc569), seq 1369090411, win 64240, options [mss 1460,sackOK,TS val 4273508704 ecr 0,nop,wscale 7], length 0
01:00:00.000000 IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 60)
    10.244.0.8.8181 > 10.244.0.1.50848: Flags [S.], cksum 0x161f (incorrect -> 0xa1f1), seq 3241063482, ack 1369090412, win 65160, options [mss 1460,sackOK,TS val 2635468111 ecr 4273508704,nop,wscale 7], length 0
01:00:00.000000 IP (tos 0x0, ttl 64, id 6591, offset 0, flags [DF], proto TCP (6), length 52)
    10.244.0.1.50848 > 10.244.0.8.8181: Flags [.], cksum 0x1617 (incorrect -> 0xcd50), ack 1, win 502, options [nop,nop,TS val 4273508704 ecr 2635468111], length 0
01:00:00.000000 IP (tos 0x0, ttl 64, id 6592, offset 0, flags [DF], proto TCP (6), length 159)
    10.244.0.1.50848 > 10.244.0.8.8181: Flags [P.], cksum 0x1682 (incorrect -> 0x0c04), seq 1:108, ack 1, win 502, options [nop,nop,TS val 4273508704 ecr 2635468111], length 107
01:00:00.000000 IP (tos 0x0, ttl 64, id 60690, offset 0, flags [DF], proto TCP (6), length 52)
    10.244.0.8.8181 > 10.244.0.1.50848: Flags [.], cksum 0x1617 (incorrect -> 0xccde), ack 108, win 509, options [nop,nop,TS val 2635468111 ecr 4273508704], length 0
01:00:00.000000 IP (tos 0x0, ttl 64, id 60691, offset 0, flags [DF], proto TCP (6), length 189)
    10.244.0.8.8181 > 10.244.0.1.50848: Flags [P.], cksum 0x16a0 (incorrect -> 0x19e7), seq 1:138, ack 108, win 509, options [nop,nop,TS val 2635468111 ecr 4273508704], length 137
01:00:00.000000 IP (tos 0x0, ttl 64, id 6593, offset 0, flags [DF], proto TCP (6), length 52)
    10.244.0.1.50848 > 10.244.0.8.8181: Flags [.], cksum 0x1617 (incorrect -> 0xcc5d), ack 138, win 501, options [nop,nop,TS val 4273508704 ecr 2635468111], length 0
01:00:00.000000 IP (tos 0x0, ttl 64, id 60692, offset 0, flags [DF], proto TCP (6), length 52)
    10.244.0.8.8181 > 10.244.0.1.50848: Flags [F.], cksum 0x1617 (incorrect -> 0xcc54), seq 138, ack 108, win 509, options [nop,nop,TS val 2635468111 ecr 4273508704], length 0
01:00:00.000000 IP (tos 0x0, ttl 64, id 6594, offset 0, flags [DF], proto TCP (6), length 52)
    10.244.0.1.50848 > 10.244.0.8.8181: Flags [F.], cksum 0x1617 (incorrect -> 0xcc5b), seq 108, ack 139, win 501, options [nop,nop,TS val 4273508704 ecr 2635468111], length 0

```

mauriciovasquezbernal:
@mqasimsarfraz how did you test it? The following works for me:

terminal 1:
```bash
$ sudo ig run tcpdump:foo --verify-image=false
WARN[0000] gadget signature verification is disabled due to using corresponding option
WARN[0000] gadget signature verification is disabled due to using corresponding option
RUNTIME.CONTAINERNAME          PA… IFINDEX    PAYLOAD_L… PACKET_SI… COMM 
```

terminal 2:

```bash
$ docker run --rm -it busybox sh -c "nslookup google.com"
Server:         192.168.78.1
Address:        192.168.78.1:53

Non-authoritative answer:
Name:   google.com
Address: 2800:3f0:4005:41c::200e

Non-authoritative answer:
Name:   google.com
Address: 172.217.162.142

$ docker run --rm -it busybox sh -c "nslookup google.com"
Server:         192.168.78.1
Address:        192.168.78.1:53

Non-authoritative answer:
Name:   google.com
Address: 172.217.162.142

Non-authoritative answer:
Name:   google.com
Address: 2800:3f0:4005:41c::200e
```

data is captured in terminal 1:

```bash
serene_jennings                0   16         70         70         nslookup              92747      92747 <70 bytes>
serene_jennings                0   16         70         70         nslookup              92747      92747 <70 bytes>
serene_jennings                1   16         98         98         nslookup              92747      92747 <98 bytes>
serene_jennings                1   16         86         86         nslookup              92747      92747 <86 bytes>
unruffled_ramanujan            0   18         70         70         nslookup              92843      92843 <70 bytes>
unruffled_ramanujan            0   18         70         70         nslookup              92843      92843 <70 bytes>
unruffled_ramanujan            1   18         86         86         nslookup              92843      92843 <86 bytes>
unruffled_ramanujan            1   18         98         98         nslookup              92843      92843 <98 bytes>
```

mqasimsarfraz:
> how did you test it?

I am using the same steps as you mentioned:

```
RUNTIME.CONTAINERNAME                                                  PA… IFINDEX    PAYLOAD_L… PACKET_SI… COMM                    PID        TID PAYLOAD                              
DEBU[0000] starting op "oci"                            
DEBU[0000] starting ebpfInstance                        
DEBU[0000] setting param value "targ_tid" = 0           
DEBU[0000] setting param value "targ_uid" = 4294967295  
DEBU[0000] setting param value "targ_gid" = 4294967295  
DEBU[0000] handling string param: "targ_comm"           
DEBU[0000] setting param value "targ_comm" = [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 
DEBU[0000] setting param value "snaplen" = 1500         
DEBU[0000] setting param value "targ_pid" = 0           
DEBU[0000] using socket enricher BTF spec from context  
DEBU[0000] got var "gadget_mntns_filter_map": Hash(mntnsset_18cea94e-784a-41e6-ab34-a0cbe1b76a6a)#38 
DEBU[0000] got var "gadget_sockets": Hash(gadget_sockets)#41 
DEBU[0000] got var "gadget_filter_by_mntns": true       
DEBU[0000] setting var "gadget_filter_by_mntns" to true 
DEBU[0000] creating ebpf collection                     
DEBU[0000] starting tracer "packets"                    
DEBU[0000] creating perf reader for map "packets"       
DEBU[0000] Attaching sched_cls "ingress_main"           
DEBU[0000] Attaching sched_cls "egress_main"            
DEBU[0000] starting op "LocalManager"                   
DEBU[0000] starting op "formatters"                     
DEBU[0000] starting op "env"                            
DEBU[0000] starting op "UidGidResolverInstance"         
DEBU[0000] starting op "SocketEnricherInstance"         
DEBU[0000] starting op "filter"                         
DEBU[0000] starting op "otel-metrics"                   
DEBU[0000] starting op "cli"                            
DEBU[0000] running...                                   
DEBU[0003] fanotify: got event with mntns=4026531841 pid=1103523 caller="containerd-shim" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94/log.json --log-format json --systemd-cgroup create --bundle /run/containerd/io.containerd.runtime.v2.task/moby/5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94 --pid-file /run/containerd/io.containerd.runtime.v2.task/moby/5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94/init.pid --console-socket /tmp/pty946670078/pty.sock 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94] 
DEBU[0003] PRECREATE: 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94 
DEBU[0003] fanotify: got event with mntns=4026531841 pid=1103532 caller="runc" callee="7" path=/runc args=[/proc/self/fd/7 init] 
DEBU[0003] OCIConfig enricher: failed to initialize annotation resolver: unsupported container runtime 
DEBU[0003] CREATED: 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94 
DEBU[0003] calling gadget.AttachContainer()             
WARN[0003] start tracing container "": getting network namespace of pid 1103535: no interface found 
DEBU[0003] fanotify: got event with mntns=4026531841 pid=1103562 caller="containerd-shim" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94/log.json --log-format json --systemd-cgroup start 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94] 
DEBU[0003] fanotify: got event with mntns=4026531841 pid=1103568 caller="containerd-shim" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94/log.json --log-format json --systemd-cgroup delete 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94] 
DEBU[0003] fanotify: got event with mntns=4026531841 pid=1103581 caller="containerd-shim" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94/log.json --log-format json delete --force 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94] 
DEBU[0004] DELETED: 5cb29472d303ef37a10a126a63b0bf747ee2e253e32a98437ab0727500dddf94 
DEBU[0004] calling gadget.DetachContainer()             
WARN[0004] stop tracing container "": pid 1103535 is not attached 
DEBU[0028] fanotify: got event with mntns=4026531841 pid=1104144 caller="dockerd" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --version] 
DEBU[0028] fanotify: got event with mntns=4026531841 pid=1104145 caller="containerd-shim" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/00264e274894ae65221d1c44596ff769b4d14ed91d04c203e098d2a096af895b/log.json --log-format json --systemd-cgroup exec --process /tmp/runc-process961782736 --detach --pid-file /run/containerd/io.containerd.runtime.v2.task/moby/00264e274894ae65221d1c44596ff769b4d14ed91d04c203e098d2a096af895b/96d0c79f1963da792bed9228dfd7cd08466e4e8a3d6a5a2e087defbcfb59927f.pid 00264e274894ae65221d1c44596ff769b4d14ed91d04c203e098d2a096af895b] 
DEBU[0028] fanotify: got event with mntns=4026531841 pid=1104151 caller="containerd-shim" callee="runc" path=/usr/bin/runc args=[/usr/bin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/b83077526194087dfaf47b948c04ac36324c16265f6995739ff0a3fdd2263977/log.json --log-format json --systemd-cgroup exec --process /tmp/runc-process2464064895 --detach --pid-file /run/containerd/io.containerd.runtime.v2.task/moby/b83077526194087dfaf47b948c04ac36324c16265f6995739ff0a3fdd2263977/da478ae00769ca39a82673bd4f1972977d71bd0a62a73c5ec4e720ffefe209b7.pid b83077526194087dfaf47b948c04ac36324c16265f6995739ff0a3fdd2263977] 
DEBU[0028] fanotify: got event with mntns=4026531841 pid=1104176 caller="runc" callee="6" path=/runc args=[/proc/self/fd/6 init] 
DEBU[0028] fanotify: got event with mntns=4026531841 pid=1104178 caller="runc" callee="6" path=/runc args=[/proc/self/fd/6 init] 

```

How does the debug log look for you?

mauriciovasquezbernal:
Interesting:

```bash
RUNTIME.CONTAINERNAME          PA… IFINDEX    PAYLOAD_L… PACKET_SI… COMM                    PID        TID PAYLOAD
DEBU[0001] starting op "oci"
DEBU[0001] starting ebpfInstance
DEBU[0001] setting param value "targ_pid" = 0
DEBU[0001] setting param value "targ_tid" = 0
DEBU[0001] setting param value "targ_uid" = 4294967295
DEBU[0001] setting param value "targ_gid" = 4294967295
DEBU[0001] handling string param: "targ_comm"
DEBU[0001] setting param value "targ_comm" = [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
DEBU[0001] setting param value "snaplen" = 1500
DEBU[0001] using socket enricher BTF spec from context
DEBU[0001] got var "gadget_mntns_filter_map": Hash(mntnsset_c57def76-3b1a-4696-ba14-caf4ce06a7b5)#31
DEBU[0001] got var "gadget_sockets": Hash(gadget_sockets)#34
DEBU[0001] got var "gadget_filter_by_mntns": true
DEBU[0001] setting var "gadget_filter_by_mntns" to true
DEBU[0001] creating ebpf collection
DEBU[0001] starting tracer "packets"
DEBU[0001] creating perf reader for map "packets"
DEBU[0001] Attaching sched_cls "ingress_main"
DEBU[0001] Attaching sched_cls "egress_main"
DEBU[0001] starting op "LocalManager"
DEBU[0001] starting op "formatters"
DEBU[0001] starting op "env"
DEBU[0001] starting op "UidGidResolverInstance"
DEBU[0001] starting op "SocketEnricherInstance"
DEBU[0001] starting op "filter"
DEBU[0001] starting op "otel-metrics"
DEBU[0001] starting op "cli"
DEBU[0001] running...
DEBU[0002] fanotify: got event with mntns=4026531841 pid=102132 caller="containerd-shim" callee="runc" path=/usr/sbin/runc args=[/usr/sbin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3/log.json --log-format json --systemd-cgroup create --bundle /run/containerd/io.containerd.runtime.v2.task/moby/83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3 --pid-file /run/containerd/io.containerd.runtime.v2.task/moby/83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3/init.pid --console-socket /tmp/pty1256194508/pty.sock 83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3]
DEBU[0002] PRECREATE: 83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3
DEBU[0002] fanotify: got event with mntns=4026531841 pid=102142 caller="runc" callee="7" path=/runc args=[/proc/self/fd/7 init]
DEBU[0003] OCIConfig enricher: failed to initialize annotation resolver: unsupported container runtime
DEBU[0003] CREATED: 83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3
DEBU[0003] calling gadget.AttachContainer()
DEBU[0003] tracer attached: container "" pid 102144 mntns 4026533057 netns 4026533062
DEBU[0003] fanotify: got event with mntns=4026531841 pid=102164 caller="containerd-shim" callee="runc" path=/usr/sbin/runc args=[/usr/sbin/runc --root /var/run/docker/runtime-runc/moby --log /run/containerd/io.containerd.runtime.v2.task/moby/83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3/log.json --log-format json --systemd-cgroup start 83ebad4805f8513cebeefb091c6702dc2aae01a24230a9d0d81938a96126a6f3]
romantic_nightingale           0   20         70         70         nslookup             102144     102144 <70 bytes>
romantic_nightingale           0   20         70         70         nslookup             102144     102144 <70 bytes>
romantic_nightingale           1   20         86         86         nslookup             102144     102144 <86 bytes>
romantic_nightingale           1   20         98         98         nslookup             102144     102144 <98 bytes>
```

Does trace_dns work for you on the same situation? (on main for this branch?)

mqasimsarfraz:
> Does trace_dns work for you on the same situation? (on main for this branch?)

Yes, that works fine for me. 

> interesting:

I couldn't reproduce in `minikube` node so I wonder if this has to do something with my `docker` installation. 

Seems an issue [only for tc programs](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/tchandler/tracer.go#L211). I manually printed the links and it only shows loopback interface:

```
link: &{LinkAttrs:{Index:1 MTU:65536 TxQLen:1000 Name:lo HardwareAddr: Flags:up|loopback|running RawFlags:65609 ParentIndex:0 MasterIndex:0 Namespace:<nil> Alias: AltNames:[] Statistics:0xc0005fa240 Promisc:0 Allmulti:0 Multi:0 Xdp:0xc001cfa858 EncapType:loopback Protinfo:<nil> OperState:unknown PhysSwitchID:0 NetNsID:-1 NumTxQueues:1 NumRxQueues:1 TSOMaxSegs:65535 TSOMaxSize:524280 GSOMaxSegs:65535 GSOMaxSize:65536 GROMaxSize:65536 GSOIPv4MaxSize:65536 GROIPv4MaxSize:65536 Vfs:[] Group:0 PermHWAddr: ParentDev: ParentDevBus: Slave:<nil>}}
WARN[0274] start tracing container "": getting network namespace of pid 1214518: no interface found
```

nit: Wrapped message seem to be wrong, should be "getting network interfaces" instead of "getting network namespace of pid"

Edit: `nsenter` confirms that the network namespace seem to having only loopback interface:

```
→ sudo nsenter -t 1255890  -n ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
```

mqasimsarfraz:
Not sure if we should block this PR on above. I am fine opening a separate issues for it and track it there?

mauriciovasquezbernal:
Any chance you can try out another TC gadget like https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/ci/sched_cls_drop on main to see if this issue was introduced by this PR? (I don't think so)

mqasimsarfraz:
> Any chance you can try out another TC gadget like https://github.com/inspektor-gadget/inspektor-gadget/tree/main/gadgets/ci/sched_cls_drop on main to see if this issue was introduced by this PR? (I don't think so)

Fails on main as well with `ci/sched_cls_drop` so it isn't related to this PR. Doing some testing I found, it has something to do with `docker-ce` version. If I downgrade to `<=v0.27.0` the issue is gone but as soon as I upgrade the docker (server) version to `>=v0.28.0` it starts to fail. From `v28` [release notes](https://docs.docker.com/engine/release-notes/28/#2800):

> The deprecated OCI prestart hook is now only used by build containers. For other containers, network interfaces are added to the network namespace after task creation is complete, before the container task is started.

Not sure if above is the root cause but let me open a separate issue for it.

mqasimsarfraz:
> let me open a separate issue for it.

https://github.com/inspektor-gadget/inspektor-gadget/issues/4948

flyth:
> LGTM, great!
> 
> There are still a couple of open nits:
> 
> * [#4567 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/pull/4567#discussion_r2372843621)
> * [#4567 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/pull/4567#discussion_r2279765996)
> * [#4567 (comment)](https://github.com/inspektor-gadget/inspektor-gadget/pull/4567#discussion_r2279763355)
> 
> but we can go ahead if you thin they are not relevant.

@mauriciovasquezbernal Addressed two of the remaining ones, will fix the last one in a future PR.

Thanks for the reviews! Also thanks for creating the follow up issue @mqasimsarfraz. I hope in the future the gadget will support different modes that we could easily switch to in such cases.

Issue #4586, opened 2025-06-08
[RFE] Feature Request: Add Node Field to BuildCommonData Helper Functions


amruth6002:
# Current situation

Currently, the BuildCommonData functions in both helpers.go and utils.go lack proper Node field initialization. There's a TODO comment in the code indicating that the Node field should be included, but it hasn't been implemented.

# Impact

Without the Node field being properly initialized:

Test data is incomplete when using these helper functions
Tests may not properly simulate real-world Kubernetes environments where node information is important
Any code that depends on the Node field being set might behave differently in tests than in production
Ideal future situation
Both helper functions should:

Initialize the Node field with a default value of "test-node"
Provide a WithNode helper function to allow overriding the default value
This creates more complete test data that better represents a real Kubernetes environment and satisfies the existing TODO comments.

# Implementation options

Option 1: Add default Node name and WithNode function
Update helpers.go
Set Node: "test-node" in the BuildCommonData function
Add a properly documented WithNode function
Update utils.go

Option 2: Use actual node name from test environment
Instead of using a static "test-node" value, we could use the actual node name from the test environment (if available) or fall back to a default name.

# Additional information

This change improves test case completeness without breaking existing tests, as it only adds additional information to the test data. It's a simple change that eliminates the TODO comments in the code and makes the testing infrastructure more robust.

schnrj:
@amruth6002 can i work on this?

amruth6002:
thank you @schnrj I have already raised  a pr

PR #4587, opened 2025-06-08
Add Node field to BuildCommonData helper functions

amruth6002:
# Add Node field to BuildCommonData helper functions #4586 

This PR adds proper Node field initialization to the BuildCommonData functions in both helpers.go and utils.go. Previously, these functions had a TODO comment indicating that the Node field should be included, but it wasn't implemented.

This change improves test case completeness by providing a default "test-node" value and adding a new WithNode helper function to allow overriding the default value when needed. This creates more complete test data that better represents a real Kubernetes environment.

Reviewers should check that:
The Node field is correctly initialized with "test-node" in both helper functions
The new WithNode  helper function works correctly when customizing the node name
Existing tests continue to run as expected with this change

### DIFF
--- a/integration/helpers.go
+++ b/integration/helpers.go
@@ -66,6 +66,13 @@ func WithPodLabels(podName string, namespace string, enable bool) CommonDataOpti
 	}
 }
 
+// WithNode sets the node name in the common data
+func WithNode(nodeName string) CommonDataOption {
+	return func(commonData *eventtypes.CommonData) {
+		commonData.K8s.Node = nodeName
+	}
+}
+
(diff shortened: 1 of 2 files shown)


amruth6002:
@blanquicet @mqasimsarfraz @alban @mauriciovasquezbernal please take a look

mauriciovasquezbernal (changes requested):
> The Node field is correctly initialized with "test-node" in both helper functions

This is wrong as the node can have different values. You'll need to find a way to query the node name from the API server or so, that's why we didn't add this before.

PR #4597, opened 2025-06-10
trace_exec/unit: Dont use full path

burak-ok:
Depends on which symlink is used for running python3, different values can be found in the Args value. The check is a contains check and /bin/python3 would cover at least /use/bin/python3 and /bin/python3


alban:
Could you add a comment explaining that it accepts both `/use/bin/python3` and `/bin/python3`?

---

But I wonder if we should change the behaviour of the gadget.
cc @Sefi4 @mauriciovasquezbernal 

I tested the trace_exec gadget with a custom app:
```go
package main

import (
	"os"
	"syscall"
)

func main() {
	_ = syscall.Exec(
		"./my-symlink-to-cat",
		[]string{"fake", "/dev/null"},
		os.Environ(),
	)
}
```

```bash
$ ln -s /bin/cat my-symlink-to-cat
$ go build goexecve.go
$ sudo strace -f -e execve ./goexecve
[pid 185600] execve("./my-symlink-to-cat", ["fake", "/dev/null"], 0xc0000aa008 /* 77 vars */ <unfinished ...> 
```

```
$ sudo ig run trace_exec --host -o yaml --paths|grep -E 'args:|exepath:'
args: ./my-symlink-to-cat /dev/null 
exepath: /usr/bin/cat 
```

In the execve syscall, `args[0]` is `"fake"`. But the gadget currently loses this value because it reads from `bpf_probe_read_user_str(event->args, ARGSIZE, pathname);`, see:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_exec/program.bpf.c#L146

The gadget reads the arguments from `args[1]` because of the loop `for (i = 1; i < TOTAL_MAX_ARGS; i++)`:
https://github.com/inspektor-gadget/inspektor-gadget/blob/main/gadgets/trace_exec/program.bpf.c#L157-L158

We would need to add a new field `pathname` for the first parameter of the `execve` syscall instead of putting it in `args[0]`...

alban (approved):
LGTM!

My other comment can be fixed in another PR.

PR #4603, opened 2025-06-10
operators/ebpf: Use a BTF-based solution to register gadgets

mauriciovasquezbernal:
Before the patch the GADGET_TRACER(), GADGET_MAP_ITER(), etc. macros were used to define gadgets in the ebpf code. Those macros internally created a var starting with a given prefix and used the var name to store all the information about the gadget: name, types, etc. Relying on string concatenation has some issues, for instance, extending that to support more information in the future is difficult.

This commit uses an approach similar to the one used to define eBPF maps: defining a structure that generates BTF information. It has the following advantages:
- Easier to extend later on, as adding more information is just a matter of adding fields to the struct
- Uses a well-known approach to define metadata in the program (as used for BTF maps)

### Discussions
- Is it approach worth it?
- Should we do the same for params?


pkg/operators/ebpf/helpers.go
```diff
@@ -89,3 +89,45 @@ func (i *ebpfInstance) validateGlobalConstVoidPtrVar(t btf.Type, varName string)
...
+	}
+
+	return arr.Nelems, nil
+}
+
+// same as above for __type
+func typeFromBTF(typ btf.Type) (btf.Type, error) {
+	vk, ok := typ.(*btf.Pointer)
+	if !ok {
+		return nil, fmt.Errorf("value type is not a pointer: %T", typ)
```
alban, line 113: You can also test for `f, ok := typ.(*btf.Fwd)` and then `spec.TypeByName(f.Name, &ptr)`

gadgets/trace_open/program.bpf.c
```diff
@@ -46,7 +46,14 @@ struct {
 
 GADGET_TRACER_MAP(events, 1024 * 256);
 
-GADGET_TRACER(open, events, event);
+// TODO: We need this for the compiler to emit BTF, without this, the compiler
+// emits a FWD declaration below that messes everything up. Why it doesn't
+// happen for map declarations?
+const struct event *__foo_var_foo __attribute__((unused));
+struct {
```
alban: I think you can support both ways in helpers.go `typeFromBTF`, see comment below.

I have no idea why it does not happen for map declarations.

alban (commented):
I like this approach. As you said, it's better for extensibility, and it is reusing similar pattern as other definitions in bpf code.

It also makes it easier to document the ABI.

I think we should go for it, including for parameters, etc.

About compatibility:
- **Compile-time compatibility**: we could keep the #define for GADGET_TRACER, etc. (but document it as deprecated) so that older gadgets can simply be recompiled and then the ELF file can be parsed with the new method.
- **Run-time compatibility**: we can keep the previous implementation of populateTracer, populateSnapshotters, etc. (but mark it as deprecated and to be removed at a specific date) so that new versions of ig can run old gadgets.

pkg/operators/ebpf/helpers.go
mauriciovasquezbernal: Unfortunately it doesn't work as the BTF info of the structure itself is not present.

gadgets/trace_open/program.bpf.c
mauriciovasquezbernal: > I have no idea why it does not happen for map declarations.

This is because the structure is used as key or value in maps, so the compiler generates information for them. I introduced a macro to force the compiler to emit this information.

mauriciovasquezbernal:
I updated the PR:
- The previous approach is still present and supported, even if it's deprecated and prints some warnings 
- I updated the logic to support params too, but it's using almost the same approach as before: the param name encoded in a variable name. I didn't find any good way to encode the param in a BTF structure. 
- The metadata generation doesn't work after it. I'll try to fix it.

mauriciovasquezbernal:
This is stale and I won't have the time to complete it. Closing it, if somebody wants to explore this further, feel free to base on this.

PR #4613, opened 2025-06-12
 pkg: operators: wasm: Move syscall_helpers functions to pkg/utils/syscalls.

eiffel-fl:
Before 3ba4e5500d1a ("treewide: Remove built-in gadgets.") this package was shared with builtin traceloop.
This no more makes sense today, as builtins were removed. So, let's move this file to proper location and make its functions private.


mauriciovasquezbernal (commented):
What about moving this to a helpers/utils package so it can be used for others in the future?

PR #4617, opened 2025-06-12
tree-wide: Remove legacy code for built-in gadgets

mauriciovasquezbernal:
Continue removing legacy code:
- legacy prometheus
- legacy built-in apis 
- gadgets catalog and registry
- interfaces for old operators 

There are still some things to do:
- Do we really need DeplyInfo now that the gadget catalog is gone? (it's mainly used for the server version)
- Further clean up on operators: Remove GlobalParamDescs, ParamDescs, PreGadgetRun and PostGadgetRun


pkg/operators/prometheus/bucket.go
mqasimsarfraz, line 1: I wonder if we should add a header to https://github.com/inspektor-gadget/inspektor-gadget/blob/main/docs/design/001-prometheus.md to clarify the design doc is outdated?

pkg/operators/kubeipresolver/kubeipresolver.go
mqasimsarfraz, line 1: Everything between L53-L62 should be deleted as well?

pkg/operators/operators.go
```diff
@@ -53,47 +48,10 @@ type GadgetContext interface {
...
 // MapPrefix is used to avoid clash with maps and other eBPF objects when added
 // to gadget context.
 const MapPrefix string = "map/"
 
-type Operator interface {
-	// Name must return a unique name for the operator
-	Name() string
-
-	// Description is an optional description to show to the user
-	Description() string
```
mqasimsarfraz, line 69: I see most of the operator still have `Description()` method, shouldn't it be removed?

pkg/operators/prometheus/bucket.go
mauriciovasquezbernal: Makes sense!

pkg/operators/operators.go
mauriciovasquezbernal: I think having the description could be useful later on, so I'd rather keep it!

mauriciovasquezbernal:
Failures aren't related to this PR, moving on!

PR #4619, opened 2025-06-12
gadgets: Print logs to indicate when they're ready

### LINKED ISSUE #1279: [RFE] kubectl-gadget and local-gadget should provide way to inform when traces have been initialized
Based on the [review](https://github.com/inspektor-gadget/inspektor-gadget/pull/1278#pullrequestreview-1254203781), we should consider adding way to inform when `kubectl-gadget`/`local-gadget` has initialized the traces. This can be useful for integration-tests where we start the traces and need to wait a bit before generating any tests event:

```
testSteps := []TestStep{
		traceDNSCmd,
		SleepForSecondsCommand(2), // wait to ensure local-gadget has started
		&DockerContainer{
			Name: cn,
			Cmd: "nslookup -type=a inspektor-gadget.io. 8.8.4.4 ; " +
				"nslookup -type=aaaa inspektor-gadget.io. 8.8.4.4",
		},
	}
```

### Possible Implementation
Without `-o json`, it is visible when the column titles are displayed. But with -o json, maybe we should add an event of type [READY](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/types/types.go#L99) for all gadgets.

mauriciovasquezbernal:
Some times running a gadget takes a bit long: while the image is pulled,
wasm and ebpf programs are loaded, etc. This commit introduces a mechanism
to inform the user about the different steps when running a gadget.

```bash
$ sudo ig run trace_open --verify-image=false --pull always
INFO[0000] preparing gadget info...
WARN[0004] image signature verification is disabled due to using corresponding option
WARN[0004] This gadget was built with ig v0.41.0 and it's being run with v0.0.0. Gadget could be incompatible
INFO[0004] starting gadget...
WARN[0006] image signature verification is disabled due to using corresponding option
WARN[0006] This gadget was built with ig v0.41.0 and it's being run with v0.0.0. Gadget could be incompatible
RUNTIME.CONTAINERNAME       COMM                  PID        TID  FD FNAME                        MODE           ERROR
INFO[0006] running...
```

### DIFF
--- a/pkg/gadget-context/gadget-context.go
+++ b/pkg/gadget-context/gadget-context.go
@@ -390,7 +390,7 @@ func (c *GadgetContext) LoadGadgetInfo(info *api.GadgetInfo, paramValues api.Par
 		if err := c.start(c.localOperators); err != nil {
 			return fmt.Errorf("starting local operators: %w", err)
 		}
-		c.Logger().Debugf("running...")
+		c.Logger().Infof("running...")
 	}
 
 	if c.ExtraInfo() && extraInfo != nil {
--- a/pkg/gadget-context/run.go
+++ b/pkg/gadget-context/run.go
@@ -160,11 +160,13 @@ func (c *GadgetContext) stop(dataOperatorInstances []operators.DataOperatorInsta
 }
 
 func (c *GadgetContext) PrepareGadgetInfo(paramValues api.ParamValues) error {
+	c.Logger().Info("preparing gadget info...")
 	_, err := c.initAndPrepareOperators(paramValues)
 	return err
 }
 
 func (c *GadgetContext) Run(paramValues api.ParamValues) error {
+	c.Logger().Info("starting gadget...")
 	defer c.cancel()
 
 	metricAttribs := attribute.NewSet(
@@ -183,7 +185,7 @@ func (c *GadgetContext) Run(paramValues api.ParamValues) error {
 		return fmt.Errorf("starting operators: %w", err)
 	}
 
-	c.Logger().Debugf("running...")
+	c.Logger().Infof("running...")
 
 	WaitForTimeoutOrDone(c)
 	c.stop(dataOperatorInstances)
--- a/pkg/runtime/grpc/oci.go
+++ b/pkg/runtime/grpc/oci.go
@@ -55,6 +55,7 @@ func (r *Runtime) GetGadgetInfo(gadgetCtx runtime.GadgetContext, runtimeParams *
 		in.Flags |= api.GadgetInfoRequestFlagUseInstance
 	}
 
+	gadgetCtx.Logger().Info("getting gadget info...")
 	out, err := client.GetGadgetInfo(gadgetCtx.Context(), in)
 	if err != nil {
 		return nil, fmt.Errorf("getting gadget info: %w", err)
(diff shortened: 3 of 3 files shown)


flyth:
I'm not sure if this is the way to go. It creates a lot of unnecessary overhead & complexity, IMHO.
Want to know what's going on behind the scenes (and is probably taking too long), just enable debug mode (and add more debug messages in there, if necessary). For API use we already got the lifecycle hooks that let you know about the state of the gadget.

Regarding the UX, maybe there are better ways to improve it.
On interval based datasources for example, let's not print headers upfront but only when data arrives. Instead - where possible - add a first early run instead of waiting for the first interval.

mauriciovasquezbernal:
> just enable debug mode (and add more debug messages in there, if necessary).

I think debug messages are too noisy, we need something that's shown by default. I changed the level and added some new Info messages. Json support is missing but it can be handled later on.

It's perhaps not a perfect solution, but that's the simplest I could came up with.

blanquicet:
With `ig` the "running" log is printed after the columns, which is not very nice:

```bash
$ go run -exec "sudo -E" ./cmd/ig run ghcr.io/blanquicet/gadget/trace_dns:mauricio-gadget-init-msg --verify-image=false
INFO[0000] preparing gadget info...
INFO[0002] starting gadget...
RUNTIME.CONTAINE… SRC                     DST                     NAMESERV… COMM            PID       TID QR QTYPE     NAME             RCODE    ADDRESSES  LATENCY_NS
INFO[0004] running...
ubuntuFocal       172.17.0.2:60709        10.104.255.100:53       10.104.2… http          26504     26504 Q  A         security.ubuntu…                            0ns
ubuntuFocal       172.17.0.2:60627        10.104.255.100:53       10.104.2… http          26505     26505 Q  A         archive.ubuntu.…                            0ns
ubuntuFocal       10.104.255.100:53       172.17.0.2:60627        10.104.2… http          26505     26505 R  A         archive.ubuntu.… Success  91.189.9…    315.89ms
ubuntuFocal       10.104.255.100:53       172.17.0.2:60709        10.104.2… http          26504     26504 R  A         security.ubuntu… Success  185.125.…    315.90m
```

For kubectl-gadget, isn't it too much to print "starting gadget..." and "running..." for each node. Consider mine is just a 3-node cluster, and look at the output:

```bash
$ ./kubectl-gadget run ghcr.io/blanquicet/gadget/trace_dns:mauricio-gadget-init-msg
INFO[0002] getting gadget info...
INFO[0005] starting gadget...
INFO[0008] aks-nodepool-41788306-vmss000001 | starting gadget...
WARN[0008] aks-nodepool-41788306-vmss000001 | image signature verification is disabled due to using corresponding option
K8S.NODE    K8S.NAMESP… K8S.PODNAME K8S.CONTAI… SRC                    DST                    NAMES… COMM       PID     TID QR QTYPE NAME       RCODE    ADDR… LATENC…
INFO[0008] running...
INFO[0009] aks-nodepool-41788306-vmss000000 | starting gadget...
WARN[0009] aks-nodepool-41788306-vmss000000 | image signature verification is disabled due to using corresponding option
INFO[0010] aks-nodepool-41788306-vmss000001 | running...
INFO[0010] aks-nodepool-41788306-vmss000002 | starting gadget...
WARN[0010] aks-nodepool-41788306-vmss000002 | image signature verification is disabled due to using corresponding option
INFO[0010] aks-nodepool-41788306-vmss000000 | running...
INFO[0011] aks-nodepool-41788306-vmss000002 | running...
```

mauriciovasquezbernal:
This turned to be more complex than anticipated. I agree with the feedback:
- We should print the message once (not depending on the number of nodes)
- Ideally before the columns headers

I have no a ideas how to make it happen without a lot of changes. It seems there is always one corner case:
- What about if the gadget takes longer to start on a node? i.e. what if we receive events from one node before the gadget is started in another one?
- What is considered that a gadget is "started"? eBPF programs are loaded and attached to the hooks?

I'll close this one and let somebody else fix it.

PR #4625, opened 2025-06-16
Optimize cilium/ebpf kallsyms: 27% initialization speedup for trace_dns

alban:
When running a gadget such as trace_dns, several bpf modules will be used:
1. kfilefields
2. container-hook
3. socket-enricher
4. trace_dns

When a bpf module contains tracing or kprobe programs, cilium/ebpf will read `/proc/kallsyms` to find out if the bpf programs have to be attached to a kernel symbol.

With the scenario above, cilium/ebpf will read `/proc/kallsyms` 3 times, which is not efficient.

This patch introduces `ebpf.KallsymsPreLoad()` to pre-load the cache with known symbols from IG internal bpf modules.

cc @mauriciovasquezbernal @burak-ok 

## How to use

See testing below.

## Testing done

The patch adds a printf with the message "Gadget started successfully" to indicate when IG is fully initialized. The expect command interrupts IG when the message is printed and hyperfine measures the time it takes to start IG.

```
hyperfine -w 2 --runs 10 "expect -c 'spawn sudo ig run trace_dns ; expect \"Gadget started successfully\" { puts \"\n--- OK ---\" } timeout { puts \"\n--- Timeout ---\"; exit 1 } eof { puts \"\n--- EOF ---\"; exit 1 }; exit 0'"
```

Before the patch:
```
  Time (mean ± σ):      4.159 s ±  0.323 s    [User: 0.004 s, System: 0.004 s]
  Range (min … max):    3.870 s …  4.740 s    10 runs
```
After the patch:
```
  Time (mean ± σ):      3.036 s ±  0.258 s    [User: 0.009 s, System: 0.005 s]
  Range (min … max):    2.798 s …  3.468 s    10 runs
```

Speedup: 1.123s (**27%**)

## TODO

- [ ] #4619
- [ ] https://github.com/cilium/ebpf/pull/1802
- [ ] Find a way to avoid hardcoding the symbols


alban:
Replaced by:
- https://github.com/inspektor-gadget/inspektor-gadget/pull/4653
- https://github.com/inspektor-gadget/inspektor-gadget/pull/4652

See also discussion in:
- https://github.com/cilium/ebpf/pull/1802#issuecomment-3000806170

